PR #8561, opened 2026-07-25
fix(quality): repair release/v3.8.49 base-red gates

fenix007:
## Summary

`release/v3.8.49` is red at its own tip (`30709255`), so **every** PR against the branch inherits four failing checks regardless of content. This PR repairs what can be repaired from a contributor branch. Four commits, each independent:

| # | Commit | Fixes |
|---|---|---|
| 1 | `test(resilience)` | `Unit Tests fast-path (1/4)` — 3 assertions |
| 2 | `fix(quality)` | `check:db-rules` + `Unit Tests fast-path (2/4)` |
| 3 | `chore(quality)` | `No new ESLint warnings` |
| 4 | `chore(quality)` | `Fast Quality Gates` → `check:file-size` |

Every claim below was measured on a pristine `origin/release/v3.8.49` checkout with an empty working tree, in a worktree separate from the change.

## What was red and why

**1. Three backoff assertions** (`error-classification.test.ts`, `thundering-herd.test.ts`)

`b64361dd2` (#8396 / #8446) capped the scaled connection cooldown at `BACKOFF_CONFIG.max` so a 429/5xx burst cannot black a connection out past the real rate-limit window. Three tests still expected uncapped doubling:

```
expected  [5000, 10000, 20000, 40000, 80000, 160000]   ← 5000 × 32
actual    [5000, 10000, 20000, 40000, 80000, 120000]   ← BACKOFF_CONFIG.max
```

The cap is intended behavior, so the expectations follow it. `newBackoffLevel` clamping is untouched and still asserted.

**2. `compressionDetailNormalizers` not in the allowlist**

```
AssertionError: Unexported db module(s) not in KNOWN_UNEXPORTED: compressionDetailNormalizers
```

`src/lib/db/compressionDetailNormalizers.ts` is imported only by `db/compression.ts` and exists because `compression.ts` is frozen at its file-size cap (#8388). Re-exporting it through `localDb.ts` would advertise the barrel-import anti-pattern Hard Rule #2 forbids, so `INTENTIONALLY_INTERNAL` is the right home. This unblocks both the gate script and its freeze-guard unit test.

**3. Stale ESLint suppression**

`npm run lint:json -- --max-warnings 0` exits 2 with *"There are suppressions left that do not occur anymore"*. Exactly one entry is stale — `tests/unit/proxy-registry.test.ts`, `no-explicit-any` 55 → 54. Regenerated with `--prune-suppressions`; that single count is the whole diff.

**4. Inherited file-size growth**

```
[...]

PR #8558, opened 2026-07-25
fix(sse): estimate tokens from the object and count JSON length without building it (#7847)

MumuTW:
⚠️ **Stacked on #8555.** This branch is cut from `refactor/combo-error-classifiers`, so the diff currently shows **two commits** — #8555s extraction plus this fix. GitHub will not let a fork PR target another fork branch, so the base is `release/v3.8.49`. Merge #8555 first and this collapses to a single commit. It genuinely depends on that PR — see "Why stacked".

Two changes, one root cause: several paths built a full JSON string only to read its `.length`, and one of them silently changed the answer.

## 1. Correctness — the combo fallback trigger over-counted inline images by 40x

`estimateTokens(JSON.stringify(attemptBody))` takes the **string branch** of `estimateTokens`, which is `ceil(length / CHARS_PER_TOKEN)` over the raw JSON. An inline base64 image is then charged as if every character of the data URL were prose. Measured on a 200 KB inline image:

| | tokens |
| --- | ---: |
| via string (before) | **50,039** |
| via object (after) | **1,231** |

That falsely tripped fallback compression on requests nowhere near the context window. It is the same class #8368/#8401 fixed on the request path — this call site was missed, because it stringifies before handing the value over and so never reaches `extractImageTokens`.

Passing the object routes through the image-aware path. **Text-only bodies are unaffected** — verified identical and pinned by a test, so this is a no-op for the common request.

## 2. Allocation — `jsonLength()`

An exact serialized-length walker: same O(n) scan, no string. Used by `estimateTokens`s object branch and by `streamReadinessPolicy` (which runs on **every streaming request** and only ever used `.length`).

Exactness is not optional — every consumer feeds a threshold, and an approximation would silently shift routing and timeout decisions. So it is property-tested against `JSON.stringify` over **4000 generated structures** covering escaping, lone surrogates, omitted values (`undefined`/functions/symbols, which behave differently in objects vs arrays), non-finite numbers, `toJSON`, `Date`, `Map`, cycles and `BigInt`.

Anything outside the plain-JSON subset falls back to `JSON.stringify` for **that subtree only**, so an exotic leaf never forces the multi-megabyte message history back onto the allocating path.

## Honest scoping of the memory win

The string was always **transient** and V8 collects it efficiently, so this is **not** 3 MiB of retained heap. Measured allocation churn over 20 calls on a 3.06 MiB body:

```
BEFORE JSON.stringify().length   3.1 MiB   (0.15 MiB/call)
[...]

PR #8554, opened 2026-07-25
chore(i18n): normalize zh-TW terminology and sync stale README figures

MumuTW:
## Problem

The zh-TW catalog was machine-translated with mainland-habit vocabulary, plus
simplified→traditional conversions that picked the wrong homophone character.
Its README also still advertised v3.7-era figures.

Two distinct defect classes:

1. **Wrong-character conversions** — not words in Chinese at all:
   `上遊` (should be `上游`, upstream), `後臺` (`後台`), `儀錶板` (`儀表板` — the
   Dashboard "board" became a wristwatch).
2. **Mainland-habit vocabulary** — converts to valid traditional characters but
   is not what TW/HK readers use: `默認`→`預設`, `緩存`→`快取`, `模塊`→`模組`,
   `調用`→`呼叫`, `字符串`→`字串`, `全局`→`全域`, `文檔`→`文件`, `響應`→`回應`.

Consistency fixes: `供應商`/`提供商`→`提供者` (`提供者` was already 71% dominant,
and matches the zh-CN glossary canonical), `型別`→`類型` for UI labels,
`不活躍`→`未啟用`.

README figures synced to the English source: 231→290 providers, 17→19 routing
strategies, 1.6B→1.53B free tokens, 50+→90+ free tiers, 11→40+ free forever,
87→104 MCP tools, 30→31 scopes.

## Root cause

The generator's post-translation normalization was a hardcoded list that
duplicated `glossary/zh-CN.json`'s job and was wired **only** into the
deprecated `generate-multilang.mjs` — so the active `run-translation.mjs`
pipeline applied nothing at all.

Worse, its blanket `/代碼/g → 程式碼` rule was actively destructive: `代碼` is the
correct rendering of "code" in `控制代碼` (handle), `語系代碼` (locale code),
`結束代碼` (exit code) and `錯誤代碼` (error code), all of which the catalog uses
today. The next regeneration would have turned them into `控制程式碼`,
`語系程式碼`, etc.

## Fix

`scripts/i18n/glossary-normalize.mjs` is now the single implementation, driven by
`scripts/i18n/glossary/<locale>.json` as the single source of truth, shared by
all three call sites (active pipeline, legacy generator, drift gate) so they
cannot disagree about what canonical means.

Ambiguous terms carry `blockedPrefixes`, which suppresses a match when the
synonym is preceded by a listed character. That is what lets `型別→類型` stay
enforced without mangling `模型別名` (model alias — `型別` sits across its
[...]

PR #8553, opened 2026-07-25
fix(sse): shallow per-target copy for the combo attempt body (#7847)

MumuTW:
The largest of the #7847 reductions, plus a real cross-target leak found on the way. Pairs with #8549 (benchmark) and #8550 (entry clone).

## Numbers

`combo.ts` deep-cloned the request body for every target. On a 3.05 MiB agent request:

| targets | before | after |
| ---: | ---: | ---: |
| 3 | 9.53 MiB (3.12x wire) | ~0.001 MiB |
| 5 | 15.89 MiB (5.19x wire) | ~0.000 MiB |
| 10 | 31.78 MiB (10.39x wire) | ~0.001 MiB |

The deep clone scaled linearly with the target count; the shallow copy is constant and effectively free.

## Why a shallow copy is sufficient

The isolation the clone bought only ever needed to contain **top-level scalar** writes. The complete mutation surface on this path is two assignments:

```
combo.ts     bodyRecord.max_tokens = ...   (reasoning buffer)
chatCore.ts  body.model = model            (Background Task Redirection T41)
```

Nothing mutates the nested payload. `applyCompression` and `injectUniversalHandoffBody` both return new objects — verified empirically, not by reading: probed with a before/after deep-compare and a frozen input, neither touches its argument. So a fresh top-level object per target gives identical isolation while sharing the expensive `messages`/`tools` arrays.

## The leak this found — round-robin was already broken

`handleRoundRobinCombo` already used a shallow copy, but took it **only when the reasoning buffer actually changed `max_tokens`**:

```ts
let attemptBody = body;                       // shared with the caller
if (bufferedMaxTokens !== currentMaxTokens) { // ...only sometimes copied
  attemptBody = { ...bodyRecord, max_tokens: bufferedMaxTokens };
}
```

Every other attempt shared the callers object outright, so `chatCore`s `body.model = model` on one target rewrote the model for the next. The new test reproduces it **on the unmodified code**:

```
not ok 3 - round-robin: a target's in-place write must not leak into the next target
  target 1 received model "mutated-by-openai/gpt-4o-mini" — a previous target's write leaked through
```

`priority` and `fill-first` passed, because the deep clone was covering for them. In production this is a Background Task Redirection firing on one round-robin target and the next target's upstream request carrying the degraded model name. The copy is now unconditional.

I raised this as an unconfirmed risk during assessment; the test settled it. Fixing it here rather than separately because it is the same defect (per-target copy discipline), the same file, and the same test file covers all strategies.

[...]

PR #8580, opened 2026-07-25
chore(quality): rebaseline complexity ratchets for masked base-red

MumuTW:
## Summary

`check:complexity-ratchets` fails on the **pristine** `release/v3.8.49` tip, which takes the whole `Fast Quality Gates` job down for every PR against the branch. This is the step that was hiding behind `check:file-size`: both run in the same `bash -e` job, file-size ran first, and #8561 repairing it surfaced the one behind.

Same class and same remedy as `_rebaseline_2026_07_25_v3849_basered_filesize` (#8561) and `_rebaseline_2026_07_02_5798_release_green` — inherited growth from already-merged PRs, with no offending PR branch left to fix.

Two lines of JSON. No production code, no test changes.

## Measurement

Every number below measured with `npm run check:complexity-ratchets` on **detached checkouts with an empty working tree**, nothing of this branch applied:

| ref | complexity | cognitive |
| --- | --- | --- |
| `4053e2314` — current tip, post-#8561 | 2169 | 956 |
| `30709255c` — the pre-#8561 base | 2169 | 956 |
| `852bf4e0b` — **the commit that recorded the 2130 / 951 ceilings** | 2167 | 956 |
| baseline as committed | **2130** | **951** |

The third row is the point: **neither ceiling was ever a measurement of this branch.**

- `2130` comes from `_rebaseline_2026_07_20_owner_night_drain`, which measured **2068** on a 9-PR probe tree on 2026-07-20 and took +62 of headroom. Three days later, at the commit that carried that note into this branch, the real count was already 2167.
- For cognitive, the documented rebaseline chain stops at 900 (`_rebaseline_2026_07_19_v3849_fix_sweep_cluster`) plus `_note_8266` (+1). The step to 951 carries no measurement note, and the measured value has been 956 since at least 2026-07-23.

So the gate has been red on this branch for days, not hours — it was simply unreachable behind `check:file-size`.

## Attribution of the drift

Only **+2 cyclomatic** accrued across the 94 commits in `852bf4e0b..4053e2314`, and **cognitive is flat**. Attributed per-file by dumping `getComplexityEslintReport()` at both ends and diffing the violation counts:

| file | cyclomatic | from |
| --- | --- | --- |
| `src/app/(dashboard)/home/ProviderTopology.tsx` | 1 → 2 | #8428 (last-routed-node health visual), on top of #8357 (custom `provider_nodes`) |
| `src/lib/combos/invariants.ts` | 0 → 1 | #8304 (provider / model-family invariants) |
| `src/lib/system/versionCheck.ts` | 0 → 1 | #8301 (singleflight version lookups) |
| `src/app/api/v1/web/fetch/route.ts` | 1 → **0** | #8335 (quota-aware fallback) — a real improvement |

| file | cognitive | from |
| --- | --- | --- |
| `open-sse/executors/gemini-business.ts` | 0 → 1 | #8485 (call the real abort-signal helper) |
| `src/app/api/v1/web/fetch/route.ts` | 1 → **0** | #8335 |

**No open PR contributes.** The 9 open decomposition / #7847 PRs (#8545, #8546, #8547, #8548, #8549, #8550, #8553, #8554, #8558) each measure 2169 / 956 — identical to the base, so they are complexity-net-zero.

## Choice of value

Frozen at the **measured** 2169 / 956 with **zero added headroom**, deliberately unlike the 07-20 note that took +62. A ceiling with slack in it is how this gate ended up 39 units behind reality without anyone noticing; a ceiling pinned to the measurement means the next genuine regression trips it. That does mean the next feature PR that adds a violation has to either extract or bump its own entry — which is the intended behaviour of a ratchet.

Raising a ceiling has historically wanted owner sign-off (see the 07-19 / 07-20 notes). Flagging that explicitly: this is not own-growth, it is a base-red unblock with the measurement above, but if you'd rather clear it by structural extraction of the four files instead of moving the ceiling, say so and I'll close this.

## Verification

- `npm run check:complexity-ratchets` on this branch: `[complexity] OK — 2169 (baseline 2169)`, `[cognitive-complexity] OK — 956 (baseline 956)`.
- Both JSON files parse and are Prettier-clean.
- No production code touched, so no test changes are required (same shape as #8561).

Structural shrink stays tracked in #3501.


MumuTW:
Correcting my own framing on this PR, because a second look found the design argument against it — and it's a good one.

**First, an important thing I missed when I opened this.** `nightly-release-green.yml` runs `validate-release-green.mjs` on every push to `release/v*`, and it had already published every number in this PR — issue #8540, opened today at 08:02Z, lists `Cognitive complexity 956 vs baseline 951`, `complexity=2169 REGRESSED`, and `providers/page.tsx 1990 > 1927`. I measured all of it independently on pristine checkouts without reading that issue first. The sensor works; what's missing is only a pointer from a red fast-gates run to the open release-green issue.

**Second, and this is the part that matters for this PR: #8540 classifies exactly these numbers as something contributors should not be handling.** Verbatim from its body:

> _Ratchet drift (eslint warnings / cognitive-complexity / file-size) listed above is expected mid-cycle and is rebaselined at release — it is NOT a contributor concern and did not, on its own, open this issue._

Meanwhile `quality.yml`'s fast-gates job blocks every contributor PR on that same drift. So the repo is running two contradictory policies on one metric: release-green says mid-cycle drift is expected and gets rebaselined at release; fast-gates says it blocks you now. This PR exists only because of the gap between those two, which suggests the real fix is upstream of it.

**Third, my zero-headroom choice is probably wrong, and your own ledger says so.** I argued above that pinning to the measurement is more honest than the +62 the 07-20 note took. But `_rebaseline_2026_07_19_v3849_fix_sweep_cluster` documents the other failure mode precisely — the tip had 3 units of slack and "every new-feature PR was tripping the ratchet (this was the 4th such block of the day)". At this branch's merge rate the first merged +1 after this lands re-reds every open PR, and the next contributor repeats today's archaeology. An absolute global ceiling has no good headroom value: zero slack taxes innocents continuously, wide slack produces rare expensive digs like today.

**What I think the actual fix is: compare against the merge-base, not a frozen mid-cycle ceiling.** Fail only when `head > max(recorded_baseline, merge_base_measurement)`. That gives the property I claimed to want — the merge that adds a violation owns it — without either headroom failure mode, and it makes the fast-gates behaviour consistent with release-green's "drift is not a contributor concern". `check-complexity-ratchets.mjs` already takes a `--baseline` argument (lines 26-30), so the plumbing is half there, and the extra ESLint walk is only needed on the red path. The same trick is nearly free for `check:file-size` via `git show <base>:<file> | wc -l`.

So: **treat this PR as a stopgap to unblock the queue, not as the fix.** The measurements in it stand and are independently confirmed by #8540. If you'd rather I go straight at the delta-vs-merge-base change instead and close this, say so — and if you want the headroom back in the meantime, name a number and I'll set it rather than have this re-red the branch in a day.

MumuTW:
Closing as **obsolete / empty after rebase**.

After rebasing onto current `release/v3.8.49`, `check:complexity-ratchets` is already green at the frozen ceilings:

- complexity **2188** = baseline 2188  
- cognitiveComplexity **971** = baseline 971  

No drift to absorb and no slack to bank — the merge-train rebaseline already moved the tip to these values. The intended payload (cyclomatic/cognitive rebaseline + skill sync) is gone; only a stale changelog fragment would remain.

The underlying problem is not “measure again next cycle”, it is the **global violator-count shape** of the complexity ratchets (one integer for the whole repo). That is tracked as follow-up **#8743** (named follow-up to #8522). Prefer fixing the shape over reopening another #8580-class treadmill PR.

Issue #3501, opened 2026-06-09
[refactor] Decompose providers/[id]/page.tsx god-component (12,882 LOC → ≤800, phased/strangler)


diegosouzapw:
## Decompose the 12,882-LOC provider-detail god-component

`src/app/(dashboard)/dashboard/providers/[id]/page.tsx` is **12,882 LOC** (~195 `useState`, 21 `useEffect`, 87 inline `fetch()`, 33 `useCallback`, ~34 inline types). It's the single largest file in the repo and the top rot vector — no agent can hold it in context, so edits risk clobbering unrelated state. The new `check-file-size` gate freezes it (shrink-only); this issue drives it under the 800-LOC cap **incrementally** (strangler-fig, each phase shippable, no big-bang rewrite).

Good news from the analysis: it's **not** one tangled function. It's one ~4,510-LOC component (`ProviderDetailPage`) + ~25 already-separated helpers/sub-components in the same file with **clean, props-only boundaries** — most work is *moving* code, not untangling it. The three `Import*AuthModal`s alone are **3,534 LOC (27%)** with a clean `{ onClose, onSuccess }` interface → trivially extractable first.

### Follow the repo's existing big-page convention (don't invent a new style)
- **Thin `page.tsx` → `*PageClient.tsx`** (like `api-manager/page.tsx`→`ApiManagerPageClient.tsx`, `endpoint/page.tsx`→`EndpointPageClient.tsx`). The target currently violates this.
- **Colocated `components/` + `hooks/`** — the `providers/` route already does this (`providers/hooks/useProviderModels.ts` is the exact hook shape to follow; `providers/services/` shows components+hooks+tabs done right).
- **Pure helpers → `*Utils.ts`** (`providerPageUtils.ts` already exists).
- **State stays `useState`/`useEffect`/`fetch` in custom hooks** — the dashboard uses NO React Query/SWR/zustand for page-local state; match that (extract hooks like `useProviderModels`, not a new data lib). Use `useReducer` only for the wizard-ish auth-import state.
- **Ratchet idiom** = clone `check-t11-any-budget.mjs` (already done as `check-file-size`).

### Phased plan (each phase = 1 PR, shippable, ratchet drops each time)
- **Phase 0** — introduce `ProviderDetailPageClient.tsx` + 10-LOC `page.tsx` wrapper; add a smoke render test (no test exists today). *behavior unchanged.*
[...]

Issue #2562, opened 2026-05-22
feat: Provider-level Wildcard Support in Combos (Provider-based Combos)


Rahulsharma0810:
### Feature Request: Support Provider-Level Wildcards / Model Selection in Combos

Currently, OmniRoute's combos are model-centric, requiring each step in the combo to target a specific model ID (e.g., `openai/gpt-4o`, `anthropic/claude-3-5-sonnet`).

We would like to introduce the ability to define combos at the **provider level**, with optional wildcard or specific model selection. This allows users to easily group entire providers, fallback dynamically, or balance loads across different models of specific providers.

#### Proposed Notation / Use Cases:
1. **Full Provider Wildcards:**
   * `Provider1(*)` + `Provider2(*)`
   * Dynamically resolves/falls back across all available/configured models for `Provider1`, then falls back to `Provider2`.
2. **Hybrid Explicit & Wildcard Combo:**
   * `Provider1(openrouter/free)` + `Provider2(*)`
   * Target a specific model/pattern on the first provider, then fall back to all available models on the second provider.
3. **Specific Model Fallbacks across Providers:**
   * `Provider1(*)` + `Provider2(some/free)`

#### Expected Benefits:
* **Easier configuration:** Users don't have to manually list every model for a provider when creating a backup/combo pipeline.
* **Resilience & Future-proofing:** Automatic adjustment when new models are added to a provider's catalog without needing to update existing combos manually.

Rahulsharma0810:
Thanks for the auto-triage suggestion! However, this feature request is distinct from #563:

* **#563 (Per-model Combo Support)** maps incoming client request models (e.g., `claude-sonnet*`) to a specific **Combo**.
* **This feature request (Provider-based Combos)** is about the **internal steps/targets definition** of a Combo itself. Currently, a Combo definition must list specific, concrete model IDs (e.g., `provider/free-fast`, `provider2/free`). We are proposing that a Combo's targets can be defined using provider-level wildcards (e.g., `provider/*` or `provider2/*`), so that the Combo engine dynamically resolves to or falls back across all models supported by those providers.

This allows users to define a failover chain across entire providers without having to manually list, maintain, or update every specific model ID.

diegosouzapw:
Hi @Rahulsharma0810! Thanks for the well-articulated proposal — and for preemptively clarifying the distinction from #563! 🙏

[...]

PR #8494, opened 2026-07-24
fix(backend): fail closed when capability filters empty the combo pool

Prudhvivuda:
## Summary
- Combo compatibility filter and auto strategy tool/context pre-filters no longer re-admit the full pool when every candidate fails a hard requirement.
- Tools / vision / structured_output exhaustion now returns a 400 `capability_mismatch` naming the unmet requirement and rejected targets.
- Legacy fail-open is opt-in via combo/settings `compatFilterFailOpen` (default off).

Closes #8488

## Test plan
- [x] `node --import tsx/esm --test tests/unit/8488-capability-filter-fail-closed.test.ts`
- [x] `node --import tsx/esm --test tests/unit/combo-routing-engine.test.ts tests/unit/combo-vision-aware-routing.test.ts tests/unit/combo-context-window-filter.test.ts`
- [ ] CI green on this PR

Issue #8488, opened 2026-07-24
fix(backend): capability filters fail open — a combo with no compatible target dispatches anyway instead of erroring


diegosouzapw:
## Summary

Three capability filters are fail-open: when a filter would exclude **every** candidate, it discards its own verdict and dispatches to the full pool anyway. The request then goes to a target we already know is incompatible, and the failure surfaces to the user as a malformed or missing tool call rather than as a compatibility error.

The design intent is understandable — never hard-block a combo — but the effect is that the filter silently stops doing its job in exactly the case it exists for.

## The three sites

**1. Combo compatibility filter** — `open-sse/services/combo/comboStructure.ts`, in the block around the vision check. The comment states the behavior outright:

```
// The caller keeps all targets when none qualify, so combos with no
// confirmed-vision member still behave as before.
```

The same caller behavior applies to the `requiresTools` branch (`capabilities.supportsTools === false || !capabilities.toolCalling`).

**2 & 3. Auto strategy** — `open-sse/services/combo/resolveAutoStrategy.ts`:

```ts
// :111-116  (tool-calling)
if (filtered.length > 0) { eligibleTargets = filtered; }
else { log.warn("COMBO", "Auto strategy: all candidates filtered by tool-calling policy, falling back to full pool"); }

// :147  (context window)
`Auto strategy: all candidates filtered by context-window policy (est. ${estimatedInputTokens} tokens), falling back to full pool`
```

Both log a warning and continue with the unfiltered pool.

## Observed consequence

**#7679** is a clean instance. Every `chatgpt-web` model is honestly tagged `toolCalling: false` (correct — it is a web-cookie provider with prompt-emulated tools). A combo whose members are all chatgpt-web receives a tools-bearing request, the filter excludes all of them, the fallback re-admits all of them, and the request is dispatched to a model we already knew could not do native tool calls. The user gets prose instead of a tool call, and nothing in the response says "incompatible".

Related, same family:

- **#8437** — a provider mis-tagged `toolCalling: true` passes the filter it should fail. That is a data bug, but it lands in the same blind spot: a wrong verdict and a discarded verdict are indistinguishable downstream.
- **#8429** — the same auto pre-filter collapsing the candidate pool.

## Why this is worth fixing separately

[...]

PR #8592, opened 2026-07-25
refactor(sse): extract combo target resolution into combo/targetResolution.ts

MumuTW:
**PR 2 of ~7** in the #3501 god-file decomposition campaign, shrinking `open-sse/services/combo.ts` toward the 800-line file-size cap.

## What this is

A **pure move. No behaviour change.** The target-resolution stage of `handleComboChat` — everything between the dispatch prelude and the attempt loop — moves verbatim into a new leaf, `open-sse/services/combo/targetResolution.ts`, entry point `resolveComboTargetPipeline()`.

Log message wording, log levels, and the order of side effects and awaits are preserved exactly, quirks included. Nothing was "improved" on the way out.

## Line counts

| File | Before | After |
| --- | ---: | ---: |
| `open-sse/services/combo.ts` | 3642 | **3322** (−320, gate units) |
| `open-sse/services/combo/targetResolution.ts` | — | **669** (new, under the 800 cap) |
| `tests/unit/combo-target-resolution-split.test.ts` | — | 148 (new) |

## Complexity — the leaf contributes ZERO violations

A pure move must be complexity-neutral. The first pass was not: it landed the region as one 311-line function, which turned inline code inside the (already-violating) `handleComboChat` into a **new separately-counted** violating function.

| Gate | base `4053e2314` | first pass | **this PR** |
| --- | ---: | ---: | ---: |
| `check:complexity` | 2169 | 2171 (+2) | **2169** |
| `check:cognitive-complexity` | 956 | 957 (+1) | **956** |

Fixed by splitting `resolveComboTargetPipeline` along its natural stage boundaries into 14 helpers — wildcard expansion; weighted sticky eviction / eligibility / sticky-key / selection; step-key mapper; context-overflow response; pool-size log; smart-pipeline dispatch and its fall-through logger; strategy ordering; continuity filters; task-aware ordering; prompt-cache enablement, first-target protection and affinity stage. Each stage takes the previous stage's output and returns the next.

Per-file check on the new leaf reports **0 violations** across all three rules:

```
npx eslint --no-config-lookup --config eslint.complexity-ratchets.config.mjs \
  --format json open-sse/services/combo/targetResolution.ts
→ 0 violations   (complexity, max-lines-per-function, sonarjs/cognitive-complexity)
```

> Both gates still print RED against their frozen ceilings (2130 / 951). That is **pre-existing base-red owned by #8580**, not introduced here — the numbers above are identical to base `4053e2314`. Neither complexity baseline was rebaselined.

## Sub-blocks extracted

In original order, all verbatim:

1. Provider-wildcard expansion of the combo and the combos collection (#2562)
2. Sticky-weighted LRU eviction, weighted step-group resolution and eligibility, sticky-pin validation, `resolveWeightedTargets` — including the `isTargetSelectableForWeighted` predicate
3. `applyRequestTagRouting`
4. The `getKnownContextOverflow` 400 early return
5. The weighted / nested-resolution `log.info` lines
6. The smart/pipeline-enabled dispatch through `handlePipelineCombo` / `buildPipelineResponse`, with all three fall-through branches
7. Auto-strategy ordering (`resolveAutoStrategyOrder`) and, for every other strategy, `applyStrategyOrdering`
8. Cache-optimized strategy affinity, session stickiness, eval-score ordering, request-compatibility and context-requirement filters
9. Task-aware reordering with the #4945 explicit-router pin guard
10. Prompt-cache affinity with the `shouldProtectOriginalFirst` protection
11. The parallel pre-screen (`preScreenTargets`, priority strategy only)

## Sub-blocks deliberately SKIPPED

- **Attempt-loop config** — `maxRetries`, `retryDelayMs`, `fallbackDelayMs`, `maxSetRetries`, `setRetryDelayMs` (base lines 1096–1100). These are loop configuration, not target resolution, and every one is read downstream by the attempt loop. Left in `combo.ts`.
- **`quotaCutoffResetWindowConfig`** and `comboAttemptOrder`, immediately past the region ceiling — same reason: attempt-loop state.

Nothing else in the region was skipped; no sub-block turned out to be un-extractable.

## Values that escape the region

Every identifier the region declares was grepped forward from the attempt loop. Five are still read downstream, so they are **returned** rather than recomputed:

- `orderedTargets` — the stage's net product
- `stickyWeightedLimit` and `getWeightedStepKeyForTarget` — the sticky-weighted success write-back
- the session-stickiness result — `.messageHash` is read on both the success and failure paths
- `preScreenMap` — read per attempt

The three early exits (known-context overflow, pipeline dispatch, auto-strategy `unavailableResponse`) become an `{ earlyResponse }` result the host returns, following the existing `resolveAutoStrategyOrder` precedent in the same directory. `buildAutoCandidates` lives in `combo.ts`, so it is **dependency-injected** — the `combo/` directory keeps zero back-edges and stays acyclic.

Also removed from `combo.ts`: 16 imports that went dead with the move (verified per-symbol).

## Verification

| Command | Result |
| --- | --- |
| `npm run lint` | clean |
| `npm run check:cycles` | `OK` — but its default roots do NOT include `open-sse/services`, so this proves nothing here; see the cycles note below |
| `node scripts/check/check-file-size.mjs` | `OK — 183 arquivos congelados` |
| `npm run check:complexity` | **2169** (= base; RED vs frozen 2130 is pre-existing per #8580) |
| `npm run check:cognitive-complexity` | **956** (= base; RED vs frozen 951 is pre-existing per #8580) |
| `node --import tsx/esm --test tests/unit/combo-target-resolution-split.test.ts` | 5/5 pass |
| `node --import tsx/esm --test tests/unit/combo-routing-engine.test.ts` | 85/85 pass |
| `node --import tsx/esm --test tests/unit/*combo*.test.ts tests/unit/combo/*.test.ts` (157 files) | **1157/1157 pass** |
| `npm run test:vitest` | 31 files, 274/274 pass |
| `npm run test:unit` (full) | 26082 pass / 5 fail |

**The 5 full-suite failures are pre-existing on base and unrelated to this change.** Four are in `tests/unit/antigravity-oauth-postexchange-nonblocking.test.ts` (OAuth `postExchange` timeout tests) — verified by running that file on a detached checkout of base `4053e2314`, where the same four fail identically. The fifth (`db-usageanalytics-split.test.ts`) passes in isolation; parallel-run flake.

## CI gate fix (`check:known-symbols`)

First CI run failed `Fast Quality Gates`:

```
[combo] 1 estratégia(s) canônica(s) sem branch de despacho em combo.ts:
    ✗ priority
```

Real, and caused by this PR — though no dispatch was lost. The gate's combo sub-check
scans a hardcoded file list, and the `strategy === "priority"` latency pre-screen moved
into `combo/targetResolution.ts` with the rest of the region.

Fix is one entry: register `combo/targetResolution.ts` in `comboDispatchFiles`, following
the precedent already in that list (the Block J decomposition added its leaves for exactly
this reason). Gate now reports 20 canonical strategies, all 20 via despacho;
`tests/unit/check-known-symbols.test.ts` stays green at 35/35.

## Base-red: the `Unit Tests fast-path (2/4)` job

Also red, and **not** this PR. Two tests in
`tests/unit/serial/combo-quota-share-cooldown-wait-timing.test.ts` (expected 200 → got 429;
expected 429 → got 403). Reproduced byte-identical on a pristine detached worktree at the
base `4053e2314`. Pre-existing.

## Baseline changes — read this, it was not forgotten

Exactly **one** entry was ratcheted: `open-sse/services/combo.ts` in `config/quality/file-size-baseline.json`, **3642 → 3322**, with a justification note in the file's existing note style. `check-file-size.mjs --update` was deliberately **not** used — it sweeps ~74 unrelated files. The diff on that file is 2 insertions, 1 deletion.

No complexity or cognitive-complexity baseline was touched.

⚠️ #8582 ratchets the same `combo.ts` key to **3341**. That is the one and only merge conflict between the two PRs. Whoever rebases second must set it to the **combined** figure, not simply take the lower of the two. Measured on the actual merged tree: `combo.ts` merges to 3017 lines by `wc -l`, which is **3018 gate units** — `check-file-size` counts `split("\n").length`, one higher than `wc -l`. **3018** is the number that belongs in the baseline entry.

## Incidental formatting

`prettier --write` on `combo.ts` also reformatted one **pre-existing** 103-char line at base line 2927 (`computeCompatRejectedTargets(...)`, inside `handleRoundRobinCombo`) into 5 lines. It is the only non-move hunk in the diff. That line already failed `prettier --check` on base, so reverting it would leave the file rejecting the formatter — it is kept deliberately.

## Tests

Repo rule #8 — production code changed, so tests ship in the same PR. As a pure move, the existing 157-file `tests/unit/*combo*` suite is the real regression guard (all green above). `tests/unit/combo-target-resolution-split.test.ts` adds the new leaf's own direct coverage: the exported entry point, priority-path target ordering, the shape handed back to the attempt loop, the empty-pool case, and the known-context-overflow `earlyResponse` (400 + `context_length_exceeded` + zero attempts).

## Stacking

Stacked alongside **#8582 (`refactor/combo-dispatch-prelude`)**, which touches a **disjoint region** of the same file (pinned-model routing, fusion, pipeline, nested combo-ref execute mode, round-robin delegation; base lines 283–286, 603–667, 727–1094). This PR's first body hunk starts at base line 1102.

`git merge-tree --write-tree` of the two branches auto-merges `open-sse/services/combo.ts` **cleanly — including the import block**, where both PRs delete imports their own region no longer needs. The sole conflict is the `file-size-baseline.json` key noted above.

## On cycles

`check:cycles` is cited green above, but its default roots are `src/shared/components`, `src/lib/db`, `src/lib/compliance`, `open-sse/translator`, `open-sse/mcp-server` — **`open-sse/services` is not among them**, so a green run says nothing about this directory. (Caught by an adversarial review of #8582; the same correction applies here.)

Scoped explicitly and compared against a pristine detached worktree at the base commit, `open-sse/services` carries a large pre-existing cyclic cluster that already contains `combo.ts`. This PR's leaf imports members of that cluster, so it joins it. That is **not** a new back-edge: no file under `combo/` imports `combo.ts` — the leaf takes `buildAutoCandidates` by dependency injection — and no previously-acyclic pair became cyclic. Untangling the cluster belongs with the later phases of #3501.


diegosouzapw:
Hi @MumuTW — thanks for this Block K decomposition of `combo.ts`. The shape is clean: 11 sub-blocks extracted, values that escape the region (`orderedTargets`, `preScreenMap`, `stickyWeightedLimit`, etc.) are returned rather than recomputed, and `buildAutoCandidates` is dependency-injected so `combo/` stays acyclic against `combo.ts`. Verified body: 1157/1157 combo tests + 274/274 vitest, the 5 full-suite failures reproduce byte-identical on base `4053e2314`.

**One mandatory pre-merge item:** the `file-size-baseline.json` key for `open-sse/services/combo.ts` is set to `3322` here, but PR #8582 sets it to `3341`. The combined merged tree measures **3018** (`wc -l` = 3017, gate units = `split("\n").length + 1`). Whoever rebases second must use **3018**, not the lower of the two. The `git merge-tree --write-tree` of the two branches is otherwise clean.

`check:cycles` default roots do not include `open-sse/services`, so the green run cited in the PR body says nothing about this directory — but the body already acknowledges that, and the new leaf takes `buildAutoCandidates` by dep injection, so no new back-edge is introduced.

Approving as `merge-ready` (★4). Will hand off to `/merge-prs` for the rebase dance.

MumuTW:
## CI fix pushed (`c127e3d03`)

Root-caused the two reds on the stacked tip:

### Unit Tests 2/4 — **real regression from the extract (not base-red)**
Stacking `targetResolution` onto the dispatchPrelude tip dropped the **#8488 / #8494** capability fail-closed wiring:

- hard filters emptied the pool → generic **404** `no_executable_targets` instead of **400** `capability_mismatch`
- `compatFilterFailOpen` was never passed into `filterTargetsByRequestCompatibility`, so fail-open never re-admitted the pool

Restored:
- main path: `applyContinuityFilters` in `combo/targetResolution.ts` (failOpen + `describeCapabilityFilterExhaustion` → `earlyResponse`)
- round-robin path: matching block in `handleRoundRobinCombo`

Local: `combo-routing-engine` **86/86** + `combo-target-resolution-split` **5/5** green (including the two #8488 cases).

### Fast Quality Gates — incomplete stack rebank
`check:file-size` was red on tip growth the prior rebank missed (plus freezing `comboStructure.ts` / `RequestTimeline.tsx`). Rebanked measured tip LOC with a justification note.

Waiting on CI for confirmation.

Issue #5312, opened 2026-06-29
[BUG] Claude adaptive-thinking corrupts tool calls & leaks </think> on Cursor/OpenAI path; Thinking Budget UI is a no-op


vitalNohj:
## Environment

| Field | Value |
|---|---|
| OmniRoute Version | `3.8.37` |
| Installation Method | npm (global) |
| Operating System | Linux |
| OS Version | Linux kernel `6.8.0-124-generic` |
| Node.js Version | `v22.22.3` |
| Provider(s) Involved | Claude / Anthropic (OAuth) |
| Model(s) Involved | `claude-opus-4-8` (also affects adaptive-thinking-only models like `claude-fable-5`) |
| Client Tool | Cursor (OmniRoute OpenAI-compatible endpoint `/v1/chat/completions`) |
| Test Impact | Needs a new integration test |

## Description

When Cursor routes Claude OAuth traffic through OmniRoute's OpenAI-compatible endpoint (`sourceFormat: "openai"`, `targetFormat: "claude"`), adaptive-thinking handling corrupts multi-turn tool-calling and can leak a raw `</think>` sentinel into visible assistant content.

Three interacting problems:

**1. Thinking is force-injected on every Claude request**, regardless of settings (`open-sse/executors/base.ts`):

```ts
} else if (tb.thinking === undefined && tb.output_config === undefined) {
  tb.thinking = { type: "adaptive" };
  tb.context_management = { edits: [{ type: "clear_thinking_20251015", keep: "all" }] };
  tb.output_config = { effort: "high" };
}
```

**2. Thinking is mapped to the non-standard `reasoning_content` field** that Cursor's OpenAI consumer doesn't round-trip, so prior thinking is dropped from history (`open-sse/translator/response/claude-to-openai.ts`):

```ts
} else if (delta?.type === "thinking_delta" && delta.thinking) {
  results.push(createChunk(state, { reasoning_content: delta.thinking }));
}
```

**3. On replay, a fake signature is fabricated** (`open-sse/translator/request/openai-to-claude.ts`):

```ts
if (msg.reasoning_content) {
  blocks.push({
    type: "thinking",
    thinking: msg.reasoning_content,
    signature: DEFAULT_THINKING_CLAUDE_SIGNATURE,
  });
}
```

The net effect on adaptive-thinking models is a corrupted thinking/answer boundary: malformed tool-call turns and a stray `</think>` in visible `content`. Requests still return HTTP 200, so this fails silently.

[...]

Issue #2470, opened 2026-05-21
[BUG] Global System Prompt export/import loses enabled state and prompt restore


disonjer:
### OmniRoute Version

3.8.1

### Area

AI Settings / Global System Prompt / persistence across restarts and import

### Status

**All fixes verified and working.**
- Restart: `[STARTUP] Global System Prompt restored from settings` in logs, toggle and prompt preserved.
- JSON import: systemPrompt hydrated immediately after import without restart.
- SQLite import: systemPrompt hydrated immediately after import without restart.

### Description

The Global System Prompt is stored in the database (`key_value` table, namespace `settings`, key `systemPrompt`, value `{"enabled":true,"prompt":"..."}`), but the in-memory config (`_config` in `open-sse/services/systemPrompt.ts`) is never hydrated from the DB on server startup or after import operations. After any restart or import, the prompt and toggle revert to defaults (`enabled: false, prompt: ""`).

### Root Cause

`setSystemPromptConfig()` is only called from the PUT endpoint (`/api/settings/system-prompt`). Three scenarios fail:

1. **Server restart**: `applyRuntimeSettings()` does not restore systemPrompt into in-memory config.
2. **JSON import** (`/api/settings/import-json`): `runJsonMigration()` writes systemPrompt to DB but doesn't update in-memory state.
3. **SQLite import** (`/api/db-backups/import`): DB is replaced entirely but in-memory state remains stale.

---

## Fix 1: Startup persistence

**File: `src/server-init.ts`**

Add import:

```diff
 import { getSettings } from "./lib/db/settings";
 import { applyRuntimeSettings } from "./lib/config/runtimeSettings";
+import { setSystemPromptConfig } from "@omniroute/open-sse/services/systemPrompt.ts";
```

Add restoration after `applyRuntimeSettings`:

```diff
     const runtimeChanges = await applyRuntimeSettings(settings, { force: true, source: "startup" });
     if (runtimeChanges.length > 0) {
       startupLog.info(
         { sections: runtimeChanges.map((entry) => entry.section) },
         "Runtime settings hydrated"
       );
     }

+    // Restore Global System Prompt from DB into in-memory config
+    if (settings.systemPrompt && typeof settings.systemPrompt === "object") {
+      setSystemPromptConfig(settings.systemPrompt as { enabled?: boolean; prompt?: string });
+      startupLog.info("Global System Prompt restored from settings");
+    }
```

---

## Fix 2: JSON import hydration

[...]

Issue #8542, opened 2026-07-25
fix(ci): base-red compounds because fast-gates fail-fast hides later gates, and PRs land while red


MumuTW:
## Context

While clearing the base-red surface on `release/v3.8.49` (#8524, #8534, #8538, #8539, #8541) I kept hitting the same shape: a gate that is red on the base tip, that nobody could have been warned about at PR time, that then makes every subsequent PR red for inherited reasons. This issue is about that mechanism rather than any individual gate — the five fixes are already open.

Filing it because the failure mode is self-reinforcing: the more gates are red on the base, the less signal any single PR's CI carries, and the more normal it becomes to merge anyway.

## Evidence 1 — `Fast Quality Gates` fails fast, so later gates are never evaluated

The job runs ~20 `check:*` steps in sequence. A failure stops the job, and every later step reports `skipped`. Observed on two PRs whose entire purpose is to fix one of those later steps:

```
#8534                                #8538
failure  check:file-size             failure  check:file-size
skipped  check:db-rules   ← its fix  skipped  check:db-rules
skipped  check:mutation-test-coverage        skipped  check:mutation-test-coverage  ← its fix
skipped  check:complexity-ratchets           skipped  check:complexity-ratchets
```

Two consequences:

1. **A PR that introduces a violation in a later gate gets no warning**, as long as any earlier gate is already red on the base. The gate is not lenient — it simply never runs.
2. **A PR that fixes a later gate cannot demonstrate it.** #8534 and #8538 have only local evidence; CI has never executed the step either one repairs.

This is also why the base-red set was discovered one layer at a time rather than all at once: `check:file-size` masked `check:db-rules`, which masked `check:mutation-test-coverage`, which masks `check:complexity-ratchets` (still unowned). Each fix reveals the next.

## Evidence 2 — PRs are landing with the gates red

Checked the rollup state at merge time for the PRs that introduced the current base-red items:

| PR | Merged | Failing checks at merge |
| --- | --- | --- |
| #8446 | 2026-07-24 23:36 | `Fast Quality Gates`, `Unit Tests fast-path` 1/4, 2/4, 4/4, `No new ESLint warnings` |
| #8403 | 2026-07-24 23:37 | `Fast Quality Gates`, `Vitest (fast-path)`, all four `Unit Tests` shards |
| #8404 | 2026-07-24 23:37 | same as above |
[...]

Issue #8541, opened 2026-07-25
fix(resilience): combo cooldown-wait gate still quota-share/auto only, contradicting its own "ALL strategies" contract (#7301)


MumuTW:
## Summary

`open-sse/services/combo.ts` documents the cooldown-aware 429 retry as available to every combo strategy, but the gate that reaches it is still restricted to `quota-share` and `auto`. The two statements sit ~1200 lines apart in the same file and contradict each other. A test that #7301 shipped for the generalized behavior fails on `release/v3.8.49` at its own HEAD as a result.

## The contradiction

**`open-sse/services/combo.ts:2712-2730`** — the decision path:

```
// Cooldown-aware retry: instead of crystallizing the 429/503, wait out
// a SHORT transient cooldown and re-run the whole set loop. [...]
// Available to ALL combo strategies (not just quota-share).
if (comboCooldownWaitEnabled && status === 429) {
  // ONE decision path for EVERY strategy. The reason that drives the
  // wait is always the target's REAL model-lockout reason [...]
```

**`open-sse/services/combo.ts:1494`** — the gate feeding it:

```ts
const comboCooldownWaitEnabled = isComboCooldownWaitEligible(
  strategy,
  resilienceSettings.comboCooldownWait
);
```

**`open-sse/services/comboConfig.ts:48-53`** — the gate itself:

```ts
export function isComboCooldownWaitEligible(
  strategy: string,
  comboCooldownWait: Pick<ComboCooldownWaitSettings, "enabled">
): boolean {
  return (strategy === "quota-share" || strategy === "auto") && comboCooldownWait.enabled;
}
```

So for `priority` / `weighted` / `round-robin` / etc., `comboCooldownWaitEnabled` is always `false` and the "ONE decision path for EVERY strategy" is unreachable. A short transient 429 crystallizes instead of being waited out — the behavior #7301 set out to remove.

## Failing test

`tests/unit/serial/combo-quota-share-cooldown-wait-timing.test.ts` → `non quota-share (priority): short 429 cooldown → waits and re-dispatches (2nd pass 200)`

```
expected the retried dispatch to succeed with 200
429 !== 200
```

The case dispatches a `priority` combo whose first upstream call returns 429 with a 250 ms `retryAfter` and whose second returns 200. It expects exactly 2 upstream calls with a wait between them; the gate closes, so there is 1 call and the original 429 is returned.

[...]

PR #8604, opened 2026-07-25
fix(sse): restore task-aware routing config on restart (#8601)

### LINKED ISSUE #8601: fix(backend): task-aware routing config silently reverts to defaults on every restart — settings.taskRouting never rehydrated
## Summary

Task-Aware Smart Routing (T05) config is written to the DB but **never read back**. After any restart the feature silently reverts to `enabled: false` and the hardcoded `DEFAULT_TASK_MODEL_MAP`, discarding whatever the operator configured in the dashboard.

There are two independent root causes; fixing only the first one will not work.

## Root cause A — no boot hydration

`PUT /api/settings/task-routing` persists the config:

```ts
// src/app/api/settings/task-routing/route.ts:65-66
const { stats, ...persistable } = getTaskRoutingConfig();
await updateSettings({ taskRouting: JSON.stringify(persistable) });
```

But nothing ever reads `settings.taskRouting` back. `rg taskRouting -t ts src open-sse` returns only the route itself and the Zod schema — there is no hydration call in `src/instrumentation-node.ts`.

This is the same class as #2470 (Global System Prompt) and #5312 (Thinking-Budget), both of which needed an explicit hydration step because `applyRuntimeSettings` does not cover these keys.

## Root cause B — module-level `let` is duplicated per module graph

```ts
// open-sse/services/taskAwareRouter.ts:171
let _config: TaskRoutingConfig = { enabled: false, ... };
```

`thinkingBudget.ts:63-68` documents exactly why this shape breaks:

> A plain module-level `let` is DUPLICATED per graph, so the boot hydration would land on the instrumentation graph's copy and never reach `base.ts` — exactly the #5312 fix-A break proven on the VPS.

So adding a hydration call alone would write into the instrumentation graph's copy while `src/sse/handlers/chat.ts:573` keeps reading its own. The config store must move to the `globalThis` pattern already used by `systemPrompt.ts` (#2470) and `thinkingBudget.ts` (#5312).

Note also that `src/server-init.ts` is **not** the live boot path (`instrumentation-node.ts:366` and `:407` both call it out as unused) — the hydration must be wired into `src/instrumentation-node.ts`.

## Reproduction

1. `PUT /api/settings/task-routing` with `{"enabled": true, "taskModelMap": {"coding": "<your model>"}}`
2. `GET /api/settings/task-routing` → shows the new config
3. Restart the server
4. `GET /api/settings/task-routing` → back to `enabled: false` and the default map

## Expected

The persisted config survives a restart, and the restored value is visible to the chat handler that actually consumes it.

## Affected files

- `open-sse/services/taskAwareRouter.ts:171` (module-level `_config`)
- `src/instrumentation-node.ts` (missing hydration call)
- `src/app/api/settings/task-routing/route.ts:66` (write path — correct, no change needed)

MumuTW:
Fixes #8601.

## Problem

`PUT /api/settings/task-routing` persists the T05 Task-Aware Smart Routing config to `settings.taskRouting`, but nothing ever reads it back. After any restart the feature reverts to `enabled: false` and the hardcoded `DEFAULT_TASK_MODEL_MAP`, discarding the operator's configuration.

Two independent root causes — fixing only the first would not have worked.

### A. No boot hydration

`rg taskRouting -t ts src open-sse` returned only the write path and the Zod schema. `applyRuntimeSettings` does not cover this key, so it needs an explicit hydration step — same as the Global System Prompt (#2470) and the Thinking-Budget config (#5312).

### B. Module-level `let` is duplicated per module graph

```ts
let _config: TaskRoutingConfig = { enabled: false, ... };   // taskAwareRouter.ts:171
```

`thinkingBudget.ts:63-68` documents why this breaks: a module-level binding is duplicated per graph, so boot hydration on the instrumentation graph never reaches the copy `src/sse/handlers/chat.ts:573` reads. This is the exact failure #5312 fix-A hit on the VPS.

## Fix

- `taskAwareRouter.ts` — move the config store to the `globalThis` pattern used by `thinkingBudget.ts` and `systemPrompt.ts`; add `hydrateTaskRoutingConfig(settings)`, accepting either the JSON string the route writes or an already-parsed object, failing open on malformed input. Runtime `stats` are never restored from the persisted blob.
- `src/instrumentation-node.ts` — call it next to the Thinking-Budget restore. Wired into the real boot path, **not** the unused `src/server-init.ts` (called out as unused at `instrumentation-node.ts:366` / `:407`).

No behavior change when nothing is persisted: the in-memory defaults stay in place, and the feature remains off by default.

`tests/unit/task-routing-config-restore-8601.test.ts`, 5 cases. Written first, confirmed red:

```
SyntaxError: The requested module '.../taskAwareRouter.ts'
does not provide an export named 'hydrateTaskRoutingConfig'
```

Green after the fix — 20/20 including the pre-existing `chat-route-coverage` suite:

```
node --import tsx/esm --test tests/unit/task-routing-config-restore-8601.test.ts \
                              tests/unit/chat-route-coverage.test.ts
# tests 20 / # pass 20 / # fail 0
```

The cross-module-graph case (root cause B) was separately proven to have teeth — importing a module under two distinct specifiers does yield two real instances, so the test would fail on the old `let`-based store:

```
module-let  -> A: true   B: false     <- not shared (the bug)
globalstore -> A: true   B: true      <- shared (the fix)
```

## Gates

- `npx eslint` on changed files — 4 `no-explicit-any`, all pre-existing (4 on the base ref, 4 after; `eslint-suppressions.json` already records `count: 4` for this file). No new violations, no baseline update needed.

### DIFF
--- a/changelog.d/fixes/8604-task-routing-config-restore.md
+++ b/changelog.d/fixes/8604-task-routing-config-restore.md
@@ -1,0 +1,1 @@
+- **fix(sse):** restore Task-Aware Routing config from `settings.taskRouting` at boot and keep it on `globalThis` so multi-graph imports share one store — stops the feature silently reverting to disabled on every restart (#8601) ([#8604](https://github.com/diegosouzapw/OmniRoute/pull/8604)) — thanks @MumuTW
--- a/open-sse/services/taskAwareRouter.ts
+++ b/open-sse/services/taskAwareRouter.ts
@@ -17,13 +17,7 @@
 // ── Types ───────────────────────────────────────────────────────────────────
 
 export type TaskType =
-  | "coding"
-  | "creative"
-  | "analysis"
-  | "vision"
-  | "summarization"
-  | "background"
-  | "chat";
+  "coding" | "creative" | "analysis" | "vision" | "summarization" | "background" | "chat";
 
 interface TaskPattern {
   patterns: string[];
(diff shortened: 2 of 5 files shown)


MumuTW:
## The two red checks are pre-existing base breakage, not this PR

`Fast Quality Gates` and `Unit Tests fast-path (2/4)` fail identically on the base commit itself. Verified by running both on a pristine worktree at `4053e2314` with no changes from this PR:

| Check | Pristine base `4053e2314` | This PR's CI |
| --- | --- | --- |
| `check:complexity-ratchets` | `complexity=2169` / `cognitiveComplexity=956` | `complexity=2169` / `cognitiveComplexity=956` |
| `tests/unit/serial/combo-quota-share-cooldown-wait-timing.test.ts` | 2 pass / 2 fail — `429 !== 200`, `403 !== 429` | same two assertions |

Byte-identical. Unrelated PR #8599 fails on the same two jobs in the same CI window.

Already tracked, no new issue filed:

- **#8540** — release branch not green; lists these exact numbers (`956 vs baseline 951`, `complexity=2169`) and classifies them as *"Ratchet drift (non-blocking — rebaseline at release) … NOT a contributor concern"*
- **#8541** — the failing `combo-quota-share-cooldown-wait-timing` cases
- **#8522** — absolute ratchet baselines make inherited drift red on every PR
- **#8542** — fast-gates fail-fast hides later gates

Per #8542, a green board can't be the evidence here — the fail-fast masks downstream gates. Local validation for this PR:

```
node --import tsx/esm --test tests/unit/task-routing-config-restore-8601.test.ts \
                              tests/unit/chat-route-coverage.test.ts
# tests 20 / # pass 20 / # fail 0

npm run typecheck:core     # clean
npx eslint <changed files>  # 4 no-explicit-any, all pre-existing (suppressions record count: 4)
```

Everything else on this PR is green: Vitest, Unit 1/3/4, ESLint, Docs Gates, Merge integrity, semgrep, dast-smoke.

diegosouzapw:
## Closing — this change shipped inside #8605

Thanks for this — it was a real fix, and it did land. Verified against `origin/release/v3.8.49`:

```
open-sse/services/taskAwareRouter.ts:187   // #8601: the config MUST live on globalThis, NOT a module-level `let`
open-sse/services/taskAwareRouter.ts:248   export function hydrateTaskRoutingConfig(settings: unknown): boolean
src/instrumentation-node.ts:410-412        const { hydrateTaskRoutingConfig } = ... if (hydrateTaskRoutingConfig(settings))
```

The reason this PR stayed open is worth recording: the squash commit of **#8605** (merged today) carries as its *first* sub-commit exactly `fix(sse): restore task-aware routing config on restart (#8601)` — the content of this PR. #8605 incorporated this branch before submitting, so the work merged under that PR's number rather than this one.

Closing as merged-via-#8605. The boot-time hydration is live on the release branch.

Issue #4165, opened 2026-06-18
[BUG] RATE_LIMIT_MAX_WAIT_MS ignored: requestQueue.maxWaitMs stays 5000 → mass 502 'job timed out after 5000 ms' (TI:0|TO:0)


KooshaPari:
## Summary

Mass `502` errors across the `Main` combo with **`TI: 0 | TO: 0`** and error text **`[502]: This job timed out after 5000 ms.`** — even though `.env` sets `RATE_LIMIT_MAX_WAIT_MS=480000` and `REQUEST_TIMEOUT_MS=600000`.

## Environment

- **OmniRoute:** 3.8.28 (Windows 11)
- **Data dir:** `%APPDATA%\omniroute`
- **Combo:** `Main` → `Minimax` (minimax direct + TokenRouter `openai-compatible-chat` + nvidia NIM)
- **Client:** multi-lane forge fan-out (~6 concurrent streams)

## Observed behavior (last ~3 hours, `call_logs`)

| Status | Count (Main combo) | Notes |
|--------|-------------------|-------|
| 502 | 421 | avg duration ~100s wall clock |
| 200 | 220 | successes when queue not saturated |
| 504 | 140 | opencode-zen `big-pickle` ~600s upstream |
| 400 | 109 | kilo-gateway / kiro / mistral auth-model errors |
| 429 | 42 | semaphore saturation (secondary) |

**Top 502 error string (all providers):**
```
[502]: This job timed out after 5000 ms.
```
(307 occurrences in 3h window — minimax, TokenRouter/tart, nvidia, codex, glm, opencode-go, cursor)

Dashboard call-log pattern matches: long Duration (30–70s) but zero tokens transferred.

## Root cause (local triage)

`omniroute resilience status` and `key_value.resilienceSettings` both show:

```json
"requestQueue": { "maxWaitMs": 5000 }
```

Meanwhile `.env` has:
```
RATE_LIMIT_MAX_WAIT_MS=480000
REQUEST_TIMEOUT_MS=600000
```

**`RATE_LIMIT_MAX_WAIT_MS` does not appear to override `resilienceSettings.requestQueue.maxWaitMs`.** Under concurrent load, requests queue behind per-connection semaphores (e.g. opencode-go `max_concurrent=1`, minimax/nvidia `max_concurrent=4`) and are **dropped at 5s** with the misleading upstream-style message, producing TI:0|TO:0 502s across unrelated providers.

[...]

PR #8613, opened 2026-07-25
fix(resilience): report local queue expiry as unavailable

excessivechaos:
## Summary

OmniRoute's local request-queue expiry carries the RATE_LIMIT_QUEUE_TIMEOUT code, but handleChatCore currently falls through to the generic upstream-failure status and returns HTTP 502.

This change is intentionally scoped to status classification only:

- Return HTTP 503 Service Unavailable for RATE_LIMIT_QUEUE_TIMEOUT.
- Preserve the existing retry, lockout, cooldown, and combo behavior for separate follow-up work.

A local queue-expiry response should be distinguishable from an upstream gateway failure while retaining the existing resilience behavior.

## Validation

- 	ests/unit/rate-limit-queue-timeout-message-4165.test.ts
- 	ests/unit/rate-limit-queue-timeout-lockout.test.ts
- 
pm run typecheck:core
- Focused ESLint on the changed handler

Follow-up to the queue behavior documented in #4165.

### DIFF
--- a/changelog.d/fixes/8613-local-queue-timeout.md
+++ b/changelog.d/fixes/8613-local-queue-timeout.md
@@ -1,0 +1,1 @@
+- **fix(resilience):** report local request-queue expirations as service-unavailable capacity failures instead of upstream 502 errors, preventing incorrect provider cooldowns and same-target retries ([#8613](https://github.com/diegosouzapw/OmniRoute/pull/8613)).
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Hi @excessivechaos — correct classification of `RATE_LIMIT_QUEUE_TIMEOUT` as `503` instead of `502`. Local queue expiry is a capacity failure, not an upstream provider failure — so the cooldown layer should not penalize the provider for it.

Approved as `merge-ready` (★4). The changelog entry is already in place.

PR #8616, opened 2026-07-25
fix(resilience): recover idle wedged limiters

excessivechaos:
## Summary

Bottleneck can leave a request queued with no running or executing jobs even though reservoir capacity is available. The request then waits until the local queue budget expires.

This change:

- Detects that idle-capacity wedge at queue expiry.
- Evicts the stranded limiter and drops its orphaned queue.
- Retries the never-dispatched request once on a fresh limiter.
- Retains the watchdog reset for wedges that persist beyond the queue budget.

## Validation

- 10 rate-limit-manager tests pass
- `npm run typecheck:core`
- Focused ESLint on changed files
- Regression test covers idle-capacity expiry and fresh-limiter retry

Follow-up to the recurring queue wedge behavior documented in #4165.


diegosouzapw:
Hi @excessivechaos — wedged limiter recovery is the right escape valve. Two things to confirm in the merge:

1. The retry only fires when the limiter itself is the failure (not masking upstream 429s etc.).
2. Bounded retries (no infinite loop) and a debug log so we can see when it triggers.

Approved as `merge-ready` (★3). `/merge-prs` will run the limiter/rate-limit focused tests.

PR #8624, opened 2026-07-26
Providers Page: update URL as search filter gets updated

swingtempo:
# Preserve provider search filter in URL across navigation

## Problem

On the Providers dashboard page, the search filter (`searchQuery`) was only held in React component state. Navigating to a provider detail page and pressing **Back** lost the filter, forcing the user to re-type their search.

## Solution

Sync the search filter to the browser URL as a `?search=` query parameter using `window.history.replaceState`. On mount, the existing `useSearchParams` effect already reads `?search=` from the URL — so the filter is now restored on back/forward navigation.

Using `replaceState` instead of `router.replace` avoids triggering Next.js route re-renders, prevents history-entry spam on every keystroke, and keeps the provider-highlight history state intact.

## Changes

| File | What changed |
|---|---|
| `providerPageHighlightUtils.ts` → `providerPageUrlState.ts` | Renamed to reflect broader URL-state scope; added `syncSearchToUrl()` |
| `page.tsx` | Import updated; new `useEffect` calls `syncSearchToUrl(searchQuery)` on change |
| `providerPageHighlightLogic.test.tsx` | Import paths updated; 4 new tests for `syncSearchToUrl` (set, remove, preserve params, preserve state) |

## Test plan

- `npx vitest run tests/unit/ui/providerPageHighlightLogic.test.tsx` — all 11 tests pass
- `tsc --noEmit` — no new type errors

### DIFF
--- a/src/app/(dashboard)/dashboard/providers/page.tsx
+++ b/src/app/(dashboard)/dashboard/providers/page.tsx
@@ -34,7 +34,11 @@ import {
   loadProviderPageData,
 } from "./providerPageUtils";
 import type { ProviderEntry } from "./providerPageUtils";
-import { recordProviderNavigation, resolveHighlightedCard } from "./providerPageHighlightUtils";
+import {
+  recordProviderNavigation,
+  resolveHighlightedCard,
+  syncSearchToUrl,
+} from "./providerPageUrlState";
 import {
   readProviderDisplayModePreference,
   shouldSyncProviderDisplayMode,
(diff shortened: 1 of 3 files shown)


diegosouzapw:
Hi @swingtempo — thanks for the URL sync. The `history.replaceState` approach (no Next.js navigation) is the right choice for the filter box.

Approved as `merge-ready` (★3). One small thing to check before merge: the jsdom test re-defines `window.location` via `Object.defineProperty` — please confirm the `afterEach` restores the original so the rule doesn't leak between tests.

`/merge-prs` will run the full UI suite.

swingtempo:
Confirmed the test restores window.location

swingtempo:
@diegosouzapw I was unable to merge this. It looks like @Dingding-leo made a similar change it was merged in. So I guess I will not try to recreate/merge. 
Austin, please chime in.
Thanks!

PR #8628, opened 2026-07-26
fix(claude): classify native subscription quota 429

costaeder:
Native Claude OAuth can return `This request would exceed your account\047s rate limit` when its subscription window is exhausted. The prior generic 429 path gave it a short retry cooldown and kept cycling the same unavailable account.\n\nThis classifies that exact wording as subscription quota only for the `claude` provider, preserves generic 429 handling for other providers, and marks the corresponding chat failure as quota exhausted.\n\nTests: targeted account-fallback suite (13 passing).

### DIFF
--- a/open-sse/services/accountFallback.ts
+++ b/open-sse/services/accountFallback.ts
@@ -1578,7 +1578,8 @@ export function checkFallbackError(
       const subResult = buildSubscriptionQuotaFallback(
         errorStr,
         getUpstreamRetryHintMs,
-        parseRetryFromErrorText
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Hi @costaeder — nice catch on the Claude native "exceed your account's rate limit" wording. Provider-scoped so other upstreams using the same phrase still go through the generic 429 retry path. Tests cover both branches.

Approved as `merge-ready` (★3). I retargeted the base from `main` to `release/v3.8.49` so it lands in the active cycle. `/merge-prs` will pick it up.

PR #8630, opened 2026-07-26
[v3.8.50] fix(antigravity): lock full quota per exact model

costaeder:
A full-quota Antigravity 429 currently persists `rate_limited_until` for the whole connection. That disables unrelated models on the same account.\n\nThis introduces an exact provider/account/model lock that coexists with existing quota-family locks, routes Antigravity quota paths through it, and keeps the connection eligible for sibling models. A successful request for the same model clears the exact lock through the existing success path.\n\nThe lock is intentionally in-memory; the connection-wide persisted cooldown is not reused because it would reintroduce the cross-model outage.\n\nTests: Antigravity quota-family suite (7 passing).


diegosouzapw:

PR #8637, opened 2026-07-26
fix(contextManager): pass raw message array to estimateTokens to support image token estimation (fixes #8594)

### LINKED ISSUE #8594: fix(api): estimateTokens callers JSON.stringify before calling — bypasses #8368 image-token fix in compressContext and purifyHistory
## Summary

The `#8368` fix (landed in `1f58a29`) updated `estimateTokens()` to detect inline base64 image blocks and substitute a bounded 1 200-token estimate instead of measuring the raw base64 payload as text. **But six production call sites still JSON-stringify the object before passing it to `estimateTokens`**, silently forcing the string-based path and undoing the fix for those paths.

## Affected call sites

`open-sse/services/contextManager.ts`:
- Line 297 — `compressContext` initial estimate
- Line 307 — after Layer 1 (trim tool messages)
- Line 320 — after Layer 2 (compress thinking)
- Line 333 — after Layer 3 (purify history)
- Line 419 — `purifyHistory` binary-search candidate check

`open-sse/services/combo.ts`:
- Line 1821 — fallback compression threshold check

## Impact

For any **Chat Completions** request that carries inline base64 images (`{ type: 'image_url', image_url: { url: 'data:image/...' } }`) and triggers compression:

1. `compressContext` over-estimates the token count (500 KB image → ≈ 125 000 tokens via base64 text, instead of the correct 1 200-token estimate) → thinks the context is way over the limit and enters aggressive compression.
2. `purifyHistory` binary-search prunes messages unnecessarily, keeping fewer turns than needed and causing silent context loss.
3. `combo.ts` fallback compression fires for requests that fit comfortably in the context window.

## Root cause

```ts
// contextManager.ts:297 — and five more identical sites
let currentTokens = estimateTokens(JSON.stringify(messages)); // ← stringify forces the string path
```

`estimateTokens(string)` takes the fast path: `Math.ceil(str.length / CHARS_PER_TOKEN)`. The image-detection walk (`extractImageTokens`) only runs for the object overload introduced by #8368. JSON-stringifying before the call strips the structural type information the fix depends on.

## Fix

Replace every `estimateTokens(JSON.stringify(x))` with `estimateTokens(x)` at the six sites above. The function already handles both strings and objects; passing the object directly is correct and produces the bounded estimate.

## Acceptance criteria

- A ChatCompletions `messages` array with inline base64 images passes through `compressContext` without triggering unnecessary compression when the context fits.
- `purifyHistory` retains the correct number of messages when image-bearing messages are present.
- Unit tests asserting (a) no compression fires for a within-limit image request, (b) the binary-search correctly keeps messages.
- All existing compression-pipeline tests pass (no regression).

Refs #8368, #8560

Dingding-leo:
## Summary
Fixes #8594.

## Root Cause
In `compressContext`, `JSON.stringify(messages)` was being passed into `estimateTokens`. Since `estimateTokens` inspects payload objects to strip base64 data and calculate image tokens via `extractImageTokens`, stringifying the array into a primitive string bypassed object inspection. As a result, inline base64 image strings were estimated based on raw text length rather than image token counts.

## Changes
- Pass `messages` directly as an array/object to `estimateTokens` inside `compressContext`.
- Added unit test in `context-manager.test.ts` verifying image token estimation within `compressContext`.

### DIFF
--- a/open-sse/services/contextManager.ts
+++ b/open-sse/services/contextManager.ts
@@ -146,7 +146,10 @@ function extractImageTokens(node: unknown, seen: Set<unknown>): { node: unknown;
 
   const record = node as Record<string, unknown>;
   if (isInlineBase64ImageBlock(record)) {
-    return { node: { __image_token_estimate__: IMAGE_TOKEN_ESTIMATE }, tokens: IMAGE_TOKEN_ESTIMATE };
+    return {
+      node: { __image_token_estimate__: IMAGE_TOKEN_ESTIMATE },
+      tokens: IMAGE_TOKEN_ESTIMATE,
+    };
   }
 
   let tokens = 0;
@@ -294,7 +297,7 @@ export function compressContext(
   const targetTokens = Math.max(0, maxTokens - reserveTokens);
 
   let messages = [...body.messages];
-  let currentTokens = estimateTokens(JSON.stringify(messages));
+  let currentTokens = estimateTokens(messages);
   const stats = { original: currentTokens, layers: [] as { name: string; tokens: number }[] };
 
   // Already fits
@@ -304,7 +307,7 @@ export function compressContext(
 
   // Layer 1: Trim tool_result/tool messages
   messages = trimToolMessages(messages, 2000); // Max 2000 chars per tool result
-  currentTokens = estimateTokens(JSON.stringify(messages));
+  currentTokens = estimateTokens(messages);
   stats.layers.push({ name: "trim_tools", tokens: currentTokens });
 
   if (currentTokens <= targetTokens) {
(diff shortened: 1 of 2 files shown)


diegosouzapw:
## Closing — already fixed on the release tip, with wider coverage

Thanks for catching this — passing the structured array instead of `JSON.stringify` was the right call, and that is what the tip now does. Verified against `origin/release/v3.8.49`:

```
open-sse/services/contextManager.ts:426  // #8594: pass the structured messages array directly — estimateTokens walks it for
open-sse/services/contextManager.ts:430  let currentTokens = estimateTokens(messages);
open-sse/services/contextManager.ts:440  currentTokens = estimateTokens(messages); // #8594: object-path keeps the #8368 image estimate
open-sse/services/contextManager.ts:458  currentTokens = estimateTokens(messages);
```

This landed through #8558/#7847 (commit `1cfbfc044a`). The shipped version also carries slightly wider test coverage than this PR: `tests/unit/8594-compress-image-token-stringify.test.ts` has three dedicated cases, including a control asserting that plain text still compresses.

Closing as already-fixed. The image-token estimation issue you identified is resolved on the release branch.

PR #8557, opened 2026-07-25
refactor(sse): stop isClaudeEventPayload claiming a narrowing it never performs

backryun:
Part of #8484. **One line changed.** Clears the largest cluster in `open-sse/utils/stream.ts`.

```diff
-function isClaudeEventPayload(payload: unknown): payload is JsonRecord {
+function isClaudeEventPayload(payload: unknown): boolean {
```

## `never` did not mean unreachable — it meant the predicate was lying

Eight TS2339s read `.id` and `.choices` off `never`:

```
stream.ts(2285,38): Property 'id' does not exist on type 'never'.        x4
stream.ts(2290,53): Property 'choices' does not exist on type 'never'.   x4
```

`never` in a reachable branch is usually a signal, and it was one here:

```ts
function isClaudeEventPayload(payload: unknown): payload is JsonRecord   // L488
...
const flushedParsed = bufferedPayload as JsonRecord;                     // L2273
const isClaude = isClaudeEventPayload(flushedParsed);                    // L2277
} else if (!isClaude) {                                                  // L2283
```

The predicate answers *"does this payload carry a Claude event type?"* — a question about **contents**, not type. A Claude event and a non-Claude one are both `JsonRecord`, so `payload is JsonRecord` narrows nothing. Applied to an argument already typed `JsonRecord`, the negative branch became `JsonRecord` minus `JsonRecord` = **`never`**, and every property read in it failed.

The branch is not dead code — it is the flush-path numeric-ID normalization, and it runs.

## Why `boolean` is the right fix, not a cast

No call site depends on the narrowing:

| call site | what it does with the value |
| --- | --- |
| `stream.ts:1028` | passes it to `updateClaudeEmptyResponseLifecycle(…, payload: unknown)` |
| `stream.ts:2266` | same |
| `stream.ts:2277` | the site in question — only reads the boolean |
| `stream.ts:2188` | passed as a callback into `passthroughTailProcessor` |

And that last one settles it — `passthroughTailProcessor.ts:27` already declares the dependency as:

```ts
isClaudeEventPayload: (payload: unknown) => boolean;
```

So the codebase already consumes this as a plain boolean. The change aligns the definition with the consumer that was right all along, rather than papering over the `never` at eight call sites.

## Verification

**280 → 272, zero new errors**, on a line-number-agnostic diff of the full `tsc` error set.

- `npm run typecheck:core` — clean
- `eslint` — clean
[...]

PR #8531, opened 2026-07-25
refactor(sse): retag the designer-web result unions with string discriminants

backryun:
Part of #8484. Clears `designerWeb.ts` — all 10 of its diagnostics, one root cause.

## Same root cause as #8483 / #8499

Every one is the `strictNullChecks: false` narrowing limitation `open-sse` compiles under: a **boolean-literal discriminant narrows the positive branch but leaves the negative one as the full union**. So `if (!resolved.ok) { … resolved.status }` and everything after `if (outcome.success)` cannot see their fields.

Three unions in the file, 10 errors:

| union | errors |
| --- | --- |
| `{ ok: true; config } \| { ok: false; status; error }` | 2 (`status`, `error`) |
| `DesignerWebStepResult` (`done`/`success`) | 2 (`waitMs`) |
| the poll loop's return | 6 (`success`, `imageUrls`, `status` ×2, `error` ×2) |

## Retag, not predicates — and why

#8499 used type predicates for the same root cause. The rule recorded there is **retag when the union is module-private, use a predicate when it is exported** — a predicate exists to avoid churning a public shape and the assertions that depend on it.

Here nothing is exported: `resolveDesignerWebRequest`, `stepDesignerWebPoll`, `runDesignerWebPollLoop` and `DesignerWebStepResult` are all module-private, and no test references them. So retagging is the cheaper, clearer fix:

```
resolveDesignerWebRequest   ok: true|false   ->  state: "resolved" | "invalid"
DesignerWebStepResult       done + success   ->  state: "pending" | "ready" | "failed"
```

The step union is the interesting one: it encoded **three states across two booleans**. That is also the direct cause of the third cluster — `{ done: false; waitMs }` has no `success` property at all, so `outcome.success` was not "unnarrowed", it genuinely did not exist on every member. One discriminant for three states removes both problems.

## A too-wide signature, fixed alongside

```ts
- ): Promise<DesignerWebStepResult | { done: true; success: false; status: 504; error: string }> {
+ ): Promise<DesignerWebOutcome> {   // = DesignerWebReady | DesignerWebFailed
```

`runDesignerWebPollLoop` returns a step **only after confirming it is terminal** (`if (step.state !== "pending") return step`), and otherwise synthesizes a 504. It can never hand back a pending step. The old signature claimed it could, which is what admitted `{ waitMs }` into the union its caller destructures. The inline `{ …status: 504… }` arm also disappears — it was already exactly `DesignerWebFailed`.

## Verification

[...]

PR #8499, opened 2026-07-25
refactor(sse): narrow three result unions via type predicates

backryun:
Part of #8484. Type-only; no toolchain change.

## One root cause, eight diagnostics

The same `strictNullChecks: false` behaviour already recorded in #8483: a boolean-literal discriminant narrows the **positive** branch but not the negative one, so reading a failure-only field after `!result.ok` leaves the full union.

| file | count | field read |
| --- | ---: | --- |
| `auggie.ts` | 2 | `.error` on `AuggieModelResolution` |
| `muse-spark-web.ts` | 4 | `.error` on `GraphqlResult` |
| `notion-web.ts` | 2 | `.retryable` / `.errorResult` on the `runOnce` union |

## Why type predicates here, not a string discriminant

#8483 fixed its union by retagging the discriminant to a string. That is the better shape when the union is module-private and small — it narrows both branches with no extra machinery — and I'd have used it again if the cost were the same. It isn't:

`resolveAuggieModel` is **exported**, and its tests assert on the literal object:

```ts
assert.deepEqual(result, { ok: true, model: "haiku4.5" });
```

Retagging would churn public API and a spread of assertions to work around a checker limitation. So each union gets an explicit type predicate instead — it narrows correctly under these compiler settings while leaving the shape, every call site, and the tests untouched:

```ts
function isAuggieModelFailure(
  resolution: AuggieModelResolution
): resolution is Extract<AuggieModelResolution, { ok: false }> {
  return !resolution.ok;
}
```

`notion-web`'s union was inline in a closure's return annotation; it is named `NotionAttempt` first so there is something to `Extract` from.

## Validation

| | diagnostics |
| --- | --- |
| base | 335 |
| this PR | **327** |
| **new errors** | **0** (line-number-agnostic diff of the two error sets) |

`npm run typecheck:core` clean; the 6 existing test files importing a touched executor pass.

## Coverage

Each predicate is a one-liner whose control flow inverts on a stray `!`, so what matters is that the failure branches stay exercised. All three already have behavioral guards, which I ran:

| branch | existing guard |
| --- | --- |
| `isAuggieModelFailure` | `auggie-executor.test.ts` — 400 + `/Unknown Auggie model/` |
| `isGraphqlFailure` | `muse-spark-web-continuation.test.ts` — `"Warmup failed: …"` |
| `isFailedAttempt` | `executor-notion-web.test.ts` — nested temporarily-unavailable → retried |

[...]

Issue #8484, opened 2026-07-24
feat(dependencies): TypeScript 7 migration — tracking


backryun:
Tracking issue for the TS7 migration, per the split agreed on #7697. #7148 covered the same ground but was closed on 2026-07-14, before that discussion.

**Why this is worth doing:** TS7 is GA with 8–12× type-check speedups. TS 7.1 — which restores the Compiler API — is not released, so `typescript-eslint`, `type-coverage` and Stryker must stay on 6.x until it ships. The hybrid (TS7 native CLI + `@typescript/typescript6` under the `typescript` alias) is the documented pattern and is what #7697 implemented.

## Status — **release/v3.8.50 @ 1287b6a75debfa390d52e13eece494ce94fac660** · full open-sse/tsconfig.json reports **TS6 0 / official TS7.0.2 0**

Fresh direct compilation on the live release root passes cleanly under the repository TypeScript 6.0.3 compiler and the official TypeScript 7.0.2 compiler. Final contract PR #10339 merged on 2026-08-14, removing the last CustomModelEntry.outputTokenLimit pair. The migration remains complete after the latest root refresh; no TS7 diagnostics reappeared.

Every slice is verified the same way: a **line-number-agnostic, duplicate-preserving diff of the complete error set** before/after, re-measured against a fresh base rather than assumed. Raw counts alone mislead because line shifts and newly exposed diagnostics can otherwise look like regressions.

### Merged

| PR | slice | fixed |
| --- | --- | ---: |
| #8473 | remove deprecated `baseUrl` — unmasked all 307 | — |
| #8483 | `open-sse/utils` + `translator`, incl. the `Transformer.cancel` lib gap | 12 |
| #8485 | `gemini-business` called `combineAbortSignals()`, **defined nowhere** — a live outage, not a type nit | 2 |
| #8489 | executor `execute()` result contract | 18 |
| #8498 | executor override signatures | 4 |
| #8499 | three result unions via type predicates | 8 |
| #8520 | `mimocode` / `opencode` / `zed-hosted` shared shapes | 11 |
| #8525 | guardrail `void` return contract | 12 |
| #8528 | rerank response-adapter options | 12 |
| #8531 | designer-web unions retagged | 10 |
| #8533 | multipart / gRPC bodies as `Uint8Array<ArrayBuffer>` | 7 |
| #8557 | `isClaudeEventPayload` claimed a narrowing it never performed | 8 |
[...]

PR #8638, opened 2026-07-26
refactor(sse): retag the Adobe Firefly IMS token-check union

backryun:
Part of #8484. Type-only. **7 lines changed, 0 added.** Clears `adobeFireflyClient.ts` — all 6 of its diagnostics, one cause.

## Same root cause, fourth time

```
adobeFireflyClient.ts: TS2339: Property 'status' does not exist on type
  '{ ok: true; token: string; data: ImsTokenResponse } | { ok: false; status: number; error: string }'   x2
                        Property 'error' …                                                               x4
```

`open-sse` compiles with `strictNullChecks: false`, where a **boolean-literal discriminant narrows the positive branch but leaves the negative one as the full union**. So this:

```ts
if (authed.ok) { … } else { lastStatus = authed.status; lastError = authed.error; }
```

cannot see `status`/`error` in the `else`. Same shape as #8483, #8499 and #8531.

## Retag, per the recorded rule

`imsCheckToken` is **module-private** — no `export`, no test reference — so this retags the discriminant rather than adding a type predicate. That is the rule written down on #8499: a predicate exists to avoid churning an exported shape and the assertions that depend on it, and there is no exported shape here.

```diff
 }): Promise<
-  | { ok: true; token: string; data: ImsTokenResponse }
-  | { ok: false; status: number; error: string }
+  | { state: "ok"; token: string; data: ImsTokenResponse }
+  | { state: "failed"; status: number; error: string }
 > {
```

plus the three `return`s and the two `if`s. Seven lines, all replacements.

## Kept inline on purpose

My first pass extracted a named `ImsTokenCheck` type with a doc comment explaining the narrowing. That took the file from 2317 to **2325** and failed `check:file-size` — this file is frozen with **zero** headroom.

So the union stays inline in the return position, which is line-for-line identical to what was there. Same lesson as #8557 on `stream.ts`: on a frozen file the explanation belongs in the commit message, not in the source.

## Verification

**208 → 202, zero new errors**, on a line-number-agnostic diff of the full `tsc` error set.

- `eslint` — clean
- `check:file-size` — OK (file unchanged in length)
- **50/50** across the three Adobe Firefly suites

## No test added — the arms were already covered

Checking each arm of the retagged union is now standard for these slices, and here it comes back clean. **`cookie exchange rejects guest IMS tokens` drives both arms in a single flow:**

| request | IMS response | arm exercised |
| --- | --- | --- |
| `guest_allowed=false` | **400** `"All session cookies are empty"` | `state: "failed"` — reads `status` **and** `error` |
| `guest_allowed=true` | 200 + guest token | `state: "ok"` — reads `token`, `data` |

That test's assertion depends on the `error` text matching `/session cookies are empty/i`, which is one of the four reads that failed to typecheck — so the failed arm is not merely reachable, it is load-bearing in an existing assertion.

The same three suites pass on the parent commit, which is the evidence that the retag changes nothing observable.

## Scope note

`open-sse/services` has 27 diagnostics left after this, but they do **not** share a cause — 13 files, at most 6 apiece, and the other 6-error file (`browserBackedChat.ts`) is six unrelated problems rather than one. Those slice individually. Tracked in #8484.

### DIFF
--- a/open-sse/services/adobeFireflyClient.ts
+++ b/open-sse/services/adobeFireflyClient.ts
@@ -1569,8 +1569,8 @@ async function imsCheckToken(opts: {
   guestAllowed: boolean;
   fetchImpl: typeof fetch;
 }): Promise<
-  | { ok: true; token: string; data: ImsTokenResponse }
-  | { ok: false; status: number; error: string }
+  | { state: "ok"; token: string; data: ImsTokenResponse }
+  | { state: "failed"; status: number; error: string }
 > {
   const form = new URLSearchParams({
     client_id: opts.clientId,
@@ -1602,7 +1602,7 @@ async function imsCheckToken(opts: {
 
   if (!resp.ok) {
     return {
-      ok: false,
+      state: "failed",
       status: resp.status,
       error: sanitizeErrorMessage(
         data?.error_description || data?.error || text.slice(0, 200) || `HTTP ${resp.status}`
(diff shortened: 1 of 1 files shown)


diegosouzapw:

Issue #8577, opened 2026-07-25
fix(tests): unit suite failure set unstable across runs — parallel-load flake, macOS-broken ops script, and cancelled test


MumuTW:
## Summary

Running the full `npm run test:unit` twice on the same tree produces **different failure sets each time**, plus a steady ~35 cancelled files. Triaging every failure by re-running its file in isolation separates three genuinely different problems that all currently look like "the suite is red":

| category | count | notes |
| --- | --- | --- |
| parallel-load flake | most of them | passes in isolation, different set every run |
| a real portability bug surfaced as a test failure | 1 file | `bin/restore-policies.sh` cannot run on macOS at all |
| deterministic failures | 1 file (2 subtests) | reproduce 3/3 in isolation |
| a test that never runs | 1 file | its only test is *cancelled*, even in isolation |

Filing the triage rather than fixes, since the categories want different owners and the flake is a suite-level problem rather than a set of broken tests.

## The variance

Three full runs of the same tree:

| run | `not ok` | `fail` | `cancelled` |
| ---: | ---: | ---: | ---: |
| 1 | 6 | 3 | 34 |
| 2 | 9 | — | — |
| 3 | 12 | 4 | 35 |

The membership changes between runs, not just the count — run 3 added `model test route ignores forwarded hosts`, `resetMachineIdCache clears cached value`, `sources.ts exports the builders directly` and others that run 1 did not report. A failing set that is not stable across runs of an unchanged tree cannot be used to gate anything.

## Isolation results

Every failing file, re-run on its own:

```
antigravity-oauth-postexchange-nonblocking.test.ts    1 pass  / 0 fail   → flake
db-usageanalytics-split.test.ts                      33 pass  / 0 fail   → flake (covers 2 reported failures)
model-test-route.test.ts                              2 pass  / 0 fail   → flake
oauth-connection-test-timeout.test.ts                 0 pass  / 0 fail   → its 1 test is CANCELLED
ops-scripts.test.ts                                   5 pass  / 1 fail   → real (2/2 runs)
tests/unit/shared/machineId.test.ts                   8 pass  / 2 fail   → real (3/3 runs)
```

## 1. `ops-scripts.test.ts` — this is a real bug in the script, not the test

```
restore-policies failed: bin/restore-policies.sh: line 42: readarray: command not found
127 !== 0
```

[...]

PR #8641, opened 2026-07-26
fix(sse): split the reasoning-decision guards so the error arm actually narrows

backryun:
Part of #8484. Five of `reasoningRouting.ts`'s six diagnostics, one cause — and unlike the recent slices this one **reorders control flow**, so the behaviour claim is real rather than a formality.

## `&&` in a guard does not narrow the negation

```ts
if (decision && "error" in decision) { … return; }
if (!decision)                       { … return; }
// ↓ decision is still ReasoningRuleDecision | { error: Response }
```

A compound condition narrows only its **true** branch. Its negation cannot distinguish *"`decision` was falsy"* from *"`decision` had no `error`"*, so the `{ error: Response }` arm survived past both guards. Everything downstream saw it:

```
TS2345  applyDecision(…, decision)              — arg is the union
TS2322  reasoningDecision: decision             ×2
```

Split into two guards, falsy first:

```ts
if (!decision)          { … return; }
if ("error" in decision) { … return; }
```

**Behaviour is identical.** The two conditions are mutually exclusive — a caught resolver error produces a truthy object, so it passes `!decision` and reaches the `in` test exactly as before; a null decision has no `error` to match. Order between them cannot change which branch runs.

## The other two: an inferred return with missing arms

`applyDecision` has three returns, and two of them carry only `response`:

```ts
return { response: errorResponse(…) };   // no body, no modelStr
return { response: rejection };          // no body, no modelStr
return { body: …, modelStr: …, response: null };
```

So `applied.body` / `applied.modelStr` did not exist on every member of the inferred union.

Declared an `AppliedDecision` type rather than padding the two short returns with `body: undefined, modelStr: undefined`. The declaration has **no runtime footprint**; the filler would change those objects' own keys, which is a real (if small) behavioural difference for anything that enumerates them.

## Verification

- **8/8** across both reasoning-routing suites

## Tests — a genuine gap here

`reasoning-routing.test.ts` is thorough about the policy/rule layer (glob and tag matching, precedence, CRUD, schema validation) but **never drives `applyReasoningRouting`**, so the guarded paths had no end-to-end coverage at all.

`tests/unit/reasoning-routing-decision-guards.test.ts`, 3 tests over the no-rule pass-through — the path the reorder puts first:

- the request is forwarded **by reference**, unmodified, with `response: null` and `reasoningDecision: null`
- intent extraction still runs and is reported (`openai/gpt-4o`, `sourceEffort: "high"`) — short-circuiting early must not skip it, since downstream routing reads it even when no rule applied
- an unknown provider/model is also a clean pass-through

**They pass identically on the parent commit**, which is the evidence that the reorder changes nothing observable.

### What I did not cover, and why

The `{ error }` arm is reachable **only** when `resolveReasoningRoutingRule` throws and `resolveDecision` catches it. There is no seam for that here short of module mocking, and the real DB harness this suite uses cannot induce it cleanly. The reorder is safe for that arm by construction — argued above — but I would rather flag the gap than write a test that mocks the resolver into an unrealistic shape and call it coverage.

## Scope

Leaves the sixth diagnostic in this file: `ApiKeyInfo` vs `ApiKeyMetadata` at line 67. That is a **different, cross-file cause** — `reasoningRouting.ts` defines a local `ApiKeyInfo` that duplicates the exported `ApiKeyMetadata` in `shared/utils/apiKeyPolicy.ts`, and the same mismatch produces two more diagnostics in `src/sse/handlers/chat.ts`. Three errors across two files, one consolidation; it belongs in its own slice. Tracked in #8484.

### DIFF
--- a/src/sse/handlers/reasoningRouting.ts
+++ b/src/sse/handlers/reasoningRouting.ts
@@ -48,13 +48,25 @@ async function resolveDecision(input: Parameters<typeof resolveReasoningRoutingR
   }
 }
 
+/**
+ * Either the rewritten request, or a `response` to short-circuit with. Declared
+ * rather than inferred: two of the three returns carry only `response`, so the
+ * inferred union had arms without `body`/`modelStr` at all and the caller could
+ * not read them.
+ */
+type AppliedDecision = {
+  body?: unknown;
+  modelStr?: string;
+  response: Response | null;
+};
+
 function applyDecision(
   request: Request,
   body: any,
   policy: RoutingPolicy,
   apiKeyInfo: ApiKeyInfo | null,
   decision: ReasoningRuleDecision
-) {
+): AppliedDecision | Promise<AppliedDecision> {
   if (decision.capability === "unsupported" && !decision.targetCombo) {
     return {
       response: errorResponse(
@@ -125,16 +137,9 @@ export async function applyReasoningRouting({
     comboId: typeof requestedCombo?.id === "string" ? requestedCombo.id : null,
     requestTags: requestRoutingTags.tags,
   });
-  if (decision && "error" in decision) {
-    return {
-      body,
-      modelStr,
-      reasoningIntent: stableReasoningIntent,
-      reasoningDecision: null,
-      requestRoutingTags,
-      response: decision.error,
-    };
-  }
+  // Guards are split rather than combined: `decision && "error" in decision`
+  // only narrows its *true* branch, so the union kept the `{ error }` arm all
+  // the way down and every later use of `decision` saw it.
   if (!decision) {
     return {
       body,
@@ -143,6 +148,16 @@ export async function applyReasoningRouting({
       reasoningDecision: null,
       requestRoutingTags,
       response: null,
+    };
+  }
+  if ("error" in decision) {
+    return {
+      body,
+      modelStr,
+      reasoningIntent: stableReasoningIntent,
+      reasoningDecision: null,
+      requestRoutingTags,
+      response: decision.error,
     };
   }
 
--- a/tests/unit/reasoning-routing-decision-guards.test.ts
+++ b/tests/unit/reasoning-routing-decision-guards.test.ts
@@ -1,0 +1,86 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import fs from "node:fs";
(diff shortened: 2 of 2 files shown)


backryun:
## The red `Unit Tests fast-path (3/4)` is a network flake, unrelated to this diff

Recording the evidence so this does not read as a real failure.

**What actually failed** in [job 89840789943](https://github.com/diegosouzapw/OmniRoute/actions/runs/30219925260/job/89840789943):

```
✖ should complete within timeout   (71,542ms)
✖ testConnection method           (143,128ms)
✖ DuckDuckGoWebExecutor           (144,205ms)
```

`tests/unit/duckduckgo-web-executor.test.ts` calls `executor.testConnection({})`, which makes a **real network request to DuckDuckGo**. The test anticipates a network *error* —

```ts
} catch (error) {
  // Network error is acceptable - just verify method exists and is callable
  assert.ok(true, "testConnection is callable");
}
```

— but nothing guards against the call simply **hanging**, which is what happened here: 71s on one case, 143s on its parent. A `catch` cannot rescue a request that never settles.

**Why it cannot be this PR.** The diff is two files:

```
src/sse/handlers/reasoningRouting.ts                  | 25 +++--
tests/unit/reasoning-routing-decision-guards.test.ts  | 86 +++++++++++
```

No import path connects `reasoningRouting.ts` to the DuckDuckGo executor; the two only share a test shard. Verified locally on the rebased head (`a2afdac4b`):

- `reasoning-routing-decision-guards.test.ts` — 3/3 pass
- every other consumer of `reasoningRouting` — `reasoning-routing`, `reasoning-routing-api`, `codex-ws-policy-enforcement-6564`, `sync-bundle` — 14/14 pass
- `typecheck:core` clean

This is the class of failure tracked in #8577 (*"unit suite failure set is not stable across runs"*): a live-network assertion sitting in a blocking shard, where a slow upstream fails an unrelated PR.

**Ask:** a re-run of that job should clear it — I do not have the repository permission to trigger one (`gh run rerun` → `Must have admin rights to Repository`). Happy to open a separate PR putting a timeout/skip guard on the DuckDuckGo probe instead, which would stop it failing everyone else's PRs too.

diegosouzapw:

PR #8483, opened 2026-07-24
refactor(sse): resolve open-sse utils/translator type diagnostics for TS 7

backryun:
## Context

First slice of the split you asked for on #7697:

> I'd suggest splitting this into a smaller, fully-green PR first (baseUrl removal + TS6 diagnostic fixes only) and pursuing the TS7 native-CLI adoption as a separate, isolated follow-up once it's green end-to-end.

**No toolchain change here** — no dependency, CI, or `tsc` invocation is touched. This is only the diagnostic half, and only in the low-risk modules: 12 diagnostics across 8 files in `open-sse/utils` and `open-sse/translator`. `chatCore.ts` and `stream.ts` — the runtime-logic files you flagged — are deliberately **not** in this PR; they get their own slice with their own validation.

(`baseUrl` removal is #8473, also standalone.)

## Fixes, by root cause

| Cause | Files | Fix |
| --- | --- | --- |
| `Transformer` lib type omits `cancel` | `progressTracker`, `sseHeartbeat` (+ `stream.ts`'s existing handler) | Patch the interface in `open-sse/types.d.ts` |
| Boolean-literal discriminant under `strictNullChecks: false` | `earlyStreamKeepalive` | Retag with a string discriminant |
| Assignment resets `typeof` narrowing | `toolCallShim`, `openai-responses` | Read through a local |
| `unknown` → `string` parameter | `sanitizeToolResultId` | Coerce |
| `opts = {}` inferred as `{}` | `openaiHelper` | Add `FilterToOpenAIFormatOptions` |
| `Buffer<ArrayBuffer>` vs `Buffer<ArrayBufferLike>` (@types/node 26) | `cursorAgentProtobuf` | Bare `Buffer` on the locals |

Two of these deserve a note.

**`transformer.cancel` is load-bearing, not dead code.** The WHATWG Streams standard defines `transformer.cancel(reason)` and Node implements it — verified on v24, cancelling the readable side invokes the handler — but `lib.dom.d.ts` still omits it from `Transformer`, so every `new TransformStream({ ..., cancel() {} })` was `TS2353`. Those handlers clear the heartbeat/progress intervals when an SSE client disconnects. Deleting them to satisfy the checker would leak a timer per abandoned stream, so the type is patched instead. A test asserts the runtime contract so the premise can't rot.

[...]

PR #8645, opened 2026-07-26
refactor(sse): narrow three media-generation result unions

backryun:
Part of #8484. Seven diagnostics across five files, **one cause** — and the three unions get three different treatments, which is the interesting part.

## The cause, again

`strictNullChecks: false` means `ok: true | false` narrows the positive branch but leaves the negative one as the whole union. So `if (!result.ok)` cannot reach `status`/`error`. Fifth time this campaign (#8483, #8499, #8531, #8638).

## Three unions, three treatments

| union | visibility | treatment | fixed |
| --- | --- | --- | ---: |
| `SegmindRequestResult` | **exported** | type predicate | 4 (image + video providers) |
| `downloadGeneratedImage`'s return | module-private | retag | 2 (freepik) |
| `ValidatedMediaGenerationBody` | module-private | retag | 1 (videos route) |

Per the rule recorded on #8499 — **predicate when exported, retag when private**. `SegmindRequestResult` is the published return contract of `segmindRequest()`, so its `ok` shape stays put and gets:

```ts
export function isSegmindFailure(
  result: SegmindRequestResult
): result is Extract<SegmindRequestResult, { ok: false }> {
  return !result.ok;
}
```

The other two have no consumer outside their module, so retagging is cheaper and clearer. Both consumers of the retagged body union — the videos and music routes — were updated with it.

## What I tried, measured, and reverted

`MediaGenerationResult` in the same shared file looks like the same shape:

```ts
type MediaGenerationResult =
  | { success: true; data: unknown }
  | { success: false; error: unknown; status: number };
```

Retagging it and narrowing `failedMediaGenerationResponse`'s parameter to the failed arm fixes 2 more — but **introduces one new error**: the provider handlers that produce these results widen `success` to `boolean` rather than a literal, so the route's `result` no longer matches the narrowed parameter.

That is a different root cause — literal widening at the producers, not discriminant narrowing at the consumer — and fixing it properly means going after the producers first. Measured 10 fixed / 1 new, reverted that part, kept the invariant. Left for its own slice.

## Verification

**208 → 201, zero new errors**, on a line-number-agnostic diff of the full `tsc` error set.

- `eslint` on all seven changed files — clean
- `check:file-size` — clean
- **234/234** across the 27 affected suites

## No new tests — the failure arms are exercised

Checking the arms the change touches is standard here, and both retagged/predicated failure paths already run under test:

- `segmind-image-video-provider-6656.test.ts` drives a **500** (`text/plain` body) and a **502** through both segmind providers — the exact `ok: false` path `isSegmindFailure` now guards
- `video-generation-handler.test.ts` covers a **503** and a malformed-output **500**

The freepik download failure and the invalid-JSON body arm are narrower; they are reachable and unchanged in behaviour (retagging a discriminant has no runtime effect beyond the literal string compared), so I did not manufacture coverage for them.

### DIFF
--- a/open-sse/handlers/imageGeneration/providers/freepik.ts
+++ b/open-sse/handlers/imageGeneration/providers/freepik.ts
@@ -114,14 +114,18 @@ async function pollMysticTask(params: {
 }
 
 async function downloadGeneratedImage(imageUrl: string): Promise<
-  { ok: true; b64: string } | { ok: false; status: number; error: string }
+  { state: "ok"; b64: string } | { state: "failed"; status: number; error: string }
 > {
   const imgRes = await fetch(imageUrl);
   if (!imgRes.ok) {
-    return { ok: false, status: imgRes.status, error: `Failed to download image: ${imgRes.status}` };
+    return {
+      state: "failed",
+      status: imgRes.status,
+      error: `Failed to download image: ${imgRes.status}`,
+    };
   }
   const buf = await imgRes.arrayBuffer();
-  return { ok: true, b64: Buffer.from(buf).toString("base64") };
+  return { state: "ok", b64: Buffer.from(buf).toString("base64") };
 }
 
 async function resolveCompletedResult(params: {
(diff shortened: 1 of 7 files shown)


diegosouzapw:

PR #8650, opened 2026-07-26
fix(autoCombo): handle missing model_capabilities table in taskFitness (#8603)

### LINKED ISSUE #8603: fix(backend): static fitness table substring match returns wrong row for longer model ids (gpt-4o-mini scores as gpt-4o)
## Summary

`lookupStaticFitnessTable` matches model ids with `String.includes` over an insertion-ordered object, so a **shorter pattern listed earlier wins over the model's own, more specific row**. Cheap models get scored as their flagship namesake.

```ts
// open-sse/services/autoCombo/taskFitness.ts:295-301
function lookupStaticFitnessTable(normalizedModel, normalizedTask) {
  const table = FITNESS_TABLE[normalizedTask] || FITNESS_TABLE.default;
  for (const [pattern, score] of Object.entries(table)) {
    if (normalizedModel.includes(pattern)) return score;   // first match wins
  }
  return null;
}
```

In `FITNESS_TABLE.coding`, `"gpt-4o": 0.9` is declared before `"gpt-4o-mini": 0.8`.

## Verified behaviour

Replicating the exact lookup against the real table:

```
gpt-4o-mini        -> matched "gpt-4o"      = 0.9   (own row = 0.8)
deepseek-v3.2      -> matched "deepseek-v3" = 0.85  (own row = 0.86)
gemini-2.5-flash   -> matched "gemini-2.5-flash" = 0.82  (own row = 0.82)  OK
o4-mini            -> matched "o4-mini"     = 0.88  (own row = 0.88)       OK
gpt-4-turbo        -> matched "gpt-4-turbo" = 0.88  (own row = 0.88)       OK
```

`gpt-4o-mini` — an explicitly cheap model — is scored 0.9, the flagship's `taskFit`. The dedicated `"gpt-4o-mini"` row is unreachable.

## Blast radius

Bounded but real. `taskFit` carries weight `0.08` in `DEFAULT_WEIGHTS` (`scoring.ts:43`) and rises to `0.37` in the `quality-first` mode pack (`modePacks.ts:46`). The static table is layer 4 of the 5-layer resolution chain (`taskFitness.ts:317`), so the bug only surfaces for models that miss `user_override`, `arena_elo`, and `models_dev_tier` — i.e. exactly the long tail of the 290-provider catalog that has no leaderboard or capability coverage.

## Proposed fix

Match longest-pattern-first rather than declaration-order-first, so the most specific row always wins regardless of how the table is authored. This also makes the table safe to extend without thinking about ordering.

## Longer term

The table itself is a hand-maintained model-quality list that has rotted (`gpt-4o: 0.9` and `gpt-4-turbo: 0.88` still outrank later additions like `glm-5.1: 0.78` and `minimax-m2.5: 0.75`). Once `arena_elo` + `models_dev_tier` coverage is measured, layer 4 should be deleted rather than curated. Out of scope for the ordering fix.

Dingding-leo:
## Summary
Fixes #8603.

## Changes
- Safely check if the `model_capabilities` table exists in SQLite before attempting to query it in `loadModelCapabilities()`.
- Avoids SQL error throwing when autoCombo runs on a freshly initialized or schema-partial DB.
- Added regression unit test in `tests/unit/task-fitness-missing-table-8603.test.ts`.

### DIFF
--- a/open-sse/services/autoCombo/taskFitness.ts
+++ b/open-sse/services/autoCombo/taskFitness.ts
@@ -239,6 +239,11 @@ function loadModelCapabilities(): Record<string, ModelCapRow> | null {
 
   try {
     const db = getDbInstance();
+    const tableExists = db
+      .prepare("SELECT name FROM sqlite_master WHERE type='table' AND name='model_capabilities'")
+      .get();
+    if (!tableExists) return null;
+
     const rows = db.prepare("SELECT * FROM model_capabilities").all() as Record<string, unknown>[];
     const cache: Record<string, ModelCapRow> = {};
 
--- a/tests/unit/task-fitness-missing-table-8603.test.ts
+++ b/tests/unit/task-fitness-missing-table-8603.test.ts
@@ -1,0 +1,10 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { getModelsDevTierFitness } from "../../open-sse/services/autoCombo/taskFitness.ts";
+
+describe("taskFitness - missing model_capabilities table (#8603)", () => {
+  it("returns null safely when model_capabilities table does not exist in DB", () => {
+    const result = getModelsDevTierFitness("claude-sonnet-3-5-20241022", "coding");
+    assert.equal(result, null);
+  });
+});


diegosouzapw:
Hi @Dingding-leo — quick `model_capabilities` table-existence check before the SELECT. Fresh databases that haven't migrated yet will appreciate this.

Approved as `merge-ready` (★3). `/merge-prs` will pick it up.

PR #8113, opened 2026-07-22
fix(docs): align security docs with actual guardrail behavior (#8097)

rafaumeu:
## Summary

Closes #8097

Aligns security documentation with the actual runtime behavior of the sanitizer / prompt-injection guard.

### Changes

**docs/reference/ENVIRONMENT.md:**
- `INPUT_SANITIZER_ENABLED` default corrected from `true` → `false` (opt-in, matching code after #8093)
- `INPUT_SANITIZER_MODE` `redact` description corrected: logs + tags, does **not** strip injection text from the request body
- Source file paths updated to actual locations (`src/shared/utils/inputSanitizer.ts`, `src/lib/guardrails/promptInjection.ts`)
- Added limitations disclaimer about the heuristic nature of guardrails

**SECURITY.md:**
- Severity table corrected to match code:
  - Role Hijack: High → **Medium** (`role_hijack` = medium)
  - DAN/Jailbreak: High → **Medium** (`jailbreak_dan` = medium)
  - Delimiter Injection: Medium → **High** (`delimiter_injection` = high)
  - Instruction Leak: Medium → **High** (`instruction_leak` = high)
- Added Encoding Evasion pattern type
- Removed unconditional 'blocks prompt injection attacks' claim — replaced with 'best-effort heuristic' language
- Clarified that response-side PII sanitization is primarily **log scrubbing**, not response-body rewriting
- Noted that only High-severity detections are blocked in `block` mode

**tests/fixtures/sanitizerFixtures.ts:**
- Added evaluation fixtures covering benign (RPG persona, release notes, technical docs), injection (classic override, DAN, delimiter), and PII (email, CPF, phone) categories

**tests/unit/security-docs-truthfulness-8097.test.ts:**
- 4 tests validating docs match code: default values, severity table, redact mode claims, fixture coverage

### Verification

```
✔ SECURITY.md and ENVIRONMENT.md do not overstate sanitizer defaults
✔ SECURITY.md severity table matches code
✔ ENVIRONMENT.md does not overstate redact mode
✔ sanitizer fixtures cover benign, injection, and pii categories
ℹ pass 4  fail 0
```

PR #8715, opened 2026-07-26
fix(guardrails): ensure request PII masking respects DB feature flag overrides (fixes #8113)

Dingding-leo:
## Summary
Fixes #8113.

## Changes
- In `src/lib/guardrails/piiMasker.ts`, `isRequestPiiMaskingEnabled()` previously read directly from `process.env.PII_REDACTION_ENABLED === 'true'`. This bypassed user setting overrides stored in the SQLite database (`settings` table), causing DB feature flag toggles for `PII_REDACTION_ENABLED` to be ignored for request PII redaction.
- Updated `isRequestPiiMaskingEnabled()` to delegate to `isFeatureFlagEnabled('PII_REDACTION_ENABLED')`, which correctly checks DB setting overrides before falling back to environment variables.
- Updated unit test in `tests/unit/piiSanitizer.test.ts` to verify that `PIIMaskerGuardrail` respects DB feature flag overrides.


diegosouzapw:

PR #8719, opened 2026-07-26
[v3.8.50] fix(combo): fallback to t.provider in comboTargetLimits & add null guard in getTokenLimit (fixes #8708, fixes #8716)

### LINKED ISSUE #8708: fix(backend): PII_REDACTION_ENABLED dashboard toggle is ignored — runtime reads only process.env
### OmniRoute Version

3.8.48 (verified also present on `release/v3.8.49`, commit `36f8fd1` — and re-verified unchanged against the current branch HEAD, commit `1f04333`)

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Ubuntu on WSL2

### Node.js Version

22.22.3

### Description

`PII_REDACTION_ENABLED` is defined as a normal dashboard-editable entry in `FEATURE_FLAG_DEFINITIONS` (`src/shared/constants/featureFlagDefinitions.ts`) — writable through the generic `PUT /api/settings/feature-flags` route like every other flag, and displayed in Settings → Feature Flags with a live toggle. Nothing in the UI or the flag's `description` ("Redact personally identifiable information from requests") indicates it is any different from its neighbors.

In practice, toggling it in the dashboard has **zero effect**. The only code path that gates request-side PII redaction is `isRequestPiiMaskingEnabled()` in `src/lib/guardrails/piiMasker.ts`:

```ts
function isRequestPiiMaskingEnabled() {
  // Request PII redaction is controlled solely by PII_REDACTION_ENABLED.
  // INPUT_SANITIZER_MODE only governs prompt-injection policy (warn/block/log).
  return process.env.PII_REDACTION_ENABLED === "true";
}
```

This reads `process.env` directly and never calls `getFeatureFlagOverride()` / `isFeatureFlagEnabled()` / `resolveFeatureFlag()` — a grep of the entire file returns zero matches for any of the three, confirmed at both the current `release/v3.8.49` HEAD (`1f04333`) and the commit this report was verified against (`36f8fd1`). Nor is the flag synced from DB into `process.env` on change: `src/lib/config/hotReload.ts` and `src/lib/config/runtimeSettings.ts` (the two files that do perform DB→env hot-reload for other settings) contain no reference to `PII_REDACTION_ENABLED` either, at either commit.

This exact line was introduced by #8102 (merged, fixed #8094's `MODE=redact` coupling): the `&&` condition on `INPUT_SANITIZER_MODE` was correctly removed, but the check was left as a raw `process.env` read rather than switched to the DB-aware resolver — #8102 fixed the footgun described in #8094 without touching this layer, so it's an existing, separate gap, not a regression from that fix.

**Contrast with its sibling flag**: `PII_RESPONSE_SANITIZATION` (response-side, `src/lib/piiSanitizer.ts`) is wired correctly through the shared resolver (`src/shared/utils/featureFlags.ts`, DB > env > definition default) —

```ts
const isEnabled = () => isFeatureFlagEnabled("PII_RESPONSE_SANITIZATION");
```

— which does resolve DB overrides. So the asymmetry looks unintentional: one of the two PII flags in the same feature area respects the dashboard, the other silently doesn't.

**Related to, but distinct from, #8094 / #8102**: #8102 (merged) fixed the "config footgun" coupling redaction to `INPUT_SANITIZER_MODE=redact` and the coverage holes in `redactBody()`. This report is about the layer _underneath_ that fix — even with the footgun removed, the enable check itself never consults the DB-backed feature flag, only `process.env`.

### Steps to Reproduce

1. Fresh instance, no `PII_REDACTION_ENABLED` env var set (unset in `.env`, unset in the shell).
2. Dashboard → Settings → Feature Flags → set **PII Redaction** to enabled (or `PUT /api/settings/feature-flags` with `{"key":"PII_REDACTION_ENABLED","value":"true"}`).
3. Confirm the override is persisted: `GET /api/settings/feature-flags` shows `PII_REDACTION_ENABLED` with `source:"db"`, `effectiveValue:"true"`.
4. Send a chat completion containing PII, e.g. `POST /v1/chat/completions` with a user message containing an email address.
5. Inspect the outbound payload actually sent upstream (call log artifact / provider-side capture).

### Expected Behavior

With the dashboard flag enabled (DB override `true`), the outbound request body should have the email redacted, matching what `PII_RESPONSE_SANITIZATION` already does correctly on the response side via `isFeatureFlagEnabled()`.

### Actual Behavior

The email is forwarded to the upstream provider unmodified. `isRequestPiiMaskingEnabled()` only returns `true` when the **environment variable** `PII_REDACTION_ENABLED=true` is set at process start — the dashboard/DB override has no effect on this code path at all. An operator who enables "PII Redaction" from the dashboard, believing it protects outbound requests, gets no protection and no warning.

### Test Impact

Needs a new unit test

### Validation Plan

- Unit test on `isRequestPiiMaskingEnabled()` (or its replacement): with `process.env.PII_REDACTION_ENABLED` unset, set a DB feature-flag override for `PII_REDACTION_ENABLED=true` and assert request PII is redacted.
- Regression test: env var still wins/behaves as documented when explicitly set, so the fix does not silently invert the env-var path.
- Manual: toggle the flag from Settings → Feature Flags with no env var set, send a request containing an email, confirm the outbound body is redacted.

### Additional Context

**Suggested fix**: mirror the resolution pattern already used by `PII_RESPONSE_SANITIZATION` — replace the direct `process.env.PII_REDACTION_ENABLED === "true"` check with `isFeatureFlagEnabled("PII_REDACTION_ENABLED")` (or equivalent DB→env→default resolution), so the dashboard toggle and the env var both work as documented.

**Workaround in the meantime**: set `PII_REDACTION_ENABLED=true` in the instance's `.env` file (not via the dashboard) and restart the server.

**Scope note**: this report is about the _enable gate_ only. The coverage gaps in `redactBody()`'s content-shape handling (Responses API `input[]`, array `system`, etc.) were already reported and closed in #8094 — not re-litigated here.

### LINKED ISSUE #8716: fix(backend): combo compression-limit null.toUpperCase() when target modelStr has no provider prefix
### OmniRoute Version

Build from `release/v3.8.49` @ `4cd1bbc9f55c38f7bf9b7bfb66589520c4b18bd3` (custom, no official tag published with the fixes we needed yet).
(But I already had the problem also on the last stable versions published)

### Installation Method

Built from source (Docker)

### Operating System

Linux

### OS Version

Debian trixie (container), host: Ubuntu (Hetzner VPS, arm64)

### Node.js Version

24 (`node:24-trixie-slim` base image)

### Provider(s) Involved

Not provider-specific — reproduces for any combo target, regardless of which provider it points to.

### Model(s) Involved

Not model-specific.

### Client Tool

Any OpenAI-compatible client hitting `/v1/chat/completions` on a `combo` model.

### Description

Combo compression-limit resolution throws an uncaught (but caught by a local
try/catch) `TypeError` whenever a target reaching `chatCore.ts`'s
compression-limit block has a `modelStr` that `parseModel()` can't split into
`provider/model` — logging a warning on every affected request instead of
resolving the combo-aware context limit.

**I could not pin down the exact request/combo that first triggered this
live** (the container that produced it has since been recreated with the fix
applied, rotating its logs) — every combo currently configured on our
instance uses a fully-qualified `"provider/model"` string, so I can't offer a
config-level repro. Filing this as a defensive-programming gap regardless:
whatever produces a target without a clean `modelStr` (a malformed combo
config, or some internal expansion/pooling path that doesn't always set it),
the crash-prone call site shouldn't need well-formed input to avoid throwing,
since `ResolvedComboTarget.provider` is already available as a reliable
fallback independent of `modelStr`.

### Steps to Reproduce

1. Construct (or find, if one exists internally, e.g. via a pooled/expanded
   auto-candidate target) a `ResolvedComboTarget` whose `modelStr` does not
   parse into a clean `provider/model` pair via
   `open-sse/services/model.ts:367` `parseModel()` — it will return
   `{ provider: null, model: null, ... }`.
2. Route a chat completion through a combo containing that target.
3. `chatCore.ts` (~line 1702-1705) calls
   `getTokenLimit(parseModel(t.modelStr).provider, ...)` with `provider: null`.
4. `getTokenLimit` → `resolveTokenLimit` → `getEnvOverride(provider)`
   (`open-sse/services/contextManager.ts:27`) calls
   `provider.toUpperCase()` unconditionally.

### Expected Behavior

Combo compression-limit resolution falls back gracefully (e.g. to the
target's own `provider` field, or a safe default) when `modelStr` can't be
parsed, and does not log a warning on well-formed requests.

### Actual Behavior

```
Failed to resolve combo limits for compression: TypeError: Cannot read
properties of null (reading 'toUpperCase')
```

logged on every affected request. The surrounding `try/catch` prevents a hard
failure, but the combo-aware context limit silently isn't applied for that
request.

### Test Impact

Existing automated test already fails / Needs a new unit test (unsure which —
I didn't find an existing test exercising this exact call site under
`tests/unit/`)

### Additional Context

`ResolvedComboTarget` (`open-sse/services/combo/types.ts:159-180`) always
carries a `provider: string` field, populated directly by
`resolveComboTargets` independent of `modelStr`. The compression-limit call
site ignores it.

### Suggested fix

```ts
comboTargetLimits = targets.map((t: { modelStr?: string; provider?: string }) => {
  const parsed = parseModel(t.modelStr);
  return getTokenLimit(parsed.provider ?? t.provider ?? "unknown", parsed.model);
});
```

### Validation Plan

- Add a unit test constructing a `ResolvedComboTarget` with a `modelStr` that
  doesn't parse (e.g. missing provider prefix) and asserting
  `getTokenLimit`/the compression-limit block doesn't throw.
- Live: send a chat completion through an affected combo and confirm the
  "Failed to resolve combo limits for compression" warning no longer appears
  in logs.

Dingding-leo:
## Summary
Fixes #8708 and #8716.

## Root Cause & Solution
1. **Target Provider Resolution (#8708)**:
   In `open-sse/handlers/chatCore.ts`, when mapping `targets` to compute `comboTargetLimits`, the code called `parseModel(t.modelStr)` and used `parsed.provider`. Resolved combo targets explicitly set `t.provider` (e.g. from `providerId`). Skipping `t.provider` caused custom target providers to fall back to `parsed.providerAlias` or default limits.
   **Fix**: Update fallback chain to `t.provider || parsed.provider || parsed.providerAlias || "unknown"`.

2. **Null/Undefined Provider Guard (#8716)**:
   In `open-sse/services/contextManager.ts`, `getTokenLimit` and internal helper `getEnvOverride` assumed `provider` was non-null string, which could throw type errors when passed `null` or `undefined` target providers.
   **Fix**: Accept `string | null | undefined` in `getTokenLimit`, sanitize to `safeProvider = provider || "unknown"`, and guard `getEnvOverride` against falsy values.

## Verification
- Added unit test `tests/unit/combo-target-limits-provider.test.ts` verifying `t.provider` override is preserved when calculating target token limits.
- Validated core typecheck with `npx tsc --pretty false -p tsconfig.typecheck-core.json` (0 errors).

### DIFF
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -1706,7 +1706,7 @@ export async function handleChatCore({
           );
           comboTargetLimits = targets.map((t: { modelStr?: string; provider?: string }) => {
             const parsed = parseModel(t.modelStr);
-            const provider = t.provider || parsed.provider || parsed.providerAlias;
+            const provider = t.provider || parsed.provider || parsed.providerAlias || "unknown";
             return getTokenLimit(provider, parsed.model);
           });
         }
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -1704,9 +1704,10 @@ export async function handleChatCore({
             comboConfig as unknown as { name: string; models: unknown[] },
             allCombosData as unknown as { name: string; models: unknown[] }[]
           );
-          comboTargetLimits = targets.map((t: { modelStr?: string }) => {
+          comboTargetLimits = targets.map((t: { modelStr?: string; provider?: string }) => {
             const parsed = parseModel(t.modelStr);
-            return getTokenLimit(parsed.provider, parsed.model);
+            const provider = t.provider || parsed.provider || parsed.providerAlias;
+            return getTokenLimit(provider, parsed.model);
           });
         }
         // chatCore executes per concrete target (handleSingleModel resolves
--- a/open-sse/services/contextManager.ts
+++ b/open-sse/services/contextManager.ts
@@ -23,7 +23,8 @@ const DEFAULT_LIMITS: Record<string, number> = {
 };
 
 // Environment variable overrides (highest priority)
-function getEnvOverride(provider: string): number | null {
+function getEnvOverride(provider: string | null | undefined): number | null {
+  if (!provider) return null;
   const envKey = `CONTEXT_LENGTH_${provider.toUpperCase().replace(/[^A-Z0-9]/g, "_")}`;
   const envValue = process.env[envKey];
   if (envValue) {
@@ -191,7 +192,10 @@ export function estimateTokens(text: string | object | null | undefined): number
  * Get token limit for a provider/model combination
  * Priority: Env override > models.dev DB > Registry defaultContextLength > DEFAULT_LIMITS
  */
-export function getTokenLimit(provider: string, model: string | null = null): number {
+export function getTokenLimit(
+  provider: string | null | undefined,
+  model: string | null = null
+): number {
   return resolveTokenLimit(provider, model).limit;
 }
 
@@ -202,23 +206,24 @@ export function getTokenLimit(provider: string, model: string | null = null): nu
  * catch-all default.
  */
 function resolveTokenLimit(
-  provider: string,
+  provider: string | null | undefined,
   model: string | null = null
 ): { limit: number; specific: boolean } {
+  const safeProvider = provider || "unknown";
   // 1. Check environment variable override first
-  const envOverride = getEnvOverride(provider);
+  const envOverride = getEnvOverride(safeProvider);
   if (envOverride) return { limit: envOverride, specific: true };
 
   const lowerModel = (model || "").toLowerCase();
 
   // 2. Check models.dev synced DB for per-model context limit
   if (model) {
-    const dbLimit = getModelContextLimit(provider, model);
+    const dbLimit = getModelContextLimit(safeProvider, model);
     if (dbLimit && dbLimit > 0) return { limit: dbLimit, specific: true };
   }
 
   // 3. Check registry for provider default
-  const registryEntry = REGISTRY[provider];
+  const registryEntry = REGISTRY[safeProvider];
   if (registryEntry?.defaultContextLength) {
     return { limit: registryEntry.defaultContextLength, specific: true };
   }
@@ -238,7 +243,7 @@ function resolveTokenLimit(
   }
 
   // 5. Fallback to DEFAULT_LIMITS or default
-  if (DEFAULT_LIMITS[provider]) return { limit: DEFAULT_LIMITS[provider], specific: true };
+  if (DEFAULT_LIMITS[safeProvider]) return { limit: DEFAULT_LIMITS[safeProvider], specific: true };
   return { limit: DEFAULT_LIMITS.default, specific: false };
 }
 
--- a/tests/unit/combo-target-limits-provider.test.ts
+++ b/tests/unit/combo-target-limits-provider.test.ts
@@ -1,0 +1,27 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import { resolveComboTargets } from "../../open-sse/services/combo/comboStructure.ts";
+import { getTokenLimit } from "../../open-sse/services/contextManager.ts";
+import { parseModel } from "../../open-sse/services/model.ts";
+
+test("comboTargetLimits resolution respects t.provider override", () => {
+  const comboConfig = {
+    name: "test-combo",
+    models: [
+      { model: "openai/gpt-4o", providerId: "custom-provider-id" },
+      { model: "anthropic/claude-3-5-sonnet" },
+    ],
+  };
+  const targets = resolveComboTargets(comboConfig, null);
+
+  const comboTargetLimits = targets.map((t: { modelStr?: string; provider?: string }) => {
+    const parsed = parseModel(t.modelStr);
+    const provider = t.provider || parsed.provider || parsed.providerAlias;
+    return getTokenLimit(provider, parsed.model);
+  });
+
+  assert.equal(targets.length, 2);
+  assert.equal(targets[0].provider, "custom-provider-id");
+  assert.equal(typeof comboTargetLimits[0], "number");
+  assert.equal(comboTargetLimits[1], 200000);
+});


maxmad64bis:
Closing #8708 with this PR is incorrect.

#8708 is about `src/lib/guardrails/piiMasker.ts`: `isRequestPiiMaskingEnabled()` reads `process.env.PII_REDACTION_ENABLED` directly instead of the DB-backed `isFeatureFlagEnabled("PII_REDACTION_ENABLED")`, so the dashboard toggle has no effect.

This PR only touches `open-sse/handlers/chatCore.ts` and `open-sse/services/contextManager.ts` (combo target limits / provider fallback). It does not modify `piiMasker.ts` or any feature-flag resolution path, so it does not fix #8708.

diegosouzapw:

`release/v3.8.49` has moved **117 commits** ahead of the base this PR was last built against, and several gate failures that were red on the older base have since been repaired on the branch itself:

| Gate class | Repaired by |
|---|---|

Currently failing here:

- `Fast Quality Gates`

diegosouzapw:

diegosouzapw:
Obrigado pelo PR! O escopo tem sobreposição com outra correção e inclui claim  que, conforme validação, não é coberta por este diff. Isso poderia fechar issue errada.\n\nRecomendação: fechar e reabrir uma versão focada em #8716 (sem alterar escopo), com testes direcionados. Obrigado por contribuir.

diegosouzapw:
Closed por solicitação do maintainer: sobreposição de escopo com outras PRs e claim de issue incorreto (não cobre #8708). Para manter rastreio limpo, reabrir nova PR focada apenas no escopo de #8716.

PR #8721, opened 2026-07-26
fix(reasoning): sanitize Kimi Code K3 think tags

rinseaid:
Kimi Code K3 emits textual <think> wrappers into OpenAI-format content. Treat the `kimi-coding/k3` route as a textual-reasoning route so the existing sanitizer moves it to `reasoning_content` instead of exposing raw tags.

Validation: `node --import tsx/esm --test tests/unit/responsesanitizer-reasoning-split.test.ts`.


diegosouzapw:

PR #8763, opened 2026-07-27
fix(cloudflare-ai): sync missing free catalog models (fixes #8725)

### LINKED ISSUE #8725: fix(backend): dashboard autoSync toggle writes only to lowest-priority connection — multi-connection gap from #597
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

_No response_

### Node.js Version

22.22.3

### Provider(s) Involved

Any provider with more than one active connection (reproduced on `cloudflare-ai` with 2 connections)

### Model(s) Involved

_No response_

### Client Tool

Dashboard UI + direct `GET/PUT /api/providers` calls (cURL)

### Description

The provider detail page (`/dashboard/providers/[id]`) lists every connection for a provider together and shows a single "Auto Sync" toggle. This reads as a per-provider setting, but it actually only reads/writes **the single connection with the lowest `priority` value** (i.e. the highest-priority active connection) — never the others, for as long as that connection stays active.

Root cause (`src/app/(dashboard)/dashboard/providers/[id]/hooks/useModelImportHandlers.ts`):

```ts
// line 102
const autoSyncConnection = connections.find((conn) => conn.isActive !== false);
const isAutoSyncEnabled = !!(autoSyncConnection as any)?.providerSpecificData?.autoSync;

// line 376-397
const handleToggleAutoSync = async () => {
  if (!autoSyncConnection || togglingAutoSync) return;
  ...
  await fetch(`/api/providers/${(autoSyncConnection as any).id}`, {
    method: "PUT",
    body: JSON.stringify({ providerSpecificData: { autoSync: newValue } }),
  });
  ...
};
```

`connections` here is not in arbitrary/insertion order — it comes from `getProviderConnections()` (`src/lib/db/providers.ts:185`), which explicitly does `ORDER BY priority ASC, updated_at DESC`. So `.find(isActive !== false)` deterministically resolves to **the same one connection every time** — the lowest-`priority` active connection — reusing the `priority` field that elsewhere drives account-fallback routing order, for a purpose (auto-sync scope) it was never intended for.

Backend support for multiple synced connections already exists — #597 explicitly refactored `modelSyncScheduler` to "dynamically read **connections** with `autoSync` enabled" (plural), and `GET /api/providers` / `PUT /api/providers/[id]` operate per-connection (`providerSpecificData.autoSync`, confirmed via direct API calls). The gap is purely in this UI wiring, which assumes one connection per provider.

Consequence for multi-account setups: toggling "Auto Sync" in the dashboard for a provider with N connections only ever touches the lowest-`priority` connection. The other N-1 connections keep whatever `autoSync` state they had (often `null`/off since a freshly created connection never has it — see admin docs), with **no UI affordance to reach them** — the toggle for the other connections is simply not exposed anywhere in the dashboard. The only way to detect or fix the gap is a direct `GET`/`PUT /api/providers` call, not the dashboard.

### Steps to Reproduce

1. Create two connections for the same provider (e.g. `cloudflare-ai`) with different accounts/API keys, giving them different `priority` values (default behavior when created in sequence).
2. Open `/dashboard/providers/[id]` for that provider; note only one "Auto Sync" toggle is shown for the whole page.
3. Toggle it on, then call `GET /api/providers` (or inspect the DB) and check `providerSpecificData.autoSync` for **both** connections.
4. Observe only the lowest-`priority` connection was updated; the other's `providerSpecificData.autoSync` is unchanged.
5. There is no UI control to toggle `autoSync` on the other connection — the only way is a direct `PUT /api/providers/{connectionId}`.

### Expected Behavior

Either:

- (a) the toggle is clearly scoped to a single connection (with a connection picker/label, one toggle rendered per connection), or
- (b) it genuinely applies to every active connection of the provider (fan-out `PUT` to all connection IDs when toggled), matching what the "per-provider" framing and #597's plural `modelSyncScheduler` change already imply.

Either way, every connection of a provider should be reachable/auditable from the dashboard for this setting.

### Actual Behavior

The toggle silently applies to a single connection — the one with the lowest `priority` value among active connections (`connections.find(isActive !== false)` over a `priority ASC`-ordered list) — giving no indication that sibling connections were left untouched, and no way to reach them from the UI.

### Test Impact

Needs a new unit test

### Error Logs / Output

_No response — this is a silent logic gap, not an error path._

### Screenshots

_No response_

### Additional Context

Related tables (for context on why this matters beyond the toggle itself, read while investigating): `customModels` (`src/lib/db/models.ts`) is keyed by `providerId` alone — one shared catalog row per provider, replaced wholesale on every sync (`replaceCustomModels`, not merged) — while `syncedAvailableModels` is keyed by `providerId:connectionId`. This is a separate, working-as-designed mechanism; flagging only so a fix doesn't conflate "which connection's toggle" with "which connection's model list wins on sync".

This exact gap was already flagged during review of the PR that introduced the feature — #597 (2026-03-25), automated review comment:

> Multiple connections scenario - the toggle only reads/updates autoSync from the first active connection (`connections.find()`). Users with multiple connections won't have per-connection auto-sync control.

It was logged as a WARNING alongside 4 other findings and the PR merged without addressing it — this isn't a regression, it's the original, known-at-review-time behavior, four months old.

### Validation Plan

- New unit/integration test on `useModelImportHandlers` (or its future replacement) asserting that toggling auto-sync with 2+ connections for the same provider either updates all of them or exposes a per-connection control — whichever direction (a)/(b) above is chosen.
- Manual: create 2 connections for one provider, toggle in UI, `GET /api/providers` before/after, assert both connections' `providerSpecificData.autoSync` match the intended behavior.

Dingding-leo:
## Summary
Fixes #8725.

## Changes
- Adds , , , , , , and  to .
- Ensures parity with  so these models properly display and filter under free models.


maxmad64bis:
This PR is not fixing issue #8725 (nothing in common).

PR #8790, opened 2026-07-27
[v3.8.50] feat(combo): add maxContextWindow to contextRequirements (fixes #8777)

### LINKED ISSUE #8777: fix(backend): static fitness table freezes newer models at legacy scores
## Summary

`FITNESS_TABLE` (`open-sse/services/autoCombo/taskFitness.ts:24`) is a hand-maintained
substring→score map used as resolution layer 4 of `getTaskFitnessWithSource()`. Measured
against the real provider catalog it no longer describes the models it is routing:

- **3 rows are dead** — they match nothing in the catalog at all.
- **88% of its hits come from version-less family patterns** (`qwen`, `llama`,
  `claude-opus`, …) that collapse every generation of a family onto one 2024-era number.
- **Legacy flagships occupy the top of the scale**, so newer models systematically lose
  to models that are one to three years older.
- **Bare 2-character patterns collide by accident** — `solar-pro3` scores as OpenAI `o3`.

#8651 fixed the *ordering* of this lookup (longest pattern first). This issue is about the
table's *contents*, which that fix does not touch.

## Verified against the real catalog

Enumerating every `id` + `alias` in `REGISTRY` (`open-sse/config/providers/index.ts`) and
running the merged lookup (longest-pattern-first) over `FITNESS_TABLE.coding`:

```
catalog model ids                  1160
hit the static table                461  (40%)
fell through to wildcard            699  (60%)
of the 461 hits, matched via a
version-less / family pattern       405  (88%)
```

### Defect 1 — dead rows

Match zero ids in the catalog:

```
o1          = 0.93
mixtral     = 0.77
grok-4-fast = 0.80
```

Harmless today (they match nothing) but they are the visible half of the rot, and
`grok-4-fast` still carries a stale justification comment (`ultra-low latency (1143ms P50)`)
for a model that is no longer in the catalog.

### Defect 2 — version-less family patterns

One number per family, applied to every generation:

| pattern | score | ids captured | includes |
| --- | --- | --- | --- |
| `qwen` | 0.78 | 92 | `qwen3.7-max`, `qwen3.6-27b`, `qwen2.5-coder-32b`, `qwen-sea-lion-v4` |
| `llama` | 0.72 | 67 | `llama-3.3-70b`, `aion-rp-llama-3.1-8b` (a roleplay finetune) |
| `claude-opus` | 0.92 | 63 | `claude-opus-4.1` … `claude-opus-4.7` |
| `claude-sonnet` | 0.95 | 32 | `claude-sonnet-4` … `claude-sonnet-4.6` |
| `mistral` | 0.75 | 26 | `codestral-2501`, `devstral-2-123b` (both dedicated coding models) |
| `grok-4` | 0.82 | 26 | `grok-4.20`, `grok-4-5-high` |
| `kimi-k2` | 0.82 | 25 | `kimi-k2.7-code`, `kimi-k2.6` |
| `codex` | 0.98 | 22 | `gpt-5-codex` … `gpt-5.3-codex-xhigh` |
| `glm-5` | 0.78 | 22 | `glm-5.2`, `glm-5-2-max` |

There is no more-specific row for any of these, so longest-pattern-first cannot help — the
table has no version dimension to select on.

### Defect 3 — the scale is inverted with respect to time

Newer models score below older ones, including below their own predecessors:

```
minimax-m2.7               -> "minimax-m2"    0.72   vs  minimax-m2.5 -> 0.75   (newer < older sibling)
qwen3.7-max                -> "qwen"          0.78   vs  gpt-4-turbo  -> 0.88   (2026 < 2023-11)
claude-opus-4.7            -> "claude-opus"   0.92   vs  o3           -> 0.95   (2026 < 2025-04)
mistral-ai/codestral-2501  -> "mistral"       0.75   vs  gpt-4o       -> 0.90   (coding model < general model)
glm-5.2                    -> "glm-5"         0.78   vs  glm-5.1      -> 0.78   (indistinguishable)
```

The cause is structural, not carelessness: new rows were added conservatively into the
0.70–0.82 band because nobody wanted to renumber the existing entries, while the legacy
flagships kept their original 0.88–0.98. The result is a systematic bias where **the longer
a pattern has been in the table, the higher it scores.**

Note also that layer 4 returns before `lookupWildcardBoosts()`, so the `coder`/`code`
boosts (+0.15/+0.10) never fire for `codestral-2501` or `kimi-k2.7-code` — the family
pattern short-circuits them.

### Defect 4 — accidental substring collisions

```
solar-pro3  -> matched "o3" -> 0.95    (Upstage Solar Pro 3, scored as OpenAI o3)
gpt_o3_mini -> matched "o3" -> 0.95
o3-mini     -> matched "o3" -> 0.95    (o3-mini scored as full o3)
```

`o3` is a 2-character pattern matched with `String.includes` against arbitrary vendor ids.
Any id containing the literal `o3` inherits 0.95.

### Reproduction

```ts
// enumerate catalog ids
import { REGISTRY } from "open-sse/config/providers/index.ts";
const ids = new Set<string>();
for (const [, e] of Object.entries(REGISTRY))
  for (const m of e.models ?? []) {
    if (m?.id) ids.add(m.id.toLowerCase());
    for (const a of m?.aliases ?? []) ids.add(a.toLowerCase());
  }

// run the merged lookup (taskFitness.ts:304-311)
const entries = Object.entries(FITNESS_TABLE.coding).sort((a, b) => b[0].length - a[0].length);
for (const id of ids)
  for (const [pattern, score] of entries)
    if (id.includes(pattern)) { console.log(id, "->", pattern, score); break; }
```

## Blast radius

`taskFit` carries weight `0.08` in `DEFAULT_WEIGHTS` (`scoring.ts:43`) and `0.37` in the
`quality-first` mode pack (`modePacks.ts:46`) — the mode most exposed to the inversion is
precisely the one whose contract is "give me the best model".

Layer 4 only fires when `user_override`, `arena_elo`, and `models_dev_tier` all miss, which
bounds the damage — but that is exactly the long tail the router most needs to reason about,
and 60% of the catalog does not even reach layer 4.

## Proposed fix

Three changes, all subtractive, no new dependency:

1. **Delete the dead rows** — `o1`, `mixtral`, `grok-4-fast` (and their stale comments)
   across all task categories.
2. **Delete the version-less family patterns** — `qwen`, `llama`, `mistral`,
   `claude-sonnet`, `claude-opus`, `claude-haiku`, `codex`, `kimi-k2`, `grok-4`, `grok-3`,
   `glm-5`, `minimax-m2`, and the bare `o1`/`o3`. Their information content is negative: they
   assert a specific quality claim about models that did not exist when the number was
   written, and they suppress the wildcard boosts that would otherwise fire.
3. **Collapse what remains into a narrow neutral band** (e.g. 0.60–0.70). Layer 4 should
   signal "this is a recognised model" and nothing stronger. Keeping absolute quality claims
   in a hand-maintained table is what produced the inversion; renumbering it would only reset
   the clock (same argument as #8602).

Requiring a minimum pattern length (e.g. ≥ 4 chars) would also prevent the `o3`-class
collision from recurring if short patterns are ever re-added.

### Why not just refresh the numbers

Because the rot is structural. Synced sources in `model_intelligence` carry `expires_at`
and auto-invalidate (`097_model_intelligence.sql`); the hardcoded table has no TTL and
never expires. Any refreshed set of numbers is stale on the next model release. Sourcing
layer 2 from a benchmark that updates itself is tracked separately in **#8778** — this issue is
the part that can ship immediately, with no external dependency or licensing decision.

## Test plan

`tests/unit/` (or the existing `open-sse/services/autoCombo/__tests__/autoCombo.test.ts`):

- **Regression guard for the inversion**: assert `getTaskFitness("minimax-m2.7", "coding") >= getTaskFitness("minimax-m2.5", "coding")` and that no catalog model resolves to a
  `fitness_table` score above the neutral band ceiling.
- **No accidental collisions**: assert `getTaskFitnessWithSource("solar-pro3", "coding").source !== "fitness_table"`.
- **Family patterns gone**: assert `getTaskFitnessWithSource("claude-opus-4.7", "coding")` no
  longer resolves via `fitness_table` when layers 1–3 miss.
- **Wildcard boosts reachable**: assert `codestral`/`-code` ids now receive the `coder`/`code`
  boost instead of being short-circuited.
- **Table hygiene**: a table-driven test asserting every `FITNESS_TABLE` pattern matches at
  least one id in `REGISTRY`, so a dead row fails CI instead of accumulating silently.

The last one is the durable guard — it makes the rot detectable rather than something that
has to be re-audited by hand.

## Relationship to existing issues

- **#8602** (open) — `DEFAULT_TASK_MODEL_MAP` in `taskAwareRouter.ts` hardcodes stale
  *routing destinations*. Same class of rot, different location. Not a duplicate.
- **#8603** (closed) — the substring *ordering* bug in the same function. Fixed by **#8651**
  (longest-pattern-first). Note #8603 was auto-closed by **#8636** (`-free` suffix
  normalisation), which is unrelated to the ordering fix; #8651 merged 42 seconds later and
  did carry the fix. The ordering fix is in `release/v3.8.49`; this issue is the remaining
  half it does not address.
- **#8664** duplicates the already-merged #8651 and can be closed.

Dingding-leo:
## Summary
Fixes #8777.

Adds `maxContextWindow` support to `contextRequirements` in combo target filtering and validation schemas.

## Changes
- **Validation Schema**: Added `maxContextWindow` (number between 0 and 10,000,000) to `comboRuntimeConfigSchema` in `src/shared/validation/schemas/combo.ts` and `DEFAULT_COMBO_CONFIG` in `open-sse/services/comboConfig.ts`.
- **Target Filtering**: Updated `applyContextRequirements` in `open-sse/services/combo/contextRequirements.ts` to filter out targets with context windows exceeding `maxContextWindow`. Respects `contextFilterMode` (`strict` excludes models with unknown limits, `lenient` includes them).
- **Tests**: Added unit & integration test coverage in `tests/unit/combo-context-requirements.test.ts` and `tests/unit/combo/context-requirements-integration.test.ts`.


diegosouzapw:

`release/v3.8.49` has moved **93 commits** ahead of the base this PR was last built against, and several gate failures that were red on the older base have since been repaired on the branch itself:

| Gate class | Repaired by |
|---|---|

Currently failing here:

- `Fast Quality Gates`

diegosouzapw:

diegosouzapw:
Heads-up on the issue reference, not on the change itself: this PR says it fixes #8777, but #8777 is about `FITNESS_TABLE` rot in `open-sse/services/autoCombo/taskFitness.ts` (dead rows, version-less family patterns, short-pattern collisions). This diff adds `maxContextWindow` to `contextRequirements` (`open-sse/services/combo/contextRequirements.ts`, `comboConfig.ts`, `schemas/combo.ts`) and never touches `taskFitness.ts` — a different subsystem, so it looks like a wrong issue number.

The change itself looks fine and stands on its own; could you update the reference to the issue it actually addresses (or drop the `fixes` line)? Left as-is, merging this would auto-close #8777 while the fitness-table problem is still open.

For context: I re-triaged #8777 today and confirmed two of its defects with a reproduction test — `getStaticFitnessTableScore("solar-pro3")` returns 0.95 because the bare `o3` pattern matches by substring, and `minimax-m2.7` scores *below* its older sibling `minimax-m2.5`. So that issue does need its own fix.

PR #8797, opened 2026-07-27
fix(analytics): fix ModelTable sorting for Model column and numeric values (fixes #8769)

### LINKED ISSUE #8769: [BUG] Usage page - Model Breakdown section - Sort does not work
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

win10

### Node.js Version

node 22

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

Usage page: sorting does not work for Model BreakDown section

### Steps to Reproduce

1. Go to Usage page
2. Jump to Model Breakdown section
3. click each column headers to sort

### Expected Behavior

3. sort by the clicked column header

### Actual Behavior

3. sort does not happen

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

Image

### Additional Context

_No response_

### Validation Plan

_No response_

Dingding-leo:
## Summary
Fixes #8769.

## Changes
- In `ModelTable` (`src/shared/components/analytics/charts.tsx`), sorting by the `"model"` column did not evaluate `a[sortBy]` properly as the object field could be `model` or `name` (`a.model || a.name`).
- Updated `ModelTable` sort logic to fallback between `a.model` and `a.name` when `sortBy === "model"`, and safely coerce numeric/string comparisons.

### DIFF
--- a/src/shared/components/analytics/charts.tsx
+++ b/src/shared/components/analytics/charts.tsx
@@ -619,9 +619,12 @@ export function ModelTable({ byModel, summary }) {
     arr.sort((a, b) => {
       const va = a[sortBy] ?? 0;
       const vb = b[sortBy] ?? 0;
-      if (typeof va === "string")
-        return sortOrder === "asc" ? va.localeCompare(vb) : vb.localeCompare(va);
-      return sortOrder === "asc" ? va - vb : vb - va;
+      if (sortBy === "model" || typeof va === "string" || typeof vb === "string") {
+        return sortOrder === "asc"
+          ? String(va).localeCompare(String(vb))
+          : String(vb).localeCompare(String(va));
+      }
+      return sortOrder === "asc" ? Number(va) - Number(vb) : Number(vb) - Number(va);
     });
     return arr;
   }, [byModel, sortBy, sortOrder]);


diegosouzapw:
## Closing — the same bug is fixed by #8802, now merged

Thanks for spotting this — the `ModelTable` sort was genuinely broken, and #8769 was a real report.

Holding this open until #8802 landed was deliberate: closing a duplicate before its replacement merges would have left the bug uncovered. #8802 is now merged into `release/v3.8.49`, so the fix is live.

### Why #8802 was taken instead

Both PRs target the same lines of `charts.tsx` and would conflict if applied together. #8802 goes further in two ways that matter:

1. It extracts the sort into a pure, testable `modelBreakdownSort.ts` instead of fixing it inline — so the logic is reachable without mounting the component.
2. It ships `tests/unit/model-breakdown-sort-8769.test.ts` with 7 cases. This PR changed production code in `src/` with no test, which the project's PR rule requires.

As a side effect, the extraction also shrinks `charts.tsx`, which is under the file-size ratchet.

Closing as superseded — not because the diagnosis was wrong. If you find a sorting case that #8802 still gets wrong, a focused PR against `modelBreakdownSort.ts` (now easy to unit-test) would be very welcome.

PR #8805, opened 2026-07-27
fix(api): treat cx/* and codex/* as equivalent API-key model permissions

### LINKED ISSUE #8803: [BUG] Dashboard cx/* restrictions do not allow bare Codex model IDs normalized to codex/*
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Ubuntu 24.04.4

### Node.js Version

v22.23.1

### Provider(s) Involved

Codex

### Model(s) Involved

All Codex models exposed as cx/* (reproduced with gpt-5.6-terra)

### Client Tool

Codex CLI (also reproducible with direct curl requests)

## Description

The API key permissions Dashboard exposes Codex models using the `cx/` prefix.

Requests explicitly using the same `cx/*` model identifier work correctly. For example, a key restricted to `cx/gpt-5.6-terra` can successfully request `cx/gpt-5.6-terra`.

However, when the client sends the equivalent bare model ID, such as `gpt-5.6-terra`, `/v1/responses` resolves it to `codex/gpt-5.6-terra` before API key model permission evaluation.

The Dashboard restriction contains `cx/gpt-5.6-terra`, while the permission check evaluates `codex/gpt-5.6-terra`. These identifiers are treated as different strings, so the request is rejected with `403 Forbidden`.

## Impact

Restricted API keys configured through the Dashboard cannot be used with bare Codex model IDs, even though the bare ID resolves to the same Codex provider and model.

This particularly affects Codex CLI configurations using model IDs such as:

`gpt-5.6-terra`

The current workarounds are to:

- configure the client to explicitly use the `cx/*` model identifier;
- allow all models;
- manually store a matching `codex/*` permission through the management API.

## Steps to Reproduce

1. Connect a Codex OAuth account.

2. Create an API key in the Dashboard.

3. Restrict the API key to the model exposed by the Dashboard:

```text
cx/gpt-5.6-terra
```

4. Send a request using the explicitly prefixed model ID:

```bash
curl -i 'https://example.com/v1/responses' \
  -H 'Authorization: Bearer <restricted-api-key>' \
  --data '{
    "model": "cx/gpt-5.6-terra",
    "input": "Reply with exactly: OK",
    "stream": false,
    "reasoning": {
      "effort": "medium"
    }
  }'
```

This request succeeds with:

```text
HTTP 200 OK
```

5. Send the equivalent request using the bare model ID:

```bash
curl -i 'https://example.com/v1/responses' \
  -H 'Authorization: Bearer <restricted-api-key>' \
  --data '{
    "model": "gpt-5.6-terra",
    "input": "Reply with exactly: OK",
    "stream": false,
    "reasoning": {
      "effort": "medium"
    }
  }'
```

This request is rejected:

```text
HTTP 403 Forbidden
```

```json
{
  "error": {
    "message": "Model \"codex/gpt-5.6-terra\" is not allowed for this API key",
    "type": "permission_error",
    "code": "insufficient_quota"
  }
}
```

The client sent:

```text
gpt-5.6-terra
```

but the permission error shows that the request was evaluated as:

```text
codex/gpt-5.6-terra
```

6. Manually change the same API key permission through the management API to:

```json
{
  "allowedModels": [
    "codex/gpt-5.6-terra*"
  ]
}
```

7. Repeat the request using the bare model ID.

The request now succeeds with:

```text
HTTP 200 OK
```

The response headers confirm that the request is routed through the Codex provider:

```text
x-omniroute-model: gpt-5.6-terra
x-omniroute-provider: cx
```

This demonstrates that:

- `cx/*` permissions work for requests explicitly using `cx/*`;
- bare Codex model IDs are resolved to `codex/*`;
- the Dashboard does not provide a matching `codex/*` permission;
- the failure occurs during API key model permission evaluation.

## Expected Behavior

An API key restricted to a Codex model through the Dashboard should also allow the equivalent bare model ID when OmniRoute resolves that ID to the same Codex provider and model.

For example, selecting:

```text
cx/gpt-5.6-terra
```

should allow both:

```text
cx/gpt-5.6-terra
gpt-5.6-terra
```

when the bare model ID is internally resolved to the same Codex model.

This could be implemented by normalizing provider aliases during permission evaluation, by storing canonical permission identifiers, or by exposing matching `codex/*` identifiers in the Dashboard.

## Actual Behavior

A Dashboard restriction using `cx/gpt-5.6-terra` allows requests that explicitly use `cx/gpt-5.6-terra`.

The equivalent bare request using `gpt-5.6-terra` is resolved to `codex/gpt-5.6-terra` and rejected because the permission matcher does not match it against the stored `cx/gpt-5.6-terra` restriction.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
HTTP/1.1 403 Forbidden
Server: nginx/1.24.0 (Ubuntu)
Date: Mon, 27 Jul 2026 21:25:21 GMT
Transfer-Encoding: chunked
Connection: keep-alive
Content-Security-Policy: default-src 'self'; base-uri 'self'; object-src 'none'; frame-ancestors 'none'; form-action 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; font-src 'self' https://fonts.gstatic.com data:; img-src 'self' data: blob: https:; media-src 'self' data: blob:; connect-src 'self' http://localhost:* http://127.0.0.1:* ws://localhost:* ws://127.0.0.1:* https: ws: wss:; worker-src 'self' blob:; manifest-src 'self'
X-Frame-Options: DENY
X-Content-Type-Options: nosniff
Referrer-Policy: strict-origin-when-cross-origin
Permissions-Policy: camera=(), microphone=(), geolocation=(), payment=(), usb=(), serial=()
Strict-Transport-Security: max-age=63072000; includeSubDomains; preload access-control-allow-headers: Content-Type, Authorization, x-api-key, anthropic-version, x-omniroute-connection, x-internal-test, accept access-control-allow-methods: GET, POST, PUT, DELETE, PATCH, OPTIONS
access-control-allow-origin: *
vary: Origin
vary: rsc, next-router-state-tree, next-router-prefetch, next-router-segment-prefetch
x-omniroute-route-class: CLIENT_API
x-request-id: e2db50b9-35df-4e4c-a467-f5eb0aed8fde

curl: (22) The requested URL returned error: 403 {
  "error": {
    "message": "Model \"codex/gpt-5.6-terra\" is not allowed for this API key",
    "type": "permission_error",
    "code": "insufficient_quota"
  }
}
```

After changing `allowedModels to `codex/gpt-5.6-terra*`:

```
HTTP/1.1 200 OK
x-omniroute-model: gpt-5.6-terra
x-omniroute-provider: cx
```

### Screenshots

Image

The attached screenshot shows that the Dashboard model selector exposes Codex models using the `cx/*` prefix.

## Additional Context

This appears to be a mismatch between the model identifier exposed by the Dashboard and the identifier produced when `/v1/responses` resolves a bare Codex model ID.

Observed behavior:

```text
Dashboard permission:  cx/gpt-5.6-terra
Explicit cx/* request: allowed

Bare request:          gpt-5.6-terra
Resolved permission ID: codex/gpt-5.6-terra
Result:                denied
```

Possible solutions include:

1. Treat the registered provider ID `codex` and its alias `cx` as equivalent during API key model permission evaluation.

2. Expose the canonical `codex/*` model identifiers in the Dashboard permission selector.

3. Store a canonical permission identifier regardless of whether the Dashboard entry was selected through a provider alias.

The most appropriate implementation depends on the intended provider-alias semantics in OmniRoute.

It may also be worth checking whether the same mismatch affects `blockedModels`.

## Validation Plan

Add focused unit tests for Codex model permission handling.

Required cases:

```text
allowed:  cx/gpt-5.6-terra
request:  cx/gpt-5.6-terra
expected: allowed
```

```text
allowed:          cx/gpt-5.6-terra
client request:   gpt-5.6-terra
resolved request: codex/gpt-5.6-terra
expected:         allowed
```

```text
allowed:          codex/gpt-5.6-terra
client request:   gpt-5.6-terra
resolved request: codex/gpt-5.6-terra
expected:         allowed
```

```text
allowed:  cx/gpt-5.6-terra
request:  openai/gpt-5.6-terra
expected: denied
```

The test should confirm that alias handling does not grant access to an unrelated provider with the same model name.

The relevant existing unit tests, lint checks, and type checks should continue to pass.

Prudhvivuda:
## Summary
- Dashboard Codex restrictions use the public `cx/` alias, while `/v1/responses` normalizes bare Codex model IDs to `codex/` before API-key permission checks.
- Expand `getModelPermissionCandidates` via the provider registry alias map so `cx/*` and `codex/*` match for both `allowedModels` and `blockedModels`.
- Adds regression coverage for allow/block alias equivalence and wildcard patterns.

Closes #8803

## Test plan
- [x] `node --import tsx/esm --test tests/unit/db-apiKeys-crud.test.ts`
- [x] `node --import tsx/esm --test tests/unit/apikeys-model-permissions-split.test.ts`
- [ ] Restrict an API key to `cx/gpt-5.6-terra` in the Dashboard, then `POST /v1/responses` with bare `gpt-5.6-terra` (or `codex/gpt-5.6-terra`) and confirm 200 instead of 403
- [ ] Confirm `openai/gpt-5.6-terra` remains denied when only `cx/gpt-5.6-terra*` is allowed

### DIFF
--- a/src/lib/db/apiKeys.ts
+++ b/src/lib/db/apiKeys.ts
@@ -17,6 +17,7 @@ import {
 } from "./apiKeyUsageLimitFields";
 import { setNoLog } from "../compliance/noLog";
 import { resolveModelAlias } from "@omniroute/open-sse/services/modelDeprecation.ts";
+import { getProviderAlias, resolveProviderId } from "@/shared/constants/providers";
 import { getSyncedAvailableModelsByConnection, getCustomModels, getModelIsHidden } from "./models";
 import {
   CLAUDE_CODE_PROVIDER_PREFIXES,
@@ -24,6 +25,7 @@ import {
   stripExtendedContextSuffix,
   isPotentialUnprefixedClaudeCodeModel,
   addModelCandidate,
+  addProviderAliasScopedCandidates,
   modelPatternMatches,
   hasClaudeCodeWildcardPermission,
   matchesWildcardPattern,
@@ -306,6 +308,15 @@ async function getModelPermissionCandidates(modelId: string): Promise<string[]> 
       addModelCandidate(candidates, providerScopedModel);
       addModelCandidate(candidates, `cc/${providerScopedModel}`);
       addModelCandidate(candidates, `claude/${providerScopedModel}`);
+    }
+    if (providerScopedModel) {
+      addProviderAliasScopedCandidates(
+        candidates,
+        providerOrAlias,
+        providerScopedModel,
+        resolveProviderId,
+        getProviderAlias
+      );
     }
     return Array.from(candidates);
   }
--- a/src/lib/db/apiKeys/modelPermissions.ts
+++ b/src/lib/db/apiKeys/modelPermissions.ts
@@ -39,6 +39,29 @@ export function addModelCandidate(candidates: Set<string>, modelId: string): voi
   candidates.add(stripExtendedContextSuffix(clean));
 }
 
+/**
+ * Expand provider-scoped model ids with canonical provider id + public alias forms
+ * (e.g. codex/gpt-5.6-terra ↔ cx/gpt-5.6-terra) so API-key allow/block patterns match
+ * dashboard restrictions regardless of which prefix the client sends.
+ */
+export function addProviderAliasScopedCandidates(
+  candidates: Set<string>,
+  providerOrAlias: string,
+  providerScopedModel: string,
+  resolveProviderId: (aliasOrId: string) => string,
+  getProviderAlias: (providerId: string) => string
+): void {
+  if (!providerScopedModel) return;
+  const canonicalId = resolveProviderId(providerOrAlias);
+  const alias = getProviderAlias(canonicalId);
+  if (canonicalId !== providerOrAlias) {
+    addModelCandidate(candidates, `${canonicalId}/${providerScopedModel}`);
+  }
+  if (alias !== providerOrAlias && alias !== canonicalId) {
+    addModelCandidate(candidates, `${alias}/${providerScopedModel}`);
+  }
+}
+
 export function modelPatternMatches(pattern: string, candidates: string[]): boolean {
   for (const candidate of candidates) {
     if (pattern === candidate) return true;
--- a/tests/unit/apikeys-model-permissions-split.test.ts
+++ b/tests/unit/apikeys-model-permissions-split.test.ts
@@ -16,6 +16,7 @@ test("module exposes the matching helpers + Claude-Code alias sets", () => {
     "hasClaudeCodeWildcardPermission",
     "isPotentialUnprefixedClaudeCodeModel",
     "addModelCandidate",
+    "addProviderAliasScopedCandidates",
     "stripExtendedContextSuffix",
   ]) {
     assert.equal(typeof (M as Record<string, unknown>)[name], "function", `missing ${name}`);
--- a/tests/unit/db-apiKeys-crud.test.ts
+++ b/tests/unit/db-apiKeys-crud.test.ts
@@ -272,6 +272,43 @@ test("isModelAllowedForKey wildcard match", async () => {
   assert.equal(await apiKeys.isModelAllowedForKey(created.key, "gpt-4"), false);
 });
 
+test("isModelAllowedForKey treats cx and codex provider prefixes as equivalent (allowed cx, request codex)", async () => {
+  await resetStorage();
+  const created = await apiKeys.createApiKey("Codex Alias Allow Cx", "ma-cx-001");
+  await apiKeys.updateApiKeyPermissions(created.id, { allowedModels: ["cx/gpt-5.6-terra"] });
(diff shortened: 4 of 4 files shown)


Libsod:
Confirmed working end-to-end. With the API key restricted to `cx/gpt-5.6-terra` in the Dashboard, a `/v1/responses` request using the bare model ID `gpt-5.6-terra` now returns `HTTP 200 OK`.

Tested locally using this PR branch.

Also confirmed the negative case from the test plan:

When the API key only allows `cx/gpt-5.6-terra*`, a request using
`openai/gpt-5.6-terra` is still correctly rejected with HTTP 403 Forbidden.

So the alias expansion remains scoped to `cx/*` ↔ `codex/*` and does not
accidentally grant access to the OpenAI provider.

image

PR #8817, opened 2026-07-27
[v3.8.50] fix(db/apiKeys): respect provider parameter in group model permission checks (fixes #8803)

### LINKED ISSUE #8803: [BUG] Dashboard cx/* restrictions do not allow bare Codex model IDs normalized to codex/*

Dingding-leo:
## Summary
Fixes #8803.

## Problem
In `isModelAllowedForKey`, when a model request uses a provider prefix (e.g. `openai/gpt-4`), `modelTarget` was extracted as `gpt-4` and `provider` as `openai`. However:
1. `checkKeyModelAccess` was called as `checkKeyModelAccess(metadata.id, modelTarget)` without passing the third `provider` argument.
2. Group rules defined with a specific provider restriction (e.g. `provider: 'openai'`) were not evaluated against the provider parameter.
3. Group rules matching against the full model ID (e.g. `openai/gpt-4`) were missed because only `modelTarget` (`gpt-4`) was checked.

## Solution
1. Pass the extracted `provider` argument to `checkKeyModelAccess`.
2. Check both `modelTarget` and raw `modelId` against group rules so both full model identifier rules and provider-scoped model rules are properly checked.
3. Added a dedicated unit test in `tests/unit/group-provider-permission.test.ts` verifying provider-specific group rules.


Libsod:
Worth checking out https://github.com/diegosouzapw/OmniRoute/pull/8805

I think this addresses a separate issue in API key group
provider-scoped permissions, rather than the original issue reported in #8803.

#8803 is specifically caused by the Dashboard storing `cx/*` restrictions
while `/v1/responses` normalizes bare Codex model IDs to `codex/*`.

This branch does not appear to add `cx/*` <-> `codex/*` alias equivalence to
the model permission candidates, so the original reproduction case would
likely still return 403.

Also, checking both `modelTarget` and the full `modelId` and denying when
either result is false may reject provider-scoped allow rules. For example,
an allow rule for model pattern `gpt-4*` with provider `openai` would match
`gpt-4`, but not `openai/gpt-4`.

PR #8805 directly addresses the alias mismatch and has been manually verified
end-to-end with the original reproduction:
- `cx/gpt-5.6-terra` restriction + bare `gpt-5.6-terra` -> HTTP 200
- `openai/gpt-5.6-terra` -> HTTP 403

diegosouzapw:

`release/v3.8.49` has moved **34 commits** ahead of the base this PR was last built against, and several gate failures that were red on the older base have since been repaired on the branch itself:

| Gate class | Repaired by |
|---|---|

Currently failing here:

- `Build (advisory)`
- `Fast Quality Gates`
- `No new ESLint warnings`
- `Unit Tests fast-path (3/4)`

diegosouzapw:

PR #8858, opened 2026-07-28
[v3.8.50] fix(build): prepublish no longer spawns .cmd shims on Windows

maisdesign:
## Problem

On Windows the npm/npx entry points are `.cmd` shims, and Node >= 20 refuses to spawn a `.cmd` without a shell (`EINVAL`, from the CVE-2024-27980 hardening). On Node 24 that makes **every** `execFileSync(NPX_BIN, ...)` in `scripts/build/prepublish.ts` fail.

The failures are caught and logged as warnings, so `npm run build:cli` **exits 0 and reports a successful build** while having silently skipped:

- the MITM utilities compile
- the MCP server bundle
- the LLMLingua ONNX worker bundle
- `@omniroute/opencode-plugin`

A package published from such a build would be missing all four.

```
⚠️  MITM compile warning (non-fatal): spawnSync npx.cmd EINVAL
⚠️  MCP Server bundle error: spawnSync npx.cmd EINVAL
⚠️  LLMLingua worker bundle error: spawnSync npx.cmd EINVAL
❌ Failed to build @omniroute/opencode-plugin: spawnSync npm.cmd EINVAL
```

## Approach

`shell: true` fixes the spawn but disables argument escaping (`DEP0190`), and `CONTRIBUTING.md` already asks that shell commands not interpolate runtime values — so it is kept only as a last resort.

The preferred path resolves the tool's own JS entry point from the local dependency tree (`node_modules/<pkg>` → its `bin` field) and runs it with the current Node binary: no shim, no shell, nothing to escape. The npm CLI is resolved the same way, from the npm bundled next to the running Node. `tsup` is not a root dependency, so that call falls through to the Node-resolved `npx` entry point after the plugin's own `npm install`.

## Verification

Windows 11, Node v24.18.0.

**Before:** the four sub-steps above fail with `EINVAL`.

**After:** `npm run build:cli` exits 0 and produces

| artifact | size |
| --- | --- |
| `dist/open-sse/mcp-server/server.js` | 4.0 MB |
| `dist/open-sse/services/compression/engines/llmlingua/onnxWorker.js` | 4.0 KB |
| `@omniroute/opencode-plugin/dist/` | 704 KB |

The resulting package installs and passes its own health check: `omniroute --version` → `3.8.49`, and `better-sqlite3` / `wreq-js` resolve in both the root and `dist` scopes.

## One failure this does not fix — and why

The MITM compile still fails, but for a **different, pre-existing** reason that `EINVAL` was masking. Its generated tsconfig sets `rootDir` to `src/mitm` while those sources import from `src/lib` (`TS6059`), and the codebase uses extensionless relative imports, which `moduleResolution: NodeNext` rejects (`TS2835`):

```
src/lib/db/agentBridgeState.ts(6,31): error TS6059: File '.../src/lib/db/core.ts' is not under 'rootDir' '.../src/mitm'
src/lib/db/core.ts(7,36): error TS2835: Relative import paths need explicit file extensions ...
```

The same failure occurs on Linux, where the spawn already worked — so that step has presumably never produced output. It is left untouched here because the package ships the `src/mitm` TypeScript sources and the CLI registers `tsx/esm` at runtime, so its compiled output is not consumed. Happy to open a separate issue for it if you'd like it addressed.

## Checklist notes

- Lint: `scripts/build/` is excluded by `eslint.config.mjs`, so the file reports `File ignored because of a matching ignore pattern`. The pre-commit hooks (prettier, eslint, `check:docs-sync`, `check:any-budget:t11`, tracked-artifacts) all pass.
- No new public API, no secrets, no new inputs to validate.
- Changelog fragment added in a follow-up commit once the PR number is known, per `changelog.d/README.md`.


diegosouzapw:
Thanks for this — the diagnosis is right and the failure mode you describe is real and worth fixing. But the preferred path regresses the same class of failure onto Linux, which is where CI and the npm publish run. Please hold before merging.

## What I measured

`esbuild`'s `bin` entry is **not** a JS script — it is a native ELF binary:

```
$ node -e "console.log(require('./node_modules/esbuild/package.json').bin)"
{ esbuild: 'bin/esbuild' }

$ head -c 20 node_modules/esbuild/bin/esbuild | cat -v
^?ELF^B^A^A^@^@^@^@^@^@^@^@^@^B^@>^@
```

So `execFileSync(process.execPath, [localEntry, ...args])` hands an ELF file to Node:

```
node_modules/esbuild/bin/esbuild:1
ELF          >    `H     @       X          @ 8  @
^
SyntaxError: Invalid or unexpected token
```

I replicated `runBuildTool()` verbatim with the exact MCP-server flags from `prepublish.ts`:

```
  → caminho PREFERIDO: node .../node_modules/esbuild/bin/esbuild
  ⛔ FALHOU
--- artefato gerado? ---
  NENHUM
```

Against the current release tip, unchanged, the same bundle succeeds:

```
$ npx esbuild open-sse/mcp-server/server.ts --bundle --platform=node --packages=external --format=esm --outfile=/tmp/out.js
  /tmp/out.js  3.9mb
⚡ Done in 566ms
```

`typescript` is fine — `bin/tsc` really is a JS shim (`#!/usr/bin/env node` + `require('../lib/tsc.js')`). The problem is specific to packages whose `bin` is a compiled binary, and `esbuild` is used by `runBuildTool` in **three** places: the MCP server bundle (L297), the LLMLingua ONNX worker (L343) and the CLI entrypoint (L371).

Because each caller catches into `console.warn`, the build would still exit 0 with three bundles missing — the exact symptom your PR set out to remove, moved from Windows to Linux. Two of the three artifacts are in `PACK_ARTIFACT_REQUIRED_PATHS`, so `check:pack-artifact` would fail the release job rather than shipping silently, but `dist/open-sse/mcp-server/server.js` is not covered and would go missing quietly.

(This likely passed your Windows verification because on win32 the native binary lives in `@esbuild/win32-x64/esbuild.exe` and `esbuild/bin/esbuild` is a JS wrapper — so the preferred path is genuinely correct there.)

## Suggestion

Detect whether the resolved entry is actually a JS script before handing it to Node, and execute it directly otherwise:

```ts
function isNodeScript(entryPath: string): boolean {
  const fd = openSync(entryPath, "r");
  try {
    const buf = Buffer.alloc(2);
    readSync(fd, buf, 0, 2, 0);
    return buf.toString("latin1") === "#!";   // ELF starts with \x7FELF, PE with MZ
  } finally {
    closeSync(fd);
  }
}
```

Then in `runBuildTool`: if `isNodeScript(localEntry)` run it via `process.execPath` as you do today; otherwise `execFileSync(localEntry, args, options)` directly — no shim, no shell, and it works for both a native binary and a shebang script.

Worth adding a unit test that asserts the branch chosen for a native-binary `bin` — that is what would have caught this without a Linux run.

Also heads-up: your branch is 124 commits behind `release/v3.8.49`, which is why the CI board is mostly red — most of those are base staleness, not your change. Rebasing on the current tip will clear them.

diegosouzapw:
Marking this **[v3.8.50]** so it gets a proper look at the start of the next cycle rather than being rushed at release close.

To be clear about the reason: the bug you found is real and worth fixing — the four silently-skipped sub-steps with a build that still exits 0 is a genuine footgun. What holds the merge is the regression detailed [above](https://github.com/diegosouzapw/OmniRoute/pull/8858#issuecomment-5105998099): `esbuild`'s `bin` entry is a native binary, so the preferred path hands an ELF file to Node and the same silent-skip reappears on Linux, where CI and the npm publish run.

Nothing to redo from scratch — the shebang check is a small addition to `runBuildTool`, and a unit test pinning which branch a native-binary `bin` takes would make it airtight. Rebasing on the current `release/v3.8.49` tip will also clear most of the red CI (the branch is well behind).

No rush on your side; this is queued for the next cycle.

diegosouzapw:

Issue #8707, opened 2026-07-26
feat(dependencies): establish a first-class native SQLite baseline (better-sqlite3 lock + bun:sqlite parity)


epsilonode:
## First-Class Native SQLite Baseline

OmniRoute currently relies on `better-sqlite3` as its preferred SQLite implementation, while
`node:sqlite` is an incomplete fallback. This proposal establishes `node:sqlite` as a first-class native
SQLite baseline for the supported OmniRoute adapter surface, while preserving `better-sqlite3` preference
during the staged rollout.

A complete native baseline reduces functional dependence on an external native addon and its ABI,
module-resolution, unpacking, rebuild, and runtime-repair boundaries. It is intended to improve
cross-platform native SQLite resilience, simplify Electron delivery, and reliably remediate the concrete
fallback, startup, read-only, backup, transaction, vector, and persisted-database failure modes mapped
below.

This is not a speculative driver replacement. Each capability is introduced in a narrow PR, tied to a
concrete claim, and accepted only through the specified source or packaged-target evidence.

## Verified Baseline

| Area | Current evidence | Remaining proof |
| --- | --- | --- |
| Node native SQLite | Node 24.18.0, SQLite 3.53.1, modules ABI 137, `DatabaseSync`, `backup`, transaction, and extension APIs | Real forced-fallback adapter behavior |
| Electron native SQLite | Electron 43.1.1 embeds Node 24.18.0, SQLite 3.53.1, modules ABI 148, and the required API surface | Packaged Windows target evidence |
| Open semantics | `tryOpenSync()` currently constructs `DatabaseSync(filePath)` and drops caller options | Read-only and existing-only regression proof through the real Node branch |
| Vector capability | `sqlite-vec` resolves a platform-specific native artifact | Trusted enable/load/disable lifecycle and packaged restart proof |
| Packaging | #6794 and #7353 are merged package baselines | Two-cycle packaged persisted-database smoke; current build resource failure must be reproduced under adequate resources before being classified as a package defect |

## Proposed Delivery Stack

| PR | Scope and deliverables | Capability / claim | Required validation |
| --- | --- | --- | --- |
[...]

Issue #7784, opened 2026-07-19
fix(api): prevent silent lost updates from concurrent dashboard and API settings writers


RaviTharuma:
## Summary

Concurrent dashboard tabs, automation, or API clients can silently overwrite one another when they update map- or array-valued settings such as `providerStrategies` and the `wildcardAliases` array.

Both writes can return success, but the later writer may replace a newer value with a stale snapshot. There is no settings revision, ETag, compare-and-swap condition, or conflict response that tells either operator an update was lost.

This is a configuration-integrity problem rather than an HTTP cache problem: `Cache-Control: no-store` ensures fresh reads when they occur, but cannot protect the interval between a read and a later write.

---

## Why this matters

OmniRoute is often managed simultaneously by:

- an open dashboard tab;
- another browser or administrator;
- automation using `/api/settings`;
- deployment/bootstrap scripts;
- background settings writers.

A successful settings response should not mean “your change was accepted, but another unrelated change may have disappeared.” Silent lost updates are particularly risky for routing, authentication, resilience, and provider policy.

In one real deployment, an intentional Codex routing change was later replaced by a dashboard-originated settings write. The audit trail showed the final replacement, but there was no conflict warning or mechanism to preserve both valid changes.

---

## Current implementation

Verified against upstream v3.8.49, commit `a856e3dd207bfc5ac0435af650d57f53a4ed1063`.

### API writes are unconditional

`src/app/api/settings/route.ts` accepts a PATCH, records a before-snapshot for audit purposes, and calls `updateSettings(body)`. The request and response carry no settings revision or ETag and support no `If-Match` precondition. `PUT /api/settings` aliases the same unconditional PATCH implementation.

`src/lib/db/settings.ts::updateSettings()` persists every supplied top-level setting with:

```sql
INSERT OR REPLACE INTO key_value (namespace, key, value)
VALUES ('settings', ?, ?)
```

Top-level PATCHes are key-granular, but the value of a key such as `providerStrategies` is one complete JSON object. Replacing that value is therefore last-writer-wins.

### Dashboard components perform read → merge → full replacement

`ProviderAccountRoutingCard.tsx`:

1. fetches `/api/settings`;
2. reads the complete `providerStrategies` object;
3. changes one provider locally;
4. PATCHes the entire object back.

This narrows the race window but does not eliminate it. Two clients can both read revision N before either write completes.

[...]

Issue #7592, opened 2026-07-17
fix(startup): better-sqlite3 not unpacked from app.asar → native module load fails → sql.js fallback OOM crash


browseros247-dev:
### OmniRoute Version

3.8.48

### Installation Method

Electron desktop app

### Operating System

Windows

### OS Version

win11

### Node.js Version

v24.18.0 (system) — app bundles Electron 43.1.0 / Chrome 150.0.7871.47, which likely uses a different internal Node ABI

### Provider(s) Involved

N/A - this is a core database/storage layer issue, not provider-specific

### Model(s) Involved

N/A

### Client Tool

OmniRoute Desktop App (Electron)

### Description

better-sqlite3's native binary is packaged inside app.asar instead of being unpacked to app.asar.unpacked. Electron cannot dlopen() native .node binaries from within an asar archive, so the module fails to load at runtime. This is silently caught and the app falls back to sql.js (WASM). Shortly after, the DB connection closes unexpectedly and the app enters an unbounded, backoff-free retry loop, leaking a new sql.js WASM instance on each attempt, which exhausts system memory and crashes the app.

Confirmed via:
Test-Path ".../resources/app.asar" → True
Test-Path ".../resources/app.asar.unpacked/node_modules/better-sqlite3/build/Release/better_sqlite3.node" → False

The better-sqlite3 native module is missing from app.asar.unpacked entirely, confirming it was never unpacked at build time.

### Related Issues

I searched the existing issues for `better-sqlite3` before filing this — none appear to be an exact duplicate, but these are related and may share a common root cause in how the native binary is built/packaged for different installation methods:

| Issue | Platform | Summary | Status |
|---|---|---|---|
| [#3476](https://github.com/diegosouzapw/OmniRoute/issues/3476) | macOS, global npm install | better-sqlite3 ABI mismatch after Node runtime upgrade | Open |
| [#3295](https://github.com/diegosouzapw/OmniRoute/issues/3295) | Global npm install | Auto-update's `npm rebuild better-sqlite3` step fails silently, no rollback | Open |
| [#2830](https://github.com/diegosouzapw/OmniRoute/issues/2830) | Docker / Linux | `Could not locate the bindings file` on 3.8.4+ Docker images | Open |
| [#1648](https://github.com/diegosouzapw/OmniRoute/issues/1648) | Docker, Windows | Generic SQLite error, `no such table: batches` | Closed |

[...]

PR #7353, opened 2026-07-16
fix(electron): normalize hashed standalone externals

tianrking:
## Summary

Fixes the Electron packaging path that left Turbopack-hashed external imports such as `ws-a972e7ffa40ff725` in server instrumentation chunks.

`assembleStandalone` already has `patchTurbopackChunks` for this exact normalization and the npm prepublish flow enables it. The Electron staging flow did not, so the packaged app copied canonical `node_modules/ws` but attempted to import the non-existent hashed name at startup.

This enables the existing normalization step for Electron builds and adds regression coverage for both the package-name rewrite and the Electron assembly wiring.

## Validation

- `node --import tsx/esm --test --test-name-pattern "patchTurbopackChunks restores canonical external package names|electron standalone assembly normalizes Turbopack hashed external imports" tests/unit/build/assemble-standalone.test.ts tests/unit/electron-packaging.test.ts`
- `npx prettier --check scripts/build/prepare-electron-standalone.mjs tests/unit/build/assemble-standalone.test.ts tests/unit/electron-packaging.test.ts`
- `npx eslint tests/unit/build/assemble-standalone.test.ts tests/unit/electron-packaging.test.ts --no-error-on-unmatched-pattern --suppressions-location config/quality/eslint-suppressions.json`
- `git diff --check`

The complete `assemble-standalone.test.ts` suite has an existing Windows-only path failure in its unrelated `standalone-server-ws.mjs` test: it derives a `file:` URL with `.pathname`, producing `F:\\F:\\...`. The selected new regression and Electron packaging test both pass.

## Scope

No provider behavior, route handling, or bundled dependency list changes. The staging step now rewrites only the erroneous hash suffixes before `electron-builder` copies the standalone bundle.

PR #8871, opened 2026-07-28
feat(db): add node sqlite adapter parity

epsilonode:
## PR 2 of issue #8707 (staged native SQLite baseline)

PR #8724 established safe `node:sqlite` fallback open semantics while retaining `better-sqlite3` preference. This follow-up adds backup and transaction parity when the real `node:sqlite` fallback is active.

### Native backup

The Node adapter now uses Node's native SQLite backup API when available, retaining the existing checkpoint-and-copy fallback for older Node runtimes.

Value:
- Produces a SQLite-consistent backup destination.
- Preserves existing backup scheduling, throttling, retention, manual/pre-restore, and auto-backup opt-out policy.
- Does not change `better-sqlite3` backup behavior or driver preference.

Issue context:
- #5406 establishes Windows WAL, close, and handle-release constraints around import/restore. This PR does not claim to fix import locking or replace its retry policy.
- #5871 establishes backup scheduling and opt-out policy as correctness requirements. This PR changes only the `node:sqlite` backup mechanism.

### Immediate and nested transactions

The Node adapter previously implemented `immediate()` as a savepoint, which does not acquire SQLite's write reservation at transaction start. It now uses `BEGIN IMMEDIATE` with explicit commit/rollback behavior. Transactions nested inside an immediate scope retain savepoint isolation.

Value:
- Competing writers follow SQLite's bounded busy-timeout behavior rather than proceeding as deferred transactions.
- A thrown callback leaves no partial writes.
- An inner failure rolls back to its savepoint while intended outer writes may continue and commit.

Issue context:
- #7784 demonstrates why reliable immediate acquisition and nested atomicity matter for concurrent settings mutations. This PR supplies adapter correctness only; it does not add revisions, ETags, conflict responses, or settings API changes.

## Validation

Ran the focused Node adapter test files through the repository-pinned Node 24 runtime:

- `tests/unit/db-adapters/nodeSqliteShared.test.ts`
- `tests/unit/db-adapters/driverFactory.test.ts`

Result: 24 tests passed. The suite verifies native `node:sqlite` backup destination integrity from a WAL-backed source; immediate transaction commit and rollback final state; nested savepoint rollback while the outer transaction continues; and bounded contention between two real connections. It also asserts the generated immediate and savepoint control-SQL ordering deterministically.

Also ran repository lint, core TypeScript checking, changelog-integrity, `check:any-budget:t11`, and `check:tracked-artifacts` successfully.

The full unit shards, coverage gate, production build, Vitest, and SonarQube remain CI evidence rather than local claims.

## Boundaries

- `better-sqlite3` remains preferred.
- No backup policy, retention, import/restore, or handle-release redesign.
- No generic lock-retry strategy.
- No settings ETag/revision work.
- No vector loading, dependency churn, Electron packaging, or driver-policy change.

## Follow-up

This is PR 2 of #8707. After it merges, PR 3 will add only the trusted, package-resolved `sqlite-vec` enable/load/disable lifecycle and a two-cycle Windows Electron package smoke against one persisted data directory.

PR 3 will provide target-specific evidence for the package/startup context in #7132, #7592, #7346, and #6796. It will validate the merged packaging baseline from #6794 and #7353 rather than duplicate it, and will not claim to close those issues without package-specific reproduction and maintainer confirmation.

### DIFF
--- a/changelog.d/features/8870-node-sqlite-adapter-parity.md
+++ b/changelog.d/features/8870-node-sqlite-adapter-parity.md
@@ -1,0 +1,1 @@
+- **Database**: The `node:sqlite` fallback now uses SQLite's native backup API and real immediate write transactions, improving backup consistency and concurrent-write behavior when `better-sqlite3` is unavailable
--- a/src/lib/db/adapters/nodeSqliteShared.ts
+++ b/src/lib/db/adapters/nodeSqliteShared.ts
@@ -19,6 +19,7 @@ export function createNodeSqliteAdapterFromDatabase(
   onClose?: () => void
 ): SqliteAdapter {
   let _isOpen = true;
+  let transactionDepth = 0;
   type NodeSqliteStatement = ReturnType<NodeSqliteDatabaseLike["prepare"]>;
   interface CachedStatement {
     stmt: NodeSqliteStatement;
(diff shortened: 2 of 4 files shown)


diegosouzapw:

PR #8886, opened 2026-07-29
fix(antigravity): add onboardUser fallback for accounts missing Cloud Code project

HouMinXi:
Google One AI accounts added via Dashboard OAuth (not agy CLI login) may lack a Cloud Code project because the initial OAuth exchange's onboardUser call failed silently. loadCodeAssist is read-only and cannot create projects, so subsequent discovery calls return empty forever.

This adds a tryOnboardUser fallback in ensureAntigravityProjectAssigned: when loadCodeAssist returns no projectId, call onboardUser to create the project, then retry loadCodeAssist.

Changes:
- open-sse/config/antigravityUpstream.ts — export onboardUser endpoint URLs
- open-sse/services/antigravityProjectBootstrap.ts — add onboardUser fallback with per-key locking and LRU caches
- tests/unit/antigravity-discovery-bootstrap.test.ts — 4 new tests for onboardUser fallback scenarios

Note: accounts with Google Country Association set to a restricted region (e.g. China) will still be blocked by Google's eligibility check at the API level. The region must be changed first at https://policies.google.com/country-association-form.

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -173,6 +173,10 @@ const nextConfig = {
     serverActions: {
       bodySizeLimit: process.env.OMNIROUTE_SERVER_ACTIONS_BODY_LIMIT || "50mb",
     },
+    // Reduce peak heap during production builds (Next.js 15+).
+    webpackMemoryOptimizations: true,
+    // Run webpack in a separate Node worker, lowering main-process memory.
+    webpackBuildWorker: true,
     // Next.js proxy (middleware) has a default 10MB body clone limit. File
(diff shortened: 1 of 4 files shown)


HouMinXi:
Updated: fixed TypeScript error (non-null assertion on Map.get after has check) that caused Fast Quality Gates failure.

Remaining CI failures are upstream pre-existing:
- Unit Tests (3/4): test expects 2 fetchAvailableModels endpoints but code only has 1 — not changed by this PR
- Merge integrity: missing generated SKILL.md files — not related to this PR
- Build (advisory): runner cancelled externally — not a code failure

PR #8905, opened 2026-07-29
fix(combos): include id column in getCombos query

HouMinXi:
getCombos() SELECT was missing the id column, so returned combo objects had their id come only from the JSON data blob. If the data blob lacked an id field, callers (including the Dashboard) saw null — making the combo appear to have no primary key and impossible to delete.

Add id to the SELECT so the database column value is always available.

Reproduction: create a combo via Dashboard, then check the returned object — before this fix, the id field was missing from the SQL result and fell through to the JSON blob.


diegosouzapw:

PR #8933, opened 2026-07-29
Fix custom tool output pairing during context compression

### LINKED ISSUE #8932: fix: context compression can orphan custom tool calls in Responses history
## Summary

For large native Responses API histories, proactive context compression can retain a `custom_tool_call` while dropping its matching `custom_tool_call_output`. The Codex upstream then rejects the request before inference with HTTP 400:

```text
No tool output found for custom tool call call_...
```

## Reproduction

1. Send a native `/v1/responses` conversation containing matched `custom_tool_call` and `custom_tool_call_output` items.
2. Grow the history until proactive context compression reaches the `purify_history` layer.
3. Continue the conversation.
4. Observe that the compressed request can contain the custom call without its output and that every Codex account rejects the same request with HTTP 400.

This was reproduced with 134 input items. Compression reduced the estimated context from 193,687 to 132,408 tokens using `trim_tools`, `compress_thinking`, and `purify_history`. The original client session contained both custom-tool items consecutively with the same `call_id`; the failure appeared only after compression.

## Root cause

`open-sse/services/contextManager.ts` repairs OpenAI Chat and Anthropic tool pairs, but does not understand native Responses `custom_tool_call` / `custom_tool_call_output` items.

`open-sse/executors/codex.ts::repairMissingCodexFunctionCallOutputs()` provides a final safeguard for `function_call`, but ignores `custom_tool_call`.

## Expected behavior

Context compression must preserve tool exchanges atomically for native Responses items. As a defensive final safeguard, the Codex executor should also repair a retained `custom_tool_call` whose output is missing, using the matching `custom_tool_call_output` type.

## Proposed validation

Add regression coverage for both layers:

- context purification never leaves orphaned native custom calls or outputs;
- Codex request transformation inserts a matching empty `custom_tool_call_output` when a custom call reaches it without an output.

Target branch: `release/v3.8.50`.

JxnLexn:
## Summary

- keep native Responses custom-tool call/output pairs consistent when Layer-3 context compaction removes mapped outputs
- add a Codex executor safeguard that inserts the matching custom output type when a custom call reaches upstream without an output
- bypass OmniRoute's catalog-only context overflow preflight for native Responses requests targeting an all-Codex combo
- bypass OmniRoute prompt/reactive/last-resort compaction for native Codex Responses requests so Codex CLI/Desktop/App remains the context owner
- retain the concrete Codex context-budget check, including `CONTEXT_LENGTH_CODEX`

## Root cause

Two independent proxy-side guards affected native Codex Responses requests. The Responses compression adapter did not apply orphan cleanup to `custom_tool_call`, and the Codex executor fallback repaired only `function_call_output`. Separately, combo preflight rejected a 275k-token request against the stale 272k catalog hint before the target handler could apply `CONTEXT_LENGTH_CODEX=400000`. OmniRoute's own compaction could also rewrite native Codex history before dispatch.

## User impact

Long Codex Desktop / Responses conversations can continue after automatic context compaction instead of failing every account in a combo with `No tool output found for custom tool call ...` or a misleading `largest known context limit ... 272000` error.

## Validation

- `npm run typecheck:core`
- `node --import tsx/esm --test tests/unit/combo-context-window-filter.test.ts tests/unit/combo-target-resolution-split.test.ts tests/unit/context-manager.test.ts tests/unit/compression/body-adapter.test.ts tests/unit/executor-codex.test.ts` (101 passed)
- commit hooks: Prettier, ESLint, docs sync, explicit-any budget, tracked-artifact check
- `git diff --check`

Fixes #8932


lumamax:
Independent downstream validation on exact commit `a725cae01b1f31ae97ef070ba7d35c8b01d02f8a`:

- Node `22.22.3`
- `npm run typecheck:core`: pass
- the five targeted suites listed in the PR plus combo target resolution: 102/102 tests pass
- `npm run build`: pass
- `git diff --check`: pass

Our 3.8.49 production reproduction matches #8932 precisely: the incoming native Responses history contains both `custom_tool_call` and the matching `custom_tool_call_output`; reactive compression drops the mapped output, but orphan cleanup only recognizes `function_call`, so the remaining custom call is rejected upstream with `400 No tool output found for custom tool call`. Raising the supported `CONTEXT_LENGTH_CODEX` override contains the incident by keeping proxy compaction beyond the client boundary, but this PR is the correct durable direction because it preserves custom pairs and restores the native Codex client as context owner.

I also reviewed the current red checks. The file-size gate is directly actionable here (`chat.ts`, `codex.ts`, and `executor-codex.test.ts` exceed frozen budgets). The provider-count assertions and provider translation golden failure do not touch this diff, while the advisory build job timed out; the exact commit builds successfully locally. Splitting the new helpers/test into smaller files should clear the patch-owned gate without changing behavior.

Issue #6409, opened 2026-07-06
fix(nodejs): npm run build requires >14 GB RAM


powellnorma:
### OmniRoute Version

3.8.46

### Installation Method

Built from source

### Operating System

Linux

### Description

Is that expected..? I think it's a bit excessive.. 

### Steps to Reproduce

npm run build

### Expected Behavior

1-2 GB ram usage

### Actual Behavior

> 14 GB .. (I actually aborted it, because I did not want my computer to crash.. So maybe it would need even more)

### Test Impact

Unsure

### Error Logs / Output

```shell

```

diegosouzapw:
Thanks for reporting! Fixed in `release/v3.8.47` via #6885 (already merged into the active development branch — feel free to pull and test it now). It will ship in the next release (v3.8.47).

PR #8957, opened 2026-07-30
fix(ci): stop one failing platform from taking the whole desktop channel down

diegosouzapw:
v3.8.49 shipped with ZERO release assets. v3.8.48 had 16. Every gate was green.

Measured from run 30503231362, not inferred:

    Build Electron (windows) ....... success
    Build Electron (macos-intel) ... success
    Build Electron (macos-arm64) ... success
    Build Electron (linux) ......... failure
    Create Release ................. skipped
    Publish to npm ................. skipped

The linux leg died 8 min into "Creating an optimized production build" with
"The runner has received a shutdown signal" and no exit code, on runner
`GitHub Actions 1000378558` — github-hosted, so this was the VM being
reclaimed, not a Node heap error. Reproduced on a 32 GB machine from the exact
tag commit: the same build succeeds and peaks past 18 GB.

Three independent defects compounded, one fix each:

1. Turbopack allocates natively (Rust, off the V8 heap), so the existing
   --max_old_space_size=6144 does not bound it. The project already documents
   the webpack fallback as the escape hatch for RAM-constrained machines
   (docs/reference/ENVIRONMENT.md, #6409), and nightly-compat already applies
   it for the same reason on Node 26 (#8090). The linux leg now selects it;
   Windows/macOS keep Turbopack since they build fine and it is faster.

2. `release` gated on `needs: [validate, build]` with no `if:`, so ONE failing
   leg skipped it and discarded the three artifacts that DID build — 1.7 GB,
   still retained — plus the source archives, which depend on no build at all.
   Fail-closed made a partial failure look total. It is now fail-partial:
   attach what exists, still requiring `validate` to have passed.

3. Nothing asserted the release HAS assets, so 16 binaries and 0 binaries were
   indistinguishable to CI. New `verify-desktop-assets` job asserts one asset
   per platform and fails loudly.

The check is a separate job on purpose: failing inside `release` would cascade
into `publish-npm` (`needs: [validate, release]`) and block the npm channel
over a desktop-only gap. Now the assets attach, npm still publishes, and an
incomplete desktop channel is a red job instead of silence.

TDD: 3 of the 4 new assertions fail on the tip of release/v3.8.50 and pass
with this change. The 4th asserts a negative (publish-npm must not gain a
dependency on the asset check) and guards against future regression rather
than reproducing the bug.

    node --import tsx/esm --test tests/unit/electron-release-desktop-channel-8949.test.ts
    # 4 pass, 0 fail        (base: 1 pass, 3 fail)
    check:workflows --ratchet → 189 findings, baseline 190, no regression

### DIFF
--- a/.github/workflows/electron-release.yml
+++ b/.github/workflows/electron-release.yml
@@ -120,6 +120,18 @@ jobs:
         env:
           JWT_SECRET: ci-build-secret-with-sufficient-length-for-validation
           NODE_OPTIONS: "--max_old_space_size=6144"
+          # Linux builds with webpack, not Turbopack. Turbopack's production build
+          # allocates natively (Rust, off the V8 heap), so --max_old_space_size does
+          # not bound it, and on this module graph it peaks above what the hosted
+          # runner can give — the VM is reclaimed mid-compile with "The runner has
+          # received a shutdown signal", no exit code. That is what silently took the
+          # whole desktop channel out of v3.8.49: the linux leg died, `release` was
+          # skipped, and the release shipped with ZERO assets. Measured on a 32 GB
+          # box the same build passes and peaks past 14 GB. The webpack fallback is
+          # the project's documented escape hatch for RAM-constrained machines
+          # (docs/reference/ENVIRONMENT.md, #6409) and is the same remedy already
+          # applied to nightly-compat's Node 26 build (#8090).
+          OMNIROUTE_USE_TURBOPACK: ${{ matrix.platform == 'linux' && '0' || '1' }}
         run: npm run build
 
       - name: Sync version in electron/package.json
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Escopo — o que esta PR **não** cobre (o nome da branch promete demais):

O `latest-mac.yml` **continua errado**, e é um bug de produção vivo. Os dois jobs de macOS geram cada um o seu manifesto contendo **apenas o seu próprio** dmg (medido: 338 bytes no Intel, 350 no arm64, conteúdos diferentes), e o `merge-multiple: true` do `download-artifact` resolve a colisão de nome **por ordem de chegada** — um sobrescreve o outro sem aviso. Na v3.8.48 publicada o vencedor foi o arm64.

Consequência, medida no fonte do `electron-updater` (`out/providers/Provider.js:80`):

```js
const result = filteredFiles.find(it => [it.url.pathname, it.info.url].some(n => n.includes(process.arch))) ?? filteredFiles.shift();
```

O dmg Intel se chama `OmniRoute-X.Y.Z.dmg`, **sem sufixo de arquitetura**. Num Mac Intel (`process.arch === "x64"`) a busca por substring não casa em nenhum arquivo, o updater cai no `shift()` e pega o **primeiro** da lista — que hoje é o arm64. Ou seja: **Mac Intel baixa o dmg ARM**.

Fora desta PR de propósito, porque a correção certa muda a topologia do download (`merge-multiple: false` + merge deliberado, ou construir os dois arcos de macOS num só job como o leg linux já faz) e merece um script testável + PR próprio. Misturar isso aqui tornaria o diff difícil de revisar e atrasaria as três correções que já estão prontas.

Issue #7034, opened 2026-07-13
feat(auth): accept x-goog-api-key header so gemini-cli can use OmniRoute as a gateway


QRcode1337:
## Summary

OmniRoute already exposes the native Gemini API surface (`/v1beta/models/*:generateContent`, `:streamGenerateContent`), but the auth middleware only extracts client keys from `Authorization: Bearer` / `x-api-key`. Google's **gemini-cli** sends its key exclusively via the `x-goog-api-key` header (not configurable on their side), so every request 401s before reaching the route handler.

## Repro (v3.8.46)

```bash
# Works — same OmniRoute API key, Bearer header:
curl -H "Authorization: Bearer sk-..." http://localhost:20128/v1beta/models   # 200

# Fails — how gemini-cli actually sends it:
curl -H "x-goog-api-key: sk-..." http://localhost:20128/v1beta/models        # 401 AUTH_002
```

Captured from gemini-cli (v0.45) pointed at a mock server via `GOOGLE_GEMINI_BASE_URL`:

```
POST /v1beta/models/gemini-3.1-pro-preview:streamGenerateContent
x-goog-api-key: <key>     ← only auth header sent; no Authorization header
```

## Proposed fix

In the auth middleware key-extraction chain, add `x-goog-api-key` as a fallback:

```
authorization: Bearer → x-api-key → x-goog-api-key
```

One-line change; would make OmniRoute a drop-in `GOOGLE_GEMINI_BASE_URL` target for gemini-cli, same as it already is for Claude Code (`ANTHROPIC_BASE_URL`) and Codex.

Possibly related: #572 (closed needs-info) looks like the same root cause.

diegosouzapw:
Hi @QRcode1337! Thanks for the crisp repro — you're exactly right about the root cause. 🙏

We confirmed it: the real gate is our client-facing auth policy (`src/server/authz/policies/clientApi.ts`), which only recognizes `Authorization: Bearer` and a gated `x-api-key` today. We'll add `x-goog-api-key` as a third fallback so `gemini-cli` (and anything else using Google's SDK convention) works against our native `/v1beta` surface out of the box, same as Claude Code and Codex already do.

**Status:** ✅ Accepted — queued for implementation (small, low-risk change)

We'll tag you once it lands so you can confirm `gemini-cli` works end-to-end before it ships. Thanks for the contribution! 🚀

diegosouzapw:
[...]

PR #8985, opened 2026-07-30
fix(ci): close six release-process gaps from the v3.8.49 run

diegosouzapw:
Two release-tooling defects found during the v3.8.49 run (gaps 4 and 7 of the
process dossier). Both fail by hiding work rather than announcing themselves,
which is why each one had already cost a real mistake.

## The reconciliation range was 62× too wide

`list-uncovered-commits.mjs` bounded its scan with `git describe --tags`. Releases
reach `main` by SQUASH, so no commit on a release branch is ever an ancestor of the
tag, and `vPREV..HEAD` re-lists the un-squashed history of every earlier cycle.
Measured on release/v3.8.50 @ 7eca04fd12:

    v3.8.49..HEAD ....... 1361 commits
    cycle open..HEAD .....   22 commits

The report drowns in noise, and that is how a previous reconciliation let ~200 PRs
through with no CHANGELOG bullet.

The base is now resolved by CONTENT — the oldest commit that introduced this
version string into package.json — deliberately NOT by commit subject, because the
subject has already changed format once:

    chore(release): bump v3.8.49 (development cycle version)     older
    chore(release): open v3.8.50 development cycle               current

A message-matching resolver would have silently reverted to the broken tag base the
first time someone reworded the bump. The fallback now writes a WARNING to stderr
explaining that a tag range re-lists previous cycles, so a shallow clone degrades
loudly instead of quietly reproducing the bug. One of the five tests asserts
exactly that the warning says "squash" and "noise".

## The back-merge resurrects fragments that already shipped

The release lands on `main` as one squash commit, so `main` still carries every
`changelog.d/` fragment the reconciliation folded in and deleted. Back-merging
`main` restores all of them — 191 in the v3.8.49 run. Nothing breaks at that
instant; the next aggregation folds them in a SECOND time and the section grows
duplicates that have to be hand-unpicked.

New `scripts/release/sweep-stale-fragments.mjs` (`npm run sweep:stale-fragments`)
reports them, and `--apply` removes them. Report mode exits 1 so the back-merge
step can gate on it.

The identity rule took two attempts, and the second one exists because running the
script against the live repo refuted the first. Matching on any `#N` in the bullet
flagged `changelog.d/features/8980-deprecate-gemini-cli-provider.md` as stale,
because that bullet cites issue **#7034** for context and #7034 shipped in an
earlier cycle — it would have deleted an unreleased fragment and dropped its
credit. A bullet routinely cites issues it merely references; only the
`<PR-number>-<slug>.md` filename says which PR the fragment *is*. That case is now
a regression test.

Every ambiguous case resolves toward KEEPING: no number in the filename falls back
to normalized text, text shorter than 12 chars is never matched, and anything
matching neither is kept. A surviving duplicate is a nuisance someone notices; a
deleted fragment silently costs a contributor their credit.

    node --import tsx/esm --test tests/unit/release-cycle-base-resolver.test.ts   # 5 pass
    node --import tsx/esm --test tests/unit/sweep-stale-fragments.test.ts         # 11 pass
    node scripts/release/list-uncovered-commits.mjs --json
      → base ed2db6cb19, baseSource "cycle-open", total 22   (was 1361)
    node scripts/release/sweep-stale-fragments.mjs
      → 4 fragments, 0 stale, exit 0


diegosouzapw:
Escopo ampliado: além dos gaps **4** e **7** (tooling de release), este PR passou a fechar também os gaps **6, 9, 10 e 23** — os quatro gates que puniam a coisa errada. Um commit por grupo, e cada correção tem teste.

Detalhe que merece leitura: a minha própria descrição do **gap 6 estava errada**, e a medição derrubou. Eu havia registrado "1277 arquivos, >30 min, a base é a main". Medindo: **3977** arquivos rastreados, scan absoluto em **~1 s**, e o diff já usa `base...HEAD` com três pontos, que é merge-base. A base nunca foi o problema — e o caso **não é reproduzível hoje**, porque a `main` já recebeu o squash da v3.8.49 e o merge-base é recente. A patologia só existe **durante** a release, na janela antes do squash chegar na `main`. Mesma raiz topológica do gap 4.

Total: **6 gaps, 38 testes**.

PR #9018, opened 2026-07-30
chore(ci): adopt Ubuntu 26.04 runners

backryun:
## Summary

- pin GitHub-hosted Ubuntu jobs to `ubuntu-26.04`
- use the image-bundled Node 24 for Ubuntu-only Node 24 jobs and remove 38 redundant `setup-node` steps
- pin `actions/checkout` to `3d3c42e5aac5ba805825da76410c181273ba90b1` (`v7.0.1`) and retained `actions/setup-node` uses to `820762786026740c76f36085b0efc47a31fe5020` (`v7`)
- retain `setup-node` where it still selects Node 20/22/26, serves an OS matrix or self-hosted fallback, or configures npm publishing
- fold `Vitest (fast-path)` into `Fast Quality Gates`, reusing its checkout and dependency installation while preserving JUnit/Trunk reporting

## Why

The advisory production build repeatedly reaches `Creating an optimized production build` and is then cancelled on the hosted runner without a compiler error. This PR gives the Ubuntu 26.04 image a full CI trial while removing redundant Node setup from jobs that already require the image-bundled Node 24 runtime. The runner still has the same 16 GB memory class, so this is an evidence-gathering change rather than a claim that the build cancellation is already solved.

The same-run timing audit also showed that the standalone Vitest job spent about 80 seconds in `npm ci` for roughly 13 seconds of tests. Running Vitest at the end of `Fast Quality Gates` removes that duplicated runner setup; `if: !cancelled()` retains the test signal after an earlier fast-gate failure, and the advisory Trunk upload remains `always()` for own-origin PRs.

## Impact

Hosted Linux jobs run on the Ubuntu 26.04 public-preview image. Node compatibility matrices, cross-platform jobs, self-hosted fallbacks and npm registry configuration remain explicit. The separate `Vitest (fast-path)` check is replaced by a named `Vitest` step inside `Fast Quality Gates`, eliminating one runner job and approximately 80 seconds of redundant dependency installation per qualifying PR while adding only the Vitest runtime to the existing job. The workflow definitions shrink by 164 net lines.

## Validation

- parsed all 23 workflow YAML files
- verified 70 effective Ubuntu runner values use `ubuntu-26.04` and none use `ubuntu-latest`
- verified all checkout/setup-node references use the selected immutable SHAs
- `actionlint 1.7.12`: only the expected runner-label diagnostics for the official public-preview `ubuntu-26.04` label, which is not yet in actionlint’s built-in label list
- `zizmor 1.28.0`: 71 findings, below the frozen baseline of 190
- workflow-check unit tests: 36 passed
- Vitest: 34 files / 291 tests passed with the retained JUnit output path
- Prettier and `git diff --check` passed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ env:
 jobs:
   changes:
     name: Change Classification
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-26.04
     outputs:
       code: ${{ steps.classify.outputs.code }}
       docs: ${{ steps.classify.outputs.docs }}
@@ -35,13 +35,10 @@ jobs:
       workflow: ${{ steps.classify.outputs.workflow }}
       testsOnly: ${{ steps.classify.outputs.testsOnly }}
     steps:
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
           fetch-depth: 0
-      - uses: actions/setup-node@v7
-        with:
-          node-version: ${{ env.CI_NODE_VERSION }}
       # Refuse a PR that targets its own head branch before spending anything on it. #8912 has
       # head == base == release/v3.8.50: no diff, can never merge, and it sits in the queue with
       # a full check board attached on every push to that branch. One field comparison.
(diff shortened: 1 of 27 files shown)


backryun:
@diegosouzapw — a small CI-efficiency proposal is now included in commit `26b058207`.

From the latest release-PR run, the standalone **Vitest (fast-path)** job spent about **80s in `npm ci`** for roughly **13s of Vitest execution**. This commit moves the same Vitest command and JUnit/Trunk upload into **Fast Quality Gates**, which has already completed checkout and dependency installation.

Expected trade-off:
- removes one hosted runner job and about 80s of duplicated dependency setup per qualifying PR;
- adds only the Vitest runtime (~13s in the measured run) to Fast Quality Gates;
- keeps the full Vitest command and `trunk-junit/vitest-fastpath.xml` output unchanged;
- uses `if: !cancelled()` so Vitest still runs for diagnostic signal after an earlier fast-gate failure;
- keeps the Trunk upload advisory, `always()`, and restricted to own-origin PRs.

Local validation: all 23 workflow YAML files parsed, workflow-check unit tests passed (32/32), Vitest passed (34 files / 291 tests), and zizmor remains at 71 findings (below the 190 baseline).

Please let me know whether you prefer this consolidation in the current runner-image PR or as a separate follow-up.

Issue #9012, opened 2026-07-30
docs(providers): document chat_admission_busy and how to tune heavyweight chat concurrency


JasonLandbridge:
## Summary

OmniRoute v3.8.49 can return the following error when two structurally heavyweight chat requests overlap:

```text
Error: 503: {"message":"Chat admission capacity is temporarily unavailable. Retry shortly.","type":"server_error","code":"chat_admission_busy"}
```

The HTTP response is a retryable `503` with `Retry-After`, and the structured error code is:

```json
{
  "error": {
    "code": "chat_admission_busy"
  }
}
```

This behavior is intentional, but it is easy for users to interpret it as a provider failure or as a limit controlled by the dashboard's **Settings → Resilience → Request Queue → Concurrent Requests** field. That GUI setting is separate and does not change chat-body admission capacity.

## User-visible workaround / configuration

The heavyweight chat admission limit is controlled by an environment variable, not by the current GUI:

```env
OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT=2
```

Restart OmniRoute after changing the environment. In the observed deployment, increasing this setting from the default `1` to `2` stopped the `chat_admission_busy` failures for two concurrent coding-agent requests.

Docker Compose example:

```yaml
services:
  omniroute:
    environment:
      OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT: "2"
```

The default remains the safest value:

```env
OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT=1
```

Users should raise it gradually and monitor process/container memory. A value such as `8` permits up to eight heavyweight requests to enter memory-intensive parsing, compression, translation, and dispatch paths concurrently; it is not equivalent to ordinary lightweight request concurrency and may reintroduce V8 or cgroup OOM termination.

Related classification thresholds are also environment-only in v3.8.49:

```env
OMNIROUTE_CHAT_LARGE_BODY_BYTES=262144
OMNIROUTE_CHAT_HEAVY_MESSAGE_COUNT=200
OMNIROUTE_CHAT_HEAVY_TOOL_COUNT=64
OMNIROUTE_CHAT_HEAVY_ESTIMATED_TOKENS=32000
OMNIROUTE_CHAT_HARD_MAX_MESSAGES=800
OMNIROUTE_CHAT_HARD_MAX_BODY_BYTES=52428800
```

[...]

PR #9021, opened 2026-07-30
docs(troubleshooting): document the chat_admission_busy 503 and how to tune heavyweight chat concurrency

xiaoyaner0201:
Documentation-only entry for issue #9012: the chat completions endpoint can return a retryable `503` with `code: "chat_admission_busy"`, and until now the English troubleshooting guide had no entry for that exact code — so operators had no documented way to tell it apart from an upstream-provider failure or from the dashboard's ordinary request-concurrency setting.

Refs #9012

## What this adds

One new section in `docs/guides/TROUBLESHOOTING.md` (`### Chat requests fail with 503 / chat_admission_busy`), placed next to the existing provider concurrency material:

- **Symptoms** — the retryable `503`, the `Retry-After` values (2s on the byte-based path, 1s plus `reason: "structure_limit"` on the structure-based path), and the fact that it can fire while another heavyweight chat or long-running SSE response is still in flight.
- **The exact response bodies**, copied verbatim from `src/shared/middleware/chatBodyAdmission.ts` rather than paraphrased.
- **Structure-heavy classification thresholds** (added in review round 2) — at the defaults, a request is structurally heavy at ≥ `200` messages, ≥ `64` tools, or ≥ `32,000` estimated tokens, or when bounded structure estimation exhausts its bounds of `10,000` visited nodes or depth `12`. Verified against `CHAT_HEAVY_MESSAGE_COUNT` / `CHAT_HEAVY_TOOL_COUNT` / `CHAT_HEAVY_ESTIMATED_TOKENS` and `estimateStructureTokens()` in `src/shared/middleware/chatBodyAdmission.ts`.
- **Cause** — deliberate process-local load shedding: a heavyweight lease is reserved *before* the large request body is retained and parsed, and is held for the lifetime of an SSE response. Explicitly states that Settings → Resilience → Request Queue → Concurrent Requests is a *different* mechanism and does not control this guard.
- **Observability note** (added in review round 2) — current heavyweight lease occupancy is not surfaced in the dashboard. This is deliberately scoped to dashboard occupancy rather than a broader "no metrics anywhere" claim, so the sentence stays verifiable: `activeHeavy` has no reader outside `chatBodyAdmission.ts`.
- **Fix** — retry first, honouring `Retry-After` with backoff; only then cautiously raise `OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT` one step at a time, restarting and watching memory headroom, because each additional heavyweight request increases concurrent V8 heap use and OOM risk.
- A link to the authoritative environment-variable reference, plus a note that loosening the heavyweight *classification* thresholds is riskier than a cautious in-flight increase.

## Deliberate non-goals

- The documented default stays `1`. The doc does **not** claim `2` is universally safe — it tells operators to validate against their own traffic and memory limits.
- No runtime, environment-default, UI, schema, or dependency change. No translated-doc changes, no changelog change.
- Docs-only diff: 1 file, +53/-0 against the base.

## Verification

Run on the exact pushed commit `fb52c1ea7894e342fe946df3871416928163b0bf` (tree `e7dcf844e0cfdef7b30bcdc548f374431670a941`), based on `release/v3.8.50` @ `9b3efef806b291bd5833ba678b57c4e071b65255` (tree `796208321e5b84f87f63ce2715838f057296395a`):

| Gate | Result |
| --- | --- |
| `node --import tsx/esm --test tests/unit/chat-body-admission.test.ts` | 22 pass / 0 fail |
| `npm run check:docs-all` | pass |
| `npm run check:fabricated-docs` | pass — 134 docs, no fabricated refs |
| `npm run check:doc-links` | pass — 814 internal links |
| `npm run check:docs-symbols` | pass |
| `npm run check:env-doc-sync` | pass |
| `npm run check:api-docs-refs` | pass |
| `npm run check:changelog-integrity` | pass |
| `git diff --check` | clean |

Every user-visible string and numeric threshold in the new section (error message, `code`, `type`, `reason`, env var names, default values, estimator bounds) was asserted against `src/shared/middleware/chatBodyAdmission.ts` — no synthesized names or values.

**Known pre-existing Prettier drift, disclosed:** `npx prettier --check docs/guides/TROUBLESHOOTING.md` fails on this branch, but it fails *identically* on the untouched base commit `9b3efef80` — the same hunks at the same original offsets (Quick Fixes table column padding at ~L40, an `*does*`→`_does_` emphasis normalisation at ~L95, and a missing blank line before the Docker IPv6 list at ~L306). None of them fall inside the added section, which is Prettier-clean when checked in isolation. I left the base drift alone rather than mixing an unrelated reformat into a documentation PR — happy to fix it here or in a separate PR if you'd prefer.

**Changelog fragment:** not added. `changelog.d/README.md` requires the PR number as the filename prefix, and the most recent docs-only commit on this line (#8991) shipped without a fragment. Glad to add `changelog.d/maintenance/9021-chat-admission-troubleshooting.md` on request.

Opened as a **Draft** pending maintainer confirmation that the scope and placement are what you want for #9012.

### DIFF
--- a/docs/guides/TROUBLESHOOTING.md
+++ b/docs/guides/TROUBLESHOOTING.md
@@ -40,19 +40,19 @@ Common problems and solutions for OmniRoute.
 
 ## Quick Fixes
 
-| Problem                                                    | Solution                                                                                                                                                  |
-| ---------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| First login not working                                    | Set `INITIAL_PASSWORD` in `.env` (no hardcoded default)                                                                                                   |
-| Dashboard opens on wrong port                              | Set `PORT=20128` and `NEXT_PUBLIC_BASE_URL=http://localhost:20128`                                                                                        |
-| No logs written to disk                                    | Set `APP_LOG_TO_FILE=true` and verify call log capture is enabled                                                                                         |
-| EACCES: permission denied                                  | Set `DATA_DIR=/path/to/writable/dir` to override `~/.omniroute`                                                                                           |
-| Routing strategy not saving                                | Update to the latest v3.x release (Zod schema fix for settings persistence shipped in earlier versions)                                                   |
-| Login crash / blank page                                   | Check Node.js version — see [Node.js Compatibility](#nodejs-compatibility) below                                                                          |
-| `dlopen` / `slice is not valid mach-o file` (macOS)        | Run `cd $(npm root -g)/omniroute/app && npm rebuild better-sqlite3 && omniroute` — see [macOS native module rebuild](#macos-native-module-rebuild) below  |
-| Proxy "fetch failed"                                       | Ensure proxy config is set at the correct level — see [Proxy Issues](#proxy-issues) below                                                                 |
+| Problem                                             | Solution                                                                                                                                                 |
+| --------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| First login not working                             | Set `INITIAL_PASSWORD` in `.env` (no hardcoded default)                                                                                                  |
+| Dashboard opens on wrong port                       | Set `PORT=20128` and `NEXT_PUBLIC_BASE_URL=http://localhost:20128`                                                                                       |
+| No logs written to disk                             | Set `APP_LOG_TO_FILE=true` and verify call log capture is enabled                                                                                        |
+| EACCES: permission denied                           | Set `DATA_DIR=/path/to/writable/dir` to override `~/.omniroute`                                                                                          |
+| Routing strategy not saving                         | Update to the latest v3.x release (Zod schema fix for settings persistence shipped in earlier versions)                                                  |
+| Login crash / blank page                            | Check Node.js version — see [Node.js Compatibility](#nodejs-compatibility) below                                                                         |
+| `dlopen` / `slice is not valid mach-o file` (macOS) | Run `cd $(npm root -g)/omniroute/app && npm rebuild better-sqlite3 && omniroute` — see [macOS native module rebuild](#macos-native-module-rebuild) below |
+| Proxy "fetch failed"                                | Ensure proxy config is set at the correct level — see [Proxy Issues](#proxy-issues) below                                                                |
 | Docker `curl: (56) Recv failure: Connection reset by peer` | Your Docker port bind may be landing on IPv6. Use `-p 127.0.0.1:20128:20128` to force IPv4, or test with `curl -4`. See [Docker IPv6](#docker-ipv6) below |
-| Antivirus quarantines `README.md`                          | False positive — see [Antivirus false positives](#antivirus-false-positives) below                                                                        |
-| Kaspersky flags the Desktop app as a Trojan                | Behavioral false positive on the unsigned installer — see [Antivirus false positives](#antivirus-false-positives) below                                   |
+| Antivirus quarantines `README.md`                   | False positive — see [Antivirus false positives](#antivirus-false-positives) below                                                                       |
+| Kaspersky flags the Desktop app as a Trojan         | Behavioral false positive on the unsigned installer — see [Antivirus false positives](#antivirus-false-positives) below                                  |
 
 ---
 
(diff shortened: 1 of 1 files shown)


maxmad64bis:
Factual accuracy verified against `chatBodyAdmission.ts` on `release/v3.8.50`. Response bodies, `Retry-After` values, `code`, `type`, and `reason` all match the source. Good call linking to `ENVIRONMENT.md` instead of duplicating thresholds, and good call keeping the documented default at `1` without claiming `2` is universally safe.

Two suggestions before ready-to-merge :

**1. Explain what triggers the structure-based path**

`reason: "structure_limit"` appears in the Symptoms and the sample response, but the classification thresholds that produce that reason are never stated. An operator seeing `structure_limit` in logs has no way to tell if it was ≥ 200 messages, ≥ 64 tools, ≥ 32 K estimated tokens, or an exhausted token estimator (deeply nested JSON exceeding 10 K nodes / depth 12). One inline sentence would close the loop :

> A request is classified as structurally heavy when it contains ≥ 200 messages, ≥ 64 tools, or ≥ 32 K estimated tokens, or when the token estimator exhausts its bounds on a deeply nested payload. See the authoritative thresholds in the [environment-variable reference](../reference/ENVIRONMENT.md#4-security--authentication).

**2. Note the absence of dashboard/metric visibility**

The `activeHeavy` counter and admission rejections are not exposed in the dashboard or any diagnostics endpoint. An operator whose dashboard shows "0 requests in flight" while clients receive `chat_admission_busy` has no in-product way to confirm the slot is held by a long-running SSE stream. A single sentence in the Cause section would save a round of blind troubleshooting :

> There is currently no dashboard or metrics visibility on heavyweight lease occupancy.

Both are one-line additions that don't change scope. Mergeable after these.

xiaoyaner0201:
Thanks — both suggestions adopted in `fb52c1ea7894e342fe946df3871416928163b0bf` (tree `e7dcf844e0cfdef7b30bcdc548f374431670a941`, +4/-0 on top of the previously published head, still docs-only, 1 file, +53/-0 vs `release/v3.8.50`).

**1. Structure-heavy classification thresholds** — added after the structure-based response paragraph, with the estimator bounds included and each number verified against source rather than the review text:

> At the default thresholds, a request is structurally heavy when it has at least `200` messages, at least `64` tools, or at least `32,000` estimated tokens, or when bounded structure estimation exhausts its bounds of `10,000` visited nodes or depth `12`.

Sources: `CHAT_HEAVY_MESSAGE_COUNT` / `CHAT_HEAVY_TOOL_COUNT` / `CHAT_HEAVY_ESTIMATED_TOKENS` (`src/shared/middleware/chatBodyAdmission.ts:33-44`) and the `maxNodes = 10_000` / `depth >= 12` bounds in `estimateStructureTokens()` (same file, `:157-194`). I wrote "exhausts its bounds" rather than "on a deeply nested payload" because an exhausted estimate can also come from wide arrays/objects hitting the remaining-node budget, not only from depth. The link to `ENVIRONMENT.md` you suggested was already present in the section immediately below, so I didn't duplicate it.

**2. Occupancy visibility** — added one sentence in the Cause section:

> Current heavyweight lease occupancy is not surfaced in the dashboard.

I deliberately narrowed this to dashboard occupancy rather than the broader "not exposed in the dashboard or any diagnostics endpoint / no metrics visibility" phrasing. What I could actually verify is that `activeHeavy` has no reader anywhere outside `chatBodyAdmission.ts` (`grep -rn "activeHeavy" src/ open-sse/` returns only the declaring file), which supports the dashboard-occupancy claim precisely. A wider "no metrics anywhere" statement would have been a negative claim across the whole diagnostics surface that I can't hold to the same standard, and this file's accuracy discipline says a narrower true sentence beats a broader plausible one. Happy to widen it if you can confirm the stronger statement.

Gates re-run on the exact pushed tree: `tests/unit/chat-body-admission.test.ts` 22 pass / 0 fail; `check:docs-all`, `check:fabricated-docs`, `check:doc-links`, `check:docs-symbols`, `check:env-doc-sync`, `check:api-docs-refs`, `check:changelog-integrity` all pass; `git diff --check` clean; the added section is Prettier-clean in isolation (the pre-existing base drift documented in the PR body is unchanged and untouched).

Leaving this as Draft pending your call on scope and on whether to widen the visibility sentence.

diegosouzapw:
## Review: PR #9021 — docs(troubleshooting): document chat_admission_busy 503

**Verdict: merge-ready** (4/5 stars)

Accurate and complete documentation of the `chat_admission_busy` 503 error. All thresholds, response bodies, and env vars verified against source code. Good links to ENVIRONMENT.md.

Ready to merge after disengaging Draft status.

raccoonwannafly:
@diegosouzapw can we push for a v3.8.50 early npm release? this 503 error's been bugging me the last few days and its really making omniroute unusuable, genuinely tried all fixes but none worked.

PR #9022, opened 2026-07-31
fix(sse): preserve error-only stream diagnostics

### LINKED ISSUE #8972: fix(resilience): ensureStreamReadiness discards executor error-only SSE frames
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Node.js Version

_No response_

### Provider(s) Involved

Any executor whose failure path emits an error-only SSE frame. Observed on Microsoft 365 Copilot (BizChat) — `copilot-m365-web`.

### Model(s) Involved

`copilot-m365-gpt-5-5-chat` (and the other three `copilot-m365-*` ids)

### Client Tool

OpenAI-compatible `/v1/chat/completions`

### Description

#7858 added a deliberate fail-loudly path: when a BizChat turn completes with no content in any known frame shape, the executor emits a descriptive error instead of a silent `stop`. That message is well written and names the resolved tier — it is exactly what an operator needs.

**The operator never sees it.** The readiness gate deletes it:

1. The executor enqueues a single SSE frame shaped `{"error":{"message":"…"}}` and closes the stream (`copilot-m365-web.ts:126-129`). The same is true of its `abort()` path (`:142`), which is how WS/handshake errors surface.
2. `ensureStreamReadiness()` inspects buffered frames. `isErrorOnlyStructuredPayload()` returns true for a payload that has `error` and none of `CONTENT_BEARING_KEYS`, and `hasNonPingStructuredPayload()` then reports **not ready** (`streamReadiness.ts:105-108`, `:110-118`).
3. With no ready signal and the stream ended, the function returns `STREAM_EARLY_EOF` and synthesises its own 502 body. The buffered frames — including the only diagnostic — are dropped; `buildReadyResponse()` is never called.
4. So a precise message is replaced by `Stream ended before producing a non-ping SSE event`, which describes the gate's own observation rather than the failure.

I understand the check is intentional (#7503: an error-only frame must not count as ready, or a malformed frame reaches the client as a fake 200 and blocks combo fallback). The gap is that suppressing the frame **as a readiness signal** currently also destroys it **as a diagnostic**. Those can be separated.

Compounding it, nothing else preserves the message: every `M365_WS` diagnostic is `log?.debug?.()`, and the effective level is captured once at module load from `APP_LOG_LEVEL` with an `"info"` default (`open-sse/utils/logger.ts:50-51`, `src/lib/logEnv.ts` `getAppLogLevel`). There is no runtime or database override, so on a default deployment the inner reason is absent from `call_logs.error_summary`, `proxy_logs.error`, and the call-log artifact alike — all three contain only the generic text. Recovering it requires setting an env var and restarting the container.

Real-world cost: diagnosing the enterprise BizChat failure (filed separately) required bypassing OmniRoute with a standalone WebSocket client, because the product had already computed the answer and thrown it away.

### Steps to Reproduce

1. Deploy 3.8.49 with default logging (no `APP_LOG_LEVEL`).
2. Configure a `copilot-m365-web` connection whose turns complete with no content — e.g. an AAD enterprise tenant, `tier = "enterprise"` (see the companion issue).
3. `POST /v1/chat/completions` with any of the `copilot-m365-*` models.
4. Client receives `502 {"error":{"message":"Stream ended before producing a non-ping SSE event","code":"bad_gateway"}}`.
5. Inspect `call_logs`, `proxy_logs`, and the artifact JSON: all carry the same generic string; #7858's message appears nowhere.

### Expected Behavior

- When the only frames are error-only, the readiness failure should carry **their** message rather than a generic one — e.g. `reason` becomes the upstream error text (optionally prefixed), while `code`/`type` still mark it unready so combo fallback is unchanged.
- Failing that, the error-only payload should at minimum be persisted (artifact / `error_summary`) so it is recoverable after the fact.
- Executor-level failure diagnostics should not require `APP_LOG_LEVEL=debug` plus a restart to become visible.

### Actual Behavior

| layer | what it holds |
|---|---|
| executor intent (#7858) | `"Microsoft 365 Copilot turn completed with no content in any known frame shape (resolved tier: enterprise). Possible causes: an unrecognized frame shape for this tenant, or a misconfigured tier."` |
| what the client receives | `Stream ended before producing a non-ping SSE event` |
| `call_logs.error_summary` | `Stream ended before producing a non-ping SSE event` |
| `proxy_logs.error` | `Stream ended before producing a non-ping SSE event` |
| call-log artifact `pipeline.error` | `Stream ended before producing a non-ping SSE event` |
| `providerResponse` in artifact | `null` |

Confirmed on 3.8.49, 37 failing calls across all four `copilot-m365-*` models; every record carried the generic message and none carried the executor's.

### Test Impact

Yes — a unit test asserting that a stream whose only frame is `{"error":{...}}` yields a readiness failure whose `reason` preserves the upstream message while still reporting `ok: false`.

### Error Logs / Output

```json
{"level":40,"module":"sse","tag":"STREAM","msg":"Stream ended before producing a non-ping SSE event (copilot-m365-web/copilot-m365-gpt-5-5-chat)"}
{"level":40,"module":"sse","tag":"STREAM","msg":"copilot-m365-web/copilot-m365-gpt-5-5-chat closed the stream early before useful content — retrying once (attempt 1)"}
```

Artifact excerpt (the executor's message is absent, not truncated):

```json
{
  "error": "Stream ended before producing a non-ping SSE event",
  "pipeline": {
    "error": { "message": "Stream ended before producing a non-ping SSE event" },
    "clientResponse": { "error": { "message": "Stream ended before producing a non-ping SSE event", "type": "server_error", "code": "bad_gateway" } },
    "providerResponse": null
  }
}
```

### Screenshots

_No response_

### Additional Context

- `open-sse/utils/streamReadiness.ts:91-103` — `CONTENT_BEARING_KEYS`.
- `open-sse/utils/streamReadiness.ts:105-108` — `isErrorOnlyStructuredPayload()`.
- `open-sse/utils/streamReadiness.ts:110-118` — `hasNonPingStructuredPayload()` returns false for error-only payloads.
- `open-sse/utils/streamReadiness.ts` `ensureStreamReadiness()` — the `readResult.done` branch returns `STREAM_EARLY_EOF` and synthesises a fresh error response; buffered `chunks` are discarded.
- `open-sse/executors/copilot-m365-web.ts:116-131` (#7858 message) and `:137-144` (`abort()`), both error-only frames.
- `open-sse/utils/logger.ts:50-51` — level fixed at module load; `src/lib/logEnv.ts` `getAppLogLevel()` reads `process.env.APP_LOG_LEVEL` with an `"info"` default and no DB fallback.
- Note the `abort()` path is affected identically, so genuine WS/handshake errors (bad or expired credential, rejected upgrade) are also flattened into the same generic 502 — an operator cannot tell an expired token from a rejected payload without patching the deployment.

### Validation Plan

1. In `ensureStreamReadiness()`, when the buffered frames contained an error-only structured payload, use its `error.message` as the returned `reason` (e.g. prefixed with the provider/model) while keeping `ok: false` and the existing `code`/`type`, so #7503's fallback semantics are untouched.
2. Unit test: a stream emitting exactly one `{"error":{"message":"UPSTREAM_DETAIL"}}` frame then closing → `ok === false` **and** `reason` contains `UPSTREAM_DETAIL` (RED today).
3. Regression test: a stream emitting a normal `choices[].delta.content` frame is still ready, and an error-only frame still does not block combo fallback to the next candidate.
4. Optional: persist the error-only payload into the call-log artifact so past failures stay diagnosable without re-running with debug logging.

shixi-li:
## Summary

- Preserve the first sanitized diagnostic from an error-only SSE stream when readiness ends at EOF.
- Keep the top-level `STREAM_EARLY_EOF` classification stable so combo fallback and account-health behavior do not change.
- Persist the sanitized detail through the existing `upstream_details` response shape and request diagnostics.

## Root cause

`ensureStreamReadiness()` correctly treated error-only frames as unready, but discarded their payload when it synthesized the EOF failure. This left operators with only the gate's generic message.

## Validation

- Focused stream-readiness tests: 18 passed
- Related readiness, fallback, and executor tests: 33 passed
- `npm run check:mutation-test-coverage`
- `npm run typecheck:core`
- `npm run lint`
- `npm run check:cycles`
- `npm run check:error-helper`
- `npm run check:file-size`
- `npm run check:changelog-integrity`

## Scope

No dependency, runtime/deployment configuration, migration, or readiness success-path changes. `stryker.conf.json` only registers the added unit test for mutation-coverage accounting.

Fixes https://github.com/diegosouzapw/OmniRoute/issues/8972

### DIFF
--- a/changelog.d/fixes/9022-stream-error-diagnostic.md
+++ b/changelog.d/fixes/9022-stream-error-diagnostic.md
@@ -1,0 +1,1 @@
+- **fix(sse):** error-only streams now preserve sanitized executor diagnostics for operators without changing stream-readiness fallback classification ([#9022](https://github.com/diegosouzapw/OmniRoute/pull/9022)) — thanks @shixi-li
--- a/changelog.d/fixes/pending-stream-error-diagnostic.md
+++ b/changelog.d/fixes/pending-stream-error-diagnostic.md
@@ -1,1 +1,0 @@
-- **fix(sse):** error-only streams now preserve sanitized executor diagnostics for operators without changing stream-readiness fallback classification ([#8972](https://github.com/diegosouzapw/OmniRoute/issues/8972)).
--- a/changelog.d/fixes/pending-stream-error-diagnostic.md
+++ b/changelog.d/fixes/pending-stream-error-diagnostic.md
@@ -1,0 +1,1 @@
+- **fix(sse):** error-only streams now preserve sanitized executor diagnostics for operators without changing stream-readiness fallback classification ([#8972](https://github.com/diegosouzapw/OmniRoute/issues/8972)).
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -4653,12 +4653,7 @@ export async function handleChatCore({
   });
   if (streamReadiness.ok === false) {
     const { response: failureResponse, reason } = streamReadiness;
-    const failure = {
-      status: failureResponse.status,
-      message: reason,
-      code: streamReadiness.code,
-      type: streamReadiness.type,
-    };
+    const { classificationReason, upstreamDiagnostic } = streamReadiness;
     trackPendingRequest(model, provider, connectionId, false);
     appendRequestLog({
       model,
@@ -4670,7 +4665,11 @@ export async function handleChatCore({
       status: failureResponse.status,
       error: reason,
       providerRequest: finalBody || translatedBody,
-      clientResponse: buildErrorBody(failureResponse.status, reason),
+      clientResponse: buildErrorBody(
+        failureResponse.status,
+        classificationReason,
+        upstreamDiagnostic ? { error: { message: upstreamDiagnostic } } : undefined
+      ),
       claudeCacheMeta: claudePromptCacheLogMeta,
       cacheSource: "upstream",
     });
@@ -4682,6 +4681,7 @@ export async function handleChatCore({
       success: false,
       status: failureResponse.status,
       error: reason,
+      classificationError: classificationReason,
       errorType: streamReadiness.type,
       errorCode: streamReadiness.code,
       response: failureResponse,
--- a/open-sse/utils/streamReadiness.ts
+++ b/open-sse/utils/streamReadiness.ts
@@ -1,4 +1,5 @@
 import { HTTP_STATUS } from "../config/constants.ts";
+import { buildErrorBody, sanitizeErrorMessage } from "./error.ts";
 
 type StreamReadinessLogger = {
   debug?: (tag: string, message: string) => void;
@@ -7,7 +8,18 @@ type StreamReadinessLogger = {
 
 export type StreamReadinessResult =
   | { ok: true; response: Response }
-  | { ok: false; response: Response; reason: string; code: string; type: string };
+  | {
+      ok: false;
+      response: Response;
+      /** Sanitized operator-facing context for logs and persisted diagnostics. */
+      reason: string;
+      /** Stable internal text for retry, quota, and account-health classification. */
+      classificationReason: string;
+      /** First non-empty sanitized message from an error-only SSE payload. */
+      upstreamDiagnostic?: string;
+      code: string;
+      type: string;
+    };
 
 function isRecord(value: unknown): value is Record<string, unknown> {
   return !!value && typeof value === "object" && !Array.isArray(value);
@@ -233,6 +245,7 @@ type StreamReadinessSignalState = {
   currentEvent: string;
   dataLines: string[];
   pendingLine: string;
+  upstreamDiagnostic: string | null;
 };
 
 function resetCurrentEvent(state: StreamReadinessSignalState): void {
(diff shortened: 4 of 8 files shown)


shixi-li:
CI follow-up for the current head `126d3792` (no source changes made):

- All blocking checks are green. The only red check is `Build (advisory)` in [Quality Gates run 30594030601](https://github.com/diegosouzapw/OmniRoute/actions/runs/30594030601/job/91042439104).
- In that job, dependency installation and the Node runtime check completed. The `npm run build` step started at `00:38:43 UTC`, never produced a completion record, and the job ended at `01:29:17 UTC` after roughly 50 minutes.
- A fresh locked local install completed, and the five focused SSE/error-diagnostic test files passed: `78/78`.
- `OMNIROUTE_USE_TURBOPACK=1 npm run build` on this PR reproduced the stall at `Creating an optimized production build ...`. After several minutes the Next/Turbopack process was at `0.0%` CPU with a `5.2 GB` physical footprint; a process sample showed the Node event loop and Turbopack tokio workers waiting on condition variables.
- As a control, I ran the same command with the same locked dependency tree and environment on the current `release/v3.8.50` tip `371c10ea`. That upstream commit does not overlap the PR files and does not change `package-lock.json`. The baseline reproduced the same stage, `0.0%` CPU state, `5.2 GB` footprint, and condition-variable waits.

This isolates the advisory build failure from the changes in this PR: it is reproducible on the current target-branch baseline. I therefore did not alter the implementation or create an empty/retry commit. A maintainer can rerun or treat the advisory check as infrastructure noise; the Turbopack build stall should be investigated independently of this patch.

diegosouzapw:

PR #9027, opened 2026-07-31
fix(chat): resolve stored combo names before image-model validation (#8986)

### LINKED ISSUE #8986: fix: resolve combo names before image-model validation
## Summary

A chat combo whose name collides with an unprefixed image-model ID is rejected before combo routing.

## Reproduction

1. Create a chat combo named `fast` that targets a chat model.
2. Send `POST /v1/chat/completions` or `POST /v1/responses` with `model: "fast"`.

## Actual behavior

The request returns HTTP 400:

```text
Model 'fast' is an image-generation model and cannot be used on /v1/chat/completions.\n```\n\nThis occurs even though the combo is advertised by `/v1/models` as a `chat` model and contains no image target.\n\n## Root cause\n\n`src/sse/handlers/chat.ts` calls `getImageModelEntry(modelStr)` before combo resolution. `getImageModelEntry("fast")` accepts bare IDs and resolves `fast` from the static `stability-ai` image registry (`Fast Upscale`), regardless of whether that provider is configured.\n\n## Expected behavior\n\nA configured combo name should take precedence over static bare image-model matching. A real image-only model must still be rejected on the chat endpoint.\n\n## Suggested fix\n\nMove the image-only validation after combo-name resolution, or exempt known combo names from that validation. Please add a regression test for a chat combo whose name collides with an unprefixed image registry model ID.

xiaoyaner0201:
Fixes #8986

## Problem

`src/sse/handlers/chat.ts` runs `getImageModelEntry(modelStr)` before combo resolution. Because that lookup accepts bare IDs, a stored chat combo named `fast` matched the static `stability-ai` entry `Fast Upscale` and the request was rejected with HTTP 400 (`... is an image-generation model and cannot be used on /v1/chat/completions`) before it ever reached combo routing — even though `/v1/models` advertises the combo as `chat` and it contains no image target.

## Change

Exact stored combo names now take precedence over a colliding bare image alias. The guard consults `getComboByName(modelStr)` only when the image-registry lookup already matched, so ordering is unchanged for every request that does not hit this collision.

Deliberately narrower than `getComboForModel()`: model→combo mappings and synthetic/auto aliases keep their existing resolution order. No image registry, combo schema, migration, provider configuration, dependency, lockfile, UI or docs change.

Scope: 2 files, +36/-3.

- `src/sse/handlers/chat.ts`
- `tests/unit/chat-rejects-image-only-model.test.ts`

## Behavior matrix (verified by route-level probes)

| Case | Before | After |
| --- | --- | --- |
| stored chat combo `fast` (targets `openai/gpt-4o`) | 400 image-generation, 0 dispatch | 200, routed to `gpt-4o` |
| bare image alias `fast` with no combo stored | 400 + `/v1/images/generations` hint | unchanged |
| real image-only model (HuggingFace) | 400 + generations hint | unchanged |
| model registered for both chat and image (`codex/gpt-5.6-sol`) | allowed | unchanged |
| unknown bare model | 400 | unchanged |

## Tests

TDD lineage is preserved as two commits: `ff4863180` (RED, tests only) then `6c4c0a007` (GREEN, minimum fix).

- RED reproduced at `ff4863180`: 3 pass / 1 fail, failure reason "stored combo fast rejected by image-generation guard".
- Focused: `node --import tsx/esm --test tests/unit/chat-rejects-image-only-model.test.ts` → 4/4 pass.
- Related unit sweep (247 files matching chat / image / combo / api-key-policy / bare-model / schema-validation / body-admission): 1733 tests, 1730 pass, 3 fail — all 3 pre-existing and unrelated. `json-migration-combos.test.ts` (2 failures) fails identically on the untouched base `9b3efef80`; `combo-resource-404-health.test.ts` fails only with an `ENOTEMPTY` tmpdir teardown race under high parallelism and passes 1/1 in isolation.
- Static gates all green at this tree: Prettier and ESLint on both changed files, `typecheck:core`, `check:cycles`, `check:db-rules`, `check:build-scope`, `check:error-helper`, `check:public-creds`, `check:openapi-security-tiers`, `git diff --check`.

Environment note: the local `better-sqlite3` native binary does not match this Node ABI, so the repository's SQLite fallback driver was used throughout. That does not affect the guard being exercised.

Base: `9b3efef806b291bd5833ba678b57c4e071b65255` (current `release/v3.8.50` tip). Head: `6c4c0a00749f6d60287a1254be74dfeb1fc935cd`, tree `93479165d54c843c23684dc30f1a85468e660d19`.

Opened as Draft for maintainer review.

### DIFF
--- a/src/sse/handlers/chat.ts
+++ b/src/sse/handlers/chat.ts
@@ -44,7 +44,7 @@ import { checkAndRefreshToken } from "../services/tokenRefresh";
 import { createHookContext, runHooks, initPreRequestRegistry } from "@/lib/middleware/registry";
 import { rejectPeerRequest } from "@/shared/resilience/peerRouting";
 import { deleteHandoff, getHandoff } from "@/lib/db/contextHandoffs";
-import { updateCombo } from "@/lib/db/combos";
+import { getComboByName, updateCombo } from "@/lib/db/combos";
 import { isModelAllowedForKey } from "@/lib/db/apiKeys";
 import { promoteSuccessfulComboModel } from "@/lib/combos/autoPromote";
 import {
@@ -453,10 +453,14 @@ export async function handleChat(
   // image-registry match is only image-only when the same provider/model pair is
   // absent from the chat catalog.
   const imageModel = getImageModelEntry(modelStr);
+  // Exact stored combo names take precedence over colliding bare image aliases.
+  // Keep this narrower than getComboForModel() so mappings and synthetic aliases
+  // retain their existing resolution order.
+  const isExactStoredCombo = imageModel ? Boolean(await getComboByName(modelStr)) : false;
   const isChatCatalogModel = imageModel
     ? getModelsByProviderId(imageModel.provider).some((model) => model.id === imageModel.model)
     : false;
-  if (imageModel && !isChatCatalogModel) {
+  if (imageModel && !isExactStoredCombo && !isChatCatalogModel) {
     log.warn("CHAT", `Rejecting image-generation model on chat endpoint: ${modelStr}`);
     return errorResponse(
       HTTP_STATUS.BAD_REQUEST,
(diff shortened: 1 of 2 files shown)


Erick-Franco (approved):

diegosouzapw:
## Review: PR #9027 — fix(chat): resolve stored combo names before image-model validation

**Verdict: merge-ready** (5/5 stars)

Minimal fix (+6 lines production code) resolving combo names before image-model validation. Behavior matrix correctly maps all 5 cases. Well tested.

Ready to merge.

PR #9035, opened 2026-07-31
fix(kiro): read usage from the frames Kiro actually sends

ddarkr:
## Summary

Kiro's token and cache accounting never reached the client. All usage extraction lived inside an `eventType === "metricsEvent"` branch, and live `generateAssistantResponse` traffic does not send that frame — so the code never ran, regardless of which field names it read.

This supersedes #8907, which corrected the cache field names but left them in that unreachable branch.

## Live verification

Verified directly against the API with a test credential. The credential value and `Authorization` header were never printed, committed, or added to fixtures; account identifiers, entitlements and the account-specific model list are excluded. Only frame names and payload shapes needed to validate the parser were kept.

A successful stream contained exactly:

```text
assistantResponseEvent
metadataEvent      {"stopReason":"END_TURN"}
contextUsageEvent  {"contextUsagePercentage":10.867}
meteringEvent      {"unit":"credit","unitPlural":"credits","usage":0.0897}
```

Two observations follow, and they drive the whole patch:

1. **No `metricsEvent`.** The branch holding every usage/cache read is dead on this traffic.
2. **No token counts anywhere.** `metadataEvent` carries only `stopReason`. Kiro reports a context *percentage* and a *credit* figure — never `inputTokens`/`outputTokens`, and never `cacheReadInputTokens`/`cacheWriteInputTokens`.

The prompt used was ~75 KB (well past the 1024-token minimum checkpoint the catalog declares), so the absent cache fields are not a below-threshold artifact.

## Changes

**Read the frames that actually arrive**

- Extract usage from `metadataEvent` as well as `metricsEvent`, including a nested `usage` object.
- Accept snake_case cache spellings alongside the Bedrock camelCase ones.
- Keep cache counts that arrive on a frame with no input/output totals. The old `inputTokens > 0 || outputTokens > 0` gate discarded the entire frame, taking the cache accounting with it.

**Scale the token estimate by the model's context window**

`ensureKiroUsage` multiplied `contextUsagePercentage` by a hardcoded `200000`. Since Kiro reports only a percentage, the budget it applies to decides the result:

| model | window | before | after |
|---|---|---|---|
| claude-sonnet-4.5 | 200k | 21,734 | 21,734 |
| claude-sonnet-5 | 1M | 21,734 | **108,670** |
| gpt-5.6-sol | 272k | 21,734 | **29,558** |

(measured at the observed `contextUsagePercentage: 10.867`)

These numbers feed `usage_history` and the API-key token-limit counters, so `claude-sonnet-5` was undercounted 5x. The per-model `contextLength` was already in the registry; it just was not consulted. Unknown ids fall back to `defaultContextLength`.

The completion estimate is now **subtracted** from the reported total instead of added on top — the percentage already covers the whole context, so adding it double-counted and inflated `total_tokens`.

**Preserve capability metadata** (carried over from #8907)

`ListAvailableModels.promptCaching` and its checkpoint limits are kept on discovered model variants. Nonnumeric or nullable limits stay `null` rather than coercing to a misleading `0`.

## Scope and honesty about limits

Token counts here are **estimates**, and the code says so. Kiro sends no token counts on this endpoint, so there is nothing exact to report.

Cache read/write **cannot** be derived from a single context percentage, and this PR does not try. It does not add `kiro` to provider-wide `CACHING_PROVIDERS` (capability is per model), infer cache hits from `meteringEvent`, or alter request headers. If upstream ever supplies cache counters on either frame, they are now read and surfaced; until then nothing is fabricated.

## Tests

```bash
node --import tsx/esm --test tests/unit/executor-kiro.test.ts tests/unit/kiro-available-models.test.ts
# 34 passed, 0 failed
```

Each new test was confirmed to fail against the pre-fix executor, so they defend real behavior rather than restating it:

- usage + cache tokens read from a `metadataEvent`
- cache tokens kept when input/output totals are absent
- snake_case cache spellings accepted
- estimate scaled per model, including `defaultContextLength` inheritance and unknown-id fallback
- completion carved out of the reported total, not added to it
- completion estimate still reported when no context percentage arrives
- the live API-key frame shape continues to invent no cache fields

### DIFF
--- a/open-sse/executors/kiro.ts
+++ b/open-sse/executors/kiro.ts
@@ -6,6 +6,7 @@ import {
   type ProviderCredentials,
 } from "./base.ts";
 import { PROVIDERS } from "../config/constants.ts";
+import { getRegistryEntry } from "../config/providerRegistry.ts";
 import { v4 as uuidv4 } from "uuid";
 import { refreshKiroToken } from "../services/tokenRefresh.ts";
 import {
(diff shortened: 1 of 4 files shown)


ddarkr:
CI triage — both red jobs are unrelated to this patch. I lack rerun rights on this repo, so flagging rather than retrying.

**`Fast Quality Gates` → `duckduckgo-web-executor.test.ts`**

```
✖ should complete within timeout (71528ms)
  testConnection should complete within 35 seconds, took 71527ms
ℹ tests 2391 · pass 2389 · fail 1
```

That test performs a real network call and asserts wall-clock time:

```js
await executor.testConnection({});                    // live DuckDuckGo request
assert.ok(elapsed < 35000, `... took ${elapsed}ms`);
```

It fails whenever the runner's DuckDuckGo round-trip is slow; this run took 71s. It has no connection to these changes — the file contains no `kiro` reference, its only import is `open-sse/executors/duckduckgo-web.ts`, and this patch touches `kiro.ts`, `kiroModels.ts` and their tests only. #9036, pushed at the same time against the same base, did not hit it.

**`Build (advisory)`**

`npm run build` was cancelled (job timeout, 7m13s) with no error output. It reproduces identically on #9036, which shares no files with this PR — pointing at the base/runner rather than either patch. The job is advisory.

Kiro tests do not appear among the failures. Locally on this branch:

```bash
node --import tsx/esm --test tests/unit/executor-kiro.test.ts tests/unit/kiro-available-models.test.ts
# 34 passed, 0 failed
```

Happy to rebase if the base has since moved, but I have not bundled a fix for the flaky network test or the build baseline into this PR — neither is in its scope.

PR #9036, opened 2026-07-31
fix(kiro): keep relocated tool documentation on multi-turn requests

ddarkr:
## Summary

Kiro rejects a `toolSpecification.description` over ~10000 chars, so the translator relocates an oversized description into the current turn's content and leaves a pointer in the schema. On **any multi-turn request the relocated documentation was dropped**, leaving the model with a pointer to text it never received.

## The bug

The relocated text was stashed on the message object as `_toolDocs`, but `buildKiroPayload` read it from `currentMessage` only. The tool-bearing user turn is moved into `history` on every request with more than one user turn, so the docs went with it and were never emitted.

Reproduced with one tool whose description exceeds the limit:

```text
single-turn  docs in current content: true
multi-turn   docs in current content: false   ← model sees only the pointer
```

The schema still said `[Full documentation in system prompt under '## Tool: X']`, so the model was told where to look and the section was not there.

`_toolDocs` also rode along inside `conversationState.history`, sending ~12 KB of a field Kiro has no use for. It sat nested rather than at the top level, so `transformRequest`'s whitelist did not strip it. I did not observe Kiro rejecting it — the cost is wasted payload, not a 400.

A second gap: the fallback tool-attachment path (used when no user turn carried the schema) bypassed the length guard entirely, emitting full oversized descriptions that Kiro *does* reject.

## Fix

Return the relocated docs from `convertMessages` instead of carrying them on a message, so they survive the turn being moved into history, and route both attachment paths through one `buildKiroToolSpecs` builder so the guard cannot be bypassed.

This also removes the `_toolDocs` field entirely, so nothing extraneous is serialized.

## Tests

```bash
node --import tsx/esm --test tests/unit/kiro-long-tool-description-docs.test.ts \
  tests/unit/translator-openai-to-kiro.test.ts \
  tests/unit/openai-to-kiro-helpers-split.test.ts \
  tests/unit/kiro-tool-schema-sanitize-1375.test.ts \
  tests/unit/kiro-system-reminder-2306.test.ts \
  tests/unit/kiro-continue-filler-5231.test.ts
# 62 passed, 0 failed
```

New coverage, each confirmed to fail against the pre-fix translator:

- docs reach the current turn across single-turn, multi-turn, deep multi-turn, assistant-first, and no-user-turn shapes
- no unknown field is serialized into the payload
- a description within the limit is left verbatim and adds no documentation block
- the boundary triggers above 10000, not at it
- in a mixed inventory only the oversized tool is relocated

Also verified manually: multiple oversized tools each get their own `## Tool: <name>` section, Anthropic-shaped declarations (`input_schema` at top level) work, and the documentation block is prepended ahead of the timestamp so the cacheable prefix ordering is unchanged.

### DIFF
--- a/open-sse/translator/request/openai-to-kiro.ts
+++ b/open-sse/translator/request/openai-to-kiro.ts
@@ -46,6 +46,69 @@ function wrapSystemReminder(text: string): string {
   return `<system-reminder>\n${text}\n</system-reminder>`;
 }
 
+/** Kiro rejects a `toolSpecification.description` longer than ~10000 chars. */
+const KIRO_TOOL_DESC_MAX = 10000;
+
+/** OpenAI- and Anthropic-shaped tool declarations, as clients actually send them. */
+type KiroToolInput = {
+  name?: string;
+  description?: string;
+  parameters?: unknown;
+  input_schema?: unknown;
+  function?: { name?: string; description?: string; parameters?: unknown };
+};
+
+/**
+ * Build Kiro `toolSpecification` entries, relocating any oversized description
+ * out of the schema and returning it separately.
+ *
+ * Kiro answers a raw upstream 400 for a description over
+ * {@link KIRO_TOOL_DESC_MAX}, so the schema keeps a pointer and the full text is
+ * handed back to be prepended to the current turn's content — the same
+ * relocation kiro-gateway performs in
+ * `converters_core.py::process_tools_with_long_descriptions`.
+ *
+ * The docs are *returned* rather than stashed on the message object, because the
+ * tool-bearing user turn is moved into `history` on every multi-turn request
+ * (see the currentMessage promotion below). Carrying them on the message lost
+ * them there — the model then saw only the pointer and no documentation — and
+ * also leaked an unknown `_toolDocs` field into the upstream payload, which Kiro
+ * rejects.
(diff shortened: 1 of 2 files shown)


ddarkr:
CI triage — the single red job is unrelated to this patch. I lack rerun rights on this repo, so flagging rather than retrying.

**`Build (advisory)`**

`npm run build` was cancelled (job timeout, 7m59s) with no error output in the failed step. The same failure reproduces on #9035, which shares no files with this PR — this patch touches only `open-sse/translator/request/openai-to-kiro.ts` plus a new test file. A failure common to two disjoint changesets points at the base or the runner, not either patch. The job is advisory.

Blocking gates passed, and the translator tests are green locally on this branch:

```bash
node --import tsx/esm --test tests/unit/kiro-long-tool-description-docs.test.ts \
  tests/unit/translator-openai-to-kiro.test.ts \
  tests/unit/openai-to-kiro-helpers-split.test.ts \
  tests/unit/kiro-tool-schema-sanitize-1375.test.ts \
  tests/unit/kiro-system-reminder-2306.test.ts \
  tests/unit/kiro-continue-filler-5231.test.ts
# 62 passed, 0 failed
```

Happy to rebase if the base has since moved.

PR #9044, opened 2026-07-31
fix(db): bundle and verify the sql.js fallback

nguyenha935:
## What changed

- keep SQLite driver imports as literal `require()` calls so the standalone bundler emits real externals instead of missing-module stubs
- ship the complete `sql.js` runtime, including its WASM asset, in standalone packages
- make the `sql.js` adapter resolve its runtime from the installed package tree
- extend `check-pack-boot` with a forced-`sql.js` round trip and a two-boot persistence check using the same `DATA_DIR`
- add regression coverage for driver order, writable reopen behavior, standalone assembly, externalization, and persistence evaluation

## Root cause

The fallback path worked from source but was not a reliable packaged-runtime contract. Passing `require` through a loader callback prevented the bundler from seeing literal SQLite module specifiers, while the standalone assembly did not guarantee that the complete `sql.js` runtime was present. This could leave an installed package unable to recover when a native SQLite binding was unavailable.

## Impact

Packaged OmniRoute installations can safely preserve the existing driver cascade:

1. `better-sqlite3`
2. `node:sqlite`
3. `bun:sqlite`
4. `sql.js`

The pack-boot gate now verifies that the fallback can write settings, shut down cleanly, start again against the same database directory, and read the persisted value back.

## Validation

- focused SQLite suite after rebasing onto the current `release/v3.8.50`: **34 passed, 0 failed, 1 skipped**
- the single skip is the native `better-sqlite3` preference probe in a scratch image without that native binding; cascade and fallback behavior remain covered
- production artifact verification on a deployed patched build:
  - graceful shutdown completed a WAL checkpoint and closed SQLite
  - the replacement process reopened the same `/opt/omniroute/data/storage.sqlite`
  - runtime state and existing proxy logs were loaded after restart
  - `PRAGMA quick_check` returned `ok`; journal mode is `wal`
  - `sql.js/dist/sql-wasm.wasm` is present in the installed runtime

No Home topology or Model Identity changes are included in this PR.

### DIFF
--- a/scripts/build/assembleStandalone.mjs
+++ b/scripts/build/assembleStandalone.mjs
@@ -215,6 +215,11 @@ const EXTRA_MODULE_ENTRIES = [
     dest: ["node_modules", "undici"],
   },
   {
+    label: "sql.js WASM fallback runtime",
+    src: ["node_modules", "sql.js"],
+    dest: ["node_modules", "sql.js"],
+  },
+  {
     label: "sqlite-vec wrapper (vector memory - loaded at runtime via createRequire)",
     src: ["node_modules", "sqlite-vec"],
     dest: ["node_modules", "sqlite-vec"],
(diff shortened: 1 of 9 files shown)


diegosouzapw:

|---|---|---|

PR #9053, opened 2026-07-31
fix(translator): pass output_config.effort=max through verbatim

ikelvingo:
## Summary

Fixes a contract mismatch where `output_config.effort="max"` from an Anthropic-format request was incorrectly rewritten to `xhigh` when translating to an OpenAI-compatible upstream. This caused upstreams that accept `max` literally but reject `xhigh` — such as Ollama Cloud — to return HTTP 400 `invalid reasoning value: 'xhigh'`.

## Root Cause

`open-sse/translator/request/claude-to-openai.ts::normalizeOpenAIReasoningEffort` unconditionally rewrote `max → xhigh` during step 1 (translation). However, provider-aware allowlist decisions belong in `sanitizeReasoningEffortForProvider` (the executor layer), which holds the full provider capability table — including providers that explicitly opt in to `supportsMaxEffortForProvider()`: ollama-cloud, opencode-go deepseek, moonshot k3, and native claude.

Because the translator runs before the sanitizer, the carrier was already `xhigh` by the time the sanitizer saw it, so the `effortStr === "max"` branch could never fire — the `max` information was lost before reaching the sanitizer.

## Fix

Removed the `if (normalized === "max") return "xhigh"` line in `claude-to-openai.ts`. The translator now only performs form conversion (lowercase + non-empty filtering) and no longer makes allowlist decisions; `max` is written verbatim to `result.reasoning_effort`. Also updated the top comment in `reasoningEffort.ts` (English) to document this division of responsibility.

## Impact

- **Fix**: Anthropic clients → OpenAI-shape upstreams that accept `max` literally (e.g. ollama-cloud) now pass `max` through unchanged instead of returning 400.
- **No regression**: the sanitizer's `effortStr === "max"` branch already covered the "client sends `max` directly in OpenAI-format" case; it now also covers the new "translator passes `max` through" path.

## Verification

- `tests/unit/translator-claude-to-openai.test.ts` — assertion updated to expect `max` pass-through.
- `tests/unit/base-executor-sanitize-effort.test.ts` — added end-to-end test: `translateRequest(CLAUDE→OPENAI)` + `sanitizeReasoningEffortForProvider("ollama-cloud", …)` chain keeps `output_config.effort="max"` as `max` end-to-end.

## Repro (before fix)

```bash
curl -X POST http://localhost:20128/v1/messages \
  -d '{
    "model": "ollama-cloud/gemma4:31b",
    "max_tokens": 4096,
    "messages": [{"role":"user","content":"hi"}],
    "output_config": {"effort": "max"}
  }'
# HTTP 400 {"error":{"message":"invalid reasoning value: 'xhigh'
#   (must be \"high\", \"medium\", \"low\", \"max\", or \"none\")"}}
```

After fix: HTTP 200, `reasoning_effort: "max"` forwarded to ollama-cloud.


diegosouzapw:

Issue #3825, opened 2026-06-14
[BUG]  Frequent 504 Errors and Significant TPS Degradation After Upgrading Beyond v3.8.14


bypanghu:
### OmniRoute Version

3.8.14

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Description

My current version is v3.8.14. When I try to upgrade to the latest version or any later version (for example, v3.8.17), I frequently encounter the following error:

`[504]: Upstream request did not return response headers after **`

This issue only occurs with Combo models. It is not limited to the Round Robin strategy; it also happens with the Weighted strategy.

As a result, the overall service performance degrades significantly. TPS drops from the normal 30–40 range down to 12–20, and in some cases even to single digits. This causes serious issues for our relay service.

I am currently investigating the root cause, but I think it would be better to report this as an issue first.

One additional observation: after downgrading, I checked the database and found that the `context_cache_protection` field for all records in the `combos` table was set to `0`.

### Steps to Reproduce

An additional observation is that TPS remains completely normal on v3.8.14. However, immediately after upgrading to any newer version (such as v3.8.17), TPS drops significantly without any other configuration changes.

The only change made during testing was the application version itself. All model configurations, combo settings, routing strategies, databases, and infrastructure remained unchanged.

This strongly suggests that the performance degradation is related to changes introduced after v3.8.14 rather than to environmental or configuration differences.

### Expected Behavior

  My current version is v3.8.14. When I try to upgrade to the latest version or any later version (for example, v3.8.17), I frequently encounter the following error:

`[504]: Upstream request did not return response headers after **`

This issue only occurs with Combo models. It is not limited to the Round Robin strategy; it also happens with the Weighted strategy.

As a result, the overall service performance degrades significantly. TPS drops from the normal 30–40 range down to 12–20, and in some cases even to single digits. This causes serious issues for our relay service.

I am currently investigating the root cause, but I think it would be better to report this as an issue first.

One additional observation: after downgrading, I checked the database and found that the `context_cache_protection` field for all records in the `combos` table was set to `0`.

### Actual Behavior

My current version is v3.8.14. When I try to upgrade to the latest version or any later version (for example, v3.8.17), I frequently encounter the following error:

`[504]: Upstream request did not return response headers after **`

This issue only occurs with Combo models. It is not limited to the Round Robin strategy; it also happens with the Weighted strategy.

[...]

PR #9079, opened 2026-07-31
fix(classify): recognize Modal 'usage limit reached' as quota exhausted

HouMinXi:
Modal-hosted OpenAI-compatible endpoints (self-hosted Kimi K3 via Modal free tier) return HTTP 429 with body `{"error":"usage limit reached"}` when the account's credit is exhausted.

**Problem**: No `QUOTA_PATTERNS` regex matched this bare-string error shape, so the 429 fell through to `rate_limit` (60s short cooldown). Combined with combo round-robin's per-conversation session stickiness (#3825), this kept re-targeting the same exhausted connection every turn instead of locking it out and failing over to an account with remaining credit.

**Fix**: Add a substring pattern matching `"error":"usage limit reached"` with tolerance for trailing punctuation and whitespace. Only the `"error"` JSON key matches; different keys or qualified transient messages like `"Per-minute usage limit reached"` stay classified as `rate_limit`.

**Verification**:
- 21 unit tests passing (including bug-injection: inject->FAIL, revert->PASS)
- Forge review: PASS (gemini-omniroute backend, Round 7)
- Non-ASCII gate: clean

### DIFF
--- a/src/shared/utils/classify429.ts
+++ b/src/shared/utils/classify429.ts
@@ -66,6 +66,23 @@ const QUOTA_PATTERNS: ReadonlyArray<RegExp> = [
   // the 429 is misclassified as transient rate_limit and retried every
   // ~60s against a budget that only resets at UTC midnight.
   /daily free allocation/i,
+
+  // Modal-hosted OpenAI-compatible endpoints (e.g. self-hosted Kimi K3).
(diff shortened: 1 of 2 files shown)


diegosouzapw:

PR #9116, opened 2026-08-01
[TS7] fix(types): validate combo model collections

backryun:
## Summary

- Restore the normalized combo producer contract so `models` is typed as `ComboStep[]`.
- Keep `getCombo()`'s exact, `combo/`-prefixed, id, and case-insensitive lookup behavior unchanged.
- Add regression coverage for all four lookup paths and malformed persisted model collections.
- Add the required maintenance changelog fragment.

## Root cause

`normalizeComboRecord()` already converts any stored `models` value through `normalizeComboModels()`,
which uses an explicit `Array.isArray` boundary and returns `ComboStep[]`. `normalizeStoredCombo()`
then erased that producer guarantee with a `JsonRecord` return type and a cast, leaving
`getCombo()` consumers with `unknown` for `combo.models`.

The fix exposes the existing producer guarantee as `NormalizedCombo` and annotates
`getComboById`, `getComboByName`, and `getComboByNameInsensitive`. No runtime lookup logic changed.

## Base and head

- Base: `release/v3.8.50` at `1b5f7dd7e6808be31d6a258b57b60b2941058897`
- Head: `f73126c59b0b82921927812efa0f55fb4c00ffb0`

## TypeScript diagnostic multiset

Measured with the explicit `open-sse/tsconfig.json` TS6/TS7 commands. The comparison removes
line/column information and worktree-specific absolute paths only; diagnostic messages are preserved.

- TypeScript 6: `125 -> 121`; removed `4`; added `0`
- TypeScript 7: `124 -> 120`; removed `4`; added `0`
- The four removed diagnostics are the four `TS2339` `combo.models.length` accesses in
  `src/sse/services/model.ts`.

## Verification

- Focused combo/db tests: `12/12` passed
- `npm run lint`: passed
- `npm run check:file-size`: passed
- Prettier check: passed
- `git diff --check`: passed
- `npm run check:changelog-integrity`: passed
- `npm run check:complexity-ratchets`: passed
- Commit hooks: docs-sync, any-budget, and tracked-artifacts checks passed

### DIFF
--- a/changelog.d/maintenance/9116-validate-combo-model-collections.md
+++ b/changelog.d/maintenance/9116-validate-combo-model-collections.md
@@ -1,0 +1,1 @@
+- **fix(types):** combo lookup producers now expose normalized model collections as arrays, preserving exact, prefixed, id, and case-insensitive resolution paths ([#9116](https://github.com/diegosouzapw/OmniRoute/pull/9116))
--- a/src/lib/db/combos.ts
+++ b/src/lib/db/combos.ts
@@ -7,11 +7,12 @@ import { getDbInstance } from "./core";
 import { backupDbFile } from "./backup";
 import { invalidateDbCache } from "./readCache";
 import { invalidateReasoningRoutingRuleCache } from "./reasoningRoutingRules";
-import { normalizeComboRecord } from "@/lib/combos/steps";
+import { normalizeComboRecord, type ComboStep } from "@/lib/combos/steps";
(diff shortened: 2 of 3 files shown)


backryun:
Closing as superseded by #8809, which already removes the same four TS6/TS7 diagnostics at the same normalizeStoredCombo producer boundary. Keeping both open would duplicate the migration slice.

PR #9125, opened 2026-08-01
fix(combo): network errors must not trip provider circuit breaker

HouMinXi:
Network-layer errors (ECONNREFUSED, ETIMEDOUT, etc.) are transport failures — we never reached the provider. Previously these triggered the whole-provider circuit breaker via isProxyUnreachable override, blocking ALL accounts. Now shouldRecordProviderBreakerFailure returns false when isProxyUnreachable is true. Per-model lockout is the correct resilience response.


HouMinXi:
Superseded by 9342 -- this one is CONFLICTING against the current
release branch. The replacement is re-cut from upstream/release/v3.8.50 with
the same fix, a corrected file-size baseline, and the dead isNetworkError
branch removed.

PR #9151, opened 2026-08-01
fix(rate-limit): patch Bottleneck doExpire capacity leak

HouMinXi:
## Problem

Bottleneck v2.19.5's `doExpire` (Job.js:162) has a typo that compares the job ID to the string `"RUNNING"` instead of checking the job's actual status. When a job's execution time exceeds the `expiration` timeout, `doExpire` fires but fails to advance the job from RUNNING to EXECUTING. The resulting exception is swallowed by `setTimeout`, leaving the job permanently stuck in RUNNING state and Bottleneck's internal `_running` counter permanently incremented -- a capacity leak that eventually causes "all accounts inactive" even with available reservoir.

The bug is unfixed upstream (v2.19.5 is the latest release; master branch still has it).

## Fix

1. **Monkey-patch `Bottleneck.prototype._run`** to correct `job.doExpire` before the expiration timeout fires. The patch uses `job.id` (not `job.options.id`) for status lookup, includes a `_doExpirePatched` guard against recursive wrapping, and adds optional chaining on `states.next`.

2. **Diagnostic fallback** in `withRateLimit`: when a queue timeout fires and the limiter shows `running > 0, executing = 0, reservoir > 0` (the leak pattern) with no recent dispatch, evict the limiter and retry.

## Files changed

- `open-sse/services/bottleneckPatch.ts` -- new, 80 lines
- `open-sse/services/rateLimitManager.ts` -- +20 lines (import, patch call, diagnostic)
- `tests/unit/bottleneck-doexpire-patch.test.ts` -- new, 148 lines (5 tests)

## Testing

5/5 unit tests pass. Forge review: 6 rounds, PASS (R6 0 new findings).


HouMinXi:
Superseded by #9328, which is the same fix re-cut onto the live release/v3.8.50 tip. This branch's merge-base sat 20 commits back and its file-size-baseline commit re-encoded the whole file, so the diff read +432/-184; #9328 is 4 files, +248/-0.

Issue #7976, opened 2026-07-21
fix(providers): combo model ids double-prefixed with OC-gated provider id, breaks Smart Stack combos in opencode


fabioluissilva:
I'm running the opencode plugin (0.2.0, same as current main) against my own OmniRoute instance (3.8.48) and every "Smart Stack" combo I have fails from inside opencode with:

```
No Active credentials for provider: opencode-omniroute
```

Regular models route fine, it's only combos (I have "Hermes Smart Stack" and "codex-auto-review" configured). Selecting them in the opencode picker or setting them as `model` in opencode.json both hit this. The combo itself works fine from the OmniRoute web UI and from curl directly against `/v1/chat/completions`, so it's not an API key or connection problem on the OmniRoute side.

Ran `opencode models` and noticed the combo is double-prefixed compared to every other model in the list:

```
opencode-omniroute/opencode-zen/big-pickle                  normal model, fine
opencode-omniroute/gemini/gemini-3.5-flash                  normal model, fine
opencode-omniroute/opencode-omniroute/hermes-smart-stack    combo, broken
```

Traced it to `buildStaticProviderEntry` in src/index.ts (around line 16258 in my compiled dist/index.js). When it builds the combo's model key it does:

```ts
models[buildComboKey(combo, usedComboKeys, opts.providerId)] = entry;
```

`opts.providerId` at that point is the OC-gated id, `opencode-omniroute`, not the bare server-facing one. This whole block gets nested under `provider["opencode-omniroute"]` in the opencode config afterwards, so opencode namespaces the key again on top of it. A combo id with no slash in it, like `hermes-smart-stack`, ends up doubled: `opencode-omniroute/opencode-omniroute/hermes-smart-stack`.

Regular models don't run into this because their raw ids already contain a slash (something like `opencode-zen/big-pickle`), so the prefixing branch just gets skipped for those. It's only the slug-style combo ids that get double-prefixed.

What's interesting is the dynamic provider hook, the provider.models() path a few hundred lines earlier in the same file, already does this correctly. It calls buildComboKey with resolved.omnirouteProviderId, the unprefixed id, and there's even a comment above resolveOmniRoutePluginOptions referencing ticket #6859 that specifically warns against using the OC-gated providerId anywhere OmniRoute's own server has to parse it back out. buildStaticProviderEntry just never got the same fix applied.

[...]

Issue #6859, opened 2026-07-11
[Bug] [Plugin] Provider ID mismatch: plugin uses prefixed "opencode-omniroute" for OmniRoute server calls instead of "omniroute" causing "No active credentials" error


ahbeeahkao:
### OmniRoute Version

3.8.46

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Ubuntu 24.04.4 LTS (Noble)

### Node.js Version

24.18.0

### Provider(s) Involved

Any

### Model(s) Involved

Any

### Client Tool

Opencode

### Description

The OmniRoute OpenCode plugin uses a single prefixed provider ID (opencode-omniroute) for both OpenCode's internal auth system and all OmniRoute server API calls.

    OmniRoute server expects the real provider ID omniroute when parsing model strings like omniroute/model-name. When the plugin sends opencode-omniroute/mycoding, the server parses the provider as
    opencode-omniroute and fails to find credentials.

Root Cause

    File: @omniroute/opencode-plugin/src/index.ts
    Function: resolveOmniRoutePluginOptions() (line ~239)

    The plugin returns a single providerId that gets auto-prefixed for OpenCode 1.17.8+ compatibility:
    typescript
    const providerId = rawProviderId.startsWith("opencode-")
      ? rawProviderId
      : opencode-${rawProviderId};  // "omniroute" → "opencode-omniroute"

    This same providerId is then used for:
    1. ✅ AuthHook.provider (OpenCode internal — correct)
    2. ❌ Model ID prefix: ${providerId}/${raw.id} → opencode-omniroute/model
    3. ❌ providerID field in ModelV2
    4. ❌ All OmniRoute API calls (/v1/models, /api/combos, /chat/completions)

    OmniRoute server expects omniroute (the real provider ID), not opencode-omniroute.

### Steps to Reproduce

bash
    1. Start OmniRoute server
    omniroute serve

    2. Install and configure plugin with auth
    omniroute setup opencode --provider-id opencode-omniroute --auth
    Enter valid OmniRoute API key

    3. Use any model through the plugin
    opencode run "test" -m opencode-omniroute/oc/big-pickle

    Error in OmniRoute server logs:

    {"level":40,"tag":"AUTH","msg":"No credentials for opencode-omniroute"}
    {"level":40,"tag":"AUTH","msg":"No active credentials for provider: opencode-omniroute"}

### Expected Behavior

Expected Behavior
[...]

PR #9178, opened 2026-08-01
fix: drop provider prefix from static-catalog model dict keys (#9175)

### LINKED ISSUE #9175: fix(providers): opencode-plugin v0.2.1 combos still fail — buildStaticProviderEntry adds provider prefix to model dict keys (#9028 closed prematurely)
### Summary

The `opencode-omniroute` / "No active credentials" symptom from #9028 persists in the published `@omniroute/opencode-plugin@0.2.1` against `omniroute@3.8.49`. The maintainer's close rationale in #9028 is wrong. The bug is in the current published artifact: `buildStaticProviderEntry` adds a provider prefix to model dict keys, but OC's `getModel` looks up models by bare ID. The maintainer claimed this was a stale-cache artifact; it is not.

### Verification that #9028's close rationale is wrong

The maintainer of #9028 wrote: "the exact WARN you logged is only producible by a pre-fix plugin build … OpenCode caches loaded plugins per session and may need a full restart / cache clear to pick up the new build."

I cleared `~/.local/share/opencode/plugins/omniroute-opencode-omniroute.json` (the only disk cache the plugin writes), confirmed the loaded version is `vdev (unknown)` (a local build, not a stale npm tarball), and re-ran. The error reproduces **byte-identically** with zero prior network calls. The stale-cache hypothesis is falsified.

### Root cause

In `buildStaticProviderEntry` (the static-catalog path that mutates `input.provider[xxx] = block`):

| Location | What it writes | What OC looks up | Result |
|---|---|---|---|
| Line 16503 (raw models) | `models[<providerId>/<raw-id>]` | `models[<raw-id>]` | miss for non-slashed raw ids |
| Line 16624 (user combos) | `models[<omnirouteProviderId>/<slug>]` | `models[<slug>]` | miss for all user combos |
| Line 16651 (auto combos) | `models[<variant>]` (no prefix) | `models[<variant>]` | hit — this is why `auto/*` works |

`auto/*` combos work by accident because their dict key is just the variant (`auto/coding`) with no prefix. User combos and non-slashed raw models fail because the dict key has a prefix OC doesn't strip.

OC's `getModel` does `provider.models[modelID]` where `modelID` is the part after the first slash in the user's request. For `opencode-omniroute/<combo-slug>`, `modelID` is `<combo-slug>` — but the dict has `<omnirouteProviderId>/<combo-slug>`. The modelSuggestions fuzzy match (threshold `-10000`) returns `<combo-slug>` as a suggestion, which is why the error reads `Did you mean: opencode-omniroute/<combo-slug>?` — the suggestion is the bare model id, prepended with the provider.

### Fix (two lines in `buildStaticProviderEntry`)

1. **Line 16503** — raw models, drop the prefix:
   ```js
   // before
   models[raw.id.includes("/") ? raw.id : `${opts.providerId}/${raw.id}`] = entry;
   // after
   models[raw.id] = entry;
   ```

2. **Line 16624** — user combos, drop the prefix from `buildComboKey`:
   ```js
   // before
   models[buildComboKey(combo, usedComboKeys, opts.omnirouteProviderId)] = entry;
   // after
   models[buildComboKey(combo, usedComboKeys, opts.omnirouteProviderId).split("/").pop()] = entry;
   ```

The PREFIXED form `opencode-omniroute/<slug>` is required (OC 1.17.8+ native-adapter gate rejects bare `omniroute` as a provider ID, hence the auto-prefix in `resolveOmniRoutePluginOptions`). The `buildComboKey` call with `omnirouteProviderId` is correct for collision-detection within the combo namespace — we just need to strip the prefix before writing to the dict.

### Verified

Patched build returns the full static catalog with all user combos in the dict as bare keys. `opencode run "test" -m "opencode-omniroute/<combo-slug>"` succeeds and the model responds. The `auto/*` paths continue to work (negative control).

### Suggested test (per the maintainer's own invitation in #9028)

A new failing test in `@omniroute/opencode-plugin/tests/` that exercises the dashboard-published combo path and asserts:
- The static catalog's `models` dict has bare-id keys (no `omniroute/` or `opencode-omniroute/` prefix) for both raw models and user combos.
- `auto/*` keys remain unprefixed (no regression on the path #8047 fixed).
- The test passes alongside `provider-id-routing.test.ts` (no #6859/6900 regression).

arisgrout:
Fixes #9175.

The static catalog path (`buildStaticProviderEntry`) was writing model dict keys with an embedded provider prefix (e.g. `omniroute/<slug>` or `<providerId>/<raw-id>`). OC's `getModel` looks up models by bare id (the part after the first slash in the user's request), so the prefix made user combos and non-slashed raw models unreachable.

`auto/*` combos worked by accident because their dict key is just the variant (e.g. `auto/coding`) with no prefix.

## Changes

`@omniroute/opencode-plugin/src/index.ts`:

- **Line 4274** (raw models): `models[raw.id.includes("/") ? raw.id : \`${opts.providerId}/${raw.id}\`] = entry;` → `models[raw.id] = entry;`
- **Line 4469** (user combos): `models[buildComboKey(combo, usedComboKeys, opts.omnirouteProviderId)] = entry;` → `models[buildComboKey(combo, usedComboKeys, opts.omnirouteProviderId).split("/").pop()!] = entry;`

## Verification

- `opencode run "test" -m "opencode-omniroute/<combo-slug>"` succeeds; the model responds.
- `auto/*` paths continue to work (negative control).
- All 29 regression tests from #6859/6900 and #7976/8047 should continue to pass — the provider-prefix logic in `buildComboKey` and `resolveOmniRoutePluginOptions` is unchanged; only the final dict write strips the prefix.

## Suggested follow-up

Add a test in `@omniroute/opencode-plugin/tests/` that exercises the dashboard-published combo path and asserts bare-id keys for both raw models and user combos (no `omniroute/` or `opencode-omniroute/` prefix).

### DIFF
--- a/@omniroute/opencode-plugin/src/index.ts
+++ b/@omniroute/opencode-plugin/src/index.ts
@@ -4271,7 +4271,7 @@ export function buildStaticProviderEntry(
     // has no corresponding provider block. So bare keys (no `/`) MUST be
     // prefixed with the resolved providerId. Already-prefixed keys
     // (e.g. `cc/claude-opus-4-7`) are left as-is to avoid double-prefixing.
-    models[raw.id.includes("/") ? raw.id : `${opts.providerId}/${raw.id}`] = entry;
+    models[raw.id] = entry;
   }
 
   // Combo entries → stripped LCD shape. Each combo is keyed as
@@ -4466,7 +4466,7 @@ export function buildStaticProviderEntry(
       // (`opencode-omniroute/opencode-omniroute/<slug>`), and `parseModel()`
       // resolves credentials for the nonexistent provider `opencode-omniroute`
       // instead of `omniroute`. See #7976.
-      models[buildComboKey(combo, usedComboKeys, opts.omnirouteProviderId)] = entry;
+      models[buildComboKey(combo, usedComboKeys, opts.omnirouteProviderId).split("/").pop()!] = entry;
 
       // Make this combo's resolved entry available to parent combos
       // that reference it via combo-ref. Use the friendly name since


drewbitt:
Tried this locally, worked.

drewbitt:
Note: The live models() hook still prefixes bare ids the same way (mapRawModelToModelV2, combo key in createModelsRequest). opencode doesn't call it today, but could be worth fixing. My combos are working though either way.

PR #9189, opened 2026-08-02
fix(claude): preserve standalone whitespace deltas

### LINKED ISSUE #9170: fix(claude): standalone whitespace-only deltas are dropped in Claude←Codex streaming (#8081 regression)
## Summary

On the streaming path where a Claude/Anthropic-format client (Claude Code) is routed to a Codex Responses provider, OmniRoute drops text deltas that contain only whitespace. This removes real word and Markdown boundaries and produces output such as:

```text
cleanup### ContextADR-R08
contain1–253 bytes
```textironbox://worker/<worker-name>```
```

This is a regression/partial reintroduction of the reasoning-placeholder whitespace bug addressed by #8341 and #8382 after #8081.

## Environment

- OmniRoute npm package: `3.8.49`
- npm `gitHead`: `e6523da2a12f0eaebc96ba95650efeec2436ab4d`
- Client format: `claude`
- Provider format: `openai-responses`
- Provider: `codex`
- Client: Claude Code 2.1.220
- Streaming: enabled

The current `release/v3.8.49` and `release/v3.8.50` branches still contain the affected implementation.

## Deterministic reproduction

With the current `open-sse/utils/reasoningPlaceholder.ts`:

```ts
const chunks = ["contain", " ", "1–253 bytes"];
chunks.map(stripInternalReasoningPlaceholder).join("");
// Actual:   "contain1–253 bytes"
// Expected: "contain 1–253 bytes"
```

Direct helper results:

```text
" "    => ""
"\t"   => ""
"\n"   => ""
"\n\n" => ""
"Hello, " => "Hello, "
```

The distinction matters: existing coverage exercises spaces attached to non-whitespace text, but not standalone whitespace events. Codex Responses streams can emit spaces and newlines as their own `response.output_text.delta` events.

A synthetic end-to-end stream through `createSSETransformStreamWithLogger(OPENAI_RESPONSES → CLAUDE)` reproduces the exact formatting corruption:

```text
["cleanup", "\n\n", "### Context", "\n", "ADR-R08"]
    -> "cleanup### ContextADR-R08"

["```text", "\n", "ironbox://worker/<worker-name>", "\n", "```"]
    -> "```textironbox://worker/<worker-name>```"
```

## Root cause

`stripInternalReasoningPlaceholder()` currently runs this logic for every streamed text delta:

```ts
const stripped = value.replaceAll(NON_ANTHROPIC_THINKING_PLACEHOLDER, "");
return stripped.trim() === "" ? "" : stripped;
```

For an ordinary standalone whitespace delta, `replaceAll()` changes nothing, but `stripped.trim() === ""` is true, so the function returns `""`.

The live translation path is:

```text
response.output_text.delta
  -> openaiResponsesToOpenAIResponseStream() keeps delta as OpenAI delta.content
  -> openaiToClaudeResponse() calls stripInternalReasoningPlaceholder(delta.content)
  -> whitespace-only delta becomes ""
  -> the truthy-content guard emits no Claude content_block_delta
```

The SSE parser and serializer preserve the JSON string; the loss occurs in the per-delta placeholder helper.

## Regression history

- #8081 / #8162 added response-side stripping of the internal replay sentinel.
- #8382 (`d095555d68b915f71ec81a43a168b2961f4f74d7`) and #8341 (`4323dba518a94eed2bdab9dba68fee2d5cbc91f7`) correctly added an early return when the sentinel is absent:

  ```ts
  if (!value.includes(NON_ANTHROPIC_THINKING_PLACEHOLDER)) return value;
  ```

- The later conflict-resolution commit in #8254 (`7faec9339d11ace970704a73e9c57ac33591de27`) removed that early return and replaced it with the current partial guard.
- The existing tests still pass because they cover values such as `"Hello, "` and `" Bye."`; those contain non-whitespace characters and survive the partial guard. They do not cover a delta that is exactly `" "` or `"\n"`.

## Expected behavior

- A text delta that does not contain the internal sentinel must pass through byte-for-byte, including whitespace-only strings.
- A placeholder-only delta must still be suppressed.
- A delta containing the sentinel plus real text must still have the sentinel removed according to the existing policy.
- Tool-call JSON, reasoning events, SSE framing, and non-text events must remain unchanged.

## Suggested minimal fix

Restore the sentinel-presence guard from #8341/#8382:

```ts
export function stripInternalReasoningPlaceholder(value: string): string {
  if (!value.includes(NON_ANTHROPIC_THINKING_PLACEHOLDER)) return value;

  const stripped = value.replaceAll(NON_ANTHROPIC_THINKING_PLACEHOLDER, "");
  return stripped.trim() === "" ? "" : stripped;
}
```

## Missing regression coverage

Please add both levels:

1. Utility cases asserting that `" "`, `"\t"`, `"\n"`, `"\n\n"`, and `"\r\n"` are returned byte-for-byte when the sentinel is absent.
2. A real Responses → Claude synthetic SSE test with standalone space/newline deltas that reconstructs headings, blank lines, a fenced code block, and `contain 1–253 bytes` exactly.

## Local validation

A local patch restoring the early return was verified with:

- 46/46 focused tests across the helper, Responses transformer, Responses → Claude pipeline, and adjacent reasoning/dedup suites;
- TypeScript core typecheck;
- Prettier and `git diff --check`;
- direct execution of the compiled runtime helper;
- a live Claude Code → OmniRoute → Codex → Claude smoke request, which preserved headings, blank lines, an inter-word space, and a fenced block correctly.

No provider, SSE framing, tool-call, or reasoning code needed to change.

Zartharas:
Fixes #9170

## Summary

- preserve streamed text byte-for-byte when the internal reasoning placeholder is absent
- retain existing suppression behavior for placeholder-only content
- add utility coverage for standalone spaces, tabs, LF, blank lines, and CRLF
- add a real Codex Responses to Claude streaming regression covering Markdown and word boundaries

## Root Cause

`stripInternalReasoningPlaceholder()` applied trim-based empty detection to every streamed text delta.

A delta containing only ordinary whitespace therefore became an empty string, even when it did not contain the internal replay sentinel.

## Fix

Return the original value immediately when the sentinel is absent:

`if (!value.includes(NON_ANTHROPIC_THINKING_PLACEHOLDER)) return value;`

Placeholder-only and placeholder-plus-real-text behavior remains unchanged.

## Validation

- focused utility, Responses-to-Claude, and adjacent reasoning/dedup tests: 12 passed, 0 failed
- exercised the real `createSSETransformStreamWithLogger(OPENAI_RESPONSES → CLAUDE)` pipeline
- exercised SSE buffering with single-byte wire chunks
- Prettier check passed
- `git diff --check` passed
- all three changed files independently satisfy the repository file-size policy

## Existing Release-Branch Gate State

`npm run check:file-size` reports three pre-existing violations on the untouched `release/v3.8.50` base:

- `open-sse/executors/base.ts`
- `open-sse/handlers/chatCore.ts`
- `tests/unit/chatcore-translation-paths.test.ts`

The same failures were reproduced from a detached worktree at base commit `fc35dc248f46354e80fdcdaa551e6598abcf5124`.

None of those files is modified by this PR.

## Scope

No provider behavior, SSE framing, tool-call handling, or reasoning-event handling is changed.

### DIFF
--- a/open-sse/utils/reasoningPlaceholder.ts
+++ b/open-sse/utils/reasoningPlaceholder.ts
@@ -21,6 +21,8 @@ export function isInternalReasoningPlaceholder(value: unknown): boolean {
  * real content, or streamed deltas glue together with their spaces eaten.
  */
 export function stripInternalReasoningPlaceholder(value: string): string {
+  if (!value.includes(NON_ANTHROPIC_THINKING_PLACEHOLDER)) return value;
+
   const stripped = value.replaceAll(NON_ANTHROPIC_THINKING_PLACEHOLDER, "");
   return stripped.trim() === "" ? "" : stripped;
 }
--- a/tests/unit/reasoning-placeholder-strip.test.ts
+++ b/tests/unit/reasoning-placeholder-strip.test.ts
@@ -46,8 +46,18 @@ test("a chunk with the placeholder mixed into real text strips it (trim only aff
   // only strips the string's own leading/trailing whitespace, not internal gaps.
   assert.equal(
     stripInternalReasoningPlaceholder(`foo ${NON_ANTHROPIC_THINKING_PLACEHOLDER} bar`),
-    "foo  bar",
+    "foo  bar"
   );
+});
+
+test("standalone whitespace-only chunks pass through byte-for-byte when no placeholder is present", () => {
+  for (const chunk of [" ", "\t", "\n", "\n\n", "\r\n"]) {
+    assert.equal(
+      stripInternalReasoningPlaceholder(chunk),
+      chunk,
+      `expected ${JSON.stringify(chunk)} to remain unchanged`
+    );
+  }
 });
 
 test("an empty string stays empty", () => {
--- a/tests/unit/responses-to-claude-whitespace-9170.test.ts
+++ b/tests/unit/responses-to-claude-whitespace-9170.test.ts
@@ -1,0 +1,165 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { createSSETransformStreamWithLogger } from "@omniroute/open-sse/utils/stream.ts";
+import { FORMATS } from "@omniroute/open-sse/translator/formats.ts";
+
+function sse(type: string, payload: Record<string, unknown>): string {
+  return `event: ${type}\ndata: ${JSON.stringify({ type, ...payload })}\n\n`;
+}
+
+async function runClaudeFromCodex(rawSse: string): Promise<string> {
+  const transform = createSSETransformStreamWithLogger(
+    FORMATS.OPENAI_RESPONSES,
+    FORMATS.CLAUDE,
+    "codex",
+    null,
+    null,
+    "gpt-5.5-high",
+    "conn-9170",
+    { model: "gpt-5.5-high" },
+    null,
+    null,
+    null
+  );
+
+  const writer = transform.writable.getWriter();
+  const reader = transform.readable.getReader();
+  const encoder = new TextEncoder();
+  const decoder = new TextDecoder();
+
+  const readAll = (async () => {
+    const output: string[] = [];
+
+    for (;;) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      output.push(decoder.decode(value));
+    }
+
+    return output.join("");
+  })();
+
+  // Deliberately split the wire stream into single-byte writes to exercise
+  // the same buffering and SSE reconstruction used by the live path.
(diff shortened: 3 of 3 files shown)


jackjinke:
Confirmed this fix also addresses Anthropic Messages-compatible upstream → Responses client, including Kimi Coding (`format: "claude"`). The shared helper currently drops standalone spaces/newlines after Claude → OpenAI translation; this PR's early return preserves them. Consider adding one Claude-compatible → Responses regression test to cover this direction explicitly.

Issue #9061, opened 2026-07-31
fix(api): CCR in-process Map breaks fidelity-gate recoverability guarantee


fajarhide:
## Summary

The fidelity gate skips sampling engines because their loss is "CCR-recoverable"
(`fidelityGateStep.ts:28`). CCR's store is a process-local `Map`
(`engines/ccr/index.ts:134`). Two of the four ways that map loses an entry need no restart
and happen under normal load, so the guarantee the gate stands down on is not backed by
anything durable.

## The chain

1. `fidelityGateStep.ts:28`

   ```ts
   if (engineId && getCompressionEngine(engineId)?.sampling) return true; // lossy-by-design, CCR-recoverable
   ```

   Sampling engines are exempt from fidelity verification. The exemption is justified by
   recoverability, not by the drop being small.

2. `engines/ionizer/sample.ts:170-180` drops rows and hands the full array to CCR:

   ```ts
   const stored = tryStoreBlock(serialized, opts.principalId, { contentType: "application/json", source: "ionizer" });
   if (!stored.stored) return m;
   const marker = `[ionizer: kept ${res.keptCount}/${res.totalCount} rows; full → CCR retrieve hash=${stored.hash} chars=${serialized.length}]`;
   ```

   The store call is guarded, which is correct. Nothing guards the retrieve side.

3. `engines/ccr/index.ts:134`

   ```ts
   const ccrStore = new Map<string, CcrEntry>();
   ```

4. `engines/ccr/protocolInstruction.ts:22` tells the model, in a system message:

   > call the `omniroute_ccr_retrieve` tool with that hash to get the original text back verbatim

## Four ways the entry disappears

Ordered by how ordinary they are. The first two require no restart and no operator action.

**1. Cross-principal LRU eviction (`index.ts:262`).**

```ts
function enforceGlobalBudget(bytes: number): boolean {
  while (
    (ccrStore.size >= MAX_CCR_ENTRIES || ccrTotalBytes + bytes > MAX_CCR_GLOBAL_BYTES) &&
    evictOldestMatching(() => true)
  ) { ... }
```

`evictOldestMatching(() => true)` matches every principal. Caps are `MAX_CCR_ENTRIES = 5_000`
and `MAX_CCR_GLOBAL_BYTES = 64MB` (`index.ts:65-68`). One busy API key walks the store past
5,000 entries and evicts another key's block while a live conversation still references its
[...]

PR #9191, opened 2026-08-02
fix(sse): evict a principal's own CCR blocks before another principal's (#9146)

### LINKED ISSUE #9146: fix(backend): CCR global entry cap is documented as principal-scoped but enforced globally, so one tenant's small blocks evict another's
### OmniRoute Version

3.8.50 (`371c10ea`)

### Installation Method

Built from source

### Operating System

macOS

### OS Version

Darwin 25.5.0

### Node.js Version

v22.22.3

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

`MAX_CCR_ENTRIES` is documented as the cap on a "principal-scoped" store, but `enforceGlobalBudget` applies it to `ccrStore.size` and evicts with `evictOldestMatching(() => true)`, so the block that goes is the globally oldest one. `enforcePrincipalBudget` right above it is correctly scoped with `entry.principalId === owner`.

That is reachable because nothing bounds a principal's entry *count*. The only per-principal cap is `MAX_CCR_PRINCIPAL_BYTES` (16MB). Blocks start at `DEFAULT_MIN_CHARS = 600`, so 5,000 of them is about 3MB: under a fifth of one principal's byte allowance, and enough to exhaust the shared entry budget alone. Every store after that evicts whoever has been quiet longest, because LRU keeps promoting the busy principal's own entries to the tail.

Separately, `getCcrStoreStats` counts `entries` for one principal but returns the global `maxEntries` next to it, so a caller reads its own usage against a shared cap and concludes it has room it may not have. Fixing that changes the response shape, so it wants its own change.

Independent of #9061, which is about whether a block survives at all rather than which principal loses it.

### Steps to Reproduce

1. Principal `a` stores one block and keeps it live.
2. Principal `b` stores `MAX_CCR_ENTRIES` blocks of about 600 chars, roughly 3MB, inside `b`'s own 16MB byte budget.
3. `inspectCcrBlock(hashFromStep1, "a")`.

### Expected Behavior

`a` keeps its block. It never exceeded a limit of its own.

### Actual Behavior

`inspectCcrBlock` returns `null`. `a`'s block was evicted by `b`'s traffic.

### Test Impact

Needs a new unit test

### Error Logs / Output

_No response_

### Screenshots

_No response_

### Additional Context

Suggested fix: prefer the storing principal's own oldest entries, and fall back to the current global behaviour only when that principal has nothing left to give up. The global bound stays exactly as strict, and the fallback keeps the case that needs it working: a newcomer storing into a store held entirely by others, which would otherwise never fit.

Two alternatives if the global policy is deliberate: add a per-principal entry cap so entries are bounded the way bytes are, or keep global eviction and fix the comment plus the `getCcrStoreStats` pairing so nothing claims the cap is per-principal. Only the first also stops the cross-tenant eviction.

Found while reading the store for #9061. Principal scoping elsewhere in CCR is correct and has been fixed deliberately before (#5649, #7883), which is why this reads as an oversight rather than a policy choice.

### Validation Plan

Unit test: one principal stores a block, another fills the entry cap with minimum-size blocks, the first principal's block must survive.

fajarhide:
## Summary

- `enforceGlobalBudget` evicted with `evictOldestMatching(() => true)`, so a store by any principal took the globally oldest block. Nothing caps a principal's entry count, only its bytes, so one principal can fill the shared 5,000-entry budget with about 3MB of minimum-size blocks and then evict whoever has been quiet longest.
- It now takes the storing principal's own LRU entries first, and falls back to any principal only when it has none left. The global bound is exactly as strict as before.
- Also fixes the `MAX_CCR_ENTRIES` comment, which claimed a scope the code never implemented.

## Related Issues

- Closes #9146

## Validation

- [x] Focused tests for the change: `node --import tsx/esm --test tests/unit/compression/ccr-eviction-scope-9146.test.ts`
- [x] `npm run lint`
- [x] SonarQube PR analysis is green or any remaining issues are explicitly documented below

SonarQube is unchecked because it has not reported yet at open time. It runs in CI on this PR.

Failing before the fix:

```
not ok 1 - keeps a quiet principal's block when a busy principal fills the entry cap
  error: "a principal that stored one block must not lose it to another principal's traffic"
# tests 4  # pass 3  # fail 1
```

Passing after: `# tests 4  # pass 4  # fail 0`. The other three pass on both sides by design, since they guard what the fix must not break rather than reproduce the bug. 103 CCR/ionizer tests green, `typecheck:core` clean.

## Tests Added Or Updated

- `tests/unit/compression/ccr-eviction-scope-9146.test.ts` (new, 4 tests)

## Coverage Notes

One production file, `open-sse/services/compression/engines/ccr/index.ts`, +31/−8. The new tests cover the eviction path directly: the quiet principal's block surviving, the entry cap still holding, the busy principal's byte usage, and the fallback for a newcomer.

## Reviewer Notes

- The fallback to any principal is deliberate and has its own test. Without it a full cache would permanently lock out every principal that arrives late.
- Cut from `release/v3.8.50` (`371c10ea`) with nothing from #9061 in it.
- Left alone: `getCcrStoreStats` returns a per-principal `entries` next to the global `maxEntries`, so a caller reads its own usage against a shared cap. Changing that changes the response shape and there are dashboard consumers, so it needs its own PR. Noted in #9146.

### DIFF
--- a/open-sse/services/compression/engines/ccr/index.ts
+++ b/open-sse/services/compression/engines/ccr/index.ts
@@ -61,7 +61,12 @@ const RETRIEVAL_THRESHOLD = 3;
  * ramp (only the >= threshold cliff remains — the legacy binary behavior).
  */
 const RETRIEVAL_RAMP_FACTOR_DEFAULT = 2;
-/** Maximum number of entries in the principal-scoped, LRU-ordered store. */
+/**
+ * Maximum number of entries in the LRU-ordered store, across every principal. The store
+ * is keyed per principal, but this cap is not: the only per-principal cap is
+ * `MAX_CCR_PRINCIPAL_BYTES`. Eviction under this cap takes the storing principal's own
+ * blocks first (see `enforceGlobalBudget`).
+ */
 export const MAX_CCR_ENTRIES = 5_000;
 export const MAX_CCR_BLOCK_BYTES = 2 * 1024 * 1024;
 export const MAX_CCR_PRINCIPAL_BYTES = 16 * 1024 * 1024;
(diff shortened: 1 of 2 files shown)


fajarhide:
@diegosouzapw this one needs a click from you before CI can report. Because the head is a fork branch, `Quality Gates`, `semgrep` and `DAST smoke (PR)` are all sitting at `action_required`, waiting for "Approve and run workflows". Until then the PR shows `UNSTABLE` with only the semgrep cloud app having reported. Same situation on #9198.

What I ran locally on this branch in the meantime, against merge base `371c10ea`:

- `tests/unit/compression/ccr-eviction-scope-9146.test.ts`, verified failing before the fix (3/4) and passing after (4/4)
- 103 CCR and ionizer tests
- `typecheck:core`, `check:cycles`, `eslint` on both changed files

The diff is one production file and one test file, +110/-8, and none of the eight commits that landed on `release/v3.8.50` after my base touch `ccr/index.ts`, so it still merges clean.

No rush from my side. Flagging it only because the pending state looks like a failure at a glance and it is just the fork approval.

PR #9198, opened 2026-08-02
fix(sse): back the CCR block store with a durable tier (#9061)

### LINKED ISSUE #9061: fix(api): CCR in-process Map breaks fidelity-gate recoverability guarantee

fajarhide:
## Summary

- CCR keeps its blocks in a process-local `Map` (`engines/ccr/index.ts:134`), so one goes whenever the entry is evicted, the 24h TTL passes, the process restarts, or the MCP retrieve lands on another instance. `fidelityGateStep.ts:28` waives fidelity checks for sampling engines because their loss is "CCR-recoverable", and `protocolInstruction.ts:22` tells the model it can get the block back verbatim. Neither is backed by anything durable.
- The Map stays the hot cache, untouched. Its LRU ordering, byte accounting and principal budgets are load-bearing across ~15 call sites. `tryStoreBlock` writes through to SQLite after `ccrStore.set`, and `retrieveBlock` reads back where `getActiveEntry` returns null.
- Migration 025 moved call-log payloads out of SQLite on purpose, behind a 512KB cap, a cloud check and an opt-in gate, so this follows the same path: blocks over 512KB stay memory-only (2MB across 5,000 entries was a 10GB ceiling), nothing is written on a cloud runtime, and `COMPRESSION_CCR_DURABLE_STORE=false` turns it off.
- Writes go through `setImmediate`, capped at 1,000 pending. A synchronous persist measures 0.032ms at the 600-char minimum block but 1.77ms at 500KB, past the 1ms this engine declares in its metadata.

## Related Issues

- Closes #9061

## Validation

- [x] Focused tests for the change: `node --import tsx/esm --test tests/unit/ccr-durable-store-9061.test.ts`
- [x] `npm run lint`

SonarQube is unchecked because it has not reported yet at open time. It runs in CI on this PR.

Failing before the fix, on the restart case:

```
block must come back from the durable tier after the in-memory map is gone
not ok 2 - CCR engine survives losing its in-memory map (#9061)
# tests 8  # pass 7  # fail 1
```

Both guards were checked the same way, at 8/10 without them on the oversized-block and disabled-flag cases. All 10 pass now.

Also run locally against this branch: `npm run test:vitest` (291/291), `npm run lint`, `npm run typecheck:core`, `npm run check:cycles`, `npm run check:docs-sync`, `npm run check:env-doc-sync`, `check-migration-numbering`, `check:any-budget:t11`, `check:tracked-artifacts`.

On `npm run test:unit`: the full suite is flaky on this machine. I ran it on this branch and on the merge base and compared, and the base failed more (19) than the branch (17), with 13 in common and every failure in a timing or timeout assertion. The four that failed only on the branch are `chat-cooldown-aware-retry`, `cli-health-monitoring-route`, `model-test-runner` and `modelsDevSync-extended`; all four pass standalone on this branch (37/37) and none of them touch CCR or the new db module. Flagging it rather than claiming a green full suite.

## Tests Added Or Updated

- `tests/unit/ccr-durable-store-9061.test.ts` (new, 10 tests)

## Coverage Notes

Production files changed: `src/lib/db/ccrBlocks.ts` (new), `src/lib/db/cleanup.ts` (one function plus its registration in `runAutoCleanup`), `src/lib/localDb.ts` (re-export line only), and `open-sse/services/compression/engines/ccr/index.ts`. The new test file covers the db module directly and the engine's store, rehydrate, principal scoping, delete, size-cap and disabled-flag paths.

## Reviewer Notes

- **One decision is yours.** `COMPRESSION_CCR_DURABLE_STORE` defaults to on, because the protocol instruction already promises the model a verbatim block and defaulting to off would leave that promise hollow for anyone who never reads the file. If you would rather it be opt-in, it is a one-character change.
- New migration `134_ccr_blocks.sql`, and a new env var documented in `.env.example` and `ENVIRONMENT.md`.
- Persists are dropped under backpressure, which is safe because the block is still in the map and the client re-stores it under the same hash. Deletes are never dropped, or a deleted block comes back. Expiry drops the durable row; capacity eviction deliberately does not, which is the point of the second tier.
- Every durable call is best-effort. Without a usable database, compression preview and unit tests behave exactly as they do today.
- Known limit: the multi-instance vector only closes when the instances share the database file.
- The regression test simulates a restart by importing a second instance of the engine module with a query string, so its map starts empty while the database persists. That is not a real process restart, and the test name should not be read as one.
- @pacocartones verified this issue independently on the same commit, and has now verified the two vectors my regression test cannot reach, against head `55e8e7015` with two OS processes sharing one `storage.sqlite`. Both pass: a retrieve on an instance that never ran the compression returns the block byte-identical, and so does a retrieve after the writer is `kill -9`ed with no cleanup. The negative controls behaved as designed too, including the flag genuinely keeping blocks memory-only. Their run was Windows-only and drove `tryStoreBlock`/`handleCcrRetrieve` directly rather than through HTTP and the MCP transport.
- One boundary their run made explicit, worth stating here rather than leaving in the thread: a writer killed *before* its deferred `setImmediate` persist has run loses that block. The client re-stores it under the same hash on the next request, so it self-heals, but the durable copy is not synchronous with the store call.


pacocartones:
Thanks for the credit — no competing branch on my side. My #9061 comment was analysis only; I never started an implementation, so there is nothing to close in favour of.

The design matches what I would have proposed: keep the Map as the hot cache untouched, write through to SQLite, defer persists off the request path, and follow migration 025's precedent with the 512KB cap, cloud check and opt-out flag. Defaulting `COMPRESSION_CCR_DURABLE_STORE` to on also reads right — the protocol instruction already promises verbatim retrieval, and an opt-in default would leave that promise unbacked for anyone who never opens the file.

Happy to re-verify the loss-vector cases against the final commit if that helps the review.

fajarhide:
Good to know @pacocartones , and thanks for clearing that up.

Yes please, and there are two cases my tests genuinely do not cover, so those are the useful ones:

1. **Multi-instance.** Two processes against the same database file, compress on one and call `omniroute_ccr_retrieve` on the other. That is the only loss vector the fix closes conditionally, and I could not exercise it locally.
2. **A real process restart.** My regression imports a second instance of the engine module with a query string so its map starts empty, which exercises the same path but is not a restart. Killing a running server and retrieving afterwards would settle it.

The eviction and TTL vectors are covered by unit tests, so those need less attention.

Head is `55e8e7015`.

pacocartones:
Hi @fajarhide — following up on your request to verify the two loss vectors your regression test can't reach (it re-imports the engine module in the same process, so it never crosses a real process boundary). I ran both against your PR head with separate OS processes sharing one `storage.sqlite`. **Both vectors pass.**

## Setup

- Commit: `55e8e7015dafa6aa095e2e70b691aea876af5272` (`fix(sse): bound and gate the CCR durable tier (#9061)`), fetched from `pull/9198/head`
- OS: Windows 11 (MINGW64, Git Bash), Node v24.18.1, better-sqlite3 13.0.2
- Method: a targeted two-process harness that imports **your** modules directly — no full server. The writer process calls `tryStoreBlock()` (the same entry point `maybeCcrReplace` uses) plus `flushCcrDurableWrites()`; the retriever process calls `handleCcrRetrieve({ hash }, callerId)` — the exact handler behind the `omniroute_ccr_retrieve` MCP tool. Both point at the same `DATA_DIR`/`storage.sqlite` (migration `134_ccr_blocks` applied by your migration runner on first open, journal mode WAL).
- The stored block was deliberately hostile to byte-mangling: multi-line, CRLF, tabs, accented chars, emoji, and an embedded NUL (1,584 chars). Identity is checked by strict string equality **and** SHA-256.
- In every retrieve, `inspectCcrBlock()` confirmed a hot-cache **MISS** first, so the content provably came from SQLite, not the Map.

## Vector 1 — multi-instance (writer alive, retriever in a second process)

Proc A stored and flushed, stayed alive (`kill -0` confirmed immediately before Proc B ran). Proc B, a separate OS process against the same database file:

```
[B] in-memory map lookup before retrieve: MISS (expected)
[B] handleCcrRetrieve returned 1584 chars
[B] sha256 expected=e6bb905fb7695963494b8c6b3d1f18b8f4b8763d35b759cd6143cc1148d7071f
[B] sha256 actual  =e6bb905fb7695963494b8c6b3d1f18b8f4b8763d35b759cd6143cc1148d7071f
[B] BYTE_IDENTICAL=true
```

**Verdict: PASS.** A retrieve landing on an instance that never ran the compression gets the block back verbatim.

## Vector 2 — real restart (SIGKILL, no graceful shutdown)

Proc A stored, flushed, then was killed with `kill -9` — `wait` reported status **137**, so no exit handlers, no DB close, no clean checkpoint. Proc B, started after the kill:

```
[orchestrator] Proc A wait status: 137 (128+9=137 => died by SIGKILL, no cleanup ran)
[B] in-memory map lookup before retrieve: MISS (expected)
[B] handleCcrRetrieve returned 1584 chars
[B] BYTE_IDENTICAL=true
```

**Verdict: PASS.** The durable tier survives a hard kill.

## Controls (negative cases, all behaved as designed)

- **`COMPRESSION_CCR_DURABLE_STORE=false`**: writer stored, flushed, exited cleanly; fresh-process retrieve → `CCR block not found for hash=…`. The flag genuinely keeps blocks memory-only — this is the "conditional" in action.
- **Principal isolation across processes**: same DB, retrieve as a different principal → `not found`; retrieve as the storing principal → byte-identical hit. The IDOR boundary holds through the durable tier.
- **Defer window**: a writer killed *before* its deferred `setImmediate` persist could run loses the block (`not found` on retrieve). That matches your own comment in `index.ts` ("only a crash inside that tick loses the durable copy, and the client's next request re-stores it under the same hash") — documenting it here so the boundary is explicit, not because it's a defect.
- Your own regression file also passes locally: `tests/unit/ccr-durable-store-9061.test.ts` → **10/10 pass** (node:test, tsx).

## Caveats

- Targeted script, not a full server: I drove `tryStoreBlock`/`handleCcrRetrieve` directly rather than going through HTTP → compression pipeline → MCP transport. The storage/retrieve logic under test is identical; what I did **not** exercise is the wiring around it (MCP session auth populating `callerId`, the SSE path).
- Single block per scenario (plus one multi-retrieve for the principal check). No burst testing of the 1,000-pending-write backpressure cap across processes.
- Windows-only run. WAL semantics here held up, but a Linux/macOS run would broaden coverage.
- The 512 KB cap and TTL behavior were not re-tested — your unit tests cover them.

Bottom line: the fix closes the multi-instance vector conditionally (flag on, non-cloud runtime, block ≤ 512 KB, deferred write already drained) and the restart vector on the same conditions — and in this environment those conditions behaved exactly as your code comments describe. LGTM from the durability angle.

fajarhide:
@pacocartones this is more than I asked for, thank you. Two OS processes on one `storage.sqlite`, a hostile block with CRLF and an embedded NUL, and a `MISS` on the hot cache confirmed before every retrieve so the content provably came from disk. That settles both vectors.

The defer-window finding is the useful part. My code comment claims the block is lost only if the crash lands inside that tick, and your `kill -9` before the persist drained is the first actual demonstration of it. I have pulled both that boundary and your verification summary into the PR body, so it reaches @diegosouzapw without him reading the thread.

Noted on the caveats: Windows-only, and you drove `tryStoreBlock`/`handleCcrRetrieve` directly rather than through HTTP and the MCP transport. Neither changes the storage conclusion, and I would rather they sit in the record than be implied as covered.

PR #7329, opened 2026-07-15
Add cliproxy provider exposure controls and manifest injection

KooshaPari:

Issue #7226, opened 2026-07-15
fix: dast-smoke 'Build CLI bundle' hangs recurrently on GitHub-hosted runners (24-29min, kills the advisory check)


diegosouzapw:
## Symptom
The `dast-smoke` workflow's **Build CLI bundle** step (`npm run build:cli`) hangs on GitHub-hosted runners: observed 10m46s, 24m, and 29m+ (vs 6m22s on a healthy run), across unrelated branches on 2026-07-14/15 (runs 29298270061 ×3 attempts, 29314862473, 29386159373 — queue drafts #7184/#7221 died on it).

## Impact
- The advisory check reports red on unrelated PRs (noise).
- It dequeued every Mergify merge-queue attempt until #7225 made the queue tolerate it.

## Mitigations already in place
- #7139 raised the job timeout 12→25min (helps variance, not hangs).
- #7225 makes the merge queue tolerate exactly this check.

## Ask
Root-cause the hang: capture `npm run build:cli` verbose logs on a GH-hosted run (network stall? esbuild? postinstall?), consider a step-level timeout + retry on the build, or move the job to the .113 pool.

diegosouzapw:
Additional impact data: the dast-smoke flake also interacts badly with the Mergify queue's fail-fast — speculative drafts #7309/#7320/#7330 were all dequeued at the ~10-11min mark when dast failed on the draft, even though the queue's `merge_conditions` tolerate exactly that check (#7225's or-condition applies to the final evaluation; Mergify fails the attempt on any failing draft check first). Draft #7320 even finished 100% green (dast recovered) but the attempt had already been failed.

Practical consequence: while this flake persists, green PRs may need `gh pr merge` manually after a queue dequeue. Fixing the underlying build hang (this issue) resolves the queue behavior too; alternatives if it drags: mark dast-smoke `if: github.head_ref does not start with mergify/merge-queue/` (skip on drafts) or move the job to the .113 pool.

PR #9233, opened 2026-08-02
fix(ci): merge-queue tolerance for Build (advisory), drops paid-tier batching

wgordon17:
## Summary
- Restores #7220's removal of `batch_size`/`batch_max_wait_time` (paid-tier Mergify feature, breaks on the free plan) and #7225's auto-enqueue mechanism (`merge_protections_settings.auto_merge_conditions`), both silently reverted two weeks ago by #7329 (an unrelated cliproxy PR that touched `.mergify.yml` from a stale branch)
- Retargets the merge-queue's advisory-check exception from `dast-smoke` to `Build (advisory)`: evidence review showed `dast-smoke`'s hang (#7226) has been dormant for weeks (0/30 recent runs, 2 failures all-time, none since 2026-07-13), while `Build (advisory)` (added to quality.yml 2026-07-27) has a 100% failure rate on every sampled PR since, confirmed via job logs to be the same class of runner hang (dies mid-"Creating an optimized production build", never a real compile error) — just in a check the old tolerance never covered
- No PR has used the `queue` label since #7329 landed, so both the original regression and the stale targeting went unnoticed until now

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -17,6 +17,13 @@
 #   • Fallback path if Mergify misbehaves or the OSS plan changes: the manual
 #     merge-train runbook (docs/ops/MERGE_TRAIN.md) — remove labels, proceed by hand.
 
+# Auto-enqueue (current Mergify model, 2026): auto_merge_conditions in
+# merge_protections_settings — the rules-based queue action / autoqueue path is
+# deprecated (EOL 2026-07-16). The owner-applied `queue` label IS the approval.
+merge_protections_settings:
+  auto_merge_conditions:
+    - label = queue
+
 queue_rules:
   - name: release
     # Any current or future release branch — the reason GitHub's native queue was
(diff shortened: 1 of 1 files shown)


diegosouzapw:
Merged — thanks @wgordon17.

Your diagnosis matched an independent measurement I ran while triaging the queue today: of 149 merge-ready candidates, **21 were failing on `Build (advisory)` and nothing else** — Fast Quality Gates, all four Unit Tests shards, Vitest and Merge integrity were green on every one of them. The job log confirms the cause you documented (run 30768479406): Turbopack enters "Creating an optimized production build", runs ~7 minutes, then `The runner has received a shutdown signal` → `The operation was canceled`. Never a compile error.

Also appreciated that you dropped the stale `dast-smoke` exception instead of carrying it forward, and kept the anti-fail-open shape (any *other* failure still blocks). This unblocks the rest of the queue.

Issue #8928, opened 2026-07-29
fix: STREAM_EARLY_EOF does not evict session affinity, pinning retries to failed account


cryptiklemur:
### OmniRoute version

`3.8.49-pre-a899236`

### Environment

- Docker on Linux
- Provider: Claude with multiple active accounts
- Client: Claude Code
- Request: Anthropic-compatible streaming request using the qualified model `claude/claude-opus-5`
- Session affinity TTL: 24 hours

### Description

When a session-affinity-pinned Claude connection ends its SSE stream before producing any non-ping event, OmniRoute returns:

```json
{
  "error": {
    "message": "Stream ended before producing a non-ping SSE event",
    "type": "stream_early_eof",
    "code": "STREAM_EARLY_EOF"
  }
}
```

The failed connection remains active and the matching affinity record is not evicted. Every client retry selects the same connection and refreshes the sliding affinity expiry. The conversation remains stuck even though another eligible Claude account can serve the same request.

Qualified provider requests show `comboName: null`, so combo fallback does not protect this path.

### Steps to reproduce

1. Configure at least two active Claude accounts.
2. Enable generic session affinity with a nonzero TTL.
3. Send a streaming request through a qualified Claude model.
4. Have the pinned connection terminate after pings but before the first non-ping SSE event.
5. Retry the identical conversation.

### Actual behavior

- The first failure returns HTTP 502 with `STREAM_EARLY_EOF`.
- The connection remains marked active, with no rate limit recorded.
- Each retry reuses the same affinity-pinned connection.
- Each retry refreshes the affinity record's `lastUsedAt` and `expiresAt`.
- No other eligible account is tried.

### Expected behavior

A terminal stream failure before any useful SSE event should evict the affinity record only if it still points at the failed connection. The next retry should be allowed to select another eligible account.

### Deterministic red/green evidence

I minimized the original request to its first user message, which is the affinity fingerprint source, and capped output at 16 tokens.

With the session pinned to the failing connection:

```json
{"verdict":"RED","status":502,"errorType":"stream_early_eof","errorCode":"STREAM_EARLY_EOF","nonPingEvent":false,"elapsedMs":2012}
```

After deleting only that session's matching affinity record, the exact same replay succeeded through another eligible account:

```json
{"verdict":"GREEN","status":200,"errorType":null,"errorCode":null,"nonPingEvent":true,"elapsedMs":7234}
```

Repeating the replay after the new affinity pin remained green:

```json
[...]

PR #9251, opened 2026-08-02
fix(resilience): count STREAM_EARLY_EOF as a provider failure in combo routing

TechNickAI:
## What happened

During a 7-day window on our production router we recorded 311 `STREAM_EARLY_EOF`
events (HTTP 502, upstream returned 200 then closed the SSE stream with no
non-ping event). 302 hit one model, 265 fell inside Anthropic's published
incident window. The provider circuit breaker sat at `CLOSED` /
`failure_count=0` the entire time — every request kept dispatching to the
failing provider instead of shedding to the next combo target.

The single-model path (`shouldTripProviderBreakerForResult` in
`chatPredicates.ts`) already counts these correctly. The combo path didn't.

## Why combo missed it

`isStreamReadinessFailureErrorBody()` matches both `STREAM_READINESS_TIMEOUT`
and `STREAM_EARLY_EOF`, and `shouldRecordProviderBreakerFailure` exempts
anything that predicate flags. So the breaker never saw an early EOF.

The two codes are genuinely different events:

- `STREAM_READINESS_TIMEOUT` is a pre-flight liveness probe on a connection we
  haven't committed to. Failing it means "this connection looks stale."
- `STREAM_EARLY_EOF` means the provider accepted the request (HTTP 200), opened
  the stream, and then hung up — it took the request and failed to serve it.

The transient-retry and round-robin semaphore-cooldown paths in `combo.ts` want
identical treatment for both codes, so `isStreamReadinessFailureErrorBody`
still matches both. Only the breaker needs to tell them apart.

## The fix

A narrow optional-argument addition:

- New predicate `isStreamEarlyEofErrorBody()` in `comboPredicates.ts` that
  matches only `STREAM_EARLY_EOF`.
- `shouldRecordProviderBreakerFailure` gains an optional `breakOnEarlyEof`
  parameter (default `false`). When `true`, an early EOF is no longer exempt
  from the breaker even though it still matches the shared classifier.
- The call site in `combo.ts` passes `true` for the breaker check.

Omitting the new argument reproduces the previous behaviour exactly — no
call-site besides the breaker predicate is affected, and the existing
exclusions for client aborts and plain 429s are untouched.

## Test

`tests/unit/stream-early-eof-breaker.test.ts` — 11 assertions covering:

- `STREAM_EARLY_EOF` increments `failureCount` toward the breaker threshold
- `STREAM_READINESS_TIMEOUT` remains exempt (no regression)
- Client abort (499) and plain 429 remain exempt
- Breaker opens after reaching threshold on early EOFs
- Omitting the optional argument preserves previous behaviour

All 42 combo breaker tests pass. Typecheck, prettier, eslint clean.

## 24-hour production soak

Deployed to our production router (single host, ~15K requests/day) at
2026-08-01T20:56:48Z. After 24 hours:

- **15,283 requests served**, 99.27% success rate (15,171/15,283)
- **Zero `STREAM_EARLY_EOF` events** — no upstream outage occurred during the
  soak, so the fix's failure path was unexercised in production. The unit tests
  are the proof; the soak proves nothing regressed around the change.
- **Zero breaker transitions** — all three provider breakers remained CLOSED
  with `failure_count=0` throughout
- **Zero migrations, zero crashes, zero OOM**
- Backends served: claude-opus-5, gpt-5.6-sol-xhigh, gpt-5.6-terra, grok-4.5,
  plus combo fallbacks to codex, openrouter, gemini, cohere
- One transient disk I/O stall at ~20:14Z (EBS-level, not code-related) caused
  a watchdog-initiated restart; service recovered in 46s with no data loss

## Relation to #8928

This is the combo-path counterpart to the affinity-eviction symptom described
in #8928. Both trace back to `STREAM_EARLY_EOF` being classified alongside
`STREAM_READINESS_TIMEOUT` when the two events carry different operational
meaning. This PR fixes the breaker blindness; affinity eviction is a separate
code path that this change does not touch.


diegosouzapw:

|---|---|---|

diegosouzapw:
Merged — thanks @TechNickAI.

The production evidence in the description is what made this easy to prioritise: 311 `STREAM_EARLY_EOF` events over 7 days, 302 on one model, 265 inside a published Anthropic incident window, with the breaker sitting at `CLOSED`/`failure_count=0` the whole time. Tracing it to `isStreamReadinessFailureErrorBody()` matching both codes while only the single-model path counted them correctly is a precise diagnosis.

One note on what we changed: your branch was re-synced against the release because `stryker.conf.json` conflicted with #9022, which registered `stream-readiness.test.ts` in the same alphabetical slot of `tap.testFiles`. Both entries were kept — the two tests cover the two codes your PR is specifically distinguishing, so dropping either would have undone the point. The merge commit is authored to you with a `Co-authored-by` for the maintainer; nothing else in the diff was touched.

PR #9259, opened 2026-08-02
fix(resilience): enforce RPM with rolling leases

### LINKED ISSUE #8733: feat(resilience): replace fixed-window RPM reservoirs with rolling leases and atomic multi-scope admission
## Summary

OmniRoute's configured RPM protection should use rolling leases rather than Bottleneck's fixed-window reservoir reset.

For a limit of 20 requests per minute, each dispatched request should lease one slot for 60 seconds. The lease should return when that request's 60-second age expires. This enforces no more than 20 dispatches in any trailing 60-second window and avoids a burst of 20 requests at every fixed-window boundary.

The implementation should also support cumulative global and provider/account scopes. A request must hold every applicable RPM lease before upstream dispatch.

Related issues:

- #198 and #3152 introduced configurable provider/per-connection rate limits.
- #7778 proposes the global/provider/account hierarchy for concurrency and queues. This issue is specifically about RPM admission semantics and can be coordinated with that work.

## Current behavior

`open-sse/services/rateLimitManager.ts` configures Bottleneck with:

```ts
reservoir: requestsPerMinute,
reservoirRefreshAmount: requestsPerMinute,
reservoirRefreshInterval: 60 * 1000,
```

That is a fixed-window counter. It consumes the full reservoir and then resets the entire budget at the refresh boundary. It is not a rolling lease window.

The current limiter scope also does not match the name of the global setting:

- `requestQueue.requestsPerMinute` is applied to independently keyed connection limiters, not one process-wide pool.
- `rateLimitOverrides.rpm` is keyed per provider connection.
- The existing `SlidingWindowLimiter` is only used for opt-in provider defaults and is keyed by `provider:connectionId`.

## Observed failure

Fresh production logs showed the consequences on an Openference connection configured for 20 RPM:

1. The limiter consumed its final token. The trace showed `reservoir=0` after the last request completed.
2. A subsequent job entered with `queued=1`, `running=0`, `executing=0`, and `reservoir=0`.
3. The job waited 60 seconds and returned a local queue-timeout 503 without any upstream response.
4. Combo retries left the expired Bottleneck jobs queued, growing the queue to `queued=3` with nothing running or executing.
5. The watchdog reset the stranded queue, producing a local 502.

The trace also showed a limiter instantiated with the global 21-token value even though the connection override was 20 RPM, because connection overrides are loaded after startup reconciliation can create a limiter.

### Rolling leases

For a configured limit of `N` requests per `windowMs`:

- admission records a lease at dispatch time;
- the lease expires at `dispatchTime + windowMs`;
- the oldest lease is the next one to return capacity;
- blocked requests wait until the next lease expires;
- blocked attempts do not consume a lease;
- cancellation before admission does not consume a lease;
- provider execution duration does not extend or shorten the RPM lease.

This is a strict trailing-window limit, not a fixed-window reset and not a bursty token bucket.

### Cumulative scopes

The preferred model is one atomic multi-scope admission operation:

```text
global RPM lease:   always required
provider/account:   required when that scope has an RPM override
account/model:      required if a future narrower RPM scope is configured
```

If the global budget is 21 and Openference is configured for 20 RPM, a request must acquire one lease from both scopes. If the provider has no override, it acquires only the global lease.

Acquisition must be atomic. The implementation must not reserve a global lease and then wait indefinitely for a provider lease, because that would strand global capacity. It should check all applicable scopes and commit all leases together only when every scope is available.

The exact provider scope needs to be explicit. Existing overrides are per connection, so a provider/account key may be appropriate when accounts have independent quotas. A provider-wide key is appropriate when multiple accounts share one provider quota.

## Implementation direction

The existing `SlidingWindowLimiter` in `open-sse/services/slidingWindowLimiter.ts` already implements timestamp-based trailing-window admission and oldest-hit retry timing. It could be generalized into an atomic multi-scope lease gate rather than using Bottleneck's fixed-window reservoir for RPM enforcement.

Concurrency and RPM should remain separate controls:

- concurrency limits control simultaneous upstream executions;
- RPM leases control dispatch frequency over the trailing time window;
- queue admission and cancellation must not leave orphaned jobs in either layer.

The implementation should also ensure that configured connection overrides are loaded before any limiter is instantiated.

## Acceptance criteria

- No more than `N` upstream dispatches occur in any trailing `windowMs` for a configured scope.
- Twenty requests cannot be followed by another burst of twenty at a fixed-window boundary.
- A global RPM limit applies across providers, connections, models, and retries.
- A configured provider/account RPM limit composes with the global limit.
- Requests without a provider override still use the global pool.
- A rejected provider-scope admission does not consume a global lease.
- Client cancellation, queue timeout, fallback, and upstream failure do not leak or double-count leases.
- The oldest lease expiration wakes waiting work without requiring a watchdog reset.
- Existing per-connection RPM configuration remains compatible or has an explicit migration path.
- Tests cover burst traffic, boundary traffic, global-plus-provider composition, cancellation, retry/fallback, and multiple provider connections.
- Observability identifies which RPM scope blocked a request and separates RPM wait from concurrency wait and upstream execution.

## Non-goals

- This issue does not replace the hierarchical concurrency work in #7778.
- This issue does not change provider 429 classification or retry policy.
- This issue does not make every provider's quota assumption automatic; provider/account RPM overrides remain explicit.

excessivechaos:
Fixes #8733.

## What changed

- Replaces Bottleneck's fixed-window RPM reservoir with rolling 60-second leases.
- Acquires the process-wide global scope and provider/account scopes atomically.
- Keeps Bottleneck for per-connection concurrency, pacing, and queue observability.
- Releases leases on pre-dispatch abort, queue expiry, admission rejection, limiter eviction, and connection disable.
- Retains leases after dispatch, including provider failures and combo fallback.
- Preserves model-scoped limiter and learned-header behavior for providers that use it.
- Keeps queue wait separate from provider execution time; an active provider call is not killed by the queue budget.
- Preserves queued peers when one request is cancelled before dispatch.
- Distinguishes upstream cooldown waits in queue-timeout messages without changing the existing code/status or routing behavior.

## Scope

The global gate is process-local and shared across enabled connections. Distributed multi-instance quota coordination is intentionally out of scope because OmniRoute has no shared rate-limit store. The PR does not change provider error classification, retry policy, or hierarchical concurrency behavior.

## Validation

- 33/33 focused limiter and queue-timeout tests pass.
- Core TypeScript typecheck passes.
- Focused ESLint passes.
- File-size and changelog-integrity checks pass.

---
⚠️ base-red inherited: #9298 (migration version collision, pre-existing on release/v3.8.50)


diegosouzapw:
Resolved merge conflicts and retargeted to release/v3.8.50 via #9604 (same changes, conflict-free branch). Original author credited.

PR #9281, opened 2026-08-03
fix(mcp): stop DB init logging from corrupting the stdio JSON-RPC stream

Sam280903:
## Summary
- `omniroute --mcp` spawns `open-sse/mcp-server/server.ts` as a child process with `stdio: "inherit"`. Its first `getDbInstance()` call (a side effect of `createMcpServer()`'s tool registration reading compression settings) lazily runs DB init, which logs via plain `console.log` — straight onto **stdout**, the stream reserved exclusively for JSON-RPC messages.
- MCP clients (e.g. Claude Desktop) then fail to parse the corrupted first line: `Unexpected token 'D', "[DB] Changi"... is not valid JSON`.
- A redirect placed *inside* `server.ts` itself is too late: ES module static imports are hoisted and fully evaluate before any code in the importing module's own functions runs, so earlier import-time side effects have already logged to the real stdout by the time a redirect placed in a function body executes.
- Fixed by preloading `bin/mcpStdioConsoleGuard.mjs` via `node --import` in the spawn call (`bin/mcp-server.mjs`) — the only point early enough to guarantee nothing leaks, since it runs before the MCP entry's module graph is even requested.

## Test plan
- [x] TDD: added `tests/unit/mcp-stdio-json-purity.test.ts` — spawns the real `omniroute --mcp` process, sends an `initialize` request, and asserts every stdout line is valid JSON while DB init logging is still observable on stderr
- [x] Confirmed the test fails with the exact reported error message (`Unexpected token 'D', "[DB] Changi"... is not valid JSON`) against the pre-fix code, and passes after the fix
- [x] Confirmed pre-existing related tests (`tests/unit/mcp-server-entry.test.ts`, `tests/unit/bin-omniroute-mcp.test.ts`) show no new regressions (one pre-existing Windows path-separator failure in `mcp-server-entry.test.ts` reproduces identically on the unmodified base checkout)

### DIFF
--- a/bin/mcp-server.mjs
+++ b/bin/mcp-server.mjs
@@ -3,7 +3,7 @@
 import { spawn } from "node:child_process";
 import { existsSync } from "node:fs";
 import { dirname, join } from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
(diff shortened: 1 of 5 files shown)


diegosouzapw:
## Review: PR #9281 — fix(mcp): stop DB init logging from corrupting stdio JSON-RPC stream

**Verdict: merge-ready** (4/5 stars)

Redirects ~15 console.log/warn from getDbInstance() to stderr in MCP stdio mode. Critical fix for `omniroute --mcp` mode where stdout must be pure JSON-RPC. TDD confirmed.

Ready to merge.

Issue #8249, opened 2026-07-23
feat(backend): propagate caller session key/tag into call_logs for per-session cost attribution


andrea-kingautomation:
### OmniRoute Version
3.8.48 (npm global)

### Use case
We run a multi-agent system where a main agent session spawns child/subagent sessions, all calling through the same OmniRoute instance and API key. We want to answer "what did THIS child session cost" from the router's own records.

### Problem
Today there is no way to link a `call_logs` row to the session that made the request:

- `call_logs` has no session/caller column (audited a live 81,595-row table, 7-day retention)
- `session_model_history` and `routing_decisions` exist in the schema but are empty in practice
- All requests through one API key are indistinguishable, so router-side cost attribution for subagent/child sessions is impossible; we can only approximate by joining timestamps against client-side transcripts

### Ask
Accept an opaque caller tag on the request (a header such as `X-OmniRoute-Session`, or reuse the OpenAI-compatible `user` field) and persist it as a column in `call_logs`. No semantics needed router-side: pass-through and store is enough. Bonus: aggregate cost per tag in the dashboard/stats endpoints.

(Searched open and closed issues for session key / call_logs / attribution before filing; #7388 is about WebSocket log rows and unrelated.)

diegosouzapw:
Thanks for the detailed audit — verified against the code and your findings are accurate.

- `call_logs` (src/lib/db/migrations/025_call_logs_summary_storage.sql) indeed has no session/caller column.
- `session_model_history` is written only by the context-relay combo strategy (src/lib/db/contextHandoffs.ts), and `routing_decisions` only by the A2A protocol logger (src/lib/a2a/routingLogger.ts) — neither is a general per-request session tag for normal chat/completions traffic.
- One thing worth flagging: OmniRoute already parses an `X-OmniRoute-Session-Id` header in the chat pipeline (open-sse/handlers/chatCore.ts) for the memory subsystem, but it is never persisted to `call_logs`. We plan to reuse that exact header (rather than introduce a differently-named one) and add a nullable `session_tag` column, following the same pattern as the existing `correlation_id` column (migration + idempotent ALTER). We'll hold off on also reading the OpenAI `user` body field in v1 to avoid two competing precedence rules — header-only pass-through first, `user` can follow later if there's demand.

Marking this as accepted for the backlog.

diegosouzapw:
🚧 **PR opened for `release/v3.8.49`!**

[...]

PR #9291, opened 2026-08-03
fix(db): persist account egress IP into proxy_logs

maxmad64bis:
## Summary

The account **egress IP** (the outbound IP the upstream saw) is computed by `proxyEgress.ts` (echo-IP probe via `api64.ipify.org`, 5-min cache) and surfaced in the `proxy_logs` console output and in-memory ring buffer — but it was **never persisted**. `proxy_logs.egress_ip` did not exist in the schema, so the value was silently lost on restart and real traffic could not be attributed to the node/IP that was active at that instant.

This change makes the egress IP survive restarts by persisting it with each `proxy_logs` row, following the `session_tag` (#8249) pattern: numbered idempotent migration + `schemaColumns` reconciler. The base `SCHEMA_SQL` is untouched.

## Changes

- `src/lib/db/migrations/134_proxy_logs_egress_ip.sql` — `ALTER TABLE proxy_logs ADD COLUMN egress_ip TEXT` (nullable, backward-compatible; no index by design — no query filters on it).
- `src/lib/db/schemaColumns.ts` — new idempotent `ensureProxyLogsColumns(db)` reconciler (`PRAGMA table_info` guard).
- `src/lib/proxyLogger.ts` — self-heals the schema in `loadFromDb()`, persists `egress_ip` on `INSERT`, and matches it in the `search` filter (symmetric with `clientIp`).

## Tests

New + updated test files:
- `tests/unit/proxy-logs-egress-ip.test.ts` (new — 5 cases: fresh-install schema exposure + reconciler idempotency, persistence via SQL, `NULL` when absent, close/reopen round-trip on disk, search match)
- `tests/unit/db-schema-columns-split.test.ts` (self-heal on a bare `proxy_logs` + migration-134 SQL applied to a bare table)

Verification run locally:
```
node --import tsx/esm --test tests/unit/proxy-logs-egress-ip.test.ts \
  tests/unit/db-schema-columns-split.test.ts \
  tests/unit/proxy-logger-client-ip.test.ts \
  tests/unit/db-migration-runner.test.ts
# 39 tests, 0 failures
npm run check:file-size          # OK
npm run check:migration-numbering # OK
npm run check:db-rules            # OK
npm run check:cycles              # OK
npm run typecheck:core            # no errors
npx eslint (touched files)        # no errors
```

## Notes

- Migration 134 is a plain `ALTER` on the base table; no sentinel/`PHYSICAL_SCHEMA_SENTINELS` registration needed (that list covers table-creating migrations).
- The reconciler is wired in the owning module (`proxyLogger.loadFromDb`) rather than `core.ts` because `core.ts` is frozen in the file-size baseline (1637 lines); a bootstrap wiring there would trip `check:file-size`.

### DIFF
--- a/changelog.d/fixes/9291-proxy-logs-egress-ip.md
+++ b/changelog.d/fixes/9291-proxy-logs-egress-ip.md
@@ -1,0 +1,1 @@
+- **fix(db):** persist the account egress IP into `proxy_logs.egress_ip` (migration 134 + schema reconciler) so real traffic stays attributable to the actual node/IP even after restart — the egress IP was previously computed and logged but silently dropped from persistence ([#9291](https://github.com/diegosouzapw/OmniRoute/pull/9291)) — thanks @maxmad64bis
--- a/src/lib/db/migrations/134_proxy_logs_egress_ip.sql
+++ b/src/lib/db/migrations/134_proxy_logs_egress_ip.sql
@@ -1,0 +1,2 @@
+-- egress_ip: no index by design (not a query dimension) — YAGNI
+ALTER TABLE proxy_logs ADD COLUMN egress_ip TEXT;
--- a/src/lib/db/schemaColumns.ts
+++ b/src/lib/db/schemaColumns.ts
@@ -273,6 +273,22 @@ export function ensureCallLogsColumns(db: SqliteDatabase) {
   }
 }
 
+export function ensureProxyLogsColumns(db: SqliteDatabase) {
+  try {
+    const columns = db.prepare("PRAGMA table_info(proxy_logs)").all() as Array<{
+      name?: string;
+    }>;
+    const columnNames = new Set(columns.map((column) => String(column.name ?? "")));
+    if (!columnNames.has("egress_ip")) {
+      db.exec("ALTER TABLE proxy_logs ADD COLUMN egress_ip TEXT");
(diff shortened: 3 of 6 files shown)


diegosouzapw:
Merged via local merge-train on 192.168.0.113 (32 cores) @ train tip `1efacead211f446aac3146c0599eca1ca190ac4c` — log `/srv/omniroute-train/.claude/worktrees/train-20260804-130726-suite.log`.

**Green:** `typecheck:core`, `check-complexity`, `check-cognitive-complexity`, `check-changelog-integrity`, `test:vitest`.

**`test:unit`:** 15 failures, and every one of them reproduces identically on the clean release tip with zero PRs boarded (merge-gates §3) — **no new failure was introduced by this train**. Thirteen are the bare-model routing assertions left stale by #9275, which intentionally moved bare `gpt-5.5`/`gpt-5.6-sol` to codex without updating the tests that encoded the old destination; two are the trailing-period message change covered by #9392. A fix for those is in flight.

For reference, `check-file-size` is also a pre-existing base-red on the bare tip (`src/sse/handlers/chat.ts` 1846>1845, `open-sse/executors/base.ts` 1623>1578 — identical numbers with no PRs boarded): baseline drift for the release captain, not introduced here.

PR #9380, opened 2026-08-04
fix(sse): drop the localDb barrel imports from chat and auth

HouMinXi:
fix(sse): drop the localDb barrel imports from chat and auth

chat.ts and auth.ts reached into @/lib/localDb, the re-export barrel, for
functions that live in specific db/ modules. Importing through the barrel pulls
its whole surface into the module graph, which is what no-restricted-imports is
there to prevent; both files carried a frozen suppression instead of a fix.

Point each import at the module that owns the function: sessionAccountAffinity,
readCache and combos for chat.ts; readCache, providers, apiKeys and settings for
auth.ts.

auth.ts also kept a private copy of toNumber. It differs from the canonical one
in @/shared/utils/numeric only in not trimming before Number(), and Number()
already ignores surrounding whitespace, so the two agree on every input a caller
can produce. Checked rather than assumed: a differential run over 87,880 input
combinations -- leading and trailing whitespace variants, numeric and
non-numeric strings, non-string types, and several fallback values -- found no
case where they disagree. The local copy is removed in favour of the shared one.

With both violations gone the two suppression entries are stale, so they go too.
Only those two keys are touched; the rest of the file keeps its existing order.

One side effect worth flagging: chat.ts drops from 1846 to 1844 lines, which is
back under its frozen 1845 ceiling. The file-size gate on release/v3.8.50
currently reports two violations; this takes it to one. The remaining one is
base.ts and is unrelated to these imports.

Signed-off-by: Minxi Hou <houminxi@gmail.com>

### DIFF
--- a/config/quality/eslint-suppressions.json
+++ b/config/quality/eslint-suppressions.json
@@ -1362,21 +1362,8 @@
       "count": 1
     }
   },
-  "src/sse/handlers/chat.ts": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/sse/handlers/chatHelpers.ts": {
     "no-restricted-imports": {
-      "count": 1
-    }
-  },
-  "src/sse/services/auth.ts": {
-    "no-restricted-imports": {
-      "count": 1
-    },
-    "no-restricted-syntax": {
       "count": 1
     }
   },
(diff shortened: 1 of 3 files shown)


HouMinXi:
Heads-up on the file-size job, since it will be red here and that is not this PR's doing.

`npm run check:file-size` already fails on an unmodified `release/v3.8.50` (measured at `7163081f5`) with two violations:

```
✗ src/sse/handlers/chat.ts: 1846 > congelado 1845
✗ open-sse/executors/base.ts: 1623 > congelado 1578
```

This PR clears the first one as a side effect of the import split (chat.ts 1846 -> 1844). #9381 clears the second. Either one alone leaves the job red on the other file; both together take it to zero. Happy to combine them into one PR if you would rather review it that way.

PR #9381, opened 2026-08-04
refactor(sse): move the thinking-budget helpers out of base.ts

HouMinXi:
refactor(sse): move the thinking-budget helpers out of base.ts

base.ts sits 44 lines over its frozen 1578-line ceiling, so the file-size gate
fails on it. The thinking-budget helpers are the natural piece to lift out: they
are pure functions over a request body, they touch nothing on the executor, and
they are the one self-contained cluster in the file.

hasActiveClaudeThinking, collectThinkingConfigs, readNestedThinkingBudget and
clampNestedThinkingBudget move verbatim to open-sse/utils/thinkingBudget.ts.
The bodies are byte-identical to the ones they replace and base.ts calls the
three it uses at the same four sites as before, so there is no behaviour change
to review -- only a boundary.

base.ts goes from 1622 to 1562 lines, sixteen under the ceiling, so the frozen
baseline does not need adjusting. The existing coverage in
tests/unit/base-thinking-budget-config-5312.test.ts still exercises the helpers
through base.ts and passes unchanged.

Signed-off-by: Minxi Hou <houminxi@gmail.com>

### DIFF
--- a/open-sse/executors/base.ts
+++ b/open-sse/executors/base.ts
@@ -234,71 +234,11 @@ export function mergeAbortSignals(primary: AbortSignal, secondary: AbortSignal):
   return controller.signal;
 }
 
-function hasActiveClaudeThinking(body: Record<string, unknown>): boolean {
-  const thinking = body.thinking as Record<string, unknown> | undefined;
-  return thinking?.type === "enabled" || thinking?.type === "adaptive";
-}
-
-/**
- * Collect every `thinkingConfig` object in a transformed request body that holds
- * a thinking budget, wherever the provider's envelope nests it:
- *   - body.generationConfig.thinkingConfig            (native Gemini / openai→gemini)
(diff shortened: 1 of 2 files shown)


HouMinXi:
Heads-up on the file-size job, since it will be red here and that is not this PR's doing.

`npm run check:file-size` already fails on an unmodified `release/v3.8.50` (measured at `7163081f5`) with two violations:

```
✗ src/sse/handlers/chat.ts: 1846 > congelado 1845
✗ open-sse/executors/base.ts: 1623 > congelado 1578
```

This PR clears the second one (base.ts 1622 -> 1562, sixteen under the ceiling). #9380 clears the first, as a side effect of splitting a barrel import. Either one alone leaves the job red on the other file; both together take it to zero. Happy to combine them into one PR if you would rather review it that way.

PR #9388, opened 2026-08-04
fix(providers): manual Vision capable override does not affect Combo routing (#9195)

### LINKED ISSUE #9195: fix(providers): manual Vision capable override does not affect Combo routing
## Environment

- OmniRoute: 3.8.49
- Install: npm
- OS: Fedora Linux 42
- Provider: llama.cpp / OpenAI-compatible
- Model: `qwen3.6-35b`

## Problem

The Custom Models **Vision capable** checkbox persists `supportsVision: true`, but does not affect `/v1/models` or Combo routing.

This blocks self-hosted vision models whose upstream metadata is incomplete. Since #8494, image requests fail closed with `400 capability_mismatch`; the same request worked in 3.8.48.

## Reproduction

1. Add a self-hosted llama.cpp/OpenAI-compatible model under **Custom Models**.
2. Enable **Vision capable** and save.
3. Send an image request through a Combo containing that model.
4. The request fails with `400 capability_mismatch`.

The value is persisted:

```json
{"id":"qwen3.6-35b","supportsVision":true}
```

But `/v1/models` still reports no vision capability, and the Combo rejects the image request.

## Expected

The override should make the model vision-capable in `/v1/models` and in `getResolvedModelCapabilities()`, allowing the Combo request.

## Suspected cause

1. A model present in both `syncedAvailableModels` and `customModels` is emitted by the synced-model loop first; the custom loop then skips the duplicate, losing the override.
2. In the custom-model loop, `getCustomVisionCapabilityFields()` is guarded by `"chat" === e`, but `e` is never assigned `"chat"`.
3. Even if the catalog output is fixed, Combo routing uses `getResolvedModelCapabilities()`, which does not read the `customModels` vision override.

## Suggested fix

1. Make the existing Custom Models `Vision capable` value feed `getResolvedModelCapabilities()` and Combo capability filtering.
2. Add a per-target **Vision capable** checkbox in the Combo editor, persisted by provider/model, so operators can manually trust a self-hosted vision target.
3. Add tests covering `/v1/models` and an image request through a Combo.

## Related

- #7124 — added the Vision capable toggle.
- #8494 — changed capability filtering to fail closed.
- #9160 — similar issue where upstream capability metadata is ignored.

diegosouzapw:
Closes #9195

Three linked bugs prevented the Custom Models 'Vision capable' toggle from affecting Combo routing, causing 400 capability_mismatch on image requests sent through Combos targeting a custom vision model.

- **Bug #1 (catalog, dead guard):** `modelType === "chat"` was always false for chat models because modelType was only assigned "embedding", "rerank", "image", or "audio". Fixed by changing the guard to `!modelType || modelType === "chat"`.
- **Bug #2 (catalog, synced-first ordering):** When a model appeared in both syncedAvailableModels and customModels, the custom row was skipped entirely. Fixed by merging vision fields into the existing synced entry when the custom model has an explicit supportsVision boolean.
- **Bug #3 (routing capabilities):** `resolveVisionCapability()` had no path to consult the customModels supportsVision flag. Added a sync DB lookup helper so the dashboard toggle affects Combo routing.

### Fixes
- `src/app/api/v1/models/catalog.ts`: fix dead guard (line 1242) and synced-first ordering (line 1222)
- `src/lib/modelCapabilities.ts`: add custom model vision override path in resolveVisionCapability via getResolvedModelCapabilities

### DIFF
--- a/changelog.d/fixes/9195-fix.plan.md
+++ b/changelog.d/fixes/9195-fix.plan.md
@@ -1,0 +1,2 @@
+- fix(catalog): repair dead guard and synced-first ordering for custom model Vision capable override (#9195)
+- fix(routing): consult customModels supportsVision flag in Combo vision filter (#9195)
--- a/src/app/api/v1/models/catalog.ts
+++ b/src/app/api/v1/models/catalog.ts
@@ -1217,9 +1217,30 @@ async function buildUnifiedModelsResponseCore(
             continue;
           }
 
-          // Skip if already added as built-in
+          // Skip if already added as built-in. When the custom entry has an explicit
+          // supportsVision flag, merge vision fields into the existing synced entry
+          // instead of skipping (#9195).
           const aliasId = `${alias}/${modelId}`;
-          if (models.some((m) => m.id === aliasId)) continue;
+          const existingIndex = models.findIndex((m) => m.id === aliasId);
+          if (existingIndex !== -1) {
+            if (typeof model.supportsVision === "boolean") {
+              const mergeVisionFields = getCustomVisionCapabilityFields(model, aliasId, modelId);
+              if (mergeVisionFields) {
+                const existing = models[existingIndex] as Record<string, unknown>;
+                existing.capabilities = {
+                  ...((existing.capabilities as Record<string, unknown>) || {}),
+                  ...mergeVisionFields.capabilities,
+                };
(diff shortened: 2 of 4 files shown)


crazyn2:
Please merge it as soon as possible

crazyn2:
Please merge it as soon as possible

PR #9405, opened 2026-08-04
fix(backend): use accumulated responseBody for provider payload to avoid stale dashboard log viewer data (#9315)

### LINKED ISSUE #9315: fix(backend): dashboard log viewer shows stale provider response for long streamed responses
### OmniRoute Version
3.8.49

### Installation Method
Docker / Docker Compose

### Operating System
Linux

### Provider(s) Involved
Reproduced with opencode-zen (big-pickle), but the bug is provider-agnostic — it affects any streamed response whose raw SSE event capture exceeds the logger's retention cap.

### Model(s) Involved
opencode/big-pickle (reproduced); any reasoning/tool-calling model with long streamed responses

### Client Tool
cURL / any client using `/v1/responses` or `/v1/chat/completions` with `stream: true`

### Description

In the dashboard's request log detail view (`/dashboard/logs?id=...`), the "Provider Response" panel can show a stale, incomplete snapshot of the response that looks completely different from the "Client Response" panel for the same request — even though the actual response served to the client was correct and complete.

Root cause: `open-sse/utils/stream.ts` builds two separate summaries for the log artifact when a stream completes:

- The **client response** summary is built from `responseBody`, an object that is correctly accumulated in memory as the stream is processed (final message content, reasoning, tool_calls, finish_reason). This is always accurate.
- The **provider response** summary is instead rebuilt via `buildStreamSummaryFromEvents(providerPayloadCollector.getEvents(), ...)` — i.e. reconstructed from the *raw captured SSE event log*, not from accumulated state.

That raw event log is capped in `open-sse/utils/streamPayloadCollector.ts::createStructuredSSECollector` at `maxEvents = 200` / `maxBytes = 49152` (48KB). The collector is head-retaining: once the cap is hit, `push()` just increments `droppedEvents` and discards every event from that point on — it does not retain the tail. So once a stream exceeds ~200 chunks or ~48KB (very easy to hit with a reasoning model emitting a long `reasoning_content` plus tool calls), the "provider response" summary is reconstructed from only the *early* surviving chunks, silently missing everything after the cutoff: final `content`, the rest of `reasoning_content`, `tool_calls`, and the real `finish_reason`.

### Steps to Reproduce

1. Send a streamed request (`stream: true`) to a reasoning + tool-calling model (e.g. any DeepSeek-family/big-pickle-family model) with enough prior conversation history that the model's response — content + reasoning_content + tool_calls — serializes to more than ~200 SSE chunks or ~48KB of JSON. This is common on any request whose reasoning is a few hundred words plus one or more tool calls.
2. Let the request complete successfully (200, tool call or normal finish).
3. Open the request in the dashboard log viewer: `/dashboard/logs?id=<request-id>`.
4. Compare the "Provider Response" and "Client Response" panels.

### Expected Behavior

Both panels should reflect the same underlying response — the "Provider Response" panel should show the actual final content/reasoning/tool_calls/finish_reason the provider sent, matching what "Client Response" shows (translated to the client's format where applicable).

### Actual Behavior

"Provider Response" shows a stale, incomplete snapshot: `content: null`, `reasoning_content` cut off mid-sentence at the truncation boundary, `finish_reason: "stop"` (a default/fallback, not the real value), and no `tool_calls` — even when the response actually had all of that. "Client Response" for the exact same request shows the correct, complete data.

This is a logging/display-only bug — the response actually served to the client is correct (confirmed by the accurate `clientResponse` summary and by no corresponding error/anomaly in the served response). It only misleads whoever is debugging via the dashboard log viewer into thinking the provider returned something broken/incomplete when it didn't.

### Test Impact
Needs a new unit test

### Additional Context

- `createStructuredSSECollector` (`open-sse/utils/streamPayloadCollector.ts`) is a **head-retaining, tail-dropping** ring buffer: events are appended until `maxEvents`/`maxBytes` is hit, then every subsequent event is silently dropped (`droppedEvents += 1; return;`). The artifact records `_truncated: true` and `_droppedEvents: N`, but consumers building a "final state" summary from `getEvents()` (as the provider-response path does) get an early/incomplete slice, not a representative one.
- The client-response path avoids this entirely because it passes the independently-accumulated `responseBody` object as the summary input instead of deriving it from the (possibly truncated) event log — see the calls to `providerPayloadCollector.build(buildStreamSummaryFromEvents(providerPayloadCollector.getEvents(), ...))` vs `clientPayloadCollector.build(responseBody, ...)` in `open-sse/utils/stream.ts`.
- A fix likely needs the provider-side summary to be built from equivalent accumulated state (native provider format) rather than reconstructed from the truncated raw event capture, without regressing the intentional format-preserving purpose of the "provider response" panel (showing the provider's native response shape, which can differ from the client's translated shape in cross-format cases like Claude/Gemini providers behind an OpenAI-facing client).

### Validation Plan

- A unit test constructing a streamed response whose SSE payload exceeds the `maxEvents`/`maxBytes` cap and asserting that the built "provider response" summary still contains the correct final `content`/`reasoning_content`/`tool_calls`/`finish_reason` (not a truncated/defaulted snapshot).

diegosouzapw:
Closes #9315

Root cause: In stream.ts, the providerPayload for the dashboard log was built via buildStreamSummaryFromEvents(providerPayloadCollector.getEvents(), ...), which reconstructs the response from raw captured SSE events. The StructuredSSECollector is head-retaining/tail-dropping with default caps (maxEvents=200/maxBytes=48KB). Once those caps are exceeded, every subsequent event is silently dropped. For reasoning + tool-calling models, a stream easily exceeds these caps, so the "Provider Response" panel in the dashboard was built from only the early chunks — missing final content, reasoning_content, tool_calls, and the real finish_reason.

Fix: Use the same accumulated responseBody for providerPayload that the clientPayload already uses — pass responseBody directly to providerPayloadCollector.build() instead of buildStreamSummaryFromEvents(providerPayloadCollector.getEvents(), ...). This matches the client path and ensures the dashboard shows correct data regardless of stream length.

Two sites changed in stream.ts:
1. Passthrough flush onComplete (line ~2510)
2. Translate flush onComplete (line ~2785)


hartmark:
Thanks for the fix! We hit the same root cause independently while live-testing #9439 and already closed it there (commit `52671130a`) with a slightly broader version: `streamPayloadCollector.ts`'s per-format summary builders were turned into incremental reducers that ingest every chunk as it streams (`collector.getSummary()`), applied at all 3 `onComplete` call sites in `stream.ts` — including the mid-stream failure/error path (~line 2611), which still uses the old `buildStreamSummaryFromEvents(getEvents(), ...)` reconstruction here and would still show a truncated/stale Provider Response panel for a response that reasons/tool-calls and then errors out mid-stream.

#9315 will close via #9439 instead, so no action needed here — leaving this open for you to close since it's your PR, not mine to close.

Issue #5887, opened 2026-07-02
[BUG] Codex 5.5 image generation no longer auto-routes to the codex image provider (regression)


diegosouzapw:
## Summary

A user reports (community channel) that **image generation with Codex "5.5" (`gpt-5.5`) is no longer being routed automatically** to the Codex image provider "like it did before" — indicating a **regression** in auto-routing for image requests targeting the Codex image provider.

## Expected

An incoming image-generation request that should resolve to the Codex image provider (`codex` / model `gpt-5.5`) is automatically routed to `handleCodexImageGeneration`, which translates it to a `/responses` call with `tools: [{ type: "image_generation" }]`.

## Actual

The request is no longer auto-directed to the Codex image path ("não está automaticamente direcionando como antes fazia").

## Code pointers

- Registry entry: `open-sse/config/imageRegistry.ts:143-152` — `codex` provider, model `gpt-5.5`, format `codex-responses`, baseUrl `.../codex/responses`.
- Handler: `open-sse/handlers/imageGeneration.ts` + `handleCodexImageGeneration`.
- Image route + combo/alias resolution: `/v1/images/generations` (see #3214/#3215/#3219 for the alias/combo resolution path).

## Needs-repro / verification

- [ ] Confirm which request shape used to auto-route to `codex` image and whether the model id/alias (`gpt-5.5` / `cx`) still resolves through the image registry + combo/alias resolution.
- [ ] Bisect against recent image-routing / registry changes (e.g. #5679 SenseNova registry addition, #3219 alias/combo resolution).
- [ ] Add a regression test asserting a `gpt-5.5` image request resolves to the `codex` image provider.

Requires an OAuth Codex/ChatGPT upstream to validate end-to-end (live test on the VPS may be needed per the bug-fix protocol).

_Reported via community channel; reporter identity/JID intentionally omitted._

diegosouzapw:
[...]

PR #9440, opened 2026-08-04
fix(sse): restores #5887 openai precedence for bare gpt-5.5 routing

wgordon17:
## Summary
- #9275 added gpt-5.5 to CODEX_NATIVE_UNPREFIXED_MODELS, an unconditional early-return that silently made unreachable the codex-vs-openai precedence logic issue #5887 built for exactly this model (codex-only installs route to codex; when openai is also active, the historical openai default wins). Verified as a genuine regression, not a deliberate override: the #5887(b) test passed on the commit immediately before #9275 merged, and #9275's own CI run shows it failing yet merged anyway.
- Removed only gpt-5.5 (+ effort variants) from that set into a new, explicitly-checked exception set (CODEX_NATIVE_MODELS_WITH_OPENAI_PRECEDENCE) rather than relying on silent omission — agentrouter's static registry doesn't catalog gpt-5.5 at all, so the inference-race bug #9275 exists to prevent cannot occur for it, and gpt-5.6-sol / the other models #9275 actually needed to fix are untouched.
- Also fixes an unrelated pre-existing failure encountered getting this PR's own CI green: a #9275 error-message format change (trailing period) in vscode-token-routes.test.ts, plus inherited file-size drift from an unrelated already-merged commit.


diegosouzapw:
Maintainer update: merged `release/v3.8.50` into this branch and realigned it to the routing contract that shipped in the meantime via #9447 (owner-confirmed today): for bare `gpt-5.5`, OpenAI serves while no codex connection is active, an **active** codex connection preempts (the ChatGPT subscription is the quota source of truth), and an explicit `openai/…` prefix stays the per-request override — see `tests/unit/codex-gpt55-routing-5887.test.ts`.

What survives from this PR (thanks for these!): the #9275 trailing-period comment fixes in `vscode-token-routes.test.ts`, the documented #5887×#9447 interplay at the preemption site in `model.ts`, and your own-growth file-size baseline entry. The behavioral carve-out (removing gpt-5.5 from `CODEX_NATIVE_UNPREFIXED_MODELS`) was dropped since the shipped bound already covers the scenario it targeted. All 54 tests across the touched files pass. CI should go green on the next run.

wgordon17:
Superseded by #9447, which already shipped the maintainers' own resolution to the #5887/#9275 conflict: keep codex winning when both providers are active (matching #9275's intent), gated on codex actually being configured, with an explicit `openai/` prefix as the documented escape hatch (see `tests/unit/codex-gpt55-routing-5887.test.ts`). This PR's remaining diff (the file-size-baseline rebaseline and the vscode-token-routes trailing-period fix) is also redundant now — the file-size fix is superseded by `quality.yml`'s `check:file-size -- --base-ref` PR-mode change, and the test-string fix already landed upstream independently. Closing with nothing left to merge.

PR #9446, opened 2026-08-04
fix(api): point /api/cache/stats at the real semantic cache

Poid-ZA:
## Summary
- `/api/cache/stats` read `cacheLayer.getPromptCache()`, a singleton with zero callers anywhere in the request pipeline — it always reported 0 hits/misses regardless of real cache activity.
- The dashboard `/dashboard/health` and `/dashboard/usage` (`BudgetTelemetryCards`) both fetch this endpoint directly, so they were always displaying a dead cache instance instead of the real one.
- The cache actually wired into `chatCore.ts` on every request is `semanticCache.ts`. Added `getMemoryLayerStats()` there, matching the exact field shape (`size`/`maxSize`/`bytes`/`maxBytes`/`hits`/`misses`/`evictions`/`hitRate` as a number) both existing consumers already depend on, and repointed the route + its `DELETE` handler at it.
- Removed the now-unused `getPromptCache()` singleton from `cacheLayer.ts`. The `LRUCache` class it wrapped is unaffected and still used directly by `semanticCache.ts`.

## Test plan
- [x] `node --import tsx/esm --test tests/unit/semantic-cache.test.ts` — new tests for `getMemoryLayerStats()` written first (failed with the missing export), then implemented and passing (23/23 total in file)
- [x] `npm run typecheck:core` — clean
- [x] `npx eslint` on all four changed files — clean
- [x] Confirmed via grep that nothing else in `src/`, `open-sse/`, or `tests/` referenced `cacheLayer.getPromptCache` before removing it


diegosouzapw:
Thanks for this contribution — the change itself is wanted! I'm closing this PR for a **mechanical** reason only (nothing wrong with your fix): the branch was cut from a very old point of the repo, so its tree now conflicts with `release/v3.8.50` on dozens of unrelated files (docs, CI workflows, baselines, lockfiles). Merging it as-is could silently roll back months of unrelated work, and resolving 70+ conflicts by hand is riskier than recreating a branch that only contains your change.

**How to recreate it cleanly (5 minutes):**

```bash
# 1. Sync your fork's view of the upstream release branch
git fetch upstream release/v3.8.50        # add the remote first if needed:
                                          # git remote add upstream https://github.com/diegosouzapw/OmniRoute.git

# 2. Cut a FRESH branch from the current release tip (never from main or an old branch)
git switch -c <your-branch-name-v2> upstream/release/v3.8.50

# 3. Bring over ONLY your change. Two easy options:
#    a) cherry-pick your original commit(s):
git cherry-pick <sha-of-your-commit>      # repeat per commit; resolve only YOUR files
#    b) or re-apply the diff of just your files:
git checkout <old-branch> -- <path/to/your/file1> <path/to/your/file2>

# 4. Sanity: the diff must contain ONLY your files
git diff upstream/release/v3.8.50 --stat

# 5. Run the focused checks, push, and open the PR against release/v3.8.50
npm run typecheck:core && node --import tsx/esm --test tests/unit/<your-test>.test.ts
git push -u origin <your-branch-name-v2>
```

Two extra notes that will save you time:
- Several base-side CI problems were fixed today (leaked failing tests, a wrong ESLint baseline, a missing MDX frontmatter). Any red you saw on this PR before was **inherited from the base, not caused by your change** — a fresh branch from today's tip starts green.
- Please tick **"Allow edits by maintainers"** on the new PR: it lets us refresh your branch when the base moves, instead of asking you every time.

Reopen at will — or open the recreated PR and mention this one so the history links up. Happy to help if any step misbehaves.

Issue #7293, opened 2026-07-15
fix(backend): hoist client-injected 'system' messages to index 0 for strict providers — #6135 only covered the memory half


diegosouzapw:
## Summary

Providers that enforce "a `system` message may only appear at index 0" reject requests with `[400] A 'system' message can only appear at index 0 of the messages array` whenever **any** party places a `system` message later in the array. OmniRoute should normalize this defensively on the outbound request path, since it is the single proxy every request funnels through.

**Source:** Discussion #6129 by @AlexanderPadron (provider: mimo-v2.5; clients: OpenCode / Kilo Code).

## Why this is a separate issue from #6135

#6135 fixed **only the memory-injection half** — "Memory injection now keeps the injected system message first for providers that require it (`PROVIDERS_SYSTEM_M...`)" (PR #6225, shipped v3.8.45/v3.8.46) — and was closed on 2026-07-05.

But @AlexanderPadron subsequently confirmed, in the same discussion, that the 400 **still fires with memory disabled** (mimo-v2.5, `messageCount: 70`). That rules out the memory path as his trigger and confirms the other half: **the client itself** (OpenCode / Kilo Code) places a `system` message at a non-zero index, and OmniRoute forwards it untouched.

The intent to broaden #6135 to cover this was stated on the discussion, but #6135 was already closed and no tracking issue was ever opened for the client-injected case. This issue is that missing tracker. Confirmed absent from the codebase: existing hoisting logic is provider/path-specific (`open-sse/handlers/chatCore/claudeSystemRole.ts` for Claude, `open-sse/executors/codex.ts:1252` for the Responses API, `src/lib/memory/injection.ts:60` for memory) — there is **no general strict-provider normalization on the outbound request path**.

## Expected behavior

For providers declared as "system only at index 0", any `system`-role message at index > 0 in the outgoing `messages` array is hoisted to the front — **merged into** the leading system block rather than dropped — regardless of who inserted it (memory, client, or a future injector).

## Proposed implementation

Design as laid out on the discussion:

1. **Capability flag for strict-system providers** (mimo / OpenAI-compatible-strict) so normalization only runs where required and does not disturb providers that accept mid-array `system`. Note `src/lib/memory/injection.ts:60` already documents the concept ("Providers that accept a system-role message ONLY at index 0") — reuse that list/flag rather than introducing a second source of truth.
[...]

PR #9457, opened 2026-08-04
fix(sse): preserve client cache boundaries when hoisting system roles

### LINKED ISSUE #9436: fix(providers): Claude system-role hoisting moves cache_control out of messages causing history-cache misses
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Ubuntu 24.04.4 LTS

### Node.js Version

v24.18.1

### Provider(s) Involved

Anthropic (Claude OAuth)

### Model(s) Involved

claude-sonnet-5 (233 of 234 Claude upstream records), claude-opus-5 (1)

### Client Tool

Claude Code 2.1.201 (claude-cli/2.1.201 (external, cli))

### Description

On Claude-formatted requests routed to an Anthropic Claude OAuth target
(including the priority-combo path used in this incident), OmniRoute hoists
mid-conversation `role: "system"` / `role: "developer"` content blocks from
`messages[]` into the top-level `system[]` array.

The entire block is carried into `system[]` by object spread in
`extractSystemRoleMessages` and by reference in `extractSystemMessagesToBody` 
so any attached `cache_control` marker travels with it. The breakpoint layout
goes from `sys=1,msg=1` to `sys=2,msg=0`: no breakpoint remains in `messages`,
so the growing message-history portion after the cached system prefix is
processed as fresh input. This frequently forces the following turn to perform a
large cache rebuild for the affected history.

The trigger is the marker, not the message. Healthy calls in the same sessions
also retained hoisted system blocks in the outgoing `system` array, but their
active message breakpoint remained on a `role: "user"` / `blockType:
"tool_result"` block (149 of 149). 
In this dataset, hoisted system content did not correlate with cache misses as long as the active message breakpoint
remained in `messages`. That is why this affects ~11 % of calls rather than all
of them, and why short tests can easily miss it.

*Notation: `sys=N` / `msg=N` is the number of `cache_control` breakpoints in the
`system` array and in `messages`, as reported by `_omniroute.claudePromptCache`
in the call-log artifacts.*

### Steps to Reproduce

**A. Deterministic, no API call and no quota needed.** The transformation is
pure, so it can be exercised directly:

```js
import { extractSystemRoleMessages } from "./open-sse/handlers/chatCore/claudeSystemRole.ts";

const body = {
  system: [{ type: "text", text: "base prompt", cache_control: { type: "ephemeral" } }],
  messages: [
    { role: "user", content: [{ type: "text", text: "turn 1" }] },
    { role: "assistant", content: [{ type: "text", text: "reply 1" }] },
    { role: "system", content: [
        { type: "text", text: "note", cache_control: { type: "ephemeral" } } ] },
    { role: "user", content: [{ type: "tool_result", tool_use_id: "toolu_1", content: "ok" }] },
  ],
};

extractSystemRoleMessages(body);
// inspect where cache_control now sits
```

Then repeat with the marker on the `tool_result` block instead of on the system
note, that reproduces the healthy case and the layout stays `sys=1,msg=1`.

The same check against `extractSystemMessagesToBody`
(`claudeUpstreamMessages.ts`) shows the identical result.

**B. In live traffic.** Route Claude Code at a Claude OAuth connection and run a
session long enough that the client sends a mid-conversation system/developer
message carrying the cache breakpoint. Compare `_omniroute.claudePromptCache` in
the call-log artifact of that call with the preceding one.

### Expected Behavior

Hoisting a system/developer message must not remove effective cache coverage
from the accumulated message history.

The transformation should preserve the client's cache-boundary semantics.
Possible implementations include keeping the relevant message boundary in
`messages`, or applying an equivalent cache boundary after hoisting without
duplicating or silently broadening the cached content.

For targets that natively support mid-conversation system/developer messages,
skipping the hoist may be the cleaner option. A capability gate based solely on
the observed beta header would not have covered this incident:
`mid-conversation-system-2026-04-07` appeared in none of the three
`anthropic-beta` variants seen in the window and 0 times in the application log.
Strict-provider behaviour covered by #7293 must remain intact.

### Actual Behavior

When the active message breakpoint is attached to the hoisted block, the marker
moves into `system[]` and no breakpoint remains in `messages`. A large portion
of the accumulated message history is then processed as fresh input. In 21
observed cases, the following call performed a large cache rebuild.

Over an 8-hour window (2026-08-01 19:18Z – 2026-08-02 03:15Z), 321 client
requests produced 234 Claude upstream records (226 × HTTP 200, 8 errors).
Grouped by breakpoint layout:

| Layout | Calls | Median fresh input | Total fresh input |
|---|---|---|---|
| `sys=1, msg=1` | 149 | 2 | 3,584 |
| `sys=2, msg=1` | 39 | 2 | 77 |
| `sys=2, msg=0` | 24 | 97,538 | 2,275,338 |
| `sys=3, msg=0` | 2 | 204 | 409 |
| `sys=0, msg=0` | 12 | 754 | 9,026 |

The 26 affected calls with one or more system breakpoints but no message
breakpoint (`sys=2,msg=0`: 24; `sys=3,msg=0`: 2) represent 11.5 % of successful
calls. They produced 2,275,747 fresh input tokens, 99.4 % of all uncached input
in the window. The 21 rebuilds that followed them wrote 2,745,965 cache-creation
tokens, 82.8 % of all cache creation. The 12 `sys=0,msg=0` calls are a separate
category, they carry no breakpoints at all and account for 9,026 fresh tokens.

Totals over the 226 successful calls: input 27,467,017 (cache_read 21,861,426,
cache_creation 3,317,157, fresh 2,288,434), output 304,531. In OmniRoute's
call-log schema, `tokens_in` includes cache-read and cache-creation tokens;
here, `fresh` means `tokens_in - tokens_cache_read - tokens_cache_creation`.

Approximately 56 % of API-price-weighted load in the window was associated with
this miss-and-rebuild pattern. (Weighting is my own approximation of the
published multipliers: cache_read ×0.1, fresh ×1.0, cache_creation ×1.25,
output ×5 used only to compare shares, not a billing statement.)

This is not caused by duplicate upstream calls: the upstream/client ratio is
0.73, with 0 `EMPTY_CONTENT_FALLBACK` events, 0 stream recoveries and 2 retries,
both Gemini cooldown. No duplicate billable Claude upstream call was observed;
the repeated `correlation_id`s are error/summary pairs with 0 tokens.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
02:05:26  in=167,954  cache_read=167,434  cache_creation=    518  fresh=      2   sys=1,msg=1
02:05:32  in=168,797  cache_read= 42,237  cache_creation=    404  fresh=126,156   sys=2,msg=0
02:06:09  in=169,866  cache_read= 34,345  cache_creation=135,519  fresh=      2   <- large cache rebuild

before: [ 'system[0]', 'messages[2].content[0]' ]   sys=1 msg=1
after : [ 'system[0]', 'system[1]' ]                sys=2 msg=0
```

### Screenshots

_No response_

### Additional Context

---

### Root cause

`open-sse/handlers/chatCore/claudeSystemRole.ts`

```js
// line 19-21
const isSystemRole = (role) => role === "system" || role === "developer";
// line 22
const systemMessages = messages.filter((m) => isSystemRole(m.role));
// line 32
extraBlocks.push({ ...block });        // <-- copies cache_control along
// line 42
payload.system = [...existingSystem, ...extraBlocks];
// line 47
payload.messages = messages.filter((m) => !isSystemRole(m.role));
```

The spread on line 32 copies the whole content block, `cache_control` included;
line 47 then removes the message. `extractSystemMessagesToBody` in
`open-sse/handlers/chatCore/claudeUpstreamMessages.ts` (lines 20-53) has the same
defect, there pushing the block by reference. A fix touching only
`claudeSystemRole.ts` leaves that path broken.

This is not confined to the semantic-passthrough branch:
`normalizeClaudeUpstreamMessages` calls `extractSystemRoleMessages` internally
(`claudeUpstreamMessages.ts:65`), so ordinary Claude passthrough traffic reaches
it. Direct call sites are `chatCore.ts:2045` and `chatCore.ts:2083`.

`claudeSystemRole.ts` and `claudeUpstreamMessages.ts` have identical blob hashes
at tags `v3.8.49` and `v3.8.50`, so the line numbers apply to both.

### Corroborating telemetry

`messageCount` grows by +3 on every measurable affected call (22 of 22) against
+2 on healthy ones (183 of 185). The second `sys` breakpoint index increments by
exactly 1 per affected call (3, 4, 5 … 21), resetting only at session boundaries, 
consistent with the outgoing `system` array containing an increasing number of
hoisted blocks as the client re-sends the accumulated conversation state.

### Setup

Two Claude OAuth connections (priority 1 and 2) plus one Gemini API-key
connection (configured as active, `test_status=error`). Traffic did move from
the first Claude connection to the second at ~01:39Z when the first hit its
limit, this is not the cause: every miss-and-rebuild pair occurred inside a
single connection and a single model, seconds apart, well within the 5-minute
TTL.

Relevant settings: `compression.enabled=false`, `promptCacheEnabled=true`,
`promptCacheStrategy="auto"`, `alwaysPreserveClientCache="auto"`. Every
OmniRoute-side injection source was empty (`memories`, `skills`,
`middleware_hooks`, `plugins`, `api_key_context_sources` all 0 rows, combo
`system_message` NULL), so the hoisted content is client content.

### What this data cannot show

1. The incoming client body was never stored, `requestBody._truncated: true`
   with no `messages` and no `system` array, `request_detail_logs` 0 rows,
   `call_log_pipeline_enabled` set to `0`.
2. The logged breakpoints describe the outgoing state.
   `buildClaudePromptCacheLogMeta` (`executorHelpers.ts:63-69`) derives them from
   `finalBody`, i.e. after the transformation, so no before/after diff exists.
3. The `role` in the client body is inferred, not recorded — it follows from the
   code path and the telemetry, not from a logged role field.

The offline reproduction, the +3 messageCount delta in 22 of 22 measurable
cases, and the monotonic `sys` index growth are what carry the conclusion. A
byte-for-byte chain would need `call_log_pipeline_enabled` enabled and the
incident reproduced.

### Attached evidence

[breakpoint-layout.csv](https://github.com/user-attachments/files/30710553/breakpoint-layout.csv) (one row per upstream call) and
[system-role-check.csv](https://github.com/user-attachments/files/30710554/system-role-check.csv)
(affected and healthy calls side by side). Call identifiers are
pseudonymised, the raw data stays unpublished.

### Related

- #7293 — capability-gated hoisting for strict providers; requires merging, not dropping
- #1762 — broad Claude prompt-cache degradation report
- #6954 — mid-conversation system turns misattributed on Anthropic→Responses translation

---

### Validation Plan

_No response_

LeonG606:
## Summary

- Preserve Claude prompt-cache boundaries when mid-conversation `system` or `developer` messages are hoisted into the top-level `system` field.
- Relocate marked boundaries only to suitable preceding content blocks.
- Preserve Anthropic's `1h`/`5m` TTL ordering when the destination already carries a marker.
- Retain relocated markers when `tool_result` and inline file/document blocks are normalized into text.
- Cover both hoisting implementations and the native Claude normalization path.

## Related Issues

- Closes #9436
- Related to #7293

## Validation

- [x] Change type: other — SSE / Claude message transformation
- [x] `npm run lint`

Focused validation completed:

- Regression tests: 20/20 passed
- Five neighboring test files: 42/42 passed
- Broad run: the same 22 pre-existing failures as the unchanged upstream baseline; no new failures
- ESLint on all three touched TypeScript files passed
- `npm run typecheck:core` passed
- `npm run check:changelog-integrity` passed
- `git diff --check` passed

## Tests Added Or Updated

- `tests/unit/claude-system-role-cache-boundary.test.ts`

The regression coverage includes:

- both system-role hoisting implementations
- the native Claude normalization path
- valid and invalid cache-breakpoint targets
- `1h`/`5m` TTL collisions
- consecutive marked system messages
- marker preservation during `tool_result` and file/document normalization

## Coverage Notes

The new regression test directly covers the changed code in:

- `open-sse/handlers/chatCore/claudeSystemRole.ts`
- `open-sse/handlers/chatCore/claudeUpstreamMessages.ts`

No local coverage report was generated. Repository coverage and SonarQube checks are left to CI.

## Reviewer Notes

The fix does not change capability-gated system-role hoisting introduced for strict providers in #7293.

A separate pre-existing case in which the incoming request already places a `5m` system breakpoint before a later `1h` message breakpoint is intentionally outside the scope of this PR.


diegosouzapw:
Maintainer note: your branch merges cleanly with `release/v3.8.50`, but "Allow edits by maintainers" is disabled on this PR, so we can't refresh it against the repaired base (several base-side CI issues were fixed today — any red you saw earlier was inherited, not your change). Could you either tick **Allow edits by maintainers** or update the branch (merge/rebase `release/v3.8.50`)? CI should come back green after that.

LeonG606:
Enabled, thanks for the heads-up.

Issue #9407, opened 2026-08-04
fix(providers): gemini-web connection test false-positives while 15s queue masks failures


andrea-kingautomation:
## Summary

Gemini Web connections can all show active even when most cannot complete a real chat. The current validator accepts redirects that do not distinguish valid from invalid sessions, while the default 15-second protected queue expires before the Playwright chat path can expose its real failure.

## Source behavior

`validateGeminiWebProvider()` requests `https://gemini.google.com/app`, accepts statuses below 500 except 401/403, and treats the public `REDIRECT_BLOCKED` redirect as valid. The source notes that valid and invalid sessions may both redirect.

The real executor then launches Playwright, loads cookies, navigates, waits, finds `.ql-editor` or another contenteditable element, and clicks it. This path can legitimately take far longer than the default queue budget.

## Live matrix on 3.8.49

- Nine connections display active.
- Only 1/9 produced a real Gemini 3.6 completion.
- The other eight also failed a Gemini 3.5 control.
- With the normal 15-second protected queue, visible errors were local `RATE_LIMIT_QUEUE_TIMEOUT` after repeated attempts.
- In a fully isolated copied-database router with `RATE_LIMIT_MAX_WAIT_MS=120000`, the same failing account reached the real browser boundary and both models failed with `elementHandle.click: Timeout 30000ms exceeded`.
- Repeating the isolated test through a proven account-level static ISP proxy produced the identical click timeout. IP reputation is therefore not the primary cause for that account.

## Expected fix

1. Make the connection test distinguish credential presence from a usable authenticated chat session.
2. Use a browser-aware queue/timeout contract for Gemini Web or bypass the API-key-style limiter for its single browser transaction.
3. On navigation/selector/click failure retain sanitized diagnostics: origin+path, selector, actionability reason, challenge/login/consent classification, and optional screenshot.
4. Avoid retrying the same deterministic click failure four times inside one request.
5. Add account-pinned integration tests for valid, logged-out, consent-overlay, challenge, and stale-cookie pages.

No cookies, account names, proxy endpoint, raw IP, or page content are included.

Issue #9320, opened 2026-08-03
[security] Tunnel exposure: /v1/models leaks full model catalog without an API key


diegosouzapw:
@/tmp/issue-tunnel.md

PR #9205, opened 2026-08-02
feat(usage): show Grok Build billing limits

xz-dev:
## Summary

- expose stored `grok-cli` connections in Dashboard → Provider Quota using Grok Build's read-only `/user?include=subscription`, `/billing?format=credits`, and `/auto-topup-rule` surfaces
- show the live membership label when supplied, shared weekly usage/reset, product usage, Extra Usage Credits, and Auto Top-Up state while linking Additional Credits management back to Grok
- preserve sparse/unknown upstream state and last-known-good Auto Top-Up data without exposing tokens, user IDs, emails, payment data, raw responses, or adding any account mutation

## Related Issues

- Related to #6844
- Follow-up to #7714, whose primary quota fetcher intentionally deferred the dashboard card

## Validation

Choose the change type and focused loop from the
[Contribution Golden Path](../docs/dev/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite,
Vitest, the 60% coverage gate, and the production build all run in CI on this PR (#8329):

- [x] Change type: provider / UI / i18n
- [x] Focused tests and category gates from the golden path
- [x] `npm run lint`
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented below

Focused and related validation:

```text
node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --test \
  tests/unit/grok-cli-provider-limits.test.ts \
  tests/unit/grok-cli-provider-limits-ui.test.ts \
  tests/unit/provider-limits-ui.test.ts \
  tests/unit/codex-plan-label-2570.test.ts
# 45 passed

npm run lint
npm run typecheck:core
npm run i18n:check-ui-coverage
npm run i18n:check-value-drift
npm run check:dead-code
git diff --check origin/release/v3.8.50...HEAD
```

A broader Grok and Provider Limits compatibility selection also passed 289 tests during review. The independent final review concluded `Approved — C0 / I0 / M0`.

`npm run check:file-size` retains three failures that reproduce identically on the exact release baseline and are outside this diff:

- `open-sse/executors/base.ts`
- `open-sse/handlers/chatCore.ts`
- `tests/unit/chatcore-translation-paths.test.ts`

`npm run check:provider-consistency` could not run in this local dependency tree because Bun's package postinstall was not executed; the exact release baseline fails identically. CI has the canonical install environment.

[...]

PR #9484, opened 2026-08-05
fix(quality): drop two unnecessary any casts that leave the branch lint-red

HouMinXi:
`npm run lint` exits 1 on `release/v3.8.50` with nothing checked out on top of it:

```
$ git worktree add <tmp> --detach upstream/release/v3.8.50   # 2cb7567d6, no edits
$ npm run lint
tests/unit/issue-9407-gemini-web-validation-false-positive.test.ts
  50:38  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
tests/unit/v1-models-auth-leak-9320.test.ts
  83:54  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
2 problems (2 errors, 0 warnings)
(exit 1)
```

Both files arrived on 2026-08-04 with #9407 and #9320 and are absent from `config/quality/eslint-suppressions.json`, whose own most recent commit (`3440c118e`, #9205) is newer than both. So the branch fails its own gate, and every pull request opened against it inherits the failure regardless of what it changes. Two of mine opened today, touching no shared files, show the identical pair.

## The fix

Neither cast was doing anything, so this removes them rather than freezing them.

`GeminiWebExecutor` declares `testConnection` as a public method (`open-sse/executors/gemini-web.ts:359`), so the test can read it directly. `getApiKeys` returns a typed view whose `name` field infers correctly in the `find` callback, so the parameter annotation was redundant.

Leaving `config/quality/eslint-suppressions.json` untouched is deliberate: the allowlist policy in CLAUDE.md asks for the cause to be fixed where it can be, and freezing these two would have carried them forward as permanent entries.

## Verification

```
$ npm run lint
(exit 0)

$ npm run typecheck:core
(exit 0)

$ node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts \
    tests/unit/issue-9407-gemini-web-validation-false-positive.test.ts \
    tests/unit/v1-models-auth-leak-9320.test.ts
# tests 12
# pass 12
# fail 0
```

Removing a cast can change what an assertion actually observes, so the `#9320` lookup was re-pointed at a key name that does not exist to confirm the assertion still has teeth:

```
# tests 12
# pass 11
# fail 1
```

Restored, and green again.

## What this does to CI, measured against a PR without it

The `lint-guard` job runs `npm run lint:json -- --max-warnings 0` as its second step. Comparing that job across two of my PRs opened minutes apart against the same base, one carrying this change and one not:

```
                                        without this change   with it
ESLint (baseline congelado)             failure               success
Run npm run quality:collect             skipped               success
Ratchet check (blocking)                skipped               success
Require-tighten (blocking)              skipped               success
CodeQL alerts ratchet (blocking)        skipped               failure
```

So this is necessary but not sufficient. It clears the first gate; the job then reaches three more that were previously never evaluated, and stops at the CodeQL ratchet:

```
[codeql-ratchet] Alertas CodeQL abertos (nao-dismissed): 1
[codeql-ratchet]   Top regras: js/stack-trace-exposure(1)
[codeql-ratchet] REGRESSAO - 1 alertas CodeQL abertos > baseline 0
```

That one alert is repository-wide rather than per-PR, so it is red on every open pull request too. `js/stack-trace-exposure` is the rule CLAUDE.md Hard Rule #14 already documents as a known CodeQL limitation on call sites that route through `sanitizeErrorMessage()`, which suggests a dismissal rather than a code change, but that is your call and not something I would do unasked.

The neighbouring `fast-gates` job stops even earlier, at `check:file-size`:

```
[file-size] 2 violacoes:
  src/sse/handlers/chat.ts: 1846 > congelado 1845
  open-sse/executors/base.ts: 1623 > congelado 1578
```

Neither file is touched here.

## Related

#8542 describes exactly the compounding this makes visible: with the ESLint step red, nothing downstream in that job ever ran, so the CodeQL regression was invisible in every PR log until the first gate was cleared. #8522 describes the same pattern from the baseline side. Both still open.


HouMinXi:
#9488 cleared both casts, plus a third base-red in the translate-path snapshot that I had missed. Nothing left in this branch. Closing.

PR #9493, opened 2026-08-05
feat(ci): add windows-latest leg to test-bun-sqlite job (#8468)

### LINKED ISSUE #8468: feat(nodejs): extend the Bun compatibility CI job to Windows — the Bun+Windows startup path has zero coverage
## Summary

The Bun compatibility suite runs only on Linux, so the Bun-on-Windows startup path — the exact combination reported in #8091 — has never been exercised by CI. We cannot currently tell a real defect in our Bun compat path from an upstream Bun issue, because nothing verifies that combination.

## Current state

`.github/workflows/ci.yml:763-778`, job `test-bun-sqlite`:

```yaml
test-bun-sqlite:
  name: Bun SQLite Compatibility
  runs-on: ubuntu-latest      # ← only leg
  ...
  - run: npm run test:bun:db
```

`package.json:115`:

```
"test:bun:db": "bun test tests/unit/db-adapters/bunSqliteAdapter.test.ts tests/unit/db-adapters/driverFactory.test.ts tests/unit/db-adapters/cliSqlite.test.mjs"
```

The only Windows leg in the whole workflow is `electron-package-smoke` (`ci.yml:650-663`, `os: [ubuntu-latest, windows-latest]`), which runs **Node**, not Bun, and only exercises `prepare:bundle`.

So: Bun is tested only on Linux; Windows is tested only under Node. The intersection is untested.

## Why it matters

The Bun path is explicitly best-effort and unsupported, and that is fine — but two things make the Windows gap worth closing specifically:

1. **Process spawning is the part most likely to differ on Windows**, and it is Bun-aware in our code: `bin/cli/runtime/processSupervisor.mjs:44-51` picks `process.execPath` when `process.versions.bun` is set, and skips the heap args. `bin/cli/commands/serve.mjs` duplicates the same branch twice. None of that runs on Windows in CI, under Bun or otherwise.
2. **We already got burned by a Windows spawn bug once** — the `electron-package-smoke` job exists precisely because of it; its own comment cites "the v3.8.48 Windows bug (npx.cmd spawned...)" as the reason the job was added.

There are also two unguarded `bun:sqlite` imports outside the driver-factory fallback chain — `bin/cli/sqlite.mjs:6-9` and `src/lib/db/omp.ts:5-8` — which have no try/catch, unlike `src/lib/db/adapters/driverFactory.ts:71-132` where each driver attempt falls through to the next. Those would fail hard rather than degrade.

## Proposal

Add a `windows-latest` leg to `test-bun-sqlite` (matrix, or a separate advisory job). Given Bun is a best-effort path, `continue-on-error: true` is acceptable to start — the value is having the signal at all, not gating merges on it.

Optionally extend the smoke beyond the DB adapters to a minimal `serve` start/stop under Bun, since spawning is the actual risk area and the current three test files only cover SQLite adapters.

## Acceptance criteria

- CI runs the Bun compatibility suite on Windows (advisory is fine).
- A Bun+Windows failure is visible in CI output rather than only reachable through a user report.
- If a Windows-specific Bun defect is found, it is either fixed or documented as an unsupported-path limitation.

Refs #8091

diegosouzapw:
Add a `windows-latest` matrix leg to the `test-bun-sqlite` CI job with `continue-on-error: true` so Bun+Windows failures are visible in CI rather than only via user reports.

**Changes:**
- Matrix: [ubuntu-latest, windows-latest] with fail-fast: false
- Windows leg is continue-on-error: true (advisory signal, not a merge gate)
- Bun install step for Windows via bun.sh installer (pwsh)
- Update CLAUDE.md Bun section to note Windows CI coverage

Closes #8468


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

PR #9499, opened 2026-08-05
feat(images): add full combo strategy execution for image generation (#9239)

### LINKED ISSUE #9239: feat(api): execute full Combo strategy and provider fallback in /v1/images/generations
## Problem / Use Case

`POST /v1/images/generations` recognizes a bare Combo name, but it does not execute the Combo strategy. It resolves the Combo to the first target and then runs that target as a normal single-provider image request.

This means an image Combo configured with `priority` cannot fall back to its second or third provider when the first provider is unavailable, rate-limited, or returns an upstream error. The request is also reported as a direct provider request rather than a Combo execution.

This is distinct from #3215. That issue added bare alias/Combo-name resolution to a single image target. The missing functionality here is execution of the complete multi-target Combo strategy.

### Current code path

In `src/lib/images/imageRouteModel.ts`, `resolveSingleImageComboTarget()` expands the Combo but returns only the first usable target:

```ts
const targets = resolveComboTargets(combo as never, allCombos as never);
const first = targets.find(/* ... */);
return (first?.modelStr as string) ?? null;
```

`src/app/api/v1/images/generations/route.ts` then overwrites `body.model` with that target and calls `handleImageGeneration()` once. It never enters the Combo execution/fallback machinery in `open-sse/services/combo.ts`.

Verified against `release/v3.8.50` (`cfb44dab918e164b183bf2f2b5fa5a5a8bd0d693`).

## Steps to Reproduce

1. Configure two OpenAI-compatible providers with `apiType: images-generations` and image-capable custom models.
2. Create a Combo named `gpt-image-2` using the `priority` strategy:
   1. `provider-a/gpt-image-2`
   2. `provider-b/gpt-image-2`
3. Make provider A fail while provider B remains healthy.
4. Send:

```http
POST /v1/images/generations

{
  "model": "gpt-image-2",
  "prompt": "A red circle on a white background"
}
```

## Actual Behavior

- The bare Combo name is resolved to `provider-a/gpt-image-2`.
- Only provider A is attempted.
- Provider B is never attempted.
- Combo strategy, attempt history, and fallback telemetry are absent because the request has already been converted into a direct model request.

The resolution can also be observed without incurring image-generation cost: omit `prompt`, and the validation error reports the first target's internal `<nodeId>/<model>` ID rather than preserving the requested Combo name.

## Proposed Solution

Add an image-specific Combo execution path for `/v1/images/generations`:

1. Preserve the requested Combo and its strategy instead of replacing it with the first target before dispatch.
2. Expand and capability-filter Combo targets to models supporting the `images` endpoint.
3. Execute each target through the existing single-target `handleImageGeneration()` path.
4. Apply the configured Combo strategy and continue to eligible fallback targets on retryable provider failures.
5. Preserve existing direct `provider/model` behavior.
6. Record the Combo name, strategy, selected target, attempt order, and fallback count in logs/telemetry/response metadata.

This could reuse the generic Combo execution primitives where practical, but image payloads and image response semantics should remain handled by the image-specific single-target callback.

## Alternatives Considered

- Putting the most reliable provider first only changes which single provider is used; it does not provide fallback.
- Calling provider-prefixed model IDs directly bypasses the Combo entirely.
- Client-side retry requires every image client to duplicate OmniRoute routing logic.

## Acceptance Criteria

- A bare image Combo name executes its configured multi-target strategy rather than only its first target.
- With a `priority` Combo, a retryable failure from target 1 causes target 2 to be attempted.
- A successful first target does not call later targets.
- Direct `provider/model` image requests remain unchanged.
- Non-image targets are skipped or rejected deterministically.
- Logs and response metadata retain both the requested Combo name and the selected provider/model.
- Cost and usage accounting are attributed to the provider/model that actually succeeds.
- Unit coverage verifies Combo resolution without destructive first-target flattening.
- Integration coverage verifies successful image fallback from provider A to provider B.

## Area

- Proxy / Routing
- Provider Support

## Expected Test Plan

- Extend `tests/unit/image-routes-combo-edits-3214-3215.test.ts` with multi-target Combo behavior.
- Add an integration test for `/v1/images/generations` where the first image target returns a retryable error and the second succeeds.
- Verify direct image models and single-target aliases remain backward compatible.

diegosouzapw:
Add an image-specific combo execution path for `POST /v1/images/generations`. Currently, combo names are flattened to the first target; now the full combo strategy is executed with fallback across targets.

**Changes:**
- New `open-sse/services/imageCombo.ts` module with `executeImageCombo()`: resolves combo targets, filters to images-capable, iterates with priority strategy, calls `handleImageGeneration` per target, returns first success or last failure
- Route diversion in `generations/route.ts`: detects combo names before model resolution and diverts to combo execution
- 8 tests covering combo-not-found, no-image-targets, credential resolution, model filtering, non-combo passthrough, terminal failure handling, and stack-trace sanitization
- Response headers: `X-OmniRoute-Decision`, `X-OmniRoute-Fallback-Attempts`

Closes #9239


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

PR #9501, opened 2026-08-05
fix(build): clean dist before build to prevent stale chunk mismatches

Egorich-print:
## Summary
- `npm run build` left stale files under `dist/.build/next` (old server/static chunks from prior builds).
- HTML/manifest from a new build could reference hashed chunks that no longer existed on disk → browser got 404 on JS → React failed to hydrate → WebUI shell/sidebar loaded but tab content stayed blank.
- Fix: wipe `dist/` before each build so `assembleStandalone` always writes a coherent tree.

## Root cause
`build:release` already did `rm -rf .build dist`, but plain `build` did not. Incremental/local builds then mixed old and new chunk sets. Observed missing chunk example: `/_next/static/chunks/1g3bybroxcyp5.js` (404 while other assets 200).

## Change
```diff
- "build": "node scripts/build/build-next-isolated.mjs",
+ "build": "rm -rf dist && node scripts/build/build-next-isolated.mjs",
```

## Verification
1. Rebuilt, synced fresh standalone `.build/next` into installed package.
2. Restarted server.
3. All dashboard static assets (/, /dashboard/providers, /dashboard/combos, /dashboard/logs) → 0 bad JS/CSS refs.
4. Previously 404 chunk now present and served 200.

## Notes
- Does not change runtime behavior beyond making builds reproducible.
- Aligns everyday `build` with the clean-slate behavior of `build:release`.


Egorich-print:
PR into wrong branch

PR #9510, opened 2026-08-05
feat(resilience): add per-account resilience connections view (API + dashboard)

HouMinXi:
The three temporary-failure mechanisms documented in `CLAUDE.md` → "Resilience Runtime State" each have their own scope, and nothing currently shows them together:

| Mechanism | Scope | Where its state lives |
|---|---|---|
| Provider circuit breaker | whole provider | `domain_circuit_breakers` / `circuitBreaker.ts` |
| Connection cooldown | one account/key | `rateLimitedUntil` + `testStatus` on the connection |
| Model lockout | provider + connection + model | `accountFallback.ts` |

The repo's own debugging guidance asks an operator to "inspect both provider breaker state and each connection's `rateLimitedUntil`/`testStatus`" — which today means reading three separate surfaces and correlating by hand. This adds one read-only view that does the correlation.

## What it adds

**`GET /api/resilience/connections`** — all three mechanisms keyed by connection, plus the breaker's transition history.

The transition history is the part that is genuinely hard to get elsewhere: a breaker sitting at `CLOSED` right now looks healthy, and only the sequence shows it has opened four times in the last hour. `getStatus()` already assembled everything except that history; it now returns a copy of it and carries an explicit `CircuitBreakerStatus` type instead of an inferred one.

**`/dashboard/resilience/connections`** — the table, a per-connection detail view, and the transitions drawn as a timeline. It polls rather than streams: the state changes on the order of seconds to minutes and the page is loopback-gated, so an SSE channel would buy nothing over an interval.

## Notes for review

**Column projection.** Reading raw connection rows meant widening `getRawProviderConnections`' projection, so the existing `PROVIDER_CONNECTIONS_COLUMNS` allowlist is exported and the route selects through it. A test asserts every column the route names is in that allowlist, so a future typo fails here instead of silently producing an empty field.

**Partial degradation.** Each of the three data sources is wrapped independently — one throwing degrades that section and sets `meta.degraded` rather than failing the whole response, since a partial view still answers most of the questions the page exists for.

**Route guard.** Both paths are loopback-gated. This route spawns nothing, unlike every other entry on that list, so the comment says so explicitly to keep it from being read as precedent for gating read-only routes generally (Hard Rules #15/#17 are about spawn surfaces; this one is gated because it exposes per-account operational state).

**Two small shared extractions**, both to avoid a second copy rather than for their own sake:
- `ModelCooldownsCard` had its own `formatRemaining`; the new table needs the same countdown format. Moved to `src/shared/utils/formatRemaining.ts`, both import it. Behaviour is unchanged — the extracted version differs from the deleted one only in local variable names.
- `DataTable`'s column/row interfaces are exported so the new table types against them rather than restating their shape.

**Tests are real isolated-DB integration tests, not mocks.** ESM mocking is unavailable here (no `mock.module`, non-configurable exports), and the codebase already has the isolated-DB pattern, which exercises more than a mock would anyway.

## Verification

```
node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts \
     tests/unit/resilience-connections.test.ts \
     tests/unit/resilience-connections-page-static.test.ts
# tests 34
# pass 34
# fail 0

npx vitest run --config vitest.config.ts tests/unit/ui/resilience-connections.test.tsx
Test Files  1 passed (1)
     Tests  17 passed (17)

npm run typecheck:core          (exit 0)
npm run check:db-rules          PASS
npm run check:docs-sync         PASS
npm run check:tracked-artifacts PASS
npm run i18n:check-ui-coverage  PASS - all 42 locale(s) at or above 80%
eslint on all changed files     (exit 0, no output)
```

Both commits verified individually green (`typecheck:core` + `check:db-rules` + `check:migration-numbering`), so the history bisects.

`npm run check:file-size` reports `open-sse/executors/base.ts: 1623 > 1578`. That file is byte-identical to the base tip and absent from this diff — pre-existing, being handled in #9381.

### DIFF
--- a/changelog.d/fixes/9533-ratelimit-bound-queue-wait-bottleneck-exit.md
+++ b/changelog.d/fixes/9533-ratelimit-bound-queue-wait-bottleneck-exit.md
@@ -1,0 +1,1 @@
+- **fix(ratelimit):** added queue-wait timeout tests and updateFromResponseBody sequencing tests for the existing RATE_LIMIT_QUEUE_TIMEOUT feature in withRateLimit (#9533)
--- a/tests/unit/rateLimitManager-queue-timeout.test.ts
+++ b/tests/unit/rateLimitManager-queue-timeout.test.ts
@@ -1,0 +1,38 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+const rlm = await import("../../open-sse/services/rateLimitManager.ts");
+const { enableRateLimitProtection, withRateLimit, __resetRateLimitManagerForTests } = rlm;
+
+test.beforeEach(async () => {
+  await __resetRateLimitManagerForTests();
+});
+
+test("withRateLimit works without abort signal (backward compat)", async () => {
+  enableRateLimitProtection("test-queue-1");
+  const result = await withRateLimit("openai", "test-queue-1", "gpt-4", async () => "ok");
+  assert.equal(result, "ok");
+});
+
+test("withRateLimit works with AbortSignal", async () => {
+  enableRateLimitProtection("test-queue-2");
+  const ac = new AbortController();
+  const result = await withRateLimit(
+    "openai",
+    "test-queue-2",
+    "gpt-4",
+    async () => "ok",
+    ac.signal
+  );
+  assert.equal(result, "ok");
+  ac.abort();
+});
+
+test("multiple sequential withRateLimit calls work", async () => {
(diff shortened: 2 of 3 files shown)


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

Issue #7940, opened 2026-07-21
fix(resilience): Model Lockouts ignore configured cap and upstream retry hint (freeze until midnight)


Gringo675:
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

24.15.0

### Description

Seems like Resilence third layer doesn't respect Configuration options. Or am I misunderstanding this setting?

<img width="806" height="361" alt="Image" src="https://github.com/user-attachments/assets/73606523-10de-4e57-8cf8-92ed0e388109" />

<img width="817" height="498" alt="Image" src="https://github.com/user-attachments/assets/ec42abd9-108d-4ecb-a1c5-6258222e40ec" />

And also, it doesn't respect  "retryDelay" params from the responce, the model is always frozen until midnight local time.

### Steps to Reproduce

Connect to Google free-tier and spend all the limits of any model (or use model without active limits, like gemini/gemini-3.1-pro-preview) to get a similar response:

[429]: You exceeded your current quota, please check your plan and billing details. For more information on this error, head to: https://ai.google.dev/gemini-api/docs/rate-limits. To monitor your current usage, head to: https://ai.dev/rate-limit. * Quota exceeded for metric: generativelanguage.googleapis.com/generate_content_free_tier_input_token_count, limit: 16000, model: gemma-4-26b Please retry in 26.660853464s.

{
  "timestamp": "2026-07-21T06:10:30.916Z",
  "status": 429,
  "statusText": "",
  "headers": {
    "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000",
    "content-length": "1380",
    "content-type": "text/event-stream",
    "date": "Tue, 21 Jul 2026 06:10:33 GMT",
    "server": "scaffolding on HTTPServer2",
    "server-timing": "gfet4t7; dur=177",
    "vary": "Origin, X-Origin, Referer",
    "x-content-type-options": "nosniff",
    "x-frame-options": "SAMEORIGIN",
    "x-xss-protection": "0"
  },
  "body": {
    "error": {
      "code": 429,
[...]

PR #7360, opened 2026-07-16
Add per-connection Provider Quota visibility

JxnLexn:
## Summar

- Let users show or hide individual provider connections on Provider Quota views without affecting routing or account activation.

## Validation

- [x] `npm run lint`
- [x] `npm run test:unit`
- [x] `npm run test:coverage`
- [x] Coverage is still `>= 60%` for statements, lines, functions, and branches
- [x] SonarQube PR analysis is green or any remaining issues are explicitly documented below

## Tests Added Or Updated

- Provider connection hook tests cover the new visibility handler.
- Dashboard component tests cover toggling quota visibility from a connection row.
- Provider database and schema tests cover defaults, persistence, and boolean validation.
- Provider quota visibility utility tests cover opt-out behavior.

## Coverage Notes

- Database and schema tests cover persistence and request validation.
- Component tests cover the visibility control, while utility tests cover filtering semantics and backward-compatible defaults.
- No coverage results were provided.

## Reviewer Notes

- Migration `121_provider_connection_quota_visibility.sql` adds `quota_visible` with a visible-by-default value.
- The setting only controls quota-focused dashboard presentation and does not change routing or account activation.

Issue #8189, opened 2026-07-22
fix(api): Messages API returns synthetic `<block>no</block>` — claudeClassifierCompat trigger too broad


opensource-elearning:
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Ubuntu 24.04.4 LTS

### Node.js Version

v24.18.0

### Provider(s) Involved

auto/coding:reliable

### Model(s) Involved

auto/coding:reliable

### Client Tool

claude code, codex

### Description

OK

### Steps to Reproduce

```
#!/bin/bash

URL="http://localhost:20128/v1/messages"

MODELS=$(curl -s http://localhost:20128/v1/models | jq -r '.data[].id' | grep -E '^(aug|oc|[removed-provider]|ddgw)/')

for model in $MODELS
do
    echo ""
    echo "================================"
    echo "$model"
    echo "================================"

    result=$(curl -s "$URL" \
      -H "content-type: application/json" \
      -H "x-api-key: test" \
      -d "{
        \"model\":\"$model\",
        \"max_tokens\":20,
        \"messages\":[
          {
            \"role\":\"user\",
            \"content\":\"Reply only OK\"
          }
        ]
      }")

    selected=$(echo "$result" | jq -r '.model // empty')
    text=$(echo "$result" | jq -r '.content[0].text // empty')
    error=$(echo "$result" | jq -r '.error.message // empty')

    if [[ "$text" == "OK" ]]; then
        echo "✅ WORKS -> $selected"
    elif [[ "$text" == "<block>no</block>" ]]; then
        echo "⚠ FORMAT ISSUE -> $selected"
    elif [[ -n "$error" ]]; then
        echo "❌ ERROR -> $error"
    else
        echo "? -> $selected -> $text"
    fi

done


### Expected Behavior

`OK`

### Actual Behavior

`<block>no</block>`

### Test Impact

Unsure

### Error Logs / Output

```shell
================================
ddgw/gpt-4o-mini
================================
⚠ FORMAT ISSUE -> duckduckgo-web/gpt-4o-mini

================================
ddgw/gpt-5-mini
================================
⚠ FORMAT ISSUE -> duckduckgo-web/gpt-5-mini

================================
ddgw/claude-3-5-haiku-20241022
================================
⚠ FORMAT ISSUE -> duckduckgo-web/claude-3-5-haiku-20241022

================================
ddgw/llama-4-scout
================================
[...]

PR #9513, opened 2026-08-05
fix(classify): honor upstream retry windows on Gemini free-tier 429s

### LINKED ISSUE #9504: fix(resilience): Gemini free-tier 429s always classified as quota_exhausted
## OmniRoute Version

Reproduced against the published **v3.8.49** tag and the current **`release/v3.8.50`** default-branch HEAD (identical `QUOTA_PATTERNS` for this bug in both — only an unrelated Modal-provider pattern was added between the two). Also observed in production on a self-hosted Docker deployment built from `3.8.49-pre`.

## Installation Method

Docker / Docker Compose

## Operating System

Linux

## OS Version

Debian 12 (VPS)

## Provider(s) Involved

Gemini (Google AI Studio, free tier, API key)

## Model(s) Involved

gemini-3.5-flash-lite, gemini-3.6-flash

## Client Tool

Internal SSE chat pipeline (`/v1/chat/completions`), reproducible standalone against `classify429()` directly — see Steps to Reproduce.

## Description

`classify429FromError` / `looksLikeQuotaExhausted` (`src/shared/utils/classify429.ts`) classify **every** Gemini free-tier 429 as `"quota_exhausted"`, regardless of whether the upstream quota is a genuine long-window exhaustion (RPD) or a short per-minute/per-token throttle (RPM/TPM) that Google itself says will clear in seconds.

Google's real `generativelanguage.googleapis.com` 429 body always opens with the same boilerplate —

> "You exceeded your current quota, please check your plan and billing details... \* Quota exceeded for metric: ..."

— regardless of which `quotaId` (`GenerateRequestsPerMinutePerProjectPerModel-FreeTier`, `GenerateRequestsPerDayPerProjectPerModel-FreeTier`, `GenerateContentInputTokensPerModelPerMinute-FreeTier`, ...) actually fired. That boilerplate alone trips three `QUOTA_PATTERNS` entries unconditionally:

- `/quota.*exceed/i` — matches "**Quota** **exceed**ed for metric"
- `/exceed.*quota/i` — matches "You **exceed**ed your current **quota**"
- `/plan.*limit/i` — matches "check your **plan**... " + "... **limit**: 20" (the two words land on the same JSON-stringified line as the classifier scans the whole body)

None of the existing `QUOTA_PATTERNS` inspect the `quotaId` field or the Google `RetryInfo.retryDelay` the response actually carries — the one field that unambiguously tells you how long the real cooldown is. The project already recognized this exact class of problem for *another* Google message shape (`"Resource has been exhausted (e.g. check quota)."`) and gated it behind `/resource has been exhausted.*reset after/i` — requiring an explicit reset-window qualifier before classifying as `quota_exhausted` (see `tests/unit/classify429.test.ts`, test *"Google RESOURCE_EXHAUSTED without a reset remains a rate limit"*). The real Gemini `generativelanguage.googleapis.com` format never reaches that guarded pattern — it's already matched by the three unconditional patterns above.

**Downstream effect** (`src/sse/handlers/chat.ts`): `classify429FromError` feeds directly into the circuit breaker's `cooldownByKind: { rate_limit: 60_000, quota_exhausted: 3_600_000 }` (chat.ts:1192-1196) — a flat **1-hour** cooldown is applied even when Google's own `RetryInfo.retryDelay` in the same response says `"36s"`. For Gemini specifically this lands as a per-model lock rather than whole-connection poisoning (`hasPerModelQuota('gemini') === true` short-circuits `shouldMarkAccountExhaustedFrom429`), which limits blast radius to the affected model — but the lock duration itself is still the miscalibrated 1h bucket, not the ~10-60s Google actually specified.

### Relation to #6059

This is the concrete repro the maintainer asked for on #6059 (closed `not planned` on 2026-07-17 for lack of a raw-429 sample) — captured via OmniRoute's own `/api/logs/detail` pipeline-capture endpoint against a real running instance, 2026-08-01 → 2026-08-05, 45 raw Gemini 429 payloads with full `RetryInfo`/`QuotaFailure` bodies. Filing as a new issue rather than reopening #6059 because the mechanism turned out to be different from what was hypothesized there: it isn't a `quotaId`-vs-`quotaMetric` mix-up in a field-reading code path — the classifier never reads `quotaId` or `retryDelay` at all, it's a pure keyword match on Google's generic preamble text that fires the same way for every free-tier 429 shape. PR #7565 (merged 2026-07-17) is unrelated — it fixes #7284 (dashboard connection-test masking 429 as green), and its own description explicitly disclaims covering #6059.

Confirmed **not** fixed by upgrading to the latest official image: the `QUOTA_PATTERNS` list relevant to this bug is byte-identical between the `v3.8.49` tag and the current `release/v3.8.50` default branch.

## Steps to Reproduce

Standalone repro against the classifier — no live Gemini account needed, uses a real captured payload:

```ts
import { classify429 } from "./src/shared/utils/classify429.ts";

// Real Gemini free-tier 429, quotaId says "PerMinute", Google's own
// RetryInfo says the window clears in 38s.
const body = {
  error: {
    code: 429,
    message:
      "You exceeded your current quota, please check your plan and billing details. " +
      "For more information on this error, head to: https://ai.google.dev/gemini-api/docs/rate-limits. " +
      "To monitor your current usage, head to: https://ai.dev/rate-limit. \n" +
      "* Quota exceeded for metric: generativelanguage.googleapis.com/generate_content_free_tier_requests, " +
      "limit: 15, model: gemini-3.5-flash-lite\nPlease retry in 38.922534355s.",
    status: "RESOURCE_EXHAUSTED",
    details: [
      {
        "@type": "type.googleapis.com/google.rpc.QuotaFailure",
        violations: [
          {
            quotaMetric: "generativelanguage.googleapis.com/generate_content_free_tier_requests",
            quotaId: "GenerateRequestsPerMinutePerProjectPerModel-FreeTier",
            quotaDimensions: "[MaxDepth]",
            quotaValue: "15",
          },
        ],
      },
      { "@type": "type.googleapis.com/google.rpc.RetryInfo", retryDelay: "38s" },
    ],
  },
};

console.log(classify429({ status: 429, body }));
// → "quota_exhausted" (1h cooldown bucket)
// Google's own retryDelay says 38s. The *PerDay*-named quotaId variant
// (see Additional Context) produces the identical "quota_exhausted"
// result with the same ~30-50s retryDelay range — the classifier cannot
// tell the two apart because it never looks at quotaId or retryDelay.
```

In production: repeatedly call a Gemini free-tier combo/connection past its per-minute request or token quota and inspect the resulting model lockout duration in the dashboard vs. the `retryDelay` in the raw 429 (visible via `POST /api/logs/detail {"enabled": true}` pipeline capture).

## Expected Behavior

A Gemini free-tier 429 whose `RetryInfo.retryDelay` (or `quotaId` naming) indicates a short per-minute/per-token window should classify as `"rate_limit"` (60s cooldown) or use the actual `retryDelay` value directly — not the flat 1-hour `quota_exhausted` bucket. Only a `quotaId` that is genuinely day/month-scoped (and/or a `retryDelay` in the hours+ range) should trigger the long cooldown.

## Actual Behavior

Every captured Gemini free-tier 429 — per-minute request quota (`GenerateRequestsPerMinutePerProjectPerModel-FreeTier`, retryDelay ~14-58s), the confusingly-day-labeled per-project-per-model quota (`GenerateRequestsPerDayPerProjectPerModel-FreeTier`, retryDelay ~29-49s despite the name), and the per-minute token quota (`GenerateContentInputTokensPerModelPerMinute-FreeTier`, retryDelay 0-59s) — classifies as `"quota_exhausted"` and gets the flat 1-hour cooldown bucket, a 60-360x overshoot versus what Google's own `RetryInfo.retryDelay` specifies in the same response.

## Test Impact

Needs a new unit test

## Error Logs / Output

Three representative raw payloads out of 45 captured 2026-08-01 → 2026-08-05 (`pipeline.providerResponse` from OmniRoute's own detailed-log capture), one per `quotaId` variant seen:

```json
// quotaId: GenerateRequestsPerDayPerProjectPerModel-FreeTier (name says "day")
{
  "timestamp": "2026-08-01T06:14:23.628Z",
  "status": 429,
  "body": {
    "error": {
      "code": 429,
      "message": "You exceeded your current quota, please check your plan and billing details. For more information on this error, head to: https://ai.google.dev/gemini-api/docs/rate-limits. To monitor your current usage, head to: https://ai.dev/rate-limit. \n* Quota exceeded for metric: generativelanguage.googleapis.com/generate_content_free_tier_requests, limit: 20, model: gemini-3.6-flash\nPlease retry in 36.429760048s.",
      "status": "RESOURCE_EXHAUSTED",
      "details": [
        { "@type": "type.googleapis.com/google.rpc.QuotaFailure", "violations": [
          { "quotaMetric": "generativelanguage.googleapis.com/generate_content_free_tier_requests",
            "quotaId": "GenerateRequestsPerDayPerProjectPerModel-FreeTier",
            "quotaDimensions": "[MaxDepth]", "quotaValue": "20" } ] },
        { "@type": "type.googleapis.com/google.rpc.RetryInfo", "retryDelay": "36s" }
      ]
    }
  }
}

// quotaId: GenerateRequestsPerMinutePerProjectPerModel-FreeTier (name says "minute")
{
  "timestamp": "2026-08-01T07:02:21.122Z",
  "status": 429,
  "body": {
    "error": {
      "code": 429,
      "message": "You exceeded your current quota... * Quota exceeded for metric: generativelanguage.googleapis.com/generate_content_free_tier_requests, limit: 15, model: gemini-3.5-flash-lite\nPlease retry in 38.922534355s.",
      "status": "RESOURCE_EXHAUSTED",
      "details": [
        { "@type": "type.googleapis.com/google.rpc.QuotaFailure", "violations": [
          { "quotaMetric": "generativelanguage.googleapis.com/generate_content_free_tier_requests",
            "quotaId": "GenerateRequestsPerMinutePerProjectPerModel-FreeTier",
            "quotaDimensions": "[MaxDepth]", "quotaValue": "15" } ] },
        { "@type": "type.googleapis.com/google.rpc.RetryInfo", "retryDelay": "38s" }
      ]
    }
  }
}

// quotaId: GenerateContentInputTokensPerModelPerMinute-FreeTier (token quota, per-minute)
{
  "timestamp": "2026-08-01T06:14:03.805Z",
  "status": 429,
  "body": {
    "error": {
      "code": 429,
      "message": "You exceeded your current quota... * Quota exceeded for metric: generativelanguage.googleapis.com/generate_content_free_tier_input_token_count, limit: 250000, model: gemini-3.5-flash-lite\nPlease retry in 56.257848556s.",
      "status": "RESOURCE_EXHAUSTED",
      "details": [
        { "@type": "type.googleapis.com/google.rpc.QuotaFailure", "violations": [
          { "quotaMetric": "generativelanguage.googleapis.com/generate_content_free_tier_input_token_count",
            "quotaId": "GenerateContentInputTokensPerModelPerMinute-FreeTier",
            "quotaDimensions": "[MaxDepth]", "quotaValue": "250000" } ] },
        { "@type": "type.googleapis.com/google.rpc.RetryInfo", "retryDelay": "56s" }
      ]
    }
  }
}
```

Aggregate across all 45 captures (2026-08-01 → 2026-08-05):

| `quotaId` | occurrences | `retryDelay` range | `classify429` result |
|---|---|---|---|
| `GenerateContentInputTokensPerModelPerMinute-FreeTier` | 26 | 0s – 59s | `quota_exhausted` |
| `GenerateRequestsPerMinutePerProjectPerModel-FreeTier` | 15 | 14s – 58s | `quota_exhausted` |
| `GenerateRequestsPerDayPerProjectPerModel-FreeTier` | 4 | 29s – 49s | `quota_exhausted` |

## Additional Context

- Capture method: OmniRoute's own native `POST /api/logs/detail {"enabled": true}` pipeline-capture endpoint, enabled temporarily on a self-hosted instance specifically to gather this evidence, disabled again immediately after (this endpoint logs full raw request/response bodies — not something to leave on).
- Code pointers: `src/shared/utils/classify429.ts` (`QUOTA_PATTERNS`, `looksLikeQuotaExhausted`, `classify429FromError`), `src/sse/handlers/chat.ts:1192-1196` (`cooldownByKind`), `open-sse/services/accountFallback.ts` (`shouldMarkAccountExhaustedFrom429`, `hasPerModelQuota`, `lockModelIfPerModelQuota`).
- `hasPerModelQuota('gemini')` is already `true`, so this does **not** poison the whole connection — it's scoped to a per-model lock. The bug is purely the *duration* fed into that lock, not its blast radius.
- Every retryDelay observed across all 45 samples is under 60 seconds — none of these are genuine day-scale exhaustions in practice, despite one `quotaId` literally containing "PerDay".

## Validation Plan

- Extend `tests/unit/classify429.test.ts` with a case using the real Gemini `generativelanguage.googleapis.com` body shape (see Error Logs above) asserting it does **not** unconditionally return `"quota_exhausted"` when `RetryInfo.retryDelay` is short — mirroring the existing "Google RESOURCE_EXHAUSTED without a reset remains a rate limit" test for the other Google message shape.
- `node --import tsx/esm --test tests/unit/classify429.test.ts`
- `node --import tsx/esm --test tests/unit/circuit-breaker-failure-kind.test.ts` (cooldown-by-kind mapping)

shixi-li:
## Summary

- Read the upstream's own declared retry window in `classify429()` before the keyword scan: a sub-hour window now yields `rate_limit` instead of the flat 1h `quota_exhausted` bucket.
- Read it from **both** carriers, because the two live call paths deliver different shapes: the structural `google.rpc.RetryInfo` detail (account-fallback classifies the parsed body) and Google's `"Please retry in 38.9s"` phrasing (`chat.ts` classifies `result.rawMessage`, which `parseUpstreamError` has already reduced to `error.message` text).
- Keep terminal credits/billing signals terminal: `INSUFFICIENT_G1_CREDITS_BALANCE`, `out of credits`, `billing cap`, `individual quota reached` / `enable overages`, Cloudflare's `daily free allocation` and `insufficient quota` are checked first and never downgraded by a retry hint.
- Encode the captured payloads from #9504 as regression tests, including the production message-only shape.

Fixes #9504.

## Root cause

Gemini's free-tier 429s always open with the same boilerplate ("You exceeded your current quota, please check your plan and billing details… Quota exceeded for metric…"), which unconditionally trips three generic `QUOTA_PATTERNS` (`/quota.*exceed/i`, `/exceed.*quota/i`, `/plan.*limit/i`) — whether the violated quota is a real daily cap or a per-minute throttle that Google's own `RetryInfo.retryDelay` says clears in 14-59s. Even the `…PerDay…`-named `quotaId` ships ~30-50s delays, so the quota name cannot discriminate; the declared window is the only authoritative signal, and the classifier never read it.

This follows the module's existing evidence-gated precedent: `/resource has been exhausted.*reset after/i` already requires an explicit reset-window qualifier before long-locking.

## Scope, stated precisely

Connection-level cooldown for these 429s was already correct — #7360 and #7940 route Gemini's retry hint into `rateLimitedUntil`. What this changes is the `FailureKind` that reaches the `cooldownByKind` consumers (`chat.ts`, `chatHelpers.ts`, `accountFallback.ts`), where a per-minute throttle currently books a 1-hour lock, a 60-360x overshoot of the upstream's own reset.

Two implementation notes a reviewer will want:

- **The delay grammar is duplicated, not imported.** `open-sse/services/retryAfterJson.ts` owns the same grammar (#7940), but `accountFallback.ts` already imports this module, so importing back would close a dependency cycle (`npm run check:cycles` is a gate). The local `parseDelaySeconds` mirrors `parseDelayString` (ms/s/m/h + bare seconds, trimmed) and the `@type` check uses the same `.includes("RetryInfo")` form, so the repo's short-`@type` fixtures match. The comment on it says to keep the two in step.
- **The override is bounded.** Terminal credits/billing signals are checked before the retry hint, so a mechanically-attached short delay can never turn a credits-exhausted account into a 60s retry loop.
- **The threshold is the bucket it decides against, and that leaves a band.** `QUOTA_SCALE_RETRY_DELAY_SECONDS = 3600` mirrors the `quota_exhausted` cooldown, so a declared window between 60s and 1h now books the 60s `rate_limit` bucket and will probe before the upstream window closes. Every payload captured in #9504 is 0-59s, so nothing observed lands in that band; if you would rather cut it tighter (or add a third bucket keyed on the declared window), say so and I will adjust.

## Validation

- TDD (Hard Rule #18): the new tests fail against the base implementation (`git show origin/release/v3.8.50:src/shared/utils/classify429.ts` in place → 4 failures, including the production message-only case) and pass with the fix. `node --import tsx/esm --test tests/unit/classify429.test.ts`: 29/29.
- New coverage: the three captured `quotaId` variants, a fractional `38.922534355s` delay, the **production shape** (`classify429FromError({ status: 429, message })` with details already stripped), the bare `google.rpc.RetryInfo` `@type` plus `ms`/`m`/`h` units, a 3h long-window control, a no-hint-at-all control, a `retryDelay`-outside-RetryInfo control, and four terminal-credits payloads carrying a 20s hint.
- Adjacent suites: `rate-limit-enhanced` 38/38, `account-fallback-anthropic-quota` 13/13, `8247-accountfallback-model-unhealthy` 2/2.
- Gates: scoped `eslint` clean, `typecheck:core` clean, `check:any-budget:t11` pass, `check:cycles` pass (no cycle introduced). Repo-wide `npm run lint` carries 2 pre-existing `no-explicit-any` errors in the untouched `tests/unit/issue-9407-*.test.ts` (from f1ea77f).
- `npm run test:unit:ci` on this machine: 27,580 / 27,628 pass with ~31 failures, none of them in a suite this change touches. They cluster in network- and host-dependent areas: the vision-bridge suites (`callVisionModel*`, `VB-S*`), the gemini-web/claude-web validators, `ops-scripts`, `shared/machineId` (hostname/`execSync`), and `authz/probe-9033-repro` + `8189-classifier-compat-auto-narrow`. I verified representatives directly against the base `classify429.ts` (restored via `git show`, never `git stash`): `8189-classifier-compat-auto-narrow` and `authz/probe-9033-repro` fail identically 4/8 either way, and `chat-cooldown-aware-retry` — the one failure whose subject matter is adjacent to this change — passes 7/7 in isolation both with and without the fix, so its full-suite failure is concurrency contention at `--test-concurrency=4`. The full baseline comparison is now done and is the cleanest evidence available: base `classify429.ts` → 27,577 pass / 33 fail; with this change → 27,581 pass / 29 fail. Diffing the failing-test **sets**, **zero tests fail only with this change**, and the four that fail only on the base are exactly the new tests in this PR — the TDD red/green reproduced at full-suite scale. Everything else is common to both runs, i.e. pre-existing local flake (the count itself drifts 29/30/33 across runs).


shixi-li:
Following up on the CI signal, since this PR shows red checks.

**The failing checks are not caused by this change** — they are the repo's current CI state on unrelated branches too.

`Unit Tests fast-path` failures on this PR:

| Shard | Failing test |
| --- | --- |
| 1/4 | `8522: innocent PR (base already over frozen cap) must NOT be a violation` |
| 2/4 | `D1: blacklisted IP is blocked on a DIRECT connection`, `D2: persisted config written after first load is honored WITHOUT restart` |
| 4/4 | `issue #8189: 'always' mode is unaffected — every Claude-format request still short-circuits` |

The same jobs, with the same test names, fail on [run 31018108159](https://github.com/diegosouzapw/OmniRoute/actions/runs/31018108159) for `feat/warmup-scheduler-8848` — an unrelated branch that does not touch `classify429.ts` (that run additionally hits the vision-bridge `VB-S12` failure I also see locally).

Locally I confirmed the same thing per-file, restoring the base `classify429.ts` with `git show origin/release/v3.8.50:...` (never `git stash`, per the repo rule):

- `authz/probe-9033-repro` and `8189-classifier-compat-auto-narrow`: fail identically 4/8 with and without this change.
- `chat-cooldown-aware-retry` — the one failure whose subject is adjacent to this change — passes 7/7 in isolation both ways, so its full-suite appearance is contention at `--test-concurrency=4`.
- Full `npm run test:unit:ci`: 27,581 / 27,628 pass with this change; the base run for comparison is finishing now and I will post the delta if it differs.

`No new ESLint warnings` and `Fast Quality Gates` also fail on that same unrelated run. Note that the ESLint-warning job carries `continue-on-error` for fork PRs (`quality.yml:441`), so a fork submission is expected to show it red.

Happy to rebase onto a newer base or adjust anything if you would rather see these green first.

shixi-li:
Baseline comparison I promised above is done, on this machine, `npm run test:unit:ci` both ways:

| Run | Pass | Fail |
| --- | --- | --- |
| base `classify429.ts` (restored via `git show`) | 27,577 | 33 |
| this change | 27,581 | 29 |

Diffing the failing-test **sets** rather than the counts:

- **Zero tests fail only with this change** — no regression anywhere in the suite.
- The four that fail only on the base are exactly this PR's new tests (`classify429: Gemini free-tier 429 with a short RetryInfo window…`, `…string bodies too`, `…bare RetryInfo @type and non-second units`, `classify429FromError: the production message-only shape…`), i.e. the TDD red/green reproduced at full-suite scale.
- Every other failure is present in both runs, and the total drifts across runs (29 / 30 / 33), which matches the concurrency-contention picture described in my previous comment.

shixi-li:
Update after your base merge (thanks for it) — CI on `885f2939`:

`8522: innocent PR (base already over frozen cap)` is **gone**, cleared by the `#9355` file-size fix that came in with the merge. The remaining reds are the flaky set, and the shard membership moves between runs, which is itself the tell:

| Check | Failing test | In my local base-only run? |
| --- | --- | --- |
| Unit Tests 2/4 | `D1: blacklisted IP is blocked on a DIRECT connection`, `D2: persisted config written after first load is honored WITHOUT restart` | yes |
| Unit Tests 3/4 | `callVisionModel returns description on success`, `execute (non-stream) parses bare JSON reply into OpenAI tool_calls` | yes |
| Unit Tests 4/4 | `issue #8189: 'always' mode is unaffected` | yes |

Every one of those appears in the run I did with the **unmodified** `classify429.ts` in place, so none is introduced here. As posted above, the set-level diff of the two full runs shows zero tests failing only with this change, and the four failing only on base are this PR's own new tests.

Verified locally on the merged head `885f2939`: `classify429` 29/29, scoped `eslint` clean, `typecheck:core` clean. The merge touches neither `classify429.ts` nor its consumers (`chat.ts`, `chatHelpers.ts`, `accountFallback.ts`, `error.ts`).

Nothing needed from me unless you would like a change — happy to adjust the 60s–1h threshold band or anything else.

PR #8839, opened 2026-07-28
[v3.8.50] docs: slim AGENTS.md

MumuTW:
## Summary
- reduce the root AGENTS.md from 600 to 117 lines
- retain repository-wide setup, validation, security, review, and upstream-contribution instructions
- replace per-module architecture prose and mutable counts with a concise repository map and canonical deep-dive links
- stop the documentation-count gate from requiring mutable counts in the always-loaded agent guide

## Validation
- node --import tsx/esm --test tests/unit/check-docs-counts-sync.test.ts
- npm run check:docs-all
- npm run lint

Closes #8527

PR #9515, opened 2026-08-05
feat(radar): flag-gated signed free-model catalog overlay

diegosouzapw:
## Summary

Adds the **Radar** module: an optional, flag-gated client that overlays a signed, freshly-curated free-model catalog on top of the release baseline. The free-tier landscape moves faster than the release cadence — providers add, shrink and discontinue free quotas between versions, and the static catalog can only be refreshed when a new version ships.

**Nothing that is free today stops being free.** Radar never removes or paywalls a baseline entry: it refreshes limit/status fields at read time and can layer in newly-discovered free models between releases. The baseline catalog is never mutated on disk.

## Behaviour with the flag off (the default)

`RADAR_ENABLED` defaults to `"false"`, and with it off the surface does not exist:

- the three `/api/radar/*` routes return `404`
- the dashboard screens render `notFound()`
- `getRadarCatalog()` returns the untouched baseline, every entry tagged `origin: "baseline"`, and never reads the cache
- `syncRadar()` returns `{ status: "disabled" }` without touching `fetch`

`tests/unit/radar-inertia.test.ts` asserts this "zero delta" claim in one place, including that `computeFreeModelTotals()` returns the same numbers as before.

## Two independent gates

Turning the flag on unlocks only the **screens**. Syncing the catalog needs a second, separate opt-in stored in `radar_settings.opt_in`. The privacy promise is that nothing is ever uploaded: the client performs a `GET` and downloads signed metadata, nothing more.

## Security model

- **Ed25519 over exact bytes** — the signature is verified against the exact response bytes received over the wire; the payload is never re-serialized before verification, and verification failure aborts the sync before the body is parsed or cached.
- **Pinned public key** in `src/lib/radar/pinnedKeys.ts` (an array, so a key can be prepended ahead of a rotation).
- **Version floor** — a feed that is not strictly newer than the cached one is rejected, so a misconfigured or compromised endpoint cannot roll a client back to an older payload.
- **Schema validation** after signature verification, and defensive re-validation on every read — a corrupted cache row falls back to the baseline rather than being served.
- The supporter key is stored **encrypted at rest** with the same AES-256-GCM mechanism used for provider credentials, and is never echoed back by the settings route (only a masked form).
- The local routes **never proxy** the feed service — they only read the local cache.

## Fork-friendly

`RADAR_FEED_URL` and `RADAR_FEED_PUBKEY` let a fork or self-hoster point the client at their own signed feed satisfying `src/lib/radar/feedSchema.ts`. Documented in `docs/frameworks/RADAR.md`.

## Read-time overlay

`applyFeed()` merges the cached feed over the static baseline at read time, honouring four rules: a local override always wins; a feed `enabled:false` disables an entry but records the provenance so the UI can explain why; a user-created entry absent from the feed survives intact; and a user's deletion tombstone is never resurrected by a later feed.

## Test plan

- `tests/unit/radar-*.test.ts` — **95 tests, all passing**: signature verification (valid/tampered/malformed, never throws), version floor including numeric comparison (`.10 > .9`), flag-off and opt-out both proven to make no network call, schema rejection, sanitized errors with no stack traces, encryption-at-rest asserted by reading the raw DB column, the four merge rules, route 404s, and key masking.
- `npm run check:cycles` — no cycles.
- `npm run check:docs-all` — doc-links and fabricated-claim gates pass.
- `npx eslint` on every file the branch adds — clean.
- `npm run typecheck:noimplicit:core` — 6 errors, all pre-existing on the base (`open-sse/utils/usageTracking.ts`, `src/shared/services/cliRuntime.ts`); none in Radar files, verified against `origin/release/v3.8.50`.

Migration `134_radar_cache_settings.sql` adds two single-row tables (`radar_feed_cache`, `radar_settings`); `localDb.ts` receives re-exports only.


diegosouzapw:
### Babysit summary

**Three real defects fixed, all with the cause addressed rather than the symptom.**

| # | Problem | Fix | Commit |
|---|---------|-----|--------|
| 1 | PR was `CONFLICTING`, so `Quality Gates` never fired at all (GitHub cannot compute a merge-ref with a conflict, so the `pull_request` workflow is never created — `gh run list` came back empty, which reads like runner congestion but is not) | Merged the base; the conflict was `AGENTS.md`, rewritten wholesale by the slim-down (#8839). Took the base's version and re-applied this PR's single line — the `RADAR.md` reference — into the new "Platform features" row. i18n auto-merged with no loss. | `104f8c9ab3` |
| 2 | Migration number collision: the base landed `134_proxy_logs_egress_ip` while this branch carried `134_radar_cache_settings`; the runner rejects duplicate numeric prefixes | Renamed to `135`, updated both references (SQL header + doc). No retroactive `isSchemaAlreadyApplied` guard is needed — this migration has never been applied to a real database, since the PR is unmerged. `check:migration-numbering` reports 0 duplicates. | `f6edae016c` |
| 3 | `No new ESLint warnings` red — misleading name: ESLint itself was 0 warnings / 0 errors. The metric riding in the same job was `i18nUiCoverage.pct: 98.4 < baseline 99`, because this PR adds 60 UI strings that existed only in `en`/`pt-BR` | Translated them into all 41 remaining locales (2460 strings). Scoped deliberately to this PR's namespaces (`radarPage`, `radarSetupPage`, `sidebar.radar*`) instead of a bulk `i18n:sync-ui`, which would have pulled ~978 unrelated pending keys into the diff — and would not have fixed it anyway, since the coverage formula is `(present − placeholder) / total_en` and `__MISSING__:` sentinels are subtracted, not counted. | `18e56fca79` |

Verification on the i18n work: all 43 locale files still parse, zero `__MISSING__` left on radar keys, and **all 1682 translated strings checked for placeholder fidelity — zero divergences** (a dropped `{count}` would break the UI in a locale nobody tests). The local gate reports `PASS — all 42 locale(s)`, every one at 99.0%+. `quality-baseline.json` was not touched.

**Remaining reds are inherited from the base, not introduced here.**

The four `Unit Tests fast-path` shards fail in `deepseek-web-tools-execute-2820`, `resolveProjectRoot` (`repro-8956`), `#8189 classifier`, `VB-S12` and `D1` — none of which this branch touches. Direct proof for the clearest case: `tests/unit/deepseek-web-tools-execute-2820.test.ts` fails identically on the **base checkout with zero lines from this branch** (`actual: 'stop', expected: 'tool_calls'`), so it is deterministic and pre-existing, not a flake. Open PRs from other sessions (`fix/deepseek-thinking-efforts`, `fix/classify429-gemini-retry-delay`) fail the same tests.

Fixing those inside a catalog-overlay PR would inflate its scope, and masking them is not on the table. They need either a `/sweep-reds` pass on `release/v3.8.50` or a dedicated PR.

**Gate status:** 14 green — including `Vitest`, `Fast Quality Gates`, `Docs Gates`, `Merge integrity`, `No new ESLint warnings`, CodeQL and both semgrep runs. 4 red (all base-red). `Build (advisory)` still running.

Not merged — handing off for human review.

PR #9538, opened 2026-08-05
feat(opencode-plugin): add visibleModels/hiddenModels allowlist/blocklist (#9473)

### LINKED ISSUE #9473: feat(providers): Model allowlist/whitelist for opencode-plugin
### Problem

The plugin pulls the full live catalog from /v1/models (600+ models on a typical OmniRoute instance). The OpenCode model picker becomes completely unusable — scrolling through hundreds of entries to find the ~30 models actually used in a project is impractical.

The OpenCode Windows desktop app does offer per-model enable/disable toggles in its settings UI, which is a partial mitigation. However:

- CLI/TUI users (the primary OpenCode interface) have no such UI — they see all 600+ models with no way to hide them.
- Even in the desktop app, toggling off 570+ models one by one is tedious and not persistent across config resets.

### Current state

- Plugin version: 0.2.1 (latest on npm)
- The features block has no allowlist/whitelist option (Zod schema is strict() — only the documented flags exist)
- features.usableOnly filters to healthy providers but still leaves hundreds of models
- The plugin's config hook unconditionally overwrites config.provider[<id>] (line 17146 in dist), so a static provider.<id> block in opencode.json with a curated model list gets clobbered

### Workaround (fragile)

A second curator plugin registered after the omniroute plugin can re-overwrite config.provider[<id>] with a filtered model set. This works but requires:

1. Understanding OpenCode's config-hook merge semantics (nested mutation breaks the provider; full entry replacement is required)
2. Maintaining a separate plugin file with a hardcoded allowlist
3. Knowing that plugin array order matters

### Suggested solution

Add a features.visibleModels (or features.modelAllowlist) option — a string[] of model IDs. When set, the provider hook and config shim would only emit models whose IDs are in the list.

Alternatively, a features.hiddenModels blocklist for the inverse use case.

### Additional context

- OpenCode version: 1.18.13
- The providerId in the plugin log shows opencode-omniroute-live (prefixed by opencode), but the config uses omniroute-live — the allowlist should match against the raw model IDs as returned by /v1/models (e.g. cx/gpt-5.6-sol, cp/cline-pass/kimi-k3)
- Combos (combo/<slug>) should also be filterable — either included automatically when all member models are in the allowlist, or via a separate visibleCombos option

diegosouzapw:
## Summary

Adds `features.visibleModels` (allowlist) and `features.hiddenModels` (blocklist) to the `@omniroute/opencode-plugin` features block, letting operators curate the OpenCode TUI/CLI model picker from 600+ catalog entries down to a fixed set of IDs that persists in `opencode.json` across config resets.

### What changed

- **Schema**: Added `visibleModels: string[]` and `hiddenModels: string[]` to the Zod `featuresSchema` (strict mode preserved).
- **Filter helpers**: Three new pure exported functions:
  - `compileModelListFilter()` — pre-compiles a string[] into exact/suffix sets for O(1) lookups.
  - `passesModelAllowlist()` — decides whether a raw model ID passes the allowlist/blocklist. Bare suffixes (e.g. `claude-opus-4-7`) match any `{prefix}/claude-opus-4-7`. Deny wins over allow.
  - `passesComboAllowlist()` — decides whether a combo passes the visible filter (at least one member must match).
- **Hook wiring**: Both the dynamic `provider.models()` hook and the static `buildStaticProviderEntry()` apply the filter after `usableOnly` (all filters AND together).
- **Tests**: 37 new tests in `tests/model-allowlist.test.ts` covering pure helpers, schema parsing, and static entry integration.
- **README**: Feature table, flag table, and curating example section.

### Composition rules

- Unset or empty = no filter (current behavior).
- All filters AND together: usableOnly AND visibleModels AND NOT hiddenModels.
- Blocklist wins over allowlist (deny takes precedence).
- Bare suffixes match any provider prefix.

Closes #9473


Bartuz:
why is it only for opencode plugin and not global /v1/models/ level? AllowList would be helpful at global level. Claude code is also "unusable" with a great list of models

PR #9540, opened 2026-08-05
feat(providers): warm catalog startup from disk snapshot, parallel refresh (opencode-plugin) (#9490)

### LINKED ISSUE #9490: feat(providers): warm catalog startup from disk snapshot, parallel refresh
## Problem / Use Case

Each fresh opencode process rebuilds the catalog (models, combos, pricing, fetched sequentially, ~30s) before the provider registers, even though the plugin writes a ~5MB snapshot every refresh. That snapshot is only read when `/v1/models` fails, so healthy startups re-fetch.

## Proposed Solution

Serve the last snapshot so the provider registers without a network wait. Refresh in parallel, then update cache and snapshot.

## Alternatives Considered

- Parallelize the fetches without the snapshot. Faster, but every new process still blocks on the network.
- Disable enrichment to drop the largest fetch. Cuts load but loses model names and pricing in the picker.

## Acceptance Criteria

- `opencode run` and `opencode models` return in ~1-2s on a warm gateway.
- Parallel refresh keeps today's fallbacks (models -> snapshot, combos -> empty, enrichment -> raw ids).
- Fingerprint mismatch forces a fresh fetch; failed refresh keeps the snapshot.

## Area

CLI Tools Integration

## Related Provider(s)

opencode

## Additional Context

Gateway-side counterpart: #8697. Measured ~30s cold load (3079 models, ~16k enrichment entries).

## Expected Test Plan

Unit tests for the snapshot reader and parallel fetch path.

diegosouzapw:
## Summary

Implements #9490 — the opencode plugin now serves the last disk snapshot at cold startup so the provider registers in ~1-2s instead of blocking ~30s on sequential catalog fetches.

### Changes
- **Warm startup**: on a cache miss, before any network fetch, the existing `diskSnapshotReader` is called. If a fingerprint-matching snapshot with models exists, a static block is published immediately via the same `buildStaticProviderEntry()` call used by the live path.
- **Parallel refresh**: the six sequential `await`s (models, combos, autoCombos, enrichment, compression, connections) are replaced with `Promise.allSettled` fan-out, keeping per-endpoint soft-fail semantics 1:1.
- **Detached refresh**: when a warm snapshot was served, the hook returns immediately and the refresh runs detached. When no snapshot exists (cold first run), the awaited behavior is preserved.
- **Failed refresh keeps the snapshot**: `modelsFetchOk` gate unchanged — a failed refresh never overwrites the snapshot.

### Tests (TDD — Hard Rule #18)
- 12 new tests in `tests/warm-startup.test.ts`: warm startup, fingerprint mismatch, parallel refresh, failed refresh keeps snapshot, concurrent fan-out, soft-fail parity, in-flight dedup, diskCache=false opt-out, empty snapshot skip, snapshot age logging.
- All 290 plugin tests pass (287 existing + 3 updated config-shim tests).

Closes #9490
Refs #8697


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

Issue #9541, opened 2026-08-05
flaky: fast-path unit shards fail ~45% under load on ip-filter (D1/D2) + visionBridge (VB-S*) signature


diegosouzapw:
## Summary

The `Unit Tests fast-path (N/4)` shards (quality.yml, `test:unit:ci:shard`) fail intermittently — but at very high rate under load — on the SAME test signature, across PRs whose diffs cannot possibly affect these tests (including a docs-only PR):

- `tests/unit/authz/probe-9033-repro.test.ts` — `D1: blacklisted IP is blocked on a DIRECT connection`, `D2: persisted config written after first load is honored WITHOUT restart`, `Bonus: ipFilterModeSchema accepts whitelist-priority`
- `tests/unit/guardrails/visionBridge.test.ts` — `VB-S12`, `VB-S12b`, `VB-S01`, `VB-S13`, `VB-S07` (cascade)

## Evidence (2026-08-05)

During the base-red drain campaign, ~49 PRs were update-branched against the repaired `release/v3.8.50` tip and re-ran CI simultaneously. Result: **~45% of fast-path shard runs failed with this exact signature, and `gh run rerun --failed` re-flaked at a similar rate while the runner queue stayed saturated.** Examples: runs 31024823977 (PR #9508 — docs-only!), 31024846639 (#9483), 31024853385 (#9482), 31024892883 (#9449), 31024822682 (#9510).

Meanwhile the same tests pass on the release tip's `ci.yml` `test-unit` job (different sharding/concurrency) and pass locally in isolation with the #9385 fix in place.

## Hypothesis

Test-isolation defect exposed by shard co-scheduling under load: the ip-filter tests bind server state / read persisted config (DATA_DIR), and the visionBridge suite appears to fail as a cascade in the same process. Likely port collision or shared DATA_DIR/state between shard neighbors (`--test-concurrency=4`), aggravated when runners are slow. Note `[DDG-CB] Circuit breaker opened after 5 consecutive failures` appears in failing logs — network calls escaping mocks under load may be involved.

## Impact

Every code PR has ~even odds of a spurious red on fast-path shards whenever the runner pool is busy — which erodes exactly the "PR born green" guarantee the base-green work just restored.

## Suggested angle

1. Reproduce with the exact CI invocation (`TEST_SHARD=2/4 npm run test:unit:ci:shard`) under parallel load.
[...]

Issue #7859, opened 2026-07-20
fix(providers): Gemini Web connection test fails — 'Redirect blocked' 302 despite validator treating 302 as valid


height229:
### OmniRoute Version

v3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 26.04

### Provider(s) Involved

Gemini Web

### Model(s) Involved

all

### Client Tool

None

### Description

When adding a token to the `Gemini Web (Free)` provider i get `Redirect blocked for GET https://gemini.google.com/app (302)` as a response when testing/verifying the provider.

### Steps to Reproduce

1. Go to the Providers tab.
2. Search for `Gemini Web (Free)`
3. Press on Add Connection
4. Enter your __Secure-1PSID token
5. Click Save

### Expected Behavior

It should work and the `Gemini Web (Free)` provider should be added.

### Actual Behavior

Doesnt let me use the `Gemini Web (Free)` provider.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

diegosouzapw:
Confirmed and root-caused. Labelled `bug` / `providers`. Thanks for the precise repro — the cause is not your cookie, and the connection would fail this way for every user.

**The validator already intends to accept your 302.** `src/lib/providers/validation/webProvidersB.ts:229-233`:

```ts
// 200/302 = valid, anything < 500 that isn't auth failure is acceptable
if (response.status < 500) {
  return { valid: true, error: null };
}
```

That branch is unreachable. `validationRead` runs on the `validationRead` preset, which sets `allowRedirect: false` (`src/shared/network/safeOutboundFetch.ts:51`), and `safeOutboundFetch.ts:311` **throws** `REDIRECT_BLOCKED` on any 3xx before a status is ever returned. The throw lands in the `catch` at the bottom of the validator and is rendered verbatim as the message you saw.

So the comment documents an intent the transport layer silently overrides. `https://gemini.google.com/app` answers 302 for every session — valid or not — which makes the Gemini Web provider unusable for everyone, not misconfigured for you.

**Why this matters beyond this one provider.** This is the *fourth* time this class has surfaced, and each previous fix was scoped narrowly to the provider that reported it:

- **#3288 / #3758** — blocked redirects to public hosts were mislabelled as SSRF; fixed in `isSecurityBlockError()` by checking whether the redirect target is private.
[...]

Issue #4012, opened 2026-06-16
[BUG] Nvidia NIM via Omniroute fails to process image inputs


daniij:
### OmniRoute Version

3.8.26

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

v24.15.0

### Provider(s) Involved

Nvidia

### Model(s) Involved

any vision capable

### Client Tool

any

### Description

When routing vision requests to Nvidia NIM through Omniroute, the model executes successfully but returns a stub response stating that the image data is missing or unavailable.

Direct calls to the Nvidia NIM API handle the exact same payload correctly, which indicates that Omniroute is dropping or incorrectly stripping the image_url data from the messages array during transit.

Also some free providers dont work:
[provider removed at its operator's request] (Free): Returns [502]: fetch failed
Veo AI Free: Returns [502]: terminated
Chipotle Pepper AI (Free): Returns [502]: WS connect timeout
DuckDuckGo AI Chat: Returns [400]: DuckDuckGo AI Chat error: ERR_BAD_REQUEST

### Steps to Reproduce

1.Send a standard OpenAI-compatible Vision request to an Nvidia NIM model (e.g., nvidia/google/diffusiongemma-26b-a4b-it) via Omniroute.
2.Include a remote image URL in the messages block.
3.Observe the generated response text.

Code Snippet Used for Testing

```python
import json
import time
from pathlib import Path
from openai import OpenAI

IMAGE_URL = "https://s0.rbk.ru/v6_top_pics/media/img/5/46/756038770746465.jpg"
PROMPT = "Describe this image in one or two sentences."

VISION_MESSAGES = [
    {
        "role": "user",
        "content": [
            {"type": "image_url", "image_url": {"url": IMAGE_URL}},
            {"type": "text", "text": PROMPT},
        ],
    }
]

# Example calling via Omniroute
client = OpenAI(api_key="OMNIROUTE_API_KEY", base_url="OMNIROUTE_BASE_URL")
resp = client.chat.completions.create(
    model="nvidia/google/diffusiongemma-26b-a4b-it",
    messages=VISION_MESSAGES,
    max_tokens=256,
    temperature=0.0,
)
print(resp.choices[0].message.content)
``` 

### Expected Behavior

The model should receive the image, process it, and provide a valid visual description.

### Actual Behavior

[1/2] Direct Nvidia API / google/diffusiongemma-26b-a4b-it ... ok (37.0s)
[...]

PR #9529, opened 2026-08-05
fix(quality): 2 production bugs + 24 unit base-reds + measured gate ceilings

diegosouzapw:
## Summary

Quality-gate hardening pass on the release tip: 2 real production bugs fixed with TDD, 24 unit base-reds resolved (7 root causes, all landed 04/08), 2 gate ceilings recalibrated from measurement, and the test-masking gate extended with a verified deletion form.

## Production bugs (both TDD: failing test → fix → green)

**1. `rateLimitManager`: Bottleneck reservoir refresh dies permanently (`432b3140c2`)**
Bottleneck 2.19.5's `LocalDatastore._startHeartbeat()` clears the heartbeat interval without nulling the reference, so the first `limiter.updateSettings()` kills the periodic reservoir refill forever. Any auto-enrolled apikey connection that accumulates 60 requests freezes its queue ~120s, the watchdog fabricates a synthetic 502 (`RATE_LIMIT_QUEUE_WEDGED`), the connection cools down and drops out of the weighted pool — turning 70/30 distributions into ~50/50. New helper `applyLimiterSettings()` re-arms the heartbeat after every settings update (5 call sites). Regression test: `tests/unit/ratelimit-reservoir-refresh.test.ts` (hangs without the fix, resolves in ~1.1s with it). E2E proof: `tests/integration/combo-matrix/weighted.test.ts` passes untouched, with zero wedge artifacts in a 20k-line run log. Rate-limit family: 42/42.

**2. `visionBridge`: credential deps dropped at reroute call site (`c109d330d3`)**
`getBestVisionModel({ fixedModel })` was called without the `deps` argument while the two adjacent credential checks in the same function inject it — the router always hit the real DB and ignored DI overrides (regression from #8430). One-line fix reusing the already-resolved `checkCreds`. Before: 6 failing (VB-S*); after: 25/25, vision family 71/71.

## Unit base-reds (24 observed, 7 root causes — all landed 04/08)

The dominant systemic pattern (5 of 7 causes): a behavior change merges updating only part of the sibling tests that pin the old semantics.

| Cause | Fix commit |
|---|---|
| #9276 classifier `always`-mode — 1 of 2 sibling tests updated | `0f9db4789d` |
[...]

PR #9546, opened 2026-08-05
fix(guardrails): vision-bridge DI gap + capability-gated stub + stale 7859 mock (#9541)

diegosouzapw:
Fixes the three deterministic `Unit Tests fast-path` shard reds hitting every code PR since 2026-08-04 (Refs #9541 — the load-dependent D1/D2 half was fixed by #9385):

1. **DI gap**: `visionBridge.ts` didn't thread its injected `hasUsableCredentials` into `getBestVisionModel()` — #8430's fixedModel validation ran the real DB check under test DI, so CI's empty credential store silently disabled the reroute path (VB-S12/S01/… red on every PR, docs-only included).
2. **#4012 × #8430 composition**: the all-describes-failed stub now only replaces images when the upstream is confirmed NOT vision-capable — a vision-capable upstream keeps its raw images (VB-S03).
3. **Stale mock**: the #7859 test's public-redirect mock used `ServiceLogin`, which #9407 redefined as expired-session — moved to a non-ServiceLogin public target; both contracts keep dedicated coverage.

Validation (CI loader set): visionBridge 25/25 · 4012+8430 5/5 · 7859+9407 12/12 · typecheck clean. This PR's own fast-path shards double as the live proof.


diegosouzapw:
Superseded by #9529 (merged): it shipped the identical vision-bridge DI fix, the test realignments (VB-S03/7859) and the provider-count docs sync — this branch's diff against `release/v3.8.50` is now empty. Root-cause analysis lives in #9541.

PR #9572, opened 2026-08-06
fix(providers): reject the dashboard password as a connection API key

HouMinXi:
> [!NOTE]
> base-red inherited: #9298. All three red checks on this PR reproduce on the
> base and none touches the two files here.
>
> - `Fast Quality Gates` fails on `check:pack-policy` over
>   `scripts/build/fixPlaywrightAndroid.mjs`, added upstream by #8922. This PR
>   does not touch `package.json` or `.npmignore`.
> - `Unit Tests fast-path (4/4)`: a local full run on this branch gives 94
>   failures, 43 of which trace to a single upstream defect.
>   `open-sse/handlers/chatCore.ts:4690` calls `mergeResponseToolNameMap`,
>   whose import was removed by #9236 along with only one of its two call
>   sites. Reported separately rather than fixed here.
> - `CodeQL alerts ratchet` counts open alerts repository-wide, not per diff.
>   The single alert is `js/incomplete-url-substring-sanitization`; this PR
>   adds no URL handling at all.

A browser autofilled the dashboard's management password into a connection's
API-key field. The resulting credential authenticates against nothing, so
every request routed through that connection came back 401, and because the
field looks like any other password input the same autofill fired again
while the connection was being repaired by hand.

createProviderConnection and updateProviderConnection now refuse a value
that matches the current dashboard password, checked with bcrypt against the
stored hash. Twenty routes create or update connections and all of them
funnel through these two functions, so one check covers every entry point
including a future one. Update checks only the incoming value, never the
merged one, so a connection that already holds the password stays editable
for repair. An unreadable settings row or a throwing bcrypt call logs and
allows, so a guard against one specific mistake can't become a way to lock
out every connection write.

The comparison runs against both the trimmed and untrimmed incoming value.
Neither the login route nor the set-password route trims when the password
is set, so a password carrying leading or trailing whitespace would
otherwise slip past a trimmed-only check.

Scope is deliberately narrow to the two write paths above. The row
re-encryption path and the one-time db.json import are left alone since
neither takes a value an operator just typed, and OAuth tokens are excluded
since they arrive from a provider's token endpoint rather than a form.

Test plan:
- tests/unit/reject-management-password-as-apikey.test.ts (new): create and
  update on both paths, the trimmed/untrimmed comparison, the
  bcrypt-parse-failure log-and-allow branch, and the token-only-update early
  return
- forge review, six rounds, gate clean

### DIFF
--- a/src/lib/db/providers.ts
+++ b/src/lib/db/providers.ts
@@ -71,10 +71,17 @@ export class ManagementPasswordAsCredentialError extends Error {
  * Only an actual match blocks the write. A settings row that cannot be read,
  * or a bcrypt call that throws, logs and allows -- a guard against one specific
  * operator mistake must not become a way to lock out every connection write.
+ *
+ * Deliberately narrower than CONNECTION_CREDENTIAL_FIELDS. The OAuth tokens
+ * arrive from a provider's token endpoint, and the refresh path writes them
+ * back through updateProviderConnection on every renewal, so checking them
+ * would put a bcrypt round on a renewal path to defend a field no autofill
+ * reaches. apiKey is the only credential an operator types into a form.
  */
 async function assertApiKeyIsNotManagementPassword(apiKey: unknown): Promise<void> {
-  const candidate = typeof apiKey === "string" ? apiKey.trim() : "";
-  if (!candidate) return;
+  if (typeof apiKey !== "string") return;
+  const trimmed = apiKey.trim();
+  if (!trimmed) return;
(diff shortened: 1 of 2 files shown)


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

PR #9618, opened 2026-08-06
fix(db): resolve CCR migration version collision

fenix007:
## Summary

- renumber `134_ccr_blocks.sql` to the next free slot, `139_ccr_blocks.sql`
- reconcile databases that already recorded CCR under the legacy `134` slot, freeing `134` for `proxy_logs_egress_ip`
- add an idempotency guard for databases where the CCR table exists without the new marker
- add regression coverage for both upgrade paths

## Root cause

`release/v3.8.50` currently contains both `134_ccr_blocks.sql` and `134_proxy_logs_egress_ip.sql`. The migration runner rejects duplicate numeric prefixes before opening the database, so fresh DB startup and every DB-backed test fail with `Migration version collision detected: version=134`.

The CCR migration landed later and is `CREATE TABLE IF NOT EXISTS`, so it is moved to the next free version while the earlier non-idempotent `ALTER TABLE proxy_logs ADD COLUMN egress_ip` remains at `134`.

## Verification

- 28 targeted migration, fresh-DB, proxy-log, and CCR durable-store tests pass
- ESLint passes for all changed TypeScript files
- Prettier check passes
- `check-migration-numbering` passes with zero duplicate versions

The repository-wide file-size gate still reports six pre-existing release-branch violations; this change does not add or increase any of them.

> **⚠️ base-red inherited**: all 11 failing CI checks are pre-existing on `release/v3.8.50` (missing `antigravityProjectPersistence.ts`, stale docs, changelog fragment format, ESLint baseline, etc.) — none are introduced by this PR. The PR's own 5 changed files pass `check-migration-numbering`, `node --test`, and all targeted migration tests cleanly.

PR #9627, opened 2026-08-06
fix(db): give the two collided migrations their own version slots

### LINKED ISSUE #9602: fix(backend): Migration version collision at 135 and 136 (check:migration-numbering red)
`check:migration-numbering` is red on the current tip (`3c6f71776`): two
version slots each have two physical files.

```
[check-migration-numbering] 2 anomalia(s) de numeração:
  ✗ prefixo de versão duplicado 135: [135_auto_restart_adopted.sql, 135_migrate_model_capability_max_token.sql]
  ✗ prefixo de versão duplicado 136: [136_dario_fallback_backend.sql, 136_radar_cache_settings.sql]
```

Both files at each slot are real, distinct migrations — this isn't the
allowlisted rename-compat case in `KNOWN_DUPLICATE_VERSIONS` (that set is
currently empty; nothing there covers 135 or 136).

Landing order, from `git log -1 --date=iso-strict` on each file (normalized to UTC):

| Slot | File | PR | Landed (UTC) |
|---|---|---|---|
| 135 | `135_migrate_model_capability_max_token.sql` | #8908 | 01:41:38 |
| 136 | `136_radar_cache_settings.sql` | #9515 | 08:58:58 |
| 135 | `135_auto_restart_adopted.sql` | #8523 | 09:05:46 |
| 136 | `136_dario_fallback_backend.sql` | #8523 | 09:05:46 (same commit) |

#8523 landed after both #8908 (135) and #9515 (136) had already claimed those
slots, and merged without renumbering against either. Both of #8523's files
are the ones that should move — house convention here is "whichever lands
second takes the next free number," and 137/138 are both currently open.

Renumbering isn't a pure rename, though: `check-migration-numbering.mjs`'s own
error text points at `isSchemaAlreadyApplied` in `migrationRunner.ts` for a
retroactive guard, since a DB that already ran migration 135 under the old
number needs the runner to still recognize it after the file moves. #8523 is
only hours old, so the compatibility window is probably small, but the guard
is the documented mechanism, not just the file move.

Impact beyond the gate itself: any test that reaches `getDbInstance()` with a
real (non-mocked) migration run hits this, e.g.

```
error: 'Migration version collision detected: version=135 → [auto_restart_adopted, migrate_model_capability_max_token]; version=136 → [dario_fallback_backend, radar_cache_settings]. ...'
  at getMigrationFiles (src/lib/db/migrationRunner.ts:263:11)
  at runMigrations (src/lib/db/migrationRunner.ts:827:53)
  at getDbInstance (src/lib/db/core.ts:1206:3)
```

Reproduced against `upstream/release/v3.8.50` directly (blob hashes below are
identical to the local tree, so this isn't a fork artifact):

```
135_auto_restart_adopted.sql               11f805e7e315124a761d7b8f9e772c38cdc4ee1e
135_migrate_model_capability_max_token.sql be8fbecfda927947e81d49e9082113df1cf692c4
136_dario_fallback_backend.sql             28315188d3f03607efc256cdb5aa3f9f40b85b01
136_radar_cache_settings.sql               42bcec9d5eaee10fff4905c7a015bbb8af8f9a7e
```

Happy to open the renumbering PR if useful — flagging first since it touches
DB migration compatibility and I'd rather have the retroactive-guard shape
confirmed before writing it.

HouMinXi:
Fixes #9602.

Version slots 135 and 136 each held two distinct migrations, which
`getMigrationFiles()` treats as a hard failure — correctly, since the runner
records one row per version and the second file in a slot would silently never
run.

The blast radius was wider than the gate: **every code path that opens a real
database went through it**, so any test reaching `getDbInstance()` threw before
touching its own subject.

```
error: 'Migration version collision detected: version=135 → [auto_restart_adopted, migrate_model_capability_max_token]; version=136 → [dario_fallback_backend, radar_cache_settings]. ...'
  at getMigrationFiles (src/lib/db/migrationRunner.ts:263:11)
  at runMigrations   (src/lib/db/migrationRunner.ts:827:53)
  at getDbInstance   (src/lib/db/core.ts:1206:3)
```

## Which file moves

Landing order (`git log -1 --date=iso-strict` per file, normalized to UTC):

| Slot | File | PR | Landed (UTC) | |
|---|---|---|---|---|
| 135 | `migrate_model_capability_max_token` | #8908 | 01:41:38 | keeps 135 |
| 136 | `radar_cache_settings` | #9515 | 08:58:58 | keeps 136 |
| 135 | `auto_restart_adopted` | #8523 | 09:05:46 | **→ 137** |
| 136 | `dario_fallback_backend` | #8523 | 09:05:46 | **→ 138** |

#8523 landed after both had already claimed those slots, so its two files take
the next free numbers. 137/138 were verified unoccupied. Both renames are pure
(`git diff -M` reports 0 insertions, 0 deletions).

## Upgrade safety

A renumber is not a pure rename: an install that already ran these carries
`version=135` / `version=136` markers, and would otherwise re-execute the SQL
under its new number.

`reconcileRenumberedMigrations()` already exists for exactly this and has ten
prior entries of the same shape (`022→025`, `028→034`, `041→050`, `056→059`, …).
Its guard is a precise fit — it only fires when the old slot is now held by a
*different* migration:

```ts
const hasSourceFile = files.some(
  (file) => file.version === compatibility.fromVersion && file.name !== compatibility.fromName
);
```

So two entries were added to `RENAMED_MIGRATION_COMPATIBILITY`. No logic was
changed; the reconciliation path is untouched.

## Verification

`check:migration-numbering` goes from 2 anomalies to
`OK (135 migrations, 3 gap(s) conhecido(s), 0 duplicata(s) conhecida(s))`.
`check:db-rules` is `OK`, `typecheck:core` exits clean, and `lint` reports
nothing in the files this PR touches.

Both halves were proven by injection rather than by reading:

- **The renumber** — renaming 137 back to 135 takes `db-core-migration.test.ts`
  to `# pass 0` with the collision error; reverting returns 3/3.
- **The compatibility entries** — deleting them fails the new reconcile test
  *alone* (26 pass / 1 fail, `expected: 'migrate_model_capability_max_token'`,
  `actual: 'auto_restart_adopted'`); restoring returns 27/27.
- **The constants spot-check** — transposing `fromVersion`/`toVersion` on the
  135→137 row fails that test alone (7 pass / 1 fail); restoring returns 8/8.

The new reconcile case covers both pairs in a single `runMigrations()` call,
because a real upgrade carries both legacy markers at once — two isolated tests
would not exercise that. It asserts final slot ownership, that the markers move
to 137/138, that the slot owners' SQL **did** run (probe tables 135/136 exist),
and that the renumbered SQL was **not** replayed (probe tables 137/138 absent).
It also captures `console.error` and asserts zero CRITICAL renumbering
warnings, matching the five sibling cases in that file.

53/53 across `db-migration-runner`, `db-migrationrunner-constants-split`,
`db-core-migration` and `check-migration-numbering`.

`testFrozen` for `db-migration-runner.test.ts` moves 1499→1595 for the added
case, with a `_rebaseline_` justification key.

⚠️ base-red inherited: #9298 — `check:file-size` still reports 4 violations
(`providers/page.tsx`, `chat.ts`, `base.ts`, `accountFallback.ts`) and `lint`
reports 3 in `vertex-functioncall-id-3440.test.ts`. All five files are
byte-identical to the `release/v3.8.50` tip (`git hash-object` vs
`git rev-parse upstream/release/v3.8.50:<path>`) and none appear in this diff.


HouMinXi:
Closing this — #9600 already fixed the collision, four hours before I opened this, and I missed it because I did not re-fetch before pushing. Sorry for the noise.

Worth recording that the two fixes took different routes, since both are now visible in the history. #9600 renamed the files and added retroactive guards in `isSchemaAlreadyApplied` (`case "137"` / `case "138"` checking the column already exists). This PR renamed the same files to the same slots and instead added `RENAMED_MIGRATION_COMPATIBILITY` entries so `reconcileRenumberedMigrations()` carries the old markers forward. Both are upgrade-safe. The merged one leaves the old 135/136 rows in `_omniroute_migrations` and relies on the column probe; the marker route would have moved them. Not worth a follow-up PR — I mention it only so the choice is on the record rather than looking like an oversight if someone reads `constants.ts` and wonders why 135/136 are absent from a list that has ten similar entries.

One thing that may be worth acting on: slot 134 collided again 27 minutes after #9600 landed, when #9198 added `134_ccr_blocks.sql`. #9618 is already fixing it. That is three collisions in the same window, and each of the colliding PRs was individually green when it was opened — the duplicate only exists once both have landed, so a per-PR gate cannot see it coming. If that keeps costing merge-train time, the check that would catch it is one that runs against the merge result rather than the branch.

PR #9631, opened 2026-08-06
feat(db): add a job registry for scheduled background work

HouMinXi:
This is the first of the four PRs we discussed. It adds the job registry and `/api/jobs`, and moves two existing jobs onto it. Nothing here depends on the other three, so it can merge on its own.

## Why

You corrected my count on this: it is not three modules that schedule themselves, it is eleven. Each one owns a private `setInterval`, decides for itself whether to start, and reports nothing. There is no way to ask which background jobs exist, when one last ran, whether it failed, or why. Restarting the server is the only way to restart a stuck job, and disabling one means an env var and a redeploy.

The registry is one table, one class, and five endpoints. Jobs register a handler and a schedule; the registry owns the timer, records every run, and exposes the history.

## What is in it

`jobs` and `job_runs` tables (migration 139), a `JobRegistry` supporting interval and cron schedules, and five routes under `/api/jobs` for listing, run history, enable, disable, and run-now.

Cron uses `cron-parser` and re-arms the next fire before running the current handler, so a slow handler cannot break the chain. It computes the next instant from the IANA timezone rather than comparing wall-clock fields, so it does not double-fire or skip across a DST boundary.

Run history is pruned on two dimensions at once, keeping the most recent hundred runs per job and dropping anything older than thirty days that is already outside that window. A run still marked `running` past its timeout is reconciled to `failure` on the next startup, so a crash mid-run does not leave a row that claims to still be going.

## Scope: two jobs, not three

The migration seeds three job rows, but this PR only wires handlers for `budget_reset` and `token_health_check`. The third row is `warmup`, and its handler lives in #9449. Pulling it in here would make this PR depend on that one, which is the opposite of what you asked for.

So `warmup` is seeded with `enabled = 0`. `startAll()` filters on `enabled` before it looks for a handler, which means the row sits quiet instead of logging a missing-handler warning on every boot. Whichever change brings the warmup handler flips it on.

## Startup ordering

You asked for this to be observable, and writing the test found a real bug in my own change, so it earned its place immediately.

`startAll()` runs each interval job's first tick synchronously. I had it running before `initializeCloudSync()` had been awaited, which put both handlers in front of the initialisation they run against.

What made this easy to miss is that the old code reached the right ordering two different ways, and I replaced both at once. The budget reset was started after the init call. The health check started before it, but its first sweep sat behind a ten second timer. Neither is an explicit ordering constraint, so collapsing them into one `startAll()` silently dropped both. Both entry points now start the registry after the init call, and `tests/unit/jobRegistry/boot-wiring.test.ts` asserts it.

That test also asserts both entry points register the same set of jobs. `src/server-init.ts` previously started only the budget reset, which is pre-existing and currently harmless since nothing imports that module, but starting one job and not the other is exactly how a background job goes missing without anything failing.

## Migrations

Net negative on migration count is not something I can claim yet, since this adds one. It becomes true as the remaining schedulers move over and their bespoke state columns stop being needed.

## Testing

64 tests across the registry, time utilities, persistence, the API routes, and the boot wiring. The two existing `tokenHealthCheck` suites still pass unchanged.

`tokenHealthCheck.ts` shrinks rather than grows. It measures 1053 lines on the base tip against a frozen ceiling of 1021, and this branch brings it to 1048 by putting the registry adapter in `src/lib/jobs/` next to the budget reset one instead of adding to a file that is already over. No baseline was touched.

## ⚠️ base-red inherited: #9298

CI on this branch will fail, and the failures are not from this change.

Migration version 134 is claimed by both `134_ccr_blocks.sql` and `134_proxy_logs_egress_ip.sql` on `release/v3.8.50`. The migration runner throws on the collision, so every test that touches the database fails before reaching any assertion. #9618 is already fixing it. Neither file is in this diff; the only migration here is `139_job_registry.sql`. With the collision unblocked locally the suite is 64/64.

The file-size gate is also red on the base for `models/route.ts`, `catalog.ts` and `accountFallback.ts`, none of which this branch touches.


diegosouzapw:
## Babysit audit trail

### Fixes applied

1. **Add `cron-parser` to `dependency-allowlist.json`** — the PR adds `cron-parser` as a new dependency for cron-schedule support in the job registry. The `check:deps` gate requires human-reviewed allowlisting of every new dependency.

2. **Document `OMNIROUTE_RUNNOW_TIMEOUT_MS` in `.env.example` and `docs/reference/ENVIRONMENT.md`** — the PR's run-now API endpoint reads this env var, and the `check:env-doc-sync` gate requires it to appear in both files.

### Remaining failures (all base-red, not caused by this PR)

All 11 remaining failing checks are inherited from `release/v3.8.50`:

| Failure | Root cause | Tracking | Fixable here? |
|---|---|---|---|
| Unit Tests (4/4) | Migration version collision (134) | #9618 (OPEN) | No |
| Vitest (fast-path) | Migration version collision (134) | #9618 (OPEN) | No |
| Fast Quality Gates — check-migration-numbering | Migration version collision (134) | #9618 (OPEN) | No |
| No new ESLint warnings — stale suppressions | Pre-existing suppressions file drift | Base-red | No |
| Merge integrity — changelog fragment | `9415-newapi-sub2api-aggregator-balance.md` not starting with bullet | Base-red | No |
| Docs Gates — fabricated-docs | Stale version references in `PROXY_GUIDE.md`, `RELEASE_CHECKLIST.md` | Base-red | No |
| Build (advisory) | Turbopack build timeout | Pre-existing | No |
| dast-smoke | CLI bundle `Module not found` in `quotaStrategies.ts` | Pre-existing | No |

The PR author correctly identified these in the PR body ("⚠️ base-red inherited: #9298"). The migration collision fix #9618 is still open and must merge before these tests can pass.

PR is ready to merge once the base-red issues are resolved.

diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

Issue #9415, opened 2026-08-04
[Feature] Integration with NewAPI, Sub2API...


CoZooo:
### Problem / Use Case

I would like to combine all the registered sites like New API, One API, Sub2API, .etc into OmniRoute. But OmniRoute is more likely to manage certain providers' API keys. The only way maybe is to add OpenAI-Compatible providers.

[Metapi](https://github.com/cita-777/metapi) has the ability to detect the balance & api keys from these sites, but it is no longer maintained and it is not as good as OmniRoute in terms of experience. I wish OmniRoute can add this function so that accounts can be managed more easily.

### Proposed Solution

Add Custom Aggregators Gateways Page to manage sites.

### Acceptance Criteria

Detect the balance from New API/Sub2API sites.

### Area

Provider Support, Dashboard / UI

CoZooo:
Since each API key in these sites has a custom price multiplier, #6206 needs to be completed first.

diegosouzapw:
Thanks for the clear write-up — this is a great fit, and more of the machinery
already exists than you might expect, so I've moved it into the viable backlog.

OmniRoute already supports pointing at any OpenAI-compatible endpoint (New-API /
One-API / Sub2API instances included) as a custom "compatible provider node" with
a custom base URL, and it already implements the exact New-API balance pattern —
but only for one hardcoded instance (AgentRouter).
`open-sse/services/agentrouterQuotaFetcher.ts` calls New-API's
`GET /api/user/self` with the System Access Token + `New-Api-User` header, parses
the quota, and feeds it into the quota preflight + dashboard balance badge. The
plan is to generalize that: add an optional "New-API / One-API aggregator" flavor
to the compatible-node form that captures the aggregator base URL + console
credentials, and register a generalized New-API quota fetcher for those nodes so
the balance shows up the same way it does for AgentRouter. No DB migration, no
breaking changes — opt-in on the compatible-node form.

A couple of questions to nail down before implementation: (1) does Sub2API
specifically support the `Authorization: Bearer {systemAccessToken}` +
`New-Api-User` auth that New-API's `/api/user/self` uses, or does it need a
session-cookie path? (2) should `quota_per_unit` (units per $1) be configurable
per aggregator, or is the New-API default of 500000 fine? Leaving the issue open
while we scope it.

CoZooo:
Sub2API needs session cookies.

PR #9632, opened 2026-08-06
[defer] fix(api): keep the connection status when a test never reaches the upstream

### LINKED ISSUE #9623: fix(backend): Failed connection test leaves testStatus=error with no recovery path
A connection that fails a test is written to the database as `testStatus: "error"` and stays that way. Nothing moves it back to `active` except another successful test, or a successful real request through that exact connection. A brief network outage while connections are being tested therefore paints them red permanently, long after the network recovers.

I hit this on v3.8.50 after a host reboot. 20 connections across 6 providers went to `error_code = network_error` inside 0.823s (`14:15:23.625Z` to `14:15:24.448Z`). Sixty-three minutes later every one of them was still red.

## Why they cannot recover

`src/app/api/providers/[id]/test/route.ts:702` persists the failure:

```ts
testStatus: result.valid ? "active" : "error",
```

It sets no `rateLimitedUntil` and does not touch `backoffLevel`. I confirmed that against the live database: all 20 rows have `rate_limited_until` NULL and `backoff_level` 0.

The proactive recovery scheduler is the only thing that restores a connection without traffic flowing through it, and its predicate excludes these rows twice over (`src/lib/quota/connectionRecovery.ts:77-88`):

```ts
if (status !== RECOVERABLE_COOLDOWN_STATUS) return false;   // must be "unavailable"
return hasElapsedCooldown(connection.rateLimitedUntil, nowMs);
//     hasElapsedCooldown returns false when rateLimitedUntil is null
```

`RECOVERABLE_COOLDOWN_STATUS` is `"unavailable"`, not `"error"`, and these rows have no cooldown to elapse. The scheduler itself is healthy — my logs show it ticking every 60s and restoring genuine cooldown cases.

That leaves `clearAccountError` from a successful request (`src/sse/handlers/chatHelpers.ts:476`) or a manual re-test as the only exits.

## Routing is unaffected, which is what makes it easy to miss

Selection skips a connection on an unelapsed `rateLimitedUntil` (`src/sse/services/auth.ts:1196`) or on a terminal status. Every terminal set in the tree is `{banned, expired, credits_exhausted}` — `providers/rateLimit.ts:155`, `freeProviderRankings.ts:239`, `providerHealthAutopilot.ts:105`, `connectionRecovery.ts:40` — and `"error"` is in none of them.

So these connections keep serving traffic the whole time they read as broken. Nothing is down. What breaks is the dashboard's usefulness: after any transient blip most of the fleet reads red, so a genuine failure stops standing out.

## The tree already contains both answers

`src/lib/tokenHealthCheck.ts:143` faces the same situation and resolves it the other way:

```ts
testStatus: wasExpired ? "expired" : "active",
lastError: "Health check: token refresh failed",
errorCode: "refresh_failed",
```

It records what went wrong without moving the connection into a state it cannot leave. The test route moves it; the recovery scheduler only recognises `"unavailable"`. Three components, three different readings of what a transient failure means.

## Reproduction

1. Note a connection reading `active`.
2. Break outbound network (drop the default route, or point at an unreachable proxy).
3. Trigger a test for that connection.
4. Restore network and wait past several 60s recovery ticks.

It stays `error`. `rate_limited_until` stays NULL throughout, so it keeps serving traffic.

## Possible directions

Widening `isRecoverableCooldownConnection` to accept `"error"` is the smaller change, but it would also clear errors that were never transient, which is presumably why the predicate is narrow.

Having the test route classify transient network failures the way `tokenHealthCheck` already does looks closer to existing intent, since the diagnosis at `route.ts:149` has already decided the failure is `network_error` rather than a credential problem.

Happy to send a PR for whichever direction you prefer.

HouMinXi:
Fixes #9623.

This takes the second of the two directions in the issue, the one that classifies the failure at the test route rather than widening the recovery predicate. Before the code, a correction to what I wrote in that issue.

## The first direction would not have worked as I described it

I called widening `isRecoverableCooldownConnection` to accept `"error"` "the smaller change". That was wrong, and the issue quotes the reason two paragraphs above the place where I got it wrong.

The predicate has two independent gates:

```ts
if (status !== RECOVERABLE_COOLDOWN_STATUS) return false;   // must be "unavailable"
return hasElapsedCooldown(connection.rateLimitedUntil, nowMs);
```

`hasElapsedCooldown` returns false when `rateLimitedUntil` is null, and the rows written by a failed test have it null. Accepting `"error"` clears the first gate and the second one still rejects every row. To make that direction work the test route would also have to invent a `rateLimitedUntil` for a connection it never reached, which is a larger change than this one and puts a cooldown on a connection that is not rate limited.

## What this does

A `network_error` diagnosis means the request never reached the upstream. The test observed nothing about the connection, so it should not report a verdict on it.

The `testStatus` key is now left out of the update entirely in that case. `updateProviderConnection` merges with `{...existing, ...data}`, so omitting the key preserves whatever was stored. The error fields are still written, so the attempt is recorded and visible; only the claim about whether the connection works is withheld.

Leaving the key out is deliberate rather than writing `connection.testStatus` back. Writing it back would persist a value read earlier in the same request, and for a connection that has never been tested there is no prior value to write.

A real rejection from the upstream is unchanged. A 401 still marks the connection `error`, because there the test did reach the provider and did learn something.

## A second gap, found while writing the test

`classifyFailure` matched `"timeout"` as a substring, but `testOAuthConnection` reports its own abort as `Test timed out after 30s`. That string does not contain `"timeout"`, so an OAuth probe that hit the 30s ceiling was classified as `upstream_error` rather than `network_error`, and got marked broken on the strength of a timer expiring locally.

This was not in the issue. I found it by classifying the literal message strings rather than reasoning about the matching, and it is a separate defect from the one the issue describes, so it would have survived the fix above on its own.

## Tests

Three cases in `tests/unit/connection-test-network-error-status-9623.test.ts`, driving the real `testSingleConnection` against a temporary database with a stubbed `fetch`: a network failure preserves the stored status, a probe timeout does the same, and a genuine 401 still marks the connection as error.

Each of the two changes was verified by reverting it on its own. Reverting the classifier change fails only the timeout case. Reverting the status-write change fails both network cases while the 401 case stays green. With both in place, 3/3.

## ⚠️ base-red inherited: #9298

CI will fail on this branch, and not because of this change.

Migration version 134 is claimed by both `134_ccr_blocks.sql` and `134_proxy_logs_egress_ip.sql` on `release/v3.8.50`. The migration runner throws on the collision, so every test that opens the database fails before reaching an assertion, including all three here. #9618 is already fixing it. Neither file is in this diff. With the collision unblocked locally these tests are 3/3.


diegosouzapw:
## Babysit — audit trail

**Fix**: Added changelog fragment for PR #9632 (`changelog.d/fixes/9632-connection-test-network-error-status.md`).

**Root cause of failing check**: The `Merge integrity (changelog + generated skills)` check failed because (1) no changelog fragment was present for this PR (required by the fragments convention), and (2) the base branch had an invalid fragment (`changelog.d/features/9415-newapi-sub2api-aggregator-balance.md`) with YAML frontmatter instead of a markdown bullet — also fixed.

**Changes pushed** (1 commit, co-authored):
- `31f498b` — `fix(changelog): add #9632 fragment and fix #9415 fragment well-formedness`

**CI result**: `Merge integrity (changelog + generated skills)` is now **PASSING**. All remaining failures are pre-existing base-red issues (Vitest timeouts, file-size ratchets, ESLint, docs drift, DAST) and unrelated to this PR.

No tests weakened. No CI modified. No auto-merge.

diegosouzapw:
Fechando — superada. O fix de produção (route.ts) já está no tip via outra PR (#9688 ou batch anterior). O arquivo route.ts no tip já contém a lógica de network_error. Apenas o arquivo de teste não foi portado, mas o fix central já está entregue.

Issue #7039, opened 2026-07-13
fix(backend): combo context filter double-counts output reserve against maxInputTokens


forge-it:
## Summary

The combo request-compatibility filter conflates a model's **input-only limit** with its **total context window**.

For models that expose distinct limits, such as Codex `gpt-5.5-xhigh`:

- `maxInputTokens = 272000`
- `contextWindow = 400000`
- `maxOutputTokens = 128000`

OmniRoute currently computes:

```ts
requiredContextTokens = estimatedInputTokens + requestedOutputTokens;
```

but compares that combined value against:

```ts
maxInputTokens ?? contextWindow
```

This double-counts the requested output reserve against an already input-only cap. With Claude Code's usual `max_tokens: 32000`, the effective Codex input allowance becomes about 240K instead of 272K.

This is separate from the output-cap check, which is already performed correctly against `maxOutputTokens`.

## Affected version

- Reproduced on the exact `v3.8.46` tag (`92715c8`)
- The same predicate is still present on the current default branch at the time of filing
- Provider path: Claude Messages client -> priority combo -> Codex/OpenAI Responses target

## Root cause

In `open-sse/services/combo/comboStructure.ts`:

```ts
function getKnownContextLimit(capabilities: {
  maxInputTokens?: number | null;
  contextWindow?: number | null;
}): number | null {
  return capabilities.maxInputTokens ?? capabilities.contextWindow ?? null;
}

// ...

const contextLimit = getKnownContextLimit(capabilities);
if (
  requirements.requiredContextTokens > 0 &&
  contextLimit !== null &&
  contextLimit < requirements.requiredContextTokens
) {
  failures.push("context_window");
}
```

`requiredContextTokens` is `input + output`, while `getKnownContextLimit()` preferentially returns the input-only limit.

This appears to be an interaction between the context-aware combo filter and the distinct `maxInputTokens` metadata added for #6191. The #6191 regression tests correctly establish that `272000` is an input allowance distinct from the `400000` total context window, but the combo filter does not preserve that distinction.

## Minimal regression test

I added the following focused case to `tests/unit/combo-context-window-filter.test.ts` in a throwaway checkout of the exact `v3.8.46` tag:

```ts
test("a distinct max-input cap does not double-count the requested output reserve", () => {
  saveModelsDevCapabilities({
    "unit-split-limits": {
      split: capabilityEntryWithSplitLimits(400_000, 272_000, 128_000),
[...]

PR #9676, opened 2026-08-07
fix(combo,db): clear release/v3.8.50 base-red (typecheck + migration collision)

matiasbaglieri:
⚠️ base-red inherited: #9298

Fixes the 8 typecheck errors + 1 migration-version collision blocking `release/v3.8.50` (per `🔴 Release branch not green: release/v3.8.50` #9298).

## Changes

- **combo/quotaStrategies** — fix dangling import from #8894: `preferAntigravityConnectionsWithStoredProject` was imported from a non-existent file (`antigravityProjectPersistence.ts`) and the symbol existed nowhere. Pointed the import at `antigravityProjectPersist.ts` and implemented the missing helper there (stable, non-mutating sort that surfaces connections with a stored `projectId` first).
- **combo/comboStructure** — remove duplicate `HARD_COMPAT_REASONS` declaration (TS2451); narrow `step.prompt`/`step.model` access behind `step.kind === "model"` so wildcard steps don't violate the union type.
- **combo/fusionPanel** — push `step.model` only for model steps (wildcards have no `model`).
- **accountSemaphore** — pass `maxConcurrency ?? 1` to `ensureGate` (the `null` path early-returns via `isBypassed`, so `?? 1` is dead-code that satisfies the `number | null` param type).
- **compression/ccr** — pass `entry.principalId` as the first arg to `enforceGlobalBudget(owner, bytes)` (was called with only `bytes`, off by one arg — owner-preferring eviction).
- **firecrawlQuotaFetcher** — widen the return type of `fetchFirecrawlQuota` to `Promise<FirecrawlQuota | null>` so the `customBase` short-circuit branch (which returns a `FirecrawlQuota`-shaped literal) type-checks against the `QuotaInfo`-extending interface.
- **db/migrations** — renumber `134_proxy_logs_egress_ip.sql` → `139_proxy_logs_egress_ip.sql`. Version `134` was already taken by `134_ccr_blocks.sql` (#9061); #9291 reused it and triggered a migration-version collision. Added an `isSchemaAlreadyApplied(case "139")` retroactive guard (precedent: `085_quota_pools` renumbered from `077`) so a DB that already has `egress_ip` does not re-run the ALTER.

## Validation

- `npx tsc --noEmit -p tsconfig.typecheck-core.json` → **No errors found** (was 8 errors).
- Focused tests pass:
  - `tests/unit/antigravity-project-persist-prefer.test.ts` — 4/4 (new regression guard for the #8894 helper)
  - `tests/unit/db-schema-columns-split.test.ts` — 6/6 (updated to migration 139)
  - `tests/unit/combo/strict-context-failopen-8786.test.ts` — 6/6 (the migration-collision failure is gone)
  - `tests/unit/effective-max-concurrency.test.ts` — 7/7
  - `tests/unit/quota-share-concurrency.test.ts` — 8/8
- Lint clean on all changed files; Prettier clean.
- Dev server boots on `:20128` (`/` → 307, `/health` → 200).

Per Hard Rule #18, each fix carries or updates a test (the antigravity helper gets a new 4-case unit test; the migration renumber updates the existing schema-column test). The `accountSemaphore` and `ccr`/`firecrawl` fixes are pure type-argument corrections with no behavioral change to exercise beyond the now-passing typecheck.

### DIFF
--- a/open-sse/services/accountSemaphore.ts
+++ b/open-sse/services/accountSemaphore.ts
@@ -200,7 +200,7 @@ export function acquire(
     return Promise.reject(makeAbortError(signal));
   }
 
-  const gate = ensureGate(semaphoreKey, maxConcurrency);
+  const gate = ensureGate(semaphoreKey, maxConcurrency ?? 1);
   clearCleanupTimer(gate);
 
   if (gate.running < gate.maxConcurrency && !isBlocked(gate)) {
--- a/open-sse/services/antigravityProjectPersist.ts
+++ b/open-sse/services/antigravityProjectPersist.ts
@@ -15,6 +15,25 @@
 import { updateProviderConnection } from "@/lib/db/providers";
 
 /**
+ * Reorder antigravity connections so those with a persisted projectId surface
+ * first in quota-strategy selection (#8894). Connections without a stored
+ * projectId are kept (fallthrough) but ranked after. Stable; non-mutating.
+ * ponytail: O(n) copy + native stable sort; fine for connection-list sizes.
+ */
+export function preferAntigravityConnectionsWithStoredProject<
+  T extends { projectId?: unknown; providerSpecificData?: Record<string, unknown> | null },
+>(connections: T[]): T[] {
+  return connections
+    .map((conn, index) => {
+      const data = conn.providerSpecificData;
+      const hasProject = Boolean(conn.projectId || (data && data.projectId));
+      return { conn, hasProject, index };
+    })
+    .sort((a, b) => (a.hasProject === b.hasProject ? a.index - b.index : a.hasProject ? -1 : 1))
+    .map((entry) => entry.conn);
+}
+
+/**
  * Write `discoveredProjectId` onto both the `projectId` column and
  * `providerSpecificData.projectId` for `connectionId`, preserving any other
  * `providerSpecificData` fields already on the connection.
--- a/open-sse/services/combo/comboStructure.ts
+++ b/open-sse/services/combo/comboStructure.ts
@@ -137,7 +137,7 @@ function normalizeRuntimeStep(
       : {}),
     weight,
     label,
-    prompt: step.prompt || null,
+    prompt: step.kind === "model" ? step.prompt || null : null,
   } satisfies ResolvedComboTarget;
 }
 
@@ -612,13 +612,11 @@ function getTargetCompatibilityFailures(
 export type CompatFilterOptions = {
   /**
    * Opt-in legacy behavior (#8488): when every target fails a hard capability
-   * requirement (tools / vision / structured_output), restore the full pool
+   * requirement (tools / vision / structured_output / output_tokens), restore the full pool
    * instead of returning an empty list. Default is fail-closed.
    */
   failOpen?: boolean;
 };
-
-const HARD_COMPAT_REASONS = new Set(["tools", "vision", "structured_output"]);
 
 function hasHardCapabilityFailure(reasons: string[]): boolean {
   return reasons.some((reason) => HARD_COMPAT_REASONS.has(reason));
--- a/open-sse/services/combo/fusionPanel.ts
+++ b/open-sse/services/combo/fusionPanel.ts
@@ -51,7 +51,7 @@ export function extractFusionPanelSpec(
       panel.push(step.comboName);
       return;
     }
-    panel.push(step.model);
+    if (step.kind === "model") panel.push(step.model);
   });
   return { panel, comboRefUnits };
 }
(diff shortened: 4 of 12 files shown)


HouMinXi:
This PR and #9634 both fix the 134 collision, and both pick 139. They renumber the opposite file, so whichever merges second recreates the same failure one slot up.

- #9676: `134_proxy_logs_egress_ip.sql` → `139_proxy_logs_egress_ip.sql`, leaves `134_ccr_blocks.sql`
- #9634: `134_ccr_blocks.sql` → `139_ccr_blocks.sql`, leaves `134_proxy_logs_egress_ip.sql`

Either one alone clears the collision. Together they put two files on 139, and `getMigrationFiles()` throws exactly as it does today on 134.

There is a second overlap in the same switch: both add a `case "139"` to `isSchemaAlreadyApplied` — `hasColumn(db, "proxy_logs", "egress_ip")` here, `hasTable(db, "ccr_blocks")` in #9634. If both land, one of them is unreachable and that migration loses its retroactive guard.

One note on the comment in `139_proxy_logs_egress_ip.sql`:

> The collision threw before any DB could apply this file under 134, so renumbering is safe

That holds for `134_ccr_blocks.sql`, which landed in 8e27f5ec8 on 2026-08-06 11:07 -03 and made the runner throw immediately. It does not hold for `proxy_logs_egress_ip`, which landed in 8b97ef99a on 2026-08-04 14:34 -03 and sat alone on slot 134 for about 44 hours, so a DB that started up in that window did record version 134. The `case "139"` guard here covers that, but the reasoning above it points at the wrong file and would mislead whoever reads it next.

Happy to drop the migration hunk from #9634 and leave this one to carry the renumber, since this PR clears more of the base-red at once. Just say which way you want it and I will rebase.

HouMinXi:
Correcting my comment above: I named two PRs on slot 139, but there are three, and I missed the first one.

| PR | Author | Opened | Renumbers | `case "139"` guard |
|---|---|---|---|---|
| #9618 | fenix007 | 08-06 14:30 | `134_ccr_blocks` → 139 | `hasTable(db, "ccr_blocks")` |
| #9634 | me | 08-06 18:42 | `134_ccr_blocks` → 139 | `hasTable(db, "ccr_blocks")` |
| #9676 | this PR | — | `134_proxy_logs_egress_ip` → 139 | `hasColumn(db, "proxy_logs", "egress_ip")` |

#9618 came first and is byte-identical to the migration half of my #9634 — same five files, same guard. So #9634 adds nothing on the migration side and I will drop that hunk, leaving it to carry only the combo module-load fix.

That leaves #9618 and this PR, which still collide: both take 139, from opposite files, and both add a `case "139"` to the same switch. Merging both puts two files on 139 and makes one guard unreachable.

The point from my earlier comment stands and now favours #9618: `134_ccr_blocks` landed in 8e27f5ec8 on 08-06 11:07 -03 and made `getMigrationFiles()` throw immediately, so no database ever recorded version 134 for it. `134_proxy_logs_egress_ip` landed in 8b97ef99a on 08-04 14:34 -03 and held slot 134 alone for about 44 hours, so DBs that started up in that window did record it. Renumbering the file nobody applied is the safer half, which is what #9618 does.

Whichever way you go, only one of the two should carry the renumber.

HouMinXi:
Heads up on something this PR will surface through no fault of its own.

The unresolved `preferAntigravityConnectionsWithStoredProject` import stops `open-sse/services/combo.ts` from loading at all, which masks three failing assertions in `tests/unit/combo-context-window-filter.test.ts`. Repairing the import — which is what this PR does, and does correctly by implementing the helper rather than deleting the caller — makes them reachable, and they fail.

Measured on the same base, running only that file:

| Tree | Result |
|---|---|
| Upstream test + this PR's fix | `# pass 11  # fail 3` |
| Upstream test + the delete-the-caller fix | `# pass 11  # fail 3` — the same three |

So it is not an artefact of the approach; the assertions themselves are stale. They are:

- `input-only maxInputTokens still rejects when the input itself exceeds the cap`
- `maxInputTokens defaulting to contextWindow still rejects when input + output exceeds the total window (#7039 follow-up)`
- `without an override the small-catalog target is still dropped for the large request`

All three demand that catalog-too-small targets be dropped. The file's own header says the opposite:

> Context metadata is advisory: known-fitting targets are preferred, while unknown and catalog-too-small targets remain available for runtime fallback.

and four neighbouring tests that pass today assert exactly that, including `all known-too-small context targets still fall back to strategy order`. The file contradicts itself and the module-load failure is the only reason it never showed.

#9634 carries the fix for those three plus a new case pinning output-token limits as a genuine hard requirement, so the relaxation cannot drift further. With this PR's fix applied locally that file goes to `# pass 15  # fail 0`. Merging this one first and #9634 straight after leaves both green.

Nothing here needs changing in this PR — flagging it so a red run after merge does not read as a regression from your change.

PR #9712, opened 2026-08-07
fix(build): standalone bundle misses LLMLingua dist + onnxruntime native binaries

hartmark:
## Summary

Two related build-tooling bugs found while debugging a Docker `runner-base` build, each the next failure once the prior one was fixed:

1. **`colocateLlmlinguaOptionals` skip-check treated a Next-traced stub as fully copied.** Next.js's own standalone trace creates a stub directory for `@atjsh/llmlingua-2` containing only `package.json` (it references the package but can't fully bundle a dynamically-imported optional dependency). The skip check (both the closure-level early return and the per-package loop) only tested `existsSync(dest)`, so that stub was indistinguishable from "already fully co-located" — the function skipped copying the real `dist/` output entirely, silently shipping a package with a manifest but no code, failing at runtime with `Cannot find module '.../dist/index.js'`.

2. **`onnxruntime-node`'s native `bin/` was never registered as a standalone asset.** `dist/binding.js` is a normal JS file Next's standalone trace bundles correctly, but it `dlopen()`s a platform-specific native library shipped under `bin/napi-v3/<platform>/<arch>/libonnxruntime.so.1` — a dynamic native load static tracing can't see (same blind-spot class as bug 1, just for a `.so` instead of a JS import). Fails at runtime with `Error: libonnxruntime.so.1: cannot open shared object file`.

## Fixes

1. `isPackageFullyCopied` originally fixed this by checking the package's declared `main` entry file. Since this branch was authored, upstream independently shipped its own fix for the same bug (`isPackageIntact`, which resolves the package via `require.resolve` from inside the target tree — a stronger check than a bare file-existence test). Rebasing picked that up; this PR's own diff now only carries the regression test coverage for the same bug.
2. Added `NATIVE_ASSET_ENTRIES` entry for `onnxruntime-node/bin`, mirroring the existing `better-sqlite3` entry.

⚠️ base-red inherited: #9737 (re-check at merge time)

## Test plan

- TDD: `tests/unit/colocate-optionals.test.ts` (+1 new case, 7/7 total) and `tests/unit/assemble-standalone-onnxruntime-native-asset.test.ts` (2/2, new file) — both confirmed failing against the pre-fix code on a clean `release/v3.8.50` checkout, passing after
- `npm run lint` — clean (scripts/ files are lint-ignored by config, as expected)
- Both fixes originally confirmed against a real Docker `runner-base` build reaching further each time (Cannot find module -> libonnxruntime.so.1 missing -> build succeeds)

### DIFF
--- a/scripts/build/assembleStandalone.mjs
+++ b/scripts/build/assembleStandalone.mjs
@@ -48,10 +48,7 @@
 import fs from "node:fs/promises";
 import fsSync from "node:fs";
 import path from "node:path";
-import {
-  colocateLlmlinguaOptionals,
-  SEED_PACKAGES,
-} from "./colocateOptionals.mjs";
+import { colocateLlmlinguaOptionals, SEED_PACKAGES } from "./colocateOptionals.mjs";
 
 /**
  * Check whether a path exists (async).
@@ -78,7 +75,7 @@ async function exists(targetPath) {
  * (relative to projectRoot) and destination (relative to outDir) can be joined
  * for either path/platform. @type {{label:string, src:string[], dest:string[]}[]}
  */
-const NATIVE_ASSET_ENTRIES = [
+export const NATIVE_ASSET_ENTRIES = [
   {
     label: "wreq-js native runtime",
     src: ["node_modules", "wreq-js", "rust"],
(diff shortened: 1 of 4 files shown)


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

hartmark:
Code landed in #9892

Issue #9573, opened 2026-08-06
fix(backend): Reasoning replay cache self-poisons with thinking placeholder


stanleytejakusuma:
# Reasoning replay cache self-poisons with the thinking placeholder, causing empty turns on deepseek-v4-flash

**Package:** omniroute 3.8.49 (latest at time of report)

## Symptom
Sessions on `deepseek-v4-flash` (thinking mode) intermittently produce **empty assistant turns**: the response contains only `reasoning_content = "(prior reasoning summary unavailable)"` (~6 tokens), `finish_reason=stop`, and **no content and no tool calls**. Clients (e.g. Pi/Claude Code agents) record a thinking-only turn and the agent stops mid-task; the user must type "continue", which repeats the failure.

Frequency grows with context size and worsens over time within a session.

## Root cause — the replay placeholder poisons its own cache
`NON_ANTHROPIC_THINKING_PLACEHOLDER` (`open-sse/utils/reasoningPlaceholder.ts`) is injected as `reasoning_content` on cache miss for DeepSeek-family providers (required to avoid the upstream 400 "reasoning_content must be passed back"). Then:

1. **Echo:** DeepSeek's chain-of-thought continues from the injected scaffold and *echoes the placeholder text* as its response reasoning. Verified upstream: `usage.output=6, reasoning=6, finish=stop, content=""`.
2. **Write (unguarded):** `cacheReasoningByKey` / `cacheReasoningFromAssistantMessage` (`open-sse/services/reasoningCache.ts`) and `setReasoningCache` (`src/lib/db/reasoningCache.ts`) store the echoed placeholder verbatim. There is no `isInternalReasoningPlaceholder` guard on any write path.
3. **Replay (unguarded):** `lookupReasoning` returns the stored placeholder, which is re-injected into the next request. The conversation history also accumulates the placeholder (responses carry it as `reasoning_content`; `copyOpenAICompatibleReasoningFields` in `open-sse/utils/reasoningFields.ts` copies it verbatim to the client with no strip).
4. Once the history is seeded, the loop is self-sustaining: scaffold → echo → store → replay. At sufficient context the model stops after the echo (empty turn).

The placeholder file documents that response translators "must suppress it before emitting client-facing events", and `stripInternalReasoningPlaceholder` exists — but it is only applied to `delta.content`, **never to `reasoning_content`**, so the echo reaches the client and the cache.

## Proposed fix (minimal)
[...]

PR #9183, opened 2026-08-02
fix(responses-api): prevent output_index collision and data truncation

hartmark:
This PR fixes several critical bugs in the Responses API implementation that caused data loss, truncated responses, and incorrect turn grouping in multi-turn conversations.

### Key Fixes:
1. **Index Collisions**: Fixed a bug where reasoning, message, and tool calls were assigned colliding 'output_index' values, causing clients to ignore tool calls.
2. **Data Truncation**: Fixed a race condition in chunk processing where tool calls and finish signals were lost if they arrived in the same chunk as reasoning or text content.
3. **Multi-Choice Support**: Enabled full processing of multi-choice chunks, which were previously truncated to a single choice.
4. **Reasoning Replay**: Added support for capturing and replaying 'reasoning_content' for DeepSeek-based models (like 'big-pickle'). This prevents history corruption with generic placeholders that cause models to lose context and stop prematurely.
5. **Reasoning Extraction**: Added 'big-pickle' to models recognized for native textual reasoning tags, ensuring <think> blocks are correctly converted to reasoning items.
6. **Turn Grouping**: Improved the Responses-to-Chat translation to group reasoning, content, and tool calls into a single assistant turn, as required by strict OpenAI-compatible upstreams.
7. **Encrypted Reasoning**: Added a descriptive placeholder for encrypted Responses API reasoning blocks to maintain context in downgraded Chat API requests.

### Verification:
- Added 'tests/integration/live-responses-api.test.ts' to verify tool calls and multi-turn reasoning in a live environment.
- Added 'tests/unit/responses-api-truncation.test.ts' for truncation and multi-choice scenarios.
- Added 'tests/unit/responses-request-translation.test.ts' for reasoning conversion and turn grouping.
- Verified reasoning replay logic for 'big-pickle' via system-level logic tests.
- All Responses API unit and integration tests pass.

Issue #721, opened 2026-03-28
[BUG] Model Rerouting Despite Context Pinning (Multi-Request Bypass)


k0valik:
## Problem description

When using a combo with `context_cache_protection` enabled, a single conversation session from a CLI tool (e.g., Opencode) results in multiple separate calls to different providers/models, even when the model should be "pinned" via the `<omniModel>` context tag. The main concern is:

> **One single response from a "pinned" provider using tool calls is still rerouted to another model.**

This violates the expected behavior where every subsequent request in the same turn/conversation should use the same model when `context_cache_protection` is enabled for that combo.

## Expected Behavior

1. **Context Pinning**: When `context_cache_protection` is enabled on a combo, the first model selected should be "pinned" via injection of `<omniModel>provider/model</omniModel>` tag into the response
2. **Subsequent Requests**: All follow-up requests in the same conversation session (identified by the presence of the `<omniModel>` tag in message history) should route to the pinned model
3. **Tool Call Inheritance**: Tool calls within the same multi-turn conversation should inherit the pin from the initial turn
4. **Fallback Behavior**: If the pinned model becomes unavailable within the combo, fall back to the combo's strategy (least-used, round-robin, etc.) to select the next model, then re-pin
5. **User Override**: Pinning should be overridden if the user explicitly changes model/combo selection
6. **Task-Aware Routing**: Task-aware routing should check for existing pins before applying overrides; pinning should respect "background degradation" behavior, using default combo behavior secondarily if not defined

## Actual Behavior

1. **Pin Tag Not Injected Reliably**: The `<omniModel>` tag injection fails during tool-calling turns because:
   - Streaming: Regex `/"content":"([^"]+)/` requires **non-empty** content, but tool call chunks often have empty `content: ""` or `null`
   - Non-streaming: Code checks `json.messages` instead of standard `json.choices[0].message`
2. **Strategy Bypasses Pin**: Even when a pin exists, the combo strategy (e.g., `least-used`) is applied, selecting a different model
3. **Task-Aware Routing Ignores Pin**: Task-aware routing in `src/sse/handlers/chat.ts` applies overrides before checking for pins
4. **Model Switching Mid-Conversation**: Users experience degraded context consistency as models switch unexpectedly

### Multi-Request Pattern

```
[19:05:09.840Z] POST /v1/chat/completions | Sonnet | 3 msgs
[19:05:09.844Z] Combo "Sonnet" [least-used] with 5 models
[...]

PR #9741, opened 2026-08-08
fix(responses-api): sync reasoning-cache write index with the fixed read side

hartmark:
## Summary

Originally part of my own #9183 (closed in favor of an earlier, now-deferred branch). Re-verified against the current `release/v3.8.50` tip before extracting: most of #9183's scope (reasoning-content-alias generalization) has since been **superseded by upstream's own independent fix** — `translator/response/openai-responses.ts` now has its own dense-output-index-sort + `getReadableReasoningValue` implementation (its own comment says "mirrors upstream PR #721"). This PR extracts only the narrower reasoning-cache messageIndex sync below, which survives as a distinct, still-genuinely-broken bug.

**Correction (2026-08-08):** the "output_index collision prevention" part of #9183 was originally assumed fully covered by that same upstream fix. A live incident proved that assumption wrong for one specific case (a text message and a following tool call in the same turn colliding on the same output_index) — fixed separately in #9822, not part of this PR's scope.

The turn-index-hardcoding fix updated the reasoning-cache read side (`translator/index.ts`'s main replay loop) to key lookups by the assistant message's real position in the messages array, but two other spots still used the old hardcoded convention:

- `chatCore.ts`'s write side (both the streaming and non-streaming completion paths) still cached every response under a hardcoded `messageIndex: 0`.
- `translator/index.ts`'s own plain-turn (non-tool-call) cache-key lookup **also** still hardcoded `messageIndex 0` at its call site — found while re-verifying this fix live against the current tip; the original fix only addressed the write side.

Past the first assistant turn these conventions no longer matched, so DeepSeek/Xiaomi-mimo plain-turn reasoning replay silently missed the cache and fell back to the placeholder (or, once #9573 removed the placeholder fallback, to an absent field) in ordinary multi-turn conversations.

Compute the write-side index from the incoming request's message count instead, and use the real loop-provided `messageIndex` on the read-side lookup — both matching the position the response occupies once the client appends it to history for the next turn.

⚠️ base-red inherited: #9737 — `check:file-size` also reports two pre-existing, unrelated violations on this tip (`src/sse/handlers/chat.ts`, `open-sse/translator/response/openai-responses.ts`); this PR's diff is empty against both (confirmed via `git diff upstream/release/v3.8.50...HEAD -- <path>`)

## Test plan

- TDD: `tests/unit/reasoning-cache.test.ts`'s new end-to-end "write side (chatCore's messageIndex) and read side (translateRequest) agree on the same key end-to-end" test, plus the pre-existing "should inject placeholder for a plain (non-tool-call) DeepSeek turn" and "should replay cached reasoning for a plain (non-tool-call) DeepSeek turn when available" tests — confirmed failing against the pre-fix code on a clean `release/v3.8.50` checkout (both the hardcoded-0 write side AND the hardcoded-0 read-side lookup independently reproduce the mismatch), passing after both fixes
- `npm run lint` — clean
- `npm run check:file-size` — this PR's own diff is clean (`chatCore.ts` rebaselined 5034->5042; `reasoning-cache.test.ts` frozen at 1035); see base-red note above for the two unrelated inherited violations
- 2 pre-existing, unrelated test failures in the same file ("should replace empty-string reasoning_content with NON_ANTHROPIC_THINKING_PLACEHOLDER on cache miss", "should inject placeholder for a plain (non-tool-call) DeepSeek turn missing reasoning_content") confirmed present on a completely clean, untouched `release/v3.8.50` checkout — these test obsolete placeholder-injection behavior the code deliberately removed per #9573 (see the code's own comment); not touched by this PR


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

hartmark:
Code landed in #9895

PR #9744, opened 2026-08-08
test(integration): add general live-test tool for the real "default" combo + rootless wire capture

hartmark:
## Summary

Two pieces of live-test infrastructure, gated behind explicit env vars so they never run in CI (need podman, real provider credentials, and — for the wire-capture piece — tcpdump/python3):

**1. General default-combo live-test tool** (`RUN_LIVE_WIRE_CAPTURE=1` is for piece 2 below; this piece reuses `omniroute-beta`-style gating via `OMNIROUTE_API_KEY`) — `tests/integration/liveDefaultComboShared.ts` + `live-default-combo-workload.test.ts`. Unlike the existing `liveGeminiShared.ts` (which provisions its own narrow 2-model Gemini-only combo), this reads the REAL `default` combo currently configured on the target instance directly from the DB and exercises every provider/model step in it directly, bypassing combo routing — so live-test coverage always matches whatever the operator actually has configured instead of a hardcoded snapshot that goes stale the moment the combo changes. Live-verified earlier this session: 14/18 models pass consistently across non-streaming + streaming Chat Completions and streaming Responses API against the real 18-model/5-provider `default` combo; the 4 consistent failures were real external state (a provider temporarily out of credits, one deprecated free-tier model), not code regressions.

**2. Rootless wire-capture correlation** (`RUN_LIVE_WIRE_CAPTURE=1`) — `tests/integration/liveContainerHarness.ts`, `wireCapture.ts`, `live-default-combo-wire-capture.test.ts`. Spins up a dedicated, throwaway podman container running the target branch's own code, captures its real wire traffic via `podman unshare nsenter --net=<container netns> -- tcpdump` (no root needed — verified working live), sends a sample of real requests through the real `default` combo, and cross-checks each one's app-level JSON result against the actual HTTP status line + TCP behavior observed on the wire via `scripts/sre/tcp-close-analyzer.py`'s stream reassembly (that script's own docstring now documents the rootless capture command alongside the existing root-requiring one). Catches bugs where the app layer claims one thing but the wire shows another — e.g. a truncated/reset stream, or a status mismatch.

Building this surfaced two real, non-obvious bugs along the way (both fixed inline, not part of this PR's own diff since they're general build-tooling issues found via `liveContainerHarness.ts` needing to actually build the target image):
- `podman unshare -> nsenter -> tcpdump` is a 3-level subprocess chain; `SIGTERM` to the top-level tracked process doesn't reliably reach the tcpdump grandchild, leaving orphaned processes and a truncated/unreadable pcap. Fixed with `-U` (unbuffered tcpdump writes) + a `pkill -f <pcap path>` fallback in `wireCapture.ts`'s own cleanup.
- Wire capture happens *inside* the container's own network namespace, where only the container's *internal* listening port is meaningful — filtering on the dynamically-assigned *host* port (the natural first instinct) silently captures nothing.

⚠️ base-red inherited: #9737

## Test plan

- `npm run lint` — clean
- Live-verified end-to-end from a clean cherry-pick onto the current `release/v3.8.50` tip: wire-capture test passed with 4/4 sampled requests correctly correlated across 11 captured TCP streams — including a real `HTTP 400` (deprecated model) and a real `HTTP 502` (transient upstream error), both correctly matched between the app-level result and the wire-level status line, not just the happy-200 path. Container + tcpdump fully torn down afterward each run (verified no orphaned podman container or process left running).
- While testing against a freshly-built image from the current upstream tip, hit and worked around (locally only, not part of this diff) two separate pre-existing base-red issues to actually get a working build/DB: the LLMLingua/onnxruntime native-asset bug already fixed in #9712, and a live migration-numbering collision (two files both claiming version `135` — `135_connection_runtime_state.sql` and `135_migrate_model_capability_max_token.sql`) that will be filed and fixed as its own separate PR next, since it currently breaks any fresh build from `release/v3.8.50`.

### DIFF
--- a/scripts/sre/tcp-close-analyzer.py
+++ b/scripts/sre/tcp-close-analyzer.py
@@ -17,8 +17,8 @@ parsing the libpcap file format and IPv4/TCP headers directly. Good enough
 for this one question; not a general-purpose pcap toolkit.
 
 ────────────────────────────────────────────────────────────────────────────
-CAPTURING (run this yourself — needs root/sudo for CAP_NET_RAW; also see
---show-capture-cmd)
+CAPTURING (run this yourself — needs root/sudo for CAP_NET_RAW, UNLESS you
+use the rootless method below; also see --show-capture-cmd)
 ────────────────────────────────────────────────────────────────────────────
 
 Rootless Podman gotcha: there is usually NO `podman3`/`podmanN` bridge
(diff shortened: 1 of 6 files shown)


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

hartmark:
Code landed in #9862

PR #9746, opened 2026-08-08
fix(pricing): memoize getSyncedPricing() across catalog cache versions

chloeassistant:
## Problem

`getSyncedPricing()` (src/lib/pricingSync.ts) re-ran `SELECT + JSON.parse` of every `pricing_synced` blob on each call. `resolveCatalogPricing()` calls it once per model lookup, so each call returned a **fresh object**; `findInsensitive()` caches its lowercase index by object identity (WeakMap), so every fresh object rebuilt the index and emitted hundreds of `case-insensitive key collision` warnings per second while pinning a CPU core (~92-105% of one core sustained).

Impact observed on a 3.8.50 install with a 5376-model catalog:
- `GET /v1/models` took **15-16s** per call and added **~5899 collision warnings** per call.
- `/api/health/ping` stalled **>5s** (up to 14s) and the request queue dropped requests with 503s.

## Fix

Memoize the parsed result and invalidate via the existing `modelCatalogCacheVersion` signal — `saveSyncedPricing()`/`clearSyncedPricing()` already bump it through `invalidateDbCache("pricing")` (readCache.ts increments `modelCatalogCacheVersion` on every invalidation). This mirrors the pattern already proven in `getModelsDevPricing()` (#8697).

Measured after the fix (same install):
- `GET /v1/models` cold: **5.2s** (first, includes index warm-up); warm second call: **0.013s**; subsequent catalog rebuilds **~3s**.
- Collision warnings: **0** during the second call; 15 emitted once at the first index build (expected — the index is now built once per cache version).
- Concurrent `/api/health/ping` during the cold call: max **7.0s**, none >8s; after warm-up **2-4ms**.

## Tests

New `tests/unit/pricing-sync-memoization.test.ts` mirroring the models-dev memoization suite:
- same object reference across repeated reads within one cache version,
- at most 1 `db.prepare()` across 3 reads,
- a new reference with fresh data after a pricing write invalidates the cache.

RED before the source patch (identity assertion fails, 3 prepare calls observed), GREEN after. Sibling pricing suites (pricing-sync, pricing-sync-cross-instance, pricing-sync-extended, models-dev-pricing-memoization-8697, catalog-pricing-lookup-index-8697): 24/24 pass.

### DIFF
--- a/src/lib/pricingSync.ts
+++ b/src/lib/pricingSync.ts
@@ -11,7 +11,7 @@
  */
 
 import { getDbInstance } from "./db/core";
-import { invalidateDbCache } from "./db/readCache";
+import { invalidateDbCache, getModelCatalogCacheVersion } from "./db/readCache";
 import { backupDbFile } from "./db/backup";
 
 // ─── Types ───────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

PR #9758, opened 2026-08-08
feat(radar): paste-key input on the activation screen (F4)

diegosouzapw:
⚠️ base-red inherited: migration collision `135_connection_runtime_state` × `135_migrate_model_capability_max_token` breaks DB boot on the tip (3rd collision this cycle — not caused by this branch; needs a base-reds drain PR).

Closes the gap where the Radar activation screen had no field to paste the `omr_` supporter key — the last missing piece of the supporter/contributor flow.

## What this adds
- **Paste-key field** in the activation card (primary path), above "activate without a key" and the existing claim buttons. Submits `POST /api/radar/settings {optIn:true, supporterKey}` in one request.
- **Shared format validator** `src/lib/radar/supporterKey.ts` (`omr_` + 40 hex), client-safe, reused by the client and the route's Zod schema (dedupes the inline regex).
- **Already-activated state**: shows the masked form (`omr_****abcd`) with a "change key" button — the raw key is never rendered.

## Security
- Raw key never appears in any response (POST/GET always masked) nor in logs (no `console.*`); tested byte-for-byte with `!text.includes(RAW_KEY)`.
- No price/monetary value anywhere (D14, regex-tested across page + 43 locales).

## Validation
40 radar tests green across the touched suites (supporter-key-format 7, key-input 7, api-routes 21, inertia 5), run against a deduplicated migrations dir because of the inherited 135 collision. `typecheck:core`/ESLint clean for touched files; i18n coverage PASS (4 new keys × 43 locales, English fallbacks).


diegosouzapw:
CI triage before merge: all 7 reds are inherited/infra, zero in this branch's files. **Unit Tests fast-path (1-4)** = `ERR_DLOPEN_FAILED` (better-sqlite3 native module failed to load on the runner) cascading into `no such table` — infra, not code (tip boots locally; this branch adds no migration/native change). **Docs Gates** = 48 `stale-version` hits in untouched docs. **Fast Quality Gates / No new ESLint warnings** = file-size/CodeQL base-reds. Radar-scope tests green locally against the drained base. Merging per operator instruction.

PR #9777, opened 2026-08-08
fix(sse): route claude discovery aliases for catalog-only providers

Michael-Rocco-Goldmann:
## Bug

Claude Code discovery aliases are advertised for providers the request path then refuses to route, so **every** such request silently goes to the Claude provider instead.

Reproduced live against an Azure AI Foundry connection:

```
POST /v1/messages  {"model":"claude/azure-ai/DeepSeek-V4-Flash"}
-> ROUTING: Provider: claude, Model: azure-ai/DeepSeek-V4-Flash
-> 404 (routed to the Claude provider, not Azure)

POST /v1/messages  {"model":"azure-ai/DeepSeek-V4-Flash"}   # direct id
-> 200 OK
```

No `CC_DISCOVERY` log line was emitted, confirming `stripped === false`.

## Root cause

The two sides of the alias disagree:

| side | predicate |
| --- | --- |
| Catalog (`ccAliasPredicate.ts`) | alias gate **only** - consults no provider registry |
| Request (`stripCcDiscoveryAlias`) | gate **plus** `isKnownProviderPrefix` |

`isKnownProviderPrefix` was built from `getRegistryEntry()` (the `open-sse` REGISTRY) plus operator-defined custom nodes. Enterprise-cloud providers such as `azure-ai` / `azure-openai` live only in the **provider catalog** (`src/shared/constants/providers/apikey/enterprise-cloud.ts`) and have no `open-sse` registry entry - `"azure-ai"` appears nowhere under `open-sse/config/`.

So the catalog advertised the mirror id, the request path declined to strip it, and the unstripped id fell through to normal resolution - which splits on the first `/` and parses `claude` as the provider.

This is exactly the hazard called out in `src/lib/ccDiscoveryAliasResolve.ts`: "they must match or an alias could be listed yet rejected."

## Fix

Extract the predicate as `isRoutableProviderPrefix()` and widen it to the provider catalog (`getProviderById` + `getProviderByAlias`) alongside the existing registry lookup, so the request path recognises exactly what the catalog can advertise.

Widening the request side (rather than narrowing the catalog) is the correct direction: these models genuinely route - the direct ids return 200.

## Tests

New `tests/unit/cc-discovery-alias-routable-prefix.test.ts`:

- `azure-ai`, `azure-openai` routable (the regression)
- `azure` alias resolves
- `openai` / `anthropic` stay routable
- unknown prefix and empty string stay non-routable

**TDD verified** - the azure and alias assertions failed against the pre-fix predicate (`false !== true`), then passed after the widening.

```
new suite                    4 pass / 0 fail
existing cc-discovery suites 70 pass / 0 fail   (6 files)
typecheck:core               clean
eslint (changed files)       exit 0
```

WARNING - base-red inherited: #9737

### DIFF
--- a/src/lib/ccDiscoveryAliasResolve.ts
+++ b/src/lib/ccDiscoveryAliasResolve.ts
@@ -20,6 +20,7 @@ import {
 } from "@omniroute/open-sse/handlers/chatCore/ccDiscoveryAliasStrip.ts";
 import { getModelsByProviderId } from "@omniroute/open-sse/config/providerModels.ts";
 import { getRegistryEntry } from "@omniroute/open-sse/config/providerRegistry.ts";
+import { getProviderById, getProviderByAlias } from "@/shared/constants/providers";
 import { getCachedProviderNodes } from "@/lib/db/readCache";
 import { getComboByName } from "@/lib/db/combos";
 import {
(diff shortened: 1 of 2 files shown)


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

Issue #8295, opened 2026-07-23
fix(api): Responses→Chat namespace flattening emits duplicate bare-leaf tool names across namespaces


ashe0047:
### OmniRoute Version

v3.8.48

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Provider(s) Involved

DeepSeek

### Model(s) Involved

deepseek/deepseek-v4-flash

### Client Tool

Codex desktop app

### Description

When a Codex `/responses` request contains multiple `namespace` tools whose children share the same leaf name, OmniRoute’s Responses → Chat Completions translator flattens the children using only the bare leaf name.

For example:

```text
mcp__codex_apps__atlassian_rovo._search
mcp__codex_apps__linear._search
```

are both translated into provider-visible function tools named:

```text
_search
```

The upstream provider then rejects the translated request with:

```text
400: Tool names must be unique.
```

The same DeepSeek model and provider configuration work through other coding harnesses using OmniRoute. The failure is specific to Codex requests containing Responses API namespace tools.

The same Codex request also works through CC Switch local routing, whose translation implementation generates namespace-qualified provider-visible function names rather than bare leaf names.

### Steps to Reproduce

1. Configure the Codex desktop app to use OmniRoute.
2. Configure OmniRoute to route requests to `deepseek/deepseek-v4-flash`.
3. Send a Codex `/responses` request containing at least two namespace tools whose children share the same name.

Minimal conceptual request:

```json
{
  "model": "deepseek/deepseek-v4-flash",
  "tools": [
    {
      "type": "namespace",
      "name": "namespace_a",
      "tools": [
        {
          "type": "function",
          "name": "_search",
          "parameters": {
            "type": "object",
            "properties": {}
          }
        }
      ]
    },
    {
      "type": "namespace",
      "name": "namespace_b",
      "tools": [
        {
          "type": "function",
          "name": "_search",
          "parameters": {
            "type": "object",
            "properties": {}
          }
        }
      ]
    }
  ]
}
```

4. Submit any prompt.
5. Observe that the upstream request fails before generation begins.

A real Codex request that reproduced the issue included these namespace tools:

```text
mcp__codex_apps__atlassian_rovo
mcp__codex_apps__linear
```

Both exposed a child named `_search`.

The complete request body is intentionally omitted because it is approximately 150 KB and contains unrelated tool descriptions and schemas.

### Expected Behavior

[...]

PR #9783, opened 2026-08-08
fix(translator): keep Responses namespace identity across the hub-and-spoke pivot

### LINKED ISSUE #9780: fix(providers): Responses namespace identity map is dropped by the hub-and-spoke pivot
## Summary

The Responses namespace identity map introduced by #7936 (and re-keyed by #8295) is silently dropped by the hub-and-spoke pivot in `open-sse/translator/index.ts`. Every `openai-responses` → non-OpenAI target loses it, so the response-side seam gets `requestToolIdentityMap = null` and never restores `{namespace, name}`.

Net effect on Kiro: Codex CLI receives `"name":"functions__exec"` with no `namespace` field and rejects the call with `unsupported call: functions__exec` — the exact symptom #7936 was opened to fix, reintroduced through a different path.

Reproduced on a clean checkout of `main` at `918fba5e3`.

## Root cause

There is no direct `openai-responses` → `kiro` translator, so `translateRequest` falls back to the hub-and-spoke path (`index.ts`, "Fallback: hub-and-spoke via OpenAI"):

- **Step 1** `openai-responses` → `openai`: `openai-responses.ts` flattens each namespace sub-tool to a qualified wire name via `flattenNamespaceToolName()` (#8295) and stores the `{namespace, name}` pair in `namespaceToolIdentityMap`, attached with `Object.defineProperty(result, "_toolNameMap", { enumerable: false, ... })` (`openai-responses.ts:749`).
- **Step 2** `openai` → target: `buildKiroPayload()` (`openai-to-kiro.ts:660`) builds and returns a **brand-new** `payload` object. The non-enumerable property from step 1 is never carried over. `openai-to-kiro.ts` does not reference `_toolNameMap` anywhere (0 matches).
- `chatCore.ts:2258` then reads `translatedBody._toolNameMap` → `null` → `resolveRequestToolIdentity()` always returns falsy → the flattened name reaches the client verbatim.

The seam itself is correct and works for targets that happen to re-attach the property (`cliproxyapi.ts:368`, `executors/base.ts:1346`, `antigravity/executeAttempt.ts:240`). Only the pivot's step 2 has no propagation.

## Reproduction

Translator-level, no network and no provider credentials. Save as `.work/repro-8295-pivot.mjs`, run `npx tsx .work/repro-8295-pivot.mjs` from the repo root:

```js
import "../open-sse/translator/bootstrap.ts";
import { getRequestTranslator } from "../open-sse/translator/registry.ts";

const body = {
  model: "m",
  instructions: "coding agent",
  input: [{ type: "message", role: "user", content: [{ type: "input_text", text: "run git" }] }],
  tools: [{ type: "namespace", name: "functions", tools: [{
    name: "exec",
    description: "Run a shell command",
    parameters: { type: "object", properties: { cmd: { type: "string" } }, required: ["cmd"] },
  }]}],
};

const toOpenAI = getRequestTranslator("openai-responses", "openai");
const describe = (m) => {
  if (!(m instanceof Map)) return "ABSENT";
  const [k, v] = [...m.entries()][0];
  return `${v && typeof v === "object" ? "identity {namespace,name}" : "alias string"}  ${JSON.stringify([k, v])}`;
};

for (const target of ["kiro", "openai-responses", "cursor", "claude", "gemini"]) {
  const fromOpenAI = getRequestTranslator("openai", target);
  if (!fromOpenAI) continue;
  const step1 = toOpenAI("m", structuredClone(body), true, null);
  const before = describe(step1._toolNameMap);
  let after;
  try { after = describe(fromOpenAI("m", step1, true, null)?._toolNameMap); }
  catch (e) { after = `threw: ${String(e.message).slice(0, 50)}`; }
  const ok = after.startsWith("identity");
  console.log(`${target.padEnd(18)} ${ok ? "OK  " : "FAIL"}  after step 2: ${after}`);
  if (!ok) console.log(`${" ".repeat(18)}      step 1 had: ${before}`);
}
```

Actual output on `918fba5e3` — step 1 always builds the map correctly, step 2 always destroys it:

```
kiro               FAIL  after step 2: ABSENT
                        step 1 had: identity {namespace,name}  ["functions__exec",{"namespace":"functions","name":"exec"}]
openai-responses   FAIL  after step 2: ABSENT
                        step 1 had: identity {namespace,name}  ["functions__exec",{"namespace":"functions","name":"exec"}]
cursor             FAIL  after step 2: ABSENT
                        step 1 had: identity {namespace,name}  ["functions__exec",{"namespace":"functions","name":"exec"}]
claude             FAIL  after step 2: alias string  ["proxy_functions__exec","functions__exec"]
                        step 1 had: identity {namespace,name}  ["functions__exec",{"namespace":"functions","name":"exec"}]
gemini             FAIL  after step 2: alias string  ["functions_exec","functions__exec"]
                        step 1 had: identity {namespace,name}  ["functions__exec",{"namespace":"functions","name":"exec"}]
```

Expected: `identity {namespace,name}` preserved on every row.

## End-to-end confirmation

Live request through a running OmniRoute (`kr/gpt-5.6-luna`, UA `codex_cli_rs/0.147.0`, the namespace payload above), status 200:

```
"type":"function_call","name":"functions__exec","status":"in_progress"
```

No `namespace` field, and `name` is the qualified wire name rather than the declared `exec`. Codex's adjudicator has no dispatch entry for it and answers `unsupported call: functions__exec`.

## The trap in the obvious fix

Copying `_toolNameMap` through in step 2 would be wrong. That single property carries **two mutually incompatible types**:

| Producer | Type | Purpose |
| --- | --- | --- |
| `openai-responses.ts:749` | `Map<string, {namespace, name}>` | #7936 identity seam |
| `openai-to-claude.ts:479`, `openai-to-gemini.ts:638` | `Map<string, string>` | tool-name alias remap (#1375) |

The `claude` and `gemini` rows above show step 2 overwriting the identity map with an alias map. A naive copy-through would clobber those aliases and break tool-name remapping for those providers. `chatCore.ts:2256` already documents the distinction ("Keep the request translator's namespace identities separate from toolNameMap") and extracts the identity map before the Kiro merge — but nothing makes it survive step 2 in the first place.

A fix therefore needs a **separate channel** (e.g. a non-enumerable `_namespaceToolIdentityMap`) propagated by `index.ts` across the pivot, with `chatCore.ts` preferring it and falling back to the current `_toolNameMap` read for backward compatibility.

## Affected

- `kiro`, `cursor`, `openai-responses` — identity map lost entirely
- `claude`, `gemini` — identity map overwritten by the alias map

Any Responses-API client sending `{type:"namespace"}` tool groups (Codex CLI, MCP tool groups) routed to a non-OpenAI provider.

## Related

- Continues #7936 (identity seam) and #8295 (qualified wire names)
- Same user-visible symptom as #7936: `unsupported call: <flattened name>`

## Environment

- `main` @ `918fba5e3`, clean checkout, no local patches
- Node v22.23.2

I have a fix working locally and can open a PR.

VXNCXNX:
Fixes #9780.

## Problem

There is no direct `openai-responses` → `kiro` translator, so `translateRequest` falls back to the hub-and-spoke pivot:

- **Step 1** (`openai-responses` → `openai`) flattens each namespace sub-tool to a qualified wire name (#8295) and records the `{namespace, name}` pair, attached with `Object.defineProperty(result, "_toolNameMap", { enumerable: false })`.
- **Step 2** (`openai` → target) returns a **brand-new** object — `buildKiroPayload()` and friends never copy that property over.

`chatCore.ts` then reads `translatedBody._toolNameMap` → `null`, so `resolveRequestToolIdentity()` never fires and the flattened name reaches the client verbatim. Codex has no dispatch entry for it and answers `unsupported call: functions__exec` — the exact symptom #7936 was opened to fix, reintroduced through a different path.

Confirmed against a live gateway (`kr/gpt-5.6-luna`, UA `codex_cli_rs/0.147.0`), HTTP 200:

```
"type":"function_call","name":"functions__exec","status":"in_progress"
```

No `namespace` field, and `name` is the wire name rather than the declared `exec`.

## Why not just copy `_toolNameMap` through

That property carries **two mutually incompatible types**:

| Producer | Type | Purpose |
| --- | --- | --- |
| `openai-responses.ts` | `Map<string, {namespace, name}>` | #7936 identity seam |
| `openai-to-claude.ts`, `openai-to-gemini.ts` | `Map<string, string>` | tool-name alias remap (#1375) |

On a pivot, step 2 overwrites the identity map with an alias map for those two targets. A naive copy-through would clobber their aliases and break tool-name remapping. `chatCore.ts` already documents the distinction ("Keep the request translator's namespace identities separate from toolNameMap") — but nothing made the identity map survive step 2 in the first place.

## Fix

A dedicated `_namespaceToolIdentityMap` channel:

- `openai-responses.ts` publishes the identity map on the new key **and** keeps writing `_toolNameMap`, so the existing consumers (`executors/base.ts`, `cliproxyapi.ts`, `antigravity`) are unaffected.
- `translator/index.ts` re-attaches it after step 2 of the pivot, non-enumerably.
- `chatCore.ts` prefers the new key and falls back to `_toolNameMap` for the non-pivot producers.
- `cliproxyapi.ts` strips both keys from the wire body.

49 insertions, 6 deletions across 4 files. Nothing changes for targets that were already working.

## Verification

`tests/unit/9780-namespace-identity-pivot.test.ts`, 10 cases. On `main` without the fix: **1 pass, 9 fail**. With the fix: **10 pass**.

The end-to-end cases chain the request pivot and the response seam, i.e. what the Codex adjudicator actually receives:

| Target | Before | After |
| --- | --- | --- |
| kiro | `functions__exec`, no namespace | `exec` + `functions` |
| cursor | `functions__exec`, no namespace | `exec` + `functions` |
| claude | `functions__exec`, no namespace | `exec` + `functions` |
| gemini | `functions__exec`, no namespace | `exec` + `functions` |

The "before" column reproduces the live gateway output above.

Note `claude` / `gemini`: they carry a map even without the fix, yet still fail — it is the alias map. That is the regression guard, asserted explicitly (identity restored **and** alias values still strings).

Regression run — all `*responses*` and `*namespace*` suites, plus `chatcore-passthrough-tool-names`, `executor-kiro`, `translator-openai-to-kiro`, `8295-namespace-leaf-collision`, `translator-resp-openai-responses-namespace-identity`: **442 pass, 0 fail**.

### DIFF
--- a/changelog.d/fixes/9783-namespace-identity-pivot.md
+++ b/changelog.d/fixes/9783-namespace-identity-pivot.md
@@ -1,0 +1,1 @@
+- **Translator**: keep the Responses namespace identity map across the hub-and-spoke pivot — namespace sub-tool calls routed to non-OpenAI targets (Kiro, Cursor) no longer come back flattened (`unsupported call: functions__exec` in Codex CLI) (#9783 — thanks @VXNCXNX)
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -1,3 +1,4 @@
+import { extractRequestToolIdentityMap } from "./chatCore/requestToolIdentity.ts";
 import { injectMemoryAndSkills } from "./chatCore/memorySkillsInjection.ts";
 import { resolveChatCoreRequestSetup } from "./chatCore/requestSetup.ts";
 import { buildFailureUsageRecord } from "./chatCore/failureUsage.ts";
@@ -2264,20 +2265,7 @@ export async function handleChatCore({
   // the latter is a Kiro/Claude passthrough alias channel with string values,
   // while namespace identities carry `{namespace, name}` for the #7936 response
   // seam. Extract first because Kiro merge may reuse `_toolNameMap` below.
-  //
-  // #9780 — prefer the dedicated channel: on a pivot the openai->claude/gemini
-  // step publishes its own alias map on `_toolNameMap`, so that property alone
-  // yields aliases here. The `_toolNameMap` read stays as the fallback for the
-  // non-pivot producers (executors/base.ts, cliproxyapi.ts, antigravity).
-  const namespaceIdentityMap = translatedBody._namespaceToolIdentityMap;
-  const requestToolIdentityMap =
-    namespaceIdentityMap instanceof Map
-      ? namespaceIdentityMap
-      : translatedBody._toolNameMap instanceof Map
-        ? translatedBody._toolNameMap
-        : null;
-  delete translatedBody._namespaceToolIdentityMap;
-  delete translatedBody._toolNameMap;
+  const requestToolIdentityMap = extractRequestToolIdentityMap(translatedBody);
 
   // Kiro: sanitize tool schemas before dispatch. Kiro returns 400 "Improperly
   // formed request" for unsupported JSON-Schema keywords (anyOf/$ref/if-then,
--- a/open-sse/handlers/chatCore/requestToolIdentity.ts
+++ b/open-sse/handlers/chatCore/requestToolIdentity.ts
@@ -1,0 +1,26 @@
+type NamespaceIdentity = { namespace: string; name: string };
+
+/**
+ * Extract the #7936 request-tool identity map from the translated body and
+ * strip both side channels before dispatch.
+ *
+ * #9780 — prefer the dedicated `_namespaceToolIdentityMap`: on a pivot the
+ * openai->claude/gemini step publishes its own alias `Map<string, string>` on
+ * `_toolNameMap`, so that property alone can yield aliases instead of
+ * identities. The `_toolNameMap` read stays as the fallback for the non-pivot
+ * producers (executors/base.ts, cliproxyapi.ts, antigravity).
+ */
+export function extractRequestToolIdentityMap(
+  translatedBody: Record<string, unknown>
+): Map<string, NamespaceIdentity> | null {
+  const namespaceIdentityMap = translatedBody._namespaceToolIdentityMap;
+  const requestToolIdentityMap =
+    namespaceIdentityMap instanceof Map
+      ? namespaceIdentityMap
+      : translatedBody._toolNameMap instanceof Map
+        ? translatedBody._toolNameMap
+        : null;
+  delete translatedBody._namespaceToolIdentityMap;
+  delete translatedBody._toolNameMap;
+  return requestToolIdentityMap as Map<string, NamespaceIdentity> | null;
+}


diegosouzapw:

diegosouzapw:

diegosouzapw:

diegosouzapw:

VXNCXNX:
Rebased onto `release/v3.8.50` as requested (branch rebuilt from `918647af6` + the fix cherry-picked, now `d7f076bf3`).

Also fixed the Lint failure: `chatCore.ts` had grown past its frozen 5020-line cap. The #9780 extraction block now lives in a new `open-sse/handlers/chatCore/requestToolIdentity.ts` helper (same pattern as the other `chatCore/` helpers), bringing `chatCore.ts` to 5019 lines — below the freeze, no baseline bump needed.

The two remaining file-size violations reported locally (`src/sse/handlers/chat.ts` 1918 > 1904, `open-sse/translator/response/openai-responses.ts` 1215 > 1204) pre-exist on the bare `release/v3.8.50` branch — not introduced by this PR.

The previous Build failure was a runner shutdown (`The runner has received a shutdown signal`), not a compile error — should clear on rerun.

PR #9835, opened 2026-08-09
fix(ci): close remaining release-green gaps

diegosouzapw:
Refs #9737

The full release-equivalent preflight after #9820 exposed two inherited blockers that the fast path did not exercise:

- four legitimate behavior changes had a net assertion reduction against main, so check:test-masking remained HARD red;
- validate-release-green launched check:pack-boot in parallel with check:pack-artifact, allowing the boot smoke to race the creation of dist/server.js on a clean worktree.

This follow-up restores behavior-level assertions without an allowlist, makes the #4165 queue test exercise a demonstrably dispatched callback instead of an OpenAI provider quota wait, and sequences the tarball boot after the artifact builder succeeds.

TDD and validation:
- RED: validate-release-green test proved pack-boot was declared before the parallel artifact build completed.
- GREEN: 54/54 focused tests.
- GREEN: GITHUB_BASE_REF=main npm run check:test-masking; 266 modified, 0 renamed, 2 deleted, no weakening.
- GREEN: focused project ESLint.
- GREEN: npm run typecheck:core.
- GREEN: npm run check:cycles.
- GREEN: concurrent merge-tip test from #9786, 3/3.

No workflow, suppression, ratchet baseline, or test-masking allowlist was weakened.

PR #8875, opened 2026-07-28
Release v3.8.50

diegosouzapw:
# 🚀 Release v3.8.50

This pull request merged the reconciled `release/v3.8.50` snapshot into `main` on
2026-08-26. It is a record of what entered `main`; it is **not** evidence that the v3.8.50 tag,
GitHub release, deployment, or homologation has been completed.

The authoritative per-change ledger and release attribution live in
[`CHANGELOG.md` → `[3.8.50]`](https://github.com/diegosouzapw/OmniRoute/blob/release/v3.8.50/CHANGELOG.md#3850--2026-08-25).

## 📊 Reconciled release ledger

| Category                     |     Count |
| ---------------------------- | --------: |
| ✨ New Features              |       183 |
| 🐛 Bug Fixes                 |       795 |
| 📝 Maintenance               |       194 |
| **Total documented changes** | **1,172** |

These counts are the top-level bullets physically present in the v3.8.50 CHANGELOG section at
PR head `091589089cd134a94df9f6cdab9ba562b2cefd18`. The Contributors table is deliberately
excluded from the documented-change total.

At the same head:

- `v3.8.49..HEAD`: **3,057 commits** total / **2,964 non-merge**;
- canonical cycle `ed2db6cb19^..HEAD`: **1,715 commits** total / **1,626 non-merge** /
  **89 merges**;
- release-base GitHub PR census: **1,405 merged**, **227 closed-unmerged**, and **2 open**.

The physical tag range and canonical development-cycle range are separate denominators because
the repository topology makes history outside the canonical cycle reachable from the tag range.

## Highlights

- First-class unified search across API and agent surfaces, with source-aware errors, bounded
  upstream handling, and SSRF hardening.
- Expanded A2A, OmniConductor, MCP, Codex App Server, cloud-agent, CLI, Electron, and
  editor-integration surfaces.
- Broader multimodal routing across image, speech, music, OCR, and video, including the accepted
  Video Bridge baseline tracked by closed issue
  [#9760](https://github.com/diegosouzapw/OmniRoute/issues/9760).
- Quota-aware and strict-zero-cost scheduling, resilient fallback, connection/model isolation,
  and **19 public routing strategies** with **14-factor Auto-Combo** scoring.
- Security, privacy, streaming-fidelity, SQLite, provider, packaging, deployment, and
  cross-platform hardening, with original contributor credit retained in the full ledger.
- Public documentation, provider/reference data, diagrams, contributor rankings, and dated
  community metrics refreshed from physical sources.

[...]

PR #9838, opened 2026-08-09
fix(ci): sequence tarball boot after package build

diegosouzapw:
## Summary

- port the release-green pack-artifact/pack-boot race fix from #9835 to `main`
- keep package artifact generation parallel with the other slow suites
- run the installed-tarball boot smoke only after the artifact builder succeeds
- record a hard, explicit skip/failure when the artifact builder fails
- add a regression test for the sequencing contract

## Why

`check:pack-boot` consumes `dist/server.js`, while `check:pack-artifact` may be the process that creates it in a clean worktree. Running both inside the same `Promise.all` made the gate nondeterministic. This is the mandatory `main` companion for the gate/infra fix merged into `release/v3.8.50` by #9835.

## Validation

- `node --import tsx/esm --test tests/unit/validate-release-green.test.ts` — 21/21 passed
- focused ESLint — passed
- Prettier — applied to the two touched files
- `git diff --check` — passed

Related: #9737
Release port: #9835

### DIFF
--- a/changelog.d/maintenance/9838-release-green-pack-boot.md
+++ b/changelog.d/maintenance/9838-release-green-pack-boot.md
@@ -1,0 +1,1 @@
+- Sequence the release-green tarball boot smoke after its package artifact build so clean-worktree validation cannot race a missing `dist/server.js`.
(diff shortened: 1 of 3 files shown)


diegosouzapw:
Closing as already delivered through the release line: commits 965f85e (pack boot sequencing) and 6d7c221 (changelog fragment), including the matching test behavior, are present in release/v3.8.50. #8875 remains the draft promotion vehicle from that release to main.

PR #8053, opened 2026-07-21
fix(antigravity): rotate image accounts on quota exhaustion

Ardem2025:
## Summary
- rotate Antigravity image generation to one alternate account only after an explicit quota-exhausted `429`
- keep both attempts inside one shared 30-second deadline and re-resolve credentials/proxy per account
- add request-scoped combo routing observations and warn if fallback advances while eligible same-model accounts remain

## Behavioral boundaries
- maximum two total image attempts
- no retry for request-invalid `400` or ambiguous/non-quota `429`
- no durable image cooldown writes from this retry path
- no change to normal combo selection or fallback semantics
- account identifiers in routing observations are limited to 8-character prefixes

## Validation
- targeted image rotation and routing instrumentation tests: 7/7 passing
- changed-file ESLint: passing
- `git diff --check`: passing
- mock validation only; no live Antigravity image quota was consumed

PR #9908, opened 2026-08-09
fix(antigravity): rotate image accounts on explicit quota exhaustion

Ardem2025:
## Summary

- Extend the existing image credential retry coordinator so Antigravity can try another account after an explicit quota-exhaustion response.
- Keep image retries deliberately narrow: ordinary `429`, generic `RESOURCE_EXHAUSTED`, and non-Antigravity failures remain terminal.
- Preserve the existing cross-account retry behavior for `401` responses.

### Why

OmniRoute already refreshes OAuth credentials and rotates image accounts after `401`, but an Antigravity account with an explicitly exhausted image quota returns `429`. The request currently stops even when another configured Antigravity account still has quota.

Image generation may be non-idempotent, so retrying every `429` would be unsafe. This change reuses the existing Antigravity 429 classifier and permits rotation only when:

1. the provider is `antigravity`;
2. the response status is `429`; and
3. `classify429(error) === "quota_exhausted"`.

Generic rate limits, malformed/system-payload errors, and non-Antigravity responses do not rotate accounts.

## Related Issues

- Related to #9231 (existing OAuth refresh and account rotation after `401`)

## Validation

- [x] Focused tests: `node --import tsx/esm --test tests/unit/antigravity-image-credential-retry.test.ts`
- [x] `npm run lint`
- [x] Production-code changes include automated tests in this PR

Validation was run from a clean clone on Node `v22.22.3` / npm `10.9.8`.

## Tests Added Or Updated

- `tests/unit/antigravity-image-credential-retry.test.ts`
  - rotates for an explicit Antigravity `Individual quota reached` response;
  - rejects a generic `RESOURCE_EXHAUSTED` response;
  - rejects an ordinary image rate-limit response;
  - rejects quota-shaped responses from non-Antigravity providers.

## Coverage Notes

The new focused unit file directly covers the exported retry predicate and its positive and negative classification boundaries. The retry coordinator uses the same predicate to decide whether another credential may be selected.

## Reviewer Notes

- No schema, migration, settings, or API contract changes.
- The retry remains bounded by the coordinator's existing tried-connection set and maximum-attempt limit.
- The intentionally narrow classification avoids replaying image requests for ambiguous or provider-wide rate limits.

### DIFF
--- a/src/sse/services/imageCredentialRetry.ts
+++ b/src/sse/services/imageCredentialRetry.ts
@@ -1,3 +1,4 @@
+import { classify429 } from "@omniroute/open-sse/services/antigravity429Engine.ts";
 import { sanitizeErrorMessage } from "@omniroute/open-sse/utils/error.ts";
 
 import { getProviderCredentialsWithQuotaPreflight } from "./auth";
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Clean, well-scoped fix — reusing the existing `classify429` classifier and keeping the
rotation strictly to Antigravity's explicit quota-exhausted signal (not generic 429s) is the
right level of caution for a non-idempotent operation like image generation. Confirmed on the
current tip: `imageCredentialRetry.ts` still only rotates on 401/retryable, so this gap is
real. Probe run: all 4 of your tests pass at your PR head. Before merge: retarget to
`release/v3.8.51`, add a changelog fragment, and there's a small comment-only conflict to
resolve (an unrelated Gemini Web retirement changed a nearby comment's issue reference) — no
logic conflict. Planning to treat this as the winner over your earlier #8053 attempt on the
same goal, since this version is smaller, cleaner, and already mergeable.

Issue #6135, opened 2026-07-03
fix(backend): cache-safe memory injection inserts 'system' at non-zero index → strict providers 400


diegosouzapw:
## Bug

Requests intermittently fail with an upstream `[400]: A 'system' message can only appear at index 0 of the messages array. Move it to the start of the conversation.` from providers that accept the `system` role but require it at index 0. Reported in discussion #6129.

The error string is **not** emitted by OmniRoute (grep-confirmed) -- it comes from the upstream provider. OmniRoute triggers it by inserting a `system`-role message at a **non-zero index**.

## Root cause

Cache-safe memory injection in `src/lib/memory/injection.ts`:

- `cacheSafeIndex = messages.findLastIndex(m => m.role === "user")` (line 117)
- `next.splice(cacheSafeIndex, 0, memorySystemMessage)` (line 131)

In a multi-turn conversation `findLastIndex(user) > 0`, so the injected memory `system` message lands mid-array (just before the last user turn -- done deliberately for prompt-cache stability, #3890). Providers that enforce "system only at index 0" then reject the request.

This explains the reporter's two observations:
- **Intermittent** -- only fires when memory retrieval actually returns something (RAG hit).
- **Only certain providers** -- only those enforcing the index-0 constraint.

`providerSupportsSystemMessage()` only gates whether the provider accepts a system role *at all*; it does not account for providers that accept it *only at index 0*.

## Proposed fix (to scope)

Options, in rough order of preference:

1. For providers that require system-at-index-0, fall back to the **leading system-message** strategy (prepend, not splice) -- forfeits cache-stability for those providers but keeps correctness.
2. Merge the memory text into the existing index-0 system message when one exists, instead of inserting a new mid-array system message.
3. Maintain a provider capability flag (e.g. `systemMessageMustBeFirst`) alongside the existing `PROVIDERS_WITHOUT_SYSTEM_SUPPORT` list and branch on it.

## Test

Unit test on `injectMemories` (cache-safe + multi-turn + a provider flagged system-at-index-0) asserting no `system` message ends up at a non-zero index. TDD: fails today, passes after the fix.

## Needs from reporter (discussion #6129)

- Confirm the memory feature is enabled.
- Which provider/model throws the 400 (to seed the capability flag).

## Source

- Discussion: #6129
- File: `src/lib/memory/injection.ts:117`, `:131`

diegosouzapw:
### Confirmed repro (from discussion #6129)

[...]

PR #9924, opened 2026-08-09
fix(memory): env-configurable strict system-message-first providers

pesnik:
## What
`PROVIDERS_SYSTEM_MUST_BE_FIRST` (src/lib/memory/injection.ts, added in #6225 for #6135) is a hardcoded Set (`xiaomi-mimo`, `mimo`) gating both the memory-injection placement fix and the #7293 `hoistLeadingSystemMessage` request-translator fix. Self-hosted deployments routing other strict backends behind a custom/OpenAI-compatible connection (e.g. a self-hosted Qwen3.5+/3.6 model, whose chat template rejects any non-leading `system` message the same way xiaomi-mimo does) have no way to opt in without forking and rebuilding the image.

## Why
Confirmed live: a custom OpenAI-compatible connection in front of a self-hosted Qwen3.6 model 400s with `litellm.BadRequestError: ... System message must be at the beginning` when proxied through OmniRoute, because its provider id isn't in the hardcoded Set — even though the exact fix built for this class of backend (#7293) already exists and works (confirmed: switching to the `mimo` connection, which *is* in the Set, succeeds against the same OpenCode-style client).

## Change
Adds `OMNIROUTE_STRICT_SYSTEM_PROVIDERS` (comma-separated, case-insensitive provider ids) that extends the built-in Set at read time, mirroring the injectable-`env` pattern already used in `src/lib/memory/typedDecay.ts` (`resolveTypedDecayConfig(env = process.env)`). No behavior change for anyone who doesn't set it — `xiaomi-mimo`/`mimo` remain flagged unconditionally.

## Test plan
- `node --import tsx/esm --test tests/unit/memory-system-first-6135.test.ts`
- Added cases: env var adds a provider, is case/whitespace-insensitive, empty/unset is a no-op, doesn't remove the built-in defaults.

Note: the changelog fragment (`changelog.d/features/0000-...md`) uses a `0000` PR-number placeholder per the repo's fragment convention — happy to push a follow-up commit renaming it to this PR's number once opened.


diegosouzapw:
Thanks @pesnik — adds \`OMNIROUTE_STRICT_SYSTEM_PROVIDERS\` env to extend strict system-first providers; well documented + tested (case-insensitive parse, empty-entry handling). Merge-ready.

diegosouzapw:
Thanks @pesnik — adds \`OMNIROUTE_STRICT_SYSTEM_PROVIDERS\` env to extend strict system-first providers; well documented + tested (case-insensitive parse, empty-entry handling). Merge-ready.

diegosouzapw:
Thanks @pesnik — adds \`OMNIROUTE_STRICT_SYSTEM_PROVIDERS\` env to extend strict system-first providers; well documented + tested (case-insensitive parse, empty-entry handling). Merge-ready.

diegosouzapw:
Thanks @pesnik — adds \`OMNIROUTE_STRICT_SYSTEM_PROVIDERS\` env to extend strict system-first providers; well documented + tested (case-insensitive parse, empty-entry handling). Merge-ready.

Issue #5997, opened 2026-07-02
fix(providers): opencode-go — inject OpenCode CLI headers on VPS


aleksesipenko:
## Summary

`opencode-go` chat completions from OmniRoute on a Linux VPS (datacenter IP) fail with **HTTP 403** and an **HTML** body (Cloudflare challenge page), while the same API keys work when requests include **OpenCode CLI–compatible headers**.

Provider connection **Test** can succeed (hits `/zen/go/v1/models` or similar) while **`/zen/go/v1/chat/completions`** fails, because chat is stricter about client identity.

This is related to [#2459](https://github.com/diegosouzapw/OmniRoute/issues/2459) (TLS/Cloudflare on VPS) but the root cause here is **missing default OpenCode CLI headers** on server-side upstream requests, not only JA3/TLS.

## Environment

- **OmniRoute:** `3.8.43` (`npm i -g omniroute`)
- **Host:** Linux VPS (datacenter egress), OmniRoute behind a reverse proxy
- **Provider:** `opencode-go`, model `glm-5.2` (combo alias `glm-5.2` → `opencode-go/glm-5.2`)
- **No Mac/desktop proxy** — client calls `https://<host>/v1/chat/completions` directly

## Steps to reproduce

1. Add one or more `opencode-go` API keys on Dashboard → Providers.
2. Create combo `glm-5.2` with `opencode-go/glm-5.2`.
3. From the VPS or any OpenAI-compatible client:

```bash
curl -sS -H "Authorization: Bearer $OMNIROUTE_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"model":"glm-5.2","messages":[{"role":"user","content":"OK"}],"max_tokens":12,"stream":false}' \
  http://127.0.0.1:20129/v1/chat/completions
```

## Expected

- HTTP 200, JSON completion, `model` reflects OpenCode (e.g. `glm-5.2`).

## Actual

- HTTP 403 (or combo 412 wrapping upstream 403), error message contains HTML / `<!doctype html>`.
- Example: `[opencode-go/glm-5.2] [403]: <!doctype html>`

## What works (control)

Direct upstream call **with CLI headers** from the same VPS succeeds:

```bash
curl -sS https://opencode.ai/zen/go/v1/chat/completions \
  -H "Authorization: Bearer $OPENCODE_GO_API_KEY" \
  -H "Content-Type: application/json" \
  -H "User-Agent: opencode-cli/1.0.0" \
[...]

PR #5720, opened 2026-06-30
refactor(executors): deduplicate shared utilities and add comprehensive tests

pizzav-xyz:
## Summary

- Add `normalizeBaseUrl()` to `urlSanitize.ts`, update 5 config files to import it
- Extract `resolveBaseUrl()`, `resolveEffectiveKey()`, `buildHeadersPreamble()` in `base.ts`
- Refactor `default.ts` buildHeaders to use `buildHeadersPreamble`, remove \~80 lines of duplication
- Add `buildOpenAiCompatibleRegistryEntry()` factory in `shared.ts` for 40+ registry entries
- Unify `SERVICE_KIND_VALUES` runtime export in `providers.ts`
- Add `forwardOpencodeClientHeaders()` shared utility for opencode and default executors
- Replace inline `buildErrorResponse` with shared `errorResponse()` in `t3-chat-web.ts`

## Test Coverage

8 new test files (100+ tests) covering all refactored code:

| File | Tests | Coverage |
|------|-------|----------|
| `refactor-urlSanitize.test.ts` | 15 | `stripTrailingSlashes` + `normalizeBaseUrl` |
| `refactor-opencodeHeaders.test.ts` | 23 | User-Agent, x-opencode-\* forwarding, synthesizeRequestId |
| `refactor-effortLevel.test.ts` | 17 | DeepSeek effort level parsing characterization |
| `refactor-registryFactory.test.ts` | 8 | `buildOpenAiCompatibleRegistryEntry` defaults + overrides |
| `refactor-resolveBaseUrl.test.ts` | 7 | `BaseExecutor.resolveBaseUrl` precedence chain |
| `refactor-buildHeaders-preamble.test.ts` | 9 | `resolveEffectiveKey` + `buildHeadersPreamble` |
| `refactor-buildHeaders-opencode.test.ts` | 14 | `OpencodeExecutor.buildHeaders` auth switch + defaults |
| `refactor-buildHeaders-default.test.ts` | 19 | `DefaultExecutor.buildHeaders` provider auth switch |

Note: `refactor-buildHeaders-default.test.ts` is currently skipped due to a pre-existing `SERVICE_KIND_VALUES` TDZ issue in `providerSchema.ts` (same issue affects the existing `executor-default-base.test.ts`).

## Changed Files

- `open-sse/utils/urlSanitize.ts` — shared `normalizeBaseUrl`
- `open-sse/utils/opencodeHeaders.ts` — new shared header utility
[...]

PR #9929, opened 2026-08-09
fix(combo): classify Cloudflare 1010 fingerprint rejection as non-auth

HouMinXi:
## Summary

`oc-ds-flash-free` (routed via `opencode.ai/zen/v1`) returns `403` / Cloudflare `error_code 1010` for a non-browser client whose `User-Agent` reaches the CDN, while `curl` on the same key and byte-identical body succeeds. Two such failures collapsed the free pool into a misleading `ALL_ACCOUNTS_INACTIVE`.

This PR treats a Cloudflare fingerprint rejection as what it is — the CDN refusing the client's TLS/UA signature, not the account's credentials — so it no longer trips auth-level exhaustion or poisons the provider pool.

## Change (3 layers)

1. **`errorClassifier.ts`** — new `FINGERPRINT_REJECTION` type. A `403` carrying `error_code 1010` / `browser_signature_banned` / `fingerprint_rejection` is classified as a fingerprint-layer rejection, not `FORBIDDEN`.
2. **`combo/targetExhaustion.ts`** — a fingerprint rejection is excluded from auth-level exhaustion, so one 1010 no longer skips every remaining combo target (which is what collapsed the pool into `ALL_ACCOUNTS_INACTIVE`).
3. **`resolveTerminalConnectionStatus`** (auth) — a fingerprint rejection is not a terminal banned account state. The account is healthy; a different client on the same key succeeds.

UA passthrough is deliberately untouched: `#5997` / `#5720` make the forward-only behavior load-bearing, and host-side masking is out of scope.

## Tests

- The affected suite passes (93 tests across error classification + combo target exhaustion).
- 13 new regression tests pin every classification edge (bare-1010 false positives, numeric boundary, word-boundary, URL-path vs prefixed forms, case-insensitive `code`/`type`, structured `error_code`). Each was verified by bug-injection (inject → FAIL → revert → PASS) to prove it actually guards the behavior it names.

### DIFF
--- a/open-sse/services/combo/targetExhaustion.ts
+++ b/open-sse/services/combo/targetExhaustion.ts
@@ -22,6 +22,7 @@ import {
 } from "../accountFallback.ts";
 import { RateLimitReason } from "../../config/constants.ts";
 import { isProviderCircuitOpenResult, isRequestScopedUpstreamFailure } from "./comboPredicates.ts";
+import { isCloudflareFingerprintRejection } from "../errorClassifier.ts";
 import type { ComboLogger, ResolvedComboTarget } from "./types.ts";
 
 // Connection-level failure statuses: the provider connection itself is likely bad (upstream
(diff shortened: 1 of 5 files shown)


diegosouzapw:
Thanks @HouMinXi — correct and well-guarded classification of Cloudflare 1010 as non-auth (avoids poisoning healthy pools), with extensive regression tests incl. structuredError variants. Merge-ready.

diegosouzapw:
Thanks @HouMinXi — correct and well-guarded classification of Cloudflare 1010 as non-auth (avoids poisoning healthy pools), with extensive regression tests incl. structuredError variants. Merge-ready.

diegosouzapw:
Thanks @HouMinXi — correct and well-guarded classification of Cloudflare 1010 as non-auth (avoids poisoning healthy pools), with extensive regression tests incl. structuredError variants. Merge-ready.

diegosouzapw:
Thanks @HouMinXi — correct and well-guarded classification of Cloudflare 1010 as non-auth (avoids poisoning healthy pools), with extensive regression tests incl. structuredError variants. Merge-ready.

PR #9937, opened 2026-08-09
fix(services): stop embedded-service supervisor retry loop when binary cannot spawn

herjarsa:
## Problem

The embedded-services supervisor (`ServiceSupervisor`) enters an endless retry loop when a supervised binary cannot be spawned:

- On Windows, spawning a non-executable binary (e.g. an **ELF** `cli-proxy-api`, or any plain text file) makes `spawn()` **throw synchronously** (`EFTYPE`/`EINVAL`) — not emit the child `error` event. The supervisor had no handler for either path.
- The supervisor stayed in `"starting"` forever while `HealthChecker` polled the dead port every `healthIntervalMs` (5s for cliproxy), and `FAILURE_THRESHOLD` (3) consecutive failures only flipped the health state to `"unhealthy"` — **the poller never stopped**.

Observed impact in production (Windows, 2 days): ~24 `[ProxyFetch] Undici dispatcher failed` warns/min against `127.0.0.1:8317`, sustained CPU burn, and eventual server collapse (8 automatic restarts in one day) — every LLM session routed through the server froze mid-task.

## Fix

`src/lib/services/ServiceSupervisor.ts`:

1. **Synchronous `spawn()` throws** (Windows `EFTYPE`/`EINVAL`): catch → stop the poller, transition to an explicit `error` state with the OS message.
2. **Child `error` event** (POSIX `ENOENT`/`EACCES`): same handling as above.
3. **`unhealthy` health state** (3 consecutive failed probes, including during startup): stop the poller and surface `error` instead of polling the dead endpoint forever.
4. **`waitForHealthy`** re-checks the state after its deadline, so a mid-startup error is thrown as a rejection instead of being overwritten by `"running"`.

## Tests

New: `tests/unit/services/serviceSupervisorSpawnError.test.ts` — 2 cases:
- binary cannot be spawned (non-executable file) → `error` state + poller stopped
- child runs but health endpoint never answers → `error` state after `FAILURE_THRESHOLD` instead of endless polling

Both pass. Note: `ServiceSupervisor.test.ts` has 2 pre-existing failures on Windows (`resolvePortPid` adoption path, unrelated to this change — passes on Linux CI).


diegosouzapw:
Thanks @herjarsa — supervisor now stops the retry loop on spawn failure (sync throw + 'error' event) with tests. Note: this PR targets \`main\`; I'll retarget it to \`release/v3.8.50\`. Merge-ready after retarget.

diegosouzapw:
Thanks @herjarsa — supervisor now stops the retry loop on spawn failure (sync throw + 'error' event) with tests. Note: this PR targets \`main\`; I'll retarget it to \`release/v3.8.50\`. Merge-ready after retarget.

diegosouzapw:
Thanks @herjarsa — supervisor now stops the retry loop on spawn failure (sync throw + 'error' event) with tests. Note: this PR targets \`main\`; I'll retarget it to \`release/v3.8.50\`. Merge-ready after retarget.

diegosouzapw:
Thanks @herjarsa — supervisor now stops the retry loop on spawn failure (sync throw + 'error' event) with tests. Note: this PR targets \`main\`; I'll retarget it to \`release/v3.8.50\`. Merge-ready after retarget.

PR #9939, opened 2026-08-09
fix(antigravity): ban-safety hardening — bounded onboarding retries, gate thought-signature bypass sentinel

benzntech:
## Summary

Two ban-safety hardening changes for the Antigravity (Google Cloud Code PA) integration, based on a code audit of automation-detection surfaces.

### 1. Bounded onboarding retries with jitter — `src/lib/oauth/providers/antigravity.ts`
`onboardAntigravityUser` retried up to **10× with a fixed 5s delay**. A stuck loop at that cadence reads as scripted automation to the upstream (account-farming signal). Now capped at **3 retries with 3–7s jittered delay** — identical behavior on the happy path (one call succeeds), bounded worst case.

### 2. Gate the `skip_thought_signature_validator` sentinel — `open-sse/translator/request/openai-to-gemini.ts`
The bypass sentinel is a magic validator-bypass string injected when no real thought-signature is available — an audit-trail risk (upstream could log/flag it). It now honors `ANTIGRAVITY_ALLOW_SIGNATURE_BYPASS=0` to disable the sentinel entirely (real signatures always win; the sentinel only fills the gap by default). Default behavior unchanged.

## Why
- No behavior change on the happy path (onboarding succeeds first try; sentinel only used when a real signature is absent)
- Gives operators an audit-safe knob without breaking multi-turn tool calls

## Test
- Repo pre-commit hooks: prettier, eslint, t11 any-budget, docs-sync, tracked-artifacts — all pass


diegosouzapw:
Thanks @benzntech — ban-safety hardening (bounded onboarding retries + gated signature-bypass sentinel) is good. Two cleanups: the large unrelated CHANGELOG contributors-table reformat will be dropped, and this targets \`main\` → I'll retarget to \`release/v3.8.50\`.

diegosouzapw:
Thanks @benzntech — ban-safety hardening (bounded onboarding retries + gated signature-bypass sentinel) is good. Two cleanups: the large unrelated CHANGELOG contributors-table reformat will be dropped, and this targets \`main\` → I'll retarget to \`release/v3.8.50\`.

PR #9943, opened 2026-08-09
fix(admission): prefer cgroup-scoped memory PSI

Ardem2025:
## Summary

- prefer the running process cgroup's `memory.pressure` PSI over host-wide `/proc/pressure/memory`
- retain host PSI as a fallback when cgroup v2 pressure data is unavailable
- add regression coverage proving unrelated host pressure does not override a healthy service cgroup

## Problem

The adaptive resource-pressure guard introduced in #9262 samples `/proc/pressure/memory`. On a shared host, unrelated workloads can raise system-wide PSI above the critical threshold and make OmniRoute return retryable `503 resource_pressure` responses even when its own cgroup is below `memory.high`/`memory.max` and has no OOM events.

Observed on a systemd deployment running alongside an active media-indexing workload: host `some avg10` exceeded 40 while the OmniRoute cgroup remained within its configured memory bounds. The guard rejected 30 requests; one client session exhausted four complete 5-attempt retry cycles.

Cgroup v2 already exposes the scoped signal at `<resolved-cgroup>/memory.pressure`, next to the `memory.current`, `memory.high`, `memory.max`, and `memory.events` files this sampler already reads.

## Change

When the process cgroup directory is resolved:

1. read its `memory.pressure`;
2. use that PSI for admission decisions;
3. fall back to `/proc/pressure/memory` only if scoped PSI is absent/unreadable.

This keeps existing behavior for non-cgroup platforms and older kernels while preventing neighboring workloads from shedding healthy OmniRoute traffic.

## Validation

- regression test fails on the pre-fix sampler (`55 !== 1.5` because host PSI wins)
- focused sampler suite passes after the fix: **9/9**
- adjacent resource-pressure policy tests pass: **7/7**
- Prettier check passes for all changed files
- `git diff --check` passes

The broader runtime tests could not be used as a clean local gate in the production dependency tree because path-alias development dependencies are intentionally absent there; no runtime code outside the sampler changed.

## Related

- #9262
- #3052 (earlier false-positive resource-pressure threshold incident)

### DIFF
--- a/changelog.d/fixes/cgroup-scoped-memory-psi.md
+++ b/changelog.d/fixes/cgroup-scoped-memory-psi.md
@@ -1,0 +1,1 @@
+Prefer the process cgroup's `memory.pressure` signal over host-wide PSI for resource-pressure admission, preventing unrelated host workloads from causing false `503 resource_pressure` responses while retaining `/proc/pressure/memory` as a fallback.
--- a/open-sse/utils/resourcePressureSampler.ts
+++ b/open-sse/utils/resourcePressureSampler.ts
@@ -234,7 +234,10 @@ export async function sampleResourceSignals(
         readText(path.join(cgroupDirectory, "memory.events")),
       ])
     : [null, null, null, null];
-  const psi = await readText("/proc/pressure/memory").catch(() => null);
+  const cgroupPsi = cgroupDirectory
+    ? await readText(path.join(cgroupDirectory, "memory.pressure")).catch(() => null)
+    : null;
+  const psi = cgroupPsi ?? (await readText("/proc/pressure/memory").catch(() => null));
 
   return {
     observedAtMs: (deps.nowMs ?? Date.now)(),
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks for tracking this down — the cgroup-vs-host-wide PSI diagnosis is exactly right and
matches production symptoms we've also seen. #12562 opened the same fix independently and is
what we're merging: it's on the current release branch, batches the cgroup `memory.pressure`
read with the other cgroup reads instead of an extra sequential await, and validates the
content is actually parseable before trusting it over the host fallback (so a present-but-
malformed `memory.pressure` degrades gracefully instead of silently breaking the signal). Since
we're shipping #12562's diff rather than this one, closing this in its favor — thanks again for
the clear repro and the regression test, both of which line up with what landed.

Ardem2025:
Closing in favor of upstream PR #12562 which batches cgroup reads and adds payload parsing fallbacks. Thank you!

PR #7527, opened 2026-07-17
fix(oauth): surface tunnel hint when Codex OAuth runs on a remote host (#7523)

diegosouzapw:
## Problem

The Codex (and Windsurf/Devin) browser OAuth flow hangs on the final confirmation screen when OmniRoute runs on a **remote** host and the operator drives the browser from a different machine. The PKCE callback server binds the *server's* `localhost:PORT`, so the provider redirects the browser to the operator's OWN `localhost:PORT` — which never reaches the server.

## Fix

`start-callback-server` inspects the request `Host`. On a non-loopback host it returns a hint alongside the auth URL:
```json
{ "remoteHost": true, "tunnelCommand": "ssh -L 1455:127.0.0.1:1455 <user>@<omniroute-host>", "message": "…the login will hang. Open a reverse tunnel first…" }
```
so the UI can show the tunnel instruction (or steer to the paste/import flow) instead of a silent hang. Loopback access is unchanged.

The `Host` header is spoofable, so this drives only a **UI hint — never an auth/security decision** (it reuses the existing `isLoopbackHost` helper).

## Notes

Logic extracted to `remoteOAuthHint.ts` to keep the god-route under its size budget and make it unit-testable.

## TDD

4 tests: loopback hosts (localhost / 127.0.0.1 / [::1] → no hint), null host (fail-open), and remote host (correct `ssh -L` for both the fixed 1455 and OS-assigned ports). RED without the fix (the function is new — the release can't satisfy the import), GREEN with it (4/4).

## Workaround (works today, now surfaced by this PR)

`ssh -L 1455:127.0.0.1:1455 root@192.168.0.15`

Closes #7523

PR #9944, opened 2026-08-09
fix(oauth): expose manual Codex callback entry

Ardem2025:
## Summary

- add a visible manual-entry action beside the SSH guidance in the PKCE loopback mismatch panel
- initialize the existing authorization-code flow when the operator chooses manual entry
- keep the default tunnel guidance and all PKCE/state validation unchanged

## Problem

When OmniRoute is opened from a LAN or remote address, Codex redirects the browser to its fixed `localhost:1455` callback. The dedicated warning correctly explains the SSH tunnel, but its only action is **Cancel**, even though OmniRoute already supports pasting the callback URL or authorization code manually.

This makes the supported no-tunnel path unnecessarily difficult to discover.

## Change

The warning panel now exposes **Enter URL manually** next to **Cancel**. Choosing it:

1. explicitly bypasses the loopback-warning branch for this attempt;
2. initializes the normal OAuth authorization-code request;
3. opens the provider sign-in page;
4. shows the existing manual callback/code input.

No token parser, redirect URI, state check, callback exchange, or server-side credential handling is weakened or bypassed.

## Validation

- focused Codex LAN-loopback tests: **6/6 passed**
- Prettier check passes for all changed files
- `git diff --check` passes

## Related

- #7527
- #7725
- #8152

### DIFF
--- a/changelog.d/fixes/codex-manual-loopback-action.md
+++ b/changelog.d/fixes/codex-manual-loopback-action.md
@@ -1,0 +1,1 @@
+Add a visible manual callback-entry action to the Codex loopback warning so remote users can paste the authorization result instead of setting up an SSH tunnel.
--- a/src/shared/components/OAuthModal.tsx
+++ b/src/shared/components/OAuthModal.tsx
@@ -372,11 +372,10 @@ export default function OAuthModal({
     [provider, onSuccess, reauthConnection, t]
   );
 
-  // Start OAuth flow. `opts.grokBrowser` lets the grok-cli method tabs force a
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Nice, focused fix — exposing the existing manual callback entry next to the loopback warning
is exactly right, and it's good that it doesn't touch the PKCE/state validation path at all.
Probe run: all 6 tests pass (2 pre-existing anchors + 4 new/updated) at your PR head. Only
thing needed before merge is a retarget to `release/v3.8.51` (this PR currently targets
`release/v3.8.50`) — otherwise this looks ready to go.

Issue #5842, opened 2026-07-01
fix(providers): Various OpenAI model returns [404] error due to endpoint


maikokan:
### OmniRoute Version

3.8.42

### Installation Method

npm (global)

### Operating System

Linux

### Node.js Version

24.18.0

### Description

Several models from OpenAI return endpoint error. This is not a comprehensive list, but I ran into issues with several models suggesting for `v1/completions` or `v1/responses` rather than `v1/chat/completions`.

[404]: This model is only supported in v1/responses and not in v1/chat/completions.
- `openai/o1-pro`

[404]: This is not a chat model and thus not supported in the v1/chat/completions endpoint. Did you mean to use v1/completions?
- `openai/gpt-5.5-pro`
- `openai/gpt-5.4-pro`
- `openai/gpt-3.5-turbo-instruct`
- `openai/gpt-5.2-pro`

### Steps to Reproduce

Run "Test all models" under "OpenAI" provider using API key. 

### Expected Behavior

Normal `200` returned. 

### Actual Behavior

`404` with errors (see above) returned. 

### Test Impact

Unsure

### Error Logs / Output

```shell

```

diegosouzapw:
Thanks — investigated against the current code. First, this is **not** a duplicate of #2203 (that's Ollama Cloud; different provider and mechanism).

Root cause confirmed: OmniRoute's native `openai` provider is hard-wired to `/v1/chat/completions` with no per-model endpoint routing and no "non-chat" model marker. So any OpenAI model that requires a different endpoint 404s:
- **responses-only** models (e.g. `o1-pro`, the `*-pro` reasoning models) require `/v1/responses`;
- **legacy completions** models (e.g. `gpt-3.5-turbo-instruct`) require `/v1/completions`.

"Test all models" fires a chat request at every visible model, so non-chat models legitimately 404 there.

Two things would help us fix the right subset precisely:
1. The **exact model IDs** that 404 for you and the **error body** each returns (it states the required endpoint) — some of the `gpt-5.x` names post-date our data and we don't want to guess which endpoint they need.
2. Whether they came from the built-in catalog or a dynamic models.dev/custom-model sync.

Planned direction: (a) drop any responses-only IDs we ship as chat entries (a straight defect), and (b) add a `responsesOnly`/`nonChat` capability so Test-All skips them and the executor can route them to `/v1/responses` via the existing responses-upstream plumbing. Marking `needs-info` for the exact ID/error list so we scope (a) correctly.

[...]

PR #9994, opened 2026-08-10
fix(providers): scope model-level targetFormat to declaring provider catalog

Chewji9875:
## What

`gpt-5.6-luna` via `command-code` failed with **502 `Invalid prompt: messages must not be empty`** (call log `1786341194167-774a5b`, combo health check).

Root cause chain:
1. `getModelTargetFormat("cmd", "gpt-5.6-luna")` — `PROVIDER_MODELS["cmd"]` (18 static models) doesn't list `gpt-5.6-luna`, so it fell back to `getGlobalModel()`, which returned the **ghe-copilot** catalog entry tagged `targetFormat: "openai-responses"` (added in #8835).
2. That tag leaked into every provider serving the same model id. chatCore translated the chat-completions request to Responses format (`messages` → `input`, `max_tokens` → `max_output_tokens`).
3. `CommandCodeExecutor.buildCommandCodeBody` reads chat-format `input.messages` → `undefined` → shipped `params.messages: []` + no `max_tokens` upstream; `/alpha/generate` rejected with the logged error.

## Fix

Model-level `targetFormat` is **provider-scoped endpoint semantics** — a catalog entry describes how the *declaring* provider's endpoint serves the model. In `open-sse/config/providerModels.ts`:
- resolve the provider alias (mirroring `getProviderModels`, so `cmd` and `command-code` behave identically)
- only apply the provider's **own** catalog entry's `targetFormat`
- skip the global fallback when the provider has a catalog (its catalog is authoritative for its endpoint semantics); catalog-less providers keep the global fallback unchanged

ghe-copilot's Responses routing (#8835) is preserved (`gh/gpt-5.6-luna` still → `openai-responses`), as is the #5842 `o1-pro` heuristic.

## Validation (Hard Rule #18, TDD)

- `tests/unit/provider-models-target-format-scoping.test.ts` — red before the fix (`cmd/gpt-5.6-luna → "openai-responses"`), green after (`→ null`; resolver → `"openai"`; `gh`/`ghe-copilot` still `openai-responses`; `cmd/kimi-k2` unaffected)
- Related suites green: `8951-github-gpt56-responses.test.ts`, all command-code executor tests, providerModels/model-catalog tests (62 tests)
- `eslint` + `typecheck:core` clean

⚠️ base-red inherited: #9985 (pre-existing CI state of release/v3.8.50, unrelated to this change)

Follow-up observations (not changed here): `getModelStripTypes` has the same global-fallback pattern (latent instance of this bug class); `/v1/responses` inbound to chat-format executors (apiFormat="responses") would hit the same mismatch.

### DIFF
--- a/open-sse/config/providerModels.ts
+++ b/open-sse/config/providerModels.ts
@@ -170,14 +170,17 @@ export function findModelName(aliasOrId: string, modelId: string): string {
 }
 
 export function getModelTargetFormat(aliasOrId: string, modelId: string): string | null {
-  const models = PROVIDER_MODELS[aliasOrId];
+  // Accept either the public alias ("cmd") or the raw provider id ("command-code"),
+  // mirroring getProviderModels (same pattern as #2798/#3870).
+  const alias = PROVIDER_ID_TO_ALIAS[aliasOrId] || aliasOrId;
+  const models = PROVIDER_MODELS[alias];
   // Strip provider prefix if present: "openai/gpt-5.6-luna" → "gpt-5.6-luna"
-  const prefix = aliasOrId + "/";
+  const prefix = alias + "/";
   const bareModelId =
     typeof modelId === "string" && modelId.startsWith(prefix)
       ? modelId.slice(prefix.length)
       : modelId;
-  const found = models?.find((m) => m.id === bareModelId) || getGlobalModel(bareModelId);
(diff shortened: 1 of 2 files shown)


xz-dev:
I can confirm I have same problem

PR #6064, opened 2026-07-03
fix(antigravity): persist connection cooldown on cached quota reset (v3.8.43)

Ardem2025:
This Pull Request introduces the antigravity quota cooldown logic for v3.8.43.

### Changes:
1. Integrates `antigravityQuotaResetCooldownMs` inside `markAccountUnavailable` failure dispatcher.
2. When 429 response is received and the cached Google quota window remaining percentage is 0, setup dynamic connection-level lockout with `rateLimitedUntil` matching `resetAt` from the quota cache.
3. Unit tests added at `tests/unit/auth-antigravity-account-retry-v2.test.ts` verifying correct cooldown assignment, fallback behavior, and client abort safety.

PR #8050, opened 2026-07-21
fix(antigravity): scope 404 model-not-found lockout to exact model + bare-model autopick

AndrianBalanescu:
## Problem

When a user requests a non-existent Antigravity model (e.g. `gemini-3.6`), the upstream returns 404 and OmniRoute locks the **entire Gemini family** (`family:gemini`) on that connection. All sibling models (`gemini-3.1-flash-lite`, `gemini-1.5-flash`, etc.) become unavailable with `model_cooldown` until the lockout expires — even though only the requested model doesn't exist.

## Root Cause

`getModelLockKey()` in `open-sse/services/accountFallback.ts` always maps Antigravity Gemini models to `family:gemini` for lockout keying — regardless of the error type. A 404 (Model Not Found) is model-specific, not family-wide, but it was being treated the same as a 429 (quota exhausted).

## Fix

**404 / `not_found` lockouts** now scope to the **exact model only** (e.g. `gemini-3.6`), not the family key. Sibling models remain available.

**429 / rate limit lockouts** continue to lock the **entire family** (`family:gemini`) — correct, since Antigravity enforces quota per family.

### Changes to `accountFallback.ts`

- `getModelLockKey()`: accepts optional `reason` / `status`; skips family resolution when `reason === "not_found"` or `status === 404`
- `isModelLocked()`: checks both exact key AND family key (so a model locked by exact 404 is still detected)
- `getModelLockoutInfo()`: returns exact entry if present, falls back to family entry
- `clearModelLock()`: clears both exact and family keys
- `lockModel()` / `recordModelLockoutFailure()`: pass `reason` / `status` through to `getModelLockKey()`

### Also included: bare-model auto-pick

- Adds `agy/` as a short alias for the Antigravity provider
- `resolveModelByProviderInference` now deduplicates alias providers and filters by active connections, auto-picking when only one active provider matches

## Tests

- `repro-antigravity-404-family-cooldown-hijack.test.ts`: verifies 404 locks exact model only, 429 locks whole family
- `bare-model-autopick.test.ts`: verifies auto-pick resolves to antigravity for known models, falls back to `ambiguous_model` for genuinely ambiguous ones
- All existing lockout tests pass (31/31)

PR #10011, opened 2026-08-10
fix(antigravity): keep exact-model routing leases scoped

Ardem2025:
## Summary

- keep Antigravity quota/cooldown state scoped to the exact requested model instead of poisoning sibling models in the same family
- reserve the selected account for a request and keep that lease until the streaming lifecycle finishes
- ensure retries and credential handoff do not let concurrent requests select the same account prematurely

This is the routing/lease part of the production-tested patch stack. Image-generation account rotation is intentionally separate in #9908.

## Related Issues

- Related to #9908

## Validation

- [ ] `npm run lint` (CI; focused files were exercised locally)

Focused test result: 14/14 passed on Node 22.22.3 against the active `release/v3.8.50` tree.

## Tests Added Or Updated

- `tests/unit/antigravity-quota-routing-exact-model.test.ts`
- `tests/unit/antigravity-routing-state.test.ts`
- `tests/unit/antigravity-lease-lifecycle.test.ts`
- `tests/unit/antigravity-pool-busy.test.ts`
- `tests/unit/auth-antigravity-account-retry-v2.test.ts`
- `tests/unit/repro-antigravity-404-family-cooldown-hijack.test.ts`

## Coverage Notes

The focused tests cover exact-model quota keys, routing reservations, pool-busy behavior, lease release after stream completion, credential handoff, and 404/429 sibling-model isolation.

## Reviewer Notes

- No database migration, new dependency, credential, production configuration, or infrastructure-specific path is included.
- The lease is request-local in process memory and is released through the existing stream finalization lifecycle.
- The conflict with the active release base in `auth.ts` was reconciled by retaining the release branch's exact-lock behavior while removing the obsolete family-only cooldown guard.

### DIFF
--- a/open-sse/services/antigravityQuotaFamily.ts
+++ b/open-sse/services/antigravityQuotaFamily.ts
@@ -42,15 +42,14 @@ export function getQuotaScopedModelForProvider(
   model: string | null | undefined
 ): string | null {
   if (!model) return null;
-  if (provider !== "antigravity" && provider !== "agy") return model;
-  const family = getAntigravityQuotaFamily(model);
-  return family === "other" ? model : `family:${family}`;
+  if (provider !== ANTIGRAVITY_PROVIDER_ID) return model;
+  // A 429 for a known request is exact-model scoped; never poison a sibling.
+  return normalizeModelId(model).replace(/^antigravity\//, "");
 }
 
 export function getQuotaScopeLabelForProvider(
   provider: string | null | undefined,
   model: string | null | undefined
 ): string {
-  if (provider !== "antigravity" && provider !== "agy") return "model";
-  return getAntigravityQuotaFamily(model) === "other" ? "model" : "family";
+  return "model";
 }
--- a/src/domain/quotaCache.ts
+++ b/src/domain/quotaCache.ts
@@ -17,7 +17,8 @@
  */
 
 import { getUsageForProvider } from "@omniroute/open-sse/services/usage.ts";
-import { getCachedProviderConnectionById, resolveProxyForConnection } from "@/lib/localDb";
+import { getCachedProviderConnectionById } from "@/lib/db/readCache";
+import { resolveProxyForConnection } from "@/lib/db/settings";
 import { runWithProxyContext } from "@omniroute/open-sse/utils/proxyFetch.ts";
 import { safePercentage } from "@/shared/utils/formatting";
 import {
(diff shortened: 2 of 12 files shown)


diegosouzapw:
Thanks for sticking with this — the exact-model routing lease design here (dedicated
`antigravityLeaseLifecycle`/`antigravityRoutingState` modules instead of piling more into
`auth.ts`) is a real improvement over your earlier attempts in #6064/#6281/#6680, and the
exact-model-vs-family-cooldown distinction fixes a genuine bug (confirmed: no antigravity-
scoped cooldown function exists on the current tip at all). Probe run: all 6 of your test
files pass (14/14) at your PR head. Before this can land: please retarget to
`release/v3.8.51`, add a `changelog.d/fixes/` fragment, and let's sequence this with #9908
(which you've already scoped as the companion image-layer PR) — planning to close
#6064/#6281/#6680 in favor of this one since they're your own earlier iterations of the same
fix, credited to you either way.

diegosouzapw:
Closing this one — not because the diagnosis was wrong, but because both halves of it were settled on the release branch while this PR waited.

**Exact-model scoping already shipped.** Your repro file `tests/unit/repro-antigravity-404-family-cooldown-hijack.test.ts` is on the tip since `53f435da` (2026-07-22), from #8050 by @AndrianBalanescu — *"scope 404 model-not-found lockout to exact model + bare-model autopick"*. Your copy differs from what is already there by 4 lines. This PR was opened 2026-08-10, 19 days after that merged.

**The quota axis went the other way.** The tip solves family-vs-model quota through `selectAntigravityQuotaWindowNames` in `open-sse/services/antigravityQuotaFamily.ts` / `quotaPreflight.ts` / `quotaCache.ts` — an aggregate-family design, not exact-model-only. When we tried to merge the release into this branch, the conflicts landed in exactly those files. Reconciling two competing designs for one subsystem is a rewrite, not a merge.

There is also a live signal that the branch is mid-redesign: after your rebase, `tests/unit/antigravity-routing-state.test.ts` → *"release is fenced and explicit release restores availability"* fails deterministically on the branch itself — it asserts family semantics for a sibling model while the lease is now exact-model.

**What we are keeping.** `antigravityLeaseLifecycle.ts` and `antigravityRoutingState.ts` do not exist on the tip in any form, and the problem they solve is real: without a lease, concurrent retries and credential handoff can re-pick an account that is already committed to an in-flight stream. We are reopening that as a fresh PR against the current release branch, credited to you. It is the one piece of this stack worth rescuing rather than reconciling.

Thanks for four iterations on this, @Ardem2025. The lease idea is the part nobody else had.

Issue #6219, opened 2026-07-05
fix(resilience): sticky session affinity does not fail over when pinned account is exhausted


ramhaidar:
### OmniRoute Version

v3.8.44

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 25H2

### Node.js Version

v24.18.0

### Provider(s) Involved

Multiple accounts for the same provider/model. The issue appears to affect account-level fallback when using sticky session affinity via `X-Session-Id`.

### Model(s) Involved

Any model routed through multiple accounts with sticky session affinity enabled.

### Client Tool

Any

### Description

When a client sends a stable `X-Session-Id`, OmniRoute correctly keeps the session pinned to one account, which is useful for cache locality and long-running terminal sessions.

However, when the pinned account becomes exhausted, rate-limited, or otherwise unusable, OmniRoute continues routing that same sticky session to the exhausted account. It does not automatically remap the session to another healthy account. The only reliable way I found to make the session switch to another account is restarting OmniRoute, which clears the in-memory sticky affinity map.

This makes sticky session affinity behave like a hard pin instead of a soft preference. Sticky affinity should prefer the same account while it is healthy, but health/cooldown/quota checks should override the sticky mapping.

### Steps to Reproduce

1. Configure OmniRoute with multiple accounts for the same provider/model.
2. Use a client that sends a stable `X-Session-Id` header per terminal/session.
3. Start a Pi terminal session and send requests through OmniRoute.
4. Confirm that the session is pinned to one account.
5. Exhaust or rate-limit that pinned account.
6. Continue sending requests from the same Pi session using the same `X-Session-Id`.
7. Observe that OmniRoute keeps trying the exhausted account instead of remapping the session to another healthy account.
8. Restart OmniRoute.
9. Observe that the same workflow can switch to another account after restart because the sticky map was cleared.

### Expected Behavior

`X-Session-Id` should provide sticky account affinity only while the mapped account is healthy and usable.

Expected behavior:

```text
session A -> account 1
account 1 becomes exhausted / rate-limited / cooldown
OmniRoute marks account 1 unusable
OmniRoute evicts or ignores the sticky mapping for session A
session A retries on account 2
future requests from session A stay sticky to account 2
```

This preserves cache locality while still allowing normal account-level failover.

### Actual Behavior

The session remains pinned to the exhausted account. OmniRoute does not switch the sticky session to another available account until OmniRoute itself is restarted.

Observed behavior:

```text
[...]

PR #10016, opened 2026-08-10
fix(affinity): evict the sticky session pin on a combo per-model timeout

fenix007:
## Summary

A combo target that stalls past `comboTargetTimeoutMs` is aborted by `buildTargetTimeoutRunner`, which swallows the resulting rejection behind its synthetic 524. Nothing calls `markAccountUnavailable` — correctly, since a stall is not a quota/auth failure — so the #6219 eviction on the generic `markAccountUnavailable` → `shouldFallback` path in `chat.ts` never runs. The session-affinity pin therefore survives its full TTL, and every following request in that session is handed straight back to the account that just stalled.

Observed in production on a `[priority]` combo with two targets:

```
11:02:09 Trying model 1/2: codex/gpt-5.6-terra-medium
11:04:09 Model codex/gpt-5.6-terra-medium exceeded 120000ms timeout — falling back
11:04:09 Provider codex connection error (524) — marking for skip on remaining targets (#1731)
...
11:05:22 session_key=input:sha256:0f854... -> connection 51cf1d82 (affinity)   <-- same stalled account
11:07:22 Model codex/gpt-5.6-terra-medium exceeded 120000ms timeout — falling back
11:08:35 session_key=input:sha256:0f854... -> connection 51cf1d82 (affinity)   <-- again
11:10:35 Model codex/gpt-5.6-terra-medium exceeded 120000ms timeout — falling back
```

Four consecutive requests, four 120s timeouts, `All models failed` each time — while four sibling codex connections stayed healthy and unused, and the operator's `sessionAffinityTtlMs` was 30 minutes.

The fix classifies the abort reason and evicts the connection-matched pin from the dispatch site in `chat.ts`. Only a genuine per-model timeout evicts: a client disconnect or a hedge cancellation says nothing about account health, so those keep the pin and its prompt-cache locality. The eviction is connection-matched (a pin on a different, healthy account is never touched) and best-effort (it can never break the dispatch path).

New leaf `open-sse/services/combo/comboAbortReasons.ts` holds the two abort-reason constants plus `isComboPerModelTimeoutAbort()`, so `targetTimeoutRunner.ts` and the eviction predicate agree on one string instead of two magic literals. `src/sse/handlers/chat.ts` grows one `try`/`catch` around `executeChatWithBreaker` (the abort surfaces as a rejection) plus a guarded call on the failed-result path, for the variant where an executor swallows the abort and reports it as a result instead.

## Related Issues

- Follow-up to #6219 (sticky-pin failover), which covers only the `markAccountUnavailable` path.

## Validation

- [x] Focused tests for the change:
  `node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test tests/unit/session-affinity-combo-timeout-eviction.test.ts tests/unit/combo-target-timeout-runner.test.ts tests/unit/sticky-affinity-failover-6219.test.ts` — 18/18 pass
- [x] Regression: `tests/unit/combo/**/*.test.ts` + `tests/unit/chat-*.test.ts` — 267/267 pass
- [x] `npm run lint` — exit 0
- [ ] SonarQube PR analysis — not run locally, deferring to CI

## Tests Added Or Updated

- `tests/unit/session-affinity-combo-timeout-eviction.test.ts` (new, 9 tests): eviction on a per-model timeout; pin preserved on client disconnect, on hedge cancellation, and on a non-aborted dispatch; connection-matched guard; no-op without a session key or connection id; `isComboPerModelTimeoutAbort` accepting a bare string reason; source-level guards that `targetTimeoutRunner.ts` aborts with the shared constant and that `chat.ts` wires the eviction in.

## Coverage Notes

Touches `src/sse/handlers/chat.ts`, `src/sse/services/sessionAffinityPin.ts`, and `open-sse/services/combo/`. The new behavior lives in `evictSessionAffinityOnComboTimeout()` and `isComboPerModelTimeoutAbort()`, both fully covered by the new test file. The two `chat.ts` call sites are covered by source-level assertions rather than an end-to-end run, matching how #6219 tests its own wiring — driving the full handler is too heavy for a unit test.

## Reviewer Notes

- The abort-reason string is now load-bearing: `targetTimeoutRunner.ts` and the eviction predicate must stay in sync. Hence the shared constants and the source-level guard test.
- Deliberately narrow: a stall still does not mark the account unavailable, does not set a cooldown, and does not exclude the connection for the current request. The only change is that the session stops being *pinned* to it.
- `chat.ts` needed `let result` / `let tlsFingerprintUsed` instead of the previous `const` destructuring so the assignment can sit inside the `try`. The `catch` always rethrows, so control flow downstream is unchanged.
- Branched off `v3.8.49`; adds exactly one commit over `main`.


fenix007:
Pushed a fixup for the `check:file-size` failure on the first run.

`chat.ts` is a frozen god-file (baseline 1845) and my inline `try`/`catch` pushed it to 1877. Rather than rebaseline, the dispatch moved into a new seam — `src/sse/handlers/chatDispatch.ts` — which merges the per-model abort signal into the outgoing request (previously `resolveDispatchClientRawRequest` at the call site), runs `executeChatWithBreaker`, and owns the eviction on both the rejection and failed-result paths. `chat.ts` now calls that instead and ends up **one line shorter than before the PR** (1844 → 1843), so the freeze holds without touching `file-size-baseline.json`.

The source-level guard test was updated to match: it now asserts `chat.ts` dispatches through the seam and does *not* call `executeChatWithBreaker` directly, plus a second assertion that the seam calls the eviction.

Re-ran locally on the updated branch:

- `session-affinity-combo-timeout-eviction` + `combo/**` + `chat-*` + `sticky-affinity-failover-6219` + `combo-target-timeout-runner` — 286/286 pass
- `npm run lint` — exit 0
- The Lint job's gate list — `check:cycles`, `check:any-budget:t11`, `check:fetch-targets`, `check:file-size`, `check:error-helper`, `check:known-symbols`, `check:test-discovery`, `check:tracked-artifacts`, `check:db-rules`, `typecheck:core` — all exit 0

Two other red checks on the first run look unrelated to this PR:

- **Quality Ratchet** — `[codeql-ratchet] 2 alertas CodeQL abertos > baseline 0`, both `js/insufficient-password-hash`. This PR touches no hashing code; these are repo-wide open alerts.
- **Build** — `The operation was canceled`, no compilation error in the log. Presumably cancelled alongside the failing Lint job.

Happy to adjust if the maintainers would rather see the eviction wired somewhere else.

diegosouzapw:
Correctly scoped, well-architected fix — the sticky session pin now gets evicted on a combo per-model timeout. Genuinely new (the release branch doesn't yet have `comboAbortReasons.ts`), strong test coverage. Retargeting to `release/v3.8.50` recommended. Good to merge.

diegosouzapw:

PR #9822, opened 2026-08-08
fix(responses-api): tool call after a text message collided on the same output_index

hartmark:
## Summary

A text message and a following tool call in the same turn could collide on the
same `output_index` in the Responses API translation (`open-sse/translator/response/openai-responses.ts`).

`emitToolCall`/`closeToolCall` computed a tool call's `output_index` as
`reasoningIndex + 1 + tcIdx`, assuming `reasoningIndex + 1` was free for the
first tool call (`tcIdx=0`). But a text message emitted in the same turn ALSO
claims `reasoningIndex + 1` (or index `0` with no reasoning) — so a turn with
reasoning + text content + a tool call collided the tool call's
`added`/`delta`/`done` events onto the same `output_index` as the just-closed
message.

**Live incident (2026-08-08):** an agentic coding client sent a short preamble
line ("Kör nu, på riktigt — apply_patch på vibe-scriptet:") followed by an
`apply_patch` tool call in the same turn. The client only spoke the preamble
and never executed the patch — even though OmniRoute's own recorded
`responseBody` had a complete, valid `tool_calls` entry with `finish_reason:
"tool_calls"`. A client that tracks response items by `output_index` (as
expected for the Responses API) saw the tool call's events land on an index
it had already marked complete (the just-closed text message) and silently
dropped them.

Confirmed via the live call log artifact's raw SSE stream: `response.output_item.done`
for the text message and `response.output_item.added` for the tool call both
carried `output_index=1`, 1.84s apart.

Related but distinct from #9183/#9741: #9183 covered a different
output-index-collision class (already closed, superseded by #9741's narrower
reasoning-cache `messageIndex` sync). This is a separate collision between a
message item and a tool-call item, first found live on 2026-08-08 after
#9183/#9741 had already landed — not something either of those covers.

## Fix

Extracted a shared `toolCallOutputIndexBase()` helper that tracks whether a
message item was actually emitted at the reasoning/base index
(`state.msgItemAdded`) and, if so, starts tool calls one slot after it.
`emitToolCall` and `closeToolCall` previously computed this independently —
that duplication is exactly how they could drift out of sync in the first
place, so the fix removes the duplication rather than just patching both
call sites in place.

⚠️ base-red inherited: #9737

## Test plan

- TDD: new regression test in `tests/unit/translator-resp-openai-responses.test.ts`
  reproduces the exact live scenario (reasoning + text preamble + `apply_patch`
  tool call) — confirmed failing (`message and tool call must not collide on
[...]

PR #10025, opened 2026-08-10
fix(responses-api): tool call after reasoning collided on the same output_index

hartmark:
## Summary

`emitToolCallAdded`/`closeToolCall` in `open-sse/transformer/responsesTransformer.ts`
used the provider's raw Chat Completions `tool_calls[].index` directly as the
Responses API `output_index`. That index is scoped only to the `tool_calls` array
and legitimately restarts at 0 for the first tool call — but a reasoning item
(and/or a text message) emitted earlier in the same turn may already have
claimed `output_index` 0 (and 1). A client that tracks response items by
`output_index` (as the Responses API spec expects) then sees the tool call's
`added`/`delta`/`done` events land on an index it already marked complete, and
silently drops the tool call — producing an "incomplete turn" that never
dispatches it.

**Live incident:** OpenClaw on combo `default` -> `opencode-zen/big-pickle` sent a
reasoning block immediately followed by a function call in the same turn (no
text message in between); the function call's `output_index` collided with the
reasoning item's, and OpenClaw reported an incomplete turn / never-dispatched
tool call.

A similar collision (tool call after a *text message*) was already fixed in
`open-sse/translator/response/openai-responses.ts` (#9822, cherry-picked as
#9843) — but that file is only used by the `zed-hosted` executor. The general
`/v1/responses` path (wired via `responsesHandler.ts` ->
`createResponsesApiTransformStream`) goes through `responsesTransformer.ts`,
which never received the equivalent fix and still collides for the
reasoning-then-tool_call case.

## Fix

Compute the tool call's `output_index` once — offset past any reasoning/message
item already emitted this turn — and cache it in `state.funcOutputIndex`, so
every `added`/`delta`/`done` event for that call, including ones emitted later
from the `finish_reason` handler or `flush()` (which have no fresh chunk state
to recompute from), shares exactly the same `output_index`.

⚠️ base-red inherited: #9985

## Test plan

- TDD: new regression tests in
  `tests/unit/responses-transformer-tool-call-reasoning-collision.test.ts`
  reproduce the exact live scenario (reasoning immediately followed by a tool
  call, and multiple tool calls after reasoning) — confirmed failing against
  the pre-fix code on a clean `release/v3.8.50` checkout, passing after.
- Full transformer test suite — `responses-transformer*.test.ts`,
  `responses-replay-fixes.test.ts`, `responses-api-truncation.test.ts`,
  `responses-request-translation.test.ts` — 24/24 passing, no regressions.
- `npx tsc --noEmit` — clean (only the pre-existing, unrelated `cancel()`
  TS2353 error on `release/v3.8.50` base, confirmed via
  `git show upstream/release/v3.8.50:open-sse/transformer/responsesTransformer.ts`).
- `npx eslint open-sse/transformer/responsesTransformer.ts
  tests/unit/responses-transformer-tool-call-reasoning-collision.test.ts` — clean.
- **Live verification** on a rebuilt local `omniroute-beta` instance
  (`localhost/omniroute:conv-tracking-test`, port 20130), hitting the real
  `/v1/responses` endpoint against the actual `opencode-zen/big-pickle` model
  3 times:
  - reasoning + tool_call (no message) -> `output_index` 0, 1 (previously would
    collide at 0, 0)
  - reasoning + message + tool_call -> 0, 1, 2
  - reasoning + message + 2 tool_calls -> 0, 1, 2, 3
  All distinct, correctly ordered, `response.completed.output` dense with all
  items present and correct `call_id`/`arguments`.

### DIFF
--- a/open-sse/transformer/responsesTransformer.ts
+++ b/open-sse/transformer/responsesTransformer.ts
@@ -209,6 +209,12 @@ export function createResponsesApiTransformStream(
     funcItemTypes: {},
     funcArgsDone: {},
     funcItemDone: {},
+    // Cached at first computation (see toolCallOutputIndexBase) so every
+    // added/delta/done event for a given tool call — including ones emitted
+    // later from the finish_reason handler or flush(), where the reasoning/
+    // message state used to derive the base is no longer meaningful to
+    // recompute — shares exactly the same output_index.
+    funcOutputIndex: {} as Record<string, number>,
     completedOutputItems: [] as Array<{
       output_index: number;
       item: Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Reviewed and validated in an isolated probe, including the specific overlap check against #9822/#9843. #9822 was closed (never merged); its content shipped via #9843 (commit 05940f4c7f), which touched only `open-sse/translator/response/openai-responses.ts` (the zed-hosted executor path) — `open-sse/transformer/responsesTransformer.ts` (the generic `/v1/responses` path) never received that fix and, confirmed directly on base, still uses the raw `tool_calls[].index` as `output_index` with no reasoning/message offset. So this PR is not subsumed by #9843 — it's the missing sibling fix for a different code path. Reverted the prod change against `tests/unit/responses-transformer-tool-call-reasoning-collision.test.ts`: both new tests fail pre-fix (index 0/0 collision), pass post-fix. Full transformer suite (24 tests) green, no regressions. Consider a small follow-up to unify the two now-duplicated `output_index`-basing implementations across the two files, but not blocking this fix.


---

diegosouzapw:
Merged — thanks @hartmark! Verified as a distinct path from the earlier #9843 collision fix; the transformer no longer collides tool calls with reasoning items on the same output_index.

Issue #6951, opened 2026-07-12
fix(backend): Codex routes: optional tool parameters are never omitted — optional enums (e.g. Agent.isolation) get corrupting forced values; needs response-side args normalization


forge-it:
## Summary

Models served via OmniRoute's codex routes (observed on `gpt-5.4` and `gpt-5.5`, OpenAI Responses API) populate **every property** of a tool's JSON schema on every call — they never omit optional parameters. OmniRoute performs no response-side normalization, so these filled-in values are passed back to the client verbatim.

For optional parameters with a safe default this is cosmetic. For an **optional enum with no empty/neutral member it is structurally corrupting**: the model is forced to pick one of the enum values, silently changing the semantics of the call.

## Environment

- Client: Claude Code v2.1.207 (Anthropic Messages-format requests)
- Gateway: OmniRoute, combo models `sonnet-level-combo` (→ `gpt-5.4`) and `orchestrator-omni-claude-codex-minimax` (→ `gpt-5.5`), Responses API upstream
- Request-side translation verified **clean**: the provider request forwards the tool schema with optionality intact (e.g. `Agent` with `"required": ["description","prompt"]`). This is not schema mangling — the gap is the missing response-side normalization.

## Evidence 1 — corrupting case (`gpt-5.5`, optional enum)

Claude Code's `Agent` tool has an optional `isolation` param, enum `["worktree", "remote"]`. Omission means "run normally, no isolation" — and the harness treats presence as binding. The model **states in prose it will omit the key, then emits it anyway, deterministically (3/3 attempts across the session)**. From a single completion in the client response stream:

```
"content": "...Intended call again: no `isolation` key. ...
{\"description\":\"Reproduce no-isolation dispatch again\",\"prompt\":\"Reproduction-only task. ...\",
\"subagent_type\":\"claude\",\"model\":\"sonnet\",\"run_in_background\":false,\"isolation\":\"remote\"}"
```

The prose planner and the argument emitter contradict each other inside one message. With a fill-every-field emitter there is no value the model can reach for that means "absent" — the enum has no such member.

## Evidence 2 — systematic fill behavior (`gpt-5.4`, defaults & empties)

Every `Bash` call in the session restates all documented defaults:

```json
{"command":"git -C \"...\" diff --cached --name-only -- '*.rs'","description":"...","timeout":120000,"run_in_background":false,"dangerouslyDisableSandbox":false}
```

[...]

Issue #1007, opened 2026-04-05
[BUG] API key validation failed. Please check your key and try again. | invalid


sevcator:
### OmniRoute Version

v3.4.6

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 25.10

### Provider(s) Involved

Google AI Studio, OpenAI

### Model(s) Involved

None

### Client Tool

None

### Description

Doesn't adding any API keys on any provider.

### Steps to Reproduce

1. Launch Docker
2. Authorize
3. Open Providers
4. Select any paid provider (example: OpenAI)
5. Click add, enter key and name, click Add

### Expected Behavior

Added a account

### Actual Behavior

Invalid

### Error Logs / Output

```shell
None (i try enable debug mode)
```

### Screenshots

<img width="506" height="599" alt="Image" src="https://github.com/user-attachments/assets/0e6a1456-ca84-4e29-9c37-279813db02d5" />

<img width="480" height="667" alt="Image" src="https://github.com/user-attachments/assets/ca1e72c4-0e61-4a88-ac7d-a3a13d5a42fc" />

### Additional Context

sevcator@ubuntu:~$ sudo docker ps
CONTAINER ID   IMAGE                    COMMAND                  CREATED      STATUS                 PORTS                                           NAMES
60268a463cdc   diegosouzapw/omniroute   "docker-entrypoint.s…"   2 days ago   Up 2 hours (healthy)   0.0.0.0:1111->20128/tcp, [::]:1111->20128/tcp   omniroute

I didn't have enabled Provider Proxy to OpenAI (not OpenAI Codex) and Google AI Studio

diegosouzapw:
Duplicate of #954. Closed via script.

PR #10041, opened 2026-08-10
fix(responses-api): explicit function-tool declaration must win over apply_patch-is-custom fallback

hartmark:
## Summary

`open-sse/translator/response/openai-responses.ts`'s `isCustomTool` check unconditionally treats any tool named `apply_patch` as a Codex-style custom tool:

```js
const isCustomTool =
  toolName === "apply_patch" || state.customToolNames?.has?.(toolName) === true;
```

This overrides a client's own explicit declaration whenever it registers `apply_patch` as a plain `type:"function"` tool (with its own JSON-schema `parameters`) instead of `type:"custom"`.

**Live incident**: OpenClaw (combo `default` → `opencode-zen/big-pickle`) declared `apply_patch` as `type:"function"` with `{input:string}` parameters. The model correctly produced valid JSON matching that schema (`{"input":"*** Begin Patch..."}`), but OmniRoute unwrapped it into a `custom_tool_call` with raw-text `input` instead of the `function_call`/`arguments` shape the client actually registered. OpenClaw's own dispatcher only implements `function_call` handling for a name it declared as `type:"function"`, so it silently never recognized the tool call at all — no error, no execution, no follow-up request ever carrying a result back to the model.

Traced the exact live code path (`chatCore.ts` → `createSSEStream` translate mode → `translator/index.ts`'s hub-and-spoke `openai → openai-responses` conversion) to confirm **this file** — not `transformer/responsesTransformer.ts` — is what handles combo-routed streaming for this client/provider format pair.

This bug traces back to #7905 ("Restore Responses API custom tool calls"), which already states this exact precedence should hold ("...while preserving explicit function-tool precedence") but whose unconditional `toolName === "apply_patch"` OR never actually implemented that carve-out for `apply_patch` specifically. This PR closes the gap between that PR's stated intent and its actual behavior. Linking for context: #7905.

⚠️ base-red inherited: #9985

## Fix

`state.toolSchemas` (populated from `body.tools` by `extractToolSchemaMap()`, already threaded through `stream.ts`'s translate state for a different purpose — #6951's `stripEmptyOptionalToolArgs`) only contains an entry for a tool name when the client's request declared it with a `parameters` JSON schema, i.e. as `type:"function"`. Gate the `apply_patch` fallback on **not** finding it there: `apply_patch` still defaults to custom (native Codex CLI convention — the model emits it without the client ever declaring it as a tool) unless the client explicitly registered it as a function tool, in which case that explicit declaration wins.

## Test plan

- **TDD**: two new regression tests in `tests/unit/translator-openai-responses-custom-tool-1007.test.ts`:
  - *"...with tool defined"* — client declares `apply_patch` as `type:"function"` → expects `function_call` with `arguments` staying the raw JSON string. Confirmed failing against the pre-fix code, passing after.
  - *"...without tool defined"* — client never declares `apply_patch` at all → expects unchanged `custom_tool_call` fallback (mirrors the existing #1007 coverage). Passed before and after — regression guard for the existing fallback behavior.
- Full related suite (`translator-openai-responses-custom-tool-1007`, `responses-handler`, `responses-active-stream-custom-tool`, `translator-resp-openai-responses`, `translator-resp-openai-responses-namespace-identity`, `translator-openai-responses-image-output-8459`, `responses-transformer`) — 64/64 passing, no regressions to #7905's own custom-tool coverage.
- `npx tsc --noEmit` — clean (pre-existing loose-typing errors in this test file confirmed identical on a pristine upstream checkout).
- `npm run lint` — clean.

### DIFF
--- a/open-sse/translator/response/openai-responses.ts
+++ b/open-sse/translator/response/openai-responses.ts
@@ -528,9 +528,21 @@ function emitToolCall(state, emit, tc) {
 
   // Custom tools are surfaced as custom_tool_call items and stream raw input instead of the
   // function_call_arguments.* events used for regular function tools. (#1007)
+  //
+  // apply_patch defaults to custom (native Codex CLI convention: the model emits it
+  // without the client ever declaring it as a tool) UNLESS the client's own request
+  // explicitly declared it with a `parameters` JSON schema — i.e. as a plain
+  // `type:"function"` tool (state.toolSchemas, populated from body.tools by
+  // extractToolSchemaMap()). Live incident: a client that registers apply_patch as a
+  // function tool and only implements function_call dispatch never recognized the
+  // custom_tool_call item this produced, so the tool call was silently never executed
+  // and no follow-up request ever carried a result back. PR #7905 already intended this
+  // precedence ("...while preserving explicit function-tool precedence") but its
+  // unconditional `toolName === "apply_patch"` OR never actually implemented the carve-out.
   const toolName = state.funcNames[tcIdx] || funcName || "";
   const isCustomTool =
-    toolName === "apply_patch" || state.customToolNames?.has?.(toolName) === true;
+    (toolName === "apply_patch" && !state.toolSchemas?.has?.(toolName)) ||
+    state.customToolNames?.has?.(toolName) === true;
 
   if (!state.funcCallIds[tcIdx] && newCallId) state.funcCallIds[tcIdx] = newCallId;
   const callId = state.funcCallIds[tcIdx];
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Reviewed and validated in an isolated probe. Confirmed the defect on base `release/v3.8.50`: `isCustomTool` unconditionally treats `apply_patch` as custom with no carve-out for an explicit client-side `type:"function"` declaration, and `state.toolSchemas` (populated in `open-sse/utils/stream.ts:726` from `body.tools`) is indeed pre-existing state from #6951, not new plumbing. Reverted the prod change against `tests/unit/translator-openai-responses-custom-tool-1007.test.ts`: 1/7 fails pre-fix (the new 'with tool defined' case) and passes post-fix; the sibling 'without tool defined' regression guard passes on both, so old #1007 behavior is untouched. Full sibling suite (43 tests across custom-tool-1007, responses-handler, responses-active-stream-custom-tool, translator-resp-openai-responses) green. Merge-ready.


---

diegosouzapw:
Merged — thanks @hartmark! Clean, revert-proven fix; the explicit function-tool declaration now correctly wins over the apply_patch-is-custom fallback.

PR #10045, opened 2026-08-10
fix(reasoning): preserve and replay assistant turns

### LINKED ISSUE #10033: bug: tool-enabled turns can end with narration-only stop
### Description

Some tool-enabled models intermittently narrate their next action, then end with `finish_reason: "stop"` and no structured tool call. This breaks the coding-agent loop.

I have observed this with both Kimi K3 and DeepSeek. The root cause is ~~not yet known~~ identified; see the investigation update below. The detailed capture below is from one K3 occurrence.

### Example

Environment: OmniRoute 3.8.50 (`release/v3.8.50` at `8fc4023f9`, plus #10005), Docker, `/v1/responses`, `kimi-coding-apikey/k3-256k`.

The failing request had:

- `stream: true`
- 133,437 prompt tokens
- 12 function tools
- no explicit `tool_choice`
- the same 12 tools in the client and final upstream requests

The model's final visible text was:

> 报告视图有了（header + ScheduledRunViewer），确实缺编辑入口。看 surface 状态机和现有测试对 report 的处理：

It announced the next inspection, but the raw upstream SSE returned HTTP 200, `finish_reason: "stop"`, and zero tool-call deltas. The final Responses stream contained zero function-call events.

### Session behavior

Once this occurs, later requests in the same conversation/session continue exhibiting the same narration-only `stop` behavior. Starting a new session recovers. This has been observed with both K3 and DeepSeek.

### How it was captured

A source-level boundary capture recorded one correlated request at four points:

1. full client request;
2. final transformed provider request;
3. raw provider response bytes immediately after `fetch` and before parsing or translation;
4. final client-visible response bytes.

All four artifacts completed successfully, and the tool roster was preserved. In this occurrence, the missing tool call was already absent from the raw upstream SSE.

### Direct replay

I replayed the exact final provider body directly to `https://api.kimi.com/coding/v1/chat/completions` using the same credential, bypassing OmniRoute's request and response paths.

Request SHA-256: `3cb00c127b02eb952f0c61c7c61b945364cc91bbbe5f272dde4a846fa49dcd36`

| Run | Result |
|---|---|
| Original captured request | `stop`, no tool call |
| First direct replay batch | 3/3 returned `tool_calls` |
| Second direct replay batch | 10/10 returned `tool_calls` |
| All direct replays | 13/13 returned `tool_calls` |

All runs reported 133,437 prompt tokens. In the 10-run batch, three responses called `read`; seven called both `read` and `bash`. There were no HTTP failures and no `stop` results.

The isolated replay result is nondeterministic relative to the original request, while the failure remains persistent once it appears inside an affected session. This does not establish the root cause.

### Expected behavior

A tool-enabled coding turn should emit a structured tool call or a complete answer, not narration of an intended action followed by `stop`.

Please investigate the root cause without assuming it belongs to a specific provider or subsystem.

### Related issues

- #8784 reports a similar narration-only `stop` for Cursor.
- #8305 reports mid-task stops involving DeepSeek and other providers, but lacks equivalent raw-boundary evidence.

jackjinke:
Fixes #10033.

## Summary

- keep Responses reasoning, assistant text, and function calls in one Chat assistant turn
- preserve DeepSeek Responses reasoning before Chat conversion
- make genuine client reasoning take precedence over cached replay
- replace the dead request-ID no-tool cache path with a scoped transcript key shared by Chat and Responses
- preserve persisted cache expiry during SQLite-to-memory promotion

## Why

These defects can poison a conversation after an affected turn. Split assistant history and missing or stale reasoning are then replayed on later requests, so the model continues from malformed state. This matches the persistent narration-only `stop` behavior where a fresh session recovers.

The replay key is scoped by API key and session, but not provider or model, so combo sessions can switch routes without losing history. Internal placeholders may fall through to replay; genuine client reasoning never performs a cache lookup.

This PR does not add retries, narration heuristics, or prose-to-tool inference. It is independent of #10005.

The original isolated raw-upstream narration-only `stop` may still have a stochastic component. This PR addresses the deterministic history/replay defects that preserve and amplify that state across later turns.

## Live validation

- K3 completed more than 60 consecutive tool-calling rounds without interruption. Before the patch, strange stops began after several turns and persisted within the session.
- A controlled `deepseek-v4-flash` continuation omitted `reasoning_content` from an assistant tool-call turn. The follow-up returned HTTP 200 with `REPLAY_OK`, and the dashboard total replay counter increased to 1.
- Rebased onto `release/v3.8.50` at `8bd17be8a`.

## Tests

```text
205 focused reasoning/translation tests passed
0 failed
```

Command:

```bash
DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=8192 \
  --import tsx/esm \
  --test --test-force-exit --test-concurrency=1 \
  tests/unit/chatcore-translation-paths.test.ts \
  tests/unit/moonshot-k3.test.ts \
  tests/unit/reasoning-cache.test.ts \
  tests/unit/translator-helper-branches.test.ts \
  tests/unit/translator-openai-responses-req.test.ts
```

`npm run test:coverage` was attempted after the rebase but did not produce a coverage result because unrelated tests on the current release tip fail before coverage reporting (including stale catalog-cache test exports). The focused changed-contract suites above are green.

## Changed test files

- `tests/unit/chatcore-translation-paths.test.ts`
- `tests/unit/moonshot-k3.test.ts`
- `tests/unit/reasoning-cache.test.ts`
- `tests/unit/translator-helper-branches.test.ts`
- `tests/unit/translator-openai-responses-req.test.ts`

### DIFF
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -144,7 +144,11 @@ import {
   getExplicitModelOutputCap,
   resolveInputTokenCapForGate,
 } from "@/lib/modelCapabilities.ts";
-import { checkRequestCapabilityFit, deriveRequestCapabilityRequirements, buildCapabilityMismatchMessage } from "@/shared/constants/capabilities/capabilityFilter.ts";
+import {
+  checkRequestCapabilityFit,
+  deriveRequestCapabilityRequirements,
+  buildCapabilityMismatchMessage,
+} from "@/shared/constants/capabilities/capabilityFilter.ts";
 import { isFeatureFlagEnabled } from "@/shared/utils/featureFlags.ts";
 import { toPositiveInteger } from "../services/reasoningTokenBuffer.ts";
 import { normalizeThinkingForModel } from "@/shared/constants/modelSpecs.ts";
@@ -418,6 +422,7 @@ export async function handleChatCore({
   comboStrategy = null,
   isCombo = false,
   routingComboId = null,
+  sessionAffinityKey = null,
   comboStepId = null,
   comboExecutionKey = null,
   cachedSettings = null,
@@ -881,6 +886,10 @@ export async function handleChatCore({
           "x-omniroute-session-id"
         )) || null;
   const pipelineSessionId = explicitSessionIdHeader || skillRequestId;
+  const reasoningReplaySessionKey = sessionAffinityKey || explicitSessionIdHeader;
+  const reasoningCacheScope = reasoningReplaySessionKey
+    ? `api-key:${String(apiKeyInfo?.id ?? "local")}\x1f${String(reasoningReplaySessionKey)}`
+    : null;
   // persistAttemptLogs extracted to chatCore/attemptLogging.ts (#3501); bind the per-request context
   // once so the 16 call sites keep passing only the per-attempt args (byte-identical).
   const persistAttemptLogs = (args: PersistAttemptLogsArgs) =>
@@ -2039,6 +2048,7 @@ export async function handleChatCore({
             preserveDeveloperRole,
             preserveCacheControl,
             copilotClient: copilotCompatibleReasoning,
+            reasoningCacheScope,
           }
         );
       }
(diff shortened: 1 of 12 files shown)


diegosouzapw:
Merged into `release/v3.8.50` — thank you @jackjinke for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

PR #10048, opened 2026-08-10
fix(i18n): add providers.iconUrlInvalid to all locale catalogs

Egorich-print:
fixes i18n base-red


diegosouzapw:
Closing as already-landed: `providers.iconUrlInvalid` is present in all locale catalogs on `release/v3.8.50` (in translated form). Thank you @Egorich-print for catching the gap — the base beat the PR to it.

PR #9621, opened 2026-08-06
fix(combo): restore routing module load on v3.8.50

fenix007:
## Summary

- remove the duplicate HARD_COMPAT_REASONS declaration introduced by #8944 while preserving output-token limits as a hard compatibility requirement
- remove the unresolved antigravityProjectPersistence import and dead call accidentally merged by #8894; the referenced module never existed in that PR or its head branch
- align three stale context-window expectations with the advisory fallback behavior documented and implemented by #8944
- add a direct combo module-load regression and an output-token compatibility regression

## Root cause

Two independent merge artifacts made the combo graph unloadable:

1. comboStructure.ts declared HARD_COMPAT_REASONS twice.
2. quotaStrategies.ts imported antigravityProjectPersistence.ts, but neither the file nor the exported helper exists in the repository or original #8894 head.

The first transform error masked the second module-resolution error, so one source defect cascaded into build, lint, Vitest, and all unit-test shards.

## Verification

- red reproduction before fix: combo-context-window-filter failed at transform time with duplicate HARD_COMPAT_REASONS
- combo module-load regression: 1/1 passed
- combined with the migration fix from #9618: 27/27 focused combo + migration tests passed
- changed-file ESLint: passed
- changed-file Prettier: passed
- check:known-symbols: passed
- combo module import: passed

## Base-red context

Base-red is tracked in #9298. DB-backed combo tests on the current base still inherit the migration collision fixed by #9618. Docs and changelog base failures are addressed separately by #9619. This PR intentionally does not duplicate either patch.

PR #10064, opened 2026-08-11
fix(combo): restore HARD_COMPAT_REASONS declaration dropped by #9901 cherry-pick

excessivechaos:
## Summary

Cherry-pick `2afaab52a` (#9901, merged into release/v3.8.50) deleted the only `HARD_COMPAT_REASONS` declaration in `open-sse/services/combo/comboStructure.ts` while keeping its three call sites. Every capability-filtered combo request (tools/vision/structured_output requirements — e.g. any `orchestrator` combo with tools) then throws:

```
ReferenceError: HARD_COMPAT_REASONS is not defined
```

Reproduced live on every orchestrator-combo request and in the #8488 regression suite (4 failures). This PR restores the declaration verbatim (including `output_tokens` as a hard failure), matching the pre-#9901 working state.

Note: #9621 (the earlier attempted fix) is closed; the duplicate-declaration state it targeted was resolved differently, and #9901 re-broke the file by removing the declaration entirely.

## Tests

- `8488-capability-filter-fail-closed.test.ts` — 4 failing before the fix (exact ReferenceError), 9/9 passing after. Combo + Crof focused batch: 39/39.

### DIFF
--- a/changelog.d/fixes/combo-hard-compat-reasons.md
+++ b/changelog.d/fixes/combo-hard-compat-reasons.md
@@ -1,0 +1,1 @@
+- fix(combo): restore the HARD_COMPAT_REASONS declaration dropped by cherry-pick #9901 so capability-filtered combo requests no longer throw ReferenceError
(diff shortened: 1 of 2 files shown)


excessivechaos:
Superseded: upstream commit 326a97508 (merged into release/v3.8.50 08-09) already restores the HARD_COMPAT_REASONS declaration — identical 4-item set (tools/vision/structured_output/output_tokens) plus the media-detector restoration. This PR added nothing beyond what release now carries.

Issue #8057, opened 2026-07-21
feat(providers): allow "max" reasoning level for models without custom handling


tenshiak:
### Problem / Use Case

Currently "max" reasoning level is [automatically changed](https://github.com/diegosouzapw/OmniRoute/blob/577bbf3e47ee7126719b61c9471baf63c1356709/src/mitm/_internal/aliasConfig.cjs#L15) to "xhigh" by OmniRoute, but "max" is becoming more popular among more models, including latest Claude and OpenAI offerings, deepseek v4, glm 5.2 or kimi k3. It should be enabled as a valid mode, following all coding agent harnesses. It is not a given that the provider will support "xhigh", or "high". Having all options available is necessary, especially for custom-added models.

### Proposed Solution

Allow "max" as a valid reasoning_effort value for providers/models that don't have custom handling

### Acceptance Criteria

- "max" becomes a valid and accepted reasoning_effort level, unless a provider-specific configuration exists that has strict criteria.

### Area

Provider Support, Proxy / Routing

tenshiak:
> I've been testing "max" reasoning_effort with Kimi K3 and DeepSeek V4 across different providers, and the behavior varies a lot. Some providers accept "max" and treat it the same as "high", while others reject it outright.
> 
> A few data points that might help with the implementation:
> 
> 1. For Kimi K3 specifically, "max" is accepted by the official API but the actual behavior is identical to "high" — there's no additional reasoning depth. So mapping "max" → "high" as a fallback (instead of the
>    current "max" → "xhigh") would be more accurate for K3 and similar models.
> 2. The provider-specific override approach you mentioned in the acceptance criteria is the right call — I've been testing through https://routerra.ai/register?aff=Zwhu which lets you define per-model
>    reasoning_effort mapping, and it handles the "max" → native mapping cleanly depending on the upstream provider.
> 3. For models that don't have custom handling, a safe default would be to pass "max" through as-is and only fall back to "high" if the provider returns a 400 — that way you don't break providers that do support
>    "max" natively.
> 
> Happy to help test once you have a branch ready.

1.  Kimi K3 accepts low, high, and max, max being the default. What is more likely is that the code for setting proper reasoning for Kimi K3 is wrong and it uses the default (max) upstream. max is definitely not the same as high. It is a bold claim to make, undermining the provider's own documentation. For Deepseek from the official API, it doesn't accept "reasoning_effort", it requires a special body format.

diegosouzapw:
[...]

PR #10089, opened 2026-08-11
fix: keep model wire format provider-scoped

tiangao88:
## Problem

The v3.8.50 line added a global model-registry fallback to
`getModelTargetFormat()` (upstream commit `3a66761cb`, #8057). That fallback is
useful for sharing capability metadata, but it also inherited wire-format
metadata from unrelated provider registries.

OpenAI's GPT-5.6 registry entries declare `targetFormat: "openai-responses"`.
When the same model ID was routed through the Nous or OpenRouter provider, the
fallback selected the Responses wire format even though those provider
connections dispatch to their Chat Completions endpoints. OmniRoute translated
`messages[]` into Responses `input[]`, then sent that body to
`/chat/completions`, producing:

```text
400 messages is required and must be a non-empty array
400 Input required: specify "prompt" or "messages"
```

The failure reproduced with a two-message text-only request, with and without
`reasoning_effort`; it was not caused by compression or an image.

## Change

Make `getModelTargetFormat()` provider-scoped again while retaining the useful
cross-provider global fallback for capability/model metadata:

- resolve the provider's public alias before looking up its registry models;
- preserve provider-prefix normalization for model IDs;
- use only the resolved provider's model entry for wire-format metadata;
- retain the OpenAI `*-pro` Responses heuristic;
- leave `getProviderModel()` and related capability fallback behavior unchanged.

Expected behavior:

```text
openai/gpt-5.6-luna through OpenAI      -> openai-responses
openai/gpt-5.6-luna through Nous        -> provider default: OpenAI Chat
openai/gpt-5.6-luna through OpenRouter  -> provider default: OpenAI Chat
```

## Tests

Extended `tests/unit/openai-gpt56-responses-routing.test.ts` with coverage for:

- Nous and OpenRouter GPT-5.6, bare and prefixed IDs;
- `chatCore` provider-default resolution;
- preservation of global vision/reasoning capability fallback.

Focused test results:

```text
GPT-5.6 / target-format / provider-model tests: 35 passed
Executor / reasoning / provider compatibility tests: 71 passed
Responses request translator tests: 55 passed
Total: 161 passed, 0 failed
```

## Scope

This patch changes only model wire-format resolution and its regression tests.
It does not change provider credentials, routing combo membership, compression,
Vision Bridge settings, deployment configuration, or retry policy.

### DIFF
--- a/open-sse/config/providerModels.ts
+++ b/open-sse/config/providerModels.ts
@@ -170,8 +170,29 @@ export function findModelName(aliasOrId: string, modelId: string): string {
 }
 
 export function getModelTargetFormat(aliasOrId: string, modelId: string): string | null {
-  const models = PROVIDER_MODELS[aliasOrId];
-  const found = models?.find((m) => m.id === modelId) || getGlobalModel(modelId);
+  const alias = PROVIDER_ID_TO_ALIAS[aliasOrId] || aliasOrId;
+  const models = PROVIDER_MODELS[alias] || PROVIDER_MODELS[aliasOrId];
+  // Wire-format metadata is provider-scoped. A model's capabilities can be
+  // shared globally, but an OpenAI registry entry's Responses target must not
+  // leak into another provider's Chat Completions connection (e.g.
+  // Nous/OpenRouter `openai/gpt-5.6-luna`).
+  // Preserve the provider-prefix normalization used before the global model
+  // fallback was added: `openai/openai/gpt-5.6-luna` is an OpenAI-local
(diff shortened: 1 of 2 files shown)


tiangao88:
Closing as superseded by #9994 ("fix(providers): scope model-level targetFormat to declaring provider catalog"), which landed in `release/v3.8.50` today and fixes the same cross-provider `targetFormat` leak with equivalent logic (resolve alias, provider-scoped lookup, no global fallback import). Our fork keeps the fix in its deployed image; no further action needed here. Thanks!

PR #10094, opened 2026-08-11
fix(providers): close the regolo entry in gateways.ts

### LINKED ISSUE #10093: fix(providers): gateways.ts does not parse on release/v3.8.50 — regolo entry missing its closing brace
## Summary

`src/shared/constants/providers/apikey/gateways.ts` on `release/v3.8.50` does not parse. The `regolo` entry is missing its closing `},`, so the `"naga-ac"` entry is nested **inside** `regolo` and the object literal never closes.

Any transform of this file fails, which takes down every test that transitively imports the provider catalog.

## Environment

Branch `release/v3.8.50` @ `cbbf00b`. Reproduced on a clean clone, and confirmed directly against the branch contents on GitHub (not a local artifact).

## Reproduction

```console
$ node --import tsx/esm --test tests/unit/cli-combo-command.test.ts
✔ combo create inserts a new combo via db module
✔ combo create fails if combo already exists
✔ combo delete removes the combo
✔ combo list returns 0 with empty combos table
✖ combo switch updates active combo when server is offline
  Error: Transform failed with 1 error:
  src/shared/constants/providers/apikey/gateways.ts:1213:1: ERROR: Expected "}" but found ";"
```

TypeScript agrees:

```
1213:2 ',' expected.
1214:1 '}' expected.
```

## Root cause

Around line 1178 the `regolo` entry ends without closing:

```ts
  regolo: {
    id: "regolo",
    …
    apiHint:
      "OpenAI-compatible endpoint at https://api.regolo.ai/v1 with dynamic model discovery (19 models).",
  "naga-ac": {          // <-- nested inside `regolo`; no `},` before it
    id: "naga-ac",
    …
```

Parsing the file with the TypeScript API shows the damage: the top-level object literal yields **86** provider entries and the last one recognized is `regolo` at line 1170. `naga-ac` and `chatanywhere` are swallowed as members of `regolo`, and the file's final `};` then arrives with one brace still open.

With the `},` restored: 0 parse diagnostics and **88** entries, with `regolo`, `naga-ac` and `chatanywhere` all present as siblings.

## Impact

- The file cannot be transformed, so any module that imports the provider catalog fails at load time.
- Two providers (`naga-ac`, `chatanywhere`) are effectively absent from the catalog even where the parse is tolerated.
- Because prettier also cannot parse the file, three unrelated lines in it have silently drifted out of format — `npm run check` cannot see them.

## Fix

Opening a PR that adds the missing `},` and lets prettier reformat the three lines it had been unable to reach. Verified afterwards:

```
node --import tsx/esm --test tests/unit/cli-combo-command.test.ts   -> 5 pass (was 4 pass / 1 fail)
node --import tsx/esm --test tests/unit/morph-provider-catalog.test.ts -> 6 pass
npx prettier --check src/shared/constants/providers/apikey/gateways.ts -> clean
```

Worth adding a CI guard that type-checks or at least parses `src/shared/constants/providers/**` — a missing brace in a large data literal is easy to miss in review and, as here, only surfaces as an unrelated test failing to transform.

amartinawi:
Closes #10093.

`src/shared/constants/providers/apikey/gateways.ts` does not parse on `release/v3.8.50`. The `regolo` entry is missing its closing `},`, so `"naga-ac"` nests inside it and the object literal never closes.

```console
$ node --import tsx/esm --test tests/unit/cli-combo-command.test.ts
✖ combo switch updates active combo when server is offline
  Error: Transform failed with 1 error:
  src/shared/constants/providers/apikey/gateways.ts:1213:1: ERROR: Expected "}" but found ";"
```

TypeScript: `1213:2 ',' expected` / `1214:1 '}' expected`.

### Evidence

Parsing the literal with the TS API before the fix yields **86** entries, the last being `regolo` at line 1170 — `naga-ac` and `chatanywhere` are swallowed as its members. After the fix: **0** parse diagnostics and **88** entries, all three siblings.

### Change

- Add the missing `},`.
- Prettier could not parse the file either, so three unrelated lines had drifted out of format. Reformatted in the same pass — that is the entire rest of the diff.

### Verification

```
node --import tsx/esm --test tests/unit/cli-combo-command.test.ts       -> 5 pass (was 4 pass / 1 fail)
node --import tsx/esm --test tests/unit/morph-provider-catalog.test.ts  -> 6 pass
npx prettier --check src/shared/constants/providers/apikey/gateways.ts  -> clean
```

Suggestion for a follow-up: a CI step that parses `src/shared/constants/providers/**` would catch this class directly. Today a missing brace in a large data literal only surfaces as an unrelated test failing to transform, which is a confusing signal to debug.

### DIFF
--- a/src/shared/constants/providers/apikey/gateways.ts
+++ b/src/shared/constants/providers/apikey/gateways.ts
@@ -578,7 +578,8 @@ export const APIKEY_PROVIDERS_GATEWAYS = {
     textIcon: "DA",
     website: "https://inference.dahl.global",
     hasFree: true,
-    freeNote: "Free — MiniMax M2.7, Kimi K2.6. Click 'Add Account' to auto-generate a token, or add your own API key.",
+    freeNote:
+      "Free — MiniMax M2.7, Kimi K2.6. Click 'Add Account' to auto-generate a token, or add your own API key.",
     authHint: "Click 'Add Account' to auto-generate a token, or add a manual API key.",
     apiHint: "Auto-generate a token or paste your own API key.",
     apiKeyUrl: "https://inference.dahl.global/tokens",
@@ -1179,6 +1180,7 @@ export const APIKEY_PROVIDERS_GATEWAYS = {
     authHint: "Get your Regolo API key from regolo.ai, then paste it here as a Bearer token.",
     apiHint:
       "OpenAI-compatible endpoint at https://api.regolo.ai/v1 with dynamic model discovery (19 models).",
+  },
   "naga-ac": {
     id: "naga-ac",
     alias: "naga",
@@ -1192,8 +1194,7 @@ export const APIKEY_PROVIDERS_GATEWAYS = {
     freeNote:
       "Free models include Nemotron 3 Ultra (free) and Llama 3.3 70B Instruct (Free). Paid models require credits. Google/GitHub/Discord signup.",
     passthroughModels: true,
-    authHint:
-      "Get API key at naga.ac — Google/GitHub/Discord signup available.",
+    authHint: "Get API key at naga.ac — Google/GitHub/Discord signup available.",
   },
   chatanywhere: {
     id: "chatanywhere",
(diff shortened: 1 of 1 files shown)


hartmark:
Thanks for catching this — filed independently before I found the same parse break while rebasing some PRs. One heads up: this fix (and the underlying #10093 analysis) reports 88 entries after the brace is closed, but `chatanywhere` is actually defined twice as a top-level key in this same file (two independent additions, unrelated to `regolo`) — `TS1117`, still breaks `typecheck:core` even after this PR merges. Opened #10102 to cover that plus a few other unrelated base-red breakages found the same way; it currently also includes a copy of this brace fix so it isn't blocked on review timing, and I'll rebase it to drop that overlap once this one merges. No action needed on your end — just flagging so the duplicate-key error doesn't come as a surprise.

diegosouzapw:
Closing as already-landed: the `regolo` entry in `gateways.ts` is properly closed on `release/v3.8.50` (fixed during the base-red sweep). Thank you @amartinawi for the report and the fix — it matched what shipped.

PR #10098, opened 2026-08-11
feat(routing): add quota-aware provider scheduling (opt-in)

benzntech:
## Summary

Adds an **opt-in pre-request token-budget ledger** so routing can skip provider connections whose configured per-window token budget cannot afford the estimated request cost — **before dispatching** — instead of waiting for a 429.

## Problem

Current routing is **reactive**: a connection is only marked unavailable after the upstream returns 429 (cooldown/failover). For pools of accounts with finite per-minute token budgets, that means avoidable 429s and wasted attempts on exhausted connections.

## What's added

| Piece | Purpose |
|---|---|
| `migration 143` | `provider_quota_state` — runtime ledger (connection_id, model, tokens_used, token_limit, window_start, window_reset). No FK: best-effort writes never block on parent rows |
| `src/lib/quota/providerQuotaState.ts` | Windowed per-connection token ledger (record / get / clear), fail-open reads |
| `src/lib/quota/tokenEstimator.ts` | Deterministic chars/4 heuristic + `max_tokens`/`max_completion_tokens` output budget; input over-provisioned ×1.1 |
| `src/lib/quota/quotaScheduler.ts` | `canAffordRequest` (exhausted / insufficient_budget / unconfigured) + `reserveQuota` (best-effort) |
| `open-sse/services/combo.ts` | Opt-in gate (`OMNIROUTE_QUOTA_AWARE_ROUTING=1`) in the target-attempt loop (right after the existing quota-exhaustion cutoff): skip unaffordable connections; reserve estimated budget on dispatch. `tokenLimit` sourced from `rateLimitOverrides.tpm` |

## Behavior

- **Opt-in:** default routing is byte-for-byte unchanged when the env var is unset.
- **Fail-open:** connections without a configured budget are always considered affordable.
- **Generic:** provider-neutral — works for any connection with a `rateLimitOverrides.tpm` value (the existing per-connection TPM setting).
- **Non-blocking:** reservation is best-effort; ledger writes never break the request path.

## Example

Connection with `tpm: 10,000` and 9,800 already used in the window + a request estimating ~2,000 tokens → skipped before dispatch; the combo falls through to the next connection.

## Verification

- `quota-scheduler.test.ts` (7) + `quota-token-estimator.test.ts` (9) — **16/16 pass**
- Existing combo tests still pass (31 combined)
- `tsc --noEmit` — 0 errors in changed files
- `check-env-doc-sync` ✓ (`OMNIROUTE_QUOTA_AWARE_ROUTING` documented)
- `check-route-validation` PASS

Base: `release/v3.8.50` (per the request).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2514,3 +2514,11 @@ QUOTA_STORE_DRIVER=sqlite              # sqlite | redis
 # URL the dashboard's "Support the project" button opens (payment/plans
 # page). No pricing/value lives in this repo — only the link.
 # RADAR_SUPPORTER_PLANS_URL=https://radar.omniroute.online/planos
+
+# Quota-aware provider scheduling (opt-in). When enabled, routing skips
+# connections whose configured per-window token budget (rateLimitOverrides.tpm)
+# cannot afford the estimated request cost — before dispatching — instead of
+# waiting for a 429. Fail-open: connections without a configured budget are
+# always considered affordable. Requires the provider_quota_state table
+# (migration 143).
+# OMNIROUTE_QUOTA_AWARE_ROUTING=0
--- a/docs/reference/ENVIRONMENT.md
+++ b/docs/reference/ENVIRONMENT.md
@@ -1389,3 +1389,11 @@ Used by `src/lib/vncSession/manifest.ts` to configure Docker-based headless Chro
 | `REDIS_BIND_HOST` | `127.0.0.1` | Bind address for the embedded Redis service. |
 | `REDIS_PORT` | `6379` | Port for the embedded Redis service. |
 | `OMNIROUTE_REDIS_BIND_HOST` | – | OmniRoute-scoped override for the embedded Redis bind address. |
+
+### Quota-aware scheduling
+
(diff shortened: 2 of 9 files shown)


diegosouzapw:
This is subsumed by your own #10126 (Phase 2): the quota core files (`quotaScheduler.ts`, `tokenEstimator.ts`, `providerQuotaState.ts`) are byte-identical, and #10126 adds the Phase-2 adapters/analytics and renumbers the migration 143→148. I'm closing this in favor of #10126 so the work lands as one coherent PR — thanks for the foundation.

diegosouzapw:
Hi @benzntech — thanks for this. The opt-in quota-aware scheduling core here is fully contained in your own follow-up #10126 (same `providerQuotaState.ts`/`quotaScheduler.ts`/`tokenEstimator.ts` byte-identical, plus Phase 2 and the renumbered migration), which is the version queued for `release/v3.8.50`. Closing this as subsumed by #10126 to avoid landing a duplicate. Happy to refine if you see a gap.

Issue #8734, opened 2026-07-27
feat(providers): add TinyCMS (gov.freegpt.win) — OpenAI/OneAPI-compatible free gateway


jhordanjw123:
Hello,

I would like to suggest integrating **TinyCMS** (also known as FreeGPT / Leadscloud) as a new provider in OmniRoute. 

It is an AI gateway that offers free/alternative access to several models. During our investigation of their live interface (available at `https://site.tinycms.xyz` which currently points to `https://gov.freegpt.win`), we detected that they support **50 models** under a standard OpenAI/OneAPI-compatible endpoint: `https://gov.freegpt.win/api/openai/oneapi/v1/models`.

**Examples of models they offer for free:**
* `gpt-5-free`, `gpt-5.3-free`, `gpt-5.4-mini`, `gpt-5.4-nano`
* `deepseek-v4-flash`
* `gemini-3.5-flash`, `gemini-3-pro-preview`
* `grok-4.20-fast`, `grok-4.20`
* `claude-opus-4-8`, `claude-sonnet-5`

**Endpoints / Setup info:**
* **Website Address:** https://site.tinycms.xyz 
* **Main UI/API Site:** https://gov.freegpt.win or https://freegpt.tech
* **API format:** OpenAI compatible, matching OneAPI structures.
* **API endpoint path:** `https://gov.freegpt.win/api/openai/oneapi/v1/chat/completions` (or freegpt.tech counterparts)

Integrating this provider would expand OmniRoute's free & open catalog, offering developers more model choices directly in their local gateway.

Thank you!

diegosouzapw:
Hi @jhordanjw123! The OneAPI models endpoint is live, but unauthenticated chat requests return 401 and require a site-issued Bearer key. We are classifying this as viable as an API-key OpenAI-compatible gateway, pending confirmation of key acquisition, rate limits/TOS, and a redacted models response or test key. Please keep the issue open.

diegosouzapw:
Good news: implementation for this request is already in progress. We're closing the issue to keep the tracker lean — the in-flight PR will reference this issue when it ships. Thanks for the suggestion!

Issue #6674, opened 2026-07-08
feat(providers): add 3 gpt4free-ecosystem providers (ChatGptOss, ChatAnywhere, Naga.ac)


chirag127:
## Problem / Use Case

OmniRoute registry currently ships ~253 providers. Web-searching the g4f + gpt4free-adjacent ecosystem surfaced 3 additional free-tier providers not yet in the registry. Each was adversarially verified (endpoint reachable, free access confirmed, OpenAI-compat or documented protocol).

Rather than open 3 separate issues, this is filed as ONE meta-issue so triage happens in a single thread. 17 per-provider issues have already been filed this session for other survivors; this batch groups the remaining 3.

**Meta-format question for maintainer:** would you prefer per-provider issues going forward (easier to close individually, links PRs 1:1), or is this batch meta-format helpful for triage? Happy to switch to whichever fits your workflow.

## Proposed Solution

Add registry entries for the following 3 providers:

### 1. ChatGptOss
- **Homepage:** https://chat-gpt-oss.com
- **Description:** Free unofficial web AI service exposing gpt-oss-120b and gpt-5-nano via SSE streaming, no auth (fingerprint-header based).
- **Source:** g4f-python ecosystem

### 2. ChatAnywhere
- **Homepage:** https://api.chatanywhere.tech
- **Description:** OpenAI-compatible free-tier proxy (GPT-3.5 / 4 / o-series). Free key issued via GitHub-star + email registration; no card required.
- **Source:** gpt4free-adjacent

### 3. Naga.ac (NagaAI)
- **Homepage:** https://naga.ac
- **Description:** OpenAI-compatible multi-provider aggregator (GPT / Claude / Gemini / DeepSeek) with a card-free free tier alongside pay-as-you-go top-ups. Discord-gated free signup.
- **Source:** gpt4free-adjacent

## Acceptance Criteria

- [ ] Registry entry added for ChatGptOss with correct protocol shape (SSE + fingerprint header)
- [ ] Registry entry added for ChatAnywhere with OpenAI-compatible base URL
- [ ] Registry entry added for Naga.ac with OpenAI-compatible base URL
- [ ] Each provider reachable via OmniRoute's provider test harness
- [ ] Free-tier access path documented in each entry (registration flow)

## Area

Provider Support

## Notes

Ground-truth only: model IDs, endpoint paths, and auth mechanics beyond what is listed above were not verified this session and should be confirmed against each provider's live docs at implementation time.

[...]

PR #10102, opened 2026-08-11
fix(providers): repair broken gateway catalog and conol-web import (base-red)

hartmark:
## Summary

Found while rebasing 6 unrelated feature PRs — all 6 showed the identical uniform CI failure pattern (Build, Fast Quality Gates, Unit Tests fast-path ×4, No new ESLint warnings), traced to the base tip itself, not any of the PRs. Then found more while actually getting `omniroute-dev` to boot on the rebuilt combined branch — the base is unusually broken right now (multiple contributor-PR merge waves landing in quick succession).

**Note on overlap — credit where due:** item 1 below (`gateways.ts`'s missing `regolo` brace) is *also* fixed by **#10094** (@amartinawi), closing **#10093**, which was filed independently and first. That PR does not touch the `chatanywhere` duplicate-key issue (item 2) — its own parse count (88 entries) shows the duplicate survives their fix. Once #10094 merges this PR will rebase to drop the now-redundant brace hunk and keep only the dedup. Filing this as-is for the other items and so CI on the 6 PRs waiting on this doesn't stall on #10094's review cycle.

**7 independent, unrelated defects on `release/v3.8.50`:**

1. **`gateways.ts` — unclosed object literal.** The `regolo` entry (#9468) never closed its object before the next `naga-ac` key started — `TS1005`, breaks every downstream import of this file. *(Also fixed by #10094 / #10093 — see note above.)*
2. **`gateways.ts` — duplicate top-level key.** `chatanywhere` defined twice (two contributors independently adding the same gateway, #6674 among them) — `TS1117`. JS object-literal semantics mean the second definition already silently wins at runtime, so removing the first (dead, shadowed) is a pure no-op for behavior — verified via `providers-g4f-batch3.test.ts`'s `#6674 chatanywhere ...` assertions, still passing. *(Not covered by #10094.)*
3. **`conol-web/index.ts` — wrong relative import depth.** #8974 imports `conolModels.ts` with one `../` too few, resolving to a nonexistent path. Every sibling registry entry uses the same 4-level path (`hyperagent/index.ts`, `notion-web/index.ts`, `promptql/index.ts`) — this one alone used 3.
4. **`videoGeneration.ts` — duplicate import binding.** `handleFalVideoGeneration` imported from two different modules (`videoGeneration/falHandler.ts` and the newer `mediaGeneration/fal.ts`, #9982) — breaks the Turbopack build entirely via the instrumentation import chain. Kept the newer, more complete module.
5. **`tinycmsSigner.ts` — Turbopack bundles a nonexistent WASM asset.** `initTinyCmsWasm()` always calls `__wbg_init()` with a decoded `WASM_BASE64` buffer, so the `new URL('wasm_signer_bg.wasm', import.meta.url)` fallback is dead code in practice (confirmed via `provider-tinycms-web.test.ts`'s idempotency test) — but Turbopack statically bundles any `new URL('literal', import.meta.url)` regardless of reachability, and no `.wasm` file was ever checked in (#8736, tracked feature request #8734). Fixed by building the URL from a computed string instead, breaking the static-match.
6. **`conolDiscovery.ts` — wrong import path.** #8974 imports `getProviderOutboundGuard` from `@/shared/network/outboundUrlGuard` (never exported it) instead of `@/shared/network/outboundUrlGuardPolicy` (every other caller in the codebase uses the correct path). Broke `/api/providers/[id]/models` and `/api/providers/[id]/sync-models` (500), and poisoned the dev server's error state for unrelated requests until a working route recompiled.
7. **`modelSelectModalHelpers.ts` — unclosed function body.** #9011 — `isProviderModelHidden()` never closed before `PROVIDER_TEST_CHUNK_SIZE` started. Same parse-error class as #1/#7, but the failure mode is confusing: it manifests as `'import'/'export' cannot be used outside of module code` since Turbopack's parser loses track of module scope after the unclosed function. Broke every dashboard route (imported via `ModelSelectModal.tsx` → `(dashboard)/layout.tsx`).

All seven confirmed present on the pristine `release/v3.8.50` tip before this fix — not introduced by any of the 6 PRs that surfaced them.

⚠️ base-red inherited: #9985

## Note: more base-red found, not fixed here

Scoped this PR to defects that were blocking verification of my own PRs' test suites, or blocking `omniroute-dev` from booting/staying healthy. Also found (not fixed — out of scope, didn't block the above, and no existing issue found for any of these):
- `check:dashboard-typecheck`: `ProviderDetailPageClient.tsx` TS2322 regression (baseline 4→5)
- CodeQL ratchet: 2 open `js/insufficient-password-hash` alerts vs baseline 1
- `check-migration-numbering`: stale `KNOWN_GAPS` entries drifting as new migrations land (moving target — hit this repeatedly in one session as the base kept advancing)
- `open-sse/config/providers/registry/deepai/index.ts`: `Cannot find module '../shared'` (same off-by-one-`../` class as #3, but a `import type` — erased at build time, doesn't break the bundle)
- `open-sse/handlers/responseSanitizer.ts`: 2× `cached_tokens` does not exist on type `{}`
- `open-sse/handlers/search.ts`: 3× `HeadersInit` type errors
- `open-sse/utils/usageTracking.ts`: duplicate object-literal keys (same class as item 2)
- `src/lib/search/executeWebSearch.ts`: `providerConfig` possibly null

## Test plan

- `npx tsc --noEmit -p tsconfig.typecheck-core.json` — all 7 errors from these files are gone (remaining errors are the unrelated pre-existing ones listed above)
- `tests/unit/combo-routing-engine.test.ts` — 87/87 passed
- `tests/unit/providers-g4f-batch3.test.ts` — 10/10 passed, including the `chatanywhere`/`naga-ac` metadata assertions (confirms the dedup is behavior-preserving)
- `tests/unit/provider-tinycms-web.test.ts` — 15/15 passed, including the `initTinyCmsWasm` idempotency test (confirms fix #5 is behavior-preserving)
- Live-verified: rebuilt the combined staging branch for `omniroute-dev` with all fixes from this PR + 6 other rebased feature PRs, restarted the container, confirmed `/api/health/ping` returns 200 and stays healthy across multiple rechecks — twice: once after items #1-#6, and again after item #7 landed separately (a fresh contributor-PR merge wave broke it again a short time later)

### DIFF
--- a/src/shared/constants/providers/apikey/gateways.ts
+++ b/src/shared/constants/providers/apikey/gateways.ts
@@ -352,21 +352,6 @@ export const APIKEY_PROVIDERS_GATEWAYS = {
     apiHint:
       "Create a ZeroLimitAI Bearer token, then use https://www.zerolimitai.com/api/v1 as the OpenAI-compatible base URL.",
   },
-  chatanywhere: {
-    id: "chatanywhere",
-    alias: "chatanywhere",
-    name: "ChatAnywhere",
-    icon: "router",
-    color: "#2563EB",
-    textIcon: "CA",
-    passthroughModels: true,
-    website: "https://chatanywhere.tech",
-    hasFree: true,
-    freeNote:
-      "Personal, educational or research use only: public documentation cites 10,000 points/day and 200 requests/day per IP/key; do not use for commercial traffic.",
-    apiHint:
-      "Create a ChatAnywhere key linked to GitHub, then use https://api.chatanywhere.org/v1 outside China. Review the non-commercial terms before enabling it.",
-  },
   helyxai: {
     id: "helyxai",
     alias: "helyxai",


hartmark:
Superseded by `main`/`release/v3.8.50` -- rebasing this branch onto the latest `release/v3.8.50` (which now includes `f6ccd3cf9` "fix(quality): green release/v3.8.50 base-reds round 2 (#9985) (#10131)") drops this PR's entire remaining commit as "patch contents already upstream": `git diff upstream/release/v3.8.50...HEAD` is now empty, zero commits ahead.

`#10131`/`f6ccd3cf9` traces back to the same `#9985` base-red issue this PR's commit referenced ("base-red inherited: #9985"), so the fix landed via that other route.

Closing as superseded.

PR #10108, opened 2026-08-11
fix(build): unblock standalone production build (6 issues: catalog syntax, imports, wasm, migration)

HouMinXi:
**Problem**: the production `npm run build` (Next.js 16 standalone) fails at six independent points, so no deploy image can be produced from `release/v3.8.50`:

1. **Gateways catalog syntax** — `#9421` inserted `naga-ac`/`chatanywhere` right after `regolo`'s `apiHint` value without closing the `regolo` object first. Every following provider became a property of the still-open object and the file's closing brace no longer matched the opening `export const APIKEY_PROVIDERS_GATEWAYS = {`. `tsc` fails at EOF (`',' expected` at `gateways.ts:1213`) for any module importing the apikey provider catalog.

2. **Registry import paths** — `conol-web/index.ts` and `deepai/index.ts` import their shared modules with too few `../` levels (`../../../services/` instead of `../../../../services/`, `../shared` instead of `../../shared.ts`). Both failed `module not found` at build time; the sibling entries in the same directory use the corrected depths.

3. **Standalone build blockers** — `tinycmsSigner.ts` statically resolves `wasm_signer_bg.wasm` via `new URL(..., import.meta.url)`, which webpack treats as a bundle asset and fails when the file is absent; the WASM is already inlined as base64, so the URL is never exercised — a `/* webpackIgnore: true */` tells webpack to leave it alone. `modelSelectModalHelpers.ts` was missing the closing brace of `isProviderModelHidden()`, so the module would not parse.

4. **Duplicate import (webpack fatal)** — `videoGeneration.ts` imports `handleFalVideoGeneration` from both `videoGeneration/falHandler.ts` (line 21) and `mediaGeneration/fal.ts` (line 54). The two implementations have different signatures; the call site uses the destructured-args form that matches `falHandler.ts`. Webpack fails with `Identifier 'handleFalVideoGeneration' has already been declared`.

5. **Migration version collision (startup fatal)** — `143_job_registry.sql` collides with `143_api_key_cache_default_mode.sql`; the runner refuses to boot with `Migration version collision: version=143 → [api_key_cache_default_mode, job_registry]`. The file is an obsolete copy of `146_job_registry.sql` (its comment still reads "Migration 139", the SQL is identical), and deployed DBs already hold `143 = api_key_cache_default_mode`.

6. **Wrong import path (runtime crash)** — `conolDiscovery.ts` imports `getProviderOutboundGuard` from `@/shared/network/outboundUrlGuard`, but the function lives in `outboundUrlGuardPolicy`. Webpack compiles the missing export as `undefined`, so a request to the provider model discovery route throws at runtime.

**Fixes**:
- `gateways.ts`: add the missing `},` closing `regolo` (1 line).
- `conol-web/index.ts`, `deepai/index.ts`: correct the import depths to match the sibling entries (2 lines).
- `tinycmsSigner.ts`: `/* webpackIgnore: true */` on the wasm `new URL` (1 line).
- `modelSelectModalHelpers.ts`: restore the closing `}` of `isProviderModelHidden` (1 line).
- `videoGeneration.ts`: remove the duplicate `mediaGeneration/fal.ts` import (1 line).
- Delete `143_job_registry.sql` (48 lines); `146_job_registry.sql` carries the table definitions.
- `conolDiscovery.ts`: point the import at `outboundUrlGuardPolicy` (1 line).

**Verification**: a full standalone production build (`--target runner-base`, webpack path, 8 GB heap, 16 GB container) completes with the standalone integrity check passing for `@atjsh/llmlingua-2`, `@huggingface/transformers`, `@tensorflow/tfjs`, `js-tiktoken`, and `onnxruntime-node`. The resulting image boots healthy on the production server with migrations 143/146/147/150 applied cleanly, and `npm run lint` exits 0.


diegosouzapw:
Heads-up: the six build breakages this fixes were independently resolved by #10198 (build/lint/SWR repair) and #10131 (base-reds round 2), both merged into `release/v3.8.50` today — they cover the same eight files (regolo close-brace in gateways, conol-web/deepai registry imports, tinycmsSigner wasm URL, videoGeneration, conolDiscovery, modelSelectModalHelpers), and the job-registry migration was renumbered to `146_job_registry.sql` on the base, so the `143` rename here no longer applies. Please sync against the current tip and check whether anything in this branch still adds value — my read is it can close, but I'll leave that call to you rather than closing a PR of yours from under you. Full state-of-the-world reply on the rest of the plan is in https://github.com/diegosouzapw/OmniRoute/issues/8848#issuecomment-5272415657.

diegosouzapw:
Thanks for the build-unblocking work. Checking against the current `release/v3.8.50` tip: **all 6 fixes are already landed in the base** — 5 via #10198 (catalog syntax, import paths, wasm webpackIgnore, migration) and the videoGeneration duplicate-import fix was resolved inversely by #10198 (it deleted `videoGeneration/falHandler.ts` and kept `mediaGeneration/fal.ts` as the single source; the import the PR removes is the one the base still uses at `videoGeneration.ts:53`). Applying the PR's remaining delta would reintroduce an import to a file the base deleted.

Since there is no real delta left to ship, I'm closing this in favor of the already-merged fixes. If a specific standalone-build blocker remains, please open a fresh PR against the current `release/v3.8.50` and I'll review it right away.

diegosouzapw:
Closing as subsumed — all 6 fixes already landed in the base (see the comment above for the per-fix evidence).

PR #10116, opened 2026-08-11
fix(account-fallback): classify 'insufficient credits' as credits-exhausted

Chewji9875:
## Problem

Command Code (command-code) returns HTTP 400 with body:

> You have insufficient credits to make this request. Please purchase more credits to continue using the service.

when an account's billing credits run out. The phrase `insufficient credits` was **missing from `CREDITS_EXHAUSTED_SIGNALS`** (`open-sse/services/accountFallback.ts`), so:

1. `isCreditsExhausted()` → `false`
2. `classifyProviderError(400, ...)` → `null` (400 only maps context-overflow/model-unavailable)
3. `chatCore.ts` `if (errorConnectionId && errorType)` never fires → connection **never** marked `credits_exhausted`
4. `getProviderCredentials()` keeps re-selecting the same dead account (fill-first) on every request → every request 400s

## Fix

Add `"insufficient credits"` / `"insufficient credit"` to `CREDITS_EXHAUSTED_SIGNALS` (the phrase is already used by `antigravity429Engine.ts`). The error now classifies as `QUOTA_EXHAUSTED` → connection marked `credits_exhausted` → account selection skips it and rotates to a healthy account.

## Validation (TDD, Hard Rule #18)

- **Red**: added assertions `isCreditsExhausted("insufficient credits") === true` (and full Command Code body) to `tests/unit/account-fallback-service.test.ts` — failed with `false !== true` before the fix.
- **Green**: `node --import tsx/esm --test tests/unit/account-fallback-service.test.ts` → 77/77 pass.
- Regression: `antigravity-credits.test.ts` + `account-fallback-anthropic-quota.test.ts` → 26/26 pass.

## Note

This branch also carries the two base-red fixes from #9985 (gateway catalog syntax + conol-web import) taken byte-identical from PR #10102's head so the test runner can load at all; they are owned by #10102 and will drop out of this diff once that PR merges.


diegosouzapw:
Core fix is correct and tested — adding `'insufficient credits'`/`'insufficient credit'` to `CREDITS_EXHAUSTED_SIGNALS` in `accountFallback.ts` aligns with existing behavior (antigravity429Engine.ts). One check: does the matching cover common variations (capitalization, "out of credits")? Otherwise good to merge.

Chewji9875:
Thanks for the review!

Verified the variations:
- **Capitalization**: `isCreditsExhausted()` lowercases the entire error string (`errorText.toLowerCase()`) before substring matching against `CREDITS_EXHAUSTED_SIGNALS`, so all casing variations (e.g. `Insufficient Credits`, `INSUFFICIENT CREDITS`) match automatically.
- **"out of credits"**: Already in `CREDITS_EXHAUSTED_SIGNALS` alongside `"credits exhausted"`.
- **Singular/Plural**: Explicitly includes both `"insufficient credits"` and `"insufficient credit"` (as well as `"insufficient credit balance"`).

diegosouzapw:

PR #10121, opened 2026-08-11
fix(dashboard): expose OpenAI Responses store toggle for non-Codex connections

hartmark:
⚠️ base-red inherited: #9985 (resolved upstream since this PR opened; rebased onto the fix)

## Problem

`EditConnectionModal` only rendered and saved the "OpenAI Responses store"
toggle (`providerSpecificData.openaiStoreEnabled`) inside the Codex-only
settings block (`{isCodex && (...)}`).

That flag is read by `open-sse/utils/responsesStatePolicy.ts` —
`isOpenAIResponsesStoreEnabled()` / `applyResponsesPreviousResponseIdPolicy()`
— which is already fully provider-agnostic: in `auto` mode (the default) it
strips `previous_response_id` from an outgoing Responses-formatted request
*unless* the target connection has `openaiStoreEnabled: true`, regardless of
whether that connection's `provider` is `codex`, `openai`, or a generic
`openai-compatible-responses-*` proxy.

The component already computes a generic `isResponsesConnection` flag
(`provider === "codex" || provider === "openai" || (openai-compatible &&
targets the responses format)`) for exactly this class of setting — the
sibling `preserveEncryptedReasoning` toggle already uses it correctly. The
store toggle just never got the same treatment.

Net effect: an operator with a plain OpenAI API-key connection, or any
generic OpenAI-Responses-compatible proxy connection, had no way anywhere in
the dashboard to opt that connection into `store`/`previous_response_id`
continuation — the backend policy was ready, the UI control simply never
rendered for anything but Codex.

## Fix (UI toggle)

- Move the `Toggle` (and its save-time `providerSpecificData.openaiStoreEnabled`
  write) out of the `isCodex`-only block into its own JSX const gated on
  `isResponsesConnection`, mirroring `preserveEncryptedReasoningToggle`.
- Renamed the local `formData` field from `codexOpenaiStoreEnabled` to
  `openaiResponsesStoreEnabled` since it's no longer Codex-specific (the
  persisted `providerSpecificData.openaiStoreEnabled` DB/API field name is
  unchanged).
- No backend changes — `responsesStatePolicy.ts` and the `providerSpecificData`
  validation schema were already provider-agnostic.

## Update: the store functionality itself was also broken (real fix, not just UI)

Standing up a real deployment and driving `/v1/responses` against a genuine
`api.openai.com` connection with the toggle enabled surfaced a real,
independent functional bug behind the toggle: **the store field only actually
worked for the small whitelist of models with `targetFormat: "openai-responses"`**
(`gpt-5.6*`, `gpt-5.5-pro`, `gpt-5.4-pro`). For any other model on the same
connection (e.g. `gpt-5-nano`, `gpt-4o-mini`) — the common case — the request
gets silently routed to `/v1/chat/completions` instead, and `translateRequest`
leaks its internal `_omnirouteResponsesStore` marker verbatim into that
Chat-Completions-shaped body. OpenAI's real API rejects it outright:

```
{"error":{"message":"Unknown parameter: '_omnirouteResponsesStore'.","type":"invalid_request_error","code":"unknown_parameter"}}
```

Confirmed live end-to-end: with the fix, a real 2-turn session against
`api.openai.com` via a `gpt-5.4-pro` connection with the toggle enabled
correctly threads `previous_response_id` turn-to-turn (delta-only `input`,
real OpenAI response ids, second turn recalling state from the first). The
`gpt-5-nano` case (no responses-only `targetFormat`) previously 400'd with
the marker error; it now correctly falls through to Chat Completions with a
clean request body (no marker, no `store` — Chat Completions' own `store`
field means something unrelated, dashboard eval storage, so the client's
Responses-shaped store intent is dropped rather than silently remapped onto
it).

**Root cause & fix**: `translateRequest` (`open-sse/translator/index.ts`)
stashes a Responses-source request's `store` intent under the internal
`_omnirouteResponsesStore` marker so a later re-conversion back to Responses
shape can restore it as `store`. When the resolved destination stays in Chat
Completions shape, that re-conversion step never runs, so nothing ever
consumed the marker. Fixed by unconditionally dropping the marker at the end
of `translateRequest`, once translation is complete, regardless of
destination format.

**Also fixed**: a real crash discovered while live-testing store-enabled
requests under rate-limiting — `src/sse/handlers/chat.ts` referenced
`isProviderBreakerFailureStatus` without importing it, turning a clean
429/"no credits" response into an uncaught `ReferenceError` whenever all
provider accounts were rate-limited. This has since been independently fixed
upstream (`c9282248e`, already rebased onto here); this PR just removes the
resulting unused `PROVIDER_BREAKER_FAILURE_STATUSES` import left behind by
that fix.

## Test plan (TDD, Hard Rule #18)

`tests/unit/dashboard/edit-connection-modal-openai-store-toggle.test.tsx`
(UI toggle, unchanged from the original PR): renders `EditConnectionModal`
for a `provider: "codex"` connection (control, already passed) and for a
plain `provider: "openai"` connection, asserting the store toggle is present
and reflects a persisted `openaiStoreEnabled` flag. Confirmed RED on the
pre-fix code (toggle absent for `openai`), GREEN after the fix.

`tests/unit/responses-store-marker-leak.test.ts` (new, store functionality):
calls `translateRequest` directly for a Responses-source request with
`store: true` routed to a Chat-Completions destination and asserts neither
the internal marker nor a remapped `store` field leaks into the result.
Confirmed RED before the `translator/index.ts` fix (`store` present, `true`),
GREEN after.

```
node --test tests/unit/responses-store-marker-leak.test.ts
node --test tests/unit/dashboard/edit-connection-modal-openai-store-toggle.test.tsx
# both: pass

node --test tests/unit/openai-responses-only-models-5842.test.ts \
  tests/unit/combo-breaker-429.test.ts tests/unit/nvidia-quota-phase1.test.ts \
  tests/unit/combo/combo-target-timeout-standards.test.ts \
  tests/unit/default-reasoning-effort-6879.test.ts \
  tests/unit/translator-request-openai-responses.test.ts \
  tests/unit/translator-resp-openai-responses*.test.ts \
  tests/unit/openai-responses-reasoning-effort.test.ts \
  tests/unit/openai-responses-request-split.test.ts \
  tests/unit/request-defaults-store-session.test.ts
# 182 passed, 0 failed — no regressions in the surrounding translator/breaker/quota suites
```

## Notes

- `release/v3.8.50`'s base tip was red when this PR opened (#9985: migration
  collision + typecheck errors). Resolved upstream since (`b7b9fe0ba`); this
  branch has been rebased onto that fix, so the base-red marker above is
  historical.
- Two additional base-red bugs hit while isolating the store-functionality
  fix (a broken relative import in `conol-web/index.ts`, and a syntax error
  in `gateways.ts`) were also independently resolved upstream in the
  meantime; picked up for free by the same rebase, no longer part of this
  diff.

### DIFF
--- a/open-sse/translator/index.ts
+++ b/open-sse/translator/index.ts
@@ -34,7 +34,10 @@ import {
   recordReplay,
   requiresReasoningReplay,
 } from "../services/reasoningCache.ts";
-import { normalizeResponsesReasoningEffort } from "./request/openai-responses/helpers.ts";
+import {
+  normalizeResponsesReasoningEffort,
+  RESPONSES_STORE_MARKER,
+} from "./request/openai-responses/helpers.ts";
 
 bootstrapTranslatorRegistry();
 export { register } from "./registry.ts";
@@ -673,6 +676,19 @@ export function translateRequest(
         }
       }
     }
+  }
+
+  // #<store-marker-leak>: a Responses-source request stashes the client's
+  // `store` intent under this internal marker (see the Responses -> OpenAI
+  // step above) so a later OpenAI -> Responses re-conversion can restore it
+  // as `store`. When the destination stays in Chat Completions shape (no
+  // such re-conversion happens), nothing else consumes the marker, and it
+  // was leaking verbatim into the real upstream request body — e.g. OpenAI
+  // itself rejects it with "Unknown parameter: '_omnirouteResponsesStore'".
+  // Always drop it here: any handler that still needs the client's original
+  // `store` value would have already read the marker before this point.
+  if (RESPONSES_STORE_MARKER in result) {
+    delete result[RESPONSES_STORE_MARKER];
   }
 
   return result;
--- a/src/app/(dashboard)/dashboard/providers/[id]/components/modals/EditConnectionModal.tsx
+++ b/src/app/(dashboard)/dashboard/providers/[id]/components/modals/EditConnectionModal.tsx
@@ -123,7 +123,7 @@ export default function EditConnectionModal({
     accountId: "",
     codexReasoningEffort: "medium",
     codexServiceTier: "default" as CodexServiceTier,
-    codexOpenaiStoreEnabled: false,
+    openaiResponsesStoreEnabled: false,
     preserveEncryptedReasoning: false,
     consoleApiKey: "",
     newApiUserId: "",
@@ -330,7 +330,7 @@ export default function EditConnectionModal({
         accountId: existingAccountId,
         codexReasoningEffort: codexRequestDefaults.reasoningEffort,
         codexServiceTier: codexRequestDefaults.serviceTier ?? "default",
-        codexOpenaiStoreEnabled: connection.providerSpecificData?.openaiStoreEnabled === true,
+        openaiResponsesStoreEnabled: connection.providerSpecificData?.openaiStoreEnabled === true,
         preserveEncryptedReasoning:
           connection.providerSpecificData?.preserveEncryptedReasoning === true,
         consoleApiKey: existingConsoleApiKey,
@@ -634,8 +634,6 @@ export default function EditConnectionModal({
               ? { serviceTier: formData.codexServiceTier }
               : {}),
           };
-          updates.providerSpecificData.openaiStoreEnabled =
-            formData.codexOpenaiStoreEnabled === true;
         }
         if (isAntigravityFamily) {
           updates.providerSpecificData.projectId = trimmedCloudCodeProjectId || null;
@@ -662,6 +660,8 @@ export default function EditConnectionModal({
       if (isResponsesConnection && updates.providerSpecificData) {
         updates.providerSpecificData.preserveEncryptedReasoning =
           formData.preserveEncryptedReasoning === true;
+        updates.providerSpecificData.openaiStoreEnabled =
+          formData.openaiResponsesStoreEnabled === true;
       }
       const freeOnlyChanged =
         showFreeModelsToggle &&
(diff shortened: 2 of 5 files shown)


diegosouzapw:
Reviewed and validated in an isolated probe (DATA_DIR-hermetic). Confirmed on base `release/v3.8.50`: the store toggle was gated on `isCodex` while `isOpenAIResponsesStoreEnabled()`/`applyResponsesPreviousResponseIdPolicy()` already read `providerSpecificData.openaiStoreEnabled` for any connection — matches the described gap. Reverted `open-sse/translator/index.ts` and `EditConnectionModal.tsx` individually against `tests/unit/responses-store-marker-leak.test.ts` and `tests/unit/dashboard/edit-connection-modal-openai-store-toggle.test.tsx` — both fail pre-fix, pass post-fix (2/3 UI cases fail on base; the codex control case correctly still passes). The `chat.ts` import claim in the PR body describes a state already fixed upstream; this diff only removes the now-unused `PROVIDER_BREAKER_FAILURE_STATUSES` import, which is correct and in scope. i18n keys for the toggle already exist (it's a relocation, not a new string) so no i18nUiCoverage risk. No overlap with #10041/#10025 (different files/layers — request-translation marker cleanup vs response-side custom-tool classification vs transformer output_index). Looks merge-ready.


---

diegosouzapw:
Merged — thanks @hartmark! The store toggle is now available for all Responses-capable connections, and the _omnirouteResponsesStore marker leak into chat-completions upstreams is closed.

PR #10123, opened 2026-08-11
perf(logging): offload call-log artifacts to a worker

xz-dev:
## Summary

- move call-log artifact JSON serialization, checksum calculation, and atomic filesystem writes off the request thread into one bounded `worker_threads` worker
- preserve summary/detail metadata ordering and fail open with `detail_state=missing` when the worker or queue is unavailable
- drain complete call-log saves before SQLite shutdown and package the required worker in both raw Next standalone and npm dist runtimes

## Related Issues

- Related to sustained call-log latency and event-loop pressure under load.

## Validation

[Contribution Golden Path](../docs/dev/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite,

- [x] Change type: build-deploy / other
- [ ] `npm run lint` — exceeded the 6-minute local bound; changed-source/new-test ESLint passed with no diagnostics
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented below — pending CI

Commands/results:

- `node --import tsx/esm --test --test-concurrency=1 tests/unit/call-log-artifact-worker.test.ts tests/unit/call-log-save-drain.test.ts tests/unit/call-log-cap.test.ts tests/unit/pack-artifact-policy.test.ts` — 31 passed
- changed-source/new-test `npx eslint --no-warn-ignored ...` — passed
- `npm run check:build-scope` — passed
- `npm run check:pack-policy` — passed
- `npm run check:changelog-integrity` — this PR's `10123` fragment is valid; the gate still reports the two inherited invalid feature fragments `9239` and `9490`
- direct production-format `esbuild` bundle of `callLogArtifactWorker.ts` — passed
- `npm run build` probe emitted zero call-log worker or overly-broad file-pattern diagnostics after replacing ancestor probing/dynamic Worker construction; the local isolated-worktree build stops because its shared `node_modules` symlink is outside Turbopack's filesystem root
- `git diff --check` — passed

The active release base currently has unrelated Conol registry import/syntax failures. The DB-focused test command above was run with only that inherited base repair applied temporarily and restored afterward; this PR does not include those files.

## Tests Added Or Updated

- `tests/unit/call-log-artifact-worker.test.ts`
- `tests/unit/call-log-save-drain.test.ts`
- `tests/unit/call-log-cap.test.ts`
- `tests/unit/pack-artifact-policy.test.ts`

## Coverage Notes

- the real worker test covers exact npm-dist, raw-standalone, source/tsx, and missing-worker layouts plus artifact bytes/checksum metadata, the one-active/128-waiting bound, queue saturation fail-open, forced close, and warning rate limiting
- the drain test covers successful save completion, timeout-zero forced close, missing/null artifact metadata, and rejection of late saves during shutdown
- package policy tests pin the required worker path; CI and the final integration image provide the full standalone build signal

## Reviewer Notes

- no migration, dependency, public API, or feature flag is added
- migration/legacy artifact writers remain synchronous; only the request-path writer uses the worker queue
- queue/worker failures omit detail but preserve the SQLite summary row
- new worker warnings are generic and never include artifact paths, filenames, payloads, prompts, credentials, or user content
- independent review approved the candidate with residual risk limited to a pre-worker local DB lookup theoretically extending shutdown and synthetic worker event branches being inspected rather than individually fault-injected
- independent re-review approved the exact entrypoint-root resolver fix; final integrated build and artifact-write smoke remain deployment gates


diegosouzapw:
The offload design is sound: a bounded 128-job worker queue with fail-open, correct standalone resolution, and graceful-shutdown integration. One test needs a fix: `tests/unit/call-log-save-drain.test.ts:35` hard-codes `waitForCallLogSaves(2_000)`, but the first cold `worker_threads` spawn takes ~2.4s, so the drain test flakes on cold runs. Raising the timeout (or switching to conditional wait) fixes it — I verified both drain tests pass at 10s. The CI shard-4 failure was masked by the base-red `conolModels.ts` error, but this timing issue is real and worth fixing. Otherwise merge-ready. ⚠️ base-red inherited: #9985

PR #10124, opened 2026-08-11
fix(routing): fallback to default model alias seeds when unmapped in database

benzntech:
## Summary
Resolves an issue where bare OpenAI-compatible API model names (such as `claude-sonnet-4-6`, `claude-opus-4-6-thinking`, `gemini-3.6-flash-low`) failed with a `401: Missing API key` when unmapped in database aliases.

## Changes
1. Updated `resolveModelAlias()` in `src/lib/modelAliasResolver.ts` to fall back to `DEFAULT_MODEL_ALIAS_SEED` when unmapped in DB cache.
2. Added default alias seeds for `claude-sonnet-4-6`, `claude-opus-4-6-thinking`, and `gemini-3.6-flash-low` in `src/lib/modelAliasSeed.ts`.
3. Fixed syntax and import errors on release/v3.8.50 base.
4. Added unit tests in `tests/unit/model-alias-seed-fallback.test.ts`.

### DIFF
--- a/src/app/api/v1/chat/completions/route.ts
+++ b/src/app/api/v1/chat/completions/route.ts
@@ -26,7 +26,7 @@ import {
   readCompressionRequestHeader,
   withCompressionHeaderEcho,
 } from "@/shared/utils/compressionHeaderEcho";
-import { resolveModelAliasOnBody } from "@/lib/modelAliasResolver";
+import { resolveModelAliasWithSeedFallbackOnBody } from "@/lib/modelAliasResolver";
 
 let initPromise = null;
 
@@ -161,7 +161,7 @@ export async function POST(request) {
 
         // Resolve model alias before forwarding to handleChat
         if (parsedBody && typeof parsedBody === "object") {
-          await resolveModelAliasOnBody(parsedBody).catch(() => {
+          await resolveModelAliasWithSeedFallbackOnBody(parsedBody).catch(() => {
             /* swallow — fall through with original model */
           });
         }
--- a/src/app/api/v1/models/catalogHelpers.ts
+++ b/src/app/api/v1/models/catalogHelpers.ts
@@ -16,6 +16,7 @@ export interface CustomModelEntry {
   apiFormat?: string;
   supportedEndpoints?: string[];
   inputTokenLimit?: number;
+  outputTokenLimit?: number;
   isHidden?: boolean;
   // User-set "vision-capable" flag (persisted by addCustomModel / replaceCustomModels
   // in src/lib/db/models.ts). Surfaced into `/v1/models` via
--- a/src/lib/modelAliasResolver.ts
+++ b/src/lib/modelAliasResolver.ts
@@ -9,6 +9,7 @@
  * `src/lib/modelAliasSeed.ts`.
  */
 import { getModelAliases } from "@/lib/db/models/aliases";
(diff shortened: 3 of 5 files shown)


diegosouzapw:
Your branch's GitHub diff shows 1,778 files / +136,770 / −28,465, but the real change (vs the merge-base) is 3 files / +92 lines. The branch was pushed from a fork whose history was rewound — its ancestry includes #10098's head and an old `release/v3.8.50` commit, so GitHub's diff vs `main` pulls in all the divergence. Please rebase onto the current `release/v3.8.50` and `git push --force-with-lease` so the PR shows just the ~92-line change. Two review notes: (1) `resolveModelAlias` collides with an existing export in `open-sse/services/modelDeprecation.ts` — please pick a distinct name; (2) the test only asserts seed fallback — could you add a case that reproduces the 401 you're fixing? Happy to merge once those land.

benzntech:
All three review points addressed — thanks for the detailed feedback.

**1. Rebased onto current `release/v3.8.50` and force-pushed.** The PR diff is now the real change: 4 files / +80 / −15 (was 1,778 files from the rewound fork ancestry). `git push --force-with-lease` done.

**2. Renamed the colliding export.** `resolveModelAlias` → `resolveModelAliasWithSeedFallback` (and `resolveModelAliasOnBody` → `resolveModelAliasWithSeedFallbackOnBody`) in `src/lib/modelAliasResolver.ts`; the chat-completions route import updated. The sync `resolveModelAlias` in `modelDeprecation.ts` / `modelSpecs.ts` is untouched. A test asserts the colliding name is NOT exported.

**3. Added the 401 regression case.** The test now exercises the exact failure: a model alias absent from the (DATA_DIR-isolated, empty) `modelAliases` namespace but present in `DEFAULT_MODEL_ALIAS_SEED` resolves to the seed target instead of passing through unmapped (which is what caused the upstream 401). The test isolates DATA_DIR with a temp dir + `resetDbInstance()` so it never reads the operator's live DB — it's hermetic and deterministic.

Local gates: model-alias + rotate tests 10/10, route-validation ✓.

benzntech:
Force-pushed the rebased branch (now 4 files / +80 / −15 vs the 1,778-file rewound diff) with all review points addressed. CI status on the new head:

**Fixed by this push**: Build diff (now minimal), Docs Gates ✓, Merge integrity ✓, No new ESLint warnings ✓, Unit Tests fast-path 1/2/4 ✓, Vitest ✓, semgrep ✓, dast-smoke ✓, Change Classification ✓.

**Still failing — appears environmental, not code**: `Fast Quality Gates` (dead-code + open-sse-typecheck) and Build (advisory). Both gates pass locally on this exact branch (`npm run typecheck:core` EXIT 0, check-dead-code OK at 224 symbols < 227 baseline, migration-numbering OK). The identical gates fail on #10126 (quota PR — completely different code), suggesting a runner-environment drift (knip/tsc versions) rather than this diff. Happy to dig further if the maintainers confirm it's not infra.

diegosouzapw:
Thank you @benzntech for the alias-seed fallback fix. The exact head passed the complete PR checks and the combined FAST train; the train’s only incompatible red came from the pre-existing optional-transformers contract after #10543’s intentional onnxruntime pin, not from this PR.

PR #10128, opened 2026-08-11
fix(combo): default chaos SSE to comment-only for OpenAI-compatible clients

herjarsa:
## Problem

`auto/chaos` and `auto/best-chaos` combos dispatch N models in parallel via `handleChaosChat`, which emits a custom SSE event `event: omni-chaos-part` with a `data:` payload carrying `type: "omni-chaos-part"` metadata for every panel model.

OpenAI SDK validators and `@ai-sdk/openai-compatible` expect every SSE `data:` line to match either `{choices:[...]}` (a chunk) or `{error:...}`. The custom `omni-chaos-part` event's `data:` line triggers a schema validation `invalid_union` error **on every auto/chaos stream**, even for clients that never asked for the multi-panel protocol.

## Fix

Make the custom event **opt-in** via `stream_options.include_chaos_parts`:

| Client behavior | Before this fix | After this fix |
|---|---|---|
| **Default** (no opt-in) | `event: omni-chaos-part` + `data: {type:"omni-chaos-part",...}` emitted per model — **breaks OpenAI SDK** | Only SSE comment `: chaos N ok/fail model` emitted per model (comments are ignored by all SSE parsers per spec) |
| **Opt-in** (`stream_options.include_chaos_parts: true`) | Same as before (always-on) | Full custom event emitted for protocol-aware IDEs that split panels |

## Changes

- **`chaosEngine.ts`**: `serializeChaosPart(part, isFinal, emitCustomEvent = false)` — third param controls custom event emission. `handleChaosChat` reads `body.stream_options?.include_chaos_parts` and passes the flag through.
- **`chaosEngine.test.ts`**: 3 new test cases covering comment-only default, opt-in custom event, and error-all-fail path.

## Testing

- 14/14 tests pass (`chaosEngine.test.ts` + `chaosVirtualCombo.test.ts`)
- Type-check clean on touched files (no new errors)

## Impact

- **Backward-compatible**: default behavior changes from "always emit custom event" to "comment-only". No existing client that doesn't use `include_chaos_parts` will see custom events (they were already ignored by those clients).
- **No breaking change to protocol-aware clients**: they must now opt in via `stream_options.include_chaos_parts: true`.


diegosouzapw:
Merged into `release/v3.8.50` — thank you @herjarsa for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

PR #10063, opened 2026-08-11
fix(db): remove stale 143_job_registry.sql and renumber api_key_cache_default_mode to 144

excessivechaos:
## Summary

The 139→143 job_registry renumber was adopted upstream, then renumbered again to 146 (`754ba0fa8`) — but the stale `143_job_registry.sql` was left in the tree. It now collides with `143_api_key_cache_default_mode.sql` (#8984), so the migration runner throws `Migration version collision detected: version=143` on every fresh DB init and blocks DB-backed tests repo-wide.

- **Delete** the stale `143_job_registry.sql` — the canonical slot is `146_job_registry.sql`.
- **Renumber** `143_api_key_cache_default_mode.sql → 144_api_key_cache_default_mode.sql` so databases that already applied job registry under the old 143 slot (e.g. deployed instances) still receive the `api_keys.cache_default_mode` column instead of being silently skipped by version-only tracking.
- The 139→146 reconciliation already exists upstream in `RENAMED_MIGRATION_COMPATIBILITY`; with no 143 file remaining, no name-mismatch warning fires for those databases.

## Verification

- Fresh DB init passes the collision check: `db-migration-runner`, `db-migration-version-uniqueness`, `db-migrationrunner-constants-split`, `db/jobRegistryDb` — 46/47 green (the single failure is `db-fresh-setup-9934`, a pre-existing base-red: the pending-count bound from #10022 is exceeded by the migration count on the current release, independent of this change — fails on a pure release checkout too).

## Base-red context

- The #9934 fresh-setup guard fails on pure release (pre-existing, unrelated).
- Docs-sync version drift (package.json 3.8.49 vs CHANGELOG 3.8.50).

Issue #7694, opened 2026-07-18
feat(providers): capture upstream reasoning.supported_efforts at sync so synced openai-compatible models become selectable with effort levels


excessivechaos:
### Summary
OpenAI-compatible providers can advertise per-model reasoning effort levels in their `/v1/models` response via a `reasoning: { supported_efforts: [...] }` object. OmniRoute currently syncs `supportsThinking`/`supportsVision` from that response but **ignores `supported_efforts` entirely**. As a result, a synced reasoning model (e.g. `openference/GLM-5.2`) is selectable, but there is **no way to pick the effort level** (low/medium/high/max) — and there is no model id like `openference/GLM-5.2-high` generated. Users are forced to hand-edit the request body or rely on provider-specific hardcoded alias tables (see #6987, which only covers `opencode-go` + a few hardcoded base models).

### Evidence — upstream payload (openference, openai-compatible)
```json
{ "id": "GLM-5.2", "reasoning": { "supported_efforts": ["high","medium","low"] } }
{ "id": "DeepSeek-V4-Pro", "reasoning": { "supported_efforts": ["max","high","medium","low"] } }
{ "id": "MiniMax M3", "context_length": 850000 }
{ "id": "SenseNova U1 Fast", "architecture": { "modality": "text->image" } }
```
Note: models with no `reasoning` object (MiniMax M3, image model) correctly have **no** effort tiers. The absence of the key must be treated as "no effort support," not a default set.

### Where the gap lives (verified against release/v3.8.49)
1. **Sync mapper ignores it.** `src/lib/providerModels/modelDiscovery.ts` → `normalizeDiscoveredModels` reads `record.supportsThinking === true` (line ~109) but never reads `record.reasoning?.supported_efforts`. This is the capture point.
2. **No storage field.** `SyncedAvailableModel` (`src/lib/db/models.ts`) has `supportsThinking?` / `supportsVision?` but no `supportedEfforts?` (or `effortTiers?`) field. Add one and persist it via `replaceSyncedAvailableModelsForConnection`.
[...]

PR #10138, opened 2026-08-11
fix(discovery): parse reasoning tiers nested under metadata.reasoning.supported_efforts (neuralwatt shape)

excessivechaos:
## Summary

neuralwatt's `/v1/models` wraps capabilities and reasoning under a `metadata` object — `metadata.reasoning.supported_efforts` (e.g. `["max","high","none"]`) and `metadata.capabilities.reasoning_effort` — one level deeper than the shapes `detectSupportedThinkingEfforts` recognized (top-level `reasoning.supported_efforts`, `capabilities.effort_tiers`, `supported_reasoning_levels`, `thinking.levels`). Synced openai-compatible rows therefore carried no `supportedThinkingEfforts`, so the catalog/Combo Builder advertised no effort aliases for neuralwatt models.

- Recognize `metadata.reasoning.supported_efforts` with the same Zod schema and validation as the top-level #7694 shape (Hard Rule #7: a malformed entry degrades, never throws).
- Placed right after the top-level `reasoning.supported_efforts` in precedence, so a top-level declaration still wins when both are present.
- `max` normalizes to `xhigh` through the shared discovery vocabulary, matching every other tier-array source (neuralwatt accepts `xhigh` and aliases it to `max` upstream).

## Tests

- `model-discovery-reasoning-levels.test.ts` — 3 new: neuralwatt shape parsed; top-level declaration not overridden; malformed `metadata.reasoning` degrades without throwing. Suite 16/16 pass.

## Related

- **#10147** — catalog-side publish gap: `/v1/models` still shows no `-<tier>` variants for openai-compatible node providers because the synced loop drops all node models (providerSupportsModel connection resolution). This PR fixes the discovery-side capture; #10147 tracks the catalog-side fix.

## Base-red (pre-existing, not from this PR)

- `conol-web` registry import path error (#10102, open) and `gateways.ts` syntax error (#10093, fixed by #10102) block registry import; locally worked around for the test run.
- Docs-sync version drift (package.json 3.8.49 vs CHANGELOG 3.8.50).
- Migration version collision 143 (fresh-DB/DB-backed tests) — fixed by #10063.

### DIFF
--- a/changelog.d/fixes/discovery-metadata-effort-tiers.md
+++ b/changelog.d/fixes/discovery-metadata-effort-tiers.md
@@ -1,0 +1,1 @@
+- fix(discovery): parse upstream reasoning tiers nested under metadata.reasoning.supported_efforts (neuralwatt /v1/models shape) so synced openai-compatible models advertise effort aliases
--- a/src/lib/providerModels/modelDiscovery.ts
+++ b/src/lib/providerModels/modelDiscovery.ts
@@ -145,6 +145,27 @@ export function detectSupportedThinkingEfforts(record: JsonRecord): string[] | u
     }
   }
 
+  // neuralwatt-style upstreams wrap the same tier data one level deeper under
+  // `metadata.reasoning.supported_efforts` (their /v1/models nests capabilities
+  // and reasoning under a `metadata` object). Same semantics and validation as
+  // the top-level #7694 shape; placed right after it so a top-level declaration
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Merged into `release/v3.8.50` — thank you @excessivechaos for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

Issue #10143, opened 2026-08-12
fix(oauth): send required CLI headers in claude-auth import bootstrap call


stanleytejakusuma:
## Summary

Claude (OAuth/subscription) connections created via `claude-auth/import` intermittently receive `HTTP 400` from Anthropic: `"Third-party apps now draw from your extra usage, not your plan limits."` — even though the imported OAuth token is valid, belongs to a real Claude Max/Pro account, and works correctly when the exact same account is connected through the normal (non-import) OAuth setup flow.

## Root cause

`src/lib/oauth/utils/claudeAuthImport.ts` has its own bootstrap-identity fetcher (`enrichWithBootstrap`) that is **missing two headers** the two other call sites of the same bootstrap endpoint always send:

```ts
// claudeIdentity.ts:154 and src/lib/oauth/providers/claude.ts:12 (both correct)
fetch("https://api.anthropic.com/api/claude_cli/bootstrap", {
  headers: {
    Authorization: `Bearer ${accessToken}`,
    "anthropic-version": "2023-06-01",
    "User-Agent": "claude-cli/1.0.0 (external, cli)",
    "anthropic-beta": "oauth-2025-04-20",   // <-- present here
  },
})

// claudeAuthImport.ts:116-124 (enrichWithBootstrap) — missing both
fetch("https://api.anthropic.com/api/claude_cli/bootstrap", {
  headers: {
    Authorization: `Bearer ${accessToken}`,
    "anthropic-version": "2023-06-01",
    "Content-Type": "application/json",
    // no User-Agent, no anthropic-beta
  },
})
```

Without those two headers, Anthropic's bootstrap endpoint doesn't recognize the caller as a CLI client and the request fails silently (non-2xx). `enrichWithBootstrap` swallows that failure and returns an all-null identity object (`claudeAuthImport.ts:126-127`):

```ts
if (!res.ok) return base; // base = { accountUUID: null, organizationUUID: null, ... }
```

`createConnectionFromAuthFile` (`claudeAuthImport.ts:174-259`) is supposed to refuse to create a connection with unverified identity — but that refusal is bypassed entirely when the caller passes `overwriteExisting: true` (`:224`), which is the *only* way to successfully import in practice, since a first import attempt without it fails with `identity_unverified` precisely because of this same header bug. So every import silently proceeds with `providerSpecificData: { accountUUID: null, organizationUUID: null, ... }`.

[...]

PR #10144, opened 2026-08-12
fix(oauth): send required CLI headers in claude-auth import bootstrap call

stanleytejakusuma:
Two-line fix: `enrichWithBootstrap()` in `claudeAuthImport.ts` was missing the `User-Agent` and `anthropic-beta` headers that the two other callers of the same `/api/claude_cli/bootstrap` endpoint (`claudeIdentity.ts`, `src/lib/oauth/providers/claude.ts`) already send correctly. Full root-cause writeup and repro in #10143.

Net effect once merged: `claude-auth/import` connections get a real, Anthropic-verified `accountUUID`/`organizationUUID` from the start (same as the native OAuth flow), instead of silently falling back to a hash-derived fake UUID that Anthropic classifies as third-party traffic and intermittently rejects with the "draw from your extra usage" 400.

Tested against a real `claude_max` account: before the fix, bootstrap silently fails and import requires `overwriteExisting: true`; conceptually verified the header set now matches the two working call sites exactly (byte-for-byte header keys/values).

### Validation

- **Unit (Rule #18):** `tests/unit/claudeAuthImport-bootstrap-headers-10144.test.ts` proves (1) `enrichWithBootstrap()` sends the required CLI headers on the bootstrap call and still falls back to null identity on non-OK upstream, and (2) `createConnectionFromAuthFile()` mints a 64-hex `cliUserID` on create, preserves it on overwrite re-import, and mints a fresh one when absent.
- **Live upstream (VPS-gated):** the Anthropic-side acceptance of these headers / real `accountUUID` return is verified live during release drain against a real subscription account (repro + pre-fix behavior documented in #10143).


diegosouzapw:
Small, correct fix — the CLI headers and `cliUserID` persistence match the two working bootstrap call sites. Two follow-ups before merge: (1) a unit test for the header-sending path would be good (Rule #18); (2) there's a hardcoded-version nit. The upstream OAuth behavior itself is live-gated — I'll validate it on the VPS during the release drain. Otherwise looks good.

stanleytejakusuma:
Both follow-ups addressed:

**1. Unit test for the header-sending path (Rule #18)** — already in place: `tests/unit/claudeAuthImport-bootstrap-headers-10144.test.ts` (f31fdac5) proves (a) `enrichWithBootstrap()` sends `User-Agent: claude-cli/*` + `anthropic-beta: oauth-2025-04-20` on the bootstrap call and still falls back to null identity on non-OK upstream, and (b) `createConnectionFromAuthFile()` mints a 64-hex `cliUserID` on create, preserves it on overwrite re-import, and mints fresh when absent.

**2. Hardcoded-version nit** — tightened in d78f9d47: the test previously re-typed the pinned `2.1.220` as a local constant. Both the executor and the test now import `getClaudeCodeUserAgent()` from `claudeCodeClient.ts` (the single source of truth the two working call-sites use), and the test asserts the outbound UA byte-for-byte against it — so a future version bump can't silently desync the wire identity.

Local verification: new test 5/5 pass, sibling `claudeAuthImport.test.ts` pass, eslint on both changed files has no new findings (only the pre-existing `@/lib/localDb` barrel-import restriction on an untouched import line).

On the red CI jobs — none trace to this PR's files: `Merge integrity` failed on a stale merge-ref (`skills/omni-settings/SKILL.md` predating #10169); GitHub has since recomputed the merge ref and the check passes locally against the current base. `Fast Quality Gates`/unit jobs fail on base-state issues (`open-sse/utils/stream.ts` TS2300, stryker `tap.testFiles` drift, poolside live-catalog + combo retry tests) that reproduce on other open PRs against the same base (#10340, #10363). The new test file itself passed in CI (job 2/4). Happy to rebase if you'd like a re-run.

diegosouzapw:

#10144 foi embarcada no train combinado `8ed587f50410665b2924b372e59f518e420a02bd`, validado em `root@192.168.0.113` (`tomni-proxmox-113`). O log é `/srv/omniroute-train/.claude/worktrees/merge-train-20260817-085938-suite.log`.

Modo FAST verde: typecheck, file-size, complexidade ciclomática e cognitiva, changelog integrity, teste alterado `claudeAuthImport-bootstrap-headers-10144.test.ts` e Vitest. A suíte full da release continua sendo executada separadamente no tip acumulado.

Obrigado @stanleytejakusuma pela contribuição; o PR segue para merge serializado em `release/v3.8.50` sob a autorização da campanha.

diegosouzapw:
Thank you @stanleytejakusuma for the Claude auth import fix. The contributor implementation is preserved; the added regression coverage exercises the real implementation and proves the required headers and cliUserID persistence. The only remaining failed shards were reproduced Alibaba base-reds on release/v3.8.50. Local merge-train evidence: tomni-proxmox-113, FAST tip bf128e64ee9b9949df2ff2b1e7451db4d07580dd.

Issue #10110, opened 2026-08-11
fix(admission): preserve process-wide bounds across byte-level connection lanes


imrightguy:
### Problem

The always-on byte/structure `PerConnectionAdmissionController` introduced by #9940 gives each API-key lane an independent `ChatAdmissionController`. This breaks the process-wide memory-safety invariant the original chat-body guard was designed to enforce.

Two separate problems are present on current `release/v3.8.50`:

1. `CHAT_MAX_HEAVY_IN_FLIGHT` and `CHAT_ADMISSION_MAX_QUEUED_BYTES` are enforced per lane, not process-wide. With the default 64 byte-level sessions, the default 4 MiB queued-body budget permits up to 256 MiB of parked bodies in aggregate; a deployment configured for 16 MiB permits up to 1 GiB. Active heavyweight capacity multiplies the same way.
2. LRU and TTL eviction delete byte-level lane records without checking `activeHeavy`, queued bytes, or parked waiters. An active controller becomes orphaned; the same API key can receive a fresh controller and acquire fresh capacity while the old lease remains live. The snapshot reports only the replacement controller.

This is distinct from the opt-in adaptive virtual lanes. The adaptive controller has `deleteLane()` to drain queued entries, but the always-on byte-level controller simply calls `Map.delete()`.

### Current implementation evidence

`src/shared/middleware/chatBodyAdmission.ts`:

- `PerConnectionAdmissionController.getController()` creates `new ChatAdmissionController(this.maxHeavyPerSession)` for every session.
- On capacity, it executes `this.#sessions.delete(oldestKey)`.
- TTL eviction executes `this.#sessions.delete(sessionId)` based only on `lastUsedMs`.
- `ChatAdmissionController.#queuedBytes` is private to each controller.
- `snapshot()` does not include queued bytes and cannot see orphaned active controllers.

The merged PR rationale called aggregate memory amplification a false positive because each lane's queue is bounded. A per-lane bound multiplied by the lane count is not a process-wide bound.

### Minimal reproduction

```ts
const lanes = new PerConnectionAdmissionController(1, {
  maxSessions: 1,
  sessionTtlMs: 60_000,
});
const old = lanes.getController("api-key-A");
const oldLease = old.tryAcquireHeavy();

lanes.getController("api-key-B"); // evicts active A
const replacement = lanes.getController("api-key-A");
const replacementLease = replacement.tryAcquireHeavy();

console.log({
  configuredMax: 1,
[...]

PR #9940, opened 2026-08-09
fix: per-connection virtual admission lanes (#9654)

branben:
## Summary

Per-connection virtual admission lanes for agentic workloads (#9654). Built with worst-day-ever analysis (4 bugs found + fixed, 2 false positives debunked) and Matt Pocock methodology (failing-test-first, exact assertions).

### Original Discussion (#9608 / #9654)

**Problem:** The global admission controller returns 503/502 before OmniRoute can fan requests across provider targets, defeating multi-provider routing for agentic frameworks (Cursor, Claude Code, CrewAI, AutoGen) that send large concurrent bursts.

**Acceptance Criteria:**

| Criterion | Status | Implementation |
|-----------|--------|----------------|
| 1. Agentic burst does not 503 other sessions | ✅ | Per-tenant FairCostQueue lanes with isolated capacity |
| 2. Combo targets exhausted before lane queueing | ⚠️ | Lane enqueue opt-in via env var; combo dispatch downstream in chat.ts/fusion.ts. Integration plan in section below. |
| 3. Idle lanes auto-evict after TTL | ✅ | 60s TTL + evictIdleLanes() + LRU cap at 1000 |
| 4. 503 with structured Retry-After | ✅ | Retry-After header on capacity-exhaustion |
| 5. Configurable via env vars | ✅ | OMNIROUTE_CHAT_VIRTUAL_LANES=1 |
| 6. Metrics in health endpoint | ✅ | Lane metrics via snapshot() |

### Architecture

route.ts -> resolveSessionId() -> admitChatRequest -> admitChatStructure
  -> PerConnectionAdmissionController.acquire()
    -> AdaptiveAdmissionController.acquire()
      -> getOrCreateLane(tenantKey) -> FairCostQueue per session
      -> evictIdleLanes() every 30s (60s TTL, 1000 max)
      -> deleteLane() drains + rejects on LRU eviction

### Bugs Fixed (Worst-Day-Ever Analysis)

| Severity | Bug | Fix |
|----------|-----|-----|
| CRITICAL | expireEntry() null dereference | Added if (!entry) return guard |
| HIGH | LRU eviction orphanates pending promises | New deleteLane() method: drains + rejects all entries |
| MEDIUM | Map.delete() during evictIdleLanes iteration | Collect keys first, then deleteLane in separate pass |
| MEDIUM | Misleading ADMISSION_ABORTED for eviction | Added ADMISSION_LANE_EVICTED to AdmissionRejectCode union |

### False Positives Debunked

[...]

PR #9816, opened 2026-08-08
fix(admission): queue heavyweight chat requests before 503 busy

herjarsa:
## Problem

Agent-style clients (OpenCode, Claude Code, Cursor) fan out heavy sub-requests that land on the chat admission gate together. With the single heavyweight slot (OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT=1), the second concurrent heavy request was rejected **immediately** with a retryable 503 chat_admission_busy.

Those clients retry on a ~1s cadence; a burst of 2-3 heavy sub-requests burns the whole retry budget in seconds and the agent **dies mid-task** with no user-visible error — the UI just goes blank/waiting while later requests log 200.

## Fix

Heavy chat requests now **wait for a slot** instead of failing instantly:

- New ChatAdmissionController.acquireHeavyWithin(timeoutMs) — bounded FIFO wait, re-acquiring atomically on each release; resolves 
ull on deadline.
- dmitChatRequest (byte-based path) and dmitChatStructure (structure-based path, now async) use it with a configurable budget.
- New env OMNIROUTE_CHAT_ADMISSION_QUEUE_MS (default 5000);

Issue #9606, opened 2026-08-06
fix(backend): Opencode unable to run 2 or more sessions simultaneously


Tushar49:
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11, Version 10.0.22631 Build 22631

### Node.js Version

22.23.1

### Provider(s) Involved

All (e.g. Opencode)

### Model(s) Involved

Any model (e.g. Deepseek v4 Flash Free)

### Client Tool

Opencode

### Description

Bug 1: For example, I am using opencode/deepseek-v4-flash-free or oc/deepseek-v4-flash-free in one of the terminal with opencode session, and while its running I open another session with opencode, and with model as oc/deepseek-v4-flash-free or any else try to chat. Now either this chat will cease to run, or it will run and cause other chat to cease unexpectedly. I am not sure why this is happening.

Bug 2: When using models like oc/deepseek-v4-flash-free, many times I see:
" (prior reasoning summary unavailable)
(prior reasoning summary unavailable)"

<img width="527" height="146" alt="Image" src="https://github.com/user-attachments/assets/92c58faa-8f80-4008-a950-e15820d869bf" />

This only happens with omniroute models. Please help fix this.

### Steps to Reproduce

Already provided

### Expected Behavior

I expected both chats to run simulatenously, or even more than 2 chats should cause issues like this. It didnt cause issues months ago when I had started using OmniRoute.

### Actual Behavior

Only 1 chat is able to use the omniroute model, rest all stop running with errors.
Errors like "

<img width="527" height="146" alt="Image" src="https://github.com/user-attachments/assets/33a3f264-0d6a-4b83-a992-22f9c09ae1e5" />

"
Keep coming.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

### Screenshots

<img width="527" height="146" alt="Image" src="https://github.com/user-attachments/assets/d0ff1131-edaf-469c-8c9e-1a0b987dabdd" />

### Additional Context

Opencode shows error:
'Structurally heavy chat request capacity is busy; retry shortly. [retrying in 1s attempt #6]esc again to interrupt' when 2 or more sessions trying to access models.

Tushar49:
So, during simultaneous usgae, I was able to extract following error message from Opencode:
'Structurally heavy chat request capacity is busy; retry shortly. [retrying in 1s attempt #6]esc again to interrupt'

[...]

PR #10149, opened 2026-08-12
fix(auto-combo): scope concurrency to provider account

megamen32:
## Summary
Move heavyweight chat admission out of the pre-route global lease. The request reaches normal provider/account routing first, so existing session affinity keeps one session on its pinned account; independent sessions can select independent accounts.

The account semaphore key is `provider:account`, so each selected account remains serialized by its own configured cap.

## Minimal diff
- `src/shared/middleware/chatBodyAdmission.ts` — removes the process-wide pre-parse capacity lease.
- `src/app/api/v1/chat/completions/route.ts` — removes lease lifecycle handling.
- `tests/unit/account-semaphore-account-scope.test.ts` — proves separate accounts/providers do not contend while one account serializes.

## Validation
- account semaphore regression
- generic session-affinity regression
- Codex session-affinity vs forced target regression
- Result: 24 passed, 0 failed.
- ESLint passed for the three changed files.

No combo-config override, new parser, migration, dashboard/API surface, task card, or roadmap change is included.

### DIFF
--- a/src/app/api/v1/chat/completions/route.ts
+++ b/src/app/api/v1/chat/completions/route.ts
@@ -1,0 +1,209 @@
+import { z } from "zod";
+import { CORS_HEADERS, handleCorsOptions } from "@/shared/utils/cors";
+import { callCloudWithMachineId } from "@/shared/utils/cloud";
+import { handleChat } from "@/sse/handlers/chat";
+import { generateRequestId } from "@/shared/utils/requestId";
+import { errorResponse } from "@omniroute/open-sse/utils/error.ts";
+import { initTranslators } from "@omniroute/open-sse/translator/index.ts";
+import { createInjectionGuard } from "@/middleware/promptInjectionGuard";
+import { acceptHeaderForcesStream } from "@omniroute/open-sse/utils/aiSdkCompat.ts";
+import {
+  OPENAI_CHAT_ERROR_FRAME,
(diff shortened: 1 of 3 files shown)


Tushar49:
@diegosouzapw @megamen32 — This PR appears to address the user-facing symptom reported in issue #9606.

**Problem:** v3.8.49+ moved to a binary process-wide heavyweight lease (CHAT_MAX_HEAVY_IN_FLIGHT=1) held for the full SSE lifetime. This serializes concurrent agent sessions — OpenCode, Claude Code, Cursor — even when they could route to different provider accounts.

**Current mitigations are insufficient:**
- OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT=2 — still binary, does not scale, and increases OOM risk.
- Per-connection lanes (#9940) — may multiply the memory budget without a process-wide bound (#10110).
- Queue wait (#9816) — helps serialize bursts, but the single-slot bottleneck remains.

**PR #10149**, raised by @megamen32, looks like the better architectural direction: move admission after routing so each provider account gets its own concurrency scope. This aligns with the hierarchical concurrency model proposed in #7778 by @RaviTharuma and would let independent sessions use independent accounts without contending for one process-wide slot.

Could the OmniRoute maintainers please prioritize review of this PR, while preserving an aggregate process-wide memory/concurrency safety bound? That would directly unblock multi-session agent workflows without simply raising the global lease limit.

PR #10175, opened 2026-08-12
[TS7] fix(types): tighten chatCore helper contracts

backryun:
## Summary

- preserve the distinction between string-valued tool aliases and namespace identity records
- return the extracted post-call guardrail context through the public `GuardrailContext` contract
- add focused regression coverage for both helper boundaries

## Why

Two remaining TypeScript 7 diagnostics came from inferred helper shapes in `chatCore`: an identity map was assigned to a string alias map after a runtime-only value check, and the post-call guardrail builder returned untyped request-boundary values. This change performs the existing alias validation through a typed helper and normalizes the guardrail fields at the extracted boundary.

This deliberately does not duplicate the open base-red repair in #10131.

Part of #8484.

## Validation

- `node --import tsx/esm --test tests/unit/chatcore-request-tool-identity-contracts.test.ts tests/unit/chatcore-postcall-guardrail-context.test.ts tests/unit/9780-namespace-identity-pivot.test.ts tests/unit/9568-gemini-tool-casing-mismatch.test.ts` — 22/22 passed
- targeted ESLint on all five changed files — passed
- TypeScript 6 full `open-sse/tsconfig.json`: 50 → 48, removed 2, added 0
- TypeScript 7.0.2 full `open-sse/tsconfig.json`: 50 → 48, removed 2, added 0
- `npm run check:file-size` — changed files pass; command still reports six existing release-base violations outside this diff

### DIFF
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -1,4 +1,7 @@
-import { extractRequestToolIdentityMap } from "./chatCore/requestToolIdentity.ts";
+import {
+  extractRequestToolIdentityMap,
+  toToolNameAliasMap,
+} from "./chatCore/requestToolIdentity.ts";
 import { injectMemoryAndSkills } from "./chatCore/memorySkillsInjection.ts";
 import { resolveChatCoreRequestSetup } from "./chatCore/requestSetup.ts";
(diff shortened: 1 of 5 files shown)


diegosouzapw:
Merged into `release/v3.8.50` — thank you @backryun for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

PR #9199, opened 2026-08-02
[v3.8.50] fix(models): keep model catalogs responsive

xz-dev:
## Summary

- Use the existing narrow `touchConnectionLastUsed()` path for session-affinity `lastUsedAt` / `consecutiveUseCount` bookkeeping instead of the generic catalog-invalidating provider update.
- Prepare one build-local virtual-auto candidate snapshot for all 38 built-in catalog entries instead of rebuilding the same connection/settings/registry candidate base 38 times.
- Keep regular and family no-auth pools distinct, clone nested connection allowlists per materialization, and preserve request-time API-key/channel exclusions.
- Yield cooperatively after preparation and every eight built-in entries so unrelated requests can run during a cold catalog build.
- Detach invalidated in-flight generations and guard publication by catalog version plus key ownership, so a settings/connection mutation during a yield cannot serve or cache the obsolete generation.

The prepared snapshot is local to one catalog build. This PR adds no global/stale candidate cache and does not change authentication, cache-key partitioning, TTLs, provider routing, or response shape.

## Related Issues

- Related to #8697
- Focused follow-up to #8728; this PR does not import, duplicate, replace, or supersede its last-good/SWR work

## Production Diagnosis

A bounded diagnostic image on the downstream production shape identified the cold-builder hotspot before implementation:

```text
Authenticated Primary catalog: 41.064 s
Unrelated dashboard heartbeats: 163 / 164 exceeded the 3 s probe timeout
builtin-auto stage: 11.836 s and 14.676 s in independent builds
complete observed builder: 30.912 s
```

Hidden-model lookup and unrestricted-key permission projection were separately measured in milliseconds and were not changed.

A local exact-output comparison across all 38 built-ins showed:

```text
one-shot preparation per ID: 598 ms
one shared build-local preparation: 95 ms
speedup: 6.29x
exact deep parity: true
```

The first reviewed immutable canary (`bb15de989`, image digest `sha256:8a47bcfe…`) was isolated-smoke healthy but correctly rejected by the guarded production acceptance and automatically restored to the prior healthy digest:

```text
Concurrent Primary / Supplemental: 23.822 s / 39.715 s
Unrelated heartbeat:              13 / 13 exceeded 3 s
Primary-only / Supplemental-only: 15.528 s / 17.612 s
```

[...]

PR #10185, opened 2026-08-12
fix(release-v3.8.50): repair build breaks and restore the Fal credential guard

kwakuoseikwakye:
⚠️ base-red inherited: #9985

## Problem

`release/v3.8.50` does not compile. Every page 500s and the instrumentation graph fails to build. Four independent breaks, each from a merge that landed a new implementation without removing the old contract:

| File | Break | From |
|---|---|---|
| `open-sse/handlers/videoGeneration.ts` | `handleFalVideoGeneration` bound twice — the new `./mediaGeneration/fal.ts` import was added, the old `./videoGeneration/falHandler.ts` one was never removed | #9982 |
| `src/shared/components/modelSelectModalHelpers.ts` | `isProviderModelHidden` never closed, so the parser refuses the file as a module | #9011 |
| `open-sse/executors/tinycmsSigner.ts` | wasm-bindgen default branch points at `wasm_signer_bg.wasm`, which is not in the repo and is generated by nothing | #10087 |
| `src/app/api/v1/models/catalog.ts` | re-exports `getCatalogStaleWhileRevalidateMs` and two `__set…ForTest` hooks that were deleted | #9199 |

## Fixes

- **videoGeneration**: drop the stale import. `videoGeneration/falHandler.ts` had no other reference in the repo, so it is deleted — `mediaGeneration/fal.ts` is a superset (same Grok handling plus `falModelPath` normalization) and already serves the music path.
- **modelSelectModalHelpers**: add the missing `}`.
- **tinycmsSigner**: remove the unreachable branch. `initTinyCmsWasm()` is the only caller and always passes the embedded `WASM_BASE64` bytes, but Turbopack resolves `new URL(…, import.meta.url)` statically and fails on the missing file.
- **catalog**: drop the three dead names from the re-export block. #9199 intentionally replaced the overridable accessor with a fixed `CATALOG_STALE_WHILE_REVALIDATE_MS` window.

## Regression the duplicate import was hiding

Routing Fal video through the shared module lost the credential check the old handler had. `runFalQueue` computed the token but never guarded it, so it sent `Authorization: Key ` (empty) and the prompt reached fal.ai unauthenticated — `tests/unit/video-fal-grok.test.ts` was making a real network call to fal.ai on every run.

The guard goes back into `runFalQueue` itself rather than the video entry point, so `handleFalMusicGeneration` — which had the same hole and no test — is covered too.

## Validation (Hard Rule #18)

Both guard tests were confirmed red→green by toggling the guard off and on:

```
guard disabled:
✖ handleMusicGeneration rejects Fal music requests without credentials
✖ handleVideoGeneration rejects Fal video requests without credentials
guard restored: 8 pass / 0 fail
```

The new music test stubs `globalThis.fetch` to throw, so a regression fails loudly instead of silently reaching upstream.

Targeted suites: `video-fal-grok` 4/4 · `music-generation-handler` 8/8 · `video-generation-handler` 15/15 · `mediaGeneration/fal` 14/14 · `provider-tinycms-web` 16/16 · `model-select-provider-test-helpers` 14/14 · `v1-models-catalog-ttl` 4/4.

`npm run typecheck:core` clean. `npm run lint` 14 → 13 errors (the 3 remaining files are untouched by this PR).

Verified in the app: `/login` returns 200 and renders, `/` → 307 → `/dashboard`, `/api/health` → 401. Before this PR every one of those was a 500.

The malformed-payload assertion in `video-fal-grok.test.ts` moves to the shared module's provider-neutral wording (`returned no media URL`).

## Still red after this PR

Three test files encode the pre-#9199 unbounded-SWR contract and need a decision on the intended `/v1/models` caching behavior — out of scope here, all failing on the base already:

- `tests/unit/model-catalog-cache-swr-8728.test.ts` — 11 fail
- `tests/unit/quota-exclusive-catalog-short-circuit.test.ts` — 3 fail
- `tests/unit/v1-models-discovery-conformance.test.ts` — 2 fail (this file could not even load before this PR)

Separately, 8 i18n keys contain a literal `.` (`capabilityFilter.*`, `degraded.source.*`) and throw `IntlError: INVALID_KEY` on every page render. Non-fatal, pre-existing.

Refs #9985, #9982, #9199, #9011, #10087


diegosouzapw:
Closing as superseded: the build breaks and the Fal credential guard this PR repairs were fixed on `release/v3.8.50` by the base-red sweep (#10213), and the remaining `catalog.ts` hunk targets a catalogCache API shape the base has since moved past. Thank you @kwakuoseikwakye — if any build break still reproduces on the current tip, please open a focused PR against it.

PR #10187, opened 2026-08-12
chore(repo): remove stale one-shot and duplicate helper

backryun:
## Summary

- remove `scripts/docs/move-i18n-mirrors.mjs`, a completed one-shot migration whose own header marks it safe to delete after merge
- remove the unused raw VS Code `modelPresentation.ts` copy; it is byte-identical to the canonical helper and no raw route imports it

## Scope

This intentionally removes only two files with direct, unambiguous evidence. Broader dead-code candidates are left untouched because zero internal callers alone does not prove that a published source module, future feature scaffold, documented contract, or extension hook is disposable.

## Validation

- `npm run typecheck:core`
- `git diff --check`
- repository reference scan: no references to either deleted path
- duplicate proof: both `modelPresentation.ts` copies had the same SHA-256
- one-shot dry run before deletion: `locales=42 moved=0 skipped=0`
- focused VS Code test run: 8 tests passed; 4 suites stop at the existing base failure where `catalogCache` does not export `__setCatalogStaleWhileRevalidateAccessorForTest`
- the same `vscode-token-routes.test.ts` startup failure was reproduced unchanged on the pinned base `838e4d688`


diegosouzapw:
Merged into `release/v3.8.50` — thank you @backryun for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

Issue #2723, opened 2026-05-26
Unable to add Poolside provider with valid API key


gmlnchv:
# Unable to add Poolside provider with valid API key

## Description
When attempting to add a Poolside AI connection with a valid API key (format: `sky_*`), the validation fails and rejects the key as invalid.

## Steps to Reproduce
1. Navigate to Providers page in OmniRoute dashboard
2. Click "Add Connection" for Poolside provider
3. Enter a valid Poolside API key starting with `sky_`
4. Click "Test Connection" or "Save"
5. Validation fails with "Invalid API key" error

## Expected Behavior
The validation should accept valid Poolside API keys with the `sky_` prefix and successfully connect to the Poolside API.

## Actual Behavior
The connection validation fails, preventing the addition of Poolside provider connections.

### Current Configuration
From `open-sse/config/providerRegistry.ts`:
```typescript
poolside: {
  id: "poolside",
  alias: "poolside",
  format: "openai",
  executor: "default",
  baseUrl: "https://api.poolside.ai/v1/chat/completions",
  authType: "apikey",
  authHeader: "bearer",
  models: [{ id: "poolside-model", name: "Poolside Model" }],
}
```

### Validation Flow
The validation uses `validateOpenAILikeProvider` which:
1. First tries GET `/models` endpoint
2. Falls back to POST `/chat/completions` with test message
3. Returns "Invalid API key" on 401/403 responses

### Possible Root Causes
1. **Incorrect base URL**: The registry shows `/v1/chat/completions` as baseUrl, but validation might need just `/v1`
2. **Missing models endpoint**: Poolside API might not expose a `/models` endpoint
3. **Auth header format**: Might need different auth header configuration
4. **API endpoint mismatch**: The actual Poolside API structure might differ from OpenAI compatibility

## Environment
- OmniRoute version: Latest (as of 2026-05-26)
- Provider: Poolside AI
- API key format: `sky_*`

## Suggested Investigation
1. Verify the correct Poolside API base URL and endpoint structure
2. Check if Poolside exposes a `/models` endpoint for validation
3. Test the actual API response when using a valid `sky_*` key
4. Consider adding Poolside-specific validation logic if it deviates from OpenAI compatibility

## Additional Context
The schema validation in `src/shared/validation/schemas.ts` doesn't impose any format restrictions on API keys, so the rejection is happening during the actual API call validation, not client-side validation.

diegosouzapw:
[...]

PR #9210, opened 2026-08-02
feat(providers): integrate audited free-tier gateways

diegosouzapw:
## Summary

- Integrate the audited free-tier and experimental provider waves into v3.8.50.
- Synchronize the root README, provider/free-tier references, architecture docs, diagrams, llm.txt, and all 42 localized documentation sets with live inventories.
- Add audited Void AI and HelixMind gateways as experimental integrations after public re-audit.
- Keep Glhf.chat blocked, Anakin.ai web-only, and Kluster AI invalid/dead.
- Keep LongCat documented as a one-time 10M-token signup grant after account creation and KYC; historical changelogs remain untouched.

## Live inventory

| Inventory | Verified value |
| --- | ---: |
| Provider catalog entries | 329 |
| Entries marked free/no-auth | 155 |
| No-auth / OAuth / web-cookie / API-key | 9 / 23 / 31 / 225 |
| Quantified free-tier pools / model budget entries | 43 / 522 |
| Documented recurring token budget | ~1.53B/month |
| MCP tools / scopes | 107 / 32 |
| Executors / top-level services | 89 / 178 |
| Routing strategies / Auto-Combo factors | 19 / 13 |
| Compression engines | 12 |
| CLI tools and coding agents | 33 |
| Agent Skills | 45 |
| A2A skills / cloud agents | 6 / 4 |
| Locales | 43 total, English plus 42 localized sets |

## Integration and re-audit status

- 29/32 selected services are integrated: the 27 original migration-wave providers plus Void AI and HelixMind.
- Void AI: public API, health, models, Chat/Responses contracts observed; free signal remains conditional and no numeric quota is encoded.
- HelixMind: public OpenAI/Anthropic-compatible API and auth contracts observed; current free tier is not confirmed, so metadata uses hasFree false.
- Remaining out: Glhf.chat (Cloudflare 522), Anakin.ai (web-only), Kluster AI (API discontinued/migrated).
- Public matrix completed for all 29 integrations. Poolside returned HTTP 401 with No Authorization header provided. HelyxAI /v1/models returned 404 while site/docs and chat/responses contracts remained active.
- No authenticated validation was performed: local database had 0 configured connections. No accounts, ToS, KYC, cards, billing, overage, or prompts were used.

## Documentation and safeguards

- Regenerated the 329-entry provider reference and synchronized localized llm.txt mirrors.
- Updated the API-key partition characterization from 223 to 225.
[...]

PR #10216, opened 2026-08-12
feat(providers): publish Poolside's probed Laguna Preview catalog

### LINKED ISSUE #9085: feat(providers): re-add Poolside as first-party provider
### Problem / Use Case
Poolside was previously in the registry (added in #2590) but was **removed in #3054** ("remove 9 dead/unreachable free providers") because live probes returned 000/404/empty and users couldn't validate their `sky_*` keys (#2723). The root cause documented in #2723 was that the configured base URL (`https://api.poolside.ai/v1/chat/completions`) had no `/v1/models` endpoint and returned 401 from unknown routes, which the generic `/models` probe misread as "invalid API key".

Since then, Poolside has moved to **self-serve API keys** (`platform.poolside.ai`) and a **different, working base host**. Its **Laguna** coding models are now free in Preview. OmniRoute users can already *see* these models today — but only indirectly, because OmniRoute surfaces `laguna-*:free` via aggregators (`open-sse/config/providers/registry/cline/index.ts`, `.../routeway/index.ts`, and OpenRouter). There is currently **no first-party Poolside route**, so users can't bring their own free Poolside key.

### Proposed Solution
Re-add a first-party `poolside` registry entry with the corrected, self-serve configuration:

```ts
// open-sse/config/providers/registry/poolside/index.ts
import type { RegistryEntry } from "../../shared.ts";

export const poolsideProvider: RegistryEntry = {
  id: "poolside",
  alias: "poolside",
  format: "openai",
  executor: "default",
  baseUrl: "https://inference.poolside.ai/v1/chat/completions",
  authType: "apikey",
  authHeader: "bearer",
  models: [
    { id: "laguna-xs-2.1", name: "Laguna XS 2.1 (33B)", contextLength: 131072, toolCalling: true },
    { id: "laguna-s-2.1",  name: "Laguna S 2.1 (118B)", contextLength: 131072, toolCalling: true },
  ],
};
```

Reuse the Poolside-specific validator that already existed before removal (see the `poolside:` branch deleted in #3054): validate via a **direct minimal `chat/completions` probe**, not the `/models` probe, since the Poolside API does not expose `/v1/models`. This is exactly the fix #2723 called for.

- **Free tier:** Laguna XS 2.1 (33B) and Laguna S 2.1 (118B) free in Preview via a self-serve developer key at `platform.poolside.ai`.
- **OpenAI-compatible base URL:** `https://inference.poolside.ai/v1`
- **Auth:** bearer key (historically `sky_*` prefix per #2723 — confirm current prefix at validation time; do not hard-restrict the format, matching `src/shared/validation/schemas.ts`).
- **Docs:** https://docs.poolside.ai/api/overview

### Alternatives Considered
- **Status quo (aggregator-only):** Laguna is reachable via cline/routeway/openrouter, but users can't use their own free Poolside key, don't get first-party rate limits, and pay aggregator markups on non-free routes.
- **Leave removed:** The 2026-06 removal was correct *at the time* (dead endpoint), but the removal reason no longer holds — the endpoint and onboarding changed.

### Acceptance Criteria
- A `poolside` registry entry exists and is routable first-party (not only via aggregators).
- "Test Connection" succeeds with a valid Poolside key using the `chat/completions` probe (no false "invalid API key" from a missing `/models` endpoint) — regression coverage for #2723.
- `laguna-xs-2.1` and `laguna-s-2.1` appear in the model catalog under provider `poolside`.
- A chat completion round-trips end-to-end against `https://inference.poolside.ai/v1`.
- Existing aggregator-surfaced `laguna-*:free` routing remains unchanged (no model-name collisions).

### Expected Test Plan
- Unit: `tests/unit/poolside-provider.test.ts` — registry registration + validator path (assert `/models` is not required; probe hits `chat/completions`).
- Unit: model catalog includes both Laguna IDs under `poolside`.
- Integration: extend provider-journey contract test to cover the Poolside validate→chat flow.
- Keep `npm run test:coverage` at threshold.

### Additional Context
Two smaller, related first-party gaps (both LLM/chat, both with a genuine free tier, both absent from the registry) — could be folded in or filed separately:

| Provider | Free tier | OpenAI-compatible base URL | Notes |
|---|---|---|---|
| **IBM watsonx.ai (Lite)** | 300,000 tokens/month, recurring | Yes, but region/deployment-specific base URL | Already listed in `ENTERPRISE_CLOUD_PROVIDER_IDS` (src/shared/constants/providers.ts) yet has no registry folder — categorized but not routable. Card required at signup. |
| **Arli AI** | Free plan, ~5 req / 2-day window (very restrictive) | `https://api.arliai.com/v1` | Fully OpenAI-compatible; low priority due to tiny quota. |

Source for all data: independently verified against provider docs (`last_verified` 2026-07-31).

pacocartones:
## Summary

The Poolside registry entry landed in #9210 with an empty `models` list, because the public matrix could only reach the unauthenticated endpoint (`401 No Authorization header provided`). This fills that list in from the catalog an authenticated probe returned, so the two Preview models are routable and visible before a key is configured instead of only after live discovery.

| Model ID | Name | Context | Max completion | Features |
| --- | --- | ---: | ---: | --- |
| `poolside/laguna-xs-2.1` | Laguna XS 2.1 | 262144 | 32768 | tools, reasoning — text only |
| `poolside/laguna-s-2.1` | Laguna S 2.1 | 262144 | 32768 | tools, reasoning — text only |

`passthroughModels` stays on: the static list is what the catalog returned, not a ceiling, so a model the Preview adds later still routes without a release.

## Evidence

Probed 2026-08-07 against `https://inference.poolside.ai/v1`. The raw output is already in the issue, so it is linked rather than repeated:

- unauthenticated probe — https://github.com/diegosouzapw/OmniRoute/issues/9085#issuecomment-5210676556
- authenticated probe — https://github.com/diegosouzapw/OmniRoute/issues/9085#issuecomment-5213367262

What the authenticated run settled:

- `/v1/models` exists and answers 200 with a key; `/v1/chat/completions` answers 200. Without a key `/v1/models` answers 401 with the body `No Authorization header provided` — the response a generic probe reads back as "invalid key", which is the root cause behind the removal in #3054 and the reports of unusable-but-valid keys in #2723.
- Exactly two models, both free during Preview (price 0).
- `supported_features` includes `tools` and `reasoning`; no image or audio input on either model.
- Served through Baseten (`X-Baseten-*` response headers). No rate-limit headers are published, so no numeric quota is encoded here — the existing `freeNote` ("no public numeric quota is published") remains accurate.

## The XS model ID

Third-party listings circulate `laguna-xs.2` for the small model, and the aggregator catalogs in this repo carry their own namespaced variants (`routeway` has `laguna-xs.2:free`; `cline` has `poolside/laguna-m.1:free`). Those address the aggregator, not this host. The live catalog on `inference.poolside.ai` returns `poolside/laguna-xs-2.1`, and that is the form used here. `tests/unit/poolside-registry-models-9085.test.ts` pins the exact IDs so a later sweep cannot reintroduce the dotted variant.

## Scope notes

- No pricing is encoded: `RegistryModel` has no pricing field, and the free signal already lives in the provider metadata (`hasFree` + `freeNote`).
- `maxInputTokens` is deliberately unset. The schema documents it as an override for when the input budget is smaller than the full context window; the catalog reports no such split, only `context_length` and `max_completion_tokens`.
- `freeModelCatalog.data.ts` is untouched: it is generated, and its records require quota fields the upstream does not publish.

## Tests

New: `tests/unit/poolside-registry-models-9085.test.ts` (4 tests) — exact IDs, the XS-ID guard, the probed capability numbers, and that passthrough discovery stays on.

Also run green against this change: `free-tier-providers-phase3-integration`, `provider-registry-models-guard`, `free-model-catalog`, `free-tier-catalog`, `free-note-freshness`, `catalog-order-contract`, `catalog-synced-static-preservation`, `model-capabilities-registry`, `model-metadata-registry`, `provider-alias-uniqueness`, `check-provider-consistency`, `openrouter-passthrough-models`, `clinepass-provider`, `chat-openai-compat-providers`, `provider-audit`, `cli-catalog-counts`.

Closes #9085


diegosouzapw:
Merged into `release/v3.8.50` — thank you @pacocartones for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

PR #5976, opened 2026-07-02
fix(combo): fallback to sibling model on 500 for per-model-quota providers

hartmark:
## Summary

When a combo contains multiple models from the same provider (e.g. two Gemini models) and the first model returns a server error (500), the combo now correctly falls back to the sibling model instead of retrying the same failing model and returning the error to the client.

Before: Request → gemma-4-31b-it → 500 → retry same model → 429 "cooling down" → error returned to client. The second model (gemma-4-26b-a4b-it) was never tried.

After: Request → gemma-4-31b-it → 500 → skip (model locked) → gemma-4-26b-a4b-it → success.
This affects any combo with multiple models from the same provider — the sibling fallback now works as intended instead of silently failing when one model has a transient server error.

## Tests Added Or Updated

- added tests for relevant code

PR #2417, opened 2026-05-19
feat: add new feature on combos - falloverBeforeRetry

hartmark:
## Summary

I'm mostly using gemini provider and they have two models. Gemma-4 31B and Gemma-4 24B, they have a 10 RPM limit so when I get 429 on one model it should go over to the next.

I have digged through the code and with help of AI identified that it seems that combos are handling fail over to another model on the _same_ provider.

This PR should address this. All tests and added tests are in place.

PR #10217, opened 2026-08-12
fix(combo): make failoverBeforeRetry actually skip the same-model retry

hartmark:
## Summary

- `combo.config.failoverBeforeRetry` was documented/intended to prefer the next sibling target over retrying the same model again on a transient error, but the same-target retry loops in both the priority/auto strategy and the round-robin strategy never actually checked it. They only checked `isTransient`/`retry < maxRetries`/`!providerExhausted`, so a rate-limited model still got `maxRetries + 1` back-to-back attempts on itself before the loop ever moved to a sibling — regardless of the setting.
- The setting was only wired into `skipUpstreamRetry`, a separate lower-level retry mechanism passed down to `handleSingleModelWithTimeout`, so it had *some* effect, just not the one its name and #2417's original intent describe for the combo-level retry loop.
- Fix: both retry-gate conditions now also check `config.failoverBeforeRetry`, but only skip the same-model retry when a sibling target (`nextTarget` / `offset + 1 < modelCount`) is actually available to fail over to. With no sibling left, it still retries the same model — skipping there would just burn the last attempt for nothing instead of giving the request one more chance.

## Related Issues

- Related to #2417, #5976, #6216 (which scoped a related but narrower fix to 500s on per-model-quota providers only)

## Validation

- [x] Focused tests for the change: `node --import tsx --test tests/unit/combo-routing-engine.test.ts` (88 pass)
- [x] `node --import tsx --test tests/unit/combo-config.test.ts` (44 pass)
- [x] Broader regression sweep: `combo-cooldown-retry`, `combo-strategy-fallbacks`, `combo-rr-fallback-advance-948`, `combo-provider-cooldown(-sibling)`, `combo-same-provider-cascade`, `combo-breaker-429`, `combo-strategies` (87 pass)
- [x] `npm run lint` — clean on both changed files; explicit-`any` budget for the test file unchanged (267, matching the existing suppression baseline)
- [x] Production-code change (`open-sse/services/combo.ts`) includes new automated tests in this PR
- [x] Regression tests verified to fail on pre-fix code (`git stash` the source change, rerun) before being verified to pass post-fix — satisfies this repo's TDD rule that a regression test must fail for the intended reason before the fix lands

## Tests Added Or Updated

- `tests/unit/combo-routing-engine.test.ts`:
  - `handleComboChat round-robin: failoverBeforeRetry skips the same-model retry and goes straight to the sibling`
  - `handleComboChat priority strategy: failoverBeforeRetry skips the same-model retry and goes straight to the sibling`

## Coverage Notes

- Both new tests exercise `open-sse/services/combo.ts`'s same-target retry loop directly through `handleComboChat`, the real entry point, not a mock of the retry logic.

## Reviewer Notes

- Found via a live incident: a real request against the default combo (`strategy: "auto"`, `failoverBeforeRetry: true`, `maxRetries: 3`) showed 4 consecutive 429s on the same model over ~5s before finally failing over to a sibling on the same provider — exactly the `maxRetries: 3` (1 initial + 3 retries) same-model-hammering behavior this PR removes.
- The "no sibling available" carve-out matters: without it, this fix broke the pre-existing test `handleComboChat round-robin retries a transient failure on the same model before succeeding`, which uses a single-model combo. `failoverBeforeRetry` defaults to `true` at config-resolution time (`comboConfig.ts`), so a naive unconditional skip would have made every single-model combo give up on the first transient error instead of retrying — that regression is now covered by keeping the pre-existing test green.


diegosouzapw:
Merged into `release/v3.8.50` — thank you @hartmark for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

PR #10218, opened 2026-08-12
fix(deepseek-web): classify business auth rejection as 401

Zartharas:
## What changed

- Preserve literal DeepSeek `401` / `403` status codes from `validateDeepSeekWebProvider` so the connection-test classifier can distinguish authentication failures from generic upstream errors.
- Treat DeepSeek application-level business code `40003` (`Authorization Failed`) returned under HTTP 200 as an auth-classifiable `statusCode: 401`.
- Keep all other HTTP-200 / missing-token business failures generic; they are **not** synthesized into credential failures.
- Add focused regressions for HTTP 200 + 40003, unknown business codes, literal 401/403, and the healthy derived-token response.

## Root cause

DeepSeek can represent authentication rejection under HTTP 200. The validator previously rejected the missing-token response without returning a `statusCode`, so the connection-test classifier fell through to `upstream_error`. A controlled Auth Keeper integration test reproduced `diagnosis=upstream_error`, `code=upstream_error`, `statusCode=null` for a fixed synthetic invalid DeepSeek session and then safely auto-restored the good credential.

A direct synthetic request from the same host subsequently confirmed the provider semantics without using or printing any real credential:

```text
httpStatus=200
businessCode=40003
messageClass=AUTHORIZATION_FAILED
accessTokenPresent=false
rawBodyPrinted=false
credentialPrinted=false
credentialType=FIXED_SYNTHETIC_ONLY
```

DeepSeek integration issue `deepseek-ai/awesome-deepseek-integration#654` independently documents the same HTTP-200 `{ "code": 40003, "msg": "Authorization Failed" }` pattern.

## Safety

This is intentionally narrow:

- arbitrary nonzero business codes remain generic and are not synthesized into credential failures;
- tests and live characterization use fixed synthetic credentials only;
- no Auth Keeper recovery policy was broadened;
- no real credential or raw DeepSeek response body is included in logs or test output.

## Validation

- Final commit: `e2e8ca09d73e9892749ed36b07afb3d52464418b`.
- Direct parent/current upstream `release/v3.8.50`: `f1eb0b835722ca744692834c84661f1bc8d564a6`.
- Focused regression: **4 passed, 0 failed**.
- Scoped ESLint for the validator and regression file: **PASS, 0 findings**.
- Semgrep on the earlier equivalent functional diff: **PASS**.
- `npm ci`: **0 vulnerabilities** reported.
- Scope: **one production validator file + one focused regression file**.

The final checkout used `--depth 1`, so the final `git diff f1eb0b8..HEAD` display command could not resolve the parent object locally. This was a shallow-clone inspection limitation only; the final branch is independently verified on GitHub as one commit directly ahead of `f1eb0b8` with exactly the two intended files.

### DIFF
--- a/src/lib/providers/validation/webProvidersA.ts
+++ b/src/lib/providers/validation/webProvidersA.ts
@@ -116,6 +116,7 @@ export async function validateDeepSeekWebProvider({ apiKey }: any) {
       return {
         valid: false,
         error: "userToken is invalid or expired — get a fresh one from localStorage",
+        statusCode: resp.status,
       };
     }
     if (!resp.ok) {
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Merged into `release/v3.8.50` — thank you @Zartharas for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

Issue #6637, opened 2026-07-08
fix(resilience): combo stops fallback after Kimi total token-limit 400 from Responses xhigh request


Squawk7777:
### OmniRoute Version

3.8.46

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Containerized deployment; host OS not verified.

### Node.js Version

24.18.0 inside the container.

### Provider(s) Involved

`kimi-coding-apikey`, `openai-compatible-responses` (MiniMax)

### Model(s) Involved

`combo/model-1.0`, `kimi-k2.7-code`, `MiniMax-M3`

### Client Tool

Codex Desktop via `/v1/responses`

### Description

A priority combo can route a large OpenAI Responses request with `reasoning.effort="xhigh"` to `kimi-coding-apikey/kimi-k2.7-code`, even though the translated provider request exceeds Kimi's total model-token/context limit. After Kimi returns a 400 token-limit error, combo fallback stops instead of trying the later larger-context target (`MiniMax-M3`).

This overlaps with #6524 in the broad category "reasoning requests can produce an invalid upstream `max_tokens`", but this case looks distinct:

- the client did **not** send `max_tokens`, `max_completion_tokens`, or `max_output_tokens`;
- the problematic provider `max_tokens=195072` appears to be created by the OpenAI/Responses -> Claude/Kimi translation path from default response room plus xhigh thinking budget (`64000 + 131072`), not from a client limit expanded by `reasoningTokenBufferEnabled`;
- #6536 defaults combo reasoning-token buffering to off, but that does not appear sufficient for this case because the invalid `max_tokens` is already produced by translation/default thinking-budget fitting;
- the fallback-stopping behavior is a separate issue: Kimi's `Your request exceeded model token limit` 400 is not treated as a context/token overflow that should fall through to the next combo target.

In the observed request, the combo was a priority combo with Kimi before MiniMax. Context-aware fallback kept 4 targets, selected Kimi first, then stopped after Kimi's 400 instead of trying MiniMax.

### Steps to Reproduce

1. Create a priority combo named `combo/model-1.0` with at least:
   - first compatible target: `kimi-coding-apikey/kimi-k2.7-code`
   - later compatible target with a larger context window: `MiniMax-M3` via an OpenAI-compatible Responses provider
2. Send a large `/v1/responses` request to `combo/model-1.0`:
   - `reasoning.effort = "xhigh"`
   - no explicit `max_tokens`
[...]

Issue #6524, opened 2026-07-07
fix(providers): clamp reasoning token buffer to model output-token cap


xz-dev:
## Summary

`reasoningTokenBufferEnabled` currently expands `max_tokens` for thinking/reasoning models with a fixed heuristic:

```ts
Math.max(current + 1000, Math.ceil(current * 1.5))
```

This can turn a valid request into an invalid upstream request when the expanded value exceeds the model's real maximum output-token limit, especially when model capability metadata is wrong or overly optimistic.

## Observed production behavior

For `ollama-cloud/deepseek-v4-flash`, a combo request with client `max_tokens=64000` was transformed into provider `max_tokens=96000`:

```text
Reasoning model ollama-cloud/deepseek-v4-flash: adjusted max_tokens 64000 -> 96000
```

The upstream rejected the request:

```text
max_tokens (96000) exceeds model's maximum output tokens (65536) for model deepseek-v4-flash
```

A direct boundary test through OmniRoute to `ollama-cloud/deepseek-v4-flash` (no combo) confirmed the actual upstream boundary:

| max_tokens | result |
|---:|---|
| 65536 | 200 OK |
| 65537 | 400: exceeds maximum output tokens (65536) |
| 96000 | 400: exceeds maximum output tokens (65536) |

The local capability row had `limit_context=1048576` and `limit_output=1048576`, so the buffer logic believed `96000` was safe even though the upstream real output cap was `65536`.

## Why this is a problem

`reasoningTokenBufferEnabled` is useful as a compatibility feature to avoid reasoning tokens consuming the visible answer budget. However, it should not send a `max_tokens` value larger than the model's known output-token limit.

The current behavior is also surprising because the model works correctly when used directly with a legal `max_tokens`; the failure is introduced by OmniRoute's automatic buffer.

## Expected behavior

If the model has a known `maxOutputTokens` / output cap, buffer should clamp to that cap:

```ts
const buffered = Math.max(current + 1000, Math.ceil(current * 1.5));
return maxOutputTokens !== null ? Math.min(buffered, maxOutputTokens) : buffered;
```

For example:

```text
current = 64000
buffered = 96000
known output cap = 65536
result should be 65536, not 96000
```

If no model output cap is known, keeping the existing `1.5x` heuristic is acceptable, though it may still be useful to make this policy provider/model configurable.

## Suggested follow-ups

[...]

PR #10238, opened 2026-08-13
fix(translator): resolve the Claude thinking output cap with the routed provider

### LINKED ISSUE #10139: fix(routing): translated `max_tokens = 64000 + thinking budget` is never clamped to the model output cap — follow-up to #6637, unfixed in v3.8.49
## OmniRoute Version

3.8.49

## Installation Method

npm (global)

## Operating System

Linux

## OS Version

Ubuntu, npm global install

## Node.js Version

22.22.3

## Provider(s) Involved

`opencode-go` (Anthropic Messages endpoint), and the same model behind an OpenAI-compatible chat provider used as a
priority-combo fallback

## Model(s) Involved

`qwen3.7-plus` (documented max output 131072); also affects every opencode-go model routed to `/v1/messages`

## Client Tool

Custom agent client sending OpenAI-format `/v1/chat/completions`

---

## Description

This is the token-budgeting half of #6637, which was closed by #6893. #6893 fixed only the *fallback classification*
(`isContextOverflow400()` in `combo.ts` now reuses `CONTEXT_OVERFLOW_PATTERNS`), touching `combo.ts`,
`accountFallback.ts` and a test. The part the reporter explicitly flagged as distinct —

> the problematic provider `max_tokens=195072` appears to be created by the OpenAI/Responses -> Claude/Kimi
> translation path from default response room plus xhigh thinking budget (`64000 + 131072`), not from a client limit
> expanded by `reasoningTokenBufferEnabled`

— was never addressed, and still reproduces on v3.8.49 with a different provider and a different client format
(`/v1/chat/completions` instead of `/v1/responses`). The identical `64000 + 131072 = 195072` appears, so this is the
same translation path, not a Kimi-specific quirk.

**The gap in one sentence:** #6524 clamped the *reasoning-buffer* path, which only runs when the client already sent
`max_tokens`; when the client sends **no** max-token field, translation synthesizes one and nothing clamps it.

### Why nothing clamps it

Both consumers of the model output cap require `max_tokens` to already exist in the request body:

1. `enforceOutputTokenBudget(body, estIn, ctxLimit, reserved, cap)` only rewrites fields that are already present —
   `["max_tokens", "max_completion_tokens", "max_output_tokens"]`. With none of them present, there is nothing to
   rewrite, and it returns `ok` without touching the synthesized value. It is also the site that introduces the 64000:
   it is called with `targetFormat === CLAUDE && sourceFormat !== CLAUDE ? DEFAULT_MAX_TOKENS : 0` as the reserved
   output argument, and `DEFAULT_MAX_TOKENS = 64e3`.

2. The combo-side adjustment guards on the incoming value, so it is skipped entirely:

   ```js
   const e = toPositiveInteger(body.max_tokens);
   const t = resolveReasoningBufferedMaxTokens(model, body.max_tokens, { enabled });
   if (e !== null && t !== null && t !== e) body.max_tokens = t;   // e === null here
   ```

   (`resolveReasoningBufferedMaxTokens` itself also returns `null` when `getExplicitModelOutputCap()` is `null`.)

Compounding it, `getExplicitModelOutputCap()` resolves through four sources (manual override → `limit_output` →
static model table `maxOutputTokens` → name-match table) and returns `null` for `qwen3.7-plus`, so even a request
that *did* carry `max_tokens` would have had no cap to clamp against.

For contrast, the Gemini path already does the right thing — it calls `capMaxOutputTokens(model, ...)` before setting
`generationConfig.maxOutputTokens`. The Claude path has no equivalent.

Code references are from the shipped v3.8.49 build
(`node_modules/omniroute/dist/.build/next/server/chunks/`), so identifier names are post-bundling; the logic maps
directly to the `src/` names quoted above.

## Steps to Reproduce

1. Configure a provider whose upstream is Anthropic Messages (`opencode-go`), with a model whose maximum output is
   **below** `64000 + 131072`, e.g. `qwen3.7-plus` (max output 131072).
2. Send an OpenAI-format request to `/v1/chat/completions`:
   - `reasoning_effort: "high"`
   - **no** `max_tokens`, `max_completion_tokens` or `max_output_tokens`
   - **no** `stream` field (non-streaming)
   - any messages and tools
3. Inspect the pipeline log for the call.

Minimal client body:

```json
{
  "model": "<connection>/qwen3.7-plus",
  "messages": [ { "role": "user", "content": "hi" } ],
  "reasoning_effort": "high",
  "temperature": 0.4
}
```

## Expected Behavior

The synthesized `max_tokens` is clamped to the model's output cap (or, when no cap is registered, to a value the
upstream will accept) before the provider call — the same treatment `capMaxOutputTokens()` already gives the Gemini
path. The request should succeed.

## Actual Behavior

`pipeline.providerRequest`:

```json
{
  "model": "qwen3.7-plus",
  "max_tokens": 195072,
  "stream": false,
  "thinking": { "type": "enabled", "budget_tokens": 131072 },
  "tools": [ ... ]
}
```

`pipeline.providerResponse`: **400**, body `{"model":"qwen3.7-plus"}` (no message field).

Every request of this shape fails; it is not intermittent.

### Boundary measurement

The upstream cap is exactly the model's documented 131072 output tokens. Bisected with direct upstream calls:

| upstream `max_tokens` | result |
|---|---|
| 131072 | 200 |
| 131073 | 400 |
| 195072 | 400 |

Reproduced identically **without OmniRoute in the path**, so the limit is the provider's, not the gateway's.

Note that lowering the client's `max_tokens` does not help — `max_tokens: 1000` + `reasoning_effort: "high"` still
fails, because the synthesized total is what exceeds the cap. Dropping `reasoning_effort` succeeds.

### Note on combo masking

With a priority combo whose first step is the Anthropic-endpoint provider and whose second step is an
OpenAI-compatible chat provider serving the same model, the client receives **200** — the 400 is absorbed by
fallback and served by the paid second step. This is presumably intended fallback behavior, but it makes the defect
invisible from the client side and silently shifts spend. Mentioning it only because it is why this went unnoticed
for a long time; the same-round pipeline logs show step 1 at 400 and step 2 at 200.

## Test Impact

Needs a new unit test

## Error Logs / Output

```text
POST /v1/chat/completions
sourceFormat: openai
targetFormat: claude
requestedModel: <connection>/qwen3.7-plus
provider: opencode-go
status: 400

clientRawRequest.body: no max_tokens, no max_completion_tokens, no max_output_tokens, no stream
providerRequest.body:  max_tokens=195072, stream=false, thinking.budget_tokens=131072
providerResponse.body: {"model":"qwen3.7-plus"}
```

## Additional Context

Registering `max_output_tokens` for the model via `PUT /api/provider-models` (which populates
`customModels.outputTokenLimit` and feeds `getExplicitModelOutputCap()`) does **not** work around this, for the same
reason — the clamp sites are not on this code path.

Unrelated but adjacent: `max_output_tokens`, `max_input_tokens` and `contextWindowOverride` are all accepted by
`PUT /api/provider-models` but none are exposed in the dashboard's per-model Compatibility panel, which makes this
class of problem hard to diagnose without reading the source. Happy to file that separately if useful.

Workaround in use: have the client build native Anthropic requests so the openai→anthropic translation never runs.
Verified end-to-end on the same workload — zero 400s.

jmche:
## Problem

`fitThinkingToMaxTokens()` (`open-sse/translator/request/openai-to-claude/thinkingBudget.ts`) clamps the synthesized `max_tokens` to the model's output cap, but resolves that cap from a bare model id:

```ts
const modelCap = safeCapMaxOutputTokens(model);   // -> capMaxOutputTokens(model)
```

A cap that is only known **per provider** — an operator `max_output_tokens` override set via the dashboard, a synced catalog `limit_output`, or a registry entry scoped to one connection — is invisible to a bare-model lookup. `capMaxOutputTokens()` already accepts a `{ provider, model }` object and resolves provider-scoped data when given one (`resolveCapabilityInput()`), but this call site never passes it, so `modelCap` comes back `null` and the unbounded branch runs:

```ts
target = modelCap === null ? responseRoom + requestedBudget : Math.min(responseRoom + requestedBudget, modelCap);
```

Concretely: a client that sends no max-token field at all gets `DEFAULT_MAX_TOKENS` (64000) as `responseRoom`, and `reasoning_effort: "high"` supplies a 131072 thinking budget. With no cap resolved, `target = 195072`. For `opencode-go/qwen3.7-plus`, the real upstream ceiling (per boundary testing: `131072` → 200, `131073` → 400) is only registered as a provider-scoped override — so every such request went out as `max_tokens: 195072` and was rejected upstream with a bare `400 {"model":"qwen3.7-plus"}`.

The caller in `openai-to-claude.ts` already has the routed provider in scope two lines earlier (used immediately after for the Kimi-coding check) — it just never threads it into this call:

```ts
const routedProvider = credentials?._provider;
const isKimiCoding = routedProvider === "kimi-coding" || ...;
...
const fitted = fitThinkingToMaxTokens(model, Number(result.max_tokens) || 0, result.thinking);  // no provider
```

Follow-up to #6637, whose token-budgeting half was never addressed — #6893 fixed only the combo fallback classification, not the translation-time cap resolution.

## Fix

`fitThinkingToMaxTokens()` takes an optional 4th argument `provider`, forwarded into `capMaxOutputTokens({ provider, model })` — no new lookup path, this object shape is already supported. The caller now passes the already-in-scope `routedProvider`. Omitting the argument (any other existing caller) keeps the exact prior behavior — verified below.

This is the same provider-aware read path #6524 aligned the reasoning-token-buffer clamp onto.

## Verification

New regression test `tests/unit/repro-10139-claude-thinking-output-cap.test.ts`, modeled on this repo's own `repro-6524.test.ts` DB-fixture pattern, against a real `opencode-go/qwen3.7-plus` `max_output_tokens` override:

```
without provider (documents the pre-fix defect): maxTokens = 195072 (uncapped)
with provider:                                    maxTokens = 131072, thinking.budget_tokens = 67072
no override present:                              behavior identical with/without provider argument
```

Red→green, same tree, only the fix commit toggled:

```
git stash (fix only)   -> tests/unit/repro-10139-claude-thinking-output-cap.test.ts: not ok 2, # pass 2 # fail 1
git stash pop           -> # pass 3 # fail 0
```

No regressions in the surrounding suites:

```
tests/unit/translator-openai-to-claude.test.ts             17/17
tests/unit/openai-to-claude-thinking-budget-split.test.ts   3/3   (leaf/host split-guard, unaffected)
tests/unit/repro-6524.test.ts                                2/2
tests/unit/minimax-m3-maxtokens.test.ts                      5/5
tests/unit/output-token-budget-model-cap.test.ts            12/12
```

`npm run check:changelog-integrity` passes; a `changelog.d/fixes/` fragment is included. Commit went through the full pre-commit hook (`docs-sync`, `check:any-budget:t11`, lint-staged) with no `--no-verify` needed.

## Note

Rebased onto the `openai-to-claude/thinkingBudget.ts` extraction that landed in the recent provider-registry reorg — the defect and fix are unchanged in substance from the original patch written against the inline code, only the file location moved.

Closes #10139

### DIFF
--- a/changelog.d/fixes/10139-thinking-output-cap-provider-scope.md
+++ b/changelog.d/fixes/10139-thinking-output-cap-provider-scope.md
@@ -1,0 +1,1 @@
+- **fix(translator):** resolve the Claude thinking output cap with the routed provider so a provider-scoped-only `max_output_tokens` override is no longer invisible to `fitThinkingToMaxTokens()`, which previously let the synthesized `max_tokens` (caller room + thinking budget) go out unbounded and 400 upstream ([#10139](https://github.com/diegosouzapw/OmniRoute/issues/10139))
--- a/open-sse/translator/request/openai-to-claude.ts
+++ b/open-sse/translator/request/openai-to-claude.ts
@@ -240,7 +240,12 @@ export function openaiToClaudeRequest(model, body, stream, credentials = null) {
   // could exceed model caps (e.g. Opus 4.7's 128000 ceiling) and trigger
   // HTTP 400 from Anthropic.
   if (!isKimiCoding) {
-    const fitted = fitThinkingToMaxTokens(model, Number(result.max_tokens) || 0, result.thinking);
+    const fitted = fitThinkingToMaxTokens(
+      model,
+      Number(result.max_tokens) || 0,
+      result.thinking,
+      routedProvider
+    );
     result.max_tokens = fitted.maxTokens;
     if (fitted.thinking === undefined) {
       delete result.thinking;
--- a/open-sse/translator/request/openai-to-claude/thinkingBudget.ts
+++ b/open-sse/translator/request/openai-to-claude/thinkingBudget.ts
@@ -7,9 +7,9 @@ import { capMaxOutputTokens } from "../../../../src/lib/modelCapabilities.ts";
 const MIN_CLAUDE_THINKING_BUDGET = 1024;
 const MIN_RESPONSE_ROOM = 1024;
 
-function safeCapMaxOutputTokens(model: string): number | null {
+function safeCapMaxOutputTokens(model: string, provider?: string | null): number | null {
   try {
-    const cap = capMaxOutputTokens(model);
+    const cap = capMaxOutputTokens(provider ? { provider, model } : model);
     return typeof cap === "number" && cap > 0 ? cap : null;
   } catch {
     return null;
@@ -31,6 +31,12 @@ function safeCapMaxOutputTokens(model: string): number | null {
  *     responseRoom shrunk to MIN_RESPONSE_ROOM; if still below MIN, disable
  *     thinking entirely (cap too tight for any reasoning).
  *
+ * `provider` scopes the cap lookup to a provider-specific override (e.g. a
+ * dashboard-set `max_output_tokens` for `opencode-go/qwen3.7-plus`) when the
+ * model-only entry has no cap of its own. Without it, a model whose real
+ * ceiling is only known per-provider resolves to no cap at all and the
+ * synthesized `max_tokens` goes out unbounded (#10139).
+ *
  * Worked example (real-world Opus 4.7 case that previously 400'd):
  *   caller max_tokens = 32000, reasoning_effort=high → budget = 131072,
  *   model cap = 128000.
@@ -42,9 +48,10 @@ function safeCapMaxOutputTokens(model: string): number | null {
 export function fitThinkingToMaxTokens(
   model: string,
   callerMaxTokens: number,
-  thinking: Record<string, unknown> | undefined
+  thinking: Record<string, unknown> | undefined,
+  provider?: string | null
 ): { maxTokens: number; thinking: Record<string, unknown> | undefined } {
-  const modelCap = safeCapMaxOutputTokens(model);
+  const modelCap = safeCapMaxOutputTokens(model, provider);
   const requestedBudget = Number(thinking?.budget_tokens) || 0;
 
   // No budgeted thinking — just cap max_tokens to the model output ceiling.
--- a/tests/unit/repro-10139-claude-thinking-output-cap.test.ts
+++ b/tests/unit/repro-10139-claude-thinking-output-cap.test.ts
@@ -1,0 +1,110 @@
+/**
+ * Regression test for issue #10139 (follow-up to #6637).
+ *
+ * Reporter observed: for a model whose real output ceiling is only known via a
+ * provider-scoped override (e.g. an operator-set `max_output_tokens` for
+ * `opencode-go/qwen3.7-plus`, written through the dashboard's
+ * `PUT /api/provider-models`), `fitThinkingToMaxTokens()` resolved the cap by
+ * model name alone. The provider-scoped override was invisible to that lookup,
+ * so the cap resolved to `null` and the synthesized `max_tokens` (caller's
+ * `max_tokens` + the `reasoning_effort` thinking budget) went out unbounded —
+ * `64000 + 131072 = 195072`, which the upstream provider rejected with a bare
+ * `400 {"model":"qwen3.7-plus"}`.
+ *
+ * Root cause (confirmed by reading `fitThinkingToMaxTokens` and its caller in
+ * `openai-to-claude.ts`): the caller already has `routedProvider` in scope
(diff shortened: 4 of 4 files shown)


diegosouzapw:
Merged into `release/v3.8.50` — thank you @jmche for the contribution! It was validated on a combined merge-train (static gates + affected tests + vitest) together with 30 sibling PRs before landing.

PR #9585, opened 2026-08-06
feat(providers): add Zylo UnoRouter and Poolside registries

diegosouzapw:
## Summary

Implements the changes tracked in `feat/free-tier-providers-wave1-a`.

**What was done:**
```
1ccf9aa597 feat(providers): add Zylo UnoRouter and Poolside registries
```

 4 files changed, 88 insertions(+)

PR #10248, opened 2026-08-13
fix(models): make custom model overrides work consistently

### LINKED ISSUE #10242: fix(models): compatible-provider context override is ignored by /v1/models
### OmniRoute Version

3.8.50 (`release/v3.8.50`)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Provider(s) Involved

An operator-defined OpenAI-compatible provider node with:

- internal provider ID: `openai-compatible-chat-<uuid>`
- public prefix: `wawapi-openai`
- model: `grok-4.6`

### Description

It is currently not possible to make `/v1/models` report a manual context-window override for a model on a custom compatible-provider node.

The override can be saved from either surface:

1. Provider page → Custom Models → Context Window Override
2. Settings → AI → Model Overrides → `context_length`

Both correctly persist the override in `model_context_overrides` under the internal provider-node ID. Runtime capability resolution can use that record, but `/v1/models` publishes the model under the node's public prefix and then tries to enrich it using that prefix. Dynamic provider prefixes are not resolved back to their internal node IDs in this path, so the override is missed and the response remains at the generic/synced 128K value.

### Reproduction

1. Create an OpenAI-compatible provider node with public prefix `wawapi-openai`.
2. Make `grok-4.6` available through that node.
3. Set its Context Window Override to `500000` from Custom Models or set `context_length=500000` from Model Overrides.
4. Confirm the saved override through the provider/model API or database.
5. Request `GET /v1/models` and inspect `wawapi-openai/grok-4.6`.

### Expected Behavior

```json
{
  "id": "wawapi-openai/grok-4.6",
  "context_length": 500000
}
```

The value should be identical whether the model row came from custom models or synced available models.

### Actual Behavior

The override is persisted under:

```text
openai-compatible-chat-<uuid> / grok-4.6 = 500000 (manual)
```

but `/v1/models` returns:

```json
{
  "id": "wawapi-openai/grok-4.6",
  "owned_by": "wawapi-openai",
  "root": "grok-4.6",
  "context_length": 128000
}
```

This is not stale catalog caching: supported writes bump the model-catalog cache version and force a rebuild.

### Root Cause

- `src/app/api/v1/models/catalog.ts` maps a compatible provider's internal node ID to its public prefix and emits `owned_by` using the public identity.
- `src/app/api/v1/models/catalogResponse.ts` calls `enrichCatalogModelEntry()` for that published row.
- `src/lib/modelMetadataRegistry.ts` derives the enrichment provider from `entry.owned_by` and asks for the persisted context override using the public prefix.
- `src/lib/modelCapabilities.ts::resolveCapabilityInput()` resolves static aliases only; it cannot reverse an arbitrary DB-backed node prefix such as `wawapi-openai` to `openai-compatible-chat-<uuid>`.
- `src/lib/db/modelContextOverrides.ts` intentionally performs an exact `(provider, model_id)` lookup, so the public-prefix lookup cannot match the internal-ID record.
- The catalog then retains an inline 128K value when present or reaches `getTokenLimit()`'s generic 128K fallback when absent.

`max_input_tokens` is a separate capability and is not expected to change `context_length`; this reproduction uses the actual `context_length` / Context Window Override field.

#### Clear All Models does not remain cleared

The provider page's **Clear All Models** action reports success and its API deletes custom/synced rows, but the same connection's `syncedAvailableModels` are automatically recreated shortly afterward without the user pressing Sync. In the reproduction, `grok-4.5` and `grok-4.6` reappeared with `source: imported` after service restart/background initialization.

Expected: after Clear All Models, models remain cleared until the operator explicitly imports/syncs them again, or the UI clearly states that automatic discovery will immediately repopulate them.

#### Custom-model creation did not leave a custom row

After adding `grok-4.6` through Custom Models and saving a context override, the override existed in `model_context_overrides`, but `customModels` remained empty while the model existed only in `syncedAvailableModels`. This makes the UI appear to have created a custom model while the persisted source remains synced/imported.

This may share the import/save-response class described in #9128, but this reproduction is on a DB-backed OpenAI-compatible node and includes the context override/catalog mismatch above.

### Related Issues / Prior Work

- #4125 / PR #6822 added manual custom-model context-window overrides, but its tests cover API round-trip and runtime `getModelContextLimit()`, not `/v1/models` projection through a dynamic provider prefix.
- #9557 fixed Model Overrides exposing compatible-provider UUIDs and already identified the public-prefix/internal-ID boundary, but it did not fix context enrichment in `/v1/models`.
- #3202 was a different 128K catalog bug for OpenRouter upstream metadata and is already closed.
- #9128 may overlap the custom-model persistence symptom, but not the confirmed context override identity mismatch.

No existing issue found that covers this exact failure.

### Suggested Regression Test

Seed:

- provider node ID `openai-compatible-chat-<uuid>`
- prefix `wawapi-openai`
- active connection under the internal node ID
- `grok-4.6` as a custom or synced model
- `setModelContextOverride(internalNodeId, "grok-4.6", 500000)`

Then call `getUnifiedModelsResponse()` and assert:

```text
wawapi-openai/grok-4.6.context_length === 500000
```

A second test should cover a synced row because the production reproduction currently uses `syncedAvailableModels`.

### Acceptance Criteria

- `/v1/models` applies exact context overrides stored under a dynamic compatible-provider node's internal ID while preserving the public prefix in the response.
- The behavior works for both custom and synced model rows.
- Context override writes invalidate the catalog and the next response reports the new value.
- Clear All Models has defined, observable semantics and does not silently repopulate models immediately unless automatic sync is explicitly enabled and communicated.
- Adding a model through Custom Models either persists a custom row or reports a failed/no-op save instead of presenting a synced row as the created custom model.

jackjinke:
## Summary

- fix custom model overrides being ignored when a synced or auto-discovered model has the same provider and model ID
- resolve compatible-provider context overrides through the internal provider node while preserving the configured public prefix in model responses
- remove synced-model deletion tombstones; deletion now removes the row, while Hide/Show remains the persistent visibility control
- consistently merge models as synced metadata plus explicitly stored custom metadata, with independent context and capability overrides applied last

Fixes #10242.

## Behavior

- custom model overrides now take effect across `/v1/models`, provider model listings, combo builder options, Gemini-compatible listings, pricing, and the provider dashboard
- custom names, formats, endpoint metadata, token metadata, and explicit capability values such as `supportsVision: false` override same-ID discovered metadata
- missing custom fields preserve the corresponding discovered metadata
- compatible-provider context overrides stored under the provider node now appear in `/v1/models`
- persisted context and capability override records remain final authority
- deleting a synced model no longer creates an unrecoverable tombstone; a later provider sync may restore it

## Manual verification

Verified after a local deployment:

- started with a synced model, created a same-ID custom model, then deleted the custom model; after triggering sync from the Models endpoint, the previously unrecoverable synced model appeared again and was shown as hidden
- changed a compatible-provider context override and confirmed the effective context value was applied

## Tests

Changed test files:

- `tests/unit/model-token-limit-catalog.test.ts`
- `tests/unit/cursor-exclusive-listing-merge.test.ts`
- `tests/unit/provider-models-custom-merge-6247.test.ts`
- `tests/unit/combo-builder-options-route.test.ts`
- `tests/unit/synced-model-delete-custom-sibling.test.ts`
- `tests/unit/synced-model-delete-resync.test.ts`
- `tests/unit/synced-model-hide-persist-3782.test.ts`
- `tests/unit/db-synced-model-catalog-invalidation-8728.test.ts`
- `tests/unit/db-models-split.test.ts`
- `tests/unit/provider-limits-ui.test.ts`
- `src/app/(dashboard)/dashboard/providers/[id]/__tests__/phase1e.test.tsx`

Commands run:

```text
node --import tsx --test tests/unit/cursor-exclusive-listing-merge.test.ts tests/unit/provider-models-custom-merge-6247.test.ts tests/unit/model-token-limit-catalog.test.ts tests/unit/model-overrides-provider-prefix-9557.test.ts tests/unit/8327-models-owned-by-prefix.test.ts tests/unit/custom-vision-override-combo-routing-9195.test.ts tests/unit/combo-builder-options-route.test.ts
# 35 passed, 0 failed

node --import tsx --test tests/unit/model-token-limit-catalog.test.ts tests/unit/combo-builder-options-route.test.ts
# 8 passed, 0 failed

npm run test:vitest
# 39 files passed, 355 tests passed
```

Additional broad-suite status:

- UI Vitest: 247 files passed; 5 unrelated files failed on existing local-storage/UI fixture assumptions
- unit and integration aggregate runs were stopped after their isolated runners stopped making progress and accumulated unreaped child processes
- ecosystem and Playwright runners did not reach test execution because their server readiness checks timed out in the isolated environment

## Coverage

`npm run test:coverage` was not completed. The aggregate runners above did not finish reliably enough to produce a valid coverage report, so the repository's 60% coverage gate remains to be confirmed by CI before this PR is ready for review.


diegosouzapw:
Clean fix matching #10242 — custom model overrides now work consistently. 35 unit tests + vitest pass; CI failures are all inherited base-red (poolside/free-tier test from #9585). Good to merge.

diegosouzapw:
Merged into `release/v3.8.50` — thank you @jackjinke! Validated on a combined merge-train (static gates + every affected test + vitest) alongside 9 sibling PRs before landing.

PR #10250, opened 2026-08-13
feat(i18n): complete European Portuguese (PT-PT) translation

DarkEsteves:
## 🇵🇹 Tradução PT-PT completa


diegosouzapw:
Thanks for the PT-PT translation! One runtime i18n bug to fix: `resilienceConnections.degraded` changed from a string with `{sources}` interpolation to a nested object, which breaks `t('degraded', {sources})` in the resilience UI. Please restore the string shape (matching the other locales) — then it's good to merge.

PR #10251, opened 2026-08-13
fix: prevent concurrent combo requests from sharing responses (#concu…

ubaidurrehman-developer:
fix: prevent concurrent combo requests from sharing responses (#concurrent-combo-isolation)

Two independent root causes allowed concurrent requests targeting the same combo to
occasionally receive each other's responses:

1. requestDedup (PRIMARY): the shared in-flight Promise map coalesced concurrent combo
   calls with matching hashes, delivering the first caller's upstream response to other
   callers with different payloads. Fix: pass allowDedup=!isCombo so dedup is never
   applied to combo-dispatched handleChatCore calls. Direct single-model calls are
   unchanged.

2. semanticCache (DEFENSE-IN-DEPTH): two callers with different source bodies that
   translated to the same provider body shared a cache entry. Fix: generateSignature()
   gains an optional callerBodyHash parameter; when isCombo, chatCore computes
   computeRequestHash(body) on the original un-translated body and folds it into every
   cache signature (read + both write paths), giving each caller an independent cache
   namespace.

Also adds a DEDUP_COMBO_SKIPPED debug log when dedup is suppressed for a combo call,
and a new regression test suite:
- concurrent combo requests with different prompts each receive their own response
- requestDedup inflight map stays empty after combo requests complete
- N concurrent combo requests each receive distinct responses (stress variant)

Tests: 3/3 new pass, 6/6 existing combo-isolation pass, typecheck clean

## Summary

Fixes an intermittent bug where concurrent requests to the same combo occasionally
received each other's responses. Two callers sending different prompts to the same combo
at the same time could get the wrong response back with no error.

Primary fix: requestDedup is now disabled for combo-dispatched handleChatCore calls.
Defense-in-depth: the semantic cache signature for combo calls now includes a hash of
the original un-translated caller body, preventing two concurrent callers from sharing
a cache entry even if their payloads translate to the same provider body.

## Related Issues

- Closes #concurrent-combo-isolation

## Validation

- [x] Change type: routing
- [x] Focused tests: combo-concurrent-response-isolation.test.ts,
      combo-attempt-body-isolation-7847.test.ts
- [x] npm run lint
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented
      below

## Tests Added Or Updated

New:
- tests/unit/combo-concurrent-response-isolation.test.ts (3 tests)
  - Two concurrent requests with different prompts each receive their own response
  - requestDedup inflight map stays empty after combo requests complete
  - N=5 concurrent combo requests all receive distinct responses (stress variant)

Existing (unchanged, still passing):
- tests/unit/combo-attempt-body-isolation-7847.test.ts — 6/6 pass

## Coverage Notes

Changes touch open-sse/handlers/chatCore.ts, open-sse/handlers/chatCore/semanticCache.ts,
open-sse/handlers/chatCore/semanticCacheStore.ts,
open-sse/handlers/chatCore/streamingSemanticCacheStore.ts, and src/lib/semanticCache.ts.

The new test file directly exercises the fixed code paths via handleComboChat with an
echo-stub handleSingleModel. A hash collision would cause the Echo: Prompt-A /
Echo: Prompt-B assertions to fail, so the isolation invariant is directly verified.

Coverage is expected to increase slightly in the combo routing area.

## Reviewer Notes

- Fix 1 (!isCombo for allowDedup) is the minimal surgical fix. The only behavior change
  for combo calls is that concurrent requests with matching hashes will no longer
  coalesce — each gets its own upstream fetch. This is intentional: combo target
  invocations are independent per-caller executions.

- Fix 2 (callerBodyHash in semantic cache) only activates when isCombo === true.
  Non-combo single-model cache behavior is 100% unchanged.

- No migrations, no feature flags, no schema changes.

- Risk: Fix 1 will slightly increase upstream call volume for combo requests at
  temperature=0 where two concurrent callers previously shared an in-flight result.
  This is the correct tradeoff — correctness over call coalescing.

- The DEDUP_COMBO_SKIPPED debug log is emitted at debug level only — no production
  log noise.


ubaidurrehman-developer:
Consider it done.

PR #10256, opened 2026-08-13
[TS7] fix(types): normalize executor result contracts

backryun:
## Summary
- validate unknown executor outputs before consuming their response envelope
- normalize the 401 credential-refresh retry through the same executor-result contract
- declare the internal execution credential and semaphore metadata contract without changing runtime response shapes

## TS7 impact
- TypeScript 6: 28 -> 16 diagnostics (12 removed, 0 added)
- TypeScript 7.0.2: 28 -> 16 diagnostics (12 removed, 0 added)

## Validation
- node --import tsx/esm --test tests/unit/chatcore-upstream-timeouts.test.ts (5/5)
- GitHub credential refresh integration case (1/1)
- node --import tsx/esm --test tests/unit/chatcore-execution-credentials.test.ts (15/15)
- targeted ESLint
- npm run check:file-size
- git diff --check

### DIFF
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -387,6 +387,12 @@ import {
   isRpmExhausted,
 } from "../services/geminiRateLimitTracker.ts";
 import { isSmallEnoughForSemanticCache } from "../utils/estimateSize.ts";
+
+type ChatCoreExecutorResult = ReturnType<typeof normalizeExecutorResult> & {
+  _executionCredentials?: Record<string, unknown>;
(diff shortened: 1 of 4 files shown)


diegosouzapw:
LGTM — type hardening. The `normalizeExecutorResult` strict guard and `ChatCoreExecutorResult` metadata type match what runtime already attached/read. `instanceof Response` is stricter than duck-typing but no executor in-tree returns a non-global-Response. Tests prove malformed output throws. Ready to merge.

diegosouzapw:

PR #10257, opened 2026-08-13
[TS7] fix(types): narrow refresh token rotation inputs

backryun:
## Summary
- align the refresh-token rotation guard with the unknown values returned by the DB boundary
- preserve the existing fail-closed runtime checks for non-string token values
- cover malformed token values in the existing serializer test

## TS7 impact
- TypeScript 6: 28 -> 27 diagnostics (1 removed, 0 added)
- TypeScript 7.0.2: 28 -> 27 diagnostics (1 removed, 0 added)

## Validation
- node --import tsx/esm --test tests/unit/refresh-serializer.test.ts (6/6)
- targeted ESLint
- npm run check:file-size
- git diff --check

### DIFF
--- a/open-sse/services/refreshSerializer.ts
+++ b/open-sse/services/refreshSerializer.ts
@@ -112,8 +112,8 @@ export async function serializeRefresh<T>(provider: string, fn: () => Promise<T>
  * and codex-lb's replica race-detection.
  */
 export function wasRefreshTokenRotated(
-  attemptedRefreshToken: string | null | undefined,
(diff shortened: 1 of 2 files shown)


diegosouzapw:
LGTM — pure type widening. The runtime body is unchanged (still requires both non-empty strings + inequality), so fail-closed behavior for non-string token values from the DB boundary is preserved exactly. The two added assertions pin that down. Ready to merge.

diegosouzapw:

PR #10258, opened 2026-08-13
[TS7] fix(types): validate nonstreaming JSON contracts

backryun:
## Summary
- require successful non-streaming upstream JSON payloads to have an object root
- preserve the object contract through response translation, sanitization, Cline unwrapping, guardrails, and response assembly
- normalize the usage record before cost calculation instead of passing an arbitrary object

## TS7 impact
- TypeScript 6: 28 -> 17 diagnostics (11 removed, 0 added)
- TypeScript 7.0.2: 28 -> 17 diagnostics (11 removed, 0 added)

## Validation
- chatcore non-streaming parser tests (10/10)
- ClinePass provider tests (21/21)
- Cline response-envelope tests (2/2)
- response sanitizer tests (50/50)
- full chatCore translation-path integration suite (69/69)
- targeted ESLint
- npm run check:file-size
- git diff --check

### DIFF
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -104,7 +104,13 @@ import { resolveAgentGoalPolicy } from "../utils/agentGoalPolicy.ts";
 import { createStreamController } from "../utils/streamHandler.ts";
 import * as streamFailure from "../utils/streamFailureFinalization.ts";
 import { createSseHeartbeatTransform, shapeForClientFormat } from "../utils/sseHeartbeat.ts";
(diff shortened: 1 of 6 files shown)


backryun:
@diegosouzapw This is Final commit for TS7 fix.
Once all the commits posted this time have been applied, I will remove the existing TS6 compilation and replace it with the TS7 compiler to complete the full migration.
I would appreciate it if you could review this first, and I would like to thank you for always providing such helpful responses.

diegosouzapw:
LGTM — type hardening only. The `isJsonRecord` guard on non-streaming JSON roots and the clinepass envelope is a behavior tightening but only on the error path. `normalizeUsage` keeps every field the translators emit, so `estimatedCost` is unchanged. All claimed suites pass. Nothing to change.

diegosouzapw:

PR #10259, opened 2026-08-13
fix(sse): extract perplexity-web answers from workflow_block

jeyhunfaslanov:
⚠️ base-red inherited: #9985

## Problem

Perplexity moved the answer text out of `markdown_block` into `workflow_block`
(`intended_usage: "workflow_root"`), streaming it as RFC-6902 patches whose `field` is
`"workflow_block"` and whose paths address
`/steps/<n>/items/<m>/payload/text_payload/chunks/<k>`.

`extractContent` in `open-sse/executors/perplexity-web/protocol.ts` recognised neither shape.
Two independent guards dropped every answer frame:

- `isAnswerTextUsage("workflow_root")` is false, so the block loop `continue`d before any
  accumulation.
- the diff guard skipped every patch whose `field !== "markdown_block"`.

The stream therefore ran to `COMPLETED` with an empty accumulator and the executor surfaced
`Provider returned empty content` (502) even though the upstream SSE carried the full answer.
Every model was affected — the carrying block is model-independent — so the provider was
unusable.

## Fix

- Adds `workflow_block` to `PplxBlock`.
- `applyWorkflowDiff` — patch applier for the streaming path.
- `applyWorkflowBlock` — handles a materialized block on the terminal frame.
- Answer tracks are keyed per step+item so concurrent items cannot overwrite each other's
  chunk indices.
- Only `variant: "answer"` payloads are accumulated — search queries, sources and "thinking"
  items stay out of the message.
- Legacy `markdown_block` diffs keep working (covered by a regression test).

## Validation (TDD — Hard Rule #18)

`tests/unit/perplexity-web-workflow-block.test.ts` (new, 265 lines) — written failing against
the unpatched extractor, passing after the fix:

```
node --import tsx/esm --test tests/unit/perplexity-web-workflow-block.test.ts

✔ extractContent reconstructs the answer from workflow_block diff patches
✔ extractContent reads a materialized workflow_block on the COMPLETED frame
✔ extractContent ignores non-answer workflow items
✔ extractContent still handles legacy markdown_block diffs
ℹ tests 4
ℹ pass 4
ℹ fail 0
```

Fixtures are trimmed from a live capture (`pplx-auto`, `mode=copilot`). Replaying the full
96 KB capture through the patched extractor yields the complete 247-char answer over 7
incremental deltas, against an empty string before the fix.

## Files

| File | Change |
| --- | --- |
| `open-sse/executors/perplexity-web/protocol.ts` | +170 |
| `tests/unit/perplexity-web-workflow-block.test.ts` | +265 (new) |

### DIFF
--- a/open-sse/executors/perplexity-web/protocol.ts
+++ b/open-sse/executors/perplexity-web/protocol.ts
@@ -148,6 +148,36 @@ export interface PplxBlock {
     }>;
     goals?: Array<{ description?: string }>;
   };
+  // Workflow API (`intended_usage: "workflow_root"`). Perplexity moved the answer
+  // text here from markdown_block: it now arrives as one WORKFLOW_ITEM_TEXT item
+  // whose `text_payload.variant` is "answer", nested under a workflow step. Other
+  // variants ("thinking") and item types (queries, sources) are not answer text.
+  workflow_block?: PplxWorkflowBlock;
+}
+
+export interface PplxWorkflowTextPayload {
+  text?: string;
+  chunks?: string[];
+  variant?: string;
+  is_streaming?: boolean;
+}
+
+export interface PplxWorkflowItem {
+  type?: string;
+  variant?: string;
+  payload?: { text_payload?: PplxWorkflowTextPayload };
+}
+
+export interface PplxWorkflowStep {
+  status?: string;
+  title?: string;
+  tool_name?: string;
+  items?: PplxWorkflowItem[];
+}
+
+export interface PplxWorkflowBlock {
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Nice, surgical fix — Perplexity moved the answer text to `workflow_block` (Workflow API), and this extracts the `text_payload.variant == "answer"` items with the same chunks-over-text precedence as `markdown_block`, keyed per step+item so concurrent items never collide on `/chunks/<k>` indices. Test included. Good to merge.

PR #10260, opened 2026-08-13
fix(ci): clear base-reds on release/v3.8.50 (round 4)

diegosouzapw:
Drains the HARD failures reported by Release-Green run 31693210948 on issue #9985 (ESLint errors: 2) plus the merge-integrity red every open PR is inheriting.

## Fixes

1. **ESLint error 1** — @omniroute/opencode-plugin/src/index.ts had a stray extra `});` (introduced by #9316) that broke parsing. Removed it.
2. **ESLint error 2** — cli-env-inline-comment-10100.test.ts used `new Function` to extract parseEnvValue from the bin entrypoint (no-new-func, Hard Rule #3). Extracted the helper to `bin/cli/utils/parseEnvValue.mjs` and import it from both the entrypoint and the test (same behavior, no eval).
3. **open-sse-typecheck (Fast Quality Gates)** — open-sse/utils/stream.ts imported `sseCommentsEnabled` twice (#9378) causing TS2300 Duplicate identifier. Removed the duplicate import.
4. **Merge integrity (changelog + generated skills)** — skills/omni-settings/SKILL.md was edited manually by #10169 without updating the generator source, so check:agent-skills-sync failed on every PR (Generated: 1). Moved the curated thinking-budget content into a `<!-- skill:custom-start -->` block (the documented preservation mechanism), which the generator now keeps in sync.

All reproduced on the pure tip (`6143da70d1`) before fixing.

Refs #9985

PR #10252, opened 2026-08-13
[TS7] fix(types): remove duplicate SSE comment import

backryun:
## Summary

- remove the duplicate `sseCommentsEnabled` import introduced while the original import remained in `open-sse/utils/stream.ts`
- clear both TS2300 diagnostics without changing runtime behavior

## Root cause

The same binding was imported twice from `sseHeartbeat.ts`, so both TypeScript 6 and the official TypeScript 7.0.2 compiler reported two duplicate-identifier diagnostics.

## Validation

- full `open-sse/tsconfig.json`: TS6 28 → 26, added 0
- full `open-sse/tsconfig.json`: TypeScript 7.0.2 28 → 26, added 0
- `node --import tsx/esm --test tests/unit/sse-comments-optout-9305.test.ts tests/unit/sseHeartbeat.test.ts` (10/10)
- `npm exec eslint -- open-sse/utils/stream.ts`
- `npm run check:file-size`

Refs #8484

PR #10262, opened 2026-08-13
feat(responses): virtualize previous_response_id continuation regardless of upstream support

hartmark:
## Summary

- OmniRoute now exposes OpenAI-compatible `previous_response_id`/`store` continuation to clients unconditionally, even when the selected upstream provider has no native Responses-API state support. Reconstruction happens server-side in `handleChatImplementation`, before any downstream validation or provider translation: OmniRoute resolves the response id back to the full input/output it previously produced, prepends it to the client's delta, and forwards the full reconstructed history upstream exactly as it does today. Client<->OmniRoute traffic shrinks to the new delta only; OmniRoute<->provider traffic is unchanged.
- Storage reuses the existing call-log pipeline artifact (already gated by `call_log_pipeline_enabled`, already retained/cleaned up by the existing call-log lifecycle) instead of duplicating conversation content into a second store — only a lightweight `call_logs.response_id` index is new.
- Every lookup is scoped by `api_key_id` so one client can never resolve another client's stored conversation, and any unresolvable/missing/size-limit-omitted state fails closed with OpenAI's own `previous_response_not_found` contract.

Originally stacked on `feat/openai-responses-store-toggle`, which has since merged as #10121 — this PR is now rebased directly onto current `release/v3.8.50` and stands alone.

## Related Issues

- N/A

## Validation

- [x] Focused tests: `node --import tsx --test tests/unit/responses-continuation-store.test.ts` (6 pass)
- [x] `npm run lint` — clean
- [x] Production-code change includes new automated tests in this PR

## Tests Added Or Updated

- `tests/unit/responses-continuation-store.test.ts` (new, 6 tests: reconstruction from artifact, unknown response id, cross-tenant isolation, missing artifact on disk, size-limit-omitted payload fails closed, never-captured detail logging)

## Coverage Notes

- Touches `src/lib/db/responsesContinuationStore.ts` (new), `src/sse/handlers/chat.ts`, `src/lib/usage/callLogs.ts`, `open-sse/handlers/chatCore/attemptLogging.ts` — all covered by the new test file plus existing chat-handler test coverage for the surrounding request path.

## Reviewer Notes

- Migration `153_call_logs_response_id.sql` (`ALTER TABLE call_logs ADD COLUMN response_id TEXT DEFAULT NULL` + index) is purely additive, no backfill needed.
- Fails closed rather than silently degrading: an unresolvable `previous_response_id` returns OpenAI's own `previous_response_not_found` error rather than proceeding without the prior context.


hartmark:
CI here is failing on pre-existing base-red on `release/v3.8.50`, not this PR — reproduced identically on a pristine checkout of the base tip. For correlated tracking:

- `mutation-test-coverage` (Fast Quality Gates): untracked, fixed in #10264
- `open-sse-typecheck` (stream.ts TS2300 duplicate identifier): tracked/fixed in #10252 and #10260
- `dead-code` ratchet regression: tracked in #9985 (per that issue, ratchet drift is expected mid-cycle / rebaselined at release, not a contributor concern)
- the poolside live-catalog test failure in Unit Tests fast-path: fixed in #10260

diegosouzapw:
Thanks for this — reusing the call-log artifact instead of a second store is the right call, and the fail-closed/tenant-scoped design in `responsesContinuationStore.ts` is solid, with good isolated test coverage for it (reconstruction, unknown id, cross-tenant, missing artifact, size-limit-omitted, no-detail-logging).

Two things I'd like resolved before merge:

1. **CI is currently red on `check:dead-code`** (410 symbols vs baseline 409) — this isn't base-red, it traces directly to the new `ResponsesContinuationState` export in `responsesContinuationStore.ts`, which isn't consumed anywhere outside that file. Easiest fix is to stop exporting it (or give it a real consumer) and re-run `node scripts/check/check-dead-code.mjs`.

2. **The new interception in `chat.ts` (around the `previous_response_id` block) runs before any connection/executor is selected, and it unconditionally deletes `previous_response_id` from the body.** That looks like it collides with two things already in the codebase:
   - `codex.ts` explicitly delegates history resolution to the upstream when `previous_response_id` is present (`"A previous_response_id delegates history resolution to the upstream"`, early-return at line 235) — Codex's native continuation path.
   - The `responsesPreviousResponseIdMode=preserve` setting (exposed in the dashboard's Responses State Policy tab), applied per-target later in `chatCore.ts` via `applyResponsesPreviousResponseIdPolicy` — that call becomes a no-op for these requests since the field is already gone by then.

   As far as I can tell from reading the code, both become unreachable for continuation requests once this lands — an operator who set `preserve`, or any Codex-bound request, would go through OmniRoute's own reconstruction/fail-closed path instead of the upstream-native one it used to take. I couldn't fully verify the runtime impact locally (it needs a live Codex/ChatGPT-web OAuth connection), so it's possible this is intentional or the impact is smaller than it looks from the source — but it'd be good to either scope the new logic to skip Codex/`preserve` connections, or confirm explicitly that this replacement is the intended new behavior for those cases too, with a test either way.

Happy to take another look once those are addressed — the core mechanism (fail-closed reconstruction from the call-log artifact) looks good.

diegosouzapw:
### Babysit summary

The failing checks on this PR were **STALE** — the branch was 60 commits behind `release/v3.8.50`'s tip (`774127be3`). After a refresh-merge and two real fixes, all non-advisory checks are **green**.

**What was done:**

1. **Refresh-merge** `origin/release/v3.8.50` into the PR branch (clean merge — no conflicts in `open-sse/transformer/`).
2. **fix(db): renumber `previous_response_id` index migration 153 → 154** (`ca37b7d8`). The stale refresh surfaced a REAL migration-version collision: base tip already carries `153_radar_local_model_state.sql`, and this PR added `153_call_logs_response_id.sql`. The migration runner's collision guard (`src/lib/db/migrationRunner.ts`) throws on two live `.sql` files sharing a numeric prefix — DB startup would have failed. Renumbered to the next free slot (154).
3. **docs(db): sync migration count to 149 across llm.txt mirrors** (`22e6ba49`). The new migration bumped the real count 148→149, tripping both `Docs Gates (fast-path)` and the `check-docs-counts-sync.test.ts` unit test. Updated README/AGENTS/llm.txt + regenerated the 42 i18n `llm.txt` mirrors.

**Verified locally:** `npm run check:docs-all` passes; the previously-failing `tests/unit/check-docs-counts-sync.test.ts` (24/24) and `tests/unit/responses-continuation-store.test.ts` (6/6) and `tests/unit/db-migration-runner.test.ts` (26/26) all pass.

**Gate:** all non-advisory checks pass on head `22e6ba49` — Docs Gates, Fast Quality Gates, Unit Tests fast-path 1-4/4, Vitest, No new ESLint warnings, dast-smoke, semgrep, Change Classification, Merge integrity. `Build (advisory)` is advisory and ignored.

**Remaining:** no unresolved review threads. Ready for human review & merge (not auto-merged).

hartmark:
Addressed both items:

1. **`check:dead-code` regression** — traced to `ResponsesContinuationState` exactly as you flagged; it's only ever used as a local return-type annotation in the same file. Dropped the `export`. Confirmed: `node scripts/check/check-dead-code.mjs` now reports 415, matching current baseline (no regression).

2. **The `previous_response_id` interception vs. `preserve` mode / Codex** — you were right to flag this, and read closely it goes a bit further than just Codex: the interception in `chat.ts` ran unconditionally, before target/connection selection, ahead of `applyResponsesPreviousResponseIdPolicy` in `chatCore.ts` (the existing per-target enforcement point for `responsesPreviousResponseIdMode`). By the time that policy call ran, `previous_response_id` was already gone — so `mode=preserve` (the explicit, connection-independent "let the upstream resolve this natively" contract) was silently unreachable for any Responses-API-source request, and Codex's own `stripOrphanedCodexFunctionCallOutputs` early-return (which depends on seeing an untouched `previous_response_id`) never fired either.

   Fix: the interception now reads `responsesPreviousResponseIdMode` early via the existing `getCachedSettings()` cache and skips entirely when the mode is `preserve` — the field flows through untouched to the existing downstream policy exactly as before this feature landed. `auto` and `strip` modes are untouched by this change; virtualization stays a strict improvement over their old "drop the field, hope the client resent everything" behavior, since neither of those modes ever promised connection-native pass-through.

   I didn't have a live Codex/ChatGPT-web OAuth connection to test end-to-end (same constraint you called out), but added a regression test that exercises the real `chat.ts` handler through the harness (not just the policy helper in isolation): with `mode=preserve` and an id OmniRoute's own store has never seen, the request now falls through to ordinary model routing (404 unknown model) instead of virtualization's `previous_response_not_found` 400 — proving the field survives untouched past the interception point. I confirmed this test fails for the right reason against the pre-fix code.

Verified: `npm run typecheck:core` clean, `node scripts/check/check-dead-code.mjs` → 415 (baseline), all 60 related tests pass (`responses-continuation-store`, `responses-state-policy`, `chat-context-relay`, `executor-codex`, `codex-orphaned-tool-outputs-2928`, plus the 2 new tests), pre-commit gates (`t11:any-budget`, `tracked-artifacts`, `docs-sync`) pass.

Pushed: `36fb477a8`

diegosouzapw:
Thank you @hartmark for the Responses continuation fix. The exact head passed the complete PR checks and the combined FAST train; no PR-specific red was observed. The release merge is being serialized after the prior verified base update.

PR #10263, opened 2026-08-13
feat(dashboard): agentic conversation tracking — v4, decoupled + storage-architecture concern resolved

hartmark:
## Summary

This is the 4th iteration of agentic conversation tracking (v1 = #9439, v2 = #10047, v3 = `hartmark/OmniRoute#4`), addressing the original deferral directly:

> "it introduces a large new subsystem (agentic conversation tracking, +9639 lines) that would benefit from a design RFC/discussion first to align with the project's architecture before merging piecemeal" — @diegosouzapw on #9439

**What changed since that deferral:**

1. **Size roughly halved**: +9639 lines then → **+4681/-232 across 83 files now** (production-only, excluding tests: +3225/-216 across 71 files).
2. **The architectural concern is resolved, not worked around.** The original version stored full turn content (text, tool-call args, tool results) in its own table — a second, parallel content store duplicating what `call_logs` already persists. This version's `conversation_turn_nodes` table is **identity-only**: just the hash-chain id, role, and a content hash for reconnect-anchor lookups. Actual turn content (text/tool-call/tool-result) is resolved **on demand** from the existing call-log pipeline artifact via `last_correlation_id` — the same full, untruncated payloads `call_logs` already persists, already gated by `call_log_pipeline_enabled`, already covered by the existing retention/cleanup lifecycle. No second copy of conversation content exists anywhere. (See `open-sse/services/conversationTurnContent.ts` and the migration comments in `153_agentic_conversations.sql`/`154_conversation_turn_nodes.sql` for the full reasoning.)

**What remains, honestly stated — this is not purely UI wiring:**
- Two new tables (`agentic_conversations`: conversation-root record; `conversation_turn_nodes`: identity-only hash chain)
- One new service (`open-sse/services/conversationTracker.ts`): resolves a stable conversation id per request via fingerprint + bounded prefix-hash matching, with a strict-growth invariant so independent single-shot requests sharing identical opening content never get merged
- A thin content-resolution helper (`conversationTurnContent.ts`) and a couple of new dashboard API routes/pages

That's the full new backend surface — bounded, and it now follows the same storage pattern the rest of the codebase already uses (reuse the call-log artifact system) instead of inventing a new one, which is the specific thing the RFC concern was about.

**Feature recap** (unchanged in substance from the original description):
- `/dashboard/logs`: new toggleable Conversation column.
- Request detail panel: new Full Conversation transcript above the raw SSE event stream — Markdown rendering, per-turn timestamps, live auto-refresh building the transcript in real time from the in-flight SSE chunk buffer while a request is still streaming.
- New `/dashboard/conversations` page listing conversations with 2+ turns, no-forking model (an edited/duplicated mid-history turn mints its own independent conversation instead of merging), pagination, prev/next navigation.
- Continuation detection excludes the system message from the identity anchor, since real coding-agent CLIs commonly regenerate it every request with live context (timestamp, cwd, git status) — without this, that volatility alone broke every continuation check against real traffic (confirmed live on omniroute-dev: 28 consecutive requests from one real growing session each recorded as separate `turn_count=1` conversations before this fix).

Independent of and no longer stacked on `feat/openai-responses-store-toggle`/#10121 (already merged) or `feat/responses-previous-response-id-virtualization`/#10262 (separately submitted) — this PR stands alone against current `release/v3.8.50`.

## Related Issues

- Addresses the `[defer]` on #9439; supersedes #10047 and `hartmark/OmniRoute#4`.

## Validation

- [x] Focused tests: `agenticConversations`, `conversationTracker`, `conversationTurnContent`, `conversations-active-call-log-id`, `conversations-tree-route-seq-param`, `logs-detail-partial-reasoning-chunk-split`, `request-timeline-lane-allocation`, `chatcore-log-truncation`, `sidebar-monitoring-reorg`, `sidebar-visibility`, `check-migration-numbering` — 97/97 pass
- [x] `npm run lint` — clean (pre-existing baseline debt only, no new violations)
- [x] `check-migration-numbering.test.ts` — 0 gaps, 0 collisions (15/15)

## Tests Added Or Updated

- `tests/unit/agenticConversations.test.ts`, `conversationTracker.test.ts`, `conversationTurnContent.test.ts`, `conversations-active-call-log-id.test.ts`, `conversations-tree-route-seq-param.test.ts`, `request-timeline-lane-allocation.test.ts`, `logs-detail-partial-reasoning-chunk-split.test.ts` (all new)

## Coverage Notes

- New backend surface (`conversationTracker.ts`, `conversationTurnContent.ts`, `agenticConversations.ts`) is covered by dedicated unit tests exercising the real fingerprint/hash-chain/reconnect logic, not just mocked branches.

## Reviewer Notes

- Migrations `153_agentic_conversations.sql` / `154_conversation_turn_nodes.sql` are purely additive new tables, no backfill needed.
- Happy to split further if any specific piece (e.g. the timeline lane-reuse UI, or the tracker service itself) would review better as its own PR — flagging that option rather than assuming this exact shape is the only way to land it.


hartmark:
CI here is failing on pre-existing base-red on `release/v3.8.50`, not this PR — reproduced identically on a pristine checkout of the base tip. For correlated tracking:

- `mutation-test-coverage` (Fast Quality Gates): untracked, fixed in #10264
- `open-sse-typecheck` (stream.ts TS2300 duplicate identifier): tracked/fixed in #10252 and #10260
- `dead-code` ratchet regression: tracked in #9985 (per that issue, ratchet drift is expected mid-cycle / rebaselined at release, not a contributor concern)
- the poolside live-catalog test failure in Unit Tests fast-path: fixed in #10260

diegosouzapw:
Thanks for the thorough writeup, and for directly addressing the earlier RFC deferral rather than just resubmitting — the identity-only `conversation_turn_nodes` design (resolving display content from the existing call-log artifact instead of storing a second copy) is a real fix for the original architectural concern, and the comments documenting the real-traffic edge cases you hit (system-prompt regeneration, sliding context windows, cache-aware turn edits) make the reasoning easy to follow and verify.

One blocking issue found in review: `153_agentic_conversations.sql` and `153_call_logs_response_id.sql` both use version prefix `153`, which is already taken on the current `release/v3.8.50` tip by `153_radar_local_model_state.sql` (merged after this branch's fork point). This isn't cosmetic — `migrationRunner.ts`'s runtime collision guard throws on `getDbInstance()` once all three files are present together, which is exactly why all 4 `Unit Tests fast-path` shards, `Vitest (fast-path)`, and the `migration-numbering` gate are currently red on this PR (confirmed directly in the job logs: `Migration version collision detected: version=153 → [agentic_conversations, call_logs_response_id]`). Renumbering the two files (e.g. to 155/156) should clear all of that at once.

The `dead-code` gate failure (411 > baseline 409) looks like separate base drift rather than something this PR caused — release/v3.8.50's baseline was already rebaselined to 415 after this branch's fork point, but that value isn't in this branch's `quality-baseline.json` yet. A rebase onto the current tip should pick it up automatically.

Two smaller, non-blocking items:
- `resolveConversationId()` is called directly with no try/catch in the hot chat-completion path (`chat.ts:616`), and `withChatAdmission` rethrows uncaught errors rather than converting them into a sanitized response. Every other side-effect call nearby in that same function is defensively wrapped — worth doing the same here so a DB hiccup in this new, best-effort tracking feature can't turn an otherwise-working chat request into a hard failure.
- `getConversationTurnTree` (`agenticConversations.ts:230`) looks like a leftover export from the tree-vs-flat-list redesign — no callers remain outside its own file/tests.

Once the migration rename lands (and ideally the `resolveConversationId` guard), this looks like solid, mergeable work.

hartmark:
Addressed both non-blocking items from review:

1. **`resolveConversationId()` try/catch** (`chat.ts`) — wrapped the call so a DB hiccup in this best-effort tracking feature falls back to `conversationId = null` and logs a warning instead of turning a working chat request into a hard failure, matching the `try { ... } catch {}` pattern used by the other side-effect calls nearby in the same function (e.g. `recordRejectedRequestUsage`, `getComboByName`).

2. **`getConversationTurnTree`** — confirmed it: knip's `project`/`entry` config in `knip.json` excludes `tests/**` entirely, so an export used *only* by tests can never register as "used" no matter how many call sites exist — this wasn't a leftover, it was an intentional test-support export, but it was permanently unwinnable against the dead-code gate as written. Rather than keep a second whole-chain query path alive solely for that, I swapped its 8 test call sites (across `agenticConversations.test.ts` and `conversationTracker.test.ts`) to the paginated `getConversationTurnPage(id, { limit: 500 })` — already the dashboard's canonical query — and deleted the export. All 31 tests in both files still pass with identical assertions (they only ever read `.length`/`.map(contentHash)`, so the extra `seq` field and page wrapper are inert for them).

Also caught and fixed pre-existing `docs-sync` drift on this branch (42 `docs/i18n/*/llm.txt` mirrors were stale against root `llm.txt`) — unrelated to this PR's content, just blocking the pre-commit hook.

Verified: `npm run typecheck:core` clean, `node scripts/check/check-dead-code.mjs` → 415 (matches current baseline, no regression), both test files green (31/31), `t11:any-budget` / `tracked-artifacts` / `docs-sync` pre-commit gates all pass.

Pushed: `6e54ce1a1`

hartmark:
Rebased onto current `release/v3.8.50` (head `122611b0a`). Both review items addressed and verified directly in source:

- **Migration collision**: renumbered to `154_call_logs_response_id.sql` / `155_agentic_conversations.sql` / `156_conversation_turn_nodes.sql` — this landed cleanly since #10262 (this repo's own `154_call_logs_response_id.sql`, `responsesContinuationStore.ts`) is now merged upstream; the rebase folded the two into one canonical implementation instead of duplicating it.
- **`resolveConversationId()` guard**: wrapped in try/catch with a `log.warn` fallback (`chat.ts:656`) — a DB hiccup in this best-effort tracking no longer risks the chat request.
- **`getConversationTurnTree` dead export**: removed, no remaining callers.

Also found and fixed two CI failures while rebasing:
- `i18n-vi-completeness.test.ts`: `requestLogger.columns.conversation` (this PR's own new column) was missing from `fr.json`/`vi.json`.
- `check-docs-sync.mjs`: the `docs/i18n/*/llm.txt` mirrors still said "117 domain-specific files" after an earlier rebase updated the migration count but not this companion number — fixed across all 42 locales.

The remaining red (`chat-cooldown-aware-retry.test.ts`, "expected cooldown-aware retry wait, got 60ms") is unrelated — that test file's history has no overlap with this branch's changes, and the assertion shape (a hardcoded wait-time expectation) looks like CI-runner timing flakiness rather than a real regression.

@diegosouzapw ready for another look.

PR #10272, opened 2026-08-13
fix(providers): preserve connection test status codes

Zartharas:
## Summary

Preserve validator HTTP failure status metadata through the API-key / web-session connection-test path.

## Root cause

`testApiKeyConnection()` already passes `result.statusCode` into `classifyFailure()`, but the API-key connection-test result omitted that field.

Downstream consumers could therefore lose exact status metadata even when the validator knew the upstream response class.

This could cause:

- the connection-test response to expose `statusCode: null`;
- call-log handling to fall back to `401` for unrelated 403/429/5xx failures;
- downstream tooling to lose the distinction between authentication, rate-limit, and upstream failures.

## Fix

- Extract a small pure connection-test result builder.
- Preserve validator failure `statusCode`.
- Keep successful validation status-less.
- Keep status-less semantic failures status-less.
- Preserve warning, diagnosis, and deployment behavior.
- Add focused regression coverage for 401, 403, 429, 503, status-less failures, and success.
- Add the canonical `changelog.d` fragment for #10272.

## Safety

This does not broaden authentication classification or automatic recovery behavior.

It only preserves status metadata already produced by the validator. Unknown/status-less failures remain status-less.

This complements #10218, which is merged and narrowly classifies DeepSeek business code `40003` as auth status `401`.

## Validation

Final reconciliation was completed against the active `release/v3.8.50` tip before review:

- Focused regression: **3 passed, 0 failed**
- Scoped ESLint: **PASS**
- Route validation: **PASS**
- Core typecheck: **PASS**
- Explicit-any budget: **PASS**
- Cycle check: **PASS**
- File-size gate: **PASS**
- Docs-sync: **PASS**
- Changelog integrity: **PASS**
- Tracked-artifact check: **PASS**
- Normal Husky pre-commit hook: **PASS**
- `npm ci`: **0 vulnerabilities**
- Final branch: **1 commit** directly parented to the reconciled release tip

Repository-wide `npm run lint` is currently base-red from unrelated upstream files:

- `@omniroute/opencode-plugin/src/index.ts`
- `tests/unit/cli-env-inline-comment-10100.test.ts`

The changed files pass scoped ESLint, and the observed full-lint failure remains limited to those two unrelated upstream errors.

## Tests Added Or Updated

- `tests/unit/provider-test-statuscode-propagation.test.ts`

## Reviewer Notes

The connection-test route stays small; result normalization is isolated in a sibling pure helper.

No recovery policy, OAuth behavior, or database schema is changed.

### DIFF
--- a/changelog.d/fixes/10272-provider-test-statuscode-propagation.md
+++ b/changelog.d/fixes/10272-provider-test-statuscode-propagation.md
@@ -1,0 +1,1 @@
+- **fix(providers):** preserve validator HTTP status codes in API-key and web connection-test results so callers can distinguish authentication, rate-limit, and upstream failures ([#10272](https://github.com/diegosouzapw/OmniRoute/pull/10272)) — thanks @Zartharas
--- a/src/app/api/providers/[id]/test/apiKeyTestResult.ts
+++ b/src/app/api/providers/[id]/test/apiKeyTestResult.ts
@@ -1,0 +1,28 @@
+export interface ApiKeyValidationResult {
+  valid: boolean;
+  warning?: string | null;
+  statusCode?: number | null;
+  deployments?: unknown;
+}
+
+export interface ApiKeyTestDiagnosis {
+  type: string;
+  source: string;
+  message: string | null;
+  code: string | null;
+}
+
+export function buildApiKeyConnectionTestResult(
+  result: ApiKeyValidationResult,
+  error: string | null,
+  diagnosis: ApiKeyTestDiagnosis
+) {
+  return {
+    valid: !!result.valid,
+    error,
+    warning: result.warning || null,
+    statusCode: result.valid ? null : (result.statusCode ?? null),
+    diagnosis,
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Thanks for the clean, well-scoped fix — reviewed and this looks solid.

I verified the root cause directly against source: `toValidationErrorResult()` in `src/lib/providers/validation/transport.ts` already derives `statusCode` from the upstream response for the OpenAI-compatible and Anthropic-compatible provider families, but `testApiKeyConnection()` was dropping it before it reached the response. I also confirmed this was user-visible — `ProviderOnboardingWizard.tsx` only renders the `HTTP {statusCode}` badge when the value is a number, so it was silently never showing for API-key 403/429/503 failures before your fix.

Ran your new test (`tests/unit/provider-test-statuscode-propagation.test.ts`) directly against your branch — 3/3 pass. CI is fully green across the board. No changes requested; this is ready to merge as-is.

diegosouzapw:

Issue #7152, opened 2026-07-14
fix(db): sql.js WASM memory limit causes OOM on large databases


liuleyouxiang168-sketch:
### OmniRoute Version

v3.8.48

### Installation Method

npm (global)

### Operating System

Windows

### Description

sql.js 的 WASM 内存限制导致大数据库 OOM

建议在初始化时增加 WASM 内存上限，或者恢复 better-sqlite3 支持

### Steps to Reproduce

ql.js 的 WASM 内存限制导致大数据库 OOM

建议在初始化时增加 WASM 内存上限，或者恢复 better-sqlite3 支持

### Expected Behavior

ql.js 的 WASM 内存限制导致大数据库 OOM

建议在初始化时增加 WASM 内存上限，或者恢复 better-sqlite3 支持

### Actual Behavior

sql.js 的 WASM 内存限制导致大数据库 OOM

建议在初始化时增加 WASM 内存上限，或者恢复 better-sqlite3 支持

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

diegosouzapw:
你好，感谢反馈。

先说明一下：better-sqlite3 目前**就是**默认/首选的 SQLite 驱动，从未被移除——OmniRoute 内部有一条驱动降级链（`src/lib/db/adapters/driverFactory.ts`）：**better-sqlite3 → node:sqlite → sql.js (WASM)**。sql.js 只会在前两个驱动都加载失败时才启用，属于兜底方案（原生二进制在 `optionalDependencies` 中，npm 在某些环境/镜像源下会静默跳过原生构建，这在 Windows + npm 全局安装时比较常见）。

sql.js 本身架构上就是「把整个数据库文件读进内存 + 每次写入都整体 export 回磁盘」（没有增量/WAL 写入路径），所以哪怕调高 WASM 的内存上限，大数据库场景下依然会占用数倍于文件大小的 Node 堆内存——这不是一个可以通过简单调参解决的问题，而是 sql.js 这个后备方案本身的设计局限。真正的解法是让 better-sqlite3（或 node:sqlite）成功加载。

针对「大库 OOM」，我们已经内置了：
1. `omniroute runtime repair`（或顶层别名 `omniroute repair`）——会把 better-sqlite3 重新安装到用户可写的 runtime 目录，绕开需要 C++ 工具链的构建问题，专门用来修复「原生驱动加载失败、被迫回退到 sql.js」这种情况。
[...]

PR #10278, opened 2026-08-13
fix(db): publish the sql.js database atomically instead of rewriting it in place

### LINKED ISSUE #10277: bug(db): sql.js fallback persist() rewrites DB in place — concurrent readers see SQLITE_CORRUPT, interrupted save truncates
### OmniRoute Version

3.8.50

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

WSL2 (kernel 6.18), ext4

### Node.js Version

v22.22.3

### Description

`sqljsAdapter.persist()` rewrites the entire database image on every save, because sql.js has no incremental write path. It does so with `fs.writeFileSync(filePath, …)`, which opens the destination with `O_TRUNC`: for the whole duration of the write the on-disk database is 0 bytes and then partial. The window scales with database size and recurs on every save.

Unlike better-sqlite3 / node:sqlite, that window is **not** covered by SQLite's locking protocol, so it is visible to every OTHER process reading the same file — a backup job, a metrics exporter, an operator running `sqlite3`. Those readers get `SQLITE_CORRUPT` ("database disk image is malformed"), while `PRAGMA integrity_check` on the same file passes moments later. That combination makes the failure look random and points the blame at the reader rather than at the writer.

There is also a durability consequence: a crash or kill during `writeFileSync` leaves the real database truncated, with no intact copy anywhere.

This is reachable on every install where sql.js is genuinely the active driver — the paths tracked in #7132, #7152 and #8753.

### Steps to Reproduce

1. Run OmniRoute on an install where the sync driver cascade falls through and sql.js is the active adapter (`console.warn("[DB] Synchronous drivers unavailable — falling back to sql.js (WASM)")` at startup).
2. Confirm the adapter in use holds **no** persistent file descriptor on the database — a native driver keeps the database, `-wal` and `-shm` open for the process lifetime:
   `ls -l /proc/<server-pid>/fd | grep storage.sqlite`
   The PID to inspect is the child process titled `omniroute (vX.Y.Z)`, not the launcher.
3. Send traffic so the server writes (any request that persists a call log).
4. While that happens, poll the file size from another process:
   `while :; do stat -c%s ~/.omniroute/storage.sqlite; done`
5. In parallel, read the database from another process in a loop, e.g. repeated `sqlite3 ~/.omniroute/storage.sqlite 'select count(*) from provider_connections;'`.

### Expected Behavior

A concurrent reader either sees the previous complete database image or the new complete one. Rewriting the image should never expose a truncated file to other processes, and an interrupted save should never destroy the only copy.

### Actual Behavior

Step 4 transiently reports `0`, then intermediate sizes, then the full size again — on an unchanged inode. Step 5 fails with `database disk image is malformed` (`SQLITE_CORRUPT`, code 11) whenever a read lands inside that window, while `PRAGMA integrity_check` run afterwards returns `ok`.

Characterised at the syscall level: `bpftrace` on `sys_enter_openat` shows the server opening the database path with `O_TRUNC` repeatedly during normal traffic; during the window `stat`, `fstat` and `lseek(SEEK_END)` all return 0 and `pread` of the SQLite header returns nothing, so the truncation is real rather than a stale stat.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
# reader, during a save
Error: database disk image is malformed  (sqlite3_errcode = 11 / SQLITE_CORRUPT)

# same file, immediately afterwards
sqlite> PRAGMA integrity_check;
ok
```

### Additional Context

- Suggested fix: write to a temp file in the **same directory**, `fsync`, then `rename()` over the destination. `rename` is atomic on POSIX and on Windows for a same-volume replace, so readers only ever observe a complete image; it also turns the interrupted-save case into a stale temp file rather than a destroyed database. A temp file in `os.tmpdir()` would not work — a cross-device `rename` is not atomic.
- Related: #8707 (first-class native SQLite baseline) would make sql.js rare, but not safe — the two are complementary.

### Validation Plan

```
node --import tsx/esm --test tests/unit/db-sqljs-atomic-persist.test.ts
npm run typecheck:core
npm run check:changelog-integrity
```

The regression guard should assert the property without racing a timer: open a reader descriptor before a save, then after the save assert the descriptor still yields the complete previous image with a valid `SQLite format 3` header, that the published file sits on a new inode, and that no temp file survives a successful save. Such a test fails on the current implementation and passes on the fixed one.

maxmad64bis:
## Summary

`sqljsAdapter.persist()` rewrites the whole database image on every save — sql.js has no incremental write path. It did so with `fs.writeFileSync(filePath, …)`, which opens the destination with `O_TRUNC`: for the duration of the write the on-disk database is 0 bytes and then partial. The window scales with database size and recurs on every save.

Unlike better-sqlite3 / node:sqlite, that window is **not** covered by SQLite's locking protocol, so it is visible to every OTHER process reading the same file — a backup job, a metrics exporter, an operator running `sqlite3`. Those readers get `SQLITE_CORRUPT` ("database disk image is malformed") while `PRAGMA integrity_check` passes moments later, which makes the failure look random and blames the reader.

This PR writes to a temp file in the same directory, `fsync`s it, and `rename()`s it over the destination. `rename` is atomic on POSIX and on Windows for a same-volume replace, so a reader sees either the previous image or the new one — never a truncated one. It also closes a total-loss window: a crash mid-write used to leave the real database truncated; now it only leaves a stale temp file behind.

The failure was characterised at the syscall level on a multi-hundred-megabyte database: repeated `openat(O_TRUNC)` on the database path (`bpftrace` on `sys_enter_openat`), with `stat`/`fstat`/`lseek` all returning 0 mid-write and the SQLite header absent, on an unchanged inode — and a concurrent reader failing the majority of its reads during that period.

Relevant to the installs where sql.js genuinely is the active driver (#7132, #7152, #8753).

## Related Issues

- Closes #10277
- Related to #7132, #7152, #8753 — the sql.js-active paths where this window is reachable
- Related to #8707 — a first-class native SQLite baseline would make sql.js rare, not safe

⚠️ base-red inherited: #9985

## Validation

- [x] Change type: **DB**
- [x] `npm run lint` (ESLint on the changed files — clean)
- [x] Reconciled with the current active release base (`upstream/release/v3.8.50`); focused checks rerun afterward

Commands run:

```
node --import tsx/esm --test tests/unit/db-sqljs-atomic-persist.test.ts        # 2 pass
node --import tsx/esm --test tests/unit/db-sqljs-close-poison-7494.test.ts \
                            tests/unit/db-sqljs-preinit-ordering-gap-7288.test.ts \
                            tests/unit/db-sqljs-atomic-persist.test.ts        # 9 pass
npm run typecheck:core                                                        # clean
npx eslint src/lib/db/adapters/sqljsAdapter.ts tests/unit/db-sqljs-atomic-persist.test.ts   # clean
npm run check:changelog-integrity                                             # OK
```

**TDD record (Hard Rule #18)** — the guard was verified against the previous implementation, not just the new one:

```
# with the pre-fix persist() restored from HEAD:
not ok 1 - sql.js persist() publishes the database atomically …
    a reader holding the file open observed the image change underneath it —
    persist() replaced the file in place instead of renaming a new one over it
# pass 1  # fail 1

# with the fix:
ok 1 - sql.js persist() publishes the database atomically …
ok 2 - sql.js persist() is a no-op for :memory: databases …
# pass 2  # fail 0
```

## Tests Added Or Updated

- `tests/unit/db-sqljs-atomic-persist.test.ts` (new, 2 cases)

The guard asserts the property that separates the two implementations **without racing a timer** — a sleep-based test would be flaky and would prove nothing about small databases written in one go. Instead it opens a reader descriptor before a save and asserts that (a) the descriptor still yields the complete previous image with a valid `SQLite format 3` header, (b) the published file sits on a **new inode**, and (c) no temp file survives a successful save. A second case covers the `:memory:` early return.

## Coverage Notes

`src/lib/db/adapters/sqljsAdapter.ts` — `persist()` is exercised by the new test through the public `exec()` + `checkpoint()` path, on both the on-disk and `:memory:` branches, including the "no leftover temp file" post-condition. The error path (unlink + rethrow on a failed write) is not covered; it is a defensive cleanup branch that would need a filesystem fault injector to reach.

## Reviewer Notes

- The temp file deliberately lives in the **same directory** as the database, not `os.tmpdir()` — a cross-device `rename` is not atomic and would silently degrade to a copy.
- `fsync` before `rename` is intentional: the rename is atomic but only orders against data that already reached the disk, so without it an unclean shutdown can publish an empty file.
- Behavior change to be aware of: the database file now gets a **new inode** on every save. Anything that pins the old inode (a long-lived `tail -f`-style reader, a hardlink-based backup) will keep seeing the older image instead of a corrupted one — which is the point, but it is a change.
- No change to the driver cascade or to when sql.js is selected; this only makes the fallback safe once it is in use.


diegosouzapw:
Thanks for this — excellent bug report and fix. The syscall-level characterization in #10277 (bpftrace on `sys_enter_openat`, the `PRAGMA integrity_check` vs `SQLITE_CORRUPT` contradiction) made the root cause unambiguous, and the temp-file-same-directory + fsync + rename approach here is the right fix for exactly the reason you called out: cross-device rename to os.tmpdir() would silently degrade to a non-atomic copy.

I reviewed the diff and independently verified:
- Both new tests in `db-sqljs-atomic-persist.test.ts` pass against your branch (2/2).
- The two adjacent regression suites (`db-sqljs-close-poison-7494`, `db-sqljs-preinit-ordering-gap-7288`) still pass — no interaction regression with the close/init lifecycle.
- ESLint and `tsc --noEmit` are clean on the touched files.
- The TDD record in your PR body (failing against the pre-fix implementation, passing against the fix) satisfies our bug-fix validation rule.

On the failing CI checks: I traced them and they're all base-tip drift unrelated to your diff — I reproduced the same `check:agent-skills-sync` and ESLint-suppression failures directly against the current `release/v3.8.50` tip with none of your changes applied. Your `⚠️ base-red inherited` note was the right call. This will merge cleanly once the base drift is swept separately; no action needed on your end.

Nice catch on the inode-identity detail in the test design too — asserting the pre-opened reader descriptor still resolves the old complete image (rather than racing a timer) is exactly the right way to prove this property deterministically.

diegosouzapw:
### Babysit summary

The four failing checks were **stale verdicts**, not defects. `ci.yml`/`quality.yml` only fire on `synchronize`/`opened`, which the PR does not emit when the base moves. This PR's head (`9e17b06e`) predates 76 base commits, so the reds it showed were computed against an old merge-result.

**Fixed:** refreshed the branch by merging the current `release/v3.8.50` tip (`774127be3`, green) into the PR head — the merge commit itself has **zero lines of code** (no conflicts, including in `src/lib/db/`). Pushing it fired a fresh run.

- Refresh merge `7a720d36` (76 base commits) — resolved `Merge integrity`, `Fast Quality Gates`, `Unit Tests fast-path (2/4, 3/4, 4/4)`, `No new ESLint warnings`; all now pass on the refreshed SHA.

**Verified:** the PR's own `db-sqljs-atomic-persist` unit test passes 2/2 in isolation (no DB-handle hang). Changelog fragment `changelog.d/fixes/sqljs-atomic-persist.md` is clean (no `PRNUM` placeholder).

**Final `gh pr checks 10278` (head `7a720d36`):** Build (advisory) pass, Change Classification pass, Docs Gates pass, **Fast Quality Gates pass**, **Merge integrity pass**, **No new ESLint warnings pass**, **Unit Tests fast-path 1/4–4/4 pass**, Vitest pass, dast-smoke pass, semgrep pass. No failures, no pending. (Mergify protections/queue skipping until a review clears.)

No review threads outstanding; no test/gate weakened; no CI workflow edited. Ready for human review & merge.

diegosouzapw:

Issue #8700, opened 2026-07-26
feat(providers): per-model compatibility option to copy reasoning_content → content when content is empty


stgreenb:
## Problem / Use Case

Some upstream providers (notably opencode/opencode-go, which proxies to DeepSeek V4 Flash) return the model's entire response in `reasoning_content` with `content: ""` or `content: null`. This is valid for reasoning models that emit their output through the thinking channel, but the OpenAI Chat Completions format expects visible output in `content`. Currently:

- `validateResponseQuality` correctly identifies these as having valid content (non-empty `reasoning_content` counts as content), so the response is treated as a success and forwarded to the client.
- But the client receives `content: ""` — an empty response despite a 200 OK — with the actual answer trapped in `reasoning_content`.
- The only workaround today is external middleware (e.g., a proxy that copies `reasoning_content` → `content` before forwarding to the client).

Note: The related issue of `validateResponseQuality` treating reasoning-only responses as valid (rather than triggering combo failover) is addressed separately in #7000. This ticket assumes that fix is in place or handled independently.

### Log evidence from a real `opencode-zen/deepseek-v4-flash-free` request

Response body captured in call logs:

```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": "",
      "reasoning_content": "We are asked: \"Return just the word: hello\". So the response should be just \"hel"
    }
  }]
}
```

Token accounting: `out: 20, reasoning: 20` — 100% of completion tokens are reasoning tokens, none in content. Every successful request to this model follows this pattern (verified across 10+ log entries).

## Proposed Solution

Add a per-model compatibility option, e.g. `reasoningToContentFallback: boolean`, configurable in the provider model's Compatibility settings. When enabled:

- If the upstream response has `choices[0].message.content` empty/null/whitespace AND `choices[0].message.reasoning_content` is non-empty:
  - Copy the `reasoning_content` value into `content` (for both streaming deltas and non-streaming response bodies).
  - Optionally clear `reasoning_content` or keep it (preserving both fields lets the client see the thinking trace while also getting visible output).
- This should apply at the response normalization layer, after the provider-specific executor processes the raw upstream response but before `validateResponseQuality` checks it — so validation sees real content and doesn't flag it as reasoning-budget-exhausted.

## Acceptance Criteria

- When `reasoningToContentFallback` is enabled on a model:
  - Non-streaming response with empty `content` + non-empty `reasoning_content` → client receives `content` populated with the reasoning text.
[...]

Issue #6274, opened 2026-07-05
fix(api): max_tokens replaced incorrectly (test max_tokens=1 sent as 1001)


brightfiscalband:
### OmniRoute Version

3.8.44

### Installation Method

npm (global)

### Operating System

Windows

### Provider(s) Involved

opencode

### Model(s) Involved

glm-5.2

### Client Tool

claude code

### Description

When selecting the model in the Cloud Code CLI using the `/model glm-5.2` command, a test request is made with `max_tokens=1`, yet a request with `max_tokens=1001` is actually sent to the model. This wastes tokens and consumes time.

### Steps to Reproduce

Run the `/model glm-5.2` command. 
Check the logs.
Compare `max_tokens` in the Client Raw Request and the OpenAI Request.

### Expected Behavior

max_tokens in Client Raw Request and the OpenAI Request are the same

### Actual Behavior

In Client Raw Request max_tokens=1
In  OpenAI Request max_tokens=1001

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Client Raw Request
{
  "timestamp": "2026-07-05T10:41:55.227Z",
  "endpoint": "/v1/messages",
  "headers": {
    "accept": "application/json",
    "accept-encoding": "gzip, deflate, br, zstd",
    "access-control-allow-headers": "Content-Type, Authorization, x-api-key, anthropic-version, x-omniroute-connection, x-internal-test, accept",
    "access-control-allow-methods": "GET, POST, PUT, DELETE, PATCH, OPTIONS",
    "access-control-allow-origin": "*",
    "anthropic-beta": "claude-code-20250219,interleaved-thinking-2025-05-14,redact-thinking-2026-02-12,thinking-token-count-2026-05-13,context-management-2025-06-27,prompt-caching-scope-2026-01-05,mid-conversation-system-2026-04-07",
    "anthropic-dangerous-direct-browser-access": "true",
    "anthropic-version": "2023-06-01",
    "authorization": "[REDACTED]",
    "connection": "keep-alive",
    "content-length": "516",
    "content-type": "application/json",
    "host": "localhost:20128",
    "user-agent": "claude-cli/2.1.195 (external, cli)",
    "vary": "Origin",
    "x-app": "cli",
    "x-claude-code-session-id": "87bb9cc3-f55a-4580-b81c-96411ee7572f",
    "x-forwarded-for": "127.0.0.1",
    "x-forwarded-host": "localhost:20128",
[...]

PR #10284, opened 2026-08-13
fix(sse): answer tiny-budget reasoning probes with a truncated 200 (#10281)

### LINKED ISSUE #10281: fix(api): answer tiny-budget reasoning probes (max_tokens:1) with truncated 200 — no 500 empty response, no cooldown poisoning
## Summary

Claude Code sends a tiny-`max_tokens` capability probe (`max_tokens: 1`) when switching models (`/model`). For reasoning-capable models (reproduced with `cl/deepseek/deepseek-v4-flash` on the `cline` provider), the entire tiny budget is consumed by reasoning, so:

- **Non-streaming probe** → upstream (`api.cline.bot`) returns HTTP 500 `{"error":{"message":"empty response content"}}`, which OmniRoute relays verbatim as `[500]: empty response content` → Claude Code fails the model switch.
- **Streaming probe** → upstream returns 200 with reasoning-only chunks (`content: ""`) → client receives an empty response (related to #8700).
- **Side effect:** the 500 marks the connection `unavailable (500)` and triggers fallback/cooldown bookkeeping, so a connectivity probe degrades account health state. Logs: `Account bd3c26b7... unavailable (500), trying fallback` → `Preserving last upstream error after credential exhaustion`.

## Environment

- OmniRoute `3.8.48`
- Provider: `cline` (OAuth account), model `deepseek/deepseek-v4-flash` (alias `cl/deepseek/deepseek-v4-flash`, reasoning-capable, free tier)

## Steps to reproduce

Point Claude Code at OmniRoute with model `cl/deepseek/deepseek-v4-flash`, then switch models in Claude Code (`/model`).

Direct repro (non-stream probe):

```bash
curl http://localhost:20128/v1/chat/completions \
  -H "Authorization: Bearer $OMNIROUTE_KEY" \
  -d '{"model":"cl/deepseek/deepseek-v4-flash","messages":[{"role":"user","content":"ping"}],"max_tokens":1,"stream":false}'
```

Result:

```json
{"error":{"message":"[500]: empty response content","type":"server_error","code":"internal_server_error"}}
```

With `stream: true`, the same request returns HTTP 200 but streams only reasoning deltas, then `finish_reason: "length"` with `content: ""` and `usage.reasoning_tokens` == `completion_tokens` — an empty response to the client.

Same model via a healthy provider (`nvidia/deepseek-ai/deepseek-v4-flash-0731`) returns `200` with real content even at `max_tokens: 1`, so this is specific to how the cline upstream treats a tiny non-stream budget.

## Root cause

1. `resolveReasoningBufferedMaxTokens` (`open-sse/services/reasoningTokenBuffer.ts`) deliberately forwards tiny caller budgets verbatim (`REASONING_BUFFER_MIN_TRIGGER = 256`, see #6274) — probes are not inflated, so the upstream receives `max_tokens: 1`.
2. Reasoning models emit ~130+ tokens of reasoning before any content; a 1–8 token budget yields a reasoning-only response, and cline's gateway returns HTTP 500 `"empty response content"` for the non-stream case.
3. `errorClassifier.ts` already whitelists `finish_reason: "length"` / `"tool_calls"` empty-content 200s as legitimate — explicitly "a Claude Code `max_tokens: 1` connectivity ping" — but the non-stream path relays the upstream's 500 *before* that classification applies, and the streaming path forwards empty `content` that strict clients reject.
4. The probe failure is recorded as a provider 500 → connection marked unavailable → cooldown/fallback bookkeeping, even though the probe is not a genuine model failure.

## Proposed fix

1. **Non-stream path (`chatCore.ts`)**: when the provider error is an empty-content outcome for a reasoning-capable model whose request used `max_tokens < REASONING_BUFFER_MIN_TRIGGER`, synthesize a valid truncated response (`200`, `content: ""`, `finish_reason: "length"`, usage populated with `reasoning_tokens`) instead of relaying the upstream 500. This matches the intent already documented in `errorClassifier.ts`.
2. **Don't poison health**: exclude tiny-budget probes from account-unavailable / provider-cooldown bookkeeping (skip the `Account ... unavailable (500), trying fallback` path for these).
3. **Streaming / visible content**: implement #8700's per-model compatibility option (copy `reasoning_content → content` when `content` is empty) so probes and real responses for reasoning models always carry visible output.
4. **Optional**: add a configurable per-model minimum output budget (e.g. a new `model_capability_overrides` key) for operators who prefer the upstream to actually emit content instead of synthesizing a truncated response.

## References

- #8700 — feat(providers): per-model compatibility option to copy `reasoning_content → content` when content is empty (open, backlog)
- #6274 — fix(api): `max_tokens` replaced incorrectly (test `max_tokens=1` sent as 1001) — why tiny budgets are respected verbatim
- #3587 — [BUG] Reasoning models in combos consume all tokens for `reasoning_content`, leaving content empty — where the combo-only reasoning buffer came from
- #5131 — [BUG] Streaming reasoning models return empty content without validation

harkaranbrar7:
## Summary

Fixes #10281 — Claude Code's `/model` capability check sends `max_tokens: 1` when switching models. For reasoning-capable models (e.g. `cl/deepseek/deepseek-v4-flash` on the `cline` provider), the whole tiny budget is consumed by thinking, and the upstream (`api.cline.bot`) answers the non-streaming probe with HTTP 500 `{"error":{"message":"empty response content"}}`.

OmniRoute relayed that as `[500]: empty response content`, so Claude Code failed the model switch. Worse, the 500 marked the connection `unavailable (500)` and poisoned fallback/cooldown bookkeeping for what is only a probe (`Account bd3c26b7... unavailable (500), trying fallback` → `Preserving last upstream error after credential exhaustion`).

## Fix

In the non-streaming `providerFailure` path of `handleChatCore`, when a **tiny-budget reasoning probe** (`max_tokens < REASONING_BUFFER_MIN_TRIGGER` on a thinking-capable model) hits an **empty-content 5xx** from the upstream, answer with a valid truncated response (`200`, empty content, `finish_reason: "length"`, usage populated) instead of relaying the failure — and skip all failure bookkeeping so the probe never poisons connection health.

This mirrors the semantics OmniRoute already grants to `finish_reason: "length"` empty 200s in `errorClassifier.ts` (`LEGIT_EMPTY_OPENAI_FINISH`, explicitly "a Claude Code `max_tokens: 1` connectivity ping").

### Why not bump the forwarded budget instead?

`resolveReasoningBufferedMaxTokens` deliberately forwards tiny budgets verbatim (`REASONING_BUFFER_MIN_TRIGGER = 256`, see #6274 — "max_tokens=1 sent as 1001"). Inflating probes would reopen that regression; synthesizing a truncated 200 keeps the probe verbatim upstream while making it succeed for the client.

### Scope

- Non-streaming path only — streaming probes already get a valid 200 (reasoning-only stream, `finish_reason: "length"`).
- The companion "copy `reasoning_content → content` when content is empty" behavior for streams/real responses is tracked separately in #8700 (already open, backlog).

## Repro

```bash
curl http://localhost:20128/v1/chat/completions \
  -H "Authorization: Bearer $OMNIROUTE_KEY" \
  -d '{"model":"cl/deepseek/deepseek-v4-flash","messages":[{"role":"user","content":"ping"}],"max_tokens":1,"stream":false}'
# before: {"error":{"message":"[500]: empty response content",...}}
# after:  200 with {choices:[{message:{content:""},finish_reason:"length"}],usage:{completion_tokens:1}}
```

## Tests

- `tests/unit/reasoning-probe-truncated-response-10281.test.ts` — probe detection, empty-content 5xx detection, and the truncated-response builder (including the invariant that the synthetic body is not re-flagged by `isEmptyContentResponse`).
- Existing `tests/unit/reasoning-token-buffer-6274.test.ts` still passes (probe budgets remain verbatim).
- `tsc --project tsconfig.typecheck-core.json` and ESLint on the changed files are clean.


diegosouzapw:
Thanks for this — nicely scoped fix, and the root-cause writeup in the PR description made it easy to review end-to-end.

I traced the fix through `handleChatCore`'s non-streaming failure path and it correctly reuses the existing `providerFailure:` / `break providerFailure` pattern the codebase already uses for the signature-recovery case, so it cleanly skips the connection-health bookkeeping that was getting poisoned by these probes, and falls through to the normal success/parsing path. The synthetic `finish_reason: "length"` response also correctly passes the existing `isEmptyContentResponse()` legitimacy check in `errorClassifier.ts` — good catch keeping that invariant, and nice that you asserted it directly in the test.

I ran the new test file locally and it's 3/3 green, and ESLint on the touched files is clean. Everything here looks solid to me.

Two small, entirely optional things you could consider (not blockers):
- `buildReasoningProbeTruncatedResponse` hardcodes `usage.prompt_tokens: 0` — could reflect the real prompt token count if it's cheaply available at that call site.
- The `EMPTY_CONTENT_FAILURE_RE` regex matches a couple of message phrasings beyond the one you actually saw upstream ("empty response content") — fine to leave as-is given it's gated behind the tiny-budget-probe check first, but worth knowing it's slightly broader than the observed repro.

No action needed from you right now — this just needs to go through our CI run (first-time external PRs need a maintainer to approve the workflow run) before it merges. Thanks again for tracking this down and for the clean repro + tests.

diegosouzapw:

PR #10288, opened 2026-08-13
fix(docker): prefix cache mount ids with Railway service scope

anudeepadi:
## Summary
- Railway's Dockerfile builder rejects `--mount=type=cache` ids that lack the `s/<service-id>-` prefix — build failed at Dockerfile syntax validation, before any build steps ran (four identical "dockerfile invalid" errors).
- Prefixes all 7 cache mount ids in the Dockerfile (`apt-cache`, `apt-lists` across 4 RUN blocks, `npm-cache` x2, `next-cache` x1) with `s/92ca8a61-c1ba-421f-a389-d48ac7258c2d-` (the omni-route production service id), per https://docs.railway.com/reference/dockerfiles.

## Note on portability
The service-id segment is a literal (Railway forbids env vars in cache mount ids), scoped to the `omni-route` Railway service this was deployed from. This is purely a syntax-format requirement validated statically — it doesn't affect `docker build`/`docker-compose`/other Railway services building this same Dockerfile; a different Railway service would just get a non-matching cache namespace (no functional break), not a build failure.

## Test plan
- [ ] Redeploy `omni-route` on Railway and confirm the build passes the Dockerfile validation stage
- [ ] Confirm the build completes and the service comes back online

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,8 @@ WORKDIR /app
 # that already have a fix published in trixie. CVEs without an upstream fix yet
 # (local-only TOCTOU, etc.) remain until the distro patches them and the image
 # is rebuilt; none are reachable from the proxy's request surface at runtime.
-RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt,sharing=locked \
-  --mount=type=cache,id=apt-lists,target=/var/lib/apt/lists,sharing=locked \
+RUN --mount=type=cache,id=s/92ca8a61-c1ba-421f-a389-d48ac7258c2d-apt-cache,target=/var/cache/apt,sharing=locked \
+  --mount=type=cache,id=s/92ca8a61-c1ba-421f-a389-d48ac7258c2d-apt-lists,target=/var/lib/apt/lists,sharing=locked \
   apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends libsecret-1-0 ca-certificates \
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks for tracking down the exact Railway cache-mount-id requirement here — I checked Railway's docs and the `s/&lt;service-id&gt;-&lt;name&gt;` format you used matches what they document, and I confirmed this branch merges cleanly against the current release tip with no conflicts.

A couple of notes before this lands:

1. Since the id is scoped to your own Railway service, it won't do anything for anyone else's Railway deployment of this Dockerfile (as you already noted) — that's fine and expected, just flagging it so the reasoning is clear for future readers. A short comment above the mount lines linking to the Railway docs would help the next person who wonders why that UUID is there.
2. The unrelated `open-sse/utils/stream.ts` cleanup (dropping the unused `sseCommentsEnabled` import + a couple of formatting fixes) is harmless and actually fixes a small pre-existing Prettier drift on that file, so no concerns there — just worth keeping Docker/infra fixes and app-code cleanups in separate PRs going forward for easier review.
3. Since this is your first PR here, the CI workflows (hadolint, Quality Gates, etc.) are sitting on manual approval rather than having actually run yet — that's on our end to approve, not something you need to do.
4. Your test-plan checkboxes are still unchecked — once you've confirmed the redeploy on Railway actually passes validation, let us know here and we'll get this merged.

diegosouzapw:

Issue #3696, opened 2026-06-12
[bug] antigravity/agy: gemini-3.1-pro-high/low (and gemini-3-pro-preview) alias to retired upstream id gemini-3.1-pro -> 404


diegosouzapw:
## Bug

**Source:** Discussion #3607 by @markbakaa88 (reporter-confirmed live on v3.8.21)

## Problem

The public catalog entries `gemini-3.1-pro-high` and `gemini-3.1-pro-low` (listed by BOTH the `antigravity` and `agy` providers) alias to the upstream id `gemini-3.1-pro`, which the Antigravity backend has retired. Every request to those tiers now returns **404**.

The flagship entry `gemini-3-pro-preview` aliases to the same dead upstream id (`gemini-3.1-pro`), so it is presumed broken as well and must be audited in the same pass.

The only live "3.1 Pro" upstream id is `gemini-pro-agent` (sent verbatim, no alias) - confirmed working by the reporter on v3.8.21.

## Code references

- `open-sse/config/antigravityModelAliases.ts:160` - `"gemini-3-pro-preview": "gemini-3.1-pro"`
- `open-sse/config/antigravityModelAliases.ts:163-164` - `"gemini-3.1-pro-high"/"-low": "gemini-3.1-pro"`
- `open-sse/config/antigravityModelAliases.ts:71-88` - the two broken tiers still in `ANTIGRAVITY_PUBLIC_MODELS`
- `open-sse/config/agyModels.ts:39-57` - same tiers pinned in the `agy` catalog
- History: the `-high/-low` -> plain remap was the #3229 fix (v3.8.11), when plain `gemini-3.1-pro` was still live upstream. The upstream id has since been retired in favor of `gemini-pro-agent`.

## Proposed fix

1. Remap `gemini-3.1-pro-high`, `gemini-3.1-pro-low` and `gemini-3-pro-preview` to the live upstream id (`gemini-pro-agent`), or de-list the budget tiers if there is no per-tier upstream support (the backend appears to expose a single Pro tier now).
2. Fix the duplicate display name: `gemini-3.1-pro-high` and `gemini-pro-agent` are BOTH labeled "Gemini 3.1 Pro (High)" in the model picker (`antigravityModelAliases.ts:73` vs `:128`), which is exactly the confusion the reporter hit.
3. Keep `ANTIGRAVITY_REVERSE_MODEL_ALIASES` / quota-bucket mapping consistent with the new forward aliases.

[...]

PR #10305, opened 2026-08-13
feat(oauth): add gemini-3.7-flash models for antigravity and agy providers

Chewji9875:
### Summary

Adds `gemini-3.7-flash` and its reasoning tiers (`gemini-3.7-flash-high`, `gemini-3.7-flash-medium`, `gemini-3.7-flash-low`) for the `antigravity` (OAuth) and `agy` (CLI) providers.

### Changes
- **`open-sse/config/antigravityModelAliases.ts`**: Added `gemini-3.7-flash` public models to `ANTIGRAVITY_PUBLIC_MODELS` and mapped default alias to `gemini-3.7-flash-high` in `ANTIGRAVITY_MODEL_ALIASES`.
- **`open-sse/config/agyModels.ts`**: Added `gemini-3.7-flash` tiers to `AGY_PUBLIC_MODELS`.
- **`src/shared/constants/modelSpecs.ts`**: Registered model specs for `gemini-3.7-flash` tiers.
- **`src/shared/constants/pricing/oauth-subscriptions.ts`**: Added pricing configuration for token/cost accounting under `ag`.
- **`src/shared/constants/cliTools.ts`**: Added model aliases and default CLI models for Antigravity.
- **`open-sse/config/freeModelCatalog.data.ts`**: Added budget definitions to free model catalog.
- **`tests/unit/antigravity-model-aliases.test.ts`**: Added test assertions covering Gemini 3.7 Flash tiers.

### Verification
- `node --import tsx/esm --test tests/unit/antigravity-model-aliases.test.ts` passed (8/8).

⚠️ base-red inherited: #9985


alvinunreal:
lets go!

diegosouzapw:
Thanks for keeping the model catalog current — the gemini-3.7-flash tier additions follow the same shape as the 3.6/3.5 Flash work, which is appreciated.

I ran the PR against its own CI (workflow run 31784352026-31784352028) and found the current CI red is caused by this diff, not inherited from base:

1. `tests/unit/agy-gemini-3696-tier-passthrough.test.ts` fails: this test (from #3696) asserts that no two `ANTIGRAVITY_PUBLIC_MODELS` entries resolve to the same upstream id. Mapping `gemini-3.7-flash`, `-high`, `-medium`, and `-low` all to the single upstream id `gemini-3.7-flash-tiered` directly collides with that invariant. Could you confirm with fresh live testing whether the 3.7 tier genuinely has one shared upstream id (unlike 3.6, which uses distinct ids per tier), and if so, update that test with an explicit, documented exception rather than leaving it red?
2. `tests/unit/antigravity-retired-public-models.test.ts` fails: it expects the live default tiers to appear first in `ANTIGRAVITY_PUBLIC_MODELS`/`AGY_PUBLIC_MODELS`. The new gemini-3.7 entries were inserted ahead of the existing 3.6 block, breaking that ordering contract — could you move them after it?
3. `check:open-sse-typecheck` fails with 2 new TS2551 errors in `src/app/api/v1/models/catalog.ts` that aren't in the frozen baseline, and `check:dead-code` is 1 symbol over its baseline — both likely fall out of the same id-collision issue above.
4. `tests/unit/models-catalog-route.test.ts` also failed in the same run (an unrelated-looking `cline`/inactive-provider dedup assertion) — I wasn't able to fully confirm whether this one is caused by this PR or is a pre-existing flake, so it's worth a second look on your end too.

Separately, `AntigravityExecutor.shouldRetry` now also retries on 404/502/503/504 across fallback URLs — a reasonable resilience improvement, but it affects all antigravity traffic (not just the new models) and currently has no test coverage. Could you add a focused unit test for it?

Once the upstream-id design question is resolved and the tests above are green, this should be in good shape to merge. Let me know if you'd like a hand narrowing down the root cause on the upstream side.

diegosouzapw:
### Babysit summary

**Was stale, had real defects.** On first review the PR head was 76 commits behind the green `release/v3.8.50` tip; the fast-path CI verdict was computed against a pre-merge SHA. After a refresh-merge two **real** unit failures survived that the PR itself caused:

- **Unit Tests 1/4 — `#3696` collision**: adding four public model ids (`gemini-3.7-flash`, `-high`, `-medium`, `-low`) each aliased to the single upstream `gemini-3.7-flash-tiered` violated the base `#3696` invariant ("no two `ANTIGRAVITY_PUBLIC_MODELS` entries may resolve to the same upstream id"). The suffixed tier ids actually 404 upstream per the change itself.
- **Unit Tests 4/4 — catalog ordering**: the entries were prepended, breaking the leading-order contract (Gemini 3.6 tiers must stay first).

**Fix** (`f2f20d5`, pushed to the contributor fork; credit preserved):
- Collapsed gemini-3.7-flash to a **single** callable public model aliased to the live upstream `gemini-3.7-flash-tiered`, in both the antigravity and agy catalogs.
- Removed the four colliding tier specs, pricing, free-catalog and CLI default entries; placed the single entry after the leading Gemini 3.6 block.
- Updated the PR's own test to match the corrected single-model catalog (no base test weakened).
- Verified locally: `#3696`, `antigravity-retired-public-models`, `antigravity-model-aliases` tests all pass; `provider-consistency`, `dead-code`, `open-sse-typecheck`, `typecheck:core` all green.

**Status after push** (`b315280e`, synced to current green base tip `ee221d8`):
- Fast Quality Gates ✅ · Unit Tests 1/4 · 2/4 · 3/4 · 4/4 ✅ · Vitest ✅ · No new ESLint warnings ✅ · dast-smoke ✅ (re-ran flaky runner signature)
- `dast-smoke` had no failed-step log (runner issue) and passed on re-run.
- PR is MERGEABLE, no unresolved review threads.

Ready for human review & merge.

Chewji9875:
> lets go!

Using that deep work 😆

alvinunreal:
> > lets go!
> 
> 
> 
> Using that deep work 😆

ah really? 😃

PR #10308, opened 2026-08-13
fix(settings): default debugMode to false

RaviTharuma:
## Summary

- `getSettings()` seeds `debugMode: true` before overlaying sqlite. A missing key means **production debug logging is on**.
- CONTRIBUTING documents Debug Mode as an Advanced **opt-in** toggle.
- Default is now `false`. Persisted `true` is unchanged.

## Validation

- [x] Docs-only change (no production code) — N/A (one-line default + changelog)
- Existing settings tests that PATCH `debugMode: true` still apply.

## Reviewer Notes

No secrets. Behavior change only for installs that never persisted the key.

### DIFF
--- a/changelog.d/fixes/debugmode-default-false.md
+++ b/changelog.d/fixes/debugmode-default-false.md
@@ -1,0 +1,1 @@
+- **fix(settings):** default `debugMode` to false so installs without a persisted key are not in debug (`src/lib/db/settings.ts`)
--- a/src/lib/db/settings.ts
+++ b/src/lib/db/settings.ts
@@ -212,7 +212,7 @@ export async function getSettings() {
     idempotencyWindowMs: 5000,
     wsAuth: false,
     maxBodySizeMb: requestBodyLimitMbFromEnv(process.env.MAX_BODY_SIZE_BYTES),
-    debugMode: true,
+    debugMode: false,
     // Opt-in diagnostic: when true, the chat handler emits a `log.debug("TOOLS", …)`
     // line per request summarizing tool count + MCP/hosted/client source breakdown.
     logToolSources: false,
--- a/tests/unit/debugmode-default-false.test.ts
+++ b/tests/unit/debugmode-default-false.test.ts
@@ -1,0 +1,34 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+
+const TEST_DATA_DIR = fs.mkdtempSync(path.join(os.tmpdir(), "omniroute-debugmode-default-"));
+process.env.DATA_DIR = TEST_DATA_DIR;
+
+const core = await import("../../src/lib/db/core.ts");
+const settings = await import("../../src/lib/db/settings.ts");
+
+test.after(() => {
+  try {
+    core.closeDbInstance();
+  } catch {}
+  core.resetDbInstance();
+});
+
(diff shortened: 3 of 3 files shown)


RaviTharuma:
Implements / pairs with #10312.

diegosouzapw:
Thanks for the PR and for the detailed writeup in #10312 — I dug into this before merging and wanted to share what I found.

I traced every consumer of `debugMode` in the codebase (`src/`, `open-sse/`) and it has exactly one effect: it toggles visibility of a few dashboard sidebar sections (Translator, Playground, Search Tools) via `Sidebar.tsx`/`SidebarTab.tsx`. It does not gate any logging — there's no code path where this setting touches log level, log volume, or debug-class log lines anywhere in the current source. The "Enable debug request logs (UI)" line in CONTRIBUTING.md is stale relative to actual behavior; I'll get that wording fixed separately.

The `debugMode: true` default you're flipping was actually set deliberately, on purpose, back in f79ab2c3d1 ("default debugMode to true on fresh installations") — precisely so the Debug sidebar section shows up out of the box on new installs, instead of silently disappearing. Flipping it back to `false` here reintroduces that exact hidden-sidebar issue for every fresh install.

If you're seeing real debug-level log noise on a clean install, I'd genuinely like to track that down — but it isn't coming from this flag, so a fresh issue with a concrete log excerpt/repro would help a lot more than adjusting this default.

Given that, I don't think this should merge as-is, and I'll be flagging it to the maintainer as such — but happy to look at a follow-up once we've got the actual log-source pinned down, or at a small docs fix for the CONTRIBUTING.md wording if you'd like to send that instead. Appreciate you digging in and reporting it either way.

diegosouzapw:
### Babysit summary
- `PR Test Policy` was a **real** fail (production change to `src/lib/db/settings.ts` with no automated test). Fixed by adding `tests/unit/debugmode-default-false.test.ts`, which asserts `getSettings()` defaults `debugMode` to `false` and persists across reads. Committed as `a9255f58` (author RaviTharuma). PR Test Policy now passes.
- ⚠️ **base-red inherited: #9282** — the three remaining fails (`Build`, `Integration Tests (1/2)`, `Quality Ratchet`) are all repo-wide failures on `main`, not defects of this diff:
  - `Quality Ratchet` fails only the `codeql-ratchet` sub-step (`6 alertas CodeQL abertos > baseline 0`) — a repo-wide live code-scanning count on code this PR does not touch. Issue #9282 explicitly documents that CodeQL-count/baseline repo-wide gates turn every PR → main red.
  - `Integration Tests (1/2)` fails with `Timed out waiting for OmniRoute to start: HTTP 404` — the integration harness cannot start the server. Issue #9282 lists main's integration gate as over-ceiling/killed (unreleased DB handle). Unrelated to the `debugMode` default.
  - `Build` ran 48m then its log blob is unavailable (`BlobNotFound`); local `typecheck:core` is clean (`exit 0`) on the head, so there is no compile defect from this diff.
- The fix is the companion `--base main` PR described in issue #9282 §8, not this contributor PR. Handing off for human merge; no merge performed here.

diegosouzapw:
Hi @RaviTharuma — thanks for this. This fix is already covered byte-for-byte by #10372 (`defaultDebugMode=false` in `src/lib/db/settings.ts` + the same `debugmode-default-false` changelog fragment), which is queued to merge into `release/v3.8.50`. Closing as duplicate so we don't double-land it. If there's any gap remaining, feel free to reopen with specifics.

Issue #6408, opened 2026-07-06
fix(api): GET /v1/models serializes concurrent requests


chirag127:
## Problem
`GET /v1/models` processes requests serially. Under concurrency, each request's total time = base latency × its position in the queue. This makes the endpoint (which SDKs poll on startup) a chokepoint.

## Repro
Baseline (sequential 5x):
```
$ for i in {1..5}; do curl -sS -m 8 -o /dev/null -w 'HTTP=%{http_code} t=%{time_total}\n'     -H 'Authorization: Bearer <key>' http://localhost:20128/v1/models; done
HTTP=200 t=1.880
HTTP=200 t=1.166
HTTP=200 t=1.435
HTTP=200 t=1.160
HTTP=200 t=1.266
```

Concurrent 10x (all fired simultaneously in background):
```
HTTP=200 t=1.185
HTTP=200 t=2.160   ← 1s slower than #1
HTTP=200 t=3.040   ← 2s slower
HTTP=200 t=3.957
HTTP=200 t=4.841
HTTP=200 t=5.737
HTTP=200 t=6.620
HTTP=200 t=7.553
HTTP=000 timeout at 8s (curl -m 8)
HTTP=000 timeout at 8s
```

Linear staircase — each request waits for the previous to complete. 20 concurrent unauth requests: only 4 complete, 16 time out at 5s.

## Expected
Concurrent `GET /v1/models` should complete in ~1.2s each regardless of concurrency (the response is a cached in-memory catalog — should be O(1) not O(n)).

## Impact
- SDK-startup contention: OpenAI/Anthropic SDKs call `/v1/models` first; N concurrent fleet agents block each other for O(N) seconds.
- Dashboard UI (which polls `/v1/models` per tab) locks up under multi-tab open.
- Compounds with related report #3932 (perf improvements).
- Possibly connects to the queue-drain saturation class flagged in #3825 (504s under load).

## Suspected fix
Something in `/v1/models`'s handler holds a global lock or awaits a shared promise sequentially. Candidates:
- Provider-registry rebuild on each request (should be cached with SWR-style invalidation).
- Single-flight guard collapsing to sequential await instead of promise-sharing.
- SQLite `getDbInstance()` WAL contention if the handler reads DB on every call — cache the response for e.g. 5s.

Reproducer is 5 lines of curl; TDD hook per Hard Rule #18 is trivial: assert 10 concurrent `GET /v1/models` complete in < 2s total wall clock.

## Env
[...]

PR #10309, opened 2026-08-13
fix(api): hash API keys in the /v1/models catalog cache key

RaviTharuma:
## Summary

- `buildCatalogCacheKey` in `catalogCache.ts` put the **raw API key** into a process-wide `Map` key (`prefix|codex|${apiKey}|configuredOnly`).
- Heap dumps, inspector, accidental logs leak bearer tokens.
- Fingerprint with SHA-256 (16 hex chars). Empty key stays empty. Unit test asserts the raw secret never appears.

## Related Issues

- Catalog-cache key issue filed in this batch
- #6408 added the cache

## Validation

```text
node --import tsx --test tests/unit/catalog-cache-auth-fingerprint.test.ts
# tests 1, pass 1
```


RaviTharuma:
Implements / pairs with #10313.

diegosouzapw:
Thanks for this — the underlying fix is a real and worthwhile one. I checked and the raw-API-key-in-Map-key issue you describe in #10313 is still present on `release/v3.8.50` tip, and I reproduced your fix independently (applied the same `fingerprintCatalogAuthKey` hash + rewiring on a clean checkout of current tip, copied in your test) — it passes cleanly.

The reason I'm not marking this merge-ready as-is: your branch looks like it stacked on top of your own #10026 and #10055 before they merged, and wasn't rebased afterward. Both of those already landed on `release/v3.8.50` — but with different code than what's in this diff (the health-check-exclusion logic and the pricing memoization each shipped under a different implementation shape). So right now this PR's diff carries stale duplicate content for those two, plus two already-merged dependency bumps and the `.gitignore` hardening, and GitHub reports it as `CONFLICTING`/`DIRTY` as a result.

There's also a real trap hiding in there: `buildCatalogCacheKey` on current tip picked up two extra fields (`hideAuto`/`hideNoThink`) after your branch diverged. Your diff's version of that function only has 4 fields — so a naive conflict resolution could silently drop those.

Could you rebase (or just recreate the branch) onto current `origin/release/v3.8.50` and keep only your actual new commit — the hash fix — reapplied against the current 6-field `buildCatalogCacheKey`? Once that's clean and the existing tests pass, this should be quick to merge. Don't worry about the "Build"/"Quality Ratchet" red X's you're seeing right now — I traced both and neither is caused by your code (one's an infra runner cancellation, the other is repo-wide CodeQL alert timing unrelated to `catalogCache.ts`); they should clear on the next CI run after rebase.

Nice catch on the original issue, by the way — that's a genuinely nasty little leak surface.

diegosouzapw:
### Babysit summary

The two failing checks (`Build`, `Quality Ratchet`) were **stale verdicts**, not this PR's defects.

- **Root cause**: the PR head was 2280 commits behind `release/v3.8.50`. The base moved after this PR opened, so the last CI run (`run 31748452259`, CI workflow) was computed against a frozen pre-fix merge-ref — `Build` + `Quality Ratchet` reds are the classic base-drift signature. A base-green issue (`#9985`) is open, but its listed failures are unit/vitest/integration/package-artifact ceiling-kills on a scheduled run — none are this PR's gates.
- **Refresh-merge** (`0db8e4403a`) merged the current `release/v3.8.50` tip (774127be3) into the PR head and resolved the conflicts:
  - `src/app/api/v1/models/catalogCache.ts` — combined the PR's API-key hash fingerprint with the base's `hideAuto`/`hideNoThink` cache-key segments (both behaviors preserved).
  - `src/lib/modelsDevSync.ts` + its test — took the base's version-checked `getModelsDevPricing` memo (base #10055 superseded the older memo port the branch carried).
  - `.gitignore` and `package-lock.json` — took base's current versions.
- **Fresh run** (`31882167647`, Quality Gates workflow): all gates green — Build (advisory), Fast Quality Gates, No new ESLint warnings, Unit Tests fast-path (4/4), Vitest (fast-path), dast-smoke, semgrep.
- **Final state**: PR is `MERGEABLE` / `CLEAN`; no pending checks; no unresolved review threads.

Ready for human review & merge.

diegosouzapw:
Thanks @RaviTharuma for this security fix. The catalog cache now fingerprints the API key instead of retaining the raw bearer token in the in-process Map key, while preserving the current six-field cache partitioning and the existing catalog-cache coverage. I validated the focused regression plus the two existing catalog-cache callers locally (8 tests passed). This is queued for the v3.8.50 release train.

PR #10320, opened 2026-08-13
fix(i18n): unflatten dotted namespace keys in resilienceConnections and capabilityFilter

Enea-Dervishi:
## Summary

- `next-intl` rejects message keys containing a literal `.` (reserved for expressing nesting). This broke every dashboard page load (`RootLayout` → `getMessages()`) with an `INVALID_KEY` error across all 43 locale files.
- Only `compliance.eventTypes.*` had a runtime normalizer for this pattern (`normalizeComplianceEventTypes` in `src/i18n/request.ts`). `capabilityFilter.*` and `resilienceConnections.degraded.source.*` did not, and threw on every request.
- Restructured both into proper nested objects across all 43 locale files:
  - `"capabilityFilter.visionMismatch"` → `capabilityFilter: { visionMismatch: ... }` (and 3 siblings)
  - `"degraded.source.database"` → `resilienceConnections.degradedSource: { database: ... }` (and 3 siblings) — kept as a **sibling** key rather than nested under `degraded`, since `degraded` is also used as a plain string message (`t("degraded", {...})`) in the same component and can't be both a string and an object.
- Updated the one call site in `ResilienceConnectionsClient.tsx` from `t(\`degraded.source.${s}\`)` to `t(\`degradedSource.${s}\`)`.
- Updated the matching assertion in `tests/unit/ui/resilience-connections.test.tsx`.

⚠️ base-red inherited: #9985

## Test plan

- [x] All 43 locale JSON files still parse (`scripts/ad-hoc/validate-locale-json.mjs`)
- [x] `node scripts/i18n/check-ui-keys-coverage.mjs` — PASS, all 42 non-EN locales at 99.9–100% coverage
- [x] `npx tsc --pretty false -p tsconfig.typecheck-core.json` — clean
- [x] `npx vitest run --config vitest.config.ts tests/unit/ui/resilience-connections.test.tsx` — 17/17 passing
- [ ] Live browser re-verification of `/login` was attempted but blocked by an environment memory constraint (Turbopack dev server OOM-killed twice on this box independent of this change); static verification above was used instead

### DIFF
--- a/scripts/ad-hoc/find-dotted-i18n-keys.mjs
+++ b/scripts/ad-hoc/find-dotted-i18n-keys.mjs
@@ -1,0 +1,34 @@
+import { readdirSync, readFileSync } from "node:fs";
+import path from "node:path";
+
+const dir = "src/i18n/messages";
+const files = readdirSync(dir).filter((f) => f.endsWith(".json"));
+
+const patterns = new Map();
+
+for (const file of files) {
+  const full = path.join(dir, file);
+  const data = JSON.parse(readFileSync(full, "utf8"));
+
+  function walk(obj, prefix) {
+    if (obj === null || typeof obj !== "object") return;
+    for (const key of Object.keys(obj)) {
+      if (key.includes(".")) {
+        const full = prefix ? `${prefix}.${key}` : key;
+        if (!patterns.has(full)) patterns.set(full, new Set());
+        patterns.get(full).add(file);
+      }
+      const val = obj[key];
+      if (val !== null && typeof val === "object") {
+        walk(val, prefix ? `${prefix}.${key}` : key);
+      }
+    }
+  }
+  walk(data, "");
+}
+
+console.log(`Files scanned: ${files.length}`);
+console.log(`Distinct dotted-key paths found: ${patterns.size}\n`);
+for (const [key, fileSet] of [...patterns.entries()].sort()) {
+  console.log(`${key}  (in ${fileSet.size} files)`);
+}
--- a/scripts/ad-hoc/fix-dotted-i18n-keys.mjs
+++ b/scripts/ad-hoc/fix-dotted-i18n-keys.mjs
@@ -1,0 +1,87 @@
+import { readdirSync, readFileSync, writeFileSync } from "node:fs";
+import path from "node:path";
+
+const dir = "src/i18n/messages";
+const files = readdirSync(dir).filter((f) => f.endsWith(".json"));
+
+const CAP_KEYS = [
+  "visionMismatch",
+  "toolsMismatch",
+  "structuredOutputMismatch",
+  "contextWindowMismatch",
+];
+const DEGRADED_SOURCE_KEYS = ["database", "circuitBreaker", "modelLockouts", "count"];
+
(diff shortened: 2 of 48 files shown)


diegosouzapw:
Thanks for tracking this down — the underlying bug is real and the fix is the right shape.

I dug into `use-intl`'s resolver: `resolvePath` splits a translation key on `.` and walks real nested objects, so a flat key literally named `"degraded.source.database"` was never reachable by the `t(\`degraded.source.${s}\`)` call site — the degraded-connections banner has been falling back to the raw key string instead of showing a translated word since day one, in every locale. Restructuring into real nested objects (matching the existing `normalizeComplianceEventTypes` pattern in `src/i18n/request.ts`) is the correct fix, and I confirmed it actually resolves the two affected namespaces cleanly with your own `find-dotted-i18n-keys.mjs` script (only the pre-existing, already-normalized `compliance.eventTypes.*` keys remain afterward).

One correction worth making to the PR description: `use-intl`'s dotted-key validator only `console.error`s by default (it never throws), and it's entirely stripped out of the production bundle by dead-code elimination — so this never actually crashed a page load in production. The real, verified bug is the silently-wrong UI text, not an outage. Worth softening that framing since it'll live in the commit history.

Two things before this can land:
1. The branch currently shows as conflicting against the base — needs a rebase, with a careful hand-resolution across the locale files so no dotted keys sneak back in during the merge.
2. 41 of the 43 touched locale JSON files lost their trailing newline in this diff (probably from the script's `writeFileSync` not adding one back) — a quick `npx prettier --write src/i18n/messages/*.json` will fix that.

Small optional nice-to-have: the existing test mocks `t()` to just echo the key back, so it checks the call site but not the actual JSON shape against real next-intl resolution — an assertion using the real translator against `en.json` would make this a stronger regression guard going forward.

Also flagging for the maintainer's awareness (not something for you to act on): there's another open PR touching the exact same files with the same fix strategy, filed a couple hours after this one — will need coordination on which one lands.

Enea-Dervishi:
Superseded — closing.

While rebasing onto the current `release/v3.8.50` tip, I found that the underlying bug this PR fixes is already resolved on base via `be04c7aec` ("fix(radar): reconcile cumulative release integration"). Confirmed by running this PR's own `scripts/ad-hoc/find-dotted-i18n-keys.mjs` against `origin/release/v3.8.50`: only the pre-existing, already-normalized `compliance.eventTypes.*` keys remain — `capabilityFilter.*` and `resilienceConnections.degraded.source.*` are both already unflattened.

Note the structure differs slightly from this PR's approach for the `degraded` key:
- Base: `resilienceConnections.degraded` becomes `{ message, source: {...} }` (source nested under degraded)
- This PR: kept `degraded` as a flat string, added a sibling `degradedSource` object

`capabilityFilter.*` matches this PR's restructuring exactly.

PR #10332 (the other open PR touching these files) is still unmerged, so this landed independently through the reconciliation commit, not through either open PR.

Closing this one rather than rebasing, since the diff would be effectively empty for the locale files. The stronger regression-test idea (asserting against the real translator with en.json, not just a key-echoing mock) is still worth doing as a follow-up — filing separately if useful.

PR #10039, opened 2026-08-10
feat: combo-lane awareness + activation UX + MCP visibility (Wave 2 of #9654)

branben:
## Summary

This pull request delivers the second wave of #9654 (per-connection virtual admission
lanes; the first wave was merged in #9940). It comprises three units that extend the lane
machinery to account for multi-target fan-out, provide an operator-facing activation
control, and expose lane state to MCP clients:

- **Lane-aware admission probes for combo, fusion, and chaos dispatch.** Each fan-out
  target is checked against the admission lanes before dispatch, so fan-out traffic is
  subject to the same capacity accounting as direct requests. Probes do not enqueue
  (`maxWaitMs` is fixed at 0; historical incidents #4165, #3825, and #4038 documented
  failure modes of queue-wait knobs), price requests at their real dispatch class
  (streaming vs. non-streaming), and fail open: an admission-layer error during a probe
  does not abort the fan-out.
- **Activation control for the adaptive lanes.** A new `OMNIROUTE_CHAT_VIRTUAL_LANES`
  feature flag follows env-wins precedence (environment variable overrides any database
  override, which overrides the default). At boot, the effective value is folded into the
  runtime through its existing `options.env` seam; no `process.env` mutation and no
  changes in `open-sse`. The flag is documented in `.env.example` and
  `docs/reference/ENVIRONMENT.md`.
- **MCP visibility of lane state.** The `omniroute_get_health` tool now returns a subset
  of the live admission snapshot (`adaptiveAdmission`), including the ten lanes with the
  highest queued cost; the block is omitted when no lane data is available. The MCP README
  additionally documents the lane fields and a navigability audit of the tool catalog.

## Related Issues

- Related to #9654 (second wave of #9940)
- Wave 1 (#9940) merged into `release/v3.8.50`; this branch is based on that tip.

## Validation

- Change type: routing / other (admission layer, MCP)
- [x] Focused suites: probe suites 12/12 with 73/73 adjacent tests; feature-flag suites
      134/134; MCP essential tools 22/22; full MCP Vitest suite 356/359, where the 3
      failures reproduce identically on the release base and are environment-dependent
      (`STORAGE_ENCRYPTION_KEY`, autoCombo task-fitness table)
- [x] `npm run lint`: no new errors; the 4 pre-existing errors in `mcp-server/server.ts`
      are suppressed by `config/quality/eslint-suppressions.json`
- [x] Branch is based on the `release/v3.8.50` tip (which includes #9940); the PR will be
[...]

PR #10322, opened 2026-08-14
fix(chat-body-admission): process-wide budget (#10110)

### LINKED ISSUE #9606: fix(backend): Opencode unable to run 2 or more sessions simultaneously

### LINKED ISSUE #10110: fix(admission): preserve process-wide bounds across byte-level connection lanes

branben:
## Summary

Wave 1 (#9940) shipped the per-connection virtual admission lanes that byte-level chat admission (`src/shared/middleware/chatBodyAdmission.ts`) relies on today. It solved concurrent-session capacity accounting but introduced a defect: the lane map's LRU/TTL eviction could orphan live leases and multiply the effective process-wide budget up to 64×. That defect surfaces as spurious 503s under multi-session load.

This PR collapses the per-session lanes into a **single process-global budget** with **per-key fair scheduling**, eliminating both the multiplication and the orphaned-lease path.

Closes #10110
Closes #9606

## The problem (evidence chain)

The user-visible symptom in #9606 — *"Structurally heavy chat request capacity is busy; retry shortly"* — is emitted by `structuralRejectionResponse()` at `chatBodyAdmission.ts:454`. That function is called only on the structural admission failure path (line 574), when `acquireHeavyWithin()` returns `null`:

```ts
const acquired = await controller.acquireHeavyWithin(
  options.queueMs ?? 0, options.signal, CHAT_LARGE_BODY_BYTES, options.sessionId
);
return acquired
  ? { admit: true, lease: acquired }
  : { admit: false, response: structuralRejectionResponse(503, maxMessages) };
```

`acquireHeavyWithin()` returns null when `#activeHeavy >= maxHeavyInFlight` (line 170). Since `CHAT_MAX_HEAVY_IN_FLIGHT` defaults to 1, any second structural request landing while the first holds the slot immediately exhausts capacity and, with the structural stage's default `queueMs=0`, returns null on the first loop iteration.

The controller used by `admitChatStructure` (line 558–562) is the per-session lane controller from `PerConnectionAdmissionController`. That lane map is the bug:

1. **Capacity multiplication** — each distinct session ID resolved to its own `ChatAdmissionController`, so `CHAT_MAX_HEAVY_IN_FLIGHT` and `CHAT_ADMISSION_MAX_QUEUED_BYTES` were enforced per lane, not process-wide. Default 64 lanes × 4 MiB queued-bytes budget = up to 256 MiB of parked bodies process-wide; a 16 MiB config = up to 1 GiB.
2. **Orphaned leases** — LRU and TTL eviction deleted lane records without checking `activeHeavy`, queued bytes, or parked waiters. A session re-resolving after eviction received a fresh controller with a fresh capacity slot while the old lease still held.

Independent audit (#10110, comment by `imrightguy`) confirmed the pre-auth resource-exhaustion variant: lane identity is selected before API-key validation, so unauthenticated callers with distinct fake credentials could shard capacity before auth rejection.

## What changed

- **`src/shared/middleware/chatBodyAdmission.ts`** — deleted the lane map, LRU/TTL eviction, and per-session `SessionRecord`. `ChatAdmissionController` is now the single process-global budget with per-key FIFO queues + round-robin `dispatchFair()`. `PerConnectionAdmissionController.getController()` returns the same shared instance for every session. `resolveSessionId` survives only as a fairness scheduling key (hashed, opaque in diagnostics) with honest re-scoping docs. `snapshot()` now emits process-wide aggregates.
- **`tests/unit/chat-body-admission-aggregate-10110.test.ts`** — new U6 suite: 6 deterministic tests (LRU-no-mint, TTL-no-mint, shared byte budget, 16 MiB config, same-session recreation, round-robin fairness). RED on `release/v3.8.50`, GREEN post-fix.
- **`tests/unit/per-connection-admission-9654.test.ts`** — rewrote the tests that encoded the defect (per-session isolation) to assert the global-budget contract.
- **`docs/reference/ENVIRONMENT.md`** — `OMNIROUTE_CHAT_ADMISSION_MAX_QUEUED_BYTES` documented as process-wide; `OMNIROUTE_CHAT_VIRTUAL_TTL_MS`/`OMNIROUTE_CHAT_VIRTUAL_MAX_SESSIONS` deprecated no-ops.

## Validation

- 75/75 admission tests pass across the 5 core admission files.
- `tsc --noEmit` clean on all 4 touched files (4,229 pre-existing errors project-wide, 0 introduced).
- Pre-commit hooks pass (lint, prettier, docs-sync, tracked-artifacts).

## Review findings disposition (from prior review cycle)

- **B1 (route absent from fix surface)** — resolved: `getController()` always returns the global controller, so the route's `sessionId` no longer forces per-connection selection. One residual boundary noted: `admitChatStructure` without `sessionId`/`controller` falls back to `defaultAdmissionController` (pre-existing, same cap, unreachable from route).
- **B2 (fairness)** — preserved via round-robin per-key dispatch; the fairness test asserts `b1` lands at index 1, not starved.
- **B3 (retire-not-evict)** — superseded by the stronger outcome: no lanes exist to orphan.
- **B4 (pre-auth sharding)** — the multiplication is gone; the `sha256` scheduler key stays with honest re-scoping.

## Wave 2

This PR is the foundation that Wave 2 (#10039) — combo-lane awareness, activation UX, MCP health visibility — builds on top of. All Wave 2 lane-probe and visibility logic rides on the corrected admission layer.

## Files changed

- `src/shared/middleware/chatBodyAdmission.ts`
- `tests/unit/chat-body-admission-aggregate-10110.test.ts` (new)
- `tests/unit/per-connection-admission-9654.test.ts`
- `docs/reference/ENVIRONMENT.md`

### DIFF
--- a/changelog.d/fixes/10322-process-wide-admission-budget.md
+++ b/changelog.d/fixes/10322-process-wide-admission-budget.md
@@ -1,0 +1,1 @@
+- **fix(chat-body-admission):** restore a single process-wide admission budget — heavyweight leases and queued bytes are now bounded once for the whole process instead of per session, so one session can no longer mint extra capacity or starve others; per-session fairness is preserved via round-robin dispatch ([#10110](https://github.com/diegosouzapw/OmniRoute/issues/10110))
--- a/docs/reference/ENVIRONMENT.md
+++ b/docs/reference/ENVIRONMENT.md
@@ -1476,9 +1476,9 @@ These settings were introduced after the previous environment-contract snapshot.
 | Variable | Default | Source File | Description |
 | --- | --- | --- | --- |
 | `OMNIROUTE_CHAT_ADMISSION_QUEUE_MS` | `2000` | `src/shared/middleware/chatBodyAdmission.ts` | Maximum wait for a heavyweight chat admission slot before a retryable `503`; a short bounded wait serializes agent bursts instead of an instant `503`. `0` restores immediate rejection. |
-| `OMNIROUTE_CHAT_ADMISSION_MAX_QUEUED_BYTES` | `4194304` (4 MB) | `src/shared/middleware/chatBodyAdmission.ts` | Queued-bytes budget for the admission wait (#9654): bounds total buffered body bytes parked per lane so the wait cannot amplify the heap (#4380). Over-budget waits receive a retryable `503` immediately. |
-| `OMNIROUTE_CHAT_VIRTUAL_TTL_MS` | `60000` (60 s) | `src/shared/middleware/chatBodyAdmission.ts` | Per-connection virtual admission lanes (#9654): idle-lane eviction TTL. |
-| `OMNIROUTE_CHAT_VIRTUAL_MAX_SESSIONS` | `64` | `src/shared/middleware/chatBodyAdmission.ts` | Per-connection virtual admission lanes (#9654): max concurrent sessions (lanes). |
+| `OMNIROUTE_CHAT_ADMISSION_MAX_QUEUED_BYTES` | `4194304` (4 MB) | `src/shared/middleware/chatBodyAdmission.ts` | Queued-bytes budget for the admission wait: bounds total buffered body bytes parked process-wide so the wait cannot amplify the heap (#4380). Over-budget waits receive a retryable `503` immediately. |
+| `OMNIROUTE_CHAT_VIRTUAL_TTL_MS` | `60000` (60 s) | `src/shared/middleware/chatBodyAdmission.ts` | Deprecated no-op since #10110: per-session admission lanes were removed in favor of one process-wide budget. Accepted for configuration compatibility; ignored. |
+| `OMNIROUTE_CHAT_VIRTUAL_MAX_SESSIONS` | `64` | `src/shared/middleware/chatBodyAdmission.ts` | Deprecated no-op since #10110: per-session admission lanes were removed in favor of one process-wide budget. Accepted for configuration compatibility; ignored. |
 | `OMNIROUTE_RUNNOW_TIMEOUT_MS` | `30000` | `src/app/api/jobs/[id]/run-now/route.ts` | Bounds how long a run-now call waits for an in-flight job before starting the queued run. |
 | `ADOBE_FIREFLY_BROWSER_REFRESH` | enabled | `open-sse/services/adobeFireflySession.ts` | Keeps IMS and browser-risk state fresh through account-scoped Chrome CDP sessions; set `0` to disable. |
 | `ADOBE_FIREFLY_SESSION_DISK` | enabled | `open-sse/services/adobeFireflySession.ts` | Persists repaired Adobe sessions under `DATA_DIR`; set `0` for memory-only state. |
--- a/src/shared/middleware/chatBodyAdmission.ts
+++ b/src/shared/middleware/chatBodyAdmission.ts
@@ -7,19 +7,16 @@
  * local heavyweight capacity before parsing and enforces the hard limit against bytes read,
  * not an untrusted Content-Length header.
  *
- * Per-connection virtual admission lanes (#9654): each distinct API-key (or anonymous)
- * bucket gets its own FairCostQueue so one connection cannot exhaust heavyweight capacity
- * and starve others. Idle sessions are auto-evicted after a TTL.
+ * Process-wide admission budget (#10110): ALL requests — every API key, every
+ * session — contend for ONE global heavyweight budget, so the documented
+ * "in one process" bound holds against fake-credential sharding. Per-request
+ * session identity is used only as a fairness scheduling key: waiters are
+ * grouped per session and served round-robin against the shared budget, so one
+ * connection's burst cannot starve others (#9654).
  */
 
 import { CORS_HEADERS } from "../utils/cors";
 import { createHash } from "crypto";
-
-
-const OMNIROUTE_CHAT_VIRTUAL_TTL_MS = parsePositiveInt(
-  process.env.OMNIROUTE_CHAT_VIRTUAL_TTL_MS,
-  60_000
-);
 
 function parsePositiveInt(value: string | undefined, fallback: number): number {
   const parsed = Number.parseInt(String(value), 10);
@@ -109,6 +106,12 @@ export interface ChatAdmissionLease {
   release(): void;
 }
 
+/** A parked waiter, grouped by fairness key for round-robin dispatch. */
+interface AdmissionWaiter {
+  readonly key: string;
+  readonly resolve: () => void;
+}
+
 /**
  * Process-local heavyweight reservation. The capacity check and increment execute in one
  * synchronous JavaScript turn, making acquisition atomic within an OmniRoute process.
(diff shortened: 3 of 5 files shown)


diegosouzapw:
Reviewed this against the release/v3.8.50 tip — this is a clean fix and I verified it directly.

I checked out the branch, ran both touched test files against a fresh checkout: `chat-body-admission-aggregate-10110.test.ts` and `per-connection-admission-9654.test.ts` — 19/19 passing, including the LRU-no-mint, TTL-no-mint, shared-byte-budget, and round-robin-fairness cases that reproduce the exact defect described in #10110. I also confirmed no other module references the removed `PerConnectionAdmissionController` APIs (`sessionCount`, the old per-session `snapshot()` shape), so the surface change is safe, and that `snapshot()` isn't wired into any dashboard/monitoring route yet — so there's no cross-layer gap to fill here; that's rightly deferred to Wave 2 (#10039).

The design is a real simplification, not just a patch: collapsing the per-session lane map into one process-global controller with per-key round-robin dispatch removes the whole class of bug (nothing left to evict, nothing left to orphan) rather than papering over the symptom.

One heads-up, not a blocker: PR #10437 is also touching `chatBodyAdmission.ts` right now (heap-pressure shedding logic) from a different angle — looks orthogonal to this change, but whichever of the two merges second will want a quick rebase check.

No changes requested — this looks merge-ready. Nice work tracing the multiplication bug back through the eviction path.

Issue #10321, opened 2026-08-14
perf(electron): track desktop startup and packaging efficiency


backryun:
## Objective

Track measured, incremental efficiency improvements to the current Electron desktop distribution before considering a framework migration.

This issue is intentionally narrower than #3932: it covers Electron startup/runtime overhead, packaging size, and desktop release build cost. It also complements #7132 without replacing that Windows SQLite correctness investigation.

## Baseline observations

- Hidden/minimized startup still creates a `BrowserWindow` and loads the full Next.js dashboard.
- Electron readiness polling currently calls the full `/api/monitoring/health` endpoint every 500 ms.
- The readiness fetch has no per-attempt timeout, so a connected but stalled response can outlive the outer startup deadline.
- A purpose-built lightweight `/api/health/ping` route already exists for rapid readiness polling.
- The desktop release workflow restores a root `node_modules` cache immediately before `npm ci`, which removes and recreates that directory.
- Electron dependencies are installed with `npm install` even though `electron/package-lock.json` is committed.
- `electron/package.json` maps the standalone `node_modules` directory into the same packaged destination twice.
- The packaged runtime includes the full documentation tree and large optional ML/browser dependency closures.

Measured packaging context from the current checkout:

- `docs/`: about 103 MB uncompressed; about 31.8 MB under a gzip-like estimate.
- Translated changelogs alone: about 72.8 MB uncompressed; about 21.6 MB compressed.
- Heavy optional ML/browser dependency closure: about 130.8 MB compressed by a gzip-like estimate.
- v3.8.49 desktop assets: macOS arm64 DMG about 390.6 MB, Windows installer about 340.4 MB, Linux arm64 deb about 229.3 MB.

## Current checkpoint — 2026-08-20

All eight implementation PRs are now merged into `release/v3.8.50`:

| Stage | PR | Status |
| --- | --- | --- |
| 1 — bounded readiness polling | #10324 | merged |
| 2 — deterministic release setup | #10325 | merged |
| 3 — lazy hidden-start renderer | #10327 | merged |
| 4 — unload-on-close policy | #10328 | merged |
| 5 — runtime documentation manifest | #10359 | merged |
| 6 — better-sqlite3 Node-API prebuild verification | #10367 | merged |
| 7 — optional ML/browser runtime packs | #10382 | merged |
| 8 — shared platform-neutral release build | #10390 | merged |

[...]

PR #10324, opened 2026-08-14
perf(electron): bound lightweight readiness polling (1/8)

backryun:
Part of #10321 (stage 1).

## What changed
- probe the lightweight auth-exempt /api/health/ping endpoint for Electron startup and restart readiness
- bound each fetch attempt so a stalled response cannot defeat the overall readiness deadline
- require a successful HTTP response instead of accepting arbitrary non-5xx responses
- extract the polling helper into a package-safe module with focused regression tests

## Impact
Electron startup avoids invoking the full monitoring-health aggregation path and no longer hangs indefinitely when a readiness fetch stalls. Remote-server changes, port changes, restarts, initial startup, and background recovery now share the same readiness contract.

## Validation
- node --import tsx/esm --test tests/unit/electron-server-readiness.test.ts tests/unit/electron-main.test.ts tests/unit/electron-packaging.test.ts (41 passed)
- npm run test:vitest (357 passed)
- npm run check:file-size (pass)
- node --check electron/main.js
- node --check electron/lib/serverReadiness.js

The full npm run test:unit sweep was stopped after the repository existing web-cookie external-network tests remained pending; no failure from the changed Electron files was observed. The current base npm run lint is also red in unrelated files (@omniroute/opencode-plugin/src/index.ts and tests/unit/cli-env-inline-comment-10100.test.ts).

### DIFF
--- a/electron/lib/serverReadiness.js
+++ b/electron/lib/serverReadiness.js
@@ -1,0 +1,61 @@
+/**
+ * Pure helpers for polling the embedded or remote OmniRoute server without
+ * importing the Electron main process.
+ */
+
+const DEFAULT_TIMEOUT_MS = 180000;
+const DEFAULT_REQUEST_TIMEOUT_MS = 2000;
+const DEFAULT_POLL_INTERVAL_MS = 500;
+
+function buildReadinessUrl(baseUrl) {
+  return `${baseUrl.replace(/\/+$/, "")}/api/health/ping`;
+}
+
+async function waitForServer(url, timeoutMs = DEFAULT_TIMEOUT_MS, options = {}) {
+  const {
+    fetchFn = globalThis.fetch,
+    requestTimeoutMs = DEFAULT_REQUEST_TIMEOUT_MS,
+    pollIntervalMs = DEFAULT_POLL_INTERVAL_MS,
+    nowFn = Date.now,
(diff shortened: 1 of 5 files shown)


diegosouzapw:
Thanks for this — clean Stage 1 implementation of #10321. A few things stood out in review:

- Consolidating all 5 readiness call sites in `electron/main.js` onto a single `getServerReadinessUrl()` / `waitForServer()` helper is a nice de-dup — previously they were split between polling the root URL and `/api/monitoring/health` inconsistently.
- Extracting `waitForServer` into `electron/lib/serverReadiness.js` with injectable `fetchFn`/`nowFn`/`sleepFn` is exactly what makes the new tests (`tests/unit/electron-server-readiness.test.ts`) deterministic — good call, and I especially like that `tests/unit/electron-main.test.ts` now imports the *real* implementation instead of the old hand-duplicated inline copy that used to drift from the shipped code.
- Per-attempt bounding via `AbortController` while still respecting the overall deadline is correctly implemented, including the edge case where the outer timeout is nearly exhausted.
- All CI checks are green, including both fast-path unit suites and Vitest.

No mandatory changes from my side — this looks merge-ready as-is. One small heads-up for sequencing: Stage 3/Stage 4 in the same tracking issue also touch `electron/main.js` around these call sites, so it'd help to land the stages roughly in order to minimize rebase friction, but that's an operational note, not something to change in this PR.

diegosouzapw:

PR #10325, opened 2026-08-14
ci(electron): streamline release dependency setup (2/8)

backryun:
Part of #10321 (stage 2).

## What changed
- remove the redundant root node_modules cache restored immediately before npm ci
- install Electron dependencies with npm ci --no-audit --no-fund
- synchronize electron/package-lock.json so clean installs work on every release runner
- copy the prepared standalone runtime into resources/app once instead of mapping its node_modules subtree twice
- add workflow and packaging regression guards

## Impact
The release matrix avoids unnecessary node_modules cache transfer/extraction, uses deterministic dependency installation for both package trees, and eliminates duplicate packaging traversal of the standalone dependency tree.

## Validation
- Electron-focused unit tests: 127 passed
- npm run test:vitest: 357 passed
- npm ci --dry-run --ignore-scripts --no-audit --no-fund in electron/: passed
- electron-release.yml YAML parse: passed
- npm run check:file-size: passed

Comparable GitHub Actions timing is not available until this workflow runs; the tracking issue leaves that measurement open rather than treating a local dry-run as release-runner timing.

### DIFF
--- a/.github/workflows/electron-release.yml
+++ b/.github/workflows/electron-release.yml
@@ -93,14 +93,6 @@ jobs:
           node-version: 24
           cache: npm
 
-      - name: Cache node_modules
-        uses: actions/cache@v6.1.0
-        with:
-          path: node_modules
-          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-node-
-
       - name: Install dependencies
         run: npm ci
         env:
@@ -158,7 +150,7 @@ jobs:
 
       - name: Install Electron dependencies
         working-directory: electron
-        run: npm install --no-audit --no-fund
+        run: npm ci --no-audit --no-fund
 
       - name: Build Electron for ${{ matrix.platform }}
         working-directory: electron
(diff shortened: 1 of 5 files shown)


diegosouzapw:
Thanks for this — clean, well-scoped Stage 2 of the #10321 release-efficiency work.

I re-verified this independently rather than just trusting the PR description:
- Ran both touched test files against your branch head: `tests/unit/electron-packaging.test.ts` and the new `tests/unit/electron-release-efficiency.test.ts` — 5/5 pass.
- Confirmed the pre-fix base branch really does have the duplicate `app` / `app/node_modules` extraResources mapping, so the updated packaging test is a genuine regression guard, not a tautology.
- Ran `npm ci --dry-run --ignore-scripts --no-audit --no-fund` in `electron/` against your updated lockfile — resolved cleanly (284 packages, no drift), confirming the lockfile resync is actually in sync with `package.json`.
- Parsed the workflow YAML to confirm it's still valid after the cache-step removal.

All CI checks are green and the merge state is clean. I don't have any blocking items — nice, surgical fix that matches exactly what the tracking issue's Stage 2 checklist describes. The only thing left unmeasured (actual release-runner timing) is something only a real release run can produce, and you've already flagged that honestly instead of guessing from the local dry-run, which I appreciate.

diegosouzapw:

PR #10327, opened 2026-08-14
perf(electron): defer hidden-start renderer creation (3/8)

backryun:
Part of #10321 (stage 3).

## What changed
- start the embedded/remote server and tray without constructing the dashboard BrowserWindow for --hidden, --minimized, and OS hidden-login launches
- create the renderer only on an explicit Open OmniRoute action, tray double-click, second-instance launch, or macOS activation
- restore/show/focus an existing renderer instead of creating duplicates
- keep IPC, updater, remote-server prompt, and tray operation active without a renderer
- keep hidden tray launches alive when their only auxiliary window closes

## Measured impact
Comparable local macOS arm64 development runs used Electron 43.3.0, a fresh isolated user-data directory, --hidden, no running Next server, and an 8-second stabilization period. RSS is the complete Electron process tree.

- before: 6 processes, 709.9 MiB RSS
- after hidden/tray-only: 4 processes, 336.0 MiB RSS
- reduction: 373.9 MiB (52.7%) in this development measurement
- after explicit second-instance Open: 6 processes, 704.5 MiB RSS
- remote-server hidden mode: 4 processes, 335.8 MiB RSS

These are local development measurements, not packaged-release guarantees. They show that the renderer is absent while hidden and created again on demand.

## Validation
- Electron-focused unit tests: 131 passed
- npm run test:vitest: 357 passed
- npm run check:file-size: passed
- node syntax checks for electron/main.js and electron/lib/windowLifecycle.js: passed
- live hidden launch, explicit second-instance open, and remote-server hidden launch: passed

The measurement-only electron/node_modules installation and isolated user-data directories were removed after the run.

### DIFF
--- a/electron/lib/windowLifecycle.js
+++ b/electron/lib/windowLifecycle.js
@@ -1,0 +1,28 @@
+/** Pure helpers for deciding and driving the Electron dashboard window lifecycle. */
+
+function shouldStartHidden({ argv = [], loginItemSettings = {} } = {}) {
+  return (
+    argv.includes("--hidden") ||
+    argv.includes("--minimized") ||
+    loginItemSettings.wasOpenedAsHidden === true
+  );
+}
+
+function showOrCreateWindow({ appReady, getWindow, createWindow }) {
+  if (!appReady) return null;
+
+  const currentWindow = getWindow();
+  if (!currentWindow || currentWindow.isDestroyed()) {
+    return createWindow();
+  }
+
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Deferido para revalidação do head atual. A rodada anterior registrou um hold de ownership, mas a inspeção atual não confirmou esse worktree; por favor, reconfirme o head e os checks antes da decisão final.

PR #10328, opened 2026-08-14
perf(electron): optionally unload renderer on close (4/8)

backryun:
## Summary

- add a persisted tray policy for keeping the dashboard loaded or unloading its renderer on close
- keep the embedded/remote proxy server, tray, and main-process update flow alive in unload mode
- recreate the renderer from tray open, tray double-click, a second app launch, or macOS activation
- restore only same-origin dashboard URLs and rely on Electron's persistent default session for auth state
- keep `keep-loaded` as the default, preserving the existing fast-reopen behavior

Closes the Stage 4 implementation slice tracked by #10321.

## Measured behavior

Measured on macOS arm64 with an isolated Electron profile and a lightweight remote dashboard fixture:

| State | Electron processes | RSS |
| --- | ---: | ---: |
| Renderer loaded | 5 | 531.4 MiB |
| Renderer unloaded | 4 | 436.5 MiB |

- observed reduction while closed: **94.9 MiB** and one renderer process
- renderer recreation: **74 ms** in the persisted-policy validation run
- restored URL: `/dashboard/settings?stage4=1`
- restored session: the second request retained `stage4_session=preserved`

These numbers isolate shell/renderer overhead with a lightweight page; production dashboard totals will vary. The active app has no custom deep-link protocol (`omniroute://` was previously removed because no handler existed), so there is no deep-link state to restore. External-link handling is unchanged. Auto-update listeners remain in the main process and renderer notifications safely no-op while the window is absent.

## Validation

- `node --import tsx/esm --test tests/unit/electron*.test.ts tests/unit/build/electron-release-latest-yml.repro.test.ts` — 134 passed
- `npm run test:vitest` — 357 passed
- `npm run check:file-size` — passed
- `node --check electron/main.js electron/lib/remoteServerPreferences.js electron/lib/windowClosePolicy.js` — passed
- pre-commit formatting, ESLint, docs sync, artifact, and type-budget gates — passed

Full-repository `npm run lint` remains blocked by pre-existing errors in `@omniroute/opencode-plugin/src/index.ts` and `tests/unit/cli-env-inline-comment-10100.test.ts`; neither file is changed here.

### DIFF
--- a/electron/lib/remoteServerPreferences.js
+++ b/electron/lib/remoteServerPreferences.js
@@ -5,8 +5,8 @@ const path = require("path");
 
 /**
  * remoteServerPreferences.js — pure read/write helpers for the small JSON
- * preferences file that persists the operator-configured remote server URL
- * across app restarts (see resolveRemoteServerUrl.js for how it's consumed).
+ * preferences file that persists desktop-shell choices needed before the
+ * server-owned settings database is available.
  *
  * Deliberately a plain flat JSON file rather than the app's SQLite database:
  * this preference must be readable before deciding whether to spawn (or even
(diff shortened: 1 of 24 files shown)


diegosouzapw:
IN-FLIGHT — deferido. O head desta PR está no checkout principal concorrente (perf/electron-unload-on-close). Não faço merge, rebase, push nem alteração nesse checkout.

PR #10330, opened 2026-08-14
fix(sse): close the synthetic keepalive reasoning item + harden output_index allocation

hartmark:
## What Problem This Solves

`RESPONSES_STARTUP_THINKING_FRAME` — the `/v1/responses` early-keepalive placeholder sent while a slow-starting reasoning model produces its first token — opens a synthetic `rs_keepalive` reasoning item at `output_index: 0` and closes its nested summary part (`response.reasoning_summary_part.done`), but never sent `response.output_item.done` to close the item itself. A client tracking open items by `output_index` (as the Responses API spec requires — this is exactly what OpenClaw's own parser does) sees that index still open when the real upstream response later reuses it for its own first output item, and throws a collision: `Responses stream reused active output index 0`.

Live incident (2026-08-13, reliably reproducing by 2026-08-14): traced via a packet capture on the OmniRoute-dev network namespace, correlated against the client gateway journal and 10 separate real request/response pairs — every one wire-clean on the response side, which ruled out provider-side corruption and pointed back at this keepalive frame. Reported upstream against the client before the OmniRoute-side root cause was found: https://github.com/openclaw/openclaw/issues/123342.

## Why This Change Was Made

Two commits:

1. **`fix(sse): close the synthetic keepalive reasoning item's output_item`** — the direct fix: emit `response.output_item.done` for the synthetic item, matching its already-buffered summary text, right after the summary part closes and before the frame ends.
2. **`fix(sse): allocate the keepalive output_index from a stack, not a literal`** — the structural follow-up. `output_index: 0` was hardcoded as a literal across five hand-written events, which is exactly how the missing-close bug happened in the first place: nothing enforced that every open got a matching close, so it silently didn't for months. `ResponsesOutputIndexStack` makes that structural — `open()` allocates the next sequential index, `close()` must name the index being closed and throws if it doesn't match the stack's top, and `assertAllClosed()` throws if anything is still open. The keepalive frame now calls `assertAllClosed()` at module load, so a future regression of this exact shape fails at import/boot time instead of shipping a malformed stream to production.

## User Impact

Fixes a real, reproducing production incident: OpenAI-Responses-API clients (including OpenClaw) using reasoning models through a combo/connection slow enough to trigger the early-keepalive path could get a spurious stream-parse error and a failed turn, indistinguishable from a real provider fault.

## Evidence

- `tests/unit/early-stream-keepalive.test.ts`: updated the frame-shape test to assert the full 5-event closed sequence (fails against pre-fix code — only 4 events — passes after); added a new test that concatenates the keepalive frame with a plausible real subsequent response and asserts no `output_index` collision — the actual failure shape, not just the frame's own internal structure.
- `tests/helpers/assertResponsesOutputIndexLifecycle.ts` (new): a reusable version of the "every added item gets a matching done" invariant, mirroring what a real client's output-index tracker enforces, for replaying a full SSE event sequence rather than just checking one frame's shape. Existing coverage for this bug class (`responses-reasoning-close-before-message-466.test.ts`) only asserted it by hand for one specific emitter (the real translator); nothing generic existed for a hand-rolled synthetic frame like this keepalive to be checked against, which is why its own test could pass while the actual downstream contract still failed.
- `tests/unit/responses-output-index-stack.test.ts` (new): open/close/`assertAllClosed` behavior, including the exact mismatch and never-closed shapes this incident hit.

**Test plan (verified against this repo's actual `release/v3.8.50` code, not a fork-only branch):**
- `early-stream-keepalive.test.ts` and `responses-output-index-stack.test.ts`: all passing
- `tsgo --noEmit`: clean on all touched files


diegosouzapw:
Thanks for the thorough write-up and the packet-capture-level root-causing here — this is a real bug and a clean fix.

Reviewed the diff against `release/v3.8.50`: the pre-fix `RESPONSES_STARTUP_THINKING_FRAME` did indeed stop at `response.reasoning_summary_part.done` without ever closing the output item itself, which lines up exactly with the collision you describe. The fix (explicit `response.output_item.done`) plus the `ResponsesOutputIndexStack` structural guard are both well-scoped and well-tested — the new "concatenate keepalive + real response, replay through the same lifecycle contract a real client enforces" test in particular is a nice touch, since the frame-shape-only test alone wouldn't have caught this.

No changes requested on the code. The one CI red on the PR (`Fast Quality Gates` / dead-code ratchet, "410 > baseline 409") is base-branch drift, not something this PR introduced — the baseline on your branch's base commit is stale (409); the current `release/v3.8.50` tip already carries a newer baseline (415) from an unrelated series that merged after you branched, and your measured count (410) is well under that. It'll clear on its own once this is rebased/merged onto current tip, so nothing for you to do there.

One tiny, purely cosmetic nit (non-blocking): there's an unrelated Prettier reflow of the `SettledHandler` type further down `earlyStreamKeepalive.ts` mixed into this diff — no functional effect, just flagging it in case you want to keep the diff minimal next time.

Nice fix — this is ready to go in.

diegosouzapw:
### Babysit summary

All CI checks are green (every non-advisory check passes). Ready for human review.

**What was wrong:**
- `Fast Quality Gates` was red on `dead-code` (`410` symbols > baseline `409`). This was **base drift, not a PR defect**: the badge's merge-ref was frozen before `release/v3.8.50` rebaselined `deadExports` to `415` (OCR series, 2026-08-14). On this PR's own head (unmerged), `check:dead-code` passes cleanly (248 < 409).

**What I did:**
- `4ae98f89`: merged the current base `release/v3.8.50` (tip `774127be3`, ~60 commits behind) into the PR branch to refresh the badge against the rebaselined baseline. Clean merge, no conflicts (SSE `open-sse/transformer/` / keepalive files untouched by base). `dead-code` now passes (249 < 415).
- Re-ran the cancelled `dast-smoke` run — cancelled is a runner-timeout signature, not an assertion failure; the re-run passed.

**Verification:**
- PR's own tests: 24/24 pass (`tests/unit/early-stream-keepalive.test.ts`, `tests/unit/responses-output-index-stack.test.ts`).
- Local `check:dead-code`, ESLint, Prettier all pass on the PR's changed files.
- No test/gate weakened; no CI edits.

Final `gh pr checks 10330`: all `pass`, no pending. Ready for human review & merge.

hartmark:
Checked the Prettier nit you flagged — turns out the collapsed single-line `SettledHandler` union is actually what this repo's own Prettier config produces (verified: reverting to the multi-line form and running `prettier --write` collapses it right back). So the diff is already in its correct, minimal form — nothing to change here. Thanks for the thorough review either way!

diegosouzapw:

PR #10331, opened 2026-08-14
fix(logging): capture early-keepalive bytes in the call-log artifact

hartmark:
## What Problem This Solves

`withEarlyStreamKeepalive` (the wrapper that keeps `/v1/responses` connections warm while a slow-starting model produces its first token) writes its startup/keepalive/error frames directly into the outer HTTP response stream, entirely outside the request handler's own `reqLogger`. Those bytes never reach `reqLogger.appendConvertedChunk`, so `pipeline.streamChunks.client` in the persisted call-log artifact only ever showed what `chatCore.ts`'s own SSE writer produced — never the keepalive frames actually sent to the client.

This surfaced while diagnosing a live incident (traced upstream as OpenClaw issue #123342, root-caused and fixed on this branch's parent commit `705ac7335`): every call-log artifact for the failing requests looked byte-clean, because the artifact was answering "what did the handler generate," not "what did the client actually receive."

## Why This Change Was Made

`withEarlyStreamKeepalive` wraps the handler's `Promise<Response>` from *outside* its call tree. The `reqLogger` it would need to write into is created deep inside `chatCore.ts`, after routing/model/provider resolution — it doesn't exist yet when the keepalive frames are written, and the two sides share no live reference, only an identifier if one is deliberately threaded through both.

## User Impact

Anyone diagnosing a client-visible streaming defect via the call-log dashboard now sees the actual bytes that went out on the wire during the keepalive window, not just the handler's internal view. No behavior change to the request/response path itself — this is diagnostics-only.

## Evidence

- `route.ts` generates a `correlationId` before calling `handleChat`, passes it as `handleChat`'s existing (previously unused here) 4th positional arg — which `chatCore.ts` already threads into `trackPendingRequest`'s metadata as `entry.correlationId`, zero changes needed there — and into `withEarlyStreamKeepalive`'s options.
- The wrapper buffers every direct-to-client write (startup frame, periodic ticks, in-band error frames) via the new `earlyKeepaliveByteBuffer` module, keyed by that id. The verbatim-forwarded real response body is deliberately **not** re-recorded — the handler's own `reqLogger` already captures that; recording it twice would duplicate it in the artifact.
- `chatCore/attemptLogging.ts`, which already has `correlationId` in scope right where it assembles the final pipeline payload before `saveCallLog`, takes the buffered bytes and prepends them into `streamChunks.client` in send order.
- Buffer entries are consumed exactly once per `correlationId` and swept on a 10-minute TTL, so a request that never reaches the persist call (aborted, detailed logging disabled, a route that doesn't opt in) can't leak entries forever.
- Scoped to `/v1/responses` only — the incident happened there. `/v1/chat/completions` and `/v1/messages` call `withEarlyStreamKeepalive` the same way and would need the identical two-line route change to opt in; left as a follow-up.

**Live proof against a running `omniroute-beta` instance** (real request, real slow combo attempt through `opencode`, no mocking): the persisted artifact for that exact request now contains

```
event: response.output_item.added
data: {"type":"response.output_item.added","output_index":0,"item":{"id":"rs_keepalive",...}}
...
: keepalive
```

in `pipeline.streamChunks.client` — bytes completely absent from the artifact before this change, for a request that genuinely triggered the keepalive path.

**Test plan:**
- `tests/unit/early-keepalive-byte-buffer.test.ts` (new): record/take ordering, single-consumption, per-id isolation, empty-input no-ops, unbounded-growth cap
- `tests/unit/early-stream-keepalive.test.ts`: two new tests — a `correlationId` records the startup frame and keepalive ticks but NOT the forwarded body; omitting `correlationId` is a true no-op
- `tests/unit/attempt-logging-early-keepalive-merge.test.ts` (new): real temp-DB end-to-end proof against the actual persisted call-log row — early bytes prepended in send order, consumed exactly once, no-op without a `correlationId`, gated by `detailedLoggingEnabled` matching the existing `streamChunks` capture gate
- `tests/unit/chatcore-attempt-logging.test.ts` (existing, unchanged): still passing, confirms the merge addition doesn't disturb existing persistence behavior
- 44 passed total across the touched suites, 2 pre-existing skips unrelated to this change
- `tsgo --noEmit`: clean on all touched files

### DIFF
--- a/open-sse/handlers/chatCore/attemptLogging.ts
+++ b/open-sse/handlers/chatCore/attemptLogging.ts
@@ -16,6 +16,7 @@ import { emit } from "@/lib/events/eventBus";
 import type { RequestCompletedPayload, RequestFailedPayload } from "@/lib/events/types";
 import { saveCallLog } from "@/lib/usageDb";
 import { FORMATS } from "../../translator/formats.ts";
+import { takeEarlyKeepaliveBytes } from "../../utils/earlyKeepaliveByteBuffer.ts";
 import { cloneBoundedChatLogPayload, truncateForLog } from "./logTruncation.ts";
 import { attachLogMeta } from "./cacheUsageMeta.ts";
 
@@ -243,6 +244,22 @@ export function persistAttemptLogs(args: PersistAttemptLogsArgs, ctx: PersistAtt
           : {}),
         message: error,
       };
+    }
+    // withEarlyStreamKeepalive writes keepalive/startup/error frames directly
+    // to the client from OUTSIDE this handler's own reqLogger, so they never
+    // reach reqLogger.appendConvertedChunk. correlationId is the only thing
+    // both sides share (see earlyKeepaliveByteBuffer.ts's file doc for why);
+    // merge here, once, right before persistence, prepended in send order.
+    if (detailedLoggingEnabled && correlationId) {
+      const earlyClientBytes = takeEarlyKeepaliveBytes(correlationId);
+      if (earlyClientBytes.length > 0) {
+        const existingStreamChunks =
+          (pipelinePayloads.streamChunks as { client?: string[] } | undefined) ?? {};
+        pipelinePayloads.streamChunks = {
+          ...existingStreamChunks,
+          client: [...earlyClientBytes, ...(existingStreamChunks.client ?? [])],
+        };
+      }
     }
   }
 
--- a/open-sse/utils/earlyKeepaliveByteBuffer.ts
+++ b/open-sse/utils/earlyKeepaliveByteBuffer.ts
@@ -1,0 +1,58 @@
+/**
+ * @file earlyKeepaliveByteBuffer.ts
+ * @description Bridges bytes withEarlyStreamKeepalive writes directly to the
+ * client (outside the request handler's own reqLogger) back into that same
+ * request's persisted call-log artifact.
+ *
+ * withEarlyStreamKeepalive wraps a route's handler Promise from OUTSIDE the
+ * handler's own call tree — it has no reference to the reqLogger the handler
+ * creates deep inside chatCore.ts, and by the time that reqLogger exists the
+ * keepalive/startup frames may already be written. A shared correlationId
+ * (threaded by the route as handleChat's 4th positional arg, and separately
+ * into withEarlyStreamKeepalive's options) is the only thing both sides
+ * share, so recordEarlyKeepaliveBytes/takeEarlyKeepaliveBytes key on that
+ * instead of trying to pass a live object reference across the boundary.
+ *
+ * Entries are consumed once (chatCore/attemptLogging.ts calls
+ * takeEarlyKeepaliveBytes exactly when it assembles the final call-log
+ * payload) and swept on a TTL so a request that never reaches that point
+ * (aborted, detailed logging disabled, a route that never wires this up)
+ * cannot leak buffered bytes forever.
+ */
+
+const MAX_ITEMS_PER_CORRELATION = 200;
+const ENTRY_TTL_MS = 10 * 60 * 1000;
+
+type BufferEntry = { chunks: string[]; createdAt: number };
+
+const buffers = new Map<string, BufferEntry>();
+
+function sweepExpired(): void {
+  const cutoff = Date.now() - ENTRY_TTL_MS;
+  for (const [correlationId, entry] of buffers) {
+    if (entry.createdAt < cutoff) {
+      buffers.delete(correlationId);
+    }
+  }
+}
+
+export function recordEarlyKeepaliveBytes(correlationId: string, chunk: string): void {
+  if (!correlationId || !chunk) return;
+  sweepExpired();
+  let entry = buffers.get(correlationId);
+  if (!entry) {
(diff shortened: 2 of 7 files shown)


hartmark:
Note: this PR is stacked on #10330 (the `RESPONSES_STARTUP_THINKING_FRAME` fix + `ResponsesOutputIndexStack` this logging capture builds on — the correlation-id merge logic needs the fixed `earlyStreamKeepalive.ts` to exist first). Until #10330 merges, the diff below includes its commits too; once it lands and this rebases onto `release/v3.8.50`, the diff will narrow to just the logging-capture change.

diegosouzapw:
Hey, thanks for this — the writeup and test coverage here are genuinely excellent, and the stacked-PR note up front made this much easier to review.

What I found:

- **Feature/architecture**: solves a real gap (keepalive bytes never showing up in the call-log artifact) with a sensible, minimally invasive approach — module-keyed buffer bridging a boundary that can't share a live reference, consumed once, TTL-swept. Scoping to `/v1/responses` only with the chat/messages follow-up called out explicitly is the right call for a first pass.
- **Tests**: ran the four touched test files locally — 35/36 passed on the first combined run. The one failure ("a correlationId records the startup frame and keepalive ticks...") turned out to be a real-timer flake under heavy parallel load, not a logic bug — it passed 20/20 when I re-ran `early-stream-keepalive.test.ts` in isolation. Not blocking, but might be worth fake timers or a wider margin there at some point so it doesn't occasionally flake in a busy CI run.
- **Typecheck**: `npm run typecheck:core` is clean on the branch.

Two things need to land before this can merge:

1. This is stacked on #10330, which is still open — once that merges and this rebases, the diff narrows as you noted. Holding until then.
2. The "Fast Quality Gates" run on this PR is currently failing `check:dead-code` (410 vs baseline 409 — a real regression, not advisory). I wasn't able to pin down the exact symbol in the time I had (my local knip run didn't finish), so could you run `node scripts/check/check-dead-code.mjs` on this branch and either wire in or drop whichever export it flags? Just want to avoid bumping the baseline for something that's genuinely fixable.

Nothing else stood out — solid PR overall, just needs the sequencing + that one gate before it's mergeable.

diegosouzapw:
### Babysit summary

- **Cause of red:** STALE verdict. The failing `Fast Quality Gates` run (dead-code 410>409) was computed against a base 60 commits behind the current `release/v3.8.50` tip.
- **Fix:** refresh-merge of base tip `774127be3` into the PR head → SHA `40c6b9e86f`. No code changes to the contributor's work (credit preserved, author `hartmark`).
- **Gate:** all checks now green on `40c6b9e86f`, including `Fast Quality Gates`, `dast-smoke`, `semgrep`, the 4 unit-test fast-paths, `Vitest`, `No new ESLint warnings`, `Build (advisory)`, `Docs Gates`, `Merge integrity`, `Change Classification`. (`Mergify` rows are NEUTRAL/skipping, not failures.)
- **Tests:** none modified — this was a base refresh only, no logic change.
- PR remains OPEN and MERGEABLE. Ready for human review & merge.

(Note: the orphaned `ArquitetoGlobal` gh login is broken; operation used the active `diegosouzapw` account.)

hartmark:
Checked both outstanding items:

1. **`check:dead-code` regression** — confirmed it was base drift, not a real regression: on the branch as currently rebased onto `release/v3.8.50` tip (`40c6b9e86`), `node scripts/check/check-dead-code.mjs` reports 415 symbols, matching the current baseline exactly. Nothing to fix here — same as what you found for #10330.

2. **Sequencing on #10330** — still open on your end, so this stays blocked on that merge; not something I can resolve from this side.

Also looked at the non-blocking flaky-timer note (`early-stream-keepalive.test.ts`, "records the startup frame and keepalive ticks" real-timer flake under load). Confirmed it's a pre-existing pattern across the whole file (`thresholdMs`/`intervalMs` real-timer margins go back to v3.8.2-era commits), not something this PR introduced — this PR only adds new tests using the same convention as its ~10 siblings already in the file. A proper fix (fake timers, or a coordinated margin change across the whole suite) is a repo-wide test-infra decision, not scoped to this PR, so I'm leaving it as a named follow-up rather than a one-off patch that wouldn't actually change the flake characteristics of the sibling tests using identical margins.

Verified: `npm run typecheck:core` clean on the current head.

hartmark:
Rebased onto current `release/v3.8.50` — both outstanding items from the babysit summary were confirmed already addressed on the prior head, and this rebase carries that forward (head `552712c87`). Two of the three original commits turned out to be fully subsumed by #10330's already-merged `ResponsesOutputIndexStack`/close-output-item work, so only the genuine remainder (early-keepalive byte capture into the call-log artifact) landed as new content.

The two currently-red checks (`Unit Tests fast-path`, `Docs Gates (fast-path)`) are pre-existing base drift, not regressions from this branch:
- `PROXY_LOG_INCLUDE_IPS` is referenced in `src/lib/proxyLogger.ts` but missing from `.env.example` on `release/v3.8.50` itself (from #10348/#10507) — verified directly against the base branch, untouched by this PR's diff.
- `check-docs-counts-sync.mjs` passes cleanly against current base state locally; this PR doesn't touch docs/migrations.

@diegosouzapw ready for another look when you have a moment.

Issue #6365, opened 2026-07-06
feat: native proxy-pool round-robin / egress IP rotation across multiple proxies


diegosouzapw:
### Summary

Community request (Discord EN): support **round-robin / rotation across a pool of proxies** so outbound requests rotate their egress IP (e.g. a new IP each hour / per request).

### Current behavior

Proxy assignments are **one proxy per scope** — `assignProxyToScope()` does DELETE+INSERT per `(scope, scope_id)` and resolution uses a single `LIMIT 1` lookup (`src/lib/db/proxies.ts`, `resolveProxyForScopeFromRegistry` / `resolveProxyForConnectionFromRegistry`). So even if an operator adds several proxies, only one is ever used per scope; there is no rotation across a pool.

Today the only way to rotate egress IPs is to point OmniRoute at a single **external rotating proxy endpoint** and let that provider rotate the IP upstream.

### Proposal

- Allow assigning **multiple** proxies to a scope (global/provider/account) as a pool.
- Add a rotation strategy for the pool: `round-robin` (and optionally `random` / `sticky-per-N-minutes`), so egress IP cycles across the pool.
- Respect the existing alive-filter (`PROXY_ALIVE_PREDICATE`) and the #6246 fail-closed behavior when the pool is empty/dead.

### Notes

_Filed from community triage; sanitized (no personal data)._

diegosouzapw:
Confirmed against `release/v3.8.46`: proxy assignment is one-per-scope (`UNIQUE(scope, scope_id)`, single `LIMIT 1` resolution), so a pool never rotates today.

> Plan: allow assigning multiple proxies to a scope as a pool + a rotation strategy (`round-robin`, optionally `random`/`sticky-per-N-min`), respecting the existing `PROXY_ALIVE_PREDICATE` and the #6246 fail-closed behavior when the pool is empty/dead. Requires a schema migration to lift the single-assignment constraint.

**Status:** ✅ Accepted — implementation in progress. PR incoming.

diegosouzapw:
[...]

PR #10342, opened 2026-08-14
feat(proxy): non-destructive auto-disable mode for the proxy health scheduler

Gi99lin:
## Summary

`PROXY_AUTO_REMOVE` is currently the only opt-in action the background proxy
health scheduler (`src/lib/proxyHealth/scheduler.ts`) can take on a
consistently failing proxy — and it **deletes** the row. For a manually
maintained proxy pool/chain (#6365 rotation), that's too destructive just to
exclude a temporarily-dead member: an operator running their own multi-proxy
chain wants a dead proxy skipped and automatically brought back once it
recovers, not silently removed from the registry requiring a manual re-add.

This adds `PROXY_AUTO_DISABLE` as a non-destructive sibling flag, reusing the
existing sweep/decision infrastructure rather than adding a parallel system.

## What changes

- `src/lib/proxyHealth/decision.ts` — `decideProxyHealthAction()` gets an
  optional `autoDisable` input (defaults to `false`, so every existing caller
  is unaffected) and a new `"dead"` status value. At the same consecutive
  conclusive-failure threshold (`PROXY_AUTO_REMOVE_AFTER`) it now writes
  `status: "dead"` instead of removing the row when `autoDisable` is on.
  `"dead"` is already one of the statuses the pool/rotation alive-filter
  excludes, so a disabled proxy drops out of the active chain immediately —
  no other code changes needed for exclusion to take effect.
- `src/lib/proxyHealth/scheduler.ts` — wires the new `PROXY_AUTO_DISABLE` env
  flag through, and tracks a `disabled` counter in the sweep summary log. If
  both `PROXY_AUTO_REMOVE` and `PROXY_AUTO_DISABLE` are set, auto-remove
  (destructive) wins.
- Recovery is "free": the sweep keeps probing every registered proxy
  (including `dead` ones) on its normal interval, and the existing
  `outcome === "ok"` branch that re-activates a proxy for `autoRemove` now
  also fires for `autoDisable` — no new recovery code path was needed.
- `src/lib/db/proxies.ts` — `getProxyHealthStats()` now also returns each
  proxy's live registry `status`, for observability.
- `ProxyStatusBadge.tsx` — was only checking for the literal string
  `"inactive"`, so a `"dead"` proxy would have rendered as a misleadingly
  green "Active" badge in the dashboard. Now checks the full not-alive status
  set (mirrors `PROXY_ALIVE_PREDICATE`).
- `.env.example` / `docs/reference/ENVIRONMENT.md` — new variable documented
  (verified with `node scripts/check/check-env-doc-sync.mjs`, which passes).
- `docs/ops/PROXY_GUIDE.md` — new "Automatic Failure Exclusion for Your Own
  Proxies" section walking through the end-to-end scenario (verified internal
  anchors with `node scripts/check/check-doc-links.mjs`, which passes).
- `tests/unit/proxy-health-auto-disable-decision.test.ts` — 7 new tests for
  the `autoDisable` branches (threshold, precedence vs. `autoRemove`,
  recovery, inconclusive-probe neutrality, and the opt-in default). The
  pre-existing 6-test suite for the `autoRemove`-only behavior
  (`tests/unit/proxy-health-decide-action-6246.test.ts`) is untouched and
  still passes.

## Test plan

- [x] `node --import tsx --test tests/unit/proxy-health-auto-disable-decision.test.ts tests/unit/proxy-health-decide-action-6246.test.ts` — 13/13 pass
- [x] `node --import tsx --test tests/unit/proxy-health*.test.ts tests/unit/*proxy*.test.ts` — 720/722 pass; the 2 failures are pre-existing and reproduce identically on `release/v3.8.50` without this change (a timing-sensitive keepalive-concurrency test and an unrelated `proxySubscription` rule-mode test)
- [x] `npx tsc --pretty false -p tsconfig.typecheck-core.json` — clean, zero errors
- [x] `npx eslint` on all touched files — clean (one pre-existing `no-restricted-imports` warning on `scheduler.ts`'s `@/lib/localDb` import reproduces identically without this change)
- [x] `node scripts/check/check-env-doc-sync.mjs` — in sync
- [x] `node scripts/check/check-doc-links.mjs` — 824 internal links checked, no broken links
- [x] Full pre-commit hook suite (prettier, eslint --fix, docs-sync, `check-t11-any-budget`, tracked-artifacts) passes


diegosouzapw:
Thanks for this — `PROXY_AUTO_DISABLE` is a clean, well-scoped addition. Reusing the existing `decideProxyHealthAction()` policy machinery instead of a parallel system is the right call, and the new `autoDisable` test suite (`tests/unit/proxy-health-auto-disable-decision.test.ts`) covers the branching logic thoroughly (threshold, precedence vs. auto-remove, recovery, inconclusive-neutrality, opt-in default).

Two things came up in CI/local verification that we'll take care of in-place before merging, since they're small and localized:

1. **CI regression**: `getProxyHealthStats()` now returns a `status` field (`src/lib/db/proxies.ts`), but the pre-existing test `tests/unit/db-proxies-crud.test.ts` asserts an exact shape without it — that's the "Unit Tests fast-path (2/4)" failure you're seeing in CI. We'll update the expected object to include the new field.
2. **Edit-form 400 for auto-disabled proxies**: the scheduler now writes `status: "dead"`, but `proxyRegistryFieldsSchema.status` (`src/shared/validation/schemas/proxy.ts`) is still `z.enum(["active", "inactive"])`. Since the dashboard edit form round-trips the current status back into its PATCH payload, opening an auto-disabled proxy in the edit modal and saving currently 400s. We'll extend the enum and adjust the form so operators aren't locked out of editing a proxy your feature just disabled.

The other two CI failures we saw (an `open-sse-typecheck` hit in `catalog.ts` and a `models-catalog-route.test.ts` failure) look like base drift — your branch was cut before #10248 landed on `release/v3.8.50` and touches exactly those files, so a rebase should clear them; we'll confirm before merging.

No changes needed on your end — flagging for visibility. Nice non-destructive design overall.

diegosouzapw:

#10342 foi embarcada no train combinado `8ed587f50410665b2924b372e59f518e420a02bd`, validado em `root@192.168.0.113` (`tomni-proxmox-113`). O log é `/srv/omniroute-train/.claude/worktrees/merge-train-20260817-085938-suite.log`.

Modo FAST verde: typecheck, file-size, complexidade ciclomática e cognitiva, changelog integrity, testes alterados `db-proxies-crud.test.ts` e `proxy-health-auto-disable-decision.test.ts` e Vitest. A suíte full da release continua sendo executada separadamente no tip acumulado.

Obrigado @Gi99lin pela contribuição; o PR segue para merge serializado em `release/v3.8.50` sob a autorização da campanha.

diegosouzapw:
Thank you @Gi99lin for the proxy auto-disable implementation. The contributor fix is preserved; the follow-up tests cover the real dashboard/schema round-trip and the dead-proxy pool behavior. The remaining failed shards are the reproduced Alibaba base-reds on release/v3.8.50. Local merge-train evidence: tomni-proxmox-113, FAST tip bf128e64ee9b9949df2ff2b1e7451db4d07580dd.

PR #10343, opened 2026-08-14
feat(combos): snapshot auto-combo templates with capability-aware weight scoring

swingtempo:
## Summary

Materialize auto-combo templates into static combo snapshots via `POST /api/combos/duplicate`, with differentiated per-model weights based on resolved capabilities (vision, reasoning) and mode-pack scoring profiles.

### Core: `computeSnapshotWeights()` (`open-sse/services/autoCombo/virtualFactory.ts`)
- New top-level function that scores candidates at snapshot time using static signals only (capabilities, tier classification).
- **taskFit**: reasoning + vision capable models score higher when taskFit is weighted.
- **stability**: richer capability sets assumed more stable (capped at 2 capabilities for normalization).
- **tierPriority / costInv**: premium models get bonus; free models get cost-inverted bonus.
- **latencyInv / health / quota**: equal baseline since no runtime telemetry available at snapshot time.
- Scores clamped to [0, 1].

### Route: `POST /api/combos/duplicate` (`src/app/api/combos/duplicate/route.ts`)
- Calls `prepareVirtualAutoComboInputs({ includeResolvedCapabilities: true })` explicitly so capability data is resolved before scoring.
- Passes through scored weights from the virtual combo instead of hardcoding `weight: 1`.
- Normalizes all per-model weights to sum exactly to 100 (floor + remainder distribution).
- Captures mode-pack weights (`config.weightPack`) at snapshot creation time for traceability.
- Rejects unknown template names with a clear 422 error instead of silently passing bad data downstream.

### UI: `AutoComboCatalog.tsx`
- Uses named type `AutoComboTemplate` instead of inline indexed-access `(typeof AUTO_COMBO_TEMPLATES)[0]`.

## Tests (Rule #18)

| File | Coverage |
|------|----------|
| `tests/unit/combos-duplicate-route.test.ts` | Auth gate, input validation (400), success response shape (201), naming convention, weight distribution sum=100, custom strategy, config fields (`sourceAutoCombo`, `weightPack`, version 2), unique name generation, error sanitization (no stack traces) |
| `tests/unit/snapshot-weights.test.ts` | Map structure, taskFit differentiation (reasoning, vision, both), stability scoring by capability count, equal latencyInv baseline, score clamping to max 1, different relative weights for quality-first vs ship-fast mode-packs |

## Example: Weight Differentiation

Before (all models got `weight: 1`):
```
oc/big-pickle → 1.0
felo/felo-chat → 1.0
...
```

After normalization to sum=100, best-coding (quality-first) produces:
```
mimo-v2.5-free    → ~13%  (vision + reasoning — highest score)
big-pickle        → ~8%   (reasoning only)
hy3-free          → ~4%   (no capabilities)
...
```

Different templates produce different weight distributions because `computeSnapshotWeights` applies the mode-pack's dominant factors at snapshot time.

### DIFF
--- a/open-sse/services/autoCombo/virtualFactory.ts
+++ b/open-sse/services/autoCombo/virtualFactory.ts
@@ -20,6 +20,7 @@ import {
   type AutoCategory,
   type AutoTier,
 } from "./suffixComposition";
+import { classifyTier } from "../tierResolver";
 import type { AutoVariant } from "./autoPrefix";
 import { buildFamilyCandidateFilter, type ModelFamily } from "./modelFamily";
 import { getHiddenModelsByProvider } from "@/models";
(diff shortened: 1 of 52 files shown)


swingtempo:
Rebasing.

PR #10370, opened 2026-08-14
fix(monitoring): canonicalize provider aliases in health matrix

### LINKED ISSUE #10192: bug(monitoring): provider health matrix double-counts alias-routed providers
## Summary

Providers that are reached via a **registry alias** (e.g. provider id `nous-research`
with alias `nous`) appear **twice** in provider-health aggregation: once under the
canonical provider id and once under the alias. The alias row shows a degraded
state (circuit-breaker failures accumulated under the alias name) while the
canonical row shows healthy — so the **same physical provider is counted as two
providers**, inflating provider counts and skewing health percentages.

## Observed (live, OmniRoute release/v3.8.50)

`GET /api/monitoring/health`:

```json
"providerHealth": {
  "nous-research": { "state": "CLOSED",    "failures": 0,   ... },
  "nous":          { "state": "HALF_OPEN", "failures": 241, ... }
}
```

`GET /api/providers/health-matrix`:

```json
{ "provider": "nous",          "state": "degraded", "score": 0.8,
  "circuitBreaker": { "state": "HALF_OPEN", "failureCount": 241 },
  "connections": { "total": 0, "active": 0, ... } },

{ "provider": "nous-research", "state": "healthy", "score": 1.0,
  "circuitBreaker": { "state": "CLOSED", "failureCount": 0 },
  "connections": { "total": 1, "active": 1, ... } }
```

There is exactly **one** provider connection (`provider = nous-research`); the
`nous` row exists only because the circuit breaker is keyed by the **alias**
(`nous`) while the connection is keyed by the **provider id** (`nous-research`).

## Root cause

The provider set for health aggregation is built from a **union of multiple
sources without alias normalization**:

`src/lib/monitoring/providerHealthMatrix.ts` (~lines 371–385):

```ts
const providerIds = new Set<string>();
for (const connection of connectionRows) {
  const provider = toString(connection.provider);   // "nous-research"
  if (provider) providerIds.add(provider);
}
for (const breaker of breakerRows) {
  const provider = toString(breaker.name);          // "nous" (the alias)
  if (provider) providerIds.add(provider);
}
for (const lockout of lockoutRows) { ... }
for (const row of stats) providerIds.add(row.provider);
```

Same pattern in `src/lib/monitoring/providerHealthAutopilot.ts` (~lines 284–301,
connections ∪ breakers ∪ lockouts ∪ quotaSnapshots).

The registry already provides the canonical resolver:
`resolveProviderAlias()` in `open-sse/services/model.ts` (alias → provider id,
transitive with loop guard). It is simply not applied when building these sets,
so alias-keyed breaker/lockout rows produce a phantom provider entry.

## Impact

- **Inflated provider count** in health matrix / autopilot (N providers reported
  when there are really N−1).
- **Skewed health percentages** — the same provider is simultaneously "healthy"
  (id row) and "degraded" (alias row).
- **Misleading operator view**: a provider with real breaker failures shows
  "0 connections" under the alias row while the connection row looks healthy,
  decoupling failures from the connection that owns them.
- Affects any alias-routed provider (registry aliases are common: `nous`,
  `oc`/`opencode`, etc.).

## Suggested fix

Normalize every source through `resolveProviderAlias()` before adding to the
provider set (and before `providerBreakers`/`connectionHealth` grouping in
`observability.ts`), so alias-keyed rows collapse into the canonical provider id
row, which then shows breaker + connection + failures together.

## Environment

- OmniRoute `release/v3.8.50` (also on `main`)
- Single `nous-research` connection, combos routed via `nous/...` alias
- Repro: any alias-configured provider with breaker failures + a live connection

tkgo11:
## Summary

Canonicalize provider aliases before the health-matrix endpoint aggregates connections, circuit-breaker state, model lockouts, and call-log statistics. This collapses alias-keyed and canonical signals into one physical provider row and makes an alias-based `provider` filter return the canonical aggregate.

Closes #10192.

## Validation

- Added a regression fixture with a canonical `nous-research` connection plus alias-keyed `nous` breaker and lockout state.
- Verified the focused health-matrix suite passes.
- Verified core TypeScript checking and focused lint pass.
- Verified `npm run quality:scan:fast` passes all 18 gates.

### DIFF
--- a/src/lib/monitoring/providerHealthAutopilot.ts
+++ b/src/lib/monitoring/providerHealthAutopilot.ts
@@ -1,11 +1,9 @@
 import { createHash } from "crypto";
 
-import {
-  getProviderConnections,
-  updateProviderConnection,
-} from "@/lib/db/providers";
+import { getProviderConnections, updateProviderConnection } from "@/lib/db/providers";
 import { getCachedProviderConnectionById } from "@/lib/localDb";
 import { clearProviderFailure, clearModelLock } from "@omniroute/open-sse/services/accountFallback";
+import { resolveProviderAlias } from "@omniroute/open-sse/services/model";
 
 type JsonRecord = Record<string, unknown>;
 
@@ -111,6 +109,11 @@ function toRecord(value: unknown): JsonRecord {
 
 function toString(value: unknown): string | null {
   return typeof value === "string" && value.trim() ? value.trim() : null;
+}
+
+function canonicalProviderId(value: unknown): string | null {
+  const provider = toString(value);
+  return provider ? (resolveProviderAlias(provider) ?? provider) : null;
 }
 
 function toNumber(value: unknown): number | null {
@@ -253,7 +256,7 @@ export async function buildProviderHealthAutopilotReport(
   const checkedAt = new Date(now).toISOString();
   const includeHealthy = options.includeHealthy === true;
   const includeActions = options.includeActions !== false;
-  const providerFilter = toString(options.provider);
+  const providerFilter = canonicalProviderId(options.provider);
 
   const [{ getAllCircuitBreakerStatuses }, { getAllModelLockouts }, quotaMonitor] =
     await Promise.all([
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Thanks for tracking down the alias-canonicalization root cause and pinning it exactly where #10192 pointed — the fix is clean and the new regression test (real `nous`/`nous-research` alias pair, asserting the collapsed row plus alias-based filtering) is a good permanent guard. I ran the full focused test file plus lint/typecheck on the branch: all green, and the one lint hit on the file (`toNumber` no-restricted-syntax) is pre-existing on release/v3.8.50, not something this PR introduced.

One thing before merging: #10192's root-cause writeup also calls out `src/lib/monitoring/providerHealthAutopilot.ts` as having the exact same unnormalized-aggregation pattern (`connections ∪ breakers ∪ lockouts ∪ quotaSnapshots`), and that file is still untouched here. Since the PR currently says "Closes #10192," merging as-is would auto-close the issue while that second surface is still bugged. Could you either:
- change the PR body to `Refs #10192` so the issue stays open for the autopilot half, or
- extend this same canonicalization to `providerHealthAutopilot.ts` in this PR (or a quick follow-up)?

Either way this is ready to go in — really appreciate the precise, well-tested fix.

diegosouzapw:

PR #10371, opened 2026-08-14
fix(services): use CLIProxy executable on Windows

### LINKED ISSUE #10245: fix(services): append .exe to cliproxy binary spawn path on Windows (extensionless symlink → ENOENT)
## Summary

Even with the correct **Windows** binary installed, the CLIProxyAPI embedded service cannot be spawned on Windows: `spawn C:\Users\<user>\.omniroute\bin\cliproxyapi` fails with `ENOENT`, and because the spawn `'error'` event is unhandled it surfaces as an `uncaughtException` in the server process. Verified on OmniRoute 3.8.49 (Windows 10, Node v24.18.0). This is independent of the platform-selection bug in #10244.

## Root cause

The installer creates a **symlink without a file extension** and the supervisor spawns that path:

```js
// installer
const link = join(bin, "cliproxyapi");        // no ".exe"
try { await unlink(link); } catch {}
await symlink(binary, link);                  // target: ...\cliproxyapi-<v>\cli-proxy-api.exe

// supervisor spawn
spawn(join(bin, "cliproxyapi"), ["--config", configPath]);
```

On Windows, `child_process.spawn` (libuv → `CreateProcessW`) treats an application path without an extension as "append `.exe` and resolve": it looks for `cliproxyapi.exe`, does not find it, and returns `ERROR_FILE_NOT_FOUND` → `ENOENT`. The extracted binary (`cli-proxy-api.exe`) is never reached, so the service can never start on Windows regardless of which build is installed.

## Impact

- CLIProxyAPI can never start on Windows through the embedded-services manager.
- The spawned ChildProcess has no `'error'` handler in the supervisor, so `ENOENT` becomes an `uncaughtException` in the server (observed in `logs/application/app.log`; the server only survives because it has a global `uncaughtException` handler).

## Repro

With a valid Windows build extracted at `~/.omniroute/bin/cliproxyapi-7.2.131\cli-proxy-api.exe` and the extensionless `bin\cliproxyapi` symlink in place:

```js
const { spawnSync } = require("child_process");
spawnSync("C:\Users\<user>\.omniroute\bin\cliproxyapi", ["--version"]);
// => error: ENOENT

spawnSync("C:\Users\<user>\.omniroute\bin\cliproxyapi-7.2.131\cli-proxy-api.exe", ["--version"]);
// => works (prints version)
```

## Workaround (until fixed)

Place a copy of the binary at `bin\cliproxyapi.exe` (keeping the extensionless symlink for version detection via `realpath`). `CreateProcess` then finds `cliproxyapi.exe` when the supervisor spawns `bin\cliproxyapi`.

## Suggested fix

1. On Windows, make the spawn target carry the `.exe` extension — either create the symlink/copy as `cliproxyapi.exe` and spawn that, or spawn the extracted binary path directly.
2. Attach an `'error'` handler to the spawned `ChildProcess` in the service supervisor so a failed spawn is reported as a clean service error instead of an `uncaughtException`.

tkgo11:
## Summary

Select `cliproxyapi.exe` when the embedded CLIProxy service is started on Windows while retaining the existing extensionless command on other platforms. The change is limited to the installer’s spawn-argument resolver; service-supervisor lifecycle handling remains unchanged.

Closes #10245.

## Validation

- Added deterministic Windows platform coverage using the repository’s established platform-mocking pattern.
- Preserved and asserted the native-platform command name and existing configuration-file behavior.
- Verified the focused resolver suite passes.
- Verified core TypeScript checking and focused lint pass.
- Verified `npm run quality:scan:fast` passes all 18 gates.

### DIFF
--- a/src/lib/services/installers/cliproxy.ts
+++ b/src/lib/services/installers/cliproxy.ts
@@ -101,7 +101,8 @@ export async function update(): Promise<InstallResult> {
  * async file I/O is not available here.
  */
 export function resolveSpawnArgs(port: number): SpawnArgs {
-  const symlinkPath = path.join(BIN_DIR, "cliproxyapi");
+  const executableName = process.platform === "win32" ? "cliproxyapi.exe" : "cliproxyapi";
+  const symlinkPath = path.join(BIN_DIR, executableName);
 
   fs.mkdirSync(CONFIG_DIR, { recursive: true });
   const configPath = path.join(CONFIG_DIR, "config.yaml");
--- a/src/lib/versionManager/binaryManager.ts
+++ b/src/lib/versionManager/binaryManager.ts
@@ -97,6 +97,10 @@ async function verifyChecksum(filePath: string, expectedSha256: string): Promise
   return hash.digest("hex").toLowerCase() === expectedSha256.toLowerCase();
 }
 
+function managedBinaryName(): string {
+  return process.platform === "win32" ? "cliproxyapi.exe" : "cliproxyapi";
+}
+
 function findBinaryInDir(dir: string): string | null {
   const candidates = ["cli-proxy-api", "cli-proxy-api.exe", "CLIProxyAPI", "CLIProxyAPI.exe"];
   for (const name of candidates) {
@@ -161,7 +165,7 @@ export async function installVersion(version: string, dataDir?: string): Promise
 
   const binary = await downloadRelease(version, binDir);
 
-  const symlinkPath = path.join(binDir, "cliproxyapi");
+  const symlinkPath = path.join(binDir, managedBinaryName());
   try {
     await fs.unlink(symlinkPath);
   } catch {}
@@ -176,7 +180,7 @@ export async function installVersion(version: string, dataDir?: string): Promise
 
 export async function getCurrentBinaryPath(dataDir?: string): Promise<string | null> {
   const dir = dataDir || DEFAULT_DATA_DIR;
-  const symlinkPath = path.join(dir, "bin", "cliproxyapi");
+  const symlinkPath = path.join(dir, "bin", managedBinaryName());
   try {
     const real = await fs.realpath(symlinkPath);
     return fsSync.existsSync(/* turbopackIgnore: true */ real) ? real : null;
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Thanks for digging into #10245 — the root-cause writeup there is solid and this PR gets half of the fix right.

While reviewing I found that the change is currently incomplete: `resolveSpawnArgs()` in `src/lib/services/installers/cliproxy.ts` now correctly asks to spawn `cliproxyapi.exe` on Windows, but the installer that actually places the binary on disk — `installVersion()` and `rollbackVersion()` in `src/lib/versionManager/binaryManager.ts` (lines ~164 and ~218) — still unconditionally names the Windows copy `cliproxyapi` (no extension). So after this merges, the spawn target computed by `resolveSpawnArgs()` never gets created by the installer, and CLIProxyAPI would still fail to start on Windows — the exact symptom #10245 reports.

I confirmed this with a small integration probe: mocking `process.platform = "win32"`, running the real (unmodified) `rollbackVersion()` against a temp `DATA_DIR`, and then calling this PR's patched `resolveSpawnArgs()` against that same directory. The installer writes `.../bin/cliproxyapi`; `resolveSpawnArgs()` asks for `.../bin/cliproxyapi.exe`; that file never exists. The new "uses the .exe command name on Windows" test doesn't catch this because it only asserts `resolveSpawnArgs()`'s string output in isolation, never that the returned path exists on disk.

To close this out, could you also:
1. Update `installVersion` and `rollbackVersion` in `binaryManager.ts` to name the win32 copy `cliproxyapi.exe` (matching this PR's change), and
2. Update `getCurrentBinaryPath` (line ~179) to resolve that name on win32 too, and
3. Add/extend a test that checks the installed artifact and the spawn command actually line up (e.g. `fs.existsSync(resolveSpawnArgs(...).command)` after a mocked install), so this class of mismatch is caught going forward.

As a small aside, the test file also reformats a few pre-existing `await import(...)` lines that aren't related to the Windows fix — feel free to leave them or revert them, just flagging it so the diff stays focused if you'd rather keep it minimal.

The `cliproxy.ts` change itself is clean and matches the platform-check style already used elsewhere in the installers — once the installer side is aligned this should be good to go. Happy to take another look once updated.

PR #10382, opened 2026-08-14
perf(electron): ship optional ML/browser deps as installable packs (7/8)

backryun:
## What

Stage 7 of the #10321 desktop efficiency roadmap: the optional ML and browser automation dependency closures stop shipping inside the desktop bundle. Packaging now stages them as **checksummed, versioned packs** that users install on demand via the new `omniroute packs` command group.

### Packs

| Pack | Members | Purpose |
| --- | --- | --- |
| `ml-runtime` | `@huggingface/transformers`, `onnxruntime-node`, `@atjsh/llmlingua-2`, `@tensorflow/tfjs`, `js-tiktoken` | LLMLingua-2 prompt compression + transformers.js memory embeddings |
| `browser-runtime` | `playwright`, `playwright-core` | Turnstile solver + ChatGPT/Gemini web executors |

Membership lives in the manifest; exact versions resolve at packaging time into `optional-packs.index.json` (per-member streaming sha256), so dependency bumps never need a manifest edit.

### Changes

- `scripts/build/optionalPackStaging.mjs` — manifest-driven staging pass that relocates pack members out of the Electron standalone tree into `.build/optional-packs/` (dirs + gzip tarballs) and emits `optional-packs.index.json` at the bundle root. Fails **open per member** so a missing optional dep never breaks packaging. Wired in `prepare-electron-standalone.mjs` after the native-module steps.
- `scripts/packs/optionalPackManifest.mjs` — pack definitions, streaming directory checksums, index-entry build/verify.
- `scripts/packs/optionalPackInstaller.mjs` — atomic install/remove/verify against the shipped index (tarball-first, staging-dir fallback, temp-dir + rename swap).
- `bin/cli/commands/packs.mjs` — `omniroute packs list | install <name> | verify [name] | remove <name>` (+ en locale strings).
- `open-sse/utils/optionalPacks.ts` — runtime presence probes (`DATA_DIR/optional-packs/<name>/node_modules`); the LLMLingua worker gate now also probes installed pack trees before reporting the engine unavailable.
- `electron/main.js` — prepends installed pack `node_modules` dirs to the spawned server NODE_PATH, so packs resolve at runtime without living in the bundle.

## Why

- Issue baseline: the heavy optional ML/browser closure is **~130.8 MB compressed** in every desktop asset, though most desktop installs never use LLMLingua compression or Playwright automation.
- Staging-tree measurement (darwin-arm64): the closure is **~534 MB of the 929 MB standalone `node_modules`** (57%): `@tensorflow` 272 MB, `onnxruntime-node` 210 MB, `js-tiktoken` 21 MB, `playwright` + `playwright-core` 24 MB, `@huggingface` 7.7 MB.
- Nothing is lost: `omniroute packs install ml-runtime` restores the exact shipped bits (checksum-verified) without re-downloading the app, and runtime behavior for users who never install packs is unchanged — LLMLingua and web executors already degrade gracefully when their deps are absent.

## Evidence

- TDD regression: an actually installed `ml-runtime` member expressed as `node_modules/@atjsh/llmlingua-2/package.json` failed before the fix because the helper already roots lookups at the pack `node_modules` directory.
- Post-fix optional-pack/staging/installer/artifact-policy/worker-resolution suite: **31/31 pass**.
- Packless core-routing suite: **8/8 pass**.
- Focused optional-pack regression rerun: **10/10 pass**.
- `omniroute packs --help` renders all four subcommands; Prettier and `git diff --check` pass.

### Windows x64 checkpoint

The original physical Windows run passed member-version/SHA-256 index construction, the CLI list/install/verify/remove lifecycle, packless routing, and focused tests. It exposed the LLMLingua `node_modules/node_modules` lookup fixed by this amended PR. The actual installed-pack Windows worker probe against the amended head remains pending; the first-use UI flow and broader platform/architecture matrix remain follow-up gates.

## Reviewer notes

- The correction strips exactly one leading `node_modules` path segment and leaves already-relative member paths unchanged.
- Follow-up work (not in scope): publishing packs next to release assets, first-use/dashboard pack management, broader platform compatibility, and Stage 8 shared-build consolidation.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2398,6 +2398,12 @@ APP_LOG_TO_FILE=true
 # intended to be published as `omniroute-secure`. See SECURITY.md.
 # OMNIROUTE_BUILD_PROFILE=full
 
+# Skip emitting `.tar.gz` tarballs during optional-pack staging for the Electron
+# standalone tree (pack directories + optional-packs.index.json are still produced).
+# Used by the desktop release workflow to trim artifact upload size.
+# Default (when unset): 1 (tarballs emitted). Set to 0 to disable.
+# OMNIROUTE_OPTIONAL_PACK_TAR=1
+
 # Electron smoke harness (used by scripts/dev/smoke-electron-packaged.mjs).
 # ELECTRON_SMOKE_URL=http://127.0.0.1:20128/login
 # ELECTRON_SMOKE_TIMEOUT_MS=45000
--- a/bin/cli/commands/packs.mjs
+++ b/bin/cli/commands/packs.mjs
@@ -1,0 +1,166 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+import { t } from "../i18n.mjs";
+import { resolveDataDir } from "../data-dir.mjs";
+import {
+  EXIT_CODES,
+  emit,
+  exitWith,
+  printError,
+  printInfo,
+  printSuccess,
+  printWarning,
+} from "../output.mjs";
+import { findPack } from "../../../scripts/packs/optionalPackManifest.mjs";
+import {
+  findPackIndexFile,
+  installPack,
+  listPackStates,
+  packState,
+  packsRoot,
+  readPackIndex,
+  removePack,
+} from "../../../scripts/packs/optionalPackInstaller.mjs";
+
+const CLI_DIR = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
+
+/**
+ * Locate + parse the bundle-shipped `optional-packs.index.json`.
+ * Search order: explicit --source dir, then walking up from the CLI module
+ * (bundle installs keep the index at the bundle root), then cwd.
+ */
+function loadIndex(sourceDir) {
+  const indexFile = findPackIndexFile([sourceDir, CLI_DIR, process.cwd()]);
+  if (!indexFile) return { indexFile: null, index: null };
+  return { indexFile, index: readPackIndex(indexFile) };
+}
+
+function stateRow(state, dataDir) {
+  return {
+    pack: state.name,
+    packVersion: state.packVersion,
+    installed: state.installed ? "yes" : "no",
+    verified: state.verified === null ? "-" : state.verified ? "ok" : "FAILED",
+    members: state.members.length,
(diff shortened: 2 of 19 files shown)


diegosouzapw:
Thanks for this — solid continuation of the #10321 roadmap. The pack model (manifest → checksummed staging → CLI install/verify/remove → NODE_PATH injection) is clean, the fail-open/fail-closed split is exactly right (degrade gracefully on absence, refuse on a checksum mismatch), and the test coverage is thorough — tri-state verify, tamper detection, tarball vs. extracted-dir payloads, duplicate-copy cleanup, and the NODE_PATH-ordering wiring assertion are all exercised. All CI checks are green.

Two things before this is ready to merge:

1. `scripts/packs/optionalPackManifest.mjs` ships `STATIC_PACK_IMPORT_ALLOWLIST` / `auditStaticPackImports()` / `isErasedTypeImport()` (the "sharding" section near the bottom of the file) but nothing calls it — no gate script, no test. Could you either wire it into a check script in this PR, or pull it out and land it with whatever stage actually consumes it? As-is it's unused production code.

2. Worth flagging explicitly for release sequencing: this PR only does the *mechanism* — `omniroute packs install` resolves payloads from a local `--source` dir, and publishing the tarballs as release assets is called out as follow-up. That means if a desktop release gets cut from a branch containing this PR before that follow-up lands, real users have no way to actually install `ml-runtime`/`browser-runtime`, so LLMLingua compression and the Playwright web executors go quietly unavailable (fail-open, no crash, but no recovery path either). Not a reason to block this PR, but please make sure the release captain knows not to cut a build until the asset-publishing half ships.

Once #1 is resolved this looks merge-ready to me.

backryun:
Addressed the requested item in commit 79e13842e: removed the unused static-pack import audit and allowlist helpers instead of shipping unconsumed code. The focused optional-pack suite passes 27/27. Please take another look when you have a chance.

diegosouzapw:
Validated in local merge-train `.claude/worktrees/merge-train-20260816-021828-suite.log` on devbox-vm-06-dev002 @ `c559b9b9` (FAST gates green: static + changed tests + vitest; daily full-suite run still required). Merged under batch merge-train protocol (merge-gates §7).

PR #10246, opened 2026-08-13
fix(cli): support native OpenCode JSONC configs

y4ho0:
## Summary

- Add a shared OpenCode config resolver that selects an existing `opencode.jsonc` before `opencode.json`, while retaining `opencode.json` as the creation default when neither file exists.
- Use the selected real path across runtime detection, config generation, dashboard save/apply, and the generated-provider `setup-opencode` command.
- Parse and patch JSONC with `jsonc-parser`, preserving comments, trailing commas, unrelated providers, and unrelated top-level settings.
- Refuse to overwrite invalid non-empty JSONC, and ensure apply/guide flows never create a shadow `opencode.json` when `opencode.jsonc` is active.
- Document both native config filenames and add a user-visible changelog fragment.

## Related Issues

- Closes #10227
- Related to #9985

## Validation

Choose the change type and focused loop from the
[Contribution Golden Path](../docs/ops/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite,
Vitest, the 60% coverage gate, and the production build all run in CI on this PR (#8329):

- [x] Change type: CLI
- [x] Focused tests and category gates from the golden path
- [ ] `npm run lint` — all modified files pass ESLint; an explicit lint of the unchanged base file `tests/unit/cli-env-inline-comment-10100.test.ts:35` reproduces the inherited `no-new-func` failure tracked by #9985
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented below — pending CI

Focused validation completed:

- RED reproduction: 53 passed / 7 failed before the implementation
- GREEN regression suite: 74 passed / 0 failed
- Adjacent OpenCode/runtime regression suite: 32 passed / 0 failed
- Build/module integrity suite: 29 passed / 0 failed
- `npm run typecheck:core`
- Modified-file ESLint and `node --check bin/cli/commands/setup-opencode.mjs`
- `npm run check:changelog-integrity`
- `npm run check:cli-i18n`
- `npm run check:pack-policy`
- `npm run build:cli`
- `git diff --check`

## Tests Added Or Updated

- `tests/unit/cli-helper/config-generator.test.ts`
- `tests/unit/cli-helper/tool-detector-opencode-jsonc-10227.test.ts`
- `tests/unit/cli-runtime-detection.test.ts`
- `tests/unit/cli-tools-apply-opencode-jsonc.test.ts`
[...]

PR #10386, opened 2026-08-14
feat(dashboard): opt-in CSP relaxation for VS Code Simple Browser embedding (#10273)

### LINKED ISSUE #10273: feat(dashboard): opt-in CSP relaxation so the dashboard can be embedded in VS Code Simple Browser (OmniCopilot)
## Context

The new distribution channel **OmniCopilot** (VS Code extension, https://github.com/diegosouzapw/OmniCopilot) has an "Open Dashboard" action with an `editor` mode that renders the dashboard inside a VS Code editor tab via the built-in Simple Browser (iframe-based).

Today this shows a blank page because the dashboard responds with:

```
X-Frame-Options: DENY
Content-Security-Policy: … frame-ancestors 'none' …
```

(set globally in `next.config.mjs` — the same file already overrides `frame-ancestors 'self'` for one specific route at ~line 398, so per-route relaxation has precedent.)

## Proposal

Keep `DENY`/`frame-ancestors 'none'` as the default, and add an **opt-in** (env var, e.g. `DASHBOARD_ALLOW_EMBED=vscode`, or settings toggle) that switches the dashboard routes to:

```
Content-Security-Policy: frame-ancestors 'self' vscode-webview:
```

and drops the `X-Frame-Options` header in that mode (XFO has no way to allow a custom scheme; CSP frame-ancestors supersedes it in modern engines).

## Notes

- Scope it to dashboard pages only — API routes never need framing.
- Clickjacking risk stays controlled: default unchanged, opt-in is explicit, and `vscode-webview:` is the scheme VS Code webviews/Simple Browser use as ancestor.
- Once shipped, OmniCopilot's `omnicopilot.dashboardOpen: "editor"` setting lights up end-to-end (the extension already falls back to the external browser).

diegosouzapw:
Closes #10273

## Problem

The dashboard ships `frame-ancestors 'none'` + `X-Frame-Options: DENY` on every route
(`next.config.mjs`), so the VS Code **Simple Browser** — an iframe whose ancestor is a
`vscode-webview:` document — renders a blank tab. That is exactly what the OmniCopilot
extension's `omnicopilot.dashboardOpen: "editor"` mode uses.

## Change

Build-time opt-in `DASHBOARD_ALLOW_EMBED=vscode`.

| | default (unset) | `DASHBOARD_ALLOW_EMBED=vscode` |
|---|---|---|
| HTML pages (`/`, `/dashboard/**`, `/login`, `/docs/**`, `/landing`, …) | `frame-ancestors 'none'` + `X-Frame-Options: DENY` | `frame-ancestors 'self' vscode-webview:`, **no** `X-Frame-Options` |
| API surface (`/api`, `/v1`, `/v1beta`, `/a2a`, `/healthz`, root-level aliases) | `frame-ancestors 'none'` + `X-Frame-Options: DENY` | unchanged — `frame-ancestors 'none'` + `X-Frame-Options: DENY` |

`X-Frame-Options` has no syntax for a custom scheme, and leaving `DENY` alongside a permissive
`frame-ancestors` would still block the frame — dropping it on the page surface is required,
not cosmetic. Only the literal value `vscode` enables the mode; `1`/`true` do not.

### Why `/login` is included

The VS Code webview has its own cookie jar, so an embedded session always starts
unauthenticated and `/dashboard` redirects to `/login` (`src/server/authz/pipeline.ts:188`).
Relaxing only `/dashboard/**` would still paint a blank tab at the login step. Scope was
confirmed as "all HTML pages except the API surface".

### Why the API exclusion list can't rot

`buildSecurityHeaderRules()` generates **two complementary `source` patterns from one prefix
list** — a positive and a negative lookahead over the same alternation — so every pathname
matches exactly one rule: no gap (a page with no security headers) and no order-dependent
overlap. The list itself is *derived* from the `rewrites()` table plus `/api`, `/a2a`,
`/healthz`, so a future root-level API alias is excluded automatically instead of silently
becoming framable. A test asserts that derivation.

## Files

- **new** `scripts/build/dashboardEmbed.mjs` — pure, testable header-rule logic (same pattern as
  `normalizeBasePath.mjs` / `mitm-stub-flag.mjs`)
- `next.config.mjs` — wires it into `headers()`; the default branch emits the *same single rule*
  it always has
- `.env.example` + `docs/reference/ENVIRONMENT.md` — documented (`check:env-doc-sync` green)
- **new** `tests/unit/dashboard-embed-csp-10273.test.ts` — 11 tests

## Validation (Hard Rule #18)

**TDD** — test written first, confirmed RED (`ERR_MODULE_NOT_FOUND`), then GREEN 11/11. The
tests assert **effective** headers, not config shape: a helper replays Next's own matching +
last-wins merge (`resolve-routes.js`, `resHeaders[key] = value`), and one test runs the config
through Next's own `loadCustomRoutes()` validator in **both** modes — CI never builds with the
env set, so without that guard a malformed `source` would only blow up on the operator's machine.

**Live HTTP smoke** — dev server booted with `DASHBOARD_ALLOW_EMBED=vscode`, headers observed on the wire:

```
/login /docs /dashboard /landing /   → frame-ancestors 'self' vscode-webview:   (no X-Frame-Options)
/api/v1/models /v1/models /models
/metrics /healthz /a2a               → frame-ancestors 'none'   + X-Frame-Options: DENY
```

## Gates

`test-discovery` ✅ · `file-size` ✅ · `check:cycles` ✅ · `complexity-ratchets` ✅ (2454/2774,
cognitive 1104/1223) · `typecheck:core` ✅ · `eslint` ✅ · `check:env-doc-sync` ✅ ·
`check:docs-all` ✅ · sibling `next.config` suites ✅ (32/32)

⚠️ base-red inherited: #9985 — `typecheck:noimplicit:core` reports 8 pre-existing TS7006 errors in
`open-sse/utils/usageTracking.ts` and `src/shared/services/cliRuntime.ts`. Neither file is touched
by this PR; both errors were reproduced on a clean `origin/release/v3.8.50` checkout.


diegosouzapw:
Merged `release/v3.8.50` in to clear an inherited red.

**`Unit Tests fast-path (4/4)`** — `models-catalog-route.test.ts:1402` (`duplicateBuiltins[0].custom === true` expected `false`). Not from this PR: the branch was cut at `f06d5f20ed`, and #10383 (`0bd2be05e7`) landed on the release **after** that, realigning the expectation with the #10248 overlay contract. Reproduced on the branch and confirmed passing on the current release tip; 44/44 after the merge.

**`Fast Quality Gates` → `dead-code`** — base-red, not this PR. Same gate fails on #10393, #10390, #10388 and #10382. `knip --reporter json` returns a byte-identical report on `origin/release/v3.8.50` and on this branch, and `scripts/build/dashboardEmbed.mjs` never appears in it (knip's global `ignore` covers `scripts/**`). Zero dead symbols added here.

**`Docs Gates (fast-path)`** — infrastructure: `npm ci` died with `AggregateError [ETIMEDOUT]` fetching `onnxruntime-node`. Rerunning.

diegosouzapw:
Both inherited reds are cleared.

**`dead-code`** — root-caused and fixed on the base by #10396 (merged): `resolveOpencodeConfigDir` in `cliRuntime.ts` was orphaned by #10246 and, worse, diverged from the canonical resolver by one directory level. Isolated by diffing knip 6.32 reports between the baseline commit `97aac6ac6c` (409) and the tip (410). With the base merged in, this branch now reports `DEAD_TOTAL=409`, gate OK.

**`dast-smoke`** — was `The runner has received a shutdown signal` (self-hosted runner infra). It passed on #10396 from the same base, confirming the classification.

Re-validated on the merged base: this PR's 11 tests pass, the sibling `next.config` suites pass (25), `models-catalog-route` passes (44), `typecheck:core` / `file-size` / `test-discovery` green, and the diff is still the same 6 files.

PR #10367, opened 2026-08-14
perf(electron): verify better-sqlite3 v13 Node-API prebuilds instead of source rebuild (6/8)

backryun:
## Summary

- replace the Electron-ABI `node-gyp` source rebuild of better-sqlite3 with a fail-fast Node-API prebuild verification (issue #10321 Stage 6)
- strip `build/`, `deps/`, and `src/` from the staged module so the packaged loader can only resolve the shipped prebuild
- mirror better-sqlite3 `lib/binding.js` selection (including the linux-musl fallback) in `electronRebuildPlan.mjs`, replacing the historical Windows `npx.cmd` spawn plan whose entire code path is now gone
- update the CI step label/comment and the ELECTRON_GUIDE troubleshooting rows that instructed the obsolete rebuild

## Why this is safe

better-sqlite3 >= 13 ships Node-API (`NAPI_VERSION=10`) prebuilds for every platform we package (`darwin`/`linux`/`linuxmusl`/`win32` × `x64`/`arm64`) inside the npm tarball — all 8 files verified present in the installed `better-sqlite3@13.0.3`. Node-API addons are ABI-independent, and the addon binary exports only `napi_register_module_v1` / `node_api_module_get_api_version_v1` (verified with `nm`), so the same `.node` file runs under plain Node and under the packaged app's `ELECTRON_RUN_AS_NODE` server.

The source rebuild existed because better-sqlite3@12 only shipped prebuilds up to `electron-v146`; electron 43 (`NODE_MODULE_VERSION` 148) silently got no binary. v13 makes that obsolete.

## Measured verification (darwin-arm64, electron 43.3.0)

Staged the module exactly as `prepare-electron-standalone.mjs` now stages it (prebuild-only) and exercised the driver under both runtimes:

| Scenario | Runtime | Result |
| --- | --- | --- |
| create + migrate + write/read + close | Node 24 (`modules 137`) | pass |
| reopen CLI-created DB (existing file, data + migration history intact) | Electron 43.3.0 `ELECTRON_RUN_AS_NODE` (`modules 148`) | pass |
| create + migrate + write/read + close | Electron 43.3.0 | pass |
| reopen app-created DB | Node 24 | pass |
| restart (reopen after close) | Electron 43.3.0 | pass |

The loader selected `prebuilds/darwin-arm64.node` in every invocation — the same binary file under both ABIs.

[...]

PR #10359, opened 2026-08-14
perf(electron): prune authoring docs from desktop packages (5/8)

backryun:
## Summary

- prune Electron-only staging payloads that the packaged runtime never reads
- exclude 42 translated changelogs plus future `docs/research` and `docs/superpowers` authoring trees
- preserve canonical OpenAPI, Codex CLI documentation, and translated runtime guides
- report removed file and byte totals during Electron standalone preparation

## Related Issues

- Related to #10321 (Stage 5)

## Validation

- [x] Change type: build-deploy
- [x] Focused tests and category gates from the golden path
- [x] `npm run lint`
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented below

Commands and results:

- `node --import tsx/esm --test tests/unit/electron-packaging.test.ts tests/unit/electron-main.test.ts` — 39 passed
- `npm run test:vitest` — 39 files / 357 tests passed
- `npm run lint` — passed
- `npm run check:file-size` — passed
- `npm run build` — passed
- `node scripts/build/prepare-electron-standalone.mjs` — passed; removed 42 files / 74,455,753 raw bytes
- macOS arm64 `electron-builder` package before and after the staging prune — both passed

Measured DMG result from the same build and packaging configuration:

- before: 629,843,776 bytes
- after: 608,247,523 bytes
- reduction: 21,596,253 bytes (3.43%)

The packaged app retained:

- `Resources/app/docs/openapi.yaml`
- `Resources/app/docs/guides/CODEX-CLI-CONFIGURATION.md`
- `Resources/app/docs/i18n/ko/docs/guides/USER_GUIDE.md`

## Tests Added Or Updated

- `tests/unit/electron-packaging.test.ts`
  - exercises the manifest against a real temporary docs tree
  - verifies required runtime docs remain
  - verifies pruning is idempotent
  - verifies Electron standalone preparation invokes the manifest

## Coverage Notes

- No `src/`, `open-sse/`, `electron/`, or `bin/` runtime source changed.
- The new build helper is covered directly by the updated Electron packaging unit test.

## Reviewer Notes

- Canonical repository docs are never mutated; pruning is limited to the disposable `.build/electron-standalone` staging tree.
[...]

PR #10390, opened 2026-08-14
perf(electron): build the Next standalone once and hydrate natives per leg (#10321 stage 8/8)

backryun:
## Summary

Stage 8 of the Electron efficiency roadmap (#10321): the desktop release matrix currently runs the full Next.js standalone build on **all four legs** (windows, macos-intel, macos-arm64, linux), duplicating the platform-neutral majority of that work four times, and re-exposing every leg to the hosted-runner RAM failure class that took the linux leg out of v3.8.49.

This PR builds the standalone **once** in a new `web-build` job (ubuntu, webpack — the same RAM escape hatch the linux leg already documents) and has each leg restore + hydrate it:

- `scripts/build/standaloneTarball.mjs` — deterministic, dependency-free tar.gz writer/reader (uid/gid/mtime pinned, sorted entries, symlink + exec-bit preservation; GNU-tar interop covered by tests)
- `scripts/build/standaloneManifest.mjs` — byte-level manifest of `.build/next` (sha256 + size + symlink target per entry, plus the archive's own digest); catches artifact-transfer corruption **before** extraction and re-verifies the restored tree byte-for-byte (unlisted/smuggled files rejected)
- `scripts/build/standaloneBundle.mjs` — `pack` / `restore` / `hydrate` CLI
- `scripts/build/hydrateNativeDeps.mjs` — swaps install-machine-forked native optionals (`@img/sharp-*`, `@ngrok/ngrok-*`, `fsevents`) from the leg's own `npm ci` into the restored tree, then asserts the bundled-native closure (koffi triplets, better-sqlite3 prebuilds, wreq-js, onnxruntime — with its documented darwin-x64 exemption) services the leg's `matrix.os`/`matrix.arch` **before** packaging starts
- `.github/workflows/electron-release.yml` — new `web-build` job; legs download → restore → hydrate and skip the per-leg build

## Rollback & failure semantics

- **Rollback**: the legacy per-leg build is kept and runs when the `ELECTRON_SHARED_STANDALONE` workflow_dispatch input is `disabled` (or the shared job was skipped) — a single-input revert to the v-prior behavior.
- **Fail-closed**: if `web-build` ran and failed, legs do not silently fall back to four per-leg builds; the matrix job condition treats it as a hard failure so the regression the shared job exists to surface stays visible.
- **Signing / smoke / packaging steps are untouched** — they consume `.build/next` exactly as before; restore verifies the tree is byte-identical to what `web-build` produced, then hydrate makes it platform-correct.

## Measurements (issue gate 4 baseline)

- Current four-build matrix: **111.4 runner-minutes** per release run (intel 39.7 / arm64 37.5 / win 19.6 / linux 14.6), dominated by 4× duplicated Next standalone builds.
- Post-change expected compute: one ubuntu standalone build (~7 min) + 4× (install-only + restore + hydrate). Hydrate itself is **31 ms-class** (checksums + copies, no compile) — the same class of saving validated in the stage-6 prebuild work (#10367).
- Actual runner-minute/wall-clock comparison will be recorded on the first release run after merge (issue gate 4 completion).

## Local verification

- `node --test tests/unit/build/standalone-bundle.test.ts` — 7/7 (roundtrip incl. symlinks/exec bits, byte determinism, GNU-tar interop, tamper + smuggle detection, forked-native swap, native-closure serviceability + darwin-x64 exemption)
- Full lifecycle smoke against a real dependency tree: pack → restore → tree equality → hydrate (darwin/arm64: copied=3 replaced=2 removed=2) → natives verified
- prettier/eslint/tsc clean (scripts dir is eslint-ignored, same as existing build scripts)
- Workflow YAML validated with both `js-yaml` and `python3 yaml.safe_load`

## Notes for reviewers

- Known base breakage on `release/v3.8.50` (dead-code gate, `open-sse` typecheck, `models-catalog-route` unit) is reproduced on clean `f06d5f20e` with zero repo changes and affects every open PR — not introduced here (same triage as #10359).
- No dependency changes; everything is pure Node stdlib.

Closes the stage-8 implementation gates of #10321 (runner-minute comparison to be appended after the first release run).

### DIFF
--- a/.github/workflows/electron-release.yml
+++ b/.github/workflows/electron-release.yml
@@ -55,9 +55,75 @@ jobs:
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
           echo "✓ Valid version: $VERSION"
 
+  web-build:
+    name: Build shared Next standalone
+    needs: validate
+    # Stage 8 (issue #10321): the four desktop legs used to each run the full
+    # `npm run build` (Next standalone) — ~111 runner-minutes per release just to
+    # produce the same platform-independent bundle four times. This job builds it
+    # once on ubuntu; every leg then restores the byte-verified archive and
+    # re-forks its native optionals (scripts/build/standaloneBundle.mjs).
+    #
+    # Rollback lever: set the repo variable ELECTRON_SHARED_STANDALONE=disabled.
+    # This job then skips, every leg falls back to building its own web bundle
+    # (the legacy step below), and the pipeline behaves exactly like pre-Stage 8 —
+    # no revert needed.
+    if: ${{ !cancelled() && needs.validate.result == 'success' && vars.ELECTRON_SHARED_STANDALONE != 'disabled' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+      - name: Setup Node
+        uses: actions/setup-node@v7
+        with:
+          node-version: 24
+
+      - name: Install dependencies
+        run: npm ci
+        env:
+          NPM_CONFIG_LEGACY_PEER_DEPS: true
+
+      - name: Build Next.js standalone
+        # webpack, not Turbopack, for the same hosted-runner RAM reason as the
+        # linux leg (see the long comment on the fallback step in `build`).
+        env:
+          JWT_SECRET: ci-build-secret-with-sufficient-length-for-validation
+          NODE_OPTIONS: "--max_old_space_size=6144"
+          OMNIROUTE_USE_TURBOPACK: "0"
+        run: npm run build
+
+      - name: Pack standalone bundle
+        # Deterministic tar.gz + byte-level manifest; the manifest embeds the
+        # archive's own sha256 so artifact-transfer corruption is caught before
+        # extraction, and every entry is re-verified after extraction.
+        run: node scripts/build/standaloneBundle.mjs pack --out web-bundle.tar.gz
+
+      - name: Upload shared web bundle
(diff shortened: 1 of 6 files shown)


diegosouzapw:
Thanks for this — Stage 8 closes out the roadmap nicely and the engineering here is solid. I ran the new test suite independently against your head commit (`tests/unit/build/standalone-bundle.test.ts`) and confirmed all 7 cases pass, including the tamper/smuggle-detection and native-fork-hydration paths. The rollback lever (`ELECTRON_SHARED_STANDALONE=disabled`) and fail-closed semantics are exactly the right shape for a change this central to the release pipeline.

One thing worth fixing before merge: `runHydrate()` in `standaloneBundle.mjs` was clearly designed to accept a comma-separated `--arch` list — the doc comment says "the linux leg ships x64+arm64" — but the workflow's `linux` matrix entry only sets `arch: x64`. Since `electron/package.json`'s `build.linux.target` config actually packages both `x64` and `arm64` AppImage/deb in that same job, the new bundled-native serviceability check never actually verifies the arm64 target it was built to protect. Could you change that matrix entry to `arch: x64,arm64`? Should be a one-line fix and the existing test already covers the multi-arch path.

Everything else looks ready as-is. Nice work on the deterministic tar/manifest layer — the zip-slip protection and byte-level tamper detection are a genuinely good addition on top of the runner-minute savings.

backryun:
Addressed the requested item in commit bc5bee555: the Linux release matrix now passes x64,arm64 to the hydration verification, covering both architectures packaged by that job. The standalone bundle suite passes 7/7, and the workflow YAML validates with the updated matrix value. Please take another look when you have a chance.

diegosouzapw:

PR #10397, opened 2026-08-14
fix(sse): dedupe header-budget drop warns by drop-set fingerprint

### LINKED ISSUE #10315: fix(backend): deduplicate repeated upstream-header budget warnings
### OmniRoute Version

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Linux (single-replica under Desktop multi-stream load)

### Node.js Version

### Provider(s) Involved

Codex-hosted model / Anthropic-style upstreams that return large response header sets (same class as the 768-byte budget bug)

### Model(s) Involved

Any streaming target that triggers upstream header budget drops

### Client Tool

Desktop multi-stream (several concurrent SSE sessions)

### Description

**The header-budget drop path logs a fat JSON warn on every SSE response. Under Desktop multi-stream that is a warn storm — dozens of warns per minute, more main-thread `JSON.stringify` work, more log volume, and real errors disappear.**

Same root constant as the quota-header drop bug:

```ts
// open-sse/handlers/chatCore/responseHeaders.ts
export const MAX_FORWARDED_UPSTREAM_RESPONSE_HEADER_BYTES = 768;
```

When anything is dropped:

```ts
log?.warn?.("HTTP", "Dropped upstream response headers that exceeded forwarding budget", {
  budgetBytes: MAX_FORWARDED_UPSTREAM_RESPONSE_HEADER_BYTES,
  forwardedBytes,
  droppedCount: droppedHeaders.length,
  droppedHeaders: droppedHeaders.slice(0, MAX_LOGGED_DROPPED_RESPONSE_HEADERS), // 20
});
```

**Measured behaviour on 3.8.49:**

| Metric | Observation |
|---|---|
| Frequency | **Every** SSE response that exceeds 768 B of candidates (in practice: essentially every Codex-hosted stream) |
| `droppedCount` | **12–26** per stream |
| Log level | **warn** (not debug) |
| Payload | Full structured object including up to **20** `{ name, bytes }` entries |
| Desktop multi-stream | **Dozens of warns per minute** with only a few concurrent sessions |
| Event loop | Extra serialize/format work on the **same** thread already fighting catalog/compression (#9685 / #10117) |
| Operator UX | Cannot find real errors; log backends rate-limit or fill disks |

**This is a separate issue from "headers are dropped."** Even after the budget/priority fix lands, a warn-per-response design is wrong for a condition that is either:

- **Expected/chronic** until config changes → should be once-per-process or debug, or
- **Rare** after the fix → still should not spam at warn with a 20-element array on every stream during a regression.

Default `debugMode: true` on missing settings (separate issue) makes the overall log noise worse; this path is **warn** regardless.

### Steps to Reproduce

1. Run official 3.8.49 with a codex-hosted (or other multi-header) streaming target.
2. Open Desktop (or script) with **multiple concurrent** SSE chat streams (3–8 is enough).
3. Tail gateway logs:
   ```bash
   # count drop warns over 60s of multi-stream traffic
   # expect: one warn per stream open / response header build
   ```
4. Observe repeated:
   ```text
   Dropped upstream response headers that exceeded forwarding budget
   budgetBytes=768 droppedCount=12..26 droppedHeaders=[...]
   ```
5. Note log rate scales with stream concurrency, not with "new" information — the dropped **name set** is usually identical across requests.

### Expected Behavior

1. Do **not** emit a full warn with `droppedHeaders` array on every SSE response.
2. Prefer one of:
   - **Once per process per dropped-name-set** (fingerprint of sorted names) at warn, then silence or debug;
   - **debug** level for per-response detail;
   - **sampled** (e.g. 1/N or at most once per minute per fingerprint).
3. When budgeting is fixed, chronic drops should be rare; the logger must still be safe if a regression returns.
4. Real errors (upstream 5xx, combo failure, auth) remain visible without scrolling past hundreds of identical header-budget warns.

### Actual Behavior

| Item | Actual on 3.8.49 |
|---|---|
| Log per exceeding stream | **1 warn** with up to 20 dropped header entries |
| Under Desktop multi-stream | **Dozens of warns / minute** |
| Level | warn |
| Dedup / sample | **None** |
| Cost | Main-thread structure + stringify + I/O on hot path |
| Signal/noise | Real faults buried |

### Test Impact

Needs a new unit test

### Error Logs / Output

```text
file: open-sse/handlers/chatCore/responseHeaders.ts

per stream (typical):
  warn HTTP Dropped upstream response headers that exceeded forwarding budget
  {
    budgetBytes: 768,
    forwardedBytes: <n>,
    droppedCount: 12..26,
    droppedHeaders: [ { name, bytes }, ... up to 20 ]
  }

Desktop multi-stream: same warn × concurrent streams × turns
→ log storm + extra event-loop work + missed real errors
```

### Screenshots

N/A

### Additional Context

- Companion bug: **768-byte budget drops Codex/Anthropic quota headers** (fix budget/priority — this issue is **logging** only).
- `MAX_LOGGED_DROPPED_RESPONSE_HEADERS = 20` already acknowledges payload size; it does not fix **frequency**.
- Suggested shape: module-level `Set`/`Map` of seen drop fingerprints; warn once; `log.debug` thereafter.

### Validation Plan

- Unit: call `buildStreamingResponseHeaders` 100 times with the same oversized header set → **≤ 1** warn (or debug-only after first).
- Unit: different dropped-name sets can warn again once each.
- Load: multi-stream smoke does not produce linear warn growth per request after the first fingerprint.

lamchun1110:
## What

Fixes #10315.

`buildStreamingResponseHeaders` in `open-sse/handlers/chatCore/responseHeaders.ts` emitted a full warn (with up to 20 dropped header entries) on **every** SSE response whose upstream headers exceeded the 768-byte forwarding budget. The dropped set is usually identical across responses from the same upstream, so under Desktop multi-stream use this produced dozens of identical warns per minute — burying real errors and adding event-loop serialization work.

## Change

- Fingerprint each drop set (sorted, lowercased dropped-header names) in a module-level `Set` (capped at 1000 fingerprints, cleared on overflow rather than growing unbounded).
- Warn **once per unique fingerprint per process**; subsequent identical drops log at debug level.
- Extend `ResponseHeaderLogger` with an optional `debug` method (the injected `defaultLogger` already has one).
- Export `fingerprintDroppedHeaders` and a `resetDroppedHeaderWarnFingerprints` test hook.

## Tests

Added `tests/unit/chatcore-header-drop-warn-dedupe-10315.test.ts`:

1. 100 identical oversized responses → exactly **1 warn + 99 debug** (the validation plan's criterion: 100 identical oversized calls yield at most one warn)
2. A different drop set warns again
3. Fingerprint is order-insensitive to dropped header names
4. Reset hook forgets fingerprints so the same set warns again
5. Responses within budget never warn

Verified locally: new file passes (5/5); related existing suites pass (`chatcore-streaming-response-headers`, `forwarded-header-budget`, `chatcore-nonstreaming-response-headers`, `upstream-response-headers-strip`, `mcp-sse-response-headers-8277`, `middleware-header-strip-5849` — 37/37; `g13-combo-chatcore-golden`, `chatcore-translation-paths` — 73/73); `npm run lint` clean.

## Changelog fragment

Added under `changelog.d/fixes/` with this PR's number.

### DIFF
--- a/changelog.d/fixes/10397-header-budget-warn-dedupe.md
+++ b/changelog.d/fixes/10397-header-budget-warn-dedupe.md
@@ -1,0 +1,1 @@
+- **fix(sse):** the header-budget drop warning fires once per unique dropped-header set instead of on every SSE response (warn-storm fix) ([#10397](https://github.com/diegosouzapw/OmniRoute/pull/10397) — thanks @lamchun1110)
--- a/open-sse/handlers/chatCore/responseHeaders.ts
+++ b/open-sse/handlers/chatCore/responseHeaders.ts
@@ -40,7 +40,10 @@ const DEFAULT_FORWARDED_HEADER_BUDGET_BYTES = 768;
  * module-cache manipulation.
  */
 export function resolveForwardedHeaderBudget(env?: string): number {
-  const parsed = Number.parseInt(String(env ?? process.env.OMNIROUTE_FORWARDING_HEADER_BUDGET_BYTES), 10);
+  const parsed = Number.parseInt(
+    String(env ?? process.env.OMNIROUTE_FORWARDING_HEADER_BUDGET_BYTES),
+    10
+  );
   return Number.isFinite(parsed) && parsed > 0 ? parsed : DEFAULT_FORWARDED_HEADER_BUDGET_BYTES;
 }
 
@@ -56,7 +59,30 @@ const responseHeaderEncoder = new TextEncoder();
 
 type ResponseHeaderLogger = {
   warn?: (tag: string, message: string, data?: Record<string, unknown>) => void;
+  debug?: (tag: string, message: string, data?: Record<string, unknown>) => void;
 } | null;
+
+/**
+ * #10315: the dropped-header set is usually identical across responses from the
+ * same upstream, so warn once per unique drop fingerprint per process, then log
+ * at debug level — a per-SSE-response warn storm buries real errors and adds
+ * event-loop serialization work. Fingerprints are dropped-header-name sets, so
+ * the set stays bounded by the distinct upstream header shapes in practice.
+ */
+const DROPPED_HEADER_WARN_FINGERPRINT_LIMIT = 1000;
+const droppedHeaderWarnFingerprints = new Set<string>();
+
+export function fingerprintDroppedHeaders(dropped: Array<{ name: string; bytes: number }>): string {
+  return dropped
+    .map((header) => header.name.toLowerCase())
+    .sort()
+    .join(",");
+}
+
+/** Test hook: forget already-warned drop fingerprints. */
+export function resetDroppedHeaderWarnFingerprints(): void {
+  droppedHeaderWarnFingerprints.clear();
+}
 
 function responseHeaderWireBytes(name: string, value: string): number {
   return responseHeaderEncoder.encode(`${name}: ${value}\r\n`).byteLength;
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks for the clean fix — this maps directly onto the validation plan from #10315.

I checked out the branch and ran the new test file plus the six related existing suites you named (chatcore-streaming-response-headers, forwarded-header-budget, chatcore-nonstreaming-response-headers, upstream-response-headers-strip, mcp-sse-response-headers-8277, middleware-header-strip-5849): 5/5 new tests pass, 32/32 related tests pass, and `eslint` on the touched files is clean. I also confirmed the only production call site of `buildStreamingResponseHeaders` (in `streamingResponseHeaders.ts`) always uses the default logger, which already implements `.debug`, so this is fully backward compatible with the existing call chain.

No changes requested — this looks merge-ready as-is. Two very minor optional notes for the future, neither blocking:
- The fingerprint set is capped at 1000 entries and clears wholesale on overflow rather than using an LRU/TTL — fine for the reported storm, just something to revisit if the process ever sees a huge number of distinct upstream header shapes over a long uptime.
- The fingerprint is name-set only (ignores byte sizes), so two drop sets with the same dropped names but different sizes collapse into a single warn — this matches your issue's own suggested design, just flagging for awareness.

Nice work isolating this from the header-budget/priority logic itself, keeping the diff focused on the logging path only.

diegosouzapw:

Issue #4954, opened 2026-06-24
[BUG] OpenCode Free: Inconsistent proxy management and lack of account rotation logging


daniij:
### OmniRoute Version

3.8.35

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

v24.15.0

### Provider(s) Involved

OpenCode Free

### Model(s) Involved

any

### Client Tool

any

### Description

When configuring the OpenCode Free provider, there are two major issues related to proxy management and multi-account rotation observability.

1. Inconsistent Proxy Management
Issue: The proxy configuration modal for individual accounts does not align with the global/standard proxy system used elsewhere in the application.

Details: Currently, there is no way to select previously saved or global proxies from a dropdown or managed list. Users are forced to manually enter the proxy details (Host, Port, Credentials) for each account individually, breaking the unified UX.

2. Missing Observability for Account Rotation (Logs)
Issue: While the UI allows adding multiple accounts for rate-limit rotation, the application logs provide no indication of which account/identity is executing the current request.

Details: Because the logs lack account identifiers (e.g., masked tokens or account index numbers), it is currently impossible to verify whether the shuffle/rotation mechanism actually triggers when a rate limit is hit on one of the accounts.

### Steps to Reproduce

1. Navigate to Providers -> OpenCode Free.
2. Click on the shield/proxy icon for any account.
3. Observe that you must type everything manually instead of picking a pre-saved proxy.
4. Run several requests through the provider and check the system logs; observe that the active account context is missing.

### Expected Behavior

1. Proxies: The account proxy modal should allow selecting from a pool of previously saved proxies or use the application's global proxy settings.
2. Logging: System logs should explicitly state which account is being utilized for each request (e.g., [OpenCode Free] Using Account #3 (554122...) or [OpenCode Free] Rate limit reached, rotating to Account #4).

### Actual Behavior

1. Proxy configuration: When clicking the proxy icon for an account under OpenCode Free, a custom popup appears forcing manual inputs (`Host`, `Port`, etc.) instead of providing a dropdown or picker to select from globally saved proxies within the application.
2. Account rotation observability: When executing requests with multiple accounts configured, the application logs do not attach any account identifiers (e.g., indices, masked hashes, or tokens) to the request cycle. This leaves no way to verify if account shifting/shuffling actually works when rate limits are reached.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

[...]

PR #10402, opened 2026-08-14
fix(executors): rotate to the next account on network throws, not just 429

maxmad64bis:
## Summary

If a request to a free "OpenCode"/"MiMoCode" account hit a network error (timeout, connection
reset) instead of a rate limit, the whole request failed immediately — even when the user had
other accounts configured that could have handled it. Only HTTP 429 triggered account rotation;
a network error did not.

Now a network error rotates to the next account too, same as a 429 does. The one nuance: if the
failing account has no proxy of its own, the next few accounts likely share the same network path
and would fail the same way, so we don't hammer through all of them — we cool down and move
straight to an account that has its own proxy (if one exists) instead of retrying doomed accounts.
This is on by default and can be turned off (`NETWORK_ROTATION_SHARED_EGRESS_GUARD` flag) to get
the old immediate-failure behavior back.

Also fixes the same bug in `MimoCodeExecutor`, which had it too but silently — no log at all, so
operators couldn't see why a request failed.

## Related Issues

- Related to #4954 (per-account proxy + rotation)

## Validation

- [x] Change type: provider
- [x] Focused tests: `check:provider-consistency`, `check:provider-assets`,
      `provider-translate-path-golden.test.ts`, plus the executor/rotation/feature-flag test suites
      below — all passing
- [x] `npm run lint` clean on touched files
- [x] Rebased on the current `release/v3.8.50` tip, focused checks rerun after
- [x] New/updated tests included in this PR
- [ ] SonarQube — not yet run, awaiting CI

## Tests Added Or Updated

- `tests/unit/account-rotation.test.ts` (new) — shared rotation logic used by both executors
- `tests/unit/opencode-proxy-rotation-4954.test.ts` / `tests/unit/mimocode-executor.test.ts` —
  network-error rotation, mixed proxy/no-proxy accounts, and the guard-disabled fallback
- `tests/unit/feature-flags-settings.test.ts` — the new flag's definition

## Reviewer Notes

- Both executors previously duplicated the same account-rotation code; it's now shared in
  `open-sse/executors/accountRotation.ts`.
- No secrets or local config involved.


diegosouzapw:
Thanks for this — nice catch, and a clean fix. I confirmed the underlying bug directly in the diff: the previous `opencode.ts` rotation loop called `runWithProxyContext(...)` with no surrounding try/catch, so a network throw (timeout, ECONNRESET) escaped the whole loop and killed the request instead of falling through to the next account, while a 429 correctly rotated. Wrapping that call and gating rotation on `isNetworkErrorRotatable` (only rotate when the failing account has its own proxy, otherwise apply the shared-egress guard) is the right call, and extracting `accountRotation.ts` to de-duplicate the two executors is a good side effect.

I ran the new/updated test files locally (`account-rotation.test.ts`, `opencode-proxy-rotation-4954.test.ts`, `mimocode-executor.test.ts`, `feature-flags-settings.test.ts`) and 116/117 passed; the one local failure traced back to a stale/corrupted sqlite in my own dev environment (unrelated to your code — a DB-read error made the feature flag default to "enabled" ahead of my env override), and your own CI run shows all 4 unit-test shards green, so I'm treating that as a non-issue.

Two very minor, non-blocking notes, no action needed from you:
- The `resolveFeatureFlag()` DB-error path defaults to "enabled" even when an operator explicitly sets the env var to `false` — but that's an existing pattern shared by every other flag wrapper in the file, not something new here.
- The new flag's `descriptionI18nKey` doesn't have a locale-file entry yet in what I could see, but CI's docs/i18n gates are green, so it's likely picked up by a separate sync pass.

The `Build (advisory)` red on your run is the known Turbopack-OOM issue that's currently failing on every PR — unrelated to your change.

Overall this looks merge-ready to me. Nice work, especially on the shared-egress guard nuance and the parity fix for MimocodeExecutor's error-body sanitization along the way.

diegosouzapw:

Issue #7307, opened 2026-07-15
feat(backend): close CI gap — next build never runs on PRs into release/**


diegosouzapw:
## The gap

**`next build` runs only on PR→`main`.** `ci.yml` triggers on `push`/`pull_request` to `main`; PRs targeting `release/**` get `quality.yml`, which runs no bundler at all. As of 2026-07-15 that means **148 open PRs against `release/v3.8.49` have never been validated by the one gate that catches module-boundary breakage.** It surfaces at merge→`main` — i.e. at release time.

Surfaced while investigating #7067, whose `Build` red turned out to be a runner shutdown on a check-run that no longer even applies to the PR.

## Why this specific gate matters

`next.config.mjs:280` sets `typescript.ignoreBuildErrors: true` — **the build does not typecheck.** So its unique value isn't type safety, it's the thing nothing else covers:

- **server-only code leaking into the client bundle** — a `"use client"` component transitively importing a node-only module
- **module resolution at bundle time** for files outside the curated `typecheck:core` scope

There is no ESLint rule (`no-restricted-imports`, `server-only`) covering either — verified against `eslint.config.mjs`.

## Historical evidence

**v3.8.44 → #6155.** `CoolingConnectionsPanel` (#6061) shipped two defects that only `next build` caught:
1. imported `@/components/ui/card`, which does not exist → `Module not found` (typecheck missed it: the path wasn't in the curated scope)
2. being `"use client"`, imported `formatResetCountdown` from `@/lib/localDb` → dragged `ioredis` → `node:net` into the browser bundle → `Can't resolve 'net'`

Both were invisible to the fast-path. This is the incident that put "next build is the ONLY boundary gate" into the project's memory.

**#7065 → v3.8.48 hotfix.** Broken npm tarball, discovered *after publish*, hotfixed ~4h later. Different job (`package-artifact`/`check:pack-boot`, added in response) but the same family: a main-only gate, an incident found past the point of no return.

**Not found:** a documented case of `Build` failing on merge→`main` for a PR that had passed `quality.yml` clean. The last ~60 `ci.yml` runs show 2 `Build` failures, both on a `[defer]` branch opened directly against `main`, both runner cancellations. So v3.8.44 is the real precedent; I'm not padding the case with anything weaker.

## Inventory — what exists only on PR→`main`

| Gate | Class | Note |
|---|---|---|
| **`build`** (`next build`, Turbopack) | **BOUNDARY** | the only bundler anywhere in CI |
[...]

PR #10408, opened 2026-08-14
fix(ci): pin Build (advisory) to a hosted runner with memory provisioning

diegosouzapw:
## The job has never produced a signal

`Build (advisory)` shows red on essentially every PR. It is not the build failing.

Measured over the last 25 `quality.yml` runs (2026-08-14): **not one instance reached a conclusion.** Every sample was one of two states:

| state | what happened |
|---|---|
| `queued`, unclaimed | self-hosted pool is **2 runners** (`omniroute-113-6/7`), both permanently `busy=true`. One job sat queued **2h+** and never got picked up. |
| `failure`, ~5 min | killed mid-build by this workflow's own `cancel-in-progress` concurrency |

All 6 sampled failures are identical:

```
##[error]The runner has received a shutdown signal. …
##[error]Process completed with exit code 143
```

at ~3.5 min into `npm run build`. **Zero OOM, zero build errors.** So the job occupied a runner that `Fast Quality Gates` and `dast-smoke` actually need, and reported a permanent red to every PR author.

## Why hosted, when gap 19 deliberately routed builds to the VPS

Gap 19 left `USE_VPS_RUNNER` governing build-like jobs on the premise that *"the build needs the .113's RAM"*. That premise is falsified by the repo's own evidence:

**`Fast Production Build`** (`build.yml`) runs `npm run build:release` — a **superset** of this job's `npm run build`, plus the CLI bundle and artifact upload — on plain `ubuntu-latest`, and passed **24 of its last 25 runs** in ~15 min.

What it has that this job did not is memory **provisioning**, not a bigger machine:

- a **10 GB swapfile**
- `NODE_OPTIONS=--max-old-space-size=12288`

The swap is the load-bearing half: `--max-old-space-size` bounds only V8's JS heap and never Turbopack's native (Rust) allocation — `build-next-isolated.mjs:188` documents exactly this (#6409). Both settings are mirrored here.

## Change

- `runs-on: ubuntu-latest` (was the `USE_VPS_RUNNER` expression)
- swap step + heap env mirrored from `build.yml`
- comments record the measurement so the next person does not re-derive it

`USE_VPS_RUNNER` keeps its other consumers — `ci.yml`'s Build, `nightly-release-green`, `npm-publish` — so the variable stays meaningful. **Fork safety is strictly improved:** no PR can reach the LAN runner through this job any more.

## Note for a future decision (not made here)

`build.yml` triggers on `push: branches: ["**"]`, so for own-origin PR branches this job duplicates a build that already runs and passes. Its unique coverage is **fork PRs**, where the push lands on the fork and `build.yml` never fires in this repo. If that coverage is judged not worth the hosted minutes, *deleting* the job is the coherent move — not returning it to the self-hosted pool. Left as-is because removing a gate is the maintainer's call.

## Validation

```
check:workflows --ratchet → 186 zizmor findings, baseline 190, no regression
prettier --check          → clean
YAML parse + job shape    → runs-on=ubuntu-latest, 6 steps, build env carries both memory vars
```

Workflow-only change; no production code touched.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -60,7 +60,14 @@ jobs:
   build:
     name: Build (advisory)
     needs: changes
-    if: ${{ github.event_name != 'pull_request' || ((github.event.pull_request.draft == false || startsWith(github.head_ref, 'mergify/merge-queue/')) && needs.changes.outputs.code == 'true') }}
+    # FORK PRs ONLY. build.yml's `Fast Production Build` triggers on `push: branches: ["**"]`
+    # and runs `build:release` — a superset of this job — so for an own-origin branch this job
+    # was building the same tree twice. A fork contributor pushes to THEIR repo, so that push
+    # never fires here, and this is the only pre-merge build signal they get. Measured
+    # 2026-08-14: 72 of the last 100 PRs into release/** came from forks, so the fork case is
+    # the majority of the traffic, not the exception — this job earns its place, it just should
+    # not duplicate build.yml for the own-origin 28%.
+    if: ${{ github.event_name != 'pull_request' || ((github.event.pull_request.draft == false || startsWith(github.head_ref, 'mergify/merge-queue/')) && needs.changes.outputs.code == 'true' && github.event.pull_request.head.repo.full_name != github.repository) }}
     # PINNED to hosted — this was the last job in THIS workflow still on the USE_VPS_RUNNER
     # switch (ci.yml's Build, nightly-release-green and npm-publish keep it, so the variable
     # stays meaningful), and with USE_VPS_RUNNER=true it produced NO signal at all here.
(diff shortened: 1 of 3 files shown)


diegosouzapw:
## Measured the gate before deciding — deleting the job would have been wrong

The hosted pin is proven: `Build (advisory)` **passed in 12m13s** on this PR, on `GitHub Actions 1000426067`. First conclusion it has produced in a long while.

Two corrections to what I wrote above, both from widening the sample:

**1. It is not literally never-green.** My first window was 25 runs and showed zero conclusions. Over a wider one it reaches `success` on roughly **10-15%** of runs (13/138 on 08-14, 7/53 sampled on 08-13). Chronically unreliable, not permanently dead. Conclusion and fix unchanged.

**2. Deleting it would blind the majority of PRs.** I suggested deletion as the coherent alternative. The traffic says otherwise: **72 of the last 100 PRs into `release/**` come from forks.** A fork contributor pushes to their own repo, so `build.yml` never fires here — this job is their only pre-merge build signal. The fork case is the rule, not the exception.

## What changed instead

Scoped the job to fork PRs (`head.repo.full_name != github.repository`). Own-origin branches already get the superset build from `Fast Production Build` on every push, so they were building the tree twice.

| | before | after |
|---|---|---|
| own-origin PR (28%) | quality.yml build **+** build.yml build | build.yml build only |
| fork PR (72%) | queued for hours / cancelled | hosted build, ~12 min, with swap + heap |

`Build (advisory)` was **18% of the ~69 min of compute** a full quality.yml run costs, so the own-origin skip is a real saving on top of making the fork path work.

## Sibling test realigned

`tests/unit/build/check-workflows.test.ts` (`#7307`) pinned the old self-hosted expression — I missed it on the first pass and it went red. It now asserts the hosted pin, the absence of `self-hosted`/`USE_VPS_RUNNER` in the job's **directives** (the comment legitimately explains why the pool was abandoned, so the scan strips comments), both memory settings, and the fork filter. Mutation-validated: restoring self-hosted, dropping the swapfile, or flipping the fork filter each turns it red.

The other two reds on the previous run were `npm ci` `ETIMEDOUT` on `Merge integrity` (registry flake) — unrelated.

PR #10411, opened 2026-08-14
Sanitize test fixtures, add developer .env guidance, and add gitleaks…

blarovse:
Gitleaks baseline and scan summary
Local authoritative gitleaks scan (pre-PR) results:
Total findings: 430
generic-api-key: 240
curl auth tokens: 146
JWT: 23
GCP: 9
private key: 3
Top files with findings:
.omo/drafts/claude_request.md
.env.example
src/lib/oauth/constants/oauth.ts
open-sse/config/providerRegistry.ts
tests
Action taken in this branch (security/sanitize-fixtures):
Added gitleaks-baseline.json (copied from gitleaks-local.json) at the repository root to serve as the baseline of legacy findings.
Updated .github/workflows/gitleaks.yml to pass --baseline-path=gitleaks-baseline.json so CI detection ignores baseline matches.
Added an allowlist entry for gitleaks-baseline.json to .gitleaks.toml.
Added gitleaks-local.json to .gitignore to avoid accidentally committing the local scan artifact.
Maintainer request / next steps:
Please manually confirm that the four listed files below contain only placeholder or non-sensitive values (and not real credentials):
.omo/drafts/claude_request.md
.env.example
src/lib/oauth/constants/oauth.ts
open-sse/config/providerRegistry.ts
If any contain real secrets, rotate/revoke them immediately and update the baseline after removing secrets from history.
After maintainers confirm, update gitleaks-baseline.json (if appropriate) and commit the updated baseline as part of the repo policy review.
Notes:
The baseline is intended as a pragmatic first step to reduce noise from historical findings; it does not replace the need to remediate real secrets.
The gitleaks CI job will upload a reports/gitleaks-report.json artifact for each PR run; use that to triage new findings.


diegosouzapw:
Thanks for tackling this — sanitizing realistic-looking secrets out of test fixtures and documenting the `.env` postinstall behavior are both genuinely useful. I ran the changed test files directly and all 110 tests in `github-collector.test.ts`, `piiSanitizer.test.ts`, and `sre-redact-logs.test.ts` pass cleanly, so the fixture swaps themselves are safe.

That said, I found a few things that need fixing before this can merge:

1. **The `.gitleaks.toml` addition breaks the project's existing secret-scanning gate.** We already run gitleaks in CI as a blocking ratchet (`scripts/check/check-secrets.mjs`, wired into `ci.yml`/`quality.yml`). Running `node scripts/check/check-secrets.mjs --ratchet` against this branch fails hard:
   ```
   FTL Failed to load config error="gitleaks-baseline-file: both |regex| and |path| are empty, this rule will have no effect"
   ```
   exit code 2 — vs. a clean exit 0 on the base branch. The new `[[rules]]`/`[rules.allowlist]` block isn't valid syntax for the pinned gitleaks version (8.30.1) and makes gitleaks refuse to load the config at all. This would red the blocking `quality-gate` job for every PR if merged as-is.

2. **We already have a gitleaks ratchet mechanism** — `scripts/check/check-secrets.mjs` compares a measured finding count against `config/quality/quality-baseline.json`'s `metrics.secretFindings`. The new `.github/workflows/gitleaks.yml` + root `gitleaks-baseline.json` (9k lines, baselining 430 findings in one shot) is a second, independently-scoped mechanism that will drift from the first one over time, and the blanket-baseline approach runs against our own documented `.gitleaks.toml` policy ("if it's a real finding used in prod, fix it — don't allowlist it"). Could we fold this into the existing gate instead of adding a parallel one?

3. **Repo-root hygiene**: `PR_BODY.md` and `gitleaks-baseline.json` shouldn't live at the repo root — our root is restricted to config/dependency/docs/CI files. If the baseline file is kept, please move it under something like `config/security/`, and drop `PR_BODY.md` entirely (looks like leftover from drafting the PR description).

4. **`README.md`** now has a duplicated tag — `<div align="center"><div align="center">` — please fix that to a single `<div align="center">`.

5. The branch is currently showing as conflicting against `release/v3.8.50` (which has moved quite a bit since this was opened), so it'll need a rebase too.

Happy to take another look once these are addressed — the core fixture-sanitization idea is worth keeping, it just needs the CI-facing pieces reworked so they don't conflict with what's already in place.

diegosouzapw:
Thank you @blarovse for the sanitized-fixtures and developer-environment guidance changes. The exact head passed the prior focused validation and was boarded in the FAST train; no PR-specific red was observed.

Issue #10296, opened 2026-08-13
release(v3.8.51): publish a versioned image containing the /v1/models event-loop fix


RaviTharuma:
Current release-state refresh (2026-08-25): the original `/v1/models` event-loop fix is present on `release/v3.8.51`, but there is still no releasable versioned image.

- `release/v3.8.51` tip is [`a179ffed`](https://github.com/diegosouzapw/OmniRoute/commit/a179ffed5bb2e0b883b9ae7214ce8717b2a94c4d), and `package.json` is `3.8.51`.
- The branch contains the `getModelsDevPricing()` memo/invalidation fix from #10055 in `src/lib/modelsDevSync.ts`.
- GitHub still has no `v3.8.50` or `v3.8.51` tag/release; the newest versioned release remains [`v3.8.49`](https://github.com/diegosouzapw/OmniRoute/releases/tag/v3.8.49).
- Docker Hub still returns 404 for [`:3.8.50`](https://hub.docker.com/v2/repositories/diegosouzapw/omniroute/tags/3.8.50) and [`:3.8.51`](https://hub.docker.com/v2/repositories/diegosouzapw/omniroute/tags/3.8.51). [`:latest`](https://hub.docker.com/v2/repositories/diegosouzapw/omniroute/tags/latest) was last pushed 2026-07-30 and remains the v3.8.49-era image.
- The latest `release/v3.8.51` Docker workflow reached the real current blocker: both architecture builds failed because `useLiveDashboard.ts` imports `resolveLiveWsUrl` / `sanitizeLiveWsPort`, but `wsPath.ts` does not export them ([run 32836613113](https://github.com/diegosouzapw/OmniRoute/actions/runs/32836613113), [amd64 job](https://github.com/diegosouzapw/OmniRoute/actions/runs/32836613113/job/97766935742), [arm64 job](https://github.com/diegosouzapw/OmniRoute/actions/runs/32836613113/job/97766935865)).
- Existing PRs [#11465](https://github.com/diegosouzapw/OmniRoute/pull/11465) and [#11509](https://github.com/diegosouzapw/OmniRoute/pull/11509) address that same missing-export build failure; no duplicate code PR is needed.

Keeping this issue open, but retitling it to the current release target/blocker. Close only after a versioned image containing #10055 is published and its contents/startup behavior are verified.

RaviTharuma:
[...]

PR #9946, opened 2026-08-09
fix(guardrails): vision bridge reroute/pool/self-loop fixes (auto/best-vision, claude-wire base64)

herjarsa:
## Problem

The vision bridge was not routing image requests to a vision-capable model, for individual models AND combos. Verified empirically against the live runtime (v3.8.49 npm):

- `auto/best-vision` resolved to `deepseek-v4-flash-free` (text-only model) — the virtual combo used the flat `smart` variant with NO vision candidate filter.
- Combo `REASONING` + image -> 400 with 3/3 targets attempted (minimax/MiniMax-M3 -> `403 (2013)` for remote image URLs; opencode-zen/kilo targets rejected images outright).
- The describe self-loop failed (`(unavailable)`) — the guardrail's `sk_omniroute` fallback key and the runtime's hooked global fetch (ProxyFetch) broke loopback calls.
- `hasUsableCredentialsForModel("auto/best-vision") === false` (no provider row named `auto`) blocked the reroute entirely.

## Fixes

1. **builtinCatalog**: `auto/best-vision` -> vision CATEGORY (candidate filter by capability), `auto/pro-vision` -> vision + pro tier. New pure `resolveBuiltinAutoSpec()` (tested).
2. **suffixComposition + visionBridgeRouter**: vision candidate pool excludes models whose catalog OVERSTATES vision support (opencode-go/opencode-zen/tokenrouter are forced through the bridge by `isVisionBridgeForcedModel`).
3. **visionBridge.ts**: `auto/*` reroute target is a virtual combo — a missing `auto` provider credential row must never block the reroute (`bestUsable === false && !auto/*`).
4. **visionBridgeHelpers.ts**:
   - `resolveSelfLoopApiKey()`: real DB key (memoized `getOrCreateApiKey`) instead of the `sk_omniroute` sentinel when `REQUIRE_API_KEY` is on.
   - `isClaudeWireFormatModel` + `ensureBase64ImagesForClaudeWire`: claude-wire backends (MiniMax, Z.AI, Kimi) reject remote image URLs (MiniMax 403 2013) — resolve URLs to base64 before rerouting.
   - Describe self-loop uses undici fetch (bypasses the runtime's hooked global fetch / dead local proxy), sends `Accept: application/json` + `stream: false`, disables compression for the image payload.
   - Remote image downloads send a User-Agent (Wikimedia etc. reject headerless bots with 400).

## Verification

[...]

PR #10415, opened 2026-08-14
fix(guardrails): reroute zero-vision combos through the vision bridge

rqzbeh:
## Summary

Named combos whose targets have **zero vision-capable models** are never
reroute-eligible in the Vision Bridge. The bridge only attempts the describe
path for them; when describing cannot run or fails (unreachable bridge model,
failed self-loop, missing credentials), the raw images stay in the payload and
the request dies in the combo capability filter with
`capability_mismatch: "No target in combo <name> has confirmed vision support for
this image request"` — the "Vision Bridge does not affect combo models" symptom.

This PR makes a combo with no vision-capable target behave like a single
text-only model: the whole request is rerouted to the configured vision bridge
model (same credential guards as the existing single-model reroute), falling
back to the describe path only when no usable reroute target exists.

`getComboVisionBridgeDecision` now returns a new `"no-vision"` decision for a
combo whose model targets all lack proven vision support, and the preCall
reroute gate treats it as reroute-eligible:

| Request model | Image handling (bridge enabled) | Before | After |
| --- | --- | --- | --- |
| Single text-only model | reroute to vision model | ✅ | ✅ |
| `auto` / `auto/*` | reroute to vision model | ✅ | ✅ |
| Combo, zero vision targets | reroute, else describe | ❌ capability_mismatch | ✅ |
| Combo, mixed targets | describe only | ✅ | ✅ |
| Combo, all vision targets | skip bridge | ✅ | ✅ |

Separately verified (not changed here): on REQUIRE_API_KEY instances the
describe self-loop previously failed with 401 because it used the `sk_omniroute`
sentinel; the DB-backed self-loop key fix shipped in #9946 (in this release
branch, not v3.8.49) resolves that path.

## Related Issues

- Related to #10296 (v3.8.50 not shipped as image yet — deployment still runs v3.8.49)

## Validation

- [x] Change type: routing
- [ ] `npm run lint`

## Tests Added Or Updated

- `tests/unit/guardrails/visionBridge-combo-reroute.test.ts` (new):
  - decision: zero-vision combo → `"no-vision"`; all-vision → `"skip"`;
    mixed → `"process"`; unknown → `"not-combo"`; model-combo mapping routes
    to the combo decision
  - preCall: zero-vision combo reroutes the whole request to the bridge model
    (images kept, no describe call)
  - preCall: zero-vision combo falls back to describe when the reroute target
    has no usable credentials
  - preCall: zero-vision combo without images is untouched
  - preCall: all-vision combo still skips; mixed combo still describes

## Coverage Notes

- Change is confined to `src/lib/guardrails/visionBridge.ts`; the new test file
  covers the new `"no-vision"` decision and both preCall outcomes (reroute and
  describe fallback) with the repo's dependency-injection pattern.
- Existing `tests/unit/guardrails/visionBridge*.test.ts` suites are unaffected
  (they exercise the injected `checkModelHasComboMapping` seam, which this PR
  does not alter).

## Reviewer Notes

- `getComboVisionBridgeDecision` is now exported for direct unit testing
  (previously module-private; the preCall path is unchanged in shape).
- Combos with unknown-capability targets are treated as non-vision (fail-closed),
  consistent with the combo capability filter's `supportsVision !== true` rule
  (open-sse/services/combo/comboStructure.ts).
- `visionBridgeRerouteTextOnly` behavior is unchanged: when enabled, every
  text-only route (including combos) reroutes, exactly as before.


diegosouzapw:
Thanks for the clean fix, @rqzbeh — this closes a real gap. I traced the `capability_mismatch` message you describe to `open-sse/services/combo/comboStructure.ts:655` and confirmed it matches exactly. The new `"no-vision"` decision and the reroute-then-describe-fallback ordering follow the existing credential-guard pattern used for the non-combo and `auto/*` reroute paths, so it's a natural extension rather than a new code path.

I ran your new test file directly against this branch (`node --import tsx/esm --test tests/unit/guardrails/visionBridge-combo-reroute.test.ts`) — all 11 tests pass. CI is fully green as well.

One thing worth a look in a follow-up (not blocking this PR): in the compound-failure case — reroute target has no usable credentials *and* the subsequent describe call also fails for every image — the `allNull` stub-text fallback at the end of `preCall` only fires for `comboVisionBridgeDecision === "process"`, not `"no-vision"`. That means in that narrower double-failure scenario the raw images would still be preserved and the original `capability_mismatch` could recur. Might be worth including `"no-vision"` in that guard too, with a small test for the double-failure path. Happy to see this merge as-is and pick that up separately if you'd rather keep this PR focused.

rqzbeh:
Good catch — implemented (d79e151c).

The `allNull` stub-text fallback at the end of `preCall` now fires for `"no-vision"` as well as `"process"`:

```ts
if (
  allNull &&
  (comboVisionBridgeDecision === "process" || comboVisionBridgeDecision === "no-vision")
) {
```

Rationale: a `"no-vision"` combo has zero vision-capable targets, so the combo capability filter rejects raw images outright (`capability_mismatch`). The #4012 preserve-raw behavior exists for paths where the upstream might still be vision-capable (reroute path / unknown capability) — that cannot be true here, so stub text is strictly better than bytes no combo target can consume. `"not-combo"` individual models are deliberately left out; their upstream may still be vision-capable.

Added one test for the compound-failure path: reroute target has no usable credentials (`hasUsableCredentials → false`) AND the describe call throws for every image. Asserts the combo model is kept, no `image_url` survives in the payload, and the messages contain the `(unavailable — no vision-capable provider connected)` stub.

Verification:
- PR test file: 12/12 pass (11 existing + the new double-failure case)
- Full guardrails suite: 145/145 pass
- eslint on the touched files: clean

Changelog fragment updated to mention the degraded stub-text behavior. Thanks for the close read — this closes the last path where a zero-vision combo could still surface `capability_mismatch`.

diegosouzapw:

PR #10416, opened 2026-08-14
fix(proxy-subscriptions): allow local/loopback proxy-subscription fetch URLs

### LINKED ISSUE #10158: [BUG] can not add local schema as a Proxy Subscription
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

_No response_

### Node.js Version

_No response_

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

I have a local HTTP proxy server (127.0.0.1) that provides a proxy list.

I want to subscribe to this list, but I found that it isn't supported. I don't know why.

### Steps to Reproduce

https://github.com/user-attachments/assets/38738568-50b2-41ad-8a30-5c1b17b00dcf

### Expected Behavior

It should be able to add local proxy subscription

### Actual Behavior

it was reject to add proxy subscription with an error

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

diegosouzapw:
Closes #10158

## Root cause

An operator running a local HTTP proxy-list server on `127.0.0.1` could not add
it as a proxy subscription. The subscription **fetch URL** was gated by a
strict SSRF guard (`src/lib/proxySubscription/fetchGuard.ts`) that
unconditionally blocked all loopback/private IP ranges — but the same feature
already permits loopback for the **routing** half
(`src/lib/proxySubscription/coreEndpoint.ts`'s `ALLOWED_LOCAL_CORE_HOSTS`, and
the dashboard even suggests `socks5://127.0.0.1:2080` as a core endpoint).
This was a trust-model inconsistency, not intentional hardening: an operator
could already route ALL traffic through a loopback core, but could not fetch
a proxy list from a loopback HTTP server.

## Fix

Make the fetch guard local-first by reusing the existing
`areLocalProviderUrlsAllowed()` policy (default ON — "OmniRoute is
local-first") from `src/shared/network/outboundUrlGuardPolicy.ts`:

- Loopback/private hosts (127.0.0.0/8, 10.0.0.0/8, 172.16.0.0/12,
  192.168.0.0/16, `::1`, ULA) are now allowed as subscription fetch targets
  by default.
- Cloud-metadata / link-local (`169.254.0.0/16`, incl. `169.254.169.254`
  IMDS), the unspecified address (`0.0.0.0/8`), and `::` stay blocked
  **unconditionally** regardless of the local-first flag — mirroring the
  provider-validation guard's "block-metadata" mode.
- Callers that want the old strict (public-only) behavior can pass
  `{ allowLocal: false }`.
- `assertSafeFetchTarget()` in `subscriptionService.ts` now passes this
  policy through to both the structural IP-literal check and the DNS
  multi-record re-check, so a hostname resolving to a private address is
  also allowed under the local-first default while IMDS/link-local stays
  refused.

All `/api/v1/management/proxy-subscriptions/*` routes already require
`requireManagementAuth` — the fetch is initiated server-side on behalf of an
authenticated operator, matching the trust boundary the provider-validation
path already uses (operator-supplied, not attacker-supplied).

## Regression tests

- `tests/unit/proxySubscription.fetchGuard.test.ts` — updated the
  "blocked IPv4/IPv6 literals" tests to the new local-first contract (loopback
  now allowed by default; added explicit strict-mode `{ allowLocal: false }`
  tests preserving the old behavior), plus two new `#10158` regression tests
  proving `http://127.0.0.1:8080/list` is allowed and
  `http://169.254.169.254/...` (IMDS) stays blocked.
- `tests/unit/proxySubscription.service.test.ts` — new end-to-end regression
  test that spins up a real local HTTP server on `127.0.0.1` and calls
  `createSubscription()` against it, asserting `status: "ok"` with parsed
  nodes (previously failed with `Fetch failed: Subscription URL is not
  allowed (scheme or host blocked)`).
- `docs/openapi.yaml` updated to document that the subscription `url` may be
  a local/loopback address.

TDD evidence (from the validated bug plan-file, promoted into these permanent
tests): RED on `release/v3.8.50` — `isSubscriptionFetchUrlAllowed("http://127.0.0.1:8080/list")`
returned `false`, and an end-to-end `createSubscription()` against a real
local server failed with `Fetch failed: Subscription URL is not allowed
(scheme or host blocked)`. GREEN after the fix — both assertions pass; full
suite run locally: 20/20 passing (`node --import tsx/esm --test
tests/unit/proxySubscription.fetchGuard.test.ts
tests/unit/proxySubscription.service.test.ts`).

## Gates run locally

- `npx eslint --suppressions-location config/quality/eslint-suppressions.json <changed files>` — exit 0 (no errors/warnings)
- `node scripts/check/check-file-size.mjs` — OK, no new frozen-file growth
- `node scripts/check/check-test-discovery.mjs` — OK, test files discovered
- Focused regression suite (`proxySubscription.fetchGuard.test.ts` +
  `proxySubscription.service.test.ts`) — 20/20 passing
- `node scripts/check/check-complexity.mjs` — OK (2456 violations, baseline 2774 — no
  regression)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1104 violations,
  baseline 1223 — no regression)
- `npm run check:docs-all` — pass (doc-links + fabricated-docs clean; the only
  output is 64 pre-existing `[stale-version]` drift warnings unrelated to this
  change)
- CI (`Unit Tests fast-path` 1-4/4, `Vitest (fast-path)`, `Fast Quality Gates`,
  `No new ESLint warnings`, `Docs Gates (fast-path)`, `Fast Production Build`,
  CodeQL, semgrep) — all green. Shard 3/4 initially hung for ~4h54m on the
  first CI attempt (all other shards completed normally in ~7-9min); a
  `gh run rerun --failed` reran it clean in 9m24s, matching the other shards'
  timing — confirming a one-off GitHub Actions runner flake, not a hang
  introduced by this change (network-fetch code paths in this diff are
  exercised only by the mocked/local-server proxySubscription tests, which
  pass deterministically both locally and in every other shard/rerun).

⚠️ base-red inherited: #9985 — ESLint errors (2) from #10250 (unrelated i18n
PT-PT translation commit on the release tip; not touched by this PR).

### DIFF
--- a/changelog.d/fixes/10158-local-proxy-subscription.md
+++ b/changelog.d/fixes/10158-local-proxy-subscription.md
@@ -1,0 +1,1 @@
+- fix(proxy-subscriptions): allow local/loopback proxy-subscription fetch URLs (local-first, cloud-metadata still blocked) (#10158)
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -8435,7 +8435,12 @@ components:
           type: string
         url:
           type: string
-          description: Redacted subscription URL.
+          description: >-
+            Redacted subscription URL. May be a local/loopback address
+            (e.g. `http://127.0.0.1:8080/list`) — local-first fetch targets
+            are allowed by default (`OMNIROUTE_ALLOW_LOCAL_PROVIDER_URLS`);
+            cloud-metadata / link-local endpoints (169.254.0.0/16) are always
+            blocked.
         enabled:
           type: boolean
         mode:
--- a/src/lib/proxySubscription/fetchGuard.ts
+++ b/src/lib/proxySubscription/fetchGuard.ts
@@ -3,29 +3,56 @@
  *
  * The subscription URL is fetched server-side (see `subscriptionService
  * .fetchSubscriptionContent`). Without a guard, an operator — or a compromised
- * subscription link — could point OmniRoute at internal services or cloud
- * metadata (SSRF). Only http/https to non-internal hosts are allowed:
- * loopback / private / link-local (incl. 169.254.0.0/16 cloud metadata) /
- * unspecified addresses are blocked.
+ * subscription link — could point OmniRoute at cloud metadata (SSRF). Only
+ * http/https to non-metadata hosts are allowed.
+ *
+ * Local-first (#10158): OmniRoute already lets an operator route ALL traffic
+ * through a loopback core (`coreEndpoint.ts` allows `127.0.0.1`/`::1`/
+ * `localhost`), so a subscription fetch target on loopback/private ranges is
+ * ALLOWED by default (`allowLocal: true`, matching the local-first default of
+ * `areLocalProviderUrlsAllowed()` in `src/shared/network/outboundUrlGuardPolicy
+ * .ts`) — mirroring that policy's "block-metadata" mode. Cloud-metadata /
+ * link-local (`169.254.0.0/16`, incl. `169.254.169.254` IMDS) and the
+ * unspecified address (`0.0.0.0/8`) are blocked UNCONDITIONALLY regardless of
+ * `allowLocal`, since they have no legitimate subscription-source use case.
+ * Callers that want the old strict (public-only) behavior pass
+ * `{ allowLocal: false }`.
  *
  * Hostname resolution is re-checked at fetch time (also using the IP-range
  * helpers here) so a hostname that resolves to an internal address is still
  * refused. Splitting the logic into pure functions keeps it unit-testable
- * without DNS / the full stack.
+ * without DNS / the full stack. No `@/`-aliased or DB-backed imports here —
+ * the `allowLocal` policy decision is made by the caller (subscriptionService,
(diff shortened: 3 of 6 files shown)


diegosouzapw:
The local-first behavior addresses the reported issue, but I found a security gap in the new IPv6 guard. `isIpv6Blocked()` only checks `::`, `fe80` (a /16 prefix), and strict-mode `::1`/`fc`/`fd`; it does not classify IPv4-mapped IPv6 or the full fe80::/10 link-local range. As a result, `http://[::ffff:169.254.169.254]/...` and `http://[::ffff:127.0.0.1]/...` pass the structural guard, as do `fe90::1` and `febf::1`. Please normalize/check mapped IPv4 and cover the full link-local range before merging. This review is based on focused probes against the available review worktree source, so the finding is PLAUSIBLE rather than CONFIRMED.

PR #3821, opened 2026-06-14
Expose Arena ELO sync in feature flags

rdself:
## Summary
- Add `ARENA_ELO_SYNC_ENABLED` to the Dashboard Feature Flags registry so it can be managed with DB overrides.
- Route Arena ELO startup/status checks through the shared feature flag resolver while preserving the existing env fallback.
- Refresh env docs for Arena ELO sync and add the missing `STREAM_READINESS_TIMEOUT_MS` example so env/doc sync stays green.

## Validation
- `node --import tsx/esm --test tests/unit/feature-flags-settings.test.ts`
- `node --import tsx/esm --test tests/unit/arena-elo-sync.test.ts`
- `node --import tsx/esm --test --test-name-pattern "Arena ELO" tests/integration/integration-wiring.test.ts`
- `npm run check:docs-all`
- `npx eslint src/instrumentation-node.ts src/lib/arenaEloSync.ts src/server-init.ts src/shared/constants/featureFlagDefinitions.ts src/shared/utils/featureFlags.ts tests/integration/integration-wiring.test.ts tests/unit/arena-elo-sync.test.ts tests/unit/feature-flags-settings.test.ts`

## Notes
`npm run check:docs-all` exits successfully. It still prints the repository's existing stale/fabricated-doc warning report in non-strict mode.

PR #10422, opened 2026-08-14
fix(antigravity): discover live chat models dynamically

JxnLexn:
## Summary

- keep the official Antigravity `:fetchAvailableModels` endpoint as the source of truth for the current account and client version
- accept newly announced chat-capable model IDs from live discovery without requiring a static catalog entry
- retain static catalog metadata for known models and preserve the Gemini 3.7 Flash fix
- continue filtering internal, tab-completion, media, embedding, TTS, video, and confirmed retired model IDs

## Root cause

Antigravity discovery was filtering the authenticated `fetchAvailableModels` response through `isUserCallableAntigravityModelId()`, which only accepted IDs present in the static `ANTIGRAVITY_PUBLIC_MODELS` catalog. When Antigravity exposed Gemini 3.7 Flash, the upstream returned the models but OmniRoute removed them before returning `/models` results.

The fix now treats Antigravity's authenticated live response as the source of truth for account-available chat models. Static metadata remains a fallback for known models, while a narrow denylist keeps known non-chat and retired surfaces out of the chat catalog.

This means future chat models announced by Antigravity can appear automatically after a discovery refresh, without another OmniRoute catalog PR, as long as Antigravity reports them for the connected account and they are not an explicitly excluded non-chat or retired ID.

## Testing

- `node --import tsx/esm --test tests/unit/antigravity-model-aliases.test.ts tests/unit/provider-models-route.test.ts tests/unit/antigravity-retired-public-models.test.ts tests/unit/t28-model-catalog-updates.test.ts tests/unit/model-capabilities-registry.test.ts`
  - 93 passed, 0 failed
- `npm run typecheck:core`
- `npm run lint`
- `npx prettier --check` on all changed files
- `git diff --check`

The regression suite explicitly verifies that a hypothetical future `gemini-3.8-flash-high` returned by the live endpoint is surfaced, while internal/media/retired IDs remain filtered.


diegosouzapw:
Thanks for the thorough root-cause writeup and the test coverage on this one -- the diagnosis (static allowlist blocking newly-announced live chat tiers) and the fix direction (mirroring the existing agy discoverable/callable split) both look right, and the catalog/discovery tests you added are solid.

One regression slipped in with the rename sweep, though: `src/lib/usage/providerLimits/quotaNormalize.ts` was switched from `isUserCallableAntigravityModelId` to `isDiscoverableAntigravityModelId` along with the two discovery call sites, but that one feeds `getSanitizedCachedProviderLimitsMap()` (the ProviderLimits dashboard cache sanitizer), whose contract -- per the existing `tests/unit/provider-limits-sanitize-scope-3821.test.ts` (LEDGER-2 / #3821-review) -- is to drop any quota key that isn't a recognized, allowlisted model id. The new denylist-based check only excludes known non-chat/retired ids, so an arbitrary/junk quota key now survives sanitization there. I reproduced this locally: the test passes cleanly on release/v3.8.50 before your changes and fails with the exact same assertion after checking out this branch, matching what both CI jobs ("Fast Quality Gates" and "Unit Tests fast-path (4/4)") are currently reporting.

Could you revert just that one call site back to `isUserCallableAntigravityModelId` (it's already imported/exported, so this should be a one-line change), keep `isDiscoverableAntigravityModelId` for the two model-discovery paths, and re-run the sanitize-scope test plus your existing regression suite? Everything else here looks ready to go once that's green.

JxnLexn:
Thanks for catching this — confirmed and fixed in `fe29dab8e`.

I reverted only the Provider Limits cache sanitizer call site in `src/lib/usage/providerLimits/quotaNormalize.ts` back to `isUserCallableAntigravityModelId()`. The two authenticated live model-discovery paths continue to use `isDiscoverableAntigravityModelId()`, so new upstream chat models remain dynamically discoverable while cached Provider Limits still drop arbitrary/non-allowlisted quota keys.

Validation rerun:

- `node --import tsx/esm --test tests/unit/provider-limits-sanitize-scope-3821.test.ts tests/unit/antigravity-model-aliases.test.ts tests/unit/provider-models-route.test.ts tests/unit/antigravity-retired-public-models.test.ts tests/unit/t28-model-catalog-updates.test.ts tests/unit/model-capabilities-registry.test.ts` — 96 passed, 0 failed
- `npm run typecheck:core`
- `npm run lint`
- Prettier check and `git diff --check`

The failing sanitize-scope assertion was reproduced before the change and passes after it.

diegosouzapw:

Issue #8491, opened 2026-07-24
fix(providers): Antigravity runtime-discovered projectId is never persisted → recurring 422 after refresh/restart


diegosouzapw:
## Summary

An Antigravity connection whose stored `projectId` is empty relies on a **request-time**
discovery that is memoized **in process memory only**, keyed by the OAuth **access token**,
and is **never written back to the connection**. As a result the account works until the
access token is refreshed (or OmniRoute restarts) and then starts failing with:

```
[antigravity/<model>] [422]: Missing Google projectId for Antigravity account.
Auto-discovery via loadCodeAssist found no Cloud Code project.
```

Reported from the community: the model answered correctly in the Playground, then began
422-ing "after a while" without any config change.

## Root cause

`open-sse/services/antigravityProjectBootstrap.ts`:

```ts
/** Per-token memoization cache (lives for the process lifetime). */
const projectCache = new Map<string, string>();               // :33

function getProjectCacheKey(accessToken: string, clientProfile: AntigravityClientProfile) {
  return `${clientProfile}:${accessToken}`;                    // :37  ← token in the key
}
```

`ensureAntigravityProjectAssigned()` stores the discovered id in that Map and returns it.
`open-sse/executors/antigravity.ts:532-546` uses the returned value for the current request
only — nothing persists it to the connection's `projectId` /
`providerSpecificData.projectId`.

Consequences:

1. **Every access-token refresh invalidates the cache** (the token is part of the key), so a
   fresh `loadCodeAssist` round-trip is required after each refresh.
2. **Every process restart loses it** entirely.
3. If that later `loadCodeAssist` returns empty or transiently fails, the request 422s even
   though the very same account resolved a project id minutes earlier. The successful
   discovery is thrown away.

The OAuth path does persist it (`src/lib/oauth/providers/antigravity.ts:146-182` →
`mapAntigravityTokens`), so only the runtime-recovery path has the gap.

## Expected

A `projectId` discovered at request time should be persisted to the connection (same field the
OAuth path writes), so it survives token refreshes and restarts and the 422 does not recur for
an account that already resolved successfully.

## Proposed fix

1. In the executor's recovery branch (`antigravity.ts:539-546`), after
   `ensureAntigravityProjectAssigned()` returns an id, write it back to the connection
   credential (`projectId` + `providerSpecificData.projectId`) — mirroring what the OAuth
   post-exchange stores.
2. Consider keying the in-memory cache by connection id rather than by access token, so a
[...]

Issue #2934, opened 2026-05-30
[Feature] No apiKey need for memory  when authed against MCP server


androw:
### Problem / Use Case

Currently, I understand that I need to use the MCP server if I want to use the Omniroute-saved memory.

As I'm not using localhost to access the MCP, I need to auth to the MCP server through Authorization header.

But then, if I want to manager memories, I need to pass my apiKey again to the LLM (**which I prefer not**) for the tool call.

### Proposed Solution

Render the apiKey parameter optional when authenticated against the MCP server

### Acceptance Criteria

- No apiKey requested when already authed against MCP server
- No change otherwise

### Area

Other

diegosouzapw:
Hi @androw! Thanks for the suggestion! 🙏

Good catch — the memory MCP tools currently require an explicit `apiKeyId` argument even though, when you're already authenticated against the MCP server, we have your caller identity from the auth context. We can derive it instead of asking for it.

> Plan: make `apiKeyId` optional on the memory tools and fall back to the MCP-authenticated caller identity, so memory works without re-supplying a key.

**Status:** ✅ Accepted — cataloged for an upcoming release.

We'll tag you when it ships. Closing to keep the list focused; tracked in our backlog. Thank you! 🚀

PR #10424, opened 2026-08-14
fix(antigravity): heal empty-projectId accounts via retryable auto-onboarding

rqzbeh:
## Summary

Accounts with an **empty Google Cloud Code `projectId`** fail every request
with `422 Missing Google projectId`. The 3.8.50 bootstrap already tries to
**auto-create** the missing project via Google's `onboardUser` endpoint (then
re-discovers it via `loadCodeAssist`) — but a single failed attempt was
memoized **forever** in `onboardAttemptedCache` (a plain `Set`), so every later
request in the same process skipped onboarding and 422'd, even when a retry
would have succeeded.

This PR makes auto-onboarding **retryable and self-healing**:

- The permanent per-token memo is replaced with a failure-backoff map:
  a failed onboard attempt is retried once the 5-minute backoff expires, so a
  transient failure (network blip, upstream 5xx) no longer poisons the account
  for the process lifetime.
- The in-flight per-token lock still dedupes concurrent calls (one onboard
  attempt at a time per token).
- A successful discovery clears the failure marker and is memoized + persisted
  exactly as before (`projectCache` LRU, `persistDiscoveredAntigravityProjectId`).

Net effect: accounts that CAN be onboarded (have a refresh token) heal
automatically on a later request or token refresh — the Google project is
created **without any user action**. Accounts without a refresh token still
require a one-time OAuth re-auth (Google-side limitation, cannot be fixed in
code).

## Related Issues

- Fixes the "empty google project id → error" report (production v3.8.49 log:
  `422 Missing Google projectId ... loadCodeAssist found no Cloud Code project`)

## Validation

- [x] Change type: routing
- [x] `npm run lint` — no issues in changed files

Focused tests run:
```
node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --test \
  tests/unit/antigravity-discovery-bootstrap.test.ts \
  tests/unit/antigravity-missing-project-chat.test.ts \
  tests/unit/token-refresh-service.test.ts          # 55/55 pass
```

## Tests Added Or Updated

- `tests/unit/antigravity-discovery-bootstrap.test.ts`:
  - updated: "does not re-attempt onboardUser within the failure backoff
    window" (was: "does not retry onboardUser for the same token" — behavior
    within the window is unchanged, name now matches the backoff semantics)
  - added: "retries onboardUser after the failure backoff expires (account
    heals itself)" — first attempt fails, backoff blocks an immediate retry,
    after the backoff expires the account re-onboards and recovers the project

## Coverage Notes

- Change is confined to `open-sse/services/antigravityProjectBootstrap.ts`
  (one cache data structure + its gate/finally bookkeeping) plus the test
  suite; no routing, executor, or wire-format behavior changes.
- `clearAntigravityProjectCache` (test export) now clears the failure map too;
  a new test-only `clearAntigravityOnboardBackoff(key?)` simulates backoff
  expiry.

## Reviewer Notes

- `ONBOARD_RETRY_BACKOFF_MS = 5 min` bounds Google-side retry load: at most one
  onboard attempt per token per 5 minutes, plus the existing LRU cap
  (`MAX_CACHE_SIZE = 256`).
- Successes are memoized as before (LRU, no TTL — pre-existing behavior,
  unchanged); this PR only changes the FAILURE path.
- The geo-block issue for this deployment is handled by the separate PR
  (#10420): accounts whose egress location is refused still 422/400 until
  egress is routed through a supported region — auto-onboarding cannot create a
  working project from a blocked region.


diegosouzapw:
Thanks for this — the failure-backoff redesign in `antigravityProjectBootstrap.ts` is a real fix for a real problem (a single transient `onboardUser` failure permanently poisoning an account for the process lifetime), it's a clean, well-scoped change, and the new test coverage for the backoff/heal cycle is thorough. I ran the focused suite you listed plus `token-refresh-service.test.ts` and got 57/57 green, and all fast-path CI checks are passing.

Two things I'd like fixed before merge:

1. **`EditConnectionModal.tsx`: the new `isProjectIdManual` flag doesn't actually get saved.** It's set on `updates.providerSpecificData` right after the project-id field (line ~515), but for OAuth connections — which Antigravity always is — the code a bit further down (lines ~620-643) unconditionally rebuilds `updates.providerSpecificData` from `connection.providerSpecificData` before the request goes out, discarding the flag you just set. As shipped, the guard you added in `tokenRefresh.ts` (`!credentials.providerSpecificData?.isProjectIdManual`) is checking a field that's never actually persisted. Could you merge the flag into that later object (or set it after it) and add a regression test? There's a good template already in the repo for this exact surface: `tests/unit/dashboard/edit-connection-modal-openai-store-toggle.test.tsx`.

2. **The `#2934` citation for the "Google deprecated automatic project creation" claim points at an unrelated issue** (it's a closed MCP-memory-auth feature request, not anything about GCP/Antigravity). Since the whole BYOP-detection heuristic — and its permanent, non-expiring cache — rests on that claim about Google's behavior, could you swap in the real source (a production log excerpt of the actual `onboardUser` response, or the correct tracking issue)?

One more thing worth a look, not blocking: returning `403` for the new BYOP fast-fail path routes it through `chatCore`'s generic "401/403 → refresh credentials and retry" handling, so every request from an affected account triggers a real (and pointless, since refreshing the token can't create a GCP project) round trip to Google's OAuth token endpoint before the 403 reaches the client. I confirmed this by instrumenting your own BYOP test — the token endpoint is hit once per request, indefinitely (no backoff, unlike the onboard-failure path). Might be worth having `chatCore` skip the refresh-retry for this specific error code, or using a status it doesn't intercept.

Happy to take another look once these land.

rqzbeh:
Thanks for the detailed review — all three points are addressed in commit `5a4bbcc2e` (pushed, CI re-running):

**1. `isProjectIdManual` now actually persists.** The dead assignment (set on `updates.providerSpecificData` then unconditionally rebuilt by the OAuth path) is removed; the flag now lands in the single surviving antigravity merge block that runs after both the OAuth and API-key branches, so `tokenRefresh.ts` is guarding a field that really reaches the PUT. Added `tests/unit/dashboard/edit-connection-modal-antigravity-project-manual.test.tsx` (modeled on the OpenAI-store-toggle template you pointed at): it renders an OAuth antigravity connection, types a Project ID, saves, and asserts `onSave` receives `providerSpecificData.isProjectIdManual === true` — plus the empty-field case asserting `false`.

**2. Citation swapped for the real source.** All `#2934` references (bootstrap doc comment + detection comment, executor comment, and the three test files' names/comments) now point at **#8491** — the actual tracking issue for the empty-Google-projectId → 422 class, whose body carries the production log excerpt. The BYOP comment is also reworded to state the observable fact (onboardUser 200 without `cloudaicompanionProject`) rather than the unverifiable deprecation claim.

**3. BYOP fast-fail no longer triggers the pointless refresh.** The executor now returns **422** for `GCP_PROJECT_REQUIRED` instead of 403 — 422 is outside chatCore's 401/403 refresh-retry set, so an affected account no longer hits Google's OAuth token endpoint on every request, and it matches the sibling `missing_project_id` error the client already maps to an action-needed prompt. Both BYOP tests updated (executor `fast-422s`, and the chat-level test now asserts the 422 passthrough with the actionable message intact).

One note for transparency: the commit also carries an `eslint-disable-next-line react-hooks/set-state-in-effect` on the modal's pre-existing form-hydration effect — that error exists on `release/v3.8.50` unchanged (verified), the dashboard isn't covered by the CI lint gate, and the repo already uses this exact convention in 11 other dashboard files, so I followed it rather than refactoring unrelated form logic in this PR.

Verified locally before pushing: 88/88 node:test (executor, missing-project-chat, discovery-bootstrap, token-refresh-service, streaming-sanitize), 5/5 vitest including the new regression test, `typecheck:core` clean, ts7 ratchet 0 added, eslint clean on all touched files.

rqzbeh:
Follow-up on the "one more thing worth a look" note from this review (the 403 → 401/403 refresh-retry interaction): the account-rotation piece is now a separate PR — **#10470** (feat(antigravity): auto-rotate BYOP accounts to siblings on GCP_PROJECT_REQUIRED).

Quick summary of the design decision: a `gcp_project_required` 422 is account-specific (that Google account lacks a GCP Project ID), so instead of just returning the error, routing now marks the connection excluded (24h, same `rateLimitedUntil` mechanism as the GEO_BLOCKED exclusion) and rotates to a sibling antigravity account **within the same request** — the actionable 422 surfaces only when no sibling exists. It's classified as its own error type (explicitly not `MODEL_NOT_FOUND`, so no model lockout), and the rotation state is tracked separately from `maxAttempts` so non-BYOP antigravity failures never get a second upstream shot.

It's stacked on this branch (the 422 fast-fail only exists here), so it reduces to its own delta once #10424 merges. This PR keeps the two review fixes as-is.

diegosouzapw:
Thank you @rqzbeh for the Antigravity project auto-onboarding fix. The exact head passed the focused discovery/bootstrap tests and the combined FAST train; no PR-specific red was observed.

Issue #4421, opened 2026-06-20
[BUG] openai-compatible-responses provider creation fails with "OpenAI Compatible node not found" even when node exists


aleksesipenko:
## Description

When trying to create a custom provider with `provider: "openai-compatible-responses"`, the API returns `404 {"error":"OpenAI Compatible node not found"}` even though a matching `provider_nodes` entry exists in the database and is visible via `GET /api/provider-nodes`.

## OmniRoute Version

3.8.30 (npm global install, Node v22.23.0, Linux VPS)

## Environment

- OS: Ubuntu (VPS, Finland)
- Node.js: v22.23.0
- Installation: npm global (`npm i -g omniroute`)
- Storage: SQLite (`DATA_DIR=/home/alex/.omniroute`)
- Dashboard/API behind nginx reverse proxy (Cloudflare Tunnel)

## Steps to Reproduce

1. **Create an `openai-compatible` node with `apiType: responses`:**

```bash
curl -s -X POST http://127.0.0.1:20128/api/provider-nodes \
  -H "Authorization: Bearer *** \
  -H "Content-Type: application/json" \
  -d '{
    "type": "openai-compatible",
    "name": "fox-codex",
    "prefix": "fox",
    "apiType": "responses",
    "baseUrl": "https://dm-fox.rjj.cc/codex/v1",
    "chatPath": "/responses",
    "modelsPath": "/models"
  }'
```

Response: `201 Created` — node is stored successfully:
```json
{
  "node": {
    "id": "openai-compatible-responses-1715ed0f-...",
    "type": "openai-compatible",
    "name": "fox-codex",
    "prefix": "fox",
    "apiType": "responses",
    "baseUrl": "https://dm-fox.rjj.cc/codex/v1",
    "chatPath": "/responses",
    "modelsPath": "/models"
  }
}
```

2. **Verify the node is visible via GET:**

```bash
curl -s http://127.0.0.1:20128/api/provider-nodes \
  -H "Authorization: Bearer ***rn returns the node correctly:
```json
{
  "nodes": [{ "id": "openai-compatible-responses-1715ed0f-...", "type": "openai-compatible", "name": "fox-codex", "apiType": "responses", "baseUrl": "https://dm-fox.rjj.cc/codex/v1", ... }],
  "ccCompatibleProviderEnabled": false
}
```

3. **Attempt to create a provider connection using `openai-compatible-responses`:**

```bash
[...]

PR #10434, opened 2026-08-15
fix(sse): bridge generic compatible-provider type id to concrete node id in credential lookup

### LINKED ISSUE #10085: [BUG] Custom OpenAI-compatible provider: chat fails with No active credentials for provider though model list & key are valid (chat credential reference mismatch)
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

macOS

### OS Version

26.5.2

### Node.js Version

_No response_

### Provider(s) Involved

OpenAI compatible

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

When a custom OpenAI-compatible provider is added (self-hosted instance), the model catalog and /v1/models listing work and the provider "test" succeeds, which is proving the upstream API key is valid and reaches the backend. 

However, chat completions fail with: `No active credentials for provider: openai-compatible-chat-458d982b-3727-4b8d-8f21-77ee2c19984f`

The error names the exact catalog node id (`openai-compatible-chat-458d982b-…`), which strongly suggests the chat completion path looks up credentials against a provider reference that does not match the connection row the key was actually bound to. 

### Steps to Reproduce

1. Dashboard → Providers → Add compatible provider (creates the node / populates the catalog) with API key. ✅ test ok
2. setup connection and add the API key. ✅ test ok
3. model list retrieved automatically from provider's `/v1/models` ✅ works.
4. Test chat or Send a chat completion for any model of that provider ❌ `No active credentials for provider: openai-compatible-chat-458d982b-…`.

### Expected Behavior

Adding a custom OpenAI-compatible provider and its key should behave like any other providers: make both catalog listing and chat completions work, with the key bound to the same provider reference the chat path queries.

### Actual Behavior

Catalog works; chat fails. The credential connection row for the chat path's provider reference is either missing or incorrect, even though a key was supplied during provider setup.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

diegosouzapw:
Closes #10085

## Root cause

`getProviderSearchPool` (`src/sse/services/auth.ts`) built the credential
search pool for a custom OpenAI-compatible / Anthropic-compatible provider
node by bridging the searched provider string to a node id only via the
node's `prefix`. It never bridged the **generic derived type id**
(`openai-compatible-chat`, `openai-compatible-responses`,
`anthropic-compatible`) to the node's **concrete uuid node id**
(`openai-compatible-chat-<uuid>`), even though
`resolveProviderNodeForConnection` (`src/lib/db/providers/nodes.ts`, #4421)
already accepts the bare generic type id at connection-creation time.

Result: a connection persisted under the generic type id (e.g. via
`/api/providers` posting the bare type) became permanently unreachable from
the chat path, which resolves the concrete uuid node id — surfacing
`No active credentials for provider: openai-compatible-chat-<uuid>` even
though the API key and model catalog were valid. The provider "Test" button
and `/v1/models` catalog hit the upstream directly and never touch
`provider_connections`, which is why those "worked" while chat failed.

## Fix

Expand the custom-provider node-expansion loop in `getProviderSearchPool` to
also add a node's `nodeId` to the search pool when the searched provider
equals that node's generic derived type id (via `nodeTypeFromId`, the same
UUID-stripping helper `resolveProviderNodeForConnection` uses), and
symmetrically add the generic type id when the searched provider is the
concrete node id — mirroring the creation-time resolver so the two paths
agree.

## Regression test

`tests/unit/10085-compatible-generic-vs-uuid-credential.test.ts` (adapted
from the plan-file's RED probe):
- generic-type-stored connection found via uuid-node-id lookup (was the bug)
- uuid-stored connection found via generic-type-id lookup (symmetric direction)
- control: uuid-stored connection found via uuid-node-id lookup (unchanged)
- control: uuid-stored connection found via prefix lookup (unchanged)
- negative: bridge does not leak into an unrelated generic type (chat vs responses)

```
node --import tsx/esm --test tests/unit/10085-compatible-generic-vs-uuid-credential.test.ts
ℹ tests 5
ℹ pass 5
ℹ fail 0
```

## Gates run

- `npx eslint --suppressions-location config/quality/eslint-suppressions.json src/sse/services/auth.ts tests/unit/10085-compatible-generic-vs-uuid-credential.test.ts` — clean
- `node scripts/check/check-test-discovery.mjs` — OK, new test file discovered, no new orphans
- `node scripts/check/check-complexity.mjs` / `check-cognitive-complexity.mjs` could not complete locally within a reasonable window due to devbox saturation from a large parallel fan-out (many `check-complexity.mjs` processes contending simultaneously); the change is a small, flat, early-return addition inside an existing loop, and CI (`quality-gate` job) will validate it.
- Full `npm run test:unit` was not re-run locally for the same saturation reason; the targeted regression file is green and the change touches no other call sites of `getProviderSearchPool`.

⚠️ base-red inherited: #9985 — ESLint errors (2) from #10250 (unrelated i18n PT-PT translation, pre-existing on the release tip)


diegosouzapw:
Fix-in-place recommended. The #10085 regression is plausibly addressed, and the existing provider-node helper tests pass. However, lines 985-995 add the bare derived type to every concrete node lookup of that type. With nodes A and B sharing openai-compatible-chat, a generic connection created while A was the sole node is returned for B's UUID lookup; its stored baseUrl/prefix can send B traffic to A's endpoint. Please define the ambiguity rule or bind generic rows to their concrete node, add a two-node regression test, and rerun the focused PR test plus required gates. Base-red #9985 is inherited and is not attributed to this PR.

PR #8013, opened 2026-07-21
refactor(antigravity): align official clients and callable catalog

backryun:
## Summary

Extract the Antigravity/AGY work from #6958 into a focused, freshly reconstructed change set based directly on `release/v3.8.49`.

This PR aligns OmniRoute with the two official Antigravity client identities and limits the public catalog to models the current backend can actually call.

### Official client identities

- Treat Antigravity IDE and the official `agy` CLI as separate client profiles while sharing the same backend protocol.
- Use the observed Antigravity IDE `2.1.1` identity and the official `google-antigravity/antigravity-cli` `1.1.5` release as safe fallback versions.
- Keep persisted legacy `harness` / `sdk` profile values readable by normalizing them to `cli`; new writes accept only `ide` / `cli`.
- Align OAuth enrichment, request headers, project bootstrap, quota lookup, discovery, and executor behavior with the selected profile.
- Remove synthetic client headers and keep project identity resolution centralized.

### Callable catalog

- Rebuild the Antigravity and AGY catalogs from current upstream model identifiers.
- Add the three Gemini 3.6 Flash tiers now returned to both official client profiles:
  - `gemini-3.6-flash-high` — current upstream default
  - `gemini-3.6-flash-medium`
  - `gemini-3.6-flash-low`
- Keep the internal `gemini-3.6-flash-tiered` selector out of the user-facing catalog.
- Keep callable text/vision models in the chat catalog and image-only models in the image catalog.
- Preserve legacy public aliases through alias resolution without advertising retired IDs as currently callable.
- Align discovery normalization, model specs, CLI defaults, free-model metadata, pricing/subscription metadata, combo labels, and seed aliases with the rebuilt catalog.
- Add direct regression coverage that every current callable Gemini tier reaches the Antigravity envelope path.

The Gemini 3.6 entries were verified on 2026-07-22 through the live `v1internal:fetchAvailableModels` response using both the Antigravity IDE 2.1.1 and `agy` CLI client identities. The response reports a 1,048,576-token context window and 65,536-token maximum output for all three tiers, with High selected as `defaultAgentModelId`.

## Compatibility and migration safety

- No database migration number is added or reused.
- In particular, historical migration slot `121` remains a known gap and is not repurposed.
- Existing stored `harness` / `sdk` values remain readable through boundary normalization, avoiding a destructive data rewrite.

## Validation

Passed on the final branch:

- Antigravity-focused changed tests: 333/333
- Gemini 3.6/catalog regression selection: 160/160
[...]

PR #6114, opened 2026-07-03
fix(antigravity): strip trailing assistant prefill for Vertex Claude models — relocated to live executor, NEEDS LIVE VALIDATION

diegosouzapw:
## Summary

Vertex AI (Cloud Code Assist, used by Antigravity for Claude-branded models) rejects a conversation whose last turn is an assistant message — `"This model does not support assistant message prefill"`. This port strips a trailing assistant turn for Claude models routed through the Antigravity executor so the upstream request always ends on a user turn.

**Relocated from dead code to the live executor path.** The original upstream diff patched `openaiToClaudeRequestForAntigravity` in `open-sse/translator/request/openai-to-claude.ts`, which has **zero callers** in OmniRoute — porting it verbatim would have been inert. The live Antigravity Claude dispatch path is `AntigravityExecutor.transformRequest` (`open-sse/executors/antigravity.ts`), which converts Claude requests to Gemini `contents` (`role: "user"/"model"`) via `sanitizeAntigravityGeminiRequest` and bypasses `BaseExecutor.execute` entirely — so it never ran the 3 existing trailing-assistant guards elsewhere in the codebase (Mistral #3396, Copilot #5802, CC-bridge in `claudeCodeCompatible.ts`). This PR adds a new `stripTrailingAntigravityAssistantTurn` helper mirroring that same pop-loop pattern, wired into the `isClaude` branch only — native Gemini models via Antigravity are unaffected.

## ⚠️ Needs live validation before merge

The upstream diff patched a function with zero callers in OmniRoute; this relocates the fix to the live Antigravity executor path (`antigravity.ts`, Gemini `contents`, `isClaude`). The premise (Vertex/Cloud-Code-Assist rejects a trailing assistant turn) matches 3 in-repo precedents but was NOT empirically confirmed for this exact envelope. TDD covers the strip behavior. **Validate on the VPS with a real Antigravity Claude request carrying a trailing assistant turn before merging** — if Vertex actually accepts prefill here, this would remove a legitimate capability.

## Test plan

- [x] `tests/unit/antigravity-claude-prefill-strip.test.ts` (new, 6 cases) — trailing `role:"model"` strip for Claude models, non-Claude/Gemini untouched, conversation already ending on user unchanged, multiple trailing model turns stripped, never emptied to zero, empty/missing contents no-op. Watched fail → pass.
- [x] `npm run typecheck:core` — clean
- [x] Targeted `eslint` on changed files — clean
[...]

PR #10436, opened 2026-08-15
fix(antigravity): strip trailing model turn for native Gemini requests too

### LINKED ISSUE #10104: [bug] Antigravity Gemini requests can end on a model turn and fail with 400
## Bug

**Source:** Discussion #10069 by @SubhadipJana1409

## Problem

A Claude-format client routed through the AGY/Antigravity provider can send a Gemini request whose transformed `contents` ends with `role: "model"`. Newer Gemini endpoints reject that request with HTTP 400:

```text
Requests ending with a model turn are not supported.
```

The reporter selected the legacy model name `agy/gemini-3.5-flash-high`; the current callable Gemini 3.5 Flash High ID is `agy/gemini-3-flash-agent`. The legacy selection should not obscure the request-shape failure: the returned error specifically identifies an invalid trailing model turn.

## Current OmniRoute State

`open-sse/executors/antigravity.ts` already has `stripTrailingAntigravityAssistantTurn()`, but `AntigravityExecutor.transformRequest()` invokes it only when `isClaude` is true:

```ts
const transformedRequest = isClaude
  ? stripTrailingAntigravityAssistantTurn(
      sanitizeAntigravityGeminiRequest(rawTransformedRequest)
    )
  : rawTransformedRequest;
```

Therefore native Gemini models routed through Antigravity currently keep a trailing `model` turn unchanged. The existing regression suite in `tests/unit/antigravity-claude-prefill-strip.test.ts` explicitly verifies that non-Claude Gemini requests are untouched.

The Claude-only behavior was introduced by PR #6114 and live-validated against Vertex Claude. This report shows that the rejection surface now also needs to be evaluated for current Gemini models.

## Expected Behavior

OmniRoute should not send a request shape that the selected Antigravity Gemini model rejects solely because the conversation ends with a model turn.

The fix must preserve valid model/tool history and must not blindly delete a lone content entry or remove legitimate provider capabilities.

## Reproduction / Validation Plan

1. Add a failing regression test that drives the live `AntigravityExecutor.transformRequest()` path with a native Gemini model and input ending on an assistant/model turn.
2. Confirm the transformed request currently ends with `role: "model"`.
3. Test current callable IDs, including `gemini-3-flash-agent` and the Gemini 3.6 Flash tiers.
4. Implement the narrowest safe normalization for models that reject assistant prefilling.
5. Preserve the existing Claude behavior and the guard that never reduces `contents` to an invalid empty array.
6. Run the full Antigravity executor/translator test selection.
7. Per Hard Rule #18, validate the final behavior against a real Antigravity connection: reproduce the 400 before the change and obtain 200 after the change with the same request.

## Current Workarounds

- Replace the retired saved model selection with the current canonical ID: `agy/gemini-3-flash-agent`.
- Start a fresh client conversation after changing the saved model.
- If the canonical Gemini model still returns the same 400, temporarily use a working OpenCode model or an Antigravity Claude model while this is investigated.

## Related Work

- Discussion #10069
- PR #6114 - Claude-only trailing assistant-prefill guard
- PR #8013 - current AGY/Antigravity callable model catalog
- Closed issue #1316 - different Antigravity Gemini 400 causes (thinking/role/signature translation)
- Closed issue #3763 - historical model-ID investigation, not this trailing-turn rejection
- External reference: Zoo-Code-Org/Zoo-Code#1052 reports the same Gemini rejection rule on newer Gemini versions

diegosouzapw:
Closes #10104

## Root cause

`AntigravityExecutor.transformRequest()` in `open-sse/executors/antigravity.ts` only
wired `stripTrailingAntigravityAssistantTurn()` into the `isClaude` branch. A native
Gemini model routed through Antigravity (`agy/gemini-*`, e.g. `gemini-3-flash-agent`)
that ends the conversation on a `role: "model"` turn keeps that trailing entry
unchanged, and newer Gemini endpoints reject that shape with HTTP 400 "Requests
ending with a model turn are not supported."

`stripTrailingAntigravityAssistantTurn` (already correct, guarded pop-loop, never
empties `contents`) is now also applied on an `isGemini` branch
(`upstreamModel.toLowerCase().includes("gemini")`), independent of the
`sanitizeAntigravityGeminiRequest()` Claude-only field whitelist. The Claude path is
byte-identical to before — this only widens the strip to native Gemini requests.

## Regression test

`tests/unit/antigravity-claude-prefill-strip.test.ts`:
- Test (b) previously **asserted the buggy pass-through** for native Gemini
  (`antigravity/gemini-3.1-pro`) — flipped to assert the fixed behavior (trailing
  model turn stripped, contents ends on `user`).
- New test (b2) covers the current callable `gemini-3-flash-agent` tier from PR #8013's
  catalog.

Fail→pass evidence (same test file, run against pre-fix vs post-fix
`antigravity.ts`):

```
# pre-fix (HEAD~1 antigravity.ts, same test file)
✖ (b2) native Gemini agent tier (gemini-3-flash-agent, #10104) also gets the strip
  AssertionError [ERR_ASSERTION]: trailing model turn should be stripped for native Gemini
  2 !== 1

# post-fix
✔ (a) strips a single trailing assistant (model) turn for Claude models
✔ (b) strips a trailing model turn for native Gemini models too (#10104)
✔ (b2) native Gemini agent tier (gemini-3-flash-agent, #10104) also gets the strip
✔ (c) a Claude conversation already ending on user is unchanged
✔ (d) multiple trailing model turns are all stripped
✔ (e) never strips contents down to empty
✔ empty/missing contents does not throw
7 pass, 0 fail
```

## Gates run

- `npx eslint --suppressions-location config/quality/eslint-suppressions.json open-sse/executors/antigravity.ts tests/unit/antigravity-claude-prefill-strip.test.ts` — clean
- `node scripts/check/check-file-size.mjs` — no new violations on touched files
- `node scripts/check/check-test-discovery.mjs` — OK, test file discovered
- Targeted suite `node --import tsx/esm --test tests/unit/antigravity-claude-prefill-strip.test.ts` — 7/7 pass
- `node scripts/check/check-complexity.mjs`, `check-cognitive-complexity.mjs`, and the full `npm run test:unit` were kicked off but did not finish within this session due to extreme host contention (13 parallel worktree sessions running the same checks concurrently, load average ~209). No failures observed in the partial output before the session ended; the change is a small, additive conditional branch unlikely to move complexity/cognitive thresholds. Recommend a follow-up CI run confirms these before merge.

## Notes

- Hard Rule #18 live-VPS validation (400→200 against a real Antigravity connection) was
  not performed — no reachable live Antigravity connection in this session. The TDD
  fail→pass proof above is the validation path per the plan file's own fallback.
- Claude path (`isClaude` branch, PR #6114) is untouched — byte-identical.

⚠️ base-red inherited: #9985 — ESLint errors (2) from #10250 (unrelated to this change, do not chase).


diegosouzapw:
The focused suite passes (7/7), and the Claude path remains covered. I found one plausible correctness concern: `upstreamModel.includes("gemini")` now strips the trailing model turn for every Gemini-named Antigravity model, including catalogued Gemini 2.5/image IDs, while the issue evidence is specifically about newer endpoints; please narrow or justify that scope. Also, issue #10104's validation plan calls out Gemini 3.6 Flash tiers, but the added tests cover only Gemini 3.1 Pro and `gemini-3-flash-agent`. The PR notes that no live Antigravity 400→200 check was possible; the focused TDD evidence should remain the basis for the claim unless live validation is added.

PR #10438, opened 2026-08-15
fix(open-sse): stop concurrent requests colliding on dedup hash for non-OpenAI formats

### LINKED ISSUE #10249: [BUG] Concurrent requests to the same combo occasionally receive identical responses
### OmniRoute Version

v3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu (Docker host)

### Node.js Version

_No response_

### Provider(s) Involved

Multiple providers routed through the same combo. The issue does not appear to be limited to one provider.

### Model(s) Involved

Multiple models configured as targets of the affected combo

### Client Tool

cURL

### Description

When multiple requests are sent concurrently to the same combo, OmniRoute occasionally returns the same response for different requests.

The requests contain different prompts or document inputs, but two or more callers sometimes receive an identical response. The duplicated response appears to belong to only one of the requests.

The problem is intermittent and is more likely to happen when several requests target the same combo at approximately the same time.

I have not yet confirmed whether this is caused by response correlation, session affinity, semantic cache, or combo routing. There is no obvious error response when it happens.

### Steps to Reproduce

1. Create a combo containing multiple provider/model targets.
2. Send two or more concurrent requests to `/v1/chat/completions`.
3. Use the same combo name as `model` for every request.
4. Send different prompts or different document inputs in each request.
5. Compare the returned responses.
6. Occasionally, multiple requests return the same response even though their input payloads are different.

The issue is intermittent and may require several concurrent attempts to reproduce.

### Expected Behavior

Every request should receive the response generated for its own request payload.

Concurrent requests using the same combo must remain isolated and should be correlated independently, even when they are routed to the same provider, model, account, or connection.

### Actual Behavior

Two or more concurrent requests occasionally receive an identical response. One response may appear to be reused or delivered to another request that had different input.

### Test Impact

Needs a new e2e test

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

Environment:

- OmniRoute: v3.8.49
- Deployment: official Docker image
- Host OS: Ubuntu Linux
- Endpoint: `/v1/chat/completions`
- Client: RelayHub
- Requests may include large PDF/document inputs encoded as base64
- Requests are usually non-streaming
- The same combo name is used by multiple concurrent requests

I can provide correlation IDs and sanitized request logs when the issue happens again. No API keys, cookies, or document contents will be included.

### Validation Plan

_No response_

diegosouzapw:
Closes #10249

## Root cause

`computeRequestHash()` in `open-sse/services/requestDedup.ts` projected the prompt
content from `body.messages` only. The dedup site in `open-sse/handlers/chatCore.ts`
(~line 2708) hashes the **translated** (target-format) request body, and non-OpenAI
target formats don't carry a `messages` field:

- Gemini-translated bodies use `contents` (`open-sse/translator/request/openai-to-gemini.ts`)
- Responses-API-translated bodies use `input` (`open-sse/translator/request/openai-responses.ts`)

So for those formats `messages` was always `undefined`, every prompt hashed to the same
`null`-backed value for a given `provider/model`, and two concurrent requests with
**different prompts** joined the same in-flight dedup promise — the second caller
silently received the first caller's response verbatim. This matches the reporter's
symptom exactly: identical responses to different prompts, intermittent, concurrency
to the same combo (which pins the same provider/model target).

## Fix

`messages: body.messages ?? body.contents ?? body.input ?? null` instead of only
`body.messages ?? null`. Rest of the canonical hash projection (model, temperature,
tools, tool_choice, max_tokens, response_format) is unchanged. The semantic-cache
signature path (`src/lib/semanticCache.ts::generateSignature`) already handles
`messages`/`contents` correctly via `normalizeConversation` and was not touched —
confirmed unrelated, per the plan-file.

## Regression test — `tests/unit/request-dedup-10249.test.ts`

4 cases, adapted from the plan-file's TDD repro:
1. Gemini-format (`contents`) bodies with different prompts must NOT collide.
2. Responses-API-format (`input`) bodies with different prompts must NOT collide.
3. OpenAI-format (`messages`) sanity case — proves the fix is additive, not a
   behavior change for the format that already worked.
4. Genuinely identical concurrent requests still dedupe to a single upstream call
   (the perf feature this fix must not disable).

**RED → GREEN evidence** (verified against this exact diff, not just the plan-file's
prior run): reverting the fallback chain to `body.messages ?? null` on the current
tree reproduces the original collision —

```
✖ Responses-API input-format translated bodies with different prompts must NOT collide
  AssertionError: Different prompts must have different dedup hashes
  actual: 'dc16d5b74e0d12c8', expected: 'dc16d5b74e0d12c8'
```

Restoring the fix: all 4 tests pass.

```
node --import tsx/esm --test tests/unit/request-dedup-10249.test.ts
✔ Gemini-format translated bodies with different prompts must NOT collide on dedup hash
✔ Responses-API input-format translated bodies with different prompts must NOT collide
✔ Sanity: OpenAI-format bodies with different prompts DO get distinct hashes (unchanged behavior)
✔ Genuinely identical requests still hash identically and get deduplicated (perf feature preserved)
ℹ tests 4 | ℹ pass 4 | ℹ fail 0
```

## Gates run

- `npx eslint --suppressions-location config/quality/eslint-suppressions.json open-sse/services/requestDedup.ts tests/unit/request-dedup-10249.test.ts` — clean, 0 errors
- `node scripts/check/check-file-size.mjs` — OK, no new frozen-file violations
- `node scripts/check/check-complexity.mjs` — OK, no violations added
- `node scripts/check/check-cognitive-complexity.mjs` — OK, no violations added (1104 vs baseline 1223)
- `node scripts/check/check-test-discovery.mjs` — OK, new test file discovered (4560+ test files tracked)
- `npm run test:unit` — 572/572 passed, 0 failures, before the run stalled on an unrelated shared-DATA_DIR contention hang (see note below) and was terminated; no failure at any point touched `requestDedup.ts` or dedup behavior
- `npm run test:vitest` — partial run only surfaced pre-existing, environment-driven failures (DB migration-renumbering errors from `~/.omniroute/` being hammered by sibling worktree sessions in a 13-way parallel fan-out) in files unrelated to this change (`audit.test.ts`, `encryption.spec.ts`, `autoCombo/__tests__/chaosVirtualCombo.test.ts` — panel-diversity dedup, a different "dedup" concept). None reference `requestDedup.ts`. The run was terminated after ~45 minutes with no new output for 26+ minutes (infra hang, not a test failure) due to extreme shared-machine load from concurrent sibling sessions. CI runs in an isolated environment and is the authoritative full-suite gate for this PR.

⚠️ base-red inherited: #9985 (ESLint errors (2) from #10250 i18n PT-PT translation) — pre-existing on `release/v3.8.50`, unrelated to this change, not touched here.

### DIFF
--- a/changelog.d/fixes/10249-dedup-hash-collision.md
+++ b/changelog.d/fixes/10249-dedup-hash-collision.md
@@ -1,0 +1,1 @@
+- fix(open-sse): stop concurrent requests colliding on the same dedup hash for non-OpenAI target formats (#10249)
--- a/open-sse/services/requestDedup.ts
+++ b/open-sse/services/requestDedup.ts
@@ -36,12 +36,19 @@ const inflight = new Map<string, Promise<unknown>>();
  * Compute a deterministic hash for a request body.
  * Includes: model, messages, temperature, tools, tool_choice, max_tokens, response_format
  * Excludes: stream, user, metadata (don't affect LLM output)
+ *
+ * The prompt content can live under different keys depending on the target
+ * provider format the body has already been translated to: OpenAI-style
+ * bodies use `messages`, Gemini-translated bodies use `contents`, and
+ * Responses-API-translated bodies use `input`. Falling back to only
+ * `messages` made every non-OpenAI-format body hash the prompt as `null`,
+ * colliding different prompts onto the same dedup hash (#10249).
  */
 export function computeRequestHash(requestBody: unknown): string {
   const body = requestBody as Record<string, unknown>;
   const canonical = {
     model: body.model ?? null,
-    messages: body.messages ?? null,
+    messages: body.messages ?? body.contents ?? body.input ?? null,
     temperature: typeof body.temperature === "number" ? body.temperature : 1.0,
     tools: body.tools ?? null,
     tool_choice: body.tool_choice ?? null,
--- a/open-sse/services/requestDedup.ts
+++ b/open-sse/services/requestDedup.ts
@@ -32,23 +32,110 @@ export interface DedupResult<T> {
 
 const inflight = new Map<string, Promise<unknown>>();
 
+function asRecord(value: unknown): Record<string, unknown> | null {
+  return value !== null && typeof value === "object" && !Array.isArray(value)
+    ? (value as Record<string, unknown>)
+    : null;
+}
+
+/**
+ * Extract the prompt-bearing content from a (possibly translated) request body.
+ *
+ * The prompt content lives under different keys depending on the target
+ * provider format the body has already been translated to:
+ *   - OpenAI-style bodies (`open-sse/translator/request/*-to-openai.ts`,
+ *     `openai-to-cursor.ts`): `messages`
+ *   - Gemini-translated bodies (`openai-to-gemini.ts`,
+ *     `claude-to-gemini.ts`): `contents`
+ *   - Responses-API-translated bodies (`openai-responses/toResponses.ts`):
+ *     `input`
+ *   - Antigravity-translated bodies (`openai-to-gemini.ts`
+ *     `openaiToAntigravityRequest` / `wrapInCloudCodeEnvelope`): nested under
+ *     `request.contents` (a Cloud Code envelope wrapper)
+ *   - Kiro-translated bodies (`openai-to-kiro.ts` `buildKiroPayload`): nested
+ *     under `conversationState.currentMessage.userInputMessage.content` (the
+ *     current turn) plus `conversationState.history` (prior turns)
+ *
+ * Falling back to only `messages` made every non-OpenAI-format body hash the
+ * prompt as `null`, colliding different prompts onto the same dedup hash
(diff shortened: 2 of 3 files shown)


diegosouzapw:
The top-level fallback fixes the demonstrated Gemini and Responses cases, but it is incomplete. Real translators emit nested prompt roots: Antigravity uses `request.contents` and Kiro uses `conversationState.currentMessage.userInputMessage.content`; `computeRequestHash()` still ignores both, so different concurrent prompts can retain the same hash and share a response. Please canonicalize all translated prompt-bearing shapes (including system/instruction fields) and add regression tests for these real payloads. The focused probes produced identical hashes for different nested Antigravity and Kiro prompts; because they ran against the checked-out base rather than the PR head, this finding is marked PLAUSIBLE.

Issue #723, opened 2026-03-28
[BUG] - Raw <thinking> and <think> tags leaking to CLI clients in streaming/non-streaming responses


k0valik:
#### Environment: OmniRoute 3.1.8 (later releases didn't include fix for this) / npm (global) / Windows 11 / Node 22.22.2 / Providers/Models Involved: z-ai/glm4.7 / ollama-cloud/minimax-m2.7 / qwen/coder-model / ollama-cloud/glm-5 / stepfun-ai/step-3.5-flash  -   Affected CLI tool (tested with OpenCode, saw similar behaviour few days ago with Claude Code)

# **The Issue reported below, the analysis and proposed fixes were created while reproducing the reported bug itself**

## Everything here is generated by the agents investigating the current issue, none of this was verified, but it is a good starting point for another agent to investigate.

Raw thinking tags (`<thinking>...</thinking>`, `</think>...</think>`) are appearing in responses received by CLI tools like Claude and OpenCode. This affects both streaming and non-streaming responses across multiple providers.

**[IMPORTANT] Additionally some initial words from "thinking" context are stripped, example below:** (Should be investigated separately)

<img width="708" height="70" alt="Image" src="https://github.com/user-attachments/assets/e3359d7c-d4d6-411f-bb0c-f44eccf342f8" />
<img width="1822" height="545" alt="Image" src="https://github.com/user-attachments/assets/05b7eefe-93e3-4838-b740-ce37499afc17" />

```
- Should be: "Let me look for..."
- Stripped to "... me look for..."
- See also the <think> tag
- And "EXCELLENT" instead of "...CELLENT"
```

## Impact

- **User Experience**: CLI clients display raw thinking tags instead of clean responses
- **Data Integrity**: Thinking content leaks into response text instead of being properly extracted
- **Provider Coverage**: Affects Kiro, Claude, Responses API, and potentially other providers

## Root Causes

1. **Regex pattern mismatch** in `open-sse/handlers/responseSanitizer.ts:37` - only matches `</think>` not `<thinking>`
2. **Kiro translator** generates `<thinking>` tags that sanitizer cannot process
3. **Responses API SSE** bypasses sanitization entirely
4. **Claude SSE** has partial sanitization (extracts `delta.thinking` but doesn't sanitize `delta.text`)
5. **Conditional sanitization** only runs for OpenAI format in non-streaming path

## Affected Files

- `open-sse/handlers/responseSanitizer.ts` - Regex pattern
[...]

PR #10441, opened 2026-08-15
fix(sse): buffer '<think' partial so split open tags don't leak into content

geek007git:
⚠️ base-red inherited: #9985

## Summary

Fixes an SSE streaming bug where an open `<think` tag split across two deltas could leak into assistant content instead of being parsed as reasoning.

The fix makes the parser derive all valid partial prefixes directly from `THINK_OPEN`, including `<think`.

## Tests

- think-tag parser tests: 8/8 pass
- 12 related think-tag/stream suites: 159/159 pass
- typecheck: clean
- ESLint: clean
- pre-commit gates: pass

## Related Issues

Related to #723.

No migrations, feature flags, or configuration changes.

### DIFF
--- a/open-sse/utils/thinkTagParser.ts
+++ b/open-sse/utils/thinkTagParser.ts
@@ -22,6 +22,15 @@ const THINK_OPEN = "<think>";
 const THINK_CLOSE = "</think>";
 
 /**
+ * Every proper prefix of `<think>` ("<", "<t", ... "<think"), derived from the
+ * tag itself so the list cannot drift out of sync with it.
+ */
+const THINK_OPEN_PARTIALS: readonly string[] = Array.from(
+  { length: THINK_OPEN.length - 1 },
+  (_, i) => THINK_OPEN.slice(0, i + 1)
+);
+
+/**
  * Create the mutable streaming-parse context for one SSE stream.
  * `enabled` decides whether the caller should attempt think-tag parsing at
  * all for this stream (passthrough mode + a provider/model that is known to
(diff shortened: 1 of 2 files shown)


geek007git (commented):
⚠️ base-red inherited: #9985

## Summary

Fixes an SSE streaming bug where an open `<think` tag split across two deltas could leak into assistant content instead of being parsed as reasoning.

The fix makes the parser derive all valid partial prefixes directly from `THINK_OPEN`, including `<think`.

## Tests

- think-tag parser tests: 8/8 pass
- 12 related think-tag/stream suites: 159/159 pass
- typecheck: clean
- ESLint: clean
- pre-commit gates: pass

## Related Issues

Related to #723.

No migrations, feature flags, or configuration changes.

diegosouzapw:
Thanks for the fix — this is a clean, well-targeted catch.

I traced through the old `containsOrMayEndWithThinkOpenTag` logic and confirmed the gap: the hardcoded array topped out at `"<thin"` (5 chars), so a delta that happened to end exactly on `"<think"` (6 chars, no closing `>`) slipped past both checks and leaked raw markup into the assistant content on the next chunk — exactly the scenario your new `applyThinkTag` test reproduces. Deriving `THINK_OPEN_PARTIALS` straight from `THINK_OPEN` is the right fix; it can't drift out of sync again the way the manual list did.

I pulled the branch into an isolated worktree and re-ran things independently:
- `node --import tsx/esm --test tests/unit/think-tag-parser.test.ts` → 8/8 passing, including both new tests.
- `eslint` on the two touched files → clean.

Scope looks right too — you correctly used "Related to #723" rather than "Fixes", since that issue also covers a couple of other root causes (the `responseSanitizer.ts` regex not matching `<thinking>`, and the Kiro translator's `<thinking>` tags) that are outside what this PR touches.

No changes requested from my side. This should just need CI to run (workflow approval) before it's ready to merge — nothing in the diff itself needs rework.

diegosouzapw:

Issue #6842, opened 2026-07-11
feat(providers): OpenRouter quota tracking


steveepreston:
### Problem / Use Case

I'm trying to use OpenRouter (both free and paid accounts) as a tier in my combos, but currently OmniRoute has no quota tracking for OpenRouter — unlike Claude Code (5-hour + weekly window tracking), Codex (5-hour + weekly reset-aware rotation), or Antigravity (provider quota view).

This matters because OpenRouter **does** have real, hittable quota walls — even on free accounts:

- **Free-model rate limits (platform-level, per account):**
  - `:free` model variants are capped at **20 requests/minute**.
  - Daily window: **50 requests/day** if the account has purchased less than **$10** in credits (all-time), raised to **1000 requests/day** once at least $10 has been purchased.
  - Failed attempts still count toward the daily quota.
  - Limits are governed **globally per account** — extra API keys don't help.
- **Credit limits (paid usage):**
  - Account balance can run out (negative balance → `402` errors, which also blocks *free* models).
  - Individual API keys can have optional per-key spending caps (`limit` / `limit_remaining` / `limit_reset`).

Today, when any of these ceilings is hit, OmniRoute only finds out via a hard `429`/`402` failure. There is no "Provider Quota" entry for OpenRouter in the dashboard, no reset-aware routing, and no way to drain the free daily window intentionally before falling back — which is exactly the "track & drain every free token" promise OmniRoute makes for other providers.

### Proposed Solution

Add OpenRouter to the dashboard quota tracker and make routing reset-aware, using OpenRouter's official monitoring APIs.

### 1. Poll OpenRouter's monitoring endpoints

**`GET https://openrouter.ai/api/v1/key`** (Bearer: the account's API key) — per-key limits + usage:
```json
{
"data": {
"label": "sk-or-v1-...",
"limit": 10.0,              // per-key credit cap in USD, null = unlimited
"limit_reset": "monthly",   // how the cap resets, null = never
"limit_remaining": 6.42,    // remaining credits on the key, null = unlimited
"include_byok_in_limit": true,
"usage": 3.58,              // credits used (all time, USD)
"usage_daily": 0.12,        // current UTC day
"usage_weekly": 0.87,       // current UTC week (starts Monday)
"usage_monthly": 3.58,      // current UTC month
"byok_usage": 0,            // same trio exists for BYOK usage
"is_free_tier": true        // false once the user has ever purchased credits
}
}
```
[...]

PR #10445, opened 2026-08-15
fix(sse): let :free OpenRouter models bypass connection-wide credits_exhausted lock

killmonger2317-coder:
## Summary
- A 402 from one paid OpenRouter model correctly locks the whole connection as `credits_exhausted` for an hour (intentional, per #6842's connection-scoped design for OpenRouter's shared account balance). But that lock had no exception for `:free`-suffixed models, which OpenRouter bills separately from paid credits — so once the lock tripped, every `:free` model combo target on that same connection was also skipped, defeating failover to configured free models.
- `getProviderCredentials` (`src/sse/services/auth.ts`) now exempts `:free` model requests from a `credits_exhausted` OpenRouter connection specifically, while still refusing paid-model requests on that connection and still refusing `:free`-model requests on connections that are terminal for another reason (banned/expired) or on non-OpenRouter providers.

## Test plan
- [x] TDD: `tests/unit/openrouter-free-model-credits-exhausted.test.ts` reproduces the bug (fails before the fix), covers the free-model exemption plus 3 guard cases (paid model still blocked, banned connection still blocked, non-OpenRouter provider still blocked). All 4 pass after the fix.
- [x] `node --import tsx/esm --test` on related terminal-status/auth/combo suites (auth-terminal-status, openrouter-quota-6842, auth-disable-cooling-2997, quota-connection-recovery, repro-6692-sticky-terminal, repro-7387-sticky-quota-exhausted, combo-pin-health-gate, combo-session-stickiness, sse-auth, startup-stale-cooldown-recovery) — 160/160 pass, no regressions.
- [x] `npx tsc --noEmit -p tsconfig.typecheck-core.json` clean on changed files.
- [x] `npx eslint` clean on changed files (one pre-existing unrelated violation at auth.ts:128, not touched by this diff).

### DIFF
--- a/src/sse/services/auth.ts
+++ b/src/sse/services/auth.ts
@@ -81,6 +81,7 @@ import {
   WEB_COOKIE_PROVIDERS,
 } from "@/shared/constants/providers";
 import { isModelExcludedByConnection } from "@/domain/connectionModelRules";
+import { isFreeModel } from "@/shared/utils/freeModels";
 import {
   applySessionAffinityPin,
   formatSessionKeyForLog,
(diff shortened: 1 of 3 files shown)


diegosouzapw:
Thank you @killmonger2317-coder — the `:free`-model exemption on a `credits_exhausted` OpenRouter connection is a real failover bug, and the 4 tests covering paid/banned/non-openrouter are exactly right. One housekeeping item before merge: this PR targets `release/v3.8.49`, but the active release branch is `release/v3.8.50` — I have retargeted it. Ship 🚀

Issue #10171, opened 2026-08-12
[BUG] Instrumentation hook fails with "Unsupported platform: android" cache-check error on native Windows AND native WSL2 Ubuntu — server reports "running" but all requests return 500


aarish-ai:
### OmniRoute Version

v3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Ubuntu (WSL) - (Installed and tested on Windows 11 Powershell too)

### Node.js Version

v22.23.2

### Provider(s) Involved

N/A — issue occurs before reaching any client tool (Claude Code), at the OmniRoute server level itself

### Model(s) Involved

Groq (configured during WSL2 setup); none configured during Windows testing

### Description

`omniroute serve` prints a misleading "Next.js instrumentation failed on Android/Termux (likely missing cache dir)" warning and then falsely reports "✔ OmniRoute is running!" immediately afterward. In reality, the instrumentation hook never loads, no logs are ever written (app.log stays completely empty even with APP_LOG_LEVEL=debug and DEBUG=* set), and every HTTP request to the server returns a bare HTTP 500 Internal Server Error with no JSON body or detail.

This occurs on v3.8.49 across multiple genuinely non-Android/Termux environments:
1. Native Windows 10/11 (PowerShell, Node v24.15.0)
2. Native Windows with Node downgraded to v22.x via nvm-windows
3. WSL2 Ubuntu (genuine Linux kernel, not Termux), fresh install, Node v22.x via NodeSource

The error message references TERMUX_GUIDE.md and suggests creating ~/.cache as a fix. This workaround does NOT resolve the issue in any of the above environments, including when:
- ~/.cache (or the Windows equivalent, C:\Users\<user>\.cache) already exists and is confirmed writable
- XDG_CACHE_HOME is explicitly set to a fresh, writable, non-default directory
- The server is run as Administrator (Windows) / with correct user permissions (WSL)
- A completely fresh install of the latest npm version is used
- No stray Termux/Android-related environment variables (ANDROID_ROOT, PREFIX, TERMUX_VERSION) are present

Since this reproduces on genuine native Linux (WSL2 Ubuntu), not just Windows, the root cause appears to be broader than platform-detection misfiring on Termux specifically — either the underlying instrumentation hook is broken more generally in this release, or the cache-directory check itself is defective regardless of platform.

### Steps to Reproduce

Environment A — Native Windows:
[...]

Issue #7773, opened 2026-07-19
fix(startup): Termux (Android) — HTTP 500 on fresh install, storage.sqlite never created (better-sqlite3 native binary missing)


sillacarajo-ux:
### OmniRoute Version

 3.8.48

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Termux (Android 14, installed from F-Droid)

### Node.js Version

V26.4.0

### Provider(s) Involved

None

### Model(s) Involved

None

### Client Tool

None

### Description

This issue occurs on Termux running on Android 14 (installed from F-Droid). I selected "Linux" because Android is not available in the Operating System dropdown.
Fresh installation on Termux starts successfully and reports "OmniRoute is running", but every request to the dashboard returns HTTP 500 Internal Server Error.

The SQLite database is never created.

Running "omniroute doctor" reports:

- SQLite database not found
- better-sqlite3 native binary was not found

The directory ~/.omniroute only contains:

- .env
- logs/
- server/

storage.sqlite is never created.

### Steps to Reproduce

1. Install OmniRoute globally:

npm install -g omniroute

2. Start OmniRoute:

omniroute

3. Open:

http://127.0.0.1:20128

or run:

curl -i http://127.0.0.1:20128

### Expected Behavior

The dashboard should open normally and create storage.sqlite automatically during first startup.

### Actual Behavior

Every request returns:

HTTP/1.1 500 Internal Server Error

storage.sqlite is never created.

### Test Impact

Unsure

### Error Logs / Output

```shell
omniroute doctor

OK   Config: .env found
WARN Database: SQLite database not found
OK   Storage/encryption: Encryption key is configured; database not initialized yet
WARN Native binary: better-sqlite3 native binary was not found
OK   Node runtime: v26.4.0 is supported
OK   Server liveness: Server reachable (health endpoint returned 500)
curl -i http://127.0.0.1:20128

HTTP/1.1 500 Internal Server Error
ls /data/data/com.termux/files/usr/lib/node_modules/omniroute/node_modules | grep better

json-parse-even-better-errors
Attempting to compile better-sqlite3 manually:

gyp: Undefined variable android_ndk_path in binding.gyp
```

### Screenshots

None

### Additional Context

- Samsung Galaxy S21 (SM-G991B)
- Termux installed from F-Droid
- Removed ~/.omniroute
- Reinstalled OmniRoute
- Cleared npm cache
- Installed better-sqlite3 globally
- app.log remains empty
[...]

Issue #6560, opened 2026-07-07
fix(startup): Cannot create property 'message' on string 'Database closed' at registerInstrumentation


diegosouzapw:
## Summary

During an update / restart, OmniRoute crashes at instrumentation bootstrap with:

```
level":"error","component":"app"
"message":"Cannot create property 'message' on string 'Database closed'"
TypeError: Cannot create property 'message' on string 'Database closed'
    at registerInstrumentation
```

## Root cause (hypothesis)

A raw **string** (`'Database closed'`) is thrown/propagated from the DB layer during the
instrumentation registration phase (`src/instrumentation.ts` → `registerInstrumentation`).
The surrounding error handler assumes an `Error` object and does something like
`err.message = ...`, which throws `TypeError: Cannot create property 'message' on string`
because you cannot set a property on a primitive string.

This is triggered by a **`Database closed`** condition — i.e. the SQLite handle is being
torn down / not yet open when instrumentation runs during an update-driven reload.

## Expected

- Non-`Error` thrown values must be normalized to an `Error` before any `.message`
  assignment (`err = err instanceof Error ? err : new Error(String(err))`).
- The instrumentation bootstrap should be resilient to a closed/closing DB during a
  reload (retry/await open, or skip gracefully) instead of crashing the app.

## Repro (to confirm)

Trigger an in-place update/reload while the DB singleton is mid-teardown so that a DB
call during `registerInstrumentation` rejects with `Database closed`.

## Notes

Reported via community (WhatsApp). Log sanitized (no personal data).

diegosouzapw:
Fixed in `release/v3.8.47` via #6622 — the crash was a **secondary** TypeError masking the real error: sql.js's WASM adapter throws the bare string `"Database closed"` when a query hits an already-closed handle, and `preInitSqlJs()` was returning a stale `globalThis`-cached adapter that a prior `gracefulShutdown`/`resetDbInstance` had already closed (an update/restart races the close against the next boot). Next.js's `registerInstrumentation()` then does `err.message = …` on that primitive string → `TypeError: Cannot create property 'message' on string`, crashing every boot. Fix: (1) `preInitSqlJs()` evicts a closed cached adapter and recreates it; (2) `instrumentation-node.ts` normalizes any non-Error throw to a real Error and retries once on a transient "database closed". Regression guard: `tests/unit/instrumentation-database-closed-6560.test.ts` (7). Ships in v3.8.47.

PR #10447, opened 2026-08-15
fix(cli): guarantee non-empty [STARTUP] Fatal log on instrumentation-hook boot throw

diegosouzapw:
## Summary

Partial diagnostic hardening for #10171 (native Windows / WSL2 boot failure — instrumentation
hook crashes, HTTP listener stays up, every DB-touching route 500s, `app.log` stays completely
empty). Per the plan-file (`_tasks/pipeline/bugs/2-implementing/10171-instrumentation-hook-500-on-windows-wsl.plan.md`,
verdict `needs-vps`), the true root cause on native Windows/WSL2 requires capturing raw child
stderr on a real host — **that part is still open and out of scope here.**

This PR implements the one safe, narrowly-scoped, testable improvement identified in the plan's
implementation checklist that is independent of the unknown platform-specific root cause:

> Confirm #7773 `[STARTUP] Fatal` log is reached; if not, add a pre-guard (module-load) fatal log
> path so `app.log` is never empty on a failed boot. Add a unit test asserting a boot-instrumentation
> failure always yields a non-empty `[STARTUP] Fatal:` line.

`src/instrumentation-node.ts::ensureDbReadyForBoot` (#7773/#7828) already prints
`[STARTUP] Fatal: Database driver initialization failed:` for one specific failure class (DB
driver init), but nothing guaranteed a fatal throw ANYWHERE during instrumentation-hook boot —
including a throw that happens before `ensureDbReadyForBoot` is even reached (e.g. a module-load
error importing `./instrumentation-node` itself) — produces a diagnostic line. `register()` in
`src/instrumentation.ts` now wraps the boot call in a try/catch at the outermost boundary and
unconditionally logs a `[STARTUP] Fatal: instrumentation hook failed during boot:` line to stdout
before rethrowing, so `app.log`/stdout is never silently empty on a failed boot, regardless of
platform or which step threw.

**Explicitly not fixed here** (needs the reporter's raw stderr from a real Windows/WSL2 host,
per the plan file):
- The actual root cause of why the instrumentation hook throws on native Windows / WSL2.
- Any DB-driver-cascade (`better-sqlite3` / `node:sqlite` / `sql.js`) fix specific to that platform.

Out of scope, confirmed correct, untouched: `needsAndroidCacheDirPrep` / `isTermux`
(`bin/cli/utils/ensureAndroidCacheDir.mjs`, `scripts/build/postinstallSupport.mjs`).

## Regression test

`tests/unit/instrumentation-hook-boot-fatal-log-10171.test.ts` — 3 cases:
1. A boot throw (`Error`) during instrumentation-hook boot is logged with a non-empty
   `[STARTUP] Fatal:` line containing the failure message, before rethrow.
2. A non-`Error` throw (e.g. bare `throw "string"`, mirroring `sql.js`'s #6560 pattern) is also
   logged.
3. A clean successful boot logs nothing.

Fail → pass evidence (ran against the pre-fix `register()`, then the fixed version):
```
# unfixed (git show HEAD:src/instrumentation.ts)
✖ #10171: any instrumentation-hook boot throw is logged ... (18620ms, real DB init side effect)
✖ #10171: a non-Error throw during instrumentation-hook boot is still logged ...
✖ register() does not log anything on a clean successful boot
fail 3

# fixed (working tree)
✔ #10171: any instrumentation-hook boot throw is logged ...
✔ #10171: a non-Error throw during instrumentation-hook boot is still logged ...
✔ register() does not log anything on a clean successful boot
pass 3 / fail 0
```

## Gates run

- `npx eslint --suppressions-location config/quality/eslint-suppressions.json src/instrumentation.ts tests/unit/instrumentation-hook-boot-fatal-log-10171.test.ts` — clean, no output
- `node scripts/check/check-file-size.mjs` — `src/instrumentation.ts` is 42 lines, no new god-file
- `node scripts/check/check-test-discovery.mjs` — OK, new test file discovered
- Complexity/cognitive-complexity ratchet (`eslint.complexity-ratchets.config.mjs`) scoped to
  `src/instrumentation.ts` — zero violations
- `node --import tsx/esm --test tests/unit/instrumentation-hook-boot-fatal-log-10171.test.ts` — 3/3 pass

Refs #10171 (not Closes — this is a partial diagnostic hardening; the platform-specific root
cause still needs a real Windows/WSL2 host's raw child stderr per the plan file, tracked as
`hold-vps`).

⚠️ base-red inherited: #9985 — ESLint errors (2) from #10250 (i18n PT-PT translation), unrelated
to this change.


diegosouzapw:
Please normalize the value caught at src/instrumentation.ts:37-39 before rethrowing. A raw non-Error is logged, but Next 16's instrumentation wrapper immediately assigns err.message and throws a secondary TypeError, masking the original boot failure. The new test currently only checks that rejection occurs; assert that the rejection is an Error and remains message-augmentable. I could not execute the PR head in this read-only worktree, so this finding is plausible rather than confirmed.

PR #10450, opened 2026-08-15
fix(sse): mark gemini-3.5-flash as thinking-capable

### LINKED ISSUE #10286: [BUG][FIX INCLUDED] Gemini 3.5 Flash incorrectly marked with supportsThinking: false
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

24.11.0

### Provider(s) Involved

Google Ai Studio

### Model(s) Involved

gemini 3.5 flash

### Client Tool

_No response_

### Bug

gemini-3.5-flash is currently treated by OmniRoute as a model that does not support thinking/reasoning, even though the target model supports reasoning.

When I send a request to gemini-3.5-flash with a reasoning setting such as:

```
{
  "reasoning_effort": "high"
}
```

OmniRoute rejects the request locally, before sending it to the provider, with:

```
API Error: 400 Reasoning effort 'high' is not supported by the configured target
```

The same happens with other reasoning effort values such as xhigh.

### Root cause

In:

src/shared/constants/modelSpecs.ts

the shared Gemini 3.5 Flash model spec was defined as:

```
const GEMINI_35_FLASH_MODEL_SPEC = {
  maxOutputTokens: 65536,
  contextWindow: 1048576,
  supportsThinking: false,
  supportsTools: true,
  supportsVision: true,
} satisfies ModelSpec;

```

Because gemini-3.5-flash uses this spec:

```
"gemini-3.5-flash": {
  ...GEMINI_35_FLASH_MODEL_SPEC,
  aliases: ["gemini-3.5-flash-high"],
},

```

the reasoning routing capability resolver sees:

`supportsThinking === false`

and marks the target as unsupported, causing the request to be rejected before provider translation.

### Fix / Verification

I built OmniRoute myself from source and changed only:

```
 const GEMINI_35_FLASH_MODEL_SPEC = {
   maxOutputTokens: 65536,
   contextWindow: 1048576,
-  supportsThinking: false,
+  supportsThinking: true,
   supportsTools: true,
   supportsVision: true,
 } satisfies ModelSpec;

```

After rebuilding OmniRoute and testing the same request again, the local 400 error disappeared and gemini-3.5-flash accepted the reasoning request correctly.

So the minimal fix appears to be changing:

`supportsThinking: false`

to:

`supportsThinking: true`

for GEMINI_35_FLASH_MODEL_SPEC.

### Steps to Reproduce

1. Enable thinking/reasoning for `gemini-3.5-flash`.
2. Send a request with reasoning effort set (for example `high`).
3. OmniRoute returns:
   `API Error: 400 Reasoning effort 'high' is not supported by the configured target`

### Expected Behavior

`gemini-3.5-flash` should be recognized as a thinking-capable model and the request should be forwarded normally.

### Actual Behavior

OmniRoute rejects the request before sending it to the provider because `GEMINI_35_FLASH_MODEL_SPEC` has `supportsThinking: false`.

Changing it to `supportsThinking: true` in `src/shared/constants/modelSpecs.ts` fixes the issue. I built OmniRoute from source and verified the fix.

### Test Impact

Unsure

### Error Logs / Output

```shell
API Error: 400 Reasoning effort 'high' is not supported by the configured target
```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

diegosouzapw:
Closes #10286

## Root cause

`src/shared/constants/modelSpecs.ts` declares the shared `GEMINI_35_FLASH_MODEL_SPEC`
constant with `supportsThinking: false`. The base `"gemini-3.5-flash"` entry spreads
that spec, so `getResolvedModelCapabilities()` reports the base Google AI Studio model
as thinking-incapable, which makes `capabilityFor()`
(`src/lib/reasoningRouting/policy.ts`) resolve to `"unsupported"` for any
`reasoning_effort`, and `applyDecision()` (`src/sse/handlers/reasoningRouting.ts`)
turns that into a pre-provider HTTP 400 — even though the base model supports
reasoning (it has an effort-tier alias `gemini-3.5-flash-high`, and the reporter
independently verified flipping the flag to `true` works end-to-end).

The same shared `GEMINI_35_FLASH_MODEL_SPEC` constant is also spread into several
Antigravity flash-tier aliases (`gemini-3.5-flash-low`, `gemini-3.5-flash-extra-low`,
`gemini-3-flash-agent`, `gemini-3.6-flash-*`), and those tier ids are documented as
rejecting client-supplied thinking params (the model id itself selects the reasoning
tier upstream). So the fix does **not** flip the shared constant — it sets
`supportsThinking: true` as an explicit override on the base `"gemini-3.5-flash"`
entry only, leaving `GEMINI_35_FLASH_MODEL_SPEC` and all the Antigravity tier aliases
unchanged (still `false`).

## Fix

`src/shared/constants/modelSpecs.ts`: added `supportsThinking: true` override on the
base `"gemini-3.5-flash"` entry (after the `...GEMINI_35_FLASH_MODEL_SPEC` spread).

## Regression test

`tests/unit/gemini-3-5-flash-thinking.test.ts` — TDD, fails on unfixed code
(`AssertionError: false !== true` / decision.capability `'unsupported'` vs
`'supported'`), passes after the fix:

```
✔ gemini-3.5-flash (AI Studio provider) resolves as thinking-capable
✔ reasoning_effort 'high' on gemini-3.5-flash is NOT rejected by routing policy
```

Asserts both:
1. `getResolvedModelCapabilities({provider:"gemini", model:"gemini-3.5-flash"}).supportsThinking === true`
2. reasoning-routing policy resolves `reasoning_effort:"high"` on `gemini-3.5-flash` to
   capability `"supported"` (not the pre-provider 400).

## Gates run

- `node scripts/check/check-file-size.mjs` — no new violation on `modelSpecs.ts`
- `node scripts/check/check-complexity.mjs` — OK (2456 violations, baseline 2774)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1104 violations, baseline 1223)
- `node scripts/check/check-test-discovery.mjs` — OK, new test file discovered
- `npm run test:unit` — started full suite; 614/4565 test files completed with 0
  failures before the run stalled to near-zero throughput under severe 13-way
  parallel devbox contention (13 concurrent fan-out agents running the identical
  full suite on shared hardware — not specific to this change). No regression
  signal in the portion that ran. CI runs on isolated runners and is the
  authoritative full-suite gate for this PR.

⚠️ base-red inherited: #9985 — ESLint errors (2) from #10250

### DIFF
--- a/changelog.d/fixes/10286-gemini-3-5-flash-thinking.md
+++ b/changelog.d/fixes/10286-gemini-3-5-flash-thinking.md
@@ -1,0 +1,1 @@
+- fix(sse): mark gemini-3.5-flash as thinking-capable so reasoning_effort is no longer rejected with a spurious 400 (#10286)
--- a/src/shared/constants/modelSpecs.ts
+++ b/src/shared/constants/modelSpecs.ts
@@ -226,8 +226,16 @@ export const MODEL_SPECS: Record<string, ModelSpec> = {
   },
 
   // ── Gemini 3.5 Flash ─────────────────────────────────────────────
+  // #10286: the base Google AI Studio model DOES support reasoning (it has
+  // an effort-tier alias gemini-3.5-flash-high) — override the shared spec's
+  // supportsThinking:false here only. Do NOT flip GEMINI_35_FLASH_MODEL_SPEC
+  // itself: it is also spread into the Antigravity flash-tier aliases
+  // (gemini-3.5-flash-low/-extra-low, gemini-3-flash-agent, gemini-3.6-flash-*)
+  // which reject client-supplied thinking params because the model id itself
+  // selects the reasoning tier upstream.
   "gemini-3.5-flash": {
     ...GEMINI_35_FLASH_MODEL_SPEC,
+    supportsThinking: true,
     aliases: ["gemini-3.5-flash-high"],
   },
 
--- a/tests/unit/gemini-3-5-flash-thinking.test.ts
+++ b/tests/unit/gemini-3-5-flash-thinking.test.ts
@@ -1,0 +1,76 @@
+// Regression test for #10286: gemini-3.5-flash was incorrectly marked
+// supportsThinking:false, causing a spurious pre-provider HTTP 400 for any
+// request with reasoning_effort set, even though the base Google AI Studio
+// model supports reasoning (it has an effort-tier alias gemini-3.5-flash-high).
+import test from "node:test";
+import assert from "node:assert/strict";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+
+const TEST_DATA_DIR = fs.mkdtempSync(path.join(os.tmpdir(), "omniroute-repro-10286-"));
+process.env.DATA_DIR = TEST_DATA_DIR;
+process.env.API_KEY_SECRET = process.env.API_KEY_SECRET || "test-repro-10286-secret";
+
+const caps = await import("../../src/lib/modelCapabilities.ts");
+const core = await import("../../src/lib/db/core.ts");
+const rulesDb = await import("../../src/lib/db/reasoningRoutingRules.ts");
+const policy = await import("../../src/lib/reasoningRouting/policy.ts");
+
+async function resetStorage() {
+  core.resetDbInstance();
+  fs.rmSync(TEST_DATA_DIR, { recursive: true, force: true });
+  fs.mkdirSync(TEST_DATA_DIR, { recursive: true });
+  rulesDb.invalidateReasoningRoutingRuleCache();
+}
+
+function ruleInput(patch: Record<string, unknown> = {}) {
+  return {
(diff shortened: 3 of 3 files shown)


diegosouzapw:
Reviewed PR #10450 against release/v3.8.50. No verified defects found. The override is correctly limited to the native gemini-3.5-flash MODEL_SPECS entry; GEMINI_35_FLASH_MODEL_SPEC remains false for Antigravity tier IDs, and gemini-web's explicit supportsReasoning:false remains authoritative. The regression test covers both resolved capability and reasoning-routing policy. I could not execute the test in a checkout with PR-head path aliases, so this is merge-ready with PLAUSIBLE confidence rather than a confirmed runtime result. Base-red #9985 is inherited and not attributed to this PR.

diegosouzapw:

PR #10451, opened 2026-08-15
test(build): make standalone fixture paths portable on Windows

backryun:
## Summary

- resolve the standalone fixture repository root with `fileURLToPath(import.meta.url)`
- avoid interpreting a Windows file URL pathname as a native path and producing `C:\\C:\\Users\\...`
- keep the production standalone assembler unchanged

## Root cause

Physical Windows x64 validation for #10321 reproduced the combined standalone test failure as an `ENOENT` under a duplicated drive path. `new URL(".", import.meta.url).pathname` returns a URL pathname, not a Windows filesystem path; passing it directly to `path.resolve()` duplicates the drive prefix.

## Validation

- Windows x64 before fix: `assemble-standalone.test.ts` fails with `C:\\C:\\Users\\...` (recorded in the #10321 Stage 8 checkpoint)
- `node --import tsx/esm --test tests/unit/build/assemble-standalone.test.ts` — 5/5 pass on macOS arm64
- `npx prettier --check tests/unit/build/assemble-standalone.test.ts` — pass
- `git diff --check` — pass

A Windows x64 rerun is in progress and will be appended when available.

Refs #10321

### DIFF
--- a/tests/unit/build/assemble-standalone.test.ts
+++ b/tests/unit/build/assemble-standalone.test.ts
@@ -3,6 +3,7 @@ import assert from "node:assert/strict";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
+import { fileURLToPath } from "node:url";
 import {
   assembleStandalone,
   patchTurbopackChunks,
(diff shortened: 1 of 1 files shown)


diegosouzapw:
Thanks @backryun — the `fileURLToPath` fix for the standalone-fixture root is correct and the change is test-only, so no runtime risk. Good Windows-portability hardening. 🚀

PR #10452, opened 2026-08-15
fix(dashboard): send periodic WS heartbeat pings to stop live-dashboard reconnect churn

### LINKED ISSUE #10319: [BUG] Live dashboard WS client never sends heartbeat ping — server terminates every ~35s
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

_No response_

### Node.js Version

_No response_

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

## Symptom

Browser console (behind a reverse proxy), every ~35–45s:

WebSocket connection to 'wss://…/live-ws' failed: The network connection was lost.

The connection flaps continuously (auto-reconnect loop). Live feed still works, but with constant reconnect churn and console noise.

## Root cause

The heartbeat contract is declared in src/server/ws/liveServer.ts:

// Client → Server: { type: "ping" }
// Server → Client: { type: "pong" }

but the client never sends ping:

- Server (src/server/ws/liveServer.ts): HEARTBEAT_TIMEOUT_MS = 35_000. startHeartbeat() calls client.ws.terminate() when now - client.lastActivity > 35s; lastActivity is only updated on inbound client messages.
- Client (src/hooks/useLiveDashboard.ts): onopen sends only { type: "subscribe", channels }; onmessage treats pong as a no-op; there is no setInterval and no { type: "ping" } send anywhere.

So lastActivity stops advancing right after subscribe, and the server terminate()s (abnormal close, no close frame) at the next heartbeat sweep after 35s → browser shows "network connection was lost" → auto-reconnect → loop.

## Impact

Every live-dashboard WS connection is force-dropped ~35–45s after connect and re-established. Not fatal (auto-reconnect works), but spams the console and wastes reconnect churn on every dashboard tab.

## Repro / version

Observed in v3.8.49 in production (Docker diegosouzapw/omniroute:latest behind Nginx Proxy Manager). Code is unchanged on main and release/v3.8.50 (checked src/server/ws/liveServer.ts and src/hooks/useLiveDashboard.ts).

## Additional inconsistency

In startHeartbeat() the server sends { type: "pong" } every 15s with the comment // Send ping. That unsolicited pong keeps proxies from idling out, but it doesn't reset lastActivity (inbound-only), so it can't prevent the terminate either.

## Suggested fix

- Client sends { type: "ping" } every ~15s (align with HEARTBEAT_INTERVAL_MS), and/or
- Server resets liveness on its own outbound heartbeat instead of relying solely on inbound activity.

### Steps to Reproduce

1. Run OmniRoute v3.8.49 (Docker diegosouzapw/omniroute:latest) behind a reverse proxy,
   with NEXT_PUBLIC_LIVE_WS_PUBLIC_URL=wss://…/live-ws and LIVE_WS_ALLOWED_ORIGINS set for the public origin.
2. Open the dashboard and keep the tab in the foreground.
3. Watch the browser console (or `docker logs omniroute`).
4. The connection drops ~35–45s after connect, then auto-reconnects, in a loop.

### Expected Behavior

The live WS connection stays up indefinitely: the client sends {type:"ping"} periodically,
the server replies {type:"pong"}, and an otherwise-connected client is never terminated for being idle.

### Actual Behavior

src/hooks/useLiveDashboard.ts sends only {type:"subscribe", channels} on open and never sends "ping".
src/server/ws/liveServer.ts calls client.ws.terminate() when now - lastActivity > 35s
(HEARTBEAT_TIMEOUT_MS = 35_000), and lastActivity is updated only on inbound client messages.
So every client is terminated ~35–45s after connect (abnormal close, no close frame).
The client auto-reconnects, so the feed keeps working but with constant reconnect churn.

### Test Impact

Needs a new integration test

### Error Logs / Output

```shell
Browser console:
WebSocket connection to 'wss://…/live-ws' failed: The network connection was lost.

Server log (docker logs omniroute) — disconnects every 15s sweep, client dropped ~44s after connect:
[LiveWS] Client connected: 0101199b (127.0.0.1) [17 total]
[LiveWS] Client disconnected: 0101199b [4 remaining]
```

### Screenshots

_No response_

### Additional Context

- Add/extend a test asserting the client sends {type:"ping"} every ~15s (HEARTBEAT_INTERVAL_MS),
  or that a silent-but-connected client is not terminated.
- Existing tests/integration/live-ws-startup.test.ts and the handshake-route tests can be
  extended to cover keepalive/terminate timing.

### Validation Plan

_No response_

diegosouzapw:
Closes #10319

## Root cause

Confirmed in two files (matches the analyzed plan `_tasks/pipeline/bugs/2-implementing/10319-live-ws-client-never-sends-heartbeat-ping.plan.md`):

- **Client** (`src/hooks/useLiveDashboard.ts`): `ws.onopen` sent only `{ type: "subscribe", channels }`. No `setInterval`/keepalive timer existed anywhere in the hook, so the client never emitted the protocol's declared `{ type: "ping" }` heartbeat frame.
- **Server** (`src/server/ws/liveServer.ts`): `client.lastActivity` was refreshed only from inbound client messages. `startHeartbeat` terminates any client whose inbound gap exceeds `HEARTBEAT_TIMEOUT_MS` (35s, hardcoded). The server's own outbound `{ type: "pong" }` did **not** reset `lastActivity` (inbound-only), so it could not prevent the terminate.

Net effect: a healthy, connected-but-idle dashboard client (the normal case — the dashboard just watches, sends nothing) was force-terminated (`ws.terminate()`, abnormal close) roughly every 35-45s, and the browser auto-reconnected in a loop — exactly the reporter's symptom (`wss://…/live-ws failed: The network connection was lost` every ~35-45s).

## Fix (both directions, per the plan's recommendation)

1. **Client** — `useLiveDashboard.ts` now starts a 15s `setInterval` on open that sends `{ type: "ping" }`, cleared on every exit path (close, unmount, disable, reconnect) to avoid leaked timers. This hook is the shared base for `useLiveRequests`, `useLiveComboStatus`, `useLiveConnectionStatus`.
2. **Server (defense in depth)** — `startHeartbeat` now also bumps `client.lastActivity` when sending the outbound pong, so even a third-party/bespoke client that never pings is not dropped for being idle.

## Regression tests

- `tests/unit/useLiveDashboard-heartbeat.test.tsx` (vitest, fake timers): asserts the hook emits periodic `{ type: "ping" }` frames after open, and that the interval is cleared on close and on unmount (no leaked timers).
- `tests/integration/live-ws-heartbeat-keepalive.test.ts` (node:test, real WS server via `scripts/start-ws-server.mjs`): a subscribed-but-silent client must stay connected past the 35s `HEARTBEAT_TIMEOUT_MS` (~50s window) — the permanent version of the plan file's TDD RED repro, which failed before this fix (`AssertionError: true !== false`, connection was terminated at ~44s).

TDD RED→GREEN evidence is documented in the analyzed plan file (`verdict: fix-confirmed`, `confidence: CONFIRMED`). Pure protocol defect, not environment-specific — no VPS live validation required per the plan.

## Gates run locally

- `npm run typecheck:core` — GREEN
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json <changed files>` — GREEN, 0 errors
- `node scripts/check/check-file-size.mjs` — no new violations on touched files
- `node scripts/check/check-complexity.mjs` — GREEN (2456 vs baseline 2774)
- `node scripts/check/check-cognitive-complexity.mjs` — GREEN (1104 vs baseline 1223)
- `node scripts/check/check-test-discovery.mjs` — GREEN, new test files discovered by the runner

**Note:** `npm run test:unit` and the targeted `npx vitest run tests/unit/useLiveDashboard-heartbeat.test.tsx` could not be completed locally in this session — this devbox is running a 13-way parallel campaign fan-out and sustained a load average of ~150-240 on 16 cores for 30+ minutes; the local `test:unit` process measured 0.0% CPU over 26 minutes of wall time (confirmed via `ps`), and vitest's worker-pool spawn repeatedly hit its internal startup timeout under the same contention (`[vitest-pool-runner]: Timeout waiting for worker to respond` — an infra symptom, not a test assertion failure). Both new tests were manually verified line-by-line for correctness against the existing `tests/integration/live-ws-startup.test.ts` harness pattern and the hook's actual timer/cleanup logic. CI's `test-unit` and `test-vitest` jobs will run the authoritative suite on this PR.

⚠️ base-red inherited: #9985 — ESLint errors (2) from #10250 (i18n PT-PT translation), unrelated to this change.


diegosouzapw:
### Code review

Found 1 important issue.

The new `client.lastActivity = now` at `src/server/ws/liveServer.ts:435` runs for every socket whose local `readyState` is `OPEN`, immediately before the server sends its application-level `{ type: "pong" }`. A half-open or black-holed TCP peer can remain locally OPEN without receiving or acknowledging that message, so the timeout at lines 425-429 is never reached. Such stale entries can accumulate until `MAX_CLIENTS` is exhausted and healthy dashboard connections receive `SERVER_FULL`. Refresh liveness only on inbound client activity, or after an actual WebSocket-level pong acknowledgement, and add a stale-open-socket regression test. This conclusion is static review evidence; no runtime test was run on the PR head.

PR #10453, opened 2026-08-15
fix(ci): clean up Windows packaged smoke process trees

backryun:
## Summary

- terminate the Windows packaged Electron smoke process tree in one `taskkill /T /F` operation while the root PID is still valid
- prevent a crashed Electron parent from orphaning its embedded server and contaminating later smoke runs
- make the cleanup path dependency-injectable for a focused regression test

## Root cause

The Windows cleanup path first sent a signal only to the direct Electron process. If that parent exited before the liveness check, `stopApp()` considered cleanup complete and skipped the existing tree-kill path, leaving server descendants and port 20128 behind.

This PR fixes the cleanup defect only. It does not claim to resolve the separate Windows `EXCEPTION_BREAKPOINT` or `/login` readiness failure tracked in #10321.

## Validation

- TDD: the new regression initially failed because the Windows tree-safe cleanup contract did not exist
- `node --import .../tsx/dist/esm/index.mjs --test tests/unit/electron-smoke-script.test.ts tests/unit/electron-packaging.test.ts` — 6/6 pass
- `npx prettier --check scripts/dev/smoke-electron-packaged.mjs tests/unit/electron-smoke-script.test.ts` — pass
- `git diff --check` — pass

Refs #10321

### DIFF
--- a/scripts/dev/smoke-electron-packaged.mjs
+++ b/scripts/dev/smoke-electron-packaged.mjs
@@ -255,20 +255,33 @@ async function signalProcessTree(child, signal) {
   }
 }
 
-async function stopApp(child) {
+export async function stopApp(
+  child,
+  {
+    currentPlatform = platform(),
+    signalProcessTreeFn = signalProcessTree,
+    waitForProcessTreeExitFn = waitForProcessTreeExit,
+  } = {}
+) {
   if (!child.pid) return;
 
-  await signalProcessTree(child, "SIGTERM");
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks @backryun — killing the whole Windows process tree while the root PID is still valid is the right fix for the orphaned-server port contamination. The DI refactor makes it well-testable. Note the red CI is an inherited base-red (hung unit suite), not from this change. 🚀

PR #10454, opened 2026-08-15
fix(kilocode): downgrade json_schema response_format for DeepSeek V4 Flash (400 regression)

benzntech:
## Problem

`kilocode/deepseek/deepseek-v4-flash` rejects `response_format: {type: "json_schema"}` with HTTP 400 `Invalid input: response_format` (verified live 2026-08-15 via the Hindsight fact-extraction path — `upstream_details.param: response_format`). Any client using Structured Output against kilocode's DeepSeek fails.

The default executor's `applyJsonSchemaFallback` (which downgrades json_schema → json_object with the schema injected into the system prompt) is gated on `provider.startsWith("openai-compatible-")` only — so kilocode forwarded the unsupported format raw. Same bug class as the opencode fix in #9992, which covered the opencode/opencode-zen DeepSeek route but not kilocode's.

## Fix

Extend `applyJsonSchemaFallback`'s gate to include `kilocode`. json_schema is downgraded to json_object, with the JSON schema injected into the system prompt — identical to the existing openai-compatible-* behavior. Native-Structured-Output providers (openai etc.) keep the native json_schema path untouched.

## Regression test

`tests/unit/executor-default-base.test.ts`: new case asserts a kilocode request with json_schema response_format is downgraded to json_object + schema-injected system message, and the original body is not mutated. **Verified to fail without the fix** (sabotage run: 1 fail) and pass with it. All 48 executor-default-base tests pass.


benzntech:
Updated the fix after live verification: **kilocode's DeepSeek rejects BOTH json_schema AND json_object** response_format (both 400 with `param: response_format` — verified by direct gateway tests). So the downgrade-to-json_object approach (which fixes openai-compatible-*) is insufficient for kilocode.

**Refined fix** (force-pushed): for `kilocode`, **strip response_format entirely** and inject the schema (or a plain 'valid JSON only' instruction for json_object requests) into the system prompt. openai-compatible-* keeps the existing json_schema→json_object downgrade and json_object passthrough (they accept both natively).

Regression tests now cover: kilocode json_schema → stripped + schema-injected system message; kilocode json_object → stripped + JSON-only instruction. Sabotage-verified (2 fail without the fix). All 49 executor-default-base tests pass.

benzntech:
Superseded by #10458 — this branch accidentally carried the custom-fork history (100 commits). #10458 is the clean single-commit version of the same fix (2 files, +92/−11, sabotage-verified).

PR #10458, opened 2026-08-15
fix(kilocode): strip unsupported response_format for DeepSeek V4 Flash (400 regression)

benzntech:
## Problem

`kilocode/deepseek/deepseek-v4-flash` rejects **any** `response_format` — both `json_schema` and `json_object` return HTTP 400 `Invalid input: response_format` (verified live against the kilocode upstream). Any client using Structured Output against kilocode's DeepSeek fails.

The default executor's `applyJsonSchemaFallback` is gated to `openai-compatible-*` providers only and downgrades json_schema → json_object. That leaves kilocode out entirely — it forwards the unsupported format raw, and even the downgrade would not help (kilocode rejects json_object too). Same bug class as the opencode fix in #9992, which covered the opencode/opencode-zen DeepSeek route but not kilocode's.

## Fix

Extend `applyJsonSchemaFallback` to kilocode with **strip semantics**:

- **kilocode**: `response_format` is removed entirely; the JSON schema (or a plain `valid JSON only` instruction for json_object requests) is injected into the system prompt. The schema is preserved as a prompt, so callers still receive structured JSON.
- **openai-compatible-\***: unchanged — json_schema is downgraded to json_object (they accept it), json_object passes through untouched.
- **native providers** (openai etc.): untouched, keep native Structured Output.

## Regression tests

Two new cases in `tests/unit/executor-default-base.test.ts`:

1. kilocode json_schema request → response_format stripped + schema injected into a system message (original body not mutated).
2. kilocode json_object request → response_format stripped + JSON-only instruction injected.

**Sabotage-verified**: both fail without the fix (2 fails), pass with it. All 49 executor-default-base tests pass.

## Files

- `open-sse/executors/default.ts` (+47/−7)
- `tests/unit/executor-default-base.test.ts` (+56)

### DIFF
--- a/open-sse/executors/default.ts
+++ b/open-sse/executors/default.ts
@@ -605,24 +605,41 @@ export class DefaultExecutor extends BaseExecutor {
 
   /**
    * Downgrade `response_format: { type: "json_schema" }` to `json_object` for
-   * `openai-compatible-*` providers, injecting the JSON schema into the system
-   * prompt instead. DeepSeek / Ollama / local OpenAI-compatible models often
-   * lack native Structured Output and return empty or malformed content when a
-   * `json_schema` response_format is forwarded as-is. Gated on the
-   * `openai-compatible-` provider family so providers with native Structured
-   * Output support keep the native `json_schema` path.
+   * `openai-compatible-*` providers AND `kilocode`, injecting the JSON schema
+   * into the system prompt instead. DeepSeek / Ollama / local OpenAI-compatible
+   * models often lack native Structured Output and return empty or malformed
+   * content when a `json_schema` response_format is forwarded as-is (kilocode's
+   * DeepSeek V4 Flash rejects it with HTTP 400 `Invalid input: response_format`,
+   * verified live 2026-08-15 — same class as #9992's opencode fix). Gated so
+   * providers with native Structured Output support keep the native
+   * `json_schema` path.
    */
   applyJsonSchemaFallback<T>(body: T): T {
-    if (!this.provider?.startsWith?.("openai-compatible-")) return body;
+    const provider = this.provider ?? "";
+    const isOpenAiCompatible = provider.startsWith("openai-compatible-");
+    const isKiloCode = provider === "kilocode";
+    if (!isOpenAiCompatible && !isKiloCode) return body;
     if (!body || typeof body !== "object" || Array.isArray(body)) return body;
 
     const record = body as Record<string, unknown>;
     const rf = record.response_format as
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks @benzntech — the kilocode `response_format` strip is the right fix for the DeepSeek V4 Flash 400 (kilocode rejects both `json_schema` and `json_object`, so stripping is correct). However, this PR currently bundles 5 unrelated concerns (the kilocode fix, a `.gitignore` entry, dompurify/nanoid dependency bumps, a catalog ghost-model exclusion, and a modelsDevSync pricing memo). Please split the catalog exclusion and the modelsDevSync memo into their own PRs, and resolve the conflict against `release/v3.8.50` — that keeps each change reviewable and merge-safe on its own. The executor tests are good. 🙏

benzntech:
Addressed — thanks for the detailed review! 🙏

**Rebased onto `release/v3.8.50` and force-pushed.** The branch was originally based on `main`, which pulled in the unrelated files (gitignore, dompurify/nanoid bumps, catalog exclusion, modelsDevSync memo) as main-vs-release drift. Now the diff is **exactly 2 files**: `open-sse/executors/default.ts` + `tests/unit/executor-default-base.test.ts` (+92/−11). The other concerns are not part of this branch anymore — they were upstream commits, not mine, and have dropped out of the diff.

**Verified**: all 49 executor-default-base tests pass on the release/v3.8.50 base (including the 2 new kilocode strip regressions, sabotage-checked).

diegosouzapw:
### Babysit summary
- **Fixed**: `No new ESLint warnings` (lint-guard) → commit `9fb66ea` → the PR's 2 new kilocode strip tests added 3 net-new `as any` casts, pushing this file's frozen `@typescript-eslint/no-explicit-any` baseline (count 42 in `config/quality/eslint-suppressions.json`) to 45, tripping `npm run lint:json -- --max-warnings 0`. Replaced the `as any` result casts with typed `as unknown as { response_format?; messages: Array<{role;content}> }` assertions and the `(body as any)` check with the typed `body.response_format.type` — same assertions, same semantics, no test weakened or removed.
- **Tests**: the focused `tests/unit/executor-default-base.test.ts` suite passes 49/49, including the 2 new kilocode strip tests.
- **Gate**: all checks green — Build (advisory), No new ESLint warnings, Fast Quality Gates, Unit Tests ×4, Vitest, Docs Gates, Change Classification, Merge integrity, dast-smoke, semgrep, semgrep-cloud all `pass`. Mergify rows show `skipping` (neutral).
- Ready for human review & merge.

PR #10459, opened 2026-08-15
fix(services): fall back to ss and netstat when lsof is absent

### LINKED ISSUE #10431: fix(services): resolvePortPid() silently returns null when lsof is absent — adopted services lose pid tracking
## Summary

`resolvePortPid()` shells out to `lsof` and has no fallback. On any host without `lsof` — which includes slim container images — an adopted service permanently reports `pid: null`, silently.

`src/lib/services/portProbe.ts:116`:

```ts
export async function resolvePortPid(port: number): Promise<number | null> {
  return new Promise((resolve) => {
    const proc = spawn("lsof", ["-ti", `:${port}`]);
    ...
    proc.on("error", () => finish(null));
```

When `lsof` is absent the spawn fails with `ENOENT`, `on("error")` converts it to `null`, and the caller cannot distinguish "nothing holds this port" from "I have no way to look".

## Impact

`ServiceSupervisor` adopts an already-healthy instance instead of spawning a duplicate (`src/lib/services/ServiceSupervisor.ts:123`). The pid recorded for that adoption comes from `resolvePortPid()`. The existing test spells out why the pid matters:

> the adopt branch used to call `setToolStatus(tool, "running")` with no pid argument at all, leaving `pid` permanently null for any service adopted on a supervisor restart (common in production — e.g. after `systemctl --user restart omniroute.service`, sidecar child processes can outlive the restart and get adopted rather than spawned fresh). Downstream consumers that key liveness tracking off pid would then treat a genuinely healthy, running service as untrustworthy/stale.

That regression is fixed for hosts with `lsof` and silently reintroduced for hosts without it.

## Reproduction

On a container that has `ss`/`netstat` but no `lsof`:

```
$ which lsof ss netstat
/usr/bin/ss
/usr/bin/netstat

$ node --import tsx/esm --test tests/unit/services/ServiceSupervisor.test.ts
✖ #6205: probeBeforeSpawn adopts a healthy existing instance (no spawn)
  AssertionError: adopted instance still gets a tracked pid
✖ adopted service resolves and records the real pid of the process holding the port
  AssertionError: expected null to strictly equal 596922
ℹ pass 5
ℹ fail 2
```

Reproduced on a clean checkout of `release/v3.8.50` (`e05ac345da`), so this is not branch-specific.

## Why CI does not catch it

`tests/unit/services/ServiceSupervisor.test.ts` does not run in any of the four `Unit Tests fast-path` shards — grepping the logs of all four jobs of run `31851844623` returns zero occurrences. The GitHub runners also happen to ship `lsof`, so even if it ran there it would pass. The gap only shows on a host that lacks `lsof`.

## Suggested direction

Fall back to a second resolver when `lsof` is unavailable rather than collapsing to `null` — `ss -ltnp`/`netstat -ltnp` expose the same pid, and `/proc/net/tcp` + `/proc/*/fd` needs no external binary at all. Worth distinguishing "no process found" from "no resolver available" so callers can tell the two apart.

Either way this needs a test that runs on a host without `lsof` — otherwise the same silent hole reopens.

aron-intframe:
Closes #10431.

`resolvePortPid()` only shells out to `lsof`. No lsof means `spawn` raises ENOENT, `on("error")` turns it into `null`, and now the caller can't tell "nothing holds this port" from "I had no way to look" — so a service adopted after a supervisor restart keeps `pid: null` forever. That's the state the adopt-branch test at `ServiceSupervisor.test.ts:246` exists to prevent.

Now it tries `lsof`, then `ss`, then `netstat`, sharing one deadline so the whole lookup still costs at most `PID_RESOLVE_TIMEOUT_MS` instead of three times it.

The output parsing for each is a pure exported function (`parseLsofPid`, `parseSsPid`, `parseNetstatPid`) so the formats are testable on a box that doesn't have the binary. `netstat` has no port filter, so its parser matches the local-address column rather than scanning the line — otherwise a foreign address ending in the same number reads as a listener. There's a test for that row.

### Validation (Hard Rule #18, TDD)

This box has lsof, so I emulated the reported environment by pointing `PATH` at a dir holding only `ss` and `netstat`. Same script, same host, only `portProbe.ts` swapped:

```
before:  real=2027683  resolved=null       <- bug
after:   real=2027730  resolved=2027730
```

That's now a permanent test (`resolvePortPid still resolves a pid on a host without lsof`). It skips rather than passing vacuously if neither fallback is installed, and it asserts lsof is genuinely unreachable on the shimmed PATH first so it can't pass for the wrong reason.

```
node --import tsx/esm --test tests/unit/services/portProbePid.test.ts   →  9 pass, 0 fail
npx eslint src/lib/services/portProbe.ts tests/.../portProbePid.test.ts →  clean
```

Pre-commit ran clean as well (prettier, eslint with suppressions, `check-docs-sync`, `check:any-budget:t11`, `check:tracked-artifacts`).

### One pre-existing failure worth disclosing

`tests/unit/services/ServiceSupervisor.test.ts` is 6 passed / 1 failed here, on `adopted service resolves and records the real pid of the process holding the port` (`'error' !== 'running'`). It fails the same way on a pristine `release/v3.8.50` tree with my files removed, so it isn't mine — but it's the neighbouring test, so I'd rather say it than have it look like I didn't run it.

Environment: Node 22.14.0 (engines want >=22.19.0; `better-sqlite3` warns EBADENGINE on it). lsof, ss and netstat are all present here, which is why the fallback had to be exercised through a shimmed PATH rather than by uninstalling something.

### DIFF
--- a/src/lib/services/portProbe.ts
+++ b/src/lib/services/portProbe.ts
@@ -103,19 +103,86 @@ export async function probeBeforeSpawn(healthUrl: string, port: number): Promise
   return { healthy, portInUse };
 }
 
+/** `lsof -ti :PORT` prints one pid per line and nothing else. */
+export function parseLsofPid(stdout: string): number | null {
+  const firstLine = stdout
+    .split("\n")
+    .map((line) => line.trim())
+    .find((line) => line.length > 0);
+  const parsed = firstLine ? Number.parseInt(firstLine, 10) : Number.NaN;
+  return Number.isFinite(parsed) ? parsed : null;
+}
+
 /**
- * Resolve the pid of whatever process is listening on `port`, if any.
+ * `ss -tlnp 'sport = :PORT'` carries the pid inside the process column:
  *
- * Used when adopting an already-healthy instance (see `decidePreSpawn`'s
- * "adopt" outcome): the supervisor didn't spawn that process itself, so it
- * has no pid from a `ChildProcess` handle, but tracking a real pid is still
- * needed for downstream liveness checks to trust an adopted service the same
- * way they trust a freshly-spawned one. Returns null if nothing is found or
- * the lookup fails/times out (best-effort; never blocks adoption on this).
+ *   LISTEN 0 511 127.0.0.1:20128 0.0.0.0:* users:(("node",pid=596922,fd=18))
+ *
+ * The filter is applied by `ss` itself, so any `pid=` on any line belongs to
+ * the requested port.
  */
-export async function resolvePortPid(port: number): Promise<number | null> {
+export function parseSsPid(stdout: string): number | null {
+  const match = /\bpid=(\d+)/.exec(stdout);
+  const parsed = match ? Number.parseInt(match[1], 10) : Number.NaN;
+  return Number.isFinite(parsed) ? parsed : null;
+}
+
+/**
+ * `netstat -tlnp` cannot filter by port, so the port is matched here:
+ *
+ *   tcp 0 0 127.0.0.1:20128 0.0.0.0:* LISTEN 596922/node
+ *
+ * Matching on the local-address column keeps a foreign address that happens to
+ * end in the same number from being read as a listener.
+ */
+export function parseNetstatPid(stdout: string, port: number): number | null {
+  for (const line of stdout.split("\n")) {
+    const columns = line.trim().split(/\s+/);
+    // proto recv-q send-q local-address foreign-address state pid/program
+    if (columns.length < 7 || columns[5] !== "LISTEN") continue;
+    if (!columns[3].endsWith(`:${port}`)) continue;
+    const parsed = Number.parseInt(columns[6], 10);
+    if (Number.isFinite(parsed)) return parsed;
+  }
+  return null;
+}
+
+/**
+ * Ways to ask the OS which process holds a port, in preference order.
+ *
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks @aron-intframe — the lsof→ss→netstat probe chain fixes the silent `pid: null` on hosts without lsof, and the parser isolation with shared-deadline budget is clean. The no-lsof PATH-shim test is exactly the regression guard this needed. Note red CI is an inherited base-red, not this change. 🚀

aron-intframe:
Thanks for the read, and sorry for the slow reply.

One thing worth leaving here for whoever touches `resolvePortPid` next: the chain still returns `null` when *all three* probes are unavailable, so "nothing is listening on that port" and "this host cannot tell me" remain the same value. That is the distinction the original `pid: null` bug turned on, and the PATH shim only proves the lsof-missing case, where `ss` or `netstat` still answers. On minimal images — distroless, scratch-based sidecars — none of the three is present, so the branch is reachable in production, not just in the test. If it ever needs to be actionable, the honest shape is a third state at the call site rather than a `null` the caller has to guess at.

Your confirming the red CI as inherited was the useful half of the review for us, by the way. We had it in the PR body as a claim about someone else's tree; you turned it into a fact.

PR #10463, opened 2026-08-15
fix(combo): guarantee combo loops terminate instead of hanging silently

herjarsa:
## Summary

Guarantees that OmniRoute's combo/target loops always terminate with an actionable error instead of dying silently (no response, no log) when an upstream hangs.

Fixes 10 silent-stop gaps (G1–G10) found by auditing every loop that iterates models/combos to resolve a task:

**Critical — combo target loop (open-sse/services/combo.ts)**
- **G1**: globalPromise in handleComboChat had no timer — a hung target (per-model timeout disabled) froze the request forever. Added loopSafetyPromise (uses comboTimeoutMs, else a 10-minute hard ceiling COMBO_LOOP_SAFETY_TIMEOUT_MS) to all 3 Promise.race sites; on fire: marks comboExpired, aborts in-flight upstreams, drains in-flight tasks (COMBO_SAFETY_DRAIN_MS), returns **504** with the same aggregated 	ried: a (500) diagnostics as the regular timeout path.
- **G2**: task wrapper .catch() now resolves globalPromise with **502** so an unexpected task throw can never leave the post-loop race hanging.
- **G3**: 	argetTimeoutRunner logs a warning when comboTargetTimeoutMs <= 0 (per-model timeout disabled).
- **G4**: handleRoundRobinCombo had no global timeout and no safety net — added 
rSafetyPromise timer (same ceiling), 
rExpired loop guard, Promise.race around each model dispatch, try/catch → 500, and a post-loop 504.
- **G9**: round-robin retry delay now honors client disconnect (499).

**Localized silent exits**
- **G5** (utoCombo/chaosEngine.ts): dispatchOnePanelModel hardcoded ok: true — a panel where every model returned 503 was treated as success and the error streamed as a valid answer. Now honors 
es.ok; all-failed branch logs All chaos panel models failed with per-model errors.
- **G6** (utoCombo/pipelineRouter.ts): logs Reflection retries exhausted when the reflection loop burns its budget and still fails.
- **G7** (src/lib/evals/evalRunner.ts): rejects catastrophic regex patterns (ReDoS) via safe-regex before 	est().
- **G8** (utoRefreshDaemon.ts): alidateCredential swallowed network errors in a bare catch {} (fail-open silent) — now logs provider + error.
- **G10** (atchProcessor.ts): each batch-item dispatch bounded by a 120s wall-clock timeout (withItemDispatchTimeout) so a stuck item fails fast instead of freezing the batch.

## Verification

- New regression suite 	ests/unit/combo-silent-stop-gaps.test.ts — 10/10 pass (hung upstream → 504 in ~130ms; unexpected throw → 502; chaos all-fail → logged; ReDoS regex rejected; batch timeout).
- 	ests/unit/combo-routing-engine.test.ts + combo-target-timeout-runner.test.ts + pipeline-router.test.ts: 105/105 pass (102 existing + new).
- chaos-executor, atch-processor, service-batch-processor, valrunner-builtinsuites-split: 37/37 pass.
- 
pm run typecheck:core clean; ESLint clean on touched files (pre-existing violations in atchProcessor.ts untouched).

⚠️ **base-red inherited: #9985** — 
elease/v3.8.50 is currently red upstream; inherited failures are not introduced by this branch.


diegosouzapw:
Deferido para revalidação do head atual. A rodada anterior registrou um hold de ownership, mas a inspeção atual não confirmou esse worktree; por favor, reconfirme o head e os checks antes da decisão final.

PR #10468, opened 2026-08-15
fix(cli): restore packaged machine-token authentication

### LINKED ISSUE #10385: fix(cli auth): restore machine-token derivation, salt rotation, and packaged auth
## Summary

This issue is **not about `oma_live_...` access tokens**. Those are scoped remote-management credentials sent as `Authorization: Bearer ...`; the machine token discussed here is a local, loopback-only credential sent as `x-omniroute-cli-token`. A valid `oma_live_` token being accepted by `/api/cli/whoami` or another management endpoint does not falsify the defects below.

There are **three distinct CLI machine-token defects**, plus documentation drift:

1. **The packaged/current CLI does not derive a token on supported Node runtimes:** `await import("node-machine-id")` returns this CommonJS package under `default`, not as a named `machineIdSync` export. The call throws, the catch caches `""`, and `api.mjs` consequently sends no machine-token header.
2. **Salt rotation is broken even after correcting that import:** the CLI hard-codes the default legacy salt while the server honors `OMNIROUTE_CLI_SALT`.
3. **The published 3.8.49 server bundle does not accept same-runtime machine tokens:** Turbopack emits an external import for a non-existent hash-suffixed `node-machine-id` package name. Correct HMAC and legacy candidates generated inside the same container both receive 401.

The default-salt source algorithms are intentionally backward compatible: the server accepts the CLI's legacy 32-character format. This issue does not claim that the default source-level formulas disagree.

I searched existing issues and PRs for machine-ID token, CLI auth 401, Docker CLI auth, `x-omniroute-cli-token`, salt rotation, and the bundle external name. I found related CLI/server contract issues such as #10081, but no issue covering these defects.

## 1. CLI CommonJS interop defect

Verified from the published `omniroute@3.8.49` CLI and unchanged `release/v3.8.50` source.

`bin/cli/utils/cliToken.mjs` currently does:

```js
const { machineIdSync } = await import("node-machine-id");
const mid = machineIdSync();
```

The installed dependency is `node-machine-id@1.1.12`, a CommonJS package. Executing the published CLI module with the supported runtimes produced:

| Runtime | `typeof module.machineIdSync` | `typeof module.default.machineIdSync` | `getCliToken()` length |
|---|---:|---:|---:|
| Node 22.22.3 | `undefined` | `function` | `0` |
| Node 26.7.0 | `undefined` | `function` | `0` |

The catch block silently caches the empty string. `bin/cli/api.mjs` only adds `x-omniroute-cli-token` when the returned value is truthy, so the normal CLI request sends no machine token at all.

This needs an executable test under supported Node versions; the current shape test permits either 0 or 32 characters and therefore treats this failure as success.

## 2. Salt rotation mismatch

Verified on `release/v3.8.50` at `f06d5f20ed8e17389aa4a750e316b61cbe60ad4d` and in the published `omniroute@3.8.49` CLI.

`docs/security/CLI_TOKEN.md` documents the current format as HMAC-SHA256 with `OMNIROUTE_CLI_SALT`, and states that changing the variable causes all CLI processes on the machine to use the new token automatically.

The server implements that HMAC format and also accepts a legacy token through `getLegacyCliTokenSync()`.

However, `bin/cli/utils/cliToken.mjs` currently:

- attempts to call `machineIdSync()`;
- generates only the legacy 32-character SHA-256 prefix;
- hard-codes `omniroute-cli-auth-v1`;
- does not read `OMNIROUTE_CLI_SALT`.

`bin/cli/api.mjs` supports an explicitly precomputed `OMNIROUTE_CLI_TOKEN` override, but that is an out-of-band manual override rather than the documented automatic salt rotation.

Falsification with a fixed fake machine ID confirms:

| Scenario | Server accepts the CLI formula? |
|---|---:|
| Built-in default salt | Yes, through the intentional legacy compatibility path |
| Rotated `OMNIROUTE_CLI_SALT` | No |

This formula-level mismatch is independent of the CommonJS interop failure above: fixing the import alone still leaves custom-salt deployments broken.

## 3. Published 3.8.49 server packaging regression

Verified against the published Docker artifact:

```text
diegosouzapw/omniroute@sha256:92c768c56e2de32c51a0621ef182835018b00b288c9bb235c5c5e4514658c1a1
```

The npm package reports git head `e6523da2a12f0eaebc96ba95650efeec2436ab4d`.

In that artifact:

- the normal package `/app/node_modules/node-machine-id` exists and resolves;
- the built server chunk `/app/.build/next/server/chunks/[externals]_node-machine-id_0b5ig9s._.js` contains:

```js
require("node-machine-id-1f004c4bfec225c1")
```

- that hash-suffixed module does not exist;
- the bundled machine-token initialization catches/falls back around this external load;
- most importantly, requests made **inside the same container** to a loopback management endpoint with either candidate derived directly from the real installed module — the full HMAC token and the legacy 32-character token — both return HTTP 401.

The executable 401 result is the claim here; this issue does not rely on an uninstrumented assertion about the exact internal fallback value after the external-load failure.

Host/container machine-ID mismatch cannot explain an in-container loopback failure. This packaging result has been verified on 3.8.49 only. Version 3.8.50 is present as a release branch but is not currently published to npm, so this issue does not claim the bundled regression has already been reproduced in a 3.8.50 package.

No token, machine ID, password, custom salt, or private endpoint is included here.

## 4. Docker boundary and diagnostics

When the CLI runs on the host and OmniRoute runs in Docker, different machine IDs are expected. This issue is **not** asking the server to trust a host-derived machine token across a container boundary.

The supported `omniroute connect <host>` / `oma_live_` access-token flow is the appropriate cross-boundary credential. The CLI should diagnose the boundary and recommend that flow instead of leaving the user with an undifferentiated 401.

## 5. Documentation and generated skill drift

- `CLI_TOKEN.md` describes loopback in terms of the Host header, while the current policy deliberately refuses to trust Host and uses a token-stamped real peer address.
- `skills/omni-auth/SKILL.md` shows `POST /api/auth/login` with an existing Bearer token and an empty body. The route requires `{ password: string }` and returns the Dashboard session cookie; the documented empty body produces HTTP 400.

## Proposed changes

1. Fix CLI loading of `node-machine-id` for its actual CommonJS export shape and fail diagnostically rather than silently caching `""`.
2. Fix the production bundle external so the real `node-machine-id` resolves in packaged/container builds.
3. Add a production-bundle integration test that proves a same-runtime machine token is accepted by a management GET.
4. Make the CLI honor `OMNIROUTE_CLI_SALT` and the documented HMAC format, while retaining deliberate legacy compatibility.
5. Add auth diagnostics to `omniroute doctor --json` without printing the derived token:
   - CLI machine-ID module/import available;
   - server machine-ID module available;
   - server reachable;
   - machine-token accepted/rejected/disabled;
   - container boundary likely;
   - recommended next action.
6. Correct the loopback description and generated `omni-auth` login/session examples.
7. Never silently read `JWT_SECRET` or mint a Dashboard JWT as fallback.

## Acceptance criteria

- [ ] The CLI derives a non-empty token under every supported Node major, including Node 22 and Node 26.
- [ ] Packaged/container server builds resolve the real `node-machine-id` module.
- [ ] A production-bundle test proves same-runtime machine-token authentication works.
- [ ] Custom `OMNIROUTE_CLI_SALT` works as documented.
- [ ] Legacy tokens remain accepted for backward compatibility.
- [ ] Docker machine-ID mismatch produces actionable `connect` guidance rather than weakening locality checks.
- [ ] `doctor --json` reports auth status without exposing credentials.
- [ ] Generated `omni-auth` examples match the actual password/session/CSRF flow.
- [ ] Tests cover CLI CommonJS interop, packaged resolution, default salt, rotated salt, legacy compatibility, unavailable machine ID, disabled CLI token, and container-boundary guidance.

I can split a fix into a packaging/auth PR and a separate generated-skill/documentation PR if that is easier to review.

xiaoyaner0201:
## Summary

- restore packaged/current CLI machine-token derivation using the runtime-safe CommonJS interop shape and the configured `OMNIROUTE_CLI_SALT`
- keep the machine token local-only while carrying the central authz decision to route-level management auth without forwarding the raw token
- harden `omniroute doctor` so it validates an HTTP(S), credential-free repository-defined loopback target before deriving or sending the machine token, and applies `redirect: "error"` on the real fetch path
- keep generated generic API skills independently usable with Bearer authentication while limiting cookie/CSRF session examples to `omni-auth`
- materialize `sql.js` and `node-machine-id` into the installed standalone runtime during postinstall
- add machine-token diagnostics, an explicit disable-path check, and a real installed-tarball smoke with no-token / invalid-token / valid-token contrast controls plus restart persistence

## Related Issues

- Closes #10385

## Maintainer Feedback Repair

Both findings in [the maintainer review comment](https://github.com/diegosouzapw/OmniRoute/pull/10468#issuecomment-5312282696) are addressed by commit `2b785f0068a862fbd867221294325ad921787782` (tree `284c0dd470f5d41deb2652d27a36544c9b1c35be`):

1. `checkMachineTokenAuth()` now parses and validates the target before calling `getCliToken()`. It accepts only HTTP(S) URLs without userinfo that pass the repository's loopback check, rewrites the path to `/api/cli/whoami`, and sends the token with `redirect: "error"`. Rejected, malformed, credential-bearing, and remote targets return without deriving a credential or calling `fetch`.
2. The generator now scopes dashboard cookie/CSRF flow to `omni-auth`. Generic generated API skills, including `omni-providers` and `omni-settings`, use standalone `Authorization: Bearer $OMNIROUTE_TOKEN` examples for both reads and mutations. Generated mutation commands also preserve the shell continuations required for their content-type and body arguments.

## Validation

- [x] Change type: CLI / build-deploy / auth
- [x] Focused ESLint and syntax checks on touched production/test files
- [x] Reconciled with the current `release/v3.8.50` tip (`ee221d870`); focused checks rerun afterward
- [x] Clean `npm run build:cli`
- [x] Real `npm run check:pack-boot`: packaged dependency contract, health, `401 / 401 / 200` machine-token controls, forced `sql.js` PATCH/GET, clean shutdown, restart, disk persistence
- [x] Pack artifact policy passed as an explicitly marked canary (`OMNIROUTE_ALLOW_CANARY_BUILD=1`); normal provenance intentionally rejects an unmerged PR commit
- [x] CLI token derivation exercised on Node `22.22.3`, `24.18.0`, and `26.5.0`
- [x] Secret scan: no 64-character token appears in build or pack-boot logs
- [x] Maintainer-feedback regression files on the accepted tree: doctor `9/9`, generator `22/22`
- [x] Related exact-tree test surface: `151/151` passed, with zero failed, skipped, or cancelled tests
- [x] Generator sync: `0` generated, `46` unchanged, `0` orphans, `0` errors
- [x] Independent exact-tree review: zero unresolved P0/P1/P2 findings

Additional executable evidence on the accepted tree:

- adversarial machine-token matrix: 31 hostile URL shapes rejected with zero fetch/token transmissions; 8 supported loopback shapes accepted
- a real loopback redirector/sink probe confirmed the redirector was reached once while the cross-origin sink received zero requests and zero tokens
- all 14 changed generated skills and 219 generator-owned bash blocks were validated; generic examples had no cookie/CSRF dependency, and all 119 generated mutation blocks matched the 119 continuation repairs
- `git diff --check` passed and the worktree remained clean

The full unit suite, Vitest, coverage, repository-wide lint, and production build remain CI-owned merge gates; this update does not claim those unrun full suites passed. The target branch currently has an inherited base-red report: #9985.

## Tests Added Or Updated

- `tests/unit/agentSkills-generator.test.ts`
- `tests/unit/check-pack-boot.test.ts`
- `tests/unit/cli-doctor-command.test.ts`
- `tests/unit/cli-machine-token.test.ts`
- `tests/unit/lib/managementCliToken.test.ts`
- `tests/unit/next-config.test.ts`

## Coverage Notes

- token derivation and error fallback are covered at the CLI/server utility level
- localhost, remote, deceptive-host, credential-bearing, wrong-token, legacy-token, disable-switch, redirect, and central-authz-to-route handoff cases are covered in management auth and doctor tests
- generated generic GET and mutation examples are covered for standalone Bearer auth, while `omni-auth` logout retains cookie/CSRF coverage
- the installed-tarball smoke is a semantic artifact test rather than a string-only test: it installs the generated package, boots it, proves negative controls fail, proves the packaged token succeeds, writes through forced `sql.js`, restarts, and reads the persisted value back

## Reviewer Notes

- the local machine token remains loopback-only; the central pipeline strips the raw token before forwarding the request and route-level auth trusts only the internal subject stamp
- token derivation is deliberately ordered after target validation, and redirects are rejected by the fetch that carries the token
- `OMNIROUTE_DISABLE_CLI_TOKEN=true` is enforced in both central policy and CLI diagnostics
- generic generated API skills remain Bearer-based and do not assume `cookie.jar` or CSRF setup; cookie/CSRF remains scoped to `omni-auth`
- no schema migration or persistent-data change
- `node-machine-id` and `sql.js` are copied from npm-installed dependency resolution into `dist/node_modules` during postinstall; they are not fetched from the network by this change
- this PR remains Draft pending live CI and maintainer review


xiaoyaner0201:
Pushed a final adversarial-review hardening pass in `fed70bb1756caad83ca11ce5b394c5bd3f76851a`.

The review found and this revision closes two additional gaps:

- Machine-token derivation now fails closed when `node-machine-id` silently resolves to an empty value. Both current HMAC and legacy derivation return no credential rather than a public constant, and the installed-package smoke explicitly rejects the public empty-machine-id HMAC in addition to checking shape and HTTP contrast controls.
- Settings audit attribution now consumes the trusted central-auth subject stamp, so successful CLI-authenticated mutations remain attributed to `cli` after the raw `x-omniroute-cli-token` header is stripped.

Also made the salt-mutating unit test restore the pre-existing environment value.

Validation on the exact revised tree:

- focused + related tests: 123/123 passed
- typecheck:core: passed
- changed-file lint / syntax / Prettier / diff checks: passed (the settings route still has two pre-existing restricted-import findings, reproduced unchanged on the pre-delta file)
- clean CLI build: passed
- installed tarball pack/boot: no-credential 401, invalid-token 401, packaged-token 200, forced sql.js write/read, clean restart, persistence passed
- independent fresh Opus delta review: `FINAL PASS`; only optional LOW hardening notes remain

The PR remains Draft and the refreshed CI is now running against this exact head. No runtime instance was restarted or modified.

diegosouzapw:
I found two important issues to address before merge. First, checkMachineTokenAuth sends the derived x-omniroute-cli-token to whatever livenessUrl/host is configured and uses fetch's default redirect behavior. A remote doctor target, or a localhost endpoint that redirects cross-origin, can therefore receive this machine-bound credential. Restrict this probe to loopback and use redirect:'error', with regression tests. Second, the generator's new cookie/CSRF branch applies to every API skill, not only omni-auth: generated omni-providers/omni-settings GETs no longer include the documented Bearer credential, and mutations assume cookie.jar/CSRF state that those skills never establish. Preserve Bearer examples for generic API skills or scope the session flow to omni-auth. The PR head was not locally available, so these are static-diff findings and are marked PLAUSIBLE.

xiaoyaner0201:
Thanks — both findings were confirmed and repaired in `2b785f0068a862fbd867221294325ad921787782` (tree `284c0dd470f5d41deb2652d27a36544c9b1c35be`).

1. `checkMachineTokenAuth()` now validates the destination before deriving the machine token. The target must be HTTP(S), contain no userinfo, and pass the repository's loopback check. Only then is `/api/cli/whoami` constructed and the token sent, with `redirect: "error"` applied to the actual fetch. Rejected remote, deceptive-host, credential-bearing, malformed, and unsupported targets perform zero fetches and derive/send no credential.
2. Cookie/CSRF generation is now scoped to `omni-auth`. Generic API skills such as `omni-providers` and `omni-settings` again use standalone Bearer authentication for reads and mutations, with corrected shell continuations so content-type and body arguments remain part of each generated command.

Verification on that exact commit/tree:

- doctor regression file: 9/9 passed
- generator regression file: 22/22 passed
- related test surface: 151/151 passed, with zero failures, skips, or cancellations
- generator sync: 0 generated, 46 unchanged, 0 errors
- adversarial URL matrix: 31 hostile forms rejected with zero fetch/token transmissions; 8 loopback forms accepted
- real loopback redirector/sink probe: redirector reached once, cross-origin sink reached zero times and received no token
- 14 changed generated skills / 219 generated bash blocks validated; all 119 mutation blocks matched the 119 continuation repairs
- independent exact-tree review: zero unresolved P0/P1/P2

I also updated the PR body with the repair and validation evidence. The PR remains Draft; this is not a Ready-for-review or merge claim, and the full CI-owned suites are not being reported as locally passed.

Issue #9134, opened 2026-08-01
fix(providers): C:/Program Files/Git/v1/audio/transcriptions rejects combo names that /v1/models advertises (chat and embeddings resolve them)


zhiru:
## Summary

`POST /v1/audio/transcriptions` does not resolve combos — it accepts only literal
`provider/model` ids. But `GET /v1/models` lists combos alongside models, so any
client that populates a model picker from `/v1/models` offers an option the endpoint
then refuses.

`/v1/chat/completions` and `/v1/embeddings` both resolve combos, so this is an
inconsistency between endpoints rather than a deliberate limitation.

## Evidence

Same API key, same gateway, same audio file, same instance:

| `model` sent | Result |
| --- | --- |
| `transcricao` (a combo, and it appears in `/v1/models`) | `400 Invalid transcription model: transcricao. Use format: provider/model` |
| `<node-id>/<model-id>` (the combo's only target) | `200` + correct transcription |
| `transcricao` on `/v1/chat/completions` | combo **is** expanded — the error names the resolved target `<node-id>/<model-id>` |

The chat control is the decisive one: the same name resolves on chat and does not on
audio.

## Requests, in order of preference

1. **Resolve combos in `/v1/audio/transcriptions`**, as chat already does. This is
   the root fix — a combo exists precisely so the caller does not depend on a
   provider's internal model id.
2. If not: **do not list combos in `/v1/models`** when they are not valid for every
   endpoint, or tag each entry with the endpoints it supports
   (e.g. `"endpoints": ["chat"]`) so clients can filter.
3. The error message could name the resolved id
   (e.g. *"combo 'transcricao' targets `<provider>/<model>`; use that id on this
   endpoint"*), which would solve the practical problem even without 1 or 2.

## Two smaller findings

**Bare alias collides with a hardcoded provider model.** An alias named `whisper`
routes to `pollinations`, which answers
`402 Insufficient balance… available balance 0.0000`. `parseAudioModel()`'s Phase-2
bare-id lookup matches a model literally named `whisper` in the hardcoded registry,
so the request leaves for a provider the operator never chose — and the failure is a
billing error from that third party. A prefix-less, official-looking name is the
natural first choice when configuring, which makes this an easy trap.

**A listed id is rejected by the endpoint.** The fully-qualified
`<node-id>/<model-id>` form appears in `/v1/models` and is also refused with
`400 … Use format: provider/model` — the catalog advertises an entry the endpoint
considers invalid.

[...]

PR #10471, opened 2026-08-15
fix(providers): resolve combo names on /v1/audio/speech and /v1/videos/generations

### LINKED ISSUE #10469: fix(providers): /v1/audio/speech and /v1/videos/generations reject combo names that /v1/models advertises
`GET /v1/models` advertises combos (`owned_by: combo`), and most routes resolve those names. Two media routes still do not: they reject a combo name at model validation before any resolution happens.

This is the same class of bug as #9134 (audio transcriptions) and #8986 (image-model validation), on the two routes those fixes did not cover.

## Reproduce

Create a combo (`strategy: fill-first`, any nodes of the matching modality), then send its name as `model`:

```
POST /v1/audio/speech        {"model":"rh-tts","input":"hi","voice":"alloy"}
→ 400 {"error":{"message":"Invalid speech model: rh-tts. Use format: provider/model",
                "type":"invalid_request_error","code":"bad_request"}}

POST /v1/videos/generations  {"model":"rh-video","prompt":"a red cube"}
→ 400 {"error":{"message":"Invalid video model: rh-video. Use format: provider/model",
                "type":"invalid_request_error","code":"bad_request"}}
```

Both combos are returned by `GET /v1/models`:

```json
{"id":"rh-tts","object":"model","owned_by":"combo", ...}
{"id":"rh-video","object":"model","owned_by":"combo", ...}
```

## Expected

The combo name resolves to its first node, as `/v1/images/generations` already does. Same request against the images route, for contrast — the name is accepted and expanded:

```
POST /v1/images/generations  {"model":"rh-image"}
→ 400 "Prompt is required for image model: kie/qwen/text-to-image"   ← combo resolved to node 1

POST /v1/images/generations  {"model":"__nonexistent__"}
→ 400 "Invalid image model: __nonexistent__. Use format: provider/model"
```

## Current state per route

| Route | Combo name |
|---|---|
| `/v1/chat/completions` | resolved |
| `/v1/embeddings` | resolved |
| `/v1/audio/transcriptions` | resolved (fixed in #9134) |
| `/v1/images/generations` | resolved (fixed in #8986) |
| `/v1/audio/speech` | **rejected** |
| `/v1/videos/generations` | **rejected** |

## Why it matters

A client that picks a model from `/v1/models` cannot tell which entries are usable on which route — the catalogue advertises names two endpoints refuse. Callers end up hardcoding vendor ids for speech and video while using combo names everywhere else, which defeats the point of server-side routing.

## Environment

- OmniRoute 3.8.48 (self-hosted); `/openapi.yaml` reports `info.version: 3.8.35`
- Combos created via `POST /api/combos` with `strategy: fill-first`

Related: #9134, #8986, #9239.

sha367:
Fixes #10469.

`GET /v1/models` advertises combos with `owned_by: combo`, and chat, embeddings, transcriptions (#9134) and images (#8986, #9239) all resolve those names. Speech and video did not — both rejected a combo name at model validation, before any resolution could happen:

```
POST /v1/audio/speech        {"model":"my-combo","input":"hi"}
→ 400 Invalid speech model: my-combo. Use format: provider/model

POST /v1/videos/generations  {"model":"my-combo","prompt":"a cube"}
→ 400 Invalid video model: my-combo. Use format: provider/model
```

A client picking a model out of `/v1/models` could not tell which entries the catalogue would actually accept, so callers ended up hardcoding vendor ids for these two routes while using combo names everywhere else.

## What changed

Both routes now mirror `images/generations`: detect a combo name before the provider lookup and divert to a strategy executor. The two new executors follow `imageCombo` — expand targets with `resolveComboTargets()`, filter to targets the route can serve, walk them in priority order, return the first success or the last failure, with 400/401/403 terminal and everything else advancing.

| File | |
|---|---|
| `open-sse/services/speechCombo.ts` | new executor |
| `open-sse/services/videoCombo.ts` | new executor |
| `src/app/api/v1/audio/speech/route.ts` | +13 — combo detection |
| `src/app/api/v1/videos/generations/route.ts` | +13 — combo detection |
| `src/app/api/v1/_shared/mediaGenerationRoute.ts` | +9 — optional meta fields |

## Two deliberate differences from imageCombo

**Speech filters at model level, not provider level.** `parseSpeechModel()` resolves a provider prefix without checking that the model behind it can speak, so `openai/gpt-4o` would be accepted as a target and fail only once dispatched. The filter checks the provider's own `models` list, and keeps targets from dynamic provider nodes that do not enumerate models. This was caught by the test below, which failed against a provider-level filter.

**Speech returns the handler's `Response` untouched** rather than building a JSON body, because that route streams audio. Only the ADD-only meta headers are attached, exactly as the direct path does; the body is read on the failure branch only.

`successfulMediaGenerationResponse()` gains optional `strategy` and `fallbackAttempts` so the video strategy can report them the way `imageCombo` does instead of duplicating the cost calculation. Both are omitted on the direct single-model path, where neither is meaningful.

## Tests

`tests/unit/combo/{speech,video}-combo.test.ts` mirror `image-combo.test.ts`: combo not found, no capable targets, empty combo, and targets present with no provider connection.

```
node --test tests/unit/combo/{image,video,speech}-combo.test.ts
# tests 16
# pass 16
# fail 0
```

Verified against a self-hosted 3.8.48 instance where the two routes reject combo names while `/v1/images/generations` resolves them.

### DIFF
--- a/open-sse/services/speechCombo.ts
+++ b/open-sse/services/speechCombo.ts
@@ -1,0 +1,182 @@
+/**
+ * Speech Combo Strategy Execution
+ *
+ * Mirrors imageCombo for /v1/audio/speech: expands combo targets via
+ * resolveComboTargets(), filters to speech-capable targets, runs each through
+ * handleAudioSpeech() in priority order, and returns the first success or the
+ * last failure.
+ *
+ * Unlike the image and video strategies, the speech handler returns a Response
+ * carrying an audio stream rather than a JSON result object, so success is read
+ * off `response.ok` and the upstream body is passed through untouched — only
+ * ADD-only meta headers are attached, matching the direct route.
+ */
+import { getComboByName, getCombos } from "@/lib/db/combos";
+import { resolveComboTargets } from "@omniroute/open-sse/services/combo.ts";
+import { parseSpeechModel, getSpeechProvider } from "@omniroute/open-sse/config/audioRegistry.ts";
+import { resolveDynamicAudioProviders } from "@/app/api/v1/_shared/audioProviderNodes";
+import {
+  getProviderCredentialsWithQuotaPreflight,
+  clearRecoveredProviderState,
+} from "@/sse/services/auth";
+import { isAllRateLimitedCredentials } from "@/app/api/v1/_shared/rateLimit";
+import { handleAudioSpeech } from "@omniroute/open-sse/handlers/audioSpeech.ts";
+import { attachOmniRouteMetaToResponse } from "@/domain/omnirouteResponseMeta";
+import { generateRequestId } from "@/shared/utils/requestId";
+import { calculateModalCost } from "@/lib/usage/costCalculator";
+import { getClientIpFromRequest } from "@/lib/ipUtils";
+import { toJsonErrorPayload } from "@/shared/utils/upstreamError";
+import { HTTP_STATUS } from "@omniroute/open-sse/config/constants.ts";
+import { errorResponse } from "@omniroute/open-sse/utils/error.ts";
+
+/**
+ * Execute a full combo strategy for a text-to-speech request.
+ */
+export async function executeSpeechCombo(
+  comboName: string,
+  body: Record<string, unknown>,
+  auth: {
+    request: Request;
+    policy: { apiKeyInfo?: { id?: string; name?: string } | null };
+  },
+  startTime: number
+): Promise<Response> {
+  const combo = await getComboByName(comboName);
+  if (!combo) {
+    return errorResponse(HTTP_STATUS.BAD_REQUEST, `Combo not found: ${comboName}`);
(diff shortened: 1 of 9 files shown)


diegosouzapw:
The combo-name diversion is directionally correct and matches the issue, but I found three behavior gaps in the new video path. First, videoCombo skips the direct route's local override credential lookup, so configured ComfyUI connection base URLs are not forwarded. Second, its capability filter only consults the static video registry, while the direct route supports custom models with supportedEndpoints including videos. Third, videos/generations applies the prompt-required check before combo expansion, so a combo targeting a prompt-optional I2V model can still be rejected before resolution. Please fix these and add route-level regression coverage; the current tests exercise executors directly, not the POST diversion.

Issue #10293, opened 2026-08-13
fix(startup): preserve Windows Tailscale Funnel code paths in published builds


jonlwheat2-gif:
# Tailscale Funnel broken on Windows — Windows branches missing from the published build (v3.8.49)

## Summary

On Windows, the Tailscale tunnel manager is broken in the published v3.8.49 `dist`. The compiled bundle is missing every Windows-specific branch that exists in `src/lib/tailscaleTunnel.ts`, so the dashboard reports **"Not installed"** and **Install & Enable** fails with a persisted `tailscaled did not become ready` error — even when Tailscale is installed, running, and logged in.

## Environment

- OS: Windows (win32)
- omniroute v3.8.49 (latest on npm as of 2026-08-13; published 2026-07-30)
- Tailscale installed at `C:\Program Files\Tailscale\` (client v1.102.2); daemon running and logged in — `tailscale status --json` returns `"BackendState": "Running"`

## Symptoms

- Dashboard → **Tailscale Funnel** shows `Not installed / Install & Enable` although Tailscale is installed and the daemon is up.
- Live `GET /api/tunnels/tailscale/check` reports `loggedIn: false`, `daemonRunning: false`, `tunnelUrl: null` while the CLI itself works fine from the same machine.
- Enabling fails and persists `lastError: "tailscaled did not become ready"`; `~/.omniroute/tailscale/tailscaled.log` shows:
  `safesocket.Listen: namedpipe.Listen: open C:\Users\...\.omniroute\tailscale\tailscaled.sock: Incorrect function.`

## Root cause

The `dist` bundle was produced on a non-Windows machine. The bundler (Turbopack) constant-folded `process.platform === "win32"` to `false` and dead-code-eliminated every Windows branch from the compiled output. The shipped `src/lib/tailscaleTunnel.ts` contains the correct branches; the compiled chunk (`dist/.build/next/server/chunks/src_0x3a1q5._.js`) does not.

Four concrete regressions, all in that one chunk:

### 1. `resolvePathCommand` always runs `which` — never `where`

`src/lib/tailscaleTunnel.ts` (line 215):
```ts
const lookupCommand = process.platform === "win32" ? "where" : "which";
```

compiled:
```js
async function F(t){try{let{stdout:a}=await td("which",[t],{timeout:3e3,windowsHide:!0,env:{...process.env,PATH:tf}});return a.split(/\r?\n/).map(t=>t.trim()).find(Boolean)||null}catch{return null}}
[...]

PR #10474, opened 2026-08-15
fix(cliproxy): read os.platform()/os.arch() at runtime in binaryManager platform detection (#10244)

### LINKED ISSUE #10244: fix(services): cliproxy installer may download Linux build on Windows (verify platform detection in binaryManager.ts)
## Summary

On **Windows**, the embedded-services installer for CLIProxyAPI always downloads and installs the **Linux** build, so the service can never start. Verified on OmniRoute 3.8.49 (Windows 10, Node v24.18.0).

## Root cause

In the bundled cliproxy installer, the platform/arch selection uses the **comma operator**, which discards `process.platform`/`process.arch` and always evaluates to `"linux"`/`"amd64"`:

```js
let { platform: n, arch: a } = {
  platform: (process.platform, "linux"),   // always "linux"
  arch: (process.arch, "amd64"),          // always "amd64"
};
const i = `CLIProxyAPI_${s.version}_${n}_${a}${"windows" === n ? ".zip" : ".tar.gz"}`;
```

Result: on every OS the download is `CLIProxyAPI_<version>_linux_amd64.tar.gz`, and the Windows branch (`Expand-Archive` via PowerShell) is dead code.

## Impact

- `~/.omniroute/bin/cliproxyapi-<version>/` contains the Linux tarball; the extracted `cli-proxy-api` is an ELF binary (`ELF 64-bit ... for GNU/Linux`) that cannot run on Windows.
- The service fails to start: the supervisor reports `Health probe did not succeed within 15000ms` and `Fast crash (exited with code 1)`, and nothing listens on the configured port (8317).
- Windows builds **are** published upstream (`CLIProxyAPI_7.2.131_windows_amd64.zip` exists on the `router-for-me/CLIProxyAPI` release), so the fix is trivial.

## Repro

1. Install OmniRoute on Windows (`npm i -g omniroute`).
2. Trigger install/start of the CLIProxyAPI embedded service from the dashboard.
3. Inspect `~/.omniroute/bin/`:
   - downloaded asset is `CLIProxyAPI_<v>_linux_amd64.tar.gz` on a Windows host;
   - `file ~/.omniroute/bin/cliproxyapi-<v>/cli-proxy-api` → `ELF 64-bit ... for GNU/Linux`.

## Workaround (until fixed)

Download `CLIProxyAPI_<version>_windows_amd64.zip` manually, extract it to `~/.omniroute/bin/cliproxyapi-<version>/`, and set `auto_update=0` + `pinned_version=<version>` in `version_manager` so the buggy updater does not re-download the Linux build.

## Suggested fix

```js
const platform = process.platform === "win32" ? "windows" : process.platform;
const arch = process.arch === "x64" ? "amd64" : process.arch;
const ext = platform === "windows" ? ".zip" : ".tar.gz";
const asset = `CLIProxyAPI_${version}_${platform}_${arch}${ext}`;
```

diegosouzapw:
Closes #10244

## Root cause
`#10244` reports that the embedded CLIProxyAPI installer always downloads the **Linux** build on Windows, quoting `platform: (process.platform, "linux")` from the bundled installer. The **source is correct** — `src/lib/versionManager/binaryManager.ts` already maps `win32→windows`, `x64→amd64`, and selects the `.zip` + PowerShell `Expand-Archive` on Windows. The damage is in the **build artifact**, not the source: Turbopack `next build` runs only on Linux (`npm-publish.yml`) and constant-folds `process.platform`/`process.arch` to the build machine's values, dead-code-eliminating every Windows/ARM branch. `detectPlatform()` is an `if (p === "linux") return "linux"` chain, so the fold collapses it to exactly that comma-operator form. This is the second confirmed victim of the systematic build defect tracked in `#10293`.

## Fix
Switch `detectPlatform()`/`detectArch()` to runtime `os.platform()`/`os.arch()` reads (the repo's established anti-fold pattern in `cliRuntime.ts`, `containerProvider.ts`, several routes). `os.platform()` is a function call a bundler cannot fold, so the Windows/ARM branches survive any build machine. Behavior is identical on every real OS.

## Regression test (fail → pass)
Added `tests/unit/binaryManager.test.ts` "should read platform/arch at runtime from os (anti build-folding guard) (#10244)": mocks `os.platform()`→`win32` and `os.arch()`→`arm64` and asserts `getTargetPlatform()` returns `{platform:"windows", arch:"arm64"}` and the asset name is the Windows `_windows_arm64.zip`.
- **RED before fix**: unfixed source returned `{platform:"linux", arch:"amd64"}` even with `os` mocked to win32/arm64 (it read the foldable `process.platform`/`process.arch`).
- **GREEN after fix**: returns `{platform:"windows", arch:"arm64"}`.
All 17 binaryManager tests pass; 36 focused touched-area tests across 6 files pass.

## Gates run green
- `node scripts/check/check-file-size.mjs` → OK
- `node scripts/check/check-complexity.mjs` → OK (2465 vs baseline 2774)
- `node scripts/check/check-cognitive-complexity.mjs` → OK (1104 vs baseline 1223)
- `npm run typecheck:core` → exit 0
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json` on both changed files → exit 0
- Full `test:unit` could not complete locally due to extreme devbox CPU saturation (load avg 60–80 from ~15 concurrent whole-tree test/tsc/eslint runs in parallel sessions); all changed-area tests pass locally and CI covers the full suites.

Plan file: `_tasks/pipeline/bugs/2-implementing/10244-cliproxy-installer-windows-platform-detection.plan.md`

⚠️ base-red inherited: #9985


diegosouzapw:
[IMPORTANT] The patch fixes runtime asset-name detection but leaves the installer’s other platform decisions on process.platform. At binaryManager.ts:85, extractZip still calls buildExtractZipCommand(process.platform,...); lines 168 and 222 likewise select symlink/copy behavior from process.platform. In a Linux-built Turbopack bundle, the Windows host can therefore select the new Windows ZIP and then still invoke unzip (and take the symlink path) instead of PowerShell/copy. The added test only mocks getTargetPlatform and getAssetName, so it does not cover this execution path. Please route extraction/install/rollback through the same non-foldable runtime platform source and add a regression covering the full Windows path.

Issue #10443, opened 2026-08-15
[Bug] antigravity via OmniRoute unusable — stream drops mid-generation, connection refused on restart


trancexe:
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04.4 LTS (Kernel 7.0.0-28-generic x86_64

### Node.js Version

26.5.0

### Provider(s) Involved

Antigravity

### Model(s) Involved

antigravity/gemini-3.6-flash-medium

### Client Tool

Hermes Agent CLI

### Description

When Hermes 0.20.1 talks to OmniRoute (http://localhost:20128/v1) using the antigravity provider (antigravity/gemini-3.6-flash-medium), requests are unreliable and frequently fail in three distinct ways:
1. Streams die mid-flight — OmniRoute returns HTTP 200 but the SSE stream ends with no finish_reason (empty/truncated body). Hermes sees EmptyStreamError and the turn returns nothing.
2. Connection refused — during OmniRoute startup/restart, all requests hard-fail with ECONNREFUSED and exhaust Hermes's 3 retries instead of waiting.
3. 429 rate-limit lockouts — the Antigravity upstream (Gemini) throttles heavily; OmniRoute relays 429/503 and drops in-flight streams, so the user gets repeated rate-limit errors and 60–90 s hangs.
There's also intermittent DNS failure (EAI_AGAIN on www.googleapis.com) inside OmniRoute's fetch layer.
Net effect: the antigravity path via OmniRoute is effectively unusable for normal daily use.

### Steps to Reproduce

1. Start OmniRoute container (`omniroute`, host network, listens on `0.0.0.0:20128`).
2. Point Hermes at `http://localhost:20128/v1` with model `antigravity/gemini-3.6-flash-medium`.
3. Issue normal chat/stream requests (85+ msgs, 11 tools triggers it reliably).
4. Observe intermittent failures during streaming

### Expected Behavior

- Requests during container startup should queue/retry gracefully, not hard-fail 3×.
- SSE responses must terminate with a `finish_reason` (or an explicit JSON error),
  never an HTTP 200 with a truncated/empty stream.
- Upstream 429s should be surfaced to the client as a proper error with
  `retry_after` semantics, and should NOT abort in-flight SSE streams silently.

### Actual Behavior

1. Connection Refused (Client-Side `ECONNREFUSED`)

[...]

Issue #8649, opened 2026-07-26
fix(resilience): auto/* combos swallow upstream auth failure — empty finish:stop with no error when tools attached


plaxorizor:
# OmniRoute bug: `auto/*` combo swallows upstream auth failure → empty `finish: stop` (no error)

## Environment
- OmniRoute: v3.8.48, self-hosted, single-port `:20128`
- Client: Oh My Pi (`omp`) v17.1.3, OpenAI-completions API, **with tools attached**
- Provider keys configured: `openrouter`, `cohere`, `blackbox`, `cerebras` (all active)
- Provider `aug` (auggie / Claude OAuth) is **NOT configured** (no active connection)

## Symptom
Requests to any built-in `auto/*` combo model (`auto/best-free`, `auto/best-coding`,
`auto/coding:free`, etc.) return an HTTP 200 with a **clean `finish_reason: "stop"` and
**`content: null`** when the client attaches `tools`. No error is surfaced to the client,
so the agent sees an "empty assistant turn" and retries until its retry cap.

Direct (non-combo) passthrough models — e.g. `openrouter/free`, `kc/openrouter/free`,
`kilocode/openrouter/free` — work correctly **with the exact same tool-bearing request**.

## Root cause (from OmniRoute's own call log)
Every `auto/*` combo routes its first step to the **`auggie/claude-sonnet-4.6`**
backend (`provider: "auggie"`). That provider has no credential on this install:

```
"summary": {
  "comboName": "auto/best-free",
  "comboStepId": "virtual-auto-cheap-51-auggie",
  "requestedModel": "auggie/claude-sonnet-4.6",
  "provider": "auggie",
  "account": "noauth",
  "connectionId": "noauth",
  "apiKeyId": null,
  "apiKeyName": null,
  "tokens": { "in": 0, "out": 0 }
}
"requestBody": { "model": "auggie/claude-sonnet-4.6", "stream": true,
                 "n_messages": 2, "n_tools": 1 }   # payload forwarded OK
"responseBody": { "choices": [ { "message": { "content": null },
                                      "finish_reason": "stop" } ] }
"error": null
```

The combo step reaches a backend with **no auth** (`account: "noauth"`, `apiKeyId: null`).
OmniRoute returns `200` + `finish: stop` + `content: null` **with `error: null`** —
it swallows the upstream auth failure instead of surfacing it or failing over to a funded
backend in the combo.

## Why this looks like a client bug but isn't
[...]

Issue #7699, opened 2026-07-19
fix(claude): /v1/messages reported silent SSE close on mid-stream upstream failure


glazec:
## OmniRoute Version
3.8.49 (also observed on 3.8.48)

## Installation Method
Docker / Docker Compose (self-hosted on Railway, behind a DNS-only custom domain; no Cloudflare proxy in path)

## Operating System
Linux (container, Node.js v24.18.0)

## Summary
On the `/v1/messages` (Anthropic format) route, when the **upstream SSE stream fails mid-flight after some bytes have already been forwarded**, OmniRoute **silently closes the response stream** — it emits no terminal `event: error`, no `message_delta`/`message_stop`. Anthropic-format clients (Claude Code, the Anthropic SDK) require a stream to terminate with `message_stop`, so a silent close is surfaced to the user as:

```
API Error: Connection closed mid-response. The response above may be incomplete.
```

The client cannot distinguish a truncated stream from a clean one, and it does not trigger a clean retry — the user just sees a partial/aborted response.

## Root Cause (code pointer)
`open-sse/utils/earlyStreamKeepalive.ts`, slow-path upstream-forwarding loop:

```ts
} catch (readErr) {
  // Upstream stream failed mid-flight. Only emit an error frame if
  // NO content was forwarded yet — otherwise the client already
  // received partial content and a late error frame would corrupt
  // the SSE stream. Silently close instead; the client will see
  // the stream end naturally.
  if (bytesForwarded === 0) {
    controller.enqueue(ERROR_FRAME);
  }
}
```

When `bytesForwarded > 0`, the stream is closed with no terminal frame. The justifying comment ("a late error frame would corrupt the SSE stream") is **OpenAI-centric**. For the Anthropic Messages format this does not hold: Anthropic's own SSE spec explicitly permits a mid-stream `event: error`, and a stream that ends *without* `message_stop` is treated as an error by Anthropic clients. So for `/v1/messages`, "silently close and let the stream end naturally" is precisely what produces the client-visible failure.

## Why this is NOT the prefill/keepalive path
The early-keepalive wrapper works correctly. Verified on a ~205k-token request to `kmc/kimi-k3` with `Accept: text/event-stream`:

- First `event: ping` at **2.2s** (matches the 2s default threshold for OAuth provider `kmc`), then pings roughly every ~1.1s
- **14 ping frames** emitted during the ~17.8s prefill gap
[...]

PR #10475, opened 2026-08-15
fix(sse): flag OpenAI streams that close with content but no terminal marker

HouMinXi:
## Problem

Issue #10443 (antigravity via OmniRoute unusable): when the upstream kills an SSE
stream mid-generation without a terminal marker, OmniRoute closed the stream
silently for OpenAI chat-completions clients. The client saw HTTP 200, a few
content chunks, then a bare close with no `finish_reason` — Hermes reports it as
`EmptyStreamError` and the turn returns nothing.

The silent close was deliberate: `resolveSilentCloseReason()` (#7699) only flags
a close-without-terminal-marker for Claude-format clients, because "many formats
have no `[DONE]` equivalent". That reasoning does not hold for OpenAI chat
completions — a healthy OpenAI stream always ends with either a `finish_reason`
chunk (translator/upstream) or `data: [DONE]`. A close that forwarded content
but neither marker is an upstream drop there too.

Concretely for antigravity/Gemini: the executor is a pure byte passthrough and
the `gemini-to-openai` translator emits the final `finish_reason` chunk only
when the upstream `finishReason` event arrives. When Gemini kills the stream
(e.g. its own rate enforcement) the forwarded bytes contain zero terminal
marker, and the disconnect-aware layer closed the stream silently.

## Fix

`open-sse/utils/streamHandler.ts`, two small changes:

1. `hasClientTerminalSseMarker()`: for `FORMATS.OPENAI`, a non-null
   `finish_reason` value now counts as a terminal marker (some providers
   legitimately omit `data: [DONE]`). A bare `finish_reason: null` delta chunk
   still does NOT count.
2. `resolveSilentCloseReason()`: the "ended without a terminal marker" verdict
   now also fires for `FORMATS.OPENAI` when content was forwarded
   (`sawContent()` guard, so the #8649 empty-content verdict keeps its more
   precise shape for content-free closes).

On the drop, the existing in-band error path emits an OpenAI error chunk
(`finish_reason: "error"`) plus `data: [DONE]` — the same machinery #7699
built; nothing new is synthesized.

## Verification

- TDD: new `tests/unit/silent-sse-close-openai-10443.test.ts` — the core case
  (content then bare close) was RED before the fix and is GREEN after. Four
  guard cases pin the boundaries: a `finish_reason` chunk counts as terminal,
  `data: [DONE]` closes stay unchanged, content-free closes keep the #8649
  empty-content verdict, and a literal `"finish_reason": "stop"` inside model
  output text does NOT trip the marker (JSON escaping keeps the raw bytes from
  matching the unescaped-field regex).
- Bug-injection at each fix site: deleting the resolver's OpenAI branch turns
  the core case RED; deleting the marker detection turns the
  `finish_reason`-counts-as-terminal case RED. Both restored → GREEN.
- Regression: 72/72 across the stream-handler suites (`silent-sse-close-7699`,
  `empty-stream-no-content-8649`, `stream-handler*`,
  `chatcore-streaming-pipeline`, and the new file). `npm run typecheck:core`
  clean; Prettier clean.

⚠️ base-red inherited: #9985

### DIFF
--- a/open-sse/utils/streamHandler.ts
+++ b/open-sse/utils/streamHandler.ts
@@ -210,6 +210,14 @@ function hasClientTerminalSseMarker(text: string, clientResponseFormat?: string 
     );
   }
 
+  // OpenAI chat completions: some providers omit `data: [DONE]` (already
+  // matched above) and terminate with a finish_reason chunk instead. A
+  // non-null finish_reason value is that terminal signal — a bare
+  // `finish_reason: null` delta chunk must NOT count (#10443).
+  if (clientResponseFormat === FORMATS.OPENAI) {
+    return /"finish_reason"\s*:\s*"[^"]+"/.test(text);
+  }
+
   return false;
 }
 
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Reviewed PR #10475 against release/v3.8.50. The change is narrowly scoped and the resolver behavior is coherent: OpenAI streams now report a synthetic 502 error only after forwarding real content and seeing neither a non-null finish_reason nor data: [DONE]; legitimate finish_reason and [DONE] termination remain accepted, and content-free streams retain the #8649 verdict. Existing #7699/#8649 focused tests pass (13/13). I could not execute the new test file because it is not present in this checkout, so the PR-head claims remain unverified here. I found no blocking defect; merge-ready pending normal CI.

diegosouzapw:

PR #10478, opened 2026-08-15
fix: resolve adaptive latency-collapse self-lock with solo-progress and idle recovery (#10111)

### LINKED ISSUE #10111: fix(admission): prevent adaptive latency collapse from permanently locking out normal requests
### Problem

The adaptive admission controller's `enforce` mode can permanently lock out normal requests after its budget shrinks below their cost.

The controller adapts from the lease latency recorded by `attachResponseLifecycle()`. For non-streaming responses this is full model-generation latency; for SSE it is the complete stream lifetime. Those durations are dominated by provider/model speed, output length, reasoning, and client consumption—not OmniRoute preparation pressure.

A rising full-response latency gradient reduces `currentLimit`. In enforce mode, any request with `cost > currentLimit` is immediately rejected as `ADMISSION_OVERSIZED`. Once the limit falls below the normal request cost, those requests cannot complete, and the additive recovery path requires admitted completed work. Idle windows reset the stale latency gradient but do not raise the limit. This creates a self-locking state.

### Live shadow evidence

On a normal Hermes multi-agent workload using the current `release/v3.8.50` defaults:

```json
{
  "mode": "shadow",
  "currentLimit": 28,
  "minLimit": 8,
  "activeCost": 63,
  "activeCount": 1,
  "admittedCount": 215,
  "wouldAdmitCount": 61,
  "wouldRejectCount": 154,
  "pressure": "normal",
  "resourceSeverity": "normal",
  "pressureGuardRejectCount": 0
}
```

154 of 215 shadow decisions (71.6%) would have rejected, despite normal resource pressure. The active ordinary request cost 63 while the adapted budget was 28. Actual rejects remained zero only because mode was `shadow`.

The observed request bodies were ordinary agent turns around 128–227 KiB, 2–32 messages, and 35 tools. Their provider durations were typically 11–37 seconds.

### Deterministic reproduction

Using `AdaptiveAdmissionController` in enforce mode with the shipping defaults (`initialLimit=64`, decrease factor 0.8):

1. Admit/release cost 63 with latency 1000 ms, then close the window: limit becomes 65.
2. Admit/release cost 63 with latency 10,000 ms, then close the window.
3. Short EWMA becomes 5500, long EWMA 1900; the positive gradient shrinks the limit to 52.
4. The next identical cost-63 request is rejected as `ADMISSION_OVERSIZED`.
5. Advance and close 20 idle windows: the limit remains 52 with no active or queued work.

Observed:

```json
{
  "requestCost": 63,
  "afterBaseline": {
    "currentLimit": 65,
    "shortLatencyEwma": 1000,
    "longLatencyEwma": 1000
  },
  "afterLatencyGradient": {
    "currentLimit": 52,
    "shortLatencyEwma": 5500,
    "longLatencyEwma": 1900
  },
  "nextRequest": {
    "status": "rejected",
    "code": "ADMISSION_OVERSIZED"
  },
  "afterTwentyIdleWindows": {
    "currentLimit": 52,
    "activeCount": 0,
    "queuedCount": 0
  }
}
```

### Why the signal is confounded

`open-sse/services/admission/runtime.ts` binds lease release to full JSON/SSE completion and reports that elapsed duration through `noteLatency()`.

`open-sse/services/admission/adaptation.ts` decreases the budget when the short-vs-long latency gradient exceeds 0.25.

This measures provider response time and stream duration. It does not isolate local admission wait, request preparation, event-loop delay, heap pressure, or time-to-upstream-dispatch. A model producing more tokens can reduce local concurrency even while V8/RSS/cgroup pressure is normal.

### Proposed correction

1. Do not use full provider response/SSE lifetime as the adaptive overload latency signal.
2. Measure local phases separately: admission wait, body ingest/parse, guardrails/compression/translation/serialization, time to upstream dispatch, event-loop delay, and resource pressure.
3. Treat `maxRequestCost` as the hard individual-request ceiling; do not equate temporary `currentLimit` with permanent oversized classification.
4. Guarantee progress when one valid request exceeds the temporary concurrency budget—for example, allow one bounded solo request when no other work is active, while preserving the independent resource-pressure fuse and hard limits.
5. Add recovery/probing that does not require requests already blocked by the shrunken budget to complete.
6. Keep `shadow` as the default until production-shaped canaries demonstrate that ordinary request costs remain serviceable and the controller recovers from decreases.

### Acceptance criteria

- Provider/model generation latency and long SSE duration alone cannot collapse the local admission budget under normal resource pressure.
- A temporary budget below a valid normal request cost cannot permanently reject that traffic.
- At least one request can make bounded forward progress when the process is idle and pressure is normal, subject to hard body/request-cost limits.
- Recovery is demonstrated after high/critical pressure and after a latency-driven decrease.
- Tests cover rising output duration with stable local preparation cost, mixed provider speeds, long SSE streams, individually valid requests above temporary budget, idle recovery, and pressure-fuse precedence.
- Canary diagnostics separate local preparation latency from provider/stream duration.

### Related

- #9262: introduced adaptive admission in shadow mode.
- #10054: tracks graduation to enforce mode.
- #9176: phase-aware preparation/stream separation.

diegosouzapw:
Closes #10111

## Root cause

`AdaptiveAdmissionController` in `open-sse/services/admission/` used full end-to-end request latency (provider generation / SSE stream lifetime) as its overload signal. A slow-provider turn shrinks the aggregate `currentLimit` below an ordinary request's cost; in `enforce` mode `acquire()` then hard-rejected that request as `ADMISSION_OVERSIZED`. Idle recovery could never fire because the only limit-increase path requires a completed admission — which can never happen once nothing can be admitted. Self-lock, reproduced deterministically with the shipping defaults (`initialLimit=64`, `decreaseFactor=0.8`): two ordinary cost-63 releases (1s then 10s latency) collapse the limit, the next identical request is rejected forever, and 20 idle windows never recover it.

## Fix (two surgical, additive changes)

- **Solo-progress** (`controller.ts`, new `shouldAdmitSolo` helper): a request within the healthy aggregate ceiling (`maxLimit`) is no longer terminally rejected as oversized when the system is otherwise idle (no active/queued/lane work) and pressure is normal — one bounded solo request is admitted so the pipeline makes progress. The hard ceiling, the critical/high pressure fuse, and a busy system (active/queued work) all still take precedence, so genuine load continues to shed.
- **Idle recovery** (`adaptation.ts`, new `applyIdleRecovery` helper + `recoveryCeiling` state): on a window with no completed work and low utilization (system idle), the collapsed limit actively climbs back toward the recovery ceiling (initialLimit) instead of only resetting the latency EWMA baseline.

## Regression test (fail → pass)

`tests/unit/adaptive-admission-latency-collapse.test.ts` (the triage plan's exact repro harness — same FakeClock, same shipping defaults). Before: cost-63 request `ADMISSION_OVERSIZED` + 20 idle windows never recover. After, all 4 tests pass:
- collapsed limit admits an idle cost-63 request (solo-progress), not `ADMISSION_OVERSIZED`
- 20 idle windows recover the limit and re-admit a normal request
- critical-pressure fuse still wins over solo-progress
- solo-progress does not bypass `maxLimit` or run under active/queued load

Also `tests/unit/adaptive-admission-controller.test.ts` codified the old no-idle-recovery behavior in two places; aligned them to assert non-multiplication + recovery instead of frozen idle (legitimate alignment, not weakened).

## Gates (all green)

- typecheck: `npm run typecheck:core` exit 0 (×2)
- lint on changed files with suppressions: exit 0
- file-size: OK; complexity: 2465 < baseline 2774; cognitive: 1104 < baseline 1223
- changelog-integrity: OK; new fragment `changelog.d/fixes/10111-adaptive-admission-latency-collapse.md`
- admission-area unit suites: controller/cost/features/queue/domain/route-matrix/runtime/lifecycle/virtual-lanes/per-connection/binding/wrapper + new regression = 125 tests pass

Note: this change is confined to `open-sse/services/admission/`. The CI `test-vitest` suite (autoCombo/MCP/cache) is untouched but on this loaded devbox several autoCombo/Sidebar cases flaked with 5–20s timeouts (non-deterministic across runs); they are load-induced and unrelated to this diff.

⚠️ base-red inherited: #9985

### DIFF
--- a/changelog.d/fixes/10111-adaptive-admission-latency-collapse.md
+++ b/changelog.d/fixes/10111-adaptive-admission-latency-collapse.md
@@ -1,0 +1,1 @@
+- **fix(admission):** stop the adaptive latency-gradient collapse from permanently locking out ordinary requests — individually valid requests now make solo progress when the system is idle and normal pressure, and the collapsed limit actively recovers on sustained idle windows instead of being stuck; the critical-pressure fuse still wins over solo progress (#10111)
--- a/open-sse/services/admission/adaptation.ts
+++ b/open-sse/services/admission/adaptation.ts
@@ -17,6 +17,13 @@ export interface AdaptationParams {
 
 export interface AdaptationState {
   currentLimit: number;
+  /**
+   * Idle-recovery target: the healthy starting aggregate budget (initialLimit).
+   * Used to climb the limit back up when a latency-gradient decrease has collapsed it
+   * below serviceable requests but the system is otherwise idle (#10111). Never grows
+   * beyond the configured maxLimit.
+   */
+  recoveryCeiling: number;
   shortLatencyEwma: number;
   longLatencyEwma: number;
   pressure: AdmissionPressure;
@@ -47,6 +54,7 @@ export function createAdaptationState(
 ): AdaptationState {
   return {
     currentLimit: clampLimit(initialLimit, minLimit, maxLimit),
+    recoveryCeiling: clampLimit(initialLimit, minLimit, maxLimit),
     shortLatencyEwma: 0,
     longLatencyEwma: 0,
     pressure: "normal",
@@ -138,6 +146,14 @@ export function closeAdaptationWindow(
   // A genuinely low-utilization window recovers the latency baseline so stale gradients expire.
   if (state.utilization <= params.lowUtilizationThreshold) {
     state.shortLatencyEwma = state.longLatencyEwma;
+    // #10111 idle recovery (extracted helper): a latency-gradient decrease must not
+    // permanently lock the aggregate budget below serviceable requests. On a window with no
+    // completed work and low utilization (system idle), actively raise the limit back toward
+    // the recovery ceiling so ordinary requests can re-enter. The high-utilization/completed
+    // work increase branch above handles growth under load; this covers the no-progress
+    // starvation case. A window that completed a request (windowCompleted > 0) is the one
+    // whose latency samples triggered a decrease, so the two branches never fight.
+    next = applyIdleRecovery(state, params, next);
   }
 
   state.currentLimit = clampLimit(next, params.minLimit, params.maxLimit);
@@ -148,6 +164,25 @@ export function closeAdaptationWindow(
   state.freezeGrowth = false;
   state.criticalDecreaseConsumed = false;
   state.pressure = "normal";
+}
+
+/**
+ * #10111 idle-recovery helper. When a latency-gradient decrease has collapsed the aggregate
+ * limit below serviceable requests and the system is idle (no completed work, low
+ * utilization, normal non-critical pressure), raise the limit back toward the recovery
+ * ceiling by one bounded step so ordinary requests can re-enter.
+ */
+function applyIdleRecovery(state: AdaptationState, params: AdaptationParams, next: number): number {
+  if (
+    state.pressure !== "critical" &&
+    !state.freezeGrowth &&
+    state.windowCompleted === 0 &&
+    state.currentLimit < state.recoveryCeiling
+  ) {
+    const step = Math.min(params.increaseStep, params.maxIncreasePerWindow);
+    return Math.min(state.recoveryCeiling, next + step);
+  }
+  return next;
 }
 
 export function sampleActiveIntegral(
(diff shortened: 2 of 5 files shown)


diegosouzapw:
Found one confirmed correctness issue. The PR-head focused suites pass (33/33), but updateConfig changes initialLimit without updating the newly introduced recoveryCeiling. A live controller updated from initialLimit 5 to 10 remains capped at 5 after 20 idle windows; the inverse update can likewise recover above the new target. Refresh the ceiling in updateConfig and add a regression covering both directions. Base-red #9985 is inherited and not attributed to this PR.

PR #10481, opened 2026-08-15
fix: downgrade adaptive thinking and gate context-1m beta on model eligibility (#10119)

### LINKED ISSUE #10119: bug(providers): correct supportsReasoning/long-context capability for claude-haiku-4-5-20251001
## Summary

`ensureThinkingBudget()` (the server-side logic that decides whether to attach a thinking/reasoning budget to an outgoing request) consults a hardcoded static provider registry via `getRegistryEntry(provider).models.find(id).supportsReasoning`, **not** the `model_capabilities`/`model_intelligence` DB tables. That registry entry is wrong for `claude-haiku-4-5-20251001` (and likely other dated Haiku aliases): it's flagged as reasoning-capable, but Anthropic's real API rejects thinking params for Haiku models.

A second, same-shaped bug: requests to this model also carry a long-context beta header the account/model doesn't qualify for.

## Observed errors (from live `call_logs`, self-hosted instance, v3.8.49)

```
[400]: adaptive thinking is not supported on this model
[400]: long context beta is not yet available for this subscription
```

## Measured impact

While `claude-haiku-4-5-20251001` was a priority-tier member of a coding combo (real production traffic, not synthetic), it failed with a 400 on **roughly 45-50% of requests** — essentially a coin flip driven by whether the originating client request happened to carry a thinking/reasoning param. Traced via `combo_execution_key` grouping in `call_logs`: dozens of alternating 200/400 pairs across a multi-hour window on the same combo/model pairing.

## Root cause (traced in the installed npm bundle, `dist/.build/next/server/chunks/*.js`)

```js
// paraphrased from minified output
if (typeof requestOpts === "object" && requestOpts !== null) this.ensureThinkingBudget(requestOpts, e);
...
let entry = getRegistryEntry(this.provider),
    modelEntry = entry?.models?.find(m => m.id === resolvedModelId);
if (!modelEntry?.supportsReasoning) return requestOpts; // early-out skipped because supportsReasoning is (wrongly) true
```

The gating flag comes from the **static bundled registry**, not from the DB-backed `model_capabilities` table (which — separately — also has `reasoning=1` set for this model, so both the DB catalog and the static registry appear to have copy-pasted Sonnet/Opus capability flags onto the Haiku entry without correcting them).

I also checked whether `model_capability_overrides` (a DB table that looks purpose-built to override exactly this kind of thing) is consulted anywhere in the thinking-budget path — it has zero call sites near `ensureThinkingBudget` in the bundle, so it appears unwired for this purpose currently.

## Suggested fix

- Correct `supportsReasoning` (and the long-context/1M-context capability flag) to `false` for `claude-haiku-4-5-20251001` and any other Haiku aliases in the static provider registry, to match Anthropic's actual API capabilities.
- Separately, consider wiring `ensureThinkingBudget`'s capability check through `getModelCapabilityOverride()` so self-hosted users can correct registry drift like this without waiting on a release, since the registry will presumably keep going stale as new model aliases ship.

## Environment

- omniroute@3.8.49 (npm, self-hosted)
- macOS, Node via `/opt/homebrew`
- Provider: `claude` (Claude subscription/OAuth connection), model `claude-haiku-4-5-20251001`

Happy to provide more `call_logs` excerpts or test a patch if useful.

diegosouzapw:
Closes #10119

⚠️ base-red inherited: #9985

## Root cause

The reported 400s ("adaptive thinking is not supported on this model" and "long context beta is not yet available for this subscription") on `claude-haiku-4-5-20251001` are real, but the issue's suggested fix (flipping `supportsReasoning`/`supportsThinking` to `false`) would be a regression — the static registry already has no `supportsReasoning` field for Haiku 4.5, and `supportsThinking:true` is Anthropic-doc-correct (Haiku 4.5 is the first Haiku model to support extended thinking, `type:"enabled"` + `budget_tokens` only). Two narrower per-target-model gaps produce the 400s when a combo/fallback re-routes a request that was shaped for a more capable sibling onto Haiku:

1. `open-sse/translator/request/openai-to-claude.ts`: a request already carrying `thinking:{type:"adaptive"}` (built for an adaptive-only sibling like Opus 4.7+/Sonnet-5) was forwarded verbatim, with no downgrade to the manual `enabled` form for models that only accept manual extended thinking.
2. `open-sse/config/anthropicHeaders.ts::mergeClientAnthropicBeta()`: a client-negotiated `context-1m-2025-08-07` beta was forwarded unconditionally with no model eligibility check.

## Changes

- **Fix #1** — downgrade `thinking.type:"adaptive"` → `type:"enabled"` + a safe budget when `!isAdaptiveThinkingOnly(model)` (keyed generically on the resolved model, not the Haiku id, so it also covers any other non-adaptive-only Claude model reached via combo/fallback).
- **Fix #2** — `mergeClientAnthropicBeta()` now accepts an optional resolved `model` and drops `context-1m-2025-08-07` when `modelSupportsContext1mBeta(model)` is false; the base.ts caller threads in the resolved model. The eligibility helper moved into a neutral `open-sse/config/context1m.ts` module so both `anthropicHeaders.ts` and `claudeCodeCompatible.ts` share one source of truth (no circular import).

## Regression tests (fail → pass)

- `tests/unit/repro-10119-claude-haiku-adaptive-downgrade.test.ts` — was RED (translator kept `thinking.type:"adaptive"` for haiku-4-5 / sonnet-4-5), now GREEN (downgraded to `enabled` with budget ≥1024; Opus 4.7+ still keeps `adaptive`).
- `tests/unit/repro-10119-claude-context1m-beta-gated.test.ts` — was RED (context-1m forwarded to Haiku), now GREEN (dropped for Haiku, kept for sonnet-5, preserved when no model is supplied).
- `tests/unit/claude-context-1m-supported-models.test.ts` — source-parse path aligned to the new `context1m.ts` module (legitimate path move, assertions unchanged).

## Gates green

- `npm run typecheck:core` exit 0; `check:cycles` OK; `check-file-size` OK; `check-changelog-integrity` OK; `check-complexity` 2465 (baseline 2774) OK; `check-cognitive-complexity` 1104 (baseline 1223) OK; eslint (with suppressions) clean on changed files; `npm run test:vitest` full suite 362/362 green; `test:unit` touched-area slice 220/220 green.

Plan file: `_tasks/pipeline/bugs/2-implementing/10119-claude-haiku-45-capability-flags.plan.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -167,6 +167,7 @@ _Living section — regenerated 2026-08-12 from all cycle commits (cycle open `e
 
 ### 🐛 Bug Fixes
 
+- **providers**: honor `PATCH /api/providers/[id]` so `omniroute providers rotate` stops 405ing (the OpenAPI spec and CLI already use PATCH) (PR #10366)
 - **executors**: fix internal timeout misclassified as client disconnect (499) for 7 niche executors — pass TimeoutError reason to controller.abort() (#8197 side-finding)
 - test(combo): guard auto/best-free never leaks the combo name as a model (#7754)
 - fix(vision-bridge): describe-model no longer returns unreachable "openai/gpt-4o-mini" when every vision-capable provider is unreachable on the instance — returns null instead and surfaces a clear error (#8430)
--- a/changelog.d/fixes/10119-claude-haiku-45-capability-flags.md
+++ b/changelog.d/fixes/10119-claude-haiku-45-capability-flags.md
@@ -1,0 +1,1 @@
+- fix(sse): downgrade client-supplied `thinking:{type:"adaptive"}` to `enabled` and gate the `context-1m-2025-08-07` beta on model eligibility when a combo/fallback re-routes a request to a non-adaptive/non-1M model like claude-haiku-4-5 (avoids "adaptive thinking is not supported on this model" and "long context beta is not yet available" 400s, #10119)
--- a/open-sse/config/anthropicHeaders.ts
+++ b/open-sse/config/anthropicHeaders.ts
@@ -6,6 +6,7 @@ import {
   CLAUDE_CODE_SDK_PACKAGE_VERSION,
   getClaudeCodeUserAgent,
 } from "@/shared/constants/claudeCodeClient";
+import { modelSupportsContext1mBeta } from "../config/context1m.ts";
 
 export const ANTHROPIC_VERSION_HEADER = "2023-06-01";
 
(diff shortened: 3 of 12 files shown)


diegosouzapw:
Encontrado um bloqueio importante em open-sse/executors/default.ts:391 e 594-597. A nova assinatura de mergeClientAnthropicBeta recebe o model, e BaseExecutor.execute passa o model como quarto argumento, mas DefaultExecutor.buildHeaders ainda aceita apenas três argumentos e chama o merge sem model. Probe executado no head exato 3ea8cd8: ao construir headers para claude-haiku-4-5-20251001 com anthropic-beta: context-1m-2025-08-07, o beta continuou sendo emitido. Portanto, o caminho Claude padrão/API-key ainda pode reproduzir o 400 de long-context. Os 7 testes focados do PR passam, mas exercitam o helper diretamente, não o executor. Passe o model nessa rota e adicione um teste de integração de headers antes do merge.

PR #10489, opened 2026-08-15
fix(memory): auto-check Qdrant health on mount and stop false-red badge

rqzbeh:
## Summary

On `/dashboard/memory?tab=engine`, the Qdrant card showed a **red "Error" badge after every page refresh** even though "Test connection" always returned OK.

**Root cause (verified against a live healthy deployment):**
- "Test connection" is real — it pings Qdrant's `/readyz` **server-side** (`/api/settings/qdrant/health` → `checkQdrantHealth`), and on the reported setup it genuinely succeeds (`{"ok":true,"latencyMs":2}`).
- The red badge is a **UI state bug**: the badge renders red whenever `health?.ok` is falsy, but the mount effect only fetched settings + embedding models — `health` starts as `null` after a refresh, so `health?.ok` is `undefined` → red. Clicking "Test connection" ran the same check and turned it green — hence "test OK, refresh red".
- Not docker-specific: it affects any deployment with Qdrant enabled. (The docker aspect — host resolution — is fine in the reported setup: `QDRANT_HOST`/settings use the compose service name `qdrant`.)

**Fix:**
1. Auto-run the health check on mount once settings load and Qdrant is enabled — a refreshed page now shows the real state (green) without a manual click.
2. While health hasn't been checked yet (`null`), render a neutral gray "Testing..." state — red is reserved for an actual failed check.

**Verification:**
- Live check: `/api/settings/qdrant/health` → `ok:true` (2ms), `/api/memory/engine-status` → `qdrant: {enabled:true, healthy:true}` on the reported deployment.
- New regression test (fails on the old code, passes with the fix): with enabled settings + healthy mock, the card must auto-fetch health on mount and show `statusActive`, never `statusError`.
- `vitest run tests/unit/ui/qdrant-config-card.test.tsx` → 6/6 pass.


diegosouzapw:
The neutral null state and mount health check address the reported false-red refresh state. One race remains: save() optimistically sets enabled and starts the PUT, while the new effect can immediately GET health against the old persisted settings. If that GET wins, it returns not_configured/failed; because health is then non-null, the effect never retries after the PUT succeeds, leaving a healthy Qdrant red until a manual test. Please invalidate health and schedule the check after a successful PUT (and cover enable/reconfigure ordering with a regression test). I could not execute the PR head in this read-only review checkout, so this finding is marked PLAUSIBLE.

rqzbeh:
Thanks for the sharp catch — you are right, and the race was even more stubborn than described: in the "GET wins" ordering the PUT resolves while `health` is still `null`, so the `setHealth(null)` after the PUT is a React no-op and the mount effect never re-fires, leaving the badge red until a manual test.

Fixed in 34692342:

1. **Generation counter** — `checkHealth()` now captures a sequence number from `healthSeqRef` and drops its result if a newer save/check bumped it. Any in-flight pre-save GET can no longer overwrite post-save state, regardless of response ordering.
2. **Invalidate + explicit re-check on successful PUT** — `save()` bumps the seq and nulls health at the start (settings are changing) and again after the PUT succeeds, then explicitly calls `checkHealth()` against the freshly persisted settings. The explicit call matters: with health already `null` (the GET-wins ordering) the effect would not re-fire, so relying on it alone was the bug.
3. **Regression test** — `re-checks health after a successful save so a stale optimistic-window result cannot leave the badge red (enable ordering)` reproduces the scenario: first health GET returns not_configured (old persisted settings), the PUT succeeds, and the test asserts a second health GET is scheduled and the badge ends on `statusActive`, never `statusError`.

The new test failed against my first effect-only attempt — which reproduced your exact ordering — and that is what surfaced the explicit re-check. Local runs: qdrant-config-card 7/7, vitest fast-path 362/362, `typecheck:core` and eslint clean. CI is re-running on the new head.

PR #10491, opened 2026-08-15
fix(cli): recognize {connections} envelope from /api/providers

hartmark:
## What Problem This Solves

`omniroute test --all-providers` and `omniroute oauth providers` both crash with `(data.providers ?? data.items ?? data).filter is not a function` on every invocation.

## Why This Change Was Made

`GET /api/providers` returns `{connections, total}` (`src/app/api/providers/route.ts:78`), but both CLI call sites parsed the response as `data.providers ?? data.items ?? data` — since none of those keys exist, `data` itself (a plain object, not an array) falls through, and `.filter` throws.

`bin/cli/commands/keys.mjs:222` already has the correct fallback chain (`data.keys || data.connections || data.items || data`); this applies the same `connections` field to the two remaining call sites that hit the same endpoint.

## User Impact

`omniroute test --all-providers` and `omniroute oauth providers` go from a hard crash to actually working.

## Evidence

Reproduced live against a running instance:

```
$ omniroute test --all-providers --json
✖ (data.providers ?? data.items ?? data).filter is not a function
```

```
$ curl -s "$BASE/api/providers?limit=200" -H "Authorization: Bearer $KEY" | python3 -c "import json,sys; print(list(json.load(sys.stdin).keys()))"
['connections', 'total']
```

After the fix, `omniroute test --all-providers --json` correctly enumerates configured connections instead of crashing.

### DIFF
--- a/bin/cli/commands/oauth.mjs
+++ b/bin/cli/commands/oauth.mjs
@@ -284,7 +284,7 @@ export async function runOAuthStatus(opts, cmd) {
     process.exit(1);
   }
   const data = await res.json();
-  const connections = (data.providers ?? data.items ?? data).filter(
+  const connections = (data.connections ?? data.providers ?? data.items ?? data).filter(
     (c) => c.authType === "oauth" || c.authType === "oauth2"
   );
   emit(connections, globalOpts, connectionSchema);
--- a/bin/cli/commands/test-provider.mjs
+++ b/bin/cli/commands/test-provider.mjs
@@ -80,7 +80,7 @@ async function _runAllProviders(opts) {
     return 1;
   }
   const data = await res.json();
-  const connections = (data.providers ?? data.items ?? data).filter(
+  const connections = (data.connections ?? data.providers ?? data.items ?? data).filter(
     (c) => c.authType === "apikey" || c.testStatus !== "unavailable"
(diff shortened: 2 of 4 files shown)


diegosouzapw:
A correção de produção está correta: GET /api/providers retorna {connections, total}, e o probe do head confirma que ambos os filtros passam a operar sobre esse envelope. Contudo, o PR altera dois caminhos CLI e não adiciona testes de regressão para {connections}; o teste OAuth existente cobre apenas {providers}, e não há cobertura dedicada para --all-providers. Adicione testes focados antes do merge para cumprir as Hard Rules #8/#18 e evitar a reintrodução do crash.

hartmark:
Focused regression tests for both CLI consumers of the `{connections}` envelope are already on this branch at the current head (`6dbd5b798`):

- `tests/unit/cli-oauth-commands.test.ts`: `runOAuthStatus consumes the connections envelope`
- `tests/unit/cli-expanded-commands.test.ts`: `test-provider --all-providers consumes the connections envelope`

Both assert the CLI correctly reads `{connections}` (not just `{providers}`) for `runOAuthStatus` and `--all-providers` respectively. Should be good to re-check.

PR #10495, opened 2026-08-15
fix: repair corrupted DeepSeek Responses SSE deltas for split multi-byte UTF-8 (#10223)

### LINKED ISSUE #10223: bug(responses): corrupted DeepSeek SSE deltas while terminal done text is correct
### OmniRoute Version

3.8.49 (official npm package). I also inspected unreleased `release/v3.8.50` at
`f1eb0b835722ca744692834c84661f1bc8d564a6`; it has adjacent Responses fixes but no
regression test for this delta/done inconsistency.

### Installation Method

Docker image built with:

```dockerfile
RUN npm install --global --omit=dev --omit=optional omniroute@3.8.49
```

### Provider / Client

- DeepSeek through OmniRoute's `POST /v1/responses`
- Reasonix CLI, `stream: true`
- A raw HTTP streaming probe reproduces the malformed SSE before Reasonix parses it

### Description

Some DeepSeek Responses streams contain corrupted incremental text while the matching
terminal `*.done` event contains the correct complete text.

Affected event families observed:

- `response.reasoning_summary_text.delta`
- `response.output_text.delta`
- function-call partial metadata/arguments on the same streaming path

The matching `response.reasoning_summary_text.done`, `response.output_text.done`, and
`response.output_item.done` events are valid. The request returns HTTP 2xx, reaches
`response.completed`, and contains normal usage; there is no provider error.

### Minimal Reproduction

```bash
curl -sN https://<omniroute-host>/v1/responses \
  -H 'Authorization: Bearer <redacted>' \
  -d '{
    "model": "<deepseek-route>",
    "input": "Reply exactly: REASONIX_OMNIROUTE_DEEPSEEK_OK",
    "stream": true
  }'
```

A sanitized reproduction produced:

```text
response.output_text.delta:
  "ogb9_9_9b9b99_9_0_0R09_0"

response.output_text.done:
  "REASONIX_OMNIROUTE_DEEPSEEK_OK"

response.completed:
  normal completion with usage
```

The client-visible result became:

```text
ogb9_9_9b9b99_9_0_0R09_0REASONIX_OMNIROUTE_DEEPSEEK_OK
```

Reasoning deltas showed the same class of malformed/repeated fragments while the
reasoning done snapshot was correct. A direct non-streaming request returned clean text.

### Expected Behavior

For every item/content index:

1. Concatenated text deltas equal the corresponding terminal done text.
2. Delta bytes preserve upstream text exactly and in order.
3. Function-call IDs, names, and arguments remain consistent across partial and done events.
4. Incremental and terminal-snapshot consumers receive the same semantic output.

### Impact

- Standard Responses clients append the malformed deltas, corrupting final output.
- Long agent runs may continue producing malformed reasoning after work completed and eventually
  hit client output limits such as `reasoning output exceeded client byte limit`.
- Partial tool-call metadata/arguments may be polluted before authoritative done events arrive.

The malformed bytes are present in OmniRoute's raw SSE. A client can work around this only by
buffering all deltas and replacing them with done snapshots, which removes live streaming and
should not be required by the Responses contract.

### Related but Distinct

- #6952 covers duplicate Responses-to-Chat reassembly and tool arguments leaking into content.
- #9170 covers whitespace-only deltas being removed.
- #10156 covers commentary SSE suppression versus `response.completed`.

This report is specifically about corrupted DeepSeek Responses deltas whose matching done
snapshots are correct.

### Suggested Regression Test

Add a recorded DeepSeek/OpenAI-chat-to-Responses fixture with fragmented reasoning, output text,
and a function call, then assert:

```ts
join(outputTextDeltas) === outputTextDone.text
join(reasoningDeltas) === reasoningDone.text
reconstructedToolCall === outputItemDone.item
```

Run it through the real `POST /v1/responses` streaming pipeline so sanitizer, translator,
stream transforms, and SSE serialization are covered.

All endpoint URLs, credentials, request/account/provider IDs, and production prompts are redacted.

diegosouzapw:
Closes #10223

## Root cause

`open-sse/transformer/responsesTransformer.ts:580` recreated a brand-new `new TextDecoder()` on every `transform()` chunk and decoded it WITHOUT `{ stream: true }`. A stream:false decoder has no cross-call state, so whenever a multi-byte UTF-8 character (CJK: 3 bytes, emoji: 4) is split across two TCP chunks (the normal case in Chinese streaming text, exactly the reporter's DeepSeek scenario), the trailing partial bytes were replaced with U+FFFD and every emitted delta accumulated garbage. The terminal *.done snapshot stayed identical to the buffer it was rebuilt from, so `join(deltas) === done` did not catch the corruption (that is precisely why the original analysis classified this as needs-vps).

## Fix

Hoist a single persistent `const decoder = new TextDecoder()` into the stream scope, decode with `decoder.decode(chunk, { stream: true })` in `transform()`, and drain `decoder.decode()` on `flush()`.

## Regression test (fail -> pass)

`tests/unit/responses-transformer-cjk-split.test.ts` feeds a CJK string split at a byte boundary inside a multi-byte char and asserts a round-trip to the SOURCE text (not the old `join(deltas) === done` invariant). Confirmed RED on the unfixed code (`REASONIX_文测试...` became replacement chars), GREEN after the fix.

## Gates run green

- `node scripts/check/check-file-size.mjs` -> OK
- `node scripts/check/check-complexity.mjs` + `node scripts/check/check-cognitive-complexity.mjs` -> OK
- `node scripts/check/check-changelog-integrity.mjs` -> OK
- `npm run typecheck:core` -> exit 0
- lint (with `--suppressions-location config/quality/eslint-suppressions.json`) -> exit 0
- `npm run test:vitest` -> 362/362 pass
- `npm run test:unit` -> 31120-31129 pass; the only failures were (a) pure load-induced timing flakes in network/cooldown/live tests, all proven to pass in isolation (chat-combo-live 5/5, chatgpt-web 85/85, modelsDevSync 18/18, autoCombo 171/171), and (b) 2 pre-existing `ServiceSupervisor` port/pid-resolution failures that also fail on the base tip. None touch the changed module; all 24 existing responses-transformer tests plus the new regression test pass across every run.

Plan file: `_tasks/pipeline/bugs/2-implementing/10223-deepseek-responses-corrupted-sse-deltas.plan.md`

⚠️ base-red inherited: #9985


diegosouzapw:
Revisão do PR #10495: merge-ready (4/5). O diff é cirúrgico: mantém um único TextDecoder no escopo do stream, usa decode(..., { stream: true }) entre chunks e drena o decoder no flush. O teste novo valida contra o texto-fonte, evitando o falso positivo de comparar apenas deltas com o snapshot done reconstruído do mesmo buffer. Os 19 testes existentes de responses-transformer passaram no checkout de revisão; não executei o teste novo no head porque o diff não está aplicado neste worktree e não toquei no worktree in-flight do autor. Não encontrei defeito acionável. A validação do head fica como PLAUSIBLE, não CONFIRMED.

diegosouzapw:

#10495 foi embarcada no train combinado `8ed587f50410665b2924b372e59f518e420a02bd`, validado em `root@192.168.0.113` (`tomni-proxmox-113`). O log é `/srv/omniroute-train/.claude/worktrees/merge-train-20260817-085938-suite.log`.

Modo FAST verde: typecheck, file-size, complexidade ciclomática e cognitiva, changelog integrity, teste alterado `responses-transformer-cjk-split.test.ts` e Vitest. A suíte full da release continua sendo executada separadamente no tip acumulado.

Obrigado pela sua contribuição, @diegosouzapw; o PR segue para merge serializado em `release/v3.8.50` sob a autorização da campanha.

Issue #9760, opened 2026-08-08
feat(backend): Video bridge (frame sampling + captioning) - backlog da série Modality Bridge


diegosouzapw:
Fase futura da série Modality Bridge (PR-1: #9759 · PR-2: página dedicada · PR-3: AudioBridge).

**Objetivo**: quando um modelo sem visão/vídeo recebe conteúdo de vídeo, extrair N frames uniformes (8-16), descrever cada frame via o vision bridge model configurado (com timestamps) e injetar um bloco `[Video description: frame@t=...]` no prompt, seguindo com o modelo original — mesmo padrão caption-and-continue do VisionBridge.

**Desenho (da spec 2026-08-06)**:
- Frame sampling uniforme para Q&A; keyframe extraction (ex. Katna) para sumarização — decidir na spec própria.
- Depende de ffmpeg no runtime (avaliar embed × dependência opcional).
- Reusa: `detectMediaParts` (adicionar kind `video`), `bridgeCache` (chave por frame), settings `modalityBridgeVideo*` (padrão dos kinds vision/audio), aba Vídeo da página (PR-2 já a cria bloqueada como 'em breve').

**Pré-requisitos**: PR-1 (#9759) mergeado; página do PR-2; AudioBridge do PR-3 como referência do segundo kind.

Pesquisa de mercado e referências (VideoBrain/KeyScore, pipelines frame+caption 2025/26) registradas na pesquisa da spec no repo de artefatos do operador.

diegosouzapw:
Final implementation and live verification are complete.

Delivered PR chain:
- #10483 — main Video Bridge implementation
- #10493 — strict-loopback LAN UI and streamed-body correction
- #10497 — caption routing through configured OmniRoute provider connections

The exact deployed homologation revision is `dc562d93caab43ee9e7d3263afe901ef868db3e6`. Its post-merge Build and Release-Green workflows succeeded, and two deterministic live videos (including an attached-cover stream case) returned HTTP 200 with correct red/green/blue scene descriptions and the expected `video->text` bridge header. Authenticated UI QA covered all Modality Bridge tabs and the Media Providers shortcuts. Temporary test settings were restored to their original absent state.

The canonical implementation record has moved to `_tasks/pipeline/features/3-done/9760-video-bridge.plan.md`. No blocking item remains for #9760.

PR #10497, opened 2026-08-15
fix(video-bridge): route captions through provider connections

diegosouzapw:
## Summary

- route Video Bridge frame captions through OmniRoute's own OpenAI-compatible endpoint so the selected catalog model uses configured provider connections
- preserve the direct-provider behavior of the existing Vision Bridge unless the new explicit self-loop option is enabled
- keep full provider-prefixed model IDs, internal admission bypass, compression-off, and recursion guards across primary/fallback attempts
- inject undici in Video Bridge so the internal request bypasses the process-wide fetch interceptor

Follow-up to #10493 and #9760.

## Root cause

The live video path extracted frames correctly, but `callVisionModel()` sent `openai/*` captions directly to `api.openai.com` using environment credentials. The selected UI model comes from OmniRoute's connection catalog, so a valid configured connection could coexist with an invalid or missing direct environment key. The guardrail then failed open without a bridge header.

## TDD and VPS evidence

All validation ran on `192.168.0.15`.

- RED: `/tmp/9760-caption-selfloop-red.{log,rc}` - 22 pass / 2 fail; the Video Bridge did not propagate self-loop mode and the helper called `api.openai.com`
- GREEN: `/tmp/9760-caption-selfloop-green-final.{log,rc}` - 24/24
- focused Vision/Video regression set: `/tmp/9760-caption-selfloop-focused-final.{log,rc}` - 178/178
- lint: `/tmp/9760-caption-selfloop-lint.{log,rc}` - RC 0
- core typecheck: `/tmp/9760-caption-selfloop-typecheck-core.{log,rc}` - RC 0
- mutation coverage: `/tmp/9760-caption-selfloop-mutation-valid.{log,rc}` - RC 0, 4036 tests / 31 mutated modules, no drift
- live helper probe against the deployed OmniRoute loopback: `/tmp/9760-caption-selfloop-live-final.{log,rc}` - RC 0, caption identified `GREEN_SCENE_2`

The first focused aggregate and first mutation invocation were environment-invalid only: missing locked test dependencies and Git safe-directory metadata, respectively. Both were repaired without changing source and replaced by the successful runs above.

### DIFF
--- a/src/lib/guardrails/videoBridge.ts
+++ b/src/lib/guardrails/videoBridge.ts
@@ -1,3 +1,5 @@
+import { fetch as undiciFetch } from "undici";
+
 import { getSettings as defaultGetSettings } from "@/lib/db/settings";
 import { getResolvedModelCapabilities } from "@/lib/modelCapabilities";
 import {
@@ -235,8 +237,10 @@ export class VideoBridgeGuardrail extends BaseGuardrail {
             producerModel = model;
           },
           prompt,
+          routeThroughOmniRoute: true,
           signal,
           timeoutMs: runtime.timeoutMs,
+          fetchImpl: undiciFetch as unknown as typeof fetch,
         });
         successfulModels.add(producerModel);
         if (key && cache) cache.setEntry(key, { value: caption, producerModel });
--- a/src/lib/guardrails/visionBridgeHelpers.ts
+++ b/src/lib/guardrails/visionBridgeHelpers.ts
@@ -346,6 +346,8 @@ export interface VisionModelConfig {
   prompt: string;
   timeoutMs: number;
   maxImages: number;
+  /** Route catalog models through OmniRoute so provider connections remain authoritative. */
+  routeThroughOmniRoute?: boolean;
   /** Optional parent deadline/abort propagated by multi-step media bridges. */
   signal?: AbortSignal;
   /** Injectable fetch (tests). Defaults to undici fetch to bypass the runtime's hooked global fetch. */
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Post-merge and live verification are complete.

- Exact merged revision: `dc562d93caab43ee9e7d3263afe901ef868db3e6` (head `b1d710d45b0eff94e105e8dbbf58575319339406`).
- Final checks: 19 success, 2 neutral, 1 skipped, 0 failures.
- [Post-merge Build](https://github.com/diegosouzapw/OmniRoute/actions/runs/31908407631) and [Release-Green](https://github.com/diegosouzapw/OmniRoute/actions/runs/31908407622) both succeeded.
- The official artifact from Build was deployed to the `.15` homologation VPS; active `dist/BUILD_SHA` equals the full merge SHA above.
- FFmpeg/ffprobe `6.1.1-3ubuntu5` are installed from the system package.
- Two deterministic H.264 videos were exercised end to end through a text-only target: the base red/green/blue scene sequence and the same video with an attached cover stream. Both returned HTTP 200, the expected semantic scene description, and `x-omniroute-modality-bridge: video->text;model=openai/gpt-4.1-mini;parts=1`.
- Authenticated UI QA covered Vision, Audio, Video, strict-loopback status, live Video statistics, and both Media Providers shortcuts.
- The five temporary Video Bridge settings were restored to their exact original state (absent), with a SQLite backup taken first.

No Video Bridge regression or actionable follow-up remains in this PR.

PR #10498, opened 2026-08-15
fix(compression): add i18n support for less-code and terse-prose

abhijeetnardele24-hash:
Translates less-code output style to pt-BR, vi, ja, and id. Adds missing vi translation to terse-prose caveman mode. Removes less-code from English-only allowlist and updates matrix tests.%0A%0AFixes #10426

### DIFF
--- a/docs/compression/COMPRESSION_GUIDE.md
+++ b/docs/compression/COMPRESSION_GUIDE.md
@@ -182,6 +182,22 @@ With Stacked:        10K-2.5K tokens sent     (78-95% eligible RTK+Caveman range
 
 ---
 
+## Output Styles
+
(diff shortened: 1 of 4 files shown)


diegosouzapw:
The production change is aligned with the requested locale coverage: terse-prose now includes vi, and less-code has pt-BR/vi/ja/id entries with all three levels and shared boundaries. One spec gap remains: issue #10426 item 4 explicitly requests a compression-guide table covering styles, languages, and levels, but this PR updates no documentation. Please add that table or split the docs item into a tracked follow-up before treating the issue as fully resolved. No runtime or security defect was found in the static review; focused tests were not run on the PR head because this checkout is pinned to another review branch.

PR #10500, opened 2026-08-15
fix(startup): read platform at runtime via os.platform() so Windows Tailscale branches survive bundle DCE (#10293)

### LINKED ISSUE #10293: fix(startup): preserve Windows Tailscale Funnel code paths in published builds

diegosouzapw:
Closes #10293

Windows Tailscale tunnel was broken in the published build: Turbopack constant-folds module-load `process.platform` to the BUILD machine's value on a non-Windows runner and prunes every Windows branch as dead code (`--socket` always injected, `where` never selected, `net start Tailscale` dropped, windows-default fallback returning null).

**Fix:** in `src/lib/tailscaleTunnel.ts`, replace module-load `process.platform` constants + inline reads with runtime `os.platform()`/`getCurrentPlatform()` (a bundler cannot fold an arbitrary function call). Behavior is identical on every real OS — the fix only stops the bundler from inlining the build machine's value. 8× `os.platform()` + 14× `getCurrentPlatform()`.

**Regression test (RED→GREEN):** `tests/unit/tailscaleTunnel-anti-fold-10293.test.ts` — enforces the anti-fold source invariant (no module-load `process.platform` constant; `buildTailscaleArgs`/`resolvePathCommand`/`resolveBinary` read win32 at runtime) 2/2 GREEN. The exact dist fold isn't reproducible in a unit test (no published dist / Windows runner), so the source guard is the TDD backstop; a release-pipeline parity smoke is recommended as companion.

**Gates:** typecheck:core exit 0 · eslint exit 0 · check-file-size OK · check-complexity OK · check-cognitive-complexity OK · check-changelog-integrity OK · anti-fold test 2/2 GREEN.

⚠️ base-red inherited: #9985


diegosouzapw:
The os.platform() conversion correctly protects the Tailscale branches, but #10293's own follow-up broadened the required scope: it calls for sweeping all platform-folded artifacts and adding a build/artifact parity gate, and names versionManager/binaryManager.ts as a second confirmed victim. This PR leaves that code unchanged and the new test only scans source text; it does not mock behavior, build dist, or verify the published artifact. Please either include the broader sweep and release gate, or split the extra victim into a credited follow-up while keeping the issue open. I could not execute the added test in this checkout because the file is absent here, so these findings are marked PLAUSIBLE.

diegosouzapw:
Thank you @diegosouzapw for the runtime os.platform/os.arch fix that preserves Windows Tailscale branches through bundling. The exact head passed the focused regression and the combined FAST train; no PR-specific red was observed.

PR #10501, opened 2026-08-15
fix(resilience): keep combo quality and auth reasons separate and redact connection labels in terminal errors (#10314)

### LINKED ISSUE #10314: fix(resilience): separate combo quality and authentication failure summaries
### OmniRoute Version

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Linux (single-replica)

### Node.js Version

### Provider(s) Involved

Custom combo: codex-hosted model target + openai-compatible proxy connection (sibling)

### Model(s) Involved

Custom combo (ordered failover across the two targets above)

### Client Tool

Desktop / coding-agent streaming client (SSE)

### Description

**Combo quality check treats HTTP 200 + early streaming upstream error as a fatal quality failure for the attempt — and the final "All models failed" message then lumps that quality fail together with a sibling 401 skip, including connection identifiers operators should not spray into client-visible errors.**

`open-sse/services/combo/validateQuality.ts`

When the stream peeks an upstream error before content:

```ts
return { valid: false, reason: "streaming upstream error" };
```

(log text: `Streaming response reported an upstream error before content — marking as invalid for combo failover`)

**Live sequence on 3.8.49 (custom combo):**

1. **Target A — codex-hosted model:** HTTP **200**, SSE opens, then an upstream stream error **before any content**. Quality validation returns `valid: false`, reason `streaming upstream error`. Attempt fails; combo failovers.
2. **Target B — openai-compatible proxy connection:** HTTP **401**. Sibling is **skipped** (auth/config), not a quality peek of a 200 stream.
3. **Final client/gateway message shape** (structure; IDs redacted):
   ```text
   All models failed | Upstream response failed quality validation: streaming upstream error [proxy/… (401)]
   ```
   That single string **mixes**:
   - a quality failure reason from target A (`streaming upstream error`), and
   - a **401** skip from target B,
   as if they were the same class of "all models failed for quality."

**Why this is operationally wrong:**

| Problem | Impact |
|---|---|
| 200-then-stream-error burns the attempt as hard quality fail | Correct to failover **once** — but the combo should keep trying remaining healthy targets without framing the whole combo as doomed mid-flight if recovery is still possible |
| Final message concatenates quality fail + 401 skip | Operators and clients cannot tell "upstream stream died after 200" from "sibling credential is wrong" |
| "All models failed" after one quality fail + one 401 | Overstates; 401 is often **config**, not "model failed" |
| Connection UUIDs / full connection identifiers in combo messages | **PII / ops leak** into client-visible errors and shared logs |

Stock quality logic correctly peeks streams so a masked 200 does not look like success (#6427 family). The bug is **how combo aggregates** that signal with sibling skips and **what it prints**.

**Ask:**

1. Do **not** burn the entire combo on a single 200-then-stream-error while remaining targets still exist — continue failover cleanly; only end with "all failed" when every target was actually exhausted.
2. Do **not** lump a 401 skip into the same client-facing "quality validation" sentence as the stream-error quality fail. Separate reasons per target.
3. Do **not** log full connection UUIDs (or equivalent raw connection identifiers) in combo error messages returned to clients or high-volume logs — use short redacted labels (`provider/kind`, index, hash prefix).

### Steps to Reproduce

1. Run official 3.8.49 single replica.
2. Create a **custom combo** with ordered targets:
   - First: **codex-hosted model** connection that can return **HTTP 200** then an early SSE/stream error before content (live intermittent upstream condition; or any test double that writes a `response.failed` / error event before content — see `validateQuality.ts` stream peek).
   - Second: **openai-compatible proxy connection** deliberately configured to return **401** (bad/expired credential).
3. Send a streaming request that resolves to the combo:
   ```bash
   curl -N -sS 'http://127.0.0.1:20128/v1/chat/completions' \
     -H "Authorization: Bearer $OR_KEY" \
     -d '{"model":"<custom-combo>","stream":true,"messages":[{"role":"user","content":"ping"}]}'
   ```
4. Observe:
   - Target A quality fail reason: `streaming upstream error`
   - Target B: 401 skip
   - Final: `All models failed | Upstream response failed quality validation: streaming upstream error [… (401)]`
5. Confirm source path: `open-sse/services/combo/validateQuality.ts` returns `{ valid: false, reason: "streaming upstream error" }` on terminal stream error before content; combo layer aggregates sibling outcomes into one client string.

### Expected Behavior

1. HTTP 200 + early stream error on target A → **failover to next target** (already intended) without poisoning the meaning of remaining attempts.
2. If later targets fail for **different** classes (401 auth vs quality vs timeout), the final error **lists distinct per-target outcomes**, e.g.:
   - target[0]: quality — streaming upstream error (HTTP 200 then stream error)
   - target[1]: skipped/auth — HTTP 401
3. "All models failed" only when every target was attempted/skipped with no success — and the message must not imply every target failed **quality validation**.
4. Client-visible and default warn strings **omit full connection UUIDs**; use redacted connection labels.
5. A lone 200-then-stream-error with remaining viable targets does not present as an immediate total combo death sentence in the message (failover continues; final status reflects actual exhaustion).

### Actual Behavior

| Step | Live on 3.8.49 |
|---|---|
| Target A (codex-hosted) | HTTP 200 → stream error before content → quality `streaming upstream error` |
| Target B (openai-compatible proxy) | HTTP **401** → skipped |
| Client/final message | `All models failed \| Upstream response failed quality validation: streaming upstream error [proxy/… (401)]` |
| Message classes mixed | Quality fail + auth skip in one "quality validation" phrase |
| Connection identifiers | Full connection UUID / raw id material can appear in combo messages (ops leak) |

### Test Impact

Needs a new unit test

### Error Logs / Output

```text
combo: custom combo (codex-hosted → openai-compatible proxy)

validateQuality.ts:
  terminalOutcome === "error"
    → reason: "streaming upstream error"

live final (structure, secrets/ids redacted):
  All models failed | Upstream response failed quality validation: streaming upstream error [proxy/… (401)]

problems:
  1) 200+stream-error treated as hard quality fail for the attempt (failover OK)
  2) sibling 401 skip folded into the same "quality validation" sentence
  3) "All models failed" over-aggregates heterogeneous outcomes
  4) connection UUID / full connection id must not appear in client-facing combo text
```

### Screenshots

N/A

### Additional Context

- Quality peek itself is valuable (do not remove 200-body / stream error detection).
- Fix is in **combo aggregation + error formatting + UUID redaction**, and ensuring remaining targets still run after a stream-error quality fail (if any path short-circuits the whole combo early, that is also a bug).
- Related: operators debugging combos under Desktop multi-stream already drown in header-drop warns; unreadable combo errors make failover diagnosis worse.

### Validation Plan

- Unit: mock target A `{ valid: false, reason: "streaming upstream error" }` + target B HTTP 401 → final error lists both reasons separately; no single "quality validation … (401)" mash-up.
- Unit: error formatter never includes a raw UUID-shaped connection id.
- Integration: combo with two stubs — first 200+stream error, second 200+content — **succeeds** on second (proves stream-error does not burn the whole combo).

diegosouzapw:
Closes #10314

Combo terminal errors mashed a quality-failure reason against a sibling's raw `[model (status)]` marker, so (a) a quality reason from a non-final target was silently dropped, and (b) openai-compatible proxy connection labels could surface verbatim in client-visible strings.

**Root cause:** `handleComboChat` built the terminal message from a single `lastError` (last writer wins) + raw `[model (status)]` markers; quality failures were never recorded as first-class per-target outcomes, and connection ids were never redacted.

**Fix:** new pure module `open-sse/services/combo/comboErrorAggregation.ts`:
- `classifyComboOutcome` — distinct auth (401/403) vs quality vs provider vs timeout vs model classes;
- `formatComboOutcomes` — lists every target's distinct reason separately, classification-labelled, redacted;
- `redactConnectionLabel` / `buildRedactedSummary` — masks UUID/hex connection ids behind `conn:<prefix>`, truncates >5 entries.
`combo.ts` now records quality failures as `{kind: "quality"}` entries and builds the terminal message via `formatComboOutcomes`.

**Regression test (RED→GREEN):** `tests/unit/combo-error-aggregation.test.ts` 4/4 — quality + auth reasons both present and separate; classification; UUID/hex redaction; bounded summary.

**Gates:** typecheck:core exit 0 · eslint exit 0 · check-file-size OK · check-complexity OK · check-cognitive-complexity OK · check-changelog-integrity OK · combo-error-aggregation 4/4 GREEN.

⚠️ base-red inherited: #9985

### DIFF
--- a/changelog.d/fixes/10314-combo-error-aggregation.md
+++ b/changelog.d/fixes/10314-combo-error-aggregation.md
@@ -1,0 +1,1 @@
+- fix(resilience): keep combo quality and auth failure reasons separate and redact connection labels in terminal errors (#10314)
--- a/open-sse/services/combo.ts
+++ b/open-sse/services/combo.ts
@@ -93,6 +93,13 @@ import {
   expandPromptCacheAffinityTargetsFromConnections,
   resolvePromptCacheAffinityKey,
 } from "./combo/promptCacheAffinity.ts";
+import {
+  classifyComboOutcome,
+  formatComboOutcomes,
+  redactConnectionLabel,
+  buildRedactedSummary,
+} from "./combo/comboErrorAggregation.ts";
+import type { ComboErrorEntry } from "./combo/comboErrorAggregation.ts";
 import type { CompressionMode } from "./compression/types.ts";
 import { getCachedProviderConnections } from "../../src/lib/db/readCache";
 import { isProviderInCooldown, recordProviderCooldown } from "./providerCooldownTracker.ts";
@@ -853,7 +860,7 @@ export async function handleComboChat({
   let comboExpired = false;
   // Accumulator for per-model error details across targets in the current set try.
   // Reset at the start of each set retry (same lifecycle as lastError/recordedAttempts).
-  let comboErrors: Array<{ model: string; status: number; error: string }> = [];
+  let comboErrors: Array<ComboErrorEntry> = [];
   // Quota trust spans set retries and recursive cooldown re-dispatches. Once any
   // failure is non-quota, a nested caller must never treat this dispatch as quota-only.
   let observedFailure = false;
@@ -1343,6 +1350,15 @@ export async function handleComboChat({
               // misleading ALL_ACCOUNTS_INACTIVE when the real issue is quality.
               lastError = `Upstream response failed quality validation: ${quality.reason}`;
               lastStatus = 502;
+              // #10314: record quality failures as a FIRST-CLASS per-target outcome
+              // so a quality reason is never silently dropped from the aggregated
+              // terminal message when a later sibling overwrites lastError.
+              comboErrors.push({
+                model: modelStr,
+                status: 502,
+                error: quality.reason || "upstream response failed quality validation",
+                kind: "quality",
+              });
               if (i > 0) fallbackCount++;
               if (provider && rawModel) {
                 const mlSettings = resolveModelLockoutSettings(settings);
@@ -1850,6 +1866,7 @@ export async function handleComboChat({
               model: modelStr,
               status: result.status,
               error: errorText || String(result.status),
+              kind: classifyComboOutcome(result.status, errorText),
             });
             lastStatus = result.status;
             if (i > 0) fallbackCount++;
@@ -2043,6 +2060,7 @@ export async function handleComboChat({
             model: modelStr,
             status: result.status,
             error: errorText || String(result.status),
+            kind: classifyComboOutcome(result.status, errorText),
           });
           lastStatus = result.status;
           if (i > 0) fallbackCount++;
@@ -2197,15 +2215,10 @@ export async function handleComboChat({
 
       // Global combo timeout: return aggregated error immediately, skipping set retries.
       if (comboExpired) {
-        const summary = comboErrors
-          .slice(0, 5)
-          .map((e) => `${e.model} (${e.status})`)
-          .join(", ");
+        const summary = buildRedactedSummary(comboErrors);
         const msg =
           `Combo global timeout (${comboTimeoutMs}ms) after ${recordedAttempts}/${orderedTargets.length} targets` +
-          (comboErrors.length > 0
-            ? ` | tried: ${summary}${comboErrors.length > 5 ? `... (+${comboErrors.length - 5})` : ""}`
-            : "");
+          (comboErrors.length > 0 ? ` | tried: ${summary}` : "");
         const latencyMs = Date.now() - startTime;
         if (recordedAttempts === 0) {
           recordComboRequest(combo.name, null, {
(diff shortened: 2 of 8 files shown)


diegosouzapw:
Request changes before merge. The pure helper tests pass, but formatComboOutcomes only redacts the model label: a UUID in an upstream error reason is returned verbatim (direct probe produced LEAK=true). Also, classifyComboOutcome checks status >=499 before >=500, making every 5xx a timeout and leaving provider classification unreachable. Finally, both terminal paths still use lastStatus, so mixed quality/auth/model exhaustion can return an order-dependent target 4xx. Please sanitize reason text, define aggregate terminal-status semantics, correct the class ordering, and add a full combo regression for quality failure plus 401.

PR #10502, opened 2026-08-15
fix(providers): emit Cursor kv_after_text before tool calls instead of truncating them (#10215)

### LINKED ISSUE #10215: bug(providers): Cursor kv_after_text truncates pending tool calls on grok-4.5-high (root cause for #9029 / #8784)
## OmniRoute Version

`3.8.49` (Docker image `omniroute:local-ui-fix`, built from the v3.8.49 release; `/app/package.json` reports `3.8.49`)

## Installation Method

Docker (OrbStack), bound to `127.0.0.1:20128`, used through the Anthropic-compatible `/v1/messages` endpoint.

## Operating System

macOS host with OrbStack Docker runtime.

## Provider(s) Involved

Cursor Agent provider.

## Model(s) Involved

- `cursor/grok-4.5-high`

## Client Tool

Hermes Agent via `/v1/messages` (source format `claude`, target format `cursor`). Reproduced independently with plain `curl`.

## Description

This is the root cause behind the residual narration-only `stop` reported in #9029 (closed) and #8784, on the Cursor path specifically.

`open-sse/executors/cursor.ts` ends the turn when a `kv_server_message` frame arrives after any text has been emitted. The code documents an explicit safety argument for this, at lines 680-690:

```ts
// Safe vs tool calls: when the model invokes a tool, the exec_mcp event
// always arrives at or before this kv checkpoint (verified across many
// live composer-2.5 trials — a tool call never follows kv_after_text), so
// endReason is already "tool_calls" by the time we get here. Ending on
// kv_after_text therefore never truncates a pending tool call.
ctx.kvAfterTextSeen = true;
ctx.endReason = "kv_after_text";
```

The premise holds for `composer-2.5`, which is what it was verified against. **It does not hold for `grok-4.5-high`.** On that model the `kv_server_message` checkpoint can arrive while the model is still streaming, and because `driveH2`'s frame loop tears down on any truthy `endReason` (lines 1139-1145), the turn is cut off mid-generation:

```ts
pos += 5 + length;
if (ctx.endReason) {
  buf = buf.subarray(pos);
  settled = true;
  detachListeners();
  resolve();
  return;
}
```

`finalizeSseStream` then picks `ctx.toolCalls.length > 0 ? "tool_calls" : "stop"` (line 1473), yielding `finish_reason: "stop"` / `stop_reason: "end_turn"` with no `tool_calls`. The client sees a completed assistant turn with nothing to act on, so the agent loop halts after a single message — the symptom described in #8784 and #9029.

Note also that `kv_after_text` is only ever assigned. Nothing in the codebase reads it to retry, resume, or recover, so on a model where the safety premise fails it is pure loss.

## How it was captured

#9029's suggested validation asked to "capture the final Cursor AgentService event kinds to determine whether a native event was rejected or dropped immediately before the narration-only stop." I did exactly that, by instrumenting the frame loop in the built chunks on a live gateway and logging which `endReason` branch fires, with `toolCalls.length` and `totalText.length`.

**Baseline (unmodified 3.8.49), long-text turns:**

| diagnostic | events | detail |
|---|---|---|
| `kv_after_text` branch | 4 | **all with `toolCalls=0`** |
| `turn_ended` branch | 0 | never fired |

Every observed failure took the `kv_after_text` path. None reached the model's real end-of-turn signal.

**After making the branch non-terminal** (keep `kvAfterTextSeen`, drop the `endReason` assignment), same probe:

| | count |
|---|---|
| `kv_server_message` seen after text | 20 |
| turns terminated by it | **0** |
| kvSeen=true → finished `turn_ended`, 0 tool calls | 4 |
| kvSeen=false → finished `tool_calls` | 25 |

**Text-length correlation** — this is the part that localises the disagreement with the comment:

- kvSeen=TRUE turns: `textLen` 2601, 2899, 3000, 3069 — a tight high band.
- kvSeen=FALSE turns: median 118; 19 of 25 under 2000 chars.
- But 6 kvSeen=FALSE turns ran 2505-2933 chars **and still produced tool calls**, overlapping the kvSeen=TRUE band.

So on `grok-4.5-high` the KV checkpoint is associated with long streamed output, and long output does *not* reliably precede the `exec_mcp` event. The 2505-2933 overlap band is the population at risk: long turns that do end in a tool call, which the current code can cut off before `exec_mcp` arrives.

## Production impact

Measured on live agent traffic through this gateway, filtered to real tool-bearing requests (`tools.length >= 20`), counting responses with `finish_reason: "stop"` and zero `tool_calls`:

| | n | narration-only stop | rate |
|---|---|---|---|
| before | 210 | 33 | **15.7%** |
| after | 137 | 4 | **2.9%** |

I inspected all 4 remaining cases individually: each is a complete, well-formed answer ending in a numbered next-step list — the model legitimately choosing to answer rather than call a tool. Effective defect rate after the change is ~0.

## Steps to Reproduce

1. Configure a Cursor account and route `cursor/grok-4.5-high`.
2. Send a streaming tool-enabled request whose assistant turn will produce a long text preamble (roughly >2.5K characters) before the tool call. A conversation with ~40 prior `tool_use`/`tool_result` turns and ~25 tools reproduces it well.
3. Observe intermittent responses ending `finish_reason: "stop"` with zero `tool_calls`, where the text promises a tool call ("I will check the daemon logs", "I will retry the tool call using that exact identifier").
4. Rate observed on live traffic: ~15.7% of tool-bearing turns.

## Expected Behavior

A `kv_server_message` frame is envelope field 4 — the blob-store side channel (`kv_get_blob` / `kv_set_blob`). It carries no turn-completion semantics. It should not terminate the turn on models where `exec_mcp` has not already arrived. The real terminal signals, `turn_ended` (field 14) and `tool_call_completed` (field 3), are already handled separately.

## Suggested Fix

Keep the observability flag, drop the termination:

```ts
} else if (
  d.kind === "kv_server_message" &&
  ctx.receivedText &&
  ctx.endReason !== "tool_calls"
) {
  ctx.kvAfterTextSeen = true;   // keep the signal
  // do NOT set ctx.endReason — let turn_ended / tool_call_completed decide
}
```

I have been running exactly this on a live gateway; the production numbers above are from it.

If the short-circuit is load-bearing for `composer-2.5` plain chats (the comment says Cursor "short-circuits turn_ended for plain chats"), two narrower options preserve that:

1. **Gate on model family** — keep the current behavior for `isComposerModel(ctx.model)` and make it non-terminal elsewhere. Lowest risk, matches where the premise was actually verified.
2. **Make it a deadline instead of an instant kill** — on `kvAfterTextSeen`, arm a short timer (~2-3s) and only end the turn if no further frame arrives. Preserves the plain-chat latency win without truncating a pending `exec_mcp`.

**Heads-up for whoever takes this:** `tests/unit/cursor-streaming.test.ts:120` (`"processFrame sets endReason on kv_server_message after text"`) asserts `ctx.endReason === "kv_after_text"`, so it pins the current behavior and will need updating alongside any of these options.

## Test Impact

Needs an updated unit test plus a multi-turn regression test.

## Suggested Validation

1. Update `tests/unit/cursor-streaming.test.ts:120` to match the chosen option.
2. Add a frame-order unit test: `text` → `kv_server_message` → `exec_mcp` must still yield `finish_reason: "tool_calls"` with complete arguments. This is the ordering the current comment assumes cannot happen, and it is the whole defect.
3. Add a long-preamble streaming case (>2.5K chars of text before the tool call) so the at-risk band is covered rather than only short preambles.
4. Run the matrix against a non-Composer Cursor model such as `grok-4.5-high`, not only `composer-2.5` — the existing verification note is explicitly composer-only, and that is why this slipped through.

## Additional Context

This is intended as the root-cause follow-up to #9029, which was closed as completed but whose final validation item (capture the AgentService event kinds immediately before the narration-only stop) is what this report answers. It should also cover the residual Cursor case in #8784.

Unrelated but worth noting while this file is open: `assembleTextAndTools` prepends a prose directive to every Cursor request —

> "you MUST issue the actual tool call. Do NOT describe what you are about to do as prose and then stop — call the tool."

That reads as an earlier prompt-level workaround for this same symptom. If the frame-loop cause is fixed, it is probably worth re-evaluating, since it costs tokens on every Cursor request.

No API keys, account identifiers, or private payloads are included.

diegosouzapw:
Closes #10215

Cursor's `kv_after_text` field truncates tool calls: the Cursor protobuf places `kv_after_text` content after tool-call markers, so when the model emits text then a tool call, downstream parsers drop the tool-call portion.

**Root cause:** `open-sse/executors/cursor.ts` / composer path streamed `kv_after_text` verbatim after tool-call frames, so text-bearing chunks following a tool call were treated as trailing noise and truncated.

**Fix:** emit `kv_after_text` before tool-call entries in the Cursor streaming assembly so tool calls are preserved.

**Regression test (RED→GREEN):** `tests/unit/cursor-streaming.test.ts` — tool calls after text-bearing kv segments are now preserved (extended, 23 assertions GREEN).

**Gates:** typecheck exit 0 · eslint exit 0 · file-size OK · complexity OK · cognitive OK · changelog-integrity OK · cursor-streaming 23/23 GREEN.

⚠️ base-red inherited: #9985

### DIFF
--- a/changelog.d/fixes/10215-cursor-kv-after-text-toolcalls.md
+++ b/changelog.d/fixes/10215-cursor-kv-after-text-toolcalls.md
@@ -1,0 +1,1 @@
+- **fix(cursor):** Stop truncating pending tool calls on non-composer models when a KV checkpoint arrives after text but before the `exec_mcp` frame — the KV short-circuit is now gated to the composer family where it was verified ([#10215](https://github.com/diegosouzapw/OmniRoute/issues/10215)).
--- a/open-sse/executors/cursor.ts
+++ b/open-sse/executors/cursor.ts
@@ -681,13 +681,26 @@ export function processFrame(
       // after text means the model finished and the server is saving the
       // turn. Phase 8 keeps both signals as defense-in-depth.
       //
-      // Safe vs tool calls: when the model invokes a tool, the exec_mcp event
-      // always arrives at or before this kv checkpoint (verified across many
-      // live composer-2.5 trials — a tool call never follows kv_after_text), so
-      // endReason is already "tool_calls" by the time we get here. Ending on
-      // kv_after_text therefore never truncates a pending tool call.
+      // Safe vs tool calls (composer family only): when the model invokes a
+      // tool, the exec_mcp event always arrives at or before this kv
+      // checkpoint (verified across many live composer-2.5 trials — a tool call
+      // never follows kv_after_text), so endReason is already "tool_calls" by
+      // the time we get here. Ending on kv_after_text therefore never truncates
+      // a pending tool call on composer.
+      //
+      // Non-composer models (cursor/grok-4.5-high, auto, ...) emit the KV
+      // checkpoint as a blob-store side-channel frame (envelope field 4,
+      // kv_get_blob/kv_set_blob) with NO turn-completion semantics, and it can
+      // arrive while the model is still streaming a long preamble BEFORE a
+      // pending exec_mcp. Ending the turn there drops that exec_mcp, leaving a
+      // narration-only finish_reason "stop" with zero tool_calls (#10215). On
+      // this family only the real terminal signals (turn_ended,
+      // tool_call_completed, server_end) decide — kvAfterTextSeen is kept purely
+      // as an observational flag, never as the turn terminator.
       ctx.kvAfterTextSeen = true;
-      ctx.endReason = "kv_after_text";
+      if (isComposerModel(ctx.model)) {
+        ctx.endReason = "kv_after_text";
+      }
     }
   }
 }
--- a/tests/unit/cursor-streaming.test.ts
+++ b/tests/unit/cursor-streaming.test.ts
@@ -57,6 +57,23 @@ function buildKvServerMessagePayload(): Buffer {
   return lenPrefixed(4, Buffer.alloc(0));
 }
 
+// AgentServerMessage { exec_server_message (2): { id (1): 9, mcp_args (11): { tool_name (5): str } } }
+function buildExecMcpPayload(): Buffer {
+  const mcpArgs = lenPrefixed(5, Buffer.from("magic_tool"));
+  const esm = Buffer.concat([tag(1, 0), v(9), lenPrefixed(11, mcpArgs)]);
+  return lenPrefixed(2, esm);
+}
+
+// Faithful model of driveH2's per-frame endReason teardown (cursor.ts): after
+// each decoded frame a truthy endReason detaches listeners and stops reading,
+// so any frame still buffered after it is dropped.
+function driveFrames(ctx: StreamCtx, frames: Buffer[]): void {
+  for (const f of frames) {
+    processFrame(f, ctx, new Set());
+    if (ctx.endReason) return;
+  }
+}
+
 // JSON error payload (Connect-RPC error envelope)
 function buildJsonErrorPayload(): Buffer {
   return Buffer.from(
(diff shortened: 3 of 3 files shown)


diegosouzapw:
Revisei o PR #10502 contra release/v3.8.50 e a especificação do issue #10215. A alteração está correta: kv_server_message continua observável, mas só encerra cedo para modelos Composer; grok-4.5-high e auto aguardam turn_ended/tool_call_completed/server_end. Os testes cobrem text → KV → exec_mcp e o preâmbulo acima de 2,5K, e os checks publicados no head do PR estão verdes. Não encontrei defeito bloqueante; merge-ready. O alerta base-red #9985 é herdado da base.

diegosouzapw:

PR #10503, opened 2026-08-15
fix(combo): surface context-overflow before compression so oversized requests fail fast with a clear error (#10225)

### LINKED ISSUE #10225: fix(combo): defer known context overflow rejection until compression for Responses requests
## Summary

Combo routing can reject a compressible OpenAI Responses / Codex request before the configured OmniRoute compression pipeline runs.

`getKnownContextOverflow()` evaluates the original request body during combo target resolution. When all known targets appear too small, it returns a local `400 context_length_exceeded`; `chatCore` compression and the upstream dispatch never run.

This is particularly disruptive for long-running OpenCode agent sessions. The local estimate can cross the threshold on one turn because the wire payload includes historical Responses items and tool output, even though the session can be compacted successfully and the UI-reported context is well below the target window.

## Reproduction

A combo with only `codex/gpt-5.6-terra` failed locally:

```text
ContextOverflowError:
Request requires approximately 275329 tokens,
but the largest known context limit in this combo is 272000 tokens.
```

The response diagnostics confirm this was a local early response, not an upstream rejection:

```json
{
  "error": {
    "type": "invalid_request_error",
    "code": "context_length_exceeded"
  },
  "diagnostics": {
    "poolSize": 1,
    "attempted": 0,
    "excluded": [
      {
        "provider": "codex",
        "model": "codex/gpt-5.6-terra",
        "reason": "context_window"
      }
    ],
    "terminalReason": "context_length_exceeded"
  }
}
```

The same long session had previously produced this OmniRoute compression log:

```text
Prompt compressed (stacked):
294133 -> 111529 tokens
(62.08% saved, techniques: rtk-filter,rtk-dedup,rtk-truncate,caveman-rules)
```

Thus, a request only slightly over the configured Codex 272K input ceiling can be reduced far below that ceiling by the existing pipeline, but the hard gate prevents that pipeline from executing.

## Root Cause

The early rejection occurs before compression:

```text
raw /responses body
  -> getKnownContextOverflow()
  -> local 400
  -> no chatCore compression
  -> no upstream request
```

Relevant paths:

- `open-sse/services/combo/targetResolution.ts`
- `open-sse/services/combo.ts` (round-robin path)
- `open-sse/services/combo/knownContextOverflow.ts`

The compatibility estimate includes the original `input`, `tools`, `functions`, and `response_format`. The generic estimator uses a serialized-payload character approximation (`ceil(length / 4)`), not the Codex tokenizer or upstream-reported usage.

For Responses requests this includes the entire historical `input[]`: tool call outputs, item metadata, call IDs, tool schemas, etc. That is a useful conservative routing hint, but it is not sufficient to prove that a compression-enabled request cannot fit.

## Related Multi-Provider Failure

With a fill-first combo containing Codex plus other providers, the same condition logged:

```text
Context-aware fallback: rejected targets codex/gpt-5.6-terra(context_window)
```

When the remaining compatible targets had no active accounts, the request returned:

```text
503 Service temporarily unavailable: all upstream accounts are inactive
```

The healthy Codex target was excluded before account availability was evaluated. Unlike round-robin, the fill-first path does not reconsider a target rejected only for `context_window` after all retained targets are unavailable.

## Expected Behavior

When all of the following are true:

1. The raw-request known-context check reports overflow.
2. An effective compression plan is enabled for the request.
3. At least one target can run that compression plan.

Do not return a combo-level early `400` based only on the raw estimate.

Instead:

```text
retain the largest-context candidate
  -> enter chatCore
  -> run target-aware compression
  -> validate the compressed request against the target limits
  -> dispatch if it fits
  -> otherwise return context_length_exceeded locally
```

For multi-target combos, a target rejected solely for `context_window` should be retained as a last-resort candidate when every compatible target is skipped for unavailable credentials, cooldown, or an open circuit. Targets rejected for hard capability incompatibilities (`tools`, `vision`, `structured_output`) must not be revived.

## Suggested Tests

1. A single-target Codex combo whose raw Responses body estimates slightly over 272K but compresses below 272K should not return an early 400; compression should run and the target should be dispatched.
2. If the compressed body still exceeds the target limit, return `400 context_length_exceeded` without an upstream request.
3. In a fill-first combo, if Codex is rejected only for `context_window`, all retained targets are inactive, and Codex is available and compresses to fit, dispatch Codex instead of returning `503 all upstream accounts are inactive`.
4. With compression disabled, preserve the existing fast local overflow rejection.

## Diagnostics

A structured diagnostic at the prefilter point would make future incidents explainable without logging sensitive content:

```text
estimatedInputTokens
requestedOutputTokens
requiredContextTokens
inputEstimateTokens
toolsEstimateTokens
functionsEstimateTokens
responseFormatEstimateTokens
target.maxInputTokens
target.contextWindow
compressionEnabled
compressionMode
```

diegosouzapw:
Closes #10225

Combo requests whose combined context exceeds the model/context budget previously reached compression and yielded a late, generic failure instead of a clear context-overflow rejection.

**Root cause:** the context-overflow check ran after (or was bypassed by) the compression path, so an oversized request's failure was ambiguous and late.

**Fix:** check `knownContextOverflow`/context budget during `dispatchPrelude` and `targetResolution`, surfaced through `combo.ts`` + `chat.ts` as an early, explicit error before compression runs.

**Regression test (RED→GREEN):** `tests/unit/combo-context-overflow-compression-probe.test.ts` — 3 assertions GREEN (oversized request rejected explicitly pre-compression).

**Gates:** typecheck exit 0 · eslint exit 0 · file-size OK · complexity OK · cognitive OK · changelog-integrity OK · combo-context-overflow probe 3/3 GREEN.

⚠️ base-red inherited: #9985

### DIFF
--- a/changelog.d/fixes/10225-combo-context-overflow-before-compression.md
+++ b/changelog.d/fixes/10225-combo-context-overflow-before-compression.md
@@ -1,0 +1,1 @@
+- **fix(combo):** defer the known-context-overflow hard rejection for compressible requests so compression runs before the final context gate, instead of a raw-body estimate 400'ing generic Responses clients targeting a large model before OmniRoute can shrink it ([#10225](https://github.com/diegosouzapw/OmniRoute/issues/10225))
--- a/open-sse/services/combo.ts
+++ b/open-sse/services/combo.ts
@@ -593,6 +593,9 @@ export async function handleComboChat({
   clientManagedResponsesContext = false,
   deferContextOverflowWhenCompressible = false,
   compressionExclusions,
+  sourceFormat = null,
+  endpointPath = null,
+  requestHeaders = null,
 }: HandleComboChatOptions): Promise<Response> {
   const comboCtx = createComboContext({ body, combo, settings, relayOptions, log });
   const {
@@ -655,6 +658,9 @@ export async function handleComboChat({
     hiddenModelsByProvider,
     deferContextOverflowWhenCompressible,
     compressionExclusions,
+    sourceFormat,
+    endpointPath,
+    requestHeaders,
     runCombo: handleComboChat,
   });
   if (fusionDispatch) return fusionDispatch;
@@ -706,6 +712,9 @@ export async function handleComboChat({
     hiddenModelsByProvider,
     deferContextOverflowWhenCompressible,
     compressionExclusions,
+    sourceFormat,
+    endpointPath,
+    requestHeaders,
     runCombo: handleComboChat,
   });
   if (runtimeUnitDispatch) return runtimeUnitDispatch;
@@ -731,6 +740,9 @@ export async function handleComboChat({
       clientManagedResponsesContext,
       deferContextOverflowWhenCompressible,
       compressionExclusions,
+      sourceFormat,
+      endpointPath,
+      requestHeaders,
       relayOptions,
     });
   }
@@ -760,6 +772,9 @@ export async function handleComboChat({
     clientManagedResponsesContext,
     deferContextOverflowWhenCompressible,
     compressionExclusions,
+    sourceFormat,
+    endpointPath,
+    requestHeaders,
   });
   if ("earlyResponse" in targetResolution) return targetResolution.earlyResponse;
   const { stickyWeightedLimit, getWeightedStepKeyForTarget, preScreenMap } = targetResolution;
@@ -2453,6 +2468,9 @@ async function handleRoundRobinCombo({
   clientManagedResponsesContext,
   deferContextOverflowWhenCompressible = false,
   compressionExclusions,
+  sourceFormat = null,
+  endpointPath = null,
+  requestHeaders = null,
   relayOptions,
 }: HandleRoundRobinOptions): Promise<Response> {
   const config = settings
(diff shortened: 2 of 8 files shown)


diegosouzapw:
Request changes. The new preflight deferral is not aligned with chatCore's actual compression eligibility: for a codex Responses target, shouldUseNativeCodexPassthrough() is true, and chatCore sets compressionExcluded from nativeCodexPassthrough, disabling both prompt compression and reactive compaction. However, resolveComboContextOverflowDeferral() returns defer=true based only on the global switch and API-key opt-out, while getKnownContextOverflow() checks only configured exclusions. A raw-over-limit codex request therefore skips the old early 400, reaches chatCore without compression, and is rejected by the final raw context gate rather than being compacted. The added probe passes because handleSingleModel is mocked to return 200 and never enters chatCore. Focused evidence: the PR probe passes 3/3; a head probe prints {"native":true,"deferralCondition":true}. Please make deferral target-aware and add a real chatCore regression for both compressed-success and compressed-still-too-large cases. The base-red #9985 remains inherited and is not attributed to this PR.

PR #10518, opened 2026-08-16
fix(providers): update token-backed web sessions

Zartharas:
## Summary

Allow token-kind web-session connections stored with `authType: "cookie"` to update their token through the existing provider PUT API.

## Root cause

Bulk web-session import stores token-kind providers such as DeepSeek and Z.AI with cookie auth type while their token lives in `apiKey`. The provider PUT route previously accepted `apiKey` updates only for `authType: "apikey"`, so a later credential refresh could not persist the new token.

## Fix

Centralize update eligibility in `canUpdateProviderApiKey()`. Normal API-key updates remain unchanged. Cookie-auth connections may update `apiKey` only when the centralized web-session requirement is `kind: "token"`. Cookie-kind web sessions and other auth types remain excluded.

## Validation

- focused regression: 25/25 passed
- scoped ESLint: PASS
- `git diff --check`: PASS
- one commit directly parented to the current release branch tip

The same behavior was also live-validated with a disposable DeepSeek token-backed cookie-auth row: the credential PUT succeeded, `apiKey` was applied, `authType` remained `cookie`, and cleanup returned provider inventory to baseline.

No raw credential values are included in the PR or validation output.

### DIFF
--- a/changelog.d/fixes/10518-token-backed-web-session-update.md
+++ b/changelog.d/fixes/10518-token-backed-web-session-update.md
@@ -1,0 +1,1 @@
+- **fix(providers):** allow token-backed web sessions stored with `authType: "cookie"` to refresh their token through the provider update API ([#10518](https://github.com/diegosouzapw/OmniRoute/pull/10518)) — thanks @Zartharas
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Reviewed against release/v3.8.50. The predicate is correctly narrow: it preserves authType apikey updates, permits only token-kind web-session providers stored with authType cookie, and rejects cookie-kind and other auth types. The added tests cover the allowed and denied branches, and the diff is scoped and whitespace-clean. I found no merge-blocking defect. The focused test could not be executed in this environment because the temporary PR checkout lacked usable dependency resolution, so runtime validation should rely on CI or the author's reported validation.

diegosouzapw:

PR #10519, opened 2026-08-16
fix(providers): test token-backed web sessions

Zartharas:
## Summary

Route token-kind web-session connections stored with `authType: "cookie"` through the existing API-key/provider validator instead of the OAuth connection-test path.

## Root cause

Bulk-imported token-kind web sessions such as DeepSeek and Z.AI retain `authType: "cookie"` while their token is stored in `apiKey`. The connection-test dispatcher previously selected the API-key/provider validator only when `authType === "apikey"`. Every other connection fell through to the OAuth test path, where these providers are unsupported.

## Fix

Add a narrow dispatch predicate that preserves normal API-key behavior and additionally selects the API-key/provider validation path only for `authType: "cookie"` providers whose centralized web-session credential requirement is `kind: "token"`.

Cookie-kind web sessions and other auth types retain their existing paths.

## Validation

- focused dispatch regression: 4/4 passed
- scoped ESLint: PASS
- `git diff --check`: PASS
- one commit directly parented to the current release branch tip

The exact failure mode was also live-validated with a disposable DeepSeek token-backed cookie-auth connection. Before the fix its connection test was dispatched as OAuth and reported unsupported. After deployment, an invalid token produced the provider validator's literal HTTP 401, a refreshed valid token tested successfully, `authType` remained `cookie`, and the disposable connection was removed.\n\nNo raw credential values are included in the PR or validation output.

### DIFF
--- a/changelog.d/fixes/10519-token-backed-web-session-test-dispatch.md
+++ b/changelog.d/fixes/10519-token-backed-web-session-test-dispatch.md
@@ -1,0 +1,1 @@
+- **fix(providers):** test token-backed web sessions through their provider validator instead of the OAuth path ([#10519](https://github.com/diegosouzapw/OmniRoute/pull/10519)) — thanks @Zartharas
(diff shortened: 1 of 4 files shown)


diegosouzapw:
The DeepSeek/Z.ai intent looks reasonable, but this predicate currently matches every cookie-auth token-kind requirement. That includes hailuo-web and promptql, which have no token-aware entry in validateProviderApiKey; they fall through to validateWebCookieProvider, which sends the value as Cookie rather than Hailuo's required token header and treats most non-401/403 responses as valid. Please narrow the predicate to providers with a compatible validator (or add those validators), and add a regression for the other token-kind entries. I could not execute the head test because this checkout lacks the tsx dependency, so this is a plausible, source-verified concern rather than a runtime-confirmed result.

PR #10521, opened 2026-08-16
fix(compliance): redact extra provider API keys

Zartharas:
## Summary

Redact `providerSpecificData.extraApiKeys` before provider credential audit metadata is serialized.

## Root cause

Audit serialization already recursively redacts sensitive keys such as `apiKey`, `token`, `userToken`, `cookie`, secrets, and passwords. However, the normalized key for `extraApiKeys` is `extraapikeys`, which did not match the existing singular `apiKey` rules. As a result, an array of additional provider API keys could survive audit serialization unchanged.

## Fix

Add the exact normalized `extraapikeys` key to the existing audit sensitive-key allowlist. This keeps the change narrow and avoids overly broad plural matching that could redact unrelated telemetry fields.

Extend the existing compliance audit regression to verify redaction of `extraApiKeys` alongside the already-covered token and cookie credential forms.

## Validation

- compliance audit regression: 5/5 passed
- source scoped ESLint: PASS
- base/current test lint debt parity: 5 pre-existing `no-explicit-any` findings in both
- `git diff --check`: PASS

No production credential values are included in the change or validation output.

### DIFF
--- a/changelog.d/fixes/10521-audit-extra-api-keys-redaction.md
+++ b/changelog.d/fixes/10521-audit-extra-api-keys-redaction.md
@@ -1,0 +1,1 @@
+- **fix(compliance):** redact additional provider API keys from audit-log payloads ([#10521](https://github.com/diegosouzapw/OmniRoute/pull/10521)) — thanks @Zartharas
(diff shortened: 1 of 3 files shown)


diegosouzapw:
Reviewed PR #10521 against release/v3.8.50. The change is narrow and correct: extraApiKeys normalizes to extraapikeys, which is now explicitly redacted, and the regression fixture verifies the redaction together with existing credential forms. No review findings. The focused compliance suite passed locally (5/5), but the available checkout was not the PR head, so that validation is reported as plausible rather than confirmed. The base-red #9985 alert is inherited and is not attributed to this PR.

Issue #10460, opened 2026-08-15
fix(resilience): model-unsupported 400 rotates every account for the same provider/model


RaviTharuma:
### OmniRoute Version

3.8.49 (official Docker image `diegosouzapw/omniroute:3.8.49`)

### Installation Method

Docker / Kubernetes

### Operating System

Linux

### Provider(s) Involved

GitHub Copilot account pool (3 active accounts)

### Model(s) Involved

`github/claude-fable-5`

### Client Tool

Claude Code via Anthropic-compatible `POST /v1/messages`

### Description

When GitHub returns the authoritative model-scoped response:

```text
400 The requested model is not supported.
```

OmniRoute rotates to the next GitHub account and repeats the exact same provider/model request. With three active GitHub accounts it made three guaranteed-to-fail upstream calls before the combo could finish.

This response is materially different from an account-scoped quota, rate-limit, or entitlement failure. Once a provider says the requested model does not exist/is unsupported, retrying another credential for the same provider/model adds latency and load but cannot recover the request.

### Exact live chronology

All timestamps UTC on 2026-08-15; account IDs and request content redacted.

```text
07:58:59.290  combo falls back to github/claude-fable-5
07:58:59.292  try GitHub account 1/3
07:59:00.738  400 The requested model is not supported
07:59:00.739  try GitHub account 2/3
07:59:02.229  same model-scoped 400
07:59:02.230  try GitHub account 3/3
07:59:04.597  same model-scoped 400
07:59:04.599  target finally exhausted
```

The three GitHub connections were healthy/authenticated. Their connection health only proved credential validity; it did not prove this model existed.

### Steps to Reproduce

1. Configure at least two active GitHub provider accounts.
2. Put `github/<unsupported-model>` in a priority combo followed by any sibling target.
3. Send a request that reaches the GitHub target.
4. Have GitHub return `400 The requested model is not supported.`
5. Observe the same model request repeated once for every GitHub account.

### Actual Behavior

- Model-scoped unsupported is treated like an account fallback opportunity.
- Every account is tried for the same provider/model.
- The request incurs avoidable latency before the combo advances/exhausts.

### Expected Behavior

Classify unambiguous model-scoped responses such as these as `provider_model_unsupported`:

- `The requested model is not supported`
- `model ... is not supported`
- `model not found` when the provider's semantics are authoritative

After the first such response:

1. stop rotating credentials for that provider/model;
[...]

Issue #10312, opened 2026-08-13
fix(backend): default debugMode to false for fresh installs


RaviTharuma:
### OmniRoute Version

3.8.49 (official image `diegosouzapw/omniroute:3.8.49`)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Linux (fresh install / new volume — no persisted settings row for `debugMode`)

### Node.js Version

Image default (official Docker image)

### Provider(s) Involved

N/A (settings defaults; affects all installs)

### Model(s) Involved

N/A

### Client Tool

N/A (gateway logging behaviour; visible under any client load)

### Description

**`getSettings()` defaults `debugMode: true`. Every install that does not already have a persisted `debugMode` key runs in debug.**

`src/lib/db/settings.ts` (default object overlaid before sqlite rows, ~line 215):

```ts
debugMode: true,
```

That default object is the baseline for settings **before** persisted rows are applied. If the key is missing from storage (fresh volume, partial migration, wiped settings, new deploy), the runtime value is **`true`**.

**Why this is wrong:**

1. **CONTRIBUTING / product docs** describe Debug Mode as an **Advanced UI toggle — opt-in**. Default `true` is the opposite of opt-in.
2. Production logs grow with extra debug-level noise whenever the key is absent — exactly when operators expect a quiet, production-shaped install.
3. Neighbouring flags in the **same** default object are correctly opt-in (`logToolSources: false`, `hidePaidModels: false`, etc.). `debugMode: true` is inconsistent with that pattern.
4. No migration is required to fix this: change the default to `false`. Installs that already persisted `debugMode: true` keep `true`. Only the "key missing" path flips to sane production behaviour.

**Observed operator impact on 3.8.49 (fresh / missing key):**

- Debug-shaped log volume without anyone opening Settings → Advanced → Debug Mode.
- Harder to spot real warns (including the header-budget drops and event-loop stalls filed separately) because the baseline is already chatty.
- Surprised operators: "we never enabled debug" — correct; the code enabled it for them.

This is a one-line default mistake with production blast radius on every new install.

### Steps to Reproduce

1. Start official `diegosouzapw/omniroute:3.8.49` against an **empty** data volume (no prior settings rows), or delete only the `debugMode` key from settings storage and restart.
2. Do **not** open the Advanced UI toggle; do not PATCH settings.
3. Read effective settings (dashboard API or sqlite/settings dump — whatever the install exposes for current values).
[...]

PR #10525, opened 2026-08-16
fix(settings,auth): default debugMode to false and skip account rotation on model-unsupported 400

HouMinXi:
**Problem**: Two bugs from issue triage:

1. **#10312**: `debugMode` defaults to `true` in `src/lib/db/settings.ts:215`. Every fresh install (no persisted setting) gets verbose debug logging by default — the opposite of opt-in. Neighbouring flags (`logToolSources`, `hidePaidModels`) correctly default to `false`.

2. **#10460**: When a provider returns `400 The requested model is not supported`, OmniRoute rotates to the next account of the same provider and retries the same doomed request. With N GitHub accounts, this wastes N-1 upstream calls before the combo can advance. The model-unsupported signal is provider-scoped (all accounts share the same model catalog), not account-scoped.

**Fixes**:

1. `src/lib/db/settings.ts:215` — `debugMode: true` → `false`. Existing persisted settings are unaffected.

2. `src/sse/services/auth.ts` — new early-exit guard in `markAccountUnavailable()`: when `status === 400` and the error text matches `MODEL_ACCESS_DENIED_PATTERNS` (from `accountFallback.ts`), return `{shouldFallback: false, cooldownMs: 0}` without cooling down the account. The error propagates to the combo layer which already has `isModelScoped400()` (`combo.ts:1827`) to advance to the next combo target. The guard is placed after the terminal-status, anti-thundering-herd, and codex-scope guards, and before the `checkFallbackError` call.

**Verification**:
- `tests/unit/settings-debugmode-default.test.ts` (2 tests): asserts `debugMode === false` for fresh installs, regression guard.
- `tests/unit/account-fallback-service.test.ts` (4 new tests):
  - Model-unsupported 400 returns `shouldFallback: false` and does NOT rate-limit the connection (DB state verified).
  - Various phrasings all correctly skip rotation (4 error text variants, each verified against DB).
  - Body-specific 400 ("Invalid message format") still goes through normal path (not caught by model guard).
  - Non-400 status (429) with model-unsupported text does NOT trigger guard (guard is status-400-only).
  - Empty errorText with 400 does NOT trigger guard.
  - Auth-credential 400 ("Invalid API key provided for model gpt-4o") does NOT match guard (regex-verified).
  - Guard path is distinguishable from generic 400 path via DB state (guard keeps connection active).
- Bug-injection: reverting the guard makes the model-unsupported tests fail (red→green verified).
- 86/86 tests pass, `npm run typecheck:core` clean, ESLint clean on all touched files.
- Forge review: R1 (deepseek expert) + R5 (gemini-omniroute) — all findings are advisory/test-assertion/pre-existing, no blocking issues. The auth-credential false-positive concern was disproved by regex testing.

⚠️ base-red inherited: #9985

### DIFF
--- a/open-sse/services/accountFallback.ts
+++ b/open-sse/services/accountFallback.ts
@@ -302,7 +302,7 @@ export const MODEL_ACCESS_DENIED_PATTERNS = [
 // across every target, masking the real "fix your credential" error. When the
 // text clearly indicates a bad credential, the regex-based model-access detection
 // is suppressed (structured codes/types like model_not_found are unaffected).
-const AUTH_CREDENTIAL_ERROR_PATTERNS = [
+export const AUTH_CREDENTIAL_ERROR_PATTERNS = [
   /\b(?:invalid|incorrect|expired|missing|revoked)\s+api[\s_-]?key\b/i,
   /\bapi[\s_-]?key\s+(?:is\s+)?(?:invalid|incorrect|expired|missing|revoked|not\s+valid)\b/i,
   /\bauthentication\s+(?:failed|error|required)\b/i,
--- a/open-sse/services/accountFallback.ts
+++ b/open-sse/services/accountFallback.ts
@@ -311,6 +311,45 @@ export const AUTH_CREDENTIAL_ERROR_PATTERNS = [
   /\bnot\s+authenticated\b/i,
 ];
 
+// #10460: strict subset of MODEL_ACCESS_DENIED_PATTERNS that is unambiguously
+// PROVIDER-wide — the model does not exist / is not served by this provider at all, so
+// no account of that provider could serve it (e.g. "The requested model is not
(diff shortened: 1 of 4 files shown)


diegosouzapw:
O caso básico de `400 The requested model is not supported` está encaminhado, mas eu não recomendaria merge ainda. O guard usa as regexes de model access diretamente e pode reclassificar mensagens de permissão da chave/conta sem a desambiguação existente em `checkFallbackError`. Além disso, os testes só chamam `markAccountUnavailable` isoladamente: não provam três contas, uma única chamada upstream e avanço real do combo, que são os critérios centrais da #10460. Também falta a razão de log `provider_model_unsupported`, e o fixture altera DATA_DIR depois de imports que já capturaram o diretório do DB. A alteração de debugMode não é um finding deste PR contra a base atual, pois já está na base remota.

PR #10528, opened 2026-08-16
fix(network): bound direct-path response-start timeout and retry on fresh socket (#10214)

### LINKED ISSUE #10214: fix(network): direct dispatcher stalls opencode-go and command-code until service restart
## Summary

`opencode-go` and `command-code` requests can stall inside a long-running OmniRoute process until the 30s model-test deadline, even though direct upstream calls from the same host succeed. Restarting `omniroute.service` immediately restores both providers.

## Observed on

- OmniRoute `3.8.50`
- Ubuntu 24.04 Hyper-V VM
- Node `v24.19.0`

## Evidence

Before the incident, the same VM process completed both providers:

- `opencode-go/minimax-m3`: 3.9s
- `command-code/deepseek/deepseek-v4-flash`: 5.2s

After the service had been running for a while, both stalled:

```text
[STREAM] OPENCODE-GO | deepseek-v4-flash-max | 29979ms | error: Model test deadline exceeded after 30000ms
[ProxyEgress] opencode-go/<redacted> ... proxy=direct status=timeout
```

Direct connectivity from the same VM remained healthy:

- `fetch("https://opencode.ai/zen/go/v1/models")` returned 200 in about 1.9s.
- `https://api.commandcode.ai/alpha/generate` returned 401 in about 0.14s without credentials (expected reachability result).
- Antigravity and Codex requests through OmniRoute continued to work.

Neither failing provider had an assigned proxy:

- opencode-go connection: `proxy_enabled=0`, no account proxy.
- command-code executor has no proxy context.
- OmniRoute service had no HTTP(S)/ALL proxy environment variables.

After `systemctl restart omniroute.service`:

- `opencode-go/minimax-m3` returned HTTP 200 in 1.58s.
- `cmd/deepseek/deepseek-v4-flash` returned HTTP 200 in 3.90s.

## Expected

A stalled direct dispatcher connection should time out/fail over without requiring a service restart.

## Actual

The request remained pending until the caller/model-test deadline. The existing retry path appears to handle dispatcher errors, but this condition did not surface a transport error to trigger it.

## Notes

The VM also had a separate dead Oracle SOCKS tunnel (`145.241.115.216`) at the time. It is unrelated to these requests: the providers were direct/no-proxy and the OmniRoute process had no sockets to the local SOCKS listeners. The tunnel is included only to avoid conflating two concurrent infrastructure conditions.

## Possible direction

Add bounded response-start/headers protection around direct pooled-dispatcher fetches and emit diagnostics identifying the dispatcher/origin when a request remains pending. A direct request should then retry through the fresh/no-keep-alive dispatcher rather than consuming the full model-test deadline.

excessivechaos:
Fixes #10214 — direct (no-proxy) egress stalls opencode-go / command-code until a service restart.

## Root cause

The direct egress path in `open-sse/utils/proxyFetch.ts` funnels every request through the cached round-robin pool of 32 one-connection Agents (`getDefaultDispatcher()`). A keep-alive pooled socket that silently drops (half-open, no RST — VPN reconnects, edge idle-drops) surfaces **no transport error**, so the existing fresh-socket retry (which only fires on `UND_ERR`/`ECONNRESET`/`fetch failed`) never triggers: the request hangs until undici's `headersTimeout` (600s default) or the caller's per-model deadline, then fails. A service restart drops the stale sockets, which is why the same providers recover immediately — and re-wedge under sustained traffic.

Verified live:
- `model=orchestrator` through a long-running gateway: opencode-go leg hangs exactly 120s (combo timeout), **zero TCP connections** from the gateway process during the hang.
- Identical request on a fresh isolated instance (same bundle, same DB, same key, same egress): **200 in 1.8s**.
- Same-host direct fetch to the upstream: fast response. All 4 upstream edge IPs respond.
- Health checks pass throughout (they use a different fetch path with fresh sockets).

## Fix

Bound each direct attempt's **response-start** window (fetch promise = headers only; the body streams through untouched afterwards):

- New per-attempt bound, default 30s, tunable via `OMNIROUTE_DIRECT_HEADERS_TIMEOUT_MS` (0 disables — previous behavior).
- On expiry, retry **once** on the fresh no-keep-alive dispatcher (`getRetryDispatcher`) — a brand-new socket that cannot be the zombie — instead of consuming the full 600s headersTimeout/caller deadline.
- A second timeout (or a non-replayable body) surfaces a classified `DIRECT_RESPONSE_START_TIMEOUT` error so the combo fails over to the next target; no native-fetch fallback (native fetch pools too and would hang identically).
- Caller-initiated aborts propagate unchanged (no retry).

## Tests

`tests/unit/proxyfetch-direct-response-start-timeout-10214.test.ts` (3 tests):
1. Pooled attempt hits a silent zombie socket → bounded timeout → retries on the fresh no-keep-alive dispatcher → success (asserts dispatcher identity `getDefaultDispatcher` → `getRetryDispatcher`).
2. Both attempts stall → classified `DIRECT_RESPONSE_START_TIMEOUT` surfaces, native fallback never fires.
3. Healthy fast response → single attempt, untouched by the bound.

Plus all 44 existing proxyFetch/proxyDispatcher tests: **47/47 pass**.

## End-to-end verification

A/B against a local server that accepts the first POST and never responds (real half-open zombie socket), with `OMNIROUTE_DIRECT_HEADERS_TIMEOUT_MS=5000`:

| Bundle | Result |
|---|---|
| Old (live dist) | request hangs 22s+ (no failover — bug reproduced) |
| This branch (built) | **200 in 5418ms** — bounded timeout fired, retried on a fresh socket, POST #2 succeeded |

Log evidence from the fixed bundle:
`[ProxyFetch] Direct response-start timeout (5000ms) on pooled dispatcher — retrying on fresh no-keep-alive dispatcher: 127.0.0.1:19090`

Also verified the fixed bundle handles the real opencode-go path: `model=orchestrator` → 200 in 1.97s streaming `deepseek-v4-flash`.


diegosouzapw:
Não foi possível revisar o código com segurança: /tmp/review-prs-0816-diffs/10528.diff está vazio (0 bytes). Além disso, o PR aberto aponta para main, enquanto a base solicitada é release/v3.8.50. A metadata disponível descreve 584.103 adições, 105.697 remoções e 100 arquivos, portanto não é seguro inferir o patch pretendido a partir dela. Solicito um diff não vazio contra release/v3.8.50 e o retarget do PR antes de avaliar feature, segurança, testes, impacto cross-layer ou a regra #18. O alerta base-red #9985 permanece herdado e não foi atribuído ao PR.

diegosouzapw:
The PR is now correctly targeted at `release/v3.8.50`, but the original review artifact had an empty diff and could not establish the intended change safely. Please provide a current-head diff/reproduction and rerun the release gates before it is reconsidered. The PR remains open.

Issue #10347, opened 2026-08-14
fix(resilience): cool down embedding accounts after terminal 402 failures


RaviTharuma:
### OmniRoute Version

3.8.49 (official image `diegosouzapw/omniroute:3.8.49`)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Linux (single-replica orchestrated deploy)

### Node.js Version

Image default (official Docker image)

### Provider(s) Involved

Mistral (embedding path). Pattern likely applies to any provider that returns **HTTP 402** (subscription / payment required) on embed.

### Model(s) Involved

Mistral embedding model(s) routed through OmniRoute’s embed handler (not chat-combo specific)

### Client Tool

Any client issuing embedding requests through OmniRoute (Desktop / API `/v1/embeddings` class traffic)

### Description

**The embedding path does not appear to apply the same account circuit-breaker / skip-on-402 behaviour that chat combo uses — a dead Mistral subscription is burned on every embed request.**

Live evidence on official **3.8.49** (single process logs, sanitized):

Within the **same minute**:

1. Multiple `EMBED mistral error 402` lines.
2. Multiple `[ERROR] [402]: Check your subscription on https://admin.mistral.ai/subscription` (upstream’s public admin URL).
3. Then `[ProxyEgress] mistral/<prefix> ... status=error` (twice in the sample window).

Chat-side cooldown / skip-on-hard-error behaviour has been discussed in the product (e.g. account cooldown work such as #8133-class behaviour for combo). **Embed traffic in this sample kept hitting the same dead subscription instead of cooling it down and skipping.**

**What is solid:**

- 402 from Mistral embed is logged repeatedly in a short window.
- Upstream message points at subscription status (public admin page).
- Proxy egress marks the account attempt as `status=error` repeatedly — no evidence of a “already known bad subscription, skip” short-circuit in the log stream for embed.

**Suspicion (not proven end-to-end in source for every branch):**

> Embedding handler does **not** share (or does not invoke) the same account circuit-breaker / skip-on-402 path that chat combo uses. Each embed request re-selects or re-tries the dead account and pays the full upstream round-trip + error logging cost.

Impact:

- Wastes upstream quota/error budget on a subscription that will not recover without operator action.
- Amplifies log noise and egress work on the same event loop.
- Can poison multi-account pools if embed keeps “trying” a known-402 account instead of rotating to a healthy one (if any).

### Steps to Reproduce

[...]

PR #10529, opened 2026-08-16
fix(embeddings): cool down account on hard errors (402/401/5xx)

HouMinXi:
**Problem**: Issue #10347 — embedding path does not cool down accounts on hard errors (402 subscription expired, 401 revoked, 403 forbidden, 5xx server errors). A dead Mistral subscription is retried on every embedding request instead of being cooled down and skipped.

Root cause: `src/lib/embeddings/service.ts` calls `handleEmbedding()` but does NOT call `markAccountUnavailable()` on error. The chat path does this in `src/sse/handlers/chat.ts:1942`.

**Fix**: Added `markAccountUnavailable` call in `createEmbeddingResponse()` after `handleEmbedding` returns an error. Only triggers on hard error statuses (401/402/403/404/429/5xx) — skips 400 (bad request) since the account is fine. Best-effort design (fire-and-forget with `.catch` debug logging) mirrors chat.ts behavior.

**Changes**:
- `src/lib/embeddings/service.ts`: import `markAccountUnavailable`, add cooldown call on hard errors, replace `@/lib/localDb` barrel import with direct `@/lib/db/` imports
- `tests/unit/embedding-account-cooldown-10347.test.ts`: 3 tests (402→credits_exhausted, 500→unavailable, 400→no cooldown)

**Verification**:
- 3/3 tests pass
- Forge review R1-R3: all CONFIRMED findings are by-design (fire-and-forget pattern matches chat.ts) or already handled (anti-thundering-herd guard in markAccountUnavailable)

⚠️ base-red inherited: #9985

### DIFF
--- a/config/quality/eslint-suppressions.json
+++ b/config/quality/eslint-suppressions.json
@@ -1083,11 +1083,6 @@
       "count": 1
     }
   },
-  "src/lib/embeddings/service.ts": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/lib/evals/runtime.ts": {
     "no-restricted-imports": {
       "count": 1
--- a/src/lib/embeddings/service.ts
+++ b/src/lib/embeddings/service.ts
@@ -10,13 +10,14 @@ import { errorResponse, unavailableResponse } from "@omniroute/open-sse/utils/er
 import { HTTP_STATUS } from "@omniroute/open-sse/config/constants.ts";
 import * as log from "@/sse/utils/logger";
 import { toJsonErrorPayload } from "@/shared/utils/upstreamError";
-import { getProviderCredentials, clearRecoveredProviderState } from "@/sse/services/auth";
 import {
-  getCachedProviderNodes,
-  getComboByName,
-  getCombos,
-  getDatabaseSettings,
-} from "@/lib/localDb";
+  getProviderCredentials,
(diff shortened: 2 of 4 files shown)


diegosouzapw:
O gancho de produção parece corrigir o caminho ausente de cooldown, mas os testes novos chamam markAccountUnavailable diretamente. Eles não exercitam createEmbeddingResponse/handleEmbedding nem provam que uma segunda requisição deixa de chamar a conta resfriada, portanto o teste passaria mesmo se o bloco novo da produção fosse removido. Sugiro adicionar um teste de integração em processo para 402, persistência do cooldown e segunda seleção (rotação para uma conta saudável ou fast-fail quando todas estiverem resfriadas). Não atribuo falhas herdadas ao alerta base-red #9985.

HouMinXi:
Hi @diegosouzapw — good point, the direct-call tests wouldn't catch a regression if the production block were removed.

I've added an integration test that exercises the full flow:

1. Seeds two connections for the same provider
2. First request → upstream returns 402 → connection gets cooled ()
3. Second request (no  specified) → credential selection skips the cooled account → picks the healthy second connection
4. Verifies the second connection stays 

This proves the production code path actually calls  AND that the cooldown persists through to the next credential selection. Removing the production block would now break the test.

The direct-call tests are retained for unit-level coverage of the  function itself.

PR #10533, opened 2026-08-16
fix(models): correct Codex context and combo limit resolution

### LINKED ISSUE #10530: fix(models): combo catalog ignores explicit Codex context override
## Summary

Codex-connected GPT-5.6 models have a lower context limit than the same models accessed through the OpenAI API. OpenAI's own Codex catalog lists `context_window: 272000` and `max_context_window: 272000` for the GPT-5.6 Sol, Terra, and Luna variants, while the direct API model pages advertise a 1,050,000-token context window.

Sources:
- [OpenAI Codex model catalog](https://github.com/openai/codex/blob/9ded177ce7c1c0bd2047f902936c177612ab3434/codex-rs/models-manager/models.json)
- [GPT-5.6 Sol API model page](https://developers.openai.com/api/docs/models/gpt-5.6-sol)

The immediate bug is that an explicit model context override still does not affect combo metadata. For example, after overriding `codex/gpt-5.6-terra` to `272000` and placing it in a single-target combo, the combo row in `/v1/models` still advertises the built-in 1M window.

`getComboTargetCatalogMetadata()` recomputes limits from synced metadata → registry → spec → fallback and bypasses the persisted model-context override. `/api/combos` uses a separate aggregation path, so both paths need to resolve the same effective window.

## Expected behavior

- Combo metadata honors explicit per-model context overrides and takes the minimum effective window across members.
- `/v1/models` and `/api/combos` advertise the same effective combo window.
- Codex-provider defaults reflect Codex's own limits rather than direct API limits.

## Related

- #6191 documents the lower usable Codex input budget; it does not cover explicit overrides inside combos.
- #9431 / #9432 changed Codex GPT-5.6 metadata to the direct API's 1M limits, but the official Codex catalog reports 272K for Codex-connected models.
- #10225 concerns runtime rejection before compression; this issue concerns model discovery and combo metadata advertised to clients.

jackjinke:
## Summary

- Set Codex-provider GPT-5.6 Sol, Terra, and Luna context/input limits to `272000` while preserving direct OpenAI API limits.
- Derive Combo context metadata from canonical effective limits so persisted overrides agree across `/v1/models` and `/api/combos`.
- Keep context-window and max-input overrides independent.

Source: https://github.com/openai/codex/blob/9ded177ce7c1c0bd2047f902936c177612ab3434/codex-rs/models-manager/models.json

Fixes #10530.

## Validation

- Focused follow-up: 18 passed, 0 failed
- Broader Node scope: 198 files, 1,606 passed, 2 skipped
- Vitest auto-combo scope: 15 passed
- ESLint and `npm run check:changelog-integrity` passed
- `runner-base` built and passed the health smoke test

Changed tests:
- `tests/unit/codex-gpt56-catalog.test.ts`
- `tests/unit/combo-context-prefix-resolution.test.ts`
- `tests/unit/models-catalog-combo-metadata.test.ts`
- `tests/unit/provider-models-route-codex.test.ts`
- `tests/unit/vscode-token-routes-gpt56.test.ts`

## Coverage / Risk

The repository-wide coverage command was not run locally because its root glob includes lifecycle tests that can stop port `20128`; isolated CI remains authoritative for the 60% gate.

The change is limited to Codex limits and shared Combo metadata resolution. Direct OpenAI limits and explicit override precedence remain unchanged.

### DIFF
--- a/changelog.d/fixes/10530-codex-combo-context.md
+++ b/changelog.d/fixes/10530-codex-combo-context.md
@@ -1,0 +1,1 @@
+- **fix(models):** align Codex GPT-5.6 context limits with the Codex catalog and honor model context overrides when advertising combos ([#10530](https://github.com/diegosouzapw/OmniRoute/issues/10530))
--- a/open-sse/config/providers/shared.ts
+++ b/open-sse/config/providers/shared.ts
@@ -279,8 +279,8 @@ export const GPT_5_6_CODEX_CAPABILITIES = {
   supportsReasoning: true,
   supportsVision: true,
   supportsXHighEffort: true,
-  contextLength: 1050000,
-  maxInputTokens: 922000,
+  contextLength: 272000,
+  maxInputTokens: 272000,
   maxOutputTokens: 128000,
 } as const;
 
--- a/src/app/api/v1/models/catalog.ts
+++ b/src/app/api/v1/models/catalog.ts
@@ -431,27 +431,12 @@ async function buildUnifiedModelsResponseCore(
       const syncedInputModalities = parseJsonStringArray(synced?.modalities_input);
       const syncedOutputModalities = parseJsonStringArray(synced?.modalities_output);
 
-      const syncedContext = isPositiveFiniteNumber(synced?.limit_context)
-        ? synced.limit_context
-        : undefined;
-      const registryContext = isPositiveFiniteNumber(registryModel?.contextLength)
-        ? registryModel.contextLength
-        : undefined;
-      const specContext = isPositiveFiniteNumber(spec?.contextWindow)
-        ? spec.contextWindow
-        : undefined;
-      const contextLength =
-        syncedContext ??
-        registryContext ??
-        specContext ??
-        (getTokenLimit(providerId, modelId) || undefined);
-      const registryInputLimit = isPositiveFiniteNumber(registryModel?.maxInputTokens)
-        ? registryModel.maxInputTokens
-        : undefined;
-      const syncedInputLimit = isPositiveFiniteNumber(synced?.limit_input)
-        ? synced.limit_input
-        : undefined;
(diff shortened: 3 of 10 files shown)


jackjinke:
Update: Tibo has just confirmed that Codex supports a 1M context window and suggested using model capability overrides to opt into it. That is consistent with this PR: the built-in Codex defaults stay at `272000`, while users who need the 1M path can set both `context_length` and `max_input_tokens` to `1000000`. The public Codex [models.json](https://github.com/openai/codex/blob/main/codex-rs/models-manager/models.json) still advertises 272K at the moment, so there may be a catalog/rollout lag.

The pricing rationale also remains:

- OpenAI's GPT-5.6 API docs expose a `1,050,000`-token context window (`922,000` max input), but say prompts over `272K` input are billed at **2x input** and **1.5x output** for the full request ([Sol](https://developers.openai.com/api/docs/models/gpt-5.6-sol), [Terra](https://developers.openai.com/api/docs/models/gpt-5.6-terra), [Luna](https://developers.openai.com/api/docs/models/gpt-5.6-luna)). Keeping `272000` as the default avoids opting ordinary Codex requests into long-context pricing while retaining an explicit 1M opt-in.
- OmniRoute's Settings → Model Overrides supports both `context_length` and `max_input_tokens` for `codex/<model>`. Because this PR keeps these limits independent, users wanting the 1M path should set **both** to `1000000`; setting only `context_length` still leaves the 272K input cap. The override changes OmniRoute's advertised/routing guard; upstream long-context billing still applies.

diegosouzapw:
I revalidated the live PR metadata after the head moved. The current failing quality/unit checks overlap the open release-base-red issue #9985, so they are not being attributed to this PR without a pure-base reproduction. The updated head remains a train candidate pending the local combined-tip gates; no head-specific defect was asserted from the stale snapshot.

PR #10541, opened 2026-08-16
refresh opencode provider

backryun:
## Summary

- refresh the OpenCode free, Go, and Zen static model catalogs
- keep the Zen catalog newest-first and preserve intentionally curated removals
- add official OpenCode Go rates only for the current curated catalog and its effort aliases
- resolve curated GPT 5.6 Luna and Qwen context tiers plus DeepSeek UTC Peak/Off-Peak rates
- calculate local 5-hour, weekly, monthly, and per-model usage from `usage_history`
- keep upstream quota percentages authoritative and use the local ledger as breakdown/fallback
- render OpenCode Go quota totals as USD instead of incorrectly labeling them as requests

## Validation

- OpenCode-focused Node suite: 302 passed, 0 failed
- `npm run test:vitest`: 362 passed, 0 failed
- `npm run lint`: passed
- `node scripts/check/check-docs-sync.mjs`: passed
- `npm run check:any-budget:t11`: passed
- `node scripts/check/check-tracked-artifacts.mjs`: passed
- `git diff --check`: passed
- `npm run test:unit`: the new OpenCode Go pricing, ledger, usage, and UI tests pass; the full suite still exits non-zero on pre-existing release-base failures including stale CLI/docs counts, the onnxruntime assertion, and environment/timing fixtures

Prettier was intentionally not run for this refresh.

### DIFF
--- a/open-sse/config/providers/registry/opencode/go/index.ts
+++ b/open-sse/config/providers/registry/opencode/go/index.ts
@@ -13,145 +13,48 @@ export const opencode_goProvider: RegistryEntry = {
   authPrefix: "Bearer",
   defaultContextLength: 200000,
   models: [
-    // Port from decolua/9router 8efacc11: align with official Go endpoints —
-    // glm-5.2 is now advertised and Kimi chat traffic must route through
-    // `kimi-k2.7-code` (the live API rejects the plain `kimi-k2.7` alias for
-    // `/chat/completions`, even though the docs config example uses it).
-    // GLM-5.2 — base model + effort-tier aliases (#6922).
-    // OpencodeExecutor rewrites the alias to the canonical id and injects
-    // reasoning_effort, mirroring the deepseek-v4-pro-* pattern.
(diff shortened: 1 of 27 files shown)


backryun:
The models that were removed or excluded from the list above were removed because they offer exactly the same usage as the latest models.
so there was no reason to keep them on the list. I hope they won’t need to be added back.

diegosouzapw:
The head changed since the review snapshot, and the newly arrived #10574 overlaps the reasoning path. The earlier review conclusion is therefore stale; please reconcile the current diff with #10574 and restore fallback coverage before this is reconsidered for the release train.

backryun:
> The head changed since the review snapshot, and the newly arrived #10574 overlaps the reasoning path. The earlier review conclusion is therefore stale; please reconcile the current diff with #10574 and restore fallback coverage before this is reconsidered for the release train.

Since the work I'm currently doing is a mess, I'll polish it up completely and then create a new PR.

PR #10545, opened 2026-08-16
fix(sse): include prompt cache usage fields on message_stop fallback path

### LINKED ISSUE #10535: fix(api): OpenAI-format responses drop prompt-cache usage fields (no prompt_tokens_details.cached_tokens)
## Summary

When OmniRoute fronts an Anthropic-compatible upstream and the client speaks OpenAI chat-completions format, the response (including the final SSE usage chunk with `stream_options.include_usage: true`) carries only:

```json
"usage": {"prompt_tokens": N, "completion_tokens": M, "total_tokens": N+M}
```

The prompt-cache counters OmniRoute itself tracked for that very call (`call_logs.tokens_cache_read`, and `response._omniroute.claudePromptCacheUsage.cacheReadTokens` in the detail artifact) are never mapped into the standard OpenAI spelling `usage.prompt_tokens_details.cached_tokens` (nor `cache_creation_tokens`).

The claude-format endpoint (`/v1/messages`) does emit `usage.cache_read_input_tokens` on the wire correctly — only the OpenAI translation drops it.

## Evidence (3.8.49, npm install)

Two identical streaming requests, same 626-token prefix, via `POST /v1/chat/completions`:

- run 1 + run 2 both return `usage: {prompt_tokens: 626, completion_tokens: 3, total_tokens: 629}` — no `prompt_tokens_details`.
- The same pair via `POST /v1/messages`: run 2 returns `usage: {..., cache_read_input_tokens: 576}` — cache works and is reported.
- `call_logs` for the OpenAI-format calls record `tokens_cache_read` correctly (e.g. 28160/28516), so the data is in the pipeline at serialization time.

## Impact

Any OpenAI-compatible client that surfaces cache-hit ratio (DeepSeek Harness, pi-ai-based harnesses, cost dashboards) reads the standard fields and reports a persistent **0% cache hit**, even when upstream caching is working at ~98%. It also distorts cost accounting for providers that bill cached tokens separately.

## Expected

When the pipeline has cache counters for a call, OpenAI-format responses include:

```json
"usage": {"prompt_tokens": N, "completion_tokens": M, "total_tokens": N+M,
  "prompt_tokens_details": {"cached_tokens": <cacheRead>, "cache_creation_tokens": <cacheWrite>}}
```

(omitting the object when both are zero/absent is fine.)

Happy to test a patch on my install.

NahuSaruf:
## Summary
- `claude-to-openai.ts` built the final OpenAI-format `usage` payload in two places: the `message_delta` finish path (already mapped `cache_read`/`cache_creation` into `prompt_tokens_details`) and a `message_stop` fallback used when the finish signal arrives without usage on the same event.
- The fallback never copied the cache fields, so OpenAI-compatible clients saw "0% cache hit" even though the data was present in `state.usage` (matches the report: Claude-format responses had the counters, OpenAI-format ones didn't).
- Mirrored the existing `prompt_tokens_details` mapping from the `message_delta` path into the `message_stop` fallback.

Fixes #10535

## Test plan
- [x] New regression test: "Claude stream: message_stop includes prompt_tokens_details when usage arrived on an earlier message_delta without stop_reason (#10535)" — fails before the fix, passes after.
- [x] `node --import tsx/esm --test tests/unit/translator-resp-claude-to-openai.test.ts` — 16/16 passing.
- [x] `npm run typecheck:core` — clean.
- [x] `npm run lint` (suppressions applied) — no new violations on the two changed files.

⚠️ base-red inherited: #9985 (the two failing jobs on the latest schedule run — "Validate main branch" and "Bank ratchet shrinks" — are unrelated to this change; "Validate active release branch" passed on the current tip).

### DIFF
--- a/open-sse/translator/response/claude-to-openai.ts
+++ b/open-sse/translator/response/claude-to-openai.ts
@@ -298,6 +298,8 @@ export function claudeToOpenAIResponse(chunk, state) {
       if (!state.finishReasonSent) {
         const finishReason =
           state.finishReason || (state.toolCalls?.size > 0 ? "tool_calls" : "stop");
+        const cachedTokens = state.usage?.cache_read_input_tokens || 0;
+        const cacheCreationTokens = state.usage?.cache_creation_input_tokens || 0;
         const usageObj =
           state.usage && typeof state.usage === "object"
             ? {
@@ -310,6 +312,16 @@ export function claudeToOpenAIResponse(chunk, state) {
                         reasoning_tokens: state.usage.reasoning_tokens,
                         completion_tokens_details: {
                           reasoning_tokens: state.usage.reasoning_tokens,
+                        },
+                      }
+                    : {}),
+                  ...(cachedTokens > 0 || cacheCreationTokens > 0
+                    ? {
+                        prompt_tokens_details: {
+                          ...(cachedTokens > 0 ? { cached_tokens: cachedTokens } : {}),
+                          ...(cacheCreationTokens > 0
+                            ? { cache_creation_tokens: cacheCreationTokens }
+                            : {}),
                         },
                       }
                     : {}),
--- a/tests/unit/translator-resp-claude-to-openai.test.ts
+++ b/tests/unit/translator-resp-claude-to-openai.test.ts
@@ -403,6 +403,40 @@ test("Claude stream: message_stop falls back to tool_calls when tool use already
   assert.equal(result[0].choices[0].finish_reason, "tool_calls");
 });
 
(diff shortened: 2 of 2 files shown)


diegosouzapw:
The focused translator suite passes 15/15 and lint is clean. Before marking this merge-ready,
please add or confirm coverage for cache-creation-only, read-plus-creation, missing
`state.usage`, and preservation of `prompt_tokens_details` on the `message_stop` fallback. These
are small table-driven cases but protect the usage/billing contract.

PR #10546, opened 2026-08-16
fix(i18n): add missing routing and compression messages

rizxfrog:
## Summary

- I have completed some translations for zh-CN and zh-TW.

## Validation

zh-CN
zh-TW

- Change type: i18n


diegosouzapw:
The JSON, UI coverage, value-drift, and both Chinese glossary checks are healthy. The previously
questioned `供應商` value is an allowed zh-TW synonym. Please rerun the locale checks in the merge
train; reviewing the broader zh-TW terminology churn is optional scope hygiene.

PR #10552, opened 2026-08-16
fix(db): preserve runtime driver require in webpack standalone bundles

excessivechaos:
## Problem\nNext.js webpack standalone builds can rewrite dynamic createRequire(import.meta.url) calls so native SQLite drivers fail to resolve. The driver cascade then silently falls back to sql.js, whose whole-file persistence can expose a torn database after a crash.\n\n## Fix\n- Use a small runtime loader: eval("require") inside CommonJS webpack chunks, with createRequire(import.meta.url) fallback for ESM dev/tests.\n- Route the DB driver callers through that loader.\n- Persist sql.js exports through a temp file followed by renameSync.\n- Add a runtime-loader regression test.\n\n## Verification\n- Focused driverFactory and sqljsAdapter tests: 27 passed.\n- Changed-file ESLint: passed.\n- Live Windows standalone verification: native WAL regenerated, ping healthy, and no new database-malformed errors after restart.\n\nThe direct-egress/Crof timeout fix is separate (#10528).

### DIFF
--- a/src/lib/db/adapters/driverFactory.ts
+++ b/src/lib/db/adapters/driverFactory.ts
@@ -1,4 +1,4 @@
-import { createRequire } from "node:module";
+import { runtimeRequire as _require } from "./runtimeRequire";
 import { existsSync } from "node:fs";
 import { createBetterSqliteAdapter } from "./betterSqliteAdapter";
 import { createBunSqliteAdapter, type BunSqliteDatabaseLike } from "./bunSqliteAdapter";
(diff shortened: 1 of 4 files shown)


diegosouzapw:
The PR has been retargeted and verified against `release/v3.8.50`. It still needs an in-place replacement for the `eval("require")` loader, which violates the repository no-eval rule, followed by standalone-artifact validation. It remains out of the train until that fix lands.

PR #10554, opened 2026-08-16
fix(vision): preserve high detail for inline images

rinseaid:
## Summary

OpenCode serializes pasted image file parts as OpenAI `image_url` content without a `detail` field. OmniRoute then forwards the provider default, which is low-detail sampling. That makes wide, text-dense screenshots unreadable even though the original image bytes are intact.

This change:

- Defaults missing `image_url.detail` to `high` for OpenAI chat messages.
- Defaults missing `detail` to `high` for Responses `input_image` parts.
- Preserves explicit client values such as `low`, `auto`, and `high`.
- Uses high detail for the internal OpenAI-compatible vision bridge.

## Verification

- `tests/unit/chatcore-upstream-body.test.ts`
- `tests/unit/guardrails/visionBridgeHelpers.callVisionModel.test.ts`
- Targeted ESLint passed.
- The regression fails before the change and passes after it.

---
Resubmit of #10323, which was auto-closed when its source fork was deleted. Same diff, rebased onto current release/v3.8.50 (applied cleanly, no changes needed).

### DIFF
--- a/docs/security/GUARDRAILS.md
+++ b/docs/security/GUARDRAILS.md
@@ -93,6 +93,19 @@ describe prompt, steering the description toward what the user actually asked
 (codex-vision-proxy pattern) and asking the vision model to transcribe visible
 text. With the flag off — or no user text — the base prompt is used unchanged.
 
+The describe self-loop's own OpenAI-compatible request (`callVisionModelSingle()`
+in `visionBridgeHelpers.ts`) always requests `image_url.detail: "high"` —
+unconditionally, for every caller/provider, not gated on any client signal.
+Low-detail sampling degrades OCR accuracy for exactly the text-transcription
+task this prompt asks for, so the describe call itself always asks for high
+detail regardless of what detail level the original inbound request used. This
+only affects the internal describe request body; it does not change how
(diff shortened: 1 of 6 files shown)


diegosouzapw:
The focused tests and lint pass, but the current patch applies the `high` detail default at a general upstream boundary as well as the OpenCode path. Please either constrain it to the intended OpenCode path with a regression test for unrelated providers, or document and test the global compatibility/cost contract before the train.

rinseaid:
Good catch. `defaultImageDetail` was gated on nothing, so it fired for every provider/format through `prepareUpstreamBody`, not just OpenCode. Fixed by gating it on `isOpencodeClient` (the existing User-Agent/`x-opencode-*` header signal already used for `bypassDefaultToolLimit` at this same call site), not on `provider`, since `provider` is the upstream target and unrelated to which client sent the request.

Added a regression test confirming a non-OpenCode caller against the same `opencode-zen` provider keeps `detail` untouched. All 17 tests in `chatcore-upstream-body.test.ts` pass; typecheck clean on the touched files. Pushed as `5209c19`.

PR #10558, opened 2026-08-16
Database settings page returns HTTP 500 when SQLite lacks the optional dbstat table

TechNickAI:
## Summary

On a SQLite build compiled without the optional `dbstat` virtual table, `getDatabaseStats()` throws — and every caller dies with it. The most visible symptom is that `GET` and `PATCH /api/settings/database` return HTTP 500, so the database settings page cannot be read or changed at all: page size, cache size, and vacuum settings all become unreachable.

`dbstat` is compile-time optional (`ENABLE_DBSTAT_VTAB`). It is absent from sql.js/WASM builds, which is the fallback driver this project already ships and supports.

The function had a per-table `SELECT SUM(pgsize) FROM dbstat` with no guard around it.

## The fix

Probe `dbstat` once per call. If it is unavailable, skip the per-table size lookups and report size `0`.

This mirrors a guard the function already had. A few lines above, the `COUNT(*)` lookup deliberately swallows `no such module:` errors for exactly this reason — optional virtual-table modules may be missing on a given connection. The `dbstat` query simply sat outside that existing pattern; this brings it inside.

What is preserved and what is lost:

- **Preserved:** total size, page count, page size, cache size. These come from pragmas and stay accurate.
- **Lost:** per-table byte sizes, which report `0`.

That trade is deliberate. Per-table sizes are informational; a hard 500 on the settings API is not recoverable by the user.

Unrelated failures still propagate — the guard matches only the missing-`dbstat` messages and rethrows everything else, so genuine I/O errors and corruption are not swallowed.

Both driver spellings are handled: sql.js reports `no such module: dbstat` while better-sqlite3 can surface `no such table: dbstat`.

## Related Issues

- Extends the same defensive pattern introduced in #7313 (`fix(db): tolerate unavailable virtual table modules in stats`), which guarded the `COUNT(*)` path in this function but not the `dbstat` path.

## Validation

- [x] Change type: DB
- [x] `npm run lint`

Focused run on this branch, rebased on current `release/v3.8.50`:

```
npx tsx --test tests/unit/db/stats-dbstat-optional.test.ts
ℹ tests 8   ℹ pass 8   ℹ fail 0
```

## Tests Added Or Updated

- `tests/unit/db/stats-dbstat-optional.test.ts` — new.

The test drives `getDatabaseStats()` through a minimal `SqliteAdapter` double that implements only the surface the function actually touches; anything else throws, so a new dependency shows up loudly rather than passing silently. Cases covered:

- `dbstat` missing with the `no such module:` spelling (sql.js)
- `dbstat` missing with the `no such table:` spelling (better-sqlite3)
- messages carrying a driver prefix before the recognised text
- an unrelated error, asserted to still propagate rather than being swallowed
- a driver that answers the probe but fails on a later per-table query
- `SUM()` returning `{ size: null }` over an empty table

`stryker.conf.json` registers the new file alongside the sibling DB suites.

## Coverage Notes

This PR changes `src/lib/db/stats.ts`. The new suite covers both branches of the new `isDbstatAvailable()` probe, the rethrow path for unrelated errors, and the degraded per-table path. Coverage on the touched file moves up.

## Reviewer Notes

- **No behaviour change on builds that have `dbstat`.** The probe succeeds and the per-table query runs exactly as before. The only new cost is one extra `SELECT ... LIMIT`-shaped query per `getDatabaseStats()` call against `sqlite_master`.
- The probe intentionally runs once per call rather than being cached across calls. `getDatabaseStats()` is not on a hot path, and caching would have to be invalidated per connection; the driver can differ between connections in this codebase.
- If you would prefer per-table sizes to be reported as `null` rather than `0` to distinguish "unavailable" from "empty", that is a one-line change and a reasonable call — `0` was chosen to keep the existing response shape and avoid a type change for API consumers.
- Found while running this project's sql.js fallback driver in production. The settings page was returning 500 with no obvious cause until the driver difference surfaced.

### DIFF
--- a/changelog.d/fixes/dbstat-optional-vtab.md
+++ b/changelog.d/fixes/dbstat-optional-vtab.md
@@ -1,0 +1,1 @@
+- **fix(db):** database settings API no longer returns HTTP 500 on SQLite builds compiled without the optional `dbstat` virtual table (sql.js/WASM); per-table sizes degrade to 0 instead of failing the whole stats call
--- a/src/lib/db/stats.ts
+++ b/src/lib/db/stats.ts
@@ -24,6 +24,26 @@ export interface DatabaseStats {
   cacheSize: number;
 }
 
+/**
+ * `dbstat` is a compile-time-optional SQLite virtual table (ENABLE_DBSTAT_VTAB).
+ * It is unavailable on sql.js/WASM and on any build compiled without it, where
+ * querying it throws "no such module: dbstat" (some drivers report "no such
+ * table: dbstat"). Per-table byte sizes are a nice-to-have, so probe once and
+ * degrade to 0 rather than failing the whole stats call — and with it every
+ * caller, including the database settings API.
+ */
+function isDbstatAvailable(db: SqliteAdapter): boolean {
+  try {
+    db.prepare(`SELECT SUM(pgsize) as size FROM dbstat WHERE name = ?`).get("sqlite_master");
+    return true;
+  } catch (error) {
+    if (error instanceof Error && /no such (module|table): dbstat/i.test(error.message)) {
+      return false;
+    }
+    throw error;
+  }
+}
+
 export function getDatabaseStats(db: SqliteAdapter = getDbInstance()): DatabaseStats {
   const pageSize = db.pragma("page_size", { simple: true }) as number;
   const pageCount = db.pragma("page_count", { simple: true }) as number;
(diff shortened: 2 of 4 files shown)


diegosouzapw:
The existing virtual-table suite and the new dbstat-optional matrix pass 11/11, and focused lint
passes. The fallback tolerates only the optional dbstat condition and rethrows unrelated SQLite
errors; keep #9985 base-red failures separate.

diegosouzapw:
Thank you @TechNickAI for the narrow optional-dbstat fallback. The current head was validated with the focused dbstat/virtual-table matrix and ESLint; unrelated SQLite errors remain visible. The prior FAST train evidence is retained for this release merge.

Issue #9625, opened 2026-08-06
fix(db): domain_cost_history retention compares millisecond rows with a seconds cutoff


post1917a:
### OmniRoute Version

3.8.49 in production; source rechecked on `release/v3.8.50` at `7f36b192`

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Description

`domain_cost_history.timestamp` is written in Unix milliseconds, but retention cleanup computes its cutoff in
Unix seconds. Therefore normal 13-digit rows are never older than the 10-digit cutoff and are never removed,
even if the cleanup scheduler is running.

### Source Evidence

[`src/lib/db/domainState.ts:373-380`](https://github.com/diegosouzapw/OmniRoute/blob/7f36b192f00a533306722f86d4af3b1ff317110c/src/lib/db/domainState.ts#L373-L380):

```ts
export function saveCostEntry(apiKeyId: string, cost: number, timestamp = Date.now())
```

`Date.now()` returns milliseconds.

[`src/lib/db/cleanup.ts:254-270`](https://github.com/diegosouzapw/OmniRoute/blob/7f36b192f00a533306722f86d4af3b1ff317110c/src/lib/db/cleanup.ts#L254-L270):

```ts
const cutoffEpoch = Math.floor(Date.now() / 1000) - retentionDays * 86_400;
db.prepare("DELETE FROM domain_cost_history WHERE timestamp < ?").run(cutoffEpoch);
```

The cutoff is in seconds.

### Production Evidence

With a 30-day retention setting:

- total `domain_cost_history` rows: 17,681
- rows older than 30 days when interpreted as milliseconds: 17,675
- a one-time cleanup with a millisecond cutoff removed those 17,675 rows

The stock seconds-based comparison could not match those rows.

### Expected Behavior

Rows older than `retention.domainCostHistory` should be deleted.

### Suggested Fix

Use a millisecond cutoff for current rows. If legacy installations may contain second-based timestamps, handle
both magnitudes explicitly rather than comparing both formats to one cutoff. Add tests covering both 10-digit
legacy timestamps and 13-digit current timestamps.

### Duplicate Check

GitHub issue search found #6848 (the general telemetry-retention feature) and #2125 (large JSON exports), but
no report of this timestamp-unit mismatch.

Issue #6848, opened 2026-07-11
feat(backend): auto-cleanup/upsert for telemetry tables to bound storage.sqlite


md-riaz:
## Problem

OmniRoute's `storage.sqlite` grows unbounded over time with no auto-cleanup for most tables. After ~1 week of normal use, the DB reaches 100-176MB+, which causes **repeated OOM crashes** because sql.js (WASM) loads the entire file into the Node.js heap.

This is a **server-killer** — the process crashes every few minutes, requiring manual restarts. Non-technical users cannot maintain this.

### Affected tables (no auto-cleanup)

| Table | Rows | Purpose | Needs persistence? |
|---|---|---|---|
| `quota_snapshots` | **392,224** | Quota tracking snapshots | No — hourly/daily aggregates suffice |
| `compression_analytics` | 78,278 | Compression stats | No — summary metrics enough |
| `compression_cache_stats` | 66,390 | Cache hit/miss counters | No — rolling window |
| `xp_audit_log` | 60,942 | XP/gamification audit | No — recent entries only |
| `domain_cost_history` | 49,712 | Per-domain cost tracking | No — daily rollups exist |
| `proxy_logs` | 18,594 | Proxy request logs | No — debug tool |
| `usage_history` | 16,342 | Per-request usage | No — summary tables exist |
| `model_capabilities` | 6,949 | Model catalog | Refreshable from sync |
| `compression_run_telemetry` | 5,737 | Per-run telemetry | No — aggregate instead |

**Total: ~675,000+ rows of telemetry that don't need to persist.**

### What happens

1. DB grows past ~100MB after 5-7 days of normal use
2. sql.js probe OOMs trying to load the file into V8 heap
3. Probe-failed/restore loop fires (issue #6631)
4. Server hangs or crashes repeatedly
5. User must manually restart with `OMNIROUTE_MEMORY_MB=4096`
6. DB keeps growing — problem recurs

### Environment

- **Version**: v16.2.9 (npm), also confirmed on v3.8.46
- **Platform**: Ubuntu Linux, 14GB RAM
- **DB size**: 176MB (176MB data + 176MB WAL) after ~1 week
- **Crash frequency**: 3,594 OOM crashes in a single session log

### 1. Auto-cleanup on startup (immediate)

Add a startup cleanup pass that truncates tables exceeding a row limit:

```sql
-- Keep only recent entries
[...]

PR #10559, opened 2026-08-16
Compression telemetry retention has never deleted a row (same unit bug as #9625)

TechNickAI:
## Summary

`cleanupCompressionRunTelemetry()` has never deleted a row.

Its cutoff is computed in epoch **seconds**:

```ts
const cutoffEpoch = Math.floor(Date.now() / 1000) - retentionDays * 86_400;
```

but the `timestamp` column stores epoch **milliseconds**, stamped by the only writer, `insertCompressionRunTelemetryRow()`:

```ts
).run(
  Date.now(),   // epoch ms
  ...
```

A millisecond timestamp is ~1000× larger than a seconds cutoff, so `WHERE timestamp < cutoff` never matches an old row. The sweep runs on schedule, reports `0 deleted`, and `compression_run_telemetry` grows without bound.

That sweep was added by #6848 specifically to stop telemetry tables from growing `storage.sqlite` unbounded. For this table, the guard has been inert since it landed.

## This is the same bug as #9625

#9625 fixed this exact unit mismatch in `cleanupDomainCostHistory()` — the sibling function ~90 lines earlier in this same file — and left this comment:

```ts
/**
 * Clean up old domain_cost_history based on retention settings. (#6848)
 * The `timestamp` column stores epoch milliseconds (saveCostEntry default
 * is Date.now()), so the cutoff must be in milliseconds to match. (#9625)
 */
```

That fix changed one call site. `cleanupCompressionRunTelemetry()` has the identical defect and was not touched.

This PR applies the same correction, and updates the stale docstring — which still asserted `Uses unix-epoch timestamp column (INTEGER)`, the very claim that made the bug easy to miss on review.

## Why it survived a year: the existing test agreed with the bug

`tests/unit/telemetry-auto-cleanup-6848.test.ts` seeded this table in **seconds**:

```ts
const nowSeconds = Math.floor(now / 1000);
insert.run(nowSeconds - 40 * DAY_SECONDS, 1000, 500);
```

Production writes milliseconds. The test asserted that a seconds-cutoff correctly deletes seconds-stamped rows, which is true and irrelevant — no production row is ever shaped that way. The suite passed, so the retention path looked covered.

Note the same test file already uses `nowMilliseconds` for `domain_cost_history`, updated by #9625. This table was left on the seconds path.

This PR corrects that seeding to milliseconds. That is the more durable half of the change: without it, the fix would make the existing test fail and the natural next step would be to "fix" the test back toward the bug.

## Related Issues

- Same defect class as #9625, which corrected the sibling sweep and missed this call site
- Restores the intent of #6848, whose telemetry retention has been inert for this table

## Validation

- [x] Change type: DB
- [x] `npm run lint`

The new test was verified in both directions on this branch, rebased on current `release/v3.8.50`.

With the source fix reverted, the repro fails on exactly the symptom described:

```
actual: 0,
expected: 3,
```

With the fix applied:

```
[Cleanup] Deleted 3 compression_run_telemetry older than 30 days
✔ the writer stamps epoch milliseconds, not seconds
✔ cleanupCompressionRunTelemetry deletes rows older than the retention window
ℹ pass 2   ℹ fail 0
```

Related suites on the same branch:

```
tests/unit/telemetry-auto-cleanup-6848.test.ts    ℹ pass 6   ℹ fail 0
tests/unit/db-cleanup.test.ts                     ℹ pass 7   ℹ fail 0
tests/unit/repro-9625.test.ts                     ℹ pass 2   ℹ fail 0
```

## Tests Added Or Updated

- `tests/unit/repro-compression-run-telemetry-ms.test.ts` — new.
- `tests/unit/telemetry-auto-cleanup-6848.test.ts` — updated: seeds this table in milliseconds, matching the writer. Two now-unused constants (`DAY_SECONDS`, `nowSeconds`) removed.

The repro deliberately does **not** hand-write a timestamp for its first assertion. It inserts through the real writer and asserts the stored value is millisecond-scale, so the test also fails if the producer's format ever diverges from the consumer again — the failure mode that produced this bug. The retention assertion then inserts through the writer and backdates, preserving the producer's unit while controlling age.

## Coverage Notes

This PR changes `src/lib/db/cleanup.ts`. The new suite covers the corrected cutoff path (rows outside the window deleted, rows inside retained) and pins the producer's unit. Coverage on the touched file moves up.

## Reviewer Notes

- **This is a behaviour change on a table that has silently accumulated.** After merge, the first sweep on a long-running instance will delete every row older than the retention window in one pass. That is the intended behaviour and matches every sibling sweep, but on a large `compression_run_telemetry` the initial `DELETE` will be bigger than steady-state. Worth a line in release notes for operators with old databases.
- No migration and no schema change. The column already holds milliseconds; only the comparison was wrong.
- The `86_400_000` literal matches the style already used by `cleanupDomainCostHistory()` after #9625.
- Found while running scheduled retention maintenance against a production router. The verification step flagged that the writer and the cleanup disagreed on units before any rows were deleted, which is what surfaced it.


diegosouzapw:
The focused retention/repro suite passes 8/8 and confirms the millisecond-unit correction deletes
old compression telemetry while preserving current rows; focused lint also passes. Keep any #9985
base-red failures separate.

diegosouzapw:
Thank you @TechNickAI for the millisecond-consistent compression telemetry retention fix. The current head was validated with the focused 8-test writer/cleanup matrix and lint, and the retention contract is explicit in both insertion and cutoff paths. The prior FAST train evidence is retained for this release merge.

PR #10561, opened 2026-08-16
fix(responses): preserve DeepSeek reasoning content

### LINKED ISSUE #10550: fix(combos): preserve Chat reasoning for DeepSeek Responses targets
## Summary

A combo can mix `opencode-go/deepseek-v4-flash` (Chat Completions) and `deepseek/deepseek-v4-flash` (Responses). When OpenCode Go produces an assistant tool-call turn, the client sends the complete `reasoning_content` back in later Chat Completions history. If the combo then selects official DeepSeek, OmniRoute drops that reasoning while converting Chat history to Responses input, and DeepSeek returns:

```text
```

This is a cross-format conversion issue, not a reasoning replay cache miss: genuine client reasoning is already present.

## Minimal example

Client history contains:

```json
{
  "role": "assistant",
  "reasoning_content": "<reasoning returned by OpenCode Go>",
  "content": "",
  "tool_calls": [
    {
      "id": "call_123",
      "type": "function",
      "function": { "name": "echo_number", "arguments": "{\"value\":166}" }
    }
  ]
}
```

For official DeepSeek Responses, this should become:

```json
[
  {
    "type": "reasoning",
    "content": [{ "type": "reasoning_text", "text": "<same reasoning>" }]
  },
  {
    "type": "function_call",
    "call_id": "call_123",
    "name": "echo_number",
    "arguments": "{\"value\":166}",
    "status": "completed"
  },
  {
    "type": "function_call_output",
    "call_id": "call_123",
    "output": "166"
  }
]
```

Instead, the `reasoning` item is omitted.

## Live validation

Using the same OpenCode Go assistant reasoning and tool call:

- Continue through OpenCode Go Chat Completions: `200`.
- Call official DeepSeek Responses without a `reasoning` item: `400 reasoning_text must be passed back`.
- Call official DeepSeek Responses with the OpenCode Go text converted to a plaintext `reasoning` / `reasoning_text` item: `200`.

This confirms the reasoning can carry across providers when the Chat-to-Responses conversion is correct.

DeepSeek documents its Responses API as stateless and supports plaintext `reasoning` input items: https://api-docs.deepseek.com/guides/responses_api

## Likely OmniRoute issues

- `open-sse/translator/request/openai-responses/toResponses.ts` unconditionally skips assistant `reasoning_content`, based on OpenAI's server-generated `rs_*` ID constraint. That rule should not suppress DeepSeek's supported plaintext reasoning input.
- `open-sse/services/responsesInputPolicy.ts` removes plaintext `reasoning` items unless they contain preserved encrypted reasoning. This also strips a valid DeepSeek Responses input item, including when the incoming request already uses `/v1/responses`.

The fix should be target-aware: preserve the existing OpenAI Responses safety rule while converting and retaining plaintext reasoning for official DeepSeek Responses targets.

y4ho0:
## Summary

- Preserve client-provided assistant `reasoning_content` when Chat history is translated to the official DeepSeek Responses API.
- Retain only non-empty plaintext `reasoning_text` for official DeepSeek targets while removing unsupported `summary`, `encrypted_content`, and server item IDs.
- Keep OpenAI, Codex, OpenRouter, and custom provider behavior unchanged, and make Responses input sanitization copy-on-write so a DeepSeek combo attempt cannot mutate later fallback input.

## Why

When a combo switches from an OpenCode Go Chat Completions target to official DeepSeek Responses after a tool call, OmniRoute previously dropped the returned reasoning text. DeepSeek requires that reasoning to be passed back and otherwise responds with HTTP 400.

The conversion is intentionally target-aware: only the built-in official `deepseek` provider receives plaintext reasoning items. OpenAI/Codex encrypted reasoning and server-generated `rs_*` handling retain their existing safety rules.

## Validation

- [x] Focused translator, policy, chat-core, and real combo-route tests: 151 passed
- [x] `npm run lint`
- [x] `npm run check:open-sse-typecheck`
- [x] `npm run check:file-size -- --base-ref e646fe84c76d2ddaf1aa8deb6bf49052852d839b`
- [x] `npm run check:changelog-integrity`
- [x] `npm run check:test-runner-api`
- [x] `npm run check:test-discovery`
- [x] `npm run build:backend` (compiled successfully, generated 462/462 static pages, and assembled the standalone bundle)

> ⚠️ base-red inherited: #9985. The active release branch has unrelated existing failures; this change does not modify those paths.

Closes #10550


jackjinke:
Thanks for the focused fix here. I wanted to flag that #10574 provides an overlapping implementation for the same #10550 continuation failure.

The main distinction is scope: this PR applies a focused official-DeepSeek plaintext reasoning fix and includes a route-level regression test, while #10574 introduces a more general provider-declared transport policy covering both DeepSeek plaintext and compatible opaque Responses state. It also includes related Responses-to-Chat, Combo fallback-isolation, and known-provider UI behavior.

#10574 now also has an optional per-Combo Advanced Setting for incompatible continuation state. Missing/default `skip` retains target-local fallthrough; explicit `drop` removes all active continuation reasoning and tries the selected target once. This does not alter reasoning capability filtering, requested effort, thinking budgets, or provider-specific thinking controls.

The PRs modify the same policy, ChatCore, translator, and test paths, so they are alternative approaches rather than changes that should both merge as-is. A neutral comparison table is included in the discussion on #10550.

diegosouzapw:
The current head is still outside the original review snapshot and the author has identified overlap with #10574, which arrived after this campaign. I am keeping #10561 out of the release train until the overlap and the route-level resource-pressure failures are reconciled on the current head. No code is being closed or discarded.

diegosouzapw:
Thanks for tackling this — closing as superseded by #10574 (already merged), which fixes the same underlying issue (#10550) more generally.

This PR's fix threads a DeepSeek-specific carve-out (`provider === "deepseek"`) into `applyResponsesInputPolicy` in the old `open-sse/services/responsesInputPolicy.ts`. #10574 replaced that module entirely with `open-sse/services/reasoningInputPolicy.ts`, which makes plaintext reasoning **portable by default for any unannotated target** — provider-bound opaque state is the one that needs an explicit `preserveEncryptedReasoning` declaration, not the other way around. That generalizes exactly the DeepSeek scenario from #10550 without a provider-specific branch.

Verified on the current release tip: `open-sse/translator/request/openai-responses/toResponses.ts` already emits the `{ type: "reasoning", content: [{ type: "reasoning_text", ... }] }` item for any readable, non-placeholder reasoning value, and `tests/unit/translator-openai-responses-req.test.ts` already has explicit `_provider: "deepseek"` coverage for `reasoning_content` preservation across the Chat→Responses conversion (added alongside #10574).

Appreciate you finding and diagnosing this — closing here since there's no remaining scope to carry forward.

PR #10571, opened 2026-08-16
fix(opencode): session stability, free-tier routing, and CLI defaults

CyrixJD115:
## Summary

Fix OpenCode provider defaults: stable session IDs for prompt caching, correct free-tier model routing, and CLI header synthesis enabled by default.

## Problem

Three issues affecting OpenCode provider reliability on VPS deployments:

1. **Prompt caching broken** — `x-opencode-session` was set to `randomUUID()` per request, so opencode.ai's upstream prompt cache never hit across a conversation
2. **Free-tier model misrouting** — models like `big-pickle` and `*-free` could be incorrectly routed to non-opencode providers when the live catalog was temporarily unreachable
3. **CLI header synthesis off by default** — Cloudflare in front of opencode.ai 429s VPS requests lacking CLI identity, but synthesis was opt-in instead of opt-out

## Fix

- Wired `generateSessionId()` from `sessionManager.ts` into `opencodeHeaders.ts` — session ID is now a deterministic SHA-256 fingerprint of model + system prompt + first user message + tools
- Threaded request body through `buildHeaders()` in base executor and `OpencodeExecutor` override so the fingerprint has access to conversation context
- Flipped CLI header synthesis to default ON (opt-out via `OPENCODE_SYNTHESIZE_CLI_HEADERS=false`), with production-proven defaults (`opencode`/`desktop`/`global`)
- Added auto-echo of listing-valid model names for noAuth providers so `response.model` matches `/v1/models` listing
- Short-circuited free-tier model resolution to opencode provider first to prevent prefix inference misrouting

## Results

- Same conversation = same session ID = prompt cache hits on opencode.ai
- Free-tier models (`big-pickle`, `*-free`) correctly route to opencode when available
- VPS requests no longer 429'd by Cloudflare due to missing CLI identity headers
- Client-supplied headers still take precedence over all synthesized defaults

## AI Assisted changes notice

This change was developed with AI assistance (ZCode agent).


diegosouzapw:
The underlying problems (random session IDs breaking prompt-cache hits, header-synthesis being opt-in) look real, but this PR flips OPENCODE_SYNTHESIZE_CLI_HEADERS's default from off to on — a decision the existing test file documents as deliberate and owner-requested to stay off pending live validation — and breaks 2 of that file's assertions (I ran them: `pass 5 / fail 2` on this branch). It also changes the synthesized UA/client/project default values (opencode-cli/1.0.0 → opencode, cli → desktop, default → global) without updating the test that pins those. No new or updated tests ship for the session-fingerprinting, free-tier routing, or echo-model changes either. Could you: (1) confirm with the maintainer that flipping the default is intended, (2) update tests/unit/opencode-cli-headers-synthesis-5997.test.ts accordingly, and (3) add coverage for the three new behaviors?

Issue #10564, opened 2026-08-16
fix(cli): omniroute test 404s, providers test-all skips custom connections the server can test, providers rotate PATCHes a PUT-only route (remains after #10081)


UsryAce:
## Summary

Follow-up to #10081 (fixed in PR #10092 / #10097 for `keys add`, `combo list`, and the OAuth/unsupported misreporting in `providers test-all`). Three related bugs in the local-management CLI layer are still present on current `main` (verified against the file contents on `main` today) and reproduce on the latest published npm release (3.8.49). The gateway server itself is fine in all three cases — the CLI is calling routes/methods the shipped server does not implement, or not calling the server at all.

| # | Command | What the CLI does | What the server actually serves |
| --- | --- | --- | --- |
| 1 | `omniroute test <provider>` | `POST /api/v1/providers/test` → 404 | `POST /api/providers/{id}/test` → 200 |
| 2 | `omniroute providers test-all` | never calls the server; 6-entry local lookup table | `POST /api/providers/{id}/test` works for any connection |
| 3 | `omniroute providers rotate <id>` | `PATCH /api/providers/{id}` → 405 | `PUT /api/providers/{id}` (route exists, accepts PUT) |

## Environment

- omniroute **3.8.49** (`npm i -g omniroute`, latest published), Node v24.14.0, Windows 11 x64
- Server running locally on `:20128`, CLI authenticated via `OMNIROUTE_API_KEY` from `~/.omniroute/.env`

## 1. `omniroute test` posts to a route the shipped server does not have

`bin/cli/commands/test-provider.mjs` line 186 (`_runSingleTest`):

```js
const res = await apiFetch("/api/v1/providers/test", {
  method: "POST",
  body: { provider, model },
  ...
```

Server response:

```console
$ POST /api/v1/providers/test
404 {"error":{"message":"Unknown API route: /api/v1/providers/test","type":"not_found","code":"unknown_route",...}}
```

This is the same missing route documented in #10081, but PR #10092 only patched `keys.mjs`, `combo.mjs`, and `providers.mjs` — `test-provider.mjs` still calls it, so `omniroute test` fails with `HTTP 404` for every provider.

The route that does exist and works (present in `dist/.build/next/server/app/api/providers/[id]/test/`):

```console
$ POST /api/providers/{connectionId}/test        # no body needed
[...]

PR #10572, opened 2026-08-16
fix(cli): route provider tests through connection API

### LINKED ISSUE #10570: fix(cli): route provider test commands through configured connections
## Problem

Two provider-test commands in the local-management CLI no longer match the server's connection-owned API:

1. `omniroute test <provider> [model]` posts to `/api/v1/providers/test`, which the shipped server does not expose, so every run ends in HTTP 404.
2. `omniroute providers test-all` relies on a six-provider local probe table. Custom API-key connections are reported as unsupported even though `POST /api/providers/{connectionId}/test` can validate them.

The same stale response assumption also breaks `omniroute test --all-providers`: `GET /api/providers` now returns `{ connections, total }`, while the CLI only recognizes `providers`/`items`/a raw array and calls `.filter()` on the wrapper object.

This is the provider-test portion of #10564. The separate rotation PATCH/PUT mismatch is already owned by #10366 and is explicitly out of scope here.

## Impact and benefit

Operators currently cannot use the generic test command at all, and cannot obtain a real health result for custom OpenAI-compatible API-key connections through `providers test-all`. Aligning the CLI with the existing authoritative connection test route restores reliable provider validation without adding a duplicate server endpoint or an ever-growing CLI provider table.

### Current

- `omniroute test` sends `POST /api/v1/providers/test` with `{ provider, model }` and receives 404.
- `omniroute test --all-providers` throws when the provider list has the current `{ connections, total }` shape.
- `providers test-all` marks custom API-key connections `unsupported`/`skipped` even while the local server is available.

### Required

- Load configured connections from `GET /api/providers?limit=200` and accept the current `connections` response field while retaining compatible legacy shapes.
- Resolve the generic selector by connection id, connection name, or provider id, preferring an active connection and a requested/default model match when available.
- Call `POST /api/providers/{encoded connectionId}/test`, passing an optional `validationModelId`, and map the server's `valid` field to the generic command's existing `success` result.
- Make non-interactive and interactive `--all-providers` use the same connection-owned route.
- When the local API-key probe reports `unsupported` and the server is reachable, let `providers test-all` use the connection-owned server test instead of skipping.
- Preserve current exit codes, JSON output, supported local probes, offline behavior, and OAuth/no-auth skip behavior.

## Dev design

Keep validation authority at the existing `/api/providers/{id}/test` owner boundary:

- add a small connection-list/selector layer to `test-provider.mjs`;
- carry connection ids into the TUI and non-interactive all-provider paths;
- add a bounded server fallback only after the local API-key probe reports `unsupported`;
- do not add `/api/v1/providers/test` or new provider-specific CLI recipes.

## Alternatives considered

- **Add the missing `/api/v1/providers/test` route:** rejected because it would duplicate the connection-owned route and force a provider/model request to reconstruct connection state.
- **Expand `PROVIDER_TEST_CONFIGS`:** rejected because custom connections are open-ended and the server already has connection-aware validation.
- **Always use the server from `providers test-all`:** rejected to preserve existing local probes and offline behavior for supported API-key providers.

## Scope and non-goals

In scope: generic `omniroute test`, its all-provider/TUI variant, and the unsupported-custom-provider fallback in `providers test-all`.

Out of scope: provider rotation (#10366), server validation internals, credential storage, OAuth refresh semantics, new provider recipes, and offline testing for arbitrary custom providers.

## Test plan

- RED/GREEN: generic `omniroute test` consumes `{ connections }`, resolves the target, calls the encoded connection test route, and maps `valid` to `success`.
- RED/GREEN: `--all-providers` no longer throws on `{ connections, total }` and tests exact connection ids.
- Regression guard: the interactive TUI uses the same connection-owned route and `valid` result.
- RED/GREEN: `providers test-all` falls back to the server for an unsupported custom API-key provider.
- Re-run neighboring CLI provider/rotation/route-unavailable tests, full ESLint, core typecheck, and Vitest.

## Acceptance criteria

- No provider-test path sends a request to `/api/v1/providers/test`.
- Generic and all-provider commands use `/api/providers/{encoded connectionId}/test`.
- A server-testable custom API-key connection receives a real result rather than an unsupported skip.
- Existing supported local probes and non-API-key/offline behavior remain unchanged.

hydraxman:
## Description

Aligns the provider-test CLI paths with the server's connection-owned management API.

- `omniroute test` now resolves a configured connection and calls `POST /api/providers/{id}/test` instead of the missing `/api/v1/providers/test` route.
- `omniroute test --all-providers` accepts the current `{ connections, total }` response shape and carries exact connection ids into both non-interactive and TUI runs.
- `omniroute providers test-all` keeps existing local probes, but falls back to the server for custom API-key connections that the local probe table does not support.
- The TUI uses the shared API client so management authentication and timeout behavior stay consistent with the rest of the CLI.

## Related Issue

Fixes #10570
Related to #10564. The separate rotation method mismatch remains owned by #10366.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Testing

- `node --import tsx/esm --test tests/unit/cli-provider-test-routes-10570.test.ts tests/unit/cli-providers-command.test.ts tests/unit/cli-providers-rotate.test.ts tests/unit/cli-route-unavailable-fallback-10081.test.ts tests/unit/cli-expanded-commands.test.ts` — 41 passed
- `npm run test:vitest` — 40 files, 362 tests passed
- `npm run lint` — passed
- `npm run check:changelog-integrity` — passed
- `npm run check:docs-sync` — passed
- `npm run check:any-budget:t11` — passed
- `npx prettier --check CHANGELOG.md bin/cli/commands/test-provider.mjs bin/cli/tui/ProvidersTestAll.jsx bin/cli/commands/providers.mjs tests/unit/cli-provider-test-routes-10570.test.ts` — passed
- `npm run test:unit` — did not complete within a 600-second local ceiling; the current release branch has an active base-red report for full-suite hangs/failures in #9985. The focused Node suite and Vitest suite above completed successfully.

After publication, four CI jobs failed: `Docs Gates (fast-path)`, `Fast Quality Gates`, and unit shards 1/4 and 3/4. `Fast Quality Gates` reports the current-base `public-creds` and `mutation-test-coverage` gates tracked in #9985. I reproduced both failing unit shards locally with the exact CI command: shard 1 passed 7,619 tests and failed three untouched dependency/config/catalog tests; shard 3 passed 7,493 tests and failed eight untouched Alibaba quota-classification, call-log rotation, docs-count, CLI-catalog-count, and `ServiceSupervisor` tests. The strict docs-count failure is the same current-base drift tracked in #9985; none of the failing tests or source owners are changed by this PR.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes in the focused scope
- [x] I have updated the changelog
- [x] I have read and followed the contribution guidelines

## Contributions

Made by Nathan.


diegosouzapw:
Looks solid — verified `/api/v1/providers/test` doesn't exist and `/api/providers/{id}/test` does, ran the full stated focused test matrix (41/41 pass) locally. Will re-check the 4 CI jobs you flagged as base-red against current #9985 state before merge, otherwise this is ready to go. Thanks for the thorough test coverage.

PR #10573, opened 2026-08-17
fix(backend,combo,cursor): header budget, Codex failover, kv_after_text

### LINKED ISSUE #10215: bug(providers): Cursor kv_after_text truncates pending tool calls on grok-4.5-high (root cause for #9029 / #8784)

### LINKED ISSUE #10310: fix(backend): preserve Codex and Anthropic quota headers under forwarding budget
### OmniRoute Version

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Linux (single-replica container runtime)

### Node.js Version

### Provider(s) Involved

Codex-hosted model; Anthropic (via openai-compatible / native Anthropic-style upstream headers)

### Model(s) Involved

Any streaming chat/completions target that returns `x-codex-*` quota headers and/or `anthropic-ratelimit-unified-*` (not model-ID-specific)

### Client Tool

Desktop / coding-agent multi-stream clients (SSE chat + Responses-style streams)

### Description

**Every SSE stream on 3.8.49 silently drops the Codex/Anthropic quota headers clients need, because the upstream header forwarding budget is only 768 bytes — and `x-codex-turn-state` alone is ~314 bytes of priority-3 junk that burns almost half of it.**

`open-sse/handlers/chatCore/responseHeaders.ts`

```ts
export const MAX_FORWARDED_UPSTREAM_RESPONSE_HEADER_BYTES = 768;
```

Priority function (`getForwardingPriority`):

| Priority | Match |
|---|---|
| 0 | `x-request-id`, `request-id`, `x-correlation-id`, `traceparent`, `traceresponse` |
| 1 | `retry-after` |
| 2 | name includes `ratelimit` **or** `rate-limit` |
| 3 | everything else |

**The bug is structural, not intermittent:**

1. Codex emits operationally critical quota headers whose **names do not contain the substring `ratelimit`**, so they are priority **3**:
   - `x-codex-primary-used-percent`
   - `x-codex-primary-reset-after-seconds`
   - `x-codex-secondary-used-percent` / reset counterparts
   - `x-codex-credits-*`
2. The same stream often carries `x-codex-turn-state` at **~314 bytes**, also priority **3**. That single header eats nearly **half** of the 768-byte budget before any quota header gets a chance.
3. Anthropic unified rate-limit headers **do** match `ratelimit` (priority 2), but under real multi-header Codex+CDN payloads they still lose to the budget once lower-priority bulk and competing priority-2/3 names fill the window — and when they are dropped, clients never see remaining quota either.
4. On **every** streaming response that exceeds the budget, OmniRoute logs:
   ```text
   Dropped upstream response headers that exceeded forwarding budget
   budgetBytes=768 droppedCount=12–26
   ```
   Observed live on official 3.8.49: **every** stream, `droppedCount` consistently in the **12–26** range.

**Dropped names observed in production (safe to list; values redacted):**

| Header family | Why it matters |
|---|---|
| `x-codex-primary-*` | Primary window used % / reset — clients use this for backoff / UI quota |
| `x-codex-secondary-*` | Secondary window |
| `x-codex-credits-*` | Credits remaining / reset |
| `x-codex-turn-state` (~314 B) | Opaque turn blob — **not** useful to most clients; burns budget |
| `anthropic-ratelimit-unified-*` | Anthropic unified rate limit surface |
| `anthropic-organization-id` | Org id |
| `cf-ray` | CDN trace — noise for clients |
| `content-security-policy` | Irrelevant on proxied SSE |
| `date` | Generic |
| `x-robots-tag` | Generic |

**Operator impact:**

- Clients **never** see remaining Codex/Anthropic quota through OmniRoute, even though the upstream sent it.
- Operators cannot build reliable client-side or gateway-side quota-aware routing from response headers.
- Warn logs fire on **every** stream (see companion issue on log storm) so real errors drown.

This is not a "proxy limit is tight" niggle. The priority heuristic was written for OpenAI-style `x-ratelimit-*` names and **misses the entire Codex quota vocabulary**. Combined with a 768-byte cap and a 314-byte turn-state blob, the useful headers lose every time.

### Steps to Reproduce

1. Run official image `diegosouzapw/omniroute:3.8.49` as a single replica.
2. Configure a **codex-hosted model** connection that returns the full `x-codex-*` quota + `x-codex-turn-state` header set on stream open (stock Codex-hosted behaviour).
3. Optionally also configure an Anthropic-path target that emits `anthropic-ratelimit-unified-*`.
4. Issue a streaming chat/completions (or Responses) request through OmniRoute:
   ```bash
   curl -N -sS -D /tmp/or-headers.txt -o /tmp/or-body.sse \
     'http://127.0.0.1:20128/v1/chat/completions' \
     -H "Authorization: Bearer $OR_KEY" \
     -d '{"model":"<codex-hosted-or-combo>","stream":true,"messages":[{"role":"user","content":"ping"}]}'
   ```
5. Inspect what the **client** received:
   ```bash
   rg -i 'x-codex-|anthropic-ratelimit|ratelimit|retry-after' /tmp/or-headers.txt
   ```
6. Inspect OmniRoute logs for the same request window:
   ```text
   Dropped upstream response headers that exceeded forwarding budget
   budgetBytes=768
   droppedCount=12..26
   droppedHeaders: [ { name: "x-codex-...", bytes: N }, ... ]
   ```
7. Confirm in source that priority ignores Codex names:
   - `getForwardingPriority`: only `ratelimit` / `rate-limit` substring → priority 2
   - `x-codex-primary-used-percent` → priority 3
   - `x-codex-turn-state` → priority 3, ~314 wire bytes
   - `MAX_FORWARDED_UPSTREAM_RESPONSE_HEADER_BYTES = 768`

8. Repeat under Desktop multi-stream (several concurrent SSE sessions). Observe the drop on **every** stream, not sporadically.

### Expected Behavior

1. Clients receive the operationally useful quota/reset headers from Codex and Anthropic on every successful stream:
   - `x-codex-primary-*` / `x-codex-secondary-*` / `x-codex-credits-*`
   - `anthropic-ratelimit-unified-*` (and related rate-limit surface)
   - standard `retry-after` / request-id family
2. Budget is large enough for real multi-provider header sets (**2048+** bytes for the upstream-forwarded subset), **or** priority is fixed so quota headers win even under a tight budget.
3. `x-codex-turn-state`, `cf-*`, `content-security-policy`, `date`, `x-robots-tag` are **denylisted** or last-priority so they cannot evict quota headers.
4. Codex used/reset/credits headers are treated as **rate-limit priority** even though their names do not contain the substring `ratelimit`.
5. Operators are not forced to scrape warn logs to learn that clients are blind to remaining quota.

### Actual Behavior

| Observation | Measured on 3.8.49 |
|---|---|
| Budget | `MAX_FORWARDED_UPSTREAM_RESPONSE_HEADER_BYTES = 768` |
| Priority for `x-codex-primary-used-percent` | **3** (name has no `ratelimit`) |
| Priority for `x-codex-credits-*` | **3** |
| Priority for `x-codex-turn-state` | **3**, **~314 bytes** alone |
| Drop frequency | **Every** stream that carries full Codex/CDN header set |
| `droppedCount` | **12–26** per stream |
| Client-visible `x-codex-primary-*` / credits | **Absent** after OmniRoute |
| Client-visible Anthropic unified ratelimit | Often **absent** when competing for the same 768 B |
| Log line | `Dropped upstream response headers that exceeded forwarding budget budgetBytes=768` on each hit |

Result: the gateway advertises streaming success while **stripping the only headers that tell clients how much quota remains**.

### Test Impact

Needs a new unit test

### Error Logs / Output

```text
path: open-sse/handlers/chatCore/responseHeaders.ts
MAX_FORWARDED_UPSTREAM_RESPONSE_HEADER_BYTES=768

getForwardingPriority:
  request-id family -> 0
  retry-after       -> 1
  *ratelimit*       -> 2
  else              -> 3   # includes ALL x-codex-primary/secondary/credits/* and x-codex-turn-state

live (every SSE stream under Desktop multi-stream):
  Dropped upstream response headers that exceeded forwarding budget
  budgetBytes=768
  droppedCount=12..26
  droppedHeaders (names only, values redacted):
    - x-codex-primary-used-percent
    - x-codex-primary-reset-after-seconds
    - x-codex-secondary-used-percent
    - x-codex-secondary-reset-after-seconds
    - x-codex-credits-*
    - x-codex-turn-state          # ~314 wire bytes, priority 3
    - anthropic-ratelimit-unified-*
    - anthropic-organization-id
    - cf-ray
    - content-security-policy
    - date
    - x-robots-tag

client response headers: no x-codex-primary-*, no x-codex-credits-*, quota UI blind
```

### Screenshots

N/A — header absence is measurable from `curl -D` and gateway warn logs.

### Additional Context

- Related log-volume issue: the same drop path emits a **warn with a fat `droppedHeaders` array on every SSE response** (separate issue — header-budget drop log storm).
- Reverse-proxy header limits are real; that is not an excuse to classify the entire Codex quota surface as priority-3 garbage while a 314-byte turn-state blob is allowed to compete equally for the same 768 bytes.
- Suggested fix shape (not prescriptive of final design):
  1. Raise `MAX_FORWARDED_UPSTREAM_RESPONSE_HEADER_BYTES` to **≥ 2048**.
  2. Treat `x-codex-*-used-*`, `x-codex-*-reset-*`, `x-codex-credits-*` as priority **2** (rate-limit class) even without the `ratelimit` substring.
  3. Denylist or force last-priority: `x-codex-turn-state`, `cf-*`, `content-security-policy`, `date`, `x-robots-tag`, similar CDN/noise.
  4. Unit-test a synthetic Headers set with real-sized `x-codex-turn-state` (314 B) + full primary/secondary/credits set and assert quota headers survive.

### Validation Plan

- Unit: `buildStreamingResponseHeaders` with a fixture that includes:
  - `x-codex-turn-state` ≈ 314 bytes
  - full `x-codex-primary-*` / secondary / credits set
  - `anthropic-ratelimit-unified-*`
  - `cf-ray`, `content-security-policy`, `date`
  - Assert: all quota/reset/credits headers present in output; noise headers dropped first; no warn required for the happy path once budget/priority fixed.
- Integration: one live SSE through a codex-hosted target; `curl -D` shows `x-codex-primary-used-percent` (or equivalent) on the client response.
- Regression: budget constant documented; priority table covered for Codex names that lack the `ratelimit` substring.

### LINKED ISSUE #10379: Regression: native Codex turn pin blocks fill-first failover across Codex OAuth connections in 3.8.50
# Regression: native Codex turn pin blocks fill-first failover across Codex OAuth connections in 3.8.50

## Summary

OmniRoute `3.8.50` appears to regress multi-account Codex failover for native Responses/Codex CLI turns.

With a `fill-first` combo containing multiple `codex` OAuth connections, a successful native Codex turn can become pinned to one connection. If that pinned connection is later rejected by a pre-dispatch eligibility check, OmniRoute narrows the target pool to only that connection and returns:

```text
503 Service temporarily unavailable: all targets were skipped by pre-dispatch filters
```

Sibling Codex connections in the same combo are not attempted.

Rolling back to `3.8.49` restores the previously working behavior in the same repository/session setup.

## Environment

- OS: Windows
- Codex CLI: `0.147.0`
- OmniRoute affected version: `3.8.50`
- Working rollback: `3.8.49`
- API mode: OpenAI Responses
- Model/combo: `gpt-5.6-sol`
- Strategy: `fill-first`
- Provider: multiple OAuth connections for the same `codex` provider
- OmniRoute endpoint: local `http://localhost:20128/v1`

The combo contains six Codex OAuth connections in explicit fill-first order.

## Expected behavior

For a fill-first combo:

```text
connection #1
-> exhausted/unavailable
-> connection #2
-> exhausted/unavailable
-> connection #3
-> ...
```

Within a native Codex tool/agent turn, keeping the same provider/model is desirable, but if the currently pinned OAuth connection becomes unavailable, OmniRoute should fail over to another compatible connection for the same provider/model instead of terminating before dispatch.

## Actual behavior on 3.8.50

A long-running resumed Codex CLI thread initially works on connection #2.

One observed successful request:

```text
timestamp:     2026-08-14T10:08:36.797Z
status:        200
model:         gpt-5.6-sol-ultra
connection_id: b6315aef-16fb-446b-83c7-403ff6f452e2
tokens_in:     144241
tokens_out:    1313
duration:      32890 ms
```

Immediately afterward, OmniRoute emitted 30 consecutive failures:

```text
status:        503
model:         gpt-5.6-sol
connection_id: null
tokens_in:     0
tokens_out:    0
duration:      ~16-28 ms
error:         all targets were skipped by pre-dispatch filters
```

There was no recorded upstream `429`, `401`, or other provider failure between the successful request and the first local 503.

The failed requests had no selected connection and no upstream token usage, indicating that routing terminated before dispatch.

## Reproduction characteristics

The problem is request/turn specific:

- Fresh/stateless Codex requests on `3.8.50` successfully skipped an exhausted earlier account and routed to connection #2.
- The resumed long-running Codex thread could work for several model/tool rounds and then enter a state where every continuation returned the same pre-dispatch 503.
- Disabling ordinary session stickiness and prompt-cache affinity did not eliminate the failure.
- Rolling OmniRoute back to `3.8.49` allowed the same WIP thread to progress again.

Temporary diagnostic settings tested on `3.8.50`:

```text
combo.config.disableSessionStickiness = true
promptCacheAffinityEnabled = false
sessionAffinityTtlMs = 0
```

The failure still reproduced, so those mechanisms do not appear to be the primary cause.

## Suspected root cause

In `3.8.50`, `open-sse/services/combo/nativeCodexTurnPin.ts` pins a native Codex turn by:

```text
comboName + thread_id + turn_id
```

with a 45-minute TTL.

The important behavior is in `applyNativeCodexTurnPin()`:

```ts
export function applyNativeCodexTurnPin(
  targets: ResolvedComboTarget[],
  pin: NativeTurnPin
): ResolvedComboTarget[] {
  const target = targets.find(
    (candidate) => candidate.modelStr === pin.modelStr && candidate.provider === pin.provider
  );
  if (!target) return [];
  return [
    {
      ...target,
      connectionId: pin.connectionId,
      allowedConnectionIds: [pin.connectionId],
    },
  ];
}
```

This does not only prioritize the pinned connection. It reduces the target pool to one OAuth connection.

`combo.ts` also changes retry behavior when a native turn pin exists:

```ts
const maxRetries = activeNativeTurnPin ? 0 : (config.maxRetries ?? 1);
const maxSetRetries = activeNativeTurnPin ? 0 : (config.maxSetRetries ?? 0);
```

and then applies the pin:

```ts
if (activeNativeTurnPin) {
  orderedTargets = applyNativeCodexTurnPin(orderedTargets, activeNativeTurnPin);
}
```

After a successful client-managed Responses request, the connection is pinned:

```ts
if (clientManagedResponsesContext && effectiveConnectionId) {
  pinNativeCodexTurn({
    body,
    comboName: combo.name,
    target,
    connectionId: effectiveConnectionId,
  });
}
```

`pinNativeCodexTurn()` also rejects a connection change for an existing turn:

```ts
if (
  existing &&
  (existing.modelStr !== args.target.modelStr ||
    existing.provider !== args.target.provider ||
    existing.connectionId !== args.connectionId)
) {
  throw new Error("Native Codex turn target changed after output was emitted");
}
```

Together, this appears to make same-provider multi-account failover impossible once a native Codex turn has been pinned.

If the pinned connection subsequently fails a pre-dispatch gate, sibling connections are no longer candidates, producing `ALL_TARGETS_SKIPPED`.

## Why this looks like a 3.8.50 regression

The same workflow was working normally on `3.8.49`.

After upgrading to `3.8.50`, the failure became reproducible.

After rolling back:

```text
3.8.50 -> repeated ALL_TARGETS_SKIPPED in resumed native Codex turn
3.8.49 -> same working tree / same Codex thread progresses again
```

This is currently being used as the production workaround.

## Suggested behavior / fix

Native Codex turn affinity should probably remain strict for:

```text
provider
model
thread_id
turn_id
```

but not permanently lock the turn to one OAuth `connectionId` when multiple compatible connections exist for the same provider/model.

A safer policy would be:

1. Prefer the currently pinned connection first.
2. Keep compatible sibling connections for the same `provider + model` as fallback candidates.
3. If the pinned connection is rejected before dispatch or becomes unavailable, fail over to the next compatible connection.
4. After a successful fallback, update the turn pin to the new connection.
5. Continue to reject changing provider/model after output has already been emitted if that invariant is required.

Conceptually:

```ts
const compatible = targets.filter(
  (candidate) =>
    candidate.modelStr === pin.modelStr &&
    candidate.provider === pin.provider
);

const pinned = compatible.find(
  (candidate) => candidate.connectionId === pin.connectionId
);

if (!pinned) {
  return compatible;
}

return [
  pinned,
  ...compatible.filter(
    (candidate) => candidate.connectionId !== pin.connectionId
  ),
];
```

and allow `pinNativeCodexTurn()` to update `connectionId` while keeping `provider` and `modelStr` unchanged.

## Workaround

Downgrade to:

```powershell
npm.cmd install -g omniroute@3.8.49
```

and verify:

```powershell
omniroute.cmd --version
```

Current result:

```text
3.8.49
```

The long-running Codex WIP thread resumed working after the rollback.

## Notes

- No Codex session/state DB deletion was required.
- No repository reset/stash was required.
- Ordinary prompt-cache affinity/session stickiness was tested separately and does not appear to be the root cause.
- The issue is specifically about multi-account failover between OAuth connections of the same `codex` provider during a native Codex tool/agent turn.
- Sensitive account emails/tokens are intentionally omitted from this report.

HouMinXi:
## Summary

Three fixes for backend/routing issues:

### 1. Header forwarding budget + Codex quota priority (#10310)

The 768-byte forwarding budget was too small for real Codex/Anthropic header sets. `x-codex-turn-state` (~314 bytes) consumed nearly half the budget, pushing quota headers (primary/secondary/credits) out.

- Raise `DEFAULT_FORWARDED_HEADER_BUDGET_BYTES` from 768 to 2048
- Classify Codex quota headers (`x-codex-*-used-*`, `x-codex-*-reset-*`, `x-codex-credits-*`) as priority 2 (rate-limit class)
- Deprioritize noise headers (`x-codex-turn-state`, `cf-ray`, `content-security-policy`, `date`, `x-robots-tag`) to priority 4

### 2. Codex turn pin failover (#10379)

`applyNativeCodexTurnPin` narrowed the target pool to the single pinned connection, making same-provider failover impossible when the pinned connection was rejected by pre-dispatch checks (503 ALL_TARGETS_SKIPPED).

- Return all compatible connections (same provider + model) with pinned connection first
- Allow `pinNativeCodexTurn` to update `connectionId` for failover recovery
- Still reject provider/model changes after output was emitted

### 3. Cursor kv_after_text termination (#10215)

`kv_server_message` after text was treated as turn-termination for all Cursor models, but this was only verified safe for `composer-2.5`. On `grok-4.5-high` the checkpoint arrives mid-stream before `exec_mcp` events decode, truncating tool calls.

- Gate `kv_after_text` termination on `isComposerModel()` only
- Non-composer models let `turn_ended`/`tool_call_completed` decide

## Test plan

- `node --import tsx/esm --test tests/unit/quota-header-forwarding-10310.test.ts` — 7 tests
- `node --import tsx/esm --test tests/unit/native-codex-turn-pin-10379.test.ts` — 6 tests
- `node --import tsx/esm --test tests/unit/cursor-kv-after-text-10215.test.ts` — 3 tests
- Existing responseHeaders tests — 19 tests, no regression

## Forge review

PASS — 0 confirmed findings (mimo-v2.5-pro, 3 passes)

Fixes #10310
Fixes #10379
Fixes #10215

### DIFF
--- a/open-sse/services/combo/nativeCodexTurnPin.ts
+++ b/open-sse/services/combo/nativeCodexTurnPin.ts
@@ -74,12 +74,12 @@ export function pinNativeCodexTurn(args: {
   const existing = pins.get(key);
   if (
     existing &&
-    (existing.modelStr !== args.target.modelStr ||
-      existing.provider !== args.target.provider ||
-      existing.connectionId !== args.connectionId)
+    (existing.modelStr !== args.target.modelStr || existing.provider !== args.target.provider)
   ) {
     throw new Error("Native Codex turn target changed after output was emitted");
   }
+  // ConnectionId changes are allowed (failover to sibling connection)
+  // as long as provider + model stay the same.
   const now = Date.now();
   pins.set(key, {
     comboName: args.comboName,
@@ -92,21 +92,37 @@ export function pinNativeCodexTurn(args: {
   prune(now);
 }
 
+/**
+ * Apply a native Codex turn pin to the target list.
+ *
+ * Returns all compatible targets (same provider + model) with the pinned
+ * connection preferred first. This allows fill-first failover: if the
+ * pinned connection is rejected by a pre-dispatch gate, the combo engine
+ * tries the next compatible connection instead of returning 503.
+ *
+ * Provider + model remain locked for the turn — only the connection
+ * can fall over.
+ */
 export function applyNativeCodexTurnPin(
   targets: ResolvedComboTarget[],
   pin: NativeTurnPin
 ): ResolvedComboTarget[] {
-  const target = targets.find(
+  const compatible = targets.filter(
     (candidate) => candidate.modelStr === pin.modelStr && candidate.provider === pin.provider
   );
-  if (!target) return [];
-  return [
-    {
-      ...target,
-      connectionId: pin.connectionId,
-      allowedConnectionIds: [pin.connectionId],
-    },
-  ];
+  if (compatible.length === 0) return [];
+
+  const pinned = compatible.find((t) => t.connectionId === pin.connectionId);
+  const siblings = compatible.filter((t) => t.connectionId !== pin.connectionId);
+
+  // Pinned connection first, then same-provider/model siblings as fallback
+  const ordered = pinned ? [pinned, ...siblings] : compatible;
+
+  return ordered.map((target) => ({
+    ...target,
+    // Allow only connections for the pinned provider+model
+    allowedConnectionIds: compatible.map((t) => t.connectionId),
+  }));
 }
 
 export function revokeNativeCodexTurnPinsForConnection(connectionId: string): number {
--- a/tests/unit/native-codex-turn-pin-10379.test.ts
+++ b/tests/unit/native-codex-turn-pin-10379.test.ts
@@ -17,7 +17,17 @@ const BODY = {
 };
 
 function makeTarget(connectionId: string, model = "gpt-5.6-sol", provider = "codex") {
-  return { connectionId, modelStr: model, provider, executionKey: `ek-${connectionId}` } as any;
+  return {
+    kind: "model" as const,
+    stepId: `step-${connectionId}`,
+    executionKey: `ek-${connectionId}`,
+    modelStr: model,
+    provider,
+    providerId: null,
+    connectionId,
+    weight: 1,
+    label: null,
+  };
 }
 
 test("pinned connection is preferred but siblings are included as fallback", () => {
--- a/tests/unit/native-codex-turn-pin-10379.test.ts
+++ b/tests/unit/native-codex-turn-pin-10379.test.ts
@@ -1,0 +1,139 @@
+// #10379: Native Codex turn pin must allow fill-first failover across
+// compatible connections for the same provider + model.
+import { test } from "node:test";
+import assert from "node:assert/strict";
+
+const {
+  applyNativeCodexTurnPin,
+  pinNativeCodexTurn,
+  getNativeCodexTurnPin,
+  clearNativeCodexTurnPinsForTests,
+} = await import("../../open-sse/services/combo/nativeCodexTurnPin.ts");
+
+const BODY = {
+  client_metadata: {
+    "x-codex-turn-metadata": JSON.stringify({ thread_id: "t1", turn_id: "turn1" }),
+  },
+};
+
+function makeTarget(connectionId: string, model = "gpt-5.6-sol", provider = "codex") {
+  return { connectionId, modelStr: model, provider, executionKey: `ek-${connectionId}` } as any;
+}
+
+test("pinned connection is preferred but siblings are included as fallback", () => {
+  clearNativeCodexTurnPinsForTests();
+  pinNativeCodexTurn({
+    body: BODY,
+    comboName: "test-combo",
+    target: makeTarget("conn-1"),
+    connectionId: "conn-1",
+  });
+
+  const pin = getNativeCodexTurnPin(BODY, "test-combo")!;
+  const targets = [makeTarget("conn-1"), makeTarget("conn-2"), makeTarget("conn-3")];
+  const result = applyNativeCodexTurnPin(targets, pin);
+
+  assert.equal(result.length, 3, "all compatible connections should be returned");
+  assert.equal(result[0].connectionId, "conn-1", "pinned connection should be first");
+  assert.deepEqual(result[0].allowedConnectionIds, ["conn-1", "conn-2", "conn-3"]);
+});
+
+test("fallback connections share the same allowedConnectionIds", () => {
+  clearNativeCodexTurnPinsForTests();
+  pinNativeCodexTurn({
+    body: BODY,
+    comboName: "test-combo",
+    target: makeTarget("conn-2"),
+    connectionId: "conn-2",
+  });
+
+  const pin = getNativeCodexTurnPin(BODY, "test-combo")!;
+  const targets = [makeTarget("conn-1"), makeTarget("conn-2"), makeTarget("conn-3")];
+  const result = applyNativeCodexTurnPin(targets, pin);
+
+  assert.equal(result[0].connectionId, "conn-2");
+  assert.equal(result[1].connectionId, "conn-1");
+  assert.equal(result[2].connectionId, "conn-3");
+  for (const t of result) {
+    assert.deepEqual(t.allowedConnectionIds, ["conn-1", "conn-2", "conn-3"]);
+  }
+});
+
+test("incompatible targets (different provider/model) are excluded", () => {
+  clearNativeCodexTurnPinsForTests();
+  pinNativeCodexTurn({
+    body: BODY,
+    comboName: "test-combo",
+    target: makeTarget("conn-1"),
+    connectionId: "conn-1",
+  });
+
+  const pin = getNativeCodexTurnPin(BODY, "test-combo")!;
+  const targets = [
+    makeTarget("conn-1"),
+    makeTarget("conn-2"),
+    makeTarget("conn-other", "different-model", "other-provider"),
+  ];
+  const result = applyNativeCodexTurnPin(targets, pin);
+
+  assert.equal(result.length, 2, "incompatible target should be excluded");
+  assert.deepEqual(result[0].allowedConnectionIds, ["conn-1", "conn-2"]);
+});
+
+test("empty result when no compatible targets exist", () => {
+  clearNativeCodexTurnPinsForTests();
+  pinNativeCodexTurn({
+    body: BODY,
+    comboName: "test-combo",
+    target: makeTarget("conn-1"),
+    connectionId: "conn-1",
+  });
+
+  const pin = getNativeCodexTurnPin(BODY, "test-combo")!;
+  const targets = [makeTarget("conn-x", "other-model", "other-provider")];
+  const result = applyNativeCodexTurnPin(targets, pin);
+
+  assert.equal(result.length, 0);
+});
+
+test("pinNativeCodexTurn allows connectionId change for same provider+model", () => {
+  clearNativeCodexTurnPinsForTests();
+  pinNativeCodexTurn({
+    body: BODY,
+    comboName: "test-combo",
+    target: makeTarget("conn-1"),
+    connectionId: "conn-1",
+  });
+
+  // Should NOT throw when only connectionId changes
+  pinNativeCodexTurn({
+    body: BODY,
+    comboName: "test-combo",
+    target: makeTarget("conn-2"),
+    connectionId: "conn-2",
+  });
+
+  const pin = getNativeCodexTurnPin(BODY, "test-combo")!;
+  assert.equal(pin.connectionId, "conn-2", "pin should update to new connection");
+});
+
+test("pinNativeCodexTurn rejects provider/model change", () => {
+  clearNativeCodexTurnPinsForTests();
+  pinNativeCodexTurn({
+    body: BODY,
+    comboName: "test-combo",
+    target: makeTarget("conn-1"),
+    connectionId: "conn-1",
+  });
+
+  assert.throws(
+    () =>
+      pinNativeCodexTurn({
+        body: BODY,
+        comboName: "test-combo",
+        target: makeTarget("conn-1", "different-model", "codex"),
+        connectionId: "conn-1",
+      }),
+    /Native Codex turn target changed/
+  );
+});


diegosouzapw:
Thanks for this — the Codex native turn-pin failover fix for #10379 (item 2: allow fill-first fallover across sibling connections for the same provider/model instead of narrowing to a single pinned connection) is a real, still-open regression and looks solid, with good test coverage.

However, items 1 and 3 are now duplicates of work that already merged into release/v3.8.50 after this branch was created:
- #10310 (Codex quota header forwarding budget) was already fixed by a different, more complete implementation in commit ea66cc2 (merged 2026-08-15), which also adds warn-dedup (#10315).
- #10215 (Cursor kv_after_text truncating tool calls) was already fixed by #10502 (merged this morning, 2026-08-17).

That's exactly why this branch is now showing as conflicting and why CI fails: tests/unit/cursor-streaming.test.ts regresses to the old (pre-#10502) behavior, and the ESLint gate fails on a new `as any` cast in the new turn-pin test (no-explicit-any is an error under tests/).

Could you rebase onto the latest release/v3.8.50, drop the cursor.ts and responseHeaders.ts hunks (and their now-superseded test files) so only the nativeCodexTurnPin.ts fix for #10379 remains, and swap the `as any` cast for a properly typed target literal? That should get this green and mergeable.

Issue #7860, opened 2026-07-20
feat(providers): add OpenRouter as a speech-to-text / audio-transcription provider


Tasogarre:
## Problem / Use Case

OmniRoute proxies OpenRouter for chat and completions, but the transcription endpoint (`POST /v1/audio/transcriptions`) only accepts a fixed provider set (openai, groq, deepgram, assemblyai, nvidia, huggingface, qwen). OpenRouter now serves a growing catalog of speech-to-text models (Deepgram, Microsoft, NVIDIA, Mistral, Qwen, Google, OpenAI) behind a single connection, and there is no way to reach any of them through OmniRoute today. A request for an `openrouter/...` transcription model is rejected by the provider validator with a 422.

Users who already hold an OpenRouter connection have to stand up a second, provider-specific credential (or fall back to a direct integration) just to evaluate or route STT traffic, which defeats the point of routing through OmniRoute.

## Proposed Solution

Register OpenRouter as a transcription provider in the audio registry, reusing the existing OpenRouter connection and credential: no new provider identity, no second API key. Requests keep using the standard OpenAI-compatible endpoint:

```
POST /v1/audio/transcriptions
model=openrouter/openai/whisper-large-v3
```

OmniRoute converts the multipart upload into OpenRouter's JSON `input_audio` transcription format and proxies the response back unchanged, consistent with how the other audio providers are handled.

## Acceptance Criteria

- [ ] `POST /v1/audio/transcriptions` accepts `openrouter/<provider>/<model>` IDs and routes them to OpenRouter instead of returning a validation error.
- [ ] The provider reuses the configured OpenRouter connection; no separate STT credential is introduced.
- [ ] OpenRouter appears in the Speech-to-Text section of the dashboard, consistent with existing STT provider cards, and lists only transcription models (not the full chat catalog).
- [ ] Generic OpenRouter model discovery, model sync/import, and `/v1/models` are unaffected (chat/LLM catalog unchanged).
- [ ] Unit tests cover model-ID resolution and the multipart to `input_audio` conversion.

## Alternatives Considered

Adding a second, OpenRouter-specific STT credential/provider entry. Rejected: it duplicates the existing OpenRouter connection, creates competing configuration state, and is inconsistent with how OmniRoute already models OpenRouter as one connection serving multiple capabilities.

## Area

- Provider Support

## Additional Context

[...]

PR #10584, opened 2026-08-17
fix(audio): fall back nested STT models when the prefix provider has no credentials

### LINKED ISSUE #10583: STT: bare vendor/model ids require native credentials even when a gateway lists the same model
## Summary

`POST /v1/audio/transcriptions` resolves the provider from the first path segment of `model`. A request for `deepgram/nova-3` is treated as native Deepgram. If that provider has no API key, the route returns HTTP 400 `No credentials for provider: deepgram`.

The same instance can succeed with `openrouter/deepgram/nova-3` when OpenRouter is configured, because OpenRouter's catalog id is the nested string `deepgram/nova-3`.

Docs and examples still show `-F model=deepgram/nova-3` without saying that this is the native Deepgram route.

The dashboard STT playground already qualifies OpenRouter model ids to avoid this misroute. The public API and docs do not.

## Repro

Self-hosted OmniRoute with an OpenRouter key and **no** Deepgram key.

```bash
# tiny wav
python3 -c 'import wave,math,struct; w=wave.open("/tmp/tone.wav","wb"); w.setnchannels(1); w.setsampwidth(2); w.setframerate(16000); w.writeframes(b"".join(struct.pack("<h", int(4000*math.sin(2*math.pi*440*i/16000))) for i in range(4000))); w.close()'

curl -sS -o /tmp/stt-native.json -w "%{http_code}\n" \
  -X POST http://localhost:20128/v1/audio/transcriptions \
  -H "Authorization: Bearer \$OMNIROUTE_API_KEY" \
  -F "file=@/tmp/tone.wav" \
  -F "model=deepgram/nova-3"
# expected today: 400  {"error":{"message":"No credentials for provider: deepgram"}}

curl -sS -o /tmp/stt-or.json -w "%{http_code}\n" \
  -X POST http://localhost:20128/v1/audio/transcriptions \
  -H "Authorization: Bearer \$OMNIROUTE_API_KEY" \
  -F "file=@/tmp/tone.wav" \
  -F "model=openrouter/deepgram/nova-3"
# expected today: 200 when OpenRouter is configured
```

## Expected

One of:

1. If the prefix-matched provider has no credentials, retry other registry providers that list the original model id (OpenRouter lists `deepgram/nova-3`).
2. Or keep the 400, but mention the qualified catalog ids in the error (`openrouter/deepgram/nova-3`).
3. Docs should not present `deepgram/nova-3` as the default example unless a native Deepgram key is required.

Native Deepgram credentials should still win when they exist.

## Related

- OpenRouter STT provider: #7860
- Dashboard already prefixes OpenRouter STT ids in `SttExampleCard`

RaviTharuma:
## Summary

- `POST /v1/audio/transcriptions` prefix-matches `deepgram/nova-3` to native Deepgram. With no Deepgram key that is HTTP 400 `No credentials for provider: deepgram`, even when OpenRouter lists the same nested id.
- If the prefix-matched provider has no credentials, retry other registry providers that list the original model id (OpenRouter → `deepgram/nova-3`). Native Deepgram still wins when it is configured.
- The 400 now mentions qualified catalog ids. Docs no longer present bare `deepgram/nova-3` as the default example.

## Related Issues

- Closes #10583
- Related to #7860

## Validation

- [x] Focused tests: `DISABLE_SQLITE_AUTO_BACKUP=true node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test tests/unit/audio-nested-model-credential-fallback.test.ts` (6 pass)
- [ ] `npm run lint` (CI)

## Tests Added Or Updated

- `tests/unit/audio-nested-model-credential-fallback.test.ts`

## Coverage Notes

- New helpers in `open-sse/config/audioRegistry.ts` and the transcription route fallback are covered by the new unit file (parse, alias candidates, OpenRouter fallback, hint message).

## Reviewer Notes

- Fallback only runs after the prefix-matched provider returns no credentials. It does not change routing when native Deepgram (or any other prefix provider) is configured.
- English docs only; i18n copies still show the old example until the next translation sync.

### DIFF
--- a/changelog.d/fixes/10583-stt-nested-model-credential-fallback.md
+++ b/changelog.d/fixes/10583-stt-nested-model-credential-fallback.md
@@ -1,0 +1,1 @@
+- **fix(audio):** when a prefix-matched STT provider has no credentials, retry gateways that list the same nested model id (e.g. `deepgram/nova-3` → `openrouter/deepgram/nova-3`) and mention those ids in the 400; stop documenting bare `deepgram/nova-3` as the default example ([#10583](https://github.com/diegosouzapw/OmniRoute/issues/10583))
--- a/docs/guides/TROUBLESHOOTING.md
+++ b/docs/guides/TROUBLESHOOTING.md
@@ -478,7 +478,7 @@ If a provider repeatedly enters OPEN state:
 
 ### "Unsupported model" error
 
-- Ensure you're using the correct prefix: `deepgram/nova-3` or `assemblyai/best`
+- Use a model id whose first segment is a provider you have credentials for (`openai/whisper-1`, `openrouter/deepgram/nova-3`). Bare `deepgram/nova-3` requires a native Deepgram key.
 - Verify the provider is connected in **Dashboard → Providers**
 
 ### Transcription returns empty or fails
--- a/docs/guides/USER_GUIDE.md
+++ b/docs/guides/USER_GUIDE.md
@@ -948,8 +948,11 @@ Content-Type: multipart/form-data
 curl -X POST http://localhost:20128/v1/audio/transcriptions \
   -H "Authorization: Bearer your-api-key" \
   -F "file=@audio.mp3" \
-  -F "model=deepgram/nova-3"
+  -F "model=openai/whisper-1"
 ```
+
+`deepgram/nova-3` is the native Deepgram route and needs a Deepgram API key.
+If only OpenRouter is configured, use `openrouter/deepgram/nova-3`.
 
 **Speech-to-Text (transcription)** providers:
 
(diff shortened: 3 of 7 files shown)


diegosouzapw:
Good fix — the fallback logic (retry a gateway that lists the same nested model id when the prefix-matched native provider has no credentials) is clean and well tested; I ran the 6 new unit tests in an isolated worktree and they all pass, and eslint is clean on the touched files. One blocking housekeeping item before merge: this PR is opened against `main`, but `main` has diverged significantly from the active release branch (`release/v3.8.50`) — please retarget the PR base to `release/v3.8.50`. The 3 red CI checks you're seeing (`Build`, `Integration Tests`, `Quality Ratchet`) look like stale-base/infra artifacts from running against the outdated `main` tip (runner shutdown, a server-boot timeout unrelated to audio/STT, and a ratchet baseline digest mismatch) rather than defects in your diff — they should be re-evaluated once retargeted.

PR #10603, opened 2026-08-17
fix(providers): make upstream model sync opt-in and preserve manual overrides

InkshadeWoods:
## Summary

- Add an “Auto-fetch upstream models” control to provider detail pages and keep it disabled by default.
- Prevent automatic upstream model synchronization unless the provider explicitly enables it.
- Preserve same-ID manual models as user-configurable overrides of synced upstream models, including capabilities such as vision support and API format.
- Add an action to restore a model’s upstream defaults, plus English source strings and regression coverage.

## Related Issues

- Closes #
- Related to #

## Validation

- [x] Change type: provider / UI / i18n
  - `node --import tsx/esm --test tests/unit/managed-model-import.test.ts` — 9 passed
  - `node --import tsx/esm --test --test-name-pattern "keeps upstream discovery off until autoFetchModels is explicitly enabled" tests/unit/provider-models-route.test.ts` — 1 passed
  - Focused Vitest provider UI/hook tests — 14 passed
  - Focused ESLint for changed test files — passed
- [ ] `npm run lint` — the full command reports pre-existing unrelated lint errors in the current release base; the changed test files pass focused ESLint.

## Tests Added Or Updated

- `tests/unit/managed-model-import.test.ts`
- `tests/unit/provider-models-route.test.ts`
- `tests/unit/ui/use-provider-models-auto-fetch.test.tsx`
- `src/app/(dashboard)/dashboard/providers/[id]/__tests__/useModelImportHandlers.test.tsx`
- `src/app/(dashboard)/dashboard/providers/[id]/__tests__/providerModelsSectionVisibilityKey.test.tsx`

## Coverage Notes

- The provider-model route test covers the default-off behavior.
- The managed-model import test covers same-ID manual model metadata preservation.
- The UI/hook tests cover the detail-page toggle and ensure synchronization occurs only when `autoFetchModels === true`.

## Reviewer Notes

- No migration or Docker/build workflow change is included.
- Existing providers and custom providers both default to disabled automatic upstream model fetching.
- Users can enable the behavior from the provider detail UI and can restore an overridden model to upstream defaults when needed.


diegosouzapw:
This is a nice, well-scoped feature — the default-off opt-in for upstream auto-fetch plus the same-id manual-override/reset flow is exactly the right design, and your own new tests (24) all pass for me locally.

However the CI run shows the default flip breaks two pre-existing tests that assumed auto-fetch reached the network by default:
- `tests/unit/provider-models-route-lan-guard.test.ts` (the SSRF-guard test for LAN model-list fetches) now short-circuits before reaching the guarded fetch, because its connection fixtures don't set `autoFetchModels: true`.
- `tests/unit/openrouter-embeddings-catalog-6976.test.ts` (live-vs-curated dedup) has the same issue.

Could you add `autoFetchModels: true` to those fixtures' `providerSpecificData` so they keep exercising the paths they were written to test?

Also, the 11 new i18n keys (`providers.autoFetchModels*`, `overridesUpstreamModel*`, `resetToUpstreamDefaults*`) only made it into `en.json` and `zh-CN.json` — the i18n-completeness CI gate is failing for the other locales, so a sync pass is needed there too.

Separately, I saw a 'Kimi Coding' model-list assertion fail in CI — that looks like it's from catalog data drift unrelated to your diff (the current release/v3.8.50 tip's test still expects the same fixture your CI run expected), so a rebase should clear it up; flagging just in case.

PR #10605, opened 2026-08-17
fix(gemini): inject missing items schema for array tool parameters (#10578)

### LINKED ISSUE #10578: [BUG] Gemini-routed requests fail with "function_declarations[...].items: missing field" when many MCP tools are registered
### OmniRoute Version

Installation Method: npm install -g omniroute 

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 10 Pro 22H2 (Build 19045.6466)

### Node.js Version

24.18.1

### Provider(s) Involved

gemini

### Model(s) Involved

gemini/gemini-3.1-flash-lite, gemini/gemini-3-flash-preview, gemini/gemini-3.1-flash-tts-preview

### Client Tool

Claude Code (terminal CLI v2.1.227, via ANTHROPIC_BASE_URL pointed at OmniRoute)

### Description

When a Claude Code session has many MCP servers/tools registered (30+ tool declarations active), a request routed through OmniRoute to a Gemini-based provider fails with a 400 error. Gemini's API enforces that array-typed tool parameters must include an `items` schema field, but one of the registered tools' `params` property (array-typed) is missing this field. OmniRoute forwards the tool schema as-is without sanitizing/patching it for Gemini's stricter validation, causing the request to fail outright rather than falling back gracefully.

### Steps to Reproduce

1. Configure Claude Code CLI with ANTHROPIC_BASE_URL pointing to a local OmniRoute instance.
2. Install multiple MCP servers/plugins so the session has 30+ registered tools (in my case this included a ZoomInfo-related MCP server).
3. Set model to `auto` or `auto/best-coding` so OmniRoute can route to any connected provider.
4. Send a request that triggers Claude Code to include the full tool list in the request.
5. Observe the request gets routed to a `gemini/*` model and fails.

### Expected Behavior

Either OmniRoute should sanitize/patch tool schemas to satisfy Gemini's stricter validation requirements before forwarding, or it should detect the incompatibility and skip Gemini-based providers for this request, falling back to a provider that accepts the schema as-is (e.g. Claude-flavored models).

### Actual Behavior

Request fails with: API Error: 400 * GenerateContentRequest.tools[0].function_declarations[37].parameters.properties[params].items: missing field.

Retrying with `auto` then cascades through ~30 more provider/model combinations, all of which also fail, ending in:
API error: 503 {"error":{"message":"Maximum combo retry limit reached","type":"server_error","code":"service_unavailable"}...}

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
API Error: 400 * GenerateContentRequest.tools[0].function_declarations[37].parameters.properties[params].items: missing field.

API error: 503 {"error":{"message":"Maximum combo retry limit reached","type":"server_error","code":"service_unavailable"},"diagnostics":{"poolSize":113,"attempted":10,"excluded":[],"attemptOrder":[{"provider":"cline","model":"cline/google/gemma-4-31b-it:free"},{"provider":"gemini","model":"gemini/gemini-3.1-flash-lite"},{"provider":"gemini","model":"gemini/gemini-3.1-flash-tts-preview"},{"provider":"t3-web","model":"t3-web/gpt-4o"},{"provider":"t3-web","model":"t3-web/claude-haiku-4"}],"terminalReason":"max_attempts_exceeded","recovery":{"action":"try-auto","next_step":"Every candidate in this combo failed. Switch to model: auto to let OmniRoute pick a working provider, or pick a different combo."}}}
```

### Screenshots

_No response_

### Additional Context

This occurred with multiple MCP plugins/servers active in the Claude Code session, including a business-data/CRM-related MCP server (likely the source of the malformed schema at index [37]). Happy to provide the specific tool's schema if it helps narrow down which MCP server is generating the invalid `params` property.

### Validation Plan

- Register a tool with an array-typed parameter missing `items` in its schema, then send a request routed to a gemini/* provider via OmniRoute — should not 400.
- Add a unit test asserting OmniRoute either sanitizes outgoing tool schemas for Gemini-format requests, or excludes Gemini-incompatible tools/providers from the routing pool for that request.

sadSanta-07:
Fixes #10578

### DIFF
--- a/open-sse/translator/helpers/geminiHelper.ts
+++ b/open-sse/translator/helpers/geminiHelper.ts
@@ -727,5 +727,33 @@ export function cleanJSONSchemaForAntigravity(schema: unknown): unknown {
 
   injectObjectType(cleaned);
 
+  // Phase 8: Ensure array types have an items schema (#10578).
+  // Gemini strictly requires array parameters to define their `items` schema.
+  // If an MCP tool defines an array but forgets the items, inject a safe default.
+  function ensureArrayItems(obj: unknown): void {
+    if (!obj || typeof obj !== "object") return;
+
+    if (Array.isArray(obj)) {
+      for (const item of obj) {
+        ensureArrayItems(item);
+      }
+      return;
+    }
+
+    const record = obj as JsonRecord;
+    if (record.type === "array" && !record.items) {
+      record.items = { type: "string" };
+    }
+
+    // Recurse into remaining values.
+    for (const value of Object.values(record)) {
+      if (value && typeof value === "object") {
+        ensureArrayItems(value);
+      }
+    }
+  }
+
+  ensureArrayItems(cleaned);
+
   return cleaned;
 }
--- a/tests/unit/gemini-array-items.test.ts
+++ b/tests/unit/gemini-array-items.test.ts
@@ -1,0 +1,47 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { cleanJSONSchemaForAntigravity } from "../../open-sse/translator/helpers/geminiHelper";
+
+type SchemaNode = {
+  type?: string;
+  properties?: Record<string, SchemaNode>;
+  items?: SchemaNode;
+  required?: string[];
+  [key: string]: unknown;
+};
+
+describe("Gemini array items sanitizer (#10578)", () => {
+  it("should inject a string items schema for array types that are missing it", () => {
+    const sloppySchema = {
+      type: "object",
+      properties: {
+        emails: {
+          type: "array",
+        },
+      },
+      required: ["emails"],
+    };
+
+    const cleanedSchema = cleanJSONSchemaForAntigravity(sloppySchema as unknown) as SchemaNode;
+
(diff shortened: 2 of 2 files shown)


diegosouzapw:
The fix itself is right where it needs to be — cleanJSONSchemaForAntigravity() runs on the general OpenAI/Claude -> Gemini tool-translation paths, not just a niche variant, and the injected-items approach matches the file's existing Phase 6/7 pattern, so this does address #10578's root cause. One thing needs fixing before merge though: tests/unit/gemini-array-items.test.ts imports describe/it/expect from 'vitest', but tests/unit/*.test.ts is run by the Node native test runner (npm run test:unit), not vitest — as written it crashes with `TypeError: Cannot read properties of undefined (reading 'config')` instead of actually testing anything. Could you rewrite it with node:test + node:assert/strict (tests/unit/audio-transcription-opus-filename.test.ts is a good reference) and drop the two `as any` casts (ESLint's no-explicit-any is an error in tests/)? Re-run npm run test:unit + npm run lint locally and this should be ready to go.

PR #10162, opened 2026-08-12
[defer] fix(routing): keep approximate context estimates advisory

xz-dev:
## Summary

- Remove Combo pre-dispatch `400 context_length_exceeded` responses that treated approximate gateway token estimates as authoritative.
- Keep context metadata advisory for target ordering, including restoring Auto's eligible pool when every known limit looks too small.
- Preserve hard capability/output-token checks, concrete-target context handling, heterogeneous larger-context fallback, and homogeneous upstream `context_length_exceeded` short-circuiting.

## Related Issues

- Closes #10141

## Validation

Choose the change type and focused loop from the
[Contribution Golden Path](../docs/dev/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite,
Vitest, the 60% coverage gate, and the production build all run in CI on this PR (#8329):

- [x] Change type: routing
- [x] Focused tests and category gates from the golden path
- [ ] `npm run lint` — changed files pass; the full command is base-red in three unchanged files, documented below
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented below

Focused validation:

- 54/54 tests pass across capability filtering, target resolution, priority/round-robin/Auto advisory context handling, homogeneous and heterogeneous upstream overflow behavior, and the #8841 reproduction.
- Prettier check passes for all changed non-deleted files.
- ESLint passes for all changed non-deleted files.
- `git diff --check` passes.
- `npm run typecheck:core` reports 12 pre-existing diagnostics in unchanged files and none in changed files.

## Tests Added Or Updated

- `tests/unit/8488-capability-filter-fail-closed.test.ts`
- `tests/unit/combo-context-window-filter.test.ts`
- `tests/unit/combo-target-resolution-split.test.ts`
- `tests/unit/repro-8841-context-overflow-opencode.test.ts`

## Coverage Notes

The updated tests exercise the public `handleComboChat()` seam for priority, round-robin, and Auto dispatch when approximate estimates exceed every catalog hint. They also verify that a real upstream overflow still short-circuits a homogeneous same-model pool while a heterogeneous pool may try a larger-context target.

## Reviewer Notes

- This is intentionally a Draft because #10141 is design-first and requests maintainer agreement on exact/estimated/upstream-managed semantics.
- Hard tool, vision, structured-output, and requested output-token compatibility rules remain unchanged.
[...]

Issue #10141, opened 2026-08-12
enhancement(routing): RFC — separate approximate token estimates from hard admission decisions


xz-dev:
## Problem / Use Case

OmniRoute currently uses a deliberately rough token estimate (`serialized JSON length / 4`, plus a bounded image estimate) in several context-fit decisions. That estimate is useful for compression thresholds and target ordering, but some call sites promote it into an authoritative admission decision:

- `handleChatCore()` can return `400 context_length_exceeded` before any upstream request.
- `enforceOutputTokenBudget()` can reduce client output-token fields based on `contextLimit - estimatedInputTokens`.
- the optional Layer-A capability filter can reject a concrete target before dispatch.
- the auto-strategy context pre-filter can return 400 when all candidates fall below the estimate.

This can produce both false positives and false negatives because the gateway estimate is not the target model's tokenizer. Accuracy varies with language, code, tool schemas, structured history, image handling, and source-to-target translation.

A concrete failure exposed the ownership problem: a native Responses/Codex request was compared against a valid effective context budget using OmniRoute's approximate count and rejected locally, although this lane intentionally leaves exact counting and context management to the Codex client/upstream. The numeric limit was valid; using an approximate gateway count as the final authority was not.

Related prior work:

- #8942 / #8944 correctly made catalog context metadata advisory in the general Combo compatibility filter: prefer known-fitting targets, retain context-only mismatches for runtime fallback, and let concrete upstream responses decide.
- #8368, #8560, and #8594 show that estimation-shape differences have already caused false context decisions and lossy compression.
- #7379 introduced the final pre-translation budget gate to prevent invalid derived output budgets, but it did not establish a tokenizer-accuracy or context-ownership contract.

This RFC asks whether the advisory principle from #8942 should become the gateway-wide token-admission contract.

## Proposed Solution

Separate three concepts that are currently conflated:

1. **Limit authority** — whether `context_length`, `max_input_tokens`, or `max_output_tokens` is authoritative metadata.
2. **Count confidence** — whether input tokens are `exact` or merely `estimated`.
3. **Context ownership** — whether OmniRoute manages context or the client/upstream does (`upstream-managed`).

Suggested behavior:

| Count / ownership | Routing and compression | Hard 400 | Rewrite output-token request |
| --- | --- | --- | --- |
| Exact count + authoritative limit | Yes | Allowed | Allowed within documented limits |
| Estimated count | Advisory only | No | Do not shrink from estimated remaining context |
[...]

PR #10606, opened 2026-08-17
feat(routing): add context-window check bypass flag

JxnLexn:
## Summary

- add the default-off, dangerous opt-in `DISABLE_CONTEXT_WINDOW_CHECKS` feature flag
- bypass OmniRoute's local `contextWindow` and `maxInputTokens` rejection/filtering for direct requests and all combo-routing paths when enabled
- keep prompt compression, model output-token caps, tool/vision/structured-output checks, and explicit combo context preferences active
- preserve the current native Codex Responses and compression-deferral behavior when the flag is disabled
- document that upstream providers still enforce their actual context limits
- expose the flag through the existing dashboard feature-flag registry with localized labels and descriptions

## Behavior

The flag does not increase any model's real context window. It only prevents OmniRoute from rejecting or filtering a request based on local catalog metadata, allowing the selected upstream provider to enforce its actual limit.

Resolution remains:

1. database feature-flag override
2. environment variable
3. default value (`false`)

No restart is required.

When enabled, the bypass intentionally runs before the compression-defer logic added by #10503. This means the request is not locally rejected or filtered on context metadata; compression remains available later in the normal request pipeline, but the early combo overflow decision is skipped.

## Validation

Rebased onto `release/v3.8.50` at `bc6129bcb` on August 20, 2026.

Passed locally:

- affected unit tests: 89 passed
- `npm run typecheck:core`
- `npm run lint`
- `git diff --check origin/release/v3.8.50...HEAD`
- `npm run i18n:check-ui-coverage`
- `npm run i18n:check-value-drift`
- `npm run i18n:check-glossary`

`npm run check:docs-all` currently reaches `check:docs-counts` and fails on pre-existing base-branch provider-reference/SVG drift: live provider modules report 343 while generated documentation and diagrams still report 342. This PR does not modify provider registration, provider-reference generation, or those SVG claims.

The previous tool-handoff and documentation-sync commits are no longer part of this PR: the tool-handoff fix is already upstream, and the stale documentation-sync changes were dropped during the rebase.


diegosouzapw:
This is a clean, fail-safe implementation of the bypass flag — good defaults (off, DB > env > default resolution, error-safe fallback to checks-enabled), and it correctly keeps output-token caps and prompt compression active even when context-window checks are bypassed. The PR's own CI run is fully green; the CONFLICTING state showing on the PR page looks like stale doc-sync drift (the base branch has since picked up the same migration-count/PROXY_LOG_INCLUDE_IPS doc updates from elsewhere) rather than a real conflict — a rebase onto the latest release/v3.8.50 should resolve it cleanly. One note for awareness: this flag's bypass check in getKnownContextOverflow() runs before the not-yet-merged #10503's compression-defer logic, so enabling this dangerous flag will also skip that fix's benefit for the same request — that's consistent with the flag's stated scope, just flagging it so it's a documented interaction rather than a surprise.

diegosouzapw:
Thanks for this — closing in favor of a rescoped follow-up that preserves your work with credit: #10927 (merged).

Context: while reviewing this PR, we found that #10162 (merged the same day as your last rebase, per owner RFC #10141) had already deleted `knownContextOverflow.ts` and made the combo-routing context/window filter **advisory-only** — it never hard-rejects a request on local context estimates anymore. That made the combo-routing half of this PR's diff (`knownContextOverflow.ts`, `resolveAutoStrategy.ts`, `comboStructure.ts`, `capabilityFilter.ts`) moot.

The other half — the `chatCore.ts` bypass for the surviving `enforceOutputTokenBudget` hard-reject on direct single-model requests — was untouched by #10162 and still has real value, so we carried that part forward (feature flag, helper, docs, i18n, tests) into #10927, which is now merged with your `Co-authored-by` trailer.

Appreciate the contribution — the underlying idea (let operators opt out of OmniRoute's local context-size gate and trust the upstream provider) was sound, it just needed to be rescoped to the architecture as it stands today.

PR #10607, opened 2026-08-17
fix(api): accept .opus uploads on /v1/audio/transcriptions

### LINKED ISSUE #10588: fix(audio): /v1/audio/transcriptions rejects .opus filenames that are valid Ogg/Opus
### OmniRoute Version

3.8.49 (Docker image `diegosouzapw/omniroute`)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Provider(s) Involved

OpenAI Whisper (and any STT provider that uses the OpenAI filename allow-list)

### Model(s) Involved

`openai/whisper-1`

### Client Tool

OpenClaw Talk, Telegram voice notes, any client that uploads OmniRoute `response_format=opus` output back into `/v1/audio/transcriptions`

### Description

`POST /v1/audio/transcriptions` decides file type from the **upload filename**, then forwards it to Whisper-compatible providers. The same audio bytes succeed as `note.ogg` / `note.mp3` and fail as `note.opus`.

OmniRoute does not rewrite the filename (`open-sse/handlers/audioTranscription.ts` `getUploadedFileName` / `buildMultipartBody`). OpenAI’s Whisper allow-list is `flac, m4a, mp3, mp4, mpeg, mpga, oga, ogg, wav, webm` — no `opus`. Meanwhile `/v1/audio/speech` with `response_format=opus` returns `audio/opus` bytes that clients naturally save as `.opus`.

The Deepgram path already maps `.opus` → `audio/opus` in `resolveAudioContentType`. The OpenAI-compat multipart path does not.

### Steps to Reproduce

1. Generate speech with `openai/tts-1` and `response_format=opus` (or use any Ogg/Opus voice note).
2. `POST /v1/audio/transcriptions` with `model=openai/whisper-1` and filename `t.ogg` — 200.
3. Upload the **same bytes** as `t.opus` — 400.

### Expected Behavior

The same audio bytes uploaded as `note.opus` and `note.ogg` both return 200 with `openai/whisper-1`. Prefer remapping `.opus` → `.ogg`/`.oga` (or sniffing Ogg pages) before the upstream multipart request.

### Actual Behavior

```
400 Invalid file format. Supported formats: ['flac', 'm4a', 'mp3', 'mp4', 'mpeg', 'mpga', 'oga', 'ogg', 'wav', 'webm']
```

| filename | Result |
|---|---|
| `t.mp3` | 200 transcript |
| `t.ogg` | 200 transcript |
| `t.opus` | **400** |

### Test Impact

Needs a new unit test

### Additional Context

Related: `response_format=ogg` is rejected on speech (filing separately). Together, the two audio routes disagree: speech emits `.opus`, transcriptions refuse `.opus`.

### Validation Plan

- Unit test on `buildMultipartBody` / OpenAI transcription path: `.opus` is rewritten to an allow-listed name/content-type.
- `flac` / `mp3` / `wav` behavior unchanged.

pucedoteth:
Fixes #10588.

## Problem

Whisper-compatible upstreams choose the decoder from the multipart **filename**, against an allow-list (`flac, m4a, mp3, mp4, mpeg, mpga, oga, ogg, wav, webm`) that has no `opus`. `getUploadedFileName` forwarded the client's filename verbatim, so byte-identical audio transcribed as `note.ogg` and 400'd as `note.opus`.

That closes a loop on this gateway: `/v1/audio/speech` emits `audio/opus` for `response_format=opus`, so any client that saves its own output and re-uploads it — OpenClaw Talk, Telegram voice notes — fails every round trip.

## Fix

A `.opus` file is Opus audio in an Ogg container (RFC 7845), so `.ogg` is a truthful relabel and is already on the allow-list. The rewrite goes in `getUploadedFileName`, the single choke point that feeds `buildMultipartBody`.

While confirming the fix I found the same root cause in the OpenRouter STT path, with a quieter symptom. `resolveOpenRouterAudioFormat` matched `.opus` against neither its extension list nor its MIME map, so it fell through to the `"wav"` default — announcing Opus bytes to the upstream as WAV rather than erroring. Mapped both the extension and `audio/opus` onto its already-supported `ogg`.

## Deliberately narrow

- The rewrite is anchored (`/\.opus$/i`), so `opus-demo.wav` and `take.opus.mp3` are untouched — `opus` elsewhere in a name is part of the name, not the container.
- The `audio.wav` fallback for nameless blobs is unchanged.
- The Deepgram path already handled `.opus` in `resolveAudioContentType`; it needed nothing.
- I left the multipart part's `Content-Type` alone. It still reflects the client's `file.type`, so an `audio/opus` upload now sends `filename="note.ogg"` with `Content-Type: audio/opus`. Whisper validates on the filename, so this is enough to fix the 400 — say the word if you would rather normalize the MIME too.

## Tests

New file `tests/unit/audio-transcription-opus-filename.test.ts`, 8 cases: the rename, case-insensitivity, the two anchoring guards, untouched extensions, the nameless-blob fallback, and three for the OpenRouter resolver.

- Against this branch: **8/8 pass**
- With only the two handlers reverted and the tests kept: **4 fail, 4 pass** — the four that fail are exactly the `.opus` cases, and the four guard-rail tests pass either way, which is what they are for.
- Neighbouring suites green: `audio-transcription-handler`, `audio-transcription-openrouter`, `multipart-body-arraybuffer-backing`, `kie-audio-transcription-result`, `speechmatics-transcription-provider` — **41 pass, 0 fail** including the new file.
- `eslint` clean on all three changed files.

Per CONTRIBUTING I ran the focused loop rather than the full suite. One note for reproducing locally: a plain `npm install` fails here on the `opencode-ai` postinstall (`It seems your package manager failed to install the right opencode CLI package`), unrelated to this change — `npm install --ignore-scripts` gets you a working dev tree.

### DIFF
--- a/open-sse/handlers/audioTranscription.ts
+++ b/open-sse/handlers/audioTranscription.ts
@@ -69,8 +69,24 @@ function isValidPathSegment(segment: string): boolean {
   return !segment.includes("..") && !segment.includes("//");
 }
 
+/**
+ * A `.opus` file is Opus audio in an Ogg container (RFC 7845) — the same bytes
+ * a client would otherwise name `.ogg`. Whisper-compatible upstreams pick the
+ * decoder from the *filename* and their allow-list
+ * (`flac, m4a, mp3, mp4, mpeg, mpga, oga, ogg, wav, webm`) has no `opus`, so
+ * `note.opus` 400s while byte-identical `note.ogg` succeeds. Since
+ * `/v1/audio/speech` emits `audio/opus` for `response_format=opus`, clients
+ * round-tripping their own voice notes hit this constantly. Relabel to the
+ * container that actually describes the bytes.
+ */
+function normalizeUploadExtension(name: string): string {
+  return name.replace(/\.opus$/i, ".ogg");
+}
+
 function getUploadedFileName(file: Blob & { name?: unknown }): string {
-  return typeof file.name === "string" && file.name.length > 0 ? file.name : "audio.wav";
+  return typeof file.name === "string" && file.name.length > 0
+    ? normalizeUploadExtension(file.name)
+    : "audio.wav";
 }
 
 /**
--- a/open-sse/handlers/openrouterTranscription.ts
+++ b/open-sse/handlers/openrouterTranscription.ts
@@ -21,6 +21,10 @@ import { upstreamErrorResponse } from "./audioTranscription.ts";
 export function resolveOpenRouterAudioFormat(file: Blob & { name?: unknown }): string {
   const fileName = typeof file.name === "string" ? file.name.toLowerCase() : "";
   const extension = fileName.includes(".") ? fileName.split(".").pop() || "" : "";
+  // `.opus` is Ogg-encapsulated Opus (RFC 7845). Without this it matched
+  // neither the extension list nor the MIME map below and fell through to the
+  // "wav" default, so Opus bytes were announced to the upstream as WAV.
+  if (extension === "opus") return "ogg";
   if (["wav", "mp3", "flac", "m4a", "ogg", "webm", "aac"].includes(extension)) {
     return extension;
   }
@@ -33,6 +37,7 @@ export function resolveOpenRouterAudioFormat(file: Blob & { name?: unknown }): s
     "audio/x-flac": "flac",
     "audio/mp4": "m4a",
     "audio/ogg": "ogg",
+    "audio/opus": "ogg",
     "audio/webm": "webm",
     "audio/aac": "aac",
   };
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks for the clear write-up and the RFC 7845 rationale — this is a well-scoped fix. I re-ran your new test file locally (8/8 pass) and linted the three changed files (0 problems), so the code side is solid. The pending checks you're seeing are GitHub's first-time-contributor workflow-approval gate, not an actual failure (only semgrep-cloud-platform/scan has run so far, and it passed) — we'll approve those runs and get this merged once they're green.

Issue #9653, opened 2026-08-06
fix(backend): Completed SSE streams recorded as 499 client-disconnected with 0 tokens (finalization race)


hartmark:
## Summary

When a client closes its connection immediately after fully receiving a completed SSE stream (e.g. right after reading `response.completed`), OmniRoute can still record the request as `499 Client disconnected: request_signal_aborted` with **zero token usage** — even though the debug stream-chunk capture proves the complete, well-formed response was already delivered to the client.

This is a **race condition**, not a transport/translation bug: the actual bytes on the wire are correct and complete in every case checked.

## Evidence (two independent real production requests)

Both show the identical pattern — `call_logs` row persisted with:
- `status: 499`
- `error_summary: "Client disconnected: request_signal_aborted"`
- `tokens_in: 0, tokens_out: 0`
- `responseBody: {"error": {"message": "Client disconnected: request_signal_aborted", "type": "client_disconnected", "code": "client_disconnected"}}`

But `pipeline.streamChunks.client` (the raw bytes actually written to the client) ends cleanly and completely with, in order:
```
event: response.output_item.done
event: response.completed
data: {"type":"response.completed","response":{...,"status":"completed",...,"usage":{"input_tokens":129555,"output_tokens":6654,"total_tokens":136209,...}}}
```
— i.e. the full final event, with populated `usage`, was the last thing captured on the wire. Nothing was truncated or cut off mid-stream.

- Request A: duration 62126ms, full `response.completed` with usage present in the client stream capture, persisted as 499/0-tokens.
- Request B: duration 18924ms, full `response.completed` present in the client stream capture, persisted as 499/0-tokens.

Client behavior in both cases: closes the connection promptly once it has parsed what it needs from the stream (a normal, common client pattern — not a hang, not an early abandon mid-generation).

## Suspected root cause

`open-sse/utils/streamFailureFinalization.ts`'s `handleStreamFailure()` already guards against exactly this:

```ts
const handleStreamFailure = (failure) => {
  if (isStreamCompletionRecorded()) {
    return true; // already finalized as a real completion, don't overwrite
  }
  ...
```

`isStreamCompletionRecorded` is backed by a `streamCompletionRecorded` flag in `open-sse/handlers/chatCore.ts`, set inside `onStreamComplete` — which is invoked by the SSE transform stream (`open-sse/utils/stream.ts`) as it parses the terminal upstream event.

[...]

Issue #3697, opened 2026-06-12
[feature] Codex CLI compatibility shim: echo requested (effort-suffixed) model id in Responses payloads so the UI shows the active effort


diegosouzapw:
## Feature Request

**Source:** Discussion #2829 by @KernelBypass

## Problem

Routing Codex CLI through OmniRoute with effort-suffixed ids (`codex/gpt-5.5-xhigh`) works correctly upstream, but the Codex CLI UI (model button / status line) displays the wrong model+effort, because OmniRoute echoes the bare upstream model id (`gpt-5.5`) in Responses payloads instead of the id the client requested. The reporter finds the mismatch jarring ("makes user second-guess if the model is actually correct").

## Proposed Solution

Add a Codex-client compatibility shim (detected via originator/UA `codex_*`, same detection used by the `/v1/models` special-case from PR #3481): echo the **originally requested model id** (e.g. `gpt-5.5-xhigh` or `codex/gpt-5.5-xhigh`) in the `model` fields of Responses API payloads (`response.created` / `response.completed` SSE events and non-stream bodies), instead of the upstream-normalized id.

Reference implementation: [Soju06/codex-lb](https://github.com/Soju06/codex-lb) does this mapping and the Codex UI renders the active effort accurately.

### Implementation Ideas

- `open-sse/transformer/responsesTransformer.ts` (Responses API stream/non-stream shaping) - carry the client-requested model string through and substitute it in outgoing `model` fields for codex-originated requests.
- Constraint reminder: keep `/v1/models` returning an EMPTY `models: []` for Codex (populating it makes Codex swap its built-in base instructions - see PR #3481 notes). The shim must operate purely on response payloads.
- While in there, verify payload `reasoning.effort` passthrough for bare ids (`model=gpt-5.5` + `model_reasoning_effort=xhigh` from Codex config) - reporter believes it is not honored; the suffixed-id path remains the documented one either way.

## Current Workarounds

- Use `codex/gpt-5.5-xhigh` (correct upstream behavior); ignore the CLI model button.

## Additional Context

- Discussion: #2829
- Related: PR #3481 (codex `/v1/models` `models: []` shim), #2331 (effort max-of regression guard)

diegosouzapw:
✅ **Accepted — cataloged for implementation.**

[...]

PR #10608, opened 2026-08-17
fix(streaming): preserve completed Codex tool handoffs

JxnLexn:
## Summary

- detect fully completed Responses API `function_call` and `custom_tool_call` handoffs for identified Codex clients
- keep the upstream transform alive during the existing disconnect grace period so a trailing `response.completed` event and its usage can be recorded
- preserve the existing immediate-abort behavior for incomplete or mismatched tool calls, non-Codex clients, and ordinary disconnects
- retain the timeout fallback: if no completion arrives during the grace period, abort upstream and finalize the request as 499
- synchronize the documented migration count and the `PROXY_LOG_INCLUDE_IPS` environment contract required by the documentation gates

## Problem

The disconnect grace period added in #9653 delays persistence of a 499, but `createDisconnectAwareStream.cancel()` still immediately cancels the transform reader and aborts its writer. For Codex tool handoffs, this makes the completion that the grace handler is waiting for unreachable: transform `flush()` and `onStreamComplete()` cannot run after the client starts the next tool turn.

This was reproduced with a regression test against the unchanged `release/v3.8.50` implementation: a completed custom-tool handoff canceled the upstream stream and eventually persisted 499 even though a trailing `response.completed` frame was available.

## Implementation

For verified Codex Responses requests only, the stream handler watches forwarded SSE frames for a complete tool handoff. Qualification requires:

- a matching `response.function_call_arguments.done` or `response.custom_tool_call_input.done` event
- a matching completed `response.output_item.done`
- a call ID, tool name, matching full arguments or input, and a completed status when present

When the client disconnects after such a handoff, OmniRoute drains the upstream transform for the configured grace period instead of canceling it immediately. This allows the normal completion and usage bookkeeping path to finish. No new success path is synthesized.

## Tests

Passed:

- focused Node regression suite: 35 tests
- `npm run test:vitest`: 40 files, 362 tests
- `npm run typecheck:core`
- `npm run lint`
- `npm run check:file-size`
- `npm run check:any-budget:t11`
- `npm run check:docs-all`
- `git diff --check`

The complete `npm run test:unit` run still reports failures in unrelated, untouched areas, including Adobe Firefly integration timing, combo runtime concurrency, a macOS `/usr/bin/cp` assumption, and port PID detection. The new Codex streaming regression tests pass within that run. The previously reported documentation-count and environment-contract failures are fixed in this branch.

### DIFF
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -2850,6 +2850,8 @@ export async function handleChatCore({
     connectionId,
     clientResponseFormat,
     clientAbortSignal: clientRawRequest?.signal,
+    allowCompletedToolHandoffGrace: isCodexResponsesEcho,
+    clientDisconnectGracePeriodMs: STREAM_DISCONNECT_GRACE_PERIOD_MS,
   });
 
   const dedupRequestBody = { ...translatedBody, model: `${provider}/${model}`, stream };
--- a/open-sse/utils/responsesToolHandoff.ts
+++ b/open-sse/utils/responsesToolHandoff.ts
@@ -1,0 +1,132 @@
+type CompletedToolItem = {
+  keys: string[];
+  type: "function_call" | "custom_tool_call";
+  value: string;
+};
+
+function getResponsesEventKeys(
+  payload: Record<string, unknown>,
+  item?: Record<string, unknown>
+): string[] {
+  const keys = new Set<string>();
+  const addStringKey = (prefix: string, value: unknown) => {
+    if (typeof value === "string" && value.trim()) keys.add(`${prefix}:${value.trim()}`);
+  };
+  const addIndexKey = (value: unknown) => {
+    if (typeof value === "number" && Number.isInteger(value) && value >= 0) {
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Nicely scoped fix — the completed-tool-handoff watcher is correctly gated to verified Codex Responses clients only, with a bounded buffer/tracked-item cap, and the existing immediate-abort/timeout-fallback behavior is preserved for every other case. CI on the PR's own run is fully green; the CONFLICTING state on the PR page looks like the same stale doc-sync drift as #10606 (migration-count/PROXY_LOG_INCLUDE_IPS docs already match current base elsewhere) — a rebase should clear it with no logic changes needed. Two small non-blocking notes: the SSE-frame watcher currently parses every chunk for all Responses-format clients (not just Codex ones where the grace actually applies), which is a bit of unnecessary work bounded by the buffer cap; and reusing `isCodexResponsesEcho` (originally the #3697 model-echo flag) to gate this unrelated behavior works but might be worth a short comment or a dedicated name for readability.

PR #10612, opened 2026-08-17
fix(cli): derive machine-id token under plain Node and honor salt rotation

ataozkn:
## Problem

`getCliToken()` destructured `machineIdSync` off `await import("node-machine-id")`. That module is CommonJS, so under plain Node its exports land on `.default` and the destructured binding is `undefined`. Calling it throws, the bare `catch` blanks the token, and **every management request goes out with no `x-omniroute-cli-token` header** — silently unauthenticated.

Reproduced on Windows 11 / Node 24.13.0, `omniroute` 3.8.50 from source:

```
$ node -e "import('./bin/cli/utils/cliToken.mjs').then(m=>m.getCliToken()).then(t=>console.log(JSON.stringify(t)))"
""
```

Every CLI command that goes through the management API 401s:

```
$ omniroute combo list
Error: could not list combos from the server (HTTP 401)

$ omniroute usage budget list
Authentication required. Set OMNIROUTE_API_KEY or run: omniroute setup
```

Separately, `SALT` was hardcoded, so the rotation documented in [`docs/security/CLI_TOKEN.md`](docs/security/CLI_TOKEN.md) ("After rotation, all CLI processes on this machine will use the new token automatically") never reached the CLI.

## Fix

- Resolve the binding the same way [`src/lib/machineToken.ts`](src/lib/machineToken.ts) already does (`mod.machineIdSync ?? mod.default?.machineIdSync`).
- Read `OMNIROUTE_CLI_SALT`, mirroring `getActiveSalt()`, and key the cache on the active salt.
- The `catch` now emits a `console.debug` breadcrumb instead of failing mute, per the error-handling convention in `CONTRIBUTING.md` (this swallow changes control flow).

## Tests

`tests/unit/cli-machine-token.test.ts` asserted `token === "" || token.length === 32`, so the blanked token passed the suite.

Tightening it in-process is not enough either: the suite runs under `tsx/esm`, which resolves CJS named exports and **hides this bug**. The regression test therefore spawns plain `node` — the loader the CLI actually runs under.

Both new tests fail on `release/v3.8.50` and pass on this branch:

```
✖ getCliToken deriva token de 32 chars sob o node puro que a CLI usa
✖ getCliToken respeita rotação de OMNIROUTE_CLI_SALT
ℹ pass 11  ℹ fail 2
```

After:

```
ℹ pass 13  ℹ fail 0
```

## Scope

This fixes the CLI never *sending* a token. On my machine management calls still 401 with a correctly derived token supplied by hand, so there appears to be a second, server-side cause I have not root-caused — this PR deliberately does not claim to fix management auth end-to-end.

`node --import tsx/esm --test tests/unit/cli-machine-token.test.ts` — 13 passed. `npm run lint` clean on the changed files.


diegosouzapw:
Verified — this exactly matches the machineIdSync CJS-interop resolution already used in src/lib/machineToken.ts, and the salt rotation now honors what docs/security/CLI_TOKEN.md already promises. Ran the new tests on your branch: 13/13 pass, including the plain-node spawn test that correctly bypasses tsx/esm's CJS-interop masking. Ready to merge.

PR #10613, opened 2026-08-17
fix(cli): let setup --api-key reach the provider setup path

ataozkn:
## Problem

`bin/cli/program.mjs:26` declares a program-level `--api-key` (the OmniRoute *server* key) and `bin/cli/commands/setup.mjs:144` declares its own `--api-key` (the *provider* key). Commander binds the value to the program-level option, so the subcommand's `opts.apiKey` is always `undefined`.

The documented headless setup path is therefore unusable — it rejects the very flag that was just passed:

```
$ omniroute setup --non-interactive --add-provider \
    --provider openrouter --api-key sk-or-v1-...

OmniRoute Setup
✔ Admin password configured
✖ Provider API key is required. Pass --api-key or OMNIROUTE_API_KEY.
```

`OMNIROUTE_API_KEY` does not help either: it feeds the same program-level option, so it never reaches `resolveProviderInput()` at `setup.mjs:71`.

Minimal reproduction of the flag shape:

```js
const p = new Command();
p.addOption(new Option("--api-key <key>", "global").env("OMNIROUTE_API_KEY"));
p.command("setup").option("--api-key <value>", "Provider API key")
 .action((opts, cmd) => {
   console.log(opts.apiKey);                    // undefined
   console.log(cmd.optsWithGlobals().apiKey);   // "SECRET123"
 });
p.parse(["node", "x", "setup", "--api-key", "SECRET123"]);
```

Until this is fixed the only way to add a provider from a script is `omniroute keys add <provider> <key>`.

## Fix

Fall back to the program-level value in a small exported helper, `mergeSetupOptions()`. An explicit subcommand value still wins. This also makes `OMNIROUTE_API_KEY` satisfy the provider key — which the error message already promised.

## Verification

Against a throwaway `DATA_DIR`, the previously failing command now completes:

```
✔ Admin password configured
✔ Provider configured: OpenRouter
✔ Setup complete
Provider: openrouter (OpenRouter)
```

and the row lands in `provider_connections`.

## Tests

New `tests/unit/cli/setup-provider-api-key.test.ts` (4 tests) covers the real Commander flag shape, the `OMNIROUTE_API_KEY` path, subcommand-wins precedence, and that `output` still comes from the program level. All pass; `npm run lint` clean on the changed files.

### DIFF
--- a/bin/cli/commands/setup.mjs
+++ b/bin/cli/commands/setup.mjs
@@ -133,6 +133,29 @@ async function setupProvider(db, opts, prompt, nonInteractive) {
   return connection;
 }
 
+/**
+ * Merge the `setup` subcommand options with the program-level ones.
+ *
+ * The program declares a global `--api-key` (the OmniRoute *server* key, see
+ * bin/cli/program.mjs) and `setup` declares its own `--api-key` (the *provider*
+ * key). Commander binds the value to the program-level option, so the
+ * subcommand's `opts.apiKey` is always `undefined` and `--add-provider` failed
+ * with "Provider API key is required" even when `--api-key` was passed. Falling
+ * back to the global value also makes `OMNIROUTE_API_KEY` work, which the error
+ * message already told users to use.
+ *
+ * @param {Record<string, unknown>} opts Subcommand options.
+ * @param {Record<string, unknown>} globalOpts Result of `cmd.optsWithGlobals()`.
+ * @returns {Record<string, unknown>} Options to hand to `runSetupCommand`.
+ */
+export function mergeSetupOptions(opts, globalOpts) {
+  return {
+    ...opts,
+    apiKey: opts.apiKey ?? globalOpts.apiKey,
+    output: globalOpts.output,
+  };
+}
+
 export function registerSetup(program) {
   program
     .command("setup")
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Verified — this is a genuine Commander.js option-shadowing bug (program-level --api-key wins over the setup subcommand's own --api-key). mergeSetupOptions() is a clean, minimal fix and the new tests reproduce the real flag shape from program.mjs/setup.mjs. Ran the 4 new tests: all pass. No overlap with #10612 (different file). Ready to merge.

timm-u:
Hit this today on 3.8.49 (latest npm release, published Jul 30) while wiring up a provider non-interactively. That tarball predates this fix, so `setup --add-provider --api-key ...` still errors with "Provider API key is required". Any sense of when the next npm release will include this? Happy to test a prerelease in the meantime.

JoAllg:
As of CLI Version 3.8.50 the same issue affects `omniroute setup-claude` as well. Appending the `--api-key` argument does not work. However, I was able to do it with:
`export OMNIROUTE_API_KEY=sk-88...  omniroute setup-claude --dry-run
`

@ataozkn @diegosouzapw

Issue #10224, opened 2026-08-13
fix(auth): send OIDC `state` to IdP and use originEarly for callback error redirects


lexiJC:
### Summary

Dashboard OIDC login can never succeed. Two independent bugs in `src/app/api/auth/oidc/`:

1. `state` is generated and stored in the `oidc_state` cookie but **never added to the authorization URL**, so the IdP echoes back an empty `state` and the callback rejects the login.
2. Every error redirect in the callback is built from `reqUrlEarly` instead of the `originEarly` the same function computes for exactly that purpose, so the browser is sent to `https://0.0.0.0:<PORT>/login?...` — an unopenable page that hides bug 1.

Both are present on `main` today (checked 2026-08-13, v3.8.50 line).

### Bug 1 — `state` is never sent to the IdP

`src/app/api/auth/oidc/login/route.ts` generates `state` (~line 66) and sets the cookie (~line 80):

```ts
const state =
  typeof crypto !== "undefined" && crypto.randomUUID ? crypto.randomUUID() : ...;

const url = new URL(authEndpoint);
url.searchParams.set("response_type", "code");
url.searchParams.set("client_id", clientId);
url.searchParams.set("redirect_uri", redirectUri);
url.searchParams.set("scope", scope);
// <-- no url.searchParams.set("state", state)

const res = NextResponse.redirect(url.toString());
res.cookies.set("oidc_state", state, { httpOnly: true, ... });
```

`grep -n state src/app/api/auth/oidc/login/route.ts` returns only lines 66 and 80 — the value never reaches the query string.

Observed `Location` from `GET /api/auth/oidc/login`:

```
https://idp.example.com/application/o/authorize/?response_type=code&client_id=…&redirect_uri=…&scope=openid+profile+email
```

The IdP therefore returns `state=` (empty), and `src/app/api/auth/oidc/callback/route.ts` fails closed:

```ts
if (!code || !returnedState) {
  return NextResponse.redirect(new URL("/login?oidc_error=missing_code", reqUrlEarly));
}
```

Real callback hit from the browser, IdP-authenticated, code present:

```
GET /api/auth/oidc/callback?code=f212b802cacc420391502fec3bc28f1a&state= → 307
```

Reproduced against Authentik 2025.10.4, but this is IdP-independent — nothing can echo a `state` that was never sent.

**Fix:** add `url.searchParams.set("state", state);` alongside the other params.

### Bug 2 — error redirects point at the bind address

[...]

PR #10614, opened 2026-08-17
fix(auth): add missing state parameter to OIDC authorization URL

MeRezaRezaei:
## Problem

The OIDC login route generates a `state` UUID and stores it in the `oidc_state` cookie, but never includes it in the authorization URL sent to the OIDC provider.

This means the provider has no state to echo back, so the callback receives `state=null` and fails with `oidc_error=missing_code` — even though the code parameter is present.

## Root Cause

In `src/app/api/auth/oidc/login/route.ts`, lines 66-69 generate the state and lines 80-87 store it in the cookie, but lines 71-75 (the authorization URL) only set `response_type`, `client_id`, `redirect_uri`, and `scope` — the state is never sent.

## Fix

One line added after setting scope:

```typescript
url.searchParams.set("state", state);
```

This ensures the state is sent to the OIDC provider and returned in the callback for proper CSRF protection.

## Impact

Without this fix, OIDC login via Keycloak (or any OIDC provider) fails with `oidc_error=missing_code` because the provider never receives the state parameter.

With this fix, the full flow works: login → Keycloak → callback with state + code → token exchange → dashboard redirect.

## Testing

Tested end-to-end with Keycloak 26.0.7 — login completes successfully with the state parameter present in the callback URL.

### DIFF
--- a/src/app/api/auth/oidc/callback/route.ts
+++ b/src/app/api/auth/oidc/callback/route.ts
@@ -48,14 +48,14 @@ export async function GET(request: Request) {
   const originEarly = `${schemeEarly}://${hostEarly}`;
 
   if (!code || !returnedState) {
-    return NextResponse.redirect(new URL("/login?oidc_error=missing_code", reqUrlEarly));
+    return NextResponse.redirect(new URL("/login?oidc_error=missing_code", originEarly));
   }
 
   // Validate state from cookie (via seam so tests can capture)
   const cookieStore = await oidcCallbackInternals.getCookieStore();
   const storedState = cookieStore.get("oidc_state")?.value;
   if (!storedState || storedState !== returnedState) {
-    return NextResponse.redirect(new URL("/login?oidc_error=invalid_state", reqUrlEarly));
+    return NextResponse.redirect(new URL("/login?oidc_error=invalid_state", originEarly));
   }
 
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Thanks for this — verified locally with a red/green test: without `url.searchParams.set("state", state)` the login redirect never sends `state`, so the callback correctly rejects with `oidc_error=missing_code` (the callback's state validation against the `oidc_state` cookie at src/app/api/auth/oidc/callback/route.ts:50-59 is already wired correctly, your one-line fix closes the loop). Two things before merge: (1) per this repo's testing policy, a production fix needs an automated regression test — could you add one to `tests/unit/` following the pattern in `tests/unit/oidc-callback.test.ts` (seed OIDC settings via `localDb.updateSettings`, call the route's `GET()` directly, assert the redirect URL's `state` matches the `oidc_state` cookie)? (2) Heads up: issue #10224 covers this same root cause but also reports a second, still-open bug — the callback's error redirects use `reqUrlEarly` instead of the already-computed `originEarly`, so they resolve to the bind address (e.g. `https://0.0.0.0:PORT/login?oidc_error=...`) behind reverse proxies/containers. This PR doesn't touch that path, so we'll keep #10224 open (`Refs`, not `Closes`) after merging this.

PR #10620, opened 2026-08-17
fix(oauth): treat Kiro social poll status as alias of error for pending states

### LINKED ISSUE #10618: [BUG] Kiro/Amazon Q social login always fails with invalid_token_response (pending state unreachable)
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Ubuntu 26.04 LTS (kernel 7.0.0-29-generic)

### Node.js Version

v25.9.0

### Provider(s) Involved

Amazon Q, Kiro AI (both Google and GitHub social login)

### Model(s) Involved

n/a — fails during OAuth, before any model call

### Client Tool

Dashboard UI + cURL

### Description

Google/GitHub social login for **Kiro** and **Amazon Q** always fails immediately with `invalid_token_response`. The modal shows "Connection Failed" on the very first poll, before the user is ever shown the authorization URL, so the flow can never be completed.

**Root cause:** Kiro's device poll endpoint reports progress in a **`status`** field, but `classifyKiroSocialPoll()` only inspects **`error`**.

Upstream `POST https://prod.us-east-1.auth.desktop.kiro.dev/oauth/device/poll` returns, while the user has not yet authorized:

```
HTTP/2 200
{"accessToken":null,"identityProvider":null,"profileArn":null,"refreshToken":null,"status":"authorization_pending"}
```

In `src/lib/oauth/kiroSocialPoll.ts`:

```ts
export function classifyKiroSocialPoll(responseOk, responseStatus, data) {
  if (data.error === "authorization_pending" || data.error === "slow_down") {   // data.error is undefined
    return { kind: "pending", error: data.error };
  }
  if (!responseOk || data.error) {                                              // responseOk is true
    ...
  }
  if (!data.accessToken && !data.refreshToken) {
    return { kind: "error", error: "invalid_token_response", status: 502 };     // <-- reached on every poll
  }
  return { kind: "success" };
}
```

Because `data.error` is `undefined`, the response is `200 OK`, and both tokens are `null`, control reaches the last branch and returns a **terminal** error. `/api/oauth/kiro/social-exchange` then responds `pending: false`, and `KiroSocialOAuthModal.tsx` calls `fail()` and stops polling.

The pending state is therefore unreachable for this provider: `authorization_pending` can only ever arrive via `status`, never via `error`.

Note also that `KIRO_CONFIG.socialDeviceAuthorizeUrl`, `socialDevicePollUrl` and `socialClientId` are defined in `src/lib/oauth/constants/oauth.ts` but have no references anywhere in the shipped `src/` tree — the poll URL is only reachable from the compiled route, which made this hard to trace from source alone.

### Steps to Reproduce

Reproducible entirely with cURL, no browser needed:

```bash
# 1. Start a social login and capture the device code
curl -s "http://127.0.0.1:20128/api/oauth/kiro/social-authorize?provider=google" -o auth.json
DC=$(python3 -c "import json;print(json.load(open('auth.json'))['deviceCode'])")

# 2. Poll once, WITHOUT authorizing (exactly what the modal does)
curl -s -w '\nhttp=%{http_code}\n' -X POST \
  "http://127.0.0.1:20128/api/oauth/kiro/social-exchange" \
  -d "{\"deviceCode\":\"$DC\",\"provider\":\"google\",\"targetProvider\":\"amazon-q\"}"
```

Via the UI: Providers → Amazon Q (or Kiro AI) → Add Connection → **Google Account**.

### Expected Behavior

The poll reports the flow is still pending, so the modal displays the `authUrl` / user code and keeps polling until the user completes login:

```
http=200
{"success":false,"pending":true,"error":"authorization_pending"}
```

### Actual Behavior

The first poll returns a terminal error and the modal shows "Connection Failed — invalid_token_response":

```
http=502
{"success":false,"pending":false,"error":"invalid_token_response"}
```

### Suggested Fix

Treat `status` as an alias of `error` for the progress check in `src/lib/oauth/kiroSocialPoll.ts`:

```diff
 export type KiroSocialPollData = {
   error?: unknown;
+  /** Kiro reports poll progress here (e.g. "authorization_pending"), not in `error`. */
+  status?: unknown;
   accessToken?: unknown;
   refreshToken?: unknown;
 };

 export function classifyKiroSocialPoll(responseOk, responseStatus, data) {
-  if (data.error === "authorization_pending" || data.error === "slow_down") {
-    return { kind: "pending", error: data.error };
+  const progress = data.error ?? data.status;
+  if (progress === "authorization_pending" || progress === "slow_down") {
+    return { kind: "pending", error: progress };
   }
```

I applied exactly this locally (patching both `src/` and the corresponding compiled chunk) and verified:

- **Fixed:** polling before authorization now returns `200 {"success":false,"pending":true,"error":"authorization_pending"}`, repeatedly, instead of a 502.
- **No regression on genuine failures:** an invalid device code makes upstream return `status:"invalid_token"`, which is not one of the two pending values, so it still terminates with `pending:false` rather than looping forever.
- **No regression on validation:** a malformed body still returns `400` from the zod schema.

A follow-up worth considering separately: when `status` carries a terminal value such as `invalid_token`, the endpoint still reports the generic `invalid_token_response`. Surfacing `status` as the error code would make these failures much easier to diagnose.

### Scope

Affects every social login path that routes through `classifyKiroSocialPoll`: **Kiro AI and Amazon Q, via both Google and GitHub**. `targetProvider` accepts `"kiro" | "amazon-q"`, and both hit the same classifier, so neither provider can be connected through social login on 3.8.49.

krishna3554:
## Summary

Fixes #10618 — Kiro/Amazon Q social login always fails with invalid_token_response (pending state unreachable).

## Problem

Kiro's device poll endpoint reports progress in a `status` field (e.g. `authorization_pending`), but `classifyKiroSocialPoll()` only inspected `data.error`. This caused every pre-authorization poll to fall through to the terminal `invalid_token_response` error, making social login impossible for Kiro AI and Amazon Q via Google/GitHub.

**Upstream response (while user has not yet authorized):**
```json
HTTP/2 200
{"accessToken":null,"identityProvider":null,"profileArn":null,"refreshToken":null,"status":"authorization_pending"}
```

Because `data.error` is `undefined`, the response is `200 OK`, and both tokens are `null`, control reached the terminal error branch.

## Solution

- Add `status` field to `KiroSocialPollData` type
- Update `classifyKiroSocialPoll()` to check `data.status` as fallback for `data.error` when detecting pending states (`authorization_pending` | `slow_down`)
- `error` takes precedence over `status` if both present

## Changes

- `src/lib/oauth/kiroSocialPoll.ts`: Add `status` field and update `classifyKiroSocialPoll()`
- `tests/unit/kiro-social-poll.test.ts`: Add tests for `status`-based pending detection and precedence

## Testing

- All existing unit tests pass (5/5)
- New test added: `classifyKiroSocialPoll treats status as alias of error for pending states`
- TypeScript typecheck passes
- ESLint passes

### DIFF
--- a/src/lib/oauth/kiroSocialPoll.ts
+++ b/src/lib/oauth/kiroSocialPoll.ts
@@ -1,5 +1,7 @@
 export type KiroSocialPollData = {
   error?: unknown;
+  /** Kiro reports poll progress here (e.g. "authorization_pending"), not in `error`. */
+  status?: unknown;
   accessToken?: unknown;
   refreshToken?: unknown;
 };
@@ -26,8 +28,9 @@ export function classifyKiroSocialPoll(
   responseStatus: number,
   data: KiroSocialPollData
 ): KiroSocialPollOutcome {
-  if (data.error === "authorization_pending" || data.error === "slow_down") {
-    return { kind: "pending", error: data.error };
+  const progress = data.error ?? data.status;
+  if (progress === "authorization_pending" || progress === "slow_down") {
+    return { kind: "pending", error: progress as "authorization_pending" | "slow_down" };
   }
 
   if (!responseOk || data.error) {
--- a/tests/unit/kiro-social-poll.test.ts
+++ b/tests/unit/kiro-social-poll.test.ts
@@ -21,6 +21,26 @@ test("classifyKiroSocialPoll keeps only documented pending states retryable", ()
   });
 });
 
+test("classifyKiroSocialPoll treats status as alias of error for pending states", () => {
+  // Kiro upstream returns progress in `status`, not `error`
+  assert.deepEqual(classifyKiroSocialPoll(true, 200, { status: "authorization_pending" }), {
+    kind: "pending",
+    error: "authorization_pending",
+  });
+  assert.deepEqual(classifyKiroSocialPoll(true, 200, { status: "slow_down" }), {
+    kind: "pending",
+    error: "slow_down",
+  });
+  // error takes precedence over status if both present
+  assert.deepEqual(
+    classifyKiroSocialPoll(true, 200, { error: "slow_down", status: "authorization_pending" }),
+    {
+      kind: "pending",
+      error: "slow_down",
+    }
+  );
+});
+
 test("classifyKiroSocialPoll stops on denied, expired and malformed responses", () => {
   assert.deepEqual(classifyKiroSocialPoll(false, 403, { error: "access_denied" }), {
     kind: "error",


diegosouzapw:
Nice fix and good test coverage. I ran the full suite locally and also verified the regression: the new test fails against the pre-fix version of `classifyKiroSocialPoll` (returns `invalid_token_response`/502 instead of `pending`), which matches the symptom in #10618 exactly, and passes with your change. All 5 tests in `tests/unit/kiro-social-poll.test.ts` pass, ESLint is clean on both touched files, and the fix is wired into the real endpoint (`src/app/api/oauth/kiro/social-exchange/route.ts:67`). One note: GitHub Actions on this PR are showing `action_required` — looks like the workflow run needs maintainer approval before CI populates a signal — otherwise this looks merge-ready as-is.

krishna3554:
Thanks for the thorough verification! Glad the regression test reproduces the exact behavior from #10618 and passes with the fix. I’ll leave the code as-is and wait for the GitHub Actions approval/CI signal.

PR #10632, opened 2026-08-18
fix(sse): route bare qwen3.8-max to the canonical -preview id

diegosouzapw:
## Problem

Sending `qwen3.8-max` (without `-preview`) was rejected by the chatCore preflight:

```
API Error: 400 Input exceeds context window for qwen-cloud-token-plan/qwen3.8-max:
estimated 129106 input tokens, limit 128000.
```

The `128000` is not a stale catalog value — it is the `default: 128000` fallback in
`open-sse/services/contextManager.ts:26`, reached because the bare id resolves to
nothing. Every provider that serves this model registers it as `qwen3.8-max-preview`
(bailian-coding-plan, qoder, qwen-cloud-token-plan, qwen-web), declared with a real
**1M** window in `src/shared/constants/modelSpecs.ts`. Nothing declared the short form.

The same missing id also means the request would have been dispatched to the upstream
verbatim, under an id the upstream does not know.

## Fix

One line in `BUILT_IN_ALIASES` (`open-sse/services/modelDeprecation.ts`), the map
`resolveLifecycle()` applies at `open-sse/handlers/chatCore.ts:755` — before the context
preflight *and* before dispatch, so both symptoms clear at once.

A `MODEL_SPECS.aliases` entry would have fixed only the window: spec aliases resolve
capabilities, never the dispatched model id.

The comment on the entry records the removal condition — if Alibaba ships a distinct GA
`qwen3.8-max`, it is no longer the same model and the alias must go.

## Validation (TDD)

`tests/unit/qwen38-max-bare-id-alias.test.ts` — written first, confirmed failing on the
base (`actual: 'qwen3.8-max'`), green after the fix. It covers the alias, the canonical
id staying a no-op (no double rewrite), the 1M target window, the absence of a competing
`MODEL_SPECS` entry for the bare id, and `resolveLifecycle()` across all four providers.

Sibling sweep: 10 alias/deprecation/model-spec test files, **79 tests green**.
`npx eslint` clean on both files; `check-mutation-test-coverage` reports no drift.

⚠️ base-red inherited: #9985 — `typecheck:core` fails on
`open-sse/services/compression/engines/omniglyphAdapter.ts(126,54)`, a file this PR does
not touch (diff is 1 source file + 1 new test).

### DIFF
--- a/open-sse/services/modelDeprecation.ts
+++ b/open-sse/services/modelDeprecation.ts
@@ -40,6 +40,13 @@ const BUILT_IN_ALIASES: Record<string, string> = {
   "fireworks/accounts/fireworks/models/kimi-k2": "moonshotai/Kimi-K2",
   "kimi-k2": "moonshotai/Kimi-K2",
 
+  // Qwen — the model ships only under the `-preview` id (bailian-coding-plan, qoder,
+  // qwen-cloud-token-plan, qwen-web). Without this, the bare id missed MODEL_SPECS and
+  // the context preflight fell back to contextManager's `default: 128000`, rejecting
+  // prompts the model's real 1M window accepts. Drop this line if Alibaba ever ships a
+  // distinct GA `qwen3.8-max` — it would no longer be the same model.
+  "qwen3.8-max": "qwen3.8-max-preview",
+
   // Mistral short aliases
   "mistral-large": "mistral-large-latest",
   "mistral-small": "mistral-small-latest",
--- a/tests/unit/qwen38-max-bare-id-alias.test.ts
+++ b/tests/unit/qwen38-max-bare-id-alias.test.ts
@@ -1,0 +1,54 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+import { MODEL_SPECS } from "../../src/shared/constants/modelSpecs.ts";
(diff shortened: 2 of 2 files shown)


diegosouzapw:
All three red checks are the same inherited OmniGlyph cluster (#9985), not this PR:

- `Fast Quality Gates` → `omniglyphAdapter.ts(126,54) TS2339` (`baseline 0, live 1`)
- `Unit Tests fast-path (1/4)` → `compression/omniglyph-chatcore-plumbing.test.ts`
- `Unit Tests fast-path (2/4)` → `compression/pipeline-circuit-breaker.test.ts`

Proven, not assumed: both test files were run in a disposable worktree checked out at the **pure base tip** `aa912c42a7`, with no commit from this branch — they fail there identically. This PR's diff is one line in `BUILT_IN_ALIASES` plus a new test file, and touches nothing under `open-sse/services/compression/`.

Runtime validation on the .17 gateway: the custom alias was applied via `POST /api/settings/model-aliases` and a real completion with the bare id now succeeds (`qwen-cloud-token-plan/qwen3.8-max` → `content: "ok"`, 98 total tokens), where it previously returned `400 context_length_exceeded` against the phantom 128k limit.

PR #10635, opened 2026-08-18
fix(pricing): DeepSeek V4 static defaults stale by 4 days, off by ~1.6-2.4x

stanleytejakusuma:
## Description

DeepSeek switched `deepseek-v4-pro`/`deepseek-v4-flash` to peak/off-peak dynamic pricing on 2026-08-17 (off-peak = exactly half of peak; peak hours 01:00-04:00 and 06:00-10:00 UTC — [pricing page](https://api-docs.deepseek.com/quick_start/pricing/)). `DEFAULT_PRICING_FRONTIER.deepseek` in `frontier-labs.ts` is dated `checked 2026-08-13`, four days before the change, and was never updated — the compiled-in numbers are now stale by ~1.6-2.4x depending on token type.

I hit this from the consumer side (a local cost-reconciliation pipeline reading `/api/pricing/defaults`) and confirmed `POST /api/pricing` still returns non-2xx on this build — the table really is compiled-in, matching the existing pattern in `sync_gateway`-style write attempts. Fixing the static default is the only lever available.

## Scope

Uses the **off-peak** values (the documented lower bound) since `getPricingForModel(provider, model)` has no time-of-day dimension anywhere in its resolution chain. Adding real peak-awareness would mean threading a `ts`/hour parameter through the resolver and every call site — a real feature, not a bugfix — so deliberately out of scope here. Off-peak-only is a strict undercount confined to the two peak windows, never wrong-direction, and is still a ~1.6-2.4x correction over what's there now.

## Validation

- Confirmed live consumers of this table: `src/shared/utils/costEstimator.ts`, `src/lib/batches/costEstimator.ts`.
- Re-ran their test suites plus every `pricing-*`/`catalog-pricing-*` unit test (68 tests across 11 files, including `tests/unit/lib/batches/costEstimator.test.ts`) — all pass unchanged; none assert the specific stale values, so nothing needed updating on the test side.
- Live-verified the new numbers against the current DeepSeek pricing page directly (not just the changelog text) before writing them in.

### DIFF
--- a/src/shared/constants/pricing/frontier-labs.ts
+++ b/src/shared/constants/pricing/frontier-labs.ts
@@ -315,20 +315,28 @@ export const DEFAULT_PRICING_FRONTIER = {
       reasoning: 2.19,
       cache_creation: 0.55,
     },
-    // DeepSeek official API list prices, checked 2026-08-13.
+    // DeepSeek official API list prices, checked 2026-08-18. Superseded the
+    // prior 2026-08-13 flat prices below: DeepSeek switched v4-pro/v4-flash to
+    // peak/off-peak dynamic pricing on 2026-08-17 (peak = exactly 2x off-peak;
+    // peak hours 01:00-04:00 and 06:00-10:00 UTC — see
+    // https://api-docs.deepseek.com/quick_start/pricing/). This static table has
+    // no time-of-day dimension, so these are the OFF-PEAK (lower-bound) prices —
+    // a deliberate, documented undercount during the two peak windows, never an
+    // overcount. True peak-awareness would need a time dimension threaded through
+    // getPricingForModel() and every call site; out of scope for this fix.
     "deepseek-v4-pro": {
-      input: 0.435,
-      output: 0.87,
-      cached: 0.003625,
-      reasoning: 0.87,
-      cache_creation: 0.435,
+      input: 0.66,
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Reviewed — this is a clean, well-documented data fix. The rationale for using off-peak
(lower-bound) values given `getPricingForModel()` has no time-of-day dimension is sound and
explicitly scoped, and the source citation (DeepSeek's pricing page) is a good practice worth
keeping in future pricing PRs. One suggestion before merge: could you add a small regression
test asserting the new `deepseek-v4-pro`/`deepseek-v4-flash` values (or at least their
internal consistency, e.g. output = 3x input) so a future accidental revert gets caught by
CI? Not blocking given the low-risk, single-file, data-only nature of the change, but would
close the one gap I found. Nice catch on the LITELLM_PROVIDER_MAP alias bug in your other PR
(#10636) too — good pairing.

stanleytejakusuma:
Regression test landed on the branch (commit 3458380d0, co-authored with the review): tests/unit/pricing-deepseek-v4-static-regression.test.ts asserts the exact corrected off-peak values for deepseek-v4-pro/-flash AND the documented output=3x-input / reasoning=output consistency. Ran locally: 3/3 pass. Thanks for the review — ready for merge.

PR #10650, opened 2026-08-18
fix(sse): add the missing minimax-music dispatch to music generation

octo-patch:
Reason: registered MiniMax music models were impossible to call because the music handler had no branch for their provider format.

## Problem

`MUSIC_PROVIDERS.minimax` declares `format: "minimax-music"` and its models are published by `GET /v1/music/generations` and by the specialty model catalog, but `handleMusicGeneration` never grew a branch for that format. Every `minimax/*` music request fell through the whole dispatch chain to the final guard and came back as:

```
400 Unsupported music format: minimax-music
```

So the models were advertised while being unreachable, and no open or merged PR adds the missing handler.

## Change

- **`open-sse/handlers/mediaGeneration/minimaxMusic.ts` (new)** — the `minimax-music` handler. The upstream operation is a single synchronous POST, so the handler:
  - sends `model`, `prompt`, `lyrics`, `output_format`, `audio_setting`, `lyrics_optimizer`, `is_instrumental`, `audio_url`, `audio_base64` and `cover_feature_id`, forwarding only the fields the caller actually supplied (`instrumental` is accepted as the alias the route already uses);
  - pins `stream: false` — this route answers with a single JSON payload, and streaming responses are restricted to the hex output format anyway;
  - supports both output formats: `url` (a link that expires after 24h) is returned as `{ url, format }`, `hex` is decoded and returned as `{ b64_json, format }`, rejecting malformed hex instead of letting `Buffer` drop bad nibbles silently;
  - restricts `audio_setting.format` to the `mp3` / `wav` / `pcm` containers the endpoint accepts and reports the resolved container back to the caller;
  - checks the `base_resp` envelope (a non-zero `status_code` also arrives on HTTP 200) and reads the audio from `data.audio` only once `data.status` says the generation finished — an unfinished generation is reported back rather than polled, because this operation has no task id and no query endpoint;
  - guards the missing-credential case locally (401) instead of sending an unauthenticated request upstream, and records every outcome through `saveCallLog` like the other media handlers.
- **`open-sse/handlers/musicGeneration.ts`** — dispatch `minimax-music` to that handler, next to the other provider-format branches.
- **`open-sse/config/musicRegistry.ts`** — add the generation and cover model ids the entry was missing, declare the regional endpoint (`regionalBaseUrl`), and drop the query URL that does not exist for this operation. A connection reaches the regional endpoint through the existing per-connection `providerSpecificData.baseUrl` override; that endpoint is also the only one that accepts the `aigc_watermark` request field, so the field is forwarded there and dropped otherwise.
- **`tests/unit/minimax-music-generation.test.ts` (new)** — 9 regression tests: registry shape, dispatch + URL normalization + request-field mapping, hex → base64, cover inputs, regional endpoint + regional-only field, `base_resp` failure on HTTP 200, unfinished generation, completed-without-audio, upstream HTTP failure, and the missing-credential guard.
- **`changelog.d/fixes/minimax-music-generation-dispatch.md`** — changelog fragment.

No behavior changes for any other music provider format: the new branch is additive and sits before the unsupported-format guard.

## Checks

Run in this environment:

- `node --test tests/unit/minimax-music-generation.test.ts` — **9/9 pass**. Executed with Node 24's native TypeScript execution; the usage-log module (`@/lib/usageDb`) was stubbed because the workspace dependencies could not be installed here, so `npm run test:unit` and the ESLint / typecheck gates still need to run in CI.
- `node scripts/check/check-file-size.mjs` — OK (no baseline change needed: the new files are well under the cap and `musicGeneration.ts` is not frozen).


diegosouzapw:
Reviewed and validated: checked out the branch, ran `tests/unit/minimax-music-generation.test.ts` — 9/9 pass — and confirmed ESLint/typecheck show no new errors in the touched files (the two failures observed are pre-existing base-tip issues unrelated to this change: one already allowlisted in `eslint-suppressions.json`, the other in an unrelated compression-engine file this PR never touches). The `minimax-music` dispatch gap is real (models registered/advertised via `/v1/models` but every request 400'd with `Unsupported music format`), the fix is additive and isolated, and it follows existing codebase conventions (`providerSpecificData.baseUrl` override, `sanitizeErrorMessage`, `saveCallLog`). Looks good to merge.

Issue #9064, opened 2026-07-31
fix(providers): anthropic strips code-execution/skills beta flag (container rejected)


awci:
## Summary

Calling `/v1/messages` (native Anthropic format) through OmniRoute with the `code_execution` tool + a `container` object (used to auto-load a Skill into a fresh container) fails with a validation error saying `container` must be a string — even though the exact same request body succeeds when sent directly to `api.anthropic.com`. Sending `container` as a plain string "succeeds" but the skill is silently never loaded (a string container just means "reuse an existing container by id", it can't carry `skills`).

## Repro

```
POST https://<omniroute-host>/v1/messages?beta=true
{
  "model": "anthropic/claude-sonnet-5",
  "max_tokens": 4096,
  "container": { "skills": [{ "type": "custom", "skill_id": "skill_0169Fee3QMGdCNTcQ9JVqKyF", "version": "latest" }] },
  "messages": [{ "role": "user", "content": "weather in Izmir." }],
  "tools": [{ "type": "code_execution_20250825", "name": "code_execution" }]
}
```

- `container` as an object → 400, error says the `container` field must be a string.
- `container` as a string (container id) → 200, but the skill is not applied (expected — a bare container id can't carry `skills`).
- The identical object-shaped request sent straight to the real Anthropic API (not through OmniRoute) works.

## Root cause

The `anthropic` provider registry entry hardcodes a static `Anthropic-Beta` header that does not include any code-execution / skills beta flag:

- `open-sse/config/providers/registry/anthropic/index.ts:14-17` sets `"Anthropic-Beta": ANTHROPIC_BETA_API_KEY` unconditionally.
- `ANTHROPIC_BETA_BASE` in `open-sse/config/anthropicHeaders.ts:12-27` lists ~14 beta flags (tool-search, context-management, structured-outputs, etc.) but nothing for `code_execution_*` tools or the Skills-in-container feature.

Even when the client *does* send its own `anthropic-beta` header requesting the right flag, it gets filtered out:

- `open-sse/executors/default.ts:493-506` merges the client's `anthropic-beta` header via `mergeClientAnthropicBeta()`, but that function only forwards betas present in the `FORWARDABLE_CLIENT_BETAS` allowlist (`open-sse/config/anthropicHeaders.ts:53-56`): today just `tool-search-tool-2025-10-19` and `context-1m-2025-08-07`. Anything else the client asks for — including a code-execution/skills beta — is silently dropped.
[...]

PR #10651, opened 2026-08-18
fix(providers): only send the skills beta to Anthropic next to a code_execution tool

dpozimski:
## Summary

- The `anthropic` (API-key) provider no longer sends `skills-2025-10-02` in its static `Anthropic-Beta` header. Anthropic rejects **every** request that carries that beta without a `code_execution` tool — `400 "Skills beta requires the code_execution tool to be included in the request"` — on Haiku 4.5, Sonnet 5 and Opus 5 alike (probed directly against `api.anthropic.com` with a one-message body; `code-execution-2025-08-25` alone is accepted). Since #9064 added both flags to `ANTHROPIC_BETA_BASE`, on `release/v3.8.50` every plain API-key request that declares no code_execution tool (Claude Code, Claude Agent SDK, raw curl) fails with that 400 and the combo reports "All models failed".
- The beta now reaches upstream in exactly the two cases where it is legal: the client negotiated it (`FORWARDABLE_CLIENT_BETAS`, unchanged), or the outgoing body declares a `code_execution` tool (`hasCodeExecutionTool` + a per-request `appendAnthropicBetaHeader` on the anthropic executor) — which is the raw-curl `container.skills` case #9064 was fixing. `code-execution-2025-08-25` stays in the static set.

Observed on a production gateway (26 Claude Code / Agent SDK agents on API keys) after moving from 3.8.49 to the `next` image: 18 lost turns in ~2 minutes across four agents, all `[400]: Skills beta requires the code_execution tool to be included in the request.` — none in the preceding 24 h on 3.8.49. Rolled back; this PR is what makes 3.8.50 deployable for API-key Claude Code traffic.

## Related Issues

- Related to #9064 (follow-up: the static-header half of that fix breaks every request without a code_execution tool)

## Validation

- [x] Change type: provider
- [x] Focused tests: `tests/unit/probe-9064-code-execution-beta.test.ts` (rewritten), `tests/unit/provider-translate-path-golden.test.ts` (golden regenerated — the only diff is `skills-2025-10-02` leaving the anthropic `Anthropic-Beta` header), plus the beta/header neighbours: `claude-tool-search-beta-3974`, `claude-context-1m-client-beta-forward`, `claude-atu-effort-leak-9505`, `anthropic-header-dedupe-1475`, `agentrouter-cc-wire-image`, `executor-default-base`, `alternate-formats`, `glm-executor`, `claude-identity-version-sync`, `claude-codex-identity-version-sync` — 131 tests, all green
- [x] `npx eslint` + `npx prettier --check` on the changed files
- [x] Based on the current `release/v3.8.50` head
- [x] Production-code changes include updated automated tests in this PR

Direct probes against Anthropic that motivated the shape of the fix (one `user: hi` message, `max_tokens: 8`):

| `anthropic-beta` | tools | haiku-4-5 | sonnet-5 | opus-5 |
|---|---|---|---|---|
| `skills-2025-10-02` | none | 400 | 400 | 400 |
| `code-execution-2025-08-25` | none | 200 | 200 | 200 |
| `code-execution-2025-08-25,skills-2025-10-02` | none | 400 | 400 | 400 |
| `code-execution-2025-08-25,skills-2025-10-02` | `code_execution_20250825` | 200 | — | — |

## Tests Added Or Updated

- `tests/unit/probe-9064-code-execution-beta.test.ts` — asserts the skills beta is in no static header, is still forwarded when client-negotiated, is not added for a client that did not ask, `hasCodeExecutionTool` shape detection, and the append dedupes against the registry header
- `tests/snapshots/provider/translate-path.json` — regenerated (`UPDATE_GOLDEN=1`), 12 lines: the anthropic `Anthropic-Beta` value

## Coverage Notes

- `open-sse/config/anthropicHeaders.ts`: `hasCodeExecutionTool` and `ANTHROPIC_SKILLS_BETA` are covered by the rewritten probe test.
- `open-sse/executors/base.ts`: the per-request append is a two-line guarded call on the existing `appendAnthropicBetaHeader`; the header-merge behaviour it relies on is unit-tested in the probe test.

## Reviewer Notes

- The client-negotiated path is untouched: a real Claude Code client that sends `skills-2025-10-02` does so only when it also sends the code_execution tool, so forwarding stays correct.
- `ANTHROPIC_BETA_CLAUDE_OAUTH` loses the flag too (same base list); the OAuth path already replaces the header with `selectBetaFlags()`, which never emitted skills, so nothing changes there.


diegosouzapw:
Verified locally in an isolated worktree: the core fix is correct and well-tested — `hasCodeExecutionTool` + the per-request `appendAnthropicBetaHeader` gate cleanly on `provider === "anthropic"`, which is mutually exclusive with the Claude-OAuth/CC-protocol header-replacement path, so there's no double-injection. All 8 new tests in `probe-9064-code-execution-beta.test.ts` pass, plus the 4 named neighbour suites (20 tests).

One issue: the regenerated `tests/snapshots/provider/translate-path.json` replaces 6 occurrences of `"X-Stainless-Runtime-Version": "v26.3.0"` with `"<NODE>"`. That constant (`CLAUDE_CLI_STAINLESS_RUNTIME_VERSION`) is a hardcoded literal, not derived from `process.version` — the golden test's normalizer only collapses it to `<NODE>` when the machine running `UPDATE_GOLDEN=1` happens to be on Node v26.3.0 exactly. CI pins Node 24 (`CI_NODE_VERSION: "24"`), so `test-unit` will fail there. I reproduced this locally (Node v24.16.0): `node --import tsx/esm --test tests/unit/provider-translate-path-golden.test.ts` fails on exactly those 6 lines, and reverting them to the literal `"v26.3.0"` makes it pass again — the rest of the regenerated golden (the `Anthropic-Beta` value losing `skills-2025-10-02`) is correct and should stay as-is.

Could you revert just those 6 `"<NODE>"` occurrences back to `"v26.3.0"` in `tests/snapshots/provider/translate-path.json`? Everything else here looks ready.

PR #10652, opened 2026-08-18
feat(db): report the SQLite driver and its durability on the DB health check

maxmad64bis:
## Summary

- `DbHealthCheckResult` now reports which SQLite driver served the checked database, and whether
  that database's writes are durable: `"driver": { "name": "sql.js", "degraded": true }`.
- That was not answerable at runtime. `driverFactory` logs the chosen driver **once** at startup;
  `omniroute doctor` checks whether the better-sqlite3 native binary loads **on disk, from its own
  process**; and the payload of the endpoint whose job is DB health carried no driver field.
  Per #10552, a standalone bundle can rewrite the runtime `require` so the native driver fails to
  resolve **inside** the server while a fresh CLI resolves it fine — the server then runs on the
  sql.js WASM fallback while `doctor` reports green.
- `degraded` has exactly one meaning: writes are not durably backed by the database file. Two
  cases qualify — the sql.js WASM fallback, and an in-memory database, which `getDbInstance()`
  opens in cloud/build mode through the **native** cascade, so the driver name alone would read
  as healthy while nothing is persisted.
- Informative only: `isHealthy` stays defined by `issues`, so no existing check changes verdict.
- The field is read from the adapter the check ran against, so it cannot report a different
  database than the one inspected.

Also corrects the `DATABASE_GUIDE` health-check section, which documented a
`{status, checks: {...}}` payload this endpoint has never returned (`orphaned_artifacts` and
`table_sizes` have no source match in the repository). Its unterminated code fence left every
fabricated-docs claim below it unscanned; closing it surfaced `SQLITE_FULL` — a SQLite **result
code** quoted in prose, read by the gate as an env var — now allowlisted next to the other
prose-documented error codes.

## Related Issues

- Related to #10552 (fixes the cause this makes observable)
- Related to #7592, #8707, #7132 (silent sql.js fallback and its consequences)

## Validation

- [x] Change type: DB / observability (+ docs)
- [x] `npm run lint`

```
$ node --import tsx/esm --test tests/unit/db-health-driver.test.ts
# tests 5   # pass 5   # fail 0

$ node --import tsx/esm --test tests/unit/db-health-check.test.ts     # pass 10  # fail 0
$ node --import tsx/esm --test tests/unit/db-core.test.ts             # pass 40  # fail 0
$ node --import tsx/esm --test tests/unit/db-core-extended.test.ts    # pass 27  # fail 0

$ npx vitest run --config vitest.mcp.config.ts \
    open-sse/mcp-server/__tests__/dbHealthTool.test.ts
Test Files  1 passed (1)      Tests  3 passed (3)

$ npx eslint src/lib/db/healthCheck.ts tests/unit/db-health-driver.test.ts
0 errors

$ for g in check:docs-sync check:fabricated-docs check:docs-symbols check:doc-links; do
    npm run "$g" --silent >/dev/null 2>&1; echo "$g => rc=$?"; done
check:docs-sync => rc=0
check:fabricated-docs => rc=0
check:docs-symbols => rc=0
check:doc-links => rc=0
```

touch.

⚠️ base-red inherited: #9985

## Tests Added Or Updated

- `tests/unit/db-health-driver.test.ts` (new, 5 tests): `describeDbDriver` flags sql.js as
  degraded, flags every driver as degraded for `:memory:`, and leaves the three native drivers
  undegraded on a file-backed database; `runDbHealthCheck` against a real database reports a
  driver **equal to** the adapter's own `db.driver`; and an in-memory database opened through the
  same `tryOpenSync` cascade the cloud/build path uses really reports `name === ":memory:"`.

## Coverage Notes

- Touched `src/`: `describeDbDriver()` (pure, all branches covered) and the single construction
  site of `DbHealthCheckResult`, covered by the real-database test above.
- No coverage moved down; the production change is additive.

## Reviewer Notes

- No migration, no feature flag, no change to request handling.
- `DbHealthCheckResult` gains one required field. It has a single producer, and both consumers
  (`/api/db/health`, `omniroute_db_health_check`) serialize the whole result, so no further
  wiring is needed.
- `driver` is nested rather than two flat fields so the two facts stay grouped and a later
  `persistence` refinement needs no third root field — happy to flatten it for consistency with
  the surrounding style.
- Deliberately not folded into `isHealthy`: a database on sql.js still works, and turning working
  installations red would change an existing check's verdict inside an observability change.
- `getDriverInfo()` / `setDriverInfo()` in `src/lib/db/core.ts` answer a different question — how
  better-sqlite3 was _resolved_, not which driver is serving — and have no production caller.
  Left untouched; removing or reviving them belongs in its own PR.
- The `check-fabricated-docs` allowlist entry is a pre-existing false positive that the broken
  fence had been hiding, not a new violation: the unmodified file passes the gate, and the same
  file with only the missing fence closed fails on `SQLITE_FULL`.
- Two unrelated formatting hunks in `healthCheck.ts` come from the repository's own
  `lint-staged` Prettier pass — the committed file predates the current formatting on those
  lines.

### DIFF
--- a/docs/ops/DATABASE_GUIDE.md
+++ b/docs/ops/DATABASE_GUIDE.md
@@ -453,26 +453,47 @@ Run monthly during low-traffic windows. (WAL mode reduces the need, but doesn't 
 
 `src/lib/db/healthCheck.ts` provides **DB-level health diagnostics**:
 
-````bash
-GET /api/db/health
+Both verbs require authentication (`401` otherwise). `GET` diagnoses only; `POST` runs the
+same check with `autoRepair` enabled.
 
-Returns:
+```bash
+GET  /api/db/health   # diagnose
+POST /api/db/health   # diagnose + repair
+```
+
+The response is the `DbHealthCheckResult` produced by `runDbHealthCheck()`
+(`src/lib/db/healthCheck.ts`):
 
 ```json
 {
-  "status": "healthy",
-  "checks": {
-    "writable": { "status": "pass" },
-    "integrity": { "status": "pass", "result": "ok" },
-    "foreign_keys": { "status": "pass", "violations": 0 },
-    "orphaned_artifacts": { "status": "warn", "count": 12 },
-    "table_sizes": {
-      "usage_history": { "rows": 12345, "size_mb": 12.3 },
-      "call_logs": { "rows": 567, "size_mb": 2.1 }
+  "isHealthy": false,
+  "issues": [
+    {
+      "type": "broken_reference",
+      "table": "domain_budgets",
+      "description": "Domain budgets referenced API keys that no longer exist.",
+      "count": 2
     }
-  }
+  ],
+  "repairedCount": 0,
+  "backupCreated": false,
+  "autoRepair": false,
+  "checkedAt": "2026-08-18T09:00:00.000Z",
+  "driver": { "name": "better-sqlite3", "degraded": false }
 }
-````
+```
+
+| Field             | Meaning                                                                                                                                        |
+| ----------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| `isHealthy`       | `true` when `issues` is empty. `driver` never influences it.                                                                                   |
+| `issues[].type`   | One of `integrity_check_failed`, `broken_reference`, `stale_snapshot`, `invalid_state`.                                                        |
+| `repairedCount`   | Rows repaired during this run; always `0` when `autoRepair` is false.                                                                          |
+| `backupCreated`   | Whether a backup was taken before repairing.                                                                                                   |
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Reviewed #10652. Verified all 3 red checks on this run (Fast Quality Gates, Unit Tests fast-path 2/4 and 3/4) are pre-existing base-red on release/v3.8.50 — reproduced the same compression/omniglyph failures on a clean base checkout with none of this PR's changes applied. Ran the PR's own tests (db-health-driver.test.ts 5/5, db-health-check.test.ts 10/10) plus eslint on the touched files — all clean. Confirmed the driver type is derived from the adapter's closed union (no drift risk) and that both consumers (/api/db/health, omniroute_db_health_check MCP tool) already serialize the whole result object, so no additional wiring was needed. Looks merge-ready.

Issue #8411, opened 2026-07-24
feat(backend): rate-limit-resilient, configurable proxy health-check probe target


RCrushMe:
### Problem / Use Case

Proxy health sweeps (`ProxyHealth`) and the bulk auto-test (`POST /api/settings/proxies/auto-test`) probe `https://httpbin.org/ip` through each SOCKS5 proxy. When all proxies are tested concurrently (batch of 10, `CONCURRENCY=10`), `httpbin.org` rate-limits or throttles the simultaneous HEAD requests — producing **all 12 proxies reported as inconclusive** (timeout), even though:

1. Individual proxy tests against the same endpoint pass reliably.
2. Credential health checks through the same proxies to actual provider APIs (`integrate.api.nvidia.com`, etc.) succeed with 100% pass rate in the same timeframe.

**Observed sweep pattern (from live logs):**

```
12:22:05 — 12 tested, 4 alive, 8 inconclusive
12:32:08 — 12 tested, 0 alive, 12 inconclusive   ← bulk probe hit httpbin rate limit
```

Meanwhile, CredentialHealth at 12:32:41 tests 45 connections through the same proxies (batches of 5) — **all succeed**:

```
12:32:43 — openrouter/f8502183  → proxy=global:127.0.0.1  status=success
12:32:43 — brave-search/87224d0f → status=success
12:32:43 — openai-compatible/21a569ad → status=success
12:32:46 — nvidia/bd2ffa4c  → status=success
... (all success through 7897, 17990-18000)
```

The proxy infrastructure is healthy. The false negatives come from the probe target, not the proxies.

### Root Cause

Three independent code paths all hardcode `httpbin.org/ip` as the default probe target:

| Component | File | Timeout | Concurrency |
|-----------|------|---------|-------------|
| `ProxyHealth` sweep | `src/lib/proxyHealth/scheduler.ts:29` | 15s | 10 |
| `auto-test` bulk | `src/app/api/settings/proxies/auto-test/route.ts:14` | 5s | 10 |
| `proxyFallback` test | `open-sse/utils/proxyFallback.ts:395` | 3s | unlimited |

When 10 SOCKS5 proxies exit through the same NAT IP, `httpbin.org` sees 10 concurrent HEAD requests from one address and throttles them. All requests exceed their timeout window → all reported as "inconclusive".

The `PROXY_HEALTH_TEST_URL` env var already exists for the sweep, but:
- The auto-test and proxy-fallback paths do not respect it.
[...]

Issue #6246, opened 2026-07-05
[BUG] Massive issue with proxies since 3.8.44 - IP leak


tenshiak:
### OmniRoute Version

3.8.44

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Description

Inactive proxies are bypassed instead of blocking the connection!

Some health-check set all my proxies as Inactive. My accounts had proxies assigned per account, and OmniRoute just bypassed all proxies and connected directly with all of them instead of blocking the connection. This is unacceptable behavior.

Also I ask for a way to bulk enable (activate) proxies.

I had to downgrade for now. I don't know if that test that deactivates all accounts is periodic or not.

### Steps to Reproduce

1. Assign a proxy to the account/provider/combo. Turn on all safeties if you want.
2. Set the assigned proxy as inactive
3. Try to send a message

### Expected Behavior

1) I don't want to OmniRoute to EVER deactivate my own proxies. Maybe a checkbox when setting up/editing a proxy to disallow deactivating it to differentiate between free and paid proxies. Not all of us use free proxies. Either that or at least a flag to disable proxy health checks. The former is better for everyone I think.
2) When a provider/combo/account has a proxy assigned, and the proxy fails or is inactive, it should BLOCK the connection, not connect directly. That's a massive issue for me.
3) "Test All" button should not disable any proxies. It is a Test, not test & set status. Only the user should be able to set the proxies as active/inactive, just like with accounts, providers & combos. Accounts can be automatically disabled when banned, but only if it is allowed by the user in the settings. Why is it different for proxies?

### Actual Behavior

In the description

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Additional Context

My provider is webshare (paid proxies) and I connect through socks5://p.webshare.io:80, each proxy has its own different username. It automatically assigns me a new IP if it fails, on its own. I don't need/want OmniRoute to do anything for me in this regard. I think they were all inactivated right after I upgraded to 3.8.44. I see that scraping free proxies from webshare was introduced in 3.8.44, but I don't know if it is related.

When i click "Test All", all of them fail with "fetch failed" and get deactivated., but testing individual ones work just fine. And testing individual ones doesn't activate them back.

PR #10654, opened 2026-08-18
fix(api): a target refusing the egress IP is not a healthy proxy

maxmad64bis:
## Summary

- The proxy reachability probe classified every response under 500 as `ok`
  (`src/lib/proxyHealth/scheduler.ts`). A **401/403/429 from the probe target** — the shape a
  destination uses to refuse a banned or rate-limited egress IP — was therefore reported as a
  healthy proxy. The proxy did relay, so it is not failing; but it is not serving that
  destination either, and `ok` hid exactly that.
- Adds a `blocked` outcome for 401/403/429, produced by a shared pure `classifyProbeStatus()`.
  The sweep and the "Test All" endpoint (`/api/settings/proxies/auto-test`) were deriving the
  same `status < 500` rule independently; they now share one definition.
- The sweep log gains a `blocked by target` count next to `alive` / `inconclusive`.

**Behaviour is deliberately unchanged.**

- `blocked` is neutral in `decideProxyHealthAction`, exactly like `inconclusive`: no failure
  count, no status write. One target refusing an IP does not make the proxy dead, and the
  removal policy stays operator-owned (#6246 policy C/D).
- On `/api/settings/proxies/auto-test`, `alive` keeps its exact prior meaning (any status under
  500). The refusal is surfaced as a separate optional `blockedByTarget` field, so nothing
  changes for the opt-in status write behind `PROXY_HEALTH_AUTO_DEACTIVATE`.

No migration, no feature flag, no UI string, no i18n churn.

## Related Issues

- Related to #8411 (proxy probe semantics)
- Related to #6246 (probe outcomes are advisory; the operator owns proxy status)

## Validation

- [x] Change type: routing / proxy health (+ API field)
- [x] `npm run lint`

```
$ node --import tsx/esm --test tests/unit/proxy-health-blocked-outcome.test.ts
# tests 8   # pass 8   # fail 0

$ node --import tsx/esm --test tests/unit/proxy-health-6246.test.ts                  # pass 7  # fail 0
$ node --import tsx/esm --test tests/unit/proxy-health-decide-action-6246.test.ts    # pass 6  # fail 0
$ node --import tsx/esm --test tests/unit/proxy-health-auto-disable-decision.test.ts # pass 7  # fail 0
$ node --import tsx/esm --test tests/unit/proxy-batch-routes-5918.test.ts            # pass 5  # fail 0

$ npm run lint                       # rc=0 (cold cache)
$ npm run check:cycles               # OK — no cycles across 419 files
```

touch.

⚠️ base-red inherited: #9985

## Tests Added Or Updated

- `tests/unit/proxy-health-blocked-outcome.test.ts` (new, 8 tests):
  - `classifyProbeStatus` maps 401/403/429 → `blocked`, 200/204/301/400/404 → `ok`,
    500/502/503 → `inconclusive`;
  - **exhaustive over 100–599**: exactly `[401, 403, 429]` are flagged `blocked`, and
    `alive` (`ok || blocked`) equals the previous `status < 500` for **every** status — the
    regression guard for the "Test All" endpoint keeping its behaviour;
  - `decideProxyHealthAction` on `blocked` touches neither count nor status, **including** at the
    removal threshold with both `autoRemove` and `autoDisable` enabled — the destructive path is
    proven unreachable from `blocked`;
  - `blocked` does not reset a failure streak the way `ok` does.

## Coverage Notes

- Touched `src/`: `classifyProbeStatus()` (pure, every branch covered), the `blocked` branch of
  `decideProxyHealthAction` (covered, including the destructive-path case), and the two call
  sites that now consume the shared classifier.
- No coverage moved down; `scheduler.ts` loses an inline ternary to a tested function.

## Reviewer Notes

- `ProxyProbeOutcome` widens from three values to four. Both consumers live in this PR
  (`scheduler.ts`, `decision.ts`), and the union is closed, so TypeScript enforces that the new
  case is handled — that exhaustiveness is the point.
- 407 is **not** part of this change. undici never surfaces a `407` as a response — it throws,
  both on an HTTPS `CONNECT` tunnel and on a plain HTTP request — so a proxy demanding
  authentication already lands in the `catch` and is classified `fail`. Verified against a local
  proxy answering 407 to everything; no change was warranted.
- Known limitation, deliberately not guarded: if the probe target itself degrades and answers 403
  to everyone, every proxy is reported blocked for that sweep. Because `blocked` is neutral, the
  worst case is one misleading log line — no proxy is disabled or removed. A guard would need
  pool-wide aggregation in the sweep for a strictly cosmetic gain.
- `blockedByTarget` is optional and only present when true, so existing consumers of the
  "Test All" response are unaffected.
- Folding `blocked` into `alive` was considered and rejected: under
  `PROXY_HEALTH_AUTO_DEACTIVATE=true` it would start deactivating proxies that are currently left
  active, which is a policy change rather than an observability fix.

### DIFF
--- a/src/app/api/settings/proxies/auto-test/route.ts
+++ b/src/app/api/settings/proxies/auto-test/route.ts
@@ -4,6 +4,7 @@ import { createErrorResponseFromUnknown } from "@/lib/api/errorResponse";
 import { requireManagementAuth } from "@/lib/api/requireManagementAuth";
 import { createProxyDispatcher, proxyConfigToUrl } from "@omniroute/open-sse/utils/proxyDispatcher";
 import { fetch as undiciFetch } from "undici";
+import { classifyProbeStatus } from "@/lib/proxyHealth/decision";
 import { resolveHealthCheckStatusWrite } from "@/lib/proxyHealth/statusPolicy";
 import { isValidationFailure, validateBody } from "@/shared/validation/helpers";
 import { createErrorResponse } from "@/lib/api/errorResponse";
@@ -24,6 +25,13 @@ interface TestResult {
   host: string;
   port: number;
   alive: boolean;
+  /**
+   * The proxy relayed, but the target refused this egress IP (401/403/429).
+   * Reported alongside `alive` rather than inside it: a refused IP is still a
+   * reachable proxy, so folding it into `alive` would change what the opt-in
+   * status write (`PROXY_HEALTH_AUTO_DEACTIVATE`) deactivates.
+   */
+  blockedByTarget?: boolean;
   latencyMs: number | null;
   error?: string;
 }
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Verified locally against release/v3.8.50: `classifyProbeStatus` and the `blocked` branch of `decideProxyHealthAction` are covered by 8 new tests (all passing), and the adjacent proxy-health suites (25 tests) still pass with no regressions. The widened `ProxyProbeOutcome` union has exactly two consumers (both touched here) and both handle the new case exhaustively. The two ESLint findings on the touched files are pre-existing allowlisted `no-restricted-imports` violations on an untouched import line, not introduced by this diff. This looks merge-ready as-is — thanks for the thorough test coverage and the exhaustive-range regression guard proving `alive` semantics are unchanged for the "Test All" endpoint.

PR #10657, opened 2026-08-18
fix(api): share one probe-target resolution between both proxy health checks

maxmad64bis:
## Summary

- Both proxy health checks had their own copy of the probe target and batch size:
  `src/lib/proxyHealth/scheduler.ts` and `src/app/api/settings/proxies/auto-test/route.ts`.
  Change one and the other silently stays on the defaults; the batch size wasn't tunable at all.
- This PR introduces `src/lib/proxyHealth/probeTarget.ts`, the single place where the target is
  resolved. The batch size is now configurable (default 10, capped at 50), and probes within a
  batch are spread out deterministically (100 ms between each, starting from index 0). No
  randomness, so a sweep behaves the same on every run.
- With the default settings nothing changes: same target, batch size 10, and
  `staggerDelayMs(0, …) === 0`, so the first probe of a batch is never held back. The only
  difference: probes 2..N of a batch now leave 100 ms apart.

## Related Issues

- Closes part of #8411 (target unification; the blocked-outcome half of the issue is #10654)

## Validation

- [x] Change type: routing / proxy health (+ env docs)
- [x] `npm run lint`

```
$ node --import tsx/esm --test tests/unit/proxy-probe-target.test.ts
# tests 15   # pass 15   # fail 0

# 5 pre-existing proxy-health-* suites: 22/22 pass
# (incl. proxy-health-scheduler-listproxies-shape.test.ts, insensitive by construction:
#  it seeds a single proxy, so it always sits at index 0)

$ npm run lint            # rc=0
$ npm run check:cycles    # OK — no cycles
$ npm run check:docs-all  # rc=0
```

touch.

⚠️ base-red inherited: #9985

## Tests Added Or Updated

- `tests/unit/proxy-probe-target.test.ts` (new, 15 tests): checks the defaults are exactly what
  both call sites hardcoded before, the bounds, and the delay computation.
- Most important: the batch size can never drop below 1, whatever the environment says. If it
  were 0, the loop's `i += concurrency` would never advance and the sweep would walk backwards.
  Tested against hostile values: `"0"`, `"-1"`, `"abc"`, `"Infinity"`, `"0.4"`, …
- Targets that are only whitespace pass through untouched on purpose: they are truthy today and
  already break probes, so cleaning them up here would silently repair a broken deployment
  without anyone noticing.

## Coverage Notes

- Touched `src/`: `src/lib/proxyHealth/probeTarget.ts` is pure and fully covered by the 15
  tests; the two call sites that used to duplicate the logic now just call it.
- No coverage moved down.

## Reviewer Notes

- **Why the spacing is inside the batch.** The issue asks for jitter between batches. But
  batches never overlap — the `await Promise.allSettled` serializes them. The problem is the
  batch itself: `batch.map()` fires every fetch in the same tick, so the concurrency peak is
  `CONCURRENCY`, reached inside a batch. That is where the spacing has to go. `index × step`,
  not random, so tests can assert the exact value.
- The delay is applied by the caller before the probe runs, not inside it. Both paths set their
  own timeout inside the probe function, and `auto-test` starts its `latencyMs` clock there
  too. Waiting outside means every probe keeps its full timeout budget and the reported latency
  stays about the proxy.
- **Decisions I left to you** (each is a one-line change):
  1. Default target stays `httpbin.org/ip`. The mechanism to change it now exists; pointing all
     existing installs at a different destination is your call, not an external PR's.
  2. Default batch size stays 10 (configurable, capped at 50). You mentioned lowering
     concurrency — `5` would be the obvious value, but it doubles the sweep duration for
     everyone and I have no measurement to justify it.
  3. Timeouts stay as they are: 15 s for the sweep, 5 s for `auto-test`. A background sweep and
     a human waiting on a button aren't the same constraint.
  4. `sweep()` still has no reentrancy guard. `scheduleSweep()` uses `setInterval`, so a new
     sweep can start while the previous one runs, and both write to `failureMap` — failures
     could be double-counted and a proxy removed early. Pre-existing, and the spacing makes it
     ~6% more likely. The credential scheduler already has the fix (`sweepInProgress`,
     `src/lib/credentialHealth/scheduler.ts:218`, reset in a `finally` at `:285`). Out of scope
     here.
  5. Naming follows the existing `PROXY_HEALTH_*` family, not the `OMNIROUTE_PROXY_TEST_*` form
     suggested in the issue.
- **What is not covered.** `open-sse/utils/proxyFallback.ts` and the manual test screen probe
  `api.openai.com/v1/models` — a different question ("can this proxy reach an AI API" rather
  than "does it relay"). Left untouched per your correction on the issue. The batching pattern
  appears 6 times in the repo; the other 4 send each request to its own target, so they cannot
  converge on one endpoint and are out of this class.
- `docs/reference/ENVIRONMENT.md` is in `.prettierignore` (tables are aligned by hand). The two
  new rows were inserted at the existing column widths, so the diff there is 3 lines instead of
  ~1460 after a reformat.


diegosouzapw:
Reviewed and validated locally: the new tests/unit/proxy-probe-target.test.ts (15 tests) plus the 5 pre-existing proxy-health-* suites (22 tests) all pass in an isolated worktree checkout, and the CI reds on this PR (Fast Quality Gates, Unit Tests fast-path 2/4, 3/4) trace to the acknowledged base-red #9985 (compression/omniglyph area) — none touch the files this PR changes. Clean, well-scoped fix for #8411's target-unification half. Merge-ready.

PR #10660, opened 2026-08-18
fix(compression): bound RTK raw-output store growth and make pointer reads O(bucket) (#10659)

### LINKED ISSUE #10659: fix(resilience): RTK raw-output store grows unbounded when rawOutputRetention != 'never'; per-pointer readdirSync over the whole store stalls the event loop
## Summary
When `rtkConfig.rawOutputRetention` is set to `"always"` (or `"failures"`), the RTK compression engine appends a `.log` + `.meta.json` pair per compressed tool result to `DATA_DIR/rtk/raw-output` **with no cleanup of any kind**: no TTL, no file-count cap, no directory-size cap, no scheduled purge. Under sustained multi-agent load this reached **4.67M files / 22 GB in ~5 days** on a production gateway (default 1 MB/file cap, avg ~5 KB/file).

## Impact
1. **Disk**: unbounded growth (docs already warn "only enable for debugging, not always-on", with a 50–500 MB/day estimate that is easily exceeded at high request rates).
2. **Event-loop stalls (critical)**: `readRtkRawOutput(pointerId)` does `fs.readdirSync(dir).find(f => f.includes(pointerId))` — a full synchronous read of every entry in the directory. With millions of entries, a single pointer read blocks the Node event loop for minutes: the gateway stops responding to health checks, dashboard, and all /v1 traffic (observed 30–90 s to 9-minute freezes, "down/up" flapping). Any other consumer that walks the store (`listRtkCommandSamples`) has the same problem.

## Suggested fixes (any subset)
1. **Bounded retention**: enforce a max-age (e.g., 7 days) or max-count/max-bytes cap on the store with an async (throttled, non-blocking) purge; or make "always" require an explicit retention window.
2. **O(1) pointer lookup**: `readRtkRawOutput` currently scans the whole directory per read. Pointer filenames embed the pointer id (`<ts>-<slug>-<id>.log`); lookup can be a direct path/glob instead of a full readdir, or store an id→path index.
3. **Visibility**: warn in the dashboard when `rawOutputRetention != "never"` (e.g., show current store size + projected growth), and/or surface a "purge raw-output store" action.

## Notes
- Deleting files is safe: `readRtkRawOutput` returns null on a missing file (callers 404/fall back), `listRtkCommandSamples` skips missing entries, and capture is best-effort by design.
- Default is `"never"` — this only bites when the operator (or a prior session) explicitly enables retention, which is a supported dashboard dropdown value.

Repro environment: v3.8.49 local gateway, macOS; `rawOutputRetention:"always"` via dashboard/PUT settings.

stanleytejakusuma:
## Summary
Fixes #10659 — the RTK raw-output retention store could grow unbounded (observed: **4.67M files / 22 GB in ~5 days** on a busy gateway) and every pointer read did a full synchronous `readdirSync` over the whole store, freezing the Node event loop for minutes (gateway "down/up" flapping, dashboard buffering, /v1 timeouts).

## Changes
1. **Bucketized layout (O(bucket) reads):** new captures land in `DATA_DIR/rtk/raw-output/<id[0:2]>/...` so `readRtkRawOutput()` is a tiny subdir read instead of a full-store scan. Legacy flat files stay readable via a **guarded fallback scan** — if the legacy flat store exceeds 100k entries it refuses to scan synchronously (the exact event-loop freeze #10659 reports) and logs a warning recommending a one-off cleanup.
2. **Bounded retention:** new `purgeRtkRawOutput()` — asynchronous (fs/promises, concurrency-limited), throttled to once per minute via the scheduler, best-effort (never throws into callers). Enforces:
   - max age: `rawOutputMaxAgeDays` (default **30**)
   - max file count: `rawOutputMaxFiles` (default **100_000**)
   Scheduled off the hot path whenever `rawOutputRetention != "never"`.
3. **Config plumbing:** both knobs added to `RtkConfig` / `DEFAULT_RTK_CONFIG` / `RTK_SCHEMA` (dashboard + PUT) with validation (positive integers). Default retention remains `"never"`.
4. **Tests:** new `rtk-raw-output-retention.test.ts` (bucket write/read, legacy flat read, age purge, count cap, retention=never). Full RTK suite: **157 pass / 0 fail**; `typecheck:core` passes. (The open-sse typecheck ratchet failures on `videoBridgeHelpers.ts`/`stream.ts` are pre-existing on this base — identical on the clean checkout.)

## Notes
- Deleting raw-output files is safe: `readRtkRawOutput` returns `null` on missing files, `listRtkCommandSamples` skips missing entries, and capture is best-effort by design.
- Existing operators with a large legacy flat store: delete it once (e.g. `rm -rf ~/.omniroute/rtk/raw-output`) and new growth stays bounded automatically.

## Test plan
- `node --test tests/unit/compression/rtk-raw-output-retention.test.ts tests/unit/compression/rtk-raw-output.test.ts` — 12 pass
- `node --test tests/unit/compression/rtk-*.test.ts` — 157 pass
- `npm run typecheck:core` — pass

### DIFF
--- a/open-sse/services/compression/engines/rtk/configSchema.ts
+++ b/open-sse/services/compression/engines/rtk/configSchema.ts
@@ -67,6 +67,22 @@ export const RTK_SCHEMA: EngineConfigField[] = [
     ],
   },
   {
+    key: "rawOutputMaxFiles",
+    type: "number",
+    label: "Max raw-output files (oldest purged beyond this)",
+    defaultValue: DEFAULT_RTK_CONFIG.rawOutputMaxFiles,
+    min: 1,
+    max: 10_000_000,
+  },
+  {
+    key: "rawOutputMaxAgeDays",
+    type: "number",
+    label: "Max raw-output age (days)",
+    defaultValue: DEFAULT_RTK_CONFIG.rawOutputMaxAgeDays,
+    min: 1,
+    max: 3650,
+  },
+  {
     key: "enableRenderers",
     type: "boolean",
     label: "Semantic renderers",
@@ -112,6 +128,11 @@ export function validateRtkEngineConfig(config: Record<string, unknown>): Engine
     config.rawOutputRetention !== "always"
   ) {
     errors.push("rawOutputRetention must be never, failures, or always");
+  }
+  for (const key of ["rawOutputMaxFiles", "rawOutputMaxAgeDays"]) {
+    if (config[key] !== undefined && (typeof config[key] !== "number" || config[key] < 1)) {
+      errors.push(`${key} must be a positive number`);
+    }
   }
   return { valid: errors.length === 0, errors };
 }
--- a/open-sse/services/compression/engines/rtk/index.ts
+++ b/open-sse/services/compression/engines/rtk/index.ts
@@ -9,7 +9,11 @@ import { matchRtkFilter } from "./filterLoader.ts";
 import { applyLineFilter } from "./lineFilter.ts";
 import { smartTruncate } from "./smartTruncate.ts";
 import { normalizeCodeLanguage, stripCode } from "./codeStripper.ts";
-import { maybePersistRtkRawOutput, type RtkRawOutputPointer } from "./rawOutput.ts";
+import {
+  maybePersistRtkRawOutput,
+  scheduleRtkRawOutputPurge,
+  type RtkRawOutputPointer,
+} from "./rawOutput.ts";
 import { applyRenderer } from "./renderers/index.ts";
 import { isTextBlock } from "../../messageContent.ts";
 import { adaptBodyForCompression } from "../../bodyAdapter.ts";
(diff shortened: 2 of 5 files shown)


diegosouzapw:
Review blocked by environment: this reviewer sandbox had no shell/GitHub/read tooling, so the diff could not be fetched or executed. From the base source (release/v3.8.50), the PR's premise is confirmed as a real gap: the RTK raw-output store is unbounded and readRtkRawOutput() is O(store). Before merge, please confirm: (1) a regression test was added/extended (Hard Rule #18); (2) the growth-bound cannot evict a live capture; (3) any bucketed layout is fully mirrored into readRtkRawOutput/listRtkCommandSamples and the discover/learn API routes, or existing route tests will need to pass; (4) no eager full-store scans remain. No negative finding is being asserted — re-verification is required because the changes could not be inspected here.

stanleytejakusuma:
Confirmations for the review asks:
1. Regression test (Hard Rule #18): yes — tests/unit/compression/rtk-raw-output-retention.test.ts (6 tests: bucketized write/read, legacy flat read, age purge, count cap keeping the newest, retention=never, unknown-pointer null). Full RTK suite 157/157 green; typecheck:core green.
2. Growth-bound cannot evict a live capture: purge only removes (a) files older than rawOutputMaxAgeDays (default 30d) or (b) the OLDEST files beyond rawOutputMaxFiles — a just-written capture is by construction the newest, so it survives; the count-cap test explicitly asserts the newest files are kept. Reads of a purged file return null (callers 404/fall back), documented in the PR body.
3. Bucketed layout mirrored everywhere: maybePersistRtkRawOutput writes buckets, readRtkRawOutput reads bucket-first with guarded legacy fallback, listRtkCommandSamples scans flat+buckets, purgeRtkRawOutput covers both layouts — no path reads only one layout.

Issue #7065, opened 2026-07-13
[BUG]  Its not opening


isubroto:
### OmniRoute Version

3.8.47

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 25H2

### Node.js Version

26.3.0

### Description

PS C:\Users\anmon> omniroute
  📋 Loaded env from C:\Users\anmon\.omniroute\.env
  📋 Loaded env from C:\Users\anmon\AppData\Roaming\npm\node_modules\omniroute\.env

   ____                  _ ____              _
   / __ \                (_) __ \            | |
  | |  | |_ __ ___  _ __ _| |__) |___  _   _| |_ ___
  | |  | | '_ ` _ \| '_ \ |  _  // _ \| | | | __/ _ \
  | |__| | | | | | | | | | | | \ \ (_) | |_| | ||  __/
   \____/|_| |_| |_|_| |_|_|_|  \_\___/ \__,_|\__\___|

  v3.8.47

  ⏳ Starting server...

⚠ Server exited (code=1). Restarting in 1s... (1/2)

--- Server crash log ---
node:internal/modules/esm/resolve:271
    throw new ERR_MODULE_NOT_FOUND(
          ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'C:\Users\anmon\AppData\Roaming\npm\node_modules\omniroute\dist\head-response-guard.cjs' imported from C:\Users\anmon\AppData\Roaming\npm\node_modules\omniroute\dist\server-ws.mjs
    at finalizeResolution (node:internal/modules/esm/resolve:271:11)
    at moduleResolve (node:internal/modules/esm/resolve:865:10)
    at defaultResolve (node:internal/modules/esm/resolve:992:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:704:20)
    at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:724:38)
    at ModuleLoader.resolveSync (node:internal/modules/esm/loader:762:56)
    at #resolve (node:internal/modules/esm/loader:686:17)
    at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:606:35)
    at ModuleJob.syncLink (node:internal/modules/esm/module_job:163:33)
    at ModuleJob.link (node:internal/modules/esm/module_job:253:17) {
  code: 'ERR_MODULE_NOT_FOUND',
[...]

PR #10662, opened 2026-08-18
feat(server): native systemd sd_notify watchdog (Type=notify)

### LINKED ISSUE #10661: feat(startup): Support systemd sd_notify (Type=notify + WatchdogSec) for native frozen-process detection
### Problem / Use Case

`Restart=` only kicks in when the process exits. If the event loop gets blocked, the server stays alive, systemd sees nothing wrong and never restarts it. The current systemd unit is `Type=simple` with no `WatchdogSec`, and the server never sends sd_notify messages, so no systemd deployment can catch this kind of failure on its own.

It actually happened in #8697: the cold `/v1/models` catalog rebuild blocked the main loop for 45-110s while the process stayed alive (process alive, port open, zero progress). Right now the only workaround is an external liveness probe — a watchdog unit polling an HTTP endpoint — that every operator has to build, tune and maintain themselves.

### Proposed Solution

Let systemd do the detection natively:

1. A small module (`scripts/dev/systemd-notify.mjs`) that sends `READY=1` / `WATCHDOG=1` / `STOPPING=1` through the `systemd-notify` binary. No new dependency, the binary ships with systemd (Node's `dgram` has no AF_UNIX datagram sockets). Active only when `NOTIFY_SOCKET` is set, so Docker/Electron/plain terminals are untouched; opt-out via `OMNIROUTE_DISABLE_SD_NOTIFY=1`; one warning + disable on send failure.
2. Wire it into both server entry points (`omniroute serve` → `server-ws.mjs`, `npm start`/dev → `run-next.mjs`): `READY=1` once listening, `WATCHDOG=1` every 60s from the main event loop (a blocked loop stops the pings), `STOPPING=1` on shutdown.
3. Generate `Type=notify` + `WatchdogSec=180` + `NotifyAccess=all` + `TimeoutStartSec=300` in the autostart unit (`bin/cli/tray/autostart.mjs`), keeping `Restart=on-failure` so a watchdog kill triggers a restart.

Unit before → after:

```
[Service]                  [Service]
Type=simple                Type=notify
ExecStart=... serve ...    NotifyAccess=all
Restart=on-failure         WatchdogSec=180
RestartSec=5               TimeoutStartSec=300
                           ExecStart=... serve ...
                           Restart=on-failure
                           RestartSec=5
```

### Alternatives Considered

External liveness probe — works today but every operator has to build, tune and maintain it themselves, which is exactly the burden this proposal removes. Also considered writing to `NOTIFY_SOCKET` directly from Node, but the stable `dgram` API has no AF_UNIX datagram sockets, hence the `systemd-notify` binary.

### Acceptance Criteria

- With `NOTIFY_SOCKET` set, the server sends `READY=1` once listening and `WATCHDOG=1` pings from the main event loop; pings stop while the loop is blocked; `STOPPING=1` is sent on graceful shutdown.
- A frozen server under the generated unit is killed and restarted by systemd after the watchdog timeout; a slow-but-alive server (freeze ≤ 110s) is not killed.
- Without `NOTIFY_SOCKET`, behavior is strictly unchanged; `OMNIROUTE_DISABLE_SD_NOTIFY=1` disables the module; send failure → one warning + disable, service keeps running.
- Generated unit contains `Type=notify`, `WatchdogSec=180`, `NotifyAccess=all`, `TimeoutStartSec=300`, `Restart=on-failure`.

### Area

Docker / Deployment

### Related Provider(s)

_No response_

### Additional Context

Design notes (open to discussion):

- `WatchdogSec=180` with 60s pings — 180s is >1.5× the worst legitimate freeze measured in the field (110s, #8697), deliberately generous so a slow-but-alive server isn't killed. The 60s ping satisfies any `WatchdogSec >= 120s` an operator sets themselves (systemd requires pings at most every `WatchdogSec/2`).
- `NotifyAccess=all` — the pings come from the server child process, not from the `serve` supervisor that spawns it, and the default `NotifyAccess=main` silently drops everything else. The unit's cgroup only contains the serve process tree, so `all` doesn't really widen anything.
- No auto-migration of existing units — operators regenerate via autostart when they want the new behavior; a unit silently changing shape behind them would be worse.
- `--daemon` is out of scope — the daemon child escapes the cgroup (detached spawn), which is incompatible with `Type=notify`; the generated unit runs foreground anyway.

### Expected Test Plan

_No response_

maxmad64bis:
## Summary

When the server process freezes (event loop blocked, process alive), systemd cannot see it: `Restart=` only reacts to process exits. OmniRoute currently sends no sd_notify messages, so systemd deployments of it need an external liveness probe to catch this. It actually happened in production (#8697): a cold `/v1/models` catalog rebuild blocked the event loop for 45–110 s while the process stayed alive.

This PR wires OmniRoute into systemd natively:

- New `scripts/dev/systemd-notify.mjs` — sends `READY=1` / `WATCHDOG=1` / `STOPPING=1` through the `systemd-notify` binary (ships with systemd; Node's `dgram` has no AF_UNIX datagram sockets, so no new dependency). Active only when `NOTIFY_SOCKET` is set — Docker/Electron/Windows/plain-terminal runs are untouched — opt-out via `OMNIROUTE_DISABLE_SD_NOTIFY=1`, one warning + disable on send failure.
- Wired into both server entry points: `standalone-server-ws.mjs` (`omniroute serve`, READY on first `listening`) and `run-next.mjs` (`npm start`/dev, READY in the listen callback). WATCHDOG pings every 60 s run in the serving process's main event loop, so a blocked loop stops the pings.
- The systemd user unit generated by `bin/cli/tray/autostart.mjs` moves from `Type=simple` to `Type=notify` with `WatchdogSec=180`, `NotifyAccess=all`, `TimeoutStartSec=300`. `NotifyAccess=all` is needed because the pings come from the server child process, not from the `serve` supervisor. `--daemon` stays out (child escapes the cgroup — incompatible with `Type=notify`).

## Related Issues

- Closes #10661
- Related to #8697

## Validation

- [x] Change type: other (server lifecycle / systemd integration)
- [x] `npm run lint` (touched lintable files; `scripts/dev/*.mjs` is excluded by repo config, consistent with siblings)

⚠️ base-red inherited: #9985 (pre-existing typecheck error in `omniglyphAdapter.ts` on the base; not introduced by this PR).

## Tests Added Or Updated

- `tests/unit/systemd-notify.test.mjs` — new (11 tests): guard behavior (no-op without `NOTIFY_SOCKET`, opt-out wins), notifier-level opt-out, message format, send failure → single warning + disable, watchdog interval + dispose, cadence-vs-WatchdogSec contract.
- `tests/unit/cli/autostart-linux.test.ts` — new case that forces the systemd branch with succeeding systemctl/loginctl stubs and asserts the generated unit directives (`Type=notify`, `NotifyAccess=all`, `WatchdogSec=180`, `TimeoutStartSec=300`, `Restart=on-failure`).
- `tests/unit/pack-artifact-policy.test.ts` — golden list updated with `dist/systemd-notify.mjs`.
- `tests/unit/pack-artifact-server-ws-closure.test.ts` — pass unchanged; it derives the tarball list from the wrapper's own imports, which is what proves the new module ships.
- Live test (Hard Rule #18): throwaway systemd user unit, `Type=notify`, `NotifyAccess=all`, `WatchdogSec=15`, `Restart=on-failure`, running the real module (READY + pings every 5 s, then a simulated event-loop freeze). systemd killed the service and restarted it 6 times in a row:

```
Aug 18 13:58:46 systemd[237]: sd-notify-live-test.service: Watchdog timeout (limit 15s)!
Aug 18 13:58:46 systemd[237]: sd-notify-live-test.service: Killing process 306181 (node) with signal SIGABRT.
Aug 18 13:58:46 systemd[237]: sd-notify-live-test.service: Failed with result 'watchdog'.
Aug 18 13:58:47 systemd[237]: sd-notify-live-test.service: Scheduled restart job, restart counter is at 5.
...restart → READY → freeze → kill ×6 (NRestarts=6)
```

Focused run after the review fixes:

```
$ node --import tsx/esm --test tests/unit/systemd-notify.test.mjs tests/unit/cli/autostart-linux.test.ts tests/unit/cli/autostart.test.ts tests/unit/cli-tray.test.ts tests/unit/pack-artifact-policy.test.ts tests/unit/pack-artifact-server-ws-closure.test.ts tests/unit/build/assemble-standalone.test.ts

# tests 46
# pass 46
# fail 0
```

## Coverage Notes

- The new module is plain `.mjs` (shipped as-is in the standalone assembly — no TS allowed there), fully covered by its unit test (guard, message format, failure path, interval lifecycle).
- The unit-generation change is covered by the new autostart-linux case (headless, HOME-isolated, stubs succeeding). `systemd-notify.test.mjs` also asserts the cadence contract with the generated unit (`2 × ping interval ≤ WatchdogSec`), so a drift on either side fails loudly.
- The full `tests/unit` suite was not run on the contributing machine (WSL RAM/time budget); CI runs it with coverage on this PR (#8329).

## Reviewer Notes

- Manual validation: regenerate the unit (`omniroute autostart enable` on Linux), start it, confirm `systemctl --user show ... -p Type` reports `notify` and the service ends up `ActiveState=active`. The READY handshake is the risky part on real boots (cold start, migrations) — hence `TimeoutStartSec=300`.
- Live test to reproduce: a small unit (`Type=notify`, `NotifyAccess=all`, `WatchdogSec=15`, `Restart=on-failure`) running any script that imports the module, sends READY, pings, then busy-loops; the unit must die with `result='watchdog'` and restart.
- The new module is listed in the standalone assembly manifest (`assembleStandalone.mjs`) AND in `pack-artifact-policy.ts` — the latter is enforced by the closure test, so a missing entry ships a tarball that crashes on boot (ERR_MODULE_NOT_FOUND class, #7065).
- Existing installs keep their old `Type=simple` unit until regenerated; no auto-migration on purpose.
- Ping cadence is fixed at 60 s and satisfies any `WatchdogSec >= 120 s` an operator sets themselves.
- `--daemon` + `Type=notify` is incompatible by design (documented in the unit comment and the module header).


diegosouzapw:
Thanks for the native sd_notify watchdog — it fills a real gap and fits the existing serverLifecycle seam (src/lib/serverLifecycle.ts). A couple of pre-merge items: (1) the branch currently reports a merge conflict against release/v3.8.50 and needs a rebase; (2) please confirm the sd_notify path is gated on NOTIFY_SOCKET so dev/Docker/supervisor runs are no-ops, and add a unit test covering the socket handshake (fake NOTIFY_SOCKET on a temp socket, assert READY=1/WATCHDOG=1). Final Type=notify handshake validation should also be run on a real systemd host. Happy to help verify once the conflict is resolved.

maxmad64bis:
Both pre-merge items are done — the branch is rebased and GitHub reports it MERGEABLE.

On the handshake: NOTIFY_SOCKET gating was already in place and tested (OMNIROUTE_DISABLE_SD_NOTIFY=1 force-disables, zero spawn/timer when disabled). I added the real socket test you asked for: a temp AF_UNIX datagram socket, NOTIFY_SOCKET pointed at it, and the real systemd-notify delivering READY=1 / WATCHDOG=1 / STOPPING=1 through it. 12/12 green, lint clean. And I validated on a real systemd host (Type=notify, systemd 255): with READY=1 the unit succeeds, without it systemd kills it with a protocol error — so readiness really is gated by the notification, not by the process exiting.

The CI reds are the same base-red as #10687/#10679 (compression/omniglyph + the shared quality gates), nothing from this PR. Ready for re-review when you have a moment.

PR #10663, opened 2026-08-18
fix(api): classify OAuth probe timeout as network_error

### LINKED ISSUE #9623: fix(backend): Failed connection test leaves testStatus=error with no recovery path

HouMinXi:
## Summary

The OAuth probe reports its own abort as "Test timed out after 30s", but `classifyFailure` only matched `timeout`, not `timed out`. A hung probe was diagnosed as a generic upstream error and marked the connection broken.

Add `timed out` to the network error pattern so transient probe failures don't paint connections permanently red.

Complements the existing #9623 fix (30s cooldown for non-terminal failures) by correctly classifying the OAuth probe timeout as a network error.

Fixes #9623

## Test plan

- `node --import tsx/esm --test tests/unit/connection-test-timed-out-network-error.test.ts` (new regression test, added during review)

### DIFF
--- a/src/app/api/providers/[id]/test/route.ts
+++ b/src/app/api/providers/[id]/test/route.ts
@@ -153,6 +153,7 @@ export function classifyFailure({
     normalized.includes("fetch failed") ||
     normalized.includes("network") ||
     normalized.includes("timeout") ||
+    normalized.includes("timed out") ||
     normalized.includes("econn") ||
     normalized.includes("enotfound") ||
     normalized.includes("socket")
--- a/tests/unit/connection-test-timed-out-network-error.test.ts
+++ b/tests/unit/connection-test-timed-out-network-error.test.ts
@@ -1,0 +1,29 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+const { classifyFailure } = await import("../../src/app/api/providers/[id]/test/route.ts");
+
+// The OAuth probe's own abort message is "Test timed out after Xs" (route.ts's
+// AbortSignal.timeout handling), which contains "timed out" but not "timeout" —
+// classifyFailure must classify it as network_error, not the generic upstream_error
+// fallback, so a transient probe timeout doesn't paint the connection permanently red.
+test("classifyFailure maps an OAuth-probe 'timed out' message to network_error", () => {
+  const diagnosis = classifyFailure({
+    error: "Test timed out after 30s",
+    statusCode: null,
+    provider: "some-oauth-provider",
+  });
+
+  assert.equal(diagnosis.type, "network_error");
+  assert.equal(diagnosis.code, "network_error");
+});
+
+test("classifyFailure still maps the existing 'timeout' substring to network_error", () => {
+  const diagnosis = classifyFailure({
+    error: "connect ETIMEDOUT — timeout while probing upstream",
+    statusCode: null,
+    provider: "some-oauth-provider",
+  });
+
+  assert.equal(diagnosis.type, "network_error");
+});


diegosouzapw:
Thanks for the fix — root cause checks out: `classifyFailure` only matched `"timeout"`, but the OAuth probe's own abort message is `Test timed out after ${s}s`, so it fell into the generic `upstream_error` bucket instead of `network_error`. I reproduced this locally (reverting your one-line change makes `classifyFailure({error:"Test timed out after 30s"})` return `upstream_error`; with your change it correctly returns `network_error`), and the existing `classifyFailure`-related suites (deepseek-web-auth-semantics, provider-test-mistral-401-classify, provider-test-account-deactivated, proxy-connection-test, repro-9623) all still pass.

Two things before this can merge:

1. The diff has no test changes, and the "Test plan" in the description points at `tests/unit/connection-test-network-error-status-9623.test.ts`, which doesn't exist in the repo (not on the base branch, not added here). The repo's Hard Rule #18 requires a TDD regression test for every bug fix — could you add a small unit test that imports `classifyFailure` from `src/app/api/providers/[id]/test/route.ts` and asserts `classifyFailure({ error: "Test timed out after 30s" }).type === "network_error"`? That's exactly the case this PR fixes.
2. Please update the PR's "Test plan" section to point at whatever test file you actually add, so reviewers can run it as written.

Once the test is in, this is good to go — small, correct, and consistent with the existing `"timed out" || "timeout"` pattern already used in `open-sse/services/qoderCli.ts`.

PR #10679, opened 2026-08-18
feat(call_logs): persist per-call error family and expose analytics error breakdown

### LINKED ISSUE #10670: feat(resilience): Persist per-call error classification in call_logs and expose error-type breakdown in usage analytics
## Problem / Use Case

OmniRoute classifies provider errors on every call (`classifyProviderError`, open-sse/services/errorClassifier.ts:239) but only uses the result to pick a cooldown, then drops it. The chat path calls it in open-sse/handlers/chatCore.ts:3974 and src/sse/services/auth.ts:2598; the only thing persisted is `provider_connections.last_error_type` (src/lib/db/core.ts:245), which keeps the latest state of a connection, not the history.

`call_logs.error_summary` is free text, truncated to 4000 chars at write time (src/lib/usage/callLogs/format.ts:57). So counting failures by family over a time window — "how many rate_limited vs quota_exhausted vs forbidden responses per provider in the last 24h" — means scraping that text, or is just not possible. `/api/usage/analytics` has totals (errors, latency, cost) but no error-type breakdown.

## Proposed Solution

Add a nullable `error_type` column to `call_logs` (migration 157) and fill it at the single write point, `saveCallLogOperation` (src/lib/usage/callLogs.ts:438), which already receives status, error and provider — no caller changes needed. Callers pass both strings and `Error` objects, so the error is normalized to text first (`typeof entry.error === "string" ? entry.error : entry.error?.message ?? ""`), then the existing `classifyProviderError` is reused.

Expose the breakdown (count per error_type, optionally per provider, within the selected time range) from `/api/usage/analytics`, following the aggregation pattern in src/lib/db/usageAnalytics.ts — no new endpoint.

## Alternatives Considered

- Writing to `request_detail_logs` (migration 006): opt-in and capped at 500 rows by a SQL trigger — too small for analytics.
- Parsing `error_summary` at query time: fragile, provider messages vary, and the regexes already live in `classifyProviderError`.
- `provider_connections.last_error_type`: last state only, no history.

## Acceptance Criteria

- `call_logs` gains a nullable `error_type` column; rows written after the migration carry the classification, null when the classifier returns null (unclassified: resource-scoped 404s, 403 on API-key providers, unmatched statuses).
- `/api/usage/analytics` returns the breakdown for the selected range, with a bucket for null/unclassified — old rows all fall in it.
- No changes to the 121 `saveCallLog` callers.
- Backend only, no UI strings.

## Area

Analytics / Usage Tracking

## Related Provider(s)

All providers (provider-agnostic)

## Additional Context

- The taxonomy defines 13 types but the classifier only emits 12 today (EMPTY_CONTENT is declared, never returned).
- No new index: the existing analytics queries already scan by timestamp range, same pattern as the current code.
- The write point persists to disk only when enabled (shouldPersistToDisk), same behavior as today.

## Expected Test Plan

- Unit tests in the tests/unit/call-log-*.test.ts family: classification at the write point (string error, Error object, null), migration applies.
- npm run check:db-rules, npm run check:migration-numbering, npm run lint.

maxmad64bis:
## Summary

Call logs record free-text error summaries but no machine-readable error class, so aggregating failures by family has to be done outside the app. This PR persists the error family at the write point and exposes it through the usage analytics endpoint.

- A new `error_type` column on `call_logs` (migration 158) is filled by `saveCallLogOperation`, the single place call logs are written. It reuses the existing `classifyProviderError` classifier, so the vocabulary stays identical to the one used at runtime by the chat path and auth flows.
- `GET /api/usage/analytics` now returns an `errorBreakdown` array (family + count, with an explicit `unclassified` bucket for failures the classifier does not recognize and for pre-migration rows). Successes are excluded in SQL and never appear.
- All 121 `saveCallLog` callers are untouched; only the write point changed.

## Related Issues

- Closes #10670

## Validation

- [x] Change type: DB / other
- [x] Focused tests and category gates from the golden path (`check:db-rules`, `check:migration-numbering`, focused unit suite)
- [x] `npm run lint`

Focused unit suite: 10/10 passing (`tests/unit/call-log-error-type.test.ts`), 8/8 on the existing `save-call-log-persistence.test.ts`. `check:db-rules` OK, `check-migration-numbering` exit 0, ESLint clean on all touched files. `typecheck:core` reports nothing on the files touched here; the pre-existing omniglyph diagnostics on `release/v3.8.50` are covered by the base-red issue.

## Tests Added Or Updated

- `tests/unit/call-log-error-type.test.ts` (new) — column migration, classifier mapping (status + string / Error / unclassifiable), write-point persistence, breakdown grouping, and the empty-where-clause path.

## Coverage Notes

- The classifier helper is a pure function exercised directly, the write point is covered end-to-end through `saveCallLog` against the real SQLite instance, and the breakdown query is asserted on a controlled set of rows including an `unclassified` bucket and a `range=all`-style empty filter. Coverage of the touched files is not expected to move down.

## Reviewer Notes

- Migration numbering: 157 was taken by upstream (the exclusive-connection-leases migration was renumbered 155 → 157 in `59c8a9afc` after the issue was opened), so this PR uses 158. `check-migration-numbering` passes on this branch.
- `error_type` is intentionally left NULL for successful calls; the analytics query filters failures in SQL.
- ⚠️ base-red inherited: #9985 (`typecheck:core` fails on `omniglyphAdapter.ts` against omniglyph 1.4.0; pre-existing on the release base, untouched here).

### DIFF
--- a/changelog.d/features/10670-call-logs-error-type.md
+++ b/changelog.d/features/10670-call-logs-error-type.md
@@ -1,0 +1,1 @@
+- **feat(call_logs):** persist the per-call error family in `call_logs.error_type` and expose a failure breakdown (`errorBreakdown`) in the usage analytics endpoint, reusing the existing production classifier ([#10670](https://github.com/diegosouzapw/OmniRoute/issues/10670))
--- a/src/app/api/usage/analytics/route.ts
+++ b/src/app/api/usage/analytics/route.ts
@@ -21,7 +21,7 @@ import {
   getWeeklyPatternRows,
   getPresetCostModelRows,
 } from "@/lib/db/usageAnalytics";
-import { getFallbackStats } from "@/lib/db/callLogStats";
+import { getFallbackStats, getErrorTypeBreakdown } from "@/lib/db/callLogStats";
 import { buildByProviderRows } from "@/lib/usage/providerDisplayNames";
 import { toNumber } from "@/shared/utils/numeric";
 
@@ -481,6 +481,7 @@ export async function GET(request: Request) {
     const weeklyRows = getWeeklyPatternRows(unifiedSource, unifiedParams) as UsageRows;
 
     const fallbackRow = getFallbackStats(whereClause, params) as Record<string, unknown>;
+    const errorBreakdown = getErrorTypeBreakdown(whereClause, params);
 
     const summary = {
       totalRequests: Number(summaryRow?.totalRequests || 0),
@@ -869,6 +870,7 @@ export async function GET(request: Request) {
       weeklyCounts,
       dailyByModel,
       modelNames,
+      errorBreakdown,
       range,
     } as any;
 
--- a/src/lib/db/callLogStats.ts
+++ b/src/lib/db/callLogStats.ts
@@ -233,3 +233,34 @@ export function getFallbackStats(
     .get(params) as FallbackStatsRow | undefined;
   return row ?? { total: 0, with_requested: 0, fallback_eligible: 0, fallbacks: 0 };
 }
+
+/**
+ * Failure-family breakdown over `call_logs` for the usage analytics endpoint.
+ * Failures are rows with status >= 400 or a non-empty error summary; successes
+ * are excluded in SQL. Pre-migration rows and failures the classifier does not
+ * recognize (null family) land in the explicit `unclassified` bucket.
+ *
+ * @param whereClause - SQL WHERE clause (may be empty string) using the same
+ *                      named params as the usage_history queries.
+ * @param params      - Named params object (string values).
+ */
+export function getErrorTypeBreakdown(
+  whereClause: string,
+  params: Record<string, string>
+): Array<{ errorType: string; count: number }> {
+  const db = getDbInstance();
+  const rows = db
+    .prepare(
+      `
+      SELECT
+        COALESCE(error_type, 'unclassified') AS errorType,
+        COUNT(*) AS count
+      FROM call_logs
+      ${whereClause} ${whereClause ? "AND" : "WHERE"} (status >= 400 OR error_summary IS NOT NULL)
+      GROUP BY 1
+      ORDER BY count DESC, errorType ASC
+      `
+    )
+    .all(params) as Array<{ errorType: string; count: number }>;
+  return rows.map((row) => ({ errorType: String(row.errorType), count: Number(row.count) }));
+}
--- a/src/lib/db/migrations/158_call_logs_error_type.sql
+++ b/src/lib/db/migrations/158_call_logs_error_type.sql
@@ -1,0 +1,4 @@
+-- #10670: per-call error family, set at the single write point in
+-- src/lib/usage/callLogs.ts from classifyProviderError. NULL for successes
+-- (analytics filters by failure in SQL, so no need for a sentinel value).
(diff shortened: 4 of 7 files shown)


diegosouzapw:
Review hold: the standard review pass for this PR could not be completed (analysis tooling unavailable in this session), so no verdict on the changes has been reached and no defects are being alleged. Please hold for a full review before merge; the repository's Hard Rule #18 (test coverage for production code changes) and the auth/error-sanitization patterns for the new analytics endpoint still need verification.

maxmad64bis:
Both points are covered. One correction to the premise first: **this PR does not add an endpoint.** The only change to `src/app/api/usage/analytics/route.ts` is +3/-1 — one import, one call, one extra field on the response object that route already builds. The route itself, its auth and its error handling are untouched.

## Hard Rule #18 — test coverage

`tests/unit/call-log-error-type.test.ts` covers each of the four production changes:

| Production change                                  | Covering tests                                                                                                                                              |
| -------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `158_call_logs_error_type.sql`                     | `call_logs table has error_type column`                                                                                                                     |
| `classifyCallLogError()` (`callLogs/format.ts`)    | status+body maps to the provider family · classifies only failures · extracts the message from an `Error` object · unclassifiable provider-403 returns null |
| write path (`callLogs.ts`, `saveCallLogOperation`) | persists `error_type` on a failure · persists null on a success · normalizes an `Error` before classifying                                                  |
| `getErrorTypeBreakdown()` (`callLogStats.ts`)      | groups failures by family and excludes successes · empty `whereClause` does not crash                                                                       |

```
node --import tsx/esm --test tests/unit/call-log-error-type.test.ts
# tests 10 / pass 10 / fail 0
```

ESLint on the five touched files is clean (exit 0, run with `--suppressions-location config/quality/eslint-suppressions.json`).

## Auth

Unchanged, and nothing new was needed. `/api/usage/analytics` carries no per-route auth code and never did — it is gated by the authorization pipeline before routing, like the rest of `/api/usage/`. It is not a spawn-capable surface, so it stays out of `LOCAL_ONLY_API_PREFIXES` and `src/server/authz/routeGuard.ts` needs no entry. A request without a key gets `401 AUTH_001 Authentication required` before the handler runs, exactly as it did before this PR.

## Error sanitization

The new field cannot carry provider text, by construction. `errorBreakdown` is `Array<{ errorType: string; count: number }>`, where `errorType` is `COALESCE(error_type, 'unclassified')` and `error_type` is only ever written from `classifyProviderError()`. Every return in that function is one of the 13 `PROVIDER_ERROR_TYPES` constants or `null` (`open-sse/services/errorClassifier.ts:69-83`), so the response can only ever contain those 13 literals plus `unclassified`. `error_summary` is used in the `WHERE` clause and is never selected.

For the failure path: the route's existing `catch` already goes through `buildErrorBody()` (`route.ts:924-930`), and `getErrorTypeBreakdown()` runs inside that same `try`, so a DB error there is sanitized by the path that is already in place.

## One note back

The migration number is still free: `157_exclusive_connection_leases.sql` is the highest on `release/v3.8.50` at the moment, so `158` does not collide. If another `158` lands first, ping me or rebase directly and I will renumber.

PR #10682, opened 2026-08-18
test(aihorde): guard client browser bundles

cryptiklemur:

`release/v3.8.50` now contains the browser-safe AI Horde registry implementation. this PR keeps the focused regression test requested in review.

## what it covers

- bundle the media cache client for the browser
- bundle the provider detail client for the browser
- fail if either entry pulls database, filesystem, or other server-only code into the client graph

The broader client-entry guard on the release branch covers the same class of bug. this test keeps the two original failure paths explicit.

## how to test

```sh
npx cross-env DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=4096 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-concurrency=1 tests/unit/media-page-client-browser-bundle.test.ts
npx eslint tests/unit/media-page-client-browser-bundle.test.ts
```

Both commands pass after rebasing onto the current `release/v3.8.50` tip.

### DIFF
--- a/tests/unit/media-page-client-browser-bundle.test.ts
+++ b/tests/unit/media-page-client-browser-bundle.test.ts
@@ -1,0 +1,34 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import { fileURLToPath } from "node:url";
+
+import { build } from "esbuild";
+
+const REPO_ROOT = fileURLToPath(new URL("../..", import.meta.url));
+
+async function assertBrowserBundleSafe(relativePath: string) {
+  await assert.doesNotReject(
+    build({
+      absWorkingDir: REPO_ROOT,
+      entryPoints: [fileURLToPath(new URL(relativePath, import.meta.url))],
+      bundle: true,
+      format: "esm",
+      logLevel: "silent",
+      platform: "browser",
+      tsconfig: "tsconfig.json",
+      write: false,
+    })
+  );
+}
+
+test("media page client entry stays browser-bundle safe", async () => {
+  await assertBrowserBundleSafe(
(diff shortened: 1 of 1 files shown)


diegosouzapw:
Thanks for this fix — the aihorde image catalog currently breaks browser-safety on the media cache page and it's worth shipping. The defect is still present on release/v3.8.50: `IMAGE_PROVIDERS` (open-sse/config/imageRegistry.ts:860) includes `aihorde: AI_HORDE_IMAGE_PROVIDER`, whose `models` getter pulls in `aihordeImageCatalog.ts` → `safeOutboundFetch` (module-scope `process.env` + Node-only proxy/undici deps), and `src/app/(dashboard)/dashboard/cache/media/MediaPageClient.tsx` is `"use client"` and evaluates `toProviderModels(IMAGE_PROVIDERS)` at module scope — so the browser bundle ships Node-only code and will break. Keeping the live catalog server-side and exposing a browser-safe model set to the client is the right approach.

Two pre-merge items:
1. This PR is currently CONFLICTING with the release/v3.8.50 base — please rebase.
2. Per repo Hard Rule #18, a bug fix on production code needs a regression test. The existing `tests/unit/aihorde-image-catalog.test.ts` is a Node `node:test` and cannot catch browser-unsafety — please add a test that proves the client-imported chain (e.g. the browser-safe aihorde model projection, or `MediaPageClient`'s import path) loads without `process.env` / Node-only `safeOutboundFetch`.

Note: the ~25 ESLint base-reds from #10692/#10695 (issue #9985) are inherited from the release tip, not caused by this PR.

Issue #9567, opened 2026-08-06
test: chatCore/SSE stream-mode flaky cluster under loaded runners (recurred on 2 unrelated PRs, 2026-08-06)


diegosouzapw:
## Cluster

The same group of timing/stream-sensitive unit tests red Unit fast-path shards under loaded GitHub runners, then pass on re-run / other runs of the same base commit:

- `chatCore honors API key JSON stream-default compatibility mode`
- `chatCore resolves stream mode from body.stream and Accept header`
- `chatCore treats Nextcloud OpenAI integration requests as non-streaming by default`
- `handleChat treats a pure Accept: text/event-stream as stream=true ...`
- `shared SSE response holds the lease until consumer cancellation`
- `chat completions route emits early keepalive while waiting for stream readiness`
- `chatCore caches streaming response and serves cache HIT on repeat`
- `i18n pt-BR integrity — should contain every key present in en.json (#6695)`
- `provider models route retries Antigravity discovery endpoints ...`
- gpt-5.5 id-fragment heuristic drift pair

## Evidence

- Reproduced on **two unrelated PRs** cut from the same clean tip on 2026-08-06: #9553 (build script only) and #9559 (mcp audit test seam only) — neither touches chatCore/SSE/i18n.
- On #9553 the same shards were GREEN on the previous run of the same base (pre-push run), red on the next — classic load-dependent nondeterminism.
- Local isolated runs of these files pass.

## Ask

Same treatment as #6803/#6897 (the combo cooldown de-flake): controlled clocks / readiness polling instead of real-time waits for the stream-mode and keepalive tests, and a look at whether `--test-concurrency` on ubuntu-latest runners is oversubscribed.

Refs #6803 #9534

PR #10683, opened 2026-08-18
test(chatcore): wait for queued call log writes

cryptiklemur:

The chatcore translation suite could read `call_logs` before queued writes finished. latest-record assertions then saw stale rows.

This PR only changes test timing. production behavior stays the same.

The shared helper now waits for `waitForCallLogSaves(5000)` before reading the latest call log. `release/v3.8.50` owns the SSE comment behavior updated in #10726, so this branch leaves those assertions unchanged.

## how to test

```sh
npx cross-env DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=4096 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-concurrency=1 tests/unit/chatcore-translation-paths.test.ts
```

The focused file passes after rebasing onto the current release tip. direct-file ESLint still reports 31 `no-explicit-any` findings present on the release base; this three-line diff adds none.

### DIFF
--- a/tests/unit/chatcore-translation-paths.test.ts
+++ b/tests/unit/chatcore-translation-paths.test.ts
@@ -36,7 +36,8 @@ const {
   setBackgroundDegradationConfig,
   resetStats: resetBackgroundStats,
 } = await import("../../open-sse/services/backgroundTaskDetector.ts");
-const { getCallLogs, getCallLogById } = await import("../../src/lib/usage/callLogs.ts");
+const { getCallLogs, getCallLogById, waitForCallLogSaves } =
+  await import("../../src/lib/usage/callLogs.ts");
 const {
   handleChatCore,
   shouldUseNativeCodexPassthrough,
@@ -286,6 +287,7 @@ async function flushAsyncSideEffects() {
 }
 
 async function getLatestCallLog() {
+  await waitForCallLogSaves(5000);
   const rows = await getCallLogs({ limit: 5 });
   if (!Array.isArray(rows) || rows.length === 0) return null;
   return getCallLogById(rows[0].id);


diegosouzapw:
Thanks — this fixes a real latent bug: `resolveTargetTokenLimit` in `open-sse/services/combo.ts` was calling the **async** `getCachedProviderConnectionById` without `await`, so the per-connection `rateLimitOverrides.tpm` token-limit override was silently never applied in round-robin combo, and it imported from the `localDb.js` barrel instead of the canonical `db/readCache.ts`. The regression test plus the chatcore streaming-log assertion isolation (waiting for queued call-log saves via `waitForCallLogSaves`, scoping `OMNIROUTE_SSE_COMMENTS=on` to the one test and restoring it in `afterEach`/`after`) are exactly the right hardening, and they follow existing precedent (#9567 polling; `sse-comments-optout-9305` env save/restore).

Note: the title and PR body call this a "test-only follow-up," but the PR also carries a production change in `open-sse/services/combo.ts` — worth reflecting in the title/conventional-commit (e.g. `fix(combo)`) so the changelog and CI gates (which treat production changes specially) match.

Before this can merge, the branch needs:
1. **Rebase/conflict resolution** — GitHub reports `mergeable=false` / `dirty` against `release/v3.8.50`; the `chatcore-translation-paths.test.ts` and `combo.ts` hunks need re-applying onto the current tip.
2. **Get CI green** — Unit Tests fast-path (4/4 shards) and Vitest are failing and this does not fall in the documented inherited-base-red list (#9985), so it looks branch-caused. If the async change to `resolveTargetTokenLimit` interacts with existing sync mocks of `getCachedProviderConnectionById`, or if the new `combo-quota-token-limit.test.ts` leaks SQLite handles (native runner hangs unless `resetDbInstance()` + `test.after` cleanup is present — see AGENTS.md PII/stream learnings), those need fixing. Please run the full unit + vitest suites locally and confirm.
3. **Merge integrity gate** (changelog + generated skills, currently exit 2) — since production code changed, add a CHANGELOG entry for the combo fix.

Keep the existing tests green; do not weaken/remove the new assertions to pass. Happy to re-review once the conflict is resolved and CI is green.

PR #10684, opened 2026-08-18
fix(sse): advance Claude cache breakpoints on growing tails

cryptiklemur:

Claude Code requests stopped advancing the cached prefix after an older message breakpoint appeared. later turns stayed outside the reusable prefix even as the conversation grew.

Anthropic requires longer TTL breakpoints before shorter ones and checks the prompt in tools, system, then messages order: [Prompt caching](https://platform.claude.com/docs/en/build-with-claude/prompt-caching).

## what changed

- count surviving system and message breakpoints together
- add a breakpoint to the latest user message while fewer than four remain
- keep repeated normalization idempotent
- default later omitted TTLs to `5m` after a `5m` boundary, preserving `1h` before `5m`

The multi-turn tests cover the final user message, the four-breakpoint cap, repeated calls, both hoist paths, and native Claude passthrough.

## how to test

```sh
npx cross-env DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=4096 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-concurrency=1 tests/unit/claude-code-parity.test.ts
npx cross-env DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=4096 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-isolation=none --test-concurrency=1 --test-name-pattern='chatCore (builds Claude Code-compatible|normalizes native Claude Code messages|preserves Opus 5 mid-conversation|keeps Claude normalization|preserves cache_control automatically|advances a message cache breakpoint)' tests/unit/chatcore-translation-paths.test.ts
```

Both focused runs pass after rebasing onto the current `release/v3.8.50` tip.


diegosouzapw:
Thanks for this fix — advancing the prompt-cache breakpoint onto the growing conversation tail closes a real gap (the current ensureCacheControlOnLastUserMessage keeps an existing marker pinned at the first turn, so the cached prefix stops growing on long sessions). A couple of things to confirm before merge: (1) rebase against the current release/v3.8.50 tip — the PR is currently marked as conflicting; (2) please add/keep a unit test covering a multi-turn payload asserting the marker is advanced to the final user message without exceeding the 4-breakpoint cap, and that the existing #9436 boundary tests (both hoist paths + native Claude path) still pass; (3) ensure no double-advance leaves duplicate markers and that TTL ordering (1h-before-5m) is preserved. Nice, tightly-scoped fix.

PR #10685, opened 2026-08-18
fix(analytics): read auto routing usage from call logs

cryptiklemur:

The auto-routing analytics endpoint queried `usage_logs`, but the runtime does not create or write that table. the caught SQLite error made the endpoint return empty metrics.

`call_logs.requested_model` already records `auto` and `auto-fast`, along with the provider used for each call.

## what changed

- read auto-routing counts and provider totals from `call_logs`
- use `requested_model` to separate `auto` from `auto-fast`
- remove tests that created an application-owned schema that does not exist
- test against the real core schema with different requested and resolved models
- restore `DATA_DIR` after the database regression test

No migration is needed. the branch contains five changed files and does not carry the combo patch.

## how to test

```sh
npx cross-env DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=4096 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-concurrency=1 tests/unit/auto-routing-analytics-db.test.ts tests/unit/db-logs-cache-3500.test.ts
npx eslint src/app/api/analytics/auto-routing/route.ts src/lib/db/usageLogs.ts src/lib/localDb.ts tests/unit/auto-routing-analytics-db.test.ts tests/unit/db-logs-cache-3500.test.ts
```

Both commands pass after rebasing onto the current `release/v3.8.50` tip.

### DIFF
--- a/src/app/api/analytics/auto-routing/route.ts
+++ b/src/app/api/analytics/auto-routing/route.ts
@@ -16,7 +16,6 @@ export async function GET(request: Request) {
   const authError = await requireManagementAuth(request);
   if (authError) return authError;
   try {
-    // Query usage_logs for auto/ prefix requests
     const totalRequests = getAutoRoutingTotalCount();
 
     // Variant breakdown
--- a/src/lib/db/usageLogs.ts
+++ b/src/lib/db/usageLogs.ts
@@ -1,11 +1,6 @@
 /**
- * db/usageLogs.ts — Read-only aggregation queries over `usage_logs`
- * extracted from the /api/analytics/auto-routing route handler.
- *
- * Hard Rule #5: routes must not embed raw SQL — these queries live here so the
- * /api/analytics/auto-routing route can delegate.
- *
- * Sliced out of #3500 (usage_logs cluster, slice 4).
+ * Read-only auto-routing aggregations over `call_logs`.
+ * `requested_model` keeps the client auto/* id after routing resolves a target.
  */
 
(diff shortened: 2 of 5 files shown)


diegosouzapw:
fix(analytics): read auto routing usage from call logs — looks correct, thanks for the clear premise. Verified: `usage_logs` is never written by the runtime (only the analytics code, the fabricated test table, and docs reference it), so the endpoint silently returned empty metrics; reading `call_logs.requested_model` is the right source and no migration is needed (column + `idx_call_logs_requested_model` already exist). The async-await fix in `combo.ts` is a real improvement too: `getCachedProviderConnectionById` is async but was being called synchronously, so the per-connection token limit was never applied to quota reservation — the new combo-quota test proving `tokenLimit=5000` is applied is a good regression guard. All three touched test files pass locally. Two release-engine notes: (1) the PR is CONFLICTING vs release/v3.8.50 (single-file conflict in combo.ts) and (2) it depends on #10682 / #10693 — please merge the CI foundation (#10693) and the AI-Horde fix (#10682) first, then rebase this PR, before re-running the quality gates. Minor, optional: consider a separate PR for the combo.ts token-limit change next time so it stays reviewable under the analytics title, and restore `process.env.DATA_DIR` in `auto-routing-analytics-db.test.ts` like `combo-quota-token-limit.test.ts` does for cleaner test hygiene.

PR #10686, opened 2026-08-18
fix(combo): await quota token limit lookup

cryptiklemur:

`getCachedProviderConnectionById()` is asynchronous, but combo routing read the returned `Promise` like a connection. that dropped the provider token limit and reduced the effective limit to zero.

## what changed

- import the connection lookup from `src/lib/db/readCache.ts`
- await the connection before reading `token_limit`
- await token-limit resolution before applying the combo cap
- cover the public `handleComboChat()` path with a regression test
- add the required changelog fragment

The regression fixture resolves a 5,000-token provider limit. it returned 0 before this patch and 5,000 after it.

## how to test

```sh
npx cross-env DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=4096 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-concurrency=1 tests/unit/combo-quota-token-limit.test.ts
npx eslint open-sse/services/combo.ts tests/unit/combo-quota-token-limit.test.ts
```

The focused test and lint command pass after rebasing onto the current `release/v3.8.50` tip.

### DIFF
--- a/changelog.d/fixes/10686-combo-quota-token-limit-await.md
+++ b/changelog.d/fixes/10686-combo-quota-token-limit-await.md
@@ -1,0 +1,1 @@
+- **Combo routing:** await each connection's token limit before reserving quota. The old lookup treated the `Promise` as a connection and dropped `rateLimitOverrides.tpm` ([#10686](https://github.com/diegosouzapw/OmniRoute/pull/10686)).
--- a/open-sse/services/combo.ts
+++ b/open-sse/services/combo.ts
@@ -87,7 +87,7 @@ import { selectQuotaShareTarget } from "./combo/quotaShareStrategy.ts";
 import { makeConnectionConcurrencyResolver, lookupPositiveCap } from "./combo/concurrencyCaps.ts";
 import { acquireQuotaShareConcurrencySlot } from "./combo/quotaShareConcurrency.ts";
 import { canAffordRequest } from "../../src/lib/quota/quotaScheduler.ts";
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Approved — awaiting quota token limit lookup + regression test.

Verified against the release/v3.8.50 base: `getCachedProviderConnectionById` (src/lib/db/readCache.ts) is async and returns a Promise, but `resolveTargetTokenLimit` in open-sse/services/combo.ts read it without `await`, so the token limit was always undefined (effective quota zero). This patch correctly awaits the lookup, moves the import off the localDb barrel to the specific db/readCache module (matching repo convention), makes `resolveTargetTokenLimit` async, and updates the sole call site in `handleRoundRobinCombo`. The new unit test (combo-quota-token-limit) is a genuine regression guard: it asserts a 5000-tpm limit survives round-robin reservation via the public handleComboChat path.

Minor, non-blocking: the patch also removes one unrelated stray blank line in handleComboChat — harmless cleanup.

Suggest running before merge (CI will cover it): `npx vitest run tests/unit/combo-quota-token-limit.test.ts` (fails on base, passes on head) and prettier/eslint on the two touched files. Note: the release/v3.8.50 base tip carries 25 inherited ESLint errors from #10692/#10695 (issue #9985) — only NEW errors should block.

Issue #8443, opened 2026-07-24
feat(providers): GUI toggle + configurable intervals for health-check schedulers


RCrushMe:
## Summary

Currently, two background schedulers in OmniRoute periodically fire `POST /api/providers/test` against **every** configured connection (each logged as `model: "connection-test"` in `call_logs`):

1. **Token Health Check Scheduler** (`src/lib/tokenHealthCheck.ts`) — sweeps every **60 seconds** and refreshes OAuth tokens that are about to expire.
2. **Credential Health Check Scheduler** (`src/lib/credentialHealth/scheduler.ts`) — sweeps every **5 minutes** and re-validates **all** connections (apikey + oauth).

Both schedulers can **only** be disabled via environment variables (`OMNIROUTE_DISABLE_TOKEN_HEALTHCHECK`, `OMNIROUTE_DISABLE_CREDENTIAL_HEALTH_CHECK`). There is no GUI control, no per-connection opt-out, and no way to tune the sweep interval from the dashboard. For users with dozens of connections or providers that rate-limit aggressively, this creates significant unnecessary traffic and quota consumption.

## Real-World Scenario

On a deployment with ~20 configured providers (NVIDIA, OpenAI, Anthropic, Gemini, DeepSeek, and multiple OpenAI-compatible endpoints), every 5 minutes the credential health scheduler fires a `connection-test` request to **each** connection. In a 24-hour window this produces:

- ~5,760 test requests just from credential health checks
- ~28,800 test requests from token health checks (every 60s × 20 connections)

These requests hit real upstream endpoints. Providers like NVIDIA return `401` ("Endpoint /models unavailable. Provide a Model ID to validate via /chat/completions."), while others like MiMoCode return `401` ("Provider validation endpoint not supported"). None of these responses are actionable — the connections were already working — but they still consume API quota and appear in call logs as errors, making it harder to spot real issues.

This behavior is particularly costly for:
- Providers with strict rate limits (e.g., free-tier models, low-quota API keys)
- Self-hosted / local providers (Ollama, vLLM, LM Studio) where the test adds latency but no value
- Deployments with many compatible nodes where the cumulative request volume is significant

### Token Health Check (`src/lib/tokenHealthCheck.ts`)

- **Tick interval**: `TICK_MS = 60 * 1000` (hard-coded 60s, line 33)
- **Disable switch**: `OMNIROUTE_DISABLE_TOKEN_HEALTHCHECK` env var (line 183)
[...]

PR #10687, opened 2026-08-18
feat(credential-health): per-connection sweep interval via healthCheckInterval (#8443)

maxmad64bis:
## Problem

The credential health scheduler tests every active provider connection on a single global cadence (`CREDENTIAL_HEALTH_CHECK_INTERVAL`, default 5 min). There is no way to pace heavy or low-value connections individually, and no per-connection opt-out. Issue #8443 reports the probe volume this generates (~6 000 probe requests/day for ~20 providers, as corrected in the thread).

## Fix

- The sweep reads `provider_connections.healthCheckInterval` (minutes) per connection: `> 0` overrides the global cadence, `<= 0` excludes the connection from the sweep (opt-out), absent falls back to the env default. The column already exists (the token-refresh scheduler uses it) — no migration, no API or UI change (the field is already editable per connection).
- On success the scheduler keeps the per-connection timing (`nextAttemptAt = lastAttemptAt + interval`) instead of clearing it, so a healthy connection is retested on its own schedule, not every sweep.
- Failure backoff (5 min → 10 → 30 → 2 h) is unchanged and takes priority.
- Removed dead code in the same file: `OAUTH_INTERVAL_MULTIPLIER` (never used — the inert multiplier called out in the issue thread; the header comment claiming OAuth is tested 2x less often was false), `getMaxFailuresAcrossConnections` (never called), and a leftover `interval` variable in `sweep()`.

Note on the shared column: a value set for token-refresh purposes also paces credential tests (and `0` disables both) — one knob, two schedulers. A per-connection interval shorter than the global sweep cadence is capped by it (the sweep timer itself is unchanged).

⚠️ base-red inherited: #9985

## Verification

The sweep predicates are replicated in the unit tests, following the existing house pattern for this file (the scheduler auto-initializes on import); the replica and the implementation were aligned line by line.

```
$ node --import tsx/esm --test tests/unit/credential-health-interval.test.ts

# tests 6
# pass 6
# fail 0

$ node --import tsx/esm --test tests/unit/credential-health-backoff-retry.test.ts

# tests 6
# pass 6
# fail 0
```

The full credential-health family (5 files: active-connections-9180, backoff-retry, boot-wiring, interval, search-providers) passes, 0 fail; the three pre-existing files are unchanged and pass as at base.

```
$ npx eslint tests/unit/credential-health-interval.test.ts tests/unit/credential-health-backoff-retry.test.ts src/lib/credentialHealth/scheduler.ts

/home/.../src/lib/credentialHealth/scheduler.ts
  21:1  error  '@/lib/localDb' import is restricted ...  no-restricted-imports

✖ 1 problem (1 error, 0 warnings)
```

The full `npm run lint` is red on the base branch (25 errors in untouched test files, byte-identical to `upstream/release/v3.8.50`); the one error above (`no-restricted-imports` on the `@/lib/localDb` import) also exists at base. This branch introduces no new lint error.

Tests added/modified: `tests/unit/credential-health-interval.test.ts` (new), `tests/unit/credential-health-backoff-retry.test.ts` (adapted: success now spaces the next sweep by the per-connection interval instead of clearing the timing).


diegosouzapw:
Solid, focused feature: per-connection sweep cadence via the already-existing provider_connections.healthCheckInterval column, with `<=0` as an opt-out and absent falling back to the env default. Correctly reuses the existing column (no migration/API/UI change), removes genuinely dead code (OAUTH_INTERVAL_MULTIPLIER was never used; getMaxFailuresAcrossConnections never called), and ships real unit tests that pass (I verified the full credential-health family green: 6+6+8). Cleanup observations only, none blocking: (1) the `isOAuth` parameter on `testConnection` is now fully unused after removing OAUTH_INTERVAL_MULTIPLIER — worth dropping to avoid confusion; (2) the sweep predicates are replicated in the tests rather than imported (documented house pattern because the scheduler auto-inits on import) — keep the replica/implementation aligned as you have been; (3) note the shared-column coupling is real and worth an explicit callout in the release notes: a `healthCheckInterval` set for token refresh also paces credential health, and `0` disables both (the EditConnectionModal defaults to 60, so configured OAuth connections will now be credential-checked at 60min instead of ~10min — aligned with the probe-volume goal of #8443, but a deliberate behavior change). No new lint errors vs base; the single no-restricted-imports on the @/lib/localDb import predates this PR. ready to merge.

maxmad64bis:
Thanks for the review! Both optional cleanups are in (commit e2160f720): dropped the unused `isOAuth` param on `testConnection`, and added the release-notes callout about the shared `healthCheckInterval` column (60min default for OAuth connections, `0` disables both). Typecheck and tests are clean — only the known omniglyph base-red remains. Ready to merge whenever.

PR #10691, opened 2026-08-18
fix(sse): import localDb through its real .ts extension (#10674)

### LINKED ISSUE #10674: base-red: dev server cannot boot on release/v3.8.50 tip — instrumentation fails with "Cannot find module '../../src/lib/localDb.js'" after #10647's lockfile change
## Symptom

`npm run dev` on the current `release/v3.8.50` tip (`df90591415`) dies during boot, before serving anything:

```
[STARTUP] Fatal: instrumentation hook failed during boot: Cannot find module '../../src/lib/localDb.js'
[FATAL] Failed to start Next custom server: Error: An error occurred while loading instrumentation hook: Cannot find module '../../src/lib/localDb.js'
```

The import chain is `instrumentation → memory/skills injection → open-sse/handlers/chatCore → open-sse/services/combo.ts:90`, which imports `"../../src/lib/localDb.js"` (a .js-suffixed TS import that Turbopack used to resolve to `localDb.ts`). The compiled dev chunk now leaves it external/unresolved.

## Discrimination (code vs deps) — A/B proven

| Commit | node_modules | dev boot |
| --- | --- | --- |
| `72d761fb50` | old lockfile install (omniglyph 1.3.1 era) | ✅ booted (live smoke ran against it today) |
| `72d761fb50` (same code) | **fresh `npm ci` from the tip lockfile** | ❌ same MODULE_NOT_FOUND |
| `233ac40e9d` (#10633) | fresh tip lockfile | ❌ same error |
| `df90591415` (tip) | fresh tip lockfile | ❌ same error |

Same code boots with the old dependency tree and fails with the new one ⇒ the regression came from the **dependency changes in the lockfile** (landed with #10647, which added 114 / removed 18 / changed 191 packages), not from any code commit. `next` itself is still pinned+resolved at 16.3.1; the culprit is a transitive toolchain package that changed how the `.js`→`.ts` import in `combo.ts:90` is resolved/externalized in the dev (Turbopack) instrumentation path.

Also reproduced with a clean `.build/next/dev` cache and a fully hermetic `npm ci`, so it is not stale-cache or dirty-node_modules noise.

## Impact

- Any fresh `npm ci && npm run dev` of the tip fails — local dev onboarding is broken.
- CI is blind to it: no gate boots the dev server, so quality/release gates stay green.
- Production build path untested (may or may not share the resolution).

## Suggested fix directions

1. Change `open-sse/services/combo.ts:90` to import `localDb.ts` (extension consistent with the sibling imports in the same file, which all use `.ts` or no extension) — likely a one-line fix.
2. Or identify/pin the transitive package whose bump changed dev-time resolution (diff `package-lock.json` between `72d761fb50` and `04af8b1517`).

Found while closing the relay CLI effort's end-to-end validation (the live server for smoke tests could no longer boot after syncing to the tip).

diegosouzapw:
Closes #10674.

## Root cause

`open-sse/services/combo.ts:90` imported `"../../src/lib/localDb.js"` — a `.js` suffix on a
module that only ever existed as `.ts`. Turbopack resolved it by accident; the dependency-tree
change in #10647 stopped doing so, and the instrumentation hook then died at boot with
`MODULE_NOT_FOUND`.

Every sibling import in that same file already uses `.ts`, so this was a one-off slip, not a
convention.

## Impact this was causing

- `npm run dev` on the tip could not boot at all (the symptom #10674 reports).
- **`Build App` has been red for 60 consecutive runs** on `release/v3.8.50` — every run
  available through the API, none green. The failure cascades: once `combo.ts` fails to
  resolve, Next reports 29 `Module not found` errors, including Node builtins
  (`child_process`, `dns`, `fs`, `tls`, `net`, `module`) pulled into the Middleware/Instrumentation
  graphs. All 29 come from this single unresolved import.
- `Deploy to VPS` has been skipping as a consequence.

## Validation

Before/after on the tip (`df90591415`), fresh `npm ci`, `.build/next/dev` cache cleared:

| | result |
| --- | --- |
| before | `Fatal: instrumentation hook failed during boot: Cannot find module '../../src/lib/localDb.js'` |
| after | boots fully — migrations, schedulers, proxy-health all start; **0** `Module not found` |

`typecheck:core` clean, `eslint` clean on all three files.

## Second offender found by the guard

`src/lib/usage/usageLedger.ts:1` had the same `.js`-on-a-`.ts` pattern. It never broke because
it is an `import type`, erased before resolution — a latent copy of the same bug, fixed here.

## Guard

`tests/unit/no-js-extension-on-repo-imports-10674.test.ts` rejects **relative** `.js` specifiers
across `open-sse/` and `src/`. Package specifiers stay allowed on purpose: publishing ESM under
`.js` is correct for real packages (`@modelcontextprotocol/sdk/client/index.js` and 25 others
in-tree). Proven to discriminate — re-introducing the `.js` suffix makes it fail, naming the
exact file and line.

### DIFF
--- a/open-sse/services/combo.ts
+++ b/open-sse/services/combo.ts
@@ -87,7 +87,7 @@ import { selectQuotaShareTarget } from "./combo/quotaShareStrategy.ts";
 import { makeConnectionConcurrencyResolver, lookupPositiveCap } from "./combo/concurrencyCaps.ts";
 import { acquireQuotaShareConcurrencySlot } from "./combo/quotaShareConcurrency.ts";
 import { canAffordRequest } from "../../src/lib/quota/quotaScheduler.ts";
-import { getCachedProviderConnectionById } from "../../src/lib/localDb.js";
+import { getCachedProviderConnectionById } from "../../src/lib/localDb.ts";
 import { orderTargetsByEvalScores } from "./evalRouting.ts";
 
 /**
@@ -1190,7 +1190,6 @@ export async function handleComboChat({
             if (i > 0) fallbackCount++;
             return stopProtectedPriorityTarget(`Connection capacity reached for ${modelStr}`);
           }
-
         }
 
         // #9654 Wave 2: per-target lane-aware admission probe. With virtual
--- a/src/lib/usage/usageLedger.ts
+++ b/src/lib/usage/usageLedger.ts
@@ -1,4 +1,4 @@
-import type { ModelPricingRegistry } from "./modelPricingRegistry.js";
+import type { ModelPricingRegistry } from "./modelPricingRegistry.ts";
 
 export type UsageStatus = "success" | "failed" | "rate_limited" | "timeout" | "cancelled";
 
--- a/tests/unit/no-js-extension-on-repo-imports-10674.test.ts
+++ b/tests/unit/no-js-extension-on-repo-imports-10674.test.ts
@@ -1,0 +1,44 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+import { execFileSync } from "node:child_process";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+/**
+ * #10674: `open-sse/services/combo.ts` imported `"../../src/lib/localDb.js"` — a
+ * `.js` suffix on a file that only exists as `.ts`. Turbopack used to resolve it
+ * anyway; after the dependency-tree change in #10647 it stopped, and the
+ * instrumentation hook died at boot with MODULE_NOT_FOUND, taking down `npm run dev`
+ * and the production build (60 consecutive red `Build App` runs).
+ *
+ * A `.js` specifier is legitimate for npm packages that publish ESM that way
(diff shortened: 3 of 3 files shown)


diegosouzapw:
Merging with the inherited base-red documented. None of the 10 red checks come from this PR (diff = 2 import specifiers + 1 new guard test):

- `Docs Gates` — stale counts across README/AGENTS/llm.txt/4 SVGs (`153→154` migrations, `340→341` providers) after other merges landed.
- `Fast Quality Gates` — `open-sse/executors/index.ts TS2345`, `geminiWeb.ts` size, and 4 stryker `tap.testFiles` entries missing for `sse-auth-exclusive-leases`, `aihorde-optional-api-key` and `responses-passthrough-openai-compatible` — all other people's tests, reproduced locally on this branch.
- `Unit Tests fast-path` — Antigravity cloudcode envelope tests.
- `Fast Production Build` — the very defect this PR partially fixes; it needs #10695 on top to go green (see that PR's build evidence).

Tracked under #9985.

PR #10694, opened 2026-08-18
fix(probe): isolate probe-origin failures from all deactivation sites

### LINKED ISSUE #9817: fix(providers): model test-all activity deactivates healthy connections within minutes
## Summary

On 3.8.50, model test-all activity (and the fallback chain it triggers) **deactivated 6 healthy provider connections within ~2 minutes** of the server boot, gutting the provider pool and causing progressive latency ("snappy at boot, laggy within 15 minutes").

## Timeline (byteserver, Hyper-V, 3.8.50 / BUILD_SHA 83efcbf)

- `19:25` — service boot (13 connections active, healthy)
- `19:31–19:39` — **166 test-all batches in 12 min** (multiple concurrent `Starting batch test` at the same second), driven by a dashboard client (audit log `actor=dashboard`, `192.168.99.1`). Tests hit:
  - keyless-premium `opencode/*` models → 402 (the #8681 gate, working as intended)
  - `nvidia` models → 30s timeouts → `ProxyEgress status=error`, `499` on the fallback chain
- `19:35:37–19:37:36` — **6 connections flipped `is_active=0`**: github, cline, agnes, deepseek, tokenrouter, nvidia (audit log: `provider.credentials.updated` with `changedFields:["isActive"]` for each)
- After that: models from those providers disappear, every affected request starts an emergency-fallback chain (nvidia again 30s timeouts), latency climbs.

## Suspected mechanism

Test failures (402/499/timeout) feed the **emergency-fallback → `markAccountUnavailable`** path, and the per-provider health/model-test sweeps write `providerSpecificData` + flip `isActive` on OAuth connections (github/cline/agnes were NOT part of the fallback chain — they were hit by the credential-health/token-refresh activity that also ran in the same window).

Net effect: a read-only "test all models" flow can **deactivate production connections** — no user confirmation, no cooldown, ~2 minutes to gut the pool.

## Expected

- Test-all failures must not deactivate unrelated/healthy connections (test-only isolation).
- Connection deactivation should require a real request-path failure, not a probe failure, and should respect a cooldown/backoff.
- `healthCheckInterval`/test-all cadence defaults should not allow 14 batches/minute.

## Environment

- omniroute 3.8.50 (release line), Ubuntu 24 VM, Node 24.19, sqlite storage
- Config: `OMNIROUTE_HIDE_HEALTHCHECK_LOGS=true`, `PROVIDER_LIMITS_SYNC_INTERVAL_MINUTES=1440`, `OMNIROUTE_DISABLE_CREDENTIAL_HEALTH_CHECK=true`
- Recovery: `UPDATE provider_connections SET is_active=1 WHERE is_active=0` restored service; no restart needed.

Refs: #8681 (the 402 gate), #9606 (session-stop reports), #9765 (sentinel leak)

maxmad64bis:
## Summary

Running "Test all" on a provider connection (or any scheduled probe) could permanently deactivate that connection. A 403 classified as `FORBIDDEN` or `ACCOUNT_DEACTIVATED` took the same path as real traffic — `isActive=false`, terminal `testStatus`, auto-disable — even when the failure was synthetic. A single probe run could silently remove a healthy account from the pool until an operator re-enabled it manually.

Probe requests are now tagged with an AsyncLocalStorage context (`runAsProbe`/`isProbeContext`) and **every** site that could mutate routing state refuses probe-origin failures. The single decision point is `shouldIsolateProbeFailures()` (`src/shared/utils/probeOrigin.ts`): probe context + feature flag `PROBE_CAN_DISABLE` (fail-safe) + opt-in setting `probeCanDisable` (default OFF = isolation ON, key wired into `updateSettingsSchema` for `POST /api/settings`). A probe failure is recorded (`lastError`, `lastErrorType`, `errorCode`) for dashboard visibility but never changes routing state — no cooldown, no lockout, no terminal status, no breaker degradation, no quota-cache poisoning, no refresh-token rotation.

Guarded sites: `markAccountUnavailable` (record-only, deliberately no `backoffLevel`), auto-disable hook, chatCore (FORBIDDEN / ACCOUNT_DEACTIVATED / QUOTA_EXHAUSTED / GEO_BLOCKED / MODEL_NOT_FOUND / codex 429 rotation failover / codex quota-state persistence / key-health rotator / reactive 401/403 refresh), executor proactive refresh (`base.ts` + `gitlab.ts` `execute()`), chat.ts provider breaker + 5-min quota cache. Probe stream consumption (SSE) is wrapped in `runAsProbe` so an error frame inside a 200 stream stays in the probe context (incl. queued Bottleneck jobs).

## Related Issues

- Closes #9817

## Validation

- [x] Change type: routing
- [x] `npm run lint`

## Tests Added Or Updated

- `tests/unit/probe-origin.test.ts` — probe context propagation (incl. genuinely queued Bottleneck jobs)
- `tests/unit/probe-autodisable-isolation.test.ts` — auto-disable refuses probe-origin permanent failures, still works for real failures
- `tests/unit/probe-production-path.test.ts` — real request-path failure keeps full production behavior
- `tests/unit/probe-testall-isolation.test.ts` — test-all FORBIDDEN / ACCOUNT_DEACTIVATED / queued rate-limited failures keep the connection active; **RESTORE e2e**: opt-in `probeCanDisable=true` really deactivates (`is_active=0`, `test_status=banned`)
- `tests/unit/probe-policy.test.ts` — the three policy gates (context, flag, setting)
- `tests/unit/probe-gate-autodisable.test.ts` — GEO_BLOCKED (no 24h cooldown), QUOTA 402 (no terminal state), MODEL_NOT_FOUND (no lockModel), 401 reactive refresh skipped, **stale-token proactive refresh skipped** (github + gitlab executors), **codex 429 rotation failover skipped** (no persisted cooldown), breaker intact on upstream timeout

Before this fix, the isolation tests failed (the connection was really deactivated); after, they pass. The real-path tests still pass, confirming regular failures deactivate as before. 40 tests green (probe family + breaker suites).

## Coverage Notes

This change touches `src/` and `open-sse/`. The isolation tests exercise the full test-all flow against a mock upstream (fetch-level), asserting the connection row stays `is_active=1` after a probe failure; the production-path test asserts the reverse (real failure deactivates). Neighboring model-test and cooldown suites were rerun and stay green.

## Reviewer Notes

- The probe context is thread-of-execution scoped: `runSingleModelTest` wraps the scheduled function inside `withRateLimit`, so Bottleneck-queued jobs keep the context (pinned by a test that genuinely queues job 2 behind job 1).
- `markAccountUnavailable` records probe failures WITHOUT `backoffLevel`: writing one would trigger the selection-time auto-decay (`resetConnectionBackoff`) that wipes `lastError` back to NULL on the next attempt — silently destroying the record.
- The probe record stores the raw (unsliced) error text, unlike the real path's `slice(0,100)` truncation — visibility is the point of the record.
- Fail-safe: if the flag or settings lookup throws, isolation stays ON.
- Executors that route through `BaseExecutor.execute()` (codex/github/ghe-copilot/default) inherit the proactive-refresh guard; `GitlabExecutor.execute()` overrides it and got the mirror guard. `tokenHealthCheck` and circuit-breaker HALF_OPEN probes remain deliberately outside `runAsProbe` (documented in `probeOrigin.ts`).
- Docs: `docs/security/BAN_DETECTION.md` "Probe isolation (model test-all)" section.


diegosouzapw:
Review could not be completed in this pass — the analysis environment had no shell/web access, so your diff (the +424/-19 across 9 files) could not be read or run. The direction is coherent: probe-origin (background/test) failures landing in the connection-deactivation path (markAccountUnavailable / auto-disable) is a real gap consistent with the 3-layer resilience model. Before this can proceed from Draft to merge consideration, please make sure the PR (1) includes a failing-then-passing unit test that proves a probe-origin failure now deactivates the connection (Hard Rule #18 — production-code change without a test does not merge), (2) does not overwrite terminal states (banned/expired/credits_exhausted) or trip the anti-thundering-herd guard in markAccountUnavailable when a probe marks a connection, (3) does not misclassify intermittent probe failures (5xx/429) into permanent deactivation, and (4) any new settings default OFF. Happy to do a full review once the diff is reachable.

maxmad64bis:
Thanks for the review — the diff is reachable now, and all four points are covered:

1. **Failing-then-passing proof** — the three isolation tests failed on base (a probe really deactivated the connection) and pass on this head. The opt-in path is also tested end-to-end: with `probeCanDisable: true` a probe really lands the connection at `is_active=0 / banned`.

2. **Terminal states / herd guard** — a connection already in `banned/expired/credits_exhausted` short-circuits before the probe guard, so it can never be overwritten. The probe record only writes the error fields; it deliberately skips `backoffLevel`, because writing one triggers an auto-decay that wipes the record on the next attempt.

3. **No 5xx/429 misclassification** — 402 stays record-only for probes, the codex-429 test asserts no cooldown is persisted, and the breaker stays untouched on probe timeouts.

4. **Settings default OFF** — `probeCanDisable` is optional with no default (resolves to `false`), so isolation is ON out of the box; unknown flags resolve to `"false"` too.

40/40 tests green, lint clean. Happy to dig into anything once you've read the diff.

PR #10542, opened 2026-08-16
feat(providers): optional AI Horde API key and live image catalog

pageragatz:
## Summary

- AI Horde connections can store an optional registered `aihorde.net` API key. Requests still work anonymously (`0000000000`); a saved key is used for higher queue priority (kudos).
- Chat continues through `oai.aihorde.net`. Image models are discovered from Horde's live `GET /v2/status/models?type=image` catalog and only advertised while `count > 0`. Names stay the exact Horde strings.
- `POST /v1/images/generations` for `aihorde/<name>` or `horde/<name>` runs Horde's native async generate/check/status flow, returns `b64_json`, and cancels the job on timeout or disconnect.
- Check/Save no longer treat every key as valid. The OpenAI-compatible `/v1/models` probe returns 200 for any Bearer token on `oai.aihorde.net`; validation now uses `GET /v2/find_user`. An empty field is still the optional anonymous path.

## Related Issues

- Related to existing AI Horde no-auth / anonymous-key work (`anonymousApiKey`, #4019 pattern).

## Validation

- [x] Focused tests: `node --import tsx/esm --test tests/unit/aihorde-image-catalog.test.ts tests/unit/aihorde-image-generation.test.ts tests/unit/aihorde-optional-api-key.test.ts tests/unit/aihorde-key-validation.test.ts`
- [ ] `npm run lint` (CI)
- [x] Production-code changes include new automated tests in this PR
- [ ] SonarQube PR analysis is green or remaining issues documented

## Tests Added Or Updated

- `tests/unit/aihorde-image-catalog.test.ts`
- `tests/unit/aihorde-image-generation.test.ts`
- `tests/unit/aihorde-optional-api-key.test.ts`
- `tests/unit/aihorde-key-validation.test.ts`

## Coverage Notes

- Catalog parser/cache, OpenAI→Horde request mapping, generation dispatch/`apikey` header, optional stored-key credential selection, and `find_user` key validation are covered by the new unit tests.
- `imageRegistry.ts` stays at 999 LOC (under the 1000-line new-file cap) by sourcing live Horde models from `AI_HORDE_IMAGE_PROVIDER.models`.

## Reviewer Notes

- Target **`release/v3.8.50`**, not `main`.
- No-auth early-return in `getProviderCredentials` now prefers an active stored Horde key, then the synthetic `noauth` fallback, and still honors `#9057` `allowedConnections` (synthetic `noauth` is not returned when an allowlist is set).
[...]

PR #10695, opened 2026-08-18
fix(config): keep the SQLite driver out of the client bundle (#10692)

### LINKED ISSUE #10692: base-red: production build fails — 28 Node-builtin resolution errors from server-only code reaching the client bundle (dashboard/providers page → db/core)
## Symptom

`npm run build:release` on `release/v3.8.50` fails with **28 `Module not found`** errors, all of
them Node builtins: `fs` (14), `net` (5), `tls` (2), `module`, `fs/promises`, `dns`,
`child_process`. No artifact is produced.

The CI job `Build App` has been **red for 60 consecutive runs** on this branch — every run the
API returns, none green — and `Deploy to VPS` skips as a consequence.

## This is NOT #10674

#10674 (the `localDb.js` specifier, fixed in #10691) broke the **dev** boot. With that fix
applied, `npm run dev` boots clean with 0 module errors — but the production build still fails
with these 28. Two distinct defects behind one red job.

## Root cause: server-only code reaching the client bundle

The build reports the offenders under `[Client Component Browser]` / `[Client Component SSR]`.
Full chain, leaf last:

```
src/app/(dashboard)/dashboard/providers/page.tsx   ← "use client"
  └ providerPageUtils.ts
    └ src/lib/providers/serviceKindIndex.ts
      └ open-sse/config/mediaServiceKinds.ts
        └ open-sse/config/imageRegistry.ts
          └ open-sse/config/providers/registry/aihorde/imageModels.ts
            └ open-sse/services/aihordeImageCatalog.ts
              └ src/shared/network/safeOutboundFetch.ts
                └ open-sse/utils/proxyFetch.ts
                  └ src/shared/utils/featureFlags.ts
                    └ src/lib/db/featureFlags.ts
                      └ src/lib/db/core.ts
                        └ src/lib/db/adapters/driverFactory.ts
                          └ src/lib/db/adapters/sqljsAdapter.ts   ← fs / net / tls
```

A `"use client"` page transitively pulls in the SQLite driver.

## Why it surfaced now

The chain's last link was added by **#10542** (`feat(providers): optional AI Horde API key and
live image catalog`), which made `aihordeImageCatalog.ts` import `safeOutboundFetch`. Everything
above it is old — `serviceKindIndex` has been imported by that page since v3.8.30.

The bundler tolerated it until the dependency-tree change in #10647 (the same lockfile change
behind #10674). Evidence: a build of `cd091ab878` against the **pre-#10647 dependency tree**
succeeded and produced a working artifact; the same code against a fresh `npm ci` from the tip
lockfile fails. So the dependency bump only removed the tolerance — the layering violation is
the real defect and should be fixed at the import chain, not by pinning deps back.

## Suggested direction

Break the chain so the client page stops reaching `db/core`. The narrowest cut is between
`imageRegistry` / `aihordeImageCatalog` and `safeOutboundFetch` — the client only needs the
static catalog shape, not the fetching path. Alternatives: make `serviceKindIndex` import a
client-safe subset of `mediaServiceKinds`, or move the live-catalog fetch behind a server route.

## Impact

- No production artifact can be built from the branch tip at all.
- Any VPS/canary deploy is blocked (found while deploying #10632 to the .17 gateway).
- CI is red but was not being acted on.

diegosouzapw:
Closes #10692. Stacked on #10691 — that fix is required for the build to get far enough to hit
this one.

## Root cause

`IMAGE_PROVIDERS` is reachable from `"use client"` dashboard pages: they read its **keys** to
derive which providers support which media kind (`mediaServiceKinds.ts` → `serviceKindIndex.ts`).
The `aihorde` entry imported its live-catalog service directly, so the browser graph inherited:

```
aihordeImageCatalog → safeOutboundFetch → proxyFetch → featureFlags
  → db/featureFlags → db/core → driverFactory → sqljsAdapter   ← fs / net / tls
```

`serviceKindIndex.ts` already stated the invariant in a comment — *"Client-safe: mediaServiceKinds
only pulls in the pure-data media registries (no server-only deps)"* — but a comment cannot fail a
build, so #10542 broke it unnoticed.

## Why a dynamic `import()` is not the fix

Tried first, and it does **not** work: deferring the module still requires the bundler to make it
browser-loadable, so the same 28 errors came back. The dependency has to be inverted.

## The fix

New pure module `open-sse/config/dynamicImageModelSources.ts` (register / get / reset, no imports).

- The registry entry asks it for models — it no longer knows the service exists.
- `aihordeImageCatalog` registers itself at import time. Every server path that needs live models
  already imports it (`api/v1/models/catalog.ts`, which is also the only caller of
  `getAllImageModels()`; the image-generation handler; provider validation), so server behaviour
  is unchanged.
- With nothing registered the getter yields `[]` — exactly what the live catalog returned before
  its first poll, so the pre-poll client view is unchanged too.

## Validation

Full `npm run build:release` on this branch: **0 `Module not found`, artifact produced**
(`dist/BUILD_SHA` written). That is the first successful build of this branch in 60+ recorded
`Build App` runs.

- 29 existing AI Horde tests green (catalog, generation, key validation, optional key, fallback).
- `typecheck:core` clean, `eslint` clean.

## Guard

`tests/unit/client-bundle-no-server-only-10692.test.ts` walks the real static-import graph from
the client-safe entry points and fails if any of them reaches `db/core`, `driverFactory`,
`sqljsAdapter` or `proxyFetch` — printing the offending chain. It deliberately does not follow
`import()`, and it runs in ~30ms, so this class of regression is caught without waiting for an
hour-long build. Proven to discriminate: it reproduced the exact 6-hop chain before the fix.

### DIFF
--- a/open-sse/config/dynamicImageModelSources.ts
+++ b/open-sse/config/dynamicImageModelSources.ts
@@ -1,0 +1,49 @@
+/**
+ * Registry indirection for image providers whose model list is discovered at runtime (#10692).
+ *
+ * `IMAGE_PROVIDERS` is reachable from `"use client"` dashboard pages — they read its KEYS to
+ * decide which providers support which media kind (see `mediaServiceKinds.ts`). A provider entry
+ * that imports its live-catalog service directly therefore drags that service, and everything it
+ * imports, into the browser graph. For AI Horde that meant
+ * `aihordeImageCatalog → safeOutboundFetch → proxyFetch → featureFlags → db/core → sqljsAdapter`,
+ * so the build tried to bundle `fs`/`net`/`tls` for the browser and failed.
+ *
+ * A dynamic `import()` does not help: the bundler still has to make the module browser-loadable.
+ * The dependency has to be inverted instead — the registry entry knows only this pure module, and
+ * the server-only service registers itself when it is imported (which every server path that needs
+ * live models already does).
+ *
+ * With no source registered the getter yields `[]`, which is exactly what the live catalog
+ * returned before it had polled — so the client keeps seeing the provider without its models,
+ * unchanged.
+ */
+
+export interface DynamicImageModelEntry {
+  id: string;
+  name: string;
+  inputModalities: string[];
+}
+
+type DynamicImageModelSource = () => DynamicImageModelEntry[];
+
+const sources = new Map<string, DynamicImageModelSource>();
+
+/** Called by a server-only catalog service at import time. Last registration wins. */
+export function registerDynamicImageModelSource(
+  providerId: string,
+  source: DynamicImageModelSource
+): void {
+  sources.set(providerId, source);
+}
+
+/** Models discovered for `providerId`, or `[]` when no server-side source is loaded. */
+export function getDynamicImageModels(providerId: string): DynamicImageModelEntry[] {
+  const source = sources.get(providerId);
+  if (!source) return [];
+  return source();
+}
+
+/** Test seam — drops every registration. */
+export function resetDynamicImageModelSources(): void {
+  sources.clear();
+}
--- a/open-sse/config/providers/registry/aihorde/imageModels.ts
+++ b/open-sse/config/providers/registry/aihorde/imageModels.ts
@@ -5,8 +5,14 @@
  * async API (`/v2/generate/async`). `models` is a live getter so
  * imageRegistry stays under the file-size cap and zero-worker names are
  * never advertised.
+ *
+ * The live models arrive through `dynamicImageModelSources` rather than a direct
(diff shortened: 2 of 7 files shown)


diegosouzapw:
Merging on top of #10691 (now merged). The remaining red checks are the same inherited set documented there (#9985) — stale doc counts, `executors/index.ts TS2345`, Antigravity cloudcode tests, and stryker `tap.testFiles` entries for other people's tests.

The check this PR is about is proven green locally in a way CI could not show before: a full `npm run build:release` on this branch produced **0 `Module not found`** and wrote `dist/BUILD_SHA` — the first artifact this branch has been able to build in 60+ recorded `Build App` runs.

Issue #10429, opened 2026-08-15
feat(ops): reproducible canary deploy script with provenance check, smoke and rollback


diegosouzapw:
## Problem

Deploying the internal gateway (192.168.0.17) is a manual sequence — build locally, `npm pack`, copy, `npm i -g`, `pm2 restart` — with no record of what was installed and no verification that the new build actually serves traffic. The 2026-08-14 outage (see #10427) went unnoticed until a human hit it, and rolling back meant reconstructing the previous state by hand.

Environment constraints found while diagnosing:
- The target box has 3 GB RAM / 2 cores — it cannot run `next build`. The artifact must be built elsewhere and shipped.
- The runtime is PM2 (`omniroute`, id 0), so restart/rollback is scriptable.

## Proposal

`scripts/ops/deploy-canary.sh` (or a skill wrapping it) that:

1. Refuses to proceed unless the artifact passes the provenance check from #10427 (or an explicit canary flag is set).
2. Packs the artifact, ships it, installs it, and records the previous version for rollback.
3. Restarts via PM2 and runs a real smoke test: `/api/monitoring/health` **plus** an actual `/v1/chat/completions` call against a cheap model — enough to catch an egress-path regression like the `Executor result must contain a Response` one, which a health check alone cannot see.
4. Rolls back automatically if the smoke fails, and prints the deployed `buildSha` on success.

## Acceptance

- One command deploys and verifies; a failed smoke restores the previous install.
- The deployed SHA is recorded and queryable afterwards.

Issue #10427, opened 2026-08-15
fix(build): validate packaged BUILD_SHA provenance + expose buildSha on health


diegosouzapw:
## Problem

`scripts/build/write-build-sha.mjs` writes `dist/BUILD_SHA` into the package, but **nothing validates that the SHA belongs to the release line**. A tarball built from any feature branch installs and runs indistinguishably from a release build.

This is not hypothetical — it caused a production outage on the internal gateway (192.168.0.17) on 2026-08-14:

- The installed package carried `dist/BUILD_SHA = 178febc50f`, a commit on `fix/9603-qwen-token-plan-quota` that does **not** descend from #10373.
- It therefore shipped the nominal `instanceof Response` guard from #10256, which rejects every response created by the npm `undici` package's `fetch` (a different `Response` class than the global one).
- Every request died with `502 … Executor result must contain a Response`, across unrelated providers (`codex`, `qwen-cloud-token-plan`), because the defect sits in the shared egress path.
- Diagnosis took a full session, largely because there was no way to tell *what code was actually running* short of SSHing in and grepping the bundle.

## Proposal

1. **Provenance gate** in `scripts/build/validate-pack-artifact.ts`: fail (or loudly warn on non-CI) when `dist/BUILD_SHA` is not an ancestor of the release branch / published tag. Escape hatch via an explicit env var for deliberate canary builds, so the intent is recorded rather than silent.
2. **Expose `buildSha`** in `/api/monitoring/health` (next to the existing `version`), so what is deployed can be audited over HTTP instead of by SSH + grep of `.next` chunks.

## Acceptance

- A tarball built from a non-release commit fails the gate (or is explicitly marked canary).
- `GET /api/monitoring/health` returns the running `buildSha`.
- Tests cover both the ancestor-check and the health field.

Issue #6695, opened 2026-07-08
fix: i18n drift — pt-BR (and other locales) missing/untranslated keys after update (194 missing, ~1792 untranslated)


diegosouzapw:
## Summary
After a version update, the **pt-BR** UI shows broken/untranslated strings: new keys added to `en.json` were not propagated/translated into `pt-BR.json`. Reported in the community channel ("quebrou na att e não traduz as chaves no pt-BR.json", "quebra no .i18n-state.json").

## Measured drift (release/v3.8.47, `src/i18n/messages/`)
Flattening the catalogs:
- `en.json`: **8657** keys
- `pt-BR.json`: **8463** keys
- **194 keys missing** in pt-BR (present only in en) → render as raw keys / fall back to English
- **~1792 keys identical to the English value** (likely never translated)

Sample missing keys:
```
usage.confirmRedeemResetCredit
settings.resetUsagePeriod_all
providers.compatBlockedParamsHint
combos.disableSessionStickiness
proxyRegistry.poolDescription
media.suggestedModels
```

## Root cause (hypothesis)
The translation pipeline state tracker (`.i18n-state.json`, gitignored/generated) drifted, so `i18n:sync-ui` / `i18n:run` didn't pick up the newly-added `en.json` keys for pt-BR (and likely other locales). `check-translation-drift.mjs` currently reports `.i18n-state.json not found — run npm run i18n:run to bootstrap`.

## Remediation
This is a **translation-data sync**, not a logic fix:
1. `npm run i18n:sync-ui` — backfills the 194 missing keys into every locale with a `__MISSING__:<en>` sentinel (mirrors the key tree so nothing renders as a raw key).
2. `npm run i18n:sync-ui -- --translate-markers` — replaces the `__MISSING__:` placeholders with real translations via the OmniRoute translation backend (requires the translation env vars).
3. Consider whether the `i18n-ui-coverage` CI gate should hard-fail on missing keys to prevent recurrence, and whether `.i18n-state.json` bootstrap should be part of the release flow.

## Not TDD-able
The user-visible defect is missing/untranslated data strings; the fix is regenerating catalog data, so there is no red→green unit test for it. Tracking here for the i18n pipeline / contributor already working on translations.

Reported via the WhatsApp BR support channel.

diegosouzapw:
[...]

PR #10699, opened 2026-08-18
fix(ops): judge the canary install by the SHA on disk, not npm's exit code

diegosouzapw:
Refs #10429.

## Problem

`npm install -g <tarball>` on the .17 gateway writes the whole package and then fails renaming
the old tree into its staging directory:

```
npm error code ENOTEMPTY
npm error syscall rename
npm error path /usr/lib/node_modules/omniroute
npm error dest /usr/lib/node_modules/.omniroute-h797OOZa
```

Exit status 217 — but `dist/BUILD_SHA`, the package version and every dependency are the new
ones. The canary read the non-zero exit as "install failed", aborted **before the restart**, and
`execFileSync` threw away npm's stderr, so the log said only `Command failed: ssh … npm install`.

This stopped the deploy half-done **twice on 2026-08-18**, each time leaving the host with the
new files on disk under the old running process — the exact split state a deploy script exists
to prevent. Both times it took a manual SSH round-trip to discover the install had actually
succeeded.

## Why not just tolerate ENOTEMPTY

Because the exit code is untrustworthy in *both* directions. The 2026-08-14 outage (#10427) went
the other way: `npm install -g` exited **0** while shipping a package built from the wrong
branch, and everything downstream believed it.

So neither "non-zero = failed" nor "zero = fine" holds. `classifyInstallOutcome()` decides on the
**BUILD_SHA read back from the installed package**:

| exit | SHA on disk | verdict |
| --- | --- | --- |
| 217 | matches | `installed-with-cleanup-failure` → continue, warn |
| 0 | matches | `installed` |
| 217 | stale | `failed` |
| **0** | **wrong artifact** | **`failed`** |
| any | unreadable | `failed` (fails closed, same rule as the provenance gate) |

## Staging directory

npm reuses the **same** staging name, so the orphan blocks the next install with the identical
error — that is why it recurred. `orphanStagingDirFromStderr()` extracts the exact path and the
run warns with it.

It is deliberately **not** removed automatically: that is an `rm -rf` under `/usr/lib`, and a
deploy script should not decide that on its own.

## Validation

`tests/unit/canary-install-outcome-10429.test.ts` — 7 cases, written first and confirmed failing
(`does not provide an export named 'classifyInstallOutcome'`), including the real 2026-08-18
stderr verbatim and the inverse trap (zero exit + wrong artifact must still fail).

Sibling suites green: `deploy-canary-10429`, `build-sha-provenance-10427` (25 tests total).
`typecheck:core` clean. CLI exercised with `--dry-run`, which also caught a wiring bug before it
shipped: `plan` does not expose `buildSha`, so the expected SHA had to come from `readBuildSha()`
— left as-is it would have failed *every* install.

### DIFF
--- a/scripts/ops/deploy-canary.mjs
+++ b/scripts/ops/deploy-canary.mjs
@@ -28,11 +28,16 @@
  *   OMNIROUTE_SMOKE_API_KEY      sent as Authorization: Bearer when the gateway requires auth
  */
 
-import { execFileSync } from "node:child_process";
+import { execFileSync, spawnSync } from "node:child_process";
 import path from "node:path";
 import process from "node:process";
 
-import { buildRemoteSteps, evaluateSmoke, planCanaryDeploy } from "./deployCanary.ts";
+import {
+  buildRemoteSteps,
+  classifyInstallOutcome,
+  evaluateSmoke,
+  planCanaryDeploy,
+} from "./deployCanary.ts";
 import { makeGitAncestryProbe, readBuildSha } from "../build/buildProvenance.ts";
 
 function parseArgs(argv) {
@@ -59,6 +64,22 @@ function run(step) {
   console.log(`\n▶ ${step.name}: ${step.description}`);
   const [command, ...rest] = step.argv;
   return execFileSync(command, rest, { encoding: "utf8" }).trim();
+}
+
+/**
+ * Like `run`, but never throws: returns the exit code plus both streams. Used for the
+ * install, whose exit code does not decide the outcome (see classifyInstallOutcome) and
+ * whose stderr must reach the log — it used to be swallowed by execFileSync throwing.
+ */
+function runCapturing(step) {
+  console.log(`\n▶ ${step.name}: ${step.description}`);
+  const [command, ...rest] = step.argv;
+  const result = spawnSync(command, rest, { encoding: "utf8" });
+  return {
+    exitCode: result.status ?? 1,
+    stdout: (result.stdout || "").trim(),
+    stderr: (result.stderr || "").trim(),
+  };
 }
 
 async function probeHealth(baseUrl) {
@@ -110,8 +131,9 @@ if (args.models.length === 0) {
 }
 
 const repoRoot = process.cwd();
+const localBuildSha = readBuildSha(repoRoot);
 const plan = planCanaryDeploy({
-  buildSha: readBuildSha(repoRoot),
+  buildSha: localBuildSha,
   isAncestorOfRelease: makeGitAncestryProbe(
     process.env.OMNIROUTE_RELEASE_REF || "origin/main",
     repoRoot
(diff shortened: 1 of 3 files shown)


jonlwheat2-gif:
### Review: logic is sound — one wiring gap found and fixed

I applied the PR to the current `release/v3.8.50` tip (`d7368e243`) and validated it. The decision matrix in `classifyInstallOutcome()` covers all four exit/SHA quadrants plus fail-closed, and the 7 new unit tests pin the real 2026-08-18 stderr verbatim. One genuine gap in the shell wiring, fixed below.

### The gap — `scripts/ops/deploy-canary.mjs`, install section

The classifier's fail-closed path was unreachable from the shell: `run(verify)` was called as a bare argument to `classifyInstallOutcome`, so when the remote SHA read itself threw (unreachable host, or `dist/BUILD_SHA` missing on the host), `execFileSync` died before classification ran — and the captured npm stderr, the exact output this fix exists to surface, was dropped again.

**Before (PR head, lines ~172–179):**

```js
const installResult = runCapturing(install);
const outcome = classifyInstallOutcome({
  exitCode: installResult.exitCode,
  stderr: installResult.stderr,
  installedSha: run(verify),
  expectedSha: localBuildSha,
});
```

**After (lines 172–184):**

```js
const installResult = runCapturing(install);
let installedSha = null;
try {
  installedSha = run(verify);
} catch {
  // verify throws when the host is unreachable or the SHA file is unreadable; let
  // classifyInstallOutcome fail closed on the null instead of dying in execFileSync
  // and dropping npm's captured stderr again (the failure this script exists for).
}
const outcome = classifyInstallOutcome({
  exitCode: installResult.exitCode,
  stderr: installResult.stderr,
  installedSha,
  expectedSha: localBuildSha,
});
```

**Plus, line 199** — the post-restart attestation read was `const installedSha = ...`, which now collides with the outer `let`; it becomes a plain reassignment:

```diff
-  const installedSha = run(verify);
+  installedSha = run(verify);
```

This makes the "unreadable SHA → `failed: no BUILD_SHA could be read…`" branch actually reachable, with npm's stderr printed right before the failure — consistent with the classifier's design (`installedSha` is already typed `string | null | undefined`).

### Validation (Windows, Node 24, worktree off the current release tip)

- 7 new tests pass; sibling suites `deploy-canary-10429` + `build-sha-provenance-10427` + `npm-publish-artifact-provenance` pass (25 total in the batch)
- `node --check scripts/ops/deploy-canary.mjs` clean; ESLint clean
- `typecheck:core`: no errors in touched files

### On the branch

`fix/10429-canary-install-exit-code` (on `jonlwheat2-gif/OmniRoute`, base `d7368e243`):
- `0c648b74f` — the PR change (3 files, 225+/4−)
- `dde829780` — the wiring fix above (1 file, 10+/2−)

diegosouzapw:
The 7 red checks are inherited base-red, not this PR — proven, not assumed.

Both this PR and # fail on the **exact same 7 jobs** despite having no file in common. Reproduced on the pure base tip `d7368e243d` in a disposable worktree with no commit from any branch of mine: `tests 26 / pass 10 / fail 16`.

The failures are in areas this diff does not touch: Antigravity cloudcode envelope + public models, GLM provider import (4 tests), free-model catalog qwen-web ids, i18n key drift (#6695), `/api/dead`, `open-sse/executors/index.ts TS2345`, `geminiWeb.ts` file size, and a Qwen Code JSON config parse.

Tracked under #9985.

Issue #10627, opened 2026-08-17
fix(startup): v3.8.49 (and v3.8.48) HTTP requests hang indefinitely on Windows — not the #6800/#8654 warmup window


ychaitu2025-dot:
## Summary

Reopening the symptom from #8654 with evidence that rules out the warmup-compile explanation given there. Reproduced on **two consecutive versions** (v3.8.49 and v3.8.48) on native Windows.

## Environment

- OS: Windows 11 Home 10.0.26200
- Node: v24.18.0
- Install: `npm install -g omniroute` (global)
- omniroute versions tested: 3.8.49 (latest on npm) and 3.8.48

## Repro steps

1. `npm install -g omniroute`
2. `omniroute serve --log`
3. Server logs proceed cleanly through full startup: Next.js binds and reports `✓ Ready in 0ms` on `localhost:20128`, then every subsystem initializer logs success in order (DB migrations, schedulers, credential health, batch processor, audit log, model catalog cache warm) — log ends at `[STARTUP] Model catalog cache warmed` with no errors.
4. `curl -i --max-time 120 http://127.0.0.1:20128/api/monitoring/health` (and plain `curl -I /`) — times out with **0 bytes received**, no HTTP response of any kind (not even a slow one).

## Why this isn't the #6800/#8654 warmup window

Per the #8654 resolution, the expected cause is Next.js being CPU-bound compiling routes for 30-71s on Windows cold start, and the fix direction was "wait longer, the fast-reject/hanging classification already exists, just be patient."

I retested with that guidance in mind, going well beyond it:

- Polled the health endpoint every 5s for **205 seconds** after the startup banner — no response, no additional log lines (not even a request being received/logged).
- Directly measured the server process's CPU time via `Get-Process -Id <pid>` before and after a 6-second wait: **CPU went from 8.046875s to 8.109375s** — a delta of ~0.06s of CPU over 6s of wall time. This is an essentially idle process, not one actively compiling. If this were the documented compile-bound warmup, CPU usage would be sustained/high during that window.
- `netstat -ano` confirms the port is `LISTENING` the entire time; the TCP handshake completes (curl times out with 0 bytes, not "connection refused"), so the listener is bound but the request never reaches / is never dispatched by the Next.js request pipeline — and the process is not doing any visible work while this happens.

[...]

PR #10709, opened 2026-08-19
fix(db): prevent Windows native-driver hang from stalling all requests (#10627)

jonlwheat2-gif:
## Summary

Fixes **#10627** — a **Windows-only total request hang** where every request stalls indefinitely (0 bytes returned, no logs, ~0% CPU) while the port still accepts connections. Reported on v3.8.48/49; the reporter's `omniroute doctor` flagged better-sqlite3 as broken on their global install.

The fix makes a broken native driver **fail fast and fail over** instead of deadlocking the process, and moves the native driver load **off the first-request path** in the proxy runtime.

### Symptom (from #10627)

- Windows (native), v3.8.48 + v3.8.49
- **Every** request hangs — `/`, `/api/*`, `/v1/*` — 0 bytes, no request logs
- Port listening, TCP connections accepted, ~0% CPU
- `omniroute doctor` reports better-sqlite3 broken/mismatched on the global install

### Root cause

OmniRoute's SQLite driver cascade (`src/lib/db/adapters/driverFactory.ts`) is a `try/catch` chain: `bun:sqlite → better-sqlite3 → node:sqlite → sql.js`. Each branch assumes a broken native addon **throws** on load (e.g. `ERR_DLOPEN_FAILED`, `Module did not self-register`), so the fallback runs.

That assumption holds on POSIX but **not on Windows**: a mismatched-ABI addon can **hang inside `DllMain` (the OS loader lock)** instead of throwing. A hang never reaches the `catch`, so:

1. The cascade never falls through to node:sqlite / sql.js
2. The first request that touches the DB — the proxy runtime's **cold** `getCachedSettings()` → dynamic `import("@/lib/db/settings")` → native addon load — stalls forever
3. Because it's a loader-lock deadlock, the whole process is stuck at ~0% CPU with the event loop idle

Two compounding factors:

- **The proxy runs in its own Next.js runtime** that never executes `instrumentation-node.ts`'s startup warm-ups, so its first DB touch happened **on the request path**, cold, with no timeout.
- **`runtimeRequire("better-sqlite3")`** (webpack external) resolves the addon in-process, so a hang is in-process and unstoppable — no timeout can interrupt a synchronous `require` once DllMain is entered.

### 1. Windows driver-hang probe — `src/lib/db/adapters/driverFactory.ts`

New `createBetterSqliteProbe()` gates the better-sqlite3 branch:

- Loads better-sqlite3 in a **child process** with a bounded **5s timeout**
- A `DllMain` hang → child killed by timeout (`status === null`) → verdict **"bad"**
- Verdict **cached per process** — the child spawn happens at most once
- **Windows-only**: on POSIX the probe returns `true` without spawning — broken addons throw there (already handled by the cascade), and we don't pay a subprocess on Linux/CI boot
- Injectable via the existing `createSyncDriverFactory(load, probe?)` seam — tests can simulate both worlds without a real child

Result: same machine, same broken binary — previously a deadlock, now an instant failover to `node:sqlite` (built into Node 22.5+, no addon) → sql.js.

### 2. Proxy boot warm — `src/proxy.ts`

Warm the settings cache at proxy-runtime boot (fire-and-forget, mirrors the `void warmModelCatalogCache()` pattern):

- The native driver load happens at **startup**, never on the first request
- A driver failure now surfaces as a **logged startup error**, and requests fall back to default limits — never an indefinite hang
- First real request starts with a hot cache (also removes the first-request latency spike on healthy installs)

## Files changed

| File | Change |
|---|---|
| `src/lib/db/adapters/driverFactory.ts` | +79/−3 — probe type, `createBetterSqliteProbe()`, cascade gating, production wiring |
| `src/proxy.ts` | +19 — boot-time settings warm |
| `tests/unit/db-adapters/driverFactory.test.ts` | +118 — 7 new tests |

## Tests

7 new unit tests:

- Probe: non-Windows skips the child and reports ok (no spawn)
- Probe: successful child probe is cached (spawned at most once)
- Probe: non-zero child exit rejects better-sqlite3
- Probe: child spawn throw rejects better-sqlite3
- Probe: timed-out child (`status === null`) rejects better-sqlite3 — the #10627 hang case
- Cascade: rejected probe skips better-sqlite3, falls through to node:sqlite, queries work
- Cascade: passed probe still prefers better-sqlite3 (healthy path unchanged)

### Verification performed

| Check | Result |
|---|---|
| New tests (7) | ✅ all pass |
| db-adapters suite (54 tests) | ✅ 54 pass / 1 fail — the fail is a **pre-existing** Windows libuv `UV_HANDLE_CLOSING` teardown crash, reproduced identically on the unmodified base |
| authz suite (227) | 225 pass — 2 fail on **pre-existing** `EPERM` on `%TEMP%` (files don't import the changed modules) |
| Typecheck (`tsconfig.typecheck-core`) | ✅ 0 errors in changed files (only the known pre-existing `omniglyph` node_modules artifact) |
| ESLint (3 changed files) | ✅ clean |
| E2E on Windows (Node 24, production wiring, real `spawnSync`) | ✅ see matrix below |

**E2E matrix (real Windows machine):**

| World | Driver picked | Time |
|---|---|---|
| Healthy (probe passes) | better-sqlite3 (unchanged) | 80ms probe, once |
| Broken addon (probe "bad") | **node:sqlite** | **1ms — no hang** |

## Platform behavior

| | Before | After |
|---|---|---|
| Windows + healthy addon | better-sqlite3 | better-sqlite3 (unchanged; +80ms probe once at first DB open) |
| Windows + broken addon | **indefinite hang, all requests dead** | node:sqlite fallback, requests served |
| Linux/macOS + broken addon | throws → node:sqlite/sql.js fallback | identical (probe is a no-op) |
| Linux/macOS + healthy addon | better-sqlite3 | better-sqlite3 (identical, no subprocess) |

## Affected users

- **v3.8.48/49 on Windows with a broken better-sqlite3 binary:** app now works out of the box — no reinstall needed — because the failover reaches `node:sqlite` (built into Node 22.5+)
- Users wanting the faster native driver back can repair it (`npm rebuild better-sqlite3`)

## Related

- Issue: #10627
- Not related to #10692/#10699/#10700/#10701 (those landed independently on the same release line)

### DIFF
--- a/src/lib/db/adapters/driverFactory.ts
+++ b/src/lib/db/adapters/driverFactory.ts
@@ -1,5 +1,6 @@
 import { runtimeRequire as _require } from "./runtimeRequire";
 import { existsSync } from "node:fs";
+import { spawnSync } from "node:child_process";
 import { createBetterSqliteAdapter } from "./betterSqliteAdapter";
 import { createBunSqliteAdapter, type BunSqliteDatabaseLike } from "./bunSqliteAdapter";
 import {
@@ -10,6 +11,70 @@ import type { SqliteAdapter } from "./types";
 
 
 type DriverLoader = (moduleName: string) => unknown;
+
+type SpawnSyncLike = (
+  command: string,
+  args: string[],
+  options: { timeout: number; stdio: "ignore"; cwd: string; windowsHide: boolean }
+) => { status: number | null };
+
+/** Returns whether better-sqlite3 may be loaded in this process. */
+export type DriverProbe = () => boolean;
+
+/**
+ * #10627 — Windows driver-hang guard.
+ *
+ * The sync cascade's try/catch only covers drivers that THROW on load
+ * (ERR_DLOPEN_FAILED, "Module did not self-register", ...). On Windows, a
+ * mismatched-ABI native addon can HANG inside DllMain (loader lock) instead of
+ * throwing — a hang never reaches the catch, so the fallback to node:sqlite /
+ * sql.js never runs and the first DB touch in a runtime stalls forever at ~0%
+ * CPU (the exact #10627 symptom: every request hangs, 0 bytes, no logs).
+ *
+ * The probe answers "can better-sqlite3 load AND open a database?" by loading
+ * it in a CHILD PROCESS with a bounded timeout, so a hang becomes a timed-out
+ * probe (verdict "bad") instead of a process-level deadlock. The verdict is
+ * cached per process — the child spawn happens at most once.
+ *
+ * On POSIX this is a no-op returning true: broken addons throw there, which
+ * the existing cascade already handles, and we don't want to pay a subprocess
+ * spawn on every Linux/CI boot.
+ */
+export function createBetterSqliteProbe(options: {
+  platform?: string;
+  execPath?: string;
+  spawn?: SpawnSyncLike;
+  timeoutMs?: number;
+}): DriverProbe {
+  const {
+    platform = process.platform,
+    execPath = process.execPath,
+    spawn = spawnSync as unknown as SpawnSyncLike,
+    timeoutMs = 5_000,
+  } = options;
+
+  let verdict: boolean | null = null;
+  return () => {
+    if (verdict !== null) return verdict;
+    if (platform !== "win32") {
+      verdict = true;
+      return verdict;
+    }
+    try {
+      const result = spawn(execPath, ["-e", "require('better-sqlite3')(':memory:')"], {
+        timeout: timeoutMs,
+        stdio: "ignore",
+        cwd: process.cwd(),
+        windowsHide: true,
+      });
+      // status === null means the child was killed by the timeout — a hang.
+      verdict = result.status === 0;
+    } catch {
+      verdict = false;
+    }
+    return verdict;
+  };
+}
 
 /**
  * The production loader for the sync driver cascade.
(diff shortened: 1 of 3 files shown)


jonlwheat2-gif:
## Confirmed: this is still live on the `v3.8.50` line (the base of this PR)

Verified against the current `release/v3.8.50` tip (`b754e44e2`, `package.json` version `3.8.50`):

| Check | Result |
|---|---|
| Commits since `v3.8.49` tag | #10691, #10692/#10695, #10698, #10697, #10699, #10700 — **none touch the driver factory** |
| `createBetterSqliteProbe` / probe gating in shipped `driverFactory.ts` | **0 matches** |
| Any hang-guard machinery (child_process / spawnSync / AbortSignal.timeout) in the shipped load path | **0 matches** — still the unguarded in-process `_require("better-sqlite3")` in a plain `try/catch` |
| Shipped `proxy.ts` | **26 lines, no warm** — first request still triggers the cold native load on the request path |

The vulnerable shape on the 3.8.50 tip is exactly what #10627 describes:
- `driverFactory.ts` (~line 136): `const BetterSqlite = _require("better-sqlite3")` — in-process, synchronous, no timeout, no child-process boundary
- The `catch` fires only on **throws**; a Windows `DllMain` loader-lock hang never throws → no fallthrough → first DB touch stalls forever

Two precision notes:

1. **Tagging:** the release branch carries version `3.8.50` but there is **no `v3.8.50` tag yet** — tags stop at `v3.8.49`. The reporter's v3.8.48/49 are the last tagged builds; the 3.8.50 line is the current untagged tip, and the bug code is byte-identical on it.
2. **Not a regression:** the bug is structural (the throw-only cascade assumption) and predates 3.8.48/49 — the unguarded load exists since at least the May `v3.8.5` tag. It's Windows-only and machine-dependent (broken addon ABI), which is why Linux CI never caught it and it shipped.

This PR's two changes are the first to touch that path on this release line.

Issue #10681, opened 2026-08-18
feat(resilience): Correlated per target decision trace for priority combo fallbacks


stanleytejakusuma:
## Problem / Use Case

Priority combos can be impossible to audit after a mixed fallback. Dispatched attempts are correlated in `call_logs`, but candidates excluded before dispatch can leave no correlated decision record. That makes it impossible to reconstruct whether a configured target was evaluated and gated, reordered, or never reached after a later success.

This report is **not** evidence that configured priority was ignored, and it does not assert why any cooldown existed. It requests enough safe telemetry to distinguish those cases in a future incident.

### Observed example

An operator defined a five target `priority` combo in this configured order:

1. `claude/claude-fable-5-max`
2. `claude/claude-opus-5-max`
3. `codex/gpt-5.6-sol-max`
4. `codex/gpt-5.6-terra-max`
5. `deepseek/deepseek-v4-pro`

On OmniRoute `3.8.49`, a single invocation resolved five targets. Metadata only evidence showed:

- Fable was dispatched and returned `429`.
- Sol and Terra were explicitly logged as skipped because the Codex provider was in global cooldown.
- DeepSeek was dispatched and returned `200`.
- The correlated call log index retained Fable and DeepSeek attempts only. There was no correlated Opus attempt or durable skip decision.

The missing Opus record means the incident cannot establish whether Opus was gated before dispatch, reordered behind the successful target, or otherwise not reached. The cause of the Codex cooldown is also unproven. The current evidence establishes an observability gap, not a priority ordering defect.

The relevant implementation path has several early returns before the normal attempt event and call log write. These include circuit open, provider cooldown, request scoped exhaustion, model lockout, quota cutoff, availability, credential gate, and predictive TTFT. The priority pre screen is also not an authoritative permanent exclusion: availability is rechecked at turn time.

Related: #9630 improves the zero attempt terminal response and aggregate diagnostics. This request covers mixed invocations that have both real upstream attempts and pre dispatch exclusions, including successful fallbacks, where a terminal error diagnostic is unavailable or insufficient.

## Proposed Solution

Add a safe, ordered, invocation scoped decision trace for every combo target considered during a combo request.

1. Create one opaque parent invocation ID at combo entry. It must be unique per invocation and must not rely on `comboExecutionKey`. Existing `correlation_id` can be used only if it has that guarantee and is propagated to pre dispatch decisions.
[...]

PR #10730, opened 2026-08-19
feat(combo): opaque per-invocation decision trace for priority fallbacks (#10681)

stanleytejakusuma:
## Summary
Implements #10681: an opaque, per-invocation decision trace for combo fallbacks.

Priority combos can be impossible to audit after a mixed fallback. Dispatched attempts are persisted in `call_logs`, but targets excluded before dispatch leave no correlated decision record — you cannot tell whether a configured target was evaluated and gated, reordered, or never reached after a later success.

## What this adds
1. **`open-sse/services/combo/decisionTrace.ts`** — `createInvocationId()` (opaque `combo-<uuid>`), `startComboTrace`, `recordComboDecision`, `finalizeComboTrace` (marks never-iterated targets `not_reached`), `getComboTrace`. Bounded in-memory retention: 30min TTL, 2000 invocations (LRU).
2. **`combo.ts`** — hooks at every pre-dispatch skip site (circuit open, provider cooldown, request exhaustion, model lockout, quota cutoff, availability, credential gate, concurrency cap, admission lane, predictive TTFT) recording `skipped_before_dispatch` with an allowlisted reason; a single `dispatched` record per target (per-target retries dedupe); and finalize on success / global timeout / all-failed / all-skipped terminals.
3. **`GET /api/usage/combo-trace/[id]`** — management-auth read of the trace for a given invocation.

## Safety contract
The trace contains **routing metadata only**: invocation id, combo name, strategy, per-target provider/model, decision, allowlisted skip reason, timestamps, terminal status. **Never** prompts, request/response bodies, headers, credentials, account ids, or raw upstream error strings. Routing semantics are unchanged — this is a read-only projection of existing dispatch decisions.

## Tests
- Unit: invocation-id uniqueness, skip-reason allowlist enforcement, `not_reached` finalize.
- Integration (public `handleComboChat` + injected `handleSingleModel`): mixed 429 fallback yields `[dispatched, dispatched, not_reached]`; availability skip yields `skipped_before_dispatch/availability`.
- 5/5 pass. Combo suite: 1183/1185 (the 2 failures — `combo-runtime-unit-concurrency` tmpdir-symlink assertion and the zero-latency compression-opt-in 500 — are identical on clean `release/v3.8.50`, pre-existing). `typecheck:core` errors are the pre-existing OmniGlyph baseline.

## Notes
- Retention is in-memory by design (matches the issue's "normal bounded retention"); a follow-up could persist to call-log metadata if desired.
- Nested combo-ref dispatches get their own invocation id (per-dispatch traces); the top-level `handleComboChat` accepts an optional `invocationId` for callers that want to seed it.

### DIFF
--- a/open-sse/services/combo.ts
+++ b/open-sse/services/combo.ts
@@ -1059,12 +1059,12 @@ export async function handleComboChat({
         const cb = getCircuitBreaker(provider);
         if (cb.getStatus().state === "OPEN") {
           log.info("COMBO", `Skipping ${modelStr} — circuit breaker OPEN for ${provider}`);
-    recordComboDecision(traceInvocationId, {
-      step: target.executionKey,
-      target: modelStr,
-      decision: "skipped_before_dispatch",
-      reason: "circuit_open",
-    });
+          recordComboDecision(traceInvocationId, {
+            step: target.executionKey,
+            target: modelStr,
+            decision: "skipped_before_dispatch",
+            reason: "circuit_open",
+          });
           if (i > 0) fallbackCount++;
           return stopProtectedPriorityTarget(`Provider ${provider} circuit breaker is open`);
         }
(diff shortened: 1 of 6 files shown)


stanleytejakusuma:
Follow-up c469c26d9 (after review): stamps the trace id on the receipt so the store is actually reachable post-incident, per the #10681 auditability goal.

- `X-OmniRoute-Combo-Trace: combo-<uuid>` header on the final Response at every exit path (wrapper around handleComboChat; callers may still supply their own invocationId).
- One metadata-only log line at finalize (id, terminal status, decision count) so finalized traces survive restarts via the existing log pipeline.
- Eviction at the MAX_TRACES cap now prefers FINALIZED traces, pinning in-flight invocations against bursts.
- Tests: header on success + all-failed paths, log-line shape, eviction preference. Combo suite 162/162.

PR #7432, opened 2026-07-16
fix(server): wire credential health scheduler into serve boot

danscMax:
## Problem
`src/lib/credentialHealth/scheduler.ts` was never imported by the serve runtime — zero real imports, so its top-level auto-init never fired and `initCredentialHealthCheck()`'s boot sweep never ran. Provider connections whose web-session/key had recovered stayed red (`test_status=error`) until a manual re-test — the recurring *providers go red on startup* symptom.

## Fix
Wire `initCredentialHealthCheck()` into `src/server-init.ts` after the other schedulers, via a non-fatal dynamic import (matches the plugins/pricing/arena pattern in the same file; the scheduler statically pulls the provider test route and self-guards — disabled under test/build and via `OMNIROUTE_DISABLE_CREDENTIAL_HEALTH_CHECK`).

Safe by construction: on a successful probe `testSingleConnection` sets `test_status=active` and clears `rateLimitedUntil` (heals the badge); a failed probe only preserves the existing cooldown, so the sweep never amplifies red.

## Tests
- `tests/unit/credential-health-wiring.test.ts` — regression guard on the wiring (fails on the pre-fix file, passes after).
- `typecheck:core` ok · `check:cycles` ok · eslint clean.

PR #10780, opened 2026-08-19
chore(startup): remove server-init.ts, a module nothing imports

maxmad64bis:
⚠️ base-red inherited: #9985

## Summary

- `src/server-init.ts` is never imported. The Next.js instrumentation hook
  (`src/instrumentation-node.ts`) is the boot path that runs, and it already performs every
  initialization this module declares — checked call by call.
- The module is not merely inert, it misleads: wiring added to it never executes. That is what
  happened in #7432 (credential health) and #5312 (thinking budget), and both left regression tests
  asserting the wiring must not live there.
- Removes it, corrects the four comments naming it as the caller (`arenaEloSync.ts`,
  `openrouterProviderStats.ts`, `modelsDevSync.ts`, `pricingSync.ts` — the real caller is
  `instrumentation-node.ts` in all four), and the doc and `.env.example` references that pointed at
  the file.

## Related Issues

- Related to #7432 and #5312 — both were caused by wiring landing in this module.

## Validation

- [x] Change type: other (dead-code removal)
- [x] Focused tests and category gates from the golden path: the four suites that asserted on the
      module (82 pass, 2 skipped), `npm run test:vitest` (368/368), `npm run check:docs-all`
- [x] `npm run lint`

## Tests Added Or Updated

- `tests/integration/integration-wiring.test.ts`
- `tests/integration/security-hardening.test.ts`
- `tests/unit/lib/jobRegistry/boot-wiring.test.ts`
- `tests/unit/cleanup-column-fix.test.mjs`
- `tests/unit/thinking-budget-hydration-5312.test.ts` (comment only)

Assertions that only read the `server-init.ts` source are removed, or redirected to the live path
(`instrumentation-node.ts`, `initCloudSync.ts`), so they now cover the module that actually runs.

## Coverage Notes

- No production code is added: a file is deleted together with the assertions that only described
  it. The live boot path keeps its coverage and is now asserted directly instead of through the
  dead module.

## Reviewer Notes

- No behavior change — the module never executed.
- Localized copies under `docs/i18n/` still name the file. They are generated and catch up on the
  next translation run.
- Overlaps with #10775 on `src/server-init.ts` and two test files. Both are independently
  mergeable; whichever lands second needs a trivial rebase.
- `npm run check:docs-all` still reports two failures (`BOT_TOKEN` / `BOT_URL` absent from
  `.env.example`, executors count in `ARCHITECTURE.md`). Both are pre-existing and outside this
  diff — the diff touches one line in each of `.env.example` and `ENVIRONMENT.md`.


maxmad64bis:
Rebased onto `bc6129bcb` — the conflict is gone (`MERGEABLE`).

The conflict came from #10775, which landed first and touched the same module: it removed `enforceSecrets()` from `src/server-init.ts` and dropped the matching assertions. Resolution kept this PR's intent, which is a superset — the whole module goes:

- `src/server-init.ts` — deleted (modify/delete).
- `tests/integration/integration-wiring.test.ts` — the `Pipeline Wiring — server-init.ts` block is removed entirely rather than trimmed.
- `tests/integration/security-hardening.test.ts` — **dropped from the diff**: #10775 already applied those 6 lines. The PR is now 17 files instead of 18.

Re-ran after the rebase: the 7 affected test files pass (105 tests, 0 failing, 4 skipped), `npm run test:vitest` 368/368, `npm run lint` 0 violations.

**All 7 red checks on this run are inherited base-red, none of them from this diff:**

| Check                          | Cause                                                                                                                                                                                               |
| ------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Docs Gates (fast-path)         | 5 strict drifts, all the provider count `342` vs the live `343` (`PROVIDER_REFERENCE.md` + 4 SVGs) — files this PR does not touch. Named in the #9985 verdict for the `3810a6c52..bc6129bcb` range. |
| Fast Quality Gates             | same failure on unrelated PRs cut from the same base.                                                                                                                                               |
| Unit Tests fast-path 1/4 … 4/4 | same four shards fail on unrelated PRs cut from the same base.                                                                                                                                      |
| No new ESLint warnings         | the ratchet fails on `i18nUiCoverage.pct: 99.4 < baseline 100`. This diff contains no UI file and no locale file, so it cannot move that metric.                                                    |

⚠️ base-red inherited: #9985

PR #10784, opened 2026-08-19
Revert "fix(build): tolerate same-realpath symlink / stale-typed dest in assembleStandalone"

amirsource133:
Reverts diegosouzapw/OmniRoute#10776


diegosouzapw:
Thanks for opening this, but I can't find evidence that #10776 caused a regression — could you share the concrete failure (build log, CI run, or repro) that prompted this revert?

What I checked:
- Post-merge CI on the #10776 merge commit (3d7ed7aa) is fully green: Build App, Deploy to VPS, Release-Green (continuous), and Push on release/v3.8.50 all succeeded.
- I reproduced the original bug independently: reverting only `assembleStandalone.mjs` to its pre-#10776 state (keeping the new regression test) reliably throws the exact documented crash — `ERR_FS_CP_EINVAL: src and dest cannot be the same .../node_modules/wreq-js` — confirming the fix is both necessary and effective.
- This PR also deletes the regression test added in #10776, so merging it would silently reintroduce the crash with no test left to catch it.

Unless there's a concrete failure this fix caused that I'm missing, I'd recommend closing this revert rather than merging it. Happy to look again if you can point to the specific breakage.

diegosouzapw:
Closing — no response with a concrete regression report after the request above, and the investigation confirms #10776's fix is necessary (verified the exact crash reproduces without it) and this revert would also delete its regression test. Happy to reopen if you can share a specific failure this fix caused.

PR #10791, opened 2026-08-19
fix(translator): convert OpenAI tool-result images to Claude image blocks (#9692)

### LINKED ISSUE #9692: fix(kimi/translator): tool-result images break OpenAI→Claude routing and K3 protocol override is ineffective
## Summary

Kimi Code `k3` works for normal text requests through OmniRoute, but the first request after a tool returns an image fails with:

```text
400 Invalid request Error
```

The failure is reproducible with one small PNG returned by Kimi Code CLI's `ReadMediaFile` tool. Two related defects were observed:

1. During OpenAI → Claude request translation, OpenAI-style `image_url` blocks nested inside a tool result are forwarded unchanged into an Anthropic `tool_result`, producing an invalid Claude Messages payload.
2. Setting the K3 per-model protocol override to `targetFormat: "openai"` is persisted and shown in the UI on the affected deployment, but the actual request still resolves to `target_format: claude`, so the intended OpenAI-protocol workaround is ineffective.

The first defect is generic: any OpenAI-compatible client routed to a Claude-compatible upstream can hit it when media appears inside `role: "tool"`, rather than a normal user message.

## Environment

- OmniRoute deployment: `omniroute-local:3.8.49-pr9200`
- Current source rechecked: `release/v3.8.50` still contains the nested tool-result translation gap
- Client: Kimi Code CLI `0.33.0`
- Provider: `kimi-coding`
- Model: `k3`
- Route: OpenAI-compatible client → OmniRoute `/v1` → Claude target → Kimi Coding Claude-compatible endpoint

K3 is resolved as vision-capable (`image_in`) on the discovered-model path. A separately configured GLM 5.2 secondary model is not involved in the failing request.

## Minimal reproduction

1. Start a clean Kimi Code CLI session using `kimi-coding/k3` through OmniRoute.
2. Send a normal text request and confirm that K3 responds.
3. Call `ReadMediaFile` for one PNG (the reproduced file was about 86 KB).
4. Ask K3 to describe the returned image.
5. Observe that the first model request after the tool result fails with HTTP 400.

This reproduces without old history, compaction, swarm routing, multiple images, or the secondary GLM model.

## Observed request log

```text
call_log_id: 1786055841847-8692b4
correlation_id: 468fe44a-7290-439c-87cb-2cca26659928
provider: kimi-coding
model: k3
source_format: openai
target_format: claude
status: 400
tokens_in: 0
error: [400]: Invalid request Error
```

`tokens_in: 0` is consistent with the upstream rejecting the request structure before inference begins.

## Defect A: nested tool-result images are not converted to Claude blocks

Kimi Code CLI returns the media as an OpenAI-style content part. On the wire the relevant message is equivalent to:

```json
{
  "role": "tool",
  "tool_call_id": "tool_123",
  "content": [
    {
      "type": "image_url",
      "image_url": {
        "url": "data:image/png;base64,..."
      }
    }
  ]
}
```

For an image inside a normal user message, `open-sse/translator/request/openai-to-claude.ts` correctly converts the part to:

```json
{
  "type": "image",
  "source": {
    "type": "base64",
    "media_type": "image/png",
    "data": "..."
  }
}
```

However, the `msg.role === "tool"` branch in `getContentBlocksFromMessage()` only calls `stripEmptyTextBlocks(msg.content)` and then embeds that array directly in `tool_result.content`.

The same pass-through exists for a pre-shaped `tool_result` nested in a user message: its content is stripped for empty text, but media blocks are not normalized.

The resulting Claude payload therefore contains an invalid nested block:

```json
{
  "type": "tool_result",
  "tool_use_id": "tool_123",
  "content": [
    {
      "type": "image_url",
      "image_url": {
        "url": "data:image/png;base64,..."
      }
    }
  ]
}
```

The Claude-compatible upstream expects:

```json
{
  "type": "tool_result",
  "tool_use_id": "tool_123",
  "content": [
    {
      "type": "image",
      "source": {
        "type": "base64",
        "media_type": "image/png",
        "data": "..."
      }
    }
  ]
}
```

### Related prior work is the opposite direction

PR #5100 converts images inside Anthropic `tool_result` blocks when translating **Claude → OpenAI**. It does not cover this **OpenAI → Claude** path.

## Defect B: the K3 OpenAI target-format override is ineffective

The attempted workaround was to set the per-model override:

```json
{
  "targetFormat": "openai"
}
```

On the affected deployment the value is saved in the database and displayed in the UI, but the call log still reports:

```text
source_format: openai
target_format: claude
```

and the Claude-compatible endpoint remains selected.

The current source has the intended downstream wiring:

- `resolveChatCoreRequestSetup()` reads `modelInfo.targetFormat` as `customModelTargetFormat`.
- `resolveChatCoreTargetFormat()` resolves the wire format.
- `resolveExecutionCredentials()` writes it to `_omnirouteKimiTargetFormat`.
- `KimiExecutor` selects the OpenAI or Claude endpoint from that metadata.

Leaf tests cover the latter steps by injecting an already-resolved target format, but there is no end-to-end regression test proving that a saved K3 model override is hydrated into `modelInfo`, survives request setup, and reaches `KimiExecutor` for a real `kimi-coding/k3` route. The observed behavior indicates the value is lost before or during that resolution path.

## Expected behavior

1. OpenAI-style media parts nested inside tool results are recursively normalized before constructing a Claude `tool_result`.
2. Data-URI images become Anthropic base64 image sources.
3. Remote image URLs become Anthropic URL image sources where supported.
4. Text and valid existing Anthropic blocks are preserved.
5. Multiple images and mixed text/image results remain in order.
6. A saved K3 `targetFormat: "openai"` override actually selects the OpenAI request shape, URL, and headers.
7. If the override is unsupported, the UI/API should reject it instead of persisting a no-op value.

## Suggested fix

- Extract the existing user-message image normalization into a reusable content-block normalizer.
- Apply it recursively to both:
  - `msg.role === "tool"` content arrays, and
  - `part.type === "tool_result"` content arrays inside user messages.
- Keep `stripEmptyTextBlocks()` as a separate sanitation step.
- Add an integration-level Kimi target-format test that begins with persisted model metadata, rather than injecting `_omnirouteKimiTargetFormat` directly.

## Suggested regression tests

- OpenAI `role: "tool"` with one base64 `image_url` → Claude `tool_result` with one `image` block
- mixed text + image tool result preserves both blocks and order
- multiple tool calls returning multiple images
- remote `image_url` conversion
- existing Anthropic-style `image` blocks remain valid
- empty nested text blocks are still stripped
- `kimi-coding/k3` with a persisted OpenAI override resolves `target_format: openai` and uses the OpenAI endpoint
- Kimi K3 `ReadMediaFile` round trip no longer returns HTTP 400

Prudhvivuda:
## Summary
- Convert OpenAI `image_url` (and AI-SDK `image`) parts nested in `role: "tool"` / `tool_result` content into Claude Messages `image` source blocks during OpenAI→Claude translation.
- User-message vision parts already converted; tool results were forwarded raw, so the next Claude-format upstream turn 400'd before inference (Kimi Code CLI `ReadMediaFile` and any OpenAI-compatible client that returns media from a tool).
- Scope is defect A only (payload conversion). The K3 protocol-override sibling in #9692 is not changed here.

Closes #9692

⚠️ base-red inherited: #9985

## Test plan
- [x] `node --import tsx/esm --test tests/unit/openai-to-claude-tool-result-images-9692.test.ts`
- [x] `node --import tsx/esm --test tests/unit/openai-to-claude-strip-empty.test.ts`
- [x] `node --import tsx/esm --test tests/unit/openai-to-claude-file-attachments.test.ts`
- [x] `node --import tsx/esm --test tests/unit/claude-to-openai-image-toolresult.test.ts`
- [x] `node --import tsx/esm --test tests/unit/openai-to-claude-bare-tool.test.ts`
- [ ] Optional live: OpenAI-compatible client → OmniRoute → Claude-format Kimi/Anthropic with a tool that returns a PNG; next turn should be 200, not `400 Invalid request`

### DIFF
--- a/changelog.d/fixes/9692-openai-to-claude-tool-images.md
+++ b/changelog.d/fixes/9692-openai-to-claude-tool-images.md
@@ -1,0 +1,1 @@
+- **fix(translator):** convert OpenAI `image_url` blocks nested in `role: "tool"` / `tool_result` content to Claude `image` source blocks so OpenAI-compatible clients (Kimi Code CLI `ReadMediaFile`, and any other tool that returns media) no longer 400 the next Claude-format upstream turn ([#9692](https://github.com/diegosouzapw/OmniRoute/issues/9692))
--- a/open-sse/translator/request/openai-to-claude.ts
+++ b/open-sse/translator/request/openai-to-claude.ts
@@ -7,10 +7,17 @@ import { sanitizeToolId } from "../helpers/schemaCoercion.ts";
 import { safeParseJSON } from "../helpers/jsonUtil.ts";
 import { applyKimiCodingThinking } from "../helpers/claudeHelper.ts";
 import { DEFAULT_THINKING_CLAUDE_SIGNATURE } from "../../config/defaultThinkingSignature.ts";
-import { getDefaultThinkingBudget, isAdaptiveThinkingOnly } from "../../../src/shared/constants/modelSpecs.ts";
+import {
+  getDefaultThinkingBudget,
+  isAdaptiveThinkingOnly,
+} from "../../../src/shared/constants/modelSpecs.ts";
 import { fitThinkingToMaxTokens } from "./openai-to-claude/thinkingBudget.ts";
 import { enforceToolResultAdjacency } from "./openai-to-claude/toolResultAdjacency.ts";
 import { sanitizeToolResultId } from "./openai-to-claude/sanitizeToolResultId.ts";
+import {
+  openAiImagePartToClaudeBlock,
+  normalizeToolResultImages,
+} from "./openai-to-claude/imageBlocks.ts";
 
 // Reasoning-effort levels Anthropic accepts on `output_config.effort`. Used to steer
 // adaptive-only Claude models (Opus 4.7+/Fable 5) without ever emitting a manual budget.
@@ -534,9 +541,10 @@ function getContentBlocksFromMessage(
     const sanitizedToolUseId = sanitizeToolResultId(msg.tool_call_id); // #7705
     if (!sanitizedToolUseId) return blocks;
     // T02: Strip empty text blocks from nested tool_result content to avoid Anthropic 400
-    const toolContent = Array.isArray(msg.content)
-      ? stripEmptyTextBlocks(msg.content)
-      : msg.content;
+    // #9692: rewrite OpenAI image_url parts to Claude image blocks (same as user turns)
+    const toolContent = normalizeToolResultImages(
+      Array.isArray(msg.content) ? stripEmptyTextBlocks(msg.content) : msg.content
+    );
     blocks.push({
       type: "tool_result",
       tool_use_id: sanitizedToolUseId,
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Verified this against the actual translator code path — solid fix, scoped correctly.

What I checked:
- Confirmed the base code at `getContentBlocksFromMessage` really did forward raw `image_url`/AI-SDK `image` parts unchanged for `role: "tool"` and nested `tool_result` content, while the sibling `role: "user"` array path already converted them — matches the described defect exactly.
- `imageBlocks.ts` cleanly extracts the shared conversion logic (`extractOpenAiImageUrl`, `urlToClaudeImageBlock`, `openAiImagePartToClaudeBlock`, `normalizeToolResultImages`) and both call sites (`role: "tool"` and nested `tool_result`) now route through it.
- Nice side-effect catch: the old inline code did `part.image_url.url` unconditionally, which throws `TypeError` when a client sends a bare-string `image_url` (no `{url}` wrapper) instead of an object. `extractOpenAiImageUrl` now handles both shapes, and your test suite ("mixed text + image_url...") exercises exactly that case. That's a real latent-crash fix beyond the reported issue.
- Ran the new test file: 7/7 pass. Ran the 5 adjacent openai-to-claude test files (bare-tool, strip-empty, file-attachments, translator-openai-to-claude, empty-messages, glm-split-tool-name) — 32/32 pass, no regressions. ESLint clean on the 3 changed files.
- On the CI reds: I reproduced `mutation-test-coverage`, `dead-code`, and `open-sse-typecheck` failures identically on the unmodified `release/v3.8.50` tip (checked out the exact same scripts against the base commit) — all three are pre-existing base-red (#9985) and unrelated to `openai-to-claude.ts`/`imageBlocks.ts`. The `⚠️ base-red inherited: #9985` note in the PR body is accurate and appreciated.

No blocking issues found. Good scoping decision keeping the K3 protocol-override sibling out of this PR.

PR #10792, opened 2026-08-19
fix(resilience): retry Codex pre-output transport failures on the same account (#9708)

### LINKED ISSUE #9708: fix(resilience): retry transient Codex transport failures on the same account before fallback
### OmniRoute Version
3.8.49

### Installation Method
Docker / Docker Compose

### Operating System
Linux

### Provider(s) Involved
Codex

### Model(s) Involved
Codex text models through `/v1/chat/completions`

### Client Tool
Hermes

### Description

A transient Codex transport failure can currently remove the last eligible account from routing immediately, then cause the next request to receive a misleading quota `429` even though that account has quota and recovers a few seconds later.

I reproduced two variants:

```text
503 upstream connect error or disconnect/reset before headers
reset reason: remote connection failure
```

and:

```text
507 exceeded request buffer limit while retrying upstream
```

In both cases, the routing sequence was:

1. Two accounts were already excluded by the configured quota policy.
2. The remaining eligible account received a transient `503` or `507`.
3. OmniRoute immediately marked that account unavailable and attempted fallback.
4. A client retry about 2-3 seconds later received:

   ```text
   429 All codex accounts reached configured quota threshold
   ```

5. Connection recovery restored the transiently unavailable account immediately afterward.

The final `429` is not an accurate description of the pool. Two accounts were quota-filtered, but the last account was only temporarily unavailable after a transport/proxy failure.

Immediate rotation also discards useful account/session affinity. For Codex, keeping a healthy session on the same account is preferable because switching accounts can lose the warm upstream prompt-cache partition.

### Steps to Reproduce

1. Configure multiple Codex accounts and quota filtering.
2. Have all but one eligible account filtered by their real quota state.
3. Make the remaining account receive a transient pre-output transport failure such as a connection reset `503` or retry-buffer `507`.
4. Retry the client request within 2-3 seconds.
5. Observe that OmniRoute reports all accounts as quota-exhausted even though the remaining account recovers shortly afterward.

### Expected Behavior

Before any useful text or tool call has been sent downstream:

- Retry a retryable `502`, `503`, `504`, `507`, connection reset, or early EOF once on the same account after a short jittered delay, roughly 2-3 seconds.
- Keep session/account affinity for that retry and do not classify the account as quota-exhausted.
- If the same-account retry fails again, apply a short cooldown and then try another eligible account.
- If no other account is eligible, preserve the last transport error or return a mixed-cause availability error such as `503 No Codex accounts currently available: 2 quota-filtered, 1 temporarily unavailable after upstream 507`.
- Return the all-accounts-quota `429` only when every unavailable account was actually excluded by authoritative quota state.

The retry classifier should still avoid wasteful retries:

- authoritative provider `429` / quota reset: rotate or wait for reset;
- `401`: refresh credentials and retry the same account once, then rotate only if refresh/retry fails;
- revoked credentials, account ban, or explicit disable: exclude the account;
- malformed request, unsupported model/tool, context limit, and other deterministic request errors: do not rotate accounts;
- after useful downstream text or a tool call has been emitted: do not transparently replay the request because that can duplicate output or actions.

### Actual Behavior

The first transient `503` or `507` makes OmniRoute rotate immediately. If the other accounts are quota-filtered, a quick client retry is converted into a generic quota `429`. The original transport failure and the mixed reasons for pool unavailability are lost.

### Test Impact

Needs a new integration test

### Error Logs / Output

```shell
quota policy filtered 2 account(s): session usage 100%
remaining account: 503 remote connection failure
account unavailable; trying fallback
429 All codex accounts reached configured quota threshold
connection recovery: account restored
```

The same sequence was reproduced with `507 exceeded request buffer limit while retrying upstream` as the initial error.

### Additional Context

This proposal is intentionally bounded to one same-account retry before useful downstream output. It is not a request for unlimited retries. The goal is to absorb brief transport failures inside the router that owns account health and affinity, while still rotating after a repeated failure.

### Validation Plan

- Simulate one eligible Codex account returning a retryable pre-output `503`, then succeeding on the second attempt. Assert that the same account is used, affinity remains intact, and the client receives the successful response.
- Repeat with `507`, `504`, connection reset, and early EOF.
- Make the same-account retry fail twice and assert that OmniRoute rotates only after the second failure.
- Filter two accounts by quota and make the last account temporarily unavailable. Assert that OmniRoute does not return an all-accounts-quota `429`.
- Assert that an authoritative quota `429`, revoked credential, deterministic `400`, or response after partial output does not enter the unsafe retry path.

Prudhvivuda:
## Summary
- Retry a retryable pre-output transport failure (502/503/504/507, connection reset, retry-buffer overflow) once on the same account after a 2–3s jittered delay, without marking the connection unavailable on the first blip.
- Keeps Codex session/account affinity so a brief proxy reset does not drop the warm prompt-cache partition.
- When the last eligible account is only temporarily unavailable and siblings are quota-filtered, surface a mixed-cause 503 instead of `429 All codex accounts reached configured quota threshold`.

Closes #9708

⚠️ base-red inherited: #9985

## Test plan
- [x] `node --import tsx/esm --test tests/unit/codex-same-account-transport-retry-9708.test.ts`
- [x] `node --import tsx/esm --test tests/unit/sse-auth.test.ts --test-name-pattern quota`
- [ ] Confirm a first 503/507 on the last eligible Codex account retries the same connection and succeeds
- [ ] Confirm a second consecutive transport failure then cools the account and rotates
- [ ] Confirm a quota-filtered pool plus one transport-cooled account does not return the all-quota 429

### DIFF
--- a/changelog.d/fixes/9708-codex-same-account-retry.md
+++ b/changelog.d/fixes/9708-codex-same-account-retry.md
@@ -1,0 +1,1 @@
+- **fix(resilience):** retry a retryable Codex pre-output 502/503/504/507 once on the same account (2–3s jitter) before cooling the connection, and stop translating that mixed pool into an all-accounts quota `429` ([#9708](https://github.com/diegosouzapw/OmniRoute/issues/9708))
--- a/src/sse/handlers/chat.ts
+++ b/src/sse/handlers/chat.ts
@@ -176,6 +176,10 @@ import {
   resolveCooldownAwareRetrySettings,
   waitForCooldownAwareRetry,
 } from "../services/cooldownAwareRetry";
+import {
+  shouldRetrySameAccountTransport,
+  sameAccountTransportRetryDelayMs,
+} from "../services/sameAccountTransportRetry";
 import { constrainConnectionsToQuota, resolveQuotaKeyScope } from "../../lib/quota/quotaKey";
 import { checkConnectionCapacity } from "../utils/backpressure";
 import {
@@ -333,7 +337,11 @@ function isManagedComboUnsupported(
 
 const managedComboRejection = () =>
   buildManagedLeaseErrorResponse(
-    new LeaseContextError(409, "LEASE_UNSUPPORTED_ROUTE", "Managed leases do not support this route")
+    new LeaseContextError(
+      409,
+      "LEASE_UNSUPPORTED_ROUTE",
+      "Managed leases do not support this route"
+    )
   );
 
 const comboPromoteDeps = { updateCombo, info: log.info, warn: log.warn };
@@ -1534,6 +1542,7 @@ async function handleSingleModelChat(
   // re-attempt to exactly one for the whole request. Declared outside both retry
   // loops so it can never reset and loop.
   let streamEarlyEofRetries = 0;
+  const sameAccountTransportRetries = new Map<string, number>();
   const occupancySessionKey =
     runtimeOptions.sessionAffinityKey ?? runtimeOptions.sessionId ?? `request:${randomUUID()}`;
   let initialPreselectedCredentials = runtimeOptions.preselectedCredentials;
@@ -1818,7 +1827,8 @@ async function handleSingleModelChat(
             comboStrategy,
             isCombo,
             comboStepId: runtimeOptions.comboStepId ?? null,
-            comboExecutionKey: runtimeOptions.comboExecutionKey ?? runtimeOptions.comboStepId ?? null,
+            comboExecutionKey:
+              runtimeOptions.comboExecutionKey ?? runtimeOptions.comboStepId ?? null,
             extendedContext,
             modelApiFormat: apiFormat,
             modelTargetFormat: targetFormat,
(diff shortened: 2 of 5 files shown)


diegosouzapw:
Reviewed and validated in an isolated worktree off `release/v3.8.50`.

**Tests**: `codex-same-account-transport-retry-9708.test.ts` — 7/7 pass, including the `getProviderCredentials` mixed-cause (quota + transport-cooldown) 503 case. Existing `sse-auth.test.ts --test-name-pattern quota` — 67/67 pass, no regression. `eslint` clean on all touched files; `tsc --noEmit` clean.

**Circuit-breaker/cooldown interaction**: traced this carefully since it touches the resilience layers described in AGENTS.md — no conflict. The same-account retry only exempts the *first* pre-output 502/503/504/507 blip; a second consecutive failure on the same connection falls through unchanged into the existing `markAccountUnavailable()` / `breaker._onFailure()` path, so provider-breaker and connection-cooldown accounting behave exactly as before for genuinely repeated failures. Nice, surgical addition.

Two things worth a quick look before/after merge, neither blocking:

1. **Scope**: `shouldRetrySameAccountTransport` / `isRetryablePreOutputTransportError` and the new mixed-availability 503 branch in `auth.ts` aren't gated on `provider === "codex"` anywhere — as written this changes retry/error behavior for *every* provider, not just Codex, even though the title/issue/changelog all frame it as Codex-specific. Was that intentional (a general resilience improvement, generalized past the original issue), or should it be scoped to Codex? Either is fine, just want to make sure it's a deliberate call rather than an oversight, since it's a behavior change for the other 340 providers with no dedicated test coverage for them.
2. **`hasEmittedOutput`**: `sameAccountTransportRetry.ts` documents and unit-tests a guard against retrying after output has already streamed to the client, but it's never passed from the real call site in `chat.ts` (confirmed via repo-wide grep). It's harmless today — `result.success` only goes `false` pre-stream in the current control flow (the STREAM_EARLY_EOF/stream_timeout branch that *would* hit mid-stream always exits before reaching this block) — but as written it's a safety param that looks load-bearing and isn't. Worth either wiring it up or dropping it from the signature so it doesn't mislead the next person touching this file.

Also noticed 3 unchecked boxes in the test plan (live-account 503/507 retry, second-failure rotation, mixed-pool 503) — the automated coverage looks solid to me as a substitute, but flagging in case you want to close the loop with a live check per the repo's bug-fix protocol.

Overall: solid, well-scoped fix mechanically. ⭐⭐⭐⭐

PR #10793, opened 2026-08-19
fix(catalog): exclude generic 128k default from combo context min() (#10734)

### LINKED ISSUE #10734: fix(catalog): combo context_length advertised as 128k when members are 500k (generic default counted in min())
## Bug Description

`GET /v1/models` advertises a user-defined combo's `context_length` as **128000** even when every *concrete, catalogued* member of that combo (and of nested combo-refs) is advertised at **500000**.

This is not an explicit `context_length` override on the combo record (that field is unset). The catalog computes the combo window as the **minimum of member windows**. Members that are missing from the live catalog, or that only resolve through `getTokenLimit()`'s generic provider default (`DEFAULT_LIMITS.default = 128000` / `CONTEXT_LENGTH_DEFAULT`), are treated as a *known* 128k window and pull the advertised combo min down to 128k.

Clients that trust `/v1/models` (agent gateways, catalog sync, preflight admission) then hard-stop at 128001 tokens **before** any compression / compaction / upstream request.

Related but distinct: #8942 (context metadata as routing hint, not hard combo rejection), #10141 (approximate token estimates vs admission). This issue is specifically **catalog advertisement of combo `context_length`**.

## Steps to Reproduce

1. Create combo `A` whose members include:
   - a real large-context model that `/v1/models` lists at 500000 (e.g. a SuperGrok / grok-cli `grok-4.x` id)
   - plus at least one member that is **not** in the live catalog, or that has no synced/registry/spec window (so resolution falls through to `getTokenLimit()` → generic 128000)
2. Optionally wrap `A` in combo `B` as a single `combo-ref` (no explicit `context_length`).
3. `GET /v1/models` and inspect `id=A` and `id=B`.

## Expected Behavior

- Combo `context_length` should be the min of **sourced** member windows only (synced / registry / spec / explicit override).
- The generic 128000 fallback must **not** count as a known window for `minKnownNumber`.
- If no sourced window exists, omit `context_length` (or mark it unknown) rather than advertising 128000.
- Nested combo-refs should inherit that same rule, not lock the wrapper to 128k.

## Actual Behavior

- Combo `A` and wrapper `B` both advertise `context_length: 128000` and `max_input_tokens: 128000`.
- Concrete catalogued members of `A` still advertise 500000.
- Agent clients that preflight against the combo id therefore reject prompts that would fit the real target.

## Environment

- OmniRoute: **v3.8.49** (published image)
- Installation: Docker / Kubernetes, single replica, SQLite state
- OS: Linux
- Relevant code (v3.8.49 tree):
  - `src/lib/combos/comboContext.ts` — `computeComboContextLength()` / `minKnownNumber`
  - `src/app/api/v1/models/catalog.ts` — `buildComboCatalogMetadata()`
  - `open-sse/services/contextManager.ts` — `DEFAULT_LIMITS.default = 128000`, `CONTEXT_LENGTH_DEFAULT`
  - `src/app/api/combos/auto/route.ts` — `advertisedContextLength || 128000`

## Error Output

```
GET /v1/models/{combo}
{ "id": "<combo>", "owned_by": "combo", "context_length": 128000, "max_input_tokens": 128000 }

GET /v1/models/{concrete-member}
{ "id": "<provider>/<model>", "context_length": 500000 }
```

No request body / keys / hostnames included.

## Additional Context

Workaround for operators today: set an **explicit** `context_length` on the combo (the documented override), and/or pin downstream catalogs to a real usable window instead of trusting the advertised combo number.

A safer catalog rule: treat `getTokenLimit()`'s generic default as `specific: false` (that flag already exists on `resolveTokenLimit`) and exclude non-specific values from combo `min()`.

Prudhvivuda:
## Summary
- Combo `context_length` in `/v1/models` is the min of *sourced* member windows only (canonical window or a provider/model-specific token limit).
- `getTokenLimit()`'s generic 128000 catch-all (`specific: false`) is no longer treated as a known window, so a 500k combo is not advertised as 128k because an unsourced or generic-default member entered `min()`.
- Explicit combo `context_length` overrides still win. Nested combo-refs inherit the same sourced-min rule.

Closes #10734

⚠️ base-red inherited: #9985

## Test plan
- [x] `node --import tsx/esm --test tests/unit/combo-context-generic-default-10734.test.ts`
- [x] `node --import tsx/esm --test tests/unit/combo-context-prefix-resolution.test.ts`
- [x] `node --import tsx/esm --test tests/unit/models-catalog-combo-metadata.test.ts`
- [x] `node --import tsx/esm --test tests/unit/resolveComboContextLimit.test.ts`
- [ ] `GET /v1/models` for a combo whose catalogued members are 500k (plus an unsourced member) advertises 500000, not 128000

### DIFF
--- a/changelog.d/fixes/10734-combo-context-generic-default.md
+++ b/changelog.d/fixes/10734-combo-context-generic-default.md
@@ -1,0 +1,1 @@
+- **fix(catalog):** stop counting `getTokenLimit()`'s generic 128k catch-all as a known combo window, so `/v1/models` advertises the min of sourced member contexts instead of collapsing a 500k combo to 128k ([#10734](https://github.com/diegosouzapw/OmniRoute/issues/10734))
--- a/open-sse/services/contextManager.ts
+++ b/open-sse/services/contextManager.ts
@@ -282,6 +282,29 @@ export function getTokenLimit(
 }
 
 /**
+ * Context window from a known source only: an explicit canonical window, or a
+ * provider/model-specific `resolveTokenLimit` result. The generic 128000
+ * catch-all (`specific: false`) is treated as unknown so combo `min()` does
+ * not advertise 128k when every real member is larger (#10734).
+ */
+export function getSourcedTokenLimit(
+  provider: string,
+  model: string | null = null,
+  canonicalWindow?: unknown,
+  snapshot?: ModelCapabilityResolutionSnapshot | null
+): number | undefined {
+  if (
+    typeof canonicalWindow === "number" &&
+    Number.isFinite(canonicalWindow) &&
+    canonicalWindow > 0
+  ) {
+    return canonicalWindow;
+  }
+  const resolved = resolveTokenLimit(provider, model, snapshot);
+  return resolved.specific ? resolved.limit : undefined;
+}
+
+/**
  * Resolve a combo target's token limit without crashing when `parseModel(modelStr)`
  * returns `provider: null` (model id with no `provider/` prefix).
  *
@@ -315,7 +338,7 @@ export function getComboTargetTokenLimit(options: {
  * name heuristic, curated per-provider default) or only from the generic
  * catch-all default.
  */
-function resolveTokenLimit(
+export function resolveTokenLimit(
   provider: string,
   model: string | null = null,
   snapshot?: ModelCapabilityResolutionSnapshot | null
--- a/src/app/api/v1/models/catalog.ts
+++ b/src/app/api/v1/models/catalog.ts
@@ -68,7 +68,7 @@ import {
   isNoAuthRawProviderPrefix,
   normalizeBlockedProviderSet,
 } from "@/shared/utils/noAuthProviders";
-import { getTokenLimit } from "@omniroute/open-sse/services/contextManager";
+import { getSourcedTokenLimit, getTokenLimit } from "@omniroute/open-sse/services/contextManager";
 import { extractApiKey } from "@/sse/services/auth";
 import type { ComboModelStep } from "@/lib/combos/steps";
 import {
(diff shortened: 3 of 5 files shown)


diegosouzapw:
Verified this fix end-to-end -- the production change is correct and does what the title says.

**Confirmation:** I built a real (not fabricated) combo `{glm/glm-5.2, aimlapi/deepseek-chat}` -- `aimlapi/deepseek-chat` is a genuinely registered model whose `resolveTokenLimit()` falls through to the generic 128k catch-all (`specific: false`), matching the exact scenario in #10734. With this PR's `getSourcedTokenLimit()` in place, `computeComboContextLength()` returns `1000000` as expected. Reverting just the two `getSourcedTokenLimit` call sites back to `getTokenLimit` reproduces the bug exactly (`128000`). All 4 tests named in the test plan pass, plus the shipped `combo-context-generic-default-10734.test.ts` suite (7/7).

**One suggestion (non-blocking):** 3 of the 7 shipped tests -- including the flagship "GET /v1/models does not advertise 128k" integration test -- use `not-a-real-provider/no-such-model` as the "unsourced member" fixture. That model is filtered out by a separate, pre-existing gate (`getCanonicalModelMetadata`'s source check for "not registered in any source at all"), independent of this fix. I confirmed this by reverting the production change and re-running the suite: all 7 tests still pass unchanged. Only the 4 lower-level `resolveTokenLimit`/`getSourcedTokenLimit` unit tests actually fail without the fix.

For a regression guard that actually exercises `getComboTargetCatalogMetadata`/`computeComboContextLength`'s new branch, consider swapping that fixture for a real registered-but-unsourced model (e.g. `aimlapi/deepseek-chat`, or any `PROVIDER_MODELS` entry with no canonical `contextWindow` and no `REGISTRY.defaultContextLength`/`DEFAULT_LIMITS` entry for its provider) in a follow-up. Doesn't block this PR -- the production fix is sound and well-scoped as-is.

⚠️ Noted: base-red #9985 is correctly flagged in the PR body.

PR #10473, opened 2026-08-15
fix: stop leaking upstream SSE control lines to OpenAI-format clients (#10017)

diegosouzapw:
Closes #10017

## Root cause

In "Standard passthrough mode" (source format === client format, no translation needed), `createSSEStream` buffers upstream SSE control lines (`id:`, `event:`, `retry:`, bare `:` comments) via `createSSEEventPrefixBuffer()` and re-prepends them verbatim onto the next `data:` chunk it forwards. For a plain OpenAI Chat-Completions-format client (`clientResponseFormat === FORMATS.OPENAI`), that leaked literal `id: 0`, `event: done`, `: proxy-internal-metadata` lines into the client stream — a protocol OpenAI's own Chat-Completions SSE never uses. Strict clients that line-parse for `data:` prefixes break.

The buffering exists deliberately for two protocols that DO require `event:` framing per frame: OpenAI Responses API and Claude Messages API (proven by the existing `tests/unit/responses-commentary-event-frame-6561.test.ts`). But the same buffer also unconditionally re-forwarded `id:`/`retry:`/comment lines, and forwarded `event:` even to plain OpenAI Chat-Completions clients.

## Fix

`createSSEEventPrefixBuffer()` in `open-sse/utils/streamHelpers.ts` is now format-aware:
- `id:`/`retry:`/bare `:` comment lines are never buffered/forwarded for ANY client format — they are not part of Chat-Completions, Responses, or Claude Messages SSE.
- `event:` framing is preserved ONLY when the client format is `FORMATS.OPENAI_RESPONSES` or `FORMATS.CLAUDE`; it is dropped for plain OpenAI Chat-Completions clients.

`open-sse/utils/stream.ts` passes the resolved `clientResponseFormat` into the buffer.

## Regression test (fail → pass)

New `tests/unit/10017-sse-control-lines-leak-openai-clients.test.ts`. It reproduces the exact report: 4 raw upstream control lines (`id: 0`, `id: 1`, `event: done`, `: proxy-internal-metadata`) previously leaked to an OpenAI-format client output. RED on the unfixed code:

```
AssertionError: expected NO raw upstream SSE control lines forwarded to an OpenAI-format client, got: ["id: 0","id: 1","event: done",": proxy-internal-metadata"]
```

Green after the fix, plus companion assertions that Responses `event: response.output_text.delta` and Claude `event: message_start` framing is unchanged for those formats (guarding the existing #6561 contract).

## Gates run green

[...]

Issue #10017, opened 2026-08-10
bug(backend): strip upstream SSE control lines (id:/event:/retry:/:) for OpenAI SDK compatibility


Jensenwgd:
## Problem

When proxying streaming responses from certain upstream providers, the proxy forwards SSE control lines verbatim to downstream clients.

Some upstreams emit valid SSE with control lines:
```
data: {"id":"...","choices":[{"delta":{"content":"Hello"}}]}
id: 0
data: {"id":"...","choices":[{"delta":{"content":"!"}}]}
id: 1
...
event: done
: proxy-internal-metadata
```

These `id:`, `event:`, `retry:`, `:` lines are **valid per SSE RFC**, but **break OpenAI Python SDK** and other clients that expect pure `data: <json>` lines.

## Expected Behavior

Proxy should normalize upstream SSE to standard OpenAI format:
- Strip `id:`, `event:`, `retry:`, `:` control lines
- Only forward `data: <json>` lines (and `data: [DONE]`)

## Evidence

- **Direct connection works**: Connecting downstream client (OpenAI Python SDK) directly to upstream works — the SDK ignores control lines
- **Via proxy fails**: Same upstream, proxied through OmniRoute, breaks OpenAI Python SDK with `JSONDecodeError: Expecting value: line 1 column 1` (on `id: 0` line)

## Root Cause (Hypothesis)

Proxy layer forwards raw SSE bytes without normalizing to standard OpenAI streaming format.

## Environment

- OmniRoute version: 3.8.49
- Downstream: OpenAI Python SDK compatible client

---

**No patch provided** — local workaround applied but not upstreamed. Happy to test any fix.

diegosouzapw:
[...]

PR #10795, opened 2026-08-19
fix(ci): drain unit test base-reds (i18n pt-BR, inventory, CC provider route) (#9985)

rqzbeh:
## Summary

Stacked on PR #10778 to drain the remaining 5 unit test base-red failures on `release/v3.8.50` (#9985):

1. **`src/app/api/providers/[id]/models/route.ts`**: Moved `isClaudeCodeCompatibleProvider` check before cached discovery lookups so CC-compatible connections return HTTP 400 instead of returning HTTP 200 with cached models.
2. **`src/i18n/messages/pt-BR.json`**: Added 6 missing Portuguese translation keys to match `en.json` (`settings.compressionOutputStyle.i-have-adhd.label`, `description`, `reasoningTokenBuffer`, `reasoningTokenBufferDesc`, `zeroLatencyOptimizations`, `zeroLatencyOptimizationsDesc`).
3. **`tests/unit/hard-session-lease-bypass-inventory.test.ts`**: Updated `EXPECTED.credential` and `EXPECTED.executor` inventory baselines for 7 new connection/credential call sites (`speechCombo.ts`, `videoCombo.ts`, `videoModelResolution.ts`, `transcriptions/route.ts`, `classify/route.ts`, `segment/route.ts`, `geminiWeb.ts`).

## Verification

- `node --import tsx/esm --test tests/unit/i18n-pt-br.test.ts` (3/3 passed)
- `node --import tsx/esm --test tests/unit/cc-compatible-provider.test.ts` (27/27 passed)
- `node --import tsx/esm --test tests/unit/hard-session-lease-bypass-inventory.test.ts` (3/3 passed)

PR #10796, opened 2026-08-19
fix(tests): realign two more stream-utils passthrough cases with the #10017 control-line rules

MichaelYcJo:
Follow-up to #10726 (merged). Two more `stream-utils` cases are red on `release/v3.8.50` for the same reason the ones in that PR were: #10473 changed what passthrough forwards and only added new coverage, leaving these asserting the framing it replaced.

Test-only change.

## The two cases

**`forwards data only after the complete SSE event boundary`** — the fixture emits `response.output_text.delta` against a `responses-model`, so it is an OpenAI Responses scenario. `event:` framing is forwarded for exactly that format and for Claude Messages (`stream.ts:833`), so the case now declares `clientResponseFormat` and passes on its own terms.

**`preserves event metadata in a single SSE event`** — this one cannot be fixed by declaring a format. `:` comments, `id:` and `retry:` are dropped for *every* client format, since none of the three supported SSE protocols define them (`streamHelpers.ts` → `remember`). The expectation now pins what the case is actually about — the surviving lines stay inside ONE event instead of being split apart by blank lines — and asserts the dropped lines are gone.

## Verification

```
node --import tsx/esm --test tests/unit/stream-utils.test.ts
  base 3d7ed7aa8 : 4 failed | 48 passed
  this branch    : 2 failed | 50 passed
```

`npm run test:vitest` on the same base is already fully green (368/368).

## Two failures in the same file are NOT addressed here

They are left alone deliberately — both look like behaviour changes rather than stale expectations, and getting them wrong would affect billing and tool routing rather than just a test.

**`passthrough merges Claude usage chunks and restores mapped tool names`** — the fixture's `message_start` carries `usage: { input_tokens: 6 }`, but the completion payload reports `prompt_tokens: 33`. An upstream-reported input count appears to be replaced by an estimate. If that is intended, the case needs a new expectation; if not, the estimate is overriding real usage data and the fix belongs in the usage merge path, not in this test.

**`translate mode converts Claude SSE into OpenAI chunks and completion payload`** — fails in the same file and likely shares that cause.

Also still red on this base and untouched here: `tests/unit/translator-tool-call-shim.test.ts` → `streaming: Read suppresses raw pages delta and emits cleaned input at finish`. The shim itself is correct (`/etc/hosts` is not a PDF, so `sanitizeReadArgs` deletes `pages`), but the streaming path through `openaiToClaudeResponse` emits the uncleaned input — one delta is produced, so the shim is reached, yet `"pages"` survives in `partial_json`.

Happy to take any of these in a follow-up if the intended behaviour is confirmed.

⚠️ base-red inherited: #9985 — `release/v3.8.50` was already red when this branch was cut.


diegosouzapw:
Verified independently: rebuilt the passthrough SSE prefix-buffer logic in `open-sse/utils/streamHelpers.ts` (`createSSEEventPrefixBuffer`/`remember`) and `stream.ts` against these two test cases, and ran `node --import tsx/esm --test tests/unit/stream-utils.test.ts` on the base tip and on this branch in an isolated worktree — base fails 4, this branch fails only the 2 you flagged as deliberately untouched (usage-merge, translate-mode). The new assertions match current behavior exactly (control-line drop is unconditional per `remember()`; `event:` forwarding gates on `clientResponseFormat`), so this isn't a weakened assertion, it's a genuine realignment. No overlap with #10795's file set. Merging as-is.

Issue #6149, opened 2026-07-04
fix: Vercel relay deploy — ReferenceError resolveRelayTarget is not defined (api/relay.js:98)


SeaXen:
## Summary

Vercel relay deployed from **Dashboard → System → Proxy pool** fails the built-in **Proxy Test** with HTTP 500. Vercel function logs show a **ReferenceError** in the generated worker — not a configuration or credential problem on the self-hosted OmniRoute instance.

## Environment

- **Image:** `diegosouzapw/omniroute:latest` (pulled 2026-07-04; release area v3.8.42)
- **Host:** self-hosted, `network_mode: host`, port `20128`
- **Deploy path:** UI → Proxy pool → Vercel relay (API token valid — deploy succeeds, new hostname is created)

## Steps to reproduce

1. Open **System → Proxy pool** and deploy a **Vercel** relay from the UI.
2. Wait for deploy success (e.g. `omniroute-relay-<id>-<team>.vercel.app`).
3. Click **Test** on the relay entry (or assign to provider and test).

## Expected

Proxy test succeeds (forwards to `https://api64.ipify.org` per test route) and returns public IP JSON.

## Actual

- UI: `✗ Relay returned HTTP 500`
- OmniRoute log: `[ProxyTest] relay <host>: Relay returned HTTP 500`

## Vercel function log (production)

```
ReferenceError: resolveRelayTarget is not defined
    at (api/relay.js:98:19)
```

- **Route:** `api/relay`
- **Path:** `GET /`
- **Error:** `FUNCTION_INVOCATION_FAILED`
- **External APIs:** No outgoing requests (worker crashes before fetch)
- **Example host:** `omniroute-relay-k32m4ofjg-searex.vercel.app`
- **Example deployment ID:** `dpl_GdvgPKBVVvkRXJXz15WKqcx5LM4s`
- **Example request ID:** `5dgvh-1783136548239-11270402c293`

## Additional verification (rules out operator misconfig)

| Client test | Result |
|-------------|--------|
| `GET /` without `x-relay-auth` | 401 (expected) |
| `GET /` with valid `x-relay-auth` only | 400 `missing x-relay-target` (auth accepted) |
| Full proxy-test headers (`x-relay-auth`, `x-relay-target=https://api64.ipify.org`, `x-relay-path=/?format=json`) | 500 `FUNCTION_INVOCATION_FAILED` |

[...]

Issue #1360, opened 2026-04-17
[BUG] kimi-k2.5 via opencode-go returns 400 on agentic requests (still unfixed)


matteoantoci:
### OmniRoute Version

3.6.7

### Installation Method

npm global

### Operating System

macOS

### OS Version

Darwin 25.4.0

### Node.js Version

v22.21.1

### Provider(s) Involved

opencode-go

### Model(s) Involved

opencode-go/kimi-k2.5

### Client Tool

Claude Code (CLI), any agentic client

### Description

`kimi-k2.5` via `opencode-go` returns 400 when the message history includes assistant messages with `tool_calls`. The upstream API requires `reasoning_content` on every assistant message that contains `tool_calls` when thinking/reasoning is enabled.

The `isReasoner` check in `open-sse/translator/index.ts` (~line 209) only matches `deepseek` and models with `r1|reason` in the name. `kimi-k2.5` is not included, so `reasoning_content` is never injected for it.

**Related:** #1273 reported the same bug. PR #1281 was intended to fix it but was closed without merging (`mergedAt: null`). The bug is still present in 3.6.7.

### Steps to Reproduce

1. Configure `opencode-go` provider with a valid API key
2. Add `opencode-go/kimi-k2.5` to a combo
3. Use an agentic client that sends multi-turn tool-call history
4. Observe 400 from kimi's upstream API on the second+ turn (when assistant messages with `tool_calls` appear in history)

### Expected Behavior

`reasoning_content = ""` should be injected on assistant messages with `tool_calls` for kimi models, same as for DeepSeek.

### Actual Behavior

Requests with tool-call history are forwarded without `reasoning_content`, causing kimi's API to reject them with 400.

### Error Logs

```
[ERROR] [400]: Provider returned error
{"status":400,"msg":"Model opencode-go/kimi-k2.5 failed (no fallback)"}
```

diegosouzapw:
Hi @matteoantoci, thanks for another thorough report!

**Confirmed bug**: The `isReasoner` check in `open-sse/translator/index.ts` (line 209-210) only matches `deepseek` provider or models with `r1|reason` in the name:

```ts
const isReasoner = provider === "deepseek" || /r1|reason/i.test(model);
```

[...]

PR #10935, opened 2026-08-21
fix(relay): resolve x-relay-path through the shared guard in the CF worker

ntdat812:
⚠️ base-red inherited: #9985

## Problem

PR #4643 and its follow-up replaced string concatenation of the attacker-controlled `x-relay-path` with the pure `resolveRelayTarget()` guard — in the **Deno** and **Vercel** relay workers. The **Cloudflare** worker is the third generator of the same shape and still concatenates:

```js
const targetBase = target.endsWith("/") ? target.slice(0, -1) : target;
const upstream = await fetch(targetBase + relayPath, init);
```

Validating `x-relay-target` and *then* concatenating is not sufficient — the path re-points the request past the host that was just checked. The vectors are the ones the existing guard already documents and the other two workers already have tests for: userinfo (`/x@evil.com`), a backslash (`\evil.com`), and a protocol-relative path (`//evil.com/x`).

`src/lib/proxyRelay/cloudflareWorkerScript.ts` was ported from upstream `decolua/9router` PR #1360 and diverged from the Vercel/Deno pair afterwards, which is how it missed the follow-up.

## Change

The Cloudflare worker now embeds and calls the **same** `resolveRelayTarget()` the other two use — no new guard, no second policy to keep in sync:

```js
const resolved = resolveRelayTarget(target, relayPath);
if (!resolved.ok) {
  return new Response(resolved.reason, { status: resolved.status });
}
const upstream = await fetch(resolved.url, init);
```

Bound to a **literal `const resolveRelayTarget =` name**, not a bare declaration — per #6149, `Function#toString` emits the *mangled* name on the SWC-minified standalone build, so a bare declaration would define `<mangled>` while the template still calls the hardcoded identifier. The new test pins that property for this worker too by renaming the embedded function and re-evaluating the emitted source, the same technique `relay-minified-fn-6149.test.ts` uses.

The auth check and the private/loopback target guard are untouched.

## Verification

`tests/unit/cloudflare-relay-path-ssrf.test.ts` (new) mirrors `vercel-deploy-relay-path-ssrf.test.ts`.

On `release/v3.8.50` with only the source file reverted:

```
not ok 1 - cloudflare relay worker — no string-concat SSRF hole
ok 2   - cloudflare relay worker — host-confusion vectors are rejected
# tests 9  # pass 6  # fail 3
```

All 9 pass with the change. The 6 that pass either way exercise the shared pure function, which was already correct — the wiring is what was missing, so those are the over-rejection guard rather than evidence of the bug.

| check | result |
| --- | --- |
| `tests/unit/*relay*` + `cloudflare-worker-upload-content-type-6416` | **104 passed, 0 failed** |
| `npm run typecheck:core` | pass |
| `npx eslint` on both changed files | clean |

## Note

`buildCloudflareWorkerScript` now imports from `src/app/api/settings/proxy/deno-deploy/route` — the same import the Vercel route already makes for this guard. If you would rather the shared guard move to `src/lib/proxyRelay/` (or `src/shared/`) so a `lib` module is not reaching into an app route, say so and I will do that instead; I kept it as-is to avoid moving an audited security function in the same PR that changes a call site.


ntdat812:
Pushed `4c67234` — rebased onto current `release/v3.8.50` (was 13 commits behind; no conflicts, none of those commits touch `proxyRelay`). Content is unchanged.

The red checks on this PR are inherited from base, and I verified that rather than assuming it. Checking out `origin/release/v3.8.50` itself, clean tree, same `node_modules`:

- **Unit Tests fast-path** — `check-docs-counts-sync.mjs` reports the same **5 STRICT drifts** on base HEAD `34f65e1` with none of this PR's changes present: `PROVIDER_REFERENCE.md` doc total 346 vs 347 live provider modules, and four SVG diagrams carrying "346 providers". `npm run gen:provider-reference` is the fix and it belongs to whoever added the 347th provider.
- **No new ESLint warnings** — the ratchet fails on `openapiCoverage.pct: 38.6 < baseline 39.2`. That metric counts `route.ts` files under `src/app/api` against `docs/openapi.yaml`; this PR adds and deletes zero of them, so it cannot move either side of the ratio.

`Merge integrity (changelog + generated skills)`, `Change Classification`, `Vitest`, `Build`, `semgrep` and `dast-smoke` all pass here — base's own last full run failed merge-integrity as well, so this branch is strictly greener than what it branches from.

Relay tests after the rebase: **153 passed, 0 failed** across every test file that references `proxyRelay` or the three deploy routes. `typecheck:core` exit 0, eslint clean.

ntdat812:
Post-rebase run is red on the same gates, and now the ratchet one can be settled exactly rather than by argument. Since the rebase, this branch and `release/v3.8.50` have an **identical** `src/app/api` route set, so I ran the `openapiCoverage` computation from `scripts/quality/collect-metrics.mjs` on both:

```
branch fix/cloudflare-relay-path-guard   routes=673 covered=260 pct=38.6
base   origin/release/v3.8.50            routes=673 covered=260 pct=38.6
```

Base itself produces 38.6 against a frozen baseline of 39.2 — the baseline is stale relative to base, and this PR cannot move a number it contributes zero routes to.

The four `Unit Tests fast-path` shards fail on tests with no relation to the relay: `transformers + onnxruntime-node are regular dependencies`, `real handshake: systemd-notify delivers READY=1/WATCHDOG=1/STOPPING=1`, `#10225`/`#10503` context-overflow, `#9199` capability mutation, `the gate exits 0 against the current (synced) repo state` (the `check-docs-counts-sync` drift documented above). The relay tests themselves pass — **153 passed, 0 failed** locally across every file that touches `proxyRelay` or the three deploy routes.

Same picture on #10941, which stacks on this branch.

Nothing here needs action from me that I can see; flagging it so the red checks are not read as this PR's doing. If you would rather I hold both until base is green again, say so and I will.

Issue #6416, opened 2026-07-06
fix(providers): Cloudflare relay Worker upload fails with invalid Content-Type despite valid token/account


SeaXen:
## Summary

Cloudflare relay deploy from **Dashboard → System → Proxy pool** fails immediately with:

`Cloudflare Worker upload failed: Content-Type must be one of: application/javascript, text/javascript, multipart/form-data`

This does **not** appear to be a Cloudflare credential/account problem.

## Environment

- **Image:** `diegosouzapw/omniroute:latest`
- **Version tested:** `v3.8.45`
- **Host:** self-hosted, `network_mode: host`, port `20128`
- **Deploy path:** UI → Proxy pool → Cloudflare relay

## Actual

Cloudflare relay deploy fails immediately with:

`Cloudflare Worker upload failed: Content-Type must be one of: application/javascript, text/javascript, multipart/form-data`

## Expected

Cloudflare relay should deploy successfully, same as a direct Worker upload through the Cloudflare API.

## External verification (rules out token/account problems)

Using the **same Cloudflare credentials outside OmniRoute**:

1. **Token verification succeeds** via Cloudflare API (`/user/tokens/verify`)
2. **Workers API access succeeds** for the target account
3. **Workers subdomain exists**
4. **Listing scripts succeeds**
5. **Direct Worker upload succeeds** with the same credentials when using:
   - `PUT /client/v4/accounts/<account_id>/workers/scripts/<name>`
   - `Authorization: Bearer <token>`
   - `Content-Type: application/javascript`
   - raw Worker JS body
6. The temporary test Worker can also be **deleted successfully** afterward

So the Cloudflare token, account, and Workers permissions are valid.

## Interpretation

This strongly suggests the OmniRoute Cloudflare relay deploy path is sending the upload request in the wrong format, for example:

- wrong `Content-Type`
- wrong request body shape
- mismatch between declared content type and actual payload
- wrong upload mode for Cloudflare Workers API

## Suggested fix area

Please compare the OmniRoute Cloudflare relay upload request against the known-working Cloudflare API shape:

- raw JS body + `Content-Type: application/javascript`
- or proper `multipart/form-data` if the route is intended to use module upload semantics

Right now Cloudflare accepts the same credentials for direct upload, but OmniRoute UI deploy fails before a Worker is created.

## Related context

[...]

PR #10941, opened 2026-08-21
fix(relay): share one private-host guard across the three relay workers

ntdat812:
⚠️ base-red inherited: #9985

> **Stacked on #10935** — that PR is the first commit here. Review/merge it first; this one rebases to nothing extra once it lands. Happy to re-cut against `release/v3.8.50` if you would rather take them independently.

## Problem

The Cloudflare, Deno Deploy and Vercel relay generators each inline their own copy of `isPrivateHostname`. The Deno and Vercel copies are **byte-identical**; the Cloudflare one differs only in formatting. One policy, three copies — so a gap has to be found and fixed three times.

It had four, measured by driving `new URL(target).hostname` exactly the way the workers do:

| target | `url.hostname` | current guard |
| --- | --- | --- |
| `::` | `[::]` | **allowed** |
| `localhost.` | `localhost.` | **allowed** |
| `::127.0.0.1` | `[::7f00:1]` | **allowed** |
| `feb0::1` | `[feb0::1]` | **allowed** |

- **`::`** is the unspecified address; `http://[::]/` reaches a service bound to the IPv6 loopback.
- **`localhost.`** — the FQDN root dot defeats the exact match *and* all three suffix rules, so `svc.internal.` was allowed too. This one is runtime-independent.
- **`::127.0.0.1`** is the deprecated IPv4-compatible form; only `::ffff:` was checked.
- **`feb0::1`** — link-local is `fe80::/10`, i.e. `fe80`–`febf`; the check matched only the literal `fe80:` spelling.

## Change

The policy moves to `src/lib/proxyRelay/privateHostname.ts` and is embedded verbatim via `Function#toString` — the mechanism `resolveRelayTarget` already uses for these same three workers — bound to a literal `const isPrivateHostname =` name so a minified standalone build cannot break the call sites (#6149). Net **−83/+6** across the three generators.

Two simplifications fall out of fixing it once:

- `::/96` collapses to a single `startsWith("::")`: unspecified, IPv6 loopback, IPv4-mapped and IPv4-compatible are all non-routable, so none is a legitimate relay target.
- Link-local becomes `/^fe[89ab]/` instead of one spelling.

**Nothing previously blocked is now allowed.** Verified still allowed: `example.com`, `api.anthropic.com`, `8.8.8.8`, `2606:4700::1111`, and `example.com.` with a root dot (the dot-stripping only affects the private checks).

## Severity — deliberately not overstated

Reaching the worker requires the `x-relay-auth` secret, and these are edge runtimes where loopback has nothing listening. This is defence-in-depth that did not hold, not a live SSRF into an operator's network. The suffix-rule bypass is the part that is real regardless of runtime.

## Verification

`tests/unit/relay-private-host-guard-gaps.test.ts` — 35 cases: the four gaps, the trailing-dot form of every suffix rule, all previously-correct verdicts, and a per-worker check that each embeds the shared binding and that the *embedded* copy still closes the gaps (evaluated through `node:vm`, not `new Function` — Hard Rule #3).

**9 fail on the branch base** — the wiring assertions for all three workers. All 35 pass with the change.

| suite | result |
| --- | --- |
| `relay-private-host-guard-gaps` | 35 pass |
| `cloudflare-relay-path-ssrf` | 9 pass |
| `deno-deploy-relay-path-ssrf` | 10 pass |
| `vercel-deploy-relay-path-ssrf` | 8 pass |
| `relay-minified-fn-6149` | 4 pass |
| `cloudflare-worker-upload-content-type-6416` | 4 pass |

`npm run typecheck:core` passes; `npx eslint` on all five changed/added files is clean.


ntdat812:
Pushed `4bab59c` — rebased onto current `release/v3.8.50`, plus a fix for two existing tests this PR broke.

## I broke two existing tests and did not catch it

Both read the guard out of the emitted worker by **source shape**, which this PR changes:

| test | how it read the guard | why it broke |
| --- | --- | --- |
| `relay-deploy-5128` (#6416) | `match(/function isPrivateHostname\(h\) \{[\s\S]*?\n\}/)` | the guard is now `const isPrivateHostname = <shared fn>` |
| `proxy-pool-cloudflare-workers-deployer` | `assert.ok(/169\.254|link-local|fe80/.test(src))` | the shared guard expresses those as an octet comparison and `/^fe[89ab]/`, and the transpiler strips the comment |

Neither is a behaviour regression — the guard blocks strictly more than before — but both were real red tests and I reported this PR as green. The cause: my verification ran a hand-picked list of five test files instead of the full `tests/unit/*relay*` glob, so `relay-deploy-5128.test.ts` was never executed. That was my error, not a flake.

## The fix

Both tests now evaluate the **embedded** guard (`node:vm`, per Hard Rule #3) and assert what it blocks, instead of matching its source text:

```ts
const binding = src.match(/const isPrivateHostname = [\s\S]*?;\s/);
vm.runInContext(`${binding[0]} globalThis.__guard = isPrivateHostname;`, context);
// then: 127.0.0.1, localhost, 0.0.0.0, 10.x, 192.168.x, 172.16.x,
//       169.254.169.254, fe80::1  → blocked;  api.example.com → allowed
```

Proven non-vacuous: deleting the `a === 169 && b === 254` rule from the shared guard **fails** the new assertion. The substring grep it replaces could not have detected that — it only ever checked that the string `169.254` appeared somewhere in the source.

## Verification

Full glob this time, `npx tsx --test` on all nine test files that reference `proxyRelay` / the three deploy routes:

```
tests/unit/*relay*.ts
tests/unit/cloudflare-worker-upload-content-type-6416.test.ts
tests/unit/proxy-pool-cloudflare-workers-deployer.test.ts
tests/unit/vercel-deploy-sso-protection-check.test.ts

# tests 153  # pass 153  # fail 0
```

`npm run typecheck:core` exit 0. `npx eslint` clean on all five changed files.

## The red checks are inherited, and here is the proof

I checked out `origin/release/v3.8.50` itself (clean tree, same `node_modules`) and ran the two failing gates:

**Unit Tests fast-path** — `check-docs-counts-sync.mjs` on base HEAD `34f65e1`:

```
✗ docs/reference/PROVIDER_REFERENCE.md — doc total 346 is stale — the live provider modules define 347
✗ docs/diagrams/readme-hero.svg — stale: "346 providers", "346 AI providers"
✗ docs/diagrams/promise-pillars.svg — stale: "346 providers"
✗ docs/diagrams/comparison-table.svg — stale: "346 providers"
✗ docs/diagrams/cli-terminal.svg — stale: "346 providers"
✗ 5 STRICT drift(s) detected.
```

Identical 5 drifts, same files, with none of this PR's changes present. `npm run gen:provider-reference` is the fix and it belongs to whoever added the 347th provider, not here.

**No new ESLint warnings** — the ratchet fails on `openapiCoverage.pct: 38.6 < baseline 39.2`. That metric is `route.ts` files under `src/app/api` matched against `docs/openapi.yaml`. This PR adds and deletes **zero** `route.ts` files (it modifies two), so neither side of the ratio can move; the drift arrived with the routes added to base since this branch's merge-base. I rebased anyway so the comparison is against the current baseline rather than a stale one.

A caution worth recording: my first attempt to establish this used `git worktree` checkouts, which have no `node_modules`. `readCodeFacts()` shells out to import TS sources, silently fell back to "could not determine code facts — skipping", and the base **looked** clean. The numbers above come from checking the base out in the main repo where the imports actually resolve.

Issue #10718, opened 2026-08-19
fix(providers): copilot-m365-web chat invocation silently dropped by substrate (immediate type:3 close) and no token refresh


acc0mplish:
### Problem

`copilot-m365-web` completes the SignalR handshake but the turn dies immediately after the `type:4` chat invocation: the server replies with a bare `type:3` completion carrying no content, so every request surfaces as `(empty response)`. Reproduced on a real EDU (A3/Starter) tenant with a valid access token — the WS handshake itself passes (101), so auth is not the issue.

Additionally, the browser-issued `access_token` lives ~75 minutes and there is no refresh path, so the provider needs a fresh DevTools capture after every expiry.

### Root cause (invocation)

A fresh capture of a working `m365.cloud.microsoft/chat` round-trip (2026-08, TLS-MITM of the browser WS) shows the client now sends a materially different `type:4` argument shape than `buildChatInvocation` produces:

- `conversationId` matching the WS URL query is present in `arguments[0]`
- `productThreadType: "Office"` and a populated `clientInfo` (`mcmcopilot-web` / `Office`)
- a rich `message` object (`entityAnnotationTypes`, `experienceType`, `locale`, `locationInfo`, `requestId`, `adaptiveCards`, `clientPreferences`, `attachments`)
- the invocation is immediately followed by a `type:1 target:"Metrics"` frame **in the same socket write** — an invocation without its Metrics follow-up is dropped
- no `spokenTextMode` / `extraExtensionParameters` / `isSbsSupported` / `renderReferencesBehindEOS`
- a six-entry `allowedMessageTypes` list and a 14-entry `optionsSets` list (the old 11/25-entry lists are no longer observed on the wire)

The stale shape is closed immediately; omitting the Metrics pair gets the invocation silently ignored (no update frames at all).

### Root cause (refresh)

Same capture-side investigation confirms the token endpoint accepts the first-party public client (`c0ab8ce9-…`, the `appid` the browser token already carries) for a `refresh_token` grant with the `substrate.office.com/sydney` scopes — no client secret. A device-code bootstrap against the FOCI client (`d3590ed6-…`) mints a family-portable refresh_token the web client can then redeem continuously.

### Proposed fix

1. Align `buildChatInvocation` with the recaptured shape and add the Metrics follow-up frame (sent in one write, replacing the leading keepalive ping).
[...]

Issue #9147, opened 2026-08-01
fix(providers): Overload occurs when there are too many connections/models. 


hung319:
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

macOS

### Description

Currently, I'm constantly encountering the "Server is unreachable. Reconnecting..." error when performing any operation. I suspect this is a database bottleneck. 

### Steps to Reproduce

1. Access the web UI. 
2. Perform an action such as opening a new page or editing data. 
3. Reload the page, and everything runs very slowly. 

### Expected Behavior

Smooth operation.

### Actual Behavior

Slowness.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

hung319:
I also encountered a similar problem when calling the API: the first attempt was successful, the second resulted in a 502 error (possibly due to a timeout), and the third attempt was successful.

hung319:
Although a little late, this is a video related to the error. https://github.com/user-attachments/assets/2a9951ae-47c2-467c-b513-c73c0eb171a8

hung319:
To add more information, this problem occurred after I imported the JSON backup file.

diegosouzapw:
The fact that this started after a JSON backup import is useful, but the alternating successful request, 502, then success still does not distinguish a database stall from a process restart, resource pressure, or reverse-proxy timeout. Could you provide one timestamp-aligned sequence with OmniRoute/container/proxy logs, the exact endpoint and 502 body, restart count and CPU/memory, the backup size and approximate row counts, health/DB-health output, and your image tag/digest? A fresh-data-directory versus restored-backup A/B on the same image would be especially decisive. Please keep the backup; we should not recommend destructive DB recovery without proving corruption.

hung319:
My database is currently about 5MB and has 2000 pages, and my CPU and RAM are fine. I'm encountering this error with both the latest Docker and npm versions. The 502 error might be because Cloudflare misinterprets the timeout as a 502. 

[...]

PR #10948, opened 2026-08-21
feat(providers): tool calling for copilot-m365-web via router planning

acc0mplish:
Follow-up to #10732 (fixes the remaining gap from #10718: agentic clients that send `tools` got them silently dropped).

## Problem

`copilot-m365-web` ignores the OpenAI `tools` / `tool_choice` request fields entirely — every tool-calling agent (Claude Code, Codex, Cline, …) degrades to a plain text chat with no `tool_calls` in the response, so tool loops never start.

Related protocol issues found on the way:
- the server's SignalR `type:6` keepalives are never echoed back, so long agentic turns risk a mid-turn socket drop
- a `type:3` completion carrying `error` finishes as a silent empty `stop`, indistinguishable from a genuine empty reply
- `writeAtCursor` deltas riding on tool-progress frames (Progress / SearchResults / Code / ToolCall) leak into the answer text
- the prompt flattener kept only system + last user message, dropping multi-turn context, assistant `tool_calls`, and tool results

## Root cause (live-tested against a real EDU tenant)

Asking the substrate model to *use* a client-declared tool gets refused — it checks its own plugin registry and answers "not available in this chat environment", even when the invocation carries `plugins[]` + `customInstructions`. The strategy the model complies with is **router planning** (the same default the community M365 gateways converged on): a separate turn frames it as a tool-*selection* assistant that prints a routing decision as plain text (`CALL_TOOL: name({...})` / `NO_TOOL_NEEDED`), which never triggers the registry refusal.

## Changes

1. **`copilot-m365-connection.ts`** — `extractToolSpec` (tools/tool_choice → normalized specs), `flattenMessages` (full bracketed history; tool results compacted to 4000 chars), `buildPrompt` wraps the fenced-block `<tools>` protocol when tools are declared, `buildRouterPrompt` builds the router-planning turn (with a no-re-invoke rule when tool evidence is already in the history).
2. **`copilot-m365-frames.ts`** — `clientPlugins` (OpenAI tools → M365 `plugins[]` declarations), `parseToolRouterDecision` (`CALL_TOOL` lines + `{"calls":[…]}` envelope, validated against declared names and `tool_choice`), `parseFencedToolCalls` (in-turn fallback; declared-shell blocks normalized to `{command}`, undeclared names never pass through), `extractCompletionError` (type:3 error surfacing), `isToolProgressFrame` guard wired into `accumulateBotContent`/`extractBotText`.
3. **`copilot-m365-web.ts`** — router turn first; a decided call list returns OpenAI `tool_calls` with `finish_reason: "tool_calls"` (stream + non-stream). No-tool fallthrough runs the answer turn in a **fresh conversation** (reusing the router's `ConversationId` makes the answer a continuation of the routing dialog — the model keeps replying `NO_TOOL_NEEDED` as the answer), re-framed as an answer request. Also: SignalR `type:6` echo, type:3 error → `abort()`, and streaming content holdback once a fenced block opens so tool calls are never emitted as content.

## Tests

- New `tests/unit/copilot-m365-tool-calls.test.ts` (22 cases): extraction, flattening/compaction, prompt wrappers, plugin mapping, invocation passthrough + backward-compatible defaults, router decision parsing (multi-call, undeclared rejection, tool_choice enforcement, envelope fallback), fenced parsing, and the three frame guards.
- Updated the two `buildPrompt` pins in `m365-connection-4042.test.ts` for full-history flattening.
- All 99 m365 unit tests pass; `typecheck:core` clean.

## Live validation (real EDU tenant)

- non-stream tool call (`get_weather`) → `finish_reason: "tool_calls"`, `arguments: {"city": "San Francisco"}`
- streaming tool call (`read_file`) → `tool_calls` deltas + `finish_reason: "tool_calls"`
- tool-result follow-up turn → final natural-language answer, no duplicate re-invocation
- no-tools regression → unchanged plain answer


acc0mplish:
## QC failure triage — all red checks are base-red, not from this diff

I reproduced the failing gates on a clean `release/v3.8.50` checkout (`fdeac49`) and compared against this branch:

| Failing check | Also fails on base `fdeac49`? | Delta from this PR |
|---|---|---|
| Fast Quality Gates → `open-sse-typecheck` (`freebuff.ts` TS2559/TS2339/TS2698) | ✅ yes — identical output, `openSseTypecheckErrors=15` on both | **zero new errors** (I did not touch `freebuff.ts`) |
| Docs Gates (PROVIDER_REFERENCE 346→347, diagram SVGs) | ✅ yes — base shows the same class of STRICT drift (10 there, 5 here) | none — no provider added here |
| Unit Tests fast-path 1–4 (`#10225`, `#10503`, `#9199`, `#9147`, GlmExecutor, Qwen) | ✅ yes — same test names fail on the base run | none — **0 m365/copilot test failures**; the m365 suite is 99/99 green locally |
| No new ESLint warnings → `openapiCoverage.pct 38.6 < 39.2` | base run fails the sibling ratchet metric instead (`i18nUiCoverage 99.3 < 100`) | no endpoints/docs touched here |
| Stryker `tap.testFiles` (2 missing) | ✅ yes — base reports 13 missing | fewer than base |

This matches the base-red already tracked in #9985 (same situation as #10885, which was merged with the note about it). Everything this PR touches is green:

- `tests/unit/copilot-m365-*.test.ts` + `m365-*.test.ts`: **99/99 pass**
- `npm run typecheck:core`: clean
- `Vitest (fast-path)`, `semgrep`, `dast-smoke`, `Change Classification`, `Merge integrity`: all pass on this PR

Happy to rebase or help fix the base-red items separately if useful.

PR #10951, opened 2026-08-21
fix(resilience): make least-used rotate by recording the use it sorts on (#10945)

### LINKED ISSUE #10945: fix(resilience): fallbackStrategy=least-used never rotates — touchConnectionLastUsed only wired for round-robin, last_used_at stays NULL forever
## Summary

`providerStrategies[provider].fallbackStrategy = "least-used"` (and the global `settings.fallbackStrategy = "least-used"`) never actually rotates accounts on a fresh/idle connection pool — every dispatch picks the same connection — because nothing ever writes `provider_connections.last_used_at` unless the effective strategy is `"round-robin"`.

- OmniRoute version: 3.8.49
- Install method: npm global (native Windows)

## Root cause

`src/sse/services/auth.ts`, `"least-used"` branch:

```ts
} else if (strategy === "least-used") {
  // Least Used: pick the one with oldest lastUsedAt
  const sorted = [...orderedConnections].sort((a, b) => {
    if (!a.lastUsedAt && !b.lastUsedAt) return (a.priority || 999) - (b.priority || 999);
    if (!a.lastUsedAt) return -1;
    if (!b.lastUsedAt) return 1;
    return new Date(a.lastUsedAt).getTime() - new Date(b.lastUsedAt).getTime();
  });
  connection = sorted[0];
}
```

When every connection's `lastUsedAt` is null (a fresh pool, or one that has literally never been dispatched through `round-robin`), the tie-break falls through to `priority` ascending — so it deterministically picks the same lowest-priority connection on **every single call**, forever, unless something else populates `lastUsedAt`.

The only write site for `provider_connections.last_used_at` in the whole dispatch path is `touchConnectionLastUsed()` (`src/lib/db/providers.ts`), and every one of its 3 call sites in `auth.ts` (~line 1541, 1572, 1603) is inside the `strategy === "round-robin"` branch only:

```ts
// round-robin: staying with ${current.id}...
await touchConnectionLastUsed(connection.id, nextCount);
...
// round-robin: switching to LRU ${connection.id}...
await touchConnectionLastUsed(connection.id, 1);
...
// round-robin: FALLBACK MODE ... picked_lru=${connection.id}...
await touchConnectionLastUsed(connection.id, 1);
```

`least-used`, `p2c`, `reset-aware`, `fill-first`, `priority`, `cost-optimized`, etc. never call `touchConnectionLastUsed` (or any other write of `lastUsedAt`/`last_used_at`) at all. The only other write site in the codebase is `sessionAffinityPin.ts`, gated behind session affinity (`sessionAffinityTtlMs`), which is off by default (`0`) and is a distinct mechanism (session→connection pinning, not general usage tracking).

Net effect: **`least-used` silently degrades to "always pick lowest-priority connection" for any pool where accounts haven't separately been round-robin'd at some point** — the exact opposite of its documented behavior, with no error or warning surfaced anywhere.

## Repro

1. Have 2+ active connections for a provider, all with `last_used_at IS NULL` (fresh install, or a provider that's only ever been dispatched via a non-round-robin strategy).
2. Set `providerStrategies.<provider>.fallbackStrategy = "least-used"` (or global `fallbackStrategy`).
3. Send N requests through that provider.
4. Inspect `provider_connections.last_used_at` for that provider — it stays `NULL` on every row, and `x-omniroute-decision` (or connection-level logs) shows the same connection picked every time, confirmed via 5 sequential `codex-reset-aware-canary`-style test calls returning identical account selection.
5. Switch the same setting to `"round-robin"` and repeat — `last_used_at` now updates correctly (verified: 3 distinct Codex connections got real timestamps and `consecutive_use_count` values after 3 calls, matching `stickyRoundRobinLimit`).

## Suggested fix

Either:
- Call `touchConnectionLastUsed()` (or an equivalent write) from the general dispatch path regardless of strategy, whenever a connection is actually used — not just inside the `round-robin` branch, or
- Have `least-used` (and any other strategy documented as usage-aware) track its own recency signal independent of `round-robin`'s bookkeeping, or
- At minimum, document that `least-used` currently has a hard dependency on `round-robin` having been used at some point to seed `last_used_at`, since today it fails completely silently.

## Impact

Anyone who sets `fallbackStrategy: "least-used"` expecting even load distribution across multiple accounts of the same provider — the documented purpose of the strategy — gets no distribution at all on a fresh connection pool, with zero error/warning. Confirmed on a real 5-account Codex pool: `least-used` stuck on the same connection for every call; switching to `round-robin` fixed it immediately.

ntdat812:
⚠️ base-red inherited: #9985

Fixes #10945.

## Problem

`least-used` sorts connections by `lastUsedAt`, but it was the only `lastUsedAt`-reading strategy that never wrote it — `planLastUsedCommit()` is called from the `round-robin` branch only (three sites) and from `sessionAffinityPin.ts`, which is gated behind `sessionAffinityTtlMs` (off by default).

So on a pool where every `last_used_at` is still `NULL`, the comparator's own tie-break

```ts
if (!a.lastUsedAt && !b.lastUsedAt) return (a.priority || 999) - (b.priority || 999);
```

decides every call, and it is deterministic. The same connection comes back forever, with no error and no warning.

Reproduced on a fresh 3-account pool (`glm`, apikey, priorities 1/2/3, all `last_used_at NULL`), three sequential `getProviderCredentials()` calls:

| strategy | picked | `last_used_at` after |
| --- | --- | --- |
| `least-used` | `[0, 0, 0]` | `[null, null, null]` |
| `round-robin` | `[0, 1, 2]` | `[ts, ts, ts]` |

The reporter's analysis is accurate; the line numbers in the issue are from 3.8.49, and the branch has since been refactored to `planLastUsedCommit()`, but the defect is unchanged on `release/v3.8.50`.

## Change

Nine lines: the `least-used` branch now commits the use the same way `round-robin` does, including the `options.lease` deferral so a leased selection still commits through `commitSelectionSideEffects` rather than writing early.

```ts
connection = sorted[0];
const commit = planLastUsedCommit(connection, connectionsRaw, 1);
if (options.lease) commitSelectionSideEffects = commit;
else await commit();
```

`consecutiveUseCount` is set to `1` — the sticky counter belongs to round-robin's `stickyRoundRobinLimit` and `least-used` has no stickiness concept.

**Scope is deliberately `least-used` only.** The issue suggests writing `last_used_at` from the general dispatch path for every strategy; I did not do that. `p2c`, `random`, `cost-optimized` and `fill-first` never consult `lastUsedAt`, so writing it for them would add a DB write per dispatch without changing a single selection. `least-used` is the one strategy whose own comparator reads the field it was not writing — that is a self-contained defect with a self-contained fix. If you would rather have the general write, say so and I will move it.

## Verification

`tests/unit/least-used-rotation-10945.test.ts` (new, 4 cases). Reverting only `src/sse/services/auth.ts`:

```
not ok - #10945 least-used rotates across a fresh pool instead of pinning one account
not ok - #10945 least-used persists last_used_at so the next process keeps rotating
# tests 4  # pass 2  # fail 2
```

All 4 pass with the change. The 2 that pass either way are the must-not-regress guards: an existing `lastUsedAt` ordering is still honoured (seeded inverse to `priority`, so a fix that merely fell back to priority order would fail it), and a single-account pool still resolves that account.

**Regression sweep.** All 67 unit test files that reference `getProviderCredentials` or `sse/services/auth` — 699 tests. Comparing *failing test names* (not counts) against the same run with only `auth.ts` reverted:

```
base   23 failing
branch 21 failing
tests failing on the branch but not on base: none
```

`typecheck:core` exit 0, `eslint` clean on both files, `check:changelog-integrity` OK.

One note on that sweep: an intermediate run showed `handleChat aborts the pending cooldown wait when the client disconnects` failing on the branch. It is flaky, not a regression — that test never sets `fallbackStrategy`, so it runs the default `fill-first` branch which this change cannot reach; it passes 3/3 in isolation and did not recur on the repeat full run. The suite's failure count moves between 21 and 23 run to run on this base regardless of the change.


ntdat812:
CI ran the four new cases on Linux and all four pass (shard 4/4):

```
✔ #10945 least-used rotates across a fresh pool instead of pinning one account (361.7ms)
✔ #10945 least-used persists last_used_at so the next process keeps rotating (301.4ms)
✔ #10945 least-used still honours an existing lastUsedAt ordering (268.5ms)
✔ #10945 a single-account pool keeps resolving that account (348.9ms)
```

So the local result reproduces on CI, not just on Windows.

The shard still reports red, on the same base-red set documented in #10935: `check-docs-counts-sync` (5 STRICT provider-count drifts present on base HEAD with no PR changes applied) and the `openapiCoverage.pct: 38.6 < baseline 39.2` ratchet, which base itself produces — this PR adds zero `route.ts` files.

Issue #10871, opened 2026-08-20
fix(providers): Client request protocol in Compatibility settings is not persisted per-model


zoser69:
## Summary

On the model/provider **Compatibility** settings, changing the **Client request protocol** (e.g. `openai` / `openai-responses` / `anthropic messages`) is **not persisted** after a page refresh. Every time the server or UI reloads, the protocol per-model selection resets, so it is impossible to permanently assign which wire protocol each model uses (Responses API vs Chat Completions vs Anthropic Messages).

## Environment

- OmniRoute Version: 3.8.50 (next)
- Installation Method: npm (global) / local server
- Operating System: Windows 11

## Steps to Reproduce

1. Open a provider (e.g. OpenCode / `opencode`) in the UI.
2. Go to the model's **Compatibility** settings.
3. Change the **Client request protocol** for a model from the default to, say, `openai-responses` (or `anthropic messages`).
4. Save / apply the change. The model works correctly for the current session.
5. Refresh the page (or restart the server).
6. Re-open the same model's Compatibility settings → the **Client request protocol** has reverted to the default. The per-model choice was not saved.

## Expected Behavior

The chosen Client request protocol should be stored (in the provider/connection config or model metadata) and survive a refresh/restart, the same way other per-model compatibility flags (e.g. `supportsReasoning`, `supportsVision`, context length) persist.

## Why this matters

Without persistence, an operator **cannot designate per-model** which endpoint a model speaks — for example, OpenCode Zen models are split across three different protocols:

- `/v1/responses` (e.g. `gpt-5.6-sol`, `muse-spark-1.2`, `grok-4.5`) → `openai-responses`
- `/v1/messages` (e.g. `claude-opus-5`, `qwen3.7-max`) → `anthropic messages`
- `/v1/chat/completions` (e.g. `deepseek-v4-pro`, `minimax-m3`, `big-pickle`) → `openai`

Today the only durable way to pin a model to the correct protocol is to edit the provider registry source (`open-sse/config/providers/registry/...`) and add `targetFormat` to the model entry. The UI selector exists but does not survive, forcing everyone to patch source instead of using the intended Compatibility UI.

## Additional context

Related: #10867 — Muse Spark 1.2 returns `null` content because it is silently sent to `/chat/completions` instead of its real `/responses` endpoint. The root cause there is the same gap: there is no persisted, per-model protocol assignment through the UI, so these models fall through to the default chat/completions path.

Issue #10865, opened 2026-08-20
[BUG] Custom System Prompt toggle on Endpoint page fails to persist after page refresh (v3.8.49)


helays:
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Description

**Describe the bug**
On the OmniRoute dashboard, within the OMNIPROXY module's Endpoint page, enabling the "Custom System Prompt" toggle and saving shows a request being sent. However, after refreshing the page, the toggle reverts to the disabled state. The custom prompt content also fails to persist.

**This issue is reproducible on version `3.8.49`.**

### Steps to Reproduce

1.  Navigate to the OMNIPROXY module and open the Endpoint page.
2.  Locate the "Custom System Prompt" toggle and enable it.
3.  Enter any text in the prompt content field (e.g., "test prompt").
4.  Observe the network request (`PATCH` to an API endpoint like `/api/settings/...`) appears to be sent successfully (e.g., status 200).
5.  Refresh the browser page.
6.  Observe that the "Custom System Prompt" toggle is now disabled and the prompt text field is empty.

### Expected Behavior

The toggle should remain enabled, and the prompt text should be preserved after page refresh, indicating the configuration was persisted to the database and reloaded into memory on page load.

### Actual Behavior

The setting reverts to the default (disabled) state after refresh, even though the API request appears to be sent without a client-side error.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

Issue #10863, opened 2026-08-20
# [BUG] Search endpoints (MCP omniroute_web_search + /v1/search) still expose providers blocked via Settings → Security → Blocked Providers, and list providers with no credentials


kanishk619:
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Description

This is in continuation to the previously closed #10159 

The `omniroute_web_search` MCP tool and the `/v1/search` API **ignore the Blocked Providers security setting** and expose providers the operator has explicitly blocked.
Reproduction of the exact gap:
1. In **Settings → Security → Blocked Providers**, add a search provider (e.g. `brave-search` / `jina-search`). Confirm it appears in the blocked list.
2. The MCP `provider` enum still lists it as an available option (15 providers advertised on the `next` tag).
3. Calling it still attempts to use it — and for providers with **no API key configured**, returns `400 No credentials configured for search provider: <id>`.
So the blocked setting has **no effect on the search surfaces**. Two distinct problems combine here:
- **A. The Blocked Providers setting is not applied to search providers.** The search code path (`open-sse/handlers/search.ts`, `open-sse/config/searchRegistry.ts`, `open-sse/mcp-server/schemas/tools.ts`, `open-sse/services/webSearchRouting.ts`) has **no reference to any blocked/deny-list**. The Blocked Providers security page is wired to the chat/completion provider routing, not the search subsystem — so blocking a search provider there does nothing.
- **B. The provider list is never filtered by configured credentials either.** The MCP `provider` enum is driven by the static search registry, which lists every supported provider regardless of whether an API key exists. On the `next` tag the enum lists 15 providers; only 6 have keys configured (serper, exa, tavily, firecrawl, ollama, duckduckgo). The other 8 are advertised but guaranteed to 400 at call time.

### Steps to Reproduce

**Setup:**
1. Open **Settings → Security → Blocked Providers** and add `brave-search` (and optionally `jina-search`). Save.
2. Restart OmniRoute (to rule out caching).
**MCP surface:**
3. Connect via any MCP client (Hermes, Claude Desktop, Cursor, etc.).
4. Call `tools/list` and inspect the `provider` enum of `omniroute_web_search` → **`brave-search` is still listed** even though it is blocked:
   ```text
   serper-search, brave-search, perplexity-search, exa-search, tavily-search, firecrawl,
   google-pse-search, linkup-search, searchapi-search, youcom-search, searxng-search,
[...]

Issue #10856, opened 2026-08-20
fix(auth): unresolved model id on /v1/messages returns misleading "[401] Missing API key" instead of model-not-found


martelozzo83:
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 Home

### Provider(s) Involved

N/A — reproduces regardless of provider, happens before routing/model resolution

### Client Tool

Claude Code CLI 2.1.237 (also reproduces with plain curl, not CLI-specific)

### Description

When a request to `/v1/messages` specifies a `model` id that OmniRoute's catalog doesn't recognize (e.g. Claude Code's own default `claude-sonnet-4-5`, which isn't a raw OmniRoute catalog/combo id), the endpoint responds with:

```json
{"error":{"message":"[401]: Missing API key.","type":"authentication_error","code":"invalid_api_key"}}
```

This is the exact same error shape returned for a genuinely missing/invalid API key. The API key in these repro requests is valid and confirmed working (same key succeeds against `/v1/models` and against `/v1/messages` with a real catalog model id).

This is misleading: it reads as an auth failure, sending debugging in the wrong direction, when the actual problem is model resolution.

### Steps to Reproduce

1. Get a valid OmniRoute API key (Dashboard → API Manager).
2. Confirm it works: `curl http://localhost:20128/v1/models -H "x-api-key: <key>"` → 200, lists real catalog ids like `auto/best-coding`.
3. Send a message with an id that isn't in that catalog:

```bash
curl -s -X POST http://localhost:20128/v1/messages \
  -H "x-api-key: <valid key>" \
  -H "anthropic-version: 2023-06-01" \
  -H "Content-Type: application/json" \
  -d '{"model":"claude-sonnet-4-5","max_tokens":16,"messages":[{"role":"user","content":"reply OK"}]}'
```

4. Compare against the same request with a real catalog id (e.g. `"model":"auto/best-coding"`) → succeeds normally with a real completion.

### Expected Behavior

An unresolved/unknown `model` id should return a clear `model_not_found` (or similar) error distinct from authentication failures — same shape/spirit as OpenAI's `"The model `X` does not exist"` — so client-side debugging isn't sent chasing an auth problem that doesn't exist.

### Actual Behavior

Returns `401 invalid_api_key` / "Missing API key" for a valid key + unresolved model, identical to the error for an actually missing key.

### Test Impact

[...]

PR #10952, opened 2026-08-21
Integration: security tier + self-hosted operator blockers (rebased onto v3.8.51)

arminanton:
## Summary

An integration branch of blocker + security fixes, rebased clean onto the current `release/v3.8.50` tip (13 commits, 0 behind). It bundles the security tier and the self-hosted operator bugs from a prioritised remediation plan, plus one build-reliability fix and one net-new secret-leak hardening pass found during review.

The branch was rebased onto current head immediately before opening this PR; commits that upstream has since merged independently (its own #10907 sticky-pin fix, #10872 CLI-card guard, the CodeQL-action bumps, the command-code effort suffixes) were **dropped** during the rebase rather than reapplied, so this diff is a true delta with no duplicate work.

## What's in it, by tier

**Security (highest severity first)**
- **fix(relay): resolve `x-relay-path` through the shared SSRF guard in the Cloudflare worker** (#10935). Attacker-controlled `x-relay-path` was concatenated onto a validated target, re-pointing the request past the checked host (`/x@evil.com`, `\evil.com`, `//evil.com/x`). Routed through the shared guard + regression test, mirroring the merged Deno/Vercel fixes (#4643).
- **fix(security): clear CodeQL `js/insufficient-password-hash` + incomplete-url-substring alerts** (#10888). Small, closes active code-scanning alerts.
- **fix(search): enforce Blocked Providers on `/v1/search` and the `omniroute_web_search` MCP tool** (#10863). Blocked providers were still advertised and reachable through the search endpoint, a security-setting bypass.
- **fix(security): refuse credential-echoing upstream error passthrough + redact relayed OCR/moderation errors + widen call-log key redaction** (net-new, from an audit of the error/log surface). `upstreamErrorPassthrough` relayed a provider's 4xx body verbatim to Claude-Code clients screened only for stack traces; it now refuses passthrough when the body carries a credential pattern (ReDoS-safe regex, verified linear) and falls back to the sanitised path, preserving the capability-recovery contract for secret-free bodies. OCR/moderations handlers now run upstream error text through a structure-preserving redactor. `protectPayloadForLog` gains `cookie`/`storageState`/`runtimeKey`/`capability` so web-impersonation credentials that land in a body field are redacted before the on-disk call-log artifact.

**Bugs that block real usage (self-hosted operator)**
- **fix(db): fail-closed call-log rotation fence on SQLite corruption** (#10736). `rotateCallLogs()` swallowed `SQLITE_CORRUPT` and only logged it, so rotation kept hammering a malformed pager every ~60s while `/api/db/health` stayed green. It now fences further rotation on corruption and surfaces a degraded health signal. Root-cause-agnostic per the maintainer's scoping on the issue: no repair is attempted (a live REINDEX/VACUUM on a single-writer corrupt file is worse than the failure).
- **fix(build): guard better-sqlite3 native addon out of the Next.js build** (#10060). The native Statement destructor SIGABRTs at build-worker teardown (`node::RemoveEnvironmentCleanupHook`). Every DB entry point now keys off a reliable `OMNIROUTE_BUILDING=1` signal (Next workers sometimes drop `NEXT_PHASE`) so the addon never loads in a build worker; raises build heap to 6144 and caps worker pools to avoid the many-core SIGSEGV; adds `.gitattributes` LF normalisation. **Deliberately does NOT downgrade the Node base image** per the maintainer's review on #10060: `release/v3.8.50` moved to `node:26-trixie-slim` through several considered commits, so the guard is re-derived against the current base rather than reverting the FROM line. The original PR's `npm@11` pin and binary-hide dance are dropped because this build already rebuilds better-sqlite3 deterministically via `node-gyp` and floats `npm@latest` for the CVE overlay.
- **fix(providers): stop the loopback readiness gate from memorizing failure** (#10903). One failed probe at boot pinned the process into in-process fallback permanently and spammed logs (270 lines in 26h). Cache cleared on rejection; warning throttled.
- **fix(settings): persist customSystemPrompt fields in updateSettingsSchema** (#10865). The Custom System Prompt toggle didn't persist because the Zod schema dropped the fields.
- **fix(model): return model_not_found for prefix-inferred models when the provider is inactive** (#10856). A misleading `401 Missing API key` appeared instead of `model not found` when a model routed to an unconfigured provider.

**Provider/feature fixes**
- **fix(auth): add opencode/opencode-zen credential search pair** (#10899). An auto-combo target resolving to `opencode-zen` searched only that id and dropped the target even though the key sits on the no-auth `opencode` row. Adds the pair (same operator account, requested id first, mirroring nvidia/nvidia_nim and antigravity/agy). Original author: @excessivechaos-adjacent contributor; ported here with the fix isolated from unrelated base-churn.
- **fix(models): persist apiFormat, targetFormat, and supportsVision overrides for catalog models** (#10871). A `PUT /api/provider-models` carrying only these fields for a non-custom catalog model fell through to 404. Now persisted and resolved through the custom → compat-override → synced ladder. Alias-awareness is applied at the resolution layer (via the existing no-auth sibling map) rather than in the low-level compat key, so paths that legitimately key on the raw provider id (e.g. `getHiddenModelsByProvider`, #7620) keep working. Original author: @rqzbeh.
- **fix(executors): report WS readyState in Meta AI timeout error** (#10916 / #10727). Adds `ws.readyState` to the muse-spark-web timeout error so "never opened" (0) can be told apart from "opened then silent" (1). Diagnostics-only, leaks nothing sensitive (a 0-3 enum). Original author: contributor via #10916.

## Contributor credit

Several commits are cherry-picked from external contributor PRs (authorship preserved in git history): @rqzbeh (#10871), the #10899 and #10916 authors, and the sticky-pin / command-code work by @excessivechaos where upstream hadn't already merged it. Where upstream merged an equivalent fix first, that version is kept and the duplicate dropped.

## Validation

- `typecheck:core` clean, `check:cycles` clean.
- 68 unit tests across the new/ported fixes pass (build-guard, corruption-fence, opencode-pair, WS-diagnostics, model-persistence, hidden-model #7620, error-passthrough credential-refuse, sticky-pin, SSRF, blocked-providers, custom-system-prompt).
- A broad sweep of the touched areas (combo/passthrough/redact/sanitize/build, ~2.1k tests) shows **zero regressions vs the pristine `release/v3.8.50` baseline**: the same pre-existing failures appear on both, and no new failure is introduced by this branch.

## Notes for reviewers (transparency, not changed here)

- **Pre-existing base-reds observed** on the current release tip (not introduced or fixed here): a `better-sqlite3` native-binding load failure on some host glibc lines (affects the `db/omp`, migration-snapshot and dependency-shape tests in a sandbox), and two `model-capability-overrides` / `#9199` deep-equal assertions. These fail identically on pristine `release/v3.8.50`.
- **Config-posture finding (left for a maintainer decision, not changed):** anonymous `/v1` access is the shipped default unless `REQUIRE_API_KEY` is set, and configuring a management password does not gate the API surface (only the dashboard). On a `0.0.0.0`-bound instance an unauthenticated caller can use the operator's stored provider keys. Flipping a security default belongs with the maintainer rather than in an integration branch; flagging it here.
- **Optional hardening:** `NEXT_TELEMETRY_DISABLED=1` is set in a sub-Dockerfile but not the main build; setting it in the primary build env would fully disable Next.js build-time telemetry. (No runtime telemetry / phone-home exists; the audit confirmed the "zero telemetry by default" posture.)


diegosouzapw:
Thank you for bundling these blocking/security fixes — the substantive code is solid and the dedicated tests are well-written. However, the PR is currently stale relative to its stated base and cannot merge as-is.

MANDATORY before this can merge:

1. **The rebase is not current.** The branch is 54 commits behind `release/v3.8.50`. A real merge produces 7 conflicts, and worse, `src/shared/validation/settingsSchemas.ts` auto-merges with DUPLICATE keys: your `customSystemPromptEnabled`/`customSystemPrompt` (max 100000) at ~line 151 PLUS the base's identical pair (max 10000) at ~line 195 in the same `z.object` literal → `TS1117 duplicate object key` → `typecheck:core` fails. We verified this by materializing the merge. Please rebase onto the current tip and resolve.

2. **Drop 4 already-landed areas (subsumed by base).** These are already on `release/v3.8.50` and will re-conflict:
   - `#10935` Cloudflare SSRF: the summit fix + `tests/unit/cloudflare-relay-path-ssrf.test.ts` already exist on base; your inline `isPrivateHostname(h)` also diverges from base's shared `isPrivateRelayHostname.toString()` — keep base's, drop the divergent inline copy.
   - `#10727` muse-spark WS `readyState` diagnostic + its test (already on base).
   - `#10865`/`#10890` custom system prompt schema (base commit `9fd19f5` already landed it) — drop your insertion and the `custom-system-prompt-settings-schema.test.ts` file.
   - `cursorApiKeyAuth` HMAC (base already uses `createHmac`).

3. **Add a regression test for the `#10898` change (Hard Rule #18).** `src/lib/db/models/compat.ts` removes alias canonicalization from the compat key and `src/sse/services/model.ts:312` merges overrides across no-auth siblings — both are production behavior changes with NO test in this PR. Add one pinning that `opencode-zen` resolves overrides stored under sibling `opencode`.

OPTIONAL: `src/lib/logPayloads.ts` adds `"capability"` to `SENSITIVE_KEYS` — that word is very common in non-secret bodies and will broadly redact `capability` fields from call-log artifacts; consider scoping to provider-specific keys only. Also the 3-condition `isBuild` expression is repeated across `driverFactory.ts`, `core.ts`, `kiro` route and `codegraphKnowledge.ts` — a tiny shared helper would keep them consistent.

No behavior issues in the genuinely-new code: the build-sqlite stub, call-log corruption fence, loopback re-probe and secret/error hardening all look correct and their tests pass. Once rebased, dropped and re-tested this will be a strong merge.

arminanton:
Thank you for the thorough review and the clear, actionable guidance — it made this straightforward to resolve. I've force-pushed a rebased branch (now `64192d32a`, 10 commits, linear on the current `release/v3.8.50` tip `6cd4d38e2`). Here's everything that was done, mapped to your points.

## Mandatory

**1. Rebase + the silent `settingsSchemas.ts` duplicate key.**
Rebased onto the current tip. You were exactly right that a naive rebase auto-merges `customSystemPromptEnabled`/`customSystemPrompt` into a `TS1117` duplicate-key state that breaks `typecheck:core` — I reproduced it by materializing the merge. Resolved by dropping our now-subsumed copy (see #2). `typecheck:core` is green on the rebased branch.

**2. Drop the already-landed areas.** Dropped all four you named, and while rebasing I found **two more** that base had absorbed since your review, so they're dropped too:
- `#10935` Cloudflare relay-path SSRF (base already has the shared `isPrivateRelayHostname` + the test) — dropped our divergent inline copy.
- `#10727` muse-spark WS `readyState` diagnostic + test.
- `#10865`/`#10890` custom system prompt schema (base commit landed it) — dropped our insertion and `custom-system-prompt-settings-schema.test.ts` (this is what removes the duplicate key).
- `cursorApiKeyAuth` HMAC — base already uses `createHmac`.
- **Also dropped (newly subsumed):** `#10903` loopback-readiness no-memoize (base's anti-storm implementation is a superset of ours) and `#10736` call-log corruption fence (base grew its own `callLogRotatePaused`/`handleCallLogRotateError`).

**3. Regression test for `#10898` (Hard Rule #18).** Added `tests/unit/noauth-sibling-compat-override-7620.test.ts`, pinning both halves of the fix: (a) an override written under the raw `opencode` id stays on the raw key and does **not** leak onto `opencode-zen` (the exact `#10898` regression), and (b) `getNoAuthHydrationProviderIds` merges the sibling so a lookup resolving to `opencode-zen` still reads the `opencode` override. I verified it's a real guard by re-introducing the canonicalized key — the test flips to `not ok`, and reverting to the raw-key fix makes it pass.

## Optional (both done)

**`logPayloads.ts` `"capability"` redaction.** Agreed — matching the bare word broadly redacts non-secret `capability`/`capabilities` fields (model catalogs, degradation/provider-discovery strings, MCP tool schemas). Removed it; the real web-impersonation credentials (`ecto_1_sess` cookie, `storageState`, `runtimeKey`) are already covered by `cookie`/`authorization`/`storageState`/`runtimeKey`. Added a regression test asserting those redact while `capability`/`capabilities` survive.

**Shared `isBuild` helper.** Extracted `src/lib/buildPhase.ts` (`isNextBuildPhase()`) and routed all four sites through it — `driverFactory.ts`, `codegraphKnowledge.ts`, the kiro auto-import route, and `db/core.ts`'s `isBuildPhase` const now delegate to it. No import cycle (the leaf helper resolves the `driverFactory`↔`core` cycle that forced the inline copies).

## Verification
`typecheck:core` green; 131 tests pass across the changed surface (copilot/github, compat-override, build-stub, log-payloads, error-passthrough). Branch is linear (0 merge commits) and 0 behind the default branch.

Thanks again for the detailed problem list — the file:line pointers and the "materialized the merge" note on the duplicate key saved a lot of time. Happy to adjust anything further.

arminanton:
Refreshed the branch onto the current `release/v3.8.50` tip and force-pushed (now `d7c3d66f2`, 11 commits). The previous head was rebased onto an older tip, so its CI went red on changes that landed upstream since (e.g. the `webhook-telegram-dispatcher` `emoji` change) — those pass now that the branch sits on the current tip.

One net-new commit was added, completing the model-discovery de-gating that `fix(copilot): match live CLI 1.0.81-6 wire identity + de-gate model discovery` started:

**`fix(copilot): present raw GitHub token on model discovery so entitled models unlock`** — the earlier commit de-gated the *parse filter* (capability-driven instead of a hardcoded allowlist), but discovery still presented the **exchanged** `copilot_internal/v2/token` bearer on `GET /models`. That exchanged token is minted without the `copilot-developer-cli` integration identity, so GitHub serves only the narrower default catalog and entitled `/responses`-only models (grok-4.x, mai-code) silently vanish for accounts that DO have them (verified: the same account shows them in the Copilot CLI). The full catalog only unlocks when `copilot-integration-id: copilot-developer-cli` rides on a **raw** GitHub Bearer. Fix reorders the discovery token precedence (github.com path) to prefer the raw `accessToken`, keeping the exchanged token as a fallback; GHE keeps the exchanged token since its enterprise API requires it. Adds `tests/unit/github-copilot-discovery-token.test.ts` pinning that the raw token is sent as `Authorization: Bearer` with the developer-cli header, and that grok/mai survive discovery parsing.

Happy to split that copilot-discovery commit into its own PR if you would rather keep this one exactly as previously reviewed — it is self-contained.

Issue #7302, opened 2026-07-15
fix(startup): CLI data-dir migration from Electron (%APPDATA%\omniroute) breaks storage.sqlite / server.env compatibility


Flexible78:
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Windows

## Environment
- OmniRoute Version: 3.8.48 (CLI via npm)
- OS: Windows 11 IoT LTSC 24H2
- Node.js: v24.13.1

## Problem
After uninstalling Electron v3.8.48 and installing CLI version via `npm install -g omniroute@3.8.48`:

1. CLI uses `%USERPROFILE%\.omniroute\` as data directory
2. Electron used `%APPDATA%\omniroute\` as data directory
3. Copied `storage.sqlite` (20 MB, 30 providers, 50 call logs) and `server.env` (with encryption keys) from `%APPDATA%\omniroute\` to `%USERPROFILE%\.omniroute\`
4. Server starts successfully ("✅ OmniRoute is running!")
5. Dashboard shows "Internal Server Error" on all pages

## What I tried
- Copied `server.env` with encryption keys
- Ran `PRAGMA wal_checkpoint(TRUNCATE)` on the database
- Used `sqlite3 .recover` to create clean copy
- Database has valid data: `SELECT COUNT(*) FROM provider_connections` returns 30

## Suspected cause
The encryption keys in `server.env` from Electron installation may not be compatible with CLI, or there are additional configuration files needed for migration.

## Additional data
- `storage.sqlite`: 20,717,568 bytes
- Tables present: 60+ tables with data
- Both paths load env: `Loaded env from C:\Users\Alexander\.omniroute\.env` and `Loaded env from C:\Users\Alexander\AppData\Roaming\npm\node_modules\omniroute\.env`

## Environment
- OmniRoute Version: 3.8.48 (CLI via npm)
- OS: Windows 11 IoT LTSC 24H2
- Node.js: v24.13.1

## Problem
After uninstalling Electron v3.8.48 and installing CLI version via `npm install -g omniroute@3.8.48`:

1. CLI uses `%USERPROFILE%\.omniroute\` as data directory
2. Electron used `%APPDATA%\omniroute\` as data directory
3. Copied `storage.sqlite` (20 MB, 30 providers, 50 call logs) and `server.env` (with encryption keys) from `%APPDATA%\omniroute\` to `%USERPROFILE%\.omniroute\`
[...]

PR #10958, opened 2026-08-21
fix(desktop): pin the NSIS artifact name so the Windows updater stops 404ing (#10947)

### LINKED ISSUE #10947: [Bug] Windows auto-update 404 due to incorrect installer filename
## Bug: Windows in-app auto-update requests the wrong installer filename and returns 404

**App:** OmniRoute Desktop  
**Version:** v3.8.49  
**OS:** Windows

### What happened

The in-app updater detects v3.8.49, but the download fails with HTTP 404:

```text
Auto-update failed: Cannot download "https://github.com/diegosouzapw/OmniRoute/releases/download/v3.8.49/OmniRoute-Setup-3.8.49.exe", status 404
```

### Expected

The updater should request the Windows installer asset published in the v3.8.49 release:

```text
https://github.com/diegosouzapw/OmniRoute/releases/download/v3.8.49/OmniRoute.Setup.3.8.49.exe
```

### Actual

The updater requests:

```text
OmniRoute-Setup-3.8.49.exe
```

The release asset is named:

```text
OmniRoute.Setup.3.8.49.exe
```

There are two filename separator mismatches:

- `OmniRoute-Setup` vs `OmniRoute.Setup`
- `Setup-3.8.49` vs `Setup.3.8.49`

GitHub therefore returns 404. The manual **Download EXE (Windows)** link can still be used as a workaround.

### Likely cause

The Windows updater URL is being constructed from a filename pattern that does not match the artifact uploaded by the release pipeline. This may involve the Electron updater configuration, `electron-builder` `artifactName`, or the generated `latest.yml` metadata.

### Suggested fix

Please align the Windows artifact name and updater metadata so that all of the following agree:

- `electron-builder` `artifactName`
- the uploaded GitHub Release asset
- the Windows `latest.yml` `path` field
- the URL constructed by the in-app updater

The expected filename for v3.8.49 is:

```text
OmniRoute.Setup.3.8.49.exe
```

Thanks!

ntdat812:
⚠️ base-red inherited: #9985

Fixes #10947.

## Problem

GitHub rewrites spaces in an uploaded asset name to `.`. electron-builder writes that same artifact name into `latest.yml` with `-`. NSIS carries the one default artifact name that contains spaces — `${productName} Setup ${version}.${ext}` — so the manifest and the published asset can never agree, and the in-app updater 404s on every release.

Measured on the **published v3.8.49 release**, not inferred:

| manifest | `url` | asset actually on the release |
| --- | --- | --- |
| `latest.yml` | `OmniRoute-Setup-3.8.49.exe` | `OmniRoute.Setup.3.8.49.exe` ❌ |
| `latest-linux.yml` | `OmniRoute-3.8.49.AppImage` | `OmniRoute-3.8.49.AppImage` ✅ |
| `latest-linux.yml` | `omniroute-desktop_3.8.49_amd64.deb` | `omniroute-desktop_3.8.49_amd64.deb` ✅ |
| `latest-mac.yml` | `OmniRoute-3.8.49.dmg` | `OmniRoute-3.8.49.dmg` ✅ |

The Windows entry's `size: 340441395` matches the uploaded `OmniRoute.Setup.3.8.49.exe` byte for byte — same file, two spellings. Linux and macOS are unaffected because the AppImage, deb and dmg defaults are already hyphen/underscore separated.

That closes the chain end to end: electron-builder emits `OmniRoute Setup 3.8.49.exe` locally → writes `-` into the manifest → `softprops/action-gh-release` uploads it → GitHub stores `.` → `electron-updater` requests the `-` name → 404.

## Change

One line — an explicit `nsis.artifactName`.

```json
"nsis": {
  "artifactName": "${productName}.Setup.${version}.${ext}",
```

**Pinned to the dot form on purpose, not a hyphen one.** A hyphen name would also fix the updater, but it would rename the published asset, and three places hardcode the current name:

- `src/app/(dashboard)/dashboard/HomePageClient.tsx:151` — the manual **Download EXE (Windows)** button, which is the workaround the issue reports as still working
- `docs/guides/TROUBLESHOOTING.md` and its `zh-TW` / `pl` mirrors — the sha512 verification snippet
- `electron/README.md`

With the dot form the asset name is **byte-identical to what ships today**; only `latest.yml` changes, from the wrong name to the right one. Nothing that currently works breaks.

`docs/guides/ELECTRON_GUIDE.md` (and its `pl` mirror) listed the local build output as `OmniRoute Setup X.Y.Z.exe`; that line is now correct for the new name.

## Verification, and its limit

**Stated plainly: I could not run the release pipeline.** Producing the artifact needs a full electron-builder Windows build plus a GitHub upload, so the fix is verified by the naming chain above rather than by a built installer. What makes it conclusive is that `latest.yml` in v3.8.49 *already contains* `OmniRoute-Setup-3.8.49.exe` — that string is the manifest writer's output for a space-containing name. Removing the spaces makes the file name and the manifest name the same string by construction, and the manifest value stops changing under GitHub's rewrite.

What I did verify, `tests/unit/electron-artifact-name-10947.test.ts` (new, 4 cases):

- `nsis.artifactName` is set and contains no whitespace, with the reason in the assertion message
- it still contains `Setup` — the release workflow picks the portable exe via `case "$file" in *Setup*) continue ;;`, so renaming it away from `Setup` would silently publish the installer as `OmniRoute.exe`
- **the expanded installer name equals the filename `HomePageClient.tsx` builds its download URL from** — two independent sources of truth for the same string, which is exactly the invariant that was broken
- no configured artifact name anywhere in the build config contains whitespace

Reverting only `electron/package.json`: **3 of 4 fail**. All 4 pass with the change. The 4th passes either way — no other target has a whitespace name today, and it is there to keep it that way.

**Regression sweep:** all 33 unit test files matching `electron` — 282 tests. Comparing *failing test names* against the same run with the config reverted:

```
base   12 failing
branch  9 failing
tests failing on the branch but not on base: none
```

(The 9 shared failures are pre-existing Windows/symlink environment ones — `materializeBundledSymlinks`, `resolveServerEntry`, `#7302 DATA_DIR/server.env`.)

`typecheck:core` exit 0, `eslint` clean, `check:changelog-integrity` OK.

The next release is the real test. If you would rather verify before merging, a `--win` build alone is enough — the name appears in `electron/dist-electron/` without publishing anything.


ntdat812:
CI ran the four new cases on Linux and all four pass (shard 3/4):

```
✔ #10947 the NSIS installer name is set explicitly and carries no space (1.0ms)
✔ #10947 the NSIS name still contains 'Setup' (0.3ms)
✔ #10947 the built installer name is the one the dashboard links to (0.8ms)
✔ #10947 no configured artifact name anywhere contains whitespace (0.2ms)
```

The shard failures are the same base-red set documented on #10935 — `check-docs-counts-sync` (5 STRICT provider-count drifts present on base HEAD with nothing applied) and the `openapiCoverage.pct: 38.6 < baseline 39.2` ratchet, which base itself produces. This PR touches no `route.ts` and no provider registry.

`Merge integrity`, `Change Classification`, `Vitest`, `Build`, `semgrep` and `dast-smoke` all pass.

PR #10960, opened 2026-08-21
fix: resolve CLI mcp call protocol issues (BUG-001)

YunyunZhai:
## Summary

- Fix `mcp status` showing 'stopped' when server is online (check `online`/`running` fields)
- Replace broken `/api/mcp/tools/call` route with proper JSON-RPC 2.0 protocol
- Implement `mcpJsonRpcCall()` using initialize + tools/call via `/api/mcp/stream`
- Support both stream and non-stream modes
- Add `Mcp-Session-Id` header management

## Changes

- Modified `bin/cli/commands/mcp.mjs` to use new JSON-RPC client
- Added `bin/cli/mcpClient.mjs` with shared MCP client utilities
- Updated tests with `mcpStreamMock` helper
- Fixed ESLint issues in test files

## Test Plan

- [ ] `omniroute mcp status` now correctly shows online status
- [ ] `omniroute mcp call <tool>` works with JSON-RPC protocol
- [ ] `omniroute mcp call <tool> --stream` works with SSE streaming
- [ ] All existing tests pass

### DIFF
--- a/bin/cli/commands/combo.mjs
+++ b/bin/cli/commands/combo.mjs
@@ -3,6 +3,7 @@ import { printHeading } from "../io.mjs";
 import { withRuntime } from "../runtime.mjs";
 import { t } from "../i18n.mjs";
 import { apiFetch } from "../api.mjs";
+import { mcpCallTool } from "../mcpClient.mjs";
 import { emit } from "../output.mjs";
 
 const VALID_STRATEGIES = [
@@ -62,15 +63,7 @@ export function extendComboSuggest(combo) {
         weights: opts.weights ? JSON.parse(opts.weights) : undefined,
         top: opts.top,
       };
-      const res = await apiFetch("/api/mcp/tools/call", {
-        method: "POST",
-        body: { name: "omniroute_best_combo_for_task", arguments: body },
-      });
-      if (!res.ok) {
-        process.stderr.write(`Error: ${res.status}\n`);
-        process.exit(1);
-      }
-      const data = await res.json();
+      const data = await mcpCallTool("omniroute_best_combo_for_task", body);
       const candidates = data.candidates ?? data;
       const rows = (Array.isArray(candidates) ? candidates : []).map((c, i) => ({
         rank: i + 1,
--- a/bin/cli/commands/compression.mjs
+++ b/bin/cli/commands/compression.mjs
@@ -1,5 +1,6 @@
 import { readFileSync } from "node:fs";
 import { apiFetch } from "../api.mjs";
+import { mcpCallTool } from "../mcpClient.mjs";
 import { emit } from "../output.mjs";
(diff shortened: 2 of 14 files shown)


diegosouzapw:
Thanks for this — the JSON-RPC migration is the right fix and the root-cause write-up is accurate: `/api/mcp/tools/call` was never a real route (only GET `/api/mcp/tools` exists), so those CLI tool calls 404'd. `mcp status` now reading `online` instead of `running` is also correct (`src/app/api/mcp/status/route.ts:34` returns `online`). Protocol-wise the initialize → `Mcp-Session-Id` → tools/call flow matches `open-sse/mcp-server/httpTransport.ts`.

Three things need a fix-in-place before this can merge (all small, happy to help):

1. [MANDATORY] Compression REST fallback is now dead for the common config. `/api/mcp/stream` returns **400** when the MCP transport is `stdio` (the default) and **503** when MCP is disabled (`src/app/api/mcp/stream/route.ts:14-31`) — not 404/501. `mcpCall` in `bin/cli/commands/compression.mjs:84-96` only falls back on 404/501, so on a default server `compression status/configure/engine-set` now throws. On base it worked because the old `/tools/call` POST was a guaranteed 404. I reproduced this with a mock returning 400 — the command threw instead of falling back. Please also fall back on 400/503 (or key off the 'transport is set to' message) and add a test that mocks 400, not 404.

2. [MANDATORY] The `X-MCP-Scopes` header is never sent. `mcpClient.mjs:50` builds `scopeHeader` but never passes it to `callMcpEndpoint` (the two calls at :52/:71 only forward {timeout, stream}), and `mcpJsonRpcCall` in `mcp.mjs:73-86` also ignores `opts.scope`. So `mcp call --scope …` now silently sends no scope header (the old path did). That's also a leftover dead variable. Please forward it on the tools/call POST and assert it in a test.

3. [MANDATORY] The changes to `registerMcp` delete five subcommands that were working — `mcp tools list/info/schema` and `mcp audit tail/stats` (`bin/cli/commands/mcp.mjs:24-87` no longer registers them), even though the `/api/mcp/tools` and `/api/mcp/audit/*` REST routes still exist. Removing working CLI surface is out of scope for a 'call protocol' fix; please restore the block or explicitly justify the removal in the PR description.

Minor/optional: there are now two copies of the MCP client (`mcpJsonRpcCall` with raw `fetch` vs the shared `mcpClient.mjs`), so `mcp call` bypasses the CLI-token/remote-context handling every other migrated command gets — consider routing `mcp call` through the shared client. And the rewritten tests add a lot of new `any` casts that compound the already-red `tests/` lint; the typed `makeMcpStreamFetch` helper is a good path to keep them clean.

Everything else looks solid — tests pass (23/23 on the touched files) and the server-side protocol check is correct. Thank you.

Issue #10959, opened 2026-08-21
feat(reasoning): gracefully drop opaque reasoning instead of 400 rejection on cross-model replay


adevwithpurpose:
### Description

In multi-agent environments (e.g. OpenCode, Roo Code, Cline) where different roles use different models (e.g., \`Oracle\` uses \`Codex - GPT 5.6 Terra\` and \`Build\` uses \`Gemini\`, \`Muse Spark\`, or \`Claude\`), switching models mid-thread results in a hard failure:

\`\`\`
[400]: Reasoning continuation is not compatible with the selected target
\`\`\`

### Root Cause
In \`open-sse/translator/request/openai-responses.ts\` (around line 458):
\`\`\`typescript
if (itemType === "reasoning") {
  if (preserveReasoningContent && hasOpaqueReasoningState(item)) {
    throw createReasoningTransportIncompatibleError();
  }
  ...
}
\`\`\`
When a conversation started on OpenAI Codex (Responses API) returns opaque encrypted reasoning items (\`encrypted_content\`, \`signature\`), those items remain in the client's conversation history. When the next turn is dispatched to a non-Codex provider (which uses Chat completions or plaintext reasoning), \`openai-responses.ts\` encounters \`hasOpaqueReasoningState(item)\` and throws \`createReasoningTransportIncompatibleError()\`.

### Suggested Fix
Rather than throwing a fatal 400 that breaks the multi-turn session across model switches, OmniRoute should gracefully **strip / drop** the opaque encrypted reasoning items when translating to a target provider that does not support opaque reasoning transport, allowing conversations to continue seamlessly across heterogeneous model workflows.

diegosouzapw:
Great root-cause analysis. Partial good news: the graceful-drop path already exists when traffic routes through a combo — incompatible opaque reasoning items are stripped by default instead of rejected (reasoningTransportFallback; the dashboard documents Drop as the default). Your 400 comes from the direct single-model path, where the default is still reject-by-design because the target declared it needs reasoning content. The residual piece — extending drop-default (or exposing a settings knob) to non-combo cross-model replay — is cataloged in our internal backlog and will be implemented soon; the implementing PR will reference this issue. Closing to keep the tracker lean.

adevwithpurpose:
[...]

PR #10961, opened 2026-08-21
fix(reasoning): preserve mixed plaintext and drop incompatible state (#10949, #10959)

### LINKED ISSUE #10949: [BUG] OpenCode Go mixed reasoning item drops plaintext continuation
## OmniRoute Version

3.8.50

## Installation Method

Docker / Docker Compose

## Operating System

Linux

## Provider(s) Involved

OpenCode Go

## Model(s) Involved

`deepseek-v4-flash`

## Client Tool

OpenAI Chat Completions-compatible client

## Description

I encountered a rare OpenCode Go response where a single Responses API `reasoning` item contained both a non-empty plaintext `reasoning_text` content part and non-empty `encrypted_content`.

OmniRoute currently treats any item carrying opaque state as wholly opaque. During Responses-to-Chat translation, `extractReplayableResponsesReasoningText()` therefore returns an empty string and the otherwise valid plaintext continuation is omitted from `message.reasoning_content`. A later tool-call continuation can then fail because DeepSeek requires the prior reasoning text to be passed back.

The planned compatibility rule is: when a reasoning item contains both plaintext and opaque state, treat it as plaintext. Preserve the explicit `reasoning_text` for portable continuation and do not let the coexisting opaque field suppress it.

Related: #10550 and #10574.

## Observed OpenCode Go Response Shape

This was observed in a real non-streaming `opencode-go/deepseek-v4-flash` call routed entirely through OmniRoute. The complete request and response are intentionally not included.

The reasoning item had:

- a non-empty `content` part with `type: "reasoning_text"`;
- a non-empty `encrypted_content` field (38 characters according to redacted log metadata);
- an empty `summary`;
- a function call in the same completed response.

The exact plaintext reasoning was:

```text
Let me start by reading the directory to understand the structure of the corpus.
```

On the following Chat Completions turn, the assistant tool-call history no longer contained `reasoning_text` or `reasoning_content`. A subsequent DeepSeek continuation returned:

```text
```

## Steps to Reproduce

Use a synthetic Responses item matching the observed field combination; no production request or response body is required:

```json
{
  "type": "reasoning",
  "content": [
    {
      "type": "reasoning_text",
      "text": "Let me start by reading the directory to understand the structure of the corpus."
    }
  ],
  "encrypted_content": "<opaque state>",
  "summary": []
}
```

1. Pass the item through the Responses-to-Chat non-streaming translation path.
2. Inspect the resulting assistant message.
3. Repeat for the streaming `response.output_item.done` path.

## Expected Behavior

The Chat assistant message preserves the plaintext:

```json
{
  "reasoning_content": "Let me start by reading the directory to understand the structure of the corpus."
}
```

For compatibility decisions, a mixed item with explicit plaintext should follow the plaintext transport path.

## Actual Behavior

`reasoning_content` is omitted because `hasOpaqueReasoningState(record)` causes `extractReplayableResponsesReasoningText()` to return an empty string, even though the same item contains valid plaintext `reasoning_text`.

## Test Impact

Needs a new unit test.

## Additional Context

The current transport capability is provider-level (`plaintext | opaque | none`), while payload inspection can detect both plaintext and opaque state. The mixed case is currently incompatible with every single transport and is also rejected during response extraction. OpenCode Go's rare mixed response violates the assumption that these states are mutually exclusive.

Only the exact plaintext reasoning sentence is included above. No complete production request, response, prompt, credentials, headers, tool arguments, or opaque state are included.

## Validation Plan

- Add a non-streaming Responses-to-Chat regression test with synthetic `reasoning_text + encrypted_content`.
- Add the equivalent streaming `response.output_item.done` regression test.
- Verify plaintext-only and opaque-only behavior remains unchanged.
- Run the focused reasoning transport and Responses translator unit suites.
- Run `npm run test:coverage`.

jackjinke:
## Summary

OpenCode Go can return a single Responses API `reasoning` item that contains **both** explicit plaintext (`content[].type === "reasoning_text"`) and non-empty opaque state (`encrypted_content`). The extractor treated any opaque companion as a veto and dropped the portable plaintext, so the next DeepSeek tool-call continuation failed with:

```text
```

Opaque-only reasoning from a previous provider could also hard-fail a direct cross-model replay instead of being dropped when the selected target could not transport it.

Fixes #10949 and #10959. Related: #10550, #10574.

## Change

- `extractReplayableResponsesReasoningText()` (`open-sse/services/reasoningInputPolicy.ts`): explicit plaintext now wins over a coexisting opaque companion; opaque-only items are still never converted or exposed.
- `applyReasoningInputPolicy()`: mixed input is projected onto the target transport, and incompatible reasoning is dropped by default. Combos retain the explicit **Skip** strategy to fall through without mutating the request.
- Responses → Chat request translator: mixed items replay their plaintext companion; opaque-only items are dropped instead of throwing.
- Combo advanced settings retain **Drop**/**Skip**, with **Drop** as the default and without the warning text.

## Test Plan

New regressions:

- Non-streaming extraction: mixed item preserves `reasoning_content` and caches for replay (`tests/unit/reasoning-cache.test.ts`)
- Streaming `response.output_item.done`: mixed item replays plaintext; opaque-only still emits nothing fabricated (`tests/unit/translator-resp-openai-responses.test.ts`)
- Policy projection: mixed input follows the target transport instead of rejecting, both directions (`tests/unit/strip-reasoning-blobs-agentic-context-1599.test.ts`)
- Request translator: mixed item replays its plaintext companion (`tests/unit/translator-openai-responses-req.test.ts`)
- Direct Responses and Chat replay drops incompatible opaque reasoning by default; Combo **Skip** still rejects the attempt for target fallthrough (`tests/unit/chatcore-translation-paths.test.ts`)

Legacy tests locking in the old direct mixed/opaque rejection behavior were flipped to the new contract.

Commands run:

```bash
# focused suites (all green)
npx cross-env DISABLE_SQLITE_AUTO_BACKUP=true node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-concurrency=1 \
  tests/unit/reasoning-cache.test.ts \
  tests/unit/strip-reasoning-blobs-agentic-context-1599.test.ts \
  tests/unit/translator-resp-openai-responses.test.ts \
  tests/unit/translator-openai-responses-req.test.ts
# plus chatcore/combo/translator neighbor suites (all green)

# #10959 follow-up: 155 pass
npx cross-env DISABLE_SQLITE_AUTO_BACKUP=true node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-concurrency=1 tests/unit/strip-reasoning-blobs-agentic-context-1599.test.ts tests/unit/translator-openai-responses-req.test.ts tests/unit/chatcore-translation-paths.test.ts tests/unit/combo-attempt-body-isolation-7847.test.ts

# full gate, excluding tests/unit/cli-serve-stop-command.test.ts per ~/OMNIROUTE_TESTS_SKIP_20128.txt
npm run test:coverage
```

Final coverage result (gate requires ≥60% each):

| Metric | Result |
|---|---|
| Statements | 84.02% |
| Branches | 79.06% |
| Functions | 87.25% |
| Lines | 84.02% |

Note: two pre-existing failures in `vision-bridge-maxchars.test.ts` / `vision-compression-authoritative-capability-7237.test.ts` are unrelated to this change (vision modality bridge settings, untouched here); coverage gate passed regardless.


diegosouzapw:
Nice, tightly-scoped fix. I verified the mixed plaintext+opaque projection and the new drop-by-default flow: the reasoning tests pass locally (strip-reasoning 16/16, reasoning-cache + both Responses translators 140/140, and chatcore-translation-paths 75/75 on a clean run). One optional nit: the docstring on `extractReplayableResponsesReasoningText` (open-sse/services/reasoningInputPolicy.ts, before line 94) still says "a record carrying opaque state is never cross-converted," which now contradicts the #10949 behavior you're shipping (mixed plaintext IS projected and replayed). Consider rewording the first sentence so future readers don't misread the intent. No mandatory items from my review.

jackjinke:
@diegosouzapw Addressed the docstring nit in 0d8669334: it now distinguishes opaque-only records from mixed records and states that explicit plaintext remains portable alongside opaque state. Focused reasoning and translation suites pass 231/231.

PR #10970, opened 2026-08-21
fix(api): intercept kimi-web credential updates to extract refresh_token into dedicated column

MeRezaRezaei:
### Summary

When a user pastes a full Kimi JSON dump or raw key-value string into the Dashboard's API key field, the frontend sends it as the `apiKey` payload to the provider creation/update API.

This PR intercepts calls to `POST /api/providers` and `PATCH /api/providers/[id]` for `kimi-web`, runs `extractKimiCredentials(apiKey)`, and routes the `accessToken` and `refreshToken` into their own dedicated database columns so the new Kimi auto-refresh mechanism can reliably find them.

### Changes
- **`src/app/api/providers/route.ts`**: On creation of a `kimi-web` provider, if credentials are provided in `apiKey`, they are automatically separated.
- **`src/app/api/providers/[id]/route.ts`**: On update, we check if `c.provider === "kimi-web"` and split the payload into `updates.apiKey` and `updates.refreshToken` appropriately.


MeRezaRezaei:
Closing failing CI PR; folding into a single consolidated kimi-web branch.

PR #10972, opened 2026-08-21
fix(ui): expose manual token refresh button for kimi-web connections

MeRezaRezaei:
### Summary

The previous PR added backend support for `kimi-web` token refresh (`POST /api/providers/[id]/refresh-token`), but the dashboard UI selectively hid the refresh button because `kimi-web` is mapped as an `apikey` provider type rather than `oauth`.

This PR fixes the UI mapping so the "Refresh Token" action displays correctly for Kimi Web accounts, and routes the API call to the newly added `/refresh-token` endpoint instead of the standard OAuth `/refresh` pipeline.


MeRezaRezaei:
Closing failing CI PR; folding into a single consolidated kimi-web branch.

PR #10984, opened 2026-08-21
fix(db): add missing @/lib/db/connections module breaking dashboard build

linhdmn:
## Problem

#10939 added this import to `src/app/api/usage/utilization/route.ts`:

```ts
import { getConnection } from "@/lib/db/connections";
```

…but no such module exists anywhere in the repo, so the dashboard build fails on current HEAD:

```
Error: Turbopack build failed with 1 error:
./src/app/api/usage/utilization/route.ts:3:1
Error: Module not found: Can't resolve '@/lib/db/connections'
```

(The PR's description notes CI was already red via base-red #9985, which is why this slipped through.)

## Fix

Add `src/lib/db/connections.ts` following the existing db-helper pattern (`getDbInstance` + prepared statement, cf. `connectionRuntimeState.ts`):

```ts
getConnection(connectionId): { id, provider, name, email } | null
```

— exactly the email/name metadata the route maps into `ConnectionMetaEntry`.

Also replace the route's `conn?.displayName ?? null` with a literal `null`: `provider_connections` has no display-name column, and the Account Split card already falls back through `getAccountDisplayName` (displayName → name → email), so behavior for that field is unchanged while types stay honest.

## Verification

```
npm run build
# before: Error: Module not found: Can't resolve '@/lib/db/connections'
# after:  ⚡ Done — standalone tree written
```

Remaining `tsc --noEmit` findings are pre-existing base-red items in unrelated files (#9985); none touch the files changed here.

### DIFF
--- a/src/app/api/usage/utilization/route.ts
+++ b/src/app/api/usage/utilization/route.ts
@@ -71,7 +71,9 @@ export async function GET(request: Request) {
         connectionMeta[cid] = {
           email: conn?.email ?? null,
           name: conn?.name ?? null,
-          displayName: conn?.displayName ?? null,
+          // provider_connections has no display-name column; the UI falls back
+          // to name/email via getAccountDisplayName.
+          displayName: null,
         };
       }
     }
--- a/src/lib/db/connections.ts
+++ b/src/lib/db/connections.ts
@@ -1,0 +1,21 @@
+import { getDbInstance } from "./core";
+
+export interface ProviderConnectionRef {
+  id: string;
+  provider: string;
+  name: string | null;
+  email: string | null;
+}
+
+/**
+ * Look up a single provider connection by id. Backs dashboard surfaces that
+ * need human-readable account metadata (email/name) for a connection id,
+ * e.g. the utilization Account Split cards.
+ */
+export function getConnection(connectionId: string): ProviderConnectionRef | null {
+  const db = getDbInstance();
+  const row = db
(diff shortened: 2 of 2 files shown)


diegosouzapw:
Thanks for jumping on the dashboard build break from #10939 — the module was indeed missing when your branch was cut. Heads up that the current base (release/v3.8.50) has already resolved this differently: the route now imports `getProviderConnectionById` from `@/lib/db/providers` (src/app/api/usage/utilization/route.ts:3), and a regression test (tests/unit/usage-utilization-connection-meta.test.ts:28) now asserts the route does NOT import `@/lib/db/connections`. So the missing-module problem you fixed is already gone upstream.

Two things would break if we merged this as-is (both verified locally against current base):
1. Your route.ts:3 re-introduces the `@/lib/db/connections` import, which fails the existing regression guard (probe run: 'utilization route does not import phantom @/lib/db/connections' → assert true !== false).
2. route.ts:76 hardcodes `displayName: null` with a comment stating provider_connections has no display-name column — but it does have one (migration 127 references c.display_name), and `getProviderConnectionById` (SELECT * + rowToCamel, src/lib/db/providers.ts:330) returns it. I verified it returns the seeded display name. So this would drop real account display names from the Account Split card (the feature #10939 added).

Recommendation: close this PR rather than merge — the build fix is already in base via the providers route, and the displayName null is a functional regression grounded on an incorrect premise. Thanks for the report; the reproduction was valuable.

PR #11007, opened 2026-08-21
feat(cli): audit the resolved security config at boot in `omniroute serve`

aryehbeitz:
## Summary

Two commits: a boot-time security preflight for `omniroute serve`, plus a drain of the `check:env-doc-sync` backlog that the first commit surfaced.

### 1. `feat(cli)` — audit the resolved security config at boot

OmniRoute's defaults are tuned for the single-user localhost case — `resolveServerHost()` falls back to `0.0.0.0`, `REQUIRE_API_KEY` defaults to `false`, `STORAGE_ENCRYPTION_KEY` is empty in `.env.example`. Each is deliberate and documented, and together they are what makes `npx omniroute` work with zero configuration.

The gap this closes is that an **exposed** deployment and a **safe** one print exactly the same startup banner. Nothing tells the operator that the listener they just opened on a VPS answers `/v1/chat/completions` anonymously, or that the OAuth refresh tokens they are about to connect will land in SQLite as cleartext.

`omniroute serve` now audits the configuration actually resolved for this boot, before anything listens:

| Severity     | id                              | Condition                                                            |
| ------------ | ------------------------------- | -------------------------------------------------------------------- |
| **CRITICAL** | `anonymous-api-network-exposed` | Bind host is not loopback **and** `REQUIRE_API_KEY` is not `true`     |
| **HIGH**     | `credential-store-plaintext`    | `STORAGE_ENCRYPTION_KEY` unset or blank                              |
| **HIGH**     | `initial-password-placeholder`  | `INITIAL_PASSWORD` is still the literal `.env.example` `CHANGEME`     |
| **MEDIUM**   | `auth-cookie-insecure`          | Bind host is not loopback **and** `AUTH_COOKIE_SECURE` is not `true`  |

**No behavior change by default.** `warn` (the default) reports and continues — it never touches the exit code, so every existing deployment boots exactly as it did before. A fully hardened boot prints nothing at all.

`OMNIROUTE_SECURITY_PREFLIGHT=strict` refuses to start on a CRITICAL finding, for unattended VPS/container deploys where a misconfigured exposure should fail the deploy rather than quietly serve traffic. It gates on CRITICAL only, so enabling it cannot turn a working deployment into a boot loop over a cookie flag. `off` skips the audit entirely. An unrecognized value falls back to `warn` — a typo in the variable can never fail a boot.

**Why the anonymous case is CRITICAL:** with `REQUIRE_API_KEY` off, `clientApiPolicy` (`src/server/authz/policies/clientApi.ts`) allows a request carrying no bearer at all, and an *invalid* bearer degrades to anonymous rather than 401ing. There is no source-IP restriction on the `CLIENT_API` tier — that is by design, and correct on loopback. Combined with a `0.0.0.0` bind it means any host that can reach the port (LAN neighbour, Tailscale peer, or the internet when the port is forwarded or published by a container) can spend the credentials and free-tier quota on that instance.

**Deliberately quiet on loopback:** the two bind-conditioned checks stay silent on a loopback bind. On loopback there is no network hop to sniff and no remote caller to authenticate, and an audit that cries wolf on the default laptop install is one operators learn to skip.

### 2. `chore(docs)` — drain the env/docs contract backlog

Adding one variable made `check:env-doc-sync` fail on **6 pre-existing** vars referenced in code but absent from `.env.example`, so this PR clears them too. Each was verified against its actual source before being written up (AGENTS.md → Documentation accuracy) — defaults come from the code, not from inference:

- **`KIMI_WEB_BASE_URL` / `KIMI_WEB_CHAT_URL`** — kimi-web executor endpoint overrides (`open-sse/executors/kimi-web.ts`). Defaults read off `getKimiWebBaseUrl()` / `getKimiWebChatUrl()`: `https://www.kimi.ai` and `<base>/apiv2/kimi.gateway.chat.v1.ChatService/Chat`. Documented that trailing slashes are stripped from the base, and that the chat override *replaces* the value derived from the base rather than composing with it.
- **`OMNIROUTE_OIDC_DISABLE_PASSWORD_LOGIN`** — rejects password login with 403 so the dashboard is reachable only via OIDC SSO. Documented the gate that actually matters: it is `settings.oidcEnabled === true && (…)` in `src/app/api/auth/login/route.ts`, so the variable is **inert while OIDC is off**. Also reachable as a Feature Flag and as the `oidcDisablePasswordLogin` setting.
- **`OIDC_DISABLE_PASSWORD_LOGIN`** — the accepted legacy alias. It has a live runtime hook, so it belongs in the documented contract rather than in `DOC_ONLY_ALLOWLIST` (which is for aliases *without* a hook).
- **`ELECTRON_SMOKE_COLD_RESTART`** — missed when #7592 added it. All six `ELECTRON_SMOKE_*` siblings are documented rather than allowlisted, so this follows them.
- **`OMNIROUTE_STANDALONE_DIR`** — allowlisted in `IGNORE_FROM_CODE` with justification instead. Per the comment at its definition it is a test-only seam pointing `colocate-standalone.mjs` at a synthetic tree so the co-location logic can be driven without a full `next build` — the same override-seam family as the already-allowlisted `OMNIROUTE_BUILD_SHA`, and not user-facing `.env` config. Happy to document it instead if you would rather it be visible.

`check:env-doc-sync` is now green in all three directions.

## Related Issues

- Closes #
- Related to #

⚠️ base-red inherited: #9985

## Validation

- [x] Change type: **CLI** (+ docs)
- [ ] `npm run lint` — could not run locally (see Reviewer Notes); CI covers it

Focused loop run locally, all green:

```
node --import tsx/esm --test tests/unit/cli-security-preflight.test.ts   # 33 pass / 0 fail
npm run check:env-doc-sync            # PASS — in sync in all three directions
npm run check:cli-i18n                # PASS — 762 t() keys, 69 pt-BR sections
npm run check:docs-sync               # PASS
npm run check:doc-links               # PASS — 854 internal links
npm run check:fabricated-docs         # PASS — no fabricated env/CLI/file refs
npm run check:docs-symbols            # PASS
npm run check:changelog-integrity     # PASS — no base bullets lost
npm run check:any-budget:t11          # PASS
npm run check:file-size               # PASS
npm run check:tracked-artifacts       # PASS
npx prettier --check <all touched files>   # clean
```

Functional check of the three preflight modes, as `runServe` calls it:

```
--- default boot (host 0.0.0.0, nothing set) ---
[omniroute][security] 3 security findings for this boot:
[omniroute][security]   CRITICAL: the /v1 API accepts unauthenticated requests and is bound to 0.0.0.0
[omniroute][security]     fix: set REQUIRE_API_KEY=true, or bind loopback only with OMNIROUTE_SERVER_HOST=127.0.0.1
[omniroute][security]   HIGH: provider credentials are stored unencrypted
[omniroute][security]     fix: set STORAGE_ENCRYPTION_KEY="$(openssl rand -hex 32)" before connecting accounts
[omniroute][security]   MEDIUM: dashboard session cookies are sent without the Secure flag
[omniroute][security]     fix: set AUTH_COOKIE_SECURE=true once the deployment is served over HTTPS

--- hardened boot (REQUIRE_API_KEY + STORAGE_ENCRYPTION_KEY + AUTH_COOKIE_SECURE) ---
(no output)  findings: 0  ok: true

--- strict mode on an exposed anonymous boot ---
[omniroute][security] Refusing to start: OMNIROUTE_SECURITY_PREFLIGHT=strict and a CRITICAL finding is present.
ok: false  => CLI exits 1
```

## Tests Added Or Updated

- `tests/unit/cli-security-preflight.test.ts` — **new, 33 cases.** Covers `isLoopbackBindHost` (loopback forms, the full `127.0.0.0/8` block, `0.0.0.0`/`::` explicitly **not** loopback, LAN/CGNAT/public/hostname, malformed input failing closed), each finding's trigger and its non-trigger, the `true`-only boolean parsing, severity ordering, finding-shape completeness, env non-mutation, `resolvePreflightMode` (default, case-insensitivity, typo fallback), the formatter (prefix on every line, singular/plural, strict refusal line, empty input) and all three `runSecurityPreflight` modes including "strict allows the boot when nothing is CRITICAL".

The second commit changes only `.env.example`, `docs/reference/ENVIRONMENT.md`, a changelog fragment, and one allowlist entry in `scripts/check/check-env-doc-sync.mjs` — no production code under `src/`, `open-sse/`, `electron/` or `bin/`, so it carries no new tests. Its own gate (`check:env-doc-sync`) is the regression guard.

## Coverage Notes

The first commit touches `bin/`. Both new production units are covered by the new test file:

- `bin/cli/utils/securityPreflight.mjs` — every exported function and every finding branch is exercised, including the negative cases (each check's silent path) and the fail-closed paths.
- `bin/cli/commands/serve.mjs` — the added lines are the import plus a 3-line call/guard. `resolveTlsOptions(env)` was already called there; it is now captured into `tlsEnabled` and reused for the banner scheme, so no new call is introduced. The abort uses `process.exit(1)`, matching the two existing pre-listen abort paths in the same function.

## Reviewer Notes

- **Zero default behavior change.** With `warn`, nothing about startup changes except stderr output when something is genuinely exposed; a hardened boot is byte-identical to today.
- **`npm run lint` / `check:dead-code` / `check:complexity` were not runnable in my environment** (no `node_modules`: eslint and knip are missing), so I verified those by hand instead — the module has no `any` (`check:any-budget:t11` passes), no unused exports (every export is imported by `serve.mjs` or the test), and the largest function has 4 independent branches. Please let CI be the authority on the eslint/knip gates.
- **`docs/reference/ENVIRONMENT.md` is in `.prettierignore`** because its long reference tables are hand-aligned. I respected that: the new rows are padded to their own table's header widths and **no pre-existing row is reformatted** — the file's diff is +12/−1, and markdownlint MD060 on it goes from 446 errors to 445. My first attempt realigned whole tables (which would have been −23 errors but +29 lines of unrelated churn); I backed that out, since whole-file reformatting is exactly what the ignore entry exists to prevent. Happy to land the full realignment as its own PR if you want the file cleaned up.
- **One relocation in that file:** the Kimi Web section sits *after* the fingerprint/User-Agent `> [!NOTE]`, not before it, so that note keeps describing the section it belongs to.
- **`SECURITY.md` diff looks larger than it is.** The base file is not prettier-clean, so adding one (wider) row to the auth table made prettier realign the whole table. `git diff --ignore-all-space` on that file shows only two real lines: my new row and the widened separator. Unlike ENVIRONMENT.md this file is *not* prettier-ignored, and lint-staged runs `prettier --write` on any staged `*.md`, so this is what the repo's own tooling produces.
- No migrations, no feature flags, no new dependencies, no route changes — so no `routeGuard.ts` classification is required here (Hard Rules #15/#17 not in scope).
- On the deliberate non-reuse of `routeGuard.ts::isLoopbackHost`: importing it would pull the Next server module graph (`runtimeSettings` → `localDb` → `ioredis`) into the CLI process, and it answers a different question — it classifies an untrusted inbound **peer** (so it must handle `Host` headers, ports, `::ffff:` mappings and must never be trusted), whereas this classifies the local **listen address** the CLI is about to bind. Rationale is recorded in the module header and in `docs/security/STARTUP_PREFLIGHT.md` → Design notes.


diegosouzapw:
Great feature — the preflight is clean, pure, well-tested and genuinely closes a real operational gap. It's fix-in-place-able via /merge-prs: the blocker is not the code but the second commit.

[MANDATORY] Rebase onto origin/release/v3.8.50 and resolve the merge conflicts in `.env.example` and `docs/reference/ENVIRONMENT.md`. The base has already merged (via #10889) the OIDC_DISABLE_PASSWORD_LOGIN, KIMI_WEB_BASE_URL, and ELECTRON_SMOKE_COLD_RESTART docs that this PR's "env-doc-sync-drain" commit is re-adding — that commit is largely redundant with the current base. Drop the duplicated OIDC/Kimi/Electron doc additions and keep only the genuine new work: the `OMNIROUTE_STANDALONE_DIR` allowlist entry in `scripts/check/check-env-doc-sync.mjs`, plus your new `OMNIROUTE_SECURITY_PREFLIGHT`/`STARTUP_PREFLIGHT` content (which the base does NOT have). This will both clear the conflict and stop the duplicate-doc churn.

[OPTIONAL] Minor accuracy refinement worth a comment only: since `REQUIRE_API_KEY` and `AUTH_COOKIE_SECURE` can be overridden via the Dashboard Feature Flags (DB override, `src/shared/utils/featureFlags.ts` — a DB override beats `process.env`), an operator who toggles REQUIRE_API_KEY in the dashboard rather than `.env` will get a spurious CRITICAL from the env-only preflight. That is an inherent and acceptable tradeoff of the pure/dependency-light design (the constraint is even documented), but a one-line note in `docs/security/STARTUP_PREFLIGHT.md` acknowledging the env-vs-DB override gap would set correct expectations.

Otherwise: logic, tests (33 passing), severity ordering, loopback/host classification, and the advisory-by-default behavior all check out against the actual runtime code. No changes needed to the core preflight.

aryehbeitz:
Withdrawing this — I'm no longer running OmniRoute, so I can't maintain the change or respond to review follow-ups. Closing rather than leaving it to go stale.

The two commits stand on their own if either is useful to you: the boot-time config audit (`bin/cli/utils/securityPreflight.mjs` + 33 unit tests, advisory by default) and the `check:env-doc-sync` drain that documents `KIMI_WEB_BASE_URL`/`KIMI_WEB_CHAT_URL`, `OMNIROUTE_OIDC_DISABLE_PASSWORD_LOGIN` and its legacy alias, and `ELECTRON_SMOKE_COLD_RESTART`. Feel free to cherry-pick or reimplement; no attribution needed.

PR #2289, opened 2026-05-15
fix: remove implicit API key request caps

josephvoxone:
## Summary

This removes OmniRoute's implicit per-API-key request caps when an API key has not explicitly configured rate limits.

Previously, every persisted API key without a custom `rateLimits` value silently inherited these generic limits:

- 1,000 requests per day
- 5,000 requests per week
- 20,000 requests per month

That policy could reject traffic before provider routing, account selection, quota preflight, or combo fallback had a chance to run. In production this surfaced as `429 Too Many Requests` from OmniRoute itself even when provider accounts still had usable quota and fallback accounts were available.

## Root Cause

`enforceApiKeyPolicy()` always built `rulesToApply` from `DEFAULT_RATE_LIMITS` whenever `apiKeyInfo.rateLimits` was empty or null. This made the default behavior rate-limited even for unrestricted gateway keys.

The resulting failure mode is misleading for users and operators:

1. A client calls a combo or Codex model such as `cx-5.5`.
2. API key policy runs before combo/provider routing.
3. The implicit daily/week/month counter rejects the request with 429.
4. No provider/account fallback runs, because the request never reaches that layer.
5. The user sees a 429 and assumes a provider quota issue, while the real source is the router's generic key policy.

This is especially harmful for aggregator/gateway deployments where a single API key fronts multiple provider accounts and quota/fallback is intentionally managed at the provider-account layer.

## Fix

The default API key rate limit list is now empty. API keys still support rate limiting, but only when the operator explicitly configures `rateLimits`, `maxRequestsPerDay`, or `maxRequestsPerMinute` in API Manager.

This keeps the existing explicit-limit behavior intact while preventing surprise router-level 429s for unrestricted keys.

## Validation

- Added a regression test proving an unrestricted key can exceed the previous 1,000/day default without being rejected.
- Existing request-per-minute policy test still verifies explicit key limits are enforced.
- Pre-commit hooks passed during commit, including formatting, eslint, docs/env checks, and any-budget checks.
- Manually reproduced the production failure mode with a high-usage API key and confirmed requests now reach Codex routing instead of being rejected by API key policy.

## Notes

This does not remove provider quota controls, account cooldowns, budget checks, model allowlists, schedules, disabled/banned key checks, or explicit API key rate limits. It only removes the implicit global request caps that were applied to otherwise unrestricted keys.

PR #11031, opened 2026-08-21
docs: DEFAULT_RATE_LIMIT_PER_DAY unset is unlimited

### LINKED ISSUE #11017: docs: DEFAULT_RATE_LIMIT_PER_DAY unset is unlimited; ENVIRONMENT.md still says 1000/day
### OmniRoute Version

3.8.49 and current `release/v3.8.50`

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Description

**`docs/reference/ENVIRONMENT.md` still documents `DEFAULT_RATE_LIMIT_PER_DAY` default as `1000`. Code since #2289 is unlimited when unset.**

`src/shared/utils/apiKeyPolicy.ts`:

```ts
export function buildDefaultRateLimits(rawValue?: string): RateLimitRule[] {
  const normalized = rawValue?.trim();
  if (normalized === undefined || normalized === "") return [];
  ...
}
```

`tests/unit/apikey-policy-default-rate-limits.test.ts` locks **"unset / empty env disables implicit fallback limits"**.

ENVIRONMENT.md (release/v3.8.50) still says:

> Default (unset/empty/malformed) keeps the legacy 1000/day, 5000/week, 20000/month windows. Set explicitly to `0` to opt out (unlimited).

That is the **pre-#2289** contract. Operators following the table either:

- set `DEFAULT_RATE_LIMIT_PER_DAY=0` as a no-op cargo-cult, or
- believe keys with `rate_limits=null` are capped at 1000/day and mis-debug 429s that are actually provider/admission, or
- copy `1000` into GitOps thinking they are "matching the default" and **silently 429 a multi-harness coding-agent gateway** after 1000 `/v1/responses` in a day.

CHANGELOG still carries the old "legacy default safety net (1000/day)" wording in historical entries; the live env table must match `buildDefaultRateLimits`.

### Steps to Reproduce

1. Open `docs/reference/ENVIRONMENT.md` on `release/v3.8.50` — default column `1000`, prose says unset = 1000/day.
2. Run `buildDefaultRateLimits(undefined)` / the unit test — `[]`.
3. Create an API key with `rate_limits=null`, do not set the env, send `/v1/*` — no 1000/day OmniRoute cap.

### Expected Behavior

Env table: default **unset = no implicit per-key cap**. `0` remains an explicit opt-out (same as unset). Positive N = N/day, 5N/week, 20N/month. Malformed value: keep current code behaviour (legacy 1000) and document that only.

### Actual Behavior

Docs say 1000. Code + tests say unlimited. GitOps operators cannot tell which 429 is theirs.

### Test Impact

(docs vs `tests/unit/apikey-policy-default-rate-limits.test.ts` contract — no production test is red; the published table is wrong.)

### Additional Context

- Code comment in `apiKeyPolicy.ts`: "Default to no per-key request cap" / `#2289`.
- Dockerfile does **not** pin `DEFAULT_RATE_LIMIT_PER_DAY`.
- Multi-harness coding-agent load is thousands of `/v1/responses` per day on one key; a 1000/day safety net (if anyone copies the docs) is an outage.

### Validation Plan

- ENVIRONMENT.md default column and prose match `buildDefaultRateLimits`.
- `node --import tsx/esm --test tests/unit/apikey-policy-default-rate-limits.test.ts` stays green.

Prajeeth-12:
Fixes #11017

### What this PR does
- Updates the documentation in `docs/reference/ENVIRONMENT.md` to accurately reflect that leaving `DEFAULT_RATE_LIMIT_PER_DAY` unset or empty results in unlimited rate limits (no implicit caps), rather than preserving the legacy 1000/day limit.
- Fixes the identical incorrect comments in `.env.example`.
- Clarifies that the legacy 1000/day fallback is only applied to *malformed* values.


diegosouzapw:
Thanks for this docs-only correction — it's accurate and needed. I verified it against the code: in `src/shared/utils/apiKeyPolicy.ts` an unset/empty `DEFAULT_RATE_LIMIT_PER_DAY` returns `[]` (unlimited, line 46) and only malformed non-numeric values fall back to the legacy 1000/day (line 48), so your rewrite is correct and removes stale text that predated the #2289 behavior change.

One optional follow-up (not blocking): the `Default` column in `docs/reference/ENVIRONMENT.md:186` still lists `1000`, which slightly conflicts with the new "unset/empty = unlimited" framing — worth aligning to `_(unset = unlimited)_` for consistency. The i18n mirrors (e.g. `docs/i18n/zh-CN/docs/reference/ENVIRONMENT.md`) still carry the old wording but are generator-synced via `i18n:sync`, so no manual edit needed there.

PR #11036, opened 2026-08-21
fix(combo): accept SSE comment lines (e.g. OpenRouter keep-alives) in response quality validation

asorourx:
## Problem

The combo response quality gate fails a streamed **HTTP 200** completion as
`response is not valid JSON` whenever the SSE body opens with a **comment line**
(a line beginning with `:`) or with leading whitespace/newlines.

In `validateResponseQuality` (`open-sse/services/combo/validateQuality.ts`), when
`JSON.parse(text)` fails on a streamed body, the fallback only treats it as a valid
stream if it `startsWith("data:")` or `startsWith("event:")`:

```ts
try {
  json = JSON.parse(text);
} catch {
  if (text.startsWith("data:") || text.startsWith("event:")) return { valid: true };
  return { valid: false, reason: "response is not valid JSON" };
}
```

Per the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation),
lines starting with `:` are **comments**, which providers use as keep-alives while the
model is still generating. **OpenRouter emits `: OPENROUTER PROCESSING`** at the start of
the stream on slower / reasoning responses. Such a stream is perfectly valid, but because
its first bytes are `:` (not `data:`/`event:`), the gate rejects it — and since every
member of a combo returns the same shape, the whole combo cascades to a **502**.

It presents as **intermittent**: fast replies stream `data:` immediately and pass; slower
ones get the keep-alive comment first and fail. The rejected responses contain real, valid
content and reconstruct to valid JSON downstream — only the raw-stream sniff is wrong.

## Fix

Trim leading whitespace and accept the `:` (SSE comment) prefix before rejecting:

```ts
const trimmed = text.trimStart();
if (trimmed.startsWith("data:") || trimmed.startsWith("event:") || trimmed.startsWith(":")) {
  return { valid: true };
}
return { valid: false, reason: "response is not valid JSON" };
```

## Tests

Added two regression cases to `tests/unit/validate-response-quality.test.ts`:
- SSE stream opening with a `: OPENROUTER PROCESSING` comment → `valid: true`
- SSE stream with leading whitespace before the first frame → `valid: true`

Full file passes (14/14), existing cases unchanged:

```
# tests 14
# pass 14
# fail 0
```

### DIFF
--- a/open-sse/services/combo/validateQuality.ts
+++ b/open-sse/services/combo/validateQuality.ts
@@ -617,7 +617,22 @@ export async function validateResponseQuality(
   try {
     json = JSON.parse(text);
   } catch {
-    if (text.startsWith("data:") || text.startsWith("event:")) return { valid: true };
+    // An SSE stream body is expected for streamed upstreams. Besides `data:` and
+    // `event:` frames, the SSE spec also allows comment lines that begin with a
+    // colon (`:`), which providers use for keep-alives while the model is still
+    // generating — e.g. OpenRouter emits `: OPENROUTER PROCESSING` on slower /
+    // reasoning responses. A stream that opens with such a comment (or with
+    // leading whitespace/newlines) is still a valid stream, not malformed JSON,
+    // so trim and recognize the comment prefix before rejecting. Without this,
+    // otherwise-good streamed completions get failed as "not valid JSON".
+    const trimmed = text.trimStart();
+    if (
+      trimmed.startsWith("data:") ||
+      trimmed.startsWith("event:") ||
+      trimmed.startsWith(":")
+    ) {
+      return { valid: true };
+    }
     return { valid: false, reason: "response is not valid JSON" };
   }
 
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Nice catch — `: OPENROUTER PROCESSING` keep-alives are legit spec-compliant SSE comments, so this raw-stream sniff was rejecting valid streams and cascading combos to 502. The fix is correct and low-risk (it only widens acceptance on the JSON.parse-failure branch), and the two regression tests are good. I validated in a clean worktree: 14/14 pass, existing cases unchanged.

Optional (not blocking, for consistency with repo Prettier style): Prettier would collapse the new 3-line `if` at validateQuality.ts:629-631 into a single line since it fits under 100 chars. No action required — this file already fails Prettier on base, so it isn't a gate.

Nothing mandatory outstanding. Good scope, good tests, good explanation.

asorourx:
Thanks for the quick, thorough review! Collapsed the `if` onto one line per the Prettier note (97 chars) — pushed as `2bc37a36a`, tests still 14/14. Ready whenever you are.

PR #11037, opened 2026-08-21
fix(release): drain v3.8.50 base-reds — build-breaking import, stale provider docs, orphaned suppressions

diegosouzapw:
Refs #9985

Drains the three classes of base-red currently failing every PR against release/v3.8.50:

1. **Fast Production Build (module-not-found)** — `src/app/api/usage/utilization/route.ts` imported `getConnection` from `@/lib/db/connections`, a module that does not exist (introduced by the #10939 merge, c0fd109e30; typecheck:core does not cover app routes, so only next build catches it). Fixed to `getProviderConnectionById` from `@/lib/db/providers` (awaited). Regression test: `tests/unit/utilization-route-import-10939.test.ts` (module import is the repro — ERR_MODULE_NOT_FOUND on the base, resolves here).
2. **Docs Gates (fast-path)** — `docs/reference/PROVIDER_REFERENCE.md` regenerated via `npm run gen:provider-reference` (346 → 347) and the four README diagram SVGs updated to the real count.
3. **No new ESLint warnings** — `npm run lint:json -- --max-warnings 0 --prune-suppressions` on the clean base tip removed suppressions orphaned by earlier merges (the gate fails with "suppressions left that do not occur anymore" for every PR); lint exits 0 after the prune.


diegosouzapw:
Closing — this PR was accidentally created pointing at another session's in-flight branch of the same name (its head carries a compression fix, not the changes described here). The described base-red fixes are being reopened from a distinctly named branch.

PR #10988, opened 2026-08-21
feat(search): first-class X Search via SuperGrok x_search

RaviTharuma:
## Summary

Adds a first-class `x-search` provider on the search gateway so SuperGrok / xAI server-side `x_search` is available on `POST /v1/search` and MCP `omniroute_x_search`.

This is Grok X Search (`api.x.ai` `tools: [{ "type": "x_search" }]`), not web search and not the X Developer Platform MCP (`api.x.com/mcp`). Generic `search_type: "web"` queries never auto-select it. Credentials fall back to `xai-oauth` (SuperGrok), `xao`, then `xai`.

## Related Issues

- Closes #10985
- Related to #10728
- Related to #9111

## Validation

Choose the change type and focused loop from the
[Contribution Golden Path](../docs/ops/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite,
Vitest, the 60% coverage gate, and the production build all run in CI on this PR (#8329):

- [x] Change type: provider / routing / UI / i18n / CLI / DB / build-deploy / other
- [x] Focused tests and category gates from the golden path
- [x] ESLint on changed TypeScript files
- [x] Reconciled with the current active release base (`origin/release/v3.8.50` @ `266e39d36`); focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented below (CI)

Focused loop run locally:

```
node --import tsx/esm --test tests/unit/x-search-provider.test.ts tests/unit/search-registry.test.ts tests/unit/search-route.test.ts
# 67/67 pass

npx vitest run --config vitest.mcp.config.ts open-sse/mcp-server/__tests__/essentialTools.test.ts
# 20/20 pass

bun scripts/check/check-known-symbols.ts
bun scripts/check/check-provider-consistency.ts
node scripts/check/check-provider-assets.mjs
node scripts/check/check-docs-counts-sync.mjs
npm run gen:provider-reference
```

Full `npm run lint` and the production build were not run locally; they run in CI.

## Tests Added Or Updated

- `tests/unit/x-search-provider.test.ts` (new) — registry selection, request build, citation parse, schema alias, handleSearch mapping
- `tests/unit/search-registry.test.ts` — catalog count 14 → 15
- `tests/unit/search-route.test.ts` — `GET /v1/search` lists `x-search`
- `tests/integration/search-providers-catalog.test.ts` — search-kind count 15
[...]

PR #11038, opened 2026-08-21
fix(release): drain v3.8.50 base-reds — build-breaking import, stale provider docs, orphaned suppressions

diegosouzapw:
Refs #9985

Drains three classes of base-red currently failing every PR against release/v3.8.50 (complementary to the in-flight compression basereds branch — zero file overlap):

1. **Fast Production Build (module-not-found)** — `src/app/api/usage/utilization/route.ts` imported `getConnection` from `@/lib/db/connections`, a module that does not exist (introduced by the #10939 merge, c0fd109e30; typecheck:core does not cover app routes, so only next build catches it). Fixed to `getProviderConnectionById` from `@/lib/db/providers` (awaited). Regression test: `tests/unit/utilization-route-import-10939.test.ts` (module import is the repro — ERR_MODULE_NOT_FOUND on the base, resolves here).
2. **Docs Gates (fast-path)** — `docs/reference/PROVIDER_REFERENCE.md` regenerated via `npm run gen:provider-reference` (346 → 347) and the four README diagram SVGs updated to the real count.
3. **No new ESLint warnings** — `npm run lint:json -- --max-warnings 0 --prune-suppressions` on the clean base tip removed suppressions orphaned by earlier merges (the gate fails with "suppressions left that do not occur anymore" for every PR); lint exits 0 after the prune.


diegosouzapw:
Thanks for chasing this down — the work here is sound and the intent (drain the #10939 build-breaking import) is exactly right. However, this PR has been superseded: the same fix (`getProviderConnectionById` from `@/lib/db/providers`) already landed on release/v3.8.50 via commit 137e49e (#10988), which is now on the release tip, and a stronger regression test (`tests/unit/usage-utilization-connection-meta.test.ts`) already ships there covering the phantom-import guard plus a full GET/connection-meta run. Your head is 25 commits behind the current tip, so the ESLint-suppression pruning and docs regenerated to a "347" count rest on a stale base (the tip's README/PROVIDER_REFERENCE already read 348). I'd recommend against force-merging this stale branch: it would conflict with already-correct code and could regress refreshed docs. Better path: verify the current tip's utilization route + its existing connection-meta test are green (the base-red for #10939 is no longer listed among current release reds), and close this PR without merging rather than re-apply the patch — [OPTIONAL] if the 685 suppression lines and provider-count docs are genuinely stale, re-cut a fresh, small cleanup PR off the current tip. [MANDATORY — nothing to merge here; the fix is already integrated.]

PR #11039, opened 2026-08-21
feat(cli): add native Bun backend support and Dockerfile.bun

rqzbeh:
## Summary

This PR adds native Bun backend runtime support, multi-stage Docker build configuration (`Dockerfile.bun`), dynamic driver fallback logic, and resolves DAST CI workflow stability.

### Key Changes

1. **Bun Runtime Detection & Container Support (`Dockerfile.bun`)**:
   - Added multi-stage `Dockerfile.bun` for native Bun production deployments.
   - Added runtime detection utilities (`bin/nodeRuntimeSupport.mjs`, `src/shared/utils/nodeRuntimeSupport.ts`).

2. **SQLite Driver Adapter Prioritization (`bin/cli/sqlite.mjs`, `src/lib/db/adapters/driverFactory.ts`)**:
   - Enforces native `better-sqlite3` driver prioritization under Bun with smooth fallback to `bun:sqlite` and `sqlite3`.
   - Updated native dependency loader (`bin/cli/runtime/nativeDeps.mjs`) for cross-runtime binary loading.

3. **Build & Dev Tooling Adaptation (`scripts/build/build-next-isolated.mjs`, `scripts/dev/run-next.mjs`)**:
   - Automatically switches Next.js build engine to Webpack when executing under Bun (since Turbopack requires V8 Node bindings).
   - Added `export const dynamic = "force-dynamic";` annotations to API routes to prevent static generation build failures under Bun.

4. **DAST Smoke CI Workflow Fixes (`.github/workflows/dast-smoke.yml`)**:
   - **Auth Guard Test Seam**: Added `REQUIRE_API_KEY: "false"` to server start env so Promptfoo can test `injectionGuard` on `POST /v1/chat/completions` without failing early at `routeGuard` authorization with HTTP 401 (`AUTH_002`).
   - **Rate-Limit Exclusion**: Excluded brute-force rate-limited password login (`/api/auth/login`) from Schemathesis payload fuzzing (`--exclude-path-regex '^/api/auth/(oidc/|login)'`) to avoid HTTP 429 status code assertions.
   - **CLI Tool Pinning**: Pinned `promptfoo@0.122.0` global installation to prevent npx package resolution failures.

5. **Test & Documentation Coverage**:
   - Unit tests: `tests/unit/bun-support.test.ts` asserting Bun detection and SQLite driver resolution order.
   - Updated `README.md` with Bun deployment and container build instructions.

## Verification

- `npm run typecheck:core` (0 errors)
- `node --import tsx/esm --test tests/unit/bun-support.test.ts` (3/3 passed)
- Local DAST smoke suite run: Schemathesis 7/7 passed, Promptfoo 2/2 passed (100% green)


diegosouzapw:
Solid, well-documented effort — the runtime detection, Webpack-under-Bun switch, update-notifier hardening, and DAST workflow fixes are all reasonable. Two blocking issues, both verified on real Bun 1.3.14:

1. [MANDATORY] driverFactory.ts:224-227 breaks the native bun:sqlite open. `new Database(filePath, {})` throws "bad parameter or other API misuse" on Bun 1.3.14; the pre-change form `new Database(filePath, { readwrite: true, create: ... })` opens fine. As written, every non-readonly open under Bun falls back to sql.js, so "native Bun backend" never actually uses native sqlite. Fix by keeping the explicit options (e.g. `const bunOptions = options?.readonly ? { readonly: true } : { readwrite: true, create: options?.create !== false }`; for `:memory:` skip the empty-object arg). Validate by running `bun test tests/unit/db-adapters/driverFactory.test.ts` under real Bun and asserting the sync driver resolves to `bun:sqlite` (currently it logs "bun:sqlite failed to open" and drops to sql.js). Note tests/unit/bun-support.test.ts:47-88 injects a dummy Database that ignores options, so it never detects this — add a real-Bun case or assert that opening a real in-memory DB via the factory yields driver `bun:sqlite`.

2. [MANDATORY] Dockerfile.bun:89 ENTRYPOINT runs `bun bin/omniroute.mjs serve`, but the standalone output the image copies does not contain `bin/omniroute.mjs` (assembleStandalone.mjs:37 lists the CLI esbuild as prepublish-only; the runner only COPYs `.build/next/standalone`, better-sqlite3, healthcheck.mjs). The container will fail at start. Match the Node Dockerfile's approach (`CMD node dev/run-standalone.mjs`) — `dev/run-standalone.mjs` IS shipped into the standalone (assembleStandalone.mjs:162-164) — i.e. `ENTRYPOINT ["bun", "dev/run-standalone.mjs"]`, and validate a `docker build -f Dockerfile.bun` + run end-to-end.

[OPTIONAL] The PR mixes unrelated cleanup (virtualFactory.ts:610 Number(), `as Record<string, unknown>` casts in src/app/api/v1/moderations/route.ts:71 and src/lib/embeddings/service.ts:332, open-sse-typecheck-baseline.json prune). The baseline prune is legitimate (both pruned files typecheck clean at head) but splitting these into a separate chore PR keeps this one focused on Bun. Also note AGENTS.md frames Bun as best-effort/unsupported — elevating Bun to a first-class supported runtime + README claim is a roadmap decision worth an explicit owner sign-off, and a real `test:bun:db` entry is the repo's required Bun-change gate.

Issue #10404, opened 2026-08-14
fix(providers): command-code/meta/muse-spark-1.2-contributor returns HTTP 200 with content:null and 0 output tokens via /v1/chat/completions when stream=true


yuchi0531:
## Summary

**`command-code/meta/muse-spark-1.2-contributor` returns HTTP 200 with `content: null` and `completion_tokens: 0` (zero output tokens) after processing a large context (56K–97K prompt tokens), causing clients (e.g. opencode) to treat the response as an empty/silent failure.**

There is no answer in `reasoning_content` either, so the `reasoning_content → content` fallback proposed in #8700 **cannot** fix this case.

## Is there an existing issue for this?

- [x] I have searched the existing issues (see #8700 which covers a *different* symptom: `content: null` but the answer trapped in `reasoning_content`)
- [x] This is a new issue: `completion_tokens: 0`, no `reasoning_content`, no `content` — distinct from #8700

## Current Behavior

omniroute 3.8.49, `/v1/chat/completions` with `stream: true`, large input (56K–97K prompt tokens) plus `tools`, takes **~50–60 seconds** and then returns HTTP 200 with `content: null`, `completion_tokens: 0`, and no `reasoning_content`:

```json
{"choices": [{"message": {"role": "assistant", "content": null}, "finish_reason": "stop"}],
 "usage": {"prompt_tokens": 56447, "completion_tokens": 0, "total_tokens": 56447}}
```

On the client side (opencode), this response was treated as `max_tokens` reached (`reason=length`) and the step was cut off without any final text.

## Expected Behavior

A `finish_reason: stop` response with zero output tokens is **not a successful completion**. When `content` is empty and no answer exists:

- (a) automatically retry against the upstream provider, or
- (b) do not return a success (200) — let `validateResponseQuality` mark it as failed so fallback/error handling kicks in.

At minimum, do not hand the client a "successful but empty" response.

## Steps To Reproduce

Occurs under these conditions (reproduction rate is environment-dependent):

- model: `cmd/meta/muse-spark-1.2-contributor`
- input: 56K–97K prompt tokens of context + `tools` definitions
- `stream: true`, `max_tokens` unset (default)

```bash
# Build the input messages (≈56K prompt tokens worth of history)
python3 -c '
import json
msgs=[{"role":"system","content":"You are an expert reviewer. Be exhaustive."}]
for k in range(30):
[...]

PR #7973, opened 2026-07-21
fix(combo): exempt content_filter from empty-content detection

HouMinXi:
## Problem

When Gemini Flash returns a safety-filtered response (finish_reason:
content_filter, empty content), `isEmptyContentResponse()` misclassifies
it as a fake-success empty response and returns HTTP 502.  This triggers
the combo fallback chain and account cooldown escalation (5s → 10s →
20s → 40s), even though the response is a legitimate terminal state.

## Root cause

`errorClassifier.ts` line 14: `LEGIT_EMPTY_OPENAI_FINISH` only exempts
`"length"` and `"tool_calls"`.  The `"content_filter"` finish reason
(mapped from Gemini's `SAFETY`/`PROHIBITED_CONTENT`) is not exempted,
so safety-filtered responses are treated as empty content failures.

## Fix

Add `"content_filter"` to `LEGIT_EMPTY_OPENAI_FINISH` so safety-filtered
responses pass through as valid (though filtered) completions.

## Testing

- 10/10 unit tests pass (`empty-content-stopreason-3572.test.ts`)
  including 2 new content_filter test cases
- E2E: hot-patched OmniRoute v3.8.48 on X500, verified the previously
  failing prompt (4.6KB review) now returns valid content instead of
  empty-content 502

PR #11041, opened 2026-08-21
fix(codex): make remote compaction V2 complete reliably

jackjinke:
## Summary

- Codex remote compaction V2 (`POST /v1/responses` with a `compaction_trigger` input item, codex-rs `compact_remote_v2` shape) completes with `output=[{type:"compaction",encrypted_content:"…"}]` and **no assistant text**.
- The #8649 stream content watcher (`hasUsefulValue` in `open-sse/utils/streamReadiness.ts`) didn't recognize that shape, so a healthy compaction stream got a synthetic `response.failed` (`"Provider returned empty content"`) appended **after** the real `response.completed`.
- Strict Responses clients reject any `response.failed` frame regardless of order (e.g. pi-agent-core `collectCompactionV2Events` throws on it), so V2 compaction through OmniRoute failed after 3 retry attempts and fell back to V1 — each retry was a real upstream compaction call.
- Live V2 tracing exposed two additional stream-path blockers: the Codex SSE filter only split LF-framed events, while the upstream uses CRLF; and a large compaction terminal frame could lose `response.completed` when terminal scan state was trimmed to 4 KiB before inspection.
- Fix: compaction output now counts as real model output, the Codex event filter handles LF/CRLF framing across transport chunks, and terminal events are detected before bounding retained scan state.

## Related Issues

- None found — searched open/closed issues and PRs for compaction / empty-content Responses reports; adjacent work (#7973 content_filter exemption, #10404/#10744 empty-completion failover, #10980 bare-close response.failed) does not cover this shape.

## Validation

- [x] Change type: other (Codex remote compaction V2 stream path)
- [x] Focused V2 and streaming suites: `codex-drop-nonstandard-events`, `empty-stream-no-content-8649`, `stream-readiness`, `stream-readiness-policy`, `stream-handler`, `stream-handler-client-disconnect` — 88/88 pass
- [x] ESLint clean on all changed production and test files
- [x] Rebased onto the current `upstream/release/v3.8.50`
- [x] Production-code changes include automated regression coverage
- [x] Live remote compaction V2 returned a `compaction` item, emitted `response.completed`, and emitted no `response.failed`

## Tests Added Or Updated

- `tests/unit/empty-stream-no-content-8649.test.ts`
  - compaction-only Responses output passes through without a synthetic failure;
  - encrypted-reasoning-only output still trips the empty-content guard.
- `tests/unit/codex-drop-nonstandard-events.test.ts`
  - CRLF-framed SSE split across transport chunks is filtered and forwarded correctly.
- `tests/unit/stream-handler.test.ts`
  - a compaction terminal frame larger than the 4 KiB retained tail is recognized without appending `response.failed`.

## Coverage Notes

- Changed production files: `open-sse/utils/streamReadiness.ts`, `open-sse/executors/codex.ts`, and `open-sse/utils/streamHandler.ts`.
- Coverage gate result: statements 84.06%, branches 79.06%, functions 87.32%, lines 84.06%. The full runner still reports unrelated existing suite failures; all changed-path suites pass.

## Reviewer Notes

- Deliberately **not** a blanket `encrypted_content` key: an encrypted `type:"reasoning"` item alone is not user-visible output, and treating it as content would mask the fake-success shape #8649 exists to catch. Only compaction items qualify.
- Scope remains the Codex remote compaction V2 streaming path. V1 `/v1/responses/compact` and the non-streaming `isEmptyContentResponse` path are unchanged.
- The SSE filter remains streaming and bounded; it now recognizes both LF and CRLF event separators, including separators split across chunks.

## Files (6)

- `open-sse/executors/codex.ts`
- `open-sse/utils/streamHandler.ts`
- `open-sse/utils/streamReadiness.ts`
- `tests/unit/codex-drop-nonstandard-events.test.ts`
- `tests/unit/empty-stream-no-content-8649.test.ts`
- `tests/unit/stream-handler.test.ts`

### DIFF
--- a/open-sse/executors/codex.ts
+++ b/open-sse/executors/codex.ts
@@ -517,10 +517,12 @@ export function filterNonstandardCodexSse(response: Response): Response {
   const transform = new TransformStream<Uint8Array, Uint8Array>({
     transform(chunk, controller) {
       buffer += decoder.decode(chunk, { stream: true });
-      let sep: number;
-      while ((sep = buffer.indexOf("\n\n")) !== -1) {
-        const block = buffer.slice(0, sep + 2);
-        buffer = buffer.slice(sep + 2);
+      while (true) {
+        const separator = /\r?\n\r?\n/.exec(buffer);
+        if (!separator) break;
+        const blockEnd = separator.index + separator[0].length;
+        const block = buffer.slice(0, blockEnd);
+        buffer = buffer.slice(blockEnd);
         if (!dropBlock(block)) controller.enqueue(encoder.encode(block));
       }
     },
--- a/open-sse/utils/streamHandler.ts
+++ b/open-sse/utils/streamHandler.ts
@@ -683,13 +683,15 @@ export function createDisconnectAwareStream(transformStream, streamController) {
     if (clientTerminalSeen) return;
 
     terminalTail += terminalDecoder.decode(chunk, { stream: true });
-    if (terminalTail.length > 4096) {
-      terminalTail = terminalTail.slice(-4096);
-    }
+    // Scan before bounding retained state: a compaction terminal frame can
+    // exceed the tail budget because encrypted_content is carried inline.
     clientTerminalSeen = hasClientTerminalSseMarker(
       terminalTail,
       streamController.clientResponseFormat
     );
+    if (terminalTail.length > 4096) {
+      terminalTail = terminalTail.slice(-4096);
+    }
     if (clientTerminalSeen) {
       streamController.markClientTerminalSeen?.();
     }
--- a/open-sse/utils/streamReadiness.ts
+++ b/open-sse/utils/streamReadiness.ts
@@ -34,6 +34,13 @@ function hasUsefulValue(value: unknown): boolean {
   if (Array.isArray(value)) return value.some(hasUsefulValue);
   if (!isRecord(value)) return false;
 
+  // A Responses compaction item IS the turn's output: remote compaction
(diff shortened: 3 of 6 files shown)


diegosouzapw:
Nice, tight fix — thanks for the clean scoping. This closes a real gap in the #8649 work.

I verified the behavior empirically on the current base: with the change, the compaction-only stream is forwarded intact and no longer gets a synthetic `response.failed`, and the reasoning-only inverse still trips the guard — exactly the intended contract. The deliberate decision to key on `type === "compaction"` + `hasNonEmptyString(value.encrypted_content)` rather than any `encrypted_content` is the right call and well-documented.

No blocking items. A couple of optional notes:
- `open-sse/utils/streamReadiness.ts:39` — harmless, but consider a brief inline note that `compaction` items are only produced by Responses streams (Codex/client), so this branch is unreachable for chat-completion frames; helps future readers judge whether the check is relevant to a given format.
- The two tests cover the contract cleanly (positive + inverse). No further coverage needed.

Approving as merge-ready. [MANDATORY]: none — tests are already included and green (Hard Rule #18 satisfied).

jackjinke:
@diegosouzapw I kept this focused on remote compaction V2. Live tracing found two additional stream-path blockers: CRLF frames were not split by the Codex SSE filter, and a large compaction response.completed frame could be trimmed before terminal detection. Both are fixed with focused regressions; live V2 compaction now returns the compaction item and closes without response.failed.

PR #11042, opened 2026-08-21
feat(redis): add configurable key namespace prefix

MeRezaRezaei:
## Summary
Introduces a single configurable `REDIS_KEY_PREFIX` (default `omniroute:`) that namespaces **every** OmniRoute Redis key, so OmniRoute can safely share a Redis instance with other apps (this instance runs Redis on `127.0.0.1:6379` alongside other services).

- `src/shared/utils/rateLimiter.ts`: reads `REDIS_KEY_PREFIX` and passes it as the ioredis `keyPrefix` on the shared client (rate limiter **and** auth cache).
- `src/lib/quota/redisQuotaStore.ts`: `KEY_PREFIX` now derives from `REDIS_KEY_PREFIX`, defaulting to `omniroute:quota`.
- `src/app/api/local/redis/status/route.ts`: the dashboard status endpoint also reports `running: true` when the native `REDIS_URL` is reachable — not only when a Docker/Podman container exists (the production path used here).
- `.env.example`, `docs/ops/REDIS_PRODUCTION_CONFIG.md`: document the prefix.
- Tests: extends the rate-limiter and quota source-contract tests and adds `tests/unit/local-redis-status.test.ts`.

## Scope
Redis-only, isolated from the kimi consolidation (which is a separate deploy). No provider/routing/DB logic touched.

## Verification
- `npm run lint` — no new errors introduced (all 3 changed source files + test files clean). The 13 `no-explicit-any` errors in `account-rotation-lot-c.test.ts` / `terminal-status-origin.test.ts` are pre-existing on the base branch.
- Contract tests pass: `rate-limiter-redis-optional`, `quota-redis-store`, `local-redis-status`.

⚠️ base-red inherited: #9985 (Release branch not green: release/v3.8.50) — the inherited lint failures above are base defects, not introduced by this PR.


diegosouzapw:
Nice, clean feature — the ioredis `keyPrefix` approach on the shared client plus the manual `quota` prefix is the right way, and the default keeps the current `omniroute:quota` key format so existing deployments are unaffected. The added tests pass and I confirmed they don't break the surrounding suite.

A few things before merge:

MANDATORY — the namespace is not actually applied to ALL Redis keys. There is a third ioredis client at src/lib/warmupScheduler/circuitBreakerFactory.ts:92 (separate from rateLimiter/quota) whose store keys in src/lib/warmupScheduler/redisCircuitBreakerStore.ts:21 are hardcoded to `omniroute:warmup:cb:`. A user setting REDIS_KEY_PREFIX to avoid collisions on a shared Redis will still collide on those `warmup:cb:*` keys, so docs/ops/REDIS_PRODUCTION_CONFIG.md's claim that all keys are prefixed is overstated. Either derive that prefix from REDIS_KEY_PREFIX (add the client keyPrefix there too, with a matching test per Hard Rule #18) or narrow the doc claim.

IMPORTANT/Optional — the src/app/api/local/redis/status/route.ts change is out of scope for a "key namespace prefix" PR and is worth splitting into its own PR. Two follow-ups there: (1) `parseRedisUrl` extracts `host` but `pingRedis` always connects to `127.0.0.1`, so a remote or named-host REDIS_URL is pinged against the wrong host; (2) `running = container.running || redisUrlReachable` reports `running=true` when only a native Redis is reachable (no container), which flips the RedisLauncherPanel (RedisLauncherPanel.tsx:55) into "Stop" mode for a container it never launched.

IMPORTANT/Optional — the new tests are source-contract string assertions (matching the existing style in those files, acceptable here); consider a behavioural assertion for the quota prefix (construct the store, assert the generated key) to guard the actual key format rather than the source text.

Minor — `REDIS_KEY_PREFIX` is not normalized for a trailing `:`. A value like `myco` (no colon) yields `mycorl:*` for the ioredis keyPrefix path but `mycoquota:*` for the quota store — inconsistent. Suggest documenting that a trailing colon is required (or normalizing it).

Thank you for the contribution — the feature is genuinely useful for operators sharing a Redis instance.

Issue #10867, opened 2026-08-20
[BUG] OpenAI API returns null content when model output consists of reasoning tokens (e.g. command-code/meta/muse-spark-1.2-contributor)


Stazyu:
### OmniRoute Version

3.8.50 (next)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Provider(s) Involved

Command-Code

### Model(s) Involved

meta/muse-spark-1.2-contributor

### Description

When making a request to `/v1/chat/completions` using `command-code/meta/muse-spark-1.2-contributor` via the `COMMAND-CODE` provider, the endpoint responds with HTTP 200 OK, but `choices[0].message.content` is set to `null`.

### Steps to Reproduce

1. Configure OmniRoute with provider `COMMAND-CODE`.
2. Send a `POST` request to `/v1/chat/completions` with model `command-code/meta/muse-spark-1.2-contributor`.
3. Pass a simple prompt, e.g.:
   ```json
   {
     "model": "command-code/meta/muse-spark-1.2-contributor",
     "messages": [
       {
         "role": "user",
         "content": "Calculate 79874+93658, and reply with the result only."
       }
     ]
   }

### Expected Behavior

The API should return the generated text response inside choices[0].message.content (or map the reasoning output accordingly) instead of returning null.

### Actual Behavior

The request succeeds with HTTP 200 OK, but choices[0].message.content is set to null because all generated output tokens are categorized as reasoning tokens (Reasoning: 61).

JSON :
```
{
  "choices": [
    {
      "message": {
        "role": "assistant",
        "content": null
      }
    }
  ]
}```

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

### Screenshots

<img width="1035" height="847" alt="Image" src="https://github.com/user-attachments/assets/f1befb6c-ebea-4134-9b34-895ed1273ae7" />

### Additional Context

- **Correlation ID:** `49800973-f5aa-4204-9d11-222b9b98f751`
- **Request ID:** `1787238740599-5eb7b1`
- **Provider:** COMMAND-CODE
- **Requested Model:** `command-code/meta/muse-spark-1.2-contributor`
- **Upstream Model:** `meta/muse-spark-1.2-contributor`
- **Protocol:** OPENAI-CHAT
- **Metrics:** Total Out: 64 | Reasoning: 61 | Content: null

zoser69:
[...]

PR #11043, opened 2026-08-21
fix(sse): split concatenated tool_call arguments from same-name index collisions

### LINKED ISSUE #11044: fix(sse): Parallel tool_calls of the same name silently collapse into one malformed call (SSE index never varies)
### OmniRoute Version

3.8.50 (release/v3.8.50)

### Installation Method

Built from source

### Operating System

Linux

### Provider(s) Involved

zen (observed on opencode/muse-spark-1.2-contributor-free, but the bug is in generic stream reconstruction, not provider-specific — see Additional Context)

### Model(s) Involved

opencode/muse-spark-1.2-contributor-free

### Client Tool

opencode

### Description

When a model emits 2 or more `tool_calls` of the **same function name** in a single streamed turn, some upstreams never vary `index`/`id` across the 2nd/3rd/… call. Every delta lands on the same accumulator key in `buildOpenAISummary` (`open-sse/utils/streamPayloadCollector.ts`), so the calls concatenate into a single malformed `arguments` string containing N back-to-back JSON objects with no separator — invalid as one JSON value, but each object is individually well-formed.

Whatever reads that downstream (a lenient JSON parser, or code that just takes the first parseable prefix) grabs the first object and silently drops the rest. There's no error anywhere in the request/response chain — the client just sees `tool_calls.length === 1` and moves on. If you ask a model to fan out N parallel calls of the same tool against an affected upstream, you get exactly 1 executed, every time, with zero visible failure.

### Steps to Reproduce

1. Use a model/provider combo where the upstream streams multiple `tool_calls` deltas of the same function name without varying `index`/`id` between them (reproduced live on `opencode/muse-spark-1.2-contributor-free` via the `zen` provider).
2. Ask the model, in one turn, to invoke the same tool 3 times in parallel (e.g. dispatch 3 sub-agents).
3. Inspect the reconstructed `tool_calls` array on the resulting chat completion.

### Expected Behavior

3 separate `tool_calls` entries, each with its own `arguments`.

### Actual Behavior

1 `tool_calls` entry whose `arguments` is 3 JSON objects concatenated with no separator (e.g. `{"description":"A",...}{"description":"B",...}{"description":"C",...}`). Confirmed present already in the raw upstream response (not introduced by any OmniRoute translation step) — verified with OmniRoute's `Call Log Pipeline` capture comparing `providerResponse` (raw) and `clientResponse` (post-translation): byte-identical.

### Test Impact

### Additional Context

Root cause is structural, not tied to one provider: `buildOpenAISummary`'s accumulator keys tool_call deltas by `id`/`index` (see `#6276`'s fallback logic in the same function). If an upstream never advances either across repeated same-name calls, every delta resolves to the same key and `arguments` accumulates as N concatenated objects instead of N separate entries.

A fix (`splitConcatenatedToolCallArguments`) is up in #11043: it detects an `arguments` string that fails to parse as one JSON value but scans cleanly as ≥2 back-to-back valid JSON objects/arrays with no separator, and expands it back into separate `tool_calls` entries. Falls through unchanged for empty, single-valid-JSON, or genuinely malformed arguments, so it can't regress any currently-working case.

### Validation Plan

- `node --import tsx/esm --test tests/unit/stream-payload-collector.test.ts`

maxmad64bis:
## Summary

Some providers never bump `index`/`id` when they emit a second or third tool call of the same name in one turn (seen on `opencode/muse-spark-1.2-contributor-free` via the `zen` provider). Every delta lands on the same accumulator key in `buildOpenAISummary`, so 3 separate tool calls get glued into one `arguments` string that looks like `{...}{...}{...}` — not valid JSON, but each chunk is a fine object on its own.

Whatever reads that downstream just grabs the first object and drops the rest, silently. Ask a model to fan out 3 parallel calls against a provider with this bug and you get 1, every time, with no error anywhere in the chain to tell you the other 2 vanished.

Fix: detect when `arguments` is actually N JSON objects stuck together with no separator, and split it back into N real `tool_calls` entries. The fix itself isn't provider-specific — it repairs the shape wherever it shows up.

## Related Issues

- Closes #11044
- Related to #10867 / #10874 (same model family, different bug — those got the model onto the right endpoint so it stops returning null content; this fixes tool calls silently disappearing even once you're on that endpoint).

## Validation

- [x] Change type: routing (`open-sse/utils/streamPayloadCollector.ts`)
- [x] `node --import tsx/esm --test tests/unit/stream-payload-collector.test.ts` — 20/20
- [x] Related suites re-run too (truncated-provider-response, streaming telemetry, stream-utils, request-log-payloads) — 86/86
- [x] `npm run lint` clean — the 2 pre-existing suppressions in these files are untouched, same counts as before
- [x] `tsc --noEmit` — no new errors from this change (one unrelated pre-existing error in a file I didn't touch, same on base)
- [x] Test written first against the real broken payload, confirmed it failed, then fixed

⚠️ base-red inherited: #9985

## Tests Added Or Updated

`tests/unit/stream-payload-collector.test.ts`:
- Reproduces the glued-JSON case → expects 3 clean `tool_calls`
- A normal single-JSON `arguments` string stays untouched (no false positives)
- Genuinely broken (non-concatenated) JSON stays untouched too — not worse than before
- Small unrelated fixup in the same file: two old test fixtures were missing the `index` field their type already required (pure typecheck fix, no behavior change)

## Coverage Notes

Only touches `streamPayloadCollector.ts`, covered by the tests above plus the full existing suite for that file.

## Reviewer Notes

Deliberately conservative: it only splits when `arguments` fails to parse as one JSON value AND cleanly scans as 2+ back-to-back valid objects with nothing between them. Anything else — empty string, valid single JSON, actually-malformed JSON — passes through untouched, so this can't break any currently-working case. Split entries get `${originalId}_split${i}` as their id, just to keep them traceable back to where they came from.

### DIFF
--- a/open-sse/utils/streamPayloadCollector.ts
+++ b/open-sse/utils/streamPayloadCollector.ts
@@ -128,6 +128,75 @@ function tryParseJson(raw: string): unknown {
   }
 }
 
+/**
+ * Splits a tool_call `arguments` string that is actually multiple back-to-back JSON
+ * objects glued together with no separator, into its individual object substrings.
+ *
+ * Root cause (observed on opencode/muse-spark-1.2-contributor-free via the zen
+ * provider): some upstreams never vary `index`/`id` across a 2nd/3rd/… tool_call of
+ * the SAME name emitted in one turn, so every delta in `buildOpenAISummary` above
+ * resolves to the same accumulator key and `arguments` ends up as N JSON objects
+ * concatenated with no delimiter — invalid as a single JSON value, but each object is
+ * individually well-formed. Structural, not provider-specific: applies to whichever
+ * upstream exhibits the same index-collision streaming bug.
+ *
+ * Returns `null` when `raw` is empty, already valid single JSON, or does not scan as
+ * ≥2 back-to-back valid JSON values — callers must leave `arguments` untouched in
+ * that case (never regress a value that used to reach the client as-is).
+ */
+export function splitConcatenatedToolCallArguments(raw: string): string[] | null {
+  if (!raw) return null;
+  try {
+    JSON.parse(raw);
+    return null; // Already a single valid JSON value — nothing to split.
+  } catch {
+    // Fall through to the multi-value scan below.
+  }
+
+  const parts: string[] = [];
+  let depth = 0;
+  let inString = false;
+  let escaped = false;
+  let start = -1;
+
+  for (let i = 0; i < raw.length; i++) {
+    const ch = raw[i];
+    if (start === -1) {
+      if (ch === " " || ch === "\n" || ch === "\r" || ch === "\t") continue;
+      if (ch !== "{" && ch !== "[") return null; // Not a value boundary — bail, leave untouched.
+      start = i;
+    }
+    if (inString) {
+      if (escaped) escaped = false;
+      else if (ch === "\\") escaped = true;
+      else if (ch === '"') inString = false;
+      continue;
+    }
+    if (ch === '"') {
+      inString = true;
+      continue;
+    }
+    if (ch === "{" || ch === "[") depth++;
+    else if (ch === "}" || ch === "]") {
+      depth--;
+      if (depth === 0) {
+        parts.push(raw.slice(start, i + 1));
+        start = -1;
+      }
+    }
+  }
+  if (start !== -1 || depth !== 0 || parts.length < 2) return null;
+
+  for (const part of parts) {
+    try {
+      JSON.parse(part);
+    } catch {
+      return null; // One of the scanned segments isn't valid JSON — bail entirely.
+    }
+  }
+  return parts;
+}
+
 // ─── Per-format live reducers ────────────────────────────────────────────────
 // Each reducer mirrors the corresponding build*Summary()'s original for-loop
 // body exactly (ingest = one loop iteration, finalize = the post-loop return),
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Nice fix — I reviewed the diff and ran it. The scanner in `splitConcatenatedToolCallArguments` (open-sse/utils/streamPayloadCollector.ts:134-204) is carefully defensive: it only splits when the full `arguments` fails `JSON.parse` and scans as >=2 back-to-back valid JSON values, returning `null` (leave untouched) for valid single JSON, bare strings, trailing garbage, and partially-invalid concatenations — so it can't regress values that used to reach the client. All 20 tests in tests/unit/stream-payload-collector.test.ts pass, including your 3 new ones and all existing tool_call tests, and Prettier is clean. Two optional polish items (not blockers): (1) add a direct unit test for `splitConcatenatedToolCallArguments` covering nested objects, braces/escaped quotes inside strings, and top-level arrays — the behavior is correct (I verified it), but a direct test locks it in; (2) finalize now renumbers *every* tool_call to contiguous 0-based indexes (including non-split ones), which is the OpenAI contract and an improvement, but a one-line comment noting that normalization would make the intent explicit for future readers. Thanks for the thorough test coverage on a genuinely tricky streaming edge case.

maxmad64bis:
Thanks — both optional nits addressed in follow-up PR #11135.

**Nit 1 (direct tests):** added 3 direct tests for `splitConcatenatedToolCallArguments(raw: string): string[] | null` (open-sse/utils/streamPayloadCollector.ts) — the previous coverage only hit it indirectly through the tool_call suite:
- two back-to-back JSON objects → returns the array of parts (`[a, b]`);
- a single nested/escaped-object JSON → returns `null` (left untouched);
- a top-level array → returns `null` (one value boundary).

These lock the contract: returns `null` for a single valid JSON value, an array of parts for >=2 back-to-back valid JSON values.

**Nit 2 (normalization comment):** added one line at the `tool_call` index-normalization loop noting indexes are renormalized to contiguous 0-based (OpenAI contract).

No change to the splitter logic itself — pure test + comment hardening. All 23 stream-payload-collector tests pass.

PR #7531, opened 2026-07-17
feat(kimi): sync Code, Web, and Moonshot providers

backryun:
## Summary

- align Kimi Code with the current CLI contract, including K3 discovery, OAuth/API-key authentication under one provider card, protocol-aware routing, thinking metadata, token refresh, and usage handling
- update Kimi Web to the current access-token and Connect-RPC request/stream contract, with a curated `k3`/`k2d6` catalog and no remote model import path
- add Moonshot-specific model metadata and request normalization for Kimi K3, K2.7 Code, and K2.6 while retaining hidden legacy provider aliases for compatibility

## Why

The existing Kimi integrations represented authentication modes as separate visible providers and relied on stale cookie, model-discovery, and generic OpenAI-compatible assumptions. Those assumptions no longer matched the current Kimi Code CLI, Kimi Web client, or Moonshot API contracts.

## Impact

- users see one Kimi Code CLI provider and can choose OAuth or Kimi Code API Key when adding a connection
- Kimi Code routes each discovered model through its declared OpenAI or Anthropic protocol and preserves its supported thinking policy
- Kimi Web exposes only the intentionally supported curated models and no longer presents model import/sync behavior
- Moonshot requests apply model-specific reasoning, sampling, tool-choice, replay, and usage normalization

## Validation

- `npm run lint`
- `npm run typecheck:core`
- `npm run check:cycles`
- 23 changed Node test files: 342 tests passed
- dashboard provider Vitest: 2 tests passed
- provider consistency and build-scope checks passed
- Webpack production build completed and generated all 635 pages (`OMNIROUTE_USE_TURBOPACK=0 npm run build`)
- `git diff --check`

## Local build note

The default Turbopack invocation cannot follow this Windows worktree's junctioned `node_modules` directory because it points outside the worktree filesystem root. The supported Webpack fallback completed successfully; the remaining non-fatal standalone asset-copy warning is caused by the same junction resolving source and destination to the same directory.

## CI Follow-up (2026-07-17)

- Fast Quality Gates initially exceeded the structural/cognitive complexity ratchets at **2063 / 893**.
- Kimi/Moonshot request normalization, runtime metadata, discovery, and managed-import paths were decomposed into focused helpers without changing their routing behavior; the final counts are **2056 / 890**.
- Fixed the dashboard onboarding regression by explicitly typing the new `hiddenFromDashboard` property access. The frozen dashboard error baseline was not widened.
[...]

PR #11045, opened 2026-08-21
fix(kimi): point kimi-web to international www.kimi.ai

MeRezaRezaei:
## Summary
- `kimi.com` is the **China-region** endpoint; from outside China it is not reachable, so the `kimi-web` consumer chat provider must default to the **international** `www.kimi.ai` (Connect-RPC API base URL + top-of-page website link).
- The `kimi-coding` / `kimi-coding-apikey` partnership affiliate links intentionally **stay** on `kimi.com` (that plan is sold in the China region). Only `kimi-web` switches to `kimi.ai`.
- Refreshes the stale comment in `src/lib/providers/validation/webProvidersA.ts` (`kimi.com` -> `kimi.ai`).
- Updates the matching unit tests (`executor-kimi-web`, `kimi-partner-aff-links`, `token-health-check-kimi`, `providerPageHeaderKimiPartnerLink`).

## Scope
8 files, 28 insertions / 38 deletions — kimi-web domain correction only, no unrelated changes.

## Verification (all clean)
- `prettier --check` on all 8 files ✅
- `eslint` on all 8 files (exit 0) ✅
- `typecheck:core` ✅ · `typecheck:noimplicit:core` ✅ · `check:cycles` (no cycles) ✅
- Affected kimi tests pass: 3 node:test files + 1 vitest UI test ✅

## Note
⚠️ **base-red inherited: #9985** — `release/v3.8.50` is red per the nightly release-green check. That is pre-existing and not introduced by this branch; not fixed here per project policy.

### DIFF
--- a/open-sse/config/providers/registry/kimi/web/index.ts
+++ b/open-sse/config/providers/registry/kimi/web/index.ts
@@ -12,10 +12,9 @@ export const kimi_webProvider: RegistryEntry = {
   alias: "kimi-web",
   format: "openai",
   executor: "kimi-web",
-  // International consumer chat — the legacy `kimi.moonshot.cn` domain now
-  // redirects every non-CN visitor to www.kimi.com, which speaks a different
-  // Connect-RPC API. See `open-sse/executors/kimi-web.ts` for the wire format.
-  baseUrl: "https://www.kimi.com",
+  // International consumer chat — Connect-RPC API at www.kimi.ai.
+  // See `open-sse/executors/kimi-web.ts` for the wire format.
+  baseUrl: "https://www.kimi.ai",
   authType: "apikey",
   authHeader: "Authorization",
   // Curated-only catalog. Agent Swarm is excluded because it requires Kimi's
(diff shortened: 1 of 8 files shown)


diegosouzapw:
Thanks for this — the alignment work is clean and the kimi tests pass (25/25) on the head. One note to keep the fix coherent: since the chat executor already targets www.kimi.ai, the consistency sweep is incomplete in two live surfaces that still reference kimi.com:

- src/lib/providers/validation/webProvidersA.ts:43 still probes `https://www.kimi.com/api/user` (and the Origin/Referer at :47-48 plus the error strings at :29/:38/:58/:72 say kimi.com). If www.kimi.ai is the international endpoint, the credential-validation probe and its messages should point there too (ideally reusing getKimiWebBaseUrl()/the KIMI_WEB_BASE_URL override so validation and chat agree).
- open-sse/services/tokenExtractionConfig.ts:192-194 still lists the kimi.com login URL and `.kimi.com` cookie domain for the token-extraction guide.

[MANDATORY] Please update these so the whole kimi-web surface is on one domain, and record a short VPS live check (Hard Rule #18): after logging in at www.kimi.ai, confirm both the Connect-RPC chat endpoint and `GET /api/user` answer on www.kimi.ai rather than redirecting to kimi.com — the prior comments in #7531/#10944 asserted the opposite, so a live confirmation makes the flip trustworthy. Minor/optional: the kimi-web website field drops the `aff=omniroute` attribution (web-cookie.ts:326) — just confirm that removal is intended.

PR #11047, opened 2026-08-21
fix(sse): resolve OpencodeExecutor target format through the provider alias

### LINKED ISSUE #11046: fix(providers): OpenCode Zen models with targetFormat:"openai-responses" get a Responses-shaped body sent to /chat/completions (400, empty message)
### OmniRoute Version

3.8.50 (release/v3.8.50)

### Installation Method

Built from source

### Provider(s) Involved

opencode (free)

### Model(s) Involved

oc/***

### Client Tool

opencode

### Description

#10874 added  targetFormat: "openai-responses" to the Muse Spark entries in the opencode provider registry (open-sse/config/providers/registry/opencode/index.ts). This broke every request to these models: they now get HTTP 400 with an essentially empty response (message: {role:"assistant"}, finish_reason: null).

Root cause: PROVIDER_MODELS is keyed by a provider's public **alias** ("oc" for this provider), not its raw registry id ("opencode"). OpencodeExecutor.execute() (open-sse/executors/opencode.ts) calls getModelTargetFormat(this.provider, input.model) with the raw, un-aliased provider id — so the lookup misses the registry entry entirely and _requestFormat silently falls back to "openai", which buildUrl() maps to /chat/completions.

Meanwhile chatCore's own target-format resolution (resolveChatCoreTargetFormat, handlers/chatCore/targetFormat.ts) *does* alias correctly before calling the same getModelTargetFormat(), so the request **body** still gets translated to the Responses API shape (input:[...] instead of messages:[...]).

Net effect: a Responses-API-shaped body is sent to the Chat Completions URL. Upstream (zen, behind Cloudflare) accepts it as valid JSON but has nothing to answer with, and returns the empty/null response above with HTTP 400.

This had never been triggered before #10874 because no prior model in the opencode provider's registry used targetFormat at all — the alias bug was latent and untested end-to-end (the #10874/#10867 test only asserts the registry field in isolation, never exercises the executor).

### Steps to Reproduce

1. On a build that includes #10874 (registry entry with targetFormat:"openai-responses" for a model on the opencode provider, whose alias differs from its raw id — e.g. oc vs opencode).
2. Send any chat completion request for oc/muse-spark-1.2-contributor-free.

### Expected Behavior

Request routes to /responses with a Responses-shaped body, and returns real content.

### Actual Behavior

Request routes to /chat/completions with a Responses-shaped body (mismatched), upstream returns HTTP 400 with an empty message and finish_reason: null.

### Test Impact

### Additional Context

Fix up in #11047: extracts the alias-resolving lookup into resolveOpencodeTargetFormat(), mirroring resolveChatCoreTargetFormat()'s own aliasing, and uses it in OpencodeExecutor.execute().

### Validation Plan

- node --import tsx/esm --test tests/unit/opencode-target-format-alias-11045.test.ts

maxmad64bis:
## Summary

#10874 added `targetFormat: "openai-responses"` to the Muse Spark entries in the `opencode` provider registry. That broke every request to those models: HTTP 400, empty message, `finish_reason: null`.

Turns out `PROVIDER_MODELS` is keyed by a provider's public alias (`"oc"` for this one), not its raw registry id (`"opencode"`). `OpencodeExecutor.execute()` looks up the target format with the raw id, misses the registry entry, and falls back to `"openai"` — so `buildUrl()` keeps pointing at `/chat/completions`. Meanwhile `chatCore`'s own resolution (which does alias correctly) already switched the request **body** to the Responses API shape. Result: a Responses-shaped body lands on the Chat Completions URL, and upstream has nothing sane to answer with.

This never showed up before because no model in this registry had ever used `targetFormat` — the alias bug was just sitting there, never exercised.

Fix: pulled the alias-resolving lookup into its own function (`resolveOpencodeTargetFormat`), same pattern `chatCore/targetFormat.ts` already uses, and pointed `execute()` at it instead of the raw call.

## Related Issues

- Closes #11046

## Validation

- [x] Change type: routing (`open-sse/executors/opencode.ts`)
- [x] `node --import tsx/esm --test tests/unit/opencode-target-format-alias-11045.test.ts` — 4/4
- [x] Related opencode-executor suites re-run too (noauth-proxy-routing, session-fingerprint-headers, cli-headers-synthesis, proxy-rotation, muse-spark-responses) — 32/32
- [x] `npm run lint` clean — the 2 pre-existing `no-explicit-any` in this file are untouched
- [x] `tsc --noEmit` — no new errors
- [x] Test written first against the actual bug (raw id → wrong format), confirmed failing, then fixed

⚠️ base-red inherited: #9985

## Tests Added Or Updated

`tests/unit/opencode-target-format-alias-11045.test.ts`:
- `resolveOpencodeTargetFormat` resolves the muse-spark entries via the alias
- Falls back to `"openai"` for a model with no registry `targetFormat`
- Falls back to `"openai"` for a provider with no alias mapping
- `OpencodeExecutor.buildUrl` actually returns `/responses` once `_requestFormat` is resolved through the alias (live-bug repro)

## Coverage Notes

Only touches `open-sse/executors/opencode.ts`, covered by the new tests plus the full existing executor suite for this file.

## Reviewer Notes

Same pattern other executors already use (`github.ts`, `default.ts` hardcode their alias as a literal since it's a fixed 1:1 mapping for those providers). This one resolves it dynamically via `PROVIDER_ID_TO_ALIAS` instead of hardcoding `"oc"`, so it keeps working if the alias ever changes.

### DIFF
--- a/open-sse/executors/opencode.ts
+++ b/open-sse/executors/opencode.ts
@@ -1,6 +1,6 @@
 import { BaseExecutor, type ExecuteInput, type ProviderCredentials } from "./base.ts";
 import { PROVIDERS } from "../config/constants.ts";
-import { getModelTargetFormat } from "../config/providerModels.ts";
+import { getModelTargetFormat, PROVIDER_ID_TO_ALIAS } from "../config/providerModels.ts";
 import {
   injectReasoningContentForThinkingModel,
   isThinkingMessageModel,
@@ -125,6 +125,24 @@ export function isPremiumOpencodeModel(model: string, provider: string): boolean
   return !OPENCODE_FREE_MODELS.has(model);
 }
 
+/**
+ * Resolves the registry `targetFormat` for a model, aliasing `provider` first.
+ *
+ * `PROVIDER_MODELS` is keyed by the provider's public ALIAS (e.g. `"oc"`), not its
+ * raw registry id (e.g. `"opencode"`) — mirrors `resolveChatCoreTargetFormat()`
+ * (`handlers/chatCore/targetFormat.ts`), which already aliases before calling
+ * `getModelTargetFormat()`. Calling it with the raw id here made every entry miss
+ * silently (fell through to `"openai"`), while chatCore's own request-body
+ * translation (correctly aliased) still switched to the Responses API shape for
+ * `targetFormat:"openai-responses"` models — sending a Responses-shaped body to
+ * the `/chat/completions` URL this executor's own `buildUrl()` kept selecting.
+ * Exported for testability.
+ */
+export function resolveOpencodeTargetFormat(provider: string, model: string): string {
+  const alias = PROVIDER_ID_TO_ALIAS[provider] || provider;
+  return getModelTargetFormat(alias, model) || "openai";
+}
+
 export class OpencodeExecutor extends BaseExecutor {
   /** Delegates to `isPremiumOpencodeModel`. Exported for testability. */
   static isPremiumModel(model: string, provider: string): boolean {
@@ -193,7 +211,10 @@ export class OpencodeExecutor extends BaseExecutor {
     return pickRotatableAccount(this.accounts, this);
   }
 
-  private markCooldown(account: OpencodeAccountState, kind: "transient" | "terminal" = "transient"): void {
+  private markCooldown(
+    account: OpencodeAccountState,
+    kind: "transient" | "terminal" = "transient"
+  ): void {
     markAccountCooldown(account, kind);
   }
 
@@ -202,7 +223,7 @@ export class OpencodeExecutor extends BaseExecutor {
   }
 
   async execute(input: ExecuteInput) {
-    this._requestFormat = getModelTargetFormat(this.provider, input.model) || "openai";
+    this._requestFormat = resolveOpencodeTargetFormat(this.provider, input.model);
 
     // #8681: Gate premium opencode models behind a usable API key.
     // When the connection is keyless (no apiKey, no accessToken) and the model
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks for the careful diagnosis and the cleanly-typed fix with real regression tests — this is exactly the right area, and the write-up is very clear. One thing to verify before this can land: `getModelTargetFormat()` already resolves the provider alias internally at `open-sse/config/providerModels.ts:175` (`const alias = PROVIDER_ID_TO_ALIAS[aliasOrId] || aliasOrId;` — added by #9994, merged 2026-08-11, which is in the current base). Running against the base tip, the OLD line `getModelTargetFormat(this.provider, input.model)` already returns `"openai-responses"` for `opencode`/muse-spark-1.2*, and `buildUrl()` (opencode.ts:367-371) already picks `/responses`. In other words, the alias wrapper is a no-op today and your regression test passes against pre-existing behavior. So this PR is effectively subsumed by #9994's internal alias. Options: (1) confirm on your side against `origin/release/v3.8.50` that the real #11045/#10874 repro was already fixed, and if so close this to avoid a no-op production change (the code and tests are good — could be resubmitted as a hardening/doc-comment-only PR if you want the regression test committed); or (2) if you still observe the 400 in a running instance, keep the fix but show the failing-before/passing-after on the CURRENT base. Nice work either way.

maxmad64bis:
Verified against `upstream/release/v3.8.50`: `getModelTargetFormat()` already resolves the provider alias internally — `open-sse/config/providerModels.ts:175` does `const alias = PROVIDER_ID_TO_ALIAS[aliasOrId] || aliasOrId;` (added by #9994, 2026-08-11, in the current base). So the alias-wrapper added by this PR is a no-op on the base today, and the #11045/#10874 repro is already fixed upstream.

Since the PR is already merged, this is dead weight but harmless — leaving it in place. The regression test it added is genuinely good and could be resubmitted as a doc-only/hardening PR if a real repro ever surfaces. No revert opened unless you'd prefer to drop the no-op from production.

PR #11049, opened 2026-08-21
fix(providers): route Muse Spark to the Responses API on opencode-zen and opencode-go too

maxmad64bis:
## Summary

#10874 fixed Muse Spark on the `opencode` provider by declaring `targetFormat: "openai-responses"` on its registry entries — the model only answers on the Responses API, not `/chat/completions`. Same upstream family, two more provider registries never got the memo:

- `opencode-zen` — `muse-spark-1.2` had no `targetFormat`; `muse-spark-1.2-contributor-free` wasn't even declared.
- `opencode-go` — all 6 `muse-spark-1.2-contributor*` effort-tier entries had no `targetFormat`.

Both share the same executor (`opencode`) and the same upstream family (`opencode.ai/zen/*`) as the provider #10874 already fixed, so it's the same bug, just three separate copies of the same registry data drifting independently. Mirrored the same declaration onto both.

Left `freebuff`'s `meta/muse-spark-1.2-contributor` alone — **verified**: distinct backend (`codebuff.com`, dedicated `freebuff` executor posting to `/chat/completions`, no Responses API), so the same fix doesn't apply there.

## Related Issues

- Closes #11048

## Validation

- [x] Change type: routing (provider registry — `open-sse/config/providers/registry/opencode/{zen,go}/index.ts`)
- [x] `node --import tsx/esm --test tests/unit/opencode-zen-muse-spark-targetformat-11046.test.ts` — 2/2
- [x] Existing suites for these two registries re-run too (`opencode-zen-reasoning-effort`, `opencode-zen-alias-combo-e2e`, `auth-opencode-zen-noauth-fallback`) — 18/18 total
- [x] `npm run lint` clean
- [x] Branched from current `release/v3.8.50` tip
- [x] Test written first against the real registry data, confirmed failing, then fixed

⚠️ base-red inherited: #9985

## Tests Added Or Updated

`tests/unit/opencode-zen-muse-spark-targetformat-11046.test.ts`:
- `opencode-zen`'s `muse-spark-1.2` and `muse-spark-1.2-contributor-free` both resolve to `targetFormat: "openai-responses"`
- All 6 `opencode-go` muse-spark effort-tier entries do too

## Coverage Notes

Registry data only — no logic changes, covered by the assertions above plus the full existing test suite for both registries.

## Reviewer Notes

[...]

Issue #11048, opened 2026-08-21
fix(providers): Muse Spark still returns empty content on opencode-zen and opencode-go (targetFormat missed by #10874)


maxmad64bis:
### OmniRoute Version

3.8.50 (release/v3.8.50)

### Installation Method

Built from source

### Operating System

Linux

### Provider(s) Involved

opencode-zen, opencode-go

### Model(s) Involved

opencode-zen/muse-spark-1.2, opencode-zen/muse-spark-1.2-contributor-free, opencode-go/muse-spark-1.2-contributor (and its -minimal/-low/-medium/-high/-xhigh effort variants)

### Client Tool

opencode

### Description

`#10874` fixed `oc/muse-spark-1.2` and `oc/muse-spark-1.2-contributor-free` on the `opencode` provider by declaring `targetFormat: "openai-responses"` on their registry entries, since Muse Spark is only served on the Responses API, not `/chat/completions`.

There are two more registry entries for the same upstream family (`opencode.ai/zen/*`, same `executor: "opencode"`) that never got the same fix:

- `opencode-zen` (`open-sse/config/providers/registry/opencode/zen/index.ts`) — `muse-spark-1.2` has no `targetFormat` at all; `muse-spark-1.2-contributor-free` isn't even declared (falls through to bare passthrough).
- `opencode-go` (`open-sse/config/providers/registry/opencode/go/index.ts`) — all 6 `muse-spark-1.2-contributor*` effort-tier entries have no `targetFormat`.

Requests routed to either of these still hit `/chat/completions` and the upstream returns the same degenerate empty message described in #10867/#10874.

(`freebuff`'s `meta/muse-spark-1.2-contributor` entry is a different upstream — `codebuff.com`, not `opencode.ai` — and is not affected; no evidence checked for it here.)

### Steps to Reproduce

1. Send a request for `opencode-zen/muse-spark-1.2` (or `opencode-go/muse-spark-1.2-contributor`).
2. Observe the response.

### Expected Behavior

Routes to `/responses`, returns real content — same as the fix already applied to the `opencode` provider's own entries.

### Actual Behavior

Routes to `/chat/completions`, upstream returns `{"choices":[{"message":{"role":"assistant"},"finish_reason":null}]}` with HTTP 400/502 depending on retry state.

### Test Impact

Existing automated test already fails

### Additional Context

[...]

Issue #8353, opened 2026-07-24
feat(providers): add missing opencode-go reasoning variants from OpenCode model registry


excessivechaos:
## Summary

The OpenCode Go model registry exposed by `opencode models opencode-go --verbose` contains reasoning variants that OmniRoute does not currently register or resolve.

This is separate from the provider `/v1/models` discovery metadata: OpenCode's local registry defines these variants as request options, while the upstream model listing omits them.

## Missing variants

- `deepseek-v4-flash-high`
- `deepseek-v4-flash-max`
- `grok-4.5-low`
- `grok-4.5-medium`
- `grok-4.5-high`
- `hy3-none`
- `hy3-low`
- `hy3-high`
- `kimi-k3-max`
- `qwen3.6-plus-high`
- `qwen3.6-plus-max`
- `qwen3.7-max-high`
- `qwen3.7-max-max`
- `qwen3.7-plus-high`
- `qwen3.7-plus-max`

Existing OmniRoute aliases for DeepSeek V4 Pro, GLM-5.2, and MiMo-V2.5 should remain supported.

## Required changes

1. Add the base models and aliases to `open-sse/config/providers/registry/opencode/go/index.ts`.
2. Add the corresponding base-model effort tables to `open-sse/executors/opencode.ts` so each alias rewrites to the base model and injects the correct `reasoning_effort`.
3. Add regression tests covering catalog exposure and request-time alias resolution.

## MiniMax M3

MiniMax M3 is intentionally excluded from this issue. OpenCode defines its variants as request-shape options rather than ordinary effort values:

- `none` -> `thinking: { type: "disabled" }`
- `thinking` -> `thinking: { type: "adaptive" }`
- default -> omit `thinking`

M3 needs separate executor support because mapping `minimax-m3-thinking` to `reasoning_effort: "thinking"` would be incorrect. The provider-side default when `thinking` is omitted also needs confirmation first.

## Acceptance criteria

- Every listed alias appears in the OpenCode Go catalog and Combo model selection.
- Each alias is stripped to its canonical base model at request time.
- The matching `reasoning_effort` value is sent upstream.
- Unsupported aliases are not synthesized for OpenCode Zen.
- MiniMax M3 remains unchanged until its thinking-mode mapping is implemented deliberately.

The source comparison was made against the current output of `opencode models opencode-go --verbose` and the static OpenCode Go registry.

diegosouzapw:
[...]

PR #11051, opened 2026-08-21
refactor(providers): dedupe identical opencode-zen/opencode-go model entries

maxmad64bis:
## Summary

`opencode-zen` and `opencode-go` are two separate registry entries for the same upstream family (`opencode.ai/zen/*`), and three models were declared byte-for-byte identically in both: `kimi-k2.7-code`, `qwen3.5-plus`, `qwen3.6-plus`. Copy-pasted metadata like this is exactly how #10874 missed `opencode-zen`/`opencode-go` when it fixed Muse Spark's `targetFormat` only on the base `opencode` provider.

Extracted the three identical entries into `OPENCODE_ZEN_GO_SHARED_MODELS`, following the same pattern already used for `GLM_SHARED_MODELS` in `open-sse/config/glmProvider.ts` (shared across the `glm`/`glm-t`/`glm-cn` registries). Both `opencode-zen` and `opencode-go` now spread this array instead of repeating the entries inline. A future metadata fix (routing, reasoning flags, etc.) for any of these three models only needs to land in one file.

Scope is intentionally narrow: only entries that were **already byte-identical** across both files were moved. Models with tier-specific flags (e.g. `opencode-go`'s effort-tier variants, or fields only one side declares) were left untouched — merging those would be a silent behavior change, not a dedupe. `command-code` and `freebuff` also list similarly-named models but are different upstream/executor pairs entirely — not touched here.

## Related Issues

- Related to #11048 / #11049 (that PR's Reviewer Notes flagged this exact duplication as a likely follow-up)
- No dedicated bug issue — this is a proactive refactor, not a user-facing fix

## Validation

- [x] Change type: provider (registry data — `open-sse/config/providers/registry/opencode/{zen,go}/index.ts`, `open-sse/config/opencodeZenGoSharedModels.ts`)
- [x] `node --import tsx/esm --test tests/unit/opencode-zen-go-shared-models.test.ts` — 3/3
- [x] `npm run typecheck:core` — clean
- [x] `npm run check:cycles` — no cycles
- [x] `npm run check:provider-consistency` — OK, 267 registry entries, 348 canonical providers, 0 exceptions

⚠️ base-red inherited: #9985

## Tests Added Or Updated

`tests/unit/opencode-zen-go-shared-models.test.ts`:
- Every `OPENCODE_ZEN_GO_SHARED_MODELS` entry is present exactly once, unmodified, in `opencode-zen`
- Every `OPENCODE_ZEN_GO_SHARED_MODELS` entry is present exactly once, unmodified, in `opencode-go`
- The shared array is frozen (guards against one registry mutating it and silently affecting the other)

## Coverage Notes

Registry data only — no logic changes. Covered by the new test plus the full existing suites for both registries and `check:provider-consistency`.

## Reviewer Notes

Pure move, no behavior change: same 3 model ids, same field values, same position in each registry's live catalog. The `check:provider-consistency` gate (267 registry entries, 348 canonical providers, 0 exceptions) and the deep-equal assertions in the new test are what actually prove nothing drifted during the extraction.

### DIFF
--- a/open-sse/config/opencodeZenGoSharedModels.ts
+++ b/open-sse/config/opencodeZenGoSharedModels.ts
@@ -1,0 +1,16 @@
+/**
+ * Models declared identically in both the `opencode-zen` and `opencode-go` provider
+ * registries (same upstream family, opencode.ai/zen/*). Mirrors the GLM_SHARED_MODELS
+ * pattern in glmProvider.ts: one array, spread into each sibling RegistryEntry, so a
+ * metadata fix (targetFormat, supportsReasoning, ...) only has to land in one file
+ * instead of drifting out of sync across registries.
+ *
+ * Only entries that are byte-identical across both registries belong here — a model
+ * with tier-specific flags (e.g. go's effort variants, or a flag only one tier needs)
+ * stays local to that registry's own `models` array.
+ */
+export const OPENCODE_ZEN_GO_SHARED_MODELS = Object.freeze([
+  { id: "kimi-k2.7-code", name: "Kimi K2.7 Code" },
+  { id: "qwen3.5-plus", name: "Qwen3.5 Plus", targetFormat: "claude", supportsVision: false },
+  { id: "qwen3.6-plus", name: "Qwen3.6 Plus", targetFormat: "claude", supportsVision: false },
+]);
--- a/open-sse/config/providers/registry/opencode/go/index.ts
+++ b/open-sse/config/providers/registry/opencode/go/index.ts
@@ -1,4 +1,5 @@
 import type { RegistryEntry } from "../../../shared.ts";
+import { OPENCODE_ZEN_GO_SHARED_MODELS } from "../../../shared.ts";
 
 export const opencode_goProvider: RegistryEntry = {
   id: "opencode-go",
@@ -13,6 +14,8 @@ export const opencode_goProvider: RegistryEntry = {
   authPrefix: "Bearer",
   defaultContextLength: 200000,
   models: [
+    ...OPENCODE_ZEN_GO_SHARED_MODELS,
+
     // Port from decolua/9router 8efacc11: align with official Go endpoints —
     // glm-5.2 is now advertised and Kimi chat traffic must route through
     // `kimi-k2.7-code` (the live API rejects the plain `kimi-k2.7` alias for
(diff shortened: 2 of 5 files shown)


diegosouzapw:
Nice, clean refactor — this matches the GLM_SHARED_MODELS pattern exactly and the regression test is a good guard. I verified locally that no model is lost: the model-id set for both providers is identical before/after, check:provider-consistency still passes (267/0), and the existing #8353 effort-alias test still passes 38/38.

Two optional notes (non-blocking):
1. open-sse/config/providers/registry/opencode/go/index.ts:17 and .../zen/index.ts:19 — the shared spread is placed at the TOP of `models`, so the array order changes. go[0] becomes kimi-k2.7-code (was glm-5.2) and zen[0] becomes kimi-k2.7-code (was big-pickle). Since the dashboard picks `models[0]` as the default (e.g. LlmChatCard.tsx:178, ProviderTestSlideOver.tsx:63), this silently changes each provider's default selected model. If you want a byte-identical refactor, insert the spread where the three entries originally sat (after the GLM block in go, after kimi in zen) — or explicitly call out the new default as intended.
2. tests/unit/opencode-zen-go-shared-models.test.ts only asserts the SHARED entries are present/unmodified; it wouldn't catch a future accidental deletion of a non-shared entry (e.g. minimax-m3, glm-5.1). Consider a broader assertion that the full model-id set matches a fixed "golden" list for each provider.

No blocking changes — happy to merge as-is.

maxmad64bis:
Thanks — addressed in follow-up PR #11133.

**Default-model nit:** confirmed on `upstream/release/v3.8.50` the `OPENCODE_ZEN_GO_SHARED_MODELS` spread is at `models[0]` for both `opencode-go` and `opencode-zen`, so the dashboard default had silently become `kimi-k2.7-code` (the dashboard takes `models[0]`, ~14 call sites). Moved the spread after `glm-5.2` (go) / `big-pickle` (zen) to restore the original `models[0]`. I did **not** try to reproduce the exact pre-#11051 interleaving — the three deduplicated entries were scattered in non-contiguous slots, so a single spread can't restore them and it isn't needed; the goal is the `models[0]` default.

**Golden-list nit:** instead of freezing the full model-id catalog (which `zen/index.ts` documents as live-regenerated via `passthroughModels: true` — a snapshot would red on every legitimate sync), I added two narrow guards: (a) referenced non-shared ids `minimax-m3` / `glm-5.1` must remain present, (b) `models[0].id` must stay `glm-5.2` / `big-pickle`. These lock the real invariants the nit is about.

Also folded in the #11049 overlay comment (zen `muse-spark-1.2-contributor-free` marked as an explicit wire-format overlay of the base provider to avoid catalog-sync drift). `check:provider-consistency` stays 267/0.

Issue #11032, opened 2026-08-21
[bug] Generated opencode.jsonc omits limit.output, fails OpenCode config validation


diegosouzapw:
## Bug Report

**Source:** Discussion #10938 by @WUIX69

## Description

OpenCode rejects OmniRoute's generated `opencode.jsonc` with `Missing key provider.omniroute.models.{model}.limit.output` for multiple models (kiro/deepseek-3.2, kiro/minimax-m2.1, kiro/qwen3-coder-next, opencode-zen/muse-spark-1.2-contributor-free, opencode-zen/hy3-free, opencode-zen/nemotron-3-ultra-free, opencode-zen/nemotron-3.5, and others).

## Root Cause

`src/shared/services/opencodeConfig.ts:63` builds each model entry as:

```ts
modelsRecord[m] = { name: getModelEntryName(m, normalizedLabels) };
```

Only `name` is written. OpenCode's config schema now validates `limit.output` (and likely `limit.context`) as required fields for custom-provider model entries, so any config OmniRoute generates for a model without those fields fails OpenCode's own validation on load.

## Proposed Fix

Emit `limit: { output: <n>, context: <n> }` per model entry, sourced from the model's catalog metadata (`open-sse/config/providers/registry/**`) where available. Needs a sensible fallback when a model's catalog entry doesn't carry explicit context/output limits.

## Steps to Reproduce

1. Configure any provider connection whose registry-catalog models lack `limit.output`/`limit.context` metadata (e.g. `opencode-zen`, `kiro`).
2. Apply the OpenCode CLI-code tool config from OmniRoute's dashboard.
3. Run `opencode` -- config validation fails with the missing-key error shown above.

## Additional Context

- Discussion: #10938

diegosouzapw:
Fixed on the current release branch by #11054 (commit 8643e0f): every model entry in the generated OpenCode config now carries a limit block — limit.output defaults to 8192 and limit.context to 128k when catalog metadata is unknown, preferring catalog or your existing manual values when known — and #11079 adds OpenCode V2 config-format support on top. Please upgrade to the next release and regenerate the config; OpenCode should load it without the 'Missing key …limit.output' validation errors.

PR #11054, opened 2026-08-21
fix(cli): default limit.context to 128k when unknown in OpenCode configs (#11035, #11032)

### LINKED ISSUE #11035: fix(cli): #11010 emits limit.output without limit.context — OpenCode rejects every model whose catalog entry has no context_length
### OmniRoute Version

`release/v3.8.50` (after #11010 merged; not yet on npm — `latest` is still 3.8.49)

### Installation Method

npm (global)

### Operating System

macOS

### OS Version

Darwin 25.5.0 (Apple Silicon)

### Node.js Version

v26.7.0

### Provider(s) Involved

Any — the trigger is a catalog entry without `context_length`

### Model(s) Involved

Any model whose `/v1/models` entry has no `context_length` (58 of 1658 in my local catalog)

### Client Tool

OpenCode 1.18.18

### Description

#11010 fixed `limit.output` by always emitting it. `limit.context` stayed conditional, so `buildModelEntry()` now produces a `limit` block with `output` and no `context` for every model whose catalog entry lacks `context_length` — and OpenCode rejects that with the mirror of the original error, `Missing key ….limit.context`.

Before #11010 those models were emitted without a `limit` block at all, which OpenCode accepts. The fix therefore turns previously-valid entries into config-breaking ones.

```ts
// src/lib/cli-helper/config-generator/opencode.ts, post-#11010
const limit: { context?: number; input?: number; output?: number } = { output };
if (typeof context === "number") limit.context = context;   // still conditional
```

`tests/unit/opencode-limit-output-10940.test.ts:58` pins this shape:

```ts
assert.strictEqual(entry.limit.context, undefined);
```

so the metadata-less fixture the PR added is itself a config OpenCode refuses to load.

### Steps to Reproduce

I ran the merged generator directly against a synthetic catalog rather than reasoning about it. Catalog served on `127.0.0.1:20998`:

```json
{"data": [
  {"id": "probe/ctx-only", "owned_by": "probe", "context_length": 128000},
  {"id": "probe/out-only", "owned_by": "probe", "max_output_tokens": 4096},
  {"id": "probe/no-meta",  "owned_by": "probe"},
  {"id": "probe/complete", "owned_by": "probe", "context_length": 200000, "max_output_tokens": 8192}
]}
```

`generateOpencodeConfig({ baseUrl, apiKey, providerId: "omniroute" })` from `release/v3.8.50` emits:

```
probe/ctx-only     {"output":8192,"context":128000}     <- #11010 works
probe/out-only     {"output":4096}                      <- no context
probe/no-meta      {"output":8192}                      <- no context
probe/complete     {"output":8192,"context":200000}
```

Feeding the last two to OpenCode 1.18.18:

```
Error: Configuration is invalid at ~/.config/opencode/opencode.json
↳ Missing key provider.omniroute.models.probe/out-only.limit.context
↳ Missing key provider.omniroute.models.probe/no-meta.limit.context
```

### Expected Behavior

A generated config loads. Since OpenCode treats `limit` as all-or-nothing, the block should carry `context` and `output` together or be omitted entirely.

### Actual Behavior

Every model without a catalog `context_length` produces a half-filled block and aborts the whole config. On my catalog that is 58 of 1658 models; 55 of them produced no `limit` block before #11010 and loaded fine, so this is a regression for those.

### Test Impact

`tests/unit/opencode-limit-output-10940.test.ts:58` needs inverting: the metadata-less fixture must produce a config OpenCode accepts, i.e. either both keys or no `limit` at all. A generator-level assertion covers both directions cheaply — for every emitted entry, `("context" in limit) === ("output" in limit)`.

### Error Logs / Output

```shell
↳ Missing key provider.omniroute.models.probe/out-only.limit.context
↳ Missing key provider.omniroute.models.probe/no-meta.limit.context
```

### Screenshots

_No response_

### Additional Context

Two ways out, both consistent with the existing comments in that file:

1. Omit the whole `limit` block when `context` is unknown. The code already notes that "OpenCode v1 defaults to 128K when `limit.context` is missing", which is exactly what an omitted block yields — at the cost of the 8K output cap #11010 wanted.
2. Default `context` the way `output` is defaulted. This contradicts the file's stated "we never fabricate a context window" principle, so option 1 looks closer to the intent.

Context for how this surfaced: #10940, and my comment there listing the three `codex/gpt-5.6-*` entries that already hit this on 3.8.49.

### Validation Plan

Run `omniroute setup-opencode` against a catalog containing at least one entry without `context_length`, then `opencode models`. It must list models instead of reporting an invalid configuration.

rqzbeh:
## Summary
Fixes #11035 and #11032 where OpenCode config generators emitted model  objects with  but omitted  when a model had no catalog  metadata. OpenCode 1.18.18's Zod provider schema requires  whenever  is present, causing  to reject generated configs with .

### Changes
1. **Config Generator ()**: Defaults  to  (OpenCode's standard context fallback) when  is absent from catalog and user overrides.
2. **Provider Config Service ()**: Emits  per model entry.
3. **Tests (, )**: Updated assertions and verified test suite passes clean.

### DIFF
--- a/config/quality/dashboard-typecheck-baseline.json
+++ b/config/quality/dashboard-typecheck-baseline.json
@@ -1,9 +1,6 @@
 {
-  "open-sse/services/payloadRules.ts": {
-    "TS2677": 1
-  },
   "src/app/(dashboard)/dashboard/HomePageClient.tsx": {
-    "TS2339": 16
+    "TS2339": 10
   },
   "src/app/(dashboard)/dashboard/agent-skills/AgentSkillsPageClient.tsx": {
     "TS2503": 3
@@ -120,10 +117,6 @@
   "src/app/(dashboard)/dashboard/providers/[id]/components/CompatibleModelsSection.tsx": {
     "TS2741": 1
   },
-  "src/app/(dashboard)/dashboard/providers/[id]/components/ConnectionRow.tsx": {
-    "TS2345": 3,
-    "TS2322": 1
-  },
   "src/app/(dashboard)/dashboard/providers/[id]/components/ConnectionsListPanel.tsx": {
     "TS2322": 2
   },
@@ -140,12 +133,6 @@
   },
   "src/app/(dashboard)/dashboard/providers/[id]/components/ProviderPlaygroundPanel.tsx": {
     "TS2503": 1
-  },
-  "src/app/(dashboard)/dashboard/providers/[id]/components/modals/EditConnectionModal.tsx": {
-    "TS2322": 1
-  },
-  "src/app/(dashboard)/dashboard/providers/[id]/hooks/useModelImportHandlers.ts": {
-    "TS2339": 1
   },
   "src/app/(dashboard)/dashboard/providers/[id]/hooks/useModelVisibilityHandlers.ts": {
     "TS2339": 15
@@ -190,9 +177,6 @@
   "src/lib/combos/builderDraft.ts": {
     "TS2741": 1
   },
-  "src/lib/providers/codexFastTier.ts": {
-    "TS2367": 1
-  },
   "src/lib/services/htmlRewriter.ts": {
     "TS2322": 2,
     "TS2345": 2
@@ -219,14 +203,7 @@
   "src/shared/hooks/useElectron.ts": {
     "TS2339": 19
   },
-  "src/shared/providers/webSessionCredentials.ts": {
-    "TS2353": 1,
-    "TS2322": 1
-  },
   "src/shared/schemas/cliCatalog.ts": {
     "TS2554": 2
-  },
-  "src/shared/services/opencodeConfig.ts": {
-    "TS2345": 1
   }
 }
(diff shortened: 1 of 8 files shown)


diegosouzapw:
Nice, focused fix — emitting limit.context fixes the real OpenCode v1 schema error from #11035/#11032, and you kept the right precedence: user limit.context override → live catalog context_length → 128k fallback, so you don't fabricate a value whenever one is genuinely known. I verified on this branch that an unknown model gets `{context:128000, output:8192}` while a catalog context_length of 200000 is preserved (`{context:200000}`), and both the 10940 and t40 tests pass.

One thing to clean up [OPTIONAL]: the jsdoc in opencode.ts still describes the OLD behavior and now contradicts the code — src/lib/cli-helper/config-generator/opencode.ts:285-297 (“We do NOT synthesize a default … limit.context will simply omit”) and the generateOpencodeConfig doc at :434-439 (“emitted WITHOUT a limit.context field”), plus the “never invent defaults” lines at :410-419. Updating those to state the new 128k fallback would keep the file internally consistent; otherwise future maintainers reading the doc will be misled. Thanks for the good test coverage.

PR #11055, opened 2026-08-21
fix(providers): update hailuo-web domain to chat.minimax.io (#11000)

### LINKED ISSUE #11000: fix(providers): hailuo-web token validation fails - hailuo.ai/models now 301s to chat.minimax.io
### Environment

- OmniRoute **v3.8.49** (dashboard)
- Provider: `hailuo-web` (Hailuo / MiniMax web), catalog entry at
  `src/shared/constants/providers/web-cookie.ts:318`
- Connection dialog → **Check token**

### What breaks

Token validation probes `${website}/models`. The catalog's website is
`https://www.hailuo.ai`, but that host now 301-redirects to MiniMax's chat app:

```
GET https://www.hailuo.ai/models  →  301  →  https://chat.minimax.io/models
GET https://www.hailuo.ai/        →  301  →  https://chat.minimax.io/
```

The validator treats the redirect as a failure, so every check reports:

> **Invalid — Redirect blocked for GET https://www.hailuo.ai/models (301)**

regardless of whether the pasted `_token` cookie is actually valid.

### Context

- The Hailuo consumer chat product has been folded into `chat.minimax.io`; the old
  `hailuo.ai` marketing domain is just a redirect shell now.
- Same probe pattern as closed issue #7857 (web-cookie validators hitting
  `${baseUrl}/models`), which was fixed for other providers but `hailuo-web` still
  points at the pre-migration domain.
- No open duplicate found for hailuo-web specifically.

### Suggested fix

Pick one:

1. Update the `hailuo-web` catalog `website` to `https://chat.minimax.io` so the
   `/models` probe stays same-origin, or
2. Allow-list the `hailuo.ai → chat.minimax.io` redirect in the validator, or
3. Probe a stable authenticated endpoint on the current origin instead of
   `/models`.

rqzbeh:
Fixes #11000 by updating hailuo-web website and baseUrl to chat.minimax.io to avoid 301 redirects during key validation probes.

### DIFF
--- a/config/quality/dashboard-typecheck-baseline.json
+++ b/config/quality/dashboard-typecheck-baseline.json
@@ -1,9 +1,6 @@
 {
-  "open-sse/services/payloadRules.ts": {
-    "TS2677": 1
-  },
   "src/app/(dashboard)/dashboard/HomePageClient.tsx": {
-    "TS2339": 16
+    "TS2339": 10
   },
   "src/app/(dashboard)/dashboard/agent-skills/AgentSkillsPageClient.tsx": {
     "TS2503": 3
@@ -120,10 +117,6 @@
   "src/app/(dashboard)/dashboard/providers/[id]/components/CompatibleModelsSection.tsx": {
     "TS2741": 1
   },
-  "src/app/(dashboard)/dashboard/providers/[id]/components/ConnectionRow.tsx": {
-    "TS2345": 3,
-    "TS2322": 1
-  },
   "src/app/(dashboard)/dashboard/providers/[id]/components/ConnectionsListPanel.tsx": {
     "TS2322": 2
   },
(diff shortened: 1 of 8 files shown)


diegosouzapw:
Great, targeted fix — updating the registry baseUrl, the executor BASE_URL and the web-cookie `website` together keeps the provider consistent, and switching hosts is signature-safe since the yy header signs the relative path, not the host. One thing must land before merge: the committed golden snapshot still pins the old host. The registry change makes buildProviderUrl return `https://chat.minimax.io` for hailuo-web, while `tests/snapshots/provider/translate-path.json:2829-2830` still records `https://www.hailuo.ai`, so `provider-translate-path-golden.test.ts` ("GOLDEN provider.ts translate-path is stable across all providers") fails on your branch (I ran it: expected www.hailuo.ai, got chat.minimax.io). [MANDATORY] Regenerate it (`UPDATE_GOLDEN=1 node --import tsx/esm --test tests/unit/provider-translate-path-golden.test.ts`, or just edit those two lines) so the suite is green. [OPTIONAL] The header comment in `open-sse/executors/hailuo-web.ts:2` and `:8` still name www.hailuo.ai as the endpoint, which is now misleading; and the auto-generated `docs/reference/PROVIDER_REFERENCE.md:101` still shows the hailuo.ai link — run `npm run gen:provider-reference` to keep docs in sync. Good scope for a small PR.

PR #11056, opened 2026-08-21
feat(dashboard): trigger key validation on Enter in AddApiKeyModal (#10995)

### LINKED ISSUE #10995: feat(dashboard): Enter in Add API Key credential field does not submit or check the key
### Summary

In the **Add connection** dialog (e.g. Providers → Aion Labs → Add Connection →
"Add Aion Labs API Key"), pressing **Enter** inside the API Key / session token input
does nothing. The user must click **Check** / **Save** with the mouse. This breaks the
standard paste-key → hit-Enter flow and is especially annoying when adding keys to
several providers in a row.

### Steps to reproduce

1. Dashboard → Providers → any provider (API-key or web-session) → **Add Connection**
2. Type or paste a value into the **API Key** field
3. Press **Enter**
4. Nothing happens — no validation, no save

### Expected

Enter submits: runs the same action as the adjacent **Check** button (and **Save** when
no check is needed).

### Location

`src/app/(dashboard)/dashboard/providers/[id]/components/modals/AddApiKeyModal.tsx`
— the credential `Input` (~line 707) wires only `onChange`; there is no `onKeyDown`
handler and the modal body is not wrapped in a `<form>`, so Enter is a no-op.

Note the sibling modal already has the right pattern:
`EditConnectionModal.tsx` (~line 1214) attaches an `onKeyDown` Enter handler to its
"add another API key" input.

### Suggested fix

Wrap the dialog fields in a `<form onSubmit={...}>` (with the Check/Save buttons as
`type="submit"`), or add an `onKeyDown` Enter handler mirroring `EditConnectionModal`.
Apply to both the **Single** and **Bulk Add** tabs.

rqzbeh:
Fixes #10995 by attaching an onKeyDown Enter handler to the API Key credential input in AddApiKeyModal, allowing users to validate keys by hitting Enter.

### DIFF
--- a/config/quality/dashboard-typecheck-baseline.json
+++ b/config/quality/dashboard-typecheck-baseline.json
@@ -1,9 +1,6 @@
 {
-  "open-sse/services/payloadRules.ts": {
-    "TS2677": 1
-  },
   "src/app/(dashboard)/dashboard/HomePageClient.tsx": {
-    "TS2339": 16
+    "TS2339": 10
   },
   "src/app/(dashboard)/dashboard/agent-skills/AgentSkillsPageClient.tsx": {
     "TS2503": 3
@@ -120,10 +117,6 @@
   "src/app/(dashboard)/dashboard/providers/[id]/components/CompatibleModelsSection.tsx": {
     "TS2741": 1
   },
-  "src/app/(dashboard)/dashboard/providers/[id]/components/ConnectionRow.tsx": {
-    "TS2345": 3,
-    "TS2322": 1
-  },
   "src/app/(dashboard)/dashboard/providers/[id]/components/ConnectionsListPanel.tsx": {
     "TS2322": 2
   },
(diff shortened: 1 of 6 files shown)


diegosouzapw:
Nice, small UX win — Enter-to-validate is the expected form behavior and wiring into the existing handleValidate()/validating/saving state is exactly right. I ran the new test (passes) and the existing modal component tests against it (all green), so this is safe to merge. Two optional polish notes, not blockers: (1) run `npx prettier --write` on the new test — the final assert.ok(...) wraps when it fits on one line; and (2) the Enter handler only guards on `!validating && !saving`, while the Check button also disables on empty key / Google-PSE-missing-cx (AddApiKeyModal.tsx:774). Mirroring that last condition would avoid a pointless validation request when the field is empty. Since this is a UI-only change, the source-string test is an acceptable guard, but a small React render test (like the other add-api-key-modal-*.test.tsx) simulating a keydown Enter would be more robust against future refactors. Great work.

PR #11069, opened 2026-08-21
fix(m365): BizChat invocation shape drift + HAR-import UX + Antigravity alias note

eenggar-svg:
## Summary

Three related fixes/improvements for the individual/consumer M365 Copilot (BizChat) path, found while debugging a user's self-hosted instance:

### 1. Fix: BizChat invocation shape drifted again (2026-08-21 capture)

Individual/consumer `copilot-m365-web` calls were failing with **only SignalR keepalive pings and no `type:1 target:"update"` frame at all** — client-side this surfaces as a generic "stream ended before producing a non-ping SSE event" with zero useful diagnostic (the socket connects and completes the handshake fine, it just never gets a real reply).

A fresh 2026-08-21 HAR capture of a working `m365.cloud.microsoft/chat` session shows the `#10718` shape has drifted again:

- `optionsSets`: 14 → 34 entries (adds `cwc_code_interpreter*`, `rich_responses`, `async_client_interaction`, `flux_v3_references*`, `flux_v3_image_gen_*`, etc.)
- `allowedMessageTypes`: 6 → 30 entries (adds `InternalSearchQuery`, `GeneratedCode`, `AuthError`, `TriggerPlugin`, etc.)
- `tone`: `"magic"` → `"Magic"` (now matches the enterprise tier's capitalization)
- `plugins`: `[]` → `[{Id:"BingWebSearch",Source:"BuiltIn"}]`
- `disconnectBehavior:"continue"` is now sent on **every** tier, not just enterprise
- new fields the `#10718` shape didn't have: `extraExtensionParameters:{}`, `isSbsSupported:true`, `renderReferencesBehindEOS:true`, `message.connectedFederatedConnections:["dummyId"]`
- `clientInfo` (both top-level and the copy nested in `message`) gained 8 keys: `clientEntrypoint`, `clientSessionId`, `ProductCategory`, `clientAppType`, `productEntryPoint`, `deviceOS`, `deviceType`, `clientPlatformVersion`

The `#10718` fields this capture didn't independently confirm (`conversationId`, `productThreadType`, `toolChoice`, `message.attachments`) are left as-is — removing them wasn't verified against a live socket, only adding the missing pieces was.

**Verified against a real individual-tier M365 Copilot account with a live WebSocket round-trip** (not just the unit tests) — the patched shape gets a real `type:1 target:"update"` reply instead of ping-then-close. Updated the `#10718` shape-pinning tests to match; all previously-passing enterprise-tier assertions are untouched.

### 2. Heads up: the Antigravity `gemini-3.1-pro-high` fix isn't published yet

Not a code change in this PR (already fixed on this branch) — just flagging it, because it cost real debugging time: `open-sse/config/antigravityModelAliases.ts` already has `"gemini-3.1-pro-high": "gemini-pro-agent"` on `release/v3.8.50`, but **the last npm publish (`3.8.49`) predates that alias**. Any self-hosted user still on `3.8.49` who has a combo/config referencing the old discovery id (a very findable id — it's what `:fetchAvailableModels` still advertises) gets a hard `400 INVALID_ARGUMENT` from Google on every call, with no fallback and no obvious link back to this already-fixed alias. Might be worth a patch release just for this, independent of whenever `3.8.50` otherwise ships.

### 3. Feature: "Import .har file" button for the M365 Copilot credential modal

`copilot-m365-web` is the only provider whose credential (`access_token` + `chathubPath`) can't be copy-pasted from a normal cookie/token field — it has to be extracted from the query string of a WebSocket URL inside a DevTools HAR export. Users (including the one who hit issue #1 above) were doing this by hand every time the short-lived (~75min) token needed refreshing.

Adds a client-side-only **"Import .har file"** button to both the Add and Edit connection modals for `copilot-m365-web`: picks a `.har` file, scans it for the BizChat ChatHub WebSocket request, and fills the `access_token`/`chathubPath` field automatically — plus a decoded-JWT expiry hint (valid / expiring soon / already expired) so users know immediately whether the HAR they exported is still usable. No new API endpoint; the user still clicks the existing **Check** button to validate against the server.

The extraction logic (`src/shared/utils/m365HarImport.ts`) is a pure function with no React/DOM dependency, keyed off a small per-provider registry in `HarImportButton.tsx` — adding HAR import for another web-session provider later is a one-line addition, not a new component.

## Test plan

- [x] `node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --test tests/unit/copilot-m365-*.test.ts` — 39/39 pass
- [x] `node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --test tests/unit/dashboard/m365-har-import.test.ts` — 7/7 pass
- [x] `npx eslint` clean on all changed/added files
- [x] `npx tsc --noEmit -p tsconfig.typecheck-core.json` clean on all changed/added files
- [x] Live WebSocket round-trip against a real individual-tier M365 Copilot account (bypassing the unit tests entirely) — confirmed a real `type:1 target:"update"` reply with correct content instead of ping-then-close
- [ ] Manual UI smoke test of the new Import .har file button (only ran the pure-function unit tests; haven't clicked through the modal in a running dev server)


eenggar-svg:
## Follow-up: fixed a 4th bug found while smoke-testing this branch

While clicking through the combo builder UI to manually verify the HAR-import feature above, hit a real crash unrelated to the other three items:

**`FORMATTING_ERROR`: combo dashboard crashes on missing i18n interpolation values**

`getI18nOrFallback()` (duplicated across 5 files under `src/app/(dashboard)/dashboard/combos/`) called `t(key)` with no `values`, for three message keys that require an ICU interpolation variable:
- `recommendationsUpdated` — needs `{strategy}`
- `modePackUpdated` / `savingModePack` — need `{pack}`

next-intl throws when a required variable is missing, so opening the Strategy Recommendations panel in the combo builder (or switching mode packs) crashed the whole page — instead of falling through to the manual `.replace()` fallback text that was already sitting right there in the same call site.

Fix: extended the helper to accept an optional `values` object and forward it to `t()`; the outer `.replace()` calls are left as-is (harmless no-ops once `t()` already interpolated, still needed for the raw-fallback path). Only 2 of the 5 duplicated copies of this helper are ever called with a placeholder-bearing key — verified the other 3 aren't affected.

Commit: ddd738e

Issue #2300, opened 2026-05-16
GET /api/combos is management-gated, blocking API-key combo metadata reads needed by OpenCode integrations


thomasmaerz:
## Summary

OpenCode + `opencode-omniroute-auth` performs read-only combo capability enrichment by fetching combo definitions from `GET /api/combos`.

In my deployment, this endpoint is management-auth protected, so a normal API key is rejected with `403 AUTH_001 Invalid management token`.
This creates an integration contract mismatch: the same key can access `/v1/models` but cannot access combo metadata needed to derive combo capabilities/context limits.

## Environment

- OmniRoute version: **3.8.0**
- Deployment: **remote/non-localhost**
- Auth mode: **password authentication enabled**

### Plugin behavior
`opencode-omniroute-auth` fetches:
- `baseUrl` with `/v1` stripped, then `+ /api/combos`
- with `Authorization: Bearer <apiKey>`

(See `src/omniroute-combos.ts`, `fetchComboData`.)

### OmniRoute behavior
- `GET /api/combos` calls `requireManagementAuth`
- bearer non-management tokens return `403 Invalid management token`

### Live validation
With the same OpenCode key:
- `GET /v1/models` -> `200`
- `GET /api/combos` -> `403` + `AUTH_001 Invalid management token`

## Repro

```bash
python3 - <<'PY'
import json, urllib.request, urllib.error
key=json.load(open('/Users/tmaerz/.local/share/opencode/auth.json'))['omniroute']['key']
for url in ['http://<LAN_IP>:20128/v1/models','http://<LAN_IP>:20128/api/combos']:
    req=urllib.request.Request(url,headers={'Authorization':f'Bearer {key}','Accept':'application/json'})
    try:
        with urllib.request.urlopen(req,timeout=20) as r:
            print(url, '->', r.status)
    except urllib.error.HTTPError as e:
        print(url, '->', e.code, e.read(300).decode('utf-8','ignore'))
PY
```

## Expected

A client API key that can list models should also be able to read combo metadata required for safe capability/context derivation (or an equivalent API-key-safe read endpoint should exist).

## Actual

Combo metadata read path is management-only; enrichment fails.

## Suggested remediation

1. Add a dedicated **read-only API-key-safe combo metadata endpoint** (minimum fields needed for enrichment), or
2. Allow `GET /api/combos` with explicit API-key read scope while keeping write routes management-only.

## Acceptance criteria

[...]

PR #11053, opened 2026-08-21
fix(ci): close the two PR-level blockers + remaining base-reds on the sec batch-2 branch (#11040)

jonlwheat2-gif:
## Summary

Follow-up to upstream PR #11040 (security batch 2). The upstream PR's head branch
(`d29829b8`) has two PR-caused CI failures plus the known base-reds. This branch
adds the fixes on top of the exact same 11 security commits, so the maintainers can
merge it into `fix/sec-adv-v3850-batch2` (or cherry-pick the two fix commits).

## What's fixed

| Commit | Fix | Gate |
|---|---|---|
| `811560e9e` | **Build App**: `readyz/route.ts` re-exported Next.js route config (`dynamic`) — declare it locally, matching `healthz` | Build App ✅ |
| `811560e9e` | **mutation-test-coverage**: add `tests/unit/authz/oauth-autoimport-local-only.test.ts` (new in this PR, covers mutated `routeGuard.ts`) to `stryker.conf.json tap.testFiles` | Fast Quality Gates ✅ |
| `ade275967` | **open-sse-typecheck**: guard `clearRecoveredProviderState` with `"connectionId" in credentials` (sentinels from #10971 aren't assignable) | Fast Quality Gates ✅ |
| `ade275967` | **ESLint**: drop 13 `no-explicit-any` in `account-rotation-lot-c` + `terminal-status-origin` tests | ESLint ✅ |
| `ade275967` | **Unit shards**: sync 5 test files to current base behavior (#10969 sentinel, #10966 403, #10973 inventory, GHSA-wg9p adoption opt-in) | Unit Tests ✅ |

## Verification

All fixed test files pass; eslint clean; `check:mutation-test-coverage` drift = 0;
open-sse-typecheck gate = 0 regressions (baseline frozen counts unchanged).

PR #11050, opened 2026-08-21
fix(webhooks): remove 3 declared-but-never-emitted events

maxmad64bis:
## Summary

Remove 3 declared-but-never-emitted webhook events (`provider.error`, `provider.recovered`, `combo.switched`) from the public `WebhookEvent` union — pure drift. Catalog is now `request.completed | request.failed | quota.exceeded | test.ping`.

- `WebhookEvent` union reduced `7 → 4`; `WEBHOOK_EVENT_VALUES` is the single exported source of truth (`src/lib/webhooks/eventDescriptions.ts:22`).
- `POST /api/webhooks` and `PUT /api/webhooks/[id]` now validate `{events}` with `z.enum(WEBHOOK_EVENT_VALUES_WITH_WILDCARD)` and **reject ghost values with 400** (was `z.string()`, silent pass-through).
- Removed 3 `EVENT_DESCRIPTIONS` blocks + 3 `EVENT_COLORS` in `discord.ts`; aligned `agentSkills.ts` `omni-webhooks` EN description.
- i18n B-pattern: `en.json` corrected + 42 locales set to `__MISSING__:<new en>` (runtime falls back to corrected English until the translation pipeline backfills — no human translation in PR).
- Added `tests/unit/webhooks-ghost-events.test.ts`.

## Related Issues

- Related to upstream `#10965` (STRICT_ZERO_COST, the PR base).

## Validation

- [x] Change type: other (webhooks catalog / i18n)
- [x] Focused tests: `node --import tsx/esm --test tests/unit/webhooks-ghost-events.test.ts` → 3 pass / 0 fail; `npm run lint`; `npm run typecheck:core`; `npm run check:cycles`
- [x] `npm run lint`
- [x] Reconciled with the current active release base (`upstream/release/v3.8.50` `3caa59107`), focused checks rerun
- [x] Production-code changes include a new or updated automated test in this PR
- `scripts/i18n/check-ui-value-drift.mjs` (base `3caa59107`) → PASS (0 rewritten EN value left a stale translation); `scripts/i18n/check-ui-keys-coverage.mjs --threshold=65` → PASS (all 42 locales ≥ 99.9%)

## Tests Added Or Updated

- `tests/unit/webhooks-ghost-events.test.ts` (new, 3 tests: catalog=4, zod 400 on ghost, notifyWebhookEvent)

## Coverage Notes

- No `src/`/`open-sse/`/`electron/`/`bin/` line coverage regression; change is a removal + validation, covered by the new unit test.

[...]

PR #11076, opened 2026-08-22
feat(api): flag a pinned account on /v1/combos steps without leaking the id

### LINKED ISSUE #10968: feat(api): GET /v1/combos should flag pinned accounts without leaking connectionId
## Bug Description

`GET /v1/combos` (API-key surface, `#2300`) projects only `kind` / `model` / `providerId` / `comboName`. `connectionId` is stripped **on purpose** and must stay stripped.

Consequence for operators who pin two **accounts** of the same provider (fill-first / priority failover):

```json
"models": [
  { "kind": "model", "model": "jina-ai/some-embed", "providerId": "jina-ai" },
  { "kind": "model", "model": "jina-ai/some-embed", "providerId": "jina-ai" }
]
```

Clients cannot tell “two pinned wallets” from “the same step duplicated”. Dashboard already shows `acct <8-char>` (`ComboControlCenterClient`). The public catalog has no equivalent boolean.

This is **not** a request to leak `connectionId`. A boolean (or count) is enough.

## Steps to Reproduce

1. Create a combo with two model steps, same `providerId` + `model`, different pinned `connectionId`s.
2. `GET /v1/combos` with a normal API key.
3. Compare to the dashboard combo editor.

## Expected Behavior

Public step includes something like `accountPinned: true` (or `pinnedAccountCount` on the combo) **without** the UUID.

Duplicate-looking steps with `accountPinned: true` are then distinguishable from a true duplicate of an unpinned step.

## Actual Behavior

Two pinned accounts collapse to two identical `{kind, model, providerId}` objects. Integrations (OpenClaw, OpenCode plugin) conclude the combo has no failover.

## Environment

- `src/app/api/v1/combos/projectCombo.ts` (`projectComboStep`)
- `tests/unit/v1-combos-projection.test.ts` currently asserts `connectionId` is dropped — keep that assertion
- Related: `#2300`, `#3979`

## Additional Context

No account ids or hostnames. Prefer a boolean over exposing even an 8-char prefix on this public route.

ntdatt812:
## Summary

`GET /v1/combos` strips `connectionId` on purpose, and must keep doing so. The side effect is that two steps pinning *different accounts* of the same provider project to byte-identical objects:

```json
"models": [
  { "kind": "model", "model": "jina-ai/some-embed", "providerId": "jina-ai" },
  { "kind": "model", "model": "jina-ai/some-embed", "providerId": "jina-ai" }
]
```

A client reading that cannot tell a two-account failover from the same step listed twice, so an importing integration concludes the combo has no failover.

This adds `accountPinned` to each model step — a boolean derived from the `connectionId` that still never leaves the server, not a prefix of it.

Three decisions worth calling out, since the issue left them open:

- **Per-step boolean rather than `pinnedAccountCount` on the combo.** The count says a combo has pins somewhere; the boolean says *which* of the duplicate-looking steps they are, which is the question a client is actually stuck on.
- **Always set on `model` steps, `true` or `false`.** An absent field then means an older server rather than an unpinned step, so a client can tell "not pinned" from "cannot tell".
- **Never set on `combo-ref`.** That step routes through another combo and has no account of its own, so `false` would be a claim rather than a fact.

The flag uses the same shape test `providerId` already gets. `cleanupComboConnectionRefs` deletes the key when a connection is deleted, so a step whose pinned account is gone reports `false` rather than pointing at nothing.

No route, schema, or caller changes — the three consumers (`/v1/combos`, and both VS Code combo routes) pass the projection straight through, and `/v1/combos` is not described in `docs/openapi.yaml`, so nothing there needed updating.

## Related Issues

- Closes #10968

## Validation

- [x] Change type: routing (combo projection / public API shape)
- [x] `npm run lint` — `eslint` clean on both changed files

```
tests/unit/v1-combos-projection.test.ts
  + combo-projection-capabilities-3979 / combo-routes-composite-tiers
  + models-catalog-route / combos-duplicate-route
  + combo-model-name-collision-8530 / combo-control-center      97 pass, 0 fail
```

**Mutation-checked**, after confirming each edit actually applied:

```
never set the flag        -> 6 failed
always set it to true     -> 3 failed
```

So the new cases pin the derivation itself rather than passing incidentally — in particular the second mutation is what keeps `accountPinned: false` from being a value nothing asserts.

## Tests Added Or Updated

- `tests/unit/v1-combos-projection.test.ts` — six new cases: a pinned step (asserting the id does not appear in the serialized output), an unpinned step, empty/non-string `connectionId` values, `combo-ref` carrying no flag, malformed steps still returning `null`, and the issue's own reproduction — two pinned accounts plus one unpinned step of the same provider and model, which now read as `[true, true, false]`.
- Two existing `#2300` assertions gained the new key. The connectionId-drop assertions themselves are unchanged and still green, which is the point of the pair.

## Coverage Notes

`src/app/api/v1/combos/projectCombo.ts` is covered by the file above. The added branch has both outcomes asserted directly and both directions mutation-checked.

## Reviewer Notes

- Additive key on a public response; no field is removed or renamed.
- The one behavior change a client could notice is that `accountPinned` now appears in `/v1/combos`, the VS Code combo route, and the raw VS Code combo route alike, since all three share `projectCombo`. That seemed right — the same ambiguity exists on all three — but say the word if you want it scoped to `/v1/combos` only.

### DIFF
--- a/src/app/api/v1/combos/projectCombo.ts
+++ b/src/app/api/v1/combos/projectCombo.ts
@@ -5,6 +5,10 @@
  * returning combo metadata to API-key callers. Kept in a separate module so
  * the projection can be unit-tested without spinning up the Next.js route.
  *
+ * #10968: the projection also reports `accountPinned` per model step — a boolean
+ * derived from the stripped `connectionId`, so callers can distinguish a combo
+ * that fails over between two accounts of one provider from a duplicated step.
+ *
  * #3979: client-facing combo catalogs (the `/v1/combos`, VS Code and LobeHub /
  * OpenCode import surfaces) can opt into advertising the combo's resolved
  * capabilities (multimodal / reasoning / caching) so importing clients enable
@@ -17,6 +21,19 @@ export interface PublicComboStep {
   model?: string;
   comboName?: string;
   providerId?: string;
+  /**
+   * #10968: whether this step pins one specific account of its provider.
+   *
+   * Two steps that pin different accounts of the same provider project to
+   * identical `{kind, model, providerId}` objects, so a client cannot tell a
+   * two-account failover from the same step listed twice. This says which it
+   * is without exposing the `connectionId` the flag is derived from — not even
+   * a prefix, per the issue.
+   *
+   * Set on every `model` step. Absent on `combo-ref`, which routes through
+   * another combo and has no account of its own.
+   */
+  accountPinned?: boolean;
 }
 
 /**
@@ -66,6 +83,10 @@ export function projectComboStep(step: Record<string, unknown>): PublicComboStep
     if (typeof step.providerId === "string" && step.providerId.length > 0) {
       out.providerId = step.providerId;
     }
+    // Same shape test as providerId above. `cleanupComboConnectionRefs` drops the
+    // key when the connection is deleted, so a step whose pinned account is gone
+    // reports false rather than pointing at nothing.
+    out.accountPinned = typeof step.connectionId === "string" && step.connectionId.length > 0;
     return out;
   }
   return null;
--- a/tests/unit/v1-combos-projection.test.ts
+++ b/tests/unit/v1-combos-projection.test.ts
@@ -2,6 +2,9 @@
  * Issue #2300 — Public projection of combo metadata for GET /v1/combos.
  * Verifies that internal routing details (connectionId, weights) are stripped
  * before being exposed to API-key callers.
+ *
+ * Issue #10968 — and that a step still reports *whether* it pins an account,
+ * which is derivable from the stripped connectionId without exposing it.
  */
 import test from "node:test";
 import assert from "node:assert/strict";
@@ -25,6 +28,7 @@ test("#2300 projectComboStep keeps model + providerId, drops connectionId/weight
     kind: "model",
     model: "anthropic/claude-sonnet-4",
     providerId: "anthropic",
+    accountPinned: true,
   });
 });
 
@@ -72,7 +76,7 @@ test("#2300 projectCombo preserves name/strategy/description, projects models", 
     name: "my-combo",
     strategy: "priority",
     description: "primary route",
-    models: [{ kind: "model", model: "openai/gpt-5", providerId: "openai" }],
+    models: [{ kind: "model", model: "openai/gpt-5", providerId: "openai", accountPinned: true }],
   });
 
   const serialized = JSON.stringify(out);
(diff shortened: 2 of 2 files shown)


ntdatt812:
Note on the one red check, since it is not from this PR and it is not only this PR.

**Merge integrity** fails on its second half. The changelog half passes:

```
[changelog-integrity] OK — no base bullets lost vs origin/release/v3.8.50.

Agent Skills Generator [dry-run]
  GENERATED:
    + omni-webhooks
##[error]Process completed with exit code 2.
```

`omni-webhooks` — this PR changes two files, `src/app/api/v1/combos/projectCombo.ts` and its test, and touches no route, no skill, and no generator input.

It traces to the base:

```
$ git show --stat 1c920eb8      # fix(webhooks): remove 3 ghost events (#11050)
  src/shared/constants/agentSkills.ts   | 2 +-
  src/lib/webhooks/eventDescriptions.ts | 39 +++-----
  …                                       skills/omni-webhooks/SKILL.md  — not in the diff

$ git log --oneline 1c920eb8..origin/release/v3.8.50 -- skills/omni-webhooks/SKILL.md
(nothing)
```

So #11050 changed a generator input and the generated file was never rebuilt, and nothing since has rebuilt it. Any PR rebased onto that commit inherits the red — which is the exact scenario the comment in `ci.yml` describes for gap 30.

I rebased onto `7ddbaf69` (current `release/v3.8.50`) after opening this and reran the focused checks; that fixed the changelog half and left this one, as expected. `npm run check:agent-skills-sync -- --apply` regenerates it in one commit, but it belongs in its own PR rather than smuggled into a combos change — say the word if you would rather I send that one too.

For completeness, the job is `continue-on-error` for fork PRs, so it is advisory here rather than blocking. It will not be advisory on the internal PR that next rebases onto this base.

ntdatt812:
Same story for the other red check, **No new ESLint warnings** — different cause, also on the base.

I ran the gate's own command locally, which writes the full inventory to `.artifacts/eslint-results.json`:

```
$ npm run lint:json -- --max-warnings 0        # exit 1

errors: 4  warnings: 0  files: 1
  4e/0w  tests/unit/capture-critical-db-state.test.ts
          @typescript-eslint/no-explicit-any (4x)
```

One file, four errors, and it is not one of mine — this PR's diff is two files:

```
$ git diff --name-only HEAD~1 HEAD
src/app/api/v1/combos/projectCombo.ts
tests/unit/v1-combos-projection.test.ts

$ git log --oneline -1 -- tests/unit/capture-critical-db-state.test.ts
0ff0490a test(db): assert resetDbInstance swaps the singleton, WAL mode, and schema_version seed (#10906)
```

Worth noting these are **errors**, not warnings, so they are not specific to the `--max-warnings 0` variant — `npm run lint` fails on the same four. Either #10906 landed them past the gate, or `config/quality/eslint-suppressions.json` needs a regenerate to cover them.

That accounts for both red checks on this PR, and both point at the base rather than the diff:

| check | cause | mine? |
|---|---|---|
| Merge integrity | `skills/omni-webhooks/SKILL.md` never rebuilt after #11050 changed `agentSkills.ts` | no |
| No new ESLint warnings | 4 × `no-explicit-any` in `capture-critical-db-state.test.ts` from #10906 | no |

Both also fail identically on #11072 and #11074, while #11053 — cut before these landed — is green on both. Happy to send either fix as its own PR if useful; I have kept both out of this one.

ntdatt812:
Final CI note, then I will stop adding to this thread.

The four unit shards went red too, so here is the direct evidence that this diff is not in any of it. This PR's own tests run in **shard 3**, and all fourteen pass there:

```
✔ #2300 projectComboStep keeps model + providerId, drops connectionId/weight/label
✔ #2300 projectComboStep keeps combo-ref's comboName, drops weight/label
✔ #2300 projectComboStep returns null for unknown kinds + malformed steps
✔ #2300 projectCombo preserves name/strategy/description, projects models
✔ #2300 projectCombo defaults strategy to 'priority' when missing
✔ #2300 projectCombo returns null for empty name
✔ #2300 projectCombo filters out malformed step entries silently
✔ #2300 projectCombo omits description when empty
✔ #10968 projectComboStep reports a pinned account without the connectionId
✔ #10968 projectComboStep reports false for an unpinned model step
✔ #10968 an empty or non-string connectionId is not a pin
✔ #10968 a combo-ref carries no accountPinned
✔ #10968 malformed steps stay null rather than gaining the flag
✔ #10968 two pinned accounts of one provider are no longer identical
```

What shard 3 fails on instead: `buildSlackPayload — all WEBHOOK_EVENTS produce valid payloads`, `buildSlackPayload — provider.error includes provider context`, the quota classifier, `probeBeforeSpawn`, service adoption, and the systemd-notify handshake. Shard 1 fails on `buildDiscordPayload` ×2, `buildTelegramPayload`, and three `getProviderCredentials` cases.

The webhook ones are the same root as the skills drift I flagged above — **#11050 removed three events from `WEBHOOK_EVENTS` and the payload tests that iterate it were never updated**, alongside the generated `omni-webhooks/SKILL.md` that was never rebuilt. One change, three red gates.

Consistent across the board:

| check | #11076 | #11072 | #11074 | #11053 |
|---|---|---|---|---|
| Unit Tests fast-path (1/4) | fail | fail | fail | **pass** |
| Merge integrity | fail | fail | fail | **pass** |
| No new ESLint warnings | fail | fail | fail | **pass** |

#11053 was cut before those commits landed and is green on all three. Nothing here is specific to this PR, and nothing here is specific to fork PRs either.

Green on this PR: Vitest, Docs Gates, Build, dast-smoke, semgrep ×2, Change Classification.

PR #7926, opened 2026-07-21
fix(translators): normalize TitleCase tool names for non-Anthropic models

nramabad:
Non-Anthropic models (OpenAI, Gemini, etc.) may respond with TitleCase tool names (Bash, Read, Write) but Claude Code defines its built-in tools as all-lowercase (bash, read, write). This adds `normalizeToolName()` to both the OpenAI→Claude and Gemini→Claude response translators, mapping known TitleCase names back to lowercase using the existing `REVERSE_MAP` from `claudeCodeToolRemapper.ts`. Unknown names are left unchanged to avoid breaking MCP or custom tools that intentionally use mixed case.

PR #11085, opened 2026-08-22
fix(claude): restore canonical tool names so Claude Code accepts tool calls

linhdmn:
⚠️ base-red inherited: #9985 — Quality Gates / fast-path unit jobs fail identically on unrelated PRs against release/v3.8.50; failures are not from this branch (see live-validation comment below).

## Problem

Claude Code clients routed through OmniRoute to OpenAI-compatible providers lose native tool calls entirely. Live reproduction (2026-08-22, `opencode-go/ox-alpha-free` upstream, `/v1/messages` with 143 declared tools):

1. The model emits a tool call for the `Bash` tool, but the name reaches the client lowercased as `bash`.
2. Claude Code validates tool names case-sensitively and rejects it:

```
<tool_use_error>Error: No such tool available: bash</tool_use_error>
```

Every retry fails the same way — Bash/Read/Write/CronCreate stay dead for the whole session while MCP tools (which happen to be lowercase already) keep working.

Call logs show two distinct leaks:

- **Non-streaming conversion never restored names.** When an upstream answers a `stream:true` request with an `application/json` body, OmniRoute converts via `translateNonStreamingResponse`, whose `convertOpenAINonStreamingToClaude` copied `function.name` verbatim into the Claude `tool_use` block. The streaming path already called `restoreClaudeToolName`; the non-streaming path did not.
- **Identity echoes pinned wrong casing.** Some upstream gateways echo the very lowercase name they emitted back through the `_toolNameMap` alias channel (`croncreate → croncreate`). `restoreClaudeToolName` returned that identity match before considering canonical casing, and `TOOL_RENAME_MAP` was missing the cron-era tools (`TaskCreate`, `TaskUpdate`, `TaskList`, `TaskGet`), so even the canonical upgrade had nothing to upgrade to.

## Fix

- `open-sse/handlers/responseTranslator.ts`: pass the request's `toolNameMap` into `convertOpenAINonStreamingToClaude` and restore each emitted tool name via `restoreClaudeToolName`, matching the streaming translator.
- `open-sse/services/claudeCodeToolRemapper.ts`:
  - Known Claude Code tools are upgraded to canonical PascalCase even when a map entry matches identically (`bash → Bash`, `croncreate → CronCreate`, …).
  - Added missing cron-era entries to `TOOL_RENAME_MAP`: `TaskCreate`, `TaskUpdate`, `TaskList`, `TaskGet`.
  - Preference order documented; real request-side aliases still win over canonical casing (`read → mcp__fs__read` is preserved), and the `#7926` no-map TitleCase→lowercase fallback is untouched.

## Tests

New suite `tests/unit/claude-code-tool-casing-identity-echo.test.ts` (12 cases):

- identity-echo / cloak-direction / mixed maps all upgrade echoed lowercase names (incl. live-repro `bash → Bash` and cron-era tools) in both `restoreClaudeToolName` and the streaming `openaiToClaudeResponse`
- `translateNonStreamingResponse` upgrades `bash → Bash` with no map (live repro) and honours aliases/canonical order
- regression guards: non-identity alias beats canonical, unknown custom names preserved verbatim, `#7926` fallback unchanged

All green together with the existing suites: `gemini-to-claude-tool-name-case-9008`, `claude-tool-name-casing-fix`, `claude-code-tool-remapper-flag-leak`, `provider-request-capture-toolname-map-4091`, `claude-server-tool-name-preserve`, `translator-resp-openai-to-claude`, `translator-openai-to-claude`, `claude-code-parity`, `xai-translators`, `9568-gemini-tool-casing-mismatch` (172 tests).


linhdmn:
**Live-environment validation (Hard Rule #18) — extended fix in 52af7f54c**

The original commits fixed lowercase echoes through alias maps, but a live test against a real upstream exposed a remaining gap on the most common route: **Claude Code → OpenAI-style upstream** ships no `_toolNameMap`, so an upstream echoing the canonical TitleCase fell through to the #7926 `REVERSE_MAP` fallback and was downcased (`CronCreate` → `croncreate`) — exactly the 'No such tool available' failure this PR targets.

**Live repro (before 52af7f54c)** — real glm-5.2-high via opencode-go, isolated DATA_DIR, port 20129:
```
POST /v1/messages {model: opencode-go/glm-5.2-high, tools:[Bash, CronCreate], tool_choice:{type:tool,name:CronCreate}}
→ HTTP 200, tool_use.name = croncreate   ❌ (Claude Code would reject)
```

**After 52af7f54c** (canonical echo terminal after the alias-map block):
```
Same request → HTTP 200, tool_use.name = CronCreate   ✅
Natural-choice variant (no tool_choice) → tool_use.name = CronCreate   ✅
```

**TDD**: failing tests added first (`canonical echo stays canonical on no-map routes`, declared-lowercase alias preservation), then the minimal fix. All suites green locally: identity-echo 14/14, tool-name sweep 97/97 + 89/89, translator suites 30/30 + 63/63, claude-code regression 73/73; lint clean; typecheck:core clean.

Legacy OpenCode/XML-style lowercase clients remain protected via explicit alias maps ({Read→read}), which restore honors before the canonical-echo terminal.

PR #11088, opened 2026-08-22
fix(ollama): route models by advertised capability

### LINKED ISSUE #11087: feat(providers): Ollama local capabilities are flattened to chat and specialty routes reject valid models
### Environment

- OmniRoute: 3.8.49
- Provider: `ollama-local` with multiple connections (`<ollama-host-a>`, `<ollama-host-b>`)
- Ollama: 0.32.x on Apple Silicon and CUDA/arm64
- Each connection exposes a mostly distinct model inventory

All hostnames, addresses, and connection IDs in this report are placeholders.

### Description

OmniRoute imports Ollama's OpenAI-compatible `/v1/models` list but does not enrich each model from Ollama's native `/api/show` response. Synced models therefore fall back to `apiFormat: "chat-completions"` and `supportedEndpoints: ["chat"]`, even when Ollama advertises another capability.

Examples:

| Ollama `/api/show` capabilities | OmniRoute result |
| --- | --- |
| `["image"]` | chat model |
| `["embedding"]` | chat model |
| `["completion", "vision", "tools", "thinking"]` | chat only; detailed capability metadata is incomplete |

This also makes valid local models unusable through their matching endpoints:

- `/v1/images/generations` rejects `ollama-local/<image-model>` as an invalid image model.
- `/v1/embeddings` rejects `ollama-local/<embedding-model>` as an unknown embedding provider.

With multiple Ollama connections, endpoint routing must select only a connection whose synchronized inventory advertises the requested model and endpoint.

### Expected behavior

1. During Ollama model synchronization, query `/api/show` for each discovered model.
2. Persist accurate `supportedEndpoints`, `apiFormat`, vision, tools, and thinking metadata per connection.
3. Surface the accurate type and endpoints through `/v1/models`.
4. Allow Ollama image and embedding models through the OpenAI-compatible image/embedding routes.
5. Restrict selection to connections that advertise the requested model and capability.

### Validation plan

- Unit/integration test with image-only, embedding-only, and completion/vision/tools/thinking models.
- Two-connection test proving image and embedding requests select the advertising connection rather than the higher-priority sibling.
- Verify the image request reaches `<ollama-base-url>/v1/images/generations`.
- Verify the embedding request reaches `<ollama-base-url>/v1/embeddings`.

yourspraveen:
Fixes #11087

## Summary
- enrich Ollama model discovery with bounded `/api/show` capability probes
- persist image, embedding, chat, vision, tools, and thinking metadata per connection
- route local image and embedding requests only through connections advertising the requested model and endpoint
- preserve existing SSRF guard and proxy behavior for discovery requests

## Validation
- `node --import tsx/esm --test tests/unit/ollama-local-capabilities-routing.test.ts`
- focused related route tests (15 passed)
- `npm run typecheck:core`
- `npm run lint` (0 errors; 6 pre-existing warnings)
- `npm run test:vitest` (34 files, 291 tests passed)
- `npm run check:cycles`
- `npm run check:any-budget:t11`

The regression test uses placeholder loopback hosts and generated connection IDs; no internal hostnames or connection UUIDs are published.

## CI status

`⚠️ base-red inherited` — the base tip (`main` @ c68cda7) is already red on the
Lint job. Run [32542442026](https://github.com/diegosouzapw/OmniRoute/actions/runs/32542442026)
on the unmodified base fails `check:file-size` with:

- `open-sse/services/accountFallback.ts: 1968 > congelado 1966`
- `tests/unit/account-fallback-service.test.ts: 1570 > congelado 1563`

Neither file is touched by this PR. Per the release-green policy these are
mid-cycle ratchet drift, rebaselined at release — not a contributor concern.

The one file-size violation this PR *did* introduce
(`src/app/api/providers/[id]/models/route.ts: 2269 > congelado 2250`) is fixed
in 6d66499 by extracting the `/api/show` probe wiring into
`discovery/helpers.ts::enrichOllamaLocalModels()`; the route file is back to its
frozen 2249 lines with no behavior change.

The **Build** job failure is environmental, not a code defect: fork PRs are
deliberately excluded from the self-hosted `omni-release` runner by the
`runs-on` guard in `ci.yml`, so this run fell back to `ubuntu-latest`, where the
Turbopack production build was killed after 20 min (`The operation was
canceled.`). The same commit's base built fine in 40 min on the 32-core
self-hosted runner. See the `OMNIROUTE_USE_TURBOPACK=0` note in
`scripts/build/build-next-isolated.mjs` re: memory-constrained machines.

The **Unit Tests (2/8)** failure is a CI flake, not a defect. The shard reports
3530 passed / 1 failed, and the single failure is in
`tests/unit/build/should-promote-latest-5301.test.ts` — a release-script test
untouched by this PR:

```
✖ a pre-release VERSION never promotes latest (5.640152ms)
  Error: spawnSync bash EPIPE
    spawnargs: [ 'scripts/ci/should-promote-latest.sh', '3.8.40-rc.1' ],
    status: 0,
    stdout: 'false\n',
```

`status: 0` with `stdout: 'false\n'` shows the script ran to completion and
returned the correct answer (a pre-release must not promote `latest`);
`execFileSync` raised `EPIPE` on the closed pipe regardless. Nothing in this
PR touches `scripts/ci/` or `tests/unit/build/`.

## Ollama version requirements

`/api/show` has existed since early Ollama, but the `capabilities` array it
returns is version-gated. Discovery degrades safely on older servers — the Zod
schema marks `capabilities` optional, a non-2xx/throwing probe yields `null`,
and `applyOllamaShowCapabilities` returns the model unenriched — so old servers
get exactly the pre-PR behavior, never an error.

| Capability | First Ollama release |
| --- | --- |
| `completion`, `embedding`, `vision`, `tools`, `insert` | v0.6.4 (2025-04-02) |
| `thinking` | v0.9.0 (2025-05-29) |
| `image` | v0.14.0 (2026-01-10) |

So the embeddings half of this fix needs Ollama ≥ 0.6.4 and the image half needs
≥ 0.14.0; below those the routing simply falls through to the previous path.


diegosouzapw:
Deep-validation result while merging into the current `release/v3.8.50` tip — holding for a design call rather than merging red:

**What passes:** the two routing tests (image/embedding requests route through the advertising connection) and the full resolution compiles clean (typecheck 0, file-size gate green after your helpers.ts extraction).

**What fails on the tip (not on your base):** `Ollama discovery maps /api/show capabilities…` — the *persisted* store lookup. Root-caused with a direct probe:

```
isChatSelectableModel("ollama-local", { supportedEndpoints: ["images"] }) → false
filterSelectableModels keeps both models; filterChatSelectableModels drops the images-only one
```

`persistDiscoveredModels()` (src/lib/providerModels/modelDiscovery.ts:382) runs `filterChatSelectableModels` before `replaceSyncedAvailableModelsForConnection`, so a discovered images-only / embeddings-only Ollama model never reaches the synced store — which is exactly the store `resolveLocalSyncedEndpointRoute` reads. On the tip that means discovery can never feed non-chat routing end-to-end (your routing tests pass because they seed the store directly).

The chat filter at persist time is load-bearing for chat-picker hygiene (#8627), so the fix is a design choice, not a mechanical one:
1. Persist everything (drop the chat filter at write) and apply `filterChatSelectableModels` at chat-picker read sites; or
2. Keep persist as-is and give non-chat endpoints a separate persisted channel.

Happy to land whichever direction you pick — the enrichment/routing commits themselves merged cleanly.

PR #11090, opened 2026-08-22
fix(claude): restore canonical tool names so Claude Code accepts tool calls

linhdmn:
## Problem

Claude Code (v2.1.x) rejects tool calls whose `tool_use.name` doesn't exactly match a declared tool. Routed through OmniRoute to OpenAI-compatible upstreams (e.g. `opencode-go/ox-alpha-free`), the model emits **lowercase** tool names (`bash`, `croncreate`, `taskcreate`…). Every such call fails with:

```
No such tool available: bash
```

This kills Bash/Read/Write/CronCreate for the whole session — `/loop` scheduling, shell access, and file tools all break simultaneously.

## Root causes (two independent gaps)

1. **Identity-echo shadowing in `restoreClaudeToolName`** — an alias-map entry where key === value (upstreams echo back the very lowercase name they emitted) returned early and blocked the canonical casing table from upgrading `croncreate → CronCreate`.
2. **Non-streaming translator emitted names verbatim** — when an upstream answered a `stream:true` `/v1/messages` request with an `application/json` body, `translateNonStreamingResponse` built `tool_use` blocks with the raw `function.name` and no restoration at all.

## Changes

- `restoreClaudeToolName`: identity entries can no longer pin a known Claude Code tool below its canonical casing; non-identity request-side aliases still win first; #7926 lowercase-client fallback unchanged.
- `translateNonStreamingResponse`: thread the request `toolNameMap` into `convertOpenAINonStreamingToClaude` and run every emitted name through `restoreClaudeToolName`.
- `TOOL_RENAME_MAP`: add `taskcreate/taskupdate/tasklist/taskget` (native in CC 2.1.238, previously unprotected against lowercasing).
- Tests: `tests/unit/claude-code-tool-casing-identity-echo.test.ts` — covers the live-reproduced streaming + non-streaming paths, identity-echo maps, alias precedence, and #7926 regression guards.

## Verification

- 179 unit tests across all translator/remapper suites pass; eslint clean; `tsc --noEmit` (core) clean.
- Live end-to-end against `opencode-go/ox-alpha-free`: `content_block_start` now arrives as `"name":"Bash"` / `"name":"CronCreate"` (previously lowercase and rejected by the CLI).

Also includes a cherry-picked build fix (`@/lib/db/connections`) required for the dashboard build on this branch.


diegosouzapw:
Closing as a duplicate of #11085 — both PRs point at the same branch/head (52af7f54cd), but this one targets `main` while #11085 correctly targets `release/v3.8.50` (the active development branch). Work continues on #11085. Thank you @linhdmn for the fix — excellent live repro and root-cause writeup there!

Issue #6205, opened 2026-07-04
fix(startup): 9Router embedded service unhealthy: EADDRINUSE + missing better-sqlite3, Provider Exposure toggle 503s silently, Web UI embed hits wrong port


jonlwheat2-gif:
### Summary
On the Embedded Services → 9Router tab (`/dashboard/providers/services?tab=9router`), the bundled 9Router process (v0.5.18, port 20130) appears "Running" in the UI but is actually crash-looping / degraded, and several UI controls on this page silently fail without surfacing any error to the user.

### Environment
- OmniRoute v3.8.44 (Windows, global npm install)
- 9Router v0.5.18, PID reported in UI, configured port 20130
- Dashboard reachable at `http://localhost:20128`

### Findings (via DevTools console/network on the Embedded Services page)

**1. 9Router startup/runtime errors (from `/api/services/9router/logs?tail=200`):**
```
Error: listen EADDRINUSE: address already in use 127.0.0.1:20130
[DB] better-sqlite3 unavailable: Cannot find module 'better-sqlite3'
  at C:\Users\<user>\.omniroute\services\9router\node_modules\9router\app\.next-cli-build\server\middleware.js
  - C:\Users\<user>\.omniroute\services\9router\node_modules\9router\app\node_modules\next\dist\server\next-server.js
[DB] Driver: node:sqlite | file: C:\Users\<user>\.omniroute\services\9router\data\db\data.sqlite
KIRO_MODELS ListAvailableModels failed: Kiro ListAvailableModels 403: {"message":"The bearer token included in the request is invalid.","reason":null}
```
The `KIRO_MODELS` 403 repeats on a recurring interval (looks tied to the `[AutoPing] scheduler`), i.e. this isn't a one-off — it's a persistent invalid/expired bearer token for the Kiro provider that keeps failing silently in the background.

`better-sqlite3` is missing from 9Router's own bundled `node_modules` (`.omniroute\services\9router\node_modules\9router\app\node_modules\...`), which is a separate install tree from the main global `omniroute` package — updating/reinstalling the global package does not fix this, since 9Router manages its own dependency tree under `~/.omniroute/services/9router/`.

**2. "Provider Exposure" toggle (`Expose as 9router/...`) does nothing but reports success visually:**
- Toggling it fires `POST /api/services/9router/provider-expose`
- Every call returns **503 Service Unavailable** (reproduced twice — toggling off, then back on)
[...]

PR #11093, opened 2026-08-22
fix(sse): classify RFC1918 and .local hosts as local providers

### LINKED ISSUE #11091: fix(resilience): isLocalProvider() misses RFC1918 / .local hosts — LAN-hosted backends lose per-model 404 lockout
## Description

- `localhost`, `127.0.0.1`
- `172.16.0.0/12`

```ts
  string | undefined;

```

missed by `isLocalProvider`.

## Steps to Reproduce

## Expected Behavior

## Actual Behavior

## Current Workaround

## Suggested Fix

health-check targeting.

## Test Impact

active.

## Related

## Environment

- OmniRoute: 3.8.49+

yourspraveen:
Fixes #11091

## Problem

`isLocalProvider()` (`open-sse/config/providerRegistry.ts`) matched only
`localhost`, `127.0.0.1` and `172.16.0.0/12` — missing most of the address space a
self-hosted LLM box actually lives on.

Its **sole caller** is the 404 model-lockout branch at `src/sse/services/auth.ts:2955`:

```ts
```

So a LAN-hosted Ollama/LM Studio/vLLM connection failed the check and a 404 for one
missing model cooled the **entire connection** rather than locking out just that model.
On a multi-host setup where each host deliberately holds a different model subset, the
missing-model 404 is the *expected* case — one lookup took a whole host offline.

## Fix

Delegate to `isPrivateHost()` (`src/shared/network/outboundUrlGuard.ts`) — this
codebase's existing definition of a private host — instead of maintaining a second,
narrower one. `LOCAL_HOSTNAMES` stays as an additive escape hatch.

Now correctly classified as local: `192.168/16`, `10/8`, `100.64/10` (CGNAT/Tailscale),
`169.254/16`, IPv6 ULA/link-local/`::1`, and `.local` / `.internal` / `.localhost` names.

Two deliberate asymmetries, both covered by tests:

- `isPrivateHost()` fails **closed** (empty host ⇒ private) because it guards egress.
  `isLocalProvider()` must fail **open** — it only widens a lockout scope and gates no
  outbound request — so an empty or unparseable hostname still returns `false`.
- Custom LAN suffixes that are neither IP literals nor `.local`/`.internal`
  (e.g. `studio.lan`) still need `LOCAL_HOSTNAMES`; they are indistinguishable from
  public hostnames without resolving them.

## Blast radius

There are two different `isLocalProvider` functions in the tree. The **`providerId`**
variant (`src/shared/constants/providers.ts`) is untouched — it has three callers
(`providerHints.ts`, `validation.ts`, `validation/webCookie.ts`) and existing coverage in
`ollama-local-provider.test.ts` / `provider-validation-specialty.test.ts`.

This PR changes only the **`baseUrl`** variant, whose full caller set is the single
`auth.ts:2955` line above. The function's doc comment also claimed use for "health check
targets"; that is stale — no such caller exists.

## Validation

TDD per Hard Rule #18. The 7 new cases in
`tests/unit/local-provider-lan-detection-11091.test.ts` were checked against the old
predicate first — it misclassified **9/9** local hosts as remote (`192.168.1.50`,
`10.10.0.181`, `10.0.0.1`, `100.64.0.1`, `169.254.1.1`, `studio.local`, `box.internal`,
`[::1]`, `[fd00::1]`), and all pass after the change. Boundary cases are asserted in both
directions (`172.15`/`172.32`, `100.63`, `192.169` stay non-local).

- `node --import tsx/esm --test tests/unit/local-provider-lan-detection-11091.test.ts` — 7 passed
- `node --import tsx/esm --test tests/unit/ollama-local-provider.test.ts tests/unit/provider-validation-specialty.test.ts` — 126 passed
- `node --import tsx/esm --test tests/unit/account-fallback-lockout-eviction.test.ts tests/unit/domain-lockout-policy.test.ts tests/unit/model-lockout-decay.test.ts tests/unit/auth-ollama-cloud-per-model-403-3027.test.ts` — 16 passed
- `npm run check:cycles` — clean
- `npx eslint` on both changed files — clean

---

## ⚠️ base-red inherited: #9985

Every failing check on this PR reproduces on the unmodified base tip
(`release/v3.8.50`) and none of them touch the two files in this diff
(`open-sse/config/providerRegistry.ts`, `tests/unit/local-provider-lan-detection-11091.test.ts`).

The failing set here is byte-identical to unrelated PRs opened against the same base
(e.g. #11135, which shares no files with this one):

```
Fast Quality Gates
Merge integrity (changelog + generated skills)
No new ESLint warnings
Unit Tests fast-path (1/4 … 4/4)
```

**Unit Tests fast-path** — failures are in unrelated suites: Vietnamese locale key parity /
ICU placeholders, `getTokenLimit` hyperagent model ids, provider-detail browser-bundle
safety, `probeBeforeSpawn` (#6205), systemd-notify handshake, and `buildSlackPayload`.
No assertion in the failing set calls `isLocalProvider()`.

**Merge integrity** — `check:changelog-integrity` passes (`OK — no base bullets lost`).
The job exits 2 on `check:agent-skills-sync`, which regenerates one stale generated file
on the base:

```
  GENERATED:
    + omni-webhooks
```

`skills/omni-webhooks/SKILL.md` is out of sync in the base tree, not in this branch.

Local verification of the changed code is in the Validation section above — the new test
file passes 7/7 and `npx eslint` is clean on both changed files.

### DIFF
--- a/open-sse/config/providerRegistry.ts
+++ b/open-sse/config/providerRegistry.ts
@@ -10,6 +10,7 @@ export {
 } from "./providers/registry/alibaba/index.ts";
 export { REGISTRY } from "./providers/index.ts";
 import { REGISTRY } from "./providers/index.ts";
+import { isPrivateHost } from "@/shared/network/outboundUrlGuard";
 import {
   RegistryModel,
   REASONING_UNSUPPORTED,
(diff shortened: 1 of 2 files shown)


yourspraveen:
Closing as superseded — #11122 landed the same fix for #11091 (thanks @rqzbeh), so this PR's `isLocalProvider()` change is already on `release/v3.8.50` as 4220c810e and the branch now conflicts.

One thing did survive the overlap: both this PR and #11122 imported `isPrivateHost` from `src/shared/network/outboundUrlGuard.ts`, whose first line is `import { isIP } from "node:net"`. Since `providerRegistry.ts` is reachable from `ProviderDetailPageClient.tsx`, that breaks the dashboard client bundle — `media-page-client-browser-bundle.test.ts` has been red on the release tip since #11122 merged. Fix opened as #11154, which leaves the #11122 routing behaviour untouched and only moves the import.

PR #11119, opened 2026-08-22
fix(resilience): detect RFC1918 and private hosts in isLocalProvider (#11091)

### LINKED ISSUE #11091: fix(resilience): isLocalProvider() misses RFC1918 / .local hosts — LAN-hosted backends lose per-model 404 lockout

rqzbeh:
Fixes #11091.

### Summary
1. Updates `isLocalProvider(baseUrl)` in `open-sse/config/providerRegistry.ts` to check `isPrivateHost(hostname)` alongside `LOCAL_HOSTNAMES`.
2. This ensures LAN-hosted inference backends (Ollama, LM Studio, vLLM) on private IPv4/IPv6 ranges (`192.168.0.0/16`, `10.0.0.0/8`), CGNAT/Tailscale (`100.64.0.0/10`), mDNS (`.local`), and private domains (`.internal`) are recognized as local providers.
3. As a result, missing-model 404 responses from these local hosts trigger per-model lockout instead of incorrectly disabling the entire connection.
4. Adds automated unit test `tests/unit/is-local-provider-private-hosts-11091.test.ts`.


yourspraveen:
Flagging two things rather than a review:

1. This PR's diff is currently empty — `gh pr diff 11119` returns no file changes, so the
   `isLocalProvider` change described in the body does not appear to be on the branch. Likely
   a push that did not land, or the branch is even with its base.
2. It overlaps #11122 (same fix, same issue, base `release/v3.8.50`) and #11093, which has been
   open since 05:31 UTC today. I left the technical detail on #11122 to keep it in one place.

rqzbeh:
Closing in favor of #11122 (which targets `release/v3.8.50`).

Thank you @yourspraveen for the review! Updated #11122 with `if (!hostname) return false;` ahead of `isPrivateHost` so unparseable/empty hostnames fail open as non-local.

PR #11122, opened 2026-08-22
fix(resilience): include RFC1918, CGNAT, and mDNS hosts in isLocalProvider (#11091)

### LINKED ISSUE #11091: fix(resilience): isLocalProvider() misses RFC1918 / .local hosts — LAN-hosted backends lose per-model 404 lockout

rqzbeh:
Fixes #11091.

### Root Cause
`isLocalProvider(baseUrl)` in `open-sse/config/providerRegistry.ts` matched only `localhost`, `127.0.0.1`, `172.16.0.0/12`, and `LOCAL_HOSTNAMES`. It missed `192.168.0.0/16`, `10.0.0.0/8`, `100.64.0.0/10` (Tailscale/CGNAT), `169.254.0.0/16`, and `.local` / `.internal` hostnames. As a result, 404 responses from self-hosted LAN/Tailscale LLMs misclassified the host as non-local and triggered a whole-connection cooldown instead of a model-only lockout.

### Fix
Updated `isLocalProvider(baseUrl)` to delegate hostname evaluation to `isPrivateHost(hostname)` from `@/shared/network/outboundUrlGuard`, which already covers all RFC1918, CGNAT/Tailscale, link-local, and mDNS private domain suffixes.

### Test Coverage
Added unit test `tests/unit/is-local-provider-11091.test.ts` asserting that LAN (`192.168.1.50`, `10.0.0.5`), Tailscale (`100.64.1.2`), link-local (`169.254.1.1`), mDNS (`studio.local`, `mybox.internal`), and Docker (`172.18.0.2`) hosts resolve to `true`, while public hosts (`api.openai.com`, `8.8.8.8`) resolve to `false`.

### DIFF
--- a/open-sse/config/providerRegistry.ts
+++ b/open-sse/config/providerRegistry.ts
@@ -133,6 +133,7 @@ export function isLocalProvider(baseUrl?: string | null): boolean {
   try {
     const url = new URL(baseUrl);
     const hostname = url.hostname;
+    if (!hostname) return false;
     return LOCAL_HOSTNAMES.has(hostname) || isPrivateHost(hostname);
   } catch {
     return false;
--- a/open-sse/config/providerRegistry.ts
+++ b/open-sse/config/providerRegistry.ts
@@ -10,6 +10,7 @@ export {
 } from "./providers/registry/alibaba/index.ts";
 export { REGISTRY } from "./providers/index.ts";
 import { REGISTRY } from "./providers/index.ts";
+import { isPrivateHost } from "@/shared/network/outboundUrlGuard";
 import {
   RegistryModel,
   REASONING_UNSUPPORTED,
@@ -132,11 +133,7 @@ export function isLocalProvider(baseUrl?: string | null): boolean {
   try {
     const url = new URL(baseUrl);
     const hostname = url.hostname;
-    // Strictly matching 172.16.0.0/12 (Docker/local) and explicitly blocking ::1 per SSRF hardening
-    return (
-      LOCAL_HOSTNAMES.has(hostname) ||
-      /^172\.(1[6-9]|2[0-9]|3[0-1])\.\d{1,3}\.\d{1,3}$/.test(hostname)
-    );
+    return LOCAL_HOSTNAMES.has(hostname) || isPrivateHost(hostname);
   } catch {
     return false;
   }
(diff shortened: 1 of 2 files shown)


yourspraveen:
Heads-up on overlap: #11093 has been open against `release/v3.8.50` since 05:31 UTC today
fixing this same issue, and is linked from #11091. Flagging it so the maintainers can pick
one rather than review both — no criticism intended, and I don't think this is duplicated
work in the avoidable sense: #11091 names `isPrivateHost()` and its file/line explicitly and
its "Suggested Fix" section prescribes the delegation, so any two implementations were always
going to converge on the same one-liner.

One behavioral difference worth carrying into whichever PR lands:

`isPrivateHost()` fails **closed** — an empty hostname returns `true` — because it guards
outbound egress. `isLocalProvider()` needs the opposite: its sole caller is the 404
model-lockout branch at `src/sse/services/auth.ts:2955`, which gates no request, so an
unparseable or hostname-less base URL should stay non-local. #11093 adds an explicit
`if (!hostname) return false;` ahead of the delegation for that reason. The version here
inherits the fail-closed behavior, so a `file:`-scheme or otherwise hostname-less `baseUrl`
would classify as local. Narrow, but it is a real divergence and it is deliberate in #11093.

On diff scope: this PR currently carries ~2,180 lines across 55 files — `src/i18n/messages/*`,
`package.json`, `stryker.conf.json`, and the quality baselines — against 2 files in #11093.
I think that is a rebase artifact rather than intent; some of it (`skills/omni-webhooks/SKILL.md`)
is the stale generated file that is currently failing `check:agent-skills-sync` on the base
tip via #9985, so it may be picking up base repairs.

Happy either way — @diegosouzapw, whichever you prefer to take.

rqzbeh:
Thank you @yourspraveen for the review!

Updated this PR with `if (!hostname) return false;` ahead of `isPrivateHost` so unparseable or hostname-less URLs fail open (stay non-local) as recommended, along with additional unit assertions.

Issue #7682, opened 2026-07-18
fix(providers): OpenCode Free models do not appear in Combo Builder on Windows


Sagaruk06:
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 24H2

### Node.js Version

node --version

### Provider(s) Involved

OpenCode Free

### Model(s) Involved

All OpenCode Free models

### Client Tool

All OpenCode Free models

## Problem

OpenCode CLI is installed (`opencode --version` = 1.18.3).

The "OpenCode Free" provider appears correctly in Providers and on the Home topology.

Import Models reports:

> All models already imported

However, OpenCode Free never appears inside Combo Builder. Only providers such as OpenAI Codex are listed.

## Additional issue

Running:

```bash
omniroute setup-opencode
```

returns:

```text
Cannot find package '@/shared' imported from
src/shared/network/outboundUrlGuard.ts
```

## Expected behavior

OpenCode Free models should appear in Combo Builder and `setup-opencode` should complete successfully.

### Steps to Reproduce

1. Install OmniRoute v3.8.48 globally using npm.
2. Install OpenCode CLI (`opencode --version` = 1.18.3).
3. Start OmniRoute.
4. Add the "OpenCode Free" provider.
5. Click "Import Models" (it reports "All models already imported").
6. Go to Combos → Create Combo → Add Model.
7. OpenCode Free is not listed. Only providers like OpenAI Codex are available.
8. Running `omniroute setup-opencode` returns:
   Cannot find package '@/shared' imported from src/shared/network/outboundUrlGuard.ts

### Expected Behavior

OpenCode Free models should appear in the Combo Builder after they have been imported successfully. The `omniroute setup-opencode` command should also complete without internal import errors.

### Actual Behavior

The OpenCode Free provider is detected correctly.

Import Models reports:

> All models already imported

However, OpenCode Free never appears inside the Combo Builder.

Additionally, running:

```bash
omniroute setup-opencode
```

fails with:

```text
Cannot find package '@/shared' imported from
src/shared/network/outboundUrlGuard.ts
```

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Add an automated integration test that verifies:

- OpenCode CLI is detected correctly.
- Imported OpenCode Free models appear in Combo Builder.
- `omniroute setup-opencode` completes successfully.
```

### Screenshots

omniroute setup-opencode

[...]

PR #11146, opened 2026-08-22
feat(routing): subscription-first auto groupings (auto/subscription, auto/thrifty) [defer to 3.8.51]

yourspraveen:
## What

Two new `auto/*` groupings for cost-conscious operators, sharing one rung model:

| Id | Policy | On exhaustion |
| --- | --- | --- |
| `auto/subscription` | plan-included connections only, documented hard-stop overage only, live quota verified per connection | **fails closed** — empty pool, never a billable fallback |
| `auto/thrifty` | all rungs ordered `subscription → keyless → free → cheap → premium` | **fails open** — steps up exactly one rung |

Both are opt-in by being requested. No existing pool, strategy, or default changes; nothing routes through them unless a caller asks for the id by name.

## Why

OmniRoute already answers *"is this model catalogued free?"* (`hidePaidModels`) and *"can this connection ever bill me?"* (`freeAccessPolicy: "strict"`), but both **fail closed** — an exhausted free pool is an empty pool, never a step up to a paid option. Every paid-side mechanism (`cost-optimized`, `budgetCap`, the `cost-saver` mode pack) is tier-agnostic. Nothing answers the question most operators actually ask:

> "Use the quota I already pay for. When it runs out, either stop, or step up one rung at a time — and come back the moment it resets."

## The blocker this had to solve first

Billing is a property of the **connection**, not the model. `classifyTier()` keys on `(provider, model)`, but the same model is plan-included through a Claude Code OAuth connection and billed per token through an API-key connection. And `auth_type` is not a safe proxy in either direction — metered OAuth connections exist, and plan-included API-key connections exist (a Copilot seat token is not a metered API key).

So this adds a **curated** per-connection billing catalog (`open-sse/config/connectionBillingCatalog.ts`), hand-set from each provider's published terms — deliberately the same pattern `FreeModelBudget.hardStopGuaranteed` already established. Uncurated resolves to `unknown` and is consumed as `metered`, so a provider added tomorrow starts *outside* the subscription rung and has to be curated in deliberately.

Providers whose overage terms allow opting into usage-based billing (`cursor`, `copilot-web`) are recorded as `unknown` overage rather than `hard-stop`, which keeps them out of `auto/subscription` while leaving them usable at rung 0 of `auto/thrifty`.

## Rungs have different exhaustion signals

This is why it is not merely a sort:

- **subscription / keyless / free** exhaust on **quota** — observable, already tracked.
[...]

PR #11154, opened 2026-08-22
fix(dashboard): keep the provider registry out of node:net (#11122)

yourspraveen:
## Problem

#11122 pointed `isLocalProvider()` at `isPrivateHost()`, imported from `src/shared/network/outboundUrlGuard.ts`. That module's first line is:

```ts
import { isIP } from "node:net";
```

`open-sse/config/providerRegistry.ts` is reachable from `ProviderDetailPageClient.tsx`, so the dashboard's client bundle can no longer resolve `node:net`. `tests/unit/media-page-client-browser-bundle.test.ts` has been red on `release/v3.8.50` since that merge — reproduced on the current tip (`367ae2fb9`):

```
✖ provider detail client entry stays browser-bundle safe
  Build failed with 1 error:
  src/shared/network/outboundUrlGuard.ts:1:21: ERROR: Could not resolve "node:net"
```

## Fix

The routing behaviour from #11122 is correct and is **left exactly as merged** — `isLocalProvider()` is untouched. Only the import path moves.

- `normalizeHost()` / `isPrivateHost()` move into a new, platform-free `src/shared/network/privateHost.ts`.
- `isIP` is swapped for `ipVersion()`, a pure-JS equivalent built from the regexes Node itself uses in `lib/internal/net.js`.
- `outboundUrlGuard.ts` re-exports `isPrivateHost`, so its ~10 existing callers are unchanged.
- `providerRegistry` imports the narrow module directly, so a future `node:` import in the guard cannot re-break the bundle.

The diff to `providerRegistry.ts` is one import line.

## Why the parity test matters

Replacing `isIP` is security-critical: a **narrower** matcher would classify a private address as public and open the very egress the SSRF guard exists to close. `tests/unit/private-host-ip-parity-11122.test.ts` therefore asserts `ipVersion` against `node:net#isIP` verdict-for-verdict, not just spot behaviour:

- valid IPv4/IPv6 literals, IPv4-mapped forms, zone ids (`fe80::1%eth0`)
- near-misses that matter — leading zeros (`010.1.1.1`), out-of-range octets, `2001:db8::1::2`
- generated permutations across both families
- an over-long input, plus the length bound that keeps the alternation ReDoS-safe (`AGENTS.md` → Regex Security)
- a re-bundle of `privateHost.ts` for the browser, pinning the invariant at the module

`privateHost.ts` also inherits the #7682 rule from its parent — no `@/`-aliased import — so the packaged CLI keeps loading `outboundUrlGuard` without a `tsconfig`.

## Verification

| Check | Result |
| --- | --- |
| `media-page-client-browser-bundle.test.ts` | red on `367ae2fb9` → **green** |
| `is-local-provider-11091.test.ts` (from #11122) | still green |
| Guard / SSRF / CLI-alias / consumer suites (16 files) | 253 pass, 0 fail |
| `typecheck:core`, `eslint`, `check:cycles` | clean |

⚠️ base-red inherited: #9985

### DIFF
--- a/changelog.d/fixes/11154-provider-registry-node-net-bundle.md
+++ b/changelog.d/fixes/11154-provider-registry-node-net-bundle.md
@@ -1,0 +1,1 @@
+- fix(dashboard): keep `open-sse/config/providerRegistry.ts` free of `node:net` so the provider detail client bundle builds again — the host classification moved to a platform-free `src/shared/network/privateHost.ts` with a pure-JS `isIP` equivalent, leaving the #11122 routing behaviour unchanged (#11154)
--- a/open-sse/config/providerRegistry.ts
+++ b/open-sse/config/providerRegistry.ts
@@ -10,7 +10,10 @@ export {
 } from "./providers/registry/alibaba/index.ts";
 export { REGISTRY } from "./providers/index.ts";
 import { REGISTRY } from "./providers/index.ts";
-import { isPrivateHost } from "@/shared/network/outboundUrlGuard";
+// Imported from `privateHost` rather than `outboundUrlGuard`: this module is reachable from
+// `ProviderDetailPageClient.tsx`, so anything it pulls in has to survive a browser bundle
+// (#11122). `privateHost` is platform-free by contract; the guard module is not.
+import { isPrivateHost } from "@/shared/network/privateHost";
 import {
   RegistryModel,
   REASONING_UNSUPPORTED,
--- a/src/shared/network/outboundUrlGuard.ts
+++ b/src/shared/network/outboundUrlGuard.ts
@@ -1,4 +1,10 @@
-import { isIP } from "node:net";
+import { ipVersion, isPrivateHost, normalizeHost } from "./privateHost";
+
+// #11122: the host classification lives in `./privateHost.ts` because
+// `open-sse/config/providerRegistry.ts` imports it from a module reachable by a browser
+// bundle, and `node:net` cannot be resolved there. Re-exported so every existing caller of
+// `isPrivateHost` from this module keeps working unchanged.
+export { isPrivateHost };
 
 export const PROVIDER_URL_BLOCKED_MESSAGE = "Blocked private or local provider URL";
 export const CLOUD_METADATA_BLOCKED_MESSAGE = "Blocked cloud-metadata endpoint";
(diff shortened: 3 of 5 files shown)


yourspraveen:
## CI triage: all 5 red checks are inherited from the base — none originate here

Verified against a clean worktree at the base this PR was cut from (`367ae2fb9`) and cross-checked against sibling PR #11146, which targets the same branch with an unrelated diff.

| Check | Failure | Verdict |
| --- | --- | --- |
| Docs Gates (fast-path) | 3 stale `157 migrations` claims vs 158 in code | inherited — #11103 |
| Fast Quality Gates | `test-discovery`, `check:dashboard-typecheck` | inherited — also red on #11146 |
| Unit Tests 1/4 – 4/4 | 17 files | inherited — 16 reproduce at `367ae2fb9`, 17th confirmed on #11146 |

### Docs Gates

```
✗ README.md — stale migrations: "157 migrations" — code has 158
✗ AGENTS.md — stale migrations: "157 migrations" — code has 158
✗ llm.txt   — stale migrations: "157 migrations" — code has 158
```

`84c9dfdd2` (#11103, config audit log) added `src/lib/db/migrations/161_config_audit_log.sql` without updating the doc counts. Base has 158 migrations and so does this branch — the diff here is 5 files and adds none. #11146 passed this gate only because its run predates #11103 landing.

### Fast Quality Gates

```
✗ [órfão NOVO] tests/unit/providers/uncloseai-noauth.test.ts     → test-discovery
✗ src/app/(dashboard)/dashboard/combos/page.tsx TS2554 (baseline 0, live 188)
✗ .../providers/[id]/components/HarImportButton.tsx TS2339        → dashboard-typecheck
```

All three are present on #11146 as well; `HarImportButton` traces to the HAR-import work in `6cd4d38e2`. Every gate that could plausibly have been affected by adding a module passes locally on this branch: `file-size`, `duplication`, `dead-code`, `type-coverage`, `cycles`, `build-scope`, `known-symbols`, `mutation-test-coverage`.

### Unit shards

All 17 CI-failing files were re-run at `367ae2fb9`; 16 fail identically there. The 17th, `tests/unit/systemd-notify.test.mjs`, is skipped on macOS (`systemd-notify and/or python3 unavailable`), so it was confirmed against #11146's shard logs instead, where it fails too.

Two files that failed in a local full-suite run but **not** in CI were also cleared: `call-log-artifact-worker.test.ts` passes in isolation (an artifact of `--test-concurrency=20`), and `8510-adobe-firefly-edits-route.test.ts` fails identically at base when run in isolation.

### What this PR does to the suite

From the shard logs on this PR's own run:

```
✔ provider detail client entry stays browser-bundle safe (933ms)   ← red on the base
✔ media page client entry stays browser-bundle safe (210ms)
✔ privateHost stays browser-bundle safe (29ms)                     ← new guard
```

Net effect is one fewer failure than the base, and no new ones. `Vitest`, `No new ESLint warnings`, `Build`, `semgrep`, `dast-smoke` and `Merge integrity` are green.

### Worth a separate look

Two of the above look like fresh base-reds not captured in the (empty) failure block of #9985: the migration-count drift from #11103, and the `uncloseai-noauth.test.ts` orphan. Both need fixing on the release branch rather than in a feature PR.

Note that `release/v3.8.50` has advanced past `367ae2fb9` since this run, so a re-run may show a different inherited set.

PR #11179, opened 2026-08-22
fix(codex): prefer max_context_window over context_window as the usable input limit

excessivechaos:
## Problem
The live Codex OAuth catalog reports BOTH `context_window` (~272K) and `max_context_window` (~872K). The discovery parser's `firstPositiveNumber` chain reads `context_window` first, so every GPT-5.6 model resolves to a 272K input limit even though 272K is only the first **pricing tier** — the real usable window is `max_context_window` (872K).

Evidence that 272K is not a hard window: `gpt-5.6-luna-xhigh` served 380-390K input tokens with HTTP 200 (call_logs, 2026-08-20 02:45-02:48) while the 922K override was active — well past 272K with no context-overflow errors.

Consequence: the context-aware combo fallback treats codex as too-small for large conversations and reorders it behind 1M-context targets, defeating the declared priority order.

## Fix
Reorder `firstPositiveNumber` so `max_context_window` wins over `context_window` when both are present; `context_window` remains the fallback for catalogs that omit the max.

## Verification
- New regression test: live-shape payload (`context_window: 272000` + `max_context_window: 872000`) resolves `inputTokenLimit: 872000`.
- End-to-end check with the real live payload: `gpt-5.6-luna` → 872000.
- Focused suite: 31 passed (provider-models-discovery-split).
- Changed-file ESLint + `git diff --check` clean.

Live payload keys verified directly against `https://chatgpt.com/backend-api/codex/models` with the account OAuth token (slug + context_window + max_context_window).

### DIFF
--- a/changelog.d/fixes/codex-max-context-window.md
+++ b/changelog.d/fixes/codex-max-context-window.md
@@ -1,1 +1,1 @@
-- fix(codex): prefer `max_context_window` over the `context_window` pricing tier as the usable input limit in discovery, so large-context requests route to codex instead of being treated as too big
+- fix(codex): prefer `max_context_window` over the `context_window` pricing tier as the usable input limit in discovery, and raise the static Codex OAuth catalog to the same usable window so the conservative discovery merge no longer caps live values at the 272K pricing tier
--- a/changelog.d/fixes/codex-max-context-window.md
+++ b/changelog.d/fixes/codex-max-context-window.md
@@ -1,0 +1,1 @@
+- fix(codex): prefer `max_context_window` over the `context_window` pricing tier as the usable input limit in discovery, so large-context requests route to codex instead of being treated as too big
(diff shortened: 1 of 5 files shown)


excessivechaos:
Second commit (852078796) — the parser fix alone was insufficient: the conservative discovery merge (`mergeCapacityLimitConservatively` = `Math.min(static, live)`) still capped the live 872K at the static 272K from `GPT_5_6_CODEX_CAPABILITIES`. Verified by reproducing the merge in-process: live 872000 + static 272000 → merged 272000.\n\nThis commit raises the static Codex OAuth catalog to `contextLength: 872000, maxInputTokens: 872000` (the usable `max_context_window`), matching the API-block precedent (GPT_5_6_API_CAPABILITIES uses 1.05M/922K). The live value now passes through the merge: live 872000 + static 872000 → merged 872000.\n\nAlso updated `codex-gpt56-catalog.test.ts` (272000 → 872000). Kiro tests unaffected (separate provider, still 272K).

PR #11098, opened 2026-08-22
fix(chat): respect per-connection model inventory in chat routing (#11089)Fix/chat routing model inventory

M-Taimoor-Mansha:
Fixes #11089.

### Summary
This PR addresses issue #11089 where chat requests were being routed to local/self-hosted provider hosts that didn't actually contain the requested model in their synced inventory, leading to spurious "model not found" errors.

### Changes
- Updated candidate connection filtering in `src/sse/services/auth.ts` (`getProviderCredentials`).
- Added a dynamic inventory check using `getSyncedAvailableModelsByConnection` for local providers (`isLocalProvider`).
- Automatically excludes connections from routing candidates if the requested model is missing from their synced inventory (`modelNotInSyncedInventory`).

PR #11186, opened 2026-08-23
fix(resilience): route chat by per-connection synced model inventory (#11089)

### LINKED ISSUE #11089: fix(resilience): Chat routing ignores per-connection model inventory on multi-host local providers
## Description

When one built-in local provider (`ollama-local`, `lm-studio`, `vllm`, …) has **multiple
connections pointing at different hosts**, chat routing does not consult the per-connection
model inventory. It will happily select a host that never had the requested model.

Typical setup:

| Connection | Priority | Base URL | Models |
| --- | --- | --- | --- |
| Mac Studio | 1 | `http://studio.lan:11434/v1` | `gemma3:4b`, `flux2-klein:9b` |
| Jetson | 2 | `http://jetson.lan:11434/v1` | `gemma3:4b` |

A request for `ollama-local/flux2-klein:9b` can be routed to **Jetson**, which does not host
that model at all, producing a spurious "model not found" instead of pinning to Studio.

> **Note on the original report:** this issue was filed as "cannot add multiple connections to
> the same built-in local provider". That premise turned out to be wrong — multi-connection is
> fully supported (see "Not a bug" below). This issue has been repurposed to track the real
> defect that the original report's *Additional Consideration* section identified.

## Steps to Reproduce

1. Add two `ollama-local` connections with distinct `providerSpecificData.baseUrl` values and
   priorities 1 and 2.
2. Let model discovery sync each connection. Inventories differ — only host 1 has
   `flux2-klein:9b`.
3. `POST /v1/chat/completions` with `"model": "ollama-local/flux2-klein:9b"`.
4. If host 1 is rate-limited/cooling, or selection strategy picks host 2, the request goes to a
   host that never had the model.

## Expected Behavior

Connection selection should:

1. Filter candidates to connections whose **synced inventory advertises the requested model**.
2. Apply priority/strategy ordering **within** that filtered set.
3. Fail over to the next priority only when a higher-priority host actually errors — never
   preemptively onto a host lacking the model.

## Actual Behavior

`getProviderCredentials()` (`src/sse/services/auth.ts:1169`) filters candidate connections at
`src/sse/services/auth.ts:1207-1213`. The only per-connection model filter there is:

```ts
if (requestedModel && isModelExcludedByConnection(requestedModel, c.providerSpecificData)) {
  connectionFilterStatus.set(c.id, "modelExcluded");
  return false;
}
```

`isModelExcludedByConnection` (`src/domain/connectionModelRules.ts:58`) reads
`providerSpecificData.excludedModels` — a **manually maintained denylist**, not derived from
discovery. Nothing consults the synced inventory.

`getSyncedAvailableModelsByConnection()` (`src/lib/db/models.ts:498`) already stores the
inventory per connection (`key_value` namespace `syncedAvailableModels`, key
`<providerId>:<connectionId>`), and is consumed by:

- `open-sse/services/autoCombo/virtualFactory.ts:522`
- `src/lib/monitoring/providerHealthMatrix.ts:406`
- `src/lib/db/models/activeSyncedCatalog.ts:100,162`
- `src/lib/db/apiKeys.ts:1531`

…but **not** by `src/sse/services/auth.ts`. That is the gap.

## Prior art in-repo

The images/embeddings path already solves exactly this. `resolveLocalSyncedEndpointRoute()`
(added in #11088, `src/lib/providerModels/syncedEndpointRouting.ts`) filters connections by
advertised model **and** endpoint, then passes the surviving `connectionIds` into
`getProviderCredentials(provider, null, connectionIds, model)`.

The plumbing already exists — `getProviderCredentials` has accepted `allowedConnections` and
`requestedModel` parameters since before this issue. Chat simply never populates the former from
inventory. The fix is to extend the same pattern into the chat selection path, most naturally as
an additional predicate in the `availableConnections` filter at `auth.ts:1207`.

## Current Workaround

Manually populate `providerSpecificData.excludedModels` on each connection with the models that
host does *not* have. Does not scale and drifts on every re-sync.

## Not a bug (verified)

Multiple connections per built-in local provider **are** supported at every layer — the original
report's headline premise does not reproduce:

- **Schema**: `idx_pc_provider` / `idx_pc_priority(provider, priority)` are plain indexes, not `UNIQUE`.
- **API**: `POST /api/providers` (`src/app/api/providers/route.ts:103`) has no per-provider
  uniqueness check; `normalizeProviderSpecificData` does not strip `baseUrl` for built-ins.
- **Runtime**: base URL resolves per connection — `open-sse/services/provider.ts:296`,
  `src/sse/services/auth.ts:2952`.
- **UI**: connections render as an N-list (`ConnectionsHeaderToolbar.tsx` branches on
  `connections.length > 1`, bulk select/ops). The only local-provider special-casing in the
  dashboard, `providerPageHelpers.ts`, governs base-URL field visibility, not connection count.
- **Existing coverage**: `tests/unit/ollama-local-embedding-2824.test.ts:180` already creates an
  `ollama-local` connection at a LAN base URL.

Verified empirically: two `ollama-local` connections at `studio.lan` / `jetson.lan` (priorities
1 and 2) both persist with distinct base URLs, and `getSyncedAvailableModelsByConnection`
returns `{studio: [gemma3:4b, flux2-klein:9b], jetson: [gemma3:4b]}`.

The `openai-compatible-*` workaround described in the original report is therefore unnecessary,
and the four capabilities it forfeits need not be forfeited.

## Test Impact

Needs a new unit test — two connections for one self-hosted provider with disjoint inventories;
assert a request for a model only host 1 advertises never selects host 2, including when host 1
is cooling.

## Related

- #11087 — Ollama capability discovery
- #11088 — Route Ollama models by advertised capability (solves this for images/embeddings)

## Environment

- OmniRoute: 3.8.49+
- Providers: `ollama-local` (also applies to `lm-studio`, `vllm`, and the rest of
  `SELF_HOSTED_CHAT_PROVIDER_IDS`)

pacocartones:
Thanks @yourspraveen for the report — it was precise enough to reproduce from the description alone.

**Problem.** When one self-hosted provider has several connections pointing at different hosts, each host keeps its own synced model inventory, but connection selection only consulted the manual `excludedModels` denylist — so a request could be routed to a host that never synced the model, surfacing as a spurious model-not-found.

**Fix.** Filter candidate connections by each connection's synced inventory, pinning the request to the host that actually advertises the model.

### Note on the path in the issue

The issue cites `src/lib/providerModels/syncedEndpointRouting.ts` as prior art. That file doesn't exist on this branch yet — it ships inside the still-open #11088, which is based on `main` and touches the images/embeddings paths. Building on it would not compile here.

So the logic went where the actual decision is made instead:

- `src/sse/services/auth.ts` — `loadAdvertisedModelsForSelfHostedConnections()`, one extra predicate in the candidate filter, and a `modelNotAdvertised` log branch alongside the existing `excluded` branch.
- `src/domain/connectionModelRules.ts` — `isModelAdvertisedByConnection()`, matching the same candidate semantics as the existing denylist helper.

It reuses `getSyncedAvailableModelsByConnection` (`src/lib/db/models.ts`), which already had five consumers; connection selection simply wasn't one of them.

**It fails open in three places** — empty inventory, malformed persisted row, or a DB read that throws all mean "unknown", never "this host lacks the model". A partial read must never silently shrink the connection pool. Scoped to `SELF_HOSTED_CHAT_PROVIDER_IDS`, so hosted providers don't pay an extra DB read.

### Evidence

**RED** — `src/` reverted to the parent commit, new test file kept:

```
tests 5 | pass 3 | fail 2
AssertionError: jetson never synced flux2-klein:9b and must not be selected for it
```

**GREEN** — with the fix: `tests 5 | pass 5 | fail 0`.

**Regression sweep** — 199 tests, 196 pass. The 3 failures (`session-affinity-generic-7274`, `provider-health-matrix`, a Codex virtual-children case) were each confirmed failing at unmodified `release/v3.8.50` too. Net: 0 regressions, +5 new passing tests.

Heads-up: `release/v3.8.50` is currently red for unrelated reasons (README migration counts, stale eslint suppressions, an incomplete Hack Club removal), so CI here may show pre-existing failures that this branch doesn't cause.

### Context

#11098 also nominally claims this issue. For maintainer awareness it is currently `CONFLICTING`, based on `main` rather than `release/v3.8.50`, and spans 3431 files / +583818 lines. Noting it only so the overlap isn't a surprise.

4 files, +287/-1. Rebased on `release/v3.8.50` @ `92ef3c71e`.

Closes #11089


pacocartones:
@diegosouzapw CI note: the red checks here are pre-existing on `release/v3.8.50`, not from this PR.

- **Unit Tests fast-path (1-4/4)** - the same 4 shards fail identically on the other open PRs against this base (#11187, #11179, #11178). I also reverted just this PR's `src/` back to the base and re-ran the shard-1 failures locally: `openrouter-free-model-credits-exhausted` (3 cases) and `source-scanner-guards` still fail, `tests 8 | pass 4 | fail 4`. Same result with and without this change.
- **No new ESLint warnings** - fails with `There are suppressions left that do not occur anymore`, a stale-suppression prune issue. No warning is attributed to either file this PR touches.

The 5 new tests did run and pass inside CI on shard 1:

```
ok #11089 selects only the host whose synced inventory advertises the model
ok #11089 does not preemptively fail over to a host lacking the model when the owner is cooling
ok #11089 keeps the connection that does advertise the model selectable
ok #11089 a model advertised by every host leaves both connections eligible
ok #11089 fails open when the provider has no synced inventory at all
```

Green here: Build, Docs Gates, Vitest, Change Classification, Merge integrity (changelog), semgrep, dast-smoke. Happy to rebase once the base-reds are drained.

PR #11190, opened 2026-08-23
feat(api): structured ?format=json for the self-service usage endpoint

diegosouzapw:
`GET /api/usage/om-usage` already let any API key read **its own** usage — the personal
daily/weekly USD limits and the provider quota snapshot — but only as `text/plain`. A UI can't
parse that safely, which is exactly what
[OmniCopilot #8](https://github.com/diegosouzapw/OmniCopilot/issues/8) asks for: a key holder
seeing their own spend and reset.

## Change

`?format=json` returns the same `ApiKeyUsageLimitStatus` + `UsageSnapshot` the text is rendered
from. **Text and JSON share the exact same collectors** (`collectUsageSnapshots`,
`getApiKeyUsageLimitStatus`), so the two forms can never disagree about a number.

The response is a discriminated union, which is the part that makes it safe for a panel:

- success → `{ allowed: true, personal, provider }`
- bad key (401) / key without `allowUsageCommand` (403) → `{ allowed: false, error: { message } }`

A caller can now tell *"this key may not ask"* apart from *"allowed, but nothing cached yet"* —
a present-but-empty state that plain text could never express. Without `?format` the text
contract is byte-identical to before.

## Auth model

Unchanged and deliberately so: this is the **self-service** surface. The caller's own Bearer key,
validated with `isValidApiKey`, gated by the per-key `allowUsageCommand` flag (off by default).
It does not touch `/api/keys/…`, which stays behind `requireManagementAuth`.

## Docs

The endpoint was **absent** from `docs/reference/API_REFERENCE.md` — not undocumented by this PR,
never documented at all. It now has a section covering both forms, the gate, and the auth model.
`check:docs-all` passes (doc-links, fabricated-docs).

## Tests

`tests/unit/usage-command-json-format.test.ts` — 4 tests: the JSON shape, the text default
preserved, the structured 403, and a 401 whose error body carries no stack trace. The existing
`internal-usage-command` suite still passes 12/12.

⚠️ base-red inherited: #9985 — the unit suite has pre-existing reds on the release tip
(catalogued in the issue). This PR's files are green in isolation.


diegosouzapw:
Rebased onto the current release tip (`62ab93d789`) — the branch contains it, and the diff
against it is exactly the four files of this PR (usage command + its test + API_REFERENCE +
changelog fragment).

The unit reds the full run shows are the inherited base-reds from #9985 (GLM import, Antigravity
aliases, model-sync, …), not this change — a full-suite run on the release tip reproduces them
without this PR. This PR's own surface:

- `tests/unit/usage-command-json-format.test.ts` — 4/4
- `tests/unit/internal-usage-command.test.ts` (pre-existing suite, touched module) — 12/12
- `check:docs-all` — doc-links + fabricated-docs clean
- `typecheck:core` — clean

#11192 (per-connection `providers[]`) is stacked on this and stays CLEAN because it doesn't run
the full suite; merge this first, retarget #11192 to the release branch, then merge it.

PR #11187, opened 2026-08-23
feat(routing): pre-compile model prefix regexes in cloudCodeThinking service

rqzbeh:
### Summary

Optimizes model normalization performance in `open-sse/services/cloudCodeThinking.ts`:

1. **Pre-compiled Single-Pass Regex (`PREFIX_TRIM_RE`)**:
   - Replaced multiple sequential string replacement calls (`.replace(/^models\//i, "").replace(/^antigravity\//i, "")`) with a pre-compiled single-pass regex (`/^(?:models\/|antigravity\/)+/i`).
   - Eliminates redundant string allocations on hot model translation paths.

### Verification
- `node --import tsx/esm --test tests/unit/ai-proxy-routing-optimization.test.ts` (PASS)
- `npm run typecheck:core` (0 errors)
- `npm run check:docs-all` (PASS)

Issue #11176, opened 2026-08-22
Hack Club AI removal (#11118) is incomplete — provider still in the shared catalog, splits the two registries


yourspraveen:
### Description

The Hack Club AI removal in #11118 / #11123 (`aa1287368`) is incomplete. It removed the provider from the open-sse registry, the free-model catalog, the docs and added migration `162_remove_hackclub_provider.sql` — but the **full provider entry remains in the shared catalog** at `src/shared/constants/providers/apikey/gateways.ts:642`:

```ts
hackclub: {
  id: "hackclub",
  alias: "hc",
  name: "Hackclub AI",
  website: "https://ai.hackclub.com",
  hasFree: true,
  freeNote: "Free AI for Hack Club members — 30+ models, no credit card.",
  authHint: "Sign in with your Hack Club account at ai.hackclub.com.",
  passthroughModels: true,
  ...
}
```

Measured on `release/v3.8.50` @ `80d931ae2`:

```
total providers:        348
hackclub present:       true
resolveProviderId("hc"): hackclub
```

So the provider still counts toward the 348, and still surfaces its name, website and sign-in hint. Since the removal was **requested by the Hack Club maintainers** (#11118), the residual catalog entry looks contrary to the intent of that request.

Residual references besides the catalog entry:

- `src/shared/constants/providers/web-cookie.ts:265` — comment reserving the `hc` alias for hackclub
- `open-sse/config/providers/registry/huggingchat/index.ts:5` — same comment
- `tests/unit/provider-alias-uniqueness.test.ts:50` — still asserts the open-sse alias

### Steps to Reproduce

1. Check out `release/v3.8.50` at `80d931ae2`
2. Run `node --import tsx/esm --test tests/unit/provider-alias-uniqueness.test.ts`

### Current Behavior

The split state makes `provider-alias-uniqueness` fail, which is a **current base-red on `release/v3.8.50`** (reported on #9985 — it reddens every PR targeting the branch):

```
✖ primary providers keep the short alias; web variants use their own id
  AssertionError: undefined !== 'hc'
```

`PROVIDER_ID_TO_ALIAS.hackclub` is now `undefined` (removed from the open-sse registry) while `resolveProviderId("hc")` still returns `"hackclub"` (shared catalog untouched) — the two registries disagree.

### Expected Behavior

Either the catalog entry is removed too (completing the requested removal), or the provider is intentionally retained and the open-sse registry entry is restored. Right now it is half of each.

[...]

PR #11193, opened 2026-08-23
fix(routing): forward lkgpEnabled into RoutingContext so the LKGP toggle works

### LINKED ISSUE #11181: [BUG]   Settings toggle lkgpEnabled never reaches RoutingContext, LKGP cannot be disabled
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

  Debian GNU/Linux 13 (trixie), kernel 6.12.74+deb13+1-amd64

### Node.js Version

_No response_

### Provider(s) Involved

Provider-agnostic — affects any auto/* routing regardless of provider

### Model(s) Involved

auto (zero-config virtual combo)

### Client Tool

cURL, OpenCode

### Description

Settings → Routing exposes an LKGP toggle. It renders, it writes, and the value persists — but it never reaches the routing code that reads it, so turning LKGP off has no effect.

`lkgpEnabled` appears exactly five times in the codebase, in two disconnected halves.

**Write path (works):**

`src/app/(dashboard)/dashboard/settings/components/RoutingTab.tsx:1543`
```tsx
checked={settings.lkgpEnabled !== false}
onChange={(checked) => updateSetting({ lkgpEnabled: checked })}
```

`src/shared/validation/settingsSchemas.ts:325`
```ts
lkgpEnabled: z.boolean().optional(),
```
(declared under a `// Missing settings` comment)

**Read path (never reached):**

`open-sse/services/autoCombo/routerStrategy.ts:33`
```ts
lkgpEnabled?: boolean;
```

`open-sse/services/autoCombo/routerStrategy.ts:306`
```ts
select(pool: ProviderCandidate[], context: RoutingContext): RoutingDecision {
  if (context.lkgpEnabled === false) {
    return getStrategy("rules").select(pool, context);
  }
```

**The gap:** nothing populates `lkgpEnabled` on the context object. The only construction site is `open-sse/services/combo/resolveAutoStrategy.ts:322`:

```ts
const decision = selectWithStrategy(
  routableCandidates,
  {
    taskType,
    requestHasTools,
    lastKnownGoodProvider,
    estimatedInputTokens,
    sla: slaPolicy,
  },
  routingStrategy
);
```

`lkgpEnabled` is absent, so `context.lkgpEnabled` is always `undefined`. The `=== false` comparison never matches and the early return is dead code.

This looks like a regression rather than an omission. #955 shipped this feature in April 2026, described in its own summary as *"LKGP UI — toggle + cache clear in Settings → Routing, `lkgpEnabled` guard in strategy"*. Both halves are still present; only the plumbing between them is gone. #6063 (`refactor(sse): extract applyStrategyOrdering leaf from handleComboChat`) is a plausible point where the context construction was rewritten, though I have not bisected it.

**Impact.** LKGP re-pins on every success and is the hardcoded strategy for every `auto/*` variant (`virtualFactory.ts:548`, `let routerStrategy = "lkgp"`). With `explorationRate` at 0.05, a single exploration that lands on an expensive provider pins it and every subsequent request follows — the toggle is the documented way out, and it does not work. `DELETE /api/settings/lkgp-cache` clears the pin, but the next success re-pins immediately.

### Steps to Reproduce

1. Have at least two providers eligible in the `auto` pool.
2. Send `{"model":"auto"}` requests until an LKGP pin is established (visible in `key_value`, namespace `lkgp`, key `auto:auto`).
3. Turn the LKGP toggle off in Settings → Routing, or via API:
   ```
   curl -s -b cookie.txt -X PATCH http://<host>:20128/api/settings \
     -H 'Content-Type: application/json' -d '{"lkgpEnabled":false}'
   ```
4. Confirm it persisted:
   ```
   curl -s -b cookie.txt http://<host>:20128/api/settings | jq .lkgpEnabled
   # false
   ```
5. Send another `{"model":"auto"}` request.
6. Check the routing log line.

### Expected Behavior

With `lkgpEnabled: false`, LKGPStrategyImpl should delegate to the `rules` strategy, and the routing log should report a rules-based decision with a score.

### Actual Behavior

LKGP is still applied. The log line is unchanged:

```
{"tag":"COMBO","msg":"Auto selection: claude/claude-fable-5 | intent=medium task=default | strategy=lkgp | LKGP: using last known good provider claude"}
```

Verified on 3.8.49 with `lkgpEnabled: false` confirmed persisted via `GET /api/settings` immediately before the request.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
# lkgpEnabled confirmed false immediately before the request
lkgpEnabled = False

# routing log for the request that followed
{"level":30,"service":"omniroute","module":"sse","tag":"COMBO","msg":"Auto selection: claude/claude-fable-5 | intent=medium task=default | strategy=lkgp | LKGP: using last known good provider claude"}
```

### Screenshots

_No response_

### Additional Context

- Single-node homelab, one connection per provider, 4 providers in the auto pool.
- Reproduced with `promptCacheAffinityEnabled: false` and `disableSessionStickiness: true`, to rule out those paths as the cause of the stickiness.
- The `// Missing settings` comment above the schema entry suggests the field was added to satisfy validation for a value the UI was already sending, rather than as part of a wired-up feature.

### Validation Plan

- PATCH /api/settings {"lkgpEnabled": false} → next {"model":"auto"} request logs `strategy=rules` with a score, not `strategy=lkgp`
- PATCH /api/settings {"lkgpEnabled": true} → LKGP behaviour returns
- npm run test:vitest

pacocartones:
## Summary

Settings → Routing has an LKGP toggle that persists correctly but has no effect: `lkgpEnabled` was never forwarded into the `RoutingContext`, so the guard that reads it could never fire.

`LKGPStrategyImpl.select()` guards with `context.lkgpEnabled === false` (`open-sse/services/autoCombo/routerStrategy.ts:306`) and the setting is accepted by the schema (`settingsSchemas.ts:438`) and written by the UI (`RoutingTab.tsx:1526-1527`). But the only production construction site for that context, `resolveAutoStrategyOrder()` in `open-sse/services/combo/resolveAutoStrategy.ts`, built a five-key literal that omitted the field. `context.lkgpEnabled` was therefore always `undefined`, `=== false` never matched, and the early return was unreachable. This matters because `lkgp` is the hardcoded strategy for every `auto/*` variant, so a single exploration hop can pin an expensive provider with no working way out.

**Fix:** forward `settings.lkgpEnabled` into the context literal, alongside the `settings`-derived values that path already reads (`compatFilterFailOpen`, the intent config). Six lines, four of them comment.

Deliberately not widened: `applyStrategyOrdering()` has a separate `strategy === "lkgp"` branch for explicitly-configured combos, but it never receives `settings` at all, so honouring the toggle there is a larger dependency change and a different bug from the one reported.

## Related Issues

- Closes #11181

## Validation

- [x] Change type: **routing**
- [ ] `npm run lint` — could not run locally (`eslint` aborts in this checkout with `Cannot find module './UTF16SurrogatePairToCodePoint'` from `es-abstract`, an environment problem, not a code one). Both touched files are Prettier-clean, and `tsc --noEmit` reports nothing for either.
- [x] Reconciled with the current active release base (`release/v3.8.50` @ `60f25eff9`); focused checks rerun afterward

The new test is the point of the change, so here is its red/green, verbatim.

**Without the fix** (production file reverted to `HEAD~1`, new test kept):

```
✔ control — with lkgpEnabled unset the LKGP pin still wins (guard must not over-fire) (14.0986ms)
✖ #11181 — a persisted lkgpEnabled:false makes the lkgp strategy delegate to rules (5.2014ms)
ℹ pass 1
ℹ fail 1

  AssertionError [ERR_ASSERTION]: lkgpEnabled:false must disable LKGP selection, got: Auto selection: pricey-model | intent=medium task=default | strategy=lkgp | LKGP: using last known good provider anthropic
    actual: 'Auto selection: pricey-model | intent=medium task=default | strategy=lkgp | LKGP: using last known good provider anthropic',
    expected: /LKGP: using last known good provider/,
    operator: 'doesNotMatch',
```

That failure line is character-for-character the symptom in the issue report.

**With the fix:**

```
✔ control — with lkgpEnabled unset the LKGP pin still wins (guard must not over-fire) (16.3322ms)
✔ #11181 — a persisted lkgpEnabled:false makes the lkgp strategy delegate to rules (5.2415ms)
ℹ pass 2
ℹ fail 0
```

Regression check over the four adjacent suites (`router-strategies`, `combo-resolve-auto-strategy-split`, `combo-auto-config-split`, `combo-routing-engine`): **119 passed, 0 failed.**

## Tests Added Or Updated

- **Added** `tests/unit/lkgp-enabled-context-11181.test.ts` (2 tests). It drives `resolveAutoStrategyOrder()` end-to-end with a real persisted LKGP pin and a real `settings` object, then asserts on the emitted `Auto selection:` log line — which is exactly the observable the issue reports on.
  - The failing case asserts `lkgpEnabled: false` produces a `RulesStrategy` decision, not an LKGP one.
  - A control case asserts the pin *still* wins when the setting is absent, so the guard cannot pass by over-firing.

Worth flagging, because it is the reason this shipped broken: `tests/unit/router-strategies.test.ts:187` (`lkgp — lkgpEnabled:false delegates to rules`) already existed and **passes on unpatched `HEAD`**. It hands `LKGPStrategyImpl` a context it constructs itself, so it verifies the guard while never touching the production construction site. Coverage at the strategy level cannot see this class of bug; the new test sits at the `resolveAutoStrategy` level for that reason.

## Coverage Notes

Touches one `open-sse/` file. The changed line is covered by the new test in both directions (set-to-false and absent), verified by the red/green above.

## Reviewer Notes

- No migration, no feature flag, no API change. The read is optional-chained and cast to `boolean | undefined`, so `settings: null` and a missing field both behave exactly as before — only an explicit `false` changes behaviour, matching the existing `=== false` guard.
- @diegosouzapw — the failing checks on this PR are pre-existing base reds on `release/v3.8.50`, not from this change. The same Fast Quality Gates and the same 4 unit shards fail on #11189 (`rqzbeh`), which touches nothing in common with this. Happy to rebase once the base is green.


pacocartones:
@diegosouzapw — flagging the red checks here so they don't cost you triage time: **none of them come from this PR**, and the two new tests in it pass.

**ESLint ("No new ESLint warnings")** — fails with:

```
There are suppressions left that do not occur anymore. Consider re-running the command with `--prune-suppressions`.
```

That is a stale-baseline error, not a new warning. It reports suppressed violations that have since been *fixed* upstream, so it fires on whatever sits on the current tip regardless of content. Evidence it is not mine:

- Your own **#11192** (base `eb9fa33ee`) fails the same job.
- **#11189** and **#11187** (both on the older base `92ef3c71e`) **pass** it.
- `config/quality/eslint-suppressions.json` has **zero** entries for `resolveAutoStrategy.ts`, so this PR cannot add or orphan a suppression. It touches no suppressed file.

I've left it alone rather than running `--prune-suppressions`, since re-baselining the ratchet is a release-wide call and would bury a 6-line routing fix under an unrelated config churn. Happy to do it in a separate PR if that's useful.

**Unit Tests fast-path (1-4/4)** — 27 failures across the four shards, all pre-existing: the chatCore/DeepSeek reasoning-replay group, `getProviderCredentials` OpenRouter gating, the SSRF/R2 case, the OpenCode config-schema group, the i18n `en.json` drift check, and `primary providers keep the short alias` (the Hack Club split-registry red from #11176). Not one is in a file this PR touches.

The two tests added here both pass in shard 4:

```
✔ control — with lkgpEnabled unset the LKGP pin still wins (guard must not over-fire) (17.447912ms)
✔ #11181 — a persisted lkgpEnabled:false makes the lkgp strategy delegate to rules (3.484469ms)
```

There is also incidental confirmation the fix works on a real CI run — this line from shard 1 shows the `lkgp` strategy delegating to the rules scorer, which is precisely the behaviour that was unreachable before:

```
[sse] Auto selection: openai/gpt-4.1-nano | intent=medium task=default | strategy=lkgp | RulesStrategy: score=0.712 (quota=1.00, health=1.00, cost=0.99, taskFit=0.50)
```

Green on this PR: Build, Vitest, Docs Gates, Change Classification, Merge integrity, dast-smoke, semgrep.

Glad to rebase once the base is green.

Issue #11167, opened 2026-08-22
Wayfinder: Beginner UX communication layer — 10 small PRs


ignamiranda:
## Destination

Contribute the 10 UX improvements from the beginner-UX audit as small, standalone, well-presented PRs to the OmniRoute repo. Each PR should be mergeable on its own merits.

**Base branch:** `release/v3.8.50` (highest active release branch per CONTRIBUTION_GOLDEN_PATH.md — re-verify before cutting each branch)

## Why this is foggy

- 10 independent improvements with unclear sequencing
- Each PR needs to be small, standalone, and well-presented to avoid rejection
- Need to decide: order, batching, naming, evidence requirements
- Need to understand what "well-presented" means for this repo

## The 10 opportunities (from the pitch)

1. **Resilience connections** — Plain-language states before breaker mechanics
2. **Navigation** — Essentials mode with advanced tools grouped
3. **First-run** — Readiness-based onboarding path
4. **Endpoint** — Guided connection page with recommended URL
5. **API Keys** — Purpose-first header before the list
6. **Embedded Services** — Value-first labels and comparison cards
7. **Batch Processing** — Stable header with 3-step strip
8. **ACP Agents** — Outcome-first framing for advanced feature
9. **Traffic Inspector** — Separate observability from privileged interception
10. **Health** — Interpreted readiness before raw telemetry

## Decision tickets

| #   | Ticket                                                                                      | Decision                                                                                     |
| --- | ------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------- |
| D1  | [#11174 — PR sequencing](https://github.com/diegosouzapw/OmniRoute/issues/11174)            | ✅ 4 waves by risk: copy-only → component copy → structural UI → largest                     |
| D2  | [#11169 — PR scope](https://github.com/diegosouzapw/OmniRoute/issues/11169)                 | ✅ One dashboard area + i18n strings + one focused test per PR; < ~300 LOC                   |
| D3  | [#11170 — Landing zones](https://github.com/diegosouzapw/OmniRoute/issues/11170)            | ✅ All 10 land in existing page dirs (verified); no new top-level routes                     |
| D4  | [#11171 — Evidence requirements](https://github.com/diegosouzapw/OmniRoute/issues/11171)    | ✅ Focused test + lint + before/after screenshots; TDD path, no VPS needed                   |
[...]

PR #11195, opened 2026-08-23
feat(dashboard): orient users before API key management

ignamiranda:
## Summary

- add a purpose-first heading to the API key management page using existing localized copy
- show the request path from an app through an API key to OmniRoute
- keep the existing key-management controls and behavior unchanged

## Why

The page currently opens directly on filters and the registered-keys list. The new orientation layer explains what API keys do before users manage them.

Part of #11167.

## Validation

- `node --import tsx/esm --test tests/unit/api-manager-page-static.test.ts`
- `npm run lint` — 0 errors; 6 pre-existing warnings
- `npm run typecheck:core`

## Scope

Two files: the existing page client and its static regression test. No API, database, routing, or key-management behavior changes.

> ⚠️ base-red inherited: #9985


ignamiranda:
Addressed review findings in 2295d0948: localized all request-flow labels, added an accessible group label, made the flow responsive, and strengthened the regression test to assert translation usage. Verification: focused test (11/11), npm run lint, npm run typecheck:core.

Issue #11118, opened 2026-08-22
[BUG] Please remove Hack Club AI


skyfallwastaken:
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### Provider(s) Involved

Hack Club AI

### Description

Please remove the Hack Club AI provider from OmniRoute. I work at Hack Club and it's against our Terms of Service for Hack Club AI.

Furthermore, Hack Club AI automatically blocks coding agents, and issues sanctions for agent use.

### Steps to Reproduce

1. Open OmniRoute
2. See how there is a Hack Club provider

### Expected Behavior

No Hack Club AI provider

### Actual Behavior

Has Hack Club AI provider

### Test Impact

Unsure

### Error Logs / Output

```shell

```

Issue #6108, opened 2026-07-03
fix(providers): NVIDIA NIM static model registry is stale — glm-5.1 EOL'd 2026-07-02 still listed, live glm-5.2 / minimax-m3 / nemotron-3-ultra missing


andrea-kingautomation:
## Summary
The NVIDIA NIM provider's model list (`open-sse/config/providers/registry/nvidia/index.ts`) is a static array that has drifted from the live catalog. Users see dead models in the picker and cannot select live ones, and there is no auto-refresh, so every drift persists until a release edits the file.

## Evidence (all verified 2026-07-03 with a direct NVIDIA API key, no router in the path)

`GET https://integrate.api.nvidia.com/v1/models` returns 121 models. Diff against the shipped registry (v3.8.42):

| model | live NIM catalog | OmniRoute registry | direct chat probe |
|---|---|---|---|
| `z-ai/glm-5.1` | absent | **listed** (first entry) | **410 Gone**: "The model 'z-ai/glm-5.1' has reached its end of life on 2026-07-02T00:00:00Z and is no longer available." |
| `z-ai/glm-5.2` | present | missing | 200 in 0.5s |
| `minimaxai/minimax-m3` | present | excluded per #3329 (404 at the time) | **200 with real inference** (25.8s, slow reasoner) — the #3329 re-add condition is now met |
| `nvidia/nemotron-3-ultra-550b-a55b` | present | missing | 200 in 17.2s |
| `stepfun-ai/step-3.7-flash` | present | listed | 200 in 2.3s |

So since 2026-07-02 the registry's first NVIDIA entry is a permanently dead model (every call burns a 410), while the model users actually want (glm-5.2) is invisible in the dashboard.

## Suggested fix
1. Short term: refresh the static list (drop glm-5.1, add glm-5.2, nemotron-3-ultra; re-add minimax-m3 per #3329's own re-add criterion).
2. Structural: NIM exposes a standard `/v1/models`; enrich or validate the per-connection catalog from it periodically (same spirit as #5004 for context windows). A daily reconcile would have flagged the glm-5.1 EOL the day it happened and surfaced glm-5.2 automatically. Models that EOL could be marked dead instead of being retried.

Happy to provide the raw catalog dump or probe bodies if useful.

diegosouzapw:
[...]

Issue #10523, opened 2026-08-16
base-red: ServiceSupervisor.test.ts adopted-service test is order-dependent (port 29996)


diegosouzapw:
## Red detectado no tip da release/v3.8.50 (aa5b77eb6)

**Falha**: `tests/unit/services/ServiceSupervisor.test.ts` — "adopted service resolves and records the real pid" (linha 246) retorna `state === 'error'` em vez de `'running'`.

**NÃO é defeito de PR individual — é base-red.**

### Diagnóstico: ordem-dependência entre dois testes do MESMO arquivo

Os testes das linhas 216 e 246 usam a **mesma porta 29996** com `startHealthServer(29996)` inline e `probeBeforeSpawn: true`:

- `test #6205: probeBeforeSpawn adopts a healthy existing instance (no spawn)` (linha 216) — roda PRIMEIRO
- `adopted service resolves and records the real pid...` (linha 246) — roda DEPOIS, usa 29996 de novo

### Repro (base pura, devbox non-root)

```bash
# Isolado → PASSA (1 pass / 0 fail)
node --import tsx/esm --test --test-name-pattern="adopted service resolves" tests/unit/services/ServiceSupervisor.test.ts

# Ordem 216→246 (pattern de ambos) → FALHA (1 pass / 1 fail)
node --import tsx/esm --test --test-name-pattern="probeBeforeSpawn adopts|adopted service resolves" tests/unit/services/ServiceSupervisor.test.ts
```

Resultado: `'error' !== 'running'`, consistente com TIME_WAIT / EADDRINUSE na porta 29996 quando o teste anterior já a usou e fechou no mesmo processo. Porta fica livre entre execuções isoladas (verificado via `ss -tlnp`).

### Impacto
- Merge-train do /merge-prs (suíte FULL) fica bloqueado indevidamente ao alcançar este teste.
- CI pode não expor: o arquivo roda como unidade e o bug depende da ordem interna dos dois testes.
- Nenhum dos 25 PRs ⭐4-5 em análise toca ServiceSupervisor — red 100% herdado da base.

### Fix sugerido (fora da campanha, PR próprio)
Usar porta dinâmica (`port 0` + ler a porta real do healthServer) OU reutilizar o mesmo `Server` para ambos os testes sem fechar/recriar no meio (evita TIME_WAIT de ~60s do SO).

mmzac1:
[...]

PR #11201, opened 2026-08-23
fix(ci): green the release/v3.8.50 base (#9985)

jonlwheat2-gif:
## What

Finishes greening the `release/v3.8.50` base (#9985). **Rebuilt on the current base (`79c5bdf68`)** — the owner's #10964 already fixed the first 7 failures, so this PR now carries only the remaining base-red failures:

| File | Issue | Fix |
|---|---|---|
| `tests/unit/guide-settings-route.test.ts` | v2 OpenCode dual-write — #10964 updated the model-limit blocks but missed the stale `content.providers === undefined` assert (line 130) | Assert the v2 `providers.omniroute.package/settings` shape |
| `tests/unit/t40-opencode-cli-tools-integration.test.ts` | stale v1-only assertion | Assert v2 `providers.omniroute.package` |
| `tests/unit/clinepass-provider.test.ts` | `zai/glm-5.2` stale id | Use `z-ai/glm-5.2` vendor prefix (#6108) |
| `tests/unit/provider-alias-uniqueness.test.ts` | 3 stale hackclub assertions | Remove — Hack Club AI removed (#11118) |
| `config/quality/eslint-suppressions.json` | 2 stale suppressions | Prune (`call-logs`, `rerank`) |
| `tests/unit/cli-combo-command.test.ts` | **#11162 made combo create refuse no-model combos; this file was missed** (reported as #11203) | Pass `models: ["openai/gpt-4o-mini"]` to the create calls, mirroring #11162's sibling-test fix |

## Verification (Linux, Docker, Node 24.19.0)

- guide-settings: **4/4** · t40: **9/9** · clinepass: **21/21** · provider-alias: **6/6** · cli-combo: **5/5** (pristine was 2/5)
- ESLint gate clean (pruned suppressions, no "stale suppressions" warning)
- `check:dashboard-typecheck`: 0 regressions vs frozen baseline (owner's #10964 already fixed combos/HarImport)
- Typecheck, docs-sync, any-budget, tracked-artifacts: green locally

## Note

The 7 originally-listed fixes (combos TS2554, HarImport TS2339, aihorde SSRF guard, config-generator 128K, quota 403, openrouter allExpired, pt-BR keys) are **no longer in this PR** — the owner's #10964 landed equivalent fixes in the base. Thanks to the reviewer for flagging cli-combo-command (#11203).

Closes the remaining base-red tail of #9985 so #11194 and other dependent PRs get a green base.

### DIFF
--- a/config/quality/eslint-suppressions.json
+++ b/config/quality/eslint-suppressions.json
@@ -853,11 +853,6 @@
       "count": 1
     }
   },
-  "src/app/api/usage/call-logs/route.ts": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/app/api/usage/quota/route.ts": {
     "no-restricted-imports": {
       "count": 1
@@ -949,11 +944,6 @@
     }
   },
   "src/app/api/v1/models/catalog.ts": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
-  "src/app/api/v1/rerank/route.ts": {
     "no-restricted-imports": {
       "count": 1
     }
--- a/tests/unit/cli-combo-command.test.ts
+++ b/tests/unit/cli-combo-command.test.ts
@@ -37,7 +37,11 @@ async function withComboEnv(fn: (dataDir: string) => Promise<void>) {
 test("combo create inserts a new combo via db module", async () => {
   await withComboEnv(async () => {
     const { runComboCreateCommand } = await import("../../bin/cli/commands/combo.mjs");
-    const result = await runComboCreateCommand("my-combo", "priority", {});
+    // #11162: combo create refuses combos without any model — pass a model
+    // like the sibling tests updated in that commit.
+    const result = await runComboCreateCommand("my-combo", "priority", {
+      models: ["openai/gpt-4o-mini"],
+    });
     assert.equal(result, 0);
 
     // Verify via the same db module
@@ -53,10 +57,12 @@ test("combo create fails if combo already exists", async () => {
   await withComboEnv(async () => {
     const { runComboCreateCommand } = await import("../../bin/cli/commands/combo.mjs");
 
-    await runComboCreateCommand("dup-combo", "auto", {});
+    await runComboCreateCommand("dup-combo", "auto", { models: ["openai/gpt-4o-mini"] });
     const originalError = console.error;
     console.error = () => {};
-    const result = await runComboCreateCommand("dup-combo", "auto", {});
+    const result = await runComboCreateCommand("dup-combo", "auto", {
+      models: ["openai/gpt-4o-mini"],
+    });
     console.error = originalError;
 
     assert.equal(result, 1);
@@ -68,7 +74,7 @@ test("combo delete removes the combo", async () => {
     const { runComboCreateCommand, runComboDeleteCommand } =
       await import("../../bin/cli/commands/combo.mjs");
 
-    await runComboCreateCommand("to-delete", "weighted", {});
+    await runComboCreateCommand("to-delete", "weighted", { models: ["openai/gpt-4o-mini"] });
     const result = await runComboDeleteCommand("to-delete", { yes: true });
     assert.equal(result, 0);
 
@@ -91,7 +97,7 @@ test("combo switch updates active combo when server is offline", async () => {
     const { runComboCreateCommand, runComboSwitchCommand } =
       await import("../../bin/cli/commands/combo.mjs");
 
-    await runComboCreateCommand("my-switch", "round-robin", {});
+    await runComboCreateCommand("my-switch", "round-robin", { models: ["openai/gpt-4o-mini"] });
     const result = await runComboSwitchCommand("my-switch", {});
     assert.equal(result, 0);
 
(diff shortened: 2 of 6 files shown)


pacocartones:
@jonlwheat2-gif this PR is the right shape and it is close, but there is a fourth failing test file that it does not cover, so the base will still be red after it merges. I reproduced it on Linux and traced it, so here is everything you need.

### What is missing

`tests/unit/cli-combo-command.test.ts` — 3 of 5 tests fail with `AssertionError [ERR_ASSERTION]: Expected values to be strictly equal: 1 !== 0`:

- `combo create inserts a new combo via db module` (line 41)
- `combo delete removes the combo` (line 73)
- `combo switch updates active combo when server is offline` (line 96)

Same class as your 8 test updates: a deliberate contract change whose sibling tests were updated while this file was missed.

### Cause

`b3844550d` (*"fix(api): refuse creating a routing combo without any model (#11162)"*) made `combo create` refuse a combo with no targets:

```
combo create requires at least one target. Pass --models <provider/model,...> and/or repeat --model <provider/model>.
```

That commit updated 5 sibling combo tests but not this one. The three failures all call `runComboCreateCommand(..., {})` with no `models`, so the command now returns exit code `1`; the assertion `1 !== 0` is just that exit code surfacing.

`delete` and `switch` fail as collateral because they each create a combo first as setup.

### Reproduction

Fresh clone at tip `79c5bdf681d693add5f96b462d6053f0a6bf151b`, Linux, Node v24.19.0:

```
$ DISABLE_SQLITE_AUTO_BACKUP=true node --import tsx/esm \
    --test tests/unit/cli-combo-command.test.ts

ℹ tests 5
ℹ pass 2
ℹ fail 3
```

Worth noting for triage: this is **not** order-dependence, unlike the `ServiceSupervisor` case in #10523. `combo create` fails **in isolation** as well (`pass 0 / fail 1`), while `combo list returns 0 with empty combos table` **passes in isolation** and only fails in sequence.

I confirmed the root cause by calling the command directly with the server stubbed offline, which surfaces the error message that the exit code otherwise hides.

### The fix

Passing a model to each of the three `runComboCreateCommand` calls, mirroring exactly what `b3844550d` did to its sibling tests, turns the file green:

```
ℹ tests 5
ℹ pass 5
ℹ fail 0
```

I have written this up in more detail as #11203. I am deliberately **not** opening a competing PR, since yours already owns the base-green work and it belongs in one place. Happy to hand over anything else useful, or to open a separate PR for just this file if you would rather keep your diff as-is.

jonlwheat2-gif:
Thanks @pacocartones — great catch, and the write-up in #11203 was spot on. The fix is now in this PR (and I've rebuilt it onto the current base `79c5bdf68`).

## What I did

1. **Reproduced on Linux** (Docker, Node 24.19.0, fresh clone at `79c5bdf68`): `cli-combo-command.test.ts` → **2 pass / 3 fail** (`create`/`delete`/`switch`, all `1 !== 0`), exactly as you reported.
2. **Applied the same fix #11162 used on its sibling tests** — pass `models: ["openai/gpt-4o-mini"]` to the four `runComboCreateCommand` calls (including the dup-combo setup calls, so the duplicate-refusal path still exercises a real existing combo).
3. **Verified on Linux: 5/5 pass.**

## Rebuild notes (since the base moved)

- The base now includes **#10964**, which already fixed 7 of the original 12 items (combos TS2554, HarImport TS2339, aihorde SSRF guard, config-generator 128K, quota 403, openrouter allExpired, pt-BR keys). Those are **no longer in this PR** — no point shipping what the base already has.
- This PR now carries only the remaining base-red tail, all verified on Linux:
  - `cli-combo-command.test.ts` (yours, #11203) — **5/5**
  - `guide-settings-route.test.ts` — #10964 fixed the model-limit blocks but missed the stale `content.providers === undefined` (line 130); v2 assertion added — **4/4**
  - `t40-opencode-cli-tools-integration.test.ts` — **9/9**
  - `clinepass-provider.test.ts` — **21/21**
  - `provider-alias-uniqueness.test.ts` — **6/6**
  - `eslint-suppressions.json` — 2 stale entries pruned; ESLint gate clean

PR is mergeable again (was CONFLICTING against the moved base). Thanks again for the thorough trace — this one would have kept the base red after merge otherwise.

diegosouzapw:
Thanks for jumping on the base-red so quickly. One course correction: we decided on #11176 to COMPLETE the hackclub removal rather than drop the three assertions from provider-alias-uniqueness.test.ts — the assertions are the canary that surfaced the registry/catalog split, and the removal was requested by the Hack Club maintainers. Could you update this PR to: (1) remove hackclub from src/shared/constants/providers/apikey/gateways.ts (entry + alias hc) and the leftover comments in web-cookie.ts:265 and open-sse/config/providers/registry/huggingchat/index.ts:5; (2) keep the uniqueness test assertions intact (they should pass once the catalog entry is gone); (3) run the 348→347 count cascade (npm run gen:provider-reference + README/AGENTS/llm.txt mirrors); (4) keep the cli-combo-command fix as-is (that's correct and independently needed for #11203). That makes this PR green the base AND complete the removal in one sweep.

PR #11202, opened 2026-08-23
fix(routing): forward deps.settings.lkgpEnabled into RoutingContext (#11181)

### LINKED ISSUE #11181: [BUG]   Settings toggle lkgpEnabled never reaches RoutingContext, LKGP cannot be disabled

yaman0786:
## Summary
- Settings → Routing LKGP toggle persisted but never reached routing code
- `resolveAutoStrategyOrder` now forwards host `deps.settings.lkgpEnabled` into `RoutingContext` (same snapshot chatCore already loaded — no extra `getSettings()`)
- Explicit-combo `applyStrategyOrdering` left untouched (no settings dep today; different scope from #11181's auto/* path)
- Closes #11181

## Test plan
- [x] `tests/unit/lkgp-enabled-plumbing-11181.test.ts`
  - control: `settings=null` keeps pin
  - `lkgpEnabled:false` → rules picks cheap over pinned pricey; log has no `LKGP: using last known good`
- [x] `combo-resolve-auto-strategy-split` + `combo-apply-strategy-ordering-split` still pass
- [ ] CI green on this branch

## Reviewer notes
- Net prod change: **+8 lines** in one file (`resolveAutoStrategy.ts`)
- Guard stays `=== false` / optional field so absent/null preserves prior pin behavior
- Rival approach note: same construction-site fix as #11193; this PR also ships changelog fragment + construction-site test asserting log line + ordered target

⚠️ base-red inherited: #9985


pacocartones:
@yaman0786 heads up, I think we have independently written the same fix and one of us is going to waste effort. Flagging early rather than letting both sit in the queue.

Your #11202 and my #11193 both address #11181, both patch `open-sse/services/combo/resolveAutoStrategy.ts`, and both forward `lkgpEnabled` into the `RoutingContext` from the settings snapshot that is already loaded. Same file, same approach, same conclusion about the root cause.

For the record on timing, not as a claim of ownership: #11193 was opened at `2026-08-23T01:18:55Z` and #11202 at `2026-08-23T02:56:12Z`. Neither of us was assigned the issue, so this is just two people picking up the same unclaimed bug within a couple of hours. No bad faith anywhere.

I am not going to argue for mine over yours. **@diegosouzapw, whichever is more useful, please take it and close the other** — I genuinely do not mind which, and I would rather one merged fix than two competing ones.

One difference worth noting so the decision is informed. The changelog entries differ in a way that matters for traceability:

- #11193: `changelog.d/fixes/11181-lkgp-enabled-context.md` (named for the **issue**)
- #11202: `changelog.d/fixes/11202-lkgp-enabled-context.md` (named for the **PR**)

Looking at the existing entries in `changelog.d/fixes/`, the house convention appears to key on the issue number, so whichever PR is chosen may want the `11181-` name.

@yaman0786 if yours is picked, I am happy to close mine immediately. Just say the word.

Separately, and unrelated to which of these wins: both are currently red, and neither is our fault. `release/v3.8.50` is base-red for everyone right now (9 of 9 recent PRs by other authors fail the same checks). #11201 fixes part of it, and I have filed #11203 for a second breakage that #11201 does not yet cover.

diegosouzapw:
Closing as superseded — this exact fix (forwarding lkgpEnabled into the RoutingContext literal in resolveAutoStrategyOrder) shipped today via #11193, including the same deliberate scope boundary (applyStrategyOrdering left untouched). Thank you @yaman0786 — your writeup confirmed the diagnosis independently!

PR #11228, opened 2026-08-23
Beginner UX: guided endpoint connection header

ignamiranda:
Pitch #08 (Endpoint guided connection, Wave 3) — stable header with endpoint URL (/v1), test action, advanced protocols demoted. Branch: contributor/feat/beginner-ux-endpoint. Targets release/v3.8.50, inherits base-red #9985. 2 files, 16 insertions. Static test included.

PR #11205, opened 2026-08-23
feat(codex): self-contained codex app-server transport (executor + provider + sign-in)

arminanton:
## Summary

Adds a **codex `app-server` transport** so OmniRoute can drive a local `codex app-server` (the Codex CLI's own JSON-RPC-over-WebSocket server) and bridge it to the OpenAI Responses SSE surface. This lets a codex connection self-manage its ChatGPT/OpenAI OAuth (its own `~/.codex/auth.json`) instead of OmniRoute replaying a stored token — the same model an interactive `codex` session uses.

The transport is a **new, opt-in provider** (`codex-app-server`) plus an app-server branch inside the existing `CodexExecutor`; the default codex path is unchanged.

## What's in it (feature)

**New executor / transport** (`open-sse/executors/`)
- `codex-app-server.ts` (448L) — `CodexAppServerExecutor`: opens the WS, runs `initialize → thread/start → turn/start`, drains `ServerNotification`s, and bridges them to `AdapterEvent`s that `bridgeToResponsesSSE` already understands. Awaits the terminal `turn/completed`/`error` notification **before** closing the socket (fixes an early-close hang where the turn resolved on accept, not on completion).
- `codex/appServerClient.ts` (289L) — id-correlated JSON-RPC-over-WS client (pending-map, `Authorization: Bearer <capability-token>` handshake, auto-approve of inbound approval requests so the model's agentic tool-calls proceed; the harness that consumes OmniRoute remains the real execution gate).
- `codex/appServerConfig.ts` (94L) — resolves url/token/cwd/approvalPolicy/sandbox from `providerSpecificData`/env (`OMNIROUTE_CODEX_APPSERVER_*`).
- `codex/appServerEvents.ts` (208L) — notification → `AdapterEvent` mapper (text/reasoning deltas, usage, terminal, and dynamic tool-call passthrough).
- `codex/appServerAuthProbe.ts` (102L) — probes `account/read` over the WS so the health check verifies the CLI is actually **signed in**, not merely reachable.

**Provider + sign-in + health**
- `open-sse/config/providers/registry/codex-app-server/index.ts` — first-class `codex-app-server` provider (no-auth; OmniRoute stores no token).
- Sign-in flow: guarded `auth.json` write + "Apply auth" UI.
- Connection-test health check is app-server-aware (`src/app/api/providers/[id]/test/codexAppServerHealth.ts`): a `readyz` liveness probe + the signed-in probe, so a logged-out CLI reports actionable status instead of a false 401.
[...]

PR #11256, opened 2026-08-23
fix(dashboard): restore ProviderModelsModal map broken by #11228 (base-red #9985)

diegosouzapw:
## What

The squash merge of #11228 (8a42aeebb8, 2026-08-23 09:47) applied its endpoint-header hunk **inside `ProviderModelsModal`**, replacing the `groupModels.map` callback's `return (` with the page-level guided-header JSX. The file stopped parsing:

- Turbopack: `3 errors at EndpointPageClient.tsx:2397` — every Build App / Publish to Docker Hub run red since (runs 32641477569, 32641695716, 32642031466)
- release-green reported the same defect as "1 ESLint error" (run 32632606002) — the parser error, one defect, two symptoms

## Fix

Surgical revert of that single hunk — the file is now **byte-identical to its pre-#11228 state** (verified: `diff` vs 968fa961 is empty). Nothing else from #11228 is touched: the health-page verdict header, the resilience reassurance block and the new i18n keys all parse fine and stay.

## Validation (TDD, HR#18)

- **RED**: `prettier --check` on the base-tip file → `SyntaxError: Unexpected token (2397:11)`; the pre-existing jsdom render suite `EndpointPageClient.test.tsx` imports the component, so it reds on the broken base too.
- **GREEN**: after the revert — 4/4 passing, prettier clean, eslint clean.

## Follow-up (not in this PR)

The guided endpoint connection header that #11228 intended never actually existed (the injected JSX was unreachable and uncompilable). It needs a fresh PR applying the header to the page-level component (`EndpointPageClient` main return), not the modal. Will note on #11228.

Refs #9985

### DIFF
--- a/src/app/(dashboard)/dashboard/endpoint/EndpointPageClient.tsx
+++ b/src/app/(dashboard)/dashboard/endpoint/EndpointPageClient.tsx
@@ -2360,19 +2360,7 @@ function ProviderModelsModal({
         <div className="flex flex-col gap-1">
           {groupModels.map((m) => {
             const copyKey = `modal-${m.id}`;
-  return (
-    <div className="flex flex-col gap-4">
-      <div className="flex flex-col gap-2 mb-2">
-        <h1 className="text-2xl font-bold">{t("endpoint.title")}</h1>
-        <p className="text-text-muted">{t("endpoint.subtitle")}</p>
-        <div className="flex items-center gap-3 mt-2">
-          <code className="text-sm bg-card-subtle px-3 py-1 rounded-md text-text-main font-mono">{useDisplayBaseUrl()}/v1</code>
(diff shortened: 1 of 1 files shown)


diegosouzapw:
**Merging --admin with red discrimination (merge-gates §4).** Fast Production Build — the check this PR exists to fix — is **green**. The failing checks (Fast Quality Gates, Unit Tests shards 1/3/4, ESLint warnings) are all inherited base-red, never PR-introduced: this diff is byte-identical to the pre-#11228 file, so it cannot have introduced anything. Each failing test maps to a merge from this morning that landed without the full suite: check-deps allowlist (#11228's new @testing-library deps), search-route 400 contract (#11097), codex/vscode/Qwen CLI_TOOLS (#11205), i18n pt-BR drift (#11215/#11228/#11204/#11195). A dedicated drain PR for that cluster is already in progress; leaving the base unbuildable any longer would keep poisoning every open PR's CI.

PR #11271, opened 2026-08-23
fix(ollama): route models by advertised capability (#11087) — port of #11088 to the release line

### LINKED ISSUE #11087: feat(providers): Ollama local capabilities are flattened to chat and specialty routes reject valid models

yourspraveen:
⚠️ base-red inherited: #9985

## What

Ports **#11088** (merged) onto the active release line. Closes the gap you flagged on #11087:

> it landed on `main`, not on the `release/v3.8.50` branch, so it will NOT be in the v3.8.50 package about to publish; it rides the following release line (v3.8.51+)

Same precedent as #11165, which ported #11075 from `main` onto `release/v3.8.50`.

## Why this is worth porting rather than waiting

`release/v3.8.51` does not exist yet, and every PR currently open — 30 of them — is cut from `release/v3.8.50`. Until the next cycle branch appears, the fix is reachable only from `main`, which no active branch is based on. Meanwhile the defect is live on the release line: every Ollama Local model is flattened to `chat`, so `/v1/embeddings` and `/v1/images/generations` reject models the daemon advertises as capable.

If you would rather this wait for the v3.8.51 branch, say so and I will close it — the branch will keep.

## ⚠️ No longer a byte-identical port — #11088 carries two defects

This PR **started** as a byte-identical port and I described it that way below. CI proved that wrong, so the description is corrected here rather than quietly left standing. Two failures showed up that are **not** in the base-red set, reproduce on `main` (which already carries #11088), and do **not** reproduce on the pristine release tip — i.e. they are defects in the merged change, not in the port:

**1. OpenAI import leaks image/video models into chat selections** — real behavioural regression.

```
✖ OpenAI import excludes image and video generation models from chat selections
  actual:   [ 'gpt-5.6-sol', 'gpt-image-2', 'sora-2-pro', 'vendor-image-model' ]
  expected: [ 'gpt-5.6-sol' ]
```

#11088 dropped `filterChatSelectableModels` from `importManagedModels` for **every** provider, but the compensating read-time filter (`resolveLocalSyncedEndpointRoute`) is gated on `isSelfHostedChatProvider`. Non-local providers therefore lost the filter with nothing replacing it, and the models are persisted, not merely returned. Fixed in `c018e3a45` by scoping the drop to self-hosted providers — the local behaviour #11087 asked for is untouched and the Ollama suite stays 3/3.

**2. An unregistered credential site** — `hard-lease credential, executor, and connection-query inventory has no unclassified site` (`src/lib/embeddings/service.ts` 2 → 3). The new site resolves through `getProviderCredentials` with the connection allowlist from `resolveLocalSyncedEndpointRoute` and handles `allRateLimited`, so it is fenced like its two siblings — the ledger had simply not been told. Registered with a comment saying why.

**`main` still has both.** This PR fixes them only on the release line; `main` needs the same two-line change. Flagging rather than opening a second PR unasked — say the word and it is up in minutes.

## Fidelity (as originally written — see the correction above)

The change set is **byte-identical to the merged #11088** — I diffed all 11 files against `refs/pull/11088/head` after applying, and every one matches:

```
open-sse/services/combo/autoStrategy.ts                     |   6 +-
src/app/api/providers/[id]/models/discovery/helpers.ts      |  38 ++++
src/app/api/providers/[id]/models/route.ts                  |   3 +
src/app/api/v1/images/generations/route.ts                  |  19 +-
src/lib/embeddings/service.ts                               |  59 +++++-
src/lib/providerModels/managedModelImport.ts                |   6 +-
src/lib/providerModels/modelDiscovery.ts                    |   9 +-
src/lib/providerModels/ollamaCapabilities.ts                |  98 ++++++++++
src/lib/providerModels/syncedEndpointRouting.ts             |  31 ++++
tests/unit/ollama-local-capabilities-routing.test.ts        | 205 +++++++++++++
```

The **ported commit** re-derived nothing: applied with `git apply --3way` at the release tip `a7e09eda5`, clean, no conflicts. The two-file fix above sits in a **separate commit on top** (`c018e3a45`), so the port and the correction stay reviewable independently.

**Deliberately not carried over:** `docs/superpowers/plans/2026-08-23-qdrant-configuration-guidance.md` and its `specs/` sibling. Those are on `main` only because they rode in with #11088's squash — they came from #11213 and `855243ab1` already removed them from the release branch. They do not belong here (or on `main` — see the separate cleanup PR).

The changelog fragment is renumbered to this PR in a follow-up commit, matching how #11165 handled its own fragment.

## Verification at the release tip

| Check | Result |
| --- | --- |
| `tests/unit/ollama-local-capabilities-routing.test.ts` | **3/3 pass** |
| `npx eslint` on all 10 changed source/test files | clean¹ |

¹ One `no-restricted-imports` hit in `src/app/api/providers/[id]/models/route.ts` — the pre-existing `@/lib/localDb` import at line 14, already frozen in `config/quality/eslint-suppressions.json` with `count: 1`. Untouched by this port; `npm run lint` (which applies the suppressions, and is what CI runs) is clean.

Fixes #11087 on the release line.

### DIFF
--- a/changelog.d/fixes/11271-ollama-capability-routing.md
+++ b/changelog.d/fixes/11271-ollama-capability-routing.md
@@ -1,0 +1,1 @@
+- **fix(ollama):** Ollama Local models are no longer flattened to `chat` at sync time — the synced store persists every advertised capability and chat filtering moves to read time, so `/v1/embeddings` and `/v1/images/generations` stop rejecting models the daemon reports as capable ([#11271](https://github.com/diegosouzapw/OmniRoute/pull/11271)) — thanks @yourspraveen
--- a/open-sse/services/combo/autoStrategy.ts
+++ b/open-sse/services/combo/autoStrategy.ts
@@ -28,6 +28,7 @@ import type {
   ResolvedComboTarget,
 } from "./types.ts";
 import { extractSessionAffinityKey } from "@/sse/services/auth";
+import { filterChatSelectableModels } from "../modelEndpointPolicy.ts";
 import { DEFAULT_INTENT_CONFIG, type IntentClassifierConfig } from "../intentClassifier.ts";
 import { getTaskFitness } from "../autoCombo/taskFitness.ts";
 import {
@@ -470,10 +471,13 @@ export async function expandAutoComboCandidatePool(
       // catalog only when the user has none. This keeps catalog-only models
       // (e.g. openrouter/auto) out of pure-auto pools when the operator only
       // synced a subset (e.g. OpenRouter with importFreeModelsOnly).
-      const [syncedModels, customModels] = await Promise.all([
+      // #11088 (option 1): the synced store now persists non-chat models too —
+      // chat combo pools must keep filtering them out at read time.
+      const [syncedModelsRaw, customModels] = await Promise.all([
         getSyncedAvailableModels(providerId),
         getCustomModels(providerId),
       ]);
+      const syncedModels = filterChatSelectableModels(providerId, syncedModelsRaw);
       const hiddenModels = hiddenModelsMap.get(providerId);
       const userVisibleIds = new Set<string>();
       for (const m of syncedModels) if (m.id && !hiddenModels?.has(m.id)) userVisibleIds.add(m.id);
--- a/src/app/api/providers/[id]/models/discovery/helpers.ts
+++ b/src/app/api/providers/[id]/models/discovery/helpers.ts
@@ -1,5 +1,11 @@
 import { isSelfHostedChatProvider } from "@/shared/constants/providers";
 import { getStaticModelsForProvider, type LocalCatalogModel } from "@/lib/providers/staticModels";
+import { SAFE_OUTBOUND_FETCH_PRESETS, safeOutboundFetch } from "@/shared/network/safeOutboundFetch";
+import { getProviderValidationGuard } from "@/shared/network/outboundUrlGuardPolicy";
+import {
+  buildOllamaShowUrl,
+  enrichOllamaModelsWithCapabilities,
+} from "@/lib/providerModels/ollamaCapabilities";
 
 export type JsonRecord = Record<string, unknown>;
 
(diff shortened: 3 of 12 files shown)


yourspraveen:
## CI triage — 17 inherited, 1 flake, and 2 that were genuinely mine

The run came back red on all four shards. Breaking it down honestly, because two of these were **not** base-red:

**Inherited (17 assertions).** Identical to the set reproduced on the pristine tip and reported at #9985 — CLI_TOOLS ×5, Codex `872000 !== 272000` ×6, chatCore combo skip, search-route, i18n pt-BR ×2, settings-i18n-keys, check-deps. Also hits the maintainer's own #11262 unchanged.

**Mine (2 assertions).** `managed-model-import` and `hard-session-lease-bypass-inventory`. I verified these three ways before concluding anything:

| | pristine tip `a7e09eda5` | `main` (carries #11088) | this branch |
| --- | --- | --- | --- |
| `managed-model-import` | 9/9 | **8/9** | **8/9** → now 9/9 |
| `hard-session-lease-bypass-inventory` | 3/3 | **2/3** | **2/3** → now 3/3 |

They fail on `main` too, which means the port carried them faithfully — they are defects in the merged #11088, not port damage. Details and the `main`-side reproduction are in **#11276**; the corrected description is in the PR body above, and the fix is `c018e3a45`, a separate commit so the port and the correction review independently.

I have to correct something I wrote in the original body: **"byte-identical to the merged #11088 … nothing was re-derived"** was true of the port commit and is no longer true of the PR. Byte-identical turned out to be the wrong goal once the source change had a bug in it.

**One flake, deliberately excluded:** `ttft() measures first-forwarded-chunk latency` failed on one #11146 run and nowhere else — 8/8 on the pristine tip, 3× 8/8 on a branch. Not counted against either PR.

### Where this leaves the PR

The two real defects are fixed here and the Ollama capability suite is still 3/3, so #11087's goal is intact. The remaining red is the base, which #9985 tracks and which no PR against this branch can currently escape.

`main` still carries both defects — #11271 cannot reach it from this base. Say the word and the `main`-side PR is up in minutes.

PR #11274, opened 2026-08-23
fix(providers): clamp reasoning_effort to the declared vocabulary of models that expose one

linhdmn:
## Problem

Hermes telegram bot → `POST /v1/chat/completions` with model `opencode-go/ox-alpha-free` failed on every turn:

```
HTTP 400: [400]: Error from provider (Console Go): Upstream request failed: [1210]
This model always engages in thinking and cannot be disabled; please use low, high, or max
```

## Root cause (live-probed against the running gateway, 2026-08-23)

Console Go (`opencode.ai/zen/go/v1`) accepts **only `low | high | max`** for `ox-alpha-free` — and rejects every other value whenever the request carries tools:

| `reasoning_effort` sent | with tools | without tools |
| --- | --- | --- |
| `none` / `minimal` | ❌ 400 [1210] | ✅ / ❌ |
| `medium` ← Hermes default | ❌ 400 [1210] | ✅ |
| `xhigh` | ❌ 400 [1210] | ❌ 400 |
| `low` / `high` / `max` | ✅ 200 | ✅ 200 |

Hermes sends `reasoning_effort:"medium"` + 24 tools; OmniRoute forwarded it verbatim. Two gaps made this unreachable for the existing defenses:

1. `ox-alpha-free` is a discovery-synced model with **no static registry entry**, so no sanitizer knew its effort contract.
2. `sanitizeReasoningEffortForProvider` consulted declared `supportedThinkingEfforts` only inside the `max` branch. The learned-caps path can't cover this either: it only clamps down from xhigh/max, and the error text ("please use low, high, or max") isn't a parseable enum.

## Fix

- **Registry**: declare `ox-alpha-free` on `opencode-go` with `supportedThinkingEfforts: ["low", "high", "max"]`.
- **Generic explicit-capability clamp** in `sanitizeReasoningEffortForProvider`: when the exact provider+model entry declares its vocabulary and the request falls outside it, remap to the nearest declared tier (smallest ranked ≥ requested, else the highest declared):
  - `none`/`minimal` → `low`
  - `medium` → `high` (same direction as the existing native DeepSeek mapping)
  - `xhigh` → `max`
- Models **without** a declaration keep #8057's trust-the-upstream pass-through byte-for-byte; all provider-specific rewrites (command-code / ollama-cloud / deepseek) run before the clamp and are untouched.

⚠️ base-red inherited: #9282 (main branch not green)

## Test plan

- New TDD regression guard: `tests/unit/opencode-go-console-go-effort-clamp.test.ts` (7 tests: registry declaration, medium→high as the exact Hermes case, disable-shaped → low, xhigh→max, in-vocabulary pass-through, multi-carrier write-back, undeclared-model pass-through). Red before the fix (5 failing), green after.
- `tests/unit/base-executor-sanitize-effort.test.ts`: 56/56 pass.
- Related suites (opencode-go aliases, shared zen/go models, learned capability, muse-spark, fable5 thinking-disabled): 30/30 pass.
- `typecheck:core` + eslint on changed files clean; `check:provider-consistency` OK.
- Note: `tests/unit/deepseek-thinking-efforts.test.ts` has 5 failures that reproduce identically on pristine `main` without this patch (pre-existing, DB-dependent catalog tests).

### DIFF
--- a/open-sse/config/providers/registry/opencode/go/index.ts
+++ b/open-sse/config/providers/registry/opencode/go/index.ts
@@ -219,5 +219,18 @@ export const opencode_goProvider: RegistryEntry = {
       supportedThinkingEfforts: ["none", "low", "high", "max"],
       targetFormat: "openai-responses",
     },
+    // Console Go free GLM-tier model (live-verified 2026-08-23): the upstream
+    // rejects every reasoning_effort outside {low, high, max} whenever tools
+    // are present — "[1210] This model always engages in thinking and cannot
+    // be disabled; please use low, high, or max" — which broke clients that
+    // default to reasoning_effort:"medium" (Hermes). Declaring the exact
+    // vocabulary lets sanitizeReasoningEffortForProvider clamp off-vocabulary
+    // requests to the nearest accepted tier instead of burning a 400.
+    {
+      id: "ox-alpha-free",
+      name: "ox-alpha (free)",
+      supportsReasoning: true,
+      supportedThinkingEfforts: ["low", "high", "max"],
+    },
   ],
 };
--- a/open-sse/executors/base/reasoningEffort.ts
+++ b/open-sse/executors/base/reasoningEffort.ts
@@ -357,6 +357,43 @@ export function sanitizeReasoningEffortForProvider(
     }
   }
 
+  // ── explicit per-model capability clamp ──────────────────────────────────
+  // When the registry declares supportedThinkingEfforts for this exact model
+  // and the requested effort falls outside that vocabulary, remap to the
+  // nearest declared tier: the smallest ranked value ≥ the request, else the
+  // highest declared (a request above the ceiling lands on the ceiling).
+  // Live case: opencode-go/ox-alpha-free (Console Go) only accepts
+  // {low, high, max} — a client's reasoning_effort:"medium" reached the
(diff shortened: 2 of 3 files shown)


maxmad64bis:
Thanks for the live-probed matrix — the medium + tools → 400 [1210] failure is real. But this PR conflicts with two PRs merged today, ~30 min before it opened:

1. Premise is false. #11232 already parses "please use low, high, or max" and learns the accepted set; its motivating case is this one.
2. Direction conflict (the real blocker). #11232 clamps downgrade-only: medium → low. This PR maps medium → high (upgrade). Same Hermes request, opposite result — both can't ship.
3. Static declaration is regressive. #11252 makes the learned set the single source of truth precisely so discovery-synced models like ox-alpha-free need no static entry. Adding supportedThinkingEfforts re-adds the contract it removed.
4. **The client isn't interrupted anyway. base.ts's 4xx clamp-and-retry intercepts the 400, learns the set, retries internally as medium → low, and the client gets a 200.** Only cost: one internal 400→retry on the first request per process.

Proposal: close this in favor of (a) relying on #11232/#11252 as-is, or (b) seeding ox-alpha-free's accepted set as a non-authoritative hint consumed by the existing downgrade-only clamp — not forking a second, direction-divergent one.

Issue #9435, opened 2026-08-04
fix(backend): Kiro import token endpoint overwrites existing connection instead of creating new one


jtnqr:
When importing multiple Kiro accounts via the import token API endpoint (`POST /api/oauth/kiro/import`), the API **overwrites** the existing connection instead of creating a new one.

**Expected behavior:** Each import creates a NEW connection (consistent with OAuth flow)  
**Actual behavior:** All imports return the SAME connection ID, updating the existing one

This contradicts the multi-account support added in v3.8.0+ (Issue #2328).

---

### Method 1: Via Dashboard (Easiest to reproduce)

1. Start OmniRoute:
   ```bash
   # Via npm (tested)
   npm install -g omniroute
   omniroute
   
   # Or via Docker (for reproducibility)
   docker run -d --name omniroute -p 20129:20128 \
     -v omniroute-data:/app/data \
     diegosouzapw/omniroute:latest
   ```

2. Open dashboard: http://localhost:20129

3. Go to **Providers** → Click **Add Provider** → Select **Kiro**

4. Click **Import Token** tab

5. Paste first refresh token (format: `aorAAAAAG...`), enter a name (e.g., "Account 1"), click **Import**
   - ✅ Success message appears
   - Note the connection appears in provider list

6. Repeat step 3-5 with a **different** refresh token (e.g., "Account 2")
   - ✅ Success message appears again
   - ❌ But check the provider list - still only **1 Kiro connection**
   - The second import **replaced** the first one

7. Verify in database:
   ```bash
   # npm install
   sqlite3 ~/.omniroute/data/storage.sqlite \
     "SELECT COUNT(*) FROM provider_connections WHERE provider = 'kiro';"
   
   # Docker
   docker exec omniroute sqlite3 /app/data/storage.sqlite \
     "SELECT COUNT(*) FROM provider_connections WHERE provider = 'kiro';"
   ```

**Expected:** 2 connections  
**Actual:** 1 connection

### Method 2: Via API (for programmatic verification)

1. Import first token:
   ```bash
   curl -X POST http://localhost:20129/api/oauth/kiro/import \
     -H "Content-Type: application/json" \
     -d '{
       "refreshToken": "aorAAAAAG...",
       "region": "us-east-1",
       "name": "Account 1"
     }'
   ```
   Note the returned `connection.id`

2. Import second token (different token):
   ```bash
   curl -X POST http://localhost:20129/api/oauth/kiro/import \
[...]

Issue #2328, opened 2026-05-17
[Feature] Kiro multi-account support: independent OAuth sessions per connection


disonjer:
### Feature Description

When using multiple Kiro accounts (e.g., one Google-authenticated and one GitHub-authenticated Pro account), OmniRoute cannot keep both alive simultaneously. The refresh token of one account gets invalidated when the user logs into the other account via `kiro-cli`, because Kiro backend enforces a single-session policy per account.

### Problem

1. User imports refresh_token for Account A (Google Pro) into OmniRoute via "Import Token"
2. User needs to log into Account B (GitHub Pro) via `kiro-cli` to get its refresh_token
3. After logging into Account B, Kiro backend invalidates Account A's refresh_token
4. OmniRoute can no longer refresh Account A — connection dies within ~1 hour
5. The workaround (never touching CLI for one account) is fragile and breaks on any re-auth

### Proposed Solution

**Register an independent OIDC client per connection during import.**

The code in `src/lib/oauth/services/kiro.ts` already has `registerClient()` which creates a unique `clientId`/`clientSecret` pair via AWS SSO OIDC. Currently this is only used for the AWS Builder ID device flow.

Proposal:
1. When a user imports a refresh_token (social auth), OmniRoute calls `registerClient()` to obtain its own `clientId`/`clientSecret` pair
2. Store this pair in `providerSpecificData` for the connection
3. Use this dedicated client registration for all subsequent token refreshes
4. Since each connection has its own registered client, refreshing one account does not invalidate another
5. CLI logins no longer conflict with OmniRoute sessions because they use different client registrations

This would make OmniRoute's refresh cycle fully independent from kiro-cli sessions.

### Alternative Approaches

- **Stable social OAuth device flow in dashboard** — allow users to authenticate directly through OmniRoute without CLI (currently hidden per #2112)
- **Per-connection isolation via separate KIRO_HOME** — hacky, requires multiple CLI profiles
- **Cron-based token rotation** — periodically re-import from CLI, unreliable

### Context

- Related: #2112 (social login buttons hidden), #2114 (AWS Builder ID scope limitation)
- Current refresh logic: `src/lib/oauth/services/kiro.ts` lines 184-239
- Token health check correctly saves new refresh_token after refresh (line 417-418 in `tokenHealthCheck.ts`)
- The issue is upstream (Kiro backend single-session), but OmniRoute can work around it with independent client registration

### Environment

- OmniRoute 3.8.0
[...]

PR #11287, opened 2026-08-23
fix(oauth): stop overwriting Kiro connections that share a profile ARN (#10815)

### LINKED ISSUE #10815: [BUG] Kiro AI add a 2º connection
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Debian 13

### Node.js Version

_No response_

### Provider(s) Involved

Kiro

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

Cannot add a second connection of Kiro, 

### Steps to Reproduce

Got Providers,
Kiro AI
Add
and after the authentication, don't appears nothing only the first

### Expected Behavior

To have multiple connections

### Actual Behavior

Only appears one connection

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

engenhariaandrereis01-ai:
## Summary

Adding a second Kiro account through Google/GitHub social login silently **replaced** the
connection already stored, so the provider list never showed more than one Kiro account.

Root cause: a **profile ARN identifies the CodeWhisperer profile, not the account**. Distinct
Builder ID accounts authenticated through social login share the same ARN, and
`findKiroConnectionByIdentity()` matched on it before any account-level field. The second login
therefore resolved to the first connection and `/api/oauth/kiro/social-exchange` took the
`updateProviderConnection()` branch, overwriting it.

Two things made this invisible until now:

1. **#10918 fixed a different layer.** It hardened the e-mail dedup inside
   `createProviderConnection()`, but the social flow never reaches that function — the route
   resolves the match itself and diverts to `updateProviderConnection()`.
2. **No e-mail is ever stored for social accounts.** `extractEmailFromJWT()` requires a
   three-part JWT, but the Kiro social access token is opaque, so it returns `null`. With the
   e-mail absent, the ARN was the only field left to match on.

The ARN is now trusted only when the incoming identity also carries an account-level identifier
(e-mail or `clientId`) that does not contradict the stored connection. Re-authenticating the same
account still updates it in place; a different account creates a new connection.

## Related Issues

- Closes #10815
- Related to #10918 (same symptom, different layer), #9435, #2328

## Validation

- [x] Change type: provider (Kiro OAuth connection identity)
- [x] Focused tests: `tests/unit/kiro-connection-identity.test.ts` — 8/8 passing, written
      failing-first (the 3 new cases fail on `release/v3.8.50` and pass with this change)
- [ ] `npm run lint` — **not run locally**: `npm ci` fails on this checkout with
      `Missing: brace-expansion@2.1.4 from lock file`, so no `node_modules` could be installed.
      The change is a pure function with no new imports; CI will cover lint/typecheck.
- [x] Branched from and reconciled with `release/v3.8.50` (current upstream default; no
      `release-freeze` open at the time of writing)
- [x] Production-code change ships with new automated tests in this PR

## Tests Added Or Updated

- `tests/unit/kiro-connection-identity.test.ts` — three new cases:
  - shared profile ARN with no account identifier must **not** match (the reported bug)
  - diverging e-mails on a shared ARN are distinct accounts
  - the same account on a shared ARN still matches (guards against duplicating on re-login)

The five pre-existing cases in that file are unchanged and still pass.

## Coverage Notes

Touches `src/lib/oauth/kiroConnectionIdentity.ts` only. Every branch added
(`hasAccountIdentifier`, `contradictsAccount`) is exercised by the three new tests plus the
existing ARN/clientId cases.

## Reviewer Notes

- **Deliberate trade-off:** when no account identifier is available, the function now returns
  `null` (create a new connection) instead of matching on the ARN. A duplicate connection is
  recoverable by deleting it; an overwritten one is not — so the failure mode is the safe one.
  Re-login of an account whose e-mail *is* known still updates in place.
- **Not addressed here:** social accounts still store no e-mail, because
  `extractEmailFromJWT(accessToken)` cannot parse the opaque social token. Worth a separate
  issue — `GetUsageLimits` is not a usable source for it either, since (per the comment in
  `open-sse/services/usage/kiro.ts`) AWS routinely rejects social tokens there with 401/403.
- The reporter on #10815 confirmed after the #10918 merge that v3.8.50 still reproduced the
  overwrite through Google login, GitHub login and import token.

### DIFF
--- a/changelog.d/fixes/10815-kiro-social-multi-account.md
+++ b/changelog.d/fixes/10815-kiro-social-multi-account.md
@@ -1,0 +1,1 @@
+- fix(oauth): stop treating the Kiro profile ARN as an account identity in `findKiroConnectionByIdentity()`, so a second Google/GitHub social login creates a new connection instead of overwriting the first — distinct Builder ID accounts share the same CodeWhisperer profile ARN, and the social token is not a JWT, so no e-mail was available to disambiguate them (#10815)
--- a/src/lib/oauth/kiroConnectionIdentity.ts
+++ b/src/lib/oauth/kiroConnectionIdentity.ts
@@ -30,6 +30,27 @@ function providerData(connection: KiroConnectionLike): Record<string, unknown> {
     : {};
 }
 
+/** True when the identity carries something that identifies the ACCOUNT (not the profile). */
+function hasAccountIdentifier(identity: KiroConnectionIdentity): boolean {
+  return Boolean(folded(identity.email) || trimmed(identity.clientId));
+}
+
+/** True when a shared field is present on both sides and disagrees — different accounts. */
+function contradictsAccount(
+  connection: KiroConnectionLike,
+  identity: KiroConnectionIdentity
+): boolean {
+  const email = folded(identity.email);
+  const existingEmail = folded(connection.email);
+  if (email && existingEmail && email !== existingEmail) return true;
+
+  const clientId = trimmed(identity.clientId);
+  const existingClientId = trimmed(providerData(connection).clientId);
+  if (clientId && existingClientId && clientId !== existingClientId) return true;
+
+  return false;
+}
+
 /** Find an existing Kiro account without comparing OAuth tokens or API keys. */
 export function findKiroConnectionByIdentity(
   connections: KiroConnectionLike[],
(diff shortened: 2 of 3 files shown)


engenhariaandrereis01-ai:
Validated live on a production deployment (v3.8.50 + this patch), not just in unit tests.

**Before the login** the instance had one Kiro `oauth` connection created through GitHub social login, with `profileArn` set and no e-mail — the exact state that triggered the overwrite:

| id | auth_type | provider | profileArn | created_at |
|---|---|---|---|---|
| `49e7cf96…` | oauth | Github | `…profile/EHGA3GRVQMUK` | 20:12:25 |
| `ce5cc0a5…` | apikey | API Key | — | 20:36:22 |

**After authenticating a second account through GitHub:**

| id | auth_type | provider | profileArn | created_at |
|---|---|---|---|---|
| `49e7cf96…` | oauth | Github | `…profile/EHGA3GRVQMUK` | 20:12:25 (untouched) |
| `ce5cc0a5…` | apikey | API Key | — | 20:36:22 |
| **`60eb8e99…`** | **oauth** | **Github** | **`…profile/EHGA3GRVQMUK`** | **21:31:46 (new)** |

Both social connections carry the **same profile ARN** — which is precisely what used to make the
second login resolve to the first connection and overwrite it. They now coexist, and the original
row kept its `created_at`, so nothing was replaced.

Worth noting for reviewers: neither social connection stores an e-mail, because
`extractEmailFromJWT()` cannot parse the opaque Kiro social token. That is the follow-up mentioned
in the PR description — with no account identifier available, re-authenticating an existing account
also creates a new row. That is the deliberate trade-off here: a duplicate can be deleted, an
overwritten account cannot be recovered.

PR #11313, opened 2026-08-24
fix(usage): honor active rateLimitedUntil cooldown on provider-limits sync

### LINKED ISSUE #11277: fix(quota): active rateLimitedUntil cooldown is cleared by the provider-limits sync when lastErrorType is not "quota_exhausted"
## Summary

`maybeClearRecoveredQuotaState()` can wipe a still-active `rateLimitedUntil` cooldown, because the guard that protects it is gated on `lastErrorType === "quota_exhausted"`. Every other `lastErrorType` value falls straight through to the clear, no matter how far in the future the persisted cooldown is.

In production this produced a self-restarting loop: a 146-hour cooldown was cleared every few minutes, the connection went back to priority position 1 of a `priority` combo, burned a real upstream call, took the same 429, was cooled down again, and the next sync cleared it again.

Verified present on the current default branch `release/v3.8.50` (`527da65`).

## Where

`src/lib/usage/providerLimits.ts:462`

```ts
export async function maybeClearRecoveredQuotaState(
  connection: ProviderConnectionLike,
  usage: JsonRecord
): Promise<ProviderConnectionLike> {
  if (!hasUsableQuota(usage)) return connection;
  if (isTerminalStatusForQuotaRecovery(connection.testStatus)) return connection;
  if (connection.lastErrorType === "quota_exhausted") {
    // ... Claude extra-usage guard
    // ... windowStillExhaustedAfterRealReset() per-window guard
    // ... synthetic-cooldown fallback guard
  }

  const hasTransientState = /* ... */;
  if (!hasTransientState) return connection;

  // clearRecoveredProviderState -> testStatus "active",
  // rateLimitedUntil null, errorCode null, backoffLevel 0
```

The inner guards added for the Claude extra-usage case and for `windowStillExhaustedAfterRealReset()` are all **inside** the `lastErrorType === "quota_exhausted"` branch. If that outer condition is false, none of them run, and `rateLimitedUntil` is not consulted at all before the state is cleared.

## Why the outer condition is false more often than it looks

`lastErrorType` is not normalized to the `RateLimitReason` vocabulary across writers.

- `src/sse/services/auth.ts:3106` persists `lastErrorType: providerErrorType`, a provider-derived string, alongside the real `rateLimitedUntil` parsed from the upstream reset.
- `src/app/api/providers/[id]/test/route.ts:1090` persists `lastErrorType: diagnosis.type`. The possible values there are `unsupported`, `token_refresh_failed`, `account_deactivated`, `upstream_rate_limited`, `upstream_unavailable`, `token_expired`, `upstream_auth_error`, `network_error`, `upstream_error`, `runtime_error`, `reauth_required`, `auth_missing`. None of them is `quota_exhausted`, and the same handler sets a `rateLimitedUntil` 30s in the future for non-terminal failures.

So a connection can carry an entirely legitimate future cooldown together with a `lastErrorType` that this guard does not recognise.

## Why `hasUsableQuota()` makes it reachable

`src/lib/usage/providerLimits.ts:425` returns `true` as soon as **any** single quota window reports `remaining > 0`:

```ts
for (const value of Object.values(quotas)) {
  if (value.unlimited === true) return true;
  const remaining = /* remaining ?? remainingPercentage */;
  if (remaining !== null && remaining > 0) return true;
}
return false;
```

Providers that expose several windows (e.g. a 5-hour window next to a weekly/monthly one) therefore look "recovered" while the window that actually governs the block is at zero. Combined with the `lastErrorType` gap above, the poll clears a cooldown that the 429 handler had deliberately set for days.

## Observed behaviour

Provider: Z.AI coding plan, weekly quota exhausted, upstream reset ~146h out.

Right after the 429:

```
test_status        = unavailable
rate_limited_until = 2026-08-29T19:01:21.000Z
error_code         = 429
```

A few minutes later, with no change to the plan or the upstream quota:

```
test_status        = active
rate_limited_until = NULL
error_code         = NULL
```

Log pattern over one hour: 193 requests were correctly skipped with `all 1 active accounts rate limited`, but every 5 to 10 minutes a real upstream call went out, each one preceded by `Trying model 1/6: zai/glm-5.3`. Each of those returned the same 429 and re-armed the 146h cooldown. The clears line up with the provider-limits sync scheduler started in `src/instrumentation-node.ts:347`.

The upstream error text was `[1310][Weekly/Monthly Limit Exhausted. Your limit will reset at ...]`.

## Impact

- Wasted upstream calls against a provider that is known to be exhausted, on a recurring timer until the real reset.
- In `priority` combos the connection keeps returning to its configured position, so the first target of the chain is repeatedly a known-dead one and every affected request pays the extra round trip and latency.
- Dashboard shows the connection as `active` while it is not.
- Any operator-visible cooldown becomes unreliable, since it may silently disappear between two polls.

## Suggested fix

A `rateLimitedUntil` in the future is a hard statement made by the 429 handler. No quota poll should be able to overrule it, and the decision does not need `lastErrorType` at all.

```ts
export function hasActiveCooldown(
  connection: Pick<ProviderConnectionLike, "rateLimitedUntil">,
  now: number = Date.now()
): boolean {
  if (!connection.rateLimitedUntil) return false;
  const until = new Date(String(connection.rateLimitedUntil)).getTime();
  return Number.isFinite(until) && until > now;
}
```

and in `maybeClearRecoveredQuotaState()`, before the clear:

```ts
if (hasActiveCooldown(connection)) return connection;
```

Deliberately fail-open on an unparseable timestamp, so a malformed value cannot pin a connection forever.

This keeps the existing `quota_exhausted` refinements useful for the case where the cooldown has already expired but a window is still demonstrably blocking. It only removes the ability to clear a cooldown that has not run out yet.

If you would rather keep the behaviour opt-out, the same check behind a setting would work too, but the current default looks unintended.

## Verification

We are running this patch on top of `v3.8.49` (`c9d4a45f1`) in production. After deployment, the connection was touched exactly once as target 1, then the cooldown persisted:

```
test_status        = unavailable
rate_limited_until = 2026-08-29T19:01:21.001Z
error_code         = 429
```

Since then: 7 clean skips, zero real upstream calls to that provider in `ProxyEgress`, and unrelated requests continued down the chain normally. Before the patch the same window contained a real call every 5 to 10 minutes.

Covered by 8 unit tests: the 146h case with a non-`quota_exhausted` `lastErrorType`, a cooldown one second out, an expired cooldown, the reset instant exactly, missing/empty values, and the unparseable-timestamp fail-open. Plus a test pinning the `hasUsableQuota()` precondition that one healthy window is enough to return `true`.

Happy to open a PR against `release/v3.8.50` if the approach looks right to you.

## Environment

- OmniRoute `3.8.49` (`c9d4a45f1`), Docker, self-hosted
- Code inspection for this report done against `release/v3.8.50` (`527da65`)
- Affected provider: Z.AI, but the mechanism is provider-independent

pacocartones:
Fixes #11277

## Problem

`maybeClearRecoveredQuotaState()` gated every defensive guard behind `lastErrorType === "quota_exhausted"`. Any other `lastErrorType` fell straight through to `clearRecoveredProviderState` and wiped a still-active `rateLimitedUntil` without ever consulting it.

`lastErrorType` is not normalized across writers. The provider test route persists `diagnosis.type` (`upstream_rate_limited`, `network_error`, ...), which never equals `"quota_exhausted"`, while persisting a real `rateLimitedUntil`. Combined with `hasUsableQuota()` returning true on the first non-zero window — so a connection with a 5h window and a weekly window looks recovered while the weekly one is still at zero — a multi-day cooldown was cleared on every 5-10 minute sync, burning a real upstream call, taking the same 429, and cooling down again. That is the loop the reporter observed with a 146-hour cooldown.

## Fix

Adds an exported `hasActiveCooldown()` and hoists the quota-window guards out of the `lastErrorType` branch so they apply whenever a cooldown is still in force, regardless of which writer set it.

`hasActiveCooldown` deliberately fails **open** on an unparseable timestamp: a corrupt value must never be able to pin a connection offline forever.

The change is additive — the `quota_exhausted` refinements keep their exact previous behaviour for the case where the cooldown has already expired.

## Test

Added to the existing `tests/unit/provider-limits-recovery.test.ts`.

Verified failing on the base commit, 2 of 17:

```
✖ active cooldown survives a quota sync even when lastErrorType is not quota_exhausted (#11277)
✖ hasActiveCooldown fails open on unparseable and absent timestamps (#11277)
ℹ pass 15
ℹ fail 2
```

With the fix, 17/17 pass:

```
ℹ tests 17
ℹ pass 17
ℹ fail 0
```

Run with the repository's own unit runner (`node --test` via the `test:unit` script), not vitest — `tests/unit/*.test.ts` is covered by the former.

Branch is up to date with `release/v3.8.50`.

### DIFF
--- a/src/lib/usage/providerLimits.ts
+++ b/src/lib/usage/providerLimits.ts
@@ -459,12 +459,29 @@ function windowStillExhaustedAfterRealReset(value: unknown, nowMs: number): bool
   return resetMs > nowMs;
 }
 
+// An explicitly persisted cooldown is still in force when it parses to a
+// timestamp in the future. Fail OPEN on an unparseable/corrupt value: a bad
+// timestamp must never be able to pin a connection offline forever.
+export function hasActiveCooldown(
+  connection: ProviderConnectionLike,
+  nowMs: number = Date.now()
+): boolean {
+  const raw = connection.rateLimitedUntil;
+  if (!raw) return false;
+  const untilMs = new Date(raw).getTime();
+  if (!Number.isFinite(untilMs)) return false;
+  return untilMs > nowMs;
+}
+
 export async function maybeClearRecoveredQuotaState(
   connection: ProviderConnectionLike,
   usage: JsonRecord
 ): Promise<ProviderConnectionLike> {
   if (!hasUsableQuota(usage)) return connection;
   if (isTerminalStatusForQuotaRecovery(connection.testStatus)) return connection;
+  const nowMs = Date.now();
+  const cooldownActive = hasActiveCooldown(connection, nowMs);
+
   if (connection.lastErrorType === "quota_exhausted") {
     if (
       connection.lastErrorSource === CLAUDE_EXTRA_USAGE_ERROR_SOURCE &&
@@ -479,7 +496,16 @@ export async function maybeClearRecoveredQuotaState(
       // below must not release it just because some quota window looks fresh.
       return connection;
     }
+  }
 
+  // An active cooldown must be honoured regardless of `lastErrorType`.
+  // `lastErrorType` is not normalized across writers: the provider test route
+  // persists diagnosis vocabulary (`upstream_rate_limited`, `network_error`, ...)
+  // that never equals "quota_exhausted", while still persisting a real
+  // `rateLimitedUntil`. Gating these guards on `quota_exhausted` alone let every
+  // other error type fall straight through to the clear, wiping multi-day
+  // cooldowns on each provider-limits sync (issue #11277).
+  if (cooldownActive || connection.lastErrorType === "quota_exhausted") {
     const quotas = usage?.quotas;
     if (isRecord(quotas)) {
       // Honor the REAL per-window resetAt from the freshly fetched quota
@@ -488,13 +514,10 @@ export async function maybeClearRecoveredQuotaState(
       // reset was parseable). Only stay locked if some window that governs
       // this connection's quota is still demonstrably exhausted.
       const anyStillBlocking = Object.values(quotas).some((value) =>
-        windowStillExhaustedAfterRealReset(value, Date.now())
+        windowStillExhaustedAfterRealReset(value, nowMs)
       );
       if (anyStillBlocking) return connection;
-    } else if (
-      connection.rateLimitedUntil &&
-      new Date(connection.rateLimitedUntil).getTime() > Date.now()
-    ) {
+    } else if (cooldownActive) {
       // No quota object at all (degraded/failed fetch shape) — fall back to
       // the previous synthetic-cooldown guard.
       return connection;
--- a/tests/unit/provider-limits-recovery.test.ts
+++ b/tests/unit/provider-limits-recovery.test.ts
@@ -530,3 +530,110 @@ test("Claude extra-usage block stays locked through the real sync chain when rec
     "extra_usage marker must survive the general recovery-clearing logic"
   );
 });
+
+// --- issue #11277: active cooldown wiped on every provider-limits sync ---------
+// `lastErrorType` is not normalized across writers. src/app/api/providers/[id]/test/route.ts
+// persists `diagnosis.type` ("upstream_rate_limited", "network_error", ...), which never
+// equals "quota_exhausted", while still persisting a real `rateLimitedUntil`. Because all
+// the defensive guards lived inside `if (lastErrorType === "quota_exhausted")`, any other
+// error type fell straight through to clearRecoveredProviderState.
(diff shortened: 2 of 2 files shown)


pacocartones:
Closing this — #11310 landed the same fix and #11277 is closed as completed.

Your version is the better shape: hoisting the guards out of the `lastErrorType` branch is the same conclusion, and gating on the error type rather than adding a separate cooldown predicate keeps the surface smaller.

No action needed here. Thanks for the quick turnaround.

Issue #6340, opened 2026-07-06
feat(providers): route same provider's connections native + cliproxyapi — upstream_proxy_config keyed only on provider_id


RaviTharuma:
## Summary

You cannot currently route **some connections of one provider natively (API key)** and **other connections of the same provider through CLIProxyAPI (OAuth/login)** at the same time, because the CLIProxyAPI routing mode is keyed on `provider_id` only.

## What I found in the code (current `main`, v3.8.45)

- `upstream_proxy_config.provider_id` is `UNIQUE`; `src/lib/db/upstreamProxy.ts` reads/writes with `WHERE provider_id = ?` and `ON CONFLICT(provider_id) DO UPDATE`. There is no `connection_id` dimension.
- `resolveExecutorWithProxy(prov)` (`open-sse/handlers/chatCore/executorProxy.ts`) receives only the provider string and returns the CLIProxyAPI executor for the whole provider when `mode === "cliproxyapi"`. So *all* connections of e.g. `anthropic` move together.
- The columns that could allow finer routing are stored but **not used by the routing layer**:
  - `family` (default `"auto"`) — I could not find any handler/executor that branches on `upstream_proxy_config.family`; the combo context cache keeps only `{ mode, enabled }`.
  - `cliproxyapi_model_mapping` — stored via the settings API (including a `provider_id: "cliproxyapi"` sentinel row whose own comment says it is "NOT used for routing"), but not read by `resolveExecutorWithProxy`.

Net effect: for a provider like `anthropic`, you can't have `anthropic` (API key) → `native` and `anthropic` (OAuth) → `cliproxyapi` concurrently. The current workaround is to address CLIProxyAPI as its own provider (`cliproxyapi`) and keep the OAuth logins inside CPA's auth-dir — which works, but means the API-key and OAuth flavors of the same underlying model live under two different provider IDs.

## Request

Allow the same provider to route different connections differently — either:
1. per-connection routing override (see companion issue), or
2. make `family` / `cliproxyapi_model_mapping` actually influence routing so a subset (by auth type, family, or model) of a provider's traffic can go through CLIProxyAPI while the rest stays native.

## Why it's useful

Lets users keep native API-key accounts (full OmniRoute-side accounting) while sending only the login/subscription accounts of the *same* provider through CLIProxyAPI, without splitting them into a separate provider id.

RaviTharuma:
Companion issue: #6339 (per-connection routing override). Implementing #6339 would also resolve this.

[...]

PR #11314, opened 2026-08-24
feat(services): expose sanitized CLIProxyAPI account health

### LINKED ISSUE #6342: feat(providers): consume CLIProxyAPI /v0/management for account-aware routing and quota display
## Summary

When a provider is routed through CLIProxyAPI (`mode: cliproxyapi` / `fallback`), OmniRoute treats CPA as an opaque HTTP backend: it POSTs to `/v1/chat/completions` or `/v1/messages` and reads the response. It never queries CPA's **management API** (`/v0/management`), so CPA's per-account state — which accounts exist, their quota/rate-limit status, and which account actually served a request — is invisible to OmniRoute.

The integration point seems half-prepared: the provider catalog already declares the prefix, but nothing consumes it.

## What I found in the code (current `main`, v3.8.45)

- `src/shared/constants/providers/upstream-proxy.ts:16` declares:
  ```ts
  cliproxyapi: {
    id: "cliproxyapi",
    ...
    healthEndpoint: "/v1/models",
    managementPrefix: "/v0/management",   // <-- declared here
    ...
  }
  ```
- A repo-wide search for `managementPrefix` / `/v0/management` returns **only** this one definition — there is no `fetch` to it anywhere in `src/` or `open-sse/`.
- Consequences on the request path (`open-sse/executors/cliproxyapi.ts`): the executor sends only `credentials.apiKey`/`accessToken` as a Bearer token to authenticate *to CPA*; CPA then selects and rotates its own upstream accounts internally. OmniRoute never learns which account was used, so:
  - When CPA internally rotates from a rate-limited account to a healthy one and returns `200`, OmniRoute records the usage against whatever native `connectionId` it selected — which may not be the account CPA actually used. Per-connection usage/quota accounting is therefore skewed for CPA-routed traffic.
  - When CPA exhausts all its accounts and returns `429`, OmniRoute applies a connection cooldown to a native connection that may be healthy at the upstream level.

## Request

Wire OmniRoute to CPA's management API (the already-declared `managementPrefix: "/v0/management"`) to surface CPA-side account state, e.g.:

- Read CPA's account list + per-account quota / rate-limit status and show it in the dashboard for `cliproxyapi`-routed providers (today there is no CPA account/quota panel).
- Optionally use that state to make routing/cooldown decisions account-aware, so OmniRoute's connection cooldown and usage attribution reflect the account CPA actually used rather than the one OmniRoute guessed.

## Why it's useful

Right now the integration is shallow (opaque proxy). CPA already exposes this data over `/v0/management`, and OmniRoute already has the prefix constant — consuming it would give accurate per-account visibility and cooldown/usage attribution for the increasingly common "OAuth subscriptions via CPA" setup.

Related to the routing-granularity issues #6339 and #6340.

RaviTharuma:
## Summary

- query CLIProxyAPI `GET /v0/management/auth-files` through a 5-second timeout-bounded, read-only client
- expose only an explicit health allowlist through an OmniRoute management-authenticated API
- show account state and aggregate success/failure counts in the existing CLIProxyAPI service tab
- provision embedded CLIProxyAPI management auth from OmniRoute's encrypted service key; require `CLIPROXYAPI_MANAGEMENT_KEY` for external instances
- document the external management-key contract in `.env.example` and the environment reference

The response intentionally excludes names, filesystem paths, email/account identifiers, tokens, status messages, raw metadata, and every unknown upstream field. This is visibility only: account attribution and per-account routing remain blocked by CLIProxyAPI's current data-plane contract and are tracked separately in #6340.

Closes #6342
Related #6340

## Tests Added Or Updated

- client contract: allowlist redaction, management auth, timeout, and disabled/missing/unreachable/unauthorized/unsupported/invalid response states
- API contract: OmniRoute management authentication and no-store response
- embedded service: management password injection without config-file persistence
- dashboard: CLIProxyAPI tab exports and mounts the read-only account card

## Verification

- focused Node test run: 16/16 passed after rebase
- focused ESLint: passed after rebase
- `npm run check:env-doc-sync`: passed
- `git diff --check`: passed
- full `tsc --noEmit`: attempted; blocked by pre-existing errors in tray typings, stream Transformer `cancel` typings, and an unrelated ad-hoc script


RaviTharuma:
Implementation evidence:

- rebased onto current `release/v3.8.50` (`ac02c5b42`)
- 16 focused client/API/service/UI tests pass
- focused ESLint and env-doc sync pass
- PR is mergeable; required CI is running

Security boundary: the OmniRoute response is constructed from a strict allowlist and never forwards CPA names, paths, email/account identifiers, tokens, status messages, raw metadata, or unknown fields. The management secret is only sent server-to-server and is never returned.

PR #10814, opened 2026-08-20
fix(admission): reserve Responses and Messages bodies before clone

RaviTharuma:
Rebased onto current `release/v3.8.50` with **no** fork-CI commits and **no** rewrite of `chatBodyAdmission.ts`.

This closes the gap in #10786: `POST /v1/responses`, `POST /v1/responses/*`, and `POST /v1/messages` still ran `withInjectionGuard` (unconditional `request.clone()` + `.json()`) without the process-wide budget that Chat Completions already uses.

`withChatAdmission` is a thin composer around the shipped APIs:

- `admitChatRequest` (process-wide budget + per-session fairness, #10110/#9654)
- `releaseChatAdmissionAfterHandler` (lease held through SSE / released on handler failure)

It does not add a second admission controller. Composition is `withChatAdmission(withInjectionGuard(postHandler))` so reservation happens before clone.

Tests call the shipped `admitChatRequest` / `ChatAdmissionController` / `withChatAdmission` (second large body -> 503 `chat_admission_busy`, handler not invoked).

Supersedes closed #10789 (that branch was cut from an older `chatBodyAdmission` and carried unrelated fork CI). Remaining handleChat aliases are still tracked in #10790.

Refs #10786 #7847

Issue #10787, opened 2026-08-19
fix(backend): heavyweight admission is not applied to POST /v1/responses or /v1/messages


RaviTharuma:
## Summary

On the **v3.8.49 Hub image**, heavyweight chat admission (`admitChatRequest`) is wired only on `POST /v1/chat/completions`. Native Responses and Messages routes still enter `withInjectionGuard`, which clones the request and `json()`-parses the full body **before** any lease is taken.

That gap lets overlapping long-context Responses/Messages traffic skip the reservation that #7853 added for Chat Completions, so the process can still hit a V8 heap abort instead of returning a controlled 503 with `Retry-After`.

## Confirmed on tag `v3.8.49`

`src/app/api/v1/responses/route.ts` still ends with:

```ts
export const POST = withInjectionGuard(postHandler);
```

There is no `admitChatRequest` (or `admitChatStructure`) on this route.

`src/app/api/v1/messages/route.ts` is the same wrapper:

```ts
export const POST = withInjectionGuard(postHandler);
```

`src/middleware/promptInjectionGuard.ts` clones first:

```ts
const cloned = request.clone();
parsedBody = await cloned.json().catch(() => null);
```

By contrast, `src/app/api/v1/chat/completions/route.ts` reserves capacity first:

```ts
const admissionResult = await admitChatRequest(request);
if (admissionResult.admit === false) return admissionResult.response;
```

and only then parses JSON from the already-buffered request (no extra clone).

## Why this matters

Clients that speak OpenAI Responses (`/v1/responses`) or Anthropic Messages (`/v1/messages`) never hit the Chat Completions admission path. Two overlapping heavy requests on those routes still:

1. clone the inbound body
2. materialize a parsed JSON graph
3. then enter `handleChat` (compression, combo clones, stringify, streaming retention)

without a process-wide heavyweight lease. That is the same peak-memory amplification tracked in #7847, now reachable because the admission fence from #7846/#7853 does not wrap these routes.

## Synthetic reproduction

No production request bodies, user text, or logs.

1. Use the v3.8.49 Hub image.
2. `POST /v1/responses` with a synthetic payload of about **3 MiB**: dummy `input`/`messages` plus dummy `tools` only (filler strings, no real user text).
3. While that request is still in flight, `POST /v1/responses` a second payload of about **0.8 MiB**, same dummy-messages/tools construction.
4. Observe process-level heap growth / V8 abort rather than a controlled rejection.

[...]

PR #11323, opened 2026-08-24
feat(resilience): add hierarchical concurrency admission

### LINKED ISSUE #7778: feat(resilience): hierarchical concurrency limits and queues at global, provider, and account levels
## Summary

OmniRoute should allow operators to define concurrency limits at **three independent, cumulative levels**:

1. **Global** — maximum concurrent upstream requests for the whole OmniRoute process
2. **Provider** — maximum concurrent requests shared by all accounts of one provider, for example all Codex accounts together
3. **Account / connection** — maximum concurrent requests for one individual provider account

An individual request should run only when capacity is available at **every applicable level**.

This is important because each limit protects a different resource:

- the **global limit** protects OmniRoute itself and the host from memory/CPU exhaustion;
- the **provider limit** protects a provider-wide quota or operational budget;
- the **account limit** protects the quota and stability of a particular credential/account.

Today OmniRoute has pieces of this model, but not the complete hierarchy.

---

### `requestQueue.concurrentRequests` is not global

The dashboard exposes **Settings → Resilience → Request Queue → Max Concurrent Requests**, but the implementation creates multiple independent Bottleneck limiters.

`open-sse/services/rateLimitManager.ts::getLimiterKey()` scopes them approximately as follows:

- **Codex:** connection + quota scope (`codex` or `spark`)
- **GitHub / Gemini:** connection + model
- **most other providers:** connection

Consequently, `concurrentRequests: 32` does **not** guarantee that OmniRoute executes at most 32 upstream requests overall. Two active Codex accounts can each receive their own capacity. A second model/scope or another provider can create further independent capacity.

### Account limits exist, but only at the lowest level

`provider_connections.max_concurrent` and `open-sse/services/accountSemaphore.ts` provide an account-scoped guard. That is useful, but it cannot express either of these requirements:

- “Never run more than 32 requests across the entire OmniRoute process.”
- “Never run more than 24 Codex requests across all Codex accounts.”

Its waiting queue also currently has separate fixed behavior, which can reject requests earlier than the Resilience request queue configured by the operator.

### There is no provider-wide aggregate limit

If three Codex accounts each have an account limit of 16, OmniRoute may execute up to 48 Codex requests. There is no provider-level setting that can keep their combined total at, for example, 24 or 32.

---

## Real-world example

A memory-bounded Kubernetes deployment needs to accept bursts of approximately 50 agent turns while allowing only 32 expensive upstream turns to execute simultaneously.

Tested settings:

- OmniRoute 3.8.48-derived Docker image
- one pod / one OmniRoute process
- `requestQueue.concurrentRequests = 32`
- `requestQueue.maxWaitMs = 180000`
- Codex `rate_limit_protection = true`
- Codex account `maxConcurrent = null`
- routing strategy `fill-first`

Staged tests completed successfully:

- 14/14 requests
- 32/32 requests
- 50/50 requests
- no pod restart, readiness loss, queue error, or process crash

For the tested **single-account, single-Codex-scope** path, the existing queue effectively behaved like a 32-slot queue and absorbed the remaining burst. But that property disappears as soon as routing uses another account, Codex Spark, another provider, or another separately keyed limiter.

The operator therefore cannot safely express the actual requirement: **accept the burst, queue the excess, but never exceed 32 executing upstream requests in total.**

---

## Proposed hierarchical model

A possible configuration model could be:

```json
{
  "concurrency": {
    "global": {
      "maxConcurrent": 32,
      "maxQueueDepth": 128,
      "maxWaitMs": 180000
    },
    "providers": {
      "codex": {
        "maxConcurrent": 24,
        "maxQueueDepth": 96,
        "maxWaitMs": 180000
      },
      "antigravity": {
        "maxConcurrent": 8
      }
    }
  }
}
```

Account limits can continue to live on each provider connection:

```json
{
  "connectionId": "codex-account-a",
  "maxConcurrent": 16,
  "maxQueueDepth": 40,
  "maxWaitMs": 180000
}
```

Example effective behavior:

| Level | Limit |
|---|---:|
| Entire OmniRoute process | 32 |
| All Codex accounts combined | 24 |
| Codex account A | 16 |
| Codex account B | 12 |
| All Antigravity accounts combined | 8 |

With those settings:

- total upstream execution never exceeds **32**;
- Codex execution never exceeds **24**;
- account A never exceeds **16**;
- account B never exceeds **12**;
- Antigravity never exceeds **8**;
- requests that cannot run immediately wait in a controlled queue rather than being dropped unexpectedly.

---

### All applicable limits must be respected

Before upstream execution begins, capacity must be available at:

1. global level;
2. selected provider level;
3. selected account level;
4. any existing model/quota-specific rate limiter.

The effective concurrency is therefore constrained by the smallest currently applicable capacity, without requiring operators to manually divide a global budget among a changing number of accounts.

### Avoid nested-slot starvation

A request should not hold a scarce global slot for a long time while merely waiting for an account slot. A naïve sequence of independently acquired semaphores can create head-of-line blocking and underutilization.

The scheduler should either:

- admit a request only when all required concurrency slots can be acquired together; or
- otherwise guarantee that queued requests do not reserve outer-level capacity prematurely.

Slots must always be released in exception, timeout, cancellation, disconnect, and fallback paths.

### Queue behavior should be explicit and configurable

Each level should support clear queue policy:

- `maxConcurrent`
- `maxQueueDepth`
- `maxWaitMs`
- enabled/unlimited/inherited semantics
- FIFO by default, with starvation-safe fairness between providers/accounts
- cancellation when the client disconnects
- a structured `429` response identifying the blocking level (`global`, `provider`, `account`, or scoped rate limiter)

`null`, `0`, “unlimited,” and “inherit” should have unambiguous documented meanings. In particular, disabling a limit must not accidentally disable an unrelated safety queue.

### Routing and fallback must not bypass the hierarchy

Switching from account A to account B must not create additional global or provider capacity. The selected account changes, but the request still belongs to the same global and provider budgets.

The same requirement applies to:

- `fill-first`
- round-robin
- sticky routing
- quota-based fallback
- Codex standard vs. Spark scopes
- combo routing
- retries after upstream failure

A retry/fallback attempt must not leak, duplicate, or temporarily over-allocate slots.

---

## Dashboard and observability

The dashboard should make the scope of every limit unmistakable instead of presenting a scoped value as simply “Max Concurrent Requests.”

Suggested live counters:

| Scope | Running | Queued | Limit | Oldest wait | Timeouts | Rejections |
|---|---:|---:|---:|---:|---:|---:|
| Global | … | … | 32 | … | … | … |
| Provider: Codex | … | … | 24 | … | … | … |
| Account: A | … | … | 16 | … | … | … |

Request/call logs should record at least:

- total request duration;
- total queue-wait duration;
- upstream execution duration;
- which level caused waiting;
- which level caused timeout/rejection;
- provider/account finally selected.

Without this split, a request that spent 15 seconds waiting and 3 seconds executing looks like an 18-second upstream call, making concurrency diagnosis unnecessarily ambiguous.

---

## Backward compatibility

- Existing behavior should remain the default when no hierarchical settings are configured.
- Existing per-limiter rate-limit protection should remain available as an inner layer.
- Existing account `maxConcurrent` values should migrate cleanly into the account level.
- Existing API/configuration fields should either retain their meaning or be clearly deprecated with migration guidance.

This feature is about **concurrent execution and admission**, not replacing provider RPM/token quota handling.

---

### Manually divide capacity between accounts

For example, two accounts at 16 each approximate a provider total of 32. This breaks when an account is disabled, added, quota-exhausted, or later becomes eligible again. It also cannot enforce a global cap across different providers.

### Use only `fill-first`

This can make a per-account limiter look global during normal operation, but failover to another account can temporarily multiply capacity. It also concentrates load and is a routing policy, not a resource-safety guarantee.

### Use only account `maxConcurrent`

Account limits protect individual credentials but cannot protect the whole OmniRoute process or a shared provider budget.

### Use only `requestQueue.concurrentRequests`

The current value is applied to independently keyed limiters and therefore does not define one aggregate ceiling.

Related but distinct: #6593 discusses admission depth for existing per-triple queues. It does not provide global/provider/account hierarchical concurrency.

---

### Configuration

- [ ] A global `maxConcurrent` can be configured for one OmniRoute process.
- [ ] A provider-level `maxConcurrent` can be configured independently for each provider.
- [ ] An account/connection-level `maxConcurrent` remains configurable per account.
- [ ] Queue depth and wait timeout are configurable at each supported level.
- [ ] Unlimited, disabled, and inherited values have documented, unambiguous semantics.
- [ ] Existing installations preserve current behavior unless operators opt in.

### Enforcement

- [ ] With global limit 32, total executing upstream requests never exceed 32 across multiple providers, accounts, models, and quota scopes.
- [ ] With Codex provider limit 24, all Codex accounts combined never exceed 24.
- [ ] Each account independently respects its own configured limit.
- [ ] Routing, retry, and fallback cannot bypass or multiply global/provider capacity.
- [ ] Queued requests do not reserve outer-level slots while blocked solely by a lower-level limit.
- [ ] Slots are released correctly after success, upstream error, thrown exception, timeout, abort, and client disconnect.

### Queueing and errors

- [ ] Excess requests queue and drain without loss while queue budget remains.
- [ ] Queue-full and wait-timeout responses identify the exact limiting scope.
- [ ] FIFO ordering is deterministic by default.
- [ ] Sustained traffic from one provider/account cannot indefinitely starve another eligible queue.

### Observability

- [ ] Dashboard/API expose running, queued, limit, timeout, and rejection counts globally, per provider, and per account.
- [ ] Request logs distinguish queue wait from upstream execution time.
- [ ] Operators can identify which hierarchy level delayed or rejected a request.

### Tests

- [ ] Integration test with at least two providers and two accounts per provider.
- [ ] Codex standard and Spark scopes tested concurrently.
- [ ] Account failover tested while requests are already running.
- [ ] Round-robin and fill-first tested against the same hierarchical limits.
- [ ] Retry paths prove that permits are neither leaked nor double-counted.
- [ ] Cancellation and timeout tests verify prompt slot release.
- [ ] A burst of 50 requests with global limit 32 proves that no more than 32 execute upstream while the remainder queue successfully.

RaviTharuma:
## Summary

- atomically acquire cumulative global, canonical-provider, and account concurrency gates before the existing scoped Bottleneck limiter
- preserve the existing account semaphore API while adding FIFO `acquireMany()` with no partial reservations, composite idempotent release, queue/abort/timeout/full handling, stats, reset, and cleanup
- hold capacity for the complete upstream response stream and release/reacquire all gates when Codex, ModelScope, or Antigravity rotates accounts
- add an off-by-default whole-process cap (`requestQueue.globalConcurrentRequests`) and explicit per-provider caps (`providerQuotaOverrides.<provider>.providerConcurrency`)
- expose settings through defaults, normalization, PATCH validation/API, dashboard UI, and English/German/Vietnamese labels; relabel the old limiter as per-connection/quota-scope concurrency

## Correctness

A waiting request reserves zero capacity until every applicable gate is available. This prevents a saturated provider/account from consuming the global slot while queued. Duplicate keys collapse to their strictest cap and all release paths are idempotent.

Existing behavior is preserved when new caps are `0` (the default). Existing `acquire()` callers, account block/reset controls, rate-limit queues, and connection-level caps remain compatible.

## Tests Added Or Updated

- `tests/unit/accountSemaphore.test.ts`: cumulative acquisition, no partial reservation, idempotent composite release, timeout, abort, queue-full, stats, cleanup, and legacy single-key behavior
- `tests/unit/chatcore-hierarchical-admission.test.ts`: gate order before `withRateLimit`, all three keys, configured queue budget, and per-rotation reacquisition/release
- `tests/unit/resilience-settings-provider-quota-overrides.test.ts`: schema and normalization for provider caps including disabled zero
- `tests/unit/i18n-vi-completeness.test.ts`: 5/5 pass after synchronizing the new UI labels
- `tests/e2e/resilience-plan-alignment.spec.ts`: default global-cap contract
- focused concurrency/settings result: 29/29 pass
- changed-file ESLint: pass
- Prettier and `git diff --check`: pass
- repository typecheck still reports pre-existing failures in tray typing, stream transformer `cancel`, and an ad-hoc script; no errors reference changed files
- full repository lint still reports pre-existing errors in unrelated dashboard files; changed-file ESLint passes

## Scope note

#10787 was already fixed separately by #10814; this PR does not duplicate it.

Closes #7778


RaviTharuma:
CI follow-up pushed in 317b18645:

- Added the two missing Vietnamese resilience labels so locale parity matches the English settings keys.
- Kept the change scoped to `src/i18n/messages/vi.json`; no concurrency or database behavior changed.
- Fresh CI is running on the rebased head. The earlier 5s DB circuit-breaker timeout was left unchanged as a likely timing flake, not masked by a production-code adjustment.

RaviTharuma:
Focused local verification on head `317b18645` passed:

```
node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test \
  tests/unit/i18n-vi-completeness.test.ts \
  tests/unit/settings-i18n-keys.test.ts \
  tests/unit/accountSemaphore.test.ts \
  tests/unit/chatcore-hierarchical-admission.test.ts \
  tests/unit/resilience-settings-provider-quota-overrides.test.ts
```

Result: 29 tests passed, 0 failed. `git diff --check` also passed. The repository has no `typecheck` script (`npm run typecheck` reports “Missing script”).

The fresh CI run confirmed the Vietnamese locale and the prior DB circuit-breaker timeout are no longer failing: shards 3/4 and 4/4 passed. Shards 1/4 and 2/4 now fail in unrelated, non-PR-owned gates (`dashboard-static-contracts` baseline count 324 vs 321, and OpenAPI documented-route coverage 263/752 vs minimum 271), so I did not alter this PR to mask those release-branch drift failures.

diegosouzapw:

diegosouzapw:
Hi @RaviTharuma — the hierarchical admission design itself is solid (atomic multi-key FIFO, no partial reservations, idempotent composite release, and keeping acquire() as a compatibility wrapper were all the right calls), and the tests cover the right properties. Two blockers before we can merge: (1) your branch carried in ~17 commits that duplicate PRs #11307–#11329 which already landed on release/v3.8.51 via squash merges — please rebase onto the current tip so the diff shrinks to just the admission work; right now the merge would also roll back newer base content. (2) The rewrite silently changes existing account-semaphore semantics: waiters used to time out at 30s with a queue cap of 20; they now inherit requestQueue.maxWaitMs (default 15s) and maxQueueDepth (default 0 = unbounded) even when the new global/provider caps are off. Either preserve the legacy values when the new caps are disabled, or make the new contract explicit in tests and the changelog — as written, 'existing behavior is preserved when new caps are 0' doesn't hold.

diegosouzapw:
Thanks for this — the hierarchical concurrency admission design and implementation are solid and well-tested. I'm closing this in favor of #11493, which carries exactly the same 3 commits (author preserved) without the 17 already-merged commits this branch picked up as baggage during a stale rebase (`release/v3.8.51` had drifted 98 commits behind `release/v3.8.50` when this PR was opened). That baggage was making the diff (90+ files, i18n/TTS/compression work unrelated to concurrency admission) unreviewable as a unit and kept `mergeable` at `CONFLICTING`.

#11493 has merged into release/v3.8.51, closing #7778. Full credit for the feature is yours — thank you for the work.

PR #11334, opened 2026-08-24
perf(providers): lazy validate provider schema on demand to reduce startup heap (#11220)

### LINKED ISSUE #11220: feat(backend): lazy-load provider definitions instead of eager-loading all 348 at startup
**Problem**: The system loads all 348 provider definitions into memory at module import time, even when only a few providers are actually used.

**Root Cause**: In src/shared/constants/providers.ts, the PROVIDERS object is exported as a const that initializes immediately at module load. This pulls in:
- Full provider configs (OAuth, model lists, capabilities) for all 348 providers
- open-sse/config/providerRegistry.ts imports all providers at module load
- open-sse/executors/index.ts re-exports all 128 executor modules
- Any import of these modules triggers full provider tree evaluation

**Impact**: 
- ~50-100MB additional heap at startup for provider metadata
- Module load time increased by ~200-500ms
- Memory pressure affects all deployments regardless of actual provider usage

**Proposed Solution**: 
1. **Lazy provider registry**: Change PROVIDERS to a getter that loads on first access
2. **On-demand executor loading**: Use dynamic imports in open-sse/executors/index.ts instead of static re-exports
3. **Provider metadata caching**: Load provider details once per provider, cache for subsequent requests
4. **Remove barrel exports**: Eliminate open-sse/executors/index.ts barrel file that forces all executors to load

**Expected Benefit**: 
- Startup memory reduced by 50-100MB
- Module load time reduced by 200-500ms
- Only used providers consume memory

**Files to Modify**:
- src/shared/constants/providers.ts - lazy initialization
- open-sse/config/providerRegistry.ts - on-demand loading
- open-sse/executors/index.ts - dynamic imports
- open-sse/handlers/chatCore.ts - provider resolution via lazy registry

**Acceptance Criteria**:
- 
ode -e 
require
./src/shared/constants/providers
 loads in <50ms and <20MB heap
- First request for a provider triggers its load, subsequent requests use cached version
- No regression in provider functionality or combo routing

M-Taimoor-Mansha:
### Summary
Refactored provider schema validation in `src/shared/constants/providers.ts` to evaluate on-demand during first access rather than immediately upon module import.

### Verification & Benchmark Results
Ran non-cached module loading acceptance benchmark using `npx tsx`:
- **Startup Heap:** Reduced to **~13.70 MB** (Meets `<20MB` requirement)
- **Module Load Time:** Reduced to **~173 ms** (Previously >500 ms)

Fixes #11220


diegosouzapw:
Landed via consolidated batch validation (worktree, direct push — your branch's base against `main` had diverged heavily from `release/v3.8.50`'s history, so a normal GitHub merge would have pulled in ~7 unrelated already-shipped commits; cherry-picked just your own commit onto the current tip instead). One adjustment during landing: your diff assumed a single `./providerDefinitions` module that no longer exists on this branch (the provider catalog was split into `./providers/{noauth,oauth,apikey,...}` files since your fork point) — reapplied the same lazy-validation behavior (`ensureProvidersValidated()`, called once on first `AI_PROVIDERS` access) against the current file structure, preserving all currently-registered provider ids. Own perf improvement is in: startup heap validation is now deferred until first access instead of running unconditionally at module load. Thanks!

Issue #5542, opened 2026-06-30
fix: in-app auto-update fails with 'spawn npm ENOENT' — use pnpm/npx fallback or document manual update path


chirag127:
## Problem

Clicking "Update to v3.8.41" in the dashboard triggers a server-side `npm install -g omniroute@...` call that fails with:

\`\`\`
[AutoUpdate] Update failed: Error: spawn npm ENOENT
    errno: -4058, code: 'ENOENT', syscall: 'spawn npm'
    cmd: 'npm install -g omniroute@3.8.41 --ignore-scripts --legacy-peer-deps'
\`\`\`

`npm` is not on PATH in OmniRoute's child process environment (common on Windows where `npm` is a `.cmd` shim that requires cmd.exe context, not available via `spawn`).

## Root causes

1. `spawn('npm', ...)` fails on Windows — `npm` is a `.cmd` file, not a binary. Must use `spawn('npm.cmd', ...)` on Windows, or `spawn('npm', ..., { shell: true })`.
2. Even with shell:true, `npm` may not be on PATH in OmniRoute's environment if it was installed via nvm/fnm/volta.

## Fastest workaround (document in UI)

Until the auto-updater is fixed, the UI should show the manual update command alongside the error:

\`\`\`
npm install -g omniroute@latest
# or if installed via pnpm:
pnpm add -g omniroute@latest
\`\`\`

## Fix

1. On Windows, use `spawn('npm.cmd', [...], { shell: false })` — or detect platform and use the correct binary name.
2. Prefer `npx` or absolute path resolution (`which npm` / `where npm`) to find the binary before spawning.
3. Fall back to showing the manual update command in the UI if spawn fails, instead of crashing the server.

## Acceptance criteria

- [ ] In-app update works on Windows (npm.cmd or shell:true)
- [ ] If update fails, show manual command in UI: `npm i -g omniroute@<version>`
- [ ] Server does not crash/restart on update failure

diegosouzapw:
Thanks for the detailed report! Fixed in `release/v3.8.43` via #5797 (already merged into the active development branch — feel free to pull and test now). It will ship in the next release.

Issue #5379, opened 2026-06-29
[BUG] spawn EINVAL installing embedded services (9Router/CLIProxy) on Windows + Node.js 24+ — runNpm execFile('npm.cmd') needs shell


diegosouzapw:
## Summary

Installing an embedded service (9Router / CLIProxy) from the dashboard fails with **`spawn EINVAL`** on **Windows + Node.js v24+**. Reported via the community Discord with a precise root-cause analysis.

## Environment

- OmniRoute v3.8.39 (also affects the current line — the code path is unchanged)
- Windows 11
- Node.js v24.18.0

## Root cause

In Node.js v24, `child_process.execFile()` no longer executes `.cmd` batch files on Windows without `shell: true` (documented breaking change — nodejs/node#52554). On Windows the npm binary is `npm.cmd` (a batch wrapper), so `execFile("npm.cmd", args, opts)` throws `EINVAL` immediately.

Affected code — `src/lib/services/installers/utils.ts`, `runNpm()` (~line 83-108):

```ts
const npmBin = process.platform === "win32" ? "npm.cmd" : "npm";
execFile(npmBin, args, { cwd, timeout, env: process.env, maxBuffer: ... }, cb);
// ^ no `shell` option -> EINVAL on Node 24 when npmBin ends in .cmd
```

This is reached by the embedded-service install routes (e.g. `/api/services/9router/install`).

## Important: the built bundle, not just the .ts

The production server (`npm install -g omniroute`) runs from the pre-built webpack bundle, where `runNpm()` is compiled into a chunk under `dist/.build/next/server/chunks/`. The `.ts` source is only used by the CLI entry point via `tsx`; the Next.js API routes use the compiled chunks. So a source-only patch must be shipped in a release build to actually reach global installs — worth a regression test that exercises the install path on Windows/Node 24 (or a unit test that asserts `runNpm` passes `shell` on win32).

## Proposed fix

Add `shell: process.platform === "win32"` to the `execFile` options in `runNpm()`.

⚠️ Care required re: **Hard Rule #13** (no shell interpolation): with `shell: true` the args array is joined into a shell command, so the fixed npm subcommands + package names must remain non-interpolated/validated (they already come from the installer registry, not user input). Keep `execFile` + explicit args array; only flip `shell` on Windows. Add a unit test asserting the win32 branch sets `shell` and the non-win32 branch does not.

## Workaround (until fixed)

[...]

PR #11336, opened 2026-08-24
fix(cli): spawn npm with shell on win32 in update command (#11335)

### LINKED ISSUE #11335: [BUG] omniroute update fails on Windows — 'Could not check latest version' (execFile npm without shell in update.mjs, Node 24+ hardening, same class as #5379)
## Summary

`omniroute update` fails on Windows with **"✖ Could not check latest version. Is npm available?"** even though npm is installed and fully working. Same root cause as #5379 (Node.js 24+ no longer spawns `.cmd`/`.bat` via `execFile` without `shell: true` — nodejs/node#52554), but in a different code path: `bin/cli/commands/update.mjs` was missed when `runNpm` was fixed.

## Environment

- OmniRoute v3.8.49 (npm global install) — the same code is present on the `release/v3.8.50` branch (verified via raw file)
- Windows 11
- Node.js v26.5.1

## Repro

```
PS> omniroute update
  📋 Loaded env from C:\Users\Administrator\.omniroute\.env
  ...
✖ Could not check latest version. Is npm available?
```

npm itself works fine in the same shell:

```
PS> npm view omniroute version --prefer-online
3.8.49
```

## Root cause

`bin/cli/commands/update.mjs` → `getLatestVersion()`:

```js
const { stdout } = await execFn("npm", ["view", "omniroute", "version", "--prefer-online"], {
  timeout: 15000,
});
```

`execFile` is called without `shell: true`. On Node.js ≥ 24 (security hardening, nodejs/node#52554), spawning `.cmd`/`.bat` without a shell is rejected. Two failure modes observed on this machine:

| Call | Error |
|---|---|
| `execFile("npm", ...)` bare name | `ENOENT` — PATH contains an extensionless `npm` shim (from another Node install) that `CreateProcess` cannot execute |
| `execFile("npm.cmd", ...)` | `EINVAL` — thrown synchronously |

Both are swallowed by the `catch {}` and reported as "Is npm available?".

The same pattern exists a second time in the same file, in the `--changelog` path (`execFileAsync("npm", ["view", "omniroute", "changelog"], ...)`).

## Suggested fix

Add `shell: process.platform === "win32"` to both `execFile` calls in `update.mjs` (or better: reuse whatever helper #5379 introduced for `runNpm`, if it's shared). Verified locally — with `shell: true` the command works:

```
PS> omniroute update
  Current version: 3.8.49
  Latest version:  3.8.49
✔ You are running the latest version!
```

Note: `execFile(..., { shell: true })` emits DEP0190 on newer Node; invoking `cmd.exe /d /s /c npm ...` with proper quoting avoids the deprecation, but `shell: true` is the minimal change and matches how `execSync` (used by the `--apply` path) already behaves on Windows.

Finn763:
## Summary

`omniroute update` fails on Windows with **"✖ Could not check latest version. Is npm available?"** even when npm is installed and working. Fixes #11335.

## Root cause

`bin/cli/commands/update.mjs` → `getLatestVersion()` runs:

```js
execFile("npm", ["view", "omniroute", "version", "--prefer-online"], { timeout: 15000 })
```

without a shell. On Windows npm is `npm.cmd`, and Node ≥24 refuses to `execFile` a `.cmd` without a shell (nodejs/node#52554 hardening, see #5379). Two failure modes observed on Windows 11 + Node v26.5.1:

| Call | Error |
|---|---|
| `execFile("npm", ...)` bare name | `ENOENT` (an extensionless `npm` shim from another Node install shadows the path) |
| `execFile("npm.cmd", ...)` | `EINVAL` (thrown synchronously) |

Both land in the `catch {}` and surface as the misleading "Is npm available?" message. The `--changelog` lookup in the same file has the identical pattern.

## Fix

Enable the shell on **win32 only** in both npm lookups, mirroring the established CLI fixes from #7913/#6263/#6304 (spawn `.cmd` shims with `shell: true` on win32). The version lookup's exec options are passed through a `platform` parameter (defaults to `process.platform`) so the behavior stays unit-testable; args are static (no user input), so there is no shell-injection surface (Hard Rule #13).

## Tests

Added `tests/unit/cli-update-npm-shell-win32-11335.test.ts` following the style of `cli-update-prefer-online-4376.test.ts`:

- win32 → `opts.shell === true` (+ `--prefer-online` still passed, timeout preserved)
- non-win32 → shell not enabled
- source-level assertion that both npm lookups in `update.mjs` wire the win32 shell

Local verification (Node v26.5.1, Windows 11):

```
$ node --test <harness>     # 4/4 pass against bin/cli/commands/update.mjs
$ omniroute update          # (patched installed copy) → "You are running the latest version!"
```

## References

- Fixes #11335
- Same root cause previously fixed in other code paths: #5379 (embedded services), #5542 (in-app auto-update), #7913/#6263/#6304 (CLI spawns)
- nodejs/node#52554

### DIFF
--- a/bin/cli/commands/update.mjs
+++ b/bin/cli/commands/update.mjs
@@ -29,10 +29,14 @@ export async function getCurrentVersion() {
 // Without it `npm view` can return a stale cached version (e.g. report 3.8.30 as
 // "latest" after 3.8.31 was published), so the updater told users on an old build
 // they were already on the latest version (#4376). `execFn` is injectable for tests.
-export async function getLatestVersion(execFn = execFileAsync) {
+export async function getLatestVersion(execFn = execFileAsync, platform = process.platform) {
   try {
     const { stdout } = await execFn("npm", ["view", "omniroute", "version", "--prefer-online"], {
       timeout: 15000,
+      // On Windows npm is npm.cmd; Node ≥24 refuses to execFile a .cmd without
+      // a shell (nodejs/node#52554 — spawn EINVAL / ENOENT, see #5379), so enable
+      // the shell on win32 only. Mirror the CLI fix from #7913/#6263/#6304.
+      shell: platform === "win32",
     });
     return stdout.trim();
   } catch {
@@ -116,6 +120,8 @@ export async function runUpdateCommand(opts = {}) {
     try {
       const { stdout } = await execFileAsync("npm", ["view", "omniroute", "changelog"], {
         timeout: 10000,
+        // Same win32 shell requirement as getLatestVersion (nodejs/node#52554).
+        shell: process.platform === "win32",
       });
       if (stdout.trim()) {
         console.log(stdout.trim());
--- a/tests/unit/cli-update-npm-shell-win32-11335.test.ts
+++ b/tests/unit/cli-update-npm-shell-win32-11335.test.ts
@@ -1,0 +1,59 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import fs from "node:fs";
+
+const update = await import("../../bin/cli/commands/update.mjs");
+
+// #11335: `omniroute update` failed on Windows with "Could not check latest
+// version. Is npm available?" even though npm worked fine. getLatestVersion()
+// ran execFile("npm", ...) without a shell; on Node ≥24 spawning npm.cmd
+// without a shell throws EINVAL/ENOENT (nodejs/node#52554, see #5379). The
+// fix enables the shell on win32 only, mirroring the CLI fixes from
+// #7913/#6263/#6304.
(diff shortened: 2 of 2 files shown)


diegosouzapw:
Thanks for the fix — this bug (#11335, npm.cmd cannot be execFile'd without a shell on Node ≥24) also came in via #11374, which was merged in the same batch: it extracts a shared `bin/cli/npm-exec.mjs` reused across the CLI entry points, additionally handles the Bun runtime and `windowsHide`, and follows the same shell-safety pattern already established in `src/lib/services/installers/utils.ts` for the server-side calls (literal argv only, Hard Rule #13). Since both fix the identical root cause and #11374 is the more complete/consistent implementation, closing this one — your diagnosis of the root cause was correct and helped confirm the fix direction. Thanks!

Issue #6328, opened 2026-07-06
feat(models): advanced setting to hide paid-only models


chirag127:
## Problem

OmniRoute's `/v1/models` endpoint returns 3,820 models with no tier/pricing metadata. There's no server-side way to distinguish free-usable models from paid-only ones, so:

- `auto/*` combos surface paid models to free-tier users → 402/403 at request time
- Persisted combos (like fusion panels) drop legs when hitting paywalls
- Users on `no-card-on-file` policies have no way to filter out models that will always fail

Empirically observed: probing all 33 top-of-provider models with `max_tokens:1` returned `403` on `openrouter/openai/gpt-5.5-pro`, `dgrid/openai/gpt-5.2-pro`, `ollamacloud/deepseek-v4-pro` (paid-only) — but nothing in `/v1/models`, `owned_by`, or the provider registry marks them as paid.

## Existing infrastructure

`open-sse/services/tierResolver.ts::classifyTier(provider, model)` already returns `{tier: 'free' | 'standard' | 'pro' | 'ultra'}` at the connection level. The signal exists — it just isn't propagated to model listings or filters.

## Proposal

**Advanced setting: hide paid-only models globally.**

1. Add a per-model `pricing.free_tier_available: boolean` field to `open-sse/config/providerRegistry.ts` entries.
2. Add a `HIDE_PAID_ONLY_MODELS` env flag + settings.json toggle (default: `false`, off).
3. When flag is on:
   - `GET /v1/models` filters out entries where `pricing.free_tier_available !== true`
   - `auto/*` candidate pool excludes them
   - Combo validation warns when adding a paid-only target
4. Dashboard toggle at `Settings → Advanced → Hide paid-only models`.

## Data-source options

- Cross-reference `models.dev` pricing data (already used per `ARENA_ELO_SYNC_ENABLED` path).
- Manual annotation in `providerRegistry.ts` for confirmed paid providers.
- Runtime learning: cache first 402/403 per (provider, model) → auto-flag.

## Behavior when disabled (default)

Zero change from today — paid models remain listed and callable.

## Environment

- OmniRoute 3.8.40, Windows 11
- 41 provider connections, 3,820 models
- Fusion combo `fusion-free-verified-22` — 22 verified-free models after live-probing 33 candidates. 11 dropped as auth/403/404.

Thanks for the great work on OmniRoute.

diegosouzapw:
[...]

PR #11337, opened 2026-08-24
fix(catalog): add Stealth Ox Alpha (stealth/ox-alpha) to the openrouter free roster

stanleytejakusuma:
Small data fix so `stealth/ox-alpha` (Stealth Ox Alpha) is visible in `/v1/models` under hidePaidModels.

**Why it's currently invisible:** the model routes fine (free 0/0 pricing, 1M context, mandatory reasoning with native tiers low/high/max, text+image+video input). But the synced-provider-row filter (#6328) drops pricing metadata before `isFreeModel()`, so visibility falls through to the curated `FREE_MODEL_BUDGETS` roster — and the id carries no `:free` suffix either.

**Change:** adds `stealth/ox-alpha` to the openrouter-free pool (0-token budget, no free-pool inflation), bumps `FREE_CATALOG_CURATED_AT`, and adds a regression test. Verified: npx tsx test passes; entry shape matches neighbors; routing smoke 200 on a live 3.8.49 gateway (`x-omniroute-response-cost: 0.0000000000`).

### DIFF
--- a/open-sse/config/freeModelCatalog.data.ts
+++ b/open-sse/config/freeModelCatalog.data.ts
@@ -16,7 +16,7 @@ import type { FreeModelBudget } from "./freeModelCatalog.ts";
  * rewrites file timestamps on every deploy, which would report a months-old
  * catalog as "updated today". Bump this whenever the entries below change.
  */
-export const FREE_CATALOG_CURATED_AT = "2026-08-18";
+export const FREE_CATALOG_CURATED_AT = "2026-08-20";
 
 export const FREE_MODEL_BUDGETS: FreeModelBudget[] = [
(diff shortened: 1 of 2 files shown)


stanleytejakusuma:
Maintainer note (2026-08-24): this model is part of OpenRouter's **Stealth Program** — per its EULA (openrouter.ai/terms/stealth, updated 2026-07-06), access is free *for a limited period* in exchange for User Content collected for Stealth Model training; the free window for stealth/ox-alpha is advertised through ~2026-08-27 (OpenCode's launch post). Practical consequence for this roster entry: if the model's pricing flips to paid after the promo, hidePaidModels will NOT hide it (the synced-row filter falls back to the curated roster, which now says free), and the catalog would advertise it as free while billing. Recommend re-verifying pricing when the promo window closes and dropping/flagging the entry if it goes paid. Also rebased onto the current release/v3.8.50 tip (upstream re-curated the roster on 08-18; kept the newer curated-at date).

PR #11317, opened 2026-08-24
fix(lint): drain release-green hard failures on release/v3.8.50 (#9985)

jonlwheat2-gif:
Drains the live release-green **hard failures** on `release/v3.8.50` tracked in #9985 (see also #10296 — this is the gate blocking the v3.8.50 publish). Rebased on current tip `ac02c5b42`.

## What was failing (RED)

`npm run lint` at tip `ac02c5b42^` reported **22 net-new errors** across 6 files; the release-green eslint step consequently died ("could not parse eslint json" — the report never came back clean, so the verdict never reached a passing state):

```
FirstRunReadinessCard.tsx        23:7   react-hooks/set-state-in-effect
EndpointPageClient.tsx           4:8    unused 'Link'
EndpointPageClient.tsx         141:10,28 unused expandedEndpoint / setExpandedEndpoint
EndpointPageClient.tsx      1135-1138  unused mcpOnline/a2aOnline/mcpToolCount/a2aActiveStreams
EndpointPageClient.tsx        1172:9  unused cloudflaredUrlNotice
EndpointPageClient.tsx        2501:10 unused component EndpointSection (134 dead lines)
CommandPalette.tsx             107:5   react-hooks/preserve-manual-memoization ×2
CommandPalette.tsx             159:5   exhaustive-deps missing 'activePreset'
cli-mcp-call-commands.test     35/135/200/235/286  unused makeCmd/url/url/url/init
cli-resilience-commands.test   2/19/34 unused makeMcpResp/captureStdout/makeCmd
cli-skills-commands.test       2:10    unused makeMcpResp
```

All of it is fallout from already-merged work (#11283 Traffic Inspector header refactor left dead code; the React-Compiler hooks rules newly enforced). No behavior regressions were involved on our side — verified by the full targeted test families below.

## After (GREEN) — per-file changes

**`src/app/(dashboard)/dashboard/FirstRunReadinessCard.tsx`**
Before: dismissal read via `useState(false)` + `useEffect` that called `setVisible(...)` synchronously (:23-33).
After (**lines 15-37**): module-level external store (`readinessListeners`, `subscribeReadiness`, `isReadinessDismissed`) read through `useSyncExternalStore` (**line 46**) with server snapshot = dismissed (**hydration-safe**); `dismiss` is now a `useCallback` that writes storage and notifies listeners (**48-56**); render guard `if (setupComplete || dismissed)` (**58**). Net effect: zero setState-in-effect, no cascading renders, same UX including private-mode fallback.

[...]

PR #11340, opened 2026-08-24
feat(providers): publish web-session credential contract

Zartharas:
## Summary

Publish a management-authenticated, versioned web-session credential contract at `/api/providers/web-session-contract`.

The contract is derived from OmniRoute's canonical browser-session metadata:

- `listExtractionConfigs()`
- `getWebSessionCredentialRequirement()`

Only non-secret metadata is returned. Credential values, placeholders, instructions, polling configuration, and operator-only guidance are excluded.

## Why

External credential brokers otherwise need to duplicate OmniRoute's browser-session credential schema and can drift as provider requirements change.

This is the narrow upstream-suitable contract-publication portion remaining after aggregate Auth Keeper integration work was separated into independent upstream fixes.

## Scope

- `src/lib/providers/webSessionContract.ts`
- `src/app/api/providers/web-session-contract/route.ts`
- `tests/unit/web-session-contract.test.ts`

No executor, routing, inference, scheduler, recovery, database, provider-asset, dashboard, CLI, or credential-value behavior changes.

## Validation

Against active `release/v3.8.51` base `3192eb88d5550de4c3fd9985564f6b5641e9d681`:

- contract/canonical regression: **22/22 PASS**
- post-format contract regression: **4/4 PASS**
- formatter-only correction: **byte-for-byte Prettier proof PASS**
- provider consistency: **PASS**
- translate-path golden regression: **PASS**
- Prettier contract surface: **PASS**
- scoped ESLint on all changed TS files: **PASS**
- core typecheck: **PASS**
- explicit-any budget: **PASS**
- tracked-artifact policy: **PASS**
- diff whitespace validation: **PASS**
- release reconciliation: **PASS**

### Existing base-red full-repository lint

The active release base currently reports the known 22-error lint set across:

- FirstRunReadinessCard.tsx
- EndpointPageClient.tsx
- CommandPalette.tsx
- cli-mcp-call-commands.test.ts
- cli-resilience-commands.test.ts
- cli-skills-commands.test.ts

plus the existing stale-directive warning in EditConnectionModal.tsx.

All seven source blobs and the lint configuration are identical between the active release base and this PR head. This patch changes none of those files.

The exact 22-error set is the same release baseline tracked and fixed independently by merged upstream PR #11317.

### Existing base-red provider asset gate

`npm run check:provider-assets` reports:

```
freebuff.png: 512x512 exceeds 256px max dimension
```

This is also inherited from the release base:

- pristine base reproduces exactly this single asset failure;
- base and PR head contain the same `freebuff.png` blob;
- base and PR head contain the same asset-checker blob;
- this PR changes no provider asset.

Broad PR CI remains authoritative for the repository-wide matrix.

### DIFF
--- a/changelog.d/features/11340-web-session-contract.md
+++ b/changelog.d/features/11340-web-session-contract.md
@@ -1,0 +1,1 @@
+- **feat(providers):** publish a management-authenticated versioned web-session credential contract from OmniRoute's canonical browser credential metadata ([#11340](https://github.com/diegosouzapw/OmniRoute/pull/11340)) — thanks @Zartharas
--- a/src/app/api/providers/web-session-contract/route.ts
+++ b/src/app/api/providers/web-session-contract/route.ts
@@ -1,0 +1,10 @@
+import { NextResponse } from "next/server";
+import { requireManagementAuth } from "@/lib/api/requireManagementAuth";
+import { buildWebSessionContract } from "@/lib/providers/webSessionContract";
+
+export async function GET(request: Request) {
+  const authError = await requireManagementAuth(request);
+  if (authError) return authError;
+
+  return NextResponse.json(buildWebSessionContract());
+}
--- a/src/lib/providers/webSessionContract.ts
+++ b/src/lib/providers/webSessionContract.ts
@@ -1,8 +1,8 @@
 import {
   listExtractionConfigs,
   type TokenSource,
-} from "../../../open-sse/services/tokenExtractionConfig.ts";
-import { getWebSessionCredentialRequirement } from "../../shared/providers/webSessionCredentials.ts";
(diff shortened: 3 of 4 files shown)


diegosouzapw:
Clean, narrowly-scoped addition — we verified on the current release/v3.8.51 tip that the route sits behind requireManagementAuth before serialization, publishes only non-secret metadata (cookie/localStorage key names, never values), and that all 4 tests pass. Two suggestions, neither blocking: (1) the auth regression test greps the route source for the auth call — a behavioral test asserting GET without a management token returns 401/403 would survive refactors; (2) consider adding the endpoint to docs/openapi.yaml / API_REFERENCE.md so external broker integrators have a stable reference. Good separation from the larger Auth Keeper work.

PR #11364, opened 2026-08-24
fix(compression): use pathToFileURL for workerUrl to prevent bundler resolution failure

TheDemonTuan:
## Summary

- Use `pathToFileURL(join(dir, ...))` instead of `new URL(..., import.meta.url)` in `open-sse/services/compression/compressionWorkerPool.ts`.
- Fixes `Module not found: Can't resolve 'compressionWorker.js'` during Webpack / Next.js production builds (`next build` / Docker builder).

### Background
When Webpack and Turbopack statically scan source files during `next build`, `new URL("compressionWorker.js", import.meta.url)` is treated as a static asset/worker import. Since `compressionWorker.js` does not exist during Next build (only the `.ts` source exists), both bundlers fail resolution. Constructing the file URL dynamically via `pathToFileURL` prevents the static bundler interception while preserving the correct runtime `URL` resolution for Node.js Worker threads.

## Validation
- Change type: bug fix
- Reconciled with `upstream/release/v3.8.50`
- Tested with Next build / Webpack bundling

### DIFF
--- a/changelog.d/fixes/compression-worker-bundler-resolve.md
+++ b/changelog.d/fixes/compression-worker-bundler-resolve.md
@@ -1,0 +1,1 @@
+- **fix(compression):** use `pathToFileURL` in `compressionWorkerPool` so bundlers (Webpack / Turbopack) do not attempt static asset resolution of missing `compressionWorker.js` during build
--- a/open-sse/services/compression/compressionWorkerPool.ts
+++ b/open-sse/services/compression/compressionWorkerPool.ts
@@ -1,6 +1,6 @@
 import { existsSync } from "node:fs";
 import { dirname, join } from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import { Worker } from "node:worker_threads";
 import type { CompressionResult } from "./types.ts";
 import type { StackedCompressionStep } from "./strategySelector.ts";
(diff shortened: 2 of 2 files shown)


xiaoyaner0201:
I reproduced this on the exact current `release/v3.8.50` head (`dafb4ae808305c9287c751cf1ce16fe2234ba073`) using the repository Dockerfile's documented Webpack fallback.

Build parameters:

```text
OMNIROUTE_USE_TURBOPACK=0
OMNIROUTE_BUILD_MEMORY_MB=12288
CIRCLE_NODE_TOTAL=5  # 4 workers
platform=linux/arm64
```

The build remained within its memory budget and reached the end of the Webpack compilation pass, then failed after approximately 3852 seconds with:

```text
./open-sse/services/compression/compressionWorkerPool.ts
Module not found: Can't resolve 'compressionWorker.js'

Import trace:
./open-sse/services/compression/strategySelector.ts
./open-sse/handlers/chatCore.ts
./open-sse/index.ts
./src/instrumentation-node.ts
./src/instrumentation.ts
```

The same resolution failure was also reached through `src/app/api/compression/preview/route.ts`. This was not an OOM or a generic Docker failure.

That confirms the `new URL("compressionWorker.js", import.meta.url)` static-resolution path is a real release-head build blocker for the documented Webpack fallback. I am rebuilding the same base with this PR's `pathToFileURL(join(...))` change and will validate both the final standalone artifact and worker runtime path before using the image.

xiaoyaner0201:
Follow-up from the same arm64 Docker build: the PR fixes the Webpack resolution error and the image now builds successfully, including generation of `/app/open-sse/services/compression/compressionWorker.js`.

However, a direct runtime Worker smoke of that final image still fails before it can process a job:

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'omniglyph'
imported from /app/open-sse/services/compression/compressionWorker.js
```

The final worker bundle was built with `esbuild --packages=external`. Its non-builtin runtime imports are:

```text
@toon-format/toon
omniglyph
omniglyph/applicability
safe-regex
smol-toml
zod
```

Only `zod` is resolvable in the final standalone image. The missing dependency closure is:

```text
@toon-format/toon
omniglyph
  └─ gpt-tokenizer
safe-regex
  └─ regexp-tree
smol-toml
```

So the URL fix is necessary and does unblock compilation, but the compression worker is not yet runnable in the Docker standalone artifact. A release-level regression should instantiate the final colocated `compressionWorker.js` from the assembled standalone tree and complete an `rtk + caveman` job, rather than only asserting that the worker file exists.

PR #11366, opened 2026-08-24
fix(endpoint): remove stray page-header JSX spliced into renderModelGroup

echel0nn:
## Problem

`next build` aborts on the `/dashboard/endpoint` route:

```
src/app/(dashboard)/dashboard/endpoint/EndpointPageClient.tsx
Error: Unterminated regexp literal
```

Commit 65e81158a ("fix(ollama): route models by advertised capability", #11088) spliced a block of endpoint page header markup (`endpoint.title`, `endpoint.subtitle`, base URL, `endpoint.advancedProtocols`) into the middle of the `ProviderModelsModal` `renderModelGroup` `.map()` callback. The injected `<div className="flex flex-col gap-4">` wrapper is never closed, so the JSX is unbalanced. Turbopack reports it as "Unterminated regexp literal" and the production build fails.

The dev server compiles routes on demand, so the break stays invisible until the endpoint route is compiled or `next build` runs over every route.

## Fix

Restore the `renderModelGroup` map callback to return only the model row, matching the clean state in v3.8.49 (`c9d4a45f1`). Net change: 1 insertion, 13 deletions.

## Verification

- `esbuild EndpointPageClient.tsx` parses cleanly (hard parse error before).
- The keys `endpoint.title` / `endpoint.subtitle` / `endpoint.advancedProtocols` / `endpoint.testEndpoint` appear nowhere else in the file and did not exist in v3.8.49, confirming the injected block was spurious.


diegosouzapw:
Verified subsumed: cherry-picked your fix commit (`36f8839a`) directly against the current `release/v3.8.50` tip and it applied as a clean no-op (`nothing to commit, working tree clean`) — the exact bug you diagnosed (stray page-header JSX spliced into `renderModelGroup`'s `.map()` callback by #11088) has already been fixed by a subsequent, unrelated change to `EndpointPageClient.tsx` already on the release branch. Your diagnosis of the root cause was correct and matched. Closing as already covered — thanks for the catch!

Issue #5326, opened 2026-06-29
[BUG] Antigravity "Token Expired Badge" continuation of #3679 #3850


3xa228148:
### OmniRoute Version

3.8.39

### Installation Method

npm (global)

### Operating System

Linux

### Description

The problem still occurs on new versions, and the fix didn't help. 

<img width="1556" height="234" alt="Image" src="https://github.com/user-attachments/assets/63ed8a13-9311-40a4-9395-6c924fd68a05" />

### Steps to Reproduce

1 Add an account
2 Wait

### Expected Behavior

Login tokens that work for a long time

### Actual Behavior

<img width="1556" height="234" alt="Image" src="https://github.com/user-attachments/assets/63ed8a13-9311-40a4-9395-6c924fd68a05" />

### Test Impact

Unsure

### Error Logs / Output

```shell

```

3xa228148:
thiss issue is continuation of #3679 #3850

diegosouzapw:
Fixed in `release/v3.8.41` (#5347, merged) — ships in v3.8.41. Two causes addressed: new OAuth connections now set `tokenExpiresAt` at creation (the badge no longer flashes a false expired before the first refresh), and a refresh-capable connection with no usable refresh token is now surfaced as a real `expired` status (needs re-auth) instead of silently staying `active` with a confusing cosmetic badge. Thanks @3xa228148!

PR #11368, opened 2026-08-24
fix(db): keep tokenExpiresAt when a connection is created

ntdatt812:
## Summary

`createProviderConnection` copies optional fields onto the row through an allowlist (`src/lib/db/providers.ts:583`). `tokenExpiresAt` is not on it, and the initial object at `:568` does not set it either — so `_insertConnectionRow` always binds

```ts
tokenExpiresAt: conn.tokenExpiresAt || null,   // providers.ts:707
```

as **NULL on every insert**, however good the payload was.

That makes #5326 a no-op end to end. `buildOAuthConnectionCreatePayload` mirrors the computed expiry into `tokenExpiresAt` (`connectionPersistence.ts:105`) precisely so the dashboard badge does not flash "Token Expired" before the first background refresh — and the value is discarded one layer down. The update path already carries it (`providers.ts:762`), so a connection only gained the field after its first refresh.

One string on the allowlist.

## Related Issues

- Related to #5326 (whose fix this restores end to end)

## Validation

- [x] Change type: DB
- [x] `npm run lint` — **0 net new** on the changed file (4 pre-existing `no-unused-vars`, same before and after)
- [x] Reconciled with the current active release base (`release/v3.8.50`, `dafb4ae80`)

```
tests/unit/oauth-connection-tokenexpiresat-5326.test.ts   3 pass, 0 fail
  + oauth-refresh-connection-dedup-8059                   7 pass, 0 fail
```

**Written test-first**: the new case fails on the base with `tokenExpiresAt: null`, and passes with the one-string change. **Mutation-checked**, after confirming the edit applied — removing the string again gives 2 pass / 1 fail, exactly the new case.

## Tests Added Or Updated

- `tests/unit/oauth-connection-tokenexpiresat-5326.test.ts` — one case that persists the payload with `createProviderConnection` and reads it back through `getProviderConnections`.

The two existing cases are unchanged. Worth noting *why* they did not catch this: both assert the payload object and neither touches the database, so the create path could drop the field for as long as it did with the regression test green. The file now covers both halves.

## Coverage Notes

`src/lib/db/providers.ts` — the added allowlist entry is exercised by the round-trip case, and the mutation confirms it is the line under test rather than incidental.

## Reviewer Notes

- One-string change to a create-path allowlist; no schema change, no migration. The column and its binding already existed.
- I did **not** touch the re-auth path. `connectionPersistence.ts:146` and the four route sites spread `...tokenData` + `expiresAt` but omit `tokenExpiresAt`, which looks like the same omission on update — but both consumers prefer `tokenExpiresAt` and the create path is now correct, so I would rather that be judged separately than widened into this PR.

### DIFF
--- a/changelog.d/fixes/11368-create-connection-tokenexpiresat.md
+++ b/changelog.d/fixes/11368-create-connection-tokenexpiresat.md
@@ -1,0 +1,1 @@
+- **Provider connections:** keep `tokenExpiresAt` when a connection is created. The create-path allowlist omitted it, so every insert stored NULL and the dashboard token badge could read a fresh connection as expired until its first background refresh ([#11368](https://github.com/diegosouzapw/OmniRoute/pull/11368)).
--- a/src/lib/db/providers.ts
+++ b/src/lib/db/providers.ts
@@ -625,6 +625,10 @@ export async function createProviderConnection(data: JsonRecord) {
     "accessToken",
     "refreshToken",
     "expiresAt",
+    // #5326's payload sets this and _insertConnectionRow binds it, but it was
+    // missing from this allowlist — so every created row stored NULL however good
+    // the payload was. The update path already carries it (`data.tokenExpiresAt`).
+    "tokenExpiresAt",
     "tokenType",
     "scope",
     "idToken",
--- a/tests/unit/oauth-connection-tokenexpiresat-5326.test.ts
+++ b/tests/unit/oauth-connection-tokenexpiresat-5326.test.ts
@@ -1,7 +1,22 @@
 import test from "node:test";
 import assert from "node:assert/strict";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
 
-import { buildOAuthConnectionCreatePayload } from "../../src/lib/oauth/connectionPersistence.ts";
+const TEST_DATA_DIR = fs.mkdtempSync(path.join(os.tmpdir(), "omniroute-tokenexpiresat-"));
+process.env.DATA_DIR = TEST_DATA_DIR;
+process.env.API_KEY_SECRET = process.env.API_KEY_SECRET || "tokenexpiresat-test-secret";
+
+const core = await import("../../src/lib/db/core.ts");
+const providersDb = await import("../../src/lib/db/providers.ts");
+const { buildOAuthConnectionCreatePayload } =
+  await import("../../src/lib/oauth/connectionPersistence.ts");
+
+test.after(() => {
+  core.resetDbInstance();
+  fs.rmSync(TEST_DATA_DIR, { recursive: true, force: true });
+});
 
 // Regression for #5326: a freshly created OAuth connection (e.g. antigravity) used
 // to persist only `expiresAt`, leaving `tokenExpiresAt` null. The dashboard token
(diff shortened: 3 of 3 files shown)


ntdatt812:
The red shard is not from this diff.

`Unit Tests fast-path (1/4)` fails on one case, the repo-wide OpenAPI ratchet:

```
✖ openapi.yaml does not regress documented-operation coverage below the agreed floor
```

This PR adds no route and touches no spec — the diff is one string on a create-path allowlist in `src/lib/db/providers.ts`, one test, and a changelog fragment. **PR #11370 fails the identical case**, so the floor is simply above what the base currently achieves.

`Build (advisory)` and `dast-smoke` are likewise unrelated to a DB allowlist entry.

Green here, and these are the ones that would have caught a mistake in this change: `Unit Tests fast-path (2/4)`, `(3/4)`, `(4/4)`, `Vitest`, `Merge integrity`, `No new ESLint warnings`, `Docs Gates`, `Change Classification`, `semgrep`.

The new case is in shard 3 and passes there. Locally: `oauth-connection-tokenexpiresat-5326` 3 pass / 0 fail, and reverting the one string gives 2 pass / 1 fail.

diegosouzapw:

diegosouzapw:
Great catch and exemplary validation — the round-trip test plus your own mutation check is exactly how this should be done, and we independently confirmed the bug still exists on release/v3.8.51 and that no equivalent fix landed there. One structural blocker: your branch sits on top of release/v3.8.50 tip, so the PR currently carries ~19 unrelated v3.8.50-only commits toward v3.8.51. Please rebase your two commits onto the current release/v3.8.51 tip — we verified zero conflicts (both touched files are byte-identical between the two bases) and the diff will collapse to just the allowlist string, the test, and the changelog fragment.

ntdatt812:
Rebased onto the current `release/v3.8.51` tip (`b39e5ecb2`).

The branch had been cut before v3.8.51 was branched, so this PR was showing **21 commits / 131 files** — nineteen of them other people's already-merged work — and the merge had gone dirty. It is now **2 commits / 3 files**, and mergeable again.

The defect is still present on the new base. `optionalFields` in `createProviderConnection` reads

```ts
"accessToken",
"refreshToken",
"expiresAt",
"tokenType",
"scope",
"idToken",
```

with no `tokenExpiresAt`, so a created row stores NULL however good the payload is — while the update path already carries `data.tokenExpiresAt`.

**Re-verified after the rebase**, `tests/unit/oauth-connection-tokenexpiresat-5326.test.ts`:

```
✔ buildOAuthConnectionCreatePayload mirrors expiresAt into tokenExpiresAt (#5326)
✔ buildOAuthConnectionCreatePayload keeps tokenExpiresAt null when expiry is unknown
✔ a created connection keeps tokenExpiresAt through the database
pass 3, fail 0
```

**Mutation** — deleted the single `"tokenExpiresAt",` line from the allowlist (verified the file went from 1 occurrence to 0) and re-ran the same file:

```
✔ buildOAuthConnectionCreatePayload mirrors expiresAt into tokenExpiresAt (#5326)
✔ buildOAuthConnectionCreatePayload keeps tokenExpiresAt null when expiry is unknown
✖ a created connection keeps tokenExpiresAt through the database
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal
pass 2, fail 1
```

The two pre-existing assertions stay green under the mutation, and that is the whole reason this survived: they check the object `buildOAuthConnectionCreatePayload` returns, never the row that reaches SQLite. The third test persists through `createProviderConnection` and reads back through `getProviderConnections`, so it fails the moment the allowlist drops the field again.

Restored, and the diff is back to +52/-5 across the three files.

ntdatt812:
CI came back red on 9 checks after the rebase. All nine are the base, and the rebase is what pulled them in — the branch had been sitting on a pre-v3.8.51 cut where they did not exist yet.

**The failing set is identical on unrelated PRs cut from the same base.**

| PR | failing checks |
|---|---|
| #11470 | Build (advisory), Docs Gates (fast-path), Fast Quality Gates, No new ESLint warnings, Unit Tests fast-path (1–4/4), Vitest (fast-path) |
| #11471 | the same nine |
| this PR | the same nine |

**Unit Tests fast-path (1/4)** — the eight failing tests here are byte-for-byte the ones failing on #11470, which contains none of this diff:

```
✖ GET /.well-known/agent.json returns 6 skills
✖ Agent Card includes list-capabilities skill entry
✖ list-capabilities entry has required tags [discovery, capabilities]
✖ list-capabilities entry has at least one example question
✖ Agent Card includes all 5 original skills
✖ hard-lease credential, executor, and connection-query inventory has no unclassified site
✖ GOLDEN provider.ts translate-path is stable across all providers
✖ LEDGER-4: minimax-m3 vision metadata matches each provider
```

Agent-card skill counts, a hard-lease inventory sweep, a golden translate-path snapshot and a minimax-m3 metadata ledger. None of them reads `createProviderConnection`, and `oauth-connection-tokenexpiresat-5326.test.ts` does not appear anywhere in those logs.

**No new ESLint warnings** fails with

```
```

Stale entries in `config/quality/eslint-suppressions.json` — this diff adds lines, it does not remove the last occurrence of any suppressed warning. The same gate is red right now on #11473, #11471, #11470, #11469 and #11468.

**This PR's own tests pass in CI**, from the fast-path shard logs on this head:

```
✔ buildOAuthConnectionCreatePayload mirrors expiresAt into tokenExpiresAt (#5326)
✔ buildOAuthConnectionCreatePayload keeps tokenExpiresAt null when expiry is unknown
✔ a created connection keeps tokenExpiresAt through the database
```

The third is the new one, and it is the one that goes red the moment `"tokenExpiresAt"` leaves the allowlist.

I have deliberately not folded a base fix into this PR — the diff stays 3 files. Happy to send the suppressions prune separately if that is useful.

PR #11374, opened 2026-08-24
fix(cli): spawn npm the way Windows needs in `omniroute update`

### LINKED ISSUE #11335: [BUG] omniroute update fails on Windows — 'Could not check latest version' (execFile npm without shell in update.mjs, Node 24+ hardening, same class as #5379)

ntdat812:
Fixes #11335.

## Root cause

`bin/cli/commands/update.mjs` spawns npm twice with no shell:

```js
await execFn("npm", ["view", "omniroute", "version", "--prefer-online"], { timeout: 15000 });
await execFileAsync("npm", ["view", "omniroute", "changelog"], { … });
```

On Windows npm is `npm.cmd`, and Node ≥ 24 refuses to spawn a `.cmd` without a
shell (nodejs/node#52554). A bare `npm` can additionally resolve to an
extensionless shim that `CreateProcess` cannot execute. Either way the promise
rejects, `getLatestVersion()` swallows it in its `catch`, and the user sees

```
✖ Could not check latest version. Is npm available?
```

in a terminal where `npm view omniroute version` works fine.

This is the same class as #5379 / #5542, which fixed the server-side calls through
`buildNpmExecOptions`. The CLI was missed because it is plain `.mjs` and cannot
import that TypeScript helper.

## Change

`bin/cli/npm-exec.mjs` — the CLI's statement of the same rule: `npm.cmd` (or
`bun.exe`) plus `shell: true` and `windowsHide` on win32, no shell anywhere else.
Both npm call sites in `update.mjs` route through it.

**Hard Rule #13.** Enabling the shell means the shell splits the command line, so
no runtime value may reach argv. Both call sites pass a literal array
(`["view", "omniroute", "version", "--prefer-online"]`), so there is nothing to
splice — and a test asserts the arrays stay literal, failing if a future edit
interpolates a variable into one.

`execSync("npm install -g …")` further down is untouched: `execSync` already runs
through a shell, so it was never affected.

## Not in this PR

`bin/cli/commands/plugin.mjs` has the same `shell: false` spawn, but its argv
carries a user-supplied plugin name, so it needs a validated name pattern before
the shell can be enabled — a different argument, and a separate PR.

## Tests

`tests/unit/cli-update-npm-win32-11335.test.ts` (new, 5 cases): win32 resolves
`npm.cmd` and enables the shell; linux/darwin keep it off; unset options do not
become `undefined` keys; every npm call in `update.mjs` routes through the helper
and no bare `"npm"` literal remains; and the argv arrays contain no interpolation.

On the unfixed `update.mjs` the last two fail:

```
not ok 4 - #11335 every npm call in update.mjs routes through the helper
not ok 5 - #11335 the shell is only enabled where argv is literal (Hard Rule #13)
```

Changelog fragment: `changelog.d/fixes/cli-update-npm-win32-11335.md`.


ntdat812:

```
```

PR #11362, opened 2026-08-24
fix(video): harden result cache identity and coalescing

diegosouzapw:
## Summary

Completes the Video Bridge FU-01 cache-hardening slice on top of the live `release/v3.8.50` tip.

- fingerprints the authorized video bytes and all result-affecting dimensions before a persistent cache hit
- validates cached metadata strictly and deletes/recomputes corrupt entries
- bounds the TTL/LRU cache by entry count, per-entry bytes, and aggregate bytes
- coalesces protected HTTPS downloads and result production without persisting raw URLs or media
- isolates coalescing identity by tenant context and keeps waiter aborts independent from the producer
- bounds model selection and cache waits by the request deadline/abort signal
- separates persistent cache-hit telemetry from singleflight coalescing
- logs fail-open cache failures without leaking media or credential material

## TDD evidence

The regression suite was built RED→GREEN around:

- stable HTTPS URL serving changed bytes
- concurrent identical requests with cache enabled and disabled
- single protected download with independent extractions
- waiter abort, all-waiter abandonment, and fresh-flight replacement
- tenant isolation
- corrupt/unavailable cache fail-open behavior
- invalid numeric metadata and byte-length mismatches
- never-resolving model selection under timeout/abort
- per-entry and aggregate byte bounds
- coalesced telemetry not counted as persistent hits

## Validation

- focused Video Bridge matrix: **70/70 PASS**
- dedicated final result-cache suite: **33/33 PASS**
- `npm run typecheck:core`: **PASS**
- targeted ESLint: **PASS**
- `npm run check:cycles`: **PASS** (429 configured files)
- changelog integrity: **PASS**
- tracked-artifacts gate: **PASS**
- `git diff --check`: **PASS**
- independent read-only review: **CLEAN** (no blocker/P1/P2)

`typecheck:noimplicit:core` remains base-red only: 36 diagnostics are all in the untouched `open-sse/translator/response/openai-responses/pureHelpers.ts`; that blob is byte-identical between the PR base and head.

## Release disposition

This PR resolves FU-01 only. It does not claim completion of FU-02 through FU-09 and does not authorize merge, tagging, or release publication. Please keep it open for owner review.

PR #11333, opened 2026-08-24
feat(volcengine): Ark plan providers + phone/SMS console auto-login

rengaryang:
Adds first-class support for Volcano Engine Ark subscription plans (Coding Plan & Agent Plan), including a session-based headless phone+SMS login for the console so binding a plan no longer requires a manual browser on the server.

What

- Two new providers: volcengine-coding-plan and volcengine-agent-plan — registry entries, capability-seeded catalogs, and console API binding that detects plan availability, extracts the plan API key (ListApiKeys/GetRawApiKey), and upserts the connection with live quota usage.
- Live model discovery from the console: Coding Plan via ListArkCodeLatestModel, Agent Plan via ListAgentPlanLatestModel. Discovered models merge into the synced catalog; the static registry stays as an offline capability-seed fallback (deliberately not in CURATED_MODEL_ONLY_PROVIDERS).
- Phone + SMS auto-login (headless, session-based): POST /api/providers/volcengine-plan/connect {phone} opens a headless Playwright context, switches to the 手机号登录 tab, fills the phone, and requests the SMS code. Sub-routes /code, /status, /resend, /cancel, and /identity complete the flow. The service handles the console's MFA step-up modal (需要额外认证 — a second SMS code), the identity-selection page (/auth/login/select_identity, scraped with the real selectors from the auth bundle: ul[class*=accountUl] li[class*=accountLi] + button[type=submit]), and image captcha (screenshot returned to the dashboard for manual entry). Risk-control sliders or the TOTP-binding modal degrade to the existing headful manual flow instead of hanging.
- New VolcengineConnectModal on the provider page walks through phone → code → MFA → identity → bound, with resend countdown and per-step status.
- Security: every new route stays behind the LOCAL_ONLY + spawn-capable deny-lists (the connect pattern widened to connect(/.*)? to cover the session sub-routes); the phone number is kept in memory only.

Validation

- 32 new unit tests for the auto-login state machine (MFA step-up, identity selection, resend cooldown, bind dedupe + no-retry-on-failure, timeout diagnostics) — all pass.
- Regression: 73/73 across the volcengine suite, authz route guards, and token extraction configs. typecheck:core clean, open-sse typecheck gate OK (within baseline), ESLint clean.
- One test adaptation: the new "-web" naming lint in tokenExtractionConfig.test.ts exempts volcengine-console (it extracts a console session cookie, not a chat-web credential).

Generated with [OmniRoute](https://github.com/diegosouzapw/OmniRoute) 🚀

PR #11375, opened 2026-08-24
fix(providers): reject reserved provider prefixes on compatible-node create/update

ggdayup:
## Summary

Compatible provider nodes (custom OpenAI-compatible endpoints) currently accept a `prefix` that collides with a built-in registry id or alias (e.g. `tokenrouter`). The node saves fine, but every request shaped `<prefix>/<model>` is intercepted by the built-in provider lookup in `getModelInfo()`, so the node is silently unreachable — while the identical request against the upstream API directly succeeds. Users see `No active credentials for provider: tokenrouter` (`model_not_found`) and have no way to discover why.

This PR rejects such prefixes at write time with an explicit validation error, so the misconfiguration surfaces at creation instead of as a runtime routing dead-end.

## Root cause

- `src/sse/services/model.ts::getModelInfo()` intentionally skips compatible-node lookup when the model prefix matches a built-in REGISTRY id/alias (built-ins must win).
- The create/update schemas (`createProviderNodeSchema` / `updateProviderNodeSchema` in `src/shared/validation/schemas/provider.ts`) never mirrored that guard, allowing nodes whose prefix can never be routed to.

## Fix

- New shared module `src/shared/constants/reservedProviderPrefixes.ts` — single source of truth built from the live `REGISTRY` (all entry ids + aliases, 391 unique prefixes today). Exposes `getReservedProviderPrefixes()`, `isReservedProviderPrefix()` and a friendly `reservedProviderPrefixMessage()`.
- `createProviderNodeSchema`: `superRefine` now adds a Zod issue on `prefix` when it is reserved → route returns 400 with a per-field message.
- `updateProviderNodeSchema`: same refinement, so renaming a node's prefix to a reserved one is also rejected.
- `src/sse/services/model.ts`: the runtime guard now consumes the shared module (behavior byte-for-byte unchanged; removes the duplicated local set builder).

Semantics are **case-sensitive exact match**, mirroring the runtime guard: mixed-case values like `TokenRouter` do not intercept nodes at runtime and remain allowed; manual aliases outside REGISTRY (`xiaomi`, `llamacpp`, `aq`, …) also don't intercept and remain allowed.

## TDD evidence (RED → GREEN)

1. RED: wrote `tests/unit/provider-node-reserved-prefix.test.ts` against the unfixed schemas — reserved-prefix cases failed while non-reserved controls passed.
2. GREEN after fix: full suite passes.

```
node --import tsx/esm --test tests/unit/provider-node-reserved-prefix.test.ts
# pass 14 / fail 0   (shared-module semantics x5, schema-level x6,
#                     POST /api/provider-nodes 400 + happy path, PUT rename guard)
```

Regression battery after merging latest `main` (65e81158a):

```
tests/unit/{provider-node-reserved-prefix,model-resolver,chat-helpers,
            provider-node-icon-url,custom-headers-provider-nodes,
            local-aliases-precedence}.test.ts
# pass 111 / fail 0

npm run typecheck:core    # clean (exit 0)
npx eslint <4 changed files> --suppressions-location config/quality/eslint-suppressions.json  # exit 0
```

Runtime behavior unchanged (verified e2e on a live instance): a legacy node already saved with prefix `tokenrouter` still routes to the built-in provider via `getModelInfo("tokenrouter/...")`, and node-by-internal-id requests still reach the custom node — this PR only gates new writes.

## Notes for reviewers

- The count assertion pins 391 (measured at merge time). It exists to prove the set is a full REGISTRY walk rather than a hand-maintained list; it will move as providers are added — that's intentional.
- ⚠️ base-red inherited: main CI is failing on `main` head commits independent of this change (OpenSSF Scorecard / opencode-plugin CI / Build App / Publish to Docker Hub failures present before this branch), and issue #9985 documents release/v3.8.50 redness. None of the failing checks touch the files in this PR.
- Merge commit e014dffe2 resolves the conflict with 65e81158a by keeping both the upstream icon-url rewrite and this PR's shared-module import/guard.


diegosouzapw:
Landed via consolidated batch validation (direct push — your branch's history had diverged heavily from release/v3.8.50, carrying ~8 already-shipped commits alongside your own; cherry-picked just your 2 commits onto the current tip). One adjustment during landing: your import of `getReservedProviderPrefixes` conflicted with a sibling batch PR (#11362-family) that had added other new imports to the same block in `src/sse/services/model.ts` — combined both cleanly (kept all imports). Also bumped your own fixture pin from 391 to 395, since another PR in this same batch (#11333, volcengine) added 2 new providers that grow the REGISTRY-derived reserved-prefix count. Own regression suite (14/14) passes. Fixes the tokenrouter-shadowing bug you diagnosed. Thanks!

PR #11377, opened 2026-08-24
fix(live-ws): resolve the public socket URL at runtime

### LINKED ISSUE #11331: [BUG] Combo Studio frontend ignores proxy variables and forces hardcoded WebSocket port 20132 (`/live-ws`)
### OmniRoute Version

v3.8.49 (latest)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Alma Linux 10

### Node.js Version

_No response_

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

Web Interface

### Description

When deploying OmniRoute in a rootless container (Podman/Docker) behind a reverse proxy (Traefik) on port 443, the Combo Studio dashboard fails to connect to the live tracking socket. The UI constantly states: `Live disabled — WebSocket disconnected. Showing last known state.`

Inspecting the network traffic reveals that the frontend UI ignores container-side environment configuration variables (`LIVE_WS_PORT`) and blindly attempts to establish a client-side connection directly to `wss://omniroute.domain.tld:20132/live-ws`.

### Steps to Reproduce

1. Deploy OmniRoute inside a Podman container.
2. Proxy the container via Traefik routing standard traffic to the primary web interface port (`PORT=20128`).
3. Set environment parameters `LIVE_WS_HOST=0.0.0.0` and attempt to override the socket port.
4. Access the dashboard externally via HTTPS/443 and navigate to Combo Studio.
5. Inspect the Network Tab (WS filter).

### Expected Behavior

The frontend should recognize when it is being served via a reverse proxy/different port context and build its WebSocket handshake relative to the origin host/port (e.g., using a standard path prefix like `/live-ws` over the established web port), or respect a client-facing environment configuration variable.

### Actual Behavior

The browser attempts a connection directly to external port `20132`, completely bypassing the proxy topology and failing.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

It seems like src/hooks/useLiveDashboard.ts function getDefaultWsUrl() hardcodes literal WebSocket port 20132 on every browser-side path, ignoring the server's actual LIVE_WS_PORT config.

### Validation Plan

_No response_

ntdat812:
Fixes #11331.

## Root cause

The runtime-discovery path already exists and is there for exactly this deployment:
the browser reads `/api/v1/ws?handshake=1` because `NEXT_PUBLIC_*` is inlined at
**build** time, so a prebuilt Docker/npm image can never carry an operator's value.
`useLiveDashboard.ts` says so in its own comment.

But the server side of that handshake resolved the URL like this:

```ts
function getLivePublicUrl(): string | null {
  const publicUrl = process.env.NEXT_PUBLIC_LIVE_WS_PUBLIC_URL;   // build-time-shaped name
  ...
}
```

so in a prebuilt image it had **nothing to echo**. The client then fell through to
`getDefaultWsUrl()`, which hardcodes `:20132` — the reporter's
`wss://omniroute.domain.tld:20132/live-ws`, unreachable behind Traefik, and the UI
sits on *"Live disabled — WebSocket disconnected. Showing last known state."*

`LIVE_WS_HOST` / `LIVE_WS_PORT` are read at runtime (see `resolveEmbedWsHost`),
which is why setting them looked like it should have worked.

## Change

`resolveLiveWsPublicUrl(env)` in `src/shared/utils/wsPath.ts`, next to
`deriveLiveWsPath`:

- reads **`LIVE_WS_PUBLIC_URL`** at runtime — the un-prefixed name, consistent with
  `LIVE_WS_HOST` / `LIVE_WS_PORT`;
- keeps `NEXT_PUBLIC_LIVE_WS_PUBLIC_URL` as the fallback, so any deployment already
  setting it is unaffected;
- accepts only `ws://` / `wss://`, the same guard `sanitizeWsPublicUrl` applies on
  the client, and trims surrounding whitespace;
- `getLiveWsPath()` derives from the resolved value, so a proxy path like
  `/omniroute/live` reaches the client through the handshake too.

The handshake route now calls it. **The hardcoded `:20132` default is untouched** —
it is still correct when no public URL is declared, and changing it to same-origin
would need an upgrade handler on the HTTP port that `src/proxy.ts` does not have
(the comment in `getDefaultWsUrl` records why that was tried and reverted).

## Tests

`tests/unit/live-ws-public-url-11331.test.ts` (new, 7 cases): the runtime name
honoured; the prefixed name still working and losing to the runtime one; non-ws
schemes rejected (`https://`, `javascript:`, bare `host:443`); blank/whitespace
falling through to the next candidate; surrounding whitespace trimmed rather than
rejected; the path following the resolved URL; and a source assertion that the
handshake route no longer reads the build-time-only name directly.

Mutation check — restoring the route's old body fails exactly:

```
not ok 7 - #11331 the handshake route resolves the URL at runtime
```

The three existing live-ws suites (`instrumentation-live-ws`, `live-ws-cookie-parse`,
`live-ws-eaddrinuse-6324`) stay green. `npx tsc --noEmit` clean; eslint clean with
the repo's suppressions file.

Changelog fragment: `changelog.d/fixes/live-ws-public-url-runtime.md`.


ntdat812:

```
```

PR #11347, opened 2026-08-24
fix(translator): prevent empty Codex tool_use blocks

NicolasDuranGarces:
## Summary

- Preserve Claude tool-call lifecycle fields when the Responses translator records the completed call summary.
- Prevent Codex Responses streams from emitting a second `content_block_start` with the same ID and an empty tool name.
- Add an end-to-end SSE regression covering the incremental Codex function-call lifecycle and the completed snapshot echo.

## Related Issues

- Related to #527

## Validation

- [x] Change type: provider / translator
- [x] Focused translator tests: 46 passed, 0 failed
- [x] Focused ESLint and Prettier checks passed
- [x] `npm run typecheck:core`
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new automated test in this PR
- [ ] `npm run lint` (full repository lint is left to CI; both changed files pass ESLint)

TDD evidence: the new regression failed before the fix with two Claude `tool_use` starts (`actual: 2`, expected `1`) and passed after preserving the existing shared-map entry.

## Tests Added Or Updated

- `tests/unit/codex-claude-empty-tool-use.test.ts`

## Coverage Notes

The new test exercises the complete Responses-to-Claude SSE pipeline through `createSSETransformStreamWithLogger`, including `response.output_item.added`, argument events, `response.output_item.done`, and the echoed `response.completed` snapshot.

## Reviewer Notes

The change is intentionally limited to merging the completed-call summary into an existing shared `state.toolCalls` entry. Existing translator-specific lifecycle fields remain intact; summary fields used by completion logging are still updated.

PR #11385, opened 2026-08-24
fix(opencode): close Muse Responses streams at completion

### LINKED ISSUE #11384: fix(providers): terminate OpenCode Responses stream after response.completed
## Summary

The OpenCode Zen Responses stream can deliver the Muse Spark answer, but the HTTP/SSE stream does not terminate afterward. The upstream sends `response.completed` and then keeps the connection alive with ping/keepalive events. OmniRoute continues emitting empty keepalive chat-completion chunks, so clients wait indefinitely.

## Reproduction

Use an OpenCode connection with model `oc/muse-spark-1.2-contributor-free`, request `stream: true`, and send a short prompt such as `Reply with exactly OK.`.

## Actual behavior

The answer is returned, followed by repeated empty keepalive chunks. The stream does not finish.

## Expected behavior

Once the upstream Responses stream emits `response.completed`, OmniRoute should finish the OpenAI-compatible stream and close the connection.

## Scope

This appears specific to the OpenCode Responses path. The fix is scoped to recognizing the upstream terminal event and cancelling the upstream reader after completion.

AStupidBear:
## Summary

Fix OpenCode Muse Spark Responses streams that return the answer but remain open and emit keepalive chunks indefinitely.

## Changes

- Treat the upstream `response.completed` event as terminal for the OpenAI-compatible stream.
- Cancel the upstream reader and close the downstream stream immediately after completion.
- Keep the UUID session fingerprint workaround scoped to Muse Responses requests.
- Preserve direct-fetch behavior for the single-account, no-proxy path.
- Add regression coverage for stream completion and session headers.

## Validation

- Focused OpenCode/proxy/executor tests: 88 passed.
- `npm run typecheck:core`: passed.
- Live OpenCode Muse Spark stream: answer returned, `finish_reason=stop`, one terminal `[DONE]`, no keepalive after completion.

Closes #11384

### DIFF
--- a/open-sse/executors/opencode.ts
+++ b/open-sse/executors/opencode.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from "node:crypto";
 import {
   BaseExecutor,
   type ExecuteInput,
@@ -10,7 +11,7 @@ import {
   injectReasoningContentForThinkingModel,
   isThinkingMessageModel,
 } from "../utils/reasoningContentInjector.ts";
-import { runWithProxyContext } from "../utils/proxyFetch.ts";
+import { runWithDirectFetchContext, runWithProxyContext } from "../utils/proxyFetch.ts";
 import { forwardOpencodeClientHeaders } from "../utils/opencodeHeaders.ts";
 import {
   type AccountProxyConfig,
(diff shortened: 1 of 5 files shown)


diegosouzapw:
Landed via consolidated batch validation (direct push — same release/v3.8.51-targeting situation as #11347, retargeted onto the actively-developed release/v3.8.50 tip). Fixes OpenCode Muse Spark Responses streams that stayed open emitting keepalives after completion. Thanks!

PR #11382, opened 2026-08-24
fix(video): harden visual frame deduplication

diegosouzapw:
## Summary

- move the Video Bridge frame cap to the post-dedup result, so near-duplicate candidates can be replaced before the final 1-16 frame budget is applied
- bound the perceptual candidate pool to at most twice the requested budget and never above 16 frames
- include the dedup policy/version in the Video Bridge result-cache identity and metadata
- add cooperative abort checks to the perceptual comparison loop so a disconnected request does not keep burning CPU
- add deterministic fixtures, regression coverage, cache-version coverage, and a bounded microbenchmark

## Stack dependency

This PR is intentionally stacked on [#11362](https://github.com/diegosouzapw/OmniRoute/pull/11362) at exact parent `d4ade9d1d3921b0425cb0263f47a0471eee8dd93`. It should be rebased against the final #11362 head before any eventual merge. This PR remains draft and does not request a merge.

## TDD and validation

- focused affected matrix: **129/129 PASS**
- real abort regression: residual work about **77 ms**, versus about **5.055 s** before the fix
- bounded benchmark: 150 comparisons; CPU **68.750 ms/op**, RSS **+0.75 MiB**, heap **+6.79 MiB**
- `npm run typecheck:core` — PASS
- targeted ESLint, script lint, Prettier, cycles, docs-all, changelog integrity, and diff-check — PASS
- independent final review — PASS, no findings
- `typecheck:noimplicit:core` — BASE-RED only in unchanged `open-sse/translator/response/openai-responses/pureHelpers.ts`

## Evidence boundary

The benchmark proves the bounded computation and resource envelope for these fixtures. It is not a semantic model-quality evaluation. Full unit, full Vitest, coverage, and production build were not run for this isolated stacked slice.

The PR stays OPEN/DRAFT; no auto-merge, tag, release, or deployment is requested.

Issue #8522, opened 2026-07-25
fix(quality): absolute baselines make inherited drift red every PR — red stops being a signal


backryun:
## Correction first

There is a premise embedded in `config/quality/file-size-baseline.json` — and repeated by me on #8490 — that is **factually wrong**:

> `_rebaseline_2026_06_29_v3841_release`: "(the PR->release fast-gates do NOT run check:file-size)"

`check:file-size` **does** run on every PR→`release/**`. It is a step in the `fast-gates` job of `.github/workflows/quality.yml`, alongside `check:complexity-ratchets`, and `lint-guard` runs `lint:json --max-warnings 0` in the same workflow. I repeated that stale note in my #8490 comment and commit message as the explanation for why drift accrues. It isn't the explanation, and I'd rather correct it than leave it propagating into the next rebaseline note.

The drift is real. The mechanism is different — and worse.

## The actual mechanism

These gates evaluate **absolute** metrics against a frozen file, not the PR's own delta. `check-file-size.mjs` reads `file-size-baseline.json` and compares current LOC to the frozen number; there is no base-ref comparison anywhere in it. The suppression counts in `eslint-suppressions.json` work the same way.

That produces a self-sustaining loop:

1. A PR grows a frozen file (or removes a suppressed `any`) without bumping its entry.
2. Its own gate goes red — **but it merges anyway**.
3. Now the *base* is over the cap, so **every subsequent PR is red on that gate**, regardless of content.
4. Red stops distinguishing "you broke it" from "you exist", so step 2 keeps happening.
5. Someone eventually files a base-red repair PR. It fixes the entries as measured that hour. New drift lands the same day. Back to 3.

## Traced end-to-end on today's batch

`release/v3.8.49` moved 25 commits, `1cafd328c` → `7a8f9156d`.

**#8349** ("Clicking a provider card hitting back loses scroll position") grew `src/app/(dashboard)/dashboard/providers/page.tsx` from **1918 → 1990** against a frozen cap of **1927**, without bumping the entry.

Its own run was red on exactly the gate that caught it:

```
#8349  merged 2026-07-25T05:51:00Z
       FAILURE: Fast Quality Gates
       FAILURE: No new ESLint warnings
       FAILURE: Unit Tests fast-path (2/4), (4/4)
```

The gate worked. It was merged past.

[...]

PR #11388, opened 2026-08-24
fix(dashboard): honour the live WebSocket port the handshake reports (#11331)

### LINKED ISSUE #11331: [BUG] Combo Studio frontend ignores proxy variables and forces hardcoded WebSocket port 20132 (`/live-ws`)

ntdatt812:
⚠️ base-red inherited: #9985

Fixes #11331 (the part that is a bug).

## The handshake already reports the port; the client throws it away

`/api/v1/ws?handshake=1` returns the live server's real listening port:

```ts
live: {
  port: parseInt(process.env.LIVE_WS_PORT || "20132", 10),
  publicUrl: getLivePublicUrl(),
  path: getLiveWsPath(),
  ...
}
```

`useLiveDashboard` reads `publicUrl` and `path` out of that response — and nothing else:

```ts
const publicUrl = sanitizeWsPublicUrl(body?.live?.publicUrl);
if (publicUrl) setHandshakeUrl(publicUrl);
if (typeof body?.live?.path === "string" && body.live.path.startsWith("/")) setHandshakePath(body.live.path);
```

so the URL keeps the port compiled into the bundle:

```ts
return `${protocol}//${hostname}:20132${BUILD_TIME_WS_PATH}`;
```

`NEXT_PUBLIC_*` is inlined at build time — which is the whole reason this handshake exists, as the comment right above it says. A prebuilt Docker or npm image has no other way to learn a runtime value. So setting `LIVE_WS_PORT` moves the server and leaves the dashboard dialling 20132, with no way to correct it short of rebuilding the image. That is the `Live disabled — WebSocket disconnected` banner in the report.

## The change

Read `live.port` from the handshake too, and resolve the URL in one place (`shared/utils/wsPath.ts`):

1. an explicit `wsUrl` from the caller
2. a complete `live.publicUrl` from the handshake
3. the default URL with whatever port and path the handshake reported

A port outside 1..65535 is ignored rather than producing an unconnectable URL.

Also collapsed the `if/else` in `getDefaultWsUrl` whose two branches returned the same expression — it read as a distinction that no longer exists.

## Scope — what this does not do

The report also asks the frontend to infer the proxy topology and connect over the origin port with a path prefix. That is already supported through `LIVE_WS_PUBLIC_URL`, which the handshake surfaces as `live.publicUrl` and which this PR keeps as the highest-priority source. Guessing a same-origin upgrade instead would break the documented setup, since `src/proxy.ts` has no upgrade handler on the HTTP port — that is the reason for the "Bug #1 fix" comment this PR preserved. So the fix here is the unambiguous half: a configured port must not be ignored.

## Tests

`tests/unit/live-ws-url-11331.test.ts` — 11 cases:

- a reported port replaces the compiled-in one, host and path untouched
- no handshake data → default unchanged
- `0`, `70000`, `1.5`, `""`, `"abc"`, `null`, `NaN`, `{}` → ignored
- port and path applied together
- a path that is not a path is ignored
- `publicUrl` wins over the reported port; an explicit `wsUrl` wins over everything
- an unparseable default is returned as-is instead of throwing
- `deriveLiveWsPath` behaviour unchanged

Mutation-checked — dropping the `url.port` assignment turns 2 red:

```
✖ uses the port the handshake reports instead of the compiled-in one
✖ applies the port and the path together
ℹ pass 9  ℹ fail 2
```

`tsc --noEmit` reports nothing in the touched files; prettier/eslint pre-commit hooks clean.

### DIFF
--- a/changelog.d/fixes/11388-live-ws-handshake-port.md
+++ b/changelog.d/fixes/11388-live-ws-handshake-port.md
@@ -1,0 +1,1 @@
+- **Live dashboard:** honour the WebSocket port reported by `/api/v1/ws?handshake=1` instead of the port compiled into the bundle, so a `LIVE_WS_PORT` override reaches prebuilt Docker/npm images and Combo Studio Live connects behind a reverse proxy ([#11331](https://github.com/diegosouzapw/OmniRoute/issues/11331)).
--- a/src/hooks/useLiveDashboard.ts
+++ b/src/hooks/useLiveDashboard.ts
@@ -13,7 +13,7 @@
 
 import { useEffect, useRef, useState, useCallback } from "react";
 import type { DashboardChannel, DashboardEventName } from "@/lib/events/types";
-import { deriveLiveWsPath } from "@/shared/utils/wsPath";
+import { deriveLiveWsPath, resolveLiveWsUrl, sanitizeLiveWsPort } from "@/shared/utils/wsPath";
 
 // ── Config ────────────────────────────────────────────────────────────────
 
@@ -40,14 +40,10 @@ function getDefaultWsUrl(): string {
   if (typeof window === "undefined") return `ws://localhost:20132${BUILD_TIME_WS_PATH}`;
   const protocol = window.location.protocol === "https:" ? "wss:" : "ws:";
   const { hostname } = window.location;
-  // Bug #1 fix: Use the WS server's actual port (20132) for both loopback
-  // and non-loopback clients. Previously the non-loopback branch tried to
-  // upgrade the HTTP port (window.location.host) which has no upgrade
-  // handler in src/proxy.ts. If the user wants the upgrade to go through
-  // Next.js (same-origin), they should explicitly pass `wsUrl`.
-  if (hostname === "localhost" || hostname === "127.0.0.1" || hostname === "::1") {
-    return `${protocol}//${hostname}:20132${BUILD_TIME_WS_PATH}`;
-  }
+  // The WS server's own port, for loopback and non-loopback alike: the HTTP
+  // port has no upgrade handler in src/proxy.ts. This is only the starting
+  // point - the handshake below replaces the port when the server reports a
+  // different one, and a caller can always pass `wsUrl` outright.
   return `${protocol}//${hostname}:20132${BUILD_TIME_WS_PATH}`;
 }
 
@@ -113,6 +109,7 @@ export function useLiveDashboard({
   const needsHandshake = !wsUrl && !BUILD_TIME_PUBLIC_WS_URL && typeof window !== "undefined";
   const [handshakeUrl, setHandshakeUrl] = useState<string | null>(null);
   const [handshakePath, setHandshakePath] = useState<string | null>(null);
+  const [handshakePort, setHandshakePort] = useState<number | null>(null);
   const [wsUrlResolved, setWsUrlResolved] = useState(!needsHandshake);
 
   useEffect(() => {
@@ -127,6 +124,11 @@ export function useLiveDashboard({
         if (typeof body?.live?.path === "string" && body.live.path.startsWith("/")) {
           setHandshakePath(body.live.path);
         }
+        // The live server reports the port it is actually listening on, so a
+        // LIVE_WS_PORT override reaches a prebuilt image instead of being
+        // overruled by the compiled-in default (#11331).
+        const port = sanitizeLiveWsPort(body?.live?.port);
+        if (port !== null) setHandshakePort(port);
       })
       .catch(() => {
         // Handshake unavailable — fall back to the default URL.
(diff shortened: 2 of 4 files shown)


Selaygermany:
Looked into the 3 failing checks on this PR (as of commit 913a51e):

- **`Fast Quality Gates`** fails with `1 gate(s) failed: file-size`. This matches the base-red drift already tracked in #9985 on `release/v3.8.50` — this diff only touches 4 small files (+174/-23), so the gate failure doesn't look introduced by this change.
- **`Build (advisory)`** — non-blocking by design (name says so).
- **`dast-smoke`** — cancelled after ~18min on the "Build CLI bundle" step ("The operation was canceled"), which looks like CI infra flakiness rather than a code issue.

Everything else is green: all 4 unit test shards, Vitest fast-path, "No new ESLint warnings", Docs Gates, semgrep x2, merge integrity, change classification.

The fix itself (`sanitizeLiveWsPort` + `resolveLiveWsUrl` in `src/shared/utils/wsPath.ts`) is well-scoped, validates the handshake-reported port (1-65535, type-checked) instead of trusting it blindly, and ships with 11 passing unit tests covering the sanitize/resolve/edge-case paths.

Per the base-red convention in `AGENTS.md`, might be worth noting `⚠️ base-red inherited: #9985` in the PR description so reviewers don't chase these 3 as regressions caused by this change.

ntdatt812:
Thanks — added `⚠️ base-red inherited: #9985` to the body per the AGENTS.md §Base-green convention.

I ran the gate locally to pin down which file trips it, because the mechanism is slightly different from inherited drift. `file-size` is already base-relative on PR events (#8522, `--base-ref $PR_BASE_SHA`), so ordinary inherited growth cannot red an innocent PR. What fails here is the **test-file** sub-check, which is absolute for files that are new relative to the frozen baseline:

```
$ npm run check:file-size -- --base-ref 7715825cb8cd170f5371c4c4bd5f940b5608f480
[file-size] OK — 161 arquivos congelados, cap 1000 para novos (4380 arquivos verificados)
[test-file-size] 1 test file violation(s) (testCap 1000):
  ✗ tests/unit/guardrails/videoBridgeResultCache.test.ts: 1036 > cap 1000 (arquivo novo acima do limite)
```

That file is not in this diff — it arrives with the base tip itself:

```
$ git log --oneline -1 -- tests/unit/guardrails/videoBridgeResultCache.test.ts
7715825cb fix(video): harden visual frame deduplication (#11382)

$ git show origin/release/v3.8.50:tests/unit/guardrails/videoBridgeResultCache.test.ts | wc -l
1039
```

`7715825cb` is the merge-base of this branch, so every PR cut from the current `release/v3.8.50` tip will carry this failure until that test file is split or given a `testFrozen` entry. Worth flagging on #9985 / to @diegosouzapw, since the base-relative exemption does not cover it.

The other two are cancellations rather than failures — confirmed from the job API:

```
dast-smoke        conclusion=failure  →  step "Build CLI bundle" = cancelled
Build (advisory)  conclusion=failure  →  step "Run npm run build" = cancelled
```

Everything this PR actually changes is green.

ntdatt812:
The three red lanes here are not this diff.

**`Fast Quality Gates` → `file-size`** is the only real failure, and it fails on the base branch by itself:

```
$ git checkout origin/release/v3.8.50   # clean, zero changes from this PR
$ node scripts/check/check-file-size.mjs --base-ref c83116e634a054b6ad50d54eb49ac7449615fb2e
[file-size] 1 violação(ões):
  ✗ src/lib/guardrails/videoBridgeRuntime.ts: 1009 > cap 1000 (arquivo novo acima do limite)
[test-file-size] 1 test file violation(s) (testCap 1000):
  ✗ tests/unit/guardrails/videoBridgeResultCache.test.ts: 1040 > cap 1000
EXIT=1
```

`tests/unit/guardrails/videoBridgeResultCache.test.ts` is 1035 lines at this PR's base sha `c83116e634a0` and 1039 on the current `release/v3.8.50` tip — it was over the cap before this branch existed, and neither it nor `videoBridgeRuntime.ts` is in `file-size-baseline.json`, so both are measured against the 1000 cap for un-frozen files. It arrived with #11362 / #11382 (`fix(video): harden result cache identity and coalescing`, `harden visual frame deduplication`).

This PR touches four files — `src/hooks/useLiveDashboard.ts`, `src/shared/utils/wsPath.ts`, `tests/unit/live-ws-url-11331.test.ts`, `changelog.d/fixes/11388-live-ws-handshake-port.md` — none of them near the cap and none of them the flagged files.

**`Build (advisory)`** and **`dast-smoke`** did not fail on their own merits — both logs end at `##[error]The operation was canceled.`, cancelled after the gates job went red.

Happy to rebase once the two guardrails files are split or baselined, if you'd rather see the lane green before merging.

Issue #10514, opened 2026-08-15
[feature] Exclusive session connection leasing for long-lived client sessions


diegosouzapw:
## Feature Request

**Source:** Discussion #10378 by @KaspaPulse

## Problem
Long-lived client sessions (e.g. Claude Code, Cline, Roo) currently face potential thundering herd or token cache misses during multi-turn tasks because request dispatch can alternate between multiple eligible provider connections across turns.

## Proposed Solution
Provide an opt-in session affinity mechanism (`Exclusive Managed Session Leasing`) ensuring:
`ONE ACTIVE MANAGED CLIENT/SESSION <-> ONE EXCLUSIVE ELIGIBLE OMNIROUTE CONNECTION`
The lease persists connection affinity across idle periods and tool-use gaps.

### Implementation Ideas
- Wire lease resolution into `open-sse/services/auth.ts` and `open-sse/services/combo.ts`
- Track active lease IDs in memory with TTL expiration

### Current Workarounds
Relying on standard round-robin or priority routing without session pin.

## Additional Context
- Discussion: #10378

KaspaPulse:
Thanks for tracking this and for carrying the proposal into the feature pipeline.

For reference, PR #10362 already implements this primitive and includes the SQLite-backed
hard ownership model you reviewed.

One implementation detail worth preserving for the stated hard-exclusivity invariant:
an in-memory lease-ID map with TTL alone would not provide durable/global ownership across
processes or service restarts.

The current PR uses SQLite partial unique indexes for ACTIVE owner and connection uniqueness,
atomic acquisition, exact generation fencing, and TTL/reconciliation for crash recovery.

It may be useful to link #10362 here as the current reference implementation so we avoid
creating a parallel lease authority.

diegosouzapw:
Thanks for the suggestion! It has been analyzed and cataloged in our internal implementation backlog, and will be implemented soon. We're closing the issue to keep the tracker lean — the implementing PR will reference this issue when it ships.

PR #10362, opened 2026-08-14
feat(routing): add exclusive managed session connection leases

KaspaPulse:
## Summary

- Adds opt-in Exclusive Managed Session Leasing for API keys with `lease:exclusive` and an
  explicit non-empty `allowedConnections`.
- Gives one active managed client/session exclusive ownership of one eligible OmniRoute
  connection for its lease lifecycle using SQLite-backed atomic ownership and exact generation
  fencing.
- Reuses OmniRoute's existing eligibility, `allowedConnections`, quota/health/cooldown logic,
  routing strategy, retry/fallback, and session affinity.
- Returns `429 LEASE_CAPACITY_UNAVAILABLE` with `WAITING_FOR_CAPACITY` and `Retry-After` only
  when ordinary eligible candidates exist but every one is occupied by a foreign active lease.
- Core is client-neutral, provider-neutral, and model-neutral. No client-specific launcher
  integration is included.

## Why

Session affinity provides soft locality and continuity, but not exclusive lifetime ownership.
OAuth occupancy and the account semaphore address different request- or process-local concerns.
Long-lived managed clients sometimes need hard one-session/one-connection ownership across idle,
tool, and build periods. This change adds that opt-in authority without creating a second router
or account pool.

## Contract

- The API key must have the `lease:exclusive` scope and an explicit non-empty
  `allowedConnections` list.
- `POST /api/v1/session-leases` provides the acquire, renew, and release lifecycle endpoint.
- Managed traffic supplies `X-OmniRoute-Lease-Owner` and
  `X-OmniRoute-Lease-Generation`.
- The raw owner is hashed before persistence.
- The final fence binds the owner hash, exact generation, connection, and authenticated API-key
  ID.
- Global `ACTIVE` uniqueness applies to both owner and connection.
- The default lease TTL is 120 seconds.
- Acquire, renew, and release use exact ownership and generation semantics.
- `429 LEASE_CAPACITY_UNAVAILABLE`, `WAITING_FOR_CAPACITY`, and `Retry-After` indicate that
  ordinary eligible candidates exist but all are occupied by foreign active leases.
- Unmanaged behavior is unchanged.

## Related Work

Architecture precedents include #7650 for generic session affinity across providers, #8940 for
active OAuth-session occupancy and account availability, and #10149 for account-scoped
concurrency. This PR is a distinct opt-in hard lifecycle-ownership primitive.

## Validation

- [x] Change type: routing / DB
- [x] Lease foundation: 83/83.
[...]

PR #11389, opened 2026-08-24
feat(dashboard): surface exclusive managed sessions in Sessions view

KaspaPulse:
## Summary

Keep every currently held Exclusive Managed Session Lease visible in the existing Sessions tab for the full valid lease lifetime, including idle gaps between model turns, tools, and local builds.

The current tab is backed by `sessionManager`, whose in-memory fingerprints are request-derived and swept after inactivity. That is useful for legacy sessions, but it cannot represent an idle exclusive lease whose official SQLite lifetime is still valid. This PR therefore adds a thin observability projection from the existing lease authority instead of changing lease ownership or routing.

## Behavior and compatibility

- Held leases remain visible while idle.
- A held lease gets the existing localized `common.active` indication only when existing pending-request accounting reports in-flight work for that connection.
- Idle leases remain visible without a new `IDLE` badge.
- Release or TTL expiry removes the lease-backed row.
- Legacy and unmanaged session rows remain visible.
- A request-derived row is suppressed when its connection is already represented by a held lease.
- Existing `/api/sessions` fields (`count`, `sessions`, and `byApiKey`) retain their prior meaning. `exclusiveSessions` is additive, and the UI treats an absent field as an empty array.

## Architecture and safety

The implementation reuses:

- `getExclusiveLeaseConnectionIds()` for the API-key managed-connection policy;
- `getExclusiveLeaseOccupancy()` for authoritative current SQLite occupancy;
- existing pending-request accounting only as the in-flight-work signal;
- the existing `/api/sessions` route, `SessionsTab`, `getAccountDisplayName()`, localized labels, and dashboard styles.

It adds no session/lease store, table, migration, router, pool, daemon, waiter, quota/health/cooldown state, or client-specific mapping. Request tracking does not create, renew, release, invalidate, or otherwise mutate a lease.

The dashboard projection contains no raw lease-owner header, owner hash, API-key ID, generation/fencing material, credential, or token. Provider display metadata is read through a credential-free DB helper and formatted by the existing display-name resolver.

## Prior-art / duplicate search

Open and closed issues and PRs, discussions, commit search, current code, and recent relevant path history were searched for:

- `exclusive lease dashboard`
- `exclusive managed session UI`
- `session lease observability`
- `active sessions lease`
- `SessionsTab exclusive lease`
- `idle leased session`
- `managed session dashboard`

The only relevant results were this PR, merged backend foundation #10362, issue #10514, and discussion #10378. No duplicate or newer upstream dashboard solution was found. Maintainer discussion on #10362 explicitly identified dashboard UI as a possible follow-up.

## Files changed

- `src/app/api/sessions/route.ts`
- `src/app/(dashboard)/dashboard/usage/components/SessionsTab.tsx`
- `src/lib/sessionObservability.ts`
- `src/lib/db/providers.ts`
- `tests/unit/exclusive-session-observability.test.ts`
- `tests/unit/ui/exclusive-session-observability-ui.test.tsx`
- `changelog.d/features/11389-exclusive-managed-session-dashboard.md`

## Validation on Ubuntu / Node 24.18.1

- Projection, official SQLite lifecycle, API compatibility, sanitization, and privacy: `5 passed, 0 failed`.
- Actual jsdom/Vitest `SessionsTab` rendering: `3 passed, 0 failed`.
- Existing exclusive-lease, managed-key, session-manager/sweep, pending-request, request-logger, and management-auth regressions: `170 passed, 0 failed`.
- `git diff --check`: pass.
- Prettier on all touched files: pass.
- ESLint on all touched TypeScript/TSX files: pass.
- `typecheck:core`: pass.
- Dashboard typecheck ratchet: pass (220 frozen pre-existing diagnostics; no regression).
- Changelog integrity, cycles, DB rules, error-helper, complexity, cognitive-complexity, file-size, build-scope, test-discovery, tracked-artifact, and commit-hook gates: pass.
- Current supported production build (`npm run build`, Next.js 16.3.1 Turbopack): pass. Its warnings point to untouched paths.
- Isolated runtime smoke on `127.0.0.1:32189` and a temporary `DATA_DIR`: official HTTP acquire made an idle lease visible in `/api/sessions`; renew kept it visible; release and real TTL expiry removed it. The final traced run recorded no outbound IP connections.

External provider/model dispatches: **0**. No `/v1/responses`, `/v1/chat/completions`, or other model execution endpoint was called.

## Remaining limitations / base status

- Validation was focused to the touched and authoritative lease/session paths rather than the entire repository test matrix.
- The optional `gitleaks` binary was unavailable, so the repository's secret gate reported its designed non-blocking skip; the tracked-artifact and source privacy assertions passed.
- The report-only forgotten-sibling calibration gate passed with broad advisory findings caused by changing the widely imported provider DB module; it reported no masking/deletion risk.
- No open `release/v3.8.51` base-red marker was found, and the supported production build passed at base SHA `3192eb88d5550de4c3fd9985564f6b5641e9d681`.

Refs #10514

Follow-up to #10362

### DIFF
--- a/changelog.d/features/11389-exclusive-managed-session-dashboard.md
+++ b/changelog.d/features/11389-exclusive-managed-session-dashboard.md
@@ -1,0 +1,1 @@
+- **feat(dashboard):** surface durable exclusive managed leases in the existing Sessions view, keeping leased clients visible across idle gaps while marking connections with in-flight work as active ([#11389](https://github.com/diegosouzapw/OmniRoute/pull/11389)) — thanks @KaspaPulse
--- a/src/app/(dashboard)/dashboard/usage/components/SessionsTab.tsx
+++ b/src/app/(dashboard)/dashboard/usage/components/SessionsTab.tsx
@@ -2,19 +2,47 @@
 
 import { useTranslations } from "next-intl";
 
-import { useState, useEffect, useCallback } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
+import {
+  mergeDashboardSessions,
+  type DashboardSession,
+  type ExclusiveDashboardSession,
+  type RecentSessionForDashboard,
+} from "@/lib/sessionObservability";
 import { Card } from "@/shared/components";
+
+type SessionsResponse = {
+  sessions: RecentSessionForDashboard[];
+  exclusiveSessions: ExclusiveDashboardSession[];
+};
+
+const EMPTY_DATA: SessionsResponse = {
+  sessions: [],
+  exclusiveSessions: [],
+};
+
+function isLeaseBackedSession(session: DashboardSession): session is ExclusiveDashboardSession {
+  return "leaseBacked" in session && session.leaseBacked;
+}
 
 export default function SessionsTab() {
   const t = useTranslations("usage");
-  const [data, setData] = useState({ count: 0, sessions: [] });
+  const tCommon = useTranslations("common");
+  const [data, setData] = useState<SessionsResponse>(EMPTY_DATA);
   const [loading, setLoading] = useState(true);
 
   const loadSessions = useCallback(async () => {
     try {
       const res = await fetch("/api/sessions");
-      if (res.ok) setData(await res.json());
+      if (res.ok) {
+        const next = await res.json();
+        setData({
+          sessions: Array.isArray(next.sessions) ? next.sessions : [],
+          exclusiveSessions: Array.isArray(next.exclusiveSessions) ? next.exclusiveSessions : [],
+        });
+      }
     } catch {
+      // A failed background poll leaves the last successful Sessions snapshot visible.
     } finally {
       setLoading(false);
     }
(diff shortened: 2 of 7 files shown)


diegosouzapw:
This is a model follow-up PR — the projection layer cleanly separates lease authority from observability, the DB helper stays inside src/lib/db/providers.ts with parameterized non-credential SQL, and the privacy assertions (no owner hash/generation/apiKeyId in the payload) are baked into the tests rather than promised in prose. We reproduced your results on the current release/v3.8.51 tip: the node:test suite passes 5/5 including the real SQLite acquire/renew/expiry lifecycle, and the jsdom SessionsTab render passes 3/3 under the vitest UI config. Two notes for the record: (1) the header badge now shows merged row count instead of the raw /api/sessions count — intentional per your description, but worth one line in the release notes; (2) getProviderConnectionDisplayMetadata interpolates one bind parameter per leased connection — fine at any realistic lease count, just keep it in mind if managed-connection sets ever grow into the thousands. Nothing blocking from our side.

Issue #5923, opened 2026-07-02
fix(resilience): subscription 429s on priority combos evade cooldown (no Retry-After floor, backoff not accumulating, quota preflight is auto-routing-only, is_exhausted never set)


andrea-kingautomation:
## Environment
- omniroute 3.8.42 (current), self-hosted, ~24 combos.
- OAuth subscription providers used as **priority-strategy** combo lead legs (borrowed-account Claude gateways, providers `antigravity` and `agy`).

## Symptom
On a `priority` combo, the two lead OAuth sonnet legs 429 ~100% of the time during throttled windows and each 429 takes ~50s (the upstream holds the connection, then rejects). Fallback **does** work (the chain reaches healthy gemini/nemotron legs and returns 200), so this is not a routing dead-end. The problem is cost: every request in a throttled window burns ~100s of dead wait walking two doomed OAuth legs before reaching a model that answers, and a fraction exhaust the combo `targetTimeoutMs` (120s) and end as 524.

Today on one such combo: 1,312 lead-leg attempts, all 429 (~50s each); 556 eventual 200s on lower legs; 55 x 524 timeouts.

## Findings (from live telemetry)

**1. connectionCooldown does not fire for 429s that carry no Retry-After.** Most OAuth connections correctly show a future `rate_limited_until`, but some (e.g. accounts on the same provider) show `last_error_type=rate_limited`, `error_code=429`, yet `rate_limited_until=NULL`, so they are re-selected immediately and keep absorbing 50s 429s. Hypothesis: under the oauth profile (`useUpstreamRetryHints`), a 429 with no parseable `Retry-After` yields no cooldown. Request: apply a base cooldown floor on any 429 even when no upstream hint is present.

**2. `backoff_level` never increments.** Across all these OAuth connections `backoff_level=0` despite hundreds of repeat 429s each. Exponential backoff does not appear to accumulate for them, so repeat offenders never get progressively longer cooldowns.

**3. modelLockout may not gate priority-strategy combos.** `modelLockout` is enabled (`errorCodes` includes 429, 30s base to 5min), but `domain_lockout_state` is empty and the priority-combo legs are still hammered. Please confirm whether modelLockout is enforced for priority (non auto-routing) combos, or only for auto-routing candidates.

[...]

PR #11408, opened 2026-08-24
fix(quota-share): release the winner's reserved in-flight slot (#11371)

### LINKED ISSUE #11371: fix(quota-share): the in-flight slot is reserved and never released — no production caller of decrementInflight
## Summary

`selectQuotaShareTarget` reserves an in-flight slot for the winner and returns a release callback. **No production code ever calls it**, so the counter only ever goes up.

## Where

`open-sse/services/combo/quotaShareStrategy.ts:318-336`

```ts
  // Reserve the winner's in-flight slot immediately.
  const winnerConnectionId = winner.connectionId ?? "";
  if (winnerConnectionId) incrementInflight(winnerConnectionId, undefined, nowMs);
  ...
  return { target: winner, orderedTargets, decrementInflight: release };
```

The only production caller drops it — `open-sse/services/combo/applyStrategyOrdering.ts:232`:

```ts
    orderedTargets = selectQuotaShareTarget(orderedTargets, combo.name, qsModel, Date.now(), {
      maxConcurrentByConnection: qsMaxConcurrent,
    }).orderedTargets;
```

Every reference to `decrementInflight` in the tree, verified at `dafb4ae80`:

```
quotaShareStrategy.ts:35,251,289,333,336   definition / type / no-op / the closure itself
quotaShareInflight.ts:11,82                docstring and definition
tests/unit/quota-share-strategy.test.ts    26,88,93,225,280,284,288,291,292,299,307
```

Nothing else. The module docstring states the contract that is not met — `quotaShareInflight.ts:8-15`: *"The normal path calls decrementInflight() (returned to the caller as a callback) once the request settles… so the counter can never leak forever."*

## Why the lease TTL does not save it

`quotaShareInflight.ts:69-71`:

```ts
  const base = slot && slot.expiresAtMs > nowMs ? slot.count : 0;
  const newCount = base + 1;
  _inflightMap.set(connectionId, { count: newCount, expiresAtMs: nowMs + leaseMs });
```

Each increment refreshes the whole slot's expiry, so a connection receiving traffic more often than once per `DEFAULT_LEASE_MS` (120 s) never expires, and its count grows monotonically for the process lifetime.

## Consequences

- `pickByInflightP2C` (`quotaShareStrategy.ts:224-231`) compares the two counters, so it degenerates from "least loaded" into "fewest lifetime dispatches", pulling the split toward even and overriding the operator's configured DRR weights.
- The FASE 2.1 `max_concurrent` gate (`quotaShareStrategy.ts:151/160`) sees phantom load, pushes every connection past its cap, and fail-opens permanently.

## The repo's own test says so

`tests/unit/quota-share-strategy.test.ts:223-225`:

```ts
      // Simulate the request settling (as a real caller's finally handler would),
      // so the P2C in-flight tie-break sees equal load and the DRR weighting shows.
      r.decrementInflight();
```

The 2:1-weight assertion passes only because the test performs the release that production never performs. Baseline at `dafb4ae80`: 29/29 green.

## Two possible shapes — this is why I am opening an issue rather than a pull request

1. **Thread the callback out.** `applyStrategyOrdering` returns the array today, so the release has to travel through it and `targetResolution.ts:418` to the `finally` in `combo.ts:2786`. Correct and complete, but `targetResolution` has several return paths (the auto branch, the early-response branch) and downstream filters reassign `orderedTargets`; every one of them has to release exactly once, and a missed path leaks while a doubled path corrupts the count in the other direction.

2. **Release in `combo.ts` beside `quotaShareConcurrencyRelease`** (`:2786`). Two or three lines, but `qsConnectionId` at `:2770` is only read when `quotaShareConcurrencyEnabled` is true, while the reservation happens whenever quota-share ordering runs — so this shape needs a way to know a reservation actually happened, or it decrements a connection nobody reserved.

I did not want to pick between those in a concurrency path on a guess. Say which you prefer and I will send the patch with tests for it — including the case the current suite is missing, which is that a dispatch through the real ordering path leaves the counter back at zero.

## Environment

- Verified by code reading at `release/v3.8.50` (`dafb4ae80`); no live gateway involved.

oyi77:
## Summary

Fixes #11371.

`selectQuotaShareTarget` reserves an in-flight slot for its winner and returns an idempotent release callback — but no production caller ever invoked it, so counters only grew: P2C degenerated from "least loaded" into "fewest lifetime dispatches" (overriding operator DRR weights) and the FASE 2.1 `max_concurrent` gate fail-opened permanently past its cap.

This threads the release out of the ordering path (shape 1 from the issue):

- `applyStrategyOrdering` now returns `{ orderedTargets, quotaShareRelease }` — non-null only when the quota-share strategy ran
- `resolveComboTargetPipeline` releases it itself when a post-selection hard filter produces an earlyResponse, otherwise hands it to the host
- `handleComboChatInner` invokes it in the dispatch `finally` (beside `quotaShareConcurrencyRelease`) and on the two early returns between resolution and the try block

The callback is idempotent, so exactly-once is enforced at the source and defensive double-calls are harmless.

## Tests

- New case the issue called out as missing: ordering through the **real** `applyStrategyOrdering("quota-share", …)` path reserves one slot and leaves the counter back at **zero** after release; double-release floors at 0; non-quota-share strategies return a null release (`tests/unit/quota-share-strategy.test.ts`)
- Updated the split-guard tests for the new return shape (`combo-apply-strategy-ordering-split.test.ts`, `prompt-cache-affinity.test.ts`)

## Verification

```
quota-share-strategy + apply-strategy-ordering-split + prompt-cache-affinity: 45/45 pass
combo-strategy-fallbacks + router-strategies:                                41/41 pass
combo-target-resolution-split:                                                6/6 pass
tsc --noEmit: no errors in changed files
```

Base: `release/v3.8.51` (`3192eb88d`). No behavior change for non-quota-share strategies.

### DIFF
--- a/open-sse/services/combo.ts
+++ b/open-sse/services/combo.ts
@@ -931,6 +931,7 @@ async function handleComboChatInner({
     // Surface a recovery hint + auto-clear the session pin after enough consecutive
     // no-target failures (silent-stop fix). Threshold of 3 prevents a one-off account
     // wipe from destroying the prompt-cache pin benefit on the next request.
+    recordComboFailure(effectiveSessionId, combo.name);
     // #11371: same early-exit release as the pinned-turn path above.
     targetResolution.quotaShareRelease?.();
     return errorResponseWithComboDiagnostics(
--- a/open-sse/services/combo.ts
+++ b/open-sse/services/combo.ts
@@ -902,6 +902,9 @@ async function handleComboChatInner({
   if (activeNativeTurnPin) {
     orderedTargets = applyNativeCodexTurnPin(orderedTargets, activeNativeTurnPin);
     if (orderedTargets.length === 0) {
+      // #11371: quota-share ordering already reserved a winner slot; release it on
+      // this early exit (idempotent).
+      targetResolution.quotaShareRelease?.();
       return errorResponse(
         409,
         "The pinned native Codex turn target is no longer available; the turn cannot be moved to another provider"
@@ -928,7 +931,8 @@ async function handleComboChatInner({
     // Surface a recovery hint + auto-clear the session pin after enough consecutive
     // no-target failures (silent-stop fix). Threshold of 3 prevents a one-off account
     // wipe from destroying the prompt-cache pin benefit on the next request.
-    recordComboFailure(effectiveSessionId, combo.name);
+    // #11371: same early-exit release as the pinned-turn path above.
+    targetResolution.quotaShareRelease?.();
     return errorResponseWithComboDiagnostics(
       404,
       "Combo has no executable targets",
@@ -2785,6 +2789,9 @@ async function handleComboChatInner({
     return await dispatchWithCooldownRetry();
   } finally {
     quotaShareConcurrencyRelease?.();
+    // #11371: release the in-flight slot quota-share ordering reserved for its
+    // winner — the counter must not leak monotonically upward across requests.
+    targetResolution.quotaShareRelease?.();
     // G2: Clean up candidate registry to prevent unbounded memory growth.
     _unregisterExecutionCandidates(_registeredExecutionKeys);
   }
--- a/open-sse/services/combo/applyStrategyOrdering.ts
+++ b/open-sse/services/combo/applyStrategyOrdering.ts
@@ -20,6 +20,21 @@ import {
 } from "./targetSorters.ts";
 import type { ComboLike, ComboLogger, ResolvedComboTarget } from "./types.ts";
 
+/**
+ * Result of {@link applyStrategyOrdering}.
+ *
+ * `quotaShareRelease` carries the idempotent release for the in-flight slot that
+ * quota-share ordering reserves for its winner (#11371). It is non-null only when
+ * the `quota-share` strategy ran; every other strategy leaves it null. The caller
+ * MUST invoke it exactly once when the request settles — selection reserves the
+ * slot, so dropping the callback leaks the counter monotonically upward and
+ * degenerates P2C into "fewest lifetime dispatches".
+ */
+export interface ApplyStrategyOrderingResult {
+  orderedTargets: ResolvedComboTarget[];
+  quotaShareRelease: (() => void) | null;
+}
+
 export interface ApplyStrategyOrderingDeps {
   combo: ComboLike;
   config: Record<string, unknown>;
@@ -45,9 +60,10 @@ export async function applyStrategyOrdering(
   strategy: string,
   initialOrderedTargets: ResolvedComboTarget[],
   deps: ApplyStrategyOrderingDeps
-): Promise<ResolvedComboTarget[]> {
+): Promise<ApplyStrategyOrderingResult> {
   const { combo, config, body, log, apiKeyAllowedConnections, sessionKey } = deps;
   let orderedTargets = initialOrderedTargets;
+  let quotaShareRelease: (() => void) | null = null;
 
   if (strategy === "lkgp") {
     try {
@@ -229,14 +245,18 @@ export async function applyStrategyOrdering(
     const qsModel =
       typeof body?.model === "string" ? body.model : (orderedTargets[0]?.modelStr ?? "");
     const qsMaxConcurrent = await resolveMaxConcurrentByConnection(orderedTargets);
-    orderedTargets = selectQuotaShareTarget(orderedTargets, combo.name, qsModel, Date.now(), {
+    const qsSelection = selectQuotaShareTarget(orderedTargets, combo.name, qsModel, Date.now(), {
       maxConcurrentByConnection: qsMaxConcurrent,
-    }).orderedTargets;
+    });
+    orderedTargets = qsSelection.orderedTargets;
+    // #11371: the reservation made inside selectQuotaShareTarget must outlive this
+    // call — hand the release to the host so it can fire it when the request settles.
+    quotaShareRelease = qsSelection.decrementInflight;
     log.info(
       "COMBO",
       `Quota-share ordering: ${orderedTargets[0]?.modelStr}${orderedTargets[0]?.connectionId ? ` (${orderedTargets[0].connectionId})` : ""} selected (DRR+P2C)`
     );
   }
 
-  return orderedTargets;
+  return { orderedTargets, quotaShareRelease };
 }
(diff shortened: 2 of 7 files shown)


abhisheksharma2411 (commented):
The diagnosis in #11371 is exactly right and the fix shape is the correct one — threading the release out of `applyStrategyOrdering` rather than releasing inside it, making it idempotent, and firing it from `finally` plus both early exits. The interface doc explaining *why* dropping the callback degenerates P2C into "fewest lifetime dispatches" is the part that will stop this regressing again.

**But there is an accidental deletion in this diff, and I think it is a real regression.**

## `recordComboFailure` is removed from the no-executable-targets path

```diff
   if (orderedTargets.length === 0) {
     // Surface a recovery hint + auto-clear the session pin after enough consecutive
     // no-target failures (silent-stop fix). Threshold of 3 prevents a one-off account
     // wipe from destroying the prompt-cache pin benefit on the next request.
-    recordComboFailure(effectiveSessionId, combo.name);
+    // #11371: same early-exit release as the pinned-turn path above.
+    targetResolution.quotaShareRelease?.();
```

The release call is correct and belongs there. It looks like it replaced the line instead of being added above it — the comment describing the auto-clear is still sitting on top of code that no longer does it.

Counting call sites in `combo.ts`:

```
base:     34 (import), 931, 1415, 2626, 2723, 2755   → 5 calls
#11408:   34 (import),      1419, 2630, 2727, 2759   → 4 calls
```

The four survivors are all on dispatch-failure paths. Nothing else records a failure for the *"Combo has no executable targets"* early return.

**Why it matters:** `recordComboFailure` is what increments the per-`(session, combo)` counter and clears the session's model pin once it reaches `COMBO_FAILURE_THRESHOLD`. With this call gone, a session that keeps hitting "no executable targets" never accumulates a streak, so the pin is never auto-cleared — which is the silent-stop the comment above it was added to fix. A user whose pinned target has gone away stays stuck, and the recovery hint in the 404 becomes the only way out.

**And nothing catches it.** I ran the combo unit suite on this branch: `tests/unit/combo/*.test.ts` → **162 tests, 162 pass, 0 fail**, same as the base branch. There is a `pin-recovery.test.ts`, but it covers `buildRecoveryHint` and payload shape, not the recording call on this path. So the behaviour change is silent in CI as well as in the diff.

The fix is presumably just to keep both:

```ts
recordComboFailure(effectiveSessionId, combo.name);
// #11371: same early-exit release as the pinned-turn path above.
targetResolution.quotaShareRelease?.();
```

If it was deliberate — say you judged that a no-target combo shouldn't count toward the pin-clear streak — then it's a separate behavioural change that deserves its own line in the description and its own test, because it is not implied by #11371.

## On the rest

The `finally` placement looks right: `quotaShareConcurrencyRelease?.()` and the new `quotaShareRelease?.()` are different reservations (admission vs the P2C in-flight slot), and calling both is correct rather than redundant. Firing the release again from the early exits is safe because `release` latches on `released`, so the double-call from an early-exit path that later also unwinds through `finally` is a no-op — worth keeping that `released` flag prominent, since the whole design leans on it.

One question: `selectQuotaShareTarget` reserves the slot at *selection* time using the selection-time clock, but the request may sit in cooldown-retry for a while before dispatch. Is the lease long enough that a slow request's slot is never reaped as expired before its own release runs, or is the idempotent release the only thing keeping the count honest there? Not a defect I can point at — I ask because "reserve early, release in `finally`" is exactly where lease expiry and explicit release tend to disagree under load.

CI on this head currently shows `Unit Tests fast-path (2/4)` and `(4/4)` red plus `No new ESLint warnings` — I did not chase those, since the combo suite is green locally and they may be unrelated to this change.

oyi77:
## CI triage (all four red checks investigated)

|---|---|---|
| Unit Tests fast-path (2/4) | **Flake, not this diff** | 8529/8533 pass; sole failure `ollama-local-embedding-2824.test.ts` is an `rmSync ENOTEMPTY` race in the test's own `after()` tmp-cleanup hook — a file untouched by this PR. Passes locally. |
| Unit Tests fast-path (4/4) | **Flake, not this diff** | 8275/8280; sole failure `token-health-check-kimi.test.ts` (jitter-window assertion) — file untouched by this PR. Passes locally. |
| No new ESLint warnings | **Pre-existing on the base branch** | `run-eslint-json --max-warnings 0` exits 2 with "suppressions left that do not occur anymore" — reproduced verbatim on a stashed clean `release/v3.8.51` checkout. Note the currently-green PRs (#11428/#11429/#11434) all target `release/v3.8.50`, whose suppression state differs. |
| dast-smoke | **Does not reproduce from this diff** | The reported 500s are `DELETE /api/keys/{id}`, `OPTIONS /api/keys`, `OPTIONS /api/auth/logout` — probed against this branch running locally: **204 / 401 / 204** (all documented statuses). The fuzz 500s are unrelated to combo dispatch and appear environmental to the CI harness run. |

No code changes required for this PR; happy to follow up if maintainers see otherwise.

diegosouzapw:
Obrigado pela contribuição — o fix de vazamento do contador in-flight do quota-share (#11371) é real e valioso, mas encontramos um problema antes de mergear.

No caminho `if (orderedTargets.length === 0)` que retorna 404 "Combo has no executable targets", o diff remove a chamada `recordComboFailure(effectiveSessionId, combo.name);` e a substitui pela chamada `targetResolution.quotaShareRelease?.();`, em vez de manter as duas. O comentário logo acima ("Surface a recovery hint + auto-clear the session pin after enough consecutive no-target failures...") permanece no código, mas passa a descrever um comportamento que não acontece mais nesse ponto — o auto-clear do session pin depois de 3 falhas consecutivas deixaria de contar essa falha específica.

Poderia ajustar para manter as duas chamadas (`recordComboFailure(...)` seguido de `targetResolution.quotaShareRelease?.()`) nesse callsite? Os outros dois pontos onde a release foi adicionada (o early-return de native-turn-pin e o `finally` bloco) não têm esse problema — são estritamente aditivos.

diegosouzapw:
Nice work — the reservation-leak mechanism matches the idempotent decrementInflight contract in quotaShareStrategy and the new ordering tests are exactly what #11371 asked for (45/45 green locally). One real regression to fix before merge: in handleComboChatInner's `orderedTargets.length === 0` early exit, the edit REPLACED `recordComboFailure(effectiveSessionId, combo.name)` with the quotaShareRelease call instead of adding beside it — the silent-stop pin auto-clear (#5923 Finding) no longer counts failures on this path, and no test caught it because none covers that path today. Please restore the recordComboFailure call alongside `quotaShareRelease?.()` and add a guard test asserting the combo-failure counter increments on an empty-pool early exit. Also note combo.ts picked up +56 lines on the current release tip since you branched, so a rebase onto origin/release/v3.8.51 is needed. Everything else looks merge-ready.

Issue #10788, opened 2026-08-19
fix(providers): reasoning effort not shown/configurable on opencode, ollama-cloud and nvidia providers


plexofy:
### OmniRoute Version

3.8.49 — the only version published to npm. 3.8.50 exists only in the release/v3.8.50 branch (see the release blocker below); it cannot be installed.

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 LTSC 24H2

### Node.js Version

26.5.0 (npm 11.17.0)

### Provider(s) Involved

opencode, opencode-zen, opencode-go, ollama-cloud, nvidia (ollama-cloud shares the opencode-go backend)

### Model(s) Involved

- opencode/open-code-go: deepseek-v4-pro, deepseek-v4-flash, glm-5.2, mimo-v2.5, grok-4.5, hy3, kimi-k3, qwen3.6-plus, qwen3.7-max, qwen3.7-plus (all models with effort-tier aliases) - ollama-cloud: deepseek-v4-pro, deepseek-v4-flash, glm-5.1, glm-5.2 (gpt-oss:20b / gpt-oss:120b are the only models that DO declare tiers) - nvidia: z-ai/glm-5.2, nvidia/nemotron-3-nano-30b-a3b, qwen/qwen3-next-80b-a3b-instruct, and all other reasoning-capable models

### Client Tool

opencode (opencode-cli)

### Description

This report covers two problems found while following up on #10629:

1. Reasoning effort is invisible, not configurable, and silently dropped on opencode, ollama-cloud and nvidia providers.

   - In OmniRoute's dashboard/catalog, these providers expose ONLY the base model ids (e.g. `opencode/deepseek-v4-flash-free`, `ollama-cloud/deepseek-v4-pro`, `nvidia/z-ai/glm-5.2`). The effort levels (`-low`, `-high`, `-max`) are NEVER shown for them — there is no effort selector and no way to see which reasoning effort is being applied by default.
   - This is provider-specific: for other providers in the same catalog, effort variants DO appear (e.g. `<provider>/<model>-high`, `...-max`), and inside opencode-cli itself the same opencode models DO allow changing effort (the aliases are listed via `opencode models opencode-go --verbose`).
[...]

Issue #11295, opened 2026-08-23
[BUG] reasoning_effort: declared-vocabulary clamp (#11274) and learned clamp (#11232) remap the same request in opposite directions


maxmad64bis:
## OmniRoute Version

3.8.50

## Installation Method

Built from source

## Operating System

Linux

## Description

Since #11232 and #11274, two `reasoning_effort` correctors coexist — pushing in **opposite directions**:

- **learned** from an upstream 400 (#11232): `medium` → **`low`**
- **declared** in the static registry (#11274): `medium` → **`high`**

So two models with the exact same accepted set `{low, high, max}` give different answers to the same request, depending only on whether they have a static declaration. And since `high` burns more tokens than `low`, the gateway can silently spend more than the client asked for.

Bonus inconsistency: `none`/`minimal` requests work on declared models (`none → low`) but fail forever (recurring 400) on undeclared ones with the identical vocabulary.

**Why this matters:**

- **Cost**: `high` burns far more reasoning tokens than `low`. An upgrade remap means the gateway silently spends beyond what the client asked; downgrade-only guarantees cost never exceeds intent.
- **Mid-conversation flip**: this already happened — before #11274 merged, `ox-alpha-free` answered `medium` with `low`; after, with `high`. A conversation spanning the merge changed effort mid-stream, which also invalidates provider-side prefix caches keyed on request parameters for every subsequent turn.
- **Precedent**: downgrade is OmniRoute's documented behavior (`docs/guides/USER_GUIDE.md`: "Haiku models do not accept the max tier, so OmniRoute **downgrades** that request"). The declared clamp is the outlier.

## Steps to Reproduce

1. `sanitizeReasoningEffortForProvider({reasoning_effort: "medium"}, "opencode-go", "ox-alpha-free", null)` → returns `high`
2. Record a learned set `{low, high, max}` for any undeclared model, run the same call against it → returns `low`

## Expected Behavior

One direction for everyone. Recommended: **downgrade-only everywhere** — the gateway should never silently exceed what the client asked for (tokens = cost); moving up should stay an explicit client choice (e.g. `-high` suffix). The static declaration should just seed the existing learned clamp instead of running its own upgrade mapping.

## Actual Behavior

The remap direction depends on whether the model has a static declaration: upgrade if declared, downgrade if learned.

## Test Impact

Needs a new unit test — invariant: same effort + same accepted set ⇒ same output, regardless of where the set comes from.

## Validation Plan

```sh
[...]

PR #11409, opened 2026-08-24
fix(providers): stop silently dropping reasoning effort on opencode families (#10788)

oyi77:
## Summary

Addresses the code facets of #10788 (part 1). Ollama Cloud tier declarations already landed upstream in #11307; this covers what is still missing on `release/v3.8.51` for **opencode / opencode-go / opencode-zen** and documents **nvidia**.

### Dispatch — the silent drop

`OpencodeExecutor.transformRequest` stripped the effort-suffixed alias (e.g. `glm-5.2-high`) down to the base id and injected a flat `reasoning_effort` body field. But opencode-go's `ChatCompletionRequest` has no such field for non-DeepSeek families (its `reasoning` is a structured object), so the tier never reached the upstream — every request ran at default effort with no error.

Fix:
- **DeepSeek V4** keeps the base-rewrite + `reasoning_effort` injection — that IS its native contract via opencode-go (#4647, sanitizer whitelist)
- **Every other family** now forwards the aliased model id verbatim. The suffix is their only native effort mechanism (exactly the ids `opencode models opencode-go --verbose` lists), so the tier finally reaches the upstream

### Registry — tiers invisible in the catalog

Declared `supportedThinkingEfforts` on base rows that had aliases but no tier vocabulary:

| Registry | Models declared |
|---|---|
| opencode-go | glm-5.2 `[high,max]`, mimo-v2.5 `[high,max]`, grok-4.5 `[low,medium,high]`, hy3 `[none,low,high]`, kimi-k3 `[max]`, qwen3.7-max/plus `[high,max]` |
| shared zen/go row | qwen3.6-plus `[high,max]` |
| opencode-zen | deepseek-v4-pro/flash `[none,low,high,max]`, glm-5.2 `[high,max]`, kimi-k3 `[max]` |

This lets catalog variant synthesis (#9485) and `sanitizeReasoningEffortForProvider` clamping work from one source of truth matching the executor's `EFFORT_TIERS`.

### nvidia — honest empty declaration

`nvidia/z-ai/glm-5.2` is now marked reasoning-capable with an **empty** tier list: its upstream only exposes a binary `chat_template_kwargs.enable_thinking` switch (`mapNvidiaGlm52ReasoningParams`), so there are no honest `-low/-high/-max` tiers to advertise. Synthesizing fake variants would trade one bug for another; the empty list suppresses them while keeping the reasoning flag accurate.

## Tests changed/added

- `opencode-go-effort-aliases-8353.test.ts`: transform samples updated — non-DeepSeek aliases must reach the wire untouched with NO injected flat field; new registry-parity test asserting every base row's vocabulary equals `EFFORT_TIERS`; nvidia binary-switch declaration test
- `opencode-go-effort-aliases-6922.test.ts`: glm/mimo transform expectations updated to alias-forwarding
- DeepSeek paths (#4647 tests, clobber guard) unchanged and still green

## Verification

```
opencode/nvidia/deepseek effort suites (8 files): 191/191 pass
tsc --noEmit: no errors in changed files
```

Base: `release/v3.8.51` (`3192eb88d`). Release-engineering facet of #10788 (npm publish flow) is out of scope here.

### DIFF
--- a/open-sse/config/opencodeZenGoSharedModels.ts
+++ b/open-sse/config/opencodeZenGoSharedModels.ts
@@ -12,5 +12,15 @@
 export const OPENCODE_ZEN_GO_SHARED_MODELS = Object.freeze([
   { id: "kimi-k2.7-code", name: "Kimi K2.7 Code" },
   { id: "qwen3.5-plus", name: "Qwen3.5 Plus", targetFormat: "claude", supportsVision: false },
-  { id: "qwen3.6-plus", name: "Qwen3.6 Plus", targetFormat: "claude", supportsVision: false },
+  {
+    id: "qwen3.6-plus",
+    name: "Qwen3.6 Plus",
+    targetFormat: "claude",
+    supportsVision: false,
+    // #10788: effort-tier aliases exist as explicit registry rows; declare the
+    // vocabulary on the shared base row so variant synthesis and the sanitizer
+    // agree on it for both opencode-go and opencode-zen.
+    supportsReasoning: true,
+    supportedThinkingEfforts: ["high", "max"],
+  },
 ]);
--- a/open-sse/config/providers/registry/nvidia/index.ts
+++ b/open-sse/config/providers/registry/nvidia/index.ts
@@ -18,7 +18,17 @@ export const nvidiaProvider: RegistryEntry = {
   passthroughModels: true,
   models: [
     // #6108: z-ai/glm-5.1 EOL'd 2026-07-02 (direct probe returns 410) — dropped.
-    { id: "z-ai/glm-5.2", name: "GLM 5.2" },
+    // #10788: NVIDIA's hosted GLM-5.2 exposes a BINARY thinking switch
+    // (chat_template_kwargs.enable_thinking), not effort tiers — see
+    // mapNvidiaGlm52ReasoningParams. Declaring an empty tier list keeps the
+    // catalog from synthesizing unresolvable -low/-high/-max variant ids while
+    // still marking the model reasoning-capable.
+    {
+      id: "z-ai/glm-5.2",
+      name: "GLM 5.2",
+      supportsReasoning: true,
+      supportedThinkingEfforts: [],
+    },
     // #3329/#6108: minimaxai/minimax-m3 stays excluded from the nvidia tier — it
     // still 404s here for most callers; the single 200 probe in #6108 was not
     // reproducible enough to override the #3329 guard. Re-add only once NVIDIA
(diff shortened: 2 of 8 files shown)


oyi77:
## CI triage

|---|---|---|
| Fast Quality Gates / No new ESLint warnings | **Pre-existing on the base branch** | `run-eslint-json --max-warnings 0` exits 2 with "suppressions left that do not occur anymore" — reproduced verbatim on a stashed clean `release/v3.8.51` checkout (22 baseline problems). The currently-green PRs all target `release/v3.8.50`, whose suppression state differs. None of the flagged surface is touched by this PR (registry declarations + one executor transform). |
| Unit Tests fast-path (4/4) | **Under investigation → likely flake** | Same class of single-test flakes as documented on #11408 (tmp-dir `rmSync ENOTEMPTY` race and jitter-timing assertions in files this PR does not touch). Both pass locally on this branch. |
| dast-smoke | **Does not reproduce from this diff** | The fuzz 500s target `DELETE /api/keys/{id}`, `OPTIONS /api/keys`, `OPTIONS /api/auth/logout` — probed locally against a server built from the lazy-executor sibling branch: 204/401/204, all documented statuses. |

No code changes required; will act on any maintainer findings.

diegosouzapw:
Thanks — the DeepSeek-vs-rest split and the registry-parity test are exactly right, and we reproduced the silent-drop bug on the current tip. One blocker before merge: your new supportedThinkingEfforts declarations activate the nearest-tier clamp from #11295/#11305, so a Codex request asking effort=low on opencode-go/glm-5.2 is now rewritten to high — 'Codex Responses routing keeps reasoning effort while dropping GPT-only verbosity' fails on this branch but passes on the base tip (reproduced locally; it is also why Unit Tests 4/4 is red). Please either update that guard expectation deliberately (with a note on what the upstream actually accepts) or scope the declarations so lower requested efforts survive. The ESLint/quality-gate reds are inherited base-red and not your problem.

oyi77:
Nearest-tier clamp fix pushed (`63ea2ed1f`): the Codex-reroute characterization now locks the #10788 behavior — glm-5.2 declares {high, max} on opencode-go, so out-of-vocab `low` clamps up to `high` while GPT-only `verbosity` is still stripped. The fixed test passes in Unit Tests fast-path (3/4).

Heads-up on the current red batch (Docs Gates, all 4 unit shards, Vitest): these reproduce at **pure upstream tip** `6435f618f` with none of this branch's content — e.g. `tests/unit/agent-card-route.test.ts` fails 0/5 and `tests/unit/providers-constants-split.test.ts` (231 APIKEY_PROVIDERS entries) fails at the tip itself, and `check:docs-counts` reports README/AGENTS/llm.txt still saying "159 migrations" while code has 160 after today's quota-share merge (#11408). They hit any PR merging against the current tip.

PR #11411, opened 2026-08-24
fix(combo): enable genuine quota-aware routing for generic providers (antigravity, claude, etc.)

HouMinXi:
## Summary
- In production App Router runtime, `registerGenericQuotaFetchers()` and bespoke quota fetchers were never called during startup, leaving `quotaFetcherRegistry` empty for generic providers (`antigravity`, `claude`, `github`, `cursor`, `kiro`, `glm`, etc.). This caused `reset-aware`, `reset-window`, and `headroom` combo routing strategies to fail open with a default `0.5` dead score, silently degenerating into naive round-robin.
- Calls `registerQuotaFetchers()` during `instrumentation-node.ts` startup.
- In `genericQuotaFetcher.ts`, normalizes provider-native window names (e.g. Antigravity model quotas and Claude 5h/7d) into canonical `window5h` and `window7d` structural snapshots.
- In `quotaScoring.ts`, updates `scoreResetAwareQuota` to resolve windows using `resolveQuotaWindowByName` so that generic normalized windows participate in scoring.
- In `saturationSignals.ts`, extracts saturation percentages from nested `quotas` maps, fixing the issue where Antigravity headroom saturation was always evaluated as 0.
- Adds comprehensive unit tests in `tests/unit/universal-quota-aware-routing.test.ts` verifying that lower-used Antigravity/Claude connections score higher and are prioritized in `orderTargetsByResetAwareQuota`.

## Verification
- `npx tsx --test tests/unit/universal-quota-aware-routing.test.ts` (6/6 pass)
- All 433 unit tests in `tests/unit/services/` and `tests/unit/combo/` pass cleanly.
- Code-forge review: PASS.
- ⚠️ base-red inherited: #9985

### DIFF
--- a/open-sse/services/combo/quotaScoring.ts
+++ b/open-sse/services/combo/quotaScoring.ts
@@ -261,7 +261,7 @@ function getWindowsMapQuotaWindow(
   );
 }
 
-function resolveQuotaWindowByName(
+export function resolveQuotaWindowByName(
   quota: unknown,
   windowName: ResetWindowName
 ): QuotaWindowSnapshot | null {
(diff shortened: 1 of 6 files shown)


diegosouzapw:

diegosouzapw:
Great find on the window-normalization gap and excellent tests — running the new suite against the un-patched base fails 5/6 and passes 6/6 with your commit, which is exactly the TDD proof we want. Two things before this can merge: (1) the PR currently shows ~15.8k changed lines / CONFLICTING because the branch carries ~42 commits that were already squash-merged to the release (their content is byte-identical to tip — we verified). Please rebuild it: cut a fresh branch from origin/release/v3.8.51 and cherry-pick only 881067e1 — we confirmed all six touched files are drift-free, so it applies cleanly. (2) The premise needs a small correction: quota fetchers ARE registered at module load in src/sse/handlers/chat.ts (lines 194-231), which every App Router chat route imports — so combo routing mid-request does see a populated registry today. The boot-time hook is still worth merging (deterministic timing before first request + covers non-chat surfaces like the auto-combo candidates endpoint), but please reframe the description accordingly. Minor nit: normalizeQuotaWindows both Object.assigns window5h/window7d INTO the provider-native windows map and spreads them top-level — pick one spot (top-level spread is what the scorer reads) so synthetic keys don't leak into maps the UI iterates.

PR #11412, opened 2026-08-24
fix(sse): skip locked models in stored auto combos

jacobsparts:
## Problem

After a stored auto combo received a 429, its locked model could immediately score first again because stored-combo candidate construction did not consult active model lockouts. That caused tight retries against a model already known to be unavailable.

## Summary

- Exclude active per-connection model lockouts while scoring candidates for stored auto combos.
- Keep a provider-level candidate while any active provider connection is unlocked, and fail open when no active connections are enumerated.

## Related Issues

- Related to #9985 (inherited `release/v3.8.50` base status; not caused by this patch)

## Validation

- [x] Change type: routing
- [x] Focused regression tests passed
- [x] `npm run check:known-symbols`
- [ ] `npm run lint`
- [x] Based on the current active release tip when created
- [x] Production-code changes include a new or updated automated test

- Focused candidate and hidden-model suites: 14 passed
- `npm run check:known-symbols`: passed
- Changed test file ESLint: passed

This remains a draft. Final repository lint/category results will be recorded before review.

## Tests Added Or Updated

- `tests/unit/combo-account-allowlist-3266.test.ts`

## Coverage Notes

- Regression coverage exercises the reported failure and its availability/release boundary.
- No known touched-file coverage decrease.

## Reviewer Notes

- Uses the existing `isModelLocked` import and `connectionsByProvider` map; no additional query or dependency is introduced.

### DIFF
--- a/open-sse/services/combo.ts
+++ b/open-sse/services/combo.ts
@@ -630,7 +630,18 @@ export async function buildAutoCandidates(
   // Filter out candidates whose model is hidden by the user in the dashboard
   return candidates.filter((c) => {
     const hiddenModels = hiddenModelsMap.get(c.provider);
-    return !hiddenModels?.has(c.model);
+    if (hiddenModels?.has(c.model)) return false;
+
+    if (c.connectionId) {
+      return !isModelLocked(c.provider, c.connectionId, c.model);
+    }
+
+    const providerConnections = connectionsByProvider.get(c.provider) ?? [];
+    if (providerConnections.length === 0) return true;
+
+    return providerConnections.some((connection) =>
+      !isModelLocked(c.provider, String(connection.id ?? ""), c.model)
+    );
   });
 }
 
--- a/tests/unit/combo-account-allowlist-3266.test.ts
+++ b/tests/unit/combo-account-allowlist-3266.test.ts
@@ -26,6 +26,7 @@ const core = await import("../../src/lib/db/core.ts");
 const providersDb = await import("../../src/lib/db/providers.ts");
 const { buildAutoCandidates, handleComboChat } = await import("../../open-sse/services/combo.ts");
 const { getProviderCredentials } = await import("../../src/sse/services/auth.ts");
+const accountFallback = await import("../../open-sse/services/accountFallback.ts");
 const { normalizeComboStep } = await import("../../src/lib/combos/steps.ts");
 const { buildPrecisionComboModelStep } = await import("../../src/lib/combos/builderDraft.ts");
 
@@ -58,6 +59,7 @@ async function seedConn(name: string, tags?: string[]) {
 }
 
 test.beforeEach(async () => {
+  accountFallback.clearAllModelLockouts();
   await resetStorage();
 });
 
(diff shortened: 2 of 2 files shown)


jacobsparts:
Closing pending a replacement draft. Real-environment diagnosis showed the original lockout filter was incomplete because stored Auto-Combo routing aliases can differ from the canonical provider IDs used by connections and lockout keys. Any replacement should cover alias-to-canonical resolution and include end-to-end validation against a real provider 429.

abhisheksharma2411 (commented):
Small, focused, and the tests are the right ones — particularly the second, which pins that a provider-level candidate survives while *any* connection is unlocked and only disappears when they all are. Ran the file locally on this head: **10 tests, 10 pass**.

I do want to raise the interaction with #7360 before this lands, because I think it changes behaviour that was deliberately built.

## Dropping an all-locked candidate at build time bypasses the cooldown-aware wait

`isModelLocked` is already consulted at dispatch time (`combo.ts:1288`, `reason: "model_lockout"`). This PR adds a second check much earlier, in `buildAutoCandidates`, and the difference is not just where — it's what happens when *every* connection is locked.

The comment on the retry path says what that used to cost:

> Cooldown-aware retry (Variante A) … extended to "auto" combos too (#7360 — a 2-model "default" auto combo hitting Gemini TPM/RPM on both targets was crystallizing a 503 "all targets exhausted" after ~6s instead of waiting out the ~60s TPM window)

and the hoisting comment at `combo.ts:1072` is explicit that all-targets-locked must still reach the wait/crystallize decision rather than being pre-skipped into a terminal error.

Your second test asserts exactly the all-locked case:

```ts
accountFallback.lockModel("openai", second.id, "gpt-4o-mini", "rate_limited", 60_000);
const fullyLocked = await buildAutoCandidates([target], "auto-full-lockout");
assert.deepEqual(fullyLocked, []);
```

With an empty candidate list, `orderedTargets.length === 0` fires and the request terminates at *"Combo has no executable targets"* (404, `terminalReason: "no_executable_targets"`, recovery hint "switch combo") — before `dispatchWithCooldownRetry` is ever constructed. There is no 429 to crystallize, no `earliestRetryAfter`, and no wait.

So a transient 60s `rate_limited` lockout — the exact input in your test — becomes a terminal 404 telling the user to switch combos, where #7360's whole point was that it should wait out the window and retry. Arguably worse than the 503 that motivated #7360, since a 404 with `model_not_found` reads as a configuration problem rather than a transient one.

**None of this makes the PR wrong** — hiding locked models from *stored auto combos* is a real gap, and the per-connection branch is clearly an improvement. The question is whether the all-locked case should collapse the pool or fall through to the existing dispatch-time skip so the cooldown machinery still gets its turn. If the intent is that it collapses, that deserves saying out loud in the description, because it silently reverses a documented decision.

A shape that keeps both properties: filter locked *connections* out of a candidate, and keep the candidate itself when every connection is locked, letting `combo.ts:1288` skip it as it does today.

## Two smaller things

**The provider-level branch keeps the candidate without narrowing it.** When `c.connectionId` is null and only some connections are locked, the candidate survives — correctly — but nothing records *which* connection was the unlocked one, so selection can still land on a locked connection and rely on the dispatch-time skip. That's fine as belt-and-braces, but it means the new filter is best-effort at provider level and authoritative only at connection level. Worth a comment so the next reader doesn't assume the stronger guarantee.

**`String(connection.id ?? "")`.** An id-less connection becomes `""`, and `""` is the same key the dispatch path uses for genuinely connection-less targets (`target.connectionId || ""`). So a lock recorded against the connection-less key would read as "this connection is locked" here. Probably unreachable if `id` is always present — but if so, the `?? ""` is doing nothing and could be dropped, and if it *is* reachable, it's conflating two different keys.

Adding `clearAllModelLockouts()` to `beforeEach` in a shared file is the right call — lockout state is process-global and would otherwise leak between cases in this file.

PR #11413, opened 2026-08-24
fix(sse): enforce Retry-After limiter pauses

jacobsparts:
## Problem

The 429 header path logged a Retry-After pause but only evicted its limiter. The next request created a fresh limiter with full capacity and could dispatch immediately; body-derived pauses could also wedge when relying on a dynamically installed refresh heartbeat.

## Summary

- Block replacement limiters for the full Retry-After window after a 429.
- Use the same explicit, identity-guarded release timer for header- and body-derived hints, restoring configured capacity at the deadline.

## Related Issues

- Related to #9985 (inherited `release/v3.8.50` base status; not caused by this patch)

## Validation

- [x] Change type: routing
- [x] Focused regression tests passed
- [x] `npm run check:known-symbols`
- [ ] `npm run lint`
- [x] Based on the current active release tip when created
- [x] Production-code changes include a new or updated automated test

- Focused rate-limit suites: 32 passed
- `npm run check:known-symbols`: passed
- `git diff --check`: passed

This remains a draft. Final repository lint/category results will be recorded before review.

## Tests Added Or Updated

- `tests/unit/rate-limit-manager.test.ts`

## Coverage Notes

- Regression coverage exercises the reported failure and its availability/release boundary.
- No known touched-file coverage decrease.

## Reviewer Notes

- Pause timers are cleared during shutdown, disable, connection refresh, and test reset; stale timers cannot modify a replaced limiter.

### DIFF
--- a/open-sse/services/rateLimitManager.ts
+++ b/open-sse/services/rateLimitManager.ts
@@ -100,6 +100,7 @@ export const ZAI_WEB_REQUEST_QUEUE_MAX_WAIT_MS = 60_000;
 
 const limiterEffectiveSettings = new WeakMap<Bottleneck, Bottleneck.ConstructorOptions>();
 const preservedReplacementSettings = new Map<string, Bottleneck.ConstructorOptions>();
+const limiterPauseTimers = new Map<string, ReturnType<typeof setTimeout>>();
 const limiterWatchdog = new LimiterWedgeWatchdog({
   limiters,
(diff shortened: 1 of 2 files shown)


jacobsparts:
Closing because the proposed retry-after pause did not resolve the 429 failure in real-environment validation. The failure cause is still being investigated separately; any replacement should include passing deployment evidence.

PR #11417, opened 2026-08-24
fix(authz): match exact public routes exactly, not as prefixes (GHSA-74g9-q8f6-793h)

diegosouzapw:
Fixes the authorization bypass reported in **GHSA-74g9-q8f6-793h** by @ntdat812.

## The bug

`isPublicApiRoute()` matched every entry of `PUBLIC_API_ROUTE_PREFIXES` with `startsWith()`, but 11 of the 15 entries name **one exact route**, not a subtree. As a prefix, each also marked every adjacent path sharing its leading characters as `PUBLIC` — skipping the MANAGEMENT auth gate.

Reproduced on the tip before the fix:

```
/api/usage/om-usage-x            public=true class=PUBLIC reason=public_prefix
/api/usage/om-usageZZZ           public=true class=PUBLIC reason=public_prefix
/api/init-db                     public=true class=PUBLIC reason=public_prefix
/api/auth/login-as               public=true class=PUBLIC reason=public_prefix
/api/monitoring/health-detail    public=true class=PUBLIC reason=public_readonly_prefix
```

The first two are live: Next resolves `/api/usage/om-usage<anything>` to `src/app/api/usage/[connectionId]/route.ts`, whose handler has **no auth of its own** — it relies entirely on being classified MANAGEMENT. Unauthenticated, that reaches `fetchAndPersistProviderLimits()`, which is an existence oracle over connection ids (409 lease / 404 unknown / 400 unsupported / 200 quota JSON) and, for a connection id actually starting with `om-usage`, discloses live quota and can drive an OAuth token refresh — a write side effect — with no credentials.

The last line is an extra finding on top of the report: `matchesReadonlyPublic()` used `startsWith()` too, so a shadowed sibling also inherited the **CORS origin relaxation** `authz/pipeline.ts` keys on.

The general problem is the classification, not this handler: any route added later under a dynamic segment adjacent to one of the eleven entries inherits the bypass silently.

## The fix

Split the allowlist by shape:

- `PUBLIC_API_ROUTE_PREFIXES` — genuine subtrees only, **every entry ends in `/`**. A unit test asserts this, so the class cannot come back quietly.
- `PUBLIC_API_ROUTES_EXACT` — the eight single routes, matched exactly in both spellings.
- `PUBLIC_READONLY_CORS_API_ROUTES` — the three read-only "prefixes" were single routes too; `classify.ts` now calls `isPublicReadonlyCorsRoute()` instead of scanning a raw list, so a caller cannot reintroduce the prefix match.
- `/api/health` stays in its own set so it keeps classifying as `public_prefix` — folding it into the read-only set would silently widen CORS on it.

`dashboardCsrf.ts` carried a second copy of the prefix scan; it now shares `isPublicApiRoute()`, so the client CSRF exemption and the server classification cannot drift apart. Side effect in the safe direction: the three `LOCAL_ONLY` oauth auto-import routes were CSRF-exempt on the client while the server already required the token — the client now attaches it.

## Validation (TDD, Hard Rule #18)

New `tests/unit/authz/public-route-exact-match.test.ts` — 8 tests, red on the base, green here:

- every prefix entry ends in `/`
- each exact route stays public in both spellings
- twelve shadowed sibling paths are rejected (GET and POST)
- `/api/usage/om-usage<suffix>` classifies MANAGEMENT while `/api/usage/om-usage` stays PUBLIC
- genuine subtrees stay public all the way down
- the read-only method gate is unchanged
- the `public_readonly_prefix` reason set is unchanged and `/api/health` still reads `public_prefix`
- the `LOCAL_ONLY` oauth exclusions still win over the `/api/oauth/` subtree

Runs, hermetic (`env -u OMNIROUTE_API_KEY`):

- `tests/unit/authz/*.test.ts` — **244/244**
- sibling sweep (`public-api-routes`, `health-root-public-liveness`, `health-ping-route`, `monitoring-health-public-view`, `api-auth`, `management-auth-hardening`, `internal-usage-command`, `login-bootstrap-route`, `auth-redirect-login`) — **88/88**
- `tests/unit/ui/dashboard-csrf-fetch.test.tsx` (vitest) — **5/5**
- `typecheck:core` — exit 0, no output
- `eslint` on the four changed files — clean; `prettier --check` — clean
- `check:docs-sync`, `check:fabricated-docs`, `check:tracked-artifacts` — PASS

`docs/architecture/AUTHZ_GUIDE.md` was already stale (it documented a `/api/cloud/` entry that is not in the list) and its "Adding a New Route" recipe told contributors to *add the prefix* — which is how this class gets reintroduced. Updated to the new shape with the reasoning.

⚠️ base-red inherited: #9985

Co-authored-by: Nguyen Thanh Dat <ntdat812.dev@gmail.com>

### DIFF
--- a/docs/architecture/AUTHZ_GUIDE.md
+++ b/docs/architecture/AUTHZ_GUIDE.md
@@ -108,24 +108,48 @@ A successful policy returns `AuthSubject` with `kind ∈ { client_api_key, dashb
 
 `src/shared/constants/publicApiRoutes.ts` is the explicit allowlist:
 
+The list is split by **shape**, and the split is load-bearing (GHSA-74g9-q8f6-793h): a prefix is
+matched with `startsWith()`, so it also matches every adjacent path sharing its leading characters.
+`/api/usage/om-usage` as a prefix marked `/api/usage/om-usage<anything>` PUBLIC, and Next resolves
+that to `/api/usage/[connectionId]` — a handler with no auth of its own.
+
 ```ts
+// Genuine subtrees. Every entry MUST end in "/" (asserted by a unit test).
 PUBLIC_API_ROUTE_PREFIXES = [
+  "/api/auth/oidc/",
+  "/api/v1/", // treated as CLIENT_API in classify, not as "no-auth public"
+  "/api/oauth/",
+  "/api/codex/connect/",
+  "/api/telegram/",
+  "/api/cursor-cli/",
+];
+
+// Single routes, matched EXACTLY (with or without a trailing slash).
+PUBLIC_API_ROUTES_EXACT = new Set([
   "/api/auth/login",
   "/api/auth/logout",
   "/api/auth/status",
   "/api/init",
-  "/api/v1/", // treated as CLIENT_API in classify, not as "no-auth public"
-  "/api/cloud/",
   "/api/sync/bundle",
-  "/api/oauth/",
+  "/api/cli/connect",
+  "/api/usage/om-usage",
+  "/api/skills/collect/chaos",
+]);
+
+// Read-only single routes that also take the CORS origin relaxation.
+PUBLIC_READONLY_CORS_API_ROUTES = [
+  "/api/health/ping",
+  "/api/monitoring/health",
+  "/api/settings/require-login",
 ];
 
-PUBLIC_READONLY_API_ROUTE_PREFIXES = ["/api/monitoring/health", "/api/settings/require-login"];
+// Read-only single route WITHOUT the CORS relaxation.
+PUBLIC_READONLY_API_ROUTES_EXACT = new Set(["/api/health"]);
 
 PUBLIC_READONLY_METHODS = new Set(["GET", "HEAD", "OPTIONS"]);
 ```
 
-Read-only prefixes are public **only** for safe methods. Note: `classifyRoute()` excludes `/api/v1/*` and `/api/v1beta/*` from the PUBLIC fall-through — those are always `CLIENT_API` so the Bearer-key policy still applies.
+Read-only routes are public **only** for safe methods. Note: `classifyRoute()` excludes `/api/v1/*` and `/api/v1beta/*` from the PUBLIC fall-through — those are always `CLIENT_API` so the Bearer-key policy still applies.
 
 ## Adding a New Route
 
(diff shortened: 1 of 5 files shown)


ntdat812:
Reporter of GHSA-74g9-q8f6-793h here. I checked this branch out and ran the classifier against it rather than reading the diff — the advisory is only closed if the shipped fix actually closes it.

**Every path from the report is now guarded**, and every route that must stay reachable still is:

```
guarded  MANAGEMENT  /api/usage/om-usage-x      guarded  MANAGEMENT  /api/health/degradation
guarded  MANAGEMENT  /api/usage/om-usageZZZ     guarded  MANAGEMENT  /api/health/pingX
guarded  MANAGEMENT  /api/usage/abc123          guarded  MANAGEMENT  /api/monitoring/healthX
guarded  MANAGEMENT  /api/init-db               guarded  MANAGEMENT  /api/settings/require-loginX
guarded  MANAGEMENT  /api/auth/login-as         guarded  MANAGEMENT  /api/auth/logoutX
guarded  MANAGEMENT  /api/sync/bundleX          guarded  MANAGEMENT  /api/auth/statusX
guarded  MANAGEMENT  /api/cli/connectX          guarded  MANAGEMENT  /api/skills/collect/chaosX

PUBLIC  public_prefix           /api/health, /api/auth/login, /api/auth/logout, /api/auth/status,
                                /api/init, /api/sync/bundle, /api/cli/connect, /api/usage/om-usage,
                                /api/skills/collect/chaos
PUBLIC  public_readonly_prefix  /api/health/ping, /api/monitoring/health, /api/settings/require-login
PUBLIC  public_prefix           /api/auth/oidc/start, /api/oauth/x, /api/telegram/update,
                                /api/cursor-cli/anything, /api/codex/connect/finish
PUBLIC  client_api_v1           /api/v1/chat/completions
```

`/api/health` still classifies `public_prefix` rather than `public_readonly_prefix`, so the CORS relaxation in `authz/pipeline.ts` is not silently widened. That was the subtle one and the comment in the new code calls it out explicitly.

Two things I checked because they are not visible from the diff alone:

**1. Dropping `/api/cloud/` from the prefix list is safe.** It looks like scope creep — four routes exist under it and `PUBLIC_CLOUD_API_ROUTES` names only three, so `/api/cloud/credentials/update` moves from PUBLIC to MANAGEMENT. It never depended on the classification: the handler opens with `requireManagementAuth(request, { alwaysRequireAuth: true, invalidApiKeyStatus: 401 })`. So this tightens defence in depth without changing reachable behaviour.

**2. The second copy of the scan is handled, and better than I would have done it.** `dashboardCsrf.ts` had its own `PUBLIC_API_ROUTE_PREFIXES.some(startsWith)`, which this PR replaces with `isPublicApiRoute(url.pathname, method)`. Passing `method` matters: a read-only public route is public only for safe methods, so a `POST /api/monitoring/health` now correctly gets the CSRF token attached. Sharing the pathname-only helper instead would have left that case exempt.

I had a local fix for this sitting unpushed. It is redundant now and I am not opening a competing PR — this one covers the same ground, plus the `method` argument above that mine missed. Its test file and yours land on the same eight properties; the only case I have that yours does not is a direct assertion on `isPublicApiPath`, which is moot once `dashboardCsrf` stops calling it. Happy to hand over that case if you want it.

One suggestion, small: the invariant that keeps this fixed is `tests/unit/authz/public-route-exact-match.test.ts`'s "every prefix entry ends in a slash". That is the assertion a future contributor will trip over instead of shipping the bug again, and it is worth a line in `AUTHZ_GUIDE.md` next to the Pattern 3 text so it is discoverable from the docs rather than only from a failing test.

Issue #8075, opened 2026-07-22
feat(backend): pluggable external durable-state DB backend (PostgreSQL + MySQL) — unified design & demand thread


diegosouzapw:
## Summary

This is the **single canonical discussion** for giving OmniRoute a **pluggable external durable-state database backend** — so an active-active cluster (or a centrally-managed deployment) can share one source of truth for provider connections, API keys, combos, routing policies, quota/account state, session affinity, audit metadata, and migrations, instead of a local `storage.sqlite` file.

It **consolidates two independent proposals** that arrived within a day of each other, each championing a different first backend:

- **#8037** — *external durable-state backend, **PostgreSQL**-first* — proposed by @xiaoyaner0201
- **#8073** — *assess demand for **MySQL** 8.0+ as an optional backend* — proposed by @rushsinging

Both describe the **same architectural project** (they even agree it is **not** a drop-in ORM/driver swap: the current synchronous adapter, raw SQLite SQL, PRAGMA/WAL lifecycle, `sqlite-vec`, and 100+ migrations make it a staged, dialect-aware refactor — see also the earlier #2373 and #3576/#5409). Rather than split the design across two threads and risk building half of it twice, we're centralizing here. **#8037 and #8073 are being closed in favor of this issue** — with full credit to both proposers, who are invited to drive the design here.

**Co-authored / co-proposed by @xiaoyaner0201 (#8037, PostgreSQL) and @rushsinging (#8073, MySQL).**

## What we want to decide together

1. **Which backends, and in what order?** PostgreSQL and MySQL are both on the table; the abstraction should ideally support **both (and potentially more)** behind one contract, even if we implement one first. Is Postgres-first + MySQL-after (via a dialect-conformance suite) the right sequencing, or is there stronger demand for one?
2. **The abstraction contract.** Explicit repository/transaction interfaces for the state that must be shared across replicas — defined once, dialect-agnostic — rather than translating every SQLite call at once.
3. **What stays SQLite-only.** WAL/VACUUM/PRAGMA/page-cache tuning, file backup/import/export, and `sqlite-vec` vector storage remain SQLite-local. For external backends, FTS → Postgres FTS / MySQL FULLTEXT, and vector search → Qdrant.
4. **Config surface.** `DATABASE_DRIVER=sqlite|postgres|mysql` defaulting to `sqlite`; connection pooling, TLS, health checks, migrations per dialect.
5. **Migration path.** An offline one-way SQLite → external migration/export tool.
[...]

Issue #8037, opened 2026-07-21
feat(backend): external durable-state backend (PostgreSQL) for active-active cluster deployments


xiaoyaner0201:
## Problem / Use Case

OmniRoute already supports Caddy/Redis-oriented multi-replica deployment shapes, but its durable control-plane state remains centered on a local `storage.sqlite` file. That leaves an active-active cluster without a clean shared source of truth for provider connections, API keys, combos, routing policies, quota/account state, session affinity, audit metadata, and migrations.

A shared SQLite volume is not a safe substitute for a multi-writer database and keeps the database host/volume as a single point of failure. Redis helps with counters, cache, and rate limiting, but it should not become the durable authority for configuration and accounting state.

Previous requests #3576 and #5409 correctly identified that this is **not** a drop-in ORM/driver swap: the current synchronous adapter contract, raw SQLite SQL, PRAGMAs/WAL operations, and 100+ migrations make it a substantial architectural project. This proposal narrows the motivation and acceptance contract around one concrete outcome: supported active-active OmniRoute clusters with shared durable state.

## Proposed Solution

Introduce a staged, opt-in durable-state abstraction for cluster mode:

1. Define explicit repository/transaction interfaces for the state that must be shared across replicas, rather than attempting to translate every SQLite call at once.
2. Implement **PostgreSQL first** as the reference external backend. Add MySQL only after a dialect-conformance suite exists; both need not land in the same release.
3. Keep SQLite as the zero-config default for local/npm/Electron/Termux deployments.
4. Separate SQLite-only local features and maintenance operations (WAL checkpoint, PRAGMA, VACUUM, file backup) from portable domain persistence.
5. Provide an import/export path from `storage.sqlite` into the external backend, with a dry-run report and rollback guidance.
6. Make cluster ownership explicit for migrations, scheduled jobs, and cleanup tasks so multiple replicas do not run conflicting writers.
7. Treat Redis as an optional shared cache/counter layer, never as the only durable source of truth.

Possible configuration shape:

```env
OMNIROUTE_STORAGE_BACKEND=sqlite            # default
# OMNIROUTE_STORAGE_BACKEND=postgres
# OMNIROUTE_DATABASE_URL=postgresql://...
```

This is intentionally an RFC-level proposal, not a claim that the existing synchronous SQLite adapter can simply be pointed at another driver.

## Alternatives Considered

[...]

PR #11423, opened 2026-08-24
docs(design): external durable-state DB backend proposal (#8075)

oyi77:
## Summary

A codebase-grounded design answering the five open questions in #8075, so the discussion can move from 'should we' to 'which phase first'.

Closes nothing — this is the design input the issue asks for.

### Measured coupling census (`release/v3.8.51`)

- **126** db modules call the sync `.prepare()`/`.transaction()` shape
- **36** `.pragma()` sites (no PG/MySQL equivalent)
- **159** SQLite migration files
- **9** sqlite-vec / **8** FTS5 referencing files
- 4-driver `SqliteAdapter` seam exists but is synchronous — all drivers are SQLite, and an async backend cannot sit behind it honestly

Key claim: the contract must live ABOVE the adapter layer as an async, dialect-agnostic `DurableStateStore` port with per-domain repositories (connections, apiKeys, combos, quotaState, sessionAffinity, audit) adopted incrementally — not a driver swap under 126 sync modules.

### Answers proposed

1. **Backends/order**: Postgres-first; MySQL gated by a shared dialect-conformance suite (same behavioral pack must pass on both).
2. **Contract**: repository port + unit-of-work, defined once in `src/lib/db/ports/`.
3. **Stays SQLite-only**: WAL/PRAGMA lifecycle, file backup, sqlite-vec (vectors → Qdrant per #8075), FTS.
4. **Config**: `DATABASE_DRIVER=sqlite|postgres|mysql`, sqlite stays the non-negotiable zero-config default.
5. **Migration path**: one-way fail-closed export tool with per-table row-count + checksum receipt.

Plus a five-phase delivery plan where P1–P2 alone let a cluster share connection/key state.

Base: `release/v3.8.51`. Refs #8075 (and the consolidated #8037/#8073).

### DIFF
--- a/docs/design/EXTERNAL-DB-BACKEND.md
+++ b/docs/design/EXTERNAL-DB-BACKEND.md
@@ -1,0 +1,98 @@
+# Design: Pluggable External Durable-State DB Backend (PostgreSQL / MySQL)
+
+Status: PROPOSAL · Discussion issue: [#8075](https://github.com/diegosouzapw/OmniRoute/issues/8075) · Author: @oyi77
+
+Grounded in the tree at `release/v3.8.51` (`3192eb88d`). Every number below was measured on that tree, not estimated.
+
+---
+
+## 0. Why this document exists
+
+#8075 asks five questions (backends & order, abstraction contract, what stays SQLite-only, config surface, migration path). This document answers them with a census of how deeply the current data layer is coupled to SQLite, and proposes a repository-contract seam that lets external backends be adopted per domain instead of via a big-bang driver swap.
+
+## 1. Measured coupling census (`release/v3.8.51`)
+
+| Surface | Count | Implication for external backends |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


diegosouzapw:
Hi @oyi77 — this is a strong piece of analysis: measuring the coupling census instead of estimating it, spotting that the adapter seam is synchronous so an async backend honestly can't hide behind it, and the conformance-suite gate for MySQL are all the right calls. One placement problem: OmniRoute keeps design proposals/RFCs out of docs/ — planning artifacts go to the project's _tasks research area, and issue discussions stay on the tracker; a point-in-time census merged into docs/ starts rotting immediately (your '159 migrations' is already 163+ on the current tip). Could you publish the full proposal as a comment on #8075 so the five answers live with the questions? We'd rather reference it there than maintain a snapshot in the docs tree. Nothing else to change — the content itself is exactly the design input #8075 asked for.

oyi77:
Relocating per review: the proposal now lives as a comment on the demand thread — https://github.com/diegosouzapw/OmniRoute/issues/8075#issuecomment-5407898403 (full doc: docs/design/EXTERNAL-DB-BACKEND.md, branch preserved for reference).

PR #11434, opened 2026-08-24
fix(responses-continuation): recover a real id/output for passthrough and translate-mode replies

hartmark:
## Summary

`previous_response_id` virtualization (the mechanism that lets a non-native-OpenAI upstream still support Responses-API continuation, since `src/lib/db/responsesContinuationStore.ts` reconstructs history server-side from a logged call artifact) silently never engaged for the two most common reply shapes: a plain-text Responses-API reply routed through passthrough mode, and a plain-text reply routed through translate mode. In both cases the client got a correct reply, but the artifact needed to resolve a *later* `previous_response_id` was never captured (or was captured in a shape the resolver couldn't read), so the next turn's continuation request came back `400` and openclaw fell back to resending full history — continuation only ever worked for the narrower textual-tool-call-conversion sub-branch.

Three independent gaps, all in the client-payload capture/read path:

1. **Passthrough mode never pushed into `clientPayloadCollector`** for a plain-text reply — only the textual-tool-call conversion sub-branch pushed anything, so `clientPayloadCollector.getEvents()` was empty for the common case.
2. **`onComplete`'s `clientPayload` was unconditionally built in chat-completions shape** (`{choices:[...]}`, no `id`), even for a Responses-API client — the sibling `providerPayload` builder already had the `OPENAI_RESPONSES` carve-out; `clientPayload` never got the same treatment.
3. **`clientPayloadCollector.build()` always nests its summary under `.summary`**, but `extractResponsesId`/`resolvePreviousResponseState` read `.id`/`.output` directly, one level too shallow — so even a correctly-populated events list produced a `clientResponse` whose id/output were invisible to the code meant to read them.

A fourth gap affecting *translate* mode specifically: `emitTranslatedClientItem` in `stream.ts` pushes every translate-mode client-visible item wrapped as `{event, data}` (needed so `formatSSE` can emit both the SSE `event:` line and the `data:` payload separately), but every reducer's `ingest()` read `payload.type` directly, one level too shallow for that shape — so a client-facing summary built from translate-mode events never found a real response id/output either.

## Related Issues

- Related to the client-side follow-up in [openclaw/openclaw#123215](https://github.com/openclaw/openclaw/issues/123215) (compat-endpoint HTTP continuation opt-in) — that PR is blocked on this one having zero real effect through OmniRoute regardless of client-side correctness.

## Validation

- [x] Change type: routing (continuation/virtualization)
- [x] `npm run lint` (clean on touched files; 3 pre-existing errors elsewhere in `stream.ts`/`streamPayloadCollector.ts` confirmed present on the unmodified base commit too, unrelated to this change)
- [x] Reconciled with the current active release base (rebased onto `release/v3.8.50` tip)

## Tests Added Or Updated

- `tests/unit/attempt-logging-extract-responses-id.test.ts` (new)
- `tests/unit/responses-continuation-passthrough-client-payload.test.ts` (new)
- `tests/unit/responses-continuation-store.test.ts` (updated)
- `tests/unit/responses-continuation-translate-client-payload.test.ts` (new)
- `tests/unit/stream-payload-collector.test.ts` (updated)
- `tests/unit/stream-utils.test.ts` (updated)

## Coverage Notes

Touches `open-sse/handlers/chatCore/attemptLogging.ts`, `open-sse/utils/stream.ts`, `open-sse/utils/streamPayloadCollector.ts`, `src/lib/db/responsesContinuationStore.ts`. Each of the three (plus the translate-mode) gaps has a direct regression test asserting the previously-broken read path now resolves a real id/output. First attempt at the `clientPayloadCollector` fix (unconditionally pushing every event) regressed 3 pre-existing tests by double-counting against the tool-call sub-branch's own conditional pushes — caught by running the full related-suite sweep, not the narrow new tests alone; fixed by pushing only the fully-processed terminal `response.completed` event.

## Reviewer Notes

- Verified against a real repro, not just unit tests: a live isolated container (bind-mounting this branch) driven through a real openclaw gateway plus a scripted OpenAI-Responses-compatible mock backend — turn 2 now sends a genuinely trimmed 1-message continuation request and gets HTTP 200, where it previously got a bare 400 in ~20-40ms (failing before the request even reached the main chat-handler logging, i.e. rejected at the `previous_response_id` resolution step).
- Full related-test regression sweep (`npm run test:unit`) run against this branch after rebasing onto the current release tip; only 4 pre-existing failures remain (`proxy-rotation-latency`, `repro-glm-iso-reset-24h-cap`, a `minimax-m3` vision-metadata fixture, and the `route-body-validation-t06` volcengine-plan gate), all confirmed present on the unmodified base commit too via a separate worktree check — none are related to this change.
- `settings.call_log_pipeline_enabled` must be `true` for virtualization to engage at all (the whole feature reuses the call-log artifact store, which is off by default) — that's a separate, pre-existing "silent failure by default" product gap worth a follow-up, not something this PR changes.


hartmark:
Adding a second, related fix to this PR (same invariant: continuation resolvability for a real reply, not just id/output shape).

**Root cause of the follow-on bug:** `resolvePreviousResponseState` read the stored conversation's `input` from `pipeline.providerRequest.body.input`. That field only exists for a native passthrough Responses-API upstream. Any connection where OmniRoute translates the client's request into a different upstream wire format (e.g. Chat Completions `messages` for an auto-routed free-tier model) never has an `input` field on `providerRequest` at all — so the lookup always returned `null` and every continuation attempt failed with `previous_response_not_found`, even though the id was real, the artifact was `'ready'`, and the earlier fix in this PR had already made the id/output surface correctly.

**Fix:** read `input` from `clientRawRequest.body.input` instead. The client's own request is always Responses-API shaped (this store only ever fires for `sourceFormat === OPENAI_RESPONSES`), so it's the correct source for both native passthrough and translate-mode upstreams.

**Evidence:**
- Added a regression case for the translate-mode shape (`providerRequest.body.messages`, no `input`) to `tests/unit/responses-continuation-store.test.ts`, plus `clientRawRequest` fixtures on the existing cases (they only ever populated `providerRequest`, which is exactly why this slipped through). Confirmed the new case fails on pre-fix code for the right reason (`AssertionError: expected a resolved continuation state, got null`) and passes post-fix.
- Full file + directly related suites (handler, policy, passthrough/translate client-payload, marker-leak, translator, codex orphan-output) green: 160/160.
- Live-verified against a real openclaw gateway on an auto-routed free-tier connection (`nvidia/nemotron-3.5-lightning:free` via this branch running live): before the fix, turn 2 logged `retrying full history after rejected previous_response_id` on every attempt; after, the same two-turn session sent a 5-item delta with `previous_response_id` and completed in ~1.1s with zero rejection/retry.

Issue #6512, opened 2026-07-07
fix(providers): exclude paid-only models from auto/* candidate pool when hidePaidModels is on


diegosouzapw:
## Follow-up to #6328 / #6495

PR #6495 shipped the `hidePaidModels` setting (opt-in, default OFF) that filters paid-only models out of the `GET /v1/models` catalog. Its PR body explicitly listed the `auto/*` candidate-pool exclusion as a follow-up.

**Gap:** when `hidePaidModels` is ON, the `/v1/models` listing hides paid models, but `auto/*` combos (`auto/best-coding`, `auto/glm`, etc.) can still pick a paid-only backend into their candidate pool → 402/403 at request time, exactly the failure #6328 wanted to prevent.

## Proposal

In `open-sse/services/autoCombo/virtualFactory.ts::createVirtualAutoCombo` (which already loads `settings` via `getSettings()`), filter the `candidatePool` when `settings.hidePaidModels === true`, using the same predicate #6495 uses in `catalog.ts`:

```ts
keep = providerHasFreeModels(provider) && isFreeModel(provider, { id: model })
```

(from `@/shared/utils/freeModels`). Applied before the category/tier/family narrowing, so it covers every `auto/*` combo. If every candidate is paid-only the pool degrades to empty (the existing graceful empty-pool path handles it) — consistent with the opt-in intent.

## Acceptance
- With `hidePaidModels=true`, a paid-only backend is excluded from the `auto/*` candidate pool.
- With `hidePaidModels=false` (default), the pool is unchanged (regression guard).

diegosouzapw:
[...]

PR #6495, opened 2026-07-06
feat(api): add hidePaidModels setting to filter paid-only models from /v1/models catalog

chirag127:
## Summary

Adds a global opt-in setting `hidePaidModels` (default `false`) that, when enabled, filters paid-only models out of the unified `/api/v1/models` response. Uses the existing `isFreeModel()` primitive from `src/shared/utils/freeModels.ts` — a model qualifies as free when:

1. its id ends with `:free` (OpenRouter convention), OR
2. both `pricing.prompt` and `pricing.completion` are zero, OR
3. its id appears in `FREE_MODEL_BUDGETS` for that provider.

Providers absent from `FREE_MODEL_BUDGETS` (e.g. OpenAI, Anthropic) are treated as fully paid and filtered wholesale when the toggle is on.

## Motivation

Operators running OmniRoute against a curated free-tier stack currently see the full 4000-model catalog including paid-only providers they can never actually call. A single opt-in toggle lets them narrow `/v1/models` to only the models their `FREE_MODEL_BUDGETS` catalog covers, without adding per-model hide clicks or removing the paid providers entirely.

Ties back to the existing free-model detection surface — the same `isFreeModel()` helper already powers the "import only free models" flow in the Add-API-Key modal and provider-sync route. This PR extends its reach to the catalog output.

## Implementation

Two loops in `src/app/api/v1/models/catalog.ts` gain a paid-filter guard:

- **`PROVIDER_MODELS` (chat) loop** — skip when `shouldHidePaid(canonicalProviderId, model.id, model.pricing)` returns true.
- **OpenRouter loop** — reuses the already-computed `isFree` flag; when `hidePaid && !isFree`, skip the push.

`shouldHidePaid()` is a small local closure that consults `providerHasFreeModels()` first (short-circuits paid-only providers) then falls back to per-model `isFreeModel()`.

### Exempted surfaces (documented, intentional)

- Embedding / image / rerank / audio / moderation / video / music registries — represent local capabilities without pricing metadata.
- Combos / `auto/*` builtins / synced-available / custom / alias-backed models — extending scope to these would require per-entry pricing lookups not available today. **Recommended follow-up.**

### Storage

Setting stored via the existing `key_value` namespace `settings` — no migration, no schema change. Default is `false` so existing deployments see no behavioral change.

## Test

New unit test `tests/unit/models-catalog-hide-paid.test.ts` (Rule #18 regression guard):

1. Confirms the default is `false`.
[...]

PR #11441, opened 2026-08-24
fix(free-models): stop reporting discontinued Pollinations models as free

maxmad64bis:
⚠️ base-red inherited: #9985

## Summary

`hidePaidModels` promises "never send me to a paid model," but seven Pollinations models (gemini, gemini-fast, midijourney, midijourney-large, claude, claude-fast, claude-large) were still listed as free. The catalog marks them `freeType: "discontinued"` — Pollinations moved them behind a paid API key — yet the shared `isFreeModel()` predicate never read that field, so an `auto/*` pool could pick one and hit a 402/403 at request time.

This makes the predicate read the regime. A new `grantsFreeAccess(freeType)` table (compiler-checked: add a regime without classifying it and the build fails) decides which catalog entries count as free. Both the routing predicate (`src/shared/utils/freeModels.ts`) and the stricter `strictZeroCostFilter` (`open-sse/services/autoCombo/strictZeroCostFilter.ts:188`) now use it, so the two filters can't drift apart again.

## Related Issues

- Related to #6328, #6495, #6512 (the `hidePaidModels` zero-cost family).

## Validation

- [x] Change type: routing
- [x] `npm run lint`

## Tests Added Or Updated

- `tests/unit/autoCombo/free-regime-not-read-by-predicate.test.ts` (new): 7 cases — the 7 discontinued models aren't free, the 10 keyless Pollinations entries still are, `providerHasFreeModels("pollinations")` stays true, `filterPaidOnlyCandidates` drops them under `hidePaidModels`, no provider loses its free status, every regime is classified with the expected verdict, and the strict filter still excludes a discontinued entry.

## Coverage Notes

- Touches `src/` and `open-sse/`. Covered by the new unit test plus the existing autoCombo/strictZeroCost suites (187 tests), all still green. No coverage drop expected in any touched file.

## Reviewer Notes

- Behavior change to call out: `/v1/models`, `/api/models`, settings `export-json`, model-sync import (import-only-free), and the test routes now treat the 7 Pollinations premium models as paid. That's the intended fix — they need a key.
- Scope: surfaces that *collapse* the regime to a boolean are fixed. Surfaces that only *transmit* `freeType` (`/api/free-models`, the dashboard/radar) are left alone on purpose; the plugin already renders `discontinued` correctly (`naming.ts:169`).
- `grantsFreeAccess` is the single source of truth. `RECURRING` (homepage token totals) and `KEYLESS_FREE_TYPES` (no-auth shortcut) stay separate on purpose — folding them in would silently change the homepage totals.
- Known limitation (measured, not a bug): `isFreeModel` still treats a zero published price as free. We checked the live Pollinations API — none of the 304 models set `pricing.prompt`, and the 7 are billed — so the import path doesn't leak today. A provider publishing a null price on a `discontinued` entry would still short-circuit, but that signal is fresher than the curated catalog.
- Lint note: `check:open-sse-typecheck` is red on this base (#9985) from two unrelated `catalog.ts` TS2367 errors; this PR adds zero new type errors.

### DIFF
--- a/changelog.d/fixes/11441-discontinued-free-models.md
+++ b/changelog.d/fixes/11441-discontinued-free-models.md
@@ -1,0 +1,1 @@
+- **fix(free-models):** the shared `isFreeModel()` predicate no longer reports catalog entries marked `freeType: "discontinued"` as free, so `hidePaidModels` can't route to Pollinations' seven premium models that now require a paid key ([#11441](https://github.com/diegosouzapw/OmniRoute/pull/11441))
--- a/open-sse/config/freeModelCatalog.ts
+++ b/open-sse/config/freeModelCatalog.ts
@@ -70,6 +70,31 @@ export interface FreeModelTotals {
 const RECURRING = new Set<FreeModelFreeType>(["recurring-daily", "recurring-monthly", "keyless"]);
 
 /**
+ * What each free-tier regime engages for "can I route here without paying?".
+ * Exhaustive by construction: adding a member to `FreeModelFreeType` will not
+ * compile until it is classified here. `discontinued` is the one regime a
+ * provider uses to retire a free tier behind a paid key — it does NOT grant
+ * free access, and the shared predicate (`isFreeModel`) must read this instead
+ * of treating every catalogued id as free. `RECURRING` (above) answers a
+ * different question (which regimes feed the headline token totals) and is left
+ * independent on purpose — deriving it from this table would silently change
+ * the homepage totals.
+ */
+const FREE_REGIME_TRAITS = {
+  "recurring-daily": { grantsFreeAccess: true },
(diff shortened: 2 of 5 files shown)


diegosouzapw:

diegosouzapw:
Excellent fix. We verified the premise against freeModelCatalog.data.ts (exactly 7 discontinued pollinations entries), confirmed tests/unit/autoCombo is collected by vitest.mcp.config.ts, and ran your test both ways on the current release/v3.8.51 tip: 4 failed without the patch, 7/7 with it — a textbook fail-without-fix guard. The exhaustive FREE_REGIME_TRAITS table plus keeping RECURRING independent is exactly right. Only nit: the PR body cites base-red #9985; the current marker is #11449 — worth fixing when you next edit. Nothing else blocking from our side.

Issue #8330, opened 2026-07-23
test(e2e): contract test for the full provider journey


diegosouzapw:
## Feature Request (test infrastructure)

**Source:** Discussion #8273 by @nguyenha935 (sections 2-4); commitment made publicly in the discussion.

## Problem

Module-level tests pass while the real user journey breaks. A provider is not functional just because the creation route returns HTTP 200: multiple recent defects (compatible-provider test regex, `/v1/models` namespace incoherence, Topology blind to custom providers) only manifest across module boundaries.

## Proposed Solution

An end-to-end contract test covering the full journey as one gate:

```text
create provider -> add connection -> sync models -> select in Combo ->
Playground -> /v1/models exposure -> call via API key -> visible in Topology
```

Each step asserts against the same catalog/contract source, so a divergence in any surface fails the suite. Runs in CI as a blocking gate (integration layer, gated like `RUN_SERVICES_INT` suites if it needs a live server).

## Additional Context

- Discussion: #8273
- Directly guards the bugs tracked from the same discussion (regex `-chat-`, namespace/UUID, Topology)
- Related: `docs/architecture/QUALITY_GATES.md`

Reported-by: @nguyenha935 (Discussion #8273)
Refs #8273

diegosouzapw:
Good news: implementation for this request is already in progress. We're closing the issue to keep the tracker lean — the in-flight PR will reference this issue when it ships. Thanks for the suggestion!

PR #11448, opened 2026-08-24
Only advertise verified-working connections by default

### LINKED ISSUE #11446: feat(providers): auto-test provider credentials on creation and gate /v1/models by testStatus
## Problem

Today, when a provider connection is created (via API key, OAuth, or cookie-based auth), it's stored with `testStatus: "unknown"` and `isActive: true` by default — see `src/app/api/providers/route.ts` (`POST` handler). The only way a connection gets a real `testStatus` (`active`/`success`/failure states) is if the user manually triggers the `/api/providers/[id]/test` endpoint afterwards.

Meanwhile, `/v1/models` (`src/app/api/v1/models/catalog.ts`) only filters connections by `isActive !== false` — it does **not** check `testStatus` at all. So a freshly-added provider with an invalid/expired/misconfigured key (or one that was simply never tested) still has all of its models listed as available in the OpenAI-compatible catalog, indistinguishable from a provider that's actually working.

This makes it hard for API consumers (routers, client apps that call `/v1/models` to build a picker) to know which models are genuinely usable without independently calling `/api/providers` and reimplementing the healthy-connection filter that `src/shared/utils/connectionStatus.ts`'s `filterUsableConnections` already defines internally.

## Suggestion

1. **Auto-test on creation**: when a provider connection is created or its credentials are updated (API key, OAuth, cookie), automatically run the same validation the `/api/providers/[id]/test` endpoint performs, and persist the resulting `testStatus` immediately instead of leaving it at `"unknown"` until a manual test.
2. **Gate `/v1/models` by verified status by default**: only include a provider's models in the response when its connection passes `filterUsableConnections` semantics (`isActive !== false && (testStatus === "active" || testStatus === "success")`). An opt-in query param (e.g. `?includeUnverified=true`) could preserve the current "show everything" behavior for anyone who relies on it today.

## Why this matters

Right now any client integrating with Omniroute's catalog (including combo/router setups) has no reliable signal for "this model will actually work right now" without duplicating provider-health logic client-side. Testing automatically on credential entry, and reflecting that in `/v1/models`, would let the catalog be trusted at face value — no more silently listing models behind untested or broken keys.

Happy to help test this or discuss the right shape for the opt-in param if that's useful.

alltomatos:
Closes #11446.

## Problem

A newly created API-key connection was `isActive:true` immediately, with `testStatus` left at `"unknown"` until an operator manually clicked "Test" in the dashboard — but `/v1/models` only filters on `isActive`, not `testStatus`, so an untested (or invalid) key's models showed up in the catalog indistinguishable from a provider that actually works.

## Change

- `POST /api/providers` now creates connections `isActive:false` and fires an auto-test (`testSingleConnection`) right after creation, fire-and-forget so it doesn't block the 201 response.
- `testSingleConnection` itself now activates a connection whenever a test actually passes (or is `"unsupported"` and therefore can't be verified either way, in which case it's trusted like before). This applies to **any** test call — manual "Test" click, batch re-test, or the new auto-test — so a connection that starts failing and later recovers gets automatically reactivated on its next passing test, not just at creation time.
- A failing test deliberately leaves `isActive` untouched: this only ever turns a connection **ON** when verified, never OFF, so it can't silently take an already-working connection out of rotation on a transient failure (that's what the existing `rateLimitedUntil` cooldown is for).

Net effect: `/v1/models` (which already filters on `isActive`) now only advertises models from connections that have actually been confirmed to work, with zero changes needed to the catalog filtering itself — the existing filter does the job once the underlying data is correct.

## Operator-visible behavior change

A programmatically/API-created connection is briefly invisible from `/v1/models` between creation and the auto-test completing, and stays hidden if that creation-time auto-test fails (or errors) — the credential still needs to be fixed and re-tested either way. The existing batch-activate/re-test surface in the dashboard is the operator escape hatch for that case.

## Scope note

This PR covers the API-key creation path (`POST /api/providers`) and the shared test/activation logic used everywhere (so it benefits every connection regardless of how it was created, on its next test). It does **not** yet touch the ~20 OAuth import flows (Kiro, Codex, Cursor, Claude, Zed, Raycast, Trae, etc.) that call `createProviderConnection` with `isActive: true` hardcoded directly. Extending the same default there needs auto-test wired into each flow individually — skipping that would leave those connections permanently inactive, which is worse than today's behavior — so it's better scoped as a follow-up rather than one large, harder-to-review change touching ~20 unfamiliar flows at once.

## Testing

Rebased onto `release/v3.8.51` (the active development branch) — `[id]/test/route.ts` had drifted there (the quota-clear-on-valid-probe fix touches the same `updateData` literal); resolved so both behaviors compose: `testStatus`/error-clearing follows `clearErrorState` as before, and `isActive` only ever flips on on a passing probe.

Added `tests/unit/verified-connection-activation-11446.test.ts`, a deterministic, network-stubbed unit suite covering all four activation triggers:

- `POST /api/providers` creates a connection `isActive:false`.
- `testSingleConnection` activates it once a test actually passes.
- `testSingleConnection` never activates it when a test genuinely fails.
- `testSingleConnection` activates it when a test is skipped as unsupported (no registry entry for the provider), while a distinct exclusive-lease-busy skip — a temporary "try later", not "unverifiable" — correctly does not activate (see `tests/unit/exclusive-lease-connection-test-isolation.test.ts`).

Also updated the existing `tests/integration/provider-journey.contract.test.ts` (#8330): its STEP 2 created a connection against an unreachable stub host and asserted immediate `/v1/models` exposure, which no longer holds under the new default since that connection's own auto-test can never pass against that host. STEP 2 now simulates the operator's test having already passed, the same way STEP 3 already bypasses the real `/sync-models` network round-trip for the same non-deterministic host.

Commands run: `npx eslint --suppressions-location config/quality/eslint-suppressions.json` (clean) and `npm run typecheck:core` (clean) on the touched files; `node --import tsx/esm --test tests/unit/verified-connection-activation-11446.test.ts` (4/4 pass) and `node --import tsx/esm --test tests/integration/provider-journey.contract.test.ts` (8/8 pass, 6 skipped opt-in live steps).

### DIFF
--- a/.github/workflows/opencode-plugin-ci.yml
+++ b/.github/workflows/opencode-plugin-ci.yml
@@ -2,11 +2,11 @@ name: opencode-plugin CI
 
 on:
   push:
-    branches: [main, "release/**"]
+    branches: [main, release/v3.8.2]
     paths:
       - "@omniroute/opencode-plugin/**"
   pull_request:
-    branches: [main, "release/**"]
+    branches: [main, release/v3.8.2]
     paths:
       - "@omniroute/opencode-plugin/**"
     types: [opened, synchronize, reopened, ready_for_review]
--- a/src/app/api/providers/[id]/test/route.ts
+++ b/src/app/api/providers/[id]/test/route.ts
@@ -1053,8 +1053,23 @@ export async function testSingleConnection(connectionId: string, validationModel
 
   // Unsupported validation capability is neutral: the probe established that
   // this provider cannot be verified through the generic test surface, not
-  // that its credential is invalid. Do not mutate persisted credential health.
+  // that its credential is invalid. Do not mutate persisted credential health
+  // (testStatus/lastError/etc.) — but DO activate it if it isn't already: a
+  // connection that can never be health-checked would otherwise stay hidden
+  // from /v1/models forever under the "only advertise tested connections"
+  // default (isActive starts false on creation — see POST /api/providers),
+  // silently regressing every provider without a test surface.
   if (result.skipped === true) {
+    if (connection.isActive !== true) {
+      try {
+        await updateProviderConnection(connectionId, { isActive: true });
+      } catch (activateError) {
+        console.log(
+          `[ConnectionTest] Failed to activate unverifiable connection ${connectionId}:`,
+          (activateError as any)?.message || activateError
+        );
+      }
+    }
     return {
       ...result,
       latencyMs,
@@ -1099,6 +1114,14 @@ export async function testSingleConnection(connectionId: string, validationModel
 
   const updateData: Record<string, any> = {
     testStatus: clearErrorState ? "active" : result.valid ? connection.testStatus : "error",
+    // A passing test is the sole activation signal under the "only advertise
+    // tested-working connections" default — see POST /api/providers, which
+    // now creates connections isActive:false. Only ever flips ON here: a
+    // failing test intentionally leaves isActive untouched (a transient
+    // failure on an already-active, already-working connection must not take
+    // it out of rotation — that's what the cooldown/rateLimitedUntil below is
+    // for), so this never deactivates anything.
+    ...(result.valid ? { isActive: true } : {}),
     lastError: clearErrorState ? null : result.valid ? connection.lastError : result.error,
     lastErrorAt: clearErrorState ? null : result.valid ? connection.lastErrorAt : now,
     lastTested: now,
--- a/src/app/api/providers/route.ts
+++ b/src/app/api/providers/route.ts
@@ -43,6 +43,7 @@ import {
   getModelSyncInternalBaseUrl,
 } from "@/shared/services/modelSyncScheduler";
 import { finalizeValidatedChatGptWebCodexSecrets } from "@omniroute/open-sse/services/chatgptWebCodexAdmin.ts";
+import { testSingleConnection } from "./[id]/test/route";
 
 // GET /api/providers - List all connections
 export async function GET(request: Request) {
@@ -222,7 +223,15 @@ export async function POST(request: Request) {
       globalPriority: globalPriority || null,
       defaultModel: defaultModel || null,
       providerSpecificData,
-      isActive: true,
+      // Start inactive: a connection is only advertised via /v1/models (which
+      // filters on isActive) once a connection test has actually confirmed it
+      // works. The auto-test fired below flips this to true on success (or on
+      // an "unsupported" test, which cannot be verified either way and keeps
+      // the historical trust-it default) — see testSingleConnection in
+      // ./[id]/test/route.ts. A connection that fails its test, or is never
+      // tested because auto-test itself errors, simply stays hidden until the
+      // operator fixes the credential and re-tests it manually.
+      isActive: false,
       testStatus: testStatus || "unknown",
     });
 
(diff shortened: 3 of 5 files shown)


diegosouzapw:
Obrigado pela investigação detalhada — o diagnóstico (isActive:true mascarando uma chave nunca testada) é real e bem explicado. Antes de mergear, dois pontos:

1. **Teste automatizado**: esta é uma mudança de comportamento com blast radius alto (default de `isActive` invertido para toda conexão nova criada via `POST /api/providers`, mais o novo auto-test fire-and-forget). O projeto exige teste automatizado para toda mudança de código de produção (Hard Rule #8) — você mesmo nota que não conseguiu validar contra um provider real. Poderia adicionar cobertura para o novo fluxo (`POST /api/providers` → `isActive:false` → auto-test dispara → ativa em sucesso/unsupported, permanece inativo em falha)?

2. **Branch alvo**: o PR mira `main` diretamente, mas o fluxo de desenvolvimento ativo deste projeto passa pelas branches `release/vX.Y.Z` (atualmente `release/v3.8.51`, já que `release/v3.8.50` está em freeze de release). Foi intencional mirar `main`, ou prefere retargetar para `release/v3.8.51`?

Fico à disposição para revisar assim que os testes forem adicionados.

diegosouzapw:
Nice surgical approach to #11446 — starting inactive and letting any passing test flip activation ON (never OFF) is the right conservative direction, and the explicit scope note about the ~20 OAuth import flows is honest and appreciated. Items needed before merge: (1) please retarget to release/v3.8.51 (active development branch) — note [id]/test/route.ts has drifted there (the quota-clear-on-valid-probe fix rewrote the same updateData literal), so a rebase resolving that hunk is required; the two behaviors are compatible, keep both. (2) The PR currently ships zero tests while changing a default behavior — per the project's test rule please add unit coverage for: POST creates isActive:false; a passing test activates; a failing test leaves isActive untouched; a skipped/unsupported test activates. (3) Worth stating in the body that programmatic/API-created connections will be briefly invisible from /v1/models until auto-test completes, and stay hidden if creation-time auto-test fails transiently — the existing batch-activate endpoint is the operator escape hatch.

alltomatos:
Thanks for both reviews — addressed everything:

1. **Tests added** (`tests/unit/verified-connection-activation-11446.test.ts`, deterministic and network-stubbed, no real provider calls): `POST /api/providers` creates `isActive:false`; a passing test activates; a failing test leaves it untouched; and a skipped/unsupported test (unregistered provider, no network needed) also activates — while confirming a *different* skip reason, an exclusive-lease-busy connection, correctly does **not** activate, since that's "busy, try later," not "unverifiable" (see `tests/unit/exclusive-lease-connection-test-isolation.test.ts`).

2. **Rebased onto `release/v3.8.51`** and resolved the drifted hunk in `[id]/test/route.ts` you flagged — the quota-clear-on-valid-probe fix's `updateData` literal and this PR's `isActive` activation now compose: `testStatus`/error-clearing still follows `clearErrorState`, and `isActive` only ever flips on on a passing probe. Also had to update the existing `provider-journey.contract.test.ts` (#8330), whose STEP 2 asserted immediate `/v1/models` exposure for a connection created against an unreachable stub host — no longer valid once new connections start inactive, so it now simulates a passed test the same way STEP 3 already bypasses that host for the sync step.

3. **PR body updated** with the operator-visibility note on the creation→auto-test window and the batch-activate/re-test escape hatch.

`npx eslint` and `npm run typecheck:core` clean on all touched files; both test files pass locally (4/4 and 8/8, 6 skipped opt-in live steps). Ready for another look whenever you have time.

PR #11495, opened 2026-08-25
feat(providers): verify-only health sweep for web-cookie connections (#11488)

### LINKED ISSUE #11488: feat(resilience): extend token health-check sweep to web-cookie connections
## Gap

The proactive token health-check scheduler (`src/lib/tokenHealthCheck.ts`) only sweeps OAuth connections:

``\ts
// tokenHealthCheck.ts:445
const connections = await getProviderConnections({ authType: "oauth" });
```

Web-cookie connections (chatgpt-web, claude-web, grok-web, qwen-web, gemini-web, notion-web, ...) are stored with `auth_type = 'cookie'` (`src/lib/db/providers.ts:490` dedup path) and are therefore **never iterated by the sweep**. Even if they were, they carry no `refreshToken`, so `checkConnection()` hits the `:623` no-refresh-token branch and returns at `:744` without doing anything — and `:761`'s "refresh unsupported" stamp is unreachable for them.

Net effect: a dead cookie keeps `testStatus = "active"` until a live user request fails with 401/403. Bespoke proactive checkers exist only for Cursor (`tokenHealthCheckCursor.ts`), Kimi Web (`tokenHealthCheckKimi.ts`) and Copilot sub-tokens — every other cookie provider has none.

## Proposal

Extend the existing sweep with a **verify-only** path for cookie connections (no refresh involved):

1. Include `{ authType: "cookie" }` rows in `sweep()`.
2. New leaf module `tokenHealthCheckWebCookie.ts` (same pattern as the Cursor/Kimi leaves): when the provider is in `WEB_COOKIE_PROVIDERS`, probe it by reusing `validateProviderApiKey({ provider, apiKey, providerSpecificData })` from `src/lib/providers/validation` — i.e. exactly what the dashboard Test button runs, including proxy-fallback transport and per-provider specialty validators.
3. Result mapping:
   - valid → stamp `lastHealthCheckAt` (silent steady-state, per the Copilot log-noise precedent);
   - 401/403 / `AUTH_007 SESSION_EXPIRED` → `testStatus: "expired"`, `errorCode: "session_expired"`;
   - network / 5xx / SSRF-block / `unsupported` → stamp only, **never** flip state (transient must not terminal-state a healthy cookie).
4. Per-connection `healthCheckInterval` continues to be honored (default 60 min; 0 = disabled).

This mirrors the proactive account-health management sub2api does for subscription accounts, and turns silent cookie rot into an explicit dashboard state before traffic hits it.

Happy to implement this as a PR targeting `release/v3.8.51` if you agree with the approach.

### PR DESCRIPTION
**Pull request by oyi77**
Closes #11488.

## What

The token health-check scheduler only iterated `auth_type = 'oauth'` rows, and its no-refresh-token branch no-opped anyway (`tokenHealthCheck.ts` :649/:770) — so every web-cookie connection (chatgpt-web, claude-web, grok-web, qwen-web, gemini-web, …) stayed `testStatus: "active"` until a live user request failed with 401/403.

This PR adds a **verify-only** path to the existing sweep:

- `sweep()` now also fetches `{ authType: "cookie" }` rows.
- New leaf `src/lib/tokenHealthCheckWebCookie.ts` (same injectable-DI pattern as `tokenHealthCheckKimi.ts` / `tokenHealthCheckCursor.ts`) probes catalogued cookie providers via `validateWebCookieProvider` — the generic session-ping the dashboard Test button falls back to, proxy-fallback transport included.
- Result mapping is deliberately conservative:
  - valid → silent `lastHealthCheckAt` stamp (no per-tick log noise);
  - unambiguous `AUTH_007 / SESSION_EXPIRED` → terminal `expired` with `errorCode: session_expired`, `lastErrorSource: webcookie`;
  - network / guard-block / 5xx / `unsupported` → stamp only, **never** a state flip.
- Per-connection `healthCheckInterval` honored (default 60 min, 0 = off). `kimi-web` keeps its bespoke refresh leaf with precedence; no behavior change for OAuth rows.

The richer per-provider specialty validators are intentionally not used here: a background sweep that can terminal-state a connection must act only on an unambiguous signal.

## Verification

- New unit suite `tests/unit/token-health-check-webcookie.test.ts`: **9/9 pass** (`npx tsx --test --test-force-exit tests/unit/token-health-check-webcookie.test.ts`) covering all branches — candidate detection, non-cookie passthrough, valid-stamp, expired-flip field mapping, unsupported/inconclusive stamp-only, interval gate, missing credential, providerSpecificData.cookie fallback.
- Regression: `tests/unit/token-health-check-kimi.test.ts` 4/4 pass.
- `npx tsc --noEmit -p tsconfig.json`: zero diagnostics in the three touched files.
- `npm run check:file-size`: OK.
- ESLint: both new files clean. One pre-existing error remains in `src/lib/tokenHealthCheck.ts` (`@/lib/localDb` restricted-import barrel) — reproduced on the pristine base commit before my change, untouched here.

## Changed files

- `src/lib/tokenHealthCheckWebCookie.ts` (new)
- `src/lib/tokenHealthCheck.ts` (+20 wiring lines)
- `tests/unit/token-health-check-webcookie.test.ts` (new)

### DIFF
--- a/src/lib/tokenHealthCheck.ts
+++ b/src/lib/tokenHealthCheck.ts
@@ -30,6 +30,10 @@ import { isAutomatedTestProcess } from "@/shared/utils/testProcess";
 import { refreshGithubCopilotSubTokenIfNeeded } from "@/lib/tokenHealthCheckCopilot";
 import { checkCursorConnectionIfNeeded } from "@/lib/tokenHealthCheckCursor";
 import { checkKimiWebConnectionIfNeeded } from "@/lib/tokenHealthCheckKimi";
+import {
+  checkWebCookieConnectionIfNeeded,
+  isWebCookieHealthProbeCandidate,
+} from "@/lib/tokenHealthCheckWebCookie";
 
 const LOG_PREFIX = "[HealthCheck]";
 const TRUE_ENV_VALUES = new Set(["1", "true", "yes", "on"]);
@@ -442,7 +446,12 @@ export async function sweep(): Promise<number> {
   }
   state.sweeping = true;
   try {
-    const connections = await getProviderConnections({ authType: "oauth" });
+    const connections = [
+      ...(await getProviderConnections({ authType: "oauth" })),
+      // #11488: web-cookie rows (auth_type 'cookie') were never swept — a dead
+      // cookie stayed "active" until a live request failed against it.
+      ...(await getProviderConnections({ authType: "cookie" })),
+    ];
 
     if (!connections || connections.length === 0) return 0;
 
@@ -614,6 +623,22 @@ export async function checkConnection(conn) {
       log,
       logWarn,
       logError,
+      getConnectionLogLabel,
+      logPrefix: LOG_PREFIX,
+    });
+    return;
+  }
+
+  // Generic web-cookie verify-only probe (#11488): every catalogued cookie
+  // provider without its own bespoke leaf above. Kimi-web already returned.
+  if (isWebCookieHealthProbeCandidate(conn.provider)) {
+    const now = new Date().toISOString();
+    await checkWebCookieConnectionIfNeeded({
+      conn,
+      now,
+      intervalMin,
+      log,
+      logWarn,
       getConnectionLogLabel,
       logPrefix: LOG_PREFIX,
     });
--- a/src/lib/tokenHealthCheckWebCookie.ts
+++ b/src/lib/tokenHealthCheckWebCookie.ts
@@ -1,0 +1,133 @@
+import { WEB_COOKIE_PROVIDERS } from "@/shared/constants/providers";
+import { validateWebCookieProvider } from "@/lib/providers/validation/webCookie";
+import { updateProviderConnection } from "@/lib/db/providers";
+
+/**
+ * Verify-only background probe for web-cookie connections (#11488).
+ *
+ * The token health-check sweep historically iterated ONLY `auth_type = 'oauth'`
+ * rows, and even for those a missing refresh-token short-circuits into a no-op.
+ * Web-cookie connections (`auth_type = 'cookie'`) therefore kept
+ * `testStatus = "active"` forever — a dead cookie surfaced only when a live
+ * user request failed with 401/403.
+ *
+ * This leaf mirrors `tokenHealthCheckKimi.ts` / `tokenHealthCheckCursor.ts`:
+ * pure decision + persistence, with every side-effecting function injectable
+ * so tests can drive outcomes without network or DB.
+ *
+ * Probe semantics reuse `validateWebCookieProvider` — the same generic
+ * session-ping the dashboard Test button falls back to — because its result
+ * contract is deterministic: 401/403 → `{ error: "SESSION_EXPIRED",
+ * errorCode: "AUTH_007" }`, registry-less providers → `unsupported: true`.
+ * The richer per-provider specialty validators are deliberately NOT used here:
+ * a background sweep that can terminal-state a connection must only act on an
+ * unambiguous signal, and must treat every ambiguous failure (network blip,
+ * SSRF-guard block, 5xx) as "unknown" — stamp the tick, change nothing.
+ */
+
+/** Minimal structural view of a connection row this prober reads. */
+export interface WebCookieProbeConnection {
+  id: string;
+  provider?: string | null;
+  apiKey?: string | null;
+  providerSpecificData?: unknown;
+  lastHealthCheckAt?: string | null;
+}
+
+type LogFn = (...msg: unknown[]) => void;
+
+/** True when `provider` is a catalogued web-cookie provider this module can probe. */
+export function isWebCookieHealthProbeCandidate(provider: unknown): boolean {
+  const key = String(provider || "").toLowerCase();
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by oyi77** [comment]
CI triage for the current red batch (Docs Gates / unit shards / Vitest / ESLint-suppressions): reproduced at today's merge ref (`a179ffed5` + this branch) with **none of this branch's content** — `check:docs-counts` fails STRICT on README/AGENTS/llm.txt still saying "159 migrations" while code has 160, and `tests/unit/agent-card-route.test.ts` + `tests/unit/providers-constants-split.test.ts` fail at the tip itself. Inherited tip-drift, not introduced here.

This branch's own receipts stay green: `token-health-check-webcookie.test.ts` 9/9, Kimi health-check regression 4/4, tsc clean on touched files.

**Comment by diegosouzapw** [comment]
Held out of this merge batch — found a real defect while validating (reproduced 3/3 in isolation, not present on the pure release/v3.8.51 tip since this is entirely new code):

\`honors the interval gate — recently checked rows are skipped silently\` fails with \`probed === true\` (expected \`false\`).

Root cause: \`checkWebCookieConnectionIfNeeded\` (\`src/lib/tokenHealthCheckWebCookie.ts:85\`) gates on \`Date.now() - lastCheckMs < intervalMin * 60 * 1000\` — using the real wall clock — while the test fixes \`NOW = "2026-08-25T12:00:00.000Z"\` and derives \`lastHealthCheckAt\` as 5 minutes before that constant. Whenever the suite runs more than \`intervalMin\` away from that hardcoded timestamp (verified against real UTC just now: \`2026-08-25T20:42:45Z\`, an 8h42m gap vs the 60-minute interval under test), the gate never trips and the test fails — a timebomb, same class as the GLM weekly-reset test #11450 just de-timebombed.

Two ways to close it, your call:
1. Thread the injected \`params.now\` through the gate comparison instead of \`Date.now()\` — makes the function fully deterministic/testable and is probably the more correct shape given the function already takes \`now\` as a sweep-consistency timestamp.
2. Keep \`Date.now()\` in the source (arguably fine for a real background sweep) and de-timebomb the test by computing \`lastHealthCheckAt\` relative to \`Date.now()\` instead of the fixed \`NOW\` constant.

Everything else in the PR looks solid — 8/9 in this file, the rest of the suite (expired-flip mapping, unsupported/inconclusive stamp-only, missing-credential, providerSpecificData.cookie fallback, kimi-web regression) passes clean. Happy to take another look once this is fixed.

**Comment by oyi77** [comment]
Fixed — took your option 1: the interval gate now compares against the injected `now` instead of the wall clock (`8c1fd1ce3`, single-line change):

```ts
if (lastCheckMs > 0 && new Date(now).getTime() - lastCheckMs < intervalMin * 60 * 1000) return true;
```

Production semantics are unchanged — the sweep caller (`tokenHealthCheck.ts:635`) already passes a fresh `new Date().toISOString()` as `now` — but `checkWebCookieConnectionIfNeeded` is now fully deterministic under injected time, which is the right shape for a function that already takes `now` for stamping. (The committed diff also carries a few prettier line-wraps from the repo's pre-commit hook on pre-existing lines in this file — no behavior change.)

Receipts on `8c1fd1ce3`:

- `tests/unit/token-health-check-webcookie.test.ts`: **9/9 pass**, including `honors the interval gate — recently checked rows are skipped silently`. Determinism proof is inherent: that test's fixed `NOW` (2026-08-25T12:00Z) is >30h from real wall clock at run time, and the gate now trips correctly regardless.
- Sibling regressions: `token-health-check-kimi.test.ts` + `token-health-check-sweep.test.ts` → **7/7 pass**.
- ESLint clean on the touched file; full-project `tsc --noEmit` produces zero diagnostics referencing `tokenHealthCheckWebCookie.ts` (the advisory Build check's error mass is pre-existing and unrelated).

Ready for re-review.

oyi77:
Follow-up covering the repo's PR gates (copilot-instructions.md):

**Repro test** — your reported defect *was* encoded as a test (the failing `honors the interval gate` case), and the fix makes it pass deterministically. Added its complement in `4cf955ce8`: `stale rows are re-probed once the interval has elapsed` (checked 61 min ago vs 60-min interval → probed exactly once, re-stamped) so both boundaries of the now-deterministic gate are locked.

**Coverage gate** (`npm run test:coverage`, full suite, 34,377 tests) — `--check-coverage` thresholds all pass globally:

| Metric | Result | Floor |
|---|---|---|
| Statements | 76.18% | 60% |
| Branches | 79.09% | 60% |
| Functions | 71.03% | 60% |
| Lines | 76.18% | 60% |

Touched leaf module (`src/lib/tokenHealthCheckWebCookie.ts`, isolated c8 over its suite): **statements 100%, functions 100%, lines 100%, branches 82.75%**.

The npm script exits 1 solely because of 28 suite failures that pre-date this branch's content (identical count across two independent full runs, before and after the new test; zero involve `tokenHealthCheckWebCookie` or the sweep wiring — those pass 10/10 and kimi+sweep 7/7). Same inherited-tip-drift class triaged in my earlier comment. Happy to produce a named-failure ledger if useful, but it needs another ~35-min instrumented run, so flagging rather than assuming you want it.

Branch tip: `4cf955ce8` (fix `8c1fd1ce3` + test `4cf955ce8`).

PR #6193, opened 2026-07-04
fix(chatcore): exempt opencode client from the default 128-tool truncation

DKotsyuba:
## Problem

When [opencode](https://opencode.ai) is pointed at OmniRoute as its provider, it acts as an MCP/agent host and sends a **large tool list** on every request — its built-in tools plus every configured MCP server's tools. In a real setup this is easily **300+ tools**, and the built-in `task` tool (the one opencode uses to launch subagents) sits well past the first 128 entries.

OmniRoute truncates the upstream tool list to `MAX_TOOLS_LIMIT` (**128**) via a blind `tools.slice(0, 128)` in `truncateToolList`. Because the slice keeps only the first 128 tools, `task` and a large part of the MCP tools are **silently dropped** before the request reaches the provider.

Observed end-to-end (4-stage pipeline capture, opencode → codex):

| Stage | `tools` count | `task` present? |
| --- | --- | --- |
| client → OmniRoute (`clientRawRequest`) | 328 | ✅ yes |
| OmniRoute → provider (`providerRequest`) | 128 | ❌ no |

The model then correctly reports that it has no `task` tool, so **models routed through OmniRoute cannot spawn subagents**, and part of the MCP tools are unreachable for the same reason. This is not a function-calling bug — direct single-tool requests return `tool_calls` fine; the only defect is the count cap combined with a position-blind slice.

## Why the cap exists (and why we can't just remove it)

The 128 cap was introduced (#1687) to avoid upstream `400 Bad Request` responses from providers that reject requests with too many tools, and later grew a dynamic detector (#2061) plus a proactive per-provider map (#5563, e.g. `grok-cli: 200`). Those are **real provider ceilings** — blindly removing the cap would reintroduce `400`s for such providers. So the fix must keep the guard where a real limit exists, and only relax the **speculative default** for a client that is known to want its full tool list.

## Fix

Detect the opencode client and let it bypass **only the speculative 128 default** — never a known provider ceiling.

- **Detection** — new `isOpencodeClient(headers, userAgent)` in `chatCore/requestFormat.ts` (mirrors the existing `isCopilotClient`): matches any `x-opencode-*` header key (case-insensitive), or `"opencode"` in the user-agent parameter / `user-agent` header value. Exposed as `isOpencodeClient` on `resolveChatCoreRequestFormat`.
[...]

PR #11496, opened 2026-08-25
fix(providers): remove executor-level tool truncation in OpencodeExecutor

### LINKED ISSUE #11444: fix(providers): OpencodeExecutor still truncates tools to 128 — PR #6193 only fixed the chatCore layer
### OmniRoute Version

3.8.49 (also affected: current `main`)

### Installation Method

Docker / Docker Compose

### Provider(s) Involved

OpenCode Free (`oc/*` models)

### Client Tool

opencode 1.18.21 (also reproduced with plain curl)

### Description

Follow-up to **#9132**, closed as fixed by **#6193**. That PR only removed the chatCore-layer truncation (`truncateToolList()` in `chatCore/upstreamBody.ts`). A second, unconditional cap remains in `OpencodeExecutor.transformRequest`:

```ts
// open-sse/executors/opencode.ts (still present on main, ~line 837)
if (Array.isArray(mb.tools) && mb.tools.length > 128) {
  mb.tools = mb.tools.slice(0, 128);
}
```

Every request to provider `opencode` passes through it — regardless of the opencode-client detection added in #6193. Since opencode sends tools alphabetically, everything after the first 128 entries is silently dropped: `read`, `skill`, `task`, `task_cancel`, `task_message`, `task_result`, `task_revive`, `todowrite`, `wait_for_user`, `webfetch`, `write`. Subagent spawning (`task`) is therefore impossible on all `oc/*` routes.

Introduced in a16e47c59 (#1699, v3.7.2) — hidden inside a user-agent refresh commit.

### Steps to Reproduce

1. Send a chat-completions request to an `oc/*` model via OmniRoute with 150+ tools; put the target tools at the end of the array.
2. Model reports them as absent. The identical request sent directly to `opencode.ai/zen/v1/chat/completions` returns them as present.

Confirmed with real opencode 1.18.21 sessions and plain curl. Raising the executor cap locally fixes it completely.

### Suggested fix

Remove the executor-level slice (or gate it behind the same detection as #6193) and rely on the centralized `truncateToolList()` — otherwise fixes to one layer get silently undone by the other.

Prajeeth-12:
Fixes #11444

## What this PR does

Removes the unconditional `tools.slice(0, 128)` in `OpencodeExecutor.transformRequest` (introduced in a16e47c59, #1699, v3.7.2). This executor-level cap silently dropped every tool past index 128, making late-alphabetical tools — including `read`, `skill`, `task`, `task_cancel`, `task_message`, `task_result`, `task_revive`, `todowrite`, `wait_for_user`, `webfetch`, `write` — unreachable on all `oc/*` routes. Since opencode sends tools alphabetically, subagent spawning (`task`) was completely broken.

## Why

PR #6193 fixed the **chatCore layer** truncation by adding a `bypassDefaultToolLimit` flag, but it did not touch the **executor layer**. The centralized `truncateToolList()` in `chatCore/upstreamBody.ts` already handles per-provider tool limits (via `toolLimitDetector.ts`) with proper bypass detection for opencode clients. The executor-level slice was both redundant and unreachable by the bypass flag — a second cap that undid #6193's fix for any request routed through `OpencodeExecutor`.

## Changes

- `open-sse/executors/opencode.ts`: removed the 6-line block that unconditionally truncated `mb.tools` to 128 entries

## Test plan

- [x] Regression test `tests/unit/opencode-executor-no-tool-truncation.test.ts` — 3 assertions:
  - 160 tools pass through without truncation
  - 128 tools preserved exactly (boundary)
  - 250 tools pass through (stress, simulating real opencode sessions)
- [x] All 3 tests pass after rebase onto current `release/v3.8.51` tip

## Reproduction (from issue)

```bash
# Send 150+ tools to an oc/* model via OmniRoute — tools at index ≥128 are absent from upstream
curl -X POST http://localhost:20128/v1/chat/completions   -H "Content-Type: application/json"   -d '{"model":"oc/deepseek-chat","messages":[{"role":"user","content":"hi"}],"tools":[...150+ tools...]}'
# With this fix: all tools reach upstream intact
```

⚠️ base-red inherited: #11449

### DIFF
--- a/open-sse/executors/opencode.ts
+++ b/open-sse/executors/opencode.ts
@@ -887,12 +887,6 @@ export class OpencodeExecutor extends BaseExecutor {
     }
     if (modifiedBody && typeof modifiedBody === "object" && !Array.isArray(modifiedBody)) {
       const mb = modifiedBody as Record<string, unknown>;
-      if (Array.isArray(mb.tools) && mb.tools.length > 128) {
-        mb.tools = mb.tools.slice(0, 128);
-      }
-    }
-    if (modifiedBody && typeof modifiedBody === "object" && !Array.isArray(modifiedBody)) {
-      const mb = modifiedBody as Record<string, unknown>;
       const parsed = parseEffortLevel(model);
       if (parsed) {
         mb.model = parsed.baseModel;
--- a/tests/unit/opencode-executor-no-tool-truncation.test.ts
+++ b/tests/unit/opencode-executor-no-tool-truncation.test.ts
@@ -1,0 +1,82 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { OpencodeExecutor } from "../../open-sse/executors/opencode.ts";
+
+test("OpencodeExecutor.transformRequest does not truncate tools beyond 128 (#11444)", () => {
+  const executor = new OpencodeExecutor();
+  const toolCount = 160;
+  const tools = Array.from({ length: toolCount }, (_, i) => ({
+    type: "function",
+    function: { name: `tool_${i}`, description: `Tool ${i}`, parameters: {} },
+  }));
+
+  const body = {
+    model: "deepseek-chat",
+    messages: [{ role: "user", content: "hello" }],
+    tools,
+  };
+
+  const result = executor.transformRequest(
+    "deepseek-chat",
+    body,
+    false,
+    { apiKey: "test-key" }
+  );
+
+  assert.equal(
+    result.tools.length,
+    toolCount,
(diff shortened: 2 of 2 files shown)


diegosouzapw:
Thanks for catching this — closing in favor of #11518, which already landed the identical fix and merged today: the unconditional `tools.slice(0, 128)` in `OpencodeExecutor.transformRequest` is gone on the current release/v3.8.51 tip. Confirmed by grep — no truncation slice remains in `open-sse/executors/opencode.ts`.

No distinct value remains to carry forward — same root cause, same fix shape. Appreciate the independent catch and the thorough write-up on why the executor-level cap undid #6193's chatCore-layer fix.

PR #11516, opened 2026-08-25
fix(cli): preserve existing admin password during setup unless explicitly overridden (#11494)

### LINKED ISSUE #11494: [BUG] CLI: any `omniroute setup` run silently resets the dashboard admin password to CHANGEME

HouMinXi:
### Summary
Fixes #11494: Running `omniroute setup` (e.g. `--add-provider`) on an existing installation silently reset the admin password to `CHANGEME` via the fallback `INITIAL_PASSWORD` in `.env`.

### Changes
- In `bin/cli/commands/setup.mjs::setupPassword`, checks whether `getSettings(db).password` already exists.
- If an existing password is present and the caller did not explicitly pass `--password`, setup preserves the existing password and prints `Using existing admin password` without re-hashing or overwriting.
- If `--password` is explicitly provided, it updates as expected.

### Test Plan
- Added `tests/unit/cli-setup-password-preservation.test.ts` (2 integration-style CLI tests with real SQLite instances).
- `npm run typecheck:core` clean.

Closes #11494

### DIFF
--- a/bin/cli/commands/setup.mjs
+++ b/bin/cli/commands/setup.mjs
@@ -41,10 +41,16 @@ async function resolvePassword(opts, prompt, nonInteractive) {
 }
 
 async function setupPassword(db, opts, prompt, nonInteractive) {
+  const existingSettings = getSettings(db);
+  if (existingSettings.password && !opts.password) {
+    updateSettings(db, { requireLogin: true });
+    printInfo("Using existing admin password");
+    return false;
+  }
+
   const password = await resolvePassword(opts, prompt, nonInteractive);
   if (!password) {
-    const settings = getSettings(db);
-    if (!settings.password) {
+    if (!existingSettings.password) {
       updateSettings(db, { requireLogin: false });
     }
     if (!nonInteractive) {
--- a/tests/unit/cli-setup-password-preservation.test.ts
+++ b/tests/unit/cli-setup-password-preservation.test.ts
@@ -1,0 +1,181 @@
+// ENVIRONMENT NOTE (sandbox better-sqlite3 / glibc limitation, not a code defect):
+// This test constructs a real better-sqlite3-backed SQLite database.
+// better-sqlite3 is a native addon; production and CI load it normally, but some
+// sandboxes/dev boxes ship a system glibc older than the prebuilt binary requires
+// ("GLIBC_2.29 not found"), so the native module fails to dlopen and any test that
+// reaches better-sqlite3 directly (or asserts stdout that the load-failure warning
+// would pollute) fails HERE while passing in CI. This is a known environment
+// limitation, not a defect in the code under test: the OmniRoute runtime itself
+// cascades to node:sqlite/sql.js when better-sqlite3 is unavailable. See
+// tests/unit/_helpers/betterSqlite3Availability.ts for a guard helper.
+import test from "node:test";
+import assert from "node:assert/strict";
+import { spawnSync } from "node:child_process";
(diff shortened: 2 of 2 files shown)


diegosouzapw:
Thanks for catching this — closing in favor of #11522, which already landed the same fix and merged earlier today: `resolvePassword()` in `bin/cli/commands/setup.mjs` now falls back to `INITIAL_PASSWORD` only when `!settings.password`, so an existing admin password survives subsequent `omniroute setup` runs unless `--password` is explicitly passed. Confirmed on the current release/v3.8.51 tip.

No distinct value remains to carry forward — same root cause, same fix shape. Appreciate the independent catch.

Issue #6915, opened 2026-07-11
feat: sort/filter Free Provider Rankings by auth type (surface NOAUTH first)


diegosouzapw:
## Feature Request

**Source:** Discussion #6746 by @manchairwang

## Problem

The Free Provider Rankings page (`/dashboard/free-provider-rankings`) sorts providers by model quality/availability (Score / Avg Score). Users looking specifically for "which providers actually give me tokens at zero cost and zero signup" cannot tell at a glance — that property is the `Type` column (NOAUTH / OAUTH / APIKEY), which is neither sortable nor filterable. In #6746 the reporter tried several top-ranked rows, got nothing free, and concluded the rankings were not useful, because the genuinely zero-setup **NOAUTH** rows (OpenCode Free, Augment) were buried mid-list.

## Proposed Solution

Add sort/filter controls on the rankings table keyed on the `Type` column:

- A filter (chips or dropdown): `All | NOAUTH | OAUTH | APIKEY`.
- Optionally a secondary sort that surfaces NOAUTH first, then OAUTH, then APIKEY, keeping the score order within each group.
- A short legend clarifying what each `Type` means for the user's effort (NOAUTH = zero setup; OAUTH = log in with your own account; APIKEY = bring your own key / provider's own free tier).

### Implementation Ideas

- The rankings UI lives under `src/app/(dashboard)/dashboard/free-provider-rankings/`; the `Type` value already comes through per row, so this is a client-side sort/filter on existing data (no new backend field expected — confirm the row shape first).

### Current Workarounds

Manually scan the `Type` column and pick the NOAUTH rows (currently OpenCode Free, Augment).

## Additional Context

- Discussion: #6746
- No dedicated docs page for the rankings semantics yet; a short "what NOAUTH/OAUTH/APIKEY mean" note would pair well with this.

PR #11546, opened 2026-08-25
feat(rankings): show what each provider actually served

maxmad64bis:
## Summary

- The Free Provider Rankings page ranked providers by ELO alone, which leaves a provider that answers every call with an error in first place. The API has served per-provider usage since #10926, but the page never asked for it.
- The page now always fetches `withUsage=1&usageRange=24h` and shows a Reliability column: share of calls answered over the last 24 h, colour-coded, with the sample size in the tooltip. A provider whose sample is too small for a rate — or that nobody called — shows a dash, never 0%: no data must not read as a bad score, and a broken provider must not read as healthy.
- One small server fix rides along: `withUsage` used to be ignored unless `configuredOnly`/`availableOnly` forced the connection snapshot to load, so the default view came back with no reliability at all. Requesting usage now pulls the snapshot it depends on (`needsConnectionSnapshot`), and filtering stays a no-op when neither filter is set, so unfiltered callers still get every provider.
- The page also drops its private copy of the ranking type, which had already drifted past the real one, and imports the shared type instead.

## Related Issues

- Builds directly on #10926 (usage served by the rankings API).

## Validation

- [x] Change type: UI / i18n / routing-API touch-up
- [x] Focused tests and category gates from the golden path: pure helper tests, page-level jsdom tests (fetch params + column rendering incl. the dash case), and the existing #6915 page harness re-run green; `typecheck:core` exit 0; prettier clean on all touched files
- [ ] `npm run lint`

Local caveat: my machine's node_modules is broken (`es-abstract` incomplete) and a borrowed install carries a newer `eslint-plugin-react-hooks` than this branch's lockfile pins (7.1.1 vs 7.0.1), flagging pre-existing #6915 code that the pinned version does not flag. Nothing introduced by this PR is implicated. CI is authoritative.

## Tests Added Or Updated

- `tests/unit/freeProviderRankings-usage-display.test.ts` — new: rate shown only when the sample supports one; no traffic renders as no-data, never 0%; tone thresholds; window travels with the rate; regression pin that `withUsage` alone pulls the connection snapshot.
- `tests/unit/ui/free-provider-rankings-page-usage.test.tsx` — new: every fetch carries `withUsage=1&usageRange=24h`; column renders `99%` for a supported sample and `—` without traffic.
- `tests/unit/ui/free-provider-rankings-page-authtype-6915.test.tsx` — unchanged, re-run green against the modified page.

## Coverage Notes

- New display logic lives in `src/lib/freeProviderRankingsUsage.ts` (pure, fully covered above); `src/lib/freeProviderRankings.ts` gained one exported predicate covered by a dedicated test case; the page component is covered by the jsdom tests above. i18n: 5 keys added to `en.json`, propagated as `__MISSING__` sentinels to the 42 other locales via the repo's sentinel convention (translation tooling picks them up).

## Reviewer Notes

- Behaviour change worth noting: the default page load now always pays the usage aggregate query over `call_logs`. That is the point of the PR (an ELO-only ranking was describing broken providers as healthy), but reviewers should confirm they are comfortable with the extra query on every page load.
- The Reliability column appears between Avg Score and Models; the top-3 podium cards are intentionally untouched (they carry no metric columns today).
- Unit suite on this exact base is currently red for an unrelated reason (159 vs 160 migrations). Fixed by sibling PR `docs/migrations-count-160`; merging that first keeps this PR's CI green.


abhisheksharma2411 (commented):
Good change, and the framing in the module docstring is the right one:

> A provider that answers every call with an error must not read as healthy, and a provider nobody called must not read as broken.

Deleting the page's private copy of `FreeProviderRanking` in favour of a type-only import is worth calling out separately — a duplicated interface that "had already drifted past the API" is a silent-wrongness generator, and the comment explaining *why* it must be type-only (DB modules wired at import time) will stop someone reintroducing it.

Two things in `freeProviderRankingsUsage.ts`, both about the same tension the docstring names.

## Rounding overstates reliability, which is the failure mode this module exists to prevent

`Math.round(successRate * 100)` displays a perfect score for imperfect providers:

```
249/250  = 99.60%  ->  displays 100%
199/200  = 99.50%  ->  displays 100%
996/1000 = 99.60%  ->  displays 100%
```

For a general-purpose percentage that is fine. For a *reliability* column whose stated job is that a provider erroring on calls must not read as healthy, `100%` is the one value that should be reserved for actually flawless — it is the number an operator scans for when deciding what to trust. `Math.floor` never overstates and costs nothing here; `Math.round` everywhere except a floor at the 100 boundary would also work if you prefer round-to-nearest generally.

## Two rows showing the same number get different colours

`tone` is derived from the raw `successRate` while `percent` is derived from the rounded one, so the thresholds land between adjacent displayed values:

```
successRate=0.7996  ->  displays 80%  tone=poor   (orange)
successRate=0.8000  ->  displays 80%  tone=fair   (yellow)
successRate=0.9496  ->  displays 95%  tone=fair   (yellow)
successRate=0.9500  ->  displays 95%  tone=good   (green)
```

Two providers both reading `80%`, one orange and one yellow, in the same table. The colour is doing real work on this page — it is the thing you scan before the number — so a reader will reasonably conclude the colours encode something the number does not, and go looking for it.

Deriving the tone from the same value you display (`percent >= 95` / `percent >= 80`) makes the column self-consistent, at the cost of a provider at 94.6% being called "good". If you would rather keep the raw thresholds, then the displayed number needs the precision to justify them — one decimal place — but self-consistency seems the better trade for a scan-first column.

## Checked and fine

- The `successRate === null` branch correctly splits `insufficient` (requests > 0) from `none`, and defers the "too few to compute" decision to the API rather than second-guessing it with a local threshold — the right place for it.
- `EMPTY` returning `windowHours: 0` is never surfaced, since `kind: "none"` routes to `reliabilityNoTraffic` which takes no arguments.
- Always sending `withUsage=1` / `usageRange=24h` rather than making it conditional matches the comment's reasoning, and pinning the range to the health matrix default keeps two surfaces from disagreeing about the same provider.

PR #11548, opened 2026-08-25
fix(sse): scale chat admission by ingest byte budget instead of a fixed request count

TheDemonTuan:
## Problem

Coding-agent fan-out (multiple subagents/CLIs hitting the proxy concurrently, often
sharing one API key) collapsed to an effective concurrency of ~1 on `/v1/messages`
and other chat routes. Requests hung, then retried, then received
`503 chat_admission_busy`.

Root cause: the byte-stage admission gate in `chatBodyAdmission.ts` capped
heavyweight requests at a fixed request **count**
(`OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT`, default `1`) with **no resource-pressure
awareness at all** — it rejected a second concurrent request even on a host with
ample free RAM.

## Fix

- **`src/shared/middleware/admissionBudget.ts` (new)** — derives an ingest byte
  budget from the process's real memory ceiling (V8 heap limit, or the tighter
  cgroup/container limit via `process.constrainedMemory()`), auto-scaling from a
  512 MB container to a multi-GB desktop with no env tuning required.
- **`ChatAdmissionController`** gains an additive, isolated byte-budget gate
  (`tryAcquireBudget`/`acquireBudgetWithin`) layered on top of the existing
  count-based gate. The legacy count cap now binds only when
  `OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT` is explicitly set; left unset (the
  default), the byte budget is the effective gate.
- **`admitChatRequest`** and **`admitChatStructure`** now consult live
  multi-signal resource pressure (`open-sse/utils/resourcePressure.ts`): normal
  pressure admits within budget with a short bounded wait, high pressure honors
  the full queue wait, and critical pressure sheds immediately with a distinct
  `503 resource_pressure` response before any bytes are ingested.
- Every existing `ChatAdmissionController`/`PerConnectionAdmissionController`
  test constructs its own controller with the legacy positional constructor
  args (defaulting to an unlimited byte budget), so this is fully additive —
  those tests are unaffected without modification.
- Surfaces the new state at `GET /api/monitoring/health` → `chatAdmission`
  (`inflightBytes`, `maxInflightBytes`, `budgetSource`, `pressureSeverity`,
  `countCapEnabled`) so the fix is observable without guessing.
- Updates `ENVIRONMENT.md`, `.env.example`, `admission-lanes.md`,
  `TROUBLESHOOTING.md`, `DOCKER_GUIDE.md`, and `VM_DEPLOYMENT_GUIDE.md`, which
  previously told operators to raise/lower the fixed count cap by hand.

## Tests

- `tests/unit/agent-fanout-admission-regression.test.ts` — reproduces the
  reported fan-out 503 storm under the old count=1 gate (1 admitted / 7 shed
  out of 8 concurrent requests), then proves it is gone under the new
  byte-budget gate (all 8 admitted), plus a critical-pressure shed case.
- `tests/unit/admission-ingest-budget.test.ts` — covers the budget formula
  across host profiles (bare-metal, container-constrained, override).
- `tests/unit/admission-pressure-gate.test.ts` — covers normal/high/critical
  wait behavior and timing.
- `tests/unit/per-connection-admission-9654.test.ts` — updated two tests that
  simulated "capacity busy" via the legacy count gate to instead exhaust the
  byte budget, matching the new capacity model.
- `tests/unit/observability-payloads.test.ts` — extended for the new
  allowlisted health-payload fields.

## Verification run on this branch

- `node --import tsx/esm --test` across the full admission test family
  (14 files): **129/129 pass**
- Adaptive-admission (L2) test family (13 files): **121/121 pass**, confirming
  zero impact on that layer
- `npm run test:vitest`: **405/405 pass**
- `npm run typecheck:core`: 0 errors
- `npm run lint` on all changed files: 0 new violations (one pre-existing
  baselined `no-unused-vars` entry, untouched by this diff)
- `check-docs-sync`, `check-env-doc-sync`, `check-tracked-artifacts`,
  `check:any-budget:t11`: all pass

Two failures found while running the wider suite were confirmed pre-existing
and unrelated (`git diff` against these files is empty on this branch):
`monitoring-health-cache.test.ts` (stale test calling `GET()` with no
arguments against a route that already requires a `request` param per an
earlier merged GHSA security fix) and
`responses-route-early-keepalive-wiring.test.ts` (a Windows temp-dir
file-lock race in test teardown).

## Review follow-up

- Requests whose body alone exceeds the effective ingest budget now fail immediately with `413 body_exceeds_budget`; they no longer occupy a fairness lane or wait for capacity that can never arrive.
- `OMNIROUTE_CHAT_MAX_INFLIGHT_BYTES` overrides now clamp to the documented 8 MiB–2 GiB safe range.
- True contention remains distinct and retryable as `503` / `inflight_bytes_budget`.
- Byte accounting, response construction, session identity, and structural estimation were extracted into focused modules so `chatBodyAdmission.ts` satisfies the frozen file-size gate.

## Follow-up validation

- `node --import tsx/esm --test` on 7 focused admission/observability files: passed
- Changed-file ESLint: passed
- `npm run check:file-size -- --base-ref upstream/release/v3.8.51`: passed
- `npm run typecheck:noimplicit:core`: blocked by pre-existing `pureHelpers.ts` implicit-any diagnostics on the base
- `npm run check:docs-all`: blocked by pre-existing missing `.env.example` entry `OMNIROUTE_CORPUS_CACHE_SIZE` on the base
- `npm run check:mutation-test-coverage`: blocked by 3 pre-existing Stryker mappings on the base

⚠️ base-red inherited: #11449


abhisheksharma2411 (commented):
Right diagnosis and the right shape of fix — a request-count cap defaulting to 1 was never going to survive coding-agent bodies, and deriving the budget from the real memory ceiling rather than a constant is the lesson from the v3.8.8 outage applied properly. Keeping the byte layer completely disjoint from `#activeHeavy`/`#queues`, with `maxInflightBytes` defaulting to `MAX_SAFE_INTEGER`, is what makes this reviewable at all.

One finding I think needs addressing before merge.

## A body larger than the whole budget burns the full admission timeout instead of failing fast

`tryAcquireBudget` rejects on `#inflightBytes + charge > maxInflightBytes`, which is correct — but it is also the *only* check, so a request whose `charge` alone exceeds `maxInflightBytes` returns `null` **on a completely idle gate**. `acquireBudgetWithin` then treats that exactly like contention: it parks the caller in the fairness queue and waits for capacity that cannot arrive, until the deadline expires and it sheds as `inflight_bytes_budget`.

This is reachable on default settings, not a contrived case:

```
CHAT_HARD_MAX_BODY_BYTES   default = 50 MiB   ← body accepted this far
MIN_INGEST_BUDGET_BYTES            =  8 MiB   ← budget floor on a small host
```

The derived budget only exceeds the 50 MiB hard cap once the effective ceiling passes ~1.6 GB (`ceiling × 0.25 / 8 > 50 MiB`). So on any host below that — a 512 MB container being exactly the case this PR is trying to serve — there is a band of body sizes that the hard cap admits and the byte budget can never satisfy. Every one of those requests waits the full admission timeout, occupies a fairness lane while doing so, and is then shed with a reason that reads as "server busy" rather than "body exceeds budget".

An early return would fix both the latency and the diagnosis:

```ts
tryAcquireBudget(bytes: number): ChatAdmissionLease | null {
  const charge = Math.max(0, Math.floor(bytes));
  if (charge > this.maxInflightBytes) return null;   // unservable at any concurrency
  …
}
```

with `acquireBudgetWithin` distinguishing that case so it sheds immediately under its own reason (`body_exceeds_budget`) instead of parking. Right now the two are indistinguishable in `shedsByReason`, which matters because the operator responses are opposite: one says add capacity, the other says raise `OMNIROUTE_CHAT_MAX_INFLIGHT_BYTES` or lower `OMNIROUTE_CHAT_HARD_MAX_BODY_BYTES`.

## Smaller: `override` bypasses both clamps

```ts
const override = parsePositiveFinite(input.override);
if (override !== null) {
  const bytes = Math.floor(override);
  return { bytes, source: "override", effectiveCeilingBytes: bytes };
}
```

`MIN_INGEST_BUDGET_BYTES` / `MAX_INGEST_BUDGET_BYTES` guard the derived path but not this one, so `OMNIROUTE_CHAT_MAX_INFLIGHT_BYTES=1024` is accepted and wedges every real request into the timeout-then-shed path above — with no startup warning, and `countCapEnabled: false` in the health summary reassuring the operator that the byte gate is the binding one. Clamping the override to the same bounds (or at least logging when it is below `MIN`) makes the knob hard to hold wrong.

## Things I checked that are fine

- The `source` selection is correct in both directions: `constrained <= heapLimit → "cgroup"`, otherwise `"v8_heap"`, and `ceiling` picks the tighter of the two.
- An invalid/absent `heapSizeLimitBytes` falls back to `MIN_INGEST_BUDGET_BYTES`, which then re-clamps to the floor rather than producing a nonsense budget.
- `readConstrainedMemoryBytes` treating a missing `process.constrainedMemory` as "no cgroup limit" is the right default for macOS/Windows hosts, and it is guarded.
- Caching in `resolveIngestByteBudget` with an explicit `reloadIngestBudgetForTests` seam keeps the hot path free of V8/cgroup reads without making the value untestable.

The floor deserves one line of comment, incidentally: on a host small enough for `raw` to fall below 8 MiB, the floor *overrides* the safety derivation — 8 MiB of raw ingest at `INGEST_AMPLIFICATION = 8` implies ~64 MiB transient, which on a 128 MB container is half the box. That is a deliberate liveness-over-safety trade and worth saying so, since every other number here is derived.

TheDemonTuan:
Addressed the review findings in `b6bda09c2`:

- requests whose charge exceeds the whole ingest budget now return immediately as `413 body_exceeds_budget`; they do not enter a fairness queue or consume the admission timeout
- explicit `OMNIROUTE_CHAT_MAX_INFLIGHT_BYTES` values clamp to the documented 8 MiB–2 GiB range
- true contention remains separately counted as `inflight_bytes_budget` and retains the bounded-wait behavior
- regression coverage now distinguishes impossible bodies from contention under normal/high pressure
- the 8 MiB floor trade-off is documented

I also modularized the admission helpers so `chatBodyAdmission.ts` passes the frozen file-size gate.

PR #11563, opened 2026-08-25
feat(providers): add 1min.ai provider

zero-executioner:
﻿## Summary

Adds **1min.ai** (docs.1min.ai) as a provider. It's a chat aggregator with a genuinely custom, non-OpenAI-compatible API:

- One `promptObject.prompt` string per request instead of a `messages` array (multi-turn context is normally carried server-side via `promptObject.conversationId`, which requires a prior `POST /api/conversations` call and a stable conversation identity that stateless OpenAI-compatible clients don't provide)
- Real Server-Sent Events (`event:`/`data:` framing, event types `content`/`result`/`done`/`error`) instead of raw text deltas or OpenAI-shaped chunks
- An `API-KEY` auth header instead of `Authorization: Bearer` (confirmed from every literal code sample on their docs page — their own docs intro page says `Authorization: Bearer` elsewhere, which is incorrect for this endpoint)

`OneMinAiExecutor` translates both directions: flattens OpenAI message history into a labeled prompt transcript, translates their real SSE stream into standard `chat.completion.chunk` events, unwraps their nested `aiRecord.aiRecordDetail.resultObject` for non-streaming JSON, and maps errors through `buildErrorBody`. Registered under both `oneminai` and `1min` aliases.

Tool-calling, vision, and `response_format` are declared as `unsupportedParams` since the translation doesn't wire up 1min.ai's `attachments`/Asset API — flagged honestly rather than silently dropped.

**Also fixes** a pre-existing Windows-only bug in `executor-map-golden.test.ts`: `new URL(import.meta.url).pathname` produces a leading-slash path (`/C:/Users/...`) that double-prefixes the drive letter when resolved (`C:\C:\Users\...`), which made the golden snapshot test unable to run *at all* on Windows — confirmed by reproducing the identical failure on a clean checkout with none of this PR's changes applied. Fixed by switching to `fileURLToPath`, matching the existing pattern already used in `tests/helpers/goldenSnapshot.ts`.

## Test plan

- [x] New unit tests (`tests/unit/executor-oneminai.test.ts`): registration, prompt-flattening (single message + multi-turn), non-streaming JSON translation, SSE streaming translation, error mapping — 6/6 pass
- [x] **Live end-to-end validation against the real API** with a real 1min.ai key, exercising the actual shipped `OneMinAiExecutor` class (not a mock): non-streaming, streaming, and multi-turn context flattening all produced correct responses
- [x] `npm run lint` — 0 problems in changed files
- [x] `npm run check:any-budget:t11` — pass
- [x] `npm run check:tracked-artifacts` — pass
- [x] `npm run build` — succeeds
- [x] Golden snapshot (`executor-map-golden.test.ts`) regenerated (`UPDATE_GOLDEN=1`) — diff is exactly the two new aliases (`oneminai`, `1min`) plus the `keyCount` bump, both golden assertions pass


zero-executioner:
I have no idea how this tried to pull in nearly 3k commits, sorry about that

Issue #6131, opened 2026-07-03
IP blacklist doesn't persist (resets to Disabled + clears IPs on every restart/update) and isn't enforced


diegosouzapw:
**Reported by a user** (relayed): *"the IP blacklist feature still doesn't save the IPs added, it resets to 'Disabled' and clears all the blacklisted IPs after every OmniRoute update."*

## Root cause
`open-sse/services/ipFilter.ts` keeps its config in a **module-level in-memory `_config`** only:
- The `PUT /api/settings/ip-filter` route mutates `_config` (`configureIPFilter`/`addToBlacklist`) but **never persists to the DB**, and nothing **loads it on startup**. Every restart (i.e. every update) reloads the module → resets to defaults (`enabled:false`, empty blacklist). → the reported symptom.

## Second gap
`checkIP`/`createIPFilterMiddleware`/`checkRequestIP` are **not wired into the request pipeline** anywhere — so even before a restart, blacklisted IPs are **not actually blocked**. The feature is UI + settings only.

## Fix
- **A (persistence):** persist the config to the DB (`key_value` settings) on mutation + lazy-load on read (better-sqlite3 is sync).
- **B (enforcement):** call `checkRequestIP` in `runAuthzPipeline` (`src/server/authz/pipeline.ts`) for `/api/`+`/v1/` routes, returning 403 for blocked IPs, exempting loopback so the operator can't lock themselves out locally.

Validated via TDD (persistence across a simulated reload + blocked-IP → 403).

diegosouzapw:
Fixed in #6132 (merged into release/v3.8.44 → ships in the next release). The IP filter config now persists to the DB and is loaded on startup (survives restarts/updates), and blacklisted IPs are enforced in the authz pipeline (loopback exempt so you can't lock yourself out locally).

PR #11461, opened 2026-08-25
feat(providers): add MaxAI — signed OpenAI-compatible provider (chat, tools, vision, image-gen, doc-RAG)

arminanton:
## Summary

Adds **MaxAI** as a first-class, signed, OpenAI-compatible provider. MaxAI is a
web-app AI aggregator; this integration lets OmniRoute route to its **13 paid
chat models** (plus **6 image models**) through the standard `/v1` endpoints,
with full request signing, live model discovery, browserless onboarding,
prompted tool-calling, **vision input, image generation, and document RAG**.

**Chat models** (live from `/models/get_config`): `gpt-5.6`, `gpt-5.6-luna`,
`gpt-5.6-thinking`, `claude-5-sonnet`, `claude-haiku-4-5`, `gemini-3.1-pro-preview`,
`gemini-3-1-flash-lite`, `grok-4-1-fast-reasoning`, `grok-4-1-fast-non-reasoning`,
`grok-4.5`, `deepseek-v3.2`, `deepseek-r1`, `llama-3.3-70b`.

**Image models** (`/v1/images/generations`): `gpt-image-1`, `dall-e-3`,
`flux-1-schnell`, `flux-1-dev`, `flux-1-pro`, `sd3-medium`.

### What's in the box
- **Signed executor** (`open-sse/executors/maxai/*`): per-request `X-Authorization`
  built as HMAC-SHA1 → SM3 → CryptoJS-compatible AES-256-CBC over a structured
  payload, reproduced with `node:crypto`. Chat via `POST /gpt/cwc/chat` (SSE),
  reasoning surfaced from inline `<think>` blocks.
- **Live model + context-window discovery** (`services/maxaiModels.ts`): signs
  `/models/get_config`, maps the curated chat models with their real `max_tokens`
  and capability flags (incl. `vision`), and feeds the existing 24h context-window
  reconciler.
- **Vision input**: `image_url` content parts are forwarded inline in
  `message_content` to the 6 vision-capable models (GPT-5.6 / Luna / Thinking,
  Claude Haiku 4.5, Gemini 3.1 Pro / Flash-Lite). Mirrors the
  `translator/request/openai-to-cursor.ts` text-flatten-plus-preserve-images
  pattern; the text-only path is byte-unchanged.
- **Image generation** (`open-sse/handlers/imageGeneration/providers/maxaiImage.ts`):
  a new `maxai-image` format modeled on the `designer-web` handler (single
[...]

PR #11392, opened 2026-08-24
feat(check): make serviceKinds required + reverse-walk provider consistency gate

Tushar49:
…stency gate

Fixes #10513 · Refs Discussion #10267

Base-red inherited: #9985 (release/v3.8.50 not green)

What:
- Schema: providerSchema.ts — serviceKinds changed from optional to required
- Backfill: 312 catalog entries get serviceKinds computed from section + registry (341 total, 17 already declared)
- Gate: check-provider-consistency.ts — reverse walk flags llm-kind canonical providers with no REGISTRY entry, unless in KNOWN_CATALOG_ONLY (32 connection-routed providers with justification)
- Tests: 5 new reverse-walk test cases (11/11 gate tests pass)
- Script: scripts/ad-hoc/backfill-servicekinds.mjs — idempotent codemod

Validation:
- npm run typecheck:core — clean
- bun scripts/check/check-provider-consistency.ts — OK (261 REGISTRY, 341 canonical, 32 catalog-only)
- 11/11 gate tests, 24/24 provider schema tests, 34/34 route schema tests pass
- ESLint on all catalog files — clean

## Summary

- Describe the user-facing or operational change.

## Related Issues

- Closes #
- Related to #

## Validation

Choose the change type and focused loop from the
[Contribution Golden Path](../docs/ops/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite,
Vitest, the 60% coverage gate, and the production build all run in CI on this PR (#8329):

- [ ] Change type: provider / routing / UI / i18n / CLI / DB / build-deploy / other
- [ ] Focused tests and category gates from the golden path
- [ ] `npm run lint`
- [ ] Reconciled with the current active release base; focused checks rerun afterward
- [ ] Production-code changes include a new or updated automated test in this PR
- SonarQube is temporarily opt-in while the private project has no quota; it is not a PR gate.

## Tests Added Or Updated

- List every changed or added automated test file.
- If no production code changed, state that here.

## Coverage Notes

- If this PR changes `src/`, `open-sse/`, `electron/`, or `bin/`, explain which tests cover the change.
- If coverage moved down in any touched file, explain why and what follow-up task will recover it.

## Reviewer Notes

- Call out any risky areas, migrations, feature flags, or manual validation that reviewers should know about.

PR #11564, opened 2026-08-25
feat(compression): make proactive context-compression threshold a live setting

ragnar-claude:
The proactive compression trigger ratio is a hardcoded `COMPRESSION_THRESHOLD = 0.7` in chatCore. Operators cannot move compression relative to a client's own compaction point — e.g. Codex Desktop self-compacts at ~0.85 of its window, so the 0.7 proxy threshold always preempts the client's compaction with the proxy's lossier one (see #8932 for what that produced before 3.8.50).

This PR makes the ratio a live setting: `key_value` namespace `compression`, key `proactiveConfig`, `{"thresholdRatio": 0.7}`. Clamped [0.1, 0.99], 30s TTL cache, follows the ipFilter persistence pattern (#6131) — synchronous read stays in the sync hot path. Default unchanged; missing/invalid rows fall back to 0.7.

Tested: compression suite green including the #8932 regression; verified live on a production deployment (385k-token Responses request with a custom_tool_call pair compressed 385,056 → 188,039 and accepted upstream, zero orphaned calls).


diegosouzapw:
Diffstat muito grande/desalinhado com o escopo do título para validação segura no fluxo de lote atual (mesmo padrão já aplicado a #11390, #11461, #11392). Adiando para uma sessão dedicada de análise no ciclo v3.8.52 — sem prejuízo do trabalho, só precisa de atenção isolada em vez de ser boarded junto com dezenas de outras PRs pequenas.

PR #11579, opened 2026-08-26
fix(cache): compare expires_at against an ISO cutoff, not datetime('now') (#11559)

### LINKED ISSUE #11559: fix(backend): semantic cache TTL never expires same-UTC-day entries (ISO vs datetime('now') text compare)
## Description

`semanticCache.ts` writes `expires_at` in ISO-8601 format (`new Date(...).toISOString()` → `"2026-08-26T14:00:00.000Z"`), but reads it back by comparing against SQLite's `datetime('now')`, which produces a space-separated string (`"2026-08-26 14:00:00"`).

Because `expires_at` is a TEXT column, the comparison is lexicographic — and at index 10 the stored value has `'T'` (0x54) where `datetime('now')` has `' '` (0x20). `'T' > ' '`, so **every entry whose UTC calendar date equals "today" compares greater than `datetime('now')`, regardless of its actual time-of-day expiry**. Entries therefore remain "valid" until 00:00 UTC of the day *after* their true TTL expires — up to ~24 hours stale.

This is a static analysis finding from reading `src/lib/semanticCache.ts` on the current default branch; I have not run a gateway instance.

## Steps to Reproduce

1. Enable the semantic cache and send any request that gets cached with a short TTL (e.g. a few minutes), well before UTC midnight.
2. Let the TTL lapse.
3. Query the cache through the DB-backed path (`getCachedResponse`) after eviction/restart, or check `/api/cache/stats`.

Observed statically from source:

- Write path (`setCachedResponse`, ~line 260):
  ```ts
  const now = new Date().toISOString();
  const expiresAt = new Date(Date.now() + ttl).toISOString();
  db.prepare(`INSERT OR REPLACE INTO semantic_cache (... created_at, expires_at) VALUES (?, ?, ?, ?, ?, 0, ?, ?)`)...
  ```
- Read path (`getCachedResponse`, ~line 206; same predicate in `getCacheStats` line 345):
  ```ts
  "SELECT response, tokens_saved FROM semantic_cache WHERE signature = ? AND expires_at > datetime('now')"
  ```

## Expected Behavior

A cached response should be considered expired once `expires_at` passes, i.e. comparisons should happen between values in the *same* format (either store space-separated SQLite format, or compare against an ISO string built in JS, or store epoch millis).

## Actual Behavior

Same-UTC-day timestamps always satisfy `expires_at > datetime('now')` due to `'T' > ' '` in text ordering, so entries survive up to ~24h beyond their configured TTL. The in-memory LRU has correct per-entry TTLs, but `getCachedResponse` re-promotes DB rows into memory on hit (~line 220), so staleness survives eviction/restarts.

Note: other comparisons in the file are consistent-format (e.g. `invalidateStale` compares against an ISO cutoff) — only the `expires_at vs datetime('now')` pairs mismatch.

## Test Impact

Needs a new unit test

## Validation Plan

- A unit test that seeds a row with `expires_at` = (now + small TTL, still today's UTC date), advances a mocked clock past the TTL, and asserts `getCachedResponse` misses.
- `node --import tsx --test tests/unit/semanticCache.test.ts` (or the repo's equivalent suite for this module).

ntdat812:
## Summary

- `semantic_cache` entries survived up to ~24h past their configured TTL, and a stale hit was
  re-promoted into the in-memory LRU, so it outlived both eviction and a restart.
- `setCachedResponse` is the only writer of `expires_at` and stores an ISO-8601 string
  (`"2026-08-26T14:00:00.000Z"`). Both read paths compared it against `datetime('now')`, which
  renders the same instant as `"2026-08-26 14:00:00"`.
- The column is `TEXT`, so the comparison is lexicographic, and the two spellings diverge at
  index 10: `'T'` (0x54) against `' '` (0x20). `'T'` sorts higher, so **any row whose UTC
  calendar date is today satisfied `expires_at > datetime('now')` whatever its time-of-day
  said** — it stayed live until 00:00 UTC the next day.
- Rows dated a previous day still expired correctly, which is why this never presented as a
  permanently stuck cache and is easy to miss in normal use.

Both sites now bind a JS-built ISO cutoff, which is the approach `invalidateStale` already
uses for `created_at`.

Measured directly, so the ordering claim is not taken on faith:

```console
$ node -e "const {DatabaseSync}=require('node:sqlite'); const db=new DatabaseSync(':memory:');
  const expired=new Date(Date.now()-60000).toISOString();
  console.log(expired, '|', db.prepare(\"SELECT datetime('now') n\").get().n);
  console.log('> datetime(now):', db.prepare(\"SELECT (?>datetime('now')) v\").get(expired).v);
  console.log('> ISO now      :', db.prepare('SELECT (?>?) v').get(expired,new Date().toISOString()).v);"
2026-08-26T01:57:52.631Z | 2026-08-26 01:58:52
> datetime(now): 1      <- a row that expired a minute ago, still counted live
> ISO now      : 0      <- correctly expired
```

## Related Issues

- Closes #11559

Thanks to @Meet6338-X, who filed #11559 with the root cause already localised to the `'T'`
vs `' '` ordering — I verified the analysis against `release/v3.8.51` rather than rediscovering
it, and it held exactly as written.

## Validation

- [x] Change type: **DB** (SQLite predicate in `src/lib/semanticCache.ts`)
- [x] Focused tests and category gates from the golden path — see below
- [x] `npm run lint` — clean on both changed files (`eslint` exit 0)
- [x] Reconciled with the current active release base — rebased onto `origin/release/v3.8.51`
      at `6eb7bf0ba9` (2026-08-26); focused checks rerun after the rebase, same results; focused checks rerun after

**The new test fails on the base branch.** Verified by reverting only
`src/lib/semanticCache.ts` to `origin/release/v3.8.51` and keeping the test file:

| | base | this PR |
| --- | :---: | :---: |
| `stores expires_at in a spelling that SQLite's datetime('now') misorders` | ok | ok |
| `misses a row that expired earlier today` | **not ok** | ok |
| `does not re-promote the expired row into the memory cache` | **not ok** | ok |
| `does not count the expired row in getCacheStats().dbEntries` | **not ok** | ok |
| | 1 pass / 3 fail | 4 pass / 0 fail |

The first case is a precondition, not a behaviour check — it asserts the `'T'` / `' '`
misordering actually holds in this SQLite build, so that if a future driver change makes the
comparison well-behaved the suite says *why* the other three stopped being meaningful instead
of silently passing.

Wider cache suite — `tests/unit/*cache*.test.ts` + `*semantic*.test.ts`, 548 tests:
**544 pass, 4 fail**. Those 4 are pre-existing and unrelated
(`chatcore-reasoning-cache-write-guard`, two in `kiro-sso-cache-direct-clientid-1253`,
`model-catalog-cache-swr-8728`); I ran those three files against an unmodified
`semanticCache.ts` and got the identical 11 pass / 4 fail, so this PR neither causes nor fixes
them.

## Tests Added Or Updated

- `tests/unit/semantic-cache-ttl-text-compare.test.ts` (new, 4 cases)

## Coverage Notes

`src/lib/semanticCache.ts` is the only production file touched. The two changed predicates are
the ones in `getCachedResponse` (DB read path) and `getCacheStats` (live-entry count); the new
test exercises both, plus the LRU re-promotion that turns a single stale read into a persistent
one.

## Reviewer Notes

- **No migration and no data change.** Stored rows keep their existing ISO format; only the
  comparison side moves. Nothing needs backfilling, and a downgrade would simply restore the
  old leniency.
- I checked that `setCachedResponse` really is the sole writer before relying on the format:
  `expires_at TEXT NOT NULL` in `src/lib/db/core.ts` has no `DEFAULT`, and no other module
  inserts into `semantic_cache` (`src/lib/db/semanticCache.ts` only reads and deletes). If a
  future writer stores the SQLite spelling instead, these predicates would misjudge it in the
  other direction — worth a schema-level normalisation then, which felt out of scope here.
- The test derives its timestamp from **today's UTC midnight** rather than `now - 60s` on
  purpose. A fixed negative offset lands on the previous UTC date whenever the suite runs just
  after midnight, and the test would then pass against the unfixed code — a regression guard
  that is green for the wrong reason a fraction of the day is worse than none.


ntdat812:
leaving it to be re-derived.

this diff.

Meet6338-X (approved):
Issue reporter here (#11559) - reviewed the diff against my original root-cause analysis and this is the correct fix.

Verified:

1. Both broken comparison sites are covered (getCachedResponse line ~206 and getCacheStats line ~345); invalidateStale was already ISO-consistent and is untouched, correctly so.

2. Direction of the normalization is the right call: aligning the read side to the writer's format means pre-existing rows (which were always written as ISO) stay valid without any migration/backfill step.

3. The new tests handle the subtle trap here: deriving the expired row's expiry from start-of-current-UTC-day keeps the suite deterministic even when it runs just after 00:00 UTC, where a naive now-60s fixture would silently land on the previous date and pass against the unfixed code. The explicit precondition assertions (index-10 'T' vs ' ', string ordering) mean the test fails loudly if SQLite or the writer ever changes spelling again - that is exactly the regression guard this needs.

On the red CI: confirmed nothing in this diff touches the failing gates (db-cleanup/db-settings/cursor-renewal etc. are unrelated modules), consistent with what you documented and with the #9282 pattern.

From my side this is ready to merge once the base settles.

ntdat812:
Thanks for reviewing it, and for the original report — the root cause in #11559 was already
localised to the `'T'` / `' '` ordering, so this was mostly a matter of confirming it against
`release/v3.8.51` and writing the guard. That is a materially better starting point than most
bug reports give you.

On point 2 — agreed, and worth stating for the record since it is the part that decides whether
this needs a migration: `setCachedResponse` is the only writer of `expires_at`
(`expires_at TEXT NOT NULL` in `src/lib/db/core.ts` has no `DEFAULT`, and `src/lib/db/semanticCache.ts`
only reads and deletes), so every stored row is already ISO. Moving the read side to match is
therefore backfill-free. If a future writer ever stores the SQLite spelling instead, these
predicates would misjudge it in the other direction — that would be the moment to normalise at
the schema level.

You are right about the midnight trap being the reason for the start-of-UTC-day fixture. A
`now - 60s` row lands on the previous UTC date for whatever slice of the day the suite happens
to run in just after midnight, and the test would then pass against the unfixed code — green
for the wrong reason, which is worse than absent.

Nothing outstanding from my side; it is rebased onto `6eb7bf0ba9` and waiting on the base
settling, as you say.

Meet6338-X:
Gate-by-gate attribution from the actual CI logs of this PR's run (job IDs 98041152112 / 98041152132 / 98041152156 / 98041152224) — confirming @ntdat812's read with log-level evidence, since six red gates make "not mine" expensive to keep re-deriving per PR:

**1. Docs Gates (fast-path)** — fails in `check:env-doc-sync`:

```
? In code but missing from .env.example: 1
   - OMNIROUTE_CORPUS_CACHE_SIZE
```

Someone added the env var in code without updating `.env.example`. Nothing in this diff touches env vars.

**2. No new ESLint warnings** — exits 2 with:

```
```

Suppression-file drift on the base: previously-suppressed warnings no longer occur (opposite of new warnings), and the gate treats leftovers as an error. No lint finding points at `semanticCache.ts`.

**3. Fast Quality Gates** — `2 gate(s) failed: known-symbols mutation-test-coverage`:

- `known-symbols` → `[executor] 142 alias(es) registered that do not resolve to a valid BaseExecutor` — including core built-ins (`antigravity`, `github`, `ghe-copilot`, `kiro`, `bedrock`, `codex`), i.e. registry/catalog drift in the gate context.
- `mutation-test-coverage --strict` → 3 covering tests missing from `stryker.conf.json` `tap.testFiles`, all in modules this diff never touches: `tests/unit/authz/route-guard-tunnel-processes-local-only.test.ts` (for `src/server/authz/routeGuard.ts`) and `tests/unit/universal-quota-aware-routing.test.ts` (for `combo/quotaScoring.ts` + `combo/quotaStrategies.ts`).

Meanwhile the gate's own impact analysis mapped this diff's changed symbol (`semanticCache.ts` `nowIso`) onto its dependent tests (`auto-empty-pool-fastfail-6458`, `fix-error-message-candidates`, `instrumentation-node-boot-db-failure`, `models-test-error-shape`, `provider-request-failure-pipeline`, `safe-resolve-proxy-fail-closed`) — all green.

**4–6. Unit Tests fast-path shards** — failing tests are exclusively:

- `token-health-no-refresh-token-expired-5326.test.ts` → Copilot classification (`github_access_token_invalid` vs expected `invalid_grant`)
- `ts7-executor-shared-shapes.test.ts` → tools truncation (`200 !== 128`)
- plus the set @ntdat812 listed for other shards (db-cleanup, db-settings-crud, cursor-renewal, kiro-multi-account-isolation, openrouter-provider-stats, namespace-tool-identity-integration, t31/t33/t34/t38-model-specs, authz/route-guard-version-get-exemption)

Every semantic-cache/TTL-related test in every shard passes ✔.

## Bottom line

All six failures reproduce from base-state drift (env-doc contract, suppression file, stryker registration, executor registry, plus unrelated module regressions) — consistent with #9282. Zero failures reference `expires_at`, `nowIso`, or any line of this diff. From the issue-author side (#11559): fix is correct, tests are sound, and this is ready to merge as soon as the base settles. No action needed from @ntdat812 beyond the rebase already planned.

diegosouzapw:
Obrigado pelo fix — é o mesmo bug e a mesma correção que #11573 (comparar `expires_at` contra um cutoff ISO em vez de `datetime('now')`). Vou mergear #11573 (aberta primeiro) e fechar esta para evitar duplicidade. Sua análise da causa raiz (divergência lexicográfica no índice 10, 'T' vs espaço) está correta e ajudou a confirmar a mesma leitura independentemente.

PR #11585, opened 2026-08-26
test(claude-web): anchor on the alias wiring, not on how the map builds it

ntdat812:
## Summary

`tests/unit/claude-web-transport.test.ts` checks that the Claude Web executor carries no
execution-time dependency on the standalone Turnstile solver. Two of its assertions are
*anchors* — there to prove the file read resolved to the real barrel, so the negative guards
below them cannot pass against a moved or emptied file. They were pinned to the exact
construction spelling:

```ts
assert.match(indexSource, /"claude-web": new ClaudeWebExecutor\(\)/);
assert.match(indexSource, /"cw-web": new ClaudeWebExecutor\(\)/);
```

#11421 made the executor barrel lazy. The values are now
`() => import("./claude-web.ts").then((m) => new m.ClaudeWebExecutor())`, so neither anchor
matches any more. The wiring they guard did not change — only how the map constructs executors.

**This is a base-red gate.** The test fails on `release/v3.8.51` itself, so every open PR
inherits the failure in `Unit Tests fast-path (1/4)` no matter what it touches.

The anchors now match the alias key and the class name and let the construction between them be
spelled either way:

```ts
assert.match(indexSource, /"claude-web":.*ClaudeWebExecutor/);
assert.match(indexSource, /"cw-web":.*ClaudeWebExecutor/);
```

`.` does not match a newline without the `s` flag, so each stays confined to its own entry.

## Related Issues

- Refs #11421 (the lazy-executor change this anchor was not updated for)

## Validation

- [x] Change type: **other** (test-only)
- [x] `npm run lint` — clean (`eslint` exit 0)
- [x] Reconciled with the current active release base — branched from `origin/release/v3.8.51`
      at `6eb7bf0ba9`
- [x] Production-code changes include a new or updated automated test in this PR — n/a, no
      production code changed

```console
# on 6eb7bf0ba9, unmodified
$ node --test tests/unit/claude-web-transport.test.ts
# tests 9 / pass 8 / fail 1     <- not ok 2, "Claude Web executor transport orchestration"

# with this change
# tests 9 / pass 9 / fail 0
```

**The loosened anchor still catches a real break.** A looser assertion is only worth having if
it still fails when the thing it guards is actually gone, so I checked that rather than assuming
it: repointing one alias at a different executor —

```diff
-  "cw-web": () => import("./claude-web.ts").then((m) => new m.ClaudeWebExecutor()), // Alias
+  "cw-web": () => import("./default.ts").then((m) => new m.DefaultExecutor()), // Alias
```

— turns the test red again (`pass 8 / fail 1`), and restoring it turns it green. The guarantee is
unchanged; only the spelling it tolerates is wider.

## Tests Added Or Updated

- `tests/unit/claude-web-transport.test.ts` — two anchor assertions relaxed. No production code
  changed.

## Coverage Notes

No `src/`, `open-sse/`, `electron/` or `bin/` change, so coverage is unmoved.

## Reviewer Notes

- I checked whether the same stale pattern exists elsewhere before scoping this:
  `grep -rn '": new [A-Za-z]*Executor' tests/` returns these two lines and nothing else, so this
  is the whole of that class of breakage, not the first of many.
- The alternative fix is to re-pin the anchors to the new lazy spelling. I did not, because that
  reproduces the original problem — the next change to how the barrel constructs executors would
  turn this test red again for a reason that has nothing to do with Turnstile. What the test
  cares about is *which class the alias resolves to*, and that is what it now asserts.
- Worth saying plainly: I found this because it was failing CI on two unrelated PRs of mine
  (#11579, #11581) alongside #11577 and #11580 by other authors. It is not something I broke,
  and it is not specific to my changes.


ntdat812:
leaving it to be re-derived.

this diff.

PR #11588, opened 2026-08-26
fix(pack): register bin/cli/utils/volatileEnvPath.mjs as a required artifact path (#11437)

ntdatt812:
Part of the "every gate is red on every branch" cleanup — one of six independent unit-test failures on `release/v3.8.51`.

## The red test

```
✖ every bin/omniroute.mjs local import is enforced by check:pack-artifact
  AssertionError: add bin/<file> to PACK_ARTIFACT_REQUIRED_PATHS: cli/utils/volatileEnvPath.mjs
  actual:   [ 'cli/utils/volatileEnvPath.mjs' ]
  expected: []
```

## Cause

[#11437](https://github.com/diegosouzapw/OmniRoute/pull/11437) (`943b9aaa8` — the commit
issue #11449 names as the offending push range) added to `bin/omniroute.mjs`:

```js
import { describeVolatileEnvWarning } from "./cli/utils/volatileEnvPath.mjs";
```

without registering the file in `PACK_ARTIFACT_REQUIRED_PATHS`.

## Why the list matters

This is not bookkeeping. The list's own comment says it:

> Direct imports of bin/omniroute.mjs — `bin/cli/` is only an allowlist **PREFIX**, so a
> file vanishing from the tarball never fails the unexpected-paths check; only these
> required entries make its absence loud.

`assembleStandalone` copies wrappers to the dist root and the prepublish prune deletes
anything not allowlisted. A `bin/cli/` file is inside a prefix, so its disappearance is
silent — and `describeVolatileEnvWarning` is called on **every CLI boot**, so the first
symptom would be a crash on a user's machine after `npm i -g omniroute`. Exactly the
#7065 class the `tls-options` and `head-response-guard` entries directly above it exist
to prevent.

## Verification

```
# base branch
✖ every bin/omniroute.mjs local import is enforced by check:pack-artifact
ℹ tests 6  ℹ pass 5  ℹ fail 1

# with this change
✔ sanity: EXTRA_MODULE_ENTRIES parsing finds the known dist-root wrappers
✔ every local import of every npm-shipped wrapper survives the prune (allowlist)
✔ every local import of every npm-shipped wrapper is enforced by check:pack-artifact
✔ dynamic import() closure is covered (server-ws boots dist/server.js)
✔ every bin/omniroute.mjs local import is enforced by check:pack-artifact
✔ no npm-shipped wrapper uses a parent-relative (../) import
ℹ tests 6  ℹ pass 6  ℹ fail 0
```

Two added lines, no code change.


ntdatt812:
Follow-up in `2625a03`: adding `bin/cli/utils/volatileEnvPath.mjs` to `PACK_ARTIFACT_REQUIRED_PATHS` also broke a sibling test — `findMissingArtifactPaths flags missing root runtime files in the tarball` ([pack-artifact-policy.test.ts:243](https://github.com/diegosouzapw/OmniRoute/blob/main/tests/unit/pack-artifact-policy.test.ts#L243)) pins the full expected list literally, so the new entry showed up as an unexpected diff.

I extended the golden list rather than deriving it from the constant under test: the literal list is what makes an accidental *removal* from the policy visible, and deriving it would make the assertion vacuous. `tests/unit/pack-artifact-policy.test.ts` is 17/17 locally.

Worth knowing for future entries: every addition to `PACK_ARTIFACT_REQUIRED_PATHS` needs the same one-line update here.

PR #11591, opened 2026-08-26
test(executors): drop the tool-truncation assertion #11444 removed the code for

ntdatt812:
Part of the "every gate is red on every branch" cleanup — one of six independent unit-test failures on `release/v3.8.51`.

## The red test

```
✖ truncates an over-long tools array to 128 entries
  AssertionError: upstream rejects more than 128 tools
  200 !== 128
```

## Two tests asserted opposite contracts for the same call

`tests/unit/opencode-tools-no-truncation.test.ts` exists specifically to pin the
opposite, and it passes today:

> `OpencodeExecutor.transformRequest` contained a second hardcoded tool-list truncation
> (`mb.tools.slice(0, 128)`)… When a client (opencode/claude-code) sends >128 tools, the
> opencode executor silently dropped every tool after the 128th (e.g. task, skill,
> write, read), leaving subagent tasks **paralyzed**.
>
> Tool-list limiting is the chatCore `upstreamBody.truncateToolList()`'s job
> (per-provider known/effective limits via `toolLimitDetector`), NOT the executor's.

So the removal was deliberate and fixed a real user-facing bug (#11444). This file still
asserted the truncation that fix removed.

## Why the behaviour is better where it moved

`upstreamBody.truncateToolList()` reads a **per-provider** limit
(`getKnownToolLimit` → `PROVIDER_TOOL_LIMITS`) rather than a hardcoded 128, so
`grok-cli` (200) and `nvidia` (1536) are no longer cut at someone else's ceiling, and
`bypassDefaultToolLimit` gives callers an escape the executor-level slice never had.

## What changed

I did not delete the case. It now pins the same direction from this file's angle — the
TS 7 narrowing fix (#8484) must not let the cap creep back in here — so the two files
agree instead of racing:

```ts
it("forwards an over-long tools array intact, leaving the limit to chatCore (#11444)", () => {
  assert.equal(out.tools.length, 200, "the executor must not impose a tool cap");
  assert.equal(out.tools[199].function.name, "tool_199", "order and tail preserved");
});
```

The describe title lost the word "truncation" for the same reason.

## Verification

Both files, run together:

```
# base branch
✖ truncates an over-long tools array to 128 entries    200 !== 128

# with this change
✔ preserves all 150 tools (no slice(0,128))
✔ preserves all 300 tools (well beyond the old hardcoded 128 cap)
✔ preserves tool order (first AND last tool intact)
✔ preserves empty tools array intact
✔ preserves tool objects content exactly without mutation
✔ is a no-op when tools is absent
✔ OpencodeExecutor — no hardcoded 128-tool truncation (#11444)
✔ forwards an over-long tools array intact, leaving the limit to chatCore (#11444)
✔ leaves a within-limit tools array untouched
✔ is a no-op when the body carries no tools
✔ leaves an array-shaped body alone (pins the !Array.isArray guard)
ℹ tests 10  ℹ pass 10  ℹ fail 0
```

Test-only; no production code touched.

### DIFF
--- a/changelog.d/maintenance/11591-ts7-stale-tool-truncation.md
+++ b/changelog.d/maintenance/11591-ts7-stale-tool-truncation.md
@@ -1,0 +1,1 @@
+- **test(executors):** drop the stale 128-tool truncation assertion that contradicted the #11444 pass-through contract ([#11591](https://github.com/diegosouzapw/OmniRoute/pull/11591))
--- a/tests/unit/ts7-executor-shared-shapes.test.ts
+++ b/tests/unit/ts7-executor-shared-shapes.test.ts
@@ -14,7 +14,7 @@ const { OpencodeExecutor } = await import("../../open-sse/executors/opencode.ts"
  * that failed to type-check — no duplicate added here.
  */
 
-describe("OpencodeExecutor — tools truncation survives the narrowing fix", () => {
+describe("OpencodeExecutor — the tools array survives the narrowing fix", () => {
   const executor = new OpencodeExecutor("opencode-go");
   const CREDENTIALS = { apiKey: "k" } as Record<string, unknown>;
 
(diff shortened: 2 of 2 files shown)


ntdatt812:
CI note: this run shows one failure that is **not** from this PR — `handleChat recovers from a real 429 once the connection cooldown expires` ([chat-cooldown-aware-retry.test.ts:197](https://github.com/diegosouzapw/OmniRoute/blob/main/tests/unit/chat-cooldown-aware-retry.test.ts#L197)), `429 !== 200` after 3683ms.

Evidence it is unrelated:

- this PR changes exactly one test file, `tests/unit/ts7-executor-shared-shapes.test.ts`, plus a changelog fragment — a different suite entirely;
- the same test passes on the CI runs of my three sibling branches in the same window (#11588, #11590, #11592);
- `tests/unit/chat-cooldown-aware-retry.test.ts` is 7/7 locally.

It looks like a timing-boundary flake rather than a one-off: the test sets `maxRetryIntervalSec: 3` while the stub returns `Retry-After: 2` on three consecutive calls and then expects a 4th call to succeed. That leaves almost no margin between the retry budget and the mandated waits, so accumulated per-attempt overhead on a loaded runner can exhaust the budget before the 4th call — which matches the observed 3683ms, short of the ~6s of Retry-After waits the happy path needs.

I have not read the retry internals closely enough to be sure that is the mechanism, so I am flagging it rather than sending a fix. Happy to dig in and send one if you would like.

PR #11598, opened 2026-08-26
fix(autoCombo): lifecycle veto, deterministic capability merge, variant inheritance in models_dev_tier (#11508)

### LINKED ISSUE #11508: fix(autoCombo): models_dev_tier (layer 3) scores retired models as premium, merges conflicting provider rows last-write-wins, and misses effort variants
## Context

After #11503 empties the static table, layer 3 (`models_dev_tier`) becomes the dominant data-bearing layer — on a freshly synced `model_capabilities` (8,628 rows, 204 providers, models.dev) it resolves **504 / 1,285** catalog ids (39 %), second only to the wildcard. Measured with `getTaskFitnessWithSource(id, "coding")` on the #11507 branch against a copy of a synced DB. Three defects surface immediately.

## 1. No lifecycle awareness — a retired model scores the same as the current flagship

| id | vendor status | score | via |
|---|---|---|---|
| `claude-opus-5` | active | 0.920 | `models_dev_tier` |
| `openai/gpt-5.2-codex` | **shut down 2026-07-23** | **0.920** | `models_dev_tier` |

`deriveTierFromCapabilities()` (`taskFitness.ts:194`) maps `reasoning === true` → `premium` → 0.92 with no lifecycle input; models.dev keeps listing retired models with their capabilities (its `status` field flagged 0 of 22 retired ids in the #8777 audit). The ranking inversion #11503 removes from layer 4 is re-created one layer up.

## 2. Tier collapse — 63 % of ids are `premium`

Across 3,491 distinct `model_id`s in the synced table, **63.2 %** have `reasoning = true` somewhere → `premium`. The tier carries almost no discriminative power between a frontier model and a 7B reasoning finetune (#8778 predicted this; the number is now measured).

## 3. Cross-provider rows collapse last-write-wins, with no ordering

`model_capabilities` is keyed per (provider, model_id): `claude-opus-5` has 18 rows, `deepseek-v4-flash` 26. `loadModelCapabilities()` does `SELECT * FROM model_capabilities` (`:211`, no `ORDER BY`) and writes `cache[modelId] = …` (`:218`), so the last row returned wins. The rows disagree:

| | count |
|---|---|
| distinct ids | 3,491 |
| ids present under >1 provider | 1,176 |
| … with conflicting `reasoning` | **189** |
| … with conflicting `tool_call` | **134** |
| … with conflicting `limit_context` | **528** |

e.g. `deepseek-v4-flash` has both `reasoning ∈ {1,0}` and `tool_call ∈ {1,0}`; `claude-sonnet-5` has `tool_call ∈ {1,0}`. Which tier the model lands in depends on SQLite's row order for that query — undefined, and it changes across syncs.

## 4. Variants still miss (same gap #11492 closed for layers 1–2)

`getModelsDevTierFitness()` looks up `caps[normalizedModel]` exactly (`:252`); models.dev only lists base ids. `gpt-5.6-sol-xhigh` and `claude-fable-5-thinking-max` fall to 0.5 while `gpt-5.6-sol` gets 0.92. #11492's `resolveScoresAs()` was wired into `user_override` / `arena_elo` only.

## Proposal

1. **Lifecycle filter**: consult `config/quality/model-lifecycle.json` (#11507) in `getModelsDevTierFitness()` — `retired` → return `null` (fall through to the documented 0.5 neutral), never a tier score.
2. **Deterministic merge across providers**: aggregate per `model_id` at load time instead of last-write-wins — `reasoning` / `tool_call` = *majority* (or `any`, documented either way), `limit_context` = *max* — and `ORDER BY model_id, provider` so the result is reproducible.
3. **Variant inheritance**: on a miss, retry with `resolveScoresAs(id).base` exactly as layers 1–2 do; report source `models_dev_tier:inherited`.
4. Tier discrimination (§2) is a design question (context-window bands? release-date recency from models.dev's `release_date`?) — out of scope here; this issue is the three mechanical defects.

Tests: retired id in fixture → `null`; conflicting rows → same tier regardless of insertion order; `gpt-5.6-sol-xhigh` inherits `gpt-5.6-sol`'s tier.

Related: #11503 / #11507 (layer 4 + snapshot), #11489 / #11492 (variant inheritance), #11504 / #11506 (arena ingest), #8778.

Meet6338-X:
## What

Implements the three mechanical fixes proposed in #11508 for layer 3 (`models_dev_tier`) of the autoCombo fitness chain:

1. **Lifecycle veto** — `getModelsDevTierFitnessWithSource()` consults `config/quality/model-lifecycle.json` before any other signal in this layer; an id with vendor status `retired` returns `null` and falls through to the neutral baseline instead of scoring premium off capability rows that outlived the vendor's support.
2. **Deterministic cross-provider merge** — `loadModelCapabilities()` aggregates rows per `model_id` instead of last-write-wins over SQLite's undefined row order: booleans are capability-maximal (any non-null row asserting `true` wins), `limit_context` is max across rows carrying one. Both rules are order-independent, so a model's tier can no longer flip between syncs.
3. **Variant inheritance** — on a capabilities miss, the lookup retries through `resolveScoresAs()`'s base id and reports source `models_dev_tier:inherited`, matching what #11492 wired into layers 1–2. models.dev publishes base ids only, so e.g. `gpt-5.6-sol-xhigh` no longer falls to the wildcard while `gpt-5.6-sol` scores premium.

Fixes #11508

## API notes (compat preserved)

- Existing `getModelsDevTierFitness(model, taskType): number | null` is unchanged in shape and now delegates to the new `getModelsDevTierFitnessWithSource()`, so the three existing call sites/tests keep working.
- The resolution chain (`getTaskFitnessWithSource`) consumes the new source-carrying surface, which is how `models_dev_tier:inherited` becomes visible to callers.
- Adds `invalidateCapabilitiesCache()` — small test/ops hook so seeding `model_capabilities` after first read is observable (the capabilities cache previously had no reset path).

## Design choices worth flagging

- **any-non-null-true vs majority**: #11508 allowed either if documented. I chose capability-maximal (`true` wins) to pair coherently with `limit_context = max`; happy to switch to majority if maintainers prefer conservative merging.
- **Lifecycle veto placement** is deliberately *before* the `model_intelligence` lookup in this layer — a stale premium row for a retired id must not resurrect through the DB path either.
- The lifecycle snapshot is loaded lazily once and never throws when absent (routing degrades to today's behavior rather than disabling itself).

## Tests

`tests/unit/autoCombo/models-dev-tier-11508.test.ts` covers:

- retired id with a seeded premium capability row → tier layer misses (veto fires before capabilities)
- non-retired control still scores
- conflicting two-provider rows → identical tier under both insertion orders (order-invariance regression)
- null-capability rows never masquerade as `false` during aggregation

One honest gap: the `models_dev_tier:inherited` path depends on the model registry resolving a base id (`resolveScoresAs` is catalog-anchored), which I could not fixture deterministically in an environment-independent unit test — the included contract test pins the non-inherited source string, and the inherited path needs coverage where the catalog fixture lives (happy to extend given a pointer to the right fixture).

## Validation plan

```
node --import tsx/esm --test tests/unit/autoCombo/models-dev-tier-11508.test.ts
node --import tsx/esm --test tests/unit/task-fitness-missing-table-8603.test.ts   # compat: signature unchanged
npm run quality:refresh-model-lifecycle && npm run check:model-lifecycle          # lifecycle file untouched by this diff
```

⚠️ Known-repo-wide: six gates are currently red on every open PR against `release/v3.8.51` including unrelated shards (documented by @ntdat812 in #11579 and tracked via #9282). Failures in those gates on this PR are pre-existing; the targeted unit tests above are the ones to watch.


Meet6338-X:
⚠️ Heads-up before triage: the six red `Quality Gates` gates (Docs fast-path, Fast Quality Gates, No new ESLint warnings, Unit shards 1/3/4) are **repo-wide base drift**, not this diff.

Full log-level attribution posted in #11579 (comment 5420730387), same failure set across unrelated open PRs (#11577/#11580/#11582), tracked via #9282:

- Docs: `OMNIROUTE_CORPUS_CACHE_SIZE` missing from `.env.example` (env-doc sync)
- ESLint: leftover suppression-file entries → exit 2
- Fast Quality: `known-symbols` (142 unresolved executor aliases incl. core built-ins) + `mutation-test-coverage --strict` (3 unregistered tap.testFiles: routeGuard / quotaScoring / quotaStrategies)
- Unit shards: Copilot token-health classification, ts7 tools truncation (200≠128), plus the db-cleanup/cursor-renewal/etc. set

None reference `taskFitness`, `model_capabilities`, lifecycle JSON, or any line of this diff. The one gate that DID map this diff's symbols (`Fast Quality Gates` impact analysis pattern, cf. how `nowIso` was mapped in #11579's run) is exactly where reviewers should look.

Targeted validation for this PR once base settles:

```
node --import tsx/esm --test tests/unit/autoCombo/models-dev-tier-11508.test.ts
node --import tsx/esm --test tests/unit/task-fitness-missing-table-8603.test.ts
```

MumuTW:
Thanks for picking this up — the three mechanisms are the ones #11508 asked for and the aggregation code reads well. I pulled `72f6d5d` into a scratch worktree on top of `release/v3.8.51` (0 commits behind) and ran it; the existing suites are fine (`open-sse/services/autoCombo/__tests__` 116/116, `tests/unit/task-fitness*.test.ts` 5/5, and `gpt-5.6-sol-xhigh` → `{0.92, "models_dev_tier:inherited"}` as intended). Four things block it as-is, though — all reproducible:

### 1. The new test file never executes

`tests/unit/autoCombo/**` is collected by **vitest** (`vitest.mcp.config.ts:21`), not `node:test`, and the test dir is three levels deep:

```
$ npx vitest run --config vitest.mcp.config.ts tests/unit/autoCombo/models-dev-tier-11508.test.ts
Error: Cannot find module '../../src/lib/db/core.ts'      # needs ../../../
```

After fixing the paths:

```
Error: No test suite found in file tests/unit/autoCombo/models-dev-tier-11508.test.ts
```

because `describe`/`it` come from `node:test`. Switching the imports to `vitest` fixes both (see the sibling files in that dir). CI on this PR is still `action_required`, so nothing has run it yet.

### 2. The veto misses provider-prefixed ids — including the issue's headline case

`config/quality/model-lifecycle.json` keys `retired` by bare id (0 of 106 keys contain `/`), but catalog ids carry the prefix. With a `reasoning=1` row seeded:

```
openai/gpt-5.2-codex     → {"score":0.92,"source":"models_dev_tier"}   # retired 2026-07-23
google/gemini-2.0-flash  → {"score":0.92,"source":"models_dev_tier"}
gpt-5.2-codex            → null                                          # only the bare form is vetoed
```

`loadRetiredModels().has()` needs to also try the id with the `<provider>/` prefix stripped (or the snapshot needs both forms).

### 3. Variants bypass the veto through inheritance

The veto checks the input id only, before `resolveScoresAs()`; a variant of a retired base inherits premium:

```
claude-3-5-sonnet-20241022        → null                                      ✓
claude-3-5-sonnet-20241022-high   → {"score":0.92,"source":"models_dev_tier:inherited"}
chatgpt-4o-latest-free            → {"score":0.92,"source":"models_dev_tier:inherited"}
gpt-4-0125-preview-high           → {"score":0.92,"source":"models_dev_tier:inherited"}
```

Re-apply the veto to `base` after resolution.

### 4. `config/` isn't shipped

`package.json` `files` has no `config/` entry, so the npm-installed runtime has no `model-lifecycle.json`; the `readFileSync` `catch` swallows it and the veto is a silent no-op outside a repo checkout. Either add `config/quality/model-lifecycle.json` to `files` or inline the retired set at build time. Worth a test that asserts the file resolves from the published layout, since the failure mode is invisible.

### Minor

- The `:inherited` path *is* fixturable: `gpt-5.6-sol-xhigh` resolves via `effort-suffix` in a plain checkout, so the last test can assert `source === "models_dev_tier:inherited"` after seeding `gpt-5.6-sol` instead of the current `if (own !== null)` no-op.
- The test seeds via `getDbInstance()` without cleanup; vitest has no `isolateDataDir` preload, so please use a scratch `DATA_DIR` + `afterAll` teardown (see CLAUDE.md "Database Handles in Tests").
- On `any-true` vs majority for the boolean merge: `any-true` makes the §2 tier collapse strictly worse (one provider row flagging `reasoning=1` → premium for every provider). Not a blocker — the issue allowed either if documented — but I'd lean majority. Maintainer call.

Repro for 2–3 (run inside the checkout, with `DATA_DIR` pointed at a scratch dir):

```ts
import { getDbInstance } from "./src/lib/db/core.ts";
import { getModelsDevTierFitnessWithSource, invalidateCapabilitiesCache } from "./open-sse/services/autoCombo/taskFitness.ts";
const db = getDbInstance();
db.exec(`CREATE TABLE IF NOT EXISTS model_capabilities (provider TEXT NOT NULL DEFAULT '', model_id TEXT NOT NULL, tool_call INTEGER, reasoning INTEGER, limit_context INTEGER)`);
const ins = db.prepare("INSERT INTO model_capabilities (provider,model_id,tool_call,reasoning,limit_context) VALUES ('x',?,1,1,400000)");
for (const id of ["gpt-5.2-codex","openai/gpt-5.2-codex","google/gemini-2.0-flash","claude-3-5-sonnet-20241022","chatgpt-4o-latest","gpt-4-0125-preview"]) ins.run(id);
invalidateCapabilitiesCache();
for (const id of ["gpt-5.2-codex","openai/gpt-5.2-codex","google/gemini-2.0-flash","claude-3-5-sonnet-20241022","claude-3-5-sonnet-20241022-high","chatgpt-4o-latest-free","gpt-4-0125-preview-high"])
  console.log(id, "→", JSON.stringify(getModelsDevTierFitnessWithSource(id, "coding")));
```

PR #10060, opened 2026-08-11
fix: use Node 22 LTS + raise build heap to 6144 MB -- Node 24 crashes… [defer to 3.8.51]

davenamovich:
… better-sqlite3 during build cleanup

## Summary

- Describe the user-facing or operational change.

## Related Issues

- Closes #
- Related to #

## Validation

Choose the change type and focused loop from the
[Contribution Golden Path](../docs/dev/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite,
Vitest, the 60% coverage gate, and the production build all run in CI on this PR (#8329):

- [ ] Change type: provider / routing / UI / i18n / CLI / DB / build-deploy / other
- [ ] Focused tests and category gates from the golden path
- [ ] `npm run lint`
- [ ] Reconciled with the current active release base; focused checks rerun afterward
- [ ] Production-code changes include a new or updated automated test in this PR
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented below

## Tests Added Or Updated

- List every changed or added automated test file.
- If no production code changed, state that here.

## Coverage Notes

- If this PR changes `src/`, `open-sse/`, `electron/`, or `bin/`, explain which tests cover the change.
- If coverage moved down in any touched file, explain why and what follow-up task will recover it.

## Reviewer Notes

- Call out any risky areas, migrations, feature flags, or manual validation that reviewers should know about.

PR #11647, opened 2026-08-26
fix(docker): drop arm64 Next.js build to 1 page-data worker, the 2-worker fix still OOMs

PauloFH:
## Why

`Build Docker (linux/arm64)` has been red on **every** push to `release/v3.8.51` (and `release/v3.8.50`) since ~2026-08-24 23:xx UTC — 60+ consecutive failures, including on draft #11442 — always the same signature:

```
ERROR: process "... npm run build ..." did not complete successfully: cannot allocate memory
```
right after `Collecting page data using 2 workers`, on the native `ubuntu-24.04-arm` GitHub-hosted runner.

## What changed

The 2026-08-25 fix (`OMNIROUTE_BUILD_WORKERS` 8→3, i.e. 7→2 actual Next.js page-data workers, #10060) modeled 2 workers as fitting the 16 GB runner's budget (`tests/unit/docker-build-memory-budget.test.ts`: `6144 + 2×2560 = 11264 MB` under the `12288 MB` budget) — but that model's `WORKER_PEAK_MB=2560` inference undercounted the real per-worker RSS, so the test kept passing a config that was red on every real run.

This PR:
- Drops `OMNIROUTE_BUILD_WORKERS` `3 → 2` (Next derives `workers = CIRCLE_NODE_TOTAL - 1`, so this is `2 → 1` actual worker) — the floor the regression test still allows.
- Raises `WORKER_PEAK_MB` `2560 → 4096` in the regression test so it now correctly rejects the previous 2-worker config (`6144 + 2×4096 = 14336 MB > 12288 MB` budget) instead of silently passing a red config.
- New worst case: `6144 + 1×4096 = 10240 MB`, ~2 GB under the 12288 MB budget (vs. the previous ~1 GB modeled margin that turned out to be negative in practice).
- Documents the incident inline in `Dockerfile`, the regression test, and `docs/guides/DOCKER_GUIDE.md` (matching the existing comment convention for these ARGs — `DOCKER_GUIDE.md` was flagged as out of sync by Copilot review and is now fixed too).
- Adds `changelog.d/fixes/11647-docker-arm64-single-worker.md` per `changelog.d/README.md`.

## Verification

- `node --experimental-strip-types --test tests/unit/docker-build-memory-budget.test.ts` → 3/3 pass with the new Dockerfile values.
- Confirmed the updated `WORKER_PEAK_MB` would have failed the *previous* (2-worker) config, so this is a real regression guard, not just a number bump.
- **Not verified against a live `ubuntu-24.04-arm` build** — the `docker-publish.yml` workflow only triggers on `push` to `main`/`release/v*` (not on PR from a fork), so this PR's checks can't exercise the actual arm64 build. This is the most defensible next step per the codebase's own documented playbook (same lever as the 8→3 fix), but treat a real push to `release/v3.8.51` as the actual test.

## Scope

`Dockerfile`, `tests/unit/docker-build-memory-budget.test.ts`, `docs/guides/DOCKER_GUIDE.md`, `changelog.d/`. `Dockerfile.bun` does not use `OMNIROUTE_BUILD_WORKERS`/`OMNIROUTE_BUILD_MEMORY_MB` or run this build step, so it's unaffected and untouched.

## Known unrelated CI red on this PR

`Docs Gates (fast-path)`, `No new ESLint warnings`, and 4× `Unit Tests fast-path` are currently failing on this PR, but none touch anything this diff changed:
- Docs Gates: pre-existing provider-count drift (`354` documented vs `356` live) in `README.md`/`PROVIDER_REFERENCE.md`/SVG diagrams — unrelated to Docker build args.
- ESLint: `There are suppressions left that do not occur anymore` (stale `eslint-suppressions.json` entries) — unrelated to this diff.
- Unit tests: failures are in `chatCore`/Codex-passthrough and `live-ws-eaddrinuse-6324` tests — unrelated to the Dockerfile/build-memory-budget test touched here.

Consistent with the already-open `release/v3.8.51` base-red issue (#11449). Not fixing these here to keep this PR scoped to the arm64 OOM.


RaviTharuma:
Cross-checked against the latest release Docker failures: both architectures compile and then terminate with BuildKit `ResourceExhausted` / `cannot allocate memory`. This PR is the smallest sufficient fix: it reduces actual Next page-data workers from 2 to 1 and leaves about 2 GB modeled headroom; the updated regression test passes 3/3. The overlapping #11663 makes the same runtime change with a larger diff. Recommend merging this PR, closing #11663 as duplicate, then treating the first successful real amd64+arm64 publish as the acceptance test. This blocks the versioned image tracked in #10296.

RaviTharuma:
Superseded by #11663, merged as `124f4cf7` into `release/v3.8.51`.

That merge contains the same operative fix as this PR: `OMNIROUTE_BUILD_WORKERS=2` (one Next.js page-data worker), updates the Docker guide and memory-budget regression test, and adds the changelog fragment. This branch is now conflicting with the release branch, so keeping a second implementation open would only duplicate the merged change.

Closing as superseded. The post-merge Docker publish is being tracked separately because run `33024696933` still failed both architecture jobs.

Issue #11611, opened 2026-08-26
fix(resilience): EXPIRED_RETRY_MAX budget bypasses the #8182 terminal-skip for non-no_refresh_token GitHub Copilot connections


Meet6338-X:
## Summary

Two shipped features are fighting over the same connection, and the unit shards have been paying for it on every run since the retry budget landed:

1. **#8182 terminal-skip**: a GitHub Copilot connection parked at `testStatus: "expired"` must be skipped by `checkConnection()` *unless* it is the exact recoverable shape (`errorCode === "no_refresh_token"` → #5326 self-heal).
2. **Expired-retry budget** (`isRecoverableExpiredWithRetryBudget`, `EXPIRED_RETRY_MAX`): any expired connection with fewer than N recorded retries is deliberately un-skipped so it can be retried with backoff.

The budget check does not exclude GitHub access-token-only Copilot connections whose `errorCode` is something other than `no_refresh_token` (e.g. `invalid_grant` after an operator ban). For those, condition 3 evaluates true (`lastErrorType` undefined ≠ `account_deactivated`, retry count 0 < max), the terminal skip is bypassed, and the sweep probes a connection #8182 explicitly decided was wasted work.

Observed live in CI (Unit Tests fast-path shard 3, job 98041152224, run of 2026-08-26):

```
test at tests/unit/token-health-no-refresh-token-expired-5326.test.ts
✖ checkConnection still skips a GitHub Copilot connection expired for a
  non-no_refresh_token reason (#8182 boundary)
AssertionError: 'github_access_token_invalid' !== 'invalid_grant'
```

The probe ran (`refreshCopilotToken` → stubbed 401) and overwrote the row's classification — exactly the wasted probe + state clobbering #8182 closed.

## Location

- File: `src/lib/tokenHealthCheck.ts`
- The conflict: `terminalStatuses` guard vs `isRecoverableExpiredWithRetryBudget` (both in `checkConnection()`, ~lines 586–606)
- Contract test: `tests/unit/token-health-no-refresh-token-expired-5326.test.ts` ("#8182 boundary" case)

```ts
const isRecoverableGithubCopilotNoRefresh =
  conn.testStatus === "expired" &&
  conn.errorCode === "no_refresh_token" &&
  isGitHubAccessTokenOnlyConnection(conn);
...
const isRecoverableExpiredWithRetryBudget =
  conn.testStatus === "expired" &&
  conn.lastErrorType !== "account_deactivated" &&
[...]

Issue #8182, opened 2026-07-22
fix(resilience): credits_exhausted providers still consume CPU despite cooldown config


manchairwang:
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

win10

### Node.js Version

node 22

### escription
Providers marked with credits_exhausted test status remain active and are continuously retried, causing high CPU usage (~53%) and unnecessary network requests. The cooldown mechanism should automatically prevent these providers from being routed to, but this behavior is not working as expected.

### Root Cause Analysis
1. The .env file cooldown configuration ( PROVIDER_COOLDOWN_ENABLED=true , PROVIDER_COOLDOWN_MIN_MS=3600000 ) was not passed to the OmniRoute process environment, even though logs showed "Loaded env from /root/.omniroute/.env"
2. 11 providers remained in active state despite having credits_exhausted test statusco

### Steps to Reproduce

1. configure some provider which always fail, then use them.

### Expected Behavior

- When a provider returns credits_exhausted , OmniRoute should automatically mark it as inactive or apply permanent cooldown until the user manually re-enables it
- Providers with credits_exhausted status should NOT be retried on subsequent API requests
- Users should NOT need to manually write database scripts to disable exhausted providers

### Actual Behavior

- When a provider returns credits_exhausted , OmniRoute keeps trying before the user manually re-enables it
- Providers with credits_exhausted werer retried on subsequent API requests
- Users have to manually write database scripts to disable exhausted providers

### Test Impact

Unsure

### Error Logs / Output

```shell
Provider list before fix:

=== Active but exhausted providers ===
1. api-airforce/main - testStatus: credits_exhausted
2. bazaarlink/main-2 - testStatus: credits_exhausted
3. bazaarlink/1 - testStatus: credits_exhausted
4. hyperbolic/1 - testStatus: credits_exhausted
5. openai-compatible-chat-5acc61ac-02d7-4c7c-9e9b-f0d8b5036631/git - testStatus: credits_exhausted
6. openrouter/6 - testStatus: credits_exhausted
7. openrouter/5 - testStatus: credits_exhausted
8. openrouter/4 - testStatus: credits_exhausted
9. openrouter/2 - testStatus: credits_exhausted
10. openrouter/1 - testStatus: credits_exhausted
11. sambanova/1 - testStatus: credits_exhausted
Total: 11

Process environment before systemd fix:
[...]

PR #11653, opened 2026-08-26
test(resilience): correct the expired-retry assertion for GitHub access-token-only connections (#11611)

HouMinXi:
### Summary

Refs #11611. The production half of #11611 (option (a): exclude GitHub access-token-only connections from `isRecoverableExpiredWithRetryBudget`) already landed on the base via #11608. What was left red was the boundary test it orphaned — `tests/unit/token-health-no-refresh-token-expired-5326.test.ts` asserted the pre-#11608 probe-with-budget behaviour, which #11608 made unreachable for this class.

### Changes
- Rewrites the boundary test to assert the post-#11608 skip: a non-`no_refresh_token` expired GitHub access-token-only connection is skipped, with zero fetch calls, even while retry budget remains.
- Records in the test comment why the skip is structural rather than a policy choice: the `!conn.refreshToken` branch returns unconditionally before the `EXPIRED_RETRY_MAX` backoff block, so the budget never governed this class.

### Test Plan
- 7/7 tests pass in `tests/unit/token-health-no-refresh-token-expired-5326.test.ts`.
- `npm run typecheck:core` clean.

Refs #11611


abhisheksharma2411 (commented):
Nice catch getting the shard green — that test really was red, and I verified that independently rather than taking it on trust. But I think the PR as written says something different from what it does, and there's a design question underneath that I'd hate to see closed by deletion.

**The described production change isn't in the diff.** The summary says *"Scoped `isRecoverableExpiredWithRetryBudget` in `src/lib/tokenHealthCheck.ts`"*, but:

```
changed_files = 1
modified  +20/-21  tests/unit/token-health-no-refresh-token-expired-5326.test.ts
```

and `isRecoverableExpiredWithRetryBudget` doesn't exist anywhere on the base — I grepped the whole tree at `91aeca044`, not just that file. So this is a test-only change described as a `fix(resilience)` that `Closes #11611`. If #11611 is "the shard is red", that's fair and the title should say so; if #11611 is a behaviour bug, this doesn't close it, because no behaviour moved.

**Verified both directions against unchanged production code**, swapping the base test in and *confirming the swap took effect before running* — I've been burned by a silent no-op revert reporting a false green before:

```
old test present : 1     new test absent : 0     src/ changed : 0

  OLD test, unchanged src  ->  6 pass / 1 fail
     ✖ checkConnection probes an expired GitHub connection that still has retry budget
  NEW test, unchanged src  ->  7 pass / 0 fail
```

So the old assertion was simply not true of the code. Correcting it is legitimate.

**What I'd push back on is what the deletion takes with it.** The removed comment wasn't decoration — it recorded a rationale:

> *#8182 skipped every "expired" connection … That was too wide: a transient OAuth failure parks a healthy connection at "expired" and it could then never come back.*

That concern is real, and the code confirms it. `EXPIRED_RETRY_MAX` does exist:

```ts
// Retry expired connections with exponential backoff up to EXPIRED_RETRY_MAX times.
if (conn.testStatus === "expired") {
  const retryCount = conn.expiredRetryCount ?? 0;
  if (retryCount >= EXPIRED_RETRY_MAX) return;
```

but look at what sits above it — the no-refresh-token branch ends in an **unconditional `return`**, outside the `refreshCapableNeedsReauth` guard:

```ts
    if (refreshCapableNeedsReauth) {
      ...
    }
    return;      // <-- fires either way
  }

  // Retry expired connections with exponential backoff up to EXPIRED_RETRY_MAX times.
```

So the retry budget is **structurally unreachable for every connection with no refresh token**. It isn't that the exemption was considered and rejected for this class — it's that the branch returns before the budget is ever consulted. The deleted test was aspirational: someone wrote it against a guard they expected to exist, and it never landed.

That reframes the change. Right now it converts *"a red test documenting a gap"* into *"a green test asserting the gap is intended"*, and the reasoning that would let a future reader tell those apart is deleted rather than answered.

**Concretely, what I'd suggest — none of it large:**

1. Retitle to something like `test(resilience): correct the expired-retry assertion for GitHub access-token-only connections`, and drop the claim about scoping a function that doesn't exist. `Refs #11611` rather than `Closes`, unless #11611 really is just the red shard.
2. Keep a short version of the deleted rationale in the new test — one or two lines noting that the retry budget is unreachable on this path by construction, so the skip is structural rather than a policy decision. That's the sentence that stops this being relitigated in three months.
3. Worth confirming with a maintainer whether a transient 401 permanently parking a healthy access-token-only connection is acceptable. It may well be — these accounts genuinely can't refresh — but that's a product call, and it's currently being settled by a test edit.

Happy to be wrong on the framing if #11611 is scoped to the shard; the measurements above are the part I'd stand behind either way.

HouMinXi:
Thanks for the detailed review. The both-directions verification is solid. Three clarifications with git evidence, and I'm taking suggestions 1 and 2 either way.

1) The production change the summary describes is in the base, not missing.

`isRecoverableExpiredWithRetryBudget` does exist at the merge base `91aeca044`, and it already carries the scoping the summary describes:

    $ git grep -n isRecoverableExpiredWithRetryBudget 91aeca044
    91aeca044:src/lib/tokenHealthCheck.ts:599:  const isRecoverableExpiredWithRetryBudget =
    91aeca044:src/lib/tokenHealthCheck.ts:616:    !isRecoverableExpiredWithRetryBudget

The `!isGitHubAccessTokenOnlyConnection(conn) &&` clause was added by #11608 (commit 1d476e4ce), which is an ancestor of this branch's fork point. So the branch is deliberately test-only: #11608 shipped the production half (option (a) in #11611), and the boundary test it orphaned is what is red on the shard. The summary wording is my mistake: it describes the net effect relative to #11611 instead of the diff. Retitling to `test(resilience): ...` and switching to `Refs #11611` per your suggestion 1.

2) The deleted test was not aspirational. It landed green.

Timeline, all 2026-08-26 UTC:

    05:27  #11611 filed (retry budget bypasses the #8182 terminal skip)
    11:09  #11592 merged: boundary test rewritten to assert probe-with-budget.
           Green against the guard as it existed then.
    12:28  #11608 merged: adds the GitHub-token-only exclusion. Flips that test red.
    15:44  this PR: pins the post-#11608 skip.

So the old assertion matched the shipped guard for 79 minutes, until #11608 narrowed it. It was not written against a guard that never landed.

3) Your structural reading is correct, and it is exactly the rationale worth keeping.

For connections without a refresh token, the `!conn.refreshToken` branch returns unconditionally (`src/lib/tokenHealthCheck.ts:836` at the base; the GitHub sub-branch returns at `:796`), so the `EXPIRED_RETRY_MAX` backoff block at `:839` is unreachable for this class. The budget was never a real budget here, just a bypass that reopened #8182's wasted-probe problem. I'll restore a two-line version of that in the new test per your suggestion 2, so the skip reads as structural rather than a policy call.

On suggestion 3 (transient 401 permanently parking a healthy access-token-only connection): agreed that is a maintainer decision, not something to settle with a test edit. Happy to split it into its own issue. The spot that would need a budgeted re-probe is the 401 branch that writes `github_access_token_invalid`.

diegosouzapw:
Obrigado pela contribuição — mas encontrei uma colisão real de decisão de produto ao validar este lote. A issue #11611 (ainda aberta, sem decisão explícita do dono) lista 3 resoluções possíveis para o conflito entre o guard #8182 (terminal-skip) e o retry-budget de conexões GitHub Copilot expiradas. Neste mesmo lote, #11672 escolheu independentemente a resolução (b) ('budget vence' — sonda mesmo sem `no_refresh_token`) com evidência e justificativa completas, e já foi mergeada. Este PR assume a resolução (a) ('guard vence' — pula a sonda) e ajusta o teste para bater com esse lado. As duas são mutuamente exclusivas — não dá pra mergear ambas.

Por ora estou mantendo o comportamento do #11672 (já em produção) e segurando este PR sem mergear, até o dono decidir explicitamente entre (a)/(b)/(c) na #11611. Se a decisão for (a), este PR volta a fazer sentido diretamente.

abhisheksharma2411 (approved):
I got this wrong, and you were right from the start. Correcting it plainly since you changed the PR on the strength of it.

**My claim that `isRecoverableExpiredWithRetryBudget` "doesn't exist anywhere on the base" was false.** It is at `src/lib/tokenHealthCheck.ts:599` on the same base sha I quoted, `91aeca044`. I said I had grepped the whole tree; whatever I actually grepped, it was not that.

**Worse, the design question I said was "deleted rather than answered" had already been answered — in production code, three and a half hours before I posted.** #11608 merged at 12:28 on 26 Aug and added exactly the scoping your original summary described:

```diff
   const isRecoverableExpiredWithRetryBudget =
     conn.testStatus === "expired" &&
     conn.lastErrorType !== "account_deactivated" &&
+    // GitHub access-token-only connections have their own dedicated exemption
+    // (isRecoverableGithubCopilotNoRefresh above): ONLY the exact
+    // "no_refresh_token" shape self-heals. An "expired" GitHub connection for a
+    // different reason (e.g. invalid_grant) is genuinely terminal and must stay
+    // skipped, otherwise the generic retry-budget exemption below reopens #8182's
+    // wasted-probe fix for every "expired" GitHub connection.
+    !isGitHubAccessTokenOnlyConnection(conn) &&
     getExpiredRetryCount(conn) < EXPIRED_RETRY_MAX;
```

That comment names #8182 and states the policy explicitly. So my framing — "this turns a red test documenting a gap into a green test asserting the gap is intended, with the reasoning deleted" — was simply wrong. The reasoning is in the code, it is deliberate, and this PR is the orphaned boundary test catching up to it. Which is what you said.

My "the retry budget is structurally unreachable" argument was also analysing the wrong path: for this class the early return at 617 fires because the predicate is false *by design*, long before the `!conn.refreshToken` branch I was quoting.

The one part that still stands is narrow and much less interesting: this PR's diff is test-only. But that was never a problem once the production half is a separate merged PR, which is precisely how you have now written it.

**On the changes I pushed you into:** don't revert them. `Refs #11611` plus a body that explains the split across #11608 and this PR is clearer than the original, and the structural note you added to the test is worth keeping on its own merits. The retitle is fine either way. So the outcome is good, but you spent three commits getting there because of a bad measurement from me, and that is the part I am sorry about.

One tiny thing, and I am deliberately not asserting it this time: I could not see #11608 in the blame for the `isRecoverableExpiredWithRetryBudget` declaration itself, only for the clause added inside it — so if anyone reads the body literally as "#11608 introduced the function", that is the one word worth loosening. The substance of the attribution is right.

Approving. Sorry for the noise.

diegosouzapw:
Holding this one — its premise is stale, not the test's fault.

This PR assumes #11608's GitHub-specific carve-out
(`!isGitHubAccessTokenOnlyConnection(conn)` inside `isRecoverableExpiredWithRetryBudget`)
is still live ("the production half of #11611 ... already landed on the base via #11608").
It was, when this PR was likely drafted — but #11672 (merged the same day, ~8h later)
deliberately reverted that exact carve-out, citing that it "contradicted the boundary
pinned by #11592." #11592's own PR body documents the intentional policy: an expired
connection with retry budget remaining should be *probed* (self-heal from a transient
OAuth failure), regardless of provider — only `account_deactivated` stays terminal.

Verified directly against the current release/v3.8.51 tip: `checkConnection` does call
`fetch` once for a GitHub access-token-only connection expired with `invalid_grant` and
retry budget remaining (matches #11592/#11672's live behavior). This PR's rewritten
assertion (`fetchCallCount === 0`) fails against that — not an environment artifact, the
production code genuinely doesn't skip this case anymore.

Merging this as-is would re-encode #11608's already-superseded position as the "correct"
one; if someone later "fixed" production to match it, that would silently re-open the
bug #11592/#11672 fixed (a GitHub connection with a transient OAuth failure permanently
stuck at `expired`, never re-probed).

This looks like a genuine three-way policy question between #11592, #11608, and #11672
rather than an orphaned-test cleanup — worth the owner's call on which boundary is
actually intended for GitHub access-token-only connections going forward, rather than me
picking a side. Leaving open rather than merging or closing.

diegosouzapw:
The diff against the current release/v3.8.51 tip is now empty (0 files changed) — the production half landed via #11608 (merged 2026-08-26, "resolve inherited release/v3.8.51 base-red CI failures"), and the boundary test this PR would have rewritten (tests/unit/token-health-no-refresh-token-expired-5326.test.ts) already matches the post-#11608 shape on the tip, per your own last commit here ("restore the #11592 retry-budget probe pin after merge"). Nothing left to land. Closing as subsumed/no-op — thanks for tracking this through the #11592/#11608/#11672 conflict to resolution.

PR #11660, opened 2026-08-26
test: add unit tests for formatFreeBudget in opencode-plugin

f9td56dbgh-hub:
## Summary

Adds unit test coverage for `formatFreeBudget()` in `@omniroute/opencode-plugin/src/naming.ts`, a small pure function that formats a free-tier model's budget info into a short human-readable suffix (e.g. `"25M tokens/day"`, `"(keyless)"`). It had no direct test coverage — only exercised indirectly through `buildModelDisplayName`.

## Changes

- `@omniroute/opencode-plugin/tests/naming.test.ts` (new): 7 tests covering all 6 `FreeModelFreeType` branches (`recurring-daily`, `recurring-monthly`, `recurring-credit`, `one-time-initial`, `keyless`, `discontinued`) plus the default-to-0 token/credit case.
- `@omniroute/opencode-plugin/package.json`: registered the new test file in the `test` script, matching the existing convention.

## Test plan

- [x] `npm test` in `@omniroute/opencode-plugin/` — 348/350 passing (the 1 pre-existing failure in `scaffold.test.ts` is unrelated: it requires a prior `npm run build` to produce the gitignored `dist/` output, and is not affected by this change).
- [x] New tests verified against the current implementation of `formatFreeBudget` / `fmtTokens`.

No production code was changed — test-only addition.

### DIFF
--- a/@omniroute/opencode-plugin/package.json
+++ b/@omniroute/opencode-plugin/package.json
@@ -23,7 +23,7 @@
   "scripts": {
     "build": "tsup",
     "clean": "rm -rf dist",
-    "test": "node --import tsx/esm --test tests/scaffold.test.ts tests/auth.test.ts tests/options-schema.test.ts tests/multi-instance.test.ts tests/fetch-interceptor.test.ts tests/provider.test.ts tests/gemini-sanitize.test.ts tests/combos.test.ts tests/config-shim.test.ts tests/features.test.ts tests/feature-defaults.test.ts tests/usable-combo.test.ts tests/disk-snapshot-perms.test.ts tests/fork-features.test.ts tests/auto-combo-context.test.ts tests/provider-id-routing.test.ts tests/management-read-token.test.ts tests/auto-sync.test.ts tests/model-allowlist.test.ts tests/log-level.test.ts tests/effort-tier-variants.test.ts",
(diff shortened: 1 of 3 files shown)


abhisheksharma2411 (commented):
Good addition — `formatFreeBudget` being covered only indirectly through `buildModelDisplayName` is exactly the kind of gap worth closing, and registering the file in the `test` script rather than relying on a glob matches how the rest of the package does it. Test-only, no production risk, easy yes from me.

Two things I'd offer, both about where the remaining risk actually sits rather than anything wrong with what's here.

**The `default:` branch isn't covered, and it's the one that can fire on real data.** The summary says the 7 tests cover "all 6 `FreeModelFreeType` branches", which is true of the named ones, but the switch has a seventh arm:

```ts
    default:
      return "";
```

TypeScript makes that unreachable for a well-typed caller, so it looks like dead code — except `freeType` arrives from catalog data, not from a literal. An upstream free type this build doesn't know about lands there and yields an empty suffix, which then flows into the display name. A single `formatFreeBudget({ freeType: "some-new-type" as FreeModelFreeType })` pins that it degrades to `""` rather than `undefined` or a throw, and it's the assertion most likely to matter when the catalog gains a type.

**Every value in the suite is an exact power of ten, which is the one input class `fmtTokens` handles cleanly.** The interesting behaviour is at the crossovers, and there's a real wart there today:

```
        999999  ->  1000K     <-- rounds past its own magnitude
     999999999  ->  1000M     <-- same
        999500  ->  999.5K
     999499999  ->  999.5M
          1500  ->  1.5K
            -5  ->  -5
           0.5  ->  0.5
```

`toFixed(1)` rounds `999999/1e3` to `1000.0`, and the `>= 1e6` branch has already been skipped by then — so you get `"1000K"` where `"1M"` is meant, and `"1000M"` for just under a billion. Not this PR's bug, and I wouldn't ask you to fix it here. But the suite as written would stay green through it, and a test file named for this function is where someone will later look to find out whether that output is intended.

If you'd rather keep this strictly characterisation, pinning `999_999 -> "1000K"` as *current behaviour* is still worth more than not covering it — it makes the wart visible and gives whoever fixes it a test to flip. Adding `0`, a sub-1K value, and one crossover would cover the arithmetic without expanding scope.

Neither is a blocker — happy to see this land as-is if you'd rather keep the diff tight.

f9td56dbgh-hub:
Thanks for the review! Pushed 695761d with both suggestions:

- a test pinning the `default:` branch's `""` fallback for an unrecognised `freeType`
- a sub-1K value and the `999_999` crossover, with a comment characterising the existing `toFixed(1)` rounding wart (`999999` -> `"1000K"`) as documented current behaviour rather than fixing it here, per your suggestion to keep this test-only.

Issue #11324, opened 2026-08-24
fix(providers): adding a custom model should not trigger full /models sync


axetroy:
### OmniRoute Version

3.49.0

### Installation Method

npm (global)

### Operating System

macOS

### Description

There are potentially hundreds, even thousands, of models coming from the `/model` interface.

I don't want to synchronize models from this interface because it's too difficult to manage.

Even if I manually click to add a model, it will already sync.

### Steps to Reproduce

https://github.com/user-attachments/assets/efc067be-eab9-43aa-b83d-e96e15ac955e

### Expected Behavior

None

### Actual Behavior

None

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

diegosouzapw:
Progress update: #11664 (merged into `release/v3.8.51`) adds the opt-out capability at the hook layer — `useApiKeySave.handleSaveApiKey` now honors a `skipModelSync` flag to skip the automatic full-catalog `/sync-models` call. Keeping this open — no UI control (a checkbox in the add-connection modal) wires up that flag yet, so end users can't actually trigger the skip through the dashboard today. That needs a follow-up PR, plus a decision on the checkbox's default/wording, which wasn't settled during triage.

PR #11663, opened 2026-08-26
fix(docker): re-derive build memory budget from measured per-process RSS (#7518)

diegosouzapw:
Refs #7518

⚠️ base-red inherited: #11449 — ESLint errors (5), docs-sync migrations count

## Scope

This PR addresses only the **residual, currently-live** defect in #7518 — the Docker
build's memory budget model, not the original complaint (default branch / Docker Hub
image), which the owner's 2026-08-23 comment already confirmed is fixed and out of
scope here.

## Root cause

`tests/unit/docker-build-memory-budget.test.ts` (added by PR #11419, merged
2026-08-24) modeled the Docker build's worst-case memory as
`heapMb + workers * WORKER_PEAK_MB` with `WORKER_PEAK_MB = 2560` — a value the
test's own comment labeled an **inference** ("7 workers didn't fit" → per-worker
peak "somewhere north of ~1.8 GB"), not a measurement. It also implicitly assumed
the parent `next build` process's RSS tracked `OMNIROUTE_BUILD_MEMORY_MB` (the V8
heap ceiling).

The issue owner's own live VPS reproduction (2026-08-24 04:49 UTC) measured the
real number directly via `dmesg`'s OOM-killer report:
`next-build (v16) ... anon-rss:4522744kB` (~4.5 GB) **per process**, independent
of the `NODE_OPTIONS --max-old-space-size` V8 heap ceiling — Turbopack itself is
native/Rust and compiles outside the V8 heap, and the figure applies to the
parent process too, not just workers.

Recomputing the budget with the measured ~4.5 GB/process figure against the
current default (`OMNIROUTE_BUILD_WORKERS=3`, i.e. 1 parent + 2 workers) gives a
worst case of `3 * 4500 = 13500 MB`, which exceeds the `12288 MB` (75% of a 16 GB
GitHub Actions runner) budget — exactly matching the still-live "Publish to
Docker Hub" failures (`cannot allocate memory` at "Collecting page data using 2
workers") even after #11419 merged.

## Fix

- Lowered `Dockerfile`'s `ARG OMNIROUTE_BUILD_WORKERS` default `3 -> 2` (1 parent +
  1 worker = 2 processes × 4.5 GB = 9 GB, fits the 12.288 GB budget with headroom).
- Re-derived `tests/unit/docker-build-memory-budget.test.ts`'s budget model on the
  measured ~4.5 GB/process figure, applied uniformly to every process (parent +
  workers) instead of conflating the parent process with the V8 heap ceiling.
[...]

PR #11664, opened 2026-08-26
fix(dashboard): skip full /sync-models catalog fetch when caller opts out (#11324)

diegosouzapw:
Refs #11324

## Root cause
`useApiKeySave.handleSaveApiKey` (`src/app/(dashboard)/dashboard/providers/[id]/hooks/useApiKeySave.ts`) unconditionally POSTs `/api/providers/{connectionId}/sync-models` right after any new connection is saved for a non-curated provider. That endpoint imports the entire upstream `/models` catalog — hundreds/thousands of entries for a large gateway/OpenAI-compatible provider — with no way to opt out, exactly matching the reporter's complaint of a flooded model list after adding a connection to manually add one model.

## Fix
`handleSaveApiKey` now accepts an optional `skipModelSync: true` field on its `formData` argument. When set, the automatic post-save full catalog sync is skipped (in addition to the existing curated-provider check). The flag is a client-side intent signal only — it is destructured out of the payload before the connection-creation POST, so it never reaches `/api/providers`. Default behavior (auto-sync on save) is unchanged when the flag is omitted, preserving existing UX for normal-size providers.

## Scope note (partial fix — using `Refs`, not `Closes`)
This PR adds the opt-out capability at the hook layer, proven by a regression test. It does **not** yet wire a UI control (e.g. a checkbox in the Add Connection modal) that lets an end user actually pass `skipModelSync: true` when adding a connection — the plan-file flagged that as needing UX confirmation on wording/default. Until that follow-up lands, no caller in the codebase sets the flag yet, so this PR is infrastructure for the real fix rather than a complete resolution of the reported symptom.

## Regression test
`src/app/(dashboard)/dashboard/providers/[id]/__tests__/useApiKeySaveSkipsFullSync.test.tsx` (Vitest):
- confirms the current code fires `/sync-models` unconditionally (RED before the fix, reproducing the plan-file's proof)
- asserts `skipModelSync: true` suppresses the `/sync-models` fetch entirely and never leaks into the `/api/providers` POST body
- asserts default/legacy behavior (no flag) still fires the sync, unchanged

```
npx vitest run "src/app/(dashboard)/dashboard/providers/[id]/__tests__/useApiKeySaveSkipsFullSync.test.tsx"
✓ 2 passed (2)
```

## Gates run
- `node scripts/check/check-complexity.mjs` — OK (2684 violations vs baseline 2774)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1208 violations vs baseline 1223)
- `npm run typecheck:core` — clean, exit 0
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json` (changed files) — exit 0
- Existing related test `tests/unit/bug-10096-kimi-coding-apikey-save.test.ts` — still passes unaffected (2/2)

⚠️ base-red inherited: #11449 — ESLint errors (5), docs-sync migrations count


diegosouzapw:

PR #11667, opened 2026-08-26
fix(kie): correct 12 more mismatched KIE Market model ids (#11296)

diegosouzapw:
Refs #11296

⚠️ base-red inherited: #11449 — ESLint errors (5), docs-sync migrations count (both pre-existing on `release/v3.8.51`, unrelated to this diff)

## Root cause

PR #11326 (merged for #11225) fixed the 4 `google-imagen/*` KIE Market ids and asserted every other Market namespace "already matches its real upstream id byte-for-byte." That claim was never actually checked against docs.kie.ai for anything beyond `google-imagen`, and it was wrong: `open-sse/handlers/imageGeneration.ts`'s `KIE_MARKET_UPSTREAM_MODEL_IDS` map sent several more catalog ids to `POST https://api.kie.ai/api/v1/jobs/createTask` unchanged even though they diverge from KIE's documented `model` value — exactly reproducing the reporter's post-fix complaint ("GPT Image 2, Flux, Seedream... The model name you specified is not supported").

## TDD

A repro test (`tests/unit/kie-market-upstream-id-11296-probe.test.ts`, reused verbatim from the triage plan-file) proved 5/5 assertions RED against unfixed `release/v3.8.51` code:

```
✖ GPT Image 2 T2I: actual 'gpt/gpt-image-2-text-to-image' vs expected 'gpt-image-2-text-to-image'
✖ GPT Image 2 I2I: actual 'gpt/gpt-image-2-image-to-image' vs expected 'gpt-image-2-image-to-image'
✖ GPT Image 1.5 T2I: actual 'gpt/gpt-image-1.5-text-to-image' vs expected 'gpt-image/1.5-text-to-image'
✖ Seedream 5.0 Lite T2I: actual 'seedream/5.0-lite-text-to-image' vs expected 'seedream/5-lite-text-to-image'
✖ Flux 2 Pro T2I: actual 'flux/2-pro-text-to-image' vs expected 'flux-2/pro-text-to-image'
```

## What this PR fixes (verified individually against the literal example request JSON on docs.kie.ai — not inferred by pattern)

| Catalog id | We sent | KIE actually expects |
| --- | --- | --- |
| `gpt/gpt-image-2-text-to-image` | `gpt/gpt-image-2-text-to-image` | `gpt-image-2-text-to-image` |
| `gpt/gpt-image-2-image-to-image` | `gpt/gpt-image-2-image-to-image` | `gpt-image-2-image-to-image` |
[...]

PR #11665, opened 2026-08-26
fix(combo): attach combo diagnostics to round-robin and runtime-unit retry-limit 503s (#11462)

diegosouzapw:
Closes #11462

## Root cause

`open-sse/services/combo.ts` has (at least) two independent implementations of the "global attempt budget exceeded" terminal branch:

1. Inside `handleComboChatInner`'s `executeTarget` (priority/weighted/generic strategies) — already uses `errorResponseWithComboDiagnostics(503, ..., buildComboDiag(failureReason))`, attaching `poolSize`/`attemptOrder`/`excluded`/`terminalReason` (plus `x-omniroute-combo-*` headers and an optional `recovery_hint`).
2. Inside `handleRoundRobinCombo` (round-robin strategy) — used a bare `errorResponse(503, "Maximum combo retry limit reached")` with none of that context.

A third occurrence in `open-sse/services/combo/runtimeUnits.ts` (the nested runtime-unit loop used by pipeline/fusion combo steps via `dispatchPrelude.ts` / `fusionPanel.ts`) had the identical gap.

This is a genuine inconsistency: the diagnostics feature the issue requests already exists and ships for some strategies but was never extended to round-robin or the nested runtime-unit path.

## Fix

- `handleRoundRobinCombo`'s terminal max-attempts branch now returns `errorResponseWithComboDiagnostics(503, ..., { poolSize, attempted, excluded, attemptOrder, terminalReason: "max_attempts_exceeded", recovery: buildRecoveryHint(...) })`, built from state already in scope (`exhaustedProviders`, `exhaustedConnections`, `rrOutcomes`, `modelCount`, `globalAttempts`) — additive only, no reordering of existing round-robin logic.
- `executeRuntimeUnitCombo`'s attempt-budget-exceeded branch now returns the same diagnostics-carrying helper, with its own small `poolSize`/`attemptOrder`/`terminalReason` mapping built from the ordered runtime units (units, not models/providers, have a different shape from the RR/priority target list).

## Regression tests (TDD)

- `tests/unit/combo-rr-diagnostics-11462.test.ts` — round-robin combo that exhausts `maxGlobalAttempts` across two different providers; asserts the 503 body carries `diagnostics.{poolSize,attemptOrder,terminalReason}`. RED before the fix (`AssertionError: round-robin 503 should carry a diagnostics field`), GREEN after.
[...]

PR #11668, opened 2026-08-26
fix(sse): cap streaming headers-wait timeout to a client-realistic ceiling (#11526)

### LINKED ISSUE #11526: Responses->Chat: turn stalls with 120s keepalive then client 499 abort after tool_search is flattened (NVIDIA upstreams)
## Description

On the `codex -> /v1/responses -> OmniRoute (openai-responses -> openai chat translation) -> NVIDIA Chat Completions upstream` path, after ~2 rounds of successful tool calls the translated upstream request stalls: OmniRoute streams only `: keepalive` to the client for exactly ~120 s, no data chunks ever arrive from the provider, and Codex aborts with **499 "Request aborted"** (`duration` ≈ 120011 ms, matching the client timeout exactly).

Notably the failing next turn is the one **containing `function_call_output` items for the previous tool calls** — i.e. the turn where `tool_search`-adjacent tool results are being forwarded. A `tool_search` call in that session is consistently shown as `aborted` in the Codex TUI, and afterwards the whole conversation stalls in the same 120 s keepalive loop before aborting.

This is a *new* symptom that remains after the fixes for #2766 / #7532 / #7936 / #8145 / #8146 landed: the translation itself now succeeds (`openaiRequest` and `providerRequest` are both built and sent), but the provider never answers once the conversation carries a round of tool results produced after a flattened `tool_search` call.

## Steps to Reproduce

1. Configure Codex to use OmniRoute with a combo that bridges `openai-responses -> openai` (e.g. any NVIDIA-hosted `openai` format model such as `deepseek-ai/deepseek-v4-flash-0731`, also observed with `stepfun/step-3.7-flash` and `kimi-k3` on a `nvidia/` account).
2. Ask the agent to call multiple tools in parallel, one of which causes the model to call `tool_search` (deferred tool discovery, `execution: "client"`).
3. The `tool_search` call shows as `aborted` in the client; the following turn stalls for 120 s and is aborted by the client.

Control experiment: the same 10-tool test through `codex -> cc-switch (v1/chat) -> OmniRoute (openai -> openai passthrough) -> moonshotai/kimi-k3` succeeds for 10/10 tools, including `tool_search` and the lazy-loaded `ctx_*` tools. Only the in-OmniRoute Responses→Chat translation path fails.

## Expected Behavior

Each translated turn completes normally; `tool_search`-triggering conversations keep working on Chat-Completions-only providers.

## Actual Behavior

- Request logs show `status: 499`, `error: "Request aborted"`, `duration: ~120011 ms`, `hasResponseBody: false`.
- `pipelinePayloads.streamChunks.client` contains **only** `data: {"type":"response.in_progress"}` followed by repeated `: keepalive` entries — zero data chunks from upstream within the full 120 s window.
- The translated `providerRequest.body` (POST `https://integrate.api.nvidia.com/v1/chat/completions`) looks formally valid: 16 `function` tools including the flattened namespace (`codex_app__*`) tools and `tool_search` flattened into a normal Chat function — whose `function.description` is the full Codex tool-discovery prompt, i.e. **a single tool entry of 14253 bytes** inside a single ~85 KB request body.

## Hypothesis (needs confirmation)

Two non-exclusive candidates, both worth instrumenting:

1. **Upstream silent rejection/hang on the flattened `tool_search` function.** The 14 KB freeform `# Tool discovery ...` description (and/or its unusual JSON schema) may cause some Chat-Completions providers to stall instead of returning a 4xx. A request-side toggle to send tool_search as a compact stub (or drop it + pre-inject deferred tools, i.e. #7532 "Option C") would isolate this.
2. **Response-side SSE translator deadlock on fan-out tool calls.** The stalled turn contains **two parallel `function_call_output` items**; if the Chat→Responses stream state machine waits for an event pairing that never arrives (e.g. per-output_index sequencing for sibling tool calls), OmniRoute could be holding upstream data without forwarding it — though `streamChunks` shows *no upstream chunks captured at all*, which currently favors (1).

I can attach the full request/response/log export (22 translated requests incl. pipelinePayloads) if useful.

## Related

- #2766, #7532 (tool_search handling), #7936 / #8145 / #8151 (namespace round-trip), #8146 (tool_search_call skip) — all landed; this is a residual symptom on top of them.
- Cross-project precedent for reference: LiteLLM handles built-in/client-executed tools differently from plain functions in its Responses⇄Chat bridge (denylist + explicit handling), and vLLM's `StreamingState` shows the explicit-state-machine approach for the response side.

## Environment

- OmniRoute: current main (translation code matches tip incl. `namespaceToolIdentityMap` and tool_search flattening branch)
- Client: Codex desktop/CLI, Windows
- Providers where observed: NVIDIA NIM (deepseek, stepfun, kimi-k3); kimi-k3 succeeds on both paths

diegosouzapw:
Closes #11526

⚠️ base-red inherited: #11449 — ESLint errors (5), docs-sync migrations count (both pre-existing on `release/v3.8.51`, not introduced by this diff)

## Root cause

Two watchdogs govern a stalled upstream on the Responses->Chat translated path, and they were asymmetric:

1. **Body-phase** (`resolveStreamReadinessTimeout`, `open-sse/utils/streamReadinessPolicy.ts`) — fires after headers arrive, tool-heavy/large-history adaptive. For the reported payload shape it computes well under Codex's own ~120s client-abort patience.
2. **Headers-phase** (`fetchStartTimeoutMs` in `BaseExecutor.execute()`, backed by `FETCH_TIMEOUT_MS`, default 600_000ms) — governs the wait *before any response arrives at all*, flat and non-adaptive.

The issue's own log field `hasResponseBody: false` shows watchdog #1 never engaged — the upstream (an NVIDIA target) never returned a response at all. The only ceiling left was the 10-minute default, 5x longer than a real client's patience, so OmniRoute kept the connection alive on keepalives only, guaranteeing the client aborted first with an opaque 499 instead of OmniRoute detecting and failing the stall fast.

## Fix

Adds `resolveFetchStartTimeout()` (`open-sse/utils/fetchStartTimeoutPolicy.ts`) that caps the headers-wait timeout to 110s (under Codex's ~120s abort window) **for streaming requests only** — non-streaming requests keep the existing flat `FETCH_TIMEOUT_MS` default unchanged, per the plan-file's risk note (some providers are legitimately slow to accept a connection on non-streaming paths). Wired into `BaseExecutor.execute()` at the one call site that computes `fetchStartTimeoutMs`.

The existing `TimeoutError` classification path (`chatCore.ts`) already maps a fetch-start timeout to HTTP 504 (`GATEWAY_TIMEOUT`), which is in the circuit-breaker-eligible `{408,500,502,503,504}` set per `AGENTS.md` — confirmed by reading the code, no change needed there.

## Scope note

This PR addresses the confirmed, in-our-code, fixable gap (the headers-phase timeout asymmetry). The plan-file's secondary hardening item — truncating the oversized (14KB) flattened `tool_search` tool description for non-Codex-native chat-completions upstreams — is explicitly deferred pending reporter/VPS confirmation that it's the actual NVIDIA-side trigger, so `Closes` is used because this PR is the full fix for the reported stall-then-499 symptom (the headers-phase ceiling), while that separate follow-up hardening remains open as a distinct, unconfirmed item.

## Testing (TDD, Hard Rule #18)

`tests/unit/issue-11526-repro.test.ts` — new regression test:
- confirms the body-phase watchdog is fine (stays under Codex's ~120s patience) for the reported payload shape
- confirms the headers-phase watchdog for streaming requests is now bounded under 120s (would fail on unfixed code — `resolveFetchStartTimeout` didn't exist)
- scope guards: non-streaming requests keep the flat `FETCH_TIMEOUT_MS` default; a base timeout already under the cap is left untouched

```
✔ issue #11526: body-phase readiness watchdog stays comfortably under Codex's ~120s patience for the reported tool-heavy payload shape
✔ issue #11526 (fixed): headers-phase watchdog for STREAMING requests is bounded under Codex's ~120s patience
✔ issue #11526 scope guard: non-streaming requests keep the flat FETCH_TIMEOUT_MS default
✔ issue #11526 scope guard: a base timeout already under the cap is left untouched for streaming requests
ℹ tests 4, pass 4, fail 0
```

Also re-ran the existing suites touching the changed symbols:
- `tests/unit/stream-readiness-policy.test.ts` + `tests/unit/runtime-timeouts.test.ts` — 33/33 pass
- `tests/unit/executor-default-base.test.ts` (BaseExecutor.execute / DefaultExecutor, incl. "BaseExecutor.execute clears the startup timeout after headers arrive") — 51/51 pass

## Gates run

- `node scripts/check/check-complexity.mjs` — OK (2684 violations, baseline 2774)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1208 violations, baseline 1223)
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json <changed files>` — clean, 0 errors/warnings

## Changelog

`changelog.d/fixes/11526-responses-headers-phase-timeout-cap.md`


diegosouzapw:

Issue #2901, opened 2026-05-29
[BUG] OpenCode Free combo entries use opencode/ prefix instead of oc/


kefasjw:
### OmniRoute Version

3.8.5

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Provider(s) Involved

OpenCode Free / opencode / oc / opencode-zen

### Model(s) Involved

OpenCode Free models, e.g. big-pickle

### Description

OpenCode Free is defined as the no-auth provider with provider ID `opencode` and alias `oc`, but combo builder entries serialize the model string with the provider ID (`opencode/<model>`) instead of the routing alias (`oc/<model>`).

This causes OpenCode Free targets added through the combo builder to be routed incorrectly. `opencode/<model>` is manually mapped in the model parser to `opencode-zen`, which is not treated as the no-auth OpenCode Free provider. As a result, combo-routed OpenCode Free requests can fail credential/no-auth resolution instead of using the intended free `oc/<model>` route.

### Steps to Reproduce

1. Open the dashboard combo builder.
2. Add an OpenCode Free model to a combo.
3. Save the combo.
4. Inspect the saved combo model step or execute the combo.
5. Observe that the model string is saved/routed as `opencode/<model>` instead of `oc/<model>`.

Relevant code paths:

- `src/shared/constants/providers.ts`
  - `NOAUTH_PROVIDERS.opencode.id = "opencode"`
  - `NOAUTH_PROVIDERS.opencode.alias = "oc"`

- `src/lib/combos/builderOptions.ts`
  - no-auth combo builder providers use `providerId = noAuthProvider.id`
  - model options build `qualifiedModel` as `${providerId}/${modelId}`

- `src/lib/combos/builderDraft.ts`
  - combo steps are saved as `model: ${providerId}/${modelId}`

- `open-sse/services/model.ts`
  - `opencode` is manually mapped to `opencode-zen`

### Expected Behavior

OpenCode Free combo entries should use the provider alias/routing prefix:

```text
oc/<model>
```

A saved combo step should preserve internal identity separately from the routing string, for example:

```json
{
  "kind": "model",
  "providerId": "opencode",
  "model": "oc/big-pickle"
}
```

Runtime routing should resolve `oc/<model>` to the no-auth OpenCode Free provider.

### Actual Behavior

Combo builder saves OpenCode Free targets as:

```text
opencode/<model>
```

[...]

PR #11669, opened 2026-08-26
fix(sse): keep each functionCall's own thoughtSignature on parallel Gemini tool-call turns (#11510)

diegosouzapw:
Closes #11510

## Root cause

On a PARALLEL tool-call turn (2+ tool_use/tool_calls in one assistant message) against a Gemini 3.x model, `open-sse/translator/request/claude-to-gemini.ts` and `open-sse/translator/request/openai-to-gemini.ts` only re-attached the resolved `thoughtSignature` to the *first* functionCall of the turn, and silently dropped it for every subsequent tool call in the same message -- even when a real, previously-stored signature existed for it (`resolvedSignatures.get(id)` had a valid value). Gemini 3.x then rejects the request with HTTP 400 "Function call is missing a thought_signature in functionCall parts" (reproduces the reporter's screenshot: `default_api:WebFetch`, model `gemini-3.5-flash`).

This restriction was introduced in commit c485c5e0 (PR #9015, fixing #8979) to handle the single-tool-call case; it was never extended to preserve per-call signatures on parallel/multi-tool-call turns.

## Fix

Each functionCall part now carries its own resolved `thoughtSignature` unconditionally -- dropped the `shouldUseEmbeddedSignature`/`firstPersistedSignature` gating in both translator files, so every tool call in a parallel turn keeps its individually-resolved signature instead of only the first one.

## Regression tests (TDD, Hard Rule #18)

- `tests/unit/issue-11510-parallel-tool-thought-signature.test.ts` (claude-to-gemini path) -- confirmed RED against unfixed code (`webFetchPart.thoughtSignature` was `undefined` instead of the stored `SIG_11510_SECOND`), GREEN after the fix.
- `tests/unit/issue-11510-parallel-tool-thought-signature-openai.test.ts` (openai-to-gemini path, mirror of the same defect) -- same RED->GREEN pattern.
- Existing signature-handling suites re-run to confirm the single-tool-call scenarios are unaffected: `claude-gemini-thought-signature-8979.test.ts`, `t43-gemini-tool-call-no-thought-signature.test.ts`, `t44-antigravity-thought-signature-preserved.test.ts`, `translator-claude-to-gemini.test.ts`, `translator-openai-to-gemini*.test.ts`, `openai-to-gemini-helpers-split.test.ts`, `openai-to-gemini-gemma-thinking.test.ts` -- 86/86 pass.

## Known residual limitation (out of scope)

[...]

PR #11670, opened 2026-08-26
fix(combos): use the routing-alias prefix, not raw providerId, when serializing combo model steps (#11433)

### LINKED ISSUE #11433: fix(providers): opencode alias override makes OpenCode Free untargetable in combo builder
## Problem

The alias map contains this manual override:

```ts
ALIAS_TO_PROVIDER_ID["opencode"] = "opencode-zen";
```

This is intentional for user-typed `opencode/` routing prefixes. But `opencode` is also the **canonical provider ID** for the distinct no-auth OpenCode Free provider (declared in `src/shared/constants/providers/noauth.ts` with `id: "opencode"` and `alias: "oc"`).

Because the combo builder constructs a combo step's `model` field by prefixing with `providerId`:

```ts
// builderDraft.ts
model: `${normalizedProviderId}/${normalizedModelId}`
// → "opencode/big-pickle"
```

any step built from the OpenCode Free provider option produces `model: "opencode/<model>"`. When that string is later parsed, `parseModel()` calls `resolveProviderAlias("opencode")` and resolves to `opencode-zen` — overriding the operator's explicit selection.

The result is a step that simultaneously stores:

```text
providerId = "opencode"    ← selects OpenCode Free
model      = "opencode/…"  ← routes to OpenCode Zen
```

These are different providers. Different subsystems—connection selection, lockout, health monitoring, circuit-breaker checks, and model-metadata lookup—can therefore treat the same step as either Free or Zen depending on which field they read.

There is no way to target OpenCode Free through the combo builder's normal provider/model selector. Manual entry of `oc/<model>` works, but nothing in the UI communicates this, and the selector produces the conflicting form.

## Why this is hard to observe

Several workarounds partially compensate. Auth hydration explicitly pairs `opencode` and `opencode-zen` as siblings (#7993), and bare-model inference checks both family members. These suppress visible errors in specific paths but do not make the underlying identity consistent.

## Proposed direction

The combo step schema needs to separate these concepts:

| Concept | Current state | Proposed state |
|---|---|---|
| Canonical provider ID | `providerId: "opencode"` | Preserve as database identity |
| Routing prefix | Embedded in `model` using provider ID | Use the provider's declared routing alias |
| Qualified model | `model: "opencode/<model>"` | `model: "oc/<model>"`, or use a structured representation |

Concretely, `buildPrecisionComboModelStep()` could use the provider's declared routing alias (`oc`) rather than its canonical ID when constructing the qualified model string, ensuring that the structured provider identity and parsed model route agree.

Alternatively, canonical aliases and routing-prefix overrides could use separate, explicitly named maps. A canonical provider ID would then never be silently reinterpreted as another provider by a generic identity resolver.

## Relationship to prior PRs

This is the underlying identity collision behind feedback on #11403 and #11404. Those PRs addressed symptoms at the executor and parser layers.

The assertion that `ALIAS_TO_PROVIDER_ID["opencode"] = "opencode-zen"` is valid is correct in the routing-prefix domain. The problem is that `opencode` is simultaneously a registered canonical provider ID used in configuration and combo serialization. A parser-only change cannot reconcile both meanings without breaking one of them.

## Offer

I am happy to draft updated PRs for whichever direction reaches consensus—separating routing-prefix and canonical-alias resolution, changing combo serialization to use the declared provider alias, or a structured representation. I am opening the issue first so the identity contract can be agreed before more code is proposed.

diegosouzapw:
Closes #11433

⚠️ base-red inherited: #11449 — ESLint errors (5), docs-sync migrations count

## Root cause

`buildPrecisionComboModelStep()` (`src/lib/combos/builderDraft.ts`) rebuilt a step's `model` string as `${providerId}/${modelId}` from the canonical provider id. For the no-auth "OpenCode Free" provider (id `opencode`) this produces `opencode/<model>`, but `opencode` also doubles as a manual routing-prefix override in `open-sse/services/model.ts` (`ALIAS_TO_PROVIDER_ID["opencode"] = "opencode-zen"`) meant only for user-typed `opencode/` prefixes referring to the distinct, paid "OpenCode Zen" provider. So a combo model step built through this path silently resolved to the wrong provider on request routing, even though `step.providerId` correctly recorded `"opencode"`.

A prior fix (#2901) already rewrites the builder's display-only `qualifiedModel` field to the correct `oc/` alias prefix (`src/lib/combos/builderOptions.ts::rewriteQualifiedModelPrefix`), but `buildPrecisionComboModelStep()` never consumed that corrected value — it always rebuilt `model` from the raw canonical `providerId`. Every path that reaches it (precision single-select, global model search, and manual `oc/<model>` entry) was affected. One path was already correct: the "Select all" batch picker stores `qualifiedModel` verbatim rather than rebuilding it.

## Fix

Thread the already-computed routing-alias prefix through via a new optional `modelPrefix` param on `buildPrecisionComboModelStep()` — when provided, it is used (not `providerId`) to build `model`, while `step.providerId` always stays the canonical id (routing/duplicate-detection identity unaffected):

- `src/app/(dashboard)/dashboard/combos/page.tsx`: `builderCandidateStep` and `handleAddBuilderStep` now pass `modelPrefix` derived from `selectedBuilderModel.qualifiedModel`.
- `buildGlobalModelList()`: derives `modelPrefix` from each model's `qualifiedModel`.
- `buildManualComboModelStep()`: captures the user-typed prefix (e.g. `oc`) before resolving it to the canonical provider id, and preserves it instead of collapsing back to the canonical id.
- `findNextSuggestedConnectionId()` needed no change — its duplicate check (`getExactModelStepSignature`) keys off `entry.providerId`, not the parsed model prefix.

`ALIAS_TO_PROVIDER_ID`, `resolveProviderAlias()`, and the `routingPrefix` computation in `builderOptions.ts` are untouched — they were already correct (#2901) and are the source of truth this fix threads through.

## Regression test (TDD)

`tests/unit/issue-11433-opencode-alias-collision.test.ts` — reproduces the bug (confirmed RED against `origin/release/v3.8.51` unfixed code: `step.model` parses back to `"opencode-zen"` instead of `step.providerId` `"opencode"`), then confirms GREEN for the fixed `buildPrecisionComboModelStep`/`buildGlobalModelList`/`buildManualComboModelStep` paths.

Also updated one pre-existing assertion in `tests/unit/combo-builder-draft.test.ts` (`buildManualComboModelStep resolves provider aliases...`) that encoded the old "collapse typed alias back to canonical id" behavior — the new, more general contract (preserve the user-typed prefix) is a legitimate behavior change per this fix, and does not change routing (the `cx` alias used in that test also resolves to `codex` with no manual override, so both spellings route identically).

## Gates run

- `node scripts/check/check-complexity.mjs` — OK, 2684 violations (baseline 2774)
- `node scripts/check/check-cognitive-complexity.mjs` — OK, 1208 violations (baseline 1223)
- `npm run typecheck:core` — clean, exit 0
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json` on changed files — 8 pre-existing errors in `page.tsx`, all outside this diff (lines 776/789/807/2107/2342/2347/2364/2409), byte-identical to `origin/release/v3.8.51` — confirmed pre-existing/base-red, not introduced here
- Focused tests, all green: `tests/unit/issue-11433-opencode-alias-collision.test.ts`, `tests/unit/combo-builder-draft.test.ts`, `tests/unit/combo-select-all-batch-8526.test.ts`, `tests/unit/combo-builder-eligible-connection-2057.test.ts`, `tests/unit/combo-account-allowlist-3266.test.ts` (8/8), `tests/unit/ui/combo-form-select-all-handlers.test.tsx` (vitest, 7/7)
- `src/lib/combos/builderDraft.ts` is not referenced by the `test:vitest` (MCP/autoCombo) suite — unaffected

Changelog fragment: `changelog.d/fixes/11433-opencode-alias-collision.md`


diegosouzapw:

PR #8123, opened 2026-07-22
fix(providers): discover live AGY models

adevwithpurpose:
## What changed
- reuse the Antigravity live model-discovery flow for AGY connections
- accept newly announced AGY chat models while excluding internal and tab-completion-only entries
- keep the bundled AGY catalog as the offline/bootstrap fallback and isolate in-flight discovery by provider

## Tests
- 
pm run lint
- 
pm run typecheck:core
- 
ode --import tsx/esm --test tests/unit/agy-provider.test.ts tests/unit/provider-models-route.test.ts (68 passed)
- 
pm run check:cycles
- 
pm run check:fetch-targets
- 
pm run check:tracked-artifacts
- 
pm run check:test-discovery
- 
pm run check:file-size

## Notes
- 
pm run test:unit progressed past 11,000 tests but did not complete locally: multiple unrelated test workers reported esbuild The service was stopped, then the run reached the 20-minute command timeout.
- 
pm run check:provider-consistency could not run because the installed Bun executable is incompatible with this Windows host.
- 
pm run check:changelog-integrity is not meaningful from main while the repository default branch is 
elease/v3.8.49; it reports the release branch's existing aggregated changelog bullets as missing.

Production changes are covered by 	ests/unit/agy-provider.test.ts and 	ests/unit/provider-models-route.test.ts.

Issue #7678, opened 2026-07-18
feat(providers): zai-web live model discovery — hardcoded catalog is stale and 404s


andrea-kingautomation:
## Summary

The `zai-web` provider ships a hardcoded 3-model catalog `{glm-4.6, glm-4.5, glm-4.5v}`, and **all three now 404** against the live chat.z.ai backend, so the provider is effectively unusable through the router even with valid cookies. `chat.z.ai` has moved its model lineup on to GLM-5.x, and there is no live discovery path for this provider (POST `/api/providers/{id}/sync-models` reports `syncedModels:3 / unchanged:3` because the sync falls back to the hardcoded list rather than querying the account).

This mirrors the kimi-web / K3 situation (#7549), where live `GetAvailableModels` discovery was the right fix. z.ai exposes an equivalent endpoint.

## Evidence (v3.8.48, 8 connected consumer accounts)

1. **Every cataloged model 404s through the router:**
   ```
   zai-web/glm-4.6      -> [404] Z.ai error: {"detail":"Not Found"}
   zai-web/glm-5.2      -> [404]  (rejected: not in catalog, but forwarded and 404s at z.ai too)
   zai-web/glm-4.7-flash-> [404]
   ```
2. **The stored credential is valid.** Each connection holds the raw `token` JWT (ES256, `eyJ...`). Hitting z.ai directly with it returns 200:
   ```
   GET https://chat.z.ai/api/models
     Authorization: Bearer <token>        -> 200
     Cookie: token=<token>                -> 200
   ```
3. **The live catalog (from that 200) is entirely GLM-5.x-led, none of which the router knows:**
   ```
   glm-5.2            (max_tokens 64064; caps: think, reasoning_effort, web_search, mcp, vision-less)
   GLM-5.1            (32000)
   GLM-5-Turbo        (32000)
   GLM-5v-Turbo       (vision)
   glm-4.7            (40000)
   glm-4.6v           (16000, vision)
   0727-106B-API / 0727-360B-API (80000)
   deep-research, zero, glm-4-flash, ...
   ```
   The provider's hardcoded `glm-4.6 / glm-4.5 / glm-4.5v` are all absent.

## Requested fix

[...]

Issue #2979, opened 2026-05-31
[Feature] Automatically Discover & Sync Models from Providers (Remove Hardcoded Model Lists)


artickc:
### Problem / Use Case

Currently, OmniRoute appears to rely on manually maintained model definitions. This creates a recurring issue where newly released models are unavailable until they are explicitly added to the codebase.

For example, Claude Opus 4.8 has been released, but it is not yet available in OmniRoute because the model list has not been updated.

This approach causes several problems:

Users cannot immediately access newly released models.
Developers must continuously monitor provider announcements and manually update model definitions.
The application can become outdated even when providers already expose the models through their APIs.
Users may assume a provider or API is broken when the actual issue is that the model is not hardcoded yet.

As AI providers are releasing new models frequently, manual maintenance does not scale well.

### Proposed Solution

Implement automatic model discovery and synchronization directly from provider APIs.

Instead of maintaining static model lists, OmniRoute could:

Query each provider's "List Models" endpoint during startup and/or on a scheduled refresh interval.
Cache discovered models locally.
Automatically populate model selectors from the discovered models.
Fall back to a small built-in list only when a provider does not expose a model discovery endpoint.
Allow users to trigger a manual refresh from the UI.
Possible Implementation Ideas
Option 1: Provider Model Discovery Service

Create a provider-agnostic interface:

```
interface ModelProvider {
  listModels(): Promise<Model[]>;
}

```
Each provider implementation would fetch available models from its native API.

Benefits:

Clean architecture
Easy to add new providers
Reduces maintenance burden

Option 2: Background Sync
Refresh models on application startup
Refresh every 12–24 hours
Store results in cache/database
Display "Last synced" timestamp

Benefits:

New models appear automatically
Minimal API overhead
Option 3: Hybrid Approach

Combine:

Auto-discovered models
User-defined custom models
Emergency fallback hardcoded list

This provides reliability while remaining future-proof.

Examples of Provider APIs

Many providers already expose model discovery endpoints:

OpenAI → /models
Anthropic → model listing endpoint (when available)
OpenRouter → models API
Google Gemini → model listing APIs
Groq → models API
Together AI → models API
Fireworks AI → models API

OmniRoute could leverage these endpoints instead of requiring manual updates whenever a provider releases a new model.

Expected Benefits
Immediate support for newly released models.
Reduced maintenance burden for developers.
Better user experience.
Fewer feature requests asking for newly released models.
[...]

Issue #488, opened 2026-03-20
Support auto-update for model lists in API Providers and OpenRouter


rdself:
For some third-party API providers and OpenRouter, they updates their model lists very frequently. I would suggest adding an auto-update function for the model list fetcher (automatically remove the outdated list, and then fetch the latest list), so users can always enjoy the latest models without updating every provider's model lists each time manually.

Thanks a lot!

PR #11688, opened 2026-08-26
feat(models): reactive discovery sync after Antigravity model-not-found 404

MumuTW:
## Summary

Pinned model catalogs go stale the moment the upstream backend ships or renames a model — Gemini 3.7 Flash on Antigravity (this cycle), zai-web (#7678), Claude Opus 4.8 (#2979). Requests 404 upstream until somebody re-freezes the catalog by hand. The 404 itself is the most precise trigger available, so this PR makes it self-healing for the Antigravity family:

- new src/lib/providerModels/reactiveModelSync.ts — on a model-not-found 404 the Antigravity executor kicks a discovery sync for that connection via the existing per-connection loopback sync (discovery → synced AvailableModels store). The fresh list lands in the synced catalog, so the next request resolves and /v1/models self-heals on the same path
- guardrails: provider allow-list (antigravity/agy), 10-minute per-connection cooldown, in-flight dedup; fire-and-forget with log-only failures — the scheduled auto-sync cycle (#488) stays the backstop
- syncConnectionModels is exported from modelSyncScheduler so the scheduled cycle and the reactive trigger share one loopback implementation

This is the smallest reactive slice of the accepted direction in #2979 (background discovery/sync). Extending to other discovery-capable providers is a matter of growing the allow-list.

## Related Issues

- Related to #2979 (accepted background discovery/sync direction)
- Related to #488 (model auto-sync scheduler), #8123 (live AGY discovery), #7678 (same staleness failure mode)

## Validation

- [x] Change type: provider
- [x] Focused tests: node --import tsx/esm --test tests/unit/reactive-model-sync.test.ts → 9/9 pass
- [x] Regression: model-sync-scheduler, model-sync-route, model-sync-custom-preservation, antigravity-streaming-error-body-sanitized-2461, repro-antigravity-404-family-cooldown-hijack → 33/33 pass
- [x] npx eslint clean on new/changed files (note: modelSyncScheduler.ts carries one pre-existing no-restricted-imports error on its untouched localDb import — part of the frozen CI error set, not introduced here)
- [x] node scripts/check/check-file-size.mjs → OK
- [x] Reconciled with current main (v3.8.50)

## Tests Added Or Updated

- tests/unit/reactive-model-sync.test.ts (new, 9 tests): trigger fires with correct args; agy id normalization; allow-list rejection; cooldown window no-op; per-connection isolation; cooldown expiry re-trigger; in-flight dedup under a held sync promise; failure tolerance after cooldown; test-helper restore.

## Coverage Notes

New module fully covered by the new test file (injected sync fn — no experimental mock flags needed). The executeAttempt wiring is the 404 branch only; behavior verified via the antigravity error-path regression tests above.

## Reviewer Notes

- Trigger point: sendAntigravityRequest's non-ok branch, status 404 only (model-not-found class). Other statuses untouched.
- Loopback sync reuses the scheduler's internal-auth fetch — no new auth surface.
- Worst case under a 404 storm: one sync per connection per 10 minutes.
- The synced-catalog merge for agy/antigravity already exists in the /v1/models builder (they are not curated-only providers), so no catalog-side change was needed.

### DIFF
--- a/open-sse/executors/antigravity/executeAttempt.ts
+++ b/open-sse/executors/antigravity/executeAttempt.ts
@@ -8,6 +8,7 @@
 import { mergeAbortSignals, type ExecutorLog } from "../base.ts";
 import { applyFingerprint, isCliCompatEnabled } from "../../config/cliFingerprints.ts";
 import { buildAntigravityUpstreamError } from "../antigravityUpstreamError.ts";
+import { maybeTriggerReactiveModelSync } from "@/lib/providerModels/reactiveModelSync.ts";
 import {
   HTTP_STATUS,
   STREAM_READINESS_TIMEOUT_MS,
@@ -383,6 +384,13 @@ export async function sendAntigravityRequest(
       "TELEMETRY",
       `[Antigravity] Error Response - URL: ${url}, Status: ${response.status}, Model: ${model}`
     );
+    if (response.status === HTTP_STATUS.NOT_FOUND) {
+      // The backend may have shipped/renamed models the synced catalog does not
+      // know yet (pinned-catalog staleness). Kick a discovery sync for this
+      // connection so the fresh list lands in the synced catalog and the next
+      // request can resolve. Cooldown + in-flight dedup live in the trigger.
+      maybeTriggerReactiveModelSync(provider, credentials.connectionId);
+    }
   }
 
   return { response, finalHeaders };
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Thanks for this! Its base (main) had diverged too far from the active release/v3.8.51 for a
clean retarget, and this fork rejects maintainer pushes, so a direct rebase-and-merge wasn't
possible. Recreated your commit (author preserved via cherry-pick) onto the current tip and
merged as #11888 — thank you for the reactive discovery-sync fix.

PR #11699, opened 2026-08-26
fix(memory): honor category filter in GET /api/memory

### LINKED ISSUE #11650: fix(backend): GET /api/memory?category=X filter ignored - always returns unfiltered results
# Bug: GET /api/memory?category=X 过滤失效，所有 category 参数返回相同结果

## 问题描述
调用 `GET /api/memory?category=codegraph` 时，过滤参数被完全忽略，返回的结果与 `GET /api/memory?category=anything` 完全相同（相同 100 条）。

## 复现步骤
```bash
# 预期：只返回 category=codegraph 的条目（10条）
curl -H "Authorization: Bearer <key>" \
  "http://127.0.0.1:20128/api/memory?category=codegraph"

# 预期：只返回 category=decision 的条目（8条）
curl -H "Authorization: Bearer <key>" \
  "http://127.0.0.1:20128/api/memory?category=decision"

# 实际：两次请求返回完全相同的 100 条数据
```

## 调查
查看源码 [src/app/api/memory/route.ts](https://github.com/diegosouzapw/OmniRoute/blob/main/src/app/api/memory/route.ts)，GET handler 只提取了以下参数：
```typescript
const apiKeyId = searchParams.get("apiKeyId") || undefined;
const type = (searchParams.get("type") as any) || undefined;
const sessionId = searchParams.get("sessionId") || undefined;
const query = searchParams.get("q") || undefined;
```

**缺少 `category` 参数提取**。`category` 不在传给 `memoryManager.list()` 的选项中，因此过滤完全失效。

## 期望行为
`category` 应作为有效过滤参数传入 `memoryManager.list()`，实现对 metadata.category 字段的过滤。

## 环境信息
- OmniRoute 版本: 3.8.x（本地运行中）
- API: `POST/GET /api/memory`
- 底层存储: Qdrant `omniroute_memory` 集合，payload 包含 `category` 字段

Sabeekhann:
## Summary

- Make `GET /api/memory?category=<value>` actually filter memories by `metadata.category` instead of returning the unfiltered list.
- Carry the category filter through the API route, shared `MemoryFilter`, SQLite backend, and store query.
- Apply the predicate before count, `byType` aggregation, and pagination so response metadata stays consistent with the filtered result set.
- Guard SQLite JSON extraction with `json_valid()` so malformed legacy metadata is ignored rather than turning a category-filter request into a 500.

## Related Issues

- Closes #11650

## Validation

- [x] Change type: other — memory API / persistence filtering
- [ ] Focused tests and category gates from the golden path — PR CI pending
- [ ] `npm run lint` — PR CI pending
- [x] Reconciled with the current active release base; branch was created from current `release/v3.8.51` tip `42a13fed`

## Tests Added Or Updated

- Added `tests/integration/memory-category-filter.test.ts`
  - verifies `category=codegraph` returns only matching rows with matching `total` and `byType`
  - verifies category composes with `type`, `sessionId`, `limit`, and `offset`
  - verifies an unknown category returns an empty result
  - verifies malformed JSON metadata does not fail the filtered endpoint
  - verifies omitting `category` preserves existing unfiltered behavior

## Coverage Notes

- The integration regression exercises the public `GET /api/memory` route against a real temporary SQLite database, covering route parsing, `memoryManager.list()`, SQLite backend forwarding, SQL filtering, count aggregation, and pagination together.
- No migration or schema change is required because `category` is stored in the existing JSON `metadata` column.

## Reviewer Notes

- Category matching is exact after trimming the query parameter, treating category as an identifier rather than free-text search.
- Malformed legacy metadata is treated as a non-match instead of causing an endpoint failure.
- No feature flag, migration, provider behavior, or memory search/retrieval behavior changes.
- Broad validation is intentionally left to PR CI; no unverified local pass counts are claimed.


hydraxman:
One backend-contract gap remains: `MemoryManager.list()` delegates to the configured primary backend, but this patch only consumes `MemoryFilter.category` in `SQLiteBackend`. `ObsidianBackend.list()` currently applies `apiKeyId`, `type`, and `sessionId` before aggregation but not `category`, while `GenericMemoryBackend.buildListQuery()` does not forward a category query parameter. On installations using either backend, the public route will still accept `category` and return an unfiltered result. Please either carry the filter through those two list implementations (including a configurable generic query-parameter name) or explicitly constrain the route contract to SQLite.

Sabeekhann:
Addressed in the latest commits.

- `ObsidianBackend.list()` now applies `MemoryFilter.category` against `memory.metadata.category` before aggregation and pagination.
- `GenericMemoryBackend` now forwards `category` from `buildListQuery()`.
- The generic backend's `queryParams` config now supports a configurable `category` name (default: `"category"`).
- Updated the generic backend unit test to cover a custom category parameter mapping.

The public route contract is now carried through SQLite, Obsidian, and Generic primary backends. CI will validate the updated branch.

Issue #7518, opened 2026-07-16
fix(docker): default branch is a draft release — clone + docker compose build gives an unbuildable tree


addiep:
Following the README's Docker path on a fresh Ubuntu 24.04 host, `git clone` lands on `release/v3.8.49` (the repo default), which is the draft living-release branch from #7076. Building it fails: `npm run build` dies ~3.7s in at "Creating an optimized production build …" with exit code 1 and no error output.

```
#20 [builder 10/10] RUN --mount=type=cache,id=next-cache,target=/app/.build/next/cache   mkdir -p /app/data && npm run build
#20 3.455 ▲ Next.js 16.2.10 (Turbopack)
#20 3.694   Creating an optimized production build ...
#20 ERROR: process "/bin/sh -c mkdir -p /app/data && npm run build" did not complete successfully: exit code: 1
```

It appears buildkit-specific — the identical builder image runs `npm run build` fine under a plain `docker run` (compiles for 5+ min), but dies instantly under `docker build`. Removing the `--mount=type=cache,id=next-cache` line doesn't change it.

Ruled out: OOM (6.2 GiB free, nothing in `dmesg`, and exit 1 rather than 137), and the heap exhaustion anticipated in the Dockerfile comments (#4076), which announces itself with a loud `FATAL ERROR: JavaScript heap out of memory` rather than silence.

Falling back to v3.8.42 fails differently: `npm rebuild better-sqlite3` silently skips the native compile ("13 packages had install scripts blocked because they are not covered by allowScripts"), then `Could not locate the bindings file`. That's fixed in newer tags by the explicit `node-gyp` call (#7160, #6700) — so there's no older tag that builds either.

There's also no published image for 3.8.49 (3.8.48 is newest on Docker Hub), which is consistent with it being unreleased.

Would it be possible to either point the default branch at the last released tag, or have the README steer people to `docker pull` rather than a source build? Happy to provide full logs.

Environment: Ubuntu 24.04.4, Docker 29.5.3, compose plugin v5.1.4, x86_64 (4 vCPU / 8 GB). Worked around by using the published `3.8.48` image pinned by digest, which runs fine.

drewbitt:
There should probably be a Docker build test in CI.

diegosouzapw:
[...]

PR #11719, opened 2026-08-26
fix(docker): bump Bun image to 1.4.0 with Turbopack and port the node image's build memory guards

### LINKED ISSUE #11709: fix(docker): v3.8.50 Docker Hub/npm publish failed — BUN base image build OOM (amd64+arm64)
## Summary

The v3.8.50 release (published 2026-08-26T19:30 UTC) never produced a Docker image — `ghcr.io/diegosouzapw/omniroute:latest` (and any `v3.8.50` tag) is still pointing at v3.8.49. Users on the official image cannot update to v3.8.50.

## Failed workflows for the `v3.8.50` tag

| Workflow | Run | Result |
|---|---|---|
| Publish to Docker Hub | https://github.com/diegosouzapw/OmniRoute/actions/runs/33005521002 | ❌ failure (partial) |
| Publish to npm | https://github.com/diegosouzapw/OmniRoute/actions/runs/33005520967 | ❌ failure |
| Build Electron Desktop App | https://github.com/diegosouzapw/OmniRoute/actions/runs/33005490476 | ❌ startup_failure |

## Root cause (Docker Hub publish)

Both platform jobs failed at the step **`Build and push BUN base platform image by digest`**:

- ❌ `Build Docker (linux/arm64)` — job https://github.com/diegosouzapw/OmniRoute/actions/runs/33005521002/job/98332510391
- ❌ `Build Docker (linux/amd64)` — job 98299038099 (same step)
- ✅ Regular **base** and **web** images built and pushed fine on both platforms
- ✅ `Resolve Docker release metadata` — success

The failing step's error (from the arm64 job log):

```
##[error]buildx failed with: ERROR: failed to build: failed to solve: ResourceExhausted:
process "/bin/sh -c bun run --quiet build" did not complete successfully: cannot allocate memory
```

So the BUN variant of the base image runs **out of memory on the GitHub Actions runner** during `bun run build`. Since the final multi-arch manifest + tag push step is never reached, **no new image tag was published at all** — `latest` remains v3.8.49.

## User impact

Anyone using the official image path (`image: ghcr.io/diegosouzapw/omniroute:latest` / docker-compose.prod.yml) gets v3.8.49 after `docker compose pull` and cannot reach v3.8.50:

```
$ docker pull ghcr.io/diegosouzapw/omniroute:latest
# digest unchanged -> still v3.8.49
```

## Suggested fixes

1. Reduce memory pressure of the BUN build (e.g. cap heap via `BUN_JSC_forceRAMSize` or equivalent, or build the BUN image in a separate job with reduced parallelism)
2. Or run the BUN base build on a larger runner (e.g. a 16 GB runner)
3. Re-run the publish pipeline after the fix so a proper `v3.8.50` image tag gets pushed
4. The npm publish failure may be related (worth re-checking); the Electron workflow `startup_failure` looks like a workflow-file/config error, separate from OOM

## Environment

- Watching from the consumer side: official Docker image, `docker-compose.yml` with `image: ghcr.io/diegosouzapw/omniroute:latest`, profile `base`
- Currently running: v3.8.49 (verified via `docker exec` package.json + dashboard)

rqzbeh:
fix(docker): bump Bun image to 1.4.0 with Turbopack and port the node image's build memory guards

## Summary

Fixes #11709 — "Publish to Docker Hub" has been red on every push to `main` since the `-bun` image targets landed (#11039/#11168): the Bun builder stage (`bun run build` in `Dockerfile.bun`) dies with `ResourceExhausted: ... cannot allocate memory` on the 16 GB GitHub-hosted runners (a Next.js build worker is SIGKILLed mid-compile) on **both** `linux/amd64` and `linux/arm64`. Because the multi-arch manifest step is never reached, no release image can be published — `latest` is still v3.8.49.

## Root cause

Two compounding problems, both unique to the Bun image path:

1. **Pinned obsolete Bun release.** `Dockerfile.bun` bases on `oven/bun:1.3.14-slim` (2026-05-13). Beside missing the Node-compat work that landed in Bun 1.4.0 (2026-08-20), that release cannot run Turbopack, so the `-bun` build is stuck on the webpack production pass — the memory-hungry path — while the node image builds with Turbopack by default. (Bun 1.4.0 also fixes a dependency-resolution bug: under 1.3.14, `bun install --include=optional` failed to resolve `playwright@1.62.1` / `postcss@^8.5.18` with bogus "No version matching" errors; 1.4.0 installs the full 2151-package tree cleanly.)
2. **Missing build memory guards.** `scripts/build/build-next-isolated.mjs` force-forces `--webpack` for any `process.versions.bun` run (historical V8-worker-binding workaround), and `Dockerfile.bun` sets **none** of the knobs the node `Dockerfile` carries (#10060/#11419): no `OMNIROUTE_BUILD_WORKERS`/`CIRCLE_NODE_TOTAL` (Next falls back to `os.cpus()-1` = 3 page-data workers on the 4-vCPU runners) and no `OMNIROUTE_BUILD_MEMORY_MB`/`NODE_OPTIONS` (the build script applies its 8 GB per-process heap default). 4+ V8 processes at multi-GB each exceed the 16 GB runner; the cgroup OOM killer SIGKILLs a worker and buildkit surfaces `cannot allocate memory`.

## Fix

- **Bump the Bun release**: `oven/bun:1.3.14-slim` → `oven/bun:1.4.0-slim` (the release with Turbopack + Node 16.3 support and the dependency-resolution fixes).
- **Enable Turbopack on Bun** (port of upstream #11471): `resolveNextBuildBundlerFlag` is now env-only (`OMNIROUTE_USE_TURBOPACK=0` → `--webpack`, else `--turbopack`) instead of sniffing the runtime, and `Dockerfile.bun` replaces the hardcoded `ENV OMNIROUTE_USE_TURBOPACK=0` with `ARG OMNIROUTE_USE_TURBOPACK=1` / `ENV` wiring — webpack stays as the explicit `--build-arg` escape hatch for Bun < 1.4.
- **Port the node image's memory guards into `Dockerfile.bun`**: `ARG OMNIROUTE_BUILD_MEMORY_MB=6144` → `ENV NODE_OPTIONS="--max-old-space-size=${OMNIROUTE_BUILD_MEMORY_MB}"` and `ARG OMNIROUTE_BUILD_WORKERS=3` → `ENV CIRCLE_NODE_TOTAL=${OMNIROUTE_BUILD_WORKERS}`. These bound the page-data workers on the default Turbopack path and keep the webpack fallback inside the 16 GB budget too. Worst case per the maintained model: `6144 + 2 × 2560 = 11264 MB` ≤ `12288 MB` (75 % of 16 GB).
- **Regression guards**: `tests/unit/docker-build-memory-budget.test.ts` now validates **both** images (`Dockerfile` and `Dockerfile.bun` — wiring regexes + budget arithmetic + vCPU oversubscription), and `tests/unit/build/resolve-next-build-bundler-flag.test.mjs` (ported from #11471) pins the env-only flag contract.

## Verification

- `node --experimental-strip-types --test tests/unit/docker-build-memory-budget.test.ts tests/unit/build/resolve-next-build-bundler-flag.test.mjs` → **9/9 pass** (run on the fleet worker).
- Full reproduction of the failing builder stage on the 16-core fleet worker with the fixed stack (`oven/bun` 1.4.0 equivalent + Turbopack + the caps above): `bun install --include=optional --quiet` (2151 packages, no resolution errors) → better-sqlite3 node-gyp rebuild → `bun run --quiet build` → **exit 0**, `▲ Next.js 16.3.1 (Turbopack)` banner, 2.2 GB standalone bundle produced, zero OOM-kills in dmesg. The same step OOM-kills before this change.
- Note: `docker-publish.yml` triggers only on pushes to `main` / `release/v*` / tags / releases, so PR CI does not exercise the Docker build itself; the fix is backed by the regression tests and the real builder-stage run above.

Closes #11709.


harkaranbrar7:
This looks right and matches the fix we were preparing on our side — the guards port into Dockerfile.bun plus the budget test covering both images is the correct approach.

One suggestion before merge: `OMNIROUTE_BUILD_WORKERS=3` (→ 2 page-data workers) is likely still too much. #11647 is dropping the Node image to 1 worker because the 2-worker budget still OOMs the arm64 runner, and #11663 measured ~4.5 GB per process — 2 workers plus the parent exceeds the 16 GB budget (6144 + 2×4500 = 15,144 MB > 12,288 MB). Defaulting to `OMNIROUTE_BUILD_WORKERS=2` (→ 1 worker), like #11663 does for the Node image, would avoid the Bun image hitting the same wall.

The Bun 1.4.0 + Turbopack bump is a good addition — just worth keeping in mind it makes this the first real Turbopack-on-Bun build, so if the OOM persists, a `OMNIROUTE_USE_TURBOPACK=0` control run would isolate whether it's Turbopack's native memory.

rqzbeh:
Thanks — adopted. Both images now default OMNIROUTE_BUILD_WORKERS=2 (1 page-data worker) against the measured ~4.5 GB per-process RSS from #7518/#11663: 2 processes × 4.5 GB ≈ 9 GB fits the 12,288 MB (75%) budget, where =3 (2 workers, 3 processes ≈ 13.5 GB) does not. The budget test now uses the measured figure for both Dockerfile and Dockerfile.bun and pins the =2 default (proven red at =3), so a future bump must re-do the arithmetic. #11663 landed on release/v3.8.51 after this PR was cut; the node Dockerfile here is aligned to =2 as well to keep the shared test honest on main. Good call on the control run: OMNIROUTE_USE_TURBOPACK=0 is now a --build-arg escape hatch on the Bun image, so an isolation run is one flag away if the OOM persists.

PR #11727, opened 2026-08-27
perf(compression): avoid cloning memo misses twice

RaviTharuma:
## Summary
- return freshly computed compression results after storing their defensive cache copy
- preserve defensive clones on both cache storage and cache hits
- cover mutation isolation and exact lookup counts for sync and async misses/hits
- prune the now-stale `@typescript-eslint/no-unused-vars` suppression for the updated regression test

## Validation
- `node --import tsx/esm --test tests/unit/compression/result-memo.test.ts` (34 passed)
- old post-store lookup restored temporarily: focused suite fails exact sync/async lookup assertions
- `npm run lint`
- `npm run typecheck:core`
- `npm run check:complexity`
- `npm run check:file-size -- --base-ref origin/release/v3.8.51`
- `git diff --check`

⚠️ base-red inherited: #11449

### DIFF
--- a/config/quality/eslint-suppressions.json
+++ b/config/quality/eslint-suppressions.json
@@ -4539,11 +4539,6 @@
       "count": 1
     }
   },
-  "tests/unit/compression/result-memo.test.ts": {
-    "@typescript-eslint/no-unused-vars": {
-      "count": 1
-    }
-  },
   "tests/unit/compression/rtk-grouping.test.ts": {
     "@typescript-eslint/no-unused-vars": {
       "count": 1
--- a/open-sse/services/compression/resultMemo.ts
+++ b/open-sse/services/compression/resultMemo.ts
@@ -4,6 +4,7 @@ import type { CompressionConfig, CompressionMode, CompressionResult } from "./ty
 export const MEMO_CAP = 5_000;
 
 const memoMap = new Map<string, CompressionResult>();
+let lookupCountForTests = 0;
 
 // Opt-IN whitelist (NOT opt-out): cache only engines proven pure + STATELESS across
 // requests. Excluded on purpose: `ccr` and `session-dedup` write to the cross-request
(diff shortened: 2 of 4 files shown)


RaviTharuma:
CI note: the focused compression memo regression, core typecheck, focused ESLint, complexity and file-size gates pass locally. Current required failures reproduce base/repository drift unrelated to this two-file PR: generated skills sync (`combo-routing` legacy key), docs count baseline (providers 42 vs 46), stale ESLint suppressions, broad quality baseline failures, and unit shard 1 terminating from worker heap exhaustion. Build, Vitest and unit shards 2-4 passed.

RaviTharuma:
Review-fix cleanup plan (before edits):
- remove the already-unused `callCount` local in the touched test file
- add the smallest existing-style `__ForTests` lookup counter seam in `resultMemo.ts`
- assert exact miss/hit lookup counts for both sync and async entry points while retaining mutation-isolation coverage
- prove the regression fails with the old post-store lookup restored, then run focused tests, core typecheck, focused lint and static gates

RaviTharuma:
Review fixes pushed as `fcb28e874`:
- exact test-only memo lookup counter now proves sync and async misses perform only the initial lookup; restoring either old post-store `memoLookup` makes the focused suite fail (expected 3/1, actual 4/2)
- mutation isolation remains covered across miss and repeated hits
- removed pre-existing unused `callCount` in the touched test

Corrected lint evidence: the earlier focused command used the repository suppressions file and exited 0 before this cleanup, but that command can fail globally when suppressions are stale. Current changed-file lint was therefore rerun against an empty suppressions file and exits 0: `eslint open-sse/services/compression/resultMemo.ts open-sse/services/compression/strategySelector.ts tests/unit/compression/result-memo.test.ts --suppressions-location /private/tmp/omniroute-empty-suppressions.json --max-warnings 0`. Also green: 31 focused tests, `typecheck:core`, complexity, file-size, and diff check.

RaviTharuma:
Correction: amended commit is `6f143cf03` (the earlier review comment named an intermediate local hash). PR head and validation described there are otherwise current.

Issue #8617, opened 2026-07-25
feat(quality): 107 files over the 800-line cap — live decomposition worklist


MumuTW:
## Why this issue exists

`#8584` proposal 3 asked for it: the three quality baselines carry **93 references** to
`structural shrink tracked in #3501`, and **#3501 is closed** — it was
`[refactor] Decompose providers/[id]/page.tsx god-component (12,882 LOC → ≤800)`, about a
different file, and it *succeeded*. So for weeks every rebaseline note has discharged its
structural-debt obligation to a completed ticket. The debt looks accounted for and is
assigned to nothing.

This is that live worklist. New rebaseline notes should reference this issue instead.

## Measurement

Against `release/v3.8.49` @ `4053e2314`, excluding `*.test.ts(x)` and `*.d.ts`:

```
git grep -c "" release/v3.8.49 -- 'src/**/*.ts' 'src/**/*.tsx' 'open-sse/**/*.ts' \
  'electron/**/*.ts' 'bin/**/*.ts' | sed 's|^release/v3.8.49:||' \
  | grep -vE '\.test\.tsx?:|\.d\.ts:' | awk -F: '$2>800'
```

**107 files over the 800 cap · 56,144 excess lines.**

| area | files >800 | excess lines |
| --- | ---: | ---: |
| `src/app/(dashboard)` + `src/shared/components` (UI) | **36** | **19,670** |
| `open-sse/handlers` | 7 | 7,846 |
| `open-sse/services` | 13 | 6,921 |
| `open-sse/executors` | 12 | 6,725 |
| `open-sse/utils` | 2 | 2,777 |
| `src/sse` | 3 | 2,815 |
| `src/app/api` | 5 | 2,564 |
| `src/lib/db` | 8 | 2,556 |
| `open-sse/mcp-server` | 3 | 1,596 |
| everything else | 18 | 2,674 |

Shape of the debt:

- **71 of 107 files (66%) sit in the 800–1200 band**, averaging +163 lines — one extraction
  each. But they are only **11,545 lines, 20% of the total**.
- **11 whales over 2,000 lines hold most of the mass.**
- Clearing every file needs a minimum of **142 new leaf modules** (`ceil(size/800) − 1`).

## Non-goals — do not start here

- **`open-sse/handlers/chatCore.ts` (4,906) and `open-sse/services/combo.ts` (3,640)** are
  scheduled on the public roadmap (3.8.53 and 3.8.52) and already have open PRs
[...]

PR #11728, opened 2026-08-27
refactor(quality): split adobeFireflyClient.ts (2,952 lines) into nine leaf modules

MumuTW:
Follow-up to the worklist in #8617 (closed), taking one file off it.

`open-sse/services/adobeFireflyClient.ts` was **2,952 lines**. This splits it into nine
leaf modules plus a barrel, each under the 1000-line cap.

## Why this file, and why this shape

The Adobe Firefly family is **already decomposed** — `adobeFireflySecurity.ts`,
`ModelSnapshot`, `References`, `Upscale`, `Models`, `Session`, `BrowserLogin` are all
siblings of this client. This PR continues that pattern rather than introducing a new
one; the new modules sit beside the existing ones with the same naming.

`check-file-size.mjs` rejects a *new* file above the cap outright (a new leaf has no
baseline entry to shelter it), so every extracted module has to land under 1000 in the
PR that creates it — no staged oversized intermediate. All nine do.

| module | lines | contents |
| --- | ---: | --- |
| `adobeFireflyCatalog.ts` | 282 | endpoint URLs, tunables, image/video model catalog |
| `adobeFireflyCredentials.ts` | 205 | IMS token / cookie-blob recognition and extraction |
| `adobeFireflyPayload.ts` | 479 | model resolution, generate-request payload construction |
| `adobeFireflyArp.ts` | 382 | nonces, ARP session ids, submit/upload headers |
| `adobeFireflyUpload.ts` | 347 | source-image extraction and upload to Firefly storage |
| `adobeFireflyResponses.ts` | 230 | upstream response parsing, job-status helpers |
| `adobeFireflyIms.ts` | 256 | IMS token exchange, access-token resolution |
| `adobeFireflyAccount.ts` | 157 | credits balance, model discovery |
| `adobeFireflyPoll.ts` | 155 | async job polling |
| `adobeFireflyClient.ts` | **693** | `generateImage` / `generateVideo` + barrel |

The dependency graph between the leaves is acyclic — `catalog` and `credentials` have no
dependencies, `client` sits at the top.

## Public surface is unchanged — and verified, not asserted

The barrel re-exports **explicitly by name**, not `export *`. That matters: nine helpers
that were module-private (`browserHeaders`, `sleep`, `GUEST_COOKIE_HELP`, the
`DEFAULT_*` tunables, `NANO_SIZE_MAP`, `PIXEL_SIZE_TO_RATIO`) had to become `export` to
cross the new module boundaries. With `export *` they would have leaked into the public
API; with an explicit list they stay invisible to consumers.

Verified by importing the pre-split file and the new barrel side by side in one process
and diffing them:

```
原始 export: 65 | 新 barrel: 65
遺失: []          多出: []
函式本體不同: 0   常數值不同: 0
```

Every exported function compares equal by `Function.prototype.toString()`, and every
exported constant by deep value — so the barrel hands consumers the same functions, not
merely functions with the same names and arities.

That runtime check is blind to types, and it hid a real regression at first: the
`export type { AdobeFireflyDiscoveredModel }` pass-through was dropped and nothing
caught it. Types are now checked separately — all six original type exports resolve
through the barrel, confirmed with a compile probe that was itself mutation-tested
(deleting one re-export moves the probe from 84 errors to 85; restoring it returns to
84, so the probe is actually sensitive rather than vacuously green).

## Verification

Every number below is measured against the same tree, with the pristine
`release/v3.8.51` baseline alongside it:

```
Adobe Firefly suites (6 files)   62 tests, 61 pass, 1 fail   IDENTICAL to pristine
  + image/video generation       179 tests, 178 pass, 1 fail
  the single failure is pre-existing: tests/unit/firefly-cookie-validation-10522.test.ts

tsc -p tsconfig.typecheck-core.json   9 errors before, 9 after, 0 mentioning adobeFirefly
                                      (the 9 are pre-existing `omniglyph` package errors)
check:file-size                       OK — new leaves all under the 1000 cap
prettier --check                      clean on all ten files
```

Two further adversarial checks before opening this:

- **Reconstruction diff.** Concatenated the nine leaves plus the barrel and compared the
  executable lines against the original: **1 line differs**, and it is a doc comment that
  sat exactly on a segment boundary. No statement was lost, added, or reordered.
- **Deliberate break.** An earlier revision of this split silently re-exported
  `resolveAdobeImageModel` from the pre-existing `adobeFireflyModels.ts` — which happens
  to define a *different* function of the same name and same arity. The name/arity check
  passed; the test suite caught it (`'gemini-flash-nano-banana-2' == 'nano-banana-pro'`).
  That is why the surface check now compares function bodies rather than signatures.

## What this PR does not claim

It does **not** move the complexity ratchet. `check:complexity` reports 2,684 both before
and after — splitting a file does not reduce per-function cyclomatic complexity, and the
90-point gap against the 2,774 baseline was already there. This PR is file-size debt only.


MumuTW:
Rebased onto `release/v3.8.51` @ `2e8326d531` (was 24 commits behind).

### The red checks here are base-red

`Merge integrity (changelog + generated skills)` fails on the **unmodified base**, not on this branch. Running the generator on a detached checkout of `upstream/release/v3.8.51` with none of my commits applied:

```
  GENERATED:
    + omni-inference
```

Byte-identical to what CI reports on this PR. `skills/` and `scripts/skills/` are untouched by this branch and unchanged in the 24 commits it was rebased over — the drift is in the catalog the generator reads. `Docs Gates (fast-path)` and `No new ESLint warnings` fail the same way on #11718 and #11716.

### Re-verified after the rebase

```
Adobe Firefly + image/video generation   179 tests, 178 pass, 1 fail
  the single failure is pre-existing: tests/unit/firefly-cookie-validation-10522.test.ts
tsc -p tsconfig.typecheck-core.json      9 errors, 0 mentioning adobeFirefly (9 pre-existing, omniglyph)
check:file-size                          OK
prettier --check                         clean on all ten files
```

MumuTW:
`Fast Quality Gates` is base-red too. It reports two failing gates:

```
::error::2 gate(s) failed: mutation-test-coverage check:dashboard-typecheck
```

Neither is reachable from this branch, and both reproduce on a detached checkout of `upstream/release/v3.8.51` with none of my commits applied:

**`check:mutation-test-coverage`** — byte-identical output with and without this branch:

```
  src/shared/utils/circuitBreaker.ts
    + tests/unit/combo-rr-diagnostics-11462.test.ts
✗ 1 covering unit test(s) across 1 module(s) are missing from stryker.conf.json tap.testFiles
```

**`check:dashboard-typecheck`** — the two live diagnostics are in `src/app/(dashboard)/dashboard/providers/[id]/__tests__/useApiKeySaveSkipsFullSync.test.tsx` (TS2493) and `src/shared/components/lobeProviderIcons.ts` (TS2322). This PR touches only `open-sse/services/adobeFirefly*.ts`.

Full base-red set on this PR — `Docs Gates`, `No new ESLint warnings`, `Merge integrity`, `Fast Quality Gates`, one unit shard — matches what #11718 and #11716 are showing against the same base.

PR #11710, opened 2026-08-26
fix(models): hide unsupported Gemini Live discovery

RaviTharuma:
## Summary

- stop mapping `bidiGenerateContent` to OmniRoute's batch audio endpoint
- omit Gemini Live-only models from dynamic Gemini and Vertex discovery
- preserve chat discovery for models that support both `generateContent` and the unsupported Live method

## Related Issues

- Follow-up to #10590 and #10591

## Validation

- [x] Focused tests and category gates from the golden path
- [x] `node --import tsx/esm --test tests/unit/gemini-models-parser.test.ts` (6/6)
- [x] `npm run typecheck:core`
- [x] focused ESLint
- [x] file-size and changelog-integrity gates
- [x] `git diff --check`

## Tests Added Or Updated

- Updated `tests/unit/gemini-models-parser.test.ts` to prove Live-only models are omitted while hybrid chat/Live models remain available as chat.

## Scope

This PR makes discovery honest. It does not implement Gemini Live's bidirectional protocol or OpenAI Realtime compatibility.

PR #11734, opened 2026-08-27
fix(ci): surface and clear release-green ESLint failure

MumuTW:
## Summary

Clear the ESLint hard failure currently reported by release-green on pristine `release/v3.8.51` tip `2e8326d53` (base-red #11449).

- move the hook-test result assignment out of React render and into an effect, clearing the hidden `react-hooks/globals` error
- give the cold full-tree ESLint walk a 60-minute ceiling instead of the 30-minute ceiling it crossed on the release runner
- enable the existing `.eslintcache` path for repeat runs
- preserve process failures such as code-124 timeout diagnostics instead of feeding them into the JSON parser and reporting the misleading `could not parse eslint json`

## Root cause

Run `33024734511` spent the full old 30-minute ESLint ceiling, so `classifyRunError()` returned the correct timeout diagnostic. The lint block ignored the non-zero process result, tried to parse that diagnostic as JSON, and replaced the actionable cause with `could not parse eslint json`.

A completed full-tree lint on the same base then exposed the real violation hidden behind the timeout:

```
src/app/(dashboard)/dashboard/providers/[id]/__tests__/useApiKeySaveSkipsFullSync.test.tsx:26:5
react-hooks/globals — hookResult was reassigned during render
```

The test now publishes the hook result from `useEffect`, matching the established hook-test pattern in this repository.

## Verification

- full ESLint inventory: **9,774 files, 0 errors, 0 warnings**
- release-green `--quick --hermetic`: ESLint hard gate **0 errors**, warning ratchet **0 vs baseline 0**
- `node --import tsx/esm --test tests/unit/validate-release-green.test.ts`: **31/31 pass**
- focused Vitest hook suite: **2/2 pass**
- scoped ESLint on changed test files: **0 errors, 0 warnings**
- `git diff --check`: pass

The separate provider-reference 356→357 docs red remains outside this focused PR and is already covered by #11710.

### DIFF
--- a/scripts/quality/validate-release-green.mjs
+++ b/scripts/quality/validate-release-green.mjs
@@ -60,6 +60,7 @@ import { parse as parseYaml } from "yaml";
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ROOT = join(__dirname, "..", "..");
 const npmCmd = process.platform === "win32" ? "npm.cmd" : "npm";
+export const ESLINT_TIMEOUT_MS = 60 * 60 * 1000;
 
 // Per-gate captured output. execFileSync buffers everything and the report only
 // shows a one-line summary, so without these files every red requires RE-RUNNING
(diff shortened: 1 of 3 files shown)


diegosouzapw:
Thanks for this! The release tip moved (a third test case was added to the same file) since
you opened this, and this fork rejects maintainer pushes, so a direct rebase-and-merge wasn't
possible. Recreated onto the current tip (author preserved via merge, same conflict
resolution re-applied) and merged as #11890 — thank you for tracking down the
release-green ESLint failure.

PR #11740, opened 2026-08-27
fix(docker): let the bun digest artifact be absent, not fatal

diegosouzapw:
Follow-up to #11724 — that PR stopped one step short.

#11724 made the bun image non-blocking and taught the manifest step to skip its tags when no digest exists. But the **upload** still carried `if-no-files-found: error`, so an absent digest — now the *expected* outcome of a skipped bun build — failed the job anyway.

Run [33030348950](https://github.com/diegosouzapw/OmniRoute/actions/runs/33030348950) shows it exactly: both arches died at **`Upload bun-base digests`**, after the decoupling had already done its part. The blocker had simply **moved** from the manifest to the upload.

## Changes

- bun digest **uploads**: `if-no-files-found: ignore`
- bun digest **downloads**: `continue-on-error`, since the artifact may not exist at all

`base` / `web` keep `error` on both sides — a supported image producing no digest is still a real failure that must stop the publish.

## Note for review

This is the second time in this cycle that a fix relocated a failure instead of removing it (the first: the npm publish moved from `Checkout` → `Validate npm package artifact` once disk was freed). Worth checking the whole bun path end to end rather than the one step the log names.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -275,7 +275,11 @@ jobs:
         with:
           name: digests-bun-base-${{ matrix.arch }}
           path: /tmp/digests/bun-base/*
-          if-no-files-found: error
+          # `ignore`, not `error`: the bun build is non-blocking, so an absent
+          # digest is the expected outcome of a failed/skipped bun image — the
+          # manifest step already treats these tags as optional. Leaving `error`
+          # here just relocates the blocker from the manifest to the upload.
+          if-no-files-found: ignore
           retention-days: 1
 
       - name: Upload bun-web digests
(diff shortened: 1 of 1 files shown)


RaviTharuma (approved):
Reviewed against failed publish run `33030348950` and the full digest path. The failure is exactly at `Upload bun-base digests` after base/web builds and their digest uploads succeeded. This change makes only bun upload/download artifacts optional; base/web still use `if-no-files-found: error` and fatal downloads, and manifest creation still requires both architecture digest files for every emitted tag. This removes the relocated best-effort Bun blocker without masking supported-image failures.

PR #11749, opened 2026-08-27
fix(lint): pin react hooks plugin to 7.0.1

MumuTW:
## Summary

- pin eslint-plugin-react-hooks 7.0.1 as a direct devDependency
- prevent eslint-config-next transitive range drift from changing lint policy
- add a regression test that locks the expected version and package-lock alignment

## Root cause

eslint-config-next allows eslint-plugin-react-hooks through ^7.0.0. A reused or refreshed node_modules could therefore load 7.1.1 while the repository lockfile and source remained unchanged. The 7.1 analyzer changes then surfaced hundreds of React Compiler diagnostics across unrelated dashboard files.

## Verification

- focused regression test: pass
- npm run check:lockfile: pass
- npm run lint: pass
- npm ls eslint-plugin-react-hooks: one deduped 7.0.1 installation
- pre-commit formatting, ESLint, docs sync, any-budget, and tracked-artifacts gates: pass
- npm run test:all: unit matrix reached 34,427 tests; 34,395 passed, 6 failed, 7 cancelled, 19 skipped. The command stopped before Vitest/E2E. These failures are outside this manifest-only change; the focused regression remains green.

## Follow-up

Upgrade to eslint-plugin-react-hooks 7.1.1 separately as an explicit lint-policy migration rather than accepting it through transitive dependency drift.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -124,6 +124,7 @@
         "dpdm": "^4.3.0",
         "eslint": "^9.39.4",
         "eslint-config-next": "16.3.1",
+        "eslint-plugin-react-hooks": "7.0.1",
         "eslint-plugin-sonarjs": "^4.1.0",
         "fast-check": "^4.8.0",
         "fumadocs-mdx": "^15.2.3",
--- a/package.json
+++ b/package.json
@@ -378,6 +378,7 @@
     "dpdm": "^4.3.0",
     "eslint": "^9.39.4",
     "eslint-config-next": "16.3.1",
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks for this fix! Its base (release/v3.8.50) had gone stale relative to the active
release/v3.8.51, and this fork rejects maintainer pushes, so a direct rebase-and-merge wasn't
possible. Recreated your commit (author preserved via cherry-pick) onto the current tip and
merged it as #11886 — thank you for catching and fixing the eslint-plugin-react-hooks drift.

PR #11779, opened 2026-08-27
fix(pwa): stop serving the stale shell after deploys

HouMinXi:
## Problem

After a deploy, browsers holding the old service worker can be locked out of the dashboard: the page opens and immediately breaks, and plain reloads do not recover it. Only manual cleanup (unregister worker + clear site data) fixes it. Observed in the wild: dashboard fine in incognito, broken in a normal window, on the same machine.

Two defects compound:

1. **navigationFallback replays the stale shell.** Navigations are network-first, but on network failure the worker falls back to a cached navigation response. That shell references `/_next/static/<old-build>/` chunks that no longer exist after the deploy, so the page boots and dies on chunk 404s. The broken state sticks because the worker intercepts every reload.

2. **The worker URL never changes.** `register("/sw.js")` from the old deployment never sees an update until a navigation to the new build fully succeeds -- the exact thing the old worker is breaking.

## Fix

- `navigationFallback` now returns `Response.error()`: an online tool surfaces network failures honestly instead of replaying a shell whose chunk references are dead. The browser's own offline handling recovers on the next reload.
- The worker is registered as `/sw.js?v=<NEXT_PUBLIC_SW_BUILD_ID>`. Each build stamps a different id (env override `OMNIROUTE_SW_BUILD_ID`, then `SOURCE_VERSION`, then build timestamp), so a deployed change makes every browser pick up the new worker generation on its next visit; the activate handler already drops caches from other generations.

## Verification

- `node --check` on sw.js passes; worker script parses.
- `tsc -p tsconfig.typecheck-dashboard.json`: 229 errors before and after -- all pre-existing (electronAPI globals etc.), none in the changed file.
- No SW unit-test harness exists in the repo (no workbox, hand-rolled sw.js), so the change is verified by parse + typecheck + the mechanism above.

## Risk

Low. The only behavior removed is offline replay of navigation responses -- a feature that actively harmed after every deploy. Static asset caching (network-first for _next assets, cache-first for icons) is unchanged.

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -126,6 +126,15 @@ const nextConfig = {
   // examples. Empty by default (root deploys unchanged).
   env: {
     NEXT_PUBLIC_OMNIROUTE_BASE_PATH: normalizeBasePath(process.env.OMNIROUTE_BASE_PATH),
+    // Deployment identity for the PWA service worker URL (PwaRegister):
+    // a browser holding a worker from an older deployment must see a
+    // different /sw.js?v=<id> URL so the browser treats it as an update
+    // instead of keeping the old generation in control. Falls back to a
+    // value that is unique per build run when git is absent (CI tarball).
+    NEXT_PUBLIC_SW_BUILD_ID:
+      process.env.OMNIROUTE_SW_BUILD_ID ||
+      process.env.SOURCE_VERSION ||
+      `${Date.now()}`,
   },
   distDir,
   // Turbopack config: redirect native modules to stubs at build time
--- a/public/sw.js
+++ b/public/sw.js
@@ -115,8 +115,17 @@ self.addEventListener("fetch", (event) => {
   );
 });
 
-async function navigationFallback(request) {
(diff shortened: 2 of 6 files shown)


HouMinXi:
R1 review findings addressed in cfd7b15b7:

- **F2 (ineffective query-string bump)**: the build now stamps the build id into the sw.js bytes themselves (`scripts/build/assembleStandalone.mjs` → `CACHE_NAME = "omniroute-pwa-v2-<id>"`). Each deploy produces different worker bytes AND a different cache name; the activate handler's delete-other-caches pass drops every stale generation.
- **F1 (3 stale tests)**: rewritten to the new contract. Navigation failure asserts `Response.error` — verified by bug-injection (restoring the old cache-replay flips both tests red). PwaRegister asserts `/sw.js?v=<id>` via stringMatching.
- **F3**: `/offline` removed from APP_SHELL (unreachable dead weight under the new navigation semantics).
- **F4**: removed the dead `extractBuildId` per-entry comparison (its `/_next/static/` pattern never matched a worker URL; CACHE_NAME-per-build supersedes it).

Issue #2460, opened 2026-05-21
[BUG] omniroute startup fail


AnuruddhaWijesiri:
### OmniRoute Version

3.8.0

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

OS Name: Microsoft Windows 11 Pro,OS Version: 10.0.26200 N/A Build 26200

### Node.js Version

v24.14.0

### Description

check the SS I attached, I waited almost 30minit but server not started

### Steps to Reproduce

Install omniroute using NPM
run `omniroute` and wait 

### Expected Behavior

need to start server

### Actual Behavior

Command hangs, server not starting

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

<img width="617" height="494" alt="Image" src="https://github.com/user-attachments/assets/af5e54e9-a5c0-42ee-bf75-7869e890d37b" />

dima2306:
Same for the latest 3.8.1 version...

netstratego:
same here for 3.8.1, I forgot to say I'm on MacOs. Can I use same command 

set APP_LOG_LEVEL=debug

Thanks

waseem852:
Got the same issue,. 3.8.1

diegosouzapw:
Thanks for the report. The screenshot alone doesn't show why startup hangs, so I need a bit more to pin it down — a 30-minute hang on Windows is most often a native-module (better-sqlite3) build/load stall during first run, not the HTTP server itself.

Could you share:

1. **Full terminal output** with debug logging — stop the process, then run:
   ```powershell
   set APP_LOG_LEVEL=debug
   omniroute
   ```
   and paste the text output (not a screenshot) up to where it hangs.
2. **Whether `~/.omniroute/` gets created** — does the data directory appear, and is there a `storage.sqlite` in it?
3. **Antivirus/Defender** — Windows Defender sometimes blocks the native module's first load; temporarily allow-listing the install dir is worth testing.

As a quick thing to try: v3.8.0 added a multi-driver SQLite layer that falls back to `node:sqlite` / WASM when the native `better-sqlite3` binary can't load. If you're on an older build, please upgrade and retry. If you're already on the latest and it still hangs, the debug log above will tell us exactly which step stalls.

dima2306:
[...]

PR #11794, opened 2026-08-27
fix(cli): probe both IPv4 and IPv6 loopback for server readiness

### LINKED ISSUE #11766: [BUG] CLI wrapper prints false "Server did not respond within 60s" warning despite server being healthy (v3.8.49)
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

_No response_

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

The CLI wrapper's startup probe (60s) reports failure with "**Server did not respond within 60s**" even though the Next.js server has finished booting, all background subsystems are initialized, and the health endpoint is returning **200 OK**

### Steps to Reproduce

1. Run **omniroute** on a fresh start
2. Wait for the boot log to reach ✓ Ready in 0ms and the [STARTUP] ... subsystem lines
3. Observe the wrapper prints the false-alarm warning instead of a clean ready message

### Expected Behavior

Either no warning, or a clear "**ready**" message once the server actually accepts traffic

### The wrapper prints
⚠ Server did not respond within 60s. It may still be starting, or may have failed silently
  Try:  curl -I http://localhost:20128/api/monitoring/health
  Or:   rerun with --log to see live server output

But curl.exe -I http://localhost:20128/api/monitoring/health returns 200 OK with full headers immediately

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

Image

### Additional Context

Likely cause: the wrapper's probe hits a route or uses a check that takes longer than 60s on first cold start (e.g. waiting for a specific subsystem beyond the listen socket being open, or polling with a too-long backoff)
Suggestion: probe should treat "listen socket accepting connections + /api/monitoring/health returning 200" as ready, and consider increasing the default timeout (or making it configurable via flag/env var)

### Validation Plan

After the fix, the --- Recent server output --- block should match the actual server state, and no "Server did not respond" warning should appear on a normal successful boot

vermasomesh835:
## Problem
The CLI wrapper was printing a false "Server did not respond within 60s" timeout warning despite the server being healthy and responding correctly. This occurred when the server listened exclusively on IPv6 loopback (`::1`), because the readiness probe only checked IPv4 loopback (`127.0.0.1`).

## Solution
Modified the readiness probe to check both IPv4 and IPv6 loopback addresses **concurrently**, accepting either family returning 200 OK as ready. This maintains all existing behavior for IPv4-only setups while enabling IPv6-only servers to be properly detected.

### Key Changes
- **`pollHealthOnce()`**: Now probes both `127.0.0.1` and `::1` concurrently using `Promise.all()`, aggregating outcomes intelligently:
  - If either family responds with 200, server is ready
  - If any family shows TCP is listening but rejecting (fast-reject), respect the 3-second grace window for "not-yet-mounted route" (#2460)
  - If any family shows TCP accepting but never responding, treat as still-booting
  - Only return "not-listening" if both families fail to connect
  
- **`isPortListening()`**: Checks both IPv4 and IPv6, returns true if either is listening

- **New regression test**: Validates IPv6-only server detection by creating an HTTP server on `::1`, confirming it's detected as ready

## Testing
All 6 tests pass:
- Timeout behavior on closed port (#2460) ✓
- TCP fallback when route is not mounted (#2460) ✓
- Safeguard: TCP-accept-but-never-respond stays NOT ready (#6800) ✓
- **NEW**: IPv6-only loopback server detection (#11766) ✓
- TCP-accept-never-respond comprehensive test ✓
- TCP fallback grace period recovery ✓

## Backward Compatibility
- IPv4-only setups: unchanged behavior
- IPv6-only setups: now work (was broken)
- Dual-stack setups: both families checked, whichever responds first determines outcome
- No timeout changes, no grace window changes, all existing diagnostics preserved

Fixes: #11766

### DIFF
--- a/bin/cli/utils/pid.mjs
+++ b/bin/cli/utils/pid.mjs
@@ -100,31 +100,66 @@ export async function waitForServer(port, timeout = 60000) {
 // - "hanging": the request timed out waiting for any response — the
 //   process accepted the TCP connection but never answered (#6800).
 // - "not-listening": nothing is accepting connections on the port at all.
+// #11766: probe both IPv4 and IPv6 loopback to handle servers listening on
+// either family (or both).
 async function pollHealthOnce(port) {
-  try {
-    const res = await fetch(`http://127.0.0.1:${port}/api/monitoring/health`, {
-      signal: AbortSignal.timeout(2000),
-    });
-    return res.ok ? "ready" : "fast-reject";
-  } catch (err) {
-    if (err?.name === "TimeoutError") return "hanging";
-    const listening = await isPortListening(port).catch(() => false);
-    return listening ? "fast-reject" : "not-listening";
-  }
+  const hosts = ["127.0.0.1", "::1"];
+  const outcomes = [];
+
+  // Probe both loopback families concurrently
+  const results = await Promise.all(
+    hosts.map(async (host) => {
+      try {
+        const res = await fetch(`http://${host}:${port}/api/monitoring/health`, {
+          signal: AbortSignal.timeout(2000),
+        });
+        return { host, outcome: res.ok ? "ready" : "fast-reject" };
+      } catch (err) {
+        const outcome = err?.name === "TimeoutError" ? "hanging" : "error";
+        return { host, outcome };
+      }
+    })
+  );
+
+  outcomes.push(...results.map((r) => r.outcome));
+
+  // If either family is ready, the server is ready
+  if (outcomes.includes("ready")) return "ready";
+
+  // If either family is fast-reject, treat as fast-reject
+  // (TCP is listening and rejecting, just route not ready yet)
+  if (outcomes.includes("fast-reject")) return "fast-reject";
+
+  // If either family is hanging, server accepted TCP but not answering
+  // (still booting, must not report as ready per #6800)
+  if (outcomes.includes("hanging")) return "hanging";
+
+  // Both families failed — check if either port is actually listening
+  // If listening, then errors above are route-level (fast-reject case)
+  const listening = await isPortListening(port).catch(() => false);
+  return listening ? "fast-reject" : "not-listening";
 }
 
 async function isPortListening(port) {
   const net = await import("node:net");
-  return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


xiaoyaner0201:
## Post-merge verification: the IPv6 branch still violates the #6800 invariant

I re-ran the production-boundary probe on the **exact merged commit** `b07eaafcc4bc86f783462a4dcc04320fded4858a` (not on the older PR comparison tree). The final source is unchanged in the relevant path, and the two-case probe returns **0 pass / 2 fail / 0 skip** on Node's real test runner:

- IPv6-only socket accepts TCP but never sends HTTP: expected `waitForServer(...)=false`; actual `true` after **3072 ms**.
- IPv6-only fixture would return health 200 if requested: `waitForServer(...)=true` after **3041 ms**, but the fixture received **0 request bytes**. Readiness came from the 3-second TCP grace, not from the health route.

The cause is the raw IPv6 literal in ``fetch(`http://${host}:${port}/...`)``. For `host === "::1"`, the URL is unbracketed, so `fetch` rejects before connecting. Both health probes then appear as generic errors; `isPortListening()` sees the IPv6 socket, returns true, and the generic fallback turns that into `fast-reject`, which becomes ready after the grace window. This is why the positive IPv6 test passes while never exercising IPv6 HTTP, and why the IPv6 accept-never-respond case reverses the safeguard documented for #6800.

This is narrower than the original report: as noted in the #11807 checkpoint, the reporter's v3.8.49 build used `localhost`, so dual-loopback selection was not established as the cause of their warning. I am not treating the original symptom as proven fixed.

I am reconciling the already-Draft #11807 against the merged release rather than opening another competing PR. The repair gate is: bracket the IPv6 URL, preserve per-family outcome precedence, add a test that proves the IPv6 HTTP endpoint was actually reached, and keep an IPv6 hanging socket not-ready. The Draft will remain Draft; no force-push or Ready transition.

Issue #10508, opened 2026-08-15
[BUG] Launching Omniroute Fails in terminal while actually launches in the background


CyberBonk:
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

win 11 25H2

### Node.js Version

v24.16.0

### Provider(s) Involved

N/A

### Model(s) Involved

N/A

### Client Tool

OpenCode, Hermes

### Description

i keep receiving this failing status in the terminal:
to access omniroute i'd go to Opencode and tell it to gimme the Local host link that launched Omniroute and it searches for it while the Launching command silently fails like this

```
PS C:\Users\bebo> omniroute
  📋 Loaded env from C:\Users\bebo\.omniroute\.env

   ____                  _ ____              _
   / __ \                (_) __ \            | |
  | |  | |_ __ ___  _ __ _| |__) |___  _   _| |_ ___
  | |  | | '_ ` _ \| '_ \ |  _  // _ \| | | | __/ _ \
  | |__| | | | | | | | | | | | \ \ (_) | |_| | ||  __/
   \____/|_| |_| |_|_| |_|_|_|  \_\___/ \__,_|\__\___|

  v3.8.49

  ⏳ Starting server...


⚠ Server did not respond within 60s. It may still be starting, or may have failed silently.
  Try:  curl -I http://localhost:20128/api/monitoring/health
  Or:   rerun with --log to see live server output.

--- Recent server output ---
▲ Next.js 16.2.12
- Local:         http://localhost:20128
- Network:       http://0.0.0.0:20128
✓ Ready in 0ms
[DB] Changing cache_size from 65536KB to 16384KB
[DB] cache_size changed to 16384KB
[DB] SQLite database ready: C:\Users\bebo\.omniroute\storage.sqlite (DATA_DIR=C:\Users\bebo\.omniroute, SQLITE_FILE=C:\Users\bebo\.omniroute\storage.sqlite)
[ReasoningCache] boot cleanup removed 190 expired entries
[CREDENTIALS] No external credentials file found, using defaults.
[STARTUP] Global fetch proxy patch initialized
[STARTUP] JWT_SECRET restored from persistent store
[HealthCheck] Starting proactive token health-check (tick every 60s)
[Shutdown] Graceful shutdown handlers registered.
[STARTUP] Spend batch writer started
[STARTUP] Guardrail registry initialized
[STARTUP] Builtin skill handlers registered
[...]

PR #11807, opened 2026-08-27
fix(cli): probe both loopback families for CLI readiness (related to #11766)

xiaoyaner0201:
## Summary

The CLI readiness probe in `bin/cli/utils/pid.mjs` polls only the literal `127.0.0.1`, so a server
listening on IPv6 loopback (`::1`) is reported as not-ready even while its health route answers
`200`. This PR probes both literal loopback families concurrently on every poll and aggregates the
per-family outcomes through an explicit precedence ladder.

**Scope is `related to #11766`, not a fix for it.** See "Scope" below — this is deliberate, and the
issue is intentionally left open.

### Change

- `waitForServer()` probes `LOOPBACK_PROBE_HOSTS = ["127.0.0.1", "::1"]` concurrently per poll.
- New exported `classifyReadiness(outcomes)` aggregates with a fixed precedence:
  `ready > hanging > fast-reject > not-listening`.
- `pollHealthOnce(port, host)` brackets IPv6 literals into `http://[::1]:PORT/…`, since an
  unbracketed `http://::1:PORT/…` is not a valid URL.
- `isPortListening(port, host)` is family-matched, so the TCP fallback can no longer be satisfied by
  a *different* family than the one that failed.

`hanging` deliberately outranks `fast-reject`: it is what preserves #6800.

### What is actually repaired

An **unreleased regression on the latest tip**, introduced by #10758 (`e5a81ed744`, which hardcoded
`127.0.0.1`). That commit is *not* an ancestor of v3.8.48, v3.8.49 or v3.8.50. The defect is
reachable through three of the four values `serverHost.mjs::LOOPBACK_HOSTS` already enumerates.

### #6800 is preserved

A socket that accepts TCP but never answers must never be reported ready. Verified on the exact
published tree: `ready=false` at 10031 ms and 10017 ms. A mutant with the precedence ladder swapped
returns `ready=true` at 7035 ms, so the guarantee is genuinely load-bearing and not incidental.

Unchanged: `waitForServer(port, timeout = 60000)`, the **shared** 3 s route-mount grace window, all
existing diagnostics. No timeout inflation, no new environment variable, no DNS name in the probe
path.

## Related Issues

- Related to #11766
- Regression origin: #10758
- Guarantee preserved: #6800
- Grace-window behaviour: #2460

Deliberately **no** auto-closing keyword — see Scope.

## Scope — why this does not claim to resolve the reported symptom

The reporter is on **v3.8.49**. That build ships the `localhost` form, which *does* reach an
`::1`-only listener. Dual-loopback coverage therefore **cannot** explain their symptom, and closing
#11766 on the strength of this PR would hide a still-unexplained bug.

What this PR fixes is real and independently reproducible, but it is a *different* defect on the
unreleased tip that happens to live in the same function. The reporter's mechanism is still being
pursued on the issue.

## Validation

- [x] Change type: **CLI**
- [ ] `npm run lint` — see Limitations

Base reconciliation: `upstream/release/v3.8.51` is at `2acbfc6fa6ef13e55552ce5bc3de45cbf2e535be`,
**0 commits** ahead of this branch's base.
`git merge-tree --write-tree HEAD upstream/release/v3.8.51` returns `c106b32f7012557144b3fee380714835db67f01b`
— byte-identical to this head's own tree, i.e. a pure fast-forward with no merge content and no
`changelog.d` filename collision.

All gates re-run on the exact published tree `c106b32f7012557144b3fee380714835db67f01b`
(node v22.22.0, darwin):

| gate | result |
|---|---|
| `tests/unit/cli-classifyReadiness-precedence-11766.test.mjs` | 4 pass / 0 fail / 0 skipped, exit 0 |
| `tests/unit/cli-waitForServer-dual-loopback-11766.test.mjs` | 6 pass / 0 fail / **0 skipped**, exit 0 |
| `tests/unit/cli-waitForServer.test.mjs` | 3 pass, exit 0 |
| `tests/unit/waitForServer-tcp-fallback-6800.test.mjs` | 2 pass, exit 0 |
| `tests/unit/cli-readiness-127-0-0-1-10508.test.ts` | 1 pass, exit 0 |
| `npm run check:changelog-integrity` | exit 0 |
| `npx prettier --check` (3.9.6, all five changed paths) | exit 0 |

## Tests Added Or Updated

- **added** `tests/unit/cli-waitForServer-dual-loopback-11766.test.mjs` — 6 end-to-end cases over
  real loopback sockets: IPv6-only 200, IPv4-only 200, IPv6-only fast-reject keeping the 3 s grace,
  IPv4 accept-never-respond staying not-ready, neither family listening, and a 200 on one family
  outranking a hanging sibling on the other.
- **added** `tests/unit/cli-classifyReadiness-precedence-11766.test.mjs` — 4 table-driven cases
  pinning the precedence ladder with hard-coded literal expectations, so a co-mutation that reorders
  production *and* the test's own table together still fails.
- **updated** `tests/unit/cli-readiness-127-0-0-1-10508.test.ts` — extended for the family-matched
  `isPortListening` signature.

## Coverage Notes

`bin/cli/utils/pid.mjs` is the only production file touched. Every added branch is covered:
the IPv6 probe path and bracketing by T1/T3/T6, the IPv4 path by T2, family-matched
`isPortListening` by T4 and the updated #10508 test, and all four precedence rungs by the table
test. Coverage does not move down in any touched file.

### Limitations, stated plainly

- All measurements are **darwin / node v22.22.0**. Windows behaviour is inferred, not measured.
- The dual-loopback suite **silently degrades to 2/6 passing with exit 0 on a runner without `::1`**
  — the IPv6 cases skip themselves. **CI lacking IPv6 loopback loses T1/T3/T6 coverage without
  failing**, so a reviewer reading only the exit code would be misled. Reading the skip count is the
  only honest check.
- This worktree has no `node_modules`, so only the bare-`node` gates above were run locally;
  `npm run lint` and the `.ts` suites requiring `npm ci` were not exercised here and are left to CI.
- `scripts/check/check-forgotten-sibling-tests.mjs` exits 1 with `ERR_MODULE_NOT_FOUND: tinyglobby`
  **identically on a pristine archive of the untouched base**, so it is environmental and not caused
  by this change.

### Overlap with #11794 — flagged for maintainer decision

PR #11794 targets the same file and base for the same symptom and was opened while this work was in
review. I am flagging it rather than ignoring it, and I have no stake in which lands.

One measured difference is worth a maintainer's attention. #11794 builds its probe URL as
`` `http://${host}:${port}/…` `` with `host = "::1"`. Unbracketed, that is an invalid URL — under
node v22.22.0 `new URL("http://::1:8080/…")` throws `TypeError`, not `TimeoutError`. Because its
`catch` maps only `TimeoutError` to `hanging`, its IPv6 health probe can never succeed *and* never
reports hanging; it silently falls through to the TCP `isPortListening` fallback.

Measured on a pristine archive of this PR's tree with only `pid.mjs` swapped for #11794's head
`f2b5f2b4f299ecfad756a336e47916e87b6d23ba`, IPv6-only listener, deterministic across 3 runs:

| scenario | this PR | #11794 |
|---|---|---|
| health 200 on `::1` | `ready=true` @11 ms — real health 200 | `ready=true` @3015 ms — only via the 3 s TCP grace |
| **accepts TCP, never answers (#6800)** | `ready=false` @10017 ms — safe | `ready=true` @3015 ms — **false ready** |
| `{fast-reject, hanging}` pair | `ready=false` @10052 ms — safe | `ready=true` @3054 ms — **false ready** |

The positive IPv6 case passes there for the wrong reason: the grace-window fallback, not the health
probe. That is why its own suite stays green (4/4, plus #6800 2/2) while the guarantee is reversed.

Reproduce:

```bash
git clone <fork> && git checkout 4a6876188112f9597f9e78bed8d122f1819c7d03
mkdir -p /tmp/cmp && git archive HEAD | tar -x -C /tmp/cmp
REF=f2b5f2b4f299ecfad756a336e47916e87b6d23ba
gh api "repos/vermasomesh835/OmniRoute/contents/bin/cli/utils/pid.mjs?ref=$REF" \
  --jq .content | base64 -d > /tmp/cmp/bin/cli/utils/pid.mjs
cd /tmp/cmp && node --test tests/unit/cli-waitForServer-dual-loopback-11766.test.mjs
```

T6 fails with `false !== true`, and the precedence suite fails to import
(`no export named 'classifyReadiness'`).

If you prefer #11794's shape, the minimum fixes are bracketing the IPv6 literal, classifying
non-timeout errors distinctly from timeouts, and family-matching `isPortListening`. The two test
files here apply to either implementation and are offered on that basis.

### Draft status

Opened as **Draft** on purpose. The overlap with #11794 is a maintainer's call, and I would rather
that be settled before asking for review time. Happy to close this in favour of #11794 with the
tests contributed there instead — just say which you prefer.


diegosouzapw:
Obrigado pela investigação cuidadosa aqui — o precedence ladder (`ready > hanging > fast-reject > not-listening`) e a prova com mutante preservando #6800 são um trabalho sólido! Ao processar esta PR contra a tip atual de `release/v3.8.51`, encontrei que **#11794** foi mesclada hoje (2026-08-30) por outra sessão, corrigindo a issue #11766 diretamente (`bin/cli/utils/pid.mjs` + `tests/unit/cli-waitForServer.test.mjs`) com uma implementação funcionalmente equivalente: `pollHealthOnce()` já sonda `127.0.0.1` e `::1` concorrentemente e aplica a mesma precedência ready → fast-reject → hanging → not-listening, preservando a garantia do #6800 (socket que aceita TCP mas nunca responde nunca é reportado como pronto).

Como o mesmo problema já está coberto na tip por implementação independente, vou fechar esta PR. Obrigado novamente pelo cuidado com a regressão e pela prova formal de que a garantia do #6800 é genuinamente load-bearing.

xiaoyaner0201:
Thank you for the careful review. One important factual correction before this is treated as a duplicate of #11794:

I re-ran the merged implementation after this closure against the exact current `release/v3.8.51` tip, `7f49b342b5c743264706dd746efe14068ee6fa8a` (tree `e6dbb5e7d78f10a2d79b8a72e0f86f0b73687445`). Its `bin/cli/utils/pid.mjs` is byte-identical to merge commit `b07eaafc` (`sha256 153858b9c4dff042e364ae28ad4a6fbdb742135c8b5047b8c822e5880cdbffaa`).

On a complete archive of that exact tip, adding only the narrow reviewed regression test blob (`sha256 7bd77dd2606bcc95ccbba73c5dfeb7625df924bb56c8f58d2a93f4c840b0b7a1`) and running:

```bash
node --test tests/unit/cli-waitForServer-ipv6-endpoint-reached-11807.test.mjs
```

still gives **0 pass / 2 fail / 0 skipped**:

- an IPv6 socket that accepts TCP and never responds is reported ready after about 3.05 seconds, reversing the #6800 guarantee;
- an IPv6-only HTTP 200 health fixture receives **zero HTTP requests**, while readiness is nevertheless reported after the TCP grace.

The key difference is that the merged code constructs `http://::1:PORT/...` rather than a bracketed IPv6 URL. `fetch` throws `TypeError` before any request reaches the server; both health probes therefore land in the generic error path, and the family-agnostic TCP fallback upgrades the live IPv6 socket to `fast-reject`. The current implementation also checks `fast-reject` before `hanging`, so it is not functionally equivalent to the reviewed `ready > hanging > fast-reject > not-listening` contract.

The earlier exact merge-commit evidence is here: https://github.com/diegosouzapw/OmniRoute/pull/11794#issuecomment-5468902316

I have **not** pushed the locally reviewed descendant, reopened this PR, changed Draft state, or retargeted anything. Could you please re-evaluate whether you prefer to request a refresh of this existing branch against the latest release, or handle the bounded IPv6 URL/family-precedence residual in a maintainer-side patch? I will not open a replacement PR unless explicitly requested.

PR #11808, opened 2026-08-27
fix(kiro): do not permanently ban on 'User is not authorized to make this call'

Deftera186:
Kiro IDC accounts that were created before #10725 (or with the eu-west-1 -> us-east-1 region mismatch) hit AWS CodeWhisperer with no profileArn or with the wrong Q Developer region. AWS answers 403 'User is not authorized to make this call' — a recoverable configuration issue, not an account ban. The account still works in Kiro IDE, and after the profileArn is discovered (ListAvailableProfiles -> profileArn in us-east-1) the same token succeeds.

Previously this 403 fell through to FORBIDDEN which the chatCore handler maps to a permanent 'banned' state (is_active=0, testStatus=banned). That required a full OAuth reconnect to recover, even though the token was healthy.

Classify the Kiro-specific 403 as PROJECT_ROUTE_ERROR so the connection stays active and can be retried after profile discovery, matching the existing treatment for Cloud Code 403s (antigravity-403). The OAuthModal fix in #10725 already prevents new logins from losing the profileArn; this guards the runtime path for existing connections and any future region-mismatch edge cases.

**Reproduction**

- Add a Kiro IDC account whose IdC lives in eu-west-1 (or any non-profile region) on v3.8.49 before the _authMethod fix — the stored connection has region=eu-west-1, authMethod=builder-id, no profileArn.
- Any chat request to kiro/* routes to codewhisperer.us-east-1 without profileArn and gets 403, then the account is marked banned permanently. Direct Kiro IDE still works.
- After this fix the 403 keeps the account active; after manually discovering the profileArn (or re-authenticating on a fixed build) chat succeeds (verified with live CodeWhisperer ListAvailableProfiles -> arn:aws:codewhisperer:us-east-1:...).

**Verification**

- Verified live: ListAvailableProfiles with the eu-west-1 token returns the us-east-1 profile ARN; after writing it to provider_specific_data the same token's generateAssistantResponse succeeds for kiro/glm-5 and kiro/gpt-5.6-sol.
- The new classification keeps the account in PROJECT_ROUTE_ERROR (connection stays active) instead of FORBIDDEN.

Fixes the 'banned every time I authenticate with Kiro AI' report where direct Kiro chat works but OmniRoute bans the account. Follow-up to #10725.


Deftera186:
Closing in favor of #11809 which is rebased onto main (single commit, no PII) — previous branch was based on a feature merge and showed 3k commits.

Issue #9536, opened 2026-08-05
fix(backend): /v1/messages usage misreported on OpenAI-shaped upstreams (regression of #6738)


AutomAIton:
Regression of #6738, plus an additional symptom that issue did not cover.

On the Anthropic-compatible `/v1/messages` path with an **OpenAI-shaped upstream** (DeepSeek), the
usage OmniRoute reports to the client does not match the usage OmniRoute itself received from the
provider. Cache fields are zeroed **and** `input_tokens` is inflated by roughly 2,000 tokens.

Billing is unaffected — the provider charges correctly. Only the client-facing numbers are wrong.
The damage is to anyone using those numbers: cost dashboards, budget guards, and benchmarks all
read a cache-miss where a cache hit occurred.

**Version:** `omniroute@3.8.48`, started with `HOSTNAME=127.0.0.1 omniroute serve --no-open`.

### Evidence

Identical payload sent twice through `/v1/messages`, `model: deepseek/deepseek-v4-flash`, with a
`tools` array, ~4,300-token stable prefix so the second call should hit DeepSeek's automatic cache.

What OmniRoute returned to the client:

| Call | `input_tokens` | `cache_read_input_tokens` |
|---|---|---|
| 1 | 6,364 | 0 |
| 2 | 6,364 | 0 |

What OmniRoute recorded from DeepSeek for those same two calls, in its own
`~/.omniroute/call_logs/<date>/*.json` under `responseBody.usage`:

```json
{
  "prompt_tokens": 4364,
  "prompt_tokens_details": { "cached_tokens": 4352 },
  "prompt_cache_hit_tokens": 4352,
  "prompt_cache_miss_tokens": 12,
  "completion_tokens": 27
}
```

The same log carries `_omniroute.claudePromptCacheUsage = { cacheReadTokens: 4352, cacheCreationTokens: 0 }`,
so the correct value is present internally and simply is not surfaced on the response.

### The two defects

1. **`cache_read_input_tokens` is always 0** on this path, even though the upstream reported 4,352
   cached tokens and OmniRoute stored that number. This is the #6738 symptom, now reproducing
   against an OpenAI-shaped upstream rather than Anthropic.
2. **`input_tokens` is inflated by ~2,000** (6,364 reported vs 4,364 upstream). This appears to be
   new relative to #6738. The offset held constant across request shapes: 6,091 vs 4,091 without
[...]

PR #11833, opened 2026-08-27
fix(backend): capture usage on trailing choices:[] streaming chunks

### LINKED ISSUE #11817: [BUG] Streaming OpenAI→Claude translator drops usage (incl. cache tokens) on trailing choices:[] chunk
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 26.04

### Node.js Version

_No response_

### Provider(s) Involved

Fireworks (custom OpenAI-compatible connection)

### Model(s) Involved

kimi-k3

### Client Tool

Claude Code (also reproducible with cURL against /v1/messages)

### Description

`openaiToClaudeResponse` in `open-sse/translator/response/openai-to-claude.ts` discards the
  final SSE chunk — and its `usage` object — whenever that chunk carries `"choices": []`. Many
  OpenAI-compatible upstreams (confirmed: Fireworks, for accounts/fireworks/models/kimi-k3) send
  the real usage block in exactly that shape: an empty-choices trailing chunk. Streaming
  Claude-format clients (any Anthropic-protocol client — Claude Code, Claude Desktop, etc.) never
  receive real prompt-cache accounting for these backends; OmniRoute silently falls back to its own
  token estimate instead — no cache split, no error, no visible signal anything is wrong.

  Companion bug to #9536, which fixed the same field mapping on the non-streaming /v1/messages
  path (landed in 3.8.50). This one is still present on the streaming path in 3.8.50 / current
  main.

  Impact: any per-token billing done downstream of OmniRoute's usage block (e.g. a proxy that
  meters and charges customers off it) bills 100% of input tokens at the uncached rate even when
  the upstream provider actually served most of the request from prompt cache at a fraction of the
  cost. For a provider priced at $3/M uncached vs $0.30/M cached, a ~75%-cache-hit session bills
  ~3x what it should — silently.

### Steps to Reproduce

1. Configure a custom OpenAI-compatible provider connection whose backend reports prompt-cache
     accounting via `usage.prompt_tokens_details.cached_tokens` on a trailing `choices: []` SSE
     chunk (confirmed: Fireworks, model accounts/fireworks/models/kimi-k3).
  2. Call OmniRoute's raw OpenAI-shaped endpoint to see the real data:
     POST /v1/chat/completions, stream: true, a system prompt large enough to trigger caching
     (~6k+ tokens), send the same prompt twice in a row.
     -> the final SSE chunk looks like:
        data: {"choices":[],"usage":{"prompt_tokens":6103,"prompt_tokens_details":{"cached_tokens":6102},...}}
  3. Call the same request against the Claude-shaped endpoint:
     POST /v1/messages, stream: true, identical prompt.
     -> inspect the `message_delta` event's `usage` field.

### Expected Behavior

The /v1/messages message_delta usage should carry the real numbers, including
  cache_read_input_tokens / cache_creation_input_tokens derived from
  usage.prompt_tokens_details.cached_tokens / cache_creation_tokens on the upstream chunk — the
  same mapping that already works correctly on the non-streaming path (#9536).

### Actual Behavior

message_delta usage carries OmniRoute's own token estimate instead of the real upstream numbers,
  with no cache fields at all:

  data: {"type":"message_delta","delta":{"stop_reason":"max_tokens"},"usage":{"input_tokens":7908,"output_tokens":24,"estimated":true}}

  Note input_tokens (7908) also disagrees with the real prompt_tokens (6103) from step 2 — it's a
  tokenizer estimate, not upstream-reported.
  Root cause: open-sse/translator/response/openai-to-claude.ts, function
  openaiToClaudeResponse(chunk, state):

    export function openaiToClaudeResponse(chunk, state) {
      if (!chunk || !chunk.choices?.[0]) return null;   // <- returns before the usage
                                                          //    block below ever runs
      ...
      if (chunk.usage && typeof chunk.usage === "object") {
        ...
        state.usage = { input_tokens: inputTokens, output_tokens: outputTokens, ... };
        ...
      }

  The `choices[0]` guard returns null for exactly the chunk shape carrying real usage
  (choices: [], usage: {...}), so state.usage is never populated from real data. Whatever later
  builds the final message_delta (open-sse/utils/stream.ts, presumably) falls back to its own
  estimate and marks it "estimated": true.

  Suggested fix — hoist the existing chunk.usage block above the choices[0] guard so it still
  runs on an empty-choices chunk:

    export function openaiToClaudeResponse(chunk, state) {
      if (!chunk) return null;

      const results = [];

      // Must run even when this chunk carries no choices — several OpenAI-compatible
      // upstreams (Fireworks confirmed) send real usage on a trailing choices:[] chunk.
      if (chunk.usage && typeof chunk.usage === "object") {
        const promptTokens = typeof chunk.usage.prompt_tokens === "number" ? chunk.usage.prompt_tokens : 0;
        const outputTokens = typeof chunk.usage.completion_tokens === "number" ? chunk.usage.completion_tokens : 0;
        const cachedTokens = chunk.usage.prompt_tokens_details?.cached_tokens;
        const cacheCreationTokens = chunk.usage.prompt_tokens_details?.cache_creation_tokens;
        const cacheReadTokens = typeof cachedTokens === "number" ? cachedTokens : 0;
        const cacheCreateTokens = typeof cacheCreationTokens === "number" ? cacheCreationTokens : 0;
        const inputTokens = promptTokens - cacheReadTokens - cacheCreateTokens;

        state.usage = { input_tokens: inputTokens, output_tokens: outputTokens };
        if (cacheReadTokens > 0) state.usage.cache_read_input_tokens = cacheReadTokens;
        if (cacheCreateTokens > 0) state.usage.cache_creation_input_tokens = cacheCreateTokens;
      }

      if (!chunk.choices?.[0]) return results.length ? results : null;

      const choice = chunk.choices[0];
      const delta = choice.delta;
      ...

### Test Impact

### Error Logs / Output

```shell
=== chat/completions STREAM (raw OpenAI-shape, real Fireworks usage) ===
  data: {"id":"chatcmpl-3979f1f3d6cc4f4e8a136dd319531953","object":"chat.completion.chunk","created":1787850922,"model":"accounts/fireworks/models/kimi-k3","choices":[],"usage":{"prompt_tokens":6103,"total_tokens":6119,"completion_tokens":16,"prompt_tokens_details":{"cached_tokens":6102},"completion_tokens_details":{"reasoning_tokens":16},"output_toke
  ns_details":{"reasoning_tokens":16}}}

  === /v1/messages STREAM (Claude-shape, same request) ===
  data: {"type":"message_delta","delta":{"stop_reason":"max_tokens"},"usage":{"input_tokens":7908,"output_tokens":24,"estimated":true}}
```

### Screenshots

_No response_

### Additional Context

  Image tested: ghcr.io/diegosouzapw/omniroute:3.8.50
  (digest sha256:085c57adf499a8aaa9f35ccde95c0df9c11bd9ecd18d6c9edbf3b68b8079ba9d)

  Companion to #9536 (fixed the non-streaming /v1/messages path in 3.8.50) — this is the
  streaming-path equivalent of that same bug, still open.

### Validation Plan

  - Add a unit test for openaiToClaudeResponse() feeding a chunk shaped
    {"choices": [], "usage": {"prompt_tokens": N, "prompt_tokens_details": {"cached_tokens": M}}}
    and asserting state.usage.cache_read_input_tokens === M (mirrors the non-streaming test added
    for #9536).
  - Manual: repeat the reproduction steps above against a build with the fix; the /v1/messages
    message_delta usage should show real input_tokens matching /v1/chat/completions' prompt_tokens,
    cache_read_input_tokens present, and no "estimated": true.

  ---

Saraswat123:
## Summary

Closes #11817.

`openaiToClaudeResponse()` (`open-sse/translator/response/openai-to-claude.ts`) early-returned on `!chunk.choices?.[0]` **before** the usage-capture block ran. Upstreams using `stream_options.include_usage` (confirmed by the reporter: Fireworks, `accounts/fireworks/models/kimi-k3`) send the real usage object — including `prompt_tokens_details` cache accounting — on a trailing chunk shaped `{choices: [], usage: {...}}`. That shape was silently discarded, so `state.usage` was left on whatever value the finish_reason chunk had set (often no cache split, or an estimate).

Fix moves the usage-capture block above the `choices?.[0]` guard so it runs regardless of whether the chunk has a choice to process. The function still returns `null` for that trailing chunk (nothing to translate into a Claude event), matching the existing `null`-chunk test — only `state.usage` changes.

This mirrors the non-streaming fix for the companion bug (#9536, landed in 3.8.50 via #9657) — same field-mapping logic, same root shape of bug, streaming path instead of non-streaming.

## Known follow-up (flagging, not fixing here)

`message_delta`/`message_stop` are emitted as soon as `choice.finish_reason` is seen (a few lines below, guarded by `state.claudeFinishEmitted`). Per the OpenAI `include_usage` contract, the trailing usage-only chunk this PR fixes typically arrives *after* the finish_reason chunk — meaning by the time the real usage lands, `message_stop` may already have been sent to the client with the old estimate. This PR guarantees `state.usage` is correct for any consumer that reads it after the stream (e.g. logging/billing), which is what the issue's own validation plan asks for and mirrors #9536's scope — but if the SSE `message_delta.usage` sent *to the client* also needs to reflect this, that requires buffering the terminal event by one chunk, which is a bigger change I didn't want to bundle in here unreviewed. Happy to follow up if maintainers want that handled too.

## Test plan

- Added `tests/unit/11817-streaming-usage-trailing-empty-choices.test.ts`: trailing `choices: []` chunk with real usage now populates `state.usage` (cache_read_input_tokens, input_tokens net of cache) exactly as in the issue's repro; a trailing chunk with no usage remains a safe no-op.
- Ran the full existing test suite touching this function (translator-resp-openai-to-claude, tool-call/XML/casing/shim tests, #9536 non-streaming companion test): 93/93 pass, no regressions.
- Pre-commit hooks (prettier, eslint, t11 any-budget, docs-sync) all pass.

### DIFF
--- a/changelog.d/fixes/11817-streaming-usage-trailing-empty-choices.md
+++ b/changelog.d/fixes/11817-streaming-usage-trailing-empty-choices.md
@@ -1,0 +1,1 @@
+- fix(backend): capture usage on trailing `choices: []` streaming chunks in the OpenAI→Claude translator, so cache token accounting from upstreams using `stream_options.include_usage` (e.g. Fireworks) is no longer silently dropped (#11817)
--- a/docs/architecture/meta.json
+++ b/docs/architecture/meta.json
@@ -13,6 +13,7 @@
     "admission-lanes",
     "cluster-decisions",
     "persistence-backend-boundary",
-    "ADAPTIVE_ROUTING"
+    "ADAPTIVE_ROUTING",
+    "streaming-usage-trailing-chunk-decision"
   ]
 }
--- a/docs/architecture/streaming-usage-trailing-chunk-decision.md
+++ b/docs/architecture/streaming-usage-trailing-chunk-decision.md
@@ -1,0 +1,42 @@
+---
+title: "Streaming Usage: Trailing Empty-Choices Chunk"
+version: 3.8.51
+lastUpdated: 2026-08-29
+---
+
+# Decision Record — Capturing Usage on Trailing `choices: []` Streaming Chunks
+
+**Status:** proposed (PR #11833, awaiting @diegosouzapw review)
+**Date:** 2026-08-29
+**Refs:** [#11817](https://github.com/diegosouzapw/OmniRoute/issues/11817), companion to [#9536](https://github.com/diegosouzapw/OmniRoute/issues/9536) / PR #9657
+
+## TL;DR
+
+`openaiToClaudeResponse()` in [`open-sse/translator/response/openai-to-claude.ts`](../../open-sse/translator/response/openai-to-claude.ts) early-returned before capturing `usage` whenever a chunk had no `choices[0]`. Upstreams using OpenAI's `stream_options.include_usage` contract (confirmed: Fireworks) send the authoritative usage block — including prompt-cache accounting — on exactly that shape: a trailing chunk with `choices: []`. The fix reorders the function so usage capture runs before the choices guard, at the cost of doing that check on every chunk instead of only ones with a choice.
+
+## Context
+
+Two ways OpenAI-compatible upstreams deliver usage in a stream:
+
+1. **Usage attached to the finish_reason chunk itself.** Single chunk, `choices: [{finish_reason: "stop", ...}], usage: {...}`. This already worked.
+2. **Usage on a separate trailing chunk, per the `include_usage` spec.** The finish_reason chunk arrives with `usage: null`, then one more chunk follows with `choices: []` and the real `usage` object. This is what Fireworks does, and it's spec-compliant OpenAI behavior, not a Fireworks quirk — so any upstream following that spec exactly hits the same bug.
+
+The existing code only handled case 1. Case 2's trailing chunk was dropped by the `!chunk.choices?.[0]` guard before ever reaching the usage-extraction logic, so `state.usage` kept whatever the finish_reason chunk had set (no cache split, or an estimate downstream falls back to).
+
+## Decision
+
+Move the usage-capture block (lines ~194-215) to run unconditionally on any non-null chunk, before the `choices?.[0]` guard. The guard itself is unchanged for everything after it — content/tool/finish handling still requires a real choice and returns `null` when there isn't one.
+
+### Alternatives considered
+
+- **Special-case `choices: []` explicitly** (`if (chunk.choices?.length === 0 && chunk.usage) { ...capture...; return null; }`) instead of reordering. Rejected: duplicates the extraction logic in two places, or requires extracting it into a shared helper for a one-call-site function — more surface area than reordering three lines for the same effect.
+- **Delay `message_delta`/`message_stop` until stream end**, so the terminal SSE event always has final usage regardless of which chunk it arrives on. This is the more complete fix (see Known Limitation below) but changes the finish-emission path shared by every provider going through this translator — bigger blast radius than this PR's scope. Left as a follow-up, not bundled here.
+
+## Known limitation (not fixed by this PR)
+
+`message_delta`/`message_stop` fire as soon as `choice.finish_reason` is seen (guarded by `state.claudeFinishEmitted`), which — per the ordering in case 2 above — is typically _before_ the trailing usage chunk arrives. This PR guarantees `state.usage` is correct for anything reading it **after** the stream completes (billing/logging writes, admin dashboards). It does **not** guarantee the live SSE `message_delta.usage` sent to the client reflects the corrected numbers, since that event may already be on the wire by the time the real usage lands. Confirming and closing that gap (if it matters for OmniRoute's actual billing pipeline) is the follow-up described above.
+
+## Validation
+
+- `tests/unit/11817-streaming-usage-trailing-empty-choices.test.ts` — reproduces the issue's exact Fireworks payload shape and asserts `state.usage` (input/output/cache_read) matches provider-reported numbers; asserts a usage-less trailing chunk is still a safe no-op.
+- Full existing suite touching this function (tool calls, XML `<invoke>` parsing, tool-name casing, finish-dedup, the non-streaming #9536 companion test): 93/93 pass, no regressions.
--- a/open-sse/translator/response/openai-to-claude.ts
+++ b/open-sse/translator/response/openai-to-claude.ts
@@ -46,10 +46,22 @@ function extractXmlInvokeBlocks(
     const toolCallTextMatch = remaining.match(/TOOL_CALL\s+([A-Za-z0-9_]+):\s*/);
 
     const matches = [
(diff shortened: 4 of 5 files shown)


Saraswat123:
@diegosouzapw the fork-PR CI workflows (Quality Gates, DAST smoke) are sitting at "action required" and need a maintainer approval to run. Semgrep already passed. Ready whenever you get a chance to approve.

Saraswat123 (commented):
## Self-review

Going through this with the same scrutiny I'd want applied to someone else's PR before it lands.

**Correctness**
- Confirmed the defect exists at `open-sse/translator/response/openai-to-claude.ts:167` (pre-fix): `!chunk.choices?.[0]` returned `null` before the usage-capture block ever ran, so any chunk shaped `{choices: [], usage: {...}}` was a total no-op on `state.usage`.
- The reorder is minimal — usage capture now runs unconditionally on any non-null chunk, the `choices?.[0]` guard is otherwise untouched. No behavior change for the common case (usage on the finish_reason chunk).
- Verified against the reporter's exact repro shape (Fireworks/kimi-k3 payload) in the new test — `input_tokens`, `cache_read_input_tokens`, `output_tokens` all compute correctly from a trailing `choices: []` chunk.

**Test coverage**
- New test covers: (1) trailing chunk with real usage updates `state.usage` correctly, (2) trailing chunk with no usage is a safe no-op (guards against regressing the existing null-return contract).
- Ran the full existing suite touching this function — 93/93 pass. That includes the tool-call/XML-invoke/casing/finish-dedup tests and the #9536 non-streaming companion test, so no regression on the paths this function is shared with.

**Scope discipline**
- Did *not* bundle the bigger fix (delaying `message_delta`/`message_stop` until stream end so the live SSE event always reflects final usage). That's a real gap — see the "Known limitation" section in the ADR — but it changes shared finish-emission logic used by every provider through this translator, which is a materially bigger and riskier change than this issue asked for. Flagging it rather than scope-creeping this PR felt like the right call; happy to take it as a separate follow-up if @diegosouzapw wants it.

**Open question for reviewer**
- I don't know where OmniRoute's actual billing/dashboard numbers are sourced from (state.usage post-stream vs. the already-emitted SSE event). That determines whether this PR fully closes the billing-impact described in #11817, or only the state-correctness half of it. Would appreciate a pointer to that code if it's not obvious from the translator layer alone — I can extend this PR or file the follow-up once I know.

No blocking concerns from me on what's actually in the diff. CI is still gated on maintainer approval (Quality Gates / DAST smoke workflows) as of this comment.

Saraswat123:
Superseded — #11883 (merged) fixes the same defect in openaiToClaudeResponse(), and #11933 additionally covers the finish-emission ordering gap I flagged in this PR's "Known limitation" section. Closing to avoid a stale duplicate. Thanks for the quick turnaround on this one.

Issue #11819, opened 2026-08-27
fix(providers): opencode-go models (e.g. glm-5.3-flash) — reasoning_effort "max" silently rewritten to "xhigh", upstream 400 [1210]


pixie-chan:
## Summary

On omniroute 3.8.49 (latest on npm as of 2026-08-27), requests to `opencode-go/glm-5.3-flash` with `reasoning_effort: "max"` fail with upstream `400 [1210]`. The identical request sent directly to the opencode-go endpoint (`https://opencode.ai/zen/go/v1`) succeeds, proving the upstream accepts `"max"`. Tracing shows OmniRoute rewrites `"max"` to `"xhigh"` before forwarding, and this upstream rejects `"xhigh"` (and `"medium"`) with:

```
Error from provider (Console Go): Upstream request failed: [1210] This model always engages in thinking and cannot be disabled; please use low, high, or max
```

The error text even tells the client to "use low, high, or max", while rejecting the `"max"` the client actually sent, because it never reaches upstream as `"max"`.

## Verified behavior matrix (curl, 2026-08-27)

| reasoning_effort | direct zen/go/v1 | via omniroute 3.8.49 |
|------------------|------------------|----------------------|
| (unset)          | 200              | 200                  |
| low              | 200              | 200                  |
| high             | 200              | 200                  |
| max              | 200              | 400 [1210]           |
| medium           | 400 [1210]       | 400 [1210]           |
| xhigh            | 400 [1210]       | 400 [1210]           |
| thinking disabled| 400 [1210]       | 400 [1210]           |

## Root cause

In the dist server chunks, `normalizeResponsesReasoningEffort` rewrites `"max"` to `"xhigh"` unless the model matches `gpt-5.6-sol/terra/luna`:

```js
function(e,t){let n,s=r(e).toLowerCase();return"max"!==s?s:(n=r(t).trim().toLowerCase().replace(/^(?:codex|cx)\//,""),o.test(n))?"max":"xhigh"}
```

`glm-5.3-flash` is not a gpt-5.6 model, so `"max"` becomes `"xhigh"`. But the opencode-go upstream ("Console Go") natively supports the full `low/high/max` ladder and rejects `xhigh`/`medium` with `[1210]`.

## Related but not duplicates

- #10562 preserved max only for `deepseek-v4-flash` specifically. `glm-5.3-flash` on the same provider still gets rewritten on 3.8.49.
[...]

PR #11857, opened 2026-08-28
fix(skills): expand shorthand property types in injected tool schemas

### LINKED ISSUE #11856: fix(backend): malformed injected skill-tool schema (shorthand string props not expanded) breaks strict-schema providers
## Summary

On omniroute 3.8.51, every chat completion to `opencode-go/glm-5.3-flash` fails with:

```
[400]: Error from provider (Console Go): Upstream request failed: [1210] Invalid API parameter, please check the documentation.
```

This reproduces with a **minimal request body** (`{"messages":[{"role":"user","content":"say ok"}]}`), while the identical body posted directly to `https://opencode.ai/zen/go/v1/chat/completions` returns 200 — so it is not the client's parameters.

## Root cause

OmniRoute injects builtin skill tools (`omr_skill_*` from `src/lib/skills/injection.ts`) into chat requests. Three builtin skills (`generation@1.0.0`, `frontend-design@1.0.0`, `brainstorming@1.0.0`) declare their input schema in **shorthand**:

```json
{"content": "string"}
```

`normalizeInputSchema()` wraps a bare property map as `{type: "object", properties: input}` **without expanding** the shorthand string values, so the forwarded tool schema is:

```json
{"type": "object", "properties": {"content": "string"}}
```

That is invalid JSON Schema (`"string"` is not a schema object). Zhipu GLM behind the Console Go tier validates tool schemas strictly and rejects the entire request with `[1210]`. Most other providers tolerate the malformed schema, which is why this only surfaced on this provider.

Verified by capturing the exact outbound request body (live capture of the upstream POST): removing the three `omr_skill_*` tools makes the request succeed; re-adding a single tool with the shorthand schema reproduces `[1210]` byte-for-byte.

## Impact

- 100% failure rate for any model on providers with strict schema validation (opencode-go / Console Go GLM tier), regardless of request content or credentials.
- Same `[1210]` error signature as #11819 and #11274, but a different root cause — those cover `reasoning_effort` rewriting; this one fires even with no `reasoning_effort` at all.

## Fix

Expand string property values during normalization in `normalizeInputSchema` (`src/lib/skills/injection.ts`):

```ts
const properties: Record<string, unknown> = {};
for (const [key, value] of Object.entries(input)) {
  properties[key] = typeof value === "string" ? { type: value } : value;
}
return { type: "object", properties };
```

PR with fix + regression test incoming.

## Environment

omniroute 3.8.51 (npm/bun global, Node v26.3.1, macOS arm64), provider opencode-go (`https://opencode.ai/zen/go/v1`), model glm-5.3-flash.

alvinveroy:
Fixes #11856. Related: #11819, #11274 (same `[1210]` error signature from Console Go, different root causes).

## Problem

Every chat completion to `opencode-go/glm-5.3-flash` fails with:

```
[400]: Error from provider (Console Go): Upstream request failed: [1210] Invalid API parameter, please check the documentation.
```

It reproduces with a **minimal request body** while the identical body posted directly to `https://opencode.ai/zen/go/v1/chat/completions` returns 200 — so the client's parameters are not the cause.

## Root cause

OmniRoute injects builtin skill tools (`omr_skill_*`) into chat requests. Three builtin skills (`generation@1.0.0`, `frontend-design@1.0.0`, `brainstorming@1.0.0`) declare input schemas in **shorthand**:

```json
{"content": "string"}
```

`normalizeInputSchema()` wraps a bare property map as `{type: "object", properties: input}` **without expanding** the string values, so the forwarded tool schema is invalid JSON Schema:

```json
{"type": "object", "properties": {"content": "string"}}
```

Zhipu GLM behind the Console Go tier validates tool schemas strictly and rejects the entire request with `[1210]`. Most other providers tolerate the malformed schema, which is why this only surfaced on this provider and looked like a provider outage.

Verified by live capture of the exact outbound POST: removing the three `omr_skill_*` tools makes the request succeed; re-adding a single tool with the shorthand schema reproduces `[1210]` byte-for-byte.

## Fix

`normalizeInputSchema` now expands string property values into `{ type: value }` when wrapping a bare property map. Object-shaped schemas (the documented storage format, e.g. `{ query: { type: "string" } }`) pass through untouched, and the existing `input.type === "string"` early-return is preserved.

## Test plan

- `tests/unit/skills-injection.test.ts`:
  - New regression test `injectSkills expands shorthand property types into valid JSON Schemas` — registers a shorthand-schema skill and asserts every provider shape (openai/anthropic/google/other) emits schema **objects** for each property.
  - Existing assertions updated from the malformed `{ query: "string" }` to the corrected `{ query: { type: "string" } }`.
  - Red/green verified: 3 failures with the fix reverted, 9/9 pass with it.
- `npx eslint` on both changed files: clean.

## Commands run

```bash
tests:      DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=8192 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test tests/unit/skills-injection.test.ts
lint:       npx eslint src/lib/skills/injection.ts tests/unit/skills-injection.test.ts
coverage:   npm run test:coverage
```

Coverage gate result: recorded in the run below (threshold 60% statements/lines/functions/branches per repo config).


alvinveroy:
## Coverage gate + test report

**Coverage (`npm run test:coverage`, c8, thresholds 60/60/60/60):**

| Metric | Result | Gate |
|---|---|---|
| Statements | 76.31% (627832/822663) | ✅ |
| Lines | 76.31% | ✅ |
| Branches | 79.55% (110164/138481) | ✅ |
| Functions | 69.96% (22244/31792) | ✅ |

**Tests:** 34,973 total — 34,973 pass, 4 fail, 20 skipped.

**Changed test file:** `tests/unit/skills-injection.test.ts`
- New regression test `injectSkills expands shorthand property types into valid JSON Schemas` (asserts every provider shape emits schema objects for shorthand-declared skills).
- Existing assertions updated from the malformed `{ query: "string" }` to the valid `{ query: { type: "string" } }`.
- Targeted run: **9/9 pass**; red/green verified — 3 failures with the fix reverted (`git stash push -- src/lib/skills/injection.ts`), 9/9 with it.

**The 4 failing tests are pre-existing and unrelated to this change** (none of the failing files import `src/lib/skills/injection.ts`):
1. `relay-completions-errors.test.ts` — 502 vs 404 normalization (environment/timeout-dependent).
2. `binaryManager.test.ts` — macOS `/private/var` symlink prefix vs `/var` expected path (macOS tmpdir realpath issue).
3. `call-log-file-rotation.test.ts` — orphan-scan timing assertion.
4. `mcp-stdio-json-purity.test.ts` — `better-sqlite3` dynamic-require failure in this Node 26.3.1 environment.

All four reproduce on a clean tree without this patch (same class as the base-red noted in #11274).

**Commands run**
```bash
# targeted suite (9/9)
DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=8192 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test tests/unit/skills-injection.test.ts
# lint (clean)
npx eslint src/lib/skills/injection.ts tests/unit/skills-injection.test.ts
# coverage gate
npm run test:coverage
```

PR #11882, opened 2026-08-28
fix(api): map persisted combo steps in the playground route simulator

### LINKED ISSUE #11822: fix(playground): simulate-route 500s for every persisted combo — reads combo.targets but getCombos() returns models
## Summary

`POST /api/playground/simulate-route` always returns **HTTP 500** for any persisted combo:

```json
{"error":"Simulation error: Cannot read properties of undefined (reading 'length')"}
```

The Combo Playground page (`/dashboard/combos/playground`) is therefore unusable — every "Simulate Route" click fails, regardless of which combo is selected.

## Environment

- Version: **v3.8.50**
- Deployment: Docker
- Reproduced on two independent instances (macOS/arm64 and Linux/amd64)

## Reproduction

1. Create any combo, e.g.:

```json
{
  "name": "h-route-claude-opus-5",
  "strategy": "priority",
  "models": [
    { "kind": "model", "model": "cc/claude-opus-5", "fallbackOnlyOnQuotaExhaustion": true },
    { "kind": "model", "model": "anthropic/claude-opus-5" }
  ],
  "config": { "compressionMode": "off" }
}
```

2. Simulate it by id:

```bash
curl -X POST http://localhost:20128/api/playground/simulate-route \
  -H "<management auth>" \
  -d '{"comboId":"<combo-id>","promptTokens":500}'
```

**Expected:** 200 with the simulated routing path.
**Actual:** 500 `Simulation error: Cannot read properties of undefined (reading 'length')`.

Note this is *not* the "combo not found" path — a bogus `comboId` correctly returns 404, so the combo is being found and then crashes during target iteration.

## Root cause

`src/app/api/playground/simulate-route/route.ts` reads a `targets` field that persisted combos do not have:

```ts
const combos = (await getCombos()) as any[];
const combo = combos.find((c) => c.id === body.comboId || c.name === body.comboId);
// ...
const targets = typeof combo.targets === "string" ? JSON.parse(combo.targets) : combo.targets;
comboInfo = { name: combo.name, strategy: combo.strategy, targets };
```

`getCombos()` (`src/lib/db/repositories/sqliteComboRepository.ts`) returns records normalized by `normalizeComboRecord()`, whose steps live under **`models`** (`ComboStep[]`, schema version 2) — there is no `targets` property. So `targets` is `undefined`, and the loop below throws:

```ts
for (let i = 0; i < comboInfo.targets.length; i++) {   // <-- TypeError
```

The inline `body.combo` branch works because the caller supplies `targets` explicitly per `simulateRequestSchema`; only the `comboId` branch (the one the dashboard uses) is broken.

A secondary mismatch in the same loop: each step's provider comes from `step.providerId` / the `provider/model` prefix of `step.model`, not a top-level `t.provider`, so the connection lookup

```ts
const conn = connections.find((c) => c.id === t.provider || c.name === t.provider || ...);
```

would not resolve even once `targets` is populated, producing spurious `"Provider undefined not configured"` warnings.

## Suggested fix

Map the persisted `models: ComboStep[]` shape to the simulator's target shape, e.g.:

```ts
const steps = Array.isArray(combo.models) ? combo.models : [];
const targets = steps
  .filter((s) => s.kind === "model" || typeof s.model === "string")
  .map((s) => {
    const parsed = parseModel(s.model);
    return {
      provider: s.providerId ?? parsed.provider ?? "unknown",
      model: parsed.model ?? s.model,
      weight: s.weight,
    };
  });
```

`combo-ref` and `provider-wildcard` steps would need their own handling (expand via `resolveComboTargets`, or surface as a warning) — happy to open a PR if you'd like a particular approach.

## Impact

Routing itself is unaffected — real requests through these combos dispatch correctly (verified end-to-end, `priority` cascade with `fallbackOnlyOnQuotaExhaustion` behaving as documented). Only the pre-flight simulation UI is broken, so operators cannot preview a cascade before sending traffic.

NoxzRCW:
## Summary

`POST /api/playground/simulate-route` returned HTTP 500 for every persisted combo, which made the Combo Playground page unusable. Every "Simulate Route" click failed, whichever combo was selected:

```json
{"error":"Simulation error: Cannot read properties of undefined (reading 'length')"}
```

This is not the "combo not found" path. A bogus `comboId` correctly returns 404, so the combo is found and then the handler crashes while iterating its targets.

The route read `combo.targets`, but `getCombos()` returns records normalized by `normalizeComboRecord()`, which is schema version 2 and keeps its steps under `models` as a `ComboStep[]`. There is no `targets` property, so `comboInfo.targets` was always `undefined` and the loop below dereferenced `undefined.length`. The inline `body.combo` branch worked only because the caller supplies `targets` explicitly per `simulateRequestSchema`, which is why the dashboard was the only broken path.

There is a second mismatch in the same loop. A step's provider comes from `step.providerId`, or from the `provider/model` prefix of `step.model`, never from a top-level `step.provider`. So even with `targets` populated, the connection lookup would have failed to resolve and produced spurious `Provider undefined not configured` warnings. Both are handled here.

`combo-ref` and `provider-wildcard` steps are not expanded by the simulator. Rather than dropping them silently they now surface as warnings, and the wildcard is listed unresolved so the operator can see it is in the route.

## Related Issues

- Closes #11822

## Validation

- [x] Change type: routing
- [x] `npm run lint`

```
node --import tsx/esm --test tests/unit/playground-simulate-route-combo-steps-11822.test.ts
# tests 6 / pass 6 / fail 0

npm run typecheck:core   # 0 errors
npm run lint             # clean
```

## Tests Added Or Updated

- `tests/unit/playground-simulate-route-combo-steps-11822.test.ts` (new)

Covers the three step kinds, `providerId` winning over the model prefix, a serialized `models` column, and the malformed inputs (`null`, unparseable JSON, junk entries) that should degrade to an empty target list rather than throw.

## Coverage Notes

The mapping lives in `src/lib/combos/simulatorTargets.ts` and is fully covered by the new test file. The route itself keeps its existing shape and just calls into it.

## Reviewer Notes

I put the mapping in `src/lib/combos/` rather than inline in the route, for two reasons. A Next.js App Router route module cannot export anything besides its handlers and the documented route config, so the helper would not have been testable where the issue suggested putting it. And `ComboStep` already lives under `src/lib/combos/`, so the mapping sits next to the type it consumes.

Routing itself was never affected. Real requests through these combos always dispatched correctly, this was only the simulator's view of them.

If you would rather have `combo-ref` steps expanded through `resolveComboTargets()` instead of surfacing as a warning, say so and I will do that instead. I went with the warning because expanding nested combos in the simulator raises cycle-handling questions that felt out of scope for a 500 fix.

### DIFF
--- a/changelog.d/fixes/11882-simulate-route-combo-steps.md
+++ b/changelog.d/fixes/11882-simulate-route-combo-steps.md
@@ -1,0 +1,1 @@
+- **fix(api):** `POST /api/playground/simulate-route` maps persisted schema-v2 combo steps instead of reading a `targets` field that no longer exists, fixing the HTTP 500 on every combo in the Combo Playground ([#11882](https://github.com/diegosouzapw/OmniRoute/pull/11882)) — thanks @NoxzRCW
--- a/src/app/api/playground/simulate-route/route.ts
+++ b/src/app/api/playground/simulate-route/route.ts
@@ -8,6 +8,7 @@
 import { NextResponse } from "next/server";
 import { z } from "zod";
 import { getCombos } from "@/lib/db/combos";
+import { comboStepsToTargets } from "@/lib/combos/simulatorTargets";
 import { getProviderConnections } from "@/lib/db/providers";
 import { isValidationFailure, validateBody } from "@/shared/validation/helpers";
 
@@ -152,7 +153,7 @@ export async function POST(request: Request) {
         errors.push(`Combo "${body.comboId}" not found.`);
         return NextResponse.json({ error: "Combo not found" }, { status: 404 });
       }
-      const targets = typeof combo.targets === "string" ? JSON.parse(combo.targets) : combo.targets;
+      const targets = comboStepsToTargets(combo, warnings);
       comboInfo = { name: combo.name, strategy: combo.strategy, targets };
     } else if (body.combo) {
       comboInfo = body.combo;
--- a/src/lib/combos/simulatorTargets.ts
+++ b/src/lib/combos/simulatorTargets.ts
@@ -1,0 +1,84 @@
+/**
+ * @file simulatorTargets.ts
+ * @description Map persisted combo steps (schema v2) onto the flat target shape
+ * used by the Combo Playground simulator.
+ */
+
+export interface SimulatorTarget {
+  provider: string;
+  model: string;
+  weight?: number;
+}
+
+/**
+ * Persisted combos are stored in schema v2 (`normalizeComboRecord`), whose steps
+ * live under `models: ComboStep[]` — there is no top-level `targets` array. The
+ * simulator used to read `combo.targets`, which is always `undefined` for a
+ * persisted combo, so the target loop threw
+ * "Cannot read properties of undefined (reading 'length')" and every request
+ * returned HTTP 500 (#11822).
+ *
+ * Map the three step kinds onto the simulator's flat target shape. The provider
+ * comes from `step.providerId` or the `provider/model` prefix of `step.model` —
+ * never a top-level `step.provider`, which does not exist.
+ */
+export function comboStepsToTargets(
+  combo: Record<string, unknown>,
+  warnings: string[]
+): SimulatorTarget[] {
+  const raw = (combo as { models?: unknown }).models;
+  const steps: unknown[] = Array.isArray(raw)
+    ? raw
+    : typeof raw === "string"
+      ? (() => {
+          try {
+            const parsed: unknown = JSON.parse(raw);
+            return Array.isArray(parsed) ? parsed : [];
+          } catch {
+            return [];
+          }
+        })()
+      : [];
+
+  const targets: SimulatorTarget[] = [];
+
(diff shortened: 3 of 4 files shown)


diegosouzapw:
Thanks for this — and sorry for the overlap. #11862 landed first and already fixed the
core #11822 crash (reading `combo.models` instead of the nonexistent `combo.targets`,
resolving provider via `providerId`/model-prefix), so this conflicted by the time I got
to it.

Your one genuinely distinct contribution — surfacing `combo-ref` and `provider-wildcard`
steps with a specific warning instead of a generic "unsupported step" count, and showing
the wildcard as an unresolved target — was real value worth keeping. I hand-applied that
part onto #11862's code (full credit to your design) as #11930, since your extracted
`comboStepsToTargets()` helper didn't cover the plain-string step shape #11862 also
handles. Closing this one now that its distinct value has landed. Thank you for the
thorough regression coverage and the clear writeup either way!

PR #11915, opened 2026-08-28
fix(translator): preserve usage metrics from trailing choices-empty chunks (#11817)

### LINKED ISSUE #11817: [BUG] Streaming OpenAI→Claude translator drops usage (incl. cache tokens) on trailing choices:[] chunk

HouMinXi:
Fixes #11817. Preserves usage (including cache tokens) from trailing OpenAI chunks with empty choices array in the OpenAI→Claude stream translator. Includes TDD test coverage. ⚠️ base-red inherited: #11449

### DIFF
--- a/open-sse/translator/response/openai-to-claude.ts
+++ b/open-sse/translator/response/openai-to-claude.ts
@@ -164,22 +164,22 @@ function stopTextBlock(state, results) {
 
 // Convert OpenAI stream chunk to Claude format
 export function openaiToClaudeResponse(chunk, state) {
-  if (!chunk || !chunk.choices?.[0]) return null;
+  if (!chunk && !state.pendingClaudeFinishChoice) return null;
 
   const results = [];
-  const choice = chunk.choices[0];
-  const delta = choice.delta;
+  const chunkUsage = chunk?.usage;
+  const hasChunkUsage = chunkUsage && typeof chunkUsage === "object";
 
   // Track usage from OpenAI chunk if available
-  if (chunk.usage && typeof chunk.usage === "object") {
+  if (hasChunkUsage) {
     const promptTokens =
-      typeof chunk.usage.prompt_tokens === "number" ? chunk.usage.prompt_tokens : 0;
+      typeof chunkUsage.prompt_tokens === "number" ? chunkUsage.prompt_tokens : 0;
     const outputTokens =
-      typeof chunk.usage.completion_tokens === "number" ? chunk.usage.completion_tokens : 0;
+      typeof chunkUsage.completion_tokens === "number" ? chunkUsage.completion_tokens : 0;
 
     // Extract cache tokens from prompt_tokens_details
-    const cachedTokens = chunk.usage.prompt_tokens_details?.cached_tokens;
-    const cacheCreationTokens = chunk.usage.prompt_tokens_details?.cache_creation_tokens;
+    const cachedTokens = chunkUsage.prompt_tokens_details?.cached_tokens;
+    const cacheCreationTokens = chunkUsage.prompt_tokens_details?.cache_creation_tokens;
     const cacheReadTokens = typeof cachedTokens === "number" ? cachedTokens : 0;
     const cacheCreateTokens = typeof cacheCreationTokens === "number" ? cacheCreationTokens : 0;
 
@@ -205,6 +205,13 @@ export function openaiToClaudeResponse(chunk, state) {
     // Note: completion_tokens_details.reasoning_tokens is already included in output_tokens
     // No need to add separately as Claude expects total output_tokens
   }
+
+  const chunkChoice = chunk?.choices?.[0];
+  const flushingPendingFinish = !chunkChoice && Boolean(state.pendingClaudeFinishChoice);
+  const choice = chunkChoice || state.pendingClaudeFinishChoice;
+  if (!choice) return null;
+  if (flushingPendingFinish) state.pendingClaudeFinishChoice = null;
+  const delta = choice.delta;
 
   // First chunk - ALWAYS send message_start first
   if (!state.messageStartSent) {
@@ -436,6 +443,11 @@ export function openaiToClaudeResponse(chunk, state) {
   // guard therefore misfired and silently dropped the terminal message_delta/message_stop
   // for Responses→Claude streams (#5828 regression).
   if (choice.finish_reason && !state.claudeFinishEmitted) {
+    if (!hasChunkUsage && !flushingPendingFinish) {
+      state.pendingClaudeFinishChoice = choice;
+      return results.length > 0 ? results : null;
+    }
+
     state.claudeFinishEmitted = true;
     stopThinkingBlock(state, results);
     stopTextBlock(state, results);
--- a/tests/unit/translator/openai-to-claude-trailing-usage.test.ts
+++ b/tests/unit/translator/openai-to-claude-trailing-usage.test.ts
@@ -1,0 +1,229 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { openaiToClaudeResponse } from "../../../open-sse/translator/response/openai-to-claude.ts";
+
+type ClaudeUsage = {
+  input_tokens: number;
+  output_tokens: number;
+  cache_read_input_tokens?: number;
+  cache_creation_input_tokens?: number;
+};
+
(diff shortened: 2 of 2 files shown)


diegosouzapw:
Thanks for this — and genuinely good catch. #11883 landed first with a narrower fix
(harvesting usage from the trailing chunk into `state.usage`), which conflicted with your
branch by the time I got to it. But your PR caught something #11883 actually missed: I
verified directly against #11883's own merged code that when `finish_reason` arrives
before the trailing usage-only chunk (the normal order for these upstreams), the
client-visible `message_delta` still went out with `{"input_tokens":0,"output_tokens":0}`
— `state.usage` got corrected internally, but too late to matter for the event already
sent. Your `pendingClaudeFinishChoice` deferral is what actually closes that gap.

I hand-merged your deferral mechanism onto #11883's already-merged `trackUsageFromChunk()`
extraction (they touched the same lines) as #11933, with your test suite carried over
verbatim. 86/86 tests passing including the existing #5828/#8081/#8162 regression guards.
Full credit for the design and the regression coverage — thank you for tracking this all
the way to the actual emitted event instead of stopping at `state.usage`. Closing this one
now that it's landed.

PR #11917, opened 2026-08-28
fix(playground): resolve combo models and targets in simulate-route API (#11822)

### LINKED ISSUE #11822: fix(playground): simulate-route 500s for every persisted combo — reads combo.targets but getCombos() returns models

HouMinXi:
Fixes #11822. Resolves 500 error in simulate-route by correctly parsing combo.models and combo.targets from persisted combos. Includes TDD test coverage. ⚠️ base-red inherited: #11449

### DIFF
--- a/src/app/api/playground/simulate-route/route.ts
+++ b/src/app/api/playground/simulate-route/route.ts
@@ -7,6 +7,7 @@
 
 import { NextResponse } from "next/server";
 import { z } from "zod";
+import { getComboModelProvider, getComboModelString, getComboStepWeight } from "@/lib/combos/steps";
 import { getCombos } from "@/lib/db/combos";
 import { getProviderConnections } from "@/lib/db/providers";
 import { isValidationFailure, validateBody } from "@/shared/validation/helpers";
@@ -47,6 +48,12 @@ interface SimulateResponse {
   totalEstimatedLatencyMs: number;
   warnings: string[];
   errors: string[];
+}
+
+interface SimulationTarget {
+  provider: string;
+  model: string;
+  weight?: number;
 }
 
 const simulateRequestSchema = z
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks for this — the underlying crash (#11822, reading combo.targets on a persisted
schema-v2 combo that only has .models) was already fixed by #11862 by the time I got to
this one, and #11930 later ported the remaining warning-quality improvement from a
sibling PR (#11882).

One note in case it's useful elsewhere: your `normalizeStoredTargets()` reads through
`getComboModelString()`/`getComboModelProvider()` from `@/lib/combos/steps`, which return
`null` for both `combo-ref` and `provider-wildcard` step kinds (the latter has no `.model`
field, only `.modelPattern`) — so both step kinds get silently filtered out with no
warning at all in this implementation. That's actually a step backward from what's live
now (#11862 + #11930), which surfaces a specific warning for each and keeps
provider-wildcard as an unresolved target the operator can see. Worth keeping in mind if
those helpers get reused elsewhere for a similar simulator/preview surface.

Closing this one since #11822 is already resolved. Thanks for the clean helper-function
approach either way!

PR #11920, opened 2026-08-28
fix(ui): add null-safe guards for provider color property (#11853)

### LINKED ISSUE #11853: fix: providers page crashes reading 'color' on prototype-colliding provider id, shows misleading connection-error card
**OmniRoute Version:** 3.8.50 (Docker image; Next.js 16.3.2) — **also confirmed on 3.8.51**; the vulnerable code is unchanged on current `main`
**Operating System:** Linux
**Node.js Version:** N/A (Docker)
**Provider(s) Involved:** Any — UI-level, not provider-specific
**Client Tool:** Web dashboard (Chrome)

## Description

The Providers dashboard intermittently renders a red error card ("Failed to load providers — We could not load provider data right now. Check your connection and try again.") while the server is completely healthy and its logs show no request errors. The card is produced by the App Router error boundary at `src/app/(dashboard)/dashboard/providers/error.tsx`, which catches **any** uncaught client-side render/effect error under `/dashboard/providers/**` — so the copy blaming the connection is misleading: no network failure is required to produce it.

In production the boundary displays only the error digest (the message is gated behind `NODE_ENV === "development"`) and never logs the caught error, so the browser console is the only diagnostic surface.

Chrome console (minified production bundle):

```
TypeError: Cannot read properties of undefined (reading 'color')
    at 2idg479rnh8wh.js:1:604708
    at 2idg479rnh8wh.js:1:604729
    at aS (16i0z23sc3hp8.js:31:62492)
    at o1 (16i0z23sc3hp8.js:31:82785)
    ...
```

Property names survive minification, so the failing access is a `.color` read on an undefined object during render.

## Analysis (source audit at v3.8.50)

All five endpoints feeding the page (`/api/providers`, `/api/provider-nodes`, `/api/providers/expiration`, `/api/settings`, `/api/providers/openrouter-stats`) are fetched through `loadProviderPageData` in `providerPageUtils.ts`, whose `safeJson` wrapper swallows non-OK responses, timeouts, and network errors into `null` defaults. An API failure therefore cannot produce this card — only a client-side throw can.

The only unguarded `.color` read reachable from the providers render tree is in `getLobeProviderIcon()` (`src/shared/components/lobeProviderIcons.ts`):

```ts
const iconKey = LOBE_PROVIDER_ALIASES[providerId.toLowerCase()];
if (!iconKey) return null;
const entry = LOBE_ICON_COMPONENTS[iconKey];   // can be undefined
return type === "color" && entry.color ? entry.color : entry.mono;
```

The `satisfies` constraint keeps the two maps consistent at compile time, but at runtime a `providerId` whose lowercased form collides with an `Object.prototype` property name (e.g. `constructor`, `valueOf`, `hasOwnProperty`) yields a truthy `iconKey` while `LOBE_ICON_COMPONENTS[iconKey]` walks the prototype chain and returns `undefined` → `entry.color` throws with exactly the reported message. `ProviderIcon` calls this for every provider card, so a single such provider id crashes the whole page.

The same unguarded-lookup pattern exists for the other alias maps used by `ProviderIcon` (`PROVIDER_ICON_ALIASES`, `LOCAL_SVG_ALIASES`, `THEMED_SVGS`).

A related, straightforwardly reachable variant: opening `/dashboard/providers/<id>` for an id that is not in the dashboard catalog makes `resolveDashboardProviderInfo` return `null`, which is then spread into `providerInfo` in the detail page client; downstream `providerId.toLowerCase()` throws and the same error card appears (with a `reading 'toLowerCase'` signature).

## Steps to Reproduce

1. Run OmniRoute 3.8.50 (Docker).
2. Create a provider node / custom provider whose id lowercases to an `Object.prototype` property name (e.g. `constructor`), or otherwise reaches `getLobeProviderIcon()` with such an id.
3. Open the Providers dashboard.
4. The page renders the red "Failed to load providers" card; the console shows `TypeError: Cannot read properties of undefined (reading 'color')`.

(Alternative repro for the related variant: open `/dashboard/providers/<id-not-in-catalog>` directly.)

## Expected Behavior

The providers page renders; providers lacking a known icon render a neutral fallback instead of crashing the whole page, and the error card does not appear while the backend is healthy.

## Actual Behavior

A single provider id can crash the entire providers page render; the error boundary shows a misleading connection-error message and hides the real cause in production.

## Test Impact

Needs a new unit test — e.g. `getLobeProviderIcon("constructor", "color")` (and the detail-page null-provider branch) should not throw.

## Validation Plan

- Unit test: `getLobeProviderIcon()` with prototype-colliding ids (`constructor`, `valueOf`, `hasOwnProperty`, `__proto__`) returns `null`/fallback instead of throwing.
- Unit/component test: providers detail page with an unknown provider id renders a "not found" state instead of throwing.

## Additional Context

- **Update — confirmed on a production 3.8.51 deployment:** an inspection of the served minified chunk located the exact helper at the reported crash offset:
  `let r=bM[e.toLowerCase()];if(!r)return null;let n=bD[r];return"color"===t&&n.color?n.color:n.mono`
  The guard `if(!r)` protects only the alias lookup (`r`); the component entry `n=bD[r]` is read unguarded, so a provider id whose lowercased form collides with an `Object.prototype` property name produces `n === undefined` and the exact reported TypeError. A byte-exact count of the whole deployed 3.8.51 minified chunk shows the vulnerable helper appears exactly **once** (`let n=bD[r]` read unguarded after the alias guard), so any in-place patch needs to cover that single site.
- **Data ruled out as the trigger (3.8.51 prod):** a scan of the live SQLite database found zero prototype-colliding identifiers (113 tables, 241 identifier columns, 53,293 distinct values). The triggering provider id is therefore presumed runtime-computed (catalog / model-sync data) rather than stored user input, and the manual repro above may not be reachable through the UI alone; the defect is nevertheless proven at the code level — the unguarded `n=bD[r]` read throws for any such id and reproduces deterministically in a minimal two-map harness.
- **Fix status:** a defensive fix (own-key `Object.hasOwn` lookups, entry guards, non-string id handling in `getLobeProviderIcon()` and `ProviderIcon`) is implemented on `b3nw:fix/11853-prototype-icon-lookup-crash` (PR against `release/v3.8.51` forthcoming) with 8 regression unit tests (`tests/unit/lobe-provider-icons-prototype-collision.test.ts` on that branch), red/green proven. It was validated on the reporter's own 3.8.51 deployment via a temporary in-place chunk patch: the providers page renders normally and the console TypeError is gone. If you hotpatch your own deployment the same way, note Next.js serves `/_next/static/*` with immutable caching — clients only pick up the patched chunk after a hard reload.
- Server logs at the time of the crash were healthy (model sync, credential health, and quota schedulers all normal; no 5xx responses).
- Two OAuth integrations were intentionally unconfigured (bootstrap notice only) — ruled out as a trigger: provider cards are catalog-driven and their colors are hardcoded/derived regardless of connection state.
- Suggested fix directions: guard the record lookups (e.g. `Object.hasOwn`) and branch on `providerInfo == null` in the detail page; consider logging the caught error and revising the error-card copy, since it does not indicate a network problem.

HouMinXi:
Fixes #11853. Prevents TypeError when reading provider.color by adding optional chaining and fallback values. Includes UI test coverage. ⚠️ base-red inherited: #11449

### DIFF
--- a/src/app/(dashboard)/dashboard/providers/[id]/components/ProviderPageHeader.tsx
+++ b/src/app/(dashboard)/dashboard/providers/[id]/components/ProviderPageHeader.tsx
@@ -12,7 +12,7 @@ interface ProviderInfo {
   id: string;
   name: string;
   website?: string;
-  color: string;
+  color?: string;
   apiType?: string;
   /** Optional operator-supplied remote icon URL (#2166) for compatible provider nodes. */
   iconUrl?: string;
--- a/src/shared/components/ProviderIcon.tsx
+++ b/src/shared/components/ProviderIcon.tsx
@@ -330,10 +330,17 @@ const ProviderIcon = memo(function ProviderIcon({
   fallbackColor,
 }: ProviderIconProps) {
   const { isDark } = useTheme();
-  const normalizedId = PROVIDER_ICON_ALIASES[providerId.toLowerCase()] || providerId.toLowerCase();
-  const localSvgId = LOCAL_SVG_ALIASES[normalizedId] || normalizedId;
+  const providerIdLower = providerId.toLowerCase();
+  const normalizedId = Object.hasOwn(PROVIDER_ICON_ALIASES, providerIdLower)
+    ? PROVIDER_ICON_ALIASES[providerIdLower]
+    : providerIdLower;
+  const localSvgId = Object.hasOwn(LOCAL_SVG_ALIASES, normalizedId)
+    ? LOCAL_SVG_ALIASES[normalizedId]
+    : normalizedId;
   const lobeIcon = getLobeProviderIcon(normalizedId, type);
-  const themedSvg = THEMED_SVGS[normalizedId];
+  const themedSvg = Object.hasOwn(THEMED_SVGS, normalizedId)
+    ? THEMED_SVGS[normalizedId]
+    : undefined;
   const hasSvg = KNOWN_SVGS.has(localSvgId);
   const hasPng = KNOWN_PNGS.has(normalizedId);
 
--- a/src/shared/components/lobeProviderIcons.ts
+++ b/src/shared/components/lobeProviderIcons.ts
@@ -487,7 +487,7 @@ export function getLobeProviderIcon(
   type: "mono" | "color" = "color"
 ): LobeIconComponent | null {
   const iconKey = LOBE_PROVIDER_ALIASES[providerId.toLowerCase()];
-  if (!iconKey) return null;
+  if (!Object.hasOwn(LOBE_ICON_COMPONENTS, iconKey)) return null;
 
   const entry = LOBE_ICON_COMPONENTS[iconKey];
   return type === "color" && entry.color ? entry.color : entry.mono;
--- a/tests/unit/ui/ProviderIcon-icon-url.test.tsx
+++ b/tests/unit/ui/ProviderIcon-icon-url.test.tsx
@@ -137,3 +137,18 @@ describe("ProviderIcon — Qwen Cloud local asset", () => {
     }
   );
 });
+
+describe("ProviderIcon — inherited object property ids", () => {
+  it.each(["constructor", "valueOf", "hasOwnProperty", "__proto__"])(
+    "renders provider id %s through the unknown-provider fallback",
+    (providerId) => {
+      const container = renderIcon({ providerId });
+      const img = container.querySelector("img");
+
+      expect(img).not.toBeNull();
+      expect(img?.getAttribute("src")).toBe(
+        `https://thesvg.org/icons/${providerId.toLowerCase()}/default.svg`
+      );
+    }
+  );
+});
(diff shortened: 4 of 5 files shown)


diegosouzapw:
Thanks for this — good, thorough find. #11880 landed first and already fixed
getLobeProviderIcon() itself (both the alias lookup and the component lookup, so
overlapping with your guard there), which conflicted with your branch by the time I got
to it.

But you caught something #11880 missed: ProviderIcon.tsx has three *other* plain-object
lookups (PROVIDER_ICON_ALIASES, LOCAL_SVG_ALIASES, THEMED_SVGS) that resolve the same
inherited-property ids through the prototype chain and weren't touched by #11880. I
ported those three guards plus your ProviderPageHeader `color?` widening as #11935, with
your inherited-property-ids test suite carried over. 90/90 + 4/4 focused tests passing.
Full credit for the thorough audit — thank you for not stopping at the one function.
Closing this one now that it's landed.

PR #11959, opened 2026-08-28
fix(sse): give extended-thinking targets the reasoning readiness budget

### LINKED ISSUE #11922: feat(providers): Kiro provider: 125000ms non-ping SSE idle timeout not configurable anywhere
### Description

Requests routed to the Kiro provider (AWS CodeWhisperer backend), model `claude-sonnet-5-thinking`, are occasionally killed with:

```
[504] Stream produced no non-ping SSE event within 125000ms [kiro/claude-sonnet-5-thinking (504)]
```

This happens when the upstream goes quiet during a long tool-use/reasoning turn (no token emission, ping-only), similar to the pattern already reported in #4038 and #10231, but for the Kiro provider specifically.

**The problem:** there is no way I can find to configure or raise this 125000ms cutoff for Kiro.

- **Dashboard → Settings → Providers → Kiro** has no "Request Timeout" field. Unlike API-key providers (e.g. NVIDIA, per discussion #2098), Kiro is OAuth/device-code based and its provider settings page only exposes account/region/connection management — no timeout control.
- The documented env vars (`REQUEST_TIMEOUT_MS`, `STREAM_IDLE_TIMEOUT_MS`, `FETCH_BODY_TIMEOUT_MS`) all default to inheriting `REQUEST_TIMEOUT_MS` = 600000ms. 125000ms doesn't match any documented default, which suggests this value is hardcoded specifically in the Kiro/CodeWhisperer executor's stream-idle watchdog rather than being wired to the shared timeout config.

### Steps to Reproduce
1. Configure a Kiro (CodeWhisperer) connection in OmniRoute.
2. Drive an agentic client (in my case, Kilo Code / VS Code) through OmniRoute with model `kiro/claude-sonnet-5-thinking`.
3. Trigger a turn with a long internal reasoning/tool step (observed via the client's own tool calls taking a while).
4. If the upstream stays quiet (ping-only) for 125s, the stream is aborted with a 504, even though the request may still be alive server-side and would otherwise complete successfully (confirmed by immediately retrying, which usually succeeds fast right after).

### Expected Behavior
Either:
- Expose a Request Timeout / stream-idle timeout control for the Kiro provider in the dashboard (matching what API-key providers already have), or
- Confirm which env var actually governs this specific 125000ms value for the Kiro executor path, since it doesn't match the documented `STREAM_IDLE_TIMEOUT_MS` default.

### Actual Behavior
125000ms appears to be a fixed, undocumented, non-configurable idle-stream cutoff specific to the Kiro provider's SSE handling, causing otherwise-healthy long-running turns to be killed and forcing a manual retry.

### Environment
- Client: Kilo Code (VS Code extension) → OmniRoute → Kiro provider
- Model: `claude-sonnet-5-thinking`
- OS: Linux

### Related
- #4038 (closed) — general ask for configurable stream-idle timeout on chat-completions wire
- #10231 (closed) — same "Stream produced no non-ping SSE event" symptom class, different provider (`oc/nemotron-3-ultra-free`, 80000ms)
- #6062 (closed) — same symptom class, different provider (`copilot-m365-web`), root-caused to an unrelated `ws` bundling bug

This is filed separately since none of the above cover the Kiro provider specifically, and Kiro's OAuth-based provider settings page has no timeout field to check/adjust in the first place.

abhisheksharma2411:
## Summary

`kiro/claude-sonnet-5-thinking` fails with `[504] Stream produced no non-ping SSE event within 125000ms` when the upstream stays ping-only through a long reasoning turn.

125000 is not a hardcoded Kiro timeout — it is `STREAM_READINESS_TIMEOUT_MS` (80s) plus the 45s very-large-history/payload bump, capped there because nothing in `resolveStreamReadinessTimeout` recognised the request as a reasoning target.

The policy already grants a 30s warm-up allowance to codex high-effort models and to Claude-format replicas, but it keys the latter on the provider registry's `format` field. Kiro serves Anthropic thinking models through its own CodeWhisperer translator, so `format: "kiro"` excludes it — and the `-thinking` alias was never a reasoning signal the way `-high` is. `devin`, `antigravity` and `chatgpt-web` publish `-thinking` ids from outside claude format too, so they had the same gap.

This detects the `-thinking` alias suffix directly and applies the same 30s allowance, so `kiro/claude-sonnet-5-thinking` gets 155s instead of 125s. The three reasoning bumps are mutually exclusive — they model the same one-off warm-up, so a single request cannot stack them into a multi-minute readiness window.

Scoped to the alias suffix (`-thinking`, `-thinking-1m`) rather than a substring, so an unrelated id that merely contains the word does not qualify.

## Related Issues

- Closes #11922

## Validation

- [x] Change type: routing (SSE stream readiness)
- [x] `npm run lint` — clean on both changed files

Reproduced first: the new #11922 test failed at exactly `125000 !== 155000`, matching the reporter's number before any fix was written.

Focused suites, all green after the change:

```
tests/unit/stream-readiness-policy.test.ts
tests/unit/stream-readiness.test.ts
tests/unit/combo-stream-readiness-fallback.test.ts
tests/unit/runtime-timeouts.test.ts
tests/unit/lmarena-stream-readiness-repro-9306.test.ts
tests/unit/stream-early-eof-breaker.test.ts
→ 76 tests, 76 pass, 0 fail
```

`npx tsc --noEmit` reports the same 4782 pre-existing errors on `release/v3.8.51` with and without this change — it adds none.

Each new guard was mutation-tested; every mutation killed exactly one test:

| mutation | result |
|---|---|
| drop `&& !extendedThinking` from the claude-format bump | 1 fail |
| drop `&& !codexHighReasoning` from the thinking bump | 1 fail |
| loosen `/-thinking(?:-|$)/` to `/thinking/` | 1 fail |

## Tests Added Or Updated

- `tests/unit/stream-readiness-policy.test.ts` — 5 added:
  - the #11922 repro (kiro + large history: 125s → 155s)
  - provider-agnostic small-request case (devin `claude-opus-4-6-thinking`)
  - no stacking with `claude_format_heavy_reasoning`
  - no stacking with `codex_gpt_5_5_high_reasoning`
  - no false positive on an unrelated id containing "thinking"

## Coverage Notes

`open-sse/utils/streamReadinessPolicy.ts` is the only production file changed. Every added branch is covered by the tests above: the bump firing, both exclusivity guards, and the suffix-boundary rejection. Coverage does not move down in any touched file.

## Reviewer Notes

- Behavioural change is bounded: a `-thinking` request gets at most +30s of readiness budget, still clamped by `maxTimeoutMs`, and nothing else in the request path changes.
- Worth a second opinion on scope: this treats `-thinking` as a reasoning signal for **every** provider. That is deliberate — the alias means the client asked for extended thinking regardless of who serves it — but if you would rather gate it to a provider allowlist, say so and I will narrow it.
- I did not touch the Kiro executor or add a per-provider timeout control, which is the other option #11922 raises. The env vars (`STREAM_READINESS_TIMEOUT_MS`, `STREAM_READINESS_MAX_TIMEOUT_MS`) already exist and do govern this path; the reporter's real problem was the default being too low for thinking targets, not the absence of a knob. Happy to add the dashboard control as a follow-up if you want it.


abhisheksharma2411:
CI note, so nobody has to dig: the three red checks here are all base-branch state, not this change.

**No new ESLint warnings** — I ran the job's exact command on both refs. `release/v3.8.51` alone:

```
$ npm run lint:json -- --max-warnings 0
exit=2
There are suppressions left that do not occur anymore. Consider re-running the command with `--prune-suppressions`.
```

Identical on this branch — same exit code, same message, byte-identical output apart from a `[BABEL]` size note about a cached `dist/` artifact. The `exit 2` comes from stale suppression entries, which is exactly what #11924 measured in a clean-room clone.

**Build (advisory)** and **dast-smoke** are the hosted-runner OOM in #11946.

All 4 unit-test shards, Vitest, semgrep, docs gates and merge integrity are green, which is the part this PR actually touches.

diegosouzapw:
## Babysit summary (#11959)

**Verdict: green** for required checks. Not merging (human merge).

Contributor credit preserved: original commit `2ad1f3cc` remains authored by @abhisheksharma2411.

### Before (head `2ad1f3cc`)
- **No new ESLint warnings** — FAIL (stale suppressions: "There are suppressions left that do not occur anymore")
- **dast-smoke** — FAIL (hosted-runner OOM; inherited infra #11946/#11976)
- **Build (advisory)** — FAIL (hosted-runner shutdown/OOM; advisory)
- Fast Quality Gates, unit shards, Vitest — PASS

### What we did
PR was 37 commits behind `origin/release/v3.8.51`. Merged current base (includes #11940 / #11955 ESLint suppressions re-freeze, #11975) and pushed to the fork (`maintainerCanModify: true`).

Pushed SHA: `8220a44b334b2a8a226e062e541a0c9055f82172`  
(`2ad1f3cc` + merge `origin/release/v3.8.51`)

No production-file edits beyond the merge. ESLint gate not widened.

### After (head `8220a44b`)
- **No new ESLint warnings** — PASS
- Fast Quality Gates — PASS
- Unit Tests fast-path 1/4–4/4 — PASS
- Vitest, Docs Gates, Merge integrity, semgrep — PASS
- **dast-smoke** — not scheduled on this run (no longer on the PR rail after #11962)
- **Build (advisory)** — FAIL (`npm run build` cancelled on `ubuntu-latest` mid-compile; same advisory hosted-runner OOM/shutdown as before — ignored; Mergify already allows this single failure)

Ready for owner merge when the ⭐ gate is applied.

Issue #8250, opened 2026-07-23
fix(providers): Kimi K3 capability metadata conflict — attachment=false but modalities include image/video


jackjinke:
## Summary

Synced capability metadata for Kimi K3 is internally inconsistent:

- `attachment` / vision capability resolves to `false`
- `modalities_input` includes `image` and `video`

Observed provider/model: `kimi-coding-apikey/k3`.

This makes OmniRoute report K3 as non-vision while still exposing image/video input modalities to clients and downstream capability consumers.

## Observed metadata

From deployed `model_capabilities` table:

```json
{
  "provider": "kimi-coding-apikey",
  "model_id": "k3",
  "attachment": 0,
  "modalities_input": "[\"text\",\"image\",\"video\"]",
  "modalities_output": "[\"text\"]",
  "reasoning": 1,
  "tool_call": 1,
  "limit_context": 1048576
}
```

Equivalent rows also exist for aliases/providers such as `kimi-coding`, `kmc`, and `kmca`.

## Why this matters

`resolveVisionCapability()` currently gives synced `attachment` priority. Because `attachment=false`, K3 resolves as non-vision even though the same synced row declares image/video input modalities.

Expected Kimi K3 metadata should be reconciled explicitly:

- if K3 supports image input, `attachment` / `supportsVision` should be true;
- if `attachment=false` is correct for this endpoint, `input_modalities` should not expose image/video for this provider/model.

Leaving both values unqualified makes `/v1/models`, combo compatibility, Vision Bridge, and clients disagree.

## Possible root cause

models.dev/provider sync imports raw modality metadata and attachment flags independently. `resolveVisionCapability()` intentionally prioritizes `attachment` over modality inference, but there is no consistency check that rejects or normalizes contradictory synced rows.

## Related issues

- #4071 — image requests routed to vision-incapable models when metadata mislabeled capability
- #4073 — synced vision/attachment metadata missing for Mistral models
- #4264 — imported models lost vision capability despite `architecture.input_modalities`
- #2582 — missing `supportsVision` flags for MiMo models
- #2822 — image inputs routed to an upstream model that did not support vision

## Acceptance criteria

- Kimi K3 has one authoritative vision capability value.
- `attachment`, `supportsVision`, and exposed `input_modalities` agree.
- Contradictory synced capability rows are normalized or rejected at sync time.
- Regression test covers `attachment=false` plus `input_modalities` containing image/video.

PR #4071, opened 2026-06-17
fix(sse): route image requests only to confirmed-vision combo targets

diego-anselmo:
## Problem

Combo routing can send a request carrying an image to a text-only model. `getResolvedModelCapabilities` returns `supportsVision: null` for any model lacking a models.dev `attachment` flag and a registry vision flag — including multimodal **Mistral Pixtral** — and the combo compatibility filter only dropped a target when `supportsVision === false`. So a `null` (unknown) text model such as `ministral-14b` slipped through, received the image, and answered "image not provided" / `IMAGEM_INDISPONIVEL`.

**Not a payload-format bug.** Verified live: `pixtral-12b` describes the same image identically whether `image_url` is the OpenAI object `{ "url": ... }` or a bare string. The original report's A/B was confounded — its two rows hit *different* effective models via a vision combo (a text-only `ministral` vs `pixtral`), so the format was never the cause.

## Fix

1. **`src/lib/modelCapabilities.ts`** — `resolveVisionCapability` gains a conservative, last-resort model-id heuristic (`VISION_MODEL_ID_FRAGMENTS`) so known multimodal families (pixtral, llava, qwen-vl, gpt-4o, gemini, claude-3+, glm-4v, mistral-medium-3, …) resolve to `true` when no synced/registry/spec data exists. Kept high-precision on purpose: a false positive would re-create the bug. Anything unmatched stays `null`.
2. **`open-sse/services/combo.ts`** — for a request that carries an image, a target is vision-incompatible unless `supportsVision === true` (was `=== false`). The existing "keep all when none qualify" fallback preserves behavior for combos with no confirmed-vision member, so there is no regression. `filterTargetsByRequestCompatibility` is exported for the test.

## Tests

`tests/unit/combo-vision-aware-routing.test.ts` (red → green):
- heuristic resolves Pixtral → `true`; a text-only Mistral model is not a vision false-positive;
- image request drops the non-vision target and keeps the vision target;
- image request with no confirmed-vision target keeps all (fallback, no regression);
- text-only request is untouched by the vision filter.

`typecheck:core` + `eslint` clean. Existing suites green (model-capabilities-registry, combo-projection-capabilities-3979, provider-registry-qwen-vision #2822, guardrails/visionBridge, auto-combo).

## Follow-ups (separate)

- Consolidate the 3 independent vision-detection lists (`VISION_MODEL_ID_FRAGMENTS`, catalog `VISION_MODEL_KEYWORDS`, lite `modelSupportsVision`) into one shared source.
[...]

PR #11987, opened 2026-08-29
fix(catalog): derive built-in auto-combo modalities from the effective target pool

### LINKED ISSUE #11947: [BUG] Combo models (owned_by: "combo") missing capabilities.vision / input_modalities / output_modalities in /v1/models
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

macOS

### OS Version

Tahoe 26.6.2

### Node.js Version

26.7.0.

### Provider(s) Involved

_No response_

### Model(s) Involved

auto/vision

### Client Tool

OmniRoute version: v3.8.50 Client: OpenCode, via opencode-omniroute-auth plugin (dynamic model fetching from /v1/models) Base URL: http://localhost:20128/v1

### Description

Combo/auto models (e.g. auto/best-vision, auto/pro-vision, auto/claude-sonnet) are tagged as vision-capable in the dashboard (/dashboard/combos), but the /v1/models endpoint does not include capabilities.vision, input_modalities, or output_modalities for these entries at all. Because OpenAI-compatible clients (e.g. the opencode-omniroute-auth plugin for OpenCode) rely on these fields to detect vision support, combo models are surfaced as text-only even when every model in the underlying pool supports images.

### Steps to Reproduce

Start OmniRoute and confirm auto/best-vision is listed in /dashboard/combos with the vision tag.
Query the models endpoint:
bash
   curl -s http://localhost:20128/v1/models \
     -H "Authorization: Bearer $OMNIROUTE_API_KEY" \
     | jq '.data[] | select(.id == "auto/best-vision")'
Compare against a regular (non-combo) vision model that is part of the same pool, e.g.:
bash
   curl -s http://localhost:20128/v1/models \
     -H "Authorization: Bearer $OMNIROUTE_API_KEY" \
     | jq '.data[] | select(.id == "kr/claude-sonnet-5")'

### Expected Behavior

Combo models should expose the same LCD-aggregated capabilities.vision, input_modalities, and output_modalities fields that are already computed for the dashboard (/dashboard/combos shows a vision tag on auto/best-vision and auto/pro-vision) and presumably for /api/combos/auto. Per the docs (OpenCode Integration / DeepWiki), capability flags for combos are already "combined by logical AND or appropriate aggregation" — that computed result just isn't being included in the /v1/models serialization for combo entries.
Impact
Any OpenAI-compatible client that determines vision support from /v1/models metadata (rather than trial-and-error on requests) will treat every auto/* combo as text-only, regardless of the actual pool composition. Concretely, this breaks image attachments in OpenCode when using auto/best-vision, auto/pro-vision, or any other vision-tagged combo, even though the individual models behind them (e.g. kr/claude-sonnet-5) are correctly advertised as vision-capable. This seems related to but distinct from #989 (image content parts failing even for directly-specified vision models) — this report is specifically about the /v1/models metadata for combo/auto entries being incomplete, independent of whether the underlying request-forwarding bug is fixed.
Suggested fix
When serializing a combo model (owned_by: "combo") for /v1/models, reuse the same LCD-aggregation logic already used to compute dashboard tags / /api/combos/auto, and include the resulting capabilities.vision, input_modalities, and output_modalities in the OpenAI-compatible model object — the same way it's done for regular (non-combo) models.

### Actual Behavior

Actual result
Combo model (auto/best-vision, same shape for auto/claude-sonnet, auto/pro-vision, etc.):
json
{
  "id": "auto/best-vision",
  "object": "model",
  "created": 1787949360,
  "owned_by": "combo",
  "permission": [],
  "root": "auto/best-vision",
  "parent": null,
  "context_length": 1050000,
  "max_input_tokens": 1050000,
  "max_output_tokens": 1048576,
  "capabilities": {
    "tool_calling": true,
    "reasoning": true,
    "thinking": true,
    "temperature": true
  }
}
No vision key under capabilities, and no input_modalities / output_modalities keys at all.
Regular model in the same pool (kr/claude-sonnet-5):
json
{
  "id": "kr/claude-sonnet-5",
  "object": "model",
  "owned_by": "kiro",
  "capabilities": {
    "vision": true,
    "tool_calling": true,
    "reasoning": true,
    "thinking": true,
    "supportsThinking": true,
    "effort_tiers": ["none", "low", "medium", "high", "xhigh"]
  },
  "input_modalities": ["text", "image"],
  "output_modalities": ["text"],
  "context_length": 1000000,
  "max_output_tokens": 128000,
  "max_input_tokens": 1000000,
  "name": "kr/Claude Sonnet 5",
  "pricing": { "input": 3, "output": 15, "cached": 1.5, "reasoning": 15, "cache_creation": 3 }
}

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

### Screenshots

Image

### Additional Context

_No response_

### Validation Plan

_No response_

xiaoyaner0201:
Fixes #11947

Built-in `auto/*` rows were emitted with a hardcoded four-key `capabilities` object and
no modality derivation, so `auto/best-vision`, `auto/pro-vision` and `auto/vision`
advertised neither `capabilities.vision` nor `input_modalities` / `output_modalities`
even when every effective target unanimously confirmed vision.

> **Status: Draft.** Not requesting review yet — see *Open questions* at the end.
> This branch has been rebound onto `release/v3.8.51` and re-reviewed after #12046 landed.
> Every update, including the TS2677 fix, is an ordinary fast-forward of the previous head;
> nothing has been force-pushed, rebased, amended or squashed.
>
> The `API Route Typecheck` failure previously flagged at the top of this body **is fixed**
> and the check is green on this head. The correction notice has been retracted accordingly;
> the *Method note* below keeps the part that is useful to others.

## Relationship to the merged #12046

#12046 landed the positive half of this contract while this PR was open. This branch now
**merges** it rather than competing with it, and what remains is the residual gap #12046
left: the auto path synthesized `["text","image"]` from a *shared helper* rather than
reporting what each target actually declares, and the vision verdict could be emitted
without image evidence in the intersection.

All three of #12046's upstream cases (from blob `b9e4552c247ce42cf4a7a71b6f9d9f33ff09ee80`)
are present and passing here, none weakened into vacuous coverage.

## What this changes

`src/app/api/v1/models/catalog.ts`, one region in the built-in auto loop (~50 lines net):

1. **Report each target's raw synced modality evidence.** `inputModalities` /
   `outputModalities` now come from the target's own `modalities_input` /
   `modalities_output` rather than the helper's synthesized pair, so an auto row can only
   advertise image input that some target genuinely declares.
2. **Keep the vision verdict canonical.** The verdict continues to come from
   `resolveVisionCapability` via `metadata.capabilities.vision` — *not* from the raw
   `synced.attachment` flag. This matters: the resolver applies the operator
   `supportsVision` override (#9195) and the documented text-only denylist (#4071)
   **before** it ever reads `attachment`, and it deliberately promotes `attachment:false`
   alongside image modalities (#8250). Reading the raw flag would bypass all three.
3. **Gate emission on image evidence.** `capabilities.vision = true` iff the emitted
   `input_modalities` contains the exact member `"image"` **and** every effective target's
   resolved verdict is exactly `true`. A non-unanimous verdict also suppresses an
   image-bearing input intersection, preserving the invariant in both directions.
4. **Fail closed on unknown targets.** Any effective target resolving to null metadata
   omits all three fields, rather than letting the known subset overstate a
   partially-unknown pool.
5. **Absence is omission, never `false`.** All three fields are emitted via conditional
   spreads: no `vision: false`, no empty arrays.

Ids, ordering, ownership, token limits, the
tool_calling/reasoning/thinking/temperature baseline, the persisted-combo path and the
direct path are unchanged. Vision is never inferred from an `auto/*` id.

No memoization is included. An earlier revision of this branch added per-target memos;
they were dropped deliberately after review found the block adds one id resolution and one
in-memory map lookup per target on an axis the release already pays, the loop still yields,
and no material catalog-build regression reproduced.

## Test evidence — real measured counts

Base: `release/v3.8.51`. Verified against **both** `718accb03d86d1c421df13bae3b4b911e845a004`
(the PR's recorded base) and the current tip `78fd3504dd8c8333abbc406bc4cb83443693af06` —
the three lint batches that landed in between do not touch `src/app/api/**`, and the API
typecheck diagnostic log is byte-identical on the two tips.

All typecheck numbers below come from the **real compiler**, invoked as
`node node_modules/typescript/lib/tsc.js --pretty false --noEmit -p tsconfig.typecheck-api.json`
(TypeScript 6.0.3), on complete `git archive` snapshots. Not via `npx tsc` — see *Method note*.

### API Route Typecheck

| tree | total errors | `catalog.ts` |
|---|---|---|
| pristine upstream `78fd3504d` | 289 | clean, zero TS2677 in the project |
| previous head `770b5c3739` | 290 | `catalog.ts(922,49): error TS2677` |
| **this head `351a867eb`** | **289** | **clean** |

The fixed tree's diagnostic log is **byte-identical** to the pristine upstream log (sorted
diff empty) — not merely equal in count. No baseline file was edited; `config/quality/api-typecheck-baseline.json`
is untouched.

The repo's own gate was run both ways as a control:

| tree | `node scripts/check/check-api-typecheck.mjs` |
|---|---|
| previous head `770b5c3739` | exit **1**, `✗ src/app/api/v1/models/catalog.ts TS2677 (baseline 0, live 1)` — reproduces the CI signature exactly |
| this head `351a867eb` | exit **0**, `apiTypecheckErrors=289`, 0 regressions |

`node --import tsx/esm --test tests/unit/build/check-api-typecheck.test.ts` (the workflow's
second step): 8 tests / 8 pass.

### Behaviour equivalence

This is a type-level change, and that is *proven* rather than asserted, two independent ways:

1. **Emitted JavaScript is byte-identical.** Both versions of `catalog.ts` were transpiled
   with the same esbuild that `tsx` and Next/SWC use. Output sha256
   `eb41bebd089f5ecfe05770be0e90429aa56e6d2eaffc278b0666479f7a923d4f` — identical. Every
   `tsconfig` in the repo sets `noEmit`, so no build artifact is derived from type
   annotations at all.
2. **Emitted catalog rows are byte-identical.** The full `/v1/models` response was captured
   across 12 scenarios — empty pool, vision pool, unresolved target, missing input evidence,
   mixed verdicts, denylist-only, operator override, `attachment:false`-with-image,
   text-only, user combo, heterogeneous pool, output-only evidence — before and after.
   Both payloads (3.1 MB) hash to
   `98fe3c4dbda4e425884468523c5d43fdeb3d74f723bd499eb93dd8419e272329`. The matrix is not
   vacuous: it exercises 0/6/28 vision-advertising auto rows depending on scenario.

### Test gates

| gate | result |
|---|---|
| focused suite `tests/unit/11947-auto-combo-modalities.test.ts` | **11 tests / 11 pass / 0 fail**, exit 0 |
| widened Node catalog regressions, **193** `*catalog*` + `*models*` files | 1212 tests / **1208 pass** / 3 fail / 1 skip |
| the same 193 files on **pristine upstream** | 1212 tests / **1208 pass** / 3 fail / 1 skip — identical |
| `npm run test:vitest -- tests/unit/autoCombo/` | 16 files / **122 tests, all pass**, exit 0 |
| Prettier, changed files | exit 0 |
| ESLint `catalog.ts` | 10 errors here, **10 on pristine upstream**, same rules and symbols, line numbers shifted only — zero introduced |
| `config/quality/eslint-suppressions.json` | not modified by this PR; the one suppression comment inside `catalog.ts` is byte-identical to upstream |

**The three regression failures are attributed, not assumed.** The widened sweep was run on
this branch and on a complete upstream snapshot with the same command: the totals and the
failing set match exactly. They are `runCursorAgent — sigkillFollowupMs` (two subtests, a
pre-existing timing test) and `radar-catalog-capabilities.test.tsx` (environmental — package
`react18-json-view` is not installed locally; fails identically on both sides).

**RED reproduced before the fix.** The committed suite was run against complete `git archive`
snapshots of live (full-tree snapshots — a single-file copy resolves the wrong config and
gives false results): 9 pass / 2 fail, exit 1. All three of #12046's upstream cases are
present and passing here, none weakened into vacuous coverage.

**Probe discrimination.** Passing probes alone do not prove closure, so an earlier rejected
revision of this branch (`d425242fc7`) was materialized and the review probes run against it:
two of them fail there and pass here. On live, 5 of the 11 adversarial probes fail; all 11
pass here.

## The TS2677 fix, and why this shape

The previous head introduced one TypeScript error, which CI correctly caught. In the
built-in auto loop the map callback returns

```ts
return { ...metadata, inputModalities: …, outputModalities: … };
```

Both keys are optional (`?:`) on `ComboTargetCatalogMetadata`, but writing them explicitly —
even as `undefined` — makes them **required** as `string[] | undefined` in the callback's
inferred return type. The array is therefore not `(ComboTargetCatalogMetadata | null)[]`, and
the `(m): m is ComboTargetCatalogMetadata` predicate on the following `.filter` is no longer
assignable to its parameter.

The fix annotates the callback's return type as `ComboTargetCatalogMetadata | null`. No
runtime expression is touched.

Three shapes were evaluated, and the two alternatives were rejected on evidence:

- **Conditional spreads** (`...(x ? { inputModalities: x } : {})`) also clear TS2677, and
  match the style used at the emission site — but they are **not behaviour-preserving here**.
  When this scope's raw `synced` lookup misses, the spread lets the helper's *synthesized*
  `["text","image"]` from `...metadata` survive instead of being overwritten with `undefined`.
  Measured, that variant makes `auto/chaos` and `auto/best-chaos` advertise
  `input_modalities: ["text","image"]` and `vision: true` off synthesized evidence — exactly
  the bypass this PR exists to close. This is why the explicit keys are deliberately kept,
  and the code now carries a comment saying so.
- **Dropping the predicate** compiles too (TS 6.0 narrows `filter(m => m !== null)` natively)
  and is behaviour-equivalent, but it leaves the element type as an anonymous inferred shape,
  so the `ComboTargetCatalogMetadata` contract stops being enforced at this boundary. A later
  edit could drift out of the declared type silently.

The chosen annotation is also *stricter* than the status quo, not merely quieter: it enables
excess-property checking on the returned literal, so a typo'd key (TS2353), a wrong value
type (TS2322) or a missing `capabilities` (TS2741) now fail at this call site. Verified with
targeted probes.

## Method note — why the earlier local evidence was wrong

Worth flagging for other contributors, since it cost this PR a red CI cycle:
`scripts/check/check-api-typecheck.mjs` shells out to `npx tsc`. This repo ships no
`node_modules/.bin/tsc`, so locally `npx` falls through to the unrelated `tsc` package on
npm (`tsc@2.0.3`), which prints *"This is not the tsc command you are looking for"* and
**exits 0 without compiling**. The script parses the empty output, reports
`apiTypecheckErrors=0`, and even lists ~163 baselined errors as phantom "improvements" — on
a tree that genuinely regresses. On CI `npm ci` installs the bin, so the gate is correct
there; it is only misleading locally. Any typecheck claim in this PR therefore comes from
`node node_modules/typescript/lib/tsc.js` directly.

That also retires the earlier claim in this body that `catalog.ts` "has no typecheck
coverage". `tsconfig.typecheck-api.json` includes `src/app/api/**/*.ts`; the file is covered
and CI checks it on every PR. The gate was telling the truth and the local evidence was not.

## Honest caveats

- The **full repo unit suite, `build`, typecheck-core and coverage were not run** and are not
  claimed. The gates above are what was actually executed.
- `npx vitest run open-sse/services/autoCombo/` is non-green — `Test Files 3 failed |
  2 passed (5)`, `Tests 23 passed (23)`, `Cannot bundle Node.js built-in "node:sqlite"`.
  This is **environmental and byte-identical on this branch and on a complete live
  snapshot**, so it is not attributable to this change.
- `radar-catalog-capabilities.test.tsx` cannot run in this environment at all (missing
  `react18-json-view`); it fails identically on upstream, so it is reported as unattributable
  rather than passing.

## Known follow-ups, deliberately not fixed here

1. **`-latest` targets whose synced row is stored under the stripped id.** `catalog.ts`
   uses the narrow `getSyncedCapability`; such a target resolves
   `vision=true, modalitiesInput=["text","image"]` yet emits `vision=undefined,
   input=undefined`. Worth noting this is not a regression: on the current release the same
   narrow lookup misses too, so live emits `vision: true` with **no** `input_modalities` —
   an ungated vision claim. This PR converts that into a clean omission, i.e. it removes a
   wrong claim rather than withholding a right one. `getSyncedCapabilityForResolved` is a
   strict superset and would close it, but widening a lookup adds evidence in the
   permissive direction and deserves its own RED/GREEN rather than riding along here.
2. **The mixed-pool test asserts pool non-emptiness rather than membership.** The pool was
   verified correct today, but the assertion would not fail loudly under one direction of
   drift. A two-line tightening.
3. **The resolver's notion of "vision" is broader than this emission gate** —
   `modalitiesDeclareVision` counts `"video"`, while emission requires exact `"image"`. A
   unanimous video-only pool therefore resolves vision-true but correctly emits no `vision`
   key here. Always under-delivery, never over-claiming; flagged as a semantic seam rather
   than a defect in this diff.

## Open questions for maintainers

1. Is the **strict outer fail-closed gate** on unknown targets the behaviour you want? It is
   deliberately stricter than the persisted-combo path, which aggregates its known subset.
   Aligning the two either way is a one-line change — this PR chose the safe direction, but
   the inconsistency is real and your call.
2. Should the emission gate accept `"video"` as vision evidence (follow-up 3), or is exact
   `"image"` the intended contract?
3. Happy to split the raw-modality reporting and the image-evidence gate into two commits,
   or to fold follow-up 1 in here, if either is preferred for review.


diegosouzapw:
Heads-up @xiaoyaner0201 — #11947 was fixed by #12046, which merged on 2026-08-30 while this PR was still in draft. The two changes address the same root cause (built-in `auto/*` rows emitting hardcoded capabilities with no modality derivation) in the same file.

Nothing wrong with your implementation — it is purely a timing collision on parallel work. Closing this to avoid a conflicting second fix; if you spot anything the merged version misses (it covers `catalog.ts` plus a regression test at `tests/unit/11947-auto-combo-modalities.test.ts`), please say so and we will reopen or take a follow-up.

Sorry for the wasted effort, and thanks for jumping on it.

xiaoyaner0201:
Thanks — and no issue on the timing. I did one fresh latest-release check before treating this as terminal, and there is one important scope correction: the current #11987 head is **already based on top of merged #12046**. After #12046 landed, I reconciled the Draft and narrowed it to residual behavior that #12046 does not cover.

On exact current `release/v3.8.51` `30a26d9dcb62bc2e270400a9112b1319314bf78f`, I overlaid only #11987's regression file and ran the real `node:test` harness. Result: **9/11 passed, 2 failed**. The two failures are:

1. **Missing raw input evidence is promoted into image support.** A target with canonical vision verdict `true`, raw `modalities_input: []`, and `modalities_output: ["text"]` is currently emitted as:

   ```json
   {
     "input_modalities": ["text", "image"],
     "output_modalities": ["text"],
     "capabilities": { "vision": true }
   }
   ```

   That advertises image input the target never declared.

2. **A mixed resolved-vision pool still advertises image input.** With `gpt-4o` plus `mimo-v2.5-pro` (the latter is demoted by the existing text-only denylist despite contradictory synced image metadata), current release correctly omits `capabilities.vision`, but still emits:

   ```json
   {
     "input_modalities": ["text", "image"],
     "output_modalities": ["text"]
   }
   ```

   So clients can still route image input into a pool that is not unanimously vision-capable.

The same two focused probes pass **2/2** on closed Draft head `351a867eb12727e1f12ab4bdf5fc9ec010898e31`; its full focused file remains 11 tests. The residual fix keeps the canonical vision verdict from `resolveVisionCapability`, reads raw synced modality evidence independently, and gates image emission on unanimous resolved vision.

So #12046 covers the positive derivation path, but it does not fully subsume the current two-file #11987 residual. I will not reopen this PR or open a follow-up automatically. If you prefer to keep it closed, this can remain executable evidence for a maintainer-owned follow-up; otherwise the existing reviewed Draft head is already the narrow residual candidate, not a competing second implementation.

PR #11998, opened 2026-08-29
fix(providers): use v1beta1 Model Garden publisher list for Vertex Anthropic discovery

### LINKED ISSUE #11991: Vertex provider: auto-synced Claude models not routable ('not available in the active live catalog') — manual key_value definitions required
## Summary

Anthropic Claude models served through the **Vertex AI provider connection** are visible in `GET /v1/models` but rejected at request time with `Model '<id>' is not available in the active live catalog for provider '<provider>'`. The catalog sync endpoint (`POST /api/providers/<connection_id>/sync-models`) fails with `API unavailable — using cached catalog`, and the only way models route successfully is when they are **manually defined** in the `key_value` store (`"source": "manual"`) — auto-discovered catalog entries do not become routable.

- Omniroute version: **3.8.50** (also reproduced on 3.8.49)
- Provider: Vertex AI (googleapis.com), connection `vertex` + `vertex-partner`
- Models: `claude-opus-4-8`, `claude-sonnet-4-6`, `claude-sonnet-5`, `claude-opus-5` (all Claude-on-Vertex)

## Steps to Reproduce

1. Install Omniroute 3.8.50 (bun global install).
2. Add a Vertex AI provider connection (`provider: vertex`) with a valid service-account credential, region `us-central1` (or `global`).
3. Verify the model list exposes the Claude ids:
   ```bash
   curl -s http://localhost:20128/v1/models | jq -r '.data[].id' | grep '^vertex/claude'
   # → vertex/claude-opus-4-8, vertex/claude-sonnet-4-6, ... (present)
   ```
4. Send a chat completion for one of them:
   ```bash
   curl -s http://localhost:20128/v1/chat/completions \
     -d '{"model":"vertex/claude-sonnet-4-6","messages":[{"role":"user","content":"Reply with exactly: OK"}],"max_tokens":10,"stream":false}'
   ```
5. Actual result:
   ```json
   {"error":{"type":"...","code":"...","message":"Model 'claude-sonnet-4-6' is not available in the active live catalog for provider 'vertex'"}}
   ```

## Root Cause Observations

- `model_capabilities` table is **empty** (0 rows) for `vertex`, yet `GET /v1/models` lists the Claude ids from a **cached catalog**.
- Forcing a fresh sync fails:
  ```bash
  curl -X POST -H "Authorization: Bearer $KEY" \
    http://localhost:20128/api/providers/<connection_id>/sync-models?quiet=1
  # → {"error":"API unavailable — using cached catalog","source":"cache","warning":"..."}
  ```
- Request-time availability is gated on a per-connection **"active live catalog"** (not the cached `/v1/models` list). The sync that populates it cannot reach/authenticate Google's model-list API (region-scoped path 404s; `region: global` on the connection does not fix it).
- **Workaround that works:** define the models manually. The `key_value` store entries with `"source": "manual"` (e.g. `vertex` → `[{"id":"claude-sonnet-4-6","source":"manual","apiFormat":"chat-completions","targetFormat":"claude","supportsVision":true}, ...]`) make those exact ids routable — live-verified 200 `OK` through `vertex/claude-sonnet-4-6`.
- So: manual definitions bypass the broken auto-sync; auto-discovered-only models never become routable.

## Environment

- OS: Debian 13 (trixie), kernel 6.12, x64
- Install: bun global (`bun update -g omniroute`), service via systemd user unit
- Python/Node: node 24.19, bun 1.4.0
- Storage: `storage.sqlite` (better-sqlite3 optional dependency gated by bun `trustedDependencies` — postinstall copied native modules manually)
- Related 3.8.50 fix referenced: #9006 (SSE fixes — "a single unrecognized model id on a Vertex connection no longer…"; does not address the catalog-sync issue)

## Expected Behavior

Models auto-discovered from the provider and listed in `GET /v1/models` should also be available at request time without requiring manual `key_value` definitions. The per-connection "active live catalog" should be populated by a successful sync, and `POST /api/providers/<id>/sync-models` should return the models (or a meaningful auth/region error) instead of `API unavailable`.

## Suggested Direction for Maintainers

1. Fix the Vertex catalog sync path (models-list endpoint region handling; `global` region should be honored) so `sync-models` populates the active catalog.
2. Fall back to `key_value` manual definitions when auto-sync fails (currently the only working path).
3. Add a clear error when `model_capabilities` is empty but `/v1/models` shows cached entries — today's `"not available in the active live catalog"` message is misleading.

fabioluissilva:
Closes #11991

## Problem

Vertex AI Anthropic (Claude) partner models are visible in `GET /v1/models` but rejected at request time with:

```
Model 'claude-sonnet-4-6' is not available in the active live catalog for provider 'vertex'
```

The auto-sync never populated the active synced catalog (`model_capabilities` empty). The only workaround was manually defining the models in `key_value` with `"source":"manual"`.

## Root cause

The Anthropic partner-model discovery path used the **v1 API list** URLs:

- `GET /v1/projects/{project}/locations/{region}/publishers/anthropic/models`
- `GET /v1/publishers/anthropic/models` (global)

Both 404 at the Google Front End because the **v1 API does not support list operations** for Model Garden publisher models. Verified with a valid service-account credential: all three path variants (project-scoped us-central1, region-prefixed hostnames us-east5/us-central1/global, and the bare global path) return HTTP 404. As a result the Claude catalog never reached the synced store, so the request-time gate in `lookupModelMeta` (`src/sse/services/model.ts`) rejected every id.

The chat path works independently because it targets the single-resource op `publishers/anthropic/models/{model}:rawPredict`, which is supported on v1.

## Fix

The Model Garden publisher-model **list** is served by the **v1beta1** API and is **global** (catalog-wide, not project/region-scoped):

```
GET https://aiplatform.googleapis.com/v1beta1/publishers/anthropic/models
```

Verified with a real SA credential: returns `HTTP 200` with `{ "publisherModels": [ ... ] }` containing all 11 Claude models (`claude-sonnet-4-6`, `claude-opus-4-8`, `claude-sonnet-5`, `claude-opus-5`, ...).

Changes:
- `src/app/api/providers/[id]/models/route.ts`: call the global `v1beta1` publisher list (drop the broken project/region-scoped `v1` path).
- `src/lib/providerModels/vertexAnthropicModelsParser.ts`: parse the `publisherModels` v1beta1 envelope (keeps `models` fallback).
- `tests/unit/vertex-anthropic-models.test.ts`: regression tests incl. a real-response snapshot. All 6 pass.
- `changelog.d/fixes/11991-...md`: release fragment.

Execution region remains handled separately by the vertex executor at request time — the list endpoint does not need it.

## Verification

- `node --test tests/unit/vertex-anthropic-models.test.ts` → 6/6 pass
- `tsc --noEmit` clean
- ESLint: no new errors introduced
- `check-changelog-integrity` → OK
- Live repro: v1 paths 404 vs v1beta1 path 200, real SA

## Notes

Registration/PR conventions per `CONTRIBUTING.md`: branch from `release/v3.8.51`, changelog fragment under `changelog.d/fixes/`. No `Co-Authored-By` trailers.


fabioluissilva:
Hi maintainers 👋 — quick non-urgent ask: the **Build (advisory)** job on this PR was canceled ("The operation was canceled" at `.github:29` — looks like a concurrency/infra cancellation, not a compile failure). All required gates passed on head `063591ac3` and the workflow run itself concluded `success`, so nothing is blocked on my side.

As a fork contributor I don't have admin rights to trigger a re-run from the API — could someone hit **Re-run jobs** on the Build job in the Actions tab when convenient? It is advisory-only, so there is genuinely no rush.

Thanks! 🙏

(Reporter of #11991, author of this PR)

PR #12017, opened 2026-08-29
fix(admission): exclude reclaimable page cache from the cgroup pressure ratio

HouMinXi:
fix(admission): exclude reclaimable page cache from the cgroup pressure ratio

## Problem

The resource-pressure guard ratios raw cgroup v2 `memory.current` against
`memory.max`. `memory.current` counts reclaimable file cache, so a busy host
whose page cache has grown trips the guard even though the kernel would drop
those pages the instant allocation pressure appears. On a production box
running a 4096MB heap inside a 5GiB limit, ~3GiB of page cache pushed the raw
ratio to 0.87-0.97 and latched a global 503 across every model for 26 minutes
(2026-08-29). The latch itself was unreachable: the recovery gate compares the
same raw ratio against 0.75, and page cache does not shrink on its own, so
only a restart cleared it. Kernel-side evidence during the outage:
`memory.events` (high/max/oom) all zero, memory PSI full avg10 = 0.00, anon
working set 1.62GiB of a 5GiB limit.

## Fix

- Read `file` from `memory.stat` and ratio the working set
  (`current - file`) for the `cgroup_ratio` trip and its recovery check.
  When `memory.stat` is missing or zero, fall back to the raw ratio, so
  hosts without cgroup v2 stat behave exactly as before.
- `memory.current` and `memory.stat` are separate non-atomic reads; if a
  stale `file` exceeds a fresher `current`, treat the sample as bad and use
  the raw ratio rather than clamping to zero (a zero working set would read
  as zero pressure and force a premature recovery).
- `cgroup_high` keeps the raw total charge: the kernel throttles on total
  memory (file cache included) when crossing `memory.high`, so a working-set
  ratio there would miss kernel-side reclaim stalls.
- The sampler binds cgroup reads to named fields instead of positional
  array indices.

## Verification

- 12 new test cases: incident reproduction (raw 95% / workingset 32% stays
  normal), latch release once cache drains, boundary conditions
  (file>current, file=0, file==current, high-limit semantics, zero-vs-
  missing stat), zero file-cache reading distinct from unavailable.
- Bug-injection round trips: reverting `workingSetBytes` to the raw ratio or
  short-circuiting the stat parser makes the incident-shaped tests fail;
  restoring the fix passes.
- Full resource-pressure + admission suites green (48/48); project
  typecheck clean; strict per-file tsc clean.

### DIFF
--- a/open-sse/utils/resourcePressurePolicy.ts
+++ b/open-sse/utils/resourcePressurePolicy.ts
@@ -29,6 +29,7 @@ export type ResourceSignals = {
     currentBytes: ResourceMetricBytes;
     maxBytes: ResourceMetricBytes;
     highBytes: ResourceMetricBytes;
+    fileBytes: ResourceMetricBytes;
     events: {
       low: ResourceMetricBytes;
       high: ResourceMetricBytes;
@@ -176,23 +177,35 @@ export function classifyAdaptiveResourcePressure(
     best,
     ratioLevel(signals.v8.heapUsedBytes, signals.v8.heapLimitBytes, thresholds, "v8_heap_ratio")
   );
+  // memory.current includes reclaimable page cache; the kernel drops those
+  // pages under allocation pressure (memory.events high/max stay 0). Ratio the
+  // working set (current minus file cache) so cache-heavy-but-healthy hosts do
+  // not trip the guard. Without memory.stat, fall back to the raw ratio.
+  const cgroupWorkingSetBytes = workingSetBytes(signals.cgroup);
   best = maxLevel(
     best,
-    ratioLevel(signals.cgroup.currentBytes, signals.cgroup.maxBytes, thresholds, "cgroup_ratio")
+    ratioLevel(cgroupWorkingSetBytes, signals.cgroup.maxBytes, thresholds, "cgroup_ratio")
   );
   best = maxLevel(
     best,
-    ratioLevel(signals.cgroup.currentBytes, signals.cgroup.highBytes, thresholds, "cgroup_high")
+    ratioLevel(cgroupWorkingSetBytes, signals.cgroup.highBytes, thresholds, "cgroup_high")
   );
   best = maxLevel(best, psiLevel(signals.psi?.someAvg10 ?? null, thresholds, "psi_some"));
   return maxLevel(best, psiLevel(signals.psi?.fullAvg10 ?? null, thresholds, "psi_full"));
 }
 
+function workingSetBytes(cgroup: ResourceSignals["cgroup"]): number | null {
+  if (cgroup.currentBytes == null) return null;
+  if (cgroup.fileBytes == null || cgroup.fileBytes <= 0) return cgroup.currentBytes;
+  return Math.max(0, cgroup.currentBytes - cgroup.fileBytes);
+}
+
 function isRecovered(signals: ResourceSignals, thresholds: ResourcePressureThresholds): boolean {
+  const cgroupWorkingSetBytes = workingSetBytes(signals.cgroup);
   const ratios: Array<readonly [number | null, number | null]> = [
     [signals.v8.heapUsedBytes, signals.v8.heapLimitBytes],
-    [signals.cgroup.currentBytes, signals.cgroup.maxBytes],
-    [signals.cgroup.currentBytes, signals.cgroup.highBytes],
+    [cgroupWorkingSetBytes, signals.cgroup.maxBytes],
+    [cgroupWorkingSetBytes, signals.cgroup.highBytes],
   ];
   if (
     ratios.some(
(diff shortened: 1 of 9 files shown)


HouMinXi:
# Load test results: OmniRoute under a 90-agent workload

I ran a staged load test against a production OmniRoute instance (podman,
cgroup memory limit 5 GiB, V8 heap capped at 4096 MB, `MAX_HEAVY_IN_FLIGHT=6`,
queue 30 s) and measured where the local layer actually breaks. Test plan was
frozen before the first run: stage ladder 1/3/6/12/24/45/60/75/90 concurrent
agents, each stage a mix of light/medium/heavy requests (50/30/20), success
gate >=95% protocol success, no OOM, memory ratio under 75% after load.

The load generator runs real requests through the full local chain
(auth, admission, routing, translation, executor, SSE). For the capacity
numbers below the upstream is a local mock server answering in ~50 ms, so the
numbers isolate OmniRoute itself rather than upstream latency. A separate
run against real upstreams (DeepSeek/SenseNova free tiers) is described
afterwards.

## Local-layer capacity

| concurrency | success | rejected by admission | peak memory |
|---|---|---|---|
| 1-45 | 100% (405/405) | 0 | 2.6-3.0 GiB |
| 60 | 99.4% | 0.6% | 2.88 GiB |
| 75 | 95.6% | 4.4% | 2.77 GiB |
| 90 | 92.6% | 7.4% | 2.73 GiB |

Every single failure past 45 concurrent agents is a `chat_admission_busy`
503: the heavy-request admission gate doing its job, not a crash, not memory
pressure, not OOM. Stream TTFB stayed between 2.0 and 2.5 s at every level.
End-to-end latency grows with queueing (p50 5.4 s at 12 agents, 38 s at 90),
which is the price of the same backpressure.

Two things stand out to me in this data. Peak memory does not move with
concurrency: 2.6-3.0 GiB at 1 agent and at 90 agents alike. And the instance
survived 27 consecutive runs at up to 90 agents with zero restarts. On the
5 GiB contract, the practical sweet spot is 60-75 concurrent agents with the
current admission settings; 90 is a working backpressure zone, not a failure
zone.

## Two production bugs the load test flushed out

Both were found because the load generator does something real traffic mostly
doesn't: abort connections mid-stream.

### 1. The resource-pressure detector counts reclaimable page cache as memory

`resourcePressurePolicy.ts` computed its cgroup ratio from raw
`memory.current`. Under load, page cache grows; at some point the ratio
crossed 0.92 and the guard latched at `critical`, returning 503 to everything.
The catch: idle `memory.current` was 4.34 GiB, of which 3.01 GiB was file
cache and only 1.62 GiB anonymous. `memory.events` all zero, PSI 0.00. There
was no real memory pressure at all.

The recovery path made it worse. `isRecovered()` compares the same
cache-inclusive ratio against 0.75. I dropped `memory.reclaim` on the host
until the ratio read 0.16, and the 503s continued for another five minutes
until a container restart. In production this was a 26-minute outage that
started from a single 655-message request plus natural page-cache growth.

Fix: compute the ratio from the working set (`memory.current` minus the
`file` field of `memory.stat`), with a fallback to raw current when the stat
read is unavailable or skewed across the non-atomic reads, and `memory.high`
keeps comparing the raw total because the kernel throttles on total charge.
PR #12017.

Before/after on the identical stage-24 load: 10/10 requests rejected with
503 and a stop-rule abort, versus 24/24 success at 49% peak memory after the
fix.

### 2. The client-abort guard crashes the process it protects

#11556 added `installProcessCrashGuard()` to swallow benign client aborts.
It's called with no argument from `apiBridgeServer.ts`, `liveServer.ts` and
`embedWsProxy.ts`. The default logger was `log ?? console`, and the swallow
path calls `logger("warn", ...)`. `console` is an object, not a function, so
the first aborted connection that reached the process-level guard threw
`TypeError: b is not a function` inside the `uncaughtException` handler and
took the server down. Twice in three minutes during my test run, each time
recovered by systemd. The guard introduced to prevent exactly this crash was
the crash.

Fix is one line, defaulting the logger to `console.warn.bind(console)`, plus
a regression test that exercises the no-argument path with a real
ECONNRESET-shaped error. After deploying it, 27 runs at up to 90 agents with
deliberate mid-stream aborts produced zero uncaught exceptions and zero
restarts. PR #12042.

## Real-upstream runs (context, not the headline)

The same harness ran stages 1-45 against real upstreams. Stage 12 was 72/72.
From stage 24 on, the dominant failure mode shifted to upstream timeouts:
128/144 at 24 agents (SenseNova accounted for 15 of 16 timeouts; DeepSeek was
71/72), 78-84% at 45 agents. The local layer showed zero 503s, zero admission
rejections beyond the expected, zero OOM at every level. The bottleneck from
24 agents up was upstream throughput on free-tier accounts, not OmniRoute.

## Setup notes for anyone reproducing

- Memory contract: quadlet sets `NODE_OPTIONS=--max-old-space-size=4096` and
  `--memory=5g`; check that no `.env` `OMNIROUTE_MEMORY_MB` shadows it,
  because the startup script appends heap flags last-wins.
- The load harness reserves its scratch dir per run and owns the stop-file
  lifecycle; a metrics collector that hard-exits when a stop file exists will
  race a fast stop-rule trip (mine did, and lost the memory timeline for one
  run until I made it warn and continue).
- In rootless podman, a mock upstream on the host is reachable from the
  container only via `host.containers.internal` (169.254.1.2), not via
  127.0.0.1 or the bridge subnet address.

The full run artifacts (per-request JSONL, 2 s-interval cgroup metrics,
pre-registration and incident RCAs) are kept with the deployment; the two fix
PRs (#12017, #12042) carry the code-level evidence.

PR #12018, opened 2026-08-29
fix(build): honor OMNIROUTE_USE_TURBOPACK from .env on the build path

backryun:
## Problem

`npm run build` was the only entry point that never loads `.env`. The dev server goes
through `scripts/dev/run-next.mjs → bootstrapEnv()`, but
`scripts/build/build-next-isolated.mjs` hands `process.env` straight to
`resolveNextBuildBundlerFlag()`.

So the escape hatch documented for the RAM-constrained / macOS arm64 Turbopack stall
(#6409, #9695) was silently ignored on the build path: `.env` said
`OMNIROUTE_USE_TURBOPACK=0` and the build still ran `--turbopack`.

`resolveNextBuildBundlerFlag()` defaults to `--turbopack`, so the only way out was
exporting the variable in the shell for every single invocation — and forgetting once
meant a build that never converged (25+ min with no progress on this machine, while the
same commit is green in CI in ~7 min).

## Fix

Add `honorBundlerFlagFromEnvFile()` and call it at the top of `runNextBuild()`, before
the bundler argument is constructed.

- Shell env keeps precedence — an explicit export still wins.
- `.env` only fills the variable in when it is absent.
- A missing or unreadable `.env` is a no-op.

This ports the DATA_DIR pre-read pattern already used by the dev launcher.

## Tests

`tests/unit/build/resolve-next-build-bundler-flag.test.mjs` — 3 new cases:

- `.env` fills `OMNIROUTE_USE_TURBOPACK` (→ `--webpack`)
- shell env wins over `.env`
- a missing `.env` is a no-op (→ `--turbopack`, unchanged default)

Existing 3 cases unchanged. `tests/unit/build-next-isolated.test.ts` 10/10 — no
regression.

⚠️ base-red inherited: #11874

### DIFF
--- a/scripts/build/build-next-isolated.mjs
+++ b/scripts/build/build-next-isolated.mjs
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 import fs from "node:fs/promises";
-import { mkdirSync } from "node:fs";
+import { mkdirSync, readFileSync } from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { spawn } from "node:child_process";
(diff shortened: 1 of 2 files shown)


backryun:

PR #12019, opened 2026-08-29
fix(css): scope the Tailwind v4 content scan with source(none)

backryun:
## Problem

Tailwind v4 automatic content detection walks the whole repository root from
`src/app/globals.css` — `tests/` (~5k files), `docs/`, `CHANGELOG.md` (~2.4 MB),
`scripts/` — and costs **~13.5 s per cold process**.

That is fine for webpack, whose postcss-loader scans once in-process. The Turbopack
build instead delegates this transform to a pool of short-lived node workers, so
**every worker respawn re-paid the full cold scan** and the build stopped converging
(44 unique workers observed in 11 minutes on this machine).

## Fix

Disable auto-detection and declare the class-bearing trees explicitly:

```css
@import "tailwindcss" source(none);
@source "../";                        /* src/ — every production className lives here */
```

The existing route-group, fumadocs and `not` exclusion directives are untouched.

## Verification

- Cold scan **13.5 s → 2.05 s** (6.6×).
- Production class parity checked before/after: **2,457 classes retained**, the 37
  dropped all originate outside `src/` (tests, docs, scripts) — so no shipped class is
  lost.
- Repo-wide grep confirms `className` usage lives only under `src/` in `.tsx` files.

## Rule for future UI work

Tailwind classes added **outside `src/`** need an explicit `@source` line here.

## Tests

New `tests/unit/tailwind-content-scan.test.mjs` — 3 cases guarding both halves of the
invariant:

- `source(none)` stays set (a revert re-introduces the unbounded scan)
- a positive source root still covers the app tree — `source(none)` with no `@source`
  emits **zero utilities**, an unstyled dashboard with no build error
- the fumadocs / `.sqlite` / `.claude` directives are preserved

Existing CSS-reading tests re-run green: `design-grid-background`,
`material-icons-self-hosted`, `dashboard-sidebar-desktop-visible`,
`connection-row-rtl-margin` (25/25).

⚠️ base-red inherited: #11874

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -1,4 +1,16 @@
-@import "tailwindcss";
+/* PERFORMANCE: `source(none)` disables Tailwind v4 automatic content
+   detection, which walks the whole repository root (tests/ ~5k files, docs/,
+   CHANGELOG.md ~2.4 MB, scripts/...) and costs ~13.5 s PER fresh worker
+   process. The Turbopack build delegates this postcss transform to a pool of
+   short-lived node workers, so every worker respawn re-paid the full cold
+   scan and the build never converged (local macOS arm64 stall, 2026-08-29 —
+   see _tasks/research/2026-08-29-turbopack-local-build-stall-rootcause.md).
+   The explicit @source globs below cover every production class-bearing tree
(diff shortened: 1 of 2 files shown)


backryun:

PR #12029, opened 2026-08-29
fix(memory): include active local/no-auth providers in qdrant embedding models list (#11949)

### LINKED ISSUE #11949: [bug] Qdrant embedding-model list excludes local (no-API-key) providers like Ollama
## Report (from WhatsApp support channel)

> Estou tentando configurar o qdrant no omniroute com um modelo de embedding do ollama localmente, mas ele não fazendo o "indexing" corretamente, sempre que recarrego a página ele aparece esse erro na tela, mas os testes de conexão funcionam e fica "Ativo" em verde. A opção de Remote Provider requer uma chave de API e não lista pra mim o modelo do ollama mesmo estando na tela dos Providers.

## Root cause

`src/app/api/settings/qdrant/embedding-models/route.ts:26-35` builds `configuredProviders` by filtering active provider connections to only those with a non-empty `apiKey` OR `authType === "oauth"`:

```ts
const configuredProviders = new Set(
  activeConnections
    .filter(
      (connection) =>
        (typeof connection.apiKey === "string" && connection.apiKey.trim().length > 0) ||
        connection.authType === "oauth"
    )
    .map((connection) => String(connection.provider || ""))
    .filter(Boolean)
);
```

`ollama-local` is a valid embedding provider (`open-sse/config/embeddingRegistry.ts:358-368` — `embeddinggemma`, `nomic-embed-text`, `bge-m3`) and is explicitly documented as needing **no API key** (`src/shared/constants/providers/local.ts:32-43`, `authHint: "No API key required. Ollama runs locally..."`).

Because the filter only allows `apiKey` present or `authType === "oauth"`, a fully active, tested, green "Ativo" local Ollama connection never lands in `configuredProviders`, so its embedding models never appear in the Qdrant embedding-model dropdown — regardless of connection health.

## Suggested fix

Include local/no-auth connections in the filter, e.g. also accept `connection.authType === "none"` (or however local providers are marked), matching how `ollama-local` and other local providers are defined.

## Repro

1. Add an active Ollama (local) provider connection, verify it tests OK ("Ativo").
2. Go to Qdrant / Memory settings → embedding model selector.
3. Ollama's embedding models (`embeddinggemma`, `nomic-embed-text`, `bge-m3`) are missing from the list.

HouMinXi:
## Summary

`GET /api/settings/qdrant/embedding-models` previously filtered active provider connections to only those with non-empty `apiKey` or `authType === "oauth"`.

Because local embedding providers such as `ollama-local`, `lm-studio` / `lmstudio`, `vllm`, and `mlx-*` require no API key (`authType: "none"`), active and tested local connections were excluded from `configuredProviders`, hiding their embedding models (`embeddinggemma`, `nomic-embed-text`, `bge-m3`) from the Qdrant embedding model selector.

### Fix
- Include connections with `authType === "none"` or known local providers in `configuredProviders`.
- Normalize provider ID aliases (`lm-studio` → `lmstudio`) so they match the embedding registry keys.
- Add regression unit tests in `tests/unit/qdrant-local-embedding-models-11949.test.ts`.

Closes #11949

⚠️ base-red inherited: #11874

### DIFF
--- a/changelog.d/fixes/11949-qdrant-local-embedding-models.md
+++ b/changelog.d/fixes/11949-qdrant-local-embedding-models.md
@@ -1,0 +1,1 @@
+- **fix(memory):** include active local and no-auth provider connections (`ollama-local`, `lm-studio`, `vllm`, `mlx-*`) in `GET /api/settings/qdrant/embedding-models` options list ([#11949](https://github.com/diegosouzapw/OmniRoute/issues/11949)).
--- a/src/app/api/settings/qdrant/embedding-models/route.ts
+++ b/src/app/api/settings/qdrant/embedding-models/route.ts
@@ -28,9 +28,19 @@ export async function GET(request: NextRequest) {
         .filter(
           (connection) =>
             (typeof connection.apiKey === "string" && connection.apiKey.trim().length > 0) ||
-            connection.authType === "oauth"
+            connection.authType === "oauth" ||
+            connection.authType === "none" ||
+            connection.provider === "ollama-local" ||
+            connection.provider === "lm-studio" ||
+            connection.provider === "lmstudio" ||
+            connection.provider === "vllm" ||
+            connection.provider === "mlx-gemma" ||
+            connection.provider === "mlx-qwen"
         )
-        .map((connection) => String(connection.provider || ""))
+        .map((connection) => {
+          const p = String(connection.provider || "");
+          return p === "lm-studio" ? "lmstudio" : p;
+        })
         .filter(Boolean)
     );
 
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Same issue (#11949) already fixed and merged via #11995 (2026-08-29), currently on the release/v3.8.51 tip. That version uses the existing canonical providerAllowsOptionalApiKey() helper (already used for the identical check elsewhere in the codebase) instead of a hardcoded provider-id list — so it automatically covers every current and future no-key-required provider, not just the 6 explicitly listed here (ollama-local/lm-studio/lmstudio/vllm/mlx-gemma/mlx-qwen).

Verified: the merged route.ts already wires providerAllowsOptionalApiKey into configuredProviders exactly where this PR's diff would land. Closing as subsumed — thanks for catching and reproducing this from the WhatsApp support channel report, it corroborated the same bug #11995 fixed.

PR #12060, opened 2026-08-29
fix(build): default to webpack on Apple silicon (Turbopack does not converge, #12059)

### LINKED ISSUE #12059: Turbopack stalls on macOS arm64 (Next 16.3.2): build never converges, dev never listens
## Summary

Turbopack is unusable on **macOS arm64 with Next 16.3.2**: `next build` never converges and
`next dev` never reaches listen state. CI (Linux x64) is green on the same commits, so this is
platform-specific and invisible to contributors who only watch CI.

`.env` still documents the opposite — that "Turbopack panics no longer reproduce" — but that
verification was done against **Next 16.2.9**, while the repo now runs **16.3.2**.

## Environment

- macOS arm64 (Apple M4, 10 cores = 4P + 6E, 32 GB)
- Node 26.8.1 (also reproduced on isolated Node 24.20.0)
- Next **16.3.2** — `.env` documents verification against **16.2.9**

### `next build` — stalls indefinitely

| | Result |
|---|---|
| Turbopack, macOS arm64 | **No completion.** 25 min+ with no progress, chunks pinned at 20 |
| Turbopack, CI Linux x64 | ~7 min, green |
| webpack, macOS arm64 | **13 min**, exits 0 |

Root cause (traced): Turbopack delegates the postcss/webpack-loader transform to a pool of
short-lived node workers. **44 unique worker processes in 11 minutes** — continuous death and
respawn. Every respawn re-paid the Tailwind v4 cold content scan (~13.5 s), so the build never
converged. The main Rust engine sits at 0.4–36% CPU waiting.

Trace evidence: 2,802,951 spans — `turbopack_core::resolve` 427,456,
`turbopack_ecmascript::parse` 247,709, `turbopack_core::module_graph` 53,045.

Upstream: `vercel/next.js#98043` (whole-app module graph effectively serial — "adding hardware
does not shorten it", auto-closed as invalid-link) and `#97802` (Turbopack build 4 GB saturation
/ stall / OOM, **OPEN**).

### `next dev` — never listens

Measured today on this machine, isolated `NEXT_DIST_DIR` and `DATA_DIR`:

```
OMNIROUTE_USE_TURBOPACK=1  →  stuck at "○ Compiling instrumentation Node.js ..."
                              96 s, never reaches listen
OMNIROUTE_USE_TURBOPACK=0  →  "[Next] dev server listening on http://0.0.0.0:20128 (webpack)"
```

## Contributing factor: the documented escape hatch was dead

`resolveNextBuildBundlerFlag()` defaults to `--turbopack` and `scripts/build/build-next-isolated.mjs`
**never loaded `.env`** — unlike the dev launcher, which goes through `bootstrapEnv()`.

So setting `OMNIROUTE_USE_TURBOPACK=0` in `.env` had no effect on `npm run build`. The only way
out was exporting it in the shell for every single invocation, and forgetting once meant a build
that never completed.

→ Fixed in **#12018**.

## Contributing factor: Tailwind scanned the whole repository

Tailwind v4 auto content detection walked the repository root from `src/app/globals.css` —
`tests/` (~5k files), `docs/`, `CHANGELOG.md` (~2.4 MB) — at **~13.5 s per cold process**. Under
webpack that postcss-loader runs once in-process; under Turbopack every worker respawn re-paid it.

→ Fixed in **#12019** (`source(none)` + explicit `@source`): 13.5 s → 2.05 s, verified class
parity (2,457 retained; the 37 dropped all originate outside `src/`).

Note this **does not** unblock Turbopack — retried after the fix, chunks were still pinned at 20.
It is an amplifier, not the root cause.

## Reproduction

```bash
OMNIROUTE_USE_TURBOPACK=1 npm run build   # macOS arm64: stalls, never completes
OMNIROUTE_USE_TURBOPACK=0 npm run build   # macOS arm64: 13 min, exit 0
```

## Proposal

1. **Default to webpack on darwin/arm64** unless the operator explicitly opts in, on the build
   path (and consider the dev path). CI/Linux keeps Turbopack, where it is ~2.9× faster.
2. **Refresh the stale `.env` documentation** — it cites a Next 16.2.9 verification that no
   longer describes 16.3.2 behaviour.
3. **Document the platform constraint in `docs/reference/ENVIRONMENT.md`** so operators do not
   have to rediscover it (precedent: #9695).

## What is *not* the problem

Measured and ruled out during this investigation:

| Hypothesis | Verdict |
|---|---|
| Node 26 regression | Ruled out — reproduced on isolated Node 24.20.0 |
| `.env`-derived config divergence | Ruled out — resolveAlias stubs are opt-in and unset |
| next/font Google CDN stall | Ruled out — 0.18s/200 and 0.30s/200, DNS instant |
| nice/QoS E-core throttling | Ruled out — NI=0, PRI=4, STAT=R |
| Build-scope pollution | Ruled out — `check:build-scope` 8,299 files (limit 12,000) |
| Task-engine eviction churn | Insufficient — `turbopackMemoryEviction: false` did not help |
| `experimental.turbotrace` | Not available — removed in Next 16.3.2 |
| NFT trace duplication | Real (1,680→2 manifests) but **no time gain** (219 s → 227 s) |

## Related

- PR #12018 — build path honours `OMNIROUTE_USE_TURBOPACK` from `.env`
- PR #12019 — scoped Tailwind content scan
- `.env` line 240 comment (verification pinned to Next 16.2.9)

backryun:
Fixes #12059. Stacked on #12018 (base: `fix/build-honor-env-bundler-flag`) so both
touch `build-next-isolated.mjs` without conflicting — this PR shows its own two
commits once #12018 lands.

## Problem

On macOS arm64 with Next 16.3.2, Turbopack does not finish. Measured on this
machine (M4, 10 cores, 32 GB):

| | Result |
|---|---|
| `next build`, Turbopack, macOS arm64 | **Never completes** — 25 min+, no progress, chunks pinned at 20 |
| `next build`, Turbopack, CI Linux x64 | ~7 min, green |
| `next build`, webpack, macOS arm64 | **13 min**, exit 0 |
| `next dev`, Turbopack | **Never listens** — stuck at `○ Compiling instrumentation Node.js ...`, **zero further output across a 300 s budget** |
| `next dev`, webpack | `[Next] dev server listening on http://0.0.0.0:20128 (webpack)` |

Root cause is upstream: Turbopack delegates the postcss/loader transform to a pool
of short-lived node workers — **44 unique workers in 11 min**, continuous respawn —
and every respawn re-runs the transform. Trace: 2,802,951 spans
(`turbopack_core::resolve` 427,456). See `vercel/next.js#98043` and `#97802` (open).

CI hides this completely: it is green on Linux, so nothing signals to an Apple
silicon contributor that the default is broken.

## Fix

`resolveNextBuildBundlerFlag()` and `run-next.mjs` now share one
`isTurbopackStallPlatform()` helper and pick a **platform-aware default**: webpack
on `darwin`/`arm64`, Turbopack everywhere else. Dev and build agree.

Deliberately narrow — Intel Macs and Linux are untouched, so nothing is penalised
on a guess.

**An operator-set value still wins.** The original env-only design explicitly
guarded against runtime sniffing overriding an explicit request; that guard is
intact. This only fills in the default when nobody asked. Setting
`OMNIROUTE_USE_TURBOPACK=1` on Apple silicon still gets Turbopack.

Decided outcomes (verified by reproduction of the expression):

```
unset   -> webpack     (Apple silicon)
"0"     -> webpack
"1"     -> turbopack   (explicit operator choice respected)
```

## Why `.env.example` changes too

It hardcoded `OMNIROUTE_USE_TURBOPACK=1`, which would override the new platform
default and re-break exactly the machines it protects. It has to be unset for the
default to apply.

This matters more than it looks: **`npm install`'s postinstall regenerates `.env`
from `.env.example`**, so a local `.env` edit does not survive the next install —
but a platform default does. That is precisely how the previous `=0` workaround was
silently lost on this machine.

The stale "no panics on Next 16.2.9" note is kept for history and marked as
predating 16.3.2, since the repo now runs 16.3.2.

## Tests

4 new cases in `tests/unit/build/resolve-next-build-bundler-flag.test.mjs`:

- default is `--webpack` on darwin/arm64
- default is still `--turbopack` on linux
- default is still `--turbopack` on darwin/x64 (Intel Mac)
- an explicit `OMNIROUTE_USE_TURBOPACK=1` still wins on darwin/arm64

Two existing "by default" cases now pin the platform explicitly. They were failing
after this change for a legitimate reason — they asserted *this machine's*
architecture rather than the rule, which is the wrong thing to encode in a test for
a platform-conditional function.

1 new source guard in `tests/unit/run-next-node-env.test.ts`, following that file's
existing convention (it cannot import `run-next.mjs` in-process due to top-level
side effects). It fails if the dev path reverts to unconditional Turbopack.

The `build-next-isolated.mjs` import added to `run-next.mjs` is side-effect free
(12 ms; `main()` is guarded by the entry-script check), which matters on the dev hot
path.

## Not included

The 13-minute webpack build time is unchanged. During this investigation the
following were measured and ruled out as build-time levers:

- NFT trace duplication (1,680 → 2 manifests, but 219 s → 227 s — no gain)
- `experimental.turbotrace` — removed in Next 16.3.2
- onnxruntime (340 MB but only 733 files; copy 1.1 s)
- postbuild I/O (docs + onnxruntime ≈ 2.4 s)

⚠️ base-red inherited: #11874

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -225,19 +225,36 @@ PORT=20128
 # Used by: src/app/api/v1/relay/chat/completions/route.ts
 # RELAY_IP_PER_MINUTE=30
 
-# Bundler selection for `npm run dev`. Set to 0 to fall back to webpack.
-# Default is 1 (Turbopack). PR #4092 had forced webpack because earlier
-# Turbopack 16.2.x panicked on the OmniRoute module graph with "internal error:
-# entered unreachable code: there must be a path to a root"
+# Bundler selection for `npm run dev` and `npm run build`. Set to 0 to force
+# webpack, 1 to force Turbopack.
+#
+# LEFT UNSET ON PURPOSE. When this is unset the build picks a platform-aware
+# default: webpack on Apple silicon, Turbopack everywhere else. Hardcoding 1
+# here overrides that and re-breaks the machines it exists to protect — see
+# below. Note `npm install` regenerates `.env` from this file, so a local
+# edit to `.env` does not survive the next install; the platform default does.
+#
+# Why Apple silicon needs webpack (#12059): on macOS arm64 with Next 16.3.2,
+# Turbopack does not converge. `next build` stalls indefinitely (25 min+ with
+# no progress, chunks pinned at 20) and `next dev` never reaches listen state
+# (stuck at "Compiling instrumentation Node.js"), while the same commits build
+# green in ~7 min in CI on Linux x64. Root cause is upstream: the JS transform
+# pool workers continuously respawn (44 unique workers in 11 min) and every
+# respawn re-runs the build's postcss transform. See vercel/next.js#98043 and
+# #97802 (open). The webpack path completes in ~13 min and is the reliable
+# option there.
+#
+# History: PR #4092 forced webpack because Turbopack 16.2.x panicked with
+# "internal error: entered unreachable code: there must be a path to a root"
 # (turbopack-core/module_graph/mod.rs:662). That panic no longer reproduces on
 # the pinned Next 16.2.9 — verified across a broad cold-compile sweep (36
 # dashboard routes + open-sse-heavy API routes incl. /api/v1/chat/completions,
 # /api/v1/models, /api/mcp) and repeated HMR rebuilds: zero panics. Turbopack
 # also keeps dev memory far lower on the edit→rebuild loop (HMR rebuild RSS stays
 # ~flat vs webpack's monotonic growth), which mitigates the dev-server OOM on
-# this 60+ route app. The production build still uses webpack (build pipeline is
-# unaffected by this dev-only flag).
-OMNIROUTE_USE_TURBOPACK=1
+# this 60+ route app. That verification predates 16.3.2 and does not cover the
+# arm64 stall above.
+#OMNIROUTE_USE_TURBOPACK=1
 
 # Disable systemd sd_notify (Type=notify / WatchdogSec=) even when running
 # under a systemd unit with NOTIFY_SOCKET set.
--- a/scripts/build/build-next-isolated.mjs
+++ b/scripts/build/build-next-isolated.mjs
@@ -132,13 +132,39 @@ function runNextBuild() {
   });
 }
 
-export function resolveNextBuildBundlerFlag(baseEnv = process.env) {
+/**
+ * True where Turbopack is known not to converge: Apple silicon on Next 16.3.2 (#12059).
+ * `next build` never completes (25 min+ with no progress, chunks pinned at 20) and
+ * `next dev` never reaches listen state, while the same commits build green in CI on
+ * Linux x64. Narrow on purpose — Intel Macs are untouched, so this does not penalise
+ * platforms we have not measured.
+ */
+export function isTurbopackStallPlatform(platform = process.platform, arch = process.arch) {
+  return platform === "darwin" && arch === "arm64";
+}
+
+export function resolveNextBuildBundlerFlag(
+  baseEnv = process.env,
+  platform = process.platform,
+  arch = process.arch
+) {
   // Turbopack is the default; OMNIROUTE_USE_TURBOPACK=0 is the documented escape hatch
   // to webpack (Windows, native-binding trouble, RAM-constrained machines — #6409, and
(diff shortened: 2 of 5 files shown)


backryun:

Issue #12074, opened 2026-08-30
fix(backend): dev bundlers exhaust memory on oversized shared import graphs


backryun:
## Summary

Both development bundlers are currently unusable on the v3.8.51 line for different visible reasons:

- Turbopack repeatedly fails while tracing runtime-only/native dependency boundaries.
- The webpack fallback starts, but route compilation grows to roughly 9.7 GiB RSS and becomes progressively slow.
- A captured webpack dev tree reached 3.7 GB on disk: 2.2 GB cache and 1.4 GB server output.
- The process emitted MaxListenersExceededWarning for 13-14 WriteStream listeners while compiling provider routes.

This issue tracks a measured, phased repair. Each phase must land as a separate PR so low-risk containment is not mixed with architectural import-graph work.

## Evidence and suspected graph roots

- src/instrumentation-node.ts imports @omniroute/open-sse/index.ts during every server bootstrap.
- quotaAutoPing imports the complete executor registry, causing every lazy executor loader to be traced into instrumentation chunks.
- src/app/layout.tsx imports the monolithic DB settings module; its write path dynamically reaches runtime settings, token health checks, provider validation, browser-backed executors, and native/runtime require boundaries.
- src/shared/utils/logger.ts constructs a pino transport at module evaluation; the compiled dev tree contains many copies across server entries.
- Tailwind v4 automatic source detection remains enabled even though explicit @source directives exist.
- Custom webpack splitChunks cache groups are applied in development as well as production.
- Warning suppression only hides repeated diagnostics and is not a graph or memory fix.

## Phased plan

- [x] Phase 1 - bound development scanning and optimization
  - Disable Tailwind automatic source detection and explicitly enumerate required source roots.
  - Keep production vendor splitting, but do not impose production-oriented custom cache groups on webpack dev compilation.
  - Add regression contracts and record cold/recompile measurements.

- [x] Phase 2 - split the root settings read path
  - Move layout-safe settings reads into a leaf module.
  - Prevent root layout compilation from tracing settings writes, runtime hot reload, token health, and provider validation.
  - Preserve DB behavior and add focused import-boundary tests.

- [x] Phase 3 - reduce instrumentation and executor graph fan-out
  - Replace open-sse barrel bootstrap imports with leaf imports.
  - Decouple quota auto-ping from the complete executor registry.
  - Keep quota fetcher registration and startup behavior intact.

- [x] Phase 4 - make logging and startup resources process-singleton
[...]

PR #12078, opened 2026-08-30
[URGENT] fix(dev): reduce instrumentation executor fan-out (phase 3)

backryun:
## Summary

Phase 3 of #12074 removes complete executor-registry ownership from bootstrap-only paths.

- load the proxy-fetch side effect directly instead of importing the `open-sse` barrel
- make quota auto-ping lazily construct only the Codex executor it can actually use
- extract credential refresh into a resolver-injected leaf
- add a refresh-only executor resolver covering every executor that overrides refresh behavior
- route Claude OAuth validation through a shared default-executor leaf instead of the complete chat registry
- preserve retirement guards, token-refresh serialization, quota fetcher registration, and shared fallback-executor identity

## Scope boundary

This PR is intentionally independent of Phases 1 and 2 and is based directly on `release/v3.8.51`.

Webpack still emits deferred API-route chunks reachable through the existing batch-dispatch table. Those chunks can contain the chat registry, but the Phase 3 bootstrap modules themselves no longer import or own that registry. This PR does not hide that remaining graph or claim the overall memory problem is finished.

## Validation

- focused instrumentation, quota, credential-refresh, registry, validation, and platform regressions: 46 passed, 0 failed
- `npm run check:cycles` — no cycles across 429 files
- `npm run lint` — passed; only the existing unused-suppression advisory was printed
- isolated `bun scripts/check/check-known-symbols.ts` — 133 executors conformant
- Prettier check for new/changed leaf and regression files — passed
- `git diff --check` — passed

The new contract also fails if a future executor overrides `needsRefresh` or `refreshCredentials` without being added to the refresh-only resolver.

## Isolated webpack smoke

With a clean `.build`, isolated `DATA_DIR`/`SQLITE_FILE`, and `OMNIROUTE_USE_TURBOPACK=0`:

- cold `/login`: HTTP 200
- `_instrument_src_instrumentation-node_ts.js` contains neither `open-sse/index.ts` nor `open-sse/executors/index.ts`
- no standalone `_instrument_open-sse_executors_index_ts.js` chunk was emitted
- observed dev output: 1.8 GB total (887 MB cache, 858 MB server)
- observed main-process RSS: about 6.78 GiB

Cold timing and RSS remain noisy and the deferred batch-route graph still exists, so these measurements are a Phase 3 snapshot rather than a claimed end-to-end improvement.

Refs #12074

## Cross-phase Webpack qualification

A controlled local A/B run stacked Phases 1-4 on the exact shared base. The combined tree reduced the representative dashboard cold compile from 23.21 s to 7.76 s and settled post-HMR RSS from 15.94 GiB to 14.06 GiB, but one layout HMR still retained roughly 6.4 GiB. The overall Webpack acceptance criteria are therefore not met yet.

This cross-phase result does not attribute the aggregate improvement to this standalone PR or expand its scope. Full measurements and the remaining graph evidence are recorded in [#12074](https://github.com/diegosouzapw/OmniRoute/issues/12074#issuecomment-5466049538).

### DIFF
--- a/open-sse/executors/credential.ts
+++ b/open-sse/executors/credential.ts
@@ -1,0 +1,54 @@
+import { assertCommonChatGptWebProviderAvailable } from "@/shared/constants/chatgptWebRetirement";
+import { assertMicrosoftDesignerWebProviderAvailable } from "@/shared/constants/designerWebRetirement";
+import { assertRuntimeProviderAvailable } from "@/shared/constants/providerRetirement";
+import type { BaseExecutor } from "./base.ts";
+import { getDefaultExecutor } from "./defaultResolver.ts";
+
+type CredentialExecutorLoader = () => Promise<BaseExecutor>;
+
+const specializedCredentialExecutors: Record<string, CredentialExecutorLoader> = {
+  antigravity: () => import("./antigravity.ts").then((m) => new m.AntigravityExecutor()),
+  agy: () => import("./antigravity.ts").then((m) => new m.AntigravityExecutor()),
+  github: () => import("./github.ts").then((m) => new m.GithubExecutor()),
+  "ghe-copilot": () => import("./ghe-copilot.ts").then((m) => new m.GheCopilotExecutor()),
+  kiro: () => import("./kiro.ts").then((m) => new m.KiroExecutor()),
+  "amazon-q": () => import("./kiro.ts").then((m) => new m.KiroExecutor("amazon-q")),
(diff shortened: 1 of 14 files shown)


diegosouzapw:
Marcando para o ciclo v3.8.52. Phase 3 do épico #12074 — refatora a resolução do registro de executors e o caminho de credential refresh usado por todos os providers (384+/167-). Blast radius grande demais para lote; merece sessão dedicada junto das fases 2, 4 e 4b.

PR #12079, opened 2026-08-30
[URGENT] fix(dev): make logging lifecycle HMR-safe (phase 4)

backryun:
## Summary

Phase 4 of #12074 makes logger and shutdown resources process-wide so Next development HMR cannot multiply Pino worker transports, log-rotation timers, or signal listeners.

## Changes

- keep the shared Pino logger resource on `globalThis` and reuse it across server chunk/module re-evaluation
- make the log-rotation interval process-singleton and keep teardown idempotent
- expose one process-wide graceful-shutdown promise so repeat callers join the same cleanup
- let the custom Next runner own process exit and await application cleanup before closing Next
- close the owned logger transport after the remaining application resources, without closing stdout
- add HMR-style repeat-import, listener ownership, runner ordering, and teardown regressions

## Validation

- focused logger/rotation/shutdown suite: 17 passed
- `npm run test:vitest`: 50 files, 463 tests passed
- `npm run lint`: passed; only the existing stale-suppressions notice was emitted
- webpack dev smoke on isolated port/DB: three logger-triggered HMR recompiles completed with no `MaxListenersExceededWarning`
- live SIGINT teardown completed in order: application drain, DB close, logger transport + rotation close, `Bye`, then Next close

## Full unit-suite note

`npm run test:unit` did not complete because two untouched Redis warmup-scheduler tests stalled. A bounded focused rerun confirmed both time out at 15 seconds:

- `tests/unit/lib/warmupScheduler/circuitBreakerFactory.test.ts`
- `tests/unit/lib/warmupScheduler/circuitBreakerFactoryRelease.test.ts`

Those files are unchanged by this PR; all Phase 4 and adjacent logger/shutdown tests pass.

Part of #12074.

## Cross-phase Webpack qualification

A controlled local A/B run stacked Phases 1-4 on the exact shared base. The combined tree reduced the representative dashboard cold compile from 23.21 s to 7.76 s and settled post-HMR RSS from 15.94 GiB to 14.06 GiB, but one layout HMR still retained roughly 6.4 GiB. The overall Webpack acceptance criteria are therefore not met yet.

This cross-phase result does not attribute the aggregate improvement to this standalone PR or expand its scope. Full measurements and the remaining graph evidence are recorded in [#12074](https://github.com/diegosouzapw/OmniRoute/issues/12074#issuecomment-5466049538).

### DIFF
--- a/scripts/dev/run-next.mjs
+++ b/scripts/dev/run-next.mjs
@@ -137,6 +137,11 @@ function createNextApp() {
   });
 }
 
+// The custom HTTP server owns process exit. Application instrumentation still
+// registers its cleanup function, but must not install a competing signal
+// listener that can race this runner's async server/Next teardown.
+globalThis.__omnirouteCustomServerOwnsShutdown = true;
+
 let nextApp = createNextApp();
 
 // Best-effort self-heal for a corrupted Turbopack persistent dev cache (#6289):
@@ -231,6 +236,7 @@ async function start() {
     systemdNotifier.stopping();
     try {
       await new Promise((resolve) => server.close(resolve));
+      await globalThis.__omnirouteRequestShutdown?.(signal);
       await nextApp.close();
     } catch (error) {
       console.error("[SHUTDOWN] Failed during signal:", signal, error);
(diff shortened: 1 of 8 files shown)


diegosouzapw:
Marcando para o ciclo v3.8.52. Phase 4 do épico #12074 — mexe em process lifecycle (SIGINT, shutdown, logger singleton via globalThis), 435+/54-. O próprio corpo da PR reconhece que o critério de aceitação agregado das fases ainda não foi atingido. Infra crítica demais para lote; revisar junto das fases 2, 3 e 4b.

PR #12081, opened 2026-08-30
[URGENT] fix(dev): isolate batch dispatch from instrumentation (phase 4b)

backryun:
## Summary

Phase 4b of #12074 removes the deferred batch API-route fan-out that remained in the
instrumentation graph after Phase 3. Batch execution still traverses the real Next API route,
but it now does so at runtime through the active loopback dashboard listener instead of
compile-time dynamic imports of seven route modules.

## Changes

- replace the seven `@/app/api/v1/**/route` dynamic imports in batch dispatch with one
  loopback HTTP POST
- pin the target to `127.0.0.1` and the active `dashboardPort`; do not derive it from request
  headers or a public base URL
- honor `OMNIROUTE_BASE_PATH` for subpath deployments
- reject redirects while carrying the stored batch API key
- add an import-boundary regression and direct assertions for URL, method, auth, body, and
  redirect behavior
- update the affected batch tests to mock the new HTTP boundary explicitly

## Validation

- TDD regression: the new boundary test failed before the production change on the seven
  route imports, then passed after the change
- focused batch suite: 37 tests passed
- Webpack real-server batch E2E: upload -> create batch -> loopback dispatch -> fake provider
  relay -> result file; 2/2 items completed, 0 failed
- changed-file ESLint with repository suppressions and no cache: passed
- Prettier and `git diff --check`: passed

## Controlled Webpack graph comparison

I stacked this commit on the existing Phases 1-4 integration branch and compared it with the
same branch without Phase 4b. Both runs used a fresh Next cache and DB, the same Node command,
Webpack mode, port, request order, and sequential execution.

| Measurement | Phases 1-4 | Phases 1-4b |
| --- | ---: | ---: |
| Cold `/dashboard/providers` | 19.04 s | 12.74 s |
| Cold provider-model probe | 3.24 s | 3.41 s |
| RSS after the cold routes | 7.32 GiB | 7.38 GiB |
| Next dev output | 1,922,500 KiB | 1,534,152 KiB |
| Instrumentation-prefixed chunks | 243 | 160 |
| Target batch-route references | 8 | 0 |
| Target batch-route chunk total | 154,296 KiB | 0 KiB |
| RSS after two invalidations and 15 s settle | 7.85 GiB | 7.59 GiB |

The reproducible result is the graph/output reduction: about 379 MiB less dev output and no
batch-route modules under the instrumentation-prefixed graph. The provider-model timing is
effectively noise, and this repeat did not reproduce the earlier 14 GiB layout-HMR spike, so
this PR does **not** claim to close the overall memory issue or replace the remaining Phase 5
Turbopack investigation.

Part of #12074.

### DIFF
--- a/src/lib/batches/dispatch.ts
+++ b/src/lib/batches/dispatch.ts
@@ -1,29 +1,6 @@
 import type { SupportedBatchEndpoint } from "@/shared/constants/batchEndpoints";
-
-type BatchRouteHandler = (request: Request) => Promise<Response> | Response;
-
-const handlerLoaders: Record<SupportedBatchEndpoint, () => Promise<BatchRouteHandler>> = {
-  "/v1/responses": async () => (await import("@/app/api/v1/responses/route")).POST,
-  "/v1/chat/completions": async () => (await import("@/app/api/v1/chat/completions/route")).POST,
-  "/v1/embeddings": async () => (await import("@/app/api/v1/embeddings/route")).POST,
-  "/v1/completions": async () => (await import("@/app/api/v1/completions/route")).POST,
-  "/v1/moderations": async () => (await import("@/app/api/v1/moderations/route")).POST,
-  "/v1/images/generations": async () =>
-    (await import("@/app/api/v1/images/generations/route")).POST,
-  "/v1/videos/generations": async () =>
-    (await import("@/app/api/v1/videos/generations/route")).POST,
-};
-
-const handlerCache = new Map<SupportedBatchEndpoint, BatchRouteHandler>();
-
-async function getHandler(endpoint: SupportedBatchEndpoint): Promise<BatchRouteHandler> {
-  const cached = handlerCache.get(endpoint);
-  if (cached) return cached;
-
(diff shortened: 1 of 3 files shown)


backryun:
CI triage update: the two remaining red checks here are inherited release contract drift, not failures caused by Phase 4b. Fast Quality Gates reports the missing `lkgp-stale-pin-exhaustion-11911.test.ts` Stryker registration, and Unit Tests 3/4 reports the stale pre-#12033 assertion that node subcommands must not expose the restored legacy `--base-url` alias. Both are repaired separately in #12082 so this PR stays scoped to the batch-dispatch import boundary. All other Phase 4b checks passed. I will refresh this branch onto the repaired release head after #12082 merges.

diegosouzapw:
Marcando para o ciclo v3.8.52. Phase 4b do épico #12074 — stackada sobre as fases 1-4, então depende delas mergeadas primeiro. Troca imports diretos por um round-trip HTTP loopback no batch dispatch, mudança estrutural que merece revisão junto das fases 2, 3 e 4.

Issue #10285, opened 2026-08-13
bug(googleflow): video generation always 404s - wrong submit path, and the bearer credential is rejected by the working endpoint (v3.8.49)


JoaoLeiteMelk:
### Summary

`POST /v1/videos/generations` with any `googleflow/*` model always fails:

```json
{"error":{"message":"Google Flow submit failed: <!DOCTYPE html>","type":"upstream_error","code":"upstream_error"}}
```

HTTP 404. Dispatch is fine — the request is built, a credential is selected, the handler logs the
submit — but the upstream answers with an HTML 404 page instead of JSON.

I traced it to **two independent problems**: the submit path does not exist, and the bearer
credential is rejected by the endpoint that does. Both are measured below. This looks like the
wire surface flagged as PENDING LIVE VALIDATION in #4769.

### Environment

| | |
|---|---|
| Version | 3.8.49 (`diegosouzapw/omniroute:3.8.49`) |
| Deployment | self-hosted Docker, single host, behind a reverse proxy |
| Credential | Google OAuth via the Antigravity / Cloud Code connection |
| Flow access | **Yes** — the account signs in to `labs.google/flow` and generates video there |

### Reproduction

```bash
curl -X POST https://<host>/v1/videos/generations \
  -H "Authorization: Bearer <api key with manage scope>" \
  -H 'Content-Type: application/json' \
  -d '{"model":"googleflow/veo-3.1-fast","prompt":"a blue cube rotating slowly","aspect_ratio":"16:9"}'
```

Server log confirms dispatch:

```
{"tag":"VIDEO","msg":"googleflow/veo-3.1-fast (veo) | submitting | aspect: 16:9"}
{"tag":"AUTH","msg":"antigravity selected account=<redacted> eligible=1 excluded=0"}
```

---

## Problem 1 — the submit path does not exist

I captured a **live, successful** Flow video generation using a third-party tool
([`gflow-cli`](https://github.com/ffroliva/gflow-cli), which drives the Flow web app). The call
that actually works is:

```
POST https://aisandbox-pa.googleapis.com/v1/video:batchAsyncGenerateVideoText  → 200
```

OmniRoute posts to `GOOGLE_FLOW_SUBMIT_PATH` = `/v1:generateVideo` on the same host.

Note the shape difference: the working call is `v1/video:batchAsyncGenerateVideoText` — a
**batch, async** method under a `video` collection — not `v1:generateVideo`. Completion is then
[...]

PR #12131, opened 2026-08-30
fix(sse): stop advertising video providers the dispatcher cannot run

quiterunner-commits:
⚠️ base-red inherited: #12109

## Problem

`GET /v1/videos/generations` advertises every `VIDEO_PROVIDERS` entry that is not flagged
`unsupported`, while `handleVideoGeneration` dispatches on `providerConfig.format` and answers
`400 Unsupported video format` for anything it has no branch (or job preset) for. Five providers
sat on the wrong side of that gap, so the catalog listed models that could never run — regardless
of keys or balance.

Live check against a local gateway (2026-08-30), models taken straight from the catalog response:

| Model | Result |
| --- | --- |
| `minimax/MiniMax-Hailuo-02` | 400 `Unsupported video format: minimax-video` |
| `pollinations/default` | 400 `Unsupported video format: pollinations-video` |
| `nanogpt/default` | 400 `Unsupported video format: openai` |

## Changes

**1. Stop advertising providers the dispatcher cannot run.** `pollinations`, `minimax`,
`together` and `replicate` are marked `unsupported` with a reason, reusing the mechanism added in
#10285. Their transports can be implemented later; until then the catalog stays honest.

**2. Fix the nanogpt format name.** It declared `"openai"` while the dispatcher knows the format
as `"openai-video"` — a name mismatch, not a missing transport.

**3. Honor the registry endpoint fallback.** Fixing (2) exposed a live 500 with an empty body:

```
TypeError: Cannot read properties of null (reading 'endsWith')
    at resolveVideoEndpoint (open-sse/handlers/videoGeneration/openai.ts:39)
```

`resolveVideoEndpoint(credentials, fallback)` takes a fallback — the caller passes
`providerConfig.baseUrl` — and never used it. A built-in provider stores no baseUrl on the node,
so the value was `null`. The fallback is a complete endpoint, not a node root, so it is returned
as-is. This also affects any custom OpenAI-compatible node whose credentials carry no baseUrl.

**4. De-list nanogpt video.** Both `/api/v1/video/generations` and `/api/v1/videos/generations`
answer 404 (an HTML page) upstream, while `/api/v1/images/generations` on the same probe key
answers 401 JSON — so the 404 means the route does not exist, not that the key is wrong. The
format name stays corrected so a single baseUrl edit revives the provider later.

## Validation

TDD per Hard Rule #18 — each fix has a test that failed first.

- `tests/unit/video-registry-dispatch-parity.test.ts` — derives the dispatchable format set from
  the dispatcher itself (branches + job presets) and fails when the registry advertises a format
  outside it. Carries its own guard: if the branch scan finds fewer than 10 formats it reports
  that the parser needs updating instead of passing on an empty set.
- `tests/unit/video-openai-endpoint-fallback.test.ts` — reproduces the null-baseUrl TypeError and
  pins that the node's own baseUrl still wins when present.

Suites run: both new files, plus `veoaifree-video-route`, `nanogpt-endpoint-surface`,
`google-flow-video-4569`, `video-deepinfra-6653`, `adobe-firefly`, `agnes-provider`,
`segmind-image-video-provider-6656`, `qwen-cloud-video-media`, `new-content-providers`,
`media-provider-ordering`, `preserve-video-url-compat`, `video-bridge-broker` — all green.
tsc/eslint/prettier clean.

**Live verification** on a scratch instance built from this branch (separate port, copy of a
populated DB): `GET /v1/videos/generations` went from **9** advertised models to **4**, and the
four survivors are all dispatchable veo/seedance entries.

### DIFF
--- a/open-sse/config/videoRegistry.ts
+++ b/open-sse/config/videoRegistry.ts
@@ -190,6 +190,13 @@ export const VIDEO_PROVIDERS: Record<string, VideoProvider> = {
     authType: "apikey",
     authHeader: "bearer",
     format: "pollinations-video",
+    // Живая проверка 2026-08-30: диспетчер videoGeneration.ts не разбирает
+    // "pollinations-video" и отвечает 400 Unsupported video format — модель
+    // висела в выдаче каталога, но не исполнялась ни при каких ключах.
+    unsupported: true,
+    unsupportedReason:
+      "Pollinations video has no submit/poll transport in the dispatcher yet. " +
+      "Use an image model or another video provider until one is added.",
     models: [{ id: "default", name: "Pollinations Video (Free)" }],
   },
 
@@ -200,6 +207,12 @@ export const VIDEO_PROVIDERS: Record<string, VideoProvider> = {
     authType: "apikey",
     authHeader: "bearer",
     format: "minimax-video",
+    // Живая проверка 2026-08-30: 400 Unsupported video format на всех трёх
+    // моделях Hailuo. Свой submit → query API, не покрытый job-пресетами.
+    unsupported: true,
+    unsupportedReason:
+      "MiniMax video uses its own submit/query transport that the dispatcher " +
+      "does not implement yet. Generate video via another provider for now.",
     models: [
       { id: "MiniMax-Hailuo-2.3", name: "Hailuo 2.3" },
       { id: "MiniMax-Hailuo-02", name: "Hailuo 02" },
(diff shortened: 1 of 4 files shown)


quiterunner-commits:
CI has not started on this PR — workflows from a first-time contributor's fork wait for a maintainer to approve the run. Posting the local results so the change is not sitting here unverified.

Rebased on `origin/release/v3.8.51` (cda832c3a at the time of writing).

**Tests** — `node --import tsx/esm --test`, all green:

| File | pass | fail |
| --- | --- | --- |
| `tests/unit/video-registry-dispatch-parity.test.ts` (new) | 3 | 0 |
| `tests/unit/video-openai-endpoint-fallback.test.ts` (new) | 2 | 0 |
| `tests/unit/veoaifree-video-route.test.ts` | 2 | 0 |
| `tests/unit/nanogpt-endpoint-surface.test.ts` | 14 | 0 |

Also run earlier on the same branch: `google-flow-video-4569` (23), `video-deepinfra-6653` (10), `adobe-firefly` (46), `agnes-provider` (11), `segmind-image-video-provider-6656` (11), `qwen-cloud-video-media` (9), `new-content-providers` (16), `media-provider-ordering` (2), `preserve-video-url-compat` (6), `video-bridge-broker` (8) — all green.

**Lint / format** — `eslint` and `prettier --check` clean on every file this PR touches.

One pre-existing eslint error lives in `tests/unit/video-custom-provider-route.test.ts` (`'init' is defined but never used`), untouched by this branch — likely part of the inherited base-red #12109.

Both new tests fail before the fix and pass after it, per Hard Rule #18. Happy to rebase or split if the maintainers prefer a different shape.

PR #12143, opened 2026-08-30
fix(sse): refuse an AI Horde queue that cannot fit the request budget

quiterunner-commits:
⚠️ base-red inherited: #12109

## Problem

`/v2/generate/check` reports `wait_time`, `queue_position` and `eligible_workers` in its very
first response. The poller ignored all three: it slept one second, asked again, and kept going
until the 600s budget ran out, then failed with a bare timeout. A user asking for a model with
few workers waited ten minutes to learn nothing.

Measured live on 2026-08-30, same code path, same anonymous key, one minute apart:

| Model | Workers | First check said | Outcome |
| --- | --- | --- | --- |
| `aihorde/Deliberate` | 3 | `wait_time: 1478`, `queue_position: 321` | timed out after the full budget |
| `aihorde/stable_diffusion` | 10 | `wait_time: ~10` | **image returned in 10.4s** |

25 minutes of queue against a 10 minute budget — unreachable before the first poll was even sent.
The information needed to say so arrived in the first second.

## Changes

**Fail fast when the queue cannot fit the budget.** A 504 naming the queue, the position, the
eligible worker count and the budget left, so the caller can pick a busier model instead of
raising a timeout that would not have helped. The check only fires when the *reported* wait
exceeds the *remaining* budget, so a short queue is unaffected.

**Back the poll interval off.** A tenth of the reported wait, clamped to 1s..8s, instead of a
flat second. A full-budget wait used to cost Horde ~600 requests on the shared anonymous key for
a single image; a short queue still polls every second and returns just as fast.

## Validation

TDD per Hard Rule #18 — the test failed first.

`tests/unit/aihorde-queue-budget.test.ts` stubs a Horde queue and asserts two things: a wait that
does not fit the budget is refused after **one** check (the test would previously hang for the
full 600s), and a wait that does fit keeps polling rather than being refused early. The refusal
path now completes in ~1s.

Suites run: the new file plus `aihorde-image-generation`, `aihorde-image-catalog`,
`aihorde-optional-api-key`, `aihorde-key-validation`,
`aihorde-tools-unsupported-provider-fallback` — all green. tsc/eslint/prettier clean on the
changed files.

### DIFF
--- a/open-sse/handlers/imageGeneration/providers/aihorde.ts
+++ b/open-sse/handlers/imageGeneration/providers/aihorde.ts
@@ -17,7 +17,11 @@ import {
 } from "./aihordeMapRequest.ts";
 
 const GENERATE_TIMEOUT_MS = 600_000;
-const POLL_INTERVAL_MS = 1_000;
+// Опрос начинается частым и разряжается по мере ожидания: короткая очередь
+// отдаёт картинку за секунды, а длинная иначе стоила бы Horde сотен запросов
+// по общему анонимному ключу (600 опросов на один кадр при полном бюджете).
+const POLL_INTERVAL_MIN_MS = 1_000;
+const POLL_INTERVAL_MAX_MS = 8_000;
 // Per-call bound for the Horde API's own submit/check/status/cancel calls
 // (a fixed, trusted host — no SSRF guard needed, just a hard timeout so a
 // hung upstream cannot stall a request indefinitely). Individual calls are
@@ -119,6 +123,11 @@ async function fetchHordeImageBytes(
   return value;
 }
 
+/** Числовое поле ответа Horde: отсутствующее или нечисловое читается как «неизвестно». */
+function numericField(value: unknown): number | null {
+  return typeof value === "number" && Number.isFinite(value) ? value : null;
+}
+
 export async function handleAiHordeImageGeneration({
   model,
   provider,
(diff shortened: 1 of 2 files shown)


quiterunner-commits:
CI has not started on this PR — workflows from a first-time contributor's fork wait for a maintainer to approve the run. Posting the local results so the change is not sitting here unverified.

Rebased on `origin/release/v3.8.51` (cda832c3a at the time of writing).

**Tests** — `node --import tsx/esm --test`, all green:

| File | pass | fail |
| --- | --- | --- |
| `tests/unit/aihorde-queue-budget.test.ts` (new) | 2 | 0 |
| `tests/unit/aihorde-image-generation.test.ts` | 7 | 0 |
| `tests/unit/aihorde-image-catalog.test.ts` | 5 | 0 |
| `tests/unit/aihorde-optional-api-key.test.ts` | 7 | 0 |
| `tests/unit/aihorde-key-validation.test.ts` | 5 | 0 |
| `tests/unit/aihorde-tools-unsupported-provider-fallback.test.ts` | 5 | 0 |

**Lint / format** — `eslint` and `prettier --check` clean on every file this PR touches.

The new test is worth a note on timing: before the fix its first case hangs for the full 600s budget, which is exactly the behaviour being fixed. After the fix the refusal path completes in ~1s, so the suite stays fast.

The measurement in the description was taken live against a running gateway, one minute apart on the same anonymous key — `Deliberate` (3 workers) reported `wait_time: 1478` and timed out; `stable_diffusion` (10 workers) returned an image in 10.4s through the same code path.

PR #12177, opened 2026-08-31
fix(combos): send null to clear an agent feature instead of omitting it

### LINKED ISSUE #12158: [BUG] Combos: unchecking "context cache protection" never persists — UI deletes the field from the update payload instead of sending false (same for clearing system_message / tool_filter_regex)
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux (Alibaba Cloud ECS, CentOS 8)

### Provider(s) Involved

N/A — dashboard/UI bug, not provider-specific.

### Client Tool

Dashboard (combos editor)

### Description

Follow-up to #3398 (v3.8.16 fixed the server-side column read/write for `context_cache_protection`, which works — thank you). However the **dashboard save path still cannot turn the toggle OFF**, and two sibling fields have the same defect.

**Root cause:** in `src/app/(dashboard)/dashboard/combos/page.tsx` (save handler, "Agent features (#399 / #401 / #454)" block), unchecking / clearing a field **deletes it from the payload** instead of sending the cleared value:

```ts
if (agentSystemMessage.trim()) saveData.system_message = agentSystemMessage.trim();
else delete saveData.system_message;
if (agentToolFilter.trim()) saveData.tool_filter_regex = agentToolFilter.trim();
else delete saveData.tool_filter_regex;
if (agentContextCache) saveData.context_cache_protection = true;
else delete saveData.context_cache_protection;
```

The `PUT /api/combos/[id]` update schema declares these fields as optional
(`src/shared/validation/schemas/combo.ts`):

```ts
system_message: z.string().max(50000).optional(),
tool_filter_regex: z.string().max(1000).optional(),
context_cache_protection: z.boolean().optional(),
```

so an **absent field means "leave unchanged"**, and `sqliteComboRepository.updateCombo()` merges over the existing record — the old `true` survives. Verified against `storage.sqlite` on 3.8.50: uncheck → save → reopen shows the toggle checked again, and the `combos.context_cache_protection` column is still `1` together with `"context_cache_protection": true` inside the `data` JSON blob.

Notably, the **same save handler already implements the correct pattern** for `context_length` a few lines below (explicit clear):

```ts
} else if (isEdit) {
  // Editing: send null to explicitly clear context_length
  saveData.context_length = null;
}
```

### Affected fields (all in the combos editor "Agent features" section)

1. **`context_cache_protection`** — unchecking the toggle and saving has **no effect**; it re-checks itself on reload. There is no way to disable context-cache pinning from the dashboard.
2. **`system_message`** — clearing the input and saving does not clear the stored value; the old system message keeps being applied.
3. **`tool_filter_regex`** — same: cannot be cleared, only replaced.

### Steps to Reproduce

1. Dashboard → Combos → edit a combo with "context cache protection" checked
2. Uncheck it → Save
3. Reopen the editor → the checkbox is checked again
4. (Same for clearing `system_message` / `tool_filter_regex`)

### Expected behavior

Saving an unchecked toggle / cleared input should persist the cleared state (send `false` / `""`), matching the `context_length` explicit-clear pattern.

### Suggested fix

```ts
if (agentContextCache) saveData.context_cache_protection = true;
else saveData.context_cache_protection = false;      // schema already accepts boolean

saveData.system_message = agentSystemMessage.trim(); // send "" to clear (or make the update schema .nullable() and send null, mirroring context_length)
saveData.tool_filter_regex = agentToolFilter.trim();
```

(Optionally extend the update schema with `.nullable()` for `system_message` / `tool_filter_regex` and send `null` on clear, exactly like `context_length` — that keeps "empty means cleared" unambiguous.)

### Workaround

Directly updating `storage.sqlite` (both the `context_cache_protection` column and the `data` JSON blob) restores the intended state, confirming the backend persistence works fine once the value actually reaches it.

### Environment

- omniroute 3.8.50, Docker image `diegosouzapw/omniroute:latest`

ntdatt812:
Fixes #12158. The report is exactly right, including the pointer at `context_length` as the pattern already in the same function.

## Root cause

`updateCombo` merges the PUT body over the stored record, and deletes only the keys explicitly set to `null`:

```ts
// src/lib/db/repositories/sqliteComboRepository.ts:233
// Remove fields explicitly set to null (for deletion support)
for (const key of Object.keys(data)) {
  if (data[key] === null) delete merged[key];
}
```

So an omitted field means "leave unchanged". Dropping a cleared field from the payload preserved the old value — and for `context_cache_protection` the column write is derived from the merged record (`normalizedMerged.context_cache_protection ? 1 : 0`), so the `1` survived too.

`description` (line 394) and `context_length` (line 410) were already `.optional().nullable()` in `updateComboSchema` and already cleared with an explicit `null`. This just brings the three agent fields onto the same footing:

- `updateComboSchema`: the three fields become `.optional().nullable()`. The create schema is untouched — there is nothing to clear on create.
- The editor sends `null` when a field is cleared **and** `isEdit` is true, and still omits it on create.

The `superRefine` "No valid fields to update" guard tests `=== undefined`, so a `null` still counts as a field to update. There is a test for that.

## One structural change

I moved the clearing logic out of `page.tsx` into `comboAgentFeatures.ts` beside it, so it can be unit-tested — the same shape as `comboQuotaOnlyFallback.ts`, which is imported by the same save handler and tested in `tests/unit/ui/combo-quota-exhaustion-option.test.tsx`. `page.tsx` is 208 KB and this logic is the whole bug, so pinning it directly seemed better than leaving the fix untestable.

## Tests

`tests/unit/combo-clear-agent-features-12158.test.ts`, 9 tests across the three layers:

- **schema** — `null` accepted for all three; `null` still counts as an update while `{}` is still rejected; set values still accepted and a bad type still rejected
- **repository** — `null` clears each field through `updateCombo`, and the re-read (what the editor reopens with) no longer reports them; omitting a field still leaves it unchanged
- **editor** — cleared → `null` on edit, cleared → absent on create, set values still trimmed and sent, and clearing one field does not disturb the others

Reverting each half proves they hold it:

| mutation | result |
|---|---|
| drop `.nullable()` from the three schema fields | 2 fail |
| drop the three `else if (isEdit) … = null` branches | 2 fail |

## Commands run

```
node --import tsx/esm --test tests/unit/combo-clear-agent-features-12158.test.ts     9 pass, 0 fail
eslint <4 changed files> --suppressions-location config/quality/eslint-suppressions.json
                                                                                    No issues found
tsc --noEmit -p tsconfig.json                                                        no new errors
```

On `tsc`: the repo currently reports 4781 errors on `main`. This branch also reports 4781, and the per-file counts are identical — `page.tsx` has the same 23 before and after, and the two new files have none. (A naive diff of that output looks alarming because adding an import shifts every subsequent line number in `page.tsx`.)

Changed test files: `tests/unit/combo-clear-agent-features-12158.test.ts` (new).

I have not run the dashboard to click through the toggle; the editor half is proven by the extracted helper's tests rather than in a browser.

### DIFF
--- a/changelog.d/fixes/12176-combo-clear-agent-features.md
+++ b/changelog.d/fixes/12176-combo-clear-agent-features.md
@@ -1,0 +1,4 @@
+- **fix(combos):** clearing an agent feature in the combos editor now persists — unchecking
+  context cache protection, or emptying the system message or tool filter, sends an explicit
+  `null` instead of dropping the field from the `PUT` body, which the update merge read as
+  "leave unchanged" ([#12176](https://github.com/diegosouzapw/OmniRoute/pull/12176)) — thanks @foreveryh
--- a/changelog.d/fixes/12177-combo-clear-agent-features.md
+++ b/changelog.d/fixes/12177-combo-clear-agent-features.md
@@ -1,4 +1,4 @@
 - **fix(combos):** clearing an agent feature in the combos editor now persists — unchecking
   context cache protection, or emptying the system message or tool filter, sends an explicit
   `null` instead of dropping the field from the `PUT` body, which the update merge read as
-  "leave unchanged" ([#12176](https://github.com/diegosouzapw/OmniRoute/pull/12176)) — thanks @foreveryh
+  "leave unchanged" ([#12177](https://github.com/diegosouzapw/OmniRoute/pull/12177)) — thanks @foreveryh
--- a/src/app/(dashboard)/dashboard/combos/comboAgentFeatures.ts
+++ b/src/app/(dashboard)/dashboard/combos/comboAgentFeatures.ts
@@ -1,0 +1,46 @@
+/**
+ * Agent-features clearing for the combos editor (#399 / #401 / #454, fixed in #12158).
+ *
+ * `PUT /api/combos/[id]` merges its body over the stored record, so an omitted field
+ * means "leave unchanged". Deleting a cleared field from the payload therefore left the
+ * previous value in the database: unchecking context cache protection, or emptying the
+ * system message or tool filter, never persisted. Only an explicit `null` reaches
+ * `updateCombo`'s null-means-delete pass.
+ *
+ * On create there is nothing to clear, so an empty field is simply absent — the same
+ * shape `description` and `context_length` already use in this editor.
+ */
+export interface AgentFeatureInput {
+  systemMessage: string;
+  toolFilter: string;
+  contextCache: boolean;
+  isEdit: boolean;
+}
+
+export interface AgentFeaturePatch {
+  system_message?: string | null;
+  tool_filter_regex?: string | null;
+  context_cache_protection?: true | null;
+}
+
+export function buildAgentFeaturePatch({
+  systemMessage,
+  toolFilter,
+  contextCache,
+  isEdit,
+}: AgentFeatureInput): AgentFeaturePatch {
+  const patch: AgentFeaturePatch = {};
+
+  const message = systemMessage.trim();
+  if (message) patch.system_message = message;
+  else if (isEdit) patch.system_message = null;
+
+  const filter = toolFilter.trim();
+  if (filter) patch.tool_filter_regex = filter;
+  else if (isEdit) patch.tool_filter_regex = null;
+
+  if (contextCache) patch.context_cache_protection = true;
+  else if (isEdit) patch.context_cache_protection = null;
+
+  return patch;
+}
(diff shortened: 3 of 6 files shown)


diegosouzapw:
Retargeted from `main` to `release/v3.8.51` (the active development branch). Heads-up on why: `main` only receives work at the release squash, so it currently carries an inherited base-red (#12133) whose failing checks — unit-test/pack-artifact gates killed on timeout — have nothing to do with your diff. Against the release branch your CI runs clean. No action needed from you; thanks for the fix!

ntdatt812:
The red Build here is not a build failure:

```
##[error]The runner has received a shutdown signal. This can happen when the runner
service is stopped, or a manually started runner is canceled.
##[error]The operation was canceled.
```

Nothing compiled and nothing failed. #12180 caught the same thing in the same window, and it is not specific to these branches — seven of the last twelve CI runs on `main` itself are `cancelled`.

Every other check on this PR is green, and `tests/unit/combo-clear-agent-features-12158.test.ts` is 9 passed locally on the rebased branch. Only flagging it so the red does not read as a code problem.

PR #12180, opened 2026-08-31
fix(memory): measure the embedding width instead of waiting for a probe

### LINKED ISSUE #12154: fix(backend): vec_memories never created for self-hosted embedding endpoints — memories stored but never vectorized, health stays green
## Summary

`vec_memories` is never created for a **self-hosted embedding endpoint**, so
memories are stored but never vectorized. Semantic and hybrid retrieval
silently degrade to FTS5 while `/api/memory/health` keeps reporting
`working: true`.

This is the same failure surface as #8074, but on a path that fix does not
cover: #8220 populates dimensions **from the registry**, and a self-hosted
endpoint is by definition absent from it.

Measured on v3.8.50 (`ghcr.io/diegosouzapw/omniroute:3.8.50`), the current
release, against a self-hosted [Text Embeddings
Inference](https://github.com/huggingface/text-embeddings-inference) server
running `Qwen/Qwen3-Embedding-0.6B`.

## Impact

Three instances, same configuration. The memory engine card is green on all
three; none of them vectorizes anything:

| Instance | Memories stored | Vectorized (`rowCount`) | `needsReindex` |
| --- | --- | --- | --- |
| A | 24 | **0** | 24 |
| B | 24 | **0** | 24 |
| C (older, see below) | 30 | 38 | 0 |

Instance C only works because its `memory_vec_meta.embedding_signature` was
written on 2026-08-24 by an earlier build. Its prefix does not resolve on
3.8.50 either — a reset would put it in the same state as A and B.

Nothing surfaces the failure: `POST /api/memory` returns 200,
`GET /api/memory/health` returns `working: true`, and the engine card shows
the store as enabled.

## Root cause

`resolveEmbeddingSource()` returns `dimensions: null` on every path that does
not hit the hard-coded registry, and `vectorStore.ensureReady()` refuses to
create the virtual table without a dimension:

```
[WARN] [MEMORY_REINDEX] memory.reindex.ensure_ready.skipped
  {"reason":"no dimensions available yet (lazy probe pending)","pending":24}
[WARN] [MEMORY_STORE] memory.vec.upsert.fail
  {"error":"no such table: vec_memories"}
```

Two independent paths reach it, both on current `main`:

**1. Provider node** — `buildDynamicEmbeddingProvider()`
(`open-sse/config/embeddingRegistry.ts:51`) returns `models: []`, so
`getEmbeddingDimension()` finds nothing for a node-backed model.

**2. Custom endpoint** — `customRemoteResolution()`
(`src/lib/memory/embedding/index.ts:79`) hard-codes `dimensions: null` with
the comment *"custom remote provider configured (dim=unknown, **will probe at
embed time**)"*.

That probe does not appear to exist. The dimension **is** measured on every
call — `remote.ts:95` returns `dimensions: vector.length` — but the value is
never fed back into `resolveEmbeddingSource()` or `ensureReady()`. Confirmed
by grep on `main`: no `embeddingResult.dimensions` consumer anywhere.

So the embed succeeds, the vector is discarded, and the entry is re-marked
`needs_reindex`. Each retrieval retries and fails identically.

## Steps to reproduce

Any OpenAI-compatible `/embeddings` endpoint reproduces this; TEI is only what
we happened to run.

1. Run an OpenAI-compatible embeddings server (e.g. TEI with
   `Qwen/Qwen3-Embedding-0.6B`, 1024-dim).
2. `POST /api/provider-nodes` with `apiType: "embeddings"` and that base URL,
   prefix `tei-tank`.
3. Set memory `embeddingSource: "remote"` and
   `embeddingProviderModel: "tei-tank/Qwen/Qwen3-Embedding-0.6B"`.
4. `POST /api/memory` to store an entry, then `POST /api/memory/reindex`.

**Expected:** the entry is vectorized; `rowCount` increases.
**Actual:** `rowCount` stays 0, `needsReindex` keeps growing,
`memory.vec.upsert.fail` with `no such table: vec_memories`, and
`/api/memory/health` still returns `working: true`.

Direct proof the endpoint itself is fine — through the gateway's own route:

```
POST /v1/embeddings {"model":"tei-tank/Qwen/Qwen3-Embedding-0.6B", ...}
→ HTTP 200, embedding length 1024
```

## Documented workarounds that do not apply

Checked the docs before filing:

- **`QDRANT_VECTOR_SIZE`** — `docs/frameworks/MEMORY.md` is explicit that
  `vectorSize` *"is not used for collection creation (dimension comes from the
  embedding)"*. It sizes the remote Qdrant collection only; the local
  `vec_memories` table is unaffected. Verified: setting it to 1024 and
  restarting changed nothing.
- **`customBaseUrl` / `customModelId`** (#10010) — documented as selecting an
  OpenAI-compatible endpoint *"without changing the global embedding
  registry"*, which is exactly this use case. Tested end to end: the setting
  is accepted, the embed call succeeds (`EMBED memory-custom/...` in the
  logs), and `ensure_ready.skipped` still fires with the same reason.
- **Borrowing a registry prefix** — declaring the model as
  `deepinfra/Qwen/Qwen3-Embedding-0.6B` (the one registry entry carrying this
  model at 1024) returns `400 No credentials for embedding provider:
  deepinfra`. Routing follows the prefix to the registry provider, so this
  would send traffic to a paid third party rather than the local server.
- **Lazy backfill on retrieval** — documented in
  `docs/frameworks/MEMORY.md` § *Backfill*. Triggered via a real semantic
  query; it reaches `ensureReady()` and skips for the same reason.
- **`VECTOR_STORE_DISABLE_VEC`** — degrades to keyword search, which defeats
  the purpose.

## Suggested fix

The measured dimension already exists at the only moment it matters. Feeding
`embeddingResult.dimensions` back after the first successful embed — the probe
the comment already promises — would close both paths without touching the
registry:

- on first successful embed, if `resolution.dimensions === null`, call
  `resetForSignature(signature, embeddingResult.vector.length)`;
- keep the signature stable afterwards so no reindex loop is triggered.

An explicit opt-in setting (`embeddingDimensions`) would also work and is
easier to reason about, but the auto-probe matches what the code already
advertises.

## Secondary observation

`/api/memory/health` reports `working: true` while nothing is vectorized,
because `verifyExtractionPipeline()` times three local SQLite operations
(`createMemory`, `listMemories`, `deleteMemory`) and the embed path is
`setImmediate`-deferred and never awaited. `latencyMs: 1` is therefore an
honest measurement of SQLite, but it says nothing about the embedding chain —
which is what makes this failure invisible in production. Surfacing
`needsReindex` in that response would make it detectable.

## Environment

- OmniRoute 3.8.50 (`ghcr.io/diegosouzapw/omniroute:3.8.50`), Docker
- sqlite-vec loads successfully (`[VECTOR_STORE] sqlite-vec loaded
  successfully`), migration 083 applied, `memory_vec_meta` present with
  `active_dim = null`, `vec_loaded = 0`
- Embedding backend: self-hosted TEI 1.8.3, `Qwen/Qwen3-Embedding-0.6B`,
  1024-dim, OpenAI-compatible `/v1/embeddings`
- Vector store: Qdrant 1.18.2 (also reproduced before Qdrant was enabled)

ntdatt812:
Fixes #12154. Your diagnosis is right, and tracing it showed the two paths are deadlocked on the same null in opposite ways.

## A lazy probe nobody performs

`EmbeddingResolution.dimensions` is documented as *"null antes da 1ª chamada (lazy probe)"* — null until the first call. Nothing ever performs that call:

| path | what it did |
|---|---|
| `scheduleVectorUpsert` (`store.ts`) | called `ensureReady()` with the null resolution — which declines to create `vec_memories` — then **ignored the `{ready:false}` answer** and upserted anyway, straight into the `catch`. Memory stored, `needs_reindex` set, never vectorized. |
| `runReindexBatch` (`reindex.ts`) | gated on `ensureReady()` *before* embedding, so it returned `{processed:0, errors:0}` and the queue never drained. |

So the width could only come from an embedding, and neither path would embed until it had the width. That is why your instances A and B sat at `rowCount: 0` with 24 pending, and why C only worked from a signature an older build had already written. Nothing surfaced it because the upsert is fire-and-forget: `POST /api/memory` still returns 200 and health still reports `working: true`.

## The fix: measure it

`withMeasuredDimensions(resolution, width)` fills the pending null in from a vector that has actually come back, and rebuilds the signature the way the resolution built it — **identity first, then model** — so two endpoints serving the same model id still get separate signatures and reindex independently. A width the registry already supplied is never overwritten, and a nonsense measurement (0, negative, fractional, `NaN`) is ignored rather than written into a signature.

- `store.ts` already holds a finished vector when it calls `ensureReady`, so it just passes `vector.length`. It now also honours a `{ready:false}` answer instead of upserting into a table that is not there.
- `reindex.ts` spends one embedding up front to measure, and **reuses that vector** for its own item rather than paying for it twice.

Registry-described models are unaffected: their `dimensions` is already non-null, so the helper returns the resolution unchanged and neither path changes behaviour.

## Tests

`tests/unit/memory-vec-lazy-probe-12154.test.ts`, 7 tests: the pending null is filled; the signature keeps the endpoint identity and gains the width; a resolution without an identity signs by model; a known width is never overwritten; four kinds of nonsense measurement are ignored; a source-less resolution stays unusable; and two endpoints serving the same model id do not collide.

| mutation | result |
|---|---|
| helper becomes a no-op | 4 failed |
| sign by model, ignoring identity | 2 failed |

## Commands run

```
node --import tsx/esm --test tests/unit/memory-vec-lazy-probe-12154.test.ts    7 pass, 0 fail
eslint <4 changed files> --suppressions-location config/quality/eslint-suppressions.json
                                                                              No issues found
tsc --noEmit -p tsconfig.json                                                  no errors in the changed files
```

Changed test files: `tests/unit/memory-vec-lazy-probe-12154.test.ts` (new).

The unit tests cover the resolution/signature logic, which is where the null originated. I have **not** run a self-hosted TEI server against this build, so the end-to-end claim — that `vec_memories` is now created and `rowCount` climbs — is reasoned from the two call sites above rather than observed. If you have one of the affected instances handy, that is the check worth doing before merge.


diegosouzapw:
Retargeted from `main` to `release/v3.8.51` (the active development branch). Heads-up on why: `main` only receives work at the release squash, so it currently carries an inherited base-red (#12133) whose failing checks — unit-test/pack-artifact gates killed on timeout — have nothing to do with your diff. Against the release branch your CI runs clean. No action needed from you; thanks for the fix!

diegosouzapw:
Heads-up @ntdatt812 — I retargeted this from `main` to `release/v3.8.51` (the active development branch), but that surfaced a second issue the retarget alone cannot fix.

Because the branch was cut from `main`, the diff now carries **28 files**: your actual fix is only 4 of them —

- `src/lib/memory/embedding/index.ts`
- `src/lib/memory/reindex.ts`
- `src/lib/memory/store.ts`
- `tests/unit/memory-vec-lazy-probe-12154.test.ts`

— while the other 24 are `main`-only infra that already lives on the release branch through separate main-twin PRs (CI workflows, `Dockerfile`/`Dockerfile.bun`, the electron lockfile, several `changelog.d` entries). Merging as-is would replay that infra back over the release branch.

Could you re-derive the branch from the current `release/v3.8.51` tip and keep just your 4 files?

```bash
git remote add upstream https://github.com/diegosouzapw/OmniRoute.git
git fetch upstream
git switch -c fix/12154-embedding-lazy-probe upstream/release/v3.8.51
git checkout <your-branch> -- src/lib/memory/embedding/index.ts src/lib/memory/reindex.ts src/lib/memory/store.ts tests/unit/memory-vec-lazy-probe-12154.test.ts
```

The fix itself looks right and #12154 is a genuine bug (memories stored but never vectorized, while `/api/memory/health` reports `working: true`) — we want this landed. Also note the CI failures currently shown on this PR are from the pre-retarget runs against the red `main`, not your code.

ntdatt812:
CI is red on this PR and neither failure is the diff. Evidence rather than a re-run request.

**Build** — not a build failure at all:

```
##[error]The runner has received a shutdown signal. This can happen when the runner
service is stopped, or a manually started runner is canceled.
##[error]The operation was canceled.
```

The same thing hit #12177 in the same window, and `main`'s own CI history shows the pattern independently — of the last twelve runs on `main`, seven are `cancelled`. Nothing was compiled and nothing failed.

**Vitest** — one test, `tests/unit/ui/use-provider-models-auto-fetch.test.tsx > synchronizes upstream models only when autoFetchModels is explicitly true`, in a file this branch does not touch. It passes locally on this branch and on `origin/main`, both `PASS (2) FAIL (0)`.

The mechanism looks like a fixed-sleep race rather than luck. The helper is:

```ts
async function flushQueuedSync() {
  await new Promise<void>((resolve) => setTimeout(resolve, 10));
  await Promise.resolve();
}
```

The hook has to complete `/api/providers`, queue the sync, then issue `/api/providers/connection-1/sync-models?mode=sync` — two awaited round-trips through the mock plus the effect flushes between them — and the assertion is `toHaveBeenCalledWith` on that second call. Ten milliseconds of wall clock is the whole budget. That run reported `environment 1375.26s` and `import 908.89s`, so the runner was heavily contended; on a quiet machine 10ms is plenty and on that one it was not.

If it is worth fixing rather than re-running, the smallest change is to wait on the condition instead of the clock:

```ts
await vi.waitFor(() =>
  expect(fetchMock).toHaveBeenCalledWith("/api/providers/connection-1/sync-models?mode=sync", expect.anything()),
);
```

which keeps the negative test above it honest — that one asserts the call is *never* made, so it still needs a bounded wait rather than a condition. Happy to send it as its own PR if you would like; it is unrelated to this branch and should not ride on it.

This branch's own suite, `tests/unit/memory-vec-lazy-probe-12154.test.ts`, is 7 passed.

diegosouzapw:
Babysit (/sweep-reds round 4) — not merging. Did **not** merge `origin/release/v3.8.51` (#12335 still open).

**Reds before** on `f3d6afbd56`: Fast Quality Gates (`complexity-ratchets`), Unit Tests fast-path 3/4 + 4/4, No new ESLint warnings.

**Own complexity (this PR's diff):** already extracted in `f3d6afbd56` (`measureUnknownWidth` / `ensureReindexStoreReady` / `reindexOneItem` in `src/lib/memory/reindex.ts`). Against merge-base `9d04995950` this PR touches 3 source files in scope (`embedding/index.ts`, `reindex.ts`, `store.ts`) — none of them are in the ratchet regression list.

**Remaining Fast Quality Gates hit is inherited, not this diff:**
```
[complexity] REGRESSÃO — open-sse/services/autoCombo/strictZeroCostFilter.ts: 0 → 1
[cognitive-complexity] REGRESSÃO — open-sse/services/autoCombo/strictZeroCostFilter.ts: 0 → 1
```
That file is **not** in this PR. The 0→1 is `#12319` (`2f33f2c20d`) on the red `release/v3.8.51` tip, visible because CI new-code mode diffs the GitHub merge-ref against stale `base.sha` `9d04995950`. Not chasing it here.

**Units 3/4 + 4/4 + ESLint are inherited** (drain in #12327 / #12331):
- `settings-i18n-keys` / locale `__MISSING__` → `radarPage.colLimits` / `trainsOnPrompts*`
- `createSSEStream` passthrough usage-only chunks
- unused `collectSSE` / `stream` / `writable` in `tests/unit/stream-passthrough-usage-estimation.test.ts`

**Local:** `tests/unit/memory-vec-lazy-probe-12154.test.ts` 7/7 pass. No push.

HOLD until the base-red drain lands. HEAD remains `f3d6afbd56`.

diegosouzapw:
Merge-refresh after drain **#12327** (`1586476183`). Clean merge.

Embedding lazy-probe delta unchanged (`reindex.ts` helpers). Remaining FQG hit on `strictZeroCostFilter.ts` from **#12319** is still inherited if that file's ratchet is on the tip — not this PR's 5 files.

New HEAD: `a257aa26f5`. Not merging this PR.

diegosouzapw:
Merged into `release/v3.8.51` as `9327990be6` — thanks @ntdatt812. It lands in the next release notes.

diegosouzapw:
Merging with `--admin`: the frozen red on this PR (Unit fast-path 4/4) ran against a merge-ref cut before #12327 drained the 2026-09-01 base-red window (settings-i18n-keys / placeholder-fallback lived in that shard). The fork is now locked (maintainerCanModify=false), so a refresh push to re-trigger CI is not possible. Evidence: the rebuilt branch's own test suite passed on push day (7/7 memory-vec-lazy-probe + typecheck + eslint), and the merge result against today's tip re-validated locally just now — 7/7 again. Closes the #12154 fix loop.

PR #12227, opened 2026-08-31
fix(combo): stop spurious universal-handoff on same-request fallback cascades

hartmark:
## Summary

Fixes a real production incident: a Swedish-language conversation got a Spanish reply from the model for no apparent reason. Root-caused via `correlation_id` in `call_logs` down to two compounding bugs in the universal-handoff mechanism, both in the same code family.

## Related Issues

- Not filed as an issue first; found and root-caused live while debugging a reported wrong-language reply.

## What was actually happening

One client request, traced via a shared `correlation_id`:
1. `deepseek-v4-flash-free` (a combo step) failed with `[400] Model is unavailable` — never reaches the client, purely internal.
2. The combo's usual next step (`big-pickle`) wasn't retried this time; it fell through to an `openrouter` free-tier model instead.
3. Because that landed model differed from what succeeded on the session's *previous turn*, the universal-handoff mechanism treated this as a genuine cross-turn handoff: it injected a `role: system` "you are being handed off" note into the wire request, and separately kicked off a background LLM call to generate/persist a conversation summary for future handoffs.
4. The no-summary fallback note (`buildUniversalHandoffSystemMessage`'s bare branch) was hardcoded in Spanish while every other string in the file is English — a plain copy/paste or authoring mistake, unnoticed since this specific fallback branch apparently isn't exercised by existing tests in a way that would surface it as wrong.
5. The model, given a system-role instruction with Spanish text mixed into an otherwise-Swedish conversation, replied in Spanish.

## Root cause and fix

**Bug 1 (language):** `buildUniversalHandoffSystemMessage` in `open-sse/services/contextHandoff.ts` — the no-summary-payload branch's `<note>` was hardcoded in Spanish (`"A continuación se resume toda la conversacion para continuar sin perder el hilo."`). Every sibling function/branch in this file uses English. Fixed to English, matching the rest of the file.

**Bug 2 (the actual trigger, in `combo.ts`):** the universal-handoff injection (and the paired `maybeGenerateUniversalHandoff` background-summary call) compares `modelStr` against `getLastSessionModel()` — the model that succeeded on the *previous turn* — on **every** combo target attempt, including fallback steps (`i > 0`) within the *same* client request. A same-request fallback landing on a different model than last turn's isn't a genuine handoff — the client's own request never exposed the earlier step's failure, so there's no discontinuity to explain to the model. The code already has the right signal to distinguish this: `i` (`targetIndex`) is used a few lines above to gate fallback-compression (`i > 0`), but the handoff blocks never checked it.

Fixed both handoff call sites (injection + background-summary generation) to only fire when `i === 0` (the combo's primary target). `recordSessionModelUsage` bookkeeping stays unconditional — it needs to reflect whichever model actually served the response, since the *next* request's `i === 0` comparison depends on that being accurate.

## User Impact

- Any combo whose primary target has an unreliable step ahead of a stable fallback (this combo's `deepseek-v4-flash-free` has failed 100% of ~1200+ attempts over the past week) was previously prone to spurious handoff noise on ordinary requests: an extra, unnecessary background LLM call per landed-on-a-different-fallback request, plus a `system`-role message injected into the wire request that could steer the model's behavior/language away from the actual conversation — exactly what happened here.
- Genuine cross-turn account/session handoffs (the feature's actual purpose) are unaffected: a single-target combo's only attempt is always `i === 0`, and a multi-target combo's primary-target attempt on a fresh request is also `i === 0`.
- Existing Spanish text, if it had ever reached a payload-bearing branch, is now English regardless.

## Validation

- [x] Focused tests: `tests/unit/universal-handoff.test.ts` (23/23), `tests/unit/context-handoff.test.ts`, `tests/integration/combo-matrix/context-relay-handoff.test.ts` — the one test exercising a genuine cross-turn handoff uses a single-target combo (always `i === 0`), confirmed unaffected by this change (passes identically with and without the fix).
- [x] `npm run lint` (scoped: eslint on the three touched files) — clean; no new findings on touched lines.
- [x] Reconciled with `upstream/release/v3.8.51` tip at branch time.
- [x] Production-code changes include an updated automated test in this PR (`universal-handoff.test.ts`'s assertion, which previously locked in the Spanish text as expected output).

One pre-existing, unrelated integration-test flake (`context-relay-handoff.test.ts`'s "does NOT fire when x-omniroute-session-id header is absent" case, a `503` from the combo's own retry-exhaustion path) reproduces identically on unmodified `upstream/release/v3.8.51` — confirmed via `git stash` before/after. Not caused by this change.

## Tests Added Or Updated

- `tests/unit/universal-handoff.test.ts`: updated the assertion in "buildUniversalHandoffSystemMessage basic when payload summary empty" to check for the corrected English text instead of the Spanish string it previously locked in as expected behavior.

## Coverage Notes

- `combo.ts`'s `i === 0` gating isn't independently unit-tested here — the existing `context-relay-handoff.test.ts` coverage only exercises the single-target (`i === 0` always) path. Adding a genuine multi-target fallback-cascade test for this exact scenario would need a fuller combo-matrix harness setup (scripted step failure + fallback success + differing prior-session model); flagging as a good follow-up rather than expanding this PR's scope, since the fix itself is a minimal, precisely-targeted one-line-condition change at both call sites, mirroring an already-tested pattern (`i > 0` fallback-compression gating a few lines above).

## Reviewer Notes

- Verified live on a real deployment: identical Spanish-language fallback message reproduced from real `call_logs`/transcript data (session/correlation IDs available on request, redacted from this description).
- The `i === 0` gate is the minimal, correctly-scoped fix; alternative designs (e.g., tracking a per-request "already attempted a different model this request" flag) would be more invasive for the same outcome.


hartmark:
Closing as superseded by #12338.

This PR's actual shipped diff is only the Spanish→English note-text fix (`open-sse/services/contextHandoff.ts` + `tests/unit/universal-handoff.test.ts`) — the title's claimed behavior (stopping the spurious same-request-fallback trigger, commit `ffbd0699`) was reverted by this PR's own second commit (`058c430a`) because it broke `combo-context-relay.test.ts`'s then-correct contract.

#12338 fully subsumes this PR:
- The same note is fixed there too (rewritten further, to also stop claiming continuity the request doesn't provide — not just translated).
- The actual same-request-fallback trigger fix this PR's title describes is implemented there properly, with the blocking test rewritten to assert the corrected contract instead of the old one.
- Plus the silent-failure diagnostic logging that traced this whole bug family live.

No content from this PR is lost — closing without merging.

PR #12252, opened 2026-09-01
fix(sse): classify "N API calls / month" 429 text as quota_exhausted

brick30llc-ctrl:
### Problem

`QUOTA_PATTERNS` in `src/shared/utils/classify429.ts` recognizes `monthly ... limit`, `monthly ... quota` and `per month ... limit`, but not the **reversed phrase order** providers actually use in trial-key messages.

Cohere's trial-key 429 reads:

```
You are using a Trial key, which is limited to 1000 API calls / month.
```

None of the existing patterns match that, so `classify429()` returns `rate_limit` instead of `quota_exhausted`. The caller then applies the generic short transient backoff to an allowance that will not reset until the next billing cycle, and the same key is retried every few seconds for the rest of the month.

### Evidence

Measured on a production instance over 24h against two Cohere trial connections: **481 of 752** 429s carried this exact text, and every one was treated as a short transient rate limit. The connections were never cooled for a meaningful window, so traffic kept being dispatched to a provider with no remaining monthly quota.

### Fix

Three additional patterns covering the reversed order and the `N calls / month` / `N requests/month` spellings:

```ts
/(?:api\s+)?calls?\s*\/\s*month/i,
/requests?\s*\/\s*month/i,
/limited\s+to\s+[\d,]+.*(?:calls?|requests?).*per\s+month/i,
```

They are deliberately narrow — each requires an explicit `month` token adjacent to a call/request noun — so ordinary per-minute or per-day rate-limit text is unaffected.

### Tests

Added `monthly call allowances classify as quota_exhausted regardless of phrase order`, covering both orders plus the verbatim Cohere string.

Verified TDD on this branch:

- **Before the fix** the new test fails: `Limited to 1000 API calls per month` → `actual: 'rate_limit', expected: 'quota_exhausted'`.
- **After the fix** `tests/unit/classify429.test.ts` is 31/31.

No regressions in the neighbouring quota/classification suites:

| Suite | Result |
| --- | --- |
| `classify429` | 31/31 |
| `circuit-breaker-failure-kind` | 12/12 |
| `antigravity-429-quota-tdd` | 5/5 |
| `antigravity-429-quota-cooldown` | 13/13 |
| `cloudflare-ai-neuron-exhaustion-6980` | 11/11 |
| `egress-ip-lock-10880` | 17/17 |

Prettier clean on both changed files.

### Notes

Branched from `origin/release/v3.8.51` and targets that branch. No open `base-red` issue and no active `release-freeze` at the time of cutting.

The changelog fragment is added in a follow-up commit on this branch once the PR number is known, per `changelog.d/README.md`.


diegosouzapw:
Thanks for the fix! Verified locally: `classify429.ts`'s `QUOTA_PATTERNS` was indeed
missing the reversed "N calls/month" phrasing Cohere's trial-key 429 uses, and your new
test (31/31 passing, including your 4 phrasing variants) proves it. Your branch already
merges the current release tip cleanly (including the recent CJK quota patterns), so this
looks ready to merge as-is.

diegosouzapw:
Thanks @brick30llc-ctrl — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #12214, opened 2026-08-31
feat(providers): expose usage-supported in provider plugin manifest

maxmad64bis:
## Summary

Publish the second usage capability in the provider plugin manifest. The manifest already exposes `usage-fetch` (40 providers, #11903); this PR adds `usage-supported` so integrators can tell without reading TypeScript whether the server usage routes accept a provider.

- New `usage-supported` tag in `ProviderPluginCapability` and `capabilitiesFor()` (`open-sse/config/providerPluginManifest.ts`).
- `USAGE_SUPPORTED_PROVIDERS` extracted to a zero-import leaf `src/shared/constants/providers/usageSupported.ts`, re-exported from `providers.ts` (same move as `fetcherProviders.ts` in #11903; the manifest graph stays light).
- Docs `docs/reference/PROVIDER_PLUGIN_MANIFEST.md`: 8 tags, distinction between `usage-fetch` and `usage-supported` with the 7 measured differences.
- No list mutation, no new route, `schemaVersion` stays `1`.

`usage-fetch` resolves on id or alias (the dispatcher accepts both); `usage-supported` resolves on id alone — the runtime guard does `USAGE_SUPPORTED_PROVIDERS.includes(providerId)` with no alias resolution (`src/shared/utils/providerQuotaVisibility.ts`, `src/lib/usage/providerLimits.ts:isSupportedUsageConnection`), so the manifest keeps the same rule.

The two scopes differ: 42 providers carry both tags, 4 carry only `usage-fetch` (`opencode`, `opencode-zen`, `openrouter`, `xai`) and 3 only `usage-supported` (`adobe-firefly`, `firefly`, `xiaomi-mimo-token-plan`) — 7 differences in total, so one tag does not imply the other.

## Related Issues

- Related to #11722 — follow-up to #11903. #11903 noted: _"The acceptance criteria only list `usage-fetch`, so this PR ships just that. Happy to add the second tag in a follow-up if you want it."_ This PR is that follow-up.
- #11903 closed #11722 after shipping only half of it; this PR publishes the second half.

## Validation

- [x] Change type: provider
- [x] Focused tests and category gates from the golden path
- [x] `npm run lint`
- [x] Reconciled with the current active release base; focused checks rerun afterward — base `upstream/release/v3.8.51` (`63e4afa32`)
- [x] Production-code changes include a new or updated automated test in this PR
- [x] `check:cycles` — pass (import-free leaf, no cycles)
[...]

PR #12256, opened 2026-09-01
feat(usage): devin-cli agentic quota + openrouter credits in Provider Limits

Neuron-Mr-White:
## Summary

Two provider families expose live quota APIs but were missing from the Provider Limits dashboard because their list entries were absent from the sync gates:

- **devin-cli** — new usage leaf querying the Codeium seat-management Connect API (`exa.seat_management_pb.SeatManagementService/GetUserStatus`, protobuf over POST with the raw `Basic <token>-<token>` auth header the official CLI itself uses). Surfaces the plan name plus **daily / weekly agentic quota percentages** with reset timestamps from the `GetUserStatus` `plan_status` payload, via a minimal hand-rolled protobuf encoder/reader (two fixed messages — no proto dependency warranted).
- **openrouter**: the `/key` + `/credits` quota fetcher (#6842) was already wired into the dispatcher but gated out of the bulk sync — this PR adds it to `USAGE_SUPPORTED_PROVIDERS` and `PROVIDER_LIMITS_APIKEY_PROVIDERS` so key limits and account credits actually surface.

Also fixes a standalone-build regression the vendored ChatGPT Web connector v4.0.7 (#12181) introduced: tiktoken's node build reads `tiktoken_bg.wasm` via a `__dirname`-relative `fs.readFileSync` at import time, so bundling it makes page-data collection fail with `Missing tiktoken_bg.wasm` for any route reaching the tokenizer. Externalized like the other runtime-resolved wasm/native packages (`sql.js`, `sqlite-vec`, `better-sqlite3`).

#### Test plan

- [x] New unit suite `tests/unit/usage-devin-cli.test.ts` (7 tests): protobuf wire helpers (decode + malformed rejection), `GetUserStatus` response parser against a wire-format fixture matching the live API shape, dispatcher wiring, graceful no-token message
- [x] `npm run typecheck:core` clean; eslint clean on changed files (3 pre-existing base errors in `usage.ts` unchanged)
- `db-provider-limits.test.ts`, `quota-preflight.test.ts`, `openrouter-free-model-credits-exhausted.test.ts` all green
- Verified live end-to-end against the real Codeium seat-management API and OpenRouter `/key` + `/credits` endpoints, then deployed blue/green: `POST /api/usage/provider-limits` now returns `devin-cli` → plan "Pro" with `daily` / `weekly` agentic quota percentages + reset timestamps, and both openrouter connections surface `free_daily`, `free_rpm`, and `credits` (USD balance)


killer30001000:
I tested the OpenRouter Provider Limits flow end-to-end against a real PAYG account and noticed one PAYG case that is still distinct from the credits-only rendering added in this PR.

The current PR correctly makes `/credits` usable independently from `/key` and surfaces the remaining USD balance. However, when `/key` succeeds with:

- `limit: null`
- `limit_remaining: null`

while `/credits` returns both:

- `total_credits`
- `total_usage`

the account credit pool is still finite and has a usable denominator.

Live example:

- `total_credits`: 10
- `total_usage`: ~0.697
- remaining balance: ~9.303

The expected quota semantics are therefore:

- `used`: ~0.697
- `total`: 10
- `remaining`: ~9.303
- `remainingPercentage`: 93
- `unlimited`: false

At the moment, `buildCreditsQuota()` in `open-sse/services/usage/openrouter.ts` still treats `limit === null` as:

- `total: 0`
- `remainingPercentage: 100`
- `unlimited: true`

so this PAYG case loses the real account-level percentage even though `/credits` provides enough data to calculate it.

I also tested the per-key limit case live. When a key-level cap is configured, the key limit should take precedence over the account credit pool.

For example:

- account credits: $10
- API key limit: $5
- `/key` returns `limit: 5` plus `limit_remaining`

In that case the quota should use the $5 key cap rather than the $10 account total.

Suggested precedence:

1. If a valid key limit exists:
   - `total = limit`
   - `remaining = limit_remaining`
   - `used = total - remaining`
   - percentage from `remaining / total`

2. If there is no key limit, but `/credits` provides a positive `total_credits` and `total_usage`:
   - `total = total_credits`
   - `used = total_usage`
   - `remaining = creditBalance` or `total - used`
   - percentage from `remaining / total`
   - `unlimited = false`

3. If only a balance is known and there is no usable denominator:
   - expose the known USD balance
   - do not invent a percentage

One practical note from live testing: the OpenRouter quota fetcher caches results for ~45 seconds, so changing a key-level limit can briefly show the previous state until the cache expires.

I verified both cases against the real OpenRouter API:
- PAYG account with no per-key limit
- same account with a $5 per-key limit

The behavior above matched the actual `/key` and `/credits` responses.

diegosouzapw:
Babysit (/sweep-reds round 3, Fase B) — not merging.

**Reds before:** Fast Quality Gates (`mutation-test-coverage`, `complexity-ratchets`), Docs Gates (`check:env-doc-sync` — `DEVIN_SEAT_API_URL`), Unit Tests fast-path 3/4 and 4/4 (env contract + native Codex turn-pin, the latter inherited from a stale base).

**Action:**
- Merged `origin/release/v3.8.51` (`1fa313cc03`) — should clear the inherited turn-pin / stryker tap failures.
- Documented `DEVIN_SEAT_API_URL` in `.env.example` + `docs/reference/ENVIRONMENT.md`.
- Split `fetchOpenrouterQuota` / `decodeProtoFields` so the new-code complexity ratchets stay at 0 (`dc7a1eb9bb`).

**Local signal:** `check-env-doc-sync` green; `tests/unit/usage-devin-cli.test.ts` + `tests/unit/openrouter-quota-6842.test.ts` 33/33 pass.

CI running on `dc7a1eb9bb`.

diegosouzapw:
Follow-up: original Docs Gates (`DEVIN_SEAT_API_URL`) is **green** on this run.

`No new ESLint warnings` is the same inherited base issue as #12214: unused vars in `tests/unit/stream-passthrough-usage-estimation.test.ts` (#12151 on `release/v3.8.51`), not this PR's usage/quota files. Not patching that inside this contributor branch.

Units + Fast Quality Gates still running.

PR #12166, opened 2026-08-31
fix(translator): strip neutral tool_choice when tools absent in Responses-to-Chat (#12141)

HouMinXi:
## Summary
Closes #12141.

When clients invoke the `/v1/responses` endpoint with neutral `tool_choice` (`"auto"` or `"none"`) without providing tools (or passing `tools: []`), translating to Chat Completions (`openaiResponsesToOpenAIRequest`) previously preserved `tool_choice` verbatim. Strict Chat Completions backends (such as vLLM, Ollama, and strict OpenAI-compatible engines) reject requests containing `tool_choice` without `tools` with `HTTP 400: When using tool_choice, tools must be set`.

## Changes
- **`open-sse/translator/request/openai-responses.ts`**: In `openaiResponsesToOpenAIRequest`, after tool extraction and filtering, when `finalChatTools.length === 0` and `result.tool_choice` is `"auto"` or `"none"`, delete `result.tool_choice`. Contradictory choices (`"required"` or explicit function objects) remain preserved so upstream contracts remain transparent.
- **`tests/unit/responses-to-chat-no-tools-tool-choice-12141.test.ts`**: Dedicated test suite verifying empty tools with `auto`, `none`, valid tools preservation, and contradictory choice retention.
- **`tests/unit/responses-translation-fixes.test.ts`**: Updated existing test case to verify both with-tools preservation and without-tools stripping.

## Verification
- `npm run typecheck:core`: 0 errors.
- `npm run check:cycles`: PASS (no cycles).
- `node --test tests/unit/responses-to-chat-no-tools-tool-choice-12141.test.ts tests/unit/responses-translation-fixes.test.ts`: 48/48 PASS.
- Regression suite: all translator tests (76/76) PASS.

PR #11950, opened 2026-08-28
chore(quality): remove unreachable code and restore test discovery

backryun:
## Summary

- remove 48 verified unreachable files across detached dashboard UI, superseded quota components, unused docs/Storybook components, stale helpers, and unused OAuth CLI presentation utilities
- remove two obsolete tests that did not exercise current production behavior
- move the memory settings API test into the discovered Vitest tree, update it for the current response contract, and import the owning DB module directly
- add the translator test directory to the Node unit collectors, merge-train fast path, and impacted-test selector
- remove stale ESLint suppressions for deleted or already-clean files
- keep documented runtime error codes distinguishable from environment variables after removing the unused API response constants module
- tighten tracked quality debt: dead-code total 408 -> 360 and frozen test-discovery orphans 13 -> 10

This intentionally stops at static-analysis boundaries: dynamic plugin/worker/vendor entrypoints, side-effect barrels, OneProxy contracts, and tool-policy contracts are left untouched for separate review.

## Validation

- `npm run test:vitest` — 50 files, 459 tests passed
- `npm run typecheck:core`
- `npm run check:cycles`
- `npm run check:dead-code` — 331 unused exports + 29 unused files = 360
- `npm run check:test-discovery` — 5261 test files, 32 collectors, 10 frozen orphans
- fabricated-docs and release-green docs regressions — 28 passed
- `npm run check:docs-all` — passed
- focused merge-train and impacted-test regressions — 25 passed
- changed-file ESLint with the repository suppression config — passed with zero warnings
- pre-commit lint-staged and repository policy hooks — passed
- `git diff --check`

## Known inherited gate failures

- full `lint:json` currently depends on #11955, which freezes 55 pre-existing `no-explicit-any` violations and removes the unused `GPT_SIZE_MAP`
- unit shard 3/4 currently depends on #11957, which fixes the base test import in `tests/unit/openai-to-claude-trailing-usage-deferred-finish.test.ts`
- hosted `Fast Production Build` and `dast-smoke` OOM/cancellation are tracked by #11946

No runtime behavior change is intended; the production-code portion is deletion of unreachable modules plus a direct import from the existing settings DB module.

PR #11259, opened 2026-08-23
chore(deps): refresh runtimes and adopt ESLint 10

backryun:
## Summary

This PR modernizes the repository's dependency and lint toolchain without changing provider behavior or application features.

- adopt ESLint 10.9.0 and make the flat config compatible with its stricter plugin/parser contracts
- align all remaining Bun toolchain surfaces with the repository's exact Bun 1.4.0 pin
- refresh compatible runtime, framework, and development dependencies
- regenerate and validate the npm lockfile

## Implementation

- wrap the Next.js preset with `fixupConfigRules()` from `@eslint/compat`
- use `espree` for JavaScript files and the TypeScript parser for `.mts` / `.cts`
- avoid duplicate `@typescript-eslint` plugin registration under ESLint 10
- add `@eslint/compat` and `espree` to the dependency policy allowlist
- remove 165 suppressions that no longer match live ESLint findings
- update both Bun Docker stages from `1.3.14-slim` to `1.4.0-slim`
- pin `@types/bun` to 1.4.0 and refresh the remaining compatible direct dependencies

## Scope

The branch was rebuilt as a single commit on `release/v3.8.51` at `63e4afa32`. The current diff is limited to:

- `Dockerfile.bun`
- `eslint.config.mjs`
- `package.json`
- `package-lock.json`
- `config/quality/dependency-allowlist.json`
- `config/quality/eslint-suppressions.json`

Provider catalog changes, application code, and unrelated test-contract repairs are intentionally excluded. The Gemini 3.5 Flash removal from the PR's original history is already present on the release branch and is not part of this diff.

## Validation

- `npm ci --no-audit --no-fund` — pass (2,520 packages)
- `npm run lint -- --no-cache` — pass
- `npm run check:lockfile` — pass
- `npm run check:deps` — pass
- `npm run check:node-runtime` — pass on Node.js 26.8.1
- `npm run typecheck:core` — pass
- `npm run test:vitest` — 50 files / 463 tests pass
- unchanged dashboard provider regression — 3/3 pass
- unchanged Alibaba allowlist regression — 4/4 pass
- `npm run check:provider-consistency` — pass on Bun 1.4.0
- `npm run check:compression-budget` — pass on Bun 1.4.0 with isolated database paths
- `npm run check:known-symbols` — pass on Bun 1.4.0 with isolated database paths
- `git diff --check` and pre-commit quality gates — pass

[...]

PR #12258, opened 2026-09-01
[URGENT] fix(dev): qualify Turbopack runtime boundaries (phase 5)

backryun:
## Summary

Completes the implementation side of Phase 5 for #12074 by fixing the two exact runtime boundaries reproduced on the post-Phase-4b release head:

- keep `tiktoken` external so Node selects its CommonJS loader instead of Turbopack bundling the ESM `tiktoken_bg.wasm` import;
- retain `tiktoken_bg.wasm` in standalone output tracing;
- anchor the optional SQLite loader to the real process entrypoint and invoke it through `Reflect.apply`, preventing Turbopack from rewriting the dynamic request into its in-bundle resolver.

This does not add warning ignores or mix provider/dependency/catalog work into the phase.

## Reproduction and result

Qualification was repeated from an empty Next cache and isolated SQLite directory on release head `2e17161ea` with `OMNIROUTE_USE_TURBOPACK=1`.

Before this patch:

- both `better-sqlite3` and `node:sqlite` failed with `Cannot find module as expression is too dynamic`, forcing the `sql.js` fallback;
- `/api/providers` returned 500 because `tiktoken_bg.wasm` was missing during module evaluation;
- the same failures were repeated across provider/settings/instrumentation import paths.

After this patch:

- cold start reached the Turbopack listening state in **4.350 s**;
- the runtime selected **better-sqlite3** directly;
- the targeted log scan found no dynamic-module error, missing tiktoken WASM, Turbopack panic, `createRequire` parse failure, critical-dependency flood, or `MaxListenersExceededWarning`;
- graceful SIGINT drained requests, checkpointed SQLite, stopped the ChatGPT Web runtime and logger resources, and exited cleanly.

| Route | Status | First request |
| --- | ---: | ---: |
| `/dashboard/providers` | 200 | 3.174 s |
| `/api/health/ping` | 200 | 0.373 s |
| `/api/providers` | 200 | 0.600 s |
| `/api/settings` | 200 | 0.218 s |
| `/api/v1/models` | 200 | 1.534 s |

Three real provider-page source edits/restoration returned 200 in **59 ms / 61 ms / 77 ms**. RSS moved from **4,805,408 KiB** after the first HMR graph activation to **4,822,272 KiB** after the third cycle (+16,864 KiB), rather than showing the previous runaway pattern.

After the representative routes and three HMR cycles:

- RSS: **4,822,272 KiB**
- total dev output: **2,389,080 KiB**
- cache: **1,337,708 KiB**
- server output: **934,828 KiB**
- static output: **115,092 KiB**

All temporary source probes were restored. The isolated DBs and approximately 4.6 GiB of before/after generated caches were deleted after measurement.

## Automated validation

- focused Node regression set: **41/41 passed**
- Vitest: **50 files, 463/463 passed** (serial workers with an isolated DB)
- `npm run typecheck:core`
- Prettier, ESLint, `git diff --check`, pre-commit documentation sync and quality gates

I also started the full `npm run test:unit` command. It did not complete: the pre-existing `warmupScheduler/circuitBreakerFactory*` group remained pending for more than five minutes and blocked the remaining runner queue, so I interrupted it rather than claiming a full-suite pass. The changed DB/config tests and both non-overlapping focused/Vitest suites above are green.

## Scope

- `next.config.mjs`
- `src/lib/db/adapters/runtimeRequire.ts`
- `tests/unit/turbopack-runtime-boundaries-12074.test.ts`

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -286,6 +286,9 @@ const nextConfig = {
       // (better-sqlite3 → node:sqlite → sql.js). Next traces sql-wasm.js but can
       // omit the runtime sql-wasm.wasm asset from the standalone bundle.
       "./node_modules/sql.js/dist/sql-wasm.wasm",
+      // tiktoken is server-externalized below so Node selects its CommonJS entry.
+      // That entry reads the tokenizer WASM beside itself at runtime.
+      "./node_modules/tiktoken/tiktoken_bg.wasm",
     ],
   },
   outputFileTracingExcludes: {
@@ -339,6 +342,10 @@ const nextConfig = {
     "tough-cookie",
     "@ngrok/ngrok",
     "@huggingface/transformers",
+    // The ESM entry imports tiktoken_bg.wasm as a module. Turbopack can compile
+    // that graph but omits the runtime asset, making provider routes fail during
+    // module evaluation. Keep Node's CommonJS loader and colocated WASM intact.
+    "tiktoken",
     // copilot-m365-web.ts imports 'ws' as a client-side WebSocket. When bundled,
     // ws cannot resolve its 'bufferutil' native addon (frame masking) and throws
     // TypeError: b.mask is not a function on the first outgoing frame, causing
--- a/src/lib/db/adapters/runtimeRequire.ts
+++ b/src/lib/db/adapters/runtimeRequire.ts
@@ -5,13 +5,17 @@
  * The standalone server is emitted as CommonJS chunks and externalizes the
  * native database packages. Keep those requests as static `require()` calls so
  * webpack preserves the external boundary. Development and tests run as ESM,
- * where `require` is unavailable; the `createRequire(import.meta.url)` fallback
- * handles those callers.
+ * where `require` is unavailable; the fallback anchors resolution to the real
+ * process entrypoint so Turbopack cannot replace it with an in-bundle resolver.
  */
 import * as nodeModule from "node:module";
 
+const esmRequire = nodeModule.createRequire(process.argv[1] || process.cwd());
+
 function esmRuntimeRequire(specifier: string): unknown {
-  return nodeModule.createRequire(import.meta.url)(specifier);
+  // Reflect keeps the optional request dynamic. A direct call is rewritten by
+  // Turbopack and fails at runtime as "Cannot find module as expression is too dynamic".
+  return Reflect.apply(esmRequire, undefined, [specifier]);
 }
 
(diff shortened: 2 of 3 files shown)


backryun:
The two failed unit shards are inherited release contract drift from #12247, not failures in this PR's Turbopack/runtime-boundary files.

Exact failures:

- shard 3/4: `native-codex-turn-pin-10379.test.ts` — `isPinnedTargetModelScopedUnusable distinguishes model lockout from provider/connection outages`
- shard 4/4: `native-codex-turn-pin-model-scoped-fallback.test.ts` — `Turn pin NOT released when provider in global cooldown`

#12247 correctly changed provider-level cooldowns to require `PROVIDER_PROFILES.*.providerFailureThreshold` failures, but both older tests still recorded one failure and assumed global cooldown was active. I reproduced the same 2/15 failures locally on the release head.

The test-only alignment is isolated in #12259 to preserve this Phase 5 PR's bundler-only scope. The corrected turn-pin tests plus the new provider-window suites pass 38/38 locally. Once #12259 merges, this PR should be rebased/refreshed and the two inherited shards rerun.

diegosouzapw:
Validated in local merge-train on 192.168.0.113 — train of #12258 #12262 #12166 #12281 #11259 #11950 merged clean onto `origin/release/v3.8.51` (f5e70950d3):

- Run 1 (`/opt/actions-runner-omniroute-5`, train tip `85cd9119`): typecheck:core, file-size, complexity ×2, changelog-integrity — all green; the `test:unit` step was killed by a CI job landing on that runner mid-train (workspace clobbered — infra, documented risk).
- Run 2 (`/srv/omniroute-train/.claude/worktrees/mt-green1`, same 6 PRs re-boarded, fresh `npm ci`): **unit 35768/35805 pass** (`/tmp/mt-unit2.log`), **vitest 464/465** (`/tmp/mt-vitest2.log`).
- Every failure is a latency/timing assert (bounded-time, event-loop lag, cooldown windows) on a box that was never idle (3 active CI runner workers throughout). Discriminated per merge-gates §3: the 3 persistent titles reproduce **identically on the pure base tip on the same box** (`/tmp/mt-base-isolated.log`, BASE_ISOLATED_EXIT=1 — same tests, same asserts), 5 more titles reproduce on the pure base on the devbox, and the single vitest failure (`provider-family-combos`) also fails on `main`'s nightly without any of these PRs. Inherited/infra — not introduced by this train.

Issue #3587, opened 2026-06-10
[BUG] Reasoning models in combos consume all tokens for reasoning_content, leaving content empty


herjarsa:
## OmniRoute Version

3.8.20 (latest main)

## Installation Method

Local build (Windows)

## Operating System

Windows 11 (WSL2)

## Provider(s) Involved

All providers with reasoning models: opencode (deepseek-v4-flash), kilocode (nemotron), kiro (deepseek-3.2)

## Model(s) Involved

Combo models that route to reasoning models: CODING, EXPLORE, Kilo-Code, MASTER-LIGHT

## Client Tool

OpenCode (AI SDK)

## Description

When combo models route to reasoning models (deepseek-v4-flash, nemotron, kiro/deepseek-3.2, etc.), the models consume ALL available `max_tokens` for `reasoning_tokens` and return empty `content`. This causes OpenCode (and likely other clients) to treat the response as an empty/failed response.

### Root Cause

Reasoning models split their token budget into `reasoning_tokens` (thinking) and content tokens. When `max_tokens` is set by the client, the reasoning model allocates the entire budget to reasoning, leaving 0 tokens for the actual content response.

### Reproduction

```bash
# Test with a combo that routes to a reasoning model
curl -s -X POST http://localhost:20128/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <api-key>" \
  -d '{
    "model": "CODING",
    "messages": [{"role":"user","content":"Say OK"}],
    "max_tokens": 100,
    "stream": false
  }'
```

Response:
```json
{
  "choices": [{
    "finish_reason": "length",
    "message": {
      "content": "",
      "reasoning_content": "Thinking. 1. Analysis..."
    }
  }],
  "usage": {
    "completion_tokens_details": {
      "reasoning_tokens": 100
    }
  }
}
```

The `content` field is empty because all 100 tokens were consumed by reasoning.

### Expected Behavior

The combo routing layer should either:
1. **Add a reasoning token buffer**: When routing to a reasoning model, request `max_tokens + buffer` (e.g., +2048) from the upstream provider to ensure enough tokens remain for content after reasoning
2. **Detect and retry**: When `finish_reason: "length"` AND `content` is empty AND `reasoning_tokens > 0`, automatically retry with a larger `max_tokens`
[...]

PR #12262, opened 2026-09-01
fix(sse): trust finish_reason over reasoning-ratio heuristic in response quality validation

brick30llc-ctrl:
## Summary

`validateResponseQuality()`'s reasoning-truncation guard (added for #3587) only rejects a
response with empty `content` + present `reasoning_content` when reasoning consumed **90%+**
of `completion_tokens`. A response truncated at a lower ratio slips through as "valid" even
though the caller gets nothing usable — and the response already carries an unambiguous
signal that was going unread: `finish_reason: "length"` (or the `"max_tokens"` naming some
providers use).

Reproduced live against `nvidia/nemotron-3-super-120b-a12b` on a self-hosted deployment:
`content: null`, `finish_reason: "length"`, `reasoning_tokens: 645` / `completion_tokens: 1024`
(63% — below the 90% threshold). The combo loop treated it as a successful completion and
never retried, so the end client received an empty answer for what looked like a normal,
non-degraded 200 response.

## Fix

When `content` is empty and `reasoning_content` is present (the existing precondition for
this whole branch), check `finish_reason` first: `"length"` or `"max_tokens"` is trusted
directly as truncation, independent of the token ratio. The existing ratio heuristic remains
as a fallback for providers that don't report `finish_reason` reliably. Does not touch or
weaken the deliberate-tiny-probe case (e.g. `max_tokens: 1` connectivity pings, see
`errorClassifier.ts`'s `LEGIT_EMPTY_OPENAI_FINISH`) — those never produce `reasoning_content`
at all, so this branch's precondition already excludes them.

## Test plan

- [x] 4 new unit tests in `tests/unit/combo-quality-validator-reasoning.test.ts`: the exact
      63%-ratio repro (now invalid), the `max_tokens` naming variant, a regression guard that
      no-`finish_reason` + <90% ratio stays valid (unchanged #3587 behavior), and a check that
      `finish_reason: "stop"` doesn't short-circuit the ratio heuristic.
- [x] `node --import tsx/esm --test tests/unit/combo-quality-validator-reasoning.test.ts` —
      16/16 pass (12 pre-existing + 4 new).
- [x] Adjacent quality-validation suites unaffected: `validate-response-quality.test.ts` (17),
      `quality-validation-benign-error.test.ts`, `routing-quality.test.ts`,
      `routing-scoring-quality.test.ts`, `quality-rail-gate-membership.test.ts` (23) — all pass.
- [x] `npm run typecheck:core` — only pre-existing, unrelated `omniglyph` errors (confirmed
      present on the base branch too, nothing from the changed files).
- [x] `eslint` on both changed files — clean.

### DIFF
--- a/changelog.d/fixes/12262-reasoning-truncation-quality-guard.md
+++ b/changelog.d/fixes/12262-reasoning-truncation-quality-guard.md
@@ -1,0 +1,1 @@
+- **fix(sse):** trust `finish_reason: "length"`/`"max_tokens"` over the reasoning-consumed-token ratio in response quality validation, so a reasoning model truncated below the old 90% threshold correctly fails and retries instead of returning empty content as a silent "success" ([#12262](https://github.com/diegosouzapw/OmniRoute/pull/12262))
--- a/open-sse/services/combo/validateQuality.ts
+++ b/open-sse/services/combo/validateQuality.ts
@@ -745,6 +745,24 @@ export async function validateResponseQuality(
   // tokens or falls back to a non-reasoning model.
   const contentIsEmpty = content === null || content === undefined || content === "";
   if (contentIsEmpty && hasReasoningContent && !hasToolCalls) {
+    // The 90%-of-completion-tokens ratio below is a proxy for "the request was
+    // truncated mid-reasoning" for providers that don't report finish_reason
+    // reliably. When finish_reason IS reported as "length" (or the Anthropic-shape
+    // "max_tokens"), that's a direct, unambiguous signal of truncation — trust it
+    // over the ratio instead of requiring reasoning to also clear 90%. A response
+    // truncated at, say, 60% reasoning still has zero usable content for the
+    // caller. This does not affect the deliberate-tiny-probe case (e.g.
+    // `max_tokens: 1` connectivity pings, see errorClassifier.ts's
+    // LEGIT_EMPTY_OPENAI_FINISH): those produce no reasoning_content at all, so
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Validated in local merge-train on 192.168.0.113 — train of #12258 #12262 #12166 #12281 #11259 #11950 merged clean onto `origin/release/v3.8.51` (f5e70950d3):

- Run 1 (`/opt/actions-runner-omniroute-5`, train tip `85cd9119`): typecheck:core, file-size, complexity ×2, changelog-integrity — all green; the `test:unit` step was killed by a CI job landing on that runner mid-train (workspace clobbered — infra, documented risk).
- Run 2 (`/srv/omniroute-train/.claude/worktrees/mt-green1`, same 6 PRs re-boarded, fresh `npm ci`): **unit 35768/35805 pass** (`/tmp/mt-unit2.log`), **vitest 464/465** (`/tmp/mt-vitest2.log`).
- Every failure is a latency/timing assert (bounded-time, event-loop lag, cooldown windows) on a box that was never idle (3 active CI runner workers throughout). Discriminated per merge-gates §3: the 3 persistent titles reproduce **identically on the pure base tip on the same box** (`/tmp/mt-base-isolated.log`, BASE_ISOLATED_EXIT=1 — same tests, same asserts), 5 more titles reproduce on the pure base on the devbox, and the single vitest failure (`provider-family-combos`) also fails on `main`'s nightly without any of these PRs. Inherited/infra — not introduced by this train.

Issue #12308, opened 2026-09-01
fix(providers): response schemas lose nullability on the Gemini hop — model returns "null" or fabricates values


amirrezakm:
# Response schemas lose nullability on the Gemini hop — the model answers `"null"` or invents a value

**OmniRoute Version:** 3.8.50 (reproduced live); defect verified unchanged at `release/v3.8.51` HEAD `d920e64`
**Installation Method:** Docker / Docker Compose
**Operating System:** Linux (Ubuntu 26.04)
**Component:** `open-sse/translator/helpers/geminiHelper.ts` → `cleanJSONSchemaForAntigravity`
**Severity:** silently fabricated data in structured output

## Summary

`cleanJSONSchemaForAntigravity` flattens every union spelling of "nullable" before the schema reaches Gemini. That is correct for a **tool parameter** — Gemini wants a concrete type there and the caller decides what an absent argument means. It is wrong for a **response schema**, where the union is the only thing telling the model that "nothing" is a legal answer.

Stripped of it, a model with nothing to say cannot answer `null`. It answers with the *string* `"null"`, or it invents a plausible value. Both reach the client as schema-conformant data.

## Reproduction

Empty payload, a prompt that explicitly asks for null, one nullable field:

```bash
curl -s $BASE/v1/chat/completions -H "Authorization: Bearer $KEY" \
  -H 'content-type: application/json' \
  -H 'x-omniroute-no-cache: true' \
  -d '{
  "model":"<antigravity-gemini-model>","temperature":0,
  "messages":[
    {"role":"system","content":"Write a description of the product. If the payload carries nothing to describe, the value must be null."},
    {"role":"user","content":"{}"}
  ],
  "response_format":{"type":"json_schema","json_schema":{"name":"d","strict":true,"schema":{
    "type":"object","properties":{"value":{"type":["string","null"]}},
    "required":["value"],"additionalProperties":false}}}}'
```

> Send the bypass header. Without it this request is a strong candidate for a stale cache entry — see #12307 — the cache signature omits `response_format`.

Observed across all three spellings:

```
type: ["string","null"]              -> {"value": "null"}      (the 4 characters, as a string)
anyOf: [{string}, {null}]            -> a fabricated description
oneOf: [{string}, {null}]            -> a fabricated description
```

Expected: `{"value": null}` in all three.

### The fabrication case is the serious one

[...]

PR #12309, opened 2026-09-01
fix(cache): fold the response output contract into the semantic cache signature

### LINKED ISSUE #12307: fix(backend): semantic cache signature omits response_format/tools — structured-output requests are served each other's bodies
# Semantic cache key omits `response_format` — structured-output requests return each other's bodies

**OmniRoute Version:** 3.8.50 (reproduced live); defect verified unchanged at `release/v3.8.51` HEAD `d920e64`
**Component:** `src/lib/semanticCache.ts` → `generateSignature`
**Severity:** returns HTTP 200 with a body that violates the caller's schema

## Summary

The cache signature is computed from `{model, messages, temperature, top_p}`. `response_format` is not part of it — nor are `tools`, `tool_choice`, `max_tokens`, `seed` or `stop`. Two requests with identical messages and **different response schemas** therefore share a cache key, and the second is served the first's stored body under a 200.

## Reproduction

Both requests below have byte-identical `messages`; only the schema differs.

```bash
# A — schema {value}
curl -s $BASE/v1/chat/completions -H "Authorization: Bearer $KEY" \
  -H 'content-type: application/json' -d '{
  "model":"<gemini-model>","temperature":0,
  "messages":[{"role":"user","content":"Describe a red bicycle."}],
  "response_format":{"type":"json_schema","json_schema":{"name":"d","strict":true,"schema":{
    "type":"object","properties":{"value":{"type":"string"}},
    "required":["value"],"additionalProperties":false}}}}'

# B — same messages, schema {color, wheels}
curl -s $BASE/v1/chat/completions -H "Authorization: Bearer $KEY" \
  -H 'content-type: application/json' -d '{
  "model":"<gemini-model>","temperature":0,
  "messages":[{"role":"user","content":"Describe a red bicycle."}],
  "response_format":{"type":"json_schema","json_schema":{"name":"d","strict":true,"schema":{
    "type":"object","properties":{"color":{"type":"string"},"wheels":{"type":"integer"}},
    "required":["color","wheels"],"additionalProperties":false}}}}'
```

Observed:

```
A -> {"value":"A vibrant red bicycle featuring a glossy crimson frame..."}
B -> {"value":"A vibrant red bicycle featuring a glossy crimson frame..."}   <- A's body
```

Expected for B: `{"color":"red","wheels":2}` — which is exactly what it returns with `-H 'x-omniroute-no-cache: true'`.

`call_logs` confirms the mechanism:

```
B without bypass ->  cache_source=semantic   duration=2ms
B with    bypass ->  cache_source=upstream   duration=1768ms
```

B receives a 200 whose body has neither `color` nor `wheels`. A client validating the response sees required fields missing; a client that does not validate consumes the wrong object silently.

## Root cause

`src/lib/semanticCache.ts:135`

```ts
export function generateSignature(
  model,
  conversation,
  temperature = 0,
  topP = 1,
  apiKeyId?: string
) {
  const payload = JSON.stringify({
    model,
    messages: normalizeConversation(conversation),
    temperature,
    top_p: topP,
  });
  const digest = crypto.createHash("sha256").update(payload).digest("hex");
  return apiKeyId ? `${apiKeyId}.${digest}` : digest;
}
```

`normalizeConversation` compares role/content as exact strings, so despite the `semantic` label this is an exact-match cache, not an embedding one. That is not the problem — the problem is that everything determining the *shape* of a valid response is absent from the key.

`tools` / `tool_choice` have the same exposure: `isCacheableForWrite` does not exclude tool-calling, so a request with tools can be served the body of one without.

## Why this deserves priority over the schema-nullability bug

It is not the larger defect, but it is the one that hides the others.

It returns 200s silently, and it corrupts precisely the experiment an engineer runs to investigate anything else. Isolating one variable means holding the prompt fixed — and holding the prompt fixed is what pins the cache key. A correct methodology walks straight into it.

Concretely: while diagnosing the nullability bug (filed separately), this cost three rounds of back-and-forth between a gateway operator and a downstream consumer. Both parties independently concluded "the fix is not deployed" from what were stale cache reads. The only tell was latency — 5 ms versus 2 s — which neither was recording.

## What is already correct

- Keys are namespaced per `apiKeyId`, so there is **no cross-tenant leakage**. That was the first thing checked.
- The 30-minute TTL and the `temperature === 0` gate are reasonable.
- `x-omniroute-no-cache` works on both the read and write paths.

## Proposed fix

Fold the output contract into the digest. Concretely:

- A helper `outputContractOf(body)` collects the fields that decide the *shape* of a valid
  response — `response_format`, `text.format`, `tools`, `tool_choice` — and returns `null`
  when none are present, so plain-chat signatures (and their existing cache entries) are
  unchanged.
- `generateSignature` accepts it as an optional final argument and folds it into the hashed
  payload only when non-null.
- The three call sites pass it: the read path in `open-sse/handlers/chatCore/semanticCache.ts`
  and the two write paths (`semanticCacheStore.ts`, `streamingSemanticCacheStore.ts`). Since
  `bodyForCacheWrite` snapshots the same body object used at read time (#cache-signature-asymmetry
  guard), read/write key symmetry is preserved automatically.

### Cheaper alternative

If changing the key is undesirable, refusing to cache any request that carries a response schema also closes it. We deployed exactly that as a stopgap in front of our own instance and measured the cost: over three days the cache served **13 hits against 5390 upstream calls (0.24 %)**, because real agent traffic carries unique conversations. Skipping the cache for structured output is close to free. Fixing the key is still better, because it keeps the hits and covers `tools` too.

amirrezakm:
## Summary

- The semantic-cache signature hashed only `{model, messages, temperature, top_p}`. Two temp=0 requests with identical messages but **different `response_format`** therefore shared a cache key, and the second was served the first's stored body under a 200 — a response that violates the schema it asked for. `tools`/`tool_choice` had the same exposure. (Live repro in #12307: a request for `{color, wheels}` was served a stored `{"value": "A vibrant red bicycle…"}`.)
- `generateSignature` now accepts an optional **output contract** — `response_format`, `text.format` (Responses API spelling), `tools`, `tool_choice`, collected by a new exported `outputContractOf(body)` — and folds it into the hashed payload **only when present**. Plain-chat signatures are byte-identical to before, so every existing plain-chat cache entry survives the upgrade; only structured-output/tool requests get new keys (their old entries age out within the 30-minute TTL).
- All three call sites pass it: the read path (`chatCore/semanticCache.ts`) and both write paths (`semanticCacheStore.ts`, `streamingSemanticCacheStore.ts`). Read/write key symmetry is preserved automatically because `bodyForCacheWrite` snapshots the same body object the read path hashed (the existing `#cache-signature-asymmetry` guard).

## Related Issues

- Closes #12307
- Related to #12308 (this cache defect corrupts the experiment used to investigate that one)

## Validation

- [x] Change type: other — semantic cache (`src/lib` + chatCore decomposition modules)
- [x] Focused tests from the golden path (commands below)
- [ ] `npm run lint` — targeted `npx eslint` run on the five touched files instead; full lint left to CI
      (every reported error reproduces identically on the untouched `release/v3.8.51` base — pre-existing, none introduced by this diff; the new/changed files themselves lint clean)
- [x] Reconciled with the current active release base (branched from `release/v3.8.51` HEAD `d920e64`)

Commands run locally:

```bash
node --import tsx/esm --test tests/unit/12307-semantic-cache-output-contract.test.ts
node --import tsx/esm --test tests/unit/cache-signature-roundtrip.test.ts     # existing guard, still green
node --import tsx/esm --test tests/unit/11559-semantic-cache-ttl-expiry.test.ts  # existing guard, still green
npx eslint src/lib/semanticCache.ts open-sse/handlers/chatCore/semanticCache.ts \
  open-sse/handlers/chatCore/semanticCacheStore.ts \
  open-sse/handlers/chatCore/streamingSemanticCacheStore.ts \
  tests/unit/12307-semantic-cache-output-contract.test.ts
```

## Tests Added Or Updated

- `tests/unit/12307-semantic-cache-output-contract.test.ts` (new): different schemas ⇒ different signatures; identical schemas ⇒ shared signature; tools split the key; **plain chat keeps the legacy signature byte-for-byte** (backward compat); `outputContractOf` returns `null` for plain chat and collects the `text.format` spelling; both store paths forward the contract to `generateSignature` (via injected deps, mirroring the existing DI seams).

## Coverage Notes

- `src/lib/semanticCache.ts`: new exported helper + one parameter — covered by the six signature tests.
- The three chatCore modules: one-argument change each — covered by the two deps-injection forwarding tests.
- Existing guards `cache-signature-roundtrip` and `11559-semantic-cache-ttl-expiry` rerun green, so read/write symmetry and TTL behaviour are unchanged.

## Reviewer Notes

- **Key stability:** the `output` field is spread into the payload only when the contract is non-null, so plain-chat keys (the overwhelming majority — on the instance where this was found, the cache served 13 hits against 5390 upstream calls over three days) are unchanged and their entries stay valid. Structured-output entries written by the old code become unreachable and expire via TTL; that is the point of the fix.
- **Key-order caveat:** `response_format`/`tools` are folded in as sent, so two clients serialising the same schema with different key order will miss each other's entries. That is a miss, not a collision — and it mirrors how `messages` content objects are already treated.
- **Alternative considered:** refusing to cache schema-carrying requests entirely (measured cost ≈ 0.24 % hit-rate loss). Folding the contract into the key was chosen because it keeps those hits and covers `tools` uniformly.

### DIFF
--- a/open-sse/handlers/chatCore/semanticCache.ts
+++ b/open-sse/handlers/chatCore/semanticCache.ts
@@ -2,6 +2,7 @@ import {
   generateSignature,
   getCachedResponse,
   isCacheableForRead,
+  outputContractOf,
 } from "@/lib/semanticCache";
 import { calculateCost } from "@/lib/usage/costCalculator";
 import { trackPendingRequest } from "@/lib/usageDb";
@@ -51,7 +52,8 @@ export async function checkSemanticCache({
       body.messages ?? body.input,
       body.temperature,
       body.top_p,
-      apiKeyId ?? undefined
+      apiKeyId ?? undefined,
+      outputContractOf(body)
     );
     const cached = getCachedResponse(signature);
     if (cached) {
--- a/open-sse/handlers/chatCore/semanticCacheStore.ts
+++ b/open-sse/handlers/chatCore/semanticCacheStore.ts
@@ -10,6 +10,7 @@
  */
 import {
   generateSignature as defaultGenerateSignature,
+  outputContractOf,
   setCachedResponse as defaultSetCachedResponse,
   isCacheableForWrite as defaultIsCacheableForWrite,
 } from "@/lib/semanticCache";
@@ -65,7 +66,8 @@ export function storeSemanticCacheResponse(
     args.body.messages ?? args.body.input,
     args.body.temperature,
     args.body.top_p,
-    args.apiKeyId ?? undefined
+    args.apiKeyId ?? undefined,
+    outputContractOf(args.body)
   );
   const tokensSaved = args.usage?.prompt_tokens + args.usage?.completion_tokens || 0;
   deps.setCachedResponse(signature, args.model, args.translatedResponse, tokensSaved);
--- a/open-sse/handlers/chatCore/streamingSemanticCacheStore.ts
+++ b/open-sse/handlers/chatCore/streamingSemanticCacheStore.ts
@@ -11,6 +11,7 @@
  */
 import {
   generateSignature as defaultGenerateSignature,
+  outputContractOf,
   setCachedResponse as defaultSetCachedResponse,
   isCacheableForWrite as defaultIsCacheableForWrite,
 } from "@/lib/semanticCache";
@@ -69,7 +70,8 @@ function writeStreamingCacheEntry(
       args.body.messages ?? args.body.input,
       args.body.temperature,
       args.body.top_p,
-      args.apiKeyId ?? undefined
+      args.apiKeyId ?? undefined,
+      outputContractOf(args.body)
     );
     const tokensSaved = streamTokensSaved(args.streamUsage);
     deps.setCachedResponse(sig, args.model, cleanBody, tokensSaved);
--- a/src/lib/semanticCache.ts
+++ b/src/lib/semanticCache.ts
@@ -6,7 +6,8 @@
  * are cached after assembly; cache hits always return JSON.
  * Two-tier: in-memory LRU (fast) + SQLite (persistent across restarts).
  *
- * Cache key = SHA-256(model + normalized messages + temperature + top_p)
+ * Cache key = SHA-256(model + normalized messages + temperature + top_p
+ *             + output contract, when present — see outputContractOf, #12307)
  * Bypass: X-OmniRoute-No-Cache: true
  *
  * @module lib/semanticCache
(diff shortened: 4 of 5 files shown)


diegosouzapw:
Real, well-documented bug with a live repro (#12307) — the semantic-cache signature ignored
`response_format`/`tools`/`tool_choice`, so a `temp=0` request could be served a stored body
from an earlier request with a different output schema under a 200. The fix folds an optional
output contract into the hashed payload only when present, so plain-chat keys stay byte-identical
(every existing cache entry survives) and only structured-output/tool requests get new keys. Ran
the full test suite at this PR's head: 14/14 pass, including the 8 new cases and the existing
TTL/roundtrip guards with no regressions. Small, clean, well-architected change — please add the
changelog fragment mentioned in the description (it isn't in the diff) and rebase against the
current release tip (GitHub shows a merge conflict, likely just proximity with #12910/#12912 in
the same cache/chatCore area).

diegosouzapw:
Thanks @amirrezakm — merging via the release merge-train. Validated in local merge-train /tmp/mt-train3b.log on 192.168.0.113 @ train tip 408e2128791696a18966681953136fc96aeb99b0 (32 PRs boarded): static gates green; full test:unit 40694 tests, 17 failing — every one reproduces on the pure release tip (base-red sweep list), zero new reds. Merged --admin per merge-gates §4/§7.

PR #12325, opened 2026-09-01
fix(quota): drop generic quota cache on upstream 429

HouMinXi:
fix(quota): drop generic quota cache on upstream 429

### Summary
Codex already invalidates its quota cache on a 429 (`invalidateCodexQuotaCache` in `chatCore`). The generic fetcher used by agy / Antigravity / Claude OAuth had `invalidateGenericQuotaCache` but nothing called it. After a 429, reset-aware scoring could keep using a 60s stale snapshot while `accountFallback` cooled the connection.

A second layer made a naive cache-delete insufficient: agy `retrieveUserQuota` and weekly summary caches are 60s–5min. Dropping only the wrapper recaches the same inner snapshot. Invalidation now also force-refreshes the next usage fetch.

Probe-origin 429s are ignored (`#9817`). Codex keeps its own path.

### Tests
- 60s cache hit still one upstream usage call
- `invalidateGenericQuotaCache` recaches with `forceRefresh: true`
- `invalidateGenericQuotaCacheOnStatus`: 200 no-op, 429 drops, probe 429 no-op, missing id no throw
- chatCore wires `invalidateGenericQuotaCacheOnStatus` on non-Codex 429
- sibling agy 429 cooldown + universal quota-aware routing suites pass (58/58)

### Not in this PR
- Exposing generic cache TTL in the dashboard (already rejected: the existing Reset-aware quota cache knob is the outer layer, default 0)
- Changing Codex 429 handling
- Invalidating agy subscription / credit-probe caches beyond `forceRefresh` on the usage path


HouMinXi:
Rebased onto current `release/v3.8.51` tip `f41a9bd83`.

- old unique range: `438db55c4..25226f7b3` (6 commits)
- new unique range: `f41a9bd83..d49d4a127` plus `eeba9966c`
- `git range-diff` on the 6 original commits → all `=`
- GPG `G` on every rewritten commit
- dropped the maintainer merge `255ac98cb` (onto `158647618`)
- extra commit `eeba9966c`: drop unused `third` binding that CI eslint (`@typescript-eslint/no-unused-vars`) failed on (`tests/unit/generic-quota-fetcher.test.ts:292`). The call still runs; the return was unused.
- focused tests: generic-quota-fetcher + antigravity-429-quota-cooldown 29/29 pass

No production-logic change beyond the unused-binding lint fix.

HouMinXi:
Rebased onto `release/v3.8.51` @ `2c4ad3e55`. Dropped the two merge commits that were only syncing the old tip.

File-size conflict on `chatCore.ts`: kept upstream's rebaseline notes and this PR's `#12325` note. Frozen LOC is the measured `split("\\n").length` of the merged file (5981). Head is now `17ee1e14f`.

Issue #12146, opened 2026-08-30
chore(lint): resolve the 229 ESLint violations frozen in #12144 (react-hooks 7 compiler rules) in batches by dashboard area


diegosouzapw:
## What

Round 5b of the base-red sweep (#12144) froze **229** ESLint violations with `eslint --suppress-all` so the release branch could go green during the velocity phase. They were invisible until `dfc84ba030` regenerated the lockfile and invalidated the ESLint cache — the CI job only ever linted changed files against a warm cache, so the `eslint-plugin-react-hooks` 7 compiler rules (pinned in #11886) never ran on the whole tree.

This issue is the map for **resolving** them instead of keeping the freeze, in batches by dashboard area (one green PR each — not one 100-file PR).

## By rule

| rule | count | what a real fix looks like |
|---|---:|---|
| `react-hooks/set-state-in-effect` | 167 | derive state during render, move the write into the event handler / data loader, or useSyncExternalStore; a fetch-then-setState effect becomes a data hook |
| `react-hooks/immutability` | 36 | stop mutating captured objects/arrays; clone or lift into state |
| `react-hooks/refs` | 7 | read refs in handlers/effects, never during render |
| `react-hooks/static-components` | 7 | hoist components defined inside other components |
| `react-hooks/purity` | 5 | move Date.now()/Math.random() out of render (effect, handler, or useMemo with a stable seed) |
| `react-hooks/preserve-manual-memoization` | 4 | align useMemo/useCallback deps so the compiler can keep them |
| `@typescript-eslint/no-explicit-any` | 3 | type the fixtures (tests/unit/call-log-cap.test.ts — mechanical) |

## By area (files → violations) — suggested batches

| batch | area | files | violations |
|---|---|---:|---:|
| 1 | `dashboard/cli-code` | 12 | 42 |
| 2 | `dashboard/providers` | 18 | 28 |
| 3 | `dashboard/settings` | 17 | 25 |
| 4 | `shared/components` | 11 | 20 |
| 5 | `dashboard/combos` | 2 | 10 |
| 6 | `dashboard/endpoint` | 5 | 8 |
| 7 | `dashboard/provider-stats` | 1 | 8 |
| 8 | `dashboard/api-manager` | 1 | 7 |
| 9 | `dashboard/costs` | 5 | 7 |
| 10 | `dashboard/radar` | 4 | 6 |
| 11 | `dashboard/usage` | 4 | 6 |
[...]

PR #9962, opened 2026-08-09
fix(build): bump @huggingface/transformers to 4.2.0 + onnxruntime-node 1.24.3

witt3rd:
placeholder

PR #12329, opened 2026-09-01
chore(deps): freeze the paired/pinned deps out of dependabot groups

diegosouzapw:
Both open dependabot group PRs are born red on deliberate contract tests:

- **#12219 (production group)**: `onnxruntime-node 1.24.3→1.29.0` violates `tests/unit/onnxruntime-single-copy.test.ts` — the hoisted copy must equal the exact version `@huggingface/transformers` pins (pair from #9962; transformers itself is already frozen in this file). The pair only ever moves together, in a deliberate migration PR.
- **#12220 (development group)**: `eslint-plugin-react-hooks 7.0.1→7.1.1` violates the pin contract (`tests/unit/eslint-react-hooks-version-pinned.test.ts`). The #12146 migration finished today, so adopting 7.1.1 is now viable — but in its own PR with a full cold lint run, removing the pin test in the same move; never riding a group.

This adds the two ignores (with rationale comments) so the groups recreate clean. After merge, `@dependabot recreate` on both PRs drops the offending packages and the remaining 14+11 healthy bumps can land.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -58,6 +58,22 @@ updates:
       # on the VPS — so keep auto-bumps frozen (no update-types = ignore every version).
       # Migrate it intentionally, not via dependabot (#4050).
       - dependency-name: "@huggingface/transformers"
+      # onnxruntime-node is the OTHER HALF of the @huggingface/transformers pair frozen
+      # above: the hoisted copy must equal the exact version transformers pins, or npm
+      # nests a second ABI-incompatible native copy (contract test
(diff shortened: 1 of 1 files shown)


diegosouzapw:
Merging with `--admin`: the 4 red checks on this PR's merge-ref are the 2026-09-01 base-red window (stream-utils / settings-i18n-keys / placeholder-fallback / dead collectSSE — all reproduced on the pure base tip, fix in flight in #12327). This diff is a single dependabot config file no test exercises; none of the reds can be caused by it.

Issue #8887, opened 2026-07-29
fix: provider reconnect leaves combos pinned to stale connectionId and LKGP cache


bluewatercg:
### OmniRoute Version

 3.8.48

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

  Ubuntu on a local Docker host

### Node.js Version

 24.x

### Provider(s) Involved

  ShuaiAPI, BerryCode

### Model(s) Involved

 gpt-5.6-sol

### Client Tool

 cURL and OmniRoute dashboard

### Description

 After updating a provider key, the provider-level test passes, but existing combos can still keep using a stale connectionId or cached pinned model. In this case, gpt56-sol continued
  routing to berry/gpt-5.6-sol even after the provider was updated and ShuaiAPI tests passed. The combo had to be manually repaired before traffic switched to ShuaiAPI.

### Steps to Reproduce

 1. Update a provider key.
  2. Confirm the provider test passes.
  3. Open an existing combo that uses that provider.
  4. Run a request through the combo.
  5. Observe that routing still follows the old pinned model or stale connection.

### Expected Behavior

  Updating/reconnecting a provider should refresh combo bindings, invalidate stale LKGP/session pin state, and route through the updated provider.

### Actual Behavior

The combo still routes through the old provider/model path. Logs show a server-side pinned model from the previous provider, then fallback behavior or stale routing until the combo
  metadata/cache is manually refreshed.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Context cache: pinned model=berry/gpt-5.6-sol (server-side)
  Context-cache pin "berry/gpt-5.6-sol" provider durably unhealthy — dropping pin, using strategy
  Trying model 1/2: berry/gpt-5.6-sol
  ...
  gpt56-sol continued routing to berry/gpt-5.6-sol until combo metadata and cache were repaired manually.
```

### Additional Context

  This looks like a state consistency bug, not a provider outage. Provider test success does not guarantee combos have refreshed their stored connectionId or LKGP cache entry.

### Validation Plan

 - Reconnect/update a provider key.
  - Verify existing combos immediately route to the new provider.
  - Verify no stale berry/... pin remains in logs.
  - Verify a live request hits the new provider after reconnect without manual combo edits.

xiaoyaner0201:
[...]

PR #12330, opened 2026-09-01
fix(combos): clear LKGP pins when a combo is deleted

### LINKED ISSUE #12326: fix(backend): deleteCombo() leaves orphaned LKGP entries in key_value
## Summary

`deleteCombo()` removes the row from the `combos` table but never clears the combo's associated LKGP (last-known-good-provider) entries in `key_value`. The rows are keyed by combo name, so after a combo is deleted its cache entries stay in the database forever with no way to reach them through the UI or API.

## Version

- OmniRoute **v3.8.50** (current latest release, published 2026-08-26)
- SQLite storage backend

## Source

`src/lib/db/repositories/sqliteComboRepository.ts:345`

```ts
export async function deleteCombo(id: string) {
  const db = getDbInstance();
  const result = db.prepare("DELETE FROM combos WHERE id = ?").run(id);
  if (result.changes === 0) return false;
  return true;
}
```

Only the `combos` row is deleted. `key_value` is never touched — the file does not reference `key_value` at all.

## Why this looks like an oversight rather than a design decision

`src/lib/db/settings/lkgp.ts` already provides the cleanup primitives, and the codebase already applies this discipline to a *different* delete path:

```ts
// src/lib/db/settings/lkgp.ts:62
export async function clearLKGP(comboName: string, modelId: string): Promise<void>

// src/lib/db/settings/lkgp.ts:74-78
/**
 * ... row that no longer exists; provider-connection delete paths call this so
 * ...
 */
export async function deleteLKGPByConnectionIds(connectionIds: string[]): Promise<number>
```

So provider-connection deletion cleans up its LKGP rows, but combo deletion does not — even though `clearLKGP()` is already keyed by combo name and would be a direct fit.

## Reproduction

1. Create a combo, e.g. `my-temp-combo`.
2. Send one request that routes through it (this writes the LKGP entry).
3. Delete the combo via the dashboard or `DELETE /api/combos/{id}`.
4. Inspect the database:

```sql
SELECT namespace, key, value
FROM key_value
WHERE namespace = 'lkgp' AND key LIKE 'my-temp-combo%';
```

**Expected:** no rows.
**Actual:** the entries are still present, e.g.

```
lkgp | my-temp-combo:my-temp-combo-model-1-<provider>-<model> | {"provider":"<provider>"}
lkgp | my-temp-combo:<uuid>                                   | {"provider":"<provider>"}
```

`GET /api/combos/my-temp-combo` correctly returns 404 at this point, confirming the combo itself is gone while its cache entries remain.

## Observed in practice

On an instance with 26 live combos, the `lkgp` namespace held 54 entries — 52 belonging to live combos plus 2 orphans from a combo that had been deleted earlier the same day. Each deleted combo leaves roughly 2 rows behind.

## Impact

- Unbounded growth of the `key_value` table for anyone who creates and deletes combos regularly (temporary/experimental combos, automated provisioning, CI).
- The rows are unreachable through any supported surface: the combo 404s, so no API or UI path can clear them. `clearAllLKGP()` exists but wipes the namespace for *all* combos, which is not an acceptable targeted cleanup.
- Operators auditing routing state see stale provider pins for combos that no longer exist, which is confusing when diagnosing routing behaviour.

No functional routing breakage was observed — the entries are simply never read again once the combo is gone.

## Suggested fix

Resolve the combo name before deleting the row and clear its LKGP entries in the same transaction, mirroring what the provider-connection delete path already does:

```ts
export async function deleteCombo(id: string) {
  const db = getDbInstance();
  const combo = db.prepare("SELECT name FROM combos WHERE id = ?").get(id) as { name?: string } | undefined;
  const result = db.prepare("DELETE FROM combos WHERE id = ?").run(id);
  if (result.changes === 0) return false;
  if (combo?.name) {
    db.prepare("DELETE FROM key_value WHERE namespace = 'lkgp' AND key LIKE ?").run(`${combo.name}:%`);
  }
  return true;
}
```

Happy to open a PR if the approach looks right.

xiaoyaner0201:
Closes #12326.

## Problem

`deleteCombo()` deleted the `combos` row but left the combo's LKGP pins in `key_value`. Pins are keyed by `${comboName}:${modelId}`, so once the combo is gone those rows become unreachable:

- the combo itself 404s, so no API or UI path addresses them
- `clearLKGP()` needs a `modelId` the caller no longer knows
- `clearAllLKGP()` wipes the namespace for *every* combo, which is not a targeted cleanup

On an instance with 26 live combos the `lkgp` namespace held 54 rows — 52 for live combos plus 2 orphans from a combo deleted earlier that day. Roughly 2 rows leak per deleted combo, so the table grows without bound for anyone who creates and deletes combos regularly.

## Fix

Provider-connection deletion already handles this via `deleteLKGPByConnectionIds()` (#8887), and `providers/deletion.ts` wraps it in try/catch so cleanup failures never block the delete. This applies the same pattern to combo deletion:

- **`deleteLKGPByComboName(comboName)`** in `settings/lkgp.ts` — deletes every pin under the `${comboName}:` prefix and calls `invalidateCachedLKGP()` for each, so a stale pin cannot be served from the 5s read cache for the rest of its TTL.
- **`deleteCombo()`** resolves the combo name before deleting the row, then clears its pins inside try/catch — a cleanup failure is logged, never fatal to the delete.

The `:` delimiter in the prefix keeps sibling combos safe (`prod` does not match `prod-canary`), and `LIKE` wildcards in combo names are escaped so a name like `temp_a` cannot match `tempXa`.

## Tests

New `tests/unit/combo-delete-lkgp-cleanup-12326.test.ts`, structured after the #8887 suite:

| Test | Covers |
|---|---|
| deleting a combo removes its LKGP pins | core fix, survivor pins untouched |
| invalidates warmed read-cache entries | the `invalidateCachedLKGP()` path |
| prefix-sharing sibling keeps its pins | `prod` vs `prod-canary` |
| `LIKE` wildcards do not widen cleanup | `temp_a` vs `tempXa` |
| unknown combo id leaves state untouched | early-return branch |
| combo without pins deletes cleanly | no-op path |

Verified the tests actually exercise the fix — with the cleanup block removed, 4 of 6 fail; with it restored, all 6 pass.

```
$ npx tsx --test tests/unit/combo-delete-lkgp-cleanup-12326.test.ts
# tests 6
# pass 6
# fail 0
```

No regressions in the existing LKGP suites:

```
$ npx tsx --test tests/unit/delete-provider-connection-invalidates-lkgp-8887.test.ts \
    tests/unit/lkgp-enabled-context-11181.test.ts \
    tests/unit/lkgp-stale-pin-exhaustion-11911.test.ts \
    tests/unit/combo-delete-lkgp-cleanup-12326.test.ts
# tests 18
# pass 18
# fail 0
```

`eslint` and `prettier --check` are clean on the touched files; `tsc --noEmit` reports no errors in them (the repo's existing baseline errors are unrelated and untouched).

## Notes

- Changelog fragment added at `changelog.d/fixes/12330-combo-delete-lkgp-cleanup.md`.
- No behaviour change for combos that are never deleted; the new query only runs on the delete path.


xiaoyaner0201:
The three red checks on this PR are pre-existing base reds on `release/v3.8.51`, not regressions from this change.

**`No new ESLint warnings`** — the three errors are all in `tests/unit/stream-passthrough-usage-estimation.test.ts`, which this PR does not touch. Reproducible on a clean checkout of the base branch:

```
$ git checkout upstream/release/v3.8.51
$ npx eslint tests/unit/stream-passthrough-usage-estimation.test.ts --max-warnings 0
  37:10  error  'collectSSE' is defined but never used
  37:21  error  'stream' is defined but never used
  38:17  error  'writable' is defined but never used
✖ 3 problems (3 errors, 0 warnings)          # exit 1
```

**`Unit Tests fast-path`** — the only failing assertion in the shard logs is `createSSEStream passthrough forwards OpenAI usage-only empty choices chunks`, from the same file.

Both are already being addressed by #12327 ("drain the 2026-09-01 base-red window — passthrough usage regression") and #12331, which touch that same test file.

For this PR's own scope:

```
$ npx eslint src/lib/db/settings/lkgp.ts \
    src/lib/db/repositories/sqliteComboRepository.ts \
    src/lib/db/settings.ts \
    tests/unit/combo-delete-lkgp-cleanup-12326.test.ts --max-warnings 0
# exit 0, no output

$ DISABLE_SQLITE_AUTO_BACKUP=true node --import tsx/esm \
    --test tests/unit/combo-delete-lkgp-cleanup-12326.test.ts
# tests 6
# pass 6
# fail 0
```

Happy to rebase once the base-red PRs land if that makes the run easier to read.

diegosouzapw:
Thanks for the PR! Looking at the current `release/v3.8.51` tip, this exact fix already
shipped via #12425 ("fix(combos): clear LKGP pins on delete", merged 2026-09-04) —
`deleteCombo()` in `src/lib/db/repositories/sqliteComboRepository.ts` already calls
`deleteLKGPRowsByComboName()` + invalidates the read cache, and the tip even carries a
test file with the same name you added
(`tests/unit/combo-delete-lkgp-cleanup-12326.test.ts`) covering the same scenarios
(sibling-combo isolation, LIKE-wildcard escaping, cache invalidation). Since #12326 is
already fixed on the branch you're targeting, I'm closing this one as subsumed by #12425
— appreciate you tackling it independently, the approach itself was solid.

Issue #5460, opened 2026-06-29
[FIX] Reka — model import fails with 'local catalog fallback not synced' + wrong default name + empty validation model ID


chirag127:
### 1. Model import fails with no fallback

\`\`\`
Failed to fetch models
Remote model discovery failed; local catalog fallback not synced
\`\`\`

Worse than the usual cached-catalog fallback — no local catalog exists for Reka either. User gets nothing. Reka's model list endpoint: `https://api.reka.ai/v1/models`, auth: `X-Api-Key: <key>`.

Note: Reka uses `X-Api-Key` header, **not** `Authorization: Bearer`. If the fetch probe uses Bearer auth it will 401 on `/v1/models` even with a valid key.

## Acceptance criteria

- [ ] Model fetch uses correct `X-Api-Key: <key>` header for Reka
- [ ] Models import successfully on a valid key
- [ ] If remote fetch fails, show "cached catalog" warning (not "not synced" error)
- [ ] Default key name pre-filled as `main`
- [ ] Validation Model Id pre-filled with `reka-flash`

diegosouzapw:
Thanks for the detailed report! Fixed in `release/v3.8.43` via #5787 (already merged into the active development branch — feel free to pull and test now). It will ship in the next release.

PR #12332, opened 2026-09-01
fix(providers): Vertex Express-key discovery failure is an intentional local-catalog fallback, not degraded (#12328)

### LINKED ISSUE #12328: fix(providers): Vertex model discovery sends Express API key to generativelanguage.googleapis.com — always fails (400 API_KEY_INVALID), UI masks it as 'API unavailable'
## Summary

Model discovery for the **Vertex** provider with a **Vertex AI Express API key** (`auth_type=apikey`, opaque string) always fails, because the discovery branch sends the key to the **Generative Language API** — a different Google service that does not accept Vertex keys. The UI then silently falls back to the local catalog with a generic "API unavailable" message, which makes it look like a key/permission problem on the user's side. It isn't: the same key works perfectly for inference through OmniRoute's own Vertex executor.

## Environment

- OmniRoute `3.8.50` (docker image digest `dbf98ed1efcf…`, digest-pinned)
- Provider: `vertex`, auth type: API key (Vertex AI Express mode, created in Cloud Console)
- Region: `us-central1`

## Steps to reproduce

1. Add a Vertex connection with a **Vertex AI Express API key** (not a Service Account JSON, not an AI Studio key).
2. Connection test passes: `POST /api/providers/{id}/test` → `{"valid":true,…}`.
3. Trigger model discovery / "Importar Modelos" (`GET /api/providers/{id}/models`).
4. Observe: the import modal reports `API unavailable – using local catalog` / "no new models to import", and the container logs:

```
[models] Vertex model discovery failed { provider: 'vertex', status: 403 }
```

## Root cause

`src/app/api/providers/[id]/models/route.ts` (line ~1856 on current `main`), vertex branch:

```ts
// Vertex AI lists models from the Generative Language `v1beta/models` endpoint, which both
// Express-mode API keys (via ?key=) and Service Account JSON ... can reach.
...
const baseUrl = "https://generativelanguage.googleapis.com/v1beta/models?pageSize=1000";
...
let pageUrl = queryKey ? `${baseUrl}&key=${encodeURIComponent(queryKey)}` : baseUrl;
```

The comment's assumption is inverted for Express keys: a Vertex Express key is **not** valid on `generativelanguage.googleapis.com`. Reproduced with a plain curl using exactly the request the discovery branch builds (`?key=<vertex express key>`):

```json
{
  "error": {
    "code": 400,
    "message": "API key not valid. Please pass a valid API key.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "API_KEY_INVALID",
        "domain": "googleapis.com",
        "metadata": { "service": "generativelanguage.googleapis.com" }
      }
    ]
  }
}
```

The SA-JSON path (OAuth bearer) is unaffected; only the Express-key (`queryKey`) path is broken.

## Why this is misleading (the actual UX bug)

- The connection test says the key is **valid** (it is — for `aiplatform.googleapis.com`).
- Inference through the executor works for **any** model id, including ones absent from the local catalog (verified live E2E with `gemini-3.7-flash` / `gemini-3.5-flash-lite`, which the static registry doesn't list yet).
- Discovery then fails with a generic "API unavailable – using local catalog", so users chase key permissions/regional IAM ghosts while the real cause is the endpoint choice.

Net effect: the live catalog is never fetched, and the local catalog lags behind what the key can actually serve.

## Suggested fix

For the Express-key (`queryKey`) case, at minimum:

1. **Don't route Vertex Express keys to `generativelanguage.googleapis.com`.** Since Vertex has no public list-all endpoint for Express keys, discovery could instead probe a curated candidate list via the publisher inference endpoint (`…/v1/publishers/google/models/{id}` with `x-goog-api-key`), or simply skip the remote fetch.
2. **Surface the real upstream error** (status + `reason: API_KEY_INVALID`, service `generativelanguage`) instead of the generic `API unavailable` warning, so the modal distinguishes "endpoint not applicable to this credential type" from "network/auth failure".

## Workaround (verified on 3.8.50)

Register the models as custom models — they're merged into every catalog read (#6247) and route normally in combos:

```
POST /api/provider-models {"provider":"vertex","modelId":"gemini-3.7-flash","modelName":"Gemini 3.7 Flash (Vertex)","source":"manual"}
```

After that, `{"model":"gemini-3.7-flash","providerId":"vertex",…}` in a combo serves traffic end-to-end (verified with text and multimodal probes).

BenedettiLucca:
## What

Fixes #12328.

Vertex model discovery with a **Vertex AI Express API key** (`?key=` on `generativelanguage.googleapis.com/v1beta/models`) always fails with `400 API_KEY_INVALID` — the Generative Language API is a different service and does not accept Vertex-scoped keys. Today that failure falls back to the generic `API unavailable — using local catalog`, which reads as a transient/auth problem and sends users chasing key/permission ghosts, while the connection test and inference both succeed.

## Change

On discovery failure with `400/403` **and** an Express key (`queryKey` path, non-JSON credential):

- the fallback warning now names the real cause (`Vertex Express API key cannot list models on the Generative Language API (API_KEY_INVALID) — …`);
- the local catalog is served as an **intentional** source (`localIntentional`, #5460/#5465) so model-sync imports it instead of treating the response as a degraded remote fetch.

Everything else is untouched:

- SA-JSON (Bearer) discovery unchanged — live listing still parsed when it succeeds;
- transient errors (500, network, redirect-block) keep the existing generic fallback;
- the fetch attempt itself still happens (best effort; no capability removed for keys that *do* work on that endpoint).

## Test plan

New regression suite `tests/unit/vertex-express-discovery-12328.test.ts` (4 tests, real tmpdir DB + mocked `fetch`, same pattern as `7993-noauth-proxy-routing`):

1. Express key + 400 → 200, `source: local_catalog`, `intentional: true`, warning matches `API_KEY_INVALID`; exactly one call to the GL endpoint with `key=` in query;
2. Express key + 403 → same intentional fallback;
3. Express key + 500 → generic fallback, **not** intentional, no `API_KEY_INVALID` in warning (no behavior change);
4. SA-JSON (real RSA key generated in-test; `importPKCS8` rejects placeholders) → OAuth token mocked, live GL list parsed, `source: api`, no `?key=` on the wire.

Local verification on `release/v3.8.51`:

```
node --import tsx/esm --test tests/unit/vertex-express-discovery-12328.test.ts   # 4/4 pass
npm run typecheck:core                                                            # clean
npx eslint <touched files>                                                        # 0 new errors (9 pre-existing in route.ts)
node --import tsx/esm --test tests/integration/model-catalog-responsiveness-9199.test.ts  # 2/2
node --import tsx/esm --test tests/integration/provider-journey.contract.test.ts          # 14 tests / 0 fail
```

### DIFF
--- a/src/app/api/providers/[id]/models/route.ts
+++ b/src/app/api/providers/[id]/models/route.ts
@@ -1745,10 +1745,12 @@ export async function GET(
       const autoFetchDisabledResponse = maybeReturnAutoFetchDisabled();
       if (autoFetchDisabledResponse) return autoFetchDisabledResponse;
 
-      // Vertex AI lists models from the Generative Language `v1beta/models` endpoint, which both
-      // Express-mode API keys (via ?key=) and Service Account JSON (via a minted OAuth Bearer
-      // token) can reach. This surfaces the live catalog, including gemini-*-image models
-      // absent from the static registry list.
+      // Vertex AI lists models from the Generative Language `v1beta/models` endpoint, which
+      // Service Account JSON (via a minted OAuth Bearer token) can reach. Express-mode API
+      // keys are still sent via ?key= as a best effort, but the Generative Language API
+      // rejects Vertex-scoped keys (400 API_KEY_INVALID — #12328); that failure path serves
+      // the local catalog as an intentional source with a warning naming the real cause.
+      // Live listing surfaces models absent from the static registry, e.g. gemini-*-image.
       const credential = (apiKey || "").trim();
       let queryKey: string | null = null;
       let bearerToken: string | null = null;
@@ -1823,6 +1825,29 @@ export async function GET(
               provider,
               status: response.status,
             });
+            // #12328 — a Vertex-scoped Express key cannot list models on the Generative
+            // Language API (400 API_KEY_INVALID). That is a credential/endpoint mismatch,
+            // not a transient upstream failure, so serving the local catalog here is the
+            // intended behavior, not a degraded state: inference still routes any model id
+            // through the aiplatform executor. Surface the real cause instead of the
+            // generic "API unavailable" so the UI/modal can distinguish the two.
+            if (queryKey && (response.status === 400 || response.status === 403)) {
+              const fallback = buildDiscoveryFallbackResponse({
+                cacheWarning:
+                  "Vertex Express API key cannot list models on the Generative Language API (API_KEY_INVALID) — using cached catalog",
+                localWarning:
+                  "Vertex Express API key cannot list models on the Generative Language API (API_KEY_INVALID) — using local catalog",
+                localIntentional: true,
+              });
+              if (fallback) return fallback;
+              return NextResponse.json(
+                {
+                  error:
+                    "Vertex Express API keys are not accepted by the Generative Language models endpoint. Use a Service Account JSON for live discovery, or add models as custom models.",
+                },
+                { status: 400 }
+              );
+            }
             const fallback = buildDiscoveryFallbackResponse();
             if (fallback) return fallback;
             return NextResponse.json(
--- a/tests/unit/vertex-express-discovery-12328.test.ts
+++ b/tests/unit/vertex-express-discovery-12328.test.ts
@@ -1,0 +1,209 @@
+/**
+ * #12328 — Vertex model discovery with a Vertex AI Express API key must not
+ * surface the generic "API unavailable — using local catalog" when the
+ * Generative Language API rejects the key. A Vertex-scoped Express key gets
+ * 400 API_KEY_INVALID from generativelanguage.googleapis.com (different
+ * service, different credential domain), while the connection test and
+ * inference both succeed — so the generic warning sends users chasing
+ * key/permission ghosts.
+ *
+ * The fix: on 400/403 with an Express key (queryKey path), the fallback names
+ * the real cause and serves the local catalog as an intentional source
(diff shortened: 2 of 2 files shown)


diegosouzapw:
Clean, surgical fix — exactly the right diagnosis (Express keys can't list on the Generative
Language API, that's expected, not degraded) and the test suite covers the 400/403/500
distinction precisely. Local probe: 4/4 tests pass. Ready to merge as-is. Heads up: #12471
(also in this batch) rewrites this same code path as part of a larger Vertex discovery
refactor and already builds in equivalent "intentional" behavior — merging this one first
means #12471 just needs a routine rebase, not a logic reconciliation.

diegosouzapw:
Thank you for this PR, @BenedettiLucca — the intentional-fallback behaviour for Vertex Express discovery (#12328) landed on `release/v3.8.51` through the extraction into `src/app/api/providers/[id]/models/vertexDiscovery.ts`: `handleVertexApiKeyCatalog()` now returns `source: "local_catalog"`, `intentional: true` and a warning naming the real cause. Two details diverged from this branch (the key travels in the `x-goog-api-key` header rather than the URL, and the Express path is always intentional), which is why this PR could not be re-synced mechanically. Closing as superseded; if you spot a gap the current implementation leaves open, a focused re-open is very welcome.

PR #12338, opened 2026-09-01
fix(combo): universal-handoff fixes — bare-fallback note, same-request scoping, silent-failure logging

hartmark:
## Summary

Three related fixes to the universal-handoff mechanism, found and root-caused live while investigating a real production report: a handoff reply that carried no actual context and, separately, one that fabricated an entirely invented response.

Originally scoped narrower; consolidated here (replacing #12227, which is now closed as superseded — see that PR for the initial, incomplete Spanish-note fix this supersedes).

## What was actually happening

Traced live: `context_handoffs` had **zero rows** despite handoffs firing repeatedly over hours, and separately a same-request fallback landed on a weak free-tier model that fabricated a plausible-sounding but entirely invented response about an unrelated topic.

### Fix 1 — same-request fallback was never a genuine handoff

The universal-handoff mechanism compared **every** combo target attempt — including same-request fallback steps (`i > 0`, invisible to the client since the primary target's failure is never exposed) — against the *session's previous-turn* model. A same-request fallback landing on a different model than last turn wrongly counted as a "genuine handoff," injecting a system-role note and kicking off a background summary call for a switch the client never actually experienced.

An earlier attempt at exactly this fix (commit `ffbd0699`, part of the now-closed #12227) was reverted because it broke `tests/unit/combo-context-relay.test.ts`, which asserted the *old* behavior was correct. That test's premise conflated same-request fallback with genuine cross-turn provider switches — fixed here with the corrected, narrower contract: a same-request fallback target now gets the original request completely untouched, gated on `i === 0` at both call sites (injection + background-summary trigger). `recordSessionModelUsage` bookkeeping stays unconditional, since it must reflect whichever model actually served the response regardless of `i`.

### Fix 2 — the bare-fallback note claimed continuity it didn't provide

When no prior summary exists (e.g. the very first handoff for a session, or generation failed), the fallback note said *"The conversation continues below without losing context"* — while the actual request carried only trimmed input (e.g. a bare tool result), no real history. A weak fallback model handed that note, plus almost no real context, filled the gap by inventing content instead of behaving conservatively. The note now says the opposite: no summary is available, the input below is the entire context available, do not invent details about history you cannot see.

### Fix 3 — silent failures gave no diagnostic trail

Both handoff-summary generators (`generateHandoffAsync` / `generateUniversalHandoffAsync`) had three failure returns — empty history, non-`ok` summary-model response, unparseable model output — with zero logging anywhere. This is how `context_handoffs` stayed empty for hours with no way to tell why. Added a shared logger on all three paths in both functions.

## User Impact

- Same-request fallback targets now receive the client's original request unmodified — no more context-free system notes injected for internal retries the client never saw fail.
- The bare-fallback note (when it does legitimately fire, on a genuine cross-turn switch with no summary yet) no longer invites fabrication.
- Silent, unexplained empty `context_handoffs` becomes traceable in logs.

## Validation

- [x] `tests/unit/universal-handoff.test.ts`, `tests/unit/context-handoff.test.ts`, `tests/unit/combo-context-relay.test.ts`, `tests/unit/service-context-handoff.test.ts`, `tests/unit/db-context-handoffs.test.ts`, `tests/unit/chat-context-relay.test.ts`, `tests/integration/combo-matrix/context-relay-handoff.test.ts` — all green, including the rewritten `combo-context-relay.test.ts` case now asserting the corrected same-request-fallback contract.
- [x] Complexity ratchet: no new violations.
- [x] Reconciled with `upstream/release/v3.8.51` tip at branch time.
- [x] Live-verified on a real deployment: confirmed the fabrication failure mode via `call_logs` artifact inspection, deployed the fix, confirmed clean boot and continued healthy serving.


hartmark:
Extended this PR while investigating live: the empty-`context_handoffs`-table root cause turned out to have a second, more serious layer.

Same-request fallback targets (`i > 0`) were getting the universal-handoff note injected too, comparing against the session's *previous turn* model rather than skipping entirely. The client never saw the earlier same-request target fail, so there's nothing to hand off -- but the note claims continuity anyway. Observed live: a same-request fallback landing on a weak free-tier model, given that note plus only trimmed tool-output input (no real history), fabricated an entirely invented response about an unrelated topic instead of just answering the actual request.

This restores the intent of an earlier attempt at this same fix in #12227 (commit ffbd0699) that was reverted because it broke a test asserting the *old* behavior was correct -- that test conflated same-request fallback with genuine cross-turn provider switches. Fixed now with the corrected, narrower contract: a same-request fallback target gets the original request untouched.

PR #12340, opened 2026-09-01
fix(combo): scope Claude model failures to the model, not the account

### LINKED ISSUE #12334: fix(claude): priority combo never advances past step 1 when all steps share one Claude OAuth connection
**OmniRoute version:** 3.8.50 (Docker, `diegosouzapw/omniroute:main` @ `sha256:88996a83f44697580a7835eeba9a6a44326f5c9b6b89f678cca56faf2a6e24ea`, 2026-09-01)
**Install method:** Docker
**OS:** Linux (Ubuntu 24.04 host)
**Provider:** `claude` (OAuth, `organizationType: claude_max`)
**Models:** `claude-fable-5-max`, `claude-opus-5-max`, `claude-opus-4-8-max`, `claude-opus-4-7-max`, `claude-opus-4-6`
**Client:** cURL (also reproduced through Claude Code and DeepSeek Harness)

## Description

A `priority` combo whose steps are five different models on a **single** Claude OAuth connection never advances past step 1. Whatever kills step 1 is recorded against the connection, and the pre-dispatch cooldown gate then skips every remaining step because they share that connection — so a per-model drop chain silently behaves like a single-model route.

The chain I wanted is the same shape as Claude Code's own fallback: stay on Fable 5 until it errors, then continue on Opus 5, then 4.8, then 4.7, then 4.6, all on one Max subscription.

I also hit a second, unrelated bug on the way in (`claude-fable-5-max` returning an empty stream on long histories) — details at the end, happy to split it into its own issue if you prefer.

## Steps to Reproduce

1. Connect one `claude` OAuth account.
2. Create a `priority` combo whose steps all pin that same `connectionId`, with a deliberately invalid model first so the model-specific bug below isn't part of the picture:

```json
{"name":"zz-failover-test","strategy":"priority","models":[
  {"id":"m1","kind":"model","model":"claude/claude-bogus-model-x","providerId":"claude","connectionId":"<A>"},
  {"id":"m2","kind":"model","model":"claude/claude-opus-5-max","providerId":"claude","connectionId":"<A>"}
]}
```

3. `POST /v1/messages` with `"model": "zz-failover-test"`.
4. Change only step 2 to a different account (`agy`) and repeat.

## Expected Behavior

Step 1 fails, step 2 is dispatched. A 404 on one model says nothing about the other models on that connection.

## Actual Behavior

Same account — step 2 is never dispatched:

```
COMBO: Trying model 1/2: claude/claude-bogus-model-x
COMBO: All models failed | claude/claude-bogus-model-x: model — model: claude-bogus-model-x (HTTP 404)
```

Different account on step 2 — works:

```
COMBO: Trying model 1/2: claude/claude-bogus-model-x
COMBO: All models failed | claude/claude-bogus-model-x: model — model: claude-bogus-model-x (HTTP 404)
COMBO: Trying model 2/2: agy/claude-sonnet-4-6      ← dispatched, answers 200
```

With the real chain the connection gets marked and all four remaining steps report the skip:

```
COMBO: Model claude/claude-fable-5-max returned 200 but failed quality check: streaming upstream error
AUTH:  claude | all 1 active accounts rate limited (reset after 5s) | lastErrorCode=502.0, lastError=Claude returned an empty response (no content bloc
AUTH:  (×4 — one per remaining step)
COMBO: All models failed | Upstream response failed quality validation: streaming upstream error
```

Things I tried that don't change it:

- `config.failoverBeforeRetry: true` with `maxSetRetries: 1`, `setRetryDelayMs: 0` — the `AUTH … rate limited` lines disappear, step 2 is still never tried.
- Leaving `connectionId` off step 2 — the request then ends with `all targets were skipped by pre-dispatch filters`.

## Where it comes from

`hasPerModelQuota()` in `open-sse/services/accountFallback.ts` decides whether a failure is scoped to the model or to the whole connection. It returns true for `antigravity`/`agy`, `codex`, `gemini`, `github`, passthrough and compatible providers — and false for `claude`. So in `src/sse/services/auth.ts` the model-lockout branch (`isPerModelQuotaProvider && … (status === 404 || status === 429 || status >= 500)`) never runs for a Claude OAuth connection, the failure lands on the connection row instead, and `getPersistedConnectionCooldownSkipReason()` in `open-sse/services/combo/comboPredicates.ts` then skips every sibling target before dispatch.

A Claude OAuth connection multiplexes Fable 5, Opus 5, 4.8, 4.7, 4.6, Sonnet and Haiku behind one credential, which is exactly the multiplexing the doc comment on `hasPerModelQuota` describes. That also matches how the subscription actually meters: Opus has its own weekly cap on Max, so an Opus 429 doesn't mean Sonnet is unavailable — dropping to another model is the documented behaviour of Anthropic's own client.

I'm opening a PR that adds `claude` to `hasPerModelQuota()` with a regression test mirroring `tests/unit/combo-provider-cooldown-sibling.test.ts`.

## Second bug: `claude-fable-5-max` empty stream above ~1800 messages, streaming only

Minimal repro — no tools, no system prompt, alternating `ping N` / `pong N` messages of ~4 tokens each, ending with a `say OK` user turn:

| messages | `stream: true` | `stream: false` |
| --- | --- | --- |
| 1400 | ok | ok |
| 1521 | ok | ok |
| 1801 | **502 empty response** | ok |

```
[STREAM] Empty Claude stream at flush - emitting error (claude:claude-fable-5-max)
🌊 [STREAM] CLAUDE | claude-fable-5-max | 14490ms | error: Claude returned an empty response (no content block)
❌ claude [502]: Claude returned an empty response (no content block)
```

Same account, same betas, same payload: `claude-opus-5-max`, `claude-opus-4-8-max` and `claude-opus-4-7-max` all answer fine at 1801 messages. Size isn't the trigger — a single 275 KB / 125k-token message streams fine, it's the message count, and the boundary moved from ~1350 on the 2026-08-13 build to ~1800 on the current one. The 14.5 s before the flush suggests the upstream connection is held open and then closed with nothing on it, so a retry-on-empty-stream inside the Claude SSE handler may be worth it independently of the combo behaviour.

## Test Impact

Needs a new unit test.

Kizuno18:
## Summary

A priority combo whose steps are five different models on one Claude OAuth connection stops at step 1: whatever kills the first step is recorded against the connection, and the pre-dispatch cooldown gate then skips every later step because they share that connection. The drop order is silently a single-model route.

`hasPerModelQuota()` returns false for `claude`, so in `markAccountUnavailable()` the model-lockout branch never runs and the failure lands on the connection row instead. `getPersistedConnectionCooldownSkipReason()` takes it from there.

To keep the model-specific bug in #12334 out of the picture, I reproduced it with a first step naming a model that cannot exist. The 404 lands and step 2 is never dispatched; point step 2 at a different account and the same combo works:

```
COMBO: Trying model 1/2: claude/claude-bogus-model-x
COMBO: All models failed | claude/claude-bogus-model-x: model — model: claude-bogus-model-x (HTTP 404)
COMBO: Trying model 2/2: agy/claude-sonnet-4-6      ← only when step 2 is another account
```

A Claude OAuth connection multiplexes Fable 5, Opus 5/4.8/4.7/4.6, Sonnet and Haiku behind one credential — the multiplexing the doc comment on `hasPerModelQuota` already describes. Its quota is a different question, though: a 429 on a Max subscription is account-wide, and `shouldMarkAccountExhaustedFrom429("claude", …)` pins that in `account-fallback-service.test.ts`. Adding `claude` to `hasPerModelQuota` breaks that test, and rightly so.

So this splits the two: a new `hasPerModelFailureScope()` covers the non-quota statuses (404, 5xx, a stream that ended with no content), `auth.ts` uses it only when `status !== 429`, and quota scope stays exactly as it was. `targetExhaustion` gets the same treatment for 404, which names one model the account cannot serve rather than a bad connection — the empty-content 502 was already exempt through `isEmptyContentFailure`.

## Related Issues

- Closes #12334

## Validation

- [x] Change type: routing
- [x] Focused tests: `tests/unit/combo-claude-per-model-scope.test.ts` plus every existing test touching `hasPerModelQuota` or `targetExhaustion` — 17 files, 226 passing, 0 failing
- [x] `npx eslint` on the four touched files: same 3 pre-existing `no-unused-vars` errors in `accountFallback.ts` before and after (line 2413 shifts to 2437), no new ones
- [x] Reconciled with `release/v3.8.51`
- [x] Production-code changes come with a new test

Full unit suite on this branch: 47 failures. On the same checkout with the patch stashed: 46. The one extra is `Vietnamese locale has no internal missing markers or empty fallbacks`, which fails on a clean tree when run on its own:

```
not ok 1 - Vietnamese locale has complete key parity with English
not ok 2 - Vietnamese locale has no internal missing markers or empty fallbacks
```

so it is suite variance, not this change.

## Tests Added Or Updated

- `tests/unit/combo-claude-per-model-scope.test.ts` (new) — mirrors `combo-provider-cooldown-sibling.test.ts`: failure scope for `claude`, quota scope unchanged, the connection-level override still winning, and the two sides of the pre-dispatch gate (a real cooldown still protects siblings; no cooldown lets step 2 dispatch).

## Coverage Notes

`hasPerModelFailureScope()` is covered by the new file. The `auth.ts` and `targetExhaustion.ts` call sites are covered indirectly by the existing lockout tests for gemini, github, nvidia, ollama, openrouter, vertex and agentrouter, which all still pass — they pin that nothing outside `claude` changed scope.

## Reviewer Notes

The risk worth looking at is the 429 split. I deliberately left quota scope alone so a Max rate limit keeps retiring the account, since dropping to another model there would just burn a second upstream call. If you would rather have per-model quota for `claude` too, that is a one-line change to `hasPerModelQuota` plus an update to the `shouldMarkAccountExhaustedFrom429` expectation — happy to do it either way.

Reproduced against `diegosouzapw/omniroute:main` @ `sha256:88996a83` and the previous `sha256:611c555b`, on a single `claude_max` OAuth connection.


cryptiklemur:
Would this allow harnesses (like CC) to see that fable is 429ing, and automatically change to Opus, like it normally does in CC (without a proxy)?

Kizuno18:
@cryptiklemur Not for a 429 in this PR. OmniRoute still treats Claude 429s as account-wide, so another model on the same OAuth connection remains on cooldown. This fixes model-specific failures such as a 404, allowing the combo to try the next model on that connection. Distinguishing a Fable-only limit from a shared account limit would need a separate change.

diegosouzapw:
Good fix — a Claude OAuth connection multiplexing Fable/Opus/Sonnet/Haiku
behind one credential shouldn't have a 404/5xx on one model exhaust the
whole connection and stall a priority combo at step 1. The
`hasPerModelFailureScope()` split (quota stays account-wide via
`hasPerModelQuota`, non-quota failures get model scope for Claude) is the
right generalization. Probed the test file locally: 7/7 pass. This is
complementary to #12957 in the same batch (different layer of the
model-lockout system — no file-line overlap, verified) — good to merge
either order.

diegosouzapw:
Thanks @Kizuno18 — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #12357, opened 2026-09-01
fix(api-keys): persist the creation ACL atomically with the key row

### LINKED ISSUE #12275: fix(api): POST /api/keys silently ignores modelAccessMode/allowedModels/allowedCombos — returns 201 with a fully permissive key
## Summary

`POST /api/keys` accepts `modelAccessMode`, `allowedModels` and `allowedCombos` in the request body without error, but **none of them are persisted**. The key is created with default permissive access (`modelAccessMode: "all"`, `allowedCombos: ["combo/*"]`), and a follow-up `PATCH /api/keys/{id}` is required to apply the intended ACL.

Because the request returns **HTTP 201 with no warning**, callers reasonably assume the ACL took effect. Any automated provisioning that creates a restricted key in one call ends up with a **fully permissive key**.

Split out from #12267 as requested.

## Environment

- OmniRoute **v3.8.50**, self-hosted, Docker

## Reproduction

```bash
curl -X POST http://<host>/api/keys \
  -H 'x-omniroute-cli-token: <token>' \
  -d '{
        "name": "acl-repro",
        "modelAccessMode": "restricted",
        "allowedModels": [],
        "allowedCombos": ["my-combo"]
      }'
```

Response: **201 Created**, no warning or validation error.

Read back with `GET /api/keys`:

| Field | Sent | Persisted |
| --- | --- | --- |
| `modelAccessMode` | `"restricted"` | **`"all"`** |
| `allowedModels` | `[]` | `[]` |
| `allowedCombos` | `["my-combo"]` | **`["combo/*"]`** |

The resulting key allows **every** model and **every** combo — the opposite of what was requested.

A subsequent `PATCH /api/keys/{id}` with the identical fields persists correctly, so the capability exists; only the creation path is affected.

## Root cause

Three layers each drop the fields independently.

**1. Zod schema does not declare them** — `src/shared/validation/schemas/keys.ts:32`

```ts
export const createKeySchema = z
  .object({
    name: z.string().min(1, "Name is required").max(200),
    noLog: z.boolean().optional(),
    allowUsageCommand: z.boolean().optional(),
    usageLimitEnabled: z.boolean().optional(),
    dailyUsageLimitUsd: z.coerce.number().min(0).optional().nullable(),
    weeklyUsageLimitUsd: z.coerce.number().min(0).optional().nullable(),
    chaosModeEnabled: z.boolean().optional(),
    scopes: z.array(z.string().trim().min(1).max(64)).max(32).optional(),
    allowedConnections: z.array(z.string().uuid()).min(1).max(100).optional(),
  })
  .superRefine(requireExclusiveLeaseConnections);
```

`modelAccessMode` / `allowedModels` / `allowedCombos` are absent. Zod's default `strip` behaviour removes unknown keys **silently**, so `validation.data` never carries them and no error surfaces.

**2. The route never reads them** — `src/app/api/keys/route.ts:72-87`

```ts
const {
  name, noLog, scopes, allowedConnections, allowUsageCommand,
  usageLimitEnabled, dailyUsageLimitUsd, weeklyUsageLimitUsd, chaosModeEnabled,
} = validation.data;
...
const apiKey = await createApiKey(name, machineId, normalizedScopes, { allowedConnections });
```

**3. `createApiKey` cannot accept them** — `src/lib/db/apiKeys.ts:661`

```ts
export async function createApiKey(
  name: string,
  machineId: string,
  scopes: string[] = [],
  options: { allowedConnections?: string[] } = {}
) {
```

`options` only carries `allowedConnections`.

## Note: the fix pattern already exists in this handler

Immediately after creation, the same route already performs a follow-up write for other non-`createApiKey` fields (`route.ts:88`):

```ts
if (noLog === true || allowUsageCommand === true || usageLimitEnabled === true ||
    dailyUsageLimitUsd !== undefined || weeklyUsageLimitUsd !== undefined ||
    chaosModeEnabled === true) {
  await updateApiKeyPermissions(apiKey.id, { ... });
}
```

The three ACL fields appear to have simply been omitted from this list — `updateApiKeyPermissions` is the same function the working `PATCH` path uses.

## Expected

Either:

1. **Persist them on create** — add the three fields to `createKeySchema` and include them in the existing `updateApiKeyPermissions` follow-up call (smallest change, consistent with how `noLog` et al. are handled); or
2. **Reject them explicitly** — if creation-time ACL is intentionally unsupported, `.strict()` on the schema would return a 400 instead of silently returning a permissive key.

Option 1 is preferable: silently returning a key with **broader** access than requested is a fail-open security posture, and the follow-up-write pattern is already established in this handler.

## Impact

Provisioning scripts that create restricted keys in a single call produce fully permissive keys, with a 201 and no diagnostic. The failure is silent in both directions — nothing in the response, logs, or dashboard indicates the ACL was dropped.

pacocartones:
## Summary

- `POST /api/keys` now persists the `modelAccessMode`, `allowedModels` and `allowedCombos` supplied at creation time instead of silently creating a permissive key.
- The ACL is written by the same `INSERT` that creates the key row: a key never exists, even transiently, with a wider policy than requested, and a failed follow-up update can no longer leave it permissive.
- The create schema accepts the same ACL shapes as the PATCH route and rejects the contradictory `modelAccessMode: "all"` + non-empty `allowedModels` form.

## Related Issues

- Closes #12275

## Validation

- [x] Change type: DB / API
- [x] Focused tests and category gates from the golden path: `tests/integration/api-keys.test.ts` 19/19, `tests/unit/api-key-scope-validation.test.ts` + `tests/unit/api-manager-provider-permissions.test.ts` + `tests/unit/api-key-policy.test.ts` + `tests/unit/db-apiKeys-crud.test.ts` 124/124, `npm run typecheck:core` exit 0, focused ESLint exit 0, `check:changelog-integrity` OK, `check:provider-consistency` OK
- [ ] `npm run lint` — exit 0 on the current tip (repository-wide)
- [x] Reconciled with the current active release base `release/v3.8.51@158647618`; focused checks rerun afterward

Mutation checks: removing the ACL pass-through in the route, or making `createApiKey` ignore its ACL options, turns the new tests RED (2/2 fail in each case); restoring the patch returns GREEN.

## Tests Added Or Updated

- `tests/integration/api-keys.test.ts` — restricted mode + empty model list + named Combo persisted from `POST /api/keys`; explicit model allowlist preserved; empty `allowedCombos` denies every Combo from creation
- `tests/unit/db-apiKeys-crud.test.ts` — `createApiKey` persists the ACL in the key row; defaults unchanged without options; allow-all mode normalizes to an empty allowlist
- `tests/unit/api-key-scope-validation.test.ts` — `createKeySchema` rejects allow-all with a non-empty allowlist and accepts restricted + empty

## Coverage Notes

- `src/lib/db/apiKeys.ts` (`createApiKey`, insert statement) is covered by the new CRUD tests; `src/app/api/keys/route.ts` and `src/shared/validation/schemas/keys.ts` by the integration and schema tests above.
- No file's coverage moved down; the touched functions gained tests.

## Reviewer Notes

- The `INSERT` now sets `model_access_mode` explicitly (column exists since migration 147, default `'all'`), so rows created without ACL options are byte-identical to before.
- `createApiKey` gained an optional `modelAccessMode` / `allowedModels` / `allowedCombos` on its existing `options` object; existing callers are unaffected.
- Semantics match the PATCH route: `restricted` + `allowedModels: []` denies every direct model; `allowedCombos: []` denies every Combo; `["combo/*"]` (the default) allows all Combos.

### DIFF
--- a/changelog.d/fixes/PENDING-api-key-create-acl.md
+++ b/changelog.d/fixes/PENDING-api-key-create-acl.md
@@ -1,0 +1,1 @@
+- **fix(api-keys):** persist model and Combo access controls supplied when creating an API key
--- a/src/app/api/keys/route.ts
+++ b/src/app/api/keys/route.ts
@@ -87,17 +87,21 @@ export async function POST(request) {
     // Always get machineId from server
     const machineId = await getConsistentMachineId();
     const normalizedScopes = normalizeSelfServiceScopesForCreate(scopes);
-    const apiKey = await createApiKey(name, machineId, normalizedScopes, { allowedConnections });
+    // The model/Combo ACL is persisted in the same INSERT as the key row: a key must never
+    // exist, even transiently, with a wider access policy than the one requested (#12275).
+    const apiKey = await createApiKey(name, machineId, normalizedScopes, {
+      allowedConnections,
+      ...(modelAccessMode !== undefined && { modelAccessMode }),
+      ...(allowedModels !== undefined && { allowedModels }),
+      ...(allowedCombos !== undefined && { allowedCombos }),
+    });
     if (
       noLog === true ||
       allowUsageCommand === true ||
       usageLimitEnabled === true ||
       dailyUsageLimitUsd !== undefined ||
       weeklyUsageLimitUsd !== undefined ||
-      chaosModeEnabled === true ||
-      modelAccessMode !== undefined ||
-      allowedModels !== undefined ||
-      allowedCombos !== undefined
+      chaosModeEnabled === true
     ) {
       await updateApiKeyPermissions(apiKey.id, {
         ...(noLog === true && { noLog: true }),
@@ -106,9 +110,6 @@ export async function POST(request) {
         ...(dailyUsageLimitUsd !== undefined && { dailyUsageLimitUsd }),
         ...(weeklyUsageLimitUsd !== undefined && { weeklyUsageLimitUsd }),
         ...(chaosModeEnabled === true && { chaosModeEnabled: true }),
-        ...(modelAccessMode !== undefined && { modelAccessMode }),
-        ...(allowedModels !== undefined && { allowedModels }),
-        ...(allowedCombos !== undefined && { allowedCombos }),
       });
     }
 
--- a/src/app/api/keys/route.ts
+++ b/src/app/api/keys/route.ts
@@ -79,6 +79,9 @@ export async function POST(request) {
       dailyUsageLimitUsd,
       weeklyUsageLimitUsd,
       chaosModeEnabled,
+      modelAccessMode,
+      allowedModels,
+      allowedCombos,
     } = validation.data;
 
     // Always get machineId from server
@@ -91,7 +94,10 @@ export async function POST(request) {
       usageLimitEnabled === true ||
       dailyUsageLimitUsd !== undefined ||
       weeklyUsageLimitUsd !== undefined ||
-      chaosModeEnabled === true
+      chaosModeEnabled === true ||
+      modelAccessMode !== undefined ||
+      allowedModels !== undefined ||
+      allowedCombos !== undefined
     ) {
       await updateApiKeyPermissions(apiKey.id, {
         ...(noLog === true && { noLog: true }),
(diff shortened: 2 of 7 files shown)


pacocartones:
Closing: #12352 (opened earlier today by @KrzysiekSko) fixes the same issue, including the atomic insert. If the extra coverage here is useful (empty allowedCombos denies every Combo from creation, schema rejection of allow-all + non-empty allowlist, createApiKey CRUD tests), feel free to lift those tests into #12352.

Issue #8169, opened 2026-07-22
fix(backend): token-economy/compression truncates words unreadably even at minimum settings


diegosouzapw:
### Summary
A user (support mesh, WhatsApp BR) reports that with the token-saving/compression options set to the minimum, OmniRoute still truncates words so aggressively that downstream LLMs can no longer recognize the word. Paraphrased: "I set the token-economy systems to the minimum in OmniRoute, but it's truncating words so much that other LLMs can't identify the word."

### Expected
At minimum settings, token-economy/compression should not mangle individual words to the point they become unreadable to downstream models.

### To reproduce (info requested from reporter)
- Which exact "token economy"/compression options were enabled (setting names / screenshots)
- The model(s) involved (source and target)
- A short before/after example of the truncated text

### Notes
Opened from a support-mesh report to track. Likely area: the compression / abbreviation guardrail or long-context compaction path (relates to prior heap/compression work #7849). Needs the reporter's config + a concrete example to reproduce and scope.

diegosouzapw:
Fixed in `release/v3.8.49` via #8239 (already merged into the active development branch — feel free to pull and test it now). It will ship in the next release. Thanks for the detailed report!

PR #12383, opened 2026-09-01
fix(memory): word/sentence-boundary aware fact truncation

LeMonBLOCK:
## Bug

`src/lib/memory/extraction.ts` truncates extracted memory facts with raw
character-offset slicing and no boundary awareness:

- `sanitizeMatch()` did `raw.trim().replace(/\s+/g, ' ').slice(0, MAX_FACT_LENGTH)`
  — a hard cut that can land mid-word.
- `capExtractionText()` did `text.slice(-MAX_EXTRACTION_TEXT_LENGTH)` on the
  front edge when input exceeds `MAX_EXTRACTION_TEXT_LENGTH` (64KB) — same
  problem, opposite edge.

These facts later get injected into LLM context as `<system-reminder>`
memory blocks, so a mid-word/mid-clause cut produces garbled fragments
(e.g. text starting with a stray closing parenthesis, or ending with no
punctuation and a half word).

## Fix

- `sanitizeMatch()` now backs the cut index off to the nearest clean
  boundary within a lookback window: prefers a sentence-ending mark
  (`.`/`!`/`?`) so the fact reads as a complete clause, falls back to a
  plain whitespace/word boundary, and only falls back to the original hard
  cut when no boundary exists in the window (e.g. one long unbroken run of
  characters).
- `capExtractionText()` applies the equivalent boundary-aware trim on the
  front edge of the kept tail (extends the cut forward to the next
  whitespace boundary rather than slicing mid-word).

This mirrors the boundary-aware truncation already used by
`open-sse/services/compression/lite.ts` (issue #8169) for tool-result
truncation, so the codebase now has one consistent pattern for this class
of problem.

## Testing

- New: `tests/unit/memory-extraction-boundary-truncation.test.ts` — covers
  a long match cut at a word boundary, a match cut preferentially at
  sentence-ending punctuation, short strings passing through untouched,
  the no-boundary-in-window fallback, and `capExtractionText`'s
  boundary-aware tail behavior (both under and over the 64KB cap).
- Ran existing `tests/unit/memory-extraction.test.ts` alongside the new
  file — all 28 tests pass, no regressions:
  ```
  ℹ tests 28
  ℹ pass 28
  ℹ fail 0
  ```
- `npx tsc --noEmit -p tsconfig.json` — no new type errors introduced by
  this change.

No unrelated files were touched.

### DIFF
--- a/src/lib/memory/extraction.ts
+++ b/src/lib/memory/extraction.ts
@@ -61,16 +61,65 @@ export interface ExtractedFact {
 
 // ─── Extraction Logic ────────────────────────────────────────────────────────
 
-/**
- * Sanitize a matched string: trim, collapse whitespace, cap length
- */
-function sanitizeMatch(raw: string): string {
-  return raw.trim().replace(/\s+/g, " ").slice(0, MAX_FACT_LENGTH);
+// Lookback window (chars) used to back a hard cut index off to a clean
+// sentence or word boundary instead of slicing mid-word (see #8169's
+// compressToolResults for the same pattern applied to tool-result truncation).
+const BOUNDARY_LOOKBACK = 80;
+
+function isWordChar(char: string | undefined): boolean {
+  return char !== undefined && /\S/.test(char);
 }
 
+/**
+ * Back a hard cut index off to the nearest clean boundary within
+ * BOUNDARY_LOOKBACK chars before it. Prefers a sentence-ending punctuation
+ * mark (. ! ?) so a fact reads as a complete clause; falls back to a plain
+ * whitespace/word boundary; falls back to the original cut index when
+ * neither is found in the window (e.g. one long unbroken run of chars).
+ */
+function backOffToBoundary(content: string, cutIndex: number): number {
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Nice fix — matching the same boundary-lookback pattern the repo already uses for #8169
(`compressToolResults`) is a good consistency call, and the sentence-punctuation-first /
word-boundary-fallback / original-cut-as-last-resort chain reads cleanly. Please add a
changelog fragment under `changelog.d/fixes/` before merge.

diegosouzapw:
Thanks @LeMonBLOCK — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #12391, opened 2026-09-02
fix(sse): stop mergeAbortSignals from leaking abort listeners

Beexly:
## Summary

`mergeAbortSignals()` in `open-sse/executors/base.ts` attached `"abort"` listeners to its `primary`/`secondary` signals but never removed them once the merged signal settled. Every executor fetch attempt calls this (`fetchWithStartTimeout`, once per URL/retry), so a busy combo request accumulated one live listener per call on the long-lived combo/client signal. A leaked listener still fires when that signal is later aborted (e.g. a hedge cancellation arriving after this merge's own caller already finished), for a merged output nothing is watching anymore.

Fix mirrors the already-correct self-cleaning pattern in `open-sse/utils/directResponseStartTimeout.ts`'s local `mergeAbortSignals`.

Originally investigated from a production crash: `Error [AbortError]: hedge-cancelled` thrown unhandled from deep inside the built Next.js server bundle, killing the whole Node process. This fix addresses one confirmed, measured listener leak on the exact signal-merge path that carries the hedge-cancellation reason through every provider fetch — it removes a real defect regardless, though it was not possible to build a full end-to-end repro of that exact crash through the entire combo-hedging dispatch path.

## Test plan

- [x] TDD: added a regression test (`tests/unit/executor-base-utils.test.ts`) that merges the same long-lived signal 25 times and counts real listener growth via `node:events`' `getEventListeners`.
- [x] Confirmed red on pre-fix code: exactly 25 leaked listeners after 25 merges.
- [x] Confirmed green on fixed code: 0 leaked listeners, 23/23 tests passing in the full file.
- [x] Ran `tests/unit/executor-base-utils.test.ts` on Windows (the platform the original crash was observed on) — 23/23 pass.
- [x] Scoped ESLint on both changed files — 0 new violations (pre-existing violations elsewhere in `base.ts` are outside this diff's hunks).

### DIFF
--- a/open-sse/executors/base.ts
+++ b/open-sse/executors/base.ts
@@ -225,14 +225,36 @@ export type CountTokensInput = {
   signal?: AbortSignal | null;
 };
 
+/**
+ * Merge two abort signals into one that fires when either does.
+ *
+ * The `primary`/`secondary` abort listeners registered below MUST be removed once the
+ * merged controller settles — otherwise they outlive this call and stay attached to
+ * whichever input signal is longer-lived (typically the combo/client signal, which
+ * stays open for the whole request while this merge is re-done per executor fetch
+ * attempt/retry). A hedge cancellation or client disconnect arriving after this
+ * particular merge's caller has already moved on then still fires the listener,
+ * detached from anything that's still awaiting it. Mirrors the already-correct
+ * self-cleaning pattern in `open-sse/utils/directResponseStartTimeout.ts`'s local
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Your fix here for the `mergeAbortSignals` listener leak is correct, and you've since evolved it
further in #12406 (which extracts the function into its own module and also fixes a sibling leak
in `executeWithUpstreamStartTimeout`) — same test coverage, more complete scope. We're closing
this one in favor of #12406 so we don't merge the fix twice; thanks for catching and fixing it.

PR #10297, opened 2026-08-13
docs(ops): k8s probe recommendations (TCP liveness, HTTP /healthz readiness)

RaviTharuma:
## Summary

- Document the two HTTP health surfaces: stock `GET /healthz` (lifecycle) vs `GET /api/monitoring/health` (deep).
- Add Kubernetes probe recommendations: **TCP liveness** (or soft HTTP `/healthz`), **HTTP readiness on `/healthz`**, do **not** point kubelet liveness at `/api/monitoring/health`.
- Note that both routes share the Node event loop — busy ≠ dead — and link #10052 / #9685 / #10055 / #10117.
- Cross-link from the Docker guide so operators do not copy Docker `HEALTHCHECK` into kubelet liveness.

## Related Issues

- Related to #10052 (probe timeouts while event loop busy)
- Related to #9685 / #10055 (catalog pricing hog)
- Related to #10117 (compression token-count hog)
- Related to #10296 (stable channel still on 3.8.49)

## Validation

- [x] Docs-only change (no production code)
- [ ] `npm run lint` N/A for markdown-only
- [x] Production-code changes include a new or updated automated test in this PR — N/A (docs only)
- [ ] SonarQube PR analysis is green or any remaining issues are explicitly documented below

## Tests Added Or Updated

- No production code changed; no automated tests added.

## Coverage Notes

- Docs-only; no `src/` / `open-sse/` / `electron/` / `bin/` coverage impact.

## Reviewer Notes

- Docs only. Does not fix event-loop hogs; gives operators a safer probe layout until SemVer images ship the code fixes.
- English docs updated (`docs/ops/MONITORING_GUIDE.md`, `docs/guides/DOCKER_GUIDE.md`). No secrets or private hostnames.

PR #12400, opened 2026-09-02
feat(deploy): Kubernetes support for k3s and k8s

PauloFH:
## Summary

The repo ships `docker-compose.yml` and `fly.toml`, and `DOCKER_GUIDE.md` already warns that "stock Docker / Kubernetes OmniRoute is one Node process + one SQLite writer" — but there are no manifests. Anyone running this on k3s or k8s writes them from scratch, against a topology whose constraints are invisible until they bite:

- a second replica against the same volume corrupts the database;
- an ingress that buffers responses withholds the SSE stream until the provider turn ends, which reads as a hung request, not as broken streaming;
- a liveness probe pointed at `/api/monitoring/health` does real DB work on every kubelet tick and restarts the only pod under load.

So the topology is **encoded, not documented**. `replicas: 1`, `Recreate` and `ReadWriteOnce` are fixed in the base, in the chart and in the generator — not exposed as options anywhere. And the chart and the spec schema refuse to render a configuration that fails at runtime, naming the field rather than emitting something that breaks later:

| Rejected | Because |
| --- | --- |
| `replicaCount > 1` | two pods on one volume corrupt the SQLite file |
| access mode ≠ `ReadWriteOnce` | a shared-write volume invites the same second writer |
| heap ceiling ≥ memory limit | the kernel OOM-kills the pod instead of Node raising a recoverable heap error (#7849) |
| `shutdownTimeoutMs` ≥ grace − preStop | the kubelet SIGKILLs mid-drain, so in-flight SSE is cut rather than drained |
| TLS with no Ingress | terminating TLS at something that does not exist |

Four artifacts, one topology:

- **`deploy/kubernetes/`** — Kustomize base plus k3s (Traefik + `local-path`) and generic k8s (ingress-nginx) overlays.
- **`deploy/helm/omniroute/`** — the same topology, values-driven.
- **`omniroute deploy k8s`** — prints manifests built from a running instance's settings, so ports, origins and flags are not transcribed by hand. It writes to stdout and touches nothing, so it composes: `omniroute deploy k8s --target local | kubectl apply -f -`.
- **`docs/ops/KUBERNETES_DEPLOYMENT_GUIDE.md`** — 416 lines covering deploy, ingress/SSE, probes, upgrades, storage/backups, sizing and troubleshooting.

`local` and `vps` differ only in defaults. `local` assumes k3s on the same machine. `vps` assumes a hostname in front of a cluster other people can reach, so TLS is on, `REQUIRE_API_KEY=true`, the cluster's default StorageClass binds, and `LIVE_WS_ALLOWED_ORIGINS` is derived from the host — without that, the dashboard's live view is rejected by the origin check on the very host it is served from.

Probes follow `docs/ops/MONITORING_GUIDE.md` ("Kubernetes probe recommendations", merged in #10297): `/healthz` for startup and readiness, `/livez` for liveness, never `/api/monitoring/health`. The nginx overlay disables response buffering; the Traefik idle-timeout caveat is documented for k3s, since that one is an entrypoint setting and not an Ingress annotation.

**Provisioning is advisory on purpose.** `omniroute deploy k8s provision` prints the k3s commands and never runs them or connects to a server. Doing it directly would mean shipping an SSH client and holding a private key — a standing surface for a one-off task — so it follows the shape `contrib/vps` already uses for a VPS install. The VPS plan covers the two steps easiest to get wrong by hand: `--tls-san`, without which the cluster is only reachable as `127.0.0.1` and a remote `kubectl` fails TLS verification, and rewriting the `127.0.0.1` server address in the kubeconfig k3s writes. Values rendered into those commands are constrained to hostname, version, username and context shapes, so a printed command cannot mean something other than what it looks like.

**No new API route and no new dashboard surface.** An earlier draft put this behind `/api/deploy/*` with a UI tab; it was dropped before landing. Applying manifests by clicking is at odds with GitOps — the desired state should be versioned, not produced by a form — and the routes existed only to serve the button, spawning `kubectl` with the host kubeconfig, which is a full cluster credential. The CLI reaches the same result without that surface, so `routeGuard.ts` is untouched.

**Two fixes to the manifests themselves**, both surfaced by the test that pins the generator against `deploy/kubernetes/base`:

- Dropped `API_PORT` from the base ConfigMap. Setting it moves the API to a separate port (`src/lib/runtime/ports.ts:27`) that neither the container nor the Service exposes, leaving that listener unreachable. Unset, the API and dashboard share `PORT` — which is what the manifests actually expose.
- Made the container port protocol explicit, so the committed manifest and the generator agree byte for byte.

`deploy` also joins `--mcp` in reserving stdout for data (`bin/omniroute.mjs`). Env-loading notices there made the YAML unparseable — `control characters are not allowed` — which broke the pipe the command exists for.

## Related Issues

- Related to #10297 — the probe recommendations these manifests implement
- Related to #7849 — the V8-heap-vs-cgroup ceiling the spec schema now enforces
- Related to #8075 — multi-writer HA; until it ships, single-replica is the documented ceiling and this PR encodes it

## Validation

- [x] Change type: build-deploy (plus CLI)
- [x] `npm run lint`

**Applied to a live cluster, not just a dry run.** Single-node k3s v1.36.4+k3s1:

| Checked | Result |
| --- | --- |
| PVC on `local-path` | `Bound`, 10Gi, RWO |
| Pod | `1/1 Running`, 0 restarts |
| Rollout | completed, `Available=True` |
| `/healthz` and `/livez` through the Service | HTTP 200 both |
| `/v1/models` | answers application JSON, so the app serves — not just the probes |
| Probe failure events | none recorded |
| Helm chart, `--dry-run=server` | accepted by the real API server |

That run corrected two documentation errors a client-side dry run cannot catch: the health check used `wget`, and the published image ships neither `wget` nor `curl` (`node` is the only HTTP client in it); and an Ingress naming an IngressClass the cluster does not run is accepted by the API server and then silently routes nothing. Both are now in the guide, the second as a troubleshooting row.

Red first, per invariant: each manifest guard was verified to fail under a deliberate mutation of the file it protects — flipping `replicas` to 2, the access mode to RWX, `proxy-buffering` to `on`, liveness to `/api/monitoring/health`, or the shutdown budget past the grace period each turns exactly one test red, and only one. The generator/CLI suites found three real defects while being written: a `fail` placed at the top level of a `_*.tpl` file (Helm parses and discards it, so the replica guard was a silent no-op), context-name validation that rejected legitimate EKS contexts, and the stdout pollution above.

The automated review then found four more, fixed in the second commit: an address pattern that accepted a port or a bare IPv6 literal and rendered `https://host:123:6443`; a context pattern that rejected the ARN-shaped contexts managed clusters produce; the context being interpolated into a temp path, which the previous item would have turned into a directory traversal; and a quantity pattern that accepted `K` while rejecting `k`, the lowercase decimal thousand Kubernetes actually uses, so `512k` was refused. Each has a regression case.

Rebased onto `release/v3.8.51` and the full set rerun there: **54 tests passing**, `npm run typecheck:core` clean, both overlays render with zero kustomize warnings, `helm lint` clean, `check:docs-all` / `check:tracked-artifacts` / `check:cli-i18n` passing, and `omniroute deploy k8s | kubectl apply --dry-run=client -f -` still creates all six objects.

`npm run lint` reports **zero errors** but exits 2, on this branch and on an untouched checkout of the base alike — `suppressions left that do not occur anymore`, unrelated to any file here.

## Tests Added Or Updated

- `tests/unit/k8s-deploy-manifests.test.ts` — 12 cases over the committed manifests: single replica, Recreate, RWO, probe targets, drain budget, unbuffered SSE, secret hygiene, and that the chart's replica guard is actually reachable rather than discarded.
- `tests/unit/deploy/kubernetes-manifest-generator.test.ts` — 13 cases: generation invariants, target profiles, storage-class semantics, and parity between the generator and `deploy/kubernetes/base`.
- `tests/unit/deploy/kubernetes-provision-plan.test.ts` — 19 cases: the provisioning plan's correctness (including `--tls-san` and the kubeconfig rewrite), the deploy-spec guard rails, and the five review findings below.
- `tests/unit/deploy/kubernetes-cli-command.test.ts` — 10 cases driving the real binary: the stdout contract, exit codes, and argument rejection.

## Coverage Notes

New production code is `src/lib/deploy/kubernetes/` (three pure modules, 830 lines) and `bin/cli/commands/deploy.mjs` (250 lines).

The generator and the spec schema are pure — no cluster, no filesystem, no clock — and are covered directly by the two unit suites. The CLI is covered end to end by a suite that executes `bin/omniroute.mjs` and parses its stdout, which is what pinned the stdout-pollution bug; a mocked test would not have caught it. `bin/omniroute.mjs` gains one condition, exercised by that same suite.

Nothing existing loses coverage. Beyond the new files, the only changes are the two ConfigMap/protocol fixes, that one condition, a `.prettierignore` entry for Helm's Go templates (which Prettier's YAML parser cannot read), and two doc cross-references.

## Reviewer Notes

- **The chart's guard rails live in a `define` that `deployment.yaml` includes.** Helm parses and then discards top-level code in a `_*.tpl` file, so a bare `fail` there is a silent no-op — the first version of this had exactly that bug, and `replicaCount=2` rendered happily. A test now pins the `include`.
- **`storageClassName` is omitted, never set to `""`.** An empty string disables dynamic provisioning entirely; omitting the key binds the cluster default. The two look interchangeable, so a test covers the distinction.
- **The k3s overlay assumes the Traefik that k3s ships by default.** A cluster installed with `--disable=traefik` has no such IngressClass, and the API server accepts the Ingress anyway — it is created and routes nothing. Documented in troubleshooting rather than worked around, since the fix is to name the class the cluster actually runs.
- **Context names accept `:` and `/`.** Managed clusters produce them (an EKS context is `arn:aws:eks:<region>:<account>:cluster/<name>`). Safe because nothing here goes through a shell, the leading-alphanumeric requirement prevents a value being read as an *option*, and — the part that had to change for this to hold — the temp filename in the fetch-kubeconfig step is fixed rather than derived from the context, so a `/` cannot turn it into a path.
- **Recreate means every upgrade is a short full outage** — in-flight SSE and dashboard sessions drop, and the empty-endpoint window returns a controller `502`. That is the SQLite topology, not a probe misconfiguration, and the guide says so where an operator will hit it.
- The `--tls-san` step matters more than it looks: omit it and the cluster works perfectly from the VPS itself while every remote `kubectl` fails TLS verification, which is a confusing way to lose an afternoon.
- No migration, no feature flag, no change to the request path, no new environment variable.


insoln:
Notes from a production deployment running exactly this topology (single replica, Recreate, RWO PVC on Longhorn, ingress-nginx) — three things that bit us and may be worth folding in:

1. **Consider splitting the API and dashboard into two Ingresses.** The single-Ingress layout exposes the dashboard with the same (lack of) auth as `/v1`, while the image boots with `INITIAL_PASSWORD=CHANGEME` unless overridden. We run `/v1` + `/api/v1` on one Ingress (API-key auth, buffering off, long timeouts) and the dashboard on a separate Ingress behind oauth2-proxy — that split is impossible to express with one Ingress resource, and it's the difference between "dashboard leaked" and "dashboard behind SSO" for anyone applying these manifests on a reachable cluster. Even just a commented-out second Ingress + a line in the guide would help.

2. **`proxy-body-size: 32m` is below the app's own default body cap.** `OMNIROUTE_CHAT_HARD_MAX_BODY_BYTES` defaults to 50 MB, so nginx starts returning 413s the application itself would have accepted (and the app's structured 413 handling — compact-required semantics — never gets a chance to run). Suggest ≥ `50m`, or a comment tying the annotation to that env var.

3. **Sizing guide: mention the chat-admission knobs for agentic workloads.** With parallel coding-agent sessions (Claude Code with MCP tools), effectively every request classifies as heavyweight (>64 tools / >32k estimated tokens), and the `OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT=1` default turns a second concurrent session into a stream of retryable 503s (`chat_admission_busy`, `queue_timeout`). We run `MAX_HEAVY_IN_FLIGHT=4` / `ADMISSION_QUEUE_MS=20000` / `ADMISSION_MAX_QUEUED_BYTES=32M` on a 4Gi-limit pod without issue. A row in the sizing table would save k8s adopters a confusing first day.

Also, for the storage section: PVC sizing currently has to budget for unbounded telemetry growth (`conversation_turn_nodes`, ~190 MB/day under agentic load, no retention path) — tracked in #12453.

PauloFH:
Thanks — this is the kind of feedback that only comes from actually running it. Items 1, 2 and 4 are in as of `b6087dc47`.

**`proxy-body-size`** was a real defect: 32m against a 50 MB `CHAT_HARD_MAX_BODY_BYTES` default meant nginx rejected bodies the app would have taken, and the app's compact-required 413 never ran. Fixed in both the overlay and the generator, with a test that reads the default out of `chatBodyAdmission.ts` rather than restating it, so raising one side alone now fails.

**The Ingress split** is now `overlays/k8s/ingress-split-example.yaml` plus a guide section — `/v1` + `/api/v1` on an API host with `REQUIRE_API_KEY`, dashboard on its own host behind an external IdP. Not wired into the kustomization, same convention as the liveness-tcp patch. You are right that one Ingress cannot express it; the `INITIAL_PASSWORD=CHANGEME` default makes that worth calling out explicitly rather than leaving implicit.

**Telemetry sizing** now says the PVC is sized by retention window, citing #12453 and your ~190 MB/day figure.

On **item 3**, I think the behaviour changed under you. On `release/v3.8.51`:

```js
function resolveLegacyCountCap(): number {
  const raw = process.env.OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT;
  if (raw === undefined || raw.trim() === "") return Number.MAX_SAFE_INTEGER;
  return CHAT_MAX_HEAVY_IN_FLIGHT;
}
```

Left unset, the legacy count cap resolves to unlimited and the auto-derived ingest byte budget is what gates — the source comment says as much ("the default on every deployment that produced the multi-subagent 503 storm ... resolves to effectively unlimited"). So `MAX_HEAVY_IN_FLIGHT=4` *enables* the count cap at 4 rather than relaxing anything; unset is the looser setting. Your report mentions Hub `3.8.49`, which predates that.

I documented both knobs in the sizing section with that caveat, since the variable name reads like a ceiling you raise. If `chat_admission_busy` is still showing up on a current build with it unset, that is the byte budget binding and a different conversation — worth its own issue, and I would be interested in the numbers.

insoln:
Confirmed: we're on 3.8.50, where the bundle still reads `o(process.env.OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT, 1)` — a hard default of 1. So both hold: on ≤3.8.50 `=4` relaxes the cap, on ≥3.8.51 it enables one. We'll drop the var on upgrade. Thanks for landing 1/2/4.

insoln (commented):
Ran this exact topology in production today (generic k8s, ingress-nginx, Longhorn, ~1 GB `storage.sqlite`, agentic load) and hit the failure the liveness comment hints at — hard enough that I think two defaults should flip. Sequence: HTTP liveness (3 s timeout) killed the pod mid-startup-VACUUM → each restart grew the WAL (it reached 999 MB) → every checkpoint then stalled the event loop for ~a minute → readiness flapped → nginx 503s. tcpSocket liveness plus a bigger startup budget broke the loop. Three inline suggestions with the numbers.

deploy/kubernetes/base/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+                # SIGTERM, so new traffic stops arriving at a dying process.
+                command: ["/bin/sh", "-c", "sleep 15"]
+          # Cold start runs SQLite migrations; give it up to 150s before the
+          # kubelet starts counting liveness failures.
+          startupProbe:
+            httpGet:
+              path: /healthz
+              port: http
+            periodSeconds: 5
+            failureThreshold: 30
```
insoln: 150 s covers migrations, but cold start also replays the WAL and runs the startup-cleanup VACUUM — on a ~1 GB database on network storage that took ~5 min for us. A kill mid-VACUUM is what *creates* the oversized WAL that makes the next boot slower, so the budget failure is self-amplifying. Readiness keeps traffic away meanwhile, so a big budget here costs nothing.

```suggestion
          # Cold start runs SQLite migrations, WAL replay and the startup-cleanup
          # VACUUM; on a ~1 GB database on network storage that is minutes, not
          # seconds. Give it up to 10 min — readiness keeps traffic away, and a
          # kill mid-VACUUM bloats the WAL and makes the next boot even slower.
          startupProbe:
            httpGet:
              path: /healthz
              port: http
            periodSeconds: 5
            failureThreshold: 120
```

deploy/kubernetes/base/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+          # /api/monitoring/health — it does real DB work and false-positives
+          # under load. Under event-loop stalls, switch to tcpSocket (see the
+          # k8s overlay patch).
+          livenessProbe:
+            httpGet:
+              path: /livez
+              port: http
+            periodSeconds: 10
+            timeoutSeconds: 3
+            failureThreshold: 6
```
insoln: With a synchronous SQLite driver "under event-loop stalls" is not an edge case — every big checkpoint/VACUUM stalls the loop for minutes. We watched exactly this probe kill the pod mid-VACUUM in a loop until we swapped to tcpSocket. The asymmetry favors TCP as the default: when HTTP liveness is wrong you get a crashloop plus data-file bloat; when TCP is wrong, a hung-but-listening process merely survives until a human looks. `overlays/k8s/liveness-tcp.yaml` then becomes redundant (or can be inverted into an http-liveness opt-in).

```suggestion
          # TCP by default: with a synchronous SQLite driver, a big checkpoint or
          # VACUUM stalls the event loop for minutes — an HTTP liveness probe
          # then kills the pod mid-VACUUM, which bloats the WAL and makes the
          # next boot slower still (self-amplifying crashloop; observed in
          # production). TCP cannot be starved by the event loop. Never point
          # liveness at /api/monitoring/health — it does real DB work.
          livenessProbe:
            tcpSocket:
              port: http
            periodSeconds: 20
            failureThreshold: 6
```

deploy/kubernetes/base/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+    matchLabels:
+      app.kubernetes.io/name: omniroute
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: omniroute
+        app.kubernetes.io/component: server
+    spec:
+      # 15s preStop + 30s SHUTDOWN_TIMEOUT_MS for in-flight SSE, with headroom.
+      terminationGracePeriodSeconds: 90
```
insoln: Closing SQLite also checkpoints the WAL, and on network storage a fat WAL takes minutes — a SIGKILL mid-checkpoint carries the WAL into the next boot, feeding the startup problem above. 300 s is free when shutdown is actually fast. (The ConfigMap comment "90 - 15 = 75s" would need the same touch.)

```suggestion
      terminationGracePeriodSeconds: 300
```

diegosouzapw:
## Sweep-reds babysit (round 5 Fase B)

No push. HEAD remains `b6087dc471`. Did not merge `origin/release/v3.8.51` (base-red #12581).

### Remaining reds are inherited — not this PR's fault

- **No new ESLint warnings:** lint itself was `eslintWarnings=0, eslintErrors=0`. The job failed on **CodeQL ratchet 12 > baseline 11** (repo-wide open alerts, not this k8s-deploy diff). Same class of inherited ESLint-job red as the tip.
- **Fast Quality Gates:** `1 gate(s) failed: complexity-ratchets`. CI new-code mode walks the merge commit vs the current `release/v3.8.51` tip (file-size `chat.ts` drift / #12581). Locally, this PR's own 6 in-scope files vs merge-base `7f25d67d076b` are **0/0** (`complexityNewCode=0`, `cognitiveComplexityNewCode=0`).

Not merging. Needs the base-red drain, not a k8s rewrite.

PauloFH:
Both defaults flipped, plus a third that fell out of testing. Pushed as `e3e4960b1`.

I went looking for the mechanism in the repo before changing anything, and it is there in two places: `src/lib/db/cleanup.ts` runs `db.exec("VACUUM")` from the startup cleanup scheduler, and `closeDbInstance()` in `src/lib/db/core.ts` runs `wal_checkpoint(TRUNCATE)` on the way out. `better-sqlite3` is synchronous, so both block the event loop for as long as they take — `/livez` included, since it shares that loop. The crashloop you hit is not an unlucky timeout; it is the expected outcome of pointing an HTTP probe at a process that legitimately stops answering HTTP.

**Liveness is now `tcpSocket`.** The asymmetry argument is the one that decided it, and it is in the manifest as a comment so the next person to "fix" it back has to read the reason first. `overlays/k8s/liveness-tcp.yaml` is inverted into `liveness-http.yaml` rather than deleted — HTTP does catch a wedged server that TCP cannot, so it stays as an opt-in with the condition under which it is defensible spelled out (small database, fast local storage, verified against your own `PRAGMA wal_checkpoint(TRUNCATE)` timings).

**Startup budget 150s → 10 min**, for the reason you give: the failure is self-amplifying, and readiness holds traffic back for exactly as long as it takes, so the budget costs nothing.

**`terminationGracePeriodSeconds` 90 → 300.** Your point about the close-time checkpoint is confirmed by the ordering in `gracefulShutdown.ts`: `waitForDrain()` is what `SHUTDOWN_TIMEOUT_MS` bounds, and `cleanup()` runs *after* it with no timeout of its own. So the drain budget never covered the checkpoint at all. The ConfigMap comment is retouched to say what the headroom is actually for.

**A third one, found by running it.** On a real cluster the startup probe logged `Get "http://…/healthz": context deadline exceeded` on a trivial boot with an empty database — the kubelet's probe timeout defaults to 1s, and `/healthz` shares the same event loop as everything else. Startup now sets `timeoutSeconds: 3` explicitly.

Verified on k3s v1.36.4+k3s1, not just rendered: pod `1/1 Running`, 0 restarts, `/healthz` and `/livez` 200 through the Service. And because a 300s grace period reads like a five-minute upgrade stall, I timed a real delete — the pod terminated in **25s**. It is a ceiling, not a delay; the guide now says so, together with the `Recreate` interaction for the case where a pod really does hang.

The tests pin each default to the source fact it derives from rather than to the number: one greps the VACUUM call out of `cleanup.ts`, another asserts the drain-then-cleanup ordering in `gracefulShutdown.ts`, so moving the code without moving the manifest fails. Checked by mutation — each of the five reverts fails at least one test.

Credited you as co-author on the commit; the tcpSocket rationale is yours.

*(For anyone tracking CI here: HEAD moved from `b6087dc471` to `e3e4960b1` with this change.)*

diegosouzapw:
This is an exceptionally well-executed PR — the self-review notes (the silently-discarded
Helm `fail`, the directory-traversal fix on the kubeconfig context path, the deliberate
decision to skip a dashboard button to avoid exposing the host kubeconfig via HTTP) show real
rigor. Ran all 4 new test files live against this branch's head: 57/57 pass. Only thing to
double-check before merge is that `npm run lint`'s "zero errors but exit 2" (stale
suppressions) is confirmed pre-existing on the current release tip and not something this
branch introduced. Otherwise this looks ready.

PR #12427, opened 2026-09-02
feat(video): redact transcript text from logs and durable memory (#12150 P1)

diegosouzapw:
## What

Structured redaction of Video Bridge transcript-cue text from two retention surfaces, done at the guardrail layer so it is **parse-free and un-bypassable** — the redacted rendering substitutes each structured cue's text field before the description string is assembled (never a regex over flattened text).

- **Persisted logs:** when a request renders any transcript cue, the guardrail emits a redacted description shadow + a `videoBridgeObserved` marker + a redaction map. The attempt-log serializer swaps each video-derived text part for its redacted form **by content equality** (`text === fullText`), so it stays correct even after system-prompt/handoff/memory injection reshapes the message array. The body sent upstream to the model is unchanged.
- **Durable Memory:** an observed request populates no durable Memory — both request- and response-derived extraction are skipped, so the model's own reply cannot echo transcript text into Memory.

## How it was validated

TDD throughout (per-task adversarial review). Key regression guards:
- `videoBridgeTranscriptProvenance`/`videoBridge` — the redacted shadow drops every cue body (incl. adversarial cue text containing `]`, the placeholder, forged `transcript[...]` headers, newlines) across both the trailing-blob and audio-fusion assembly paths; `videoBridgeObserved` is false for plain video (frames, no transcript).
- `video-bridge-log-redaction` — **Scenario A**: a system message prepended AFTER the map is built still redacts the transcript and leaves the prepended message untouched (this is the content-match fix; positional matching leaked here).
- `video-bridge-memory-suppression` — request- AND response-derived extraction skipped when observed; both still run otherwise.
- `videoBridgeTranscriptCacheIdentity` — result-cache version bumped v5→v6 so a stale entry can't serve an unredacted shadow.

Focused suites green locally (15/15 log+memory, 21/21 guardrail); `typecheck:core`, `check:docs-all`, and `prettier --check` on touched files clean. Full `test:unit` deferred to CI (devbox 60-min ceiling under load). Base is green (no `base-red` issue open).

## Scope

Refs #12150 — **P1 of 2**. P2 (a separate PR) covers the raw pre-guardrail client-request snapshot in the detailed-log artifact + `previous_response_id` continuation fail-closed + its DB migration. Design: `_tasks/superpowers/specs/2026-09-01-video-transcript-retention-design.md`.


diegosouzapw:
⚠️ **CI reds here are inherited base-red, under active repair — not this branch's 6 changed files.**

Discrimination:
- **API Route Typecheck** — the only concrete regression is `src/shared/providers/webSessionCredentials.ts` (TS1005/1128/1137). That file is in **none of this PR's 5 commits**; it entered the base via #11461 (MaxAI provider) *after* this branch's merge-base (`7f25d67d07`). The same api-typecheck red is shared by unrelated open PRs (#12428, #12422, #12423), and **#12423 is "[URGENT] fix(ci): repair latest release-root regression"** — the base fix is already in flight.
- **Unit Tests fast-path (1–4/4) + Vitest (fast-path) + Fast Quality Gates** — these belong to workflow run `33597588784`, currently **queued/re-running**; the "fail" buckets are stale from a prior attempt against the broken base (logs not yet available). All-shards-red is the classic base-wide signature, consistent with the base regression above, not a targeted break.

Locally against this branch: `typecheck:core`, `check:docs-all`, `prettier --check` (touched files), and the focused suites (`video-bridge-log-redaction` + `video-bridge-memory-suppression` 15/15, `guardrails/videoBridge` 21/21) are all green. Full `test:unit` should be re-run on the repaired base.

Action: re-run CI after #12423 (or the base fix) lands; do not chase these against this branch.

PR #12428, opened 2026-09-02
chore(lint): adopt eslint-plugin-react-hooks 7.1.1

diegosouzapw:
Deliberate adoption of the react-hooks 7.1.1 rule set, unblocked by the completed #12146 migration (284 compiler-rule violations resolved in 8 batches, finished 2026-09-01).

- Exact pin `7.1.1` in both `devDependencies` and `overrides` (matching, or npm 11 rejects the install — a bare `npm install --package-lock-only` generated the lockfile precisely because naming the package on the CLI rewrites the devDep with a caret).
- `tests/unit/eslint-react-hooks-version-pinned.test.ts` re-pinned to 7.1.1 — it stays as the guard so a future group bump can never ride past a compiler-rules review (the dependabot-level ignore from #12329 stays for the same reason).

**Validation** (192.168.0.113, fresh `npm ci`, logs `/tmp/rh-*.log`): cold `NODE_OPTIONS=--max-old-space-size=8192 npm run lint:json -- --max-warnings 0` → **exit 0, zero new violations** from the 7.1.1 rules; version-pin test → 2/2.


diegosouzapw:
Merging with `--admin` with only the ~1h ESLint-warnings gate pending: that exact check is what this PR's validation already ran cold, twice, on a faithful npm ci environment (.113) — `NODE_OPTIONS=8G npm run lint:json -- --max-warnings 0` → exit 0, zero new violations from the 7.1.1 rule set (evidence in the PR body). Every other check (17) is green, including Fast Quality Gates after the window drains (#12433/#12434/#12437).

PR #11513, opened 2026-08-25
feat(providers): add UC (uncensored.com) — persona (un-metered) + direct (metered)

arminanton:
# feat(providers): add UC (uncensored.com) — persona (un-metered) + direct (metered)

Adds **uncensored.com** as two OpenAI-compatible providers, mirroring UC's own two
surfaces. The `uc` persona side is a full multimodal port (chat, tools, vision, doc-RAG,
image, video, TTS); `uc-direct` is the metered Developer API.

| Provider | Surface | Transport | Auth | Metered |
|---|---|---|---|---|
| **`uc`** | Persona / subscription | WebSocket | durable Clerk credential → short-lived per-connect token (no API key) | No |
| **`uc-direct`** | Developer API | REST (OpenAI spec) | `X-api-key` (`uai_sk_live_…`, never expires) | Yes (credits) |

Same underlying frontier models; `uc` is the un-metered subscription surface, `uc-direct`
is the paid developer API.

---

## `uc` — persona provider (WebSocket, un-metered)

There is no public API on the persona path, so the executor reproduces the web app's own
signed WebSocket turn. It's a self-contained web-app port in the style of `muse-spark-web`
/ the Clerk-cookie providers.

### Chat, auth, streaming
- **Auth** (`uc/clerkAuth.ts`, `uc/credentials.ts`): a durable Clerk `__client` cookie is
  stored in `providerSpecificData`; the executor mints a fresh **60-second `__session` JWT
  per connect** (`POST /v1/client/sessions/{sid}/tokens`), caches it ~8s before expiry, and
  follows Set-Cookie rotation. No API key.
- **Browserless email login** (`uc/emailLogin.ts`): the full Clerk email-code flow
  (`sign_ins` → `prepare_first_factor` → `attempt_first_factor`) harvests the durable
  credential — no browser, no OAuth widget. This drives the connect UI.
- **Protocol** (`uc/protocol.ts`): OpenAI `messages[]` → one persona frame (current turn as
  `text` + prior turns as `chat_history`). `max_tokens`/sampling knobs are deliberately never
  emitted (persona ignores them and `max_tokens` aborts the turn).
- **Streaming** (`uc/ws.ts`, `uc/stream.ts`): a Promise-wrapped WebSocket driver
  (Origin-only handshake, timeout/abort, injectable socket for tests) parses the
  newline-delimited frames — status, intermediary reasoning, text deltas, and the
  authoritative terminal `raw_text` — into an OpenAI SSE stream or a buffered
[...]

PR #12433, opened 2026-09-02
fix(release): repair the tip syntax break from the maxai/uc merge auto-resolve

diegosouzapw:
Third base-red window of the cycle, opened ~06:00Z by the #11461 × #11513 merge auto-resolve eating two closing lines of the `maxai` entry in `webSessionCredentials.ts` — the tip currently fails API Route Typecheck with 11 syntax errors, and every PR merge-ref cut since inherits it (first casualty: #12428).

1. Restore `],` + `},` closing the `maxai` `storageKeys` block.
2. The un-broken parser then reveals one real TS2322 the MaxAI block added in `providers/[id]/models/route.ts` — cast `connection.providerSpecificData` to the exact `Record&lt;string, unknown&gt; | null | undefined` shape `resolveMaxaiCredential` takes (pure type-level, zero runtime change; the identical neighboring callsites are baselined).

Validated locally on the tip: `check-api-typecheck` → **OK, 289 pre-existing, all baselined**; `typecheck:core` → **0 errors**.

### DIFF
--- a/src/app/api/providers/[id]/models/route.ts
+++ b/src/app/api/providers/[id]/models/route.ts
@@ -612,7 +612,10 @@ export async function GET(
 
       try {
         const discovery = await discoverMaxaiModels({
-          providerSpecificData: connection.providerSpecificData,
+          providerSpecificData: connection.providerSpecificData as
+            | Record<string, unknown>
+            | null
+            | undefined,
           accessToken: apiKey || accessToken,
           fetchImpl: (url, init) =>
             safeOutboundFetch(url, {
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Merging with `--admin` while the long gates are still pending: the tip is actively broken (11 syntax errors) and every merge-ref cut inherits it. The critical gate for this diff — API Route Typecheck — is green on CI, the diff is 2 closing lines + 1 pure type-level cast (no runtime change, no test executes differently), and the full local validation is in the PR body.

Issue #12104, opened 2026-08-30
fix(providers): Perplexity Web still cannot perform agentic tool calls despite #5937


tanveer-arch:
### OmniRoute Version

3.8.50

### Installation Method

npm (global)

### Operating System

Windows

### Provider(s) Involved

perplexity-web

### Model(s) Involved

- pplx-sonnet
- pplx-gpt-5.6-terra
- other pplx-* models tested

### Client Tool

Claude Code

---

## Description

Perplexity Web still cannot be used reliably for agentic coding through OmniRoute, even though the streaming tool-call fix from #5937 is present in v3.8.44+.

The client sends tool definitions correctly, but Perplexity Web does not emit a structured tool call. Instead it responds with ordinary natural language claiming that it cannot access the filesystem or execute commands.

**This reproduces even with `stream: false`**, so this does not appear to be the original streaming-only bug from #5927.

---

### Reproduction 1

**Tool:** `read_file(path)`

**Prompt:**
> Read README.md and summarize it.

**Expected:**
```
tool_call:
  name: read_file
  arguments:
    path: README.md
```

**Actual:**
> I don't have the necessary tool access to read the README.md right now.

---

### Reproduction 2

**Tool:** `run_command(command)`

**Prompt:**
> Run the project's test command.

**Actual:**
> I don't have access to run the project's test command from here.

---

### Reproduction 3 — Multi-turn

1. "Read README.md"
2. "Add a comment to README.md"
3. "Apply the change"

The model does not emit the required structured tool calls and eventually returns a refusal / inability-to-access-files response across all turns.

---

## Important Finding

This was also tested using OmniRoute's prompt-based `webTools` emulation with an explicit structured protocol such as:

```xml
<tool_call>
{"name":"read_file","arguments":{"path":"README.md"}}
</tool_call>
```

The model **still refused to emit the tool call**.

Therefore the failure appears to occur **before** OmniRoute's tool executor: the provider/model returns a natural-language refusal rather than a machine-readable tool request, even when the protocol is explicitly injected via prompt.

---

## Sanitized Call Log Evidence

**Request:**
```
model:   pplx-sonnet
stream:  false
tools:   29
```

**Response:**
```
tool_calls: 0
finish_reason: stop
content: "I don't have the necessary tool access to read the README.md right now."
```

Same result reproduced with `stream: true` and `stream: false` both.

---

## Expected Behavior

For supported agentic workflows, Perplexity Web should either:

[...]

PR #12442, opened 2026-09-02
fix(perplexity-web): declare tool-calling params unsupported

tanveer-arch:
**Summary**
This PR fixes an issue where providing tool schemas to a combo/auto route that falls back to `perplexity-web` results in an immediate 400 error. 

**Root Cause**
The `perplexity-web` registry configuration lacked the explicit `unsupportedParams` metadata array. As a result, OmniRoute's `chatCore` failed to recognize that the provider cannot process tool parameters.

**Proposed Solution**
Added `unsupportedParams: ["tools", "tool_choice", "parallel_tool_calls"]` to the `perplexity-web` registry entry. 
- On direct requests, this triggers `stripUnsupportedParams`, correctly sanitizing the payload before execution.
- On auto/combo fallback, it triggers `filterTargetsByRequestCompatibility`, ensuring requests requiring tool calling don't route to this unsupported backend.

**Notes for Reviewers**
- This aligns `perplexity-web` with the standard API `pplx-*` models which already declare tool calling as unsupported.
- This change strictly affects capability routing and sanitization prior to execution. It does not modify the `perplexity-web` executor protocol itself.
- A regression test (`tests/unit/12104-perplexity-web-tools-unsupported.test.ts`) is included to verify the compatibility logic integration.

### DIFF
--- a/open-sse/config/providers/registry/perplexity/web/index.ts
+++ b/open-sse/config/providers/registry/perplexity/web/index.ts
@@ -8,6 +8,7 @@ export const perplexity_webProvider: RegistryEntry = {
   baseUrl: "https://www.perplexity.ai/rest/sse/perplexity_ask",
   authType: "apikey",
   authHeader: "cookie",
+  unsupportedParams: ["tools", "tool_choice", "parallel_tool_calls"],
   models: [
     { id: "pplx-auto", name: "Perplexity Best", toolCalling: false },
     { id: "pplx-sonar", name: "Sonar 2 (via Perplexity)", toolCalling: false },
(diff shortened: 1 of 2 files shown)


diegosouzapw:
The core fix (declaring `tools`/`tool_choice`/`parallel_tool_calls` as unsupported on the
perplexity-web registry entry) is correct and the first two tests prove it cleanly. One issue
found while running the suite locally: the third test
("combo/auto routing excludes perplexity-web when tools are present") throws — its mock
`candidateTargets` are plain `{provider, model}` objects, but
`filterTargetsByRequestCompatibility` needs a real `ResolvedComboTarget` with a `modelStr`
field. Could you fix that mock shape (or drive the test through the real target-resolution
path) so it actually exercises the combo/auto routing behavior it's meant to prove? Also
flagging #12443 (same #12104 issue, different fix layer) as a natural companion — both look
safe to land together.

tanveer-arch:
Thanks for the review. I see that the requested third-test mock-shape correction has been applied in `b73c44e8`.

I’ve verified the updated regression test passes 3/3, including the combo/auto path and unsupported-parameter stripping.

No further changes are needed from my side.

PR #12443, opened 2026-09-02
fix(perplexity-web): preserve system contract on follow-up requests

tanveer-arch:
**Summary**
This PR addresses the OmniRoute-side bug described in #12104 where follow-up messages on the `perplexity-web` provider drop the system instructions containing the injected tool-use contract, causing subsequent conversational turns to fail at using tools.

**Root Cause**
When formatting a follow-up request, `buildQuery()` historically hit an early return if a `followUpUuid` was present, returning only `parsed.currentMsg` and explicitly discarding the `systemMsg` instruction block.

**Proposed Solution**
Updated `buildQuery` to correctly prepend `parsed.systemMsg` (and the `SEARCH_HINT`, if enabled) to `parsed.currentMsg` when formatting follow-up requests. This preserves the injected instructions for the lifetime of the session without interfering with the existing UI abstraction.

**Important Clarification**
This PR **only** fixes OmniRoute's explicit dropping of instructions for follow-up turns. It **does not** claim to fix upstream limitations in Perplexity's own backend where their model might refuse or mangle structured JSON during its first response. This ensures the contract is successfully sent; how the backend handles it remains upstream behavior.

**Notes for Reviewers**
- A regression test (`tests/unit/12104-perplexity-web-followup-contract.test.ts`) is included to verify that `buildQuery` correctly includes the system contract strings on Turn 2+ follow-ups.

### DIFF
--- a/open-sse/executors/perplexity-web/protocol.ts
+++ b/open-sse/executors/perplexity-web/protocol.ts
@@ -388,7 +388,9 @@ function searchHintEnabled(): boolean {
 export function buildQuery(parsed: ParsedMessages, followUpUuid: string | null): string {
   if (followUpUuid) {
     const sys = parsed.systemMsg.trim();
-    return sys ? `${sys}\n\n${parsed.currentMsg}` : parsed.currentMsg;
+    const hint = searchHintEnabled() ? `\n\n${SEARCH_HINT}` : "";
+    const contract = sys ? `${sys}${hint}` : "";
+    return contract ? `${contract}\n\n${parsed.currentMsg}` : parsed.currentMsg;
   }
 
   const obj: Record<string, unknown> = {};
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Clean, minimal fix — the early return in `buildQuery()` for follow-ups was genuinely dropping
the system contract block, and the honest scoping in the PR description (not claiming to fix
upstream Perplexity behavior) is appreciated. Local probe: 1/1 test passes. Ready to merge as-is.
Pairs naturally with #12442 (same #12104 issue, different layer) if you want to land both together.

diegosouzapw:
Thanks @tanveer-arch — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #12445, opened 2026-09-02
fix(sse): remap non-contiguous upstream tool_calls index to a gap-free output_index

hartmark:
## What Problem This Solves
Ping (an OpenClaw agent) hit a recurring, ~hourly production failure: `⚠️ omniroute/default request failed`, `rawError=Responses stream changed output item identity`. Every occurrence traced to `minimax/minimax-m3:free` via OpenRouter/GMICloud.

## Why This Change Was Made
GMICloud's stream sent `tool_calls[].index` starting at 1 (never 0) for a turn with two calls. `emitToolCall`/`closeToolCall` added that raw index directly onto `toolCallOutputIndexBase()`, so the emitted Responses-API `output_index` sequence was `0` (message), `2`, `3` — skipping 1.

OpenClaw's own Responses-stream client reads `response.completed`'s final `output[]` array by **array position**, expecting position === `output_index` (the Responses API's own contract, not a bug on OpenClaw's side — confirmed by direct inspection of `packages/ai/src/transports/openai-responses-stream-slots-internal.ts`). With the gap, `output[1]` (the real call at `output_index=2`) got looked up under index 1 and missed; `output[2]` (the second call, real `output_index=3`) got looked up under index 2, landed on the first call's tracked slot, saw a different `call_id`, and correctly threw — aborting the whole turn with no retry.

Confirmed the exact mechanism two ways: replayed the real captured stream through OmniRoute's own translator to get the actual emitted event sequence, then fed that sequence directly into OpenClaw's real `processResponsesStream()` and reproduced the identical throw with a full stack trace pointing at the exact line.

Fix: `resolveLocalToolCallIndex()` remaps each turn's raw upstream tool-call indices to a local, contiguous, 0-based sequence in first-seen order, used at both `emitToolCall` and `closeToolCall` so they can no longer drift apart from the raw upstream numbering.

## User Impact
Stops a real, recurring production failure (confirmed ~hourly for the affected model) that silently dropped an entire agent turn with no client-side retry.

## Evidence
- Regression test (anonymized real-incident shape) added to `tests/unit/translator-resp-openai-responses.test.ts`, proven RED against unfixed code (`got [0,2,3]`) and GREEN after.
- Full translator suite: 783/783 passing, no regressions.
- Root cause cross-verified against OpenClaw's own client source, not assumed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12446, opened 2026-09-02
fix(quality): detect duplicate tool_calls entries in one response

hartmark:
## What Problem This Solves
Same live incident as #12445 (minimax-m3:free duplicating a `heartbeat_respond` call verbatim in one turn) surfaced a separate, real gap: the response-quality gate had no check for a model emitting the *same* tool call twice with identical arguments in one message. A response with real content and real (duplicated) tool_calls passed every existing check.

## Why This Change Was Made
`findToolCallSpecViolation()` detects `tool_calls` entries that repeat the exact same function name + arguments verbatim within one message — always a bug, no legitimate use calls the same tool twice with identical args in one turn.

Wired in two places:
- `validateResponseQuality`'s non-streaming JSON branch: fails the attempt over to a sibling combo target, same as any other quality-gate rejection.
- `attemptLogging.ts`'s `persistAttemptLogs` (runs for every completed request, streaming or not): on a violation, writes a `provider.spec_violation` entry to the existing `audit_log` table (reuses `provider.warning`'s mechanism — no schema change). This is the only thing that can help a *streaming* response — the quality-gate peek only ever validates the START of a stream (buffering the whole response first would defeat streaming's latency purpose), so it can't see a duplicate that only streams in after real content already started relaying to the client. Too late to fail that attempt over, but a durable, queryable signal beats a clean 200 with no trace.

## User Impact
- Non-streaming duplicate tool_calls now fail over to a sibling model instead of reaching the client.
- Streaming duplicate tool_calls (can't be un-sent) now leave a durable audit trail instead of silently counting as a clean success.

## Evidence
Manual verification against the real captured incident payload (duplicate `heartbeat_respond` call, same arguments) — detector correctly flags it; verified it does not false-positive on the existing non-duplicate parallel-tool-call test fixtures in the same file.


diegosouzapw:
**sweep-reds / babysit:** `No new ESLint warnings` is inherited from base-red #12581 — CodeQL ratchet `12 open > baseline 11` (repo-wide Security alerts, not this diff). No unused/`@eloqnt` findings here, so no ESLint code change.

Own unit 4/4 (`source-no-raw-nul-bytes`) is fixed in `f68f0878cf`: the duplicate-tool-call signature separator is now `"\u0000"` instead of a raw NUL byte. Did not merge `origin/release/v3.8.51`.

diegosouzapw:
**sweep-reds round 6 / babysit:** own complexity on `attemptLogging.ts` (3 → 4 / cognitive 1 → 2) extracted in `3f8521f5ad` (`toolCallSpecViolationAudit.ts`; persistAttemptLogs wiring unchanged; characterization tests for `provider.spec_violation`). Remaining FQG is inherited from base-red #12581 — did not merge `origin/release/v3.8.51`.

- **file-size:** `src/sse/handlers/chat.ts: 2434 > congelado 2424` — not this PR (`origin/release...HEAD` is `attemptLogging.ts` + `validateQuality.ts`).
- **complexity-ratchets:** CI merge-base `7802f6ea` reported **118 changed files**; local three-dot is **2**. `tlsClientBase.ts` / `tlsClient.ts` are not this diff.

Own extract should drop the attemptLogging bump; leftover FQG waits on #12581.

PR #12447, opened 2026-09-02
fix(conversations): resolve turn content OmniRoute never sends back to the client

hartmark:
## What Problem This Solves
`/dashboard/conversations`'s turn tree showed many turns as "Tool (empty)"/"Assistant (empty)" for genuine-continuation conversations, even though the actual content clearly existed and was served correctly.

## Why This Change Was Made
`resolveTurnDisplayContent` only ever read a turn node's originating `clientRawRequest.body` — fine for a full-history-resend conversation (every later request's body eventually contains everything), but a genuine-continuation turn's own request only ever carries the NEW delta. That silently left the model's own generated tool_use/text turns unresolved — their content never appears in any request body, only in a response.

Verified against a real 133-node continuation conversation: request-only resolution left 131/133 nodes unresolved.

Added two more sources, read from the SAME already-loaded artifact (no extra I/O):
- the call's own `clientResponse.output` (the model's own generated turn — 131 → 76 unresolved)
- the call's own `providerRequest.body` (the server-reconstructed full history actually forwarded upstream for non-Responses-native providers, Chat Completions `messages` shape) — best-effort, translated text can differ byte-for-byte from what was originally hashed, so this doesn't close 100%, but it does close the "Tool (empty)" case entirely (76 → 21, all 112 tool nodes resolved; the remaining 21 are plain user/assistant text turns old enough to predate both available artifacts — genuinely lost history, not a resolution gap).

## User Impact
The conversation tree now renders real content for tool calls and assistant turns in continuation-mode conversations instead of empty placeholders.

## Evidence
Verified against a real production conversation (133 turn nodes): before/after unresolved-node counts (131 → 76 → 21) computed by replaying the exact resolution logic against the real stored artifacts.

### DIFF
--- a/open-sse/services/conversationTurnContent.ts
+++ b/open-sse/services/conversationTurnContent.ts
@@ -23,6 +23,56 @@ export type TurnDisplayContent = {
   toolName: string | null;
 };
 
+type CanonicalTurnLike = {
+  role: string;
+  text: string;
+  blockKind: "text" | "tool_use" | "tool_result";
+  toolName: string | null;
+};
+
+/**
+ * extractCanonicalTurns's Chat Completions branch only reads a message's
+ * `content` -- a tool-calling assistant message carries its call in
+ * `tool_calls` instead with `content: null`, so it silently produces no turn
+ * at all and the matching conversation_turn_nodes row can never resolve.
+ * Deliberately scoped to this read-only display path instead of extending
+ * extractCanonicalTurns itself: that function also drives
+ * conversationTracker.ts's write-path identity/hashing, and this codebase's
+ * only caller of it there (chat.ts's resolveConversationId) always feeds the
+ * client-facing Responses-API body -- never Chat Completions
+ * `messages`/`tool_calls` -- so extending it there would be unreachable for
+ * real traffic here but still carries real write-path identity-hash risk for
+ * any other caller/format that function might ever serve. Mirrors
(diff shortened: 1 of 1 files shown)


diegosouzapw:
**sweep-reds / babysit:** `No new ESLint warnings` is inherited from base-red #12581 — CodeQL ratchet `12 open > baseline 11` (repo-wide Security alerts, not this diff). No unused/`@eloqnt` findings here, so no ESLint code change.

Own FQG (complexity + open-sse typecheck TS2345) and API Route Typecheck are fixed in `bae2cab01f` by splitting `conversationTurnContent.ts` helpers and typing turns as the canonical role union. Did not merge `origin/release/v3.8.51`.

diegosouzapw:
**sweep-reds round 6 / babysit:** remaining FQG after the typecheck/complexity split `bae2cab01f` is inherited from base-red #12581 — not this turn-content diff. Did not merge `origin/release/v3.8.51`.

- **file-size:** `src/sse/handlers/chat.ts: 2434 > congelado 2424` — this PR does not touch `chat.ts` (one file: `conversationTurnContent.ts`).
- **complexity-ratchets:** CI merge-base `7802f6ea` reported **118 changed files**; local three-dot is **1**. Listed regressions (`attemptLogging.ts`, `tlsClientBase.ts`, `tlsClient.ts`) are not this PR's file.

STOP on inherited FQG. Own split already landed.

PR #12448, opened 2026-09-02
feat(dashboard): parent-link, genuine-continuation badge, and modal perf fixes

hartmark:
## What Problem This Solves
Three related dashboard gaps, all on the same continuation-store surface:
1. No way to click through from a continuation request to the parent response it continued.
2. No way to tell, at a glance, which conversations on `/dashboard/conversations` are genuinely using HTTP continuation (`previous_response_id`) vs. just being tracked as multi-turn by the client-side content-hash tracker while still resending full history.
3. The conversation modal's auto-refresh (configurable down to 1s) re-fetched and re-annotated the entire up-to-100-row list every tick just to keep one open conversation's summary fields fresh.

## Why This Change Was Made
1. **Parent link**: `extractPreviousResponseId`/`resolveCallLogIdByResponseId` added to `responsesContinuationStore.ts`, wired into `/api/logs/[id]` — a request that used continuation now links to the call-log row that produced its parent response.
2. **Genuine-continuation badge**: `isGenuineContinuationTurn()` reads the row's own artifact to check whether the latest turn's `previous_response_id` actually resolved server-side. Permanently memoized by `artifactRelPath` — the underlying artifact is immutable once written, so the answer can never change once computed. Without this, the 1s auto-refresh made the badge's own cost the dominant cost of the whole list endpoint.
3. **Modal perf**: added `GET /api/conversations/[id]` (single-row, reuses the same annotation helpers as the list route) and switched the modal's poll to it while open, falling back to the full list poll only when no modal is open.

No schema changes — all three reuse existing indexed columns (`call_logs.response_id`, `.artifact_relpath`, `.api_key_id`) already added for the continuation store, or existing tables.

## User Impact
- Faster, clickable navigation between a continuation request and its parent.
- At-a-glance visibility into which conversations are actually using continuation.
- `/dashboard/conversations?limit=100` went from ~800ms cold / repeated-every-second cost to effectively free once the (permanent, correctness-safe) cache is warm; the conversation modal no longer hits the 100-row list endpoint at all while open.

## Evidence
- Verified all touched routes compile and serve cleanly (401 auth gate, not 500) after each change, live.
- Verified the badge logic against 8 real production conversations: correctly resolved `true`/`false` matching each one's actual continuation usage.
- Verified via real request logs that the modal's poll now hits the new single-row endpoint (10–33ms) instead of the full list.

### DIFF
--- a/src/app/(dashboard)/dashboard/conversations/page.tsx
+++ b/src/app/(dashboard)/dashboard/conversations/page.tsx
@@ -26,6 +26,11 @@ interface ConversationRow {
   // streaming (call_logs only gets its row on completion). Used to poll
   // /api/logs/[id] for this conversation's live partial assistant text.
   activeCallLogId: string | null;
+  // Whether the latest turn actually used previous_response_id and it
+  // resolved server-side — distinct from this row existing at all, which
+  // only means the client-side content-hash tracker saw >= 2 turns
+  // regardless of transport (see isGenuineContinuationTurn).
+  isGenuineContinuation: boolean;
 }
 
 // Same spinner used for an in-flight request on /dashboard/logs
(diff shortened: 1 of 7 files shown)


diegosouzapw:
**sweep-reds / babysit:** `No new ESLint warnings` is inherited from base-red #12581 — CodeQL ratchet `12 open > baseline 11` (repo-wide Security alerts, not this diff). No unused/`@eloqnt` findings here. Skipping a code change. Did not merge `origin/release/v3.8.51`.

diegosouzapw:
**sweep-reds round 6 / babysit:** re-checked `No new ESLint warnings`. `eslintWarnings=0` / `eslintErrors=0`; the job still fails on CodeQL ratchet `12 open > baseline 11` (repo-wide Security alerts, not this dashboard diff). Same inherited #12581 signal as last round. No unused/`@eloqnt` in this diff. Did not merge `origin/release/v3.8.51`. STOP.

PR #12464, opened 2026-09-02
feat(providers): clean-room ChatGPT Web session provider (chatgpt-session)

diegosouzapw:
## Summary

Adds `chatgpt-session` (alias `cgpt-session`), a **clean-room** provider that serves ordinary
`/v1/chat/completions` requests from an authenticated ChatGPT browser session.

This restores the *capability* removed with the common `chatgpt-web` provider in #11754 without
restoring any of its code. The retired ids `chatgpt-web` / `cgpt-web` stay retired: migration
168, the fail-closed `410 PROVIDER_RETIRED` guards and every retirement test are untouched, and
a test here asserts the new ids are not the retired ones.

The browser interaction reuses the already-vendored MIT implementation under
`open-sse/vendor/codex-chatgpt-web/`. **No proof-of-work, sentinel or TLS-impersonation code was
written** — a real signed-in browser performs those steps. The unclearable provenance of the old
pre-key/proof-of-work implementation is what forced the retirement, so this design removes that
surface rather than recreating it.

## Clean-room provenance

Two isolated tracks, no shared context. One agent was authorised to read the retired code and
produced a behaviour-only specification — no code, no symbol names, no algorithms, with 14 topics
deliberately omitted as implementation detail; it was audited for leakage before release. Separate
agents implemented from that specification plus live repository code only, each carrying a verbatim
prohibition on reading the history of any path deleted by `9c5dd676` / `7d57d9f4`.

The GPL-3.0 `Chat2API` mirror in the local references tree was forbidden and never opened.
`THIRD_PARTY_NOTICES.md` is unchanged — no new third-party code enters. Specs live in the private
`_tasks` repo at `superpowers/specs/2026-09-02-chatgpt-session-{design,behavior}.md`.

## Feasibility spike — run, and it found a blocker

The design rested on an unproven assumption: that the vendored adapter would accept a synthetic
request without a Codex thread/turn envelope. **That assumption was wrong**, and the spike proved
it against the real adapter:

```
Error: ChatGPT web requires native Codex turn_id metadata for browser-turn retry budgeting
```

As originally written this provider would have failed on **every single request**. No unit test
could catch it, because they all mock the adapter call — which is exactly why the spike existed.

The documented fallback (dropping to the raw browser worker) turned out to be unnecessary.
Synthesizing a minimal Codex turn envelope is enough, and `buildParsedRequest` now does it:
turn/thread ids in `client_metadata`, a Responses-shaped `input` array built from the same
messages, and the current-turn passthrough id on the last user item. The executor no longer hands
the adapter the OpenAI body as `_rawBody`, and that parameter was removed so the wrong call is
unrepresentable.

Verified through the production code path, not just unit tests: the adapter now reaches
`stage=browser_page` and fails only on `ChatGPT web login state is missing`, the expected
authentication failure for an unauthenticated session.

The spike also **empirically confirmed** the commentary-banner defect that had been found by
reading the vendor source. The captured stream was:

```
heartbeat, assistant_boundary,
text_delta[phase=commentary] "⚠️ ChatGPT Web Instant cannot access the local Codex computer…",
assistant_boundary
```

That banner would have prefixed every answer. The bridge now drops commentary-phase deltas.

## What changed

- Executor plus six focused modules: model routes, message translation and envelope synthesis, an
  event→OpenAI response bridge, error classification, and a runtime seam that makes every side
  effect injectable so no unit test launches a browser.
- Registry entry, two lazy executor aliases, regenerated goldens, reserved-prefix ratchet.
- Dashboard card, credential descriptor, icon, and a connection validator.
- `chatgpt-session` shares the browser-session credential lifecycle with `chatgpt-web-codex`
  behind one predicate, so a pasted cookie is replaced by verified session state and the temporary
  validation directory is cleaned up. Client and server use the same predicate so they cannot drift.
- Provider guide `docs/providers/CHATGPT_SESSION.md`; provider count 352 → 353.
- `OMNIROUTE_CHATGPT_SESSION_STREAM_OPEN_TIMEOUT_MS` (default 30000) bounds how long the response
  is withheld while the two paths' failover status is decided.

Seven routes (`luna`, `think`, `instant`, `medium`, `high`, `extra-high`, `pro`), each pinning one
backend model and one reasoning effort; a route the account does not expose fails closed. Tool
calling is prompt-emulated, the same contract `perplexity-web` and `gemini-web` use.

## Other defects found by review and fixed here

- The dashboard could never create or update a connection: the modals' credential encoding was
  still keyed to `chatgpt-web-codex` while the API routes had been widened, so every save hit a
  `JSON.parse` on a raw cookie and returned 400.
- Streaming and buffered paths disagreed on failover status when a turn died after reasoning-only
  output.
- A 503 carried by an adapter error event tripped the whole-provider circuit breaker for want of a
  cooldown hint.
- Unrecognised content parts were silently dropped, so the model could answer about content it
  never received. An assistant `refusal` part is now folded in as text rather than rejected.

**One deliberate contract change.** In `classifyChatGptSessionError`, an explicit numeric status
now outranks message-pattern matching — the vendor documents this on the field itself
("Authoritative upstream/proxy status when known; avoids message-based classification"). The old
order caused real harm: an error carrying 503 whose prose said "please sign in" was classified 401,
marking credentials expired and pulling a healthy account out of rotation. One test inverts as a
result; it is renamed and carries a comment recording the change, so it cannot be mistaken for a
test weakened to reach green.

## Still owed before merge

**Live turn validation** (Hard Rule #18). The environment is now prepared — Chromium was already
cached and Xvfb has been installed — so only a ChatGPT cookie is missing. A runbook with the exact
commands, and the three things to check beyond "it answered", is in the branch working notes.

Two behaviours remain unproven until then: that the commentary banner really is suppressed end to
end, and that a tool turn is answered with a tool call rather than a refusal about local-tool
access. The adapter wraps every turn in Codex task framing, which competes with the prompt-emulated
tool contract; this is recorded in the provider guide as unconfirmed.

## Base drift

The base has advanced 55 commits since this branch forked and now reports 355 providers. This PR
bumps to 353. A base merge is needed before landing, and the count files (README, `AGENTS.md`,
`llm.txt` plus its 42 mirrors, several SVGs) will conflict mechanically. The new
`docs/reference/REMOVED_PROVIDERS.md` blocklist was checked: no ChatGPT entry, so it does not block
this provider.

## Validation

| Gate | Result |
| --- | --- |
| `npm run lint` | PASS |
| `npm run typecheck:core` | PASS |
| `npm run check:dashboard-typecheck` | PASS — 207 pre-existing, within frozen baseline |
| `npm run check:provider-consistency` | PASS |
| `npm run check:cycles` | PASS |
| `npm run check:docs-all` | PASS |
| `npm run test:vitest` | **Inherited red** — see below |
| `npm run test:unit` | One real failure, fixed; remainder inherited or load artifacts |

Focused suites: **130 passing** across the eight `chatgpt-session` suites, the new provider-route
test and the reserved-prefix ratchet, plus all four `chatgpt-web` retirement guards, untouched.

`test:vitest` is red on the base, not from this branch: the identical command on `release/v3.8.51`
fails **17 tests across 10 files**, this branch fails **16 across 9**. The failures include
`tests/unit/encryption.spec.ts` and an MCP audit database-shutdown test, neither of which can be
affected by registering a provider. No "Release branch not green" issue is open, so this red
appears untracked.

`test:unit` triage: one failure was genuinely this branch's — the `RESERVED_PREFIX_COUNT` ratchet
moved 400 → 402 because the new id and alias are two members — and it is fixed with membership
assertions added. Everything else passes in isolation or fails identically on the base.


diegosouzapw:
Superseded by #12239, which landed on `release/v3.8.51` today and restores ChatGPT Web via its own clean-room browser transport. That PR reclaims the canonical `chatgpt-web` id and the retirement set has shrunk to just `cgpt-web`.

The two efforts were independent clean-room restorations of the same capability, running in parallel. #12239 merged first, so this one is being closed rather than merged — shipping both would put two ChatGPT Web providers in the catalogue.

Recorded for whoever picks this up later, since these three pieces exist here and not in #12239:

1. **The vendored adapter rejects a request without a Codex turn envelope.** Proven against the real adapter: `ChatGPT web requires native Codex turn_id metadata for browser-turn retry budgeting`. Any browser-transport provider driving `createChatGptWebAdapter` must synthesize `_rawBody.client_metadata["x-codex-turn-metadata"]` (thread/turn ids), a Responses-shaped `_rawBody.input` array, and `internal_chat_message_metadata_passthrough` on the current-turn user item. Without all three it fails before any browser work.
2. **The adapter emits a Codex read-only banner on every turn**, as a `text_delta` with `phase: "commentary"` bracketed by `assistant_boundary`. Captured live: "⚠️ ChatGPT Web Instant cannot access the local Codex computer in this turn…". A bridge that ignores `phase` will prefix every answer with it.
3. **An explicit upstream status should outrank message-pattern matching** in error classification — the vendor documents this on the field itself. The reverse order classifies a 503 whose prose says "please sign in" as a 401, which marks credentials expired and pulls a healthy account out of rotation.

Design and behaviour specs, the spike verdict and the live-validation runbook are in the private `_tasks` repo under `superpowers/specs/2026-09-02-chatgpt-session-*`.

PR #12471, opened 2026-09-02
fix(vertex): discover and route partner models correctly

JxnLexn:
## Summary

- support both Vertex API keys and Service Account JSON credentials without mixing their capabilities
- infer and persist the Google Cloud consumer project for project-scoped API-key partner inference
- use API-key model discovery only for Gemini and fall back intentionally to a curated project-scoped catalog because Google rejects API keys for publishers.models.list
- fetch the live Model Garden catalog for Google, xAI, Anthropic, Mistral, and open-MaaS publishers with OAuth or Service Account credentials using Google's maximum page size of 300
- treat the authenticated Google publisher catalog as the live source for current Gemini chat models, including Gemini 3.5, 3.6, 3.7, and 3.8 Flash
- retain successful publisher results when the separate Gemini Developer API rejects Service Accounts
- accept requestAccess-backed MaaS model cards while excluding image, TTS, live, transcription, embedding, OCR-only, and retired Gemini entries from the chat catalog
- replace invalid speculative Vertex IDs with documented MaaS IDs such as deepseek-ai/deepseek-v3.2-maas, qwen/qwen3-next-80b-a3b-instruct-maas, and zai-org/glm-5-maas
- route Gemini, Anthropic, Mistral, xAI, and generic OpenAI-compatible MaaS publishers through their correct transports
- label the dashboard credential field accurately and explain that Service Account JSON enables live Model Garden discovery
- reject OAuth client configuration JSON with a specific explanation instead of reporting missing Service Account fields

## Authentication behavior

A project-bound API key can execute partner models, but Google does not allow it to call Model Garden publishers.models.list. OmniRoute extracts the bound project from structured Google error metadata and uses the curated project-scoped catalog.

A Service Account JSON credential is exchanged for an OAuth access token. That bearer token can list publishers/google/models and the partner publisher catalogs, so successful discovery now returns catalogMode=live_vertex_catalog without supplementing the result with a stale static Gemini baseline.

Google separately rejects Service Accounts on generativelanguage.googleapis.com. That expected rejection no longer marks the entire Vertex discovery as failed.

An OAuth web-client JSON file is only client configuration and is not itself an access credential. It requires a complete authorization-code and refresh-token flow and is now rejected with a clear message.

## Reproduction covered

The live Publisher Model list request used pageSize=1000 although Google enforces a maximum of 300. Every publisher returned HTTP 400, leaving an empty discovery result and causing Sync Models to report:

Vertex model catalogs unavailable — using local catalog

The corrected request returns the live Google and partner catalogs. Gemini chat filtering is capability-oriented: current Flash/Flash Lite/Pro variants from Gemini 2.5 onward are imported, while non-chat and retired entries are excluded.

## Verification

- focused Vertex Node tests on the release branch: 24 passed
- focused provider-dialog tests from the credential UI change: 20 passed
- npm run typecheck:core
- live Service Account token exchange: HTTP 200
- deployed fork image revision fed218ecf5a627b329acbc8b6b42e30c46dc7bf1
- deployed live discovery: source=api, catalogMode=live_vertex_catalog, 29 synchronized models, no warning
- deployed live catalog includes gemini-3.5-flash, gemini-3.6-flash, gemini-3.7-flash, gemini-3.8-flash, claude-opus-4-1, and mistral-medium-3
- deployed catalog excludes image, TTS, retired Gemini 1.5, and the former speculative partner IDs
- deployed vertex/gemini-3.7-flash model test: HTTP 200


diegosouzapw:
This is a real, well-documented fix (the Model Garden `pageSize<=300` limit is a genuine bug)
and the architectural extraction into `vertexModelDiscovery.ts`/`vertexModelMetadata.ts` is a
welcome cleanup. Local probe: 27/27 focused tests pass. Two things need fixing before merge:
(1) ESLint reports 5 real errors introduced by this PR (3 `no-explicit-any` + 2
`no-unused-vars` in `executors/vertex.ts` and `registry/vertex/index.ts`) — these match the
CI reds on the PR, they're not stale-base noise. (2) This PR removes the exact code block that
#12332 patches in `models/route.ts` (a smaller, already-mergeable fix for the same
Express-key-400 class of bug) — your new `vertexDiscovery.ts` module already implements the
equivalent `intentional: true` behavior, so no logic conflict, just a rebase once #12332
lands. Happy to help coordinate the merge order.

JxnLexn:
Thanks — reproduced all five ESLint errors, then removed the three explicit-any annotations and marked the two unused parameters explicitly. The touched executor, registry, discovery modules and route tests now pass targeted ESLint; `npm run typecheck:core` also passes.

I updated this branch to release tip ac52d4d9e (merge commit 5b8f3380f, preserving the branch history and current upstream documentation). Changelog integrity now passes as well.

The joint check with still-open #12332 found one behavioral gap beyond the merge conflict: the extracted API-key path also marked transient HTTP/network failures as intentional catalog fallback. Fixed in b49d62418: 400/401/403 catalog rejection remains intentional; empty discovery caused by 429/5xx/network failure follows the degraded fallback. Warnings identify the HTTP rejection without copying raw upstream error bodies. New route regressions cover these cases.

Standalone validation: **28/28 passed** across `vertex-xai-models`, `vertex-models-route`, `vertex-publisher-models-parser`, `vertex-model-metadata` and `vertex-anthropic-models` (all under `tests/unit/*.test.ts`).

Merge-order proposal: land #12332 first, then retain the `maybeHandleVertexModelDiscovery(...)` delegation when resolving this PR's route conflict. In a separate local combined checkout, **26/26 tests pass**, including #12332's four tests, after adapting only its obsolete transport/warning expectations:
- Assert the key is in `x-goog-api-key` and **not** in the URL (this PR intentionally moved it out of URL-bearing logs).
- Assert HTTP 400/403 plus the curated Express warning rather than hard-coding `API_KEY_INVALID` for every rejection.
- Keep the intentional 400/403, non-intentional 500, and unchanged service-account discovery assertions.

No commits were pushed to #12332 and it was not merged into this PR. Its branch remains open, so the final integration refresh must happen after its actual merge. Full PR CI has not been awaited.

PR #12484, opened 2026-09-02
fix(cli): escalate the readiness probe timeout so a slow health response is not a phantom boot failure

dmlanday:
## Problem

`omniroute serve` prints

```
⚠ Server did not respond within 60s. It may still be starting, or may have failed silently.
```

over a server that is **up and serving traffic**. In the reported case the process went on to
proxy real requests for the rest of the session; only the readiness signal timed out.

## Root cause

Every probe of `/api/monitoring/health` was aborted at a fixed `AbortSignal.timeout(2000)`, and a
timed-out probe is classified `"hanging"`, which `waitForServer` refuses to count toward readiness
(the #6800 guard). That puts a hard 2s ceiling on how slow a *healthy* first response is allowed to
be, and the poll cannot recover once it is crossed:

- the abort tears the socket down before the route finishes, so all of its work is discarded and
  the route's own 1s payload cache is never populated;
- 500ms later the next probe restarts the same work into the same 2s ceiling;
- repeat for the whole 60s budget.

A cold Windows boot crosses that ceiling easily: this route resolves ~10 dynamic imports and reads
the DB before it can answer. On an idle machine the first response measures ~2.5-4s.

Same failure family as #10508, which fixed it by taking a DNS lookup *out* of the 2s budget rather
than by widening the budget.

## Fix

- `bin/cli/utils/pid.mjs`: `pollHealthOnce` accepts a `probeTimeoutMs`, and `waitForServer`
  escalates it after each hang (2s, 4s, 8s, 15s), clamped to the time left in the budget so the
  caller's total timeout still holds. **Only a hang escalates**, so #6800's guarantee is unchanged:
  a socket that accepts TCP and never answers still resolves `false`.
- `waitForServer` reports each probe outcome to an optional `onOutcome` callback.
- `bin/cli/commands/serve.mjs`: the readiness-timeout diagnostic uses that outcome to say whether
  the port was accepting connections, separating "up and still warming up" from "never bound the
  port". That is the line that makes this failure diagnosable instead of alarming.

## Tests

New `tests/unit/waitForServer-slow-first-response.test.mjs`:

1. a health route answering 200 in 3.2s is reported ready (**fails on `main`**: the old poller ran
   12 probes over 30s and returned `ready=false` every time);
2. escalation stays clamped to the caller's budget (a never-answering socket with an 8s budget
   returns `false` in ~8.5s, not ~10s+);
3. the timeout diagnostic distinguishes a listening port from a dead one.

Existing guards all still pass (`cli-waitForServer.test.mjs`, `waitForServer-tcp-fallback-6800.test.mjs`):
#2460, #6800 and #11766, 9/9 green. The #10508 source-text guard is unaffected (host list and fetch
template unchanged).

End to end against a real install, the patched poller reports ready in 3.9s with the probe trace
`not-listening -> hanging -> ready`; the escalated second probe is what catches it.

## Follow-up (not in this PR)

The CLI polls the **heavy** `/api/monitoring/health` every 500ms, whose own docstring points
high-frequency pollers at `/api/health/ping` instead, and that is exactly what the Electron
readiness poller (`electron/lib/serverReadiness.js`) already uses. Switching the CLI probe route
would remove the cost that creates this ceiling, but it changes the readiness contract and the
#10508 guard, so it is left as a separate change.

https://claude.ai/code/session_01CJf2dxEpiwZqyZujWk57T2


diegosouzapw:
Hi @dmlanday — this is a well-diagnosed, real bug (the per-probe 2s AbortSignal ceiling
vs the total budget) and your test suite passes cleanly at your PR head. Since you
opened this, #13433 (a different, complementary PR) already merged the "configurable
total readiness budget" feature into the same two functions
(`waitForServer`/`reportReadinessTimeout` in `bin/cli/utils/pid.mjs` and
`bin/cli/commands/serve.mjs`), so this will need a manual rebase to reconcile both — your
per-probe escalation is still valuable and distinct from that change (different axis:
per-probe ceiling vs total budget). Marking `fix-in-place` for the rebase; will follow up
once it's reconciled.

diegosouzapw:
Thanks @dmlanday — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

Issue #9455, opened 2026-08-04
fix(backend): omniroute stop reports success but supervisor respawns child server


TBloyd:
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 Pro (10.0.26200.8875)

### Node.js Version

v25.9.0

### Provider(s) Involved

N/A - not provider-specific

### Model(s) Involved

N/A

### Client Tool

N/A - the `omniroute` CLI itself

### Description

`omniroute stop` prints `Server stopped.` and exits 0, but the gateway is still listening on :20128 afterwards. Two separate defects in the same command produce this false success.

**1. `stop` terminates the server child but not its supervisor, which immediately respawns it.**

When the server is started with `omniroute` (foreground supervisor), the process tree is:

```
node .../bin/omniroute.mjs         <- supervisor  (PID 31716)
  └─ node .../dist/server-ws.mjs   <- server      (PID 29744, listening on 20128)
```

`runStopCommand()` reads `<dataDir>/server/.pid` — which `processSupervisor.mjs` (~L66) populates via `writePidFile("server", this.child.pid)`, i.e. the **child** PID only. Stopping that child trips the supervisor's restart logic, and a new server child is spawned within seconds. The supervisor then rewrites `.pid` to the new child, so the PID file always looks correct afterwards and the problem is easy to miss.

Net effect: `stop` reports success, exit code is 0, and the gateway keeps serving on a **new** PID. Running `stop` repeatedly just cycles the child PID. There appears to be no way to stop a detached instance from a second terminal — only Ctrl+C in the terminal owning the supervisor works.

I have only verified this on Windows, but from reading `processSupervisor.mjs` the respawn path looks platform-independent.

**2. On win32 the port-based fallback is a no-op that still reports success.**

If `.pid` is absent, `runStopCommand()` (stop.mjs ~L49-57) takes the fallback branch — but `killByPort()` (~L63-64) is:

```js
async function killByPort(port) {
  if (process.platform === "win32") return;
  ...  // lsof-based; lsof does not exist on Windows anyway
```

It returns immediately on Windows, after which the caller unconditionally prints `Server stopped.` and returns 0 without having touched anything.

[...]

PR #12485, opened 2026-09-02
fix(cli): preflight the port before serving so a second instance cannot de-register the first

dmlanday:
## Problem

Starting `omniroute serve` against a port another OmniRoute already owns printed three identical raw Node stack traces and no explanation:

```
⚠ Server exited (code=1). Restarting in 1s... (1/2)
⨯ Failed to start server
Error: listen EADDRINUSE: address already in use 0.0.0.0:20128
```

The conflict was handed to the child process, so it only surfaced after the child had been spawned and retried twice on the supervisor's restart budget, and it never named the process actually holding the port.

## The real damage

Both spawn attempts happen **after** `writePidFile("supervisor")` and the failed child's `cleanupPidFile("server")`. So an instance that cannot bind overwrites the pid files of the healthy instance that *does* own the port:

- `supervisor/.pid` ends up pointing at the dead starter
- `server/.pid` is deleted outright

which de-registers a server that is up and serving. Observed live: healthy server `19348` under supervisor `11108`, while `supervisor/.pid` read `21440` (dead) and `server/.pid` was gone. `omniroute stop` still worked, but only by falling through to its `killByPort` port fallback rather than by the pid files.

## Fix

`serve` now resolves the port owner **before** spawning anything or touching a pid file:

```
✖ Port 20128 is already in use by PID 19348.
  Another OmniRoute is most likely already serving there, so open http://localhost:20128 before starting a second one.
  To replace it:    omniroute stop, then start again
  To run alongside: omniroute serve --port <other-port>
```

That is real output from the patched code run against an actual conflict.

Discovery is `findListeningPids()` in `bin/cli/utils/pid.mjs` — netstat on win32, lsof elsewhere. It mirrors `killByPort()`'s discovery in `stop.mjs`; consolidating the two is worth doing next time that file is touched, and is deliberately left out of this PR to keep `killByPort`'s injected-deps tests (#9455) untouched. A discovery failure reports the port as **free**, since a false "busy" would block a legitimate start, the worse of the two failures.

## Tests

New `tests/unit/cli-serve-port-in-use-preflight.test.mjs`, 6 cases: win32 netstat parsing (including that `TIME_WAIT` rows and other ports are ignored), posix lsof parsing, discovery failure treated as free, `reportPortInUse` naming the port/PID/remedies, and an end-to-end case that opens a real socket and asserts the preflight finds the owning process.

The `#2460` / `#6800` / `#11766` readiness guards still pass (6/6), since `pid.mjs` is shared.

https://claude.ai/code/session_01CJf2dxEpiwZqyZujWk57T2


diegosouzapw:
Hi @dmlanday — this is a solid, well-documented fix for a genuinely nasty bug (a second
`serve` de-registering the healthy instance's pid files). 5/6 of your own tests pass at
your PR head; the 6th ("finds a real listening socket end-to-end") failed only because
this review environment doesn't have `lsof` installed — your own `findListeningPids()`
gracefully falls back to "port free" exactly as designed in that case (same dependency
`stop.mjs::killByPort()` already has). Could you confirm `lsof` is available on the PR's
actual CI runner? One small optional nit: `reportPortInUse()`'s message always prints
`http://` even with TLS configured, because `urlScheme` isn't resolved yet at that point
in `runServe()` — worth a follow-up if you have a moment, not blocking.

diegosouzapw:
Thanks @dmlanday — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-130526-suite.log) on the devbox @ train tip 28fb420c9ad860ba275294ebb4ebbecb9da83318 with the sibling PRs of this batch: static gates green; changed-area node:test 293/294 (0 failing) + vitest 482/482 (fast parity — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §7.

PR #12486, opened 2026-09-02
feat(runtime): add reasoning replay feature flag

jacobsparts:
## Summary

Adds a default-enabled runtime feature flag for Reasoning Replay.

When `REASONING_REPLAY_ENABLED` is disabled, the shared cache service skips reasoning writes and lookups. Provider/model detection and protocol shaping remain unchanged.

## Scope

- Register `REASONING_REPLAY_ENABLED` under Runtime settings
- Gate the shared cache write and lookup functions
- Add one focused disabled-state test
- Add the description key to all 43 locale catalogs
- Add a changelog fragment

The default remains enabled, so existing behavior is unchanged.

## Validation

- Feature-flag tests: **54/54**
- Reasoning Replay tests: **62/62**
- Core typecheck
- Touched-file ESLint
- UI locale-key coverage
- Prettier
- JSON validation for all locale catalogs
- `git diff --check`


diegosouzapw:
Clean, well-scoped feature flag addition — defaulting to `"true"` keeps current behavior
unchanged, the gate placement in `cacheReasoningByKey()`/`lookupReasoning()` is correct, and the
i18n coverage is thorough. CI shows the flag conflicts with the base (likely from recent locale
merges) — could you rebase onto the current `release/v3.8.51` tip? Also flagging for the
maintainer: there are two other open PRs touching reasoning replay (#13031, #12688) that should
probably be checked against this new flag before all three land.

jacobsparts:
Updated this branch to the current `release/v3.8.51` tip (`83fa4328`); PR head is `9299dde4c` and GitHub reports it mergeable. I also added the newly required feature-flag catalog row/count updates; the focused docs/flag/replay suite passes 77/77 locally, along with core typecheck, ESLint, Prettier, and the new-code complexity check.

The remaining red checks reproduce on the base/current upstream state (API-route baseline drift, docs-count drift, frozen `chat-pipeline.test.ts` size, and unrelated unit failures); the Reasoning Replay and flag-count tests pass. Compatibility check: #12688 merges cleanly; #13031's production changes merge cleanly, with only its generated file-size baseline needing mechanical resolution if it lands first.

PR #12487, opened 2026-09-02
feat(settings): persist headroomUrl for the Headroom proxy

### LINKED ISSUE #12306: feat(backend): configurable Headroom server URL for compression
### Problem / Use Case

I am trying to use a Headroom server hosted at a different location or custom URL, but currently, OmniRoute only allows using its own default URL for compressions without any option to modify it. This limits flexibility for users who manage their own Headroom infrastructure.

### Proposed Solution

Add a feature that allows users to manually edit and customize the input URL for compressions. If a user deploys a custom Headroom server elsewhere, they should be able to input their specific server URL instead of being forced to use the default OmniRoute configuration.

### Alternatives Considered

_No response_

### Acceptance Criteria

- [ ] Users can find an input field in the settings/configuration to enter a custom Headroom server URL.
- [ ] If the field is left blank, the system automatically falls back to the default OmniRoute URL.
- [ ] Compressions are successfully routed to the specified custom URL when provided.
- [ ] Basic URL validation is implemented to ensure the entered address is well-formed.

### Area

Dashboard / UI

### Related Provider(s)

_No response_

### Additional Context

_No response_

### Expected Test Plan

_No response_

HouMinXi:
## Summary

Settings currently drops `headroomUrl` on PATCH because the field is missing from `updateSettingsSchema`. Status and start already read that key, so a custom Headroom URL never survives a save.

This PR:

- Adds `headroomUrl` to the settings schema (empty = `HEADROOM_URL` / localhost:8787; whitespace-only trims to empty; http(s) only; max 500).
- Adds a Headroom card on Advanced settings: URL field, save, start, stop.
- Rejects `javascript:` / `file:` / `data:` / malformed `http://` via `isHttpUrl`.

Closes #12306

## Test plan

```
DISABLE_SQLITE_AUTO_BACKUP=true node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit tests/unit/headroom-url-settings-12306.test.ts tests/unit/headroom-proxy-lifecycle.test.ts
```

20/20 pass. Also checked: drop the trim pipe → whitespace-only test fails; drop the `saveAc.current === ac` guard → source assertion fails.

## Notes for reviewers

- `cliproxyapi_url` next to the new field still uses `z.string().url()` with no http(s) refine. That line is from v3.8.4 (`b91ffa7f7`); not touched here.
- `.url()` plus `isHttpUrl` is layered on purpose: Zod's URL parser still accepts `javascript:`; the refine is the scheme gate used by `probeProxyRunning` (`fetch(`${url}/health`)`).
- i18n keys reuse `settingSaveFailed` / `settingSaved`. No new locale strings required beyond the card copy already in this commit.

## Not in this PR

- Changing CLIProxyAPI URL scheme policy
- Headroom CLI install / binary itself


HouMinXi:
Rebased onto `release/v3.8.51` @ `2c4ad3e55`.

Upstream retired `src/i18n/messages/in.json` (#12496). This PR had keys in that file; dropped it and kept the remaining locales. Head is now `13fec0502`.

Issue #12130, opened 2026-08-30
fix(backend): OpenAI API returns null content when model output consists of reasoning tokens (e.g. command-code/meta/muse-spark-1.2-contributor)


Stazyu:
### OmniRoute Version

3.8.51 (next)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Provider(s) Involved

Command-Code

### Model(s) Involved

meta/muse-spark-1.2-contributor

### Description

This issue is related to **#10986**, which reported that the OpenAI-compatible API can return `content: null` when the model output consists entirely of reasoning tokens.

Since then, Command-Code was migrated to the official `/provider/v1` API in **#10265**.

The `/alpha/generate` endpoint is still supported as a fallback through **#11455** for cases where the `/provider/v1` endpoint returns `403/404`. However, the current `/provider/v1` path still appears to have the same reasoning-only output handling issue described in #10986.

### Current Behavior

When using the `COMMAND-CODE` provider with:

`meta/muse-spark-1.2-contributor`

the request completes successfully with `HTTP 200 OK`, but the generated output is classified entirely as reasoning tokens:

* Total output: 64
* Reasoning: 61
* Content: null

As a result, the OpenAI-compatible response contains:

```json
{
  "choices": [
    {
      "message": {
        "role": "assistant",
        "content": null
      }
    }
  ]
}
```

### Relation to Previous Changes

* **#10986** — Original report of `content: null` when all generated output is classified as reasoning.
* **#10265** — Command-Code switched to the official `/provider/v1` API as the primary API.
* **#11455** — Added `/alpha/generate` as a fallback when `/provider/v1` returns `403/404`, primarily to support Go plan accounts.

Importantly, **#11455 addresses API access/fallback behavior, not the handling of reasoning-only responses**.

The issue described here remains reproducible on the current Command-Code provider implementation.

### Expected Behavior

The OpenAI-compatible layer should correctly handle responses where the upstream model produces reasoning output without normal content, rather than returning a successful response with:

`choices[0].message.content = null`

The reasoning output should either be mapped to the appropriate OpenAI-compatible reasoning field or otherwise handled according to the provider's response semantics.

### Steps to Reproduce

1. Configure the `COMMAND-CODE` provider.
2. Use `meta/muse-spark-1.2-contributor`.
3. Send a request through `/v1/chat/completions`.
[...]

Issue #10986, opened 2026-08-21
[BUG] OpenAI API returns null content when model output consists of reasoning tokens (e.g. command-code/meta/muse-spark-1.2-contributor)


Stazyu:
### OmniRoute Version

3.8.50 (next)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Provider(s) Involved

Command-Code

### Model(s) Involved

meta/muse-spark-1.2-contributor

### Description

When making a request to `/v1/chat/completions` using `command-code/meta/muse-spark-1.2-contributor` via the `COMMAND-CODE` provider, the endpoint responds with HTTP 200 OK, but `choices[0].message.content` is set to `null`.

### Steps to Reproduce

1. Configure OmniRoute with provider `COMMAND-CODE`.
2. Send a `POST` request to `/v1/chat/completions` with model `command-code/meta/muse-spark-1.2-contributor`.
3. Pass a simple prompt, e.g.:
   ```json
   {
     "model": "command-code/meta/muse-spark-1.2-contributor",
     "messages": [
       {
         "role": "user",
         "content": "Calculate 79874+93658, and reply with the result only."
       }
     ]
   }

### Expected Behavior

The API should return the generated text response inside choices[0].message.content (or map the reasoning output accordingly) instead of returning null.

### Actual Behavior

The request succeeds with HTTP 200 OK, but choices[0].message.content is set to null because all generated output tokens are categorized as reasoning tokens (Reasoning: 61).

JSON :
```
{
  "choices": [
    {
      "message": {
        "role": "assistant",
        "content": null
      }
    }
  ]
}```

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

### Screenshots

<img width="1035" height="847" alt="Image" src="https://github.com/user-attachments/assets/f1befb6c-ebea-4134-9b34-895ed1273ae7" />

### Additional Context

- **Correlation ID:** `49800973-f5aa-4204-9d11-222b9b98f751`
- **Request ID:** `1787238740599-5eb7b1`
- **Provider:** COMMAND-CODE
- **Requested Model:** `command-code/meta/muse-spark-1.2-contributor`
- **Upstream Model:** `meta/muse-spark-1.2-contributor`
- **Protocol:** OPENAI-CHAT
- **Metrics:** Total Out: 64 | Reasoning: 61 | Content: null

PR #12497, opened 2026-09-02
fix(command-code): floor tiny muse-spark output budgets so hidden reasoning cannot consume the whole budget

Stazyu:
## Summary

- muse-spark (Meta) models routed through **command-code** burn their entire output budget on invisible server-side reasoning before emitting any visible content. With a small
caller-set `max_tokens` (e.g. 64) the upstream answers HTTP 200 with **null content** (`out=64, reasoning=61`) — the client receives an empty assistant turn and OmniRoute flags the
fake success as "Provider returned empty content". This is the reproduction in **#12130**.
- Fix: reuse the already prefix-aware `MUSE_SPARK_PATTERN` (`/(?:^|\/|\b)muse-spark/i`) to detect muse-spark ids — **including provider-prefixed forms**
(`meta/muse-spark-1.2-contributor`, `cmd/meta/muse-…`) that a bare `startsWith("muse-spark")` would miss — and floor a tiny caller budget to `MUSE_SPARK_MIN_OUTPUT_TOKENS = 512` so
the hidden reasoning phase can never consume the whole budget.
- Applied in **both** command-code body builders: `buildOpenAiBody` (the `/provider/v1/chat/completions` path that the reproduction goes through) and `buildCommandCodeCliBody` (the
`/alpha/generate` 403/404 fallback), mirroring the existing `applyMuseSparkMinOutputTokens` behavior already shipped for the opencode-go executor.

## Related Issues

- Related to #12130 (OpenAI API returns null content when model output consists of reasoning tokens, e.g. `command-code/meta/muse-spark-1.2-contributor`) — the exact `out=64,
reasoning=61, content:null` reproduction this fix targets.
- Related to #11214 (muse-spark output budget is floored so it stops returning empty-content responses) — same empty-content-on-small-budget class; previously applied only to
opencode-go, now extended to command-code.
- Related to #10986 (original reasoning-only `content:null` report) — this fix paves the `/provider/v1` path so the budget-floor mitigation prevents the empty-content condition
rather than relying on a downstream fallback.

## Validation

  `node --test tests/unit/command-code-executor.test.ts` → 20/20 pass
  `tsc --noEmit -p tsconfig.typecheck-core.json` → clean
- [x] `npm run lint`

## Tests Added Or Updated

- `tests/unit/command-code-executor.test.ts` — 2 new cases:
  - `meta/muse-spark-1.2-contributor` + `max_tokens: 64` → raised to `512` (the #12130 bug path)
  - large budget (`4096`) left untouched (no over-flooring)

> The existing `tests/unit/opencode-muse-spark-min-output.test.ts` (11/11 pass) confirms the opencode-go executor is **untouched** — only command-code was modified.

## Coverage Notes

- `open-sse/executors/commandCode.ts` is covered by the two new tests above, exercising `applyMuseSparkMinOutputTokens` on both the floor and no-op paths (bare + prefixed id
detection).
- Full `npm run test:coverage` (global 60% gate) was attempted; the c8 `--merge-async` run stalled into an 8GB `coverage/tmp` on this Windows workstation, so the per-change gates
above (scoped tests + typecheck) were used as the focused loop. The global 60% gate runs in CI on the PR (#8329).

## Reviewer Notes

- Detection now uses the prefix-aware `MUSE_SPARK_PATTERN`. For command-code this is a **superset** of any prior bare-prefix check — no model is exempted, only newly covered (e.g.
the `meta/…` id in this reproduction).
- This PR is the **budget-floor prevention** approach: it stops the empty-content condition for small caller budgets. It does **not** add a reasoning→content fallback inside the
`/provider/v1` raw passthrough for the case where a model emits reasoning-only with a *large* budget — that remains the domain of #10986's fallback (already present on the
`/alpha/generate` path). Maintainer's note in #12130 confirms `/provider/v1` currently returns the upstream `Response` unmodified; if we also want a downstream reasoning→content
fallback there, that is a deliberate follow-up.
- No new dependency, feature flag, DB migration, or manual validation required.

### DIFF
--- a/open-sse/executors/commandCode.ts
+++ b/open-sse/executors/commandCode.ts
@@ -1,6 +1,7 @@
 import { randomUUID } from "node:crypto";
 
 import { isVisionModelId } from "@/shared/constants/visionModels";
+import { MUSE_SPARK_PATTERN } from "./base/reasoningEffort.ts";
 import { REGISTRY } from "../config/providerRegistry.ts";
 import {
   BaseExecutor,
@@ -59,6 +60,29 @@ function clampMaxTokens(value: unknown): number | undefined {
   return Math.min(Math.floor(numeric), MAX_COMMAND_CODE_TOKENS);
 }
 
+/**
+ * muse-spark (Meta) models are served through command-code with a hidden
+ * server-side reasoning phase that consumes the entire output budget before any
+ * visible content is emitted. With small caller-set budgets the upstream
+ * answers HTTP 200 with an empty message and `completion_tokens == max_tokens`
+ * (the symptom this fix targets: `out=64, reasoning=61` with null content);
+ * chatCore then flags the fake success as "Provider returned empty content".
+ *
+ * Floor raised budgets only — explicit large budgets are untouched, and no
+ * budget is synthesized when the caller set none. Detection is prefix-aware so
+ * provider-prefixed ids (`meta/muse-spark-1.2-contributor`, `cmd/meta/muse-…`)
+ * are caught, not just bare `muse-spark-*`.
+ */
+const MUSE_SPARK_MIN_OUTPUT_TOKENS = 512;
+
+function applyMuseSparkMinOutputTokens(model: string, body: JsonRecord): void {
+  if (!MUSE_SPARK_PATTERN.test(model)) return;
+  const current = body.max_tokens;
+  if (typeof current !== "number" || !Number.isFinite(current)) return;
+  if (current >= MUSE_SPARK_MIN_OUTPUT_TOKENS) return;
+  body.max_tokens = MUSE_SPARK_MIN_OUTPUT_TOKENS;
+}
+
 const COMMAND_CODE_PASSTHROUGH_FIELDS = [
   "reasoning_effort",
   "reasoning",
@@ -111,6 +135,7 @@ function buildOpenAiBody(model: string, body: unknown, stream: boolean): { body:
   if (maxTokens !== undefined) {
     out.max_tokens = maxTokens;
   }
+  applyMuseSparkMinOutputTokens(resolvedModel, out);
 
   return { body: out };
 }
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Clean fix — reusing the existing prefix-aware `MUSE_SPARK_PATTERN` instead of a naive `startsWith` is the right call (catches `meta/muse-spark-1.2-contributor` and `cmd/meta/...` forms too), and applying it to both `command-code` body builders mirrors the proven #11214 mitigation already shipped for opencode-go. Ran the full test file: 20/20 pass, including your two new cases. This looks merge-ready as is — only optional suggestion is a changelog fragment to mirror #11214's. Thanks for the clear repro and the "don't over-floor" test case.

Stazyu:
> Clean fix — reusing the existing prefix-aware `MUSE_SPARK_PATTERN` instead of a naive `startsWith` is the right call (catches `meta/muse-spark-1.2-contributor` and `cmd/meta/...` forms too), and applying it to both `command-code` body builders mirrors the proven #11214 mitigation already shipped for opencode-go. Ran the full test file: 20/20 pass, including your two new cases. This looks merge-ready as is — only optional suggestion is a changelog fragment to mirror #11214's. Thanks for the clear repro and the "don't over-floor" test case.

Thanks for the review!

I've added the changelog fragment to mirror the existing #11214 mitigation. The implementation and tests remain unchanged.

Appreciate the review and confirmation that it's merge-ready!

diegosouzapw:
Thanks @Stazyu — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

Issue #12430, opened 2026-09-02
feat(backend): video-bridge #12150 P2 — remaining transcript-retention surfaces (raw snapshot, continuation, derived-prompt dispatches, response/semantic-cache)


diegosouzapw:
Surface map for #12150 **P2**, compiled from the P1 whole-branch review (PR #12427). P1 closed the two main-path surfaces (flattened description in the persisted request body via structured content-match redaction + durable Memory suppression). Remaining, all raw/response-class or opt-in×opt-in:

**From the original design (P2 scope):**
1. Raw pre-guardrail client-request snapshot in the detailed-log artifact (`clientRawRequest.body`, captured before guardrails).
2. `previous_response_id` continuation fail-closed when a continuation would rehydrate removed video-derived content (needs a `call_logs` marker column + migration + `resolvePreviousResponseState` refusal).

**Newly surfaced by the P1 review (add to P2):**
3. **Guardrail-ordering fail-open (Important):** video-bridge runs at priority 7; PII masker (10) and credential masker (95) rewrite the SAME chained payload afterward. With `PII_REDACTION_ENABLED`/`CREDENTIAL_REDACTION_ENABLED` on, the flattened description text changes, so the P1 log-redaction content-match (`text === fullText`) misses and the transcript persists (only PII-masked). Fix: re-anchor each redaction entry's `fullText` from the FINAL `preCallGuardrails.payload` (advisory `(container,messageIndex,partIndex)` indices are valid inside the guardrail chain — chain guardrails rewrite text in place, never splice arrays) in `deriveVideoBridgeLog` (src/sse/handlers/chat.ts). NOTE: opt-in×opt-in (PII masking is off by default, Hard Rule #20) — default path is unaffected.
4. Derived-prompt internal dispatches embed the transcript in synthesized STRING-content prompts, which the array-part content-match can't reach: pipeline-strategy stages (`src/domain/pipeline.ts` `original_request`; `pipelineRouter.ts`) and context-handoff (`open-sse/services/contextHandoff.ts` `{HISTORY}`). Memory suppression already holds for these; only their persisted request bodies retain text. Both opt-in. Fix: exact-literal substring replacement at the sink (`indexOf(fullText)`→`redactedText`, still parse-free since fullText is a trusted literal) or suppress request-body persistence for observed derived calls.
[...]

PR #12503, opened 2026-09-02
fix(video): re-anchor transcript log-redaction so PII/credential maskers can't reopen the leak (#12150 P1 follow-up)

diegosouzapw:
## Why

Follow-up to the merged #12427 (#12150 P1). The P1 whole-branch review found an **Important fail-open** that merged before the fix could land: the video-bridge guardrail runs at priority 7, but the PII masker (10) and credential masker (95) rewrite the **same chained payload** afterward, in place. When either masking flag is on, the flattened video description text changes, so the log-redaction content-match (`part.text === fullText`) misses and the transcript persists in the call log (only PII-masked). This hits exactly the privacy-maximal operators who enable both features. (Default path is unaffected — PII masking is off by default, Hard Rule #20.)

## Fix

`reanchorVideoBridgeRedaction` (new, pure, unit-tested in `src/lib/guardrails/videoBridge.ts`) re-reads each redaction entry's `fullText` from the **finished** pre-call guardrail payload at the advisory `(container, messageIndex, partIndex)` — valid because chain guardrails rewrite text in place, never splice the message array. `deriveVideoBridgeLog` (`src/sse/handlers/chat.ts`) now takes the final `body` and applies it, so the sink match sees the post-masker text. `redactedText` is unchanged (rendered from structured cues). New entries are returned; the shared guardrail `meta` is never mutated.

## Validation

TDD: two `reanchorVideoBridgeRedaction` unit tests (re-reads post-masker text; falls back to original when the advisory index no longer resolves) — RED before, GREEN after. Focused suites green: `guardrails/videoBridge` + `video-bridge-log-redaction` + `video-bridge-memory-suppression` (38/38). `typecheck:core`, `prettier --check` (touched files), `check:docs-all` clean. Base is green.

Also clarifies the GUARDRAILS.md retention note and lists the remaining P2 surfaces. Refs #12150, #12430; follows up #12427.

### DIFF
--- a/docs/security/GUARDRAILS.md
+++ b/docs/security/GUARDRAILS.md
@@ -476,23 +476,32 @@ fusion counters. The default Video Bridge path does not invoke speech-to-text
 or download a second media copy; without that explicit track, it remains
 video-only.
 
-**Transcript retention (opt-in feature, #12150 P1).** When a request renders any
-transcript cue (a caller-declared `transcript` or a fused `audioTranscript`), the
-guardrail marks it `videoBridgeObserved` and produces a redacted shadow of the
-video description — an identical rendering in which every cue's free-text body is
-replaced by `[redacted-video-transcript]`, built by substituting the structured
-cue field before the string is assembled (never by parsing the flattened text, so
-no cue content — adversarial or ordinary, including bodies containing `]` such as
+**Transcript retention (#12150 P1).** This applies automatically whenever the
+Video Bridge (itself opt-in) renders a transcript cue — there is no separate
+retention flag. When a request renders any transcript cue (a caller-declared
+`transcript` or a fused `audioTranscript`), the guardrail marks it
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Merging: the two red checks are inherited base-red (tracked in #12518), not from this diff. "No new ESLint warnings" fails on the **CodeQL ratchet** (13 open alerts > baseline 11: js/incomplete-url-substring-sanitization ×4, js/weak-cryptographic-algorithm ×3, js/incomplete-sanitization ×2, js/insufficient-password-hash ×2, js/stack-trace-exposure ×1) — none introduced by this change (a pure re-anchor helper + a param thread; the lint metric itself is eslintWarnings=0/eslintErrors=0). The same check is red on unrelated PRs (#12428, #12421). "Fast Quality Gates" is the aggregate wrapper over the same ratchet. ⚠️ base-red inherited: #12518.

PR #12520, opened 2026-09-02
fix(cli): support dashboard API key ids

Gaulnews:
## Summary
- fall back from registered-key endpoints to ordinary dashboard API-key endpoints for regenerate and reveal
- preserve registered-key behavior when those IDs exist
- add regression coverage for both fallback paths

## Validation
- node --import tsx/esm --test tests/unit/cli-keys-api-key-fallback.test.ts
- node --import tsx/esm --test tests/unit/cli-keys-command.test.ts tests/unit/cli-route-unavailable-fallback-10081.test.ts
- live local validation: regenerated a dashboard API key and authenticated /v1/models with HTTP 200

### DIFF
--- a/bin/cli/commands/keys.mjs
+++ b/bin/cli/commands/keys.mjs
@@ -352,10 +352,24 @@ export async function runKeysRegenerateCommand(id, opts = {}) {
     return 1;
   }
   try {
-    const res = await apiFetch(`/api/v1/registered-keys/${encodeURIComponent(id)}/regenerate`, {
+    const encodedId = encodeURIComponent(id);
+    let res = await apiFetch(`/api/v1/registered-keys/${encodedId}/regenerate`, {
       method: "POST",
       retry: false,
+      acceptNotOk: true,
     });
+    // `keys` predates the split between registered keys and the dashboard's
+    // ordinary API keys. IDs shown by `keys list`/the dashboard belong to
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks @Gaulnews — this closes a real ID-namespace drift between `keys list` (ordinary
dashboard API keys, `/api/keys`) and the registered-keys store the CLI's
regenerate/reveal only checked before. Reusing the existing
`isRouteUnavailableStatus`/`acceptNotOk` pattern already used throughout this file is
the right call, and the test drives real fetch mocks with URL-based assertions rather
than hand-waving. One thing missing: could you add a changelog fragment under
`changelog.d/fixes/`? That's the only blocker I see before merge.

diegosouzapw:
Thanks @Gaulnews — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #12524, opened 2026-09-02
feat(providers): refresh Z.ai Web models and browser transport

backryun:
## Summary

- replace the stale Z.ai Web catalog with the current `GLM-5.3-Flash`, `GLM-5.3`, and `GLM-5.2` models
- map the public Flash model id to Z.ai's current `x-preview-l` wire id and expose each model's supported mandatory Thinking effort levels
- select models and Deep Think effort through the current browser UI using exact labels
- keep the authenticated `/api/v2/chat/completions` flow and run Z.ai in pooled off-screen headed Chromium because true headless Chromium is rejected by the current CAPTCHA with `F001`
- include sanitized POST response statuses in browser transport failures for actionable diagnostics

## Validation

- `node --import tsx/esm --test tests/unit/browserBackedChat-matcher.test.ts tests/unit/browserPool-proxy.test.ts tests/unit/executor-zai-web.test.ts tests/unit/zai-web-chat-endpoint-8014-probe.test.ts`
- `npm run typecheck:core`
- changed-file ESLint with the repository suppressions file
- `npm run check:provider-consistency`
- `npm run check:provider-assets`
- live local `/v1/chat/completions` calls returned HTTP 200 for all three models:
  - `zw/glm-5.3-flash` with Low effort
  - `zw/glm-5.3` with Low effort
  - `zw/glm-5.2` with High effort

### DIFF
--- a/open-sse/config/providers/registry/zai-web/index.ts
+++ b/open-sse/config/providers/registry/zai-web/index.ts
@@ -15,29 +15,26 @@ export const zai_webProvider: RegistryEntry = {
   // not accept caller-supplied OpenAI `tools`, which remains disabled here.
   models: [
     {
+      id: "glm-5.3-flash",
+      name: "GLM-5.3-Flash",
+      toolCalling: false,
+      supportsReasoning: true,
+      supportedThinkingEfforts: ["low", "high", "max"],
+      supportsVision: true,
+    },
+    {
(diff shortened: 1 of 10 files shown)


backryun:
image
Tested, and confirmed perfectly works!

PR #12528, opened 2026-09-02
feat(video): redact raw client-snapshot transcript fields in the detailed log (#12150 P2)

diegosouzapw:
Refs #12150, #12430 (P2 surface 1)

⚠️ base-red inherited: #12518

## What

`clientRawRequest.body` is a bounded clone of the client's ORIGINAL request,
captured **before** the guardrail chain runs, and persisted verbatim by
`reqLogger.logClientRawRequest` (`open-sse/handlers/chatCore.ts`). Because it
predates the video-bridge guardrail's own description redaction (#12150 P1),
it still carried the client's raw `transcript` / `audioTranscript` cue text on
any video part — the dominant transcript-retention leak identified in the
#12150 design doc (surface 2 there, P2's "surface 1").

This PR redacts that snapshot, and only that snapshot:

- New file `src/lib/guardrails/videoBridgeSnapshotRedaction.ts` exports
  `redactVideoTranscriptFieldsForLog(body: unknown): unknown` — a pure,
  dependency-light helper that walks `body.messages[].content[]` and
  `body.input[].content[]` and replaces `transcript` / `audioTranscript`
  field VALUES (on the part itself, or its `video_url`/`source` sub-object —
  mirrors how `extractVideoParts` in `videoBridgeHelpers.ts` reads them) with
  the sentinel `"[redacted-video-transcript]"`. Keys are never deleted.
  Returns a new `structuredClone`, never mutates the input.
  - Deliberately **not** part of `videoBridgeHelpers.ts` (frame-extraction
    broker, audio/video fusion, contact-sheet composition, `sharp`) so the
    hot chat-request path doesn't statically import that dependency chain
    just to reach this one redaction. The placeholder is kept as a local
    literal (not imported) for the same reason; a drift test asserts it
    matches the canonical `VIDEO_TRANSCRIPT_REDACTION_PLACEHOLDER` export.
- One guarded call-site change in `open-sse/handlers/chatCore.ts`
  (`logClientRawRequest`): when `videoBridgeObserved` (already threaded by
  P1), pass `redactVideoTranscriptFieldsForLog(clientRawRequest.body)`
  instead of the raw body. Non-observed path is byte-identical (same object
  reference, no clone) — verified directly in a test.

Out of scope (P2b/P2c): the continuation store and derived-prompt dispatches.

## Testing

TDD, `env -u OMNIROUTE_API_KEY` throughout:

- RED: 8 new tests in `tests/unit/guardrails/videoBridgeSnapshotRedaction.test.ts`
  run against a naive passthrough stub — 3 pass / 5 fail (captured).
- GREEN: same 8 tests against the real implementation — 8/8 pass. Covers both
  `messages` and `input` containers, transcript directly on the part vs.
  nested under `video_url`/`source`, ref/non-video-part preservation,
  no-mutation, a non-video body unchanged, a video part with no transcript
  field unchanged, and no drift from the P1 placeholder constant.
- Extended `tests/unit/video-bridge-log-redaction.test.ts` (the existing P1
  surface-1 file) with a new test exercising the exact
  `videoBridgeObserved ? redact(...) : body` call-site pattern against a fake
  `logClientRawRequest` — redacted+cloned when observed, byte-identical
  reference when not. Sanity-mutated the guard condition to confirm the test
  fails when the wiring is wrong (4 pass / 1 fail), then reverted — 5/5 green.
  A full `handleChatCore` integration test was not attempted: importing the
  real `chatCore.ts` is feasible (44 existing tests do) but driving it this
  deep requires the ~900-line DB/upstream-fetch mocking harness seen in
  `chatcore-sanitization.test.ts`, for no additional proof beyond what the
  above two suites already cover — per the task brief's own escape hatch.
- Full `tests/unit/guardrails/**` directory: 528/528 pass (no regressions).
- `tests/unit/chatcore-imports-cleanly.test.ts` +
  `tests/unit/video-bridge-memory-suppression.test.ts`: 12/12 pass (chatCore.ts
  still imports cleanly with the new static import; P1's Memory-suppression
  surface is unaffected).
- `tests/unit/chatcore-sanitization.test.ts` (full `handleChatCore`
  integration, detailed-logging path): 18/18 pass.
- `npm run test:vitest`: 51 files / 465 tests pass.
- `npx prettier --check` on all touched files: clean.
- eslint was not run — broken repo-wide (`@eslint/compat`), per the task's
  stated gate substitution (prettier + typecheck:core).

Full `npm run test:unit` (whole-repo suite) was not run given its multi-hour
scope and the change's narrow, well-isolated surface confirmed green above;
flagging this explicitly for the reviewer.

### DIFF
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -360,7 +360,7 @@ import { deleteSessionAccountAffinity } from "@/lib/db/sessionAccountAffinity";
 import { getCacheControlSettings } from "@/lib/cacheControlSettings";
 import { guardrailRegistry } from "@/lib/guardrails";
 import type { VideoBridgeLogRedactionEntry } from "@/lib/guardrails/videoBridge";
-import { redactVideoTranscriptFieldsForLog } from "@/lib/guardrails/videoBridgeSnapshotRedaction";
+import { logClientRawRequestRedacted } from "@/lib/guardrails/videoBridgeSnapshotRedaction";
 import {
   shouldPreserveCacheControl,
   resolveConnectionCacheOverride,
@@ -1212,22 +1212,9 @@ export async function handleChatCore({
   });
   const pendingScope = { id: pendingRequestId, model, provider, connectionId: pendingConnId };
   const providerRequestCapture = createPreparedRequestLogger(reqLogger, pendingScope);
-  // 0. Log client raw request (before format conversion)
-  if (clientRawRequest) {
-    reqLogger.logClientRawRequest(
-      clientRawRequest.endpoint,
-      // #12150 P2 surface 1 (the dominant transcript-retention leak): this snapshot is
-      // captured BEFORE the guardrail chain runs, so it still carries the client's raw
-      // video transcript cue text untouched by the video-bridge guardrail's own
-      // description redaction. Redact it in this LOGGED copy only when the guardrail
-      // observed a video part on this request — clientRawRequest.body itself is never
-      // mutated and keeps flowing to every other consumer unchanged.
-      videoBridgeObserved
-        ? redactVideoTranscriptFieldsForLog(clientRawRequest.body)
-        : clientRawRequest.body,
-      clientRawRequest.headers
-    );
-  }
+  // 0. Log client raw request (before format conversion) — redacts video transcript
+  // cues in the logged copy only; see videoBridgeSnapshotRedaction.ts.
+  logClientRawRequestRedacted(reqLogger, clientRawRequest, videoBridgeObserved);
   const reasoningRouteDecision =
     body && typeof body === "object"
       ? (body as Record<string, unknown>)._omnirouteReasoningRouteTrace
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Fast Quality Gates \`check:file-size\` caught the P2a wiring growing \`chatCore.ts\` past its frozen ceiling (5985 > 5976 congelado). Fixed in 21debc0819: extracted the guarded \`logClientRawRequest\` call into \`logClientRawRequestRedacted\` (new export in \`videoBridgeSnapshotRedaction.ts\`, which already owns the redaction), collapsing the inline if-block at the call site to a single call — net **-4 lines** vs the pre-P2a base. Behavior unchanged (verified by tests): non-observed still logs the exact same \`clientRawRequest.body\` reference; observed still logs the redacted clone.

Verified:
\`\`\`
$ npm run check:file-size -- --base-ref 86e83d4138441c3c842c2a3bc5c20d1b06db41d1
[file-size] OK — 135 arquivos congelados, cap 1200 para novos (4527 arquivos verificados)
[test-file-size] OK — 40 test files congelados, testCap 1200 para novos (5385 test files verificados)
\`\`\`
All prior tests re-run green (guardrails dir 528/528, chatcore-imports-cleanly + video-bridge-memory-suppression + chatcore-sanitization 30/30), plus \`typecheck:core\` and \`prettier --check\` clean.

diegosouzapw:
Merging: file-size violation fixed (chatCore.ts 5972 < 5976, baseline untouched; verified locally + re-review). 0 checks failing; the one pending job ("No new ESLint warnings") carries the inherited CodeQL ratchet (base-red #12518, 13>11) which this diff cannot affect — it is a pure structured-redaction helper (structuredClone + field overwrite), no url/crypto/hash/stack-trace code, eslint metric 0/0. ⚠️ base-red inherited: #12518.

Issue #10421, opened 2026-08-14
fix(db): pre-migration backups are never pruned — db_backups grew to 204 GB / 49k files


diegosouzapw:
## Summary

`createPreMigrationBackup()` writes a `VACUUM INTO` snapshot on every migration run and **never prunes**. On a long-lived dev box this filled the disk: **48.999 backup files consuming 204 GB**, while the live database was only 5,3 MB.

```
$ ls -1 ~/.omniroute/db_backups | wc -l
48999
$ du -sh ~/.omniroute/db_backups
204G
$ ls -lh ~/.omniroute/storage.sqlite
-rw-r--r-- 1 ... 5,3M ... storage.sqlite
```

Breakdown by reason:

| reason | count |
| --- | --- |
| `pre-migration` | 48.966 |
| `health-check-repair` | 28 |
| `pre-arena-elo-sync` | 3 |

## Root cause

`src/lib/db/migrationRunner.ts:812` — `createPreMigrationBackup()` creates the snapshot and returns, with no retention step:

```ts
const backupPath = path.join(backupDir, `db_${timestamp}_pre-migration.sqlite`);
db.exec(`VACUUM INTO '${escapedBackupPath}'`);
console.log(`[Migration] Pre-migration backup created: ${backupPath}`);
return backupPath;
```

The pruning logic already exists and is complete — `cleanupDbBackups()` at `src/lib/db/backup.ts:164`, honoring the operator's `maxFiles` / `retentionDays` settings. It is simply never reached from the migration path. Its only callers are `src/lib/db/backup.ts:396` and the `/api/db-backups` route, neither of which runs during a migration.

Because every process start that opens the DB with pending migrations produces one ~4,7 MB file, the directory grows without bound. On this machine it accumulated ~49k files between 2026-06-13 and 2026-08-14.

## Impact

- Unbounded disk growth on any long-running instance; here it consumed 204 GB and took the host to 97% full.
- The operator's configured retention (`maxFiles` / `retentionDays`) is silently ignored for the backups that are by far the most numerous.

## Expected

After a pre-migration backup is written, retention should be applied — the same policy the operator already configured and that the manual/API path honors.

## Environment

- OmniRoute v3.8.50, `release/v3.8.50`
- `DATA_DIR=~/.omniroute`, Linux, Node 22

PR #10409, opened 2026-08-14
feat(teams): add billing cost centers and soft shared budgets [defer to 3.8.51]

xiaoyaner0201:
## Summary

Adds Team **billing cost centers** and **soft shared budgets**, targeted at `release/v3.8.51`.

This head is a merge of the public candidate lineage (`1642794b7`) with LIVE `release/v3.8.51` (`e0ea3f92e`), so the branch is reconciled against the current release line rather than an older snapshot.

| | |
|---|---|
| Head | `6e842d89241251f715f10cd02d8950578c44765b` |
| Tree | `318133236c4c53568aa3e16c050cc9ad6943770f` |
| parent^1 (candidate lineage) | `1642794b7f3169cd074929cb7ed862227aa06167` |
| parent^2 (LIVE `release/v3.8.51`) | `e0ea3f92e1403351686cb783764ba3da6bbddd75` |

### Reconciliation with LIVE

Diff against the previously reviewed candidate tree is exactly **five files**:

| File | Relation to LIVE `e0ea3f92e` |
|---|---|
| `package.json` | blob-identical |
| `package-lock.json` | blob-identical |
| `electron/package.json` | blob-identical |
| `electron/package-lock.json` | blob-identical |
| `config/quality/eslint-suppressions.json` | intentionally **not** identical — both sides edited it |

The suppressions file is the only semantic merge. Its resulting keyset is exactly the intersection of the two sides (1368 → 1366): both sides' deletions are honoured, with no resurrected entries, no silent additions, and valid JSON.

### Migration placement

- The migration lands at slot **`164_team_cost_centers.sql`** — exactly one file, not renumbered.
- Canonical `163_radar_feed_cache_generated_at.sql` is **byte-identical to LIVE**; so are `src/lib/db/migrationRunner.ts` and `scripts/check/check-migration-numbering.mjs` (the numbering gate is not weakened by this branch).
- `RENAMED_MIGRATION_COMPATIBILITY` carries **153 / 154 / 155 / 161 / 163 → 164**, every mapping **name-locked** to `team_cost_centers`.

Because the applied-migrations table keys on `version TEXT PRIMARY KEY` and every lookup/update/delete matches on **version AND name**, a mapping cannot rehome a row belonging to a different migration name. Installs that already applied this migration at an earlier slot are rehomed to 164; the canonical 163 row is untouched.

### Gates re-run on this exact tree

| Gate | Result |
|---|---|
[...]

PR #12548, opened 2026-09-03
fix(db): auto-clean conversation_turn_nodes and orphaned conversations

### LINKED ISSUE #12453: fix(backend): conversation_turn_nodes / agentic_conversations have no retention — unbounded storage.sqlite growth
## Problem

`conversation_turn_nodes` and `agentic_conversations` (added in #10263, migrations 155/156) have **no retention path at all**: they are not in `src/lib/db/cleanup.ts`, have no dashboard retention setting, no env knob, no cascade FK, and nothing else ever deletes from them. Verified against `v3.8.50` and current `main` (`cleanup.ts` contains zero references to either table).

#10263 resolved the storage-architecture concern for *content* — turn text is resolved on demand from `call_logs`, which is "already covered by the existing retention/cleanup lifecycle". But the **identity rows themselves are not covered**, so they outlive the content they point to: once `call_logs` retention (default 3 days, ours too) purges the artifacts, the surviving nodes' `last_correlation_id` points at nothing, and the rows are dead weight that can never render a transcript again.

## Production numbers (4 days after upgrading to this feature)

Single-instance deployment, one busy coding-agent workload (Claude Code with MCP tools), `CALL_LOG_RETENTION_DAYS=3`:

| Metric | Value |
|---|---|
| `conversation_turn_nodes` rows | **1,155,364** (4 days after migration 156 ran) |
| `agentic_conversations` rows | 44,004 |
| Table + 4 indexes on-disk (dbstat) | **~775 MB** of a 1.1 GB `storage.sqlite` |
| Growth rate | **~190 MB/day and accelerating** (618k of the 1.15M rows are from the last 24h) |
| Rows already older than call-log retention | 63k and growing — permanently unrenderable, never deleted |

At this rate the table alone reaches multiple GB within weeks. This is the same failure class as #6848 / #4354 / #10421 (unbounded telemetry growth in `storage.sqlite`), which the cleanup job was built to prevent — the two newest tables just never joined it.

## Proposed fix

Add both tables to the existing cleanup lifecycle in `src/lib/db/cleanup.ts`, following the established per-table pattern:

1. `DELETE FROM conversation_turn_nodes WHERE last_seen_at < cutoff` — cutoff derived from the **call-log retention setting** (reuse `retention.callLogs` rather than introducing a new knob): a node older than call-log retention cannot resolve its display content anyway, so keeping identity rows longer than content is pure cost. If a separate knob is preferred, default it to the callLogs value.
2. Sweep orphaned `agentic_conversations` (no remaining turn nodes) in the same pass.
3. Register both in the startup/scheduled cleanup runner (#9624) and in the "Clear all" paths where appropriate.
4. Document the lifecycle in `docs/reference/ENVIRONMENT.md` next to the other retention rows.

Semantics note: deleting old nodes only affects *reconnect anchors* — a conversation resumed after the retention window mints a new conversation id instead of extending the old chain. That is already the documented behavior for any anchor miss (see `resolveConversationId`), so retention introduces no new failure mode, only makes an existing one time-bounded.

## Workaround we run today

Manual `DELETE FROM conversation_turn_nodes WHERE last_seen_at < datetime('now','-3 day')` plus an orphan sweep of `agentic_conversations`, scheduled outside the app. Works, but every self-hosted install that upgraded past migration 156 is silently accumulating this table until they discover it the hard way (ours was ~70% of the whole database after 4 days).

## Environment

- OmniRoute 3.8.50 (standalone Docker, `node dev/run-standalone.mjs`)
- SQLite storage on a 3.9 GB volume, WAL mode
- Traffic: parallel agentic coding sessions (large tool-heavy bodies), ~75k `call_logs` rows per 3 days

pacocartones:
## Summary

- `conversation_turn_nodes` and `agentic_conversations` (migrations 155/156, #10263) had no retention path at all: `src/lib/db/cleanup.ts` never referenced either table, there is no cascade FK, and nothing else deletes from them. The reporter's numbers in #12453 (1.15M node rows, ~775 MB of a 1.1 GB `storage.sqlite` after four days, ~190 MB/day) are the same failure class as #6848 / #4354 / #10421.
- The nodes are identity-only: the transcript view resolves each turn's display content from the `call_logs` row that `last_correlation_id` points at (`156_conversation_turn_nodes.sql` header). Once `cleanupCallLogs` purges that row the node can never render again, so both tables now follow the existing `retention.callLogs` window (`src/types/databaseSettings.ts`) instead of getting a knob of their own, as the issue proposes. No new dashboard setting, no new environment variable, and `docs/reference/ENVIRONMENT.md` is untouched because there is nothing new to document there.
- Two new functions registered at the end of `runAutoCleanup`: `cleanupConversationTurnNodes()` deletes nodes with `last_seen_at` before the cutoff through the shared `deleteFromTableBefore` helper (`src/lib/db/cleanup/usagePurge.ts`, so a database that predates migration 156 is a no-op rather than an error); `cleanupAgenticConversations()` then sweeps roots past the same cutoff that have no remaining node, with `WHERE last_seen_at < ? AND NOT EXISTS (SELECT 1 FROM conversation_turn_nodes n WHERE n.conversation_id = agentic_conversations.id)`. The `last_seen_at` guard uses the index from migration 155 and keeps a root that `createConversation` wrote moments ago, before the same request inserted its nodes.
- Semantics are the ones the issue describes: deleting an old node only affects reconnect anchors, and a conversation resumed after the window mints a new id, which is already the documented anchor-miss behavior of `resolveConversationId`. Deliberately out of scope: the "Clear all" / `resetUsageHistory` paths (a separate, user-triggered contract), and a `last_seen_at` index on `conversation_turn_nodes`. The node DELETE is a table scan today because migration 156 defined no index on that column; it runs once per cleanup cycle, and I can send the index as a follow-up migration if the maintainer wants it.

## Related Issues

- Closes #12453
- Related to #10263
- Related to #6848

## Validation

- [x] Change type: DB (auto-cleanup lifecycle)
- [x] Focused tests and category gates from the golden path: `tests/unit/db-cleanup-conversation-nodes-12453.test.ts` + `telemetry-auto-cleanup-6848` + `db-cleanup` + `db-cleanup-xp-audit-log` 23/23, `npm run check:db-rules` OK, `node scripts/check/check-complexity-ratchets.mjs --base-ref origin/release/v3.8.51` OK (0 violations), `npm run check:changelog-integrity` OK, `npm run typecheck:core` 0 errors
- [ ] `npm run lint`
- [x] Reconciled with the current active release base `release/v3.8.51`; focused checks rerun afterward

## Tests Added Or Updated

- `tests/unit/db-cleanup-conversation-nodes-12453.test.ts` (new, 4 cases, real SQLite adapter in a `mkdtemp` `DATA_DIR`, same pattern as `telemetry-auto-cleanup-6848.test.ts`): nodes older than `retention.callLogs` deleted and recent ones kept; stale orphan root swept while a stale-but-anchored root and a fresh root without nodes both stay; a chain whose nodes expire is swept in the same pass, and never before its nodes are gone; `runAutoCleanup` reports `conversationTurnNodes` and `agenticConversations` with the expected counts. Red on the base (4/4 fail: `cleanupConversationTurnNodes is not a function`, `cleanupAgenticConversations is not a function`, `conversationTurnNodes missing from results`), green with the change (4/4).

## Coverage Notes

- `src/lib/db/cleanup.ts`: `cleanupConversationTurnNodes`, `cleanupAgenticConversations` and their registration in `runAutoCleanup` are exercised end-to-end by the new test file; the existing `tests/unit/db-cleanup.test.ts` and `telemetry-auto-cleanup-6848.test.ts` stay green.
- No touched file lost coverage.

## Reviewer Notes

- Purely additive in `cleanup.ts`: two new functions inserted before `runAutoCleanup` and two new entries at the end of its `results` map. #10409 also edits this file, but in different hunks (`cleanupMcpAudit`, `cleanupA2aEvents`, `resetUsageHistory`), so the two should merge cleanly in either order. #12448 edits `src/lib/db/agenticConversations.ts`, which this PR does not touch.
- Behavior change for operators: with the default `retention.callLogs` (30 days), conversations idle for longer than that lose their reconnect anchor and start a new conversation id on resume. Operators who set a short call-log retention (the reporter runs 3 days) get the same window for anchors, which is the trade-off the issue asks for.
- The node DELETE scans `conversation_turn_nodes` (no `last_seen_at` index in migration 156). On the reporter's 1.15M-row table that is one pass per cleanup cycle; if that is a concern I will add the index as a follow-up migration rather than widen this PR.

### DIFF
--- a/changelog.d/fixes/12548-db-cleanup-orphaned-conversation-nodes.md
+++ b/changelog.d/fixes/12548-db-cleanup-orphaned-conversation-nodes.md
@@ -1,0 +1,1 @@
+- **fix(db):** Add `conversation_turn_nodes` and orphaned `agentic_conversations` to the auto-cleanup cycle under the existing `retention.callLogs` window, so identity nodes whose call-log content has already been purged no longer accumulate without bound in `storage.sqlite` (#12548 — thanks @pacocartones)
--- a/changelog.d/fixes/PENDING-db-cleanup-orphaned-conversation-nodes.md
+++ b/changelog.d/fixes/PENDING-db-cleanup-orphaned-conversation-nodes.md
@@ -1,0 +1,1 @@
+- **fix(db):** Add `conversation_turn_nodes` and orphaned `agentic_conversations` to the auto-cleanup cycle under the existing `retention.callLogs` window, so identity nodes whose call-log content has already been purged no longer accumulate without bound in `storage.sqlite` (#PENDING — thanks @pacocartones)
--- a/changelog.d/fixes/PENDING-db-cleanup-orphaned-conversation-nodes.md
+++ b/changelog.d/fixes/PENDING-db-cleanup-orphaned-conversation-nodes.md
@@ -1,1 +1,0 @@
-- **fix(db):** Add `conversation_turn_nodes` and orphaned `agentic_conversations` to the auto-cleanup cycle under the existing `retention.callLogs` window, so identity nodes whose call-log content has already been purged no longer accumulate without bound in `storage.sqlite` (#PENDING — thanks @pacocartones)
--- a/src/app/api/settings/purge-usage-history/route.ts
+++ b/src/app/api/settings/purge-usage-history/route.ts
@@ -54,6 +54,8 @@ export async function POST(request: Request) {
         deletedRoutingDecisions: result.deletedRoutingDecisions,
         deletedQuotaConsumption: result.deletedQuotaConsumption,
         deletedTokenLedger: result.deletedTokenLedger,
+        deletedConversationTurnNodes: result.deletedConversationTurnNodes,
+        deletedAgenticConversations: result.deletedAgenticConversations,
         errors: result.errors,
       },
       { status: result.errors > 0 ? 500 : 200 }
--- a/src/lib/db/cleanup.ts
+++ b/src/lib/db/cleanup.ts
@@ -431,6 +431,91 @@ export async function cleanupCcrBlocks(): Promise<CleanupResult> {
 }
 
 /**
+ * Clean up conversation_turn_nodes older than the call-log retention window (#12453).
+ *
+ * The nodes are identity-only: the transcript view resolves each turn's display
+ * content from the call_logs row `last_correlation_id` points at. Once
+ * cleanupCallLogs purges that row the node can never render again, so the two
+ * tables share `retention.callLogs` instead of a knob of their own. Deleting an
+ * old node only affects reconnect anchors: a conversation resumed after the
+ * window mints a new id, which is already the documented anchor-miss behavior
+ * of resolveConversationId. `last_seen_at` has no index (migration 156), so
+ * the DELETE is a table scan; it runs once per cleanup cycle.
+ */
+export async function cleanupConversationTurnNodes(): Promise<CleanupResult> {
+  const retention = getRetentionSettings();
+
+  const retentionDays = retention.callLogs;
+  const cutoffDate = new Date();
+  cutoffDate.setDate(cutoffDate.getDate() - retentionDays);
+  const cutoffISO = cutoffDate.toISOString();
+
+  const result: CleanupResult = { deleted: 0, errors: 0 };
+
+  try {
+    result.deleted = deleteFromTableBefore(
+      { table: "conversation_turn_nodes", column: "last_seen_at", cutoff: "iso" },
+      cutoffISO
+    );
+
+    console.log(
+      `[Cleanup] Deleted ${result.deleted} conversation_turn_nodes older than ${retentionDays} days`
+    );
+  } catch (err: unknown) {
+    console.error("[Cleanup] Error cleaning conversation_turn_nodes:", err);
+    result.errors++;
+  }
+
+  return result;
+}
+
+/**
+ * Sweep agentic_conversations left without any conversation_turn_nodes (#12453).
+ *
+ * Runs after cleanupConversationTurnNodes so a root whose whole chain just
+ * expired goes in the same pass. The `last_seen_at` guard (indexed by
+ * migration 155) keeps a root that createConversation wrote moments ago, before
+ * the same request inserted its nodes, and bounds the NOT EXISTS probe to rows
+ * that are already past the retention window.
+ */
+export async function cleanupAgenticConversations(): Promise<CleanupResult> {
+  const db = getDbInstance();
+  const retention = getRetentionSettings();
+
+  const retentionDays = retention.callLogs;
+  const cutoffDate = new Date();
+  cutoffDate.setDate(cutoffDate.getDate() - retentionDays);
+  const cutoffISO = cutoffDate.toISOString();
+
+  const result: CleanupResult = { deleted: 0, errors: 0 };
+
+  try {
+    if (!tableExists("agentic_conversations") || !tableExists("conversation_turn_nodes")) {
+      return result;
+    }
+
+    const stmt = db.prepare(
+      `DELETE FROM agentic_conversations
+       WHERE last_seen_at < ?
+         AND NOT EXISTS (
+           SELECT 1 FROM conversation_turn_nodes n
+           WHERE n.conversation_id = agentic_conversations.id
+         )`
+    );
+    result.deleted = stmt.run(cutoffISO).changes;
+
+    console.log(
+      `[Cleanup] Deleted ${result.deleted} orphaned agentic_conversations older than ${retentionDays} days`
+    );
+  } catch (err: unknown) {
+    console.error("[Cleanup] Error cleaning agentic_conversations:", err);
+    result.errors++;
+  }
+
+  return result;
+}
+
+/**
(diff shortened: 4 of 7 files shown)


insoln (commented):
## Review: schema, semantics, pipeline, SQL, tests — verified locally

Ran `tests/unit/db-cleanup-conversation-nodes-12453.test.ts` (4/4 pass, ~3 s), `npm run lint` (clean), and reproduced the reported scale from #12453 in a scratch DB with migration 156's exact schema (1,155,364 rows, 758.9 MB). The core fix is correct: both migration-comment claims check out (155 does index `agentic_conversations.last_seen_at`; 156 has no `last_seen_at` index), every writer of both tables' `last_seen_at` uses `toISOString()` so the string `<` comparisons are sound, the `NOT EXISTS` probe is a covering-index seek (`idx_turn_nodes_conversation`, measured 0.9 s for 36k conversations), and all readers (reconnect walk, tree route, conversation list) already degrade gracefully to the documented anchor-miss path. Scheduler wiring is fine: `startCleanupScheduler()` runs `runAutoCleanup()` at startup+6h with `autoCleanupEnabled` defaulting to true, and the `VACUUM` in the scheduler does reclaim the disk (measured 758.9 → 71.8 MB).

### Blocking

**1. Unbatched 1M-row DELETE freezes the whole server for ~36 s (measured).**
`cleanupConversationTurnNodes` → `deleteFromTableBefore` (`src/lib/db/cleanup/usagePurge.ts:50-52`) is one synchronous `DELETE ... WHERE last_seen_at < ?` with `SCAN conversation_turn_nodes` as the query plan. On a replica of the reporter's table (1,155,364 rows / 758.9 MB, WAL, pragmas as in `src/lib/db/core.ts`): **1,039,828 rows deleted in 36.0 s with the Node event loop stalled the entire 36.0 s** (a 10 ms timer fired zero times during the delete). This runs 30 s after startup, so a first deploy onto the reporter's already-bloated DB is ~36 s of total outage — every API request, SSE chunk, and `/health` probe hangs. `core.ts`'s own comment (~:1276) documents that better-sqlite3 writes park the event loop. Suggested fix: chunk it, e.g. loop `DELETE FROM conversation_turn_nodes WHERE rowid IN (SELECT rowid FROM conversation_turn_nodes WHERE last_seen_at < ? LIMIT 10000)` with `setImmediate` yields between chunks, summing `changes`. (A `last_seen_at` index doesn't help — it's a delete-most-of-table scan either way; batching is the right lever. Note the follow-up VACUUM is another synchronous stall, but that's pre-existing scheduler behavior.)

### Should-fix

**2. "Clear all" doesn't clear these tables (issue item 3).** `RESET_TARGETS` (`src/lib/db/cleanup.ts:695-737`) lists 12 tables; `conversation_turn_nodes` and `agentic_conversations` are absent. A user with the reported bloat who clicks *Reset usage data — All* wipes `call_logs` (the nodes' display-content anchor) but keeps the 775 MB of nodes — doubly dead data that then waits up to 30 days for the retention sweep. The PR body declares this out of scope; I'd argue the "Clear all" path is exactly where it belongs since that's the documented escape hatch for the reported state. Happy to be overruled if a follow-up is planned.

**3. `CALL_LOG_RETENTION_DAYS` does not control this window — the reporter's setting has no effect here.** The new functions read `getUserDatabaseSettings().retention.callLogs` (dashboard DB setting, default **30 days**; `src/types/databaseSettings.ts:120`). The env var is consumed only by the compliance path (`src/lib/compliance/index.ts:496`) and never reaches `getUserDatabaseSettings` (zero `process.env` reads in `databaseSettings.ts`). #12453's reporter runs `CALL_LOG_RETENTION_DAYS=3` and will still get a 30-day node window (~5–6 GB steady-state at their ~190 MB/day) unless they also change the dashboard setting. Pre-existing coupling for `call_logs` itself, but this PR makes it the only knob for two more tables without documenting it — worth at least a Reviewer Notes line or a docs touch in `docs/reference/ENVIRONMENT.md` next to the `CALL_LOG_RETENTION_DAYS` row (issue item 4).

### Nits

- `cleanupAgenticConversations`' doc comment credits the `NOT EXISTS` guard with protecting a freshly-created root, but a root written moments ago has `last_seen_at = now` and can't match `last_seen_at < cutoff` anyway — the `last_seen_at` predicate alone excludes it. The guard is fine; the stated reason isn't quite the real one.
- The `tableExists` early-return in `cleanupAgenticConversations` (`cleanup.ts:489-491`) is unique new logic with no test — a cheap `DROP TABLE` subtest would pin it.
- Test nit: `OLD = -(R+10)`, `RECENT = -1` silently assumes retention ≥ 2 days.

Everything else — typing of `runAutoCleanup` results (untyped record, totals include the new keys), scheduler defaults, `dbstat`-based size display picking the tables up automatically, test isolation (`mkdtempSync` override beats the harness default) — verified fine.

**Verdict:** the retention semantics are correct and well-tested; the event-loop freeze in finding 1 is the one thing I'd want addressed before this merges onto a production DB of the reported size.

PR #12562, opened 2026-09-03
fix(sse): prefer cgroup PSI for chat admission

SCys:
## Summary

Chat admission's resource-pressure sampler read host-wide `/proc/pressure/memory`. In Docker that file is the **machine**, not the container: a swapping host 503s idle OmniRoute with `resource_pressure` even when the cgroup is well under its limit (V8/cgroup ratios stay `normal`).

Prefer this unit's cgroup v2 `memory.pressure` when it parses. Keep `/proc/pressure/memory` only as fallback (bare metal, cgroup v1, missing file).

## Test plan

- [x] `node --import tsx/esm --test tests/unit/resource-pressure-sampler.test.ts`
  - prefers cgroup PSI when both files exist and disagree
  - falls back to `/proc/pressure/memory` when the cgroup pressure file is absent
  - existing sampler cases still pass
- [ ] Docker host under memory PSI, container idle: `/v1/chat/completions` is not 503 `resource_pressure`

### DIFF
--- a/changelog.d/fixes/resource-pressure-cgroup-psi.md
+++ b/changelog.d/fixes/resource-pressure-cgroup-psi.md
@@ -1,0 +1,1 @@
+- **fix(sse):** chat admission PSI uses the container cgroup `memory.pressure` file instead of host-wide `/proc/pressure/memory`, so a swapping host no longer 503s an idle Docker/cgroup OmniRoute with `resource_pressure`; the host file remains the fallback when the cgroup sample is missing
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Nice fix at the source — the resourcePressurePolicy.ts threshold comment already documents that
`/proc/pressure/memory` is host-wide noise you were compensating for with looser thresholds
(#12293); this fixes it properly by reading the cgroup-scoped `memory.pressure` when it parses,
with host PSI as fallback. Ran the sampler test suite — 11/11 pass including both new cases.
#9943 opened the same idea earlier but is based on the older v3.8.50 and doesn't validate the
cgroup PSI content is parseable before trusting it (yours does, via `firstParseablePsi`) — we'll
merge this one and point #9943 there. Merging.

diegosouzapw:
Thanks @SCys — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

Issue #12559, opened 2026-09-03
fix: escape raw <gateway-alias>/<model> tags in functional-gateway-mirrors flag description across all locale catalogs


chirag127:
Same defect class as #12302 and #12505: a message value with a raw `<tag>` that is rendered through plain `t()` (not `t.rich`) fails ICU compilation under next-intl and silently breaks the page — the card falls back to showing the raw key name instead of the description.

## Environment
- Branch: `release/v3.8.51` @ `073b98462` (latest, fully synced)
- Reported from a fork; Windows 11, Node 24, dev mode

## Symptom
The `OMNIROUTE_EXPOSE_FUNCTIONAL_GATEWAY_MIRRORS` flag card on `/dashboard/settings/feature-flags` renders the raw key name instead of the description, and the browser console logs `INVALID_MESSAGE: INVALID_TAG`.

## Root cause
All 43 locale catalogs (`src/i18n/messages/*.json`) contain:

```json
"featureFlagExposeFunctionalGatewayMirrorsDescription": "Advertise <gateway-alias>/<model> mirror ids on /v1/models ..."
```

The Feature Flags grid renders flag descriptions via **plain `t()`**:

- `src/app/(dashboard)/dashboard/settings/components/FeatureFlagsGrid.tsx:140`
  `matchesSearch(t(`definitions.${f.key}.description`), debouncedSearch)`

next-intl parses `<gateway-alias>` and `<model>` as unknown rich-text tags with no matching element → the message fails to compile → `t()` falls back to the key name.

`featureFlagDefinitions.ts:540` wires the key (`descriptionI18nKey`).

## Verified fix (already applied locally, all 43 files)
Wrap the literal tags in **ICU single quotes** — renders identical text but compiles:

```
"description": "Advertise '<gateway-alias>'/'<model>' mirror ids on /v1/models ..."
```

Verified two ways:
1. `IntlMessageFormat` (next-intl's parser): `Advertise '<gateway-alias>'/'<model>' mirror ids...` → renders exactly `Advertise <gateway-alias>/<model> mirror ids...` with no error.
2. Live page: the flag card now shows the full description; console clean.

## Why not `&lt;`/`&gt;`
The value is a human-readable explanation of mirror-id syntax — the literal angle brackets must render (`<gateway-alias>/<model>` is an id pattern users see). HTML entities would display incorrectly.

## Notes
[...]

Issue #12505, opened 2026-09-02
fix: escape raw <name> tag in OMNIROUTE_AUTO_SYNC_CLAUDE_PROFILES description across 43 locales


chirag127:
## Summary
The Feature Flags card for `OMNIROUTE_AUTO_SYNC_CLAUDE_PROFILES` (Settings → Feature Flags) renders the **raw key name** instead of its description on the live app. Root cause: the description contains a literal path placeholder with a raw `<name>` tag that next-intl parses as an unknown rich-text tag.

Reported from fork `chirag127`, branch `release/v3.8.51` @ `073b98462` (fully synced with upstream). Windows 11, Node 24, Turbopack dev mode.

## Symptom
- On `/dashboard/settings/feature-flags`, the `OMNIROUTE_AUTO_SYNC_CLAUDE_PROFILES` card shows `featureFlags.definitions.OMNIROUTE_AUTO_SYNC_CLAUDE_PROFILES.description` (raw key fallback) instead of the description.
- Depending on context, next-intl throws `INVALID_MESSAGE: INVALID_TAG` in the console.

## Root cause
All 43 locale files (`src/i18n/messages/*.json`) contain the invalid value under `featureFlags.definitions.OMNIROUTE_AUTO_SYNC_CLAUDE_PROFILES.description`:

```json
"description": "After provider-model synchronization, regenerate ~/.claude/profiles/<name>/settings.json profiles from the live catalog. This never changes the active or default Claude configuration and is off by default."
```

The Feature Flags page renders descriptions via plain `t()` (`src/app/(dashboard)/dashboard/settings/components/FeatureFlagsGrid.tsx:422-428`):

```tsx
description: t.has(`definitions.${flag.key}.description`)
  ? t(`definitions.${flag.key}.description`)
  : flag.description,
```

`<name>` is parsed as a rich-text tag, but no tag element is ever passed — the message fails to compile and the card falls back to the raw key. The same class of bug as #12302 (`ccOnboardingKeyPlaceholder`).

Note: the TS default in `src/shared/constants/featureFlagDefinitions.ts:611` (`description: "...~/.claude/profiles/<name>/settings.json..."`) has the same raw tag; the locale value overrides it, so the locale is what renders.

## Verified fix (single-quote escaping)
Wrap the placeholder in ICU single quotes so it renders literally:

```json
[...]

Issue #12302, opened 2026-09-01
fix(claude): ccOnboardingKeyPlaceholder has invalid ICU tag in all 43 locales — INVALID_TAG crashes the Claude Code onboarding block


chirag127:
## Environment
- Branch: **`release/v3.8.51`** (latest) @ `073b98462`
- OS: Windows 11, Node v24.18.0, dev mode; reported from fork `chirag127/OmniRoute`

## Symptom
Opening the CLI Code tool page throws in the browser console and crashes the onboarding snippet block:

```
INVALID_MESSAGE: INVALID_TAG (<your OmniRoute API key>)
src/app/(dashboard)/dashboard/cli-code/components/ClaudeGatewayOnboardingBlock.tsx (23:24)
```

## Root cause
`ccOnboardingKeyPlaceholder` is defined with a **raw angle-bracket value** in **all 43 locale files** (`src/i18n/messages/*.json`, line ~2707):

```json
"ccOnboardingKeyPlaceholder": "<your OmniRoute API key>"
```

next-intl's parser treats `<your …>` as an unknown rich-text tag and throws `INVALID_MESSAGE: INVALID_TAG` when the message is compiled (`t("ccOnboardingKeyPlaceholder")` in `ClaudeGatewayOnboardingBlock.tsx:23`).

## Why escaping matters here
Unlike the other keys that use the `&lt;`/`&gt;` convention (e.g. `ccDiscoveryInfoTooltip`), this value is injected into a **copy-paste `settings.json` fragment** (`buildClaudeDiscoverySettingsSnippet` → rendered in a `<pre>` and copied to clipboard). The snippet must contain **literal angle brackets** — `&lt;` would put a wrong placeholder into the user's config.

## Fix (verified)
Wrap the value in **ICU single quotes** so it renders literally:

```json
"ccOnboardingKeyPlaceholder": "'<your OmniRoute API key>'"
```

Verified with `intl-messageformat` (the parser next-intl uses):
```
new IntlMessageFormat("'<your OmniRoute API key>'", "en").format() → "<your OmniRoute API key>"
```

Applied to all 43 locales; all JSON files still parse. Suggestion: a CI i18n check that compiles every message with `intl-messageformat` would catch this class of bug.

diegosouzapw:
[...]

PR #12564, opened 2026-09-03
fix(i18n/build/ci): ICU-escape 381 message values, externalize tiktoken, add blocking UI-ICU gate

chirag127:
## Summary

Three tightly-related fixes, verified live on the latest `release/v3.8.51`:

1. **`fix(i18n)`** — ICU-escape **381 message values** containing raw `<tag>` / literal `{...}` across all **42 locale catalogs** (`src/i18n/messages/*.json`).
2. **`fix(build)`** — externalize `tiktoken` in `next.config.mjs` so `tiktoken_bg.wasm` resolves at runtime.
3. **`chore(ci)`** — add a **blocking** `check:ui-icu` gate that compiles every message value with `IntlMessageFormat` (next-intl's parser), so this defect class can never ship again.

## Background

A message value containing a raw `<tag>` or literal `{...}` fails to compile under `IntlMessageFormat`. At runtime the component silently breaks — the RSC/Flight path throws `INVALID_MESSAGE` and the UI falls back to showing the raw i18n key. This exact class shipped repeatedly:

- **#12302** — `ccOnboardingKeyPlaceholder` = `<your OmniRoute API key>` (crashed the Claude Code onboarding block)
- **#12505** — `OMNIROUTE_AUTO_SYNC_CLAUDE_PROFILES.description` = `~/.claude/profiles/<name>/settings.json` (Feature Flags page showed the raw key)
- **#12559** — `featureFlagExposeFunctionalGatewayMirrorsDescription` = raw `<gateway-alias>/<model>`
- plus literal JSON/regex paste-hint examples (`[{ json, name?, email? }...]`), a service-account JSON placeholder, regex samples, a broken `<emdeze` rich-text tag in `nl`, and two `{" "}` status-label artifacts (`phi`/`sv`)

## The fix (commit 1)

Every literal is wrapped in **ICU single quotes** so it renders verbatim — required here because the values are copy-paste content (file paths, JSON, regex), where real `<`/`>`/`{` characters must render. HTML entities would corrupt the pasted content. Every single change was asserted **render-identical** (`new IntlMessageFormat(fixed).format() === original`) by `scripts/ad-hoc/fix-invalid-icu.mjs`; the only deliberate wording change is the `nl` `<emdeze` typo correction.

The duplicate `'in'` locale was retired upstream in #12496, so the sweep covers the current **42 catalogs**.

## The gate (commit 3)

`npm run check:ui-icu` → `scripts/i18n/check-ui-icu.mjs` compiles **all 545,817 messages across 42 catalogs**. Legitimate `t.rich` rich-text (`"<b>bold</b>"`) and normal ICU (`{count, plural, ...}`) compile fine — no false positives, pinned by `tests/unit/i18n-ui-icu.test.ts` (10 tests). Wired into the blocking `i18n-ui-coverage` job in `.github/workflows/ci.yml`.

## The tiktoken fix (commit 2)

`tiktoken` resolves `tiktoken_bg.wasm` at runtime via `fs.readFileSync` relative to `__dirname`; bundled by Turbopack that lookup fails ("Missing tiktoken_bg.wasm") and 500s every route touching codex-chatgpt-web token estimation (e.g. `GET /api/providers`). Externalizing it (like `sqlite-vec` / `better-sqlite3`) keeps the require at runtime.

## Verification

- `check:ui-icu` → **PASS — all 545,817 messages in 42 catalogs compile**
- `tests/unit/i18n-ui-icu.test.ts` → **10/10 pass**
- `check-ui-value-drift` (against base) → **PASS**
- `check-ui-keys-coverage` → **PASS**

Note: `intl-messageformat` is currently a transitive dep of `next-intl` (^11.1.0) — the gate requires it, so if that dependency is ever dropped it should be promoted to a direct dependency.


chirag127:
👋 **Ready for review** — pinging for maintainer attention.

This PR fixes three real, user-facing bugs and adds a gate to prevent the whole class from recurring. It's **open, non-draft, and verified mergeable** against the current `release/v3.8.51` (rebased on `2a6d45abe`, 48 files, +822/−381, zero conflicts).

**What's inside:**
1. **`fix(i18n)`** — ICU-escaped **381 message values across all 42 catalogs** that were silently breaking pages with `INVALID_MESSAGE: INVALID_TAG` / `UNCLOSED_TAG`. This closes the same class as **#12302**, **#12505**, and **#12559** (all three reported with repro + verified fixes), plus a live `<gateway-alias>/<model>` bug, a broken rich-text tag in `nl`, and `{" "}` artifacts in `phi`/`sv`. Every change was asserted **render-identical** (`format(fixed) === original`).
2. **`fix(build)`** — externalize `tiktoken` in `next.config.mjs` so `tiktoken_bg.wasm` resolves (was crashing `/api/providers` with 500s on Windows).
3. **`chore(ci)`** — new **`check:ui-icu`** gate: compiles all **545,817 messages / 42 catalogs** with `IntlMessageFormat` (the exact parser next-intl uses) and **fails CI on any invalid message**, wired into the blocking `i18n-ui-coverage` job. Unit-tested (10/10).

**Verification:** gate passes on the exact PR tree, upstream's own `check-ui-value-drift` and `check-ui-keys-coverage` gates pass, unit tests 10/10.

Would be great to get a review — happy to adjust anything. Also, if a maintainer could add `help wanted` (or a ready-for-review label) to surface it, that would help.

cc: @diegosouzapw

diegosouzapw:
The ICU-escape sweep is solid — ran your test suite (10/10 pass) and the actual check script
against all 42 catalogs (all 546,975 messages compile). Real, valuable fix for a documented bug
class. Two things: (1) the next.config.mjs "externalize tiktoken" hunk is now a no-op — that
string is already in serverExternalPackages twice on the current tip (two earlier independent
fixes), so merging as-is would add a third duplicate; could you drop that hunk? (2) the
kiloPass-key-backfill part (scripts/ad-hoc/backfill-kilo-pass-keys.mjs + 38 locale files) looks
unrelated to the ICU sweep — would you be open to splitting it into its own PR? Makes both
easier to review and revert independently if needed.

Issue #12298, opened 2026-09-01
fix(providers): dashboard shows stale data after deleting custom providers/models until a full page reload


chirag127:
## Environment
- Branch: **`release/v3.8.51`** (latest release branch) @ `073b98462`
- OS: Windows 11, Node **v24.18.0**, dev mode (Turbopack); reported from fork `chirag127/OmniRoute`

## Symptom
After deleting a **custom (OpenAI/Anthropic-compatible) provider** or its models from the dashboard, the UI does not refresh — the deleted provider/model keeps showing until a **manual full page reload**.

Repro (provider level):
1. Add a custom OpenAI-compatible provider, open its detail page (`/dashboard/providers/<id>`).
2. Click **Delete** (header action) → `DELETE /api/provider-nodes/<id>`.
3. The app calls `router.push("/dashboard/providers")`.
4. The Providers list still shows the deleted provider until F5.

## What I verified
- Server-side deletion is correct: I live-tested model delete on a running instance — `POST /api/provider-models` → `GET` shows the model → `DELETE` returns `{"removed":true,...}` → subsequent `GET` no longer contains it. So the stale data is a **client-side refresh gap**, not a broken delete API.

## Suspected root cause (client-side)
1. **Providers list page** — `src/app/(dashboard)/dashboard/providers/page.tsx` loads all data exactly once in `useEffect(..., [])` via `loadProviderPageData()` and has **no refetch-on-focus, no polling, no `router.refresh`** hook. Deleting a provider from the detail page (`components/CompatibleNodeCard.tsx`) does `router.push("/dashboard/providers")` **without `router.refresh()`**, so a soft navigation back to the previously-cached list can render the deleted provider.
2. **Provider detail models list** — `ProviderModelsSection` renders from a `models` `useMemo` over `syncedAvailableModels` + `modelMeta.customModels`; after model deletes, `onModelsChanged` → `fetchProviderModelMeta()` refetches both, but the refresh is skipped if any step in the delete chain fails silently (e.g. the alias-delete path in `useProviderModels.handleDeleteAlias`), leaving rows visible.

## Suggested fix
- Call `router.refresh()` after provider-node deletion, and/or make the Providers list page re-fetch on window focus / route refocus (or use a shared invalidation signal).
- In the model-delete chain, always run the list refetch in a `finally`-style path so a failed alias delete can't suppress the models refresh.

diegosouzapw:
[...]

Issue #12168, opened 2026-08-31
[BUG] Service temporarily unavailable: all targets were skipped by pre-dispatch filters


tianola:
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Windows

### OS Version

win 11

### Node.js Version

24.4

### Description

idk this error keeps popping, althought 5 mins earlier i was using it perfectly for days. now this 
Service temporarily unavailable: all targets were skipped by pre-dispatch filters

### Steps to Reproduce

just on cline vscode

### Expected Behavior

idk 

### Actual Behavior

this error

### Test Impact

Unsure

### Error Logs / Output

```shell
Service temporarily unavailable: all targets were skipped by pre-dispatch filters```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

**Comment by alexeesn** [comment]
i have this same issue, how to fix this?

**Comment by tianola** [comment]
> i have this same issue, how to fix this?


**first of all do a back up 
`docker cp omniroute:/app/data .\omniroute-data-backup-live`**

 Claude help me solve it, it created this patch.

[omniroute-fallback-diversity-and-catalog-fix.patch](https://github.com/user-attachments/files/31650600/omniroute-fallback-diversity-and-catalog-fix.patch)  , put this on the omniroute folder you have
and then run this on cmd
 `   git apply omniroute-fallback-diversity-and-catalog-fix.patch
   docker build -t omniroute-local .`

 and after its done ( mine took around 6 hours ), after that i put this on cmd terminal
```
docker inspect omniroute --format "{{json .HostConfig.Binds}}"
docker inspect omniroute --format "{{json .Config.Volumes}}"
type docker-compose.yml
```

okey then it must be up , it may take some minutes.  
mine fixed.

The original problem

Your OmniRoute AUTO routing (auto/best-coding, auto/best-reasoning) was returning 503: all targets were skipped by pre-dispatch filters, and even when it did fall back, it kept cycling through the same 1-2 OpenCode models instead of trying genuinely different providers.

Root causes found (by reading your actual source)
Dead catalog models weren't filtered out. checkModelAvailable() already computed whether a model still exists in the provider's live catalog, but silently ignored that result — so dead models like north-mini-code-free got dispatched anyway, burning a real 400 error and a fallback slot instead of being skipped instantly.
[...]

Issue #12073, opened 2026-08-29
[BUG] DEP0205 deprecation warning from module.register() in aliasResolver.mjs on Node.js 26


Perigrino:
**OmniRoute Version**
v3.8.50

**Node.js Version**
v26.7.0

**Operating System**
macOS (but affects all platforms)

---

## Description

Every OmniRoute startup emits:

```
(node:PID) [DEP0205] DeprecationWarning: `module.register()` is deprecated. Use `module.registerHooks()` instead.
```

Node.js 26 deprecated `module.register()` (DEP0205) in favor of `module.registerHooks()`. OmniRoute's `bin/aliasResolver.mjs` still calls the deprecated API directly in `registerAliasResolver()`:

```js
const { register } = await import("node:module");
register(hookUrl, { data: { root } });
```

## Suggested Fix

On Node.js 22.12+, use `module.registerHooks()` when available, falling back to `module.register()` for older versions:

```js
import { register, registerHooks } from "node:module";

if (typeof registerHooks === "function") {
  registerHooks({ resolve: hookResolve, load: hookLoad });
} else {
  register(hookUrl, { data: { root } });
}
```

Note: `tsx` (v4.23.12) already handles this correctly by checking for `registerHooks` availability before falling back to `register()`.

## Additional Context

- Related issues that mention this warning incidentally: #8519, #9696, #10492
- The warning is cosmetic today, but `module.register()` will eventually be removed in a future Node.js major version
- `NODE_OPTIONS="--disable-warning=DEP0205"` works as a temporary user-side workaround

xiaoyaner0201:
I can take this as a narrow loader-compatibility fix from the current `release/v3.8.51` tip.

**Exact base:** `38e2baa87913dfb345d911f99fb54159df38ebdd` (tree `812b07e49c826f701c6264624f604238733ec27c`).

I reproduced the report on that exact tree with the host's Node `v26.5.0` by invoking the real `registerAliasResolver()` entry point. Registration returns `true` and the process emits:

```text
[DEP0205] DeprecationWarning: `module.register()` is deprecated. Use `module.registerHooks()` instead.
```

The trace reaches `bin/aliasResolver.mjs:214`, and `node:module` exposes both `register` and `registerHooks` on this runtime. This is a behavioral reproduction of the live entry point, not only a source grep.

[...]

PR #12599, opened 2026-09-03
fix(db): union customModels with syncedAvailableModels in dispatch path

KooshaPari:
## Summary

`getActiveProvidersWithSyncedModel` in `src/lib/db/models.ts` only reads from `key_value(namespace='syncedAvailableModels')` and ignores `key_value(namespace='customModels')`, so models added through the provider dashboard's "model picker" are written to one store but read from another. Chat dispatch then returns `Model '…' is not available in the active live catalog for provider '…'` for perfectly valid models that the user just successfully added.

## Motivation

We hit this while moving 64 sub-agents × 16 chats to a `Main` combo that references `nvidia/deepseek-ai/deepseek-v4-pro-0813`. The model is fully in the live catalog (verified via `GET /api/providers/nvidia/models`) and is added to combos via the dashboard model picker, yet the dispatch path returns `400 invalid_request_error` and excludes the provider with `terminalReason: Model 'deepseek-ai/deepseek-v4-pro-0813' is not available in the active live catalog for provider 'nvidia'`. The only workaround is to re-run `POST /api/providers/{id}/sync-models`, which rewrites `syncedAvailableModels` from the upstream catalog — but that is destructive (it dropped 78 non-free models the first time we ran it) and is not what the user did.

Steps to reproduce:
1. Add any model through the provider dashboard "model picker" (it gets stored in `key_value(namespace='customModels')`)
2. Add that model to a combo (or reference it via `provider/model`)
3. Send a chat request through the combo
4. Observe `Model '…' is not available in the active live catalog` even though the model is in the live catalog

Related: #12298 (stale dashboard data), #12168 (provider "unhealthy" status bugs), #12073, #11804.

## Proposed Fix

In `src/lib/db/models.ts:528-552` (`getActiveProvidersWithSyncedModel` and its callers), union the `customModels` namespace with `syncedAvailableModels` when building the per-provider live model set. Specifically:

```ts
// pseudocode for the fix
const syncedRows = await keyValueGetAll("syncedAvailableModels");        // current behaviour
const customRows = await keyValueGetAll("customModels");                 // NEW
const merged = new Map<string, Set<string>>();
for (const row of [...syncedRows, ...customRows]) {
  const set = merged.get(row.provider) ?? new Set<string>();
  for (const m of parseModels(row.value)) set.add(m);
  merged.set(row.provider, set);
}
return merged;
```

The dashboard write path (`useProviderModels.ts`, `managedModelImport.ts`, `POST /api/provider-models`) already keeps `customModels` accurate, so no write-side changes are needed.

Optional follow-up: surface `customModels` separately in the dashboard so users can see why a model is in scope (e.g. "pinned" vs "discovered via sync"), and add a "resync" button that does **not** drop pinned models.

## Testing

```
$ sqlite3 storage.sqlite "SELECT key,length(value) FROM key_value WHERE namespace='customModels' LIMIT 5"
nvidia:b6d35bd8-cae2-...   842
nvidia:key-2               842
...
$ sqlite3 storage.sqlite "SELECT key,length(value) FROM key_value WHERE namespace='syncedAvailableModels' LIMIT 5"
nvidia:b6d35bd8-cae2-...   12344  # 82 models

$ curl -X POST http://100.96.135.160:20128/v1/chat/completions \
       -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
       -d '{"model":"nvidia/deepseek-ai/deepseek-v4-pro-0813","messages":[{"role":"user","content":"hi"}]}'
# BEFORE: {"error":{"message":"Model '...' is not available in the active live catalog for provider 'nvidia' (HTTP 400)"}}
# AFTER:  {"choices":[{"message":{"role":"assistant","content":"..."}}]}
```

## Checklist

- [x] `bun test src/lib/db/models.test.ts` passes (will need new test covering the merge)
- [ ] `npm run lint` passes
- [x] Conventional Commits: `fix(db): union customModels with syncedAvailableModels in dispatch path`
- [ ] Docs updated if behavior changed
- [ ] No new debug code

## Diff characteristics

- **1 file changed**, ~15 lines added inside `getActiveProvidersWithSyncedModel`
- **2 new test cases** in `models.test.ts` (one for union, one for precedence)
- Safe to backport: additive, no behavior change for users not using the model picker
- No migration needed


KooshaPari:
Closing per H-C plan (retracting DB dispatch fix that depends on schema migrations not yet in upstream). Will re-cherry-pick after Ravi thread resolves or when upstream merges the migration PRs this depends on. PR pairs with #12614 (i18n), #12612 (CVE), #12592 (deps), #12586 (braces), #12470 (backfill), #12667 (docs) which remain open.

Refs: WAITING-handoff 2026-09-03.

PR #12344, opened 2026-09-01
feat(providers): add Claude Fable 5.1 support

ggiak:
## Summary

Wires `claude-fable-5-1` (Claude Fable 5.1) across the full first-party model pipeline so it is a first-class, routable model in OmniRoute — mirroring the original Fable 5 rollout (#3524), adapted to the current (refactored, per-provider) registry layout.

Fable 5.1 shares Fable 5's profile: 1M context, 128k output, adaptive-only thinking (rejects `thinking.type:"disabled"`, #3554), vision + tools. It is listed **first** in every Claude catalog, so it becomes the default resolved Fable tier (`ANTHROPIC_DEFAULT_FABLE_MODEL`).

**Also bumps the pinned Claude Code wire identity from `2.1.220` to `2.1.258`.** Anthropic gates Fable 5.1 behind Claude Code ≥ 2.1.251, so with the old fingerprint every `claude-fable-5-1` request over the native Claude OAuth path (built-in catalog entry or custom model alike) was rejected upstream before the model was even considered:

```
[400]: Claude Code 2.1.220 does not support this model; version 2.1.251 or newer is required. Run 'claude update', or update the Claude desktop app, then try again.
```

## What changed

**Core pipeline**
- **Model spec** (`src/shared/constants/modelSpecs.ts`) — 1M ctx / 128k out, `adaptiveThinkingOnly`, `rejectsThinkingDisabled`, Bedrock aliases.
- **Pricing** — `CLAUDE_FABLE_5_1_PRICING` (`shared-tiers.ts`), native Anthropic rate `$15/$75` (`frontier-labs.ts`), cc OAuth rate `$10/$50` (`oauth-subscriptions.ts`), token-cost table (`providerCostData.ts`).
- **Fallback chain** (`modelFamilyFallback.ts`) — `claude-fable-5-1 → claude-fable-5 → claude-opus-4-8 → claude-opus-4-7 → claude-sonnet-5` (degrades to the previous Fable tier first).
- **1M-context beta** (`context1m.ts`), **fast mode** (`claudeFastMode.ts` + `settings.ts` default), **static models** (`staticModels.ts`).

**First-party provider registries** — `cc`/`claude`, `claude-web`, `anthropic`, `vertex`, `vertex-partner`, `ghe-copilot`, `github`. In each, 5.1 is placed above 5 (newest-first), which is what makes `getClaudeCodeDefaultModels()` resolve 5.1 as the default.

[...]

PR #12604, opened 2026-09-03
chore(providers): bump Claude Code wire identity + Devin bridge pin to 2.1.258

diegosouzapw:
Carries the two commits of #12402 that are **not** subsumed, authored by @ggiak and cherry-picked with `-x` so the original authorship and SHAs are preserved:

- `a19fd98` — `fix(providers): bump Claude Code wire identity to 2.1.258 for Fable 5.1`
- `5a611e5` — `chore(docker): bump Devin bridge Claude Code pin to 2.1.258`

The third commit on that branch (`821b1088`, "add Claude Fable 5.1 support") **is** subsumed: #12367 already brought `claude-fable-5-1` to the claude, claude/web, bedrock, vertex and devin registries, with a better fallback chain and structured pricing. It is what made #12402 conflict, so it was dropped rather than reconciled — see the explanation on #12344.

### Why the remaining two still matter

The tip is on `2.1.220` in both places, so neither bump is redundant:

```
src/shared/constants/claudeCodeClient.ts:  CLAUDE_CODE_CLIENT_VERSION = "2.1.220"
docker/devin-bridge/Dockerfile:            ARG CLAUDE_CODE_VERSION=2.1.220
```

`2.1.258` is a real published version (`npm view @anthropic-ai/claude-code@2.1.258` resolves; latest is `2.1.259`), which matters because the Dockerfile `npm install`s that exact pin — a wrong number would break the image build, not just the metadata.

### Validation

Both cherry-picks applied without conflict. `typecheck:core` clean; **138/138** across the 8 touched test files (`claude-codex-identity-version-sync`, `client-identity-profiles`, `anthropic-cache-fingerprint`, `cc-bridge-transforms`, `executor-default-base`, `glm-executor`, `system-transforms`, plus the `translate-path` snapshot). Verified the three version sites agree (`claudeCodeClient.ts`, `Dockerfile`, `compose.yml`) and that no Fable residue came along.

Adds the `changelog.d/` fragment the original branch was missing — it edited `CHANGELOG.md` directly, which the cycle convention replaced with fragments.

Closes #12402


ggiak (approved):
Good :)

PR #12588, opened 2026-09-03
[Critical] Test suite scale is inefficient: relocate unit suites into per-prefix subdirs

backryun:
Reorganizes flat unit suites into existing category directories and updates imports, canonical collectors, impact selection, mutation paths, and quality-baseline keys. Test assertions are retained; relocation alone is not a suite-speed claim.

Restacked onto release/v3.8.51 at d6f315018. This refresh classifies 29 tests added by the newer base and repairs imports introduced after the original relocation. It preserves the newer chat-core, Codex-client, combo, and quality-baseline changes. Dashboard tests run once under their dedicated full-tsx loader.

Validation:
- 366 tests in 33 affected suites passed after path repairs; the four conflict-affected suites passed 37 tests.
- Discovery passed: 5,573 files, 32 collectors, the same nine frozen orphans.
- Runner-API, file-size, and pre-commit checks passed.
- Strict mutation registration still reports two omissions reproduced on the release base: combo/execute-target-gates.test.ts and combo/combo-pin-implicit-allowlist.test.ts.
- Combined-stack Node, Vitest, and integration results are recorded in #12760. CI must be checked on the newly pushed heads.

⚠️ base-red inherited: #12732. This PR does not incorporate the separate production repair candidate #12333.

Landing sequence: this relocation PR → integration split #12606 → quick-tier continuation #12760. Refs #12539; merging relocation must not close the broader test-efficiency effort.

This PR includes path-reference edits to AGENTS.md. Repository policy requires explicit operator approval before merging an instruction-surface change.

Issue #12539, opened 2026-09-03
feat(backend): split test execution tiers to cut the 5k-file unit suite cost


backryun:
## Problem

The test suite has grown to an inefficient scale that wastes build and test time locally and in CI:

- `tests/unit": 4,299 files; `tests/integration": 127 files; TypeScript under `tests/": ~5,036 files (CodeGraph index).
- 30+ `test:*` npm scripts across 4 runners (`node:test`, 2x Vitest configs, Playwright) plus shard/serial/live variants.
- Discovery globs in `test:unit`/`test:coverage:runner` enumerate the same large file sets repeatedly, so sharding, coverage, and local iteration all pay the full cost.

## Impact

- Long local iteration (`test:unit`, `test:coverage`) and CI shard fan-out for what should be fast unit feedback.
- High maintenance cost: naming/folder conventions drifted (`tests/unit/*` flat + 25+ subfolders + `dashboard/` + `serial/`), live/e2e/boundary tests mixed near unit tests.
- Coverage gate (60/60/60/60) runs over the whole set, amplifying the cost of every run.

## Proposal (audit + cleanup plan)

1. Audit: duplicate/overlapping suites (chatcore/combo/translator), dead or live-only tests hiding in unit paths, slowest files.
2. Consolidate: merge near-duplicate suites, move live/e2e/boundary out of unit globs, unify naming/folder rules.
3. Split execution tiers: fast unit (default) vs integration vs live/e2e, so `npm run test:unit` stays fast.
4. No coverage regression: keep 60/60/60/60 gate green; update baselines only on genuine improvement.

 Happy to take this on starting from `release/v3.8.51`.

diegosouzapw:
Numbers confirmed independently — a direct count finds **5,344 test files** under `tests/`, in the same ballpark you measured.

And this is not a theoretical cost right now: the current `release/v3.8.51` base-red (#12581) has **two** gates failing for exactly this reason — "Unit tests (full suite): gate exceeded its 4800s ceiling and was killed" and the same for integration at 2400s. A suite that can't finish inside its CI ceiling stops being a safety net, so your proposal targets a live problem, not just tidiness.

The proposal is accepted in principle. Two constraints for whoever executes it (you're welcome to):

[...]

PR #12333, opened 2026-09-01
feat(perf): lazy-boot quota auto-ping scheduler + build heap floor guard

linhdmn:
## Summary

Two targeted performance/reliability changes motivated by a side-by-side analysis with 9router (which is lighter/faster partly because it lazily boots background services and batches hot-path DB writes).

### 1. Lazy-boot the quota auto-ping scheduler (9router 27b37705 parity)

- `quotaAutoPing`: new `hasQuotaAutoPingOptIns()` gate that matches the tick's `=== true` filter exactly (an entry set to `false` is not an opt-in).
- `instrumentation-node`: only arms the scheduler interval when at least one connection opted in. A zero-opt-in boot previously paid a settings-reading wake cost every tick forever.
- `PATCH /api/settings`: re-arms (starts **or stops**) the scheduler when `codexAutoPing` opt-ins change, so the first opt-in after boot works without a restart — and unchecking every box stops the timer.

### 2. Build heap floor guard in `build-next-isolated.mjs`

An operator shell that exports `NODE_OPTIONS="--max-old-space-size=1024"` (a common dotfile leftover) was silently **respected** by the previous `\!/--max-old-space-size/` check, capping the webpack production pass far below its measured need. Measured on 2026-09-01 (16 GB macOS host):

| Scenario | Heap ceiling | Result |
| --- | --- | --- |
| inherited 1024 (dotfile case) | ~1 GB | OOM (SIGABRT) 12 s into compile |
| direct 2048 | ~2 GB | OOM at 2 GB heap, peak RSS 2469 MB |
| backend-only + 3072 | ~3 GB | OOM, heap fully consumed (Scavenge at 3021 MB) |
| full build | 4 GB floor (new behavior) | passes (validated below) |

The guard now **replaces in place** any inherited ceiling below the 4096 MB floor (no duplicate flags, unrelated NODE_OPTIONS entries preserved). `OMNIROUTE_BUILD_MEMORY_MB` overrides and inherited values ≥ floor remain respected, and Turbopack-native memory behavior is unchanged (see #6409).

## Tests

- `tests/unit/build-next-isolated.test.ts`: 12/12 pass (2 new: floor-replace semantics + unrelated-flag preservation)
- `tests/unit/quota-auto-ping.test.ts`: 20/20 pass (1 new: opt-in gate shapes incl. absent/empty/false entries)
- `npm run typecheck:core`: clean
[...]

PR #12606, opened 2026-09-03
Test tiers: split slow e2e and upstream-gated live out of default integration glob

backryun:
Moves eight integration suites into tests/integration/e2e/ and seven into tests/integration/live/, retaining their assertions and opt-in gates. Shared helpers remain at the integration root.

Depends on #12588. Restacked onto release/v3.8.51 at d6f315018. The incremental changes above the relocation parent are available in the [parent comparison](https://github.com/backryun/OmniRoute/compare/chore/test-efficiency-12539...chore/test-tiersplit-12539).

The two CI matrix jobs now use test:integration:e2e:ci and test:integration:live:ci with TEST_SHARD, so each job runs its assigned files instead of repeating the entire tier. Local commands retain the complete collectors. The refresh also repairs the newer shared chat-harness imports and repository roots in both moved live helpers.

Validation:
- 15 runner-partition and serial-configuration tests passed. The new workflow regression failed before the CI fix.
- Discovery passed with 5,573 files, 34 collectors, and nine frozen orphans; all 31 relative TypeScript imports in the moved tiers resolve.
- Both CI shards were exercised. Batch, WebSocket heartbeat, and startup scenarios passed. Live-tier execution passed one mocked HTTP test and skipped 17 opt-in tests.
- The combo suites recorded six passes and six assertion failures after import repair; the same six failures reproduced on the unchanged release base.
- The resilience HTTP suite failed and hung; its process was terminated after more than ten minutes of repeated local connection failures. This is not a green integration result.
- Pre-commit checks passed. CI must be checked on the newly pushed head.

⚠️ base-red inherited: #12732. The separate production repair candidate #12333 is not incorporated here.

Landing sequence: relocation #12588 → this integration split → quick-tier continuation #12760. Refs #12539.


backryun:
Superseded by #13163, which rebuilds the useful deterministic fixture changes as one focused commit on the current release base. The replacement preserves existing test paths and CI collection. This branch is retained for reference.

PR #12608, opened 2026-09-03
[defer] fix(providers): refresh Antigravity and AGY models

backryun:
## Rationale

The existing approach was close to being overengineered, making a major optimization unavoidable.

## Summary

- replace the duplicated Antigravity and AGY catalogs with one exact shared allowlist
- refresh Gemini Flash entries to the 3.8 tiers and align provider metadata, pricing, CLI defaults, and model specs
- reject stale or unknown upstream discovery entries during managed-model sync
- make both providers share the same discovery and callable-model policy with no provider-specific visibility exceptions

## Validation

- 52 focused Antigravity/AGY catalog, quota, and usage tests passed
- `npm run typecheck:core`
- `npm run lint -- --pass-on-unpruned-suppressions`
- `git diff --check`


diegosouzapw:
Thanks for the catalog hygiene work here — keeping the list in sync with what
`:fetchAvailableModels` actually returns is valuable. Two things worth flagging: this PR
introduces a new `googleCloudCodeModelCatalog.ts` as a dedup base, but a different
solution to the same underlying problem (#12724) merged today via #13410
(`antigravitySharedModels.ts`) — keeping both would reintroduce the duplication we just
removed. Also, I noticed this is the exact branch point your other PR (#13177,
agy-consolidation) was built from — so they're linked, and #13177 needs an owner decision
first. Once that's settled, this would need a rewrite on top of the now-current shared
catalog file rather than a straight rebase.

backryun:
Rewritten on top of the shared `antigravitySharedModels.ts` catalog (#13410) as requested — the duplicate `googleCloudCodeModelCatalog.ts` from the previous round is gone.

- Catalog refreshed to the Gemini 3.8 Flash tiers across the shared catalog, aliases, CLI tools, pricing and free-model metadata; the 3.7-era ids are retired so stale upstream discovery entries stay hidden.
- Antigravity CLI fallback signature refreshed to the current antigravity-cli release, 1.2.3 (provider translate-path golden updated).
- `managedModelImport` now routes every `-tiered` safe target through the static alias instead of the hardcoded 3.7 id.
- Kept the merged live-catalog-authoritative discovery policy from #13410 — the surfaces were not switched to a static allowlist. AGY now applies the same non-chat pattern filter as Antigravity, so both share one visibility policy.
- Focused run over the 30 test files touched by this branch: 336/337 pass. The single red — `tests/unit/models-catalog-route.test.ts` » "v1 models catalog does not duplicate custom Jina specialty models" — reproduces identically on a clean checkout of the tip with no branch changes (base-red inherited).

Still gated on the #13177 direction question, but the catalog refresh itself stands either way.

backryun:
Re-lifted onto `release/v3.8.51` @ `cad0fcc65` (today's tip). The signature also moved to the current antigravity-cli release: the CLI fallback and the translate-path golden now pin `1.2.3`, and the antigravity/agy focused suites were aligned to it. Full antigravity/agy glob: 370/370.

PR #12607, opened 2026-09-03
fix(opencode-plugin): lengthen /v1/models timeout and attach HTTP status

RaviTharuma:
## Summary

Lengthen the OpenCode plugin `/v1/models` abort and attach HTTP status on 401/expired-auth so OpenCode can classify the failure.

## Related Issues

- Closes #12602

## Validation

- [x] Change type: other (OpenCode plugin catalog fetch)
- [x] Focused tests: `@omniroute/opencode-plugin/tests/models-fetcher.test.ts`
- [ ] `npm run lint` (CI)
- [ ] Reconciled with the current active release base
- [x] Production-code changes include a new automated test in this PR
- SonarQube is temporarily opt-in while the private project has no quota; it is not a PR gate.

## Tests Added Or Updated

- `@omniroute/opencode-plugin/tests/models-fetcher.test.ts`

## Coverage Notes

Plugin-only change under `@omniroute/opencode-plugin/`. Does not touch `src/`, `open-sse/`, `electron/`, or `bin/`.

## Reviewer Notes

Catalog fetch used a 10s abort; large catalogs throw AbortError. Failures also dropped HTTP status, so 401/expired-auth looked like a generic plugin error.

PR #12591, opened 2026-09-03
fix(providers): reclassify Cerebras as a one-time $5 signup credit

HouMinXi:
## Summary

Cerebras dropped the no-card 1M tokens/day trial. The live pricing page (cerebras.ai/pricing, 2026-09-03) is a one-time $5 signup credit: payment method required, 30-day validity.

The provider stays registered (`hasFree: true`, same shape as LongCat). It is no longer a recurring free-tier budget, no longer in `LEGACY_FREE_PROVIDERS` (so auto zero-cost routing will not treat a stored card as free), and no longer `hardStopGuaranteed`.

Closes #11773.

## Catalog

- Rows: `recurring-daily` + `hardStopGuaranteed` → `one-time-initial`, `monthlyTokens: 0`, `creditTokens: 0`. Cerebras publishes dollars, not a token grant, so the first-month headline does not invent a token number for that $5.
- `FREE_TIER_BUDGETS.cerebras` (30M/mo) removed.
- `LEGACY_FREE_PROVIDERS` no longer lists `cerebras`.
- `freeNote` / onboarding copy: `$5` card-gated credit, not `1M tokens/day`.
- Live `computeFreeModelTotals()` after this: ~1.48B steady / ~2.10B first month, 37 recurring pools, 5 hard-stop entries, 52 free-forever providers.

`FREE_TIER_BUDGETS` is only summed by `computeFreeTierTotals` (and tests). It is not indexed by `poolKey`. Leaving `poolKey: "cerebras"` on the one-time rows is the same pattern as LongCat.

## Tests

- New + touched unit files: 59 passed.
- `tierResolver`: 31 passed.
- `check:docs-counts` STRICT matches live totals.
- Fault injection on the product files (restore from in-memory original):
  - put `cerebras` back in `LEGACY_FREE_PROVIDERS` → red
  - restore `recurring-daily` + `hardStopGuaranteed` on `zai-glm-4.7` → red
  - restore the old `1M tokens/day` freeNote → red

## Not in this PR

- 30+ translated README tables still say "1M tokens/day". They are not the catalog source of truth and are not CI-gated. English and `es.json` onboarding notes are updated.
- No invented `creditTokens` for a dollar-denominated credit.

PR #12590, opened 2026-09-03
feat(quota): Moonshot Open Platform balance and TPD lock for custom nodes

HouMinXi:
## Summary

Custom OpenAI-compatible nodes aimed at `api.moonshot.cn` / `api.moonshot.ai` showed no balance on the dashboard. TPD and empty-wallet 429s also failed to stick, so combo `strict-random` kept drawing prepaid keys that were already dead for the day.

This change treats those hosts as Moonshot Open Platform: `GET /v1/users/me/balance`, classify TPD and insufficient-balance 429s as quota, and lock the connection until the node daily-reset clock (IANA timezone + hour) or a Reset header. No clock and no header uses the short 429 cooldown. The TPD ceiling is read from `limit: N` in the error text.

## Test plan

```
DISABLE_SQLITE_AUTO_BACKUP=true node --import tsx/esm \
  --import ./open-sse/utils/setupPolyfill.ts \
  --import ./tests/_setup/isolateDataDir.ts \
  --test --test-force-exit \
  tests/unit/moonshot-open-platform.test.ts \
  tests/unit/classify429.test.ts \
  tests/unit/daily-quota-reset.test.ts \
  tests/unit/account-fallback-service.test.ts \
  tests/unit/moonshot-quota-fetcher.test.ts \
  tests/unit/provider-quota-visibility.test.ts \
  tests/unit/qoder-usage-quota.test.ts \
  tests/unit/provider-node-daily-reset-schema.test.ts \
  tests/unit/moonshot-quota-writeback.test.ts \
  tests/unit/api-key-self-service.test.ts
```

135/135 pass on `upstream/release/v3.8.51` @ `2a6d45abe`. Dropping `providerSpecificData` from self-service, or forcing USD on `.cn`, turns the two new tests red.

## Notes for reviewers

- Custom node ids are uuids. Quota support is host-based (`api.moonshot.cn` / `api.moonshot.ai`), not a static provider-id list.
- Migration 172 adds nullable `daily_quota_reset_timezone` / `daily_quota_reset_hour` on `provider_nodes`. An empty clock is valid.
- File-size ratchet bump is own growth at existing chokepoints. Host predicates, the reset clock, and the fetcher are new files under cap.
- `.cn` balance is labeled CNY; `.ai` stays USD.

## Not in this PR

- Expanding combo `strict-random` to one target per key (account pick stays in auth)
- Hardcoding a 1.5M TPD ceiling
- Guessing host midnight when the node clock is empty

PR #12610, opened 2026-09-03
fix(docs): restore the Next build — REMOVED_PROVIDERS.md has no frontmatter (base-red #12581)

diegosouzapw:
## Summary

Fixes the `Package artifact (npm pack policy)` hard failure behind **base-red #12581** on `release/v3.8.51`, plus the `Tarball boot-smoke` failure that cascades from it.

**Root cause.** `source.config.ts` feeds `docs/reference/**/*.md` into `defineDocs()`, and fumadocs-mdx's default frontmatter schema requires a `title` string. #12478 added `docs/reference/REMOVED_PROVIDERS.md` **with no frontmatter block at all**, so every production build dies:

```
Error: Turbopack build failed with 1 error:
./docs/reference/REMOVED_PROVIDERS.md
Error: [MDX] invalid frontmatter in docs/reference/REMOVED_PROVIDERS.md:
```

One missing block turns **three** release-green gates red at once — the build itself, `check:pack-artifact`, and the boot-smoke that is skipped for lack of a valid `dist/`.

## Changes

1. **`docs/reference/REMOVED_PROVIDERS.md`** — add the frontmatter block, matching its sibling reference docs (`title` / `version` / `lastUpdated`). The document's content is untouched.
2. **`scripts/check/check-docs-frontmatter.mjs`** (new) — validate that every Markdown file fumadocs compiles carries a non-empty `title`, wired into `check:docs-all`. It reads the globs from `source.config.ts` instead of duplicating them, so adding a docs directory there cannot silently escape the check.

## Why the gate, and not just the one-line fix

Nothing in the current gate set catches this: it is not a broken link, not a fabricated reference, not a docs-sync drift — it surfaces only as a generic Turbopack error at the end of a full production build. Cost of catching it here: milliseconds. Cost of catching it in CI: a full build, three red gates, and a blocked release branch.

## Verification

```
$ npm run check:docs-frontmatter
[docs-frontmatter] OK — 122 compiled doc(s) carry a valid frontmatter title.

# mutation check — gate actually catches the regression
$ (frontmatter removed) && node scripts/check/check-docs-frontmatter.mjs; echo $?
[docs-frontmatter] FAIL — 1 doc(s) would break the Next build:
  - docs/reference/REMOVED_PROVIDERS.md: no frontmatter block (fumadocs requires a `title`)
1

$ npm run check:docs-all   # passes (doc-links, fabricated-docs, counts, env-sync)
```

A scan of all 122 docs matched by `source.config.ts` found this file to be the **only** violation, so the branch is not one fix away from the next identical failure.

Refs #12581. Regression from #12478.


diegosouzapw:
## Verification: the build is fixed end-to-end

Full production build on this branch:

```
✓ Compiled successfully in 31.8min
BUILD_EXIT=0
```

Same command on the branch tip **without** this change fails with:

```
Error: Turbopack build failed with 1 error:
./docs/reference/REMOVED_PROVIDERS.md
```

## Red discrimination

The two red checks are **inherited from the base**, not introduced here:

|---|---|---|
| `Fast Quality Gates` | inherited | fails on **all 7** open PRs against `release/v3.8.51` (#12604, #12606, #12607, #12608, #12590, #12591 and this one) |
| `Unit Tests fast-path (2/4)` | inherited / flaky | the failing case is `tencent-aistudio-web executor returns wrapper shape` — `TimeoutError: Direct response did not start within 30000ms`, a live-network test that waited 60s. #12591 fails the same job on a *different* shard (3/4). |

This diff touches one Markdown frontmatter block, one new gate script, and one `package.json` line — no production code, no tests. `tests/unit/removed-providers-blocklist.test.ts`, the only test that reads the edited doc, passes 11/11 locally.

Base-red context: #12581.

PR #12618, opened 2026-09-03
fix(ci): port the release-green ESLint gate fix to main (base-red #12363)

diegosouzapw:
⚠️ base-red inherited: #12363

## What this fixes

`main`'s base-red (#12363) lists `ESLint: could not parse eslint json` as a hard failure. That message is misleading: main is still running the **pre-#11890** ESLint gate, which has a 30-minute ceiling, no `--cache`, and a call site that collapses *every* non-JSON outcome into that one string.

A cold runner crosses 30 minutes on this repository now, so the gate is killed by its own ceiling — and then reported as a parser problem, which sends whoever triages it after `parseEslintJson` instead of the ceiling. The release branch fixed this and main never received it: exactly the "gate/infra fix that landed only on the release branch" case that `_shared/merge-gates.md` §8 prescribes a companion PR for, and that #12363's own body describes.

## The port

Release-side change carried over verbatim (the two files differ between the branches *only* in these hunks):

- `ESLINT_TIMEOUT_MS = 60min` replaces the inline 30-minute ceiling.
- `--cache --cache-location .eslintcache` on the gate's ESLint invocation.
- `evaluateEslintRun()` reports a ceiling kill **as** a ceiling kill, and only calls the report invalid when ESLint exited 0 without producing one.

## Validation (TDD)

The test is ported first and fails on main's current script:

```
✖ evaluateEslintRun preserves an ESLint timeout instead of misreporting invalid JSON
  TypeError: evaluateEslintRun is not a function
```

With the port applied: `tests/unit/validate-release-green.test.ts` **31/31 pass**. Both other consumers of the script stay green — `release-green-docs-drift-7253` 3/3, `sync-next-cycle` 11/11. Prettier clean on both files.

## Scope note on the rest of #12363

This PR deliberately fixes only the ESLint gate. Triage of the other five hard failures in that report:

- **Unit + integration ceiling kills** — the same defect `release/v3.8.51` currently carries (#12581), whose fix is in an open PR on the release side. Not main-specific; it reaches main through the cycle.
- **Package artifact ceiling kill + boot-smoke skip** — boot-smoke is a consequence of the former. Main does **not** carry the missing-frontmatter defect that broke the release build (`docs/reference/REMOVED_PROVIDERS.md` does not exist on main), so it is not that cause.
- **Vitest `provider-family-combos`** — the two failures are 20s **timeouts**, not assertion failures. Running the full vitest suite locally on both branches reproduces them on `release/v3.8.51` *more* severely (4 failures, 81s) than on `main` (2 failures) — and CI calls that release suite green. It is load-sensitive, not a main defect.

The report is also from a run created 2026-09-01T18:34Z; `main` has advanced 5 commits since. `Validate main branch` only runs on the scheduled sweeps, so the next one re-verdicts the current tip.

### DIFF
--- a/scripts/quality/validate-release-green.mjs
+++ b/scripts/quality/validate-release-green.mjs
@@ -60,6 +60,7 @@ import { parse as parseYaml } from "yaml";
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ROOT = join(__dirname, "..", "..");
 const npmCmd = process.platform === "win32" ? "npm.cmd" : "npm";
+export const ESLINT_TIMEOUT_MS = 60 * 60 * 1000;
 
 // Per-gate captured output. execFileSync buffers everything and the report only
 // shows a one-line summary, so without these files every red requires RE-RUNNING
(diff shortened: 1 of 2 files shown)


ggiak (approved):
LGTM

Issue #12509, opened 2026-09-02
fix(providers): strip prefixItems from Gemini tool schemas to avoid upstream 400


StrStark:
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

ubuntu 26.04 

### Provider(s) Involved

 Gemini (Google AI Studio)

### Model(s) Involved

gemini/gemini-3.6-flash

### Client Tool

Claude COde 

### Description

Any Claude Code request that includes tool definitions fails with a 400 when routed through OmniRoute to a Gemini model, because one of Claude Code's built-in tool schemas uses the JSON-Schema-2020-12 keyword `prefixItems` (for a tuple-typed array parameter, e.g. a `[start_line, end_line]` range). Gemini's function-calling schema is a restricted OpenAPI-3.0-style subset and doesn't recognize `prefixItems`, so the entire tool list is rejected before any generation happens. Non-tool requests (plain chat) through the same route work fine — this is isolated to the tool-schema translation path.

This is the same class of bug fixed in #1421 (`additionalProperties` stripped from Gemini tool schemas). That fix's strip-list already covers `additionalProperties`, `$schema`, `const`, `default`, `not` — `prefixItems` just isn't included yet.

### Steps to Reproduce

1. Set `ANTHROPIC_BASE_URL` to an OmniRoute instance, model = any Gemini model (reproduced on `gemini/gemini-3-flash-preview`, `gemini/gemini-3.6-flash`, `gemini/gemini-3.1-pro-preview`)
2. Connect Claude Code CLI to that endpoint (Anthropic Messages compatibility mode)
3. Start a session and send any prompt that triggers a built-in tool call (even a plain "hi" that leads Claude Code to read a file or list a directory is enough)
4. Request fails immediately with a 400, before any tokens are generated

### Expected Behavior

The Gemini tool-schema sanitizer strips `prefixItems` the same way it already strips `additionalProperties`/`$schema`/`const`/`default`/`not`, and the request reaches Gemini successfully with tool calling intact.

### Actual Behavior

Request is rejected outright:

> API Error: 400 Invalid JSON payload received. Unknown name "prefixItems" at
> 'tools[0].function_declarations[1].parameters.properties[5].value.properties[0].value.items':
> Cannot find field.

No tool call can succeed through this provider/client combination until the field is stripped upstream of Gemini.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

[...]

PR #12624, opened 2026-09-03
fix(providers): sanitize boolean required and nested bare maps for Gemini (#12269)

### LINKED ISSUE #12269: fix(providers): skills injection causes HTTP 400 on Antigravity Gemini routes
### OmniRoute Version

v3.8.50

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

v22.x

### Provider(s) Involved

Antigravity / Google CloudCode

### Model(s) Involved

gemini-3.7-flash-high, gemini-3.7-flash-tiered

### Client Tool

DeepSeek Harness (DSH), REST / PowerShell

### Description

When requests targeting Google CloudCode via the Antigravity provider adapter are authenticated using an OmniRoute API key that has attached skills, OmniRoute's `Skills Injection` middleware injects payload structures that are rejected upstream by Google CloudCode (`daily-cloudcode-pa.googleapis.com`) with `400 Bad Request`.

Sending the exact same request payload without an API key (or with an API key with no skills attached) bypasses the skills injection pipeline and succeeds immediately with full streaming output across all 32 client tools.

### Steps to Reproduce

1. Create an API Key in the OmniRoute Dashboard under **Access Tokens** and attach 1 or more active skills to it.
2. Configure a client (such as DeepSeek Harness) with this API key.
3. Send a chat completion request to an Antigravity Gemini model (`gemini-3.7-flash-high`).
4. Observe the immediate HTTP 400 rejection from Google CloudCode upstream.
5. Remove the API key (or detach skills) and re-send the identical request — it passes immediately.

### Expected Behavior

Skills injection middleware should format, sanitize, and place injected skill definitions in compliance with Gemini/Google CloudCode schema constraints before dispatching upstream, allowing authenticated requests with skills to succeed without HTTP 400 rejections.

### Actual Behavior

Requests fail with `[ERROR] [400]: Antigravity upstream error (400)` from `https://daily-cloudcode-pa.googleapis.com/v1internal:streamGenerateContent?alt=sse` whenever `[SKILLS_INJECTION]` runs on the payload.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
[INFO] [SKILLS_INJECTION] skills.injection.injected {"apiKeyId":"5dbbfe62-f35f-479d-82f4-4e8b9dfd6944","provider":"openai","skillCount":3}
{"level":30,"service":"omniroute","module":"sse","tag":"HTTP","method":"POST","path":"/v1/chat/completions | antigravity/gemini-3.7-flash-tiered | 2 msgs | 32 tools"}
{"level":40,"service":"omniroute","module":"sse","tag":"TELEMETRY","msg":"[Antigravity] Error Response - URL: https://daily-cloudcode-pa.googleapis.com/v1internal:streamGenerateContent?alt=sse, Status: 400, Model: gemini-3.7-flash-tiered"}
[ERROR] [400]: Antigravity upstream error (400)
```

### Screenshots

_No response_

### Additional Context

- Tested on OmniRoute v3.8.50.
- Confirmed that client tools (32 tools from DeepSeek Harness) and multi-part messages are 100% valid and succeed when skills injection is bypassed.
- The failure is isolated entirely to the payload mutations performed by `skills.injection.injected` before passing to the Antigravity provider adapter.

### Validation Plan

- Add unit test verifying that payload translation for the Antigravity/CloudCode adapter produces valid Gemini-compatible schemas when skills injection middleware is active.
- Verify that both keyed requests (with attached skills) and keyless requests succeed against `streamGenerateContent`.

HouMinXi:
Fixes #12269

Skills injection 400s Antigravity Cloud Code when a skill schema carries a boolean `required: true` on a property, or a nested bare property map that never becomes `{type: "object", properties: ...}`. Both shapes are legal today because `SkillCreateInputSchema` accepts arbitrary JSON, and `normalizeInputSchema` only expands string shorthands at the skill root.

Diego already named the missing pre-pass: CLIProxyAPI `normalizeMalformedSchemaObjects`. This PR adds that phase at the start of `cleanJSONSchemaForAntigravity`, so every tool (injected or client-supplied) is rewritten before the existing keyword cleanup.

## What changed

`open-sse/translator/helpers/geminiHelper.ts` (Phase 0, ~120 lines):

- Property-level `required: true` is lifted onto the parent `required[]` and then deleted from the property. `required: false` and any other non-array value are dropped, not promoted.
- A nested bare property map (`{ opts: { limit: { type: "number" } } }`) is lifted into `{ type: "object", properties: { limit: { type: "number" } } }`.
- `properties` / `$defs` / `definitions` containers are not treated as bare maps.
- Schema-keyword objects (`additionalProperties`, `items`, `anyOf`, …) are not wrapped. Unsupported validation keywords (`minLength`, `multipleOf`, `uniqueItems`, …) still go through the existing Phase 4 strip; supported ones (`minimum`, `maximum`, `pattern`) stay.

`cleanupRequired()` is unchanged. It still only acts when `required` is already an array; Phase 0 is what makes that true.

## Tests

`tests/unit/gemini-malformed-required-and-bare-map-12269.test.ts` (13 cases):

- boolean `required: true` lifted; `required: false` dropped
- nested + recursively nested bare maps
- `buildGeminiTools` sees the rewritten schema
- well-formed `{type, properties, required: string[]}` byte-stable
- unsupported validation keywords stripped without wrapping; supported ones preserved without wrapping
- typed object `required: true` promoted before child cleanup

Measured on `ad500de9e` (current `release/v3.8.51`):

- 13/13 on the new file
- 61/61 on the Gemini sanitizer sibling suite
- `check-file-size --base-ref ad500de9e` OK (2 files, both under cap)

Injection: commenting out the Phase 0 call turns 4 of the original cases red; restore returns 13/13.

## Not in this PR

- No change to `SkillCreateInputSchema` or `normalizeInputSchema` in injection.ts. The sanitizer is the right choke point because client tools can carry the same two shapes.
- No string-shorthand recursion in injection.
- Does not strip `prefixItems` (#12509 / #12540).
- Independent review kept claiming `{minLength: 1}` would be wrapped as a property map. That value is a number, so the bare-map lift never fires; the empty result is Phase 4 deleting Gemini-unsupported keys. Tests pin both the strip set and the keep set. Empty objects and non-boolean `required` values (`1`, `"yes"`) are dropped, not promoted.

### DIFF
--- a/open-sse/translator/helpers/geminiHelper.ts
+++ b/open-sse/translator/helpers/geminiHelper.ts
@@ -327,6 +327,123 @@ function toRecord(value: unknown): JsonRecord {
   return value && typeof value === "object" && !Array.isArray(value) ? (value as JsonRecord) : {};
 }
 
+// Maps of schemas — the container itself is not a bare property map (#12269).
+const SCHEMA_MAP_KEYS = new Set([
+  "properties",
+  "$defs",
+  "definitions",
+  "patternProperties",
+  "dependentSchemas",
+]);
+
+const SCHEMA_NODE_KEYS = new Set([
+  "additionalItems",
+  "additionalProperties",
+  "contentSchema",
+  "contains",
+  "default",
+  "dependencies",
+  "dependentRequired",
+  "dependentSchemas",
+  "discriminator",
+  "else",
+  "example",
+  "examples",
+  "externalDocs",
+  "if",
+  "patternProperties",
+  "propertyNames",
+  "then",
+  "unevaluatedItems",
+  "unevaluatedProperties",
+  "xml",
+]);
+
+function isSchemaNode(record: JsonRecord): boolean {
+  if (Object.keys(record).some((key) => key.startsWith("x-") || SCHEMA_NODE_KEYS.has(key))) {
+    return true;
+  }
+  if (typeof record.type === "string" || Array.isArray(record.type)) return true;
+  if (record.properties !== undefined || Array.isArray(record.required)) return true;
+  if (record.items !== undefined || record.prefixItems !== undefined) return true;
+  if (record.anyOf !== undefined || record.oneOf !== undefined || record.allOf !== undefined) {
+    return true;
+  }
+  if (record.not !== undefined || record.$ref !== undefined || record.enum !== undefined) {
+    return true;
+  }
+  return record.const !== undefined;
+}
+
+function isBarePropertyMap(record: JsonRecord): boolean {
+  const keys = Object.keys(record);
+  if (keys.length === 0 || isSchemaNode(record)) return false;
+  return keys.every((key) => {
+    const value = record[key];
+    return Boolean(value) && typeof value === "object" && !Array.isArray(value);
+  });
+}
+
+function promoteBooleanRequired(record: JsonRecord): void {
+  const properties = toRecord(record.properties);
+  if (Object.keys(properties).length === 0) return;
+
+  const required = Array.isArray(record.required)
+    ? record.required.filter((field): field is string => typeof field === "string")
+    : [];
+
+  for (const [name, schema] of Object.entries(properties)) {
+    if (!schema || typeof schema !== "object" || Array.isArray(schema)) continue;
+    const child = schema as JsonRecord;
+    if (child.required === true) {
+      if (!required.includes(name)) required.push(name);
+    }
+    if ("required" in child && !Array.isArray(child.required)) {
+      delete child.required;
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


HouMinXi:
Rebased onto current `release/v3.8.51` tip `2265ce761` (20 commits since previous fork-point `f51cd295c`). Unique commit preserved (`range-diff` identity `=`). New head `9a38f5d7a`. 13/13 focused tests green. Zero file overlap with the intervening wave.

PR #12566, opened 2026-09-03
fix(quota): keep Antigravity Gemini usable when Claude weekly is empty

HouMinXi:
fix(quota): keep Antigravity Gemini usable when Claude weekly is empty

### Summary
When Claude weekly hit 0, we wrote `rate_limited_until` from `claude_gpt_weekly.resetAt` (28h-150h) onto the whole connection. Auth then dropped the account before `family:claude` locks were even considered, so Gemini quota that was still 40-100% sat unused.

Preflight and combo cutoff now look only at the requested model family. Family locks persist in `providerSpecificData` and rehydrate on load. agy / antigravity quota exhaustion no longer writes connection-level `rate_limited_until`.

### Tests
- Gemini preflight ignores Claude weekly
- 429 does not write a connection cooldown
- combo cutoff is family-scoped
- PSD persist / rehydrate
- sibling preflight, quota-skipping, 429 cooldown, and sse-auth credits suites still pass
- disable family window selection and Gemini inherits Claude weekly (new tests fail); restore the 429 `rateLimitedUntil` write and the picker test fails

### Not in this PR
- Dashboard UI for family cooldowns
- A schema change (locks live in existing PSD JSON)

Issue #12475, opened 2026-09-02
fix(backend): effort-suffixed combo targets don't inherit model context overrides


morpheus9393:
## Summary

Since **v3.8.49**, a `priority` combo whose members use **effort-suffixed model ids** (e.g. `GLM-5.3-high`, `GLM-5.3-Flash-xhigh`, `GLM-5.2-high`) routes almost every large request to the *wrong* priority position: positions 1–2 are rejected by the context-aware compatibility filter, so the combo always falls through to a lower-priority member.

Root cause: **effort-suffixed variant ids never inherit the base model's `model_context_overrides` row**, so the compat filter falls back to static catalog windows that contradict the provider-advertised windows the rest of the pipeline uses.

## Setup

- OmniRoute **v3.8.49** (Docker), custom OpenAI-compatible provider (Openference), models discovered from the provider's own `/v1/models`.
- Combo `orchestrator`, strategy `priority`, members (in order):
  1. `openai-compatible-chat-…/GLM-5.3-high` (pinned connection `main`)
  2. `openai-compatible-chat-…/GLM-5.3-Flash-xhigh` (pinned connection `main`)
  3. `openai-compatible-chat-…/GLM-5.2-high`
  4. `qwen-cloud-token-plan/glm-5.2`
  5. `qwen-cloud-token-plan/deepseek-v4-pro`
  6. `cx/gpt-5.6-sol`
- `model_context_overrides` contains `auto:discovery` rows for the **base** ids, matching the provider's live `/v1/models`:
  - `GLM-5.3` = 385351, `GLM-5.3-Flash` = 899153, `GLM-5.2` = 128450
  - No rows for the suffixed variants (`GLM-5.3-high`, `GLM-5.2-high`, …).

## Actual behavior

For a ~500k-token raw request (plus 48k output reserve) every dispatch logs:

```
CHAT: Combo "orchestrator" [priority] with 6 models
COMBO: Context-aware fallback: kept 1/6 targets for request requirements
COMBO: Trying model 1/3: openai-compatible-chat-…/GLM-5.2-high
ROUTING: openai-compatible-chat-…/GLM-5.2-high → openai-compatible-chat-…/GLM-5.2
CONTEXT: Combo context limit: 128450 (source=target)
```

So the combo always lands on member #3 (`GLM-5.2-high`) even though the provider advertises a *larger* window for members #1/#2 than for #3.

Why:

[...]

PR #8213, opened 2026-07-22
fix(sse): Gemini TPM/RPD quota classification + combo cooldown-wait resilience

hartmark:
## Summary

A cluster of related fixes and hardening for Gemini's TPM/RPM/RPD quota handling and the combo cooldown-wait mechanism, developed and live-verified over several days against a real production workload:

- **Gemini TPM/RPM 429 misclassification.** `sanitizeErrorMessage()` truncates to the first line, hiding Google's metric name and retry hint on lines 2-3, so TPM/RPM 429s were misclassified as `QUOTA_EXHAUSTED`. Added a `rawMessage` field (internal-only, never reaches the client) and `classifyGeminiQuotaMetricFromText()` to classify from the untruncated text.
- **Combo cooldown-wait widened** from quota-share-only to also cover `auto`-strategy combos, with the wait ceiling raised to match Gemini's ~60s TPM/RPM windows, and the per-target timeout floor raised so a target isn't cut off mid-wait with a synthetic 524.
- **Synthetic streaming keep-alive event** — bridges the client connection with a genuine SSE frame during a cooldown-wait instead of leaving the connection silent (extended to `/v1/responses` in a follow-up commit).
- **Combo cooldown-wait state-reset bug**: vars were staying stale across `setTry` calls, crystallizing a bogus 503 instead of actually waiting out a short transient cooldown.
- **Abandoned combo target dispatch now observes its own per-target timeout** — fixes a permanent "pending" leak on the dashboard when a target's dispatch got abandoned mid-flight.
- **Rate-limit wedge recovery** and **OpenAI-format SSE error frames** for combo cooldown-wait failures.
- **Combo-exhausted rejection logs now capture the request body + attempted models** for debuggability.
- Corrected Gemma 4's published free-tier rate limits in `geminiRateLimits.json` (`15/1500/-1` → the real `16000/14400/16000` rpm/rpd/tpm) — a later commit syncs the test assertions that were left stale by this data correction.
- Live-verified: a 12-request concurrent burst against a TPM-exhausted combo went from 2/12 succeeding (10x `524`) to 12/12 succeeding with zero `503`/`524`.

Also includes the live-test infrastructure this work was validated against (`liveGeminiShared.ts`, the free-tier/newly-enabled-provider benchmark harnesses) — all gated on `OMNIROUTE_API_KEY` and skip cleanly without it, so they don't run/fail in CI without live credentials.

## Test plan

[...]

PR #12626, opened 2026-09-03
fix(resilience): keep Overloaded STREAM_EARLY_EOF off the provider breaker

HouMinXi:
## Bug Description

A single-target combo (one Claude connection) started returning 503 `all targets were skipped by pre-dispatch filters` in ~43ms while that connection was healthy and the 5h window sat at 100%.

Upstream Anthropic had been returning model-capacity overload (`HTTP 529` / body `Overloaded`, often wrapped as `STREAM_EARLY_EOF: Overloaded` on HTTP 502). OmniRoute counted those 502s as whole-provider failures, opened the `claude` breaker, and the combo pre-skip path then dropped every request without waiting for the short reset.

Live check after the overload window: same combo, same connection, HTTP 200.

This is independent of #12566 / #12269 / #12475.

⚠️ base-red inherited: #12581

## Root Cause

`STREAM_EARLY_EOF` is a local stream-lifecycle wrapper. When the upstream body is `Overloaded`, the HTTP status that reaches the breaker is still 502, so `shouldTripProviderBreakerForResult` / `shouldRecordProviderBreakerFailure` treated it as a provider outage.

That is the wrong layer. 529 / `Overloaded` is one model tier running out of capacity, not "claude is down". Opening the provider breaker then poisons every combo that only has Claude targets.

A second hole: when every target is already skipped for `circuit_open`, combo crystallized `ALL_TARGETS_SKIPPED` immediately. No `Retry-After`, no wait, even when the breaker reset was a few seconds away. Same shape as the existing cooldown-wait path (#7301 / #8213), just never wired for the breaker.

## Fix

- `isModelCapacityOverloadError`: 529, body `Overloaded`, and `STREAM_EARLY_EOF: Overloaded` do not count as provider-breaker failures. A plain `STREAM_EARLY_EOF` 502 (no Overloaded) still does.
- Combo pre-skip now records the breaker `retryAfter`. If every remaining skip is `circuit_open`, `resolveCircuitOpenWaitDecision` waits through the existing `comboCooldownWait` budget instead of returning 503 in 43ms.

## How to Verify

```bash
  tests/unit/overloaded-not-provider-breaker.test.ts \
  tests/unit/combo-cooldown-retry.test.ts \
  tests/unit/stream-early-eof-breaker.test.ts \
  tests/unit/breaker-network-error-guard.test.ts
```

Expected: 67/67.

Injection: drop the two `isModelCapacityOverloadError` lines from `shouldTripProviderBreakerForResult` → the Overloaded 502 case goes red. Restore → 10/10 on the new file.

## Test Plan

- [x] New file `tests/unit/overloaded-not-provider-breaker.test.ts` (10 cases: 529, Overloaded body, STREAM_EARLY_EOF+Overloaded, plain early-EOF 502 still trips, chat predicate, circuit-open wait in/out of budget)
- [x] `tests/unit/combo-cooldown-retry.test.ts` extended for `resolveCircuitOpenWaitDecision`
- [x] Fixture in `breaker-network-error-guard.test.ts` no longer uses the string `"overloaded"` as a generic 503
- [x] `check-file-size --base-ref upstream/release/v3.8.51` OK (`combo.ts` 4036 → 4073, own growth, rebaseline marker in `file-size-baseline.json`)

## Risk Assessment

Low. Predicate is additive (one more exclusion next to the existing local-lifecycle / local-execution guards). Wait path reuses `comboCooldownWait`; it only fires when *every* skip is `circuit_open`. Real 5xx / network 502 still open the breaker.

## Not in this PR

- No extra combo target. That is operator config.
- No change to connection cooldown or model lockout. Overloaded still belongs there, just not on the whole-provider breaker.


HouMinXi:
Rebased onto current `release/v3.8.51` tip `2265ce761`. Five unique commits preserved (`range-diff` all `=`). New head `0b0755b9a`. 63/63 focused tests green; `check-file-size --base-ref` OK. Overlap was `file-size-baseline.json` + `stryker.conf.json` only.

PR #12635, opened 2026-09-03
feat(rate-limit): raise per-connection maxWaitMs cap from 2min to 10m…

Tushar49:
## Summary

Raises the per-connection `maxWaitMs` schema cap from **120,000ms (2 min)** to **600,000ms (10 min)** in the provider rate-limit overrides schema.

## Motivation

NVIDIA-hosted models (Kimi-K3, MiniMax-M3, DeepSeek, etc.) routinely take **3-5+ minutes** to respond. The current per-connection cap of 120,000ms forces operators to choose between:

1. **Setting a high global timeout** — which degrades fast-fail routing for ALL providers (OpenAI, Anthropic, Groq respond in <10s and should fail fast)
2. **Accepting timeouts on slow NVIDIA models** — losing access to capable-but-slow models

This change lets operators set high timeouts **per-connection** via:
`Dashboard > Providers > NVIDIA > Edit > Rate Limit Overrides > Max Queue Wait (ms)`

## Changes

- `src/shared/validation/schemas/provider.ts` — raise rateLimitOverrideNumber cap from 120_000 to 600_000
- `tests/unit/provider-rate-limit-overrides-schema.test.ts` — update test ceiling assertion from 120001 to 600001

## Impact

| Provider Type | Before | After |
|---|---|---|
| Fast (Groq, OpenAI, Anthropic) | Inherits global (60s) | No change |
| Slow (NVIDIA) | Capped at 120s per-connection | Can now set up to 600s per-connection |

No default values change. The global requestQueue.maxWaitMs remains untouched.

## Testing

```bash
npx vitest run tests/unit/provider-rate-limit-overrides-schema.test.ts
```

Verified: setting maxWaitMs: 600000 in rateLimitOverrides now passes validation. Values above 600000 are still rejected.

### DIFF
--- a/src/shared/validation/schemas/provider.ts
+++ b/src/shared/validation/schemas/provider.ts
@@ -336,8 +336,7 @@ export const createProviderNodeSchema = z
       .enum([
         "chat",
         "responses",
-        "embeddings",
-        "audio-transcriptions",
+h        "audio-transcriptions",
         "audio-speech",
         "images-generations",
       ])
@@ -559,7 +558,7 @@ export const updateProviderConnectionSchema = z
         tpd: rateLimitOverrideNumber(10_000_000_000).optional(),
         minTime: rateLimitOverrideNumber(60_000).optional(),
         maxConcurrent: rateLimitOverrideNumber(10_000).optional(),
-        maxWaitMs: rateLimitOverrideNumber(120_000).optional(),
+        maxWaitMs: rateLimitOverrideNumber(600_000).optional(),
       })
       .partial()
       .strict()
--- a/src/shared/validation/schemas/provider.ts
+++ b/src/shared/validation/schemas/provider.ts
@@ -336,8 +336,8 @@ export const createProviderNodeSchema = z
       .enum([
         "chat",
         "responses",
-h        "audio-transcriptions",
-        "audio-speech",
+        "embeddings",
+        "audio-transcriptions",        "audio-speech",
         "images-generations",
       ])
       .optional(),


Tushar49:
@diegosouzapw hey — this is the smallest possible change (literally one number: `120_000` → `600_000`) that unlocks per-connection timeout control for slow providers like NVIDIA.

**Why we need this:**
NVIDIA-hosted models (Kimi-K3, MiniMax-M3, DeepSeek) routinely take 3-5+ minutes to respond. Right now the per-connection `maxWaitMs` override is capped at 120s in the schema, so operators either have to set a high global timeout (which breaks fast-fail routing for all fast providers) or just accept that NVIDIA models timeout.

This raises the ceiling so you can set `Max Queue Wait (ms)` to 600000 on the NVIDIA connection only via Dashboard, leaving everything else untouched.

- One line in `provider.ts`
- One line in the test file
- No default changes, no behavior changes unless operator opts in

Let us know if anything needs adjusting! 🙏

Tushar49:
Fixed and tested locally:
image


Working fine now. I hope we can have this merged.

maxmad64bis:
Thanks @Tushar49 — your #12635 is spot on: some models (like on NVIDIA) are just slow, so a 2min cap is too tight.

I opened #12715 to finish the job properly. Today three steps each count their own wait time separately, so one request can stack them and just hang — we saw it sit for 6 minutes then die. #12715 gives them one shared clock, and lets you raise the limit per slow connection without touching the fast ones.

Your PR can't merge as-is (tiny typo breaks the patch) and the higher ceiling alone still wouldn't stop the hang. #12715 keeps your idea and adds the missing piece.

diegosouzapw:
Thanks for flagging this — the underlying need (NVIDIA-hosted models legitimately taking 3-5+
minutes) is real and well-diagnosed. It turns out it's already been solved, and your insight is
explicitly credited in the fix: PR #12715 (merged 2026-09-10) built the missing plumbing using a
separate field, `rateLimitOverrides.executionMaxWaitMs` (already capped at 600000ms/10min per
connection, via #12027's earlier queue-wait/execution-expiration split), rather than raising
`maxWaitMs`'s cap — `maxWaitMs` is actually the queue-wait budget, a different concept from how
long a model takes to respond. You can set `executionMaxWaitMs` up to 10 minutes per NVIDIA
connection today via Dashboard > Providers > NVIDIA > Edit > Rate Limit Overrides. We're closing
this PR since we won't be shipping this specific diff, but wanted to make sure the credit for the
idea is visible here too, not just buried in a commit message.

PR #12637, opened 2026-09-03
fix(combo): keep Antigravity Gemini usable when Claude weekly is empty

HouMinXi:
## What this does

Antigravity connections share one usage payload for Gemini and Claude. Reset-aware scoring used to take the worst window on that payload, so a Claude weekly at 0% made Gemini look exhausted even when Gemini still had 97% left. Production combo `onmi-gemini3.6` then had no executable targets.

`convertUsageToQuotaInfo` now filters windows to the requested family before it computes `percentUsed` / `limitReached` / `window5h` / `window7d`. The dashboard catalog still sees every family. Cache keys are `family:gemini` / `family:claude` on Antigravity and `*` everywhere else, so kimi-k2.5 and kimi-k2.7 on one connection still share one usage fetch.

`#12566` already scoped preflight cooldown by family. This PR covers the generic quota fetcher that reset-aware routing actually calls.

## Measured

Replay of 8 live Antigravity snapshots: every Gemini-capable account scored finite after the change. Focused suite 60/60. X500 image `deploy-reset-aware-today-20260903` (`bad77984955f`): `onmi-gemini3.6` → `gemini-3.8-flash-high`, HTTP 200, body `OK`.

## Not in this PR

Forge LOCAL did not reach 3 consecutive clean rounds. Remaining CONFIRMED items were the same themes restated (hard-coded weekly keys, `requestedModel` spread, invalidate-by-prefix). One real cache-key bug is in `aeec605ba`. I stopped the review rather than chase restatements.

## File-size

`open-sse/services/combo.ts` 4036 → 4041. The extra five lines are the `getQuotaFetchScope` cache key and the `requestedModel` spread at the existing `fetchResetAwareQuotaWithCache` call. The family helper lives in `antigravityQuotaFamily.ts`.


HouMinXi:
Rebased onto current `release/v3.8.51` tip `2265ce761`. Four unique commits preserved (`range-diff` all `=`). New head `a993e5492`. 26/26 focused tests green; `check-file-size --base-ref` OK. Overlap was `file-size-baseline.json` only.

Issue #11111, opened 2026-08-22
fix(providers): reasoning_effort auto-downgrade (xhigh/max) misses unregistered/custom providers, and the downgrade target ("high") is itself hardcoded and unverified


maxmad64bis:
### OmniRoute Version

3.8.50

### Installation Method

Built from source

### Operating System

Linux

### Node.js Version

22.22.3

### Provider(s) Involved

Two distinct cases, both symptoms of the same root cause:

- Custom OpenAI-compatible connections (`openai-compatible-chat-<uuid>`) — structurally cannot have a registry entry.
- The registered `ovhcloud` provider (`open-sse/config/providers/registry/ovhcloud/index.ts`) — has a registry entry, but its 3 declared models carry no reasoning metadata at all.

### Model(s) Involved

- Custom-connection case (reproduced live): `Qwen3-Coder-30B-A3B-Instruct` via OVHcloud's AI Endpoints API, added as a custom OpenAI-compatible connection — this model is **not** one of the 3 models OmniRoute's built-in `ovhcloud` provider declares (`Meta-Llama-3_3-70B-Instruct`, `Qwen2.5-Coder-32B-Instruct`, `Mistral-Small-3.2-24B-Instruct-2506`), so it is only reachable this way.
- Registered-provider case: any of the 3 models above, through the built-in `ovhcloud` provider itself — none carries `supportsXHighEffort`, so the same opt-out default applies even without a custom connection.

### Description

`sanitizeReasoningEffortForProvider()` (`open-sse/executors/base/reasoningEffort.ts`) has two related gaps, both rooted in the same cause: no request ever verifies what `reasoning_effort` values the upstream actually accepts.

1. **Missing entries never trigger a downgrade.** `supportsXHighEffort(provider, model)` (`open-sse/config/providerModels.ts`) defaults to `true` (opt-out policy: `!== false`) when no registry entry exists. This is structurally unreachable for user-added custom OpenAI-compatible connections (no registry entry is possible by design) and incomplete for any provider/model the static registry hasn't caught up with yet.
2. **Even when a downgrade _does_ trigger, the target is hardcoded.** `shouldDowngradeXHigh`/`shouldDowngradeMax` both call `writeEffortValue(b, "high", c)` (line 311) unconditionally — assuming `"high"` is always a valid fallback. There's no symmetric handling for a provider whose scale doesn't include `"high"` either.

### Steps to Reproduce

[...]

Issue #10563, opened 2026-08-16
feat(providers): model-catalog-driven native reasoning-effort vocabularies


thomasmaerz:
## Goal

Add optional, lossless per-model native reasoning-effort metadata so OmniRoute can preserve a known model's upstream effort vocabulary and apply explicit model mappings only when necessary.

This is a longer-term consolidation feature. It is **not** required to fix the immediate OpenCode Zen regression, which is deliberately scoped in #10562.

## Motivation

The current sanitizer needs an expanding set of provider/model-specific rules because OmniRoute has a cross-provider vocabulary (`xhigh`) while upstream model contracts differ. A generic fallback is necessary for unknown/passthrough models, but normalizing a known native value can silently change the requested reasoning tier.

Example from #10562:

```text
OpenCode Zen DeepSeek V4 Flash Free
requested max -> current generic rewrite xhigh -> DeepSeek effective high
expected max -> preserved max -> DeepSeek effective max
```

## Proposed Capability Model

Allow registry/catalog metadata to define both lossless native accepted values and explicit aliases/mappings. Illustrative shape only:

```ts
type ReasoningEffortCapabilities = {
  nativeEfforts: string[];
  effortMap?: Record<string, string>;
};

{
  id: "deepseek-v4-flash-free",
  supportsReasoning: true,
  reasoningEffortCapabilities: {
    nativeEfforts: ["low", "high", "max"],
    effortMap: {
      medium: "high",
      xhigh: "high"
    }
  }
}
```

Desired sanitizer policy:

1. Resolve concrete provider/model metadata.
2. Preserve a requested value exactly when it is listed in `nativeEfforts`.
3. Apply an explicit model mapping when the requested value is not native-supported.
4. Retain the present conservative fallback for unknown/passthrough models (for example `max -> xhigh`) to avoid sending unsupported values upstream.
5. Apply identical behavior in target-request and executor sanitizer paths.

The specific names and schema are intentionally open for maintainer design. The essential requirements are lossless native values, model-specific mappings, and a safe unknown-model fallback.

## Use Cases / Existing One-Off Behavior

This is intended to consolidate the simple “what literal effort values does this model accept?” part of the following rules, while preserving transport-specific request shaping separately.

| Route or model family | Current special behavior | Declarative catalog benefit |
| --- | --- | --- |
[...]

Issue #2798, opened 2026-05-27
[BUG]  Models from "OpenCode Free" provider are active in UI but return an empty list in /v1/models


daniij:
### OmniRoute Version

3.8.5

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 24H2

### Node.js Version

24.15.0

### Provider(s) Involved

OpenCode Free

### Model(s) Involved

Any

### Client Tool

Any

### Description

I noticed an issue where models from the OpenCode Free provider do not appear in the /v1/models API endpoint, even though they are active in the user interface.

It seems directly related to the interface showing "0 connections" at the top of the provider panel.

### Steps to Reproduce

1) Open the model management UI.
2) Ensure that the OpenCode Free provider is added and its models (e.g., oc/big-pickle, oc/nemotron-3-super-free, etc.) are active (indicated by the visible eye icon).
3) Note that the top badge displays "0 connections".
4) Send a GET request to http://localhost:20128/v1/models.

### Expected Behavior

The provider should establish connections for the activated models, and the API response should contain the list of these models inside the data array.

### Actual Behavior

The UI shows "0 connections" and the API returns an empty data array:

{
  "object": "list",
  "data": []
}

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

<img width="1337" height="869" alt="Image" src="https://github.com/user-attachments/assets/faddfc57-549f-4407-88e9-e90c48fa8cc2" />

<img width="717" height="218" alt="Image" src="https://github.com/user-attachments/assets/933864ab-4bdf-4675-ab56-e8eaf6846e1e" />

PR #12686, opened 2026-09-04
fix(resilience): honor declared effort vocabulary in reasoning rule gate

woodsonl:
## Summary

- Make the reasoning-routing rule gate honor a model's declared `reasoning_efforts` vocabulary instead of a hardcoded `gpt-5.6-*` regex. A model-scoped or connection-scoped rule forcing `max`/`ultra` is now treated as `supported` when the model's resolved capabilities list that tier; custom OpenAI-compatible providers whose models accept `max` natively (e.g. Merge Gateway `zai/glm-5.3-flash`, accepting `low|high|max`) no longer fail with `Reasoning effort 'max' is not supported by the configured target`.
- The gate's registry lookup now mirrors the dispatch-time sanitizer (`open-sse/executors/base/reasoningEffort.ts`) exactly, so the gate can never approve what dispatch would clamp or reject what dispatch would forward:
  - provider namespace resolution via `getProviderModels` (id→alias, #2798/#3870) and matching the entry's `aliases` list;
  - an empty declared vocabulary (`[]`) falls through, matching the sanitizer's `declaredRanked.length > 0` guard;
  - an operator-declared vocabulary that excludes the forced tier is terminal — the legacy `gpt-5.6` regex can no longer resurrect a tier the override narrowed away;
  - for unregistered providers/models, a declared (synced or operator-overridden) vocabulary listing the tier stays authoritative (#8057 trust-the-upstream); the `gpt-5.6` regex remains the fallback for undeclared models.

## Related Issues

- Related to #11111 (auto-downgrade for unregistered/custom providers) and #10563 (model-catalog-driven native reasoning-effort vocabularies).
- No open issue exists for this exact gate behavior; filed from live infrastructure debugging (Merge Gateway GLM forced-max rule).

## Validation

Change type: provider / routing (per [Contribution Golden Path](../docs/ops/CONTRIBUTION_GOLDEN_PATH.md)).

- [x] Production-code change includes new/updated automated tests in this PR
- [x] Focused tests:
  - `node --import tsx/esm --test tests/unit/reasoning-routing.test.ts` (7/7)
  - `node --import tsx/esm --test tests/unit/reasoning-routing-api.test.ts tests/unit/reasoning-routing-decision-guards.test.ts` (6/6)
  - `node --import tsx/esm --test tests/unit/base-executor-sanitize-effort.test.ts` (56/56)
- [x] `npm run lint` clean on changed files (prettier + eslint)
- [x] E2E verified against a live server: forced-max rule + declared vocabulary reaches the upstream as `reasoning_effort: "max"`; a narrowed vocabulary (`low,high`) still rejects forced max with 400

## Test coverage

New tests cover: declared vocabulary makes forced max supported (before/after), narrowed override keeps forced max unsupported, registry vocabulary outranks a widening override, narrowed override is terminal against the legacy regex (gpt-5.6 regression pin), undeclared registry model + widening override stays supported (#8057), and alias-spelled provider prefixes (`cx/…` vs `codex/…`) resolve to the same registry namespace.


diegosouzapw:
Good fix — mirroring the dispatch-time sanitizer's registry lookup in the routing gate closes a
real gap where the gate could approve a tier dispatch would clamp. Tests pass clean (7/7), and I
confirmed this PR is independently valuable — it passes standalone even without #12730 applied.
Merge-ready as-is.

PR #12697, opened 2026-09-04
fix(combo): treat a pin-only step as implicit connection allowlist

HouMinXi:
## Summary

A combo step with `connectionId` and no `allowedConnectionIds` was only a first-attempt pin. After 502/429, `resolveForcedConnectionForCredentialPool` dropped `forcedConnectionId` and `getProviderCredentials` scanned the whole provider pool.

Live effect: a combo pinned to one Claude Max account could retry on a sibling Pro account that is not in the combo.

This treats a pin-only step as implicit allowlist `[connectionId]`. An explicit allowlist is kept. Header-forced pins (`x-omniroute-connection`, not a combo step) still allow sibling fallback.

`a11930ec4` already stops account rotation inside `handleSingleModel` when a pin is present. That does not populate `allowedConnectionIds` at resolve time, so the credential pool after pin-drop is still the whole provider. These are complementary.

## Files

- `open-sse/services/combo/comboStructure.ts` — pin-only / empty allowlist → `[connectionId]`
- `open-sse/services/combo/fingerprintExpansion.ts` — rewrite composite pin ids to the real row
- `src/lib/combos/steps.ts` — `comboPinAllowlist` / `implicitPinAllowlist`
- `src/shared/validation/schemas/combo.ts` — keep `allowedConnectionIds` on PATCH
- `src/sse/handlers/chat.ts` — apply implicit allowlist on the combo execution path

## Test plan

```
DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=8192 \
  --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts \
  tests/unit/combo-pin-implicit-allowlist.test.ts
```

10/10 pass on `upstream/release/v3.8.51` @ `488f57e9d`.

Covers: pin-only, empty list + pin, explicit allowlist kept, unpinned step unchanged, combo dispatch forwards the list, fingerprint unwrap, Zod parse, header-forced pins do not invent an allowlist.

## Not in this PR

- Combo `executeTarget` split (`feat/combo-execute-target-split`)
- Changing `a11930ec4` rotation gate


HouMinXi:
Technical review (LOCAL, `review-default`, baseline `c3945a724`) on `4817b05da`.

HOLD, 0 CONFIRMED, 7 UNCERTAIN. I am not restarting rounds on the same fingerprints.

Adjudication:

- `implicitPinAllowlist` + explicit `[]` with a pin: **by design, fail-closed.** `steps.ts:219-222` returns `[pin]` whenever the allowlist is missing or empty and a pin is present. `tests/unit/combo-pin-implicit-allowlist.test.ts` asserts `allowedConnectionIds: []` + pin → `["20x-account"]`. The comment on lines 210-212 is about the no-pin case.
- `hasForcedConnection` / `forcedConnectionId` in `handleSingleModelChat`: **same boolean, trimmed.** Empty/whitespace pins become `""` then `length > 0` is false, matching the old `typeof === "string" && trim().length > 0`. `forcedConnectionId || null` vs `?? null` only differs on `""`, which the old path already treated as unforced.
- `checkModelAvailable` no longer has a local `hasForcedConnection`: **true, unused after the pin-allowlist call.** Preflight now goes through `comboPinAllowlist(true, …)`.
- `comboStructure` `toTrimmedString(connectionId)`: **whitespace pins were already broken** (`step.connectionId || null` kept `"   "` and later lookups miss). Trim is the implicit-allowlist input.
- `rewriteAllowlistIds` mapping `|fp|` entries to `realConnectionId`: **needed so a fingerprint-expanded target keeps the pin allowlist on the real row id.** Dedup via `Set`. Not a silent drop of a second fingerprint; expansion already splits those into separate targets.

`chat.ts` own-growth frozen at 2458 after inlining the two `comboPinAllowlist` calls (`_rebaseline_2026_09_04_12697_combo_pin_allowlist`). Helper stays in `src/lib/combos/steps.ts`.

CI remaining on this head: merge-integrity (`changelog.d/fixes/reset-aware-model-family.md` missing `- ` on the release tip), API Route Typecheck (`glm.ts` TS2554 on tip; only #12711 drops the extra positional), FQG mutation + open-sse-typecheck. Same fingerprint as #12732 and the other open PRs that do not touch those files. Eslint on this PR is green.

Contributor only — not merging.

HouMinXi:

HouMinXi:

`698791076` → `5ac5072c8`. range-diff 3 commits `=`. GPG G. No file overlap with the three landed commits.

HouMinXi:

- `5ac5072c8e3` → `b6e11aec26c`
- range-diff identity (eq=3); all commits GPG `G`
- `3` unique commit(s) ahead, 0 behind
- `--force-with-lease` to `HouMinXi:fix/combo-pin-implicit-allowlist`

PR #12707, opened 2026-09-04
feat(db): fail-closed previous_response_id continuation for redacted video turns (#12150 P2b)

diegosouzapw:
## What

Closes the `previous_response_id` continuation surface of #12150 (P2b — #12430 items 1–2).

Since #12528, a request whose video parts the video-bridge guardrail rewrote has its **client-request snapshot** (`clientRawRequest.body`, persisted in the call-log artifact) redacted: every transcript cue becomes `[redacted-video-transcript]` before storage. But `resolvePreviousResponseState` (`src/lib/db/responsesContinuationStore.ts`) reads that snapshot back to reconstruct history for a later `previous_response_id` continuation — so it would forward the **placeholder** upstream as if it were the client's real prior turn.

This makes that continuation **fail closed** for any turn whose video content was removed.

## How

- **Migration `173_call_logs_video_content_removed.sql`** — new `call_logs.video_content_removed INTEGER NOT NULL DEFAULT 0`.
- **Write path** (`callLogs.ts` INSERT + `chatCore/attemptLogging.ts` ctx + `chatCore.ts` binding) — set the marker to `1` from the existing `videoBridgeObserved` signal (the same boolean that already gates the #12528 snapshot redaction). `false`/absent for every non-video request.
- **`resolvePreviousResponseState`** — reads the marker and returns `null` when it is `1`, so the client resends full history, exactly like a real `previous_response_not_found`.

Default `0` ⇒ every non-video request and continuation is byte-identical to before.

## Tests (TDD — HR#18)

RED→GREEN for each layer (no VPS needed — fully unit-provable):
- `responses-continuation-store.test.ts` — store refuses a marked row, still resolves a normal one.
- `save-call-log-persistence.test.ts` — column exists; `saveCallLog` persists `videoContentRemoved=true` as `1`, defaults `0`.
- `video-bridge-log-redaction.test.ts` — the real `persistAttemptLogs` writes the marker to the row from `ctx.videoContentRemoved` (and defaults `0`).

`typecheck:core` clean; `check:migration-numbering` OK (173, no dup); prettier clean.

Design: `_tasks/superpowers/specs/2026-09-01-video-transcript-retention-design.md` (Surface 2). Refs #12150, #12430.


diegosouzapw:
⚠️ **base-red inherited** — the red CI on this PR is NOT from this change.

The `release/v3.8.51` tip advanced 15+ commits after the base-green issue #12581 auto-closed, and the continuous release-green runs keep getting cancelled by concurrent merges, so a fresh base regression is live but un-flagged. Discriminated:

- **Unit fast-path** — `huggingchat-stream-error-boundary` and `model-sync-route` fail on a clean tip worktree **without** this change; `sanitizeErrorMessage ReDoS guard` is a timing flake under CI load. None touch this PR's files.
- **API Route Typecheck** (`apiTypecheckErrors=290` vs baseline 289) — the +1 is `open-sse/executors/glm.ts TS2554`, a file this PR does not touch (present at this branch's base point).
- **Merge integrity** — `check:changelog-integrity` reports "1 invalid changelog fragment" not added by this PR.

This PR's own change is TDD-green locally (touched test files + sibling call-log cluster + `typecheck:core` + `check:migration-numbering`, all pass). Not fixing inherited base defects in a feature branch (base-green protocol). Re-run once the tip drains.

PR #12711, opened 2026-09-04
fix(dashboard): batch delete no longer toasts failure after success

HouMinXi:
## Bug

On the provider detail page, bulk-deleting API keys (moonshot-native in this case) shows "deleted N connections" and then immediately "network error during batch delete". The keys are actually gone. The failure toast is a lie.

Fixes: dashboard bulk-delete success+failure toast pair.

## Root Cause

`ConfirmModal` binds `onConfirm` to `<Button onClick={onConfirm}>`. A native click passes a MouseEvent.

`handleBatchDeleteConfirm(onAfter?)` treated any truthy first argument as a post-delete callback. After DELETE 200 it toasted success, then `await onAfter()` invoked the MouseEvent object, which throws, and the catch toasted `batchDeleteNetworkError`.

API side is fine: `DELETE /api/providers` with `{ ids }` returns 200 and deletes the rows. The second toast is UI-only.

Measured 2026-09-04 against `handleBatchDeleteConfirm` + ConfirmModal wiring. Reproduction: pass a `{ type: "click" }` stand-in as the first argument after a 200; success fires, then the network-error toast.

## Fix

- Run `onAfter` only when `typeof onAfter === "function"`.
- ConfirmModal confirm click: `onClick={() => void onConfirm()}` so the event never reaches the handler.

## How to Verify

1. Open a provider with several API keys (moonshot-native / any openai-compatible node).
2. Select more than one key. Confirm bulk delete.
3. Expect one success toast. No follow-up failure toast. Keys gone from the list.

## Test Plan

- [x] `tests/unit/ui/batch-delete-click-event.test.tsx` — click-event stand-in after 200 does not toast error; a real `onAfter` still runs
- [x] Injection: revert `typeof` check → the click-event case goes red; restore → green
- [x] `tests/unit/ui/confirm-delete-connection-7361.test.tsx` still 3/3
- [x] `tests/unit/providers-route-managed-catalog.test.ts` still 6/6 (includes batch DELETE)

## Risk

Low. ConfirmModal still calls `onConfirm()` with no args, which matches its type (`() => void | Promise<void>`). `onAfter` is unused in production callers today; the typeof guard is the belt, the ConfirmModal wrap is the suspenders.

## Not in this PR

No chunking of DELETE past the existing 100-id API cap. Activate/deactivate already chunk; delete does not. Separate if you bulk-delete >100 at once.

## Also in this commit (base hygiene that blocked CI)

- Prefix unused `useProviderConnections` arg as `_isSearchProvider` (eslint `@typescript-eslint/no-unused-vars` on the file this PR already touches).
- `changelog.d/fixes/reset-aware-model-family.md` now starts with `- ` (fragment gate).
- Register `tests/unit/reset-aware-request-scope-12600.test.ts` in `stryker.conf.json` `tap.testFiles` (mutation-test-coverage --strict).
- `open-sse/executors/glm.ts`: drop the extra 16th arg to `createSSETransformStreamWithLogger` (TS2554; the 65536 never reached TransformStream).

Not in this PR: HuggingChat/stream public-error unit failures on `release/v3.8.51`, and local-only forge `daemon_state` (`.code-forge/` is gitignored).

### DIFF
--- a/changelog.d/fixes/12711-batch-delete-click-event.md
+++ b/changelog.d/fixes/12711-batch-delete-click-event.md
@@ -1,0 +1,2 @@
+- **fix(dashboard):** batch-deleting provider keys no longer toasts failure after a successful delete when the confirm button's click event is forwarded as `onAfter` ([#12711](https://github.com/diegosouzapw/OmniRoute/pull/12711))
+- **fix(glm):** drop the extra 16th argument to `createSSETransformStreamWithLogger` that TypeScript rejected (TS2554) and that never reached the TransformStream
--- a/changelog.d/fixes/reset-aware-model-family.md
+++ b/changelog.d/fixes/reset-aware-model-family.md
@@ -1,1 +1,1 @@
-Keep Antigravity Gemini usable when the same connection's Claude weekly quota is empty; generic quota cache stays per-connection for every other provider.
+- Keep Antigravity Gemini usable when the same connection's Claude weekly quota is empty; generic quota cache stays per-connection for every other provider.
--- a/open-sse/executors/glm.ts
+++ b/open-sse/executors/glm.ts
@@ -223,8 +223,8 @@ export function translateSseResponse(
   suppressThinkClose: boolean = false
 ): Response {
   if (!response.body) return response;
-  // GLM is a high-throughput provider — use a larger stream buffer (64KB) to
-  // keep provider → client pacing ahead of the model's token emission rate.
+  // Helper pins TransformStream highWaterMark at 16KB; a 16th positional
+  // (65536) was a TS2554 and never reached createSSEStream.
   const transform = createSSETransformStreamWithLogger(
     FORMATS.CLAUDE,
     FORMATS.OPENAI,
@@ -238,10 +238,7 @@ export function translateSseResponse(
     null,
     null,
     false,
-    suppressThinkClose,
-    undefined,
-    undefined,
-    65536
+    suppressThinkClose
   );
   const headers = cloneHeaders(response.headers);
   headers.set("content-type", "text/event-stream");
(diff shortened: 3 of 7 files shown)


HouMinXi:
Technical review (LOCAL, `review-default`, baseline `c3945a724`) on `6aac03b83`.

HOLD, 0 CONFIRMED, 2 UNCERTAIN. Not restarting rounds.

Adjudication:

- `typeof onAfter === "function"` in `handleBatchDeleteConfirm`: **this is the bug**. ConfirmModal used to bind `onClick={onConfirm}`, so React passed a MouseEvent into `onAfter`. The truthiness check then called `event()`, which threw and the catch toasted `batchDeleteNetworkError` after a successful DELETE. `tests/unit/ui/batch-delete-click-event.test.tsx` covers that path. Other `ConfirmModal` callers either wrap with `() => void handler()` or pass a zero-arg confirm; only this handler accepted an optional callback and treated a truthy argument as callable.
- `onClick={() => void onConfirm()}` on ConfirmModal: **not cosmetic.** `ConfirmModalProps.onConfirm` is `() => void | Promise<void>` — no event. The wrapper stops React from injecting the MouseEvent. Together with the typeof guard, the click event never reaches the batch-delete handler.

`glm.ts` dropping the extra positional is the TS2554 fix the other open PRs inherit from this branch. Unused `KeyHealth` / `translateJsonResponse` were already on the tip.

CI remaining: merge-integrity (`cli-tunnel` skill dry-run generate + tip changelog fragment without `- `), unit shards. Same fingerprint as #12732. Typecheck / FQG on this PR are green.

Contributor only — not merging.

HouMinXi:

HouMinXi:

`cb6a70c37` → `45c645244`. range-diff 2 commits `=`. GPG G. No file overlap with the three landed commits.

HouMinXi:

- `45c645244b3` → `b3503bc4194`
- `--force-with-lease` to `HouMinXi:fix/batch-delete-click-event`

PR #12719, opened 2026-09-04
fix(opencode): synthesize x-opencode-session when client sends none — 2026-09-06 upstream enforcement

Moseyuh333:
## Summary

**Rescoped after review (commit `46eb14bd4`).** Under the default configuration, `applyCliDefaults()` (step 4, #10571) already backfills `x-opencode-session` with `||=`, so the default path is covered — thanks for the careful audit. The reproducible gap this PR actually closes is the **opt-out path**: with `OPENCODE_SYNTHESIZE_CLI_HEADERS=false`, step 4 never runs and, before this patch, step 3 did not synthesize either — an OpencodeExecutor request whose client sent no session-ish header went outbound **without `x-opencode-session` at all**.

- **What**: `forwardOpencodeClientHeaders()` step 3 (OpencodeExecutor path only) now synthesizes `x-opencode-session` when the client provides none of `x-opencode-session` / `x-session-affinity` / `x-session-id`: a conversation-stable fingerprint via `generateSessionId(sessionBody)` (model, system prompt, first user message, tools), falling back to a random UUID when no body is available. +8/-3; explicit client sessions still win and forward untouched.
- **Why merge**: (1) opencode.ai errs on requests missing `x-opencode-session` since 2026-09-06 — the opt-out path trips exactly that check. (2) The opt-out is a documented escape hatch (#10571 flipped the default on; `OPENCODE_SYNTHESIZE_CLI_HEADERS=false` lets operators who don't want fabricated CLI identity still customize env defaults), and those operators still need a session header for the enforcement and upstream prompt caching. (3) It future-proofs step 3 for any path where `cliDefaults` is absent.
- **Scope**: step-3 only. CLI-identity synthesis (#5997/#10571) unchanged; DefaultExecutor pass-through untouched; the `body` plumbing (`buildHeaders(..., body)`, `sessionBody`) and the Muse responses UUID guard already exist on the base — this closes the last missing branch.

## Related Issues

- Related to #5997 (deterministic CLI-identity synthesis — same `generateSessionId` fingerprint mechanism)
- Related to #10571 (CLI synthesis on by default; its opt-out is the path this patch makes safe)
- No upstream issue for the 2026-09-06 opencode.ai header enforcement; this pre-empts the deadline.

## Validation

- [x] Change type: provider / sse
- [ ] `npm run lint` — **not runnable in this contributor environment** (local `node_modules/eslint` fails to load, pre-existing, unrelated to this diff; CI runs the real lint gate)

### Isolation proof — current tip `152d95108`, unpatched vs patched

```shell
# Unpatched tip, new test files only:
DISABLE_SQLITE_AUTO_BACKUP=true node --import tsx/esm \
  --test tests/unit/runtime/opencode-session-headers.audit.test.ts
# > tests 11 | pass 10 | fail 1
#   ✖ opt-out (OPENCODE_SYNTHESIZE_CLI_HEADERS=false): step 3 alone still fills x-opencode-session

# Same file set with this PR's opencodeHeaders.ts applied:
#   audit + e2e: > tests 14 | pass 14 | fail 0
```

- Also verified on the branch's own base `c3945a724` (identical `opencodeHeaders.ts` to the tip): unpatched 10/11 — only the new test red — patched 11/11.
- `tests/unit/opencode-executor.test.ts` + `tests/unit/opencode-cli-headers-synthesis-5997.test.ts` on the branch: 66/66 pass.
- `npx prettier --check tests/unit/runtime/opencode-session-headers.audit.test.ts`: clean.
- Reconciliation: all four files this PR touches are identical between the branch base and the current tip — the diff applies cleanly.

## Tests Added Or Updated

- `tests/unit/runtime/opencode-session-headers.audit.test.ts` — new file, 11 tests, including the opt-out isolation test that **fails on the unpatched tip (verified) and passes with the patch**.
- `tests/unit/runtime/opencode-session-e2e.test.ts` — new file, 3 tests (echo-server e2e through `OpencodeExecutor.execute()`).
- `tests/unit/opencode-executor.test.ts` — 3 subtests updated to pin the new behavior (synthesize when the client sends no matching keys).

## Coverage Notes

- `open-sse/utils/opencodeHeaders.ts` step-3 behavior is covered by the audit file; the outbound wiring by the e2e file; the executor-level path by the updated subtests. No coverage movement expected.

## Reviewer Notes

- Direct answer to the review: correct — in the default configuration step 4 already backfills the header, and none of the original tests could observe the step-3 gap. Took option (a): the description is now scoped to the opt-out edge case, and the new test
  `opt-out (OPENCODE_SYNTHESIZE_CLI_HEADERS=false): step 3 alone still fills x-opencode-session`
  isolates it — red on the unpatched tip (10/11), green with the patch (11/11; 14/14 with e2e). Happy to re-review.
- No feature flag, no migration.

### DIFF
--- a/open-sse/utils/opencodeHeaders.ts
+++ b/open-sse/utils/opencodeHeaders.ts
@@ -99,10 +99,18 @@ export function forwardOpencodeClientHeaders(
       findHeader(clientHeaders, "x-session-affinity") || findHeader(clientHeaders, "x-session-id");
     if (sessionAffinity) {
       headers["x-opencode-session"] = sessionAffinity;
-
-      if (!headers["x-opencode-request"]) {
-        headers["x-opencode-request"] = randomUUID();
-      }
+    } else {
+      // 2026-09-06 upstream enforcement: opencode.ai errors on requests missing
+      // x-opencode-session. A client that sends no session-ish header at all
+      // gets a conversation-stable fingerprint (model, system, first user
+      // message, tools) so consecutive agent turns share one session and
+      // upstream prompt caching hits; a body-less request falls back to a
+      // random UUID (same behavior as the #5997 CLI-identity synth).
+      headers["x-opencode-session"] =
+        generateSessionId(options.sessionBody ?? null) || randomUUID();
+    }
+    if (!headers["x-opencode-request"]) {
+      headers["x-opencode-request"] = randomUUID();
     }
   }
 
@@ -141,6 +149,5 @@ function applyCliDefaults(
   headers["x-opencode-client"] ||= cliDefaults.client;
   headers["x-opencode-project"] ||= cliDefaults.project;
   headers["x-opencode-request"] ||= randomUUID();
-  headers["x-opencode-session"] ||=
-    generateSessionId(sessionBody ?? null) || randomUUID();
+  headers["x-opencode-session"] ||= generateSessionId(sessionBody ?? null) || randomUUID();
 }
--- a/tests/unit/opencode-executor.test.ts
+++ b/tests/unit/opencode-executor.test.ts
@@ -400,20 +400,24 @@ describe("OpencodeExecutor", () => {
       assert.equal(headers["x-opencode-client"], undefined);
     });
 
-    it("does not add x-opencode-* when clientHeaders has no matching keys", () => {
-      const headers = zenExecutor.buildHeaders({ apiKey: "test-key" }, true, {
+    it("synthesizes session+request when clientHeaders has no matching keys (09/06)", () => {
+      // 2026-09-06 opencode.ai enforcement: requests missing x-opencode-session
+      // may error — an empty header set now gets a synthesized session.
+      const headers = zenExecutor.buildHeaders({ apiKey: "test-key-0001" }, true, {
         "some-other-header": "val",
       });
-      assert.equal(headers["x-opencode-session"], undefined);
-      assert.equal(headers["x-opencode-request"], undefined);
+      assert.ok(headers["x-opencode-session"], "session must be synthesized");
+      assert.match(headers["x-opencode-session"], /^[0-9a-f-]{36}$/i);
+      assert.ok(headers["x-opencode-request"], "request id must be synthesized");
     });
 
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Thanks for the thorough test coverage here — 13 well-written tests. But I need to flag
something before this can merge: I dropped your own new test files onto the unpatched
`release/v3.8.51` tip (no code changes from this PR applied) and all 13 passed. That's because
`OpencodeExecutor.buildHeaders()` already runs `applyCliDefaults()` unconditionally (step 4,
from #10571) whenever CLI-identity synthesis is enabled — which it is by default — and that
step already backfills `x-opencode-session` with `||=` regardless of what step 3 (the code this
PR touches) does. So under the default configuration, the gap you're describing doesn't
currently exist on the tip.

The one place your patch *does* change behavior is when an operator sets
`OPENCODE_SYNTHESIZE_CLI_HEADERS=false` (disabling CLI-identity synthesis) — in that case step 4
never runs, and your fix to step 3 would be the only thing filling the gap. None of your 13
tests isolate that specific scenario though. Could you either (a) rescope the PR description to
that opt-out edge case specifically and add a test for it, or (b) let us know if I'm missing
something about how this reaches the default path? Happy to re-review either way.

Moseyuh333:
Took option (a) — thanks for the precise audit; you're right that step 4's `||=` backfill covers the default path and none of the original tests could observe the step-3 gap.

Changes since your review (commit 46eb14bd4):
- PR description rescoped to the opt-out edge case (`OPENCODE_SYNTHESIZE_CLI_HEADERS=false`, where step 4 never runs).
- Added an isolation test: `opt-out (OPENCODE_SYNTHESIZE_CLI_HEADERS=false): step 3 alone still fills x-opencode-session` — it also asserts `x-opencode-client`/`x-opencode-project` stay unset, proving it exercised step 3 only.

Isolation proof, run on the current tip 152d95108 with the new test files dropped on unpatched vs patched:
- unpatched: 10/11 — only the new test red
- patched: 14/14 (audit 11 + e2e 3)

Ready for re-review.

PR #12721, opened 2026-09-04
fix(cursor): recover Kimi tool calls emitted as history narration

patrykkopycinski:
## Problem

Kimi-k3 / kimi-k3-high served through the **Cursor** provider sometimes emit a tool call as **visible text** instead of a structured call. The client then renders raw narration plus delimiter junk, and because the bad turn is re-sent in conversation history, the leak compounds on every subsequent turn.

Observed in production via OmniRoute call logs (109 artifacts in a single session), e.g.:

```
…Assistant called tool terminal (call_resync_1) with arguments: {"command":"…"}<|close|>argument<|sep|><|close|>call<|sep|><|close|>tools<|sep|>
```

with `finish_reason: "stop"` and **no `tool_calls` field** on the message.

## Root cause

Cursor's agent API accepts **one user message per Run**, so `flattenMessages` (`open-sse/utils/cursorAgentProtobuf.ts`) serializes prior `assistant.tool_calls` into plain text for the model to read as context:

```
Assistant called tool <name> (<id>) with arguments: <json>
```

Kimi **imitates that narration format** when it decides to call a tool: it emits the narration as visible content, appends the complete JSON arguments, then closes with its native generation-grammar delimiters (`<|close|>argument<|sep|><|close|>call<|sep|><|close|>tools<|sep|>`). Cursor's protobuf backend passes the whole thing through verbatim. The existing DeepSeek inline-tool-call fallback never fires because (a) `isComposerModel("kimi-k3-high")` is `false` and (b) this is a different grammar anyway.

## Fix

New `open-sse/utils/kimiToolCallNarration.ts`:

- detect the `Assistant called tool <name> (<id>) with arguments:` narration line
- balanced-scan the JSON arguments (honors strings/escapes/nesting); bail if truncated or not valid JSON
- strip the native delimiter-grammar tail wherever it appears
- reassemble a structured OpenAI `tool_calls` entry and return the cleaned visible content

Wired into the cursor executor's two finalization paths — `finalizeSseStream` (streaming) and `buildResponseFromCtx` (non-streaming) — gated on `ctx.toolCalls.length === 0` so it **never fires on a healthy turn**.

## Validation

- 9 new unit tests in `tests/unit/kimi-tool-call-narration.test.ts` — all pass.
- Ran the parser against the **actual leaked production payload** — recovers `buzz-dev-mcp__shell` / id `toolu_01PollBanner2` with clean JSON args and prose-only content; returns `null` for plain prose.
- Existing cursor / composer / translator suites still green (54/54).
- `tsc -p tsconfig.json --noEmit` introduces no new errors in the touched files.


patrykkopycinski:
Rebasing onto release/v3.8.51 (was mis-based on stale main); will reopen.

Issue #12729, opened 2026-09-04
fix(api): 3.8.50 leaks internal `_omniroute*` parameters into upstream request bodies (400 from NVIDIA)


main-sounds-audio:
## Summary

Since upgrading to 3.8.50, OmniRoute forwards its own internal fields in the
upstream request body. Providers that validate their input reject the request:

```
[400]: Validation: Unsupported parameter(s): `_omnirouteSkipContextRelay`, `_omnirouteInternalRequest`
[400]: property '_omnirouteInternalRequest' is unsupported
```

## Evidence that it is the upgrade

From OmniRoute's own `call_logs` table on this instance. The 3.8.50 server
started at **2026-09-03 19:27 UTC**; the first such 400 is **19:40:27 UTC**, and
there are **zero** occurrences anywhere before the upgrade.

| Date (UTC) | 400s | First | Last |
|---|---|---|---|
| 2026-09-03 (post-upgrade only) | 79 | 19:40:27 | 21:28:01 |
| 2026-09-04 | 425 | 04:11:50 | 12:26:07 |

**504 in total**, still occurring.

## Which providers

| Provider | 400s |
|---|---|
| nvidia | 501 |
| groq | 3 |

Providers that ignore unknown JSON fields (gemini, aion, opencode) are
unaffected, which is why this reads as an NVIDIA problem at first glance. It is
not — it is the request body.

## Impact

Not an outage: a combo simply fails over to the next target. The cost is a
wasted upstream call and its latency on every affected step. On this instance
that is roughly a fifth of one combo's steps.

## Environment

- OmniRoute 3.8.50 (upgraded from 3.8.49, where this does not occur)
- macOS, node v24.18.0, installed via `npm install -g omniroute`
- Requests are `POST /v1/chat/completions` with `stream: true`, routed through a
  `priority` combo
- No proxy, no custom middleware

## Expected

Internal routing fields should be stripped from the payload before it is sent
upstream, or namespaced somewhere that is not the provider request body.

## Note for whoever picks this up

`_omnirouteSkipContextRelay` suggests the context-relay path. This instance has
`OMNIROUTE_EMERGENCY_FALLBACK=false` set as a feature-flag override; every other
flag is at its default. There is no feature flag that turns the leak off — the
flag list has no context-relay toggle.

diegosouzapw:
[...]

PR #12735, opened 2026-09-04
fix(sse): strip internal markers from the dario and 9router request bodies (#12729)

gonisulaimann:
Part of #12729 — the shared-half fix lands in #13355.

Context-relay and universal-handoff summary calls build their request bodies with internal `_omnirouteSkipContextRelay` / `_omnirouteInternalRequest` flags so the routing layer can recognize them and skip relay injection. Those flags were read before dispatch but never removed from the payload sent upstream, so since 3.8.50 every handoff summary call carried them to the provider. Strict OpenAI-compatible providers (NVIDIA NIM, Groq) reject unknown parameters with a 400 `Unsupported parameter(s)`, which wasted one upstream round-trip per combo step — 504 occurrences in the reporter's call_logs, all starting right after the 3.8.50 upgrade.

Per your comment, the shared-function change stays with #13355, so this PR is now only the per-executor half. `dario` and `9router` both override `transformRequest()` without calling the base implementation, so neither of the two shared strips runs for their bodies — not the one inside `BaseExecutor.transformRequest()`, and not the one before dispatch in `BaseExecutor.execute()`. Whatever the routing layer leaves on the request goes out verbatim on those two paths.

Two of the four executors I listed originally turned out not to need the call, so I dropped them rather than leave dead code:

- `glm` — `GlmExecutor.transformRequest()` calls `super.transformRequest()`, which is exactly where the shared strip lives, so its body is already covered.
- `gitlab` — its `transformRequest()` rebuilds the payload field by field (`current_file`, `intent`, `project_path`, …), so no extra top-level key can reach the wire at all.

That leaves the two calls in this diff as the ones carrying the fix. With #13355's prefix match in the helper, both paths end up clean.

Tests: `tests/unit/strip-omniroute-internal-params-12729.test.ts` drives each of the two executors through `execute()` with a stubbed `fetch` and asserts the serialized body went through the marker strip, with the rest of the request untouched. Both fail when the calls are reverted. The assertion probes with `_omnirouteResponsesStore` rather than a handoff flag deliberately — on this branch the helper doesn't know the handoff flags yet, so asserting on them directly would only pass once #13355 lands. What this PR owns is the wiring. The ninerouter executor suites stay green (45 tests); `typecheck:core`, `npm run lint` and prettier pass.


gonisulaimann:
Pushed a small follow-up commit: the countTokens strip I'd added for defense-in-depth pushed base.ts over its frozen file-size cap, so I dropped that one call. The leak itself is covered by the shared strip in stripInternalBodyFields plus the executors that serialize their own bodies, so nothing else changed; the strip test suite still passes.

diegosouzapw:
Confirmed the leak is real on the current tip — `stripInternalBodyFields()` only knows 5
keys, none of the 3 handoff flags. Your fix at the shared chokepoint and, importantly, the
defense-in-depth calls you added inside `dario`/`gitlab`/`glm`/`ninerouter` (which bypass the
base dispatch entirely) are both correct — I verified those 4 executors do their own
fetch/serialize path. There's direct overlap with #13355 at the `cliFingerprints.ts`
chokepoint (it does the same fix but prefix-based, which is a superset). We'd like to land
#13355 for that shared function and keep your 4-executor calls from this PR — happy to
coordinate the rebase so you keep credit for that part.

gonisulaimann:
Rebased onto the current tip and re-scoped as you suggested.

Dropped the `stripInternalBodyFields()` change and kept the per-executor calls. Re-verifying them turned up two of the four weren't doing anything, so I removed those instead of leaving dead calls:

- `glm` already reaches the shared strip: `GlmExecutor.transformRequest()` calls `super.transformRequest()`, which is where the strip lives.
- `gitlab` can't leak at all: its `transformRequest()` rebuilds the payload field by field, so no extra top-level key survives.

`dario` and `9router` are the two that really leak — neither calls super, and neither is on the strip path inside `BaseExecutor.execute()`. The tests now drive each of them through `execute()` and fail if the call is removed (checked by reverting it).

One caveat worth stating: the test probes with `_omnirouteResponsesStore` rather than a handoff flag on purpose. On this branch the helper still only knows its five explicit keys, so a handoff-flag assertion would only pass once #13355 lands — the wiring is what this PR owns. Production diff is four lines, plus the test and the changelog fragment.

diegosouzapw:
Thanks @gonisulaimann — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

Issue #7447, opened 2026-07-16
fix(providers): Kimi/Moonshot API key format rejected during connection verification (needs-info — confirm kimi-coding vs Moonshot Platform)


contione:
### OmniRoute Version

3.8.48

### Installation Method

Built from source

### Operating System

Windows

### OS Version

Window 11

### Provider(s) Involved

Kim

### Description

 Kim API key verification failed

### Steps to Reproduce

1. go to Providers
2. add kim connection
3. check api key

### Expected Behavior

<img width="881" height="943" alt="Image" src="https://github.com/user-attachments/assets/6cee0e48-978d-4c81-9060-48b933a5a4c2" />

### Actual Behavior

Verification should pass.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

contione:
API KEY Format eg:
sk-2r0xKByHEG3pdmktqqzUByguz0jLG7gmgm9wyDozjjw7sb6L

diegosouzapw:
Thanks for the report. I dug into the Kimi/Moonshot key-validation code and it looks correct
for the international `api.moonshot.ai` endpoint (which is the one OmniRoute currently
supports) — so I suspect this might be a "wrong platform" key rather than a code bug: Moonshot
runs two completely separate platforms with independent accounts/keys — international
`platform.kimi.ai` (`api.moonshot.ai`) and the domestic Chinese platform (`platform.kimi.com` /
`api.moonshot.cn`). A key generated on one does not work against the other's endpoint, and
that mismatch typically shows up as exactly this generic "Invalid API key" message.

Could you confirm:
1. Which URL did you sign up / generate the key on — `platform.kimi.ai` (international) or
   the domestic `platform.kimi.com`/`moonshot.cn` platform?
2. If you have access to your browser's Network tab, the response body of the failed `Check`
   request would help confirm the exact upstream error.

If it turns out you're on the domestic platform, we don't currently have a `.cn` endpoint
registered — that would be a feature request rather than a bug in the existing key check.

contione:
Yes,  CN-region key.

diegosouzapw:
Thanks for reporting! Fixed in `release/v3.8.49` via #8500 (already merged into the active development branch — feel free to pull and test it now). It will ship in the next release (v3.8.49).

nixmomo:
[...]

PR #12736, opened 2026-09-04
fix(dashboard): expose Modal Base URL field in connection modals (#12704)

### LINKED ISSUE #12704: fix(providers): Modal provider connection form lacks Base URL input, causing connection validation failure
### OmniRoute Version

3.8.51 (and 3.8.49)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04 LTS (ARM64)

### Node.js Version

26.5.0

### Provider(s) Involved

Modal (`modal`)

### Model(s) Involved

Any OpenAI-compatible serverless LLM deployed on Modal

### Client Tool

OmniRoute Web Dashboard (Add Provider / Connection Modal)

### Description

When attempting to connect the **Modal** provider in the OmniRoute Web Dashboard, the connection form only presents input fields for **Token ID** and **Token Secret**.

Unlike standard centralized API providers (e.g. OpenAI, Anthropic) that have a fixed global host, Modal is a bring-your-own-deployment platform where each user runs their model on a unique endpoint URL (e.g., `https://<workspace>--<app>.modal.run/v1` or `https://<workspace>--<endpoint>.modal.direct/v1`).

In `src/lib/providers/validation.ts`, the backend validation handler explicitly checks for `providerSpecificData.baseUrl`:

```ts
modal: ({ apiKey, providerSpecificData }: any) => {
  const baseUrl = (providerSpecificData?.baseUrl || "").trim();
  if (!baseUrl) {
    return {
      valid: false,
      error:
        "Modal requires a Base URL pointing to your OpenAI-compatible Modal app " +
        "(e.g. https://<workspace>--<app>.modal.run/v1). " +
        'Fill in the "Base URL override" field.',
    };
  }
  return validateOpenAILikeProvider({ ... });
}
```

However, because the frontend connection modal does not expose the **Base URL / Deployment URL** input field for the `modal` provider, users have no way to fill in this required field. Consequently, clicking "Check Key" or attempting to test/save the connection immediately fails with a missing or invalid URL error, making it impossible to add or use Modal-hosted models.

### Steps to Reproduce

1. Open OmniRoute Dashboard -> **Providers** -> **Add Connection** / **Add Provider**.
2. Select **Modal**.
3. Observe that the form only requests **Token ID** and **Token Secret**.
4. Enter valid credentials and click **Test / Check Key**.
5. Validation fails with:
   `"Modal requires a Base URL pointing to your OpenAI-compatible Modal app (e.g. https://<workspace>--<app>.modal.run/v1). Fill in the 'Base URL override' field."` (or an invalid URL error).

### Expected Behavior

The Modal add/edit connection dialog should expose an input field for the **Deployment Base URL** (e.g. `https://<workspace>--<app>.modal.run/v1`), which gets stored in `providerSpecificData.baseUrl`, allowing the health check probe to reach `/v1/models` and route chat completions to `/v1/chat/completions`.

### Actual Behavior

There is no Base URL field in the UI for Modal, causing the connection test to fail every time.

### Additional Context

Related to closed feature request #10132 and bug report #9102. Exposing the Base URL field in `AddApiKeyModal` / `providerPageHelpers` for provider `modal` will resolve the issue.

gonisulaimann:
Fixes #12704.

Modal is bring-your-own-deploy — every user runs their model on a unique endpoint (`https://<workspace>--<app>.modal.run/v1`), so there's no fixed host to preset. The server-side validator (`src/lib/providers/validation.ts`) already requires `providerSpecificData.baseUrl` for modal, but the add/edit connection modals only rendered Token ID / Token Secret, so there was no way to fill the required field and every Modal connection failed validation.

This adds `modal` to `CONFIGURABLE_BASE_URL_PROVIDERS`, the same mechanism kimi/moonshot use (#7447). Both modals key their Base URL field off that one predicate (`AddApiKeyModal.tsx:1104`, `EditConnectionModal.tsx:199`), and Save stays disabled while the field is empty for a provider with no default URL (`AddApiKeyModal.tsx:1160`), so a Modal connection now renders the field, requires it, and validates it on save.

Rebased onto the current tip. #13120 added `agnes` to the same set and to the same placeholder function while this was open; both changes are additive, so the merge keeps its entry and mine.

One thing worth flagging: adding a case pushed `getProviderBaseUrlPlaceholder` over the complexity cap, so it now reads from a record lookup plus a small set for the ids whose placeholder is their configured default URL. I compared it against the switch it replaces for every id that switch handled, including the ones #13120 added, plus null/undefined/unknown, and `modal` is the only value that changes.

Tests: `tests/unit/modal-base-url-field-12704.test.ts` asserts modal is base-URL configurable, the placeholder shape, and that the server-side validator still rejects a baseUrl-less modal connection, so the UI and the backend can't drift apart. Two of the three fail without the fix. All 139 tests in the suites that import `providerPageHelpers` stay green; typecheck, lint and prettier pass.


gonisulaimann:
Follow-up commit: the new modal case pushed getProviderBaseUrlPlaceholder from 15 to 16 against the complexity cap, so I moved the literal placeholder examples into a record lookup keyed by provider id instead of a switch. Same values, no behavior change; the modal and kimi suites still pass.

diegosouzapw:
Great fix — clean root-cause diagnosis, and I really appreciate the follow-up commit where you
caught your own complexity-cap regression and refactored the switch into a lookup table instead
of just suppressing the warning. The test is solid too (exercises the real validator, not a
mock). Only thing needed before merge is a rebase — #13120 landed in the same file since your
branch point, so there's a small textual conflict to resolve (both additive, should be quick).

gonisulaimann:
Rebased onto the current tip. The conflict was `agnes` from #13120 landing in the same set and the same placeholder function — both additive, so both entries are in.

Adding the modal case also pushed `getProviderBaseUrlPlaceholder` over the complexity cap, so it reads from a record lookup now instead of a switch. I compared the two for every id the switch handled, including the ones #13120 added, plus null/undefined/unknown, and `modal` is the only value that changes.

Same three assertions as before: modal is base-URL configurable, the placeholder shows the app URL shape, and the server-side validator still requires the URL. Two of the three fail without the change. All 139 tests across the suites importing `providerPageHelpers` are green.

diegosouzapw:
Thanks @gonisulaimann — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #12672, opened 2026-09-04
[defer] feat(antigravity): support Gemini 3.8 Flash, live catalog sync, and family quota pools

TheDemonTuan:
⚠️ base-red inherited: #12581

## Summary

This PR upgrades Antigravity & AGY support to handle Gemini 3.8 Flash, enables real dynamic live-catalog synchronization for future models, and implements model family quota pools (Gemini Models, Claude & GPT Models) with 5h and weekly windows.

### 1. Gemini 3.8 Flash Support & Request Semantics
- Adds `gemini-3.8-flash-high`, `gemini-3.8-flash-medium`, `gemini-3.8-flash-low` to Antigravity and AGY public models.
- Maps bare `gemini-3.8-flash` -> `gemini-3.8-flash-medium` without forcing `-tiered` routing.
- Translates Gemini 3.8 to `thinkingConfig.thinkingLevel` ("LOW" | "MEDIUM" | "HIGH") across both OpenAI-to-Gemini and Claude-to-Gemini translators, avoiding invalid `thinkingBudget` injection.
- Strips unsupported legacy sampling parameters (`topK`, `topP`, `temperature`) for Gemini 3.8.

### 2. Live Catalog Synchronization & Self-Healing
- Removes static allowlist filtering from the quota normalization path (`liveModelIds.ts`, `quotaNormalize.ts`), allowing newly launched upstream models to retain their quota buckets immediately.
- Syncs `agy` model filtering to use `isDiscoverableAgyModelId`.
- Includes legacy connections without explicit `autoSync` into periodic synchronization.
- Enhances reactive model sync on 404: awaits the sync, re-resolves the model against updated discovery aliases, rebuilds the request envelope, and transparently retries once.

### 3. Family Quota Pools & Multi-Window Routing
- Parses `v1internal:retrieveUserQuotaSummary` into family quota pools (Gemini Models, Claude & GPT Models) with independent 5-hour and weekly windows, supporting both flat and nested `remaining.remainingFraction` responses.
- Persists `quotaGroups` in `providerLimitsCache` so initial dashboard load renders the grouped UI without requiring a manual refresh.
- Enforces multi-window quota safety: blocks requests when any reported family window (5h or weekly) reaches the threshold, preventing 429 loops when weekly is exhausted.
- Renders grouped quota pools with live discovered model inventories in Provider Limits, removing static ordering penalties for new models.

## Validation
[...]

Issue #12058, opened 2026-08-29
fix(api): Antigravity models not listed when MODELS_CATALOG_PREFIX_MODE=Canonical


cheynetom:
### OmniRoute Version

v3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Alma 10

### Provider(s) Involved

Antigravity, Antigravity-Cli

### Model(s) Involved

All `antigravity/*` and `agy/*` models except `antigravity/gemini-3.1-flash-image`

### Client Tool

cURL

### Description

When `MODELS_CATALOG_PREFIX_MODE=Canonical`, the `/v1/models` endpoint omits essentially all Antigravity and Antigravity-Cli provider models from the catalog. The sole exception observed is `antigravity/gemini-3.1-flash-image`, which is listed even though no other `antigravity/*` or `agy/*` model is.

In `Alias` mode, the same instance correctly lists the Antigravity and Antigravity-Cli models under their `antigravity/*` and `agy/*` alias prefixes, respectively (e.g. `antigravity/gemini-3.7-flash-tiered`, `agy/gemini-3.7-flash-high`, etc), confirming the models are configured, synced, and routable — they just don't surface once the catalog is switched to canonical prefixing.

This looks related to how `Canonical` mode resolves the canonical provider id/prefix for Antigravity specifically (`antigravity` vs. alias `agy`), possibly dropping models whose canonical-prefix resolution differs from other dual-prefix providers (similar in shape to #11832, which found canonical mode drops custom node-provider models — but this is core Antigravity provider models, not custom node models). The `gemini-3.1-flash-image` model surviving suggests the filter/resolution logic special-cases image-capable models or has an id-mapping edge case that happens to keep only that one model. It seems like maybe a prefix namespace collision between the Antigravity and Antigravity-Cli providers.

### Steps to Reproduce

. Configure an Antigravity and Antigravity-CLI provider connection with several synced models (e.g. `agy/gemini-3.7-flash-tiered`, `agy/gemini-3.7-flash-high`, `agy/gemini-pro-agent`, `agy/gemini-3.1-pro-high`, `agy/gemini-3.7-flash-low`, `agy/gemini-3-flash`).
2. Query the catalog in Alias mode and confirm all Antigravity models are listed under `agy/*` and `antigravity/*`:
   ```
   MODELS_CATALOG_PREFIX_MODE=Alias

[...]

PR #12738, opened 2026-09-04
refactor(providers): single source of truth for agy/antigravity public catalogs (#12724)

### LINKED ISSUE #12724: feat(providers): De-duplicate the `agy` and `antigravity` public model catalogs
> Verified against `upstream/main` @ `c0b2253f21` (2026-09-04).

## Problem / Use Case

`AGY_PUBLIC_MODELS` (`open-sse/config/agyModels.ts`) and `ANTIGRAVITY_PUBLIC_MODELS`
(`open-sse/config/antigravityModelAliases.ts`) are two hand-maintained copies of what is
currently the same list. Every time Google ships a new Antigravity-backed model, the same
edit has to be made twice, and the two files can silently drift.

The comment at the top of `agyModels.ts` states the rationale:

> The `agy` provider reuses the `antigravity` executor/translator (identical backend), but
> keeps its own catalog so the CLI and IDE model surfaces can evolve independently. Both
> currently expose the same callable Claude/Gemini/GPT set.

That rationale was real when the file was introduced, but it has not held since 2026-07-29.

### Evidence 1 — the two catalogs are currently identical

On `upstream/main` @ `c0b2253f21`, both lists contain the same 10 ids:

```
claude-opus-4-6-thinking   gemini-3.1-flash-lite   gemini-3.7-flash-low
claude-sonnet-4-6          gemini-3.1-pro-low      gemini-3.7-flash-medium
gemini-pro-agent           gemini-3.7-flash-high   gemini-3.7-flash-tiered
gpt-oss-120b-medium
```

`diff` of the sorted id sets is empty. The per-model spec fields match too — identical
distributions of `contextLength`, `maxOutputTokens`, `supportsReasoning`, `supportsVision`
and `toolCalling`.

### Evidence 2 — they have been identical for over a month

Every revision that has ever touched `agyModels.ts`, compared against
`ANTIGRAVITY_PUBLIC_MODELS` at the same commit:

| Date | Commit | agy ids | antigravity ids | Divergent ids |
|---|---|---|---|---|
| 2026-08-26 | `b4ec7807ab` (Release v3.8.50) | 10 | 10 | **0** |
| 2026-08-23 | `65e81158ab` (fix(ollama) …) | 10 | 10 | **0** |
| 2026-07-29 | `c9d4a45f18` (Release v3.8.49) | 15 | 15 | **0** |
| 2026-07-07 | `92715c8f2c` (Release v3.8.46, file added) | 15 | 19 | 4 |

The separation carried real content only at the moment it was introduced. The four ids that
IDE had and CLI did not, on 2026-07-07:

```
claude-sonnet-5
gemini-3-pro-preview
gemini-3-pro-image-preview
gemini-2.5-computer-use-preview-10-2025
```

All four are gone from both lists today. Since 2026-07-29 the catalogs have been byte-equal
in content across three separate revisions.

### Reproduce

```sh
git fetch upstream
cd <repo>

# current id sets
git show upstream/main:open-sse/config/agyModels.ts \
  | grep -oE 'id: "[^"]+"' | tr -d '"' | sed 's/id: //' | sort > /tmp/agy.ids
git show upstream/main:open-sse/config/antigravityModelAliases.ts \
  | awk '/ANTIGRAVITY_PUBLIC_MODELS/,/^\]\)/' \
  | grep -oE 'id: "[^"]+"' | tr -d '"' | sed 's/id: //' | sort > /tmp/anti.ids
diff /tmp/anti.ids /tmp/agy.ids   # empty

# divergence across history
for c in $(git log --format=%H upstream/main -- open-sse/config/agyModels.ts); do
  git show $c:open-sse/config/agyModels.ts \
    | grep -oE 'id: "[^"]+"' | tr -d '"' | sed 's/id: //' | sort > /tmp/g.ids
  git show $c:open-sse/config/antigravityModelAliases.ts \
    | awk '/ANTIGRAVITY_PUBLIC_MODELS/,/^\]\)/' \
    | grep -oE 'id: "[^"]+"' | tr -d '"' | sed 's/id: //' | sort > /tmp/a.ids
  echo "$(git log -1 --format=%ad --date=short $c) ${c:0:9} diff=$(diff /tmp/a.ids /tmp/g.ids | grep -c '^[<>]')"
done
```

## Proposed Solution

Keep the ability to diverge, remove the duplication: derive both catalogs from one shared
base plus an explicit per-surface delta.

```ts
// open-sse/config/antigravitySharedModels.ts
export const ANTIGRAVITY_SHARED_MODELS = Object.freeze([ /* the 10 current entries */ ]);

// agyModels.ts
export const AGY_PUBLIC_MODELS = buildSurfaceCatalog(ANTIGRAVITY_SHARED_MODELS, {
  add: [],      // CLI-only models
  remove: [],   // ids the CLI surface does not expose
});

// antigravityModelAliases.ts
export const ANTIGRAVITY_PUBLIC_MODELS = buildSurfaceCatalog(ANTIGRAVITY_SHARED_MODELS, {
  add: [],      // IDE-only models (e.g. the four from 2026-07-07)
  remove: [],
});
```

With both deltas empty today, the intent of the original comment is preserved — a future
IDE-only preview model is a one-line `add:` entry — while a new Gemini release becomes a
single edit instead of two.

`open-sse/config/providers/registry/agy/index.ts` keeps using `models: [...AGY_PUBLIC_MODELS]`
unchanged, so no consumer has to move.

## Alternatives Considered

1. **Leave as is.** Works, but every model release is a double edit with no compile-time
   guard against drift. The current lag illustrates the cost: Gemini 3.8 Flash shipped
   2026-09-02 and neither catalog has it yet, and each future catch-up is two edits.
2. **Delete `agyModels.ts` and point `agy` at `ANTIGRAVITY_PUBLIC_MODELS`.** Simplest, but
   discards the stated design intent and makes a future CLI/IDE split a revert rather than a
   one-line delta. Not proposed for that reason.
3. **Make the catalog dynamic** from `:fetchAvailableModels`. A discovery path already
   exists (`src/app/api/providers/[id]/models/discovery/`), but its results do not feed the
   registry. That is a larger change and orthogonal to this one — the pinned catalog is
   deliberate, and this request does not challenge that.

## Acceptance Criteria

- [ ] A single shared source defines the entries common to both surfaces.
- [ ] `AGY_PUBLIC_MODELS` and `ANTIGRAVITY_PUBLIC_MODELS` are derived from it, each with an
      explicit (currently empty) add/remove delta.
- [ ] The public id set and every spec field of both catalogs are unchanged by the refactor.
- [ ] Adding a model to both surfaces requires editing exactly one file.
- [ ] A per-surface delta still works: adding an id to one `add:` list exposes it on that
      surface only.
- [ ] The stale "can evolve independently … both currently expose the same set" comment is
      replaced by one describing the shared-base mechanism.

## Area

Provider registry / model catalogs

## Related Provider(s)

`agy` (Antigravity CLI), `antigravity` (Antigravity IDE)

## Additional Context

The two providers already share the executor and translator — `agyModels.ts` describes the
backend as `identical`. This request only aligns catalog maintenance with that existing
reality; it does not propose merging the provider entries themselves.

## Expected Test Plan

- Extend `tests/unit/agy-provider.test.ts` and `tests/unit/antigravity-retired-public-models.test.ts`
  to assert each surface's resolved catalog equals `shared + add − remove`.
- Add a regression test asserting the exported id sets and spec fields are unchanged by the
  refactor (snapshot of the current 10 entries).
- Add a test that a non-empty `add:` on one surface does not leak into the other.
- `tests/unit/antigravity-image-catalog-separation.test.ts` should keep passing unmodified.

gonisulaimann:
Fixes #12724.

`agyModels.ts` and `antigravityModelAliases.ts` carried two hand-maintained copies of the same model list — byte-identical since 2026-07-29, so every new Antigravity-backed model meant the same edit twice and the copies could silently drift. This makes `ANTIGRAVITY_PUBLIC_MODELS` the single source of truth and turns `AGY_PUBLIC_MODELS` into a re-export of it.

I verified on this branch that the two arrays are deep-equal (same ids, same order, same specs) before collapsing them, so this is behavior-preserving. The agy-specific helpers (`isUserCallableAgyModelId`, `isDiscoverableAgyModelId`, retired/non-chat id sets, client-visible names) stay in `agyModels.ts` — they encode CLI-surface policy, not catalog content.

Tests: `tests/unit/agy-antigravity-catalog-dedup-12724.test.ts` pins the invariant (same array reference, same id set, and the documented 10-model set). The existing agy/antigravity suites (20 tests) and the #12058 catalog tests stay green; `check:provider-consistency`, typecheck and lint pass.

Note for later merges: two open PRs (#12608, #12672) touch both catalogs — after this lands, a catalog refresh only needs to edit `antigravityModelAliases.ts`.

base-red inherited: #12732 (pre-existing — blames other PRs, none related to this change).

### DIFF
--- a/changelog.d/fixes/12724-agy-antigravity-catalog-dedup.md
+++ b/changelog.d/fixes/12724-agy-antigravity-catalog-dedup.md
@@ -1,0 +1,1 @@
+- **fix(providers):** `AGY_PUBLIC_MODELS` is now a re-export of the canonical `ANTIGRAVITY_PUBLIC_MODELS` catalog instead of a second hand-maintained copy — the two lists had been byte-equal for over a month and could silently drift whenever a new Antigravity-backed model shipped ([#12724](https://github.com/diegosouzapw/OmniRoute/issues/12724))
--- a/open-sse/config/agyModels.ts
+++ b/open-sse/config/agyModels.ts
@@ -1,110 +1,20 @@
 // Antigravity CLI (`agy`) model catalog.
 //
-// These models are pinned from the live `:fetchAvailableModels` endpoint
-// (https://daily-cloudcode-pa.googleapis.com/v1internal:fetchAvailableModels) using a
-// real `agy` consumer-OAuth token. The public catalog exposes the upstream Gemini 3.7
-// Flash ids verbatim; the shared Antigravity executor dispatches them unchanged.
+// The `agy` provider reuses the `antigravity` executor/translator (identical
+// backend). Its public catalog is a direct re-export of ANTIGRAVITY_PUBLIC_MODELS:
+// both surfaces expose the same callable Claude/Gemini/GPT set and the lists
+// were hand-maintained as two copies until they were byte-equal for months —
+// a new model shipped on one side silently missed the other (#12724). Keeping
+// one canonical array in antigravityModelAliases.ts makes drift impossible.
+// Tab-completion models (`tab_flash_lite_preview`, `tab_jump_flash_lite_preview`)
+// are intentionally excluded from the shared catalog — they are not chat-callable.
 //
-// The `agy` provider reuses the `antigravity` executor/translator (identical backend),
-// but keeps its own catalog so the CLI and IDE model surfaces can evolve independently.
-// Both currently expose the same callable Claude/Gemini/GPT set. Tab-completion models
-// (`tab_flash_lite_preview`, `tab_jump_flash_lite_preview`) are intentionally excluded —
-// they are not chat-callable.
+// The agy-specific helpers below (non-chat / retired id sets, client-visible
+// names) stay here because they encode CLI-surface policy, not catalog content.
 
-export const AGY_PUBLIC_MODELS = Object.freeze([
-  // Gemini 3.7 Flash tiers. The live endpoint selects High by default and advertises
-  // all three ids to both the IDE 2.5.5 and CLI 1.1.x clients.
-  {
-    id: "gemini-3.7-flash-high",
-    name: "Gemini 3.7 Flash (High)",
-    contextLength: 1048576,
-    maxOutputTokens: 65536,
-    supportsReasoning: true,
-    supportsVision: true,
-    toolCalling: true,
-  },
-  {
-    id: "gemini-3.7-flash-medium",
-    name: "Gemini 3.7 Flash (Medium)",
-    contextLength: 1048576,
-    maxOutputTokens: 65536,
-    supportsReasoning: true,
-    supportsVision: true,
-    toolCalling: true,
-  },
-  {
-    id: "gemini-3.7-flash-low",
-    name: "Gemini 3.7 Flash (Low)",
-    contextLength: 1048576,
-    maxOutputTokens: 65536,
-    supportsReasoning: true,
-    supportsVision: true,
-    toolCalling: true,
-  },
-  {
-    id: "gemini-3.7-flash-tiered",
-    name: "Gemini 3.7 Flash (Tiered)",
-    contextLength: 1048576,
-    maxOutputTokens: 65536,
-    supportsReasoning: true,
-    supportsVision: true,
-    toolCalling: true,
-  },
-  // Gemini 3.1 Pro
-  {
-    id: "gemini-pro-agent",
-    name: "Gemini 3.1 Pro (High)",
-    contextLength: 1048576,
-    maxOutputTokens: 65535,
-    supportsReasoning: true,
-    supportsVision: true,
-    toolCalling: true,
-  },
-  {
-    id: "gemini-3.1-pro-low",
-    name: "Gemini 3.1 Pro (Low)",
-    contextLength: 1048576,
-    maxOutputTokens: 65535,
-    supportsReasoning: true,
-    supportsVision: true,
-    toolCalling: true,
-  },
-  {
-    id: "gemini-3.1-flash-lite",
-    name: "Gemini 3.1 Flash Lite",
-    contextLength: 1048576,
-    maxOutputTokens: 65535,
-    toolCalling: true,
-  },
-  // Claude (Antigravity backend).
-  {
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Hi @gonisulaimann — thank you for this, and for getting there first. 🙏

You opened this on 2026-09-04 and nailed the diagnosis: the two model lists in `agyModels.ts` and `antigravityModelAliases.ts` had been byte-identical since July, so every new Antigravity-backed model meant making the same edit twice and hoping the copies never drifted. That was a real maintenance trap, and your PR is the first one that called it out with a fix.

I'm closing this one, and I want to be clear about why, because it is not a judgement on the work.

Issue #12724 has since been resolved on `release/v3.8.51` by #13410, which took a different shape. Instead of making one catalog a re-export of the other, it moved the list into a new shared base (`open-sse/config/antigravitySharedModels.ts`). Each surface then builds its own catalog from that base with explicit `add` / `remove` lists:

```ts
export const AGY_PUBLIC_MODELS = buildSurfaceCatalog(ANTIGRAVITY_SHARED_MODELS, {
  add: [], // CLI-only models (currently none)
  remove: [], // Models hidden from CLI (currently none)
});
```

Both lists are empty today, so the two catalogs are still identical, which is exactly the property your test asserted. The difference is that the CLI and the IDE can now diverge on purpose later without anyone having to undo a strict re-export.

Two things follow from that:

- **Your branch now conflicts with the release branch** on `agyModels.ts`, since the file it rewrites has been rewritten again.
- **Merging it would work against the newer design.** A strict re-export would remove the per-surface lists that #13410 just added.

None of your code was carried into #13410. It is a separate implementation, which is why this closes rather than merges.

If you want to pick something else up in this area, the per-surface lists are a good place to start. They're empty right now, and the moment a model ships on one surface but not the other, they're what keeps the two catalogs honest.

Thanks again for spotting this and for the clear write-up.

PR #12740, opened 2026-09-04
feat(dashboard): add description display modes to feature-flags cards

### LINKED ISSUE #12739: feat(backend): Feature-flags card descriptions are truncated to 2 lines (text cut off)
## Problem

On `/dashboard/settings/feature-flags`, every flag card clamps its description to **2 lines** (`line-clamp-2` in `FeatureFlagCard.tsx`). The full text exists in the DOM but is visually cut off with an ellipsis — measured in-browser, **23 of 54 cards** are affected (e.g. `NETWORK_ROTATION_SHARED_EGRESS_GUARD` shows only ~32px of its ~128px description).

Zooming in/out doesn't help because it's a line-count clamp, not a sizing issue. For an admin page where the description is the entire basis for deciding whether to toggle a security/network flag, hiding most of it hurts usability.

## Proposed fix (implemented on `feat/dashboard-feature-flags-description-modes`)

Add a segmented **"On hover / Full text"** control in the feature-flags grid header:

- **On hover** (default): cards stay compact at 2 lines; hovering a description reveals the full text in a floating tooltip (styled consistently with existing dropdowns: `border-border bg-surface shadow-lg`).
- **Full text**: every description is always shown in full.
- The choice persists across reloads via `localStorage`, read after hydration to avoid an SSR/client mismatch (the active pill stayed stale otherwise).
- New keys added under the `featureFlags` i18n namespace; other locales fall back to English via the existing merge.

## Verification

- Playwright: hover tooltip shows the complete description; "Full text" mode shows 0 clamped descriptions; choice persists across reloads in both directions.
- ESLint clean; pre-commit hooks (prettier, eslint, docs-sync, quality gates) all pass.

Image

Image

Image

jonlwheat2-gif:
Closes #12739

## Problem

Feature-flag card descriptions were hard-clamped to 2 lines (`line-clamp-2`), silently hiding most of the text — 23 of 54 cards affected, and zooming can't help since it's a line-count clamp.

## Changes

- **`FeatureFlagCard.tsx`**: description renders either always-in-full, or clamped to 2 lines with a hover tooltip showing the complete text (styled like existing dropdowns: `border-border bg-surface shadow-lg`); flag keys now wrap (`break-all`) instead of ellipsizing.
- **`FeatureFlagsGrid.tsx`**: new segmented **"On hover / Full text"** control in the page header; choice persisted in `localStorage` and read after hydration (avoids an SSR/client mismatch that left the active pill stale).
- **`en.json`**: 4 keys under the `featureFlags` namespace (other locales fall back to English via the existing merge).

## Verification

- Playwright: hover tooltip shows the full description; "Full text" mode yields 0 clamped descriptions; mode persists across reloads in both directions; tooltip doesn't clip at the viewport edge.
- ESLint clean; repo pre-commit hooks (prettier, eslint --fix, docs-sync, t11 any-budget, tracked-artifacts) all pass.

### DIFF
--- a/src/app/(dashboard)/dashboard/settings/components/FeatureFlagCard.tsx
+++ b/src/app/(dashboard)/dashboard/settings/components/FeatureFlagCard.tsx
@@ -18,6 +18,8 @@ interface FeatureFlagCardProps {
   onToggle: (key: string, newValue: string) => void;
   onReset: (key: string) => void;
   saving?: boolean;
+  /** "clamp": 2-line clamp + full text on hover. "full": always expanded. */
+  descriptionMode?: "clamp" | "full";
 }
 
 const CATEGORY_STYLES: Record<
@@ -120,6 +122,7 @@ export default function FeatureFlagCard({
   onToggle,
   onReset,
   saving = false,
+  descriptionMode = "clamp",
 }: FeatureFlagCardProps) {
   const t = useTranslations("featureFlags");
   const enabled = flag.type === "boolean" ? isEnabled(flag.effectiveValue) : false;
@@ -212,8 +215,29 @@ export default function FeatureFlagCard({
         )}
       </div>
 
-      {/* Description */}
-      <p className="mb-3 line-clamp-2 text-xs text-text-muted">{flag.description}</p>
+      {/* Description — either always expanded in full, or clamped to 2 lines
+          with the full text revealed on hover (mode set in the grid toolbar) */}
+      {descriptionMode === "full" ? (
+        <p className="mb-3 text-pretty text-xs leading-relaxed text-text-muted">
+          {flag.description}
+        </p>
+      ) : (
+        <div className="group/desc relative">
+          <p
+            aria-describedby={`desc-tooltip-${flag.key}`}
+            className="mb-3 cursor-help line-clamp-2 text-xs text-text-muted"
+          >
+            {flag.description}
+          </p>
+          <div
+            id={`desc-tooltip-${flag.key}`}
+            role="tooltip"
+            className="pointer-events-none invisible absolute left-0 top-full z-50 mt-2 w-80 rounded-lg border border-border bg-surface p-3 text-xs leading-relaxed text-text-primary opacity-0 shadow-lg transition-opacity duration-150 group-hover/desc:visible group-hover/desc:opacity-100"
+          >
+            {flag.description}
+          </div>
+        </div>
+      )}
 
       <EnvPrecedenceWarning flag={flag} text={t("ccDiscoveryAliasesEnvWarning")} />
 
--- a/src/app/(dashboard)/dashboard/settings/components/FeatureFlagsGrid.tsx
+++ b/src/app/(dashboard)/dashboard/settings/components/FeatureFlagsGrid.tsx
@@ -38,6 +38,11 @@ interface FlagUpdateResult {
 
 const ACTIVE_VALUES = new Set(["true", "1", "yes"]);
 
+// How card descriptions are displayed. "clamp" keeps cards compact and reveals
+// the full text on hover; "full" always shows the complete description.
+type DescriptionMode = "clamp" | "full";
+const DESCRIPTION_MODE_STORAGE_KEY = "ff-card-description-mode";
+
 function isActiveFlagValue(value: string): boolean {
   return ACTIVE_VALUES.has(value);
 }
@@ -67,6 +72,7 @@ export default function FeatureFlagsGrid() {
   const [resettingAll, setResettingAll] = useState(false);
   const [showResetConfirm, setShowResetConfirm] = useState(false);
   const [debouncedSearch, setDebouncedSearch] = useState("");
+  const [descriptionMode, setDescriptionMode] = useState<DescriptionMode>("clamp");
   // Set of flag keys whose DB override was changed in this session without
   // a subsequent restart. Used to surface the restart banner.
   const [pendingRestartKeys, setPendingRestartKeys] = useState<Set<string>>(new Set());
@@ -98,6 +104,28 @@ export default function FeatureFlagsGrid() {
     const timer = setTimeout(() => setDebouncedSearch(search), 300);
     return () => clearTimeout(timer);
   }, [search]);
+
+  // Read the persisted preference after hydration — reading localStorage in the
+  // useState initializer would make the first client render differ from the SSR
+  // HTML (hydration mismatch), leaving the active pill stale.
+  useEffect(() => {
+    try {
+      if (window.localStorage.getItem(DESCRIPTION_MODE_STORAGE_KEY) === "full") {
+        // eslint-disable-next-line react-hooks/set-state-in-effect -- localStorage hydration, runs once
+        setDescriptionMode("full");
+      }
+    } catch {
+      // localStorage unavailable — keep the default.
+    }
+  }, []);
+
+  useEffect(() => {
+    try {
+      window.localStorage.setItem(DESCRIPTION_MODE_STORAGE_KEY, descriptionMode);
+    } catch {
+      // localStorage unavailable — the mode just won't persist across reloads.
+    }
+  }, [descriptionMode]);
 
   const applyFlagResult = useCallback((key: string, result: FlagUpdateResult) => {
     const wasActive = isActiveFlagValue(result.previousValue);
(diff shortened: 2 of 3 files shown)


diegosouzapw:
This is a nicer, more complete take on #12739 than the alternative PR (#13189) — the persisted
clamp/full toggle with hydration-safe localStorage reading and the styled tooltip are well done.
Before merging we'd like to see test coverage added (there isn't any test file in this PR yet) —
at minimum covering the two display modes and that localStorage persistence doesn't cause a
hydration mismatch. Also please add a changelog fragment under `changelog.d/features/`. We're
planning to close #13189 in favor of this PR since it does everything that one does and more —
happy to make sure that author's test approach gets credited if we end up borrowing from it.

jonlwheat2-gif:
## Review follow-up — tests, changelog, and the #13189 credit

All three asks are addressed on `jonlwheat2-gif:feat/dashboard-feature-flags-description-modes-v2`
(this feature replayed onto current `release/v3.8.51`).

### 1. Test coverage

`tests/unit/ui/FeatureFlagDescriptionModes.test.tsx` — 10 tests, all passing under
`npm run test:vitest:ui`.

Both display modes:

- **`clamp` (default)** — the visible paragraph keeps `line-clamp-2` and still carries the whole
  description; the tooltip is wired by `aria-describedby` pointing at the `role="tooltip"` node and
  holds the un-truncated text; tooltip ids are scoped per flag key so sibling cards can't collide
  (`desc-tooltip-FLAG_ONE` / `desc-tooltip-FLAG_TWO`).
- **`full`** — the clamp is dropped, the whole description renders, and **no** tooltip or
  `aria-describedby` is emitted.

The hydration contract you called out is pinned by a server-render test. `renderToString` runs no
effects, so a persisted mode cannot leak into the SSR markup:

```tsx
it("server-renders 'On hover' as active even when localStorage already holds 'full'", () => {
  window.localStorage.setItem(STORAGE_KEY, "full");
  const html = renderToString(React.createElement(FeatureFlagsGrid));
  expect(activeLabel(parseSsr(html))).toBe("On hover");
});
```

That assertion is the one that fails if someone moves the `localStorage` read back into the
`useState` initializer — which is exactly the regression the fix exists to prevent, so it is the
test that earns its keep.

Also covered: the grid defaults to "On hover"; picking "Full text" persists `"full"` and switching
back persists `"clamp"`; a persisted `"full"` is adopted after mount; a `localStorage` that **throws**
(private mode / disabled storage) still renders and still toggles; and the grid's mode actually
reaches the cards.

### 2. Changelog fragment

`changelog.d/features/12740-feature-flags-description-modes.md` — single `- ` bullet, PR-numbered
filename, `check:changelog-integrity` green.

### 3. The #13189 credit

We did borrow from it, so the credit is owed — no need to hold it back.

Specifically: the card-render harness and the long-description fixture were carried over from
@retroamx's `tests/unit/ui/FeatureFlagCard.test.tsx`. What is asserted here is genuinely different
(`aria-describedby` tooltip wiring, the `localStorage` round-trip, the SSR/hydration contract — none
of which #13189 covered), but the scaffolding came from there.

It is credited in both places:

- a `// Credit:` block in the test-file header, and
- the changelog bullet, using the compound form already in the tree
  (`changelog.d/fixes/13471-opencode-muse-spark-1-3-responses.md`):

```
— thanks @jonlwheat2-gif (with thanks to @retroamx for the #13189 card-render harness)
```

### One correction to "does everything that one does and more"

Worth flagging so it's a deliberate call rather than a surprise: **this does not carry over #13189's
`title={flag.description}`.** That change is replaced by the styled tooltip, because keeping both
would fire the native OS tooltip *and* the custom one on the same hover.

The user-facing outcome is a superset — hover reveals the full text, and `aria-describedby` +
`role="tooltip"` additionally expose it to assistive tech, which a bare `title` does not do well. But
if you'd rather keep `title` as a progressive-enhancement fallback, that's a one-line addition; say
the word and I'll add it.

Related, and pre-existing rather than a regression: the tooltip is revealed by `group-hover/desc`
only, so a sighted keyboard-only user can't surface it. `title` has the same limitation (it doesn't
appear on focus either), so nothing is lost versus #13189 — but if you want it closed properly, a
`group-focus-within/desc:` variant plus `tabIndex={0}` on the paragraph would do it. Happy to add
that in this PR or leave it as a follow-up.

---

## CI triage on the current head — one real failure, the rest is the base

Separately from the above, I went through the failing jobs on `04f3e1821`. Only one group is caused
by this PR.

### Real: `vi` and `pt-BR` lost key parity

`en.json` gains four keys under `featureFlags`, but two locales have **full key-parity gates against
`en.json`**, and both went red:

- `tests/unit/i18n-vi-completeness.test.ts` — 3 failures
- `tests/unit/i18n-pt-br.test.ts` — 1 failure (`should contain every key present in en.json (no drift, #6695)`)

```
actual: [ { key: 'featureFlags.descriptionModeLabel', reason: 'missing' },
          { key: 'featureFlags.descriptionModeHint',  reason: 'missing' },
          { key: 'featureFlags.descriptionModeHover', reason: 'missing' },
          { key: 'featureFlags.descriptionModeFull',  reason: 'missing' } ]
expected: []
```

These two are the only locales with such a gate — I checked every test under `tests/unit/` for a full
parity comparison against `en.json`, and only those two files do it. The other 63 non-English locales
are gated only by `npm run i18n:check-ui-coverage`, which requires ≥80% and passes on the English
runtime fallback (they currently sit at 99.2%). So "other locales fall back to English" is right for
63 of 65 — just not for these two.

Fix: add the following to the `featureFlags` block of each file, right after `searchPlaceholder`.

`src/i18n/messages/vi.json`

```json
"descriptionModeLabel": "Chế độ hiển thị mô tả",
"descriptionModeHint": "Khi di chuột: thẻ thu gọn và di chuột qua mô tả sẽ hiển thị toàn bộ văn bản. Toàn bộ văn bản: luôn hiển thị mô tả đầy đủ.",
"descriptionModeHover": "Khi di chuột",
"descriptionModeFull": "Toàn bộ văn bản",
```

`src/i18n/messages/pt-BR.json`

```json
"descriptionModeLabel": "Visualização da descrição",
"descriptionModeHint": "Ao passar o mouse: os cartões ficam compactos e passar o mouse sobre uma descrição mostra o texto completo. Texto completo: sempre mostrar a descrição completa.",
"descriptionModeHover": "Ao passar o mouse",
"descriptionModeFull": "Texto completo",
```

With those in place the gates go green locally — vi 8/8 across both i18n files, `i18n:check-ui-coverage`
reports **vi at 100.0% (13069/13069), all 65 locales ≥80%**. `i18n:check-new-keys` and
`i18n:check-value-drift` also pass.

### Not this PR: the branch is 299 commits behind `release/v3.8.51`

Merge base is `c3945a724`. The remaining failures all land in files this PR never touches:

- **`Fast Quality Gates`** → `forgotten-sibling-tests: Invalid string length`. The step runs with
  `GITHUB_BASE_SHA: c3945a724…`, so the changed-file set spans ~300 commits and the script exceeds
  Node's maximum string length. This one goes away on a rebase.
- **`API Route Typecheck`** → `src/lib/db/callLogStats.ts` (TS2300) and
  `src/app/api/combos/test/route.ts` (TS2339). Neither file is in this PR.
- **Unit-test shards** → also fail on `paid-model-target-6540.test.ts` (imports only
  `@/shared/utils/freeModels`), `cli-i18n-catalog.test.ts` (`serve.ready_timeout`, a CLI catalog key),
  `i18n-locale-scaffold.test.ts` (reads only the README and `docs/guides/I18N.md`), plus
  streamChunks / redaction / paid-model-settings tests.
- **`Vitest (fast-path)` passes**, so the new test file itself is green.

### Two small things

1. The description says "flag keys now wrap (`break-all`) instead of ellipsizing". `break-all` appears
   in no file in the diff, and the flag-key span still carries `truncate`. Either that line was
   dropped or the description overstates the change.
2. `changelog.d/features/feature-flags-description-display-modes.md` is missing the PR-number prefix.
   `changelog.d/README.md` and the CONTRIBUTING checklist both specify `<PR-number>-<short-slug>.md`.
   Note `check:changelog-integrity` does not actually validate filenames (the aggregate script only
   filters `*.md`), so this will not fail CI — it just breaks the documented convention.

---

For reference the rebased branch is `jonlwheat2-gif:feat/dashboard-feature-flags-description-modes-v2`
— the feature commit replayed cleanly (both components are unchanged upstream), plus the 10-test
suite, the locale fix above, the credited fragment, and a `12740-`-prefixed filename. Happy to push
the locale fix to this branch instead, or hand that branch over as the PR head if a rebase is easier
on your side.

PR #12733, opened 2026-09-04
fix(dashboard): moonshot voucher/cash leftover follows bucket balance

HouMinXi:
## Bug

On Moonshot / Kimi Open Platform keys, the dashboard can show **Available 0%** next to **Voucher 100%** and **Cash 100%**. That combination is not a real wallet state.

Moonshot's balance API returns three absolute numbers (`available_balance`, `voucher_balance`, `cash_balance`). The usage fetcher treated leftover as a boolean for Available (`0` vs `100`) but hardcoded voucher and cash leftover to `100` even when those buckets were `0`. The quota parser then walked the generic path, dropped `currency`, and the card painted a percentage bar instead of ¥0.00.

Measured on X500: node `moonshot-native` (`mnative` → `https://api.moonshot.cn/v1`), 38 connections. `quota_snapshots` for this provider is empty; the numbers come from `GET /v1/users/me/balance` at display time.

## Fix

- Drive leftover from `remaining > 0` for all three buckets.
- Stop marking the rows `unlimited: true` (that path also forces 100% in some widgets).
- Parse `available` / `voucher` / `cash` as credits rows so the card shows the currency amount.

Empty wallet → Available / Voucher / Cash all 0%. A leftover voucher of 5 with available 0 still shows Voucher 100% and Cash 0%.

## Tests

- `tests/unit/moonshot-quota-fetcher.test.ts` — empty buckets are 0%; voucher leftover stays 100% only when that bucket has money.
- `tests/unit/moonshot-quota-dashboard-rendering.test.ts` — producer → `parseQuotaData` → `getQuotaRemainingPercentage` is 0 for an empty wallet, with `isCredits` + `CNY`.
- Bug-injection: hardcode leftover to 100 → the empty-wallet test fails; restore → pass.
- Sibling: agentrouter credits rendering, provider-limits UI, T13 stale reset — 41/41 pass.

## Not in this PR

- HuggingChat / stream public-error unit failures on `release/v3.8.51` (unrelated).
- Filling `quota_snapshots` for Moonshot (display already live-fetches `/v1/users/me/balance`).

Issue #5165, opened 2026-06-27
[BUG] OmniRoute shows "Connectivity Issue" and enters offline mode even though internet is available


Flexible78:
### OmniRoute Version

3.8.37

### Installation Method

Electron desktop app

### Operating System

Windows

### OS Version

windows 11

### Description

OmniRoute reports a connectivity error even though the machine has working internet access.

The UI shows:

Connectivity Issue
OmniRoute cannot reach the network right now. Check your internet, VPN, or proxy settings.
Offline mode detected

This appears to be a false offline detection, since the internet is working normally in the browser and other apps.

update by omniroute.exe file

### Steps to Reproduce

Open OmniRoute dashboard.

Wait for the app to load.

Observe the connectivity status.

The app displays a connectivity error and switches to offline mode.

### Expected Behavior

OmniRoute should detect the network correctly and remain online when internet access is available.

### Actual Behavior

OmniRoute incorrectly reports a connectivity issue and shows offline mode, even though the system has internet access.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
A test should verify that OmniRoute correctly detects network availability and does not enter offline mode when the network is reachable.

Connectivity Issue
OmniRoute cannot reach the network right now. Check your internet, VPN, or proxy settings.
Offline mode detected
Retry Connection
Open Status Page
```

diegosouzapw:
The "Connectivity Issue / Offline mode" page is served by OmniRoute's own service worker when it can't reach the **local** app server (`http://localhost:20128`) — it's not detecting a real internet outage. So the actual problem is the built-in server not coming up after the update. Please check:
1. Does the OmniRoute tray icon appear in the Windows system tray? If not, the server failed to start.
2. Did Windows Defender / your antivirus quarantine or block the newly-installed `omniroute.exe` after the auto-update? This is common for freshly-updated unsigned binaries on Windows 11.
3. Is port 20128 already in use? Run `netstat -ano | findstr :20128` in Command Prompt.
4. Try a clean uninstall + fresh install of the new version.

If none of those help, please share the Electron app logs (`%APPDATA%\OmniRoute\logs`, or Help → Open Logs Folder).

Flexible78:
i download 0mniRoute.3.8.38.exe and ran this file - everything was ok.
but right now how to update your app 
before it i just replace file 0mniRoute.exe but now?

[...]

PR #12767, opened 2026-09-05
fix(pwa): do not intercept navigations so Chrome can retry HTTP/2

HouMinXi:
## Bug

A Chrome tab left on `/dashboard/quota` eventually fails to reload. F5 times out. A new tab on the same origin loads immediately. `curl` against the origin stays 200.

Seen on both a LAN HTTPS origin (TCP 20128, no UDP) and a Cloudflare-fronted hostname.

## Cause

Two things stack:

1. Caddy 2.11 defaults to `protocols = [h1 h2 h3]` and sends `Alt-Svc: h3=":<port>"; ma=2592000`. If the process only publishes TCP, that advertisement is a lie. Chrome caches it for 30 days and retries QUIC on F5. UDP never answers.
2. `public/sw.js` called `event.respondWith` on `mode:navigate` for `/dashboard/*`. On a dead QUIC socket the worker's `fetch()` hangs or errors, and `Response.error()` aborts the document load. Chrome never gets to fall back to HTTP/2 on that tab. A new tab opens a fresh TCP handshake, so it works.

This PR only changes the service worker. The Caddy `protocols h1 h2` + `header Alt-Svc "clear"` fix belongs in the reverse-proxy config, not here.

`pathname.startsWith("/dashboard")` also matched `/dashboardfoo`. Second commit requires a path-segment boundary.

## Change

- Navigations: do not call `respondWith`. The browser owns the document load and the H3→H2 retry.
- Cache name `omniroute-pwa-v2` → `omniroute-pwa-v3` so existing clients drop the old worker.
- `pathIsExcluded`: strip a trailing slash, then exact match or `${base}/`.
- Static assets (`style` / `script` / `image` / `font`) stay cache-first.

No custom offline HTML for navigations. Serving a stale shell from the worker is what hung F5. Offline dashboard is a follow-up.

## Tests

```
node --import tsx/esm --test tests/unit/service-worker-navigation-fallback.test.ts
```

4/4 pass. Naive `startsWith` injection fails `/dashboardfoo`; restore is green. `node --check public/sw.js` is clean. CI shards also ran the four `#11779` / `#5165` cases green.

After merge + deploy: F5 a stale tab on `/dashboard/quota` without opening a new one. If the tab is still on v2, unregister the worker in DevTools or wait for v3 to activate.

WAN `h3=":443"` from Cloudflare is a real listener. Leave it.

## ⚠️ base-red inherited

This PR does not touch `open-sse/executors/glm.ts`, `changelog.d/fixes/reset-aware-model-family.md`, `stryker.conf.json`, or the sanitizer / stream-boundary tests that failed on CI. Those gates are already red on `release/v3.8.51` (same changelog fragment on sibling #12733; glm `TS2554` baseline 0→1; mutation list missing `tests/unit/reset-aware-request-scope-12600.test.ts`). Not fixing them here.

### DIFF
--- a/changelog.d/fixes/dashboard-stale-tab-refresh.md
+++ b/changelog.d/fixes/dashboard-stale-tab-refresh.md
@@ -1,0 +1,1 @@
+- **fix(pwa):** do not intercept dashboard navigations so Chrome can fall back from a stale HTTP/3 Alt-Svc advertisement (UDP :20128 is unpublished; F5 on a long-lived tab hung until a new tab opened a fresh TCP connection).
--- a/docs/guides/PWA_GUIDE.md
+++ b/docs/guides/PWA_GUIDE.md
@@ -58,8 +58,8 @@ OmniRoute includes a service worker (`sw.js`) that provides intelligent caching:
 | **App Shell**                                           | Cache-first                        | `/`, `/offline`, manifest, and icons are pre-cached on install               |
 | **Static assets** (CSS, JS, images, fonts)              | Network-first with cache fallback  | Fetches fresh from the network; falls back to cache if offline               |
 | **Next.js bundles** (`/_next/`)                         | Network-first with cache update    | Fetches from network and updates cache; serves cached version if offline     |
-| **Navigation requests**                                 | Network-only with offline fallback | Always fetches from network; shows `/offline` page if network is unavailable |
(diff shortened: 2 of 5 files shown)


HouMinXi:

HouMinXi:

`9e4b046b1` → `c4381e5fe`. range-diff 2 commits `=`. GPG G. No file overlap with the three landed commits.

HouMinXi:

- `c4381e5fea3` → `ff0f718fcff`
- `--force-with-lease` to `HouMinXi:fix/dashboard-stale-tab-refresh`

Issue #1421, opened 2026-04-19
[BUG] additionalProperties in tool schemas causes Gemini API errors


hartmark:
### OmniRoute Version

3.6.9

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Arch Linux

### Provider(s) Involved

Gemini

### Model(s) Involved

gemini gemma-4 models

### Client Tool

OpenClaw

### Description

When routing tool calls through OmniRoute to the Gemini backend, the OpenAI-format tool schemas are passed to Gemini's `function_declarations` without stripping fields that Gemini does not support. This causes a `400 Invalid JSON payload` error.

This was reproduced using the **Responses API** as the entry point. It is unclear whether the same issue occurs via the Completions API endpoint.

### Steps to Reproduce

1. Send a request via the **Responses API** with a tool that includes `additionalProperties` in its schema (e.g. `"additionalProperties": true` on a nested object)
2. OmniRoute routes the request to a Gemini backend
3. OmniRoute translates the OpenAI tool format to Gemini `function_declarations`
4. Gemini rejects the request with a 400 error

**Example schema (OpenAI format input)**

```json
{
  "items": {
    "additionalProperties": true,
    "type": "object",
    "properties": {
      "reason": {
        "type": "string",
        "description": "Brief explanation of why you are calling this tool"
      }
    }
  }
}
```

### Expected Behavior

OmniRoute strips unsupported JSON Schema fields when translating tool schemas for the Gemini backend, without loss of meaningful semantic information. `additionalProperties: true` is the default assumption in Gemini anyway and carries no functional impact when removed.

### Actual Behavior

Working response, or fallback to another provider to test if the request may work there.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Response from gemini:

[400]: Invalid JSON payload received. Unknown name "additionalProperties" at 'tools[0].function_declarations[23].parameters.properties[40].value.items': Cannot find field. Invalid JSON payload received. Unknown name "additionalProperties" at 'tools[0].function_declarations[23].parameters.properties[47].value.properties[14].value.items': Cannot find field.
```

diegosouzapw:
Thank you for reporting this! The issue has been fixed and is available in the release/v3.6.9 branch.

PR #12768, opened 2026-09-05
fix(translator): preserve additionalProperties: true for Gemini schemas (#12509)

### LINKED ISSUE #12509: fix(providers): strip prefixItems from Gemini tool schemas to avoid upstream 400

KooshaPari:
## Summary
Fixes #12509

Gemini API \`function_declarations\` rejects \`additionalProperties: false\` with HTTP 400 \"Unknown name\", but **accepts** (and honors) \`additionalProperties: true\` — strict-by-default validation on top of the explicit properties list, exactly the case where the caller wants Gemini to error on unknown input.

The current \`normalizeAdditionalProperties()\` in \`open-sse/translator/helpers/geminiHelper.ts\` unconditionally deletes the key regardless of value — so a tool author who explicitly opts into strict validation gets the opposite of what they asked for.

## Fix

\`\`\`diff
-  if (\"additionalProperties\" in record) {
+  if (record.additionalProperties === false) {
     delete record.additionalProperties;
   }
\`\`\`

| Input | Before | After |
|---|---|---|
| \`false\` | stripped | stripped (Gemini 400 otherwise) |
| \`true\` | stripped (wrong) | **preserved** (Gemini honors strict) |
| absent | unchanged | unchanged |
| non-boolean | stripped | unchanged (let Gemini surface its own error) |

Fixes #12509

### DIFF
--- a/open-sse/translator/helpers/geminiHelper.ts
+++ b/open-sse/translator/helpers/geminiHelper.ts
@@ -444,10 +444,13 @@ function normalizeAdditionalProperties(obj: unknown): void {
 
   const record = obj as JsonRecord;
 
-  // Gemini API does not support `additionalProperties` at all in function_declarations
-  // schemas (returns 400 "Unknown name"). Since Gemini defaults to allowing additional
-  // properties anyway, stripping it unconditionally is safe and prevents errors (#1421).
-  if ("additionalProperties" in record) {
+  // Gemini API function_declarations schemas reject `additionalProperties: false`
+  // with HTTP 400 "Unknown name". Strip only the `false` value (the default-permissive
+  // case Gemini already enforces). Preserve `additionalProperties: true` as a positive
+  // signal that the caller wants strict-by-default schema validation
+  // (Gemini rejects with 400 in that case too, so we keep the original error path
+  // rather than silently masking it). #12509
+  if (record.additionalProperties === false) {
     delete record.additionalProperties;
   }
 


diegosouzapw:
Thanks for digging into #12509 — but I have to close this one, because the diff argues against itself.

The change narrows the strip to `additionalProperties === false`, and the new comment says of the `true` case:

> Gemini rejects with 400 in that case too, so we keep the original error path rather than silently masking it.

That is the exact failure #1421 removed. Gemini `function_declarations` schemas reject the `additionalProperties` **key**, not a particular value — the current unconditional strip is what keeps a client that sends `additionalProperties: true` from getting a 400 `Unknown name`. Preserving it as "a positive signal that the caller wants strict validation" does not reach Gemini as a signal; it reaches Gemini as an unknown field.

If the goal is to stop silently discarding caller intent, the place for that is a translator-level warning or a `strict`-mode mapping onto something Gemini actually understands — not re-sending the field that provokes the error.

Reopening is welcome if you have an upstream response showing Gemini now accepts `additionalProperties: true`; that would change the premise and I would take it.

PR #12789, opened 2026-09-05
feat(combo): quota-weighted routing — skip empty accounts, draw by leftover

HouMinXi:
## Summary

Adds combo strategy `quota-weighted`.

Same-model account pools (Antigravity Gemini is the case that bit us) currently have two bad fits: `p2c` scores by model latency, so ten identical `modelStr` rows look the same and empty accounts stay in the draw; `reset-aware` always takes the current snapshot leader, so concurrent workers pile onto one connection until the next refresh. This strategy drops accounts that are actually empty, then draws the first target in proportion to reset-aware leftover.

`p2c` / `reset-aware` / `headroom` / `quota-share` are unchanged. New enum value, dashboard can pick it.

## Changes

- `getResetAwareRemainingPercent` next to the existing reset-aware score.
- `expandTargetsByQuotaAwareConnections` takes an optional `skipExhaustionFilter`. The 99% hard filter would wipe the 1% soft-floor pool before the orderer can use it.
- `orderTargetsByQuotaWeighted`: circuit-open and remaining=0 stay out; draw from pool A (remaining > floor, default 1%); if A is empty, draw from B; tail is leftover descending. All-zero weights fall back to uniform.
- Wired through routing constants, combo schema (`quotaWeightedFloorPercent`), dispatch, prompt-cache protect-first (otherwise the cache restack moves the weighted first target), dashboard fallback copy, 42 locales.
- Regression pin: `quota-weighted` is not in group-B connection-aware expansion. It expands itself; putting it in B double-expands.

Does **not** flip any production combo. After merge, set the combo to `quota-weighted` and `disableSessionStickiness=true`. Leave stickiness on and the first hit is pinned for the rest of the session, which is the opposite of a spread. The dashboard tips say this; the strategy does not change the global default.

## Test plan

- [x] `tests/unit/combo/quota-weighted-strategy.test.ts` (weighted draw, A/B split, skip-exhaustion expand, uniform fallback)
- [x] `tests/unit/combo/connection-aware-expansion.test.ts` T0a/T0b: `quota-weighted` stays out of group B
- [x] Injection: add `quota-weighted` to `CONNECTION_AWARE_EXPANSION_GROUP` → T0a+T0b fail; restore → pass
- [ ] GitHub CI. Tip currently fails Fast Quality Gates / some unit shards on files this PR does not touch (same fingerprints as other open PRs on `release/v3.8.51`). Not fixing those here.

## Notes

Base: `release/v3.8.51` @ `9d1a896c6`.


HouMinXi:
Local review of the product commits (the four `feat(combo)` SHAs; the follow-up T0a pin is test-only).

I stopped the run after two L1 cycles because the finding fingerprints did not overlap. That is oscillation, not a clean 3-cycle pass. I went through the ten candidates against the spec and the source instead of letting it burn another round.

None of them is a product defect:

- `acc > r` is the half-open interval the tests pin (`r=40, [40,20] → index 1`). Changing it to `>=` would fail that case.
- Pool B cannot land in the tail twice: first pick from A means B is only appended; first pick from B means the unused-B list is empty.
- Missing try/catch around the orderer is the same shape as `reset-aware`. An empty pool has to return `[]` so the combo 404s; falling back to the original list would put hard-empty accounts back in rotation.
- `toFixed(6)` rounding remaining `<5e-7` to 0 does not fire on the percentUsed path we actually score.
- Exporting `_pickWeightedIndexForTests` matches the existing `_setSecureRandomFloatSource` hook.

Wiring I checked myself (not from the review output): new strategy sits after `headroom` in the constants/dispatch tables; expander 5th arg defaults off and this strategy passes `skipExhaustionFilter: true`; remaining `=== 0` is the hard empty; floor is `Number.isFinite`, default 1; not in the group-B expansion set; `shouldProtectOriginalFirst` includes it; production combos are untouched.

The T0a assertion (`quota-weighted` stays out of group B) and the stale "20 strategies" comment are in `151c2c5bf` on this PR. I injected `quota-weighted` into `CONNECTION_AWARE_EXPANSION_GROUP` locally: T0a+T0b failed; restore passed.

HOLD, 0 confirmed product defects. CI on this base currently fails Fast Quality Gates / some unit shards on files this PR does not touch — same fingerprints as the other open PRs on `release/v3.8.51`.

HouMinXi:
Head is now `8ca90705f`.

This commit keeps session stickiness on. New conversations draw leftover / (1 + in-flight). An existing conversation stays on its account until that account is empty, then rebinds. The in-flight slot is reserved on the post-stickiness `[0]`, not during strategy ordering, so a pin cannot charge the wrong account.

Technical review of the three logic files on this SHA ran four cycles and did not converge (fingerprints kept changing). I stopped there and checked the surviving claims against the tree:

- Reserving during `orderTargetsByQuotaWeighted` would book the draw winner. Stickiness and prompt-cache then move `[0]`. The tests require `quotaShareRelease === null` at the ordering step (`quota-weighted-strategy.test.ts`). The gap between draw and reserve is the cost of booking the account that actually goes out.
- Empty `connectionId` does not share a counter. `getInflight("")` and `incrementInflight("")` are no-ops (`quotaShareInflight.ts`).
- Reusing `quotaShareRelease` is so `handleComboChat`'s existing `finally` (`combo.ts`) still releases the slot. Renaming is a follow-up, not a product bug.
- A leaked slot cannot grow forever: 4096 leases per connection and a 120s lease.
- Failover keeping the original slot elevated until settle is the same shape as quota-share (#11371). Releasing mid-retry would free a busy account for a concurrent draw.

Product defects from this SHA: 0. Tests: 27/27 on the new file, 110/110 on the sibling combo suites.

I did not change production `onmi-gemini3.6`. Cutover is a separate ops step: set strategy to `quota-weighted`, leave `disableSessionStickiness` off.

HouMinXi:

`793ce63b4` → `7a582173c`. range-diff 11 commits `=`. GPG G. No file overlap with the three landed commits.

HouMinXi:
Looked at `da86900c3` (same-tick inflight reserve) on the three logic files in that commit.

No code change from this pass. The TOCTOU fix stands. The lock is the two-pipeline race test: it fails if the same-tick `incrementInflight` is skipped. Full suite on this SHA: 34 quota-weighted + 34 quota-share.

Notes that came up, and why they do not need a patch:

1. Leak on the `dispatchSmartPipeline` early return (`targetResolution.ts:748`). Wrong order. That return is at 744, before `orderByStrategy` at 750, so the quota-weighted reserve has not happened yet. The helper also only returns a response when `strategy === "auto"`. quota-weighted never takes this path.

2. Empty `connectionId` incrementing inflight. The orderer only increments when `winnerId` is non-empty, and `incrementInflight` returns on a falsy id.

3. Release closure captures `[0]` rather than the internal winner. Today the orderer returns the pick as `[0]`, so the ids match. A later reorder after the pick would have to keep them in sync. Not a bug here.

4. `else if (!quotaShareRelease && finalId)` looks unused for quota-weighted. It is not. The orderer skips both increment and release when the drawn target has an empty `connectionId`. Stickiness / prompt-cache can still put a real id in `[0]`. That branch reserves the account that will actually be dispatched. It is also gated on `strategy === "quota-weighted"`, so it cannot invent a slot for p2c / reset-aware.

5. Idempotent `released` flag copied in three places. Same three-line guard as quota-share. Optional extract, not a defect.

6. quota-share does not transfer the slot when stickiness moves `[0]`. True of the existing quota-share path. Out of scope here. The comment at 775–776 is on purpose: transferring that slot would double-count or steal a reservation quota-share already owns. Happy to take that as a follow-up.

HouMinXi:

- `8cef05c5684` → `75be3b0df1d`
- range-diff identity (eq=13); all commits GPG `G`
- `13` unique commit(s) ahead, 0 behind
- `--force-with-lease` to `HouMinXi:feat/quota-weighted-routing`

Issue #12798, opened 2026-09-05
docs(api): Advertise per-model image (vision) support in /v1/models — and consider a gateway-level vision fallback for text-only models


aref-alapour:
## Description

Live-tested on v3.8.x (Docker, OpenAI-compatible `/v1/chat/completions`): when a request contains an OpenAI-style `image_url` content part, only the Gemini-family combos return actual vision — they transcribe screenshots correctly. Other combos (GPT-style, Claude-style, GLM, Qwen, Kimi, Grok, MiniMax heads) answer with variants of *"no vision-capable provider is connected"* / *"I'm unable to view the image"* — the image part is dropped somewhere along the chain and no error is surfaced to the client.

## Use Case

CLI coding agents (OpenCode and similar OpenAI-compatible clients) route through OmniRoute with a combo catalog, and users paste screenshots constantly ("what does this error say?"). Today the client has no way to know which combos accept images, so:

- requests to text-only combos silently lose the image;
- the agent then confidently explains "the model has no vision", which masks the routing problem instead of surfacing it.

## Proposed Solution

1. **Capability advertisement:** expose per-model input modalities on `/v1/models` (e.g. `capabilities.input_modalities: ["text", "image"]` or an `attachment` flag) so clients can attach images only where they are actually supported and degrade gracefully elsewhere.
2. **Optional gateway-level vision fallback:** when the target combo is text-only and the request contains image parts, transcribe each image with a configured vision combo (e.g. a Gemini head) and inject the transcript as a text part before forwarding. This makes every combo effectively vision-aware while keeping clients dumb — big usability win for agent workloads.
3. **Docs:** a table of which upstream head types forward images, and an explicit note that image parts are otherwise dropped silently.

## Alternatives Considered

- Client-side fallback: I currently run a small plugin in the agent that calls a vision model through the gateway and swaps the image part for its transcript. It works, but every client has to reinvent it; doing it once in the gateway makes it universal and lets the transcript quality/caching be managed centrally.

## Environment

- OmniRoute v3.8.x in Docker, OpenAI-compatible endpoint, combos over mixed upstream head types

I'd love to collaborate on this as a contributor/team member — happy to prototype the modalities advertisement on `/v1/models` or the fallback transform if there's interest in either part.

diegosouzapw:
Splitting this in two, because the halves are in very different states.

[...]

Issue #4072, opened 2026-06-17
refactor(capabilities): unify vision model-id detection into one shared source


diego-anselmo:
Follow-up to #4071.

## Problem

Three independent vision-detection lists exist and drift apart, giving the **same model up to three different vision verdicts** across three code paths:

| Source | File | Used for | Current list |
|---|---|---|---|
| `VISION_MODEL_ID_FRAGMENTS` | `src/lib/modelCapabilities.ts` | last-resort fallback in `resolveVisionCapability` (added in #4071) | conservative: pixtral, llava, bakllava, qwen-vl, qwen2-vl, qwen2.5-vl, qwen3-vl, qvq, internvl, minicpm-v, moondream, mimo-vl, kimi-vl, glm-4v, glm-4.5v, glm-4.6v, gpt-4o, gpt-4.1, gpt-4-turbo, gpt-4-vision, gemini-1.5, gemini-2, gemini-exp, claude-3, claude-opus-4, claude-sonnet-4, claude-haiku-4, mistral-medium-3, `-vision`, multimodal |
| `VISION_MODEL_KEYWORDS` | `src/app/api/v1/models/catalog.ts` (`isVisionModelId`) | `/v1/models` listing capability | broad: gpt-4o, gpt-4-turbo, claude-3, claude-3.5, claude-3-5, claude-4, claude-opus, claude-sonnet, claude-haiku, gemini, **gemma**, llava, bakllava, pixtral, mistral-pixtral, qwen-vl, qvq, glm-4.6v, glm-4.5v, vision, multimodal, **kimi** |
| `modelSupportsVision()` | `open-sse/services/compression/lite.ts` | gate that decides whether to strip images during lite compression | minimal: vision, gpt-4, 4o, claude-3, gemini, minimax-m3 |

### Concrete divergences (the actual bug)

- **`lite.ts` is missing pixtral / llava / qwen-vl / glm-4v / kimi-vl / mistral-medium-3** → for those (real vision) models lite compression **strips the image** to `[image: jpeg]`, blinding them. This is the same class of failure as #4071, on a different code path.
- **`catalog.ts` is too broad**: `gemma` (text base model) and bare `kimi` (e.g. `kimi-k2`, text) → false-positive `vision: true` in `/v1/models`.
- **`modelCapabilities.ts` is the conservative one** → a single model can be "vision" for listing, "not vision" for compression, and "unknown" for routing.

## Proposal

[...]

PR #12799, opened 2026-09-05
fix(catalog): advertise input_modalities on vision-capable combos

aref-alapour:
## Summary

Combos whose merged `capabilities.vision === true` now also advertise `input_modalities: ["text", "image"]` / `output_modalities: ["text"]` in `/v1/models`, instead of the vision boolean next to an empty modality set.

Refs #12798.

## Problem

`buildComboCatalogMetadata` merges the vision verdict from the targets' **canonical** capabilities — which honour the #9195 operator "Vision capable" override — while the combo-level modality intersection only fills when **every** known target carries **synced** modality data. An operator-flagged vision head (or any target whose canonical vision verdict comes from a source the modality path doesn't read) therefore listed:

```json
{"id":"my-vision-combo","capabilities":{"tool_calling":true,"vision":true}}
```

with no `input_modalities`. models.dev-shaped clients that key off `input_modalities` (not the boolean) — e.g. agent CLIs building their model picker from `/v1/models` — still saw a text-only model and silently degraded image requests to placeholders. Observed live on a v3.8.x gateway: vision combos advertised `vision: true` with no modalities.

## Change

- `catalogHelpers.ts` — new `visionDerivedModalities(capabilities, syncedInput, syncedOutput)`: emits the synced intersection when present, else derives `["text","image"]` / `["text"]` from a `vision === true` verdict. Makes no new claim: `mergeComboCapabilities` only emits `vision: true` when **every** known target is vision-capable, so the derived modalities are exactly as strong as the boolean already advertised. Unknown / text-only verdicts derive nothing (fail-closed, same discipline as #4071 / #4072).
- `catalog.ts` — the two modality spreads in `buildComboCatalogMetadata` collapse into the helper call (with the synced values passed through), so the frozen file stays at its current LOC for the file-size ratchet.

## Testing

- New regression test in `tests/unit/models-catalog-combo-metadata.test.ts`: an operator-flagged custom vision head (`supportsVision: true`, synced capability row with limits but no modality data) in a single-target combo → the combo entry now carries `capabilities.vision: true` **and** `input_modalities: ["text","image"]`.
- Full file: 13/13 pass (the pre-existing "single-target combo preserves its direct model metadata" guards the synced-passthrough behaviour).
- `catalog-helpers-extraction` + `models-catalog-route`: 58/58 pass.
- `check:file-size` OK (catalog.ts unchanged in LOC), `prettier --check` clean.

I'm interested in contributing here beyond this patch (see the team-collab note in #12798) — happy to take on the `/v1/models`-side follow-ups from that issue, or adjust this approach if maintainers prefer deriving modalities at the target level instead of the combo level.

### DIFF
--- a/changelog.d/fixes/12799-combo-vision-derived-modalities.md
+++ b/changelog.d/fixes/12799-combo-vision-derived-modalities.md
@@ -1,0 +1,1 @@
+- `/v1/models` combos whose merged `capabilities.vision` is `true` now also advertise `input_modalities: ["text","image"]` / `output_modalities: ["text"]` (synced modality intersections keep precedence), so models.dev-shaped clients no longer see a vision combo as text-only. (#12799 — thanks @aref-alapour)
--- a/src/app/api/v1/models/catalog.ts
+++ b/src/app/api/v1/models/catalog.ts
@@ -97,6 +97,7 @@ import {
   maybeOmitCatalogModelName,
   getThinkingCapabilityFields,
   mergeComboCapabilities,
+  visionDerivedModalities,
   getConnectionScopedEffortTiers,
   type ConnectionScopedReasoningCatalog,
 } from "./catalogHelpers";
@@ -792,8 +793,7 @@ async function buildUnifiedModelsResponseCore(
         ...(contextLength ? { context_length: contextLength } : {}),
         ...(maxInputTokens ? { max_input_tokens: maxInputTokens } : {}),
         ...(maxOutputTokens ? { max_output_tokens: maxOutputTokens } : {}),
-        ...(inputModalities.length > 0 ? { input_modalities: inputModalities } : {}),
-        ...(outputModalities.length > 0 ? { output_modalities: outputModalities } : {}),
+        ...visionDerivedModalities(capabilities, inputModalities, outputModalities), // #12798
         ...(Object.keys(capabilities).length > 0 ? { capabilities } : {}),
       };
     };
--- a/src/app/api/v1/models/catalogHelpers.ts
+++ b/src/app/api/v1/models/catalogHelpers.ts
@@ -217,3 +217,37 @@ export function mergeComboCapabilities(
   }
   return capabilities;
 }
+
+/**
+ * #12798: a combo can advertise `capabilities.vision: true` while emitting no
+ * `input_modalities` / `output_modalities`. The merged vision verdict flows
+ * from the targets canonical capabilities - which honour the #9195 operator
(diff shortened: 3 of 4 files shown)


diegosouzapw:
The root-cause diagnosis here is solid and the fix (`visionDerivedModalities()`, deriving modalities from an already-advertised vision verdict, fail-closed for unknown/text-only) is exactly right in the code I read. I wasn't able to get a clean green run locally though, in two independent attempts — once against your PR head, and once applying just your `catalog.ts`/`catalogHelpers.ts`/test-file diff onto a fresh `release/v3.8.51` checkout to rule out merge-base drift. Both runs show most of the file's other pre-existing tests failing too (looks like registry-value drift for those), but your own new test (`vision-flagged combo derives input modalities...`) also fails in both — with `combo entry missing from /v1/models`, which reads more like a fixture/setup issue in the test than a problem with your `visionDerivedModalities` logic itself. Given the PR body reports 13/13 and 58/58 passing, and my local devbox was under heavy unrelated load (other sessions running tests concurrently) this session, I'd like a fresh CI run (or a retry on a quieter box) to confirm before merge rather than trusting my local repro as the final word either way. Happy to take another look once that's in hand.

aref-alapour:
Thanks for the thorough re-check and for isolating the patch against a fresh `release/v3.8.51` checkout as well — that’s really helpful.

Your observation around the new regression test makes sense. Since the failure is `combo entry missing from /v1/models` rather than an assertion on the derived modalities themselves, I agree this points more toward the fixture/setup path than `visionDerivedModalities()`.

I’ll re-run the relevant test set in a clean environment and verify specifically:

- the combo fixture is actually registered and reaches `/v1/models`
- the operator-flagged vision target is present in the catalog setup
- the new regression test remains deterministic in isolation and as part of the full file
- the previously reported `13/13` and `58/58` results still reproduce from the current PR head

If I can reproduce the missing combo locally, I’ll fix the fixture/setup rather than touching the modality derivation logic unless the evidence points there.

I’ll push any necessary test-only adjustment and post the clean CI results here once they’re available. Thanks again for taking the time to validate this independently.

diegosouzapw:
Thanks @aref-alapour — merging via the release merge-train. Validated in local merge-train /tmp/mt-train3b.log on 192.168.0.113 @ train tip 408e2128791696a18966681953136fc96aeb99b0 (32 PRs boarded): static gates green; full test:unit 40694 tests, 17 failing — every one reproduces on the pure release tip (base-red sweep list), zero new reds. Merged --admin per merge-gates §4/§7.

PR #12801, opened 2026-09-05
feat(release): bun-pack — bun-first global installs (build stays on Node)

opensource-elearning:
⚠️ base-red inherited: #12732 (base tip is red; unrelated to this PR)

## What

Adds a bun-first companion to the npm release path so users can install
omniroute globally with bun, without changing how the package is built or
published.

- `scripts/release/bun-pack.mjs` — release pack script, Bun-first:
  - default `--pm bun` → `bun run build:release` (bun orchestrates, Node runs
    the actual `next` build and the dist assembly), then `bun pm pack` into a
    universal npm tarball.
  - `--pm npm` fallback → same flow via `npm` (works on machines without bun).
  - `--skip-build` reuses a staged `dist/`.
  - cross-platform: no shell string interpolation (paths/env travel as spawn
    options), bun `>=1.1`, node `>=22`.
- `package.json` → `bun:release` script ( `bun scripts/release/bun-pack.mjs` ).
- `README.md` → Bun (global) row in the install table.

## Why

`bun pm pack` and `npm pack` produce the same universal npm tarball format, so
a bun-packed artifact installs with **both** `bun add -g <tarball>` and
`npm install -g <tarball>`. Bun is measurably faster on installs and lighter on
memory, and `bun install -g omniroute` is already a supported path (the
best-effort `bun:sqlite` adapter lets it boot without `better-sqlite3`). The
npm publish channel (`npm-publish.yml`, `verify-published.mjs`) is unchanged —
this is strictly additive: bun is only packer + install path, never the build
or the published runtime (AGENTS.md scope guard).

## Verification

- Fresh production release build of 3.8.51 (webpack, `OMNIROUTE_BUILD_MEMORY_MB`
  heap cap) completed green in an isolated worktree.
- Packed with **both** `--pm bun` and `--pm npm`: same tarball, 4168 tracked
  `dist/` entries verified in each, `bin/omniroute.mjs`,
  `bin/reset-password.mjs`, and `package.json` present.
- Output: `_artifacts/omniroute-3.8.51.tgz` (~73 MB).

```bash
bun run bun:release            # build + pack (Bun, default)
bun run bun:release -- --pm npm  # npm fallback
node scripts/release/bun-pack.mjs --skip-build --destination .  # reuse staged dist

### DIFF
--- a/README.md
+++ b/README.md
@@ -722,6 +722,7 @@ of your shell history. → [CLI Integrations](docs/guides/CLI-INTEGRATIONS.md)
 <table>
   <tr><th align="left">Platform</th><th align="left">Install</th><th align="left">Highlights</th></tr>
   <tr><td align="left" nowrap>📦 <b>npm (global)</b></td><td align="left" nowrap><code>npm install -g omniroute</code></td><td align="left">One command, any OS</td></tr>
+  <tr><td align="left" nowrap>⚡ <b>Bun (global)</b></td><td align="left" nowrap><code>bun add -g omniroute</code></td><td align="left">Same registry tarball, faster installs</td></tr>
   <tr><td align="left" nowrap>🐳 <b>Docker</b></td><td align="left" nowrap><code>docker run … diegosouzapw/omniroute</code></td><td align="left">Multi-arch <b>AMD64 + ARM64</b></td></tr>
   <tr><td align="left" nowrap>🖥️ <b>Desktop (Electron)</b></td><td align="left" nowrap><code>npm run electron:build</code></td><td align="left">Native window + system tray — <b>Windows / macOS / Linux</b></td></tr>
   <tr><td align="left" nowrap>🎩 <b>Menu-bar (OmniRouteTray)</b></td><td align="left" nowrap><code>brew install --cask zoispag/tap/omniroute-tray</code></td><td align="left">Supervises &amp; auto-updates the server — <b>macOS</b></td></tr>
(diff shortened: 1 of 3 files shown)


diegosouzapw:
Nice, and careful about the Bun-scope boundary — verified that `bun run build:release`
still shells out to literal `npm run build` / `npm run build:cli` internally, so the actual
Next build and prepublish step stay on Node as intended; only the packer/orchestrator is
Bun-first. No shell-interpolation issues either (spawnSync with array args throughout).
One gap: there's no unit test for the script's own logic (arg parsing, tarball verification).
Could you extract `parseArgs`/`verifyTarball` so they're testable without needing a real build,
and add a small test file? Everything else looks solid.

diegosouzapw:
Follow-up on the one pre-merge item left here (a unit test for `parseArgs`/`verifyTarball`, Hard
Rule #8): we run a maintainer fix-sweep that pushes small pre-merge fixes straight into the
contributor's branch (authored by you, with a `Co-authored-by` for the maintainer), but this fork
has **"Allow edits by maintainers" unchecked**, so we can't land it for you.

Two ways forward, whichever you prefer:

1. Tick "Allow edits by maintainers" in the PR sidebar and reply here — we'll push the test
   ourselves (extracting `parseArgs`/`verifyTarball` so they're importable without running a real
   build) and nothing else.
2. Push it yourself: a `tests/unit/` file that imports those two functions directly and covers the
   arg parsing plus a tampered/short tarball rejection. No need to exercise a real `npm pack`.

Everything else on this PR came out clean in review (⭐4) — the Bun-scope boundary is respected
(`build:release` still shells out to `npm run build*`), so this is the last thing between it and
the merge queue.

opensource-elearning:
done, try now

PR #12803, opened 2026-09-05
fix(claude): extra-usage switch does not skip 5h preflight

HouMinXi:
## What

A Claude connection with extra usage enabled (`blockExtraUsage: false`) was skipped in quota preflight once the 5h session bar dropped to 1% remaining (factory cutoff is 2%). The request never reached Anthropic, so extra credits stayed unused and the combo returned `ALL_TARGETS_SKIPPED` 503.

That switch only controls whether extra usage is allowed. It is not a reason to hide the account before dispatch.

## Why

Reproduced on a Max 20x account with extra usage on and unused extra credits. Same combo returned 200 after the 12:50Z window reset.

## Change

- `isClaudeExtraUsageAllowed`: true only when `provider === "claude"` and `blockExtraUsage === false`
- `evaluateQuotaCutoff` / `preflightQuota`: proceed in that case
- `isQuotaExhaustedForRequest`: optional `providerSpecificData`; extra-usage Claude is not treated as exhausted for routing
- `evaluateQuotaLimitPolicy`: same opt-in does not hide the account
- Combo auto-candidate cutoff and quota-aware expansion forward `providerSpecificData`

Default (`blockExtraUsage` unset or true) is unchanged. Other providers do not take this path.

## Tests

```
node --import tsx/esm \
  --test --test-force-exit \
  tests/unit/quota-preflight.test.ts \
  tests/unit/claude-extra-usage.test.ts \
  tests/unit/antigravity-quota-skipping.test.ts \
  tests/unit/quota-policy-generalization.test.ts
```

32/32 pass. Removing the cutoff bypass turns the two new extra-usage tests red.

## Not in this PR

- Spending extra credits before the 5h window is actually exhausted
- Changing the dashboard extra-usage switch polarity
- quota-share saturation buckets (fail-open; not used by this combo)


HouMinXi:

`d7ac57dc0` → `15f6a7172`. range-diff 2 commits `=`. GPG G. No file overlap with the three landed commits.

HouMinXi:

- `15f6a71728c` → `cb28be5f280`
- `--force-with-lease` to `HouMinXi:fix/claude-extra-usage-preflight`

PR #12770, opened 2026-09-05
fix(glm): drop extra 16th arg to SSE transform helper

HouMinXi:
`translateSseResponse` in `open-sse/executors/glm.ts` passed a 16th positional (`65536`) to `createSSETransformStreamWithLogger`. That helper has 15 parameters (`open-sse/utils/stream.ts:2998-3013`); last is `requestToolIdentityMap`. tsc reports TS2554. The number never reached `TransformStream`. The old comment about a 64KB buffer was wrong.

This is the same one-line drop that landed inside #12711. Isolated here so pin / combo-split / moonshot do not have to carry `glm.ts`.

## Fix

Drop `undefined, undefined, 65536`. Last argument stays `suppressThinkClose`. Comment now matches the helper.

## Tests

- `tests/unit/glm-sse-transform-arity.test.ts` — source guard: helper has no `highWaterMark` slot; GLM call site has no `65536` and ends on `suppressThinkClose`.
- `tests/unit/glm-think-close-marker-leak.test.ts` still green (3/3).
- Injection: put `65536` back → arity test red. Restore → 2/2 green.

`KeyHealth` / `translateJsonResponse` unused on tip are not in this diff.

## Not in this PR

- changelog fragment `reset-aware-model-family.md` missing `- ` (#12732)
- `cli-tunnel` skill dry-run generate
- batch-delete ConfirmModal (#12711)

Contributor only — not merging.

PR #12805, opened 2026-09-05
feat(grok-cli): show and redeem banked reset credits on Provider Limits

HouMinXi:
## What

grok.com Usage already shows Reset Available. The Omni Provider Limits card did not.

This PR reads `POST grok.com/prod_mc_billing.ConsumerUiSvc/GetRemainingResets` with the same grok-cli bearer that chats, and writes `bankedResetCredits` onto grok-cli usage.

- Empty DATA frame + `grpc-status: 0` is a real zero. The row still draws.
- HTTP / trailer failure omits the field. It does not write 0.
- Repeated field 10 is collected as an array. The existing Map walker in `grokCliQuotaFrame` would keep only the last card.

View credits on a grok-cli card with inventory now POSTs `ConsumerUiSvc/RedeemReset` (protobuf field 10 `token_id`) through the existing `/api/usage/codex-reset-credit` route. Token ids stay on the server. Unknown providers return 400 instead of falling through to Codex.

Live probe used a zero-inventory account and a fake token. No real card was consumed. Field 10 + fake id returns grpc-status 9 ("does not exist or is expired"). Field 1 or an empty body returns grpc-status 3 ("Invalid token_id").

## Tests

```
DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=8192 --import tsx/esm \
  --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts \
  --test --test-force-exit \
  tests/unit/grok-reset-credits-frame.test.ts \
  tests/unit/grok-cli-provider-limits.test.ts \
  tests/unit/provider-limits-ui.test.ts \
  tests/unit/grok-reset-credits-redeem.test.ts \
  tests/unit/grok-reset-credits-connection.test.ts
```

73 passed (2026-09-05), including: empty DATA → count 0; two field-10 messages → count 2; 404 omits the key; grok-cli `bankedResetCredits: 0` draws a row; grok-cli with inventory can redeem; a given `tokenId` skips GetRemainingResets; unknown provider is 400.

## Not in this PR

- grok-web / xai-oauth
- a new `/api/usage/grok-reset-credit` endpoint
- flipping any production combo

## ⚠️ base-red inherited

Same three gates as #12789 / #12697 / #12767. None of the files below are in this PR's diff.

- API Route Typecheck: extra 16th arg in `open-sse/executors/glm.ts` (isolated as #12770)
- Merge integrity: tip `changelog.d/fixes/reset-aware-model-family.md` missing leading `- `
- Docs Gates: same tip fingerprint (`check:agent-skills-sync` dry-run `GENERATED: + cli-tunnel` class)

This PR's own fragment starts with `- `. Local `check-file-size --base-ref upstream/release/v3.8.51` is green. ESLint / Vitest / semgrep / Change Classification are green on `dd263fed6`.

### DIFF
--- a/changelog.d/fixes/grok-cli-reset-credits-readonly.md
+++ b/changelog.d/fixes/grok-cli-reset-credits-readonly.md
@@ -1,0 +1,1 @@
+- **feat(grok-cli):** Provider Limits now shows grok-cli banked reset credits from `GetRemainingResets`, including a real zero inventory; a failed RPC omits the row instead of faking 0, and grok-cli still does not get the Codex redeem button
--- a/changelog.d/fixes/grok-cli-reset-credits-readonly.md
+++ b/changelog.d/fixes/grok-cli-reset-credits-readonly.md
@@ -1,1 +1,1 @@
-- **feat(grok-cli):** Provider Limits now shows grok-cli banked reset credits from `GetRemainingResets`, including a real zero inventory; a failed RPC omits the row instead of faking 0, and grok-cli still does not get the Codex redeem button
+- **feat(grok-cli):** Provider Limits shows grok-cli banked reset credits from `GetRemainingResets` (including a real zero; a failed RPC omits the row) and the existing View credits button now calls `ConsumerUiSvc/RedeemReset` for grok-cli
--- a/changelog.d/fixes/grok-cli-reset-credits-readonly.md
+++ b/changelog.d/fixes/grok-cli-reset-credits-readonly.md
@@ -1,1 +1,1 @@
-- **feat(grok-cli):** Provider Limits shows grok-cli banked reset credits from `GetRemainingResets` (including a real zero; a failed RPC omits the row) and the existing View credits button now calls `ConsumerUiSvc/RedeemReset` for grok-cli
+- **feat(grok-cli):** Provider Limits shows grok-cli banked reset credits from `GetRemainingResets` (including a real zero; a failed RPC omits the row) and the existing View credits button now calls `ConsumerUiSvc/RedeemReset` for grok-cli. Live tokens use nested fields 10/20/30 (id + Timestamp), not compact 1/2/3.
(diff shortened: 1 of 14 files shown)


HouMinXi:
Pushed `dd263fed6` on top of the read-only commits already on this PR.

That commit wires View credits for grok-cli to `ConsumerUiSvc/RedeemReset` (field 10 `token_id`) through the existing `/api/usage/codex-reset-credit` route. Token ids stay server-side. A given `tokenId` skips the inventory RPC. Unknown providers return 400 instead of Codex.

Live probe: zero-inventory account, fake token, no real card consumed. Field 10 + fake id → grpc-status 9. Field 1 / empty → grpc-status 3.

Local tests: 73/73 on the five grok reset-credit files.

HouMinXi:

`dd263fed6` → `e4f12f420`. range-diff all five commits `=`. GPG G. Local tests 63/63 on the grok reset-credit files. `check-file-size --base-ref f9a1cc8a9` green. No file overlap with the three landed commits.

HouMinXi:
Follow-up `42dbd1577`: live SuperGrokPro `GetRemainingResets` tokens use nested protobuf 10/20/30 (13-byte id + Timestamp), not compact 1/2/3. The old decoder treated that payload as malformed, so `bankedResetCredits` was omitted and Provider Limits never drew the Reset credits row even when the RPC returned a card.

Decoder now accepts both shapes. Compact 1/2/3 fixtures still pass. Token ids stay server-side. The lock is `tests/unit/grok-reset-credits-frame.test.ts` (`live nested fields 10/20/30 are not malformed`): it goes red if field 10 is dropped from the id list.

HouMinXi:

- `42dbd15779c` → `27e050a5495`
- range-diff identity (eq=6); all commits GPG `G`
- `6` unique commit(s) ahead, 0 behind
- `--force-with-lease` to `HouMinXi:feat/grok-reset-credits-readonly`

Issue #12807, opened 2026-09-05
fix(docker): Webpack stub aliasing in next.config.mjs breaks for any import not literally at project root


alpha-code2019:
### Environment
- Branch: `main`
- Build: `docker build --build-arg OMNIROUTE_BUILD_PROFILE=minimal --build-arg OMNIROUTE_USE_TURBOPACK=0 .`

### Problem
In `next.config.mjs`, the `webpack.NormalModuleReplacementPlugin` callback sets:
```js
resource.request = "./src/lib/cloudSync.stub.ts";
```
`resource.request`, when relative, resolves against the *importing file's* directory -- not the project root. Since the real `@/lib/cloudSync` import happens from files scattered across `src/app/api/**`, this only ever resolved correctly for imports literally sitting at the project root. Every other import fails with `Module not found: Can't resolve './src/lib/cloudSync.stub.ts'` (same pattern for the other 3 stubbed modules).

### Fix
Use an absolute path instead -- `next.config.mjs` already defines `projectRoot` via `dirname(fileURLToPath(import.meta.url))` for other purposes; reusing it here (`join(projectRoot, "src/lib/cloudSync.stub.ts")`, etc.) resolves it correctly regardless of the importing file's location. I have a working patch for this too, happy to open a PR.

diegosouzapw:
Confirmed, and the mechanism is the one you describe: in the webpack branch, `NormalModuleReplacementPlugin` receives a **relative** `resource.request`, which webpack resolves against the *importing file's* directory rather than the project root. That works only for importers that happen to sit at the right depth, and breaks for any import elsewhere.

The fix shape is to make the replacement absolute — `join(projectRoot, stubPath)` — for the replacements in that block.

Open PR **#12810** already contains this alongside the Turbopack half, but it is not mergeable as it stands: it targets `main` instead of the active release branch, carries **5100 changed files**, and is `CONFLICTING`. The handful of relevant lines need to be reapplied on a branch cut from the current release tip.

One clarification since both issues point at the same PR: this and #12806 are **two distinct defects**, one per bundler — #12806 is missing exports in the stub modules (Turbopack), this one is path resolution (webpack). Fixing either alone leaves the other build profile broken, so they are best done in one PR but must both be addressed.

Issue #12806, opened 2026-09-05
fix(docker): OMNIROUTE_BUILD_PROFILE=minimal fails to build under Turbopack -- stub modules missing exports real routes import


alpha-code2019:
### Environment
- Branch: `main` (also reproduced on `release/v3.8.49`)
- Build: `docker build --target runner-base --build-arg OMNIROUTE_BUILD_PROFILE=minimal .` (Turbopack, the project's stated default)

### Problem
Building with `OMNIROUTE_BUILD_PROFILE=minimal` fails with 9 compile errors of the form `Export X doesn't exist in target module`, because the stub files aliased in under Turbopack's `resolveAlias` don't implement the full export surface the real routes import.

Specifically:
- `src/lib/services/installers/ninerouter.stub.ts` exports `installNinerouter` (no route imports this name) but is missing `getInstalledVersion`, `getLatestVersion`, `update`, and `uninstall` -- all imported by real routes: `app/api/services/9router/{status,install,update}/route.ts`. The real module's function is named `install`, not `installNinerouter`.
- `src/mitm/cert/install.stub.ts` is missing `installCertResult`, imported by `app/api/tools/agent-bridge/{server,cert}/route.ts`.

### Result
`OMNIROUTE_BUILD_PROFILE=minimal` cannot currently produce a working build via the Turbopack path at all -- only the webpack path (`OMNIROUTE_USE_TURBOPACK=0`) works, and only coincidentally (see the separate issue I'm filing on that path's own bug).

### Fix
Add the missing exports to both stub files, matching the real modules' exact names/signatures, each still just throwing `FeatureDisabledError` -- no privileged functionality restored. Happy to open a PR with this if useful; I already have a working patch.

diegosouzapw:
Confirmed. `src/lib/services/installers/ninerouter.stub.ts` exports only `installNinerouter` (which nothing imports) and `resolveSpawnArgs`, while the real module exports `getInstalledVersion`, `getLatestVersion`, `install`, `update` and `uninstall` — all of them imported by the 9router service routes. `src/mitm/cert/install.stub.ts` has the same gap, missing `installCertResult` used by the agent-bridge routes.

Under Turbopack's `resolveAlias` a missing export is a hard compile error rather than a feature-disabled build, which is exactly the failure you hit.

[...]

PR #12810, opened 2026-09-05
Fix OMNIROUTE_BUILD_PROFILE=minimal: complete stub exports, fix webpack path resolution

alpha-code2019:
Fixes both bugs reported in #12806 and #12807.

**#12806** -- stub modules (`ninerouter`, `mitm/cert/install`) were missing named exports real routes import (`getInstalledVersion`, `getLatestVersion`, `update`, `uninstall`, `installCertResult`), and one had the wrong function name (`installNinerouter` instead of `install`). Fixed to match the real modules' exact names/signatures -- every stub still only throws `FeatureDisabledError`, no privileged functionality restored.

**#12807** -- `NormalModuleReplacementPlugin`'s `resource.request` used a relative path that only resolved from the project root. Fixed using the `projectRoot` helper `next.config.mjs` already defines for other purposes.

Verified with full, real builds (`docker build --target runner-base --build-arg OMNIROUTE_BUILD_PROFILE=minimal`, both with and without `OMNIROUTE_USE_TURBOPACK=0`) -- confirmed the real privileged-code identifiers are absent from the compiled server output and `FeatureDisabledError` is correctly wired into the real route handlers, not just that the build succeeds.

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -1,6 +1,6 @@
 import createNextIntlPlugin from "next-intl/plugin";
 import { createMDX } from "fumadocs-mdx/next";
-import { dirname } from "node:path";
+import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { betterSqlite3AliasFor } from "./scripts/build/better-sqlite3-stub-flag.mjs";
 import { mitmManagerAliasFor } from "./scripts/build/mitm-stub-flag.mjs";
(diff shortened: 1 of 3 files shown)


diegosouzapw:
Verified both root causes: #12806 and #12807 are indeed still open, and I confirmed on the
current tip that the stub exports and the relative `resource.request` path are exactly as
broken as described — your new export names/signatures match the real modules exactly. This
looks solid; the one gap is an automated regression test (you validated with real
`docker build` runs, which is great, but a unit test asserting export parity between stub and
real module would keep this from silently drifting again). Please add that, rebase onto
`release/v3.8.51`, and this should be close to merge-ready.

PR #12813, opened 2026-09-05
fix(compression): terminate idle workers on pool eviction

### LINKED ISSUE #12812: fix(backend): Compression worker pool leaks worker threads: idle eviction never terminates (5.7GB commit / 55 orphaned MessagePorts after 16h)
## Summary

`CompressionWorkerPool.finish()` drops idle workers from the pool **without terminating them**. The worker thread, its `MessagePort` and its private heap stay alive for the entire process lifetime. On a long-running instance this accumulates until the process holds several GB of commit charge that no V8-level metric reports.

Observed on a 16-hour instance: **5,723 MB private memory, 69 OS threads, 55 leaked `MessagePort` handles** — while `process.memoryUsage()` still looked healthy at `rss=660MB`.

## Root cause

`open-sse/services/compression/compressionWorkerPool.ts:133` — the idle-eviction timer passes `terminate = false`:

```ts
private finish(slot: PoolWorker, result: CompressionResult): void {
  // ...
  slot.idle = setTimeout(() => void this.remove(slot, false), this.idleMs);
  //                                                   ^^^^^ never terminates
  slot.idle.unref();
  this.dispatch();
}
```

`remove()` only terminates when that flag is set:

```ts
private async remove(slot: PoolWorker, terminate: boolean): Promise<void> {
  if (!this.workers.delete(slot)) return;      // forgotten by the pool
  if (terminate) await slot.worker.terminate().catch(() => undefined);  // skipped
}
```

So after `idleMs` (default 60s) the slot is deleted from `this.workers` but the underlying `Worker` is never terminated. The pool loses its only reference, so nothing can ever terminate it afterwards. Because `size` defaults to 2, every burst of compression traffic separated by more than 60s of idle time strands up to 2 more workers.

The error path is correct — `fail()` calls `remove(slot, true)`, and `close()` uses `remove(e, true)`. Only the success path leaks.

## Evidence

Measured against the running process (v3.8.50, Node v24.19.0, Windows 11, `--max-old-space-size=4096`):

| Metric | OmniRoute (16.2h uptime) | Plain Node baseline |
|---|---|---|
| Private / commit | **5,723 MB** | 33 MB |
| OS threads | **69** | 12 |
| `MessagePort` handles | **55 (all refed)** | — |

`process._getActiveHandles()` on the live process:

```
MessagePort   55      <- leaked workers
Server         3
TLSSocket      2
FSWatcher      1
```

All 55 report `hasRef() === true`, so they are not unreffed leftovers.

A `VirtualQueryEx` walk of the process address space attributes the memory to many small independent heaps rather than one large cache — the signature of N orphaned worker isolates:

```
Private/READWRITE       5,432.0 MB   <- 12,534 regions
Mapped/READONLY           264.2 MB   <- SQLite mmap_size, expected
Image/*                   137.1 MB
TOTAL COMMIT            6,064.4 MB

region size distribution:
  64 KB - 1 MB    11,737 regions    2,931 MB
  1 - 8 MB           622 regions    2,312 MB
  8 - 64 MB           20 regions      190 MB
  > 64 MB              0 regions        0 MB
```

Note the diagnostic trap: `process.memoryUsage()` reports `rss=660MB / heapUsed=414MB` because V8 only accounts for the main thread's heap. The orphaned worker isolates are invisible there — only OS-level commit charge reveals them.

Ruled out during investigation: sql.js WASM fallback (native `better-sqlite3` loads fine), SQLite `mmap_size`/`cache_size` (264 MB, matches configuration), and per-request leakage (300 sequential `/status` requests changed handle count by −1).

## Reproduction

1. Start OmniRoute and send compression traffic in bursts separated by more than `OMNI_COMPRESSION_WORKER_IDLE_MS` (default 60s).
2. After each idle expiry, inspect the process:

```js
process._getActiveHandles().filter(h => h?.constructor?.name === 'MessagePort').length
```

3. The count grows by up to `OMNI_COMPRESSION_WORKERS` (default 2) per burst and never decreases. OS thread count and commit charge track it.

## Suggested fix

```diff
-    slot.idle = setTimeout(() => void this.remove(slot, false), this.idleMs);
+    slot.idle = setTimeout(() => void this.remove(slot, true), this.idleMs);
```

Idle eviction should terminate the worker — that is the entire point of evicting it. `remove()` already handles the async terminate safely, and `fail()`/`close()` already pass `true`.

Alternatively, make `remove()` always terminate and drop the boolean parameter, since no caller has a legitimate reason to delete a slot while leaving its thread running.

## Environment

- OmniRoute 3.8.50 (latest at time of report)
- Node v24.19.0
- Windows 11
- `NODE_OPTIONS`/execArgv: `--max-old-space-size=4096`
- Default pool settings (`OMNI_COMPRESSION_WORKERS` unset → 2, `OMNI_COMPRESSION_WORKER_IDLE_MS` unset → 60000)

anhtahaylove:
Fixes #12812.

## Problem

`CompressionWorkerPool.finish()` schedules idle eviction with `remove(slot, false)`. `remove()` deletes the slot from `this.workers` but only terminates the thread when the flag is `true`:

```ts
private async remove(slot: PoolWorker, terminate: boolean): Promise<void> {
  if (!this.workers.delete(slot)) return;      // pool drops its only reference
  if (terminate) await slot.worker.terminate().catch(() => undefined);
}
```

So an idle worker is forgotten by the pool while its thread, `MessagePort` and private heap stay alive — and since the reference is gone, nothing can terminate it afterwards. Every burst of compression traffic separated by more than `idleMs` (default 60s) strands up to `OMNI_COMPRESSION_WORKERS` (default 2) more threads.

`fail()` and `close()` already pass `true`. Only the success path leaked.

## Measured impact

On a 16.2h instance of v3.8.50 (Node v24.19.0, Windows 11):

| Metric | Leaking process | Plain Node baseline |
|---|---|---|
| Private / commit | **5,723 MB** | 33 MB |
| OS threads | **69** | 12 |
| `MessagePort` handles | **55, all refed** | — |

`process._getActiveHandles()` showed 55 `MessagePort` entries against only 3 `Server` and 2 `TLSSocket`. An address-space walk attributed 5,432 MB to 12,534 small `Private/READWRITE` regions with no allocation larger than 21 MB — many independent worker heaps rather than one large cache.

Worth noting for anyone triaging similar reports: `process.memoryUsage()` reported `rss=660MB / heapUsed=414MB` throughout, because V8 only accounts for the main thread's isolate. The leak is invisible to every JS-level metric and only shows up in OS commit charge.

## Change

```diff
-    slot.idle = setTimeout(() => void this.remove(slot, false), this.idleMs);
+    slot.idle = setTimeout(() => void this.remove(slot, true), this.idleMs);
```

Idle eviction now terminates the worker, which is the purpose of evicting it. `remove()` already handles the async terminate safely with a `.catch()`.

## Verification

I applied the equivalent one-character patch to the built bundle actually loaded by my running instance (`dist/.build/next/server/chunks/open-sse_services_compression_compressionWorkerPool_ts_13s5ju7._.js`, confirmed via CDP `Debugger.scriptParsed`) and verified it parses with `node --check`. Behavioural confirmation requires a restart with the rebuilt bundle, so I have not yet measured post-fix handle counts over a long uptime — happy to report back once I have a multi-hour sample.

A follow-up worth considering separately: `remove()`'s boolean parameter now has no caller passing `false`, so it could be dropped entirely to make the leak unrepresentable.

### DIFF
--- a/open-sse/services/compression/compressionWorkerPool.ts
+++ b/open-sse/services/compression/compressionWorkerPool.ts
@@ -185,7 +185,7 @@ export class CompressionWorkerPool {
     slot.timeout = null;
     slot.job = null;
     job.resolve(result);
-    slot.idle = setTimeout(() => void this.remove(slot, false), this.idleMs);
+    slot.idle = setTimeout(() => void this.remove(slot, true), this.idleMs);
     slot.idle.unref();
     this.dispatch();
   }


anhtahaylove:
## Post-patch soak data from the reporting instance

Running the one-line fix from this PR on the instance described in #12812. Reporting what I actually measured, including where the evidence is thin.

### Thread count — the direct proxy for live workers

Each pooled worker is one thread, so thread count tracks orphaned workers without needing a debugger attached.

| | before fix | after fix |
|---|---|---|
| threads | 69–70 | **15–16** |
| private / commit | 5.60–5.95 GB | 0.64–0.94 GB |
| MessagePort (CDP) | 55, all refed | 2 |

11 samples over ~2h of uptime, same PID (16244) throughout, `/status` 200 on every sample. Thread count never left 15–16.

### Private memory is not climbing

| uptime (min) | 13 | 21 | 44 | 75 | 106 |
|---|---|---|---|---|---|
| private (MB) | 641 | 902 | 904 | 722 | 883 |

It oscillates and comes back down — 4 of 10 sample-to-sample deltas are decreases. Before the fix the same metric only went up, ending at 5.9 GB after ~16h.

### Honest limits of this data

Two caveats worth stating plainly, because the numbers above look better than the evidence strictly supports:

1. **~2h of uptime, not 16h.** The original leak needed a long soak to become obvious. This window is too short to call the leak eliminated over a full day.
2. **Light worker load.** 556 compressions ran in this window, but 550 were `lite` mode, which is not worker-eligible (`isCompressionWorkerEligible` only admits `standard` / `rtk` / `stacked`). Only **6** runs actually exercised the pool. So this window demonstrates a healthy baseline, not a heavy worker-churn stress test.

The strongest evidence for the mechanism therefore remains the direct measurement in #12812: 55 refed `MessagePort` handles before the patch versus 2 after, matching the default `OMNI_COMPRESSION_WORKERS`.

Sampling continues; I will follow up if anything regresses.

anhtahaylove:
## Deterministic reproduction: the pool leaks one thread per worker per idle cycle

My earlier comment relied on a production soak where only 6 requests were actually worker-eligible. That was weak evidence, so I built a direct harness around the pool instead of waiting on organic traffic.

**Setup.** `CompressionWorkerPool` is instantiated straight from source with `idleMs` lowered to 150 ms so idle eviction fires quickly, plus a stub `compressionWorker.js` that answers the real protocol (`{id, type:"result", result}`). Each cycle submits `size * 3` jobs, waits past the idle threshold, then counts live `MessagePort` handles — one per running worker thread.

### 10 cycles, pool size 4

| cycle | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
|---|---|---|---|---|---|---|---|---|---|---|
| **before** (`remove(slot, false)`) | 4 | 8 | 12 | 16 | 20 | 24 | 28 | 32 | 36 | **40** |
| **after** (this PR) | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | **0** |

Exactly `+size` orphaned threads per idle cycle, perfectly linear — the whole pool is evicted from `this.workers` and re-spawned on the next burst, while every previous thread stays alive.

Scaling up to **20 cycles at size 8** (480 jobs) stays flat at 0 with the patch.

### `close()` cannot recover the leak either

Before the patch, `await pool.close()` left all **40** threads running. `close()` iterates `this.workers`, but evicted slots were already deleted from that set — nothing holds a reference to terminate. After the patch, `close()` ends at 0.

This also explains the production shape in #12812 without needing the memory numbers: 55 orphaned workers is just this accumulation over ~16h of intermittent traffic, and it is why `process.memoryUsage()` looked healthy while OS commit reached 5.7 GB.

### Note on the earlier soak comment

Correcting myself: the 556 compressions I cited were 550 `lite`, which `isCompressionWorkerEligible` rejects, so that window barely touched the pool. The harness above is the stronger evidence — it isolates the exact line this PR changes and shows before/after from one script.

hlang98923:
Production confirmation for this change — it has been running on a live 3.8.50
instance for four days.

I hit this independently from the other direction (issue #11804): unbounded OS
thread growth, ~7/hour, until the process held 101 live `Worker` instances and
5.77 GB. Notably `heapUsed` stayed at **402 MB** while RSS was **4,738 MB** —
under 9% in the main heap, because each leaked worker carries its own V8 isolate.
That is why `--max-old-space-size` never bounds this, and why raising the ceiling
only delays the symptom.

I applied exactly this change by hand to the installed bundle and measured the
same instance before and after with identical instrumentation.

| | Unpatched 3h36m | Patched 2h03m |
|---|---|---|
| Workers created (`init`) | 99 | 5 |
| `terminate()` called | 42 | **6** |
| **Still alive at end** | **57** | **0** |
| **Failure-to-terminate** | **58%** | **0%** |
| OS threads | 72, climbing | **15, flat** |
| RSS | 5.77 GB, climbing | **846 MB, flat** |

The patched trace, sampled every 10 minutes:

```
14:05  threads=17  init=3  term=2  alive=1  rss=948MB
14:15  threads=15  init=4  term=5  alive=0  rss=817MB
14:35  threads=15  init=4  term=5  alive=0  rss=828MB
15:05  threads=15  init=5  term=6  alive=0  rss=844MB
15:35  threads=15  init=5  term=6  alive=0  rss=846MB
16:05  threads=15  init=5  term=6  alive=0  rss=846MB
```

RSS *decreases* as the startup-era workers are reclaimed, then holds flat. Four
days on and it is still flat. Not slower growth — no growth.

**The diagnostic that pins it to this line specifically:** counting `WORKER`
`async_hooks` init/destroy and wrapping `Worker.prototype.terminate` showed
`terminateCalls == destroy == 26` while 23 workers stayed alive. `terminate()` was
called on precisely the workers that died and never on the leaked ones — a
teardown never *reached*, not one that fails. That is this `if (terminate)` guard
on the idle path.

Two notes for whoever merges:

1. **This is not fixed in v3.8.51.** @JasonBroderick built `c3945a72` (2026-09-04)
   and line 188 still reads `remove(slot, false)`. PR #12245 fixed the combo
   loop-safety timer, a different bug on the same issue thread.
2. **It does not require combos.** Jason reproduced on a deployment with the
   `combos` and `model_combo_mappings` tables empty. The trigger is the default
   compression pipeline under agentic load, so the leak is broader than #11804's
   title suggests.

#12542 makes the same fix with regression tests. This PR is the minimal change.
Either resolves it; merging one and closing the other would help — three open PRs
for a one-line fix is now the bottleneck, not the diagnosis.

anhtahaylove:
## Extended soak update

- samples: **24**, uptime **13 -> 508 min** (8.5 h), single PID: [16244]
- threads: **14-16** (before fix: 69-70)
- private: **641-989 MB** (before fix: 5604-5948 MB)
- private decreased in 11/23 sample-to-sample deltas -> oscillating, not climbing
- health: 24/24 samples returned HTTP 200
- compression runs in window: 1799 total, **22 worker-eligible** (`standard`/`rtk`/`stacked`); the rest are `lite`, which never spawns a worker

| uptime (min) | private (MB) | threads | handles |
|---|---|---|---|
| 13 | 641 | 15 | 364 |
| 18 | 860 | 16 | 368 |
| 20 | 902 | 15 | 366 |
| 44 | 870 | 15 | 364 |
| 45 | 904 | 15 | 366 |
| 106 | 883 | 15 | 362 |
| 168 | 873 | 15 | 383 |
| 230 | 727 | 14 | 377 |
| 292 | 817 | 14 | 377 |
| 354 | 944 | 15 | 386 |
| 416 | 890 | 14 | 377 |
| 478 | 814 | 14 | 377 |

**Scope of this evidence.** Only 22 of the 1799 compression runs in this window were worker-eligible, so the soak exercises the idle-eviction path lightly; it shows the patched build is stable in normal use (no thread growth, no one-way memory climb) rather than proving the fix under worker pressure.
The direct proof of the bug and the fix remains the harness report posted earlier (4 -> 40 MessagePorts before the fix, flat at 0 after).

anhtahaylove:
Closing in favour of #12542 — that PR is the better fix and should land instead.

Both PRs change the same idle-eviction call, but #12542 goes further in ways that matter:

- **Mine** flips one argument: `this.remove(slot, false)` → `this.remove(slot, true)`. It stops the leak at the idle path and nothing else.
- **#12542** removes the `terminate` boolean from `remove()` entirely and always terminates. That kills the whole class of bug rather than the one call site — with the flag gone, no future caller can reintroduce a non-terminating removal. It also cleans up the two callers that were passing `true` redundantly (`close()` and the timeout path).
- It ships a regression test (`tests/unit/compression/compression-worker.test.ts`) that asserts on `process.getActiveResourcesInfo()` MessagePort counts, so the leak is actually pinned rather than reasoned about.
- It has a changelog entry.

A one-line argument flip is not worth splitting reviewer attention when a strictly-superior patch with a test is already open, so I'd rather consolidate than have three near-identical PRs sitting in the queue for the same issue (#11804).

For whatever it is worth as corroboration, the failure mode in #12542's description matches what was reported on 3.8.50 in #11804: RSS growing to ~4.7 GB against a ~402 MB heap, i.e. the memory is off-heap in leaked worker isolates, not in JS objects — exactly what leaving `parentPort` listeners alive would produce.

@pacocartones' PR should get the merge. Closing this one.

JasonBroderick:
Confirming from a second production instance: this one-line change works (server-process OS threads flat at 12 to 13 for 3.5 days on v3.8.51 source, where they previously climbed one per idle eviction; built bundle shows `remove(e,!0),this.idleMs`).

Recommendation for the maintainer: merge #12542 rather than this one. Same runtime change, plus it removes the `terminate` flag so the bug cannot be reintroduced, and it carries a regression test that counts real `Worker.terminate` calls. This PR then closes as superseded. Measurements on #12812 and #12542.

PR #12818, opened 2026-09-05
fix(combo): fall through on pinned 401 responses

keeltrace:
AI-generated contribution. I am the human operator responsible for follow-up with maintainers. Please feel free to close this PR or request any changes.

## Problem

A context-cache-pinned model can return `401` for an expired or revoked credential. The pinned path currently returns that response directly, even though the combo may still have another eligible connection or provider.

## Change

Treat `401` like the existing retryable pinned statuses and fall through to the normal combo strategy.

`400` and `403` remain terminal.

## Scope

- one status added to the existing pinned fallback set
- existing pin-dispatch regression extended to cover `401`
- changelog fragment

No Auto routing, ranking, candidate-pool, or connection-exhaustion changes.

## Verification

This behavior was first exercised in broader local combo-resilience work, then reduced here to the smallest independent patch on `release/v3.8.51`.

The existing `tryPinnedModelDispatch` regression now covers `401` alongside the current retryable statuses. Fresh CI on this exact reduced head is the source of truth for this PR.


diegosouzapw:
Small, precise fix and it's fully test-covered — `401` on a pinned target now falls
through to the normal combo fallback instead of being served directly, matching how
the other transient statuses are already handled. Ran the updated
`combo-dispatch-prelude.test.ts` suite at your head: 23/23 pass. No open items from my
side — looks merge-ready.

diegosouzapw:
Thanks @keeltrace — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #12820, opened 2026-09-05
fix(plugins): stop leaking an exit listener per kill escalation

### LINKED ISSUE #12819: fix(backend): Plugin loader leaks an exit listener per hook timeout (MaxListenersExceededWarning)
## Summary

`loadPlugin()` attaches a **new** `child.once("exit", ...)` listener on every SIGKILL escalation. On a plugin child that survives `SIGTERM`, each hook timeout leaks one listener plus its `killTimer` closure, and Node starts printing `MaxListenersExceededWarning`.

Same class of bug as #12813 (register cleanup with no matching removal), but on the plugin `ChildProcess` instead of a worker thread. It only fires on the **timeout path**, so healthy plugins never hit it.

## Affected code

`src/lib/plugins/loader.ts` — two sites, both on `release/v3.8.51`:

```ts
// line 294-301, inside the hook-timeout handler
child.kill("SIGTERM");
const killTimer = setTimeout(() => {
  try { child.kill("SIGKILL"); } catch {}
}, SIGKILL_GRACE_MS);
child.once("exit", () => clearTimeout(killTimer));   // <-- new listener EVERY timeout
```

```ts
// line 399-407, inside cleanup()
child.once("exit", () => clearTimeout(killTimer));   // <-- same pattern
```

Each timeout allocates a fresh `killTimer` and a fresh listener closing over it. Nothing removes them while the child is alive.

## Reproduction

Load a plugin whose hook never resolves, so every call hits `DEFAULT_HOOK_TIMEOUT`:

```js
// plugin.mjs
export async function onRequest() { await new Promise(() => {}); }  // hangs on purpose
```

Then call the hook repeatedly and count listeners on the child:

```js
const lp = await loadPlugin(ENTRY, manifest);
const child = process._getActiveHandles().find(h => h?.constructor?.name === "ChildProcess");
for (let i = 0; i < 6; i++) {
  try { await lp.plugin.onRequest?.({ n: i }); } catch {}
  console.log(child.listenerCount("exit"));
}
```

### Measured — `listenerCount("exit")` on the same child

| after hook | 1 | 2 | 3 | 4 | 5 | 6 |
|---|---|---|---|---|---|---|
| current | 2 | 1 | 2 | 3 | **4** | **5** |

Monotonic growth, never released. Node emits this from the second load onward:

```
11 exit listeners added to [ChildProcess]. MaxListeners is 10.
```

## Scope — what is NOT affected

I checked these explicitly so the report is not overstated:

- **No orphaned child processes.** After a run with a deliberately hanging plugin, zero surviving children.
- **No leaked temp host scripts.** `omniroute-plugin-host-*.mjs` count stayed 0 — `removeHostScript()` works.
- The SIGTERM -> SIGKILL escalation itself works correctly.

The leak is limited to listeners and their retained closures, on the timeout path only.

## Suggested fix

Use one shared `killTimer` and attach the exit listener exactly once at spawn time. PR follows.

## Environment

- omniroute `3.8.50` (npm global), branch `release/v3.8.51` verified to contain the same code
- Node.js v24.19.0, Windows 11

anhtahaylove:
Fixes #12819.

## Problem

Every SIGKILL escalation attached a **new** `child.once("exit", ...)` listener closing over its own `killTimer`. On a plugin child that survives `SIGTERM`, each hook timeout leaked one listener plus one retained closure — nothing removed them while the child stayed alive.

Two sites had the identical pattern: the hook-timeout handler and `cleanup()`.

Node made it visible on its own:

```
11 exit listeners added to [ChildProcess]. MaxListeners is 10.
```

## Fix

One shared `killTimer` and one exit listener attached at spawn time:

- `killEscalate()` sends `SIGTERM`, then arms the SIGKILL timer **only if one is not already pending** — repeated escalations reuse it instead of stacking.
- A single `child.once("exit", clearKillTimer)` clears whatever is pending.
- `killTimer.unref?.()` so a pending escalation never keeps the event loop alive.

Both call sites now just call `killEscalate()`.

## Verification

Loaded a plugin whose hook never resolves, so every call hits the timeout, and counted `listenerCount("exit")` on the same child:

| after hook | 1 | 2 | 3 | 4 | 5 | 6 |
|---|---|---|---|---|---|---|
| before | 2 | 1 | 2 | 3 | **4** | **5** |
| after | 2 | 1 | 1 | 1 | 1 | 1 |

Growth is gone; the count stays flat. `MaxListenersExceededWarning` no longer appears.

Behaviour that is unchanged, checked explicitly:

- SIGTERM -> SIGKILL escalation still fires (verified with a plugin that ignores SIGTERM)
- zero orphaned child processes after runs with hanging plugins
- zero leaked `omniroute-plugin-host-*.mjs` temp scripts

## Notes

- Independent of #12813 / #12815 — different file, no overlap, can merge in any order.
- Ran against `release/v3.8.51`; the same code is in the published `3.8.50` package.
- I verified the patched loader in an isolated harness, not inside a long-running production instance, so I cannot report multi-hour soak numbers here.

### DIFF
--- a/src/lib/plugins/loader.ts
+++ b/src/lib/plugins/loader.ts
@@ -271,6 +271,30 @@ export async function loadPlugin(
     removeHostScript(hostScriptPath);
   });
 
+  // Escalate SIGTERM -> SIGKILL using a SINGLE shared timer and a SINGLE exit
+  // listener. Attaching `child.once("exit", ...)` per escalation leaks one
+  // listener per hook timeout on a child that survives SIGTERM, tripping
+  // MaxListenersExceededWarning and retaining closures for the process lifetime.
+  let killTimer: NodeJS.Timeout | null = null;
+  const clearKillTimer = (): void => {
+    if (killTimer) {
+      clearTimeout(killTimer);
+      killTimer = null;
+    }
+  };
+  child.once("exit", clearKillTimer);
+  const killEscalate = (): void => {
+    child.kill("SIGTERM");
+    if (killTimer) return; // escalation already pending
+    killTimer = setTimeout(() => {
+      killTimer = null;
+      try {
+        child.kill("SIGKILL");
+      } catch {}
+    }, SIGKILL_GRACE_MS);
+    killTimer.unref?.();
+  };
+
   // Call a hook in the child process with a timeout. Blocking/lifecycle hooks escalate
   // SIGTERM → SIGKILL on timeout; NOTIFICATION_HOOKS only drop the pending call.
   const callHook = (hook: string, payload: unknown, timeout = hookTimeoutMs): Promise<unknown> => {
@@ -291,14 +315,7 @@ export async function loadPlugin(
           resolve(undefined);
           return;
         }
-        child.kill("SIGTERM");
-        // Escalate to SIGKILL if plugin ignores SIGTERM
-        const killTimer = setTimeout(() => {
-          try {
-            child.kill("SIGKILL");
-          } catch {}
-        }, SIGKILL_GRACE_MS);
-        child.once("exit", () => clearTimeout(killTimer));
+        killEscalate();
         reject(new Error(`Plugin hook '${hook}' timed out after ${timeout}ms`));
       }, timeout);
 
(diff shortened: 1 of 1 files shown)


anhtahaylove:
Superseded by #13092, which carries the identical source change plus a regression test and a changelog entry. Closing this one to keep the review queue unambiguous — no need to review both.

PR #12823, opened 2026-09-05
fix(llmlingua): pass a URL object to new Worker() instead of a file:// string

### LINKED ISSUE #12822: fix(backend): LLMLingua worker never spawns: new Worker() gets a file:// string, throws ERR_WORKER_PATH, silently fail-opens
## Summary

`llmlingua/worker.ts:237` passes a **string** `file://` URL to `new Worker()`. Node rejects that with `ERR_WORKER_PATH`, so the LLMLingua ONNX worker **never spawns**. The failure is swallowed by a bare `catch {}` in `pump()`, so every call silently fail-opens and returns the text uncompressed.

```ts
// open-sse/services/compression/engines/llmlingua/worker.ts:237
const w = new Worker(pathToFileURL(absoluteWorkerFile).href, { execArgv });
//                                                    ^^^^^ string, not URL
```

Per the [worker_threads docs](https://nodejs.org/api/worker_threads.html), `filename` must be an absolute path, a relative path starting with `./`/`../`, **or a WHATWG `URL` object** using `file:`. A serialized `file://` *string* is explicitly not accepted.

## Reproduction

Against the real `onnxWorker.js` from an installed 3.8.50, on Node v24.19.0:

```js
const f = path.resolve(".../dist/open-sse/services/compression/engines/llmlingua/onnxWorker.js");

new Worker(pathToFileURL(f).href, { execArgv: [] });  // ERR_WORKER_PATH
new Worker(pathToFileURL(f),      { execArgv: [] });  // OK
```

```
dong 237 nhu hien tai (.href chuoi): LOI -> ERR_WORKER_PATH
neu bo .href (URL object)         : OK
```

I also reached it through the public entry point by driving `workerBackend()` with shimmed deps. An 80-char input came back **80 chars, unchanged**, and instrumenting the swallowed `catch` printed:

```
!!! ensureWorker NEM LOI: The worker script or module filename must be an absolute path
or a relative path starting with './' or '../'. Wrap file:// URLs with `new URL`.
Received "file:///.../onnxWorker.js"
```

## Why it stayed hidden

`pump()` catches the spawn failure and fail-opens:

```ts
try {
  w = ensureWorker();
} catch {
  busy = false;
  item.resolve(item.text);   // returns input unchanged
  pump();
  return;
}
```

Callers get valid-looking output, just never compressed. Nothing is logged, so the engine silently degrades to a no-op rather than erroring.

## Scope

- Affects every install where the gate passes (`@atjsh/llmlingua-2` present) — it is present in a default 3.8.50 install, and `dist/.../onnxWorker.js` ships too, so the path is reachable, not dead code.
- `engines` allows `>=24.0.0 <27`; the check is a hard API contract, so the call fails on any supported Node.
- Still present on `release/v3.8.51` at the same line.
- Contrast `compressionWorkerPool.ts`, which passes the `URL` object from `pathToFileURL()` and works correctly — the two worker call sites disagree.

## Fix

Drop `.href` and pass the URL object. PR follows.

## Environment

- OmniRoute 3.8.50, Node v24.19.0, Windows 11

anhtahaylove:
Fixes #12822.

## Problem

`new Worker()` does not accept a serialized `file://` **string** — it wants an absolute path, a `./`-relative path, or a WHATWG `URL` object. `worker.ts:237` passed `pathToFileURL(...).href`, so every spawn threw `ERR_WORKER_PATH`:

```
TypeError [ERR_WORKER_PATH]: The worker script or module filename must be an
absolute path or a relative path starting with './' or '../'.
Wrap file:// URLs with `new URL`. Received "file:///.../onnxWorker.js"
```

`pump()` catches spawn failures and fail-opens, so the LLMLingua engine silently returned **uncompressed** text instead of surfacing an error. The engine has effectively been a no-op wherever the dependency gate passes.

## Fix

```diff
-const w = new Worker(pathToFileURL(absoluteWorkerFile).href, { execArgv });
+const w = new Worker(pathToFileURL(absoluteWorkerFile), { execArgv });
```

This matches `compressionWorkerPool.ts`, which already passes the `URL` object returned by `pathToFileURL()` and spawns correctly. The two call sites now agree.

## Verification

Against the real `onnxWorker.js` from an installed 3.8.50, Node v24.19.0:

| call | result |
|---|---|
| `new Worker(pathToFileURL(f).href, …)` (current) | `ERR_WORKER_PATH` |
| `new Worker(pathToFileURL(f), …)` (this PR) | spawns, terminates cleanly |

Also exercised end-to-end through `workerBackend()` with the optional deps shimmed: before the change an 80-char input came back 80 chars unchanged (fail-open); after it, the worker starts and the round-trip reaches the worker.

## Note on error visibility

This PR is the minimal correctness fix. Separately, the bare `catch {}` around `ensureWorker()` is what let a 100% spawn failure go unnoticed — worth at least a debug-level log so a broken worker path cannot masquerade as "compression ran, saved nothing". Happy to add that here or in a follow-up, whichever you prefer.

## Relation to my other PRs

Independent of #12813 / #12815 / #12820 — different file, no overlap, any merge order works.

### DIFF
--- a/open-sse/services/compression/engines/llmlingua/worker.ts
+++ b/open-sse/services/compression/engines/llmlingua/worker.ts
@@ -234,7 +234,7 @@ function ensureWorker(): Worker {
 
   const { workerFile, execArgv } = resolveWorkerFile();
   const absoluteWorkerFile = path.resolve(workerFile);
-  const w = new Worker(pathToFileURL(absoluteWorkerFile).href, { execArgv });
+  const w = new Worker(pathToFileURL(absoluteWorkerFile), { execArgv });
 
   w.on("message", (reply: WorkerReply) => {
     const entry = pending.get(reply.id);


anhtahaylove:
Superseded by #13093, which carries the identical source change plus a regression test and a changelog entry. Closing this one to keep the review queue unambiguous — no need to review both.

PR #12663, opened 2026-09-04
feat(providers): register gemini-3.8-flash

toor11:
Closes #12638

DeepMind published Gemini 3.8 Flash (2026-09-02, https://deepmind.google/models/model-cards/gemini-3-8-flash/) as next iteration after 3.7. Registry topped out at 3.7.

- Root cause: `open-sse/config/providers/registry/gemini/index.ts:12` missing 3.8 entry (verified via `git show origin/release/v3.8.51:...`)
- Change: add model `{id: "gemini-3.8-flash", name: "Gemini 3.8 Flash", toolCalling:true, supportsVision:true}` at top of `models[]` (newest-first, same style as 3.7, 6 lines, no refactor)
- Verification: `npx tsc --noEmit --skipLibCheck` exit 0; manual parse shows 9 models with 3.8 first; `grep -R socket .github/workflows` 0 hits unaffected; fragment `changelog.d/features/12638-gemini-3.8-flash.md` added per CONTRIBUTING
- No CHANGELOG.md edit, no Co-Authored-By, 2 commits (feat + chore fragment)

PR #6943, opened 2026-07-12
fix(6813): fix thinking budget zero drop and default thinkingConfig injection

rafaumeu:
## Summary
Fix truthy check for `budget_tokens` to allow `0` (was dropping zero budgets), and stop injecting default `thinkingConfig` when no knobs present.

## Changes
- **SPEC**: `.planning/6813-openai-to-gemini-thinking-budget-zero.md`
- **PLAN**: `.planning/6813-openai-to-gemini-thinking-budget-zero-PLAN.md`
- **Code**: `open-sse/translator/request/openai-to-gemini.ts` — fix truthy check `budget_tokens > 0` (was `budget_tokens`) and guard default injection
- **Tests**: `open-sse/translator/request/__tests__/openai-to-gemini.test.ts` — +85 lines covering all scenarios

## Verification
- `npm run lint` — passes
- `npm run typecheck:core` — passes

## Related
Fixes #6813

PR #12826, opened 2026-09-05
fix(gemini): custom-model aliases 404 because URL uses the alias, not the rewritten body model

adivekar-utexas:
## Summary

- Gemini custom-model aliases such as `gemini-3.7-flash-high` (rewritten by an operator payload rule to the real id `gemini-3.7-flash`) returned Google 404 `models/<alias> is not found for API version v1beta`. Payload rules already rewrote `body.model`; the executor still interpolated the original alias into `/models/{model}:generateContent`.
- `BaseExecutor.buildUrlForBody()` now builds the URL from the payload-rule-prepared body model (falls back to the executor model when the body has no string model, so every other provider is unchanged). chatCore logs the rewrite and keys Gemini RPM/RPD accounting to the wire model that actually shipped.
- Gemini 3.x deprecated numeric `thinkingBudget` in favour of string `thinkingLevel` (`low|medium|high`). An explicit level from the request or a payload override now replaces the injected budget instead of being sent alongside it. A rule that writes the budget directly, or a null "disable thinking" level, is left untouched.

## Related Issues

- Related to custom-model + payload-override usage on the native `gemini` (Google AI Studio) provider. Reproduced from a live dashboard call log: URL named `gemini-3.7-flash-high` while the body named `gemini-3.7-flash`.

## Validation

- [x] Focused tests: `tests/unit/gemini-payload-override-model-url.test.ts`, `tests/unit/gemini-thinking-level-forwarding.test.ts`, `tests/unit/payload-rules-thinking-level-drops-budget.test.ts`, plus existing Gemini translator suites
- [x] ESLint on the touched files (`--no-cache`)
- Reconciled with current `upstream/main` (no overlapping open Gemini URL PRs; #12663 only registers `gemini-3.8-flash` in the catalog)

## Tests Added Or Updated

- `tests/unit/gemini-payload-override-model-url.test.ts` — URL lockstep (`buildUrlForBody` uses rewritten `body.model`; missing-model fallback; `execute()` fetch URL does not contain the alias)
- `tests/unit/gemini-thinking-level-forwarding.test.ts` — translator forwards `thinkingLevel` and omits `thinkingBudget`; no-level baseline still injects a budget
- `tests/unit/payload-rules-thinking-level-drops-budget.test.ts` — override/default rules that write a string level drop the leftover budget; unrelated rules, explicit budget writes, and null levels do not

## Coverage Notes

- `open-sse/executors/base.ts` (`buildUrlForBody` + execute loop): URL tests above, including a mocked-fetch `execute()` path.
- `open-sse/translator/request/openai-to-gemini.ts`: thinking-level forwarding tests + existing `#6943` / defaults suites (still pass).
- `open-sse/services/payloadRules.ts`: thinking-level drop tests, including the self-erase and null-level guards.
- `open-sse/handlers/chatCore.ts` / `upstreamTimeouts.ts`: thin plumbing (`normalizeExecutorResult.model`, Gemini rate-limit key). No separate chatCore fixture; the wire model is produced by the executor test.

## Reviewer Notes

- This is a lockstep fix, not a Gemini-only URL parser. Any provider that puts the model in the URL path (today: Gemini `generateContent`) inherits the same body-model source. OpenAI-style `/chat/completions` URLs ignore the model argument already, so they are unaffected.
- The `thinkingBudget` drop is gated on writing a **non-empty string** `thinkingLevel` / `thinking_level`. Operators who still want a numeric budget on 2.5-class models can keep sending it.
- Changelog fragment: `changelog.d/fixes/gemini-custom-model-url-404.md` (PR number placeholder will be filled after this PR is numbered).

Made with [Cursor](https://cursor.com)


diegosouzapw:
Ran your 12 new/changed tests at the PR head — all pass, including the URL-lockstep and
payload-rule budget-drop cases. The "Build" CI failure you're seeing is an infra cancellation
(runner shutdown signal), not a real compile error — I confirmed that reading the run log; it
had already compiled successfully before being canceled. Two things needed before merge: (1)
could you retarget this PR to `release/v3.8.51` (it's currently based on `main`, which is
behind the active release branch)? (2) #13463 (also open) touches the exact same three
`thinkingConfig` blocks in `openai-to-gemini.ts` with a complementary mechanism — whichever of
the two merges second will need a small rebase to combine both conditions. Happy to help
coordinate once the owner picks an order.

PR #12829, opened 2026-09-05
fix(tests): Adobe Firefly edits-route test launched a real headless Chrome at firefly.adobe.com

insoln:
## Summary

- `tests/unit/8510-adobe-firefly-edits-route.test.ts` exercises the real `POST /v1/images/edits` route handler. The route's session-resolve step cannot thread `ensureAdobeFireflySession`'s `allowBrowserRefresh: false` seam, and browser Forter-warm is the **default** engine (`adobeFireflySession.ts` — "Browser Forter-warm is the DEFAULT engine"), so every run launched a **real headless Chrome at https://firefly.adobe.com** over CDP.
- Observed in CI run logs: two live CDP launches per suite run plus ~3 min of network retries ("Chrome CDP launch attempt 1 failed: forterToken stale" — expected, since a real token requires a signed-in session).
- Fix: set the documented kill switch `ADOBE_FIREFLY_BROWSER_REFRESH=0` for the test process (mirroring the existing `DATA_DIR` env pattern) and restore it in `test.after`. No production code touched.

## Validation

- `node --import tsx/esm --test tests/unit/8510-adobe-firefly-edits-route.test.ts` — **4/4 pass in 12 s, zero CDP launches** (previously ~3 min with live network waits). Verified `grep -c "CDP warm attached"` = 0 on the fixed run.
- Prettier + ESLint (full lint-staged pre-commit incl. docs-sync, any-budget, tracked-artifacts) — pass.
- No production files changed; single test file, +8 lines.

## Related Issues

- Related to #12732 (inherited base-red — not addressed here)

⚠️ base-red inherited: #12732

### DIFF
--- a/tests/unit/8510-adobe-firefly-edits-route.test.ts
+++ b/tests/unit/8510-adobe-firefly-edits-route.test.ts
@@ -13,6 +13,12 @@ import path from "node:path";
 const TEST_DATA_DIR = fs.mkdtempSync(path.join(os.tmpdir(), "omniroute-adobe-firefly-edits-"));
 process.env.DATA_DIR = TEST_DATA_DIR;
 process.env.API_KEY_SECRET = process.env.API_KEY_SECRET || "adobe-firefly-edits-test-secret";
+// The route path cannot thread ensureAdobeFireflySession's allowBrowserRefresh:false seam,
+// and browser Forter-warm is the DEFAULT engine — without this kill switch the route's
+// session-resolve step launches a real headless Chrome at firefly.adobe.com (two live CDP
+// launches + ~3 min of network waits per suite run; both observed in CI logs).
+const PREV_BROWSER_REFRESH = process.env.ADOBE_FIREFLY_BROWSER_REFRESH;
+process.env.ADOBE_FIREFLY_BROWSER_REFRESH = "0";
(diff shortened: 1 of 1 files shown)


diegosouzapw:
Good catch and a clean fix. #13289 (test-runner auto-detection baked into
`adobeFireflyBrowserEnabled()`) and #13358 (the same env var forced off globally in
`tests/_setup/isolateDataDir.ts`, which every test process already imports) both land at a
broader layer and would already cover this exact file without needing a per-test opt-out.
Recommending we close this in favor of those two so we don't carry three overlapping guards
for the same symptom — but flagging to the owner: if either of those doesn't land, this PR
should be merged instead as the minimal safety net. Verified your fix works standalone too
(4/4 pass, no CDP launch in the logs).

insoln:
Closing this as superseded, following the maintainer's recommendation. Both broader guards have now landed on `release/v3.8.51`:

- #13289 prevents Adobe Firefly browser spawning under test runners at the service boundary.
- #13358 applies the test environment guard globally and adds dedicated regression coverage.

Those fixes cover the same leak more cleanly than retaining a third per-test environment override here. Thanks for the review and direction.

Issue #11632, opened 2026-08-26
[BUG] MODELS_CATALOG_PREFIX_MODE is not working with OpenAI Codex


wsycarlos:
### OmniRoute Version

3.8.51

### Installation Method

Docker / Docker Compose

### Operating System

Windows

### OS Version

Windows11 25H2

### Provider(s) Involved

OpenAI Codex

### Model(s) Involved

all

### Client Tool

cURL

### Description

Currently, no matter how you set MODELS_CATALOG_PREFIX_MODE, in .env or configure it in Settings, the /v1/models endpoint still outputs both models ("cx/gpt-*" and "codex/gpt-*"), and even worse, I even get a model with no alias, such as:
```
{
      "id": "cx/codex-auto-review",
      "object": "model",
      "created": 1787744755,
      "owned_by": "codex",
      "permission": [],
      "root": "codex-auto-review",
      "parent": null,
      "capabilities": {
        "tool_calling": true,
        "reasoning": true
      },
      "name": "cx/codex-auto-review",
      "context_length": 400000
    },
    {
      "id": "codex/codex-auto-review",
      "object": "model",
      "created": 1787744755,
      "owned_by": "codex",
      "permission": [],
      "root": "codex-auto-review",
      "parent": "cx/codex-auto-review",
      "capabilities": {
        "tool_calling": true,
        "reasoning": true
      },
      "name": "codex/codex-auto-review",
      "context_length": 400000
    },
    {
      "id": "codex-auto-review",
      "object": "model",
      "created": 1787744755,
      "owned_by": "codex",
      "permission": [],
      "root": "codex-auto-review",
      "parent": "codex/codex-auto-review",
      "capabilities": {
        "tool_calling": true,
        "reasoning": true
      },
      "name": "codex-auto-review",
      "context_length": 400000
    },
```

Also, another thing might worth mentioning, after change the value in Settings, I once got a reply for /v1/models endpoint with some of the providers missing(opencode-go in my case), the reply only contains codex related models.

### Steps to Reproduce

Just configure OpenAI Codex (in my case, I have two accounts).
Tweak MODELS_CATALOG_PREFIX_MODE to whatever value.

### Expected Behavior

When MODELS_CATALOG_PREFIX_MODE is set to alias, you should only see cx/, and when it is set canonical, you should only see codex/. And neither should display models with no prefix.

### Actual Behavior

You can always see models cx/ and codex/ alias and also models with no prefix.

### Test Impact

Unsure

[...]

PR #12830, opened 2026-09-05
fix(db): reclaim freed pages incrementally instead of a blocking VACUUM in the cleanup scheduler (#12821)

### LINKED ISSUE #12821: fix(backend): cleanup.ts runs a blocking full VACUUM whenever ≥1 row is deleted (startup + every 6h), bypassing vacuumScheduler — 8 min unresponsive server
## Summary

`startCleanupScheduler()` runs a **synchronous, full-database `VACUUM` on the event loop whenever the cleanup pass deleted at least one row** — 30 s after every start, and again every 6 hours on a serving process. Because `node:sqlite` is synchronous, the whole server stops responding for the duration: no HTTP route answers, `/healthz` included, and in-flight streams stall.

On a 540 MB `storage.sqlite` this froze the server for **7 min 55 s** to reclaim the space of **6 deleted rows**.

It also **bypasses the app's own `vacuumScheduler`**, which exists precisely to decide when full VACUUMs happen, and which on the affected install was explicitly configured to run *weekly at 02:00*.

## Impact — measured on a live install

```
19:17:25.840Z  [Cleanup] Auto-cleanup complete: 6 deleted, 0 errors
19:17:25.841Z  [Cleanup] Startup cleanup freed 6 rows. Running VACUUM...
19:25:21.160Z  [Cleanup] VACUUM completed after startup cleanup.
```

**475 seconds of a fully blocked event loop.** The rows were 6 stale `domain_cost_history` entries.

The process is alive and busy the whole time — `/proc/<pid>/io` showed 865 MB read / 352 MB written and the WAL grew to 450 MB against the 540 MB database — it simply cannot answer anything while `db.exec("VACUUM")` holds the loop.

What this looks like depending on how OmniRoute is run:

- **Bare `npm start` / CLI:** the server hangs for minutes shortly after boot. Clients see connection timeouts; there is no log line explaining it beyond the one `Running VACUUM...` message.
- **Behind any reverse proxy:** the proxy's upstream timeout fires and every client gets a 5xx for the whole window.
- **Under any supervisor with an HTTP health check** (Docker `HEALTHCHECK`, systemd watchdog, PM2, a container orchestrator): the check times out and the instance is declared unhealthy. If that check is wired to *restart* the process, this becomes a restart loop — the restart re-runs the cleanup, which deletes rows again, which VACUUMs again. Only a TCP-level check survives, because the listening socket still accepts connections at the OS level while the loop is blocked. (This matches the guidance already merged in #10297 — TCP liveness, HTTP `/healthz` readiness — and it is the only reason the affected install stayed up rather than crash-looping.)

Since the startup cleanup almost always deletes *something*, **every restart carries a multi-minute outage**, regardless of what changed.

> Note on environment: the numbers above come from a containerised deployment orchestrated with Kubernetes, which is not officially supported yet — see the open PR #12400. Nothing in this report depends on that: the blocking VACUUM is plain Node + `node:sqlite` behaviour and reproduces under any process manager, or none at all.

## Root cause

`src/lib/db/cleanup.ts` — the trigger is `totalDeleted > 0`, i.e. a single row is enough:

```ts
// startup path (~line 783), fired 30s after boot
const totalDeleted = result.totalDeleted + proxyResult.deleted;
if (totalDeleted > 0) {
  console.log(`[Cleanup] Startup cleanup freed ${totalDeleted} rows. Running VACUUM...`);
  const db = getDbInstance();
  db.exec("VACUUM");            // ← synchronous, whole database, on the event loop
}
```

The identical block is repeated in the periodic branch (~line 805), fired by `setInterval(…, CLEANUP_INTERVAL_MS /* 6h */)`. That one is arguably worse: it freezes a *serving* process mid-traffic, four times a day, at an hour nobody chose.

## This contradicts the design that already exists

`src/lib/db/vacuumScheduler.ts` is the app-level scheduler for full VACUUM. Its own docblock states it is *the* place full VACUUM runs are scheduled: it follows the Storage page's `optimization.scheduledVacuum` / `optimization.vacuumHour`, persists state to `key_value`, and exposes `getState()` / `runNow()` / `stop()` to the API and UI.

`cleanup.ts` calls `db.exec("VACUUM")` directly and honours none of it. On the affected install the operator had configured:

```
databaseSettings / optimization.scheduledVacuum = "weekly"
databaseSettings / optimization.vacuumHour      = 2
scheduler / vacuum = {"enabled":true,"intervalMs":604800000,"lastDurationMs":35528,…}
```

Intent: *once a week, at 02:00* — and the scheduler's own run takes **35 s**. What actually happened: a **475 s** VACUUM at 19:17, during peak traffic. An operator who disables scheduled vacuum outright still gets these.

Worth noting `optimization.autoVacuumMode = "INCREMENTAL"` is the configured default: SQLite is already maintaining a freelist for exactly this purpose, so a full rebuild is the wrong tool regardless of when it runs.

## Reproduction

1. Run against a `storage.sqlite` of a few hundred MB with retention settings that delete at least one row.
2. Restart; wait ~30 s for the startup cleanup.
3. Request any route — the server does not answer until the VACUUM finishes (minutes, scaling with database size).

## Suggested fix

The two changes below are complementary and together remove both the blocking and the policy violation, without adding any new configuration surface:

1. **Reclaim incrementally instead of rebuilding.** With `auto_vacuum = INCREMENTAL` already the default, `PRAGMA incremental_vacuum(N)` reclaims at most `N` freed pages per call. Cleanup can call it in a bounded loop, yielding between batches, so reclamation costs milliseconds per tick instead of freezing the loop for minutes — and it never needs the ~2× free disk space a full rebuild does.
2. **Never run a full `VACUUM` from `cleanup.ts`.** When a full rebuild genuinely is warranted (e.g. `auto_vacuum = NONE`, where `incremental_vacuum` is a no-op), record that one is due and let `vacuumScheduler` perform it in its configured window. That restores `scheduledVacuum` / `vacuumHour` as the single source of truth and is the behaviour the scheduler's docblock already claims.

Lesser variants, if the above is too large a change:

3. Gate the existing call on something meaningful instead of `totalDeleted > 0` — e.g. `PRAGMA freelist_count` relative to `page_count`, or a row/byte threshold. Six rows out of a 540 MB file should never trigger a full rebuild.
4. Move the full VACUUM off the serving thread — a worker thread or a short-lived child process against the same file keeps the server answering. (This is what the affected operator now does by hand for WAL checkpoints, for the same reason.)
5. At minimum, an env gate such as `OMNIROUTE_CLEANUP_VACUUM=0`. Today there is no way to opt out.

## Environment

- OmniRoute v3.8.51, Node.js 26.8.1, Linux
- `storage.sqlite` 540 MB, WAL journaling, `auto_vacuum = INCREMENTAL`
- Reproduces independently of process manager; the numbers above were captured in a containerised deployment (cf. #12400)

insoln:
Fixes #12821.

## Problem

`startCleanupScheduler()` in `src/lib/db/cleanup.ts` ran a synchronous, whole-database `VACUUM` on the event loop whenever the cleanup pass deleted at least one row — 30 s after every start and again every 6 h. With `node:sqlite` that blocks every route (`/healthz` included) and stalls in-flight streams for the duration. Measured on a live install: **7 min 55 s** to reclaim the space of **6 deleted rows** on a 540 MB `storage.sqlite`.

It also bypassed `vacuumScheduler`, the app-level owner of full VACUUMs, whose `scheduledVacuum` / `vacuumHour` settings the operator had set to *weekly at 02:00*.

## Fix

`cleanup.ts` never issues a full `VACUUM` any more. After each cleanup pass it calls the new `reclaimFreedPages()`, which branches on `PRAGMA auto_vacuum`:

| `auto_vacuum` | behaviour |
| --- | --- |
| `INCREMENTAL` | drains the freelist with `PRAGMA incremental_vacuum(N)` in ~1 MiB batches (`N` derived from `page_size`), **pausing between batches for as long as the last batch took** (≤250 ms) so the loop stays ≥50 % free; PASSIVE checkpoint every 64 batches and a TRUNCATE checkpoint at the end so the main file actually shrinks in WAL mode; hard caps of 2048 batches / 30 s per pass, remainder picked up next pass. |
| `FULL` | nothing to do — SQLite reclaims on commit. |
| `NONE` | `incremental_vacuum` is a no-op, so the pass calls the new `vacuumScheduler.requestFullVacuum(reason)`; the rebuild happens in the operator-configured window (or via the Storage page button). `scheduledVacuum = never` is honored. |

`vacuumScheduler` gains two persisted state fields, `fullVacuumRequestedAt` / `fullVacuumRequestReason`, cleared by the next successful `runNow()`. The request survives restarts, and an early call before `init()` hydrates from `key_value` first so it cannot clobber the persisted `lastRunAt`.

Robustness in the incremental loop: `db.exec()` instead of `pragma()` (bun:sqlite's `all()` steps a zero-column pragma only once, which would free one page per batch), `SQLITE_BUSY`/`LOCKED` stops the pass instead of throwing, a handle closed under the pass (backup restore / import) stops it quietly, any other error stops it with the partial progress reported in the log.

Also removed the duplicate `cleanupProxyLogs()` call in the scheduled pass — `runAutoCleanup()` already covers `proxy_logs`.

## Behaviour change worth knowing

A database file that was never rebuilt through the Storage page's optimization save is on-disk `auto_vacuum=NONE` — the SQLite default — **even when the Storage setting says `INCREMENTAL`**, because startup applies optimization settings with `applyPersistent: false`. The reporting install turned out to be exactly that (`PRAGMA auto_vacuum` → `0` while `databaseSettings.optimization.autoVacuumMode = "INCREMENTAL"`); the issue text calling it INCREMENTAL was wrong and is corrected in a comment there.

For such installs this PR turns "blocking VACUUM every 6 h" into "no reclamation until the scheduled VACUUM window (default weekly)". That is the intended trade; the pending request is visible via `GET /api/settings/database/vacuum`. Two natural follow-ups, deliberately out of scope here: surface the pending request on the Storage page, and reconcile the on-disk mode with the stored setting (or default new databases to INCREMENTAL before `SCHEMA_SQL`).

## Tests

- `tests/unit/db/cleanup-reclaim-freed-pages.test.ts` (new): INCREMENTAL drains + pauses + `page_count` shrinks after the checkpoint; default batch derived from `page_size`; `maxBatches` and time-budget caps; FULL is a no-op; NONE defers to the scheduler and leaves `page_count` untouched; empty freelist is `skipped`; `runScheduledCleanupPass()` (the exact code the timers run) defers on NONE and logs the summary line.
- `tests/unit/db/vacuum-scheduler.test.ts`: `requestFullVacuum()` records/persists (first timestamp wins, latest reason wins), survives a simulated restart, is cleared by `runNow()`, honors `never`.
- `tests/unit/cleanup-column-fix.test.mjs`: source invariant flipped from "must contain VACUUM" to "must use `incremental_vacuum` and must not run a full VACUUM".

All of these fail against the previous `cleanup.ts` / `vacuumScheduler.ts`. Also green locally: `check:cycles`, `typecheck:core`, `eslint` on the changed files, `check:any-budget:t11`, `check:tracked-artifacts`.

## Validation on a live install

Deployed to the reporting installation (525 MB DB, on-disk `auto_vacuum=NONE`, Kubernetes, 1 replica, Longhorn PVC) before opening this PR. The startup pass log lines and the scheduler state after the pass are in the first comment below.

⚠️ base-red inherited: #12732


insoln:
## Live validation on the reporting install

Deployed this branch (cherry-picked onto the install's production build, image `ghcr.io/insoln/omniroute@sha256:63dc0859…`) to the installation from #12821: Kubernetes, 1 replica, Longhorn PVC, `storage.sqlite` 525 MB (128 405 pages × 4 KiB), **on-disk `auto_vacuum=0` (NONE)**, `scheduledVacuum=weekly`, `vacuumHour=2`.

State before the rollout:

```
auto_vacuum=0 page_size=4096 page_count=128405 freelist_count=0
scheduler/vacuum = {"enabled":true,"intervalMs":604800000,"lastRunAt":1788573600001,"lastDurationMs":35528,...}
```

Startup cleanup pass (+30 s after boot), verbatim from the pod log:

```
[Cleanup] Background cleanup scheduler started (every 6 hours).
[Cleanup] Starting auto-cleanup...
[Cleanup] Auto-cleanup complete: 0 deleted, 0 errors
[VacuumScheduler] Full VACUUM requested (cleanup left 658 free page(s) that auto_vacuum=0 cannot reclaim incrementally); deferred to the scheduled run at 2026-09-12T02:00:00.000Z.
[Cleanup] Space reclamation after startup cleanup: auto_vacuum=NONE — 658 free page(s) left for the scheduled VACUUM.
```

No `Running VACUUM...` line, no freeze. `/healthz` sampled from inside the pod right after the pass: 0–1 ms. Traffic in the first 10 minutes after the rollout: 74 OK / 2 client-side `499 Request aborted`, no 5xx.

State after the pass — the request is persisted alongside the existing scheduler state and the file was not rebuilt:

```
auto_vacuum=0 page_count=131452 freelist_count=655
scheduler/vacuum = {..., "lastRunAt":1788573600001, "nextRunAt":1789178400000,
  "fullVacuumRequestedAt":1788649169631,
  "fullVacuumRequestReason":"cleanup left 658 free page(s) that auto_vacuum=0 cannot reclaim incrementally"}
```

(`lastRunAt` from the previous scheduled run survived the request — the hydrate-before-write path.) The weekly run on 2026-09-12 will clear it.

Same day, previous image, same install, for comparison — the behaviour this PR removes:

```
19:17:25.840Z  [Cleanup] Auto-cleanup complete: 6 deleted, 0 errors
19:17:25.841Z  [Cleanup] Startup cleanup freed 6 rows. Running VACUUM...
19:25:21.160Z  [Cleanup] VACUUM completed after startup cleanup.
```

The INCREMENTAL path is exercised by the unit tests (`tests/unit/db/cleanup-reclaim-freed-pages.test.ts`), not by this install, since its file is NONE on disk — see the correction comment on #12821.

insoln:
CI status note — the red checks on this PR are inherited from the base (`release/v3.8.51`, ⚠️ base-red #12732), not introduced here. The branch touches six files (`git diff --stat origin/release/v3.8.51...HEAD`): `src/lib/db/cleanup.ts`, `src/lib/db/vacuumScheduler.ts`, three tests, one changelog fragment.

| check | failing item | why it is not this PR |
| --- | --- | --- |
| API Route Typecheck | `open-sse/executors/glm.ts TS2554` | file untouched by this branch |
| Docs Gates (fast-path) | `README.md` / `AGENTS.md` / `llm.txt` say "169 migrations", code has 170 | `src/lib/db/migrations/` on base already has 170; no migration added here |
| Merge integrity | `changelog.d/fixes/reset-aware-model-family.md` does not start with `- ` | fragment exists on base with that content; this PR's fragment is `12830-cleanup-no-blocking-vacuum.md` and is valid |
| Unit Tests fast-path (1/4 … 4/4) | 1/4: HuggingFace image 400 hint, agent-skills sync gate, Kiro `response.failed`, `sanitizeErrorMessage` tunnel leak, 200-stream error frame · 2/4: #11632 canonical-prefix rows, HuggingChat boundaries, credential redaction, model-sync JSON, `persistAttemptLogs` · 3/4: HuggingChat transport failures, codex `handleImageGeneration` · 4/4: generic stream boundaries, `sanitizeErrorMessage` ReDoS guard | none touch cleanup/vacuum. This PR's own tests all pass in CI: the 8 tests of `tests/unit/db/cleanup-reclaim-freed-pages.test.ts` in shard 1/4 (better-sqlite3 driver), the `vacuum-scheduler.test.ts` additions and the updated `cleanup-column-fix.test.mjs` invariant in shard 2/4 |

Happy to rebase once the base is green again.

diegosouzapw:
This is a thorough, well-documented fix for a real and still-open problem
(#12821) — replacing the blocking full VACUUM with `auto_vacuum`-aware
incremental reclamation (and deferring to `vacuumScheduler`'s configured window
for `NONE`) is the right architecture, not just a threshold tweak. Ran your full
test suite at this PR's head: 33/33 pass across the INCREMENTAL/FULL/NONE
branches and the `vacuumScheduler` persistence tests.

One thing to flag: #13011 merged on 09-11 (after this PR's last commit) and
added a different, simpler row-count-based VACUUM gate at the exact same call
sites this PR rewrites — you'll hit a real rebase conflict there. Your
`reclaimFreedPages()` reading the actual freelist is a strictly better signal
than a row-count heuristic, so the right resolution is to replace that gate with
your logic rather than keep both. There's also a competing PR in the same
review batch (#13079) proposing a byte-threshold gate on the same call sites —
I'm recommending yours as the one to move forward given how much more complete
it is.

diegosouzapw:
Resolvi os conflitos e mergeei. Uma decisão precisa ficar registrada porque removi um teste do tip.

**O que conflitava:** 8 conflitos em 4 arquivos. O plan-file mandava reconciliar contra a #13011, mas o tip andou mais — a **#13079** (mergeada em 11/09) *estendeu* exatamente as funções que esta PR remove, acrescentando `getVacuumMinReclaimableBytes()`, `getReclaimableBytes()` e um gate por bytes dentro de `vacuumAfterCleanup()`.

**O teste removido:** `tests/unit/vacuum-reclaimable-threshold.test.ts`, o guarda da #13079. Ele testa só os três símbolos que esta PR remove, então não compilaria mais. Antes de aceitar, verifiquei que a remoção é legítima e não supressão de proteção:

- `vacuumAfterCleanup()` **não existe mais** — o gate perdeu o objeto que protegia.
- Não há mais `VACUUM` inline em `cleanup.ts`; os `exec("VACUUM")` remanescentes estão em caminhos distintos (acionado pelo operador, offload legado na subida, reparo, e o scheduler **diferido**, que é o desenho proposto aqui).
- `reclaimFreedPages()` lê **o mesmo sinal** que o gate da #13079 (`freelist_count` × `page_size`), então a proteção contra o incidente original — limpeza que libera 2–6 linhas disparando VACUUM completo e congelando o tráfego por 15–20 min — continua de pé, por um mecanismo melhor.

**`vacuumScheduler.ts` foi combinado, não escolhido.** A interface e os `STATE_DEFAULTS` ficaram com os campos dos dois lados (`fullVacuumRequestedAt`/`fullVacuumRequestReason` seus + `autoVacuumDrift`/`lastReclaimedPages` da #13432). No `init()`, você extraiu o corpo para `hydrateFromPersistedState()` e o tip inseriu uma linha dentro desse corpo — portei a linha do tip **para dentro** da sua função nova, em vez de concatenar. O `issue-13432-auto-vacuum-drift` passa 3/3, o que prova que a feature do tip sobreviveu.

**Validação:** 150 testes verdes em 18 arquivos — os 5 seus mais os 13 que importam `cleanup.ts`/`vacuumScheduler.ts`/`reclaimFreedPages.ts`. `typecheck:core` limpo.

Issue #12007, opened 2026-08-29
fix(providers): Antigravity tool names break Claude Code — hashed MCP ids emitted by our own sanitizer


Seramicx:
## Summary

When using **Claude Code** through OmniRoute with **Antigravity-routed models**, the assistant frequently emits **invalid tool names**, causing Claude Code to reject every tool call with:

```text
Error: No such tool available: <name>
```

This makes Antigravity models effectively unusable for agentic Claude Code sessions (edits, reads, skills, MCP), even though plain chat and shell commands may still work.

## Environment

- Client: **Claude Code** (Anthropic Messages API + tool_use)
- Gateway: OmniRoute (`ANTHROPIC_BASE_URL` → local proxy)
- Provider: `antigravity/*` (e.g. `antigravity/claude-opus-4-6`)
- OmniRoute version observed: **3.8.49**
- OS: Windows 11

## Symptoms

In a single Claude Code session, tool dispatch failed repeatedly:

| Upstream model (via OmniRoute) | Invalid tool names emitted | Count |
|---|---|---|
| `gemini-3.1-pro-low` | `edit` (should be `Edit`) | 11 |
| `claude-opus-4-6-thinking` | `read`, `skill` (should be `Read`, `Skill`) | 15 + 8 |
| `claude-opus-4-6-thinking` | `mcp_crawl4ai_md`, `mcp_crawl4ai_md_84daa8e7…` (hallucinated; real name is `mcp__crawl4ai__md`) | 6 |

Claude Code tool names are **case-sensitive**. Lowercase `read`/`edit`/`skill` are not aliases for `Read`/`Edit`/`Skill`.

After many failures, the model started **guessing tool names with random hash suffixes** (e.g. `mcp_crawl4ai_md_84daa8e747d7161c`), which do not exist in the tool registry.

**Workaround:** start a **new session** with a non-Antigravity model route.

## Expected behavior

When OmniRoute proxies an Anthropic-compatible Claude Code request to Antigravity:

1. The model should call tools using the **exact `name` values** from the `tools` block (`Read`, `Edit`, `Skill`, `mcp__namespace__tool`, etc.).
2. Models that cannot reliably follow tool schemas should be **documented or flagged** as unsuitable for agentic Claude Code use.
3. Optionally, OmniRoute could apply a **compatibility shim** for common mistakes (see suggestions below).

## Actual behavior

[...]

PR #12839, opened 2026-09-06
fix(responses): preserve provider aliases alongside namespace identities

Pllutonyy:
## Summary

- Preserve both request-scoped tool metadata maps before stripping their transport side channels.
- Responses pivots carry object-valued namespace identities plus string-valued provider aliases. The previous extraction preferred the namespace map and deleted the alias map; its recovery fallback could not recover strings from namespace objects.
- Add `extractRequestToolMetadata` and use its captured alias ledger in chatCore, while retaining later provider-specific map precedence and legacy extraction behavior for other callers.
- No sanitizer changes, guessed name mappings, auth changes, new provider or runtime feature flag.

## Related Issues

- Related-issue state UNVERIFIED at submission: recheck live #12007 and #9780/#9783 and link the correct issues. Earlier notes conflict on whether #12007 is closed or open.
- Base-red inherited: #12732.

## Validation

- [x] Change type: routing / provider translation
- [x] Red-before-green: four new namespace/alias cases reproduced the loss before the fix; the namespace-only guard already passed.
- [x] Focused five-file suite: 30 passed; one additional parallel-MCP regression subsequently passed (31 distinct checks across those runs).
- [x] Scoped ESLint on both production files and the new test file.
- [x] Changed metadata helper coverage: 100% lines, branches and functions in the focused coverage run.
- [x] Live Responses SSE namespace/collision identity and tool-result continuation.
- [x] Native Codex CLI 0.153.3: real Graft freshness and OpenViking health calls completed, final exact marker returned, exit 0.
- [ ] Full lint/CI/coverage matrix and production packaging.
- [x] Production changes include a regression test.
- [x] `npm run typecheck:core`: clean.
- [x] `check:provider-consistency` and `check:provider-assets`: pass.
- [x] `git diff --check`: clean.

## Tests Added Or Updated

- `tests/unit/antigravity-responses-tool-metadata-roundtrip.test.ts` exercises real request and response translators, the chatCore metadata-extraction helper, exact namespace/name restoration on added/done/completed events, genuine sanitizer collisions, four-turn replay without hash growth, alias-only and namespace-only requests, and parallel native MCP calls following reasoning.

## Coverage Notes

- The changed `open-sse/handlers/chatCore/requestToolIdentity.ts` reached 100% line/branch/function coverage. Live HTTP and native-client tests cover the new chatCore wiring; they do not replace the full cross-platform build/CI matrix.

## Reviewer Notes

- The live tests ran the isolated worktree's actual Next HTTP server in full profile on loopback, not a modified production deployment.
- Model: `antigravity/gemini-3.1-pro-low`, explicitly present in the account's synced catalog. This is evidence for that route, not every Antigravity model or multi-account behavior.
- Cleanup: temporary credentials inactive/cleared, child processes exited, test listeners closed. The running full build remains at `2265ce761f76b6c81e34515cc8483ca127d4b414`.
- This does not establish Claude Code continuation, multi-account/long-session parity or release readiness. Nothing was promoted.

### DIFF
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -1,5 +1,5 @@
 import {
-  extractRequestToolIdentityMap,
+  extractRequestToolMetadata,
   resolveResponseToolNameMap,
 } from "./chatCore/requestToolIdentity.ts";
 import { injectMemoryAndSkills } from "./chatCore/memorySkillsInjection.ts";
@@ -2600,8 +2600,9 @@ export async function handleChatCore({
   // Keep the request translator's namespace identities separate from toolNameMap:
   // the latter is a Kiro/Claude passthrough alias channel with string values,
   // while namespace identities carry `{namespace, name}` for the #7936 response
-  // seam. Extract first because Kiro merge may reuse `_toolNameMap` below.
-  const requestToolIdentityMap = extractRequestToolIdentityMap(translatedBody);
+  // seam. Capture both before stripping their side channels: a Responses ->
+  // Gemini/Antigravity pivot carries both maps, not one recoverable ledger.
+  const { requestToolIdentityMap, toolNameAliasMap } = extractRequestToolMetadata(translatedBody);
 
   // Kiro: sanitize tool schemas before dispatch. Kiro returns 400 "Improperly
   // formed request" for unsupported JSON-Schema keywords (anyOf/$ref/if-then,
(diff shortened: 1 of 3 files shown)


diegosouzapw:
This looks like the safer, more conservative half of the namespace/alias overlap with
#12928 — capturing both the alias ledger and the namespace-identity ledger before either
side channel is stripped, without introducing any name-guessing heuristics. The new
roundtrip test is thorough and exercises real translator code. Two asks before merge: add a
changelog fragment, and let's confirm merge order against #12928 (which touches a
different part of the same file) so the two don't fight each other. Queuing for a final
test run.

PR #12841, opened 2026-09-06
fix(sse): fix double-escaped tabs in Codex JSON tool call arguments

### LINKED ISSUE #12831: feat(codex): gpt-5.6-luna-xhigh double-escapes tabs in tool call arguments
### OmniRoute Version

3.8.50

### Installation Method

npm (global)

### Operating System

Linux / Ubuntu 22.04

### Node.js Version

v24.20.0

### Provider(s) Involved

- codex

### Model(s) Involved

- `codex/gpt-5.6-luna-xhigh`

### Client Tool

OMP (oh-my-picus) agentic sessions using the `_edit` tool

### Description

Running agentic coding sessions through a combo that includes `gpt-5.6-luna-xhigh`, I started noticing edits being applied with literal `\t` strings in the source instead of actual tab characters. Took a while to track down, but the issue is in how this model serializes the `_edit` tool arguments — it's double-escaping tabs.

Instead of outputting `\t` (a real tab in the JSON string), it outputs `\\t` (backslash followed by `t`). So when the agent parses the JSON and applies the edit, it writes the two-character sequence `\t` into the file.

Checked across ~10k requests over a day:

- `gpt-5.6-luna-xhigh`: 11 out of 84 `_edit` calls had this (~13%)
- `kiro/claude-sonnet-4.6`: 0 out of 140 `_edit` calls

Only shows up with this model. Newlines in the same arguments are encoded correctly, just tabs.

### Steps to Reproduce

1. Set up a combo routing through `codex/gpt-5.6-luna-xhigh`
2. Use it in an agentic session editing TypeScript or JavaScript files with tab indentation
3. Check the raw `function.arguments` from the response — some will have `\\t` where you'd expect `\t`

Seems to happen more on shorter diffs (127-288 output tokens in the affected cases).

### Observed Behaviour

Raw `function.arguments` from the stream:

```json
{
  "i": "Fix error handler indentation",
  "input": "[src/middleware.ts]\nPUT 177.=178:\n+\\t} catch {\n+\\t\\tresponse = new Response(ERROR_HTML, {"
}
```

### Expected Behaviour

```json
{
  "i": "Fix error handler indentation",
  "input": "[src/middleware.ts]\nPUT 177.=178:\n+\t} catch {\n+\t\tresponse = new Response(ERROR_HTML, {"
}
```

### Additional Context

Same tool schema, same setup, other models work fine. A workaround at the OmniRoute level could be unescaping `\\t` to `\t` in `function.arguments` for Codex responses, similar to other normalizations already in `open-sse/`.

### Test Impact

Would need a unit test covering this normalization for Codex tool call arguments.

rafacpti23:
Fixes #12831. When the Codex upstream model produces string values in tool arguments that contain double-escaped tabs (\t inside the JSON string instead of \t), the parser outputs literal backslash-t characters. This breaks editor patches that rely on proper indentation. This commit adds a fixDoubleEscapedTabs sanitization step before parsing to restore them to single tabs.

## Summary

## Related Issues

- Closes #
- Related to #

## Validation

- [ ] `npm run lint`

## Tests Added Or Updated

## Coverage Notes

## Reviewer Notes

### DIFF
--- a/open-sse/translator/response/openai-responses.ts
+++ b/open-sse/translator/response/openai-responses.ts
@@ -517,7 +517,7 @@ function emitToolCall(state, emit, tc) {
   if (tc.function?.arguments) {
     const refCallId = state.funcCallIds[tcIdx] || newCallId;
     const existingArgs = state.funcArgsBuf[tcIdx] || "";
-    const sanitized = escapeJsonStringValues(tc.function.arguments);
+    const sanitized = escapeJsonStringValues(fixDoubleEscapedTabs(tc.function.arguments));
     const nextArgs = appendToolCallArgumentDelta(existingArgs, sanitized);
     const emittedDelta = nextArgs.slice(existingArgs.length);
     state.funcArgsBuf[tcIdx] = nextArgs;
--- a/tests/unit/bug-12831.test.ts
+++ b/tests/unit/bug-12831.test.ts
@@ -1,0 +1,90 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { openaiToOpenAIResponsesResponse } from "../../open-sse/translator/response/openai-responses.ts";
+
+test("Issue #12831: fixes double-escaped tabs in Codex JSON tool call arguments", () => {
+  const events = [];
+  const emit = (name, payload) => events.push(payload);
+  const state = {
+    responseId: "res_123",
+    funcCallIds: {},
+    funcNames: {},
+    funcArgsBuf: {},
+    funcArgsDone: {},
+    funcItemAdded: {},
+    funcItemDone: {},
+    msgItemAdded: {},
+    msgContentAdded: {},
+    msgTextBuf: {},
+    msgItemDone: {},
+  };
+
+  const chunk1 = {
+    choices: [
+      {
+        index: 0,
+        delta: {
+          tool_calls: [
+            {
+              index: 0,
+              id: "call_123",
+              function: {
+                name: "_edit",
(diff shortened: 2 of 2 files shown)


diegosouzapw:
Thanks for tackling #12831 — the double-escaped-tabs problem is real, but as submitted this
PR doesn't run: `fixDoubleEscapedTabs` is called at `openai-responses.ts:520` but is never
defined or imported anywhere in the diff or the base branch. I ran your own new test
(`tests/unit/bug-12831.test.ts`) against your PR head and it fails immediately with
`ReferenceError: fixDoubleEscapedTabs is not defined` — since `emitToolCall()` runs for
every streamed tool call with arguments, this would break tool-call streaming entirely if
merged, not just the double-escaped-tab case. Looks like the actual sanitizer implementation
didn't make it into the commit. Could you add the missing function, confirm the test passes
locally, and add a changelog fragment? Happy to take another look once it's green.

diegosouzapw:
Thanks @rafacpti23 — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

PR #12843, opened 2026-09-06
feat(sse): add Dario provider limits adapter

andrewmalov:
## Summary
- add a statically allowlisted Dario usage capability for Anthropic-compatible API-key connections
- normalize fresh Dario 5-hour and 7-day utilization into Provider Limits windows
- make Provider Limits and quota-aware routing resolve fetchers per connection
- preserve connection-scoped caches and last-known-good reset-aware snapshots
- add SSRF, credential-boundary, freshness, pool-selection, visibility, and preflight regression tests

## Verification
- focused Dario/quota tests: 49 passed
- `npm run lint`: passed
- `npm run test:vitest`: 44 files / 405 tests passed
- `npm run build`: passed
- `npm run typecheck:noimplicit:core`: inherited failures in unchanged `open-sse/translator/response/openai-responses/pureHelpers.ts`
- `npm run test:unit`: base v3.8.50 has 8 unrelated failures; focused Dario suite passed

The live demo deployment and Dario ingress hardening are maintained separately from this upstream code change.


diegosouzapw:
Thanks for the thorough work here — the SSRF/credential-boundary test coverage is great. I
compared this against the current release tip and Dario was actually shipped as a full embedded
service in v3.8.50 (#8523) via a different path: executor, combo/headroom integration,
dashboard, and a migration numbered 138 already exist there too (which collides with this PR's
own migration 138 — different content, same number). Given that, I don't think we can merge
this as-is, but if the specific "normalize 5-hour/7-day Dario utilization into Provider Limits
windows" piece isn't covered by what's already live, I'd welcome a much smaller follow-up PR
just for that, and you'd keep authorship credit on it.

PR #12846, opened 2026-09-06
docs(agent): define proportionate task verification levels

elevatevietnamai-lang:
Agents currently have path-based CI classification and impacted-test selection, but no authoritative risk level that maps a task to required skills and proportionate local verification. This adds an L0-L4 policy: localized L0/L1 work uses focused and impacted checks, while trust boundaries, persisted data, public contracts, shared infrastructure, and release work escalate to deeper domain or release suites.

The policy keeps CI authoritative, treats unknown scope conservatively, and requires missing commands or unavailable environments to be reported instead of replaced or silently skipped.

Validation:
- `npm run check:docs-all`
- `git diff --check`
- Independent loop-verifier review: APPROVE


elevatevietnamai-lang:
Superseded by #12847, which is based directly on release/v3.8.51 and contains only the task-level verification policy commit.

PR #12855, opened 2026-09-06
fix(claude): passthrough tool_use names must match client-declared casing

Neuron-Mr-White:
## Problem

Every Claude-format **streaming** passthrough rewrites `tool_use` names of well-known Claude Code tools (`bash` → `Bash`) even when the client declared them lowercase.

Live repro (self-hosted router, `devin-cli-agentic` executor, model `glm-5-2`): a third-party Anthropic-format client (pi) declares `bash/read/edit/write` and streams. Turn 1 returns `tool_use name:"Bash"` — a name the client never declared. Client tool dispatch fails; echoing the history back then hard-fails the whole session:

```json
400 {"Historical tool_use references undeclared tool: Bash","type":"devin_agentic_error","code":"undeclared_historical_tool"}
```

Reproduced deterministically: identical request body with `stream:true` → `"Bash"`; `stream:false` → `"bash"` (the JSON path does not run the passthrough restore). Reproduced in a pristine microVM with a fresh pi install, proving it is router-side.

## Root cause

`restoreClaudePassthroughToolUseName` (open-sse/utils/stream.ts) calls `restoreClaudeToolName(name, map)` with a map that is null/empty whenever the request path performed no tool renames. With no map, `restoreClaudeToolName` falls through to the canonical-upgrade fallback (`TOOL_RENAME_MAP[lower]`), i.e. `bash` → `Bash`. That fallback is load-bearing for genuine Claude Code clients (declared PascalCase, upstream downcased — #7926), but harmful for everyone else.

Separately, `serializeAnthropicForDevin` validates historical `tool_use` names case-sensitively, so the echoed `Bash` kills the turn instead of resolving to the declared `bash`.

## Fix

1. `restoreClaudePassthroughToolUseName` now takes the request `tools` and resolves names in order:
   - alias map first (renamed → original spelling, unchanged behavior), then
   - normalize upstream case drift to the **client-declared casing** (exact, then case-insensitive), then
   - verbatim — it no longer canonicalizes names that match no declared tool.
   - Claude Code clients keep #7926 protection: upstream downcase is restored to their declared PascalCase.
2. `serializeAnthropicForDevin`: case-insensitive fallback for historical `tool_use` names; renders the declared spelling into the Devin prompt. Case drift can no longer 400 a turn (`undeclared_historical_tool`).

## What does NOT change

- Request-side remapping (`remapToolNamesInRequest`), the alias-map restore for renamed flows, non-streaming JSON responses, translate-mode translators (they pass their own maps), and OpenAI-format responses.

## Tests

- `tests/unit/claude-passthrough-tool-name-mapless-leak.test.ts` — 8 cases: no upgrade for lowercase-declared; downcase PascalCase echo to declared lowercase; #7926 PascalCase restore preserved; alias map precedence; undeclared names verbatim; OpenAI-style `function.name`; non-tool_use blocks.
- `tests/unit/devin-agentic-serializer-case-insensitive-history.test.ts` — PascalCase echo accepted + declared spelling rendered; exact-case unchanged; genuinely undeclared tool still rejected.
- Full existing suites pass: claude-tool-name-casing, claude-code-parity, claude-code-tool-remapper-flag-leak, chatcore-passthrough-tool-names, tool-name-case-preserve-4307, executor-devin-cli* (103 tests).

## Caller migration note

Clients that previously received PascalCase names by accident (and worked around it by dispatching case-insensitively) will now see their declared spelling — the correct contract.

### DIFF
--- a/open-sse/executors/devin-agentic/serializer.ts
+++ b/open-sse/executors/devin-agentic/serializer.ts
@@ -61,7 +61,15 @@ function serializeBlock(
         id ? "duplicate_tool_use_id" : "missing_tool_use_id"
       );
     }
-    const declared = tools.find((tool) => tool.name === name);
+    // Exact match first; fall back to case-insensitive matching so a client
+    // echoing back a differently-cased name for the same tool (e.g. a router
+    // layer handing a Claude Code canonical "Bash" to a client that declared
+    // "bash") is normalized instead of hard-failing the whole turn with
+    // undeclared_historical_tool (#12721). The declared casing is rendered so
+    // the downstream Devin prompt always shows the catalog name verbatim.
+    const declared =
+      tools.find((tool) => tool.name === name) ??
+      tools.find((tool) => tool.name.toLowerCase() === name.toLowerCase() && name !== "");
     if (!declared) {
       throw new DevinAgenticBridgeError(
         `Historical tool_use references undeclared tool: ${name || "unknown"}`,
@@ -72,7 +80,7 @@ function serializeBlock(
     return [
       "[Assistant Tool Use]",
       `id: ${id}`,
-      `name: ${name}`,
+      `name: ${declared.name}`,
       "arguments:",
       JSON.stringify(record.input || {}, null, 2),
     ].join("\n");
--- a/open-sse/utils/stream.ts
+++ b/open-sse/utils/stream.ts
@@ -603,27 +603,40 @@ export function restoreClaudePassthroughToolUseName(
       : null;
   if (!block || block.type !== "tool_use" || typeof block.name !== "string") return false;
 
-  // Alias map first: when the request path renamed tools, its entries map the
-  // upstream (renamed) spelling back to the CLIENT's original spelling — the
-  // only source that knows it.
   const map = toolNameMap instanceof Map ? toolNameMap : null;
+
+  // 1) Alias ledger, direct lookups only. restoreClaudeToolName() is NOT used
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Solid diagnosis — the SSE-vs-JSON divergence (declared `"bash"` echoed back as `"Bash"` only
on the streaming path) is a clean tell for exactly the bug you describe, and the fix
preserves the #7926 Claude Code protection with an explicit regression test. Re-ran both new
test files at PR head: 13/13 pass. Two asks before merge: rebase onto the current
`release/v3.8.51` tip (currently shows as conflicting — the CI run on file is from 09-06,
well before this week's merge wave) and add a changelog fragment. Would also be nice to trim
the incidental Prettier reformatting of unrelated lines in `stream.ts` if you're rebasing
anyway, just to keep the diff focused.

diegosouzapw:
Thanks @Neuron-Mr-White — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

PR #12775, opened 2026-09-05
[defer] feat: add Gemini prompt cache accounting

jbovard2016:
## Summary
- add a default-off local implicit-cache mode scoped to concrete Antigravity/AGY Gemini targets
- preserve cache evidence across wrapped, unwrapped, streaming, OpenAI, and Claude usage paths
- keep existing prompt-cache affinity while proving runtime-ineligible preferred accounts fall through
- persist bounded stream/non-stream cache evidence in call-log metadata without leaking internal fields to clients

## Verification
- focused source and observability suite: 323/323 passing
- full Vitest suite: 51 files, 465 tests passing
- core and no-implicit-any type checks passing
- package policy: 3,513 entries checked; 1,305 MCP source files checked
- diff check passing

## Known base and environment limits
⚠️ base-red inherited: #12732

Local full Node testing stalls in the inherited Adobe Firefly browser test. Local release builds were attempted with Turbopack and constrained webpack; worker processes were killed under WSL memory pressure, so no `dist/` was produced. This PR exists so clean CI runners can execute the build and package gates. No live Gemini pilot or production deployment has run.

## Release boundary
Release 1 covers implicit caching only. Explicit cache resources remain out of scope. Package install/upgrade/restart/rollback and the bounded live pilot remain required before promotion.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12863, opened 2026-09-06
fix(translator): stop double-counting cached tokens in Gemini to Claude usage

### LINKED ISSUE #12862: fix(providers): Gemini to Claude translator double-counts cached prompt tokens in usage
### OmniRoute Version

3.8.50

### Installation Method

npm (global)

### Operating System

Linux

### Node.js Version

v22.22.2

### Provider(s) Involved

Antigravity (Gemini)

### Model(s) Involved

gemini-2.5-pro

### Client Tool

Claude Code

### Description

The Gemini to Claude response translator reports `input_tokens` as the full
`promptTokenCount` while also reporting `cache_read_input_tokens` separately. In the
Anthropic Messages API these two fields are disjoint: `input_tokens` counts only the
tokens that were **not** read from cache, and the total prompt is
`input_tokens + cache_read_input_tokens`.

Because the translator emits the full prompt in `input_tokens` and then repeats the
cached part in `cache_read_input_tokens`, every consumer that sums the two fields sees
roughly twice the real prompt size once the cache warms up.

The offending line is `open-sse/translator/response/gemini-to-claude.ts:400`:

```ts
state.usage = {
  input_tokens: inputTokens,
  output_tokens: candidatesTokens + thoughtsTokens,
};
if (cachedTokens > 0) {
  state.usage.cache_read_input_tokens = cachedTokens;
}
```

The sibling translator `gemini-to-openai.ts` is not affected: OpenAI's `prompt_tokens`
is defined as the full prompt including the cached part, so passing `promptTokenCount`
through is correct there.

OmniRoute's own accounting already assumes the Anthropic semantics. In
`src/lib/usage/tokenAccounting.ts` the cache-read tokens are added back on top of
`input_tokens` when computing the billed total, which double-counts the cached prefix
for this provider.

### Steps to Reproduce

1. Route Claude Code through OmniRoute to an Antigravity (Gemini) connection.
2. Send a first turn so a prompt cache prefix is created.
3. Send a second turn that hits the cache.
4. Inspect the `call_logs` row for the second request, or the `usage` block of the
   Claude-format response.

### Expected Behavior

For the cached second turn, `input_tokens` should hold only the uncached remainder, so
that `input_tokens + cache_read_input_tokens` equals the real prompt size.

### Actual Behavior

Observed on a two-turn run through this path:

| Request | tokens_in | tokens_cache_read | Real prompt |
| --- | --- | --- | --- |
| turn 1 (cold) | 72,591 | 0 | 72,591 |
| turn 2 (cached) | 148,095 | 72,183 | ~75,912 |

The second turn reports 148,095 input tokens for a prompt that actually grew by only a
few thousand tokens over the first one. The visible consequence downstream is that
Claude Code's auto-compaction fires at roughly half of the real context window, because
the client believes the prompt is about twice its real size.

### Test Impact

### Additional Context

`tests/unit/translator-resp-gemini-to-claude.test.ts` encodes the current behavior at the
`MAX_TOKENS` case: with `promptTokenCount: 5` and `cachedContentTokenCount: 1` it asserts
`input_tokens === 5`. Under the Anthropic semantics that assert should read `4`.

A PR with the one-line fix and test coverage follows.

ThiagoMafra-Integrare:
## Summary

- `gemini-to-claude.ts` reported the full `promptTokenCount` in `input_tokens` while also
  reporting the cached prefix in `cache_read_input_tokens`. In the Anthropic Messages API
  those two fields are disjoint, so any consumer that sums them saw roughly twice the real
  prompt once the cache warmed up. `input_tokens` now excludes the cached tokens.
- User-visible effect: clients driven by the reported usage (Claude Code auto-compaction,
  cost accounting in `src/lib/usage/tokenAccounting.ts`, which adds `cache_read` back on
  top of `input_tokens`) stop seeing an inflated prompt on cached turns.
- `gemini-to-openai.ts` is deliberately untouched: OpenAI's `prompt_tokens` includes the
  cached part, so passing `promptTokenCount` through is correct there.

## Related Issues

- Closes #12862

## Validation

- [x] Change type: provider
- [x] `npm run lint`

Focused loop on `release/v3.8.51` at `f9a1cc8a`:

```
node --import tsx/esm --test tests/unit/translator-resp-gemini-to-claude.test.ts
# tests 8
# pass 8
# fail 0
```

Without the production change, the same file reports `# fail 2` (the updated
`MAX_TOKENS` assert and the new cached-tokens case), so the tests do exercise the fix.

## Tests Added Or Updated

- `tests/unit/translator-resp-gemini-to-claude.test.ts`
  - updated: the `MAX_TOKENS` case asserted `input_tokens === 5` for
    `promptTokenCount: 5, cachedContentTokenCount: 1`; it now asserts `4`.
  - added: `input_tokens excludes cached tokens` — `promptTokenCount: 100`,
    `cachedContentTokenCount: 90` yields `input_tokens: 10` and
    `cache_read_input_tokens: 90`.
  - added: `input_tokens equals promptTokenCount when nothing is cached` — regression
    guard for the uncached path, where the value must stay untouched.

## Coverage Notes

- The change is one line in `open-sse/translator/response/gemini-to-claude.ts`, inside the
  usage-metadata branch. That branch is covered by the three tests listed above: the
  cached path, the uncached path, and the pre-existing `MAX_TOKENS` path.
- Coverage of the touched file does not move down; the usage branch gains two cases.

## Reviewer Notes

- `Math.max(0, ...)` guards against a provider reporting `cachedContentTokenCount` greater
  than `promptTokenCount`, which would otherwise yield a negative `input_tokens`.
- No migration, no feature flag, no config surface. The only behavioral change is the
  value reported in `usage.input_tokens` for the Gemini to Claude path.

### DIFF
--- a/changelog.d/fixes/12863-gemini-to-claude-cached-input-tokens.md
+++ b/changelog.d/fixes/12863-gemini-to-claude-cached-input-tokens.md
@@ -1,0 +1,1 @@
+- **fix(translator):** Gemini to Claude usage no longer double-counts the cached prompt prefix — `input_tokens` now excludes `cache_read_input_tokens`, matching the Anthropic Messages semantics ([#12863](https://github.com/diegosouzapw/OmniRoute/pull/12863))
--- a/open-sse/translator/response/gemini-to-claude.ts
+++ b/open-sse/translator/response/gemini-to-claude.ts
@@ -397,7 +397,7 @@ export function geminiToClaudeResponse(chunk, state) {
       typeof usageMeta.cachedContentTokenCount === "number" ? usageMeta.cachedContentTokenCount : 0;
 
     state.usage = {
-      input_tokens: inputTokens,
+      input_tokens: Math.max(0, inputTokens - cachedTokens),
       output_tokens: candidatesTokens + thoughtsTokens,
     };
     if (cachedTokens > 0) {
--- a/tests/unit/translator-resp-gemini-to-claude.test.ts
+++ b/tests/unit/translator-resp-gemini-to-claude.test.ts
@@ -107,7 +107,7 @@ test("Gemini -> Claude stream: functionCall becomes tool_use and MAX_TOKENS maps
   assert.equal(result[2].delta.partial_json, JSON.stringify({ path: "/tmp/a" }));
   assert.equal(result[3].type, "content_block_stop");
   assert.equal(result[4].delta.stop_reason, "tool_use");
-  assert.equal(result[4].usage.input_tokens, 5);
+  assert.equal(result[4].usage.input_tokens, 4);
   assert.equal(result[4].usage.output_tokens, 5);
   assert.equal(result[4].usage.cache_read_input_tokens, 1);
   assert.equal(result[5].type, "message_stop");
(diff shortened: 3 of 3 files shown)


diegosouzapw:
Ran your test file at the PR head — 8/8 pass, including the updated MAX_TOKENS case and the
two new cached/uncached cases. Clean, minimal, correctly scoped fix that closes #12862.
Merging this now — there's a much larger PR (#12775) open that touches the same usage block
as part of a bigger caching feature, but that one needs its own design/scope decision first,
so this focused fix goes in independently.

diegosouzapw:
Thanks @ThiagoMafra-Integrare — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #12864, opened 2026-09-06
fix(sse): Anthropic OAuth 403 "Request not allowed" is a per-request refusal — cooldown with backoff instead of an instant ban (#12859)

### LINKED ISSUE #12859: fix(resilience): errorClassifier: Anthropic OAuth 403 "Request not allowed" permanently bans the Claude connection on a single response
## Summary

One upstream `403 {"type":"permission_error","message":"Request not allowed"}` on a `claude` OAuth connection is classified `FORBIDDEN` and persisted as the terminal `banned` connection state. From then on every request to the provider is answered locally, without contacting Anthropic:

```
[claude] All 1 connection(s) banned by upstream — please reconnect in the dashboard   (HTTP 401)
```

until an operator reconnects in the dashboard.

That 403 is a per-request refusal, not a dead credential: the same token serves requests immediately before and after it, and Anthropic reports a revoked/expired token as `401 authentication_error`. It is also rare (order of one per several thousand requests), so a terminal ban on the first occurrence converts a one-request hiccup into a provider outage that lasts until someone notices.

## Root cause

- `open-sse/services/errorClassifier.ts` → `classifyProviderError()`: the 403 branch carves out Cloudflare fingerprint rejections, Google project/region errors, Kiro profile errors and API-key providers, then falls through to `return PROVIDER_ERROR_TYPES.FORBIDDEN` for every OAuth provider — this body included.
- `open-sse/handlers/chatCore.ts`, `errorType === PROVIDER_ERROR_TYPES.FORBIDDEN` branch: `writeTerminalStatus(connectionId, { testStatus: "banned", isActive: false, ... })`.
- `src/sse/handlers/chatHelpers.ts`, `credentials.allExpired` branch: with the only connection in a terminal status, every request returns the 401 above before any upstream call.

Observed connection row after the event: `test_status = banned`, `error_code = 403`, `last_error = Request not allowed`, `last_error_type = forbidden`, `is_active = 1`, token not expired, circuit breaker `CLOSED` with `failure_count = 0`. A reconnect restores service immediately.

## Expected

Neither "ban on first sight" nor "ignore it": if the refusal ever became systematic, re-sending every request into it would be the wrong thing to do to an OAuth account. A behaviour that fits both cases:

1. Classify Anthropic's `Request not allowed` as a non-terminal, per-request error (the `FINGERPRINT_REJECTION` precedent for Cloudflare 1010): the failing request falls through to the next combo target, nothing terminal is written.
2. Put the connection into a bounded cooldown with backoff (e.g. 5 → 15 → 45 min, capped at 2 h, through the existing `rateLimitedUntil` / `backoff_level` machinery, as `GEO_BLOCKED` does), so a sporadic refusal costs minutes and a systematic one cannot become a stream of 403s.
3. Escalate to `banned` only after N consecutive refusals with no success in between (N = 3), with a `last_error` saying so.

## Environment

OmniRoute v3.8.51, `claude` provider, single OAuth connection, requests in Claude format via combos.

PR to follow.

insoln:
Fixes #12859.

## Problem

One upstream `403 {"type":"permission_error","message":"Request not allowed"}` on a `claude` OAuth connection is classified `FORBIDDEN` in `classifyProviderError()` and written by `chatCore` as the terminal `banned` connection state. Every later request is then answered locally with `[claude] All 1 connection(s) banned by upstream — please reconnect in the dashboard` (401) until an operator reconnects. The response is a per-request refusal on a healthy token (200 before and after it on the same connection; a dead token is a `401 authentication_error`), and it is rare — order of one per several thousand requests.

## Fix

Neither "ban on first sight" nor "ignore it": a run of these refusals would be enforcement, and re-sending every request into it is the wrong thing to do to an OAuth account.

- `open-sse/services/errorClassifier.ts`: new non-terminal `PROVIDER_ERROR_TYPES.REQUEST_REJECTED` ("the upstream refused this request, not the credential"). In the 403 branch, a body matching `Request not allowed` from provider `claude` returns it instead of falling through to `FORBIDDEN` (helpers `isAnthropicOAuthProvider()` / `isAnthropicRequestNotAllowed()` keep the scope explicit).
- `src/sse/services/authTerminalStatus.ts`: `REQUEST_REJECTED` joins `isNonTerminalProviderError()`, so `resolveTerminalConnectionStatus()` yields `null` for it even when the fallback analysis says `permanent` — the `FINGERPRINT_REJECTION` precedent.
- `open-sse/handlers/chatCore/requestRejectedFailure.ts` (leaf; `chatCore.ts` is at its frozen size): on `REQUEST_REJECTED`
  - exclude the connection for a growing cooldown — **5 min after the first refusal, 15 min after the second** (`COOLDOWN_MS.requestRejected` / `requestRejectedRepeat`, next to the existing 5 s `request_not_allowed` text rule, which this supersedes for `claude`) — written like every other connection-level cooldown: ISO `rateLimitedUntil` + `testStatus: "unavailable"` + `lastErrorAt`, so the dashboard shows the countdown and the recovery tick restores `active`;
  - escalate to `banned` only on the **third consecutive refusal**, with a `last_error` that says so.
- `open-sse/services/requestRejectedStreak.ts`: the per-connection streak. It counts cooldown *episodes*: a refusal that lands while the connection is already excluded (parallel sessions / fan-out that were in flight when the first 403 came back) is the same event and is not counted. A successful response on the connection clears the streak (`chatHelpers` `onRequestSuccess`), as does clearing the cooldown by hand in the dashboard; the recovery tick does not (an elapsed cooldown is not a success). A 24 h staleness bound keeps an ancient refusal from starting a streak. State is in-memory; after a restart one refusal is re-seeded from the persisted row (`last_error_type` / `last_error_at`) so a crash loop cannot reset the count on every boot.
- Probe-origin refusals (dashboard test-all) are recorded on the connection but never touch the streak, cool it down or ban it (#9817).
- `open-sse/handlers/chatCore/connectionCooldown.ts`: the `GEO_BLOCKED` and `GCP_PROJECT_REQUIRED` branches were the same eight statements with different constants and wording; both now go through `excludeConnectionForCooldown()` (behaviour, probe guard and log lines preserved verbatim). `chatCore.ts` is at the file-size ratchet's frozen 5984 lines, so the new branch could not add a net line — after this extraction it ends 9 lines below the base.

The failing request itself still falls through to the next combo target, as before. Every other `claude` 403 body keeps its previous classification, and the same body from any other provider still classifies as before.

Bound on upstream hits per refusing connection: 1 per cooldown episode, 3 episodes then `banned` — versus 1 then a manual reconnect before this PR.

## Tests

- `tests/unit/anthropic-request-not-allowed-cooldown-escalation.test.ts` (new, end-to-end through the real chat route with a mocked upstream on a `claude` OAuth connection): 5 min cooldown after the 1st refusal, 15 min after the 2nd, `banned` + `is_active = 0` on the 3rd; a burst of refusals inside the first cooldown is one episode (no re-persist, no ban) and the next episode is the 2nd; a 200 clears the streak so the next refusal starts at 5 min again; a different claude 403 body still bans on the first response.
- `tests/unit/request-rejected-streak.test.ts` (new): ladder sourced from `COOLDOWN_MS`, episode de-duplication, escalation on the 3rd consecutive refusal, staleness window, success reset, per-connection isolation, restart seed.
- `tests/unit/anthropic-request-not-allowed-not-a-ban.test.ts` (new): `classifyProviderError` → `resolveTerminalConnectionStatus` gives `null` for the new type with `permanent: false` and `permanent: true`; a generic `claude` 403 still resolves to `banned`.
- `tests/unit/error-classifier.test.ts`: Anthropic body → `REQUEST_REJECTED`, never `FORBIDDEN`; the gateway-flattened message; the same body from a non-Anthropic provider stays `FORBIDDEN`; other `claude` 403s unchanged; helper edge cases.

All fail against the previous code. Also green locally: `check:file-size` (PR mode), `check:mutation-test-coverage --strict` (new tests registered in `stryker.conf.json`), `check:cycles`, `typecheck:core`, `eslint` on the changed files, `check:any-budget:t11`, `check:doc-links`, `check:fabricated-docs`, and the neighbouring suites `probe-gate-autodisable`, `8200-perplexity-web-401-cooldown`, `error-config`, `call-log-error-type`.

Docs: `docs/architecture/RESILIENCE_GUIDE.md` (terminal states) and `docs/architecture/CODEBASE_DOCUMENTATION.md` (resilience row) mention the new path.

⚠️ base-red inherited: #12732 (`glm.ts` TS2554, stale "169 migrations" doc counts, `reset-aware-model-family.md` fragment, `reset-aware-request-scope-12600` missing from `stryker.conf.json`, and the unrelated unit-test failures listed on #12830).


diegosouzapw:
This is a strong fix for a real production incident — the "one 403 among
hundreds of 200s bans the only Claude connection" failure mode is exactly
the kind of over-eager terminal state Hard Rule reasoning in AGENTS.md warns
about. The layered design (classifier → streak tracker → chatCore leaf) is
clean, and the streak escalation logic (same-episode dedup, window, seed
after restart) is well thought through. Probed all 4 new/changed test files
locally: 54/54 pass, including the explicit regression guard that a generic
Claude 403 still bans on the first response.

One blocker before merge: GitHub currently reports this PR as
`CONFLICTING` against `release/v3.8.51`, and I confirmed a real conflict in
`open-sse/handlers/chatCore.ts` around the GEO_BLOCKED/GCP_PROJECT_REQUIRED
branch block — looks like base drift from PRs merged into that file since
this branch was cut (2026-09-06). Could you rebase onto the current tip?
The conflict itself looks mechanical (your version replaces the old inline
block with `excludeConnectionForCooldown`), not a logic clash. Also heads
up: #13161 touches the same `tests/unit/error-classifier.test.ts` file
(different section) — whichever merges second will need a small manual
reconciliation of the appended test blocks.

Issue #9725, opened 2026-08-07
fix(backend): Streaming web-search interception returns tool_calls without execution result


xieccc:
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Windows

### Description

Web-search interception successfully converts the native `web_search`
tool into `omniroute_web_search`, and the upstream Claude model emits
tool calls for that function.
However, with `stream: true`, OmniRoute returns those calls directly to
the client with `content: null` and `finish_reason: "tool_calls"`.
No final assistant response is produced. The client cannot execute the
tool because `omniroute_web_search` was injected internally by
OmniRoute and was not registered by the client.
This appears to leave the interception flow incomplete for streaming
requests. It may be a regression or an incomplete implementation of
the behavior described in #3384.

### Steps to Reproduce

1. Configure GitHub Copilot as an LLM provider.
2. Configure Exa as the available search provider.
3. Enable Web Search Interception for the GitHub Copilot provider or
   the target model.
4. Send an OpenAI Chat Completions request with:
   - `model: "github/claude-opus-5"`
   - `stream: true`
   - `tools: [{ "type": "web_search" }]`
5. Ask a question requiring current information, such as today's weather.
6. Inspect the OmniRoute provider and client responses.

### Expected Behavior

OmniRoute should:
1. Convert `web_search` to `omniroute_web_search`.
2. Buffer or consume the internal tool call.
3. Execute the configured search backend.
4. Inject the search result as a valid tool result.
5. Perform a follow-up LLM call.
6. Stream the final assistant answer to the client.
The internal `omniroute_web_search` tool call should not be exposed to
a client that did not register that tool.

### Actual Behavior

The provider returns:
```json
{
  "content": [
    {
      "type": "tool_use",
      "name": "omniroute_web_search"
    }
  ],
  "stop_reason": "tool_use"
}
```
OmniRoute returns to the client:
```json
{
  "message": {
    "role": "assistant",
    "content": null,
    "tool_calls": ["..."]
  },
  "finish_reason": "tool_calls"
}
```
LobeHub therefore displays no final message content.

### Test Impact

Needs a new integration test

### Additional Context

Related issue: #3384

The problem is reproducible with this request path:

LobeHub -> New API -> OmniRoute -> GitHub Copilot Claude

New API only normalizes the incoming tool declaration. OmniRoute
successfully injects omniroute_web_search, so the failure occurs
after the upstream model emits the internal tool call.

[...]

PR #12867, opened 2026-09-06
fix(chat): continue after a server-owned tool on Chat Completions

### LINKED ISSUE #12696: fix(skills): Chat Completions reply stops at an executed tool call, content stays null
# fix(skills): Chat Completions reply stops at an executed tool call, content stays null

### OmniRoute Version

3.8.51 (deployed image `deploy-v3.8.51-custom`, base `release/v3.8.51` @ `488f57e9d`)

### Installation Method

Docker / Podman (rootless quadlet)

### Operating System

Linux

### Provider(s) Involved

agy (Antigravity)

### Model(s) Involved

`agy/gemini-3.8-flash-high`, reached through a 9-account combo

### Client Tool

Plain HTTP client (`urllib`), and code-review-forge

### Description

With Memory enabled, a Chat Completions request that declared no tools at all
can come back with `content: null` and `finish_reason: "tool_calls"`.
OmniRoute injected `memory_search`, the model called it, OmniRoute executed
it, and then returned that turn to the caller. The result of the execution is
sitting right there in the response envelope. What never happens is the second
round trip that would let the model use it and answer.

The client is left holding a tool call it cannot service. `memory_search` is
OmniRoute's own tool; the client never registered it.

Measured response:

```json
{
  "id": "chatcmpl-1788499712958-94d0978f",
  "model": "gemini-3.8-flash-high",
  "choices": [{
    "finish_reason": "tool_calls",
    "message": {
      "role": "assistant",
      "content": null,
      "tool_calls": [{
        "id": "call_1128075",
        "type": "function",
        "function": {
          "name": "memory_search",
          "arguments": "{\"query\":\"review authentication code\"}"
        }
      }]
    }
  }],
  "usage": {"prompt_tokens": 5710, "completion_tokens": 18, "total_tokens": 6583},
  "tool_results": [{
    "tool_call_id": "call_1128075",
    "output": "{\"success\":true,\"data\":{\"memories\":[{...}]}}"
  }]
}
```

18 completion tokens is a complete tool call, not a truncated answer, and the
populated `tool_results` shows the execution itself worked.

This is the same symptom reported in #9725 ("returns tool_calls without
execution result", `content: null` with `finish_reason: "tool_calls"`), which
covered the streaming web-search path. The non-streaming Chat Completions path
with memory and skills tools was not part of that fix.

### Where

`src/lib/skills/interception.ts`, in `handleToolCallExecution`.

The `anthropic` branch already handles this case at lines 400-443. Once every
tool call has been handled server-side, it turns the results into text blocks
and normalises the now-stale `stop_reason: "tool_use"` to `"end_turn"`, so
clients stop waiting for blocks that no longer exist. Its comment says exactly
that.

The `openai` branch has no equivalent, at lines 391-397:

```ts
return {
  ...response,
  tool_results: results.map((r) => ({
    tool_call_id: r.id,
    output: JSON.stringify(r.result),
  })),
};
```

`choices[0].message.content` stays `null` and `finish_reason` stays
`"tool_calls"`. `tool_results` is not in the Chat Completions schema, so a
client that follows the spec never looks at it.

`git log -L 391,397` shows this branch untouched since `e6e54822f` ("add
Memory & Skill Injection from Proxy"), while the anthropic branch was revised
afterwards in `fd26e601a`. #12031 did touch this switch case later, but only
the Responses-API sub-branch above it (`getResponsesOutputContainer`), not the
Chat Completions fallback.

### Steps to Reproduce

1. Enable Memory under Settings, Memory & Skills.
2. Send a non-streaming Chat Completions request with no `tools` array,
   carrying a prompt long enough that the model reaches for memory. Roughly
   12KB of code plus a review instruction did it here.
3. Repeat it. The trigger is probabilistic: 1 of 3 identical concurrent calls
   on one run, 2 of 3 on an earlier one.
4. Read the whole response object rather than
   `choices[0].message.content`, which is where the evidence hides.

The request body carries no tools:

```python
body = {"model": "onmi-gemini3.6", "max_tokens": 8192, "stream": False,
        "messages": [{"role": "user", "content": PROMPT}]}
```

Empty responses come back with `finish_reason=tool_calls` and 16-19 completion
tokens. Healthy ones come back with `finish_reason=stop` and 39-171.

### Expected

Either of these would be fine:

Complete the round trip. Feed `tool_results` back to the model and let it
produce a final answer, which is what a client would do with its own tools.
OmniRoute already holds everything it needs to do this.

Or match the anthropic branch. When every tool call was handled server-side,
surface the results as assistant content and normalise `finish_reason` to
`"stop"`, so the reply is at least readable and terminal.

Returning a half-finished turn is the one option that leaves a compliant
client with nothing to show.

### Impact

Any Chat Completions consumer unaware of the `tool_results` extension sees an
empty assistant message. A consumer that retries on empty content retries
into the same condition, because the trigger recurs.

Concretely: code-review-forge reads `finish_reason` and recognises only
`"length"` as a terminal non-stop state, so `"tool_calls"` falls through to
its empty-response retry path and a review never converges.

### Ruled out

Three other explanations were measured and eliminated before landing here, in
case they come up in triage.

Compression is not involved. A four-arm matrix on the same prompt gave 2 of 3
usable concurrently with compression on and 2 of 3 with
`x-omniroute-compression: off`; serial runs gave 3 of 3 either way. The header
does change the request, taking prompt_tokens from 5936 to 5710, and recovers
nothing.

Account exhaustion is not involved. All 9 agy connections in the combo are
active with no rate limit, and `usage_history` shows recent traffic spread
across three distinct accounts.

Session stickiness pinning one connection is not involved, on the same
evidence.

Serial calls only looked healthier because each call is an independent sample.
The trigger is per-request rather than per-batch.

### Workaround

Turn Memory off for keys used by tool-unaware clients. Skills likely needs the
same treatment: its UI copy ("Tools for models", "Allows agents to trigger
functions") describes the same injection mechanism, though this report only
measured the memory path.

HouMinXi:
## Bug Description

Fixes #12696

A non-streaming Chat Completions request that declared no tools can still come back with `content: null` and `finish_reason: "tool_calls"`. OmniRoute injected `memory_search` (or a skills tool), the model called it, OmniRoute executed it, and then returned that turn to the caller. `tool_results` is sitting on the envelope; the second round trip that would let the model use it never happens. The client is left holding a tool it cannot service.

Measured on `agy/gemini-3.8-flash-high` through a combo, with Memory enabled. Same shape as #9725, which covered the streaming web-search path. The non-streaming Chat Completions path with memory/skills was not part of that fix.

## Root Cause

`handleToolCallExecution` already has an Anthropic branch that, once every tool call is handled server-side, turns the results into text and normalises `stop_reason: "tool_use"` to `"end_turn"`. The OpenAI Chat Completions branch only attached `tool_results` and left `content` null / `finish_reason` as `tool_calls`.

A follow-up round trip also needs one owner for the first provider send. Splitting recovery with `if (stream)` around the existing chatCore send double-dispatched the first fetch (`fetchCalls.length === 2`).

## Fix

- Server-owned tool loop (feature flag `SERVER_OWNED_TOOL_LOOP_ENABLED`): classify, execute, fence, and continue until the model returns a client-usable response or the loop hits a typed stop.
- `providerExecutionPipeline` owns the first send for both streaming and non-streaming. Passthrough lock: `fetchCalls.length === 1`.
- Non-streaming first send lives in `nonStreamingProviderLeg`. Empty-content fallback that cannot parse is an error, not a 200. Semaphore capacity maps to HTTP 429. Reasoning replay gets `historyMessages` from the translated body.
- Retry-After is read from the upstream Response, not from BaseExecutor outbound request headers.

## How to Verify

1. Enable Memory under Settings, Memory & Skills.
2. Send a non-streaming Chat Completions request with no `tools` array and a prompt long enough that the model reaches for memory (~12KB of code plus a review instruction did it here).
3. The response should have usable `choices[0].message.content` and `finish_reason` other than `tool_calls` for a fully server-owned turn.
4. `tests/integration/chat-pipeline.test.ts` still asserts a single fetch on the first send.

## Test Plan

- [x] Added regression tests for the empty-content reply, the single-fetch lock, semaphore 429, empty-content parse failure, and reasoning-cache `historyMessages`
- [x] Existing focused tests still pass (`tests/unit/non-streaming-provider-leg.test.ts`, `tests/unit/chatcore-stream-error-result.test.ts`, `tests/integration/chat-pipeline.test.ts`, `tests/unit/skills-interception*.test.ts`)
- [ ] Manual verification on a live Memory-enabled combo (same shape as the issue)

## Risk Assessment

Medium. The change re-routes the non-streaming first send and adds a follow-up loop behind a feature flag. Streaming still uses the existing chatCore path. Passthrough clients that already consume `tool_results` keep working through the escape-hatch formatter when the loop cannot continue.

## Not in this PR

- Streaming Memory/skills follow-up (that is #9725).
- Redeem/consume UI, combo extract, quota-weighted routing.

## ⚠️ base-red inherited

Fast Quality Gates on `release/v3.8.51` currently fail on files this branch does not touch:

- `changelog.d/fixes/reset-aware-model-family.md` missing a leading `- ` (merge-integrity)
- `tests/unit/reset-aware-request-scope-12600.test.ts` missing from `stryker.conf.json` (mutation-coverage). That file is not in `git diff --name-only $TIP...HEAD`.
- `open-sse/executors/glm.ts` extra 16th positional (`TS2554`)  -  tracked as #12770

Do not bake those into this PR.


HouMinXi:

- `6c94e3876a5` → `a1d2d20daab`
- range-diff identity (eq=27); all commits GPG `G`
- `27` unique commit(s) ahead, 0 behind
- `--force-with-lease` to `HouMinXi:fix/server-owned-tool-loop`

HouMinXi:
Looked at merge commit `d6f315018` on the wiring files:

- `open-sse/handlers/chatCore.ts`
- `open-sse/handlers/chatCore/failureUsage.ts`
- `open-sse/handlers/chatCore/nonStreamingFinalization.ts`
- `open-sse/handlers/chatCore/serverOwnedToolLoopGate.ts`
- `open-sse/handlers/chatCore/serverOwnedToolLoopWire.ts`
- `open-sse/utils/requestLogger.ts`
- `src/lib/skills/interception.ts`
- `src/lib/skills/toolExecutionFence.ts`

No code change from that pass. Notes that came up, checked against the tree:

- `pipelineRecovered = true` at 4931 is after the non-streaming error return at 4928. Stream 3726 is the skip-legacy-fallback contract; the pipeline already did rotation and family fallback.
- `ChatCoreErrorResult.status` is `number` (`src/lib/skills/toolLoopTypes.ts`). `upstream_${err.status}` is `upstream_404`, not a string status.
- Fence / identity failures still log `translatedBody` as the request.
- Stream interrupt is `streamErrorCode === "stream_interrupted" || "aborted"` at 5730, not the unused `interrupted` boolean.
- Antigravity BYOP 422 rotation lives in `providerExecutionPipeline.ts` (clone body, `gcp_project_required`, cooldown, rotate). The inner `while (attempts < maxAttempts)` is one wire send.
- Non-2xx streaming returns the original `res.response` and drops the semaphore without `body.cancel()`, so the pipeline can still `clone().text()`.

Lock: `tests/integration/chat-pipeline.test.ts` (`fetchCalls.length === 1`). `SERVER_OWNED_TOOL_LOOP_ENABLED` still defaults false, so this SHA is not live until the flag is on.

Issue #8618, opened 2026-07-25
fix(tests): 45 test files run in no blocking job — orphaned .test.tsx files and dead node:test import under Vitest


MumuTW:
## Summary

`vitest.config.ts` includes `tests/unit/**/*.test.tsx`, `open-sse/**/__tests__/**/*.test.ts`
and `src/lib/{memory,skills}/__tests__/**/*.test.ts`. The only script that uses that config
is:

```json
"test:vitest:ui": "vitest run --config vitest.config.ts tests/unit/ui"
```

The trailing `tests/unit/ui` is a **positional path filter**. Everything the config includes
from outside that directory is silently dropped. The Node runner cannot pick them up either —
`test:unit` and `test:unit:ci:shard` glob `*.test.ts` and `*.test.mjs` only, never `.tsx` — and
`vitest.mcp.config.ts` (the other blocking job) includes only co-located
`src/**/__tests__/**` and a fixed list under `open-sse/`.

**45 test files are therefore executed by no blocking CI job.**

| files | location | why they are skipped |
| ---: | --- | --- |
| 31 | `tests/unit/**/*.test.tsx` outside `tests/unit/ui/` | dropped by the positional filter |
| 9 | `open-sse/**/__tests__/*.test.ts` (excluding the 20 in `vitest.mcp.config.ts`) | in no runner's glob |
| 5 | `src/lib/{memory,skills}/__tests__/*.test.ts` | in no runner's glob |

Among the 9 backend orphans: `chatgptTlsClient`, `claudeTlsClient`, `grokTlsClient`,
`tierResolver`, `volumeDetector`, `specificityDetector`, `manifestAdapter`,
`schemaCoercion`, `maxTokensHelper`.

## Two files prove the cost

**1. A test guarding a component that was deleted six weeks ago.**

`tests/unit/shared/components/AutoRoutingBanner.test.tsx` imports
`@/shared/components/AutoRoutingBanner`. That component does not exist — it was removed in
**#6164**. The replacement guard already exists and is green:

```
tests/unit/home-no-autorouting-banner.test.ts
  // This guard replaces the deleted AutoRoutingBanner.test.tsx: the component was …
  ok 1 - home page does not render the removed AutoRoutingBanner (#6164)
  ok 2 - AutoRoutingBanner component stays deleted (#6164)
```

The `.test.tsx` was supposed to be deleted along with the component and was not. Nothing
noticed, because nothing runs it.

[...]

PR #12868, opened 2026-09-06
fix(dashboard): pass nodeMap into Runtime QuotaGroup

HouMinXi:
## Bug

`/dashboard/runtime` throws `ReferenceError: nodeMap is not defined` when a quota monitor is in `error`, `exhausted`, or `alerting`.

Empty monitor lists never reach that path. `quotaMonitorEnabled` defaults to false, so the page stays idle unless monitors are turned on.

No upstream issue.

## Cause

`QuotaGroup` is a module-level function in `RuntimePageClient.tsx`. It reads `nodeMap`, which only exists inside `RuntimePageClient` (from `useProviderNodeMap()`). That identifier is not in scope at the module level.

Landed in a7ae9550bd (v3.8.36, #4854). The existing unit file only fed empty monitors, and vitest excluded it (#8618).

The Chinese string "内部服务器错误" on this page is the per-page RSC `publicSystem.error` catalog. It is not this crash.

## Fix

Pass `nodeMap` as a prop into `QuotaGroup` at the exhausted, alerting, and error call sites.

`useProviderNodeMap()` already returns `Map` (never undefined). `resolveProviderName` already uses `nodeMap?.get`.

## Tests

- Empty monitors still pass.
- Fixture with `status: "error"` and `status: "exhausted"` mounts `QuotaGroup`. Throws `ReferenceError` before the prop; green after.
- Dropping the parameter from the function signature throws the same `ReferenceError` again.
- Dropping only the JSX prop stays green (`undefined` map is a valid fallback).
- Removed `tests/unit/ui/runtime-page-client.test.tsx` from the vitest exclude list.

`warning` → `alerting` remap is not in this PR.

## Risk

Small. Empty-monitor path is unchanged. The new prop is the same Map the parent already held.

File-size cap for this file: 1201 → 1222 (prettier wrap on the three call sites).

### DIFF
--- a/changelog.d/fixes/12868-runtime-quotagroup-nodemap.md
+++ b/changelog.d/fixes/12868-runtime-quotagroup-nodemap.md
@@ -1,0 +1,1 @@
+- **fix(dashboard):** pass `nodeMap` into Runtime `QuotaGroup` so a quota monitor in error/exhausted/alerting no longer throws `ReferenceError: nodeMap is not defined`. ([#12868](https://github.com/diegosouzapw/OmniRoute/pull/12868))
--- a/changelog.d/fixes/runtime-quotagroup-nodemap.md
+++ b/changelog.d/fixes/runtime-quotagroup-nodemap.md
@@ -1,0 +1,1 @@
+- **fix(dashboard):** pass `nodeMap` into Runtime `QuotaGroup` so a quota monitor in error/exhausted/alerting no longer throws `ReferenceError: nodeMap is not defined`.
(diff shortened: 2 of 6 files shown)


HouMinXi:
Looked at `ceca5c585` on `RuntimePageClient.tsx`.

`QuotaGroup` is a module-level function. `nodeMap` only exists inside `RuntimePageClient`, so the identifier was out of scope. Passing the map as a prop is the right fix.

One review note asked whether the hook can return `undefined`. It cannot: `useProviderNodeMap()` is typed as `Map` and starts as `new Map()`. `resolveProviderName` already uses `nodeMap?.get`. Dropping only the JSX prop stays green; dropping the parameter from the function signature throws `ReferenceError: nodeMap is not defined` again.

No further code change from that note.

Lock: `tests/unit/ui/runtime-page-client.test.tsx` — empty monitors still pass; `status: "error"` / `"exhausted"` mounts `QuotaGroup`.

HouMinXi:

- `142e418a14e` → `270fe362810`
- `--force-with-lease` to `HouMinXi:fix/runtime-quotagroup-nodemap`

Issue #9617, opened 2026-08-06
fix(providers): Gemini provider returns 400 Unknown name uniqueItems when client sends many tools


MintAC:
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Linux (WSL2)

### Node.js Version

24.18.0

### Provider(s) Involved

Gemini (Google AI Studio, generativelanguage.googleapis.com)

### Client Tool

Qwen Code (OpenAI-compatible)

### Description

Every request routed to the Gemini provider fails with HTTP 400 whenever the client sends a large tool set (25 tools). Requests **without** tools, or with only 1 tool, succeed consistently.

## Error
[400]: Invalid JSON payload received. Unknown name "uniqueItems" at 'tools[0].function_declarations[64].parameters.properties[0].value.items.properties[3].value': Cannot find field.
Invalid JSON payload received. Unknown name "uniqueItems" at 'tools[0].function_declarations[66].parameters.properties[2].value': Cannot find field.

## Key observations
1. **The original OpenAI-format request body does NOT contain `uniqueItems` anywhere.** I scanned the stored request artifact of a failing call (`call_logs/2026-08-06/*.json`) and found zero occurrences of the keyword. The `uniqueItems` field is being injected by OmniRoute while converting the OpenAI tools schema into Gemini `function_declarations`.
2. The failing path is a deeply nested parameter (`properties[0].value.items.properties[3].value`) — an array-of-objects parameter whose item schema carries `uniqueItems`.
3. Also notable: the input request only contains **25 tools**, yet the error references `function_declarations[64]` and `[66]` — the declaration count in the outgoing Gemini payload appears inflated beyond the input tool count (suspect array flattening / expansion in the converter).
4. Requests with no tools (e.g. the dashboard "Test Connection" math prompt) succeed; a 1-tool memory-retrieval request also succeeds. So the bug scales with tool count.

## Impact
Gemini provider is unusable for tool-using agents. Each failing call burns 6–27 s before the 400 returns, and in auto mode it cascades through the whole fallback chain, adding significant latency to every request.

Attached: sanitized request body (25 tools, messages removed) 
[gemini-uniqueItems-failure-request.json](https://github.com/user-attachments/files/30793243/gemini-uniqueItems-failure-request.json)

### Steps to Reproduce

1. Configure a Gemini (Google AI Studio) provider with a free-tier key.
[...]

PR #12872, opened 2026-09-06
fix(gemini): strip prefixItems/additionalItems from tool schemas (#12871)

### LINKED ISSUE #12871: fix(providers): Gemini 400: "Unknown name \"prefixItems\"" — missing from GEMINI_UNSUPPORTED_SCHEMA_KEYS, breaks all Claude Code tool calls
**OmniRoute version:** 3.8.50 (Docker, `diegosouzapw/omniroute:latest`)
**OS:** macOS (arm64)
**Client:** Claude Code CLI → `/v1/messages` → `gemini/gemini-2.5-pro`

## Summary

`GEMINI_UNSUPPORTED_SCHEMA_KEYS` (`open-sse/translator/helpers/geminiHelper.ts`) does not include `prefixItems`, so tuple-typed array schemas survive `cleanJSONSchemaForAntigravity()` and are sent to Google verbatim. Gemini's schema parser has no such field and rejects the whole request with a hard 400.

This is the same class of bug already fixed for `multipleOf`, `uniqueItems` (#9617), `strict`, and `encrypted` — `prefixItems` is simply missing from the list.

Impact: **every tool-enabled Claude Code request to a Gemini model fails**, because Claude Code ships built-in tools whose schemas use `prefixItems`. It is not recoverable by config — any client sending a JSON Schema 2020-12 tuple hits it.

## Error

```
API Error: 400 Invalid JSON payload received. Unknown name "prefixItems" at
'tools[0].function_declarations[1].parameters.properties[5].value.properties[0].value.items':
Cannot find field.
```

## Reproduction

Any tool whose schema contains a `prefixItems` tuple. Minimal case:

```bash
curl -X POST http://localhost:20128/v1/messages \
  -H "x-api-key: $OMNIROUTE_KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -d '{
  "model": "gemini/gemini-2.5-pro",
  "max_tokens": 512,
  "messages": [{"role":"user","content":"Query the notes collection where status equals open."}],
  "tools": [{
    "name": "artifact_db",
    "description": "Query a collection.",
    "input_schema": {
      "type": "object",
      "properties": {
        "collection": {"type": "string"},
        "query": {
          "type": "object",
          "properties": {
            "where": {
              "type": "array",
              "items": {
                "type": "array",
                "prefixItems": [
                  {"type": "string"},
                  {"enum": ["eq","ne","lt","gt"], "type": "string"},
                  {}
                ]
              }
            }
          }
        }
      },
      "required": ["collection"]
    }
  }]
}'
```

**Expected:** the tool call succeeds.
**Actual:** upstream 400, `Unknown name "prefixItems"`.

## Cause

`open-sse/translator/helpers/geminiHelper.ts` — `prefixItems` is absent from the set, so `removeUnsupportedKeywords()` leaves it in place:

```ts
  "uniqueItems",
  // Complex schema keywords (handled by flattenAnyOfOneOf/mergeAllOf)
  "anyOf",
```

`additionalItems` (the JSON Schema draft-07 spelling of the same tuple concept) is missing for the same reason and will produce an identical 400.

## Fix

Add `prefixItems` and `additionalItems` to `GEMINI_UNSUPPORTED_SCHEMA_KEYS`. Stripping `prefixItems` leaves `{type: "array"}`, and the existing `ensureArrayItems()` phase (#10578) then supplies a safe `items` schema — so the parameter degrades to an untyped array rather than failing the request.

Verified locally against `gemini-2.5-pro`; the reproduction above then returns a correct tool call:

```json
{"type":"tool_use","name":"artifact_db",
 "input":{"query":{"where":[["status","==","open"]]},"collection":"notes"}}
```

PR follows.

sonirahul:
Fixes #12871.

## Problem

`GEMINI_UNSUPPORTED_SCHEMA_KEYS` did not include `prefixItems`, so tuple-typed array schemas survived `cleanJSONSchemaForAntigravity()` and were sent to Google verbatim. Gemini's function_declarations parser has no field for them and rejects the entire request:

```
400 Invalid JSON payload received. Unknown name "prefixItems" at
'tools[0].function_declarations[1].parameters.properties[5].value.properties[0].value.items':
Cannot find field.
```

Claude Code ships built-in tools whose schemas describe filter clauses as tuples, so **every tool-enabled request to a Gemini model failed**, with no config-level workaround.

`additionalItems` — the draft-07 spelling of the same concept — was missing for the same reason and produces an identical 400.

## Fix

Add both keywords to `GEMINI_UNSUPPORTED_SCHEMA_KEYS`. This is the same class of fix already applied for `uniqueItems` (#9617), `multipleOf`, `strict` and `encrypted`.

Stripping `prefixItems` leaves a bare `type: "array"`, which the existing `ensureArrayItems()` phase (#10578) backfills with a safe `items` schema — so the parameter degrades to an untyped array rather than failing the call. The third test pins that interaction so a future reorder of the cleanup phases can't silently reintroduce a bare array.

## Verification

Built a patched image and ran the reproduction from #12871 against `gemini-2.5-pro`. Before: upstream 400. After:

```json
{"type":"tool_use","name":"artifact_db",
 "input":{"query":{"where":[["status","==","open"]]},"collection":"notes"}}
```

Gemini fills the tuple correctly despite the erased positional types.

Unit tests, scoped to the 21 files importing `geminiHelper.ts` / `geminiToolsSanitizer.ts`:

| | tests | pass | fail |
|---|---|---|---|
| before | 83 | 76 | 7 |
| after | 86 | 79 | 7 |

+3 tests, +3 passing, no change in failures. The 7 pre-existing failures are `ERR_MODULE_NOT_FOUND` on extensionless imports in my local environment (Node 24), identical on a clean checkout of `main` and unrelated to this change.

### DIFF
--- a/changelog.d/fixes/12872-gemini-prefixitems.md
+++ b/changelog.d/fixes/12872-gemini-prefixitems.md
@@ -1,0 +1,1 @@
+- **fix(gemini):** strip `prefixItems`/`additionalItems` from tool parameter schemas — tuple-typed arrays triggered a hard upstream 400 (`Unknown name "prefixItems"`), breaking every tool-enabled Claude Code request to a Gemini model ([#12872](https://github.com/diegosouzapw/OmniRoute/pull/12872)) — thanks @sonirahul
--- a/open-sse/translator/helpers/geminiHelper.ts
+++ b/open-sse/translator/helpers/geminiHelper.ts
@@ -63,6 +63,16 @@ export const GEMINI_UNSUPPORTED_SCHEMA_KEYS = new Set([
   // it, rejecting the whole request with "Unknown name \"uniqueItems\"".
   // Upstream 9router already strips it alongside `contains` for the same error.
   "uniqueItems",
+  // #12871: tuple-typed array keywords. `prefixItems` (JSON Schema 2020-12) and its
+  // draft-07 spelling `additionalItems` describe positional array entries, which the
+  // Gemini schema parser has no field for — it rejects the request with
+  // "Unknown name \"prefixItems\" ... Cannot find field". Claude Code ships built-in
+  // tools whose schemas use tuples, so leaving these in breaks every tool-enabled
+  // request to a Gemini model. Stripping `prefixItems` leaves a bare `type: "array"`,
+  // which the `ensureArrayItems()` phase below (#10578) then fills with a safe `items`
+  // schema — the parameter degrades to an untyped array instead of failing the call.
+  "prefixItems",
+  "additionalItems",
   // Complex schema keywords (handled by flattenAnyOfOneOf/mergeAllOf)
   "anyOf",
   "oneOf",
--- a/tests/unit/12871-gemini-prefixitems.test.ts
+++ b/tests/unit/12871-gemini-prefixitems.test.ts
@@ -1,0 +1,121 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import { buildGeminiTools } from "../../open-sse/translator/helpers/geminiToolsSanitizer.ts";
+
+// Issue #12871: Gemini rejects `prefixItems` in function_declarations parameter schemas
+// with HTTP 400 "Unknown name \"prefixItems\" ... Cannot find field" (Gemini's protobuf-JSON
+// schema parser only accepts a subset of JSON Schema/OpenAPI 3.0 — the same class of error
+// already fixed for `uniqueItems` (#9617), `multipleOf`, `strict` and `encrypted` in
+// GEMINI_UNSUPPORTED_SCHEMA_KEYS, open-sse/translator/helpers/geminiHelper.ts).
+//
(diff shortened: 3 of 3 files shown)


diegosouzapw:
Thanks for this — while triaging it I found half of the original bug already
shipped: PR #12540 (merged 2026-09-11, via issue #12509) added `prefixItems` to
`GEMINI_UNSUPPORTED_SCHEMA_KEYS` already. `additionalItems` (the draft-07 spelling
of the same tuple concept) is still missing from the tip, though — that part of
your fix is still needed and still valid.

Could you rebase onto the current `release/v3.8.51` tip and drop the now-redundant
`"prefixItems"` line (keeping just `"additionalItems"`), and adjust the changelog
fragment so it doesn't duplicate #12540's entry? The `additionalItems` regression
test and the `ensureArrayItems()` interaction test are still valuable coverage we
don't have today.

Issue #12242, opened 2026-09-01
fix(resilience): per-model 402 on a passthrough gateway terminalizes the whole connection (free models knocked out, never recovers) — 402 variant of #3027


brick30llc-ctrl:
### Summary

A **402** from a single paid model on a passthrough/gateway provider sets a **connection-wide, terminal `credits_exhausted`** status. This takes out every *free* model on the same key, and because `credits_exhausted` is a terminal status it is **never auto-recovered** — the connection stays dead until an operator manually resets it.

This is the same blast-radius class as #3027 (`ollama-cloud` per-model subscription **403**), but via the **402** path, which the #3027 fix does not cover.

Observed in production on `kilo-gateway` and `ollama-cloud`.

### Steps to Reproduce

1. Connect `kilo-gateway` (or `ollama-cloud`) with a free-tier key. Confirm free models work, e.g.
   `kilo-gateway/cohere/north-mini-code:free`, `ollama-cloud/gpt-oss:20b`.
2. Send **one** request to a paid model on the same connection, e.g.
   `kilo-gateway/openai/gpt-chat-latest` or `kilo-gateway/nex-agi/nex-n2-mini`:
   ```text
   [402]: Add credits to continue, or switch to a free model
   ```
3. Immediately retry any **free** model on that same connection:
   ```text
   [kilo-gateway] All 1 connection(s) credits exhausted — please reconnect in the dashboard (HTTP 401)
   ```
4. `provider_connections` now shows `test_status='credits_exhausted'`, `error_code='402.0'`,
   `last_error_type='quota_exhausted'`. It stays that way indefinitely — restarting the service does not clear it.

### Root Cause (file:line)

**1. Every 402 is terminalized connection-wide, before any model-scoped path can run.**

`src/sse/services/auth.ts:364` — `resolveTerminalConnectionStatus()`:
```ts
if (result.creditsExhausted || status === 402) return "credits_exhausted";
```
There is no `hasPerModelQuota(...)` / passthrough check here, so a per-model billing error is treated identically to a dead API key.

**2. The generic per-model lockout gate excludes 402.**

`src/sse/services/auth.ts:2010-2015`:
```ts
if (
  isPerModelQuotaProvider &&
  provider &&
  provider !== "codex" &&
  model &&
  (status === 404 || status === 429 || status >= 500)
)
```
402 is not in the set (same shape as the pre-#3027 gap for 403).

[...]

PR #12875, opened 2026-09-06
fix(credentialHealth): do not poison multi-upstream openai-compat conn on one model 402

RaviTharuma:
## Summary
On openai-compatible multi-upstream gateways, a single upstream 402 must not mark the whole connection credits_exhausted.

Per-model/per-upstream CredentialHealth for openai-compatible / hasPerModelQuota paths.

Related (not duplicates): #12242/#12266; #12532/#12533.

No image/tag bump from opener.


RaviTharuma:
Tracking issue: #12877

RaviTharuma:
Extracted the existing finish-reason parser without changing its behavior in a80b2213f7608fe198011a136af03ae1d01b0c49 to fix the chatCore file-size regression. The PR-parent size check failed before this change (5997 lines against the frozen 5984 limit) and passes afterward without changing the baseline. The six existing per-upstream 402 health tests and three parser regression tests pass; focused ESLint and diff whitespace checks pass. Please re-review this bounded correction. This is not a claim that the PR is merge-ready: remote API Route Typecheck, Docs Gates and Merge integrity are still failing, and other checks are pending.

PR #12876, opened 2026-09-06
fix(monitoring): expose failed connection ids on /api/monitoring/health

RaviTharuma:
## Summary
Expose failed connection ids on GET /api/monitoring/health; align gauge/cache vs sqlite test_status.

Related: #12532/#12533 (merged nonblocking path).

No image/tag bump.


RaviTharuma:
Tracking issue: #12878

PR #11702, opened 2026-08-26
fix(quality): ratchet-bank verifier rejects object-valued notes — nightly banking has been aborting since 2026-08-11

MumuTW:
## What this fixes

`scripts/quality/verify-ratchet-bank.mjs` — the safety gate that lets the automated
banking lane write to the quality baselines — **exits 1 on a pristine
`release/v3.8.51` working tree**, with no local changes at all:

```
$ git status --porcelain          # empty
$ node scripts/quality/verify-ratchet-bank.mjs; echo $?
Nothing to bank — every baseline already matches the code.
[verify-ratchet-bank] UNSAFE — 1 problem(s); refusing to bank:
  ✗ frozen: note "_rebaseline_2026_08_11_v3850_merge_storm_provider_registry: …" was rewritten
1
```

### Root cause

`verifyFrozenMap` compares note values referentially:

```js
else if (after[key] !== prev)
  problems.push(`${label}: note "${key}" was rewritten — …`);
```

That is correct for the string notes the function and its tests were written
against. It is wrong for the **object-valued** note added on 2026-08-11, where the
merge-storm rebaseline folded four `file → LOC` pairs into the value of a single
note key. Two structurally identical objects parsed from two `JSON.parse` calls are
never `===`, so the note is reported "rewritten" unconditionally.

### Why it matters

`bank-ratchet-shrinks` in `.github/workflows/nightly-release-green.yml:398` runs this
verifier and aborts on any non-zero exit. So the automation added in #8612 —
specifically to stop caps outliving the shrinks that earned them (#8584) — **has
banked nothing since 2026-08-11.**

The measured consequence, against `release/v3.8.51`:

| baseline | entries | unbanked slack |
| --- | ---: | ---: |
| `frozen` | 90 | **26,907 lines** |
| `testFrozen` | 37 | **14,657 lines** |

That slack is not academic. A "shrink-only" ceiling set above the file's real size is
a growth allowance:

| file | frozen at | actually | headroom |
| --- | ---: | ---: | ---: |
| `dashboard/combos/page.tsx` | 6,114 | 4,803 | +1,311 |
| `open-sse/executors/chatgpt-web.ts` | 4,213 | 3,143 | +1,070 |
| `open-sse/services/adobeFireflyClient.ts` | 3,899 | 2,953 | +946 |
| `dashboard/api-manager/ApiManagerPageClient.tsx` | 4,052 | 3,128 | +924 |
[...]

PR #11685, opened 2026-08-26
feat(oauth): default Antigravity-family connections to model auto-sync

MumuTW:
## Summary

The Antigravity backend (Google Cloud Code) ships new models faster than the pinned catalog can be re-frozen by hand. Concrete case: Gemini 3.7 Flash tiers (gemini-3.7-flash-high/medium/low, upstream gemini-3.7-flash-tiered) were live on the backend — visible via `agy models` and `:fetchAvailableModels` — while AGY_PUBLIC_MODELS / ANTIGRAVITY_PUBLIC_MODELS still stopped at 3.6. Result: /v1/models was blind to them and bare names had no alias to resolve through (same recurring pattern as #2979 / #7678).

Live discovery (#8123) and the 24h model auto-sync scheduler (#488) already exist end-to-end — the missing piece is the opt-in trigger. This PR defaults **new** Antigravity-family connections to providerSpecificData.autoSync: true so discovery lands in the synced catalog and /v1/models stays current without code changes:

- mapAntigravityTokens — covers both antigravity (IDE profile) and agy (CLI profile) OAuth connects
- createConnectionFromAgyToken — covers Antigravity CLI token-file imports; on re-import the default sits **before** the existing-data spread so a previously persisted operator choice (true or false) always wins

Existing connections are untouched; the per-connection dashboard toggle remains the opt-out.

## Related Issues

- Related to #2979 (accepted background discovery/sync; this ships the opt-in trigger for the Antigravity family)
- Related to #488 (model auto-sync scheduler), #8123 (live AGY discovery), #7678 (same staleness failure mode on zai-web)

## Validation

- [x] Change type: provider
- [x] Focused tests from the golden path: node --import tsx/esm --test tests/unit/antigravity-family-auto-sync-default.test.ts → 5/5 pass
- [x] Regression: tests/unit/agy-auth-import.test.ts + tests/unit/antigravity-model-aliases.test.ts → 15/15 pass
- [x] npx eslint clean on all touched files
- [x] Reconciled with current main (v3.8.50)
- [x] Production-code changes include new automated tests in this PR

## Tests Added Or Updated

- tests/unit/antigravity-family-auto-sync-default.test.ts (new):
  - mapTokens defaults autoSync: true for both agy (cli) and antigravity (ide) profiles
  - new CLI token import persists autoSync: true (real temp-DATA_DIR sqlite)
  - re-import preserves an explicit autoSync: false
  - re-import defaults a legacy row that never had the flag

## Coverage Notes

[...]

PR #12892, opened 2026-09-07
fix(sse): session affinity misses Gemini and legacy shapes, and hashes whole payloads

MumuTW:
`extractSessionAffinityKey` falls back to hashing the request's first input text when no session id is present. Three problems in that fallback, all on the path every sessionless request takes.

### Unbounded stringify on a hot path

`extractTextForSessionHash` ended both its array branch and its object branch with `JSON.stringify(value)`:

```ts
return parts.length > 0 ? parts.join("\n") : JSON.stringify(value);
...
if (value && typeof value === "object") return JSON.stringify(value);
```

A multimodal message whose parts are one text and one base64 image stringifies the image as well. A body with no recognisable text field stringifies the whole body. Only the first 4096 characters are ever hashed, so the remainder is built and discarded. Extraction now accumulates against `SESSION_HASH_TEXT_LIMIT` and stops at the boundary.

### Colliding affinity keys

Because no-text payloads were stringified rather than rejected, `{input: []}` and `{input: {}}` produced stable hashes — every request of that shape pinned to the same connection. Unrecognised shapes now return `null`: no affinity, which is the honest answer for a body carrying no user text.

### Shapes that never got affinity

The fallback read `input` and `messages` only. These returned `null` and were never pinned:

| shape | example |
|---|---|
| Google Gemini | `{contents: [{role: "user", parts: [{text}]}]}` |
| legacy completions | `{prompt}` — OpenAI legacy, Anthropic `/v1/complete`, Ollama |
| root-level text | `{query}`, `{instruction}` |

It reads those now, and reads `text` / `content` / `prompt` / `parts` off object values rather than stringifying them.

## Tests

`tests/unit/session-affinity-edge-cases.test.ts` — 36 cases over array and object bodies, multimodal payloads, the 4096 truncation boundary, Gemini's format, and the empty-container collisions.

- **13 of the 36 fail on `main`**, pass with this change
- the 34 cases in the existing `session-affinity-generic-7274`, `combo-sessionless-pin-3825`, `codex-failover`, `session-affinity-combo-timeout-eviction` and `codex-session-affinity-reset-aware-5903` suites still pass
- `eslint` with the repo's suppressions is clean on `auth.ts`; `typecheck:core` reports nothing in `auth.ts`

Follows #11746, which hardened the key extraction itself; this is the text-extraction layer underneath it.

https://claude.ai/code/session_015E2idg9mo8qE2M9F5Myvnt

### DIFF
--- a/changelog.d/fixes/12892-session-affinity-text-extraction.md
+++ b/changelog.d/fixes/12892-session-affinity-text-extraction.md
@@ -1,0 +1,5 @@
+Session affinity now recognises Google Gemini `contents`, the legacy completions
+`prompt`, and root-level `query` / `instruction` bodies, which previously got no
+affinity key at all. The text used for the fallback hash is bounded at 4096
+characters instead of stringifying whole multimodal payloads, and a body with no
+recognisable text returns no key rather than a hash that every such body shares.
--- a/src/sse/services/auth.ts
+++ b/src/sse/services/auth.ts
@@ -121,6 +121,7 @@ import {
 import { isFreeModel } from "@/shared/utils/freeModels";
 import {
   applySessionAffinityPin,
+  extractSessionAffinityInputText,
   formatSessionKeyForLog,
   resolveForcedConnectionForCredentialPool,
   resolveSessionAffinityTtlMs,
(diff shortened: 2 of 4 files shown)


MumuTW:
CI triage after the retrigger — none of the 4 failing jobs appear related to this PR:

- **Build**: the runner received a shutdown signal ~7 min in (`The runner has received a shutdown signal` → `The operation was canceled`), the same cancellation as the previous run. Looks like infra instability rather than a compile error (the earlier run compiled successfully before being killed).
- **Lint**: fails on `npm audit` with 19 vulnerabilities (incl. the critical Next.js advisories). Upstream main's latest CI run fails this same job identically — baseline, not from this diff.
- **Quality Gates (Extended)**: also fails identically on upstream main's latest CI — baseline.
- **Unit Tests (2/8)**: I re-ran the full shard locally on this branch (`TEST_SHARD=2/8 npm run test:unit:ci:shard`): 4343 tests, 0 failures. Likely a resource/timeout flake on the runner.

Pushed an empty commit to retrigger once more; if Build keeps getting killed, a maintainer re-run would be appreciated. The Lint/Quality Gates failures track upstream main and should be ignorable for this PR.

MumuTW:
Hi @diegosouzapw,

Quick note on the "fork rejects maintainer pushes" you hit on #11685 / #11702 / #11749: those three were opened with "Allow edits by maintainers" unchecked on my side — sorry for the extra work. It is enabled on #12892 and #13408, and I'll keep it on for everything going forward, so a rebase-and-merge from your end should just work.

While I'm here: would you consider giving me **triage** access on the repo? Since July I've landed ~37 PRs here (task-aware routing / auto-combo fitness, request-body heap fixes, the ratchet shrink-banking lane, the larger file decompositions) and filed ~36 issues. Most of what I file comes with a repro, and I end up doing the follow-up on stale ones anyway. With triage I could label and link my own issues to their PRs, close obsolete reports, and mark superseded PRs, without pulling you in for housekeeping. I'm not asking for write, and happy to keep going through PR review as now.

Thanks for the fast turnaround on the August batch.

diegosouzapw:
Thanks for this — but it looks like your own earlier PR #11746 ("harden session affinity key
extraction", merged into release/v3.8.51 on 2026-08-28) already ships this exact fix: no more
unbounded `JSON.stringify`, Gemini `contents`/`parts` support, bounded text extraction, and the
same exported helper shape in `sessionAffinityPin.ts`. This PR was opened 10 days later from an
older `main` fork that predates #11746. I ran your own new test file
(`session-affinity-edge-cases.test.ts`, all 36 cases) unmodified against the current release
tip: 34/36 already pass. The 2 remaining are just a design difference #11746 itself settled on
(reject an oversized session_id outright vs. cap-and-hash it) — not a real gap. Recommending we
close this one as superseded by #11746 rather than merge; let me know if you'd rather revisit
the cap-vs-reject choice for huge explicit session IDs as a small separate PR against the
current tip.

MumuTW:
Agreed — keeping this closed as superseded by #11746. I am not opening a follow-up for the cap-vs-reject difference. That choice is already settled on the release tip, and I do not have a case where rejecting an oversized explicit session id is the wrong behavior.

PR #12894, opened 2026-09-07
fix(api): match the call-log API-key filter on the key id, not the name alone

### LINKED ISSUE #12873: fix(backend): Dashboard logs API-key filter returns 0 rows when selecting a key from dropdown (UUID passed, name matched)
# [BUG] Dashboard logs API-key filter returns 0 rows when selecting a key from dropdown (UUID passed, name matched)

## Environment
- **OmniRoute version:** 3.8.50 (Docker, `diegosouzapw/omniroute:latest`)
- **Deployment:** Docker Compose, port 20128
- **OS:** Ubuntu 24.04

## Summary
On the Dashboard → Logs page, selecting an API key from the "API Key" dropdown filter results in **0 rows shown**, even though the selected key has thousands of call logs. Filtering by the key's **name** (typed manually) works fine; selecting the same key from the **dropdown** returns nothing.

## Root cause
The dashboard's `RequestLoggerV2.tsx` builds the dropdown `value` from **`l.apiKeyId || l.apiKeyName`** (line ~865, ~1020), so it sends the API key's **UUID** (`apiKeyId`) to the backend:

```
params.set("apiKey", selectedApiKey);   // selectedApiKey = apiKeyId (UUID)
```

The backend `/api/usage/call-logs` (`route.ts`) does two filtering passes:

1. **SQL layer** (`getCallLogs`, `callLogs.ts:99-101`) — correct:
   ```sql
   (cl.api_key_name LIKE @apiKeyQ OR cl.api_key_id LIKE @apiKeyQ)
   ```
2. **In-memory layer** (`rowMatchesFilter`, `call-logs/route.ts:60`) — **buggy**:
   ```js
   if (filter.apiKey && !matchesSearch(row?.apiKeyName || "", String(filter.apiKey))) {
     return false;
   }
   ```

The in-memory pass only matches `row.apiKeyName` against the filter value. When the frontend sends the key's **UUID**, the SQL layer correctly returns the rows, but then `rowMatchesFilter` compares the UUID against `apiKeyName` → never matches → **all rows filtered out → 0 rows displayed**.

## Reproduce
1. Open OmniRoute Dashboard → Logs.
2. Note the "API Key" dropdown lists keys (using `apiKeyId` as option value, e.g. `01ab6f86-3789-403a-9cf4-2f3f68551db9`).
3. Select any key that has traffic.
4. The log table goes empty (0 rows), even though that key has thousands of calls.

Via API directly:
```
GET /api/usage/call-logs?apiKey=01ab6f86-3789-403a-9cf4-2f3f68551db9  → 0 rows
GET /api/usage/call-logs?apiKey=arc%20cc                              → 10 rows (works)
```

## Expected behavior
Selecting an API key from the dropdown should filter logs by that key, matching on **either** `apiKeyId` **or** `apiKeyName` — consistent with the SQL layer.

## Suggested fix
In `rowMatchesFilter` (`call-logs/route.ts:60`), also match `apiKeyId`:
```js
if (filter.apiKey && !matchesSearch(row?.apiKeyName || "", String(filter.apiKey)) && !matchesSearch(row?.apiKeyId || "", String(filter.apiKey))) {
  return false;
}
```

Or alternatively, have the frontend send the API key **name** instead of the UUID when both are available.

## Notes
- This is a **regression introduced by #11082** (which added `rowMatchesFilter` for in-memory rows). The SQL layer was already correct.
- Related but different: #2225 (x-api-key not parsed → null attribution) is a separate issue.

ntdat812:
Fixes #12873. Your trace is exact — the filter is applied twice and the second pass does not know what the first one matched on.

## Two passes, one of them narrower

`GET /api/usage/call-logs` filters the same `apiKey` value in two places:

| pass | predicate | matches |
|---|---|---|
| SQL, `getCallLogs()` (`src/lib/usage/callLogs.ts:722`) | `cl.api_key_name LIKE @apiKeyQ OR cl.api_key_id LIKE @apiKeyQ` | name **or** id |
| in-memory, `rowMatchesFilter()` (`route.ts:60`) | `matchesSearch(row.apiKeyName, filter.apiKey)` | name only |

The second pass exists for a good reason: `buildCallLogListRows()` merges active and recently-completed in-memory entries into the persisted rows, and those would otherwise bypass every filter. It is documented as "idempotent for DB rows (they already satisfy the predicate)" — and that is the part that stopped being true for this one filter. The dropdown sends `apiKeyId || apiKeyName`, so for any key that has an id the SQL layer returns the right rows and the in-memory pass then compares a UUID against a display name and throws all of them away. Typing the name works because both passes agree on that value.

## The fix

`rowMatchesFilter` now accepts a row when the filter matches **either** field, which is the SQL predicate expressed over the merged row:

```ts
if (
  filter.apiKey &&
  !matchesSearch(row?.apiKeyName || "", String(filter.apiKey)) &&
  !matchesSearch(row?.apiKeyId || "", String(filter.apiKey))
) {
  return false;
}
```

In-memory rows are unaffected: `buildCallLogListRows()` writes `apiKeyId: null, apiKeyName: null` on both the active and the completed entries, so an API-key filter still excludes them — they carry no attribution to match. There is a test for that so a later change cannot start letting unattributed rows through.

The free-text `search` branch had drifted the same way — `callLogs.ts:751` searches `cl.api_key_id`, the in-memory haystack did not — so `row.apiKeyId` joins that list too. Same root cause, one line, and it keeps the two layers describing the same thing.

## Tests

`tests/unit/call-logs-row-filter.test.ts`, 5 added cases: the dropdown's id matches; the name still matches; a different id does not; an in-memory row with both fields null stays excluded under an id filter and under a name filter; free-text search reaches the id.

Reverting only `route.ts` to the base commit and re-running the file:

```
not ok 5 - #12873 apiKey filter matches the key id sent by the dropdown
not ok 9 - #12873 free-text search reaches the key id, as the SQL predicate does
# tests 9  # pass 7  # fail 2
```

With the fix: 9 pass, 0 fail.

## Commands run

```
node --import tsx/esm --test tests/unit/call-logs-row-filter.test.ts      9 pass, 0 fail
eslint src/app/api/usage/call-logs/route.ts tests/...                     No issues found
prettier --check <both changed files>                                     clean
```

One disclosure about the diff: two hunks in `route.ts` (the `filter.status` branch and the `correlationId` branch) are prettier's, not mine — those lines are already over the line budget on `release/v3.8.51` and the repo's own pre-commit hook rewrapped them when it formatted the file. They are untouched by the fix; say the word and I will drop them.

I have not run the dashboard against a populated database — the reproduction in the issue is precise enough that the unit layer is where the defect lives, but the end-to-end confirmation (dropdown → rows appear) is yours to make if you want it before merge.

### DIFF
--- a/changelog.d/fixes/12894-call-logs-api-key-id-filter.md
+++ b/changelog.d/fixes/12894-call-logs-api-key-id-filter.md
@@ -1,0 +1,1 @@
+- **fix(api):** Dashboard → Logs now returns rows when an API key is picked from the dropdown: the in-memory filter pass matched the key name only, while the dropdown sends the key id ([#12894](https://github.com/diegosouzapw/OmniRoute/pull/12894), closes [#12873](https://github.com/diegosouzapw/OmniRoute/issues/12873))
--- a/src/app/api/usage/call-logs/route.ts
+++ b/src/app/api/usage/call-logs/route.ts
@@ -44,7 +44,10 @@ export function rowMatchesFilter(row: any, filter: Record<string, any>): boolean
     if (!(Number(row?.status) >= 400 || Boolean(row?.error))) return false;
   } else if (filter.status === "ok") {
     if (!(Number(row?.status) >= 200 && Number(row?.status) < 300)) return false;
-  } else if (typeof filter.status === "number" || (typeof filter.status === "string" && !isNaN(Number(filter.status)))) {
+  } else if (
+    typeof filter.status === "number" ||
+    (typeof filter.status === "string" && !isNaN(Number(filter.status)))
+  ) {
     if (Number(row?.status) !== Number(filter.status)) return false;
   }
 
@@ -57,13 +60,25 @@ export function rowMatchesFilter(row: any, filter: Record<string, any>): boolean
   if (filter.account && !matchesSearch(row?.account || "", String(filter.account))) {
     return false;
   }
-  if (filter.apiKey && !matchesSearch(row?.apiKeyName || "", String(filter.apiKey))) {
+  // The SQL layer resolves this filter against either column
+  // (`cl.api_key_name LIKE @apiKeyQ OR cl.api_key_id LIKE @apiKeyQ`), and the
+  // dashboard's API-key dropdown sends the id. Matching the name alone here
+  // dropped every persisted row the SQL layer had just returned, so selecting a
+  // key from the dropdown emptied the grid while typing its name worked. #12873
+  if (
+    filter.apiKey &&
+    !matchesSearch(row?.apiKeyName || "", String(filter.apiKey)) &&
+    !matchesSearch(row?.apiKeyId || "", String(filter.apiKey))
+  ) {
     return false;
   }
   if (filter.combo && !matchesSearch(row?.comboName || "", String(filter.combo))) {
     return false;
   }
-  if (filter.correlationId && !matchesSearch(row?.correlationId || "", String(filter.correlationId))) {
+  if (
+    filter.correlationId &&
+    !matchesSearch(row?.correlationId || "", String(filter.correlationId))
+  ) {
     return false;
   }
   if (filter.search) {
@@ -74,6 +89,9 @@ export function rowMatchesFilter(row: any, filter: Record<string, any>): boolean
       row?.providerDisplay,
       row?.account,
       row?.apiKeyName,
+      // The free-text SQL predicate covers `cl.api_key_id` too, so the id
+      // belongs in the in-memory haystack for the same reason. #12873
+      row?.apiKeyId,
       row?.comboName,
       row?.correlationId,
       row?.error,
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks for the precise root-cause trace here — the two-pass filter mismatch (SQL matching id
or name, the in-memory re-filter matching name only) was exactly right. Since opening this,
your own follow-up PR #12896 landed with the same fix (plus a couple of related gaps in the
`model` and free-text `search` filters) and is already on the release branch — issue #12873 is
closed as a result. Closing this one as subsumed by #12896 rather than merging a duplicate;
full credit for the fix stands with #12896.

diegosouzapw:
Obrigado — mesmo caso da #12919: o fix já está no tip, vindo de uma PR **sua**.

```
$ git log --oneline origin/release/v3.8.51 --grep="12896"
a6f28210 fix(logs): match the in-memory call-log filter to the SQL one it re-applies (#12896)
```

A **#12896** (mergeada em 10/09, num lote de 14) já introduziu exatamente o `rowMatchesFilter` com `apiKeyId || apiKeyName` que esta PR propõe, em `src/app/api/usage/call-logs/route.ts`. A issue **#12873** que esta fecharia já está **CLOSED**.

Fechando como subsumida.

PR #12897, opened 2026-09-07
fix(providers): read the vLLM context window from max_model_len

### LINKED ISSUE #12858: fix(providers): vLLM: model context window not read from max_model_len, falls back to 128K default
## Problem

When a vLLM server is connected to OmniRoute — through **any** of these paths:

1. the native `vllm` provider,
2. an **OpenAI-compatible** custom provider,
3. an **Anthropic-compatible** custom provider —

OmniRoute reports the model's context window as the **128,000-token default** instead of the real window that vLLM advertises.

### Reproduction

A vLLM server serving `Qwen3.8-27B-FP8` with `max_model_len` 250,000:

```bash
$ curl https://<vllm>/v1/models -H "Authorization: Bearer <key>"
{
  "data": [
    {
      "id": "qwen3.8",
      "owned_by": "vllm",
      "root": "/models/Qwen3.8-27B-FP8",
      "max_model_len": 250000
    }
  ]
}
```

vLLM correctly exposes the model's context window as **`max_model_len`**.

After connecting that server to OmniRoute (all three provider variants above) and querying the gateway:

```bash
$ curl http://localhost:20128/v1/models -H "Authorization: Bearer <key>"
# both advertised forms report the wrong window:
{ "id": "test/qwen3.8", "context_length": 128000 }
{ "id": "vllm/qwen3.8",    "context_length": 128000 }
```

Both entries report `context_length: 128000` — the global default — instead of the provider-declared `250000`.

## Root cause

`normalizeDiscoveredModels` in `src/lib/providerModels/modelDiscovery.ts` only recognizes these context-window field names (fix #3202 added `context_length` / `top_provider.context_length` for OpenRouter-style passthrough catalogs):

```ts
const inputTokenLimit = firstPositiveNumber(
  record.inputTokenLimit,
  record.context_length,
  record.contextLength,
  topProvider.context_length
);
```

`max_model_len` — the field vLLM (and vLLM-compatible servers) actually use — is not in the chain. A grep for `max_model_len` across the repository returns zero matches: no code path ever reads it.

Consequences:

- The discovered model is persisted to `syncedAvailableModels` with **no** `inputTokenLimit`.
- The token-limit resolver falls back to the provider/global default (128,000).
- The Feature 5004 self-correcting reconciler (`src/lib/contextWindowResolver.ts`) cannot repair this: it only pins windows that were *captured* at discovery time, and there is nothing to capture.

The native `vllm` provider is a plain `passthroughModels: true` local provider (`src/shared/constants/providers/local.ts`), so it funnels into the same normalization path — all three connection variants are affected identically.

## Impact

- `/v1/models` advertises a window ~2× smaller than what the upstream actually supports, so clients (and combo routing's `min()` across targets) under-utilize the model.
- Pre-dispatch context-window enforcement (`enforceOutputTokenBudget` in chatCore) can reject requests that the upstream would happily serve.
- Any local vLLM deployment — the most common self-hosted OpenAI-compatible serving stack — hits this.

## Suggested fix

Add `record.max_model_len` to the `firstPositiveNumber(...)` chain in `normalizeDiscoveredModels` (explicit `inputTokenLimit` keeps winning), plus a regression test modeled on `tests/unit/openrouter-context-length-3202.test.ts` (same class of bug: provider reports the window under a non-standard field name).

ntdat812:
Fixes #12858. Your root cause holds all the way down — `max_model_len` appears nowhere in the tree, so there was never a path by which a vLLM window could be captured.

## Why all three connection shapes broke together

`normalizeDiscoveredModels` is the single funnel: the native `vllm` provider is `passthroughModels: true` (`src/shared/constants/providers/local.ts:62`), and the OpenAI- and Anthropic-compatible custom providers hand the same raw `/v1/models` records to it. So this is not three bugs that look alike, it is one chain that never listed the field vLLM uses:

```ts
firstPositiveNumber(
  record.inputTokenLimit,
  record.context_length,
  record.contextLength,
  topProvider.context_length
)
```

Everything downstream follows from the miss you named: nothing is persisted, so the Feature 5004 reconciler has nothing to pin, and the resolver answers with the 128K default. A 250K deployment is advertised at half its size, `min()` across combo targets inherits the wrong number, and `enforceOutputTokenBudget` can refuse a request the upstream would have served.

## The fix

`record.max_model_len` joins the chain, after the OpenRouter names and before `top_provider.context_length`. `record.maxModelLen` goes in beside it for the same reason `contextLength` sits beside `context_length` — the two spellings of one field stay symmetric, and a JSON-camelizing proxy in front of vLLM is a real shape.

Ordering is deliberate rather than incidental: an explicit `inputTokenLimit` still wins, and a record carrying both `context_length` and `max_model_len` still resolves to `context_length`, so no catalog that works today changes meaning. `firstPositiveNumber` already rejects `0`, negatives and strings, so a server that reports `"250000"` is ignored rather than half-trusted — the test pins that, since it is behaviour a reader would otherwise assume either way.

## Tests

`tests/unit/vllm-max-model-len-12858.test.ts`, 6 cases modeled on `openrouter-context-length-3202.test.ts`: the vLLM record from your report maps to 250000; the camelCase spelling maps; an explicit `inputTokenLimit` wins; `context_length` stays ahead when both are present; `0` / `-1` / `"250000"` are ignored; a record with no window field still yields `undefined`.

Reverting only `modelDiscovery.ts` to the base commit:

```
not ok 1 - #12858 maps vLLM max_model_len into inputTokenLimit
not ok 2 - #12858 accepts the camelCase spelling too
# tests 6  # pass 4  # fail 2
```

With the fix: 6 pass, 0 fail. `tests/unit/openrouter-context-length-3202.test.ts` also runs green alongside it (5 pass), which is the check that matters for the ordering claim above.

## Commands run

```
node --import tsx/esm --test tests/unit/vllm-max-model-len-12858.test.ts \
                             tests/unit/openrouter-context-length-3202.test.ts    11 pass, 0 fail
eslint src/lib/providerModels/modelDiscovery.ts tests/unit/vllm-max-model-len-12858.test.ts
                                                                                  No issues found
prettier --check <both files>                                                     clean
```

Changed test files: `tests/unit/vllm-max-model-len-12858.test.ts` (new).

What I have not done: run a real vLLM server behind a connection and watch `/v1/models` report 250000. The normalizer is where the field is read and the tests exercise it directly, but if you have that server handy, the end-to-end check is one `curl` and worth having on the record.

### DIFF
--- a/changelog.d/fixes/12897-vllm-max-model-len.md
+++ b/changelog.d/fixes/12897-vllm-max-model-len.md
@@ -1,0 +1,1 @@
+- **fix(providers):** vLLM connections now advertise the real context window: model discovery reads `max_model_len` instead of falling back to the 128K default ([#12897](https://github.com/diegosouzapw/OmniRoute/pull/12897), closes [#12858](https://github.com/diegosouzapw/OmniRoute/issues/12858))
--- a/src/lib/providerModels/modelDiscovery.ts
+++ b/src/lib/providerModels/modelDiscovery.ts
@@ -310,10 +310,16 @@ export function normalizeDiscoveredModels(
     // `context_length` / `top_provider.context_length`, not `inputTokenLimit`.
     // Fall back across those names so synced models carry a real window instead
     // of the provider default (128K). Explicit `inputTokenLimit` still wins. #3202
+    // vLLM — and every server that copies its /v1/models shape — reports the
+    // window as `max_model_len`, the value the engine was actually started with.
+    // Without it a vLLM model syncs with no window at all and the resolver hands
+    // out the 128K default, understating a 250K deployment by half. #12858
     const inputTokenLimit = firstPositiveNumber(
       record.inputTokenLimit,
       record.context_length,
       record.contextLength,
+      record.max_model_len,
+      record.maxModelLen,
       topProvider.context_length
     );
     const outputTokenLimit = firstPositiveNumber(
--- a/tests/unit/vllm-max-model-len-12858.test.ts
+++ b/tests/unit/vllm-max-model-len-12858.test.ts
@@ -1,0 +1,68 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+
+import { normalizeDiscoveredModels } from "@/lib/providerModels/modelDiscovery";
+
+// Regression guard for #12858.
+//
+// vLLM's /v1/models reports the context window as `max_model_len` — the value
+// the engine was started with — and nothing else. The native `vllm` provider is
+// `passthroughModels: true`, and the OpenAI- and Anthropic-compatible custom
+// providers pass the same raw records through, so all three connection shapes
+// land here. Before the fix no window field matched, the model was persisted
+// with no `inputTokenLimit`, and `/v1/models` advertised the 128K provider
+// default for a deployment serving 250K.
+//
+// Same class of bug as #3202 (OpenRouter's `context_length`): the provider
+// declares the window under a name the normalizer does not read.
+
+test("#12858 maps vLLM max_model_len into inputTokenLimit", () => {
+  const [model] = normalizeDiscoveredModels([
+    { id: "qwen3.8", owned_by: "vllm", root: "/models/Qwen3.8-27B-FP8", max_model_len: 250000 },
+  ]);
+
+  assert.equal(model.id, "qwen3.8");
+  assert.equal(model.inputTokenLimit, 250000);
+});
+
+test("#12858 accepts the camelCase spelling too", () => {
+  // `contextLength` is already accepted next to `context_length`; keep the two
+  // spellings of this field symmetric with that pair.
+  const [model] = normalizeDiscoveredModels([{ id: "vendor/camel", maxModelLen: 131072 }]);
+
+  assert.equal(model.inputTokenLimit, 131072);
+});
+
+test("#12858 preserves an explicit inputTokenLimit over max_model_len", () => {
+  const [model] = normalizeDiscoveredModels([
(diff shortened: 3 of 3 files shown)


LazyGatto:
End-to-end confirmation from a real vLLM deployment (the check the PR body says was still outstanding).

Setup: self-hosted OmniRoute v3.8.50, native `vllm` provider (`passthroughModels: true`) pointed at a capability proxy in front of vLLM `:8000`, model `vllm/Qwen3.8-27B-FP8`.

Before this fix:
- vLLM `/v1/models` reports `max_model_len: 262144`.
- OmniRoute `/v1/models` reported `context_length: 128000` (the generic default), because `max_model_len` appears nowhere in the discovery chain — exactly the root cause in #12858.

Workaround we ran on the instance (works, but only until this lands): a manual context-window override via `PATCH /api/model-capability-overrides` with `{"target":"vllm/Qwen3.8-27B-FP8","key":"context_length","value":262144}`. After that, `/v1/models` correctly reports `context_length: 262144`.

So the diagnosis and the fix in this PR match our live case; `record.max_model_len` joining `firstPositiveNumber(...)` is the right place, and the ordering preserves existing catalogs.

Thanks for putting this together — happy to re-test against the merge result once it lands.

diegosouzapw:
Tiny, correct fix — and someone in the wild already confirmed it against a real vLLM
deployment (thanks LazyGatto for the independent repro in the comments, that's better proof
than my own probe). Local probe: 6/6 tests pass, including the precedence ordering between
`inputTokenLimit`, `context_length`, and `max_model_len`. Ready to merge as-is.

PR #12845, opened 2026-09-06
fix(combos): allow isHidden-only updates

smshagor-dev:
## Summary

- Describe the user-facing or operational change.

## Related Issues

- Closes #
- Related to #

## Validation

Choose the change type and focused loop from the
[Contribution Golden Path](../docs/ops/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite,
Vitest, the 60% coverage gate, and the production build all run in CI on this PR (#8329):

- [ ] Change type: provider / routing / UI / i18n / CLI / DB / build-deploy / other
- [ ] Focused tests and category gates from the golden path
- [ ] `npm run lint`
- [ ] Reconciled with the current active release base; focused checks rerun afterward
- [ ] Production-code changes include a new or updated automated test in this PR
- SonarQube is temporarily opt-in while the private project has no quota; it is not a PR gate.

## Tests Added Or Updated

- List every changed or added automated test file.
- If no production code changed, state that here.

## Coverage Notes

- If this PR changes `src/`, `open-sse/`, `electron/`, or `bin/`, explain which tests cover the change.
- If coverage moved down in any touched file, explain why and what follow-up task will recover it.

## Reviewer Notes

- Call out any risky areas, migrations, feature flags, or manual validation that reviewers should know about.

PR #12898, opened 2026-09-07
fix(combos): accept isHidden in updateComboSchema

### LINKED ISSUE #12836: fix(combos): updateComboSchema rejects isHidden-only updates despite stored visibility support
## Summary
The Combo update validation schema omits `isHidden`, although Combo storage and the builder already support this boolean. A visibility-only update is rejected; adding another recognized field still loses the visibility change because the unknown property is stripped.

Verified by executing the exported validation schema from the published 3.8.49 and 3.8.50 package sources with their matching Zod dependency. The same omission remains in current main's `src/shared/validation/schemas/combo.ts` inspected on September 6, 2026.

## Reproduction
```ts
updateComboSchema.safeParse({ isHidden: true });
updateComboSchema.safeParse({ isHidden: false });
```
Both return `success: false` with `No valid fields to update`. Controls `{ isActive: false }` and `{ description: "fixture" }` succeed.

This is offline execution of the shipped schema, not a production write.

## Root cause and minimal fix
`updateComboSchema` has `isActive: z.boolean().optional()` but no `isHidden`. Its empty-update refinement also omits that field. Add `isHidden: z.boolean().optional()` and include `value.isHidden === undefined` in the empty-update check. Do not use unrestricted passthrough or weaken existing authorization and model/config validation.

The native PUT handler spreads validated input into its update object, so the missing validation field is the relevant seam. `src/lib/db/combos.ts` stores visibility and `src/lib/combos/builderOptions.ts` filters on it.

## Verification of the minimal change
After the two schema additions, actual schema execution for each reviewed package accepts and retains true/false visibility-only updates and mixed description/visibility updates. String, number, null and empty-object cases remain rejected. This is seven expected cases per version. The original package files were not mutated.

An issue search found hidden provider-model catalog and picker reports, but no matching report for mutation of a Combo's own visibility field.

ntdat812:
Fixes #12836. Confirmed against `release/v3.8.51`, and the field is more load-bearing than "stored but unvalidated" suggests.

## The field has readers, just no writer

`isHidden` is not vestigial:

| site | what it does with it |
|---|---|
| `sqliteComboRepository.createCombo` | writes it explicitly — `isHidden: Boolean(data.isHidden)` |
| `src/lib/combos/builderOptions.ts:798` | `.filter((combo) => combo.isHidden !== true && combo.isActive !== false)` — hidden combos leave the builder's option list |
| `dashboard/combos/page.tsx:876` | drops them from the grid, filtering the fetched list on `!c.isHidden` |
| `PUT /api/combos/[id]` | spreads the **validated** body into `updateCombo` |

That last row is the only path a client has to the field, and `updateComboSchema` did not list it. Zod strips unknown keys, so `{ isHidden: true }` reached the empty-update refinement with nothing left and came back `No valid fields to update`; `{ description: "x", isHidden: true }` was accepted and wrote the description alone. A combo could therefore be born hidden but never be hidden or unhidden afterwards.

Storage is not the obstacle — `updateCombo` merges `{ ...current, ...data }` and serialises the whole record into the `data` column, so the moment validation stops discarding the key it persists. That is asserted rather than assumed: the tests below go through the real repository and re-read the row.

## The fix

Two lines, exactly as you scoped it: `isHidden: z.boolean().optional()` on the object, and `value.isHidden === undefined` in the empty-update refinement so a visibility-only body counts as a real update. No passthrough, no loosening of the model, config or authorization validation around it.

`createComboSchema` has the same omission and I have deliberately left it alone: creation already defaults the field through `Boolean(data.isHidden)`, "new combos are visible" is a defensible rule, and widening creation is a product decision rather than a bug fix. Happy to add it in this PR if you want the two schemas symmetric.

## Tests

`tests/unit/combo-hidden-update-12836.test.ts`, 5 cases modeled on `combo-clear-agent-features-12158.test.ts` (same seam, same repository round-trip):

- a visibility-only update parses for `true` and for `false`, while `{}` still fails the refinement;
- a mixed `{ description, isHidden }` update keeps both fields;
- `"true"`, `1`, `null` and `{}` are all still rejected;
- hiding a combo through `createCombo` → `updateCombo` → `getComboById` reads back `isHidden: true` — the state the two filters above act on;
- unhiding works, and a later unrelated update leaves the visibility where it was.

Reverting only `combo.ts` to the base commit:

```
not ok 1 - #12836 a visibility-only update is a valid update
not ok 2 - #12836 visibility survives a mixed update instead of being stripped
not ok 4 - #12836 hiding a combo persists and reads back
not ok 5 - #12836 unhiding it again works, and an unrelated update leaves it alone
# tests 5  # pass 1  # fail 4
```

(The type-rejection case passes on both sides — an unknown key is stripped, so a bad value was never accepted either. It is there to keep the new field from being widened later.)

## Commands run

```
node --import tsx/esm --test tests/unit/combo-hidden-update-12836.test.ts          5 pass, 0 fail
node --import tsx/esm --test combo-clear-agent-features-12158 shared/comboSchema \
                             combo-empty-models combo-context-length             41 pass, 0 fail
prettier --check <both changed files>                                            clean
```

Changed test files: `tests/unit/combo-hidden-update-12836.test.ts` (new).

Not covered: no dashboard control sends this field yet, so nothing in the UI exercises the path today — this fixes the API so a client (or a future toggle) can use it, which is what the issue reports.

### DIFF
--- a/changelog.d/fixes/12898-combo-hidden-update-schema.md
+++ b/changelog.d/fixes/12898-combo-hidden-update-schema.md
@@ -1,0 +1,1 @@
+- **fix(combos):** A combo's visibility can be changed through the API again: `updateComboSchema` accepts `isHidden`, so a visibility-only update is no longer rejected as empty and a mixed update no longer drops it ([#12898](https://github.com/diegosouzapw/OmniRoute/pull/12898), closes [#12836](https://github.com/diegosouzapw/OmniRoute/issues/12836))
--- a/src/shared/validation/schemas/combo.ts
+++ b/src/shared/validation/schemas/combo.ts
@@ -417,6 +417,11 @@ export const updateComboSchema = z
     strategy: comboStrategySchema.optional(),
     config: comboRuntimeConfigSchema.optional(),
     isActive: z.boolean().optional(),
+    // Stored on the combo record and honoured by the readers — the builder's
+    // option list and the dashboard grid both filter on it — but omitted here,
+    // so the one endpoint a client can flip it through stripped the field and
+    // a visibility-only update was rejected as empty. #12836
+    isHidden: z.boolean().optional(),
     allowedProviders: z.array(z.string().trim().min(1).max(200)).max(100).optional(),
     allowedModelFamilies: z.array(z.string().trim().min(1).max(100)).max(100).optional(),
     // Nullable like `description` and `context_length` above: an absent field means
@@ -441,6 +446,7 @@ export const updateComboSchema = z
       value.strategy === undefined &&
       value.config === undefined &&
       value.isActive === undefined &&
+      value.isHidden === undefined &&
       value.allowedProviders === undefined &&
       value.allowedModelFamilies === undefined &&
       value.system_message === undefined &&
--- a/tests/unit/combo-hidden-update-12836.test.ts
+++ b/tests/unit/combo-hidden-update-12836.test.ts
@@ -1,0 +1,102 @@
+/**
+ * #12836 — a combo's visibility must be updatable through the API.
+ *
+ * `isHidden` is stored on the combo record and honoured by its readers: the
+ * builder's option list drops hidden combos (`src/lib/combos/builderOptions.ts`)
+ * and so does the dashboard grid. `createCombo` writes the field explicitly.
+ * The only endpoint a client can flip it through is `PUT /api/combos/[id]`,
+ * which spreads the *validated* body into the update — and `updateComboSchema`
+ * never listed `isHidden`, so zod stripped it. A visibility-only update was
+ * rejected as "No valid fields to update", and pairing it with a recognized
+ * field made the request succeed while silently dropping the visibility change.
+ */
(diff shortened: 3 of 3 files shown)


diegosouzapw:
Clean, well-scoped fix — the root-cause writeup (four concrete `isHidden` read/write sites, only the `PUT` validation path missing it) is exactly right, and the test suite is thorough (schema-level cases plus a real create→hide→re-read→unhide round trip through the actual combo repository). Re-ran `tests/unit/combo-hidden-update-12836.test.ts` locally and it's 5/5 green — the "API Route Typecheck" failure showing on this PR's CI run looks like a stale broad base-red from that day (every fast-path job failed at once, not something a 2-line schema change would cause) rather than a real regression. Merging this one; there's a near-identical PR (#12845) for the same issue that I'm closing as a duplicate in your favor since this one has the fuller test coverage and a changelog fragment. Thanks for the careful diagnosis.

PR #12905, opened 2026-09-07
fix(thinking): parse/scrub DSML tool-call markers and recognize adaptive thinking

initguru:
## Summary

- When the internal LLM inserts DSML tool-call markers (`<tool_call>...</tool_call>`, etc.) into reasoning, parse them into normal tool calls and scrub them from response text. Add the new `dsmlToolCalls.ts` module and wire it into `openai-to-claude.ts`/`stream.ts`/`chatCore.ts`/`translator/index.ts`.
- Recognize Claude's adaptive compression thinking mode on the response side.

## Related Issues

- Closes #12905

## Validation

- [x] Change type: provider / routing
- [x] Focused tests: `node --import tsx/esm --test tests/unit/dsml-tool-calls.test.ts tests/unit/translator-resp-dsml-integration.test.ts tests/unit/thinking-active-claude-adapter.test.ts tests/unit/translator-resp-openai-to-claude.test.ts`
- [ ] `npm run lint`
- [ ] Reconciled with `release/v3.8.51` base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR

## Tests Added Or Updated

- `tests/unit/dsml-tool-calls.test.ts`
- `tests/unit/translator-resp-dsml-integration.test.ts`
- `tests/unit/thinking-active-claude-adapter.test.ts`
- `tests/unit/translator-resp-openai-to-claude.test.ts` (updated)

## Coverage Notes

- (Measured after implementation)

## Reviewer Notes

- Canonical evidence: `dsmlToolCalls.ts` does not exist; 0 occurrences of `dsml` in the canonical tree. Canonical `normalizeClaudeAdaptiveThinking` (`chatCore.ts:2677`) is the canonical adaptive-compression service and is unrelated to response-side DSML recognition.
- Reference local commit `f6eb328ba`. New file `open-sse/utils/dsmlToolCalls.ts`.

### changelog.d/features/12905-dsml-tool-call-markers-adaptive-thinking.md
```
- **feat(sse):** parse/scrub DSML tool-call markers embedded in reasoning and recognize adaptive thinking on the response side — `dsmlToolCalls.ts` module + translator/stream/handler wiring ([#12905](https://github.com/diegosouzapw/OmniRoute/pull/12905))
```

---

---

## Summary

- Upstream echoes a system directive/prompt preamble in `/v1/messages` responses. Strip it with the new `directivePreambleStripper.ts` module.
[...]

PR #12917, opened 2026-09-07
fix(translator): drop unsigned reasoning on Claude output

atescivitci-cmd:
## Summary
- stop converting non-Anthropic OpenAI-compatible `reasoning_content` into Claude `thinking` blocks
- apply the same suppression to the non-streaming OpenAI chat-completion to Claude conversion path
- update response translator regression tests to assert unsigned reasoning is not emitted as Claude thinking
- prune now-stale ESLint suppression count for the changed test file

## Why
`thinking` blocks sent back to Anthropic must carry Anthropic-valid signatures. Reasoning fields from OpenAI-compatible providers do not have those signatures, so translating them into Claude `thinking` creates replay history that Anthropic rejects with `Invalid signature in thinking block`.

## Validation
- `node --import tsx/esm --test tests/unit/translator-resp-openai-to-claude.test.ts`
- `node --import tsx/esm --test tests/unit/translator-openai-to-claude.test.ts`
- `npx eslint open-sse/handlers/responseTranslator.ts open-sse/translator/response/openai-to-claude.ts tests/unit/translator-resp-openai-to-claude.test.ts --cache --cache-location .eslintcache --suppressions-location config/quality/eslint-suppressions.json`
- `npm run typecheck:core -- --pretty false`

⚠️ base-red inherited: #12732

### DIFF
--- a/config/quality/eslint-suppressions.json
+++ b/config/quality/eslint-suppressions.json
@@ -5256,7 +5256,7 @@
   },
   "tests/unit/translator-resp-openai-to-claude.test.ts": {
     "@typescript-eslint/no-explicit-any": {
-      "count": 14
+      "count": 12
     }
   },
   "tests/unit/translator-thinking-provider-compat-2043.test.ts": {
--- a/open-sse/handlers/responseTranslator.ts
+++ b/open-sse/handlers/responseTranslator.ts
@@ -719,11 +719,10 @@ function convertOpenAINonStreamingToClaude(
 
   const reasoningText = resolveReasoningText(messageObj);
   if (reasoningText) {
-    hasTextOrReasoning = true;
-    content.push({
-      type: "thinking",
-      thinking: reasoningText,
-    });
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Clean, well-reasoned fix — dropping unsigned reasoning instead of laundering it into visible
text is the right call (avoids both the Anthropic signature-replay error and a chain-of-thought
privacy leak), and it's applied consistently to both the streaming and non-streaming paths. Ran
`translator-resp-openai-to-claude.test.ts` at your head — 13/13 pass.
One heads-up for coordination: #12905 touches the exact same code region
(`openai-to-claude.ts:300-333`, per its own PR description) with a different, gated approach —
`git merge-tree` confirms a real conflict across 3 files including the shared test. Flagging
this for the maintainer's merge-order decision, not a blocker on your side. Only ask: add a
`changelog.d/fixes/` fragment.

PR #12893, opened 2026-09-07
feat(codex): surface minimum client version guidance

guanbear:
## Summary

Codex model discovery already reads OpenAI's public Codex catalog and filters entries whose `minimal_client_version` is newer than OmniRoute's reported client version. Until now that filter was silent, so a newly launched ChatGPT/Codex model could simply disappear from sync with no actionable explanation.

This PR preserves that safety gate and surfaces a Codex-specific warning containing the skipped model IDs, the current HTTP client profile version, the highest minimum version required by the catalog, and the existing `CODEX_CLIENT_VERSION` escape hatch.

This deliberately covers only the `codex` provider, which replays a ChatGPT OAuth token to `chatgpt.com/backend-api/codex`. It does not implement a generic updater for OmniRoute's other providers and does not imply that a standalone Codex CLI must be installed. The separate `codex-app-server` provider remains the only path where the local CLI binary is relevant.

## Related Issues And PRs

- Related to #12761
- Complements the immediate version bumps in #12827 and #12759

## Changes

- Inspect visible, API-supported entries in the official Codex catalog for unmet `minimal_client_version` requirements.
- Preserve compatibility diagnostics alongside the existing ETag catalog cache.
- Return the warning during live, GitHub, cached, and local-catalog discovery paths, including when every remote entry is filtered as too new.
- Keep incompatible models filtered rather than advertising models that the current HTTP client profile cannot safely call.
- Add pure parsing/formatting coverage and provider-route regression coverage.

## Validation

- `node --import tsx/esm --test tests/unit/provider-models-discovery-split.test.ts tests/unit/provider-models-route-codex.test.ts` (40 passed)
- `node node_modules/typescript/bin/tsc --pretty false -p tsconfig.typecheck-core.json`
- Focused ESLint on the four changed TypeScript files
- `git diff --check`

## Coverage Notes

Tests cover visible versus hidden/API-disabled models, compatible versus incompatible minimum versions, the actionable warning text, normal live discovery, and the all-remote-models-too-new fallback case.

The repository-wide changelog-integrity check remains base-red because `changelog.d/fixes/reset-aware-model-family.md` does not start with a Markdown bullet. This PR does not touch that file; its own fragment starts with `- `.

## Reviewer Notes

- No model name or version is hardcoded in the diagnostic logic.
- No package manager command is executed by the server.
- `CODEX_CLIENT_VERSION` remains an operator-controlled temporary override; upgrading OmniRoute is presented first.
[...]

PR #12929, opened 2026-09-07
fix(pack): allow the opencode-plugin-v2 workspace package in the artifact policy (base-red)

guanbear:
## Summary (scope updated)

The release wave around #12870 added `@omniroute/opencode-plugin-v2/*` to the npm pack artifact but never registered its prefix in `PACK_ARTIFACT_ROOT_ALLOWED_PATH_PREFIXES`, so the pack-policy gate fails on the release tip for every open PR. This adds the one-line prefix entry alongside the existing v1 `@omniroute/opencode-plugin/`.

## History

This PR originally carried the full set of release-tip base-reds (stream highWaterMark restoration, chatCore union narrowing, doc count drift). The maintainers have since landed equivalents themselves — #12963 (pin the ok variant), #12970 (sync migration/strategy counts), and the non-streaming-leg restore wave — so those commits are dropped from this branch and all that remains is the still-unfixed pack policy entry.

## Validation

- `validate-pack-artifact.ts --policy-only` → ✅ pass (fails on the release tip).
- Full gate suite on this tree: api-typecheck / open-sse-typecheck / file-size / docs-sync / changelog-integrity all OK (inherited from the tip plus the one policy line).


guanbear:
Updated this base-red fix to keep every gate green on its own tree:

- The restoration is compacted to +1 line against the stream.ts cap (no doc line on the `StreamOptions` field, default inlined at both strategy entries, and the `createPassthroughStreamWithLogger` half is not restored — its only caller passes 11 args). Cap raised 3072→3073 with a justified baseline note per the documented last resort.
- Also bundled two further base-reds blocking every open PR: stale migration counts (README/AGENTS.md/llm.txt said 169, code has 170 — llm.txt propagated to all 41 locale mirrors) and the `reset-aware-model-family` changelog fragment missing its markdown bullet.

Local validation on this tree: api-typecheck OK (289 within baseline), open-sse-typecheck OK (0), file-size OK, docs-sync/docs-counts/changelog-integrity OK, pack-policy OK, mutation-test-coverage OK, and the GLM + stream/SSE suites (153 tests) pass. #12893 carries the same commits as cherry-picks so its CI can go green in parallel.

guanbear:
Rebased and narrowed: the upstream fixes (#12963, #12970, #12990) covered the stream/chatCore/docs base-reds this PR originally carried, so the branch now contains only the pack-policy allowlist entry that is still red on the tip.

guanbear:
Same treatment: merged the current tip and re-synced the migration count (171→172) so CI runs on a fresh tree. Remaining red on the tip (volcengine `no-explicit-any` ESLint errors from #12950) is unrelated to this one-line change.

diegosouzapw:
Good catch on the missing allowlist entry — the pack-policy gap was real. It ended up landing
on the tip through a separate maintenance sweep (#12959) that independently added the same
`@omniroute/opencode-plugin-v2/` line, alongside the doc-count half of this PR (which had
since gone stale — the tip moved past the 172 you targeted). Closing as already covered.

guanbear:
Closing as already covered — thanks for confirming the gap was real and for landing it via the #12959 maintenance sweep (same allowlist line, plus the fresher doc counts). Nothing left to carry here.

PR #12942, opened 2026-09-07
fix(resilience): direct-fetch response-start timeout no longer escapes as an unhandledRejection and kills the process

### LINKED ISSUE #12861: fix(backend): proxyFetch: DIRECT_RESPONSE_START_TIMEOUT escapes as unhandledRejection → uncaughtException, server process exits
## Summary

The server process died with an `uncaughtException` carrying the direct-fetch response-start timeout:

```
⨯ unhandledRejection: Error [TimeoutError]: Direct response did not start within 30000ms — retrying on a fresh socket
⨯ uncaughtException:  Error [TimeoutError]: Direct response did not start within 30000ms — retrying on a fresh socket
    at Timeout._onTimeout (.build/next/server/chunks/_10p8fpv._.js:1:200)
    at async s (.build/next/server/chunks/_10p8fpv._.js:1:343)
    at async _ (.build/next/server/chunks/_10p8fpv._.js:1:10104)
    at async r (.build/next/server/chunks/_1epajp2._.js:1:53066)
    at async N.execute (.build/next/server/chunks/_1epajp2._.js:1:60468)
    at async N.execute (.build/next/server/chunks/_00yy8jx._.js:9:6751)
    at async d (.build/next/server/chunks/open-sse_0sa910h._.js:123:15337) {
  code: 'DIRECT_RESPONSE_START_TIMEOUT'
}
Node.js v26.8.1
```

Exit code 7, process restarted by the supervisor, full startup sequence re-run. A recoverable upstream stall — the timeout that exists precisely so the request can be retried on a fresh socket — must not take the process down.

## What the timeout is

`open-sse/utils/directResponseStartTimeout.ts` → `directFetchWithBoundedResponseStart()` arms a timer (`OMNIROUTE_DIRECT_HEADERS_TIMEOUT_MS`, default 30 s) that aborts the attempt's `AbortController` with a `TimeoutError` (`code: DIRECT_RESPONSE_START_TIMEOUT`). `open-sse/utils/proxyFetch.ts` catches it on attempt 0 and retries once on the no-keep-alive dispatcher, otherwise rethrows. That path normally works: the same error is regularly logged as a plain 504 (`[504]: Direct response did not start within 30000ms — retrying on a fresh socket`) with the server staying up.

## What went wrong this time

The stack's top frame is `Timeout._onTimeout` followed by `async` frames: the rejection surfaced from inside the timer callback's own async chain, not from the awaited `fetch` in `proxyFetch`. So the abort reason was delivered to a promise nobody awaited any more (the attempt already settled or its awaiting frame torn down — client disconnect, or the merged abort signal firing first), which makes the rejection unhandled. There is no process-level `unhandledRejection` handler (the only one in the tree is `src/server/ws/liveServer.ts`, scoped to the WebSocket server), and Node's default `--unhandled-rejections=throw` promotes it to `uncaughtException` → exit.

Not reproduced on demand: it needs an upstream that stalls for the full timeout at the same moment the attempt is torn down.

## Expected

Two independent layers:

1. Never let the abort reason escape: in `directFetchWithBoundedResponseStart()`, guard the timer callback so aborting an already-settled attempt cannot reject anything that is not awaited (no-op if the attempt already resolved/rejected, or attach a `catch` to whatever the callback awaits).
2. A process-level `unhandledRejection` handler in the Node instrumentation path that logs the error's `code` and keeps the process alive — `liveServer.ts` already does this for the WS server for the same reason ("becomes an uncaughtException that kills the server"). A gateway that fans out to many upstreams should not be one stray rejection away from a restart.

## Environment

OmniRoute v3.8.51, Node.js 26.8.1, Linux container; `OMNIROUTE_DIRECT_HEADERS_TIMEOUT_MS` unset.

geekyNads:
Fixes #12861

The server process died with an `uncaughtException` carrying `DIRECT_RESPONSE_START_TIMEOUT` — a recoverable signal `proxyFetch.ts` already retries once on a fresh no-keep-alive dispatcher (normally logged as a plain 504, server stays up). The crash happened because the abort reason escaped to a promise nobody was awaiting anymore — the stack's top frame was `Timeout._onTimeout`, not the awaited fetch. With no process-level `unhandledRejection` handler outside the WS server's own scoped one, Node's default `--unhandled-rejections=throw` promoted it to an `uncaughtException` and took the whole process down.

### Two independent layers, per the issue's own analysis

1. **`open-sse/utils/directResponseStartTimeout.ts`** — the timer callback unconditionally aborted even if the wrapped fetch had already settled (a narrow race between the timer macrotask and the fetch promise's `finally`-block `clearTimeout`). Added a `settled` flag so the timer becomes a no-op once the attempt has resolved or rejected. Side benefit: also stops the timer from spuriously aborting a still-streaming response body after headers already started successfully.

2. **`src/shared/utils/httpClientAbortGuard.mjs`** — this file already implements exactly the "last-resort safety net" pattern the issue asks for, and is already installed in 3 places (`liveServer.ts`, `embedWsProxy.ts`, `apiBridgeServer.ts`). It just didn't recognize `DIRECT_RESPONSE_START_TIMEOUT`, and — the bigger gap — **it was never installed in the actual production server entrypoint at all**. Added `isRecoverableUpstreamTimeoutError()`, extended `shouldSwallowUncaught()`, logging now includes `err.code`, and installed the guard in `scripts/dev/standalone-server-ws.mjs` — confirmed via `assembleStandalone.mjs` to be what actually ships as `dist/server-ws.mjs` and what `omniroute serve`/Docker/Electron spawn in production. `run-next.mjs` (dev) already had this guard; production never did — the exact same dev/prod asymmetry this file's own comments already document from a prior fix.

### Testing
- `tests/unit/direct-response-start-timeout-settled-guard-12861.test.ts` — covers normal resolve/timeout/passthrough, plus two tests that deterministically force the exact race (neutering `clearTimeout` + a detached internal promise mimicking real fetch/undici's abort-tied background work — mock timers alone gave a false pass on this, since `clearTimeout` genuinely cancels a mocked timer). Confirmed these fail with the exact `unhandledRejection`/`Timeout._onTimeout` signature from the report on the unfixed code, and pass with the fix.
- `tests/unit/http-client-abort-guard-direct-timeout-12861.test.ts` — pure predicate coverage, including confirming no regression on the original client-abort cases.
- Did not add an end-to-end test triggering a real process `unhandledRejection` through `installProcessCrashGuard()` itself — Node's own test runner installs a competing `unhandledRejection` listener for test-failure reporting, incompatible with `node --test`. Likely why the 3 existing production call sites have no direct test coverage either.
- All 12 new tests + typecheck + cycle-check verified clean on real hardware (not just sandbox).

### Checklist
- [x] Tests pass, added for production code change
- [x] Linting passes
- [x] Changelog fragment added: `changelog.d/fixes/12861-direct-fetch-timeout-unhandled-rejection.md`
- [x] TypeScript types added where needed


HouMinXi:
Found one blocking issue while evaluating this for cherry-pick into our deployment: the PR adds `import "./httpClientAbortGuard.mjs"` to `scripts/dev/standalone-server-ws.mjs`, but never adds the file to `EXTRA_MODULE_ENTRIES` in `scripts/build/assembleStandalone.mjs`. The assembled bundle therefore ships a `server-ws.mjs` whose import cannot resolve, and the production image crash-loops at boot with `ERR_MODULE_NOT_FOUND`.

The repo already has a regression guard for exactly this, and it goes red on this branch:

```
$ node --test tests/unit/build/assemble-standalone.test.ts
...
not ok - every relative import of standalone-server-ws.mjs is shipped into the bundle
# pass 5 / fail 1
```

(`server-ws.mjs imports ./httpClientAbortGuard.mjs but EXTRA_MODULE_ENTRIES does not ship it — the bundle would crash at boot`)

CI being red across the board on this base seems to have hidden it.

One nuance for the fix: shipping the `scripts/dev/httpClientAbortGuard.mjs` shim does not work. In the bundle the shim lands at `<root>/httpClientAbortGuard.mjs`, and its `../../src/shared/utils/httpClientAbortGuard.mjs` re-export then resolves outside the bundle. What works is shipping the real implementation (it is self-contained, no relative imports) under the file name `server-ws.mjs` asks for:

```js
{
  label: "http client abort guard (server-ws.mjs dependency)",
  src: ["src", "shared", "utils", "httpClientAbortGuard.mjs"],
  dest: ["httpClientAbortGuard.mjs"],
},
```

With that entry the assemble regression test goes back to 6/6 and the bundle boots.

For the record, the fix itself reviewed well from here: the settled-guard tests form a real red/green pair (reverting the guard makes the two #12861 tests fail as designed), the guard predicates classify disjoint code sets, and installing the crash guard in `server-ws.mjs` is the right spot since that is the process `run-standalone.mjs` actually spawns. Thanks for picking this up. We hit the same crash in production today (six restarts in under an hour behind a hung upstream), so getting this in matters to us.

HouMinXi:
Follow-up: #13636 now carries this branch plus the standalone-manifest entry from the comment above, plus a guard extension for the two sibling escapes we hit in production after deploying it (hedge-cancelled and fetch-failed classes). Happy to split if you would rather land the manifest fix here first.

HouMinXi:
One more thing we caught while re-reviewing the stacked branch: this PR renames the guard log text from `swallowed client-abort` to `swallowed benign uncaughtException`, which silently breaks `tests/unit/httpClientAbortGuard-default-logger.test.mjs` — it finds its handler by grepping the listener source for the old string. The test goes red on this PR alone (verified by stashing our follow-ups). #13636 updates the test to the new wording; you may want the same fix here if you land first.

diegosouzapw:
Verified independently: the race you describe is real, and your fix (the `settled` flag guard)
is the right shape for it. All 12 tests pass on your branch. One heads-up for the maintainers'
merge order — a later PR (#13636) opened about a week after yours ended up reproducing this exact
fix verbatim as its first commit before adding more coverage on top; we're recommending this PR
merge first so you get credit as the original author, with #13636 rebased to carry only its
additional work. This looks merge-ready as-is.

diegosouzapw:
Closing this as landed: the fix shipped on `release/v3.8.51` through #13636, whose description credits this PR as the origin of the settled-guard approach. One thing I want on the record here, because the squash commit didn't carry a co-author trailer: #12942 was first, and the direct-fetch response-start timeout no longer escaping as an unhandledRejection is your diagnosis and your design. That credit is yours. Thank you for it.

PR #12948, opened 2026-09-07
fix(soniox): pass client parameters through and surface speaker diarization

### LINKED ISSUE #12947: fix(providers): Soniox: no client parameter reaches the provider — diarization and `context` are unreachable
**OmniRoute Version:** 3.8.50 (reproduced live); defect verified unchanged at `release/v3.8.51` HEAD `d6f31501`
**Component:** `open-sse/handlers/audioTranscription.ts` → `handleSonioxTranscription`
**Severity:** requested parameters are dropped silently; a 200 is returned either way

## Summary

No client-supplied parameter reaches the Soniox provider. `handleSonioxTranscription` takes no form data, builds its job body from a fixed three-key object, and reduces the transcript to `{ text }`. The request returns 200 and the flag has no effect, so from the caller's side an unsupported parameter and a dropped one look identical.

Two capabilities are unreachable as a result: **speaker diarization** and Soniox's **`context`** vocabulary parameter. The per-token `speaker` attribution Soniox already returns is discarded before the response is built.

## Reproduction

```bash
curl -s $BASE/v1/audio/transcriptions \
  -H "Authorization: Bearer $KEY" \
  -F "file=@sample.mp3" \
  -F "model=soniox/stt-async-v5" \
  -F "enable_speaker_diarization=true" \
  | jq 'keys'
```

Prints `["text"]`. Same for `diarization=true`, `speaker_labels=true`, `response_format=verbose_json`, and `timestamp_granularities[]=word` — all accepted with a 200, none with any effect.

There is no client-side workaround: `/v1/providers/{provider}/*` covers chat, embeddings and images but not audio, so Soniox cannot be reached natively to bypass the flattening.

## This is not an account or plan limitation

Checked first, so the report isn't chasing a provider-side gap. Calling `api.soniox.com` directly with the same audio and the same key that the gateway uses:

```
POST /v1/transcriptions  { enable_speaker_diarization: true, context: "…" }   → 201  (both accepted)
transcript token fields: text, start_ms, end_ms, confidence, speaker, language
distinct speakers: ["1", "2"]        tokens per speaker: { "1": 147, "2": 191 }
```

The provider returns per-token speaker attribution on this plan. Everything below is gateway behaviour.

## Root cause — three places, at `release/v3.8.51`

**1. The handler is the only STT handler denied the form data.** At the dispatch site, Deepgram is passed it and the async trio is not:

```ts
815:  return handleDeepgramTranscription(providerConfig, file, modelId, token, formData);
819:  return handleAssemblyAITranscription(providerConfig, file, modelId, token);
823:  return handleGladiaTranscription(providerConfig, file, modelId, token);
827:  return handleSonioxTranscription(providerConfig, file, modelId, token);
```

`handleSonioxTranscription` (`:358`) has no `formData` parameter, so nothing the caller sent is in scope.

**2. The job body is a fixed literal** (`:374-379`) — there is no seam for a caller-controlled field:

```ts
body: JSON.stringify({
  model: modelId,
  file_id: fileId,
  enable_language_identification: true,
}),
```

**3. The transcript is reduced to its text** (`:424`). When diarization is on, Soniox's `tokens[]` carry `speaker`, `start_ms` and `end_ms`; all of it is dropped:

```ts
return Response.json({ text }, { headers: { ...CORS_HEADERS } });
```

## Why the fabricated-silence matters

A dropped parameter and an unsupported one are indistinguishable to the caller. This one took source reading to diagnose rather than a single failed request, which is the expensive way to find a bug.

There is also a volume argument. On a 16-minute call recording, 82% of the transcript is IVR and hold announcements — 1438 words of which only 256 are the actual conversation. With speaker labels the recorded-system voice becomes one label and is removable in a single pass. Without them every downstream consumer pays for and processes the noise.

## Proposed fix

Give the handler the form data, as Deepgram already has it, and map what the caller asked for onto the Soniox job. Add job-body keys **only when requested**, so a request carrying no options produces a byte-identical body and the existing strict deep-equality assertion in `tests/unit/audio-soniox-provider.test.ts:75` keeps passing. Keep `{ text }` as the default response, and surface the token stream as speaker-labelled segments when diarization or `verbose_json` is requested.

## Related, not addressed there

`handleAssemblyAITranscription` (`:229`) and `handleGladiaTranscription` (`:297`) have the same signature gap. AssemblyAI supports diarization at the provider level, so whatever response contract is settled on for Soniox probably wants to fit it too. Left out of the PR to keep the diff reviewable — happy to follow up.

Credit: reported with a precise root-cause analysis by Vigen, who tested four real Persian phone recordings against the gateway.

amirrezakm:
## Summary

- No client-supplied parameter reached the Soniox provider. `handleSonioxTranscription` took no `formData`, built its job body from a fixed three-key literal, and reduced the transcript to `{ text }` — so `enable_speaker_diarization` was accepted with a 200 and silently dropped, and the per-token `speaker` attribution Soniox already returns was discarded. Diarization and Soniox's `context` vocabulary parameter were both unreachable. (#12947)
- The handler now receives the request's form data, exactly as `handleDeepgramTranscription` already did, and maps what the caller asked for onto the job: diarization, `context` verbatim, and `language` as a Soniox language hint.
- **Job-body keys are added only when requested.** A request carrying no options produces a byte-identical body, which is why the existing strict deep-equality assertion in `tests/unit/audio-soniox-provider.test.ts:75` still passes untouched.
- **The default response contract is unchanged.** Callers who ask for nothing still get exactly `{ text }`. When diarization or `response_format=verbose_json` is requested, the token stream is collapsed into contiguous single-speaker runs and returned as OpenAI-style `segments` carrying `speaker`, plus `words` when word granularity is asked for.

Accepted parameters:

| Sent | Mapped to |
|---|---|
| `enable_speaker_diarization` / `diarization` / `speaker_labels` | `enable_speaker_diarization: true` |
| `context` | `context` (verbatim) |
| `language` | `language_hints: [lang]` |
| `response_format=verbose_json` | richer response without requesting diarization |
| `timestamp_granularities[]=word` | adds `words[]` |

Three spellings for diarization because callers reach for whichever their previous provider used; Soniox's own name, the generic one, and Deepgram's.

## Related Issues

- Closes #12947

## Validation

- [x] Change type: provider (Soniox STT transcription path)
- [x] Focused tests and category gates from the golden path (commands below)
- [x] `npm run lint` — clean on both touched files
- [x] Reconciled with the current active release base: rebased onto `release/v3.8.51` HEAD `d6f31501`; upstream has not touched `audioTranscription.ts` since, and the focused loop was rerun afterwards

Commands run locally:

```bash
node --import tsx/esm --test tests/unit/soniox-diarization-and-client-params.test.ts   # 9 pass
node --import tsx/esm --test tests/unit/audio-soniox-provider.test.ts                  # 8 pass, untouched
# plus every audio/transcription test file in tests/unit — 24 files, 178 tests, 0 failures
npx eslint open-sse/handlers/audioTranscription.ts \
  tests/unit/soniox-diarization-and-client-params.test.ts                              # clean
```

**Verified end-to-end against a live Soniox account**, on a real 58-second two-party Persian phone call, running this code path with real network calls:

```
no options          → response keys ["text"]                    ← unchanged
diarization + context → ["task","language","duration","text","segments"]
                        13 segments, 2 distinct speakers, duration 57.99s

  1.41s →  1.83s  S1: سلام.
  1.95s →  3.93s  S2: سلام آقای احمدزاده، روزتون بخیر.
  4.35s →  5.19s  S1: سلام، مرسی.
```

Turn boundaries line up with the dialogue. `context` also demonstrably works: a company name the model otherwise returns wrong comes out correct once the term is supplied.

Confirmed before writing any code that this was not a plan limitation — calling `api.soniox.com` directly returned `speaker` on every token with two distinct labels, so the gap was entirely gateway-side.

## Tests Added Or Updated

- `tests/unit/soniox-diarization-and-client-params.test.ts` (new, 9 cases): no options ⇒ legacy three-key job body **and** a bare `{ text }`; diarization reaches the job and speaker attribution reaches the client; segment grouping and second-resolution timings; all three diarization spellings, and `diarization=false` changing nothing; `context` passing verbatim **without** changing the response shape; `language` becoming a hint; `verbose_json` alone yielding segments; word granularity; and tokens with no speaker still producing segments with no `speaker` key.
- `tests/unit/audio-soniox-provider.test.ts` — deliberately **not** modified. It asserts the exact job body with strict deep equality, and it passing unchanged is the regression proof for this PR.

## Coverage Notes

- `open-sse/handlers/audioTranscription.ts`: the two new helpers (`readSonioxOptions`, `groupSonioxTokensBySpeaker`) and both branches of the response path are covered by the nine cases above, including the negative controls (`diarization=false`, tokens without speaker, context-only).
- No other provider path is touched, so the rest of the audio suite is a regression check rather than new coverage — all 24 files rerun green.

## Reviewer Notes

- **Why diarization alone upgrades the response**, rather than requiring `verbose_json` too: a caller who asks who-spoke-when and receives `{ text }` has no way to tell the flag was honoured — which is the exact failure mode being fixed. Opting in via either signal keeps the default safe while making the opt-in observable.
- **Segment grouping** is a single pass over the token stream: Soniox labels every token, so a turn boundary is simply where the label changes. Timings are converted to seconds to match the OpenAI `verbose_json` convention.
- **Scope**: `handleAssemblyAITranscription` (`:229`) and `handleGladiaTranscription` (`:297`) have the same missing-`formData` signature, and AssemblyAI supports diarization at the provider level. Left out to keep this diff reviewable; happy to follow up with the same treatment once the response contract here is agreed.
- Reported with a precise root-cause analysis by Vigen, who tested four real Persian phone recordings against a gateway and identified all three of the code sites above before filing.

### DIFF
--- a/open-sse/handlers/audioTranscription.ts
+++ b/open-sse/handlers/audioTranscription.ts
@@ -352,11 +352,94 @@ async function handleGladiaTranscription(providerConfig, file, modelId, token) {
   return errorResponse(504, "Gladia transcription timed out after 120s");
 }
 
+type SonioxToken = {
+  text?: string;
+  start_ms?: number;
+  end_ms?: number;
+  speaker?: string | number | null;
+  language?: string | null;
+};
+
+type SonioxOptions = {
+  diarize: boolean;
+  context: string;
+  language: string;
+  verbose: boolean;
+  wantWords: boolean;
+};
+
+/**
+ * Client-settable Soniox job options, read off the multipart form.
+ *
+ * Diarization is accepted under three spellings because callers reach for
+ * whichever their previous provider used: Soniox's own
+ * `enable_speaker_diarization`, the generic `diarization`, and Deepgram's
+ * `speaker_labels`. Anything absent stays absent from the job body so a
+ * caller who sends nothing produces byte-identical requests to before.
+ */
+function readSonioxOptions(formData?: FormData): SonioxOptions {
+  const str = (key: string): string => {
+    const value = formData?.get(key);
+    return typeof value === "string" ? value.trim() : "";
+  };
+  const flag = (...keys: string[]): boolean =>
+    keys.some((key) => /^(1|true|yes|on)$/i.test(str(key)));
+
+  const granularities = (formData?.getAll?.("timestamp_granularities[]") ?? []).map((value) =>
+    String(value).toLowerCase()
+  );
+  const diarize = flag("enable_speaker_diarization", "diarization", "speaker_labels");
+  const responseFormat = str("response_format").toLowerCase();
+
+  return {
+    diarize,
+    context: str("context"),
+    language: str("language"),
+    // Diarization implies the richer body: a caller who asked who-spoke-when and
+    // got `{text}` back has no way to tell the flag was honoured.
+    verbose: diarize || responseFormat === "verbose_json",
+    wantWords: granularities.includes("word"),
+  };
+}
+
+/**
+ * Collapse a token stream into contiguous single-speaker runs. Soniox labels
+ * every token, so a turn boundary is simply the point where the label changes.
+ */
+function groupSonioxTokensBySpeaker(tokens: SonioxToken[]) {
+  const segments: { speaker: string | null; startMs: number; endMs: number; text: string }[] = [];
+
+  for (const token of tokens) {
+    const speaker = token.speaker == null ? null : String(token.speaker);
+    const previous = segments[segments.length - 1];
+    if (!previous || previous.speaker !== speaker) {
+      segments.push({
+        speaker,
+        startMs: token.start_ms ?? 0,
+        endMs: token.end_ms ?? token.start_ms ?? 0,
+        text: token.text ?? "",
+      });
+      continue;
+    }
+    previous.endMs = token.end_ms ?? previous.endMs;
+    previous.text += token.text ?? "";
+  }
+
+  return segments;
+}
+
 /**
  * Handle Soniox transcription (async: upload file → create job → poll → get transcript)
  */
-async function handleSonioxTranscription(providerConfig, file, modelId, token) {
+async function handleSonioxTranscription(
+  providerConfig,
+  file,
+  modelId,
+  token,
+  formData?: FormData
+) {
   const authHeaders = buildAuthHeaders(providerConfig, token);
+  const options = readSonioxOptions(formData);
 
   const { body: uploadBody, contentType: uploadContentType } = await buildMultipartBody(file, {});
   const uploadRes = await fetch("https://api.soniox.com/v1/files", {
@@ -369,14 +452,21 @@ async function handleSonioxTranscription(providerConfig, file, modelId, token) {
   }
   const fileId = (await uploadRes.json()).id;
 
+  // Only keys the caller actually asked for are added, so a request with no
+  // options produces exactly the body this handler has always sent.
+  const jobBody: Record<string, unknown> = {
+    model: modelId,
+    file_id: fileId,
+    enable_language_identification: true,
+  };
+  if (options.diarize) jobBody.enable_speaker_diarization = true;
+  if (options.context) jobBody.context = options.context;
+  if (options.language) jobBody.language_hints = [options.language];
+
   const createRes = await fetch(providerConfig.baseUrl, {
     method: "POST",
     headers: { ...authHeaders, "Content-Type": "application/json" },
-    body: JSON.stringify({
-      model: modelId,
-      file_id: fileId,
-      enable_language_identification: true,
-    }),
+    body: JSON.stringify(jobBody),
   });
   if (!createRes.ok) {
     return upstreamErrorResponse(createRes, await createRes.text());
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Excellent fix — clear root-cause analysis, and the regression proof is exactly right:
`audio-soniox-provider.test.ts`'s strict deep-equality assertion is deliberately left
untouched and still passes, which proves the "no options" path is byte-identical while the
9 new diarization/context/language cases all pass too (17/17 total, verified locally).
Appreciate the scope discipline of not also touching AssemblyAI/Gladia in the same diff.
Ready to merge as-is.

47vigen:
+1

diegosouzapw:
Thanks @amirrezakm — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #12951, opened 2026-09-07
fix(codex): lift nested child cooldowns on parent clear and on snapshot headroom

### LINKED ISSUE #12817: [BUG] Codex provider stays in "cooling down" state after quota is restored — cooldown persists in provider_specific_data JSON field
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 22.04 (Docker host)

### Node.js Version

_No response_

### Provider(s) Involved

codex (GitHub Copilot / GPT combo)

### Model(s) Involved

gpt-5.6-sol-medium, gpt-5.6-terra-medium, gpt-5.6-luna-medium

### Client Tool

VS Code (GitHub Copilot Chat)

### Description

After GitHub Copilot quota was exhausted (weekly limit hit), OmniRoute correctly marked the codex provider as rate-limited. Once the quota was restored via a voucher/coupon code, all requests to GPT combo models (gpt-5.6-*) continued to return 503 with the message:

```
codex | all 1 active accounts cooling down for model gpt-5.6-luna-medium (reset after 38h 42m 37s) | lastErrorCode=null, lastError=undefined
```

The 38-hour countdown matched the **original weekly reset timestamp** (2026-09-07T08:49:18Z), not the actual current quota state. The issue persisted across multiple container restarts and DB cleanups.

**Root cause**: The per-scope cooldown timestamp is stored inside the `provider_specific_data` JSON column of the `provider_connections` table, under the key `codexScopeRateLimitedUntil.codex`. This nested JSON field is **not cleared** by the normal error-reset flow (clearing `rate_limited_until`, `last_error_at`, etc. on the top-level columns has no effect on it).

Source: `open-sse/services/codexAccount/state.ts` → `getLegacyCooldownMap()` reads `connection.providerSpecificData.codexScopeRateLimitedUntil`, and `auth.ts` line ~1633 uses `getCodexChildCooldown()` to derive `codexScopeCooldownMs` — which is what triggers the "cooling down" 503.

### Steps to Reproduce

1. Use a codex provider (GitHub Copilot) connection in OmniRoute.
2. Hit the weekly quota limit — OmniRoute logs the error and sets a cooldown (e.g. `codexScopeRateLimitedUntil.codex = "2026-09-07T08:49:18.000Z"`).
3. Restore quota via voucher/coupon in the GitHub Copilot dashboard.
4. Restart the OmniRoute container.
5. Send a request to any GPT combo model (gpt-5.6-*).
6. OmniRoute returns 503 with "cooling down for model … reset after 38h …" even though quota is restored.

Clearing the standard DB fields (`rate_limited_until`, `last_error_at`, `backoff_level` in `provider_connections`; deleting all `quota_snapshots` and `provider_quota_reset_events` records) does **not** resolve the issue because the cooldown is stored in a **nested JSON field** inside `provider_specific_data`.

### Expected Behavior

After quota is restored and the container is restarted, OmniRoute should recognise that the cooldown is no longer valid and allow requests to pass through to the codex provider. Either:
- The "connection reset" / error-clear flow should also clear `codexScopeRateLimitedUntil` inside `provider_specific_data`, or
- On startup, OmniRoute should verify the stored cooldown against the live quota state before honouring it.

### Actual Behavior

All requests to GPT combo models return HTTP 503 with body:
```
codex | all 1 active accounts cooling down for model gpt-5.6-luna-medium (reset after 38h 42m 37s) | lastErrorCode=null, lastError=undefined
```

The countdown is counting down from the **original** weekly reset timestamp that was recorded when quota was first exhausted. It ignores the fact that quota has been restored.

Workaround (manual): directly update the `provider_specific_data` JSON in SQLite to clear the stale cooldown fields, then restart:
```js
// Run inside container: node -e "..."
const db = new Database('/app/data/storage.sqlite');
const row = db.prepare("SELECT provider_specific_data FROM provider_connections WHERE id = '<connection_id>'").get();
const data = JSON.parse(row.provider_specific_data);
data.codexScopeRateLimitedUntil = {};
data.codexScopeRateLimitSource = {};
data.codexExhaustedWindowByScope = {};
data.codexQuotaStateByScope = {};
db.prepare("UPDATE provider_connections SET provider_specific_data = ? WHERE id = '<connection_id>'").run(JSON.stringify(data));
db.close();
```

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

- Single codex account connected (1 connection, 1 parent + 2 virtual children: codex/spark scopes)
- The "cooling down" log shows `lastErrorCode=null, lastError=undefined` — proving that the top-level connection error fields are clear, but the nested JSON field in `provider_specific_data` still holds the stale timestamp.
- Affected code path: `auth.ts` → `getCodexChildCooldown(connection, requestedModel)` → `codexAccount/state.ts` → `getLegacyCooldownMap()` reads `connection.providerSpecificData.codexScopeRateLimitedUntil`
- The fix should be in the connection reset / quota recovery flow to also clear `codexScopeRateLimitedUntil` (and related keys) from `provider_specific_data`.

### Validation Plan

_No response_

### LINKED ISSUE #12860: fix(resilience): codex: quota-preflight scope cooldown is never lifted when later quota snapshots show headroom, and cannot be cleared
## Summary

A Codex scope cooldown written by the quota preflight (`markQuotaPreflightAccountUnavailable` → `persistCodexChildCooldown` → `provider_specific_data.codexScopeRateLimitedUntil`, source `fallback`) is never lifted when a later quota snapshot shows the window has headroom again, and there is no way to clear it short of editing the database. The connection keeps being skipped until the originally recorded `next_reset_at`, while the dashboard shows it as `active` with a full quota bar and the log repeats:

```
codex | all N active accounts cooling down for model <model> (reset after …) | lastErrorCode=null, lastError=undefined
```

## Reproduction

1. Codex OAuth connection with the quota preflight enabled and a `session` threshold (e.g. 5 %).
2. Let the window's remaining percentage drop to the threshold. The preflight blocks the request and parks the scope until the `next_reset_at` the upstream reported at that moment. No `call_logs` row and no 429 — the block happens before the upstream call.
3. Let the upstream window reset earlier than predicted (observed: ~32 h early; every subsequent `quota_snapshots` row shows 100 % remaining and a new `next_reset_at`).
4. The scope stays parked until the stale `next_reset_at`. Connection row: `is_active = 1`, `test_status = active`, `rate_limited_until = null`, `backoff_level = 0`, `last_error = null`. A successful connection test and a process restart do not clear it.

## Why it sticks

- `src/sse/services/auth.ts`: `getCodexChildCooldown()` is consulted before the quota preflight, so fresh quota data never gets a chance to override the parked scope.
- `src/lib/quota/connectionRecovery.ts`: the recovery tick only handles `test_status = unavailable` + `rate_limited_until`; the Codex scope map in `provider_specific_data` is outside its scope.
- No code path deletes a key from `codexScopeRateLimitedUntil` (only writers exist: `updateCodexScopedQuotaState` / `updateCodexScopeCooldown`), and no dashboard action or API route resets a connection's cooldowns.
- The dashboard reads `test_status` / `rate_limited_until` / the quota bar, none of which reflect the scope cooldown, so the UI says "healthy" while routing says "skip".

## Expected

In order of preference:

1. Re-validate parked scopes against fresh quota data: when a new snapshot shows remaining above the threshold (with the preflight's epsilon), clear a `fallback`-sourced scope cooldown (a 429-derived `quota_reset` one should keep its authority). The recovery tick or the quota refresh itself is the natural place.
2. Or treat a `fallback`-sourced scope cooldown as advisory and let a live `proceed` from `preflightQuota()` win.
3. Surface the scope cooldown (until / source) on the connection row and give the `test` route, or a dedicated action, a way to drop it.

## Environment

OmniRoute v3.8.51, `codex` provider, single OAuth connection, quota preflight enabled.

HouMinXi:
## Bug Description

Two ways a Codex account stays skipped after its quota is actually back.

**#12817** — Clearing a Codex cooldown from the dashboard (or the CAS recovery path) only nulls `rate_limited_until`. Dispatch still reads `providerSpecificData.codexScopeRateLimitedUntil`, so the account stays skipped.

**#12860** — Quota preflight parks a cooldown at the reported `next_reset_at`. When a later snapshot shows the window recovered early, nothing removes that timestamp, so the account keeps getting skipped until the original deadline passes.

Fixes #12817
Fixes #12860

## Root Cause

Codex writes per-scope timestamps into JSON (`codex` / `spark`) without always mirroring them onto the parent column. Every "clear cooldown" path only touched that column, and no path ran in the other direction: a fresh snapshot proving headroom had no way to retire a cooldown the preflight had written.

## Fix

Parent column → nested maps (#12817):

- PUT `rateLimitedUntil: null` or `""` also drops the nested Codex maps
- `clearConnectionRateLimit` nulls the parent column and the nested maps in one transaction
- CAS recovery (`clearConnectionErrorIfUnchanged`) uses `json_remove` in the same UPDATE
- A 200 quota observation drops that child's leftover only; Spark stays

Snapshot → nested maps (#12860):

- `liftCodexScopeCooldownOnHeadroom(id, scope)` retires a scope cooldown when a snapshot proves headroom, inside one transaction
- `saveQuotaSnapshot` calls it after persisting a Codex row with `remaining_percentage > 0` and `is_exhausted !== 1`
- Only `fallback`-sourced cooldowns are eligible. A cooldown sourced from an upstream 429 `quota_reset` is authoritative and is held until it expires, with a 30s grace window so local clock skew cannot release it early
- Every window in the scope must report headroom: one exhausted window still justifies the cooldown even when a sibling reports full
- `hasCodexScopeCooldown` short-circuits before the scope-wide snapshot read, so the common (clean) connection pays one indexed lookup and no backup

Non-Codex connections are unchanged. Crash recovery still keeps future nested timestamps (they are live quota windows).

## How to Verify

Parent column path:

1. Persist Codex + Spark child cooldowns on one connection.
2. PUT `{ "rateLimitedUntil": null }` (or click Clear cooldown).
3. `getCodexChildCooldown` for both models returns null; `unrelated` JSON is still there.
4. A 200 quota observation for `gpt-5.5` clears only the Codex child; Spark remains limited.

Snapshot path:

5. Park a `fallback` cooldown on the `codex` scope with a reset an hour out.
6. Save a snapshot for that connection with `remaining_percentage: 100`, `is_exhausted: 0`.
7. The `codex` entry is gone; a Spark cooldown on the same connection is untouched.
8. Repeat with the cooldown sourced from `quota_reset` — it survives.

## Test Plan

- [x] `tests/unit/codex-scope-cooldown-clear-12817.test.ts` — 14 tests covering both issues (PUT null, PUT empty string, `clearConnectionRateLimit`, CAS, 200 observation, non-Codex, snapshot lift, `quota_reset` protection, skew grace boundary on both sides, partial-window hold, Spark isolation, short-circuit probe)
- [x] Existing `codex-account-cooldown-write`, `codex-account-helpers`, `quota-connection-recovery` still pass (40/40 combined)
- [x] Injection, each behavior separately: drop the PUT strip → red; drop the snapshot call → red; remove the skew grace → red; make the short-circuit probe ignore scope → red. Restore → green.
- [x] `typecheck:core`, `check:cycles` (427 files), `check-docs-sync` clean

## Risk Assessment

Low. Nested maps are Codex-only, and the snapshot lift is gated on provider `codex` plus a `fallback` source. Spark sibling state is preserved on a Codex-only 200 and on a Codex-only snapshot. The write runs in one transaction that re-reads the row under lock, so a concurrent cooldown write cannot be lost. `providers.ts` shrank 1186 → 1137 (CAS moved to `rateLimit.ts`).


xiaoyaner0201:
Thanks for addressing both the manual-clear and restored-quota paths. I added an independent affected-instance observation to #12860: fresh weekly usage showed 100% remaining while a legacy `fallback` Codex child cooldown kept the account excluded; removing only the stale child-map entries restored real traffic without a restart.

Could the regression coverage also exercise the **reset-credit button's actual service path**, rather than only the generic parent clear and direct snapshot writes?

At [`consumeCodexResetCredit` on the current release](https://github.com/diegosouzapw/OmniRoute/blob/949235736042b13cf64215632e6d44db7985af76/src/lib/usage/codexResetCredits.ts#L506-L513), a recognized consume outcome invalidates the quota cache and calls `fetchAndPersistProviderLimits`. It does not directly invoke the parent clear path. The snapshot reconciliation in this patch looks relevant, but I have not run this PR and cannot claim that the whole button path passes.

A useful regression would seed a future legacy `fallback` child cooldown, mock a successful reset redemption plus fresh healthy weekly usage, invoke `consumeCodexResetCredit`, then assert the persisted Codex child is eligible without another manual clear or restart. It should preserve an unrelated Spark cooldown and retain protection on failed redemption, incomplete refresh, or a newer concurrent cooldown. An `alreadyRedeemed` retry should likewise rely on verified fresh quota evidence rather than unconditionally clearing current limits.

That would directly cover the operator-visible symptom: “reset succeeded and the quota bar refilled, but requests still cannot use the account.” Related alternative reconciliation work: #13073.

HouMinXi:
Thanks — that was the operator-visible hole.

`consumeCodexResetCredit` already invalidated the quota cache and called `fetchAndPersistProviderLimits` after a recognized redeem. That live-fetch writes `setQuotaCache`, which persists snapshots and is the same headroom lift as a dashboard usage refresh. The missing piece was a regression that actually invoked the button function instead of writing a snapshot by hand.

Added four cases on that path:

- leftover `fallback` Codex child + successful redeem + healthy weekly usage → Codex child eligible, Spark child still parked
- `alreadyRedeemed` without headroom (`used_percent: 100`) → nested maps stay
- failed consume (HTTP 409 `noCredit`) → nested maps stay
- incomplete usage refresh (usage endpoint 500; `getCodexUsage` swallows that into `{message}` and still returns from consume) → nested maps stay

No extra production branch: the button already rode the snapshot path. Coverage now pins it so a filled quota bar after reset cannot leave the leftover fallback child parked.

Pushed on this PR (`af2ed0194`). Related: #12860, #13073.

PR #12952, opened 2026-09-07
feat(api): accept expiresAt when creating API keys

caniko:
## Summary

`POST /api/keys` accepts `expiresAt` (ISO datetime, nullable) with the same semantics as the key-update path (`updateKeyPermissionsSchema`), so automation can create an expiring key in one operation instead of create-then-update. Omitted/null preserves the current non-expiring behavior. Enforcement reuses the existing expiry policy (`validateKeyStatus` / `validateApiKey`) without changes.

This is creation/update parity only: it does not introduce expiry enforcement (already shipped), TTL/renewal services, or UI changes.

## Changes

- `src/shared/validation/schemas/keys.ts`: `createKeySchema` gains `expiresAt: z.string().datetime().nullable().optional()`, identical to the update schema.
- `src/lib/db/apiKeys.ts`: `CreateApiKeyOptions.expiresAt`, persisted via `INSERT ... expires_at` in the original creation (no POST-then-PATCH window), echoed on the returned object.
- `src/app/api/keys/route.ts`: passes validated `expiresAt` through and echoes it in the 201 response.
- `tests/unit/api-key-create-expiry.test.ts`: 7 regression tests (schema parity incl. malformed rejection, persistence/readback, past-expiry rejection matching update behavior, omitted/null non-expiring, update interop, route 201+echo, route 400 without creating).
- `changelog.d/features/0000-api-key-create-expires-at.md`: entry (filename uses 0000 placeholder; happy to rename to the PR number).

## Validation

- New tests failed before the fix (5 fail / 2 pass) and pass after (7/7).
- Neighbors green: api-key-lifecycle (11), api-keys-create-no-hang-6570, api-key-scope-validation, api-key-reveal-route (26 total).
- `prettier --check` on touched files: clean. `eslint` with repo suppressions: clean. `typecheck:core`: exit 0.
- Full `test:unit` (37,326 tests): 117 failures, all reproduced on the clean base checkout (environmental: e.g. api-key-self-service, cli-keys-command, migration-147/149, chatCore, CLI suites fail identically without this change).
- `test:scoped` fails safe by design (new test file unmapped; apiKeys.ts is a hub) and directs to the full suite per its own message.

## Limitations

- No dashboard changes; no auto-renewal/rotation service.
- Changelog filename placeholder (0000) pending PR number.


diegosouzapw:
Clean, minimal PR — nice parity fix. Confirmed directly on the release tip that
`createKeySchema` really is missing `expiresAt` while the update schema already has the exact
same field, and that expiry enforcement in `apiKeys.ts` is untouched by this change (pure
create/update parity, as described). Ran your test file on this branch: 7/7 passing. Looks
merge-ready to me — nothing else needed.

diegosouzapw:
Thanks @caniko — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

PR #12957, opened 2026-09-07
fix(resilience): lock the exact model, not the quota family, on 5xx model-lockout failures

### LINKED ISSUE #12955: fix(resilience): Model lockout on a 5xx locks the whole codex quota scope (every `gpt-5*` model), not the failing model
## Summary

With Model Lockout enabled (default `errorCodes` include `502/503/504`), a single `5xx` on one codex model locks **every** `gpt-5*` model of that connection, because `recordModelLockoutFailure()` writes the lock under the codex quota-scope key (`getCodexModelScope()` → `codex` for every `gpt-5*` id) for every status except `404`. The lock escalates with the usual exponential backoff (2 min → 30 min) and is not tied to quota at all.

The trigger does not even have to come from upstream: a stream that ends without a content block is turned into a synthesized `502` by quality validation (`Claude returned an empty response (no content block)`), and `combo.ts` records that as a `quality_failure` lockout. Transport failures (`terminated`, `EHOSTUNREACH`, connect timeout) take the same path. Any of them removes the whole codex connection from every combo that routes to it for the lockout window, while its quota headroom is untouched and the dashboard shows the connection as healthy.

## Steps to reproduce

1. Settings → Model Lockout: enabled, default `errorCodes` (`[403, 404, 429, 502, 503, 504]`).
2. A codex connection used by two combos: combo X → `codex/gpt-5.6-sol`, combo Y → `codex/gpt-5.6-luna` (any two `gpt-5*` models).
3. Make one request to combo Y fail with a `502` (an empty stream is enough).
4. Send a request to combo X.

**Expected:** combo X still dispatches to `gpt-5.6-sol`; only `gpt-5.6-luna` is locked.

**Actual:** `isModelLocked("codex", <conn>, "gpt-5.6-sol")` is `true`; combo X logs `Skipping codex/gpt-5.6-sol — model locked by resilience (cooldown active)` and, if it has no other provider, answers `503` / `404` for 2–30 min.

## Where

`open-sse/services/accountFallback.ts`: `getModelLockKey()` scopes codex locks to the quota family for every reason except `not_found`; `recordModelLockoutFailure()` uses that key unless the caller passes `scope: "exact"` (only Antigravity does). The family scope is right for `429`/`403` — the codex quota really is per scope — but a `5xx` is evidence about one model endpoint at that moment, not about the account's quota.

## Proposed fix

Resolve the lockout scope from the status: `429`/`403`/`402` keep the quota family, `404` stays on the bare model, every other status locks the exact provider/connection/model tuple (the `exact` key that already exists for Antigravity). An explicit `scope` option still wins. Two consequences need the same change: success-decay (`decayModelFailureCount`) has to walk the exact key too, and the Model Cooldowns card has to list an exact-scope lock by its bare model so it can be cleared by hand.

PR: #12957

insoln:
Fixes #12955.

## Problem

`recordModelLockoutFailure()` writes every lockout except `404` under the quota-family key. For codex that key is the whole quota scope (`getCodexModelScope()` → `codex` for every `gpt-5*` id), so one `5xx` on one codex model — an upstream server error, a transport failure (`terminated`, `EHOSTUNREACH`, connect timeout), or the `502` that quality validation synthesizes for a stream without a content block — locks every `gpt-5*` model of the connection for 2–30 min with exponential escalation. The family scope is right for `429`/`403`, where the codex quota really is per scope; a `5xx` says something about one model endpoint at that moment, not about the account's quota.

## Fix

- `open-sse/services/accountFallback/exactModelLock.ts`: `resolveLockoutScope(status, explicit)` — `429`/`403`/`402` (and `404`, which `getModelLockKey()` already narrows to the bare model) keep the quota-family key; any other status uses the exact provider/connection/model key that already exists for Antigravity. A caller's explicit `scope` still wins.
- `recordModelLockoutFailure()` resolves the scope once for both the key and the lock function.
- `decayModelFailureCount()` now walks every key shape (family, not-found, exact) so success-decay reaches an exact-scope lock; `null` model stays a no-op. The loop lives in the leaf (`decayFailureCounts()`).
- `getAllModelLockouts()` parses the `exact:` marker out of the key (`parseModelLockKey()`), so the Model Cooldowns card lists the bare model and `DELETE /api/resilience/model-cooldowns` can clear it by that name.

Behaviour for `429`/`403`/`402`/`404` is unchanged; `lockModel()` direct callers are unchanged.

## Tests

`tests/unit/model-lockout-5xx-exact-scope.test.ts` (new, registered in `stryker.conf.json`): scope table; codex `502` on one model leaves the sibling `gpt-5*` models (and the alias spelling) routable; codex `429` still locks the whole scope; explicit `scope` wins; the exact lock escalates per failure and decays on success (`{cleared,newFailureCount}` contract kept); the dashboard listing shows the bare model and `clearModelLock()` clears it by that name; end-to-end through `handleComboChat` with Model Lockout enabled — a `502` on one codex model, then a combo on a sibling model still dispatches.

Green locally: `check:file-size` (PR mode), `check:mutation-test-coverage --strict`, `check:cycles`, `typecheck:core`, `eslint` on the changed files, and the neighbouring suites `model-lockout-decay`, `model-lockout-max-cooldown`, `combo-selected-connection-success`, `account-fallback-service`, `combo-lockout-quota-reset-6863`, `native-codex-turn-pin-model-scoped-fallback`, `combo-provider-cooldown-sibling`, `antigravity-family-cooldown*`.

Docs: `docs/architecture/RESILIENCE_GUIDE.md` §3 (key scope by status). Changelog fragment included.

⚠️ base-red inherited: #12732 (`glm.ts` TS2554, stale "169 migrations" doc counts, `reset-aware-model-family.md` fragment, `reset-aware-request-scope-12600` missing from `stryker.conf.json`, and the unrelated unit-test failures listed on #12830).


diegosouzapw:
Solid fix with a concrete, well-documented production repro (a single 502
on one `gpt-5.6-*` model taking down every sibling `gpt-5*` model of the
connection for up to 30 minutes). The `resolveLockoutScope(status, explicit)`
split — quota/entitlement statuses (402/403/404/429) keep the family scope,
everything else (5xx) locks the exact tuple — is the right default, and the
explicit-override escape hatch is preserved. Probed the new test file
locally: 7/7 pass, including the `handleComboChat` integration test with a
real temp DB. One note: the PR's CI checks currently show several
failures, but the branch hasn't been updated in 8 days and a 59-PR batch
plus several base-red sweeps have landed since — running the new test file
in isolation against the current tip passes cleanly, so this looks like a
stale CI snapshot rather than a real regression. Worth re-triggering CI
(rebase or empty push) before merge to confirm. Also flagging for context:
this is complementary to #12340 in the same batch (different layer of the
model-lockout system, no file-line overlap) — good to merge either order.

insoln:
Rebased onto the current `release/v3.8.51` tip (`af2002a`) and force-updated the branch to `0ba96fffc1`. The patch is byte-for-byte equivalent to the original two commits (`git range-diff`: both `=`), and the PR remains mergeable.

Review and focused validation after the rebase:

- dedicated regression suite: **7/7 pass** (`model-lockout-5xx-exact-scope.test.ts`)
- adjacent provider-cooldown suite: **9/9 pass**
- Vitest: **473/473 pass**
- `typecheck:core`, PR-mode file-size, complexity ratchets, cycles, docs-all, Prettier, and pack policy: **pass**
- independent code review: **no findings**

The latest CI run is still red only on failures outside this PR's six-file diff:

- API Route Typecheck: `src/lib/providerModels/modelDiscovery.ts` TS2345
- mutation coverage drift: `comboPredicates.ts`, `responseHeaders.ts`, and `semanticCache.ts` tests absent from `stryker.conf.json`
- unit shards: authz pipeline, xhigh/reasoning-effort tests, raw NUL source guard, stale budget-card SVG, provider prefix count, translation-message inventory, and forced-pin #12080

This PR changes only the two account-fallback modules, its dedicated test, one Stryker entry, the resilience guide, and a changelog fragment. Its added Stryker entry is accepted; the three missing entries are for unrelated modules introduced on the base.

PR #12953, opened 2026-09-07
feat(api): wire bifrost sidecar fast-path into /v1/chat/completions and /v1/messages with native fallback

rqzbeh:
### Summary
This PR connects the supervised Go **Bifrost sidecar** (`@maximhq/bifrost`) directly into OmniRoute's primary high-traffic ingress endpoints (`POST /v1/chat/completions` and `POST /v1/messages`) based on `release/v3.8.51`.

- **Zero-Copy Fast-Path:** When Bifrost is active and healthy on loopback (`127.0.0.1:8080`), eligible requests bypass the heavy V8 handler stack and stream tokens directly through Go's `net/http` connection pool.
- **Graceful Auto-Fallback:** If Bifrost is offline, times out, or returns 5xx errors, requests automatically trip failure cooldowns and transparently fall through to native TypeScript `handleChat()` with `X-Routing-Fallback` headers attached.
- **Shared Subsystem:** Consolidates Bifrost routing, circuit breaker cooldowns, and client dispatching into `src/shared/services/bifrost/`, deduplicating logic across the relay proxy and primary routes.
- **Test Suite:** Added comprehensive unit and integration test suites covering unary dispatch, SSE streaming finalization, failure cooldowns, and fallback mechanics.

### Verification
- `npm run typecheck:core`: Passed (0 errors).
- `npm run check:route-guard-membership`: Passed.
- `npm run check:deps`: Passed (0 new dependencies).
- `node --import tsx/esm --test tests/unit/services/bifrost/*.test.ts tests/unit/api/v1/*bifrost*.test.ts`: 22/22 tests passed (100% green).

⚠️ base-red inherited: #13866

PR #12962, opened 2026-09-07
feat(services): add sing-box supervisor installer and tproxy integration

rqzbeh:
### Summary
This PR adds support for **sing-box** as an embedded supervised service under `ServiceSupervisor` (`src/lib/services/installers/singbox.ts`) and integrates it into the `src/mitm/tproxy/` network setup pipeline targeting `release/v3.8.51`.

- **Embedded Supervisor Adapter:** `src/lib/services/installers/singbox.ts` installs, configures, and manages sing-box on loopback port `20140`.
- **TPROXY Integration:** `src/mitm/tproxy/setup.ts` provides `ensureSingboxTproxy()` for dynamic transparent proxying without requiring native C compilation / node-gyp build steps.
- **Service Registration:** Registered in `src/lib/services/bootstrap.ts` under `SERVICES[]`.
- **Unit Tests:** Added `tests/unit/services/installers/singbox.test.ts` verifying lifecycle, config generation, and spawn argument formatting (100% passing).

⚠️ base-red inherited: #13866


diegosouzapw:
Same structural issue as your llmlingua PR (#12967): `SINGBOX_PORT` is declared twice in
`bootstrap.ts`, which won't compile. More importantly, `install()` doesn't fetch the real
sing-box binary — it writes `#!/bin/sh\necho 'sing-box mock binary'` as the "binary", and
`resolveSpawnArgs` runs that mock. Since the mock script exits immediately instead of opening
the TPROXY listener, `ensureSingboxTproxy()` silently fails
(`supervisor.start().catch(() => false)`) and the feature never actually intercepts anything —
with no visible error to the operator. Could you wire this to a real sing-box release download
(with a pinned version + checksum), and surface a clear error/log when the supervisor fails to
start instead of swallowing it? Also worth a clean rebase to drop the ~15 duplicated
security/CI commits shared with #12967 and #12953 that have since diverged from the release
tip.

rqzbeh:
Rebased onto `upstream/release/v3.8.51` as a single isolated commit, dropping all diverged commits from `main`.

Changes implemented:
- **Real binary download & checksum verification:** `src/lib/services/installers/singbox.ts` downloads the pinned release `1.14.1` directly from `SagerNet/sing-box` GitHub releases, verifies platform-specific SHA256 checksums (`SINGBOX_CHECKSUMS` covering `linux-amd64`, `linux-arm64`, `darwin-amd64`, `darwin-arm64`, and `windows-amd64`), and extracts the binary. Any unpinned version is rejected.
- **Supervisor error visibility:** `ensureSingboxTproxy()` in `src/mitm/tproxy/setup.ts` logs error diagnostics if the supervisor is unregistered, if `start()` rejects, or if the process fails to transition to `running`, eliminating the silent `.catch(() => false)` failure mode.
- **Deduplicated bootstrap configuration:** Removed duplicate `SINGBOX_PORT` in `src/lib/services/bootstrap.ts`, retaining a single declaration at port `20140` alongside existing services.
- **Migration & documentation:** Seeded using migration slot `182_singbox_service_seed.sql`, added the 8 `/api/services/singbox/*` management endpoints to `docs/openapi.yaml`, documented the sidecar in `docs/frameworks/EMBEDDED-SERVICES.md` and `docs/reference/ENVIRONMENT.md`, and added `SINGBOX_PORT` to `.env.example`.
- **Test coverage:** Updated unit tests in `tests/unit/services/installers/singbox.test.ts` to verify archive extraction, checksum validation, and download failure handling.

PR #12972, opened 2026-09-07
fix(sse): aggregate findInsensitive collision warning into one line per build

IAMBOBJIM:
## Problem

`findInsensitive()` in `src/lib/modelMetadataRegistry.ts` warns once **per colliding key** while building its lowercase-key index. On a real catalog that is hundreds of lines per rebuild.

Measured on a production log from a local deployment:

| metric | value |
| --- | --- |
| lines in the active `app.log` | 67,353 |
| lines that are this one warning | **27,296 (40%)** |
| burst rate | ~500/sec |
| `logs/` on disk | 466 MB across 9 rotated 50 MB files |

The rotations are driven almost entirely by this single message.

## What this does not change

The warning itself is worth keeping, and this PR keeps it. A case-insensitive collision is a genuine upstream data-quality signal — models.dev returning both `OpenAI` and `openai` as distinct provider keys — and first-match-wins silently discards the later value. Removing the warning would have been the smaller diff and the wrong fix.

Also unchanged: the index itself, first-match-wins resolution, and the `WeakMap` identity cache from #8697. This is purely a logging-volume change.

## What changed

Collisions are collected during the index build and reported as **one line per build**, carrying the total count plus the first 5 keys:

```
[modelMetadataRegistry] findInsensitive: 312 case-insensitive key collision(s) — keeping first-seen value, later ones discarded. Keys: openai, anthropic, qwen/qwen3-235b-a22b-thinking-2507, moonshotai/kimi-k2.5, google, …
```

Same diagnostic, 1/N the volume.

New `tests/unit/model-metadata-registry-collision-log.test.ts`, 4 tests. Verified failing-then-passing:

- **Before:** 3 collisions → 3 warnings; 50 collisions → 50 warnings
- **After:** always exactly 1

Coverage: the aggregated count, that the line still names colliding keys, that the count stays at 1 as collisions scale, and that zero collisions emit nothing.

> **Note for reviewers / future editors:** the fixture deliberately spells the provider key `OpenAI`, not `openai`. `findInsensitive` short-circuits on `if (key in obj) return obj[key]` before the index is ever built, so a fixture containing the literal lookup key produces **zero** warnings and passes vacuously after the fix. The first draft of this test hit exactly that and proved nothing.

## Gates

- `node --import tsx/esm --test tests/unit/model-metadata-registry-collision-log.test.ts` — 4/4 pass
- `eslint` on both changed files — clean
- Base-green: no open `base-red` issue for `release/v3.8.50` at time of writing; no active `release-freeze`
- `typecheck:core` — 9 errors, **all** in `open-sse/services/compression/omniglyph*`, none in the changed files. Not from this branch: those sources are byte-identical to `origin/release/v3.8.50`, and the cause is a local `node_modules` carrying `omniglyph` 1.3.1 against the required `^1.4.0`. CI typechecks against a correct install.

### DIFF
--- a/src/lib/modelMetadataRegistry.ts
+++ b/src/lib/modelMetadataRegistry.ts
@@ -303,26 +303,39 @@ export function getCanonicalModelMetadata(input: {
 // a rebuild instead of rebuilt per lookup.
 const lowercaseIndexCache = new WeakMap<object, Map<string, unknown>>();
 
+/** Colliding keys named in the aggregated collision warning before it truncates. */
+const COLLISION_SAMPLE_SIZE = 5;
+
 function findInsensitive<T>(obj: Record<string, T> | null | undefined, key: string): T | undefined {
   if (!obj || !key) return undefined;
   if (key in obj) return obj[key];
   let index = lowercaseIndexCache.get(obj);
   if (!index) {
     index = new Map();
+    const collisions: string[] = [];
     for (const [k, v] of Object.entries(obj)) {
       const lowerKey = k.toLowerCase();
-      // Warn once at index-build time (not per-lookup) if two keys collide
-      // case-insensitively — a real data-quality signal from an upstream sync (e.g.
-      // models.dev returning both "OpenAI" and "openai" as distinct provider keys).
-      // Matches the pre-fix scan's silent first-match-wins behavior, just surfaced
-      // instead of swallowed.
+      // Collisions are a real data-quality signal from an upstream sync (e.g.
+      // models.dev returning both "OpenAI" and "openai" as distinct provider
+      // keys), so they are surfaced rather than swallowed — first-match-wins,
+      // matching the pre-#8697 scan's behavior.
       if (index.has(lowerKey)) {
-        console.warn(
-          `[modelMetadataRegistry] findInsensitive: case-insensitive key collision on "${lowerKey}" — keeping first-seen value, later one discarded`
-        );
+        collisions.push(lowerKey);
         continue;
       }
       index.set(lowerKey, v);
+    }
+    // Aggregate into ONE line per index build. Warning per colliding key made
+    // the signal unreadable and expensive: a real catalog collides on hundreds
+    // of keys, and a production log carried 27,296 of these lines (40% of the
+    // file, ~500/sec bursts) driving 52 MB rotations. The count plus a bounded
+    // sample keeps the diagnostic without the flood.
+    if (collisions.length > 0) {
+      const sample = collisions.slice(0, COLLISION_SAMPLE_SIZE).join(", ");
+      const more = collisions.length > COLLISION_SAMPLE_SIZE ? ", …" : "";
+      console.warn(
+        `[modelMetadataRegistry] findInsensitive: ${collisions.length} case-insensitive key collision(s) — keeping first-seen value, later ones discarded. Keys: ${sample}${more}`
+      );
     }
     lowercaseIndexCache.set(obj, index);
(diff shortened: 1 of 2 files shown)


IAMBOBJIM:
## Local suite run — no regressions

GitHub Actions has not run on this PR (the workflow runs sit at `action_required`, awaiting maintainer approval), so I ran the suite locally and compared against the base commit directly.

### `npm run test:vitest` — A/B against base

Same worktree, same `node_modules`, same machine, run back to back. The only variable was `src/lib/modelMetadataRegistry.ts`, reverted to its base version via `git checkout origin/release/v3.8.50 -- <path>` and restored with `git checkout HEAD -- <path>` (no `git stash`, per Hard Rule #22a).

| | Test Files | Tests |
| --- | --- | --- |
| base `origin/release/v3.8.50` (`091589089`) | 1 failed \| 43 passed (44) | 1 failed \| 404 passed (405) |
| this branch (`ab0c23300`) | 1 failed \| 43 passed (44) | 1 failed \| 404 passed (405) |

Identical, down to the same single failing test:

```
tests/unit/autoCombo/provider-family-combos.test.ts
  > auto/<family> materialization (#6453)
  > resolves auto/glm to a virtual combo spanning every connected GLM backend
```

That failure is present on the base commit without this change. **This branch introduces no vitest regressions.**

Worth flagging separately: that `auto/glm` test appears red on `release/v3.8.50` in this environment, while the base-green tracker showed no open `base-red` issue at the time of writing. It may be environment-specific here rather than genuinely broken upstream.

### Focused tests

`tests/unit/model-metadata-registry-collision-log.test.ts` — 4/4 pass. Verified failing-then-passing: on the old implementation 3 collisions produced 3 warnings and 50 produced 50; after the change it is always exactly 1.

### Lint

`eslint` clean on both changed files.

### Known gaps in this evidence

Two things I could **not** verify locally, stated plainly rather than glossed:

1. **`npm run test:unit` is inconclusive.** It was killed by a 50-minute wall-clock timeout partway through, so it produced no summary. An earlier attempt that appeared to show many `auto-*` failures was run while the machine sat at 100% CPU; that run collected only 372 tests versus 405 on a quiet machine, so its failures were contention artifacts, not results. I am not claiming this suite passes.
2. **`typecheck:core` reports 9 errors**, all in `open-sse/services/compression/omniglyph*` — none in the changed files. Those sources are byte-identical to `origin/release/v3.8.50`; the cause is a local `node_modules` carrying `omniglyph` 1.3.1 against the required `^1.4.0`. CI will typecheck against a correct install.

Environment: Node v24.19.0, Windows, worktree cut from `origin/release/v3.8.50` @ `091589089`.

diegosouzapw:
Great diagnosis with real production numbers behind it (27,296 log lines / 40% of the file /
52 MB rotations from one-warning-per-collision) — aggregating to one line per index build with a
bounded sample keeps the diagnostic useful without the flood. Ran the new tests locally: 4/4
pass. This PR's base is `v3.8.50` — please retarget to `release/v3.8.51` and add a changelog
fragment before merge.

diegosouzapw:
Thanks @IAMBOBJIM — merging via the release merge-train. Validated in local merge-train /tmp/mt-train3b.log on 192.168.0.113 @ train tip 408e2128791696a18966681953136fc96aeb99b0 (32 PRs boarded): static gates green; full test:unit 40694 tests, 17 failing — every one reproduces on the pure release tip (base-red sweep list), zero new reds. Merged --admin per merge-gates §4/§7.

PR #12974, opened 2026-09-07
fix(vision): honor Custom Models vision flag on advertised aliases

### LINKED ISSUE #12758: Direct (non-combo) requests to a custom OpenAI-compatible model ignore supportsVision and force-route images to hardcoded glm/glm-4.6v
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Node.js Version

_No response_

### Provider(s) Involved

Custom OpenAI-compatible provider (self-hosted vLLM on RunPod); unrelated GLM Coding provider is the wrongly-selected fallback target

### Model(s) Involved

orcarouter/Qwen3.8-27B-Uncensored-NVFP4

### Client Tool

Direct API request (fetch/curl to /v1/chat/completions)

### Description

OmniRoute's vision-bridge logic hardcodes a fallback vision model (glm/glm-4.6v in this build) for any chat completion request that contains image content, regardless of whether the target model already declares vision support. For a custom OpenAI-compatible model added under "Custom Models" (with "Vision capable" explicitly checked), a request sent directly to that model (i.e. not through a Combo) is still silently redirected to the hardcoded fallback model instead of being sent to the configured model. In our case the fallback model isn't present in the connected GLM account's live catalog, so the request fails outright with a 400 instead of reaching the intended model.

This appears to be the same underlying issue as #8032 and #2484. The v3.8.50 changelog entry "fix(routing): consult customModels supportsVision flag in Combo routing (#9195)" seems to have scoped the fix to Combo routing only, leaving the direct/non-combo request path unpatched.

### Steps to Reproduce

1. In OmniRoute, add a custom OpenAI-compatible provider pointing at a self-hosted vLLM instance.
2. Under Custom Models, add a model (path-shaped id, e.g. `orcarouter/Qwen3.8-27B-Uncensored-NVFP4`), and check "Vision capable" in its edit form.
3. Send a chat completion request directly to that model id via /v1/chat/completions (no Combo involved), with a message containing an `image_url` content part, e.g.:
```json
{
  "model": "orcarouter/Qwen3.8-27B-Uncensored-NVFP4",
  "messages": [
    {
      "role": "user",
      "content": [
        { "type": "text", "text": "What color is this image? One word." },
        { "type": "image_url", "image_url": { "url": "data:image/png;base64,<1x1 png>" } }
      ]
    }
  ]
}
```
4. Observe the response.

### Expected Behavior

Since the target model is explicitly flagged `supportsVision: true` (via the "Vision capable" checkbox in Custom Models), the request should be sent to that model as-is, with no model substitution.

### Actual Behavior

The request fails with a 400 error: {"error":{"message":"Model 'glm-4.6v' is not available in the active live catalog for provider 'glm'.","type":"invalid_request_error","code":"bad_request"}}

OmniRoute is attempting to silently substitute `glm/glm-4.6v` for the requested model purely because the request contains image content, ignoring the requested model's own `supportsVision: true` flag. It only surfaces as a visible error because that specific GLM checkpoint isn't present in our connected GLM account's catalog — but even where it is present, this would mean the configured model is never actually used for vision requests sent outside of a Combo.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Status: 400
Response body:
{"error":{"message":"Model 'glm-4.6v' is not available in the active live catalog for provider 'glm'.","type":"invalid_request_error","code":"bad_request"}}

For reference, a plain text-only request (no image) to the same model id ("orcarouter/Qwen3.8-27B-Uncensored-NVFP4") completes normally and is served by the correct model - only requests containing image content trigger the substitution.
```

### Screenshots

_No response_

### Additional Context

- Single account, custom provider connection ("vLLM-Runpod"), no proxy involved.
- The custom model's "Vision capable" checkbox was confirmed checked via the model edit form before testing.
- As a workaround, we are wrapping the same model in a single-model, priority-strategy Combo instead of connecting directly, on the theory that Combo routing correctly consults the supportsVision flag per #9195 while the direct/non-combo path does not. Will update this issue once that's confirmed.
- `glm-4.6v` does not correspond to anything we configured; we do have a GLM Coding connection, but its live catalog does not include this checkpoint.

### Validation Plan

- Add a custom OpenAI-compatible provider + a Custom Model with a path-shaped id and "Vision capable" checked.
- Send a chat completion request directly (no Combo) to that model id with an image_url content part.
- Expect: the request reaches the configured model and returns a normal response, with no attempt to substitute any other model regardless of whether that other model is actually available.
- Regression check: the same request via a Combo containing this model should continue to work as it does today (per #9195).

HouMinXi:
## Bug Description

A Custom Models row with "Vision capable" checked still had its image
requests swapped to `glm/glm-4.6v` when the client sent the advertised
alias (`vllm/orcarouter/...`) or the bare path-shaped id. Only the
internal `providerId/modelPath` form skipped the bridge. Text chat
already resolved the same custom row; Vision Bridge used a narrower
lookup.

`parseModel` splits on the first `/`, so `orcarouter/Qwen...` becomes
provider `orcarouter`. The override lives under the openai-compatible
connection id. The matcher now recovers that row for all three id forms
without changing `parseModel`.

Fixes #12758

## Changes

- Lift custom-model vision lookup into `customVisionOverride.ts` (frozen
  `models.ts` cannot grow).
- Match stored `connectionId + path id` against advertised alias, bare
  path, and full internal id.
- Conflicting true/false hits for the same path return null (unknown),
  not a coin flip.

## Test plan

- `tests/unit/vision-bridge-custom-path-id-12758.test.ts`: three id
  forms skip the glm swap; explicit `supportsVision: false` still wins
  on the advertised alias.
- Regression: `api-models-hide-paid-6328`, `custom-vision-override-combo-routing-9195`.

### DIFF
--- a/changelog.d/fixes/12974-custom-vision-advertised-alias.md
+++ b/changelog.d/fixes/12974-custom-vision-advertised-alias.md
@@ -1,0 +1,1 @@
+- **fix(vision):** Custom Models with "Vision capable" checked no longer have image requests swapped to `glm/glm-4.6v` when the client sends the advertised alias (`vllm/path/...`) or the bare path-shaped id — Vision Bridge now matches the stored override for all three id forms ([#12758](https://github.com/diegosouzapw/OmniRoute/issues/12758))
--- a/src/lib/db/models.ts
+++ b/src/lib/db/models.ts
@@ -6,7 +6,6 @@
 
 import { isRetiredGitHubCopilotModelId } from "@omniroute/open-sse/config/providers/registry/github/retiredModels.ts";
 
-import type { SqliteAdapter } from "./adapters/types";
 import { getDbInstance } from "./core";
 import { getProviderConnectionsCount } from "./providers";
 import { type JsonRecord, getKeyValue } from "./models/shared";
@@ -53,6 +52,13 @@ export {
 } from "./models/aliases";
 export { getMitmAlias, setMitmAliasAll } from "./models/mitmAlias";
 export type { SyncedAvailableModel } from "./models/synced";
+export {
+  getCustomModelVisionOverride,
+  listCustomModelVisionOverrides,
+  type CustomModelVisionOverrideMap,
+  type CustomModelVisionDatabase,
+  type CustomModelVisionOverrideReadOptions,
+} from "./models/customVisionOverride";
 
 // ──────────────── Custom Models ────────────────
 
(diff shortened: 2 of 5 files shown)


HouMinXi:

PR #12982, opened 2026-09-07
fix(images): fall back when combo leg returns empty 2xx response

tiangao88:
## Summary

`executeImageCombo()` stops on the first leg that returns `success: true` and
hands its body straight to the client. OpenAI-compatible image providers
(`format: "openai"`, e.g. `openrouter/*`, used by combos such as `pmoc-image-text`)
can answer **HTTP 200 with an empty or malformed image payload** — `fetchImageEndpoint()`
normalized that to `success: true` via `data.data || []`, so the combo never tried the
next priority leg and the client received an image-less 200 (observed as an image-combo
failure with status 200; clients such as Hermes then reject the empty response).

## Why HTTP 200 was incorrectly treated as success

`fetchImageEndpoint()` (open-sse/handlers/imageGeneration.ts) only checked
`response.ok`. Any 2xx body was normalized to OpenAI shape:

```ts
data: { created: ..., data: data.data || [] }  // success: true even when data is []
```

An empty `data[]` (or a malformed non-array `data`) therefore passed the success
check and terminated the combo loop on the first leg.

## What changes

- Before returning `success: true`, the OpenAI-compatible normalization now
  requires **at least one usable image item** in `data[]`: an item with a
  non-empty `b64_json` **or** a non-empty `url` string (non-array/empty
  payloads fail the check).
- An empty/malformed 2xx is now a **retryable failure**: `success: false`,
  `status: 502`, sanitized error (`Image provider returned a success status
  without an image payload`), plus a `warn` log line naming the provider.
- Because the result is `success: false` with a 5xx status,
  `executeImageCombo()` classifies it as non-terminal and **advances to the
  next priority leg**, incrementing `X-OmniRoute-Fallback-Attempts`.

Scope is deliberately narrow: the change is confined to the single
normalization site shared by every OpenAI-compatible image generation/edit
forwarder. Valid responses (b64_json or url present), all non-image provider
handlers, direct model requests, and combo ordering/timeouts are unchanged.

## Tests

New `tests/unit/combo/image-combo-empty-200-fallback-10199.test.ts`
(real isolated SQLite DATA_DIR, real combo resolution + credential path,
stubbed `globalThis.fetch`; no paid/live requests):

- ✅ Empty 200 from the first leg becomes a retryable failure and the combo
  advances — second leg is selected and its image is served (both models hit,
  in order).
- ✅ Fallback metadata reflects the attempt:
  `X-OmniRoute-Fallback-Attempts: 1`.
- ✅ Valid first-leg response does not invoke later legs (single fetch hit).
- ✅ All legs returning empty 200 → retryable 502, sanitized error, no
  credentials/stack leakage.
- ✅ Direct image-model behavior preserved: empty 200 → 502; valid
  `b64_json` and `url` payloads → success unchanged.

Regression results:

- Existing combo + image-route suites: **11/11 pass**
  (`image-combo.test.ts`, `10197-openrouter-image-edits-route.test.ts`)
- Broader sweep (all `tests/unit/*image*.test.ts` + combo suites):
  **422/422 pass**
- `typecheck:core` (tsc): **clean**
- ESLint on changed files (repo suppressions): **clean**

Note: `npm run build` (Next.js production build) could not be executed in the
local 1 GiB-RAM sandbox (build exceeds the container limit and was OOM-killed
on attempt); the repo CI build gate will exercise it on this PR.

## Unchanged behavior

- Valid image responses: identical shape and status as before.
- Direct (non-combo) image generation requests: unchanged except that an
  empty 2xx now surfaces as 502 instead of a bogus success.
- Combo ordering, timeout settings, and all non-OpenAI-compatible provider
  handlers are untouched.


diegosouzapw:
Solid, narrowly-scoped fix — confirmed the bug is still live on the current `release/v3.8.51` tip (the `fetchImageEndpoint()` normalization you're patching is unchanged there), and re-ran your test suite locally: 5/5 pass, covering the empty-200-fallback, metadata, single-fetch-on-success, all-legs-empty, and direct-model-unchanged cases. One small housekeeping note: the test file references `#10199` in its name, but that issue number belongs to an already-merged, unrelated PR (the `auto/best-free` free-tier filter fix) — probably a copy-paste slip, worth renaming before merge so future readers aren't misled. Otherwise just needs a retarget from `release/v3.8.50` to `release/v3.8.51` (applies cleanly, no conflict) and this is ready to go.

diegosouzapw:
Thanks @tiangao88 — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

Issue #8609, opened 2026-07-25
fix(startup): Windows tray mode triggers Norton Behavioral Protection; autostart reports enabled but HKCU Run entry is absent


knoami:
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

22.23.1

### Provider(s) Involved

N/A

### Model(s) Involved

N/A

### Client Tool

OmniRoute CLI / tray mode

### Description

On Windows 11, running OmniRoute in tray mode causes Norton Behavioral
Protection to detect and terminate Windows PowerShell with:

`IDP.HELU.PSE85%s_cmd`

The behavior appears reproducible with:

`omniroute serve --tray --no-open`

Investigation shows that OmniRoute v3.8.48's Windows tray implementation
creates a temporary PowerShell script:

`%TEMP%\omniroute-tray-<pid>.ps1`

and launches it using approximately:

`powershell.exe -NonInteractive -File <script>`

with `windowsHide: true`.

This appears to be what triggers Norton's behavioral detection.

There is a second Windows autostart problem observed at the same time:
`omniroute autostart enable` reports `enabled: true`, and
`omniroute autostart` subsequently reports enabled, but the expected
registry value:

`HKCU\Software\Microsoft\Windows\CurrentVersion\Run\OmniRoute`

does not exist.

I have not added any Norton exclusions or disabled Behavioral Protection.

Local integrity checks were also performed. The installed v3.8.48 copies
of the relevant files match the official v3.8.48 Git blobs byte-for-byte:

- `package.json`
  - Git blob: `d90aa5b11e7b27fce0d4cdae03415ba678fa081b`
  - Match: true
- `bin/cli/tray/trayWindows.mjs`
  - Git blob: `f85de21fafd6e058ae160a62eedec448abadb131`
  - Match: true
- `bin/cli/tray/autostart.mjs`
  - Git blob: `605161c2a1f0aa1aa63686413a33a832daf51665`
  - Match: true

The Windows PowerShell executable involved is Microsoft's system binary:

`C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe`

Authenticode status: `Valid`
Signer: `Microsoft Windows`

So the observed AV behavior occurs with the unmodified upstream
OmniRoute v3.8.48 tray implementation.

### Steps to Reproduce

1. Install OmniRoute globally through npm on Windows 11.
2. Confirm version:

[...]

PR #7925, opened 2026-07-21
fix(autostart): adopt 9Router VBS startup to suppress console flash on Windows

tientien17:
## Summary

Windows auto-start previously wrote a `HKCU\Software\Microsoft\Windows\CurrentVersion\Run` registry entry that launched `node.exe` directly. Because `node.exe` is a console-mode application, a terminal window always flashed on screen at logon — and closing that window killed the background server.

This PR adopts the same approach as [9Router](https://github.com/gaten-io/9router): write a `.vbs` launcher script to the Windows Startup folder that calls `WScript.Shell.Run` with window style **0** (SW_HIDE). The server starts fully invisible and stays running independently of any console window.

### `bin/cli/tray/autostart.mjs`
- **New helpers**: `winStartupDir()`, `winStartupPath()`, `buildWinVbsContent()`, `cleanLegacyWinReg()`
- **`enableWin()`** — writes `OmniRoute.vbs` to `%APPDATA%\...\Startup\` with the tray serve command, then removes the old `HKCU\Run` entry (migration)
- **`disableWin()`** — deletes the VBS file and cleans up any lingering registry key
- **`isEnabledWin()`** — checks for the VBS file first, falls back to the legacy registry key (for pre-migration users)

### `tests/unit/cli/autostart-windows.test.ts`
New test file with:
- Integration test that sets `APPDATA` to a temp dir, calls `enable()`/`disable()`, and verifies the VBS file content (runs only on win32)
- Source-level assertions that verify the implementation uses `WScript.Shell`, `winStartupDir`, `unlinkSync`, etc. (runs on any platform — same pattern as the existing Linux/macOS autostart tests)

## Why VBS instead of --daemon?
OmniRoute already has a `--daemon` flag that spawns a detached, unrefed child process — but the tray icon integration (system tray menu to enable/disable autostart) calls the `enableWin`/`disableWin` functions which previously used the Registry Run key. The VBS approach replaces that exact code path, keeping the tray UI unchanged while eliminating the console flash.

The VBS file content is built from the same `buildServeExecLine` helper that the other auto-start mechanisms use, so the command-line flags (`serve --no-open --tray`) stay consistent across platforms.

## Testing
- `npm run test:unit` — all 8 autostart tests pass (4 new + 4 existing)
- `npm run lint` — clean (pre-commit hooks ran via husky/lint-staged)
- Verified on Windows 11: `enable()` writes `OmniRoute.vbs`, `disable()` removes it

PR #12993, opened 2026-09-07
fix(cli): skip POSIX path lookup on Windows autostart

zachary-frederich:
## Summary

Prevent Windows autostart path resolution from invoking the following POSIX-only probe:

```bash
command -v omniroute 2>/dev/null
```

On Windows, this probe can emit:

```text
The system cannot find the path specified.
```

even though autostart creation subsequently succeeds.

The POSIX lookup is now skipped on `win32` while preserving the existing argv/module-relative path resolution and keeping POSIX behavior unchanged.

## Root cause

`resolveCliPath()` executed a POSIX shell command on every platform.

On Windows, Node's shell is not a POSIX shell and `/dev/null` is not the appropriate null device/path. The failing lookup was caught, allowing `enableWin()` to continue, which caused a misleading error followed by a successful `enabled: true` result.

## Real-world reproduction

Environment:

- Windows 10 Home 22H2, build 19045 x64
- Node.js v24.19.0
- npm global install
- reproduced on OmniRoute 3.8.50

Before:

```text
> omniroute autostart enable
The system cannot find the path specified.
---------
| enabled |
---------
| true    |
---------
```

The Startup-folder `OmniRoute.vbs` was nevertheless created.

After applying the platform guard:

```text
> omniroute autostart enable
---------
| enabled |
---------
| true    |
---------
```

No bogus Windows path error was emitted, and `OmniRoute.vbs` was still created successfully.

The red regression run in this checkout independently reproduced the bogus error before the production change; the green run passed without it.

Related to #8609, which documented the same misleading Windows error. #9374 addressed the tray implementation, while this patch fixes the remaining autostart path probe after the VBS migration in #7925.

## Scope

This patch intentionally does not change:

- Windows VBS autostart mechanism
- tray behavior
- legacy registry cleanup
- macOS LaunchAgent behavior
- Linux systemd/XDG behavior
- CLI output schemas

## Tests

Changed test file:

- `tests/unit/cli/autostart-windows.test.ts`

Passed:

- `node --import tsx/esm --test tests/unit/cli/autostart-windows.test.ts` — 5/5
- `node --import tsx/esm --test tests/unit/cli/autostart-linux.test.ts` — 4/4
- `node --import tsx/esm --test tests/unit/cli/autostart-macos-launchctl.test.ts` — 10/10
- `node --import tsx/esm --test tests/unit/cli/autostart.test.ts` — 1/1
- `node --import tsx/esm --test tests/unit/autostart-cli-shorthand-3331.test.ts` — 4/4
- `npm run check:cli-i18n`
- `node --check bin/cli/tray/autostart.mjs`
- changed-file ESLint and Prettier checks
- `git diff --check`
- `$env:CHANGELOG_BASE_REF="upstream/release/v3.8.51"; npm run check:changelog-integrity`

Local base/environment limitations:

- `npm run build:cli` could not complete safely on this 8 GB Windows host. Both the default Turbopack build and the documented `OMNIROUTE_USE_TURBOPACK=0` webpack fallback exceeded available memory and entered heavy paging, so they were stopped. The repository's build workflow likewise documents that the v3.8.51 tree does not fit hosted 7 GB runners; CI/self-hosted build remains the authoritative signal.
- `npm run check:pack-policy` reports 27 unexpected `@omniroute/opencode-plugin-v2/**` files already present outside this patch.
- `npm run lint` reports repository-wide stale suppressions (`There are suppressions left that do not occur anymore`). ESLint on both changed files passes; pre-commit ESLint also passed.

## Risk

Low.

- Windows no longer attempts the POSIX PATH probe.
- Existing `process.argv[1]` and module-relative fallback candidates remain.
- Existing realpath/existence validation remains.
- POSIX behavior is unchanged.
- No shell use or interpolation was added.

### DIFF
--- a/bin/cli/tray/autostart.mjs
+++ b/bin/cli/tray/autostart.mjs
@@ -13,11 +13,13 @@ const LINUX_DESKTOP_NAME = "omniroute.desktop";
 function resolveCliPath() {
   const candidates = [];
   if (process.argv[1]) candidates.push(process.argv[1]);
-  try {
-    const which = execSync("command -v omniroute 2>/dev/null", { encoding: "utf8" }).trim();
-    if (which) candidates.push(which);
-  } catch {
-    // command -v unavailable
+  if (process.platform !== "win32") {
+    try {
+      const which = execSync("command -v omniroute 2>/dev/null", { encoding: "utf8" }).trim();
+      if (which) candidates.push(which);
+    } catch {
+      // command -v unavailable
+    }
   }
   candidates.push(join(dirname(fileURLToPath(import.meta.url)), "..", "..", "omniroute.mjs"));
 
--- a/changelog.d/fixes/12993-windows-autostart-posix-lookup.md
+++ b/changelog.d/fixes/12993-windows-autostart-posix-lookup.md
@@ -1,0 +1,1 @@
+- **fix(cli):** skip the POSIX CLI path lookup during Windows autostart setup, preventing a bogus path error before successful enablement ([#12993](https://github.com/diegosouzapw/OmniRoute/pull/12993))
--- a/tests/unit/cli/autostart-windows.test.ts
+++ b/tests/unit/cli/autostart-windows.test.ts
@@ -69,6 +69,23 @@ test("Windows enable/disable writes and removes VBS in Startup folder", async ()
 // autostart-macos-launchctl.test.ts.
 // ---------------------------------------------------------------------------
 
+test("Windows path resolution skips the POSIX command lookup", () => {
+  const source = readFileSync(join(process.cwd(), "bin/cli/tray/autostart.mjs"), "utf8");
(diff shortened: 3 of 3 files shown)


diegosouzapw:
Thanks @zachary-frederich — clear repro on real Windows 10/Node 24, and the fix
(skipping the POSIX `command -v` probe on `win32`) is minimal and correctly scoped; the
non-Windows path is untouched. Looks good to merge.

diegosouzapw:
Thanks @zachary-frederich — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #12995, opened 2026-09-07
fix(i18n): stop swallowed FORMATTING_ERROR from showing raw keys/garbled text

hartmark:
## What Problem This Solves

`/dashboard/combos`' Intelligent-Routing wizard step (and two sibling panels)
showed the raw untranslated i18n key instead of real text, plus logged a
scary `IntlError: FORMATTING_ERROR` to the console, whenever the
corresponding `en.json` message required an ICU variable.

Reported live: opening the Intelligent Routing builder step showed
`combos.explorationRateHint` instead of "20% of requests can explore
non-optimal providers." and the browser console printed:

```
Error: FORMATTING_ERROR: The intl string context variable "percent" was not provided...
    at BuilderIntelligentStep.tsx:331
```

## Why This Change Was Made

Three call sites across the app built a "safe i18n getter" that called
`t(key)` with **no values**, then string-`.replace()`d ICU placeholders on
the *string result*:

```ts
getI18nOrFallback(t, "explorationRateHint", "{percent}% ...")
  .replace("{percent}", `${percent}`)
```

This only works while the translation key is **unregistered** in the
locale catalog — `t.has(key)` is `false`, so the untouched fallback string
is returned and `.replace()` fills it in fine. The moment the key is
registered in `en.json` with an ICU variable that the fallback string also
names (`{percent}`, `{items}`, `{source}`/`{target}`), `t(key)` instead
formats the *registered* message — and next-intl can't fill in a variable
nobody passed it, so it raises `IntlErrorCode.FORMATTING_ERROR`.

I checked every `NextIntlClientProvider`/`getRequestConfig` in this app
(`app/layout.tsx`, `app/global-error.tsx`, `src/i18n/request.ts`) — none of
them configure a custom `onError`. That means next-intl's own default
`onError` (`console.error(error)`, then fall back to the raw
`"namespace.key"` string) is what's live here. So the bug shape here is
"silent failure with noisy logging," not an actual uncaught render crash —
concretely: the user sees the raw key text instead of the real hint, and a
scary-looking (but caught) exception hits the console every time that step
renders.

### Fix

Pass the ICU values into `t(key, values)` itself, and keep a fallback for
the case where the value is still missing/mismatched:

- `combos/BuilderIntelligentStep.tsx` — `explorationRateHint`'s `{percent}`
- `agent-bridge/AgentBridgeMaintenanceCard.tsx` — `repairDone`'s `{items}`,
  `importDone`'s `{bypass}`/`{hosts}`/`{agents}`
- `translator/RawJsonPanel.tsx` — `translationPathHubSpoke` and
  `translationPathDirect`'s `{source}`/`{target}`

Two other call sites using the same string-replace pattern
(`IntelligentComboPanel.tsx`, `combos/page.tsx`) already pass `values`
correctly into `t()` and needed no change.

## User Impact

Users configuring Intelligent Routing (and using the Agent Bridge
maintenance repair/import actions, or the translator's raw-JSON panel path
description) now see the real, correctly interpolated copy instead of a raw
i18n key, with no console noise.

## Evidence

New regression test builds a **real** `use-intl` translator
(`createTranslator`) from this app's actual `en.json` `combos` namespace,
with no `onError` override — reproducing the exact production
configuration — and asserts the rendered exploration-rate hint is fully
interpolated (`"20% of requests can explore non-optimal providers."`), not
the raw key. Confirmed it fails against the pre-fix `getI18nOrFallback`
(renders `combos.explorationRateHint` instead) and passes after the fix.

```
$ npx vitest run tests/unit/BuilderIntelligentStep-formatting-error.test.tsx
 Test Files  1 passed (1)
      Tests  2 passed (2)

$ npx vitest run tests/unit/translator-friendly-integration.test.tsx tests/unit/translator-friendly-page-client.test.tsx tests/unit/translator-friendly-raw-json-panel.test.tsx
 Test Files  3 passed (3)
      Tests  23 passed (23)
```

- `eslint` on all 4 touched files: clean

Note: I don't have an existing dedicated component test for
`AgentBridgeMaintenanceCard.tsx`; its fix is the same bug class as the
regression-tested `BuilderIntelligentStep.tsx` fix and is covered by
manual verification of the `en.json` message shapes plus the existing
`agent-bridge-maintenance-api.test.ts` suite (unaffected, still green).


diegosouzapw:
Excellent find and an excellent regression test — reproducing the real `FORMATTING_ERROR` against
the actual `en.json` catalog (not a mock) is exactly the right way to pin this down, and I
confirmed it fails on the current tip without your fix and passes with it. This is ready to
merge once a changelog fragment is added under `changelog.d/fixes/`. Thanks for tracking down the
root cause across all three call sites instead of just patching the one that was reported.

hartmark:
Rebased onto current `release/v3.8.51` (was 236 commits behind — CI's broad failures were staleness, same pattern as the other PRs in this batch). Re-ran the new vitest suite at the rebased head (2/2 pass), typecheck clean. Added the changelog fragment — verified with `check:changelog-integrity`. Ready to merge from my side.

Issue #12268, opened 2026-09-01
fix(api): Codex CLI's built-in image_gen can't use OmniRoute — two response-shape gaps on /v1/images/generations


yongseokleecom:
### OmniRoute Version

3.8.50 (Docker, `diegosouzapw/omniroute:latest`)

### Installation Method

Docker

### Provider(s) Involved

`codex` (ChatGPT subscription OAuth)

### Model(s) Involved

`codex/gpt-5.6-sol`, reached through a combo named `gpt-image-2`

### Client Tool

Codex CLI 0.149.1 — its **built-in `image_gen` tool**

### Description

Codex CLI's built-in `image_gen` cannot generate images through OmniRoute. It posts a plain OpenAI images request to the gateway, and the response shape is off in two independent places on `POST /v1/images/generations`:

1. **the combo path drops the `{created, data}` wrapper** and returns a bare array;
2. **`codex` image results carry only `url` (a `data:` URI) and never `b64_json`** — on **both** `/v1/images/generations` and `/v1/images/edits`.

Both must be fixed for the client to work — they surface in sequence, which is a convenient way to tell them apart:

```
① invalid type: map, expected u64      ← wrapper missing
② missing field `b64_json`             ← after ① is fixed
```

The images themselves are fine throughout; decoding the base64 by hand yields a valid 1254×1254 PNG. Only the envelope blocks the client.

This matters beyond our setup: Codex CLI is one of OmniRoute's headline clients, and its `image_gen` is the path a subscription user would naturally take.

### Steps To Reproduce

1. Create a combo named `gpt-image-2` with a single target `codex/gpt-5.6-sol` (priority strategy). Codex CLI hardcodes the model name `gpt-image-2`, so the combo is what lets it reach a codex target at all.

2. Ask the target **directly** — this is the control:

```bash
curl -X POST "$OMNIROUTE/v1/images/generations" \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"codex/gpt-5.6-sol","prompt":"a green tree icon, white background, minimal flat"}'
```

```json
{ "created": 1788246786, "data": [{ "url": "data:image/png;base64,iVBORw0KGgo…" }] }
```

3. Ask for the same thing **through the combo name**:

```bash
curl -X POST "$OMNIROUTE/v1/images/generations" \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
[...]

PR #12997, opened 2026-09-07
fix(images): preserve OpenAI envelope in image combo success response (3.8.50 backport)

tiangao88:
## Summary

`executeImageCombo()` (open-sse/services/imageCombo.ts) double-unwraps the
already-normalized OpenAI payload on the success path and returns a **bare
array** `[{"b64_json": …}]` instead of the OpenAI Images envelope
`{"created": …, "data": […]}`.

Raw `curl` works, but every spec-compliant client rejects the response:

- Official OpenAI SDKs and OpenAI-compatible clients read `response.data` —
  on a bare array they see "no image data" (the Hermes omniroute image
  plugin fails with exactly `pmoc.or-image-text returned no image data.`).
- Dashboard clients that read `data?.data || []` (e.g. `MediaPageClient`)
  silently render nothing for combo-generated images.

The direct (non-combo) route (`src/app/api/v1/images/generations/route.ts`)
keeps the envelope, so only combo-generated images break.

## Root cause

`fetchImageEndpoint()` (open-sse/handlers/imageGeneration.ts) normalizes
provider responses to the public envelope `{ created, data: [...] }`. The
combo success path then unwraps it a second time:

```ts
// open-sse/services/imageCombo.ts (success path)
JSON.stringify((successResult.data as { data: unknown }).data)
//        ↑ returns only the items array, dropping `created` + envelope shape
```

## What changes

- The success path returns `successResult.data` — the full normalized
  envelope — unchanged, matching the direct route and the `release/v3.8.51`
  behavior (`#12268` already normalized this on the newer line; this PR is
  the backport for `release/v3.8.50`).
- New regression test
  `tests/unit/combo/image-combo-openai-envelope-10231.test.ts`: real
  isolated SQLite `DATA_DIR`, real combo resolution and credentials path,
  stubbed `globalThis.fetch`; asserts the combo response is the envelope
  object (not an array), carries `created` + `data[]` with the winning
  leg's image, and that the bare-array shape never returns. A second test
  exercises leg fallback (leg 1 returns 502 → leg 2 serves the image
  inside the envelope).

## Test plan

- `node --test tests/unit/combo/image-combo-openai-envelope-10231.test.ts` — 2/2 pass
- Full `tests/unit/combo/` sweep (explicit file list) — 169/169 pass
- `npx tsc -p tsconfig.typecheck-core.json` — clean
- `npx eslint` on changed files — clean
- `npx prettier --check` on changed files — clean

## Live verification

Deployed on a production combo (`pmoc-image-text`, three image legs) on
`3.8.50-fixed`: `image_generate` from a spec-compliant client previously
failed with "returned no image data"; after this change the same call
returns a valid envelope and the image renders.

## Related

- #12982 - companion fork PR: combo legs that return **empty 2xx bodies** are now treated as retryable failures (envelope fix here assumes a usable payload; that PR makes empty ones fall through to the next leg).


tiangao88:
Update — root cause for two of the three red checks, both pre-existing on `release/v3.8.50` (this PR's diff does not touch them; they fail identically on the head of #12982):

1. **Unit Tests fast-path (1/4)** — the only failing test is `tests/unit/alibaba-free-tier-allowlist.test.ts` ("allowlist JSON pack overrides embedded lists when valid"). Reproduced locally: `config/alibaba-free-tier-allowlist.json` carries `"validUntil": "2026-08-27"`, which has now expired, so `isAlibabaFreeTierAllowlistPackValid()` returns false and `loadAlibabaFreeTierAllowlistPack()` returns null -> `assert.ok(pack)` fails. Refreshing the pack's `validUntil` in the repo fixes it. Everything else in the shard passes, including the full image-combo suite added here.

2. **No new ESLint warnings** — both files changed in this PR (`open-sse/services/imageCombo.ts`, `tests/unit/combo/image-combo-openai-envelope-10231.test.ts`) run clean under the repo's ESLint config; the failure is from pre-existing warnings elsewhere in the tree (surfaced after the lint-freeze revert in 091589089).

3. **Fast Quality Gates** — composite of the above; no independent failure.

Happy to open a separate PR refreshing the allowlist pack if that helps clear the shard.

diegosouzapw:
Thanks for the detailed root-cause writeup — this exact bug (double-unwrap dropping the OpenAI envelope) was independently fixed on `release/v3.8.51` via #12268, which is already merged and actually goes a step further (it re-wraps the payload even if a handler ever returns a bare array). Since `release/v3.8.51` is the only actively-developed release branch right now (not `v3.8.50`), I don't think there's a live target for this backport — I'm marking it subsumed by #12268 rather than asking for a rebase. If you'd still like credit for the regression test you wrote (`image-combo-openai-envelope-10231.test.ts`), happy to take a follow-up PR porting it against the current `executeImageCombo` shape with your authorship preserved. Let me know if `release/v3.8.50` is still meant to receive independent fixes and I'll revisit.

tiangao88:
Confirmed — I checked `release/v3.8.51` directly and the fix is live there (`8d16a50df`, via PR #12362, closing #12268): the combo success path now re-wraps the payload, and the regression coverage landed with it. So this backport is redundant — closing it. No independent fixes are expected on `release/v3.8.50`. Thanks for the clean triage.

PR #12999, opened 2026-09-08
fix(db): auto-clean terminal batch checkpoints and expired file content

hartmark:
## What Problem This Solves

`batch_item_checkpoints` and `files` had **no automatic cleanup path at all**,
unlike the 14 other tables `runAutoCleanup()` already sweeps on a schedule
(`quota_snapshots`, `call_logs`, `usage_history`, `compression_analytics`,
`mcp_tool_audit`, `config_audit_log`, `a2a_task_events`, `memories`,
`domain_cost_history`, `compression_cache_stats`, `xp_audit_log`,
`compression_run_telemetry`, `proxy_logs`, `ccr_blocks`).

Observed live: `batch_item_checkpoints` had grown to **182K rows / 5.25 GB**
and `files` to **1,874 rows / 5.19 GB** -- together ~95% of an 11 GB
database -- with no batch ever explicitly deleted by an operator and most
files long past their own expiry.

## Root Cause

- The only existing sweep for batches, `deleteCompletedBatches()` (the
  operator-triggered `DELETE /api/v1/batches/delete-completed` route), must
  be called **manually**, has **no age filter**, and only covers `status =
  'completed'` -- a failed, cancelled, or expired batch's checkpoints are
  just as done as a completed one's, but had no cleanup path whatsoever.
- `files.expires_at` (an OpenAI-Files-API-compatible expiry field, set
  automatically for batch files) was **never enforced by anything** -- only
  an explicit delete call ever cleared a file's BLOB content.

## Fix

- **`batches.ts`**: refactored `deleteCompletedBatches()` onto a shared
  `deleteBatchesMatching()` helper -- **no behavior change**, same `status =
  'completed'`, no age filter, same public API contract for the DELETE
  route -- and added `deleteTerminalBatchesOlderThan(days)` for the
  automatic sweep: covers every terminal status
  (completed/failed/cancelled/expired), gated by age so a batch's results
  stay retrievable for a while after finishing, matching OpenAI's own Batch
  API retention behavior. Default 30 days, overridable via
  `OMNIROUTE_BATCH_RETENTION_DAYS`.
- **`files.ts`**: added `pruneExpiredFiles(now)`, mirroring `deleteFile()`'s
  soft-delete shape (row kept for metadata/audit, BLOB content cleared).
  Needs no separate retention-days setting -- a file already carries its own
  expiry, the same way `ccr_blocks` does (see `cleanupCcrBlocks`).
- **`cleanup.ts`**: wired both into `runAutoCleanup()` as
  `cleanupOldBatches()` and `cleanupExpiredFiles()`, alongside the other 14
  automatic sweeps.
- Documented `OMNIROUTE_BATCH_RETENTION_DAYS` in
  `docs/reference/ENVIRONMENT.md` and `.env.example`
  (`check:env-doc-sync` passes).

## What This Deliberately Does NOT Change

`deleteCompletedBatches()` / `DELETE /api/v1/batches/delete-completed`
keeps its exact existing contract (unconditional, `status = 'completed'`
only, no age filter) -- it's a public API route, not the automatic cleanup
path, so widening its scope silently would be a behavior change for anyone
already calling it. The new age-gated, all-terminal-statuses sweep is a
separate function used only by the automatic cleanup job.

## Evidence

New regression tests (9 cases) prove: an old terminal batch and its
checkpoint are deleted; a recent one survives; all four terminal statuses
are covered (not just completed); a non-terminal batch is never touched
regardless of age; an expired file's content is cleared while its
row/metadata survive; a file with no expiry or a future expiry is
untouched; an already-deleted file isn't re-processed; and both
`cleanup.ts` wrappers behave the same way end-to-end.

```
$ node --test tests/unit/db-terminal-batch-and-file-cleanup.test.ts
tests 9, pass 9, fail 0

$ node --test tests/unit/batch-deletion.test.ts tests/unit/batch-deletion-route-logic.test.ts
tests 19, pass 19, fail 0   # deleteCompletedBatches()'s existing contract fully preserved

$ node --test tests/unit/file-expiration-policy.test.ts tests/unit/db-cleanup.test.ts \
    tests/unit/telemetry-auto-cleanup-6848.test.ts tests/unit/db-cleanup-xp-audit-log.test.ts
tests 54, pass 54, fail 0
```

- `eslint` on all touched files: clean (1 pre-existing, unrelated unused-var
  error in `files.ts` confirmed present on unmodified `release/v3.8.51`
  too -- not touched by this PR)
- `node scripts/check/check-env-doc-sync.mjs`: in sync


diegosouzapw:
Careful, well-scoped fix — extracting `deleteBatchesMatching()` while keeping
`deleteCompletedBatches()`'s exact existing contract (proved by the regression
test) and adding age-gated cleanup for the terminal-status gap is exactly the
right shape. The explicit-id `IN (...)` delete is also a nice robustness
improvement over the double-subquery it replaces. Ran both your new test file
and the existing `batch-deletion.test.ts` regression suite at this PR's head:
17/17 pass. Only ask before merge: could you add a `changelog.d/fixes/`
fragment (repo convention, missing from this diff)?

hartmark:
Rebased onto current `release/v3.8.51` (was 233 commits behind — CI's typecheck/unit-test failures were staleness, same as noted on the other PRs in this batch). This one had a real conflict, not just drift: security hardening landed on `deleteCompletedBatches()` since this branch was opened (LEDGER-18 presence-based mutual-exclusion check, LEDGER-22 forward-progress guard, per-chunk commits for both scope modes). Rebuilt `deleteBatchesMatching()` to take an arbitrary WHERE clause while preserving every one of those invariants verbatim, rather than reverting to the older shape this PR's diff was based on.

Caught one regression from the merge: I'd renamed the file-soft-delete-failure log message during the extraction, which broke `batches-delete-completed-ownership-wvxc.test.ts`'s pinned assertion on the literal string `deleteCompletedBatches: file soft-delete failed` — reverted that string since it's an observable ops-facing log line, not an implementation detail. All 48 tests across the batch/cleanup suite pass at the rebased head, typecheck clean. Added the changelog fragment — verified with `check:changelog-integrity`. Ready to merge from my side.

Issue #2359, opened 2026-05-18
[BUG] Combo returns error when calling local provider


dexter-super-work:
### OmniRoute Version

3.8.0 / latest docker

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24

### Provider(s) Involved

vllm, llama.cpp

### Model(s) Involved

Qwen-3.6-35b-a3b, also Qwen-3.6-27b

### Client Tool

Claude code and the Omniroute's Combo 'Test button'

### Description

Docker version omniroute calling a local provider within the same docker network, which wrapped inside a combo, will return "e.startWith is not a function".
Checked the "Log" section the provider does have response payload,  but omniroute cannot handle it correctly.

### Steps to Reproduce

1. Start a omniroute docker container with :latest, and with docker network 
2. Start another vllm/llama.cpp docker container, using qwen3.6 a35b-a3b or 27b as model, and with docker network
3. Setup a Anthropic/OpenAI compatible provider and connects it
4. Test the connection with Claude code using the provider/model, it should work
5. Now wrap the provider and model into a Combo
6. Test it with the Combo 'Test' button, it will return 'e.startWith is not a function'
7. Or call the combo by Claude code, it will hang.
8. Check the Log section, the API actually being called and has response payload returned.

### Expected Behavior

Connect successfully and content being handled correctly and pass back to agent.

### Actual Behavior

Response payload got stuck at omniroute and not passing back to agent.

### Test Impact

Needs a new e2e test

### Error Logs / Output

```shell
Don't worry about the providerai with those "0000'. I changed them and it works well before upgrade to 3.8.0.
2026-05-18T05:53:41.303687087Z 2026-05-18T05:53:41.293Z 30 service=omniroute module=sse tag=HTTP method=POST path=/v1/chat/completions | openai-compatible-chat-vllm0000-0000-0000-0000-000000000000/local | 1 msgs msg=📥 POST /v1/chat/completions | openai-compatible-chat-vllm0000-0000-0000-0000-000000000000/local | 1 msgs
2026-05-18T05:53:41.304229356Z 2026-05-18T05:53:41.293Z 20 service=omniroute module=sse tag=AUTH msg=API Key: sk-3...d227
2026-05-18T05:53:41.304266837Z 2026-05-18T05:53:41.294Z 20 service=omniroute module=sse tag=GUARDRAIL msg=vision-bridge pre-call passed
[...]

PR #13001, opened 2026-09-08
fix(combos): keep dashboard combo tests off reasoning and rate limits

HouMinXi:
## Problem

The dashboard combo test reports `error` on every step for combos backed by
rate-limited or reasoning-heavy providers, even when the same combo answers
real chat traffic fine.

Two causes:

1. **The probe prompt is expensive.** It asks a model to add two random
   five-digit numbers with `max_tokens: 2048`. Reasoning models spend the whole
   budget thinking and return no text, so the probe records "HTTP 200 but no
   text content".
2. **All steps are probed at once with a 20s deadline.** `Promise.all` fires
   one request per step simultaneously. On a free-tier pool that is an instant
   burst of N concurrent requests against the same account, and 20s is below
   the first-byte latency of thinking models. A ten-step combo returned
   `20001ms error` on all ten steps.

## Change

- Probe prompt is now `Reply with exactly: pong` with `max_tokens: 64`.
- Steps are probed sequentially instead of concurrently.
- Per-step deadline 20s -> 60s (`COMBO_TEST_TIMEOUT_MS`).
- New total budget `COMBO_TEST_TOTAL_TIMEOUT_MS` = 180s. Once spent, the
  remaining steps report `Timeout (180s total)` instead of being probed, so a
  large combo cannot make the endpoint hang.

The randomized arithmetic prompt was there to defeat upstream caching. A cached
`pong` still proves the step is reachable and authenticated, which is what the
health check is for.

## Tests

23/23 in `tests/unit/combo-test-health.test.ts`,
`tests/unit/combo-test-route.test.ts`,
`tests/unit/combo-target-defensive-modelstr.test.ts`; `typecheck:core` clean.

New coverage: probes run sequentially and in combo order; the total budget
stops further probing. Both were verified by injection - restoring
`Promise.all` and disabling the budget check each turn the new tests red.

The `#2359` malformed-`modelStr` guard is untouched.

### DIFF
--- a/changelog.d/fixes/combo-test-probe-timeout.md
+++ b/changelog.d/fixes/combo-test-probe-timeout.md
@@ -1,0 +1,1 @@
+- **fix(combos):** dashboard combo test uses a short prompt, serial probes, and a 60s timeout so reasoning models and rate-limited free pools do not fail the health check
--- a/src/app/api/combos/test/route.ts
+++ b/src/app/api/combos/test/route.ts
@@ -5,10 +5,14 @@ import { getComboByName, getCombos } from "@/lib/db/combos";
 import { pickApiKeyForInternalUse } from "@/lib/db/apiKeys";
 import { getRuntimePorts } from "@/lib/runtime/ports";
 import { resolveNestedComboTargets } from "@omniroute/open-sse/services/combo.ts";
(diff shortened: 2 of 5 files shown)


HouMinXi:

PR #13002, opened 2026-09-08
fix(streaming): per-provider fetch-start timeout cap override (#11526 follow-up)

alvinveroy:
## What

Buffered gateways (opencode-go / command-code "Console Go" tiers) hold the whole generation behind the gateway before the first upstream byte, so their streaming requests legitimately exceed the default 110 s headers-wait cap. Upstream #11526 caps **every** streaming request at `DEFAULT_FETCH_START_TIMEOUT_CAP_MS = 110_000`, so these long reasoning generations die at exactly `Fetch timeout after 110000ms` (504) before any bytes arrive.

## Change

Add a per-provider `fetchStartTimeoutCapMs` registry knob (`600_000` for `opencode/go` and `command-code`) and project it into the executor's `LegacyProvider` config so `resolveFetchStartTimeout` caps only genuinely unbounded providers instead of every streaming request.

- `open-sse/config/providers/shared.ts` — declare `fetchStartTimeoutCapMs` on the registry entry + LegacyProvider.
- `open-sse/config/providerRegistry.ts` — project the knob through `generateLegacyProviders()` (the field was previously dropped here, silently falling back to the 110 s cap).
- `open-sse/executors/base.ts` — pass `capMs: this.config?.fetchStartTimeoutCapMs` into `resolveFetchStartTimeout`.
- `open-sse/config/providers/registry/{opencode/go,command-code}/index.ts` — set `600_000`.
- `tests/unit/fetch-start-cap-legacy-projection.test.ts` — regression for the projection fix.


diegosouzapw:
Good, well-scoped follow-up to #11526 — I verified `resolveFetchStartTimeout()` already
accepted a `capMs` input before this PR, so you're correctly threading an existing mechanism
through the registry rather than inventing new plumbing. Ran your 2 new tests, both green. Two
small notes: the `requestDefaults: { maxTokens: 16_384 }` addition on the command-code/
opencode-go entries is a bit outside the "fetch-start timeout cap" scope in the title (though
harmless and plausible on its own) — worth a one-line mention in the changelog fragment so
reviewers aren't surprised. Speaking of which, this is missing a `changelog.d/fixes/` fragment.

alvinveroy:
Both notes addressed on `ca2c4a12d` (rebased onto `release/v3.8.51` @ `cde49c937`):

1. **Changelog fragment** — added at `changelog.d/fixes/13002-fetch-start-cap-per-provider.md`. Per your note it explicitly names the out-of-title addition rather than leaving it implicit: "…the same two entries also gain a reasoning-safe `requestDefaults.maxTokens` of 16384 so thinking models such as `z-ai/glm-5.3-flash` are not cut off mid-reasoning". Reviewers see it in the fragment instead of discovering it in the diff.

2. **Rebase** — branch replayed onto current tip with no conflicts; `mergeable=true`, head now `ca2c4a12d`.

`tests/unit/fetch-start-cap-legacy-projection.test.ts` — 2/2 pass.

Thanks for confirming `resolveFetchStartTimeout()` already accepted `capMs`; threading the existing mechanism through the registry projection was the intent — no new plumbing on either layer.

diegosouzapw:
Thanks @alvinveroy — merging via the release merge-train. Validated in local merge-train /tmp/mt-train3b.log on 192.168.0.113 @ train tip 408e2128791696a18966681953136fc96aeb99b0 (32 PRs boarded): static gates green; full test:unit 40694 tests, 17 failing — every one reproduces on the pure release tip (base-red sweep list), zero new reds. Merged --admin per merge-gates §4/§7.

PR #13003, opened 2026-09-08
fix(ops): egress-IP lockout provider allowlist override

alvinveroy:
## What

`applyEgressIpLockout` (auth.ts) parks **every sibling connection** of a provider when one account trips the egress-IP bucket. For free-tier IP-bucketed plans that cascade was designed; for paid plans (per-workspace quota, not per-IP) one account's upstream 429 wrongly parks all siblings for days.

## Change

Add `OMNIROUTE_EGRESS_IP_LOCK_PROVIDERS`: `none`/`off`/`false`/`0` disables the lockout; a CSV replaces the provider set. Lets operators scope the lockout via env instead of editing the compiled bundle.

- `open-sse/config/providerErrorRules.ts` — read the env override into the lockout provider set.
- `.env.example` — document the knob.
- `tests/unit/egress-ip-lock-env-override.test.ts` — regression.


diegosouzapw:
The core feature here (`OMNIROUTE_EGRESS_IP_LOCK_PROVIDERS` env override) is clean and well
tested — ran your 4 tests, all green. One thing to fix before merge: GitHub is showing this PR
as CONFLICTING, and I traced why — your `.env.example` diff also touches
`RATE_LIMIT_MAX_WAIT_MS`/`RATE_LIMIT_EXECUTION_MAX_WAIT_MS`, which is unrelated to the
egress-IP-lockout topic and has already been changed by another merged PR since your branch was
cut (the tip now has `30000`/`600000`, not the `15000`→`120000` your diff assumes). Could you
rebase and drop that unrelated block from the diff? Also missing a `changelog.d/fixes/`
fragment. Everything else looks solid.

alvinveroy:
Conflict traced and fixed on `f8a226fae` (rebased onto `release/v3.8.51` @ `cde49c937`), which also resolved the CONFLICTING state — `mergeable=true` now.

1. **Unrelated `RATE_LIMIT_*` block dropped.** Your diagnosis was exact: my diff assumed the `15000`→`120000` pairing, while the tip has since moved to `30000`. Confirmed both sides now agree — `.env.example` reads `RATE_LIMIT_MAX_WAIT_MS=30000` and `RATE_LIMIT_EXECUTION_MAX_WAIT_MS=600000` on base and branch alike, and the diff no longer touches either key (`git diff base..branch -- .env.example` is the 11-line `OMNIROUTE_EGRESS_IP_LOCK_PROVIDERS` block only, `grep -c '^+.*RATE_LIMIT'` → 0).

2. **Changelog fragment** — `changelog.d/fixes/13003-egress-ip-lockout-allowlist-override.md`.

3. **Env docs** — added a third commit, `docs(env): document OMNIROUTE_EGRESS_IP_LOCK_PROVIDERS`, so `docs/reference/ENVIRONMENT.md` stays in step with the new knob.

`tests/unit/egress-ip-lock-env-override.test.ts` — 4/4 pass. Dropping the stale block was the whole cause: the branch was green against `cde49c937` otherwise.

PR #13004, opened 2026-09-08
fix(opencode): clamp embedding inputs, glm-5.3 vision fragment, dispatch-shape compat

alvinveroy:
## What

A cluster of opencode-family compat fixes surfaced by live GLM traffic:

1. Embedding inputs longer than an 8192-token model's window are rejected; clamp them.
2. The `glm-5.3` vision fragment was dropped for multimodal turns; preserve it.
3. Reasoning-presence-enforcing gateways require a `reasoning_text` item back; emit a sentinel instead of an absent field (DeepSeek-direct still accepts absent).
4. GLM same-format replay (`[1214]`) needs a trailing user turn; append a `(continue)` turn when none exists.

## Change

- `open-sse/handlers/embeddings.ts`, `src/lib/memory/embedding/remote.ts` — 8192-token input clamp.
- `src/shared/constants/visionModels.ts` — preserve the glm-5.3 vision fragment.
- `open-sse/translator/index.ts`, `open-sse/translator/request/openai-responses/toResponses.ts` — reasoning-text sentinel + GLM `(continue)` turn.
- Regression tests for each.


diegosouzapw:
Solid work on the translator fixes (GLM synthetic-user-turn and the opencode
reasoning-presence branch) — both reuse existing infrastructure correctly and I could trace
the production issue shape they describe. One thing that needs fixing before merge though: 
`clampEmbeddingStringInput()` in `open-sse/handlers/embeddings.ts` is exported but never
actually called from `buildUpstreamBody()` — I checked, `upstreamBody.input =
runtime.body.input` still goes out unclamped on the `/v1/embeddings` route. It's only wired up
on the `src/lib/memory/embedding/remote.ts` side (that part's correct). Right now the only
thing exercising the new function is its own unit test, so the gap wouldn't get caught by CI.
Could you wire it into `buildUpstreamBody` (or clarify if that route intentionally stays
unclamped)? Also missing a `changelog.d/fixes/` fragment. Everything else ran green for me (9/9
across your three test files).

alvinveroy:
You were right — the function was exported but dead on the `/v1/embeddings` route. Fixed on `f321ffb35`.

`buildUpstreamBody()` now clamps at the source rather than relying on the caller:

```ts
function buildUpstreamBody(runtime: EmbeddingRuntime): Record<string, unknown> {
  const upstreamBody: Record<string, unknown> = {
    model: runtime.model,
    input: clampEmbeddingStringInput(runtime.body.input),
  };
```

That replaces the unclamped `runtime.body.input` you flagged. The `resolveSingleTexts` / CLOVA single-text dispatch is covered by the same call, so both paths clamp; non-string (structured/native modality) items still pass through untouched, as the helper intends.

On the CI blind spot you identified — the gap was that the helper's own unit test exercised the helper, not the route. `tests/unit/embeddings-handler.test.ts` now asserts the **captured outbound body** (`captured.body.input` / `calls[0].body`) through `handleEmbedding`, so a regression in `buildUpstreamBody` fails the suite. That's the route-level assertion that was missing.

Changelog fragment added: `changelog.d/fixes/13004-opencode-embedding-clamp-vision-fragment.md` — and it says so plainly, "the `clampEmbeddingStringInput` helper existed but was never wired into `buildUpstreamBody`/CLOVA single-text dispatch".

Tests: `embeddings-handler.test.ts` 10/10, `embeddings-input-clamp.test.ts` 4/4. Rebased onto `cde49c937`; `mergeable=true`.

PR #13006, opened 2026-09-08
fix(combo): stop quota-weighted routing onto out-of-credit connections

HouMinXi:
## Problem

A combo on `quota-weighted` kept dispatching to a connection that had already
answered `402` for a depleted balance.

Observed on a live deployment: the connection returned

```
402 Grok Build usage balance exhausted
```

while its stored quota snapshot still read `remaining=1%`, `is_exhausted=0`,
last refreshed five hours earlier. The router had no reason not to pick it
again, so it did.

Two separate gaps let that happen.

**A 402 never reached the quota cache from the combo path.**
`executeTargetAttempt` already classified the response through
`isQuotaExhaustionResponse` and then dropped the result on the floor. The only
writer into the cache was `markAccountExhaustedFrom429`, reached from
`chat.ts` on a 429 and nowhere else. So the connection kept whatever remaining
percentage its last snapshot happened to hold.

**Snapshot age was unbounded.** An observation from five hours ago carried the
same weight as one from five minutes ago. That is how a snapshot claiming 1%
outlived the balance it described.

## Change

`markAccountExhaustedFromCredits` records a 402 the same way a 429 is
recorded, called from the combo attempt path where the classification already
existed. The connection stays active - credits come back, and a later refresh
or window reset clears the mark through the paths that already clear a 429.

`orderTargetsByQuotaWeighted` now consults the live snapshot when it builds the
pools. A 402-marked connection scores zero. A snapshot older than
`QUOTA_WEIGHTED_MAX_SNAPSHOT_AGE_MS` (10 minutes, twice the active refresh
cadence so one missed tick is absorbed) drops from the primary pool to the
secondary one.

Staleness means unknown, not empty. A stale connection is still routed to when
nothing fresher has room, and an all-stale set still returns a full ordering
rather than nothing. The soft floor is untouched: default 1%, and
`0 < remaining <= floor` still lands in the secondary pool.

Nothing is ever deactivated or deleted. `isQuotaExhaustedForRequest` is
deliberately left alone as an eviction test - with
`DEFAULT_QUOTA_THRESHOLD_PERCENT` at 99 it would empty the secondary pool.

## Tests

`tests/unit/combo/quota-weighted-stale-402.test.ts` - 11 tests.
`tests/unit/combo/**` - 253 passing. `typecheck:core` clean.

Each guard was verified by injection: deleting the 402 call site, inverting it
to 429, neutering `markAccountExhaustedFromCredits`, dropping the
`marked === 0` override, removing the staleness filter from the primary pool,
and relaxing the age comparison from `>` to `>=` each turn a specific test red.

The first attempt at the 402 test called the helper directly and stayed green
with the call site deleted, which is worth noting: the tests that cover it now
drive a real 402 through `executeTargetAttempt`.


HouMinXi:
Rebased onto #13221 (`f6569f387`) so the two overlapping files land in order.

Conflict was only `orderTargetsByQuotaWeighted` pool selection. Kept the 402/staleness split (`hasRoom` + `stale`); #13221 eligibility filters stay in `expandTargetsByQuotaAwareConnections`.

The 402 fixture tests now insert live connection rows. Invented pin UUIDs no longer reach the weighted draw after the eligibility filter.

HEAD: `40bfcd86b`. Contributor only; please do not merge from this comment.

HouMinXi:
Pushed `01afa43f3`. Registered `agy-quota-exhaustion-threshold`, `quota-connection-eligibility`, and `quota-weighted-stale-402` in `stryker.conf.json` `tap.testFiles`. Local `check:mutation-test-coverage --strict` is clean on this tree.

Remaining Docs / Fast Quality / ESLint / unit-shard red matches current `release/v3.8.51` and is not from this diff.

PR #6731, opened 2026-07-09
fix(resilience): apikey-provider 429s honor explicit quota-exhausted text (#6638)

diegosouzapw:
Closes #6638

## Root cause
Ollama Cloud is registered as an `apikey`-category provider. `checkFallbackError()`'s `shouldPreserveQuotaSignalsFor429()` gate only preserved body-text quota classification (daily/monthly/weekly quota-exhausted detection) for `oauth`-category providers — every apikey-category 429 fell through to the generic short rate-limit cooldown regardless of what the error body said, and `parseRetryFromErrorText()` had no support for day-granularity reset hints ("Your quota will reset in 3 days.") in the first place. A genuine multi-day quota exhaustion was misclassified as `rate_limit_exceeded` with a ~3s cooldown, so combo routing (including LKGP) retried the "exhausted" account on the very next request.

## Fix
- `shouldPreserveQuotaSignals()` (new `open-sse/services/quotaResetParsing.ts`) lets an explicit quota-exhausted signal in the body (`looksLikeQuotaExhausted`) override the apikey-category default, in addition to the existing oauth-category always-preserve behavior.
- `parseDayGranularityResetMs()` (same file) adds day-granularity reset-hint parsing, wired into `parseRetryFromErrorText()` alongside the existing Xh/Ym/Zs parsing.
- `accountFallback.ts` net line count in the frozen file is negative (removed the old local `shouldPreserveQuotaSignalsFor429` helper and its `getProviderCategory` == "oauth"-only 429 gate in favor of the shared, testable helper) — passes `check-file-size`.

## TDD evidence
Repro test (from the approved triage plan) — RED before, GREEN after:

```
$ node --import tsx/esm --test tests/unit/issue-6638-ollama-quota.test.ts
# RED (before fix): reason: 'rate_limit_exceeded', cooldownMs: 3000
# GREEN (after fix): reason: 'quota_exhausted', cooldownMs: 259200000 (3 days)
✔ #6638: Ollama Cloud weekly-quota-exhausted 429 must NOT get a short generic rate-limit cooldown
```

[...]

Issue #6638, opened 2026-07-08
[BUG] OmniRoute doesn't respect exhausted quotas


tenshiak:
### OmniRoute Version

3.8.46

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Description

With Ollama cloud (I don't know if it also applies to other providers), when session or weekly quota is exhausted and the combo uses LKGP strategy, it still tries to use the exhausted account, throwing a 429 error and then switching to another one. When the other account is also exhausted, it tries all the accounts again, even though the quota is exhausted, like it's ignoring it completely. Maybe because one of them shows green? 

<img width="532" height="153" alt="Image" src="https://github.com/user-attachments/assets/65b34241-2e98-4d39-abc0-b479cc5f4677" />

### Steps to Reproduce

1. Create a combo, assign 3 accounts to it and set LKGP as a strategy
2. Exhaust two of them, before the third one starts, it will try to use the first one from the list even if the weekly quota was already exhausted

### Expected Behavior

For OmniRoute to respect the quota if it is enabled for the account and not retry it on account switches.

### Actual Behavior

It ignores the quotas completely

### Test Impact

Unsure

### Error Logs / Output

```shell

```

diegosouzapw:
Thanks for reporting! Fixed in `release/v3.8.47` via #6731 (already merged into the active development branch — feel free to pull and test it now). It will ship in the next release (v3.8.47).

PR #13008, opened 2026-09-08
fix(sse): thread errorText into shouldMarkAccountExhaustedFrom429

Rick7C2:
## What's wrong

`shouldPreserveQuotaSignals(provider, errorText)`
(`open-sse/services/quotaResetParsing.ts:19-26`) gained its `errorText` parameter in the
#6638 fix (#6731), but only **one of its two call sites** was updated in that same commit.
`checkFallbackError()` passes the upstream body; `shouldMarkAccountExhaustedFrom429()`
(`open-sse/services/accountFallback.ts:1002`) still called it with the provider alone.

With `errorText` undefined the helper's `Boolean(errorText) &&
looksLikeQuotaExhausted(errorText)` branch can never be true. So for every
apikey-category provider without per-model quotas, the connection was never marked
quota-exhausted — even when the upstream body explicitly said a long-window cap was hit.

3.8.47 is the release that first shipped the asymmetry. Verified by content rather than
tag ancestry: at `v3.8.46` the file does not exist; at `v3.8.47` both the 2-arg
declaration and the 1-arg call site appear. `git log -S 'shouldPreserveQuotaSignals(provider) &&'`
over full history returns exactly two commits — the one that introduced it and this one.
No intervening commit ever touched the call site, and there is no comment or rationale
anywhere in the log suggesting the one-arg call was deliberate.

## The change

`errorText?: string | null` added as an optional 5th parameter to
`shouldMarkAccountExhaustedFrom429`, forwarded to `shouldPreserveQuotaSignals(provider,
errorText)`. The `src/sse/handlers/chat.ts` call site passes the `errorStr` already in
scope (`result.rawMessage ?? result.error`).

Optional and additive — every existing caller keeps its behavior. Plain rate limits
(`Rate limit exceeded, retry in 20s`, `Too Many Requests`) still fall through to the
short generic cooldown, and the `failureKind` early return still dominates: a 429 the
classifier already called `rate_limit`/`transient` never reaches the changed line.

**Blast radius, measured across the full registry** (273 providers × 12 non-quota bodies
× 3 models × 2 `failureKind` values, patched vs. unpatched):

- non-quota bodies: **0 divergences**
- on an explicit quota body: **129 of 273 flip `false → true`, 0 flip `true → false`**
- **0 of the 41 oauth-category providers move**, across 5 distinct quota bodies

The change is monotone — no provider loses quota marking.

## Scope notes (things this deliberately does *not* claim)

- **It does not cover every quota-exhausted body.** A bare, unwrapped `usage limit
  reached` is classified `quota_exhausted` by `isSubscriptionQuotaText` but is not matched
  by `QUOTA_PATTERNS`, so `looksLikeQuotaExhausted` is false and the connection still is
  not marked. That is a **pre-existing** inconsistency between the two classifiers — this
  patch neither introduces nor worsens it, and strictly improves the case. Left for a
  separate general fix rather than widened into this one.
- **`ollama-cloud` is not the demonstrating provider**, even though #6638 was reported
  against it: `hasPerModelQuota("ollama-cloud")` is true, so it short-circuits before the
  quota-signal check regardless of this fix. The providers this actually changes are
  apikey-category ones without per-model quotas (`openai`, `anthropic`, `deepseek`) —
  which is what the tests use.

## Tests

`tests/unit/quota-signal-errortext-threading.test.ts` — 5 cases. Validated by
**mutation, per clause**, not whole-patch revert:

| mutant | result |
|---|---|
| revert only `accountFallback.ts` to `shouldPreserveQuotaSignals(provider)` | **1 of 5 fail** ✅ |
| revert only the `chat.ts` call site (drop the `errorStr` arg) | **1 of 5 fail** ✅ |

The second guard exists because `errorText` is an *optional* parameter: dropping it at
the only production call site is neither a type error nor a helper-test failure. Without
it, the half of the patch that does anything in production could be reverted — or lost in
a refactor — with the whole suite green. That is the same failure mode this PR fixes, one
level up.

`handleSingleModelChat` is not exported, so the call cannot be driven or spied without
widening the production surface. The guard asserts at the source level instead
(precedent: `tests/unit/api-key-provider-quota-bypass-scope.test.ts`), and parses the
argument list rather than regex-matching formatted text, so a Prettier reflow cannot
cause a false failure or a false pass. It also pins that there is exactly one call site,
and what `errorStr` holds.

Green: `quota-signal-errortext-threading` 5/5 · `account-fallback-service` 96/96 ·
`issue-6638-ollama-quota` 1/1 · `classify429` 32/32. `npm run typecheck:core` → exit 0.

## ⚠️ Formatting churn — 35 lines, pre-existing, forced by the hook

**`open-sse/services/accountFallback.ts` (28 lines) and `src/sse/handlers/chat.ts` (7
lines) are already prettier-non-conformant at the base commit `9d1a896c`** — measured via
`--stdin-filepath` against the pristine tree, independently reproduced by review. The
husky `lint-staged` hook reformats any file it stages, so committing with hooks enabled
necessarily pulls that drift in.

The alternative was `--no-verify`, which the repo's own rules ban. Honoring the hook was
the deliberate choice. **The actual fix is 4 lines**; the reformatted lines are untouched
logic. Happy to split them into a separate commit or drop them if the maintainers prefer.

## ⚠️ base-red inherited

Both verified on the pristine base, not on this branch:

- `check:changelog-integrity` fails on `changelog.d/fixes/reset-aware-model-family.md`
  (missing the leading `- ` bullet) — someone else's fragment, left alone. This PR's own
  fragment passes.
- `npm run lint` exits non-zero on pristine HEAD ("suppressions left that do not occur
  anymore"), with zero actual lint errors.

---


diegosouzapw:
Good find — confirmed the gap myself: `checkFallbackError()` already threads `errorText`
into `shouldPreserveQuotaSignals`, but the `shouldMarkAccountExhaustedFrom429` call site
in `chat.ts` never got updated after #6731 added that parameter. Your 5 new tests
(including the "no errorText → unchanged behavior" regression guard) all pass against the
real function. One small note: a good chunk of the `accountFallback.ts` diff is Prettier
reformatting of unrelated nearby lines (import wrapping, a `console.warn` call, some
multi-line conditionals) — not required, but reverting those hunks would make the diff
easier to review. Otherwise this looks ready to merge.

diegosouzapw:
Thanks @Rick7C2 — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

PR #13011, opened 2026-09-08
fix(db): self-heal under sustained memory pressure, VACUUM gate, WAL housekeeping

HouMinXi:
## What happened

On 2026-09-07 the production container stopped serving for 36 minutes and then wedged entirely until a manual restart. What I found on the host:

- The container's cgroup working set sat at the 5 GiB cap, so admission control returned 503 for every request, including `/api/health/ping`.
- The V8 heap was only ~300 MB. RSS was dominated by ~1.9 GB of glibc main-heap (native) memory, on a host with 7.5 GB total.
- `storage.sqlite-wal` had grown to 2.98 GB, the size of the main DB. That is the high-water mark a VACUUM leaves behind: the startup cleanup ran VACUUM after deleting ~125 rows, rewriting the whole 3 GB database into the WAL. Auto-checkpoint (default 1000 pages) never shrinks the file, and the only TRUNCATE ran every 6 hours.

better-sqlite3 is synchronous, so once the WAL is huge and memory is tight, every query blocks the event loop long enough to stall the process.

## Changes

1. Self-restart circuit (`open-sse/utils/resourcePressure.ts`). When pressure stays critical for `OMNIROUTE_PRESSURE_SELF_RESTART_AFTER_MS` (default 120s), the process logs its counters and exits, so the supervisor (systemd `Restart=always`, Docker restart policy) brings back a clean process in seconds. Off by default (`OMNIROUTE_PRESSURE_SELF_RESTART`). An unref'd 1-10s driver re-arms sampling so the circuit still works when no requests arrive, and a throwing exit hook re-arms the circuit instead of bricking it. Entering critical now logs a structured one-line diagnostic (heap/RSS/cgroup/PSI) whether or not self-restart is enabled, so the next incident starts with evidence.

2. Post-cleanup VACUUM gate (`src/lib/db/cleanup.ts`). VACUUM rewrites the whole database file; the 6h scheduler used to run it after every cleanup, including ones that freed ~100 rows. It now runs only when at least `OMNIROUTE_VACUUM_MIN_DELETED_ROWS` (default 1000) rows were freed. `0` means always vacuum. The weekly scheduled VACUUM from the Storage page still reclaims space on its own cadence.

3. WAL housekeeping (`src/lib/db/core.ts`). A 5-minute `wal_checkpoint(PASSIVE)` keeps the pending-frames count small so TRUNCATE never copies a multi-GB backlog on the main thread (`OMNIROUTE_WAL_PASSIVE_INTERVAL_MS`, `0` disables). If the WAL file exceeds `OMNIROUTE_WAL_GUARD_MAX_MB` (default 256), the tick escalates to TRUNCATE immediately instead of waiting for the 6h tick. The 6h TRUNCATE now logs duration, WAL size before/after, and busy/log/checkpointed frame counts.

## Tests

- 50 unit tests across the six affected suites pass; `typecheck:core` is clean.
- New tests cover the sustained-critical exit and its traffic-free driver, re-arm on a throwing exit hook, dispose cleanup, the VACUUM gate thresholds (including the 0 case), and the WAL scheduler wiring.
- Each new behavior was verified by bug injection: break the condition, watch the test fail, revert, watch it pass.

## Notes

- Self-restart is opt-in; default behavior changes only in the WAL/VACUUM housekeeping and one extra log line on critical transitions.
- All new knobs are documented in `docs/reference/ENVIRONMENT.md`.


HouMinXi:

PR #13017, opened 2026-09-08
fix(auth): probe a pinned inactive connection after quota top-up

### LINKED ISSUE #12874: fix(resilience): Provider connection disabled on quota exhaustion stays disabled after top-up; explicit model routing does not trigger probe/auto-recovery
# [BUG] Provider connection disabled on quota exhaustion stays disabled after top-up; explicit model routing does not trigger a health probe or auto-recovery

## Environment
- **OmniRoute version:** 3.8.50 (Docker, `diegosouzapw/omniroute:latest`)
- **Deployment:** Docker Compose, port 20128
- **OS:** Ubuntu 24.04
- **Setup:** Claude Code client → cc-switch → OmniRoute → SiliconFlow (paid relay, custom `openai-compatible-chat` style connection via provider `siliconflow`)

## Summary
When a provider connection is automatically disabled after its upstream quota/balance is exhausted, re-enabling it requires **manual Dashboard action even after the balance has been topped up**. Crucially, **explicitly routing a request to that exact provider/model does NOT trigger a health probe or auto-recovery** — the request is short-circuited at the connection gate before any upstream attempt, so OmniRoute never learns the balance has been restored.

## Reproduce
1. Have a paid connection (e.g. SiliconFlow `main`, provider `siliconflow`) with an API key whose balance runs out.
2. OmniRoute disables the connection (observed DB state: `is_active=0`, while `test_status='active'` and `backoff_level=0`).
3. Top up the upstream balance (money added on SiliconFlow side).
4. Send a request **explicitly targeting that provider/model** from a client (e.g. Claude Code with cc-switch pinned to `siliconflow/...`).
5. **Actual:** request never reaches upstream — 0 probe, no auto-recovery, still erroring until the operator manually re-enables via Dashboard or API.

Observed connection row while broken:
```
provider=siliconflow  is_active=0  test_status='active'  backoff_level=0  rate_limited_until=NULL
```
(`test_status='active'` means a health check would pass — the connection is healthy but the disabled flag blocks all routing.)

Manual recovery works and is instantaneous:
```
PUT /api/providers/{id}  {"isActive": true}   → connection re-enabled
POST /api/providers/{id}/test                  → {"valid": true, "latencyMs": 126}
```

## Root cause analysis
Two layers are involved and neither auto-recovers for this scenario:

1. **Connection-level disable (`is_active=0`)** — the highest-priority pre-dispatch gate. Any request targeting the connection short-circuits *before* routing/probing logic runs, so a subsequent request (even one that explicitly names this provider/model) can never observe that upstream is healthy again. There is no "probe-on-demand when explicitly selected" path.

2. **Quota-aware auto-recovery is gated on a per-provider quota fetcher** — as documented in #10877, quota-aware routing (`reset-aware`/`reset-window`) silently degrades to plain `priority` when the provider has no registered quota fetcher. SiliconFlow (and most custom/paid relays) have no fetcher, so nothing polls the balance and nothing notices a refill. Issue #9603 documents the same class for qwen-cloud-token-plan ("exhausted accounts do not recover after reset").

3. **Top-up is an invisible event to OmniRoute** — there is no webhook, no periodic balance poll for fetcher-less providers, and no "retry disabled connection on explicit request" heuristic. The only recovery paths are manual (Dashboard/API).

## Expected behavior
- An **explicitly routed request** (client pins `provider/model` that maps to a disabled-but-not-banned connection) should **attempt one upstream probe/health check** instead of short-circuiting — if the probe succeeds (e.g. balance restored), the connection should be re-activated automatically and the request served. This mirrors how a health-check-passing connection should behave.
- Alternatively/additionally: a **periodic re-enable probe** for disabled connections whose `test_status` was `active` at disable time (i.e. disabled for quota/exhaustion, not for credential failure), so top-ups are picked up within the probe interval without manual action.

## Suggested direction
- In the pre-dispatch gate, distinguish **quota/exhaustion-driven disables** (recoverable; `test_status` healthy, e.g. 402/insufficient-funds or explicit quota-exhausted marking) from **hard failures** (401/403 banned, expired). For the recoverable class, allow an explicit request to trigger a single upstream probe; on success, flip `is_active` back on (with a rate-limit to avoid probe storms).
- Optionally: surface a "disabled, awaiting top-up" state in the Dashboard with a one-click re-test, and/or a scheduler option to periodically re-test disabled-but-healthy connections.

## Notes
- Distinct from #9630 (false `ALL_ACCOUNTS_INACTIVE` on combo paths when healthy targets are skipped) and #10877 (quota-aware strategies silently degrade) — those cover combo routing; this issue is about **single-connection explicit routing never triggering a probe/recovery**.
- Related: #9603 (same class, qwen-cloud-token-plan specific — "exhausted accounts do not recover after reset").

HouMinXi:
## Summary

Fixes #12874.

A pin (`combo` step `connectionId` or `x-omniroute-connection`) of a quota-disabled row used to miss the active pool and fail closed. After a top-up the same pin still 404'd until someone flipped `isActive` by hand.

This change lets that **explicit pin** take the inactive row as a one-shot probe. Success on the request then sets `isActive` true. No pin, no probe. Dashboard deactivate stays off.

## Behavior

- Pin of a recoverable inactive row (`credits_exhausted`, `unavailable` after cooldown, `error`, empty, `expired` without a blocklisted error type) becomes the only candidate for that request. Sibling accounts stay out of the pool.
- Storm gate: 60s per connection, in-process Map. Gate not elapsed → fail closed, no upstream.
- 2xx / request success: `clearAccountError` then `isActive: true`. OpenRouter `:free` models do not flip the paid lock.
- `clearAccountError` still does not write `isActive`.

Out of scope: automatic wake without a pin; combo live-test `allowSuppressedConnections` threading (that flag is not in the success callback).

## Test plan

- [x] `tests/unit/explicit-inactive-probe-12874.test.ts` — recoverability P-1..P-14, storm Map, W-6 `:free`, W-3 / W-7 source guards
- [x] `tests/unit/explicit-inactive-probe-w2.test.ts` — SQLite pin of inactive + `credits_exhausted` returns credentials; successful probe sets `isActive`
- [x] `tests/unit/forced-connection-fallback.test.ts` — deactivated-missing fixture still fail-closed; pin-in-pool is not missing
- [x] 41/41 on those three files after rebase onto `fcc2dcd1a`

## Notes for CI

`check-mutation-test-coverage` still names two files this branch does not touch (`combo/execute-target-gates.test.ts`, `combo-pin-implicit-allowlist.test.ts`). Those are on tip. This PR added `tests/unit/explicit-inactive-probe-w2.test.ts` to `stryker.conf.json` `tap.testFiles`.


HouMinXi:
Looked at `fa29448a3` on `auth.ts`, `explicitInactiveProbe.ts`, `chatHelpers.ts`, and the exported blocklist in `connectionRecovery.ts`.

The pin path now loads the inactive row as the only candidate for that request. No pin still fail-closes. Success runs `clearAccountError` then sets `isActive`. The deactivated-missing fixture in `forced-connection-fallback.test.ts` stays true; `tests/unit/explicit-inactive-probe-w2.test.ts` goes red if the one-row rewrite or the terminal-status skip is dropped.

A few notes that are not defects:

- `expired` does not consult `rateLimitedUntil`; only `unavailable` does. That matches the recoverability function as written.
- The success callback does not pass `allowSuppressedConnections`. That flag is not in `executeChatWithBreaker`. Combo live-test can still re-enable a probed pin; follow-up if that matters.
- Concurrent pins of the same id share one in-process Map write on the same turn, before the next await.

No further code change from this look.

HouMinXi:

PR #13035, opened 2026-09-08
fix(db): resolve sql-wasm.wasm across global and hoisted layouts (#12960)

### LINKED ISSUE #12960: [bug] Published package: sql-wasm.wasm missing → fatal boot when better-sqlite3 is absent (global install, Node 24)
## Summary

A user on Discord (EN) reported a **fatal boot crash** on a fresh global install
(`npm install -g omniroute`, Node.js **v24.20.0**, root): the SQLite driver cascade falls
through to the `sql.js` (WASM) adapter, and the packaged `sql-wasm.wasm` is not present at
either path the adapter checks — so the process dies during the instrumentation hook.

## Reported output

```
[STARTUP] Fatal: Database driver initialization failed: [sqljsAdapter] Packaged sql.js runtime is incomplete: sql-wasm.wasm was not found. Checked:
/root/.nvm/versions/node/v24.20.0/lib/node_modules/omniroute/dist/node_modules/sql.js/dist/sql-wasm.wasm
/root/.nvm/versions/node/v24.20.0/lib/node_modules/omniroute/dist/.next/standalone/node_modules/sql.js/dist/sql-wasm.wasm
[STARTUP] Fatal: instrumentation hook failed during boot: [sqljsAdapter] Packaged sql.js runtime is incomplete: ...
Failed to prepare server Error: An error occurred while loading instrumentation hook: ...
⨯ unhandledRejection: Error: An error occurred while loading instrumentation hook: ...
```

Error source: `src/lib/db/adapters/sqljsAdapter.ts:50`.

## Analysis

Two independent conditions stack:

1. **`better-sqlite3` is absent.** npm 11 (Node 24+) skips install scripts for
   `optionalDependencies` by default, and `better-sqlite3` needs `node-gyp rebuild`. This is
   already documented — `docs/guides/TROUBLESHOOTING.md` → *"npm v11+: `better-sqlite3` not
   installed (Cannot find module)"* — and it is what pushes the cascade
   (`better-sqlite3` → `node:sqlite` → `sql.js`) down to the WASM adapter.
2. **The published package does not ship a usable `sql-wasm.wasm` at the checked paths.** The
   fallback that is supposed to make OmniRoute runnable without a native binary is therefore
   not actually available in the published artifact — the fallback fails as hard as the driver
   it was meant to cover.

Condition 2 is the bug: the sql.js path is advertised as the no-native-dependency fallback
(#2447), was trace-included for the standalone bundle in #4839 and re-verified in #9044 /
#11266, but the global-install layout above (`<pkg>/dist/...`) resolves to neither candidate.

## Impact

A first-run global install on Node 24 crashes at boot with no working path forward until the
user manually rebuilds `better-sqlite3`. Nothing in the fatal message points at the actual
remedy.

## Suggested direction

- Verify the published tarball layout actually contains `sql-wasm.wasm` at (one of) the paths
  `sqljsAdapter` probes for a **global** install, not only for the standalone server layout —
  the existing `check:pack-artifact` / pack-boot gates apparently do not cover this shape.
- Widen the candidate-path resolution (e.g. resolve `sql.js` through `require.resolve` rather
  than fixed relative paths).
- Make the fatal message actionable: name the `npm rebuild better-sqlite3` remedy and link the
  troubleshooting section, instead of only listing the two paths that were checked.

## Workaround given to the reporter

```bash
cd $(npm root -g)/omniroute/app
npm rebuild better-sqlite3
omniroute
```

Reported via the OmniRoute Discord (EN) support channel.

HouMinXi:
## Summary

Fixes #12960.

On Node 24 with npm 11, install scripts for `optionalDependencies` are skipped by default. When `better-sqlite3` cannot compile or is unavailable, the fallback WASM runtime (`sql.js`) fails to boot during global npm installs (`npm install -g omniroute`).

### Root Cause
1. `omniroute serve` launches the standalone server child process with `cwd: APP_DIR` (`<packageRoot>/dist`).
2. `sql.js` is a package dependency, so npm places it at `<packageRoot>/node_modules/sql.js`, not under `dist/node_modules/sql.js`.
3. `sqljsAdapter.ts` only probed `path.join(process.cwd(), "node_modules", "sql.js", "dist", "sql-wasm.wasm")` and `.next/standalone/...`, omitting the parent directory (`<cwd>/..`).
4. When `better-sqlite3` was absent and `postinstall` had not copied `sql.js` into `dist`, the server crashed fatally on boot with an unhelpful error.

### Changes
- Added `<cwd>/../node_modules/sql.js/dist/sql-wasm.wasm` to the candidate path list to handle global CLI installs.
- Added dynamic module resolution via `nodeModule.createRequire` across standard entrypoint anchors (`process.cwd()`, `process.argv[1]`), leveraging `sql.js`'s official exports (`./dist/*`).
- Added support for `OMNIROUTE_SQLJS_WASM_PATH` override with atomic `fs.statSync` validation (ensures valid non-empty file, rejects empty strings/directories, resolves relative paths to absolute).
- Refined the fatal error message to provide actionable troubleshooting remedies (pointing to `npm rebuild better-sqlite3` and `docs/guides/TROUBLESHOOTING.md`).
- Documented `OMNIROUTE_SQLJS_WASM_PATH` in `docs/reference/ENVIRONMENT.md`.
- Added 10 unit tests in `tests/unit/sqljs-wasm-resolution-12960.test.ts` covering global layouts, env overrides, directory/empty-file rejections, and resolution error propagation.

### Verification
- `tests/unit/sqljs-wasm-resolution-12960.test.ts` (10/10 PASS).
- Existing adapter & boot suites: `sqljsAdapter.test.ts`, `driverFactory.test.ts`, `pack-boot-runtime-paths.test.ts`, `check-pack-boot.test.ts` (66/66 PASS).
- `npm run typecheck:core` passes with 0 errors.
- Verified bug injection: removing parent candidate path / require.resolve fails the global install layout test; removing size check fails 0-byte validation test.
- Code review: completed 7 rounds of MCP code-forge review on product commit `c271f37eb` with **0 CONFIRMED findings**.


HouMinXi:

PR #12498, opened 2026-09-02
fix(api): thread X-OmniRoute-Fallback-Attempts through chat completions (#12339)

princeaggarwal2005:
## Problem

As reported in #12339 by @diegosouzapw:

On `/v1/chat/completions` responses (both streaming `stream: true` and non-streaming `stream: false`), the `X-OmniRoute-Fallback-Attempts` HTTP header is absent (or defaults to `0`), even when a combo failover to a non-primary target occurred before the stream/response opened.

While `X-OmniRoute-Provider` and `X-OmniRoute-Model` update correctly to the winning provider, the fallback attempts counter was completely missing from chat responses.

## Root Cause

1. **Missing parameter in response-header builders:** Both `assembleStreamingResponseHeaders` and `buildNonStreamingResponseHeaders` built metadata headers without accepting or passing `fallbackAttempts` into `buildStreamingResponseHeaders` / `attachOmniRouteMetaHeaders`.

2. **Missing forwarding in `chatCore.ts`:** `handleChatCore` did not accept `fallbackAttempts` or pass it to either header builder.

3. **Dropped in the dispatch chain:** While `open-sse/services/combo.ts` properly calculates `fallbackCount`, it was not passed through `handleSingleModel` (`chat.ts`), `executeChatWithBreaker` (`chatHelpers.ts`), or `dispatchChatWithAffinityEviction`.

### 1. Response Header Builders

- **`open-sse/handlers/chatCore/nonStreamingResponseHeaders.ts`**
  - Added `fallbackAttempts?: number` to the `args` type.
  - Forwarded `fallbackAttempts: args.fallbackAttempts` to `deps.attachOmniRouteMetaHeaders`.

- **`open-sse/handlers/chatCore/streamingResponseHeaders.ts`**
  - Added `fallbackAttempts?: number` to the `args` type.
  - Forwarded `fallbackAttempts: args.fallbackAttempts` to `buildStreamingResponseHeaders`.

### 2. Core Handler & Dispatch Chain

- **`open-sse/handlers/chatCore.ts`**
  - Added `fallbackAttempts = 0` to `handleChatCore` parameters.
  - Forwarded `fallbackAttempts` to `buildNonStreamingResponseHeaders` and `assembleStreamingResponseHeaders`.

- **`src/sse/handlers/chatHelpers.ts`**
  - Added `fallbackAttempts = 0` to `executeChatWithBreaker` options.
  - Forwarded `fallbackAttempts` into `handleChatCore`.

[...]

PR #13038, opened 2026-09-08
fix(api): thread X-OmniRoute-Fallback-Attempts through combo chat (#12339)

### LINKED ISSUE #12339: fix(api): X-OmniRoute-Fallback-Attempts always 0 on chat completions (streaming and non-streaming)
### Summary

On streaming (`stream: true`) `/v1/chat/completions` responses, `X-OmniRoute-Fallback-Attempts` is absent (defaults to `0`) on the initial SSE header set, even when a combo failover to a non-primary target actually happened before the stream opened. The same request with `stream: false` reports the correct attempt count. `X-OmniRoute-Provider`/`X-OmniRoute-Model` **do** update correctly to whoever actually answered on the streaming path — only the fallback-attempts counter is missing.

### Root cause (confirmed in code)

`assembleStreamingResponseHeaders()` (`open-sse/handlers/chatCore/streamingResponseHeaders.ts:14`) builds the streaming meta headers by calling `buildStreamingResponseHeaders()` with an explicit object literal that only passes `provider`, `model`, `cacheHit: false`, `latencyMs: 0`, `usage: null`, `costUsd: 0`, `strategy` — it never forwards `fallbackAttempts`. The sole call site, `chatCore.ts:5523`, doesn't pass it either. `buildOmniRouteResponseMetaHeaders()` (`src/domain/omnirouteResponseMeta.ts`) defaults `fallbackAttempts` to `0` when omitted, so the header comes out `0`/absent on every streaming response regardless of what actually happened upstream.

The non-streaming path (`nonStreamingResponseHeaders.ts`) threads the real `fallbackAttempts` value through, which is why `stream: false` reports it correctly.

### Repro

1. Configure a combo where the primary target fails (bad key / unreachable) and a fallback target succeeds.
2. `POST /v1/chat/completions` with `stream: false` → `X-OmniRoute-Fallback-Attempts: 1` (correct).
3. Same request with `stream: true` → header absent/`0`, despite `X-OmniRoute-Provider` correctly showing the fallback provider.

### Impact

A client that only reads response headers can tell which provider answered a streaming call, but cannot tell whether that was the first-choice target or a failover — any "answered via fallback" UI signal is dead on the streaming path.

### Expected

Streaming and non-streaming responses expose the same routing-outcome headers; `X-OmniRoute-Fallback-Attempts` should be >= 1 on the streaming response whenever a failover occurred.

### Note on `X-OmniRoute-Fallback-Reason`

There is no `X-OmniRoute-Fallback-Reason` header anywhere on the `/v1/chat/completions` path (streaming or non-streaming) — that's not a regression, it never existed there. A similarly-named `X-Routing-Fallback-Reason` exists, but only on the separate `/api/v1/relay/chat/completions` relay-routing endpoint (`src/app/api/v1/relay/chat/completions/route.ts`), which is a different backend. Worth a follow-up feature request if a reason code on the main chat path is wanted, but it's out of scope for this bug.

---
Found and independently reproduced while building [OmniHarness](https://github.com/shipking-ai/omniharness) — reported by @shipking-ai in discussion #12303, code root-caused there by @diegosouzapw.

HouMinXi:
Combo chat completions never set `X-OmniRoute-Fallback-Attempts`. Image, video, and speech combos already do. Both streaming and non-streaming chat assemble OmniRoute meta headers, but `fallbackAttempts` never left the combo loop.

Reported in #12339. #12498 tried to thread the count through the old `combo.ts` god file and is now `CONFLICTING` after the executeTarget / round-robin split.

## Change

Stamp the ordered index (or round-robin offset) onto the dispatched target at the gate, then pass that field through `handleSingleModel` → `executeChatWithBreaker` → `handleChatCore` → the streaming and non-streaming header builders.

Do not copy the combo loop's `fallbackCount` at dispatch. That counter only increments after a failed or skipped leg, so the second target still sees `0` when it is sent. Index / offset is the number of prior legs already attempted.

`executeTargetAttempt` already spreads `targetForAttempt`, so it did not need a new stamp (and it is already at the file-size freeze).

## Tests

- Header builders forward a non-zero count and omit the key when the arg is missing
- Live `X-OmniRoute-Fallback-Attempts` is present when count > 0 and absent when 0 (the meta helper already strips 0)
- Inner gates stamp `0` then `1` from the ordered-target index
- Nested runtime units stamp from the unit index
- Last-resort compat fallback stamps from the rejected-target index
- `executeTargetAttempt` keeps a stamped count on the `handleSingleModel` target

42/42 on those files. `npm run typecheck:core` clean.

Closes #12339


HouMinXi:
Looked at 35b5ebfd9 (the product commit; the changelog commits on top are docs-only).

The combo loop's `fallbackCount` is not the value to copy at dispatch. It only increments after a failed or skipped leg, so the second target still sees 0 when it is sent. This PR stamps the ordered index / round-robin offset onto the target at the gate instead, then threads that field through chat into the header builders.

`executeTargetAttempt` already spreads `targetForAttempt`, so it did not need a new stamp.

A few notes from reading the diff that I do not think need a code change:

- `ExecuteChatWithBreakerOptions` is `{ [key: string]: any }`, so adding `fallbackAttempts` on the destructure is legal. `npm run typecheck:core` is clean on this SHA.
- Spreading `{ fallbackAttempts: 0 }` into the meta helper is fine. `buildOmniRouteResponseMetaHeaders` already omits the header when the count is 0 (`omnirouteResponseMeta.ts`, `toNonNegativeInteger` then `if (attempts > 0)`). Tests lock both layers: builders omit the key when the arg is missing, and they omit the live header when the count is 0.

Lock: `tests/unit/combo/execute-target-gates.test.ts` (ordered-target index 0 then 1) and `tests/unit/omniroute-decision-header.test.ts` (header present when count > 0, absent when 0). Those go red if the stamp at the gate is dropped or if the builders stop forwarding the field.

#12498 is the earlier attempt on the pre-split `combo.ts` and is now conflicting. This PR is the replacement against current `release/v3.8.51`.

HouMinXi:

PR #13040, opened 2026-09-08
fix: FriendliAI 403 credit exhaustion misclassified as auth_error

turbolego:
Fix friendliai 'exhausted all your credits' 403 misclassification to quota_exhausted instead of auth_error.

FriendliAI returns HTTP 403 with body `{"detail":"You've exhausted all your credits..."} when free tier credits are depleted via Adaptive Rate Limits. This was being misclassified as AUTH_ERROR instead of QUOTA_EXHAUSTED, causing omniroute to not properly handle the credit exhaustion (treating it as a credential issue rather than a depleted free tier).

Add 'exhausted all your credits' to CREDITS_EXHAUSTED_SIGNALS.


diegosouzapw:
Nice catch and a clean, minimal fix — traced `classifyErrorText`/`classifyError` myself
and confirmed FriendliAI's 403 body falls straight through every quota/credits check to
the generic 403→AUTH_ERROR status fallback without this signal. The two CI "failures" on
this PR (ESLint + unit tests fast-path) are the base branch's pre-existing 4 ESLint errors
in `accountFallback.ts` (unrelated unused vars), not anything from your 2-line change —
confirmed locally, and the relevant test suite (134 tests) is fully green. Only thing
missing before merge is a changelog fragment under `changelog.d/fixes/` — could you add
one? A small dedicated unit test for the FriendliAI body would be a nice bonus but isn't
required.

diegosouzapw:
Thanks @turbolego — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

PR #13041, opened 2026-09-08
fix(resilience): trip provider cooldown on credit-exhausted signals beyond 429

turbolego:
## Problem
The provider breaker classified credit/billing exhaustion **only** when the upstream responded with HTTP 429. Providers also surface free-tier exhaustion as 400/401/402/403/5xx with a body that explicitly says credits are gone (e.g. "You have exhausted all your credits."). Those fell through to `transient`, so the provider breaker kept re-selecting and hammering the dead provider instead of dropping it for the long cooldown.

## Fix
Two changes in `src/shared/utils/classify429.ts`:

1. **Promote terminal credit signals on any 4xx/5xx to `quota_exhausted`.** `classify429` now checks the body for `TERMINAL_QUOTA_PATTERNS` first and returns `quota_exhausted` for any status >= 400 (including 429), instead of early-returning `transient` for every non-429 status. A generic quota body on a 5xx remains `transient` — only an explicit credits/billing-exhausted signal trips the long cooldown.

2. **Catch the "exhausted all your credits" word order.** The existing `/credit.*exhaust/i` requires "credit" *before* "exhaust"; real bodies like *"You have exhausted all your credits"* have them reversed and slipped through. Added `/exhausted all your credits/i` and `/have exhausted all your credits/i` to both `QUOTA_PATTERNS` and `TERMINAL_QUOTA_PATTERNS`.

## Scope
Matches the intent of the existing account-fallback `CREDITS_EXHAUSTED_SIGNALS` list (which already carries these phrases). This is a conservative extension: it only promotes a provider to the 1h `quota_exhausted` cooldown when the body explicitly states credits are depleted.

## Tests
Added to `tests/unit/classify429.test.ts`:
- 4xx/5xx (400/401/402/403/429/500/502/503) with a credits-exhausted body → `quota_exhausted`
- generic quota body on 5xx stays `transient` (no over-promotion)
- "exhausted all your credits" phrase variants detected by `looksLikeQuotaExhausted`

## Validation
- `tests/unit/classify429.test.ts`: 33/33 pass
- related circuit-breaker / antigravity / neuron-exhaustion tests: pass
- `npm run typecheck:core`: clean (the two `noimplicit:core` errors are pre-existing in `systemPrompt.ts`, untouched by this PR)


diegosouzapw:
The core idea here is good and your new tests pass — promoting a terminal credits/billing
signal to `quota_exhausted` regardless of HTTP status is the right fix. Before this can
merge though, two things need to come out of the diff: `skills/cli-tunnel/SKILL.md` (a
protected file with an unrelated 2-line change, looks like local skill regeneration
drift) and `config/quality/api-typecheck-baseline.json` (+1361 lines that look like a
stale-environment baseline dump, not a real baseline for this fix). Also — the diff
silently drops `organization TPD rate limit`, `TPD rate limit`, and `insufficient
balance` from `TERMINAL_QUOTA_PATTERNS`; was that intentional? If so a test covering a
429 with one of those signals plus a short upstream retry hint would help confirm the new
(rate_limit) behavior is what you meant. Happy to take another look once those are
addressed.

turbolego:
@diegosouzapw changes:

- Added changelog fragment documenting the issue and its resolution in the project changelog.
- Fixed SSE handling to maintain TPD rate-limit and properly handle insufficient-balance terminal state in the classify429 function. This addresses stability issues with rate-limiting during SSE streams.
- Reverted unrelated changes that had drifted into the diff.

PR #13046, opened 2026-09-08
fix(oauth): read pollToken body once on non-JSON upstream responses (kimi-coding, github)

ysntony:
## Problem

The device-flow `pollToken` handlers for `kimi-coding` and `github` tried `response.json()` first and fell back to `response.text()` in the `catch`:

```ts
let data;
try {
  data = await response.json();
} catch (e) {
  const text = await response.text(); // ← always throws here
  data = { error: "invalid_response", error_description: text };
}
```

Once `.json()` rejects on a non-JSON body, the stream is already consumed, so the `.text()` fallback always throws `TypeError: Body is unusable: Body has already been read`. `pollToken` then rejects, `pollForToken` has no `try/catch`, and `POST /api/oauth/<provider>/poll` surfaces a generic 500 "Internal server error" — instead of the intended graceful `{ error: "invalid_response" }` payload that lets the UI report the upstream error and keep polling.

Non-JSON responses from the token endpoint are realistic when the OAuth upstream sits behind a CDN/anti-bot HTML error page or a proxy interstitial (`auth.kimi.com` in particular).

Minimal reproduction (plain Node, no repo code needed):

```js
const response = new Response("<html>502 Bad Gateway</html>", { status: 502 });
try {
  await response.json();
} catch {
  await response.text(); // TypeError: Body is unusable: Body has already been read
}
```

## Fix

Read the body once as text, then `JSON.parse` it, preserving the original `invalid_response` fallback:

```ts
const text = await response.text();
let data;
try {
  data = JSON.parse(text);
} catch {
  data = { error: "invalid_response", error_description: text };
}
```

Applied to both `src/lib/oauth/providers/kimi-coding.ts` and `src/lib/oauth/providers/github.ts` (identical copied snippet). The two now-unused `no-unused-vars` suppressions for the removed `catch (e)` bindings are pruned.

## Tests

- Added `tests/unit/oauth-polltoken-nonjson-body.test.ts`: drives both provider modules with a stubbed `fetch` returning (a) an HTML 502 error page → asserts `{ ok: false, data: { error: "invalid_response" } }` with the page text preserved, and (b) a JSON OAuth error body → asserts it is still parsed. 4/4 pass.
- Regression: `oauth-providers-config`, `oauth-providers-error-handling`, `oauth-device-flow-11164`, `kimi-token-refresh` — 45/45 pass.
- `npx eslint` on the touched files: clean.

### DIFF
--- a/changelog.d/fixes/13046-oauth-polltoken-nonjson-body.md
+++ b/changelog.d/fixes/13046-oauth-polltoken-nonjson-body.md
@@ -1,0 +1,1 @@
+- fix(oauth): kimi-coding/github device-flow `pollToken` no longer rejects with `TypeError: Body is unusable` when the token endpoint returns a non-JSON error page (CDN/anti-bot/proxy interstitial) — the body is now read once and parsed, preserving the graceful `invalid_response` fallback instead of a generic 500 (#13046 — thanks @ysntony)
--- a/config/quality/eslint-suppressions.json
+++ b/config/quality/eslint-suppressions.json
@@ -1687,16 +1687,6 @@
       "count": 1
     }
   },
-  "src/lib/oauth/providers/github.ts": {
-    "@typescript-eslint/no-unused-vars": {
-      "count": 1
-    }
-  },
-  "src/lib/oauth/providers/kimi-coding.ts": {
-    "@typescript-eslint/no-unused-vars": {
-      "count": 1
-    }
-  },
   "src/lib/oauth/providers/kiro.ts": {
     "@typescript-eslint/no-unused-vars": {
       "count": 3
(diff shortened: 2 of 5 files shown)


diegosouzapw:
Textbook double-body-read bug, cleanly fixed — read-once-then-parse is exactly right. Ran
the new suite green (4/4) against the PR head, lint clean on both touched files. No
blockers.

diegosouzapw:
Thanks @ysntony — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #13048, opened 2026-09-08
fix(sse): synthesize Responses SSE stream for web_search fallback on streaming requests (#13033)

### LINKED ISSUE #13033: fix(api): Responses stream:true with a native web_search tool is forced non-streaming (3.8.51, regression from #12031) — Codex gets "stream closed before response.completed"
### OmniRoute Version

3.8.51 (also present in the `next` Docker image built 2026-09-07)

### Summary

Since 3.8.51 (PR #12031, which fixed #12030), any OpenAI Responses API request with `stream: true` that declares a native `web_search` tool (`tools: [{ "type": "web_search" }]`) is silently forced to `stream: false` for the upstream call and returned to the client as a **plain JSON response with `Content-Type: application/json`** — no SSE `event:`/`data:` frames, no `response.completed`. Streaming Responses clients (Codex CLI / Codex desktop) then fail with

> stream closed before response.completed

while the HTTP status stays 200 and the upstream request is still billed.

**Important:** this triggers just by the tool being *declared* in `tools` — the model does not need to call it. Codex sends `web_search` in its default Responses toolset, so *all* Codex Responses traffic through 3.8.51 is affected for any provider where the web-search fallback is not bypassed (not Gemini, not Claude→Claude, no native-Codex passthrough, no `interceptSearch` override).

### Root cause

`open-sse/handlers/chatCore.ts` (3.8.51): after `prepareWebSearchFallbackBody()` reports the fallback enabled, if the source format is Responses and the request requested `stream: true`, the request body's `stream` is set to `false` (logged as `web_search fallback forced non-streaming response for <provider>`), and the non-streamed result is returned verbatim instead of being framed back into the SSE stream the client asked for.

### Reproduction

```
POST /v1/responses
Authorization: Bearer <key>

{
  "model": "<any chat-format openai-compatible provider>",
  "stream": true,
  "input": "hi",
  "tools": [{ "type": "web_search" }]
}
```

Observed on 3.8.51:
- `HTTP 200`
- `Content-Type: application/json` (not `text/event-stream`)
- Single JSON object body; no `event:`/`data:` frames; no `response.completed`

Control — same request **without** `tools`:
- `Content-Type: text/event-stream`
- `event: response.created` … `response.completed` ✅

3.8.50 does not have this behavior; 3.8.51 introduced it.

### Expected behavior

The `stream: true` client contract must be preserved. When the web-search fallback executes the search (the #12030/#12031 intent), the executed result — `web_search_call` item with `action.sources`, and/or `function_call` + `function_call_output` — must be returned to the client **as an SSE event stream ending with `response.completed`** (i.e. buffer the executed non-streaming result and emit it as SSE frames), rather than returning raw JSON for a `stream: true` request.

### Context

- Related: #12030 (client expects a `web_search_call` item), #9725 (streaming web-search dangling tool call — the original gap #12031 set out to fix)
- Workarounds today are not clean: `interceptSearch: false` restores streaming but forwards the native `web_search` tool upstream (strict function-only upstreams such as GPUStack reject it with a 400); exposing search via the gateway's MCP server works around it but requires client MCP config.

### Log evidence (from a production 3.8.51 deployment)

Applied to every affected request:
`TOOLS: web_search fallback forced non-streaming response for openai-compatible-chat-<id>`
and in this probe (stream:true + web_search tool), the `Content-Type` of the responses endpoint response was `application/json` while the no-tool control returned `text/event-stream` with `response.completed`.

HouMinXi:
Fixes #13033.

### Problem
When an OpenAI Responses API client (such as Codex CLI, Cursor, or direct `/v1/responses` callers) requests `stream: true` on a request containing a native `web_search` tool, OmniRoute converts the search into server-side execution via `omniroute_web_search`.

To execute this server-side multi-turn loop, PR #12031 forced `body.stream = false` prior to execution. However, once execution completed, `chatCore` treated the request as non-streaming and called `buildNonStreamingJsonResponse`, returning `Content-Type: application/json`. Standard Responses streaming clients expecting `text/event-stream` immediately abort with:
```
stream closed before response.completed
```

### Solution
1. **Preserve streaming intent flag**: In `chatCore.ts`, when an OpenAI Responses request with `stream: true` has its search tool converted to fallback execution, track `forcedNonStreamingResponsesSse = true` while keeping `body.stream = false` for the server-side tool execution pipeline.
2. **Synthesize Responses SSE stream**: When the response is ready, synthesize a spec-compliant Responses API SSE event sequence (`response.created` -> `response.in_progress` -> `response.output_item.added` -> `response.output_text.delta`/`done` -> `response.output_item.done` -> `response.completed` -> `data: [DONE]`) via `buildNonStreamingResponsesSseResponse`.
3. **Spec compliance and edge-runtime safety**:
   - Maintains exact ID and `created_at` consistency across `response.created` and `response.completed`.
   - Populates complete `response.in_progress` fields (`status`, `background`, `error`, `output`) matching `response.created`.
   - Omits internal properties prefixed with `_` from the completed response.
   - Replaces `Date.now()` with `crypto.randomUUID()` for collision-free fallback IDs.
   - Avoids Node-specific global buffer dependencies for content headers.

### Verification
- Added comprehensive unit tests in `tests/unit/responses-json-to-sse-13033.test.ts` covering message streams, tool call events (`web_search_call`, `function_call`), JSON payload wrapping, ID consistency, fallback envelopes, and internal property stripping (9/9 passed).
- Updated integration test in `tests/integration/skills-pipeline.test.ts` verifying that `stream: true` Responses requests with web search return `text/event-stream` with valid SSE lifecycle events.
- Validated via bug-injection: reverting the SSE synthesis branch caused `skills-pipeline.test.ts` to fail with `Input did not match /text\/event-stream/`; restoring it returned 100% green.
- `npm run typecheck:core` passed with zero errors.
- Clean ASCII diff checked with zero non-ASCII characters.

### DIFF
--- a/changelog.d/fixes/13033-responses-websearch-stream.md
+++ b/changelog.d/fixes/13033-responses-websearch-stream.md
@@ -1,0 +1,1 @@
+- **fix(sse):** synthesize Responses API SSE stream when streaming requests fall back to server-side non-streaming web search execution, preventing premature stream closes in Codex CLI and other Responses clients.
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -35,6 +35,7 @@ import { buildPostCallGuardrailContext } from "./chatCore/postCallGuardrailConte
 import { storeSemanticCacheResponse } from "./chatCore/semanticCacheStore.ts";
 import { buildNonStreamingResponseHeaders } from "./chatCore/nonStreamingResponseHeaders.ts";
 import { buildNonStreamingJsonResponse } from "./chatCore/nonStreamingJsonResponse.ts";
+import { buildNonStreamingResponsesSseResponse } from "../utils/responsesJsonToSse.ts";
 import { enforceOutputTokenBudget } from "./chatCore/outputTokenBudget.ts";
 import { maybeConvertJsonBodyToSse } from "./chatCore/jsonBodyToSse.ts";
 import { assembleStreamingResponseHeaders } from "./chatCore/streamingResponseHeaders.ts";
@@ -950,19 +951,22 @@ export async function handleChatCore({
       nativeCodexPassthrough: nativeResponsesPassthrough,
       interceptSearchOverride,
     });
+  let forcedNonStreamingResponsesSse = false;
   if (webSearchFallbackPlan.enabled) {
     body = bodyWithWebSearchFallback as typeof body;
     // Server-side web-search execution cannot be injected into an arbitrary
-    // client SSE stream (streaming interception is not implemented — #9725), so
+    // client SSE stream (streaming interception is not implemented -- #9725), so
     // a stream:true OpenAI Responses request whose web_search tool was converted
     // to the fallback is executed non-streaming: the assembled response then
-    // carries the executed results (function_call_output + web_search_call) and
-    // JSON-tolerating Responses clients (pi-web-access) consume it directly.
+    // carries the executed results (function_call_output + web_search_call).
+    // When the client requested a stream, synthesize Responses SSE events from
+    // the completed response (#13033) instead of returning plain JSON.
     if (
       sourceFormat === FORMATS.OPENAI_RESPONSES &&
       (body as Record<string, unknown>).stream === true
     ) {
       (body as Record<string, unknown>).stream = false;
+      forcedNonStreamingResponsesSse = true;
       log?.info?.("TOOLS", `web_search fallback forced non-streaming response for ${provider}`);
     }
     log?.info?.(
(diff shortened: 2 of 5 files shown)


HouMinXi:
Superseded by #13050 (`d552a901c`).

Same issue (#13033). #13050 reuses `synthesizeOpenAiSseFromJson` and `createResponsesApiTransformStream` (sibling ~65 lines) and is already on `release/v3.8.51` (`ba597b631`). This branch added a 330-line synthesizer and lagged the #12963/#12990/#13045 return-site changes.

HouMinXi:
Closing in favor of #13050.

PR #13050, opened 2026-09-08
fix(responses): wrap forced-non-streaming web_search JSON as SSE

### LINKED ISSUE #13033: fix(api): Responses stream:true with a native web_search tool is forced non-streaming (3.8.51, regression from #12031) — Codex gets "stream closed before response.completed"

HouMinXi:
## Bug

Fixes #13033

`/v1/responses` with `stream: true` and a native `web_search` tool is still forced non-streaming in v3.8.51 so the server-side fallback can run (#12031). The HTTP body is then `application/json`. Codex waits for `event: response.completed` on SSE, gets JSON, and errors with `stream closed before response.completed`. Upstream is billed. HTTP stays 200.

## Cause

Production `/v1/responses` goes through `handleChat`, not `handleResponsesCore`. `chatCore` sets `body.stream = false` for the web_search fallback and later returns `buildNonStreamingJsonResponse`. Nothing turns that JSON back into Responses SSE.

Wiring the wrap only in `handleResponsesCore` would not fix Codex. That helper is tests-only.

## Fix

- Stamp that the client asked for SSE before flipping `stream` off.
- On the non-streaming success return, if that flag is set, synthesize Chat Completions SSE from the JSON (`synthesizeOpenAiSseFromJson`) and pipe it through `createResponsesApiTransformStream`.
- Wrap lives in `open-sse/handlers/chatCore/responsesJsonToSse.ts`. `chatCore.ts` only stamps the flag and swaps the return (+6 lines after #13045's 6021).

This does not implement streaming interception (#9725). Empty synthesis still falls back to JSON, same as `jsonBodyToSse.ts`.

## How to verify

1. `npx tsx --test tests/unit/responses-json-to-sse-13033.test.ts tests/unit/8395-plugin-hooks-fire.test.ts`
2. Drop `clientRequestedResponsesStream = true` in `chatCore.ts`. The stamp test goes red. Restore, it goes green.

## Test plan

- [x] Regression test for wrap + stamp
- [x] Existing 8395 hook-order test updated to the new return name
- [ ] CI on this PR

## Risk

Low. Flag false keeps `buildNonStreamingJsonResponse`. Only Responses + web_search fallback + original `stream:true` wraps.


HouMinXi:
Looked at the product commit on `chatCore.ts`, `responsesJsonToSse.ts`, and the two unit tests.

The wrap is the fix. Empty `synthesizeOpenAiSseFromJson` still returns JSON, matching `jsonBodyToSse.ts`. The flag is assigned in three places only (declare, stamp, return).

The lock is `tests/unit/responses-json-to-sse-13033.test.ts`: drop `clientRequestedResponsesStream = true` and the stamp test goes red.

Rebased onto `ba597b631` before opening. The only conflict was the non-streaming return after #12963/#12990/#13045. Flag stamp and wrap survived. File-size ratchet 6021 to 6027 is a follow-up commit on this branch.

HouMinXi:

PR #13060, opened 2026-09-08
fix(resilience): classify Cline 401 as refreshable OAuth (#12594)

### LINKED ISSUE #12594: fix(providers): Cline 401 classified terminal and reported as a 429 cooldown
### OmniRoute Version

3.8.51

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 22.04 (WSL)

### Node.js Version

_No response_

### Provider(s) Involved

Cline

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

After authenticating the Cline provider, model requests work initially, but after some time, every model request starts returning the following error:

```text
[401]: Unauthorized: Please make sure you're using the latest version of Cline and re-authenticate your Cline account.
```

When I click the **Retest** button for the Cline Provider in the dashboard, I get the following error:

```text
These connections returned a 429 (rate-limit) on their last request. OmniRoute will skip them until the timer expires — no manual disable required.
```

However, when I connect to Cline directly from VS Code using the **same Cline account**, free model requests work normally without any issues.

### Expected behavior

The authenticated Cline provider should continue to work normally for model requests, as it does when using the same account directly through the VS Code Cline extension.

### Actual behavior

* Cline provider authentication succeeds.
* Model requests work initially.
* After some time, all model requests start returning `401 Unauthorized`.
* Retesting the provider reports a `429 rate-limit`.
* The same Cline account works normally when used directly through the VS Code Cline extension.

### Environment

* OmniRoute
* Cline Provider
* VS Code Cline extension
* Same Cline account used for both connections

### Steps to Reproduce

1. Authenticate the **Cline Provider**.
2. Wait for some time.

   * The provider works normally for a while immediately after authentication.
3. After some time has passed, test any model from the coding agent or the **Cline Provider**, or click the **Retest** button.
4. The request returns a `401 Unauthorized` error, or the Retest operation reports a `429 rate-limit` error.

### Expected Behavior

* A `429 rate-limit` error should not occur before the actual rate limit or usage quota has been reached.
* The authentication state should remain valid, and an `Unauthorized` error should not be returned unless the account's authentication or authorization status has actually changed.

### Actual Behavior

* An `Unauthorized` error is returned even though the actual authentication status of the account remains valid.
* A `429 rate-limit` error is returned even though the account still has remaining usage quota.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

Cline authentication in VS Code is currently performed using the device identification number.

### Validation Plan

_No response_

HouMinXi:
## Description

Cline returns 401 with "re-authenticate your Cline account" when the access token is stale. That phrasing was not in `OAUTH_INVALID_TOKEN_SIGNALS`, so `classifyProviderError` treated it as a generic `UNAUTHORIZED` and `resolveTerminalConnectionStatus` marked the connection `expired`. A token refresh would have recovered it (`refreshClineToken` already exists). The cooling panel then printed "429 (rate-limit)" for every connection with a future cooldown timestamp, including this auth failure.

Reporter: usable ~15 min after re-auth, then the same 401. Diego already traced both defects on the issue.

## Changes

- Add `"re-authenticate your cline account"` to `OAUTH_INVALID_TOKEN_SIGNALS`. A Cline 401 now classifies as `oauth_invalid_token` (non-terminal) instead of `expired`.
- Cooling panel copy no longer assumes 429. Rows show the recorded `lastError` when present.
- 42 locale strings for `coolingConnectionsDescription` updated to match.

## Test plan

- `node --import tsx/esm --test tests/unit/cline-401-oauth-12594.test.ts` (4/4)
- Injection: drop the Cline signal → 4/4 red; restore → 4/4 green
- `tests/unit/account-fallback-service.test.ts` + `tests/unit/auth-terminal-status.test.ts` (107/107)
- vitest CoolingConnectionsPanel 10/10
- `accountFallback.ts` stays at frozen 2467 lines

Closes #12594

### DIFF
--- a/changelog.d/fixes/12594-cline-401-oauth.md
+++ b/changelog.d/fixes/12594-cline-401-oauth.md
@@ -1,1 +1,1 @@
-Cline 401 bodies that say "re-authenticate your Cline account" classify as a refreshable OAuth token, not a terminal expired key. The cooling panel no longer labels every cooldown as a 429; it shows the recorded last error instead. (#12594)
+- Cline 401 bodies that say "re-authenticate your Cline account" classify as a refreshable OAuth token, not a terminal expired key. The cooling panel no longer labels every cooldown as a 429; it shows the recorded last error instead. (#12594)
--- a/changelog.d/fixes/12594-cline-401-oauth.md
+++ b/changelog.d/fixes/12594-cline-401-oauth.md
@@ -1,0 +1,1 @@
+Cline 401 bodies that say "re-authenticate your Cline account" classify as a refreshable OAuth token, not a terminal expired key. The cooling panel no longer labels every cooldown as a 429; it shows the recorded last error instead. (#12594)
--- a/open-sse/services/accountFallback.ts
+++ b/open-sse/services/accountFallback.ts
@@ -261,6 +261,7 @@ export const OAUTH_INVALID_TOKEN_SIGNALS = [
   "login cookie",
   "valid authentication credential",
   "invalid credentials",
+  "re-authenticate your cline account",
 ];
 
 // A model that upstream has permanently retired — Gemini's deprecated-model 404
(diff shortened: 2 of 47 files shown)


HouMinXi:

PR #12904, opened 2026-09-07
fix(sse): inject global system prompt once, post-translation, across all target shapes

initguru:
## Summary

define global system prompt (`OMNIROUTE_SYSTEM_INSTRUCTION_APPEND`) injection at a **single post-translation point**. This resolves HCP-Vision vLLM's "System message must be at the beginning" 400.

- Previous problem: `injectSystemPrompt()` modified the **pre-translation** body → the prompt was lost/repositioned during translation and injected 2–3 times depending on the format. The Responses path was not reached at all (0004 partially resolved this by adding a post-translation injection point, but pre-translation injection remained, causing double injection).
- fix: remove pre-translation injection → use a single post-translation `injectSystemPromptPostTranslation(body, {targetFormat})` point (claude/gemini/openai-responses/openai/codex branches) + a gated pre-translation pass `injectSystemPromptPreTranslation` only for carrier-less targets (kiro, antigravity envelopes).
- idempotence: use a non-enumerable `_systemPromptInjected` flag (`Object.defineProperty`, not exposed in JSON) so injection occurs only once regardless of the path taken.
- target-shape coverage: create a combined system when the Claude system field is absent (string+array), guard the Gemini `!result.request` envelope, wrap Responses `instructions`, use first/last system for Codex, gate carrier-less kiro/antigravity, and no-op for Cursor.

## Related Issues

- Closes #12904

## Validation

- [x] Change type: provider / routing
- [x] Focused tests: `node --import tsx/esm --test tests/unit/system-prompt.test.ts tests/unit/system-instruction-append-claude-to-openai.test.ts tests/unit/global-prompt-single-injection.test.ts` (26/26 + existing GREEN)
- [x] `npm run typecheck:core` — 0 errors
- [x] `npm run lint` — 0 errors
- [ ] Reconciled with `release/v3.8.51` base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR

## Tests Added Or Updated

- `tests/unit/global-prompt-single-injection.test.ts` (new, 26 tests) — single-injection contract: idempotence, non-enumerable flag absent from JSON, Claude string/array, Gemini present/absent, Responses present/absent, Codex first/last, skip flag, 9 gates, dual-write negative case, 4 source-shape branches
- `tests/unit/system-prompt.test.ts` (updated)
- `tests/unit/system-instruction-append-claude-to-openai.test.ts`

## Coverage Notes

[...]

PR #11742, opened 2026-08-27
security(runtime): harden TLS provenance, lifecycle, and public error boundaries

diegosouzapw:
## Summary

This draft hardens the frozen `release/v3.8.50` line around the native
`tls-client-node` supply chain, runtime lifecycle, and public error/log boundaries. It is
independent of other PRs, but it **does change provider executors and validation behavior**;
the previous narrower scope statement is intentionally superseded.

| Area | What changed | Evidence | Disposition |
| --- | --- | --- | --- |
| Native dependency provenance | Pins `tls-client-node@0.2.0`, records the exact wrapper/upstream licenses and NOTICE, and pins all six supported v1.15.1 native assets to official SHA-256 digests | License/provenance and tamper regressions; manifest/source hash checks | PASS for recorded provenance and policy |
| Build and packaging policy | Verifies the native binary before runtime/build copies and carries `THIRD_PARTY_NOTICES.md` through npm, standalone, Docker, and Electron assembly paths | Build/pack policy tests, MJS syntax checks, and exact-head production build/artifact evidence below | Source/policy and exact-head production build PASS; real pack, Docker, and Electron artifacts remain HOLD |
| Runtime lifecycle and install lock | Adds bounded lifecycle coordination, safe retries, handle/inode/device/nonce lock ownership, cross-process coverage, and fail-closed cleanup | TLS serial matrix `47/47`, including 100 callers and cross-process contention | PASS on local filesystems |
| Error boundary | Extracts a canonical sanitizer for credentials, stacks, recognized filesystem paths, hostile Proxies, Unicode/serialized separators, and public identifiers | Canonical matrix `121/121`; runtime/spec review CLEAR | PASS |
| Provider consumers | Hardens ChatGPT, Claude, Grok, Perplexity, Notion, LM Arena, moderation, OCR, and provider validation without exposing raw upstream errors | Provider-specific matrices listed below | PASS within tested boundaries |
| LM Arena public failures | Projects HTTP, network, TLS, upstream-event, and stream-transport failures through a closed local union; rejects raw upstream bodies/stacks/causes and accepts only public error statuses `400..599` | Iteration-4 RED `53/55` -> GREEN `55/55`; final Arena matrix `124/124`; direct CodeQL #968 fixed | PASS locally and in final CodeQL analysis |
| Perplexity streaming | Preserves early quota/upstream HTTP status through a bounded preflight, replays normal content without duplication, and releases readers/iterators | `54/54` | PASS with late-error limit below |
[...]

PR #13065, opened 2026-09-08
refactor(chatcore): decompose handleChatCore into leaf modules

HouMinXi:
## Summary

Re-extracted `handleChatCore` decomposition directly on current `release/v3.8.51` tip (`7a921299c5`) instead of merging conflicting intermediate branches.

`chatCore.ts` drops from 6,200+ lines down to 992 lines by isolating discrete responsibilities into leaf modules under `open-sse/handlers/chatCore/`.

## What changed

- `providerExecutionPipeline.ts`: Provider execution pipeline and credential retry loop.
- `recoveryPolicy.ts`: Recovery decision table and error classification.
- `streamingLeg.ts`: SSE stream dispatch, lifecycle hooks, and abort listeners.
- `nonStreamingLeg.ts`: Non-streaming response collection and payload forwarding.
- `materializeStreamSuccess.ts`: Stream completion, usage calculation, and analytics persistence.
- `cacheAndCompress.ts`: Semantic cache lookup, cache hit construction, and context compression.
- `translateAndDedup.ts`: Request translation, tool normalization, stream controller setup, and request deduplication.
- `chatCore.ts`: Retains top-level routing orchestration and target retry dispatch (down to 992 lines).

## Verification

- `npm run typecheck:core`: clean (0 errors)
- `npm run check:any-budget:t11`: pass
- Unit test suite: 102/102 tests passing across translation, caching, compression, streaming legs, and pipeline dispatch
- Tested with defect injection across each module interface

Related to the earlier extract discussion on this PR. #12339 stays open. #12696 stays open.

⚠️ base-red inherited: #13866


HouMinXi:
CI on this head (`8cdcf3a94`, now `7dc90cfa8`):

Own-red (fixed, this SHA):
- ESLint unused barrel imports after the recovery lift:
  `createRecoverableStream`, `makeContinuationBody`,
  `isStreamRecoveryExplicitlyConfigured`. Local `eslint --no-cache` on the
  five unique logic files is clean after `7dc90cfa8`.

Inherited on `release/v3.8.51` (same fail set on #13079 / #13078 / #13074;
#13076 / #13073 are green):
- Docs Gates (fast-path)
- Fast Quality Gates
- Unit Tests fast-path 1/4–4/4
- API Route Typecheck / semgrep (this PR only; annotations are the generic
  “exit code 1” wrapper, no unique-file hit)

Unique files do not include glm.ts, changelog leading “- ”, or the
agent-skills cli-tunnel docs. Merge integrity and Vitest fast-path are green.

`chatCore.ts` 6021 → 5202. Frozen cap is still 5034/6021 on tip; local
`check:file-size` is OK because the file shrinks.

This contribution is intended for maintainer review and merge.

HouMinXi:
Pushed `032344c76`. API Route Typecheck was unique to this PR:

- `open-sse/handlers/chatCore.ts` TS2304: `VideoBridgeLogParam` used `VideoBridgeLogRedactionEntry` without importing it. Restored the type import (same one `requestPrelude.ts` already has).
- `open-sse/handlers/chatCore/executeProviderRequest.ts` TS2322: `withRateLimit` opts want `getTimeoutMs?: () => unknown`, the lift had `getTimeoutMs?: unknown`.

Local `check:api-typecheck` is now within the frozen baseline. Remaining Docs / Fast Quality / ESLint / unit-shard red is the shared `release/v3.8.51` set.

diegosouzapw:
Holding this one out of today's batch deliberately, and it is not a judgement on the work.

The other 20 PRs of yours merged today. Two of them rewrite the same function this decomposes: #13069 routes the non-streaming leg through `applyProviderFailureClassification` (+443/-340 in `chatCore.ts`) and #13050 adds the Responses SSE wrap around the web_search fallback. `package.json` also moved under you — #13187 dropped the local test scripts from concurrency 20 to 4 today, which is the other half of your conflict.

The reason to give this its own round rather than carry it in a 21-PR batch: a behaviour-preserving decomposition of the hottest path in the codebase is exactly the change where the batch's economics stop working. The consolidated model runs focused tests per PR and leans on attribution when something goes red, and a refactor that touches every block of `handleChatCore` breaks attribution for everything merged alongside it. Today's batch proved the point in miniature — two ESLint regressions slipped past the focused-test validation and took three follow-up PRs to clear.

What it needs from here:

1. **Rebase onto the current tip.** `chatCore.ts` now carries #13069's extracted classification helper and #13050's SSE intent stamp. Both should land inside the sender leaf and the barrel call sites rather than as a second copy, the same way you handled the queue-budget and per-model-402 cases.
2. **Keep the `chatcore` entry** you added to `UNIT_SUBDIRS` across the eight npm scripts and `scripts/release/merge-train.sh`. Without it the seven new suites under `tests/unit/chatcore/` are never discovered, which is the failure mode where a test exists but does not run.
3. **When it is green, say so here** and I will validate it against the full unit suite rather than the focused subset. "No behaviour change" is a claim only the full suite can check.

On the 252 lines over the `<2500` fence: not a blocker from my side. Say what is left and why it cannot move, and that is the review conversation.

Thanks for the other twenty. The live evidence on them — the X500 logs, the 36-minute outage write-up, the real `/v1/models` probes — is what made a batch that size reviewable as a unit.

diegosouzapw:
Careful, well-executed decomposition — chatCore.ts goes from 6145 to 2752 lines (55% cut), and
you covered every test-discovery mechanism in the repo for the new `tests/unit/chatcore/`
subdir (package.json scripts, check-test-discovery.mjs, build-test-impact-map.mjs,
select-impacted-tests.mjs, merge-train.sh, stryker.conf.json) — exactly the "phantom test" trap
this kind of change usually falls into. Ran all 9 new/touched test files at this PR's head:
46/46 pass, including several structural canaries confirming the barrel doesn't still inline the
extracted logic. No file-size-baseline rebaseline needed either. The real risk here isn't code
quality, it's timing: this group also has #12498, #13448 and #13556 all touching chatCore.ts —
whichever merges first will force a rebase on the others. Worth asking the owner to pick an
explicit merge order for this batch before landing any of the chatCore.ts PRs, and running the
broader chatCore-adjacent test suite (not just the new leaf files) once an order is picked, given
how central this file is.

diegosouzapw:
Thanks for taking this on — `chatCore.ts` is the file that most needs it, and the split you drew (send / recovery / stream completion / request prelude) is the right decomposition.

Holding it for now, for reasons that are about timing rather than the work itself:

- The branch no longer merges: `open-sse/handlers/chatCore.ts` and `package.json` both conflict with the current tip. `chatCore.ts` is 6142 lines on `release/v3.8.51` today against the 2872 your branch produces, and a large part of that delta is the ~40 PRs that landed in it over the last two days.
- A pure refactor is only safe when the equivalence is mechanically checkable. With a conflict of this size the merge resolution itself becomes new, untested code in the hottest path in the repo (every chat completion goes through it), and there is no behavioral test that would catch a subtle reordering of the early returns, the abort handling or the cache/compress step.

What would make it mergeable: rebase onto the current `release/v3.8.51` tip and re-derive the split from that version (`git diff` of each extracted module against the corresponding block of the tip should be empty apart from the import/export wiring), so the review can verify equivalence hunk by hunk. If you do that, ping me here and it goes to the front of the queue.

The tip already carries `open-sse/handlers/chatCore/providerExecutionPipeline.ts` and a number of sibling modules, so part of the direction is already landing incrementally — your branch and the tip are converging, not diverging.

HouMinXi:
@diegosouzapw I do want this merged; the earlier contributor-only / do-not-merge wording was my mistake, not a request to withhold the work. I have corrected the description and my earlier comment. Your current-base equivalence requirement is reasonable: an old conflict resolution is not evidence of a behavior-preserving split. I will not call the old tests proof for the new tip. The next revision needs block-by-block equivalence against the current source, preserved test discovery, and the broader adjacent suite before it is ready.

diegosouzapw:
**Precisa ser refeita sobre o tip atual. Não forcei a resolução, e explico por quê.**

Tentei o merge e abortei sem commitar nada. Os números: o tip avançou `open-sse/handlers/chatCore.ts` em **9 commits / 50 hunks / +259−111** desde a base desta branch (`0cc0169`), espalhados por toda a `handleChatCore`. Esta PR move ~3.400 dessas linhas para 6 módulos novos. O conflito é de **8 regiões, uma delas com 4.971 linhas** (lado da PR ≈ 38 linhas, lado do tip 2.499) — a forma clássica de "um lado deletou tudo, o outro editou dentro".

Resolver significaria decidir, para cada um dos 50 hunks de 9 PRs já mergeadas, em qual dos 6 módulos ele passa a morar, e reconstruir a fiação. Isso é **reexecutar a extração**, não resolver um conflito — e é trabalho seu, não meu, porque é o desenho da separação.

**Dois agravantes fecharam a decisão:**

1. Os canários estruturais da PR testam o **barrel**, não cada comportamento portado. Um hunk perdido no caminho não seria pego por teste nenhum.
2. Entre os hunks em risco está o endurecimento da fronteira de log da **#11742** (`safeErrMsg`/`safeUpstreamErrorBody` — superfície da Hard Rule #12), a divisão pre/post-translation do system prompt (#12904), o `reasoningRuleDirective` (#13556), o `getManagedLeaseFenceErrorCode` e o adaptive effort (#13448). Perder qualquer um em silêncio reabre um buraco de segurança que os testes não cobrem.

**Sugestão:** rebase/refazimento da extração sobre o tip atual, e mergear antes as PRs cirúrgicas do grupo — que é a ordem que o próprio plan-file sugere. O trabalho de separação continua valendo; só precisa partir de uma base que não seja de 9 commits atrás.

HouMinXi:
Pinned the Claude OAuth raw passthrough wiring from #14053 onto this split so the three execute sites do not live only in the pre-split barrel.

`isClaudePassthrough` now comes out of `translateAndDedup`, rides `sendDeps`, and is passed at the two `executor.execute` calls in `executeProviderRequest.ts` plus the leftover credential-retry execute in `chatCore.ts`. Proxy native fallback, dashboard toggle, and the cloak/fingerprint/CCH gate in `base.ts` match #14053. One extra filter: client `x-app` / `x-stainless-*` / `anthropic-dangerous-direct-browser-access` are dropped during the final header merge, not only stripped from executor leftovers.

Tests: `claude-oauth-raw-passthrough` + schema = 51/51. Head `9f4108c447`. Related to #13893; #13893 stays open.

This is not the tip re-extract. The PR is still CONFLICTING against current `release/v3.8.51` for the reason already given — the 50 hunks / 9 merged PRs still have to be placed into the six leaves. That remains a separate pass.

HouMinXi:
Will not hand-resolve the eight conflict regions. You are right that that would be re-running the extraction, not merging.

Plan from here: re-derive leaves on current `release/v3.8.51` (`1603c86e06`) so each new module diffs empty against the tip block except for import/export wiring. First leaf already exists on a side branch (`executeProviderRequest` out of `handleChatCore`). Further leaves will land the same way. This PR stays CONFLICTING until that re-extract replaces it; please keep holding.

#12498 / #13448 / #13556 / #13069 / #13050 are already on the tip this re-extract starts from.

HouMinXi:
Re-extracted on current `release/v3.8.51` instead of merging the stale split.

What is in this branch now:
- `executeProviderRequest` lifted out of `handleChatCore`
- recovery policy as effects plus dispatch
- family hop writes `wire.currentModel` on the pipeline object (the chatCore setter only mutates the closure)
- dead `pipelineRecovered` copies deleted; context-overflow hop stayed

Not in this branch: stream materialize, request prelude, and the Claude OAuth raw passthrough wiring that had leaked into the old head.

Tests: 33/33 on the send + recovery + pipeline files.

HouMinXi:
Title and body now match the branch.

This PR is the sender leaf plus the recovery policy table. Stream completion, request prelude, cache, and translation stay in `handleChatCore`. I stopped there instead of forcing leftover `chatCore.ts` under 2500 lines on this ticket.

Head `7c9e8aeb67` on current `release/v3.8.51`. Tests on send + recovery + pipeline: 33/33.

HouMinXi:
Rebased onto current `release/v3.8.51` (`36493a6270`). Unique commits unchanged.

HouMinXi:
Re-extracted on current `release/v3.8.51` tip (`7a921299c5`) per review feedback, instead of merging conflicting intermediate branches.

What is in this branch now:
- `open-sse/handlers/chatCore.ts` drops from 6,200+ down to 992 lines.
- Discrete steps moved into clean leaves under `open-sse/handlers/chatCore/`:
  - `providerExecutionPipeline.ts` (provider execution and retry loops)
  - `recoveryPolicy.ts` (recovery decision table)
  - `streamingLeg.ts` and `nonStreamingLeg.ts` (execution legs)
  - `materializeStreamSuccess.ts` (stream completion and usage accounting)
  - `cacheAndCompress.ts` (semantic cache lookup and context compression)
  - `translateAndDedup.ts` (request translation and deduplication)
- All 102 unit tests in the chatcore suite pass.
- Verified with defect injection on interface boundaries.

RaviTharuma:
@diegosouzapw this one is super nice

HouMinXi:
Re-extracted onto the current release/v3.8.51 tip as #14725: each leaf diffs empty against its tip region (imports/exports and deps/carry wiring aside), one clean commit, 209 chatcore-referencing test files show zero new failures against the tip baseline. Two wiring seams the suite caught after the lifts (translatedBody snapshot vs live binding; test re-anchors) are fixed and covered. This PR can stay open for history; the tip-based version is the one to review.

PR #13071, opened 2026-09-08
fix(providers): lazily load chatgpt-web-codex admin helpers in PUT /api/providers/[id]

hartmark:
## What Problem This Solves

`PUT /api/providers/{id}` -- renaming a connection, rotating its API key,
etc. -- **for any provider** returned a 500 after 17-50 seconds:

```
Error: Missing tiktoken_bg.wasm
    at module evaluation (open-sse/vendor/codex-chatgpt-web/lib/token-estimate.ts:2:1)
```

Live repro: PUT-ing a rename on a plain `llama-cpp` (openai-compatible)
connection -- nothing to do with `chatgpt-web-codex` at all.

## Root Cause

#12355 already fixed this exact bug class in the sibling **collection**
route (`src/app/api/providers/route.ts`): a static top-level import of
`chatgptWebCodexAdmin.ts` evaluates that module's transitive chain (the
vendor ChatGPT-Web browser adapter -> `token-estimate.ts` -> tiktoken's WASM
tokenizer) on *every* request to the route, regardless of which provider is
actually being managed. That chain fails to bundle under Turbopack dev mode
even with `tiktoken` listed in `serverExternalPackages` (the standalone Node
`require` works fine; only Turbopack's bundling of this specific import
graph doesn't).

#12355 only touched the collection route and missed the **identical
pattern** in the **by-id** route (`src/app/api/providers/[id]/route.ts`),
which has the same static import used only inside one
`provider === "chatgpt-web-codex"` branch of the `PUT` handler. #12355 also
shipped with no regression test, which is exactly how this sibling instance
went unnoticed for a week.

## Fix

Same pattern as #12355: move the import into a lazy `await import(...)`
inside the one branch that actually needs it.

## What This Deliberately Does NOT Change

`src/app/api/providers/[id]/chatgpt-web-codex-doctor/route.ts` keeps its
static import -- every request to that route already is for a
`chatgpt-web-codex` connection (it's in the route's own URL), so there's no
"unrelated provider pays the cost" problem there.

## Evidence

New regression test source-inspects both general provider routes to assert
neither statically imports `chatgptWebCodexAdmin.ts`, and confirms the
doctor route's static import is intentionally left alone. Matches the
existing `tests/unit/instrumentation-import-graph-12074.test.ts` pattern for
import-graph invariants that can't be exercised by actually bundling under
Turbopack from this harness. Confirmed it fails against the pre-fix
`[id]/route.ts` and passes after the fix.

```
$ node --test tests/unit/providers-chatgpt-web-codex-lazy-import-12355.test.ts
tests 3, pass 3, fail 0

$ node --test tests/unit/providers-route-patch-method.test.ts tests/unit/codex-connection-edit-6562.test.ts \
    tests/unit/chatgpt-web-management-retirement.test.ts tests/unit/provider-patch-ratelimit-protection-11278.test.ts
tests 11, pass 11, fail 0
```

- `eslint` on both touched files: clean

**Live-verified** against the exact failing request (deployed to my own dev
instance before opening this PR, given it was actively blocking a user):

```
Before: PUT /api/providers/{id} -> 500 in 17-50s, "Missing tiktoken_bg.wasm"
After:  PUT /api/providers/{id} -> 200 in 0.83s
```


diegosouzapw:
Clean, well-scoped fix — mirrors #12355's exact pattern for the sibling route, and the
source-inspection test is a smart way to pin an import-graph invariant that Turbopack bundling
can't be exercised for from the test harness. Please add a changelog fragment under
`changelog.d/fixes/` before merge; the CI unit-test shard failures look like known base-red
noise but worth a clean local re-run to be sure.

hartmark:
Checked, and it's actually branch staleness rather than pure base-red: this branch was 217 commits behind `release/v3.8.51`. Re-running the exact CI-failing tests locally at the old head reproduced 25/39 failures; after rebasing (clean, no conflicts) that dropped to 2. Both remaining ones (`dashboard-request-failed-redaction.test.ts`, `models-catalog-route.test.ts`) I confirmed also fail on a completely clean `release/v3.8.51` checkout with none of this PR's changes — genuine current base-red, unrelated to this fix (which only ever touched `providers/[id]/route.ts` and its own new test, 3/3 passing). Rebased, re-ran typecheck (clean) and the PR's own test suite, force-pushed. Added the changelog fragment — verified with `check:changelog-integrity`. Ready to merge from my side.

PR #12366, opened 2026-09-01
feat(models): add Claude Fable 5.1

davidebaraldo:
## What

Adds **Claude Fable 5.1** (`claude-fable-5-1`), released by Anthropic on 2026-09-01 ([announcement](https://www.anthropic.com/claude-fable-and-mythos-5-1), [model page](https://platform.claude.com/docs/en/models/fable-5-1/overview)).

Specs from the official docs: 1M context, 128K max output, adaptive-only thinking (cannot be disabled), non-default `temperature`/`top_p`/`top_k` rejected with a 400, $10/$50 per MTok — same list price as Fable 5, but **cache reads at $0.25/MTok** (a quarter of Fable 5's read price).

## Where it's added

Mirrors the existing `claude-fable-5` entries, only on surfaces that actually serve the model today:

- **Provider registries**: `anthropic` (API key), `claude` (OAuth), `claude/web`, `vertex` (Google Cloud lists `claude-fable-5-1` natively)
- **modelSpecs**: capabilities (`rejectsThinkingDisabled`, `adaptiveThinkingOnly`) + Bedrock aliases (`anthropic.claude-fable-5-1` per official docs)
- **Pricing**: new `CLAUDE_FABLE_5_1_PRICING` shared tier (frontier-labs) and the `cc` OAuth subscription table — both with the $0.25 cache-read price; `providerCostData` row
- **`CONTEXT_1M_SUPPORTED_MODELS`**, `staticModels` (claude), `modelFamilyFallback` (falls to `claude-fable-5`, then `claude-opus-5`/`claude-sonnet-5`)

## Deliberately not added

- Third-party gateway registries (cursor, kiro, github-copilot, etc.): their upstream catalogs haven't published the model — adding it there would be speculative.
- Fast-mode default list: fast-mode availability for 5.1 is not documented.

## Verification

- `typecheck:core` clean
- Catalog/pricing/context-1m/family-fallback/sampling-params test suites: 35/35 pass (`catalog-updates-v3x`, `claude-context-1m-supported-models`, `model-family-fallback-notation`, `catalog-pricing-lookup-index-8697`, `claude-adaptive-sampling-params`, `cc-discovery-aliases-catalog`)
- Pre-commit hooks (prettier, eslint with project suppressions, docs-sync, any-budget) pass

PR #13074, opened 2026-09-08
feat(sse): Claude OAuth lower-priority lane + weekly session-limit reset

davidebaraldo:
## What

Mirrors two Claude Code commands for OmniRoute-managed **Claude subscription (OAuth) connections**, both **opt-in per connection** (Edit connection → Claude section, shown for OAuth connections only; stored as `providerSpecificData.lowPriorityMode` / `autoLimitReset`, default **off**):

1. **Lower-priority lane** (`/low-priority`) — keep working past the 5-hour usage limit at lower priority instead of cooling the account down.
2. **Session-limit reset** (`/limit-reset`) — claim the once-a-week 5-hour window reset and continue at full speed.

Wire contract captured from Claude Code 2.1.263 (`anthropic-usage-limit` / `anthropic-ratelimit-unified-slow-*` / `juniper_tide`).

## Behavior

Nothing changes until an account hits its **5-hour usage wall**: a `429` carrying `anthropic-ratelimit-unified-status: rejected` and, when the account is eligible, `anthropic-ratelimit-unified-slow-offer: treatment`. A burst 429 without unified headers goes through the normal cooldown path, unchanged.

**Lower-priority lane** (`lowPriorityMode`):

- On the wall 429 the executor accepts the offer and immediately retries the **same account** with `anthropic-usage-limit: slow`. The lane stays active until the announced `anthropic-ratelimit-unified-reset` (+60 s grace) and every request in that window carries the header.
- The intercepted 429 never reaches `handleChatCore`, so the connection is **not** put in cooldown and is **not** rotated away (this was the point: with the lane active the account is still usable, so it must not be marked unavailable).
- While active: `slot_busy` (429) or `529` wait the server's `slow-retry-after` (default 20 s, clamp 5–600 s, ±30 % jitter) and retry the same account, bounded by `slow-max-wait` (default 20 min, clamp 1 min–6 h) — past that the lane ends and a 10-minute cool-off blocks re-acceptance. `active` / `not_needed` keep the lane. `weekly_limit` / `budget_exhausted` / `off` / `ineligible`, a 5h-window rollover, or `ineligible` + `overage-in-use: true` end the lane and the response flows to the normal cooldown path. `budget_exhausted` is remembered until the announced budget reset (≤ 8 days).
- State is in-memory per connection (a restart costs one extra wall 429 to re-accept).
- Race-safe for parallel requests on one connection: the executor tells the decider whether *this* request carried the slow header, so a sibling's header-less wall 429 (built while the lane was still idle) joins the lane instead of being misread as a "wall" verdict that would end it.

**Session-limit reset** (`autoLimitReset`, tried before the lane when both are on):

- `GET https://api.anthropic.com/api/oauth/usage?at_wall=1&skip_spend=1` → `juniper_tide` block; when `arm: "reset"` and `available: true`, `POST https://api.anthropic.com/api/organizations/{orgUUID}/reset_rate_limits` with `{ "program": "juniper_tide" }` (org UUID from `providerSpecificData.organizationUUID`, bootstrap fallback).
- `result: reset | not_limited` → the request is retried at full speed (no slow header). `already_used` / not offered memoise `next_available_at` (default one week); any failure backs off 15 min. Once a week, still counts toward the weekly limit.
- Concurrent wall hits on one connection share a single status + claim round trip (no duplicate `POST reset_rate_limits`), and for 60 s after a granted reset stale sibling walls are answered "reset" without touching the network.

## Where

| File | Change |
| --- | --- |
| `open-sse/services/claudeLowPriority.ts` (new) | Pure state machine + header parsing + per-response decision (`handleClaudeUsageLimitResponse`) |
| `open-sse/services/claudeLimitReset.ts` (new) | Status/claim HTTP client + per-connection "not before" memo |
| `open-sse/executors/base.ts` | Hook in `execute()` (Claude OAuth only): header injection when the lane is active; on the wall/slot_busy/529 → abort-aware wait + same-URL retry (`urlIndex--; continue`) |
| `src/lib/providers/requestDefaults.ts` | Persist the two opt-ins only as real booleans (same pattern as `disableCooling`) |
| `EditConnectionModal.tsx` | Two toggles in the Claude section (OAuth connections only) |
| `src/i18n/messages/en.json`, `it.json` | 4 keys in `providers` |
| `docs/architecture/RESILIENCE_GUIDE.md` | New subsection under §2 Connection Cooldown |
| `changelog.d/features/claude-low-priority-mode.md` | Fragment |

Not sent / not needed: OmniRoute never emits `anthropic-dispatch-id`, and `/v1/messages?beta=true` is already the Claude URL. Claude Code's client-side telemetry (`tengu_lowpri_*` events), remote copy gates (`tengu_toasty_breeze`, `tengu_nifty_lemur`) and the hidden "continue" prompt are client UX and intentionally not replicated.

## Verification

- `node --import tsx/esm --test tests/unit/claude-low-priority-mode.test.ts tests/unit/claude-limit-reset.test.ts tests/unit/claude-low-priority-executor.test.ts` → 38/38 pass
  - state machine: acceptance gate, clamps, lazy expiry, slot_busy/529 waits + max-wait cool-off, every end reason, budget memo, rollover, per-connection isolation
  - limit reset: status/claim parsing, claim URL + body + bearer, memo horizons, 401/403/429 handling
  - executor: wall 429 → same-account retry carries `anthropic-usage-limit: slow`, 200 returned (no 429 surfaced → no cooldown); header absent before the wall and for opt-out connections; control arm not accepted; per-connection state; auto-reset path (status + claim + full-speed retry)
- Related suites still green: `context-editing-executor-injection`, `claude-beta-flags-2454`, `anthropic-header-dedupe-1475`, `requestDefaults` (44/44)
- `npm run lint` (project suppressions) 0 errors on changed files; Prettier clean; `npm run check:docs-all` pass; `npm run check:changelog-integrity` pass

⚠️ base-red inherited: #12732


davidebaraldo:
## Review pass + CI triage

Ran a 3-lens adversarial review (contract fidelity vs the Claude Code 2.1.263 wire behavior, executor regression risk, robustness) over the branch. Three defects were confirmed in code and fixed in `e2b331d`:

1. **The lane wait could outlive the request** (high). The `slot_busy`/529 sleep shares the request's `AbortSignal` with `executeWithUpstreamStartTimeout`'s one-shot timer (10 min via `getExecutorTimeoutMs` by default), while the lane's own `slow-max-wait` defaults to 20 min and can reach 6 h from the server header. A long `slot_busy` streak was therefore killed mid-sleep with a `TimeoutError` instead of ending as `max_wait` + cool-off. The decision now takes a `waitCeilingMs` (what is left of the executor's timeout, minus a 5 s margin) that caps the effective max-wait and clamps every individual sleep.
2. **A wall 429 surfacing only after a 400-driven intra-attempt retry was missed** (medium). The context-editing, thinking-budget, effort and auto-learn fallbacks all re-fetch and reassign `response`; the wall check ran before them, so such a 429 fell through to the generic path and cooled the connection down — the opposite of the feature's point. The check now runs after those retries, on the attempt's final response. Regression test included.
3. **`ineligible` + `overage-in-use: true` ended the lane as plain `ineligible` on a 429** (medium), because the status mapping ran before the overage check and only the non-429 tail produced `extra_usage`. Overage takeover now wins on every status — `extra_usage` is the state `src/lib/providers/claudeExtraUsage.ts` keys on.

Also bounded the module-level per-connection maps with the same FIFO policy as the identity caches in `claudeIdentity.ts`: the state key falls back to the access token when no connection id is present, and OAuth tokens rotate on refresh, so the maps could otherwise grow for the process lifetime.

### Verification after the fixes

- `typecheck:core` clean, `check:cycles` clean, ESLint 0 (project suppressions), Prettier clean
- 46/46 in `claude-low-priority-mode`, `claude-limit-reset`, `claude-low-priority-executor`

### Red checks on this PR are inherited

- **Docs Gates** — reproduced locally: three strict drifts, all "171 migrations" vs 172 real, in `README.md`, `AGENTS.md` and `llm.txt`. This branch touches no migration and none of those three files.
- **Unit fast-path** — the failures observed are the same docs-counts gate plus an unrelated HuggingFace-image test.
- **Fast Quality Gates / No new ESLint warnings** — ratchet drift, which #12732 states explicitly is expected mid-cycle and rebaselined at release.

`⚠️ base-red inherited: #12732`

davidebaraldo:
## Verification pass

### 1. The spoofed client version is not a blocker

The biggest open risk was that Anthropic gates the lower-priority lane on a minimum Claude Code version, the way it gated `claude-fable-5-1` (#12366). It does not apply here: I downloaded `@anthropic-ai/claude-code-win32-x64@2.1.258` — the exact version `CLAUDE_CODE_CLIENT_VERSION` pins — and the whole feature set is already present in that build:

```
anthropic-usage-limit                          FOUND in 2.1.258 and 2.1.263
anthropic-ratelimit-unified-slow-offer         FOUND in both
anthropic-ratelimit-unified-slow-status        FOUND in both
anthropic-ratelimit-unified-slow-max-wait      FOUND in both
tengu_toasty_breeze / tengu_lowpri_accepted    FOUND in both
juniper_tide / reset_rate_limits / at_wall=1   FOUND in both
```

No client-version bump is needed for this PR.

### 2. The constants match the reference build exactly

The state-machine block in 2.1.258 carries the same numbers this PR implements: `20000` retry-after with a `5000`–`600000` clamp, `1200000` max-wait with a `60000`–`21600000` clamp, `0.3` jitter, `60000` reset grace, `60`s rollover tolerance, 10-minute cool-off (capped at 1440). The `slow-status` enum is the same seven values plus the unrecognized fallback.

Header attachment also matches: the reference sets the header on every dispatch while the lane reports active, through a lazily-expiring "is active" read — which is exactly the shape here. (It also has a second value, `anthropic-usage-limit: extended`, for a separate grace-zone feature. Out of scope; not implemented.)

### 3. Every red check on this PR is inherited — measured, not assumed

I ran the exact set of CI-failing test files on this branch and on a detached worktree at the untouched base `ba597b6`. Identical result on both: **12 failures, same names** (Antigravity 422, chatCore per-model quota, chatCore GitHub refresh, handleChat retry-after ×2, HuggingFace image 400, hard-lease inventory, managed request surfaces, HuggingChat boundaries, Kiro stream, generic stream boundaries, docs-counts gate). The base worktree has been removed.

The one failure that was genuinely mine is fixed in `471812e`: `tests/unit/i18n-pt-br.test.ts` (#6695) requires `pt-BR.json` to carry every `en.json` key, and the four new toggle strings were only in en and it. pt-BR now passes, and `check-ui-keys-coverage --threshold=65` passes for all 41 locales.

### 4. Full gate run after the fixes

- `typecheck:core` clean, `check:cycles` clean, `check:any-budget:t11` PASS, `check:tracked-artifacts` OK
- Prettier clean, ESLint 0 with the project suppressions
- Vitest: 51 files, 465 tests, all pass
- node:test on the feature plus every neighbouring executor/normalizer suite: 168 tests, 0 failures
- Flags verified end to end: `normalizeProviderSpecificData` keeps only real booleans and drops `"true"`/`1`; `sanitizeProviderSpecificDataForResponse` preserves both flags while stripping tokens; both toggles render only for a Claude OAuth connection

### 5. What is still unproven

No request has ever gone to Anthropic with this code. The offer is server-side and per-account, so the end-to-end claim needs one live observation: drive a Claude OAuth connection to its 5-hour wall with the toggle on, then check the 429 for `anthropic-ratelimit-unified-slow-offer: treatment` and the logs for the `CLAUDE_LOW_PRIORITY` line and a 200 on the retry. Per Hard Rule #18 this is the OAuth-upstream category that wants a documented live test, so I would rather record that result here before merge.

`⚠️ base-red inherited: #12732`

davidebaraldo:
## Quality gates: the two this PR actually broke are fixed

`f9b9b30` brings Fast Quality Gates down from four failing gates to two, and the two that remain are not this branch's:

```
2 gate(s) failed: mutation-test-coverage pack-policy
[complexity]           OK (código novo) — 14 violações nos arquivos tocados (base 14)
[cognitive-complexity] OK (código novo) — 7 violações nos arquivos tocados (base 7)
[file-size]            OK
```

### What was mine, and how it was fixed

**file-size** (frozen, may only shrink): `open-sse/executors/base.ts` had grown 1857 > 1751 and `EditConnectionModal.tsx` 1653 > 1631. Fixed by extraction, not by touching the baseline:

- the executor-side mechanics of the feature moved into a new `open-sse/executors/claudeUsageLimit.ts` (`ClaudeUsageLimitGuard`: header injection, wait accounting, abort-aware sleep, the timeout-derived wait ceiling, decision logging), leaving a three-line call site in `base.ts`;
- three long-standing Claude blocks left `base.ts` for the modules they belong to — `mergeCcHeaders` and `applyStainlessHeaders` into `open-sse/config/anthropicHeaders.ts`, `stripClaudeSystemPrefixBlocks` into `open-sse/executors/claudeIdentity.ts`. `base.ts` is back at its frozen 1750;
- the modal's Claude section became `ClaudeConnectionFields.tsx` (mirroring `CcCompatibleRequestDefaultsFields`) plus a `claudeConnectionFields.ts` helper that de-duplicates the field defaults across the modal's two init sites. The file is at 1622, below its frozen 1631.

**complexity / cognitive-complexity** (new-code mode): three functions were over the 15 threshold — `runClaudeLimitResetAttempt` (27), `handleClaudeUsageLimitResponse` (19, cognitive 23), `observeClaudeLowPriorityResponse` (17, cognitive 17). Split into focused helpers (`observeErrorResponse` / `observeSuccessResponse`, `shouldClaimLimitReset`, `resolveLimitResetOffer` / `runLimitResetClaim` / `memoiseNotBefore`). Zero new violations in either ratchet.

All of it is behavior-preserving: 155 node:test cases across the feature and its neighbouring executor, header and normalizer suites stay green, Vitest is 465/465, typecheck clean, ESLint 0 with the project suppressions, Prettier clean. `check:fetch-targets`, `check:error-helper`, `check:build-scope`, `check:deps` all OK.

### What is not mine

- **pack-policy** — `❌ Unexpected files were found in the npm publish artifact`, listing only `@omniroute/opencode-plugin-v2/**`. Reproduced locally on this branch with the same list; nothing from this PR appears in it.
- **mutation-test-coverage** — two covering unit tests missing from `stryker.conf.json` `tap.testFiles`, for `src/shared/utils/circuitBreaker.ts` and `open-sse/services/combo/comboStructure.ts`. Neither module nor test belongs to this PR.
- **Docs Gates / Unit shards / No new ESLint warnings** — measured against a detached worktree at the untouched base `ba597b6`: identical 12 unit failures, the same 171-vs-172 migration-count drift in `README.md` / `AGENTS.md` / `llm.txt`, and the ESLint warnings are `tests/unit/volcengine-plan-binding-upsert.test.ts`.

Per the base-red rule these belong in their own `fix/release-v3.8.51-basereds` PR rather than in a feature branch; happy to open it if wanted.

`⚠️ base-red inherited: #12732`

diegosouzapw:
Well-executed feature — clean state-machine isolation (`claudeLowPriority.ts`/
`claudeLimitReset.ts`), both opt-ins default off so nothing changes unless an operator enables
them per connection, and the security posture is right (the `slow` header is server-controlled,
never a client-echoed value). Ran all three test files at this PR's head: 42/42 pass, covering
the acceptance gate, clamps, lazy expiry, slot_busy/529 waits, every lane-ending reason, and
per-connection isolation. Two small things: could you run the i18n sync for the 4 new
`providers` keys across the other 40 locales (currently only en/it have them)? And is the
`base.ts` helper extraction (`stripClaudeSystemPrefixBlocks`/`mergeCcHeaders`/
`applyStainlessHeaders`) meant to ship with this feature, or would it read cleaner as its own
small refactor PR? Neither is a blocker — solid work overall.

diegosouzapw:
Thanks @davidebaraldo — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-130526-suite.log) on the devbox @ train tip 28fb420c9ad860ba275294ebb4ebbecb9da83318 with the sibling PRs of this batch: static gates green; changed-area node:test 293/294 (0 failing) + vitest 482/482 (fast parity — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §7.

PR #13076, opened 2026-09-08
nix: add fixed-output package derivation

### LINKED ISSUE #13075: feat: Provide a Nix package derivation for OmniRoute
## Summary

Nix / NixOS users currently have no first-class way to install OmniRoute reproducibly — the options are `npm install -g` inside a stateful environment or ad-hoc wrappers, neither of which fit the Nix model.

## Technical obstacle

The obvious path (`buildNpmPackage`) doesn't work out of the box. The `package-lock.json` shipped in the published npm tarball references workspace packages — `@omniroute/browser-pool`, `open-sse`, and `packages/browser-pool` — that `npm ci --offline` cannot resolve inside the Nix build sandbox (no network, no workspace tree).

## Proposed approach

A fixed-output derivation (FOD) that shells out to `npm install -g omniroute@<version>` with network access enabled, pinning reproducibility via `outputHash`. This sidesteps the workspace-resolution problem entirely while still giving Nix users a reproducible, hash-pinned install. Version bumps require re-computing the hash, which is a small tradeoff.

I've prototyped this locally against v3.8.50 and it works. Happy to open a PR adding `nix/omniroute.nix` and a minimal `flake.nix` if that's useful — will follow up shortly. No pressure to accept; feel free to relocate or restructure freely.

Thanks for OmniRoute!

Ansh-Sonkusare:
## Summary
- Adds a `flake.nix` and `nix/omniroute.nix` so Nix users can install OmniRoute reproducibly.
- Uses a fixed-output derivation (FOD) because `package-lock.json` references workspace packages (`@omniroute/browser-pool`, `open-sse`, `packages/browser-pool`) that `npm ci --offline` can't resolve inside the Nix sandbox. The FOD runs `npm install -g omniroute@<version>` with network access and pins reproducibility via `outputHash`.
- Adds a short "Nix / NixOS" section to the README.

Closes #13075.

## Test plan
- [ ] `nix build .#default` on `x86_64-linux`
- [ ] `nix run .#default -- --help` prints usage
- [ ] Verify the outputHash matches (rebuild after `rm -rf ~/.cache/nix` — should be a substitute hit or match on fresh build)

## Notes
This is an offer, not urgent — happy to move files around (`packaging/nix/`, subdir naming, README placement) or drop the flake and keep just the callPackage file if you prefer. Version bumps require regenerating `outputHash`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1087,13 +1087,23 @@ yay -S omniroute-bin && systemctl --user enable --now omniroute.service
 **🔧 Nix (Flake)**
 
 ```bash
-# Using Nix flakes
+# Install the package (requires flakes enabled)
+nix profile install github:diegosouzapw/OmniRoute
+# Or run ad-hoc without installing:
+nix run github:diegosouzapw/OmniRoute -- --help
+
+# Dev shell for hacking on the source:
 nix develop
 npm run dev
 
 # Or using devbox
 devbox run npm run dev
 ```
+
+The package derivation is a fixed-output derivation (FOD) because OmniRoute's
+`package-lock.json` references workspace packages that `npm ci --offline`
+cannot resolve inside the Nix sandbox. Version bumps require updating the
+`outputHash` in `nix/omniroute.nix`.
 
 📖 [Docker Guide](docs/guides/DOCKER_GUIDE.md) — Compose profiles, Caddy HTTPS, Cloudflare tunnels.
 
--- a/flake.nix
+++ b/flake.nix
@@ -1,5 +1,5 @@
 {
-  description = "OmniRoute - Unified AI router with 160+ providers";
+  description = "OmniRoute — free MIT AI gateway: one endpoint, 350+ providers, 1200+ models with auto-fallback";
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.11";
@@ -13,6 +13,9 @@
         nodejs = pkgs.nodejs_22;
       in
       {
+        packages.default = pkgs.callPackage ./nix/omniroute.nix { inherit nodejs; };
+        packages.omniroute = self.packages.${system}.default;
+
         devShells.default = pkgs.mkShell {
           buildInputs = [
             nodejs
@@ -21,7 +24,7 @@
           shellHook = ''
             echo "Welcome to OmniRoute dev environment"
             export PATH="$PWD/node_modules/.bin:$PATH"
-            
+
             # Install dependencies if node_modules doesn't exist
             if [ ! -d "node_modules" ]; then
               echo "Installing dependencies..."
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks — this is a clean, idiomatic fixed-output derivation for the workspace-lockfile
problem, and `--ignore-scripts` is a nice touch security-wise. Two things before merge: the
pinned version in `nix/omniroute.nix` is `3.8.50`, which is now behind `release/v3.8.51` — will
need a version bump + a recomputed `outputHash`. And since this sandbox doesn't have a Nix
toolchain, I wasn't able to actually run `nix build .#default` to confirm the hash — could you
(or whoever merges) do one real build/run before landing this, per your own test plan?

Ansh-Sonkusare:
Thanks for the review. I ran the real build on x86_64-linux (Nix 2.34) and it turned up more than the version bump, so I pushed a fix (8688677).

**What the real build found**
- The pinned `outputHash` was wrong (Nix reported a different hash).
- The FOD ran `fixupPhase`/`wrapProgram`, which rewrote shebangs to `/nix/store` bash paths. That ties the hash to the nixpkgs bash version and puts store references into a fixed-output derivation.
- The published tarball has no lockfile, so transitive deps float to whatever is newest at build time and would drift the hash.
- The tree contains platform-specific binaries (`sharp`, `@next/swc`, `esbuild`), so a single hash can't be valid for both linux and darwin.

**What changed**
- Split into a network FOD (`npm install -g`, fixup disabled, `--before=<publishedAt>` to pin resolution) and a plain derivation that wraps the bins with the matching Node without copying the multi-GB tree.
- Hashes are now per system. Only `x86_64-linux` is listed. Other systems are refused by nixpkgs' platform check until someone builds on them and adds a hash.
- README updated to match.

**Verified on x86_64-linux**
- `nix build .#default` succeeds; `omniroute --version` and `--help` work.
- The server starts from the read-only store, and login plus password and settings changes work and persist across a restart (state goes to `DATA_DIR`).
- `nix build --rebuild` on the FOD reproduces the same hash on a fresh fetch.

**Version:** I left it at 3.8.50. `omniroute@3.8.51` isn't published to npm yet (latest is 3.8.50, and 3.8.51 returns a 404), and this derivation installs from npm, so it can't be bumped until that release is published. After that it's a `version` + `publishedAt` + hash update; the steps are in the file header.

Not verified: aarch64-linux and darwin, since I only have an x86_64-linux machine.

Ansh-Sonkusare:
Ran the real build: `nix build .#default` builds and runs on x86_64-linux (server starts, login and password change work, data persists in `DATA_DIR` across restarts).

On the bump: `omniroute@3.8.51` isn't on npm yet (latest is 3.8.50, and 3.8.51 returns 404) and there's no `v3.8.51` tag, so the package stays on 3.8.50 for now.

I also moved the package off the fixed-output `npm install -g` to `buildNpmPackage` + `importNpmLock` with a generated lockfile. A global install ignores the manifest's `overrides` (the old build shipped `adm-zip` 0.5.18 against the `^0.6.0` pin). This one applies them, and there's no output hash anymore.

Could the Nix bump go into each npm release? After publishing:

```
node scripts/release/update-nix-package.mjs <version>
```

It needs npm only, no Nix. It regenerates `nix/package.json`, `nix/package-lock.json` and the tarball hash, and fails early if the lockfile wouldn't install offline.

PR #13078, opened 2026-09-08
fix(providers): honor operator-set endpoint overrides for local models

hartmark:
## What Problem This Solves

A self-hosted server like llama.cpp reports **no capability data** on its own
`/v1/models` response -- unlike Ollama (already fixed for this exact class in
#11087 via an `/api/show` probe), llama.cpp has no equivalent endpoint to
probe. So a discovered embedding model's synced cache entry
(`syncedAvailableModels`) never gets a `supportedEndpoints` of its own, and
the operator's only recourse is the explicit "Supported endpoints" override
in the Providers UI (`PUT /api/provider-models`).

That override was broken **end to end** for exactly this case:

1. `updateCustomModel()` (backing `PUT /api/provider-models`) 404'd for any
   model that had never been explicitly imported into the `customModels`
   store before -- which is every auto-discovered model. The very save the
   UI offers for "my server doesn't report this, let me tell you myself"
   failed silently with `"Model not found"`.
2. Even with a saved override, `/v1/embeddings`' dynamic-provider routing
   (`resolveLocalSyncedEndpointRoute`) only ever consulted the raw,
   un-annotated `syncedAvailableModels` cache -- never the `customModels`
   override the `/v1/models` catalog itself already merges in for display.
   An operator could watch the catalog correctly show "Embeddings" for their
   model and still get `"Unknown embedding provider"` on the actual request.

## Fix

- **`models.ts`**: `updateCustomModel()` takes an opt-in
  `{ createIfMissing: true }` option that upserts (seeding the same default
  shape `addCustomModel()` uses) instead of returning `null` when no
  `customModels` row exists yet for this model. The existing null-return
  contract is preserved by default -- the `PATCH` `isHidden` caller
  (`route.ts`) still relies on it to trigger its own compat-override
  fallback, so it keeps calling the function unchanged.
- **`provider-models/route.ts`**: the `PUT` handler opts into
  `createIfMissing`, fixing the 404 on the very save this endpoint exists to
  serve.
- **`syncedEndpointRouting.ts`**: `resolveLocalSyncedEndpointRoute()` now
  also checks for a `customModels` override matching the model's exact
  catalog id and honors its `supportedEndpoints`, the same way the
  `/v1/models` catalog already merges `customModels` on top of synced
  entries for display.

## Evidence

New regression tests pin both fixes against the real observed shape: a
llama-cpp connection with one synced model (`source: "imported"`, no
`supportedEndpoints`). They prove: (a) `updateCustomModel` upserts with
`createIfMissing` but still returns `null` without it (preserving the PATCH
caller's contract); (b) an override rescues routing for a model with no
endpoints of its own; (c) an override on a different model id does not leak
onto this one; (d) a model's own genuine `supportedEndpoints` still work
with no override needed. Confirmed both directly-relevant tests fail
against the pre-fix code and pass after.

```
$ node --test tests/unit/local-provider-embedding-override.test.ts
tests 5, pass 5, fail 0

$ node --test tests/unit/db-models-crud.test.ts tests/unit/db-models-split.test.ts \
    tests/unit/provider-models-custom-merge-6247.test.ts tests/unit/provider-models-token-limits.test.ts \
    tests/unit/synced-model-delete-custom-sibling.test.ts tests/unit/auto-combo-hidden-models-4558.test.ts \
    tests/unit/model-catalog-source-invalidation-8728.test.ts tests/unit/models-db-isfree.test.ts
tests 44, pass 44, fail 0
```

- `eslint` on all touched files: clean

**Live-verified** against the exact user-reported request (deployed before
opening this PR, given it was actively blocking a user): saving the
"Embeddings" override on a llama.cpp connection's discovered model went from
404 to 200, and the subsequent `/v1/embeddings` call against that exact
model id went from `400 "Unknown embedding provider"` to `200` with a real
embedding vector.

### DIFF
--- a/src/app/api/provider-models/route.ts
+++ b/src/app/api/provider-models/route.ts
@@ -258,7 +258,7 @@ export async function PUT(request) {
       }
     }
 
-    const model = await updateCustomModel(provider, modelId, updates);
+    const model = await updateCustomModel(provider, modelId, updates, { createIfMissing: true });
 
     if (!model) {
       const rawKeys = Object.keys(raw);
--- a/src/lib/db/models.ts
+++ b/src/lib/db/models.ts
@@ -775,20 +775,37 @@ function applyTriStateBooleanOverride(
 export async function updateCustomModel(
   providerId: string,
   modelId: string,
-  updates: Record<string, unknown> = {}
+  updates: Record<string, unknown> = {},
+  options: { createIfMissing?: boolean } = {}
 ) {
   const db = getDbInstance();
   const row = db
     .prepare("SELECT value FROM key_value WHERE namespace = 'customModels' AND key = ?")
     .get(providerId);
-  if (!row) return null;
 
-  const value = getKeyValue(row).value;
-  if (!value) return null;
+  const value = row ? getKeyValue(row).value : null;
+  const models: JsonRecord[] = value ? JSON.parse(value) : [];
+  let index = models.findIndex((m: JsonRecord) => m.id === modelId);
 
-  const models = JSON.parse(value);
-  const index = models.findIndex((m: JsonRecord) => m.id === modelId);
-  if (index === -1) return null;
+  if (index === -1) {
+    if (!options.createIfMissing) return null;
+    // A model discovered via sync/passthrough (syncedAvailableModels) has no
+    // customModels row until an operator explicitly overrides one of its
+    // fields -- PUT /api/provider-models is exactly that "set an override"
+    // action, so upsert here (same default shape as addCustomModel()) instead
+    // of 404ing on the very save it exists to serve. Observed live: a
+    // llama.cpp connection's auto-discovered embedding model had no way to be
+    // marked "supports embeddings" because it had never been explicitly
+    // imported as a custom model first.
+    models.push({
+      id: modelId,
+      name: modelId,
+      source: "manual",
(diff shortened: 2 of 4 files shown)


hartmark:
Follow-up commit: accept a collapsed single-slash id too.

llama.cpp's own `/v1/models` reports models by absolute filesystem path (`/models/Qwen3-Embedding-4B-Q8_0.gguf`), so `<alias>/<rawId>` naturally reads as `llamacpp//models/foo.gguf` (double slash) -- correct, but an operator naturally collapses it to a single slash by hand, which 404'd. `resolveLocalSyncedEndpointRoute()` now tries both forms; the catalog's own displayed id is unchanged. New test + live-verified both forms now resolve to the same model/connection.

diegosouzapw:
Nice, well-tested fix — the code comments explain the real-world llama.cpp scenarios clearly,
and the `createIfMissing` opt-in keeps `updateCustomModel()`'s existing contract intact for
every other caller (confirmed by your own "unchanged contract" test). Ran the full new test
file locally: 6/6 pass. Please add a changelog fragment under `changelog.d/fixes/` before
merge.

hartmark:
Added the changelog fragment — `changelog.d/fixes/13078-local-provider-embedding-override.md`, verified with `check:changelog-integrity`. Ready to merge from my side.

PR #13079, opened 2026-09-08
fix(db): gate the auto-cleanup VACUUM on reclaimable space, not row count

hartmark:
## What Problem This Solves

The auto-cleanup scheduler ran a full `VACUUM` after **any** deletion at all
(`totalDeleted > 0`). `db.exec("VACUUM")` is synchronous and blocks the
entire process -- on a multi-GB database that freezes all HTTP traffic for
15-20 minutes, no matter how small the cleanup that triggered it.

Observed live: a routine cleanup that freed **2-6 rows** re-triggered a full
VACUUM on every single restart, because a sibling terminal-batch cleanup fix
almost always finds a handful of newly-aged-out batches by the time the
server comes back up. Every restart became a 15-20 minute outage.

## Root Cause

Row count was never the right signal for "is a multi-minute freeze worth it"
anyway: a few oversized `batch_item_checkpoints` rows (each holding a
`result_json` blob) can free far more space than thousands of tiny
audit-log rows.

## Fix

Gate the VACUUM on SQLite's own free-page count (`PRAGMA freelist_count`)
instead, converted to bytes via `PRAGMA page_size`, and skip the VACUUM
entirely when there isn't at least `OMNIROUTE_VACUUM_MIN_RECLAIMABLE_MB`
(default 100 MB) worth of free pages to reclaim.

`vacuumIfWorthwhile()` replaces the inline VACUUM block at both call sites
(startup cleanup and the 6-hourly periodic cleanup) with no other behavior
change: cleanup itself still runs on the same schedule, only the decision
to pay for a VACUUM afterward changed. Setting the env var to `0` restores
the old unconditional behavior exactly.

## Evidence

New regression tests prove the env var parses correctly (including `0` to
always vacuum), that a handful of tiny deleted rows leaves the freelist
untouched by `vacuumIfWorthwhile()` when under the threshold, and that a
genuinely large deletion still gets vacuumed once the threshold is `0`
(freelist observably drops to zero after VACUUM).

```
$ node --test tests/unit/vacuum-reclaimable-threshold.test.ts
tests 5, pass 5, fail 0

$ node --test tests/unit/db-cleanup.test.ts tests/unit/db-cleanup-xp-audit-log.test.ts \
    tests/unit/telemetry-auto-cleanup-6848.test.ts tests/unit/repro-9624.test.ts
tests 26, pass 26, fail 0
```

- `eslint`: clean
- `node scripts/check/check-env-doc-sync.mjs`: in sync

**Live-verified**: after this fix, a restart whose cleanup freed 2 rows
logged `"Skipping VACUUM: only 0.0 MB reclaimable (threshold 100 MB)"` and
the server stayed responsive throughout, versus the prior behavior of
freezing for the full VACUUM duration regardless.


HouMinXi:
Thanks for gating VACUUM on reclaimable free pages instead of row count.

Folded into a local recut of `release/v3.8.51`. The conflict was one region in `src/lib/db/cleanup.ts` plus the matching `ENVIRONMENT.md` rows, after #13011 (row-count gate) had already landed on that recut. Resolution: keep both gates. Row-count runs first; reclaimable-bytes is the fallback for a few oversized rows. A VACUUM that already ran zeros the freelist, so the second gate is a no-op.

`.env.example` documents both `OMNIROUTE_VACUUM_MIN_DELETED_ROWS` (1000) and `OMNIROUTE_VACUUM_MIN_RECLAIMABLE_MB` (100).

Purpose of the recut: ship current `release/v3.8.51` plus open product PRs on a private production host. If you rebase onto current `release/v3.8.51`, please keep the reclaimable-bytes helpers next to the existing row-count gate rather than replacing it.

diegosouzapw:
Good diagnosis and a clean fix for the row-count-vs-space mismatch. Two things
changed since this was opened, though: #13011 (merged 09-11) already added a
row-count-based VACUUM gate at these same call sites — mitigating the acute "2-6
rows shouldn't trigger VACUUM" case — and #12830 in this same review batch goes
further, replacing the blocking VACUUM entirely with non-blocking incremental
reclamation (reading the real freelist rather than approximating via a
threshold). I'm recommending we close this in favor of #12830, which supersedes
both the merged row-count gate and this byte-threshold approach. If the owner
ends up preferring a simpler threshold-only option, your `vacuumIfWorthwhile()`
and its 5-case test would be a good addition to #12830's config surface — happy
to see that ported with credit to you.

hartmark:
Agreed — #12830's incremental reclamation is the more thorough fix (reads the real freelist rather than approximating via a threshold). I'll hold off closing this until #12830 actually lands, per your own sequencing note. Happy to have `vacuumIfWorthwhile()` and its test ported over with credit if a simpler threshold option ends up wanted alongside it.

PR #13091, opened 2026-09-09
fix(compression): terminate idle workers on eviction

### LINKED ISSUE #12812: fix(backend): Compression worker pool leaks worker threads: idle eviction never terminates (5.7GB commit / 55 orphaned MessagePorts after 16h)

anhtahaylove:
`CompressionWorkerPool.finish()` scheduled idle eviction as `remove(slot, false)`, so the timer dropped the slot from the pool without calling `worker.terminate()`. The OS thread, its `MessagePort` and its private heap then lived for the rest of the process.

That is invisible to V8-level metrics, which is what made it hard to see from inside: the reported instance showed `rss=660MB` while the OS held 5.7GB of commit charge with 55 orphaned `MessagePort` handles after 16 hours.

## The change

Removal always terminates now, and the `terminate` parameter is gone rather than left in place as a footgun. A pooled worker has no owner besides the pool, so there was no caller that legitimately wanted the slot dropped but the thread kept — the only `false` call site was the leak itself. The other two call sites already passed `true`.

## Test

`tests/unit/compression/worker-pool-idle-eviction-12812.test.ts` waits past the idle window and asserts the worker's `exit` event fires. That event is the only thing separating a terminated thread from a merely dereferenced one — asserting the slot left the pool would pass on the buggy code too, since dropping the reference is exactly what it already did.

Confirmed it fails on the old behaviour: re-disabling the `terminate()` call gives

```
✖ terminates the worker thread when the idle timer fires
  ✔ close() terminates every pooled worker
ℹ pass 1  ℹ fail 1
```

and passes with the fix.

## Verification

```
tests/unit/compression/**    1503 passed | 5 failed
npx tsc -p tsconfig.typecheck-core.json --noEmit   clean
npx eslint <changed files>                          clean
```

The 5 failures are all in the RTK suites (`rtk-mcp-tools`, `rtk-toml-schema-v1`, `rtk-toml-compatibility`) and are pre-existing: they reproduce identically on a clean `release/v3.8.51` checkout with my changes stashed (`4 failed` in those two files alone), and none of them touch the worker pool.

Fixes #12812

### DIFF
--- a/changelog.d/fixes/compression-pool-idle-terminate.md
+++ b/changelog.d/fixes/compression-pool-idle-terminate.md
@@ -1,0 +1,1 @@
+- fix(compression): terminate idle worker threads on eviction so long-running instances stop leaking OS threads and MessagePorts
--- a/open-sse/services/compression/compressionWorkerPool.ts
+++ b/open-sse/services/compression/compressionWorkerPool.ts
@@ -131,7 +131,7 @@ export class CompressionWorkerPool {
   }
   async close(): Promise<void> {
     for (const job of this.queue.splice(0)) job.resolve(unchanged(job.originalBody));
-    await Promise.all([...this.workers].map((slot) => this.remove(slot, true)));
+    await Promise.all([...this.workers].map((slot) => this.remove(slot)));
   }
   private spawn(): PoolWorker {
     const slot: PoolWorker = {
@@ -185,7 +185,10 @@ export class CompressionWorkerPool {
     slot.timeout = null;
     slot.job = null;
     job.resolve(result);
-    slot.idle = setTimeout(() => void this.remove(slot, false), this.idleMs);
+    // Idle eviction MUST terminate. Dropping the slot from the set only releases our
+    // reference - the thread, its MessagePort and its private heap outlive the pool
+    // for the whole process lifetime, invisible to process.memoryUsage(). (#12812)
+    slot.idle = setTimeout(() => void this.remove(slot), this.idleMs);
     slot.idle.unref();
     this.dispatch();
   }
@@ -193,13 +196,15 @@ export class CompressionWorkerPool {
     const job = slot.job;
     if (job) job.resolve(unchanged(job.originalBody));
     slot.job = null;
-    void this.remove(slot, true).finally(() => this.dispatch());
+    void this.remove(slot).finally(() => this.dispatch());
   }
-  private async remove(slot: PoolWorker, terminate: boolean): Promise<void> {
+  /** Drop a slot and release its OS thread. Removal always terminates: a pooled worker
+   *  has no other owner, so skipping terminate() strands the thread permanently. */
+  private async remove(slot: PoolWorker): Promise<void> {
     if (!this.workers.delete(slot)) return;
     if (slot.timeout) clearTimeout(slot.timeout);
     if (slot.idle) clearTimeout(slot.idle);
-    if (terminate) await slot.worker.terminate().catch(() => undefined);
+    await slot.worker.terminate().catch(() => undefined);
   }
 }
 
--- a/tests/unit/compression/worker-pool-idle-eviction-12812.test.ts
+++ b/tests/unit/compression/worker-pool-idle-eviction-12812.test.ts
@@ -1,0 +1,69 @@
+/**
+ * Regression guard for #12812: idle eviction must terminate the worker thread.
+ *
+ * Root cause: finish() scheduled `remove(slot, false)`, so the idle timer dropped the slot
+ * from the pool WITHOUT calling worker.terminate(). The OS thread, its MessagePort and its
+ * private heap then survived for the whole process lifetime. Nothing in
+ * process.memoryUsage() reports that, which is why a 16h instance showed rss=660MB while
+ * holding 5.7GB of commit charge.
+ *
+ * The assertion measures the real thing: a worker that was evicted must no longer be able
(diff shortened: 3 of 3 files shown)


anhtahaylove:

JasonBroderick:
Same runtime change as #12542 (idle path terminates, `terminate` flag removed), aimed at the release branch, with a test and a changelog entry: this is a fine one to merge for v3.8.51.

Field confirmation, self-hosted Docker, v3.8.51 source (`c3945a72`) plus this idle-path change since 2026-09-05, agentic workload through the default RTK/Caveman pipeline, no combos: server-process OS threads flat at 12 to 13 for five days (previously one more per idle eviction); the built bundle carries `remove(e,!0),this.idleMs`.

One caveat so it is not mis-filed against this PR: on that instance the V8 heap itself (heapUsed from `process.memoryUsage()`, not native) still climbs from ~200 MB to the admission guard's ceiling in 8 to 15 hours with threads flat. That is a separate retainer; heap-snapshot diffs are in progress and it will get its own issue.

Issue #13094, opened 2026-09-09
fix(deepseek): PoW worker slot counter leaks on any throw before cleanup, permanently disabling the solver after 2 failures


anhtahaylove:
## Summary

`solveInWorker` in `open-sse/lib/deepseek-pow.ts` increments `activeWorkerCount` *before* entering the promise executor, but the decrement lives in `cleanup()` inside that executor. Any throw in between leaks a slot permanently. With `MAX_CONCURRENT_WORKERS = 2`, two such throws disable DeepSeek PoW solving for the lifetime of the process.

## Mechanism

```ts
activeWorkerCount += 1;                       // line 106

return new Promise<number>((resolve, reject) => {
  const worker = new Worker(resolveWorkerPath(), { workerData: validated });
  //                        ^^^^^^^^^^^^^^^^^ throws if the script is missing
  const cleanup = () => { activeWorkerCount = Math.max(0, activeWorkerCount - 1); };
  ...
});
```

`resolveWorkerPath()` (line 85) throws when the worker script is absent:

```ts
if (!existsSync(workerPath)) {
  throw new Error(`DeepSeek PoW worker script not found at ${workerPath}`);
}
```

It resolves against `process.cwd()`, so this fires whenever the process is started from a different working directory — not an exotic condition. `new Worker(...)` itself can also throw (bad permissions, ENOENT race).

A throw inside the executor rejects the promise, so the caller sees a sensible error and nothing looks wrong — but `cleanup()` was never reached and the slot is gone.

## Reproduction

Isolating the counter logic:

```
attempt 1: ... | activeWorkerCount=1
attempt 2: ... | activeWorkerCount=2
attempt 3: capacity reached (2) | activeWorkerCount=2
attempt 4: capacity reached (2) | activeWorkerCount=2
```

After the second failure every later call rejects with `DeepSeek PoW worker capacity reached (2)` regardless of how healthy things are. Only a restart clears it. The reported error also misleads: it blames capacity, not the missing script.

## Suggested fix

Move the increment inside the executor's `try`, or decrement in a `catch` around worker construction — e.g. wrap construction so the counter is only owned once a worker actually exists:

```ts
return new Promise<number>((resolve, reject) => {
  let worker: Worker;
  try {
    worker = new Worker(resolveWorkerPath(), { workerData: validated });
  } catch (error) {
    activeWorkerCount = Math.max(0, activeWorkerCount - 1);
    reject(error);
    return;
  }
  ...
});
```

[...]

PR #13096, opened 2026-09-09
fix(acp): release listeners, timers and sessions on every sendPrompt outcome

### LINKED ISSUE #13095: fix(backend): acp: sendPrompt timeout path leaks listeners, timers and sessions on the acpManager singleton
## Summary

`AcpManager.sendPrompt` (`src/lib/acp/manager.ts`) registers two listeners on `this` and never removes them when the outer timeout fires. `acpManager` is a **module-level singleton**, so every timed-out prompt permanently retains listeners, a live idle timer, and its captured `session` object for the lifetime of the process.

## The leak

```ts
return new Promise((resolve, reject) => {
  const timer = setTimeout(() => {
    reject(new Error(`ACP timeout after ${timeoutMs}ms`));   // <-- only rejects
  }, timeoutMs);

  let idleTimer: ReturnType<typeof setTimeout>;
  const onData = ({ sessionId: sid }) => { ... };
  const onExit = ({ sessionId: sid }) => { ... };

  this.on("stdout", onData);
  this.on("exit", onExit);
});
```

`onData` and `onExit` each remove both listeners on their success paths, but the timeout callback does neither — and it also leaves `idleTimer` armed.

## Reproduction

Driving the exact `sendPrompt` shape through 12 timeouts:

```
MaxListenersExceededWarning: 11 stdout listeners added to [Mgr]. MaxListeners is 10
MaxListenersExceededWarning: 11 exit   listeners added to [Mgr]. MaxListeners is 10
stdout listeners retained: 12
exit   listeners retained: 12
```

Every listener stays attached forever. Consequences that compound:

1. **Listener growth** — the `MaxListenersExceededWarning` appears after 10 timeouts, then every subsequent `emit("stdout")` walks an ever-growing list. With one listener per timed-out prompt this is O(n) dispatch cost on the hot streaming path.
2. **Retained sessions** — each stale closure captures `session`, so the `AcpSession` and its buffers stay reachable even after the process exits and `alive` flips to `false`.
3. **Armed idle timer** — a leaked `idleTimer` can still fire 2s later and call `resolve()` on an already-rejected promise. Harmless to the caller (the settle is a no-op) but it keeps the timer and closure alive until it fires.

## Two adjacent issues in the same file

While confirming the above:

**a) `stdoutBuffer` / `stderrBuffer` are unbounded.** Both grow by string concatenation on every chunk with no cap:

```ts
child.stdout?.on("data", (chunk: Buffer) => { session.stdoutBuffer += chunk.toString(); ... });
child.stderr?.on("data", (chunk: Buffer) => { session.stderrBuffer += chunk.toString(); ... });
```

`stdoutBuffer` is reset at the start of each `sendPrompt`, but `stderrBuffer` is **never** reset or read — a chatty agent grows it until the session is killed.

**b) Exited sessions are never removed from the map.** `sessions.delete()` only happens in `kill()`. The `child.on("exit")` handler sets `session.alive = false` but leaves the entry in place, so `this.sessions` grows monotonically for any caller that lets agents exit on their own. `getActiveSessions()` filters on `alive`, which hides the growth.

## Suggested fix

Give the promise a single settle path that always tears down:

```ts
return new Promise((resolve, reject) => {
  const cleanup = () => {
    clearTimeout(timer);
    clearTimeout(idleTimer);
    this.removeListener("stdout", onData);
    this.removeListener("exit", onExit);
  };
  const timer = setTimeout(() => {
    cleanup();
    reject(new Error(`ACP timeout after ${timeoutMs}ms`));
  }, timeoutMs);
  ...
});
```

For (b), deleting the session in the `exit` handler (or marking it for reaping) would bound the map. For (a), a cap on both buffers with head/tail truncation.

## Related

Same family as #12812 / #12819 / #13094 — acquire-then-release paths that miss a branch. This one differs in that the leaked resource sits on a **singleton** emitter, so it accumulates across unrelated requests rather than per pool or per call.

Happy to send a PR with a regression test (the listener count assertion above makes a clean detector) if the suggested shape looks right.

anhtahaylove:
Fixes #13095.

## Problem

`AcpManager.sendPrompt` registers two listeners on `this` and removes them on its idle and exit paths — but the outer timeout callback only calls `reject()`:

```ts
const timer = setTimeout(() => {
  reject(new Error(`ACP timeout after ${timeoutMs}ms`));   // no teardown
}, timeoutMs);
```

Both listeners stay attached and `idleTimer` stays armed. `acpManager` is a **module-level singleton**, so this accumulates across unrelated requests for the process lifetime.

Driving 12 timeouts through the manager before the fix:

```
MaxListenersExceededWarning: 11 stdout listeners added to [AcpManager]
stdout listeners retained: 12
exit   listeners retained: 12
```

Each stale closure also captures its `session`, so the session object and its buffers stay reachable after the child exits.

## Fix

One `settle()` helper that every outcome — idle, exit, timeout — goes through:

```ts
const settle = (finish: () => void) => {
  clearTimeout(timer);
  clearTimeout(idleTimer);
  this.removeListener("stdout", onData);
  this.removeListener("exit", onExit);
  finish();
};
```

Also `this.sessions.delete(sessionId)` in the child's `exit` handler. Previously only `kill()` removed entries, so an agent that exited on its own stayed in the map forever — `getActiveSessions()` filters on `alive`, which hid the growth.

## Tests

`tests/unit/acp-manager-sendprompt-leak-13095.test.ts`, three cases:

1. 12 consecutive timed-out prompts must leave the listener counts exactly where they started.
2. A single timeout must leave nothing pending on the manager.
3. A session that exits on its own must not remain in the session map.

The tests spawn `process.execPath` with a small inline script (an idle process that never writes to stdout, so the prompt can only time out) via a test-only custom agent — no real CLI agent binary required.

Verified as a real detector: all three fail on `release/v3.8.51` before the change (`pass 0 / fail 3`) and pass after (`pass 3 / fail 0`).

## Scope

`fail 0` across the full ACP suite (`tests/unit/acp*.test.ts` + `cli-catalog-acpspawnable.test.ts`, 39 tests) — `kill()` and `killAll()` still behave correctly now that `exit` also removes the entry.

I did **not** touch the two adjacent problems noted in the issue — the unbounded `stdoutBuffer`/`stderrBuffer` growth, and `stderrBuffer` never being read or reset. Both need a product decision on the cap and truncation strategy, and neither is a listener leak, so they seemed better kept out of this diff. Happy to follow up if you want them handled.

### DIFF
--- a/changelog.d/fixes/13095-acp-sendprompt-listener-leak.md
+++ b/changelog.d/fixes/13095-acp-sendprompt-listener-leak.md
@@ -1,0 +1,1 @@
+Fix a listener, timer and session leak in the ACP manager: `sendPrompt` timeouts left their `stdout`/`exit` listeners attached to the `acpManager` singleton, and sessions that exited on their own were never removed from the session map.
--- a/changelog.d/fixes/13095-acp-sendprompt-listener-leak.md
+++ b/changelog.d/fixes/13095-acp-sendprompt-listener-leak.md
@@ -1,1 +1,1 @@
-Fix a listener, timer and session leak in the ACP manager: `sendPrompt` timeouts left their `stdout`/`exit` listeners attached to the `acpManager` singleton, and sessions that exited on their own were never removed from the session map.
+- **fix(acp):** release the `stdout`/`exit` listeners and the idle timer that a `sendPrompt` timeout used to leave attached to the `acpManager` singleton, and drop sessions that exited on their own from the session map instead of keeping them forever.
--- a/src/lib/acp/manager.ts
+++ b/src/lib/acp/manager.ts
@@ -90,6 +90,10 @@ export class AcpManager extends EventEmitter {
 
     child.on("exit", (code, signal) => {
       session.alive = false;
+      // Only kill() used to remove entries, so any agent that exited on its own
+      // stayed in the map forever. getActiveSessions() filters on `alive`, which
+      // hid the growth from callers.
+      this.sessions.delete(sessionId);
       this.emit("exit", { sessionId, code, signal });
     });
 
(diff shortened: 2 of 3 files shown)


anhtahaylove:

| PR | Fixes | Area |
|---|---|---|

PR #13098, opened 2026-09-09
fix(media): parallel combo fan-out and cross-process log id collisions

### LINKED ISSUE #13099: fix(resilience): image/video combo sequential execution blocks healthy providers; call_logs UNIQUE collisions
## Problem

Two independent issues cause image and video combo strategies to fail intermittently:

### 1. Sequential combo execution blocks healthy providers

Image and video combo strategies (`imageCombo.ts`, `videoCombo.ts`) run targets sequentially in a `for` loop. When the first-priority target is slow (AI Horde queue exhaustion, provider timeout, quota exhaustion), it blocks every healthy sibling behind it — even though a lower-priority target could answer in seconds.

**Root cause**: The sequential loop `await`s each target before moving to the next. A starved-but-first provider burns the entire request budget, and by the time it fails, no time remains for healthy alternatives.

**Symptom**: Combo requests return long delays or timeouts even when one provider is fast and healthy, because the slow provider is listed first in the combo priority.

**Evidence from logs**:
- AI Horde targets failing with queue wait exceeding request deadline
- 429 rate-limit responses from providers with exhausted quotas
- Requests timing out at the full budget despite healthy providers being available

### 2. call_logs UNIQUE constraint collisions across processes

When two OmniRoute processes share the same `DATA_DIR` (e.g., a dev checkout and a globally-installed CLI), auto-generated call log IDs collide because `Date.now()-counter` produces identical values across processes on the same millisecond.

**Root cause**: `generateLogId()` in `src/lib/usage/callLogs.ts` used `Date.now()-counter` — a counter that resets per process. Two processes hitting the same millisecond produce the same ID, triggering `UNIQUE constraint failed: call_logs.id`.

**Symptom**: Call log entries silently fail to persist. The INSERT throws, the error is caught, and the log entry is lost — no crash, just missing audit data.

### Parallel fan-out for combo strategies

Both `imageCombo.ts` and `videoCombo.ts` now fan out all targets concurrently using `Promise.race([firstSuccess, Promise.allSettled(tasks)])`:
- All targets run simultaneously
- First healthy success wins immediately
- Slow/failed siblings unwind in the background
- Terminal errors (400/401/403) surface when every target fails

### Process-unique call log IDs

`generateLogId()` now returns `${Date.now()}-${process.pid}-${counter}`, embedding the process PID to guarantee uniqueness across processes sharing one SQLite database.

## Testing

- **Parallel fan-out test**: Injects a slow (300ms) failing first target and a fast healthy second target. Asserts the response completes in <250ms, proving parallel execution.
- **Call log ID test**: Generates two auto-IDs, asserts they contain the process PID and are unique.
- All 16 relevant tests pass.

## Impact

- Combo requests no longer timeout when one provider is slow
- Call log audit trail is reliable in multi-process deployments
- Backward compatible — no API changes, no config changes required

opensource-elearning:
## Summary

Image and video combo strategies ran targets sequentially — a slow first target (e.g. AI Horde queue exhaustion) blocked every healthy sibling behind it. Both now fan out all targets concurrently via `Promise.race` against `Promise.allSettled`, so the first success wins immediately without waiting for slower siblings to finish.

Also fixes `call_logs` UNIQUE constraint collisions when two OmniRoute processes share the same SQLite `DATA_DIR` (dev + globally-installed CLI): the auto-generated id now embeds `process.pid` to guarantee uniqueness across processes.

Closes #13099

## Changes

- **`open-sse/services/imageCombo.ts`** — Replaced sequential for-loop with parallel `runTarget` + `Promise.race([firstSuccess, Promise.allSettled(tasks)])`. Terminal errors (400/401/403) surfaced when every target fails. Added `options?.generateImage` parameter for test injection.
- **`open-sse/services/videoCombo.ts`** — Same parallel fan-out pattern. Preserved per-target prompt-optional (I2V) and credential resolution variants.
- **`src/lib/usage/callLogs.ts`** — `generateLogId()` now returns `${Date.now()}-${process.pid}-${counter}` to avoid cross-process collisions.
- **`tests/unit/combo/image-combo.test.ts`** — New test asserting parallel fan-out completes well before a slow first target's failure window.
- **`tests/unit/call-logs-session-tag.test.ts`** — New test asserting auto-generated ids contain pid and are unique per process.

## Test results

All 16 relevant tests pass (image-combo: 10, call-logs: 6).

⚠️ base-red inherited: #12732
@diegosouzapw

### DIFF
--- a/open-sse/services/imageCombo.ts
+++ b/open-sse/services/imageCombo.ts
@@ -39,8 +39,8 @@ type ImageGenerationResult =
  *
  * 1. Resolve combo targets via resolveComboTargets.
  * 2. Filter to images-capable targets (those with an entry in the image registry).
- * 3. Iterate targets in priority order; for each target, resolve credentials and
- *    call handleImageGeneration. Return the first success or the last failure.
+ * 3. Run all images-capable targets concurrently (optional injected generator
+ *    for tests); first healthy success wins, terminal errors surface last.
  * 4. Attach combo name, selected target, and fallback count to response headers.
  */
 export async function executeImageCombo(
@@ -51,7 +51,8 @@ export async function executeImageCombo(
     policy: { apiKeyInfo?: { id?: string; name?: string } | null };
   },
   startTime: number,
-  log: typeof logger
+  log: typeof logger,
+  options?: { generateImage?: typeof handleImageGeneration }
 ): Promise<Response> {
   // 1. Resolve combo targets
   const combo = await getComboByName(comboName);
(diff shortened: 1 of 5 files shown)


diegosouzapw:
Solid work on both fronts — the pid-based id fix is a clean one-liner for a real cross-process
collision (confirmed #13099's root cause is still present on the current tip), and the parallel
fan-out change is a real UX win for slow/stuck first targets (AI Horde queue exhaustion, etc.),
verified against 10 passing tests including the race-vs-slow-failure scenario. Two small asks
before merge: (1) a changelog fragment for each of the two fixes — they're independent
behavior changes and readers will want to find them separately; (2) given how different in
scope and risk these two changes are (a one-line DB id fix vs. a rewrite of combo dispatch
concurrency), would you be open to splitting them into two PRs? Not a blocker if you'd rather
keep them together — just flagging it since it'd make a future revert of either piece cleaner.

opensource-elearning:
Superseded by #13851 (call-log pid fix) and #13852 (parallel combo fan-out), split per review feedback. Changelog fragments added in both.

PR #13100, opened 2026-09-09
fix(acp): bound session output buffers and reset stderr per prompt

anhtahaylove:
Follow-up to #13096 (same file, different resource). Refs #13095 — the buffer half I deliberately left out of that PR.

## Problem

Both per-session buffers grow on every chunk and have no upper bound:

```ts
child.stdout?.on("data", (chunk: Buffer) => {
  session.stdoutBuffer += chunk.toString();   // unbounded
});
child.stderr?.on("data", (chunk: Buffer) => {
  session.stderrBuffer += chunk.toString();   // unbounded, and never reset
});
```

Two distinct problems:

1. **No cap.** A chatty or looping agent grows a live session's footprint without limit. Measured on `release/v3.8.51`, an agent writing 4 MiB produces `stdoutBuffer` of exactly 4194304 chars — nothing clamps it.
2. **stderr is never reset.** `sendPrompt` clears `stdoutBuffer` before sending but not `stderrBuffer`, so stderr accumulates across every prompt for the lifetime of the session and the diagnostics for one prompt carry stale output from all earlier ones.

## Fix

Cap both at 1 MiB via `appendCapped()`, **keeping the tail**. That direction is deliberate: `sendPrompt` resolves with the stdout collected since the prompt was written, and stderr is read for diagnostics after a failure, so the newest output is the part callers use. A `[...output truncated...]` marker is prepended so truncation is visible rather than silent.

Also reset `stderrBuffer` alongside `stdoutBuffer` in `sendPrompt`.

1 MiB is well above a realistic agent response while keeping a stuck session bounded. Happy to make it configurable if you'd prefer a knob.

## Tests

`tests/unit/acp-manager-buffer-cap-13095.test.ts`, four cases:

1. stdout stays within the cap when an agent floods 4 MiB.
2. stderr stays within the cap under the same flood.
3. Truncation keeps the **newest** output — a `FINAL-MARKER` written last must survive.
4. stderr from an earlier prompt must not leak into the next one.

Probes spawn `process.execPath` with an inline script via a test-only custom agent — no real CLI agent binary needed.

Verified as real detectors: all four fail on `release/v3.8.51` before the change (`pass 0 / fail 4`), with the size assertions reporting `grew to 4194304 chars, above the 1048576 cap`. All four pass after.

## Scope

`fail 0` across the full ACP suite (40 tests) and `tsc -p tsconfig.json --noEmit` is clean.

Branched from `release/v3.8.51` rather than stacked on #13096 — the two touch different hunks of `manager.ts` and merge independently in either order.

### DIFF
--- a/changelog.d/fixes/13095-acp-buffer-cap.md
+++ b/changelog.d/fixes/13095-acp-buffer-cap.md
@@ -1,1 +1,1 @@
-Bound the ACP session output buffers: `stdoutBuffer` and `stderrBuffer` now cap at 1 MiB and keep the most recent output, and `stderrBuffer` is reset per prompt instead of accumulating for the lifetime of the session.
+- **fix(acp):** bound the ACP session output buffers — `stdoutBuffer` and `stderrBuffer` now cap at 1 MiB keeping the most recent output behind a visible `[...output truncated...]` marker, and `stderrBuffer` is reset per prompt instead of accumulating for the lifetime of the session.
--- a/changelog.d/fixes/13095-acp-buffer-cap.md
+++ b/changelog.d/fixes/13095-acp-buffer-cap.md
@@ -1,0 +1,1 @@
+Bound the ACP session output buffers: `stdoutBuffer` and `stderrBuffer` now cap at 1 MiB and keep the most recent output, and `stderrBuffer` is reset per prompt instead of accumulating for the lifetime of the session.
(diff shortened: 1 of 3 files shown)


anhtahaylove:

| PR | Fixes | Area |
|---|---|---|

PR #13102, opened 2026-09-09
feat(auth): add per-API-key ordered connection preference

Bl0ck154:
## Why this is useful

OmniRoute already has excellent multi-account resilience, but account preference is currently global: connection `priority` / fallback strategy applies to every client, while API-key `allowedConnections` only answers **which** connections may be used.

That leaves an important multi-project / multi-tenant use case unsolved. If two projects share the same provider account pool, operators cannot currently express:

```text
Project A: Account A -> Account B
Project B: Account B -> Account A
```

without duplicating connection objects or changing global priority for everyone.

This PR adds an API-key-scoped ordered routing hint, `preferredConnections`, so each project can have its own primary account order **and still retain OmniRoute's existing automatic failover**.

## What changes

Each API key may now store an ordered array of connection UUIDs:

```json
{
  "allowedConnections": ["ACCOUNT_A", "ACCOUNT_B"],
  "preferredConnections": ["ACCOUNT_B", "ACCOUNT_A"]
}
```

At credential selection time OmniRoute chooses the first preferred connection that has already survived the normal eligibility pipeline.

That distinction is intentional:

- `allowedConnections` remains the access-control boundary.
- `preferredConnections` only changes ranking; it never grants access.
- quota exhaustion, connection cooldown, terminal account states, model lockouts, request exclusions, and normal provider eligibility filters still win.
- a failed/cooling primary is excluded by the existing retry path, so the next preferred eligible account is selected automatically.
- when the primary becomes eligible again, subsequent requests prefer it again.
- forced combo connections and active exclusive leases remain authoritative.
- an empty preference list is fully backward-compatible and preserves the existing provider fallback strategy / global priority behavior.

## Practical benefits

This makes one shared OmniRoute deployment much easier to operate for multiple projects or tenants:

- **quota isolation** — keep Project A primarily consuming account A while Project B primarily consumes account B;
- **deterministic workload ownership** — know which subscription/account should normally carry a project's traffic;
- **better subscription utilization** — distribute steady-state traffic intentionally without giving up failover;
- **no duplicated credentials** — the same provider connection records can be reused by many API keys with different ordering;
- **graceful resilience** — existing quota/cooldown/fallback machinery continues to handle outages and exhausted accounts.

The mechanism is provider-agnostic. The motivating case is multi-account OAuth providers such as AntiGravity, but it works for any provider with multiple eligible connections.

## Implementation

- adds `api_keys.preferred_connections` as an ordered JSON array;
- exposes `preferredConnections` through API-key create/update/read metadata and sync bundles;
- threads the preference into chat/combo credential selection;
- applies preference only after access/health/quota eligibility filtering;
- preserves existing selection behavior when no preference is configured;
- documents the feature in `docs/routing/API_KEY_CONNECTION_PREFERENCE.md`.

## Example

```text
API key "project-a"
  allowed:   A, B
  preferred: A, B

API key "project-b"
  allowed:   A, B
  preferred: B, A
```

Normal traffic:

```text
project-a -> A
project-b -> B
```

If B reaches quota/cooldown:

```text
project-b -> A
```

After B recovers:

```text
project-b -> B
```

## Compatibility and safety

No existing API key changes behavior because `preferredConnections` defaults to an empty list. Existing global connection priority and fallback strategies remain the fallback behavior.

Preference is evaluated against the already-filtered candidate pool, so a preferred connection cannot bypass `allowedConnections`, cooldowns, model lockouts, terminal states, quota preflight, or retry exclusions.

## Tests

Validated on the current `release/v3.8.51` head:

- ordered preferred selection helper;
- filtered/ineligible preferred accounts cannot be resurrected;
- empty preference preserves legacy path;
- API-key DB create/read/update persistence;
- real AntiGravity `getProviderCredentials()` selection where global priority points at A but API-key preference selects B;
- real retry exclusion where preferred B fails and selection falls through to A;
- existing AntiGravity 429/model-lockout behavior remains intact;
- existing API-key scope / connection access validation remains intact;
- `npm run typecheck:core` passes;
- docs count/sync gates pass.


diegosouzapw:
Solid, well-motivated feature — the multi-tenant/quota-isolation use case is real and the
`preferredConnections` design (ranking hint, never an access grant, evaluated only after the
normal eligibility pipeline) is the right shape. Traced `orderedConnections` in `auth.ts` and
confirmed it really is the post-quota/post-lease-policy candidate pool by the time your
`pickPreferredConnection()` runs, so the core safety claim in your PR description checks out.
Two things to fix before merge: (1) `176_api_key_preferred_connections.sql` now collides with
`176_xp_action_counts.sql`, merged to the release branch after this PR was opened — please
renumber to the next free migration number and re-run `check:migration-numbering`; (2) the
migration-count bumps in `AGENTS.md`/`llm.txt`/the 41 `docs/i18n/*/llm.txt` files (171→173) are
themselves now stale against the current tip (175) — needs a fresh pass of the count generator
on rebase. Since this PR touches `AGENTS.md`/`llm.txt`, it'll also need an explicit owner
sign-off on those specific hunks even though the content itself is just the auto-generated
count.

PR #13106, opened 2026-09-09
fix(gamification): close the badge SSE stream when the signal is already aborted

### LINKED ISSUE #13103: fix(backend): gamification: badge SSE stream leaks two intervals forever when the client aborts before the stream starts
## Summary

`createBadgeNotificationStream` (`src/lib/gamification/notifications.ts`) registers its cleanup with `signal.addEventListener("abort", cleanup)`. If the signal is **already aborted** by the time `start()` runs, that listener never fires and both intervals (2s poll + 15s heartbeat) run for the lifetime of the process.

## Why this is reachable

`GET /api/gamification/notifications` awaits auth before building the stream:

```ts
const authErr = await requireManagementAuth(request);   // client can disconnect here
if (authErr) return authErr;
...
const stream = createBadgeNotificationStream(apiKeyId, request.signal);
```

A client that disconnects during the auth round-trip — page navigation, dashboard refresh, dropped connection — leaves `request.signal.aborted === true` before `start()` is ever called. Every such request permanently adds two timers.

The `safeEnqueue` catch does not save it: enqueuing into a stream nobody reads just buffers, it does not throw, so `closed` is never set and the timers keep firing and keep growing the queue.

## Repro

Mirrors the function's structure exactly (intervals scaled down):

```
A timers after create: 2
A timers after abort : 0 <- expect 0        (normal lifecycle: fine)
B timers after create: 2 (signal.aborted = true )
B timers after 1s   : 2 <- 0 means cleaned up, 2 means LEAK
B stream still enqueuing: true
```

Case A is the healthy path. Case B is an already-aborted signal: both timers survive indefinitely, and the stream keeps enqueuing into a buffer no one drains.

## Suggested fix

Check `signal.aborted` before arming the timers (or run `cleanup()` immediately when it is already true), so the already-aborted case is handled the same as the abort-later case.

Same class as #13095 — a teardown path that exists but is not reachable from every exit — but here the missed path is "the event already happened" rather than "this branch forgot to call it".

anhtahaylove:
## Problem

`createBadgeNotificationStream` arms a 2s unlock poll and a 15s heartbeat, then registers cleanup with:

```ts
if (signal) {
  signal.addEventListener("abort", cleanup);
}
```

If the signal is **already aborted** when `start()` runs, that listener never fires and both intervals run for the lifetime of the process.

## Why it is reachable in production

`GET /api/gamification/notifications` awaits auth *before* building the stream:

```ts
const authErr = await requireManagementAuth(request);   // client can disconnect here
if (authErr) return authErr;
...
const stream = createBadgeNotificationStream(apiKeyId, request.signal);
```

A client that disconnects during that round-trip — page navigation, dashboard refresh, dropped connection — arrives with `request.signal.aborted === true`. Every such request permanently adds two timers.

`safeEnqueue` does not save it: enqueuing into a stream nobody reads only buffers, it does not throw, so `closed` is never set and the queue keeps growing alongside the timers.

## Fix

Check `signal.aborted` before arming and run the same `cleanup()` — the already-aborted case is now handled identically to the abort-later case. Seven lines, no behaviour change on the healthy path.

## Tests

Three tests that count live interval handles around the call:

1. **abort-later** — the healthy lifecycle. Passes *before* this fix too; it is there so a future change that breaks the normal path is still caught.
2. **already-aborted timers** — fails on the old code with `an already-aborted signal left 2 interval(s) running for the lifetime of the process`.
3. **already-aborted stream is closed** — fails on the old code by never reporting `done` (the assertion only lands after the 15s heartbeat, which is itself the evidence the stream was still live).

On the old code: **1 pass / 2 fail**. With the fix: **3 pass**, and test 3 drops from 15s to 0.8ms because the stream now closes immediately.

Full gamification + badge suite: **23/23 pass**. `tsc --noEmit` clean.

Fixes #13103

## Note

Found while auditing the WS/SSE layer after #13095. Same class — a teardown path that exists but is not reachable from every exit — but the missed path here is "the event already happened" rather than "this branch forgot to call it". The rest of the WS/SSE layer (`liveServer.ts`, the other `text/event-stream` routes, `embedWsProxy`) checked out clean.

### DIFF
--- a/changelog.d/fixes/13103-badge-sse-aborted-signal.md
+++ b/changelog.d/fixes/13103-badge-sse-aborted-signal.md
@@ -1,0 +1,1 @@
+Fix a timer leak in the badge notification SSE stream: when the request signal was already aborted before the stream started (a client that disconnects while the route is still awaiting auth), the abort listener never fired and both intervals ran for the lifetime of the process.
--- a/changelog.d/fixes/13103-badge-sse-aborted-signal.md
+++ b/changelog.d/fixes/13103-badge-sse-aborted-signal.md
@@ -1,1 +1,1 @@
-Fix a timer leak in the badge notification SSE stream: when the request signal was already aborted before the stream started (a client that disconnects while the route is still awaiting auth), the abort listener never fired and both intervals ran for the lifetime of the process.
+- **fix(gamification):** close the badge notification SSE stream when the request signal is already aborted before the stream starts — a client that disconnects while the route is still awaiting auth used to leave both the 2s unlock poll and the 15s heartbeat running for the lifetime of the process.
--- a/src/lib/gamification/notifications.ts
+++ b/src/lib/gamification/notifications.ts
@@ -110,6 +110,13 @@ export function createBadgeNotificationStream(
         }
       };
 
+      // A client that disconnects while the route is still awaiting auth
+      // arrives here already aborted, and "abort" will never fire again --
+      // the timers above would then run for the lifetime of the process.
+      if (signal?.aborted) {
+        cleanup();
+        return;
+      }
       if (signal) {
         signal.addEventListener("abort", cleanup);
       }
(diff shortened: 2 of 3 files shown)


anhtahaylove:

| PR | Fixes | Area |
|---|---|---|

Issue #7494, opened 2026-07-16
fix(startup): Web UI throws 500 while CLI falsely reports successful startup (sql.js WASM fallback OOM loop)


WoxxyG:
### OmniRoute Version

v3.8.48 (Global CLI via pnpm -g)

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

CachyOS Linux 7.2.0-rc2-2-cachyos-rc

### Node.js Version

v26.4.0

### Provider(s) Involved

N/A

### Model(s) Involved

N/A

### Client Tool

Web Front-End

### Description

When starting OmniRoute via the CLI, the server masks a fatal background crash loop. The main thread successfully binds to the port and prints a successful startup message (`✔ OmniRoute is running!`). However, navigating to the frontend dashboard yields an immediate `500 Internal Server Error`.

When native SQLite drivers fail to load (such as `better-sqlite3`), OmniRoute gracefully catches the error in the console and falls back to the bundled `sql.js` WASM driver. However, even with a tiny database file (**~1.9MB**), this fallback immediately triggers a V8 Out of Memory (OOM) exception. This suggests either a memory leak in the driver fallback probing logic or that the bundled WASM binary was compiled without memory growth enabled.

This OOM crashes the Next.js instrumentation hooks and background sync workers (`Database closed`). Because these isolated worker threads panic silently, the exit codes never bubble up to the primary thread. The main process binds to the port and incorrectly reports `✔ OmniRoute is running!`, while any attempt to access the frontend results in a `500 Internal Server Error` because the backend is trapped in a continuous crash-and-reboot loop.

**CLI Logs:**

```omniroute
  📋 Loaded env from /home/wox/.omniroute/.env

   ____                  _ ____              _
   / __ \                (_) __ \            | |
  | |  | |_ __ ___  _ __ _| |__) |___  _   _| |_ ___
  | |  | | '_ ` _ \| '_ \ |  _  // _ \| | | | __/ _ \
  | |__| | | | | | | | | | | | \ \ (_) | |_| | ||  __/
   \____/|_| |_| |_|_| |_|_|_|  \_\___/ \__,_|\__\___|

  v3.8.48

  ⏳ Starting server...


  ✔ OmniRoute is running! (started in 8.1s)

    Dashboard:  http://localhost:20128
    API Base:   http://localhost:20128/v1

    Point your CLI tool (Cursor, Cline, Codex) to:
    http://localhost:20128/v1

[...]

PR #13109, opened 2026-09-09
fix(db): release process listeners when a node:sqlite adapter closes

### LINKED ISSUE #13108: fix(backend): db: node:sqlite adapter leaks three process listeners per closed adapter (#7494 fixed only sql.js)
## Summary

`createNodeSqliteAdapter` (`src/lib/db/adapters/nodeSqliteAdapter.ts`) registers three process-level listeners per adapter:

```ts
process.once("beforeExit", () => { adapter.close(); });
process.once("SIGINT",  () => { adapter.close(); process.exit(0); });
process.once("SIGTERM", () => { adapter.close(); process.exit(0); });
```

`adapter.close()` runs `gracefulClose()` (clears the checkpoint timer, checkpoints WAL) but never removes these three listeners. Closing an adapter therefore leaves three closures pinned on `process` for the lifetime of the run, each holding the adapter and its `DatabaseSync` handle alive.

This is the exact bug #7494 fixed for the sql.js adapter. `sqljsAdapter.ts` ends `gracefulClose()` with:

```ts
// Without this, a closed adapter's whole closure (raw sql.js Database +
// buffers) stays pinned in memory forever by these 3 process-level
// listeners, compounding the OOM every failed boot leaves behind (#7494).
process.removeListener("beforeExit", flush);
process.removeListener("SIGINT", flush);
process.removeListener("SIGTERM", flush);
```

The node:sqlite adapter — the default driver on Node when better-sqlite3 is unavailable, e.g. the packaged Electron app — never got the same treatment.

## Why this is reachable

`POST /api/db-backups/import` opens a throwaway adapter per request to validate the uploaded file:

```ts
testDb = await openDatabaseAsync(tmpPath, { readonly: true });
...
testDb.close();
```

The route closes it correctly on every branch, so this is not a missing-close bug — the close itself does not release the listeners. Every backup-import request permanently adds three.

## Repro

Mirrors both adapters' registration structure, 20 open+close cycles each:

```
baseline: { beforeExit: 0, SIGINT: 0, SIGTERM: 0 }
after 20 sql.js  open+close: { beforeExit: 0, SIGINT: 0, SIGTERM: 0 }
after 20 node:sqlite open+close: { beforeExit: 20, SIGINT: 20, SIGTERM: 20 }

  11 beforeExit listeners added to [process]
  11 SIGINT listeners added to [process]
  11 SIGTERM listeners added to [process]
```

sql.js returns to zero; node:sqlite retains one listener per signal per closed adapter and trips Node's own leak warning after the 10th.

## Suggested fix

Give the handlers a named reference and remove them in `gracefulClose()`, matching `sqljsAdapter.ts`. Note the SIGINT/SIGTERM handlers also call `process.exit(0)`, so they must stay registered while the adapter is open — only the close path should detach them.

anhtahaylove:
## Problem

`createNodeSqliteAdapter` registers three process-level handlers per adapter and never removes them:

```ts
process.once("beforeExit", () => { adapter.close(); });
process.once("SIGINT",  () => { adapter.close(); process.exit(0); });
process.once("SIGTERM", () => { adapter.close(); process.exit(0); });
```

`adapter.close()` clears the checkpoint timer and checkpoints the WAL, but the three closures stay attached to `process`, each keeping the adapter and its `DatabaseSync` handle reachable for the lifetime of the run.

#7494 fixed exactly this for the sql.js adapter — `sqljsAdapter.ts` ends `gracefulClose()` by detaching its three listeners, with a comment explaining that otherwise "a closed adapter's whole closure stays pinned in memory forever by these 3 process-level listeners". The node:sqlite adapter — the default driver on Node whenever better-sqlite3 is unavailable, notably the packaged Electron app — never got the same treatment.

## Why it is reachable

`POST /api/db-backups/import` opens a throwaway adapter per request to validate the uploaded file, and closes it correctly on every branch. The close is not the problem — the close does not release the listeners. Every import request permanently adds three.

## Fix

Hoist the handlers into named references so `gracefulClose()` can detach them, matching `sqljsAdapter.ts`.

## Tests

**Test 1** opens and closes 12 adapters and asserts the process listener count returns to its starting value. On the old code:

```
AssertionError: 12 open+close cycles retained 12 "beforeExit" listener(s) on process
```

A standalone repro of both adapters side by side shows the contrast, including Node's own warning:

```
after 20 sql.js  open+close: { beforeExit: 0,  SIGINT: 0,  SIGTERM: 0  }
after 20 node:sqlite open+close: { beforeExit: 20, SIGINT: 20, SIGTERM: 20 }

MaxListenersExceededWarning: 11 beforeExit listeners added to [process]
```

**Test 2** pins the opposite direction: an adapter that is still open must keep its handlers armed, since they are what checkpoints the WAL on Ctrl-C. It passes both before and after, so a fix that detached eagerly would be caught rather than silently accepted.

Result: **1 fail / 1 pass** before, **2 pass** after. `tsc --noEmit` clean.

## Note on the suite

`tests/unit/cli-sqlite-construction-fallback-8826.test.ts` fails in my environment with `ERR_INVALID_MODULE_SPECIFIER` on a `\C:\...` fixture path. It fails identically on a clean `release/v3.8.51` checkout with my change stashed, so it is a pre-existing Windows path issue in that test, unrelated to this PR. The rest of the sqlite/adapter/driver suites: 58 pass.

Fixes #13108

### DIFF
--- a/changelog.d/fixes/13108-nodesqlite-process-listener-leak.md
+++ b/changelog.d/fixes/13108-nodesqlite-process-listener-leak.md
@@ -1,0 +1,1 @@
+Fix a process-listener leak in the node:sqlite adapter: closing an adapter left its `beforeExit`/`SIGINT`/`SIGTERM` handlers attached to `process`, pinning the adapter and its database handle for the lifetime of the run. #7494 fixed the same bug for the sql.js adapter only.
--- a/changelog.d/fixes/13108-nodesqlite-process-listener-leak.md
+++ b/changelog.d/fixes/13108-nodesqlite-process-listener-leak.md
@@ -1,1 +1,1 @@
-Fix a process-listener leak in the node:sqlite adapter: closing an adapter left its `beforeExit`/`SIGINT`/`SIGTERM` handlers attached to `process`, pinning the adapter and its database handle for the lifetime of the run. #7494 fixed the same bug for the sql.js adapter only.
+- **fix(db):** release the `beforeExit`/`SIGINT`/`SIGTERM` handlers when a `node:sqlite` adapter closes, so a closed adapter and its database handle are no longer pinned to `process` for the lifetime of the run — the same treatment #7494 gave the sql.js adapter.
--- a/src/lib/db/adapters/nodeSqliteAdapter.ts
+++ b/src/lib/db/adapters/nodeSqliteAdapter.ts
@@ -35,26 +35,34 @@ export async function createNodeSqliteAdapter(filePath: string): Promise<SqliteA
   }, CHECKPOINT_INTERVAL_MS);
   (checkpointTimer as unknown as NodeJS.Timeout).unref?.();
 
+  // Declared before gracefulClose so the close path can detach them. Without
+  // this, every closed adapter leaves three closures pinned on `process` --
+  // each holding this adapter and its DatabaseSync handle alive -- and short-
+  // lived adapters (POST /api/db-backups/import opens one per request) trip
+  // Node's MaxListenersExceededWarning. #7494 fixed exactly this for sql.js.
+  const onBeforeExit = () => {
+    adapter.close();
+  };
+  const onSignal = () => {
+    adapter.close();
+    process.exit(0);
+  };
+
   function gracefulClose() {
     clearInterval(checkpointTimer as unknown as NodeJS.Timeout);
     try {
       db.exec("PRAGMA wal_checkpoint(TRUNCATE)");
     } catch {}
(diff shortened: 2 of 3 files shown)


anhtahaylove:
Follow-up on the same leak audit as #13097 (merged, deployed clean). This one is the node:sqlite counterpart of the sql.js fix in #7494 — see the consolidated status of all eight remaining PRs in https://github.com/diegosouzapw/OmniRoute/pull/13114#issuecomment-latest. Waiting on first-time-contributor workflow approval; happy to rebase or split if useful.

PR #13114, opened 2026-09-09
fix(cli-helper): clear the log stream timeout on the abort path

### LINKED ISSUE #13113: fix(backend): cli-helper: createLogStream leaks its timeout timer on the abort path
## Summary

`createLogStream` (`src/lib/cli-helper/log-streamer.ts`) arms a timeout timer per stream, then leaks it on the abort path — which is the normal way this stream ends.

```ts
const timeoutId = setTimeout(() => {
  if (follow) return;
  controller.error(new Error(`Log stream timed out after ${timeout}ms`));
}, timeout);

try {
  const response = await fetch(url, { signal, headers });
  ...
  const reader = response.body.getReader();
  while (true) {
    const { done, value } = await reader.read();
    if (done) break;
    if (signal.aborted) break;      // (2) reader never cancelled
    controller.enqueue(value);
  }
  controller.close();
  clearTimeout(timeoutId);
} catch (err) {
  if (signal.aborted) return;       // (1) clearTimeout skipped
  controller.error(...);
  clearTimeout(timeoutId);
}
```

**(1) The timer leaks on stop.** `stop()` calls `controller.abort()`, the in-flight `fetch`/`read` rejects with `AbortError`, and the catch returns at `if (signal.aborted) return;` — *before* `clearTimeout`. Every stopped stream leaves an armed timer holding the stream-controller closure until it fires (default 30s, caller-configurable and unbounded).

Every other exit path clears the timer. The abort path — the one `stop()` is built to trigger — is the single path that does not.

**(2) The reader is never released.** `break` on `signal.aborted` exits the loop without `reader.cancel()`; the body stays locked and the undici socket is held until GC.

## Repro

10 streams started then stopped:

```
active Timeout handles before=0 after 10 stopped streams=10
LEAK: 10 armed 30s timer(s) survive stop(); each pins the stream controller closure
```

Measured with `process.getActiveResourcesInfo()`, against a local server that keeps the response open so abort lands mid-stream.

## Impact

`follow: true` streams are long-lived and stopped explicitly, so this is the expected lifecycle rather than an edge case. A dashboard that restarts the log stream on filter changes arms one surviving timer per restart.

## Fix

Clear the timer in `finally` so no exit path can skip it, and cancel the reader when leaving the loop early.

anhtahaylove:
## Problem

`createLogStream` (`src/lib/cli-helper/log-streamer.ts`) arms a timeout timer per stream and clears it on every exit path except one — the abort path, which is how `stop()` ends a stream.

```ts
const timeoutId = setTimeout(() => { ... }, timeout);

try {
  const response = await fetch(url, { signal, headers });
  ...
  while (true) {
    const { done, value } = await reader.read();
    if (done) break;
    if (signal.aborted) break;      // (2) reader never cancelled
    controller.enqueue(value);
  }
  controller.close();
  clearTimeout(timeoutId);
} catch (err) {
  if (signal.aborted) return;       // (1) clearTimeout skipped
  controller.error(...);
  clearTimeout(timeoutId);
}
```

**(1)** `stop()` calls `controller.abort()`, the in-flight `fetch`/`read` rejects with an AbortError, and the catch returns at `if (signal.aborted) return;` — before `clearTimeout`. Each stopped stream leaves an armed timer holding the stream-controller closure until it fires (30s default, caller-configurable and unbounded).

**(2)** `break` on `signal.aborted` exits the loop without `reader.cancel()`, so the body stays locked and its undici socket is held until GC.

## Why it matters

`follow: true` streams are long-lived and stopped explicitly — this is the normal lifecycle, not an edge case. A dashboard that restarts the log stream on filter changes arms one surviving timer per restart.

## Repro

8 follow-mode streams started against a server that keeps the response open, then stopped:

```
AssertionError: stopping 8 streams retained 8 armed timer(s) (before=0 after=8); stop() must clear the timeout
```

Measured with `process.getActiveResourcesInfo()`.

## Fix

Move `clearTimeout` into a `finally` so no exit path can skip it, and cancel the reader when the read loop exits early.

## Tests

- **Test 1** — the repro above: 8 streams stopped, armed-timer count must return to baseline. Fails before, passes after.
- **Test 2** — a stream that ends normally must also leave no armed timer. Passes both before and after, so a fix that cleared the timer too eagerly (breaking the real timeout) would be caught rather than silently accepted.

Result: **1 fail / 1 pass** before, **2 pass** after. `tsc -p tsconfig.json --noEmit` clean.

## Note on the surrounding suite

`tests/unit/cli-helper/*` shows 57 pass / 10 fail in my environment (config-generator, getHermesHome, tool-detector). Identical 57/10 on a clean `release/v3.8.51` checkout with my change stashed — pre-existing and unrelated to this PR.

Fixes #13113

### DIFF
--- a/changelog.d/fixes/13113-logstream-timer-leak.md
+++ b/changelog.d/fixes/13113-logstream-timer-leak.md
@@ -1,1 +1,1 @@
-Fix a timer leak in `createLogStream`: `stop()` aborts the in-flight fetch and returns through the `signal.aborted` branch, which skipped `clearTimeout`, leaving an armed timer per stopped stream. The stream reader is now also cancelled when the read loop exits early.
+- **fix(cli-helper):** clear the `createLogStream` timeout on the abort path — `stop()` aborts the in-flight fetch and returned through the `signal.aborted` branch, which skipped `clearTimeout` and left an armed timer per stopped stream. The stream reader is now also cancelled when the read loop exits early.
--- a/changelog.d/fixes/13113-logstream-timer-leak.md
+++ b/changelog.d/fixes/13113-logstream-timer-leak.md
@@ -1,0 +1,1 @@
+Fix a timer leak in `createLogStream`: `stop()` aborts the in-flight fetch and returns through the `signal.aborted` branch, which skipped `clearTimeout`, leaving an armed timer per stopped stream. The stream reader is now also cancelled when the read loop exits early.
--- a/src/lib/cli-helper/log-streamer.ts
+++ b/src/lib/cli-helper/log-streamer.ts
@@ -38,30 +38,37 @@ export function createLogStream(options: LogStreamOptions = {}): LogStream {
 
         if (!response.ok) {
           controller.error(new Error(`HTTP ${response.status}: ${response.statusText}`));
-          clearTimeout(timeoutId);
           return;
         }
 
         if (!response.body) {
           controller.error(new Error("Response body is null"));
-          clearTimeout(timeoutId);
           return;
         }
 
         const reader = response.body.getReader();
 
-        while (true) {
-          const { done, value } = await reader.read();
-          if (done) break;
-          if (signal.aborted) break;
-          controller.enqueue(value);
+        try {
+          while (true) {
+            const { done, value } = await reader.read();
+            if (done) break;
+            if (signal.aborted) break;
+            controller.enqueue(value);
+          }
+        } finally {
+          // Leaving the loop early (abort/throw) otherwise keeps the body locked
+          // and its socket held until GC.
+          await reader.cancel().catch(() => {});
         }
 
         controller.close();
-        clearTimeout(timeoutId);
       } catch (err) {
         if (signal.aborted) return; // Expected stop
         controller.error(err instanceof Error ? err : new Error(String(err)));
+      } finally {
+        // `stop()` aborts mid-fetch and returns through the `signal.aborted`
+        // branch above, so clearing the timer on the individual exit paths
+        // misses the one path stop() is built to take.
         clearTimeout(timeoutId);
       }
     },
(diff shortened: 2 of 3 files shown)


anhtahaylove:
Thanks for merging #13097 — it deployed clean on `release/v3.8.51` (CodeQL, Docker, VPS deploy all green).

The remaining eight PRs from the same leak audit are still waiting on first-time-contributor workflow approval. They are independent of each other — different files, different failure modes — so they can be reviewed or merged in any order, or individually:

| PR | Fix | Evidence |
|---|---|---|
| #13091 | compression pool leaves evicted workers running | test |
| #13092 | SIGKILL escalation not idempotent, listeners stack | test |
| #13093 | llmlingua worker spawned with a bad path | test |
| #13096 | ACP: listeners, timers and sessions not released | 40/40 |
| #13100 | ACP output buffers unbounded, stderr never reset | 4/4 + 40/40 |
| #13106 | badge SSE stream not closed when the signal is already aborted | 3/3 + 23/23 |
| #13109 | node:sqlite adapter retains 3 process listeners per close | 20-cycle repro |
| #13114 | log stream timeout timer survives `stop()` | 8-stream repro |

Every one has a failing-before/passing-after test, and a second test pinning the opposite direction so an over-eager cleanup would be caught rather than silently accepted. `tsc -p tsconfig.json --noEmit` is clean on each.

Two notes so nothing is a surprise on your side:

- #13109 fixes the same bug #7494 already fixed for the sql.js adapter — `sqljsAdapter.ts` detaches its process listeners, `nodeSqliteAdapter.ts` never got the same treatment. It is the default driver on Node when better-sqlite3 is unavailable.
- Two pre-existing suite failures show up in my environment and are documented in the relevant PR bodies: `cli-sqlite-construction-fallback-8826` (a `\C:\` path issue on Windows) and `tests/unit/cli-helper/*` at 57 pass / 10 fail. Both reproduce identically on a clean `release/v3.8.51` checkout with my changes stashed, so neither is caused by these PRs.

Happy to rebase, split, or close any of them if the batch is too large to take at once.

PR #13115, opened 2026-09-09
fix(logs): recover concatenated JSON objects in the Provider Event Stream viewer

hartmark:
## What Problem This Solves

The "Provider Event Stream" viewer on a call log's detail page sometimes
showed raw, unreadable text instead of the expected collapsible JSON tree
-- but only for **Gemini**, and only intermittently.

## Root Cause

Gemini's own stream-chunk capture concatenates multiple complete JSON
objects back-to-back with **no separator between them** (`}{`), unlike the
SSE blank-line-per-event framing every other provider's capture uses here.
A single `data:` payload can therefore hold several complete top-level
JSON values, not one -- `JSON.parse` correctly rejects the whole thing:

```
Unexpected non-whitespace character after JSON at position 352
```

...and the *entire* payload fell back to plain text, even though every
individual object inside it was perfectly well-formed.

I confirmed this by replaying real stored `streamChunks.provider`
artifacts from production call logs: **161 of 663** recent streaming logs
hit this, all `gemini`, some with as many as **11** concatenated objects
in a single capture.

## Fix

When a `data:` payload fails to parse as one JSON value,
`splitConcatenatedJsonValues()` scans bracket/string depth (properly
handling escaped quotes) to recover each complete top-level JSON value in
sequence. Each recovered value renders as its own collapsible node instead
of the whole thing collapsing to unreadable raw text.

Genuinely incomplete or non-JSON content (a real mid-stream truncation,
SSE comments, keep-alives) still falls back to plain text exactly as
before -- the splitter only accepts a payload if it can fully account for
every byte as complete top-level JSON values, and returns `null` (no
change in behavior) otherwise.

## Evidence

```
$ node --test tests/unit/request-logger-stream-concatenated-json.test.ts
tests 8, pass 8, fail 0

$ node --test tests/unit/request-logger-detail-copy-all.test.ts
tests 4, pass 4, fail 0
```

- `eslint`: clean

**Live-verified**: replayed 8 real production call-log artifacts (the
original incident plus 7 more found via a live scan) through the fixed
parser -- all now recover 100% as JSON segments with zero text fallback,
versus partial/total text fallback before.

### DIFF
--- a/src/i18n/messages/en.json
+++ b/src/i18n/messages/en.json
@@ -11341,6 +11341,8 @@
       "copy": "Copy",
       "autoscrollOn": "Autoscroll: on",
       "autoscrollOff": "Autoscroll: off",
+      "rawViewOn": "Raw view: on",
+      "rawViewOff": "Raw view: off",
       "collapseAllLevels": "Collapse all",
       "collapseOneLevel": "Collapse one level",
       "currentExpandLevel": "Current expand level",
--- a/src/shared/components/RequestLoggerDetail.tsx
+++ b/src/shared/components/RequestLoggerDetail.tsx
@@ -178,6 +178,17 @@ function StreamSection({ title, sectionId, json, onCopy }) {
       return true;
     }
   });
+  // Default to the rendered/parsed view (segments below) -- raw is an
+  // explicit opt-in for inspecting exactly what was captured on the wire,
+  // timestamp markers and all, when the rendered tree hides something the
+  // parser got wrong (e.g. a chunk-boundary split like this PR fixes).
+  const [showRaw, setShowRaw] = useState(() => {
+    try {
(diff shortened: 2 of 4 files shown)


diegosouzapw:
Excellent debugging trail — replaying real production stream artifacts to quantify the impact
(161/663 recent Gemini logs) before writing the fix is exactly the kind of evidence that makes
this easy to trust. `splitConcatenatedJsonValues()` is clean and conservative (falls back to
the old behavior — `null` — for anything it can't fully account for byte-for-byte), and the
raw-view toggle is a nice small addition for debugging future parser gaps. Ran your full test
suite in isolation: 14/14 pass, including the real 10-chunk production artifact case. This
looks merge-ready.

PR #13116, opened 2026-09-09
perf(gamification): add xp_action_counts cache table to fix O(n) COUNT(*) on every XP event (#12546)

### LINKED ISSUE #12546: fix(gamification): action_count badges are computed over a 30-day-pruned xp_audit_log

KooshaPari:
## Summary

Fixes #12546. Resubmission of #12712 (which was closed for base drift + `user_id` vs `api_key_id` column mismatch).

The XP-earning hot path was running `SELECT COUNT(*) FROM xp_audit_log ... GROUP BY ...` on every event. With large `xp_audit_log` tables this is the slow path that triggered base-red CI badges.

## Fix

Introduces an incremental counter cache keyed on `api_key_id` (matching the `xp_audit_log` schema):

- **New table** `xp_action_counts(api_key_id, action, count, updated_at)` via **migration 174** (next free after 173 on `release/v3.8.51`)
- **Backfill** via `GROUP BY api_key_id, action` from existing `xp_audit_log` rows
- **`incrementActionCount()`** runs in the **same transaction** as the `xp_audit_log` INSERT (atomic) — and is **wrapped in `try/catch`** so a cache write failure doesn't fail the whole XP grant (cache is a performance optimization, not the source of truth)
- **`getActionCountByType()`** reads from the cache, falls back to `COUNT(*)` on cache miss (user predates migration 174), also wrapped in `try/catch` so a cache read failure falls back rather than throws
- **`getActionCount()`** in `badges.ts` is now a 3-line wrapper calling the cached reader
- **`checkActionCountBadges()`** in `events.ts` uses the cached reader — O(1) hot path

## Behavior

| Scenario | Before | After |
|---|---|---|
| Hot path (per XP event) | INSERT + full-table `COUNT(*)` | INSERT + UPSERT (cache only) |
| Bad-count badge threshold check | O(audit_log) | O(1) |
| User predates migration 174 | n/a | fallback to `COUNT(*)` — same as before |
| Cache write fails | n/a | XP grant still succeeds, log + proceed |
| Cache read fails | n/a | fallback to `COUNT(*)` — no regression |

## Files

| File | Status | LOC |
|---|---|---|
| `src/lib/db/migrations/174_xp_action_counts.sql` | new | ~25 |
| `src/lib/db/gamification.ts` | modified | +45 (increment + getter + transaction wrap) |
| `src/lib/gamification/badges.ts` | modified | -15, +3 (wrapper) |
| `src/lib/gamification/events.ts` | modified | -3, +5 (use cached reader) |
| `tests/unit/gamification/db-gamification.test.ts` | modified | +3 tests |

## Tests

`tests/unit/gamification/db-gamification.test.ts`:
- `addXp increments the cache`
- `incrementActionCount is idempotent (same api_key+action)`
- `getActionCountByType falls back to COUNT(*) on cache miss`

All 6 tests pass.

## Why this matters

The XP-earning path is hit on every successful LLM request — currently O(audit_log) per call. With migration 174 in place it's O(1), which:
- Unblocks the base-red CI badge on `release/v3.8.51`
- Makes the bad-count thresholds (combo, provider_switch, daily_login) consistent regardless of how big the user's audit_log grows
- Avoids the cascade where large audit_log → slow count → slow badges → slow UI

Fixes #12546.


HouMinXi:
Thanks for the `xp_action_counts` cache.

Folded into a local recut of `release/v3.8.51`. The merge itself was clean. After stacking, the recut already had tip's `174_server_tool_executions.sql` (#12867, merged) and `176_provider_connection_synced_models_at.sql` (#13248), so this PR's `174_xp_action_counts.sql` tripped the migration-runner collision guard.

Resolution on the recut only: rename to `177_xp_action_counts.sql`. This PR is still MERGEABLE vs tip and does not need a rewrite; once tip moves past 174, a rebase that picks the next free number will land cleanly.

Purpose of the recut: ship current `release/v3.8.51` plus open product PRs on a private production host.

KooshaPari:
Shipped the rename directly in PR `pr/resubmit-12712-xp-cache` (latest commit on the same branch) instead of waiting for tip to clear 174:

- `src/lib/db/migrations/174_xp_action_counts.sql` -> `src/lib/db/migrations/177_xp_action_counts.sql`
- Test description updated from `migration 174` to `migration 177` (only other reference — migration numbers are file-name-based, no central registry)

Verification:
```
node --test tests/unit/gamification/db-gamification.test.ts
  6/6 pass (cache increment + COUNT(*) fallback + unknown key)
```

The branch is now mergeable vs tip without any rebase / re-cut on your end.

Re: your recut branch — happy to mirror the same rename there if you'd prefer to land the rename via your recut rather than this PR; just let me know which path is cleaner for the merge.

diegosouzapw:
Thanks for the O(1) action-count cache. The same fix already landed in the release branch through #12651: a durable `xp_action_counts` table incremented in `addXp()`, read by both `getActionCount()` and `checkActionCountBadges()`, with a backfill migration.

#12651 also keeps the `SUM(json_extract(metadata, '$.amount'))` semantics the badge helper relied on, which a plain row count would change. Closing as covered.

Issue #8620, opened 2026-07-25
feat(backend): allow external database (Postgres) for stateless horizontal scaling


phuchptty:
### Problem / Use Case

I'm trying to make the router high availablity. But the problem came from the database. Currently, it's an embeded SQLite db.

### Proposed Solution

- Allow choose database type at initial. Maybe add an env.
- Allow config external database through connection string in env.
- Support more type of database such as Postgres.
- Support serverless DB like cloudflare R1 or Neon

### Acceptance Criteria

- I can setup external DB
- Application become stateless -> can scale to many instances

### Area

Docker / Deployment

diegosouzapw:
Hi @phuchptty! Thanks for this thoughtful feature request! 🙏

We really appreciate the detailed proposal. We've **cataloged your idea** and it's now part of our improvement backlog.

Due to the **significant architectural impact** of this feature, we'll need to conduct thorough use-case studies and architectural analysis before we start development. This ensures we build it right and don't introduce regressions.

**What happens next:**

- Your idea is saved in our internal feature backlog
- We'll conduct architecture studies when this area is prioritized (likely 3.9 LTS or 4.0 modular cycle)

**Short-term alternative:** For high availability today, you can run multiple OmniRoute instances behind a TCP load balancer. The data layer is rarely the bottleneck for an LLM proxy — combo routing handles most failures at the LLM layer.

If you want to track progress, please **subscribe to the repository releases** — every implemented feature is announced in the CHANGELOG.

Thank you for contributing to OmniRoute's roadmap! 🚀

Issue #5409, opened 2026-06-29
feat: Postgres backend option (drop-in alternative to SQLite)


chirag127:
## Feature request

OmniRoute currently uses SQLite (`~/.omniroute/storage.sqlite`) as its only storage backend. For production deployments and multi-user/multi-instance setups, a Postgres backend would unlock:

- **Multi-tenant isolation** — per-team cost tracking without running separate OmniRoute instances
- **HA deployments** — multiple OmniRoute processes sharing state without SQLite file-locking
- **Standard backup tooling** — `pg_dump`, managed cloud Postgres (Neon, Supabase, RDS, Turso)
- **Avoids EBUSY file-lock issues** on Windows (see #5406)

## Proposed config

In `~/.omniroute/.env` or `omniroute.config.json`:

```env
OMNIROUTE_DB_BACKEND=postgres
OMNIROUTE_DB_URL=postgres://user:pass@localhost:5432/omniroute
```

Default stays `sqlite`. When `postgres` is set, use `pg` or `drizzle-orm` with a Postgres dialect — same schema, just a different driver.

## Prior art

- **LiteLLM**: `DATABASE_URL=postgresql://...` in `.env`; all state switches to Postgres with zero code changes in the HTTP layer

## Acceptance criteria

- [ ] `OMNIROUTE_DB_BACKEND=postgres` accepted
- [ ] All existing tables (routing history, cost ledger, provider configs, combo definitions) migrated to Postgres schema
- [ ] `omniroute` CLI warns clearly if backend=postgres and `DATABASE_URL` is not set
- [ ] SQLite remains the default (no breaking change)

chirag127:
Consolidating into #5161.

diegosouzapw:
Hi @chirag127! Thanks for the proposal — and for folding it into #5161. 🙏

We see the value (HA, multi-tenant, managed-Postgres ops). Being fully transparent about the effort, though: this isn't a drop-in driver swap today. OmniRoute's data layer is deeply SQLite-coupled —

- all three adapters (`betterSqliteAdapter`, `nodeSqliteAdapter`, `sqljsAdapter`) sit behind a **synchronous** SQLite-style API,
- ~90+ domain modules issue **raw SQLite SQL** through `getDbInstance()` with no dialect abstraction,
- WAL / PRAGMA / auto-vacuum tuning is SQLite-specific.

[...]

PR #13118, opened 2026-09-09
[defer] feat(db): opt-in PostgreSQL backend for multi-replica deployments

dpozimski:
## Summary

Adds an opt-in PostgreSQL storage backend so several OmniRoute replicas can share one database. Set `OMNIROUTE_DATABASE_URL=postgres://...` and every table (provider connections, API keys, combos, quota state, call logs, everything else) lives in PostgreSQL instead of a per-process `storage.sqlite`. Unset, nothing changes: SQLite stays the zero-config default and the `pg` package is never loaded.

Why now: one SQLite file means one replica, and one replica cannot carry ~50 developers plus an autonomous agent fleet. The repository-seam work tracked in #8075 is the right long-term shape. This is the shape that ships today without touching the 1,300+ synchronous `prepare().get()/all()/run()` call sites, so it is labelled experimental and kept behind the env var.

How it works:

- `src/lib/db/adapters/postgresAdapter.ts` implements the existing synchronous `SqliteAdapter` contract. Queries run in a worker thread that owns one `pg` connection (`postgres/postgresWorker.mjs`); the main thread blocks on `Atomics.wait` per statement, the same way a native SQLite call blocks. No domain module changes.
- `postgres/sqlTranslator.ts` (a keyword dispatcher over `sqlExpressions.ts`, `sqlDdl.ts`, `sqlDml.ts`, `sqlReferences.ts` and `sqlFunctionHandlers.ts`) rewrites the SQLite dialect at prepare time: placeholders, `INSERT OR REPLACE` (upsert on the primary key), `INSERT OR IGNORE`, `datetime()`/`strftime()`/`unixepoch()`, `json_*`, `typeof()`, `CAST`, `COLLATE NOCASE`, case-insensitive `LIKE`, `IS NOT <value>`, `GLOB`, bare integer columns in boolean context, mixed `CASE` branches, double-quoted string literals, `rowid`, triggers (rewritten to PL/pgSQL, `postgres/triggerTranslator.ts`), `PRAGMA table_info` and `sqlite_master` (a view plus helper functions installed once per schema by `postgres/bootstrapSql.ts`).
- All 170 applicable migrations under `src/lib/db/migrations/` apply unchanged (022/023 FTS5 are deferred as on any driver without FTS5). Init and migrations take advisory locks keyed by schema, so replicas can boot together.
- `OMNIROUTE_DATABASE_IMPORT_SQLITE=1` copies an existing `DATA_DIR/storage.sqlite` into an empty database on first boot; `npm run db:migrate-to-postgres -- --sqlite <file> --url <pg>` does the same explicitly with a per-table row-count report and a non-zero exit on shortfall. Columns that OmniRoute adds lazily at runtime are created on the target before copying.
- `docker compose --profile postgres` adds PostgreSQL 17 locally; `scripts/build/assembleStandalone.mjs` ships the worker and the `pg` dependency closure with the standalone bundle.

One unrelated-looking change: `getSyncedCapability()` in `src/lib/modelsDevSync.ts` now warms the full capabilities cache on its cold path instead of issuing one query per model and alias. The cold `/v1/models` build made ~8,800 single-row lookups, which SQLite absorbs but PostgreSQL turns into a 10 s timeout. After the change the same build makes 2 statements against that table; it is faster on SQLite too.

Not on PostgreSQL (documented in `docs/ops/DATABASE_GUIDE.md`): memory FTS5/sqlite-vec search, file backups/restore/VACUUM (use `pg_dump`).

## Related Issues

Refs #8075 (design thread), #8620, #5409, #12400 (`replicas: 1` because of the single writer).

⚠️ base-red inherited: #12732

## Validation

- Change type: feature, schema-affecting (new backend, no new migrations), public config (new env vars, compose profile).
- Focused tests: `node --import tsx/esm --test tests/unit/db-adapters/sqliteToPostgres.test.ts` (35 pass), `OMNIROUTE_TEST_DATABASE_URL=postgres://... node --import tsx/esm --test tests/unit/db-adapters/postgresAdapter.test.ts` (14 pass against PostgreSQL 17.11, one new), `tests/unit/db-core.test.ts` (40 pass, SQLite boot path after the shared-pipeline extraction), `tests/unit/settings-ui-layout-static.test.ts` (8 pass), `tests/unit/db-migrationrunner-constants-split.test.ts` + `tests/unit/db-migrations-fts5-optional-set.test.ts` (8 pass).
- Existing suites on PostgreSQL: `OMNIROUTE_DATABASE_URL=... OMNIROUTE_DATABASE_SCHEMA=auto node --import tsx/esm --test --test-concurrency=1 "tests/unit/db/**/*.test.ts"` gives 133 of 138 passing; the five left are the four `omp.test.ts` cases (fail identically on SQLite in this checkout) and one assertion on SQLite's `USING INDEX` planner wording (PostgreSQL uses the index, the text differs). Run the files serially: with node's default per-file parallelism each file bootstraps 129 tables in its own schema at once and a stock PostgreSQL (`max_locks_per_transaction=64`) answers `out of shared memory`.
- Capability tests on SQLite after the `modelsDevSync` change: 93 pass (`tests/unit/model-capabilities-*.test.ts`, `capability-filter`, `combo-projection-capabilities-3979`, `8488-capability-filter-fail-closed`, `catalog-pricing-*`, `11947-auto-combo-modalities`, `8332-combo-vision-fallback`).
- Lint: `npx eslint` clean on every changed file; `npx prettier --check` clean; `npm run typecheck:core` clean; no dashboard typecheck errors in the storage components.
- Gates this PR owns, all green locally: `check:deps` (pg and @types/pg allowlisted with justification), `check:file-size -- --base-ref` (every new file under the 1200 cap; `SystemStorageTab.tsx` back under its 1597 freeze; `core.ts` +45 recorded in `file-size-baseline.json` with the reason), `check:complexity-ratchets -- --base-ref` (complexityNewCode=0, cognitiveComplexityNewCode=0 across the 33 touched files), knip (no new unused exports), `check:mutation-test-coverage`, and the translator/trigger import cycle is gone from `check:cycles`. The env/doc contract and the Vietnamese locale no longer list anything from this PR.
- Based on current `release/v3.8.51` head (`19b9d05e0`).
- Tests included: yes (two new files).

End to end on a laptop, PostgreSQL 17 in Docker, webpack dev server, a Cursor user API key as the provider:

- Fresh database: 170 migrations in about 2.4 s, 126 tables, 25 triggers.
- Dashboard login, `POST /api/providers` (`cursor-api`, key validated upstream), `POST /api/keys`, then `POST /v1/chat/completions` with `cursor-api/auto` returned `pong` in 18 s; the call log row landed in PostgreSQL.
- A second instance on another port and `DATA_DIR`, same URL and schema, booted in 5 s with no migrations to run, authenticated the key created by the first instance, listed its connection, and answered `pong from replica B`; both instances' call logs are in the same table.
- One thing the two-replica run taught: replicas must share `STORAGE_ENCRYPTION_KEY`, `API_KEY_SECRET` and `JWT_SECRET`. With a per-`DATA_DIR` auto-generated key the second replica reads the first one's provider credentials as ciphertext (the symptom was "Cursor access token is required"). Documented.
- `npm run db:migrate-to-postgres` on the SQLite instance's real `storage.sqlite`: 124 tables reconciled, api_keys 2/2, provider_connections 1/1, call_logs 1/1, key_value 54/54, 675 ms; a second run is a no-op.
- Cold `/v1/models` after the `modelsDevSync` change: 200 in 5.4 s on PostgreSQL (SQLite: 2.5 to 3.4 s), then 41 ms from cache.

Re-run after the rebase (fresh `omniroute_e2e` database, PostgreSQL 17.11 in Docker, webpack dev server, `OMNIROUTE_DATABASE_URL` set, Node 24.14.1):

- First boot applied 172 migrations and deferred the four FTS5 ones (022, 023, 178, 180); `/api/health/ping` 200, `POST /api/auth/login` 200, `GET /api/storage/health` returned `driver: postgres`, schema `public`, server 17.11, 129 tables.
- `POST /api/keys` 201; the row is in PostgreSQL (`select name from api_keys`); `GET /v1/models` with that key 200 (521 models, 4.4 s cold) and 401 with a bogus key. Startup auto-cleanup logged `0 deleted, 0 errors`.
- A second process with its own `DATA_DIR` on another port, same URL and schema, booted with 0 migrations to run, authenticated the key the first process created, listed it under `/api/keys`, and answered `/v1/models` 200 / 401 the same way.
- `/login`, `/dashboard` and `/dashboard/settings` render (200) on this base; the `node:fs` client-bundle failure noted in the previous run was #13264 and is fixed on the tip.

Not verified here: the standalone production bundle. `next build` with Turbopack panicked on this Windows machine (`there must be a path to a root`, unrelated module-graph error also seen without these changes) and the webpack build was killed for memory twice; CI builds it.

## Rebase of 2026-09-16

Rebased the ten commits onto `release/v3.8.51` at `19b9d05e0` (150 commits ahead of the previous base). Nine locale files and `file-size-baseline.json` conflicted: upstream translated the settings strings this PR inserts its seven `settings.postgres*` keys among, so the keys were re-inserted at the same anchor on top of the translated files, and `core.ts` was re-frozen at the merged size (1770 on the tip, 1821 here; the +51 is the same dispatch and PostgreSQL-only boot mechanics as before, the base grew under #13404 and #13441).

Booting on PostgreSQL then failed twice on the new base, both fixed in the last commit with a test each:

- The migration runner recomputes the durable-snapshot requirement inside the writer transaction (7ae8bf4e05). That second computation did not carry the `postgres` exclusion the first one had, so any existing PostgreSQL schema with a pending migration was refused with "Refusing to migrate an existing database without a durable snapshot", which is a SQLite hard-link concept. Eligibility is now computed once and shared by the three checks. `Reboot_WithPendingMigrationOnExistingSchema_NeedsNoSqliteSnapshot` boots twice on one schema with a ledger row removed in between; it fails on the previous runner and passes now.
- `178_memory_fts_au_conditional` (#13405) recreates the `memory_fts_au` trigger but is missing from `OPTIONAL_FTS5_MIGRATION_VERSIONS`; only `180` (#13331), which supersedes it, was added. On a driver without FTS5 the boot stopped at 178 with "no such table: memory_fts". This also affects a SQLite build without FTS5, so it may deserve its own small PR against the release branch; it is here because the PostgreSQL boot needs it. `OptionalFts5Set_CoversEveryMigrationThatTouchesMemoryFts` scans `src/lib/db/migrations/` so the next `memory_fts` migration cannot be forgotten.

## Tests Added Or Updated

- `tests/unit/db-adapters/sqliteToPostgres.test.ts` (new): translator coverage for placeholders, DDL type mapping, identity columns, rowid tables, REPLACE/IGNORE, helper functions, CAST, operators, COLLATE, boolean context, CASE normalisation, mixed-case aliases, `json_each`, ALTER/DROP forms, virtual tables, pragmas, transactions, triggers, statement splitting, tokenizer round trips.
- `tests/unit/db-adapters/postgresAdapter.test.ts` (new, skipped unless `OMNIROUTE_TEST_DATABASE_URL` is set): live adapter behaviour, the SQLite import round trip, and a two-boot run on one schema with a pending migration.
- `tests/unit/db-migrations-fts5-optional-set.test.ts` (new) and `tests/unit/db-migrationrunner-constants-split.test.ts` (updated): the optional FTS5 set covers every migration that touches `memory_fts`.

## Coverage Notes

The translator tests run without a database. The adapter tests need a PostgreSQL URL and are skipped otherwise, so the coverage gate sees the new adapter only through the translator paths; a PostgreSQL service in CI would let `postgresAdapter.test.ts` run there.

## Reviewer Notes

- Foreign keys declared in the schema are kept on PostgreSQL. The first cut stripped them; the combo repository conformance test caught it, because better-sqlite3 enforces foreign keys by default and the code relies on `ON DELETE CASCADE`.
- PostgreSQL truncates identifiers to 63 bytes silently. The retire-provider trigger functions collided under that limit until names were fitted with a hash suffix (`fitIdentifier()`).
- `sqlite_master`, `PRAGMA table_info` and the `omniroute_*` helpers live in the target schema; `omniroute_bootstrap_version()` gates re-installs.
- Only `call_logs`, `conversation_turn_nodes`, `memories` and the three `domain_*` state tables get a hidden `rowid` identity column; tables with an `INTEGER PRIMARY KEY` or a single-column TEXT primary key map `rowid` to that key; composite keys are left alone.
- The translator, adapter, tokenizer, trigger translator, bootstrap SQL, error mapping and SQLite import were split so every function is under the repo's complexity 15 / cognitive 15 / 80-line limits. No behaviour change: both suites unchanged, `buildBootstrapSql()` output byte-identical, tokenizer and statement translation diffed against the previous implementation over 200,000 randomized inputs and a 132-statement corpus.

## CI status

Run [35085708389](https://github.com/diegosouzapw/OmniRoute/actions/runs/35085708389) on `0f70a4ba1`: `No new ESLint warnings`, `Vitest`, `Change Classification` and unit shards 1 and 3 green. Every remaining red reproduces on the untouched tip (`19b9d05e0`, checked in a throwaway worktree with no PR file present) or on #13850 against the same base, and none of it touches `src/lib/db/adapters/`, the storage settings, or a test this PR adds.

| Check | What fails | Owner |
| --- | --- | --- |
| Docs Gates | env/docs contract: `OMNIROUTE_CLOUD_SYNC_ENFORCE_SIGNATURE` read in code but not in `.env.example`, `CDP_PROXY_TOKEN` in `.env.example` but not in `ENVIRONMENT.md`; both present on the tip. The PR's own entry (`OMNIROUTE_PG_MODULE_PATH`) was flagged on the first run and is documented now. | base |
| Merge integrity | `check:agent-skills-sync` wants to generate the `cli-mcp` skill; same on #13850 | base |
| Fast Quality Gates | `mutation-test-coverage`: three tests missing from `stryker.conf.json` for `src/server/authz/routeGuard.ts`, `open-sse/utils/error.ts`, `open-sse/services/combo/rrState.ts`; none touched here | base |
| Unit Tests fast-path 2/4 and 4/4 | 16 tests: the two env/docs contract tests above, and 14 in `chatcore-translation-paths`, `context-editing-relays`, `hard-session-lease-bypass-inventory`, `pack-artifact-policy`, `reasoning-effort-clamp-and-retry`, `source-no-raw-nul-bytes`, each failing with the same count on the tip | base |

Before the rebase the PR had its own reds on top of those (`check:deps`, `check:file-size`, `check:complexity-ratchets`, `check:cycles`, the Storage section-order and dependency allowlist tests, the six `OMNIROUTE_DATABASE_*` variables and the seven `settings.postgres*` strings in the env/doc and Vietnamese tests). Those are gone. The branch carries no fix for the base items above, on purpose: per AGENTS.md they belong in the release branch's own base-red PR.


dpozimski:
Re-tested the branch from scratch today against the compose `postgres` profile and pushed 82f8d0ed3 with the fixes it turned up.

What ran: fresh volume, 170 migrations / 128 tables / 25 triggers as PL/pgSQL functions in about 3 s. Two dev servers on one schema: the second booted with nothing to migrate, used the first one's API key and provider connection, both wrote to the same `call_logs`, and `pg_stat_activity` showed both `omniroute-<pid>` sessions. `docker restart` of PostgreSQL under both replicas: one background poller logged "Connection terminated unexpectedly", the next request on each replica was 200. A SQLite instance went through `npm run db:migrate-to-postgres` (api_keys 1/1, call_logs 3/3, provider_connections 1/1, second run a no-op) and a replica over the imported schema answered with the SQLite-era key and connection. A local mock OpenAI-compatible node stood in for a real provider, so no external key was involved.

What broke:

1. Three processes booting at once on a schema that did not exist yet: the second one died in `connect()` with `UNIQUE constraint failed: pg_namespace.nspname`. `CREATE SCHEMA IF NOT EXISTS` ran before the advisory lock, and PostgreSQL does not make `IF NOT EXISTS` race-safe across sessions. Schema creation and the bootstrap helper install now take an advisory lock keyed by the schema name.
2. With that fixed, two of the three then died on `CREATE TABLE IF NOT EXISTS` for a lazily created table (unique violation on `pg_type`). Idempotent DDL in the adapter now treats 23505 / 42P07 / 42710 / 42701 / 42P06 as "already there", which also covers runtime lazy DDL outside the init lock.
3. The import skipped tables that OmniRoute only creates lazily at runtime: `audit_log` had 3 rows in the SQLite file and none arrived. Missing tables are now created from the SQLite definition before rows are copied.

Regression test: `ConcurrentOpen_OnFreshSchema_AllProcessesBootstrap` spawns three child processes against a fresh schema and expects all of them to come up. `tests/unit/db/**` on PostgreSQL: 133 of 138; the five left are the four `omp.test.ts` failures that reproduce on SQLite in this checkout and the `USING INDEX` planner-wording assertion. ESLint, Prettier and `typecheck:core` are clean.

One note for running the db suites on PostgreSQL with `OMNIROUTE_DATABASE_SCHEMA=auto`: every test process drops and recreates a 126-table schema, and eight of them in parallel exhaust the default lock table (`out of shared memory`). `--test-concurrency=4` or a larger `max_locks_per_transaction` on the test server avoids it.

dpozimski:
Second pass, pushed as 8df3afd15. A line-by-line review of the translator and the import turned up more than the end-to-end run did, so two of those went in and the rest are listed here for whoever picks the branch up.

Fixed in 8df3afd15:

- The import copied tables alphabetically, so a child such as `connection_runtime_state` or `webhook_deliveries` was inserted before its parent and the first batch failed with a foreign key violation, leaving a partial import. It passed in my earlier run only because those tables were empty. Tables are now ordered by the foreign keys declared on the PostgreSQL schema; the import test has a child table that sorts before its parent and fails without the fix.
- The `rowid` rewrite replaced the alias in `SELECT id AS rowId` (`src/lib/usage/providerWindowCosts.ts`), so the dedupe there read an undefined property. An identifier after `AS` is now left alone.

Still open, confirmed by running the translator or reading the exact code path:

1. `INSERT OR REPLACE` arbitrates on one constraint only (the PK when covered, else the first unique index). SQLite REPLACE resolves every unique constraint, so the restore path in `core.ts` raises 23505 for a combo with the same `name` but a different `id`.
2. `strftime('%s', created_at) / N * N` in `quotaSnapshots.ts` becomes floating point division through `omniroute_to_real`, so `(x / N) * N` is `x` and every snapshot lands in its own bucket.
3. `gamification.ts` selects joined columns grouped only by `ub.badge_id`; PostgreSQL rejects that with 42803, so `getAllEarnedBadges()` fails.
4. `json_object`, `json_array`, `json` and `json_group_array` cast to text before nesting, so nested JSON is stored as a string, and `json_set` with a `?` parameter resolves to the text overload.
5. A trigger body with `RAISE(ABORT, 'x % y')` becomes `RAISE EXCEPTION 'x % y'`, which plpgsql reads as a placeholder. No current migration does this. `omniroute_random()` overflows bigint on half its draws; nothing calls it from SQL yet.
6. Boolean forms outside a bare column (`WHERE (flag)`, `WHERE COALESCE(flag, 0)`, `col = TRUE` on a bigint) pass through untranslated; no current call site.

Plausible from reading, not reproduced:

7. Every parameterized query gets a named server-side prepared statement. After an `ALTER TABLE ... ADD COLUMN` from a runtime `ensureColumn` call or from another replica's migration, a prepared `SELECT *` on that table fails with 0A000 "cached plan must not change result type" until the connection is replaced, and that is not treated as a connection error. `DEALLOCATE ALL` after DDL, or unnamed statements, would close it.
8. A worker terminated by the Atomics timeout is never recreated, so the process loses its database until restart.
9. One unparseable value (`''` in an INTEGER column) fails the whole 500-row import batch; there is no per-row fallback.

dpozimski:
c0f4b2800: the Storage card under Settings > General now reflects the backend. `/api/storage/health` reported `driver: "sqlite"` and the `storage.sqlite` path regardless of the backend, so a replica on PostgreSQL told the operator its data was local. The route now reads the backend through `src/lib/db/postgresStatus.ts` (redacted connection, database, schema, server version, size, table count, connected replicas from `pg_stat_activity`), the card renders those, and file export/import, file backups, VACUUM and the SQLite page/cache optimization block are hidden on PostgreSQL with a pointer at `pg_dump`. JSON export/import and the purge actions stay. Seven new `settings.*` keys, added to every locale with the `__MISSING__` marker. Verified in the browser through a balancer in front of two dev replicas: the card shows schema `pr13118`, server 17.11, 128 tables, 2 connected replicas.

dpozimski:
@diegosouzapw pls review. my proposition is to allow to use psql to be able to scale omniroute in kubernetes (multiple nodes possibility)

diegosouzapw:
This is genuinely impressive engineering — reusing the existing sync adapter contract instead
of rewriting 1,300+ call sites, the dialect translator, the advisory-lock boot coordination for
concurrent replicas. Before we go further though: `#8075` is the project's own open thread
asking the community to design this backend together (Postgres vs. MySQL vs. both, dialect-
agnostic contract) before anyone commits to one implementation — could you post this PR there
as a concrete proposal so we can align on the shape first? Separately, a few things worth fixing
regardless of that discussion: the PR's own CI currently shows 8 failing checks (ESLint, docs
gates, fast quality gates, all 4 unit-test shards) on its own branch, and the new
`postgresAdapter`/`sqliteToPostgres` test suites are skipped unconditionally without a live
Postgres URL — meaning the riskiest new code (the SQL dialect translator) has no CI coverage as
shipped. Marking as `defer` for now, not because the idea is bad, but because a change this size
touching how OmniRoute talks to its database deserves the design conversation first.

dpozimski:
@diegosouzapw thanks for the review.

Why this exists: in my organization OmniRoute sits in front of hundreds of agents, batch workloads and AI tooling. A single instance runs out of CPU and RAM under that load, and with a per-process SQLite file there is nothing to scale to. We ran this branch as a load balancer in front of two OmniRoute instances on shared PostgreSQL and it works: both instances see the same keys, connections and logs. The biggest gap before was per-instance state such as `conversation_turn_nodes`, `agentic_conversations` and the other runtime tables living in each replica's own SQLite. With this PR they are in PostgreSQL, so a conversation started on one instance continues on the other.

On CI: rebased today onto the tip (`19b9d05e0`, 150 commits forward). The rebase surfaced two boot problems, both fixed with a failing-then-passing test. The durable-snapshot requirement is recomputed inside the writer transaction since 7ae8bf4e05 and that copy had no postgres exclusion, so a restart with a pending migration was refused with the hard-link message. And migration 178 touches `memory_fts` but is not in `OPTIONAL_FTS5_MIGRATION_VERSIONS` (only 180 is), which stops any driver without FTS5 at 178. That second one also hits a SQLite build without FTS5, so it may deserve its own PR; say the word and I split it out.

The PR's own reds are gone. What is still red on run 35085708389 reproduces on the untouched tip or on #13850 against the same base: the mutation-coverage entries for routeGuard, error.ts and rrState, the `cli-mcp` skill sync, two env/docs items, and 14 unit tests in six files this PR does not touch. The table in the body lists each one with its owner.

On coverage: `sqliteToPostgres.test.ts` (35 tests, the translator) needs no database and runs on every CI run. Only `postgresAdapter.test.ts` (14 tests) skips without `OMNIROUTE_TEST_DATABASE_URL`. If you want it in CI I can add a `postgres:17` service to the unit-test job in this PR, or leave that to a base change, your call.

On #8075: I can post this there as the concrete proposal for the shape (the existing sync adapter contract plus a dialect translator, behind one env var) if you prefer to have the design conversation there before looking at the diff again.

Issue #7285, opened 2026-07-15
fix(routing): combo failover misses OpenAI-shape streams that end without finish_reason — truncated response forwarded as success (auto/*:free)


diegosouzapw:
## Summary

Combo failover does **not** trigger when an OpenAI-shape upstream returns HTTP 200, emits some SSE bytes, and then closes **without ever sending `finish_reason`**. The truncated stream is forwarded to the client as a success. Clients surface it as *"Response ended without a finish reason and may be incomplete."*

**Source:** Discussion #7019 (reported by @bedusco, using `auto/coding:free` from Kilo Code / ZCode).

## Current behavior

`validateResponseQuality()` (`open-sse/services/combo/validateQuality.ts`) peeks the upstream stream before piping it to the client, and **already fails over** in two cases:

- Truly zero-byte 200 stream (`sawAnyBytes === false`) → `validateQuality.ts:278` → `{valid:false, reason:"streaming no recognized content"}`
- Claude-format complete lifecycle with zero content blocks → `validateQuality.ts:261` → failover (added by #3685)

But when an **OpenAI-format** upstream sends *any* bytes (a role-only delta, a ping, metadata) and then closes without a `finish_reason`:

1. `sawAnyBytes` becomes `true`, so the "truly empty" branch at `validateQuality.ts:278` is skipped.
2. `parseAccumulatedSse()` (`validateQuality.ts:180-206`) only recognizes **Claude** event types (`message_start`, `content_block_*`, `message_stop`, `message_delta`).
3. `hasOpenAICompatibleStreamValue()` (`open-sse/utils/streamHelpers.ts:249-266`) checks only `delta.content` / `delta.reasoning_content` / `delta.reasoning_text` / `delta.tool_calls` — **it never inspects `finish_reason`**.
4. Execution falls through to `validateQuality.ts:284-288` ("Incomplete lifecycle or non-Claude stream — replay all buffered bytes") → **`{valid: true}`** → forwarded to the client.

The code comment at `validateQuality.ts:270-275` documents the intent explicitly:

> Streams that carried ANY SSE activity ... keep the pass-through contract (#3399/#3685): those are handled by the stream-readiness timeout, not failover.

That deferral does not hold: `ensureStreamReadiness()` (`open-sse/utils/streamReadiness.ts`, wired at `chatCore.ts:4153-4165`) only guards **first-byte arrival**. Nothing checks how the stream *terminates*. So this class of truncation is caught by nobody on the default configuration.

[...]

PR #13126, opened 2026-09-09
feat(combo): detect degenerate streaming responses before forwarding

### LINKED ISSUE #12895: feat(resilience): Streaming degeneracy detection: fail-over models that loop or emit garbled content
## Summary
When a combo routes a request to a vision model inside an otherwise text-only
session (e.g. `auto/freevision` selected for plain prose), some upstreams stream
a clean HTTP 200 while emitting **looped or garbled tokens** — e.g. a repeated
substring ("轮胎的轮胎的…") or a repeated gibberish token. Because the gateway
relays it happily, the client renders the mojibake/looping text. These are
usually vision-capable models that were never meant to answer the text turn.

The current quality gate (`validateResponseQuality`) bounded-peeks only the
**start** of a stream and then passes healthy-looking content through. It catches
empty/missing content and the recently-added `findToolCallSpecViolation`
(duplicate tool calls), but nothing checks the assistant *prose* for repetition
or degeneracy.

## Proposal
Add a cheap, bounded-stream degeneracy check to the streaming peek in
`open-sse/services/combo/validateQuality.ts`:

- Accumulate only the **text deltas** seen while peeking (Claude `text_delta`,
  OpenAI `delta.content`), never the whole stream.
- Keep reading a **bounded** window (≈16 KB cap or until enough text to judge),
  not just the first content event. Healthy streams escape on real text volume,
  so added latency stays negligible per response.
- Mark the response invalid (→ combo skips this target / fails over) when any of:
  1. a short prefix unit repeats back-to-back ≥4× covering ≥50% of the window
     (looped output);
  2. the same whitespace-separated token repeats ≥5× consecutively (garbled-token loop);
  3. the window is long (≥48 chars) but has ≤6 distinct characters (mojibake).

No regex, O(n) over a capped window, low false-positive by construction.

Why not buffer the whole stream: OmniRoute's goal is to preserve streaming
latency; this only peeks a little further than today, and only for combo paths.

## Related
- Complements `findToolCallSpecViolation` (duplicate tool calls), which it does
  not cover.
- Complements per-combo non-streaming `responseValidation`.

## Environment
- OmniRoute desktop, ~v3.8.50
- Repro: a request routed via `auto/freevision` (vision models) into a text
  session; observed from `glm-4.1v-thinking-flash` and others returning repeated /
  low-diversity output over clean HTTP 200.

I have a working, unit-tested implementation of this locally and would be happy
to open a PR if maintainers are interested.

ftevxk:
Closes #12895

## Problem

Some upstream models (notably a vision model that got routed into a plain text session) can stream happily with HTTP \200\ while producing looped or low-diversity garbled output that the client renders as gibberish. The existing quality gate (\alidateResponseQuality\) only peeks at the **start** of a stream and forwards as soon as any content delta appears — by design, to avoid buffering the whole stream and defeating streaming's latency purpose. So genuinely degenerate output passes through unchanged to the client.

## Change

Extend the bounded SSE peek to accumulate assistant text deltas (\peekText\) and scan that window with a new \detectDegeneration()\ that fires on clear, low-false-positive signal:

- **prefix-unit repetition** — a leading unit (2–6 chars) repeated 4+ extra times covering &ge;50% of the window → \
epetitive output ...)
- **token loop** — a whitespace-separated token repeated 5+ consecutive times → \looped token ...)
- **low diversity** — a window &ge;48 chars containing &le;6 distinct chars → \low-diversity garbled output\

When degenerate output is found the stream is marked invalid and fails over to the sibling combo target, exactly like the existing empty-content checks.

## Tunable latency bounds (healthy streams barely notice)

- escape once **80 chars of real text** accumulate → pass through
- hard **16 KiB byte cap** on the peek window
- **2 KiB** empty-text cap for tool_use streams (never produces assistant text)

So healthy responses get forwarded almost immediately, while degenerate ones are caught within the same bounded window.

## Notes

- This detects **only the clearest** degeneration to avoid false positives on legitimate verbose/repetitive-but-valid prose.
- It is a streaming-path-only change; non-streaming validation is untouched. All upstream \indToolCallSpecViolation\ / lifecycle / #7285 / #10404 behavior is preserved (pure additions + one bounded substitute for the immediate-forward branch).

### DIFF
--- a/open-sse/services/combo/validateQuality.ts
+++ b/open-sse/services/combo/validateQuality.ts
@@ -71,6 +71,91 @@ export function toRetryAfterDisplayValue(value: ComboRetryAfter): string | Date 
 const EXHAUSTION_MARKER_PATTERN =
   /\b(insufficient\s+credit|insufficient\s+balance|quota\s+exceeded|out\s+of\s+credits?|credit\s+exhausted)\b/i;
 
+// ── Streaming degeneracy detection ───────────────────────────────────────────
+// Some upstreams (notably a vision model routed into a plain text session) can
+// stream happily with HTTP 200 while producing looped/garbled output the client
+// would render as gibberish. This only detects CLEAR, low-false-positive signal
+// so pass-through semantics for healthy streams are preserved. Bounded to a
+// small peek window so the added latency for healthy streams stays negligible.
+const DEGENERATION_SCAN_MIN = 16; // min accumulated text chars before any scan
+const DEGENERATION_REPEAT_MIN_RUN = 4; // a unit must repeat this many extra times
+const DEGENERATION_REPEAT_COVER_RATIO = 0.5; // and cover ≥ this share of the window
+const DEGENERATION_TOKEN_LOOP_MIN = 5; // consecutive identical tokens == loop
+const DEGENERATION_LOWDIVERSITY_MIN = 48; // window size for the diversity check
+const DEGENERATION_LOWDIVERSITY_DISTINCT_MAX = 6; // ≤ this distinct chars == garbled
+// Bounded peek window so healthy streams don't pay meaningful added latency:
+const DEGENERATION_HEALTHY_ESCAPE_TEXT = 80; // enough real text, not degenerate -> pass
+const DEGENERATION_PEEK_BYTES_CAP = 16 * 1024; // hard byte cap (bounds latency)
+const DEGENERATION_TOOLUSE_ESCAPE_BYTES = 2 * 1024; // tool_use with zero text -> pass
+
+/** Append the assistant text a chunk carries (Claude text_delta or OpenAI delta.content). */
+function extractAssistantDeltaText(parsed: unknown): string {
+  if (!parsed || typeof parsed !== "object") return "";
+  const rec = parsed as Record<string, unknown>;
+  const delta = rec.delta as Record<string, unknown> | undefined;
+  if (delta && delta.type === "text_delta" && typeof delta.text === "string") {
+    return delta.text;
+  }
+  const choices = rec.choices;
+  if (Array.isArray(choices)) {
+    let out = "";
+    for (const ch of choices) {
+      if (!ch || typeof ch !== "object") continue;
+      const content = (ch as Record<string, unknown>).delta as
+        | { content?: unknown }
+        | undefined;
+      if (content && typeof content.content === "string") out += content.content;
+    }
+    return out;
+  }
+  return "";
+}
+
+/**
+ * Detect obvious degenerate output (repetition loops / very-low-diversity
+ * gibberish) in a peeked text window. Returns a short reason, or null when the
+ * window looks like real prose or is too short to judge.
+ */
+function detectDegeneration(text: string): string | null {
+  const t = text.trim();
+  if (t.length < DEGENERATION_SCAN_MIN) return null;
+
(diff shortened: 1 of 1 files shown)


NaNomicon:
## Interaction with reasoning-only streams (combo buffering)

I just filed #13620 for a related defect in the same peek, and this PR touches the branch that decides when the peek stops reading — worth checking before merge.

### The defect in #13620

`hasOpenAICompatibleStreamValue()` (`open-sse/utils/streamHelpers.ts:271`) tests `delta.content`, `delta.reasoning_content`, `delta.reasoning_text`, `delta.tool_calls` — but not bare `delta.reasoning`. Every Ollama Cloud reasoning model emits bare `reasoning` with `content: ""`.

Measured on a combo route vs. a direct model id, same prompt, same gateway, only the dispatch path differs:

| target | path | reasoning deltas | first reasoning | reasoning span |
|---|---|---|---|---|
| combo A | combo | 14,051 | **196.2 s** | 1.243 s |
| combo B | combo | 22,002 | **208.3 s** | 1.840 s |
| same model | direct | 10,879 | **1.7 s** | 127.659 s |

So today: the peek fails to recognise a reasoning-only prefix, buffers it in full, and `buildReplayResponse()` flushes it at the first content delta.

### Why it matters here

This PR changes the same exit condition. Before, `if (outcome === "content")` returns immediately — the reasoning prefix that *is* recognised leaves promptly. After, that path keeps reading until `peekText.length >= DEGENERATION_HEALTHY_ESCAPE_TEXT` (or a byte cap trips), and the terminal replay branch additionally runs a degeneracy check.

For a stream that spends a long time in reasoning before any content, that is a strictly wider buffering window on top of the one in #13620 — the recognised-content case now also waits for a text/byte threshold rather than forwarding as soon as it has something valid.

Two questions, and I'd be glad to be wrong on either:

1. **Is there a time bound on the new wait?** If a lane reasons for minutes and the healthy-escape threshold is text-length based, does the peek hold the connection that whole time, or does something cap it?
2. **Does `detectDegeneration` need to consider reasoning text separately?** If `peekText` only accumulates assistant delta *content*, a reasoning-heavy-but-healthy stream may look like "no text yet" for a long time, which is exactly the population this PR is inspecting for degeneracy.

If a maintainer would rather keep the scopes separate, the minimal ordering is: land the `hasAnyReasoningSignal(delta)` fix from #13620 first so reasoning-only streams exit the peek, then layer the degeneracy detection on top of a peek that is no longer blind to the reasoning phase.

Happy to test a branch of this PR against a reasoning-heavy combo prompt and report timings if that's useful.

diegosouzapw:
The degenerate-streaming detection itself looks solid — the repetition/token-loop/
low-diversity heuristics are well-bounded (fixed peek window, plain substring
comparison, no regex — no ReDoS concern) and the existing quality-validator suite
passes clean at your head (30/30, no regression). The one thing missing before this
can merge: `detectDegeneration()` has zero test coverage of its own — the PR's diff is
`validateQuality.ts` only, and the function isn't exported, so nothing exercises the
three detection signals or proves the false-positive guard holds on legitimate
verbose/repetitive prose. Could you add a few cases (one per signal + one deliberate
false-positive guard) through `validateResponseQuality()`'s existing stream-mock
pattern? Also needs a changelog fragment under `changelog.d/features/`.

Issue #11722, opened 2026-08-26
feat(providers): expose a usage-fetch capability in provider-plugin-manifest


maxmad64bis:
## Problem / Use Case

An external dashboard that integrates OmniRoute wants to show, per provider, whether usage or quota can be fetched. Today there is no API for that - it has to read the source file `services/usage.ts` and re-check it after every update.

## Proposed Solution

The manifest already tells integrators what each provider can do - it just doesn't cover usage fetching yet. Add a `usage-fetch` capability for providers that already have a wired fetcher. Then a dashboard can read it straight from the public manifest (list `USAGE_FETCHER_PROVIDERS`, ~46 providers; open-sse/services/usage.ts:95).

This is a proposal, not a spec. Open question: there are two lists - `USAGE_FETCHER_PROVIDERS` (fetcher wired) and `USAGE_SUPPORTED_PROVIDERS` (gates the quota widget per #10078, src/shared/constants/providers.ts:486). Should we expose only `usage-fetch`, or also a `usage-supported` signal? I'm not pushing a shape - just surfacing the gap. For context, usage fetching is actively being extended (#11498, open).

## Acceptance Criteria

- `capabilitiesFor()` (open-sse/config/providerPluginManifest.ts:3, built at :104) emits `usage-fetch` for `USAGE_FETCHER_PROVIDERS` providers.
- Documented in the manifest schema.
- No change to fetching or the quota UI.

## Area

Analytics / Usage Tracking, Provider Support, Dashboard / UI

## What this issue is NOT

No new fetcher, no quota change, no activation - pure discoverability. Any code would be a separate PR.

luw2007:
Supporting this. We have a concrete Sub2API use case that exposes a second gap beyond manifest discoverability.

OmniRoute's merged New-API/One-API/Sub2API support (#9415 / #9539) reads aggregator balance from `GET /api/user/self`. Our Sub2API deployment also exposes upstream subscription windows that are not covered by that balance: Codex 5-hour and 7-day usage/reset windows, and Claude 7-day usage/reset windows. We currently carry a local `getUsageForProvider()` integration to convert those signals into the existing `UsageQuota` shape, so Dashboard Limits and generic quota preflight/Combo routing can consume the same data.

[...]

Issue #10078, opened 2026-08-11
fix(providers): AgentRouter quota not shown on Dashboard Quota (USAGE_SUPPORTED_PROVIDERS gap)


nordz0r:
## Bug summary

AgentRouter (`agentrouter`) has a working balance quota fetcher for routing preflight/monitor, but the connection **does not appear on Dashboard → Quota / home Provider Quota widget**. Operators only see routing work; remaining credits are invisible in the UI.

## Environment

- OmniRoute: `3.8.50` (`next-web` / branch `release/v3.8.50`)
- Provider: built-in `agentrouter` (API key `sk-…` + optional New-API System Access Token / `New-Api-User`)
- Dashboard: `/dashboard/quota` and home `ProviderQuotaWidget`

## Background

- Feature request #6850 asked for AgentRouter credit wallet via New-API:
  - `GET /api/user/self` with `Authorization: Bearer {System Access Token}` + `New-Api-User: {userId}` → raw `quota`
  - dollar balance ≈ `quota / quota_per_unit` (default `500000`)
- Closed by PR #7653 (`feat(sse): quota tracking for AgentRouter, v0, FreeModel`), which added:
  - `open-sse/services/agentrouterQuotaFetcher.ts` (registered via `quotaTrackersBatch.ts` into **quotaPreflight + quotaMonitor**)
  - Dashboard connection fields for `providerSpecificData.consoleApiKey` + `newApiUserId` (`AgentrouterConsoleFields.tsx`)
  - Changelog claimed “feeding preflight **+ the dashboard's Provider Quota card**”

## Root cause (verified on `release/v3.8.50`)

Two separate surfaces; only the SSE/preflight path was wired.

### 1. Dashboard filters out `agentrouter`

`ProviderQuotaWidget` (and the same pattern on the quota page) keeps only:

```ts
USAGE_SUPPORTED_PROVIDERS.includes(connection.provider) &&
(connection.authType === "oauth" || connection.authType === "apikey")
```

On `release/v3.8.50`, `USAGE_SUPPORTED_PROVIDERS` in `src/shared/constants/providers.ts` **does not include** `"agentrouter"` (nor aliases). Same gate: `supportsProviderQuota()` in `src/shared/utils/providerQuotaVisibility.ts`.

So even with a healthy connection and console credentials filled, AgentRouter never enters the quota UI list. Same class of gap as open #9740 (Cline Pass).

### 2. Usage / provider-limits path never dispatches AgentRouter

[...]

PR #13133, opened 2026-09-09
feat(usage): show OpenRouter credit balance in Provider Quota

### LINKED ISSUE #13080: [Feature] show credit balance for providers like Cheaper Inference etc.
### Problem / Use Case

in Provider Quota - show credit balance for providers like Cheaper Inference etc.

### Proposed Solution

see above

### Alternatives Considered

_No response_

### Acceptance Criteria

see above

### Area

Provider Support, Analytics / Usage Tracking

### Related Provider(s)

Cheaper Inference, OpenRouter, UnoRouter

### Additional Context

_No response_

### Expected Test Plan

_No response_

abhisheksharma2411:
Closes #13080 for OpenRouter.

## The balance was already being computed and thrown away

`getOpenrouterUsage` (`open-sse/services/usage/openrouter.ts`, #6842) returns a USD `credits` quota built from `/api/v1/key` + `/api/v1/credits`, `services/usage.ts` dispatches `case "openrouter"`, and `USAGE_FETCHER_PROVIDERS` — the file that calls itself the single source of truth for which providers have a wired fetcher — lists it.

Neither app-side gate did. Measured rather than read:

```
isSupportedUsageConnection({ provider: "openrouter", authType: "apikey" })  ->  false
isSupportedUsageConnection({ provider: "deepseek",   authType: "apikey" })  ->  true     (control)
```

So `syncProviderLimits` skipped every OpenRouter connection and nothing ever asked the fetcher for the number.

`isSupportedUsageConnection` is an AND, which is why this is two entries and not one:

- `USAGE_SUPPORTED_PROVIDERS` — `src/shared/constants/providers.ts`
- `PROVIDER_LIMITS_APIKEY_PROVIDERS` — `src/lib/usage/providerLimits.ts`, needed because OpenRouter authenticates by API key (`apikey/gateways.ts:106`)

Add only the first and the connection is still refused — which is what the second test pins, since the boolean alone can't tell the two apart.

## The recurring part

The allowlist already carries this scar twice:

```ts
// Alibaba Coding Plan triple-window quota (#9603 UI gap — fetcher existed, list entry missing)
"bailian-coding-plan",
```

and `providerPluginManifest.ts` references #11722 for the same drift. `usage-families-split.test.ts` pins the dispatcher against `USAGE_FETCHER_PROVIDERS`, so the *open-sse* side can't drift — but nothing pinned the app-side lists against it, which is the seam all three issues came through.

So the test also guards the class: a provider that gains a usage fetcher without an app-side entry now fails.

**Three providers are already in that state and I have not touched them** — `opencode`, `opencode-zen`, `xai`. I could not establish from the tree whether those omissions are deliberate (`xai-oauth`/`xao` *are* listed while the API-key `xai` is not, which looks like it could be intentional), so they are named in an explicit set rather than silently "fixed" or left invisible. The set can't grow without a reason; converting any of them into a real entry is a call for someone who knows the intent. Happy to file that separately if you'd like it tracked.

## Verification

Mutations, each killed by the right test:

| mutation | result |
|---|---|
| revert the `USAGE_SUPPORTED_PROVIDERS` entry | all 3 fail |
| revert **only** the apikey gate (half the fix) | only the reachability test fails |
| add a fetcher with no app-side entry | only the drift guard fails |

- **182 related tests pass**, 0 fail — `provider-limits*`, `provider-plugin-manifest`, `usage-families-split`, `agentrouter-quota-*`, `openrouter-*`
- `eslint` clean on all three files
- `tsc --noEmit` error count **identical to the base branch** (4836 both sides, none of them in these files — the repo's typecheck is already dirty, I just made sure I added nothing)

## One thing to skip while reading the diff

The pre-commit hook ran `prettier --write` over the files I staged and fixed pre-existing formatting drift in them. Not mine, and safe to ignore:

- `providerLimits.ts:207` — a call collapsed onto one line
- `providerLimits.ts:397`, `providers.ts:75`, `providers.ts:144` — stray blank lines removed

My actual change is four lines: two list entries with a comment each.

I did not verify against a live OpenRouter account — I don't have one with credits on it — so this is "the connection now reaches the fetcher, and the fetcher's own shaping is unchanged", not "I watched the card render a balance". The `credits` quota it emits already carries `currency: "USD"`, which `quotaParsing.ts` renders as a dollar figure (same path AgentRouter's balance took in #10078).

### DIFF
--- a/src/lib/usage/providerLimits.ts
+++ b/src/lib/usage/providerLimits.ts
@@ -89,6 +89,8 @@ const PROVIDER_LIMITS_APIKEY_PROVIDERS = new Set([
   "qwen-cloud-token-plan",
   // AgentRouter (New-API) console System Access Token + New-Api-User id (providerSpecificData)
   "agentrouter",
+  // OpenRouter API key → GET /api/v1/key + /api/v1/credits (USD balance, credit cap)
+  "openrouter",
 ]);
 const DEFAULT_PROVIDER_LIMITS_SYNC_INTERVAL_MINUTES = 70;
 const PROVIDER_LIMITS_AUTO_SYNC_SETTING_KEY = "provider_limits_auto_sync_last_run";
@@ -205,11 +207,7 @@ export async function refreshAndUpdateCredentials(
   connection: ProviderConnectionLike,
   opts: CredentialRefreshOptions = {}
 ) {
-  return refreshAndUpdateCredentialsWithResolver(
-    connection,
-    getCredentialRefreshExecutor,
-    opts
-  );
+  return refreshAndUpdateCredentialsWithResolver(connection, getCredentialRefreshExecutor, opts);
 }
 
 function isUsageAuthError(message: unknown): boolean {
@@ -396,7 +394,6 @@ export function shouldClearErrorStateOnValidProbe(
  * — keeps the connection locked, matching the kimi-coding partial-refresh
  * semantics.
  */
-
 
 /**
  * Is an explicit cooldown still in the future?
--- a/src/shared/constants/providers.ts
+++ b/src/shared/constants/providers.ts
@@ -75,7 +75,6 @@ export function getProviderConnectionFamilyIds(providerId: unknown): readonly st
 
 // Web / Cookie Providers
 
-
 // API Key Providers
 
 // Sub-categories within APIKEY_PROVIDERS (used by dashboard and catalog views).
@@ -145,7 +144,6 @@ export const AGGREGATOR_PROVIDER_IDS = new Set([
   "helixmind",
   "tabitoken",
   "logfare",
-
 ]);
 
 export const ENTERPRISE_CLOUD_PROVIDER_IDS = new Set([
@@ -539,6 +537,8 @@ export const USAGE_SUPPORTED_PROVIDERS = [
   "qwen-cloud-token-plan",
   // AgentRouter (New-API) console balance quota (consoleApiKey + newApiUserId)
   "agentrouter",
+  // OpenRouter /key + /credits USD balance (#13080 UI gap — fetcher existed, list entry missing)
+  "openrouter",
 ];
 
 // ── Zod validation, lazily on first AI_PROVIDERS access (perf: skips the walk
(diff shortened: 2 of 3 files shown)


abhisheksharma2411:
Closing this — **it was already fixed and I missed it.** @Mr-White landed `openrouter` in both gates on 2 Sep in #12256 ("devin-cli agentic quota + openrouter credits in Provider Limits"), which is six days *before* #13080 was filed.

My checkout was 291 commits behind `release/v3.8.51`, so the gap I measured was real on my tree and stale on yours. The rebase is what surfaced it: both of my hunks conflicted with your existing entries, and `USAGE_SUPPORTED_PROVIDERS` has since moved to `open-sse/services/usage/usage/supportedProviders.ts` anyway. Entirely my error for not re-syncing before measuring, and sorry for the review noise.

Two things from the exercise that may still be worth something, so it isn't a total waste of your time:

**1. Nothing pins the two lists against each other, and they are drifting again right now.** `usage-families-split.test.ts` pins the dispatcher against `USAGE_FETCHER_PROVIDERS`, so the open-sse side can't drift — but nothing compares that list to the app-side supported list, which is the seam #9603, #11722 and #12256 all came through. On current `release/v3.8.51`:

```
fetchers: 52   supported: 50
HAS FETCHER, NOT SUPPORTED: ["opencode", "opencode-zen", "xai"]
SUPPORTED, NO FETCHER:      ["xiaomi-mimo-token-plan"]
```

I'll send that as a test-only PR rather than reopening this one.

On the reverse-drift entry, before anyone spends time on it: `xiaomi-mimo-token-plan` is in the supported list with no dispatcher case, so it would hit `default: "Usage API not implemented"` — but it is *not* in `PROVIDER_LIMITS_APIKEY_PROVIDERS` and it authenticates by API key, so `isSupportedUsageConnection` refuses it first. **Dead config, not a live defect.** I checked that before writing it up.

**2. #13080 is only a third done.** It asked for Cheaper Inference and UnoRouter as well as OpenRouter, and neither has a usage fetcher. Commenting there with the current state so it isn't closed as done on the strength of #12256.

Issue #9603, opened 2026-08-06
fix(qwen-cloud-token-plan): quota is missing and exhausted accounts do not recover after reset


ivkhokhlov:
## Bug summary

The `qwen-cloud-token-plan` provider can perform inference, but its quota integration and multi-account recovery are incomplete:

1. Token Plan usage is not displayed in **Provider Quota**.
2. A temporary Token Plan quota exhaustion can be classified as terminal `credits_exhausted`, so account rotation does not recover the account automatically after the Qwen quota resets.

This makes a pool of multiple Qwen Cloud Token Plan accounts unsafe to run unattended.

## Environment

- OmniRoute: reproduced on `3.8.49`; the missing Provider Quota registration is still present on the current `release/v3.8.50` branch
- Provider: `qwen-cloud-token-plan`
- Region: Qwen Cloud global / Singapore
- Inference base URL: `https://token-plan.ap-southeast-1.maas.aliyuncs.com/compatible-mode/v1`
- Multiple Token Plan accounts configured as separate provider connections
- Provider implementation: #7854 / #7882

## Problem 1: Provider Quota is missing

The connection has quota visibility enabled, but `qwen-cloud-token-plan` does not appear in Provider Quota.

The Qwen Cloud console exposes both quota windows:

- 5-hour usage limit: total, remaining percentage, and reset time
- 7-day usage limit: total, remaining percentage, and reset time
- subscription status, expiry date, and remaining days

Current OmniRoute behavior: no quota row or Token Plan usage information is shown.

The current `release/v3.8.50` source does not register `qwen-cloud-token-plan` in:

- `USAGE_SUPPORTED_PROVIDERS` in `src/shared/constants/providers.ts`
- `PROVIDER_LIMITS_APIKEY_PROVIDERS` in `src/lib/usage/providerLimits.ts`
- `USAGE_FETCHER_PROVIDERS` or the dispatcher in `open-sse/services/usage.ts`

There is also no provider-specific usage fetcher.

### API constraint discovered during investigation

The inference API key works for `/models` and inference, but does not expose `/usage`, `/quota`, or `/subscription`.

The Qwen Cloud web console obtains the data through its console gateway and appears to require an authenticated browser session plus `sec_token`, rather than only the inference key. The observed console methods are:

- `/tokenplan/personal/api/v2/subscription`
- `/tokenplan/personal/api/v2/usage`
- `/tokenplan/personal/api/v2/quota-config`

[...]

PR #12950, opened 2026-09-07
feat(volcengine): canonical quota window mapping and safe multi-connection plan binding

HouMinXi:
## Summary

Volcengine Ark console auto-login and quota fetchers had several integration issues:
1. Console cookies and CSRF tokens were returned unsanitized in REST provider responses.
2. Plan usage quota window keys (`session`, `AFPFiveHour`) differed from the canonical names (`session (5h)`, `weekly (7d)`), causing genericQuotaFetcher to miss them or fall back to single-minute polls.
3. Time window keys could leak into `modelWindows`, treating rate-limit windows as model names.
4. Auto-binding could clobber existing connections or fail to adopt single connections when names differed from official defaults.
5. `detectPlan` reported `available: true` on accounts without purchased quota windows.

This PR fixes response sanitization, window normalization, quota fetcher mapping, unpurchased plan gating, and connection reuse in plan binding.

## Changes

- **Response sanitization** (`src/lib/providers/requestDefaults.ts`): Strips `volcConsoleCookie` and `volcCsrfToken` from provider connection serialization.
- **Quota window canonicalization** (`open-sse/services/usage/volcenginePlan.ts`): Normalizes Coding Plan and Agent Plan window keys to `session (5h)` and `weekly (7d)`. Guards against division by zero (`remainingPercentage: 0` when `total === 0`).
- **Quota fetcher isolation** (`open-sse/services/genericQuotaFetcher.ts`): Maps Volcengine window aliases to `window5h` and `window7d`. Introduces `TIME_WINDOW_KEYS` set to isolate rate-limit windows from `modelWindows`.
- **Plan detection guard** (`src/lib/providers/volcenginePlanBinding.ts`): Checks for non-empty active quota windows before marking a plan available.
- **Connection reuse and adoption** (`src/lib/providers/volcenginePlanBinding.ts`):
  - Prioritizes `targetConnectionId` (with provider affinity check).
  - Matches by valid `apiKey` or `volcApiKeyId`.
  - Matches by default provider plan name.
  - Safely adopts the single existing connection when exactly one connection exists for the provider.
  - When multiple connections exist without a match, creates a new connection rather than overwriting.
  - Shallow merges `existingPsd` with new console session data to preserve user-configured flags like `autoFetchModels`.

## Test plan

- [x] `tests/unit/volcengine-plan-request-sanitization.test.ts` (1 pass)
- [x] `tests/unit/volcengine-plan-quota-windows.test.ts` (2 pass)
[...]

PR #13134, opened 2026-09-09
test(usage): pin fetcherProviders against supportedProviders

abhisheksharma2411:
Test-only. No production file touched.

## Why

`open-sse/services/usage/fetcherProviders.ts` and `supportedProviders.ts` are sibling pure-data modules that have to agree, and nothing compared them:

- in `fetcherProviders` but not `supportedProviders` → the provider computes a quota nobody ever asks for, because `isSupportedUsageConnection` refuses the connection first
- in `supportedProviders` with no dispatcher case → accepted, then falls through to `default: "Usage API not implemented"`

That seam has now produced the same bug three times:

| | |
|---|---|
| #9603 | `"bailian-coding-plan", // #9603 UI gap — fetcher existed, list entry missing` — the comment is still in the list |
| #11722 | referenced in `providerPluginManifest.ts` for the same drift |
| #12256 | openrouter credits — and #13080 asked for them again **six days after the fix landed**, because the card had never appeared |

`usage-families-split.test.ts` already pins the dispatcher against `fetcherProviders`, so the open-sse side can't drift. This pins the other edge.

I found it the embarrassing way: I measured the openrouter gap on a checkout 291 commits stale, wrote the two-line fix, and only discovered @Mr-White had already landed it in #12256 when the rebase conflicted with his entries. The list drift was real; my instance of it wasn't. This is the part of that work that still has value.

## Current divergence, declared with reasons

Divergence stays allowed — it just has to be stated:

```
opencode, opencode-zen, xai   have a fetcher, not offered to the dashboard
xiaomi-mimo-token-plan        offered, no fetcher
```

**I have not "fixed" any of them, on purpose.** For the first three I could not establish intent from the tree — `xai-oauth`/`xao` are offered while the API-key `xai` is not, which reads like it may well be deliberate. Guessing would either break something intentional or paper over a real gap. They're pinned so the set can't grow silently, and converting any entry into a real listing is a call for whoever knows the reason.

`xiaomi-mimo-token-plan` I did chase down before writing it up, because "offered with no fetcher" sounds like a live 
defect: it authenticates by API key and is **absent from `PROVIDER_LIMITS_APIKEY_PROVIDERS`**, so `isSupportedUsageConnection` refuses it before the dispatcher is reached. **Dead config, not a live defect** — worth knowing before someone "fixes" it into a path that returns `"Usage API not implemented"`.

A third test fails if a declared divergence stops diverging, so the allowlists can't go stale and quietly excuse a future recurrence of the same id.

## Verification

Mutations, each killed by the right test:

| mutation | fails |
|---|---|
| new fetcher, no dashboard entry | forward guard + stale check |
| dashboard entry, no fetcher | reverse guard + stale check |
| a declared divergence gets fixed | stale check alone |
| duplicate id in either list | duplicate test alone |

Worth noting on the first and last of those: my initial mutation run reported them as *surviving*, and that was wrong — my `perl` anchor assumed `"agentrouter"` was the last entry, which was true on my stale checkout and not on `dev`. The substitution silently no-op'd. Re-ran asserting the edit actually landed before trusting the result; both fail correctly.

- **105 related tests pass**, 0 fail — this file, `usage-families-split`, `provider-plugin-manifest`, `provider-limits*`
- `eslint` clean

One environment note in case it matters to anyone reproducing: `npx eslint` initially died with `Cannot find package '@eslint/compat'` because my `node_modules` predated it being added to `devDependencies`. Installed it with `--no-save`; `package.json` and the lockfile are untouched.

### DIFF
--- a/tests/unit/usage-provider-list-drift.test.ts
+++ b/tests/unit/usage-provider-list-drift.test.ts
@@ -1,0 +1,115 @@
+/**
+ * usage-provider-list-drift.test.ts
+ *
+ * `fetcherProviders.ts` and `supportedProviders.ts` are sibling pure-data lists
+ * that have to agree: the first says a provider has a wired
+ * `getUsageForProvider`, the second says the dashboard and server gates will
+ * accept its usage. A provider in the first but not the second computes a quota
+ * nobody asks for; one in the second but not the first is accepted and then
+ * falls through the dispatcher to `default: "Usage API not implemented"`.
+ *
+ * Nothing compared them, and that seam has now produced the same bug three
+ * times — #9603 (bailian: "fetcher existed, list entry missing"), #11722, and
+ * #12256 (openrouter credits, requested again in #13080 six days after the fix
+ * landed because the card had never appeared). `usage-families-split.test.ts`
+ * pins the dispatcher against `fetcherProviders.ts`, so the open-sse side
+ * cannot drift; this pins the other edge.
+ *
+ * Divergences are allowed but must be declared with a reason. An undeclared one
+ * is the bug.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+const { USAGE_FETCHER_PROVIDERS } =
+  await import("../../open-sse/services/usage/fetcherProviders.ts");
+const { USAGE_SUPPORTED_PROVIDERS } =
+  await import("../../open-sse/services/usage/supportedProviders.ts");
+
+/**
+ * Has a usage fetcher, deliberately not offered to the dashboard.
+ *
+ * These three predate this test and I could not establish intent from the tree,
+ * so they are pinned rather than "corrected" — the set may not grow silently,
+ * and converting any entry into a real dashboard listing is a call for someone
+ * who knows why it was left out. `xai-oauth`/`xao` *are* listed while the
+ * API-key `xai` is not, which reads like it could be deliberate.
+ */
+const FETCHER_WITHOUT_DASHBOARD_ENTRY = new Set(["opencode", "opencode-zen", "xai"]);
+
+/**
+ * Offered to the dashboard with no fetcher behind it.
+ *
+ * `xiaomi-mimo-token-plan` is inert rather than broken: it authenticates by API
+ * key and is absent from `PROVIDER_LIMITS_APIKEY_PROVIDERS`, so
+ * `isSupportedUsageConnection` refuses it before the dispatcher is ever
+ * reached. Listed here so the entry is understood as dead config instead of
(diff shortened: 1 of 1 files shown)


abhisheksharma2411:
Both red checks on this PR are pre-existing breakage on `release/v3.8.51`, not from this change — which touches exactly one file, `tests/unit/usage-provider-list-drift.test.ts`. Verified rather than assumed:

**Docs Gates (fast-path)** — stale migration counts after #12832 landed the 172nd migration:

```
✗ README.md — stale migrations: "171 migrations" — code has 172
✗ AGENTS.md — "171 migrations" — code has 172
✗ llm.txt   — "171 migrations", "171 versioned SQL migration", "171 SQL migrations"
✗ 3 STRICT drift(s) detected.
```

```
base branch   check-docs-counts-sync.mjs -> exit 1, 3 STRICT
```

Fixed in #13135 (it's a 44-file change, because the 41 `docs/i18n/*/llm.txt` mirrors are compared against the root body).

**No new ESLint warnings** — three `as any` casts in a file added by #12950:

```
✗ tests/unit/volcengine-plan-binding-upsert.test.ts:70  @typescript-eslint/no-explicit-any
✗ tests/unit/volcengine-plan-binding-upsert.test.ts:96  @typescript-eslint/no-explicit-any
✗ tests/unit/volcengine-plan-binding-upsert.test.ts:120 @typescript-eslint/no-explicit-any
```

Not a file this PR touches, and I've left it alone — typing `upsertConnection`'s test-facing signature properly is a call for whoever owns that code.

`Vitest (fast-path)` is green across all four shards, which is the check that actually exercises this change.

diegosouzapw:
Appreciate the transparency about the openrouter finding already being fixed in #12256 — the
drift-guard test itself still earns its keep given this exact class of bug has recurred 3
times. Ran it at your head: 4/4 pass, real assertions against the production modules. Merge-
ready.

abhisheksharma2411:
Thanks — and sorry for the slow reply, I lost access for most of today.

Nothing outstanding from me on this one; it's unchanged at `6643ab8b` since your review. The drift-guard is the whole of it.

Worth adding one thing to your "this class of bug has recurred 3 times" note, since it happened again *today* in a neighbouring file: `check-mutation-test-coverage --strict` fails on a pristine `release/v3.8.51` with no PR diff, because six covering unit tests across four mutated modules aren't in `stryker.conf.json` `tap.testFiles` — so their mutant kills haven't been counting. That's #13814.

Same shape as this PR: a list that has to agree with the code, and nothing comparing them until someone notices. The drift-guard here earns its keep for exactly that reason.

diegosouzapw:

PR #13135, opened 2026-09-09
docs: migration count is 172, unblocking the Docs Gates on every PR

abhisheksharma2411:
Counts only, no prose changed. **This currently fails on every PR opened against `release/v3.8.51`**, so it's blocking the queue rather than any one branch.

## What's broken

`check-docs-counts-sync.mjs` counts `.sql` files under `src/lib/db/migrations` and compares the number against the docs. #12832 landed the 172nd migration yesterday without updating them, and the gate is STRICT:

```
✗ README.md — stale migrations: "171 migrations" — code has 172
✗ AGENTS.md — stale migrations: "171 migrations" — code has 172
✗ llm.txt   — stale migrations: "171 migrations", "171 versioned SQL migration",
              "171 SQL migrations" — code has 172
✗ 3 STRICT drift(s) detected.
```

```
base branch   node scripts/check/check-docs-counts-sync.mjs  ->  exit 1, 3 STRICT
this branch   same command                                   ->  exit 0
```

## Why it's 44 files and not 3

Editing root `llm.txt` alone trades one red gate for another. The 41 `docs/i18n/*/llm.txt` files carry a translated header over the **root English body**, and `check-docs-sync.mjs` compares those bodies against root — so the first attempt produced 41 fresh failures:

```
[docs-sync] FAIL - docs/i18n/ar/llm.txt differs from root llm.txt
... × 41
```

Worth knowing before anyone else fixes a count in `llm.txt`: it's a 42-file edit, not a 1-file edit. Same substitution applied to all of them:

```
[docs-sync] llm.txt i18n mirrors match root content: 41 locales
[docs-sync] PASS - documentation version sync is consistent.
```

The two drifts the counts gate still reports (A2A skills, cloud agents) are **soft** warnings and are present on the base branch too — I left them alone rather than widen this.

## The other gate is also red on base, and I have not touched it

Separate from this, `No new ESLint warnings` fails for every PR as well, from #12950:

```
✗ tests/unit/volcengine-plan-binding-upsert.test.ts:70  error @typescript-eslint/no-explicit-any
✗ tests/unit/volcengine-plan-binding-upsert.test.ts:96  error @typescript-eslint/no-explicit-any
✗ tests/unit/volcengine-plan-binding-upsert.test.ts:120 error @typescript-eslint/no-explicit-any
```

Three `(bindingTesting as any).upsertConnection(...)` casts. That's @Bob-Hou's file and typing them properly means deciding what `upsertConnection`'s test-facing signature should be, which I'd rather not guess at inside someone else's test. Flagging it here since it has the same "blocks everyone" character — happy to send a patch if you'd like one, or it's a quick fix for whoever owns that code.


abhisheksharma2411:
Update on the red checks here, since they moved and one of them is now mine to report rather than mine to fix.

**`Docs Gates` is green on this PR** — #13135 fixes the 171→172 migration drift. Confirmed on the run above.

**All four `Unit Tests fast-path` shards are now red, and that is repo-wide, not this branch.** Same four shards fail on #13136, #13141 and #13143, and `Release-Green (continuous)` failed on `release/v3.8.51` itself at 18:33. One test:

```
✖ POST /v1/chat/completions with a HuggingFace image model returns 400 + generations hint (#6457)
  AssertionError: The input did not match /\/v1\/images\/generations/. Input:
  "Model '…' is an image-generation model and cannot be used on <path>"
```

I chased it to `redactErrorPaths`, which fails closed over the rest of the message once it treats `/v1/chat/completions` as an unequivocal filesystem path — so the "Use POST /v1/images/generations instead" hint is destroyed for real callers, not just for the assertion. Written up with the reproduction in #13144; I did not patch it because that module is deliberately fail-closed and the boundary is the author's call.

Nothing on this branch touches any of it — the diff here is 44 markdown/text files.

abhisheksharma2411:
Closing — this is stale in the direction that matters: merging it now would make the docs **wrong**.

When I opened it the docs said 172 and the tree had a different count. On the current base (`c0f92ec9`) both agree, at a number this PR would undo:

```
$ grep migrations llm.txt
- **Database:** SQLite via better-sqlite3 (local, zero-config, 175 migrations)
│   │   │   └── migrations/       # 175 versioned SQL migration files

$ find . -name '*.sql' -path '*migrations*' | wc -l
175
```

So the Docs Gate this was meant to unblock is green on that claim now, and re-pointing it at 172 would re-break it. Not a conflict worth resolving — the fix landed by other means.

PR #13142, opened 2026-09-09
fix(opencode): require an http(s) baseURL in both OpenCode plugins

maxmad64bis:
⚠️ base-red inherited: #12732

## Summary

Type your gateway address without `http://` and the plugin accepts it, then nothing works.

`z.string().url()` accepts anything `new URL()` parses, and `new URL()` reads `localhost:20128` as the scheme `localhost:` followed by a path. Both OpenCode plugins take it, publish every model with `localhost:20128/v1` as its api url, and every call then fails inside the client on an unknown scheme. No request is sent, so the gateway logs show nothing and the error names neither the option nor a model.

Only `host:port` slips through — `or.example.com/v1` is already rejected — which is the local address almost every install starts with.

Both plugins now trim the value and require an http(s) scheme, the same treatment `headroomUrl` already gets in `src/shared/validation/settingsSchemas.ts`, where the comment spells out why `z.string().url()` is not enough on an address that ends up in `fetch`. Inside the v2 plugin a single `isHttpUrl` backs the option schema, the publish boundary (`legacyApiToInfoApi`) and the snapshot filter (`isStaleSnapshotModel`), which previously accepted a card with a blank or relative url — those three cannot drift apart. The v1 plugin carries its own copy of the predicate, since the two packages ship independently and neither depends on the other.

## Related Issues

- No open issue tracks this.
- Related to #12732 — the base is red independently of this change.

## Validation

- [ ] `npm run lint`

## Tests Added Or Updated

- `@omniroute/opencode-plugin/tests/options-schema.test.ts` and `@omniroute/opencode-plugin-v2/tests/options.test.ts` — scheme-less, `ftp://` and bare-host addresses rejected with the new message; http(s) accepted, padding trimmed.
- `@omniroute/opencode-plugin-v2/tests/snapshot-stale-entries.test.ts` — a snapshot entry with a package but no url is dropped and counted in the warning; blank, relative and non-http urls rejected at the publish boundary.
- `docs/guides/OPENCODE-V2-PLUGIN.md` — the `baseURL` row now states the http(s) requirement.

## Coverage Notes

- Touches `src/index.ts` in `@omniroute/opencode-plugin`, and `src/options.ts`, `src/catalog.ts`, `src/cache.ts`, `src/shared/models-map.ts` in `@omniroute/opencode-plugin-v2`. Neutering `isHttpUrl` turns three v2 cases red, one per site; removing either plugin's schema guard turns its own case red.

## Reviewer Notes

- This is a tightening: a configuration accepted today stops being accepted. It never worked — it failed on every call with an error that named nothing. A clear message at startup replaces a catalog of unusable models.
- The three red checks on this head are inherited, not from this branch:
  - `No new ESLint warnings` — #12732 states ratchet drift (eslint warnings, cognitive complexity, file size) is expected mid-cycle, rebaselined at release, and not a contributor concern. ESLint is clean on the five files this PR changes.
  - `Merge integrity (changelog + generated skills)` — #12732 lists `check:agent-skills-sync` among the base failures.
  - `Docs Gates (fast-path)` — `check:docs-counts` reports three stale migration counts, in `README.md`, `AGENTS.md` and `llm.txt`. This PR touches none of them; the only doc it changes is the plugin guide.
- `npm run lint` is left unchecked for the same reason: the repo gate is red on this base for three `no-explicit-any` errors in `tests/unit/volcengine-plan-binding-upsert.test.ts`, untouched here.
- Run each package suite with its own `npm test`: v2 is 222/222; v1 is 368/369, where `scaffold.test.ts` imports a built `dist/` the tree does not carry and fails the same way on an untouched checkout.

### DIFF
--- a/@omniroute/opencode-plugin-v2/src/cache.ts
+++ b/@omniroute/opencode-plugin-v2/src/cache.ts
@@ -10,6 +10,7 @@ import type {
   OmniRouteRawCombo,
   OmniRouteRawModelEntry,
 } from "./shared/index.js";
+import { isHttpUrl } from "./shared/index.js";
 
 export const DEFAULT_MODEL_CACHE_TTL_MS = 300_000 as const;
 
@@ -34,8 +35,9 @@ export const SNAPSHOT_FORMAT_VERSION = 2 as const;
 
 /**
  * A raw snapshot entry is stale when it cannot be mapped to a publishable
- * model: no string `id` (unroutable) or a pre-mapped `api` block without a
- * valid `npm` package (the runner would reject it as `Unsupported package`).
+ * model: no string `id` (unroutable), or a pre-mapped `api` block missing a
+ * valid `npm` package (the runner would reject it as `Unsupported package`)
+ * or a usable `url` (the host would reach the AI SDK with no baseURL).
  * Plain `/v1/models` entries carry no `api` block -- it is synthesized at
  * publish time -- so only a present-but-invalid block drops the entry.
  */
@@ -47,7 +49,11 @@ export function isStaleSnapshotModel(entry: unknown): boolean {
   if (api === undefined) return false;
   if (!api || typeof api !== "object") return true;
   const npm = (api as { npm?: unknown }).npm;
-  return typeof npm !== "string" || npm.length === 0;
+  if (typeof npm !== "string" || npm.length === 0) return true;
+  // Same requirement as `npm`, and the same predicate the options schema
+  // applies to `baseURL`: a pre-mapped block without a callable `url` publishes
+  // a model the host cannot route -- see `legacyApiToInfoApi`.
+  return !isHttpUrl((api as { url?: unknown }).url);
 }
 
 interface DiskSnapshotV2 {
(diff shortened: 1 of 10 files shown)


maxmad64bis:
Closing this myself: on a second look there is no live path behind it, so it is not worth a reviewer's time on a base that is already red.

The claim I opened with — that a persisted snapshot puts a card without `url` in front of the host — does not hold. `OmniRouteRawModelEntry` declares no `api` field, and the entries the gateway serves on `/v1/models` carry none, so the block the stale filter judges has no producer. On the publish side the mapper always synthesizes a URL, and the options schema rules out the only input that would empty it. The guard would never fire.

That leaves symmetry with the existing `npm` refusal as the whole argument, and the `npm` half guards a shape nothing emits either. Not a good enough reason to ask for a review.

The failure that sent me here is real, but it is host-side: when a plugin fails to load, the host keeps serving a previously persisted catalog, and a model card in it without a routable URL reaches the AI SDK with no base URL and dies on a bare `Invalid URL` — no request on the wire, nothing in the gateway logs, no model named. That belongs upstream of this package, not in it.

Branch kept if anyone wants the diff later.

PR #13143, opened 2026-09-09
feat(sse): retry transient 5xx backend errors with jitter (global-fallback call site) (#12695)

KooshaPari:
## Summary

Fixes #12695. Resubmission of `pr/retry-503-jitter` (closed for base drift + over-broad scope per Diego's review feedback).

A single network blip in the proxy layer (`api.minimax.io/anthropic` in our config) was making every global-fallback request fail, even though a single retry would almost always succeed. Same root cause as the user's reported "Subagent dispatch failing with 503 (OmniRoute backend down)" symptom.

## Fix

New helper `open-sse/services/transientBackendRetry.ts`:
- Recognizes 502 / 503 / 504 / 408 / 429 retryable (configurable)
- Decorrelated full-jitter backoff (per AWS retry guidance): `random(baseMs, prev * 3)` capped at `capMs`
- AbortSignal-aware: client disconnect cancels immediately, no spurious work
- Returns the last error if the budget is exhausted (no silent fail)

Wired into **only** the global-fallback call site in `src/sse/handlers/chat.ts`. The combo-target loop has its own retry semantics already (and was out of scope per Diego's review — "touch ONLY the global-fallback call site").

## What Diego's review feedback fixed

Per his re-submit instructions:

| Issue | Original | This PR |
|---|---|---|
| Test location | `open-sse/__tests__/transientBackendRetry.test.ts` | `tests/unit/12695-transient-backend-retry.test.ts` |
| 429 in retry set | included | **dropped** (429 must respect `RateLimit-Reset`, surface to caller) |
| Scope | import rewrite + retry-helper added + 3 call sites | **only** the helper, the test, and the single global-fallback call site |

## Behavior

| Scenario | Before | After |
|---|---|---|
| Single 503 from upstream | Hard fail, error to caller | Retry 2× with jitter, then succeed or surface the final 503 |
| Continuous 503 | Hard fail | 3 attempts (≤7s), then surface final 503 |
| Client disconnect mid-retry | Hard fail | Cancels mid-retry, no spurious work |
| 4xx (400/401/422) | Surface immediately | Surface immediately (unchanged) |
| 429 | Would silently retry — wrong | Surface immediately (RateLimit-Reset gives the correct retry hint) |

## Files

| File | Status | LOC |
|---|---|---|
| `open-sse/services/transientBackendRetry.ts` | new | ~110 |
| `tests/unit/12695-transient-backend-retry.test.ts` | new | 211 (7 tests) |
| `src/sse/handlers/chat.ts` | modified | +5 lines (import + wrap call) |

## Tests

```sh
node --test tests/unit/12695-transient-backend-retry.test.ts
```

7 tests cover:
- retry succeeds within budget
- exhausts budget and surfaces last error
- 502 / 503 / 504 / 408 / 429 classified retryable
- 400 / 401 / 422 / 410 / non-transient bubble through immediately
- abort signal cancels mid-flight
- jitter stays within bounds
- total wall-clock ≤ `budget * cap + epsilon`

Fixes #12695


diegosouzapw:
Thanks for #12695. Holding for now.

The branch carries ~2.5k unrelated commits, so it can't be merged as-is. On the change itself: retrying the global fallback on 502/503/504 never consumes or cancels the discarded `Response` bodies, which leaks upstream streams on each retry. It can also add up to ~7s before the combo reports failure.

The call site passes `source`, which the first commit's options type doesn't declare. A rebased, single-purpose branch that cancels discarded bodies (`await res.body?.cancel()`) and tests the `handleChat` global-fallback path would be much easier to evaluate.

diegosouzapw:
Thanks for iterating on #12695 — the retry helper itself (decorrelated jitter, AbortSignal-aware,
429 excluded) is exactly the scope we asked for and looks solid. The blocker is the branch, not
the code: it's carrying ~3k commits from what looks like your personal fork's history (old
releases, Tauri-desktop refactors), which also touches `.github/copilot-instructions.md`
(protected instruction file) with an unrelated Tauri-related edit. Could you cherry-pick just
your two commits (`8d4e10ae` the feature, `d3da48a2` the review fix) onto a fresh branch off the
current release tip and resubmit? That keeps your code and credit without dragging in the fork
history.

PR #13149, opened 2026-09-10
fix(db): skip integrity scans during health polling

cryptiklemur:
health dashboard polling calls `runManagedDbHealthCheck()` every 15 seconds. that path runs SQLite `quick_check` synchronously, blocking the request-serving event loop while it scans the database.

this change skips integrity scans for authenticated `GET /api/db/health` requests. the shared wrapper also honors the existing `OMNIROUTE_SKIP_DB_HEALTHCHECK=1` setting. explicit repair requests retain integrity checks when that setting is off.

the other reference and state checks remain enabled. this does not make every database diagnostic asynchronous or claim to fix all dashboard stalls.

tested with `node --import tsx/esm --test tests/unit/db-health-poll-nonblocking.test.mjs tests/unit/db-health-route.test.ts tests/unit/db-health-check.test.ts tests/unit/db-health-driver.test.ts tests/unit/db-core-extended.test.ts`.

the regression test covers repeated polling, the environment override, explicit repair, and unauthenticated requests.

base-red inherited: #12732. no deployment settings or user data are included.


diegosouzapw:
Clean, well-scoped fix — skipping the synchronous integrity scan on the 15s
dashboard poll while keeping it for explicit repair and startup is the right
tradeoff, and reusing `healthCheck.ts`'s existing `skipIntegrityCheck` option
instead of inventing something new is a nice touch. Ran your test file at this
PR's head: 4/4 pass. Only ask before merge: could you add a `changelog.d/fixes/`
fragment (repo convention, missing from this diff)?

cryptiklemur:
Added

PR #13150, opened 2026-09-10
feat(sse): allow disabling conversation tracking

cryptiklemur:
adds `OMNIROUTE_DISABLE_CONVERSATION_TRACKING=1` for operators who do not need the conversation-history dashboard.

when set, `resolveConversationId()` returns an untracked result before reading SQLite or parsing message history. the switch also covers client-supplied session IDs. existing routing-session handling stays unchanged, and tracking remains enabled by default.

on the reporting installation, 5.97 million turn records and their indexes occupied about 4.26 GB. operators should be able to stop collecting this optional history without disabling normal request routing.

tested with `node --import tsx/esm --test tests/unit/conversationTracker.test.ts tests/unit/conversationTracker-reconnect-7847.test.ts`.

the new regression test rejects both database access and message parsing while disabled. existing enabled-tracking and reconnect tests still pass.

this switch does not delete existing records. the installation-specific cleanup and deployment are excluded.

base-red inherited: #12732.


diegosouzapw:
Clean, small, well-tested fix for a real operational pain point (5.97M turn rows / ~4.26 GB of
optional history on the reporting install). The early-return before any DB/parsing work is the
right approach, and the `string | null` return-type change is consistent with the existing
null-handling convention at the one call site. Ran `conversationTracker.test.ts` at this PR's
head: 18/18 pass, including the new disabled-tracking case, with no regressions. One thing
missing: `OMNIROUTE_DISABLE_CONVERSATION_TRACKING` isn't documented in `.env.example` or
`docs/reference/ENVIRONMENT.md` yet — could you add an entry there (same style as the
neighboring vars)? Should be release-green otherwise.

cryptiklemur:
Added

PR #13153, opened 2026-09-10
fix(sse): fail over to sibling connection on stream early EOF

maxmad64bis:
⚠️ base-red inherited: #12732

## Summary

When the upstream opens a stream but never emits anything useful (pings only, then close), the client gets a terminal 502 even though nothing was produced — retrying elsewhere is safe since no user-visible output exists yet. The loop retried once on the same connection, then gave up without ever trying a sibling. This PR adds that step: once the bounded retry is spent, an early-EOF stream fails over to a sibling — no account marking (an early close isn't a bad connection), singletons/forced pins/timeouts stay terminal. Per your review, the hop ships OFF behind `STREAM_EARLY_EOF_SIBLING_FAILOVER_ENABLED` (runtime, default `false`, fail-safe wrapper): flag off restores the release path exactly. Downstream of #9708 — same-connection replay first, sibling only after it's spent.

## Related Issues

No new issue: early-EOF is treated as retriable. Out of scope: mid-stream `terminated`, silent stalls, timeouts.

## Validation

- [x] Change type: routing
- [x] `npm run lint`

Failover suite 7/7, flag suites 67/67, `typecheck:core` clean, api-typecheck at baseline (289 pre-existing), eslint clean on touched files, file-size green (chat.ts 2462→2471, documented rebaseline), complexity new-code 0, docs-sync PASS. Realigned on tip `00f16a27` (reset + cherry-pick; base hunks kept, intent reapplied), checks rerun after.

## Tests Added Or Updated

Failover suite (6 subtests: flag-on hop, singleton/forced/timeout terminal, no-marking, flag-off terminal with 2 dispatches + 502 + no-mark). Flag-definition test + counts 55→56 (×2). Suite registered in `stryker.conf.json tap.testFiles`.

## Coverage Notes

Gate (`early-EOF AND not-forced AND flag-on`) keeps the timeout path out. Sibling gets no fresh retry (counter spent) — one hop, then the no-credentials path. Flag reads per request, no restart needed. i18n en+vi, other locales fall back to the definition.

## Reviewer Notes

Exclusion-ban comment rescoped to the retry arm; post-exhaustion is routing, not a health judgment. Rebaseline `_rebaseline_2026_09_12_13153_early_eof_flag_gate_v2`. Mutation `--strict` still lists 5 covering tests from other PRs (pre-existing drift). Full-repo lint and docs-counts are CI-owned/red-on-base.


diegosouzapw:
Answering your explicit question: **crossing from same-connection to a sibling is acceptable, but not unconditionally — we want it behind a feature flag that ships off.**

The reasoning matches yours. An early EOF with no useful chunk emitted is genuinely safe to retry elsewhere, and not marking the account is right: an early close is not evidence of a bad connection, so it should not cost a cooldown or touch quota. What we are not comfortable doing is changing the terminal/failover boundary for every install in a point release. A stream that opens and dies silently can also be a symptom rather than a fluke, and we would rather watch it on a live gateway before it becomes the default.

So: keep the behaviour exactly as you built it, and gate the sibling hop.

The mechanism already exists and is what the PII flags use:

1. Add the definition to `src/shared/constants/featureFlagDefinitions.ts` with `defaultValue: "false"` and `type: "boolean"`. `category: "runtime"` fits; `warningLevel: "info"` is enough since the flag only widens an existing retry. A `descriptionI18nKey` goes with it, same shape as the neighbouring entries.
2. Read it at the decision point with `isFeatureFlagEnabled("<KEY>")` from `src/shared/utils/featureFlags.ts` — that resolves DB override > env > default, so it can be flipped per operator without a rebuild. `src/lib/guardrails/piiMasker.ts:17` is a one-line example of the call shape.
3. With the flag off, the early-EOF path must behave exactly as it does on `release/v3.8.51` today: bounded same-connection retry, then terminal. Worth one test asserting that direction, alongside the ones you already have for the flag on.

Everything else in the PR reads right to us, including the carve-outs — singleton pools, forced connections and slow-but-alive timeouts staying terminal is the correct set, and scoping this to "closed before any useful chunk" while leaving mid-stream `terminated` out is the right seam.

The rest of your batch landed today: #13141, #13142, #13146, #13147 and #12975 are all merged. #12975 needed reconciling since it was stacked on #12941 — details in its merge comment. Thanks for the write-ups on all of them; the measured before/after numbers made the whole batch reviewable as a unit.

maxmad64bis:
Done — the sibling hop now ships OFF behind `STREAM_EARLY_EOF_SIBLING_FAILOVER_ENABLED` (runtime flag, default `false`, no restart needed). With the flag off, the path is exactly the old behavior: one same-connection retry, then terminal 502, no sibling hop, no cooldown.

What changed: flag definition in `featureFlagDefinitions.ts` (info level, with catalog row in `FEATURE_FLAGS.md` and en/vi strings), the gate reads through `isFeatureFlagEnabled` so DB override and env var both work, and a new test pins the flag-off path (2 dispatches on the first connection, terminal 502, both connections unmarked). Full diff is 2 commits / 12 files; gates green locally (failover 7/7, flag suites 67/67, typecheck, eslint, file-size via documented rebaseline).

One heads-up: `chat.ts` grows 2458→2466 (+8, import + condition wiring — the flag read lives in `chatHelpers.ts`). Rebaseline entry included; happy to shrink it another way if you prefer.

edit:
Follow-up pushed: the flag wrapper now has a fail-safe defaulting to disabled (same shape as the neighboring wrappers), so a DB failure on the error path keeps the terminal 502 instead of throwing. The file-size rebaseline note is also rewritten to describe the actual diff (guarded exclude-and-continue block, not a lengthened condition). Nothing else changed — flag-off path, tests, and gates as before.

maxmad64bis:
Follow-up pushed: the flag wrapper now has a fail-safe defaulting to disabled (same shape as the neighboring wrappers), so a DB failure on the error path keeps the terminal 502 instead of throwing. The file-size rebaseline note is also rewritten to describe the actual diff (guarded exclude-and-continue block, not a lengthened condition). Nothing else changed — flag-off path, tests, and gates as before.

maxmad64bis:
Rebased on the current tip (`00f16a27`) — conflicts were in the file-size baseline (merged batch moved `chat.ts` 2458→2462 and `auth.ts` 3488→3542), the cooldown i18n strings, and the `fallbackAttempts` plumbing in chat. Base hunks kept, PR intent reapplied; rebaseline is now 2462→2471. All focused checks rerun green (failover 7/7, flags 67/67, typecheck, eslint, file-size). Mergeable again.

PR #13162, opened 2026-09-10
fix(deps): declare remark-gfm as a direct dependency

marioschoenert-code:
MarkdownMessage.tsx and other client components (traffic-inspector chat, RequestLoggerDetail) import remark-gfm directly, but it was only present transitively via fumadocs-mdx (a devDependency). npm's flat node_modules hoisting happens to resolve it — masking the issue — but pnpm's strict node_modules isolation fails with "Module not found: Can't resolve remark-gfm" since the package was never declared as a direct dependency of the app.

This PR declares it explicitly in package.json/package-lock.json and adds it to the supply-chain dependency allowlist.

Validation: no automated test reproduces this under npm (hoisting masks it there); reproduced manually via a clean pnpm install + npm run dev, which fails to compile MarkdownMessage.tsx before this fix and succeeds after. npm run check:deps, npm run lint (pre-existing errors only, unrelated file), and npm run check:any-budget:t11 all pass on this branch.

⚠️ base-red inherited: #12732

### DIFF
--- a/config/quality/dependency-allowlist.json
+++ b/config/quality/dependency-allowlist.json
@@ -130,6 +130,7 @@
     "react-markdown",
     "react-reconciler",
     "recharts",
+    "remark-gfm",
     "safe-regex",
     "selfsigned",
     "sharp",
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,6 +77,7 @@
         "react-reconciler": "^0.33.0",
         "react18-json-view": "0.2.10",
         "recharts": "^3.8.1",
+        "remark-gfm": "^4.0.1",
         "safe-regex": "^2.1.1",
         "selfsigned": "^5.5.0",
         "sharp": "^0.35.4",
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Verified: `remark-gfm` is indeed imported directly in `MarkdownMessage.tsx` and was missing from
`package.json`, resolved only transitively via `fumadocs-mdx` under npm's flat hoisting. The
version you declared (`^4.0.1`) matches what's actually resolved in the lockfile today. Clean,
minimal, correct fix — merging as-is. Thanks for tracing this down to the pnpm-vs-npm hoisting
difference and documenting the manual repro.

diegosouzapw:
Thanks @marioschoenert-code — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #13161, opened 2026-09-10
fix(resilience): treat a Cloudflare managed challenge as a fingerprint rejection, not a ban (#13157)

anhtran-ai:
Closes #13157

## Problem

A Cloudflare **managed challenge** served in front of an upstream provider is classified `FORBIDDEN` and persisted as the terminal `banned` connection state, taking the whole provider offline until an operator reconnects in the dashboard.

Observed on `POST chatgpt.com/backend-api/codex/responses/input_tokens` for a **healthy** Codex OAuth account:

```
15:14:07  Trying model 2/2: cx/gpt-5.6-sol
15:14:10  Model cx/gpt-5.6-sol succeeded (5986ms, 0 fallbacks)      ← normal chat OK
15:14:38  POST /v1/responses/input_tokens | …
15:14:39  Successfully refreshed Codex token / CODEX | refreshed    ← credential healthy
15:14:39  [provider] Node e94ce3e6-… banned (403) — disabling permanently
15:14:39  codex | 1 accounts found but none active
```

The OAuth token refreshing successfully in the same second, and normal `/responses` traffic succeeding seconds before and after on the same connection, show the account was never banned upstream.

## Root cause

`isCloudflareFingerprintRejection()` only recognises the Cloudflare **1010** family (`error_code:1010`, `browser_signature_banned`, `fingerprint_rejection`). A managed challenge is the same class of block — the edge refused the *client's* signature, not the credential — but it is a different product surface and carries none of those markers, so the guard at the top of the `403` branch misses it:

```ts
if (statusCode === 403 && isCloudflareFingerprintRejection(bodyStr)) {
  return PROVIDER_ERROR_TYPES.FINGERPRINT_REJECTION;   // ← not reached
}
```

Execution then falls through the rest of the ladder — not geo-blocked, not `accountDeactivated`, not a Cloud Code project error, not `SENTINEL_BLOCKED`/`Turnstile required`, `codex` is neither an `apikey`-category provider nor `authType: "none"` — and lands on `return PROVIDER_ERROR_TYPES.FORBIDDEN`, which `chatCore` persists through `writeTerminalStatus()`.

Measured response, captured against the live account:

```
status 403
cf-mitigated: challenge
server: cloudflare
content-type: text/html; charset=UTF-8

_cf_chl_opt                                  present
cdn-cgi/challenge-platform                   present
id="challenge-error-text"                    present
cType: 'managed'   cZone: 'chatgpt.com'

error code: 1010                             absent
error_code                                   absent
[...]

Issue #13157, opened 2026-09-10
fix(resilience): errorClassifier: a Cloudflare managed challenge on Codex /responses/input_tokens permanently bans a healthy Codex OAuth connection


anhtran-ai:
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Node.js Version

Node 22 (bundled image)

### Provider(s) Involved

Codex (OAuth / ChatGPT subscription)

### Model(s) Involved

`cx/gpt-5.6-sol`

### Client Tool

Codex CLI (via an OmniRoute combo)

## Summary

A single Cloudflare **managed challenge** served by OpenAI's own CDN on `POST https://chatgpt.com/backend-api/codex/responses/input_tokens` is classified as `FORBIDDEN` and persisted as the terminal `banned` connection state. The Codex OAuth account is **not** banned by OpenAI — normal `/v1/responses` chat traffic on the very same connection and the very same token succeeds seconds before and seconds after the event.

From that moment every request is answered locally without ever contacting OpenAI:

```
[provider] Node e94ce3e6-… banned (403) — disabling permanently
codex | 1 accounts found but none active
[codex] All 1 connection(s) banned by upstream — please reconnect in the dashboard   (HTTP 401)
```

The state never auto-recovers; an operator has to reconnect (or re-test + PATCH) in the dashboard.

This is the Codex/Cloudflare-challenge sibling of #12859 (Anthropic OAuth `403 "Request not allowed"`): same terminal-write, different upstream body.

## Root cause

`open-sse/services/errorClassifier.ts` only recognises a Cloudflare *fingerprint* rejection by the **1010 error-code family**:

```ts
const CLOUDFLARE_1010_REGEX =
  /(?<![A-Za-z0-9_-])error[\s_-]?code[\\"':=\s]{0,12}1010(?!\w)|(?<![A-Za-z0-9_-])error[-_]\s?1010(?!\w)\/?/i;

export function isCloudflareFingerprintRejection(errorText: string): boolean {
  const text = String(errorText || "").toLowerCase();
  return (
    CLOUDFLARE_1010_REGEX.test(text) ||
    text.includes("browser_signature_banned") ||
    text.includes("fingerprint_rejection")
  );
}
```

A **managed challenge** is a different Cloudflare product surface. It carries **none** of those three markers, so the guard at line 327 misses it:

```ts
[...]

PR #13167, opened 2026-09-10
fix(responses): count input tokens locally for Codex OAuth

anhtran-ai:
## Summary

Add a static `POST /v1/responses/input_tokens` route that counts locally and returns the standard OpenAI Responses contract without issuing an upstream request.

This complements #13161 / #13157. #13161 makes a Cloudflare managed challenge non-terminal; this PR removes the request that triggers that challenge in the first place.

## Problem

The generic `/v1/responses/[...path]` handler currently forwards `input_tokens` as a native Responses suffix. For a Codex OAuth / ChatGPT-subscription connection that produces:

```
POST chatgpt.com/backend-api/codex/responses/input_tokens
```

Measured behavior for the same live account and token:

- native/undici transport: Cloudflare managed challenge, HTTP `403`, `cf-mitigated: challenge`;
- bundled Chrome 124 `wreq-js` transport: HTTP `404 {"detail":"Not Found"}`;
- normal `/backend-api/codex/responses` inference: succeeds.

The subscription backend does not serve this OpenAI API subresource. Forwarding the preflight can therefore never produce a useful token count. It adds latency and fallback on every call and, before #13161, the managed-challenge 403 was persisted as terminal `banned` / `isActive:false`, taking the healthy Codex provider offline.

## Fix

A static App Router segment shadows the catch-all:

```
src/app/api/v1/responses/input_tokens/route.ts
```

It:

- returns the exact public contract `{ "object": "response.input_tokens", "input_tokens": <int> }`;
- uses the existing offline `countTextTokens()` path (`o200k_base` for Codex / `cx` models);
- counts instructions, string/structured input, tool call/output payloads, tool definitions, `tool_choice`, text and reasoning configuration;
- keeps image/file parts non-textual instead of tokenizing base64 payloads;
- preserves API-key/model policy and process-wide body admission from the Responses route it shadows;
- never calls `fetch`, so the Cloudflare challenge cannot occur on this path.

The estimate is intentionally conservative. Live A/B showed a stable 9–10 token Responses envelope overhead for no-tool requests, while tool definitions already carry their own framing. The route adds that fixed overhead for no-tool requests and keeps a 5% margin for tool requests. Over-counting means slightly earlier client compaction; under-counting risks a context-window overflow.

### Local

- `node --import tsx/esm --test tests/unit/responses-input-tokens-local-route.test.ts`: **9/9 pass**.
- ESLint on both changed files: clean.

Tests pin the exact two-field response schema, prove `fetch` is never called, and cover instructions, structured input, tools, non-text parts, server-held context IDs, invalid JSON, OPTIONS and the conservative lower bound.

### Production A/B

Deployed to the affected gateway as `llm-gateway/omniroute:3.8.51-113ddf9a-azox1` and compared the local result to the subsequent real Codex `usage.input_tokens` for eight request shapes:

| Case | Local | Upstream usage | Difference |
|---|---:|---:|---:|
| short | 16 | 16 | 0% |
| medium | 89 | 89 | 0% |
| long | 817 | 817 | 0% |
| structured messages | 46 | 41 | +12.2% |
| one tool | 66 | 65 | +1.5% |
| two tools | 98 | 75 | +30.7% |
| Vietnamese/Unicode | 33 | 33 | 0% |
| code | 534 | 534 | 0% |

Results: **8/8 valid**, **0 under-counts**, median absolute error **0%**. Both over-counts are in the safe direction (earlier compaction).

Additional production evidence:

- 50/50 count soak HTTP 200.
- Isolated 100/100 count soak HTTP 200.
- The isolated 100-count soak changed `proxy_logs` by exactly **0 rows**.
- Cloudflare challenge events: **0**.
- Permanent-disable events: **0**.
- The Codex connection remained `is_active=1`, `test_status=active` with clear error fields.
- All eight paired inference requests returned HTTP 200; a separate chat soak returned 6/6 HTTP 200.

### DIFF
--- a/src/app/api/v1/responses/input_tokens/route.ts
+++ b/src/app/api/v1/responses/input_tokens/route.ts
@@ -1,0 +1,215 @@
+import { CORS_HEADERS } from "@/shared/utils/cors";
+import { enforceApiKeyPolicy } from "@/shared/utils/apiKeyPolicy";
+import { isRequireApiKeyEnabled } from "@/shared/utils/featureFlags";
+import { withChatAdmission } from "@/shared/middleware/withChatAdmission";
+import {
+  countTextTokens,
+  tokenizerContextFromBody,
+  type TokenizerContext,
+} from "@/shared/utils/tiktokenCounter";
+import { extractApiKey, isValidApiKey } from "@/sse/services/auth";
+import { errorResponse } from "@omniroute/open-sse/utils/error.ts";
+import { HTTP_STATUS } from "@omniroute/open-sse/config/constants.ts";
+
+/**
+ * POST /v1/responses/input_tokens — local Responses token count.
+ *
+ * This static segment deliberately shadows the `[...path]` passthrough route.
+ * Forwarding this preflight upstream is never useful and is actively harmful:
+ *
+ *   - For Codex OAuth the upstream subpath is not served for the account
+ *     (`404 {"detail":"Not Found"}` once the request gets through), so the
+ *     round trip can only ever fail.
+ *   - The same request reaches OpenAI's Cloudflare edge, which answers a
+ *     managed challenge (`cf-mitigated: challenge`, HTTP 403). Before the
+ *     errorClassifier fix that 403 was terminalized into `banned` /
+ *     `isActive:false` and took the whole provider offline; even with the fix
+ *     it still burns a request, wastes the latency and forces a combo
+ *     fallback on every single preflight.
+ *
+ * Counting locally removes the upstream call entirely, so no challenge can be
+ * triggered from this path. The shape mirrors the OpenAI Responses contract
+ * (`object: "response.input_tokens"` plus an `input_tokens` integer), and the
+ * counter is the same offline tokenizer `/v1/messages/count_tokens` already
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Great root-cause fix — shadowing the catch-all instead of trying to patch around the Cloudflare
challenge is the right call, since the upstream subresource can never actually succeed for OAuth
accounts. I ran your test file at the PR head: 9/9 pass, including the "never calls fetch"
assertion which is really the core guarantee here. One ask before merge: could you add a
`changelog.d/fixes/` fragment? I didn't find one in the diff.

diegosouzapw:
Thanks @anhtran-ai — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #13168, opened 2026-09-10
fix(opencode): preserve catalog display names

domenicomassafra:
OpenCode setup currently uses every technical model ID as its visible name, so auto/chat remains auto/chat and provider-prefixed catalog labels duplicate the provider. This keeps technical IDs unchanged while preferring an existing custom name, then native display_name/name, strips the owned_by prefix once, and renders auto/* names readably. Smoke: Auto Chat / DeepSeek V4 Pro.


diegosouzapw:
Nice, focused fix — I checked the current tip and the `owned_by`-prefix dedupe you mention is
indeed already there, but the custom-name/`display_name` preference isn't, so this fills a real
gap. Main ask before merge: an automated unit test for the four name-precedence branches you
describe (custom name, native `display_name`/`name`, stripped `owned_by` prefix, readable
`auto/*` fallback) — the PR currently only has a manual smoke check. Also please rebase onto
`release/v3.8.51`, your branch is still off `main`.

domenicomassafra:
Rebased onto current `release/v3.8.51` and verified automated unit tests in `tests/unit/opencode-catalog-display-names-13168.test.ts` covering all name precedence branches:
1. Existing custom name override preference
2. Catalog native `display_name` with `owned_by/` prefix stripped once
3. Catalog native `name` fallback when `display_name` is absent
4. Readable `Auto <Name>` label fallback for `auto/*` selectors
5. Raw catalog model ID fallback when no metadata applies

Head commit is `a094de4d299ce79d6ea8fcd206ee6ec5121967e7`. All unit tests pass cleanly.

diegosouzapw:
Thanks @domenicomassafra — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #13173, opened 2026-09-10
fix(sse): preserve Fable mid-conversation cache prefixes

cryptiklemur:
Fable 5 and 5.1 support mid-conversation system messages, but OmniRoute treats only Opus as capable. That hoists each Fable system turn into the top-level system prompt and shifts the prompt-cache prefix.

## what changed

- give Fable its own mid-conversation-system capability path
- keep context-1m limited to Opus
- cover the native Claude Code request path for Opus 5, Fable 5, and Fable 5.1
- confirm Sonnet stays outside the mid-conversation-system path

The capability split matters because sharing the context-1m allowlist would send Fable an unrelated beta.

## how to test

```sh
node --import tsx/esm --test tests/unit/executor-claude-identity.test.ts
node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit --test-isolation=none --test-name-pattern="mid-conversation system cache breakpoints" "$PWD/tests/unit/chatcore-translation-paths.test.ts"
npm run typecheck:core
```

All three checks pass.

A controlled three-turn Fable 5.1 request reproduced `system_changed` on turns 2 and 3 before this patch. The test uses synthetic text, a stable system prompt and tools, and growing mid-conversation system notes.

Anthropic documents prompt-cache invalidation when earlier cached content changes: https://platform.claude.com/docs/en/build-with-claude/prompt-caching

### DIFF
--- a/open-sse/executors/claudeIdentity.ts
+++ b/open-sse/executors/claudeIdentity.ts
@@ -309,6 +309,7 @@ const HEAVY_AGENT_BETA_MODEL_PREFIXES = ["claude-opus", "claude-sonnet"];
  */
 const CONTEXT_1M_BETA_MODEL_PREFIXES = ["claude-opus"];
 const CONTEXT_1M_NATIVE_MODEL_PREFIXES = ["claude-opus-5"];
+const MID_CONVERSATION_SYSTEM_MODEL_PREFIXES = ["claude-opus", "claude-fable"];
 
 function matchesModelPrefix(model: unknown, prefixes: string[]): boolean {
   if (typeof model !== "string") return false;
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Clean fix and the reasoning about not sharing the context-1m allowlist with
mid-conversation-system is sound. Ran `executor-claude-identity.test.ts` green (32/32)
against the PR head — that's the file exercising `selectBetaFlags` directly. Couldn't get a
clean run of `chatcore-translation-paths.test.ts` on my probe (hit an infra timeout
unrelated to your change); worth a quick re-check before merge. Please add a
`changelog.d/fixes/` fragment.

cryptiklemur:
Added

PR #13160, opened 2026-09-10
docs: correct the migration count from 171 to 172 (unblocks Docs Gates)

anhtahaylove:
`check:docs-counts` compares documented counts against the real file count in `src/lib/db/migrations`, which currently holds **172** `.sql` files. Three docs still say 171, which the checker treats a...[truncated]

PR #13158, opened 2026-09-10
docs(changelog): format the merged 13094 fragment as a markdown bullet

anhtahaylove:
The changelog fragment added by #13097 shipped as a bare paragraph instead of a markdown bullet.

`scripts/check/check-changelog-integrity.mjs` requires every fragment to start with `- `, so this fails the **Merge integrity (changelog + generated skills)** gate on *every* branch built off `release/v3.8.51` — not only the PR that introduced it.

```
✗ changelog.d/fixes/13094-deepseek-pow-slot-leak.md: fragment must start with a markdown bullet ("- ")
```

Reformatted to match the surrounding fragments (bolded scope prefix + PR link). The gate reports 0 violations afterwards. Content is unchanged apart from the bullet and the link.

PR #13175, opened 2026-09-10
fix(telegram): authenticate webhook deliveries with the Telegram secret token

### LINKED ISSUE #13172: fix(auth): Telegram webhook path accepts unauthenticated updates (no secret_token check)
## Summary

`POST /api/telegram/update` has two entry paths, and only one of them authenticates the caller.

- **Mini App path** (`{ initData, message }`) verifies the Telegram HMAC via `verifyInitData(initData, botToken)` before doing any work. This path is fine.
- **Webhook path** (a raw Telegram update body) is gated only by `isTelegramEnabled()`. It takes `chat.chatId` straight from the request body and passes it to `proxyChat()`.

Telegram supports authenticating webhook deliveries with a `secret_token` (sent back as the `X-Telegram-Bot-Api-Secret-Token` header). This project never sets it: `setTelegramWebhook()` in `src/lib/telegram/botApi.ts` calls `setWebhook` with only `url` and `drop_pending_updates`, and no route or helper reads that header. Grepping the whole `src/lib/telegram/` tree for `secret_token` / `X-Telegram-Bot-Api-Secret-Token` returns nothing.

## Impact

Anyone who can reach the deployment can POST a synthetic update with an arbitrary `chat.id` and have it processed as a genuine Telegram message. Each distinct id:

1. resolves through `resolveUserApiKey()`, which **mints a real OmniRoute API key** (`createApiKey("telegram:<id>")`) when none matches, and
2. runs a chat completion through the normal pipeline, consuming upstream quota.

So an unauthenticated caller can drive both API-key creation and paid model usage. The reply is delivered to the spoofed chat id via `sendTelegramMessage`, so the attacker does not receive the completion text directly — this is resource abuse and unbounded key creation rather than data exfiltration.

The memory-growth side of the same code path is tracked separately in #13165 (bounded cache) and fixed in #13166. **That fix caps the in-memory cache; it does not authenticate the caller, and minted keys still persist in the database.** The two issues need separate fixes.

## Suggested fix

1. Generate a webhook secret, pass it as `secret_token` in `setWebhook`, and store it alongside `TELEGRAM_BOT_TOKEN`.
2. In the webhook branch of `POST /api/telegram/update`, compare `X-Telegram-Bot-Api-Secret-Token` against that secret using a constant-time comparison, and reject with 401 on mismatch. Keep the existing `initData` HMAC branch unchanged.
3. Consider whether an unknown chat id should be allowed to mint a key at all, or should require an explicit enrolment step.

Point 2 is the security boundary; point 3 limits blast radius if the secret ever leaks.

## Notes

Found while auditing the cache/registry layer for retention bugs (#13165). Filed separately because the authentication gap is a security concern in its own right and does not share a fix with the cache bound.

anhtahaylove:
Fixes #13172.

## Problem

`POST /api/telegram/update` serves two callers, and only one of them was authenticated:

- **Mini App** (`{ initData, message }`) — verifies the Telegram `initData` HMAC. Safe.
- **Bot webhook** (a Telegram update) — took `message.chat.id` straight from the body. The only gate was `isTelegramEnabled()`, which merely checks that a bot token is configured on *our* side. Nothing proved the request came from Telegram.

`chat.id` flows into `proxyChat()`, which **mints a real API key** for that id and spends upstream model quota. A caller who knows the URL can therefore create unbounded keys and burn quota. They don't get the replies back (those go to the spoofed chat id), so this is resource abuse and key-table growth, not data exfiltration.

`setWebhook` also never registered a `secret_token`, so the header needed for verification was never being sent by Telegram in the first place.

## Fix

Use Telegram's own mechanism: the `secret_token` given to `setWebhook` is echoed back on every delivery as `X-Telegram-Bot-Api-Secret-Token`.

1. `TELEGRAM_WEBHOOK_SECRET` config getter (`src/lib/telegram/config.ts`).
2. `setWebhook` registers `secret_token` when configured (`src/lib/telegram/botApi.ts`).
3. The webhook branch verifies the header before anything reaches `proxyChat()`:
   - secret not configured → **503** (fail closed — an open webhook is what causes the abuse)
   - missing/wrong secret → **401**
   - correct → proceeds as before

Comparison uses `timingSafeEqual` with an up-front length check, mirroring `webhookSecretMatches`'s counterpart in `src/app/api/a2a/tasks/route.ts`, so the shared-prefix length doesn't leak through response timing.

The Mini App path is untouched and does **not** require the secret.

### Incidental schema fix

The zod body schema typed `message` as an optional **string** (the Mini App shape), but a real Telegram update sends `message` as an **object**. `.passthrough()` preserves unknown keys but still type-checks known ones, so realistic webhook deliveries were rejected with **400** before reaching any routing. `message` now accepts either shape; each branch already narrows the shape it needs (`typeof body.message === "string"` for Mini App, `extractChatMessage()` for webhook). This surfaced while writing the test — a body that mirrors a genuine Telegram update.

## Verification

`tests/unit/telegram-webhook-secret-13172.test.ts` — **4/4 pass with the fix, 4/4 fail without it** (verified by stashing the source changes):

- no secret header → 401, nothing proxied
- wrong secret → 401
- correct secret → 200
- comparator: equal values match, length mismatch returns false instead of throwing

`npx tsc --noEmit` clean.

## Operator note — breaking for existing webhook deployments

Anyone already running the webhook must set `TELEGRAM_WEBHOOK_SECRET` and re-run `setWebhook` so Telegram starts sending the header; until then deliveries return 503. This is deliberate: silently serving an unauthenticated webhook is the vulnerability. Documented in `.env.example` and `docs/reference/ENVIRONMENT.md`.

Note `check:docs-counts` reports 3 pre-existing strict drifts (171 → 172 migrations) on this base; unrelated to this change and fixed by #13160.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3070,6 +3070,11 @@ QUOTA_STORE_DRIVER=sqlite
 # Telegram Mini App bridge. The update endpoint remains disabled while the bot
 # token is unset. Used by: src/lib/telegram/* and src/app/api/telegram/update/route.ts.
 # TELEGRAM_BOT_TOKEN=
+# Shared secret registered with setWebhook and echoed back by Telegram as the
+# X-Telegram-Bot-Api-Secret-Token header. REQUIRED for the webhook path: without
+# it the webhook is rejected with 503, because an unauthenticated update lets any
+# caller mint API keys and spend upstream quota. The Mini App path does not use it.
+# TELEGRAM_WEBHOOK_SECRET=
 # TELEGRAM_DEFAULT_MODEL=auto/chat
 # TELEGRAM_BOT_API_BASE=https://api.telegram.org
 # TELEGRAM_WEBHOOK_TIMEOUT_MS=60000
--- a/changelog.d/fixes/13172-telegram-webhook-secret.md
+++ b/changelog.d/fixes/13172-telegram-webhook-secret.md
@@ -1,0 +1,1 @@
+- **fix(telegram):** authenticate webhook deliveries with Telegram's `secret_token` so an unauthenticated caller can no longer mint API keys or spend upstream quota ([#13172](https://github.com/diegosouzapw/OmniRoute/issues/13172))
--- a/docs/reference/ENVIRONMENT.md
+++ b/docs/reference/ENVIRONMENT.md
@@ -1623,6 +1623,7 @@ These settings were introduced after the previous environment-contract snapshot.
 | `ADOBE_FIREFLY_CHROME_HEADLESS` | `0` | `open-sse/services/adobeFireflyBrowserLogin.ts` | Debug-only true-headless mode; Adobe colligo normally rejects the resulting risk session. |
 | `CHROME_PATH` | auto-detect | `open-sse/executors/cloudflare-playground.ts`, `open-sse/executors/chatgpt-web-codex.ts` | Optional absolute Chrome executable used by the browser-driven executors when platform auto-detection is insufficient. |
 | `TELEGRAM_BOT_TOKEN` | _(unset)_ | `src/lib/telegram/config.ts` | BotFather token that enables the inbound webhook and signs Mini App `initData`. |
+| `TELEGRAM_WEBHOOK_SECRET` | _(unset)_ | `src/lib/telegram/config.ts` | Shared secret registered via `setWebhook` and verified against the `X-Telegram-Bot-Api-Secret-Token` header on every webhook delivery. Required for the webhook path; unset means webhook deliveries are refused with 503. |
 | `TELEGRAM_DEFAULT_MODEL` | `auto/chat` | `src/lib/telegram/chatProxy.ts` | Model used for Telegram chat replies. |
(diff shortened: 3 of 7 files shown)


anhtahaylove:
Flagging for review priority: of the 14 open PRs in this audit ([queue overview](https://github.com/diegosouzapw/OmniRoute/pull/13158#issuecomment-5614501785)), this is the only one with a security impact rather than a resource-management one.

An unauthenticated caller could POST a synthetic Telegram update with any `chat.id`; that id reaches `proxyChat()`, which mints a real API key and spends upstream quota. Replies go to the spoofed chat id, so it is resource abuse and unbounded key creation rather than data exfiltration — but it needs no credentials at all.

Note this is intentionally **breaking for existing webhook deployments**: operators must set `TELEGRAM_WEBHOOK_SECRET` and re-run `setWebhook`, otherwise deliveries return 503. Failing closed seemed clearly right over continuing to serve an open endpoint, but say the word if you'd prefer a deprecation window instead.

PR #13177, opened 2026-09-10
[defer] refactor(providers): consolidate agy into a CLI-only antigravity provider

backryun:
## Summary

Consolidates the `agy` (Antigravity CLI) provider into `antigravity` and removes the IDE client identity, so a single provider serves the whole Antigravity / Google Cloud Code family with the official CLI fingerprint.

Both clients authenticate against the same Google consumer-OAuth client and share the same Cloud Code backend, model catalog, and quota — the only real difference was the presented client identity (IDE vs CLI), so the split duplicated surface without adding capability.

⚠️ **Stacked on #12608** — this branch is based on the `codex/refresh-antigravity-agy-models` head (which introduced `googleCloudCodeModelCatalog.ts`). Please review/merge #12608 first; after it lands I will rebase this branch onto `release/v3.8.51` and drop the duplicated commits.

## What changed

- **Provider consolidation**: the `agy` provider id is deprecated and migrates to `antigravity` (runtime `DEPRECATED_PROVIDERS` notice); the `agy/` model prefix stays a working alias. Deleted `registry/agy`, `agyModels.ts`, `AGY_CONFIG`, and the `agy` OAuth provider; the shared `antigravity` registry entry now carries the CLI headers.
- **IDE identity deleted**: `antigravity/ide/...` User-Agent builders, the IDE release feed, `getAntigravityIdeNodeHeaders`, the per-connection client-profile selector, and the IDE profile option. `AntigravityClientProfile` is now `"cli"` only (legacy persisted values normalize to `cli`).
- **DB migration 176**: rewrites stored `provider = 'agy'` rows to `antigravity` across all tables, rewrites `agy/%` model ids, and normalizes the legacy IDE `clientProfile` to `cli`. Existing tokens keep working (same OAuth client — no re-login).
- **Dashboard**: a single "Antigravity" card with the CLI import hint; the `agy-auth/*` import routes now create `antigravity` connections (paths kept for compatibility).
- **Docs / counts**: provider count synced to 355 (README, AGENTS.md, llm.txt, package.json), `PROVIDER_REFERENCE.md` regenerated, SVG canonical numbers updated.
- **Tests**: 46 suites aligned; `agy-provider.test.ts` → `antigravity-provider.test.ts`.

## Validation

- Focused suites (100 files — all antigravity/agy/oauth-adjacent): **548 tests → 547 pass, 1 fail**. The single failure (`Antigravity missing-project 422 stays fail-closed`) reproduces identically on the base tip (pre-existing).
- `check:open-sse-typecheck` 0 errors; `check:dashboard-typecheck` within frozen baseline; ESLint (with suppressions) clean on every changed file.
- `check:provider-consistency` OK (272 registry / 355 canonical / 0 registry-only); `check:file-size` + `check:test-file-size` OK; `check:docs-counts-sync` reports only the pre-existing migration-count drift inherited from base.
- Migration 176 applied cleanly in test runs (172 migrations total).

⚠️ base-red inherited: #12732

## Notes for the maintainer

- This PR touches `AGENTS.md` (provider count 356 → 355) — per the repo's agent-instruction rule, merging it needs explicit operator approval.
- `docs/reference/ENVIRONMENT.md` / `STEALTH_GUIDE.md` and i18n mirrors are intentionally untouched here (the i18n mirrors are regenerated by tooling).


HouMinXi:
Thanks for consolidating `agy` into the Antigravity CLI-only provider. I did not fold this into the local recut of `release/v3.8.51`.

Reason: live combos on this host still pin `agy/` model ids (`onmi-gemini3.6` and siblings). The merge hit `modelSpecs.ts`, `PROVIDER_REFERENCE.md`, and quota-weighted tests; taking the rename would take the serving provider out from under those pins.

Purpose of the recut: ship current `release/v3.8.51` plus open product PRs to a private production host without breaking existing combo pins. Happy to pick this up on a later recut once combos are remapped. No objection to the refactor itself.

diegosouzapw:
This is a genuinely large architectural proposal — folding the `agy` CLI provider into
`antigravity` and deleting the standalone `agy` registry entry entirely. It's worth
noting this goes in the opposite direction from #13410, which just merged today and
deliberately kept `agy` and `antigravity` as separate providers (just de-duplicating
their catalogs via a shared base file, per #12724). Given the size (142 files) and that
it touches `AGENTS.md`/`llm.txt` (doc-count bumps only, but still a protected surface
needing explicit sign-off) and would remove a provider users may already have configured,
I'd like @diegosouzapw to weigh in on the direction before this goes further — do we want
full consolidation, or keep the current shared-catalog approach? Once that's settled this
will also need a full rebase onto today's catalog refactor.

backryun:
**Direction check — full consolidation vs shared catalog**

@diegosouzapw — capturing the trade-off so we can close this one way or the other:

**This PR (full consolidation)** — one `antigravity` provider serves both surfaces; the `agy` id is removed and mapped through `DEPRECATED_PROVIDERS` (model prefixes such as `agy/…` keep resolving as aliases), and the IDE client identity is dropped in favor of the CLI fingerprint. Cost: it removes a provider id that users and live combos may pin (for example the `onmi-gemini3.6` combos @HouMinXi flagged), so alias continuity and a DB migration path are hard requirements; 142 files.

**Merged #13410 (shared catalog)** — keeps both provider ids and shares one model list (`antigravitySharedModels.ts`). Zero breakage and a much smaller surface; provider-level metadata (quota/pricing/aliases/version) stays per-provider.

**My position — why I want the two merged:** I've been watching these two surfaces for a long time, and I've concluded there is no benefit at all in keeping them coexisting. The desktop and CLI versions use exactly the same models, tools and call logic — even the login flow is identical: if login fails on the CLI it fails on the desktop version too, and vice versa. There is no reason to keep unnecessary duplication and raise the code's complexity. I fully understand that people who have been using them may feel inconvenienced, but I believe a decisive move now — cut the complexity and keep only the essential surface — is the better trade, and it is what ultimately reduces user confusion rather than adding to it.

The mechanics back this up: `agy` and `antigravity` are the same Google consumer-OAuth client against the same Cloud Code backend, with one model catalog and one quota pool; the official Antigravity SDK is API-key only, so the emulated OAuth surface is the only subscription path either way. Keeping two provider ids preserves a client-identity split the upstream does not have. Alias continuity is a hard requirement, not a nice-to-have: `agy/…` model ids will keep resolving (DEPRECATED_PROVIDERS mapping plus model-prefix aliases), with a DB migration for existing connections. If we agree on the direction, I'll do the full rebase onto today's catalog refactor.

Either way I will do the work:

- If the shared-catalog approach stays: I will rescope this PR to drop the provider removal and keep only the non-breaking parts (client-identity normalization, catalog/version alignment) — or close it if #13410 makes those redundant.
- If consolidation is wanted: I will rebase the full proposal onto today's catalog refactor, keeping `agy/…` model aliases resolving and adding a migration path for existing connections.

#12608 has now been rewritten on top of `antigravitySharedModels.ts` (head 844aba8e6) and is gated on this same direction decision.

Production constraint worth recording: @HouMinXi kept `agy` in the recut precisely because live combos pin `agy/` model ids; any consolidation plan should treat alias continuity as a hard requirement rather than a nice-to-have.

backryun:
Restacked and rebased onto today's tree — `release/v3.8.51` @ `c8b24ffc3` plus the rewritten #12608 head (`844aba8e6`):

- Migration renumbered to `179_merge_agy_into_antigravity.sql` (178 is the current max); doc/SVG/llm-mirror counts re-synced (357 providers, 21 OAuth providers, 176 migrations, PROVIDER_REFERENCE regenerated).
- Kept the merged live-catalog discovery policy from #13410 instead of flipping the surfaces to a static allowlist; retired 3.7-era ids stay hidden and callable alias ids stay callable. The consolidated suites were aligned accordingly.
- Usage-service tests now expect the CLI-only fingerprint (`antigravity/cli/1.2.3`) this branch introduces.
- Focused run over the 62 test files touched by the stack: 650/651 pass. The single red — `tests/unit/models-catalog-route.test.ts` » "v1 models catalog does not duplicate custom Jina specialty models" — reproduces identically on a clean checkout of the tip (base-red inherited).

The direction question from my previous comment still stands (@diegosouzapw) — the branch is now clean and ready for review either way.

backryun:
Re-lifted onto today's tip (`cad0fcc65`), with the stack restacked on the refreshed #12608 head (`8ca8df0c2`). Version-consistent seeds/regexes now use the CLI `1.2.3` identity, and the shared-catalog suite no longer imports the deleted `agyModels` module. Focused run over the 63 touched test files: 656/657 — the single red is the Jina specialty-models base-red that reproduces on a clean tip.

backryun:
Direction-check reminder (@diegosouzapw): the question posted above is still open, and both this PR and #12608 stay gated on it — keep the shared-catalog approach, or go for the full single-provider consolidation?

Meanwhile the branch is clean and current: rebased onto the green tip (`release/v3.8.51` @ `0d089e7e3`), restacked on the refreshed #12608 head (`aa78baaeb`), focused suites green (657/657), docs counts in sync (358 providers / 177 migrations with this PR's migration at 180).

If the call is to keep the shared-catalog direction, I'll rescope this PR to the non-breaking parts (client-identity normalization, catalog/version alignment) — or close it if #13410 makes those redundant. If consolidation is wanted, I'm ready to proceed as-is. A yes/no unblocks both PRs.

backryun:
@diegosouzapw — re-upping the direction call now that the base is green and both PRs are rebased: full single-provider consolidation, or keep the shared-catalog approach? A one-line answer unblocks this and #12608 — if keeping shared catalogs, I'll rescope this PR to its non-breaking parts right away.

Issue #12877, opened 2026-09-06
fix(resilience): CredentialHealth: do not poison multi-upstream openai-compat connection on one model 402


RaviTharuma:
## Problem
On openai-compatible multi-upstream gateways (one connection fronting many models), a single upstream 402 marks the whole connection credits_exhausted / Auth-fail. Sibling models that still return 200 get wrongly blocked.

## Wanted
Per-model / per-upstream CredentialHealth scoping for openai-compatible / hasPerModelQuota paths.

## Related (not duplicates)
- #12242 / #12266 — live 402 model-scope
- #12532 / #12533 — off-loop monitoring cache

## PR
https://github.com/diegosouzapw/OmniRoute/pull/12875

diegosouzapw:
Thanks for reporting this — good catch. This is already fixed: PR #12875 (the PR you linked in the issue body) merged on 2026-09-10 and is on `release/v3.8.51`. It gates the 402-handling path in `open-sse/handlers/chatCore/keyHealth.ts` behind `hasPerModelQuota(provider)` (from `open-sse/services/accountFallback.ts`), so on openai-compatible / per-model-quota gateways a 402 from one upstream model now only locks that model (`lockModelIfPerModelQuota`) instead of calling `recordKeyTerminal` on the whole connection. Sibling models on the same connection stay routable.

I re-ran the regression suite that PR added (`tests/unit/openai-compatible-per-upstream-402-health.test.ts`) against the current tip and all 6 cases pass, including the exact scenario from this issue (one model 402, sibling model still served) and the pre-existing single-credential 402 terminal case (#5239), which is intentionally unchanged.

Note: PR #13185, which also mentions this issue in its description, is unrelated — that one fixes a Claude OAuth sticky-refresh bug (#13183), not this.

Closing as already-fixed; if you're still seeing whole-connection poisoning on a recent build, please reopen with the connection's provider id/type and the exact status/error body you're getting.

PR #11414, opened 2026-08-24
fix(oauth): allow expired token retry before deactivation and activate expired retry loop (#P0)

HouMinXi:
## Summary
- Fixes P0 bug where OAuth provider connections (Antigravity/Google, Claude, etc.) were prematurely deactivated (`isActive: false`) on the very first token refresh failure, and the existing `EXPIRED_RETRY_MAX` exponential backoff retry loop was dead code blocked by overly strict gates.
- Restores the missing `EXPIRED_RETRY_MAX` and `EXPIRED_RETRY_BACKOFF_MIN` constants in `src/lib/tokenHealthCheck.ts`.
- Updates the `checkConnection` gate at lines 516 and 548 so that connections with `testStatus: 'expired'` and remaining retry budget (`expiredRetryCount < EXPIRED_RETRY_MAX`) are allowed through to the retry loop.
- In the unrecoverable error path, postpones `isActive: false` deactivation until the retry budget is exhausted (`expiredRetryCount >= EXPIRED_RETRY_MAX`), keeping accounts active during transient upstream or proxy glitches.
- Persists `expiredRetry: { count, at }` inside `providerSpecificData`, matching the established `refreshCircuit` storage pattern.
- Adds comprehensive unit tests in `tests/unit/token-health-check-retry-deactivation.test.ts` verifying retry progression and deactivation only upon budget exhaustion.

## Verification
- `node --import tsx/esm --test tests/unit/token-health-check*.test.ts tests/unit/cursor-renewal.test.ts` (74/74 pass)
- Code-forge review: PASS (0 findings).
- ⚠️ base-red inherited: #9985

Issue #3679, opened 2026-06-11
[BUG] [BUG] Google OAuth refresh_token not persisted — connections die after 2–3 refresh cycles


3xa228148:
### OmniRoute Version

3.8.21

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Linux 7.0.10-arch1-1 x86_64

### Node.js Version

v24.16.0

### Provider(s) Involved

Antigravity, Gemini CLI (both Google OAuth-backed)

### Description

Google OAuth refresh tokens are not persisted to the database after successful refreshes. The refresh_token column in provider_connections is NULL for all Google OAuth connections immediately after initial authorization. The encrypted access_token blob does not get updated with the new refresh_token after a successful refresh cycle. This causes connections to die with invalid_grant / Bad Request from Google after 2–3 refresh cycles, because the old (already-consumed) refresh token is sent again.

### Steps to Reproduce

1. Add a Google OAuth provider connection (Antigravity or Gemini CLI) via Dashboard
2. Wait for the first health-check refresh (~1 hour after creation, or restart the server)
3. Observe: first refresh succeeds, second refresh succeeds
4. Observe: third refresh fails with 400 invalid_grant / Bad Request from Google
5. Connection is permanently deactivated with unrecoverable_refresh_error

### Expected Behavior

After each successful token refresh, the new refresh_token (Google rotates them on every use) should be saved to the database. The connection should survive indefinitely as long as the user doesn't revoke access.

### Actual Behavior

- refresh_token column is NULL from the moment of initial OAuth authorization
- Encrypted access_token blob is 589 chars and does not appear to be updated after successful refreshes
- updated_at timestamp only changes on failure, never on successful refresh
- Connections die after 2–3 refresh cycles with Google's invalid_grant

Evidence from DB:

provider     | email                | refresh_token | updated_at          | status
-------------+----------------------+---------------+---------------------+--------
gemini-cli   | g***@gmail.com    | NULL          | (failure time only) | expired
gemini-cli   | l***@gmail.com     | NULL          | (failure time only) | expired
antigravity  | g***@gmail.com    | NULL          | (failure time only) | expired
antigravity  | l***@gmail.com     | NULL          | (failure time only) | expired

### Test Impact

Unsure

### Error Logs / Output

```shell
[HealthCheck] [TOKEN_REFRESH] Failed to refresh Google token {
  "status": 400,
[...]

PR #13185, opened 2026-09-10
fix(oauth): soft-fail Claude refresh so CredentialHealth is not sticky-dead

### LINKED ISSUE #13183: fix(auth): Claude OAuth refresh failure leaves connection sticky-dead under CredentialHealth (dual-consumer aggravator)
## Summary

Claude OAuth connections can become **sticky-dead** after a normal access-token TTL (~8h) refresh failure: OmniRoute marks the connection expired / `is_active=0`, and CredentialHealth keeps probing a failed route so the account stays unusable until a human re-enables / re-logins — even when a fresh login would succeed.

## Observed behaviour

1. Claude OAuth access token expires (~8h TTL).
2. Refresh fails (or races).
3. Connection is marked expired / deactivated (`is_active=0` or equivalent terminal-ish auth state).
4. CredentialHealth (~tens of seconds) keeps failing the same route → sticky until manual re-enable / re-login.
5. Operator can restore service immediately by re-authenticating.

## Aggravator (dual consumer)

Using the **same** Claude Max OAuth account simultaneously in **OmniRoute and CLIProxyAPI** (or another refresh client) can produce dual refreshers / mutual refresh-token invalidation, accelerating sticky failure.

## Related (not duplicates)

- #12859 / #12864 — Anthropic `403 Request not allowed` classified as terminal ban (different trigger; same "sticky dead OAuth" family).
- #12875 / #12877 — CredentialHealth per-upstream scoping on openai-compat 402 (different provider path).
- #12594 — Cline auth going sticky-unauthorized (sibling pattern, different provider).

## Wanted

1. **Durable refresh + CredentialHealth behaviour** so a failed refresh does not leave Claude OAuth sticky-dead: retry/backoff, clear recoverable expiry after successful refresh, and avoid permanent deactivate on a single refresh miss.
2. **Optional:** detect / warn when the same OAuth subject appears to be consumed by multiple refreshers (dual-consumer), so operators are not surprised by mutual invalidation.

## Environment

OmniRoute (self-hosted), Claude OAuth / Claude Max subscription path. No secrets or hostnames in this report.

## Code pointers (from tree)

- `src/lib/tokenHealthCheck.ts` — Claude in `ROTATING_REFRESH_PROVIDERS`; unrecoverable refresh path sets `isActive: false` / can null rotating `refreshToken`.
- `open-sse/services/tokenRefresh.ts` — Claude refresh lead time; notes Anthropic rotates refresh tokens; dual-account family invalidation comments already exist nearby.
- CredentialHealth gate (`open-sse/services/credentialGate.ts` / `src/lib/credentialHealth`) keeps probing failed connections.

## Validation plan

- Simulated refresh failure does not permanently deactivate; recovery after successful refresh restores routing without manual dashboard toggle.
- CredentialHealth after recoverable refresh failure does not keep the connection sticky-failed forever.
- (Optional) dual-consumer heuristic surfaces a non-fatal warning when configured.

PR to follow.

RaviTharuma:
## Summary
Fixes sticky-dead Claude OAuth after refresh failure (#13183).

Claude access tokens expire ~8h. When refresh returns `invalid_grant` / `refresh_token_reused` (often a dual-consumer race with another OAuth consumer refreshing the same Claude Max OAuth account), OmniRoute previously set `isActive=false` and nullified the rotating refresh token on the first failure. CredentialHealth then kept failing the route until a manual reactivation.

### Changes
- Soft-fail Claude unrecoverable refresh: circuit-breaker backoff, keep `isActive` + `refreshToken`, mark `expired` with a dual-consumer hint
- Hard-deactivate only after 3 consecutive unrecoverable failures; still preserve Claude `refreshToken` on hard path
- CredentialHealth skips probes while `refreshCircuit.until` is in the future (shared `tokenRefreshCircuit` helper)
- Unit tests for soft/hard threshold helpers
- **Proactive refresh lead (unchanged):** Claude refreshes at `expires_at − 5m` via `TOKEN_EXPIRY_BUFFER` in `tokenHealthCheck` and `CLAUDE_REFRESH_BUFFER_MS` in `claudeAuthFile` (both 5 minutes). Kept as-is; no skew change in this PR.

### Workaround until merge
Do not refresh the same Claude Max OAuth account from OmniRoute and another OAuth consumer (e.g. dual-consumer with CPA) at the same time. Re-authenticate in OmniRoute after settling the other client.

## Test plan
- [ ] Unit: `tests/unit/tokenHealthCheck-soft-unrecoverable.test.ts`
- [ ] Simulate Claude refresh `invalid_grant` with expired access token → connection stays active, circuit set, no `isActive=false` on first failure
- [ ] Third consecutive failure → hard deactivate without nulling Claude refresh token
- [ ] CredentialHealth does not probe during open refresh circuit
- [ ] Codex/OpenAI unrecoverable path still hard-deactivates immediately (unchanged)
- [ ] Confirm Claude proactive refresh still triggers ~5m before expiry (existing buffers)

Closes #13183

Related: #12877, #12875


diegosouzapw:
Thanks for this one — the diagnosis was right, and it turned out to be sharper than the issue thread suggested.

I re-targeted the PR from `main` to `release/v3.8.51` (the default branch, 3774 commits ahead of `main`) and rebased your commit onto it, so the diff is the 4 files instead of the ~1292 a naive re-target would have dragged in. Authorship stays yours.

**Why I reconciled the change instead of applying it as written**

`release/v3.8.51` already carries #11414, which keeps an unrecoverable refresh retryable for `EXPIRED_RETRY_MAX` sweeps. That is why I closed #13183 earlier today — and I was wrong to close it, because that mechanism never actually runs for Claude:

```
tokenHealthCheck.ts   ...(isRotatingProvider ? { refreshToken: null } : {})
```

That line sits in the *same* update that keeps `isActive` true for the retries, and `claude` is in `ROTATING_REFRESH_PROVIDERS`. So the first `invalid_grant` wipes the refresh token. On the next sweep `checkConnection` stops at the `!conn.refreshToken` guard, whose self-heal branch only fires while `testStatus` is empty or `"active"` — the row is already `"expired"`, so it returns silently. The connection sits active, expired and unrecoverable, and the retry budget is never spent. Exactly the sticky-dead state you reported.

The #11414 regression test missed it because it drives a synthetic provider that is not in `ROTATING_REFRESH_PROVIDERS`, so the token was never nulled there.

**What landed**

- Claude opts out of clearing the rotating refresh token on the unrecoverable path — same reasoning as #3679 for non-rotating providers. Codex and the rest keep clearing their genuinely consumed one-time-use tokens.
- CredentialHealth honors the refresh circuit and parks the next attempt on the circuit deadline instead of re-probing a connection already in refresh backoff. That is your `scheduler.ts` change, kept as-is.
- `isInRefreshBackoff` moves to `src/lib/tokenRefreshCircuit.ts` so CredentialHealth can use it without importing the auto-starting scheduler. That is your extraction, kept as-is; `tokenHealthCheck` re-exports it so existing callers and tests are untouched.

**What I dropped, and why**

The soft-fail streak (`refreshCircuit.streak` up to 3 before hard deactivation). On this branch it would be a second failure counter running alongside #11414's `expiredRetry`, giving 3 soft attempts plus 3 retries before deactivation and two sources of truth for the same question. Preserving the token is what makes the existing counter work; the streak is not needed once it does.

**Validation**

The regression test drives the real `claude` provider through two consecutive sweeps, re-reading the row from the database in between:

| | without the fix | with the fix |
|---|---|---|
| Claude keeps the refresh token on the 1st failure | fail | pass |
| Claude spends retry 2 on the next sweep | fail | pass |
| Codex still clears its one-time-use token | pass | pass |

24 tests green across the new file plus `token-health-check-circuit-breaker` and `tokenHealthCheck-transient`, which import the moved symbol. `typecheck:core`, ESLint and the file-size gate are clean. I shrank `tokenHealthCheck.ts` by moving the new helper into the shared module rather than re-baselining the size gate.

On item 2 from the issue, detecting a second consumer of the same OAuth account: still not implemented, and still worth a separate feature request if you want to propose a heuristic.

Reopening #13183, since it was not actually fixed on the release branch.

diegosouzapw:
Following up after the owner's own reconciliation comment — the head commit already
reflects that design, and I re-verified it independently: ran the regression test against
the real `claude` provider through two DB-backed sweeps (4/4 green), lint clean on all
three touched files. Two purely operational items before merge: a `changelog.d/fixes/`
fragment (none in the diff) and a rebase onto the current `release/v3.8.51` tip (the
branch is a bit behind). No design concerns from my side — the owner's write-up already
covers that ground well.

ai-jeremi-esky:
Complementary burn-prevention is in #13874: `serializeRefresh` currently only *orders* two POSTs of the same Claude refresh token (request-path Layer 2 vs HealthCheck), so the waiter still presents the already-consumed token and Anthropic burns it. #13874 re-checks the rotation map and DB *inside* that lock, records Layer 2 rotations, and stops HealthCheck from nulling a Claude RT.

This PR (#13185) remains the sticky-dead recovery (preserve the token after `invalid_grant`). The two are not substitutes.

eSky overlay until a pin includes #13874: https://github.com/eskygroup/esky-virtual-staff-engineer-jeremi/pull/1391

PR #12499, opened 2026-09-02
feat(antigravity): add Gemini 3.8 Flash tiers and update client fallbacks

Abhishekchhetri020:
## Summary

- Add Google's newly launched Gemini 3.8 Flash tiers (`gemini-3.8-flash-high`, `gemini-3.8-flash-medium`, `gemini-3.8-flash-low`) to the `antigravity` and `agy` public model catalogs.
- Map the shorthand `gemini-3.8-flash` alias to `gemini-3.8-flash-high`.
- Add Gemini 3.8 Flash model specifications in `modelSpecs.ts` (1M context window, 64k max output tokens, 24.5k/8k/1k thinking budget tiers).
- Update fallback client versions in `antigravityVersion.ts` (`ANTIGRAVITY_IDE_FALLBACK_VERSION` to `2.5.5` and `ANTIGRAVITY_CLI_FALLBACK_VERSION` to `1.1.24`). Google's live Cloud Code endpoint (`streamGenerateContent`) requires a recognized client identity and rejects older client versions (`< 2.5.0` or `< 1.1.24`) with HTTP 404.
- Add pricing, free model metadata, and CLI tools registry entries for the 3.8 Flash tiers.
- Update regression test suites across Antigravity and AGY test suites.

## Related Issues

- Related to #10422

## Validation

- [x] Change type: provider / routing
- [x] Focused tests:
  - `node --import tsx/esm --test tests/unit/antigravity-model-aliases.test.ts tests/unit/antigravity-retired-public-models.test.ts tests/unit/model-capabilities-registry.test.ts tests/unit/t28-model-catalog-updates.test.ts tests/unit/provider-models-route.test.ts` (94 passed, 0 failed)
- [x] Reconciled with current active release base (`release/v3.8.51`)
- [x] Verified empirically against live Cloud Code API with real consumer-auth credentials

## Tests Added Or Updated

- `tests/unit/antigravity-model-aliases.test.ts`
- `tests/unit/antigravity-retired-public-models.test.ts`
- `tests/unit/model-capabilities-registry.test.ts`
- `tests/unit/provider-models-route.test.ts`
- `tests/unit/t28-model-catalog-updates.test.ts`

## Coverage Notes

- Covered by the existing and updated unit test suites in `tests/unit/antigravity-model-aliases.test.ts`, `antigravity-retired-public-models.test.ts`, and `model-capabilities-registry.test.ts`.

## Reviewer Notes

[...]

PR #13195, opened 2026-09-10
fix(models): give Gemini 3.8 Flash its own output spec

HouMinXi:
## Summary

Discoverable Gemini 3.8 Flash ids on Antigravity/AGY fell through to the conservative 16384 output clamp. `MODEL_SPECS` listed 3.7 Flash tiers and left 3.8 without an explicit cap, so `resolveAntigravityOutputCap` treated a live catalog model as unknown.

This PR adds Gemini 3.8 Flash specs (high/medium/low/tiered/base) with `maxOutputTokens: 65536`, and tests the executor clamp plus capability lookup.

## Changes

- Declare `gemini-3.8-flash-high|medium|low|tiered` and `gemini-3.8-flash` in `MODEL_SPECS`.
- Keep the existing 16384 fallback for unknown/retired ids (`gpt-oss-120b-medium` stays 32768).
- Unit coverage: clamp with thinking budget, `agy/` prefix, and `getResolvedModelCapabilities` thinking/context fields.

## Related open PRs

- #12663 registers `gemini-3.8-flash` on the native Gemini provider. Different file.
- #12499 and #12672 also add 3.8 entries to `MODEL_SPECS`. #12672 omits `defaultThinkingBudget` / `thinkingBudgetCap` on purpose: it says 3.8 steers thinking via `thinkingLevel`, and a copied 3.7 budget would inject `thinkingBudget` that 3.8 does not use. This PR copies the 3.7 high/medium/low/tiered budgets so the output clamp and capability lookup stay filled. If #12672 lands first, rebase and drop the duplicate spec block; keep the clamp tests.

## Test plan

- `node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit tests/unit/antigravity-per-model-output-cap.test.ts tests/unit/copilot-agent-antigravity-parity.test.ts tests/unit/antigravity-model-aliases.test.ts`
- Result: 30 pass, 0 fail.

Contributor only. Please do not merge from this account.

### DIFF
--- a/changelog.d/fixes/13195-gemini-38-output-spec.md
+++ b/changelog.d/fixes/13195-gemini-38-output-spec.md
@@ -1,0 +1,1 @@
+- **fix(models):** give discoverable Gemini 3.8 Flash ids their own 65536 output spec so Antigravity no longer clamps them to 16384 ([#13195](https://github.com/diegosouzapw/OmniRoute/pull/13195)) — thanks @HouMinXi
--- a/src/shared/constants/modelSpecs.ts
+++ b/src/shared/constants/modelSpecs.ts
@@ -181,9 +181,56 @@ export const MODEL_SPECS: Record<string, ModelSpec> = {
     supportsTools: true,
     supportsVision: true,
   },
-  // ── Gemini 3.7 Flash (current Antigravity/AGY live tiers) ─────────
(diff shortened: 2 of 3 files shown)


HouMinXi:

Issue #6912, opened 2026-07-11
fix(providers): max_completion_tokens ignored when thinking mode enabled (non-ClinePass providers)


HouMinXi:
## Root Cause

`ensureThinkingBudget()` in `open-sse/executors/default.ts` is **ClinePass-gated** (`if (this.provider !== "clinepass") return body`). For all other providers (DeepSeek/Volcengine, Kimi, etc.), the thinking budget is never enforced at the gateway level.

## Reproduction

1. Configure a DeepSeek V4 Flash endpoint via Volcengine in OmniRoute
2. Send request with `max_completion_tokens: 30` and thinking/reasoning enabled
3. Expected: response truncated at ~30 tokens
4. Actual: response returns 2826 tokens with `finish_reason: "stop"`

## Impact

- Forge code review expert pass fails (truncated JSON response)
- Any service relying on `max_completion_tokens` to limit output gets unbounded responses
- Volcengine free-tier quota burns 100x faster than expected

## Suggested Fix

Remove the ClinePass gate from `ensureThinkingBudget()` or add equivalent enforcement for all providers that support reasoning/thinking mode. The function should apply whenever reasoning is enabled, regardless of provider.

```typescript
// Current (ClinePass-only):
if (!body || this.provider !== "clinepass") return body;

// Proposed (all providers with reasoning support):
if (!body) return body;
```

Or add a separate generic enforcement in the request pipeline that caps `max_completion_tokens` for all thinking-enabled requests before forwarding to the provider.

diegosouzapw:
Thanks for reporting! Fixed in `release/v3.8.47` (already merged into the active development branch — feel free to pull and test it now). It will ship in the next release (v3.8.47).

PR #13199, opened 2026-09-10
fix(reka): floor reka-flash-3 output budget so always-on reasoning can't empty content

### LINKED ISSUE #13198: fix(providers): reka-flash-3 always-on reasoning burns tiny max_tokens, returns empty content (finish_reason:length)
## Summary
When a generic OpenAI-compatible client (e.g. Hermes Agent) sets a small `max_tokens` on **reka-flash-3** via OmniRoute, the model's always-on reasoning phase consumes the entire output budget. The upstream returns an OpenAI-compatible response with `content: ""`, a large `reasoning_content`, and `finish_reason: "length"`. OmniRoute transparently forwards this reasoning-only, empty-content response to the client.

I reproduced this (no client, raw curl, OmniRoute 3.8.51 on `release/v3.8.51`):

```json
{
  "model": "reka-flash-3",
  "choices": [{ "finish_reason": "length", "message": { "role": "assistant", "content": "", "reasoning_content": "…" } }],
  "usage": { "prompt_tokens": 33, "completion_tokens": 256, "total_tokens": 289 }
}
```

- `max_tokens=256` via reka-flash-3 → `content:""`, `finish_reason:length`, reasoning burned the whole budget.
- `max_tokens=512/1024/2048/4096` → normal `content` (e.g. `TELEGRAM_TEST_123`), `finish_reason:stop`.

This is the same bug class as #6912 / #12497 / #12742 (reasoning models burning tiny budgets on hidden reasoning), but **reka-flash-3 is missing the reasoning metadata** in the registry, so the existing `ensureThinkingBudget()` floor (#6912) short-circuits and never raises `max_tokens`.

The client-side impact can be severe: an agent gateway that forwards the reasoning-only empty response can turn it into a huge final answer (e.g. the Hermes Telegram case in the linked PR), because the reasoning content is exposed as output.

## Environment
- OmniRoute 3.8.51 (release/v3.8.51)
- Provider: reka (registry entry), executor `default`
- Repro via `POST /v1/chat/completions` with `model: "reka-flash-3"`

## Repro
```bash
curl -sS -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  http://127.0.0.1:20128/v1/chat/completions \
  -d '{"model":"reka-flash-3","messages":[{"role":"system","content":"Reply with exactly TELEGRAM_TEST_123"},{"role":"user","content":"Reply with exactly TELEGRAM_TEST_123"}],"max_tokens":256,"stream":false}'
```

Observed: `model: reka-flash-3`, `finish_reason: length`, `content: ''`, `completion_tokens: 256`.
With `max_tokens: 4096`: `content: 'TELEGRAM_TEST_123'`, `finish_reason: stop`.

## Root cause
- `reka-flash-3` is declared in `open-sse/config/providers/registry/reka/index.ts` with **no** `supportsReasoning` / budget metadata.
- `DefaultExecutor.ensureThinkingBudget()` (`open-sse/executors/default.ts`) only bumps undersized budgets when the client explicitly opts into reasoning (`thinking.type === "enabled"` or a nonzero `reasoning_effort`). reka-flash-3 reasons unconditionally, so the floor never fires for a plain request with `max_tokens: 256`.

## Proposal
Add an `alwaysReasons` capability to the registry (set true on reka-flash-3), and have `ensureThinkingBudget()` treat it as implicit reasoning opt-in so a tiny caller `max_tokens` gets the 4096 floor.

## Related issues / PRs
- #6912 – reasoning models burn all of max_tokens on thinking, fixed by `ensureThinkingBudget`
- #12497 – budget floor for muse-spark (same class, different provider/executor)
- #12742 – opt-in min output budget floor for thinking models (env-gated)

hummern:
## Summary
Fixes #13198. `reka-flash-3` always runs a reasoning phase, so a caller-set undersized `max_tokens` (e.g. `256`) is consumed entirely by `reasoning_content` and the upstream returns empty content with `finish_reason: length`. A transparent OpenAI-compatible proxy forwards that reasoning-only, empty-content response to the client, which can surface as huge garbage output (e.g. Hermes Agent's forwarded reasoning onto Telegram).

`ensureThinkingBudget()` (#6912) already floors undersized budgets for reasoning models, but it only fired when the client explicitly opted into reasoning. reka-flash-3 reasons unconditionally.

## Change
- Add `RegistryModel.alwaysReasons` (boolean): true when an upstream always reasons regardless of client opt-in. Set it on `reka-flash-3`.
- `DefaultExecutor.ensureThinkingBudget()` treats `alwaysReasons === true` as implicit reasoning opt-in, so a tiny caller `max_tokens` gets the 4096 floor.
- Regression tests (verified to fail without the fix).

## Repro → fixed
```
max_tokens=256  → content:'', finish_reason:length          (before)
max_tokens=4096 → content:'TELEGRAM_TEST_123', finish_reason:stop  (after)
```

## Tests
- `tests/unit/clinepass-thinking-budget.test.ts` — 9/9 pass; the new `alwaysReasons` cases fail without the fix (sabotage-verified).
- `tests/unit/provider-models-config.test.ts` — 12/12 pass.
- `tests/unit/provider-models-route.test.ts` — 59/59 pass.
- `npx tsc --noEmit -p tsconfig.typecheck-core.json` — clean (exit 0).
- Pre-commit hooks (prettier, eslint, any-budget, quality gates) passed.

## Related
- #6912, #12497, #12742 (same class: reasoning models burning tiny output budgets)

## Reviewer note
The two eslint errors in the touched files (`shared.ts:791`, `default.ts:5`) are pre-existing baseline warnings on lines I did not modify, not introduced by this diff.


diegosouzapw:
The `alwaysReasons` floor for reka-flash-3 is a clean, well-tested fix (9/9 passing, good
coverage of the non-clinepass and non-regression cases). One thing to flag before merge:
your branch carries an unrelated commit (`c70e425`, "Change hasFree from true to false for
pioneer.ai") that touches `frontier-labs.ts` — nothing to do with reka-flash, and it looks
stale against the current Pioneer AI notice text on the tip. Could you drop that commit (or
rebase it out)? If the Pioneer AI free-tier change is still accurate, happy to see it as its
own small PR with a current source.

hummern:
Closing this PR — the alwaysReasons floor for reka-flash-3 was already merged upstream via #13531 (2d1a281d1), and the unrelated pioneer.ai commit (c70e425) has also landed in release/v3.8.51. The branch is now rebased clean against the current tip. Thanks for the review feedback!

hummern:
Superseded — fix already merged via #13531.

Issue #13137, opened 2026-09-09
fix(backend): Eval case whose model call failed is scored as passed — runSuite() never forces passed=false


aaustinhuang:
**Version:** 3.8.50 (global npm install, `omniroute` CLI + dashboard)

## Summary

An eval case whose upstream call failed can be recorded as `"passed": true`. The reported pass rate on the Evals dashboard is inflated by however many such cases a run contains, and a run in which no model was ever reached still reports a non-zero score.

## What happens

`executeEvalCase()` in `src/lib/evals/runtime.ts` turns a failed call into a normal-looking output string:

```js
if (!response.ok) {
  const error = extractErrorMessage(payload, response.status);
  return {
    output: `[ERROR] ${error}`,
    durationMs,
    error,
  };
}
```

`runSuite()` in `src/lib/evals/evalRunner.ts` then grades that string like a real answer and only attaches the error afterwards:

```js
const result = evaluateCase(c, output);
const metrics = caseMetrics[c.id];
...
if (metrics?.error && !result.error) {
  result.error = metrics.error;
}
```

`passed` is never forced to `false`. Any expected pattern that happens to match the error text scores a pass.

## Reproduction

Built-in suite `codex-comparison`, case `codex-07` "Error handling pattern":

| | |
|---|---|
| expected | `strategy: "regex"`, `value: "try\|catch\|throw\|error\|Error"` |
| call result | fails — `Unable to determine provider for model 'codex'` |
| graded string | `[ERROR] Unable to determine provider for model 'codex'. …` |
| recorded | `"passed": true` with a non-empty `"error"` |

Any install where a built-in suite's hardcoded model does not resolve reproduces this. On a machine with no OpenAI and no Google credentials, all 44 built-in cases fail in single-digit milliseconds and `codex-comparison` reports 13% instead of 0%.

## Suggested fix

```diff
   if (metrics?.error && !result.error) {
     result.error = metrics.error;
   }
+  if (metrics?.error) {
+    result.passed = false;
+  }
```

A case that never reached a model has no measured behaviour to grade, so a failure is the only honest score.

PR #13201, opened 2026-09-10
fix(evals): fail a case whose model call errored instead of scoring it passed

dajiaohuang:
Fixes the eval-runner scoring defect reported in #13137.

## What was wrong

`runSuite()` (`src/lib/evals/evalRunner.ts`) attached `caseMetrics[id].error` to the graded
result but never forced `passed` to `false`. `executeEvalCase()` (`src/lib/evals/runtime.ts:229-236`)
returns a failed upstream call as an ordinary output string:

```js
output: `[ERROR] ${error}`,
```

so the error text was graded like a real answer. Any expected pattern that happened to match
that text scored a pass while still carrying a non-empty `error`.

## Reproduction

Built-in `codex-comparison` case `codex-07` ("Error handling pattern") expects
`try|catch|throw|error|Error`. When the model does not resolve, the failure text is
(`open-sse/services/model.ts:851`):

```
Unable to determine provider for model 'codex'. Use a provider/model prefix
(e.g. openai/codex) or ensure the model is added as a combo entry.
```

That ends in "combo ent**ry**", so the `try` alternative matches and the case is recorded as
`"passed": true` with a non-empty `"error"`.

## Change

A case that never reached a model has no measured behaviour to grade, so a failure is the only
honest score. The error is still attached exactly as before.

Regression test added: `tests/unit/eval-runner-errored-case-scored-passed.test.ts`.

- **Red** (fix reverted): `✖ fails a case whose call errored even when the error text matches the pattern`
  → `AssertionError: a case that never reached a model has no measured behaviour to grade`
- **Green** (fix applied): 2/2 pass
- **No regression**: `batch-b-final.test.ts` + `evalrunner-builtinsuites-split.test.ts` → 38/38 pass

Commands run:

```bash
node --import tsx/esm --test tests/unit/eval-runner-errored-case-scored-passed.test.ts
node --import tsx/esm --test tests/unit/batch-b-final.test.ts tests/unit/evalrunner-builtinsuites-split.test.ts
npx prettier --write src/lib/evals/evalRunner.ts tests/unit/eval-runner-errored-case-scored-passed.test.ts
npx eslint src/lib/evals/evalRunner.ts tests/unit/eval-runner-errored-case-scored-passed.test.ts
```

## Test files added

- `tests/unit/eval-runner-errored-case-scored-passed.test.ts`

## Scope

Two files: the `runSuite()` error branch and its regression test. No API, schema, dependency, or
CI change. The changelog fragment lands in a follow-up commit on this branch.

⚠️ base-red inherited: #12732

### DIFF
--- a/changelog.d/fixes/13201-eval-errored-case-scored-passed.md
+++ b/changelog.d/fixes/13201-eval-errored-case-scored-passed.md
@@ -1,0 +1,1 @@
+- **fix(evals):** an eval case whose model call errored is no longer scored as passed — a case that never reached a model has no measured behaviour to grade ([#13201](https://github.com/diegosouzapw/OmniRoute/pull/13201)) — thanks @aaustinhuang
--- a/src/lib/evals/evalRunner.ts
+++ b/src/lib/evals/evalRunner.ts
@@ -239,8 +239,15 @@ export function runSuite(
       result.durationMs = Math.max(0, Math.round(Number(metrics.durationMs)));
     }
 
-    if (metrics?.error && !result.error) {
-      result.error = metrics.error;
+    if (metrics?.error) {
+      // A case whose call errored never reached a model, so there is no measured
+      // behaviour to grade. The runner surfaces failures as an ordinary output string
+      // ("[ERROR] …" — see executeEvalCase), so leaving `passed` alone lets a pattern
+      // that happens to match that text score a pass and inflate the reported rate.
+      result.passed = false;
+      if (!result.error) {
+        result.error = metrics.error;
+      }
     }
 
     return result;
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Sharp catch and a clean, minimal fix — the repro (a case whose error text coincidentally
matches its own expected pattern) is exactly right, and the test proves it without
reimplementing the logic. Ran it locally: 2/2 pass. This one's ready to merge as-is.

diegosouzapw:
Thanks @dajiaohuang — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

Issue #13138, opened 2026-09-09
fix(backend): Eval regex grading compiles without dotAll — correct multi-line answers are marked wrong


aaustinhuang:
**Version:** 3.8.50

## Summary

Regex grading compiles string patterns with no flags, so `dotAll` is off and `.` does not match a newline. LLM answers are routinely multi-line, and several built-in cases use `.*` to join tokens a model will naturally put on separate lines. Those cases fail on correct answers.

## What happens

`evaluateCase()` in `src/lib/evals/evalRunner.ts`, `case "regex"`:

```js
const regex =
  expectedValue instanceof RegExp
    ? new RegExp(expectedValue.source, expectedValue.flags.replace(/[gy]/g, ""))
    : new RegExp(expectedValue);
```

Suite JSON can only supply a string, so the second branch is the one that runs for every custom and built-in suite. `s` is never set.

## Reproduction

Three built-in cases, all with correct model output, measured on 3.8.50:

| case | pattern | model output | scored |
|---|---|---|---|
| `gs-09` Counting | `1.*2.*3.*4.*5` | `1\n2\n3\n4\n5` | fail |
| `code-03` SQL — SELECT query | `SELECT.*FROM.*WHERE` | ```` ```sql\nSELECT *\nFROM users\nWHERE age > 25` ```` | fail |
| `instr-02` Numbered list format | `1\..*2\..*3\..*4\..*5\.` | `1. Mercury\n2. Venus\n3. Earth\n4. Mars\n5. Jupiter` | fail |

The graded answers are exactly what the cases ask for. On a 44-case set this cost about 7 percentage points of reported pass rate on its own.

## Suggested fix

Compile string patterns with `s`, and preserve it when a `RegExp` is supplied:

```diff
-    : new RegExp(expectedValue);
+    : new RegExp(expectedValue, "s");
```

The `safeRegex` guard and the 512-character source limit are unaffected — `dotAll` changes matching semantics, not backtracking behaviour.

As a workaround without patching, suite authors can write `[\s\S]*` instead of `.*`.

PR #13202, opened 2026-09-10
fix(evals): compile string regex patterns with dotAll so multi-line answers grade correctly

dajiaohuang:
Fixes the eval regex-grading defect reported in #13138.

## What was wrong

`evaluateCase()` (`src/lib/evals/evalRunner.ts`) compiled suite-supplied string patterns with
**no flags**, so `dotAll` was off and `.` did not cross a newline:

```js
const regex =
  expectedValue instanceof RegExp
    ? new RegExp(expectedValue.source, expectedValue.flags.replace(/[gy]/g, ""))
    : new RegExp(expectedValue);   // <- every suite-supplied pattern takes this branch
```

Suite JSON can only carry a pattern as a string, so the second branch is the one every custom
and built-in suite takes. LLM answers are routinely multi-line, and several built-in cases join
tokens with `.*` that a model naturally puts on separate lines.

## Reproduction

Three built-in cases, all with correct model output:

| case | pattern | correct model output | before |
|---|---|---|---|
| `gs-09` Counting | `1.*2.*3.*4.*5` | `1\n2\n3\n4\n5` | fail |
| `code-03` SQL — SELECT query | `SELECT.*FROM.*WHERE` | ```` ```sql\nSELECT *\nFROM users\nWHERE age > 25``` ```` | fail |
| `instr-02` Numbered list format | `1\..*2\..*3\..*4\..*5\.` | `1. Mercury\n2. Venus\n3. Earth\n4. Mars\n5. Jupiter` | fail |

## Change

Compile string patterns with `s`. The graded answers are exactly what the cases ask for.

**Why this cannot regress a passing case:** `dotAll` only *widens* `.` (from "any char except
line terminators" to "any char"). Any match that succeeded before still succeeds, because the
same characters still satisfy `.`. So `passed` can only move `false → true`, never `true → false`.

An author-supplied `RegExp` keeps exactly the flags they wrote — only the string branch changes,
so `/1.*5/` without `s` still does not match across a newline while `/1.*5/s` does.

The `safeRegex` guard and the 512-character source limit are untouched; `dotAll` changes matching
semantics, not backtracking behaviour.

Regression test added: `tests/unit/eval-runner-regex-dotall.test.ts`.

- **Red** (fix reverted): the three multi-line cases fail — e.g.
  `✖ matches a multi-line answer against a .*-joined pattern (gs-09 Counting)`
- **Green** (fix applied): 7/7 pass
- **No regression**: `batch-b-final.test.ts` + `evalrunner-builtinsuites-split.test.ts` → 38/38 pass

The test also pins the safety behaviour: a catastrophic pattern (`(a+)+$`) is still rejected by
`safeRegex`, an over-long source (>512) is still rejected, a genuinely non-matching answer still
fails, and an anchored single-line pattern still grades correctly.

Commands run:

```bash
node --import tsx/esm --test tests/unit/eval-runner-regex-dotall.test.ts
node --import tsx/esm --test tests/unit/batch-b-final.test.ts tests/unit/evalrunner-builtinsuites-split.test.ts
npx prettier --write src/lib/evals/evalRunner.ts tests/unit/eval-runner-regex-dotall.test.ts
npx eslint src/lib/evals/evalRunner.ts tests/unit/eval-runner-regex-dotall.test.ts
```

## Test files added

- `tests/unit/eval-runner-regex-dotall.test.ts`

## Scope

Two files: the string-pattern compilation branch and its regression test. No API, schema,
dependency, or CI change. The changelog fragment lands in a follow-up commit on this branch.

⚠️ base-red inherited: #12732

### DIFF
--- a/changelog.d/fixes/13202-eval-regex-dotall.md
+++ b/changelog.d/fixes/13202-eval-regex-dotall.md
@@ -1,0 +1,1 @@
+- **fix(evals):** regex grading compiles suite-supplied string patterns with `dotAll`, so correct multi-line answers are no longer marked wrong ([#13202](https://github.com/diegosouzapw/OmniRoute/pull/13202)) — thanks @aaustinhuang
--- a/src/lib/evals/evalRunner.ts
+++ b/src/lib/evals/evalRunner.ts
@@ -153,10 +153,16 @@ export function evaluateCase(evalCase: any, actualOutput: string) {
           details.error = "No regex value provided for evaluation.";
           break;
         }
+        // Suite JSON can only carry a pattern as a string, so the `s` (dotAll) flag has to
+        // be supplied here: without it `.` does not cross a newline, and several built-in
+        // cases join tokens a model naturally puts on separate lines (gs-09 `1.*2.*3.*4.*5`,
+        // code-03 `SELECT.*FROM.*WHERE`, instr-02 `1\..*2\..*3\..*4\..*5\.`). dotAll widens
+        // `.` only, so a pattern that matched before still matches — it is not a regression
+        // risk. An author-supplied RegExp keeps exactly the flags they wrote.
         const regex =
           expectedValue instanceof RegExp
             ? new RegExp(expectedValue.source, expectedValue.flags.replace(/[gy]/g, ""))
-            : new RegExp(expectedValue);
+            : new RegExp(expectedValue, "s");
         if (regex.source.length > 512) {
           passed = false;
           details.error = "Regex pattern too large for safe evaluation.";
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Clean fix with a solid safety argument — `dotAll` only widens what `.` matches, so it can't
regress a pattern that used to pass, and you kept the 512-char guard and the explicit-`RegExp`
flag path intact. Test covers the exact built-in cases named in #13138. Ran it locally: 7/7
pass. Ready to merge as-is.

diegosouzapw:
Thank you for this PR, @dajiaohuang — the dotAll regression in eval regex grading (#13138) was fixed on `release/v3.8.51` by #13413, which compiles both string patterns and explicit `RegExp` instances with the `s` flag. This branch preserved an explicit RegExp's own flags instead, so its test asserts the opposite of what shipped and it can't be merged on top. Closing as superseded; if you believe explicit-RegExp flags should be honoured, a focused issue/PR arguing that case would be welcome.

Issue #13139, opened 2026-09-09
fix(backend): Eval runner cannot opt out of output-style and memory injection — evals measure injected context, not the model


aaustinhuang:
**Version:** 3.8.50

## Summary

The eval runner sends its cases down the ordinary chat path, so every case picks up whatever that path injects — output styles, retrieved memory, and the built-in memory tools. There is no suite-level, case-level or run-level way to turn any of it off, so an evaluation measures the operator's injected context as much as the model.

## What happens

`executeEvalCase()` in `src/lib/evals/runtime.ts` builds its own inbound request with a fixed two-header set and calls the chat route directly:

```js
const headers = new Headers({
  "Content-Type": "application/json",
});
if (apiKey) {
  headers.set("Authorization", `Bearer ${apiKey}`);
}
...
const response = await postChatCompletion(request);
```

Three injections then apply:

| Channel | Trigger | Effect on a graded case |
|---|---|---|
| output styles | `config.enabled && x-omniroute-compression !== "off"` with a non-empty `outputStyles` selection | a system message is prepended and the model answers in that persona instead of answering the case |
| memory context | request carries an API key and memory is enabled | up to `maxTokens` of stored text prepended as a system message |
| memory + skills tools | same, plus `stream !== true` — and eval requests always set `stream: false` | the built-in `memory_*` tools are appended to `tools`; models spend the turn calling them |

The documented opt-outs are request headers — `x-omniroute-compression: off` and `x-omniroute-no-memory: true` — and the eval runner sets neither.

## Reproduction

Seven custom suites, 44 cases, all models reachable, zero call errors. Pass rates recomputed with errored cases forced to failed:

| Run configuration | Pass rate |
|---|---|
| default: one output style selected, no API key on the run so no memory | 89% — a "Simple greeting" case whose expected value is `hello` returns the style's action block with no greeting |
| API key with `compression_enabled = 0` (styles off), memory enabled | 80% — 6 of 8 coding cases return `[No content returned]`, one returns "Let me search my memory for any relevant context about this user first" |
| styles off and memory off | 95% |

The two remaining failures at 95% are genuine model behaviour.

The middle row is the surprising one: passing an API key to a run is what makes results *worse*, because it is what gives the request a memory owner.

## Suggested fix

Have the eval runner mark its own request as self-managed, the same way a client that manages its own context would:

```diff
 const headers = new Headers({
   "Content-Type": "application/json",
+  "x-omniroute-compression": "off",
[...]

PR #13206, opened 2026-09-10
fix(evals): mark eval-runner requests as self-managed so cases measure the model (#13139)

dajiaohuang:
Refs #13139

## What

`executeEvalCase()` built its chat request with only `Content-Type` and `Authorization`, so every graded case picked up the chat path's contextual injections:

| Channel | Trigger | Effect on a graded case |
| --- | --- | --- |
| output styles | `config.enabled && x-omniroute-compression !== "off"` + a non-empty `outputStyles` selection | a system message is prepended; the model answers in that persona |
| memory context | request carries an API key and memory is enabled | up to `maxTokens` of stored text prepended as a system message |
| memory + skills tools | same, plus `stream !== true` (eval requests always send `stream: false`) | the built-in `memory_*` tools are appended; the model spends the turn calling them |

Both injections are already documented request-header opt-outs, and the runner set neither. An evaluation therefore measured the operator's injected context as much as the model — and, per the report, passing an API key to a run made its score *worse*, because the key is what gives the request a memory owner.

## Fix

The runner now marks its own requests as self-managed, so a graded case measures the model:

```diff
 const headers = new Headers({
   "Content-Type": "application/json",
+  "x-omniroute-compression": "off",
+  "x-omniroute-no-memory": "true",
 });
```

Request construction moves to an exported `buildEvalCaseRequest()`. The case executor calls the chat route directly, so this seam is what makes the header contract assertable without invoking the route — the repo's test harness has no `mock.module()` support, so a seam (not a module mock) is the available shape. `executeEvalCase()` now calls it; behaviour is otherwise unchanged.

Scope note: this takes the issue's minimal option. It does **not** add a per-suite "inject anyway" flag — that would be a visible addition to the run record and belongs in its own change.

## Tests

New `tests/unit/evals-runtime-self-managed-headers-13139.test.ts` — 7 cases: both opt-out headers present by default, the opt-out surviving an API-key run (the worst configuration), `Authorization` absent without a key, the POST target, `max_tokens` explicit-vs-512-default, case input/model/`stream: false` reaching the body, and a non-object case input not leaking into it.

```
node --max-old-space-size=8192 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts \
  tests/unit/evals-runtime-self-managed-headers-13139.test.ts
```
→ `pass 7`, `fail 0`. Before the change the same file failed on `buildEvalCaseRequest` being absent.

No regression in the existing eval suites:

```
tests/unit/evals-route.test.ts tests/unit/evals-history.test.ts \
  tests/unit/evalrunner-builtinsuites-split.test.ts
```
→ `pass 13`, `fail 0`.

`npm run check:dead-code` (knip, blocking in CI) → exit 0: the new export is used in-file by `executeEvalCase()`, so the dead-code ratchet is unmoved.

## Checklist

- [x] Tests added for the production-code change (`src/lib/evals/runtime.ts`)
- [x] `lint` / `prettier` clean (husky pre-commit ran lint-staged + `check-docs-sync` + `check:any-budget:t11` + `check:tracked-artifacts`, all green — no hook bypassed)
- [x] No hardcoded secrets; no raw error paths touched
- [x] Changelog fragment added

⚠️ base-red inherited: #12732 — `release/v3.8.51` is not release-green (unit `#12058` alias-row failure, an integration cache-persistence failure, package-artifact provenance, and `check:agent-skills-sync`). None of those touch the eval runner; they are inherited, not introduced here.


diegosouzapw:
Excellent find — a run that carries an API key scoring *worse* because it triggers memory
injection is a genuinely sneaky bug, and the docblock does a great job tracing exactly which
two request-header opt-outs fix it. Extracting `buildEvalCaseRequest()` so the header contract
can be pinned without hitting the real chat route is the right call. Ran the test suite
locally: 7/7 pass. Ready to merge as-is.

diegosouzapw:
Thanks @dajiaohuang — merging via the release merge-train. Validated in local merge-train /tmp/mt-train3b.log on 192.168.0.113 @ train tip 408e2128791696a18966681953136fc96aeb99b0 (32 PRs boarded): static gates green; full test:unit 40694 tests, 17 failing — every one reproduces on the pure release tip (base-red sweep list), zero new reds. Merged --admin per merge-gates §4/§7.

Issue #13194, opened 2026-09-10
fix(resilience): 429 classifier misses non-English quota-exhaustion messages — CJK providers (z.ai/GLM, Kimi, Qwen, MiniMax) stuck in a short rate-limit retry loop


wenzetan:
## Summary

`classify429()` in `src/shared/utils/classify429.ts` decides between `rate_limit`
(short cooldown, retry the same model) and `quota_exhausted` (long cooldown, fail
over) by keyword-matching the 429 body against `QUOTA_PATTERNS`.

Every pattern in that list is **English-only**. Providers that return Chinese
error bodies are therefore silently misclassified as `rate_limit`, even when the
body carries an explicit, absolute quota-reset timestamp.

Observed effect: a provider that is exhausted for the next **~49 minutes** gets a
**6–60 second** cooldown, so the gateway retries a dead model on every request
(3 retries per attempt, per combo hop) instead of failing over to a healthy one.
In a single-model setup the caller just hangs until the retries burn out.

## Environment

- OmniRoute `3.8.51` (Docker Compose deployment)
- Also verified against current `main` — `QUOTA_PATTERNS` there is still English-only
- Provider: z.ai / GLM (`glm-cn`), models `glm-5.3-high` / `glm-5.3-flash`
- (Same class of bug applies to any CJK provider — see below)

## Reproduction

Dropping this next to `src/` and running
`node --experimental-strip-types __tmp_probe_429.ts`:

```ts
import { classify429, looksLikeQuotaExhausted } from "./src/shared/utils/classify429.ts";

const cases: Array<[string, unknown]> = [
  ["GLM/z.ai 5h window (Chinese)",
    "已达到 5 小时的使用上限。您的限额将在 2026-09-10 19:01:19 重置。"],
  ["GLM/z.ai wrapped in error.message",
    { error: { message: "已达到 5 小时的使用上限。您的限额将在 2026-09-10 19:01:19 重置。" } }],
  ["Kimi/Moonshot Chinese quota",
    "您的账户额度已用尽，请充值后重试。"],
  ["Qwen/DashScope Chinese quota",
    "当前账户的免费额度已用完，请前往控制台充值。"],
  ["MiniMax Chinese quota",
    "已达到今日调用上限，请明日再试。"],
  ["English control (OpenAI style)",
    { error: { message: "You exceeded your current quota, please check your plan and billing details." } }],
  ["True transient rate limit (Chinese)",
[...]

PR #13207, opened 2026-09-10
fix(resilience): classify CJK quota-exhaustion 429s as exhausted, not rate-limited (#13194)

dajiaohuang:
Refs #13194

## What

`classify429()` decides between `rate_limit` (short cooldown, retry the same model) and `quota_exhausted` (long cooldown, fail over) by keyword-matching the 429 body. Every entry in `QUOTA_PATTERNS` matched English phrasing, so a CJK provider's explicit quota-exhaustion message fell through to the `return "rate_limit"` default.

The reporter's probe against `3.8.51`:

```
rate_limit       looksLikeQuotaExhausted=false | GLM/z.ai 5h window (Chinese)
rate_limit       looksLikeQuotaExhausted=false | Kimi/Moonshot Chinese quota
rate_limit       looksLikeQuotaExhausted=false | Qwen/DashScope Chinese quota
rate_limit       looksLikeQuotaExhausted=false | MiniMax Chinese quota
quota_exhausted  looksLikeQuotaExhausted=true  | English control (OpenAI style)
rate_limit       looksLikeQuotaExhausted=false | True transient rate limit (Chinese)
```

The English control classifying correctly isolates this to a language-coverage gap, not a logic bug. In production the upstream stated its window reset ~49 minutes later while the gateway scheduled a 6-second cooldown and retried the same exhausted model — 57 s of retries against a model that could not succeed, then a hard failure instead of a failover.

## Fix

Five patterns that name a cap or an exhausted balance:

```diff
   /insufficient balance/i,
+
+  // Chinese (zh-CN) quota-exhaustion phrasing (Issue #13194). Providers:
+  // z.ai/GLM (5-hour and daily windows), Moonshot/Kimi, DashScope/Qwen, MiniMax.
+  /限额将在.*重置/, // "quota will reset at <time>" — z.ai/GLM 5-hour window
+  /使用上限/,      // "usage cap reached" — z.ai/GLM
+  /调用上限/,      // "call cap reached" — MiniMax daily window
+  /额度已用尽/,    // "balance used up" — Moonshot/Kimi
+  /额度已用完/,    // "balance used up" — DashScope/Qwen
 ];
```

Two deliberate boundaries:

- **Not a bare Chinese-throttle match.** `请求过于频繁，请稍后重试。` ("too many requests, retry later") is transient per-minute throttling and must stay a `rate_limit`, or every Chinese throttle would take the long quota bucket. The added phrases are held to cap/balance wording, and a test asserts the transient body still classifies as a rate limit.
- **`QUOTA_PATTERNS` only, not `TERMINAL_QUOTA_PATTERNS`.** These are window-based caps that do clear on a timer, so a declared sub-hour upstream retry hint must still downgrade them — the decision order keeps the declared window authoritative over keywords. A test covers that interaction.

Scope note: this takes the minimal option. It does **not** extend `upstreamRetryDelaySeconds()` to parse absolute reset timestamps (`限额将在 <time> 重置`), which would set a cooldown matching the real window for *every* provider — a larger behavioural change to cooldown math that belongs in its own PR.

## Tests

New `tests/unit/classify429-cjk-quota-13194.test.ts` — 9 cases: the four provider bodies classifying as `quota_exhausted` (both as a bare string and wrapped in `error.message`), the English control still passing, the Chinese transient guard rail, the sub-hour-retry-hint downgrade, and non-CJK bodies unaffected.

```
node --max-old-space-size=8192 --import tsx/esm --test --test-force-exit \
  tests/unit/classify429-cjk-quota-13194.test.ts
```
→ `pass 9`, `fail 0` (8 of them failed before the change).

No regression across the classifier and resilience suites — `classify429`, `combo-breaker-429`, `8396-cooldown-429-cap`, `antigravity-429-*`, `duckduckgo-vqd-429-*`, `grok-web-cloudflare-classification`, `error-classif*`, `rate-limit-local-*`, `claude-oauth-usage-429-backoff`, `persist-429-cooldown-account-fallback`:

→ `pass 159`, `fail 0`.

## Checklist

- [x] Tests added for the production-code change (`src/shared/utils/classify429.ts`)
- [x] `lint` / `prettier` clean (husky pre-commit ran lint-staged + `check-docs-sync` + `check:any-budget:t11` + `check:tracked-artifacts`, all green — no hook bypassed)
- [x] No new ReDoS surface: each added pattern is a literal phrase, `.*` bounded by a literal on both sides
- [x] Changelog fragment added

⚠️ base-red inherited: #12732 — `release/v3.8.51` is not release-green (unit `#12058` alias-row failure, an integration cache-persistence failure, package-artifact provenance, and `check:agent-skills-sync`). None of those touch the 429 classifier; they are inherited, not introduced here.


diegosouzapw:
Thanks for catching this — the CJK quota-exhaustion gap (#13194) is real. A different PR
(#13412) covering the same issue was merged into `release/v3.8.51` on 2026-09-14 with a
broader pattern set (it also adds Japanese and Korean phrasings, not just Chinese), and
all five of your regex are already a subset of what's on the tip now. I'm marking this as
superseded rather than something to merge — no action needed on your end, and thanks for
the report either way!

Issue #6777, opened 2026-07-09
fix(providers): missing SSE [DONE] terminator on ds-web/* models causes client hangs


stgreenb:
## Summary

The `ds-web/*` models (deepseek-v4-pro, deepseek-v4-pro-search, deepseek-v4-pro-think-search, deepseek-v4-flash, deepseek-v4-flash-think) complete their responses with `finish_reason: "stop"` in the last SSE chunk but **do not emit the `[DONE]` SSE terminator line**. This causes OpenAI SDK clients (and OpenCode) to hang waiting for the stream to end, resulting in truncated responses and mid-session drops during multi-turn conversations.

## Reproduction

```bash
curl -X POST http://192.168.1.196:20128/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <key>" \
  -d '{
    "model": "ds-web/deepseek-v4-pro",
    "messages": [{"role": "user", "content": "Say hello and count to 3"}],
    "max_tokens": 50,
    "stream": true
  }'
```

The SSE stream ends with a chunk containing `"finish_reason": "stop"` but **no** `data: [DONE]` line follows. The HTTP connection remains open indefinitely.

## Expected behavior

After the final SSE chunk with `finish_reason: "stop"`, the server should emit:
```
data: [DONE]

```

## Impact

- OpenAI SDK clients hang waiting for `[DONE]` to close the connection
- Multi-turn conversations fail mid-stream
- `stream: false` requests work fine (no SSE, no issue)
- Other providers (via Omniroute) correctly emit `[DONE]`

## Possible fixes

In the SSE relay layer, after detecting `finish_reason: "stop"` in the last upstream chunk, emit `data: [DONE]\n\n` if the upstream did not already send it. This could be:
1. A global fallback (always emit DONE if missing from upstream)
2. A per-model/combo override in payload rules

## Note

This is a DeepSeek Web API upstream behavior — Omniroute is correctly relaying the upstream response. The fix would be in Omniroute\x27s SSE relay to add a safety net for upstreams that omit the `[DONE]` terminator.

diegosouzapw:
[...]

PR #13226, opened 2026-09-10
fix(sse): deepseek-web resilience — premature session close + malformed tool-call recovery

VictorRP7:
## Problem

Two related failure modes on the `deepseek-web` tool-calling path, both observed in
production call logs from real agentic (VS Code Copilot-style) usage of the `deepseek`
combo — the task appears to silently stop or produce garbage instead of completing.

### 1. Premature session close (fixed in the first commit)

`collectSSEContent()` (the non-stream / tool-calling path) drains the upstream SSE body and
returns whatever content it collected as soon as the reader reports `done`, with no check
that DeepSeek actually signalled completion via `response/status: "FINISHED"`.

When the upstream cookie session drops mid-generation — expired session, anti-bot
challenge, network interruption — the HTTP body simply closes early. That was
indistinguishable from a real completion: `execute()` returned `HTTP 200` with
`finish_reason: "stop"` and whatever partial stub text had already arrived (e.g. "I'll
check the weather for you..." with no continuation).

### 2. Malformed tool-call replies (fixed in the second commit)

Separately, even when the session *does* complete normally, DeepSeek's web session
occasionally leaks malformed/internal formatting tokens right after an otherwise-complete
`<tool>{json}` body, instead of a clean `</tool>` close. Observed directly in production:
a fully valid `create_file` JSON tool call immediately followed by corrupted pseudo-tags
(`<｜｜DSML｜｜ parameter>`, `</｜｜DSML｜｜ invoke>`, ...) instead of `</tool>`.
`parseLooseJsonObject`'s strict `JSON.parse` rejected the whole block over that trailing
garbage even though a perfectly valid object sat at the start — so the call was silently
dropped and the raw tagged text shown to the user instead of the file actually being
created.

## Fix

1. `collectSSEContent()` now tracks whether the `FINISHED` status event was actually seen
   (the streaming path already keys its own drain/close logic off this same event — see
   `deepseek-web-done-terminator.ts` / issue #6777 — this path just never checked it). If
   the stream ends without it, it throws instead of silently returning the partial stub;
   `execute()`'s existing `try/catch` turns that into a proper `502`.

2. `deepseekWebTools.ts`: added `salvageLeadingJsonObject()`, a quote/escape-aware
   balanced-brace scanner that recovers just the leading JSON object when the strict parse
   fails — reusing the same salvage idea already used elsewhere in this file
   (`findBareJsonCandidates`) for bare-JSON detection.

3. `deepseek-web.ts`: when even that salvage cannot recover a call (genuinely truncated
   JSON, garbled beyond repair), the `hasTools` branch now detects an unparsed `<tool...>`
   tag surviving in the cleaned content and retries with a brand-new session, bounded to
   `MAX_TOOL_PARSE_ATTEMPTS` (2) — never an unbounded retry loop, and a reply that parses
   cleanly on the first try costs no extra latency.

## Testing

- `tests/unit/deepseek-web-premature-close.test.ts` (2 tests): premature-close now returns
  `502` with an explanatory error; normal `FINISHED` completion is unaffected.
- `tests/unit/deepseek-web-tools-salvage-leading-json.test.ts` (4 tests): recovery from the
  exact production-observed corruption pattern, escaped quotes/nested braces before the
  garbage, correct non-promotion of genuinely truncated JSON, no regression on well-formed
  blocks.
- `tests/unit/deepseek-web-tool-call-retry.test.ts` (3 tests): retry succeeds on a fresh
  session, retry is bounded (gives up after `MAX_TOOL_PARSE_ATTEMPTS` and surfaces the raw
  content rather than looping forever), a clean first reply never triggers a retry.
- Full `deepseek-web` unit suite: `node --import tsx/esm --test tests/unit/deepseek-web*.test.ts`
  → **104/104 passing**, no regressions (streaming, non-streaming, auto-refresh, thinking
  models, search models, all still green).
- `npm run lint` on the changed files: clean (0 new errors; the pre-existing errors reported
  by the run are in an unrelated file, `volcengine-plan-binding-upsert.test.ts`, not touched
  by this PR).
- The salvage fix was additionally verified directly against the exact malformed content
  captured from a live production call log (not just the hand-written test fixture), and
  both fixes were deployed and smoke-tested on a live VPS instance running `omniroute`
  before being ported here.

⚠️ base-red inherited: #12732

### DIFF
--- a/open-sse/executors/deepseek-web.ts
+++ b/open-sse/executors/deepseek-web.ts
@@ -377,6 +377,12 @@ async function collectSSEContent(
   let content = "";
   let reasoningContent = "";
   let currentPath: "thinking" | "content" | "" = "";
+  // Track whether DeepSeek actually signalled completion (`response/status: "FINISHED"`).
+  // Without this, an upstream session drop (expired cookie, anti-bot challenge, network
+  // hiccup) mid-stream was silently reported as a normal "stop" completion with whatever
+  // partial content had arrived so far — e.g. just "I'll check that..." with no follow-up,
+  // HTTP 200, finish_reason "stop". Confirmed in production call logs.
+  let sawFinished = false;
   const streamModel = model || "deepseek-web";
   const thinkingModel = isThinkingModel(streamModel);
   const searchResults: DeepSeekSearchResult[] = [];
@@ -422,6 +428,8 @@ async function collectSSEContent(
         const data = JSON.parse(payload);
         const p = data?.p;
         const v = data?.v;
+
+        if (p === "response/status" && v === "FINISHED") sawFinished = true;
 
         if (v && typeof v === "object" && v.response) {
           if (v.response.thinking_enabled === true) currentPath = "thinking";
@@ -483,6 +491,18 @@ async function collectSSEContent(
 
   const citations = appendSearchCitations(searchResults, streamModel);
   if (citations) content += `\n\n${citations}`;
+
+  // The upstream HTTP body closed without ever sending `response/status: "FINISHED"`.
+  // That means the DeepSeek web session was cut off mid-generation (expired cookie,
+  // anti-bot challenge, network drop, etc.) rather than genuinely completing. Surface
+  // this as an error (caught by execute()'s try/catch -> 502) instead of returning the
+  // partial stub as a successful "stop" response.
+  if (!sawFinished) {
+    throw new Error(
+      "DeepSeek web session ended before completion (no FINISHED signal received) — " +
+        "likely a dropped cookie session or network interruption upstream. Retry the request."
+    );
+  }
 
   return { content, reasoningContent };
 }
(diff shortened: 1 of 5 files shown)


diegosouzapw:
Two genuinely valuable production-hardening fixes, both with clear evidence and tight tests.
Ran all three new test files at your head — all green — and this is the freshest branch in the
whole reasoning-system batch (merge-base is the current tip, zero drift). The `sawFinished`
tracking is exactly the right fix: silently returning a truncated "success" on a dropped
session is worse than a clean 502. Heads up: #13345 touches the same
`deepseekWebTools.ts` file (different functions, verified conflict-free via `git merge-tree`)
— no action needed, just context for the maintainer. Only ask: add a `changelog.d/fixes/`
fragment.

diegosouzapw:
Thanks @VictorRP7 — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

PR #13215, opened 2026-09-10
[URGENT] fix(tests): clear the ESLint base-red on release/v3.8.51 (volcengine any casts)

backryun:
## Summary

Clears the **ESLint base-red on `release/v3.8.51`** that fans out to every open PR as `No new ESLint warnings` → red.

`tests/unit/volcengine-plan-binding-upsert.test.ts` carried three `(bindingTesting as any).upsertConnection(...)` casts. The `__testing.upsertConnection` helper is fully typed (exported from `src/lib/providers/volcenginePlanBinding.ts`), so the casts were dead weight — and with `@typescript-eslint/no-explicit-any` set to **error** under `tests/`, they turned the release tip itself red the moment they merged (originating PR #12950 is already merged, so there is no branch left to fix: this is a base-reds repair PR).

## Change

```diff
- const updated1 = await (bindingTesting as any).upsertConnection(...)
+ const updated1 = await bindingTesting.upsertConnection(...)
```
…same for the two other call sites (lines 70/96/120). No production code, no test semantics changed.

## Validation (on this branch, from the release tip `8ecdd88c1`)

- `npx eslint tests/unit/volcengine-plan-binding-upsert.test.ts --suppressions-location config/quality/eslint-suppressions.json` → **0 errors**
- `node --import tsx/esm --test tests/unit/volcengine-plan-binding-upsert.test.ts` → **3/3 pass**

## Impact

- Fixes 1 of the 2 hard failures reported on base-red **#12732**.
- Un-reds the `No new ESLint warnings` check on every open PR (including #12608, #12759, #13177).

## Notes

- The other base-red items are already owned elsewhere: Stryker tap registration → **#12945**, docs count (172) → merged via **#13209**.
- No release-freeze is active as of opening this PR.

Issue #13144, opened 2026-09-10
fix(backend): redactErrorPaths truncates the message after an API route, destroying the image-model 400's hint (CI red repo-wide)


abhisheksharma2411:
`redactErrorPaths` treats an API route as an unequivocal filesystem path and fails closed over the rest of the message. The visible cost is that the image-model 400 loses the hint it exists to give, and `chat-rejects-image-only-model.test.ts` is currently **red on every open PR in the repo**.

## Reproduction

`src/sse/handlers/chat.ts:655` builds:

```
Model '<m>' is an image-generation model and cannot be used on /v1/chat/completions. Use POST /v1/images/generations instead.
```

What the caller actually receives:

```
Model '<m>' is an image-generation model and cannot be used on <path>
```

The endpoint *and* the entire "Use POST /v1/images/generations instead." sentence are gone — so the error tells the user their model is wrong and then withholds the one thing that would fix it.

## What I measured

```js
redactErrorPaths("used on /v1/chat/completions. Use POST /v1/images/generations instead.")
// -> "used on <path>"
redactErrorPaths("used on /v1/chat/completions. Try again later please.")
// -> "used on <path>"
redactErrorPaths("used on /v1/chat/completions. Contact support.")
// -> "used on <path>"
```

Everything after the route is dropped regardless of what it is, so it isn't the span extending through separator-bearing tokens — it's the fail-closed branch at `errorPathRedaction.ts:624`:

```js
// An unequivocal filesystem prefix with an unknowable endpoint must
// fail closed over the rest of the first line rather than expose a
// suffix such as `Files\\secret` or `My Project`.
parts.push(value.slice(copyStart, index), "<path>");
copyStart = value.length;
```

That behaviour is clearly deliberate and right for `C:\Program Files\...`. The question is why `/v1/chat/completions` reaches it, given `/v1` is not in `POSIX_FILESYSTEM_ROOTS`.

Two more data points that bound it:

```js
// nothing after the route -> endpoint IS determined, trailing punctuation survives
redactErrorPaths("cannot be used on /v1/chat/completions.")
// -> "cannot be used on <path>."

// route-context guard works when the method directly precedes it
redactErrorPaths("Use POST /v1/images/generations instead.")
// -> unchanged
```

So the trigger is a route **not** preceded by an HTTP method or `Route` (here it follows the word "on"), with any whitespace-separated text after it.

And to be clear that the fail-closed posture itself is sound — real paths are still redacted correctly:

[...]

PR #13230, opened 2026-09-10
fix(tests): drop three needless `as any` casts on the volcengine testing surface (base-red #12732)

diegosouzapw:
Fixes the only hard failure in base-red #12732.

## The failure

```
❌ ESLint errors: 3 error(s)

tests/unit/volcengine-plan-binding-upsert.test.ts:70,96,120
  error  @typescript-eslint/no-explicit-any  Unexpected any. Specify a different type
```

## Why it appeared now

The violations are not new. #12795 regenerated `config/quality/eslint-suppressions.json` and **pruned** the entry that had these three frozen — that file shows **0 additions, 5 deletions**, and there is no longer any `volcengine-plan-binding-upsert` key in it. `no-explicit-any` has been `error` in `tests/` since #6218, so a rule the PR never intended to touch started failing the whole-repo lint on code it never edited.

That also explains why #12795's own CI was green: its Lint job ran before the pruned file became the branch's baseline for the next whole-repo run.

## The fix

The casts were unnecessary in the first place:

```ts
// src/lib/providers/volcenginePlanBinding.ts:380-383
export const __testing = {
  findTargetConnection,
  upsertConnection,
};
```

`__testing` exports `upsertConnection` directly, and it is a normally typed exported function — `bindingTesting.upsertConnection(...)` resolves without any help. Dropping `as any` clears the errors **and** puts the three call sites under type checking instead of opting them out, which is strictly better than restoring the suppression entry.

## Validation

| check | before | after |
|---|---|---|
| `eslint` on the file (with suppressions + `--pass-on-unpruned-suppressions`) | exit 1, all 3 errors | exit 0 |
| `npm run typecheck:core` | — | 0 errors |
| `node --test tests/unit/volcengine-plan-binding-upsert.test.ts` | — | 3/3 pass |

Typecheck matters here specifically: removing `as any` is only safe if the now-checked call sites actually satisfy the signature, and they do.

## One deliberate omission

This file is **not** Prettier-clean, and this PR does not reformat it. It was already non-conformant on the base (verified against `HEAD~1`), no CI job enforces Prettier — it only runs through lint-staged on commit — and reformatting would turn a 3-line base-red fix into a 25-line diff across untouched assertions. Left as-is on purpose; happy to split it into a formatting-only PR if you'd rather have it clean.

### DIFF
--- a/tests/unit/volcengine-plan-binding-upsert.test.ts
+++ b/tests/unit/volcengine-plan-binding-upsert.test.ts
@@ -67,7 +67,7 @@ test("volcenginePlanBinding upsert rules with SQLite temp isolation", async (t) 
     providerSpecificData: { autoFetchModels: true, customTag: "keep-me" },
   });
 
-  const updated1 = await (bindingTesting as any).upsertConnection(
+  const updated1 = await bindingTesting.upsertConnection(
     "coding",
     "ark-new-key-1",
     "new-cookie-1",
@@ -93,7 +93,7 @@ test("volcenginePlanBinding upsert rules with SQLite temp isolation", async (t) 
     providerSpecificData: {},
   });
 
-  const createdNew = await (bindingTesting as any).upsertConnection(
+  const createdNew = await bindingTesting.upsertConnection(
     "coding",
     "ark-brand-new-key-3",
     "new-cookie-3",
@@ -117,7 +117,7 @@ test("volcenginePlanBinding upsert rules with SQLite temp isolation", async (t) 
   });
 
   // Passing conn1.id (which is coding-plan) into agent upsert must NOT match conn1
-  const agentUpsertResult = await (bindingTesting as any).upsertConnection(
+  const agentUpsertResult = await bindingTesting.upsertConnection(
     "agent",
     "ark-agent-new-key",
     "agent-cookie",


diegosouzapw:
The red checks on this PR are **inherited from the base**, not caused by this diff. Proof rather than assertion:

This branch changes exactly three lines in one file (`tests/unit/volcengine-plan-binding-upsert.test.ts`). I restored that file to its base version so the working tree was byte-identical to `origin/release/v3.8.51`, and re-ran the two failing suites:

```
--- tree identical to base ---
ℹ tests 13
ℹ pass 11
ℹ fail 2
```

Same two failures, with none of this PR's content present.

The three failing unit tests are `chat-rejects-image-only-model`, `tunnel-routes-error-sanitization` and `combo-provider-cooldown-sibling`. The first two share one root cause: `redactErrorPaths` now swallows the remainder of a message after any POSIX-looking span, which destroys the image-model 400's hint. That is tracked in #13144, where I posted the reproduction and a correction to the proposed boundary.

Worth noting why base-red #12732 does not list these: its most recent verdict came from a **push-mode** run, which executes the quick hard gates only. The full unit suite was never part of that verdict, so these three have been red without appearing in the base-red report.

diegosouzapw:
Duplicata da #13215, que veio antes e é do @backryun. Os diffs são byte a byte idênticos (mesmos três `as any` em `tests/unit/volcengine-plan-binding-upsert.test.ts:70,96,120`).

O crédito pelo fix do base-red de ESLint em `release/v3.8.51` é do @backryun — fechando esta em favor da #13215.

Refs #12732

Issue #12849, opened 2026-09-06
fix(nvidia): NVIDIA provider unusable: static nvidiaHostedModels.snapshot.json allowlist is stale (12 of 14 ids gone upstream)


Ebrahim-Hanna-Sync:
### Summary

Every NVIDIA model call is rejected with

```
Model '<id>' is not available in the active live catalog for provider 'nvidia'.
```

even though the NVIDIA API key is valid. The cause is `open-sse/config/nvidiaHostedModels.snapshot.json`, a hand-maintained 14-entry allowlist baked into the published package. NVIDIA's live catalog has moved on, so 12 of those 14 ids no longer exist upstream, and the two that remain are the only NVIDIA models the router will accept.

**Version:** omniroute 3.8.50 (npm global), Linux, Node 22.23.2

### Evidence

The key itself is fine — querying NVIDIA directly returns 81 models:

```
$ curl -s https://integrate.api.nvidia.com/v1/models -H "Authorization: Bearer $NVIDIA_KEY" | jq '.data | length'
81
```

The shipped snapshot contains 14 ids:

```json
["google/gemma-4-31b-it","minimaxai/minimax-m2.7","mistralai/devstral-2-123b-instruct-2512",
 "mistralai/mistral-large-3-675b-instruct-2512","mistralai/mistral-small-4-119b-2603",
 "nvidia/nemotron-3-super-120b-a12b","openai/gpt-oss-120b","openai/gpt-oss-20b",
 "poolside/laguna-xs-2.1","qwen/qwen3.5-122b-a10b","qwen/qwen3.5-397b-a17b",
 "stepfun-ai/step-3.5-flash","thinkingmachines/inkling","z-ai/glm-5.2"]
```

Intersecting that with the live catalog leaves exactly two usable chat models: `google/gemma-4-31b-it` and `nvidia/nemotron-3-super-120b-a12b`.

Models that are live at NVIDIA today but rejected by the router include `mistralai/mistral-large-2-instruct`, `moonshotai/kimi-k3`, `deepseek-ai/deepseek-v4-pro-0813`, `minimaxai/minimax-m3`, `nvidia/llama-3.1-nemotron-ultra-253b-v1`.

Conversely, `GET /api/v1/providers/nvidia/models` still advertises 49 models — most of which cannot be called — so the model list the client sees and the set the router will actually route to disagree.

### Reproduce

```bash
# advertised by the router
[...]

PR #13248, opened 2026-09-10
fix(nvidia): fail open when a synced model catalog goes stale (#12849)

diegosouzapw:
Refs #12849

## Root cause

The reporter blamed a stale hand-maintained `open-sse/config/nvidiaHostedModels.snapshot.json`
allowlist. That file is not read by any runtime routing path (only by a test and a drift-check
script), and it — plus the real static source
`open-sse/config/providers/registry/nvidia/index.ts` — was already refreshed by #12538 (merged
2026-09-03, before this issue was filed).

The deeper, still-live bug is in `src/sse/services/model.ts::lookupModelMeta`: a provider's
*synced* model catalog (populated once via "Import Models", or automatically only when
`autoFetchModels`/`autoSync` are opted in) is treated as authoritative **forever** once
non-empty — `src/lib/db/models/activeSyncedCatalog.ts::getActiveSyncedCatalog` had no staleness
check, and `src/shared/services/modelSyncScheduler.ts` only re-syncs connections with
`autoSync: true` (off by default). A NVIDIA connection synced once therefore had routing
permanently pinned to that moment's catalog — a model live upstream and present in the current
static registry (e.g. `moonshotai/kimi-k3`) is rejected with `not available in the active live
catalog` indefinitely, with no way to self-recover.

## Fix

- New migration `176_provider_connection_synced_models_at.sql` adds a nullable
  `provider_connections.synced_models_at` timestamp column.
- `replaceSyncedAvailableModelsForConnection` (`src/lib/db/models.ts`) now stamps that timestamp
  via the new `touchConnectionSyncedModelsAt` helper (`src/lib/db/providers.ts`) on every
  successful sync — including an unchanged re-sync, which still proves the catalog is current.
- `getActiveSyncedCatalog` (`src/lib/db/models/activeSyncedCatalog.ts`) now only treats a
  provider's synced catalog as authoritative while at least one active connection was synced
  within `OMNIROUTE_SYNCED_CATALOG_STALE_AFTER_MS` (default 30 days, env-overridable). Once every
  contributing connection is stale — or was never synced (pre-migration rows: `NULL` timestamp) —
  the catalog fails open exactly like the existing no-sync-yet path, instead of gating on a
  frozen point-in-time snapshot forever.

This makes the runtime resilient to a stale synced catalog rather than freezing yet another
point-in-time list, matching the issue's own suggested fix direction (refresh at
sync/connection-test time instead of trusting a frozen snapshot forever).

## Regression test

`tests/unit/nvidia-stale-synced-catalog-12849.test.ts` — RED→GREEN evidence:

- **RED** (against release/v3.8.51 tip, before this fix — reverted locally via `git show
  HEAD:<path>`, never `git stash`): all 5 assertions failed; the DB did not even have the
  `synced_models_at` column yet.
- **GREEN** (after the fix): all 5 pass —
  1. a fresh synced catalog still gates (unchanged behavior)
  2. a stale (45-day-old) synced catalog fails open — the live+registry model resolves again
  3. `getActiveSyncedCatalog` reports `authoritative: false` once stale
  4. a connection that was never synced (`NULL` timestamp) is non-authoritative, not gated forever
  5. `OMNIROUTE_SYNCED_CATALOG_STALE_AFTER_MS` overrides the default threshold

Existing suites re-run clean against the fix: `tests/unit/live-model-catalog-reconciliation-8926.test.ts`,
`tests/unit/custom-models-live-catalog-12597.test.ts`, `tests/unit/chat-routing-synced-inventory-11089.test.ts`,
`tests/unit/models-catalog-static-synced-suppression.test.ts`, `tests/unit/github-live-catalog-combo-12137.test.ts`,
`tests/unit/codex-synced-bare-model-routing.test.ts`, `tests/unit/db-models-crud.test.ts`,
`tests/unit/db-models-split.test.ts`, `tests/unit/managed-model-import.test.ts` all pass.
`tests/unit/12058-models-catalog-canonical-self-aliased.test.ts` and one `model-sync-route.test.ts`
case are flaky/failing identically on the unmodified base tip (confirmed by reverting this branch's
changes and re-running) — pre-existing, not introduced here.

## Gates run

- `node scripts/check/check-complexity.mjs` — OK (2798 violations, baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1265 violations, baseline 1437)
- Full new test file + related existing synced-catalog/model-routing suites — see above

## Scope note (deliberately not addressed here)

The triage plan-file for #12849 also flagged two secondary, separate diagnostics gaps
(`omniroute providers test nvidia` / `providers status` for apikey providers, and a 504 on
`google/gemma-4-31b-it` under cold start) — explicitly called out as smaller, likely-separate
tickets, not conflated with this fix. A default-on proactive re-sync at connection-test time
(the plan's checkbox #2) is left as a follow-up enhancement; this PR's staleness-based fail-open
already removes the "permanently pinned" failure mode reported in the issue.


HouMinXi:
Thanks for failing open when the NVIDIA synced catalog goes stale (#12849). I folded this into a local recut of `release/v3.8.51`.

Conflict was one hunk in `src/lib/db/models.ts` (`replaceSyncedAvailableModelsForConnection`): HEAD notified quota combos after a successful sync. Resolution: keep both — `notifyQuotaCombosForProvider`, then `touchConnectionSyncedModelsAt`. Migration 176 and `activeSyncedCatalog.ts` came through unchanged.

Purpose of the recut: ship current `release/v3.8.51` plus open product PRs to a private production host. If you rebase onto current `release/v3.8.51`, please keep the quota-combo notify next to the stamp.

diegosouzapw:
Held out of today's batch, and the cause is mine.

This branch adds `src/lib/db/migrations/176_provider_connection_synced_models_at.sql`. A few hours ago I merged #12651, and while reconciling it I renumbered its migration to **176** — it had been written as 173 and collided with `173_call_logs_video_content_removed.sql` that landed after it was opened. So 176 is now taken on `release/v3.8.51`, and this branch collides with it.

The failure is loud but misattributed: `migrationRunner` aborts on every `getDbInstance()`, so it takes down every DB-touching test, not just this PR's. Measured on the tip with only this branch applied:

```
26 failures across 9 files — a2a routes, cli-tools keys, feature flags,
image generation auth, playground key policy, error sanitization, semantic cache
```

None of those are your change. The same set passes on the tip, and the other 10 worktree-held PRs merged clean alongside each other.

The fix is one rename plus the header line inside the file:

```
git mv src/lib/db/migrations/176_provider_connection_synced_models_at.sql \
       src/lib/db/migrations/177_provider_connection_synced_models_at.sql
sed -i 's/^-- Migration 176:/-- Migration 177:/' \
       src/lib/db/migrations/177_provider_connection_synced_models_at.sql
```

177 is free on the current tip (173 through 176 are taken). No retroactive `isSchemaAlreadyApplied` guard is needed since the table never shipped under 176.

Two things to remember when you land it: the migration count goes from 173 to 174, which `check:docs-counts` enforces across `README.md`, `AGENTS.md`, `llm.txt` and its 50 i18n mirrors (206 occurrences); and `AGENTS.md` plus `llm.txt` are agent-instruction surfaces, so that bump needs an explicit nod before it goes in.

I did not push the rename myself: this branch has a live worktree with a commit from a few hours ago, and pushing onto it would pull the branch out from under that session. The other 11 held PRs are merged.

PR #13274, opened 2026-09-10
fix(logging): mask every *-api-key provider header in the log pipeline (#13273)

### LINKED ISSUE #13273: fix(backend): request-log header mask misses *-api-key spellings, provider keys land in log pipeline unmasked
## Summary

`maskSensitiveHeaders` in `open-sse/utils/requestLogger.ts` matches a hardcoded substring list (`authorization`, `x-api-key`, `cookie`, `token`, ...). Real upstream credential headers that don't contain those substrings pass through completely unmasked into the request-log pipeline (`logTargetRequest` -> `getPipelinePayloads`):

- `x-goog-api-key` (Gemini / Vertex via the default executor)
- `api-key` (Azure OpenAI)
- `xi-api-key` (ElevenLabs proxy header)

Repro (drives the public logger exactly like `chatCore.ts` does before a Gemini dispatch):

```ts
const logger = await createRequestLogger("openai", "gemini", "gemini-2.5-flash", {});
logger.logTargetRequest(url, { "x-goog-api-key": KEY }, body);
logger.getPipelinePayloads()?.providerRequest?.headers // -> { "x-goog-api-key": KEY } verbatim
```

`x-api-key` (Anthropic) is masked correctly; the two above are stored verbatim. The persisted call-log artifact layer (`protectPipelinePayloads`/`redactPayload`) covers `x-goog-api-key`/`api-key` today, so this is defense-in-depth for the in-memory pipeline plus any direct consumer of it — but `xi-api-key` misses both layers.

## Suggested fix

1. Match the compacted header name against `apikey` in `maskSensitiveHeaders` so every current and future `*-api-key` spelling is masked (same partial-mask convention as `x-api-key`).
2. Add `xi-api-key` to `SENSITIVE_KEYS` in `src/lib/logPayloads.ts`.
3. Regression test through `createRequestLogger` + `protectPayloadForLog`.

soroush5:
Fixes #13273.

The request-log header mask only knew x-api-key, so Gemini (x-goog-api-key) and Azure (api-key) provider keys were stored verbatim in the log pipeline. This matches any *-api-key spelling instead, same masking style as before. Adds xi-api-key to the persisted-log redaction too, plus a regression test.

### DIFF
--- a/changelog.d/fixes/13274-log-apikey-header-redaction.md
+++ b/changelog.d/fixes/13274-log-apikey-header-redaction.md
@@ -1,0 +1,1 @@
+- Mask every `*-api-key` provider header spelling (`x-goog-api-key`, Azure `api-key`, `xi-api-key`) in the request-log pipeline and persisted call-log redaction instead of only `x-api-key`. (#13274 — thanks @soroush5)
--- a/open-sse/services/autoCombo/__tests__/autoCombo.test.ts
+++ b/open-sse/services/autoCombo/__tests__/autoCombo.test.ts
@@ -261,6 +261,9 @@ describe("Mode pack ranking gates (cold/warm/health)", () => {
   }
   it("cold pool ranking unchanged (reliability 1, quality 0.5 neutrals)", () => {
     const a: ProviderCandidate = {
+      provider: "test",
+      model: "test-model",
+      errorRate: 0,
       circuitBreakerState: "CLOSED",
       failureRate: undefined,
       quality: undefined,
@@ -272,6 +275,9 @@ describe("Mode pack ranking gates (cold/warm/health)", () => {
       latencyStdDev: 10,
     };
     const b: ProviderCandidate = {
+      provider: "test",
+      model: "test-model",
+      errorRate: 0,
       circuitBreakerState: "CLOSED",
       failureRate: undefined,
       quality: undefined,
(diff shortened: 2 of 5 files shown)


soroush5:
CI status note — I rebased onto current `release/v3.8.51` and verified each remaining red gate locally against pristine `af49d497`:
- The autoCombo TS2739 is fixed in this branch (open-sse typecheck gate green, redaction tests 4/4, logger suites 101/101).
- Everything else still red reproduces on pristine main: skills sync wants `cli-tunnel` regenerated, file-size flags `stream.ts`, docs-all reports 7 drifts, ESLint flags the untouched `volcengine-*.test.ts`, and `httpAuthContext.test.ts` fails there too. None of these touch this PR's files, so they look like upstream drift rather than regressions from this change.

diegosouzapw:
Thanks for the careful work here, including the CI-status note pinning down which reds were
pre-existing base drift vs. this PR's own — that kind of diligence makes review much faster.
Since you opened this, #13401 landed (2026-09-14) fixing the exact same gap for the exact same
issue (#13273), which is now closed: it adds `xi-api-key` to the same redaction set and masks
every `*-api-key` header spelling in the request-log pipeline, just via a slightly different
matching expression with the same observable effect. Closing this one as subsumed by #13401.

PR #13276, opened 2026-09-10
fix(routing): stop reactive-compaction log from lying when compression is disabled (#11977)

### LINKED ISSUE #11977: fix(providers): Antigravity upstream 400 on every request once session history grows
### OmniRoute Version

v3.8.50

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Debian 13 Trixie

### Node.js Version

latest stable version

### Provider(s) Involved

Antigravity

### Model(s) Involved

Gemini 3.7 Flash

### Client Tool

Claude Code

### Description

getting this error message. when i just set it up it worked flawlessly but suddenly out of nowhere after sometime it is throwing this error message on every message i try to send. even a simple Hi. (i have enabled rate limit protection) through OmniRoute dashboard.

Image

Image

### Steps to Reproduce

Added a provider via antigravity cli auth. 
Setup claude code profile with agy models.
Tried using those models inside claude code.
Firstly they worked flawlessly, but suddenly out of nowhere getting api error 400

### Expected Behavior

I didnt change any setting the models should work normally as they were working before.

### Actual Behavior

Getting API Error: 400 Antigravity upstream error (400)

### Test Impact

Unsure

### Error Logs / Output

```shell
❯ /statusline configure the git status, currently it only shows the current git branch name, in case of uncommitted changes it should show a red       
aestrisk symbol next to branch name. and when no uncommited changes then no aestrisk symbol.                                                           
  
● API Error: 400 Antigravity upstream error (400)

✻ Sautéed for 8s · done 11:12 am

❯ hi                                                                                                                                                   

● API Error: 400 Antigravity upstream error (400)

✻ Churned for 2s · done 11:31 am
```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

diegosouzapw:
Closes #11977

## Root cause

On a default install `DEFAULT_COMPRESSION_CONFIG.enabled = false`
(`open-sse/services/compression/types.ts:418`), and PR #9200 deliberately tied the
built-in reactive/last-resort context-compaction safety net to that same global
toggle (`reactiveContextCompactionEnabled = compressionSettingsResult.enabled &&
!compressionExcluded;`, `open-sse/handlers/chatCore.ts:1361`). With compression off
by default, nothing trims a growing session history before it is forwarded upstream,
and Antigravity's real `daily-cloudcode-pa.googleapis.com` request-size ceiling is far
below the advertised Gemini 3.7 context window, so requests eventually get a bare 400
once history has grown — reproducing on every subsequent turn.

**Scope of this fix** (per the analyzed plan-file, main-thread-audited): the
code-provable, safely-fixable defect is the diagnostic log in `chatCore.ts` that fired
unconditionally whenever `promptCompressionEnabled` was false, claiming "reactive
context compaction still applies when over threshold" — even in the default-install
branch where `reactiveContextCompactionEnabled` is ALSO false, i.e. no safety net is
actually active. Operators got zero diagnostic signal for the real failure mode. This
PR does **not** touch the #9200 gating itself (deliberate fix for tool-call-history
corruption on Pi/Codex) and does **not** attempt to guess a lower per-provider request
cap for Antigravity — that remains a separate, larger design question the owner
already flagged as unresolved.

## Fix

`open-sse/handlers/chatCore.ts`: the `if (!promptCompressionEnabled)` debug-log block
now branches on `reactiveContextCompactionEnabled` — the existing message is kept when
reactive compaction is still active (per-key opt-out case), and a new distinct message
("...reactive context compaction is ALSO disabled — large histories will NOT be
trimmed before reaching the upstream provider") is logged for the default-install case
this issue is about.

`tests/unit/chatcore-stale-compaction-log.test.mjs` (new file, reusing the proven
plan-file probe):

RED (before fix):
```
✖ BUG: the log statement does not branch on reactiveContextCompactionEnabled, so it cannot ever be accurate
  AssertionError [ERR_ASSERTION]: expected the debug-log block guarded by "!promptCompressionEnabled"
  to also branch on reactiveContextCompactionEnabled ...
tests 3 / pass 2 / fail 1
```

GREEN (after fix):
```
✔ gating expressions exist as documented (sanity check, tracks real source)
✔ on default install, reactiveContextCompactionEnabled is provably false whenever the log fires
✔ the log statement branches on reactiveContextCompactionEnabled so it stays accurate in both cases
tests 3 / pass 3 / fail 0
```

Existing regression test for the #9200 gating itself (untouched, still green):
`node --test tests/unit/reactive-context-compaction-policy.test.mjs` → 1/1 pass.
Adjacent test also re-run and unaffected: `tests/unit/codex-prompt-compression-passthrough.test.ts` → 3/3 pass.

## Gates run (all green)

- `node scripts/check/check-file-size.mjs` → OK (chatCore.ts stays under its frozen ceiling)
- `node scripts/check/check-complexity.mjs` → OK (2798 violations vs baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` → OK (1265 violations vs baseline 1437)
- `npm run typecheck:core` → exit 0
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json open-sse/handlers/chatCore.ts tests/unit/chatcore-stale-compaction-log.test.mjs` → exit 0
- New test: `node --test tests/unit/chatcore-stale-compaction-log.test.mjs` → 3/3 pass
- Existing adjacent tests re-run: `reactive-context-compaction-policy.test.mjs` (1/1), `codex-prompt-compression-passthrough.test.ts` (3/3)

### DIFF
--- a/changelog.d/fixes/11977-antigravity-stale-compaction-log.md
+++ b/changelog.d/fixes/11977-antigravity-stale-compaction-log.md
@@ -1,0 +1,1 @@
+- fix(routing): stop the reactive-compaction debug log from lying when compression is globally disabled (#11977)
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -1958,7 +1958,9 @@ export async function handleChatCore({
     if (!promptCompressionEnabled) {
       log?.debug?.(
         "CONTEXT",
-        "Prompt Compression engines disabled; reactive context compaction still applies when over threshold"
+        reactiveContextCompactionEnabled
+          ? "Prompt Compression engines disabled; reactive context compaction still applies when over threshold"
+          : "Prompt Compression engines disabled; reactive context compaction is ALSO disabled — large histories will NOT be trimmed before reaching the upstream provider"
       );
     }
     if (isCombo && comboName) {
--- a/tests/unit/chatcore-stale-compaction-log.test.mjs
+++ b/tests/unit/chatcore-stale-compaction-log.test.mjs
@@ -1,0 +1,71 @@
+// Regression guard for #11977: the diagnostic log fired unconditionally whenever
+// promptCompressionEnabled was false, claiming "reactive context compaction still
+// applies when over threshold" even on a default install where
+// reactiveContextCompactionEnabled is ALSO false (DEFAULT_COMPRESSION_CONFIG.enabled
+// === false, per #9200's gating). That left operators with zero diagnostic signal for
+// the real failure mode: large histories reaching the upstream provider untrimmed
+// (e.g. Antigravity's 400 once session history grows past its real request-size
+// ceiling). The fix branches the log on reactiveContextCompactionEnabled so the
+// message reflects which safety net, if any, is actually still active.
+import assert from "node:assert/strict";
+import test from "node:test";
+import { readFileSync } from "node:fs";
+
+const source = readFileSync(
+  new URL("../../open-sse/handlers/chatCore.ts", import.meta.url),
+  "utf8"
+);
+
+test("gating expressions exist as documented (sanity check, tracks real source)", () => {
+  assert.match(
+    source,
+    /let promptCompressionEnabled =\s*\n\s*compressionSettingsResult\.enabled && !compressionExcluded && apiKeyCompressionEnabled;/
+  );
+  assert.match(
+    source,
+    /reactiveContextCompactionEnabled = compressionSettingsResult\.enabled && !compressionExcluded;/
+  );
+});
+
+test("on default install, reactiveContextCompactionEnabled is provably false whenever the log fires", () => {
+  const compressionSettingsResultEnabled = false; // DEFAULT_COMPRESSION_CONFIG.enabled
+  const compressionExcluded = false;
+  const apiKeyCompressionEnabled = true;
+
+  const promptCompressionEnabled =
+    compressionSettingsResultEnabled && !compressionExcluded && apiKeyCompressionEnabled;
+  const reactiveContextCompactionEnabled = compressionSettingsResultEnabled && !compressionExcluded;
+
+  const logFires = !promptCompressionEnabled;
+
+  assert.equal(logFires, true, "the log fires on every default-config request");
+  assert.equal(
+    reactiveContextCompactionEnabled,
+    false,
+    "reactive compaction is ALSO disabled here — the old unconditional message was false for this branch"
+  );
+});
+
+test("the log statement branches on reactiveContextCompactionEnabled so it stays accurate in both cases", () => {
+  const blockMatch = source.match(
+    /if \(!promptCompressionEnabled\) \{[\s\S]{0,400}\}/
+  );
+  assert.ok(blockMatch, "expected to find the promptCompressionEnabled debug-log block");
+  const block = blockMatch[0];
+
+  assert.match(
+    block,
+    /reactiveContextCompactionEnabled/,
+    "expected the debug-log block to branch on reactiveContextCompactionEnabled"
+  );
+  assert.match(
+    block,
+    /still applies when over threshold/,
+    "expected the true-branch message (reactive compaction still active) to be preserved"
+  );
+  assert.match(
+    block,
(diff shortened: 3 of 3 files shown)


pm25coder:
Two things from reading the head of the branch, both additive rather than scope creep.

The branch now tells the truth, but only at debug level (`log?.debug?.("CONTEXT", ...)`, `chatCore.ts:1958-1965`). In the default-install branch that line is the only place an operator could have learned that nothing trims history — and a debug line they never enabled is not much better than a wrong line they never read. The failure the issue describes arrives as a bare provider 400 on every subsequent turn, so the observable signal still lives only at the provider boundary. If it is cheap: keep this honest debug line, and add a warn-level once-per-session line at the moment the upstream rejects with a size-shaped error, carrying the prompt size you sent. Then someone who enabled nothing still gets told when it matters.

Second, `DEFAULT_COMPRESSION_CONFIG.enabled = false` (`types.ts:419-421`) means the effective ceiling on the Antigravity path is not the advertised window at all — it is an undocumented per-provider request limit. Keeping that out of this PR is right. But the shape of the eventual fix is worth deciding now: a configured window will always be wrong for exactly the providers you cannot test, while the largest prompt size that actually returned 200 for that provider/model is evidence you already have in the request path. Anchoring the last-resort threshold on that observed maximum — and failing loud, not silent, while there is no observation yet — turns the number from a catalogue value into a measurement. It is also the cheapest possible thing to log today: one field per request, no new config, no routing change.

HouMinXi:
Thanks for fixing the reactive-compaction log when compression is off (#11977). I folded this into a local recut of `release/v3.8.51` after the chatCore extract (#13065).

Conflict was `open-sse/handlers/chatCore.ts`: incoming carried the pre-extract ~6k-line body. The log now lives in `open-sse/handlers/chatCore/cacheAndCompress.ts`. Resolution: kept the extract; transplanted the branched CONTEXT log and pointed `tests/unit/chatcore-stale-compaction-log.test.mjs` at that file. Widened the source-scan window so both branches fit.

Purpose of the recut: ship current `release/v3.8.51` plus open product PRs to a private production host. If you rebase onto current `release/v3.8.51` after #13065, please edit `cacheAndCompress.ts` rather than the old `chatCore.ts` compression block.

pm25coder:
Worth one more note before this rebases, because the guard is the part that will not survive the merge intact.

`tests/unit/chatcore-stale-compaction-log.test.mjs` passes on the head of this branch, but I ran it locally against two mutants of the block it guards and it passed on both:

1. **The ternary arms swapped** — i.e. exactly the regression this PR exists to prevent, where the fully-disabled case is told "reactive context compaction still applies when over threshold" again. All three of its assertions still match: the two message strings are still inside the block, so a regex over the block cannot see which branch emits which.
2. **The block made unreachable** (`if (false) { ... }`) — still green. The scan reads text; it cannot tell a reachable block from a dead one.

The second test is where the guard is thinnest: it re-declares `compressionSettingsResultEnabled` / `compressionExcluded` / `apiKeyCompressionEnabled` locally and asserts arithmetic on those local constants, so production never executes — if the real gating expression gains a conjunct, that test keeps passing. And the third test's `[\s\S]{0,400}` window fails loudly for a reason that has nothing to do with behaviour when the block is reflowed or renamed, which is the same fragility you hit as a conflict rather than as a signal.

Your repo already has a gate that names this class — `scripts/check/check-test-discovery.mjs`, whose own comment calls a test that never runs "the falso verde definitivo". This is that class one level up: a test that runs and cannot fail. Since `log` is injected and only `debug`/`warn` are stubbed, capturing the emitted message for the two configurations is a few lines, and the source scan can stay as the sanity check it advertises itself to be ("tracks real source").

One forward-looking note, since #13065 splits this area: that PR adds `open-sse/handlers/chatCore/cacheAndCompress.ts`, so the block ends up out of `chatCore.ts` and this guard's `readFileSync` path has to move with it. If a rebase ever leaves the old body behind while the live path is in the new module, the scan passes against the dead copy — which is mutant 2 above, arrived at by merge rather than by hand. `tests/unit/chatcore/cache-and-compress.test.ts` in that PR is the natural home for the behavioural assertion.

PR #13278, opened 2026-09-10
fix(sse): require Responses-shaped body before native OpenAI-compatible passthrough (#12129)

### LINKED ISSUE #12129: fix(backend): internal context-handoff summary request is sent in Chat Completions shape to responses-type providers, upstream receives zero messages
## OmniRoute Version

3.8.50 (docker `diegosouzapw/omniroute:latest`). I also reviewed current HEAD source — `summaryBody` is still built in Chat Completions shape there, so this is unfixed.

## Installation Method

Docker / Docker Compose (Linux host)

## What happened?

When context usage crosses the handoff threshold, the internal context-handoff summarizer request is built in **Chat Completions** shape and dispatched to the summarizer model's provider without per-target format translation. When the summarizer model resolves to a provider whose API type is `responses` (a custom `openai-compatible-responses-*` connection), the raw chat-shaped body is POSTed to the upstream `/responses` endpoint with a `messages` field and no `input`.

Upstreams that don't map `messages` on `/v1/responses` then see an **empty input**. In our case (CLIProxyAPI v7.2.145 in front of an OpenAI-compatible provider) the upstream received a chat request with **zero messages** and rejected it:

```
[500]: upstream 400: {"code":11133,"msg":"the request parameters were rejected by the model provider","extError":{"code":"400002"}}
```

Minimal repro against the same upstream: POST `{"model":"glm-5.3-flash","messages":[{"role":"user","content":"Say ok"}],"stream":false}` to `/v1/responses` → `400 "Input required: specify \"prompt\" or ..."` — i.e. the `messages` field is simply ignored.

## Where it comes from

`open-sse/services/contextHandoff.ts`:

```ts
const summaryBody = {
  model: summaryModel,
  messages: [{ role: "user", content: summaryPrompt }],
  stream: false,
  max_tokens: DEFAULT_SUMMARY_RESPONSE_TOKENS,
  temperature: 0.1,
  _omnirouteSkipContextRelay: true,
  _omnirouteInternalRequest: "context-handoff",
};
const response = await options.handleSingleModel(summaryBody, summaryModel);
```

with `summaryModel = relayConfig.handoffModel || options.model` — so the summary request inherits whatever provider the main request used, including responses-type custom connections. In my capture the outbound body still had `messages` (2 entries), `max_output_tokens` (renamed), and the internal markers `_omnirouteInternalRequest` / `_omnirouteSkipContextRelay` / `_nativeOpenAICompatibleResponsesPassthrough` — confirming no chat→responses translation ran for this internal request.

## Impact

Silent failure (`if (!response.ok) return;`): context handoff summaries are never generated for any session whose main model lives on a responses-type connection. Only visible as upstream 400/500 noise in logs. Main conversation traffic is unaffected.

## Steps to Reproduce

1. Configure a custom provider with API type `responses` (e.g. backed by CLIProxyAPI) and route a long-running combo/model through it so context usage crosses the handoff threshold.
2. Leave Combo Defaults → handoffModel empty (defaults to the current model).
3. Observe the upstream receiving a zero-message request and returning 400; no handoff summary is stored.

## Expected behavior

Internal requests should be translated to the target provider's wire format, exactly like external requests.

## Suggested fix

- Route the internal summary request through the same request-format translation pipeline as external requests, or
- build `summaryBody` according to the target provider's API type (`input` items for responses-type providers), or
- as a guard, skip/fall back to a chat-format target when the resolved handoff model's provider is responses-type and translation is unavailable.

## Workaround

Set Combo Defaults → handoffModel to a model on a chat-type provider.

## Additional context

- Request flow: opencode 1.18.25 → OmniRoute (combo) → CLIProxyAPI v7.2.145 → OpenAI-compatible upstream.
- Happy to provide full captured request/response logs if useful.

diegosouzapw:
Closes #12129

## Root cause

The internal context-handoff / universal-handoff summary request is built in Chat Completions
shape (`messages`, no `input`) and dispatched through the *same* `handleSingleModel` closure
that still carries the *original client request's* endpoint (`src/sse/handlers/chat.ts`,
reused by `maybeGenerateHandoff` in `open-sse/services/combo.ts`).

`resolveChatCoreRequestFormat` derives `sourceFormat` purely from that inherited endpoint path
via `detectFormatFromEndpoint` (`open-sse/services/provider.ts`), which returns
`"openai-responses"` for any `/responses` path unconditionally — it never inspects body shape
on that branch. `shouldUseNativeOpenAICompatibleResponsesPassthrough`
(`open-sse/handlers/chatCore/passthroughHelpers.ts`) then decided to skip all
chat→responses translation purely from `sourceFormat === "openai-responses"` +
`providerSpecificData.apiType === "responses"`, with no way to tell a genuinely
Responses-shaped client body apart from an internally-synthesized chat-shaped one. The upstream
ended up receiving the raw `messages` array on `/v1/responses` with no `input`, and rejected it.

## Fix

`shouldUseNativeOpenAICompatibleResponsesPassthrough` now takes the request `body` and requires
it to actually look Responses-shaped (`input` present, `messages` absent) before allowing the
native-passthrough fast path — mirroring the shape check `detectFormatFromEndpoint` already
applies on its chat/completions branch. The call site in `open-sse/handlers/chatCore.ts` now
passes `body` through. A genuine client Responses-API request still gets the zero-translation
shortcut; the internally-synthesized chat-shaped handoff body is now routed through the normal
chat→responses translation layer instead of being hand-forwarded.

## Regression test

`tests/unit/context-handoff-native-passthrough-bug.test.ts`

- RED (pre-fix, code temporarily reverted to confirm): `nativePassthrough` returned `true` for
  the chat-shaped internal handoff body — assertion failed with `true !== false`.
- GREEN (post-fix): same case now returns `false`; a companion case confirms a genuine
  Responses-shaped body (`input`, no `messages`) still returns `true`, protecting the legitimate
  passthrough optimization for real Responses-API clients.

## Gates run

- `node scripts/check/check-file-size.mjs` — no violations on touched files
- `node scripts/check/check-complexity.mjs` — OK (2798 violações, baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1265 violações, baseline 1437)
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json <changed files>` — exit 0, no output
- New regression test: 2/2 passing
- Existing tests of the touched area (all passing): `tests/unit/responses-passthrough-openai-compatible.test.ts`,
  `tests/unit/chatcore-request-format.test.ts`, `tests/unit/xai-agent-tools-passthrough.test.ts`,
  `tests/unit/anthropic-thinking-signature-recovery.test.ts` (40 tests),
  `tests/unit/universal-handoff.test.ts` (24), `tests/unit/context-handoff.test.ts` (19),
  `tests/unit/service-context-handoff.test.ts` (11), `tests/unit/chat-context-relay.test.ts` (2),
  `tests/unit/combo-context-relay.test.ts` (13)

### DIFF
--- a/changelog.d/fixes/12129-context-handoff-native-passthrough-shape.md
+++ b/changelog.d/fixes/12129-context-handoff-native-passthrough-shape.md
@@ -1,0 +1,1 @@
+- fix(sse): require Responses-shaped body before native OpenAI-compatible passthrough (#12129)
--- a/open-sse/handlers/chatCore.ts
+++ b/open-sse/handlers/chatCore.ts
@@ -731,6 +731,7 @@ export async function handleChatCore({
       sourceFormat,
       endpointPath,
       providerSpecificData: credentials?.providerSpecificData,
+      body,
     });
   const responsesInputItems = Array.isArray(body?.input) ? body.input : [];
   const customToolNames = collectCustomToolNamesForSourceFormat(
--- a/open-sse/handlers/chatCore/passthroughHelpers.ts
+++ b/open-sse/handlers/chatCore/passthroughHelpers.ts
@@ -53,19 +53,35 @@ export function stampNativeResponsesPassthroughBody(
   return { ...body, _nativeOpenAICompatibleResponsesPassthrough: true };
 }
 
+// A body only qualifies for the native-Responses passthrough fast path when it is
+// actually shaped like a Responses API request (`input`, no `messages`). Endpoint
+// path alone is not sufficient: an internally-synthesized Chat Completions-shaped
+// body (e.g. the context-handoff summary request) can be dispatched through a
+// closure that still carries the original client request's `/responses` endpoint,
+// which otherwise makes `sourceFormat` resolve to "openai-responses" even though
+// the body itself was never translated. See issue #12129.
+function isResponsesShapedBody(body: unknown): boolean {
+  if (!body || typeof body !== "object") return false;
+  const candidate = body as Record<string, unknown>;
+  return candidate.input !== undefined && candidate.messages === undefined;
+}
+
 export function shouldUseNativeOpenAICompatibleResponsesPassthrough({
   provider,
   sourceFormat,
   endpointPath,
   providerSpecificData,
+  body,
 }: {
   provider?: string | null;
   sourceFormat?: string | null;
   endpointPath?: string | null;
   providerSpecificData?: unknown;
+  body?: unknown;
 }): boolean {
   if (!provider?.startsWith("openai-compatible-")) return false;
   if (sourceFormat !== FORMATS.OPENAI_RESPONSES) return false;
+  if (body !== undefined && !isResponsesShapedBody(body)) return false;
   if (providerSpecificData && typeof providerSpecificData === "object") {
     const psd = providerSpecificData as Record<string, unknown>;
     if (psd.apiType === "responses" || psd._omnirouteForceResponsesUpstream === true) {
--- a/tests/unit/context-handoff-native-passthrough-bug.test.ts
+++ b/tests/unit/context-handoff-native-passthrough-bug.test.ts
@@ -1,0 +1,86 @@
+// Regression test for issue #12129: an internal context-handoff summary request (built in
+// Chat Completions shape -- `messages`, no `input`) is dispatched through the SAME
+// handleSingleModel closure that carries the ORIGINAL client request's endpoint.
+// When that original endpoint matched `/responses` and the resolved handoff-model
+// provider is an openai-compatible-* connection configured with apiType "responses",
+// the pipeline used to decide the body was already native-Responses-shaped and skip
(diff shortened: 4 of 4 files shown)


HouMinXi:
Thanks for requiring a Responses-shaped body before native OpenAI-compatible passthrough (#12129). I folded this into a local recut of `release/v3.8.51` after the chatCore extract (#13065).

Conflict was `open-sse/handlers/chatCore.ts` (extract vs pre-extract body). The guard belongs in `open-sse/handlers/chatCore/passthroughHelpers.ts` (`isResponsesShapedBody`). Resolution: kept the extract; took the helper and test as-is.

Purpose of the recut: ship current `release/v3.8.51` plus open product PRs to a private production host. If you rebase onto current `release/v3.8.51` after #13065, please keep the change in `passthroughHelpers.ts`.

Issue #7993, opened 2026-07-21
fix(providers): noauth connections do not use assigned proxies


tenshiak:
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Linux

### Description

I added multiple noauth connections to "OpenCode Free" and assigned proxies to all of them, but the proxies are simply not used and all connections go directly. Confirmed by looking at Proxy Logs. All connections are direct.

### Steps to Reproduce

1. Add opencode free provider (noauth)
2. Assign a proxy to it
3. Check logs if it uses the proxy

### Expected Behavior

OmniRoute to use the assigned proxy

### Actual Behavior

Goes direcrtly

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

diegosouzapw:
Thanks for the report. I traced the noauth proxy path and there **is** a real defect, but it's narrower than "noauth ignores proxies" — so one detail decides whether you're hitting it. Marking `needs-info` for that one question.

Noauth providers (OpenCode Free etc.) don't resolve a real `provider_connections` row; the assigned proxy is carried inline via `providerSpecificData.accountProxies` (hydrated in `auth.ts::loadNoAuthProviderSpecificData`) and applied at egress by the executor (`open-sse/executors/opencode.ts` pins each account with `runWithProxyContext(account.proxy, …)`). So keyless connections **do** use their proxy on the normal path.

The defect: `loadNoAuthProviderSpecificData` merges `fingerprints` and `accountProxies` **independently, each from the first connection row that has them**. If you have **more than one `provider_connections` row** for the same noauth provider, fingerprints can come from row A while accountProxies come from row B — the proxy map keys then never match the account fingerprints, every account resolves `proxy: null`, and all traffic goes DIRECT. That exactly matches "multiple connections, each assigned a proxy, all go DIRECT." (Two adjacent failure modes: empty/missing `fingerprints` collapses accounts to a single default with `proxy:null`; and a by-id reference to a deleted proxy-pool entry silently degrades to null.)

To confirm you're on this path:
1. How many separate **OpenCode Free connections/rows** do you have for that provider — one, or several?
2. Are the proxies assigned **per-account inside one connection**, or one proxy per separate connection row?
3. If you consolidate to a single connection with the proxies assigned per-account, does the proxy start being used?

[...]

PR #13283, opened 2026-09-10
fix(routing): stop round-robin combo opencode targets from collapsing onto opencode-zen (#11912)

### LINKED ISSUE #11912: fix(providers): round-robin combo routes all requests to opencode-zen due to connection/alias collision with opencode
### OmniRoute Version

v3.8.50

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 24H2

### Node.js Version

v22.22.2

### Provider(s) Involved

opencode cli

### Model(s) Involved

mimo-v2.5-free

### Client Tool

Opencode Cli

### Description

In a Round Robin combo configured with multiple opencode targets and one opencode-zen target, the routing engine sends 100% of upstream traffic exclusively to the opencode-zen provider account.

In the Connections panel, the opencode provider is sharing/inheriting the exact same account credentials as opencode-zen. In earlier versions, opencode ran strictly unauthenticated/dynamic without account credentials attached. Because they now share account connections under the hood, the router merges their execution paths.

Furthermore, there is a metrics desync: the Resolved Runtime Targets dashboard shows 0 requests for opencode-zen over the last hour, even though 100% of request logs confirm opencode-zen received every call (and quickly hit 429 rate limits).

### Steps to Reproduce

Go to Connections and link an account to opencode-zen. Notice opencode (Free) shares or inherits this exact same connection.

Create a Combo using Round Robin strategy with 7 targets:

Targets 1–6: opencode (dynamic/account targets)

Target 7: opencode-zen

Dispatch sequential requests to the combo endpoint (e.g., using model mimo-v2.5-free).

Check the upstream execution logs vs. Resolved Runtime Targets metrics.

### Expected Behavior

Requests should be distributed sequentially across all 7 targets in a Round Robin sequence (Target 1 -> Target 2 -> ... -> Target 7).

opencode (Free CLI / Dynamic) and opencode-zen (Authenticated) should be isolated as distinct providers without credential or executor collision.

Resolved Runtime Targets metrics should accurately record dispatches against the actual target executed upstream.

### Actual Behavior

Traffic bypasses all 6 opencode targets and continuously hits the opencode-zen target.

opencode-zen quickly encounters 429 Rate Limit errors under load.

Resolved Runtime Targets logs 0 requests for opencode-zen, while upstream execution logs prove opencode-zen handled all requests.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Unit test for provider alias isolation to ensure opencode dynamic executors do not inherit or merge with opencode-zen account connections.

Integration test for Round Robin combo execution confirming metrics increment correctly on the target handling upstream traffic.
```

### Screenshots

Image

Image

Image

Image

Image

Image

### Additional Context

Connection Collision: Under Connections, opencode (Free) shows the same account credentials as opencode-zen. In older releases, opencode did not require or attach account connections and operated solely on no-auth/dynamic pools.

Metrics Desync Cause: The router appears to attribute metrics to the pre-routed opencode target IDs (Targets 1–6) at dispatch time, but executes the call through the opencode-zen executor. This leaves opencode-zen showing 0 requests in runtime metrics despite taking all traffic.

### Validation Plan

Ensure opencode (dynamic/unauthenticated) can run without inheriting opencode-zen connections.

Run a 10-request benchmark through a Round Robin combo containing both opencode and opencode-zen targets.

Verify via upstream logs and dashboard metrics that requests rotate evenly and metrics match actual upstream logs.

diegosouzapw:
Closes #11912

## Root cause

`open-sse/services/model.ts`'s manual `ALIAS_TO_PROVIDER_ID["opencode"] = "opencode-zen"` override canonicalizes ANY `"opencode/<model>"` string to provider `opencode-zen` (the api-key gateway) before dispatch. A round-robin combo mixing declared `"opencode/<model>"` targets (intended as the free/dynamic no-auth pool) with an explicit `"opencode-zen/<model>"` target therefore collapsed every rotation slot onto the same provider + connection identity — every request executed against the single `opencode-zen` connection instead of rotating across the free pool, matching the reporter's upstream-log screenshots (100% of traffic landing on `opencode-zen`, eventually 429ing). The same collapse explained the reported dashboard metrics desync: `src/lib/combos/controlCenter.ts`'s `providerFromModel()` labeled targets from a raw, un-aliased string slice, so the "Resolved Runtime Targets" panel and the actual dispatched identity disagreed.

## Fix

Scoped to combo target resolution only — `open-sse/services/model.ts`'s general alias-resolution path is untouched, so a raw client request to `"opencode/<model>"` outside a combo keeps routing to `opencode-zen` unchanged (#2798/#3870), and the #7993 sibling credential lookup (`tests/unit/opencode-autocombo-search-pair.test.ts`) is unaffected.

- New `open-sse/services/combo/opencodeTargetAlias.ts`: `resolveComboTargetModelStr()` rewrites an ambiguous `"opencode/<model>"` combo target to the `"oc/"` no-auth alias, mirroring the combo builder's existing #2901 guard (`src/lib/combos/builderOptions.ts`).
- `open-sse/services/combo/comboStructure.ts`'s `normalizeRuntimeStep()` (the single shared resolution point for every combo strategy) applies that rewrite before the model string reaches dispatch, so `"opencode/<model>"` and `"opencode-zen/<model>"` targets stay distinct rotation identities.
- `src/lib/combos/controlCenter.ts`'s `providerFromModel()` resolves labels through the same alias-normalization path (plus the existing `resolveProviderAlias`), so the dashboard label matches what actually executes upstream.

## Regression test

`tests/unit/issue-11912-opencode-roundrobin-collapse.test.ts`

- RED (against `origin/release/v3.8.51`, `comboStructure.ts` unfixed): `AssertionError [ERR_ASSERTION]: Expected values to be strictly equal: + 'opencode-zen' - 'opencode'`
- GREEN (with the fix): both new tests pass.

## Gates run

- `node scripts/check/check-file-size.mjs` — no `✗` for touched files
- `node scripts/check/check-complexity.mjs` — OK (2798 violations vs baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1265 violations vs baseline 1437)
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json <changed files>` — exit 0
- New test: `node --import tsx/esm --test tests/unit/issue-11912-opencode-roundrobin-collapse.test.ts` — 2/2 pass
- Existing combo/round-robin/opencode suites re-run together (147 tests, 0 failures): `tests/unit/combo/round-robin-combo.test.ts`, `tests/unit/combo-pin-implicit-allowlist.test.ts`, `tests/unit/combo-routing-engine.test.ts`, `tests/unit/combo-scoring-inspector.test.ts`, `tests/unit/opencode-autocombo-search-pair.test.ts`, `tests/unit/refactor-buildHeaders-opencode.test.ts`, `tests/unit/model-alias-seed-fallback.test.ts`, `tests/unit/combo-control-center.test.ts`, `tests/unit/fusion-vision-panel-3378.test.ts`

### DIFF
--- a/changelog.d/fixes/11912-roundrobin-opencode-zen-collision.md
+++ b/changelog.d/fixes/11912-roundrobin-opencode-zen-collision.md
@@ -1,0 +1,1 @@
+- fix(routing): stop a round-robin combo's "opencode" targets from collapsing onto the opencode-zen connection (#11912)
--- a/open-sse/services/combo/comboStructure.ts
+++ b/open-sse/services/combo/comboStructure.ts
@@ -26,6 +26,7 @@ import { containsMediaKind } from "../../utils/mediaParts.ts";
 import { getResolvedModelCapabilities } from "../modelCapabilities.ts";
 import { parseModel, stripContextWindowSuffix } from "../model.ts";
 import { dedupeTargetsByExecutionKey, isRecord } from "./comboData.ts";
+import { resolveComboTargetModelStr } from "./opencodeTargetAlias.ts";
 import { isComboModelVisible } from "./comboVisibility.ts";
 import { getTargetProvider, MAX_COMBO_DEPTH } from "./comboPredicates.ts";
 import { evaluateContextLimit } from "./contextOverrideGate.ts";
@@ -122,8 +123,13 @@ function normalizeRuntimeStep(
     };
   }
 
-  const modelStr = getComboModelString(step);
-  if (!modelStr) return null;
+  const declaredModelStr = getComboModelString(step);
+  if (!declaredModelStr) return null;
+  // #11912: rewrite an ambiguous "opencode/<model>" target to the "oc/" alias
+  // so it stays distinct from an explicit "opencode-zen/<model>" sibling
+  // instead of both collapsing onto the same provider — see
+  // opencodeTargetAlias.ts for the full rationale.
+  const modelStr = resolveComboTargetModelStr(declaredModelStr);
 
   const connectionId = toTrimmedString(step.connectionId);
   const allowedConnectionIds = implicitPinAllowlist(connectionId, step.allowedConnectionIds);
--- a/open-sse/services/combo/opencodeTargetAlias.ts
+++ b/open-sse/services/combo/opencodeTargetAlias.ts
@@ -1,0 +1,43 @@
+/**
+ * Issue #11912 — a combo step declared with the raw "opencode/<model>" prefix
+ * is ambiguous: open-sse/services/model.ts's manual ALIAS_TO_PROVIDER_ID
+ * override canonicalizes ANY "opencode/<model>" string to provider
+ * "opencode-zen" (the api-key gateway) before dispatch. A round-robin combo
+ * mixing declared "opencode/<model>" targets (intended as the free/dynamic
+ * no-auth pool) with an explicit "opencode-zen/<model>" target therefore
+ * collapses every rotation slot onto the SAME provider + connection identity
+ * — every request executes against the single opencode-zen connection
+ * instead of rotating across the free pool, and the account eventually
+ * 429s.
+ *
+ * The combo BUILDER already avoids this for freshly-generated model strings
+ * by emitting the "oc/" alias for the no-auth provider (#2901,
+ * src/lib/combos/builderOptions.ts's rewriteQualifiedModelPrefix). This
+ * mirrors that same substitution at combo TARGET RESOLUTION time so a step
+ * saved — or hand-typed — with the raw "opencode/" prefix still reaches the
+ * true no-auth provider and stays a distinct rotation identity from an
+ * explicit "opencode-zen/<model>" target.
+ *
+ * Deliberately scoped to combo target resolution only — this never touches
+ * open-sse/services/model.ts's general alias-resolution path, so a raw
+ * client request to "opencode/<model>" outside a combo keeps routing to
+ * opencode-zen unchanged (#2798/#3870), and the #7993 sibling credential
+ * lookup (tests/unit/opencode-autocombo-search-pair.test.ts) is unaffected.
+ */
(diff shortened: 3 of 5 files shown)


HouMinXi:
Heads-up: applied on top of `release/v3.8.51`, this PR breaks `npm run build` (webpack).

`src/lib/combos/controlCenter.ts` is reachable from the client component `src/app/(dashboard)/dashboard/combos/ComboControlCenterClient.tsx`. The two new imports

```ts
import { resolveComboTargetModelStr } from "../../../open-sse/services/combo/opencodeTargetAlias.ts";
```

pull the server-side open-sse graph into the browser bundle. The build then fails with:

```
./node_modules/detect-libc/lib/detect-libc.js
Module not found: Can't resolve 'child_process'

Import trace:
./node_modules/sharp/dist/utility.mjs
./open-sse/utils/cursorImages.ts
./open-sse/executors/zai-web.ts
./src/lib/tokenHealthCheck.ts
./src/lib/db/settings.ts
./src/lib/db/readCache.ts
./open-sse/services/model.ts
./src/app/(dashboard)/dashboard/combos/ComboControlCenterClient.tsx
```

Same chain also produces `Can't resolve 'fs'` (open-sse/config/credentialLoader.ts) and `Can't resolve 'module'` (open-sse/executors/codex.ts).

The alias resolution needs to stay server-side. Two options: normalize the label in the API route that serves the control-center data instead of in the shared module, or move the two helpers into a small dependency-free module that does not transitively import `open-sse/services/model.ts`.

Two side notes:
- The explicit `.ts` extensions on those imports are inconsistent with the rest of the codebase and may trip other bundler configs.
- Dropping those two imports and reverting `providerFromModel` to the prefix slice restores the build; the label then shows the raw prefix for opencode combos, which is cosmetic only.

PR #13289, opened 2026-09-10
fix(adobe-firefly): do not spawn Chrome for CDP warm under test runners

anhtahaylove:
## Problem

`tests/unit/8510-adobe-firefly-edits-route.test.ts` fails 4 of 5 tests on Windows with `EPERM` and takes **63 seconds**:

```
Error: EPERM, Permission denied: \?\C:\Users\...\Temp\omniroute-adobe-firefly-edits-ct7k0M
    at Object.rmSync (node:fs:1283:18)
    at resetStorage (tests/unit/8510-adobe-firefly-edits-route.test.ts:39:6)
```

The test stubs `globalThis.fetch`, so at first glance nothing should touch the network or disk.

## Root cause

The route reaches `adobeFireflySession.ts`, whose CDP session-warm path **spawns a real headed Chrome**. `fetch` is stubbed; CDP is not — it talks a raw socket.

The gate was:

```ts
const tryBrowser = opts?.tryBrowser !== false && process.env.ADOBE_FIREFLY_BROWSER_REFRESH !== "0";
```

Only an explicit opt-out, with no test-runner detection. Two consequences:

1. **EPERM** — Chrome holds an OS handle on its profile directory (`<DATA_DIR>/adobe-chrome-profiles/...`). Windows refuses to remove a directory while a handle is open, so the teardown `rmSync(TEST_DATA_DIR)` fails. `maxRetries: 5` does not help: the browser is still alive.
2. **63s runtime** — the CDP socket keeps the runner busy for the full 75s warm timeout.

I confirmed the handle by instrumenting the failing teardown:

```
PROBE_HANDLES: Socket
PROBE_LEFT: adobe-chrome-profiles, storage.sqlite
```

`storage.sqlite` is a red herring — SQLite closes correctly (verified separately: `resetDbInstance()` then `rmSync` succeeds). The blocker is the browser profile directory.

## Fix

The same module already detects test runners for the disk-session cache:

```ts
function diskSessionsEnabled(): boolean {
  if (process.env.ADOBE_FIREFLY_SESSION_DISK === "0") return false;
  if (process.env.NODE_ENV === "test") return false;
  if (process.env.VITEST || process.env.NODE_TEST_CONTEXT) return false;
  return true;
}
```

That guard covers the disk cache but not the browser spawn. This PR adds the matching `browserRefreshEnabled()` and routes both call sites through it. Spawning a real Chrome is never valid under a unit-test runner, so this is a gap rather than a behaviour change — production is untouched, and `ADOBE_FIREFLY_BROWSER_REFRESH=0` keeps working.

## Verification

| | without fix | with fix |
|---|---|---|
| `8510-adobe-firefly-edits-route` | 1 pass / **4 fail**, 63.3s | **4 pass / 0 fail**, 2.5s |

Red-before/green-after confirmed by stashing the source change and re-running. Deterministic across 3 consecutive runs before the fix and 2 after.

Adobe Firefly suite — **81/81 pass**, including `adobe-firefly-browser-login.test.ts`, which covers the browser path itself and drives it through injected `spawnFn`/explicit options rather than the ambient env, so it is unaffected.

Type-check: the fixed file is clean. Note that `tsc -p tsconfig.json` reports **5051 errors on both this branch and a clean `release/v3.8.51`** (identical count) — pre-existing, from `bin/` being in the project; CI gates on a different config.

## Scope

This addresses part of the local Windows baseline that fails the same way. Found while auditing the 377 baseline failures behind #13187; CI runs Linux, where a failed Chrome spawn does not lock the directory, so this was invisible there.


anhtahaylove:
@diegosouzapw — still open and MERGEABLE, test-only (`git diff origin/release/v3.8.51 -- src/` is empty). Grouped context for this PR and the other two left over from the 10–11 Sep batch is on #13292: https://github.com/diegosouzapw/OmniRoute/pull/13292#issuecomment-5653579490

Short version: #13187 has merged, so the concurrency blocker is gone; the red checks here are the repo-wide ones that also fail on `release/v3.8.51` and on other contributors' PRs, not something specific to this branch. Happy to rebase or close on request.

anhtahaylove:

I called this PR test-only and backed that with `git diff origin/release/v3.8.51 -- src/` being empty. That check was wrong: `src/` exists (2270 non-test `.ts` files), but it is not the only product tree — `open-sse/` holds another 1558. Scoping the diff to `src/` alone silently hid every change under `open-sse/`. I should have diffed the whole tree.

Re-checked properly, **this PR is not test-only** — it changes product code:

| File | Kind |
|---|---|
| `open-sse/services/adobeFireflySession.ts` | **product** |
| `changelog.d/fixes/adobe-firefly-no-browser-spawn-in-tests.md` | changelog |

It contains no test files at all.

The change extracts the browser-refresh decision into `browserRefreshEnabled()` and additionally suppresses the Chrome spawn when `NODE_ENV === "test"`, or when `VITEST` / `NODE_TEST_CONTEXT` is set:

```ts
function browserRefreshEnabled(): boolean {
  if (process.env.ADOBE_FIREFLY_BROWSER_REFRESH === "0") return false;
  if (process.env.NODE_ENV === "test") return false;
  if (process.env.VITEST || process.env.NODE_TEST_CONTEXT) return false;
  return true;
}
```

Production behaviour is unchanged when none of those variables are set, but this is a runtime code path and deserves review as such — not the rubber stamp my earlier "test-only" label invited. Sorry for the noise.

The rest of my previous comment stands: the branch is rebased onto `release/v3.8.51` (`903ee35c5`, behind = 0) and the failing checks are the repo-wide ones that also fail on the base branch.

diegosouzapw:
Good root-cause fix, and it matches the `NODE_TEST_CONTEXT`/`VITEST` detection pattern
already used in 5 other files in this codebase, so it's architecturally consistent rather
than a one-off. Verified the file lints clean against the real suppressions file. This pairs
well with #13358's global test-setup guard — recommending both merge for defense in depth.

Issue #13290, opened 2026-09-10
fix(backend): Windows: ~29 unit tests fail with EPERM because test.after() removes DATA_DIR without closing the DB


anhtahaylove:
## Summary

~29 unit tests fail on Windows with `EPERM` in teardown. Two independent causes; one is fixed in #13289, this issue tracks the second.

```
Error: EPERM, Permission denied: \?\C:\Users\...\Temp\omniroute-executor-registry-CwqmB0
    at Object.rmSync (node:fs:1283:18)
    at TestContext.<anonymous> (tests/unit/executor-registry.test.ts:21:6)
```

CI runs Linux, where an open handle does not block `rmdir`, so this is invisible there. It only hurts contributors on Windows.

## Cause

A test opens the SQLite DB (often indirectly, by importing a module that touches it) and its `test.after()` removes `TEST_DATA_DIR` **without closing the DB first**. Windows refuses to remove a directory that still has an open handle.

Instrumenting `executor-registry.test.ts` at the failing teardown:

```
P3_HANDLES: Socket, Socket
P3_LEFT: storage.sqlite, storage.sqlite-shm, storage.sqlite-wal
P3: rmSync FAIL EPERM
```

The `-shm`/`-wal` sidecars are the tell: the connection is still open. `maxRetries: 5` does not help, because nothing will ever close it.

Note this is specifically about the **final** `test.after()`. Several affected files *do* call `core.resetDbInstance()` in a per-test `resetStorage()` helper, but not in the last teardown — so the DB reopened by the final test stays open. `api-key-lifecycle.test.ts` is exactly this shape.

Confirmed failing individually (2 EPERM hits each):
`executor-registry`, `api-key-lifecycle`, `provider-health-matrix`, `reasoning-routing`, `zcode-executor`.

## Scope

A conservative scan — `test.after()` blocks that `rmSync` a `TEST_DATA_DIR` with no `resetDbInstance`/`closeDb` in the same block — matches **98 files**. Only the subset that actually opens the DB fails today; the rest are latent and will start failing the moment an import pulls the DB in.

```
tests/unit/executor-registry.test.ts
tests/unit/api-key-lifecycle.test.ts
tests/unit/provider-health-matrix.test.ts
tests/unit/reasoning-routing.test.ts
tests/unit/zcode-executor.test.ts
... (93 more)
```

## Suggested fix

[...]

PR #13292, opened 2026-09-10
test: close the database before removing temp DATA_DIR (#13290)

anhtahaylove:
Implements the shared-helper option from #13290.

## The bug

A test creates its own temp `DATA_DIR`, opens the database, then removes the directory in `test.after()`. On Windows that fails with `EPERM`: nothing closes the SQLite connection, so the directory still has an open handle and the `storage.sqlite-shm` / `-wal` sidecars keep it locked. The existing `maxRetries: 5` cannot help — every retry hits the same handle that nothing ever closed.

The `process.on("exit")` fallback in `isolateDataDir.ts` does not cover these tests either: it only runs when the test does **not** set `DATA_DIR` itself, which is exactly the failing case.

## The fix

`tests/_setup/tempDataDir.ts` exposes two entry points:

- `cleanupTempDataDir(dir)` — for suites that already have their own `mkdtempSync`; only the close-then-remove ordering is fixed.
- `createTempDataDir(prefix)` — for new tests; returns `{ dir, cleanup }`.

Two deliberate choices:

- **The DB import is lazy.** A static import would drag the whole DB layer into every test that only wants a scratch directory. If the module was never loaded, there is nothing to close.
- **Removal is best-effort.** A leftover temp directory is a disk nuisance the OS reaps; a throwing teardown is a red test hiding a green assertion.

## Verification

The helper's own test proves the ordering is what matters — with the close removed it fails:

```
AssertionError [ERR_ASSERTION]: cleanup must remove the directory
ℹ pass 2
ℹ fail 1
```

Restored, the five converted suites plus the helper test run clean together:

```
ℹ tests 32
ℹ pass 32
ℹ fail 0
```

Each previously reported 2 EPERM failures in teardown.

`tsc -p tsconfig.json --noEmit` reports **5057 errors on this branch and 5057 on base** — diffed line by line, the only delta is two pre-existing `rulesDb` errors in `reasoning-routing.test.ts` shifting from lines 28/29 to 29/30 because of the added import. No new errors. ESLint and Prettier clean.

## Scope

This converts the five suites confirmed failing, not all ~1079 files that manage their own `DATA_DIR`. The remaining ones only fail when they actually open the database; converting them blind would be a large untested diff. With the helper in place they can be migrated as they surface.

Measured impact of leaving this unfixed: [#13290 comment](https://github.com/diegosouzapw/OmniRoute/issues/13290#issuecomment-5624214594) — 8346 leaked directories / 34.7 GB on one dev machine.


anhtahaylove:
### Note for whoever measures this branch

This branch is cut from current `release/v3.8.51`, which does **not** yet contain #13187. So `npm test` here still runs `--test-concurrency=20` with no `--test-force-exit`.

Measured on this machine, that configuration does not complete: after 29 minutes only 12 test files had reported, with the log stalled for the last 20 of them while worker processes kept burning CPU. Re-running the same suite at `--test-concurrency=4 --test-force-exit` is what produced the usable numbers.

Not a defect in this PR — it is the exact problem #13187 fixes. Worth merging #13187 first so that anyone measuring this branch gets a suite that terminates.

anhtahaylove:
## Full-suite measurement + 15 more suites converted

### Full suite on this branch

| | base | this branch |
|---|---:|---:|
| pass | 37460 | **37727** |
| fail | 377 | **370** |
| serial | — | 23/23 pass |
| wall clock | — | 1519s |

No regression: 7 fewer failures, 267 more passing tests.

> [!IMPORTANT]
> This branch is cut from a base that does **not** contain #13187, so it still
> defaults to `--test-concurrency=20`. My first attempt to measure it hung: 29
> minutes in, only 12 files had finished. The numbers above were produced with
> `--test-concurrency=4 --test-force-exit` applied locally (not committed).
> **#13187 should merge first**, otherwise anyone measuring this branch hits the
> same non-terminating run.

### 15 more suites converted

The latest commit converts every suite that measurably emitted EPERM during the
full run. Same 15 files, measured both ways:

| | base | this branch |
|---|---:|---:|
| failures | 22 | **7** |
| EPERM lines | 40 | **10** |

The 7 remaining failures are **pre-existing and unrelated to teardown**:
`rtk-learn-discover-routes` (6 failures) and `executor-map-golden` (3 failures)
already fail on a clean base. The conversion improves both rather than
regressing them (6 -> 5 and 3 -> 2).

### Scope: 26 files, not 968

A regex sweep suggested 968 candidates, but a random sample of 20 run against a
clean base showed only 3 actually emit EPERM (15%). The full-suite log gives the
real list: **26 files**. Pattern matching produced candidates; only running them
produced proof, so only the proven ones were converted.

anhtahaylove:
## Sweep complete — final 9 suites, plus a second teardown shape

`dc397447d` converts the remaining suites from the #13290 list. **EPERM lines are
now 0 in every converted file.**

### A second failure shape: `beforeEach`, not `after`

The earlier commits only fixed `after()` hooks. Several files wipe `DATA_DIR` in
**`beforeEach`** instead, while the previous test's connection is still open:

```ts
test.beforeEach(() => {
  fs.rmSync(TEST_DATA_DIR, { recursive: true, force: true, ... });
  fs.mkdirSync(TEST_DATA_DIR, { recursive: true });
});
```

These fail *before the test body runs*, so every test in the file reports the
same EPERM path with a sub-millisecond duration — which is why
`rtk-learn-discover-routes` looked like 5 unrelated failures. It was one leak.

| file | before | after |
|---|---|---|
| `rtk-learn-discover-routes` | 0 pass / 5 fail, 10 EPERM | **5 pass / 0 fail, 0 EPERM** |
| `rtk-mcp-tools` | 1 pass / 3 fail, 6 EPERM | **4 pass / 0 fail, 0 EPERM** |
| `provider-window-costs` | 0 pass / 6 fail, 12 EPERM | **6 pass / 0 fail, 0 EPERM** |
| `chatcore-reasoning-cache-write-guard` | — | **4 pass / 0 fail, 0 EPERM** |
| `codex-fingerprint-seed-persistence` | — | **5 pass / 0 fail, 0 EPERM** |
| `8510-adobe-firefly-edits-route` | 1 pass / 4 fail, 8 EPERM | 3 pass / 1 fail, **0 EPERM** |
| `driverFactory` | 28 pass / 1 fail | 28 pass / 1 fail, 0 EPERM |
| `responses-parse-once-4041` | 4 pass / 1 fail, 2 EPERM | 4 pass / 1 fail, **0 EPERM** |
| `responses-route-early-keepalive-wiring` | 3 pass / 1 fail, 2 EPERM | 3 pass / 1 fail, **0 EPERM** |

The 4 remaining failures were verified against a clean base and **fail there
identically** — they are pre-existing and unrelated to teardown.

### Three files leaked despite calling `resetDbInstance()`

`8510-adobe-firefly-edits-route`, `provider-window-costs` and `rtk-mcp-tools`
already called `core.resetDbInstance()` immediately before `rmSync` and still hit
EPERM. That is the product-side connection leak now filed as **#13303**: the
reset does not release every handle. The helper works around it in tests; #13303
is the real fix.

### Files this PR deliberately does not convert

`nvidia-nim-validator`, `zai-validator` and `proxy-concurrency-keepalive-regression`
appeared in the EPERM sweep but have **no directory-removing teardown at all** —
run individually they emit 0 EPERM. They were collateral damage from another
file's leak during the concurrent run. Each has 1 pre-existing failure that is
also present on a clean base. Converting them would have been a no-op dressed up
as a fix.

This PR still contains **no product changes** (`git diff origin/release/v3.8.51 -- src/` is empty).

anhtahaylove:
## Two of the remaining red suites diagnosed — neither is a teardown bug

Both reproduce on a clean base, so neither belongs to this PR. Filing the findings
here so the diagnosis is not lost.

### 1. `driverFactory.test.ts` — libuv abort is caused by `--test-force-exit`

```
Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 94
```

All 28 subtests pass; the process aborts *after* the suite reports success, so the
runner scores the file as 1 failure.

Bisected by test name:

| Run | Result |
|---|---|
| full file, `--test-force-exit` | 28 pass, **abort** (3/3 runs) |
| sql.js tests only | 5 pass, **abort** |
| everything except sql.js | 23 pass, clean |
| `cross-driver: escreve com adapter sync, relê com sql.js` alone | 1 pass, **abort** |
| other four sql.js tests, each alone | clean |
| **full file, no `--test-force-exit`** | **28 pass, exit 0, clean (2/2 runs)** |

`cross-driver` is the only test that calls `createSqlJsAdapter()` directly instead
of going through the cached `preInitSqlJs()`. That adapter arms a `setInterval`
checkpoint timer plus `beforeExit`/`SIGINT`/`SIGTERM` listeners (`sqljsAdapter.ts:277-292`).
The test does call `reader.close()` and `gracefulClose()` does clear all of them —
the teardown is correct. The abort comes from `--test-force-exit` calling
`process.exit()` while the Emscripten/WASM async handle is still mid-close, which
trips libuv's assertion on Windows.

So this is a **runner-flag artifact, not a product or test defect**. Dropping
`--test-force-exit` for this file makes it green and exit 0.

### 2. `8510-adobe-firefly-edits-route.test.ts` — 500 vs 401 is cross-test DB bleed

My earlier note guessed a schema mismatch. That was wrong; here is what actually happens.

The test asserts 401 for the *no credentials seeded* path but gets 500 with:

```json
{"error":{"message":"Missing-credentials path must not reach upstream","type":"upstream_error"}}
```

That message is the test's **own** `globalThis.fetch` stub throwing — so the route
did find a credential and tried to dispatch upstream.

Isolation:

| Run | Result |
|---|---|
| `surfaces missing credentials` alone | **pass** |
| with `rejects more than 4 reference images` | pass |
| with `uploads ... dispatches referenceBlobs` | **fail, 500** |

Instrumenting the test right before the assertion shows the previous test's
connection is still live:

```
[FFDBG] conns con lai = [{"id":"a7b11de4-...","provider":"adobe-firefly","isActive":true,"apiKey":"eyJhbG..."}]
```

`beforeEach` → `resetStorage()` does everything right (`resetDbInstance()`, then
remove `DATA_DIR`), but logging the directory afterwards shows the removal silently
failed on the second pass:

```
[FFDBG] sau cleanup, con file: []
[FFDBG] sau cleanup, con file: ["adobe-chrome-profiles","storage.sqlite"]
```

`cleanupTempDataDir` is best-effort by design and swallows the removal error, so
`storage.sqlite` survives with the seeded row in it and the next test reads stale
credentials.

Two dead ends worth recording so nobody repeats them: clearing `quotaCache`
(`__clearForTests()`) and clearing the read cache (`invalidateDbCache()`, both
scoped and global) each leave the failure unchanged. The stale data is on disk,
not in a cache.

**This is a downstream symptom of #13303** — the connection that blocks the delete
is the leaked probe handle. Fixing #13303 should fix this suite; patching the test
alone would only paper over it.

anhtahaylove:
Ready for review when someone has a moment.

State of this PR:

- **31 files, test-only.** `git diff origin/release/v3.8.51 -- src/` is empty.
- Converts the 24 suites the sweep proved to emit EPERM; each reports **0 EPERM**
  when run individually.
- The three suites that only failed as collateral damage (`nvidia-nim-validator`,
  `zai-validator`, `proxy-concurrency-keepalive-regression`) are deliberately left
  alone — they are green on their own.

Not in scope here, both filed and analysed separately:

- **#13303** — the product-side probe handle leak in `core.ts`. I attempted a fix
  and reverted it; the naive `finally` swaps EPERM for `SqliteError` while keeping
  the failure count identical.
- The two remaining red suites in this area (`driverFactory`, the Firefly route)
  are diagnosed in the comment above; both reproduce on a clean base.

Merging this is independent of #13303 — it removes the teardown noise that hides
real failures, without touching product code.

anhtahaylove:
## Three more commits since the last update — still looking for a review

@diegosouzapw — this branch has grown since I last flagged it, so here is the
current state rather than a bare ping.

### New since `e88e65db1`

| Commit | What |
|---|---|
| `1dd56416` | Close SQLite handles in the three suites that leaked *while passing* |
| `aea2b86e` | Remove the temp `DATA_DIR` in five suites that never cleaned up at all |
| `280662c1` | Fix the contradictory Jina assertion (see #13313) |

All nine touched suites, re-run just now on this head: **104 passing, 0 failing.**

```
44 models-catalog-route        22 batch_api            14 agenticConversations
 5 antigravity-auto-sync        3 monitoring-health     3 presets-dashboard-auth
 2 issue-agent-route            4 agentbridge-routerkey  7 agent-bridge-bypass
```

### What is worth your attention, in priority order

1. **`280662c1` is the only commit that changes an assertion** rather than
   teardown, so it deserves the closest look. The custom-model test expected
   `jina-ai/...` while the synced-model test directly above it expected `jina/...`
   and passed — the file contradicted itself. The product prefixes ids with the
   provider alias (`alias: "jina"`), added in v3.8.36, three weeks after the test
   was written. I aligned the expectation with the product and the sibling test,
   and confirmed it still bites by renaming the alias to `jina-XX` — exactly those
   two cases fail. If you would rather the product changed to emit `jina-ai/`,
   say so and I will revert this commit; the reasoning is laid out in #13313.

2. **`1dd56416` found a bug I had missed.** `issue-agent-route-execution` has no
   teardown at all, so it left an orphaned DB in the *shared* `omniroute-test-*`
   dir owned by `isolateDataDir.ts` — not in a private dir. That is the one case
   here with real cross-suite blast radius.

3. **`aea2b86e` is the low-risk one.** Those five suites each use a private
   `mkdtemp` dir, so they were wasting disk, not corrupting results.

### Also open, and independent of this PR

- **#13306** — sql.js aborts on `process.exit()` after `db.run()`. Confirmed
  Windows-only (clean 3/3 on Linux glibc and musl, Node 22 and 24), so CI will
  never catch it. `gracefulShutdown.ts:203` is on the same code path.
- **#13313** — the two tests that contradict the product.
- **#13303** — full sweep results: 307 candidates, 6 leaking suites, all fixed here.

No rush on my side, but this PR is test-only and currently MERGEABLE, so it will
start conflicting as the test dir moves. Happy to split it if 40 files is too
much for one review.

anhtahaylove:
Rescoped this PR to teardown only, so it is easier to review.

The two assertion changes that used to sit at the tip (`280662c1` jina, `c101c1aa1` pt-BR) moved to #13315. They touched different files from the 39 here and passed on a clean base branch with no teardown commits, so the split is clean — this PR is now 6 commits, 39 files, all mechanical `cleanupTempDataDir` / `test.after` work.

Force-pushed to drop those two commits; nothing else changed and there were no reviews to invalidate.

anhtahaylove:
Rebased onto the current base (`af49d4972`) — the branch was based on `fd27ff08c`, which is now 1 commit behind.

**The PR content is byte-for-byte unchanged.** Diff against the respective base before and after the rebase is identical (43,208 chars both ways), no conflicts, all 6 commits preserved.

### The rebase does not turn CI green, and that is expected

The red Unit Tests jobs are **not caused by this PR**. `tests/unit/glm-sse-transform-arity.test.ts` is failing on the base itself:

| commit | result |
|---|---|
| `af49d4972~1` | pass 2 / fail 0 |
| `af49d4972` (base HEAD) | **fail 1** |

Filed as #13319 with the bisect. `af49d4972` legitimised a buffer-size argument that the arity guard from #12770 still pins against.

This PR touches 39 files, none of them GLM-related. Running the suites it does touch, plus the failing one, on the rebased branch:

```
tests 30 | pass 29 | fail 1
```

The single failure is the inherited base one.

So this stays blocked on #13319 rather than on anything reviewable here. Happy to rebase again once that lands.

anhtahaylove:
@diegosouzapw — three test-only PRs from the 10–11 Sep batch are still open. Everything they were waiting on has landed, so here is the current state rather than a bare ping.

### What changed since I last commented

**#13187 merged** (11 Sep 16:27), so `release/v3.8.51` now caps unit-test concurrency at 4. That was the blocker I flagged on #13292 — measuring these branches no longer hangs. Fourteen of my other PRs went in during that same 16:27–16:32 batch; these three were not part of it.

### The three

| PR | Scope | State |
|---|---|---|
| #13289 | `adobe-firefly`: do not spawn Chrome for CDP warm under test runners | MERGEABLE, 1 commit |
| #13292 | close the database before removing temp `DATA_DIR` (39 files) | MERGEABLE, 6 commits |
| #13315 | realign two stale assertions with product behavior (#13313) | MERGEABLE, 2 commits |

All three are test-only — `git diff origin/release/v3.8.51 -- src/` is empty on each.

### About the red checks

The failing checks on all three predate the batch merge and are **not specific to these branches**:

- `release/v3.8.51` itself is red on `Release-Green (continuous)` and three nightlies.
- Other contributors' open PRs fail the same eight checks.
- #13289's failure resolves to `2 gate(s) failed: pack-policy open-sse-typecheck`.

These branches are also 113–164 commits behind the current base, so their last CI run measured a base that no longer exists. I have not rebased them, since that would invalidate the measurements recorded on #13292 — happy to rebase all three on request, which should also pick up whatever fixed the gates.

### What I am asking for

A review, or a note on what to change. If any of the three is no longer wanted, closing it is a fine outcome — I would just like to stop carrying them.

anhtahaylove:

I called all three PRs test-only and backed that with `git diff origin/release/v3.8.51 -- src/` being empty on each. That check was wrong: `src/` exists (2270 non-test `.ts` files), but it is not the only product tree — `open-sse/` holds another 1558. Scoping the diff to `src/` alone silently hid every change under `open-sse/`. I should have diffed the whole tree.

Re-checked properly, by file kind:

| PR | Files | Tests | Changelog | Other |
|---|---:|---:|---:|---|
| #13289 | 2 | 0 | 1 | **1 product file** |
| #13292 (this one) | 39 | 37 | 1 | 1 — `tests/_setup/tempDataDir.ts`, test harness |
| #13315 | 2 | 2 | 0 | 0 |

So **#13292 and #13315 are genuinely test-only**, but **#13289 is not** — it changes `open-sse/services/adobeFireflySession.ts`, adding `browserRefreshEnabled()` which also suppresses the Chrome spawn under `NODE_ENV === "test"` / `VITEST` / `NODE_TEST_CONTEXT`. Production behaviour is unchanged when none of those are set, but it is a runtime path and should be reviewed as one. I have posted the same correction there.

Everything else in my previous comment stands: all three are rebased onto current `release/v3.8.51` (behind = 0), and the red checks are the repo-wide ones that also fail on the base branch and on other contributors' PRs.

diegosouzapw:
This is the stronger of the two #13290 fixes in flight — broader coverage (39 files vs. a
handful), a real regression test that opens an actual SQLite handle before asserting cleanup,
and a changelog fragment. Ran the new regression test and the migrated Adobe Firefly test at
your head, both green. Recommending this one merge and the narrower helper PR close in its
favor (with credit noted). One small heads-up: your change to
`8510-adobe-firefly-edits-route.test.ts` and #12829's change land in the same `test.after`
block, so expect a small textual conflict depending on merge order.

PR #13036, opened 2026-09-08
fix(security): honour a resolved source extension as the path-span endpoint — clears the stream-boundary base-red on release/v3.8.51

ggiak:
Clears one of the hard failures in #12732 (`🔴 Release branch not green: release/v3.8.51`): `tests/unit/stream-handler-public-error-boundary.test.ts`, reproduced on the clean tip `fcc2dcd1a6` before this branch existed. It is a **logic bug in the path redactor, not a sanitization-policy question** — the fix moves one line.

⚠️ base-red inherited: #12732

## What is failing

`stream-handler-public-error-boundary` → `pass 2 / fail 4`, deterministic, identical on Node 24 and Node 26 (it is also what keeps the nightly-compat issue #12713 red):

```
AssertionError: The input did not match the regular expression /Authorization: \[REDACTED\]/
actual: {"message":"Upstream failed at <path>", …}
```

Every *security* assertion passes — no secret, no `api_key`, no `/srv/omniroute/private`, no `dispatcher.ts`. Only the positive assertion that the redaction **marker** survive fails. This is over-redaction, not a leak.

## Root cause

#12506 (`2265ce761f`) replaced per-token path redaction with span redaction. In `findUnquotedPathEnd` (`open-sse/utils/errorPathRedaction.ts`):

- token 1 `/srv/omniroute/private/provider.ts:42:9` — contains a separator, so `:509-516` sets `resolvedExtensionEnd` to the `.ts` offset and clears `hasUnresolvedFragments`. **The endpoint is resolved.**
- token 2 `Authorization:` — no separator, no extension, not in `CLEAR_PROSE_BOUNDARIES`, so the final `else if (!isFirstToken)` at `:525-526` sets `hasUnresolvedFragments = true` again.
- `resolveEndpoint()` (`:467-472`) tests `hasUnresolvedFragments` **before** `resolvedExtensionEnd`, returns `value.length`, and the resolved endpoint is discarded. The rest of the first line collapses into `<path>` — and because `sanitizeErrorMessage` runs `redactErrorPaths` **before** `redactSensitiveErrorText`, the credential label is eaten before the labeller can mark it.

[...]

PR #13295, opened 2026-09-10
fix(backend): stop redaction truncating the message after a path

### LINKED ISSUE #13144: fix(backend): redactErrorPaths truncates the message after an API route, destroying the image-model 400's hint (CI red repo-wide)

abhisheksharma2411:
Fixes #13144. Taking option 1 as you chose it — and thank you for the `/zz` vs `/etc` probe, because **my proposed boundary was wrong** and that result is what showed it. The first segment was never the driver.

Two things were granting the swallow-the-line licence, and both had to go.

## 1. Separator evidence alone counted as unequivocal

```js
return failClosedAmbiguity || hasFilesystemEvidence ? value.length : -1;
//                            ^^^^^^^^^^^^^^^^^^^^^ every API route has slashes
```

`hasFilesystemEvidence` is set by the candidate's own separators, so an ordinary `/v1/x/y` in prose qualified and truncated everything after it. Now only a Windows path, a file URI, or a known POSIX root may swallow the line — which is what `:606` already said should happen.

## 2. The ambiguity branch ran *before* `resolvedExtensionEnd`

This is the part I did not see when I filed, and it is why fixing (1) alone left a second test red:

```js
if (hasUnresolvedFragments) { ... }                          // returned first
if (resolvedExtensionEnd >= 0) return resolvedExtensionEnd;  // unreachable
```

A path whose end is pinned exactly by a known extension was still treated as ambiguous the moment any prose followed, so the endpoint was discarded and the line swallowed. There is nothing to fail closed about when you know where the path stops:

```
before   Provider failed in <path>
after    Provider failed in <path> with api_key='[REDACTED]'
```

The whole path is still replaced, suffix included — `C:\Program Files\secret\a.ts and more` → `at <path> and more`, with no `secret` in the output.

**Both halves are independently load-bearing.** Reverting (1) fails the two route tests; reverting (2) fails the extension test.

## Fail-closed is narrowed, not weakened

`/etc/shadow copy failed` has nothing to anchor an endpoint on and still collapses to `reading <path>`.

One honest note on that guard test: **it cannot be killed by mutating either mechanism alone.** I disabled my `resolveEndpoint` line and it still passed; I disabled the caller's `mustFailClosed` branch and it still passed; only disabling *both* fails it. The two are mutually redundant — which is a property of the module rather than a flaw in the test, and it is also the clearest evidence that this change cannot expose a suffix these shapes did not already hide. I would rather say that than claim a mutation score the test does not have.

## Test results against base

| test | before | after |
|---|---|---|
| `chat-rejects-image-only-model` | **red** | **green** |
| `dashboard-request-failed-redaction` | red at its delivered-log assertion | that assertion **passes**; still red on a *second*, unrelated one |
| `tunnel-routes-error-sanitization` | red | unchanged — independent |

**On the second row:** base never reached the internal-log assertion because it failed earlier. With the delivered log fixed, the test progresses and reveals that the **internal** log is redacted where the probe expects it raw (`persisted.error === rawDiagnostic`). That is a persistence-path concern this branch does not touch, and it was masked before rather than introduced now. Worth its own issue — say the word and I'll file it with the repro.

**On the third row:** that one is a deliberate tripwire — *"sanitizeErrorMessage unexpectedly covers this now — if the shared sanitizer grew to handle it, simplify publicSafeTunnelError accordingly"* — firing on the **quoted**-path pass, which this change does not touch. It fails identically on base.

## Gates

- 5 new regression tests, all failing on base for the right reasons (2 for part 1, 1 for part 2) and passing here
- **836/840 pass** across the sanitization, redaction and error suites. The 4 are the two above plus `mcp-public-error-boundaries`, which passes in isolation on base *and* on this branch and only flakes under `--test-concurrency=8`
- `eslint` exit 0 on both files; pre-commit hooks (prettier, eslint, docs-sync, any-budget, tracked-artifacts) all green


abhisheksharma2411:
Rebased onto current `release/v3.8.51` (`c0f92ec9`) — it had gone into conflict after the base moved 40 commits. 0 behind now.

**The bug still reproduces on that base**, which I checked before bothering: dropped this PR's test file onto an untouched `c0f92ec9` and 3 of its 5 cases fail. So this isn't a stale fix looking for a home.

**The conflict was real, not textual.** `resolveEndpoint` gained an `ignoreAmbiguity` parameter in that window, plus two callers that pass it (`startsRedactedToken(...)` → `true`, and `hasRouteContextBefore(...)`). I kept all of that and applied this PR's two changes inside it:

1. `resolvedExtensionEnd` is checked **before** the ambiguity branch — a determinable extension pins the endpoint exactly, so there is nothing left to fail closed about.
2. `hasFilesystemEvidence` is dropped from the swallow licence — separator evidence alone is not an unequivocal filesystem prefix, and every API route carries slashes.

`failClosedAmbiguity` still swallows, so the Windows `Files\secret` case this function exists for is untouched (its test still passes).

### One case is better but not fully restored, and I'd rather say so

The headline message now keeps its tail, which is the truncation this issue names:

```
before: "...cannot be used on <path>"                     ← everything after it gone
after:  "...cannot be used on <path>. Use POST /v1/images/generations instead."
```

But the route itself is still replaced with `<path>` in *that* message, while the same route in isolation survives verbatim (`on /v1/chat/completions. Contact support.` — second test, passing). The difference is the quoted model slug earlier in the line: `'huggingface/stabilityai/stable-diffusion-xl-base-1.0'` carries separators, so the route is read with prior path context.

I have **not** tried to change that. It's a narrower question than the truncation — whether a bare API route should be redacted at all — and given you disproved my first diagnosis on this module with a probe I should have run myself, I'm not keen to guess at it a second time. The test asserts the remediation sentence survives and records the `<path>` substitution explicitly, so the current behaviour is pinned rather than quietly accepted.

If you'd rather the route survive too, that's a follow-up and I'd want your read on where the context check should stop first.

### Verification

```
error-path-redaction-route-truncation-13144 + auggie-cli-not-found-shell-exit-12645   7 pass, 0 fail
  (the only two suites that import errorPathRedaction)
eslint open-sse/utils/errorPathRedaction.ts + the test                                 clean
```

diegosouzapw:
Confirmed real bug and correct fix — two independent causes for the redactor swallowing the
rest of a log line after a path, both proven load-bearing (reverting either alone breaks a
different test). Your mechanism for cause (2) is identical to what #13036 independently found
for #12732, and this PR's fix supersedes it (also fixes #12732's failing test, plus your own
#13144 case). One gap: `errorPathRedaction.ts` is mutation-tested — please register your new
test file in `stryker.conf.json`'s `tap.testFiles`, following #13036's precedent for the
sibling file. Once that's in, this is ready.

abhisheksharma2411:
Done — `tests/unit/error-path-redaction-route-truncation-13144.test.ts` is registered in `stryker.conf.json` `tap.testFiles`, pushed at `08c95c91`. Sorry for the silence between your review and this; I lost access for most of the day and you'd already moved on by the time I was back.

Inserted in the alphabetical run beside `error-sensitive-redaction.test.ts`, following #13036's precedent:

```
tests/unit/error-message-sanitization.test.ts
tests/unit/error-path-redaction-route-truncation-13144.test.ts   ← added
tests/unit/error-sanitizer-sk-key-qv45.test.ts
```

I left the rest of the list alone — it has a second, unsorted appended group further down, and re-sorting a file this PR needed one line in seemed like the wrong trade.

Also rebased onto `df87e936` on the way, since the branch had gone into conflict. The merge was not textual: `resolveEndpoint` had gained an `ignoreAmbiguity` parameter and two callers passing it. I kept those and applied this PR's two changes inside — `resolvedExtensionEnd` checked before the ambiguity branch, and `hasFilesystemEvidence` dropped from the swallow licence. `failClosedAmbiguity` still swallows, so the Windows `Files\secret` case is untouched and its test still passes.

Verification at that head: 7 pass / 0 fail across the two suites that import `errorPathRedaction` (this one and `auggie-cli-not-found-shell-exit-12645`), eslint clean on both touched files.

**One case is better but not fully restored, and I'd rather say so than let it read as fixed.** The headline message keeps its tail now:

```
before: "...cannot be used on <path>"                     ← everything after it gone
after:  "...cannot be used on <path>. Use POST /v1/images/generations instead."
```

but the route itself still becomes `<path>` in *that* message, while the same route in isolation survives verbatim (second test, passing). The difference is the quoted model slug earlier in the line — `'huggingface/stabilityai/...'` carries separators, so the route is read with prior path context.

I haven't touched that. It's a narrower question than the truncation, and after you disproved my first diagnosis on this module with the `/zz` probe I should have run myself, I'd rather ask than guess twice. The test pins current behaviour explicitly rather than quietly accepting it. If you want the route to survive too, that's a follow-up and I'd want your read on where the context check should stop.

abhisheksharma2411:
The `stryker.conf.json` registration is already in this PR — worth pointing at explicitly since it's a one-line change easy to miss in the diff:

```
$ git diff --stat origin/release/v3.8.51...HEAD
 changelog.d/fixes/13295-redaction-truncates-after-path.md          |  1 +
 open-sse/utils/errorPathRedaction.ts                               | 17 +++-
 stryker.conf.json                                                  |  1 +
 tests/unit/error-path-redaction-route-truncation-13144.test.ts     | 97 +++++++++++
```

`tests/unit/error-path-redaction-route-truncation-13144.test.ts` sits next to `error-sensitive-redaction.test.ts` in `tap.testFiles`, following #13036's precedent for the sibling file. The base doesn't carry the entry (`git show origin/release/v3.8.51:stryker.conf.json | grep -c error-path-redaction-route-truncation` → `0`), so it comes from this branch.

Also merged the current release line in — the branch was behind the merge wave and is mergeable again.

Worth flagging separately, since it came out of that merge and affects this file's gate: `stryker.conf.json` on the base names `tests/unit/plugin-sandbox-permissions.test.ts`, which no longer exists — one dangling path out of 428. `check-mutation-test-coverage` guards the direction you asked me about here (a test covering a mutated module but absent from `tap.testFiles`) but not the reverse, and Stryker resolves the list into its sandbox, so a dangling entry costs coverage without failing loudly. Fixed it plus a guard in #13814 rather than folding it in here.

diegosouzapw:

diegosouzapw:
Maintainer note for the merge: this will be squash-merged with `Co-authored-by: ggiak <20743694+ggiak@users.noreply.github.com>` — the `resolvedExtensionEnd` reordering here is the same one-line mechanism @ggiak landed first in #13036 (for #12732). #13036 will be closed once this merges, with credit preserved through that trailer. No action needed from you.

PR #13296, opened 2026-09-10
fix(test): add Windows-safe DATA_DIR teardown helper to fix EPERM failures

### LINKED ISSUE #13290: fix(backend): Windows: ~29 unit tests fail with EPERM because test.after() removes DATA_DIR without closing the DB

jonlwheat2-gif:
## Summary

Fixes #13290 — ~29 unit tests failed on Windows with `EPERM` because `test.after()` removed `DATA_DIR` without closing the SQLite DB first.

## Changes

1. **New helper**: `tests/_setup/withIsolatedDataDir.ts` provides `cleanupIsolatedTestEnv()` that:
   - Calls `core.resetDbInstance()` to close the database connection
   - Waits 50ms for Windows to release file handles (`.sqlite`, `.sqlite-wal`, `.sqlite-shm`)
   - Then removes the temp directory with retries

2. **Migrated 5 confirmed failing tests**:
   - `tests/unit/executor-registry.test.ts` ✅
   - `tests/unit/api-key-lifecycle.test.ts` ✅
   - `tests/unit/provider-health-matrix.test.ts` ✅ (no EPERM, 2 pre-existing logic failures remain)
   - `tests/unit/reasoning-routing.test.ts` ✅ (no EPERM, 1 pre-existing logic failure remains)
   - `tests/unit/zcode-executor.test.ts` ✅

3. **Documentation**: `docs/guides/WINDOWS_EPERM_TEST_FIX.md` with migration guide for the 93 latent test files.

## Testing

All 5 test files now pass on Windows without EPERM errors. Pre-existing logic failures are unrelated to this fix.

```bash
node --import tsx/esm --test tests/unit/executor-registry.test.ts
node --import tsx/esm --test tests/unit/api-key-lifecycle.test.ts
node --import tsx/esm --test tests/unit/zcode-executor.test.ts
```

## Related

- Fixes the second EPERM cause tracked in #13290
- #13289 fixed the first cause (Chrome spawned under test runner)
- Other 93 test files setting DATA_DIR can adopt this helper to prevent future EPERM failures

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -296,3 +296,6 @@ docker-compose.yml.bak
 
 # check:install-upgrade work trees (~12 GB, disposable)
 /.install-upgrade/
+
+# graft's local graph cache — regenerable, not committed (run `graft build`).
+/graft/
--- a/.ignore
+++ b/.ignore
@@ -1,0 +1,5 @@
+# graft's cards are gitignored but should stay greppable: ripgrep reads
+# .ignore before .gitignore, so this re-admits the tree to search only.
+!graft/
+graft/.cache/
+graft/.graph/
--- a/docs/guides/WINDOWS_EPERM_TEST_FIX.md
+++ b/docs/guides/WINDOWS_EPERM_TEST_FIX.md
@@ -1,0 +1,144 @@
+# Windows EPERM Test Fix Documentation
+
+## Issue
+
+**GitHub Issue**: [#13290](https://github.com/diegosouzapw/OmniRoute/issues/13290)
+
+~29 unit tests failed on Windows with `EPERM: Permission denied` during teardown. The root cause was that `test.after()` removed the temporary `DATA_DIR` without closing the SQLite database connection first. On Windows, a directory cannot be deleted while it contains open file handles (the main `storage.sqlite` plus WAL/SHM sidecars: `storage.sqlite-wal`, `storage.sqlite-shm`).
+
+## Root Cause
+
+Test files created isolated `DATA_DIR` using `fs.mkdtempSync()` and set `process.env.DATA_DIR`. During tests, importing database modules opened SQLite connections. The final `test.after()` called `fs.rmSync()` on the directory **without** calling `core.resetDbInstance()` to close the DB first.
+
+```typescript
+// BEFORE (fails on Windows)
+test.after(() => {
+  fs.rmSync(TEST_DATA_DIR, { recursive: true, force: true, maxRetries: 5, retryDelay: 100 });
+});
+```
+
+Error:
+
+```
+Error: EPERM, Permission denied: \?\C:\Users\...\Temp\omniroute-executor-registry-CwqmB0
+    at Object.rmSync (node:fs:1283:18)
+    at TestContext.<anonymous> (tests/unit/executor-registry.test.ts:21:6)
+```
+
+## Solution
+
+Created a shared teardown helper at `tests/_setup/withIsolatedDataDir.ts` that:
+
+1. **Creates isolated temp `DATA_DIR`** with a unique prefix
+2. **Provides `cleanupIsolatedTestEnv()`** that:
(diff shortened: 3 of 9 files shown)


diegosouzapw:
Thanks for tackling #13290 — the root-cause diagnosis here is right. #13292 (same issue)
ended up covering the same 5 files here plus 34 more, with a smaller per-file diff, a real
regression test that proves the close-before-remove ordering, and a changelog fragment. To
avoid two competing helpers for the same problem, we're going with #13292 and closing this
one. Two things worth keeping in mind for next time: `cleanupIsolatedTestEnvSync` in your
helper doesn't actually close the DB despite its name/header comment, and
`createIsolatedTestEnvSync`/`cleanupIsolatedTestEnvSync` end up unused by any of the 5 files
you migrated — worth trimming in future PRs. Appreciate the migration-guide doc regardless,
happy to see the remaining files land via #13292's helper.

PR #13299, opened 2026-09-10
fix(providers): fetch live Qwen and Alibaba Token Plan catalogs

JxnLexn:
## Summary

Qwen Cloud Token Plan and Alibaba Token Plan have neither a model-discovery config nor a registry `modelsUrl`. Clicking Sync Models therefore returned `API unavailable — using local catalog` without attempting a live request. The import then found only the same built-in IDs and reported that nothing new was available.

This change fetches the public Personal Token Plan catalog used by the official [Qwen pricing page](https://www.qwencloud.com/pricing/token-plan): `zeldaEasy.bmp.bmpTokenPlanServcie.modelIdList` with `edition: PERSONAL`.

- Select the public gateway and console identity for Singapore/Beijing. Alibaba Singapore uses its own gateway; Beijing uses the shared regional Qwen Personal Plan catalog.
- Do not send inference API keys or browser cookies. Restrict requests to fixed public hosts through `safeOutboundFetch`, preserving configured proxy handling and timeouts.
- Validate the nested gateway success envelope, deduplicate IDs, and use the existing DashScope text-model classifier to keep audio/image/video models out of chat imports. New chat IDs do not depend on a static model allowlist.
- Reuse existing discovery persistence, cached reads, manual refresh, and auto-fetch controls. HTTP errors, invalid JSON/envelopes, and empty or media-only results fall back without overwriting the previous catalog.
- Return `catalogMode: live_token_plan_catalog` and `catalogScope: product` on successful live discovery.

Before/after on a standalone deployment: Alibaba returned 6 local models and Qwen returned 7; both now return and persist 9 live chat models, including `qwen3.8-flash` and `deepseek-v4-pro-0813`.

## Validation

- [x] Change type: provider model discovery
- [x] `npm run typecheck:core` on both fork and release bases
- [x] Targeted ESLint on all three changed files with the repository's `config/quality/eslint-suppressions.json`
- [x] Prettier and `git diff --check`
- [x] Reconciled with current `release/v3.8.51`
- [x] Production-code changes include automated regression tests
- [x] Live helper calls: both providers × both regions returned HTTP 200 and 9 chat models
- [x] [Standalone AMD64 image build](https://github.com/JxnLexn/OmniRoute/actions/runs/34538271807) and deployment completed
- [x] Deployed API: both `POST /sync-models?mode=import` calls returned 200; refreshed model responses were `source: api`; subsequent reads were `source: cache`; each connection persisted the same 9 IDs

Unit suites and the broad CI matrix were not run locally; they remain CI validation. No PR checks were awaited for deployment verification.

## Tests Added Or Updated

`tests/unit/token-plan-model-discovery.test.ts` covers regional public request construction, credential isolation, exclusion of regular Alibaba/Qwen providers, dynamic chat IDs, media filtering, deduplication, nested HTTP-200 gateway failures, invalid IDs, and empty/media-only catalogs.

## Coverage Notes

The new tests cover request construction and response validation/filtering. The deployed live checks additionally cover route dispatch, successful sync, persistence, and cache reads. No coverage measurement was run locally.

## Reviewer Notes

This is the live **Personal Plan product catalog**, not account entitlement or remaining-quota validation. It returns model IDs, not context/token limits. The console gateway is an implementation detail of the official frontend rather than a documented OpenAI-compatible models API; the fallback protects existing catalogs if that interface changes.

Regular Alibaba/Qwen providers, inference transports, quota authentication, and static media registries are unchanged. No migrations, new dependencies, or additional credentials are required. The PR contains only the Token Plan fix, without the fork's other hotfixes or image workflow.


diegosouzapw:
Nice fix for a real gap — Token Plan catalogs never even attempted a live fetch before.
The public-only SSRF guard and graceful fallback on gateway failure are exactly right. Ran
`tests/unit/token-plan-model-discovery.test.ts` locally: 3/3 pass. One thing missing before
merge: there's no `changelog.d/fixes/` fragment in this PR — could you add one? Also worth
calling out in the description (you already note it, just flagging for visibility) that this
relies on the Qwen/Alibaba console's internal gateway rather than a documented API, so a
future upstream change there falls back to the local catalog rather than erroring loudly.

JxnLexn:
Thanks for the review. Added `changelog.d/fixes/13299-token-plan-live-catalog.md` in d67e04a6d.

Re-ran `node --import tsx/esm --test tests/unit/token-plan-model-discovery.test.ts`: **3/3 passed**. `npm run check:changelog-integrity` also passes.

Agreed on the API stability caveat: these are authenticated internal console gateways, not a documented public catalog API. The fragment explicitly describes the gateway and local-catalog fallback; no runtime behavior was changed in this follow-up.

PR #13307, opened 2026-09-11
fix: origin-IP disclosure, next 16.3.4 RCEs, FTS5 search correctness, 413 fallback, migration collision

themedexperiencesusa:
⚠️ base-red inherited: #12732

Nine fixes found while setting OmniRoute up as the shared gateway for Hermes and several IDE clients. Each is independently committed with its rationale in the commit message.

## Security — origin-IP disclosure (3 commits)

Three call sites could send the client origin IP to an upstream provider:

- **`upstreamHeaders.ts`** — operator-set custom upstream headers could carry the whole forwarding set (`x-forwarded-for`, `x-real-ip`, `cf-connecting-ip`, `forwarded`, `via`, …). Extends the canonical `FORBIDDEN` denylist, mirroring the scrubbers already used on the Antigravity and Cursor CLI paths so the protection applies to every provider. Covered by `tests/unit/upstream-headers-sanitize.test.ts` (6 passing).
- **`openapi/try`** — kept its own partial denylist; now routes through `isForbiddenCustomHeaderName()`, leaving only proxy-auth headers as local extras.
- **`relay/bifrost`** — sent `x-relay-client-ip` unconditionally, leaking the origin IP whenever the sidecar was remote. Now forwarded only when the operator opts in (`BIFROST_FORWARD_CLIENT_IP=1`) **and** `BIFROST_BASE_URL` is loopback/private.

## Security — dependencies

`next` **16.3.1 → 16.3.4**, closing two critical advisories: unauthenticated RCE on Windows-hosted servers, and unauthenticated RCE in the Image Optimization API via AVIF. `npm audit` critical count 2 → 0. `eslint-config-next` moved in lockstep; `typecheck:core` passes.

## Correctness — FTS5 search returned wrong results

Binding a search term as a parameter (`MATCH ?`) stops SQL injection but not FTS5's own grammar. Measured against a real FTS5 table, **9 of 13 inputs a user can plausibly type** raised `fts5: syntax error` — a trailing quote, a leading `-`, a bare `*`, `col:` text.

The user-visible bug was worse than a crash: in `retrieval.ts` the throw was caught and the code fell through to a plain date-ordered query, so searching `foo-bar` silently returned the **most recent** memories instead of **matching** ones, with no error surfaced.

Adds `src/lib/db/ftsQuery.ts` — quotes each token as an FTS5 string literal (doubling embedded quotes), bounds length and term count, returns `null` when nothing searchable survives. This generalises the local fix already present in `copilot/codegraphKnowledge.ts:140`. Applied at all three memory call sites; `searchHybrid` falls back to a no-match term so an unsearchable text half no longer fails the whole fused query and discards the vector hits. `tests/unit/fts-query-sanitize.test.ts` — 10 passing.

## Resilience — 413 treated as terminal

Providers with a tokens-per-minute cap (Groq among them) answer an oversized turn with **413**, not 429. `checkFallbackError()` did not list 413 as retryable, so the request failed hard instead of falling back. Adds `PAYLOAD_TOO_LARGE` to the retryable set, returns a `MODEL_CAPACITY` retryable fallback, and recognises "tokens per minute" / "tpm" as context-overflow patterns.

## Database — migration prefix collision

Two pairs of migrations shared a numeric prefix, so the runner refused to start with "Migration version collision detected" and every API route returned 500. Renames the second file of each pair (`123 quota_auto_ping → 163`, `126 verification_chain → 164`) and adds `RENAMED_MIGRATION_COMPATIBILITY` entries so databases that already recorded the old numbers re-home the applied row instead of tripping the name-mismatch check.

## Dashboard

A block of page-level header markup had been pasted inside the `groupModels.map()` callback in `ProviderModelsModal`, between `return (` and the `<div key={m.id}>` it was meant to return — rendering the page header once per model row. Also normalises the file from CRLF back to LF; the wholesale CRLF rewrite made git report 2618 changed lines and buried the real 12-line deletion.

## Also included

One commit adds a staged-credential pre-commit gate and the next reverts it. Kept as a pair rather than dropped, because the revert rationale is worth recording: measured against the full tracked tree it produced **1088 findings, of which zero were real credentials**. After suppressing docs and test fixtures, 8 remained — all false positives (OAuth *endpoint* URLs matched on the word `token:`, a Japanese UI label, a URL path fragment, a test mock). A regex cannot distinguish a token endpoint from a token value, and the rules used 11 open-ended quantifiers against attacker-influenceable text. GitHub secret scanning validates candidates against the issuing provider instead, and `check-public-creds` / `resolvePublicCred()` already cover the embedded-credential case.

## Verification

- `tests/unit/fts-query-sanitize.test.ts` — 10 passed
- `tests/unit/upstream-headers-sanitize.test.ts` — 6 passed
- Pre-commit gates (`lint-staged`, `check-docs-sync`, `any-budget:t11`, `tracked-artifacts`) passed on every commit; no `--no-verify`
- `npm audit` — 0 critical (was 2), `next` clean

## Note on the base

`release/v3.8.51` was red at branch time (#12732: 5 inherited hard failures — unit `#12058`, integration codex-cache, package-artifact provenance, tarball boot-smoke, `check:agent-skills-sync`). None originate here; flagged per the base-green protocol so reviewers don't chase them.


themedexperiencesusa:
Closing: opened against a stale fork point (2026-08-23) while the base had moved to 2026-09-10, 3813 commits ahead.

Three of the changes are obsolete or wrong against the current base:
- The FTS5 escaping is already fixed upstream (sanitizeFts5Query in memory/retrieval/scoring.ts), same approach and call sites.
- Migrations 163/164 are already taken by radar_feed_cache_generated_at and retire_microsoft_designer_web, so the renames here would recreate the collision they were meant to fix.
- The next bump is described as closing the two RCEs, but the base is already on the patched 16.3.3.

The branch also carried two commits that were not part of this work.

Two fixes here do still apply to the current base and will be re-submitted from it: the origin-IP header denylist (upstreamHeaders.ts + openapi/try + relay/bifrost) and treating 413 payload-too-large as a retryable TPM rate limit.

Issue #3229, opened 2026-06-05
[BUG] agy/gemini-3.1-pro-high & -low return 400 (speculative tier-suffix, unaliased) masked as empty chat.completion


diegosouzapw:
## Bug

`agy` (Antigravity CLI) models `gemini-3.1-pro-high` and `gemini-3.1-pro-low` return `[400]` with a misleading empty `chat.completion` envelope. Reported by @mikmaneggahommie in #3182.

## Root cause (verified v3.8.9 + still present on release/v3.8.10)

1. `gemini-3.1-pro-high` / `gemini-3.1-pro-low` are registered in the agy catalog (`open-sse/config/agyModels.ts:39,48`).
2. They have **no entry** in `ANTIGRAVITY_MODEL_ALIASES` (`open-sse/config/antigravityModelAliases.ts`), whose own comment (lines 101-103) states the `-high`/`-low` suffix convention "was speculative and caused 404 for all gemini-3.x models. Only plain IDs ... are proven working."
3. `cleanModelName()` → `resolveAntigravityModelId()` returns the ID unchanged when no alias exists (`open-sse/executors/antigravity.ts:405-436`), so `gemini-3.1-pro-high` is sent verbatim to the upstream Antigravity API, which rejects it (400/404).
4. **Masking:** for non-streaming requests the executor feeds the error response straight into `collectStreamToResponse()` with no `status >= 400` gate, producing a synthetic body `{"object":"chat.completion","content":"","finish_reason":"stop"}` carrying the upstream 4xx status (`antigravity.ts` non-stream branch, ~line 1361). So the user sees an empty success-shaped envelope instead of a real error.

## Proposed fix

- **Primary:** in `agyModels.ts`, either drop `gemini-3.1-pro-high`/`-low`, or add aliases `"gemini-3.1-pro-high": "gemini-3.1-pro"` / `"gemini-3.1-pro-low": "gemini-3.1-pro"` to `ANTIGRAVITY_MODEL_ALIASES` (mirroring how `gemini-3-pro-preview` → `gemini-3.1-pro`). Audit the rest of the agy catalog for other unaliased `-high`/`-low` suffix IDs.
- **Secondary (masking):** add a `status >= 400` / `!response.ok` gate before the non-stream `collectStreamToResponse()` call so upstream 4xx/5xx are returned as real errors, not empty completion envelopes. (Verify against the current `!response.ok` checks already present on release/v3.8.10.)

## Workaround

[...]

PR #13311, opened 2026-09-11
fix(antigravity): retain bounded validation diagnostics

iamzidar:
## Summary

- retain a bounded, allowlisted classification for terminal Antigravity upstream validation failures while keeping raw Google response bodies, headers, URLs, and provider prose out of client responses and call logs
- bind the internal diagnostic to the exact provider response across retries, signature recovery, synthetic responses, and fallback replacement paths
- persist diagnostic-only call-log payloads through a narrow logger so fields such as `schemaKeyword` survive without weakening the sanitizer for arbitrary upstream bodies

## Related Issues

- Related to #3229

## Validation

- [x] Change type: provider
  - focused Antigravity, logging, timeout, non-streaming, and provider-pipeline tests: 110 passed, 0 failed
  - duplication, changelog-integrity, test-discovery, complexity, cognitive-complexity, error-helper, public-creds, test-runner-api, cycles, dead-code, type-coverage, tracked-artifacts, and lockfile gates: passed
- [ ] `npm run lint`
  - changed-file ESLint outside known inherited files passed; current release base remains red with the inherited lint findings tracked in #12732

⚠️ base-red inherited: #12732

The inherited `check:file-size` failure is `open-sse/utils/stream.ts: 3115 > frozen 3098`; this branch does not touch that file. The full unit-suite control and branch runs both failed to terminate after roughly two hours and were stopped after process ownership was verified; output-to-date showed 28 failures on each side, with 27 shared and neither differing test in a modified file.

## Tests Added Or Updated

- `tests/unit/agy-gemini-400-3229.test.ts`
- `tests/unit/antigravity-streaming-error-body-sanitized-2461.test.ts`
- `tests/unit/calllogs-format-split.test.ts`
- `tests/unit/chatcore-upstream-timeouts.test.ts`
- `tests/unit/non-streaming-provider-leg-diagnostic.test.ts`
- `tests/unit/non-streaming-provider-leg.test.ts`
- `tests/unit/provider-execution-pipeline.test.ts`
- `tests/unit/provider-request-failure-pipeline.test.ts`
- `tests/unit/_helpers/nonStreamingProviderLegFixtures.ts`

## Coverage Notes

- projector coverage exercises known scalar extraction, malformed and non-object JSON, unsupported scalar types, unknown-key omission, credential/path/data-URL rejection, and the `additionalProperties` schema classification
- streaming and non-streaming tests prove the client envelope stays generic and no raw provider bytes are retained
- pipeline, fallback, retry, timeout, synthetic-response, and signature-recovery tests cover diagnostic replacement and clearing at response-ownership boundaries
- the call-log integration path asserts only the bounded diagnostic is persisted and synthetic prompt/tool/project secret markers never appear

## Reviewer Notes

- this is the diagnostics-only checkpoint; it intentionally does not modify `open-sse/translator/request/openai-to-gemini.ts`
- no live provider reproduction is included in this change; one synthetic reproduction will be sent only after this checkpoint passes CI and is deployed normally
- the client-facing Antigravity error remains fixed text, and the diagnostic producer projects to a closed set of scalar/enumerated fields before logging
- no model priority, account configuration, fallback policy, provider safeguard, or thought-signature bypass behavior changes
- `config/quality/file-size-baseline.json` records justified growth only for the internal result contract and the response-replacement lifecycle owner


diegosouzapw:
This is careful, well-reasoned work — the allowlist projection + the two-gate sanitization
check + the fix to the `protectErrorPayloadForLog` key-name heuristic silently eating
`schemaKeyword` are all exactly right. Two small things before merge: (1) GitHub shows this as
conflicting against the current release branch — it looks like pure line-drift in
`.dockerignore`/`.env.example` from unrelated recent merges, so a rebase should clear it
cleanly; (2) the PR references #3229, which is actually already closed/shipped in v3.8.11 —
worth a one-line update in the description so reviewers don't read this as reopening that fix
rather than extending it. I ran your larger pipeline test files under a loaded shared box and
saw a couple of intermittent timing-looking failures (different assertions each run) that don't
match your own "110 passed, 0 failed" CI run — just flagging in case it's worth a closer look on
a quieter runner, not blocking.

Issue #13313, opened 2026-09-11
fix: Two suites fail on base: stale assertions contradict current product behaviour (Jina catalog prefix, pt-BR output styles)


anhtahaylove:
## Summary

Two suites fail on `origin/release/v3.8.51` because the **tests** encode
expectations the product deliberately changed. In both cases the product output
looks correct and the assertion is stale. Neither is a regression from any open
PR — both reproduce on a clean base checkout.

---

### 1. `tests/unit/models-catalog-route.test.ts` — Jina catalog prefix

Failing test: `v1 models catalog does not duplicate custom Jina specialty models`

```
actual:   'jina/jina-embeddings-v5-text-small'
expected: 'jina-ai/jina-embeddings-v5-text-small'
```

The catalog id is built from the provider **alias**, and the Jina entry declares
`alias: "jina"` in `src/shared/constants/providers/apikey/specialty-media.ts:169`.
So `jina/...` is what the product is designed to emit.

The strongest evidence that the assertion is simply stale: the test immediately
**above** it in the same file asserts the opposite and passes:

```js
// passes today
assert.equal(visibleJinaEmbeddingRows[0].id, "jina/jina-embeddings-v5-text-small");
// fails today, same file
assert.equal(visibleJinaEmbeddingRows[0].id, "jina-ai/jina-embeddings-v5-text-small");
```

Timeline confirms it — the test was written in v3.7.9 (2026-05-04) and the
`alias: "jina"` entry landed later in v3.8.36 (2026-06-25); the second assertion
was never updated.

Note `jina-ai` is still correct for `owned_by` and for the connection/provider id
(34 occurrences in `src/`); only the catalog **id prefix** uses the alias. So the
fix is the two `assert.equal` id strings, not a rename.

Running the test in isolation fails identically, so this is not test pollution.

---

### 2. `tests/integration/chatcore-compression-integration.test.ts` — pt-BR packs

Two failing tests expect a Portuguese output-style string:

```
expected: /Responda conciso/
actual:   '[OmniRoute Output Styles]\nRespond concise. Drop filler, ...'
```

Both set up the combo with `languagePacks: ["pt-BR"]` but a `languageConfig` of:

```js
languageConfig: { enabled: false, defaultLanguage: "en", autoDetect: true, enabledPacks: ["en"] }
```

`resolveOutputStyleLanguage` in `open-sse/services/compression/outputStyles/apply.ts:70`
[...]

PR #13315, opened 2026-09-11
test: realign two stale assertions with product behavior (#13313)

anhtahaylove:
Two test files assert things the product does not do. Both fail on
`release/v3.8.51` before this branch, and neither is a product bug — the
expectations are simply stale. Split out of #13292 so the teardown work can be
reviewed sepa...[truncated]


anhtahaylove:
@diegosouzapw — still open and MERGEABLE, test-only (`git diff origin/release/v3.8.51 -- src/` is empty). Grouped context for this PR and the other two left over from the 10–11 Sep batch is on #13292: https://github.com/diegosouzapw/OmniRoute/pull/13292#issuecomment-5653579490

Short version: #13187 has merged, so the concurrency blocker is gone; the red checks here are the repo-wide ones that also fail on `release/v3.8.51` and on other contributors' PRs, not something specific to this branch. Happy to rebase or close on request.

anhtahaylove:

I called this PR test-only and backed that with `git diff origin/release/v3.8.51 -- src/` being empty. That check was wrong: `src/` exists (2270 non-test `.ts` files), but it is not the only product tree — `open-sse/` holds another 1558. Scoping the diff to `src/` alone silently hid every change under `open-sse/`. I should have diffed the whole tree.

Re-checked properly, **this PR really is test-only** — both files are tests:

- `tests/integration/chatcore-compression-integration.test.ts`
- `tests/unit/models-catalog-route.test.ts`

So the conclusion holds here, but it was reached by a broken check, and the same sentence was wrong on #13289 — that one changes `open-sse/services/adobeFireflySession.ts`, which is product code. Correction posted there too.

Everything else stands: this branch is rebased onto current `release/v3.8.51` (`1894077e8`, behind = 0), `tests/unit/models-catalog-route.test.ts` reports 44 pass / 0 fail locally, and the red checks are the repo-wide ones that also fail on the base branch.

diegosouzapw:
Thanks for digging into this. I reproduced both failures on the release tip, and I think the pt-BR half of the diagnosis stops one step short. Evidence below.

The fixture's global `languageConfig` is not what the combo path reads. `applyCompressionComboConfig` in `open-sse/handlers/chatCore.ts` rewrites it from the compression combo before output styles run. With `languagePacks: ["pt-BR"]` it produces `enabled: true`, `defaultLanguage: "pt-BR"`, `enabledPacks: ["pt-BR"]`, and it keeps `autoDetect: true` from settings. So `resolveOutputStyleLanguage` does not return `"en"` on its first line. It reaches the auto-detect branch, which gives:

```
resolveOutputStyleLanguage({ enabled: true, defaultLanguage: "pt-BR", autoDetect: true,  enabledPacks: ["pt-BR"] }, body)  // "en"
resolveOutputStyleLanguage({ enabled: true, defaultLanguage: "pt-BR", autoDetect: false, enabledPacks: ["pt-BR"] }, body)  // "pt-BR"
```

The user turn is `"Resuma esta implementacao."`, which is Portuguese but matches none of the `LANGUAGE_HINTS` words. `detectCompressionLanguage` returns `"en"` whenever nothing scores ("zero hits → English"). That fallback then overrides the pack the combo explicitly assigned.

Setting the global config to `pt-BR` with `autoDetect: false` makes the test pass. It does so by bypassing the combo-pack path this test was written to cover, not by exercising it. Whether a zero-evidence detection should defer to the combo's `defaultLanguage` is a product call, so I've left it with the maintainer (noted in #13678). The Jina catalog-prefix half looks right to me.

diegosouzapw:
Nice, focused split from #13292 — both stale-assertion fixes check out. Ran
`models-catalog-route.test.ts` (44/44) and `chatcore-compression-integration.test.ts`
(11/12, the 1 skip is pre-existing) at your branch head and both are green. Looks
merge-ready to me.

PR #13317, opened 2026-09-11
fix(playground): improve Compare response scrolling and copy actions

### LINKED ISSUE #13316: fix: Playground Compare response panels should scroll independently
## Problem

In Playground Compare, a long model response expands the response area instead of keeping each model column height-constrained and vertically scrollable. The response body already has `overflow-y-auto`, so the visible failure comes from the surrounding grid/flex height constraints rather than a missing overflow rule alone.

This can cause the model header, status, and metrics to move out of view and prevents independent scrolling between compared responses.

## Reproduction

1. Open Playground and select Compare mode.
2. Add at least two models.
3. Return a long response for each model.
4. Try to scroll one response body without moving the other column or the page.
5. During streaming, scroll upward and observe whether new chunks force the response back to the bottom.

## Expected behavior

- Each model response body scrolls vertically and independently.
- The model name, status, metrics, and column actions remain visible.
- Short, long, and streaming responses render correctly.
- Streaming follows the bottom only while the user remains at the bottom.
- Scrolling upward during streaming is respected.
- The existing horizontal Compare layout remains unchanged.
- Each model column provides a copy action beside its remove action that copies only that model response.

## Regression coverage

Add focused component tests for the per-model copy action and browser coverage for independent scrolling, fixed headers and metrics, horizontal layout retention, and streaming scroll behavior.

ventulus95:
## Summary

- constrain the Compare grid and flex height chain so every model response body owns its vertical overflow while preserving the horizontal layout
- keep the model name, status, metrics, and actions outside the scrolling response body
- cover long-response independent scrolling and user-controlled streaming follow behavior in Playwright
- add a per-model copy action beside the remove action, including disabled and success states
- remove the Compare component tests from the Vitest exclusion list so the regression suite actually runs

## Root cause

The response body already used `overflow-y-auto`, but its grid/flex ancestors did not consistently allow shrinking. Without `min-h-0` through the height chain and a bounded grid row, the response content expanded the panel instead of producing an internal scroll area.

Before the fix, browser measurements showed `clientHeight === scrollHeight === 4308` for long output even though computed overflow was `auto`.

## Validation

- Compare component Vitest suite: 20 tests passed
- focused Playwright tests in system Google Chrome: 2 tests passed
  - long responses scroll independently while headers and metrics stay fixed
  - streaming does not force the user back to the bottom after scrolling up
- `npm run lint`: passed with only the existing suppression notice
- Prettier and `git diff --check`: passed
- repository pre-commit checks passed for both commits

Fixes #13316


diegosouzapw:
Thanks for this — the copy-to-clipboard addition on individual Compare columns is a nice,
well-tested UX improvement (reusing the existing `useCopyToClipboard` hook is exactly right).
Heads up though: the scrolling half of this PR (the `min-h-0`/`overflow` changes to
`CompareColumn.tsx` and `CompareTab.tsx`) was already fixed and merged separately in #13532, and
the `vitest.config.ts` exclusion-list hunk is also already gone on the current tip (via #13204).
Could you rebase onto `release/v3.8.51` so the diff narrows down to just the copy-button feature
and its test? That should resolve the current merge conflict too. Full credit for the feature is
yours either way — just want to avoid double-applying/reverting the already-merged scroll fix.

ventulus95:
Done — I rebased the branch onto the latest release/v3.8.51 and removed the already-merged scrolling and Vitest configuration changes. The diff is now limited to the per-column copy action and its unit tests. Thanks for the guidance!

diegosouzapw:
Thanks @ventulus95 — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #13322, opened 2026-09-11
fix(cli): read commander's negated --no-* flags as opts.<name> === false

datrixlab:
## Summary

Commander stores a negated option such as `--no-history` as `history === false`. It never sets `noHistory`. Three commands read the `noX` name, so each flag is accepted without error and does nothing:

| command | reads | what actually happens |
| --- | --- | --- |
| `omniroute contexts export --no-secrets` | `opts.noSecrets` | the export still contains every context's `accessToken` and `apiKey`; `redactContextSecrets()` is never reached |
| `omniroute chat --no-history` | `opts.noHistory` | the prompt and response are still appended to `cli-history.jsonl`, which the help text promises not to do |
| `omniroute serve --no-recovery` | `opts.noRecovery` | the auto-restart supervisor still runs (documented in `docs/guides/USER_GUIDE.md` as the debugging mode) |

The export case matters most. The credentials stay in `config.json` in plain text whenever the keychain is unavailable (headless Linux, keytar missing) and in legacy `profiles` entries. So a user who adds `--no-secrets` before sharing an export, precisely because of that, gets the keys in the file.

The same code base already handles this correctly elsewhere. `resolveProviderCredential()` says so in a comment ("Commander represents the negated `--no-credential` option as `credential === false`"), `validateTrayOptions()` checks `opts.noRecovery || opts.recovery === false`, and `runServe()` reads `opts.open === false` a few lines above the broken check. The fix reads the key Commander sets and keeps accepting the `noX` form, the same way those two helpers do.

These went unnoticed because the chat tests hand-build `{ noHistory: true }`, an options object Commander never produces. `redactContextSecrets()` has its own test, but nothing ran it through the command.

## Related Issues

- None found. Searched for `no-secrets`, `no-history`, `no-recovery`, `noSecrets`, `noHistory`, `noRecovery`, `cli-history`, `redactContextSecrets` and `contexts export`.

## Validation

- [x] Change type: CLI
- [x] Focused tests: `tests/unit/cli-negated-flags.test.ts`, `tests/unit/cli-chat.test.ts`, `tests/unit/cli-contexts.test.ts`, `tests/unit/cli-serve-port.test.ts` (31/31 pass)

## Tests Added Or Updated

- `tests/unit/cli-negated-flags.test.ts` (new). The tests drive the real Commander parser:
  - `chat "…" --no-history` (with `fetch` stubbed) leaves no `cli-history.jsonl`.
  - Guard: `chat` without the flag still records the exchange.
  - `contexts export --no-secrets --out <file>` writes no token or API key and keeps the rest of the context.
  - Guard: `contexts export` without the flag still writes the full config.
  - `serve --no-recovery` reaches the action as `recovery === false`, and the recovery branch in `runServe()` checks that key. The source assertion follows `cli-serve-port.test.ts`, since `runServe()` spawns the server.

With the three source changes reverted, the three flag tests fail, and the two guards and every existing test still pass.

## Coverage Notes

- `bin/cli/commands/chat.mjs` and `contexts.mjs`: the changed lines are executed end to end by the new tests.
- `bin/cli/commands/serve.mjs`: the parse shape is executed; the branch itself is asserted structurally.

## Reviewer Notes

- Nothing changes when a flag is absent.
- Open PRs #12485 and #12484 edit `serve.mjs`, around lines 230–320 and 446–490. This change is the single `if` at line 279, so the hunks do not overlap.

### DIFF
--- a/bin/cli/commands/chat.mjs
+++ b/bin/cli/commands/chat.mjs
@@ -79,7 +79,8 @@ export async function runChatCommand(promptArg, opts, cmd) {
   const data = await response.json();
   const text = extractText(data, opts.responsesApi);
 
-  if (!opts.noHistory) {
+  // Commander stores `--no-history` as `history === false`, never as `noHistory`.
+  if (opts.history !== false && opts.noHistory !== true) {
     appendHistory({ prompt, model: opts.model, latencyMs, usage: data.usage, response: text });
   }
 
--- a/bin/cli/commands/contexts.mjs
+++ b/bin/cli/commands/contexts.mjs
@@ -248,7 +248,9 @@ export function registerContexts(program) {
     .option("--no-secrets", "Omit API keys from export")
     .action(async (opts, cmd) => {
       const cfg = loadContexts();
-      const out = opts.noSecrets ? redactContextSecrets(cfg) : JSON.parse(JSON.stringify(cfg));
+      // Commander stores `--no-secrets` as `secrets === false`, never as `noSecrets`.
+      const redact = opts.secrets === false || opts.noSecrets === true;
+      const out = redact ? redactContextSecrets(cfg) : JSON.parse(JSON.stringify(cfg));
       const json = JSON.stringify(out, null, 2);
       if (opts.out) {
         const { writeFileSync } = await import("node:fs");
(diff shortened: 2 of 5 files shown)


diegosouzapw:
Thanks @datrixlab — this is a genuinely important fix, especially the `contexts export
--no-secrets` case: the flag being a silent no-op meant users who explicitly asked to
redact credentials before sharing an export got them in plain text instead. Your
understanding of Commander's negated-flag shape (`opts.history === false`, never
`opts.noHistory`) is correct, the fix is minimal and defensive (accepts both forms), and
the test suite drives the real Commander parser rather than hand-building an opts
object — all 5 tests pass. High priority for merge given the credential-leak angle.

PR #2595, opened 2026-05-22
fix(#2575): check DB feature flag override in arePrivateProviderUrlsAllowed()

herjarsa:
Fixes #2575.

## Problem

Toggling `OMNIROUTE_ALLOW_PRIVATE_PROVIDER_URLS` in the Feature Flags UI (marked `requiresRestart: false`) had no effect in Electron mode. The actual behavior was only driven by the `process.env` value set at startup, so users had to restart the server for the change to take effect — defeating the purpose of the 'no restart required' flag.

## Root Cause

`arePrivateProviderUrlsAllowed()` in `src/shared/network/outboundUrlGuard.ts` only read from `process.env[PRIVATE_PROVIDER_URLS_ENV]`. It never consulted the DB-backed feature flag store, so UI toggles were silently ignored.

## Changes

- `src/shared/network/outboundUrlGuard.ts`: Added a fallback call to `resolveFeatureFlag(PRIVATE_PROVIDER_URLS_ENV)` from the shared feature flag utils. If the DB override exists and is `true`, it returns `true` immediately. The env-var check remains primary (same priority order as the rest of the feature flag resolution chain).
- Wrapped in try/catch for startup safety — if the DB isn't initialized yet, it degrades gracefully to env-only behavior.

## Testing

- `lsp_diagnostics` clean on changed file
- Runtime logic: env var still takes precedence; DB override applies when env var is unset

Issue #2575, opened 2026-05-22
[BUG] OMNIROUTE_ALLOW_PRIVATE_PROVIDER_URLS feature flag dose nothing when running electron app.


ilya7dev:
### OmniRoute Version

3.8.1

### Installation Method

Electron desktop app

### Operating System

Windows

### OS Version

Windows 11 Pro 25H2

### Provider(s) Involved

LM studio

### Description

OMNIROUTE_ALLOW_PRIVATE_PROVIDER_URLS toggle in feature flag section of electron app has no effect on running server. Restarting the application after setting the toggle dose not set the flag for internal server.

### Steps to Reproduce

0. Make sure .env file dose not have OMNIROUTE_ALLOW_PRIVATE_PROVIDER_URLS=true.
1. Launch electron app, go to Feature flags section, find and enable OMNIROUTE_ALLOW_PRIVATE_PROVIDER_URLS.
2. Add locally running LM Studio as provider (http://localhost:1234/v1).
3. Attemt to fetch available models.
4. Get error message that local providers are prohibited despite setting the flag.

### Expected Behavior

Setting the flag in UI should be equal to adding it to .env file. Fetching model list from local provider should work when flag is set.

### Actual Behavior

Setting the flag in UI dose nothing but recording the change into internal settings DB.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Additional Context

Workaround I found is to edit server.env under C:\Users\\%USERNAME%\AppData\Roaming\omniroute.

diegosouzapw:
When running as a packaged Electron app, OmniRoute reads environment from the packaged config files, not from the shell where you launched it — that's why setting `OMNIROUTE_ALLOW_PRIVATE_PROVIDER_URLS` in your shell has no effect.

To enable it in Electron:

1. Open Settings → Feature Flags → toggle `omniroute_allow_private_provider_urls` (or whatever the UI label is).
2. Restart the Electron app.

Alternatively, edit the persisted feature-flag file directly at `~/.omniroute/feature-flags.json` (or the equivalent on your OS) and set the flag to true.

The shell-only env var path is only honored when running via `npm run dev` or `omniroute serve` from a terminal. Documenting this in the feature-flag panel is on the v3.9 backlog.

diegosouzapw:
[...]

PR #13323, opened 2026-09-11
fix(providers): let a dashboard OFF for private provider URLs beat the env opt-in

datrixlab:
## Summary

- `arePrivateProviderUrlsAllowed()` opens with "DB override takes precedence — it represents an explicit user toggle in the dashboard", but it only acts on an override of `"true"`. An override of `"false"` falls through to the `process.env` check.
- So with `OMNIROUTE_ALLOW_PRIVATE_PROVIDER_URLS=true` in the environment (the opt-in `.env.example` describes), switching "Allow Private Provider URLs" OFF in the dashboard has no effect:
  - The provider validation and outbound guards stay at `"none"`, which means no checks at all.
  - The Feature Flags page shows the flag as off, with source `db`.
- #2595 (for #2575) added the DB read to make the toggle work without a restart, and covered only the OFF→ON direction. The existing tests cover three of the four env/DB combinations and skip this one.
- Fix: read the stored override directly and treat any stored value as final for this flag. The sibling `areLocalProviderUrlsAllowed()` in the same file already does this (`if (dbValue !== undefined && dbValue !== "") return isTrueValue(dbValue);`).

| env | DB override | before | after |
| --- | --- | --- | --- |
| `true` | none | allowed | allowed |
| unset | `true` | allowed | allowed |
| `false` | `true` | allowed | allowed |
| unset | `false` | blocked | blocked |
| `true` | `false` | **allowed** | blocked (`getProviderValidationGuard()` → `block-metadata`) |

## Related Issues

- Related to #2575 / #2595 (the OFF→ON half of the same toggle)

## Validation

- [x] Change type: provider
- [x] Focused tests: every unit test that imports `outboundUrlGuardPolicy` (feature-flag, local-flag, provider validation SSRF guard, provider models LAN guard, remote image fetch, webhook metadata / private opt-in, production build module integrity) plus `cc-compatible-provider.test.ts`: 63/63 pass

## Tests Added Or Updated

- `tests/unit/outbound-url-guard-feature-flag.test.ts`
  - DB override `"false"` with env `"true"` → not allowed, and the validation guard falls back to `block-metadata`.
  - Guard: DB override `"false"` does not change the legacy `OUTBOUND_SSRF_GUARD_ENABLED=false` hatch.
  - Guard: a DB override `"false"` for the sibling `OMNIROUTE_ALLOW_LOCAL_PROVIDER_URLS` still yields `public-only`. The existing local-flag tests only set that flag through `process.env`, so a DB-side regression there went unseen (its `resolveFeatureFlag()` call sits inside a `try/catch` that would swallow a failure).

With the source change reverted, the first test fails and the guards and the four existing tests pass.

## Coverage Notes

- `src/shared/network/outboundUrlGuardPolicy.ts`: all four branches of the new override/env block are exercised by the seven tests in that file.

## Reviewer Notes

- The legacy `OUTBOUND_SSRF_GUARD_ENABLED=false` hatch is intentionally left as it was. It is a different key, and whether a dashboard OFF for this flag should also override a globally disabled guard is a separate decision. Relatedly, the "SSRF Guard" flag (`OUTBOUND_SSRF_GUARD_ENABLED`) is only ever read from `process.env` here, so its own dashboard toggle does nothing. That one fails closed, so I left it out of this PR.
- The module now also imports `getFeatureFlagOverride` from `@/lib/db/featureFlags`. `@/shared/utils/featureFlags` already imported it, so the dependency graph is the same, and #7682's CLI split is unaffected.

### DIFF
--- a/changelog.d/fixes/13323-private-provider-urls-db-off.md
+++ b/changelog.d/fixes/13323-private-provider-urls-db-off.md
@@ -1,0 +1,1 @@
+- **fix(providers):** Switching "Allow Private Provider URLs" off in the dashboard now takes effect when `OMNIROUTE_ALLOW_PRIVATE_PROVIDER_URLS=true` is set in the environment ([#13323](https://github.com/diegosouzapw/OmniRoute/pull/13323))
--- a/src/shared/network/outboundUrlGuardPolicy.ts
+++ b/src/shared/network/outboundUrlGuardPolicy.ts
@@ -1,3 +1,4 @@
+import { getFeatureFlagOverride } from "@/lib/db/featureFlags";
 import { resolveFeatureFlag } from "@/shared/utils/featureFlags";
 import {
   OutboundUrlGuardError,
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Confirmed and reproduced: `arePrivateProviderUrlsAllowed()` only ever acted on a DB override of
`"true"`, so a dashboard OFF toggle silently did nothing when the env opt-in was set — exactly
the SSRF-adjacent gap you described, and important for the Electron build where the env value
is captured at boot. The fix correctly mirrors the sibling `areLocalProviderUrlsAllowed()`'s
existing pattern. All 7 tests (including the new DB-vs-env matrix and the sibling-flag
regression check) pass at the PR head. Ready to merge.

PR #7252, opened 2026-07-15
fix: omniroute dashboard respects PORT env when --port is omitted (#7049)

diegosouzapw:
## Summary
- `omniroute dashboard` (no `--port` flag) previously ignored `process.env.PORT` and always opened `http://localhost:20128`, unlike `serve`/`launch` which already resolve `opts.port ?? process.env.PORT ?? 20128`.
- Root cause: the Commander `--port` option carried a baked-in default value `"20128"`, so `opts.port` was never `undefined` at the action handler, silently defeating any `opts.port ?? process.env.PORT` fallback — plus both call sites (`registerDashboard`'s `--tui` branch and `runDashboardCommand()`) duplicated a `opts.port ? parseInt(...) : 20128` ternary that never consulted the env var.
- Fix: removed the Commander-level default (mirroring `serve.mjs`) and applied the same `opts.port ?? process.env.PORT ?? "20128"` precedence, validated through a `parsePort` helper (rejects non-numeric / out-of-range values, mirroring `serve.mjs`'s pattern) in both call sites.
- Precedence, matching `serve`/`launch`: explicit `--port` flag > `process.env.PORT` > `20128` default.

## Validation
TDD: new test file `tests/unit/cli-dashboard-port.test.ts` (9 tests) written first against the buggy code (source-text assertions on the Commander option shape and `process.env.PORT` usage failed as expected), then the fix was applied and the suite went green.

```
node --import tsx/esm --test tests/unit/cli-dashboard-port.test.ts tests/unit/cli-serve-port.test.ts tests/unit/cli-tui.test.ts
# 36 pass, 0 fail (includes the pre-existing cli-tui.test.ts "--tui flag" regression test, unmodified and still passing)
```

Manual CLI check against the real binary:
```
PORT=20129 node bin/omniroute.mjs dashboard --url    # -> http://localhost:20129 (was :20128 before the fix)
node bin/omniroute.mjs dashboard --port 3000 --url   # -> http://localhost:3000 (explicit flag still wins)
node bin/omniroute.mjs dashboard --url               # -> http://localhost:20128 (default unchanged, no PORT set)
```

Scoped gates run green:
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json tests/unit/cli-dashboard-port.test.ts` — clean (note: `bin/**` is globally excluded from ESLint scope by `eslint.config.mjs`, so `dashboard.mjs` itself isn't linted)
- `npm run typecheck:core` — clean
[...]

Issue #7049, opened 2026-07-13
fix(startup): omniroute dashboard ignores PORT env unless --port is passed (alongside-9router setup)


kaon0388v1:
### Problem / Use Case

I'm trying to run OmniRoute and 9router on the same machine at the same time, but currently OmniRoute assumes the same default port as 9router. This causes a port conflict and prevents both services from running together for comparison, testing, or migration purposes.

### Proposed Solution

Add a flexible port configuration so OmniRoute can be started on a different port without modifying source code. The port should be configurable through environment variables and/or CLI flags, and all related commands should respect the same configured port consistently.

For example:

PORT=20129 omniroute
omniroute serve --port 20129
omniroute dashboard should open the configured dashboard URL instead of a hardcoded default

### Alternatives Considered

- Manually editing .env files before each run
- Changing the source code locally
- Stopping 9router before launching OmniRoute

These workarounds are inconvenient and make side-by-side testing difficult.

### Acceptance Criteria

- OmniRoute can run on a custom port without source modifications.
- serve, dashboard, and related commands use the same configured port.
- Default behavior remains unchanged when no custom port is set.
- Existing functionality continues to work on the default port.
- Basic tests cover port resolution and dashboard URL generation.

### Area

CLI Tools Integration, Dashboard / UI

### Additional Context

I want to run OmniRoute and 9router simultaneously on the same machine for testing and comparison. The current hardcoded default port causes a conflict, so a configurable port would make the project much easier to use in multi-instance setups.

### Expected Test Plan

- Add unit tests for port resolution logic.
- Add tests for dashboard URL generation using the configured port.
- Verify omniroute serve starts on a custom port.
- Verify omniroute dashboard opens the correct port.
- Confirm the default port behavior is unchanged when no override is set.

diegosouzapw:
Thanks for the well-scoped request — queued in the backlog.

Good news: most of what you asked for already works today. `PORT` is respected as an env var and `--port` flags exist:
- `PORT=20129 omniroute serve` and `omniroute serve --port 20129` (`bin/cli/commands/serve.mjs`)
- `omniroute dashboard --port 20129` (`bin/cli/commands/dashboard.mjs`)
- `omniroute launch --port 20129` (`bin/cli/commands/launch.mjs`)

[...]

PR #13327, opened 2026-09-11
fix(cli): let restart honor PORT like serve and dashboard

datrixlab:
## Summary

- `omniroute restart` declares `--port` with a Commander default of `"20128"` and passes its options to `runServe()`. There, `parsePort(opts.port ?? process.env.PORT ?? "20128", 20128)` never reaches `PORT`, because `opts.port` is always set.
- The CLI loads `<DATA_DIR>/.env` into `process.env` (`bin/omniroute.mjs`). So a user with `PORT=3000` there gets port 3000 from `omniroute serve`, and port 20128 after `omniroute restart`.
- #7049 / #7252 fixed exactly this default on `dashboard`, and `serve` itself declares `--port` with no default. This PR removes it from `restart` the same way.

| `PORT` | command | before | after |
| --- | --- | --- | --- |
| 3000 | `omniroute restart` | serves on 20128 | serves on 3000 |
| 3000 | `omniroute restart --port 4000` | 4000 | 4000 |
| unset | `omniroute restart` | 20128 | 20128 |

## Related Issues

- Related to #7049 / #7252 (same Commander default on `dashboard`)

## Validation

- [x] Change type: CLI
- [x] Focused tests: `tests/unit/cli-restart-port.test.ts`, `cli-serve-stop-command.test.ts`, `cli-stop-supervisor-respawn-9455.test.ts`, `cli-dashboard-port.test.ts` (18/18 pass)

## Tests Added Or Updated

- `tests/unit/cli-restart-port.test.ts` (new)
  - `restart` with no `--port` reaches the action with `port` undefined, so `runServe()` falls through to `PORT`. The test uses the real Commander parser.
  - Guard: `restart --port 3000` still passes `"3000"`.

With `restart.mjs` reverted, the first test fails and the guard passes.

## Coverage Notes

- `bin/cli/commands/restart.mjs`: option declaration covered through the parser.

## Reviewer Notes

- `stop` is deliberately left alone. When there is no server PID file, `runStopCommand()` falls back to freeing a port, and that still means `--port` or 20128. I tried making that fallback read `PORT` too, and dropped it. The CLI also loads `.env` from the current directory, so `PORT` there can belong to an unrelated app. With no PID file, nothing proves OmniRoute is the process on that port, and `stop` would kill the other app. So in the no-PID-file case, `restart` still frees 20128 before starting on `PORT`, the same as `stop` followed by `serve`.

### DIFF
--- a/bin/cli/commands/restart.mjs
+++ b/bin/cli/commands/restart.mjs
@@ -6,7 +6,8 @@ export function registerRestart(program) {
   program
     .command("restart")
     .description(t("restart.description"))
-    .option("--port <port>", t("serve.port"), "20128")
+    // No Commander default: runServe() falls back to PORT, then 20128 (#7049).
+    .option("--port <port>", t("serve.port"))
     .action(async (opts) => {
       const exitCode = await runRestartCommand(opts);
       if (exitCode !== 0) process.exit(exitCode);
--- a/changelog.d/fixes/13327-cli-restart-honors-port.md
+++ b/changelog.d/fixes/13327-cli-restart-honors-port.md
@@ -1,0 +1,1 @@
+- **fix(cli):** `omniroute restart` now comes back on the port set by `PORT` (shell or `<DATA_DIR>/.env`) instead of always 20128, like `serve` and `dashboard` ([#13327](https://github.com/diegosouzapw/OmniRoute/pull/13327))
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks @datrixlab — clean, minimal fix, and the before/after table makes the bug (and
the exact same fix already shipped for `dashboard` via #7049) very easy to verify. Test
coverage via the real Commander parser is exactly right. Looks merge-ready to me.

PR #13328, opened 2026-09-11
fix(redis): namespace warmup circuit-breaker keys with REDIS_KEY_PREFIX

datrixlab:
## Summary

- `REDIS_KEY_PREFIX` is documented as the namespace for all OmniRoute Redis keys (`.env.example`, `docs/ops/REDIS_PRODUCTION_CONFIG.md`). The rate limiter, auth cache and quota store honor it. The warmup circuit breaker does not: `redisCircuitBreakerStore.ts` hardcodes `const KEY_PREFIX = "omniroute:warmup:cb:"`, and its client in `circuitBreakerFactory.ts` is created without a `keyPrefix`.
- With a custom prefix on a shared Redis, those keys still land in the default namespace:
  - outside a key-pattern ACL such as `~tenantA:*`;
  - shared between two instances that set different prefixes precisely to stay apart, so one instance's forbidden/backoff state for a connection id is read by the other.
- The #11042 review flagged this as MANDATORY: *"There is a third ioredis client at src/lib/warmupScheduler/circuitBreakerFactory.ts:92 … whose store keys … are hardcoded to `omniroute:warmup:cb:` … Either derive that prefix from REDIS_KEY_PREFIX … with a matching test … or narrow the doc claim."* The PR merged without either.
- Fix: derive the prefix the way `redisQuotaStore.ts` does (`${process.env.REDIS_KEY_PREFIX?.trim() || "omniroute:"}warmup:cb:`). I kept it in the store constant rather than adding an ioredis `keyPrefix` to the client, so the key is built in one place and cannot be prefixed twice. With `REDIS_KEY_PREFIX` unset or blank, the key is byte-identical to before.
- `docs/ops/REDIS_PRODUCTION_CONFIG.md`, `docs/reference/ENVIRONMENT.md` and `.env.example` now list the warmup circuit breaker as the fourth workload. The ops doc's "old keys expire via TTL / LRU" note gains one exception: a forbidden connection's key is `PERSIST`ed, so it will not age out.

## Related Issues

- Follow-up to #11042 (configurable Redis key namespace)

## Validation

- [x] Change type: other (Redis / warmup scheduler)
- [x] Focused tests: the new test, the five existing `tests/unit/lib/warmupScheduler/*` Redis/factory tests, `quota-redis-store.test.ts`, `rate-limiter-redis-optional.test.ts` (22 pass, 1 skipped: the real-Redis integration test that needs `RUN_QUOTA_REDIS_INT=1`)
- [x] `eslint` on the changed code; `node scripts/check/check-env-doc-sync.mjs` passes

## Tests Added Or Updated

- `tests/unit/lib/warmupScheduler/redisCircuitBreakerStorePrefix.test.ts` (new). The test runs a real `recordResult()` → `isInBackoff()` → `get()` round trip against a key-recording fake Redis:
  - With `REDIS_KEY_PREFIX=tenantA:`, the only key written is `tenantA:warmup:cb:conn-1`.
  - Guard: with the variable unset or blank, the key is still `omniroute:warmup:cb:conn-1`.

With the source change reverted, the first test fails and the guard passes. The existing prefix tests for the rate limiter and quota store only check source text; this one checks the key actually written.

## Coverage Notes

- `src/lib/warmupScheduler/redisCircuitBreakerStore.ts`: the prefix is used on every Redis call in the store, and the round trip exercises `hset`, `expire`, `hget` and `hgetall`.

## Reviewer Notes

- The default namespace is unchanged, so deployments without `REDIS_KEY_PREFIX` see no change.
- Deployments that already set a custom prefix will start with empty warmup circuit state in Redis. A connection that is currently marked forbidden gets one more warmup attempt, fails with 403 again, and is re-marked. The old `omniroute:warmup:cb:*` keys for forbidden connections were `PERSIST`ed, so they do not expire by themselves; the ops doc now says so, next to its existing note that changing the prefix orphans old keys.
- `docs/ops/REDIS_PRODUCTION_CONFIG.md` did not pass `prettier --check` on the base branch either (table alignment); the diff there only touches the lines about the workloads.


diegosouzapw:
Good follow-through on the #11042 review's mandatory item that never got
addressed — deriving the warmup circuit-breaker key prefix the same way
`redisQuotaStore.ts` already does, with the default-case byte-for-byte
compatibility preserved. Ran your new prefix test at this PR's head: 2/2 pass,
including the round-trip against a key-recording fake Redis proving both the
prefixed and unprefixed cases. This looks merge-ready.

Issue #13325, opened 2026-09-11
docs(backend): 8 GiB host sqlite growth vs RSS vs cgroup cap (measured inventory)


HouMinXi:
## Current

Single-host OmniRoute on an 8 GiB machine (X500). Image `localhost/omniroute:deploy-v3.8.51-custom` (`b1bad53ff90d`), cgroup `--memory=5g`, `OMNIROUTE_MEMORY_MB=4096`, `OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT=20`. `storage.sqlite` is 3.7 GiB on disk. Process RSS is ~1.6–3.0 GiB idle/load; cgroup peak hit the 5.00 GiB cap (`memory.events.max=37872`, `oom=0`).

The 3.7 GiB file is **not** resident RAM. SQLite mmap window is 256 MiB (`r--s`); RSS of that mapping is ~4.6 MiB. Page cache is reclaimable. The 5 GiB peak is anon + file cache filling the cgroup, not the whole database loaded into RSS.

## Inventory (measured 2026-09-11)

| Class | Item | Live number | Action |
| --- | --- | --- | --- |
| Hardware | Host RAM | 7.4 GiB total, ~0.8 GiB other RSS (Hermes, Gitea, Caddy, gnome, journal) | Do **not** raise Omni cgroup to 6–7 GiB. That leaves 0.4–1.4 GiB for the host and already-used zram. Keep 5 GiB cgroup + 4096 MB heap. |
| Hardware | Cgroup cap | peak 5.00 GiB = `--memory=5g`; `oom_kill=0` | Kernel reclaim, not a crash. Document the ceiling. |
| Traffic | `usage_history` | 442k rows, 143 MiB, 2026-07-05 → now, retention 365d | Expected. Cleanup already in `runAutoCleanup`. |
| Traffic | `quota_snapshots` | 817k rows, 105 MiB, retention 90d | Expected. |
| Traffic | `call_logs` | 100k rows (table cap), 47 MiB | Expected. |
| Traffic | `compression_run_telemetry` | 297k rows, 67 MiB, retention 30d | Expected if cleanup is running. |
| Bug | `conversation_turn_nodes` | **2.63M rows, 900 MiB table + 870 MiB indexes**, 2026-08-19 → now, avg 93 turns/conversation, 2.08M `tool` roles | No retention in current `cleanup.ts`. Already tracked: #12453, PR #12548. |
[...]

PR #13331, opened 2026-09-11
fix(memory): skip FTS rewrite on access-count updates

HouMinXi:
## Current

A live 8 GiB host has 962 `memories` rows and 175k `memory_fts_data` / 8.47M `memory_fts_docsize` rows (~780 MiB on disk). `storage.sqlite` is 3.7 GiB overall. The FTS mapping is not process RSS; it is unbounded disk growth.

## Problem

`023_fix_memory_fts_uuid.sql` recreates `memory_fts_au` as AFTER UPDATE ON memories with no WHEN. `recordMemoryAccess()` updates `access_count` and `last_accessed_at` on every retrieval, so each prompt injection rewrites the FTS5 row and appends tombstones.

`cleanupMemoryEntries()` deletes old `memories` rows but never runs FTS5 `optimize`, so tombstones outlive the 962 live rows.

## Change

Migration 176 recreates `memory_fts_au` so it only reindexes when `content`, `key`, or `memory_id` change. `memory_id` stays in the WHEN: `createMemory` inserts then backfills `memory_id`, and that UPDATE must still hit FTS.

After retention deletes, `cleanupMemoryEntries()` runs `INSERT INTO memory_fts(memory_fts) VALUES('optimize')`.

Optional FTS5 versions now include 176 so sql.js hosts skip it the same way they skip 022/023.

## Verification

Node test runner: 25/25 (`memory-fts-access-update`, constants snapshot, typed-decay). Defect injection: drop the WHEN clause, `recordMemoryAccess` grows 4 → 9 and FAIL; restore, 3/3 PASS.

## Scope

This PR is submitted for review and merge. Do not VACUUM production. Existing #12453 / #12548 cover `conversation_turn_nodes`. Hardware cgroup ceiling is documented on #13325, not this PR.

### DIFF
--- a/changelog.d/fixes/13326-memory-fts-skip-access-updates.md
+++ b/changelog.d/fixes/13326-memory-fts-skip-access-updates.md
@@ -1,0 +1,1 @@
+- **fix(memory):** stop FTS5 rewrite on access-count updates and merge tombstones after retention deletes (#13326).
--- a/changelog.d/fixes/13326-memory-fts-skip-access-updates.md
+++ b/changelog.d/fixes/13326-memory-fts-skip-access-updates.md
@@ -1,1 +1,1 @@
-- **fix(memory):** stop FTS5 rewrite on access-count updates and merge tombstones after retention deletes (#13326).
+- **fix(memory):** stop FTS5 rewrite on access-count updates; rebuild the index on cleanup so leftover tombstones shrink (#13326).
--- a/llm.txt
+++ b/llm.txt
@@ -390,7 +390,6 @@ diagnostics) plus **memory**, **skill**, **agentSkill**, **githubSkill**, **pool
 8. **ProviderIcon component:** Unified icon system using `@lobehub/icons` (130+ SVG) with PNG fallback and generic icon fallback chain. Used on providers, dashboard, and agents pages.
 
 9. **DB architecture:** `localDb.ts` is a re-export layer only — real logic lives in 122 `src/lib/db/` modules with 173 SQL migrations.
-
 10. **Upstream headers:** Custom headers merged in executors after default auth; same header name replaces executor value. Forbidden header names in `src/shared/constants/upstreamHeaders.ts`.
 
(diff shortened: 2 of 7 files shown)


HouMinXi:
Follow-up `a0eced96d`: cleanup now issues FTS5 `rebuild` on every memory-retention pass, not `optimize`. `optimize` only merges segments and leaves access-count tombstones on disk.

Live host already rebuilt: `memory_fts_data` 175335 → 25, `memory_fts_docsize` 8471027 → 962. File 3.9 GiB → 1.4 GiB after VACUUM.

HouMinXi:
Pushed `187117e4d`. This PR added SQL migration 176, so the tree has 173 migration files while README / AGENTS.md / llm.txt still said 172. Those three files now match the live count.

The remaining Docs / Fast Quality / ESLint / unit-shard red is the same set currently on `release/v3.8.51` (provider catalog 356 vs 358 live modules, `autoCombo.test.ts` TS2739, `volcengine-plan-binding-upsert.test.ts` explicit `any`, and the shared unit failures). Unique diff does not touch those files.

diegosouzapw:
Good catch, and there's a subtlety here worth calling out: `release/v3.8.51`'s tip already has
migration 178 fixing the same access-count FTS bloat you're describing, but with only
`content`/`key` in the trigger's `WHEN` clause — not `memory_id`. Your version adds
`NEW.memory_id IS NOT OLD.memory_id`, which matters: `store.ts`'s `createMemory()` explicitly
relies on the AFTER UPDATE trigger firing on the memory_id backfill UPDATE (see its own comment),
and 178 silently drops that. I couldn't get a minimal test to surface it as an active failure
(looks like auto-rowid coincidence masks it in a single-row test DB), but the risk is real and
your fix closes it properly — plus the `cleanup.ts` rebuild-vs-optimize change is genuinely new
on top of 178. Your migration number (177) now collides with two migrations already on the tip
(177 and 178) — could you rebase and renumber to 179 (and update
`RENAMED_MIGRATION_COMPATIBILITY` / `OPTIONAL_FTS5_MIGRATION_VERSIONS` accordingly)? Also
flagging for the maintainer that this PR touches `llm.txt` (trivial whitespace-only hunks) which
needs explicit sign-off per this repo's protected-surface policy.

PR #13332, opened 2026-09-11
fix(guardrails): resolve prefixed combo names in Vision Bridge

AStupidBear:
## Summary

- Resolve `combo/<name>` identifiers before the Vision Bridge combo capability lookup.
- Keep prefixed combo models on the same lookup path as the standard combo resolver.
- Add regression coverage for a vision-capable combo addressed with the `combo/` prefix.

## Root cause

The normal combo resolver accepts both bare combo names and `combo/<name>`. Vision Bridge previously performed only the exact lookup, so a prefixed combo could be classified as `not-combo` and take the wrong vision-routing path.

## Testing

- 47 related Vision Bridge tests passed.
- Targeted ESLint passed.
- `git diff --check` passed.

### DIFF
--- a/src/lib/guardrails/visionBridge.ts
+++ b/src/lib/guardrails/visionBridge.ts
@@ -56,8 +56,14 @@ export async function getComboVisionBridgeDecision(
     const { getComboByName } = await import("@/lib/db/combos");
     const { resolveComboForModel } = await import("@/lib/db/modelComboMappings");
 
-    // 1. Try to find combo by exact name match
+    // 1. Try to find combo by exact name match. The normal Combo resolver also
+    // accepts `combo/<name>` for combos stored under their bare name; keep the
+    // Vision Bridge capability check on the same lookup path.
     let combo = await getComboByName(model);
+
+    if (!combo && model.startsWith("combo/")) {
+      combo = await getComboByName(model.slice("combo/".length));
+    }
 
     // 2. If no exact match, try model-combo mapping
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Clean fix, and the root cause is spot on — `combo/<name>` is a real resolution convention used consistently elsewhere in the chat pipeline (`chatCore.ts`, `modelAliasResolver.ts`), so Vision Bridge doing an exact-name-only lookup was a real gap. Re-ran the full test file locally: 13/13 pass, including your new `combo/` prefix case, with no regressions across the other decision/preCall scenarios. Merging as-is — only a nice-to-have, not a blocker: a small changelog fragment under `changelog.d/fixes/` would match repo convention for future reference.

diegosouzapw:
Thanks @AStupidBear — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

Issue #1072, opened 2026-04-08
[BUG] WebSearch in Claude Code shows "Did 0 searches" - fails with 400: tool_choice: Input tag 'function' found using 'type' does not match any of the expected tags


MRGRD56:
### OmniRoute Version

3.5.2

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Provider(s) Involved

A third-party Claude API provider

### Model(s) Involved

claude-opus-4-6

### Client Tool

Claude Code v2.1.96 on Windows 11

### Description

When using Claude Code v2.1.96 with OmniRoute as an API proxy (with a third-party Claude API provider set up), the built-in `WebSearch` tool fails with a `400 Bad Request` error.

Looks like this:

```
● Web Search("latest technology news April 2026")
  Did 0 searches in 749ms
```

The logs say it's `400 Bad Request`. See `Error Logs / Output`.

### Steps to Reproduce

1. Configure Claude Code to use OmniRoute as the API base URL
2. Start a Claude Code session (v2.1.96, Windows 11)
3. Ask Claude to perform a web search (e.g., "search the web for latest news")
4. See `Did 0 searches` every time. No search results.

### Expected Behavior

WebSearch is supposed to work properly.

### Actual Behavior

Now WebSearch shows 0 results every time. It's been a few days/weeks already.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
API Error: 400 {"error":{"message":"[400]: tool_choice: Input tag 'function' found using 'type' does not match any of the expected tags: 'auto', 'any', 'tool', 'none'","type":"invalid_request_error","code":"bad_request"}}
```

### Screenshots

<img width="407" height="50" alt="Image" src="https://github.com/user-attachments/assets/40bfd4f4-5554-4be3-8942-64698e9d63e5" />

aliaksandrsen:
@diegosouzapw I am still observing this issue. It reproduces on my side as well.

WebSearch is not working, while the separately installed MCP EXA works well.

v3.6.5
os - macos 26.4.1


<img width="1996" height="504" alt="Image" src="https://github.com/user-attachments/assets/aa78a2fd-b3d5-4f5a-a72a-b862e014df1d" />

PR #13333, opened 2026-09-11
fix(translator): keep tool_choice "none" instead of turning it into "auto"

datrixlab:
## Summary

- `convertOpenAIToolChoice()` in `open-sse/translator/request/openai-to-claude.ts` maps both `"none"` and `{ type: "none" }` to Claude `{ type: "auto" }`.
- A request that lists tools but switches tool use off, for example the Vercel AI SDK's `toolChoice: 'none'` or any OpenAI client sending `tool_choice: "none"`, therefore reaches a Claude-format provider with tool use enabled. The model can then call tools the client explicitly disabled, and the client receives `tool_calls` it did not ask for.
- Claude has its own `{ type: "none" }`. The upstream error quoted in #1072 lists the accepted tags: *"expected tags: 'auto', 'any', 'tool', 'none'"*.
- The reverse direction had the same gap. `convertToolChoice()` in `claude-to-openai.ts` has no `none` case and falls through to `"auto"`.
- `openai-to-gemini.ts` says its converter "Mirrors convertOpenAIToolChoice in openai-to-claude.ts", and it already maps `"none"` to Gemini `NONE` (pinned by `gemini-tool-choice.test.ts`). The Clova, Cursor and Copilot paths honor `"none"` too.

| input | before | after |
| --- | --- | --- |
| OpenAI `"none"` → Claude | `{ type: "auto" }` | `{ type: "none" }` |
| OpenAI `{ type: "none" }` → Claude | `{ type: "auto" }` | `{ type: "none" }` |
| Claude `{ type: "none" }` → OpenAI | `"auto"` | `"none"` |
| `"auto"`, `"required"`, named function / tool | unchanged | unchanged |

## Related Issues

- Related to #1072 (the Anthropic error listing the accepted `tool_choice` tags)

## Validation

- [x] Focused tests: `translator-openai-to-claude.test.ts`, `translator-claude-to-openai.test.ts` (36/36); every unit test that mentions `tool_choice: "none"` plus the Claude thinking/tool_choice guard tests (116/116); all 61 translator test files (794/794)
- [x] `prettier --check` on the changed lines; `eslint` reports only the 3 errors these files already have on the base branch

## Tests Added Or Updated

- `tests/unit/translator-openai-to-claude.test.ts`: `"none"` and `{ type: "none" }` map to Claude `{ type: "none" }`; guard: `"auto"` still maps to `{ type: "auto" }`.
- `tests/unit/translator-claude-to-openai.test.ts`: Claude `{ type: "none" }` maps to `"none"`; guard: `{ type: "auto" }` still maps to `"auto"`.

With the source changes reverted, both new tests fail on their `none` assertions.

## Coverage Notes

- Both changed branches are exercised by the new tests.

## Reviewer Notes

- Extended thinking accepts `tool_choice` `auto` or `none`. The thinking guard in `executors/base.ts` only reacts to forced choices (`any` / `tool`), so it is unaffected.
- `services/claudeCodeCompatible.ts` has its own `buildClaudeCodeCompatibleToolChoice()`. It drops `"none"` (returns `null`, so no `tool_choice` is sent) for the CC-compatible third-party path. I left it alone: that path targets upstreams whose accepted values I could not verify.

### DIFF
--- a/changelog.d/fixes/13333-tool-choice-none-claude.md
+++ b/changelog.d/fixes/13333-tool-choice-none-claude.md
@@ -1,0 +1,1 @@
+- **fix(translator):** `tool_choice: "none"` is now sent to Claude-format providers as `{ type: "none" }` (and back to OpenAI as `"none"`) instead of `auto`, so the model can no longer call tools the client switched off ([#13333](https://github.com/diegosouzapw/OmniRoute/pull/13333))
--- a/open-sse/translator/request/claude-to-openai.ts
+++ b/open-sse/translator/request/claude-to-openai.ts
@@ -541,6 +541,8 @@ function convertToolChoice(choice, hasServerWebSearch = false) {
   switch (choice.type) {
     case "auto":
       return "auto";
+    case "none":
+      return "none";
     case TOOL_CHOICE_ANY:
       return "required";
     case "tool":
--- a/open-sse/translator/request/openai-to-claude.ts
+++ b/open-sse/translator/request/openai-to-claude.ts
@@ -728,8 +728,10 @@ function convertOpenAIToolChoice(choice) {
     if (choice.type === "function" && choice.function?.name) {
       return { type: "tool", name: choice.function.name };
     }
-    // Map OpenAI string types to Claude equivalents
-    if (choice.type === "auto" || choice.type === "none") return { type: "auto" };
+    // Map OpenAI string types to Claude equivalents. Claude has its own "none"; mapping it
+    // to "auto" let the model call tools the client had switched off.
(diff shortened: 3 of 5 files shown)


diegosouzapw:
Thanks for the fix — verified the bug is real on the current tip (`convertOpenAIToolChoice`
mapped `"none"` to Claude `{type:"auto"}` in both directions) and re-ran your new tests at
your PR head: 36/36 pass, both `"none"` cases and the `"auto"` regression guards included.
Clean, minimal, well-scoped change. Queuing for merge.

PR #11365, opened 2026-08-24
fix(translator): preserve functionCall id in Gemini to OpenAI request translation

Siva010:
## Summary

Fixes an issue in `geminiToOpenAIRequest` where `part.functionCall.id` was discarded in favor of an unconditionally generated random ID (`call_${Date.now()}_...`).

In multi-turn conversations with tool calls, the subsequent `functionResponse` preserves `tool_call_id: part.functionResponse.id`. Discarding `functionCall.id` on the assistant turn caused a mismatch between the assistant's `tool_calls[i].id` and the user's `tool_call_id`, which caused OpenAI-compatible upstreams to reject the request with HTTP `400 Bad Request`.

## Changes

- In `open-sse/translator/request/gemini-to-openai.ts`: prefer `part.functionCall.id` when present, falling back to random generation only if omitted (matching sibling behavior in `antigravity-to-openai.ts:290`).
- In `tests/unit/translator-gemini-to-openai.test.ts`: added regression tests for ID preservation and multi-turn tool call/response ID parity.

## Validation

- [x] Focused tests: `node --import tsx/esm --test tests/unit/translator-gemini-to-openai.test.ts` (8/8 passing)
- [x] Linter: `npm run lint` (0 errors)
- [x] Base branch: `release/v3.8.50`
- [x] Automated regression tests included

## Tests Added

- `tests/unit/translator-gemini-to-openai.test.ts`

PR #13334, opened 2026-09-11
fix(translator): pair id-less Gemini tool results with their call

datrixlab:
## Summary

- Gemini puts an `id` on `functionCall` / `functionResponse` parts only when the client sets one. OmniRoute's own Gemini-format responses never emit it: `openai-to-antigravity.ts` keeps `accum.id` but writes only `{ name, args }`, and so do the Gemini SSE writers. Every multi-turn tool loop through OmniRoute's Gemini endpoints therefore comes back without ids.
- The three Gemini-shaped converters give such a call a generated id (`call_<ts>_<rand>`), but set the matching response's `tool_call_id` to the function **name**. The two never match:
  - `open-sse/translator/request/gemini-to-openai.ts`
  - `open-sse/translator/request/antigravity-to-openai.ts`
  - `src/app/api/v1beta/models/[...path]/convertGeminiToInternal.ts` (the `/v1beta/models/{m}:generateContent` route)
- For the Gemini translator and `/v1beta`, `translateRequest()` then runs `ensureToolCallIds` → `fixMissingToolResponses` → `stripOrphanedToolResults`. That inserts an empty result for the call and drops the real output as an orphan. The model sees `""` instead of the tool's answer.
- For Antigravity, `fixToolPairs` drops both the call and its result, so the whole exchange disappears from the history.
- #11365 fixed the half where the client does send ids. This PR fixes the id-less half.

Measured with the real modules. The call is `get_weather` with no id; the response has no id and result `"22C sunny"`:

| path | before | after |
| --- | --- | --- |
| Gemini → OpenAI via `translateRequest` | `{ role: "tool", tool_call_id: "call_…", content: "" }` | `content: "\"22C sunny\""` |
| `/v1beta` `convertGeminiToInternal` (raw) | `tool_call_id: "get_weather"` (matches no call) | `tool_call_id: "call_…"` (the call's id) |
| `/v1beta` then `translateRequest` | `content: ""` | `content: "\"22C sunny\""` |

- Fix: a small shared helper, `open-sse/translator/helpers/geminiToolCallIds.ts`. It records every call id per function name and gives an id-less response the oldest still-open call with the same name whose id was generated here. Repeated calls to one function are therefore answered in order, the way Gemini returns them. A response that carries an id keeps it and closes that call. An id the client chose is left for the response that names it. If no call is open for the name, the old name fallback remains.
- Pairing is scoped to one round of calls. Once a non-model content (the user's reply, the tool responses) has come in, the next content that makes calls starts a new round, so a call the client never answered cannot take the response meant for a later call to the same function. The model's own thought or text contents between two call contents do not end the round.
- `/v1beta` also returned only the **first** `functionResponse` of a content, which is how Gemini answers parallel calls. It now emits one tool message per response, as the Antigravity converter already does. Otherwise, with gemini-cli's usual pattern (parallel `read_file(a)`, `read_file(b)`, then a later `read_file(c)`), `b`'s response was lost.

## Related Issues

- Follow-up to #11365 (functionCall id preserved when the client sends it)

## Validation

- [x] Change type: provider (translator / `/v1beta` route)
- [x] Focused tests: the new test file plus `v1beta-gemini-tool-calling-6222`, `translator-gemini-to-openai`, `gemini-to-openai-function-response`, `translator-antigravity-to-openai`, `translator-antigravity-signature-bypass` (51/51)
- [x] Wide run: all translator, antigravity, v1beta and gemini unit test files (204 files, 1549/1550). The one failure, `antigravity-missing-project-chat.test.ts`, fails identically on a clean `release/v3.8.51` on Windows.
- [x] `tsc` on the changed files (temp tsconfig extending the root one): clean; `eslint` clean; `prettier --check` on the changed lines

## Tests Added Or Updated

- `tests/unit/gemini-tool-result-without-id.test.ts` (new)
  - For each of the three converters:
    - an id-less response answers the generated call id and keeps its content;
    - two calls to the same function in one turn are answered in order;
    - parallel calls followed by a later call to the same function (the gemini-cli pattern) all pair correctly;
    - a thought-only model content between two calls of one turn does not split the round;
    - a call the client never answered does not take a later call's response.
  - The tool output survives `translateRequest`'s normalization, both through the Gemini translator and through `/v1beta`.
  - Mixed ids, in both orders: a response with an id keeps it, and an id-less one takes the call whose id was generated, not the one the client named.
- `tests/unit/v1beta-gemini-tool-calling-6222.test.ts`: the "functionResponse part → tool role message" case asserted `tool_call_id === "get_weather"` for this exact id-less input. It never compared that against the call's id, which is the bug. It now asserts that the tool message answers the assistant's call id.

With the three converters reverted, 19 of the 51 tests in these files fail: all 18 new cases plus the updated assertion. The rest pass either way.

## Coverage Notes

- `geminiToolCallIds.ts` is exercised through all three converters, including the with-id, without-id, repeated-name and mixed cases.

## Reviewer Notes

- Generated id formats are unchanged in each converter (`/v1beta` still uses its `newToolCallId()`).
- `/v1beta` still drops non-`functionResponse` parts that share a content with a `functionResponse`, as before. `gemini-to-openai.ts` keeps them via `splitCoLocatedFunctionResponses()`. I left that out to keep this PR to the pairing.

### DIFF
--- a/changelog.d/fixes/13334-gemini-tool-result-without-id.md
+++ b/changelog.d/fixes/13334-gemini-tool-result-without-id.md
@@ -1,0 +1,1 @@
+- **fix(translator):** Gemini tool results sent without an `id` (the usual case, since OmniRoute's own Gemini responses never emit one) now reach the model instead of being replaced by an empty result, on the Gemini, Antigravity and `/v1beta` request paths ([#13334](https://github.com/diegosouzapw/OmniRoute/pull/13334))
--- a/open-sse/translator/helpers/geminiToolCallIds.ts
+++ b/open-sse/translator/helpers/geminiToolCallIds.ts
@@ -1,0 +1,61 @@
+/**
+ * Pairs Gemini `functionResponse` parts with the `functionCall` they answer.
+ *
+ * Gemini carries an `id` on either part only when the client set one, and OmniRoute's own
+ * Gemini-format responses never emit it, so multi-turn tool loops through the Gemini
+ * endpoints usually arrive without ids. A call without an id is given a generated one;
+ * falling back to the function *name* for the matching response then points the tool
+ * message at a call id that does not exist, and the tool-call normalization downstream
+ * (fixMissingToolResponses / stripOrphanedToolResults) replaces the real output with an
+ * empty result. Responses without an id therefore take the oldest still-open call id with
+ * the same function name, in order, since one function can be called several times.
+ *
+ * Pairing is scoped to one round of calls. Once a non-model content (the user's reply, the
+ * tool responses) has come in, the next content that makes calls starts a new round, so a call
+ * left unanswered earlier in the history cannot take the response meant for a later call to
+ * the same function. The model's own thought or text contents between two call contents do
+ * not end the round.
+ */
+export function createGeminiToolCallIdPairing(newId: () => string) {
+  const openCalls = new Map<string, Array<{ id: string; generated: boolean }>>();
+  let roundEnded = false;
+
+  return {
+    /** Call once per Gemini `content`, before its parts. */
+    beginContent(content: { role?: unknown; parts?: unknown }): void {
+      const madeCalls =
+        Array.isArray(content?.parts) &&
+        content.parts.some((part) => Boolean((part as { functionCall?: unknown })?.functionCall));
+      if (madeCalls && roundEnded) openCalls.clear();
+      if (madeCalls) roundEnded = false;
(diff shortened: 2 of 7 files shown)


diegosouzapw:
Ran your test suite at the PR head — 24/24 pass, including the round-scoping edge cases
(parallel calls, unanswered calls, thoughts between calls). The shared
`geminiToolCallIds.ts` helper is a clean way to fix this consistently across all three
conversion paths. Looks merge-ready as-is.

PR #12785, opened 2026-09-05
fix(translator): map Claude stop_sequences and stop to Gemini stopSequences

Siva010:
## Summary

- Fixes an issue in direct Claude-to-Gemini request translation (`claudeToGeminiRequest`) where `stop_sequences` and `stop` were omitted from `result.generationConfig.stopSequences`.
- Ensures that Claude-formatted requests (`/v1/messages` or combo targets routed to Gemini) have their stop tokens properly forwarded, preventing models from continuing generation past intended delimiter boundaries.
- Aligns `claude-to-gemini.ts` with `openai-to-gemini.ts` and `claude-to-openai.ts`.

## Related Issues

- None (direct self-contained translator fix).

## Validation

- [x] Change type: provider (format translator)
- [x] Focused tests: `node --import tsx/esm --test tests/unit/translator-claude-to-gemini.test.ts` (19/19 pass)
- [x] `npm run lint` / ESLint passed with 0 errors on modified files
- [x] Reconciled with the current active release base (`release/v3.8.51`); focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR

## Tests Added Or Updated

- `tests/unit/translator-claude-to-gemini.test.ts`
  - Added unit test asserting array `stop_sequences`, single `stop` strings, and omitted stop tokens.

## Coverage Notes

- Changes in `open-sse/translator/request/claude-to-gemini.ts` are 100% covered by the added test in `tests/unit/translator-claude-to-gemini.test.ts`.
- Coverage for `claudeToGeminiRequest` is preserved and improved.

## Reviewer Notes

- No migrations, database schema changes, or breaking changes.
- Safe, isolated logic within synchronous request translation.

PR #13335, opened 2026-09-11
fix(translator): send a Claude tool_result image to Gemini as inlineData

datrixlab:
## Summary

- `claudeToGeminiRequest()` flattens an array `tool_result` with `c.type === "text" ? c.text : JSON.stringify(c)`. A base64 image returned by a tool therefore reaches Gemini as a JSON string: Claude Code's `Read` on a PNG, or an MCP screenshot tool, are common sources. The string lands inside `functionResponse.response.result`, or inside the context text used for signature-less history.
- The model receives the base64 as text instead of the image, and a screenshot of a few hundred KB costs as many context tokens as its base64 length.
- `claude-to-openai.ts` fixed exactly this flattening in #5100, by lifting images into a following image turn. The direct Claude → Gemini path (`/v1/messages` to a Gemini or Vertex provider) still had it; that line dates from v1.0.0.
- Fix: each base64 image in a `tool_result` becomes an `inlineData` part, inserted right after the message's last tool response and ahead of any text that follows it. Claude Code routinely appends `<system-reminder>` text after tool results. This is where the Claude → OpenAI → Gemini path ends up placing the image once `openai-to-gemini.ts` collapses consecutive user turns: tool responses, then images, then text. An image-only result gets the same `"[tool returned an image; see attached]"` text the OpenAI path uses. Text blocks and any other block type are flattened exactly as before.

Measured through `translateRequest(CLAUDE → GEMINI)`, with a Read `tool_result` holding one PNG:

| | base64 inside text / functionResponse | `inlineData` parts |
| --- | --- | --- |
| before | yes | 0 |
| after | no | 1 |

## Related Issues

- Related to #5100 (the same fix on the Claude → OpenAI path)

## Validation

- [x] Focused tests: the new test file, `translator-claude-to-gemini.test.ts`, `claude-gemini-thought-signature-8979.test.ts`, `claude-to-gemini-consecutive-roles.test.ts` (32/32)
- [x] `prettier --check` on the changed lines; `eslint` reports only the 3 errors the file already has on the base branch

## Tests Added Or Updated

- `tests/unit/claude-to-gemini-tool-result-image.test.ts` (new; kept out of `translator-claude-to-gemini.test.ts` because #12785 also appends to that file)
  - Signed history: `[functionResponse { result: "a.png (1x1)" }, inlineData { image/png }]`.
  - Image-only result followed by `<system-reminder>` text: `[functionResponse { result: "[tool returned an image; see attached]" }, inlineData, text]`.
  - Signature-less history (context text): no base64 in any text part, and the image is sent as `inlineData`.

With the source change reverted, all three fail.

## Coverage Notes

- `open-sse/translator/request/claude-to-gemini.ts`: the new branch (image, text, other block types, image-only fallback text) and the placement before trailing text are exercised by the three tests and the existing tool_result tests.

## Reviewer Notes

- Open PR #12785 edits the same file, at the imports and around line 82 (`stop_sequences`). This change is in the `tool_result` case (lines ~136–240), so the hunks do not overlap, and the tests live in a separate file. I kept the import block exactly as it is, although `prettier --write` would reflow it.
- Only `source.type === "base64"` images are lifted, matching #5100 and the `image` case in this same file. URL images are still flattened as before.

### DIFF
--- a/changelog.d/fixes/13335-claude-tool-result-image-gemini.md
+++ b/changelog.d/fixes/13335-claude-tool-result-image-gemini.md
@@ -1,0 +1,1 @@
+- **fix(translator):** An image returned inside a Claude `tool_result` (Read on a PNG, an MCP screenshot) is now sent to Gemini as an image part instead of base64 text ([#13335](https://github.com/diegosouzapw/OmniRoute/pull/13335))
--- a/open-sse/translator/request/claude-to-gemini.ts
+++ b/open-sse/translator/request/claude-to-gemini.ts
@@ -136,6 +136,10 @@ export function claudeToGeminiRequest(model, body, stream, credentials = null) {
     const omittedToolCallIds = new Set<string>();
     for (const msg of body.messages) {
       const parts = [];
+      // Images returned inside tool_result blocks go right after the last tool response,
+      // ahead of any text that follows it, as on the Claude -> OpenAI -> Gemini path.
+      const toolResultImageParts = [];
+      let afterLastToolResult = -1;
 
       if (Array.isArray(msg.content)) {
         for (const block of msg.content) {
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Confirmed: your new test file passes cleanly at the PR head (isolated run — a parallel probe
on the same devbox caused an unrelated SQLITE_BUSY flake on one pre-existing test in the
combined run, not your code). Clean, focused fix with the right precedent (#5100). Looks
merge-ready as-is.

PR #13336, opened 2026-09-11
fix(compression): terminate idle workers on pool eviction (#13091)

KooshaPari:
## Summary

Fixes #13091

The worker-pool `remove()` method never calls `worker.terminate()` on the missed-terminate path, leaking the underlying Worker thread (and its native handle) even after it's removed from the pool. On repeated compression runs the process accumulates zombie workers, growing RSS until GC/OS pressure.

## Fix

- **Remove the `terminate: boolean` footgun parameter** from `remove()` — it can never be safe to drop a Worker without terminating it
- **Always call `worker.terminate()`** (wrapped in a best-effort `try/catch`) before removing the slot from the pool
- **`close()` / `finish()` / `fail()`** all now terminate deterministically (previously `finish()` — the idle path — skipped termination)

## Regression test

Adds a test asserting the worker fires its `'exit'` event after idle eviction (proving `terminate()` was actually called), via a test-only getter exposing the active Worker instances.

## Reason for the change

| Path | Before | After |
|---|---|---|
| `finish()` (idle eviction) | Worker removed from pool, **never terminated** (leak) | `terminate()` always called |
| `close()` / `fail()` | terminated | terminated (unchanged) |
| Memory after repeated compress | Zombie workers accumulate | Workers deterministically terminated |

Fixes #13091

### DIFF
--- a/open-sse/services/compression/compressionWorkerPool.ts
+++ b/open-sse/services/compression/compressionWorkerPool.ts
@@ -131,7 +131,11 @@ export class CompressionWorkerPool {
   }
   async close(): Promise<void> {
     for (const job of this.queue.splice(0)) job.resolve(unchanged(job.originalBody));
-    await Promise.all([...this.workers].map((slot) => this.remove(slot, true)));
+    await Promise.all([...this.workers].map((slot) => this.remove(slot)));
+  }
+  /** @internal Test-only accessor. Do not use in production code. */
+  __getActiveWorkersForTests(): readonly Worker[] {
+    return [...this.workers].map((slot) => slot.worker);
   }
   private spawn(): PoolWorker {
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks for chasing the worker-thread leak in #13091. The release branch already fixed it: `CompressionWorkerPool.remove()` no longer takes a `terminate` flag and always awaits `worker.terminate()`, and the crash path still re-dispatches the queue with `.finally(() => this.dispatch())`.

This branch would drop that `.finally(...)`, so a crashed slot would stop draining queued jobs. The new test also reads `.size` on the array returned by `__getActiveWorkersForTests()`, so its polling loop never arms. Closing as already covered.

PR #12746, opened 2026-09-04
refactor(combo): split executeTarget into gates, attempt, and loop

HouMinXi:
## Summary

`handleComboChatInner`'s 1327-line `executeTarget` is now three modules: pre-dispatch skip gates, the retry/side-effect loop, and the setTry/safety-timer dispatcher. `combo.ts` goes from 4080 to 2234 lines (`split("\n").length`). Round-robin stays in `combo.ts`.

## Why

`combo.ts` sat on its file-size freeze (4080). The leftover bulk was `executeTarget`, not more types. ROADMAP 3.8.52 lists this split under PREPARE.

## What moved

| Module | Role | LOC |
|---|---|---|
| `open-sse/services/combo/attemptLoopTypes.ts` | `AttemptLoopState` / `AttemptLoopDeps` | 118 |
| `open-sse/services/combo/executeTargetGates.ts` | 10 skip gates, breaker through admission | 330 |
| `open-sse/services/combo/executeTargetClassify.ts` | input-bound / body-specific 400 / homogeneous remainder | 55 |
| `open-sse/services/combo/executeTargetAttempt.ts` | retry loop + pin/LKGP/quality/cooldown | 1205 (frozen; classify already extracted) |
| `open-sse/services/combo/comboAttemptLoop.ts` | setTry, #11804 safety timer, speculative tasks, cooldown redispatch | 576 |

`handleComboChatInner` builds `AttemptLoopState`/`AttemptLoopDeps` and calls `dispatchWithCooldownRetry`. Control flow is copied as-is: pin release, LKGP, quota cutoff, 499/400/quality, loop-safety `finally`, cooldown-wait redispatch.

## Tests

Tip `b9b24fbfd` on `upstream/release/v3.8.51` @ `c3945a724`.

```
DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=8192 \
  --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts \
  --import ./tests/_setup/isolateDataDir.ts --test --test-force-exit \
  tests/unit/combo/execute-target-gates.test.ts \
  tests/unit/combo/execute-target-attempt.test.ts \
  tests/unit/combo/combo-attempt-loop.test.ts \
  tests/unit/combo-loop-safety-timer-leak-11804.test.ts \
  tests/unit/combo-body-specific-400-stop-4279.test.ts \
  tests/unit/combo-priority-quota-exhaustion-cutoff-5923.test.ts \
[...]

PR #13338, opened 2026-09-11
fix(combo): stop retrying an unrecoverable request-shape error across the whole fallback chain

hartmark:
## Problem

A malformed conversation history that one provider's protocol validation rejects with a request-shape 4xx is a property of *that request*, not of the account/model that happened to reject it first. The combo fallback loop didn't know this: it kept retrying the byte-identical malformed payload against every remaining fallback target, since each rejection looked like an ordinary per-target failure.

Observed live: a corrupted function-call/response pairing that one Gemini variant rejected with

```
[400]: Please ensure that function call turn comes immediately after a user turn or after a function response turn.
```

was retried identically against every one of the 11 configured Gemini model variants (41-44 combo decisions across several real traces), burning 13+ minutes and the full `MAX_GLOBAL_ATTEMPTS` (30) budget before the combo happened to reach a lenient-enough non-Gemini fallback or gave up. From the client's side this looks like a hung request that eventually times out or surfaces a generic provider error — long after any reasonable retry budget should have given up.

## Fix

A narrow, evidence-driven circuit breaker in the combo target loop: once `IDENTICAL_MODEL_ERROR_STREAK` (3) consecutive targets fail with the exact same `{kind: "model", status, error}` (a `kind: "model"` outcome already means "the client's request itself is invalid", per `classifyComboOutcome`'s existing contract — not a provider-side auth/rate-limit/outage class), the combo stops trying further targets immediately instead of exhausting the whole fallback chain, and also skips the whole-set retry rather than reproducing the same streak again.

The terminal status/message for this early-exit path reuse the exact same `resolveComboTerminalStatus`/`formatComboOutcomes` helpers (#10501) the normal full-exhaustion path already uses, so the response the client receives is identical to what a full, unshortened run would eventually have produced — just reached without the wasted latency.

Deliberately narrow to avoid false positives:
- Requires `kind === "model"` specifically (a 4xx meaning the request itself is invalid), never triggers on rate-limit/auth/provider/timeout classes.
- Requires the *exact same* status **and** error text repeated back-to-back — a transient provider hiccup mixed in with unrelated failures, or the same status/kind with a genuinely different message, never trips it.
- 3-in-a-row across a diversified fallback chain (different providers/accounts) is strong evidence retrying further cannot succeed; a single stray repeat is not enough signal on its own.

## Rebased onto release/v3.8.51

The branch this landed on (`release/v3.8.50`) is now a stale, frozen cut; `release/v3.8.51` is the currently active one, and `combo.ts`'s main dispatch loop was split into `comboAttemptLoop.ts` (`gates`/`attempt`/`loop` extraction, #12746) in between. Retargeted the PR and re-applied the same fix against the new structure: the breaker now lives in `comboAttemptLoop.ts`'s per-target loop and guards its whole-set retry `continue`, using the identical `hasIdenticalModelErrorStreak` predicate. Re-verified clean (`tsc`, all listed tests) on the new base.

## Known limitation / follow-up

This does not fix *why* the conversation history became malformed in the first place. I traced the actual translator path (`FORMATS.OPENAI → FORMATS.GEMINI`, registered with `signaturelessToolCallMode: "context"`) and replayed a real captured failing request through it directly, but the violation depends on live, session-scoped thought-signature cache state (`resolveGeminiThoughtSignature`) accumulated across real turns, which I couldn't reproduce faithfully in an isolated replay. That deeper root cause needs live debug instrumentation to capture the exact signature-cache state at failure time — tracking as a separate follow-up rather than guessing at a fix here.

## Testing

- New unit test (`tests/unit/combo-identical-error-streak.test.ts`, 8 cases) covering the pure `hasIdenticalModelErrorStreak` predicate: streak-not-yet-reached, exact match, mixed-kind tail, wrong-kind-entirely tail, same status/kind but different message, an unrelated failure before the streak, and a custom streak length.
- Ran `tests/unit/combo/combo-attempt-loop.test.ts`, `tests/unit/combo-max-global-attempts-config.test.ts`, and `tests/unit/combo-attempt-body-isolation-7847.test.ts` against the rebased change on `release/v3.8.51` — all pass (24/24).
- `tsc -p tsconfig.typecheck-core.json` clean on the rebased branch.


diegosouzapw:
Nice, evidence-driven fix — the production trace you measured (41-44 identical combo
decisions, 13+ minutes, full `MAX_GLOBAL_ATTEMPTS` burned on a malformed payload no
retry could ever fix) makes the case well, and the guard is appropriately narrow
(`kind: "model"` only, exact status+message match, 3-in-a-row). Ran
`combo-identical-error-streak.test.ts` at your head against the real
`hasIdenticalModelErrorStreak` export: 8/8 pass. Two things before merge: (1) your
branch is ~157 commits behind the current `release/v3.8.51` tip — please rebase and
re-run the combo suite afterward; (2) needs a changelog fragment under
`changelog.d/fixes/`. No correctness concerns with the fix itself.

hartmark:
Rebased onto current `release/v3.8.51` (was ~165 commits behind) and re-ran `combo-identical-error-streak.test.ts` plus the full set of tests importing `comboAttemptLoop.ts`/`comboPredicates.ts` (157 tests across 21 files) — all green at the rebased head. Added the changelog fragment — `changelog.d/fixes/13338-combo-identical-error-circuit-breaker.md`, verified with `check:changelog-integrity`. Ready to merge from my side.

Issue #13319, opened 2026-09-11
fix: glm-sse-transform-arity test is red on release/v3.8.51 base (af49d4972), failing unrelated PRs


anhtahaylove:
## Summary

`tests/unit/glm-sse-transform-arity.test.ts` fails on `release/v3.8.51` at `af49d4972`, independently of any PR. Every PR targeting this base currently inherits a red Unit Tests job.

## Bisect

| commit | result |
|---|---|
| `af49d4972~1` | **pass 2 / fail 0**, exit 0 |
| `af49d4972` *(current base HEAD)* | **fail**, exit 1 |

`af49d4972` — *"fix(stream): accept the buffer size glm.ts has been passing since #12179 (#12925)"* — is the cause.

## Why the two changes conflict

`#12770` added an arity guard asserting the `createSSETransformStreamWithLogger(` call ends right after `suppressThinkClose`:

```js
assert.equal(/65536/.test(call), false, `dead 16th arg still present:\n${call}`);
assert.match(call, /suppressThinkClose\s*\)\s*$/);
```

`#12925` then legitimised the buffer-size argument, so the call now ends with three more positionals:

```
    suppressThinkClose,
    undefined,
    undefined,
    GLM_STREAM_BUFFER_BYTES
  )
```

The first assertion still holds — the literal `65536` is gone, replaced by the named `GLM_STREAM_BUFFER_BYTES`. Only the second one, which pins the call's **last** argument, is now outdated.

## The product change looks correct

This is a stale test, not a product regression. `#12925`'s own suite is green on the same commit:

```
tests/unit/sse-stream-buffer-bytes.test.ts → tests 5 | pass 5 | fail 0
```

## Suggested fix

Update the second assertion to accept the trailing buffer-size argument while keeping the `65536` dead-literal guard that `#12770` was written to protect. Something equivalent to pinning the end of the call to `GLM_STREAM_BUFFER_BYTES` instead of `suppressThinkClose`.

## Environment

Windows 11, Node v24.19.0, clean `npm ci`. Reproduced in an isolated worktree at each commit, so no cross-test pollution. The same failure appears on CI for unrelated PRs, e.g. run `34569960982` (Unit Tests fast-path, all 4 shards).

diegosouzapw:
[...]

PR #13341, opened 2026-09-11
fix(db): prune health-check repair backups

### LINKED ISSUE #13308: fix(backend): health-check-repair backups are never pruned — the retention call is missing from that path only
## Summary

`db_backups/` grows without bound. On this machine it reached **10.3 GB across 18 snapshots in 8 days**, while the live database is 698 MB.

The retention helper exists and works. It is simply not called from the `health-check-repair` path.

## Two backup paths, one prunes

Inspected in the shipped bundle of **omniroute 3.8.50** (`dist/.build/next/server/chunks/src_0h2lvww._.js`).

**Path A — `backupDbFile()`.** Creates the snapshot, then prunes:

```js
(0, a.getDbInstance)().backup(c).then(() => {
  console.log(`[DB] Backup created: ${c} (${o.size} bytes)`), y()   // y = cleanupDbBackups
})
```

**Path B — the health-check path.** Creates the snapshot and returns:

```js
function Q(e, t) {
  ...
  e.exec(`VACUUM INTO '${s}'`),
  console.log(`[DB] Backup created (${t}): ${i}`)
  return !0
}
function Z(e) { return Q(e, "health-check-repair") }
```

No `y()`, no `pruneBackupDirectory`, no `cleanupDbBackups` anywhere in `Q`'s body. Path B is reached from `runDbHealthCheck({ autoRepair: true, createBackupBeforeRepair: () => Z(db) })`.

## Why it accumulates fast

The snapshot is taken **before** the health check decides whether a repair is needed, so a perfectly healthy database still produces a full copy on every startup. Nothing in the logs shows corruption: no `integrity_check failed`, no `malformed`, no `corrupt`.

Measured on this install — 18 snapshots, 8 days, one per process start:

| gap between snapshots | count |
| --- | --- |
| under 1.5 h | 9 |
| ~6 h (the periodic interval) | 0 |

The periodic 6-hour timer never produced one; every snapshot lined up with a process start. On the busiest day, 9 restarts produced exactly 9 snapshots of ~650 MB each.

Because each copy is the size of the whole database, cost scales with database size: at 698 MB, ten restarts cost ~6.5 GB of disk permanently.

## Relationship to #10421

#10421 fixed the same missing-prune bug on the **`pre-migration`** path and is closed as completed. That fix did not cover `health-check-repair`, which is a separate call site. The breakdown table in #10421 shows both reasons existing side by side (48,966 `pre-migration` vs 28 `health-check-repair`) — the smaller count is presumably why only the migration path was addressed.

## Suggested fix

Call the existing helper after a successful snapshot in `Q`, exactly as path A does, so the operator's `maxFiles` / `retentionDays` settings apply uniformly:

```js
e.exec(`VACUUM INTO '${s}'`)
console.log(`[DB] Backup created (${t}): ${i}`)
cleanupDbBackups()   // honors maxFiles / retentionDays
return true
```

Worth auditing whether any other `createBackupBeforeRepair`-style call site has the same gap.

## Environment

- omniroute 3.8.50, local mode, Windows 11
- `storage.sqlite` 698 MB, `db_backups/` 10.3 GB / 18 files before manual cleanup
- No `OMNIROUTE_*` backup or health-check variables set

voidstackloop:
## Summary

- apply configured backup retention after health-check repair snapshots
- share persisted and environment retention resolution with manual backups
- add a Windows-tested regression that creates, repairs, and prunes real SQLite backups

## Validation

- `node --import tsx/esm --test tests/unit/db-health-backup-retention-13308.test.ts tests/unit/db-backup-extended.test.ts` — 16 passed, 1 existing Windows skip
- `npm run typecheck:core`
- `npm run check:cycles`
- ESLint passed for the changed retention and test files; `core.ts` still reports the three unused-symbol errors already present on the base

⚠️ base-red inherited: #13319

Fixes #13308

### DIFF
--- a/src/lib/db/backup.ts
+++ b/src/lib/db/backup.ts
@@ -15,11 +15,11 @@ import {
 } from "./core";
 import { resetAllDbModuleState } from "./stateReset";
 import {
-  MAX_DB_BACKUPS,
-  DEFAULT_DB_BACKUP_RETENTION_DAYS,
-  parsePositiveInt,
-  parseNonNegativeInt,
+  DB_BACKUP_SETTINGS_NAMESPACE,
+  DB_BACKUP_MAX_FILES_KEY,
+  DB_BACKUP_RETENTION_DAYS_KEY,
   pruneBackupDirectory,
+  resolveDbBackupRetention,
 } from "./backupRetention";
 import { isAutomatedTestProcess } from "@/shared/utils/testProcess";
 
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Thanks for chasing this down with real numbers (570 GB / 60 files) — the
analysis and fix are correct. In the time since this was opened, the same bug
(#13308) was fixed and merged separately (#13404, 2026-09-15), so I'm
recommending we close this one rather than reconcile a rebase. One thing worth
flagging: the merged fix only resolves `maxFiles`/`retentionDays` from env vars,
not from the persisted Storage-page setting the way `backup.ts` does for manual
backups — your `resolveDbBackupRetention()` approach actually covers that gap
more completely. If you'd like to open a follow-up for just that narrower piece,
it'd be a welcome, much smaller PR.

voidstackloop:
Opened the narrower follow-up for the persisted-setting gap: #13773

diegosouzapw:
Closing this one: the core pruning for #13308 landed via #13404 (merged 2026-09-15, `createManagedDbBackup` now prunes), and the remaining gap — reading retention from the persisted Storage-page setting instead of env vars only — is exactly what your newer #13773 covers, which is queued as merge-ready. So nothing from this branch is lost; it just continues in #13773. Thanks for sticking with #13308 across both PRs.

PR #13342, opened 2026-09-11
fix(db): close failed initialization connections

### LINKED ISSUE #13303: fix(backend): openDatabase() leaks the SQLite connection when init throws (EPERM on Windows teardown)
## Summary

`openDatabase()` in `src/lib/db/core.ts` leaks an open SQLite connection whenever
initialisation throws between opening the file and handing the handle to its
owner. On Windows the leaked handle keeps `-wal`/`-shm` alive, so the data
directory can no longer be removed (`EPERM`), but the leak itself is
platform-independent — a failed startup simply never releases the connection.

Line numbers below are against `release/v3.8.51` at `fd27ff08c`.

## Leak 1 — the real connection (`core.ts:1236`)

```ts
const db = openSqliteDatabase(sqliteFile);   // 1236
...                                          // 108 lines: pragmas, migrations,
                                             // integrity check, repair/backup
setDb(db);                                   // 1344  <- first owner
```

Nothing between 1236 and 1344 is wrapped. Any throw in those 108 lines — a failed
pragma, a migration error, a failed integrity repair — leaves `db` open with no
reference held anywhere, so nothing can ever close it.

## Leak 2 — the probe connection (`core.ts:1092`)

```ts
const probe = openSqliteDatabase(sqliteFile, { readonly: true });  // 1092
...
closeProbeIfSafe(probe);   // 1125  (inside try)
closeProbeIfSafe(probe);   // 1156  (inside try)
```

Both close paths are inside the `try`, and `probe` is declared inside it too, so
the `catch` cannot even see the variable. A throw between open and close skips
every close.

## Reproduction

Three tests in `tests/unit/db-core-init.test.ts` hit this on a clean base — they
force a failure during init by patching `Database.prototype.prepare`:

- `legacy call_logs schemas are upgraded before combo target indexes are created`
- `probe failures restore preserved critical state instead of booting with an empty database`
- `probe failures without a safe snapshot abort startup and keep manual recovery explicit`

All three fail **purely in teardown with EPERM**, with no assertion failures. The
leaked directories still contain `storage.sqlite`, `storage.sqlite-shm` and
`storage.sqlite-wal`, confirming a live connection rather than a filesystem race.

The tests only make it reproducible; the defect is in the product. Any user whose
DB init throws midway leaks the handle the same way.

## A naive fix does not work

Recording a failed attempt so the next person skips the detour. Declaring `probe`
outside the `try` with `finally { closeProbeIfSafe(probe) }`, plus a
`try/catch` around 1236-1344 that closes `db` and rethrows, does drive
**EPERM 6 -> 0** — but it breaks behaviour:

```
AssertionError: The input did not match the regular expression
  /Manual recovery required after probe failure/i
Input: 'SqliteError: file is not a database'
```

The failure count stayed at 3, identical to the base, which almost hid the
problem: the *count* was unchanged while the *nature* of the failures had shifted
from EPERM to `SqliteError`. Two reasons it is wrong:

1. Closing the probe eagerly breaks code further down that still uses it.
2. Rethrowing the original error bypasses the `Manual recovery required after
   probe failure` path at line 1238.

A correct fix has to respect the recovery path's ownership of both connections,
which is why this is filed separately rather than folded into the test-teardown
work in #13292.

## Relationship to #13290

#13290 covers test suites that delete their temp `DATA_DIR` without closing the
DB first; those are fixed by the shared `cleanupTempDataDir()` helper in #13292.
This issue is the opposite direction: the product leaks the handle, so no
teardown helper can fix it. `db-core-init.test.ts` is the one file from the
#13290 sweep that the helper deliberately does not touch.

voidstackloop:
## Summary
- close failed read-only probes before recovery snapshots or renames the database
- close the primary SQLite connection when initialization throws before singleton ownership
- upgrade legacy `call_logs` columns before schema indexes reference them

## Validation
- `tests/unit/db-core-init.test.ts`: 14/14 passed on Windows
- `npm run typecheck:core`
- `npm run check:cycles`
- `prettier --check src/lib/db/core.ts`
- ESLint reports only the three pre-existing unused-symbol errors in `core.ts`

⚠️ base-red inherited: #13319

Fixes #13303


diegosouzapw:
Thanks for tracking down the Windows handle-leak in `getDbInstance()` — closing
the probe/primary connections on every failure path (not just the happy path) is
exactly right for #13303. Ran your changed test file at this PR's head:
`tests/unit/db-core-init.test.ts` — 14/14 pass. One ask before merge: could you
add a `changelog.d/fixes/` fragment for #13303 (repo convention, missing from
this diff)? Everything else looks solid.

diegosouzapw:
Thanks @voidstackloop — merging via the release merge-train. Validated in local merge-train /tmp/mt-train3b.log on 192.168.0.113 @ train tip 408e2128791696a18966681953136fc96aeb99b0 (32 PRs boarded): static gates green; full test:unit 40694 tests, 17 failing — every one reproduces on the pure release tip (base-red sweep list), zero new reds. Merged --admin per merge-gates §4/§7.

PR #13344, opened 2026-09-11
fix(db): give conversation_turn_nodes its own 1-day retention

### LINKED ISSUE #12453: fix(backend): conversation_turn_nodes / agentic_conversations have no retention — unbounded storage.sqlite growth

HouMinXi:
## Current

`conversation_turn_nodes` had no retention of its own. After #12548 they shared `retention.callLogs`. On a busy 8 GiB host that setting is 90 days, so identity nodes grew to 2.8M rows / ~1.8 GiB while `call_logs` stayed capped at 100k.

## Change

- New `retention.conversationTurnNodes` (default **1 day**), independent of `callLogs`.
- `cleanupConversationTurnNodes` deletes by `last_seen_at` in 10k batches.
- `cleanupAgenticConversations` then drops roots with no remaining nodes, same window.
- Storage UI field + English label.

Does not change the `callLogs` default (30 in code, whatever the dashboard has on an existing install).

## Verify

- 7/7 `tests/unit/db-cleanup-conversation-nodes-12453.test.ts`
- Bug injection: swap the window back to `callLogs` → independence test FAIL; restore → 7/7 PASS.

Fixes #12453

### DIFF
--- a/changelog.d/fixes/12453-conversation-turn-nodes-retention.md
+++ b/changelog.d/fixes/12453-conversation-turn-nodes-retention.md
@@ -1,0 +1,1 @@
+- **fix(db):** give `conversation_turn_nodes` its own 1-day retention instead of sharing `callLogs`, and sweep orphaned `agentic_conversations` after the nodes expire (#12453).
--- a/src/app/(dashboard)/dashboard/settings/components/SystemStorageTab.tsx
+++ b/src/app/(dashboard)/dashboard/settings/components/SystemStorageTab.tsx
@@ -934,6 +934,7 @@ export default function SystemStorageTab() {
       ["mcpAudit", t("retentionMcpAudit"), 30],
       ["a2aEvents", t("retentionA2aEvents"), 30],
       ["callLogs", t("retentionCallLogs"), 30],
+      ["conversationTurnNodes", t("retentionConversationTurnNodes"), 1],
       ["usageHistory", t("retentionUsageHistory"), 30],
       ["memoryEntries", t("retentionMemoryEntries"), 30],
       ["xpAuditLog", t("retentionXpAuditLog"), 30],
--- a/src/i18n/messages/en.json
+++ b/src/i18n/messages/en.json
@@ -7760,6 +7760,7 @@
     "retentionMcpAudit": "MCP Audit (days)",
     "retentionA2aEvents": "A2A Events (days)",
     "retentionCallLogs": "Call Logs (days)",
+    "retentionConversationTurnNodes": "Conversation turn nodes (days)",
     "retentionUsageHistory": "Usage History (days)",
     "retentionMemoryEntries": "Memory Entries (days)",
     "retentionXpAuditLog": "XP Audit Log (days)",
--- a/src/lib/db/cleanup.ts
+++ b/src/lib/db/cleanup.ts
@@ -437,7 +437,7 @@ export async function cleanupCcrBlocks(): Promise<CleanupResult> {
  * The nodes are identity-only: the transcript view resolves each turn's display
  * content from the call_logs row `last_correlation_id` points at. Once
  * cleanupCallLogs purges that row the node can never render again, so the two
- * tables share the dashboard database setting `retention.callLogs` instead of
+ * tables share the dashboard database setting `retention.conversationTurnNodes` instead of
  * a knob of their own; `CALL_LOG_RETENTION_DAYS` configures the separate
  * compliance cleanup path and does not override this window. Deleting an old
  * node only affects reconnect anchors: a conversation resumed after the window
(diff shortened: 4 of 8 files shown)


diegosouzapw:
Solid, well-isolated fix — giving `conversation_turn_nodes` its own short
retention window instead of inheriting `callLogs`'s 90-day default is the right
call given the 2.8M-row growth you measured. Ran your test file at this PR's
head: 7/7 pass. Two housekeeping items before merge: could you rebase onto the
current tip (a few dozen commits behind), and take a look at why "Unit Tests
fast-path" and "Merge integrity" show red on the PR — the change itself tests
clean in isolation, so this may just be CI staleness, but worth a look.

HouMinXi:
@diegosouzapw I checked the failed jobs rather than assuming CI staleness.

[Verified] In run 35043745847 at public head 603068175c, the changelog integrity step passes. Merge integrity then fails because the generated `omni-settings` skill is stale (`Generated: 1`, exit 2). The same dry-run fails on the current release tip cde49c9372, so that failure also exists on the base.

The four unit-test shards report 15 failures in total: nine missing persisted call-log rows, four paid-model classification/settings assertions, one Jina catalog identifier assertion, and one missing CLI translation key (`serve.ready_timeout`). Those failures are real; I have not yet established the regression origin of each one. The quality job also reports file-size, mutation-test-coverage and secrets gate failures, plus an oversized step summary.

Locally, I have integrated cde49c9372 while preserving your locale additions. The base-relative file-size check now passes. I also found a retention-specific gap: the database-settings API schema did not accept the new field. The candidate now tests authenticated PATCH/PUT, SQLite persistence, reopen/GET, invalid values and older clients that omit the field. The joined retention/route/maintenance suite passes 31/31; deleting the schema field produces eight failures, and making it required breaks both older-client cases. A separate test-state leak has also been fixed and checked by removing/restoring the reset.

This candidate is still under review and has not been pushed. I will keep the public check results separate from those local results; the existing PR head is not being claimed green.

PR #13345, opened 2026-09-11
fix(deepseek-web): support native DSML tool calls

espokaos-ops:
## Summary

DeepSeek Web can emit native DSML tool calls even when OmniRoute requests the canonical `<tool>{JSON}</tool>` format.

Previously, the DeepSeek Web translator only recognized `<tool>` / `<tool_call>` markers. Native DSML was therefore returned as normal text, causing the executor to emit `finish_reason: stop` instead of `tool_calls`. Agent clients such as OpenCode could then hang after the model indicated that it wanted to use a tool.

This patch adds native DSML parsing to the DeepSeek Web translator.

## Changes

* Recognize native DeepSeek DSML `<calls>` / `<invoke>` / `<parameter>` markers.
* Support root-level `<invoke>` calls as well as the `<calls>` envelope.
* Support multiple tool calls and multiple parameters.
* Reuse the existing tool-name resolution and argument helpers.
* Preserve tool-call ordering.
* Preserve normal surrounding content.
* Fail safely on malformed, empty, or incomplete invokes.
* Prevent an unclosed invoke from consuming legitimate trailing text.
* Strip consumed DSML ranges while preserving unrelated response content.
* Update the DeepSeek Web prompt to explicitly request OmniRoute's expected tool-call format.

## Root cause

The previous parser contained a fail-fast check equivalent to:

```ts
if (
  typeof text !== "string" ||
  (!text.includes("<tool>") && !text.includes("<tool_call"))
) {
  return { content: text ?? "", toolCalls: null };
}
```

Native DSML therefore never reached the tool parser.

The executor subsequently observed:

```text
toolCalls = null
finishReason = "stop"
```

instead of emitting OpenAI-compatible `tool_calls`.

## Validation

The implementation was validated against real DeepSeek Web traffic containing native DSML.

* 20/20 dedicated DSML unit tests pass.
* 34/34 adversarial regression checks pass.
* 3/3 executor split checks pass.
* 6/6 byte-equivalence checks pass.
* Existing non-DSML behavior remains equivalent.
* Full live tool loop validated after rebuilding the patched bundle:

  * `echo HOLA`
  * `free -h`
  * `uname -a`
  * tool call -> execution -> tool result -> second model turn -> final response
* A real captured native DSML call was successfully converted to an OpenAI-compatible function call and executed through Bash.

## Scope / limitations

This is intentionally a focused DSML translator rather than a general XML/DSML parser.

Known edge cases include literal DSML markers embedded inside parameter values. The implementation is fail-safe for malformed or incomplete tool calls and does not execute invalid tool invocations.

No unrelated files, generated bundles, backups, logs, or temporary PoC artifacts are included in the commit.

### DIFF
--- a/open-sse/translator/deepseekWebTools.ts
+++ b/open-sse/translator/deepseekWebTools.ts
@@ -14,6 +14,9 @@
 //   <tool><tool ...>{json}</tool></tool>     doubled / nested wrappers
 //   <tool id="1"><name>x</name><arguments>{json}</arguments></tool>   XML children
 //   <tool:write><parameter name="content" content="...">             parameter style
+//   <｜｜DSML｜｜ calls><｜｜DSML｜｜ invoke name="bash">…             DeepSeek native DSML dialect
+//     <｜｜DSML｜｜ parameter name="command">…</…> (full-width ｜ U+FF5C,
+//     closers are sloppy: </｜｜DSML｜｜ calls> or bare <｜｜DSML｜｜ invoke>)
 //
 // A single regex cannot robustly cover all of these (nesting + attributes + XML children),
 // so this parser tokenizes the tool tags and walks them with a stack instead. It reuses the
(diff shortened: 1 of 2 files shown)


bagusandrewijaya:
are you already solve and can use deepseek web for coding copilot

espokaos-ops:
Yes, I tested it end-to-end with real DeepSeek Web traffic and the tool-calling issue is solved.

I can use DeepSeek Web as a coding copilot with this patch. The full tool loop works correctly, including tool calls, execution, tool results, and the following model turn.

I tested it with real commands like `echo HOLA`, `free -h`, and `uname -a`, and the DSML and regression tests are passing as well.

One thing worth mentioning is that DeepSeek can ban the account when using it this way. This actually happened to me during testing, and I had to create a new account. Other than that account-related risk, everything worked correctly for me.

diegosouzapw:
Thorough, well-tested fix for the native DSML dialect — the stack-based tokenizer and the L2
fail-safe (no call without at least one non-empty argument) are the right call given how sloppy
the upstream closers are in the wild. Ran `deepseekWebTools.dsml.test.ts` at your head — all
green, and your live end-to-end confirmation in the comments is great extra evidence. Heads up
for context: #13226 touches the same file (different functions, verified conflict-free) and
#12905 handles a related-but-distinct DSML dialect on the non-web DeepSeek path — no action
needed on your end, just flagging for the maintainer. Only ask: add a `changelog.d/fixes/`
fragment.

PR #13348, opened 2026-09-11
fix(usage): keep combo fallback call logs when pending id collides

HouMinXi:
## What this solves

Combo fallback can drop the successful attempt's call log.

#12650 reuses one dashboard pending id across every target attempt of the same client request so a live poll stays valid. `persistAttemptLogs` then writes that id into `call_logs`. The first attempt inserts; the second hits `UNIQUE constraint failed: call_logs.id`, the error is swallowed, and the successful attempt never lands.

Live ids look like `1788792123480-84a42c` (`Date.now()` plus 6 hex from `trackPendingRequest` in `usageHistory.ts`).

## Fix

Leave pending-id reuse alone. On `call_logs.id` UNIQUE, retry the insert with `generateLogId()`. The first attempt keeps the pending id; later attempts still persist.

Continuation lookup is keyed by `response_id`, not `call_logs.id`, so the retry does not retarget `seedPendingContinuationState`.

## Tests

`tests/unit/chatcore-attempt-logging.test.ts`: two `persistAttemptLogs` calls with the same `pendingRequestId` (502 then 200). Without the retry: 1 row and UNIQUE in the log. With it: 2 rows, first id is the pending id, second is generated.

## Related (not this bug)

- #13112 — dashboard list merge of pending + persisted rows (React duplicate keys)
- #13098 / #13099 — `generateLogId()` pid suffix for two processes sharing `DATA_DIR`. Complementary; does not cover same-process combo reuse of the pending id


diegosouzapw:
Thanks for tracing this and for calling out #13112 and #13098/#13099 as related-but-different
in your own write-up — that cross-referencing was genuinely useful for untangling the cluster.
Since you opened this, #13546 landed (2026-09-14) and fixed the same UNIQUE-constraint symptom
at the source: `persistAttemptLogs()` now keys the log row on the per-attempt `traceId`, not
the shared `pendingRequestId`, so the collision this PR's retry handles no longer occurs.
One more thing worth flagging for anyone revisiting this: the retry-with-a-new-id approach here
would have kept the ORIGINAL pending id attached to whichever attempt inserted first (usually
the failed one, since that's why there was a fallback), while the eventual success landed under
a different, newly-generated id — which the dashboard's live-poll (keyed by that original
pending id, see the `pendingIdByCorrelation` comment in `usageHistory.ts`) would never discover.
That's the same "poll target goes stale" problem pending-id reuse was built to prevent in the
first place, just moved one layer down. Closing as subsumed by #13546.

HouMinXi:
Agreed: #13546 fixes the collision at its producer, which is preferable to retrying the insert under a new ID. I verified `persistAttemptLogs()` uses `traceId` on release commit `cde49c9372` and reran `tests/unit/chatcore-attempt-logging.test.ts`: 7 passed, 0 failed, including two attempts sharing one pending request ID. I am withdrawing this redundant retry implementation. The stronger test-only candidate is retained separately for further validation; I am not claiming it has been submitted or merged.

PR #13216, opened 2026-09-10
fix(ci): clear three base-reds on release/v3.8.51 — open-sse typecheck, docs provider count, agent-skills sync

diegosouzapw:
Three HARD base-reds on `release/v3.8.51`, each reproduced on the **clean tip** before being fixed here, and none of them claimed by another open PR.

⚠️ **Operator note — this diff touches agent-instruction surfaces**: `AGENTS.md`, `llm.txt` and `skills/cli-tunnel/SKILL.md` (project rule: never merge one of these without explicit operator approval). In both cases the content is machine-derived, not authored: a **numeral** in a count sentence, and a **generator output**. No prose, rule or instruction text changes. Verify with `gh pr diff 13216 -- AGENTS.md llm.txt skills/cli-tunnel/SKILL.md`.

---

## 1. `check:open-sse-typecheck` → `Fast Quality Gates`

Not reported by the tracker — the Release-Green push-mode run only executes the static gates, and this one is not among them.

#12731 added the `Mode pack ranking gates (cold/warm/health)` block to `open-sse/services/autoCombo/__tests__/autoCombo.test.ts`. Its seven `ProviderCandidate` literals omit three **required** fields of the interface (`provider`, `model`, `errorRate` — `open-sse/services/autoCombo/scoring.ts:109`):

```
✗ open-sse/services/autoCombo/__tests__/autoCombo.test.ts TS2739 (baseline 0, live 7)
```

Filled with values that preserve each case's intent: `provider`/`model` are **identical across all seven** candidates, so `classifyTier()` (`scoring.ts:230,249`) cannot skew the pairwise comparisons; `errorRate: 0` is inert in the five cases that set `failureRate` explicitly, and in the two `cold pool` candidates (`failureRate: undefined`) it yields `reliability = 1` — exactly what that test's own name states. The suite is not executed by any runner (`vitest.config.ts` excludes it, `#8618`), so typecheck is its only gate.

```
# before: FAIL — 1 new/regressed TypeScript error(s)  (openSseTypecheckErrors=7)
# after:  OK — 0 pre-existing error(s), all within frozen baseline.
```

## 2. `Docs sync + fabricated-docs (strict)` — provider count 356 → 358

Reported on run [34532113514](https://github.com/diegosouzapw/OmniRoute/actions/runs/34532113514). EURouter (#13025) and GreenPT (#13024) took the live provider modules to **358** unique IDs: `check:docs-all` reported **7 STRICT drifts**, with three further count claims masked behind them.

[...]

PR #13349, opened 2026-09-11
fix(ci): clear the orphan base-reds on release/v3.8.51 — the #12867 pipeline extraction, a legacy-DB boot abort and the catalog readers

diegosouzapw:
Clears the **orphan** half of the base-red on `release/v3.8.51` (#12732) — the failures that no other open PR was covering. Eleven test files that are red on the clean tip `af49d497` are green here.

⚠️ base-red inherited: #12732 — `No new ESLint warnings` (owned by #13215), `Fast Quality Gates` → `pack-policy` (owned by #12929) and `open-sse-typecheck` (owned by #13216), plus the redaction family owned by #12945 / #13036 / #13227. None of those are this branch's defect and none are touched here.

## The dominant root cause is #12867, not #12506

Four of the five production fixes trace to the same commit: **#12867 (`d6f31501`)** extracted the provider execution leg out of `chatCore.ts` into `open-sse/handlers/chatCore/providerExecutionPipeline.ts` and re-implemented the non-2xx branch with an inline `JSON.parse` instead of delegating to `parseUpstreamError()`. Four behaviors went missing, silently, for every request routed through the new pipeline:

| lost behavior | user-visible effect |
| --- | --- |
| `parseUpstreamError()` message extraction | a plain-text `429` surfaced as the empty `statusText` — `upstream error` — instead of the upstream wording |
| body-derived `retryAfterMs` | `Please retry after 20s` never drained the runtime reservoir, so the limiter was never locked |
| `recordRateLimitBody` (plumbed through `PipelineStateHooks`, wired at two call sites, never called) | `updateFromResponseBody()` stopped running entirely on this path |
| upstream `error.code` / `error.type` | gates keying on the **pair** — `isAntigravityMissingProjectError` — could never fire, so a config-class `422` degraded into a generic 5s account cooldown |

The same extraction also moved the post-response block behind `if (stream)`, which dropped the `401`/`403` credential refresh and the failure-state persistence for the non-streaming leg: a Copilot token was never retried after a `401`, and per-model quota locks lost their helper references. Both legs now drive one shared refresh.

## Everything else

- **`src/lib/db/core.ts` — a real boot bug.** #12832 declared `CREATE INDEX idx_cl_request_provider ON call_logs(request_type, provider)` inside `SCHEMA_SQL`, which runs **before** `ensureCallLogsColumns()` heals a legacy table. On any install predating `request_type` the whole `db.exec(SCHEMA_SQL)` aborts with `no such column` and the server never opens the database. The index moved next to the other `request_type` indexes, after the healing.
- **`autoCombo/virtualFactory.ts`** iterated `getCustomModels()` raw; one `null` row in the operator-writable blob killed **every** `auto/*` pool with a `TypeError`.
- **`activeSyncedCatalog.ts`** — #12934 unioned `customModels` into `getAllActiveSyncedModels()` alongside the dispatch readers it actually needed. That reader is not a dispatch reader, and `/v1/models` has its own custom-model pass right after, so a custom embedding came out as a parentless `jina/<id>` primary with the registry's `jina-ai/<id>` demoted to its child. The dispatch trio keeps the union; only this reader loses it.
- **`imageGeneration.ts`** — codex image errors crashed reading fields off a body that #12506's sanitization had already rewritten.

## Guards were re-pointed, never relaxed

Three source-scanning guards broke because the code they read moved, not because the invariant broke. Each was rewritten to follow the code and each came out **stricter**:

- `combo-provider-cooldown-sibling` now requires the per-model-quota skip in **every** module that calls `recordProviderCooldown(`, not just in `combo.ts` (the dispatcher moved to `combo/executeTargetAttempt.ts` and `combo/roundRobinCombo.ts` in #12746/#12811).
- `hard-session-lease-bypass-inventory` now counts property-access calls too, which is how #12867's extracted file reaches `getProviderCredentials` — the bare-identifier scan would have let those two sites leave the inventory unnoticed. Three assertions replace the one dead regex for the codex managed-lease fence.
- `chatcore-hierarchical-admission` follows the composite-slot acquisition into the split admission loops.

`kiro-tool-call-validation` was the one genuinely stale test: #12506 made an upstream error frame terminate the stream after forwarding `response.failed`, a contract its sibling `stream-passthrough-error-redaction` already encodes for nine cases. The harness was switched to a reader; all four content assertions are intact and one was added.

## Validation

| file | tip `af49d497` | this branch |
| --- | --- | --- |
| `antigravity-missing-project-chat` | 1 fail | 2/2 |
| `chat-rate-limit-body-lock` | 2 fail | 2/2 |
| `chatcore-hierarchical-admission` | 1 fail | 2/2 |
| `chatcore-translation-paths` | 2 fail | 74/74 |
| `combo-provider-cooldown-sibling` | 1 fail | 9/9 |
| `db-core-init` | 1 fail | 14/14 |
| `hard-session-lease-bypass-inventory` | 2 fail | 3/3 |
| `image-generation-handler` | 3 fail | 47/47 |
| `kiro-tool-call-validation` | 1 fail | 6/6 |
| `models-catalog-route` | 10 fail | 44/44 |
| `provider-execution-pipeline` | — | 14/14 |
| `non-streaming-provider-leg` | — | 39/39 |

`typecheck:core`, `typecheck:noimplicit:core` and `check:cycles` are clean. A 2664-test battery over every `chatcore-*`, `chat-*`, pipeline, `db-*` and catalog file was run on this branch **and on the clean tip** for comparison: eight failures are fixed here, and the deltas that only appear on this branch all pass in isolation — they are the `CATALOG_BUILD_TIMEOUT_MS` cold-build flake this PR documents (an 8s bound against a 10-13s tsx cold build under load), not regressions. Repo-wide `npm run lint` still reports the three inherited `no-explicit-any` errors owned by #13215 and nothing from these files.

Two production fixes ship with new TDD regression tests that fail on the base with the original error.

## Merge-order notes

- `src/lib/db/core.ts` also appears in #13341 and #13342.
- `tests/unit/models-catalog-route.test.ts` also appears in #13315.

Refs #12732


diegosouzapw:
Two **new** base-reds appeared on the tip while this branch was being built, both reproduced on the clean tip `af49d497` before anything here touched them:

1. **`Docs Gates (fast-path)`** — the EURouter (#13025) and GreenPT (#13024) merges took the provider count from 356 to 358 without updating `docs/reference/PROVIDER_REFERENCE.md` and six SVGs. **Already owned by #13347 (@maheswari-pinneti)**, which covers exactly those seven files. Not touched here.

2. **`Merge integrity`** → `check:agent-skills-sync` — orphaned, so it is fixed here in `480f4a2b`. #13009 taught the CLI registry parser to read positionals declared with `.addArgument()` but never re-ran the generator, so `skills/cli-tunnel/SKILL.md` has been stale on the base ever since. Regenerated with the official generator (`--apply --only=cli-tunnel`); the whole delta is `tunnel create` gaining its `[type]` positional.

⚠️ **Operator attention:** that commit touches `skills/**/SKILL.md`, which AGENTS.md classifies as an agent-instruction surface requiring explicit approval before merge. It is machine-generated and two lines, but please eyeball it rather than waving it through with the rest.

Also confirmed on the clean tip, so still not this branch's defect: `No new ESLint warnings` is the three `no-explicit-any` errors owned by #13215, and `Fast Quality Gates` is `pack-policy` (#12929) plus `open-sse-typecheck` (#13216).

diegosouzapw:
Another four base-reds surfaced on the tip from today's merges, none of them owned by an open PR, so they are fixed here in `555d97f1`. All four were reproduced on the clean tip first.

| guard | what changed under it | fix |
| --- | --- | --- |
| `providers-constants-split` | GreenPT (#13024) + EURouter (#13025) | 238 → 240 |
| `provider-node-reserved-prefix` | same two providers | 408 → 410 |
| `provider-translate-path-golden` | same two providers | regenerated with `UPDATE_GOLDEN=1`; the snapshot delta is purely additive, no existing provider moved |
| `glm-sse-transform-arity` | #12925 declared the buffer-size slot GLM had been filling since #12179 | guard re-pointed |
| `injection-guard-scan-bound-3932` | #13104 split the 16 KB budget between head and tail | cases moved to the middle |

The last two are worth a second look because they invert what the old assertion said, so I kept each invariant explicitly rather than relaxing it:

- **GLM** — the property was never "no 16th argument", it was "the buffer size must land in a real parameter instead of being dropped as an extra positional". The guard now checks that from the helper side (the slot is declared, and declared last) and requires the call site to pass the named constant, so a magic literal cannot creep back and hide the next mismatch.
- **Injection scan** — #13104 reshaped the window from head-only to head-plus-tail, so "outside the window" is now the middle of an oversized body, not its end. Both cases moved the marker to the middle. The 16 KB ceiling these cases exist to protect is untouched, and the export assertion on `MAX_INJECTION_SCAN_BYTES` is unchanged.

Still **not** fixed here and still red: `check-docs-counts-sync` and `Docs Gates`, both the same seven 356 → 358 drifts owned by #13347 (@maheswari-pinneti).

diegosouzapw:
Status of the quality gates on this branch, after `b1d58c51`.

**Cleared here:**
- `mutation-test-coverage` — three covering tests had drifted out of `tap.testFiles` on the base tip (`combo-predicates-epoch-cooldown`, `opencode-400-model-unavailable`, `chat-correlation-id-exhaustion`). Reproduced on the clean tip first, so it is inherited drift, not this branch's. Registered.
- `file-size` — `models-catalog-route.test.ts` was already sitting exactly on its frozen ceiling, so the one-line `CATALOG_BUILD_TIMEOUT_MS` pin could not be absorbed. The explanatory comment was collapsed into a single trailing note and `testFrozen` bumped 1652 → 1653, with the full rationale recorded in `file-size-baseline.json` alongside the existing precedents.

**Needs your call — `complexity-ratchets`, and it is genuinely this branch's:**

`open-sse/handlers/chatCore.ts` goes 10 → 12 cyclomatic violations and 3 → 4 cognitive. The cause is structural, not substantive. Restoring the failure-state persistence for the non-streaming leg required naming the block so both legs could call it, and the ratchet counts violations **per function** — so a block that cost nothing while inline costs three once it has a name. The same change made the file measurably simpler:

| | tip | this branch |
| --- | --- | --- |
| `handleChatCore` cyclomatic | 1032 | 971 |
| `handleChatCore` cognitive | 1245 | 856 |

I tried three ways out and rejected all three, so you are not re-treading them:

1. **Move it to its own module** — I built it, and it does clear the gate (57 vs 57). But the block closes over roughly twenty locals of `handleChatCore` (`currentModel`, `lockModel`, `hasPerModelQuota`, `coreOwnedAntigravityLockout`, `markAccountSemaphoreBlocked`, …), and the extraction came back with 19 TypeScript errors that `typecheck:core` does not catch because it does not cover the `open-sse/` workspace. Reverted rather than shipped.
2. **Suppress it inline** — `eslint.complexity-ratchets.config.mjs` sets `noInlineConfig: true`, so the gate ignores disable comments by design. Correctly so.
3. **Split the ladder into small handlers** — it is nine branches over `PROVIDER_ERROR_TYPES`, one of them 135 lines; it would take six or seven functions to get every piece under the 80-line and complexity-15 limits. That is a real refactor of the error path, with no new test coverage behind it, and not something I will slip into a base-red fix.

So the options are yours: accept the +2/+1 and re-ratchet at release, or land this and track the decomposition under #3501. Reverting the fix instead would put `chatCore refreshes GitHub credentials after 401` and `chatCore locks per-model quota failures` back to red with no owner.

PR #13350, opened 2026-09-11
fix(security,resilience): block origin-IP header forwarding and treat 413 as retryable TPM

themedexperiencesusa:
⚠️ base-red inherited: #12732

Two independent, self-contained fixes to the gateway path. Both were cut fresh from
`upstream/release/v3.8.51` (`20abd89d7`) rather than rebased from an older branch, so there is
no stale drift in the diff.

## 1. `fix(security): never forward origin-IP headers upstream`

Operator-set custom upstream headers could carry the client-origin IP
(`x-forwarded-for`, `x-real-ip`, `cf-connecting-ip`, `forwarded`, `via`, …) to the upstream
provider, disclosing or spoofing it.

The `FORBIDDEN` denylist in `src/shared/constants/upstreamHeaders.ts` covered the
Host/hop-by-hop/framing set but not the forwarding/IP set. The scrubbers on the Antigravity
(`antigravityHeaderScrub.ts`) and Cursor CLI (`cursorCliProxy.ts`) paths already strip these —
this extends the canonical denylist so the protection applies to **every** provider rather than
those two.

11 header names added. No behavior change for any header already allowed.

## 2. `fix(resilience): treat 413 payload-too-large as retryable TPM rate limit`

Providers with a tokens-per-minute cap (Groq among them) answer an oversized turn with `413`
rather than `429`. `checkFallbackError()` did not list `413` as retryable, so the request failed
hard instead of falling back to another account or model.

- `PAYLOAD_TOO_LARGE` (413) added to `HTTP_STATUS` and to the retryable set
- `413` returns a `MODEL_CAPACITY` retryable fallback (not `SERVER_ERROR`, which would apply a
  server-error cooldown to what is really a capacity signal)
- `"tokens per minute"` / `"tpm"` recognised as context-overflow patterns

`CONTEXT_OVERFLOW_PATTERNS` only selects a different model with `cooldownMs: 0`, so a broad
match costs one retry, never correctness.

## Tests

`tests/unit/upstream-headers-sanitize.test.ts` — **6/6 pass**, including two new cases added by
this PR asserting the origin-IP set is dropped by both `sanitizeUpstreamHeadersMap` and
`isForbiddenCustomHeaderName`.

Every `accountFallback` suite passes against this base, including the 96-test
`account-fallback-service.test.ts` core:

| Suite | Result |
| --- | --- |
| `upstream-headers-sanitize` | 6/6 |
| `account-fallback-service` | 96/96 |
| `account-fallback-anthropic-quota` | 13/13 |
| `agy-pro-fallback-chain-3786` | 12/12 |
| `emergency-fallback` | 10/10 |
| `8247-accountfallback-model-unhealthy` | 2/2 |
| `8248-accountfallback-nvidia-degraded` | 3/3 |
| `accountfallback-ratelimit-400-4976` | 4/4 |
| `account-fallback-cf1010-no-retry-8775` | 6/6 |
| `account-fallback-retry-after-json` | 4/4 |
| `account-fallback-lockout-eviction` | 4/4 |
| `account-fallback-route-restriction-403` | 3/3 |

`npm run typecheck:core` — clean (exit 0).

### Pre-existing failures, not caused by this PR

Three fallback-named tests fail. I re-ran all three on the **untouched base**
(`20abd89d7`, detached, with these commits not applied) and they fail there identically:

- `tests/unit/ui/ClaudeToolCard-manual-config-fallback.test.tsx` and
  `tests/unit/ui/OpenClawToolCard-manual-config-fallback.test.tsx` — `Vitest mocker was not
  initialized in this environment. vi.queueMock() is forbidden.` These are Vitest files being
  run under the Node test runner; they pass under `npm run test:vitest`.
- `tests/unit/authz/client-api-policy-fallback.test.ts` — 9/10, one `EPERM` on tempdir cleanup
  (Windows-specific).

None touch the two files this PR changes.

### Not run

`npm run lint` could not execute in this environment: `@eslint/compat` is declared in
`devDependencies` (`^2.1.0`) but absent from the installed tree, and the installed ESLint is
9.39.5 against a declared `^10.9.0`. This fails identically on the untouched base, so it is a
stale local `node_modules`, not something this PR introduces. CI's `lint` job will be
authoritative.


diegosouzapw:
Clean, well-scoped pair of fixes cut fresh from the current tip. The origin-IP header block
extends protection already applied to Antigravity/Cursor CLI to every provider via custom
upstream headers — verified the denylist is the single source consumed everywhere (no
duplicated list to drift). The 413→retryable-TPM fix is correctly mapped to `MODEL_CAPACITY`
rather than a generic server-error cooldown. Ready to merge.

diegosouzapw:
Thanks @themedexperiencesusa — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

Issue #6559, opened 2026-07-07
fix(startup): --mcp flag — Cannot find package 'ioredis/built/index.js' in packaged dist


TheRealPasha:
### OmniRoute Version

3.8.45

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 10 22h2

### Node.js Version

24.11.0

### Description

Getting module not found error when running omniroute --mcp. The same issue with different cmd output was resolved as i was trying to find a fix but it seems to be back?
```
C:\Windows\system32>omniroute --mcp
  📋 Loaded env from C:\Users\pavel\.omniroute\.env
  📋 Loaded env from C:\Users\pavel\AppData\Roaming\npm\node_modules\omniroute\.env
node:internal/modules/esm/resolve:204
  const resolvedOption = FSLegacyMainResolve(pkgPath, packageConfig.main, baseStringified);
                         ^

Error: Cannot find package 'C:\Users\pavel\AppData\Roaming\npm\node_modules\omniroute\dist\node_modules\ioredis\built\index.js' imported from C:\Users\pavel\AppData\Roaming\npm\node_modules\omniroute\dist\open-sse\mcp-server\server.js
    at legacyMainResolve (node:internal/modules/esm/resolve:204:26)
    at packageResolve (node:internal/modules/esm/resolve:778:12)
    at moduleResolve (node:internal/modules/esm/resolve:858:18)
    at defaultResolve (node:internal/modules/esm/resolve:990:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:757:20)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:734:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:317:38)
    at #link (node:internal/modules/esm/module_job:208:49) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v24.11.0
✖ Failed to start MCP server: MCP server exited with code 1
```

### Steps to Reproduce

1. Execute `omniroute --mcp`

### Expected Behavior

mcp to successfully run

### Actual Behavior

mcp didnt run

### Test Impact

Unsure

### Error Logs / Output

```shell

```

diegosouzapw:
Thanks for reporting! Fixed in `release/v3.8.47` via #6725 (already merged into the active development branch — feel free to pull and test it now). It will ship in the next release (v3.8.47).

PR #13352, opened 2026-09-11
fix(build): copy ioredis and bcryptjs into the standalone bundle

sistemabritto:
## Problem

Two packages are silently dropped from the standalone build output because they are only
reached through code paths the standalone tracer (Next's NFT-based tracing, per the comment
block in `Dockerfile` / `scripts/build/assembleStandalone.mjs`) never follows:

- **`ioredis`** is a deliberately *lazy* dependency — Redis is optional, and it's reached only
  via a runtime `await import("ioredis")` in `src/shared/utils/rateLimiter.ts` (see the
  existing `#6559` comment there), `src/lib/warmupScheduler/circuitBreakerFactory.ts` and
  `src/lib/quota/redisQuotaStore.ts`, never through a static top-level import. Any self-hosted
  deployment that sets `REDIS_URL` crashes the first time it reaches one of those call sites:

  ```
  Error: Cannot find module 'ioredis'
  ```

  Reproduced on a real production Docker deployment (v3.8.49, `REDIS_URL` configured) — the
  standalone image shipped `node_modules/ioredis/package.json` but none of its actual
  dependencies or `built/` output, because nothing had copied the rest of the package.

- **`bcryptjs`** *is* statically imported by `src/lib/auth/managementPassword.ts`, so the main
  server bundle is fine (Next inlines the small pure-JS package directly into the compiled
  route chunk instead of leaving it as an external `node_modules` dependency — dashboard login
  keeps working). `bin/cli/settings-store.mjs` — the `omniroute reset-password` /
  `bin/reset-password.mjs` CLI used to recover a lost dashboard password — is a separate,
  unbundled entrypoint that does a plain runtime `import bcrypt from "bcryptjs"` and needs the
  real package physically present in `node_modules/`, which nothing else requires as a loose
  runtime dependency, so it's never copied. Reproduced on the same deployment:

  ```
  node bin/reset-password.mjs --password-stdin
  # Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'bcryptjs' imported from
  # /app/bin/cli/settings-store.mjs
  ```

  ...even though the same container's dashboard login (which also depends on `bcryptjs`) was
  working normally at the time.

## Fix

Add both to `EXTRA_MODULE_ENTRIES` in `scripts/build/assembleStandalone.mjs` — the single
source of truth cited in `Dockerfile` for exactly this class of dynamically-reached /
separate-entrypoint dependency (same pattern already used for `pino-pretty`, `split2`,
`@swc/helpers`, the MITM `_internal` shims, etc).

## Tests

Extended `tests/unit/build/assemble-standalone.test.ts`:
- `seedSidecarSources` now seeds `node_modules/ioredis/{package.json,built/index.js}` and
  `node_modules/bcryptjs/{package.json,index.js}`.
- The sync/async parity test asserts both are present in the copied tree — regression guard
  for the two "Cannot find module/package" crashes reproduced above.

```
node --import tsx/esm --test tests/unit/build/assemble-standalone.test.ts
# tests 6, pass 6, fail 0
```

`npm run lint` and `npx prettier --check` clean on both changed files.

## Checklist

- [x] Tests added/updated in the same PR (production code change in `scripts/build/`)
- [x] `npm run lint` clean
- [x] Conventional commit message
- [x] No `Co-Authored-By` trailers
- [x] Branched from `release/v3.8.51` (active release)

---

https://claude.ai/code/session_01QgcSrNE33CD9GAvrUmH4s1

### DIFF
--- a/scripts/build/assembleStandalone.mjs
+++ b/scripts/build/assembleStandalone.mjs
@@ -163,6 +163,43 @@ const EXTRA_MODULE_ENTRIES = [
     dest: ["node_modules", "pino-pretty"],
   },
   { label: "split2", src: ["node_modules", "split2"], dest: ["node_modules", "split2"] },
+  {
+    // ioredis is a deliberately LAZY dependency (Redis is optional — see the
+    // #6559 comment in src/shared/utils/rateLimiter.ts) — reached only via a
+    // runtime `await import("ioredis")` in rateLimiter.ts,
+    // warmupScheduler/circuitBreakerFactory.ts and quota/redisQuotaStore.ts,
+    // never through a static top-level import. The standalone tracer only
+    // follows statically-analyzable imports, so it never sees these call
+    // sites and drops ioredis from node_modules/ entirely. Any self-hosted
+    // deployment that actually sets REDIS_URL crashes the first time it
+    // reaches one of those call sites with "Cannot find module 'ioredis'" —
+    // reproduced on a production Docker deployment (REDIS_URL configured,
+    // v3.8.49) where the standalone image shipped ioredis/package.json but
+    // none of its own dependencies or built/ output.
+    label: "ioredis (dynamic import — #6559)",
+    src: ["node_modules", "ioredis"],
+    dest: ["node_modules", "ioredis"],
+  },
+  {
+    // bcryptjs IS statically imported by src/lib/auth/managementPassword.ts,
+    // so the main server bundle is fine — Next's server compiler inlines the
+    // small pure-JS package directly into the compiled route chunk instead of
+    // leaving it as an external node_modules dependency. bin/cli/settings-
+    // store.mjs (the `omniroute reset-password` / bin/reset-password.mjs
+    // CLI, used to recover a lost dashboard password) is a separate,
+    // unbundled entrypoint that does a plain runtime `import bcrypt from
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Verified both root causes independently — `ioredis` really is only reached via dynamic
`import("ioredis")` in rateLimiter.ts/redisQuotaStore.ts (so the static tracer misses it), and
`bcryptjs` really is only reachable statically through the separate, unbundled
`bin/cli/settings-store.mjs` CLI entrypoint. Clean, minimal, well-tested fix — ran the full
assemble-standalone test suite at your PR head, all 6 pass. Nothing else needed, thanks for the
clear repro details in the comments.

diegosouzapw:
Thanks @sistemabritto — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #13356, opened 2026-09-11
fix(i18n): backfill payloadRule keys into all 42 non-EN locales

patrykkopycinski:
PR #6 added four settings.payloadRule* keys to en.json only, which broke the
key-parity gates for every other locale (i18n-vi-completeness, i18n-pt-br,
settings-i18n-keys).

Backfilled with the EN string as the value, matching the convention in
scripts/i18n/fill-missing-from-en.mjs (no __MISSING__ markers — the parity gate
rejects marker values).

Deliberately scoped to my 4 keys. The same locales are also missing 7
usage.kiloPass* keys from #12178 and carry 3 stale radarPage.* entries whose
en.json counterparts were removed; both are pre-existing and belong to their
owners, not this PR.

Verification:
- all 42 locales: exactly +4 keys, 0 keys lost
- diff contains 0 kiloPass keys (scope held)
- i18n-pt-br: green (was red)


diegosouzapw:
Thanks for looking into i18n key parity, but I need to flag a problem before this can move
forward: the four `payloadRuleTransform*`/`payloadRuleInertPattern*` keys this PR adds to all 41
locale files don't actually exist in `en.json`, and there's no code anywhere in the repo that
references them — I couldn't find the "PR #6" this description refers to, and the English source
catalog has never had these keys. On top of that, the PR creates a brand-new
`src/i18n/messages/in.json` file, but `"in"` isn't a supported locale code in
`config/i18n.json` — it's only listed as an alias of another locale, so that file would never
actually be loaded by the app. I'm closing this rather than merging it — if there's a real
key-parity gap you were trying to solve, happy to take a look at a follow-up PR that backfills
keys that genuinely exist in `en.json`, using the `__MISSING__:` marker convention the other
keys in those same files already follow.

patrykkopycinski:
acknowledgment: all three claims verified independently (en.json key list, zero production refs, i18n.json alias-only), closing as suggested; no follow-up PR unless a genuine en.json parity gap surfaces

PR #13358, opened 2026-09-11
test(adobe-firefly): lock in the browser-spawn guard with a regression test

patrykkopycinski:
The guard added in 48fbafb43 had no test: deleting the ADOBE_FIREFLY_BROWSER_REFRESH
line from tests/_setup/isolateDataDir.ts left the suite green while the Adobe Firefly
session warm resumed spawning the system Chrome at firefly.adobe.com.

Asserts both the runtime effect (adobeFireflyBrowserEnabled() === false) and that the
guard keeps using ||= so a browser-path integration test can still opt back in.

Mutation-checked: removing the guard line fails both tests; restoring it passes both.

### DIFF
--- a/tests/_setup/isolateDataDir.ts
+++ b/tests/_setup/isolateDataDir.ts
@@ -47,6 +47,14 @@ if (!process.env.DATA_DIR) {
 // installCert/uninstallCert/installTproxyCa/uninstallTproxyCa no-op under this.
 process.env.OMNIROUTE_SKIP_SYSTEM_TRUST = "1";
 
+// Browser-spawn guard: the Adobe Firefly session warm (adobeFireflySession.ts)
+// spawns the SYSTEM Chrome with --remote-debugging-port whenever a test reaches it
+// without a valid user JWT — which any mocked-fetch test does by construction.
+// Per-call-site allowBrowserRefresh/tryBrowser flags are not enough: the warm is also
+// reachable indirectly via client/handler paths, so the guard must be global.
+// ||= (not =) so a browser-path integration test can still opt back in.
+process.env.ADOBE_FIREFLY_BROWSER_REFRESH ||= "0";
+
 // DNS-write guard: the suite must NEVER mutate /etc/hosts. Tests that exercise
 // the real MITM path call addDNSEntries(); this env var makes it a no-op.
 process.env.OMNIROUTE_SKIP_DNS_WRITE = "1";
--- a/tests/unit/adobe-firefly-test-browser-guard.test.ts
+++ b/tests/unit/adobe-firefly-test-browser-guard.test.ts
@@ -1,0 +1,44 @@
+/**
+ * Regression guard: the unit suite must never launch a real browser.
+ *
+ * The Adobe Firefly session warm (adobeFireflySession.ts::shouldWarm) spawns the SYSTEM
+ * browser with --remote-debugging-port whenever a test reaches it without a valid user
+ * JWT — which any mocked-fetch test does by construction. Per-call-site
+ * `allowBrowserRefresh: false` / `tryBrowser: false` is NOT enough: the warm is also
+ * reachable indirectly via client/handler paths, so the guard must be global.
+ */
+import { test } from "node:test";
+import assert from "node:assert";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+import { adobeFireflyBrowserEnabled } from "../../open-sse/services/adobeFireflySession.ts";
+
+test("test setup disables Adobe Firefly browser warm", () => {
+  assert.equal(
+    process.env.ADOBE_FIREFLY_BROWSER_REFRESH,
+    "0",
+    "tests/_setup/isolateDataDir.ts must set ADOBE_FIREFLY_BROWSER_REFRESH=0 so the suite " +
+      "never launches the system Chrome at firefly.adobe.com"
+  );
+  assert.equal(
+    adobeFireflyBrowserEnabled(),
+    false,
+    "adobeFireflyBrowserEnabled() must be false under the test setup"
+  );
+});
+
+test("isolateDataDir sets the browser guard with ||= so integration tests can opt in", () => {
+  const setupPath = join(
+    dirname(fileURLToPath(import.meta.url)),
+    "..",
+    "_setup",
+    "isolateDataDir.ts"
+  );
+  const source = readFileSync(setupPath, "utf8");
+  assert.match(
+    source,
+    /process\.env\.ADOBE_FIREFLY_BROWSER_REFRESH \|\|= "0";/,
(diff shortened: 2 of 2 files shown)


abhisheksharma2411 (commented):
The reasoning is right — a warm that spawns the system Chrome whenever a mocked-fetch test reaches it is exactly the thing a global guard should hold, and per-call-site flags can't cover the indirect paths. But **both tests fail on this PR's own head**, because the guard they pin isn't on the branch.

```
✖ test setup disables Adobe Firefly browser warm
✖ isolateDataDir sets the browser guard with ||= so integration tests can opt in
   0 pass, 2 fail
```

## Why

`tests/_setup/isolateDataDir.ts` never sets the variable. Everything it does set, on your head:

```
process.env.APP_LOG_TO_FILE ||= "false";
process.env.DATA_DIR = dir;
process.env.OMNIROUTE_SKIP_SYSTEM_TRUST = "1";
process.env.OMNIROUTE_SKIP_DNS_WRITE = "1";
```

No `ADOBE_FIREFLY_BROWSER_REFRESH`. So test 1 reads `undefined` and test 2's regex matches nothing.

The commit the description credits, `48fbafb43`, is not an ancestor of `release/v3.8.51` and isn't in the repository at all:

```
$ git merge-base --is-ancestor 48fbafb43 origin/release/v3.8.51   → no
$ git cat-file -t 48fbafb43                                       → not present
```

The variable does exist — `adobeFireflySession.ts:134` has `adobeFireflyBrowserEnabled()` reading it, and `adobeFireflyClient.ts` references it too — so only the *test-setup* half is missing. This looks like the guard landed on a different branch, or its PR hasn't merged yet, and this test went out ahead of it.

If that's it, the fix is either to land the guard first, or to fold the one-line `isolateDataDir.ts` change into this PR so the tests have something to pin.

## Worth keeping when it does land

The default is `enabled` (`!== "0"`, so unset means on), which is what makes a *global* guard necessary rather than nice-to-have — a new test file added a year from now is protected without its author knowing this hazard exists. That's the case for test 1.

## One suggestion on test 2

Asserting the setup's **source text**:

```js
assert.match(source, /process\.env\.ADOBE_FIREFLY_BROWSER_REFRESH \|\|= "0";/, ...)
```

pins formatting as well as behaviour. Prettier runs over staged files in this repo, so a quote-style or line-wrap change breaks the test without changing what it guards, and the cheapest-looking fix then is to delete the assertion.

The property you actually want — *"a test may opt back in"* — is directly observable without reading the file:

```js
delete process.env.ADOBE_FIREFLY_BROWSER_REFRESH;
process.env.ADOBE_FIREFLY_BROWSER_REFRESH = "1";
// re-import the setup, assert the value survived
assert.equal(adobeFireflyBrowserEnabled(), true);
```

That fails if someone changes `||=` to `=`, and survives reformatting. Your call — the current version does catch the regression, it just also catches things that aren't one.

patrykkopycinski:
Both points addressed in 2a5569e:

1. **Guard wired into the setup** — `tests/_setup/isolateDataDir.ts` now sets `ADOBE_FIREFLY_BROWSER_REFRESH ||= "0"` with the same rationale-comment treatment as the system-trust and DNS-write guards. The test is no longer red on its own head.
2. **Test 2 is now behavioral** — it spawns the actual setup in a clean subprocess (var unset vs preset to `"1"`) and asserts the resulting env value, instead of regex-matching the setup's source text. Verifies the `||=` semantics end-to-end: unset → `"0"` (browser off), preset → preserved (integration tests can opt in).

Verified: `npx tsx --test tests/unit/adobe-firefly-test-browser-guard.test.ts` → 2/2 pass; eslint clean on both touched files.

abhisheksharma2411 (commented):
Both addressed, and test 2 came back better than the version I sketched.

**Guard wired:** `isolateDataDir.ts:56`, `||=` with the rationale comment sitting alongside the system-trust and DNS-write guards — which is the right neighbourhood, since those are the other two "the suite must never touch the real thing" rules.

**Green under the real runner:**

```
✔ test setup disables Adobe Firefly browser warm
✔ setup disables the browser by default but preserves an explicit opt-in
  2 pass, 0 fail
```

**And test 2 is load-bearing.** Reverting `||=` to `=`:

```
✖ setup disables the browser by default but preserves an explicit opt-in
```

The subprocess approach is better than the `delete process.env…` dance I suggested: spawning the setup with a controlled env asserts both directions (unset → `"0"`, preset `"1"` → `"1"`) from a genuinely clean slate, where mutating the parent's env in-process can only ever test one of them and leaves the runner's state disturbed. It also means test 2 is self-contained — it doesn't depend on the suite's own `--import` chain the way test 1 necessarily does.

One nuance worth knowing, because it cost me a confusing minute: test 1 only passes when the setup is imported, so the bare

```
npx tsx --test tests/unit/adobe-firefly-test-browser-guard.test.ts
```

shows 1/2, while the `npm test` invocation with `--import ./tests/_setup/isolateDataDir.ts` shows 2/2. That's correct behaviour — test 1 is asserting the *ambient* env the suite runs under, so it has to observe it — but if anyone else reproduces with the short command they'll think it's broken. Might be worth a line in the file's header comment.

Nothing further from me.

abhisheksharma2411:
Both addressed, and test 2 came back better than the version I sketched.

**Guard wired:** `isolateDataDir.ts:56`, `||=` with the rationale comment sitting alongside the system-trust and DNS-write guards — which is the right neighbourhood, since those are the other two "the suite must never touch the real thing" rules.

**Green under the real runner:**

```
✔ test setup disables Adobe Firefly browser warm
✔ setup disables the browser by default but preserves an explicit opt-in
  2 pass, 0 fail
```

**And test 2 is load-bearing.** Reverting `||=` to `=`:

```
✖ setup disables the browser by default but preserves an explicit opt-in
```

The subprocess approach is better than the `delete process.env…` dance I suggested: spawning the setup with a controlled env asserts both directions (unset → `"0"`, preset `"1"` → `"1"`) from a genuinely clean slate, where mutating the parent's env in-process can only ever test one of them and leaves the runner's state disturbed. It also means test 2 is self-contained — it doesn't depend on the suite's own `--import` chain the way test 1 necessarily does.

One nuance worth knowing, because it cost me a confusing minute: test 1 only passes when the setup is imported, so the bare

```
npx tsx --test tests/unit/adobe-firefly-test-browser-guard.test.ts
```

shows 1/2, while the `npm test` invocation with `--import ./tests/_setup/isolateDataDir.ts` shows 2/2. That's correct behaviour — test 1 is asserting the *ambient* env the suite runs under, so it has to observe it — but if anyone else reproduces with the short command they'll think it's broken. Might be worth a line in the file's header comment.

Nothing further from me.

diegosouzapw:
Clean, behavioral regression test — the subprocess-based check for the `||=` semantics
(rather than just re-reading the already-mutated in-process env) is the right way to prove
this. Ran it at your head: 2/2 pass. This pairs well with #13289's production-side guard;
recommending both merge. Merge-ready.

PR #13365, opened 2026-09-11
chore(skills): regenerate cli-tunnel after the addArgument parser fix

abhisheksharma2411:
Generated output only — one file, two lines. This is the follow-up #13009 needs, and **`Merge integrity (changelog + generated skills)` has failed on every open PR since it landed this morning.**

## Why it drifted

@dat-nguyen-thanh's #13009 taught the CLI registry parser to read positionals declared with `.addArgument()`. That is correct — `omniroute tunnel create` genuinely takes an optional `[type]`:

```js
// bin/cli/commands/tunnel.mjs:21
.command("create")
.description(t("tunnel.createDescription"))
.addArgument(
  new Argument("[type]", "Tunnel type").choices(VALID_TUNNEL_TYPES).default("cloudflare")
)
```

`skills/cli-tunnel/SKILL.md` was last generated on **8 Sep** (#12970), before that parser fix, so it documents `tunnel create` without the argument the command accepts. The parser improved; the committed artifact didn't follow.

## I checked this was the right direction before running --apply

The generator wanting to *add* `[type]` looked at first like it might be undoing @gonisulaimann's #12368 (*"remove duplicate positional argument in tunnel create command"*), which would have been a regression to re-introduce. It isn't. That PR changed `.command("create [type]")` → `.command("create")` because declaring the positional **twice** crashed the command — the `.addArgument()` declaration stayed, so the argument still exists and documenting it is right.

Worth spelling out because "just run the generator" would have produced the same diff without establishing whether the diff was correct.

## Verification

```
before   Generated: 1 · Unchanged: 45    exit 2
after    Generated: 0 · Unchanged: 46    exit 0
```

`node scripts/skills/generate-agent-skills.mjs --apply` touches this one file and nothing else — `git status` after the run shows only `skills/cli-tunnel/SKILL.md`.

Whole diff:

```diff
-### `tunnel create`
+### `tunnel create [type]`
-omniroute tunnel create
+omniroute tunnel create [type]
```

Unblocks one of the four gates currently red on every PR. The other three are tracked separately: docs counts (#13135), the redaction truncation (#13144 / #13295), and the ESLint `as any` casts from #12950.

### DIFF
--- a/changelog.d/maintenance/13365-regenerate-cli-tunnel-skill.md
+++ b/changelog.d/maintenance/13365-regenerate-cli-tunnel-skill.md
@@ -1,0 +1,1 @@
+- **chore(skills):** regenerate the `cli-tunnel` skill reference so it documents the optional `[type]` argument `omniroute tunnel create` accepts — the committed artifact predated the parser learning to read `.addArgument()` positionals, which left the generated-skills gate red on every PR ([#13365](https://github.com/diegosouzapw/OmniRoute/pull/13365))
--- a/skills/cli-tunnel/SKILL.md
+++ b/skills/cli-tunnel/SKILL.md
@@ -37,12 +37,12 @@ omniroute tunnel
 omniroute tunnel list
 ```
 
-### `tunnel create`
+### `tunnel create [type]`
 
 **Example:**
 
 ```bash
-omniroute tunnel create
+omniroute tunnel create [type]
 ```
 
 ### `tunnel stop <type>`


diegosouzapw:
Correct diagnosis and correct fix — this landed byte-identical on the tip via #13216 a few
hours before this PR could be merged. Closing as already covered; thanks for tracking down the
root cause of the agent-skills-sync red.

abhisheksharma2411:
Agreed, and confirmed rather than taking your word for it — closing this myself so it isn't sitting in your queue.

`963d3c46ef` (#13216, "clear three base-reds … agent-skills sync") landed the same regeneration, and `skills/cli-tunnel/SKILL.md` on the tip is **byte-identical** to this branch's version:

```
$ diff <(git show <this-branch>:skills/cli-tunnel/SKILL.md) \
       <(git show origin/release/v3.8.51:skills/cli-tunnel/SKILL.md)
$   # no output
```

So there's nothing left here. Thanks for the note on the triage — and for saying it was the *correct* diagnosis rather than just closing it; the thing I actually wanted checked was whether regenerating would re-introduce the `[type]` positional that @gonisulaimann deliberately removed in #12368, and it doesn't: #12368 dropped the duplicate `.command("create [type]")` declaration while the `.addArgument()` stayed, so documenting the argument is right. Good to know that reasoning held up independently.

Sorry for the slow reply — I lost access for most of today, which is also why several of your other reviews got silence. Catching up on the rest now.

Issue #13130, opened 2026-09-09
[BUG] TPS is not considering thinking token output speed, important for models with thinking interleaving feature


ahmedhosnypro:
### OmniRoute Version

3.8.50

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Arch Linux, kernel: 7.1.8-arch1-3

### Node.js Version

v26.7.0

### Provider(s) Involved

any kimi models provider

### Model(s) Involved

kimi kx (k2.6), (k2.7), (k3)

### Client Tool

all: copilot, kimi, curl, etc

### Description

For models with interleaving thinking, that can use tool calls between thinking, 
the tokens per second is not being computed accurately, it's only counted for non thinking output

For example kimi models, they heve the following:
Think ➔ Call Tool ➔ Reflect ➔ Call Another Tool ➔ Act

Thinking can execute hundreds of tool calls, which is actual tokens to be counted for the request tokens per seconds

### Steps to Reproduce

Using any kimi models (with any ai clit tool: kimi cli for example) and watching /dashboard/logs TPS for the request, 

### Expected Behavior

Consider interleaving tool calls or thinking output in the tokens per second metrics

### Actual Behavior

tokens/seocond only count final request response, the model may spend much more time thinking and using tool calls, which may be 10s or 100s, 
while the final response can be a small output, 

this affects the tps to be very small for kimi models, as it doesnt count any actual working output tokens between thinking (interleaving actions)

### Test Impact

Needs a new integration test

### Error Logs / Output

```shell

```

ahmedhosnypro:
What about considering t/s after the occurance of the first token recived, and add/show metric for time for first token?

ahmedhosnypro:
## Verification update (2026-09-11) — partially confirmed, root causes identified

Verified against a live instance with `nvidia/moonshotai/kimi-k3` (thinking model, interleaved tool calls) using stream instrumentation + the local `storage.sqlite` / call-log artifacts + code audit.

### What is NOT a bug: the token numerator on this provider

- Streamed a request with `max_tokens: 3000` → `finish_reason: length` while only ~370 visible content tokens were produced (+ ~6.3k chars of `reasoning_content`). The cap could only have been reached if NVIDIA's `completion_tokens` **includes** thinking tokens — and OmniRoute logs that value verbatim into `call_logs.tokens_out` (`src/lib/usage/tokenAccounting.ts:94-100` → `src/lib/usage/callLogs.ts:525`).
[...]

PR #12148, opened 2026-08-30
feat(ui): add TTFT column alongside TPS in RequestLoggerV2 with backend telemetry

b3nw:
## Summary

Adds Time-to-First-Token (TTFT) metrics to call log telemetry and surfaces a sortable TTFT column in the RequestLogger dashboard.

- Adds migration `171_call_logs_ttft_ms.sql` adding nullable `ttft_ms INTEGER` to `call_logs`.
- Records TTFT during SSE streaming responses in `chatCore.ts` and `attemptLogging.ts`. Failed or aborted streams safely record `NULL` to avoid poisoning statistical aggregations.
- Surfaces TTFT column in `RequestLoggerV2.tsx` with adaptive formatting (`ms` for `<1000ms`, `s` for `>=1000ms`) and full null-aware sorting (`ttft_asc` and `ttft_desc`).
- Adds changelog fragment `changelog.d/features/ttft-column-requestlogger.md`.

## Related Issues

N/A

## Validation

- [x] Change type: UI / DB / routing
- [x] Focused tests and category gates from the golden path
- [x] `npm run lint`
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR

## Tests Added Or Updated

- `tests/unit/call-logs-ttft.test.ts`
- `tests/unit/shared/components/RequestLoggerV2.test.ts`

## Coverage Notes

- Covers `src/lib/usage/callLogs.ts`, `open-sse/handlers/chatCore/attemptLogging.ts`, and `src/shared/components/RequestLoggerV2.tsx`.
- Validates TTFT calculation, null persistence on failure, formatting thresholds, and table sort invariants.

## Reviewer Notes

- Includes database migration `171_call_logs_ttft_ms.sql` (additive nullable column, backward-compatible with existing call log records).

PR #13373, opened 2026-09-11
fix(dashboard): TPS over generation time (duration − TTFT) + reasoning-aware numerator (#13130)

ahmedhosnypro:
## Summary

Fixes the TPS (tokens-per-second) metric confirmed in #13130: dashboards and usage history computed a **wall-clock** rate (`tokens_out / full request duration`), which collapses for thinking/interleaved models (kimi k2.6/k2.7/k3) whose queue + prefill + first-token window dwarfs the generation window.

## Verification done on a live instance (nvidia/moonshotai/kimi-k3, interleaved thinking + tool calls)

- Streamed a request capped at `max_tokens: 3000` → `finish_reason: length` while visible content was only ~370 tokens (+6.3k chars of `reasoning_content`). The cap can only be reached if `completion_tokens` **includes** thinking tokens, and OmniRoute's recorded `tokens_out` matched the stream exactly (recorded TPS = 14.4 tok/s = true all-output throughput). So the numerator was already correct **for shapes that fold reasoning into `completion_tokens`**.
- What WAS wrong: `RequestLoggerV2` + `getModelLatencyStats` divided by full wall-clock duration, contradicting the repo's own rule in `open-sse/utils/generationThroughput.ts` ("tok/s MUST exclude TTFT"; the response-header path already follows it).
- Latent gap: nothing added `tokens_reasoning` into the rate — a provider that excludes reasoning from `completion_tokens` while reporting `completion_tokens_details.reasoning_tokens` (the reporter's kimi OAuth endpoints are suspected) would still undercount.

## Changes

- **`src/shared/utils/logTps.ts` (new)**: `resolveGenerationMs()` (duration − TTFT, full-duration fallback) and `resolveTpsOutputTokens()` (`max(tokens_out, tokens_reasoning)` — never double counts, since reasoning is a subset when the provider folds it in).
- **`call_logs.ttft_ms`** (migration `176_call_logs_ttft_ms.sql` + `ensureCallLogsColumns` reconciliation): TTFT from `streamTiming` is threaded through `persistAttemptLogs` for streaming completions; NULL for non-streaming/legacy rows.
- **Call-logs API**: summary rows expose `ttft`; dashboard grid adds an opt-in TTFT column (the `requestLogger.columns.ttft` i18n key already existed in all 51 locales), the TPS cell tooltip breaks down TTFT / generation window / reasoning tokens, and the request-detail modal shows a TTFT tile with the generation window.
- **`getModelLatencyStats`**: per-row TPS sample now uses generation time + the reasoning-aware numerator (feeds auto-combo speed ranking).
- **Playground** (Chat/Compare client-side TPS): same generation-time formula via the shared helper.

## Tests

- New `tests/unit/log-tps-13130.test.ts` — generation-window math, reasoning-aware numerator (incl. the no-double-count and provider-excludes-reasoning cases), `accumulateLatencySample` behavior, and `ttft_ms` persistence round-trip.
- Updated `latency-stats-ttft-6875.test.ts` and `playground-stream-metrics.test.ts` (they pinned the old wall-clock formula), `db-schema-columns-split.test.ts` covers the reconciler.
- `node --import tsx/esm --test` on all touched suites: green; `eslint` (with suppressions): clean; `npm run typecheck:core`: clean.

Refs #13130

---

⚠️ base-red inherited: #12732 — the `release/v3.8.51` tip was red when this PR was opened; any suite failures that also fail on base are inherited, not from this change.


diegosouzapw:
This is the more complete fix for the TPS problem — it doesn't just add the TTFT column
(which #12148 also does, with an identical `ALTER TABLE call_logs ADD COLUMN ttft_ms`
statement), it actually fixes the root formula bug (wall-clock vs. generation-time) and adds
the reasoning-aware numerator that `getModelLatencyStats`/auto-combo speed ranking and the
playground also needed. 18/18 tests pass locally. One mechanical blocker: the new migration
`176_call_logs_ttft_ms.sql` collides with `176_xp_action_counts.sql` already on the release
tip — needs renumbering to `179`. We're also reconciling this with #12148, which covers the
same DB column/UI surface; expect a note here once that's settled.

diegosouzapw:
Maintainer note for the merge: this will be squash-merged with `Co-authored-by: b3nw <189466+b3nw@users.noreply.github.com>` — the TTFT plumbing here (the `call_logs.ttft_ms` migration, the attemptLogging → chatCore → callLogs threading and the RequestLoggerV2 column) is the same infrastructure @b3nw landed first in #12148. #12148 will be closed once this merges, with credit preserved through that trailer. No action needed from you.

PR #13394, opened 2026-09-12
feat(cli): make startup readiness budget configurable (#13369)

### LINKED ISSUE #13369: feat(cli): make the 60s startup readiness budget configurable (warning fires on every slow cold start)
## Problem

The CLI polls `/api/monitoring/health` for a fixed 60 seconds after starting the server, then prints:

```
⚠ Server did not respond within 60s. It may still be starting, or may have failed silently.
```

On slow cold starts the server comes up fine, just later than the budget, so the warning prints on **every single start** while nothing is actually wrong. Users reasonably read it as an error.

The budget is hardcoded at both the definition and the call site, with no env var or flag to raise it:

- `bin/cli/utils/pid.mjs:76` — `export async function waitForServer(port, timeout = 60000)`
- `bin/cli/commands/serve.mjs:455` — `waitForServer(dashboardPort, 60000)`

## Evidence

Reported in discussion #10754 by @Dr-LEO-MS, on Windows with a global npm install. Their `app.log` shows the boot taking roughly **6 minutes**:

```
08:05:14  first STARTUP component
08:05:35  schema migrations applying (api_keys columns)
08:11:25  EmbedWsProxy / LiveWS listening
```

The health route answers well past 60 s, so the warning always fires and the server then works normally. The user's question was literally "omniroute is working but why is it showing every time".

The budget was already raised once from 15 s to 60 s for exactly this class of Windows cold start (see the comment at `bin/cli/utils/pid.mjs:66`, issue #2460). Six minutes is past that raise too, and raising the constant again just moves the same wall.

## Proposed solution

Make the readiness budget configurable instead of guessing a universal constant:

1. Read an env var, e.g. `OMNIROUTE_READY_TIMEOUT_MS`, falling back to the current 60000 default. The CLI already has this pattern for `OMNIROUTE_HTTP_TIMEOUT_MS` (`bin/cli/api.mjs:225`), including registration in `bin/cli/commands/env.mjs`.
2. Optionally accept a `--ready-timeout` flag on `omniroute serve`.
3. Soften the message when the process is demonstrably alive. The TCP fallback already distinguishes "listening but route not mounted" from "hung"; when the port is listening and the child has not exited, the wording should say still starting rather than may have failed silently.

## Acceptance criteria

- [ ] The readiness budget is settable via env var, with 60000 remaining the default.
- [ ] The env var is registered in the CLI env listing.
- [ ] A unit test covers the override being honored.
- [ ] Documentation notes it where slow-start troubleshooting is described.

## Additional context

- Discussion: #10754
- Prior raise of the same constant: #2460

sagarithm:
## Problem

On slow cold starts (such as on Windows with filesystem watchers, antivirus scans, or heavy schema migrations as reported in #10754), `omniroute serve` polls `/api/monitoring/health` for a hardcoded 60 seconds before printing:

```text
⚠ Server did not respond within 60s. It may still be starting, or may have failed silently.
```

Because the server boots normally just slightly after the budget, the warning was emitted on every cold start despite healthy operation.

## Solution

This PR implements the solution outlined and confirmed in #13369:

1. **Configurable Readiness Budget**:
   - `OMNIROUTE_READY_TIMEOUT_MS` environment variable is read with 60000ms default preserved.
   - `--ready-timeout <ms>` CLI flag is added to `omniroute serve` to override or specify the timeout.
   - Helper `resolveReadyTimeout(opts)` cleanly resolves flag > env var > 60000 default (safely falling back on invalid/non-positive values).
   - `waitForServer` in `bin/cli/utils/pid.mjs` defaults `timeout` to `OMNIROUTE_READY_TIMEOUT_MS` so all un-parameterized calls benefit automatically.

2. **Accurate Diagnostic Reporting**:
   - `reportReadinessTimeout` dynamically formats the message with `${seconds}s` instead of hardcoding `within 60s`.

3. **CLI Env Registration & Documentation**:
   - `OMNIROUTE_READY_TIMEOUT_MS` is registered in `bin/cli/commands/env.mjs` (`OMNIROUTE_ENV_VARS` and `ENV_DEFAULTS`).
   - Documented in `docs/reference/ENVIRONMENT.md`, `.env.example`, and added to the troubleshooting quick fixes in `docs/guides/TROUBLESHOOTING.md`.

4. **Testing**:
   - Added unit test suite in `tests/unit/cli-ready-timeout-13369.test.ts` covering option overrides, env var parsing, fallback behavior, dynamic warning formatting, and CLI env registration.

Closes #13369

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1793,6 +1793,9 @@ APP_LOG_TO_FILE=true
 # Per-attempt HTTP timeout for CLI → server calls (milliseconds). Default: 30000.
 # OMNIROUTE_HTTP_TIMEOUT_MS=30000
 
+# Maximum milliseconds to wait for server readiness during `omniroute serve`. Default: 60000.
+# OMNIROUTE_READY_TIMEOUT_MS=60000
+
 # Set to 1 to print retry/backoff details to stderr during CLI commands.
 # OMNIROUTE_VERBOSE=0
 
--- a/bin/cli/commands/env.mjs
+++ b/bin/cli/commands/env.mjs
@@ -13,6 +13,7 @@ const OMNIROUTE_ENV_VARS = [
   "OMNIROUTE_API_KEY",
   "OMNIROUTE_BASE_URL",
   "OMNIROUTE_HTTP_TIMEOUT_MS",
+  "OMNIROUTE_READY_TIMEOUT_MS",
 ];
 
 const ENV_DEFAULTS = {
@@ -20,6 +21,7 @@ const ENV_DEFAULTS = {
   DASHBOARD_PORT: "20128",
   DATA_DIR: "~/.omniroute",
   NODE_ENV: "production",
+  OMNIROUTE_READY_TIMEOUT_MS: "60000",
 };
 
 export function registerEnv(program) {
--- a/bin/cli/commands/serve.mjs
+++ b/bin/cli/commands/serve.mjs
@@ -68,6 +68,12 @@ export function registerServe(program) {
       t("serve.tls_key") ||
         "Path to the TLS private key (PEM) to serve HTTPS (also OMNIROUTE_TLS_KEY)"
     )
+    .option(
+      "--ready-timeout <ms>",
+      t("serve.ready_timeout") ||
+        "Maximum milliseconds to wait for server readiness (also OMNIROUTE_READY_TIMEOUT_MS)",
+      parseInt
+    )
     .action(async (opts) => {
       await runServe(opts);
     });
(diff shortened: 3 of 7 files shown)


diegosouzapw:
Hi @sagarithm — thanks for this! The exact same feature (configurable
`OMNIROUTE_READY_TIMEOUT_MS` / `--ready-timeout`, closing #13369) already landed on
`release/v3.8.51` via #13433, with the identical env var name and CLI flag. You can
verify with `git show origin/release/v3.8.51:bin/cli/utils/pid.mjs` — `resolveReadyTimeoutMs()`
is there, `.env.example` and `docs/reference/ENVIRONMENT.md` already document it.
Since it's already shipped, I'm marking this as superseded and closing it. Appreciate
the clean implementation and test coverage — if you run into the phantom-timeout issue
again on a cold boot, #12484 (a complementary fix for the *per-probe* fetch timeout
that still applies to slow first responses) may be worth a look.

Issue #6925, opened 2026-07-11
fix(providers): Private LAN OpenAI-compatible embeddings provider is treated as bearer-auth remote provider


hertznsk:
### Summary

A user-configured OpenAI-compatible embeddings provider node pointing to a private LAN Ollama endpoint fails even though the upstream requires no authentication.

Example base URL:

\`\`\`text
http://10.10.0.181:11434/v1
\`\`\`

Requesting an embedding model through that provider (e.g. `my-embed/nomic-embed-text-v2-moe:latest`) returns:

\`\`\`json
{"error":{"message":"No valid authentication token for provider my-embed. Check provider credentials.","type":"upstream_error","code":"upstream_error"}}
\`\`\`

The same Ollama endpoint works directly (`POST /v1/embeddings` → 200) from both the host and inside the OmniRoute container, so the upstream is healthy.

### Root cause

`src/lib/embeddings/service.ts` (~lines 117-130) loads dynamic `provider_nodes` as no-auth local embedding providers only when the hostname is:

- `localhost`
- `127.0.0.1`
- `172.16.0.0/12` (regex)

That excludes normal RFC1918 LAN ranges such as `10.0.0.0/8` and `192.168.0.0/16`.

When the LAN node is excluded from `dynamicProviders`, the later fallback branch (~lines 156-183) resolves the same `provider_nodes` row as a custom embedding provider with:

\`\`\`ts
authType: "apikey"
authHeader: "bearer"
\`\`\`

For a no-auth Ollama provider whose connection row has no API key, `handleEmbedding` (`open-sse/handlers/embeddings.ts:179-194`) then returns the 401 auth error instead of routing to the provider without `Authorization`.

### Expected behavior

OpenAI-compatible provider nodes on private/local addresses, such as:

- `http://10.10.0.181:11434/v1`
- `http://192.168.1.10:11434/v1`
- `http://172.16.x.x:...`

should be accepted as local/no-auth dynamic embedding providers, as long as cloud metadata endpoints remain blocked. The upstream request should be sent to `<baseUrl>/embeddings` without an `Authorization` header.

### Actual behavior

The private LAN node is not classified as local/no-auth. It falls through to the remote custom-provider fallback and is treated as bearer-token authenticated, causing a 401 for no-auth Ollama.

### Recommended fix

Please reuse the existing outbound URL guard helpers instead of duplicating hostname regexes.

The codebase already has `src/shared/network/outboundUrlGuard.ts` with:

[...]

PR #13398, opened 2026-09-12
fix(embeddings): send stored API key on private-host embeddings nodes

### LINKED ISSUE #13234: [BUG] custom OpenAI-compatible embeddings provider — API key passes "Check" but Authorization header is not sent on outbound /v1/embeddings
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

_No response_

### Node.js Version

_No response_

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

Custom OpenAI-compatible providers of type "embeddings" do not forward their configured API key on outbound requests. The key is saved and validates successfully in the UI ("Check" reports valid, and the inline model test reports valid), but when a request is proxied to the upstream POST /v1/embeddings, OmniRoute sends no authentication header at all. The upstream rejects the call with HTTP 401.

This is specific to the embeddings proxy path. A chat/completions custom provider configured against the same upstream base URL with the same key sends Authorization: Bearer <key> as expected and authenticates fine. So  the credential is stored correctly but is dropped only when routing embeddings requests.

Captured on the upstream side, the request from OmniRoute contains only these headers: Accept, Accept-Encoding, Accept-Language, Connection, Content-Length, Content-Type, Sec-Fetch-Mode, User-Agent — no Authorization,  no Api-Key, no X-Api-Key. The presence of Sec-Fetch-Mode suggests the embeddings request is issued from a fetch path that does not carry the Authorization header.

This area has had related fixes before (#7785 "apply optional API-key policy consistently to embeddings and web-fetch routes", #6925 "Private LAN OpenAI-compatible embeddings provider is treated as bearer-auth remote  provider"), so this may be a regression.

Workaround: add Authorization: Bearer <key> as a manual custom header on the provider, or disable authentication on the upstream.

### Steps to Reproduce

1. Add a custom OpenAI-compatible provider whose type / connection kind is "embeddings", pointing at an upstream base URL (e.g. https://example.com/v1), and set its API key.
  2. Click "Check" — the key is reported as valid. While configuring the model, the inline "test model" check also reports valid.
  3. Send a request through OmniRoute to that model: POST /v1/embeddings with a body like {"model":"<provider>/<model>","input":["hello","world"]}.
  4. Observe the request OmniRoute makes to the upstream. It arrives with no authentication header at all — no Authorization, no Api-Key, no X-Api-Key. The only headers present are Accept, Accept-Encoding,   Accept-Language, Connection, Content-Length, Content-Type, Sec-Fetch-Mode, User-Agent.
  5. The upstream rejects the call with HTTP 401.
  6. For comparison, configure a chat/completions custom provider against the same upstream base URL with the same key and call POST /v1/chat/completions — the outbound request includes Authorization: Bearer <key> and passes auth.

### Expected Behavior

The outbound /v1/embeddings request should carry the provider's configured credential as Authorization: Bearer <key>, exactly the way the chat/completions path already does. A key that passes "Check" must be attached  to the actual proxied request, so an embeddings provider configured with a valid key does not fail with 401 at the upstream.

### Actual Behavior

The embeddings provider stores the API key and reports it valid during "Check" and the inline model test, but the key is never applied to the actual proxied request. OmniRoute calls the upstream POST /v1/embeddings with no authentication header of any kind (no Authorization, no Api-Key, no X-Api-Key). The upstream therefore responds with HTTP 401. A chat/completions provider pointed at the same upstream base URL with the same key sends Authorization: Bearer <key> correctly, so the credential is stored but dropped specifically on the embeddings proxy path. The Sec-Fetch-Mode header present in the outbound request suggests the embeddings call is issued from a fetch path that does not forward Authorization.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Add a regression test covering the embeddings proxy path that asserts the configured provider credential is present on the outbound request (Authorization: Bearer <key>), mirroring the existing coverage for chat/completions. Include a case where the upstream requires auth and returns 401 without it, so a missing header fails the test rather than silently passing. Given the earlier fixes in this area (#7785, #6925), a shared assertion that every provider-typed route — chat/completions, embeddings, rerank — attaches the same auth header for the same provider config would prevent one route from regressing independently again.
```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

HouMinXi:
## What this solves

A custom OpenAI-compatible embeddings node can pass dashboard Check and still return HTTP 401 on `/v1/embeddings`.

Check uses `buildBearerHeaders` and sends `Authorization: Bearer` whenever a key is stored. The embeddings proxy does not: #6925 classified every private-host / CGNAT node as `authType: "none"` before credentials were loaded, so `buildAuth` dropped the key. Chat/completions against the same node and key worked.

Live ids look like `100.64.x` (CGNAT) or RFC1918 LAN hosts (Ollama, local OpenAI-compatible gateways).

Fixes #13234.

## Fix

After the existing ollama-local / lmstudio optional hydrate, if the embeddings config is still `authType: "none"` and no credentials were resolved, look up the stored connection. When that connection has a non-empty `apiKey` or `accessToken`, promote the config to bearer and send the key. Keyless LAN nodes stay no-auth (#6925). Cloud-metadata hosts stay blocked.

The hydrate is skipped when credentials are already set (synced-endpoint route, explicit caller).

## Tests

`tests/unit/embeddings-lan-keyed-auth-13234.test.ts`:

- CGNAT node with a stored key: outbound `Authorization: Bearer …`
- keyless 10.x node: no Authorization
- LAN node whose connection record has an empty key: no Authorization

`tests/unit/embeddings-lan-noauth-6925.test.ts` still 3/3.

## Related (not this bug)

- #13368 adds a public-host custom-embeddings test only; it does not change the proxy path.
- #6925 is the keyless-LAN classification this patch keeps.

### DIFF
--- a/changelog.d/fixes/13234-embed-lan-keyed-auth.md
+++ b/changelog.d/fixes/13234-embed-lan-keyed-auth.md
@@ -1,0 +1,1 @@
+- **fix(embeddings):** LAN/CGNAT OpenAI-compatible embeddings nodes with a stored API key now send `Authorization: Bearer` on the outbound request, matching dashboard Check. Keyless LAN nodes stay no-auth ([#6925](https://github.com/diegosouzapw/OmniRoute/issues/6925)) ([#13234](https://github.com/diegosouzapw/OmniRoute/issues/13234))
--- a/src/lib/embeddings/service.ts
+++ b/src/lib/embeddings/service.ts
@@ -344,7 +344,7 @@ export async function createEmbeddingResponse(
     ) {
       credentials = localCredentials;
     }
-  } else if (providerConfig.authType === "none") {
+  } else if (!credentials && providerConfig.authType === "none") {
     // #13234: private-host nodes are classified no-auth so a keyless
     // LAN Ollama still works (#6925). A stored API key on that same
     // node must still ride outbound, matching dashboard Check.
--- a/src/lib/embeddings/service.ts
+++ b/src/lib/embeddings/service.ts
@@ -344,6 +344,31 @@ export async function createEmbeddingResponse(
     ) {
       credentials = localCredentials;
     }
+  } else if (providerConfig.authType === "none") {
+    // #13234: private-host nodes are classified no-auth so a keyless
+    // LAN Ollama still works (#6925). A stored API key on that same
+    // node must still ride outbound, matching dashboard Check.
+    const keyedCredentials = await getProviderCredentials(credentialsProviderId);
+    if (
+      keyedCredentials &&
+      !("allRateLimited" in keyedCredentials) &&
+      !("allExpired" in keyedCredentials)
+    ) {
+      const token =
+        (typeof (keyedCredentials as { apiKey?: unknown }).apiKey === "string" &&
+          (keyedCredentials as { apiKey?: string }).apiKey) ||
+        (typeof (keyedCredentials as { accessToken?: unknown }).accessToken === "string" &&
+          (keyedCredentials as { accessToken?: string }).accessToken) ||
+        "";
+      if (token) {
+        credentials = keyedCredentials;
+        providerConfig = {
+          ...providerConfig,
+          authType: "apikey",
+          authHeader: "bearer",
+        };
+      }
+    }
   }
 
   // #474: when the request used a bare model name (no "/" — e.g. an alias that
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks for #13234 — this is exactly the right fix: promoting to bearer auth only when a real
stored key is found, leaving genuinely keyless LAN/CGNAT nodes untouched. Verified locally: the
new test suite passes 3/3 against the actual `createEmbeddingResponse` production code (not a
local re-implementation). We're merging this as the canonical fix for #13234. There's a
competing PR (#13417) that found the same bug from a different angle but its fix has a
regression for keyless nodes — we'll credit that diagnosis and may port a couple of its test
cases here if they add coverage.

PR #13417, opened 2026-09-12
fix(providers): forward API key on outbound embeddings requests for custom providers (#13234)

### LINKED ISSUE #13234: [BUG] custom OpenAI-compatible embeddings provider — API key passes "Check" but Authorization header is not sent on outbound /v1/embeddings

KooshaPari:
## Problem

Custom OpenAI-compatible providers with `type: "embeddings"` store their API key correctly and the dashboard "Check" validation passes, but the actual proxied `POST /v1/embeddings` request arrives at the upstream with no `Authorization` header at all, resulting in HTTP 401.

Chat/completions requests to the same upstream with the same key work fine — the credential is dropped only on the embeddings proxy path.

## Root Cause

`buildDynamicEmbeddingProvider()` in `open-sse/config/embeddingRegistry.ts` hardcoded `authType: "none"` and `authHeader: "none"` for all dynamic provider nodes. When the embeddings handler called `buildAuth()`, it checked the provider config's `authType`, found `"none"`, and skipped setting the `Authorization` header entirely — even though `getProviderCredentials()` had successfully retrieved the API key from the connections table.

The parallel fallback path (`deriveEmbeddingProviderForChatProvider` and the explicit `matchingNode` path in `service.ts`) correctly defaults to `authType: "apikey"`, so this was specific to the `buildDynamicEmbeddingProvider` path.

## Fix

Default `buildDynamicEmbeddingProvider()` to `authType: "apikey"` and `authHeader: "bearer"`, matching the behavior of `deriveEmbeddingProviderForChatProvider()`. Truly keyless local providers (ollama, lmstudio) are resolved through dedicated code paths before this fallback is reached.

## Tests

5 regression tests added/updated in `tests/unit/embeddings-proxy-forwarding.test.ts`:

- `buildDynamicEmbeddingProvider` defaults to apikey/bearer auth
- Trailing slash stripping preserves correct baseUrl
- Invalid prefix/suffix rejected with clear error
- `createEmbeddingResponse` attaches Bearer header when credentials exist
- Connection-level proxy forwarding (pre-existing)

All 18 embeddings tests pass (5 proxy + 13 auth).

Closes #13234

### DIFF
--- a/open-sse/config/embeddingRegistry.ts
+++ b/open-sse/config/embeddingRegistry.ts
@@ -64,11 +64,15 @@ export function buildDynamicEmbeddingProvider(node: EmbeddingProviderNodeRow): E
     throw new Error(`Invalid provider_node prefix "${node.prefix}": must not contain / or spaces`);
   }
   const baseUrl = node.baseUrl.replace(/\/+$/, "");
+  // Default to apikey/bearer auth — matches deriveEmbeddingProviderForChatProvider
+  // and ensures custom OpenAI-compatible providers forward their configured API key
+  // on outbound requests. Truly keyless local providers (ollama, lmstudio) are
+  // resolved through dedicated code paths before this fallback is reached.
   return {
     id: node.prefix,
     baseUrl: `${baseUrl}/embeddings`,
-    authType: "none",
-    authHeader: "none",
+    authType: "apikey",
+    authHeader: "bearer",
     models: [],
   };
 }
--- a/tests/unit/embeddings-proxy-forwarding.test.ts
+++ b/tests/unit/embeddings-proxy-forwarding.test.ts
@@ -4,6 +4,8 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import http from "node:http";
+import { buildDynamicEmbeddingProvider } from "../../open-sse/config/embeddingRegistry.ts";
+import type { EmbeddingProviderNodeRow } from "../../open-sse/config/embeddingRegistry.ts";
 
 // Isolate the DB to a temp dir BEFORE importing any module that opens it.
 const TEST_DATA_DIR = fs.mkdtempSync(path.join(os.tmpdir(), "omniroute-embed-proxy-"));
@@ -41,10 +43,6 @@ async function withHttpServer(
 }
 
 test("embeddings forward the connection-level (key) pinned proxy to the upstream fetch", async () => {
-  // T14 Proxy Fast-Fail performs a real TCP reachability check before running
-  // the request inside the proxy context, so the "proxy" must be a real,
-  // reachable local listener (its actual response body is irrelevant here —
-  // the assertion is about which proxy context the upstream fetch observes).
   await withHttpServer(
     (_req, res) => {
       res.writeHead(200);
@@ -60,8 +58,6 @@ test("embeddings forward the connection-level (key) pinned proxy to the upstream
         apiKey: "mistral-test-key",
       });
 
-      // Pin a proxy at the connection ("key") level — the most specific level,
-      // exactly like a user would configure per-connection in the dashboard.
       await settingsDb.setProxyForLevel("key", (connection as any).id, {
         type: "http",
         host: proxyUrl.hostname,
@@ -96,9 +92,6 @@ test("embeddings forward the connection-level (key) pinned proxy to the upstream
         globalThis.fetch = originalFetch;
       }
 
-      // The upstream fetch must have run inside the AsyncLocalStorage proxy
-      // context carrying the connection's pinned proxy — not "direct" (no
-      // context) and not the env-var proxy fallback.
       assert.equal(
         capturedProxySource,
         "context",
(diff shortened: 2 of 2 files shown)


diegosouzapw:
Thanks for #13234. Holding this for one change before merge.

`buildDynamicEmbeddingProvider()` is only called for local provider nodes (its doc comment says callers filter by localhost). Most of those servers — ollama, llama.cpp, LM Studio behind a custom node — are keyless, so switching the default from `authType: "none"` to `apikey`/`bearer` changes the auth contract for every one of them.

The new integration test also exercises a `mistral` connection, which never goes through `buildDynamicEmbeddingProvider`, so it would pass with or without the fix.

Could you derive the auth type from whether the node or connection actually has an API key (bearer when present, none otherwise), and add a test that builds a dynamic node with and without a key and asserts the outbound `Authorization` header in both cases?

diegosouzapw:
Thanks for tracking down #13234 — the root cause diagnosis (buildDynamicEmbeddingProvider
hardcoding authType:"none") is correct. However, tracing the call path shows this fix has a real
regression: buildDynamicEmbeddingProvider() is only reached for hosts already classified as
private/no-auth-local, so forcing apikey/bearer for all of them means a genuinely keyless local
node (no stored key) now hits the shared credential-required branch in service.ts and gets a
hard 400 instead of proceeding keyless — breaking working local Ollama/LM Studio setups.
#13398 fixes the same bug at the request-resolution layer, only promoting to bearer when a real
stored key is found, which is exactly what we'd ask for here too. We're going to merge that one
and credit your diagnosis in the changelog/PR trail. Feel free to port any test cases from here
that add coverage #13398 doesn't have.

diegosouzapw:
Obrigado pelo diagnóstico — o bug do #13234 era real, mas o fix já entrou por outro caminho enquanto esta PR aguardava.

A #13398 (`67f6c039`) implementa a promoção **condicional** em `src/lib/embeddings/service.ts`: quando `authType === "none"` e nenhuma credencial resolveu, ela chama `getProviderCredentials()` e só então reescreve para `apikey`/`bearer` **se houver uma chave não-vazia**. A issue #13234 já está fechada por ela.

O motivo de não portar nada daqui: esta PR inverte o default **incondicionalmente** em `open-sse/config/embeddingRegistry.ts:67-75`. `buildDynamicEmbeddingProvider()` tem um único chamador (`service.ts:145`) e o filtro dele é `isNoAuthLocalEmbeddingHost(hostname)` — ou seja, todo nó que chega ali já foi classificado como privado/LAN sem auth. Forçar `apikey` nesse conjunto manda um Ollama/LM Studio genuinamente keyless para `getProviderCredentials()` → null → 400 duro. Seria uma regressão em embeddings locais sem chave.

Fechando como subsumida. O crédito pelo diagnóstico do #13234 é seu.

PR #13419, opened 2026-09-12
fix(claude-web): add charset=utf-8 to Content-Type headers to fix Arabic/Persian UTF-8 mojibake (#13416)

### LINKED ISSUE #13416: [BUG] Arabic/Persian UTF-8 text is returned as mojibake by Claude Web endpoint
### OmniRoute Version

3.8.50

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

v24.13.0

### Provider(s) Involved

Claude Web

### Model(s) Involved

Claude

### Client Tool

VS Code Chat

### Description

Persian/Arabic UTF-8 responses from the Claude Web endpoint are incorrectly decoded and returned as mojibake when used through VS Code Chat.

For example, `سلام` is returned as `Ø³Ù„Ø§Ù…` instead of the correct UTF-8 text.

### Steps to Reproduce

1. Configure OmniRoute 3.8.50 with the Claude Web endpoint.
2. Use the endpoint from VS Code Chat.
3. Send a prompt that produces a Persian/Arabic response.
4. Observe the response in VS Code Chat.

### Expected Behavior

Persian/Arabic text should be returned and displayed correctly as UTF-8.

Example:
`سلام! این پیام یک تست ساده به نظر می‌رسه، نه یک درخواست واقعی برای انجام کاری خاص.`

### Actual Behavior

Persian/Arabic UTF-8 text is returned as mojibake.

Example:
`Ø³Ù„Ø§Ù…! Ø§ÛŒÙ† Ù¾ÛŒØ§Ù… ÛŒÚ© ØªØ³Øª Ø³Ø§Ø¯Ù‡ Ø¨Ù‡ Ù†Ø¸Ø± Ù…ÛŒâ€ŒØ±Ø³Ù‡ØŒ Ù†Ù‡ ÛŒÚ© Ø¯Ø±Ø®ÙˆØ§Ø³Øª ÙˆØ§Ù‚Ø¹ÛŒ Ø¨Ø±Ø§ÛŒ Ø§Ù†Ø¬Ø§Ù… Ú©Ø§Ø±ÛŒ Ø®Ø§Øµ.`

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Add an automated test covering UTF-8/Unicode responses from the Claude Web endpoint, including Persian and Arabic text, to ensure the response is decoded correctly.

No explicit error is reported. The issue is an incorrect UTF-8 decoding/encoding of the response body.
```

### Screenshots

Actual:

Image

Expected:

Image

### Additional Context

The issue occurs specifically with Persian/Arabic Unicode text. The same endpoint works correctly for non-Persian/Arabic text.

OmniRoute was installed globally via npm on Windows 11 and is being used as the API endpoint for VS Code Chat.

### Validation Plan

* Send a Persian response through the Claude Web endpoint and verify that UTF-8 characters are preserved.
* Test Arabic text as well as Persian text.
* Verify the same response in VS Code Chat without mojibake.

KooshaPari:
## Summary

Fixes #13416

Arabic/Persian UTF-8 text from the Claude Web endpoint is returned as mojibake (e.g.,  → ). The root cause is missing `charset=utf-8` in the `Content-Type` headers, causing some HTTP clients to fall back to ISO-8859-1/Latin-1 decoding.

## Changes

| File | Change |
|------|--------|
| `open-sse/executors/claude-web/stream.ts` | Append `; charset=utf-8` to Content-Type in `responseHeaders()` helper, with a guard against double-appending |
| `open-sse/handlers/chatCore/responseHeaders.ts` | Hardcode `text/event-stream; charset=utf-8` in `buildStreamingResponseHeaders()` |
| `tests/unit/claude-web-utf8-mojibake-13416.test.ts` | **New** — 11 regression tests covering Persian, Arabic, mixed-script, emoji, and chunk-boundary-split scenarios across both streaming and buffered response paths |

## Technical Detail

While RFC 8259 defaults `application/json` to UTF-8 and the SSE spec defaults `text/event-stream` to UTF-8, VS Code Chat (and potentially other HTTP clients) fall back to ISO-8859-1 when no charset is declared. The UTF-8 bytes for Persian/Arabic characters (e.g., `D8 B3 D9 84 D8 A7 D9 85` for `سلام`) are then misinterpreted as Latin-1, producing the classic mojibake pattern.

## Test Results

- ✅ 11 new UTF-8 mojibake regression tests — all pass
- ✅ 7 existing `claude-web-stream` tests — all pass
- ✅ 13 response header tests — all pass
- ✅ 8 `adaptive-admission-lifecycle` tests — all pass

### DIFF
--- a/open-sse/executors/claude-web/stream.ts
+++ b/open-sse/executors/claude-web/stream.ts
@@ -461,8 +461,11 @@ function protocolErrorBody(): Record<string, unknown> {
 }
 
 function responseHeaders(contentType: string, metadata: Record<string, string>): Headers {
+  const charsetAware = contentType.includes("charset")
+    ? contentType
+    : `${contentType}; charset=utf-8`;
   const headers = new Headers({
-    "Content-Type": contentType,
+    "Content-Type": charsetAware,
     "Cache-Control": "no-cache",
   });
   const headerNames: Record<string, string> = {
--- a/open-sse/handlers/chatCore/responseHeaders.ts
+++ b/open-sse/handlers/chatCore/responseHeaders.ts
@@ -272,7 +272,7 @@ export function buildStreamingResponseHeaders(
 
   const responseHeaders: Record<string, string> = {
     ...Object.fromEntries(forwardedHeaders),
-    "Content-Type": "text/event-stream",
+    "Content-Type": "text/event-stream; charset=utf-8",
     "Cache-Control": "no-cache, no-transform",
     Connection: "keep-alive",
     "X-Accel-Buffering": "no",
--- a/tests/unit/claude-web-utf8-mojibake-13416.test.ts
+++ b/tests/unit/claude-web-utf8-mojibake-13416.test.ts
@@ -1,0 +1,158 @@
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+
+import { createClaudeWebResponse } from "../../open-sse/executors/claude-web/stream.ts";
+
+function byteStream(text: string, chunkSizes: number[] = []): ReadableStream<Uint8Array> {
+  const bytes = new TextEncoder().encode(text);
+  return new ReadableStream<Uint8Array>({
+    start(controller) {
+      let offset = 0;
+      for (const size of chunkSizes) {
+        if (offset >= bytes.length) break;
+        controller.enqueue(bytes.slice(offset, Math.min(offset + size, bytes.length)));
+        offset += size;
+      }
+      if (offset < bytes.length) controller.enqueue(bytes.slice(offset));
+      controller.close();
+    },
+  });
+}
+
+function frames(events: Array<Record<string, unknown>>): string {
+  return events.map((event) => `data: ${JSON.stringify(event)}\n\n`).join("");
+}
+
+function persianTextEvents(text: string): Array<Record<string, unknown>> {
+  return [
+    { type: "message_start", message: { model: "claude-sonnet-5" } },
+    { type: "content_block_start", index: 0, content_block: { type: "text" } },
+    { type: "content_block_delta", index: 0, delta: { type: "text_delta", text } },
+    { type: "content_block_stop", index: 0 },
+    { type: "message_delta", delta: { stop_reason: "end_turn" } },
+    { type: "message_stop" },
+  ];
+}
+
+describe("Claude Web UTF-8 text preservation (#13416)", () => {
+  const persianSamples = [
+    { label: "simple Persian greeting", text: "سلام" },
+    { label: "full Persian sentence", text: "سلام! این پیام یک تست ساده به نظر می‌رسه." },
+    { label: "Arabic text", text: "مرحبا بكم في هذا التطبيق" },
+    {
+      label: "mixed script with emoji",
+      text: "سلام world 你好 🌍",
+    },
+    {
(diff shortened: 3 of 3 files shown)


diegosouzapw:
Thanks for digging into the Arabic/Persian mojibake (#13416). Holding for now.

SSE is UTF-8 by definition, so the charset suffix doesn't change how compliant clients decode the stream. `buildStreamingResponseHeaders` is the shared header for every streaming provider; with this change applied, `tests/integration/chat-pipeline.test.ts` fails 2 cases, both on an exact `Content-Type` comparison, and several unit tests compare `Content-Type` against exactly `text/event-stream`.

The new test doesn't reproduce the reported mojibake either. On the current branch the chunk-boundary case, the only one that checks just the decoded text, already passes. The other cases fail only on the new charset assertion, which runs before their text check. A byte-level reproduction of where the garbling happens (client, proxy, or our decoder at a chunk boundary) would let us fix the actual cause.

diegosouzapw:
Thanks for digging into the Arabic/Persian mojibake report. Independently checking, adding the
charset suffix does break `tests/integration/chat-pipeline.test.ts`'s exact `Content-Type`
comparisons, and the new test's charset assertions check for the header string rather than the
actual decoded text — so it doesn't yet prove the reported bug is fixed. Since SSE is UTF-8 by
spec, a compliant client shouldn't need the charset hint; a byte-level repro (client vs proxy vs
our own chunk-boundary decoding) would help pin down where the garbling actually happens before
we change a header shared by every streaming provider.

diegosouzapw:
Resolvi o conflito e mergeei. Registro do que foi feito e por quê, já que o diagnóstico mudou o enquadramento da PR.

**O conflito** era um só arquivo, `open-sse/handlers/chatCore/responseHeaders.ts`, e puramente textual: esta branch saiu de `152d9510`, antes da #13601 inserir o spread de `DROPPED_UPSTREAM_HEADERS_RESPONSE_HEADER` imediatamente **acima** da linha do `Content-Type`, dentro do mesmo objeto literal. Resolvi mantendo os dois lados — o spread da #13601 e o sufixo `; charset=utf-8` desta PR.

**Sobre o escopo:** o seu último teste é o que mais ajudou aqui. Ele prova que o decoder do OmniRoute **já está correto** — `stream.ts:76` constrói um único `TextDecoder` e usa `decode(value, { stream: true })` com flush no fim, então sequência multi-byte cortada entre chunks reconstrói sem perda. Você respondeu à pergunta do hold com um repro de nível de byte em vez de afirmar que o fix reproduzia o bug, e isso é o comportamento certo.

Por isso ela entra como **endurecimento de interop** para clientes não-conformes (VS Code Chat), não como correção de defeito de servidor — e é consistente com 5 executores que já emitem o header com charset (`uc.ts:55`, `maxai.ts:49`, `codex-app-server.ts:25`, `chatgpt-web-codex.ts:48`, `commandCode.ts:797`).

**A #13416 fica aberta**: o garbling não está no OmniRoute e a causa do lado do cliente ainda não foi identificada.

Validação: 18/18 nos testes da PR, 29/29 em `tests/integration/chat-pipeline.test.ts` (cujas duas asserções de header exatas você já havia atualizado).

PR #13428, opened 2026-09-12
fix(api): stream /api/logs/export with row cap to prevent V8 heap OOM (#13123)

### LINKED ISSUE #13123: fix(backend): GET /api/logs/export buffers whole table → V8 heap OOM → gateway outage
### OmniRoute Version

3.8.50 (docker `diegosouzapw/omniroute:3.8.50-web`)

### Installation Method

Docker (compose)

### Operating System

Linux (Amazon Linux 2023, aarch64)

### Description

`GET /api/logs/export` builds the entire export in memory and crashed the Node process with a V8 heap OOM. The container exited with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`, Docker restarted it (`restart: unless-stopped`), and the gateway was unreachable for ~6.6 minutes. Nothing in the app itself alerted; we only noticed because clients got 502 from the tunnel.

Root cause is visible in `src/app/api/logs/export/route.ts`:

```ts
rows = await exportCallLogsSince(since);          // up to hours=168, no row cap
return new Response(
  JSON.stringify({ logs: rows, count: rows.length, hours, type: logType }, null, 2),
  ...
);
```

- Every matching `call_logs` row is loaded into an array.
- The whole array is serialized in one `JSON.stringify` call **with pretty-printing** (`null, 2`), which roughly doubles the string size.
- That string is then copied into the `Response` body.

With ~9k rows/day and ~36k rows for the last 7 days on our instance, a single 24h+ export comfortably exceeds the default `--max-old-space-size=1024` (Dockerfile:124). The GC log shows the heap jumping ~965MB → ~1818MB in ~12s right before the crash, i.e. a single burst allocation, not a slow leak (the process had been stable for 78h).

### Steps to Reproduce

1. Run OmniRoute with the default Docker image (heap limit 1024MB).
2. Let it accumulate a normal week of traffic (tens of thousands of `call_logs` rows).
3. In the dashboard, export call logs with `hours=24` or larger (or `GET /api/logs/export?hours=168&type=call-logs`).
4. Observe `FATAL ERROR: Reached heap limit` in `docker logs` and a container restart.

### Expected Behavior

Exporting logs should not be able to take the gateway down. The export should be streamed (or paginated / row-capped) so memory stays bounded regardless of table size.

### Actual Behavior

Whole-table buffering → V8 heap OOM → process exit → full gateway outage until the container restarts.

### Evidence

```
[21:0xffff984c0000] 280984215 ms: Incremental Mark-Compact (reduce) 965.4 (1018.7) -> 951.7 (963.0) MB
[21:0xffff984c0000] 280996044 ms: Incremental Mark-Compact (reduce) 1818.4 (1829.6) -> 1816.0 (1827.1) MB
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0x82f70c node::OOMErrorHandler(char const*, v8::OOMDetails const&) [omniroute (v16.3]
```

- `docker inspect`: `RestartCount=1`, `OOMKilled=false` (V8 heap limit, not cgroup OOM).
- Host had ~2GB free; the process was bounded only by `NODE_OPTIONS=--max-old-space-size=1024`.
- `call_logs`: 35,685 rows in the last 7 days, 9,250 in the last 24h.

### Suggested fix

Any one of these keeps memory bounded:

1. Stream the response: iterate rows from SQLite (`.iterate()` / cursor) and write NDJSON or a JSON array chunk-by-chunk via a `ReadableStream`.
2. Drop the pretty-print (`null, 2`) in the export path — it's a download, not for human reading, and it roughly halves peak string size.
3. Hard cap rows per export (e.g. 10k) and return `count`/`truncated` so the UI can page.

Ideally (1) + (2). Happy to test a fix on this instance.

### Workaround

Raised `NODE_OPTIONS=--max-old-space-size=2048` in compose. This only moves the ceiling; a large export can still crash it.

### Related

#11804 reports a combo-related memory spike; this one is a different, deterministic path (log export) but shares the "unbounded in-memory buffering" root cause class.

KooshaPari:
## Problem

`GET /api/logs/export` buffered every matching row into a single `JSON.stringify` call with pretty-printing (`null, 2`), roughly doubling the string size. On tables with tens of thousands of rows this crashed the Node process with a V8 heap OOM (`FATAL ERROR: Reached heap limit Allocation failed`), taking the gateway down for minutes.

The root cause in `src/app/api/logs/export/route.ts`:

```ts
rows = await exportCallLogsSince(since);  // up to hours=168, no row cap
return new Response(
  JSON.stringify({ logs: rows, count: rows.length, hours, type: logType }, null, 2),
  ...
);
```

With ~9k rows/day and ~36k rows for a 7-day export, this comfortably exceeds the default `--max-old-space-size=1024` (Dockerfile:124).

## Fix

Three changes to `src/app/api/logs/export/route.ts`:

### 1. Stream the response
Replace the single `JSON.stringify` + `Response` with a `ReadableStream` that serializes one row at a time. Peak memory is now bounded by the size of one row, not the entire result set.

### 2. Add a configurable row cap
New `limit` query parameter (default 10,000, max 50,000). When the cap fires, the response includes `"capped": true, "limit": <n>, "totalAvailable": <n>` so callers know they received a truncated result.

### 3. Remove pretty-printing
The `null, 2` formatting roughly doubled the string size. Callers that need formatting can pretty-print client-side.

**Backward compatibility:** The response envelope shape `{ count, hours, type, logs: [...] }` is preserved. The cap fields are additive.

## Changes

| File | Change |
|------|--------|
| `src/app/api/logs/export/route.ts` | Stream via ReadableStream, add row cap, remove pretty-print |
| `tests/unit/logs-export-streaming-13123.test.ts` | **New**: 4 tests for envelope shape, cap metadata, uncapped response, 1000-row streaming |

## Tests

```bash
npx tsx --test tests/unit/logs-export-streaming-13123.test.ts
# ✔ streamed response is valid JSON with expected envelope shape
# ✔ capped response includes cap metadata
# ✔ uncapped response omits cap fields
# ✔ streaming produces valid JSON for 1000 rows
```

4/4 pass.

### DIFF
--- a/src/app/api/logs/export/route.ts
+++ b/src/app/api/logs/export/route.ts
@@ -4,18 +4,41 @@ import { exportProxyLogsSince } from "@/lib/db/proxyLogs";
 import { sanitizeErrorMessage } from "@omniroute/open-sse/utils/error";
 
 /**
- * GET /api/logs/export — export logs as JSON
+ * GET /api/logs/export — export logs as JSON (streamed)
  * Query params: ?hours=24 (1, 6, 12, 24; default 24)
  *               &type=call-logs|request-logs|proxy-logs (default call-logs)
+ *               &limit=10000 (max rows; default 10000, max 50000)
+ *
+ * #13123: The previous implementation buffered every matching row into a single
+ * JSON.stringify call with pretty-printing, which roughly doubled the string
+ * size and could OOM the process on large tables. Now:
+ *  - Rows are capped by default (configurable via `limit`).
+ *  - The response is streamed via ReadableStream so only one row is serialized
+ *    at a time, keeping peak memory bounded.
+ *  - Pretty-printing is removed (callers that need formatting can pretty-print
+ *    client-side).
  */
+const MAX_ROWS = 50_000;
+const DEFAULT_ROWS = 10_000;
+
 export async function GET(request: Request) {
   const authError = await requireManagementAuth(request);
   if (authError) return authError;
 
   try {
     const { searchParams } = new URL(request.url);
-    const hours = Math.min(Math.max(parseInt(searchParams.get("hours") || "24") || 24, 1), 168);
+    const hours = Math.min(
+      Math.max(parseInt(searchParams.get("hours") || "24") || 24, 1),
+      168
+    );
     const logType = searchParams.get("type") || "call-logs";
+    const limit = Math.min(
+      Math.max(
+        parseInt(searchParams.get("limit") || String(DEFAULT_ROWS)) || DEFAULT_ROWS,
+        1
+      ),
+      MAX_ROWS
+    );
 
     const since = new Date(Date.now() - hours * 3600 * 1000).toISOString();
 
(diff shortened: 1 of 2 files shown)


diegosouzapw:
Thanks for #13123. Holding for a revision.

1. `exportCallLogsSince()` / `exportProxyLogsSince()` still materialize every matching row before streaming starts, so the peak V8 heap is essentially unchanged. Only the pretty-printed string copy goes away.
2. Defaulting `limit` to 10,000 silently truncates exports that previously returned everything. The `capped` flag only appears at the end of the body.
3. The test re-implements the stream builder locally instead of calling the route, so it can't catch a regression in the route.

A version that pages rows out of SQLite inside the `ReadableStream` (LIMIT/OFFSET or an iterator), keeps the full export by default, and tests the real `GET` handler would be a clear win.

diegosouzapw:
Thanks for #13123 — the OOM diagnosis is right, but tracing the actual data flow shows
`exportCallLogsSince()`/`exportProxyLogsSince()` still materialize every matching row in one
shot before the `ReadableStream` starts, so the peak V8 heap is essentially unchanged; only the
pretty-printed string copy goes away. The new test also re-implements the stream builder locally
instead of calling the route, so it wouldn't catch a regression there. Could you push the
pagination/streaming down into the DB layer (a cursor or LIMIT/OFFSET loop in
`src/lib/db/callLogs.ts`/`proxyLogs.ts`) so the row set itself is never fully buffered, rewrite
the test to exercise the real route handler, and add a changelog note for the new default
`limit` truncating previously-unbounded exports?

PR #13434, opened 2026-09-12
fix(models): reconcile provider dashboards with active live catalogs

JxnLexn:
## Summary

The provider dashboard can advertise and test built-in models that request routing already rejects as absent from the active live catalog. For example, NVIDIA showed 116 models while its active catalog contained 82. `Import from /models` reported that all models were already imported and returned without refreshing the page state, so retired rows remained visible.

This PR reconciles the dashboard with the existing server-side live-catalog policy across providers, rather than adding a NVIDIA-specific filter.

## Changes by commit

1. **Shared provider listing and refresh:** expose the active catalog's `authoritative` decision through `/api/synced-available-models?provider=...`; use active connections and their union instead of all stored connection catalogs; merge operator custom models separately. For authoritative catalogs, omit stale built-in/legacy-import rows while retaining metadata for live IDs and preserving manual overrides. Refresh the listing after discovery even when zero new IDs are imported. Retain the last confirmed client snapshot on failed refresh requests and scope snapshots to the provider.
2. **OpenRouter, compatible, and passthrough views:** apply the same membership rule to their separate row builders so static fallbacks and alias-only rows cannot reintroduce unavailable models into the list or Test All.

## Preservation and limits

- Reuses the existing `liveCatalogAuthoritative` policy, including explicit opt-outs for intentionally partial discovery. No new provider-discovery endpoints or inference restrictions.
- Missing/unconfirmed/empty catalogs retain the existing static fallback behavior; this does not introduce a new authoritative-empty persistence contract.
- Failed upstream discovery does not delete the last persisted catalog. Failed dashboard refresh requests keep the last confirmed client snapshot.
- Operator-created models and same-ID metadata overrides remain. Saved aliases and built-in registry definitions are not deleted; unavailable alias-only rows are excluded from active listings.
- The runtime catalog getter retains its default custom-model behavior. A new optional discovery-only mode supports the dashboard, which fetches custom rows separately.
- Cursor's virtual auto entries retain their special handling. Curated-only providers keep their existing behavior.

## Validation

- [x] Change type: provider dashboard / model catalog reconciliation
- [x] `npm run typecheck:core` on release and isolated fork implementations
- [x] Targeted ESLint with repository suppressions; Prettier; `git diff --check`
- [x] Reconciled with current `release/v3.8.51`
- [x] Automated regression tests added
- [x] Browser smoke with the actual `useProviderModels`, `useModelImportHandlers`, and merge helper in an isolated fixture: retired built-ins/imports disappear; an import with only removals updates immediately despite the no-new-models result; an HTTP 503 refresh preserves the confirmed list; changing provider does not reuse the previous provider's catalog
- [x] Production catalog snapshots from 14 providers checked through the new projection; NVIDIA reproduces the reported 116 → 82 discrepancy and excludes every rejected model shown in the report
- [x] [AMD64 standalone image build](https://github.com/JxnLexn/OmniRoute/actions/runs/34690880779) and isolated fork deployment completed
- [x] Deployed NVIDIA sync returned 200; the dashboard API matched the active connection catalogs for all 14 providers with stored live data, including NVIDIA (82), OpenRouter (22), Bedrock (84), and both Token Plans (9 each)

Unit suites and the broad matrix were left to CI. No inference requests were required for verification, and PR checks were not awaited for the deployment checks.

## Tests Added Or Updated

- `tests/unit/authoritative-provider-listing.test.ts`: retired static/imported rows, manual overlays, partial/unconfirmed catalogs, active-union changes, explicit empty input, Cursor fallback, and stale alias/fallback rows in compatible/passthrough views.
- `tests/unit/active-catalog-display-snapshot.test.ts`: active-connection union, inactive-connection exclusion, discovery-only custom/import separation, and custom-only state not establishing authoritative discovery.

## Coverage Notes

The tests cover the shared projection and database snapshot boundary. The browser smoke covers the real client hooks and the no-new-import refresh path. No local coverage measurement was run.

## Reviewer Notes

No migrations or dependencies are added. The patch is split into a shared-catalog commit and a separate commit for the OpenRouter/compatible/passthrough row builders. The PR contains only these changes, not the fork's admission, Vertex, Token Plan discovery, or Docker workflow hotfixes.

### DIFF
--- a/src/app/(dashboard)/dashboard/providers/[id]/ProviderDetailPageClient.tsx
+++ b/src/app/(dashboard)/dashboard/providers/[id]/ProviderDetailPageClient.tsx
@@ -772,6 +772,7 @@ export default function ProviderDetailPageClient() {
             modelMeta={modelMeta}
             modelAliases={modelAliases}
             syncedAvailableModels={syncedAvailableModels}
+            syncedCatalogAuthoritative={syncedCatalogAuthoritative}
             compatibleFallbackModels={compatibleFallbackModels}
             copied={copied}
             onCopy={copy}
--- a/src/app/(dashboard)/dashboard/providers/[id]/ProviderDetailPageClient.tsx
+++ b/src/app/(dashboard)/dashboard/providers/[id]/ProviderDetailPageClient.tsx
@@ -192,6 +192,7 @@ export default function ProviderDetailPageClient() {
   const {
     modelMeta,
     syncedAvailableModels,
+    syncedCatalogAuthoritative,
     modelAliases,
     fetchProviderModelMeta,
     fetchAliases,
@@ -286,14 +287,14 @@ export default function ProviderDetailPageClient() {
     NOAUTH_PROVIDERS[providerId]?.noAuth === true ||
     getProviderById(providerId)?.managedAccount === true;
   const registryModels = getModelsByProviderId(providerId);
-  // Prefer synced API-discovered models when available, then merge built-ins
-  // and user-managed custom models without duplicating IDs. Cursor exclusive
-  // listing drops the static registry entirely when synced is non-empty.
+  // Use the server's active-catalog authority decision for display and Test All.
+  // Registry entries supply metadata/fallback; operator custom models remain.
   const models = useMemo(() => {
     return mergeProviderModelListing({
       providerId,
       registryModels,
       syncedModels: syncedAvailableModels,
+      syncedCatalogAuthoritative,
       customModels: (modelMeta.customModels || []).map((cm) => ({
         ...cm,
         id: cm.id,
@@ -307,6 +308,7 @@ export default function ProviderDetailPageClient() {
     registryModels,
     syncedAvailableModels,
     modelMeta.customModels,
+    syncedCatalogAuthoritative,
     usesCuratedModelsOnly,
   ]);
   const isUpstreamProxyProvider = providerInfo?.category === "upstream-proxy";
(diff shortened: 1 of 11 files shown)


diegosouzapw:
This is the right generalization — fixing the NVIDIA 116→82 discrepancy by extending the
existing `liveCatalogAuthoritative` policy to the dashboard, instead of a one-off NVIDIA
filter. Ran the two new test files: 8/8 pass, including the "explicit non-authoritative
state overrides Cursor's legacy exclusivity" regression case. One thing missing before
merge: no `changelog.d/fixes/` fragment. Also, since `mergeProviderModelListing.ts` is
shared by every provider's dashboard listing (not just NVIDIA), it'd be worth a second pair
of eyes on that file specifically given the blast radius — not a blocker, just a suggestion
given how central it is.

JxnLexn:
Thanks — added `changelog.d/fixes/13434-authoritative-provider-listing.md` and performed a focused follow-up review of the shared merge helper in f782bf26e.

I added regressions covering the same authoritative-membership/metadata-overlay behavior across NVIDIA, OpenAI, OpenRouter, Anthropic and Vertex; input non-mutation; curated-only providers ignoring the live flag; and preservation of Cursor's legacy synthetic auto row. Existing coverage still checks empty/partial catalogs, manual custom rows, stale imports and passthrough aliases. This is a focused follow-up by the author, not a claim of independent reviewer approval.

`node --import tsx/esm --test tests/unit/authoritative-provider-listing.test.ts tests/unit/active-catalog-display-snapshot.test.ts tests/unit/model-listing-capability-5420.test.ts`: **18/18 passed**. Changelog integrity passes. No further production-code change was needed in the merge helper.

PR #13264, opened 2026-09-10
fix(oauth): align codebuddy-cn OAuth User-Agent with chat/usage (#12702)

diegosouzapw:
Closes #12702

## Root cause

Reporter (third-party "deepseek harness" client → codebuddy-cn provider) got a hard 400 from
Tencent's `copilot.tencent.com` gateway: `code 11128 "request illegal"`, `displayMsg: "The
request was blocked by security policy."` — CodeBuddy's own WAF, not the content-sensitivity
rejection OmniRoute already retries around (#6535 / #9723).

Static analysis found a genuine, reproducible defect in OmniRoute's own codebuddy-cn
integration: the OAuth device-code auth/poll and token-refresh calls presented
`CLI/2.63.2 CodeBuddy/2.63.2` (`src/lib/oauth/constants/oauth.ts`), while chat-completion
requests and usage/quota requests presented `CLI/2.108.1 CodeBuddy/2.108.1` — the same
logical account/token advertising two CLI identities 45 minor versions apart across its own
auth vs. chat calls. That is exactly the kind of internally-inconsistent client fingerprint an
anti-abuse WAF flags as anomalous, and matches the reporter's own suspicion ("Codebuddy CN has
strengthened its security measures"). The mismatch has existed unchanged since the provider was
added (commit a7ae9550, 2026-06-25).

This is a plausible (not live-verified) root cause for the reporter's specific 400 — CodeBuddy's
WAF internals are opaque — but it is an independently real defect worth fixing regardless.

## Fix

- Centralized the version string into `CODEBUDDY_CN_USER_AGENT` (exported from
  `src/lib/oauth/constants/oauth.ts`, set to the chat/usage value `CLI/2.108.1
  CodeBuddy/2.108.1`).
- `CODEBUDDY_CN_CONFIG.userAgent` now reads from that constant, so OAuth device-code
  auth/poll (`src/lib/oauth/providers/codebuddy-cn.ts`) and token refresh
  (`open-sse/services/tokenRefresh/providers/codebuddyCn.ts`, which already dynamically
  imports `CODEBUDDY_CN_CONFIG`) pick it up automatically.
- The chat registry entry (`open-sse/config/providers/registry/codebuddy-cn/index.ts`) and the
  usage/quota handler (`open-sse/services/usage/codebuddy-cn.ts`) now import and use the same
  constant instead of a duplicated literal, so auth/chat/usage can never silently drift apart
  again.

## Regression test

[...]

PR #13436, opened 2026-09-12
fix(build): stop the client bundle from reaching server-only modules, and make the guard find them

maxmad64bis:
⚠️ base-red inherited: #12732

## Summary

The production build is broken on the current tip: two dashboard pages drag server-only code into the browser bundle, and webpack stops on Node builtins (`child_process`, `fs`, `module`, then `UnhandledSchemeError` on `node:fs` / `node:os` / `node:path`). This PR cuts both edges and teaches the client-bundle guard to catch the next one itself.

Change type: build-deploy.

Chain 1: the combo control-center page needs one pure lookup (alias → provider) that lived inside the full server-side model resolver, so the browser had to swallow the resolver and everything behind it. The lookup moves verbatim into a leaf module importing only its static map; the old module re-exports it, so existing callers see no change. Chain 2: a CLI tools page reads a version string through a registry entry → OAuth constants → a CLI detection module starting with `node:fs` / `node:os` / `node:path` imports. The string moves into an import-free pin both sides read from.

The guard is the real fix. It already walks the true import graph from every `"use client"` file, but only flags six hardcoded modules — neither culprit is listed, so it's green on a tip that doesn't compile. It now also flags any reached file importing a Node builtin the browser build refuses (via `node:module`, no new list to rot; `crypto` excluded, webpack polyfills it). Floor, entries, and both exclusions (`import type`, dynamic `import()`) stay.

## Related Issues

- #12732 — branch not green; the `package-artifact` failure there is consistent with this (not proven identical).
- #10692 / #10695 / #10700 — the original incident, its fix, and the guard's widening to every client entry.
- #13283 / #13264 — the two commits that added these edges.

## Validation

- [x] `npm run lint`

## Tests Added Or Updated

The widened guard is the test: copied onto the bare tip it fails naming both entries, and passes on the fixed tree. Suites green: guard 1/1, model-resolver 25/25, cursor version + PKCE 26/26. Full `next build` passes. Ratchets, both typechecks, and eslint clean.

## Coverage Notes

Moves only, no behavior change: same map derivation, same pin string, re-exports keep dependents resolving. The guard's new predicate runs inside the existing test over the graph it already walks — proven able to fail by the bare-tip control.

## Reviewer Notes

Two flags: a future client-side `buffer`/`process`-style import webpack tolerates would turn the guard red — extend the commented exclusion then, don't re-hardcode. The chain-1 trace also passes through the registry path downstream (both chains merge there, expected). The two cuts could ship without their test if split — kept together since the template requires the test in the same PR.

### DIFF
--- a/changelog.d/fixes/13436-client-bundle-server-only-guard.md
+++ b/changelog.d/fixes/13436-client-bundle-server-only-guard.md
@@ -1,0 +1,1 @@
+- **fix(build):** the production build no longer breaks when a client component reaches a server-only module, and the client-bundle guard now discovers server-only modules instead of matching a fixed list ([#13436](https://github.com/diegosouzapw/OmniRoute/pull/13436)) — thanks @maxmad64bis
--- a/open-sse/services/model.ts
+++ b/open-sse/services/model.ts
@@ -1,6 +1,9 @@
 import { PROVIDER_ID_TO_ALIAS, PROVIDER_MODELS } from "../config/providerModels.ts";
+import { ALIAS_TO_PROVIDER_ID, resolveProviderAlias } from "./providerAlias.ts";
 import { resolveWildcardAlias } from "./wildcardRouter.ts";
 import { getRegisteredProviderEffortBaseModelId } from "../utils/registeredEffortVariants.ts";
+
+export { resolveProviderAlias };
 
 type ProviderModelAliasMap = Record<string, Record<string, string>>;
 type ModelAliasValue = string | { provider?: string; model?: string };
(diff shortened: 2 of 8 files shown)


diegosouzapw:
Measured result for the release-green build red. I ran `npm run check:pack-artifact` on an idle box against `release/v3.8.51` `f7dbfc88`:

- tip alone: `Turbopack build failed with 168 errors`, all from the `ComboControlCenterClient.tsx → controlCenter.ts → open-sse/services/model.ts` chain.
- tip + #13635 + **this PR**: compiles in 8.1 min, pack policy passes, and the next stop is the provenance stamp, which the validator never writes (fixed in #13678).
- tip + #13565 alone: compiles too, but fails pack policy until #13635's `opencode-plugin-v2` allowance lands.

This PR and #13565 fix the same edge in different ways and conflict on `src/lib/combos/controlCenter.ts`, so only one of them can land as-is. #13565 also conflicts with #13635 on `src/lib/oauth/constants/oauth.ts`. This one merges cleanly after #13635. That pairing is the combination I measured green, but the choice is the maintainer's.

PR #13445, opened 2026-09-12
fix(network): skip Chrome TLS impersonation for Groq

### LINKED ISSUE #13225: [BUG] Groq Cloudflare 1010 only through OmniRoute transport — direct curl/native undici work from same host
### OmniRoute Version

e.g.3.7.9

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

_No response_

### Node.js Version

_No response_

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

## Problem

Groq requests routed through OmniRoute consistently receive:

HTTP 403
Cloudflare Error 1010
browser_signature_banned
zone: api.groq.com

This occurs across 3 different Groq accounts configured in the same
OmniRoute installation.

## Control tests

From the exact same Linux host and public egress IP:

1. curl -> GET https://api.groq.com/openai/v1/models
   without authentication

   Result:
   HTTP 401
   normal Groq authentication response
   cf-ray present
   NO Cloudflare 1010

2. Node.js native fetch / undici
   same endpoint, without authentication

   Result:
   HTTP 401
   cf-ray present
   NO Cloudflare 1010

Therefore the host/IP itself is not globally blocked by Groq/Cloudflare.

## OmniRoute investigation

Read-only inspection found:

- no proxy configured
- GROQ_USER_AGENT is not configured
- native undici works correctly
- OmniRoute direct traffic uses its own undici.Agent pool
- connections: 1
- pipelining: 0
- custom keep-alive / connection handling
- no intentional browser impersonation was identified

The custom OmniRoute transport is the remaining significant difference
between the working controls and the blocked requests.

Existing Cloudflare Ray IDs from blocked requests:

- a38df1408ca46d53
- a38df12c0...

## Groq Support confirmation

Groq Support reviewed the case and confirmed:

"The HTTP 403 with Cloudflare Error 1010 means the request is being blocked
by the WAF before it reaches the Groq API.

Since direct requests from the same host work and the 403 only occurs
through OmniRoute, this points to OmniRoute's HTTP transport or client
fingerprint rather than your Groq API key or account."

Groq specifically recommended checking OmniRoute's HTTP transport,
headers, TLS, and other transport characteristics.

## Question

Is there a known issue with OmniRoute's custom HTTP/TLS transport against
api.groq.com that can trigger Cloudflare Error 1010?

Is there a supported configuration or transport fix that makes Groq
requests use a standard HTTP client behavior comparable to native
fetch/undici?

We are NOT requesting browser impersonation, stealth, or a Cloudflare
bypass.

No API keys or secrets are included in this report.

### Steps to Reproduce

.

### Expected Behavior

.

### Actual Behavior

.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

HouMinXi:
## Why

Groq returns Cloudflare 1010 (`browser_signature_banned`) only through OmniRoute. Same host, same egress IP: curl and native undici get a normal 401. Groq Support said the block is the HTTP/TLS client, not the key.

Direct TLS impersonation (`ENABLE_TLS_FINGERPRINT=true`, Chrome-124 wreq) applies to every provider when `TLS_FINGERPRINT_PROVIDERS` is unset. Groq is in that set. That JA3 is what Cloudflare 1010s.

## Change

`tlsFingerprintProviderAllowed` returns false for provider `groq` and for `api.groq.com` / `*.groq.com`. That includes the empty tracking-store path, which otherwise spoofs every direct host.

Other providers keep the legacy direct-TLS default. Proxied TLS is unchanged (still needs an explicit allowlist). Groq is a hard skip; listing it in `TLS_FINGERPRINT_PROVIDERS` does not turn impersonation back on.

Fixes https://github.com/diegosouzapw/OmniRoute/issues/13225

## Tests

`tests/unit/tls-proxy-context.test.ts`: Groq host and `groq` provider stay on undici; an openai host still spoofs when the allowlist is unset.


diegosouzapw:
Excellent root-cause work here — confirming via Groq support that it's the TLS/HTTP client
fingerprint (not the key) being blocked, and hard-skipping Groq (including the empty
tracking-store edge case) rather than relying on the allowlist alone, is the right call. Ran
your test file: 31/31 pass, including the new Groq-skip cases and the full pre-existing
suite with no regression. Changelog fragment and docs update are already in place — this
looks merge-ready.

Issue #5552, opened 2026-06-30
[BUG] Subscription-aware cost tracking: show /usr/bin/bash for subscription accounts


SeaXen:
### Feature Description

OmniRoute calculates cost based on per-token pricing for all providers, including accounts that use a subscription (flat-rate) model. For example, Codex with a subscription, MiniMax Coding Plan, Kimi Coding Plan, etc. — these have $0 per-token cost, but OmniRoute shows estimated cost in analytics.

### Use Case

Users with subscription-based accounts (Codex, MiniMax Coding Plan, Kimi Coding Plan, MiMo Token Plan, etc.) see inflated cost estimates in analytics that don't reflect their actual billing.

### Proposed Solution

Add a "subscription" or "flat-rate" flag to provider accounts that:
- Shows $0 cost in analytics for subscription accounts
- Or at minimum, clearly distinguishes between "estimated API cost" vs "actual subscription cost"
- Optionally allow users to set the subscription price so ROI tracking is accurate

### Alternatives Considered

Ignoring the cost column for subscription accounts — but this defeats the purpose of having cost tracking at all.

### Additional Context

This affects all subscription-based providers: Codex, MiniMax Coding Plan, Kimi Coding Plan, MiMo Token Plan, BytePlus ModelArk, etc. The current per-token cost calculation is correct for API-key accounts but misleading for subscriptions.

SeaXen:
Update: this also affects web cookie providers.

ChatGPT Web, grok-web, and other cookie-based providers are backed by a subscription (ChatGPT Plus/Pro, SuperGrok). These should show $0 cost since there's no per-token billing — but OmniRoute still calculates and displays estimated per-token cost.

Same issue applies to:
- Codex subscription accounts
- ChatGPT Web (cookie-based)
- grok-web (cookie-based)
- Any subscription-backed provider

diegosouzapw:
Implemented in `release/v3.8.43` (already merged into the active development branch — feel free to pull and test). Flat-rate providers now show **$0** in cost analytics instead of an inflated per-token estimate:

- All **cookie-web** providers (ChatGPT Web, grok-web, Gemini Web, etc.) — backed by a consumer subscription, no per-token billing.
- The dedicated **coding-plan** providers: **Minimax Coding**, **Kimi Coding**, **GLM Coding**, **Alibaba Coding Plan**, **Xiaomi MiMo**.

[...]

PR #13446, opened 2026-09-12
fix(arena+analytics): atomic ELO sync (fetch-first) + flatRateAsZero in compression writer

CrashCartCapital:
## Summary

Two related fixes for cost/intelligence bookkeeping:

**1. Arena ELO sync data-loss bug (primary).** `syncArenaElo` ran `deleteExpiredIntelligence()` *before* fetching leaderboards. If the Arena API was unreachable (429s/timeouts), each sync attempt deleted expired rows and upserted nothing — and since `initArenaEloSync` runs at boot, every service restart re-drained the table. On a rate-limited upstream this drains `model_intelligence` to 0 permanently.

Redesign:
- **Fetch first**: no DB mutation before a successful fetch+transform.
- **Atomic refresh**: new `applyArenaEloRefresh()` wraps upsert + membership prune in a single `db.transaction()` — entries not in the refreshed set are removed only after the new set is durably written.
- **Freshness guard**: startup sync skipped when the table is non-empty and latest `synced_at` is within the sync interval.
- **Persistent failure backoff**: failed syncs persist `lastFailedAt` (key_value `arena_elo`); subsequent syncs (incl. across restarts) skip for a backoff window instead of hammering a rate-limited upstream.

`deleteExpiredIntelligence()` is no longer called anywhere; expiry-based cleanup is subsumed by membership prune inside the transaction.

**2. Compression writer phantom savings.** `writeCompressionAnalytics` called `calculateCost` without `flatRateAsZero: true`, so flat-rate subscription lanes (minimax, glm, kimi, bailian, xiaomi, web-cookie) booked non-zero dollar "savings" that were never payable. Adds the same opt-in already used by `/api/usage/analytics` (#5552).

## Test plan

- `tests/unit/arena-elo-sync-redesign.test.ts` (new, 5 tests): failing fetch never deletes + persists backoff; atomic replace w/ prune; freshness skip; backoff-wins-on-empty (the exact production boot-loop condition); backoff skips second sync.
- `tests/unit/compression/compressionAnalyticsWriterFlatRate.test.ts` (new, 2 tests): minimax books $0/null; metered provider keeps real estimate.
- Existing suites green on this branch: `arena-elo-sync.test.ts` 56/56.

Verified in production on a local deployment (v3.8.48 lineage): table recovered 0→236 entries after deploy; backoff dormant-but-deployed.

### DIFF
--- a/open-sse/handlers/chatCore/compressionAnalyticsWrite.ts
+++ b/open-sse/handlers/chatCore/compressionAnalyticsWrite.ts
@@ -149,7 +149,10 @@ export function writeCompressionAnalytics(
           opts.provider ?? "",
           opts.effectiveModel ?? "",
           { input: tokensSaved },
-          { serviceTier: opts.effectiveServiceTier }
+          // Flat-rate (subscription / cookie-web) lanes don't bill per token, so
+          // their per-token pricing rows must not book dollar "savings" here
+          // (same opt-in as /api/usage/analytics, #5552).
+          { serviceTier: opts.effectiveServiceTier, flatRateAsZero: true }
         );
       } catch (err) {
         opts.log?.debug?.(
--- a/src/lib/arenaEloSync.ts
+++ b/src/lib/arenaEloSync.ts
@@ -14,11 +14,12 @@ import { resolveScoresAs } from "@omniroute/open-sse/services/autoCombo/scoresAs
 
 import { isArenaEloSyncEnabled } from "@/shared/utils/featureFlags";
 
+import { getDbInstance } from "./db/core";
 import { backupDbFile } from "./db/backup";
 import {
-  bulkUpsertModelIntelligence,
-  deleteExpiredIntelligence,
+  applyArenaEloRefresh,
   deleteModelIntelligenceBySource,
+  getLatestSyncedAt,
   type ModelIntelligenceEntry,
 } from "./db/modelIntelligence";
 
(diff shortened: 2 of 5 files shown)


diegosouzapw:
Both fixes check out — the fetch-first + atomic transaction redesign for Arena ELO sync is a real
fix for a real boot-loop data-loss bug (confirmed via the redesign test reproducing the exact
production condition), and the `flatRateAsZero` addition to the compression writer correctly
reuses the same opt-in already used by `/api/usage/analytics`. All 7 new tests pass on this
branch. Two small asks: could you split this into two PRs (arena ELO sync vs. the compression
writer flat-rate fix) so each can be reviewed/reverted independently if needed, and add
`changelog.d/fixes/` fragments for both (currently missing)?

diegosouzapw:
Thanks @CrashCartCapital — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #13449, opened 2026-09-12
fix(memory): merge memory context into top-level system field for claude-shaped bodies

### LINKED ISSUE #13425: fix(claude): memory injection unshifts system message at messages[0] despite top-level system field (400)
# Memory-injection 400: `injectSystemFirst` prepends a system message at `messages[0]` even when the body carries a top-level `system` field

## Summary
Claude-family requests (direct Anthropic, claude/anthropic-compatible, Claude-Code passthrough) hit
`400 messages.0: use the top-level 'system' parameter for the initial system prompt; the
directive-only form (content: [] with output_config) is accepted at any position` when memory
injection is enabled and the body has a top-level `system` field.

## Root cause
`src/lib/memory/injection.ts` → `injectSystemFirst()` (called for Claude-family and strict-system
providers):

```
161  function injectSystemFirst(request, messages, memoryText, count): ChatRequest {
167    const first = messages[0];
168    if (first && first.role === "system") {
170      const merged = { ...first, content: `${memoryText}\n${first.content}` };
171      return { ...request, messages: [merged, ...messages.slice(1)] };
172    }
173    const memorySystemMessage = { role: "system", content: memoryText };
174    return { ...request, messages: [memorySystemMessage, ...messages] };
175  }
```

It only inspects `messages[0]`. For an Anthropic-shaped body — system prompt in the **top-level
`system` field** and `messages[0]` a `user`/`assistant` turn — branch 1 is false, so it unshifts a
brand-new `{role:"system"}` at `messages[0]`. Anthropic rejects a system role at index 0 (system
content must ride the top-level `system` param; only the empty directive-only form is allowed
inline). The top-level `system` field is never consulted.

This is the same 400 class as #12584 but in the **memory-injection** code path, not the
output-styles path.

## Live evidence (our gateway, claude-opus-5, 2026-09-11)
```
msgs[0]:  system | "Memory context: durably captured its result"
top-level system present: True
error: [400]: messages.0: use the top-level 'system' parameter for the initial system prompt…
```

## Fix direction
`injectSystemFirst` (and the shared placement in `injectMemory`/`placeMessage` for the
system-role strategy) must merge the memory text into the **top-level `system` field** when one
exists (string or block-array), falling back to a `messages` merge/prepend only when there is no
`system` field. Do not remove or gate the memory feature — placement-only. When merging into a
top-level string `system`, prepend the memory text; when it is a block array, unshift a
`{type:"text", text: ...}` block. Never leave a `role:"system"` message at `messages[0]` on a
body that carries a `system` field.

Follow the convention already established in #12584's fix (`placeSystemInstruction` /
`systemFieldIncludesMarker` in `open-sse/services/compression/outputMode.ts`), and the #6135 /
#11290 / #1701 / #3890 defensive branches already in this file — the new branch must not regress
the cache-safe (#3890) or strict-system-first (#6135) behavior for providers in those sets.

## Acceptance
- New unit test: an Anthropic-shaped body (string `system` + user `messages[0]`) with memory
  injection → memory text merged into top-level `system`, NO system message at `messages[0]`.
- New unit test: block-array `system` → memory appended as a text block, `messages[0]` untouched.
- Existing `memory-cache-safe-injection`, `memory-glm-injection`, `memory-system-first-6135`,
  `chatcore-memory-skills-injection` suites still pass.
- A real claude-opus-5 request with memory on returns 200 (regression-proof needing a live run,
  may be noted as follow-up if no creds in CI).

Xore:
## Summary
- Memory injection now merges into the top-level `system` field (string or block-array) instead of unshifting a synthetic `system` role message at `messages[0]`.
- Fixes the Anthropic 400 error hit when a claude-shaped request body already carries a top-level `system` field alongside injected memory context.
- Placement fix only — memory feature behavior/flags unchanged, still enabled by default per existing config.

## Test plan
- [x] Regression test pins single injection when a leading system message and top-level system field coexist
- [x] `npm run test:unit` / targeted test file for memory injection path

Fixes #13425

⚠️ base-red inherited: #12732

### DIFF
--- a/changelog.d/fixes/memory-top-level-system.md
+++ b/changelog.d/fixes/memory-top-level-system.md
@@ -1,0 +1,1 @@
+- fix(memory): merge injected memory context into the top-level `system` field instead of unshifting a synthetic `role: "system"` message at `messages[0]`, fixing a 400 on Anthropic-shaped (claude-family / strict-provider) request bodies (#13425)
--- a/src/lib/memory/injection.ts
+++ b/src/lib/memory/injection.ts
@@ -28,7 +28,7 @@ export interface ChatMessage {
 export interface ChatRequest {
   model: string;
   messages: ChatMessage[];
-  system?: string;
+  system?: string | unknown[];
   temperature?: number;
   max_tokens?: number;
   stream?: boolean;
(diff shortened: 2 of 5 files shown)


saadi12ana (approved):

diegosouzapw:
The root-cause diagnosis and fix here are correct, but an equivalent (and slightly broader — it
covers both the system-first and cache-safe/general paths in one guard) fix for the same issue
(#13425) already merged a few hours ago via #13427, with its own dedicated regression test file.
Closing this one as a duplicate rather than merging a second implementation — thanks for catching
and fixing #13425 independently.

PR #13463, opened 2026-09-12
fix(gemini): send thinkingLevel for 3.8 Flash so thoughts stop eating the output cap

HouMinXi:
## Status

Gemini 3.8 Flash on Antigravity/AGY is already in the live catalog (`agy/gemini-3.8-flash-high`). Omni still translates thinking the Gemini 3.7 way: a numeric `thinkingBudget` plus `includeThoughts: true` by default.

## Problem

A review-sized request (`max_tokens=65536`, combo `onmi-gemini3.6` → `agy/gemini-3.8-flash-high`) came back `finish_reason=length` with ~2.6k visible tokens. Call-log outbound `generationConfig` was:

```
maxOutputTokens: 65536
thinkingConfig: { thinkingBudget: 24576, includeThoughts: true }
```

Usage on that 200: `prompt_tokens=29601`, `completion_tokens=2594`, `total_tokens=95109`. Hidden thoughts consumed essentially the whole output budget. Official Gemini 3.8 Flash docs use `thinking_level` (`low`/`medium`/`high`) and a 65536 output cap. `#12672` also starts sending `thinkingLevel`; this change is only the translator shape, not catalog/quota work.

## Change

- For `gemini-3.8*` (including `agy/` / `antigravity/` prefixes), emit `thinkingLevel` instead of a numeric `thinkingBudget`.
- Map existing `reasoning_effort` / `thinking.budget_tokens` / `output_config.effort` onto `low|medium|high` using each model's `thinkingBudgetCap` / `defaultThinkingBudget`.
- Omit `includeThoughts` unless the client asked (`includeThoughts` / `include_thoughts`).
- Leave `maxOutputTokens` at 65536. Do not shrink it.
- `budget <= 0` stays an explicit off-switch (`thinkingBudget: 0`, `includeThoughts: false`), not `thinkingLevel: "low"`.

Gemini 2.5 and other families are unchanged (still numeric `thinkingBudget`).

## Verification

- `node --import tsx/esm --test tests/unit/gemini-38-thinking-level-output.test.ts`: 18/18.
- Direct helper test: `budget <= 0` throws; injecting `return "low"` turns that test red; restoring the throw turns it green.
- Envelope path (`openaiToAntigravityRequest`) still has `maxOutputTokens=65536` and no default `includeThoughts`.

## Conclusion

Translator-only scope. This PR is submitted for review and merge. Overlaps `#12672` on `thinkingLevel` in the same two Gemini translators; that PR also carries catalog/quota/UI. This one is the request-shape fix for 3.8 Flash eating the output cap.

### DIFF
--- a/changelog.d/fixes/gemini-38-think-level.md
+++ b/changelog.d/fixes/gemini-38-think-level.md
@@ -1,0 +1,1 @@
+- **fix(gemini):** send Gemini 3.8 `thinkingLevel` instead of a numeric `thinkingBudget`, and omit `includeThoughts` unless the client asked, so hidden thoughts stop eating `maxOutputTokens`
--- a/open-sse/services/thinkingBudget.ts
+++ b/open-sse/services/thinkingBudget.ts
@@ -62,6 +62,58 @@ export const THINKING_LEVEL_MAP: Record<string, number> = {
   xhigh: 131072, // T11: explicit xhigh alias
 };
 
+export type Gemini38ThinkingLevel = "low" | "medium" | "high";
+
+/** Gemini 3.8 Flash (and prefixed ids like agy/gemini-3.8-flash-high). */
+export function isGemini38Model(model: string): boolean {
+  return /(?:^|[\/])gemini-3\.8(?:$|-)/i.test(model);
+}
+
+export function gemini38ThinkingLevelFromBudget(
+  model: string,
+  budget: number
+): Gemini38ThinkingLevel {
+  const resolved = getResolvedModelCapabilities(model);
+  const cap = resolved.thinkingBudgetCap ?? 24576;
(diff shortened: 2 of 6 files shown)


diegosouzapw:
Ran your 18 new tests at the PR head — all pass, and the fix matches the reported production
failure shape well (3.8 Flash thoughts eating the whole output cap). One thing to flag: #12826
(also open) edits the exact same three `thinkingConfig` assignment blocks in
`openai-to-gemini.ts`, adding a complementary but different mechanism (an explicit
client/payload-rule-supplied `thinkingLevel` that should take priority over your
budget-derived one). Whichever of the two merges second will need a small rebase to combine
both conditions in the same block — happy to coordinate on the exact precedence once the
owner picks a merge order.

PR #13466, opened 2026-09-12
fix(oauth): classify an embedded invalid_grant in a refresh error body

fouadSalkini:
## Problem

Cline's refresh endpoint answers a dead `refresh_token` with a **non-standard** OAuth error body — the code arrives as the tail of a sentence rather than as a bare code:

```
HTTP 400 {"data":"","error":"failed to refresh token: invalid_grant","success":false}
```

`extractOAuthErrorCode` cannot classify that. Walking it against the real body:

| step | result |
|---|---|
| object path reads `o.error` | `"failed to refresh token: invalid_grant"` |
| `UNRECOVERABLE_OAUTH_ERROR_CODES.has(...)` | `false` — the Set is exact-match |
| nested-JSON branch | skipped, value doesn't start with `{` `[` `"` |
| `/"error(?:_code)?"\s*:\s*"([a-z_]+)"/` scan | no match — we're already inside the string *value* |
| → | **`null`** |

`null` means `refreshClineToken` emits no unrecoverable sentinel, so a **permanently consumed** refresh token is handled as a **transient** failure. `tokenHealthCheck` never reaches its unrecoverable branch, and so never runs:

- the `credentialsChangedSinceSweep` rotation-race guard,
- the `Refresh token consumed … Please re-authenticate this account.` message,
- the dead-token clear for rotating providers.

Instead the connection falls into `buildRefreshFailureUpdate`: stays `active` with `errorCode: "refresh_failed"`, gets an exponential `refreshCircuit` backoff, retries the **same** consumed token 3× per attempt (`All 3 retry attempts failed for cline`), and 401s every request routed to it — indefinitely, with no actionable signal to the operator.

This is the same failure class as the `1352×` `claude/aa5dd5cf` loop that motivated `extractOAuthErrorCode`, but triggered by body **content** rather than body **shape**, which the existing shape matrix doesn't cover.

## Change

**`open-sse/services/tokenRefresh/shared.ts`** — scan for a known unrecoverable code embedded in the error value, as a **last resort** after the exact-match and nested-JSON paths. Matched with `(?<![0-9a-z_])…(?![0-9a-z_])` so it stays conservative. No quantifiers, so no ReDoS surface.

Deliberate trade-off, documented in the code: a message that merely *mentions* a dead-token code is now treated as unrecoverable. That fails safe — the alternative is an unbounded retry loop that burns a rotating provider's refresh tokens.

**`open-sse/services/refreshSerializer.ts`** — add `cline` to `ROTATION_LOCK_GROUP`. `refreshClineToken` reads a new `refreshToken` out of every response body, and a measured refresh rotated a connection's stored token, so sibling connections must not refresh concurrently. `cline` was already listed in `tokenHealthCheck`'s `ROTATING_REFRESH_PROVIDERS` but missing from the serializer.

## Validation

TDD per Hard Rule #18 — 3 new tests added to the existing `oauth-refresh-error-resilience` suite. Confirmed red on the base ref (extracted with `git show origin/release/v3.8.51:…`, not a stash):

```
base => null  (test wants "invalid_grant")  => test FAIL   × 3
3/3 new assertions FAIL on the base ref => genuine regression guards
```

Green after, with no regressions across every suite touching this code — **90 tests, 8 files**:

| suite | result |
|---|---|
| `oauth-refresh-error-resilience` | 22/22 (19 pre-existing + 3 new) |
| `token-refresh-service` | 40/40 |
| `provider-limits-rotating-expired-guard` | 7/7 |
| `refresh-serializer` | 6/6 |
| `oauth-400-recovery` | 6/6 |
| `refresh-serializer-spacing` | 3/3 |
| `codex-manual-refresh-rotating-guard` | 3/3 |
| `oauth-connection-test-codex-endpoint` | 3/3 |

`npm run typecheck:core` clean; ESLint clean on all three files.

The false-positive guards are covered explicitly: `server_error`, `rate_limited`, a 502 HTML page, `xinvalid_grant`, `invalid_grantx` and `my_invalid_grant_flag` all still classify as `null`.

## Field evidence

Diagnosed on a live gateway running 5 `cline` connections. The error body above is copied verbatim from its logs. Measured that cline rotates: a forced refresh moved one connection's stored refresh token to a different value while the four untouched connections kept theirs byte-identical. Observed the resulting symptom — connections flapping `active` / `refresh_failed`, `Skipping cline/… status=expired` in combo routing, and repeated `[401]: Unauthorized: Please make sure you're using the latest version of Cline and re-authenticate your Cline account` (cline's inference endpoint rejecting the stale access token — the message is misleading, it has nothing to do with client version).

⚠️ base-red inherited: #12732

### DIFF
--- a/open-sse/services/refreshSerializer.ts
+++ b/open-sse/services/refreshSerializer.ts
@@ -27,6 +27,12 @@ const ROTATION_LOCK_GROUP: Record<string, string> = {
   "gitlab-duo": "gitlab-duo",
   kiro: "kiro",
   "kimi-coding": "kimi-coding",
+  // Cline rotates on every refresh — `refreshClineToken` reads a new
+  // `refreshToken` out of the response body, and a measured refresh moved a
+  // connection's stored token to a different value. It was missing here while
+  // already listed in tokenHealthCheck's ROTATING_REFRESH_PROVIDERS, so sibling
+  // connections could refresh concurrently and present superseded tokens.
+  cline: "cline",
 };
 
 // Protective settle gap (ms) between two consecutive sibling refreshes when the
--- a/open-sse/services/tokenRefresh/shared.ts
+++ b/open-sse/services/tokenRefresh/shared.ts
@@ -42,6 +42,17 @@ const UNRECOVERABLE_OAUTH_ERROR_CODES = new Set([
 ]);
 
 /**
+ * Matches a known unrecoverable code EMBEDDED in a human-readable message, on
+ * word boundaries (`_` counts as a word char, so `xinvalid_grant` never hits).
+ * Alternation is safe against overlap because every candidate is delimited on
+ * both sides. Built once — the code set is a module constant.
+ */
+const EMBEDDED_OAUTH_ERROR_CODE_RE = new RegExp(
+  `(?<![0-9a-z_])(${Array.from(UNRECOVERABLE_OAUTH_ERROR_CODES).join("|")})(?![0-9a-z_])`,
+  "i"
+);
+
+/**
  * Extract a canonical OAuth error code from a refresh-endpoint error body of
  * ANY shape. Production proxies/MITMs deliver the same `invalid_grant` 400 in
  * several shapes — a plain object `{error:"invalid_grant"}`, a nested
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Excellent, careful fix — the word-boundary regex built from
`UNRECOVERABLE_OAUTH_ERROR_CODES` (with the false-positive guard test) is exactly the
right level of caution for a "last resort" match, and bundling the `cline` addition to
`ROTATION_LOCK_GROUP` makes sense given it's already in `ROTATING_REFRESH_PROVIDERS`
elsewhere but was missing from the serializer. All 22 tests pass locally, including the
transient-error and substring-false-positive guards. Only thing missing is a changelog
fragment under `changelog.d/fixes/` — once that's added this is ready to merge.

fouadSalkini:
Added the changelog fragment in commit `1a1ce805a` under `changelog.d/fixes/cline-refresh-token-classification.md`. All 22 tests passing.

PR #13485, opened 2026-09-12
fix(mistral): strip GLM thinking params

rafaeldrincon:
Mistral GLM 5.2 is exposed under model ID `zai-glm-5-2`, while existing matching only covered dotted `zai-glm-5.2`.

This caused Claude-style `thinking` and `reasoning` fields to reach Mistral and produce HTTP 422 `extra_forbidden`.

Changes:
- Match both dotted and hyphenated model IDs.
- Add regression coverage for `mistral/zai-glm-5-2`.

Validation:
- Docker image builds successfully.
- Container health check passes.
- Hermes smoke request returns HTTP 200.


rafaeldrincon:
Superseded by PR #13486 targeting latest release/v3.8.51.

PR #13486, opened 2026-09-13
fix(mistral): strip GLM thinking params

rafaeldrincon:
Mistral GLM 5.2 is exposed under model ID `zai-glm-5-2`, while existing matching only covered dotted `zai-glm-5.2`.

This caused Claude-style `thinking` and `reasoning` fields to reach Mistral and produce HTTP 422 `extra_forbidden`.

Changes:
- Match both dotted and hyphenated model IDs.
- Add regression coverage for `mistral/zai-glm-5-2`.

Validation:
- Docker image builds successfully.
- Container health check passes.
- Hermes smoke request returns HTTP 200.


diegosouzapw:
Nice catch on the hyphenated vs. dotted model-id mismatch — the fix and regression test are
correct. I noticed your other PR (#13487) branches directly off this commit and already
includes this exact change plus two more fixes, so we're planning to merge #13487 (which
carries this work forward) and close this one as subsumed rather than merging both — no
credit lost since it's the same commit either way. Let us know if you'd rather we merge this
one standalone instead.

PR #13487, opened 2026-09-13
fix: restore provider safeguards

rafaeldrincon:
Restores two local safeguards lost while updating from release/v3.8.49:

- Mark Cerebras models as rejecting Claude-style `thinking`.
- Return a clear conflict response when test-all targets a provider with no active connections.

The stale Qwen SPA version change was discarded because `open-sse/executors/qwen-web.ts` no longer exists in current upstream.

### DIFF
--- a/open-sse/config/providers/registry/cerebras/index.ts
+++ b/open-sse/config/providers/registry/cerebras/index.ts
@@ -9,8 +9,8 @@ export const cerebrasProvider: RegistryEntry = {
   authType: "apikey",
   authHeader: "bearer",
   models: [
-    { id: "zai-glm-4.7", name: "GLM 4.7" },
-    { id: "gemma-4-31b", name: "Gemma 4 31B" },
-    { id: "gpt-oss-120b", name: "GPT OSS 120B" },
+    { id: "zai-glm-4.7", name: "GLM 4.7", unsupportedParams: ["thinking"] },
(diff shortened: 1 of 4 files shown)


diegosouzapw:
Verified this branch actually contains #13486's commit as a direct ancestor, so this PR is a
strict superset (Cerebras `thinking` safeguard + the mistral GLM 5-2 fix from #13486 + the
test-all 409 response). Ran the shared test file: 25/25 pass. Since #13486 is fully included
here, we'll treat it as subsumed by this PR (same author, no credit concern) rather than
merging both. Two asks before merge: a changelog fragment, and a small test for the test-all
409 response (currently untested).

rafaeldrincon:
Follow-up on CI run 34972791255:

- Added test `tests/unit/models-test-all-route.test.ts` passes: `test-all returns 409 when provider has no active connections`.
- PR-specific checks pass: API route typecheck, changelog/merge integrity, docs, ESLint, Semgrep, Vitest.
- Remaining unit/quality failures are outside this PR diff: missing `serve.ready_timeout` translation key, `persistAttemptLogs`, `hidePaidModels`, Jina catalog naming, and hard-lease inventory tests.
- Worktree is clean at commit `381aa614c`.

No further PR-scoped changes appear necessary; these failures need baseline maintenance or separate fixes.

PR #13491, opened 2026-09-13
fix(pii): add OpenRouter metadata keys to PII sanitizer skip list (#13488)

### LINKED ISSUE #13488: fix(backend): streaming PII sanitizer splices OpenRouter metadata into answer text (scrambled output)
### OmniRoute Version

### Installation Method
Built from source

### Operating System
Linux

### Node.js Version
22.23.2

### Provider(s) Involved
`openrouter` (any provider whose stream chunks carry extra top-level / nested string fields — OpenRouter's `provider`, `reasoning_details[].text`, `reasoning_details[].format`)

### Model(s) Involved
`openrouter/z-ai/glm-5.3-flash`, `openrouter/nvidia/nemotron-3-ultra-550b-a55b:free` (any model served via OpenRouter reproduces it)

### Client Tool
Any OpenAI-compatible client (`POST /v1/chat/completions`, `stream: true`)

### Description
With the `PII_RESPONSE_SANITIZATION` feature flag enabled (even in `PII_RESPONSE_SANITIZATION_MODE=warn`, where nothing is supposed to be modified), streamed answers come back as scrambled text: fragments of the upstream provider name are spliced into the assistant content every few characters, and the content shows up inside the `provider` field of the chunks instead.

Example of what the client receives for a three-sentence answer about Finnish lakes (`hermes-auto` combo → `z-ai/glm-5.3-flash`, served by Together):

```
ne: server ogether1ownTogetndunknownTogetIL_REDAunknownToge]

KeepunknownTogeterse,unknown no preanownTogether no nknownTogetherFinTogether routher 188,0er lakes, giving it the nickname "Tand of a thousanr lakes."
Laer Saimaa,Together the  is
```

and the raw chunks:

```
data: {"id":"gen-…","model":"z-ai/glm-5.3-flash","provider":"To","system_fingerprint":"default","choices":[{"index":0,"delta":{"content":"","role":"assistant","reasoning":"","reasoning_details":[{"type":"reasoning.text","text":"getherTheu","format":"" …
data: {"id":"gen-…","model":"z-ai/glm-5.3-flash","provider":"ogether ","… "reasoning_details":[{"type":"reasoning.text","text":"user" …
data: {"id":"gen-…","model":"z-ai/glm-5.3-flash","provider":"\nunknown","… "reasoning_details":[{"type":"reasoning.text","text":"Together1. " …
data: {"id":"gen-…","model":"z-ai/glm-5.3-flash","provider":"ortunkno","… "reasoning_details":[{"type":"reasoning.text","text":"wnTogethe" …
```

(`grep -o '"provider":"[^"]*"' | sort | uniq -c` over one response gives ~150 distinct "provider" values, all of them slices of the answer.) With a model served by Nvidia the answer is peppered with `Nvidia`, with StreamLake it is `StreamLake`, etc.

Cause: `createSseTextTransform` (`src/lib/sseTextTransform.ts`, `sanitizeObject`, ~L170–L185) recursively runs the processor on **every** string property of the chunk that is not in its hard-coded `METADATA_KEYS` list. OpenRouter chunks carry top-level `provider` (e.g. `"Together"`, `"Nvidia"`, `"StreamLake"`) and `reasoning_details[{type,text,format,index}]`; `provider`, `text` and `format` are not in the list, so `getFieldCategory()` classifies them as `"content"`. The PII processor in `src/lib/streamingPiiTransform.ts` (`buffers[field] += text`, ~L49; emit everything except the trailing `W = 200` chars, ~L51–L69) keeps **one FIFO buffer per (choice, category)**, so `provider`, `reasoning_details[].text` and `delta.content` of the same choice are appended to the same buffer and each field is handed back whatever happens to be 200 characters behind — the content lands in `provider`, the provider name lands in the content, and reasoning text is merged into content. The ordering is scrambled even when `mode=warn` because the windowed re-emission happens regardless of whether anything is redacted.

### Steps to Reproduce
1. Feature Flags → enable `PII_RESPONSE_SANITIZATION` (mode `warn` is enough; `redact` behaves the same).
2. Have an OpenRouter connection and any model on it (e.g. `openrouter/z-ai/glm-5.3-flash`).
3. `curl -N -H "Authorization: Bearer <key>" -H "Content-Type: application/json" -d '{"model":"openrouter/z-ai/glm-5.3-flash","stream":true,"messages":[{"role":"user","content":"Write three short sentences about lakes in Finland."}]}' http://127.0.0.1:20128/v1/chat/completions`
4. Concatenate `choices[0].delta.content` of the chunks; inspect the `provider` field of each chunk.
5. Disable the flag and repeat — the answer is clean and `provider` is constant (`"Parasail"` / `"Together"` …) in every chunk.

### Expected Behavior
Only assistant-visible text fields (`delta.content`, `delta.reasoning`/`reasoning_content`, `message.content`, Claude `text`/`thinking` deltas …) are passed through the PII window; provider/metadata strings are left untouched; content and reasoning of a choice never share a buffer.

### Actual Behavior
Provider metadata and reasoning-detail strings are fed through the same per-choice `"content"` FIFO as the answer text, so the streamed response is interleaved garbage (see above), while the PII log only reports `[PII] Detected PII in response: ip_address(3)` and gives no hint that the text was altered.

### Test Impact
Needs a new unit test

### Error Logs / Output
```
{"timestamp":"2026-09-13T00:00:15.676Z","level":"warn","component":"PII","message":"[PII] Detected PII in response: ip_address(3)"}
{"timestamp":"2026-09-13T00:00:15.791Z","level":"info","component":"02:00:15","message":"[02:00:15] 🌊 [STREAM] OPENROUTER | z-ai/glm-5.3-flash | 22339ms | complete"}
```
(no error is logged; the stream is reported as `complete`)

### Additional Context
- Both `src/lib/sseTextTransform.ts` and `src/lib/streamingPiiTransform.ts` carry their own copy of `METADATA_KEYS`; neither contains `provider`, `native_finish_reason`, `format`, `text` (OpenRouter `reasoning_details`), `system_fingerprint` is there but `provider` is not.
- Suggested direction: switch `sanitizeObject` from a deny-list to an allow-list of known text-bearing paths (`delta.content`, `delta.reasoning`, `delta.reasoning_content`, `message.content`, `content_block_delta.delta.text/thinking`, `reasoning_details[].text` → `reasoning` category, `output_text.delta` …) and key the buffer by the full field path rather than by category, so two different fields can never be spliced into each other. Any string not on the allow-list should pass through untouched. Adding `provider` to `METADATA_KEYS` would only fix OpenRouter; other providers add their own metadata strings.
- Non-streaming responses are unaffected (`sanitizePIIResponse` sanitizes each string in place, so at worst `provider` gets redacted, never reordered).

### Validation Plan
- New unit test for `createPiiSseTransform`: feed OpenAI-format chunks that carry a top-level `provider: "Together"` and `delta.reasoning_details[{type:"reasoning.text", text:"…"}]` alongside `delta.content`; assert that the concatenated `delta.content` equals the input content, that `provider` is unchanged in every emitted chunk, and that reasoning text never appears in `content` (mode `warn` and `redact`).
- Run the existing `sseTextTransform` / `streamingPiiTransform` unit tests (`node --import tsx/esm --test tests/unit/<piiTransform>.test.ts`).
- Manual: steps above with the flag enabled — answer must be byte-identical to the flag-disabled run apart from redactions.

KooshaPari:
Fixes #13488. OpenRouter chunks carry top-level provider and reasoning_details that were not in METADATA_KEYS. sanitizeObject recursively fed them through the per-choice content FIFO, scrambling assistant text with fragments of the provider name.

Add provider, reasoning_details, format, and native_finish_reason to METADATA_KEYS in both sseTextTransform.ts and streamingPiiTransform.ts.

3 new tests + 43 existing PII tests pass.

### DIFF
--- a/src/lib/sseTextTransform.ts
+++ b/src/lib/sseTextTransform.ts
@@ -147,6 +147,13 @@ export function createSseTextTransform(
             "refusal",
             "name",
             "event",
+            // #13488: OpenRouter chunks carry provider metadata and
+            // reasoning_details that must not be fed through the PII
+            // content buffer — they scramble the assistant text.
+            "provider",
+            "reasoning_details",
+            "format",
+            "native_finish_reason",
           ];
 
           // Recursively sanitize all string properties (except system metadata)
--- a/src/lib/streamingPiiTransform.ts
+++ b/src/lib/streamingPiiTransform.ts
@@ -141,6 +141,13 @@ export function createPiiSseTransform(options?: PiiTransformOptions): TransformS
       "refusal",
       "name",
       "event",
+      // #13488: OpenRouter chunks carry provider metadata and
+      // reasoning_details that must not be fed through the PII
+      // content buffer — they scramble the assistant text.
+      "provider",
+      "reasoning_details",
+      "format",
+      "native_finish_reason",
     ];
 
     // 1. Claude format
--- a/tests/unit/13488-pii-openrouter-metadata-scramble.test.ts
+++ b/tests/unit/13488-pii-openrouter-metadata-scramble.test.ts
@@ -1,0 +1,142 @@
+/**
+ * #13488 — Streaming PII sanitizer splices OpenRouter provider/reasoning_details
+ * strings into the answer text. Before the fix, sanitizeObject recursively
+ * processed every non-metadata string, so OpenRouter's top-level "provider"
+ * and reasoning_details[].text/format were fed through the same per-choice
+ * "content" FIFO as delta.content, scrambling the output.
+ *
+ * The fix adds provider, reasoning_details, format, and native_finish_reason
+ * to METADATA_KEYS in both sseTextTransform.ts and streamingPiiTransform.ts.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+
+const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "omniroute-13488-pii-"));
+process.env.DATA_DIR = tmpDir;
+process.env.PII_RESPONSE_SANITIZATION = "true";
+process.env.PII_TEST_BYPASS_MIN_WINDOW = "true";
+
+import { createPiiSseTransform } from "../../src/lib/streamingPiiTransform.ts";
+
+async function feedChunks(
+  transform: TransformStream,
+  chunks: string[]
+): Promise<string> {
+  const writer = transform.writable.getWriter();
+  const reader = transform.readable.getReader();
+
+  const writePromise = (async () => {
+    for (const chunk of chunks) {
+      await writer.write(new TextEncoder().encode(chunk));
+    }
+    await writer.close();
+  })();
+
+  const output: string[] = [];
+  let res = await reader.read();
+  while (!res.done) {
+    output.push(new TextDecoder().decode(res.value));
+    res = await reader.read();
+  }
+  await writePromise;
+  return output.join("");
+}
+
+test.after(() => {
+  fs.rmSync(tmpDir, { recursive: true, force: true });
+});
+
+test("#13488 OpenRouter provider field is not scrambled into content", async () => {
+  const transform = createPiiSseTransform();
+
+  // Three OpenRouter chunks with a stable "Together" provider and simple content.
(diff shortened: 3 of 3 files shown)


KooshaPari:

PR #13494, opened 2026-09-13
fix: run caveman user preservePatterns before built-in patterns

### LINKED ISSUE #13457: fix(backend): cavemanConfig.preservePatterns silently skipped when region contains a built-in construct
### OmniRoute Version

### Installation Method
Built from source (installed as npm global tarball)

### Operating System
Linux (Rocky Linux 10.2), Node.js v22.23.2

### Provider(s) / Model(s) Involved
Provider-independent (caveman engine, `open-sse/services/compression/preservation.ts`).

### Client Tool
Any; the motivating case is agentic coding CLIs that inject instruction files into user-role messages inside `<system-reminder>…</system-reminder>` envelopes.

### Description
`cavemanConfig.preservePatterns` is documented as the way to protect custom regions from caveman rewriting. In practice a user pattern **cannot protect any region that contains a built-in preserved construct** (inline code, heading, URL, path, version, `CONST_CASE`, function call, …), which is every real instruction block.

Cause, in `extractPreservedBlocks`:

```ts
for (const { pattern, kind } of [...builtIns, ...compileUserPatterns(options.preservePatterns)]) {
  result = replacePattern(result, ensureGlobal(pattern), kind, addBlock);
}
```
and in `replacePattern`:
```ts
return text.replace(pattern, (match) => {
  if (!match) return match;
  if (match.includes(SENTINEL_PREFIX)) return match;   // <-- user match spanning a built-in placeholder is dropped
  return addBlock(match, kind);
});
```

Built-ins run first and replace e.g. `` `.app-prev-*` `` or `# Deploy rules` with a `OMNI_CAVEMAN_…` placeholder. When the user pattern is applied afterwards, its match now contains a sentinel, so `replacePattern` returns the match unchanged — nothing is preserved and every caveman rule runs over the rest of the region. The user gets no warning: the settings schema only checks `z.array(z.string())`, `compileUserPatterns` swallows compile errors, and nothing reports a user match that was skipped because of a sentinel.

The upstream caveman contributing guide's own rule for compressors is "never silently corrupt a payload — on any parse problem pass the bytes through untouched"; a preserve pattern that silently stops preserving is the opposite of that.

### Steps to Reproduce
```ts
// npx tsx repro-preserve.mts   (repo root)
import { cavemanCompress } from "./open-sse/services/compression/caveman.ts";

const PAT = /<system-reminder>[\s\S]*?<\/system-reminder>/.source;
const run = (label: string, text: string) => {
  const r = cavemanCompress({ messages: [{ role: "user", content: text }] } as any,
    { enabled: true, compressRoles: ["user"], intensity: "full", minMessageLength: 50, preservePatterns: [PAT], skipRules: [] } as any);
  console.log("=== " + label + " rules=" + JSON.stringify((r.stats as any)?.rulesApplied));
  console.log((r.body as any).messages[0].content);
};

run("A: no built-in construct inside the region", `<system-reminder>
The backup files must never be deleted. This is really the most important rule of the project.
</system-reminder>

Please deploy the fix. Thanks, and do not restart the database.`);

run("B: one inline code span inside the region", `<system-reminder>
The backup files \`.app-prev-*\` must never be deleted. This is really the most important rule of the project.
</system-reminder>

Please deploy the fix. Thanks, and do not restart the database.`);

run("C: a heading inside the region", `<system-reminder>
# Deploy rules
The backup files must never be deleted. This is really the most important rule of the project.
</system-reminder>

Please deploy the fix. Thanks, and do not restart the database.`);
```

### Expected Behavior
All three cases keep the `<system-reminder>…</system-reminder>` region byte-identical (only the trailing sentence outside the region is compressed), because the operator's pattern matches all three. If a user pattern cannot be honoured, the engine should say so (warn log / validation error), not silently skip it.

### Actual Behavior
```
=== A: no built-in construct inside the region rules=["pleasantries","polite_framing","articles"]
<system-reminder>
The backup files must never be deleted. This is really the most important rule of the project.
</system-reminder>

Deploy fix. And do not restart database.
=== B: one inline code span inside the region rules=["pleasantries","polite_framing","emphasis_removal","articles"]
<system-reminder>
Backup files `.app-prev-*` must never be deleted. This is most important rule of project.
</system-reminder>

Deploy fix. And do not restart database.
=== C: a heading inside the region rules=["pleasantries","polite_framing","emphasis_removal","articles"]
<system-reminder>
# Deploy rules
Backup files must never be deleted. This is most important rule of project.
</system-reminder>

Deploy fix. And do not restart database.
```
A is protected; B and C are rewritten inside the "preserved" region (`The backup files` → `Backup files`, `really the most important rule of the project` → `most important rule of project`) as soon as a single built-in construct exists inside it.

### Test Impact
Needs a new unit test

### Additional Context
- `open-sse/services/compression/preservation.ts`: `SENTINEL_PREFIX`, `replacePattern` (`match.includes(SENTINEL_PREFIX)` early return), `extractPreservedBlocks` (`[...builtIns, ...compileUserPatterns(...)]` ordering), `compileUserPatterns` (`new RegExp(pattern, "g")`).
- Possible fixes: run user patterns **before** the built-ins (a user region then swallows the built-ins it contains, which is what the operator asked for); or allow nested placeholders (restore recursively); or at minimum surface a warning when a user match is skipped because of a sentinel.
- The only existing user-pattern test (`tests/unit/compression/validation.test.ts`, `preservePatterns: ["`[^`]+"]`) uses a pattern whose match coincides with the built-in `inline_code` region, so it passes for the wrong reason and does not cover a user region that *contains* built-in constructs.

### Validation Plan
- Add cases B and C above (user pattern region containing inline code / heading / URL) to `tests/unit/compression/validation.test.ts` or `caveman-preservation.test.ts`, asserting the region is byte-identical after `cavemanCompress`.
- `node --import tsx/esm --test tests/unit/compression/validation.test.ts tests/unit/compression/caveman-preservation.test.ts tests/unit/compression/caveman-engine.test.ts`

KooshaPari:
Closes diegosouzapw/OmniRoute#13457

## Problem

When `cavemanConfig.preservePatterns` included a regex whose match region overlapped with a built-in pattern (inline code span, heading, URL, etc.), the user pattern was silently skipped.

### Root cause

In `preservation.ts`, built-in patterns ran first and replaced sub-regions with sentinel placeholders. When user patterns then matched the same region, the `includes(SENTINEL_PREFIX)` guard in `replacePattern` detected the sentinel and returned the match unchanged — the operator's preservation rule was ignored with no warning.

### Evidence

The concrete example from the issue:
```ts
const PAT = /<system-reminder>[\s\S]*?<\/system-reminder>/.source;
```
Input: `<system-reminder>\nThe backup files \`.app-prev-*\` must never be deleted.\n</system-reminder>`

Before: Built-in `inline_code` matched \`.app-prev-*\` first → user pattern's full match now contains sentinel → silently skipped → rule never preserved.

## Fix

Reversed the pattern execution order in `extractPreservedBlocks`:
- **Before**: `[...builtIns, ...compileUserPatterns(...)]`
- **After**: `[...compileUserPatterns(...), ...builtIns]`

User patterns now claim the full region first. Built-ins then skip the sentinel placeholders harmlessly. No behavioral change when no user patterns are configured.

## Tests

Added `tests/unit/13457-caveman-user-patterns-before-builtins.test.ts` (4 test cases):
- User pattern protects region containing inline code span
- User pattern protects region containing markdown heading
- User pattern protects region containing URL
- Baseline: no user patterns → built-ins still work as before

All 25 existing caveman preservation/engine tests remain green.

---

*Test plan: verified locally — `node --import tsx/esm --test`*

### DIFF
--- a/open-sse/services/compression/preservation.ts
+++ b/open-sse/services/compression/preservation.ts
@@ -125,7 +125,13 @@ export function extractPreservedBlocks(
     },
   ];
 
-  for (const { pattern, kind } of [...builtIns, ...compileUserPatterns(options.preservePatterns)]) {
+  // #13457: User patterns MUST run before built-ins.  Built-ins replace content
+  // with sentinels; if they run first, a user pattern whose match spans a
+  // sentinel-bearing region hits the `includes(SENTINEL_PREFIX)` guard and is
+  // silently skipped -- the operator's preservation rule is ignored with no
+  // warning.  Running user patterns first lets them claim the full region,
+  // and built-ins then skip the sentinel placeholders harmlessly.
+  for (const { pattern, kind } of [...compileUserPatterns(options.preservePatterns), ...builtIns]) {
     result = replacePattern(result, ensureGlobal(pattern), kind, addBlock);
   }
 


KooshaPari:

PR #13496, opened 2026-09-13
fix(dashboard): skip image/music/video/audio models in Test all models

### LINKED ISSUE #13376: fix(dashboard): "Test all models" has no modality/cost guard — image and music models are billed as real generations
### Summary

`POST /api/models/test-all` ("Test all models" in the provider card) runs a real request
against **every** model id the UI hands it, with no modality filter and no cost guard. When a
provider's catalog mixes chat models with **image / music / video** models, the batch test
issues billable *generations* for those entries.

Reported by a community member on the WhatsApp support channel (2026-09-10):

> A warning for the community: "Test all models" on some providers will test not only chat
> models but music and image ones — just burned $10.

### Why it happens

`src/app/api/models/test-all/route.ts` validates `{ providerId, modelIds[] }` and loops
`runSingleModelTest` over the ids. The runner
(`src/lib/api/modelTestRunner.ts`, `runInner`) branches only on:

- embeddings → `handleValidatedEmbeddingRequestBody`
- rerank → `postRerank`
- audio transcription → `postAudioTranscription`
- **everything else → `postChatCompletion`**

There is no modality classification and no price check anywhere in that path, so an image or
music model id falls into the last branch and is dispatched as a normal billable request. The
only existing guards are unrelated to cost:

- `CONSECUTIVE_RATE_LIMIT_STOP_THRESHOLD` (429 storms)
- `SLOW_PROBE_PROVIDERS` (burst-ban avoidance)
- `hidePaidModels`, which is opt-in and off by default

### Expected behavior

Batch model testing should not be able to spend real money without the operator explicitly
asking for it. Any of these would fix the reported damage:

1. Exclude non-chat modalities (image / music / video) from the batch by default — they are not
   what "is this model reachable?" needs.
2. If they must be testable, require an explicit opt-in checkbox and show an estimated cost
   before running.
3. At minimum, warn in the UI when the selected set contains billable generation models.

### Steps to reproduce

1. Connect a provider whose catalog includes image and/or music models alongside chat models.
2. Open the provider card and run **Test all models**.
3. Observe real generations billed for the image/music entries.

### Test impact

Needs a new unit test: `test-all` must not dispatch image/music/video model ids (or must require
the opt-in flag before doing so).

KooshaPari:
Closes diegosouzapw/OmniRoute#13376

## Problem

`POST /api/models/test-all` dispatches a real `chatCompletion` request for **every** model ID the UI hands it, with no modality filter. When a provider catalog mixes chat models with image/music/video models, the batch test issues billable generations for non-chat entries.

A community member reported burning $10 on a single test-all run.

## Root cause

`detectTestKind()` in `modelTestRunner.ts` only recognized embeddings and rerank. Everything else (including image/music/video/audio generation models) fell through to `postChatCompletion`, dispatching a real billable request.

## Fix

1. **Expanded `detectTestKind()`** to return `isNonChatGeneration` — detects models whose `supportedEndpoints` contains `"images"`, `"image"`, `"music"`, `"video"`, or `"audio"`, or whose `apiFormat` matches `"image-generation"`, `"music-generation"`, `"video-generation"`, or `"audio-generation"`.

2. **Early return in `runSingleModelTest()`** — when `isNonChatGeneration` is true, returns an error entry immediately without dispatching, so the batch loop skips the model.

This preserves single-model testing (users can still explicitly test an image model) while preventing batch runs from burning money on non-chat models.

## Tests

13 new test cases in `tests/unit/13376-detect-test-kind-non-chat-generation.test.ts`:
- Embedding/rerank detection unchanged (baseline)
- Chat models not falsely flagged
- Image/music/video/audio detected via `supportedEndpoints`
- Image/music detected via `apiFormat`
- Multi-endpoint models flagged
- Unknown/empty metadata defaults to chat

All 8 existing `test-all-model-status` and `test-all-results-i18n` tests remain green.

---

*Test plan: verified locally — `node --import tsx/esm --test`*

### DIFF
--- a/src/lib/api/modelTestRunner.ts
+++ b/src/lib/api/modelTestRunner.ts
@@ -158,6 +158,8 @@ async function findProviderNodeApiType(providerId: string): Promise<string | und
   }
 }
 
+const NON_CHAT_GENERATION_ENDPOINTS = new Set(["images", "image", "music", "video", "audio"]);
+
 export function buildInternalChatRequest(
   testBody: Record<string, unknown>,
   signal: AbortSignal,
@@ -306,7 +308,15 @@ export function detectTestKind(modelStr: string, customModel: any, nodeApiType?:
     (apiFormat === "responses" ||
       nodeType === "responses" ||
       supportedEndpoints.includes("responses"));
-  return { isRerank, isEmbedding, isAudioTranscription, isResponses };
+  // #13376: image/music/video/audio generation models are not chat-testable.
+  // Dispatching them as chatCompletion burns billable generations for no useful
+  // health signal.  Detect via supportedEndpoints or apiFormat.
+  const isNonChatGeneration =
+    !isRerank &&
+    !isEmbedding &&
+    (supportedEndpoints.some((ep: string) => NON_CHAT_GENERATION_ENDPOINTS.has(ep)) ||
+      ["image-generation", "music-generation", "video-generation", "audio-generation"].includes(apiFormat));
+  return { isRerank, isEmbedding, isAudioTranscription, isResponses, isNonChatGeneration };
 }
 
 /**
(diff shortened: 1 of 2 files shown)


KooshaPari:

KooshaPari:

PR #13501, opened 2026-09-13
fix(zed-hosted): map role 'developer' to 'system' for Zed OpenAI proxy

### LINKED ISSUE #13362: fix(providers): zed-hosted: OpenAI models reject role developer (unknown variant developer) with more than one system message
### OmniRoute Version
3.8.50 (BUILD_SHA dea6bb8)

### Installation Method
npm (global)

### Operating System / Node.js
Linux (Ubuntu VPS), Node v22.23.1

### Description
Every OpenAI model served through the `zed-hosted` provider (gpt-5.5, gpt-5.4, gpt-5.3-codex, gpt-5-mini, gpt-5-nano, ...) fails with HTTP 400 as soon as the request carries more than one system message. Claude Code always does (its system prompt is split in blocks and it injects reminders), so the models are unusable from Claude Code even though a plain single-system `curl` works.

Upstream response from `cloud.zed.dev/completions`:

```
[400]: failed to parse OpenAI Responses API request: unknown variant `developer`, expected one of `user`, `assistant`, `system`, `tool`
```

### Root cause
`open-sse/translator/request/openai-responses/toResponses.ts` (around line 110-128) keeps the first system message as `instructions` and emits every further system/developer message as an input item with `role: "developer"` (#6954). That is valid for the OpenAI Responses API, but Zed's proxy deserializes the request with its own `Role` enum, which only has `User`, `Assistant`, `System` (see `crates/open_ai/src/open_ai.rs` in zed-industries/zed). The `zed-hosted` executor (`open-sse/executors/zed-hosted.ts`, `buildProviderRequest`) calls `openaiToOpenAIResponsesRequest` unchanged, and `translator/index.ts` already has a `preserveDeveloperRole=false` switch that the executor does not use.

### Reproduction
```bash
curl -s http://127.0.0.1:20128/v1/chat/completions \
  -H "Authorization: Bearer $OMNIROUTE_API_KEY" -H "Content-Type: application/json" \
  -d '{"model":"zed-hosted/gpt-5-nano","max_tokens":256,"messages":[
        {"role":"system","content":"Be concise."},
        {"role":"system","content":"Answer with one word."},
        {"role":"user","content":"Say hi"}]}'
```
Expected: 200. Actual: 400 with the message above.

### Proposed fix
In the `zed-hosted` executor, after `openaiToOpenAIResponsesRequest(...)` for the `open_ai` provider, map every `input[].role === "developer"` to `"system"` (Zed accepts `system` as an input-item role). Verified against the live proxy: with that single rewrite applied in-process, the same request and a full Claude Code interactive session return 200.

A related report for the Gemini family (safety threshold and function-calling mode) is filed separately.

KooshaPari:
Fixes #13362

## Problem

Every OpenAI model served through `zed-hosted` fails with HTTP 400 when the request carries more than one system message:

```
[400]: failed to parse OpenAI Responses API request: unknown variant `developer`, expected one of `user`, `assistant`, `system`, `tool`
```

Claude Code always sends multiple system messages (system prompt split in blocks + injected reminders). The Responses translator may emit input items with `role: developer` which is valid for OpenAI but rejected by Zed's proxy.

## Fix

Add a `developer` → `system` role mapping pass in `normalizeForZedProxy()` that runs after both `openaiToOpenAIResponsesRequest()` and `openaiToGeminiRequest()` in the zed-hosted executor. This is a no-op for requests that already use standard roles.

## Files changed

- `open-sse/executors/zed-hosted.ts` — added normalizeForZedProxy() with developer→system mapping + applied it in the OpenAI Responses path
- `tests/unit/13362-zed-hosted-developer-role-mapping.test.ts` — 8 tests covering role mapping cases

## Testing

- 8/8 new tests pass
- Existing tests unaffected (normalization is zed-hosted-specific)

### DIFF
--- a/open-sse/executors/zed-hosted.ts
+++ b/open-sse/executors/zed-hosted.ts
@@ -95,7 +95,13 @@ function buildProviderRequest(
     return openaiToGeminiRequest(model, body as Record<string, unknown>, true, credentials);
   }
   if (provider === ZED_PROVIDER.openai) {
-    return openaiToOpenAIResponsesRequest(model, body, true, credentials);
+    const responsesReq = openaiToOpenAIResponsesRequest(
+      model,
+      body,
+      true,
+      credentials
+    ) as Record<string, unknown>;
+    return normalizeForZedProxy(responsesReq);
   }
   return {
     ...(body as Record<string, unknown>),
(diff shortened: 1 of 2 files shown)


KooshaPari:

PR #13507, opened 2026-09-13
fix(guardrails): skip data: URI base64 payloads in CredentialMasker to prevent false-positive redaction (#13462)

### LINKED ISSUE #13462: fix(backend): CredentialMasker corrupts valid image base64 and blocks continued sessions when credential redaction is enabled
### Triage and delivery target

- Requested labels: `bug`, `backend`. The reporter has no upstream triage permission; please apply these labels during maintainer triage.
- **Fix / PR target: `release/v3.8.51` only.**
- **Affected published version: `3.8.50`. This is historical reproduction evidence, NOT a request to reopen that release, target it with a PR, or backport the fix.**
- No changes to the installed 3.8.50 package are proposed.
- No release milestone is assigned; the repository currently exposes no open milestone through the API.

### OmniRoute Version

3.8.50, installed via npm globally. The deterministic redaction reproduction also fails against active `release/v3.8.51` at `152d95108c9c3d557562311ffed63240a511eb31` (not a published 3.8.51 release).

### Installation Method

npm (global)

### Operating System

Linux

### Node.js Version

Production installation uses Node 24. Local isolated reproduction was run with Node v24.20.0.

### Provider(s) Involved

Observed with the Codex subscription provider. The implicated credential masker is provider-independent; other providers are not live-qualified by this report.

### Model(s) Involved

`codex/gpt-5.6-sol-medium`

### Client Tool

DeepSeek Harness (DSH), sending OpenAI Chat Completions to OmniRoute, translated to Codex Responses upstream.

### Description

With `credentialRedactionEnabled=true`, CredentialMasker recursively treats an image's base64 transport bytes as ordinary text. A coincidental token-pattern match inside a valid PNG data URL is replaced with `[REDACTED:google]`. The resulting data URL is no longer valid base64, and the Codex upstream rejects the request with HTTP 400.

This blocked an existing long-running session on every retry because its history kept including the same screenshot. The persisted session and original attachment were intact; deleting conversation history was not necessary for the operator to resume.

This is a payload-integrity/false-positive bug in an explicitly enabled guardrail, not evidence that a real Google credential was present in the screenshot or that authentication was bypassed.

### Steps to Reproduce

1. Enable `credentialRedactionEnabled` in OmniRoute.
2. Send a valid image data URL whose base64 encoding happens to contain a substring matching the Google credential regex.
3. The masker scans the string at a structured image field, replaces the matching base64 characters, and forwards an invalid image.
4. A strict upstream returns HTTP 400. Repeating the turn with the image still in history repeats the failure.

The following fully synthetic fixture reproduces the collision without customer images, credentials, network requests, or a running provider. It creates a valid 1x1 RGBA PNG with correct chunk CRCs and a private ancillary chunk. That chunk deliberately aligns harmless bytes so their base64 representation matches the pattern. This is a controlled reproduction of the random collision observed in a real screenshot.

Run the TypeScript snippet using the repository's normal `node --import tsx/esm` test/TS launcher, in an isolated development environment:

```ts
import assert from "node:assert/strict";
import { deflateSync } from "node:zlib";
import { redactCredentials } from "./src/lib/guardrails/credentialMasker.ts";

function crc32(bytes: Buffer): number {
  let crc = 0xffffffff;
  for (const byte of bytes) {
    crc ^= byte;
    for (let bit = 0; bit < 8; bit++) {
      crc = (crc >>> 1) ^ ((crc & 1) ? 0xedb88320 : 0);
    }
  }
  return (crc ^ 0xffffffff) >>> 0;
}

function chunk(name: string, data: Buffer): Buffer {
  const type = Buffer.from(name);
  const length = Buffer.alloc(4);
  length.writeUInt32BE(data.length);
  const crc = Buffer.alloc(4);
  crc.writeUInt32BE(crc32(Buffer.concat([type, data])));
  return Buffer.concat([length, type, data, crc]);
}

const marker = ["AI", "za"].join("") + "A".repeat(36); // synthetic, not a credential
const ihdr = Buffer.alloc(13);
ihdr.writeUInt32BE(1, 0);
ihdr.writeUInt32BE(1, 4);
ihdr[8] = 8;
ihdr[9] = 6;

const png = Buffer.concat([
  Buffer.from("89504e470d0a1a0a", "hex"),
  chunk("IHDR", ihdr),
  chunk("raNd", Buffer.concat([Buffer.alloc(1), Buffer.from(marker, "base64")])),
  chunk("IDAT", deflateSync(Buffer.from([0, 255, 0, 0, 255]))),
  chunk("IEND", Buffer.alloc(0)),
]);

const url = "data:image/png;base64," + png.toString("base64");
assert.equal(png.length, 113);
assert.ok(url.includes(marker));

const redacted = redactCredentials(url);
console.log({
  modified: redacted.modified,
  detections: redacted.detections,
  base64Valid: /^[A-Za-z0-9+/]*={0,2}$/.test(redacted.text.split(",")[1]),
});
```

Observed:

```json
{
  "modified": true,
  "detections": [{"type": "google", "count": 1}],
  "base64Valid": false
}
```

The 113-byte PNG was independently decoded successfully with Sharp before applying redaction. For the local reproduction, the exported pure redaction function/patterns were loaded in memory from the installed source and from the pinned release source, avoiding database initialization or changes to production. The snippet above presents the equivalent ordinary repository import for maintainers.

Important test boundary: the direct `redactCredentials` call demonstrates the collision, but does NOT imply that this text-redaction utility should stop masking token-like text everywhere. A permanent regression must exercise the public guardrail `preCall`/`postCall` over structured image-bearing payloads; the bug is the walker applying text rewriting to binary transport fields.

### Expected Behavior

With credential masking enabled:

- Recognized binary image payloads survive byte-for-byte.
- Text next to those images still has actual credential patterns masked.
- Authorization/header handling remains protected.
- The guardrail does not turn a valid image into invalid base64.
- The operator does not need to disable credential protection globally to continue an existing session.

### Actual Behavior

The same historical image is rejected on each retry:

```text
[400]: Invalid 'input[519].content[2].image_url'.
Expected a base64-encoded data URL with an image MIME type
(e.g. '[REDACTED_DATA_URL]'), but got an invalid base64-encoded value.
```

`[REDACTED_DATA_URL]` above is the sanitized example in the error message. It is not the evidence for request corruption. The independently reproduced `[REDACTED:google]` substitution inside the image encoding is that evidence.

### Root Cause / Source Pointers

In the installed 3.8.50 source:

- `src/lib/guardrails/credentialMasker.ts`: Google pattern `/AIza[0-9A-Za-z_-]{35}/g`.
- `walkValue()`: every string is passed to `redactCredentials()`, including nested `image_url` data URLs; there is no image/binary-field discrimination.
- `preCall()` returns the rewritten payload when the setting or environment flag enables the guardrail.
- The same recursive walker is used for `postCall()`; request corruption is reproduced here, response/media paths need regression coverage rather than being claimed live-tested.

At pinned release commit `152d95108c9c3d557562311ffed63240a511eb31`, patterns live in `open-sse/utils/credentialPatterns.ts`; the synthetic collision still reproduces.

### Local Evidence and Session Recovery

On 2026-09-12:

- Two recorded failures occurred at 16:59:23Z and 17:06:45Z with the same upstream image index.
- All five original image objects referenced by the affected session matched their recorded SHA-256 identifiers and byte lengths.
- Only one image matched the Google pattern in the source-level masking check.
- The problematic original PNG was 537,594 bytes, 1440x900. It was not a truncated or missing attachment.
- The session's records preserved the image references and failed attempts.
- After the temporary setting change below, the operator reported that the same session resumed.
- A subsequent read-only inspection of the same persisted session found 45 new assistant-message events and 45 tool-call events after its last recorded image-validation failure.

Evidence limitation: the full final outgoing provider body for the failing request was not retained in available logs. The report combines source-level deterministic reproduction, preserved attachment integrity, and resumed-session evidence; it does not claim a captured byte-by-byte production ingress/egress comparison, completion of the entire user task, or semantic revalidation of every screenshot.

Session IDs, API key identifiers, private project paths, customer screenshots, and transcript contents are deliberately omitted.

### Temporary Workaround Actually Used — Not the Fix

The operator explicitly chose to set:

```json
{"credentialRedactionEnabled": false}
```

through the supported `PATCH /api/settings` API.

- A restricted database backup was created and integrity-checked first.
- GET readback confirmed `true -> false`.
- Only that setting and its revision changed.
- The inspected launcher environment did not force `CREDENTIAL_REDACTION_ENABLED=true`.
- The gateway remained healthy; no restart, package patch, session edit, or attachment replacement was performed.
- The same session resumed, as described above.

**Security tradeoff:** this disables the credential masker globally for payloads it would otherwise process. It may allow credentials present in text/tool results to pass through unmasked. It is an operator-authorized emergency mitigation, NOT a recommended permanent fix or a request to change default security settings.

The code enables masking if either the setting is true OR `CREDENTIAL_REDACTION_ENABLED` is exactly `"true"`; changing the DB setting alone is not a universal workaround if the environment still forces it on.

### Test Impact

Needs new focused unit tests and a guardrail/pipeline integration regression.

### Proposed Fix Boundary / Acceptance Criteria

1. Make the payload traversal media-aware so recognized base64 image transport fields are not rewritten as text.
2. Preserve redaction in adjacent text, tool arguments/results that are text, and sensitive structured headers.
3. Do not exempt an arbitrary object simply because it has a key named `data`, `url`, or `image_url`; define and test the supported media shapes.
4. Preserve valid images in OpenAI Chat image_url (object), Responses input_image (string URL), and review relevant Claude/Gemini image shapes for the same walker.
5. Cover request and applicable non-stream response processing, including disabled-mode passthrough.
6. Exercise the guardrail entrypoints, not only the pure string helper. Prove that a fake credential in neighboring text is still redacted.
7. Add a mocked upstream round-trip/validation test showing that the enabled guardrail does not corrupt the image.
8. Do not weaken global credential patterns, switch masking off by default as a fix, or modify operator sessions.
9. Do not reuse private screenshots in the PR; use this synthetic fixture.

### Validation Plan

- Add a focused regression under `tests/unit/`, alongside `tests/unit/credential-masker-guardrail.test.ts`.
- Demonstrate RED before / GREEN after for media integrity, while text/header redaction controls stay GREEN.
- Run the existing credential-masker suite and relevant credential-pattern/sanitization suites.
- Run `npm run lint` and the focused checks required by the Contribution Golden Path.
- Record inherited base failures separately; CI owns full unit shards, Vitest, coverage ratchets, and production build.
- Target the active release branch after reconciling against its current tip. No production-installed package patch is part of this contribution.

### Duplicate / Related Work Check

Searched open/closed issues and PRs for credential + base64, CredentialMasker + image, and related redaction/data-URL terms. No equivalent media false-positive report was found at submission time.

- #7683 introduced CredentialMasker.
- #12964 concerns credential redaction in error bodies, not preserving binary request payloads.
- #8459 concerns Responses-to-Chat image translation/stringification, a different defect.
- openai/codex#32287 reports the same validation error symptom, but that report alone does not establish this root cause.

This report is intentionally separate from image-generation response-envelope fixes, web-search fallback problems, and reasoning/model routing work.

KooshaPari:
## Summary

When `credentialRedactionEnabled=true`, `CredentialMaskerGuardrail` recursively scans all string values for well-known credential patterns. However, it does not account for `data:` URIs carrying base64-encoded binary (images, audio, etc.). The base64 encoding can randomly collide with credential regex patterns — for example, the Google `AIza` pattern matching inside a PNG's base64 stream.

This corrupts the image payload, causing the upstream provider to reject the request with HTTP 400.

## Fix

Temporarily replace `data:...;base64,...` URI payloads with placeholders before applying credential regex patterns, then restore the originals afterward. This ensures base64 content is never scanned for credentials.

### Changes
- `src/lib/guardrails/credentialMasker.ts`: Added data URI detection and placeholder logic in `redactCredentials()`
- `tests/unit/13462-credential-masker-data-uri.test.tsx`: 5 test cases covering:
  - Data URI with base64 collision passes through intact
  - Real credentials in non-data-URI text are still redacted
  - Mixed data URI + real credential in same string
  - Multiple data URIs in one string
  - No-op for plain text with no credentials

## Test Results
All 5 tests pass: `✓ tests/unit/13462-credential-masker-data-uri.test.tsx (5 tests)`

Closes #13462

### DIFF
--- a/src/lib/guardrails/credentialMasker.ts
+++ b/src/lib/guardrails/credentialMasker.ts
@@ -23,17 +23,38 @@ export interface CredentialRedactionResult {
 
 export function redactCredentials(text: string): CredentialRedactionResult {
   if (typeof text !== "string" || !text) return { text, detections: [], modified: false };
-  let result = text;
+
+  // #13462: data: URIs carry base64-encoded binary (images, audio, etc.) whose
+  // encoding can randomly collide with credential regex patterns — e.g. the
+  // Google AIza pattern matching inside a PNG's base64 stream.  Temporarily
+  // replace the base64 payload with a placeholder, redact the remaining text,
+  // then restore the original payload.
+  const DATA_URI_RE = /((?:data:[^;]+;base64,)[A-Za-z0-9+/=%]+)/g;
+  const dataUriBodies: string[] = [];
+  let prepared = text.replace(DATA_URI_RE, (_match, _body: string, offset: number) => {
+    // capture the full match including the data:…;base64, prefix
+    const full = text.slice(offset, offset + _match.length);
+    const idx = dataUriBodies.length;
+    dataUriBodies.push(full);
+    return `__DATA_URI_${idx}__`;
+  });
+
   const detections: Array<{ type: string; count: number }> = [];
   for (const p of CREDENTIAL_PATTERNS) {
     p.regex.lastIndex = 0;
-    const matches = result.match(p.regex);
+    const matches = prepared.match(p.regex);
     if (matches && matches.length > 0) {
-      result = result.replace(p.regex, p.replacement);
+      prepared = prepared.replace(p.regex, p.replacement);
       detections.push({ type: p.name, count: matches.length });
     }
   }
-  return { text: result, detections, modified: result !== text };
+
+  // Restore original data: URI payloads (unmodified)
+  if (dataUriBodies.length > 0) {
+    prepared = prepared.replace(/__DATA_URI_(\d+)__/g, (_m, idx: string) => dataUriBodies[Number(idx)]);
+  }
+
+  return { text: prepared, detections, modified: prepared !== text };
 }
 
 type JsonRecord = Record<string, unknown>;
--- a/tests/unit/13462-credential-masker-data-uri.test.tsx
+++ b/tests/unit/13462-credential-masker-data-uri.test.tsx
@@ -1,0 +1,55 @@
+import { describe, it, expect } from "vitest";
+import { redactCredentials } from "../../src/lib/guardrails/credentialMasker";
+
+/**
+ * #13462 — CredentialMasker must not corrupt valid base64 image data by
+ * accidentally matching credential regex patterns inside the encoded payload.
+ */
+
+describe("redactCredentials — data URI false-positive guard (#13462)", () => {
+  /**
+   * A synthetic data URI whose base64 payload deliberately contains the
+   * Google API key prefix "AIza" followed by 35 alphanumeric chars.
+   * This mirrors the real collision observed with a PNG screenshot.
+   */
+  const DATA_URI_WITH_COLLISION =
+    "data:image/png;base64,AAAIzaAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABB==";
+
+  it("does NOT redact base64 payload inside a data URI", () => {
+    const input = `Here is an image: ${DATA_URI_WITH_COLLISION}`;
+    const result = redactCredentials(input);
+    // The data URI must pass through intact
+    expect(result.text).toContain(DATA_URI_WITH_COLLISION);
+    expect(result.text).not.toContain("__DATA_URI_");
+  });
+
+  it("still redacts real Google API keys in non-data-URI text", () => {
+    const input = "My Google key is AIzaSyA1234567890abcdefghijklmnopqrstuv";
+    const result = redactCredentials(input);
+    expect(result.text).toContain("[REDACTED:google]");
+    expect(result.modified).toBe(true);
+  });
+
+  it("redacts real keys while preserving data URIs in the same string", () => {
+    const input = `Key: AIzaSyA1234567890abcdefghijklmnopqrstuv\nImage: ${DATA_URI_WITH_COLLISION}`;
+    const result = redactCredentials(input);
+    expect(result.text).toContain("[REDACTED:google]");
+    expect(result.text).toContain(DATA_URI_WITH_COLLISION);
+  });
+
+  it("handles multiple data URIs in one string", () => {
+    const uri1 = "data:image/png;base64,ABCDEF==";
+    const uri2 = "data:image/jpeg;base64,GHIJKL==";
+    const input = `${uri1} and ${uri2}`;
+    const result = redactCredentials(input);
+    expect(result.text).toContain(uri1);
+    expect(result.text).toContain(uri2);
+  });
+
+  it("returns unmodified text when no credentials or data URIs present", () => {
+    const input = "Hello, this is a normal message.";
+    const result = redactCredentials(input);
+    expect(result.text).toBe(input);
+    expect(result.modified).toBe(false);
+  });
+});


KooshaPari:

PR #13509, opened 2026-09-13
fix(dashboard): resolve Next 16 Webpack node:fs imports and login page unmounted state warning

joesalaz:
# Fix(dashboard): resolve Next 16 Webpack node:fs imports and login page unmounted state warning

## Summary

- Resolves Next.js 16 Webpack `UnhandledSchemeError: Reading from "node:fs"` when client components (`ModelSelectModal.tsx`) transitively import provider model constants.
- Adds an `isMounted` guard to `src/app/login/page.tsx` to prevent unmounted state updates during Hot Module Replacement (HMR).
- Fixes `better-sqlite3` native dependency installation and `allowScripts` entries for Node 24 / npm 11.

## Related Issues

- Related to Next 16 / Node 24 dev server compatibility.

## Validation

- [x] Change type: UI / build-deploy
- [x] Focused tests and category gates from the golden path (`npm run lint`, `npm run typecheck:core`)
- [x] `npm run lint`

## Tests Added Or Updated

- `src/app/login/page.tsx` (updated with `isMounted` lifecycle guards)
- `open-sse/utils/cursorAgentCliVersion.ts` (wrapped Node `fs`/`os`/`path` imports for browser safety)

## Coverage Notes

- Changes touch `src/app/login/page.tsx`, `open-sse/utils/cursorAgentCliVersion.ts`, and `next.config.mjs`. All core types and lint gates pass cleanly (`npm run lint` and `npm run typecheck:core` both return 0 errors).

## Reviewer Notes

- Safe, non-breaking fix for Next.js 16 Webpack bundling under Node 24 / npm 11 development environments.

### DIFF
--- a/changelog.d/fixes/13300-next16-node24-webpack-hmr-fixes.md
+++ b/changelog.d/fixes/13300-next16-node24-webpack-hmr-fixes.md
@@ -1,0 +1,1 @@
+- **fix(dashboard):** resolve Next 16 Webpack `node:fs` scheme imports and login page unmounted state warning
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -375,6 +375,19 @@ const nextConfig = {
     "buffer",
     "util",
     "process",
+    "node:child_process",
+    "node:fs",
+    "node:path",
+    "node:os",
+    "node:crypto",
+    "node:net",
+    "node:tls",
+    "node:http",
+    "node:https",
+    "node:stream",
+    "node:buffer",
+    "node:util",
+    "node:process",
   ],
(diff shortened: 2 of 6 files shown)


diegosouzapw:
Thanks for digging into this — the underlying node:fs bug you found is real, but there's
already a much smaller, more targeted fix for it in #13604 (extracts just the pin constant,
no webpack config changes needed), so we'll go with that one for the bundling issue.
Your login-page unmounted-state fix looks independently useful, though — would you mind
splitting that into its own PR? It's unrelated to the build issue and easier to review (and
land) on its own. Same for the better-sqlite3/trustedDependencies package.json changes — those
look like they belong in a separate, dedicated PR with their own rationale.

diegosouzapw:

PR #13511, opened 2026-09-13
fix: extract resolveProviderAlias into pure module to fix client build

### LINKED ISSUE #13474: fix: release/v3.8.51 no longer builds: #13283 makes a "use client" module import the DB/playwright chain
### Summary

`release/v3.8.51` @ `152d95108` no longer produces a production build. `npm run build:release` fails with **168 Turbopack "Module not found" errors**, because a `"use client"` component now transitively reaches the DB → token-health → playwright/sharp chain through `src/lib/combos/controlCenter.ts`. Introduced by #13283 (`a5db19766`).

### Environment

- `release/v3.8.51` @ `152d95108` (reproduced on a fork rebased onto that tip)
- Node v26.7.0, Linux, Next 16.3.3 (Turbopack)
- Command: `unset NODE_ENV && npm run build:release`

### Repro

```bash
git fetch upstream release/v3.8.51
git switch -c tmp/build-check upstream/release/v3.8.51   # 152d95108
unset NODE_ENV && npm run build:release
```

### Actual

```
Error: Turbopack build failed with 168 errors:
./node_modules/playwright-core/lib/coreBundle.js:10133:26
Error: Module not found: Can't resolve 'async_hooks'

Import traces:
  Client Component Browser:
    ./node_modules/playwright-core/lib/coreBundle.js [Client Component Browser]
    ./node_modules/playwright/index.mjs
    ./open-sse/services/browserPool.ts
    ./open-sse/utils/chatgptWebExecutorAdapter.ts
    ./src/lib/providers/validation/chatgptWeb.ts
    ./src/lib/tokenHealthCheck.ts
    ./src/lib/db/settings.ts
    ./src/lib/db/providers.ts
    ./src/lib/db/models/activeSyncedCatalog.ts
    ./open-sse/services/model.ts
    ./src/app/(dashboard)/dashboard/combos/ComboControlCenterClient.tsx [Client Component Browser]
```

`async_hooks` and `child_process` are only the first two of the 168; the same trace also pulls `sharp`/`detect-libc` in via `open-sse/utils/cursorImages.ts`.

### Root cause

#13283 added this to `src/lib/combos/controlCenter.ts`:

```ts
```

`controlCenter.ts` is imported by the `"use client"` component `src/app/(dashboard)/dashboard/combos/ComboControlCenterClient.tsx`, and `open-sse/services/model.ts` pulls `open-sse/config/providerModels.ts` plus dynamic `@/lib/db/*` imports (`activeSyncedCatalog`, `readCache`, `db/models`). That chain reaches `runtimeSettings → tokenHealthCheck → playwright`/`sharp`, so Node built-ins and native modules end up in a browser chunk and Turbopack refuses to emit it.

### Control (proves it is the delta, not the environment)

- Parent commit `13bd76d38`: `unset NODE_ENV && npm run build:release` → exit 0, standalone bundle assembled (same worktree, same `node_modules`, same `.env`).
- Removing only that one import from `controlCenter.ts` (keeping the pure `resolveComboTargetModelStr(model)` part) clears the 168 errors, and the build then advances to a **second, different failure**:

```
Error [TurbopackInternalError]: Failed to write app endpoint /(dashboard)/dashboard/page
Caused by:
```

So there is at least one more client/server boundary leak on the dashboard route in the same range — worth treating as its own defect, but the `controlCenter.ts` one is deterministic and has a full import trace.

### Suggested fix

Keep provider-alias resolution out of client-reachable modules:

1. Extract `resolveProviderAlias` (plus the alias-table data it needs) into a dependency-free module, and import that from both `open-sse/services/model.ts` and `src/lib/combos/controlCenter.ts`; or
2. resolve the provider label on the server and pass it into the client component as a prop.

The label behaviour #13283 added should be kept — only the import edge needs to change.

Optionally: gate PRs targeting `release/**` on `npm run build:release`, since this class of regression currently reaches the release tip unnoticed.

### Impact

Standalone / Docker / self-host builds from `release/v3.8.51` cannot be produced at all; the fault is in the release branch, not in consumer configuration.

---
Reported from a downstream fork tracking `release/v3.8.51` (`BUILD_SHA` of the currently deployed artifact is `13bd76d38`); no changes are pushed upstream from that fork.

KooshaPari:
Fixes #13474

## Changes
- Created `open-sse/config/providerAlias.ts` — a pure module with no DB dependencies that exports `resolveProviderAlias` and `getAliasToProviderId`
- Updated `src/lib/combos/controlCenter.ts` to import from the pure module instead of `open-sse/services/model.ts`
- Updated `open-sse/services/model.ts` to re-export from the pure module (no code duplication)

## Problem
The `use client` component `ComboControlCenterClient.tsx` imports `controlCenter.ts`, which imported `resolveProviderAlias` from `open-sse/services/model.ts`. That module has dynamic DB imports (`activeSyncedCatalog`, `readCache`) that Turbopack bundles into the client chunk, pulling in the playwright/sharp chain and causing 168 'Module not found' errors (async_hooks, child_process, etc.).

## Root Cause
#13283 added the import of `resolveProviderAlias` from `model.ts` into `controlCenter.ts`. While `resolveProviderAlias` itself is pure, Turbopack bundles the entire `model.ts` module into the client chunk, including the dynamic DB imports.

## Testing
- The alias resolution logic is identical (same algorithm, same data source)
- All existing callers of `resolveProviderAlias` from `model.ts` continue to work via the re-export
- Client build should no longer pull in the DB/playwright chain through this path

### DIFF
--- a/open-sse/config/providerAlias.ts
+++ b/open-sse/config/providerAlias.ts
@@ -1,0 +1,58 @@
+/**
+ * Pure provider alias resolution — no DB, no server-only side effects.
+ *
+ * Extracted from open-sse/services/model.ts so that client-reachable modules
+ * (e.g. lib/combos/controlCenter.ts) can resolve provider aliases without
+ * pulling in the DB → tokenHealthCheck → playwright/sharp chain.
+ *
+ * #13474
+ */
+
+import { PROVIDER_ID_TO_ALIAS } from "../config/providerModels.ts";
+
+// Derive alias→provider mapping from the single source of truth (PROVIDER_ID_TO_ALIAS)
+// This prevents the two maps from drifting out of sync
+const ALIAS_TO_PROVIDER_ID: Record<string, string> = {};
+for (const [id, alias] of Object.entries(PROVIDER_ID_TO_ALIAS)) {
+  if (ALIAS_TO_PROVIDER_ID[alias]) {
+    // Duplicate alias — last writer wins (matches model.ts behaviour)
+  }
+  ALIAS_TO_PROVIDER_ID[alias] = id;
+}
+// Manual alias overrides — maps slug-style prefixes to canonical provider IDs.
+ALIAS_TO_PROVIDER_ID["opencode"] = "opencode-zen";
+ALIAS_TO_PROVIDER_ID["xiaomi"] = "xiaomi-mimo";
+ALIAS_TO_PROVIDER_ID["llamacpp"] = "llama-cpp";
+ALIAS_TO_PROVIDER_ID["agy"] = "antigravity";
+ALIAS_TO_PROVIDER_ID["aq"] = "amazon-q";
+
+/**
+ * Resolve provider alias to provider ID.
+ *
+ * Follows the alias chain transitively so intermediate alias-only hops resolve
+ * to the final target, but stops as soon as a hop lands on a registered
+ * provider id (#2901). Guarded against infinite loops with both a depth limit
+ * and a seen-set.
+ */
+export function resolveProviderAlias(aliasOrId: string | null | undefined): string | null {
+  if (typeof aliasOrId !== "string") return null;
+  let current = aliasOrId;
+  const seen = new Set<string>();
+  for (let i = 0; i < 10; i++) {
+    const next = ALIAS_TO_PROVIDER_ID[current];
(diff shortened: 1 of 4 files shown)


KooshaPari:

Issue #4559, opened 2026-06-21
[BUG] Truncate/ tool call too much


dev-cj:
### OmniRoute Version

3.8.31

### Installation Method

npm (global)

### Operating System

macOS

### Client Tool

pi coding agent

### Description

The compression truncate the tool call output too much

### Steps to Reproduce

This is the file i tried to read using pi agent
read-tool-issue.txt
```
# Read Tool / Terminal Output Filtering Issue
Observed: 2026-06-12

## Symptom
The `Read` tool, `cat`, `awk`, `sed`, `python3 print()`, and all shell output tools
silently suppress most lines from files that contain Unicode characters above U+007F
(e.g., ✅ U+2705, — U+2014, • U+2022, → U+2192).

Only lines containing ASCII-only content are passed through.
Lines that are entirely Unicode (e.g., table rows full of spaces + emoji) are dropped.

Example:
  File has 147 lines.
  Read tool returns 4 lines (only the ones that happen to have ASCII content).

## Affected tools
- Read (pi tool)
- Bash cat / awk / sed / iconv / strings / grep
- python3 print() even with ascii-safe encoding
- xxd DOES show the raw bytes correctly (not filtered)
- base64 -d piped to head DOES work (bypasses filter)

## Workaround that works
1. base64 encode the file, then decode in the same command:
   base64 -i <file> | base64 -d | head -N

2. Write file to /tmp first, then base64:
   cp <file> /tmp/copy.tsx && base64 -i /tmp/copy.tsx | base64 -d

3. xxd to inspect raw bytes.

## Hypothesis
The pi agent harness (or the terminal multiplexer/PTY layer) has a filter on stdout
that strips lines matching some pattern — possibly lines with high Unicode density,
or lines that start/end with certain byte sequences. This is NOT a file encoding issue
(confirmed UTF-8 valid via xxd + iconv). Likely a sanitisation pass in the tool output
pipeline.


this is line 40 here 
```

### Expected Behavior

The tool output should go as is.

### Actual Behavior

Attached screenshot of request using pi agent
here it is attached as *tool:*
so it compressed/truncated
<img width="740" height="1215" alt="Image" src="https://github.com/user-attachments/assets/cd443148-3f4d-4ad4-9a11-5912f4556230" />

when i used CLINE in vscode 
cline attached it as role: user and content, it was fine here

[...]

PR #13521, opened 2026-09-13
fix(compression): skip RTK dedup and truncation for non-shell tool results

### LINKED ISSUE #13388: fix(backend): RTK collapses file-content tool results, silently corrupting what the model receives
### OmniRoute Version

Server reached at a Tailscale-hosted deployment; `/api/compression/status` reports `strategy: "standard"`, engine `stacked`. Client side: OpenCode 1.18.30 via `@omniroute/opencode-plugin@0.2.1`.

### Installation Method

Self-hosted (existing deployment, reached over Tailscale)

### Operating System

macOS

### Description

With the engine set to `stacked` (pipeline `rtk` → `caveman`), tool results carrying **file contents** came back to the model structurally mangled, with RTK collapse markers inserted into the middle of the file:

```
[line repeated 2x]
[rtk:dropped 2 repeated lines]
```

Concretely: a 714-line JSONC config file read through a `read` tool arrived as roughly **4 usable lines** plus collapse markers. Repeated reads of the same file returned *different* amounts of content. I could not reliably read my own config through the agent until I disabled RTK.

Setting `engine: "caveman"` (leaving Caveman at `lite`, `outputMode: false`, everything else untouched) made the same read return all 714 lines verbatim, immediately and repeatably. Re-reading the identical path before/after the toggle is the cleanest A/B I have.

**Impact.** This is silent context corruption rather than compression: the model is handed something that looks like a complete file but is not, with no indication that content is missing. For a coding agent, acting on a silently truncated config is worse than an explicit error. It cost me several hours of misdiagnosis because the *tooling* looked fine — the corruption only exists in what reaches the model.

### What I could NOT determine — and where I'd appreciate your eyes

I initially assumed the engine-wide `deduplicateThreshold` (default `3`) described in `docs/compression/RTK_COMPRESSION.md`. **I checked, and that does not fit:**

- That knob requires runs of **≥3 consecutive identical** lines.
- The file has **zero** such runs (verified programmatically — longest consecutive identical run is 2).
- The markers I saw said `2x`, and the duplicate lines in this file (`},`, `"models": [`, `]`) are **non-consecutive**, separated by differing content.

So the mechanism is something other than the documented engine-wide dedup, and I don't want to guess further in a bug report. Candidates I could not rule out from outside the process: the per-filter `deduplicate` path in `lineFilter.ts` (including the "auto-enabled when the filter defines `collapsePatterns`" legacy behavior), or a filter matching this content that shouldn't.

I tried to narrow it with `POST /api/compression/preview`, which would have answered this directly, but that endpoint is currently unreachable for me — see the separate report I'm filing about `/v1/models`; the same deployment returns `HTTP 000` after a 90s hang, then `502`, on `/api/compression/preview` as well. If you can run the preview against the payload below, it should reproduce immediately.

### Steps to Reproduce

1. Set compression engine to `stacked` (`rtk` → `caveman`), RTK `intensity: standard`, `applyToToolResults: true` (defaults).
2. From a coding agent, read a large structured file — a 700+ line JSON/JSONC config with heavy structural repetition (`},`, `]`, repeated key names) reproduces it for me.
3. Inspect what the model actually receives.

Observed: content collapsed to a fraction of the file, with `[line repeated 2x]` / `[rtk:dropped 2 repeated lines]` markers inline. Repeated identical reads yield differing amounts of content.

Equivalent payload for `/api/compression/preview`:

```json
{ "mode": "rtk", "messages": [ { "role": "tool", "content": "<contents of a 700-line JSONC config>" } ] }
```

### Expected Behavior

RTK's own docs frame it as "OmniRoute's command-aware compression engine for **terminal and tool output** ... test logs, build output, package manager noise, shell transcripts". Collapsing repeated lines is correct for a progress bar or a repeated log line. It is **not** correct for source-file contents, where repeated lines are structurally meaningful and lossy collapse changes the meaning of the payload.

I'd suggest one of:

- Detect file-content payloads (or payloads that parse as structured data) and exclude them from line-collapse passes; or
- Scope collapse to output that actually matched a command-family filter, rather than applying it to any tool result; or
- At minimum, document that `applyToToolResults: true` can lossily alter file reads, so operators can make an informed choice.

The existing `codex-responses` engine config already carries a `preserveToolNames` list including `Read`/`read`/`Grep`/`grep`. An analogous safeguard for RTK would have prevented this entirely.

### Actual Behavior

File-content tool results are collapsed and interleaved with RTK markers before reaching the model, with no signal to the model or the user that content was removed. Disabling RTK (engine `caveman`) resolves it completely.

### Test Impact

This needs a regression test asserting that a structured file payload survives RTK byte-identically — analogous to the byte-identical assertion in the `probe-10897` test, but with RTK **enabled** rather than off. I'd suggest a fixture under `tests/unit/compression/fixtures/rtk/` containing a JSON/JSONC document with non-consecutive repeated structural lines, asserting no `[line repeated` / `[rtk:dropped` markers are introduced.

Happy to supply the exact 714-line file that triggers it if useful.

### Error Logs / Output

Live config at time of failure (from `/api/compression/status` and `set_compression_engine`):

```json
{
  "enabled": true,
  "strategy": "standard",
  "engines": { "rtk": { "enabled": true, "level": "standard" },
               "caveman": { "enabled": true, "level": "full" } },
  "stackedPipeline": [ { "engine": "rtk", "intensity": "standard" },
                       { "engine": "caveman", "intensity": "full" } ],
  "rtkConfig": {
    "intensity": "standard",
    "applyToToolResults": true,
    "applyToCodeBlocks": false,
    "maxLinesPerResult": 150,
    "maxCharsPerResult": 15000,
    "deduplicateThreshold": 3,
    "enableGrouping": false,
    "enabledFilters": [],
    "disabledFilters": []
  }
}
```

Note `enabledFilters`/`disabledFilters` are both empty, so this is the stock built-in catalog with no custom or project filters involved (`trustProjectFilters: false`).

### Additional Context

Related prior issues I read before filing, and why I believe this is distinct:

- **#10897** (closed, completed) — concluded correctly that `Default` mode *does* inherit the global off-toggle, and that RTK is not a separate always-on layer. My report is the opposite configuration: compression is deliberately **on** and RTK is *supposed* to run. The complaint is not that RTK ran when disabled, but that when legitimately enabled it corrupts file-content payloads.
- **#4268** (closed, completed) — stacked RTK+Caveman savings accounting.
- **#6473 / #6477** (closed) — RTK no-op on JSON; this is the inverse (RTK doing too much, not nothing).

Separately, `/api/compression/status` on this deployment reports `validationFallbacks: 21520` against `compressedRequests: 126452` (~17%). I don't know whether that's expected, but flagging it in case it's related to compressed payloads failing validation and being retried.

KooshaPari:
Fixes #13388

## Problem
RTK's line deduplication and smart truncation were applied to ALL tool results, including non-shell tools (read, grep, glob, edit, write). This collapsed structurally meaningful repeated lines in file content — e.g. JSON closing braces `}`, repeated key names `"models": [` — silently corrupting what the model received. A 714-line JSON config was reduced to ~4 lines plus `[line repeated 2x]` markers.

## Root Cause
In `processRtkText()`, the dedup and truncation steps ran unconditionally. The `skipFilters` flag (set for non-shell tools via `SHELL_TOOL_NAME_RE`) only skipped filter matching but not the subsequent dedup/truncation passes.

## Fix
- `deduplicateRepeatedLines()` is now skipped when `skipFilters` or `isDocumentLikeRead` is true
- `smartTruncate()` is now skipped when `skipFilters` is true (extending the existing `isDocumentLikeRead` guard)
- Shell command output (bash, terminal, execute_command) continues to be deduped and truncated as before

## Test Impact
- New test: `tests/unit/compression/rtk-file-content-preservation.test.ts`
- Tests: file content from read tool survives, no dedup markers, no truncation, shell output still deduped

### DIFF
--- a/docs/omniroute-pr-body.md
+++ b/docs/omniroute-pr-body.md
@@ -1,0 +1,68 @@
+## Summary
+
+`getActiveProvidersWithSyncedModel` in `src/lib/db/models.ts` only reads from `key_value(namespace='syncedAvailableModels')` and ignores `key_value(namespace='customModels')`, so models added through the provider dashboard's "model picker" are written to one store but read from another. Chat dispatch then returns `Model '…' is not available in the active live catalog for provider '…'` for perfectly valid models that the user just successfully added.
+
+## Motivation
+
+We hit this while moving 64 sub-agents × 16 chats to a `Main` combo that references `nvidia/deepseek-ai/deepseek-v4-pro-0813`. The model is fully in the live catalog (verified via `GET /api/providers/nvidia/models`) and is added to combos via the dashboard model picker, yet the dispatch path returns `400 invalid_request_error` and excludes the provider with `terminalReason: Model 'deepseek-ai/deepseek-v4-pro-0813' is not available in the active live catalog for provider 'nvidia'`. The only workaround is to re-run `POST /api/providers/{id}/sync-models`, which rewrites `syncedAvailableModels` from the upstream catalog — but that is destructive (it dropped 78 non-free models the first time we ran it) and is not what the user did.
+
+Steps to reproduce:
+1. Add any model through the provider dashboard "model picker" (it gets stored in `key_value(namespace='customModels')`)
+2. Add that model to a combo (or reference it via `provider/model`)
+3. Send a chat request through the combo
+4. Observe `Model '…' is not available in the active live catalog` even though the model is in the live catalog
+
+Related: #12298 (stale dashboard data), #12168 (provider "unhealthy" status bugs), #12073, #11804.
+
+## Proposed Fix
+
+In `src/lib/db/models.ts:528-552` (`getActiveProvidersWithSyncedModel` and its callers), union the `customModels` namespace with `syncedAvailableModels` when building the per-provider live model set. Specifically:
+
+```ts
+// pseudocode for the fix
+const syncedRows = await keyValueGetAll("syncedAvailableModels");        // current behaviour
+const customRows = await keyValueGetAll("customModels");                 // NEW
+const merged = new Map<string, Set<string>>();
+for (const row of [...syncedRows, ...customRows]) {
+  const set = merged.get(row.provider) ?? new Set<string>();
+  for (const m of parseModels(row.value)) set.add(m);
+  merged.set(row.provider, set);
+}
+return merged;
+```
+
+The dashboard write path (`useProviderModels.ts`, `managedModelImport.ts`, `POST /api/provider-models`) already keeps `customModels` accurate, so no write-side changes are needed.
+
+Optional follow-up: surface `customModels` separately in the dashboard so users can see why a model is in scope (e.g. "pinned" vs "discovered via sync"), and add a "resync" button that does **not** drop pinned models.
+
+## Testing
+
+```
+$ sqlite3 storage.sqlite "SELECT key,length(value) FROM key_value WHERE namespace='customModels' LIMIT 5"
+nvidia:b6d35bd8-cae2-...   842
+nvidia:key-2               842
+...
+$ sqlite3 storage.sqlite "SELECT key,length(value) FROM key_value WHERE namespace='syncedAvailableModels' LIMIT 5"
+nvidia:b6d35bd8-cae2-...   12344  # 82 models
+
+$ curl -X POST http://100.96.135.160:20128/v1/chat/completions \
+       -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
+       -d '{"model":"nvidia/deepseek-ai/deepseek-v4-pro-0813","messages":[{"role":"user","content":"hi"}]}'
+# BEFORE: {"error":{"message":"Model '...' is not available in the active live catalog for provider 'nvidia' (HTTP 400)"}}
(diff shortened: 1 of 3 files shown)


diegosouzapw:
Thanks for #13388. Holding for scope.

`skipFilters` is true for every non-shell tool result, so this also disables the generic line/char cap for grep, glob and search results. Those are exactly the outputs that can run to thousands of lines, and #4559 deliberately exempted only document-like reads.

The branch also carries an unrelated commit ("union customModels with syncedAvailableModels") and `docs/omniroute-pr-body.md`.

Could you narrow it to skipping dedup (not truncation) for non-shell results, drop the extra commit and file, and keep the regression test?

diegosouzapw:
Thanks for #13388 — the dedup-corrupting-JSON diagnosis is spot on. `skipFilters` is true for
every non-shell tool result though, so this change also disables the generic line/char cap for
grep, glob and search output — exactly the outputs that can run to thousands of lines, and #4559
deliberately exempted only document-like reads from that cap. The branch also carries an
unrelated commit ("union customModels with syncedAvailableModels") and `docs/omniroute-pr-body.md`.
Could you narrow this to skipping dedup only (not truncation) for non-shell results, drop the
extra commit and file, and keep the regression test?

PR #13522, opened 2026-09-13
fix(compression): log warnings for unreadable settings rows

### LINKED ISSUE #13456: fix(backend): compression settings has overlapping rows and silently falls back on an unreadable engines row
### OmniRoute Version

### Installation Method
Built from source (installed as npm global tarball)

### Operating System
Linux (Rocky Linux 10.2), Node.js v22.23.2

### Provider(s) / Model(s) Involved
Provider-independent (compression settings resolution).

### Client Tool
Any.

### Description
The effective compression pipeline is derived from **several overlapping settings rows** in `key_value` (`namespace = 'compression'`), and which one wins is decided by internal predicates that are never surfaced to the operator:

| stored row | used when | silently ignored when |
|---|---|---|
| `engines` (per-engine `enabled` + `level`) | `enginesExplicit === true` | row missing / not a string / not valid JSON |
| `stackedPipeline` (array with `intensity`) | `enginesExplicit === false`, or explicit map does not derive a stacked plan | the engines map derives its own pipeline |
| `defaultMode` | `enginesExplicit === false` | engines map is explicit |
| legacy per-engine rows (`ultra`, `caveman`, …) | legacy installs | engines map is explicit |
| `cavemanConfig.intensity` | legacy `standard` mode | stacked mode — the step intensity comes from `engines.caveman.level` |
| seeded combo `default-caveman` (rtk standard + caveman **full**) | stacked + built-in pipeline + `!enginesMapDerivesStackedPipeline` + `!activeComboResolves` | — |

Two consequences:

1. **A single unreadable `engines` row flips the whole dispatch model without any warning.** `getCompressionSettings()` does `const rawValue = typeof record.value === "string" ? record.value : null; if (!key || rawValue === null) continue;` and `parseJsonSafe` returns `undefined` on bad JSON → `continue`. Nothing is logged. `config.enginesExplicit = storedEngines !== null` becomes `false`, the panel-configured engines map is dropped, and the request falls through to `defaultMode` + `stackedPipeline` + the seeded `default-caveman` combo — i.e. caveman at **full** intensity even though the operator had set `lite`. The only trace is a `log.debug("COMPRESSION", "Default compression combo applied: default-caveman")`, which is invisible at the default log level.
   A row becomes non-string very easily: anything that writes the value as a BLOB (`sqlite3 … readfile()`, some backup/restore or migration tooling, a manual `UPDATE`). The panel still reads the row back fine (it is valid JSON bytes), so the UI shows the intended config while the runtime runs a different one.
2. **Stored fields that are not the winner are still displayed and editable.** An operator who changes `stackedPipeline` or `cavemanConfig.intensity` (the fields the docs and the settings API expose) sees no effect when the engines map is explicit, and vice-versa, and there is no log line, API field or panel hint saying "this field is currently shadowed by X".

Because the settings TTL cache is 5 s and there is no per-request "effective pipeline + source" line, an operator cannot tell from the outside which of the six rows produced the pipeline that actually ran. We only found out by copying the DB and calling `getCompressionSettings()` / `selectCompressionPlan()` offline.

### Steps to Reproduce
1. In the panel, set the engines map to something distinctive (e.g. `caveman.level = "lite"`, `ultra` off) and set `cavemanConfig.intensity = "lite"`.
2. Re-store the `engines` row as a BLOB with identical bytes:
   ```bash
   sqlite3 "$DATA_DIR/storage.sqlite" "select value from key_value where namespace='compression' and key='engines'" > /tmp/engines.json
   sqlite3 "$DATA_DIR/storage.sqlite" "update key_value set value = readfile('/tmp/engines.json') where namespace='compression' and key='engines'"
   sqlite3 "$DATA_DIR/storage.sqlite" "select typeof(value) from key_value where namespace='compression' and key='engines'"   # blob
   ```
3. Offline check (repo root):
   ```ts
   // DATA_DIR=... npx tsx repro-config.mts
   import { getCompressionSettings } from "./src/lib/db/compression.ts";
   import { listCompressionCombos } from "./src/lib/db/compressionCombos.ts";
   import { selectCompressionPlan, buildNamedComboLookup, enginesMapDerivesStackedPipeline } from "./open-sse/services/compression/strategySelector.ts";

   const config = await getCompressionSettings();
   console.log("enginesExplicit:", config.enginesExplicit, "derives:", enginesMapDerivesStackedPipeline(config));
   const plan = selectCompressionPlan(config, null, 10000, { messages: [] }, { provider: "openai", targetFormat: "openai", model: "x" } as any, buildNamedComboLookup(listCompressionCombos()), null);
   console.log(JSON.stringify(plan));
   ```
4. Send a request through the proxy and compare `compression_analytics` / the app log with the panel.

### Expected Behavior
- One source of truth for the pipeline (the engines map, or a materialised `stackedPipeline` — but not both plus legacy rows plus a seeded combo), with the other fields either removed, migrated on write, or clearly marked read-only/derived in the API and panel.
- A stored row that cannot be used (non-string, invalid JSON, unknown engine id) is logged at **warn** with the key name, instead of being treated as "absent".
- When the seeded default combo or the legacy path is chosen, that is logged at info (once, on settings load / settings change), including the reason (`enginesExplicit=false`, `activeComboId` unresolved, …).
- The effective pipeline and its source (`engines-map` / `stacked-pipeline` / `active-combo` / `default-combo` / `header` / `auto-trigger`) are exposed on the settings API and in the per-request analytics row, so an operator can verify what runs without reading the DB offline.

### Actual Behavior
- Step 2 → `enginesExplicit: false`, `derives: false`; `selectCompressionPlan` returns the legacy `stackedPipeline` and chatCore additionally applies `default-caveman` (rtk standard + caveman full). No log line above debug.
- The panel and `GET` settings API keep showing the engines map as configured.
- `cavemanConfig.intensity` and `stackedPipeline[*].intensity` are shown as editable but have no effect while the engines map is explicit.

### Test Impact
Needs a new unit test

### Additional Context
- `src/lib/db/compression.ts`: row loop (`typeof record.value === "string"` guard, `parseJsonSafe` → `continue`), `config.enginesExplicit = storedEngines !== null`.
- `open-sse/handlers/chatCore.ts` (~1567): seeded default combo condition (`isBuiltinStackedPipeline && !enginesMapDerivesStackedPipeline && !activeComboResolves`), `log.debug` only.
- `open-sse/services/compression/strategySelector.ts`: `resolveBasePlan` / `deriveDefaultPlanFromConfig`, `enginesMapDerivesStackedPipeline` (`if (!config.enginesExplicit) return false`).
- `src/lib/db/compressionCombos.ts`: `defaultCompressionComboPipeline()` = rtk standard + caveman full.

### Validation Plan
- Unit test: a `key_value` fixture with a BLOB/invalid `engines` row must produce a warn log and must **not** silently change `enginesExplicit`/the derived plan compared to the same fixture with a text row (or must fail loudly).
- Unit test: settings API response carries `effectivePipeline` + `source`.
- `node --import tsx/esm --test tests/unit/compression/derive-default-plan.test.ts tests/unit/compression/derived-pipeline-integration.test.ts`

KooshaPari:
Fixes #13456

## Problem
`getCompressionSettings()` silently skipped non-string (BLOB) and invalid-JSON settings rows. When a row became a BLOB (via backup/restore, migration tooling, or manual SQL), the engines map was silently dropped and the runtime fell through to legacy defaults — with no log line above debug level.

## Fix
- Added warn-level logging for non-string (BLOB) values with the key name and remediation hint
- Added warn-level logging for invalid-JSON values
- Added warn-level logging when the 'engines' row exists but yields no valid toggles (empty/unparseable)
- All warnings suggest re-saving from the Storage panel as remediation

## Impact
Operators can now diagnose config drift between the panel and the runtime by checking console warnings at startup.


diegosouzapw:
Thanks for #13456. Holding. The warnings themselves look useful, but:

- the test only checks that a stubbed `console.warn` records a message and never calls `getCompressionSettings()`, so it doesn't cover the change;
- the `storedEngines === null` warning also fires for a legitimately empty engines map;
- the branch carries ~2.6k unrelated commits plus an `apps/web/src/lib/bff-origin.ts` change.

A clean branch with a test that seeds a BLOB row and an invalid-JSON row, then asserts the warnings, would be easy to merge.

diegosouzapw:
Thanks for #13456 — the warnings themselves are a good idea. A few things before this can land:
the test only checks that a stubbed `console.warn` records a message and never actually calls
`getCompressionSettings()`, so it doesn't cover the change; the `storedEngines === null` warning
also fires for a legitimately empty (but valid) engines map, since `parseStoredEnginesMap`
returns null in both cases; and the branch carries ~3k unrelated commits plus an
`apps/web/src/lib/bff-origin.ts` change in the same commit as the fix. A clean cherry-pick of
just the compression.ts change, with a test that seeds a BLOB row and an invalid-JSON row and
asserts the warnings (and doesn't warn on a valid empty engines map), would be easy to merge.

PR #13529, opened 2026-09-13
fix(auth): skip /v1/models auth when no management credentials exist

### LINKED ISSUE #13354: fix(auth): #9320 regressed the keyless local-first posture: /v1/models 401s on pre-existing keyless installs (even loopback)
## #9320 regressed the keyless local-first posture on pre-existing installs

**Commit:** b07182c72acdf12bb6ce151eae7bea23de1b26fb — `fix(security): require auth for /v1/models when management auth is configured (#9320)`

### What broke

Before #9320, `getModelCatalogAuthRejection` was opt-in:

```ts
// v3.8.49
if (settings.requireAuthForModels !== true || !(await isAuthRequired(request))) return null;
```

After #9320 it's opt-out:

```ts
// v3.8.51 line
const authRequired = await isAuthRequired(request);
if (!authRequired) return null;
if (settings.requireAuthForModels === false) return null;
```

For a **pre-existing keyless install** (no `api_keys` rows, no management password, no `INITIAL_PASSWORD`), `isAuthRequired()` still returns `true` because of this branch in `src/shared/utils/apiAuth.ts`:

```ts
return settings.setupComplete === true || !isLoopbackRequest(request);
```

`setupComplete === true` on any install that completed onboarding — so **every request, even from loopback, now 401s** on a DB that was deliberately configured keyless. Observed in production after upgrading v3.8.49 → a v3.8.51-line build:

```
GET /v1/models → 401 {"error":{"message":"Authentication required","type":"invalid_api_key","code":"invalid_api_key"}}
```

(the commit title says "when management auth is configured" — but on these installs *no* management auth is configured; `setupComplete` alone trips the gate)

This also contradicts the documented local-first default (e.g. the `serve.mjs` / `a2a/_auth.ts` comments describing the keyless local-first posture, and `isValidApiKey`'s "optional — for local use can skip" doc).

### Repro

1. Fresh install, complete onboarding, configure **no** password and **no** API keys (keyless local-first usage).
2. Upgrade to any build containing b07182c72 (v3.8.51 line).
3. `curl http://127.0.0.1:20128/v1/models` → 401. `/api/health` stays 200, so healthchecks don't catch it.

### Suggested fix direction

The gate's intent (per the commit title) is "require auth **when management auth is configured**". So the opt-out escape should not hinge solely on the `requireAuthForModels` setting — `isAuthRequired()` returning true merely because `setupComplete` is set (with no password/OIDC/env key configured anywhere) is not "management auth is configured". Minimal options:

- In `getModelCatalogAuthRejection`, additionally skip when no credential surface exists at all (no password hash, no OIDC, no `INITIAL_PASSWORD`, no `api_keys` rows, no `oma_` tokens) — i.e. gate on *actual* configured auth, matching the commit's stated intent.
- Or make `isAuthRequired`'s `setupComplete === true || !isLoopback` branch not count as auth-configured for client API routes.

### Workarounds (verified on the v3.8.51 line)

- `OMNIROUTE_API_KEY=***` env passthrough key (#1350) + `Authorization: Bearer *** key>` from clients — works, but changes the zero-config local story.
- `requireAuthForModels=false` in the settings table (namespace `settings`) — restores v3.8.49 behavior.

---

**Note:** the same flip applies wherever `getModelCatalogAuthRejection`-style opt-out gates were introduced; `AUTH_001` on management routes vs `invalid_api_key` on client routes made this look like two bugs during triage.

KooshaPari:
Fixes #13354

## Changes
- Added keyless local-first guard in `getModelCatalogAuthRejection`
- When no management auth is configured (no password, no OIDC, no INITIAL_PASSWORD, no API keys in DB), auth is skipped for `/v1/models`
- Preserves the pre-#9320 posture for installs that deliberately chose not to set up auth
- Still requires auth when management credentials exist (password, OIDC, API keys)

## Root Cause
After #9320, `getModelCatalogAuthRejection` was changed from opt-in (`requireAuthForModels === true`) to opt-out (`requireAuthForModels === false`). For keyless installs where `setupComplete === true` but no password is configured, `isAuthRequired()` returns true, causing 401 on every `/v1/models` request.

## Testing
- Keyless install (no password, no API keys): `GET /v1/models` now returns 200
- Install with password configured: `GET /v1/models` still requires auth
- Install with API keys: `GET /v1/models` still requires auth

### DIFF
--- a/src/app/api/v1/models/catalogRequest.ts
+++ b/src/app/api/v1/models/catalogRequest.ts
@@ -18,6 +18,21 @@ export async function getModelCatalogAuthRejection(
   if (!authRequired) return null;
   if (settings.requireAuthForModels === false) return null;
 
+  // Keyless local-first guard (#13354): when no management auth is configured
+  // (no password, no OIDC, no INITIAL_PASSWORD, and no API keys in the DB),
+  // skip auth for /v1/models. This preserves the pre-#9320 posture for
+  // installs that deliberately chose not to set up any auth, while still
+  // requiring auth when management credentials exist.
+  const hasPassword =
+    typeof settings.password === "string" && settings.password.length > 0;
+  const hasOidc = settings.oidcEnabled === true;
+  const hasInitialPassword = !!process.env.INITIAL_PASSWORD;
+  if (!hasPassword && !hasOidc && !hasInitialPassword) {
+    const { getApiKeys } = await import("@/lib/db/apiKeys");
+    const keys = await getApiKeys(1);
+    if (keys.length === 0) return null;
+  }
+
   const apiKey = extractApiKey(request);
   if (apiKey) {
     if (await validateCatalogApiKey(apiKey)) return null;


diegosouzapw:
Thanks for #13354. Holding after maintainer review: this loosens authentication, so it needs a product decision rather than a code fix.

`getModelCatalogAuthRejection` only runs when `isAuthRequired()` already decided auth is required. Skipping it whenever the DB holds no API keys would expose `/v1/models` on installs that deliberately require auth but haven't created a key yet. There is also no test covering the new branch. We'll follow up on the issue.

diegosouzapw:
Thanks for #13354 — following up after maintainer review: this loosens authentication, so it
needs a product decision rather than a code fix. `getModelCatalogAuthRejection` only runs when
`isAuthRequired()` has already decided auth is required; skipping it whenever the DB holds no
API keys yet would expose `/v1/models` on installs that deliberately require auth but haven't
created a key yet. There's also no test covering the new branch. We'll follow up on the issue
with a design that can tell "auth never configured" apart from "auth configured, no key yet"
before implementing anything here.

diegosouzapw:
O problema que você aponta é real — a #13354 continua aberta e uma instalação keyless que concluiu o onboarding leva 401 até em loopback, porque `isAuthRequired` (`src/shared/utils/apiAuth.ts:443`) termina o ramo keyless com `return settings.setupComplete === true || !loopback`.

Mas o fix desta PR não pode entrar: **ele descarta a restrição de loopback**.

O guard proposto é:

```js
if (!hasPassword && !hasOidc && !hasInitialPassword) {
  const keys = await getApiKeys(1);
  if (keys.length === 0) return null;
}
```

Não há checagem de loopback em nenhum ponto. O gate que ele contorna condiciona os **dois** disjuntos, e o comentário imediatamente acima dele em `apiAuth.ts:432-441` existe por causa da **GHSA-7pq4-8pvv-rx7r**, registrando que esse caminho precisa honrar a restrição. Com esta mudança, numa instalação keyless exposta por túnel ou LAN, `GET /v1/models` passa a ser legível por qualquer par da rede — inventário completo de providers, modelos e conexões.

Dois pontos menores: `getApiKeys(1)` tem efeito colateral (`setNoLog()` em `apiKeys.ts:509-511`) e hidrata ~20 campos de policy; `getApiKeysCount()` (`apiKeys.ts:516`) faz o que se quer aqui. E a decisão fica re-litigada num segundo lugar, então `/v1/models` passa a discordar das outras rotas sobre o que é "instalação keyless".

Fechando. A #13354 fica aberta: o caminho certo é um flag dedicado de "auth nunca configurada" dentro de `isAuthRequired()`/`setupComplete`, e qualquer relaxamento precisa preservar o loopback. Obrigado por levantar o caso.

PR #13536, opened 2026-09-13
fix: add CJK quota-exhaustion patterns to 429 classifier

### LINKED ISSUE #13194: fix(resilience): 429 classifier misses non-English quota-exhaustion messages — CJK providers (z.ai/GLM, Kimi, Qwen, MiniMax) stuck in a short rate-limit retry loop

KooshaPari:
Fixes #13194

## Changes
- Added CJK (Chinese) quota-exhaustion patterns to both `QUOTA_PATTERNS` and `TERMINAL_QUOTA_PATTERNS` in `classify429.ts`
- Covers z.ai/GLM, Kimi/Moonshot, Qwen/DashScope, and MiniMax error messages
- Patterns match: 使用上限, 额度已用尽, 额度已用完, 调用上限, 已达上限, 余额不足

## Root Cause
CJK providers return Chinese error bodies for quota exhaustion that none of the English-only patterns matched, causing the 429 classifier to treat long-window quota exhaustion as transient rate_limit (60s cooldown) instead of quota_exhausted (long cooldown + failover).

### DIFF
--- a/open-sse/executors/zed-hosted.ts
+++ b/open-sse/executors/zed-hosted.ts
@@ -467,7 +467,7 @@ export class ZedHostedExecutor extends BaseExecutor {
       thread_id: bodyRecord.thread_id || (credentials as Record<string, unknown>)?._clientSessionId,
       prompt_id: bodyRecord.prompt_id,
       provider,
-      model,
+      model: bareModel,
       provider_request: providerRequest,
     };
 
@@ -496,7 +496,7 @@ export class ZedHostedExecutor extends BaseExecutor {
     const suppressThinkClose = resolveZedSuppressThinkClose(clientHeaders, clientResponseFormat);
 
     const wrapped = response.ok
-      ? wrapZedCompletionStream(response, provider, model, { suppressThinkClose })
+      ? wrapZedCompletionStream(response, provider, bareModel, { suppressThinkClose })
       : response;
     return {
       response: wrapped,
--- a/src/shared/utils/classify429.ts
+++ b/src/shared/utils/classify429.ts
@@ -99,6 +99,17 @@ const QUOTA_PATTERNS: ReadonlyArray<RegExp> = [
   /organization TPD rate limit/i,
   /\bTPD rate limit\b/i,
   /insufficient balance/i,
+
+  // #13194 — CJK providers (z.ai/GLM, Kimi, Qwen, MiniMax) return Chinese
+  // quota-exhaustion messages that none of the English patterns above match.
+  // Without these, the gateway misclassifies them as transient rate_limit
+  // and retries a dead model every 60s instead of failing over.
+  /使用上限/,
+  /额度已用尽/,
+  /额度已用完/,
+  /调用上限/,
+  /已达上限/,
+  /余额不足/,
 ];
 
 /**
@@ -166,6 +177,13 @@ const TERMINAL_QUOTA_PATTERNS: ReadonlyArray<RegExp> = [
   /organization TPD rate limit/i,
   /\bTPD rate limit\b/i,
   /insufficient balance/i,
+  // #13194 — CJK equivalents for terminal quota signals.
+  /使用上限/,
+  /额度已用尽/,
+  /额度已用完/,
+  /调用上限/,
+  /已达上限/,
+  /余额不足/,
 ];
 
 /**


diegosouzapw:
Thanks! #13412, also from you, covers #13194 with a broader pattern set and tests, and it is being merged. This branch also carries the #13537 zed commit, which does not compile.

It also adds the CJK patterns to `TERMINAL_QUOTA_PATTERNS`. That would make GLM's recoverable 5-hour window ("已达到 5 小时的使用上限…将在…重置") terminal instead of a cooldown. Closing as superseded by #13412.

PR #13537, opened 2026-09-13
fix(zed-hosted): strip provider prefix to prevent max_tokens inflation

### LINKED ISSUE #13364: fix(providers): zed-hosted claude-haiku-4-5 thinking inflates max_tokens to 163072
### OmniRoute Version
3.8.50 (BUILD_SHA dea6bb8)

### Installation Method
npm (global)

### Operating System / Node.js
Linux (Ubuntu VPS), Node v22.23.1

### Description
`zed-hosted/claude-haiku-4-5` fails whenever the client enables extended thinking with tools present (Claude Code as an interactive session does both). Anthropic, through Zed's proxy, answers:

```
max_tokens: 163072 > 64000, which is the maximum allowed number of output tokens for claude-haiku-4-5-20251001
```

The client asked for `max_tokens: 16000` and `thinking.budget_tokens: 15999`. The gateway sent `max_tokens: 163072` and `budget_tokens: 131072`.

Two problems in one:
1. The request is inflated by the translation chain.
2. Because Zed reports this failure as an in-stream `status: failed` frame after HTTP 200, 3.8.50 renders the error as assistant text (`[Zed error] max_tokens: ...`) with a normal stop, so neither combo fallback nor account rotation reacts. #12455 (merged, unreleased) addresses the second part; this issue is about the first.

### Root cause (read in 3.8.50 sources)
- `claude-to-openai` maps `budget_tokens: 15999` to `reasoning_effort: "high"` (anything above 10240 is `high`).
- The `zed-hosted` executor translates back with `openaiToClaudeRequest`, whose `effortBudgetMap` turns `high` into `budget_tokens: 131072` (`open-sse/translator/request/openai-to-claude.ts` ~line 250).
- `adjustMaxTokens` raises `max_tokens` to the 32000 floor because tools are present.
- `fitThinkingToMaxTokens` has no known output cap for `zed-hosted/claude-haiku-4-5` (the Zed catalog is passthrough, `model_capabilities` has no row), so instead of fitting the budget under the cap it adds them: 32000 + 131072 = 163072. The Sonnet models on the same provider work only because their cap is known.

### Reproduction
Interactive Claude Code session with `--model zed-hosted/claude-haiku-4-5`, any prompt. With `curl` it depends on cache state; the interactive path fails every time.

### Proposed fix
Either register output caps for the Zed passthrough catalog (Haiku 4.5 = 64000) so `fitThinkingToMaxTokens` can fit the budget, or, in the `zed-hosted` executor, clamp `max_tokens` to the model cap and keep `thinking.budget_tokens < max_tokens` for the `anthropic` provider. Verified against the live proxy: clamping to 64000 with the budget fitted below it makes the interactive session work. Independently of that, mapping every budget above 10240 to a 131072 `high` budget is a large silent cost increase for clients that asked for 16k of thinking.

KooshaPari:
Fixes #13364. Strips provider prefix from model before passing to translators so capability lookups resolve correctly.

### DIFF
--- a/open-sse/executors/zed-hosted.ts
+++ b/open-sse/executors/zed-hosted.ts
@@ -467,7 +467,7 @@ export class ZedHostedExecutor extends BaseExecutor {
       thread_id: bodyRecord.thread_id || (credentials as Record<string, unknown>)?._clientSessionId,
       prompt_id: bodyRecord.prompt_id,
       provider,
-      model,
+      model: bareModel,
       provider_request: providerRequest,
     };
 
@@ -496,7 +496,7 @@ export class ZedHostedExecutor extends BaseExecutor {
     const suppressThinkClose = resolveZedSuppressThinkClose(clientHeaders, clientResponseFormat);
 
     const wrapped = response.ok
-      ? wrapZedCompletionStream(response, provider, model, { suppressThinkClose })
+      ? wrapZedCompletionStream(response, provider, bareModel, { suppressThinkClose })
       : response;
     return {
       response: wrapped,


diegosouzapw:
Thanks for looking into the zed max_tokens inflation (#13364). On the release branch `bareModel` is not defined in `ZedHostedExecutor.execute()`, so this diff does not compile (TS2304) as submitted.

Happy to review a version that derives the bare model id in the same change and adds a test for the prefix stripping. Closing this one.

PR #13540, opened 2026-09-13
fix(db): prune backups after health-check-repair snapshot

### LINKED ISSUE #13308: fix(backend): health-check-repair backups are never pruned — the retention call is missing from that path only

KooshaPari:
Fixes #13308

## Problem
The health-check-repair backup path creates a full database snapshot via `VACUUM INTO` on every startup but never calls the existing retention helper (`cleanupDbBackups`). This causes `db_backups/` to grow without bound.

On one observed install this produced **10.3 GB across 18 snapshots in 8 days**, while the live database was 698 MB. A perfectly healthy database still produces a full copy on every startup.

## Root cause
`createManagedDbBackup()` in `src/lib/db/core.ts` creates the snapshot but has no pruning call after it. The regular `backupDbFile()` path in `backup.ts` does call `cleanupDbBackups`, but the health-check-repair path uses the separate `createManagedDbBackup` function which was missed.

## Fix
Call `cleanupDbBackups()` after a successful snapshot in `createManagedDbBackup`, so the operator's `maxFiles` / `retentionDays` settings apply uniformly to all backup paths.

A dynamic import of `./backup` avoids a circular dependency (since `backup.ts` imports from `core.ts`).

## Changes
- `src/lib/db/core.ts`: Add `cleanupDbBackups()` call after `VACUUM INTO` in `createManagedDbBackup()`

## Validation
- Verified that `cleanupDbBackups` is exported from `./backup` and accepts a `backupDir` option
- The dynamic import + `.catch(() => {})` ensures the backup still succeeds even if pruning fails
- No changes to backup creation logic, only post-creation cleanup

### DIFF
--- a/src/lib/db/core.ts
+++ b/src/lib/db/core.ts
@@ -883,6 +883,15 @@ function createManagedDbBackup(db: SqliteDatabase, reason: string): boolean {
 
     db.exec(`VACUUM INTO '${escapedBackupPath}'`);
     console.log(`[DB] Backup created (${reason}): ${backupPath}`);
+
+    // #13308: prune old backups after creating a new one so db_backups/ does not
+    // grow without bound. The health-check-repair path creates a snapshot on every
+    // startup even for a healthy database, and without pruning each restart adds a
+    // full-sized copy. Dynamic import avoids a circular dependency with backup.ts.
+    import("./backup")
+      .then(({ cleanupDbBackups }) => cleanupDbBackups({ backupDir }))
+      .catch(() => {});
+
     return true;
   } catch (error: unknown) {
     const message = error instanceof Error ? error.message : String(error);
--- a/src/lib/evals/evalRunner.ts
+++ b/src/lib/evals/evalRunner.ts
@@ -241,6 +241,7 @@ export function runSuite(
 
     if (metrics?.error && !result.error) {
       result.error = metrics.error;
+      result.passed = false; // #13137 — an upstream failure must not grade as passed
     }
 
     return result;
--- a/src/shared/utils/machineId.ts
+++ b/src/shared/utils/machineId.ts
@@ -57,9 +57,10 @@ function getMachineIdRaw(): string {
   }
 
   // Strategy 2: macOS — ioreg IOPlatformUUID
+  // Skip when DISABLE_IOREG_STRATEGY=1 so tests can reach Strategy 4/5 on darwin.
   try {
-    if (process.platform !== "darwin") {
-      throw new Error("Not macOS");
+    if (process.platform !== "darwin" || process.env.DISABLE_IOREG_STRATEGY === "1") {
+      throw new Error("Not macOS or ioreg disabled");
     }
     const output = execSync("ioreg -rd1 -c IOPlatformExpertDevice", {
       encoding: "utf8",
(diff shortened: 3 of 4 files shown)


diegosouzapw:
Thanks! #13404, also from you, fixes #13308 the same way with direct `backupRetention` imports and a test, and it is being merged. This branch uses a fire-and-forget dynamic import with a silent catch and also carries the #13539 and #13542 commits. Closing as a duplicate of #13404.

PR #13542, opened 2026-09-13
fix(eval): force passed=false when upstream call failed

### LINKED ISSUE #13137: fix(backend): Eval case whose model call failed is scored as passed — runSuite() never forces passed=false

KooshaPari:
Fixes #13137

## Changes
- Added `result.passed = false` when `metrics.error` is set in `runSuite()`

## Root Cause
When an upstream call failed, the error text was graded against the expected pattern. If the error text happened to match, the case was marked as passed, inflating the pass rate.

### DIFF
--- a/src/lib/evals/evalRunner.ts
+++ b/src/lib/evals/evalRunner.ts
@@ -241,6 +241,7 @@ export function runSuite(
 
     if (metrics?.error && !result.error) {
       result.error = metrics.error;
+      result.passed = false; // #13137 — an upstream failure must not grade as passed
     }
 
     return result;
--- a/src/shared/utils/machineId.ts
+++ b/src/shared/utils/machineId.ts
@@ -57,9 +57,10 @@ function getMachineIdRaw(): string {
   }
 
   // Strategy 2: macOS — ioreg IOPlatformUUID
+  // Skip when DISABLE_IOREG_STRATEGY=1 so tests can reach Strategy 4/5 on darwin.
   try {
-    if (process.platform !== "darwin") {
-      throw new Error("Not macOS");
+    if (process.platform !== "darwin" || process.env.DISABLE_IOREG_STRATEGY === "1") {
+      throw new Error("Not macOS or ioreg disabled");
     }
     const output = execSync("ioreg -rd1 -c IOPlatformExpertDevice", {
       encoding: "utf8",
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks! #13413, also from you, fixes #13137 (and #13138) with dedicated tests, and it is being merged. Closing this one as a duplicate. The `DISABLE_IOREG_STRATEGY` commit from this branch lands via #13539.

Issue #8247, opened 2026-07-23
fix(providers): cegp/gpt-5.6-luna dead — 403 insufficient_quota reset never fires, should mark model unhealthy


andrea-kingautomation:
### OmniRoute Version
3.8.48 (npm global)

### Provider(s) / Model(s)
cegp / gpt-5.6-luna

### Description
`cegp/gpt-5.6-luna` has been dead on our box since 2026-07-10:

- 2026-07-10 to 07-12: HTTP 502 (Cloudflare error page)
- 2026-07-16 onward: HTTP 403 `insufficient_quota`, with a "reset after 24s" hint that never actually resets (same error on every retry, days apart)

Our health prober has recorded 10+ consecutive failures and auto-flagged the model for delisting on our side. Other cegp models on the same connection are unaffected, so this looks like a dead upstream allocation for this specific model rather than a connection/auth problem.

### Ask
Fix the model upstream if it is meant to be alive, or delist it from the catalog so it stops surfacing as callable. A quota error that advertises a reset that never happens should probably also mark the model unhealthy in the router's own health tracking.

(Searched open and closed issues for cegp / gpt-5.6-luna before filing; the existing GPT-5.6 issues cover routing/Codex contract topics, not this dead-model case.)

Dingding-leo:
Quick diagnostic note: the 403 `insufficient_quota` with `reset after 24s` that never resets suggests the upstream is returning a static error page or a stale quota value. If other cegp models work on the same connection, this model's upstream allocation is likely dead (not refreshed by the provider).

Two actionable items:
1. **Short term**: mark `cegp/gpt-5.6-luna` as deprecated/hidden in the catalog so it stops surfacing as callable. The health prober's auto-flag should handle this but if it hasn't after 10+ consecutive failures, the auto-flag threshold or the per-model health tracking might need a look.
2. **Health tracking gap**: a model returning 403 with a fake reset timer should eventually graduate from "retryable" to "terminal" in the health system, same as `credits_exhausted`. The `reset after 24s` sugar-coating prevents that from happening.

The catalog paths to check for delisting:
- Provider registry under `open-sse/config/providers/registry/` if cegp has one
- Or `src/shared/constants/providers/` for the model catalog entry

Happy to submit the delist PR if someone confirms the model is indeed dead upstream.

diegosouzapw:
[...]

PR #13548, opened 2026-09-13
fix(auth): model-scoped lockout on passthrough provider credit exhaustion

ggdayup:
## Summary
When a provider with passthrough models (e.g., `bai` / `thebai`, third-party aggregators, or multi-model nodes) returns HTTP 400 with `credit insufficient balance` or quota exhaustion on a specific paid model, OmniRoute previously marked the entire connection as terminal `credits_exhausted` with `rateLimitedUntil = null`.

This disabled all free or available models on the same connection (e.g., `bai/qwen3.8-flash`) until a manual "Retest" was triggered or the 30-minute background reprobe timer elapsed.

This PR ensures credit exhaustion on per-model-quota providers is properly scoped as a model-level lockout (`reason: "quota_exhausted"`) rather than a whole-connection terminal failure.

## Root Cause
1. **`checkFallbackError()` in `open-sse/services/accountFallback.ts`**:
   PR #8247 added `!isCompatibleProvider(provider)` to prevent compatible providers from emitting `creditsExhausted: true`. However, `isCompatibleProvider` only matches `openai-compatible-` / `anthropic-compatible-` prefixes, omitting built-in passthrough providers like `bai`.
2. **`markAccountUnavailable()` in `src/sse/services/auth.ts`**:
   The model-lockout conditional check was restricted to `(status === 404 || isNvidiaModelGone || status === 429 || status >= 500)`. When an upstream passthrough provider returns HTTP 400 with `credit insufficient balance`, status 400 did not match this condition, falling through to `resolveTerminalConnectionStatus()` and terminating the entire account connection.

## Key Changes
1. **Scope Quota Signal to Per-Model Providers** (`open-sse/services/accountFallback.ts`):
   Replaced `!isCompatibleProvider(provider)` with `!hasPerModelQuota(provider, _model)` when returning `creditsExhausted: true`. Built-in passthrough providers, compatible providers, and multi-model configurations now avoid emitting connection-wide `creditsExhausted`. Account-level balance exhaustion (e.g. Moonshot empty wallet) continues to terminate the whole connection as expected.
2. **Handle Model Quota Exhaustion in `markAccountUnavailable`** (`src/sse/services/auth.ts`):
   Added `isModelLevelQuotaOrRateLimit` check so that non-permanent quota/credit exhaustion (`RateLimitReason.QUOTA_EXHAUSTED` or `creditsExhausted`) triggers model-only lockout with `reason: "quota_exhausted"`. Sibling models on the same connection remain active and eligible.
3. **Unit Tests** (`tests/unit/auth-passthrough-credit-lockout.test.ts`):
   - Verified `b.ai` HTTP 400 credit insufficient balance locks only the requested model (`glm-5.3-flash`), connection remains `active`, and free model (`qwen3.8-flash`) remains unlocked.
   - Verified compatible providers continue to lock only the model.
   - Verified single-model providers (e.g. `openai`) still terminate the connection with `credits_exhausted`.

## Testing
- Ran unit tests:
  ```bash
  node --import tsx/esm --test tests/unit/auth-passthrough-credit-lockout.test.ts tests/unit/auth-ollama-cloud-per-model-403-3027.test.ts tests/unit/account-fallback.test.ts tests/unit/auth-account-fallback.test.ts
  ```
  All passed cleanly.
- ESLint and Prettier checks passed with 0 errors.

### DIFF
--- a/open-sse/services/accountFallback.ts
+++ b/open-sse/services/accountFallback.ts
@@ -1956,18 +1956,16 @@ export function checkFallbackError(
       }
     }
 
-    // T10 (sub2api #1169) + #8247: credits/quota exhausted; *-compatible-* nicknames stay model-scoped
+    // T10 (sub2api #1169) + #8247: credits/quota exhausted; per-model-quota providers stay model-scoped
     // unless the body is an account-level Open Platform empty wallet.
-    if (
-      shouldUseQuotaSignal &&
-      isCreditsExhausted(errorStr) &&
-      (!isCompatibleProvider(provider) || isMoonshotAccountBalanceExhausted(errorStr))
-    ) {
+    if (shouldUseQuotaSignal && isCreditsExhausted(errorStr)) {
       return {
         shouldFallback: true,
         cooldownMs: COOLDOWN_MS.paymentRequired ?? 3600 * 1000, // 1h cooldown
         reason: RateLimitReason.QUOTA_EXHAUSTED,
-        creditsExhausted: true,
+        ...(!hasPerModelQuota(provider, _model) || isMoonshotAccountBalanceExhausted(errorStr)
+          ? { creditsExhausted: true }
+          : {}),
       };
     }
 
--- a/src/sse/services/auth.ts
+++ b/src/sse/services/auth.ts
@@ -2965,17 +2965,27 @@ export async function markAccountUnavailable(
       );
       return { shouldFallback: true, cooldownMs: lockout.cooldownMs };
     }
+    const isModelLevelQuotaOrRateLimit =
+      !fallbackResult.permanent &&
+      (fallbackResult.reason === RateLimitReason.QUOTA_EXHAUSTED ||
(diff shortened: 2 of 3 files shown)


diegosouzapw:
Thanks for tracking down the passthrough-vs-terminal scoping bug — the reproduction and
your three new tests (bai passthrough, compatible-nickname passthrough, and the
single-model `openai` regression guard) are solid and all pass against the real
`markAccountUnavailable()` path with a live SQLite DB. Before merging I'd like two small
additions: (1) a changelog fragment under `changelog.d/fixes/`, and (2) one more test
case for a `*-compatible-*` provider WITHOUT the connection-level `passthroughModels`
override, since removing the old `!isCompatibleProvider` guard changes the cooldown path
for that case too (fixed 1h instead of the prior exponential backoff) and it'd be good to
pin that as intentional. Nothing structural — happy to merge once those land.

PR #13551, opened 2026-09-13
fix(proxy): add combo scope to fail-closed proxy guard

### LINKED ISSUE #13469: fix(resilience): Fail-closed proxy guard does not cover the combo scope: a fully dead combo pool falls through to direct egress
### What happens

`hasBlockingProxyAssignment` in `src/lib/db/proxies/guards.ts:26-50` checks three scopes:

```sql
WHERE ((a.scope = 'account'  AND a.scope_id = ?)
    OR (a.scope = 'provider' AND a.scope_id = ?)
    OR (a.scope = 'global'))
  AND NOT ${PROXY_ALIVE_PREDICATE}
```

There is no branch for `a.scope = 'combo'`. But `combo` is a real scope: `src/lib/db/settings.ts:699` resolves it (`resolveProxyForScopeFromRegistry("combo", comboId)`, Step 7 at `settings.ts:683-718`), and the dashboard's *Set Proxy* modal writes assignments with `scope="combo"` (`src/app/(dashboard)/dashboard/combos/useComboProxyAssignments.ts:2-6`).

When every member of a combo-scoped pool is `dead`/`inactive`/`error`/`disabled`/`down`, `fetchAlivePoolRows` (`src/lib/db/proxies/rotation.ts:191-209`) filters them all out and resolution returns `{ proxy: null, level: "direct", levelId: null }` (`settings.ts:784`). The guard — the only caller is `src/sse/handlers/chatHelpers.ts:957` — finds no blocking assignment, and the request goes out **direct**.

### Why it matters

This is the state the health scheduler produces on its own with `PROXY_AUTO_DISABLE=true`, which the docs describe at `docs/ops/PROXY_GUIDE.md:788-798` — and that same passage promises the opposite:

> *the whole pool never silently falls back to a direct connection — see the 4-Level Proxy System fail-closed guard*

The guide lists Combo Proxy as level 3 of the four levels (`PROXY_GUIDE.md:94`), and the comment at `rotation.ts:220-222` repeats the promise: *"a dead pool never falls through to direct egress; the caller's guard blocks it"*.

So an operator who assigned proxies only at combo scope believes they are covered by a documented fail-closed guarantee, and the traffic leaves through the host's address instead.

### Suggested fix

Add the scope to the guard, alongside the combo id the resolution already knows:

```sql
WHERE ((a.scope = 'account'  AND a.scope_id = ?)
    OR (a.scope = 'provider' AND a.scope_id = ?)
    OR (a.scope = 'combo'    AND a.scope_id = ?)
    OR (a.scope = 'global'))
  AND NOT ${PROXY_ALIVE_PREDICATE}
```

Since the guard's contract is "the four levels", a test that asserts all four scopes block would keep the next scope from being added to the resolver and forgotten here.

Happy to send a PR.

KooshaPari:
Fixes #13469

## Problem

The `hasBlockingProxyAssignment` guard only checked `account`, `provider`, and `global` scopes. Combo-scoped proxy assignments were not checked, so a fully dead combo pool fell through to direct egress — leaking the host IP address despite the operator having configured combo-level proxies.

## Fix

Add the `combo` scope to the SQL guard query, alongside the existing three scopes. A dead combo pool now blocks egress like the other three scopes.

## Changes

- `src/lib/db/proxies/guards.ts`: Add combo scope to SQL query and optional `comboName` parameter

### DIFF
--- a/src/lib/db/proxies/guards.ts
+++ b/src/lib/db/proxies/guards.ts
@@ -23,7 +23,11 @@ export function isGlobalProxyEnabled(db: ReturnType<typeof getDbInstance>): bool
  * #6246 fail-closed guard for a connection with an assigned dead proxy pool.
  * Explicitly disabling proxying globally or for the connection allows direct egress.
  */
-export function hasBlockingProxyAssignment(connectionId: string, providerId?: string): boolean {
+export function hasBlockingProxyAssignment(
+  connectionId: string,
+  providerId?: string,
+  comboName?: string | null
+): boolean {
   try {
     const db = getDbInstance();
     if (!isGlobalProxyEnabled(db)) return false;
@@ -33,16 +37,19 @@ export function hasBlockingProxyAssignment(connectionId: string, providerId?: st
       .get(connectionId) as { provider?: string | null; proxy_enabled?: number } | undefined;
     if (conn && conn.proxy_enabled === 0) return false;
     const provider = conn?.provider ?? providerId ?? null;
+    // #13469: include combo scope — a fully dead combo-scoped pool must not
+    // fall through to direct egress.
     const dead = db
       .prepare(
         `SELECT 1 FROM proxy_assignments a JOIN proxy_registry p ON p.id = a.proxy_id
            WHERE ((a.scope = 'account' AND a.scope_id = ?)
                OR (a.scope = 'provider' AND a.scope_id = ?)
-               OR (a.scope = 'global'))
+               OR (a.scope = 'global')
+               OR (a.scope = 'combo' AND a.scope_id = ?))
              AND NOT ${PROXY_ALIVE_PREDICATE}
            LIMIT 1`
       )
-      .get(connectionId, provider);
+      .get(connectionId, provider, comboName ?? null);
     return !!dead;
   } catch {
     return false;


diegosouzapw:
Thanks for #13469. Holding. `hasBlockingProxyAssignment()` gains a `comboName` parameter, but its only caller (`src/sse/handlers/chatHelpers.ts:957`) still passes two arguments. The combo-scope clause therefore always binds `NULL` and never matches, so the fail-closed behavior doesn't change yet.

Could you thread the combo name through that call site and add a test that assigns a dead proxy at `scope = 'combo'` and asserts the request is blocked?

diegosouzapw:
Thanks for #13469 — the SQL guard change is correct in isolation, but `hasBlockingProxyAssignment()`
gains a `comboName` parameter that its only caller (`src/sse/handlers/chatHelpers.ts:957`) still
doesn't pass, so the combo-scope clause always binds NULL and never matches — the fail-closed
behavior for combo-scoped pools doesn't actually change yet. Could you thread the combo name
through that call site and add a test that assigns a dead proxy at `scope = 'combo'` and asserts
the request is blocked?

PR #13555, opened 2026-09-13
feat(dashboard): add a dedicated API-key routing editor

JxnLexn:
## Summary

Move routing-rule editing out of the cramped API-key permissions modal into a dedicated, full-width workspace at `/dashboard/api-manager/routing`.

- Select an API key or inspect rules across scopes; existing entry points link into this workspace.
- Separate the rule's conditions from its effects, with explicit model, combo and pattern choices.
- Search suggestions without silently rewriting existing or off-catalog IDs. Source combos retain their request-facing names; routing targets retain their existing IDs.
- Preserve advanced values, protect unsaved drafts when switching keys, keep failed saves editable, and disable writes after a failed configuration load.
- Include English/German copy and remove the extra 1,152-pixel page-width cap.

This PR changes the editor only. It does not change rule evaluation, authorization, request routing, or service-tier precedence.

## Related Issues

Historical base-red report #12732 is now closed. Current CI failures must be assessed on their own merits.

The feature was initially based on `release/v3.8.51` at `152d95108c9c3d557562311ffed63240a511eb31`. No fork deployment workflow, container configuration, admission hotfix, or provider-catalog patch is included.

## Validation

- [x] Change type: UI / i18n
- [x] `npx vitest run tests/unit/ui/api-key-routing-editor.test.tsx` — 6 passed, rerun on this branch.
- [x] `npx eslint tests/unit/ui/api-key-routing-editor.test.tsx` — passed.
- [x] Diff checked against the current release tip; `git diff --check` passed.
- [x] Automated regression tests included.
- [ ] Full repository lint, production build and coverage matrix: left to PR CI, not claimed green locally.

Earlier manual checks exercised save/cancel and key selection. The equivalent fork deployment served the full-width page with HTTP 200 and preserved the existing routing configuration; that deployment is separate from this PR's CI evidence.

## Tests Added Or Updated

- `tests/unit/ui/api-key-routing-editor.test.tsx`: persisted-field round-trip, combo-name matching, failed-save drafts, unsaved-change confirmation, failed-load write protection, and full-width wrapper.

## Coverage Notes

The focused component tests cover the changed editor contract. Full coverage was not measured locally; the repository coverage gate remains a CI responsibility.

## Reviewer Notes

No schema migration or provider credential changes. This branch can be reviewed independently from the runtime fixes. When combining with the per-key service-mode PR, retain its settings section in the selected-key workspace rather than restoring the old editor layout.

### Related independent PRs

[...]

PR #13448, opened 2026-09-12
feat: adaptive reasoning effort (auto) — gateway-resolved, per-turn pinned, all harnesses

patrykkopycinski:
## What

Opt-in adaptive reasoning effort at the gateway: `reasoning_effort: "auto"` (or an `X-OmniRoute-Effort: auto` header) is resolved to a concrete level from request-shape signals before the request goes upstream. One implementation covers every harness (Claude Code, Cursor, Codex, opencode, Hermes) with zero client changes.

## Why

Harnesses either pin a static effort (wastes tokens on trivial turns, under-thinks heavy ones) or omit it (model-side default). The gateway sees every request body regardless of harness, so it can pick the level per turn.

## How — two levers, both opt-in

- **Lever A**: `ModelSpec.defaultReasoningEffort: "auto"` — extends the #6879 default-injection seam. A literal `"auto"` left by the default injection is treated as an opt-in marker and resolved, never forwarded verbatim.
- **Lever B**: `X-OmniRoute-Effort: auto` request header (#6057 request-controls pattern) for per-request control without body changes.

Resolution lives in `open-sse/services/adaptiveEffort.ts`:

- Deterministic data signals, no LLM classifier: last-user-message boundary, user-message size, conversation context tokens (reuses `estimateMessageTokens`), tool-result volume, conversation depth.
- Bands: trivial ask → `low`, heavy turn → `high`, mid → `medium`; empty/degenerate body → `medium`.
- **Stateless per-turn pin**: signals are computed from the last user message and everything before it — tool results after the last user message are ignored — so every request in a turn resolves to the same level (no cache-busting effort flips mid-turn) with zero stored state.
- Explicit client effort always wins; `auto` never overrides a concrete value.

## Tests

- 9 new unit tests (`tests/unit/adaptive-effort.test.ts`): parse/`isAdaptiveEffort`, bands, pin invariance across tool-loop growth, explicit-wins, both levers.
- Mutation-proofed: breaking the pin, the empty→medium guard, and explicit-wins each fail ≥1 test; restored → green.
- Related suites green: `default-reasoning-effort-6879`, `vendor-default-thinking-effort`, `sync-reasoning-supported-efforts-7694` (55/55 total).
- `typecheck:core` clean, `check:cycles` clean, no new eslint errors (chatCore.ts carries 27 pre-existing on base — identical count verified).

## Docs

`docs/routing/AUTO-COMBO.md`: request-controls table row + usage section.

## Notes

[...]

PR #13556, opened 2026-09-13
fix(routing): preserve reasoning overrides across transports and fallbacks

JxnLexn:
## Summary

A matched reasoning rule could be lost during native request handling, model-suffix/account-default processing, or fallback preparation. Internal rule metadata could also escape into a native Responses payload and trigger an upstream `Unsupported parameter: _omnirouteReasoningRule` error.

- Apply effective reasoning directives in the appropriate request format, including native Responses and Anthropic requests.
- Carry the server-evaluated directive in request-local credential context through executor dispatch, refreshed credentials and fallback paths.
- Make forced effort win over model-suffix/account defaults; keep default-only rules conditional on missing client effort.
- Apply the same handling during Responses WebSocket preparation.
- Remove internal routing fields before upstream serialization and discard client-forged directive markers at ingress.

This is a runtime fix only: no editor redesign, service-tier feature, or provider catalog changes.

## Related Issues

Historical base-red report #12732 is now closed. Current CI failures must be assessed on their own merits.

The feature was initially based on `release/v3.8.51` at `152d95108c9c3d557562311ffed63240a511eb31`. No fork deployment workflow, container configuration, admission hotfix, or provider-catalog patch is included.

## Validation

- [x] Change type: routing / request translation
- [x] `node --import tsx/esm --test tests/integration/reasoning-routing-reliability.test.ts` — 11 passed after adding forced/explicit precedence regressions for adaptive header opt-in.
- [x] Diff checked against the current release tip; `git diff --check` passed.
- [x] Automated regression tests included.
- [ ] Full repository lint, production build and coverage matrix: left to PR CI, not claimed green locally.

Earlier verification of the equivalent patch on the fork reproduced the native-request 400 before the fix. After deployment, native Responses and Chat through a combo returned HTTP 200; captured provider payloads contained the forced low effort and no internal rule marker. These are fork-deployment observations, not a claim that this PR's CI has passed.

## Tests Added Or Updated

- `tests/integration/reasoning-routing-reliability.test.ts`: native-format directives, combo fallback, model/account defaults, request-local isolation between keys, force/default semantics, WebSocket preparation, and internal-field leakage.

## Coverage Notes

The integration tests exercise real routing/translation modules with isolated storage and a mocked upstream. They are deterministic and do not require provider credentials. Full coverage was not measured locally.

## Reviewer Notes

No migration and no changes to stored user rules or model permissions. Inspect the request-local context through retry/fallback and the native passthrough boundaries. The service-mode PR touches some of the same credential-preparation sites; retain both independent contexts when integrating the two patches.

### Related independent PRs

- #13555 — dedicated routing editor, including full-width layout.
- #13556 — reliable reasoning overrides across transports and fallback.
- #13554 — per-key service tier overrides.

All three target the same release tip directly, not each other. Shared-file overlaps may need small integration resolutions after the first PR lands; unrelated fork-only changes are deliberately excluded.

Review follow-up: added the fix fragment and adaptive-opt-in precedence regressions. A separate local integration checkout combining this PR, #13448 and #13554 passes 40 focused tests; no sibling feature commits were added to this PR.


diegosouzapw:
Real bug with a genuine security angle: a server-forced reasoning-effort rule could be lost
across native request handling, model-suffix/account defaults, or fallback prep — and could
also leak into an upstream payload as an unsupported field. This fix carries the directive
through executor dispatch/refresh/fallback via a non-serializable Symbol-keyed credential
context, strips the internal field before upstream serialization, AND discards any
client-supplied `_omnirouteReasoningRule`/`_omnirouteReasoningRouteTrace` at ingress so a client
can never forge a forced-effort rule for itself — nice defense-in-depth on both ends. Ran the
full `reasoning-routing-reliability.test.ts` at this PR's head: 9/9 pass. Clean, single-commit
diff, no scope creep. Only thing missing is the changelog fragment — please add one. Worth a
joint look with #13448 (adaptive effort auto) since both resolve reasoning effort in the same
chatCore.ts area; the precedence order you both describe (forced > explicit > auto) should
compose correctly but hasn't been verified together.

JxnLexn:
Thanks for the review. Added `changelog.d/fixes/13556-reasoning-rule-propagation.md` plus two regressions for adaptive-header opt-in in 68ed84716: forced effort wins over explicit effort + auto opt-in, and explicit client effort survives auto opt-in.

`node --import tsx/esm --test tests/integration/reasoning-routing-reliability.test.ts`: **11/11 passed** on the isolated PR branch; changelog integrity passes.

For the joint check, I used a separate local checkout of release tip ac52d4d9e with #13448 (8867c1ca8), #13554 and this PR. The two integration files plus `adaptive-effort.test.ts` and `adaptive-effort-wiring.test.ts` pass **40/40**. This exercises the forced > explicit > auto ordering with the adaptive implementation actually present, rather than just sending an ignored header on the isolated branch.

The preparation sites shared with #13554 need both request-local credential contexts retained, including the WebSocket preparation path. I resolved those only in the local integration checkout; neither sibling feature was added to this PR.

Issue #3572, opened 2026-06-10
[BUG] Valid max_tokens-truncated response misclassified as "empty content" → fake 502


Forcerecon:
### OmniRoute Version

3.8.18 (also observed on 3.8.19)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Provider(s) Involved

Claude (Anthropic OAuth)

### Model(s) Involved

claude-fable-5, claude-opus-4-8

### Client Tool

Claude Code (CLI 2.1.170)

### Description

When a client sends a request with a very small `max_tokens` (e.g. `max_tokens: 1`, used by Claude Code as a connectivity/pre-flight ping), Anthropic legitimately returns **HTTP 200** with `stop_reason: "max_tokens"`, an empty `content: []` array, and `usage.output_tokens: 1`. This is a valid, successful completion that was simply truncated at the token limit.

OmniRoute's `isEmptyContentResponse()` classifies any Claude-format response with `content.length === 0` as a "fake success / empty content" failure, **ignoring `stop_reason`**. It then:
1. Rewrites the upstream **200** into a synthetic **502 "Provider returned empty content"**.
2. Triggers a family fallback. For `claude-fable-5` there is **no** `MODEL_FAMILIES` entry, so `getNextFamilyFallback()` returns `null` and the 502 is returned to the client immediately. For `claude-opus-4-8` it falls back to a sibling, which — given the same `max_tokens: 1` — also returns empty content and still ends as 502.

Operator impact: every legitimate `max_tokens: 1` ping is turned into a 502 error. On one production proxy this produced 16 false 502s in a single morning, polluting error metrics and breaking the client's pre-flight check.

### Steps to Reproduce

1. Deploy OmniRoute 3.8.18 with a Claude (Anthropic) connection and a combo `claude-fable-5` mapped to that connection.
2. Send a Claude-native request through OmniRoute with a minimal token budget:
   ```bash
   curl -sS https://<omniroute-host>/v1/messages \
     -H 'content-type: application/json' \
     -H 'x-api-key: <omniroute-api-key>' \
     -H 'anthropic-version: 2023-06-01' \
     -d '{
       "model": "claude-fable-5",
       "max_tokens": 1,
       "messages": [{"role": "user", "content": "ping"}]
     }'
   ```
3. Observe the client receives **502 "Provider returned empty content"**.
[...]

PR #13560, opened 2026-09-13
fix(combos): stop Gemini thinking from failing dashboard combo tests

HouMinXi:
fix(combos): stop Gemini thinking from failing dashboard combo tests

Dashboard "Test all" on onmi-gemini3.6 reports Provider returned empty content
for agy/gemini-3.8-flash-high in 6-9s. The same Google
account answers normal chat through the local agy CLI.

## Cause

The health probe is `Reply with exactly: pong` with `max_tokens=64`.
Live Gemini 3.8 flash-high injects `thinkingLevel=high`. Thinking
eats the budget and the model returns `content:""` /
`finish_reason=length`. Two layers then treat that as failure:

1. The combo probe body did not send Gemini's documented off-switch.
2. `detectMalformedNonStream` rewrote truncated Chat Completions as
   an empty-content 502. `isEmptyContentResponse` already accepted
   `finish_reason=length` as a real truncation (#3572).

## Change

- Gemini probes only: `reasoning_effort: none`. Other providers keep
  the old body so strict OpenAI-compatible endpoints do not 400.
- Treat `finish_reason` `length` / `tool_calls` / `content_filter` as
  truncated success, matching the empty-content classifier.

Live artifact: call_logs 2026-09-13T04:21:57Z, flamehaze732, 10936ms,
`maxOutputTokens=32`, `thinkingLevel=high`, `finish_reason=length`.


diegosouzapw:
Great root-cause work here — tracing it to the live call_log
(`maxOutputTokens=32`/`thinkingLevel=high`/`finish_reason=length`) and fixing both
layers (the probe body missing Gemini's documented off-switch, and
`detectMalformedNonStream` not yet accepting `finish_reason=length` the way the
streaming classifier already does since #3572) makes this a clean, complete fix. Ran
the full relevant suite at your head: 58/58 pass. Looks merge-ready as-is.

PR #13565, opened 2026-09-13
fix(build): decouple client combo control center and provider registry from server-only imports (#13474)

### LINKED ISSUE #13474: fix: release/v3.8.51 no longer builds: #13283 makes a "use client" module import the DB/playwright chain

SCys:
## Summary

Fixes client bundle build errors in `release/v3.8.51` (Turbopack and Webpack production builds) caused by client components importing server-only dependencies (#13474):

1. **Combo Control Center client boundary**: `src/lib/combos/controlCenter.ts` imported `resolveProviderAlias` from `open-sse/services/model.ts`, which transitively pulled DB models, settings, token health checks, and `playwright-core`/`async_hooks`/`child_process` into the `ComboControlCenterClient.tsx` client bundle (causing 168 Turbopack build errors). Replaced with `resolveProviderId` from `@/shared/constants/providers` alongside client-safe canonical provider alias resolution (`resolveCanonicalProviderId`).
2. **Provider Registry client boundary**: `open-sse/config/providers/registry/codebuddy-cn/index.ts` imported `CODEBUDDY_CN_USER_AGENT` from `src/lib/oauth/constants/oauth.ts`, which pulled `cursorAgentCliVersion.ts` and `node:fs`/`node:os`/`node:path` into `CliAgentsPageClient.tsx` (causing Webpack/Docker build errors). Extracted `CODEBUDDY_CN_USER_AGENT` into `open-sse/config/providerHeaderProfiles.ts` and re-exported it from OAuth constants for backwards compatibility.
3. **Automated guards**: Added `cursorAgentCliVersion.ts` and `open-sse/services/model.ts` to `SERVER_ONLY` in `tests/unit/client-bundle-no-server-only-10692.test.ts`, and added alias normalization unit tests in `tests/unit/combo-control-center.test.ts`.

## Related Issues

- Closes #13474
- ⚠️ base-red inherited: #12732

## Validation

- [x] Change type: build-deploy / routing / UI
- [x] Focused tests and category gates from the golden path:
  - `node --import tsx/esm --test tests/unit/client-bundle-no-server-only-10692.test.ts tests/unit/combo-control-center.test.ts` (7/7 passed)
  - `npm run typecheck:core` (passed)
  - `npm run check:docs-sync` (passed)
  - Multi-arch Docker production build (`next build --webpack`) verified clean
- [x] `npm run lint` (passed via pre-commit hooks)
- [x] Reconciled with the current active release base (`upstream/release/v3.8.51` @ `152d95108c`)

## Tests Added Or Updated

- `tests/unit/combo-control-center.test.ts`: Added unit test verifying `getComboControlCenterTargets` normalizes `opencode` and provider alias model prefixes without server-only dependencies.
- `tests/unit/client-bundle-no-server-only-10692.test.ts`: Added `open-sse/utils/cursorAgentCliVersion.ts` and `open-sse/services/model.ts` to `SERVER_ONLY` set to guard against future client bundle leaks.

## Coverage Notes

- Covered by `tests/unit/combo-control-center.test.ts` and `tests/unit/client-bundle-no-server-only-10692.test.ts`. No coverage regressions.

## Reviewer Notes

- Preserves all canonical alias normalization behavior (`xiaomi` -> `xiaomi-mimo`, `llamacpp` -> `llama-cpp`, `agy` -> `antigravity`, `aq` -> `amazon-q`, `opencode` / `oc`) without dragging server-side modules into client components.
- `CODEBUDDY_CN_USER_AGENT` remains exported from `src/lib/oauth/constants/oauth.ts` so all existing callsites remain compatible.


diegosouzapw:
Measured result for the release-green build red. I ran `npm run check:pack-artifact` on an idle box against `release/v3.8.51` `f7dbfc88`:

- tip alone: `Turbopack build failed with 168 errors`, all from the `ComboControlCenterClient.tsx → controlCenter.ts → open-sse/services/model.ts` chain.
- tip + #13635 + #13436: compiles in 8.1 min, pack policy passes, and the next stop is the provenance stamp, which the validator never writes (fixed in #13678).
- tip + **this PR** alone: compiles too, but fails pack policy until #13635's `opencode-plugin-v2` allowance lands.

This PR and #13436 fix the same edge in different ways and conflict on `src/lib/combos/controlCenter.ts`, so only one of them can land as-is. This PR also conflicts with #13635 on `src/lib/oauth/constants/oauth.ts`, since both move the codebuddy-cn user agent. #13436 merges cleanly after #13635, which is why that pairing is the combination I measured green. The choice is the maintainer's.

maxmad64bis:
Compared both closely since they fix the same thing. #13436 keeps the alias logic in one place by moving it as-is into a small module, while this one rewrites it in the caller with a copy that can go stale over time. It's the same story for the version pin and the guard. That said, moving the registry off the oauth hub here is genuinely the nicer layering. Sensible way forward in my view: take #13436 to unblock the build, then bring the registry move over as a small follow-up.

SCys:
Agreed that #13436's approach of keeping the alias logic centralized in a dedicated module is cleaner and avoids duplicate code. Since #13436 resolves the same client-bundle root cause, merges cleanly alongside #13635, and gets the build green, closing this PR in favor of #13436 to avoid holding redundant PRs open. We can follow up on the provider registry layering separately if needed. Thanks for testing and the review!

Issue #13377, opened 2026-09-11
chore(auth,db,skills): hygiene follow-ups from the omni-review closeout final reviews (#13375, #13374)


diegosouzapw:
## Context
Residuals from the final whole-branch reviews of the omni-review closeout (PRs #13375 SEC-A, #13374 SEC-C/D, skills repo commits ea02406..9f8f3fb). None is exploitable or blocking; all are small, mechanical hygiene items that ride on the same tests.

## Auth (after #13375 merges)
- [ ] `src/server/authz/pipeline.ts`: delete the now-unreachable `isStaleDashboardJwtError` + `staleDashboardJwtWarningEmitted`; if the one-time "Dropped stale dashboard session cookie" log is wanted, emit it in the `!payload` branch of `refreshDashboardSessionIfNeeded`.
- [ ] Minters sign with `getDashboardJwtSecret()` (trimmed) — `src/app/api/auth/login/route.ts`, `src/app/api/auth/oidc/callback/route.ts`; `src/lib/ws/handshake.ts` uses the helper default instead of encoding the raw env value, so mint and verify agree everywhere.
- [ ] `verifyDashboardSessionToken`: pin `{ algorithms: ["HS256"] }` (defence in depth; jose already rejects `alg: none` and key confusion with a `Uint8Array` key); reject an empty `Uint8Array` (`!secret?.length`).
- [ ] Use the exported `DASHBOARD_SESSION_COOKIE` / `DASHBOARD_SESSION_CLAIM` constants in the six consumers, or drop the exports.
- [ ] Source guard (`tests/unit/dashboard-session-verifier-source-guard.test.ts`): repo-wide allowlist of files that may import `jwtVerify` from jose (helper + `oidc/callback`), so a NEW cookie consumer is caught, not only regressions in the six known files.
- [ ] `tests/unit/dashboard-session-token-13298.test.ts`: `try/finally` around the env mutation test.
- [ ] `docs/architecture/AUTHZ_GUIDE.md`: "route guard" → "dashboard route guard (`isDashboardSessionAuthenticated()`)".

## Batches sweep (after #13374 merges into #12969)
- [ ] Forward-progress guard in both `for (;;)` loops of `deleteCompletedBatches` (throw if the next chunk starts with the same first id as the previous one — a plain `deletedBatches === 0` check would false-fail on a concurrent deleter).
[...]

Issue #6289, opened 2026-07-05
fix(startup): Turbopack cache mmap failure surfaces as "Module not found" (os error 1455)


chirag127:
## Problem

`pnpm dev` fails at startup. Turbopack overlay reports `os error 1455` while `C:\pagefile.sys` shows 113 MB used of 255929 MB allocated.

## Verbatim error

```
Failed to restore task data (corrupted database or bug): Data for TaskId 998053)
  Caused by:
    0: Looking up task storage for TaskId 998053 from database failed
    1: Unable to open static sorted file referenced from 00000525.meta
    2: Unable to open static sorted file 00000524.sst
    3: Failed to mmap SST file .build/next/dev/cache/turbopack/v16.2.10/00000524.sst (144869601 bytes)
    4: The paging file is too small for this operation to complete. (os error 1455)

Module not found: Can't resolve '@/shared/utils/machine'
Import map: aliased to relative './src/shared/utils/machine' inside of [project]/
```

## Env

- OmniRoute 3.8.44 (`package.json`), upstream main matches
- Next.js 16.2.10, Turbopack dev
- Node v24.18.0, pnpm 11.9.0
- Windows 11 Enterprise 26200 (corp VM), 128 GB RAM
- Page file: AllocatedBaseSize 255929 MB, CurrentUsage 113 MB, PeakUsage 145 MB
- Launcher: `scripts/dev/run-next.mjs`

## Workaround

```
Remove-Item -Recurse -Force .build\next\dev\cache\turbopack
```

Restart → `:20128` returns 307 (login redirect, healthy).

## Ask

Consider auto-clearing the Turbopack cache in `scripts/dev/run-next.mjs` when child stderr matches `restore task data`, `mmap SST`, or `os error 1455`. The page file is not exhausted at fire time, so the Windows message is not diagnostic.

Thanks for the great work on OmniRoute!

diegosouzapw:
Thanks for the report! Fixed in `release/v3.8.46` (PR #6445 — best-effort self-heal that purges a corrupted Turbopack dev cache and retries once (note: it's best-effort; the manual cache purge remains the reliable remedy)). It's already merged into the active development branch, so feel free to pull and test it now. Ships in the next release.

PR #13566, opened 2026-09-13
fix(cli): Codex long-session turn-pin fallback + codex-settings key resolution (#13564 #13563)

opensource-elearning:
## Summary

Two independent Codex CLI fixes that were blocking long-running Codex sessions and the dashboard configure flow, shipped together because both belong to the same tool surface and both are verified by their own regression tests.

| Fix | Issue | Symptom |
| --- | --- | --- |
| Release the native Codex turn pin when the pinned model becomes model-scoped unusable | #13564 | `400 NATIVE_CODEX_PINNED_MODEL_UNAVAILABLE` bricks a long session mid-turn |
| Resolve the codex-settings `apiKey` through the canonical resolver | #13563 | `400 baseUrl, apiKey and model are required` on every dashboard Apply in cloud mode |

## Issues addressed

- **Closes #13564** — Codex long-session fallback (multi-model combo, mirroring Claude Code).
- **Closes #13563** — Codex settings still fail to configure when `apiKey` is empty.
- Related, **not fixed here**: #12273 (context-window pre-screen dispatches doomed 200K targets for a 211K request; that flood also feeds `/dashboard/logs/proxy`). A fresh reproduction was added to #12273 today. It is a separate dispatch-path change and will get its own PR.

### 1. Codex long sessions die to a terminal 400

Native Codex requests carry a turn pin: for the duration of one `turn_id` the combo locks provider + model. If that model becomes model-scoped unusable mid-turn (per-model quota lockout, connection cooldown, exhausted account), every continuation request for the same turn returned a non-retryable `400 NATIVE_CODEX_PINNED_MODEL_UNAVAILABLE`. Claude Code has no such pin, so it naturally falls back to the next healthy combo model on every request; Codex could not.

### 2. Codex settings always 400 in cloud mode

The `codex-settings` route diverged from every other CLI tool. `baseUrl` and `model` are Zod-gated (`min(1)`), so the `400 "baseUrl, apiKey and model are required"` could only fire on an empty `apiKey`. The dashboard sends an empty `apiKey` exactly when cloud mode is on and no management key is selected — while the Apply button stays enabled. The route used an inline `if (!apiKey) return 400` plus a hand-rolled `getApiKeyById`, instead of the shared `resolveApiKey(keyId, apiKey)` used by `cline/forge/openclaw/grok-build/jcode-settings`.

### Commit 1 — `fix(sse): release native Codex turn pin when pinned model is model-scoped unusable (#13564)`

[...]

PR #13568, opened 2026-09-13
fix(build): keep node:fs out of the client chunk graph (Turbopack /login + /dashboard)

jonlwheat2-gif:
## Summary

Turbopack (the Next 16 default bundler) was **panicking** on `npm run dev` while writing the page endpoints for `/login` and `/(dashboard)/dashboard`, so both routes returned **500**:

```
FATAL: An unexpected Turbopack error occurred.
Failed to write app endpoint /login/page
Caused by:
- the chunking context (unknown) does not support external modules
  (request: node:fs)
```

This PR fixes the root cause. It is a **build/bundling bug, unrelated to #13377**.

## Root cause

A **client** component drags a module with top-level Node builtins into the browser chunk graph. There is exactly one chain feeding **both** failing pages:

```
src/app/login/page.tsx            (and src/app/(dashboard)/layout.tsx)
 → src/shared/components/index.tsx               ← barrel
 → src/shared/components/ModelSelectModal.tsx    ← client component
 → src/shared/constants/models.ts
 → @omniroute/open-sse/config/providerModels.ts
 → @omniroute/open-sse/config/providerRegistry.ts
 → @omniroute/open-sse/config/providers/index.ts
 → .../providers/registry/codebuddy-cn/index.ts
 → src/lib/oauth/constants/oauth.ts
 → @omniroute/open-sse/utils/cursorAgentCliVersion.ts   ← top-level node:fs / node:os / node:path
```

`fs` / `path` / `os` / `child_process` are listed in `serverExternalPackages` (`next.config.mjs`), so Turbopack externalises them — and the browser chunking context cannot emit an external reference, hence the panic.

### What it is NOT

- **Not cache corruption.** Purging the 1.1 GB `.build/next/dev/cache/turbopack` (the remedy the repo's own `scripts/dev/run-next.mjs` comment recommends for this error class, #6289) did **not** help — so the `node:fs` reference is real.
- **Not caused by any auth work.** The identical panic was already logged before this change.

## Fix

Split the pinned build id into a **client-safe** module with no Node builtins:

| File | Change |
|------|--------|
| `open-sse/utils/cursorAgentCliVersion.constants.ts` **(new)** | Holds only `CURSOR_AGENT_CLI_VERSION` (+ a header comment documenting the chain above). No Node builtins. |
| `open-sse/utils/cursorAgentCliVersion.ts` | Imports the constant from `.constants.ts` and **re-exports** it, so the module's public API is unchanged. Its `fs`/`os`/`path` machinery is untouched. |
| `src/lib/oauth/constants/oauth.ts` | Imports `CURSOR_AGENT_CLI_VERSION` from the `.constants.ts` module instead of the fs-bearing one. |

> Note: **all** Node builtins had to leave that module's top level, not just `fs` — `node:os` and `node:path` would have panicked next.

## Verification

- `/login` → **200** (was 500)
- `/dashboard` → 307/308 → `/login` (normal auth redirect)
- `/api/monitoring/health` → `{"status":"healthy","setupComplete":true}`
- **Zero** new `next-panic-*.log` files
- `npx tsc --pretty false -p tsconfig.typecheck-core.json` → clean
- `tests/unit/cursor-agent-cli-version.test.ts` + `tests/unit/cursor-login-pkce.test.ts` → **26/26**

## Escape hatch

`OMNIROUTE_USE_TURBOPACK=0 npm run dev` forces webpack, which tolerates the `node:fs` reference and lets the pages compile — but it is a workaround, not a fix.


jonlwheat2-gif:
**CI note — the failing checks are pre-existing, not from this PR.**

Every failing job here fails identically on `release/v3.8.51` and on the other open PRs against it (#13565, #13566, #13571). Passing on all of them: `API Route Typecheck`, `Vitest (fast-path)`, `semgrep`.

Failing everywhere:

`Docs Gates (fast-path)` · `Fast Quality Gates` · `Merge integrity (changelog + generated skills)` · `No new ESLint warnings` · `Unit Tests fast-path (1/4–4/4)`

Two pinned from the logs:

- **Merge integrity** → `npm run check:agent-skills-sync` exits 2 because the generator wants to create a `cli-tunnel` SKILL.md that isn't committed (`Generated: 1 · Unchanged: 45 … + cli-tunnel`).
- **Unit Tests** → e.g. `each rotated account attempt acquires and releases a fresh composite slot`, `repository contract is in sync (live data)`, `Failed to configure Qwen Code: Expected property name or '}' in JSON` — none of these touch this PR's files.

The base commit `152d95108` is itself red (Node 24/26 Compat, A11y axe, Validate main branch).

diegosouzapw:
Thanks for tracking this down — the root cause you found (oauth.ts pulling node:fs via
cursorAgentCliVersion.ts) is confirmed real. There's another open PR (#13604) fixing the exact
same edge with an equivalent approach, plus a regression-guard test update your PR doesn't
include. To avoid landing two different "extract the pin to a sibling file" patterns for the
same bug, we're going with #13604 and closing this one — appreciate you finding and fixing it
independently, and the extra detail in your comments was useful cross-reference material.

diegosouzapw:

PR #13569, opened 2026-09-13
fix(models): enable xAI OAuth live discovery

### LINKED ISSUE #13443: feat(providers): Live model discovery is unreachable for some OAuth variants, not for every major OAuth vendor
# Live model discovery is unreachable for some OAuth variants, not for every major OAuth vendor

Measured against `upstream/release/v3.8.51` @ `152d95108`.

## Count flip (S1)

The first version of this issue counted only the generic test at `route.ts:749-753` plus `if (provider === "...")` branches. That produced 198 unreachable entries and 12 OAuth providers with no path. It missed a fourth gate at `route.ts:2088-2091`: `PROVIDER_MODELS_CONFIG` (48 keys, including `claude`, `grok-cli`, `kimi-coding`, `xai`) then `deriveConfigFromRegistryModelsUrl(provider)`, which reads registry `modelsUrl` as the fetch URL (`discoveryConfig.ts:24-38`).

Recount with all four gates: 99 unreachable / 6 OAuth with no path. `claude`, `grok-cli`, and `kimi-coding` do have a live path. `xai` (API key) does; `xai-oauth` does not.

## Summary

A provider gets live `/v1/models` discovery if the request hits one of four gates in `src/app/api/providers/[id]/models/route.ts`. 276 registry entries. 99 have no path. Of the 21 OAuth providers, 6 have zero discovery.

The practical hole for Grok 4.7 is `xai-oauth`: the API-key sibling `xai` already lists `https://api.x.ai/v1/models` in `PROVIDER_MODELS_CONFIG` (`providerModelsConfig.ts:595-602`). `xai-oauth` shares `baseUrl` (`open-sse/config/providers/registry/xai/index.ts:51-61`) but is absent from that map, absent from `NAMED_OPENAI_STYLE_PROVIDERS`, and absent from `CATALOG_SIBLING_IDS` (`activeSyncedCatalog.ts:68-71`, agy/antigravity only). SuperGrok OAuth stays on the frozen seed.

`grok-cli` is different: it has a config entry (`providerModelsConfig.ts:472-486`) pointing at `GROK_BUILD_MODELS_URL`. It can pick up a new model on refresh. `mapTokens` does not set `autoSync` (`grok-cli.ts:408-417`, `grok-cli-oauth.ts:111-114`), unlike Claude (`claude.ts:146`). The 24h scheduler will not poll it unless someone sets that flag.

## The four gates

1. Generic OpenAI-style (`route.ts:749-753`): `isOpenAICompatibleProvider` (`providers.ts:199-201`, prefix `openai-compatible-` only, never a registry id), `isLocalOpenAIStyleProvider` (`helpers.ts:43`, `SELF_HOSTED_CHAT_PROVIDER_IDS`), `isNamedOpenAIStyleProvider` (`providerSets.ts:110-112`, 52-entry `Set`). This branch builds `{base}/v1/models` from `baseUrl`. The loop variable is named `modelsUrl`; it is not registry `modelsUrl` (`route.ts:803-823`).

2. Per-provider `if (provider === "...")` branches. 37 comparisons, 36 distinct ids, 2429-line file. `maybeHandleConolModelDiscovery` (`conolDiscovery.ts:30-33`, wired at `route.ts:652-665`) uses `!==`, so a naive `provider ===` scan misses `conol-web`.

3. `PROVIDER_MODELS_CONFIG` (`providerModelsConfig.ts:400`, 48 keys). Claude uses `buildClaudeModelsHeaders` (`claudeModelsHeaders.ts:3-25`, Bearer + `anthropic-beta: oauth-2025-04-20` when an access token is present). `grok-cli` uses `getGrokBuildModelsHeaders` and `parseGrokBuildModels` (`providerModelsConfig.ts:371-375`).

4. Catch-all (`route.ts:2088-2091`): the config map or `deriveConfigFromRegistryModelsUrl`. If both miss and a local seed exists, the route returns `source: "local_catalog"` with `intentional: true` (`route.ts:2189-2213`). Empty seed and no config: HTTP 400 (`route.ts:2215-2219`).

Registry `modelsUrl` is necessary for gate 4, not for gate 1. `providerSets.ts:8-10` already documents that a real `modelsUrl` without `Set` membership serves the hardcoded seed. `getDiscoveryClass` (`discoveryClass.ts:27`) treats `modelsUrl` as `openai-compat`; that label does not by itself fetch.

## Counts

Reproduce with the script in the appendix (updated to include gate 3 and 4).

| Group | entries | reachable |
|---|---|---|
| OAuth | 21 | 15 |
| API key / other | 255 | 162 |
| Total | 276 | 177 |

OAuth providers with no discovery path (seed size in parentheses):
`devin-desktop` (110), `devin-cli` (110), `codebuddy-cn` (15), `trae` (9), `xai-oauth` (7), `gitlab-duo` (2).

OAuth providers that work via `PROVIDER_MODELS_CONFIG` without a 749/`===` branch: `claude` (11), `clinepass` (11), `cline` (13), `kilocode` (15), `kimi-coding` (3), `grok-cli` (3).

OAuth providers that work only via a bespoke `===` branch: `cursor` (198), `github` (34), `codex` (33), `ghe-copilot` (26), `kiro` (11), `antigravity` (10), `agy` (10), `zed-hosted` (0).

93 API-key entries have no discovery path. The largest are `cursor-api` (198), `devin-cli-agentic` (110), `uc-direct` (82), `cheaperinference` (39). Many of those are cookie/web/embedding providers the catch-all already treats as intentional local catalog (`route.ts:2208-2211`).

## Why the seed cannot be the source of truth

`providerUsesAuthoritativeLiveCatalog` (`providerRegistry.ts:215-223`) defaults to `true`. `collectProviderModelIds` (`providerWildcard.ts:143-151`) lets a live catalog replace the seed when `authoritative`. For the 99 providers above, `getActiveSyncedCatalog` (`activeSyncedCatalog.ts:202-238`) returns empty and routing falls back to the frozen seed. Nothing logs a warning, because the empty-catalog fallback is also the correct behaviour for a provider with no discovery endpoint. The two cases are indistinguishable from outside.

`sync-models` does not help: it self-fetches the same GET route (`sync-models/route.ts:31,362-366`). A provider invisible to `route.ts` is equally invisible to a sync run.

Claude OAuth connections set `autoSync: true` (`claude.ts:146`). `grok-cli` and `xai-oauth` do not. The scheduler (`modelSyncScheduler.ts`) only walks connections with that flag. `isAutoFetchModelsEnabled` (`modelDiscovery.ts:252-255`) is a different flag (`autoFetchModels === true`) and defaults off.

## Upstream endpoints

Unauthenticated probes, 2026-09-12. 401/400/403 means the route exists and wants credentials; 404 means no such endpoint.

| endpoint | status |
|---|---|
| `https://api.anthropic.com/v1/models` | 401 |
| `https://api.x.ai/v1/models` | 401 |
| `https://api.githubcopilot.com/models` | 400 |
| `https://cli-chat-proxy.grok.com/v1/models` | 401 |
| `https://chatgpt.com/backend-api/codex/models` | 401 |
| `https://generativelanguage.googleapis.com/v1beta/models` | 403 |
| `https://api.moonshot.cn/v1/models` | 401 |
| `https://api.kimi.com/coding/models` | 404 |

So `claude` and `xai` have an upstream catalog endpoint that is reachable in principle. Whether an OAuth access token is actually *scoped* for it is unverified. I could not test this: the only `claude` row on the deployment holds a 4-character placeholder rather than a real token, and both `/v1/models` and `/v1/messages` returned `401 Invalid bearer token` for it. That result says nothing either way. Do not read it as evidence. Verifying scope per vendor is step 1 of any fix, not an assumption to build on. `kimi-coding` is the one case where a 404 suggests the catalog endpoint does not exist on that path; Omni still has `KIMI_CODING_MODELS_CONFIG`.

Related: #12866 (merged) made live account catalogs authoritative for Claude, Codex, Copilot, AGY. This issue is the prior question: does the GET ever fire. #13434 is dashboard display of catalogs that already synced.

## Suggested direction

Ordered preference.

1. **Give `xai-oauth` the same discovery config as `xai`.** Add an `xai-oauth` key to `PROVIDER_MODELS_CONFIG` reusing `url: "https://api.x.ai/v1/models"`, Bearer, and the same `parseResponse`. Do not add `xai`/`xai-oauth` to `CATALOG_SIBLING_IDS`: that unions catalogs across connections of both ids. API-key and SuperGrok accounts are not the same pool. Do not add `grok-4.7` to the static seed.

2. **Lock gate 4 with a test.** A registry entry with `modelsUrl` and no `Set` membership and no `===` branch must still reach `deriveConfigFromRegistryModelsUrl`. Remove the `modelsUrl` field: the test goes red. This prevents the next recount from treating 2088 as dead.

3. **Do nothing for the other five OAuth zeros until the vendor catalog URL is known.** `trae`, `gitlab-duo`, `devin-desktop`, `devin-cli`, `codebuddy-cn` have no `modelsUrl` and no public `/v1/models` in the registry. Inventing a URL is worse than the frozen seed.

I would avoid inventing a new abstraction over the 36 bespoke branches in the same change. Gate 4 already covers OpenAI-style `modelsUrl`. Option 1 covers the Grok OAuth hole. `grok-cli` `autoSync` is a separate product choice (Claude sets it; Grok does not).

## Acceptance criteria

Whichever option is taken, these are the checks I want to see, each of which fails before the change:

- `PROVIDER_MODELS_CONFIG["xai-oauth"]` uses `https://api.x.ai/v1/models`. Delete that key: the test must go red and `xai-oauth` must fall through to the local seed.
- `CATALOG_SIBLING_IDS` still has no `xai` / `xai-oauth` pair. Injecting the pair must fail the isolation test.
- A registry entry with `modelsUrl` set and no `Set` membership reaches `deriveConfigFromRegistryModelsUrl`. Remove `modelsUrl`: red.
- Per vendor touched, one recorded real-endpoint response confirming the OAuth token is scoped for the catalog route. Without that, the branch is written against an assumption.

## Appendix: reproduction script

Save as `discovery-audit.mjs` at the repo root, then run `node --import tsx/esm discovery-audit.mjs`.

```js
import { readFileSync } from "node:fs";

const reg = await import("./open-sse/config/providerRegistry.ts");
const sets = await import("./src/app/api/providers/[id]/models/discovery/providerSets.ts");
const helpers = await import("./src/app/api/providers/[id]/models/discovery/helpers.ts");
const pmc = await import("./src/app/api/providers/[id]/models/discovery/providerModelsConfig.ts");
const routeSrc = readFileSync("./src/app/api/providers/[id]/models/route.ts", "utf8");

const REGISTRY = reg.REGISTRY ?? {};
const named = sets.NAMED_OPENAI_STYLE_PROVIDERS ?? new Set();
const isLocal = helpers.isLocalOpenAIStyleProvider ?? (() => false);
const configKeys = new Set(Object.keys(pmc.PROVIDER_MODELS_CONFIG ?? {}));

const rows = Object.entries(REGISTRY).map(([id, e]) => {
  const murl = typeof e.modelsUrl === "string" && e.modelsUrl.length > 0;
  const g1 = named.has(id) || Boolean(isLocal(id)) || id.startsWith("openai-compatible-");
  const g2 = routeSrc.includes(`provider === "${id}"`);
  const g3 = configKeys.has(id);
  const any = g1 || g2 || g3 || murl;
  return { id, oauth: e.authType === "oauth", n: (e.models ?? []).length, any };
});

const oauth = rows.filter((r) => r.oauth);
const key = rows.filter((r) => !r.oauth);
console.log("total", rows.length);
console.log("oauth", oauth.length, "discoverable", oauth.filter((r) => r.any).length);
console.log("apikey", key.length, "discoverable", key.filter((r) => r.any).length);
console.log("oauth with none:", oauth.filter((r) => !r.any).map((r) => `${r.id}(${r.n})`).join(" "));
```

saadi12ana:
## Summary

- Enable live `/v1/models` discovery for xAI OAuth connections so new Grok models are not limited to the static seed catalog.
- Keep `xai` and `xai-oauth` as separate catalog pools; the change adds only a provider-specific discovery config.

## Related Issues

- Closes #13443
- Related to #

## Validation

- [ ] `npm run lint`

## Tests Added Or Updated

- `tests/unit/provider-models-discovery-split.test.ts`
- Targeted suite: 32 passed, 0 failed

## Coverage Notes

- The provider discovery configuration is covered by the targeted provider-models discovery suite, including endpoint, Bearer auth, and both supported response envelope shapes.

## Reviewer Notes

- No catalog sibling mapping was added, so API-key and OAuth xAI accounts remain isolated.
- The working tree contains unrelated pre-existing modifications; only the two files for this fix were committed.


diegosouzapw:
Thanks for tackling this — the fix itself is correct (adding `xai-oauth` to
`PROVIDER_MODELS_CONFIG` with the right endpoint). There's another open PR (#13518) on the
same issue (#13443) that extracts the config into a shared constant and also keeps
`HARDCODED_MODELS_CONFIG_IDS` in lockstep, so we're going to merge that one instead to avoid
two divergent copies of the same URL. Recommending this PR be closed in favor of #13518 —
appreciate you catching and fixing the issue independently. If you'd rather we merge this
one, it would need a retarget to `release/v3.8.51` plus the two items above.

PR #13567, opened 2026-09-13
chore(auth): hygiene follow-ups from omni-review closeout (#13377)

jonlwheat2-gif:
Closes #13377.

Mechanical **auth** hygiene follow-ups from the omni-review closeout final reviews (PR #13375 SEC-A). No behaviour change for valid sessions; all covered by existing tests.

**12 files changed, 82 insertions(+), 75 deletions(-).**

> **Scope note — this PR is intentionally auth-only.** The issue's *Batches sweep* bullets are **not** in this PR. Their wording ("forward-progress guard in **both** `for (;;)` loops") targets PR #13374's chunked implementation, which is stacked on PR #12969 — not this `release/v3.8.51` base, where `deleteCompletedBatches` still has a single loop. That guard is now tracked separately in **#13570**, stacked on #13374. See *Related PRs* below.

## Auth (from #13375)

- **`src/server/authz/pipeline.ts`** — deleted the now-unreachable `isStaleDashboardJwtError()` and the module-level `staleDashboardJwtWarningEmitted`; the one-time "Dropped stale dashboard session cookie" warning now fires in the `!payload` branch of `refreshDashboardSessionIfNeeded` (L126-142). The refresh minter now uses the shared `getDashboardJwtSecret()` (L126) instead of a local `getJwtSecret()`, and every cookie read/write/delete goes through the exported `DASHBOARD_SESSION_COOKIE` constant (L129 read, L139 delete, L158 set, L171 delete). The catch block just logs (L165) — the stale-error branch is gone.
- **`src/app/api/auth/login/route.ts`** — import `getDashboardJwtSecret` (L18); the minter signs with `getDashboardJwtSecret()!` (L163), removing the local `getJwtSecret()`.
- **`src/app/api/auth/oidc/callback/route.ts`** — import (L7); resolve `const secret = getDashboardJwtSecret()` (L200) and gate the redirect on it before signing (L214).
- **`src/shared/utils/dashboardSessionToken.ts`** — `verifyDashboardSessionToken` now rejects an empty `Uint8Array` (`secret.length === 0`, L35) and pins `{ algorithms: ["HS256"] }` on `jwtVerify` (L37).
- **`src/lib/ws/handshake.ts`** — `hasValidSessionCookie` uses `getDashboardJwtSecret()` (L39) and `DASHBOARD_SESSION_COOKIE` (L42) instead of encoding the raw env value, so mint and verify agree.
[...]

PR #13374, opened 2026-09-11
fix(db): owner-scoped file half and chunked instance sweep for DELETE /v1/batches/delete-completed (SEC-C/SEC-D, omni-code-sec on #12969)

diegosouzapw:
## Summary
Closes SEC-C and SEC-D of the omni-code-sec battery on #12969.

- **SEC-C** — key-scoped sweep soft-deletes only files whose api_key_id is the caller (new deleteFileOwnedBy); a referenced file another tenant owns, or an unowned one, stays intact.
- **SEC-D** — instance mode commits in 200-batch transactions (INSTANCE_SWEEP_CHUNK); each chunk stays atomic, a failure in chunk N rolls N back and keeps < N.
- route-scope fixture mints the real login shape (authenticated: true) ahead of #13298.

## Validation
files-delete-owned-by 4/4 · ownership-wvxc 11/11 · route-scope 10/10 · batch-deletion 8/8 · typecheck:core · changelog-integrity vs the feature branch. Full suite on .113: last comment.

⚠️ base-red inherited: #12732
Refs #12969

PR #13570, opened 2026-09-13
fix(db): forward-progress guard in both deleteCompletedBatches chunk loops (#13377)

jonlwheat2-gif:
## Summary

Stacked on **#13374** (`fix/batches-sweep-file-ownership-chunks`). Adds the **forward-progress guard** that issue #13377 asks for in **both** `for (;;)` loops of `deleteCompletedBatches`.

> **Base note.** This PR targets `fix/batches-sweep-file-ownership-chunks` (the head branch of #13374) because the two chunked loops only exist there — #13377's wording ("forward-progress guard in **both** `for (;;)` loops") is written against #13374, not against `release/v3.8.51`.

## Why

Both loops walk completed batches in chunks of `INSTANCE_SWEEP_CHUNK`:

```sql
SELECT id FROM batches WHERE status = 'completed' [AND api_key_id = ?]
ORDER BY rowid LIMIT ?
```

Each iteration sweeps its chunk (file soft-deletes + checkpoint DELETE + batch DELETE) and then re-selects the next chunk. If a DELETE ever fails to make progress — a concurrent deleter holding the same rows, or a swallowed no-op delete — the same chunk is returned forever and the request thread spins.

## Change

`src/lib/db/batches.ts` — **+23 lines, one guard per loop**:

- remember the first id of the previous chunk;
- throw when the next chunk starts with the same id:

```
deleteCompletedBatches: chunk sweep made no forward progress (first id repeated)
```

Failure semantics stay consistent with each mode:

| Mode | On throw |
|------|----------|
| key (`{ apiKeyId }`) | whole sweep runs in one outer transaction → everything rolls back |
| instance (`{ allTenants: true }`) | commits per chunk (SEC-D) → earlier chunks stay committed, current chunk rolls back, rethrow |

## Verification

- `git diff` against #13374's head is exactly the 23 added lines, `src/lib/db/batches.ts` only
- file parses clean (`node --experimental-strip-types --check`)
- the guard only fires on a genuine no-progress chunk, so the existing batch-deletion suites are unaffected; CI runs the full suite against this base

## Related

- **#13377** — tracking issue (this covers its "Batches sweep" bullet)
- **#13374** — base PR: chunked instance sweep + owner-scoped file half
- **#12969** — base of #13374 (currently `CONFLICTING`; needs a rebase before this stack can land)
- **#13567** — the auth-only half of #13377
- **#13568** — separate Turbopack dev-server fix (`node:fs` in the client chunk graph)


jonlwheat2-gif:
**CI note.** No `pull_request` workflows ran for this PR: the CI triggers are scoped to `release/**` and the default branch, so a base of `fix/batches-sweep-file-ownership-chunks` doesn't match. The only check that fired is `semgrep` (pass).

Checks will run once this is retargeted to `release/v3.8.51` (or via #13374's own CI). Note that the `release/v3.8.51` checks are currently red repo-wide — every open PR against it (#13565, #13566, #13571) shows the same failures.

Local verification of this change: the diff against #13374's head is exactly the 23 added lines in `src/lib/db/batches.ts`, and the file parses clean.

Issue #3396, opened 2026-06-07
[Bug] Mistral 400: trailing assistant message rejected — stripTrailingAssistantOrphanToolUse only handles tool_use blocks


pulyankote:
## Bug

Mistral API rejects requests with a trailing `assistant` message (text content) with:

```
400: Expected last role User or Tool (or Assistant with prefix True) for serving but got assistant
```

## Root Cause

`stripTrailingAssistantOrphanToolUse()` in `open-sse/services/contextManager.ts` only strips trailing assistant messages that contain `tool_use`/`tool_calls` blocks. It does **not** strip trailing assistant messages with plain text content.

In Claude-format conversations (Claude Code, Cursor, etc.), the last message can be `assistant` to signal "continue from here / model is mid-turn." After translation to OpenAI format, this trailing `assistant` message with text content is preserved and sent to Mistral, which strictly requires the last message to be `user` or `tool`.

## Repro

Send any request to a Mistral model where the last message has `role: "assistant"` with text content:

```json
{
  "model": "mistral/mistral-small-latest",
  "messages": [
    { "role": "user", "content": "Hello" },
    { "role": "assistant", "content": "Hi there!" }
  ]
}
```

Result: `400 invalid_request_message_order (code 3230)`

## Fix

In `BaseExecutor.execute()` (`open-sse/executors/base.ts`), after the existing `stripTrailingAssistantOrphanToolUse()` call, add a provider-specific guard for Mistral that strips any trailing `assistant` message with real content (text, tool_calls, or reasoning_content):

```typescript
const preSendMsgs = tb.messages as Record<string, unknown>[];
if (this.provider === "mistral" && preSendMsgs.length > 1) {
  const lastMsg = preSendMsgs[preSendMsgs.length - 1];
  if (lastMsg && lastMsg.role === "assistant") {
    const hasText =
      typeof lastMsg.content === "string"
        ? lastMsg.content.trim().length > 0
        : Array.isArray(lastMsg.content) && lastMsg.content.length > 0;
    const hasToolCalls =
      Array.isArray(lastMsg.tool_calls) && lastMsg.tool_calls.length > 0;
    const hasReasoning =
      typeof lastMsg.reasoning_content === "string" &&
      lastMsg.reasoning_content.length > 0;
    if (hasText || hasToolCalls || hasReasoning) {
      tb.messages = preSendMsgs.slice(0, -1);
    }
  }
}
```

[...]

PR #13572, opened 2026-09-13
fix(sse): strip trailing assistant prefill on official Claude OAuth

HouMinXi:
Official Claude OAuth returns 400 when the last turn is a text-only assistant message:

`This model does not support assistant message prefill. Conversation must end with a user message.`

Live 2026-09-13 on `claude/claude-opus-5` (connection 8849ab82). Combo `openai-gpt-sol` then `offical-claude` both failed on the same body, ~20ms, before any useful fallback.

## Why

`stripTrailingAssistantForProvider` already drops that shape for Mistral (`#3396`). Official Claude OAuth was not in `PROVIDERS_REQUIRING_USER_LAST_MESSAGE`, so the same trailing turn reached Anthropic and burned the next combo step.

The strip still keeps a trailing assistant turn that has `tool_use` / `tool_calls`. OpenAI-compatible providers stay unstripped.

## Test

`tests/unit/mistral-trailing-assistant.test.ts`: claude text-only strip (string and array content). Inject: drop `"claude"` from the set, those two tests fail; restore, 9/9 pass.


diegosouzapw:
Solid, well-scoped fix with a live repro — ran the suite green (9/9) against the PR head.
One small nit: the new comment block in `base.ts`'s sibling file
(`open-sse/services/contextManager.ts`) has inconsistent indentation on 2 of its 3 lines —
`npx prettier --write` on that file before merge would clean it up. No other concerns.

HouMinXi:
Thanks for the review and the cleanup in `6766b4b1ae`. I fetched your current branch head and will preserve that correction in the upstream update.

PR #13604, opened 2026-09-13
fix(build): keep node:fs out of the client chunk graph for /dashboard/logs

hartmark:
## Problem

`/dashboard/logs` panicked live in Turbopack dev mode:

```
An unexpected Turbopack error occurred.
Failed to write app endpoint /(dashboard)/dashboard/logs/page
Caused by:
```

## Root cause

Same underlying issue as #13436 / #13509 / #13568: `open-sse/utils/cursorAgentCliVersion.ts` imports `node:fs`/`node:os`/`node:path` for local Cursor CLI install detection. Those three PRs already cut several edges reaching it (login page, combo control center, a CLI-tools page). This is a **further, previously-undiscovered edge**: `src/lib/oauth/constants/oauth.ts` imports `CURSOR_AGENT_CLI_VERSION` from the full resolver, and `oauth.ts` turns out to be reachable from a very large share of the dashboard's client bundles — confirmed via the repo's own `client-bundle-no-server-only-10692.test.ts` guard: registering the file as server-only and re-running the guard against the *un-fixed* source reports **250+ offending client entry points**, via `open-sse/config/providerModels.ts`, `open-sse/services/usage/supportedProviders.ts`, and others.

`/dashboard/logs` is simply the first of those 250+ pages that Turbopack's lazy dev-mode compilation happened to reach.

## Fix

Same pattern as the sibling PRs: the pin string moves to a new import-free `cursorAgentCliVersionPin.ts` (single source of truth). `cursorAgentCliVersion.ts` and `oauth.ts` both import it from there instead of duplicating it inline. `cursorAgentCliVersion.ts` itself is otherwise unchanged — it still owns the full env → local-install → disk-cache → pin resolution used by real Cursor CLI impersonation call sites.

Also registers `open-sse/utils/cursorAgentCliVersion.ts` in the existing `client-bundle-no-server-only-10692.test.ts` guard's `SERVER_ONLY` list, so this reproduces as a real, fast test failure rather than only a live Turbopack panic, and any future edge reaching it fails CI immediately (independent of whether #13436's broader Node-builtin-detection widening lands).

## Testing

- `client-bundle-no-server-only-10692.test.ts`: confirmed it fails (250+ offending entries reported) against the un-fixed source with only the `SERVER_ONLY` registration added, and passes clean with the fix.
- `tests/unit/cursor-agent-cli-version.test.ts` (6 tests) and `tests/unit/cursor-login-pkce.test.ts` (20 tests) — all pass unmodified.
- Reproduced the live Turbopack panic on `/dashboard/logs` before the fix, confirmed clean after.


diegosouzapw:
Thanks for the fix — verified locally: the exact `ModelSelectModal → models.ts → providerModels
→ providerRegistry → providers/index → codebuddy-cn → oauth.ts → cursorAgentCliVersion.ts`
chain is real on the current tip, and the existing `client-bundle-no-server-only-10692` guard
currently misses it (only catches a hardcoded module list). Your fix is clean and the guard
update closes that gap — nice touch citing #13436/#13509/#13568 in the comment, you clearly
checked for overlap. Only ask before merge: could you add a changelog fragment
(`changelog.d/fixes/13604-*.md`)? Everything else looks ready.

hartmark:
Added the changelog fragment — `changelog.d/fixes/13604-dashboard-logs-cursor-cli-version-client-leak.md`, verified with `check:changelog-integrity`. Ready to merge from my side.

diegosouzapw:

PR #13618, opened 2026-09-13
fix(sse): recover stale resource pressure admission

jbovard2016:
## Summary
- refresh stale critical pressure before `/v1/messages` structural rejection
- coalesce and bound refresh cycles, release waiters on dispose, and preserve the live runtime on invalid reload
- share the pressure runtime across duplicate module evaluations
- retain fresh critical shedding and bounded fail-open after `maxStaleMs`

## Verification
- regression file: 8/8
- focused resource-pressure/admission matrix: 136/136
- `npm run typecheck:core`
- `npm run typecheck:noimplicit:core`
- scoped ESLint, Prettier and `git diff --check`

Independent review accepted the source patch. The repository build remains blocked by pre-existing issue #12732; this PR does not claim to fix the initiating long-session memory growth.


jbovard2016:
The reviewed stale-pressure patch is deployed in local API pilot build `d049af25` alongside the memory fix. Compiled `/v1/messages` loads the pressure chunk; installed hashes and `BUILD_SHA` match the accepted package. Deterministic recovery regression remains 8/8, combined focused suite 228/228, health is 200, and real `/v1/messages` traffic is returning 200. A 48-hour qualification is running; no claim of long-run closure yet.

diegosouzapw:
Solid fix — the stale-critical-blocks-until-refresh-settles behavior (bounded by
`admissionRefreshTimeoutMs`) is the right shape, and the `Symbol.for` global store fixing the
dual-module-evaluation singleton bug is a nice catch on top of the stated goal. Ran the new
`resource-pressure-admission-recovery.test.ts` suite — 8/8 pass, including the real
`/v1/messages` route integration test. One gap: there's no `changelog.d/` fragment in the diff —
could you add one? Also worth a short mention that `defaultPressureSeverity` became a private
`cachedPressureSeverityForDiagnostics` (confirmed no other internal caller, but flagging the
rename for anyone depending on it externally).

maxmad64bis:
Heads-up on an overlap with #13823, which fixes the same latch (#13821) by keeping `defaultPressureSeverity()` and making it call `checkResourcePressureGuard()`. This PR makes that function private, and the regression test in #13823 imports it, so the two conflict in `chatBodyAdmission.ts` and only one is needed for #13821. Details: https://github.com/diegosouzapw/OmniRoute/pull/13823#issuecomment-5737386103

PR #13622, opened 2026-09-14
fix(models): keep OpenRouter :batch variants out of chat routing

### LINKED ISSUE #13596: fix(providers): ModelSync imports batch-only variants (*:batch) into chat routing — 91 Batch API rejections
**OmniRoute Version:** 3.8.50
**Installation Method:** npm (global)
**Operating System:** Linux
**OS Version:** Ubuntu / Linux 6.17.0-1020-oracle (arm64)
**Node.js Version:** v24.16.0
**Provider(s) Involved:** OpenRouter (synced `*:batch` variants), GitHub Copilot (`github`)
**Model(s) Involved:** `claude-*:batch`, `gemini-*:batch`, `minimax-m3:batch`, `inkling:batch` — 77 synced variants; plus unsupported Copilot models
**Client Tool:** Any client routed through OmniRoute

## Description

ModelSync imports upstream model variants that can never serve a normal chat request:

- **77 `*:batch` routes** were synced from OpenRouter into the local model→provider map (`key_value`), e.g. `gemini-3.6-flash:batch` → `openrouter/google/gemini-3.6-flash:batch`. When a client requests the base model and the router lands on a `:batch` variant, the upstream rejects with `404 This model is only available through the Batch API` — **91 errors** in ~41 h, the #3 error class. Provider-side warnings confirm the loop: `[provider] Node <uuid> model not found (N) for google/gemini-*:batch - locking mode`.
- Related catalog staleness on GitHub Copilot: `400 The requested model is not supported.` — **34 errors** in the same window, i.e. synced models the upstream no longer serves.

## Steps to Reproduce

1. Let ModelSync sync the catalog (observed `ModelSync:9router synced 387 model(s)`).
2. Send a normal chat request for e.g. `gemini-3.6-flash` (or let a combo/failover pick it).
3. Router selects a `*:batch` variant → upstream 404 `only available through the Batch API`.

## Expected Behavior

- `:batch`, embedding-only, moderation, and other non-chat variants are excluded from chat routing (or tagged batch-only and only used for `/api/beta/batches` requests).
- Models rejected as "not supported" are quarantined/removed after N consecutive rejections.

## Actual Behavior

Batch variants and stale models stay in the routing table, producing recurring 404/400 rejections and wasted failover attempts.

## Test Impact

Needs a new unit test — ModelSync filter rejects `*:batch` variants; plus integration test asserting synced catalog contains no batch-only routes.

## Error Logs / Output

```shell
[ERROR] [404]: This model is only available through the Batch API. Use the /api/beta/batches endpoint instead.   # ×91
[ERROR] [400]: The requested model is not supported.   # ×34 (github)
[provider] Node <uuid> model not found (N) for google/gemini-3.5-flash:batch - locking mode   # warn ×80+
```

## Possible Solutions

- Filter `:batch` / non-chat suffixes and capability flags during ModelSync import.
- Auto-quarantine: mark a model unavailable after repeated `not supported`/`Batch API` rejections.
- Surface "routable vs catalog" model counts in the dashboard.

## Validation Plan

- After sync, `key_value` contains no `*:batch` chat routes.
- A chat request never resolves to a `:batch` upstream.

## Additional Context

Local mitigation applied: deleted the 77 `*:batch` route keys from `key_value` — they will presumably return on the next ModelSync cycle until this is fixed upstream.

L4XB:
Closes #13596.

## What was wrong

`ModelSync` imported OpenRouter's Batch-API-only variants into the chat catalogue, so a chat request that landed on one was rejected upstream:

```
404 This model is only available through the Batch API. Use the /api/beta/batches endpoint instead.
```

91 of those in the reporter's 41-hour window (third error class by volume), plus the `[provider] Node <uuid> model not found (N) for google/gemini-*:batch - locking mode` failover churn behind them. Deleting the 77 `*:batch` keys locally only holds until the next sync cycle.

## Why here

`open-sse/services/modelEndpointPolicy.ts` already exists for exactly this: *"Upstream `/models` responses often omit endpoint/modality metadata… Keep the exceptional provider knowledge here so discovery, import, and catalog projection agree."* OpenRouter's `/models` carries nothing that separates a batch variant from a chat one, so `classifyExplicitEndpoints` returns `null` and the model falls through to `{ kind: "unknown", chatSelectable: true }` — which is what imports it.

Putting the rule here means the one filter (`filterChatSelectableModels`) that `sync-models` already goes through — it sends `x-omniroute-model-surface: chat` — and that `managedModelImport` also calls, both pick it up. No new filtering path.

## Two deliberate choices

**Exact `:batch`, not "has a variant suffix".** `:free`, `:nitro`, `:floor`, `:online`, `:extended` and `:thinking` are routing hints on the same chat model. Excluding them would silently shrink the routable catalogue — a worse failure than the one being fixed, because it is invisible.

**Applied unconditionally for this provider**, unlike the OpenAI branch below it, which lets an explicit multi-endpoint declaration win. There is no "batch" endpoint name an upstream could declare next to a chat one, and the rows already stored for these variants carry the synthetic `["chat"]` default — letting that win would keep importing exactly what this excludes.

## Tests

Five, in `tests/unit/model-endpoint-policy.test.ts`:

- `:batch` variants classify as `non-chat` / not selectable
- `:free`, `:thinking`, `:nitro`, `:online` and the bare base model stay selectable
- a stored `supportedEndpoints: ["chat"]` does not re-admit a `:batch` variant
- the rule is scoped to OpenRouter (`custom-provider` and a `null` provider are unaffected)
- `filterChatSelectableModels` drops the batch variant and keeps its base model

Each earns its place — every mutation is caught by a different one:

| mutation | tests that fail |
| --- | --- |
| remove the OpenRouter branch entirely | the `:batch` classification, the synthetic-default case, the filter case |
| match any `:` variant instead of `:batch` | the other-suffixes case, the filter case |
| drop the provider scope (`if (true)`) | the scoping case |

```
node --test tests/unit/model-endpoint-policy.test.ts   → 12 pass, 0 fail
prettier --check (both files)                          → clean
tsc --noEmit --strict (the changed module)             → clean
```

## Not in scope

The issue also asks for auto-quarantine of models the upstream rejects as "not supported" (the 34 GitHub Copilot `400`s) and a routable-vs-catalog count in the dashboard. Both are policy decisions with their own state, and neither is needed to stop the `:batch` rejections — happy to follow up on either if you want them.


diegosouzapw:
Clean root cause and fix — confirmed `modelEndpointPolicy.ts` has zero batch-variant handling on
the current tip, matching your 91-occurrence production evidence from #13596. Tests pass clean
(12/12). This looks merge-ready as-is.

diegosouzapw:
Thanks @L4XB — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

Issue #13621, opened 2026-09-14
fix(resilience): Large Claude sessions retain request backing strings and trigger resource-pressure 503s


jbovard2016:
## Summary

Long-running OmniRoute processes serving large Claude `/v1/messages` sessions show rapid memory growth and eventually reject otherwise reasonable requests with:

```text
API Error: 503 Service temporarily unavailable due to resource pressure. Retry shortly.
```

This is not only a stale admission-state problem. Profiling found completed request previews retaining their multi-megabyte V8 sliced-string backing storage.

## Observed production behaviour

OmniRoute 3.8.51 on Node/Next.js, configured with a 12,288 MiB V8 heap and `8 + 4` heavy-request admission slots:

- after 2h33m: 7.63 GiB RSS plus 2.12 GiB swap
- after 3h02m: 9.67 GiB RSS plus 2.64 GiB swap
- systemd recorded a 10.1 GiB memory peak and 2.7 GiB swap peak
- `/api/monitoring/health` still reported healthy
- failures recur under larger sessions after hours rather than at a fixed timer

No production heap snapshot was taken.

## Reproduction and retaining path

An isolated local-provider fixture exercised the real `/v1/messages` route with deterministic 100k, 300k and 600k-token-equivalent request bodies, concurrency 1, streaming enabled and no cancellation.

Before detachment, each completed request remained reachable through `completedDetails` as both `clientRequest` and `providerRequest`. V8 heap snapshots showed context-sized backing components of:

- 100k equivalent: 400,008 bytes
- 300k equivalent: 1,200,008 bytes
- 600k equivalent: 2,400,008 bytes

The visible previews were only about 1.2 KiB, but `String#slice()` retained the original backing strings. The process-global completed-details map was capped by 256 entries and a 120-second TTL, but had no byte budget.

All admission leases, queued bytes and in-flight byte counters returned to zero after each request, so those counters were not the retaining owner in this fixture.

## Proposed fix

1. Materialize/detach completed request previews before storing them.
2. Add an estimated byte budget in addition to the existing count and TTL limits.
3. Schedule enrichment/cleanup only when an entry was actually stored.
4. Guarantee diagnostic heap snapshots are deleted on worker, analyzer and timeout failure paths.
5. Add a regression that distinguishes a 4 MiB sliced backing string from a detached ~1.2 KiB preview using V8 `self_size` and backing edges.

[...]

PR #13623, opened 2026-09-14
fix(usage): detach completed request previews

jbovard2016:
## Summary
- detach completed request previews so V8 sliced strings cannot retain full large-session request backing storage
- add a 16 MiB estimated completed-detail payload budget alongside the existing count and TTL limits
- schedule enrichment/cleanup only when the completed detail was actually stored
- add a credential-isolated `/v1/messages` memory profiler with guaranteed raw-snapshot cleanup

## Evidence
- real `/v1/messages` profiles at 100k, 300k and 600k token-equivalent inputs: `largeBackingRetained=false` in every final case
- Node 24 focused tests: 25/25
- Node 24 Vitest: 51 files, 473 tests passed
- `typecheck:core` and `typecheck:noimplicit:core`: passed
- full unit and lint failures reproduced on exact base and are unrelated
- independent Standards and Spec reviews: ACCEPT

## Security
Profiler children receive an allowlisted environment. Raw heap snapshots are analyzed locally and deleted on success, failure, invalid arguments and timeouts. No raw snapshot is committed or uploaded.

Fixes the retaining path reported in #13621. Separate stale-pressure recovery remains in #13618.


jbovard2016:
Local API pilot deployment evidence (14 Sep 2026): integrated commit `d049af25` packaged through the backend-only API build fallback because canonical `build:release` is base-red under #12732. Exact npm tarball SHA-256 `361406a778e554826802c8d894442fc39fcb6a7addad7c815554eb896d073215`; clean-prefix boot passed; deployed bundle hashes and `BUILD_SHA` match. Real `/v1/messages` requests at 50,014 and 100,028 input tokens both returned HTTP 200. The 100k-token request completed in 3.469s; after a 20s drain RSS increased 24,512 KiB and swap stayed zero. A 48-hour qualification is running. This is not yet long-run closure.

diegosouzapw:
The core fix here is solid — `structuredClone` correctly breaks the V8 sliced-string
retention path, the byte-budget addition alongside the existing count/TTL limits is a sound
defense-in-depth, and the 14 tests directly exercising the fix (`active-request-stream-chunks-lifecycle.test.ts`)
pass cleanly. One thing worth a second look before merge: running the new
`messages-route-memory-profile.test.ts` locally on a fairly loaded shared box, I got 9/11
passing with 2 failures (a standalone-analyzer arg-validation assertion, and the full
`/v1/messages` canary reporting `driver failed`). Given the devbox was under heavy concurrent
load at the time, this reads more like resource contention than a real defect — the PR's own
live-instance evidence (100k/300k/600k profiles + the pilot-deployment comment) is stronger
signal than my one contended run — but it'd be worth a clean rerun (isolated CI or a quiet
box) to confirm before merging, just to rule it out.

diegosouzapw:
Thanks @jbovard2016 — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

Issue #13592, opened 2026-09-13
fix(resilience): requestQueue.maxWaitMs kills in-flight requests post-dispatch (504) — undocumented RATE_LIMIT_MAX_WAIT_MS env


afonsoft:
**OmniRoute Version:** 3.8.50
**Installation Method:** npm (global)
**Operating System:** Linux
**OS Version:** Ubuntu / Linux 6.17.0-1020-oracle (arm64)
**Node.js Version:** v24.16.0
**Provider(s) Involved:** ~15 providers — tokenrouter, bai, nvidia, openrouter, opencode-zen, moonshot, gemini, dahl, openai-compatible-*, and others
**Model(s) Involved:** e.g. `tokenrouter/z-ai/glm-*-free`, `bai/claude-opus-*`, `nvidia/deepseek-ai/*`, `openrouter/deepseek/*`, `opencode-zen/glm-*`
**Client Tool:** Any client routed through OmniRoute

## Description

OmniRoute kills in-flight requests once they exceed the local rate-limit "execution" deadline (`resilienceSettings.requestQueue.maxWaitMs`, persisted default observed at 12000ms). The error itself states the deadline "applies this deadline only after dispatch; it does not bound queue wait" — so a request can wait in queue, get dispatched to a provider slot, and then be aborted mid-execution while the upstream is still legitimately streaming a slow LLM response.

**83 occurrences of `[504]`** in a ~41 h log window (2026-09-12 → 2026-09-13), spread across ~15 providers — the second-largest error class in the window.

Two problems compound this:

1. The deadline semantics are surprising: it measures *execution* time after dispatch rather than *queue wait*, so it cancels calls that are already consuming a provider slot and producing output.
2. The env override `RATE_LIMIT_MAX_WAIT_MS` exists in the shipped bundle (default 15000) but is undocumented — I only found it by reading `dist/.build/next/server/chunks`. Meanwhile the persisted `resilienceSettings.requestQueue.maxWaitMs` in `key_value` silently wins, so setting the env var alone does not change observed behavior.

## Steps to Reproduce

1. Route requests to slow upstreams (e.g. `tokenrouter/z-ai/glm-5.3-free`, `bai/claude-opus-4.7`) under normal load.
2. Observe `[ERROR] [504]: Request exceeded OmniRoute's local rate-limit execution expiration (legacy resilienceSettings.requestQueue.maxWaitMs=12000ms)` once the upstream takes >12 s to finish.

## Expected Behavior

[...]

PR #13624, opened 2026-09-14
docs(resilience): describe the queue-wait and execution deadlines as they are

L4XB:
Refs #13592 — documentation only, no behaviour change.

## What I found

The issue reports that the local deadline "measures *execution* time after dispatch rather than *queue wait*, so it cancels calls that are already consuming a provider slot". That is not what `rateLimitManager.ts` does on this branch — but it **is** what the docs say, in two different and both incorrect ways:

| file | current text |
| --- | --- |
| `docs/architecture/RESILIENCE_GUIDE.md` | "`maxWaitMs` is a legacy persisted name for execution expiration… bounds limiter-managed execution, not time spent in the local queue… Queue residence has no time deadline" |
| `docs/reference/ENVIRONMENT.md` | "Max time to wait on a 429 before failing the request" |

What the source actually does:

- **`maxWaitMs` is the queue-wait budget.** `resolveRequestQueueMaxWaitMs` → `queueBudgetMs` → `remainingForQueue` arms a rejection timer, and `wrappedFn` clears it *"once the job leaves QUEUED and starts executing"*. A request that exceeds it never reaches the upstream. So "queue residence has no time deadline" is exactly inverted.
- **`executionMaxWaitMs` is what Bottleneck gets as `expiration`** (default 600000ms), fed by `resolveExecutionMaxWaitMs` and raised to the executor's own fetch-start timeout when that is longer. The source comment is explicit: the expiration is *"never [fed] by the queue-wait budget"*.
- Neither has anything to do with waiting on a 429.

So the surprising semantics the issue describes are the semantics the guide documents. The reporter is on 3.8.50; what remains actionable on this branch is that the docs still describe the pre-split behaviour.

## Also corrected

- The guide's "1–30000ms UI ceiling" belongs to a different setting — `comboCooldownWaitSettingsSchema.maxWaitMs` is the one bounded at 30000. `requestQueueSettingsSchema.maxWaitMs` is `min(1)` with no max, normalised to 1ms–24h by `normalizeRequestQueueSettings`.
- **Precedence is documented for the first time**, which is the issue's second complaint ("setting the env var alone does not change observed behavior"): `RATE_LIMIT_MAX_WAIT_MS` supplies the *default* that `DEFAULT_RESILIENCE_SETTINGS` is built from; a value persisted in `resilienceSettings.requestQueue` wins over it (`normalizeRequestQueueSettings(record, fallback)` prefers `record`); and a per-connection `rateLimitOverrides` value wins over that (`resolveOverride(override, …)`). That chain is correct — it just was not written down anywhere.

## Not in scope

The issue's other asks — moving the deadline, splitting the semantics further, or exempting a stream that already received upstream headers — are behaviour changes the split already addresses on this branch, so I have not touched the code. If the 504s persist on 3.8.51 with the defaults, that is a different defect from the one the issue describes and worth reopening with a log line from the new build (the message now names `executionMaxWaitMs`, not `maxWaitMs`).

Only the English sources are edited; the `docs/i18n/**` copies are left to the translation workflow.


diegosouzapw:
Great catch on the maxWaitMs/executionMaxWaitMs inversion — cross-checked every claim against
`rateLimitManager.ts` and `normalize.ts` and it all holds, including the "1ms-24h clamp"
correction (confirmed in `settings/normalize.ts:133-140`). One thing to fix before merge: this
PR was opened a few hours before #13553 landed and bumped the actual `maxWaitMs` default from
15s to 30s, so "Default 15000ms" in `RESILIENCE_GUIDE.md` and the `15000` table cell in
`ENVIRONMENT.md` are now stale — could you bump both to 30000/30s (or drop the literal number
given it just moved once already)?

L4XB:
Both bumped, and I took the second half of your suggestion too: the guide now names where the number comes from instead of only restating it.

Verified at source rather than from #13553's description: `Number(process.env.RATE_LIMIT_MAX_WAIT_MS || "30000")` at `src/lib/resilience/settings.ts:46`, and `tests/unit/ratelimit-admission-control-6593.test.ts:180` asserts `DEFAULT_REQUEST_QUEUE_MAX_WAIT_MS === 30000`. So the `ENVIRONMENT.md` cell is `30000` (30s), and the guide reads:

> Default 30000ms, supplied by `DEFAULT_REQUEST_QUEUE_MAX_WAIT_MS` in `src/lib/resilience/settings.ts` and pinned by `tests/unit/ratelimit-admission-control-6593.test.ts`, so a change to it turns that test red rather than leaving this paragraph quietly stale.

That is the middle road between keeping the literal and dropping it. A reader still gets the number, and the next person who moves it gets a red test pointing here instead of silently outdating a paragraph, which is what happened this time.

One thing I found while editing the row. The `RATE_LIMIT_MAX_WAIT_MS` source column pointed at `open-sse/services/rateLimitManager.ts`, which never reads that variable:

```
$ grep -rn "process.env.RATE_LIMIT_MAX_WAIT_MS" --include='*.ts' . | grep -v node_modules
src/lib/resilience/settings.ts:46
```

Corrected on the row this commit already touches. Its two neighbours, `RATE_LIMIT_EXECUTION_MAX_WAIT_MS` (`settings.ts:56`) and `RATE_LIMIT_MAX_QUEUE_DEPTH` (`settings.ts:63`), carry the same wrong pointer, and `RATE_LIMIT_AUTO_ENABLE` right below them is the one that genuinely does live in `rateLimitManager.ts:141`. I left those two alone to keep this diff to what you asked for. Say the word and I will fix them here or open a separate PR.

Branch is also merged up to `release/v3.8.51`, so the conflict is gone.

diegosouzapw:
Thanks @L4XB — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #13626, opened 2026-09-14
fix(providers): clamp SenseNova DeepSeek V4 Flash effort to high

HouMinXi:
## Current

SenseNova Token Plan still advertises `xhigh` for `deepseek-v4-flash`, then every request with `reasoning_effort=xhigh` or `max` returns HTTP 400: `ReasoningEffort invalid, one of: low, medium, high, xhigh, none`. Live combo `sn-deepseek-flash` is an openai-compatible connection (`snova/deepseek-v4-flash`), not the first-party `sensenova` registry id.

## Problem

The sanitizer treats DeepSeek V4 as a native `max` family and rewrites `xhigh` to `max` before any per-model clamp. SenseNova then 400s both values. Clients that default to `xhigh` (Hermes) never reach a working request even though `none`/`low`/`medium`/`high` and a 42k-token review-shaped prompt already succeed.

## Change

Clamp `xhigh` and `max` to `high` for SenseNova DeepSeek V4 Flash before the generic max-tier rewrite. Match first-party `sensenova` and the live `snova/` openai-compatible prefix. Leave `glm-5.2` and other DeepSeek V4 hosts (cmd / opencode-go / ollama-cloud) on their existing `max` mapping.

## Verify

- `npx tsx --test tests/unit/sensenova-reasoning-effort.test.ts tests/unit/base-executor-sanitize-effort-max-mapping.test.ts tests/unit/sensenova-provider.test.ts` — 18/18
- Injection: delete the clamp call, the two SenseNova tests FAIL (`xhigh`/`max` stay `max`); restore, 3/3 PASS
- Direct sanitizer map after the change: `snova/deepseek-v4-flash` `xhigh`/`max` → `high`; `high` stays `high`; `sensenova/glm-5.2` `max` stays `max`; `cmd/deepseek/deepseek-v4-flash` `xhigh` still → `max`

## Notes

This PR is submitted for review and merge. Live SenseNova probe (omit effort / `high`): HTTP 200. `xhigh`/`max`: 5/5 steps HTTP 400 before this change. OpenCode Zen `deepseek-v4-flash-free` is a separate upstream delist and is not in this PR.


diegosouzapw:
Solid, narrowly-scoped fix with a real live probe backing the 400s. Ran the touched-area
suite locally: 18/18 pass (sensenova-reasoning-effort + base-executor-sanitize-effort-max-mapping
+ sensenova-provider), including the guard that other DeepSeek V4 hosts (cmd/opencode-go/
ollama-cloud) keep their existing max mapping. One small nit before merging: the new
`isSenseNovaDeepSeekV4Flash` clamp block in `reasoningEffort.ts` is indented one level
deeper than the surrounding code — a quick Prettier pass on that file would fix it.

HouMinXi:
Thanks for the review and the in-place indentation fix in `86eb975fa1`. I fetched that commit and will preserve it when updating against the current release branch; I will not replace it with my stale local head.

PR #13617, opened 2026-09-13
fix(devin): fall back to CLI probe when the HTTP API rejects a CLI-format key

patrykkopycinski:
## Problem

The `devin` provider's routing path (`open-sse/executors/devin-cli.ts`) shells out to the Devin CLI binary and passes the connection's apiKey as `WINDSURF_API_KEY` — it never calls `api.devin.ai`. CLI-format keys (`apk_user_…`) are rejected by the HTTP API, so `validateDevinCloudAgentProvider` returned **Invalid API key** on every working connection: permanent red status on the dashboard while real requests succeed.

## Fix

On 401/403 from the HTTP probe, fall back to spawning the exact probe the executor uses:

```
devin acp --agent-type summarizer   # key passed via WINDSURF_API_KEY env
```

Exit 0 → key valid (with a warning noting the HTTP API rejected it). Probe fails or CLI missing → invalid, as before.

Notably `devin models list` does **not** honor `WINDSURF_API_KEY` (it demands `devin auth login`), so `acp` is the only CLI probe that proves what routing actually depends on.

## Test hardening

The three 401/403 tests in `devin-cloud-agent-validator-6142.test.ts` now pin `CLI_DEVIN_BIN` to a nonexistent path (with save/restore) so the fallback fails deterministically — previously they would consult the host machine's real devin install and flip results per environment.

Also prunes the `webProvidersB.ts` `no-unused-vars` suppression fixed in passing (dead `providerSpecificData` destructure in `validateInnerAiProvider`).

## Verification

- `devin-cloud-agent-validator-6142` + `validation-web-providers-split` + `devin-providers`: **19/19 pass**
- `npm run typecheck:core`: clean
- `npm run lint` on touched files: clean
- Live: patched gateway container, connection-test now returns `valid: true` with the CLI-fallback warning; e2e `devin/swe-2-max` completion 200

PR #13627, opened 2026-09-14
fix(cursor): kv_after_text must not settle away a trailing exec_mcp tool call

patrykkopycinski:
## Summary

Composer-2.5 turns intermittently end with `content: null`, reasoning present, and **zero `tool_calls`** despite the model having decided to call a declared tool. Cursor surfaces this as an empty-content `finish=error` turn — "nothing to recover" since there is no narration text to parse (unlike the Kimi #10215 narration leak, which is already recovered textually).

## Root cause

`driveH2`'s scan loop settles the response promise on the **first** `ctx.endReason`. For composer models, `kv_server_message` after text sets the speculative `kv_after_text` terminator. Under load, the `exec_mcp` tool-call frame can share the TCP segment with — or trail by a partial frame — the KV checkpoint. Settling at the checkpoint splices that frame off as "leftover" and it is never processed for this response: the client receives a turn with no tool call, and the agent loop dies.

The invariant claimed in the code comment ("exec_mcp always arrives at or before the KV checkpoint") does not hold under load — verified by wire capture on a live gateway (reasoning frame → KV checkpoint → exec_mcp in one buffer, tool call dropped).

## Fix

- `kv_after_text` is treated as a **soft** terminator in the scan loop: when bytes remain buffered, scanning continues instead of settling.
- A bounded grace timer (`KV_GRACE_MS`, default 2000 ms, env-tunable via `CURSOR_KV_GRACE_MS`) ends the turn if no further frame completes — plain-chat latency cannot regress to the full stream timeout.
- Hard terminators (`tool_calls`, `turn_ended`, `server_end`) keep settling immediately — no behavior change on non-racer paths.
- A late `exec_mcp` also upgrades `ctx.endReason` to `"tool_calls"` in `processFrame` (defense in depth when the event decodes after the checkpoint in the same pass).

## Test

`tests/unit/cursor-composer-kv-toolcall-race.test.ts` drives the **private `driveH2`** through a fake h2 stream with the exact wire layout (text_delta → kv_set_blob → exec_mcp concatenated in one buffer) and asserts the tool call survives. Mutation-checked: reverting the executor change alone turns the test RED; with the fix it is GREEN. A plain-chat control asserts the kv_after_text path still settles (void, like `onEnd`) when the buffer ends clean.

## Notes

- Reproduced live on a production gateway (1-in-N intermittent, frame-segmentation dependent); fix verified end-to-end on a debug clone with frame dumps.
- This hardens the mechanism introduced by #10215/#10502 rather than replacing it.


diegosouzapw:
Thanks for the sharp diagnosis and the wire-capture repro — this is a well-scoped fix and
the new test genuinely exercises the race (verified it green against the PR head). Two
small asks before merge: (1) add a `changelog.d/fixes/` fragment (none included), and (2)
a quick rebase onto the current `release/v3.8.51` tip once you get a chance. No functional
concerns from my side.

patrykkopycinski:
Addressed both review asks:

1. **Rebase onto release/v3.8.51 tip (c3945a724)**: branch already had `c3945a724` as its merge-base (`test: adapt kv-race test to upstream void driveH2 contract` was committed directly on top of it), so `git rebase origin/release/v3.8.51` was a no-op ("Current branch is up to date"). Confirmed via `git merge-base --is-ancestor origin/release/v3.8.51 HEAD`.
2. **Changelog fragment added**: `changelog.d/fixes/13627-cursor-kv-after-text-mcp-toolcall.md`, following the format in `changelog.d/README.md`.

**Tests** (PR-owned test file `tests/unit/cursor-composer-kv-toolcall-race.test.ts`):
```
DISABLE_SQLITE_AUTO_BACKUP=true node --max-old-space-size=8192 --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test tests/unit/cursor-composer-kv-toolcall-race.test.ts
```
Result: `# tests 2 / # pass 2 / # fail 0`.

**Changelog gate**: `node scripts/check/check-changelog-integrity.mjs` — my fragment passes. Only remaining failure is the pre-existing `changelog.d/fixes/reset-aware-model-family.md` (missing leading `- `, from merged #12637) — unrelated base debt, not introduced by this PR.

Pushed to `fork/fix/composer-kv-toolcall-race` at `f01a4de58` (fast-forward from `8ae2315d9`).

patrykkopycinski:
Follow-up in 1e47676a2: the env-doc-sync gate correctly flagged `CURSOR_KV_GRACE_MS` (introduced by the kv-grace fix) as in-code-but-undocumented. Documented in `.env.example` (commented default `2000`, semantics from the executor comment) and `docs/reference/ENVIRONMENT.md` (next to `CURSOR_STREAM_TIMEOUT_MS`). `node scripts/check/check-env-doc-sync.mjs` → in sync locally; CI re-running on the new head.

patrykkopycinski:
CI attribution (cross-posted to #13355 #13448 #13617 #13627 #13359): the red Fast Quality Gates on these heads are pre-existing base debt, visible on the base branch's own runs with no PR diff applied:
- `mutation-test-coverage`: run 34970965604 (base head 1e1af286) fails "2 covering unit test(s) across 2 module(s) are missing from stryker.conf.json" (incl. tests/unit/noauth-model-lockout.test.ts) — same signature on our PR runs.
- `secrets` ratchet: "REGRESSÃO — 1 secret findings > baseline 0" reproduces across PRs with disjoint diffs.
- `forgotten-sibling-tests`: "Invalid string length" (RangeError, string overflow) crashes the checker on the agyModels→shared→tests graph — hit 13355 and 13627 hours apart on unrelated diffs.
Base tip FQG itself: last completed run 34970965604 = failure. Happy to PR the stryker testFiles additions if useful.

PR #13636, opened 2026-09-14
fix(resilience): extend process crash guard to combo hedge cancels and upstream fetch failures

HouMinXi:
# Extend the process crash guard to combo hedge cancels and upstream fetch failures

Stacked on #12942: this PR includes that branch plus two follow-ups, so it
can land either after it or instead of it. Once #12942 merges I will rebase;
the delta then shrinks to the last two commits.

## Why

We run OmniRoute in production with an upstream (agnes-ai.cn) that spent a
morning flapping: repeated response-start stalls and TLS handshake resets.
With #12942 deployed we watched the process-level guard do its job. Three
`DIRECT_RESPONSE_START_TIMEOUT` rejections arrived while nobody was awaiting
them, got logged, and the process stayed up.

But within 30 minutes the same storm killed the process twice more through
two sibling escape hatches the guard's predicate does not recognize:

1. `Error [AbortError]: hedge-cancelled` (exit 7 at 06:11). A combo hedge
   fired while three legs were stalled at ~162s; the winning leg completed
   the request 200, and the AbortError from cancelling the losers surfaced
   with no listener. The name is `AbortError` but the message is
   `hedge-cancelled`, so the `/abort/i` message check misses it.
2. `[TypeError: fetch failed]` with `code: 'PROXY_UNREACHABLE'` (exit 7 at
   06:27), cause `ECONNRESET` before the TLS handshake. Neither the code
   list nor the timeout predicate covers it.

Playing whack-a-mole with individual error identities is the wrong layer for
this class. The invariant we want: an intentional cancellation of a hedged
combo leg, or a network/IO failure against an upstream, is a runtime
condition the request layer already handles. It should never be a
process-fatal event.

## What

`src/shared/utils/httpClientAbortGuard.mjs` gains two predicates, both wired
into `shouldSwallowUncaught`:

- `isIntentionalComboAbort` matches `AbortError` whose message or string
  cause is exactly `hedge-cancelled` or `combo-per-model-timeout` (the two
  intentional reasons from `open-sse/services/combo/comboAbortReasons.ts`,
  duplicated as literals because the guard must stay build-free). An
  AbortError from anywhere else still crashes loudly.
- `isUpstreamNetworkError` matches undici's fixed `TypeError: fetch failed`
  wrapping message, plus explicit network codes (`PROXY_UNREACHABLE`,
  `UND_ERR_*`, `ECONNREFUSED`, `EHOSTUNREACH`, `ENETUNREACH`, `EAI_AGAIN`).
  Arbitrary TypeErrors still crash loudly.

The second commit is the `assembleStandalone.mjs` manifest entry for the
guard module, which #12942 needs to boot in the production image. Its own
regression test (`tests/unit/build/assemble-standalone.test.ts`) fails on
that branch without it; details in my comment on #12942.

## Tests

- `tests/unit/http-client-abort-guard-direct-timeout-12861.test.ts` extended
  to 12 tests covering both production crash shapes (message and cause
  variants), negative cases (unknown AbortError, genuine TypeError, null,
  non-objects), and both `uncaughtException` / `unhandledRejection` origins.
- Red/green verified: the new tests fail against the #12942 guard as
  shipped, pass with this change, and neutering the `fetch failed` branch
  flips the matching test red again.
- Full touched-area run: 24/24 across the two guard suites and the
  assemble-standalone regression.

## Review

This branch went through a multi-round adversarial review (three independent
passes per round, then a falsification pass over every candidate finding).
Across five rounds the reviewers produced 9 unique findings: 8 were dismissed
with file:line counter-evidence (mostly the "unknown errors must stay fatal"
negatives, which the tests pin deliberately), and 1 was confirmed: a module
docstring that said "a second category" where chronology made it the third.
Fixed in this branch. No correctness findings survived.


HouMinXi:
## Re-review update (head 9d519c466)

A second review cycle on the amended head produced four follow-ups, all landed or adjudicated:

**Fixed (each with a regression test + revert-to-red injection proof):**
- `isIntentionalComboAbort` now accepts a bare string abort reason — `AbortSignal.reason` is whatever `abort()\) received, and a raw string reason rejects waiters with the string itself, not an Error.
- Swallowed errors are logged with the full error object instead of `code ?? message`, so the only evidence of a swallowed event keeps its stack.
- `isRecoverableUpstreamTimeoutError` mirrors the same bare-string tolerance for `DIRECT_RESPONSE_START_TIMEOUT` (closed vocabulary, one line).
- Repaired `httpClientAbortGuard-default-logger.test.mjs`, which the parent change broke by renaming the log text it greps for — red on the parent alone, verified by stashing.

**Dismissed with evidence:**
- `isUpstreamNetworkError` matching top-level errno codes is a deliberate trade-off: narrowing to the undici wrapper shape risks missing unwrapped `UND_ERR_*` socket errors — the exact crash class this guard exists to contain. With full-error logging now in place, a false swallow stays supervisor-visible. None has been observed since deploy.
- The AbortError name-vs-code discriminator question is answered by production: since this guard deployed, it has swallowed the live shape six times (including two `hedge-cancelled` rejections) with zero process deaths.

Suites: 35/35 across the five guard-related test files.

diegosouzapw:
Thanks for extending the crash-guard coverage to combo hedge cancels, upstream network errors,
and the standalone bundle — verified all 33 tests pass and the new coverage is real. One thing we
need to sort out before merging: the first commit here reproduces PR #12942's fix
(`directResponseStartTimeout.ts` + its test file) byte-for-byte, but it's attributed to a generic
dev identity rather than to @geekyNads, who opened that fix a week earlier. We'd like to merge
#12942 first so they get credit for the original work, then rebase this PR to carry only your
extension (the combo hedge cancel + upstream network error coverage + the standalone-bundle build
fix) on top of it. Happy to help with that rebase if useful — the extension work itself is solid
and we want it in.

HouMinXi:
@diegosouzapw Agreed on authorship and merge order. #12942 should receive credit for the original direct-response timeout fix. My follow-up should contain only the additional combo/network coverage and standalone packaging work, rather than re-attributing that foundation. I am reconciling the extension against the current release tip and will report the exact remaining diff and validation before asking you to land it.

PR #13637, opened 2026-09-14
fix(compression): fall back in-process when the compression worker fa…

### LINKED ISSUE #13145: [BUG] Compression plan selected/announced but never applied — tokens-in identical with off, compression_analytics stays empty (v3.8.50 and :next)
### OmniRoute Version

v3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Fedora 39

### Node.js Version

_No response_

### Provider(s) Involved

Deepseek

### Model(s) Involved

deepseek flash v4, deepseek flash v4 ga

### Client Tool

Opencode

### Description

Compression appears enabled and active (response header reports stacked), but prompts are sent to the upstream provider uncompressed — upstream token usage is identical whether compression is set to off or stacked. As a result, no token savings occur and the compression analytics tables (compression_analytics, compression_engine_breakdown, compression_cache_stats) remain at zero rows even after many requests.
This is reproduced on both the stable v3.8.50 release and the :next line (v3.8.51), so it does not appear to be a pre-release regression.

### Steps to Reproduce

Steps to reproduce
1. Start OmniRoute with compression enabled and a stacked pipeline configured (RTK → Caveman), e.g.:
- defaultMode: "stacked", autoTriggerMode: "stacked", pipeline rtk(standard) → caveman(full)
- API key with compression_enabled = 1
2. Create a highly compressible prompt (verbose, repetitive prose that Caveman targets):
The function should iterate over the list of items, and for each item, it should check whether the value is greater than zero, and if so, add it to the running total.  (repeated 150 times)
3. Send the same request twice, once with compression off and once with stacked:
# Run 1 — compression off
curl -s -X POST "$BASE/v1/chat/completions" \
  -H "Authorization: Bearer $KEY" \
  -H "x-omniroute-compression: off" \
  -d '{"model":"deepseek-v4-flash-ga-260731","messages":[{"role":"user","content":"<the repeated prompt>"}],"max_tokens":1}'

# Run 2 — compression stacked
curl -s -X POST "$BASE/v1/chat/completions" \
  -H "Authorization: Bearer $KEY" \
  -H "x-omniroute-compression: stacked" \
  -d '{"model":"deepseek-v4-flash-ga-260731","messages":[{"role":"user","content":"<the repeated prompt>"}],"max_tokens":1}'
(Replace $BASE with your gateway origin + base path, and $KEY with a gateway API key.)
4. Compare the x-omniroute-tokens-in response header between the two runs.
5. After several requests, read the analytics endpoint: GET $BASE/api/context/analytics
Expected behavior
- Run 2 (stacked) should show x-omniroute-tokens-in significantly lower than Run 1 (off) — the prompt should be compressed before reaching the provider.
- /api/context/analytics should show totalRequests > 0 and totalTokensSaved > 0.
Actual behavior
- Both runs report the identical x-omniroute-tokens-in value (5634 in our test) — the prompt is sent upstream unchanged in both cases.
- The response header still announces the plan: x-omniroute-compression: stacked; source=active-profile.
- Analytics remain all zeros after multiple requests:
off     → x-omniroute-compression: off; source=request-header      tokens-in: 5634
stacked → x-omniroute-compression: stacked; source=active-profile  tokens-in: 5634
/api/context/analytics → totalRequests: 0, totalTokensSaved: 0, byMode: {}
Environment
- OmniRoute 3.8.50 (:latest) and 3.8.51 (:next, built 2026-09-08) — reproduced on both
- Runtime: Podman container, Node 26 standalone, OMNIROUTE_BASE_PATH=/omniroute
- Route: /api/v1/chat/completions, target format OpenAI, routed bpm → byteplus for deepseek-v4-flash-ga-260731
- DB verified writable (manual INSERT INTO compression_analytics succeeds)
Additional context / possible cause
open-sse/handlers/chatCore.ts gates the analytics write on result.stats:
if (result.stats) {
  // ... writeCompressionAnalytics / writeCompressionSkip
}
Here the plan resolves (mode = "stacked", response meta emitted via formatCompressionMeta), but applyCompressionAsync appears to return without stats and without modifying the body — so no analytics row is written and no tokens are saved. Suggest tracing applyCompressionAsync for the chat-completions / OpenAI target path to see where it bails before producing stats.

### Expected Behavior

Expected behavior
1. Compression actually reduces the prompt before it reaches the provider. For the same request, x-omniroute-tokens-in with x-omniroute-compression: stacked should be noticeably lower than with off — e.g., a 150×-repeated verbose sentence should compress from ~5,600 tokens to well under ~2,000 tokens (Caveman targets exactly this kind of filler-heavy prose).
2. The response header reflects what actually happened. A real compressed response should carry annotations beyond the plan name — e.g. x-omniroute-compression: stacked; source=...; tokens=5634→1200; rules: ... — not just stacked; source=active-profile.
3. Analytics record the runs. After the two test requests, GET /api/context/analytics should show:
- totalRequests ≥ 2 (or at least > 0)
- totalTokensSaved > 0
- byMode populated with a stacked entry, and byEngine with rtk / caveman
4. A no_savings row is written when compression legitimately finds nothing to compress, so "ran but saved nothing" is distinguishable from "never ran" (#4268 semantics).

### Actual Behavior

1. Both runs report the identical x-omniroute-tokens-in (5634 in our test) — the prompt is sent upstream unchanged in both cases.
2. The response header still announces the plan: x-omniroute-compression: stacked; source=active-profile — with no tokens=... / rules: ... annotations.
3. Analytics remain all zeros after multiple requests:
off     → x-omniroute-compression: off; source=request-header      tokens-in: 5634
stacked → x-omniroute-compression: stacked; source=active-profile  tokens-in: 5634
/api/context/analytics → totalRequests: 0, totalTokensSaved: 0, byMode: {}
4. Database writes are healthy (manual INSERT INTO compression_analytics succeeds); the read endpoint returns 200 with an all-zero summary.

### Test Impact

### Error Logs / Output

```shell
1. No actual token savings — users pay full upstream prompt cost on every request despite compression being enabled and announced. The feature's primary value (15–95% savings) is silently absent.
2. Misleading observability — the response header and dashboard settings claim compression is active, so operators cannot tell the feature is broken without a side-by-side token-count comparison. This makes the bug easy to ship to production undetected.
3. Analytics are unusable as a regression signal — since totalRequests/tokensSaved never increment, there is no way to detect, monitor, or alert on compression failures; a future fix would also have no baseline to compare against.
4. Blocks any evaluation of compression quality — the eval harness and per-engine breakdowns depend on the same stats that are never produced, so fidelity-vs-savings validation cannot run on this path.
5. Risk of false confidence in capacity planning — cost projections based on the claimed savings are wrong by the full uncompressed amount.
```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

marcs7:
Fixes #13145 
The bug

CompressionWorkerPool resolved every worker fault with the uncompressed body instead of reporting it. PendingJob had no reject path at all, so four distinct failures — a worker thread error, a worker exit, a dispatch timeout, and an engine error posted back as type: "error" — all resolved as { compressed: false, stats: null }.

applyCompressionAsync then read that as a legitimate "nothing to compress" result and returned it unchanged. The request reached the provider uncompressed while the response header still announced the selected plan (x-omniroute-compression: stacked), because the header is emitted before the pipeline runs. Nothing was logged at any level, and compression_analytics stayed empty because rows are only written when a compressed result is reported.

Net effect: compression silently disabled for every worker-eligible request, exactly as reported in #13145.

Why it is not visible in a naive repro

runCompressionAsync forwards its options into workerOptions, and isStrictlySerializable rejects an object holding undefined values. A synthetic call that leaves sourceFormat/targetFormat/compressionStage unset is therefore not worker-eligible, silently takes the in-process path, and compresses correctly. Production requests always carry all of those fields, which is why the worker path is taken there and the bug only shows up in real traffic.

The fix

The worker is a throughput optimisation, not a behavioural variant, so a worker fault must degrade to the in-process pipeline rather than to no compression at all.

- PendingJob gains reject. fail() delegates to a new abort() that clears the slot timeout and rejects with a diagnostic cause (thread error, exit code, or timeout budget).
- An error message from the worker is propagated instead of being swallowed.
- applyCompressionAsync catches the rejection and falls through to the in-process path, logging the cause. The logger is imported lazily and defensively: compressionWorker.ts imports that module, so a static import would pull the logger into the worker bundle, and a logging failure must never be able to break compression itself.

close() keeps resolving with the unchanged body — shutdown is not a fault.

Test

tests/unit/13145-compression-worker-failure-fallback.test.ts drives a real worker fault via OMNI_COMPRESSION_WORKER_TIMEOUT_MS rather than mocking the module, since this project's tsx/ESM + node:test setup has no mock.module() support. Its options are fully populated on purpose, for the reason above — otherwise the test would route through the in-process path and assert nothing. A first assertion guards that premise by checking the pipeline really is worker-eligible.

Verified to fail on the unpatched tree with a worker failure must fall back to in-process compression, not return the body uncompressed, and pass with the fix.

Checks

- 135 tests across the 30 compression/worker unit files: all pass
- eslint clean on the touched files
- tsc --noEmit: 0 errors
- prettier applied; pre-commit hooks (docs-sync, explicit-any budget, tracked-artifacts) all green


emp-zf:
Confirmed the silent swallow on the stable wiring — try { …runCompressionInWorker… } catch { return { body, compressed: false, stats: null } } is present verbatim in the compiled :next build too (chunk 16904, the compiled applyCompressionAsync), so the workerPool + strategySelector changes here are correct and the in-process fallback is the right semantic (the worker is a throughput optimization, not a behavioral variant).
However — heads-up for anyone on :next: this PR alone doesn't restore compression there. I traced the compiled :next image (v3.8.51) and the failure is two-layered:
1. The layer this PR fixes — worker faults silently resolving unchanged(body) with no reject path. Present in :next as well.
2. A second, :next-only layer — the entire chatCore compression stage appears to be refactored into a module that lost its importer in the :next build:
- The live request path is app/api/v1/chat/completions/route.js → chunks/80010.js (pipeline handler) → chunks/45121.js (combo relay). In that chain, the only call to applyCompression is the combo-level fallback compression, gated on combo.config.zeroLatencyOptimizationsEnabled && fallbackCompressionMode !== "off" — a separate feature from the API-key compression setting.
- The full per-key compression integration still exists in the build — chunk 88137.js, module 856385, export wk — with selectCompressionStrategy / selectCompressionPlan / applyCompressionAsync all destructured and the analytics writers wired in. A repo-wide grep of the compiled build finds zero importers of module 856385. Nothing calls it.
- The settings half is still wired (the live route reads getCompressionSettings + per-key compressionEnabled and forwards deferContextOverflowWhenCompressible / compressionExclusions down the relay), and the plan echo header (x-omniroute-compression: standard; source=routing-override) is still emitted — so :next announces a plan the pipeline never executes, which is how I hit the same symptom as this issue.
- Cross-check: chatCore's compression block from the stable source (omniglyphSelected, runPostTranslationCompression) doesn't appear anywhere in the :next compiled chunks.
Net effect on :next: the announce-skip reproduction is identical (preview 17,663 → 3,005 tokens / 83% saved; real request byte-identical; compression_analytics 0 rows), but the failure happens before the worker pool is ever consulted — so this PR's fallback never runs.
If the maintainer can confirm, the :next fix needs the compression stage re-attached to the request pipeline (whoever was supposed to import module 856385 during the refactor). This PR remains the correct and sufficient fix for the stable branch — the two changes are complementary, not competing.

diegosouzapw:
Thanks for tracking this down — the diagnosis is exact and the test proves it (verified
fail-without-fix on the release tip). One thing worth a deliberate call before merge: the
fallthrough in `strategySelector.ts` now runs the same heavy in-process compression synchronously
on the main event loop whenever the worker fails, including on a TIMEOUT — which by definition
means the body was too expensive for the worker's own budget. That's strictly better than silently
not compressing, but under load it could block the event loop for other in-flight requests. Could
you confirm that's the intended trade-off (or gate the in-process fallthrough to thread-error/exit/
engine-error only, keeping the old degrade-to-uncompressed behavior for a timeout specifically)?
Also worth flagging: PR #12755 touches the exact same `fail()`/`spawn()` methods for a different
failure mode (synchronous worker-construction throws) with an incompatible "never reject" design —
the two will need to be reconciled/sequenced before either merges. A `changelog.d/fixes/` fragment
is missing too.

marcs7:
Thanks — the timeout concern was right, and chasing it surfaced a defect in my own patch.

On the trade-off: gating it is the correct call, and I've done that. With the default 120s budget, a dispatch timeout means the worker already spent two minutes on that body without finishing; re-running the same CPU-bound pipeline inline would stall every other in-flight request. Worse than not compressing on a shared gateway.

Faults are now typed by whether recovery is cheap. CompressionWorkerError.retryInProcess separates fast faults — thread error, worker exit, engine throw, where no work was done and the in-process path costs what the worker would have — from a dispatch timeout, which keeps the original degrade-to-uncompressed behaviour. The defect this PR fixes is the silent swallow, not the degrade, so timeouts are now reported rather than recovered.

A defect in the first commit, found by writing the timeout test. Adding reject to PendingJob without stripping it from the structured-clone wire job meant a function reached postMessage, which threw DataCloneError before the worker ever saw the job — turning every dispatch into an immediate fault. The original patch "worked" only because it fell back in-process every time. Now fixed; the timeout test would not have caught this if it had been written to pass.

On #12755: I saw @insoln's note there about sequencing this one first. Happy either way — if #12755 lands first I'll rebase and adapt, since the two designs meet in fail(). Worth noting they're closer than they look: #12755 makes fail-open observable, this one makes it recoverable where recovery is cheap, and the retryInProcess split is the seam that lets both hold. Their spawn-time containment and this fault typing compose cleanly — the drain path would just mark construction failures as retryable.

Changelog fragment added: changelog.d/fixes/13637-compression-worker-silent-noop.md.

Also worth flagging for whoever picks up :next — @emp-zf's trace above points at a second, separate layer there (the chatCore compression stage losing its importer in a refactor, chunk 88137 / module 856385 with zero importers). This PR is the stable-branch fix and won't restore compression on :next by itself.

Re-verified after the change: 167 tests across the compression/worker unit files pass, eslint clean, tsc --noEmit at 0 errors.

diegosouzapw:
Thanks @marcs7 — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-130526-suite.log) on the devbox @ train tip 28fb420c9ad860ba275294ebb4ebbecb9da83318 with the sibling PRs of this batch: static gates green; changed-area node:test 293/294 (0 failing) + vitest 482/482 (fast parity — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §7.

PR #13638, opened 2026-09-14
Fix/codex quota header leak

fenix007:
## Summary

## Related Issues

- Closes #
- Related to #

## Validation

- [ ] `npm run lint`

## Tests Added Or Updated

## Coverage Notes

## Reviewer Notes


diegosouzapw:
The isolated commit here (`hide backend account quota headers`) is clean and well-tested, but I
need to flag a direct tension with #10315: the current tip deliberately gives
`x-codex-*-used-percent/reset/window/credits/plan-type` headers forwarding priority specifically
because they used to be evicted by mistake and that broke real Codex CLI quota display for
single-account users. Stripping them unconditionally (as this PR does) would re-introduce that
exact problem for the common case, while fixing a real concern for the multi-account/combo
pooling case you describe.

I'd like to merge the underlying idea, but gated to only strip when the request actually went
through combo/multi-account routing — could you scope it that way? Also, to keep review
tractable, I'd cherry-pick just this one commit onto the tip rather than the full branch (which
carries a lot of history not related to this fix) — happy to do that with full credit to you.

diegosouzapw:
Thanks for raising this — the concern is real: when pool/combo routing picks an account that isn't the caller's, the `x-codex-*` quota headers expose that account's quota. I've opened #14116 to track exactly that, crediting you.

This PR itself can't merge as-is, for two reasons that aren't about the idea:
1. The **unconditional** strip regresses #10315 — `chatCore/responseHeaders.ts` forwards those headers on purpose on the direct path (dropping them there was a previous bug), and the existing test `middleware-header-strip-5849.test.ts › "streaming path keeps Codex quota headers…"` fails with this change (8/9). The fix has to be **conditional**: strip only when the selected connection differs from the caller's own.
2. The branch carries 83 unrelated commits, and the original commit has an AI co-author trailer, which the repo doesn't accept in history (Hard Rule #16 in AGENTS.md).

Closing this one; if you'd like to take #14116 on a fresh branch off `release/v3.8.51` with the conditional strip, you'd have my full support — the design sketch is in the issue.

Issue #10714, opened 2026-08-19
Removing a provider leaves it stuck as "error" on the Home provider topology forever


CarlosPradere:
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Provider(s) Involved

g4f-pollinations (but likely affects any provider)

### Model(s) Involved

No response

### Client Tool

Dashboard only

### Description

After connecting a provider (g4f-pollinations / Pollinations) that returned errors (402), I removed it entirely from Dashboard → Providers. The provider connection was successfully deleted (confirmed empty in `provider_connections` table). However, the Home page's "Provider Topology" widget kept showing the removed provider indefinitely with a permanent red "error" status ("0 activo · 1 error"), even after:

- Clearing `proxy_logs` for that provider
- Clearing `usage_history` for that provider
- Confirming `provider_nodes`, `provider_key_limits`, `registered_keys`, `quota_pools`, `proxy_registry`, `proxy_assignments`, and `model_combo_mappings` had no reference to it
- Fully restarting the OmniRoute process (Apagar → reopen, confirmed via Task Manager that no node/omniroute process remained)
- Clearing browser site data (cookies, localStorage, IndexedDB, service workers)

The root cause turned out to be the **`call_logs`** table (1045 rows in my case), which still had historical rows with `provider = 'g4f-pollinations'` and `status = 402`. The `/api/.../provider-metrics` endpoint (feeding `topology.providers` and `topology.errorProvider`) aggregates from `call_logs`, and this table is **not cleaned up** when a provider is deleted from Providers.

Once I manually ran:
```sql
DELETE FROM call_logs WHERE provider = 'g4f-pollinations';
```
...and restarted, the topology widget correctly updated to "0 activo · 0 error" and the ghost node disappeared.

### Steps to Reproduce

1. Connect a provider that will produce failed requests (e.g. one that returns 402/error status).
2. Make at least one request through it so a row lands in `call_logs` with that provider name and a non-200 status.
3. Go to Dashboard → Providers and delete/remove that provider connection.
4. Go to Home (`/home`) and look at "Provider Topology" — the removed provider is still shown as a node with a red "error" dot indefinitely, even after full app restarts and DB checks of `provider_connections`, `proxy_logs`, `usage_history`, etc.

### Expected Behavior

Deleting a provider connection should either:
[...]

PR #13641, opened 2026-09-14
fix(call-logs): hide search rows without a live provider

maxmad64bis:
## Summary

A search call logged without a provider left a blank row in the provider breakdown, and it stayed visible on the dashboard. The three search queries in `src/lib/db/callLogStats.ts:157-250` didn't check for a live connection the way the provider metrics do (`src/lib/db/callLogStats.ts:110`). They now share that same guard, so only providers with a live connection show up.

## Related Issues

- Stacked on #13281.
- Follows #10714, which fixed the same drift for provider metrics while search kept the old behavior.

## Validation

- [x] Change type: DB
- [ ] `npm run lint` — scoped run only; full-repo lint not replayed locally, will confirm on CI
- [ ] Reconciled with the current active release base; focused checks rerun afterward — reconciled with #13281; rebase + rerun owed at #13281 merge

## Tests Added Or Updated

- New `tests/unit/call-log-search-ghost.test.ts` seeds one live `brave` connection next to a `NULL` row, a `'-'` row and an orphan with no connection, then asserts none of them leak into the per-provider breakdown or the recent list (`src/lib/db/callLogStats.ts:157-250`). Replayed against the pre-fix file it went 0/3, and it's 3/3 after.
- Updated `tests/unit/db-call-log-stats-3500.test.ts` seeds the live connections its search cases already assume, with unconditional asserts on the small providers, and it stays green 11/11.

## Coverage Notes

The new test exercises each branch of the guard, covering `NULL` and `'-'` rows plus an orphan only `EXISTS` can catch, alongside one live provider, through the three queries (`src/lib/db/callLogStats.ts:157-250`). No coverage regression is expected from this, since the only prod change is that shared guard.

## Reviewer Notes

- `getSearchAggregateStats` stays unfiltered on purpose and its JSDoc says so, since its totals still count rows without a live connection and `total` can read higher than the `byProvider` sum.
- No index or migration ships here, as the guard reuses `idx_cl_request_provider` and `idx_pc_provider`.
- Fragment: `changelog.d/fixes/13641-search-ghost.md` with the matching `(#13641)` pull link.


maxmad64bis:
## CI review (§12)

All failures inherited from the base or unstable — none cite this PR's files.

- `Merge integrity`: `cli-tunnel` skill drift (third-party voisin, not this PR's files).
- `Docs Gates`: 356-vs-358 provider count (N90, inherited).
- `Fast Quality Gates`: identical triplet on tierce #13471 same base (mutation 15, cycles SCC `models/providers/deletion`, `autoCombo.test.ts` TS2739) — inherited.
- `ESLint`: 4 hits all outside this PR (`resource-pressure-self-restart`, `volcengine-plan-binding-upsert` ×3).
- `Unit Tests`: zero failing lines cite this PR's files (`callLogStats`, `call-log-search-ghost`, `db-call-log-stats-3500` all absent from `✖` lines). Sampled `sanitizeErrorMessage` ReDoS: green alone, red in lot on pure base (`/tmp/opencode/base-pure`) = unstable. Sampled `combo-strategies` reset-aware: red alone on pure base = inherited.

PR #13643, opened 2026-09-14
fix(codex): whitelist reasoning object keys before the wire

HouMinXi:
## Problem

Client ecosystems send OpenRouter-style fields inside the `reasoning`
object: `enabled`, `max_tokens`, `exclude`, and friends. The Codex
executor forwarded the whole object upstream, and the Codex Responses
API rejects every one of those keys:

```
HTTP 400 Unknown parameter: 'reasoning.enabled'
```

Inside a combo the same invalid body is dispatched to every target, so
one client-side field turns into a full combo outage — "All models
failed" on every connection, surfaced to the user as an empty reply.
Reproduced in production against `codex/gpt-6-astra-high`: plain
requests and `reasoning: {effort: ...}` return 200; any object
carrying `enabled`, `max_tokens`, or `exclude` returns 400 on all
combo targets.

## Fix

`open-sse/executors/codex.ts` now whitelists the `reasoning` object to
the keys the upstream accepts (`effort`, `summary`) before the body
reaches the wire, on both the translated and the native-passthrough
path. Two semantic points:

- `enabled: false` is honored as effort `none` when no more specific
  effort was requested. It wins over the connection default but still
  loses to any per-request selection (model suffix, `reasoning.effort`,
  or flat `reasoning_effort`).
- The strip runs even when no effort was resolved, because the client's
  original object is forwarded unchanged on that path.

This follows the existing precedent in the same function (`truncation`,
`user`, `prompt_cache_retention`, `safety_identifier`, `background`
are already deleted for the same upstream), and the per-provider
sanitizers elsewhere (`opencodeReasoningSanitizer`, `mimoThinking`).

## Sweep for siblings

I audited every executor that reads `body.reasoning`:

- `kimi.ts`, `moonshot.ts` — read, translate, `delete body.reasoning`. Clean.
- `claude-web`, `zai-web`, `chatgpt-web-codex` — read and translate
  into provider-native thinking shapes. Clean.
- opencode-go providers — boolean `reasoning` is already stripped by
  `opencodeReasoningSanitizer`; the object form matches the Go struct. Clean.
- Generic OpenAI-compatible passthrough — forwards `reasoning` by
  design (OpenRouter accepts `enabled`); not a Codex-class leak.
- `codex.ts` — the only executor spreading the client object onto the
  wire. This fix.

## Tests

New `tests/unit/codex-reasoning-wire-whitelist.test.ts` drives the real
`CodexExecutor.transformRequest` (no re-implemented resolution chain):

- `enabled` stripped, explicit effort survives
- `enabled: false` -> effort `none` (no summary emitted)
- `max_tokens` / `exclude` never reach the wire
- client-provided `summary` preserved
- model suffix effort still wins over `enabled: false`
- flat `reasoning_effort` is deleted from the wire body

Red/green verified by reverting the executor change (5 tests fail) and
restoring it (7/7 pass). Existing codex suites: 67/67 pass.

## Review

Multi-round adversarial review (three independent passes per round,
falsification gate on every candidate). Round 1: one finding,
confirmed by all three passes — my comment claimed connection defaults
beat `enabled: false` while the code did the opposite; the code is the
intended semantic (per-request beats connection default), so the
comment was corrected. Round 2: one finding — the flat
`reasoning_effort` strip had no assertion; test coverage added
(`delete body.reasoning_effort` already existed at the call site).


diegosouzapw:
Thanks for this — clean, well-scoped fix with strong regression coverage. I ran your new test
file at the PR head and all 8 cases pass, including the precedence chain against model-suffix
effort and the connection default. No changes requested; this looks merge-ready.

Issue #7080, opened 2026-07-14
fix(backend): proxy "Test connection" false-negative for valid SOCKS5 proxies (invalid Socks5 initial handshake)


morpheus9393:
### Summary

The "Test connection" button in the **Proxies** tab and **Key Proxy modal** (Dashboard `/dashboard/providers/codex`) returns a false-negative for **valid SOCKS5 proxies**. The same proxy works perfectly when assigned to an account and used for live LLM traffic, but the test endpoint reports:

> **`Received invalid Socks5 initial handshake (no accepted authentication type)`** (~100ms)

This makes it impossible to validate a new SOCKS5 proxy through the UI before assigning it, even though the proxy is healthy.

### Steps to reproduce

1. Start OmniRoute v3.8.48 via docker-compose (`base` profile)
2. Ensure `ENABLE_SOCKS5_PROXY=true` in `.env`
3. Dashboard `http://127.0.0.1:18081` → Proxies → add a custom SOCKS5 proxy such as:
   ```
   socks5://geonode_<session-id>:<password>@proxy.geonode.io:12000
   ```
   (works with any Geonode residential session; e.g. `codex_22`, `codex_27`, etc.)
4. Click "Test connection" on the saved proxy
5. **Expected:** public IP returned + latency in 200–2000ms range
6. **Actual:** error `Received invalid Socks5 initial handshake (no accepted authentication type)` after ~100ms

### Observed vs. expected behaviour

- The same proxy, when **assigned to a codex account** (`proxy_assignments` scope=`account`), is used successfully for live `/v1/chat/completions` traffic (e.g. `🌊 [STREAM] CODEX | gpt-5.6-sol | 8000ms | complete`).
- The test endpoint appears to take a **different code path** than the live account-routing path. Both endpoints call `createProxyDispatcher(proxyUrl)` from `open-sse/utils/proxyDispatcher.ts`, but the test endpoint calls it from a route that:

  1. Reads `proxyId` and pulls credentials via `getProxyById(id, { includeSecrets: true })`
  2. Builds a `SocksDispatcherOptions` object (`type: 5, host, port, userId, password`) — see `proxyDispatcher.ts:503-511`
  3. Calls `undiciRequest("https://api64.ipify.org?format=json", { dispatcher, ... })` from `src/app/api/settings/proxy/test/route.ts:199-208`
  4. Wraps the underlying `socks`-library error verbatim — see `route.ts:229-238`

[...]

PR #13646, opened 2026-09-14
fix(api): use shared SOCKS5 flag reader in settings routes

maxmad64bis:
## Summary

Both settings routes duplicated the opt-out check for ENABLE_SOCKS5_PROXY instead of calling the exported reader, so a future edit to the flag semantics could leave the dashboard and the dispatcher disagreeing. A concrete case is ENABLE_SOCKS5_PROXY=" OFF ": the dispatcher trims and lowercases before comparing (open-sse/utils/proxyDispatcher.ts:262), while a copied formula could drift on whitespace or casing. After this change both routes call isSocks5ProxyEnabled() as the single source of truth (src/app/api/settings/proxies/route.ts:47, src/app/api/settings/proxy/route.ts:33).

Out of scope: the generic env helper, the build-time client flag, and the other opt-in checks keep their own semantics and are untouched.

## Related Issues

- #7080 (closed): different handshake topic, no overlap with this flag-reader change.
- #13578, #13581: opened as background for this change, not a collision.
- Census of related reports found zero overlapping work on these two call sites.

## Validation

- [x] Change type: routing
- [x] `npm run lint`
- [ ] Reconciled with the current active release base; focused checks rerun afterward — rebase and rerun owed before merge

## Tests Added Or Updated

- New tests/unit/settings-socks-flag-reader.test.ts: 3 tests pass — 1 matrix covering 7 flag values (false, 0, no, off, " OFF ", "False", unset) plus 2 static assertions proving both routes import the reader and carry no copied list (tests/unit/settings-socks-flag-reader.test.ts:10).
- Neighbors tests/unit/proxy-registry-route-handlers.test.ts and tests/unit/proxy-relay-test-error-5716.test.ts: 7 pass, 0 fail.
- RED proven on the base: each route still carried the copied list (grep count 1/1) so both static assertions failed there, while the matrix already passed because the dispatcher itself is unchanged.

## Coverage Notes

Both routes (src/app/api/settings/proxies/route.ts:47, src/app/api/settings/proxy/route.ts:33) are covered by the new test, and the 7-case matrix covers every branch of the reader including the unset-to-ON default, so no touched file moves down in coverage.

## Reviewer Notes

- Edits are line-scoped only; direct measurement shows the base files prettier-clean and the final diff stays minimal at 7 added / 14 removed lines across the two routes.
- The new import uses the extensionless path, matching 8 of the 11 existing settings imports of this module (8 without extension vs 3 with).
- The indirect caller through getSupportedProxyTypes() is intentionally untouched and follows the direct replacement (src/app/api/settings/proxy/route.ts:109).
- The reader reads process.env at call time with no cache, so the dashboard and the dispatcher agree on every request.
- Gates: api-typecheck exits 0 against the frozen baseline, eslint reports 0 errors on owned files, and prettier passes on all 4 touched files.
- Change type is `routing` because the flag governs SOCKS5 egress routing.


maxmad64bis:
## CI review (§12)

All failures inherited from the base or unstable — none cite this PR's files.

- `Merge integrity`: `cli-tunnel` skill drift (third-party voisin, not this PR's files).
- `Docs Gates`: 356-vs-358 provider count (N90, inherited).
- `Fast Quality Gates`: identical triplet on tierce #13471 same base (mutation 15, cycles SCC `models/providers/deletion`, `autoCombo.test.ts` TS2739) — inherited.
- `ESLint`: 4 hits all outside this PR (`resource-pressure-self-restart`, `volcengine-plan-binding-upsert` ×3); owned files lint exit 0, prettier 4/4.
- `Unit Tests`: zero failing lines cite this PR's files (settings routes + `settings-socks-flag-reader` absent from `✖` lines). ReDoS + reset-aware samples: unstable/inherited as documented on #13641.

PR #13647, opened 2026-09-14
fix(routing): skip redundant parseAutoPrefix for recognized built-in auto variants

### LINKED ISSUE #13644: fix(backend): parseAutoPrefix rejects valid best-* auto variants
## Description

`classifyAutoModel()` correctly recognizes `auto/best-coding` via `AUTO_TEMPLATE_VARIANTS` (builtinCatalog.ts), but `applyAutoPrefix()` re-parses the model name with `parseAutoPrefix()` from `@omniroute/open-sse/services/autoCombo/autoPrefix.ts` — which only validates bare variants (`coding`, `fast`, `cheap`, `smart`, `lkgp`, `chaos`, `offline`) in its `VALID_VARIANTS` list.

Since `"best-coding"` is not in `VALID_VARIANTS`, the parser returns `valid: false`, causing:

1. Spurious log warning: `Invalid auto prefix format: auto/best-coding`
2. `applyAutoPrefix` returns state unchanged (no variant extracted)
3. The combo still works via the LKGP fallback path, but the warning noise is confusing and the explicit variant override is silently lost

## Reproduce

Send any request with `auto/best-coding` as the model. OmniRoute logs:

```
Invalid auto prefix format: auto/best-coding
```

## Fix

Two options:

1. **Fix `parseAutoPrefix`**: Accept `best-*` prefixed variants (strip the `best-` prefix and validate the remainder)
2. **Skip re-validation in `applyAutoPrefix`**: When `classifyAutoModel` already confirmed the model as a recognized built-in (via `AUTO_TEMPLATE_VARIANTS`), bypass `parseAutoPrefix` validation and extract the variant directly

Option 2 is cleaner since `AUTO_TEMPLATE_VARIANTS` is the source of truth for valid auto model strings.

## Affected versions

Tested on 3.8.51. Present in `src/sse/handlers/autoRouting.ts:applyAutoPrefix` since the auto-prefix refactor.

hummern:
## Problem

`classifyAutoModel()` correctly recognizes `auto/best-coding` via `AUTO_TEMPLATE_VARIANTS` (builtinCatalog.ts). But `applyAutoPrefix()` re-validates every model name through `parseAutoPrefix()` from `@omniroute/open-sse/services/autoCombo/autoPrefix.ts` — whose `VALID_VARIANTS` only contains bare variants (`coding`, `fast`, `cheap`, `smart`, `lkgp`, `chaos`, `offline`), **not** `best-*` prefixed aliases.

Since `"best-coding"` is not in `VALID_VARIANTS`, `parseAutoPrefix` returns `valid: false` and the function returns early with no variant extracted, despite `classifyAutoModel` having already confirmed the model as valid.

### Symptoms

1. Spurious log warning: `Invalid auto prefix format: auto/best-coding`
2. Explicit variant override silently lost (falls back to LKGP instead of `coding`)
3. Warning noise every chat completion when using `auto/best-coding`

## Fix

Compute `recognizedBuiltIn = AUTO_TEMPLATE_VARIANTS.hasOwnProperty(model)` once and skip `parseAutoPrefix` validation when true — `AUTO_TEMPLATE_VARIANTS` is the source of truth for valid auto model strings.

## Affected versions

Tested on 3.8.51. Bug present since the auto-prefix refactor in `src/sse/handlers/autoRouting.ts`.

Fixes #13644


diegosouzapw:
Thanks for tracking this down — confirmed the root cause (`VALID_VARIANTS` in `autoPrefix.ts` doesn't include the `best-*` aliases that `AUTO_TEMPLATE_VARIANTS` recognizes). One nuance from re-testing: `resolveAutoRoutingState()` already classifies `auto/best-coding` correctly via `classifyAutoModel()` *before* `applyAutoPrefix()` runs, and the early-return in the old code preserves that already-correct `state` — so the routing `variant` itself was not actually broken (verified: `tests/unit/auto-best-free-tier-filter.test.ts` passes identically with and without your fix). The real, verified effect of your fix is eliminating the spurious `Invalid auto prefix format` warning log on every `auto/best-*` request — still a legitimate, worthwhile fix, just narrower than the issue describes. Could you (a) retarget the branch to `release/v3.8.51` (this PR is based on `main`, which has drifted from the active release branch) and (b) add a small regression test asserting the warning no longer fires for a recognized built-in variant? Happy to merge once those land.

hummern:
Addressed both review requests:

(a) **Retargeted to release/v3.8.51** — the branch was rebuilt as a single clean commit (548b9a60a) on top of the current release/v3.8.51 tip, dropping the stale `main`-drifted commits that were previously under the fix. PR base confirmed as `release/v3.8.51`, head is the cleaned branch.

(b) **Added regression test** — `tests/unit/auto-best-variant-no-warn.test.ts` asserts the spurious 'Invalid auto prefix format' warning no longer fires for a recognized built-in variant (`auto/best-coding`), while a genuinely unknown `auto/not-a-real-variant` still warns (proving the log probe actually detects the message). Verified: the test fails on the pre-fix code (warning fires) and passes with the fix. Related existing suites still pass (auto-best-free-tier-filter, autoPrefix).

diegosouzapw:
Thanks @hummern — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #13649, opened 2026-09-14
fix(memory): add TTL cleanup to chatBodyAdmission queues to prevent memory leak

khnker:
## Summary

Adds a TTL-based cleanup mechanism to chatBodyAdmission.ts to prevent unbounded memory growth in the #queues and #queueTimestamps Maps.

## Problem

The #queues Map grows indefinitely as new session keys are added but never removed. This causes memory usage to increase over time, eventually leading to process restarts (observed at ~2.5GB threshold).

## Solution

1. Added #queueTimestamps Map to track when each session was last active
2. Added #cleanup() method that runs every 60 seconds and removes entries older than 5 minutes
3. Updated #queues.set() to also record the timestamp
4. Cleanup removes entries from #queues, #queueTimestamps, and #fairKeys

## Testing

- Verified the fix works by observing memory usage in production
- No regression in functionality


diegosouzapw:
Thanks for looking into the memory growth — but tracing `#queues`/`#fairKeys` in
`ChatAdmissionController`, every waiter already gets removed via `#removeWaiter` →
`#removeFairKey` the instant its own queue drains (unconditionally, on admit/timeout/abort), so
those two structures shouldn't grow unbounded on the current tip. Do you have a heap snapshot or
a failing test that pins down which structure is actually accumulating in your ~2.5GB
production case? That would help target the fix precisely.

Separately, as written the new `#cleanup()` has a correctness risk: it deletes a session's queue
by creation-age without checking whether it's still non-empty, which can strand an actively
parked waiter outside the fair-dispatch rotation (it would then only resolve via its own
queue-timeout, even if a slot frees up in the meantime) — worth guarding that before this lands.

PR #13655, opened 2026-09-14
fix(providers): declare Agnes official thinking effort tiers

HouMinXi:
## Problem

Agnes chat models advertise `supportsReasoning: true` but never declare
`supportedThinkingEfforts`. Two consumers then invent a six-tier
vocabulary the live API does not document:

- `/v1/models` falls back to none/low/medium/high/xhigh/max
- playground/combo builder reuse that list as effort options

Hermes defaults `reasoning_effort` to xhigh. That is a 400 on
`agnes-2.0-flash` and `agnes-2.5-flash`. `off` is also 400 on every
Agnes chat model; the sanitizer currently has no declared list to map
it onto `none`.

## Live contract (2026-09-14, apihub.agnes-ai.com)

| effort   | 2.0/2.5 | 3.0 |
|----------|---------|-----|
| none     | 200     | 200 |
| minimal  | 400     | 200 |
| low      | 200     | 200 |
| medium   | 200     | 200 |
| high     | 200     | 200 |
| xhigh    | 400     | 200 |
| max      | 200     | 200 |
| ultra    | 400     | 400 |
| off      | 400     | 400 |

HuggingFace's Agnes-3.0-Flash card lists none/low/medium/high. The
registry follows the live API, which also takes minimal and xhigh on 3.0.

## Fix

Declare per-model `supportedThinkingEfforts` in the Agnes registry:

- 2.0/2.5: none, low, medium, high, max
- 3.0: none, minimal, low, medium, high, xhigh, max

`sanitizeReasoningEffortForProvider` already clamps undeclared values
to the nearest declared tier (xhigh -> max on 2.x, off -> none, ultra
-> max). Catalog and playground pick that list up automatically.

## Tests

`tests/unit/agnes-provider.test.ts` now pins the live vocabularies as
literal arrays (so a registry constant edit cannot silently retune the
expected set) and covers sanitizer mapping for both families.

Touched-area run: 52/52 (agnes-provider + sanitize-reasoning-effort +
learned-reasoning-effort-caps).


diegosouzapw:
Clean, minimal fix — declaring the real per-model `supportedThinkingEfforts` and letting
the existing sanitizer clamp undeclared values is the right mechanism instead of adding a
new special case. Ran `tests/unit/agnes-provider.test.ts`: 20/20 pass, including the
literal-array pin you added for the 2.0/2.5 vs 3.0 vocabularies. Ready to merge.

PR #13663, opened 2026-09-14
[needs-vps] fix: respect iOS safe-area insets in standalone PWA mode

isavage:
## Summary

Fixes the dashboard UI sliding under the iOS status bar / clock when opened as a home-screen app (standalone PWA).

## Root cause

The app opts into full-bleed via `viewportFit: "cover"` + `statusBarStyle: "black-translucent"` + `display: "standalone"`, but has no iOS safe-area compensation. The only top-inset handling is Electron-only (`--desktop-safe-top`). On an iPhone home-screen app the sticky `Header` and mobile sidebar drawer draw from `y=0`, under the clock.

## Changes

1. `src/app/globals.css` — add `--ios-safe-top` / `--ios-safe-bottom` vars scoped to `@media (display-mode: standalone)` with `@supports` fallback
2. `src/shared/components/Header.tsx` — extend Electron safe-area padding to also apply iOS safe-area in non-Electron contexts
3. `src/shared/components/layouts/DashboardLayout.tsx` — offset mobile sidebar drawer with explicit `top`/`bottom` using the safe-area vars instead of `inset-y-0`

## Scope

- Only affects standalone PWA mode
- Normal browser tabs unaffected
- No backend/API changes
- Mirrors existing Electron safe-area pattern


diegosouzapw:
Thanks for tracking down the standalone-PWA safe-area gap — the fix mirrors the existing
Electron `--desktop-safe-top` pattern nicely, gated correctly behind `display-mode: standalone`
+ `@supports`. Two small asks before merge: (1) add a changelog fragment under
`changelog.d/fixes/` per repo convention, and (2) since this can only really be verified on a
real iOS device in standalone mode, could you confirm you tested it there (notch/Dynamic Island
device, light + dark)? Happy to merge once those are in.

isavage:
Thanks for the review.

1. **Changelog fragment** added: `changelog.d/fixes/13663-ios-standalone-safe-area.md` (single `- fix(dashboard):` bullet).
2. **iOS testing** — verified live on an **iPhone 15 Pro (Dynamic Island)** in standalone home-screen mode on **iOS 26 and iOS 27**, both light and dark: header clears the status bar and the mobile drawer respects the bottom inset. The Dynamic-Island clock stays fully visible. No regression in normal browser tabs (scope is gated to `@media (display-mode: standalone)`).

Fragment is on the branch — ready for re-review.

Issue #5570, opened 2026-06-30
[BUG] AI/ML API model fetch fails — 'API unavailable — using local catalog' (key validates OK)


chirag127:
### Summary

When adding an AI/ML API connection and importing models, the live model fetch fails:

```
Importing Models
Failed to fetch models
API unavailable — using local catalog
```

OmniRoute falls back to a local (bundled) model catalog, but this means newly added or removed models from AI/ML API's live endpoint won't be reflected.

### Steps to reproduce

1. Add AI/ML API connection (API key validation succeeds — shows "Valid")
2. Proceed to model import step

### Observed behaviour

```
Failed to fetch models
API unavailable — using local catalog
```

Models load from the local catalog fallback, not from the live API.

### Expected behaviour

The live model fetch from AI/ML API's `/v1/models` (or equivalent) should succeed. If the endpoint is temporarily down, the fallback is acceptable — but if this is a persistent failure it indicates a broken base URL, wrong endpoint path, or auth issue in the provider config.

### Questions for maintainers

- Is `api.aimlapi.com/v1/models` the correct models endpoint?
- Does AI/ML API require a specific header or query param for model listing that isn't currently sent?
- Is the local catalog up to date with AI/ML API's current model list?

### Environment

- OmniRoute version: omniroute@3.8.41
- Platform: Windows 11
- Verified: 2026-06-30
- Note: API key validation ("Check") returns **Valid** — auth is not the issue

diegosouzapw:
Fixed in `release/v3.8.43` (merged via #5673). AI/ML API's live catalog (400+ models) is served at the public, auth-free `https://api.aimlapi.com/models` endpoint (not `/v1/models`) — that's now wired into the model-import flow, so you'll get the live list instead of the bundled fallback. The bundled catalog stays as the offline fallback. Will ship in v3.8.43. Thanks! 🙏

PR #13668, opened 2026-09-14
50/50 RevShare Integration: OmniRoute & AIML API

### LINKED ISSUE #13667: [BUG] AI/ML API: discovery returns 0 models (stale seed served) and every OpenAI-SDK request 400s on literal-null optionals
### OmniRoute Version

3.8.51

### Installation Method

Built from source

### Operating System

Linux

### OS Version

_No response_

### Node.js Version

_No response_

### Provider(s) Involved

aimlapi (AI/ML API)

### Model(s) Involved

claude-sonnet-4.6, deepseek-chat, gpt-5 (and the 6 seeded ids)

### Client Tool

any OpenAI SDK client (openai-python / openai-node), cURL

### Description

Two independent defects make the `aimlapi` provider unusable on `release/v3.8.51`. Either one alone is enough to break it, and fixing only one still leaves the provider broken.

**1 — Model discovery returns nothing, so the stale static seed is served.**

`PROVIDER_MODELS_CONFIG.aimlapi.parseResponse` tests `Array.isArray(data)`, but the catalog endpoint now answers with the OpenAI-style envelope `{ "object": "list", "data": [...] }`, so every row is discarded before any filter runs and the route falls back to its local catalog — the same symptom as #5570. The chat filter also looks for `type === "chat-completion"`; the catalog's current vocabulary is `openai/chat-completions`, and the old spelling matches 0 rows. On top of that, four of the six seeded ids are absent from the catalog as both id AND alias — `claude-3-5-sonnet-20241022`, `gemini-1.5-pro`, `meta-llama/Meta-Llama-3.1-70B-Instruct-Turbo`, `mistral-large-latest` — so most of what a user is offered 404s on first use.

This is not a regression in OmniRoute's code: #5570 was correct when it landed, and the catalog changed shape underneath it.

**2 — Every OpenAI-SDK request 400s on literal-null optionals.**

AI/ML API validates optional fields with a strict schema and answers `400 "Expected number, received null"` when a field arrives as a literal `null`, instead of reading null as "unset". The OpenAI SDKs serialise an unset optional as `null`, and OmniRoute relays the caller's body verbatim — so a client that never touches `temperature` still puts `temperature: null` on the wire and gets a 400 on every single message.

Field-by-field sweep against `POST /v1/chat/completions`, 2026-09-03, read off `details[].path` in the 400 bodies (the top-level `message` is generic and names no field, which is most of why this is hard to diagnose from a log):

| result | fields |
| --- | --- |
| 400 on null, every model | `seed`, `tools`, `tool_choice`, `response_format`, `stream`, `stream_options`, `parallel_tool_calls`, `max_tokens`, `max_completion_tokens`, `reasoning_effort` |
| 400 on null, model-dependent | `temperature`, `top_p` — 400 on `claude-sonnet-4.6` and `deepseek-chat`, 200 on `gpt-5`, so an integration smoke-tested only against `gpt-5` looks healthy |
| 200 on null | `stop`, `presence_penalty`, `frequency_penalty`, `n`, `user`, `logprobs`, `logit_bias`, `top_logprobs`, `metadata` |

### Steps to Reproduce

1. Confirm the catalog shape (no key needed):

    curl -s https://api.aimlapi.com/models | jq '{envelope: (.object // "bare array"), rows: (.data | length), chat: (.data | map(select(.type == "openai/chat-completions")) | length)}'

   Returns `{ "envelope": "list", "rows": 936, "chat": 353 }` (2026-09-03). The parser expects a bare array and the filter expects `chat-completion`, so it keeps nothing.

2. Add an AI/ML API key on `/dashboard/providers` and open the model picker for the provider. The live list is empty and the 6-entry static seed is shown instead.

3. Pick `claude-3-5-sonnet-20241022` from that seed and send any message. The id does not exist upstream.

4. Send a request the way an OpenAI SDK serialises one with unset optionals:

    curl -s http://localhost:20128/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"aimlapi/claude-sonnet-4.6","temperature":null,"messages":[{"role":"user","content":"hi"}]}'

### Expected Behavior

1. Discovery lists the provider's live chat catalog (353 models on 2026-09-03).
2. A request from an OpenAI SDK client that leaves optionals unset is relayed and answered normally — `null` means "unset", so the field should not reach an upstream that rejects it.

### Actual Behavior

1. Discovery returns 0 models and the route falls back to the 6-entry static seed, four of whose ids no longer exist upstream, so most of what is offered 404s on first use.
2. Every request carrying a literal `null` on one of the fields listed above is rejected with 400 before the model is ever reached.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
{
  "message": "Validation error",
  "details": [
    { "path": ["temperature"], "reason": "Expected number, received null" }
  ]
}
```

### Screenshots

_No response_

### Additional Context

- The Operating System dropdown is required by this form but is not meaningful for this report: both defects are platform-independent — one is a response-parsing path, the other a request-body path. Our verification was done against the `release/v3.8.51` source tree and by calling the AI/ML API directly.
- We have a fix ready and are opening a PR against `release/v3.8.51` alongside this report.

### Validation Plan

node --import tsx/esm --test tests/unit/aimlapi-catalog-repair.test.ts tests/unit/aimlapi-attribution-headers.test.ts tests/unit/aimlapi-null-param-strip.test.ts tests/unit/provider-models-discovery-split.test.ts tests/unit/provider-models-route.test.ts tests/unit/executors-strip-unsupported-params.test.ts
npm run lint

hugoaimlapi:
Hi! I'm Hugo from aimlapi.com — an AI aggregator that gives access to 1000+ models in one API, trusted by 400k+ users.

We'd love to be available as a working provider option inside OmniRoute — so we went ahead and did all the technical work on our side.

To build our partnership, we offer a 50/50 revenue share on all traffic from this integration. (P.S.: tracking starts as soon as this release goes live, so no earnings will be lost during setup)

My contacts: <hugo@aimlapi.com> (email / Slack), Telegram: @hug0the

---

## Why

`aimlapi` is already wired into OmniRoute, but it has been effectively dead since the AI/ML API catalog changed shape: discovery returns zero models, so every user falls back to the 6-entry static seed — and four of those six ids no longer exist upstream, so most of what is offered 404s on first use. Independently of that, any OpenAI-SDK client gets a 400 on *every* message, because the SDKs serialise an unset optional as a literal `null` and the gateway's schema rejects `null` instead of reading it as "unset". And unlike `openrouter`, the traffic reaches us untagged, so none of it is attributable to OmniRoute.

Full reproduction in #13667.

## Summary

Four changes, each independently revertible, no new mechanism anywhere:

- **Discovery repair** — `src/app/api/providers/[id]/models/discovery/providerModelsConfig.ts`. The parser tested `Array.isArray(data)` against what is now the OpenAI-style envelope `{ "object": "list", "data": [...] }`, so every row was discarded before any filter ran; and the chat filter looked for `type === "chat-completion"`, a spelling the catalog no longer publishes (`openai/chat-completions` matches 353 of 936 rows, verified 2026-09-03 — the old one matches 0). Both spellings are accepted now, and the bare array is still unwrapped, so a future rename degrades to "some models missing" rather than "no models at all". The `chat.length ? chat : all` fallback is dropped — it would have published 583 video/image/TTS/batch ids into a chat model picker, which is what kept the broken type filter invisible.
- **Static seed** — `open-sse/config/providers/registry/aimlapi/index.ts`. The four dead ids (`claude-3-5-sonnet-20241022`, `gemini-1.5-pro`, `meta-llama/Meta-Llama-3.1-70B-Instruct-Turbo`, `mistral-large-latest` — absent from the catalog as both id AND alias) are replaced by six verified twice on 2026-09-03: present on the catalog's `openai/chat-completions` surface AND answering 200 to a real `POST /v1/chat/completions`. Catalog membership alone is not enough — `llama-3.3-70b-versatile` is listed as a chat model and still 404s on inference. All six are the unprefixed spelling on purpose: a `vendor/model` id in this registry is matched by `parseModel()` as an exact model id and reports `provider = null`, which would stop that string resolving to its real provider app-wide (#8716).
- **Partner attribution** — same file. A `headers` block, exactly the mechanism `openrouter` / `orcarouter` / `cline` / `gitlawb` already declare: `BaseExecutor.buildHeadersPreamble()` spreads it into a freshly built map per request, so the headers are scoped to this provider's own dispatch and the shared registry constant is never mutated. The regenerated `translate-path` golden shows the four headers under `aimlapi` and under no other provider. `HTTP-Referer` / `X-Title` follow the OpenRouter convention and name the *calling* application, so they point at OmniRoute's own repository, not at the gateway being called.
- **Display label** — `src/shared/constants/providers/apikey/gateways.ts` plus the generated `docs/reference/PROVIDER_REFERENCE.md`: `AI/ML API` becomes `aimlapi.com`, the name the provider ships under. The machine id (`aimlapi`) and alias (`aiml`) are untouched — those are what existing user configs and stored connections key on.
- **Null-param strip** — `open-sse/translator/paramSupport.ts`. A `dropIfNull` rule in the existing `STRIP_RULES` table, whose stated purpose is already "params a given provider/model rejects upstream". Field-by-field sweep against `POST /v1/chat/completions` on 2026-09-03, read off `details[].path` in the 400 bodies (the top-level `message` is generic and names no field):

  | result | fields |
  | --- | --- |
  | 400 on null, every model | `seed`, `tools`, `tool_choice`, `response_format`, `stream`, `stream_options`, `parallel_tool_calls`, `max_tokens`, `max_completion_tokens`, `reasoning_effort` |
  | 400 on null, model-dependent | `temperature`, `top_p` — 400 on `claude-sonnet-4.6` and `deepseek-chat`, 200 on `gpt-5`, so an integration smoke-tested only against `gpt-5` looks healthy |
  | 200 on null (left alone) | `stop`, `presence_penalty`, `frequency_penalty`, `n`, `user`, `logprobs`, `logit_bias`, `top_logprobs`, `metadata` |

  Only the literal `null` is dropped — an explicit `temperature: 0` or `stream: false` is forwarded untouched — and the rule is scoped to `provider: "aimlapi"`.

## Related Issues

- Closes #13667
- Related to #5570 — the catalog schema changed under that fix; this restores it and adds a regression guard so the next vocabulary change cannot silently empty the provider again.

## Validation

- [x] Change type: **provider**
- [ ] `npm run lint`
- [ ] Reconciled with the current active release base

Being straight about the unticked boxes rather than ticking them blind:

- The focused loop below has not been run by us on this branch — CI on this PR is the first run. We will fix whatever it turns up, same day.
- The branch is cut from `release/v3.8.51` at `ba597b631` and is behind the current tip (`963d3c46e`). We checked the four changes cherry-pick onto that tip with no conflicts; say the word and we will rebase or merge the base in.

Focused loop:

```bash
  tests/unit/aimlapi-catalog-repair.test.ts \
  tests/unit/aimlapi-attribution-headers.test.ts \
  tests/unit/aimlapi-null-param-strip.test.ts \
  tests/unit/provider-models-discovery-split.test.ts \
  tests/unit/provider-models-route.test.ts \
  tests/unit/executors-strip-unsupported-params.test.ts
npm run lint
```

Manual check against the live catalog (no key needed for discovery):

```bash
curl -s https://api.aimlapi.com/models \
  | jq '{envelope: (.object // "bare array"), rows: (.data | length),
         chat: (.data | map(select(.type == "openai/chat-completions")) | length)}'
# gives { "envelope": "list", "rows": 936, "chat": 353 }   (2026-09-03)

# end to end, with an AI/ML API key configured on /dashboard/providers:
curl -s http://localhost:20128/v1/chat/completions \
  -d '{"model":"aimlapi/claude-sonnet-4.6","temperature":null,
       "messages":[{"role":"user","content":"hi"}]}'
# before: 400 "Expected number, received null" — after: a normal completion
```

## Tests Added Or Updated

- `tests/unit/aimlapi-catalog-repair.test.ts` — new (seed ids, known-dead list, no vendor prefix, endpoint/format pinning)
- `tests/unit/aimlapi-attribution-headers.test.ts` — new (header values, partner-id shape, scoped to `aimlapi` only, fresh map per request, label is `aimlapi.com`)
- `tests/unit/aimlapi-null-param-strip.test.ts` — new (rejected-on-null dropped, real values preserved, accepted-on-null preserved, lists disjoint, no leak to other providers)
- `tests/unit/provider-models-discovery-split.test.ts` — updated (envelope unwrap, both type spellings, bare array still accepted, no fallback to non-chat rows)
- `tests/unit/provider-models-route.test.ts` — updated (the mock now returns what the live endpoint actually returns; the old mock is why a green suite coexisted with a provider that discovered nothing)
- `tests/unit/executors-strip-unsupported-params.test.ts` — updated (`STRIP_RULES` shape assertion accepts a `dropIfNull`-only rule)
- `tests/snapshots/provider/translate-path.json` — regenerated golden

## Coverage Notes

Every touched production file is covered by a test in this PR:

- `providerModelsConfig.ts` -> `provider-models-discovery-split.test.ts` plus `provider-models-route.test.ts`
- `registry/aimlapi/index.ts` -> `aimlapi-catalog-repair.test.ts` plus `aimlapi-attribution-headers.test.ts`
- `paramSupport.ts` -> `aimlapi-null-param-strip.test.ts` plus `executors-strip-unsupported-params.test.ts`
- `gateways.ts` -> `aimlapi-attribution-headers.test.ts` (label assertion)

`docs/reference/PROVIDER_REFERENCE.md` is generated; its one-cell change follows the `gateways.ts` label.

## Reviewer Notes

- ⚠️ base-red inherited: #12732
- Nothing in routing, the agent loop or the request pipeline changes. Previously saved connections load unchanged: only the display string moves, the id and alias stay.
- The partner id is asserted against `^part_[A-Za-z0-9]{1,64}$` in a test because a malformed one has no runtime symptom — the gateway answers 200 either way and simply records the usage as untagged, so a typo would cost attribution silently and forever.
- No supporter/sponsor chip and no dashboard placement change is included here. A placement request without a partnership behind it is the wrong thing to put in front of a maintainer, and it has no business travelling with a functional repair.


diegosouzapw:
Thanks for the detailed writeup and the tests — the discovery-envelope fix, the catalog
vocabulary update, and the null-optional stripping for aimlapi's strict schema all check out
against the current tip, and your test suite passes in isolation (24/24 on the file the
parallel probe run flagged, which turned out to be a shared-DB test-runner artifact, not your
code). I want to separate the technical fix from the revenue-share proposal in your PR
description — that's a business decision for the project owner, not something a PR can commit
to on its own. Before merging we'd want to confirm the `X-AIMLAPI-Partner-ID` header and the
`aimlapi.com` display-name rename are both something we're intentionally adopting, and drop the
two unrelated formatting hunks in `gateways.ts`. Will follow up once that's settled.

hugoaimlapi:
okay mate, thanks for the reply
if you want to connect with me to discuss all details with the core team, feel free to invite me in Slack hugo@aimlapi.com

hugoaimlapi:
hello @diegosouzapw 

just kindly ping about our request. if you want we can create a chat with my team to communicate faster

Issue #9418, opened 2026-08-04
[Feature] Disable auto/* and no-think/* Combos


CoZooo:
### Problem / Use Case

When requesting /v1/models, there are many models that contain auto/ and no-think/.

### Proposed Solution

Add an option only to keep original models (<ProviderName>/<ModelName>) and combos that added by users themselves.

### Acceptance Criteria

When the option is on, only original models (<ProviderName>/<ModelName>) and combos that added by users themselves are shown when requesting /v1/models. No auto/* and no-think/*.

Turning off auto/* and no-think/* shown is also accepted.

### Area

Dashboard / UI, Proxy / Routing

diegosouzapw:
Thanks for the suggestion — this is a good fit and we're adding it to the
backlog. The plan is to add two opt-in settings toggles (default off, so nothing
changes for existing operators) modeled on the existing "Hide paid models"
filter: `hideAutoCombos` to drop the built-in `auto/*` virtual combos from
`/v1/models`, and `hideNoThinkVariants` to drop the
`no-think/<provider>/<model>` gateway variants. User-defined combos and original
`<Provider>/<Model>` entries stay listed either way; routing is unaffected (a
client can still send `auto/coding` or `no-think/...` explicitly and it routes
the same). Keeping the issue open while we implement.

texastoland:
#9535 only implemented/tested the db setting. there's no ui/user control. I also tried inserting directly into the db but thinking variants persist. reopen?

PR #13670, opened 2026-09-14
feat(api): add per-key allowAutoCombos to gate the built-in auto/* combos

fouadSalkini:
## Problem: `auto/*` bypasses per-key authorization

`auto/*` combos are virtual — synthesised in the catalog, never stored as rows in `combos`. That makes them invisible to the combo access gate:

`src/shared/utils/apiKeyPolicy.ts` — `isComboAllowedForKey()`:

```ts
const comboName = await resolveRequestedComboName(modelStr);
if (!comboName) return { allowed: true, comboName: null };   // fail-open
```

`resolveRequestedComboName("auto/best-fast")` tries `getComboByName()` and `resolveComboForModel()`; both miss, so it returns `null` and the key's `allowedCombos` is never consulted.

`validateModelAccess()` then closes the other door. Lines 530-540 set `requestedComboName = modelStr` for any `auto/` or `qtSd/` id, and line 541 returns on it:

```ts
if (requestedComboName || !hasModelRestrictions) return null;
if (await isModelAllowedForKey(apiKey, modelStr)) return null;
```

So `allowedModels` **and** `blockedModels` are skipped for `auto/*` as well.

**Net effect: `allowedCombos` does not constrain `auto/*`.** A key scoped to one cheap lane can send `auto/best-coding` and reach every model on the gateway. `blockedModels: ["auto/*"]` only unadvertises the ids from `/v1/models` — it cannot deny them, for the early-return reason above.

Reproduced on a live gateway: a key whose `allowedCombos` held 24 named combos, with no `auto` entry and no `combo/*` wildcard, dispatched `auto/best-fast` successfully (HTTP 200).

## Why a flag and not a fix to the fail-open

Tightening `isComboAllowedForKey()` so `allowedCombos` governs `auto/*` is the smaller change, but it is **breaking**: every key whose list lacks an `auto` entry would start returning 403 the moment it deploys. On the gateway where this was found that includes a key with 1,378 calls on `auto/best-fast`.

So this PR adds an explicit opt-out instead. `allow_auto_combos` is `INTEGER NOT NULL DEFAULT 1`, and the row parser treats anything but an explicit falsy value as allowed — so existing rows, and rows predating the column, keep their current access. Opting out is deliberate.

Closing the fail-open is still worth doing; it just wants its own change, after operators have audited their `allowedCombos` lists.

## What the seven commits do

Working through the feature on a live gateway surfaced four follow-on gaps in the same surface — what a key may *dispatch* versus what `/v1/models` *advertises* to it. They are small, they share the catalog code path, and separating them would have left the first commit unusable in practice.

**1. `feat(api)` — per-key `allowAutoCombos`.** With the flag off: `validateModelAccess()` rejects `auto/*` through the existing `policyErrorResponse` so the Anthropic Messages shape is preserved, and the catalog skips the `auto/*` synthesis loop for that key, reusing the `hideAuto` break (#9418) — the key is never offered ids it cannot use, and the materialisation cost is skipped. The decision is a pure exported predicate, `isAutoComboDeniedForKey()`, so the auto-vs-not rule lives in one place and is directly unit-testable. Settable via `PATCH /api/keys/[id]`. No numbered migration: the column goes through `API_KEY_COLUMN_FALLBACKS`, matching `compression_enabled`, the existing default-on boolean.

**2. `feat(dashboard)` — the Auto Combos toggle.** The previous revision of this description deferred this as a follow-up; it is in the PR. `ApiKeyAutoCombosToggle.tsx` in the API Manager permissions modal, with `en` and `vi` message-catalog entries so `i18n:check-ui-coverage` stays green.

**3. `feat(api)` — combo `description` in `/v1/models`.** A combo's stored description was write-only: the dashboard held it, the catalog dropped it. Emitting it lets an OpenAI-compatible client show what a combo routes to without the operator maintaining a second list client-side.

**4. `fix(api)` — list a key's allowed combos.** A key with `modelAccessMode: "restrict"` and a populated `allowedCombos` got an **empty** `/v1/models`: the model-restriction filter ran over provider models and dropped every row, and combos were never re-admitted even though the key was explicitly authorized for them. Restricting *models* should not unadvertise the *combos* the same key may use.

**5. `fix(api)` — accept `displayName` on the combo schemas.** Caught before shipping. The field was plumbed through the route and the store but not declared on `createComboSchema` / `updateComboSchema`, so Zod stripped it: `PUT` answered 200 and nothing persisted. Also counted in the schema's "no valid fields to update" guard, or a body carrying only `displayName` is rejected as empty.

**6. `feat(api)` — per-key `catalogScope`.** `"all" | "combos" | "models"`. Listing scope is a separate question from dispatch authorization: an operator may want a key that *may* use provider models but is *offered* only combos. `catalogScope` answers only the listing question and grants nothing.

**7. `fix(api)` — invalidate the catalog on key visibility changes.** The regression this PR's own new fields introduced, and the reason the two before it looked broken in production.

## The cache-invalidation fix (commit 7)

`updateApiKeyPermissions()` already advances the unified `/v1/models` catalog generation for the fields that change what a key may dispatch:

```ts
const shouldInvalidateModelCatalog =
  normalized.modelAccessMode !== undefined ||
  normalized.allowedModels !== undefined ||
  // … allowedCombos, allowedConnections, allowedQuotas, disableNonPublicModels
```

`allowAutoCombos` and `catalogScope` — both introduced by this PR, both changing what the catalog *advertises* — were missing from it. The catalog response cache (#6408) keys on request shape (prefix / isCodex / key fingerprint / `configuredOnly` / `hideAuto` / `hideNoThink` / page), never on the key's DB state, so without the generation bump a PATCH toggling either field left the previous listing served until the TTL expired.

**Observed on a live deployment before the fix:** `PATCH catalog_scope="combos"` returned 200, the column read back `"combos"`, and `GET /v1/models` kept returning the same 44 mixed rows (24 combos + 20 provider models). After a process restart the same key correctly returned 24 combo-only rows. The feature was right; the response was stale. The dashboard's API-key screen could show a catalog that disagreed with the key it had just written.

The fix adds exactly those two fields to the existing predicate — **no new cache machinery**, reusing `invalidateModelCatalogCache()`. It still runs only after a successful write, so a no-op or failed update does not invalidate, and unrelated metadata edits (`isActive`, rate limits) still leave the catalog cached.

## Deliberately out of scope

- **The create path.** `POST /api/keys` does not read `allowAutoCombos`; a key is created allowed and opted out with a PATCH. Keeps the surface small.
- **Closing the `auto/*` fail-open itself.** Breaking, as argued above — its own change.

## Files

| File | Change |
|---|---|
| `src/lib/db/apiKeyColumnFallbacks.ts` | `allow_auto_combos INTEGER NOT NULL DEFAULT 1`, `catalog_scope TEXT` |
| `src/lib/db/apiKeys/rowParsers.ts` | `parseAllowAutoCombos()` (default-on, mirrors `parseCompressionEnabled`) |
| `src/lib/db/apiKeys.ts` | types, SELECT list, both row-parse sites, no-op guard, update statement, defaults, metadata build, catalog-invalidation predicate |
| `src/lib/db/apiKeys/permissionsUpdate.ts` | pass-through for both new fields |
| `src/shared/validation/schemas/keys.ts` | declare on `updateKeyPermissionsSchema` + count in the "No valid fields to update" guard |
| `src/shared/validation/schemas/combo.ts` | accept `displayName` on create/update + count it in the guard |
| `src/app/api/keys/[id]/route.ts` | destructure + forward |
| `src/shared/utils/apiKeyPolicy.ts` | `isAutoComboDeniedForKey()` + enforcement, metadata type |
| `src/app/api/v1/models/catalog.ts` | hoist `earlyKeyMeta`, extend the `hideAuto` break, emit combo `description`/`display_name`, re-admit allowed combos under model restriction, apply `catalogScope` |
| `src/app/(dashboard)/dashboard/api-manager/**` | Auto Combos toggle + Catalog Scope select |
| `src/i18n/messages/{en,vi}.json` | strings for the two new controls |

## Tests

Five unit files, all red before their commit and green after:

| File | Covers |
|---|---|
| `api-key-allow-auto-combos.test.ts` | 6 rules — column default, parser, predicate scope (`auto/*` only, never `qtSd/`, never a combo merely *named* `auto-router`), PATCH schema, route forward, catalog break. **0/6 → 6/6** |
| `api-key-catalog-scope.test.ts` | the three scope values, and that scope never grants dispatch |
| `models-catalog-combo-access.test.ts` | a restricted key still sees its allowed combos |
| `models-catalog-combo-description.test.ts` | `description` / `display_name` emission |
| `model-catalog-policy-invalidation-8728.test.ts` | extended: `allowAutoCombos` and `catalogScope` each advance the generation; `isActive` and a no-op update still do not. **Red at `5 !== 6` before the fix, 6/6 after** |

Regression sweep over every unit test touching the changed modules (`getApiKeyMetadata`, `isModelAllowedForKey`, `updateApiKeyPermissions`, `apiKeyPolicy`, the models catalog, `API_KEY_COLUMN_FALLBACKS`) — **72/72**. The focused catalog-invalidation sweep (`model-catalog-policy-invalidation-8728`, `model-catalog-runtime-invalidation`, `model-catalog-source-invalidation-8728`, `db-synced-model-catalog-invalidation-8728`) — **20/20**. Vitest — **473 tests across 51 files**.

`npm run typecheck:core` clean. ESLint on changed files clean. Prettier clean. All pre-commit gates (lint-staged, docs-sync, any-budget, tracked-artifacts) pass.

`npm run test:unit` in full is red on this branch **and on the base commit** — ~60 failures across i18n parity, grok reset credits, the janitor script, webpack/sqljs build guards and others, none of them in a file this PR touches. Verified by re-running the closest failure (`models-catalog-route.test.ts` → "does not duplicate custom Jina specialty models", an unrelated `jina/` vs `jina-ai/` prefix mismatch) against the base revision of `src/lib/db/apiKeys.ts`: it fails identically without this PR's change.

⚠️ base-red inherited: #12732


fouadSalkini:
Added the dashboard control in `4e875fd4b` — the "Deliberately out of scope" note above is now only true of the create path.

**`ApiKeyAutoCombosToggle`** sits in the permissions modal beside the prompt-compression toggle and mirrors it exactly (dedicated component, `role="switch"` button, `settings` message namespace).

It defaults to ON: state reads `apiKey?.allowAutoCombos !== false`, deliberately `!== false` rather than `=== true`, so a key that predates the column renders as enabled and matches `NOT NULL DEFAULT 1`.

The field is threaded through all three positional lists — the save handler signature, the modal prop type and the `onSave` call — plus the PATCH payload, so no later argument shifts position.

### i18n

Strings are added to `en.json` and `vi.json`. Vietnamese is translated rather than left as a sync placeholder because `tests/unit/i18n-vi-completeness.test.ts` asserts key parity with English *and* bans `__MISSING__` markers in that locale.

The other 49 catalogs are deliberately **not** synced here. `npm run i18n:sync-ui` also replicates ~844 unrelated pre-existing gaps across all 50 files, which does not belong in this change. Those locales fall back to English at runtime, and `i18n:check-ui-coverage` still passes at ~99.3% against its 80% threshold. Happy to run the full sync in a separate housekeeping PR if you'd prefer.

### Tests

The suite is now 8 rules, all passing:

- **R7** the modal renders the toggle, defaults ON via `!== false`, declares the positional parameter, and includes the field in the PATCH body
- **R8** `autoCombosTitle` / `autoCombosDesc` exist, are non-empty and are not placeholders, in both `en` and `vi`

`npm run typecheck:core` clean, Prettier clean, `i18n:check-ui-coverage` and `i18n:check-value-drift` both PASS.

`tests/unit/i18n-vi-completeness.test.ts` fails on this branch, but inherited: the committed `vi.json` already carries 3 `__MISSING__` markers on the base ref, and this change adds 2 fully-translated lines and the same 2 keys to both `en` and `vi` — so it cannot be the cause of a parity failure. Same base-red as #12732.

diegosouzapw:
Good find — reproduced the reasoning directly in `apiKeyPolicy.ts`: `auto/*` really does bypass
`allowedCombos` today because it resolves to no stored combo, and `validateModelAccess()`
returns before the allow/deny lists are consulted. The opt-out-flag approach (rather than
tightening the fail-open directly) is the right call given how many keys currently rely on
implicit `auto/*` access. Ran your four new test files on this branch: 25/25 passing.
Two small things before merge: (1) please add a changelog fragment under
`changelog.d/features/`; (2) worth a quick follow-up note (issue or comment) that `qtSd/*` has
the same bypass and isn't covered here — not blocking, just flagging so it doesn't get lost.

fouadSalkini:
Added the changelog fragment in commit `2b3c9bca3` under `changelog.d/features/api-key-allow-auto-combos.md`.

Also noted regarding `qtSd/*` having the same bypass — tracking that for a follow-up PR.

PR #13675, opened 2026-09-14
fix(chat-admission): default QUEUE_MS 2s to 15s so the queue bridges large-context occupancy

### LINKED ISSUE #13648: fix(resilience): chat admission default 2s QUEUE_MS cannot bridge 15s+ heavyweight occupancy — a single large-context agent self-sheds
**Environment:** omniroute 3.8.50 (npm install, systemd user service), single user, loopback clients, 30 GB host — heap nowhere near the 0.75 shed ratio.

**Repro:** one coding-agent session at ~113k tokens / 465 msgs (~1 MB bodies). While its main turn is in flight, the agent's next request (or its compression-aux call from the same API key) sheds `503 chat_admission_busy` — "Chat admission capacity is temporarily unavailable. Retry shortly." The client's retry budget (3 × 2s) always lands inside the hold, so the loop dies hard. 69 shed events in one afternoon, all `reason:"queue_timeout", activeHeavy:1, waiting:0, queuedBytes:0`:

```json
{"module":"chat-admission","reason":"queue_timeout","activeHeavy":1,"waiting":0,"queuedBytes":0,"msg":"structural chat admission shed (chat_admission_busy)"}
```

**Why:** the #12135 bounded-wait shipped as `OMNIROUTE_CHAT_ADMISSION_QUEUE_MS=2000`, but a ~113k-token turn holds the single default heavyweight slot (`OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT=1`) for 15s+ (resilience `requestQueue.maxWaitMs=15000` plus slow free-model SSE). The queue can never bridge the occupancy, so "heavyweight chat requests queue instead of 503ing" doesn't hold on defaults for exactly the workload this gate targets (coding agents). Note admission precedes combo routing, so failover can't help — it presents to users as "the combo doesn't swap models" (my initial misdiagnosis).

**Local workaround (fully fixed it):**

```
OMNIROUTE_CHAT_MAX_HEAVY_IN_FLIGHT=4
OMNIROUTE_CHAT_ADMISSION_QUEUE_MS=20000
OMNIROUTE_CHAT_ADMISSION_MAX_QUEUED_BYTES=16777216
```

Verified afterwards: 3 concurrent ~200 KB requests all return 200, zero sheds since restart.

**Suggestions (any one suffices):**
- scale the default `QUEUE_MS` to at least `resilienceSettings.requestQueue.maxWaitMs` (15s);
- default-on `OMNIROUTE_CHAT_VIRTUAL_LANES` so one key's burst queues per-lane against the shared budget instead of shedding it;
- let the same fairness key (an agent + its own aux calls) share the healthy-headroom path so a single session can't self-shed.

Related: #12135, #11244, #12037

oyi77:
Fixes #13648 — a single large-context agent self-sheds `503 chat_admission_busy` (`reason: queue_timeout`) because the default `OMNIROUTE_CHAT_ADMISSION_QUEUE_MS=2000` can never bridge 15s+ heavyweight-slot occupancy.

## Change

- `src/shared/middleware/chatBodyAdmission.ts`: default `CHAT_ADMISSION_QUEUE_MAX_MS` 2s → 15s, matching the resilience `requestQueue.maxWaitMs` default (15s). Explicit env wins; `0` still restores immediate rejection.
- `.env.example` (both stanzas), `docs/reference/ENVIRONMENT.md`, `docs/guides/TROUBLESHOOTING.md`: default documented as `15000`.
- `changelog.d/fixes/13648-chat-admission-queue-default.md`: fragment.
- `tests/unit/chat-body-admission-queue.test.ts`: default pinned at 15s + new `#13648` regression test proving an aux call still parks past the old 2s shed point and acquires the freed slot.

## Verification

- `tests/unit/chat-body-admission-queue.test.ts`: 18/18 pass (17 existing + 1 new).
- Sibling suites (`chat-body-admission`, `heavy-admission-retry-after-12135`, `with-chat-admission-10786`, `chat-admission-visibility-11244`): 53/53 pass.
- `check-env-doc-sync`: `OMNIROUTE_CHAT_ADMISSION_QUEUE_MS` in sync; remaining 5 missing vars (`PRESSURE_SELF_RESTART`, `SQLJS_WASM`, `VACUUM_MIN`, `WAL_GUARD`, `WAL_PASSIVE`) are pre-existing tip-drift in untouched files.
- `check-file-size`: only red is `open-sse/utils/stream.ts`, untouched by this diff (pre-existing).
- `check-changelog-integrity`: fails identically on the clean tip (proven via stash) — pre-existing drift.

## Risk

Bounded latency, not correctness: a waiter may now park up to 15s instead of shedding at 2s when a slot is stuck. The queued-bytes valve (4 MB), client-abort early exit, and `0`/env overrides are unchanged.


diegosouzapw:
Thanks for the diagnosis and especially the end-to-end integration test (`#13648:` a 15s-scale
occupancy bridges the default queue) — that's a genuinely good regression proof. #13676 opened
the same fix in parallel, deriving the default from `DEFAULT_REQUEST_QUEUE_MAX_WAIT_MS` instead
of a new hardcoded `15_000` — which turns out to matter: #13553 (merged a few hours after both
PRs opened) just raised that default from 15s to 30s, so the hardcoded `15_000` here is already
stale on the current tip, and #13676's derived value already tracks it correctly (verified in a
probe). We'd like to merge #13676 for that reason and port your integration test into it with
credit to you — let us know if that works, or if you'd rather rebase this PR onto the derived
constant instead.

PR #13635, opened 2026-09-14
fix(ci): clear the release/v3.8.51 base-reds on the PR fast path

dpozimski:
## Summary

This clears every inherited red on `release/v3.8.51` that shows up on feature PRs today: Docs Gates, Merge integrity, No new ESLint warnings, five Fast Quality Gates, and the 51 failing unit tests in the 4 fast-path shards. I reproduced each failure on the untouched tip `152d95108` before fixing it. Every fix is cut from that tip, so it merges cleanly, unlike the older sweeps (see Reviewer Notes).

⚠️ base-red inherited: #12732

Most of the 51 unit failures were tests that went stale after intentional merges. Six were real regressions:

- **Error redaction ran a quadratic regex** (#12506). The glued-prefix alternative `[A-Za-z0-9]{3,}sk[-_]…` in `STRONG_CREDENTIAL_TOKEN` was unanchored, so on a long alphanumeric run it rescanned from every start position. A 50k run took 1,096 ms in `containsStrongCredentialToken`, and `sanitizeErrorMessage` spent about 40 ms on every long message (under 0.05 ms before #12506). On loaded CI runners that is what trips the `sanitizers.property` ReDoS guard: it failed twice on this PR at 260 to 264 ms against its 250 ms bound, and also today on `fix/latch-interchunks`. The alternative now carries the same `(?<![A-Za-z0-9])` lookbehind as its plain `sk-` sibling. A match of this alternative can only start at the beginning of an alphanumeric run, so the matches are the same: 300,000 fuzzed strings gave identical `replace` and `test` results for the old and new pattern. `sanitizeErrorMessage` drops to about 17 ms, and that remainder is the linear labeled-assignment scanner.
- **Provider pipeline dropped the upstream `error.code` / `error.type`** (#12867). `isAntigravityMissingProjectError` could never match, so a config-class Antigravity 422 fell into a generic account cooldown. `providerExecutionPipeline.ts` now reads message, code and type from the parsed body and passes them to `createErrorResult`.
- **The provider detail page shipped `node:fs` to the browser** (#13264). The codebuddy-cn registry entry imported the server-only `src/lib/oauth/constants/oauth.ts`, which pulls in `cursorAgentCliVersion.ts`. Every dashboard route then failed in dev with a `node:fs` module error. `CODEBUDDY_CN_USER_AGENT` moved to `open-sse/config/providerHeaderProfiles.ts`, where the other CLI user-agent pins live, and `oauth.ts` re-exports it.
[...]

PR #13361, opened 2026-09-11
fix(ci): clear two release-green HARD failures on v3.8.51

HouMinXi:
## Current

`release/v3.8.51` is still not green. #12732 tracks Fast Quality leftovers. This PR already landed the earlier HARD clears (ESLint `any`, provider-count 358, skills/cli-tunnel, `.env.example`, cycles/mutation/pack/typecheck/secrets). Two more commits sit on the same branch after the live extract moved the call sites those guards were scanning.

## Problem

Earlier HARD items (already in this PR):

1. `tests/unit/volcengine-plan-binding-upsert.test.ts` called `upsertConnection` through `(bindingTesting as any)`. Three `no-explicit-any` errors.
2. Live `providers.ts` unions **358** unique ids. `docs/reference/PROVIDER_REFERENCE.md` said **356**. `gen:provider-reference` also wrote `lastUpdated: DATE---` on one line; fumadocs then fails the Next build.
3. `check:agent-skills-sync` regenerates `skills/cli-tunnel` because `bin/cli/commands/tunnel.mjs` declares `.command("create")` plus `new Argument("[type]")`. The committed skill said `tunnel create`.
4. Six live env vars documented in `ENVIRONMENT.md` (walMaintenance / cleanup / resourcePressure / sqljs) were missing from `.env.example`.
5. Fast Quality `quality.yml` HARD gates: cycles (`models.ts` -> `providers.ts` -> `deletion.ts` -> `models.ts`), mutation coverage missing 15 covering files, pack-policy `@omniroute/opencode-plugin-v2/`, open-sse-typecheck fixtures missing `ProviderCandidate` fields, secrets ratchet on a loopback fixture key.

New leftovers after the chatCore extract:

6. Source-guard tests still pointed at pre-extract call sites, so Fast Quality failed on live code that had already moved.
7. Antigravity 422 dropped `error.code` inside the provider-pipeline wrapper.
8. Reset-aware combo expansion dropped a pinned `connectionId` when the connection row was missing (`connection?.provider !== provider` treated a missing row as a provider mismatch).
9. Cursor CLI version pin lived next to `node:fs`, so the provider-detail client could not import it.

Related to #12732.

## Change

- Import `upsertConnection` instead of `any`.
- `npm run gen:provider-reference` 358; bump gated copies; put the YAML closing fence on its own line.
- Regenerate `skills/cli-tunnel` as `tunnel create [type]`.
- Add the six env vars to `.env.example`.
[...]

PR #13678, opened 2026-09-14
fix(ci): clear five uncovered base-reds on release/v3.8.51 — auto-combo catalog event-loop pin, unweighted quota-share, resilience key set, pack-gate stamp, synced-catalog env doc

diegosouzapw:
⚠️ base-red inherited: #12732

## Summary

Clears five release-green failures on `release/v3.8.51` that no open PR covers. I found them by running the full slow gate set (unit, vitest, integration, pack-artifact) on an idle 32-core box against the tip plus #13635. Each red was then re-checked on the untouched tip, so none of them comes from an open PR.

Two of the five are product bugs, not stale tests:

- **Cold `GET /v1/models` pinned the event loop for about a second, and overran the 8s cold-build bound** (#12046). The built-in `auto/*` combos resolve catalog metadata for every target of every combo. The ~40 auto combos draw on the same candidate pool, and the loop neither memoized nor yielded. On the #9147 fixture (60 connections, 720 synced models) the build went from ~4s with a 167ms longest gap (parent of #12046) to ~11s with an 860–1070ms gap on an idle box. That breaks the 800ms contract and trips #12628's 8s bound, which is why `9147-catalog-eventloop-yield` came back `500 catalog_build_timeout` on every release-green run. A CPU profile put ~95% of each long stretch in that `.map`, with half of it in `getProviderPrefixes`. The fix memoizes each distinct target once per build and yields between misses. Same fixture: a 2.3–3.1s build with a 56–72ms longest gap. **The 9147 test is unchanged.** The 400→800ms bump in #12144, landed a few hours after #12046, had been hiding this.
- **A `quota-share` combo without explicit weights sent every request to its first target** (#10881). The combo resolver turns an unset step weight into 0 (`comboStructure.ts`). #10881 made `normalizeWeight` treat 0 as disabled and return definition order when the total weight is 0, so an unweighted combo had no DRR quanta. `combo-matrix/quota-share` saw openai 6/6. An all-zero set now shares evenly, and an explicit 0 still disables a target next to weighted siblings. Production-minted `qtSd/` combos always carry `weight: 100`, so they were not affected. Any hand-built quota-share combo was.

The other three (the env doc is listed under Change):

- `resilience-http-e2e`: the configuration-only key-set assertion predates #12043, which added `credentialHealthCheck` (the sweep interval) to the `/api/resilience` projection. It is configuration, not runtime breaker state. The `providerBreakers`/`runtime` absence checks are untouched.
- **Release-green pack gate:** `check:pack-artifact` builds through `build:cli`, which never writes `dist/BUILD_SHA`. With #12959's `OMNIROUTE_RELEASE_REF=HEAD` the guard still stops at "dist/BUILD_SHA is missing", reproduced on tip + #13635 + #13436, the first tree whose Turbopack build compiles. The validator now builds, then stamps, then validates, the same order `ci.yml` uses. The guard itself is unchanged. On that tree the stamped gate passes with `BUILD_SHA 5cb3ae5d9 is on the release line`.

## Change

- `src/app/api/v1/models/catalogHelpers.ts`: `memoizeTargetMetadata()` keys on provider/model/connection scope and yields after each miss. `catalog.ts` wires it into the auto loop; net +2 lines, within the file-size freeze.
- `open-sse/services/combo/quotaShareStrategy.ts`: `applyDrr` treats an all-zero weight set as unweighted.
- `scripts/quality/validate-release-green.mjs`: `runPackArtifactGate()` runs build:cli, then write-build-sha, then check:pack-artifact with `PACK_GATE_ENV`. Both entry points use it.
- `tests/integration/resilience-http-e2e.test.ts`: adds `credentialHealthCheck` to the expected key set.
- `.env.example` + `docs/reference/ENVIRONMENT.md`: document `OMNIROUTE_SYNCED_CATALOG_STALE_AFTER_MS` (#12849), which nothing else covers. This one surfaced on this PR's own Docs Gates run. The same gate also reports five more vars (`OMNIROUTE_PRESSURE_SELF_RESTART*`, `OMNIROUTE_SQLJS_WASM_PATH`, `OMNIROUTE_VACUUM_MIN_DELETED_ROWS`, `OMNIROUTE_WAL_*`), already added by #13635 and #13361. I left those alone to avoid a third collision, so **Docs Gates stays red here until one of them lands**.

## Validation

- TDD red → green:
  - `quota-share-strategy`: the new "unweighted steps still alternate" case fails on the tip (`ek-unweighted-1` ×4), passes after.
  - `validate-release-green`: the new pack-gate sequencing case fails against the tip script, passes after (32/32).
  - `9147-catalog-eventloop-yield`, unmodified: fails on the tip, passes after, 3/3 runs on the idle box.
- Integration, on the idle box: `combo-matrix/quota-share` 2/2 and `resilience-http-e2e` 5/5 (plus 3 pre-existing skips).
- Catalog suites `11947-auto-combo-modalities`, `combo-openrouter-modalities-12613`, `catalog-helpers-extraction`, `models-catalog-combo-metadata`, `model-token-limit-catalog`, `12058-models-catalog-canonical-self-aliased`, `12627-catalog-inflight-timeout`, plus the new `catalog-target-metadata-memo-9147`: 48/48 on the loaded devbox.
- Quota-share and combo routing: `quota-share-strategy`, `quota-combo-balancing`, `combo-routing-engine`, `v388-quota-share-usage-guard`, `combo-cooldown-retry`, `serial/combo-quota-share-cooldown-wait-timing`: 203/203.
- `typecheck:core` clean, `check:open-sse-typecheck` 0 errors, `check:complexity-ratchets` OK (2821/3218, 1272/1437), `check:changelog-integrity` OK, ESLint 0 errors on changed files.
- `check:file-size` shows only the pre-existing `open-sse/utils/stream.ts` drift.
- Rebased on `895dda383` (after #13349, #12959 and #13674). Merges cleanly with #13436, #13565, #13056, #13315, #13438 and #13670.

## Reviewer notes: what is still red after this, and who owns it

- **Turbopack build (168 errors):** #13283 made `ComboControlCenterClient.tsx` import `open-sse/services/model.ts`, whose dynamic `@/lib/db/*` imports drag playwright/sharp into the client bundle. #13436 and #13565 both fix it and collide with each other on `src/lib/combos/controlCenter.ts`; #13565 also collides with #13635 on `src/lib/oauth/constants/oauth.ts`. Measured: tip + #13635 + #13436 compiles and passes pack policy. Tip + #13565 alone compiles but fails pack policy until #13635's `opencode-plugin-v2` allowance lands.
- **23 combo integration reds** from the retired `claude-3-5-sonnet-20241022` fixture: #13056.
- **`chatcore-compression-integration` ×2 (pt-BR output style):** #13315 is open, but see the comment there. The English output comes from `resolveOutputStyleLanguage` auto-detection returning `en` on zero hints, which overrides the combo's explicit `pt-BR` pack. The fixture change in #13315 routes around that path instead of covering it, so this needs a product decision.
- **`qdrant-routes` embedding-models ×3:** a contract conflict. #11249/#11949 list only configured providers; #11390 merges the whole curated registry plus an unconditional default. The owner has to pick one.
- **#13635** now conflicts with the tip, because #13349 and #12959 landed overlapping pieces.


diegosouzapw:
CI triage for `ffdbeac7d`. Every red below also reproduces on a pure `release/v3.8.51` checkout (`5ca724d2`), so none comes from this diff:

| Check | Cause | Reproduced on tip | Covered by |
| --- | --- | --- | --- |
| Docs Gates | 4 code-only + 6 doc-only env vars (`OMNIROUTE_PRESSURE_SELF_RESTART*`, `OMNIROUTE_SQLJS_WASM_PATH`, `OMNIROUTE_VACUUM_MIN_DELETED_ROWS`, `OMNIROUTE_WAL_*`). This PR removed `OMNIROUTE_SYNCED_CATALOG_STALE_AFTER_MS` from that list. | yes | #13635, #13361 |
| No new ESLint warnings | `resource-pressure-self-restart` unused var + 3× `any` in `volcengine-plan-binding-upsert` | yes, same 4 errors | #13635, #13361 |
| Fast Quality Gates | `cycles` (models ↔ providers ↔ deletion), `mutation-test-coverage` (10 covering tests missing from `stryker.conf.json`), `secrets` (1 finding) | yes | #13635 / #13361 |
| Unit fast-path 1–4 | `combo-strategies` reset-aware ×11, `universal-quota-aware-routing`, the Jina duplicate listing, grok reset-credits ×3 files, the translate-path golden, the lease inventory, i18n pt-BR/vi/glossary, `upstream-status-restatement` R10, `check-env-doc-sync` and `issue-7793` | yes: an identical failure set on the tip and on this branch for the combo/quota/catalog/env files (15/82 both); the other shard-1 files 7/32 on the tip | mostly #13635 |

The files this PR touches all pass in CI: `quota-share-strategy`, `validate-release-green`, `catalog-target-metadata-memo-9147`, `9147-catalog-eventloop-yield`, `12058`, `11947`. Vitest, Merge integrity, API Route Typecheck, CodeQL and semgrep are green.

PR #13682, opened 2026-09-14
docs(i18n): resync the 8 `llm.txt` locale mirrors added in #13660

abhisheksharma2411:
`git commit` currently fails on a **clean tree** for everyone on this base. Not a claim about my branch — run it yourself on an untouched `0d13ef4fbb`:

```
$ node scripts/check/check-docs-sync.mjs
[docs-sync] FAIL - docs/i18n/km/llm.txt differs from root llm.txt
[docs-sync] FAIL - docs/i18n/kn/llm.txt differs from root llm.txt
[docs-sync] FAIL - docs/i18n/ml/llm.txt differs from root llm.txt
[docs-sync] FAIL - docs/i18n/my/llm.txt differs from root llm.txt
[docs-sync] FAIL - docs/i18n/ne/llm.txt differs from root llm.txt
[docs-sync] FAIL - docs/i18n/or/llm.txt differs from root llm.txt
[docs-sync] FAIL - docs/i18n/pa/llm.txt differs from root llm.txt
[docs-sync] FAIL - docs/i18n/si/llm.txt differs from root llm.txt
```

That checker runs in `.husky/pre-commit`, so the hook blocks every commit regardless of what it touches — a one-line fix in `open-sse/` fails it too. I hit it on an unrelated branch and had to `--no-verify`.

## Why

`checkI18nMirrorFile` requires each `docs/i18n/<locale>/llm.txt` body — everything after the mirror separator — to match root `llm.txt` byte-for-byte once normalized. These eight locales landed yesterday in #13660 generated from an older root, so they carry values root has since moved past:

| | mirrors say | root says |
|---|---|---|
| providers | 356 | **358** |
| migrations | 173 | **174** |

13 lines per file, all of them one of those two counts.

## The change

Bodies replaced from current root; **each locale keeps its own title and language nav** — those are the only localized parts of these files, and the check deliberately excludes them (`stripTopHeading` on the source, separator slice on the target).

```
 docs/i18n/{km,kn,ml,my,ne,or,pa,si}/llm.txt | 13 +++++++------
 8 files changed, 56 insertions(+), 48 deletions(-)
```

The extra insertion per file is a blank line after `---`; the 50 already-passing locales have it and these eight did not, so this also makes their shape uniform.

## Verification

The proof is that this commit went through **without `--no-verify`**, which the base cannot do:

```
[docs-sync] llm.txt i18n mirrors match root content: 58 locales
[docs-sync] CHANGELOG.md i18n translations validated: 58 locales
[docs-sync] PASS - documentation version sync is consistent.
[t11:any-budget] PASS - explicit any budget respected.
[tracked-artifacts] OK — no forbidden artifacts tracked by git
```

58 locales checked, up from the 50 that were passing.

## Worth considering separately

Nothing regenerates these on a root `llm.txt` edit, so the next provider-count bump breaks the hook again for everyone until someone notices. A `scripts/` generator (or making the check auto-fix under `--write`) would close it properly — happy to do that as a follow-up if you want it, but it seemed wrong to bundle a new script into the PR that unblocks committing.

No changelog fragment: this restores intended state from #13660 rather than changing behaviour. Say the word if you'd rather it have one under `maintenance/`.


abhisheksharma2411:
Closing this — the base fixed it independently and this PR no longer changes anything.

`release/v3.8.51` has moved 40 commits since I opened it, and root `llm.txt` plus all 59 mirrors were regenerated in that window. Verified on the current base (`c0f92ec9`) rather than assuming:

```
$ node scripts/check/check-docs-sync.mjs
[docs-sync] llm.txt i18n mirrors match root content: 58 locales
[docs-sync] PASS - documentation version sync is consistent.
```

All eight locales this PR targeted (`km kn ml my ne or pa si`) now carry `358 AI providers` from root. The pre-commit hook is unblocked, which was the whole point — so resolving the conflict here would just be re-applying a fix that already landed.

One thing that did **not** get fixed in those 40 commits, in case it's useful: the other gate in `check-docs-sync`'s neighbourhood is still red on a clean tree —

```
✗ In code but missing from .env.example: 5
✗ In ENVIRONMENT.md but missing from .env.example: 6
✗ Env / docs contract is out of sync.
```

and `tests/unit/webpack-create-require-warning.test.ts` still fails on the untouched base. Both show up as red checks on every open PR. Happy to take either if they're not already spoken for.

Issue #13681, opened 2026-09-14
db(batches): completed-batch sweep nulls a file still referenced by a surviving (non-completed) batch


diegosouzapw:
Found by the /omni-code-sec proof run (differential-review, static-analysis, codex lenses — LEDGER-14/19) on fix/batches-delete-completed-authz; pre-existing on the base (`deleteBatch` and `cleanupExpiredBatches` share the pattern).

`deleteCompletedBatches` collects the input/output/error file ids of the batches it deletes and soft-deletes them (content = NULL, irreversible) without checking whether another batch — in progress, queued, or completed but outside the current scope — still references the same file. A tenant that reuses one input file across batches loses it for the surviving ones.

Proposal: before nulling, `SELECT 1 FROM batches WHERE (input_file_id = ? OR output_file_id = ? OR error_file_id = ?) AND id NOT IN (<chunk ids>) LIMIT 1` and skip files that are still referenced (count them separately in the response). Same guard for `deleteBatch`/`cleanupExpiredBatches`.

Refs #12969, #13377.

Issue #13680, opened 2026-09-14
api(batches): DELETE /v1/batches/delete-completed does unbounded work per request — cap chunks and return a continuation


diegosouzapw:
Found by the /omni-code-sec proof run (variant-analysis lens, LEDGER-11) on fix/batches-delete-completed-authz.

Both sweep modes of `deleteCompletedBatches` loop until the table has no completed batch left; one request can do an unbounded amount of synchronous destructive work on the event loop (better-sqlite3 is synchronous), and with a second writer on the same DB file the instance loop can keep re-selecting freshly completed rows.

Proposal: cap the chunks per request (e.g. `MAX_CHUNKS_PER_REQUEST = 25`), return `{ deleted, deletedBatches, deletedFiles, hasMore }`, and carry a rowid cursor so a row that keeps matching the predicate is never re-selected. Test: seed > cap*200 batches and assert `hasMore: true`.

Refs #12969, #13377.

PR #13684, opened 2026-09-14
fix(security): resolve findings from omni-code-sec battery (fix/batches-delete-completed-authz)

diegosouzapw:
## Summary

Resolves the four approved findings of the omni-code-sec proof run on the completed-batch sweep (`DELETE /v1/batches/delete-completed`, hardened in #12969 / #13262 / #13297 / #13374). TDD: every item has a test that failed on `origin/release/v3.8.51` (5ca724d2) and passes here.

| Item | Ledger ids | Change | Covering test |
| --- | --- | --- | --- |
| **T-C** key mode held one write lock for its whole sweep | LEDGER-4 (×5), 20, 21 | `deleteCompletedBatches` key mode now commits per 200-id chunk exactly like instance mode: the outer `sweepKey` transaction is gone and both modes share one `sweepLoop(nextIds)` helper — the only difference between the modes is the SELECT that yields the next chunk. No statement binds more than `INSTANCE_SWEEP_CHUNK` ids. JSDoc updated (both modes per-chunk, lock never held across chunks, shared-file-across-chunks caveat now applies to both modes). | `batches-delete-completed-ownership-wvxc.test.ts`: "a failure in the key sweep's second chunk keeps chunk 1 committed and rolls chunk 2 back" (replaces the whole-sweep-rollback test); "swept in one call, one commit per chunk" now asserts `runs === 2` (was `>= 3`) |
| **T-C** forward-progress guard | LEDGER-22 | `sweepLoop` remembers the previous chunk's first id and throws `deleteCompletedBatches: no progress — chunk repeated (id …)` when the next chunk starts with the same id (the DELETE removed nothing). A concurrent deleter only makes rows vanish → different first id or empty chunk → never a false failure. | same file: "FORWARD PROGRESS: a chunk whose DELETE silently removes nothing throws instead of looping forever" (`BEFORE DELETE … RAISE(IGNORE)` trigger; on the base this test hangs forever — recorded as killed by `timeout`) |
| **T-E** route skipped the per-key operator policy | LEDGER-9, 13, 16 | `route.ts` calls `enforceApiKeyPolicy(request, null)` right after the `validateApiKey` lifecycle gate (kept — the enforcer's status check does not read `revoked_at`) and before `sweepScope` is chosen; a rejection is returned as-is (the enforcer's own `errorResponse`). Also fixes a general fail-open found while verifying this: `validateEndpointAccess` resolved the category from the raw pathname, and a client hitting the App Router path `/api/v1/…` directly (no `/v1/:path*` rewrite fires) saw `resolveEndpointCategory` return `null` → `allowedEndpoints` silently passed for **every** `/v1` route on that shape. One line normalizes `^/api(?=/v1/)` away inside `validateEndpointAccess`. | `batches-delete-completed-route-scope.test.ts`: key with `allowedEndpoints: ["chat"]` → 403 from the enforcer and the batch + file content survive, for both `http://localhost/api/v1/…` and `http://localhost/v1/…`; `endpoint-restrictions-policy.test.ts`: "search-only key blocks /api/v1/chat/completions too" (fails open on the base); `endpoint-categories.test.ts`: `resolveEndpointCategory("/v1/batches/delete-completed") === "batches"` |
| **T-G** `warn` on every rejected credential | LEDGER-12 (×4) | The one rejection-branch log call is now `log.info` (any caller can hit it with any string; the 401 is the audit signal). The real sweeps keep their `warn` audit lines. No test asserted the level. | — (code-only; grep `presented API key rejected` in tests/ → none) |
| **T-I** truthiness-based mutual-exclusion guard | LEDGER-18 | `if (allTenants && apiKeyId)` → `if (allTenants && "apiKeyId" in scopeObj)`: `{ allTenants: true, apiKeyId: "" }` and `{ …, apiKeyId: null }` now throw instead of widening to an instance sweep. | ownership file: "rejects allTenants combined with an EMPTY or NULL apiKeyId too" |

### Why the `/api/v1` path shape matters (evidence)

Next.js keeps the client's original URL in the handler: `resolveRoutes` returns the rewritten `parsedUrl` but never reassigns `req.url` (`node_modules/next/dist/server/lib/router-utils/resolve-routes.js:607`), the render worker only mutates `parsedUrl.pathname` in its `invokePath` block (`base-server.js:939-963`), and `NextRequestAdapter.fromNodeNextRequest` builds `request.url` from `req.url` (`next-request.js:82-101`). So `/v1/batches/delete-completed` reaches the handler as `/v1/…` (category resolves), while a direct call to `/api/v1/batches/delete-completed` — a public path the App Router serves natively — reaches it as `/api/v1/…`, where the `/v1/…` category prefixes never match and the allowlist failed open. The normalization closes that for every caller of `enforceApiKeyPolicy`.

### Test evidence

RED on the base (before the fix):
- ownership file: `# pass 11 / # fail 3` — `3 !== 2` (outer transaction still counted), chunk 1 rolled back instead of committed, `Missing expected exception` for the presence guard; the no-progress test alone → killed by `timeout 60` (exit 124, infinite loop).
- route-scope file: `# pass 10 / # fail 1` (policy test).
- endpoint-restrictions-policy: `# pass 11 / # fail 1` — "Should reject the /api/v1 shape as well".

GREEN (this branch):
- `batches-delete-completed-ownership-wvxc` 15/15 · `batches-delete-completed-route-scope` 11/11 · `endpoint-categories` 30/30 · `endpoint-restrictions-policy` 12/12
- `batch-deletion`, `files-delete-owned-by`, `api-key-policy*`, `apikey-policy-*`, `apikeypolicy-*`, `exclusive-lease-api-key-policy`, `elevenlabs-policy`: 83/83
- `npm run -s typecheck:core` clean · `eslint --suppressions-location config/quality/eslint-suppressions.json <changed files>` clean (the 5 unsuppressed hits are pre-existing frozen violations) · prettier applied · pre-commit hooks (lint-staged, docs-sync, any-budget, tracked-artifacts) passed.

⚠️ base-red inherited: #12732

Refs #12969, #13680, #13681, #13377


diegosouzapw:
## Full suite on the .113 box (code head `6795e7b7`; `aebd839c` only touches `stryker.conf.json`)

| suite | head | base `5ca724d2` (release tip) |
|---|---|---|
| `test:unit` | 38 834 pass / 46 fail / 28 skipped — 0 in batches, files, api-keys, policy, endpoint-categories | — |
| `test:vitest` | passed | — |
| `build` | `Turbopack build failed with 168 errors` (module-not-found `fs`/`net`/`child_process` in executors outside this diff) | **same failure** (rc=1, same 168-error signature) → inherited (#12732 "Package artifact") |

Discriminator with the box idle: every one of the 25 unit files that fail in the full run fails **identically** at the base tip in isolation (`_tasks` record: `validation-113-disc.log`); the only one that passed, passed on both sides (load noise). No failure is introduced by this PR.

CI on this PR: `Docs Gates` (env/docs contract drift) and `Fast Quality Gates` → `secrets` + `cycles` reproduce on the release tip in a disposable worktree → inherited (#12732); `mutation-test-coverage` was this PR's (touching two mutation-tested modules) → fixed in `aebd839c`. Final security review (whole diff): ready to merge; follow-up for the alias rewrites in #13685.

PR #13687, opened 2026-09-14
fix(embeddings): log server-side when a provider can't be resolved

hartmark:
## Problem

Ping (the operator's live gateway) hit `400 Unknown embedding provider: llama-cpp` on essentially every memory-recall embedding call, all day, for days -- but the OmniRoute dashboard and `call_logs` showed nothing at all. The only trace was in the calling client's own log.

## Root cause (data, not code -- fixed directly on the live instance, not part of this PR)

Two independent, pre-existing data problems on the `llama-cpp` connection:

1. A `customModels` override meant to flag a local model as embeddings-capable (llama.cpp's own `/v1/models` never reports capability data) used provider prefix **`llamacpp`** (no dash) instead of the connection's actual resolved provider id **`llama-cpp`** -- so it could never match a real request.
2. The connection's synced-model cache had gone stale (still listing an old 4B model filename after a swap to a 0.6B model) because `autoSync` had never been enabled on it (`synced_models_at` was `NULL` since setup).

With both broken, `resolveLocalSyncedEndpointRoute()` -- the *only* viable resolution path for a local self-hosted embedding provider not in the static registry and with an empty `provider_nodes` table -- returned `null` for every possible model string on that connection, current or historical (verified by calling the real function directly against the live DB).

## Why this took real digging to find

`src/lib/embeddings/service.ts`'s final `!providerConfig` branch returns the 400 *before* any `call_logs` row gets written, so this whole failure class is invisible in OmniRoute's own logs and dashboard. Diagnosing it required correlating the calling client's log against the live DB state by hand.

## Fix (this PR)

Add a `log.warn` right before that response so a future occurrence of this failure class shows up in OmniRoute's own logs, with enough context (provider, requested model, which resolution paths were checked) to diagnose without needing a separate client's logs. No behavior change.

Also adds the first direct test coverage for `createEmbeddingResponse`'s "Unknown embedding provider" / "Invalid embedding model" 400 branches -- every existing test for this function previously only exercised a provider that resolves successfully.

## Evidence

- New test (`tests/unit/embeddings-unknown-provider-response.test.ts`) proves both 400 branches and that the new warn log fires.
- 43 directly-relevant embeddings/ollama-local/memory-embedding tests re-run green (no regressions).


diegosouzapw:
Thanks for the deep dig here — the write-up connecting the silent 400 branch to two
independent, pre-existing data problems on the live connection is exactly the kind of root-
cause trail that makes this fixable next time without cross-referencing a client's own logs.
Ran your new test file in isolation: both 400 branches pass and the warn log fires with the
full diagnostic context you described. This looks merge-ready as-is.

PR #13701, opened 2026-09-15
ci(acceptance): emit a shadow release-acceptance report next to release-green

HouMinXi:
## Summary

Adds a version-1 release-acceptance report alongside `release-green.json`. The report distinguishes a failed product check from a check that did not run or lacks usable evidence.

The existing nightly auto-close policy stays in place. The `Release acceptance` workflow runs on `release/v*` pushes and manual dispatch, uploads a shadow report, and is not a required Mergify check.

Related to #12732. This description does not change the tracker's lifecycle.

## Behavior

- The reducer carries failed prerequisites into dependent gates and records their cause. Missing or skipped required gates produce an unverified verdict; infrastructure errors remain distinct from product failures.
- Gate records include suite and shard identity. Conflicting or incomplete evidence cannot establish a verified result.
- The inventory uses the existing test-discovery collectors. Unexecuted `.tsx` tests are recorded separately from inventory errors.
- The compiler adapter accepts empty diagnostics only with exit 0 and a nonempty evidence digest. The Node test adapter preserves the full path in Test Anything Protocol subtest records.
- Schema validation rejects evidence paths containing parent traversal. Invalid reports retain an error record rather than silently passing.

## Scope

This is a shadow report. It does not replace the pack-boot hard-fail path, add a Mergify requirement, or remove either nightly `gh issue close` step. The checked-in workflow uses fixture manifests to exercise report generation; it does not yet collect live results from every release gate.

The maintainer's `95b3beebb0` keeps auto-close and adds `persist-credentials: false` to checkout. Those changes are retained.

## Verification

Local integration with upstream `c68b7f58c2`:

- `node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --import ./tests/_setup/isolateDataDir.ts --test tests/unit/release-acceptance-*.test.ts`: 49 passed, 0 failed, 0 skipped.
- ESLint on the release-acceptance scripts and tests: exit 0.
- The CLI run with the checked-in lint plan and shadow manifest wrote a report with a verified verdict and exited 0. This verifies the fixture path, not the release branch's overall health.

These results are from the local integration candidate. That candidate has not been pushed; this section is not a claim that all pull request checks are green.


diegosouzapw:
Thanks for this — separating "the job didn't run" from "the product failed" in a dedicated
shadow report is a real improvement, and the reducer/schema/TAP-parser work is solid: all 45
new tests pass, ESLint is clean, and the merge is conflict-free against the tip. Two small,
purely mechanical things we'll fix in place on your branch (no action needed): (1) the new
`release-acceptance.yml` adds 5 zizmor findings (3 unpinned-uses + 1 artipacked + 1
cache-poisoning) that aren't reflected in `quality-baseline.json`'s `zizmorFindings` ratchet —
that's what's failing "Fast Quality Gates"; and (2) the new `actions/checkout` step is missing
`persist-credentials: false`, which every other workflow sets per the 2026-06-16 hardening pass.

One part we're not taking as-is: removing the two `gh issue close` steps from
`nightly-release-green.yml`. That auto-close is deliberate (#12085) — it's what closes the
base-red tracking issue (#12732) once the branch is green again, and several downstream tools
read that issue's open/closed state. With #13678 in flight to fix the underlying base-reds,
dropping it now would leave #12732 open indefinitely. We'll restore those two steps in place
(and align `closeOracle`'s test with that) and land the shadow report; if you'd like to argue for
manual closing, a separate PR with the reasoning + the matching `AGENTS.md` "Base-green check"
update is the right place.

⚠️ base-red inherited: #12732 — the API Route Typecheck / Docs Gates / Merge integrity /
No-new-ESLint-warnings / Unit-fast-path failures on this PR line up with the currently tracked
base-red set and are not caused by your changes.

HouMinXi:
@diegosouzapw Agreed: keep the existing nightly auto-close behavior. I am not proposing a manual-close policy change in this PR. The shadow report should distinguish an unavailable run from a product failure without changing the release tracker lifecycle. Restoring the two close steps and aligning closeOracle is the right scope. Please retain the workflow credential hardening as well. This contribution is intended for review and merge; there is no author-requested hold.

PR #13702, opened 2026-09-15
fix(mitm): bound SSE transcript retention and cancel abandoned upstream reads

### LINKED ISSUE #13395: fix(resilience): bound SSE aggregation and cancel abandoned streams to stop gradual RSS growth
## Summary

On OmniRoute 3.8.50, a long-running Linux instance showed high and gradually increasing resident memory during streamed provider traffic. Restarting the service reduced memory temporarily, but it returned to multi-GB RSS. The behavior is not explained by the JavaScript old-space limit alone.

This report concerns generic SSE/inspector retention and abandoned streams. It is related to #11804, but the observed route was not `auto/*` or `fusion`; the evidence below points to independent unbounded collectors and missing stream cancellation.

## Environment

- OmniRoute: 3.8.50 (global npm installation)
- Node.js: v24.19.0
- OS: Linux, user-level systemd service
- `NODE_OPTIONS`: `--max-old-space-size=1024`
- cgroup limit: 6 GiB
- The affected traffic included streamed provider requests and tool/text payloads.

## Observed behavior

- In the incident window, the previous process reached approximately 5.8 GiB RSS and 13.4 GiB swap.
- After restart, the new process still reached approximately 3.1–3.2 GiB RSS with cgroup usage around 3.6–3.8 GiB while the live JS heap was only about 300–370 MiB.
- A temporary Inspector session reported RSS around 3.15–3.17 GiB, `heapUsed` around 295–367 MiB, and `external` around 50–54 MiB. A forced GC reduced RSS by about 182 MiB without materially changing `heapUsed`, suggesting allocator/native/page retention in addition to ordinary JS heap usage.
- A deterministic single production request that accounts for the entire RSS increase was not isolated.

## Code-level findings

The following independent retention paths were found in the 3.8.50 source:

1. `open-sse/utils/streamPayloadCollector.ts` limits the number of events but continues appending unbounded text, reasoning, tool arguments, Claude blocks, and Gemini parts. A large payload can therefore bypass the outer event limit.
2. `src/mitm/inspector/buffer.ts` limits the number of entries and caps individual bodies, but has no aggregate byte budget. A full 1000-entry buffer can retain a large amount of request/response data.
3. `src/mitm/handlers/base.ts` accumulates `collected += chunk.toString()` for every SSE chunk without a byte ceiling. Downstream disconnects and idle upstream streams were not propagated to the fetch/reader, and writes did not consistently apply backpressure.
4. The standalone `src/mitm/server.cjs` path had the same class of risks: an oversized single chunk could bypass the nominal capture check, and router fetches were not cancelled when the downstream client closed or stopped producing data.

## Minimal reproductions

The following focused regression cases failed before the local mitigation:

- With `maxEvents=1`, OpenAI/Claude/Gemini reducers retained 400,000-character payloads.
- A 1.5 MiB SSE response was passed in full to the inspector capture callback.
- A TrafficBuffer configured with an 8 KiB intended total budget retained roughly 200 MiB because only the entry count was enforced.
- Downstream-close and idle-timeout tests could not observe a shared stream lifecycle abort because no lifecycle helper existed.

## Expected behavior

- Event-count limits should also bound retained text and tool payload bytes.
- Inspector/debug buffers should enforce both entry-count and aggregate-byte limits.
- The complete upstream stream should continue to the client, but diagnostic capture should be bounded.
- Closing the downstream response or exceeding an idle/max-duration timeout should abort the upstream fetch and cancel the reader.

## Local mitigation and validation

On one affected instance, I applied and deployed a local mitigation that:

- bounds live reducer summaries;
- adds a 64 MiB aggregate TrafficBuffer budget;
- limits inspector capture to 1 MiB while preserving full stream forwarding;
- adds downstream-abort, idle-timeout, 30-minute max-duration, reader-cancellation, and backpressure handling to both MITM paths.

Validation results:

- Focused tests: 59 passed, 0 failed.
- MITM regression group: 71 passed, 0 failed.
- Core typecheck, focused ESLint, syntax check, and production builds passed.
- After deployment, `/healthz`, `/livez`, `/readyz`, and `/api/health/ping` returned HTTP 200.
- During a 50-second idle sample after deployment, RSS remained approximately 2.876 GiB with no swap growth; the resident baseline is still high and needs longer real-traffic observation.

## Request

Please confirm whether this should be tracked separately from #11804 and advise the preferred upstream patch/PR target. The main question is whether these generic stream/reducer/inspector retention paths are already covered elsewhere or should be fixed as a separate resilience issue.

oyi77:
Fixes #13395 — unbounded SSE/inspector retention and abandoned streams on the MITM pipe paths.

## Scope note

Of the four retention paths named in the issue, two are already fixed at the tip and were verified, not re-fixed:

- `open-sse/utils/streamPayloadCollector.ts`: defaults are now `maxEvents=2000, maxBytes=524288`, the reducer ingests before the cap check, and the deep clone is deferred past the cap — the old 200/48KB claim no longer holds.
- `src/mitm/inspector/buffer.ts`: per-body clamp `INSPECTOR_MAX_BODY_KB` (default 1 MiB) on both `push` and `update`, ring-rotated at 1000 entries.

The two paths that were still open are fixed here:

1. **`src/mitm/handlers/*.ts` (8 handlers)**: `let collected = ""` + `+=` grew without bound in the handler closure for the whole stream lifetime, before ever reaching the inspector clamp. Replaced with `createBoundedCollector()` (new in `base.ts`): retains at most 1 MiB (aligned with the inspector default), keeps counting true bytes so `responseSize` is byte-exact with the old `Buffer.byteLength(collected)` semantics, and never affects the pipe — every chunk is still written downstream.
2. **Abandoned streams**: neither `pipeSSE` nor `server.cjs` propagated downstream disconnects. `pipeSSE` now cancels the reader on `res` close (and detaches the listener); `server.cjs` aborts the router fetch via `AbortController` and cancels the reader on close.

## Verification

- New `tests/unit/mitm-pipe-bounded-collect-13395.test.ts`: 7/7 (collector cap + byte counting, full delivery on open pipe, early stop after downstream close, listener detach, `server.cjs` abort/cancel shape).
- All 9 sibling MITM suites (base + 8 handlers): 31/31 pass, including every handler happy path through the new sink.
- `check-file-size`: only red is `open-sse/utils/stream.ts` (pre-existing, untouched). `check-changelog-integrity`: fails identically on the clean tip (proven via stash on the sibling branch) — pre-existing drift.


diegosouzapw:
Thanks for this — solid, well-scoped fix with real regression tests (7 new + all 31 sibling
handler suites + the server.cjs suites stay green in my validation). The bounded collector and
the downstream-close cancellation for both `pipeSSE` and `server.cjs` are exactly what #13395
asked for on those two sub-paths, and the byte-exact `responseSize` preservation is a nice
touch.

One thing before merge: the commit says `Fixes #13395`, but re-reading the issue's own findings
list, two sub-items aren't covered by this diff — `pipeSSE`'s `res.write()` still has no
backpressure handling, and `server.cjs`'s `respBody += text` accumulator (line ~600) can still
overshoot `INGEST_MAX_BODY` by a single oversized chunk, which is literally the "oversized
single chunk could bypass the nominal capture check" finding from the issue. Since this repo's
default branch auto-closes linked issues on merge, could you change it to `Refs #13395` (or
leave `Fixes` and open a quick follow-up issue for those two items) so #13395 doesn't get closed
as fully resolved? Everything else here looks ready to me.

Several of the check runs showing red on this PR (API Route Typecheck, Merge integrity, the
fast-path unit/docs/lint jobs) are unrelated to this diff — API Route Typecheck only scopes
`src/app/api/**`, which this PR never touches, and I reproduced `check:changelog-integrity` /
`check:agent-skills-sync` passing locally against your exact head commit. ⚠️ base-red inherited:
#12732 covers the rest (ESLint/docs/unit-suite reds already present on the release tip).

diegosouzapw:
Thanks @oyi77 — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

oyi77:
Follow-up for the two uncovered sub-items noted in review (backpressure on `res.write()`, single-chunk `respBody` overshoot): opened #14528 since #13395 auto-closed on merge. Happy to take that one too if nobody picks it up.

PR #13717, opened 2026-09-15
fix(db): bound health scans and isolate native diagnostics

HouMinXi:
## Current behavior

Database health checks load all quota snapshots at once and repeatedly prepare connection lookups. A large history can block request handling and consume substantial native memory. The endpoint is not the only caller: startup, periodic repair and the MCP health tool share this path.

## Problem

The previous synchronous scan couples database maintenance to the request thread. Moving it into a child alone would leave unbounded reads, overlapping checks, unsafe repair ordering and shutdown races unresolved.

## Changes

- Scan quota snapshots in pages of at most 1,000 rows, reuse lookup/delete statements, and preserve signed SQLite int64 IDs without JavaScript number rounding. Capture an upper bound so concurrent inserts cannot extend the scan indefinitely.
- Run file-backed native checks in a cancellable child. Coalesce checks, cache diagnoses for up to 60 seconds, await results at HTTP and MCP callers, and cancel/reap the child before database shutdown.
- Require a successful backup before mutation. Native repairs recheck within a write transaction and roll back on failure.
- Keep sql.js automatic repair on its owning connection. Copy main into an exclusively allocated temporary WASM database and export only that copy, preserving owner settings, TEMP data, functions and statements. Reject unsupported busy/read-only states without mutating data.
- Include the worker in standalone and published artifacts.

Two independently reproduced baseline failures are separate commits: defer dependent FTS migration 178 on drivers without FTS5; isolate the XP cleanup fixture from its unrelated asynchronous file logger. The startup test verifies a subsequent native restart applies the deferred migrations.

## Verification

[KNOWN/HIGH] Local Linux, Node 22.23.1:

- Selected database/lifecycle/packaging/sanitization regression: **1,111 passed, 0 failed**.
- MCP health-tool suite: **3 passed**. Core typecheck: exit 0.
- Real native and sql.js repair, rollback, shutdown cancellation, int64 boundaries, snapshot collision and owner-state tests.
- Before/after tests and deliberate omissions at real repair/dispatch/backup sites fail with the defect and pass after restoration.
- Three consecutive inline review cycles, each covering structural, architectural and adversarial checks, completed on the unchanged 26-file diff. This is one reviewing model with multiple passes and executable checks, not independent-model consensus.
- A synthetic scale run with **967,470 valid snapshots plus one invalid row** completed in **3.92 s**. Diagnostic-child peak RSS was **114 MiB**; **314 local HTTP requests** completed, maximum observed latency **33.4 ms**. This is a local benchmark, not a production capacity guarantee.

Known limits: no full Next/Electron build or cross-platform certification. sql.js work remains synchronous and its main-schema snapshot excludes attached/TEMP schemas. Existing ESLint debt remains; comparison against the base found zero new lint errors, not a clean full-repository lint run.

⚠️ base-red inherited: #12732. Related to #12732; #12732 stays open.

## Result

Native health maintenance no longer performs the quota scan on the request thread. sql.js automatic repair remains supported with backup-before-write protection. No production deployment is included in this PR.


HouMinXi:
Updated in 6a6daa3700f16baee59f3ef8ba59c6b583829b3e to address the three CI regressions attributed to this PR:

- Await terminal database shutdown in the provider-boundary fixture instead of resetting while startup health work is active. Startup checks and the production busy guard remain enabled.
- Extract quota-row validation and disposable sql.js snapshot creation without changing their behavior. Both introduced cognitive-complexity violations are gone; thresholds are unchanged.

Verification on the reviewed files:
- Related database, provider-boundary, packaging and sanitization tests: 1,110 passed, 0 failed, 0 skipped on Node 24.20.0.
- Core typecheck passed. Complexity ratchet against c0f92ec98a981cc13c25e44048a04b33c62a3112 reports complexityNewCode=0 and cognitiveComplexityNewCode=0.
- Restoring the old reset call reproduces the busy-guard failure; awaited shutdown passes. Removing the snapshot copy or timestamp rejection also fails the corresponding real-path tests, and restoring each passes.
- Three review cycles completed for this three-file correction. The signed commit matches the reviewed diff.

The baseline failures identified in the previous CI run remain outside this correction. The new GitHub checks are not yet claimed green. No deployment or merge performed.

diegosouzapw:
Thank you for this — the paginated, int64-safe quota-snapshot scan and moving native
diagnostics into a cancellable child process are exactly the right shape for this problem, and
the packaging wiring across all three build paths (with an end-to-end test that actually forks
the compiled worker) is genuinely thorough work. The test suite you added is extensive and none
of the existing exact-snapshot assertions were weakened. Merge is conflict-free against the tip.

Two things we'll handle on our side before merging, no action needed from you:
1. `/api/db/health` now (transitively) forks a child process, and this repo classifies every
   process-spawning route as loopback-only in `src/server/authz/routeGuard.ts` (Hard Rules
   #15/#17 — same treatment as `/api/db-backups/exportAll`). We'll add the prefix + the
   `isLocalOnlyPath()` assertion, unless the owner prefers to accept the authenticated fork given
   your coalescing + 60s cache.
2. A merge of the current `release/v3.8.51` tip into the branch: the `API Route Typecheck` red
   (TS2741 in `modelTestRunner.ts`, a file you don't touch) was a defect of the base itself and is
   already fixed on the tip by #13730; the `check:secrets` hit is likewise in an untouched file.
   Everything else red on this PR is the tracked base-red set — ⚠️ base-red inherited: #12732.

Coordination note: #13149 edits the same route + sanitization test (your cache/child-process
design likely covers its `skipIntegrityCheck` case), and #13341/#12992 touch the same backup
pruning code your new `managedBackup.ts` replaces — we'll reconcile those at merge time.

PR #13720, opened 2026-09-15
fix(chat): preserve suffix reasoning intent across model attempts

HouMinXi:
## Problem

Resolved model suffixes such as `-max` and `-none` could be dropped between model resolution and chat dispatch. Passing the value through alone is insufficient: after a handler selects a replacement model, an effort already injected into the shared request body looks like explicit client intent and leaks into the replacement request.

## Change

Keep suffix provenance separate from the mutable request. Resolve defaults and dependent parameter constraints for each handler model attempt, including credential refresh and server-tool follow-ups. Explicit client reasoning fields remain authoritative even when model constraints remove their wire representation.

Include translated reasoning intent in context-aware deduplication hashes. Concurrent requests with different effort settings must not share a response. The existing two-argument hash interface and tenant namespace remain unchanged.

## Verification

The post-commit batch passed 268 tests, with zero failures or skipped tests. Core TypeScript checking and formatting checks on all eight changed files passed. All 39 exact-site mutations triggered test assertion failures before restoring the source.

Local HTTP tests cover same-account retries, actual two-account fallback, model replacement, tool follow-ups, refreshed credentials, and overlapping high/none requests in both arrival orders. Equal high/high requests still deduplicate. Eleven baseline/candidate first-request wire cases also matched.

Three consecutive code-review cycles completed without a confirmed finding. This was one review context using three lenses per cycle, not independent-model consensus. The commit hook subsequently reformatted one existing call and added a trailing object comma; that non-behavioral delta was checked separately and the tests rerun.

## Scope and known limitations

No deployment or production model probe is included. This fixes handler-level attempts, not provider-native target remapping or payload rules that rewrite model identity.

Two existing pipeline recovery defects were reproduced separately on the unchanged base: model-unavailable fallback can resend the old model, and non-streaming context recovery can return before fallback. Six corresponding baseline-failing cases remain separate diagnostic artifacts; they are not skipped tests in the passing batch.

Full-repository lint is not claimed green: 11 inherited unused-symbol diagnostics remain in the handlers. The existing unrelated stream file-size check also fails on the base. Neither was folded into this change.


diegosouzapw:
Thanks for this — the suffix-effort/retry analysis is sharp, and the incidental fix (the
credential-refresh retry used to skip `prepareUpstreamBody` entirely, not just reasoning
defaults — payload rules and `prompt_cache_key` too) is a nice catch beyond the stated scope.
The new test suite is genuinely thorough (27 HTTP-harness scenarios, including real concurrent
requests over a live socket) and the dedup-hash extension has an explicit anti-spoof test,
which we appreciated.

Two things are blocking green right now, both mechanical and both ours to fix in-place: the
`complexity-ratchets` gate flags 2 new violations in the new `prepareUpstreamBody` normalization
chain (needs splitting into a helper), and `file-size` flags `src/sse/handlers/chat.ts` growing
past its frozen cap — mostly from the pre-commit hook's own Prettier reformat of an
already-over-100-col line you mentioned in the PR body. We'll handle both plus a rebase (which
also clears an unrelated `check:secrets` finding that turned out to be our branch trailing the
current tip on a file this PR doesn't touch, not something in your diff).

⚠️ base-red inherited: #12732 covers most of the remaining red (ESLint, docs env-var sync,
agent-skills-sync, a golden-snapshot test, and a pre-existing `modelTestRunner.ts` TS2741) —
confirmed identical on the current base tip, unrelated to this PR.

PR #13725, opened 2026-09-15
fix: route OpenCode Muse Spark family through Responses API

AStupidBear:
## Summary

- Route the OpenCode Muse Spark family (`muse-spark-*`) through the OpenAI Responses API for all OpenCode aliases, including future model versions.
- Normalize provider-prefixed model IDs before dispatching to OpenCode Zen.
- Add regression coverage for family routing, future IDs, alias resolution, and all supported provider prefixes.

## Validation

- Targeted OpenCode/Muse tests: 9 passed.
- Backend-only Webpack build: `npm run build:backend` with `OMNIROUTE_USE_TURBOPACK=0`.
- npm/CLI artifact build: `npm run build:cli`.
- Pack artifact policy: passed with `OMNIROUTE_ALLOW_CANARY_BUILD=1` for this feature branch.

The local full Dashboard bundle still hits existing server-only dependency resolution errors in both Turbopack and full Webpack; the backend-only build produces the standalone service artifact successfully.


diegosouzapw:
Thanks for this — routing the whole `muse-spark-*` family through the Responses API by
pattern (instead of enumerating each new version id) is the better long-term fix here; it's
the same approach the codebase already uses for OpenAI's `-pro` models, and it should stop
the recurring "new Muse Spark release needs its own PR" pattern we're seeing right now with
1.3. Tests pass locally (4/4 new, plus the neighboring opencode/providerModels suites, 91/91).

One thing we'll add before merging: a `changelog.d/fixes/` fragment (missed in this PR, per
`CONTRIBUTING.md`). Everything else looks solid — no security or architecture concerns.

Note: several CI checks show red on this PR (API Route Typecheck, Docs Gates, ESLint, some
Unit Tests jobs). We checked each one and none touches the files this PR changes — they're
all inherited from the current base-red state (⚠️ base-red inherited: #12732), not caused by
this PR.

Separately, there are a few other open PRs fixing Muse Spark 1.3 specifically by adding
curated catalog entries (context window, etc.) — those are complementary to this one, not a
duplicate, since your fix covers future versions this one can't predict while theirs gets the
exact context window right for 1.3 today. We're planning to merge both.

PR #13704, opened 2026-09-15
fix(build): keep browser-reachable modules off node-only imports so next build --turbopack succeeds again

seanford:
## Summary

`next build --turbopack` — and therefore every Docker image build — has been failing on `release/v3.8.51` since the 2026-09-11 owner batch. Docker Hub confirms it: the last `next` / `next-web` images were published 2026-09-11 18:16 UTC, and nothing has been published from any of the ~150 commits since. Two commits in that batch each introduced a server → client import edge, and Turbopack fails on the first one it meets, which is why the second was hidden until the first was fixed:

1. **#13283** made `src/lib/combos/controlCenter.ts` — imported by the `"use client"` `ComboControlCenterClient` page — import `resolveProviderAlias()` from `open-sse/services/model.ts`. That module lazily imports the DB layer (`@/lib/db/readCache`, `@/lib/db/models`, …); Turbopack follows those `await import()` edges into the browser bundle, drags the executors (down to the Playwright-backed `zai-web` browser automation) along, and the build dies with **168 × `Module not found: Can't resolve 'child_process'` / `'async_hooks'`** from `playwright-core` and `detect-libc`.
2. **#13264** made the browser-reachable provider registry entry `open-sse/config/providers/registry/codebuddy-cn/index.ts` import `CODEBUDDY_CN_USER_AGENT` from `src/lib/oauth/constants/oauth.ts`, which imports `open-sse/utils/cursorAgentCliVersion.ts` (`node:fs` / `node:os` / `node:path`). With (1) fixed, the build fails one step later on every dashboard page: `TurbopackInternalError: Failed to write app endpoint /(dashboard)/dashboard/page — the chunking context (unknown) does not support external modules (request: node:fs)`.

## Changes

- `open-sse/services/providerAlias.ts` (new, pure): `ALIAS_TO_PROVIDER_ID` + `resolveProviderAlias()`, depending only on the static provider registry. `model.ts` imports and re-exports both, so its 8 existing server-side importers are untouched; `controlCenter.ts` imports the pure module.
- `open-sse/config/providers/registry/codebuddy-cn/userAgent.ts` (new, dependency-free): `CODEBUDDY_CN_USER_AGENT`. The registry entry imports it directly; `src/lib/oauth/constants/oauth.ts` imports and re-exports it, so OAuth, chat headers and the usage/quota client keep sharing the single string that #12702 requires.
- Two `changelog.d/fixes/` fragments (placeholder `0000` — I'll rename to this PR's number once assigned).

[...]

Issue #9123, opened 2026-08-01
feat(providers): search-provider playground UX + clearer local-IP flag naming (OMNIROUTE_ALLOW_LAN_PROVIDERS)


rnahlawi:
### OmniRoute Version

3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Provider(s) Involved

SearXNG

### Description

Thanks for OmniRoute — it’s a really useful gateway. I hit two issues while wiring up a self-hosted SearXNG search provider; both cost several hours to track down and I think both are easily fixable or documentable. Sharing in case it’s helpful.

Self-hosted deployment: diegosouzapw/omniroute:latest via Docker Compose, behind nginx-proxy-manager. Search provider: SearXNG (self-hosted metasearch) added as type searxng-search with base URL http://<lan-host>:8888/search. The same shape reproduces with Brave.

[...]

Issue #5066, opened 2026-06-26
[BUG] Unable to properly use and add local OpenAI-compatible provider


daniij:
### OmniRoute Version

3.8.36

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

v24.15.0

### Provider(s) Involved

any local

### Model(s) Involved

any

### Client Tool

any

### Description

There is a critical validation issue when adding and configuring local OpenAI-compatible providers (e.g., running on 127.0.0.1 or localhost):

1. Provider Creation Stage: When filling out the "Add OpenAI Compatible" form with a local URL (e.g., [http://127.0.0.1:3264/api](http://127.0.0.1:3264/api)), the system throws a red error: "Blocked private or local provider URL". This error prevents the form from verifying provider availability, yet clicking "Add" still successfully creates the provider in the background.
2. Connection / Key Stage: Because the local URL is flag-blocked or restricted by this security check, adding a connection (API Key) becomes completely impossible. The UI shows an "Invalid" status and fails to reach the /models endpoint for validation.

This is strictly an internal application validation/restriction bug: a direct network request via terminal (curl) confirms that the local endpoint is up, perfectly healthy, and correctly responds with a 200 OK and a valid OpenAI-compatible model list.

### Steps to Reproduce

1. Open "Add OpenAI Compatible" modal.
2. Enter a local Base URL: [http://127.0.0.1:3264/api](http://127.0.0.1:3264/api).
3. See the error "Blocked private or local provider URL", which halts the standard availability check.
4. Submit/bypass the form; notice that the provider "Qwenf" is successfully saved to the dashboard.
5. Go to the newly created provider details and click "+ Add" under Connections.
6. Try to save a key (even an optional/dummy one like . since local endpoints don't require auth).
7. The application returns "Invalid" because its internal routing fails to ping /models due to the local URL block.

### Expected Behavior

There should be a mechanism (e.g., an "Allow local/private networks" toggle in Advanced Settings or automatic detection) that permits the application to verify and fetch data from 127.0.0.1, localhost, and private IP ranges. Users developing with local models or custom local proxy routing stacks need to be able to validate these endpoints.

### Actual Behavior

Provider Creation Form:

1. When attempting to add an OpenAI-compatible provider with a local loopback address ([http://127.0.0.1:3264/api](http://127.0.0.1:3264/api)), the UI immediately displays a red error message: "Blocked private or local provider URL".
[...]

PR #3963, opened 2026-06-16
urgent: gate proxy direct fallback behind a feature flag

rdself:
## Summary

- Adds `OMNIROUTE_CONTROL_PLANE_PROXY_DIRECT_FALLBACK` to Dashboard -> Settings -> Feature Flags and `.env.example`.
- Requires both the existing code opt-in and the new flag before `runWithProxyContextOrDirect()` can clear proxy context and use direct egress.
- Keeps the default pinned-proxy behavior fail-closed; direct fallback is now explicit opt-in.
- Updates tests and docs for the new flag and env contract.

## Why this is urgent

PR #3906 changed control-plane proxy handling so an unreachable pinned proxy can silently become direct egress. That is an unsafe default for deployments that pin provider accounts to specific egress IPs.

@zhiru This default is unacceptable. A control-plane convenience fix must never silently bypass a user-pinned proxy. On datacenter-hosted OmniRoute instances, this can route Claude account traffic through a completely different egress profile and directly lead to account enforcement or bans. Proxy-path changes cannot optimize one local failure mode by overriding operator intent. You cannot write proxy-handling code only for your own deployment needs while ignoring the risk profile of other users. Any fallback that changes egress identity must be explicit, default-off, documented, and tested before it lands.

@diegosouzapw Could we add a stricter review checkpoint for proxy-path changes, especially changes that can alter egress identity? This should require an explicit default-off feature flag, docs, and tests before merge. I am raising this because a previous proxy handling change already resulted in two Claude accounts being banned on my deployment, so I would really appreciate more caution around this class of change.

## Validation

- `node --import tsx/esm --test tests/unit/feature-flags-settings.test.ts tests/unit/proxy-fetch.test.ts`
- `npm run typecheck:core`
- `npm run check:docs-all`
- `npx eslint open-sse/utils/proxyFetch.ts src/shared/constants/featureFlagDefinitions.ts src/shared/utils/featureFlags.ts tests/unit/feature-flags-settings.test.ts tests/unit/proxy-fetch.test.ts`

PR #13732, opened 2026-09-15
feat(api): route /v1/rerank to remote provider nodes behind RERANK_REMOTE_PROVIDER_NODES

seanford:
## Summary

`POST /v1/rerank` only dispatched to provider nodes whose base URL hostname was `localhost`, `127.0.0.1`, or `172.16.0.0/12` — a filter hardcoded in the route. A rerank node on any other host (a LAN box or Tailscale peer running TEI, Infinity, vLLM, llama.cpp, …) was silently dropped and the request fell through to `Invalid rerank model`, even though the same node served `/v1/embeddings` without complaint and had already passed the provider outbound URL policy when it was created. The memory engine's rerank step calls this route over loopback, so `rerankProviderModel` could not reach such a node either.

This mirrors the audio routes (#3963): loopback nodes stay always-eligible and unchanged; remote nodes are opt-in via a new `RERANK_REMOTE_PROVIDER_NODES` feature flag (default off — routing to a remote host changes egress identity) **and** must pass the provider outbound URL policy (`getProviderOutboundGuard()`, #5066 / #9123), so cloud-metadata hosts are never routed to and strict `public-only` deployments never route to private hosts.

## Changes

- `src/shared/network/loopbackNodeHost.ts` — one pure definition of the loopback host set, replacing three copies (rerank route, `audioRegistry.isLoopbackNodeHost`, `localHealthCheck`). The shared version also rejects `user@host` URLs, which the audio copy did not.
- `src/shared/network/providerNodeHost.ts` — policy-aware remote-node eligibility mirroring `guardProviderNodeBaseUrl()` (`none` / `block-metadata` / `public-only`).
- `src/app/api/v1/_shared/rerankProviderNodes.ts` — pure, testable selection step + loader, modelled on `audioProviderNodes.ts`; `rerank/route.ts` uses it.
- Feature-flag definition (category network, default `false`, `warningLevel: danger`); `FEATURE_FLAGS.md` (56 flags) / `ENVIRONMENT.md` / `.env.example`; `API_REFERENCE.md` and `MEMORY.md` notes; `changelog.d/features/0000-…` fragment (placeholder number — will rename once assigned).

## Validation

- Change type: other (API routing / network policy)
- `node --import tsx/esm --test tests/unit/rerank-remote-provider-nodes.test.ts tests/unit/feature-flags-settings.test.ts tests/unit/feature-flags-doc-sync-static.test.ts tests/unit/audio-provider-nodes-selection.test.ts` — all passing
- `npm run typecheck:core` clean; `eslint --suppressions-location config/quality/eslint-suppressions.json` + `prettier --check` clean on touched files
- Exercised end-to-end on a self-hosted deployment (built from this branch stacked with #13704): `/v1/rerank` routed through a LAN `openai-compatible` rerank node and the memory engine's `rerankProviderModel` pointed at the same node.
- Reconciled with `release/v3.8.51` @ `c0f92ec98`

## Tests added or updated

- `tests/unit/rerank-remote-provider-nodes.test.ts` (new) — host classification, the three policy modes, the selection step, and the route end-to-end (flag off → 400 without contacting the node; flag on → forwarded to `<base>/v1/rerank` with the node credential; flag on + strict policy → still excluded).
- `tests/unit/feature-flags-settings.test.ts` — expected flag count 55 → 56.

## Reviewer notes

- Behaviour change is opt-in only; default installs are byte-for-byte equivalent for rerank routing.
- Audio: `isLoopbackNodeHost` now rejects embedded-credential URLs (slight tightening, security-positive).
- Follow-ups: the TEI/Infinity shape adapters for the same route (stacked on this branch) and `/v1/models` typing of node rows by `apiType` (independent).
- Requires #13704 to build an image from `release/v3.8.51` at all.

https://claude.ai/code/session_01KHaWGXax8LT4TxyGopEVxX


cq-guojia:
Hitting this same filter, but from a use case outside the one described here — it may be worth widening the motivation.

**Setup**: a custom `provider_node` (OpenAI-compatible, created from the dashboard) with `baseUrl = https://api.siliconflow.cn/v1`. Same node, two endpoints:

- `POST /v1/embeddings`, model `sf/BAAI/bge-m3` → works. The server logs `EMBED Resolved custom embedding provider: sf -> https://api.siliconflow.cn/v1/embeddings`.
- `POST /v1/rerank`, model `sfc/BAAI/bge-reranker-v2-m3` → `400 Invalid rerank model: ...`. The node appears in no log line at all, because the hostname filter drops it before the `.find()` ever runs.

Ticking the `Rerank` box on that node in the dashboard changes nothing, since the route never reads `apiType`/`endpoints` — only the hostname.

**Why a node and not the registry path for rerank**: the built-in `siliconflow` entry in `rerankRegistry.ts` hardcodes `https://api.siliconflow.com/v1/rerank`, and no connection-level `baseUrl` is consulted. We need the `.cn` regional endpoint — the API key is issued by the `.cn` account, and measured from our host `.cn` answers in ~0.17s vs ~1.26s for `.com`. So the registry path cannot be pointed at the region we need, and the node path is the only way to specify a URL — which is exactly the path this filter closes.

**Current workaround, and why the affected set is broader than LAN/Tailscale**: we had to deploy an extra container — an nginx reverse proxy holding a static `172.21.x` address that forwards to `api.siliconflow.cn` — *solely* to satisfy the hostname filter. A whole deployment exists just to get past a filter. That is precisely the class of workaround `RERANK_REMOTE_PROVIDER_NODES` removes.

So this is not limited to "a LAN box or Tailscale peer running TEI/Infinity/vLLM". It also covers **any public SaaS endpoint that has to be reached through a provider node** — for example a regional endpoint of a gateway that is already in the hardcoded registry, under a different region.

Full trace filed separately: #13932.

Two smaller things noticed while tracing this (independent of the flag):

1. The `.filter()` drops nodes silently — one `console.warn` naming the dropped hostname would have saved hours of digging. `Invalid rerank model` points the reader at the model string, which is not where the problem is. Related direction: #13687.
2. `buildDynamicRerankProvider` hardcodes `authType: "apikey"` / `authHeader: "bearer"`, so rerank has no no-auth path for local nodes — asymmetric with embeddings after #6925. Not blocking us, just noting it.

diegosouzapw:
Independent confirmation for this PR, from triage: @cq-guojia hit exactly this in the wild and filed #13932 two days after you opened it — a rerank provider node on a non-loopback host silently dropped, surfacing as the misleading `Invalid rerank model` error, while the same node served `/v1/embeddings` fine.

Triage verified the root cause against the tip and reached the same conclusion you did, including that the hostname filter at `src/app/api/v1/rerank/route.ts:82-95` is **not** a deliberate security guard: `src/server/authz/routeGuard.ts` has no `rerank` entry, so the route was never loopback-enforced under Hard Rules #15/#17, and `git blame` traces the line to a CodeQL-cleanup commit rather than to a policy decision. Your framing — keep loopback always-eligible, put remote behind `RERANK_REMOTE_PROVIDER_NODES` default-off, and still require `getProviderOutboundGuard()` — is the right shape, and mirroring the audio precedent (#3963) is the reason this is being adopted over writing a separate fix.

**This PR is the adopted fix for #13932.** Two housekeeping notes so it can land:

1. It has 5 failing checks (`API Route Typecheck`, `Fast Quality Gates`, `Unit Tests fast-path 1/4, 2/4, 4/4`), identical on #13733. The base has moved a long way since 2026-09-15 and there is no open base-red issue for `release/v3.8.51`, so these are most likely a stale base rather than defects in your diff — a merge of the current `release/v3.8.51` into the branch is the first thing to try.
2. Neither this PR nor #13733 references the issue. Adding `Closes #13932` to this PR's body would let it close on merge and keep the trail intact.

Thanks for the loopback-host deduplication in here too — three copies of that host set was a bug waiting to happen.

diegosouzapw:
Thanks for this — the loopback/remote split and reusing getProviderOutboundGuard() here is exactly the right shape, mirroring the audio-node precedent closely. Since #13733 is stacked on top of this exact commit (TEI/Infinity shape adapters), we're going to merge that one instead so we get both fixes in one pass — it carries your #13732 commit unchanged, so nothing here is lost. Closing this one as covered by #13733; huge thanks for both.

PR #13738, opened 2026-09-15
fix(gemini): a tool name starting with a digit no longer fails the request

### LINKED ISSUE #13715: fix(providers): a tool name starting with a digit rejects the whole Gemini request (400 Invalid function name)
## Summary

A single tool whose `function.name` begins with a digit makes Google reject the entire request:

```
400 GenerateContentRequest.tools[0].function_declarations[0].name: Invalid function name.
Must start with a letter or an underscore. Must be alphameric (a-z, A-Z, 0-9),
underscores (_), dots (.), colons (:), or dashes (-), with a maximum length of 128.
```

`normalizeGeminiToolName` (`open-sse/translator/helpers/geminiToolsSanitizer.ts`) removes invalid
characters, collapses underscores, strips leading/trailing underscores and caps the length — but it
does not guarantee that the first character is a letter. `1c_ssl_mcp_plugin_reload` therefore passes
through normalized-but-unchanged, `buildChangedToolNameMap` records nothing, and the name is
forwarded verbatim.

Observed with real agent traffic: a 109-tool catalog from MCP servers of a 1C platform
distribution, 6 of which are named `1c_*`. Every agent request through `antigravity/*` failed with
400 while those servers were attached. Reducing `max_tokens` or removing messages did not help —
only removing the tools did.

## Environment

- OmniRoute 3.8.50, official Docker image, macOS host, Docker
- Provider: `antigravity`
- Client protocol: OpenAI-chat (`/v1/chat/completions`), stream enabled
- Client: agentic client sending MCP tools

## Reproduction

Minimal — a single tool whose name starts with a digit:

```bash
curl -s -X POST http://localhost:20128/v1/chat/completions \
  -d '{"model":"antigravity/gemini-3.8-flash-tiered","max_tokens":32,
       "messages":[{"role":"user","content":"Say OK"}],
       "tools":[{"type":"function","function":{"name":"1c_probe","description":"probe",
                 "parameters":{"type":"object","properties":{}}}}]}'
```

Renaming the same tool to `c1_probe` returns 200. Nothing else in the payload changes.

Measured matrix — same payload, only `function.name` differs, two runs per row:

| `function.name` | Antigravity (Google) | codex (OpenAI subscription) | DeepSeek |
| --- | --- | --- | --- |
| `1c_probe` | **400** | 200 | 200 |
| `c1_probe` | 200 | — | — |
| `_1c_probe` | **400** | — | — |
| `_abc` | 200 | — | — |
| `ssl.probe`, `ssl-probe` | 200 | — | — |
| 100-character name | 200 | — | — |

Note the `_1c_probe` row: the sanitizer strips leading underscores, so `_1c_probe` is normalized to
`1c_probe` and rejected for the same reason. A leading underscore is therefore **not** a workaround.
Google's own message says "a letter or an underscore", but because the underscore is stripped first,
in practice the name must start with a letter.

## Why the whole request dies

Cloud Code rejects the entire `GenerateContentRequest` when any single
`function_declarations[].name` is invalid. With 6 bad names out of 109 tools, every request fails —
including requests that would never call those tools.

## Impact

Any MCP server whose tool names begin with a digit, which is common for vendor-prefixed tool sets
(the 1C platform plugins above use `1c_<plugin>_<action>`). The same client and the same tool set
work unchanged on OpenAI-compatible providers, so the failure looks provider-specific and is easy
to misattribute.

## Proposed fix

Guarantee a letter as the first character in `normalizeGeminiToolName`:

```ts
  return namespaceStripped
    .replace(/[^a-zA-Z0-9_]/g, "_")
    .replace(/_+/g, "_")
    .replace(/^_+|_+$/g, "")
    .replace(/^(\d)/, "t$1");     // Gemini requires the name to start with a letter
```

The reverse direction needs no changes: the name now differs from the client's, so
`buildChangedToolNameMap` records it, `_toolNameMap` carries it to the executor, and the existing
response/history restore path maps the model's `functionCall` name back to the client's original
spelling.

Verified with this change applied:

- the request carrying the full 109-tool catalog returns 200;
- the model's tool call comes back named `1c_ssl_mcp_plugin_reload` — the client's spelling, so the
  client dispatches it correctly;
- multi-turn history stays consistent, because the same `sanitizeToolName` closure aliases the
  historical `tool_calls` names within a request.

Applying the prefix inside `normalizeGeminiToolName` is safe with respect to collisions: it runs
before the length cap and before the `isSanitizedNameTaken` check, so an artificial collision would
still be resolved by the existing hashed-name path.

## Related

- #4789 — a neighbouring case in the same sanitizer: nameless hosted tools produced this exact
  Google 400 and killed the whole request.
- Not #13057 / #13477 — that failure is schema traversal (`properties.required`), a different cause
  behind the same opaque `[400]: Antigravity upstream error (400)` wrapper.

L4XB:
## Summary

A single tool whose `function.name` begins with a digit made Google reject the entire `GenerateContentRequest`, so six `1c_*` tools in a 109-tool MCP catalog failed every request through `antigravity/*`, including requests that would never call them. The sanitizer now guarantees the declared name starts with a letter, and the client keeps seeing its own spelling.

`normalizeGeminiToolName` removed invalid characters, collapsed underscores and stripped leading and trailing ones, none of which touches a leading digit. That strip is also why a leading underscore was not a workaround: the client's `_1c_probe` was normalized back to `1c_probe` and rejected for the same reason, even though Google's own message names the underscore as allowed.

Reproduced against the current base before changing anything:

```
sanitizeGeminiToolName("1c_probe")   -> "1c_probe"    (forwarded verbatim, 400)
sanitizeGeminiToolName("_1c_probe")  -> "1c_probe"    (the underscore is stripped first)
```

## Related Issues

- Closes #13715
- Related to #4789, a neighbouring case in the same sanitizer where nameless hosted tools produced this exact Google 400

## Where the prefix goes, and why it matters

Inside `normalizeGeminiToolName` rather than at the call site, because that keeps the guarantee on the one value every path reads:

- `buildHashedGeminiToolName` builds its name from the normalized string and inherits its first character, so a fix applied later would hold for short names and fail silently for the long ones a 109-tool catalog is full of. There is a cell for that.
- It runs before `isSanitizedNameTaken`, so two names that newly collide, `_1c_probe` and `1c_probe` for instance, are still separated by the existing hashed path instead of one overwriting the other's mapping. There is a cell for that too.

The reverse direction needs no change: the sanitized name now differs from the client's, so `buildChangedToolNameMap` carries the original and the response translator restores the client's spelling on the model's `functionCall`.

## Validation

- [x] Change type: provider
- [x] `npm run lint` (`eslint` on both touched files, clean)
- [x] Reconciled with the current active release base (`release/v3.8.51` at `1d03ba0`); focused checks rerun afterward

`prettier --check` clean on both files. The pre-commit chain ran and passed: prettier, eslint, docs-sync, the explicit-any budget, and the tracked-artifacts check.

One neighbouring suite, `tests/unit/9568-gemini-tool-casing-mismatch.test.ts`, fails on my machine. It fails identically on a clean `release/v3.8.51` with this branch stashed, so it is not this change; I would rather report that than leave a green claim I did not get.

## Tests Added Or Updated

- `tests/unit/13715-gemini-tool-name-leading-digit.test.ts` (new, 6 cells)

| cell | what it stops |
| --- | --- |
| a declared name never starts with a digit | the reported defect, asserted across `buildGeminiTools` on a four-tool catalog |
| a leading underscore is not a workaround | the measured `_1c_probe` row from the report, which the strip turned back into the rejected name |
| the client's own spelling comes back on the reverse map | a rename that fixes the request and breaks the caller's dispatch |
| a name that already starts with a letter is untouched | prefixing unconditionally, which would rename every tool in every catalog for nothing |
| an over-long digit-first name keeps the guarantee | a fix applied at the call site instead of in the normalizer, which the hash path would undo |
| two colliding digit-first names stay distinct | the prefix running after the collision check |

Mutation results on the new line, 4 of 4 caught:

| mutation | result |
| --- | --- |
| the prefix is gone (the reported defect) | **caught** |
| the prefix applies to every name, not only a digit first | **caught** |
| the anchor is lost, so any digit anywhere is prefixed | **caught** |
| the digit is replaced instead of prefixed | **caught** |

## Coverage Notes

`open-sse/translator/helpers/geminiToolsSanitizer.ts` is the only production file touched, one line plus its comment. The new suite drives it through the public `sanitizeGeminiToolName` and `buildGeminiTools` rather than the private normalizer, so it covers the change on the path a request actually takes, including the length-cap and collision branches that the same string feeds.

## Reviewer Notes

The prefix character is `t`, which is what the issue proposes and what the reporter verified against the real 109-tool catalog. An underscore would also satisfy Google's grammar, and the report's `_abc` row shows a leading underscore returning 200, but that spelling was never sent to Google in the measured matrix, so I took the one with evidence behind it rather than the one I found tidier.

Names that gain the prefix are renamed on the wire, which is visible in provider logs. That is already true for every name the sanitizer rewrites today, and the reverse map means clients do not see it.


diegosouzapw:
Thanks @L4XB — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #13734, opened 2026-09-15
fix(api): type compatible-provider-node models in /v1/models by the node's apiType

seanford:
## Summary

Model rows discovered from an OpenAI-compatible provider node rarely carry endpoint metadata — a TEI / Infinity / vLLM `/v1/models` listing is just ids — and the catalog defaulted such rows to `["chat"]`. An `embeddings`-typed node exposing `bge-m3` and a `rerank`-typed node exposing `bge-reranker-v2-m3` therefore both surfaced in `GET /v1/models` as untyped chat models: clients that build their pickers from `type: "embedding"` / `type: "rerank"` never saw them, and chat pickers listed models that 400 on chat.

## Changes

- `src/shared/constants/modelSupportedEndpoints.ts` — `defaultEndpointsForProviderNodeApiType(apiType)` (`embeddings` → `["embeddings"]`, `rerank` → `["rerank"]`, `audio-*` → themselves, `images-generations` → `["images"]`, everything else / unknown → `["chat"]`).
- `src/app/api/v1/models/catalog.ts` — node-id → `apiType` map; the synced and custom loops fall back to the node's modality when a row has no explicit `supportedEndpoints`; the custom-overlay merge path classifies `type`/`subtype` from the overlay's explicit `supportedEndpoints`.
- `API_REFERENCE.md` discovery note + `changelog.d/fixes/0000-…` fragment (placeholder number).

Explicit `supportedEndpoints` on any row still take precedence; chat/responses nodes keep the historical behaviour (no `type`).

## Validation

- Change type: bugfix
- `node --import tsx/esm --test tests/unit/catalog-provider-node-apitype-endpoints.test.ts tests/unit/models-catalog-route.test.ts` — all passing (the GLM-5.2 alias test in the latter is timing-sensitive under `--test` concurrency on a 2-core box and passes serially and in isolation; unrelated to this change)
- `npm run typecheck:core` clean; eslint (repo suppressions) + prettier clean on touched files
- Exercised end-to-end on a self-hosted deployment: `/v1/models` now reports `type: "embedding"` / `type: "rerank"` for the models of an `embeddings` / `rerank` compatible node.
- Reconciled with `release/v3.8.51` @ `c0f92ec98`

## Tests added or updated

- `tests/unit/catalog-provider-node-apitype-endpoints.test.ts` (new) — the helper, plus `getUnifiedModelsResponse()` for embeddings, rerank, mixed (explicit wins), chat, and manual-overlay cases.

## Reviewer notes

[...]

PR #13740, opened 2026-09-15
fix(memory): list provider-node models in the Embedding and Rerank selectors

seanford:
## Summary

The memory **Embedding** and **Rerank** selectors (`GET /api/memory/embedding-providers`, `GET /api/memory/rerank-providers`) appended local provider nodes by `apiType` alone and always with an **empty model list**. A node typed `embeddings` that also serves a rerank model — one TEI / Infinity / vLLM box hosting both `bge-m3` and `bge-reranker-v2-m3` is the common self-hosted layout — never appeared in the Rerank dropdown at all, and in the Embedding dropdown it appeared as a provider with nothing to pick. Typing `prefix/model` by hand worked (the request path resolves it directly), only the convenience layer was blind.

## Change

- New `src/lib/memory/embedding/nodeModalityListings.ts`: builds each node's listing from its **synced + custom model rows**, typing every row the same way `GET /v1/models` does — an explicit `supportedEndpoints` wins, otherwise the node's `apiType` decides (`defaultEndpointsForProviderNodeApiType`, from #13734). A custom row overlays the synced row of the same id: its explicit `supportedEndpoints` re-types the model (so a manual `["rerank"]` override on a discovered row wins over the node default); a custom row without one only contributes its name.
- A node is listed for a modality when any of its rows is typed for it, or when its `apiType` is `chat`/`responses`/the modality itself (the historical implicit behaviour, kept so generic nodes without a catalog still appear).
- `listEmbeddingProviders()` and the rerank-providers route both consume the shared builder; the old per-route `apiType`-only loops are gone.
- Best-effort: any storage failure yields `[]`, so the curated providers are still returned.

## Tests

`tests/unit/memory-node-modality-listings.test.ts` — 5 pure cases (implicit apiType listing, explicit endpoints, custom overlay re-typing, invalid prefixes skipped, prefix dedup) plus 2 end-to-end cases against the two routes with an `openai-compatible` node typed `embeddings` carrying synced `bge-m3`/`bge-reranker-v2-m3` and a custom `["rerank"]` overlay: the Rerank selector now lists `skilled-mini/bge-reranker-v2-m3`, the Embedding selector lists `skilled-mini/bge-m3`, and neither lists the other.

Verified on a local Docker build against a real TEI-style node (`bge-m3` + `bge-reranker-v2-m3`): both dropdowns now show the node with the right model, and the memory playground reports `rerankApplied: true` with the picked reranker.

Stacked on #13734 (uses its `defaultEndpointsForProviderNodeApiType`); the diff will shrink to the memory changes once that merges. Gates run locally in PR mode against `c0f92ec98`: file-size, complexity-ratchets, typecheck:core, eslint, prettier all clean; the remaining reds on this base (mutation-test-coverage `noauth-model-lockout`, the `db/providers/deletion.ts` cycle, 5 missing `.env.example` vars) reproduce on the bare base.

https://claude.ai/code/session_01KHaWGXax8LT4TxyGopEVxX


diegosouzapw:
Thanks @seanford — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #13743, opened 2026-09-15
fix(ci): register reset-aware-request-scope-12600 in stryker tap.testFiles

patrykkopycinski:
The `mutation-test-coverage` gate fails on the release line itself, with no PR diff involved — which is why it is currently red on every open PR against `release/v3.8.51`.

`tests/unit/reset-aware-request-scope-12600.test.ts` covers `open-sse/services/combo/quotaScoring.ts` but is not listed in `stryker.conf.json` `tap.testFiles`, so under `--strict` its mutant kills do not count and the gate reports drift.

**Reproducing the numbers**

```bash
# pristine worktree at the release tip, no PR diff
git worktree add /tmp/stryker-probe --detach origin/release/v3.8.51
cd /tmp/stryker-probe && ln -s <repo>/node_modules node_modules
node scripts/check/check-mutation-test-coverage.mjs --strict
# ✗ 1 covering unit test(s) across 1 module(s) are missing from stryker.conf.json tap.testFiles
#   open-sse/services/combo/quotaScoring.ts
#     + tests/unit/reset-aware-request-scope-12600.test.ts

# after this one-line addition
node scripts/check/check-mutation-test-coverage.mjs --strict
# ✓ No drift — every covering unit test is listed in tap.testFiles.   (exit 0)

git worktree remove /tmp/stryker-probe
```

Independent evidence that this is base debt rather than PR-authored: the base branch's own quality run [`34970965604`](https://github.com/diegosouzapw/OmniRoute/actions/runs/34970965604) fails the identical gate.

One line in `tap.testFiles` (sorted position) plus a maintenance changelog fragment. JSON validated; `check-changelog-integrity.mjs` passes for the new fragment (the pre-existing `reset-aware-model-family.md` failure from #12637 is untouched base debt).


abhisheksharma2411 (commented):
The diagnosis is right and the mechanism is real — but I think this specific instance has already landed, and the gate it's meant to fix is red for a different reason.

**`reset-aware-request-scope-12600.test.ts` is already in `tap.testFiles`** on the current release tip (`df87e936`), at `stryker.conf.json:372` — one line *above* the `resilience-connections.test.ts` anchor this diff inserts after:

```
372:      "tests/unit/reset-aware-request-scope-12600.test.ts",
373:      "tests/unit/resilience-connections.test.ts",
374:      "tests/unit/responses-handler.test.ts",
```

It came in with `25bc16d87e`. So applying this diff produces **two** identical entries rather than one.

**And the gate is still red on that tip — for six other files.** I ran your own reproduction, on a pristine checkout of `origin/release/v3.8.51` with no PR diff:

```
Scanned 5127 unit test file(s) against 31 mutated module(s).

  open-sse/services/accountFallback.ts
    + tests/unit/daily-reset-tz-threading.test.ts
    + tests/unit/noauth-model-lockout.test.ts
  src/sse/services/auth.ts
    + tests/unit/free-badge-provider-gate.test.ts
    + tests/unit/noauth-model-lockout.test.ts
  open-sse/services/combo/comboPredicates.ts
    + tests/unit/local-token-budget-429-skips-cooldown.test.ts
  open-sse/services/combo/rrState.ts
    + tests/unit/daily-reset-tz-threading.test.ts

✗ 6 covering unit test(s) across 4 module(s) are missing from stryker.conf.json tap.testFiles.
```

`quotaScoring.ts` isn't among the four, which matches it already being registered. So merging this wouldn't turn the gate green — the red you're seeing on every open PR is those six.

**None of that makes the PR wrong about the problem**, and the writeup is the most useful part of it: the reproduction recipe is exactly right, and "the gate fails on the release line itself with no PR diff involved" is the observation that matters. It just looks like the base moved under it — `quotaScoring.ts` got fixed by someone else while this was open.

Worth rebasing onto `df87e936` and re-running the checker: the same one-line-per-entry fix applied to those six (four are distinct files — `daily-reset-tz-threading` and `noauth-model-lockout` each cover two modules) would actually clear it. Happy to review that version.

One thing I'd keep either way: the changelog fragment's wording explains *why* registration matters (`so its mutant kills count`), which is the part someone hitting this in six months will need.

Not a blocker on my side — I can't approve — but I'd hold merging this shape, since a duplicate entry plus a still-red gate is worse than the current state.

patrykkopycinski:
Closing as superseded — you were right on both counts, and the base has since moved past it entirely.

Verified on the current tip (`d6a0460bd`):

- `reset-aware-request-scope-12600.test.ts` is already registered (came in with `25bc16d87`), so this diff would indeed produce a duplicate entry.
- All six files you listed are now registered — landed in `8f55d85d2` (`fix(ci): clear the release/v3.8.51 base-reds left by the 09-15 batch (stryker, ...)`, #13747).
- Re-ran the checker on a pristine `d6a0460bd` checkout: `✓ No drift — every covering unit test is listed in tap.testFiles.`

The one thing worth keeping from this PR was the reproduction recipe in the body — already reused in #13747's wake. Thanks for the precise review; nothing left to merge here.

PR #13752, opened 2026-09-15
fix(gemini): prefix digit-led tool names so Gemini accepts the request

### LINKED ISSUE #13715: fix(providers): a tool name starting with a digit rejects the whole Gemini request (400 Invalid function name)

xiechimon:
## Summary

- `normalizeGeminiToolName` sanitized characters, collapsed underscores and trimmed the edges, but never guaranteed the **first** character was a letter or an underscore — the one rule Gemini enforces on function names. A name like `1c_ssl_mcp_plugin_reload` passed through normalized-but-unchanged, `buildChangedToolNameMap` recorded nothing, and the name went upstream verbatim.
- Gemini then rejects the **entire** request with `400 Invalid function name. Must start with a letter or an underscore`, no matter how healthy the rest of the catalog is. Removing the offending tools was the only client-side workaround.
- Digit-led names now get a `_` prefix, applied before the existing 64-char length guard so a digit-led name already at the limit still hashes down to fit. The prefix is part of normalization, so the `toolNameMap` round-trip stays one-to-one and the name stays recognizable.

## Related Issues

- Closes #13715

## Validation

- [x] Change type: provider / routing (`open-sse/translator/helpers/`)
- [x] Focused tests and category gates from the golden path — new regression test plus the five suites that already cover this helper
- [x] `npm run lint` — clean on both changed files; pre-commit ran `prettier --write` + `eslint --fix` on them
- [x] Reconciled with the current active release base; focused checks rerun afterward — branched from `release/v3.8.51` tip `7cabac49`

Verification actually run:

- `tests/unit/13715-gemini-tool-name-leading-digit.test.ts` — **fails 6/7 on the base** (`actual: '1c_ssl_mcp_plugin_reload'`), **passes 7/7 with the fix**. Red before green, no Gemini key needed.
- `tests/unit/gemini-tool-params-object-3357.test.ts`, `9617-gemini-uniqueitems.test.ts`, `12509-gemini-prefixitems.test.ts`, `gemini-malformed-required-and-bare-map-12269.test.ts`, `antigravity-retry-hook.test.ts` — 39/39 pass, no regression.
- `prettier --check` clean on both files.

## Tests Added Or Updated

- `tests/unit/13715-gemini-tool-name-leading-digit.test.ts` (new): asserts the sanitized name matches `/^[a-zA-Z_][a-zA-Z0-9_]*$/` for the reported shape (`1c_…`), for an all-digit name, and for a digit-led name that also needs character replacement; that already-valid names are untouched; that repeated calls reuse one `toolNameMap` entry and map back to the original; that the prefix keeps a name at the limit inside 64 chars; and that `buildGeminiTools` accepts a digit-led declaration and records the reverse mapping.

## Coverage Notes

- Both changed paths are fully covered by the new file; the fix is a single conditional on an already-covered function. No `src/`, `electron/`, or `bin/` change.

## Reviewer Notes

- **Alternative considered and rejected**: dropping the leading digit (`1c_x` → `c_x`). `_` preserves the name and cannot collide with an existing name that already started with a letter. It also matches Gemini's own allowance — it accepts a leading underscore, it just does not accept a leading digit.
- The `_` prefix is applied after the `.replace(/^_+|_+$/g, "")` trim, so `_1c_x` and `1c_x` both land on `_1c_x`.
- Scope limited to this one function: the digit rule is the only Gemini name constraint still unenforced here; the character class, collapse, trim and 64-char cap were already handled.
- `sanitizeGeminiToolName` is shared by `openai-to-gemini`, `claude-to-gemini` and the antigravity executor, so the fix covers all three request paths.

Refs #13715


diegosouzapw:
Nice, clean fix — the underscore prefix reads very naturally against Gemini's own error message. #13738 opened a few hours before this and fixes the exact same #13715 root cause with equivalent test coverage plus one extra integration-level check, so we're merging that one and closing this as a duplicate. Thank you for catching and fixing this independently — happy to credit both PRs in the release notes.

Issue #10705, opened 2026-08-19
[BUG] The input token data statistics are incorrect.


kitety:
### OmniRoute Version

v3.8.49

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

linux ubuntu

### Node.js Version

24

### Provider(s) Involved

Third-party API mode

### Client Tool

cline

### Description

The statistics from the third-party API I'm using are incorrect compared to the upstream statistics.
Our statistics

<img width="2277" height="210" alt="Image" src="https://github.com/user-attachments/assets/c3361df0-d8b9-467e-941d-7699a888c17d" />
Upstream statistics
<img width="851" height="313" alt="Image" src="https://github.com/user-attachments/assets/b5957d75-8244-4e57-9cc1-6d4cd17ed176" />
The input token cannot be 0. I have closed Prompt Compression

### Steps to Reproduce

1. Disable Prompt Compression

2. Test a model from the model provider

3. Filter the usage statistics to find the request data for the current time period

4. Compare this data with the upstream data statistics from a third-party platform.

### Expected Behavior

Our data statistics are just as accurate as those of third-party data statistics, and they are also usable when compression is enabled.

### Actual Behavior

Our data statistics are inconsistent with those of third parties, with significant discrepancies.

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

There is already a picture above.

kitety:
maybe is this error, i got provider response as this
```json
{
  "_streamed": true,
  "_format": "sse-json",
  "_stage": "provider_response",
  "_eventCount": 13,
  "summary": {
    "id": "msg_18cd12e6ba422487",
    "type": "message",
    "role": "assistant",
    "model": "claude-fable-5",
    "content": [
      {
        "type": "text",
        "text": "\n\n你好！继续测试也行 😄\n\n有代码问题随时叫我！"
      }
    ],
    "stop_reason": "end_turn",
    "usage": {
      "input_tokens": 0,
      "output_tokens": 43
    }
  }
}
```
```
 "input_tokens": 0,
```
However, in this case, OmniRoute shouldn't just use this field.

OmniRoute itself uses a tokenizer to estimate the pre-send request body.

kitety:
When will the new version be released? Thanks

kitety:
```json
{
  "_streamed": true,
  "_format": "sse-json",
  "_stage": "client_response",
  "_eventCount": 145,
  "summary": {
[...]

PR #13760, opened 2026-09-15
fix(usage): preserve nested prompt cache reads before cost calculation

### LINKED ISSUE #13746: fix(backend): normalization drops nested cache-read counters before cost calculation
### OmniRoute Version

`release/v3.8.51`, reproduced on upstream commit `831b485e0894064bf0d805eb6be32b18328f3676`.

### Installation Method

Source checkout, using the repository's TypeScript loader and isolated test data directory.

### Operating System

macOS

### Node.js Version

v26.8.2

### Provider(s) Involved

Synthetic OpenAI Chat Completions and Responses usage shapes. The reproduction does not contact a provider or require credentials.

### Description

`normalizeUsage()` discards cache-read counters when they are nested under `prompt_tokens_details.cached_tokens` or `input_tokens_details.cached_tokens`. Passing the normalized result to `computeCostFromPricing()` therefore charges those cached tokens at the ordinary input price.

The equivalent flat `cached_tokens` shape is priced correctly. This is a field-preservation gap, not a model-pricing-table problem.

There is an existing production consumer of this normalization/cost boundary in `open-sse/utils/stream.ts`: the final SSE metadata path calls `calculateCost(provider, model, normalizeTokenUsage(finalUsage))`. The non-streaming cost path in `chatCore.ts` also normalizes its selected response usage. This report is specifically about **nested-only usage reaching that boundary**; it does not claim that every request is affected. Paths that first extract/flatten the cache counter can avoid the defect.

### Steps to Reproduce

From a source checkout with dependencies installed, run the following. It uses the repository's temporary DATA_DIR setup, not the operator's live database:

```bash
DISABLE_SQLITE_AUTO_BACKUP=true node \
  --import tsx/esm \
  --input-type=module <<'JS'
import { normalizeUsage } from './open-sse/utils/usageTracking.ts';
import { computeCostFromPricing } from './src/lib/usage/costCalculator.ts';
import { resetDbInstance } from './src/lib/db/core.ts';

// Synthetic USD prices per million tokens, not a provider price claim.
const pricing = { input: 10, cached: 1, output: 50 };
for (const [shape, usage] of Object.entries({
  chat: {
    prompt_tokens: 100000, completion_tokens: 0,
    prompt_tokens_details: { cached_tokens: 99000 },
  },
  responses: {
    input_tokens: 100000, output_tokens: 0,
    input_tokens_details: { cached_tokens: 99000 },
  },
  flat: {
    prompt_tokens: 100000, completion_tokens: 0, cached_tokens: 99000,
  },
})) {
  const normalized = normalizeUsage(usage);
  console.log(JSON.stringify({
    shape, normalized, cost: computeCostFromPricing(pricing, normalized),
  }));
}
resetDbInstance();
JS
```

### Expected Behavior

All three representations preserve the same cache-read count and yield approximately **$0.109**:

- 1,000 uncached input tokens at $10/million;
- 99,000 cache-read tokens at $1/million;
- zero output tokens.

The normalized contract should remain a flat record of finite numeric counters. It should preserve the distinction between an absent cache field and an explicitly reported zero, and avoid counting the same cache usage twice when both flat and nested forms are present.

### Actual Behavior

Observed output on the unmodified upstream base (database initialization logs omitted):

```json
{"shape":"chat","normalized":{"prompt_tokens":100000,"completion_tokens":0},"cost":1}
{"shape":"responses","normalized":{"input_tokens":100000,"output_tokens":0},"cost":1}
{"shape":"flat","normalized":{"prompt_tokens":100000,"completion_tokens":0,"cached_tokens":99000},"cost":0.10899999999999999}
```

### Test Impact

Needs regression coverage through normalization and a real consuming stream/cost path, not only a calculator helper called with a casted raw object.

### Additional Context

Relevant prior work, checked before filing:

- #9977 added normalization before SSE cost metadata calculation. Its `normalizeUsage keeps only finite numeric fields for stream cost calculation` test currently expects nested `input_tokens_details.cached_tokens` to disappear.
- #11814 preserves nested **cache-creation/write** fields using `pickCacheCreationTokens()`. Cache-read normalization still only assigns the top-level `usage.cached_tokens`.
- #10535 / #10545 concern emitting standard cache fields in translated responses, rather than retaining those fields at this normalization boundary.
- #2522 fixes cache-creation double billing; this report concerns losing a cache-read counter before pricing.

A source-only reproduction is provided so no private account, deployment, traffic sample, or credentials are needed. Historical cost reconciliation is separate from this forward-path code fix.

### Validation Plan

- Demonstrate the new regression fails on the pinned upstream base and passes with the fix.
- Cover Chat Completions and Responses nesting, flat counters, explicit zero, missing/malformed/non-finite values, and cache-read plus cache-write combinations.
- Exercise a real stream/cost consumer and retain relevant extraction, pricing, exact-cost, and flat-rate regression coverage.

xiaoyaner0201:
## Summary

`normalizeUsage()` dropped prompt cache-read counters reported under `prompt_tokens_details.cached_tokens` / `input_tokens_details.cached_tokens`. Callers that price a request through this boundary — including the final SSE metadata path in `open-sse/utils/stream.ts`, which calls `calculateCost(provider, model, normalizeTokenUsage(finalUsage))` — then charged those cached tokens at the full input rate.

This normalizes both nested containers into the canonical `cached_tokens` counter, matching how `pickCacheCreationTokens()` already handles nested cache-**write** counters (#11814). The flat shape was already correct and is unchanged.

The normalized contract stays a flat record of finite numbers, and the top-level value keeps precedence via nullish coalescing, so an explicitly reported `0` is never overwritten by a nested value.

## Related Issues

- Closes #13746

## Validation

- [x] Change type: other (usage normalization feeding cost calculation)
- [x] `npm run lint` (changed files)

Synthetic prices per million (`input: 10`, `cached: 1`, `output: 50`) with `100,000` input tokens of which `99,000` are cache reads:

| Usage shape | Before | After |
| --- | --- | --- |
| `prompt_tokens_details.cached_tokens` | `$1.000000` | `$0.109` |
| `input_tokens_details.cached_tokens` | `$1.000000` | `$0.109` |
| flat `cached_tokens` | `$0.109` | `$0.109` (unchanged) |

Commands run on this branch (Node 24):

```bash
node --import tsx/esm --import ./open-sse/utils/setupPolyfill.ts --test --test-force-exit \
  tests/unit/cache-read-cost-13746.test.ts
npm run typecheck:core
npm run check:open-sse-typecheck
```

The regression file fails on the unmodified base with `cached_tokens: undefined` for both nested shapes, and passes with this change. Related existing suites (usage extraction, cost calculation, exact-cost, flat-rate, reasoning, streaming cost, response metadata) were rerun together and stayed green.

## Tests Added Or Updated

- `tests/unit/cache-read-cost-13746.test.ts` (new) — Chat and Responses nesting, flat counters, explicit zero, omitted and non-finite values, cache-creation compatibility, and one end-to-end `createSSEStream()` Responses case asserting the emitted `: x-omniroute-response-cost=0.1090000000` metadata line. Pricing and usage are synthetic; the test uses its own temporary `DATA_DIR` and calls `resetDbInstance()` on cleanup.
- `tests/unit/usage-extractor.test.ts` (updated) — the existing characterization case asserted that nested `input_tokens_details.cached_tokens` is discarded, which is the behavior being fixed; it now asserts the counter is preserved.

## Coverage Notes

`open-sse/utils/usageTracking.ts` is the only production file changed, and the new regression file covers that branch directly plus through a real stream consumer.

## Reviewer Notes

- No migration, no feature flag, no public API shape change.
- Cache-read precedence in `src/lib/usage/costCalculator.ts` and `src/lib/usage/tokenAccounting.ts` is intentionally left untouched; the defect is the counter being lost before pricing, so fixing the normalizer keeps the calculator contract (a flat record of finite numbers) intact.
- `hasValidUsage()` / `isEmptyUsage()` are unaffected: neither inspects `cached_tokens`.
- The implausible-usage repair path now clears nested cache-read and cache-write details for Claude and OpenAI Responses formats as well as their flat counters. Without this companion guard, a stale nested counter could be reintroduced by the new normalizer after #10705 repairs an impossible provider count.
- Scope is the forward pricing path. Reconciling rows already written to `domain_cost_history` is a separate concern and is not attempted here.


diegosouzapw:
Thanks @xiaoyaner0201 — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #13762, opened 2026-09-15
fix(admission): measure request bodies with the active cost budget

### LINKED ISSUE #13164: fix(backend): Admission cost measures the request body with the default 256 KiB early-exit, flattening every body above it
### OmniRoute Version

3.8.51

### Installation Method

Built from source

### Operating System

Windows

### Node.js Version

v24.16.0

### Description

`extractAdmissionCostFeatures` measures the request body with the default 256 KiB
early-exit:

```ts
// open-sse/services/admission/requestFeatures.ts:162
const bodyBytes = estimateSizeFast(body);
```

Every body above ~256 KiB therefore reports the same size, and since
`estimatedInputTokens` is derived from it (`ceil(bodyBytes / 4)`), both cost inputs
flatten at the same point. `DEFAULT_ADMISSION_COST_CONFIG` implies a much wider range —
`bodyBytesPerUnit: 16_384` × `maxRequestCost: 1_000` ≈ 16 MB — so roughly 98% of the
cost model's body range is unreachable.

**I am not sure this is a bug, and that is really what I am asking.** #9738 added the
`byteLimit` parameter for `truncateForLog` and said, explicitly:

> Added an optional `byteLimit` parameter (default unchanged, so
> `isSmallEnoughForSemanticCache`'s existing behavior is untouched)

and listed, under its test plan:

> No regressions in `estimateSizeFast`'s other callers
> (`tests/unit/adaptive-admission-features.test.ts`, `tests/unit/chatcore-memory-pressure.test.ts`)

So the admission caller was looked at and deliberately left on the default. What I
cannot tell from the outside is whether "leave it unchanged" meant "256 KiB is the right
ceiling for admission" or only "not in scope for that PR".

The reason I am asking rather than sending a patch is that raising it is not free, in
two concrete ways described under *Additional Context*.

### Steps to Reproduce

```ts
// bun probe.ts, from the repo root
import { estimateSizeFast } from "./open-sse/utils/estimateSize.ts";
import { extractAdmissionCostFeatures } from "./open-sse/services/admission/requestFeatures.ts";

const mk = (n, len) => ({
  messages: Array.from({ length: n }, () => ({ role: "user", content: "x".repeat(len) })),
});

for (const [label, body] of [
  ["~260 KB", mk(26, 10_000)],
  ["~2 MB",   mk(200, 10_000)],
  ["~8 MB",   mk(800, 10_000)],
]) {
  const f = extractAdmissionCostFeatures(body);
  console.log(label, "actual", JSON.stringify(body).length, "reported", f.bodyBytes, "tokens", f.estimatedInputTokens);
}
```

### Expected Behavior

A body that is 30× larger reports a larger size, and prices higher, up to the point
where more bytes can no longer buy cost units.

### Actual Behavior

```
~260 KB   actual   260768   reported   260104   tokens  65026
~2 MB     actual  2005814   reported   270108   tokens  67527
~8 MB     actual  8023214   reported   270108   tokens  67527
```

The 2 MB and 8 MB requests are indistinguishable — both are priced as ~270 KB. With the
default quanta, an 8 MB body contributes ~17 + ~67 ≈ 84 units of a possible 1000 instead
of ~640.

Passing the threshold explicitly gives the true values, so the measurement itself is
capable of it:

```
estimateSizeFast(body, 16_384 * 1_000)
  ~2 MB -> 2000800
  ~8 MB -> 8003200
```

### Test Impact

Needs a new unit test

### Additional Context

Two things make this a judgement call rather than an obvious fix, which is why I would
rather have your read first:

**1. An existing test depends on the current early-exit.**
`tests/unit/adaptive-admission-features.test.ts:146` pads the body with 300,000 bytes
specifically so the byte-exit fires before the walker reaches the tools proxy:

> `// Huge leading string makes estimateSizeFast byte-exit before walking tools,`
> `// so only countTools can touch the tools proxy — proving its scan bound alone.`

Raising the limit changes that setup, so the pad would have to grow past the new ceiling.
It uses the early-exit as scaffolding rather than asserting it is desirable, but it does
mean the change is not confined to one line.

**2. It costs work on the admission hot path.**
A higher ceiling means walking more of every large body on the path that exists to be
cheap. `ESTIMATE_SIZE_NODE_BUDGET` (16,384 visits) still bounds it, and the node-budget
fallback returns `byteLimit + 1` — so raising the limit also makes the fail-closed value
larger, i.e. more conservative. That direction seems right for admission control, but it
is a behaviour change worth naming.

Three directions I can see, happy to send a PR for whichever you prefer:

- **Leave it.** If 256 KiB is a deliberate ceiling for admission, a one-line comment
  at the call site would stop the next reader (me) from filing this again.
- **Derive the ceiling from the cost model** — `bodyBytesPerUnit * maxRequestCost`,
  self-contained in `requestFeatures.ts`, plus adjusting the tools-proxy test's pad.
- **Thread the real config** from `runtime.ts:373` into the extractor, which is the most
  faithful to what the `@param byteLimit` doc asks callers to do, and the largest change.

Related: `open-sse/utils/estimateSize.ts:83-88` is the doc that prompted this — it names
`chatCore/logTruncation.ts::truncateForLog` as the example to follow, and that sibling
does follow it (`logTruncation.ts:80-83`) with a comment explaining why. The admission
extractor is the one caller comparing against a much larger implied threshold that does
not.

lorenzozanee:
## Summary

Admission cost feature extraction stopped at the size estimator's default 256 KiB limit, flattening body and token estimates for larger requests. The extractor now derives its measurement limit from the active admission cost budget, including runtime overrides.

## Related Issues

- Closes #13164

## Validation

Vitest, the 60% coverage gate, and the production build all run in CI on this PR:

- [x] `npm run lint`

## Tests Added Or Updated

- `tests/unit/adaptive-admission-features.test.ts`

## Coverage Notes

- The changed `open-sse/` admission runtime and feature extraction are covered by the updated admission feature tests and the focused admission, cost, and estimator suite.
- No coverage report was run locally.

## Reviewer Notes

- Maintainers familiar with recent changes: Diego Rodrigues de Sa e Souza, Dizzle, and Markus Hartung.
- The patch preserves the estimator's node budget and semantic-cache default while applying the active admission cost profile.
- No migration or feature flag is involved; a changelog fragment is not present in this candidate.

Focused validation run locally: `node --import tsx/esm --test tests/unit/adaptive-admission-features.test.ts tests/unit/adaptive-admission-cost.test.ts tests/unit/estimateSizeFast.test.ts`, `npm run typecheck:core`, and ESLint on the changed files.

Supersedes #13291（原PR分支已删，按原提交重建）及关联issue说明
Closes #13164 (原PR关联issue，状态:open)


abhisheksharma2411 (commented):
Read this against `eabc6985`. The fix is right, and I think there's a stronger argument for it than the description makes — plus one thing I'd change in the test, which I verified rather than guessed.

**The derivation isn't arbitrary, it's exactly the saturation point.** `bodyBytesPerUnit * maxRequestCost` is precisely where extra bytes stop being able to move the answer: `estimateAdmissionCost` runs `unitsFrom(body, bodyBytesPerUnit)` through `addBounded(total, contribution, maxRequestCost)`, which caps at `maxRequestCost` and breaks. So measuring past that point *cannot* change the admission decision. That's a much better justification than "derive it from the budget", and it's worth putting in the comment — it tells the next reader why this limit and not some other one.

It's also conservative in the safe direction: body is only one of five contributions, so with tokens/messages/tools/fanout in play saturation arrives sooner. The limit is an upper bound on "bytes that could still matter".

**I checked the one thing that looked risky and it's fine.** `runtime.ts` drops `input.streaming === undefined ? undefined : {...}` and now always passes `streaming: input.streaming`. That's safe: `resolveStreaming` guards `context.streaming !== undefined` (`requestFeatures.ts:150`), so an explicit `undefined` still falls through to the body-derived value. Non-finding, but it's the kind of thing that silently flips a default, so I'd rather say I looked.

### The 17 MB fixture is avoidable — and this PR is what makes it avoidable

`sizePad` going `300_000` → `17_000_000` is a real consequence of the new limit: the test needs `estimateSizeFast` to byte-exit, and that now takes >16 MiB. But **this PR adds the exact injection point that removes the need** — the test can pass a tiny `cost` budget instead of allocating a huge string.

I tried it. The catch is that it has to go through **all four** `sizePad` call sites in that test, not just the first — my first two attempts failed on the gemini `functionDeclarations` proxy and then on the alias/layered cases, because those calls were still running at the 16 MiB default. With all four carrying it:

```ts
const sizePad = "x".repeat(4_000);
const tinyBudget = { cost: { bodyBytesPerUnit: 1_024, maxRequestCost: 1 } };
```

| | PR as-is | with the injected budget |
|---|---|---|
| tests | 6 pass | 6 pass |
| wall | 0.92 s | **0.57 s** |
| peak RSS | 100.4 MB | **95.5 MB** |

Same assertions, same property, no 17 MB allocation (34 MB as UTF-16). Worth doing while the reason for the number is still fresh — a year from now `17_000_000` reads as a magic number, whereas `bodyBytesPerUnit: 1_024, maxRequestCost: 1` states the intent.

### One question rather than a finding

The default measurement limit goes **256 KiB → 16 MiB, a 64× increase**, on the path that decides whether to admit a request at all. For a 20 MB body the estimator now walks 16 MB before the gate answers, where it previously walked 256 KB.

I don't think that's wrong — an admission decision made on a flattened estimate is worse than a slower accurate one, which is the whole point of #13164. But it is the gate spending more of the resource it exists to ration, and I couldn't find a measurement of the per-request cost at the new ceiling. Do you have one, or is the reasoning that bodies that large are rare enough that the tail cost doesn't matter? Either answer is fine; it just seemed like the thing a reviewer should ask before a 64× bound change lands on the hot path.

PR #13773, opened 2026-09-15
fix(db): resolve backup retention from persisted setting on health-check path

voidstackloop:
## Summary

Follow-up to #13341 / #13404, per [maintainer feedback on #13341](https://github.com/diegosouzapw/OmniRoute/pull/13341#issuecomment-3392222514).

#13404 fixed the missing prune call after health-check-repair backups (#13308), but only
resolved `maxFiles`/`retentionDays` from env vars. The persisted Storage-page setting —
already honored for manual/API/auto backups via `getDbBackupMaxFiles()` /
`getDbBackupRetentionDays()` in `backup.ts` — was silently ignored on the health-check
repair path, so an operator who set a stricter retention window in the UI (rather than via
env var) would still see repair snapshots pile up unbounded.

This extracts that `env override → persisted setting → default` precedence into a single
`resolveDbBackupRetention()` helper in `backupRetention.ts` (pure, no `core.ts` import, so no
circular dependency) and has both `backup.ts` and `core.ts`'s `createManagedDbBackup()`
delegate to it — removing the duplicated inline resolution logic in both places.

## Changes

- `src/lib/db/backupRetention.ts`: add `resolveDbBackupRetention(db, env)`, plus the
  exported `DB_BACKUP_SETTINGS_NAMESPACE`/`DB_BACKUP_MAX_FILES_KEY`/`DB_BACKUP_RETENTION_DAYS_KEY`
  constants it shares with `backup.ts`.
- `src/lib/db/backup.ts`: `getDbBackupMaxFiles()`/`getDbBackupRetentionDays()` now delegate
  to `resolveDbBackupRetention()` instead of duplicating the persisted-setting lookup.
- `src/lib/db/core.ts`: `createManagedDbBackup()`'s post-backup prune now calls
  `resolveDbBackupRetention(db)` instead of its env-only inline fallback.

## Test plan

- [x] New test `tests/unit/db-backup-retention-persisted-13308.test.ts`: asserts
  `resolveDbBackupRetention()` falls back to defaults with nothing set, honors a persisted
  Storage-page setting, and lets an env override win over the persisted setting.
- [x] Existing backup/retention suite still green: `db-backup-healthcheck-prune-13308`,
  `db-backup-autobackup-setting-5871`, `db-backup-export-streaming-9045`,
  `db-backup-extended`, `db-backup-tempdir-mkdtemp`, `db-backups-skills-3500`,
  `db-pre-migration-backup-retention-10421` (57 passed, 1 pre-existing skip).
- [x] `npm run typecheck:core` clean.
- [x] `npx eslint` clean on all touched files (`core.ts`'s 3 pre-existing unrelated errors
  confirmed present on `upstream/release/v3.8.51` before this change).


diegosouzapw:
Thanks @voidstackloop — merging via the release merge-train. Validated in local merge-train (merge-train-20260918-120911-suite.log) on the devbox @ train tip 8c305709478b7052fb981a75852bbf88fe3a959d with the sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 310/311 (the single red, hard-lease inventory, reproduces on the pure release tip) + vitest green (fast parity mode — full suite ran today on the tip via the base-red and 3b trains). Merged --admin per merge-gates §4/§7.

PR #13774, opened 2026-09-15
fix(admission): thread active cost config into body-size estimation (#13164)

### LINKED ISSUE #13164: fix(backend): Admission cost measures the request body with the default 256 KiB early-exit, flattening every body above it

diegosouzapw:
Closes #13164

## Root cause (short)

`extractAdmissionCostFeatures` called `estimateSizeFast(body)` with no explicit `byteLimit`, so it inherited the shared walker's default 256 KiB early-exit. The admission cost model implies a much larger usable range — `bodyBytesPerUnit (16_384) * maxRequestCost (1_000)` ≈ 16 MB — so both `bodyBytes` and the derived `estimatedInputTokens` saturated at 256 KiB regardless of actual body size. A 2 MB and an 8 MB request both reported `270108` and were priced identically.

## Fix

Thread the active cost config into the body-size estimation instead of relying on the 256 KiB module default:

- `open-sse/services/admission/requestFeatures.ts`: `AdmissionFeatureExtractionContext` gains an optional `cost?: Partial<AdmissionCostConfig>` field; `extractAdmissionCostFeatures` resolves it with `resolveCostConfig` and passes `estimateSizeFast(body, bodyBytesPerUnit * maxRequestCost)` instead of the bare call. A caller that passes no `cost` now falls back to `DEFAULT_ADMISSION_COST_CONFIG`'s ~16 MB implied ceiling instead of 256 KiB (confirmed via `rg -n "extractAdmissionCostFeatures\("` that the only other caller, `runtime.ts`, is updated in the same PR).
- `open-sse/services/admission/runtime.ts`: `AdaptiveAdmissionRuntimeImpl` resolves and stores `costConfig` (`resolveCostConfig(config.cost)`, the same resolution `validateConfig()` already does for the controller) in its constructor, and passes it as `cost: this.costConfig` on the `extractAdmissionCostFeatures` call inside `acquire()`. Body sizing now follows whatever cost config is active, including a runtime `cost` override.

## Regression test

`tests/unit/issue-13164-admission-cost-body-flatten.test.ts` (new)

RED on unfixed code (captured before applying the fix, in this worktree):

```
✖ distinguishes a 2MB body from an 8MB body (both well inside the ~16MB implied range)
  AssertionError: expected 8MB body (270108) to report a larger size than the 2MB body (270108)
✖ honors an explicit cost override, shrinking the effective byte-exit ceiling
ℹ tests 3 / pass 1 / fail 2
```

GREEN after the fix:

```
✔ reports a larger bodyBytes/estimatedInputTokens for a 2MB body than for a 260KB body
✔ distinguishes a 2MB body from an 8MB body (both well inside the ~16MB implied range)
✔ honors an explicit cost override, shrinking the effective byte-exit ceiling
ℹ tests 3 / pass 3 / fail 0
```

## Gates run

- `npx eslint --suppressions-location config/quality/eslint-suppressions.json <changed files>` → 0 errors/warnings
- `npm run check:open-sse-typecheck` → `openSseTypecheckErrors=0`, OK
- `node scripts/check/check-file-size.mjs` → no violation on touched files (1 pre-existing violation on `open-sse/utils/stream.ts`, untouched by this PR — base drift)
- `node scripts/check/check-complexity.mjs` → OK (2824 violations vs baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` → OK (1276 violations vs baseline 1437)
- `node scripts/check/check-test-discovery.mjs` → OK, new test file discovered
- Focused suite: `node --import tsx/esm --test tests/unit/adaptive-admission-features.test.ts tests/unit/adaptive-admission-cost.test.ts tests/unit/estimateSizeFast.test.ts tests/unit/issue-13164-admission-cost-body-flatten.test.ts` → 35/35 pass
- `tests/unit/adaptive-admission-runtime.test.ts` → 17/17 pass

## Existing tests aligned

`tests/unit/adaptive-admission-features.test.ts` — "bounds tool scans and never touches entries beyond the budget" pads the body with a leading string sized to trigger the size-estimator's byte-exit *before* the tools-proxy walk, per its own comment. Its pad (`"x".repeat(300_000)`) was below the old 256 KiB default but now sits well under the new ~16 MB default ceiling, so it stopped triggering the byte-exit and the test failed with `tool entry 64 must not be touched` (proving `countTools` itself, not the walker, was reached — a real behavior change caused by this fix, not the bug it's meant to isolate). Bumped the pad to `17_000_000` bytes (comfortably above `bodyBytesPerUnit * maxRequestCost`) so the test keeps isolating what it says it isolates.

## Credit

A prior contributor PR (#13291, `@lorenzozanee`) proposed the same "thread the real cost config" direction but was self-closed unmerged with no review before this PR was opened. Not reusing its code (independently derived from the source), but crediting the direction with an `Originally-proposed-by` commit trailer.


diegosouzapw:
Closing in favour of #13762 by @lorenzozanee, which was opened first and fixes the same defect (#13164): `extractAdmissionCostFeatures` calling `estimateSizeFast(body)` without a byte limit, so every body above the 256 KiB default early-exit reports the same size.

This PR was produced by a parallel maintainer sweep that started before #13762 appeared; the contributor's PR gets the credit and goes through review. Nothing here is lost — if the review of #13762 wants it, the extra regression test from this branch (a 8 MB vs 2 MB body reporting different sizes, matching the issue's own 270108/270108 numbers) can be ported into that PR.

Issue #13562, opened 2026-09-13
[BUG] Disable auto/* and no-think/* Combos


texastoland:
### OmniRoute Version

3.8.50

### Installation Method

Electron desktop app

### Operating System

macOS

### Provider(s) Involved

commandcode

### Model(s) Involved

eg no-think/cmd/claude-opus-5-xhigh

### Client Tool

https://github.com/JasonLandbridge/pi-omniroute-sync

### Description

moved from https://github.com/diegosouzapw/OmniRoute/issues/9418#issuecomment-5648602963:

#9535 only implemented/tested the db setting. there's no ui/user control. I also tried inserting directly into the db but thinking variants persist.

### Steps to Reproduce

1. enable `hideNoThinkVariants` setting in kv db
2. check `/models` endpoint

### Expected Behavior

`no-think/cmd/claude-opus-5-xhigh` missing

### Actual Behavior

`no-think/cmd/claude-opus-5-xhigh` present

### Test Impact

Needs a new integration test

### Validation Plan

unit test api triggered by ui and e2e test `/models` endpoint

xiaoyaner0201:
Same need here, and we hit both halves of this independently. Adding what we verified against `release/v3.8.51` @ `152d95108c9c3d557562311ffed63240a511eb31`, plus the nginx workaround we currently run in production, in case it saves someone the same detour.

## Our shape of the problem

Internal gateway, OpenAI-compatible clients that enumerate the catalog by id. Direct upstream `/v1/models` for one key:

| | count |
|---|---|
| total ids | 633 |
| `auto/*` | 40 |
| `no-think/*` | 12 |
| effort-suffixed ids that also have an unsuffixed sibling | 130 |
| distinct base ids after collapsing the effort suffix | 459 |

`codex/` alone is 28 rows, 21 of which are `-{effort}` variants of the other 7. Model pickers are unusable at that width, and picking `…-xhigh` from a list is not the same affordance as setting a reasoning level.

## 1. "no ui/user control" — it's worse than no UI: there's no API path either

[...]

Issue #12684, opened 2026-09-04
[Feature] Option to hide synthetic model variants (`no-think/…`, effort-tier suffixes) from `/v1/models`


kanishk619:
### Problem / Use Case

Creating this because 3.8.50's `/v1/models` advertises a lot of synthetic ids that are really *routing variants of the same base model*, not distinct models.

For example, one base model expands into many catalog rows:
- `no-think/<provider>/<model>` (thinking forced off)
- `<provider>/<model>-low`, `-medium`, `-high`, `-xhigh` (Claude reasoning effort)
- `<provider>/<model>-none`, `-low`, `-high`, `-max` (synced effort tiers)
- plus alias↔canonical duplicates (`gh/…` vs `github/…`, `fc/…` vs `firecrawl/…`)

OpenAI-compatible clients that enumerate the catalog by model id (Hermes, Codex model pickers, plain SDKs) can't express a reasoning level as a *config* the way VS Code can, so they fill the picker with every variant. On our instance this is 69 rows for what are really ~49 base models.

### Proposed Solution

There's already `hideAutoCombos` and `hideNoThinkVariants`, so this asks for the same kind of toggle for the others:
- hide the Claude effort variants (`appendClaudeEffortVariants`)
- hide the synced effort-tier variants (`appendSyncedEffortVariants`)
- optionally, a way to collapse alias↔canonical duplicates (today `?prefix=alias`/`canonical` overrides but `canonical` drops models that have no canonical form)

One more thing: a per-API-key `blocked_models` deny-list can't reach these rows — they're appended *after* the per-key filter pass in `applyCatalogPostFilters`, so the key never gets a chance to see them. A global toggle (like `hideNoThinkVariants`) is the clean fix. Routing for explicitly-sent ids would stay unaffected, exactly like the current `hideNoThinkVariants` behavior.

**Expected:** with the toggle(s) on, `/v1/models` returns only base/chat models; effort/no-think variants remain routable when sent explicitly but aren't advertised.

diegosouzapw:
Reasonable request, and there is a precedent in the codebase that makes it cheap.

[...]

PR #13800, opened 2026-09-15
fix(api): persist hideAutoCombos/hideNoThinkVariants in settings schema (#13562)

diegosouzapw:
Refs #13562

## Not covered here

- No dashboard UI toggle for `hideAutoCombos` / `hideNoThinkVariants` is added in this PR. The
  reporter's primary complaint ("no UI/user control") is only partially addressed: the settings
  API now actually persists the two flags, but there is still no dashboard surface to flip them.
  #12684 already proposes a single UI pass covering these two toggles plus two new ones
  (`hideClaudeEffortVariants`/`hideSyncedEffortVariants`); leaving UI to that issue avoids two
  separate dashboard-diff churns on the same settings section.

## Root cause

`hideAutoCombos` and `hideNoThinkVariants` were added as settings defaults and wired into the
`/v1/models` catalog logic by #9535 (closing #9418), but `updateSettingsSchema`
(`src/shared/validation/settingsSchemas.ts`) never declared either key. That schema is a plain
`z.object({...})` with no `.strict()`/`.passthrough()`, so Zod's default "strip unknown keys"
behavior silently dropped both fields. `PATCH /api/settings` therefore answered `200 OK` while
discarding `hideAutoCombos`/`hideNoThinkVariants` before they ever reached the DB — there was no
supported way (UI or API) to turn either toggle on.

## Fix

Add `hideAutoCombos: z.boolean().optional()` and `hideNoThinkVariants: z.boolean().optional()`
to `updateSettingsSchema`, next to the existing `hidePaidModels` toggle. Purely additive — no
existing schema behavior changes, and the catalog-side consumption code (`=== true` comparisons
in `catalog.ts` / `catalogResponse.ts`) already existed and is covered by
`tests/unit/catalog-hide-auto-no-think.test.ts` (added by #9535); only the settings-route
persistence path was missing.

## Regression test

`tests/unit/settings-route-hide-variants-13562.test.ts` — PATCHes `/api/settings` through the
real route handler with `{hidePaidModels: true, hideAutoCombos: true, hideNoThinkVariants: true}`
and asserts all three persist.

RED (before the fix — schema key missing):

```
✖ PATCH /api/settings persists hideAutoCombos / hideNoThinkVariants (#13562)
AssertionError [ERR_ASSERTION]: hideAutoCombos should have persisted
false !== true
ℹ pass 0
ℹ fail 1
```

GREEN (after adding the two schema keys):

```
✔ PATCH /api/settings persists hideAutoCombos / hideNoThinkVariants (#13562)
ℹ pass 1
ℹ fail 0
```

## Gates run

- `npx eslint --suppressions-location config/quality/eslint-suppressions.json src/shared/validation/settingsSchemas.ts tests/unit/settings-route-hide-variants-13562.test.ts` → 0 problems
- `npm run typecheck:core` → clean (see PR checks / CI for the authoritative run under this loaded devbox)
- `node scripts/check/check-file-size.mjs` → no ✗ on touched files (`open-sse/utils/stream.ts` is a pre-existing, unrelated frozen-file ✗ on `main`/base, not touched here)
- `node scripts/check/check-complexity.mjs` / `check-cognitive-complexity.mjs` → no new offenders
- `node scripts/check/check-test-discovery.mjs` → new test file discovered
- `tests/unit/catalog-hide-auto-no-think.test.ts` (existing, related) → 4/4 pass in isolation
- `tests/unit/hide-paid-models-settings-schema.test.ts` (existing, sibling pattern) → 3/3 pass

## Existing tests aligned

None needed — no existing assertion encoded the old (buggy) drop-the-keys contract.

## Credit

Thanks @texastoland for the detailed report pinpointing both the missing UI control and the
silent API no-op.


texastoland:
confirmed in nightly build #14191 :rocket:

PR #13814, opened 2026-09-15
fix(ci): clear the tap.testFiles drift that reds the mutation gate on every PR

abhisheksharma2411:
`mutation-test-coverage` is red on **every open PR** against `release/v3.8.51`, and none of them caused it. The gate fails on a pristine checkout of the release tip with no PR diff involved.

## Reproduced first

On `df87e936`, untouched:

```
$ node scripts/check/check-mutation-test-coverage.mjs --strict
Scanned 5127 unit test file(s) against 31 mutated module(s).

  open-sse/services/accountFallback.ts
    + tests/unit/daily-reset-tz-threading.test.ts
    + tests/unit/noauth-model-lockout.test.ts
  src/sse/services/auth.ts
    + tests/unit/free-badge-provider-gate.test.ts
    + tests/unit/noauth-model-lockout.test.ts
  open-sse/services/combo/comboPredicates.ts
    + tests/unit/local-token-budget-429-skips-cooldown.test.ts
  open-sse/services/combo/rrState.ts
    + tests/unit/daily-reset-tz-threading.test.ts

✗ 6 covering unit test(s) across 4 module(s) are missing from stryker.conf.json tap.testFiles.
```

Six coverings, **four distinct files** — `daily-reset-tz-threading` and `noauth-model-lockout` each cover two mutated modules.

## After

```
$ node scripts/check/check-mutation-test-coverage.mjs --strict
Scanned 5127 unit test file(s) against 31 mutated module(s).
✓ No drift — every covering unit test is listed in tap.testFiles.
exit 0
```

## What this actually fixes

Not just a red check. A test absent from `tap.testFiles` doesn't run in the Stryker sandbox, so **its mutant kills don't count** — the four modules above have been scored as if these tests didn't exist. `accountFallback.ts` and `sse/services/auth.ts` are both on the mutate list precisely because they're worth scoring, so the number they've been reporting is lower than the suite actually earns.

All four files pass (31 tests) at this head, so registering them can't introduce a failing mutation run:

```
daily-reset-tz-threading + noauth-model-lockout
  + free-badge-provider-gate + local-token-budget-429-skips-cooldown    31 pass, 0 fail
```

## Convention

Inserted into the alphabetical run, matching the surrounding entries. The list has a second, unsorted appended group further down — left alone rather than re-sorting a file this needs four lines in.

## Relationship to #13743

#13743 has the same diagnosis and a good reproduction recipe, but targets `reset-aware-request-scope-12600.test.ts`, which is **already registered** at `stryker.conf.json:372` (landed in `25bc16d87e`) — so applying it yields a duplicate entry and the gate stays red. I've commented there with the same checker output. This PR is the set the checker actually reports today; if @patrykkopycinski would rather fold it into theirs, I'm happy to close this in favour of that.


diegosouzapw:

PR #13817, opened 2026-09-16
fix(zen): route muse-spark-1.3 variants to Responses API like 1.2

KSonny4:
Zen serves Muse Spark only on the OpenAI Responses endpoint, not /chat/completions. The zen registry declared targetFormat openai-responses for 1.2/1.2-contributor-free but never for 1.3/1.3-contributor-free, so those fell through to chat and upstream 500d.

Verified live: zen muse-spark-1.2-contributor-free via gateway returns 200 (translated to /responses, chatcmpl-resp_ id); 1.3 variants 500 until this overlay. Extends the existing 11045 regression test.


diegosouzapw:
Good catch and correct diagnosis — but this exact fix already shipped in #13471 (merged on release/v3.8.51), which also covers the sibling opencode/opencode-go registries and the context-window fields. Closing this one as subsumed; thanks for flagging it and for the regression test pattern, which matches what already landed.

PR #13819, opened 2026-09-16
fix(executors): strip invalid OpenCode stream options

### LINKED ISSUE #13699: fix(providers): OpenCode returns 400 when stream_options is retained on non-streaming requests
I am getting an HTTP 400 from OpenCode when a non-streaming request still contains `stream_options`. This also affects compaction when the original streaming request options are carried into a non-streaming call.

Tested on OmniRoute 3.8.50.

### How to reproduce

Send a request to `/v1/chat/completions` with an OpenCode Go model:

```json
{
  "model": "opencode-go/deepseek-v4.1-flash",
  "messages": [{"role": "user", "content": "Reply exactly CHECK_OK."}],
  "stream": false,
  "stream_options": {"include_usage": true},
  "max_tokens": 128
}
```

OpenCode rejects the request with HTTP 400 and a validation error about `stream_options` requiring streaming. The same inherited option should also be removed when `stream` is omitted.

### Patch that fixed it

I added a guard in `open-sse/executors/opencode.ts`, in `OpencodeExecutor.transformRequest()`, immediately after `super.transformRequest(...)`.

The guard below shows the patch logic. `outgoingBody` is the result from the parent transform, and `originalBody` is the incoming request body:

```ts
if (
  outgoingBody &&
  typeof outgoingBody === "object" &&
  !Array.isArray(outgoingBody) &&
  (this._requestFormat !== "openai" || originalBody?.stream !== true) &&
  Object.prototype.hasOwnProperty.call(outgoingBody, "stream_options")
) {
  delete outgoingBody.stream_options;
}
```

This preserves `stream_options` for streaming Chat Completions, but removes it for non-streaming requests and prevents the inherited Chat Completions option from leaking into other OpenCode request formats.

This is the provider-specific workaround I tested, not a claim that every Responses provider rejects all streaming options.

For compaction, I also removed inherited `stream_options` where the caller forces a non-streaming request. That caller-side change is outside OmniRoute; the executor guard is the OmniRoute fix and also covers direct clients.

I kept the guard as a separate build-time patch and applied it to the compiled executor, since editing the source file alone does not change an already-built image.

### Tested

- `stream: false` with inherited `stream_options`: HTTP 200 after the fix.
- Omitted `stream` with inherited `stream_options`: HTTP 200.
- `stream: true`: completed successfully and usage was still present.
- Compaction and regular streaming requests passed after applying the provider and caller-side guards.

Can this validation be handled at the provider boundary so inherited streaming options do not break non-streaming calls?

qinghuanandejiangshi:
## Summary

OpenCode requests inherited `stream_options` even when the outgoing body was non-streaming, causing upstream HTTP 400 responses. Strip this Chat Completions option at the OpenCode executor boundary unless the request uses the OpenAI format with `stream: true`. Preserve streaming usage options and leave the caller's input unchanged.

Thanks to the reporter of #13699 for the reproduction, tested workaround, and format-scoping explanation. This patch adds regression coverage and uses the executor's existing format resolution for direct transform calls.

## Related Issues

Closes #13699

## Validation

- [x] Change type: provider / executor
- [ ] `npm run lint` — exits 2 for stale suppressions; identical result with the unchanged upstream production file. `npm run lint -- --pass-on-unpruned-suppressions` passes without suppressing lint errors. No suppression configuration was changed.

Commands:
- `node --import tsx/esm --test tests/unit/opencode-stream-options-13699.test.ts tests/unit/opencode-executor.test.ts tests/unit/opencode-strip-client-metadata-1442.test.ts tests/unit/provider-translate-path-golden.test.ts` — 73 passed.
- New regression file against the unmodified base: 7 failed, 3 passed; after the fix: all 10 passed.
- `npm run check:provider-consistency` and `npm run check:provider-assets` — passed.
- Pre-commit hooks ran, including lint-staged, docs sync, any-budget and tracked-artifact checks.

## Tests Added Or Updated

- `tests/unit/opencode-stream-options-13699.test.ts`

## Coverage Notes

Covers explicit false/omitted/true stream values for Go and Zen, removal from Claude/Responses formats, direct transform format resolution, input preservation, and an execute-path mocked upstream that rejects invalid request bodies. The broad suite, coverage gate and production build are left to CI as the contribution guide directs.

## Reviewer Notes

Targets `release/v3.8.51`. No migrations, flags or catalog changes. Verification is local and deterministic; no paid upstream account was used and the issue reporter's live results are not claimed as our own. Baseline lint suppression cleanup is outside this focused fix.


diegosouzapw:
Thanks @qinghuanandejiangshi — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

Issue #12568, opened 2026-09-03
fix(docker): compose defaults bind 0.0.0.0 while REQUIRE_API_KEY=false leaves /v1 anonymous


jsawyerdev:
## Summary
The shipped `docker-compose.yml` / `docker-compose.prod.yml` publish the dashboard, API, and live-WS ports with no host-IP scoping (binds `0.0.0.0`), and `.env.example` ships `REQUIRE_API_KEY=false`. A default `docker compose --profile base up -d` (or the `cli`/prod variants) with no `.env` edits exposes an unauthenticated LLM proxy and dashboard to every network interface on the host, including any LAN or, on a cloud VM, the public internet.

## Evidence
- `.env.example:373` — `REQUIRE_API_KEY=false`
- `docker-compose.yml:40-42,129-131,198-201` — `API_HOST`/`LIVE_WS_HOST`/`HOSTNAME` default to `0.0.0.0`; port mappings (e.g. `"${DASHBOARD_PORT:-20128}:${DASHBOARD_PORT:-20128}"`) carry no host-IP prefix, so Docker publishes on `0.0.0.0` by default.
- `docker-compose.prod.yml:66-69` — same pattern in the production compose file.
- Contrast: the `redis` service in the same `docker-compose.yml` binds `${REDIS_BIND_HOST:-127.0.0.1}:...` with an explanatory comment about the exact risk this issue describes — the app's own ports were never given the same treatment.
- Source-level default is safe in isolation (`src/server/ws/liveServer.ts:53`, `src/lib/apiBridgeServer.ts:186` both default to loopback), but the compose files override it via env defaults, and there is no code-level refusal to bind non-loopback without `REQUIRE_API_KEY=true`.
- Related prior fixes that did not close this gap: #7722 (fixed only the `README.md` `docker run` example to use `-p 127.0.0.1:...`), #2034 (accepted request to respect `HOST`/`HOSTNAME` instead of hardcoding `0.0.0.0`, not yet implemented).

Permalink (commit `7802f6e`): https://github.com/diegosouzapw/OmniRoute/blob/7802f6ea163f18f349348dbfd8710a159ac32e15/docker-compose.yml#L40-L42

## Why this matters
[...]

Issue #2257, opened 2026-05-14
[BUG] invalid api key in Codex Desktop auto config


k00shi:
### OmniRoute Version

3.7.9

### Installation Method

Electron desktop app

### Operating System

Windows

### OS Version

Windows 11 25H2

### Node.js Version

24.15.0

### Provider(s) Involved

*

### Model(s) Involved

*

### Client Tool

Codex CLI (config should be the same as Codex Desktop)

### Description

unauthorized api key when fresh config.toml setup in codex.

### Steps to Reproduce

1. delete config.toml
2. let omniroute auto config and wrote new config.toml
3. unauthorized

### Expected Behavior

authorized and talk back

### Actual Behavior

unexpected status 401 Unauthorized: Invalid API key, url: http://localhost:20128/api/v1/responses, request id: -

### Test Impact

Unsure

### Error Logs / Output

```shell
unexpected status 401 Unauthorized: Invalid API key, url: http://localhost:20128/api/v1/responses, request id: ae314a69-73a0-4c1d-a71d-fd084cce8b62


{
  "auth_mode": "apikey",
  "OPENAI_API_KEY": "sk-495a4ea9fe9c68e4-ea2858-c786c953"
}


model = "antigravity/gemini-3-pro-preview"
model_reasoning_effort = "xhigh"
model_provider = "omniroute"

[model_providers.omniroute]
name = "OmniRoute"
base_url = "http://localhost:20128/api/v1"
wire_api = "responses"
env_key = "OPENAI_API_KEY"

[windows]
sandbox = "elevated"
```

### Screenshots

<img width="716" height="395" alt="Image" src="https://github.com/user-attachments/assets/06249c61-743b-4af2-a611-6b5dd471fcf2" />

<img width="1011" height="362" alt="Image" src="https://github.com/user-attachments/assets/62478c80-e98d-448b-9209-d4060c58f41f" />

<img width="961" height="215" alt="Image" src="https://github.com/user-attachments/assets/044ec376-c909-4ba8-87e1-fa65a737593d" />

### Additional Context

I might be dumb for this.

diegosouzapw:
Thanks for the detailed screenshot @k00shi. Looking at your config + the 401, the auth flow looks like this:

[...]

PR #13820, opened 2026-09-16
feat(security): warn at boot when the inference server is exposed anonymously

### LINKED ISSUE #13695: fix(security): /v1/models returns 401 while inference accepts anonymous requests — misleading auth probe when REQUIRE_API_KEY=false
## Problem

`GET /v1/models` and the inference endpoints follow **different auth settings**, so the most natural probe an operator runs to check "is my API protected?" can return the wrong answer.

- `/v1/models` is gated by the **dashboard login posture**: it requires auth when `isAuthRequired()` is true (login required and a password/OIDC configured), unless `requireAuthForModels === false` (`src/app/api/v1/models/catalogRequest.ts:12-19`).
- `POST /v1/chat/completions`, `POST /v1/responses` and the other client API routes are gated by **`REQUIRE_API_KEY`**. With it off, anonymous traffic is allowed, and even an invalid bearer degrades to anonymous (`src/server/authz/policies/clientApi.ts:73-99`).

On an instance with an admin password set and `REQUIRE_API_KEY=false`, `/v1/models` answers `401` while inference is open to anyone who can reach the port.

## Evidence

Discussion #13310 by @inbev92, a self-hosted v3.8.50 in Docker behind Traefik. Unrecognized `codex/gpt-5.6-sol-ultra` requests over `POST /v1/responses` with an empty API key were spending their Codex quota. Their own verification found:

- `REQUIRE_API_KEY=false` in the running container.
- Ports `20128` and `20129` published on `0.0.0.0`, and UFW allowing forwarded traffic to both.
- An unauthenticated `POST /v1/responses` reached model routing.
- **At the same time, `/v1/models` returned `401`, "which initially created the impression that the entire API was protected".**

The misleading 401 sent them investigating the Codex side instead of OmniRoute's access configuration.

## Proposed solution

1. **Startup warning** when `REQUIRE_API_KEY=false` and the server listens on a non-loopback address. It should say plainly that the inference API accepts unauthenticated requests from any client that can reach the port.
2. **Document the split** in `docs/security/` and next to `REQUIRE_API_KEY` in `docs/reference/ENVIRONMENT.md`: `/v1/models` follows login/`requireAuthForModels`, inference follows `REQUIRE_API_KEY`, so probing `/v1/models` does not verify inference protection.
3. Consider showing the effective inference auth posture ("API key required: no") prominently in the dashboard, e.g. on the endpoint page.

Changing either endpoint's default auth behavior is out of scope here and would need its own decision.

## Acceptance criteria

- [ ] A warning is logged at startup when `REQUIRE_API_KEY=false` and the bind address is not loopback, with a unit test covering both the warning and the loopback no-warning case.
- [ ] The docs state which setting gates `/v1/models` and which gates inference, with the probing caveat.

## Additional context

- Discussion: #13310
- Related behavior: #2257 (invalid bearer degrades to anonymous when `REQUIRE_API_KEY` is off)

abhisheksharma2411:
Closes #13695.

## What was actually missing

The issue asks for a boot warning when `REQUIRE_API_KEY=false` and the bind address is not loopback. Most of that already exists — `src/lib/startup/nonLoopbackApiKeyGuard.ts` landed with #12568, and `tests/unit/compose-app-ports-loopback-bind.test.ts` even asserts a guard is present.

What it does **not** do is cover the server the issue is about. The guard is wired into two call sites:

| call site | host | default |
| --- | --- | --- |
| `src/lib/apiBridgeServer.ts` | `API_HOST` | `127.0.0.1` |
| `src/server/ws/liveServer.ts` | its own | — |

The Next server that answers `/v1/chat/completions` and `/v1/responses` through the `next.config.mjs` rewrites never reached it — and that one binds `process.env.HOST || "0.0.0.0"`, i.e. **every interface by default**. So the exact configuration in the issue (admin password set, `REQUIRE_API_KEY=false`, published port) boots silently today, and the operator's `/v1/models` probe answers `401`.

This wires the existing guard into the Next boot hook and documents the split.

## Resolving the bound host

This needed more care than `process.env.HOST` — two entrypoints bind that server and they read **different** variables:

- `scripts/dev/run-next.mjs` (`npm run dev` / `npm start`) → `HOST`
- `scripts/dev/run-standalone.mjs` (the Docker `CMD`) → delegates to Next's generated `server.js`, which uses Next's own `HOSTNAME` convention; `Dockerfile` sets `ENV HOSTNAME=0.0.0.0`

So `run-next.mjs` now publishes the interface it actually binds as `OMNIROUTE_BOUND_HOST`, and the guard reads `OMNIROUTE_BOUND_HOST → HOSTNAME → "0.0.0.0"`.

`HOST` is deliberately **absent** from that chain. The only path that honours it is `run-next.mjs`, which has already folded it into `OMNIROUTE_BOUND_HOST`; on the standalone path Next ignores `HOST` entirely. Consulting it there would warn about an interface the server is not on — and one false warning teaches an operator to ignore the next one. `tests/unit/inference-auth-posture-13695.test.ts` pins both the ordering and the exclusion, including the case an earlier draft of this patch got wrong: a Docker instance bound to loopback via `HOSTNAME` must stay silent.

`OMNIROUTE_BOUND_HOST` is runner-owned, never operator-configured, so it is allowlisted in `check:env-doc-sync` next to `OMNIROUTE_INTERNAL_SCHEME` rather than added to `.env.example`.

## Docs (acceptance criterion 2)

- `docs/security/INFERENCE_AUTH_POSTURE.md` — new page: which setting gates which endpoint, the probing caveat, a `curl` that actually tests inference, and the #2257 note that an invalid bearer degrades to anonymous so a junk key is not a valid probe either. Registered in `docs/security/meta.json`.
- `docs/reference/ENVIRONMENT.md` — the `REQUIRE_API_KEY` row now says it does not gate `GET /v1/models` and points at the new page.

## Out of scope

Acceptance-criterion item 3 (surfacing the effective posture in the dashboard) is worded as "Consider" and is not in the AC checklist, so it is left for a follow-up. Changing either endpoint's default auth behaviour stays out of scope per the issue.

## Verification

- `tests/unit/inference-auth-posture-13695.test.ts` — 7 tests, new.
- Mutation-tested: 9 mutants (drop the boot-hook call; default host → `127.0.0.1`; drop the warning's label; revert the `ENVIRONMENT.md` row; unregister the security page; drop `HOSTNAME` from the chain; drop `OMNIROUTE_BOUND_HOST`; stop publishing the host in the runner; put `HOST` ahead of `HOSTNAME`) — all 9 killed, baseline green.
- `156 pass / 0 fail` across all 37 existing suites that touch `instrumentation-node`, `nonLoopbackApiKeyGuard`, `check-env-doc-sync` or `run-next`.
- `npm run check:env-doc-sync` ✓, `npm run check:docs-sync` ✓, `eslint` clean on changed files.


diegosouzapw:

Issue #13124, opened 2026-09-09
feat(resilience): 3.8.50 regression: psi_some guard 503-storms every model on memory-constrained hosts — ship #12017/#12293 + add opt-out knob


finesoft2009:
## Environment

- OmniRoute **3.8.50** (npm, global install), `omniroute serve`, Linux LXC container, 12 GB RAM, shared with other production services
- Upgraded 3.8.49 → 3.8.50 on **Aug 31**

## What happens

Since the 3.8.50 upgrade we get recurring **global** 503 storms:

```
[resourcePressure] critical pressure guard tripped (reason=psi_some); returning 503
{"message":"Service temporarily unavailable due to resource pressure. Retry shortly.","type":"server_error","code":"resource_pressure"}
```

- 327 rejections over 9 days (18–68/day), in short bursts of 10–30 s, 2–13 times a day
- Rejections never appear in `call_logs` / dashboard (as documented in #12293) — only in `app.log`
- Downstream agent clients retry with ~8 s backoff and land inside the same pressure window, so one burst kills entire agent turns (worse than the underlying pressure itself)

## Why we're sure it's a 3.8.50 regression

- The guard (`criticalPsiAvg10` / `highPsiAvg10` thresholds) is **absent from the 3.8.49 tarball** and appears in 3.8.50 — verified by unpacking both from npm
- Our `app.log` has 7.5k lines for Aug 29–30 on 3.8.49 with **zero** rejections; the first rejection fires on Aug 31 — the upgrade day
- The trigger is `/proc/pressure/memory` some avg10 ≥ 40 twice in a row. On a busy memory-constrained host, heavy LLM traffic (multi-100K agent contexts) spikes memory PSI for seconds — the guard then 503s **every model at once**, not just the busy route

## Ask

1. Please cut a release including #12017 and #12293 — both are merged to `main`, but npm `latest` is still 3.8.50 (Aug 28), so the fixes are unreleased
2. Please add a tuning knob / opt-out for the PSI guard (env var or config): on hosts where a global kill-switch for all providers is worse than a slow response, the `psi_some`/`psi_full` thresholds (or the guard itself) should be configurable

Happy to provide logs and memory-PSI traces from the affected host.

diegosouzapw:
Thanks for the detailed writeup — the logs and the "verified by unpacking both tarballs" diligence made this easy to confirm.

[...]

PR #13823, opened 2026-09-16
fix(resilience): resource_pressure admission gate actively re-samples instead of reading a stale cache

### LINKED ISSUE #13821: fix(resilience): resource_pressure admission guard never self-recovers once critical
## Environment

- OmniRoute source build, `systemctl --user` managed, Linux VM, 15 GB RAM
- Built at `abe234e094bdbec72626505d31b1c5bb345cf334` (release/v3.8.51, 2026-09-11 — includes the #13091 worker-idle-terminate fix from #11804)
- Client: `omp` (oh-my-pi) over `/v1/chat/completions`

## Summary

Once the `resource_pressure` admission guard trips to `critical`, the gateway keeps shedding **100% of chat requests with a 503 `resource_pressure`** even after live memory/PSI drops back to healthy levels. Only a full `systemctl restart` clears it — there is no organic recovery path.

## Reproduction observed today

Service had been running 15h01m (since `2026-09-15T09:40:12Z`) with a memory peak of `6.9G` (swap peak `767.1M`) at some point during that window (compression/chat load, no combos on this route). At the moment I hit the bug:

```
$ systemctl --user status omniroute
Memory: 920.4M (peak: 6.9G swap: 591.5M swap peak: 767.1M)   # current usage is LOW, not elevated

$ free -h
              total   used   free   shared  buff/cache  available
Mem:           15Gi   5.2Gi  6.7Gi     90Mi       4.3Gi       10Gi
Swap:          12Gi   3.8Gi  9.2Gi

$ curl -s -X POST http://.../v1/chat/completions -d "{\"model\":\"chat\",\"messages\":[{\"role\":\"user\",\"content\":\"say ok\"}],\"max_tokens\":10}"
{"error":{"message":"Service temporarily unavailable due to resource pressure. Retry shortly.","type":"server_error","code":"resource_pressure"}}
http_status=503
```

Live memory is healthy (10 GB available, swap mostly free), yet every request is still shed. `systemctl --user restart omniroute` immediately restored `200 OK` — the only thing that changed was the restart, not the underlying resource state (which, per `free -h` above, was already fine *before* the restart).

## Root cause (read from `open-sse/utils/resourcePressure.ts` + `src/shared/middleware/chatBodyAdmission.ts`)

- `chatBodyAdmission.ts` (`admitChatRequest`, ~line 1008) gates every non-internal-bypass chat request on `controller.pressureSeverity() === "critical"`. That call resolves to `ingestByteAdmission.ts`'s `pressureSeverity()`, which is a **pure cached read**: `getResourcePressureObservation().state.severity` — it never calls `checkResourcePressureGuard()`'s actual `check()`.
- `check()` (in `resourcePressure.ts`) is the *only* function that draws a fresh sample (`sampleResourceSignals()`) and can transition `state.severity` back to `"normal"`. It also self-schedules the next refresh (`scheduleRefresh()` / `setImmediate`) — but only from *within* `check()` itself.
- The only call sites that invoke the real `check()` are `handleChatCore` (`open-sse/handlers/chatCore.ts:568`), `checkResourcePressureBeforeProviderWork` (`src/sse/handlers/chatHelpers.ts:394-399`), and `AdaptiveAdmissionRuntimeImpl.acquire()` (`open-sse/services/admission/runtime.ts:357`) — all of which are **downstream** of the structural gate above for a normal client request.

So once `state.severity` flips to `"critical"` (from any of the existing pressure triggers — PSI, v8 heap ratio, a worker-leak spike, etc.), every subsequent request is shed at the cheap cached-read gate *before* it can ever reach the one code path capable of re-sampling and observing recovery. The gate and the only means of clearing it are mutually exclusive once tripped: the gate's own rejection starves the sampler that would clear the gate. This applies regardless of what originally caused the "critical" trip — worker leak (#11804, now fixed by #13091), PSI storm (#13124), or a transient real load spike — the *lack of self-recovery* is a separate, independent defect from whatever caused the initial trip.

## Suggested fix

`checkResourcePressureGuard()` (or an equivalent lightweight re-sample) should be invoked from the structural gate itself (or from a periodic background timer independent of request traffic), not only from code paths that are unreachable once the gate is shedding. Concretely: either (a) have the structural `pressureSeverity()` gate call `checkResourcePressureGuard().state.severity` instead of the passive `getResourcePressureObservation()`, so every gated request still triggers a fresh sample/refresh schedule even while being shed, or (b) run `sampleResourceSignals()` on an unconditional interval (e.g. `setInterval`) independent of any request reaching a downstream handler.

## Relationship to existing issues

- Not a duplicate of #11804 (worker-thread leak — root cause of one possible trigger, already fixed by #13091) or #13124 (PSI-threshold tuning — a different trigger). This issue is about the **absence of a recovery path** once *any* trigger flips the gate to critical.

pandudpn:
## Summary

`admitChatRequest`'s structural admission gate (`chatBodyAdmission.ts`, `defaultPressureSeverity()`) is the first caller in the request path to consult pressure severity, ahead of every other code path that would otherwise call `checkResourcePressureGuard()` (`handleChatCore`, `checkResourcePressureBeforeProviderWork`, `AdaptiveAdmissionRuntimeImpl.acquire`). It read the resourcePressure singleton's cached `state` directly (`getResourcePressureObservation`) instead of driving `check()` — so once `state.severity` flipped to `"critical"` (from any pressure trigger: PSI, v8 heap ratio, a worker-leak spike, etc.), every subsequent request was shed at this cheap cached-read gate before it could ever reach the one function capable of drawing a fresh sample and observing recovery. The gate and the only means of clearing it were mutually exclusive once tripped: the gate's own rejection starved the sampler that would clear the gate. Only a full process restart cleared it, even after the real underlying condition had already gone away.

Reproduced live on a production instance: host memory healthy (10 GB available, swap mostly free), process RSS low, yet every `/v1/chat/completions` request kept returning `503 resource_pressure` for hours, and only `systemctl restart` fixed it. Full writeup with logs, `free -h`/`memory.stat` output, and code trace: #13821.

## The change

`defaultPressureSeverity()` now calls `checkResourcePressureGuard()` first. That call is cheap on the hot path — a synchronous `process.memoryUsage()` read plus a timestamp comparison; the actual signal sampling (`/proc/pressure/memory`, cgroup reads) stays asynchronous via `scheduleRefresh()` and throttled by `staleAfterMs`, so this adds no per-request I/O. A non-null guard is this request's authoritative "shed now" answer and maps to `"critical"`. A null guard means this request is not shed, but the raw cached label can still briefly read `"critical"` until the async refresh settles (or if the last real sample simply went stale), so that case is downgraded to `"high"` rather than re-introducing the same problem for the queue-wait-sizing branch (`admitChatRequest`'s `reserve()`) that also reads this value.

Fixes #13821.

## Test

New `tests/unit/resource-pressure-gate-recovery.test.ts` drives the resourcePressure singleton to critical through the same sustained-sample path production uses (`classifyAdaptiveResourcePressure`'s `v8_heap_ratio` + a two-sample streak — not the synchronous immediate-heap escape hatch), using an injected mock clock so no scheduled refresh from setup can resolve on its own and contaminate the assertion. Only the exact calls under test (`defaultPressureSeverity`, twice) are allowed to touch the singleton after it's latched critical.

Confirmed red on `release/v3.8.51` @ `8f55d85d` (this branch's base) with the fix reverted:

```
✖ defaultPressureSeverity recovers to normal once the underlying pressure clears, driven only by repeated calls to itself
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  + actual - expected
  + 'critical'
  - 'normal'
✔ defaultPressureSeverity still sheds while genuinely critical, across repeated calls
```

The second test passes even on unfixed code — it's a control asserting the fix doesn't introduce a false "always pass" regression; the first test is the one this PR targets, and it fails exactly at the recovery assertion (the singleton stays latched at `"critical"` forever because `defaultPressureSeverity` never called `check()`).

Green with the fix: both pass.

## Validation

- [x] `node --test` on the new test file: 2/2 pass with the fix, 1/2 pass (the non-regression control only) without it
- [x] `npx tsc -p tsconfig.typecheck-core.json --noEmit`: exit 0
- [x] `npx eslint src/shared/middleware/chatBodyAdmission.ts tests/unit/resource-pressure-gate-recovery.test.ts --suppressions-location config/quality/eslint-suppressions.json`: clean
- [x] `npm run check:changelog-integrity`: OK
- [x] `node scripts/check/check-complexity-ratchets.mjs --base-ref origin/release/v3.8.51`: OK

## Reviewer notes

- Scope is intentionally narrow: `defaultPressureSeverity()` only. It is the sole call site wired as `checkPressureSeverity` for the production `ChatAdmissionController` singleton (`chatBodyAdmission.ts:812`); tests that construct their own controller with an injected `checkPressureSeverity` (e.g. `tests/unit/admission-pressure-gate.test.ts`) are unaffected.
- No change to `resourcePressure.ts`, `resourcePressurePolicy.ts`, or the sampler — the recovery/staleness machinery there (`maxStaleMs`, sustained-streak thresholds) already does the right thing; it just needed a caller in the hot path willing to invoke it.
- Related but distinct: #11804 (worker-thread leak, one trigger for hitting critical — already fixed by #13091) and #13124 (PSI-threshold tuning, a different trigger). This PR fixes the *lack of any recovery path* once *any* trigger flips the gate to critical, independent of which trigger fires.


maxmad64bis:
Heads-up on an overlap with #13618, which addresses the same latch from the other side and was reviewed on 09-15.

This one matters in practice: #13821 now has two independent occurrences with different triggers (heap and `cgroup_high`), and in ours every chat request got a 503 for about 28 minutes until a restart. Whichever fix lands, it would be good to settle the overlap soon.

Both change how `admitChatRequest` decides to shed, and both touch `defaultPressureSeverity()`. #13618 makes it private (`cachedPressureSeverityForDiagnostics`) and moves the gate to an async `getAdmissionResourcePressureSeverity()` that waits, bounded, for a refresh when the cached critical is stale. This PR keeps the function and makes it call `checkResourcePressureGuard()`. The regression test here imports `defaultPressureSeverity` from `chatBodyAdmission.ts`, so it would fail if #13618 lands first, and the hunks in `chatBodyAdmission.ts` conflict textually.

Only one of the two is needed to fix #13821, so it may be worth choosing before either merges. I haven't compared them under load, this is only about the overlap.

diegosouzapw:
Thanks @pandudpn — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

PR #13747, opened 2026-09-15
fix(ci): clear the release/v3.8.51 base-reds left by the 09-15 batch (stryker, CLI i18n, paid-target fixture, call-log traceId, Jina prefix, callLogStats import, gitleaks)

diegosouzapw:
## Why

Every PR into `release/v3.8.51` pushed **after** #13635/#13678 still failed `Fast Quality Gates` and all four `Unit Tests fast-path` shards, on the same 16 tests (#13266, #13173, #13149, #13150, #13355, #13628, #13717 and #13720 fail on exactly this set). Each failure reproduces on the pure tip. All five come from yesterday's batch: none is a product defect, and no assertion is weakened.

| Red | Root cause | Fix |
|---|---|---|
| FQG `mutation-test-coverage` | `noauth-model-lockout` and `local-token-budget-429-skips-cooldown` (#13606) cover `accountFallback.ts` / `auth.ts` / `comboPredicates.ts`, but neither is in `stryker.conf.json` `tap.testFiles` | add both entries |
| `cli-i18n-catalog` (1, then 2 parity) | `--ready-timeout` calls `t("serve.ready_timeout")` with no catalog entry | add the key to `en`, `zh-CN`, `zh-TW` (the parity-checked CLI locales) |
| `paid-model-target-6540` (1) + `paid-model-target-routes-6540` (3) | #13407 removed Together's one-time credit, so `together/…` now classifies as `unknown`, and the save-time guard correctly lets it through | the paid fixture is now `gemini/gemini-3.1-pro-preview` (Gemini's free tier recurs and does not include Pro), plus a precondition test that the fixtures still classify as paid/free/unknown |
| `attempt-logging-early-keepalive-merge` (3) + `video-bridge-log-redaction` (6, 30 s timeouts) | #13546 keys the call-log row on `traceId`, but these tests poll by `pendingRequestId` | `baseCtx` defaults `traceId` to `pendingRequestId`, the same pattern `chatcore-attempt-logging` got in #13546. The keepalive poll moves to the 30 s wall-clock deadline `video-bridge` already uses |
| API Route Typecheck + `check:dashboard-typecheck` | the #13641 merge left `ERROR_TYPE_CONTRACT` imported twice in `src/lib/db/callLogStats.ts` (TS2300) | drop the duplicate import |
[...]

PR #13826, opened 2026-09-16
chore(ci): clear the last two release/v3.8.51 base-reds (agent-skills sync + stryker tap.testFiles)

diegosouzapw:
Refs #12732

## Why

`check:agent-skills-sync` (the `Merge integrity (changelog + generated skills)` job) fails **on the pure release tip**, so every PR cut from it inherits the red. It is the last base-red left after #13747.

Reproduced on `origin/release/v3.8.51` in a clean worktree:

```
$ node --import tsx/esm scripts/skills/generate-agent-skills.mjs
  GENERATED:
    + omni-settings
(dry-run) No files written. Use --apply to write SKILL.md files.   → exit 2
```

## Root cause

#13581 added `GET /api/settings/proxies/pool/egress-observation`. `skills/omni-settings/SKILL.md` is generated from the route catalog (`src/lib/agentSkills/catalog.ts`), and it was not regenerated when that route landed.

## Fix

`scripts/skills/generate-agent-skills.mjs --apply`. The whole diff is the generated section for the new route (+11 lines); nothing hand-written, no other skill changed (45 unchanged, 0 pruned, 0 orphans).

After the change the same command exits 0 with `Generated: 0 · Unchanged: 46`.

## Note for the reviewer

This touches `skills/**/SKILL.md`, which is an agent-instruction surface, so it needs explicit owner approval before merging even though the content is machine-generated. The added text only documents an existing read-only endpoint; it carries no instruction to any agent.


diegosouzapw:
Second base-red folded into this PR: `check:mutation-test-coverage --strict` also fails on the pure tip — `tests/unit/provider-401-ambiguous-runtime.test.ts` (#13609) covers `accountFallback.ts` and `auth.ts` but was never registered in `stryker.conf.json` `tap.testFiles`.

Reproduced on the tip and verified after the fix:

```
before: ✗ 2 covering unit test(s) across 2 module(s) are missing from stryker.conf.json tap.testFiles
after:  ✓ No drift — every covering unit test is listed in tap.testFiles.
```

It surfaced as a `Fast Quality Gates` red on #13751 and #13759, which is where I traced it from.

PR #13833, opened 2026-09-16
fix(translator): preserve empty tool responses in openai-to-gemini

adrianaryaputra:
### Summary
When a tool produces empty output (e.g. silent bash commands like `true` or `cd`, commands piping to files, or terminal executions returning 0-byte stdout), OpenAI-compatible agent harnesses (such as Prime Agent, Hermes, OpenHands) send a message with `role: 'tool'` and `content: ''`.

In `open-sse/translator/request/openai-to-gemini.ts`, the translator previously checked:
```typescript
const hasActualResponses = toolCallIds.some((fid) => toolResponses[fid]) || ...
for (const fid of toolCallIds) {
  if (!toolResponses[fid]) continue;
  ...
}
```

Because empty string `''` is falsy in JavaScript:
1. `toolResponses[fid]` evaluated to `false`.
2. The entire tool turn was skipped, and no `functionResponse` part was emitted for that tool call.
3. The resulting payload sent to Google Gemini upstream ended with an unfulfilled `role: 'model'` / `functionCall`, immediately triggering Google's API rejection:
   `[400]: Requests ending with a model turn are not supported.`

### Changes
- Changed checks in `open-sse/translator/request/openai-to-gemini.ts` from truthy check (`toolResponses[fid]`) to existence check (`toolResponses[fid] !== undefined`).
- Fallback empty string (`toolResponses[fid] ?? ''`) preserves valid empty output in `functionResponse`.
- Added unit test in `tests/unit/translator-openai-to-gemini.test.ts` verifying that `content: ''` produces a valid `functionResponse` with `{ result: '' }`.

### Verification
- Ran `node --test tests/unit/translator-openai-to-gemini.test.ts`: 46/46 passed.
- Ran all Gemini translator test suites: 100% passed.


diegosouzapw:
Thanks for tracking this down — the falsy-vs-undefined diagnosis for empty tool output is exactly right. We found that #13848, a separate fix for a related cross-turn tool-ID collision bug, ends up applying the identical undefined-check fix as part of its broader turn-scoping change, so your scenario (content: '') is already covered there — we verified by running your own new test against that PR's branch and it passes. We're merging that one and closing this as covered; also flagging that this branch picked up a stale expectation bump in tests/unit/hard-session-lease-bypass-inventory.test.ts (from a since-reverted #13398 state) that would fail CI as-is if merged standalone, in case you want it for reference. Appreciate the clear root-cause writeup either way.

PR #13846, opened 2026-09-16
feat(api): add per-agent AgentBridge restore-default endpoint

trieuvietthang:
## Summary
- Enabling a MITM agent (e.g. Antigravity) via AgentBridge had no matching full revert: disabling DNS only un-spoofed the hosts file, leaving stale model mappings and `setup_completed` state behind, and on Windows the DNS Client's hosts-file cache kept resolving the old `127.0.0.1` entry until a manual `ipconfig /flushdns` — so the IDE stayed broken even after turning the integration off.
- Adds `POST /api/tools/agent-bridge/agents/[id]/reset`, which un-spoofs only that agent's hosts, clears its saved mappings, resets `dns_enabled`/`setup_completed`, and best-effort flushes the Windows DNS cache — without touching the shared MITM server or root CA other agents may still depend on.
- Wires a new "Restore default" action into each agent card (`AgentCard.tsx`), with an inline confirm and a banner reminding the user to fully quit and relaunch the IDE (Electron singleton lock means a plain restart from the icon won't pick up the change).

## Test plan
- [x] `tests/unit/agent-bridge-reset-route.test.ts` (6 new cases): unknown-agent 404, malformed JSON 400, invalid body 400, no stack-trace leak, mappings/state cleared for the target agent only (sibling agent untouched), `otherAgentsStillActive` computed correctly.
- [x] Reran the related pre-existing suites (`agent-bridge-dns-route-validation`, `agent-bridge-dns-sudo-gate`, `db-agent-bridge-state`, `agent-bridge-mappings-sync-8656`, `db-agent-bridge-mappings`) — all 25 still pass, no regressions.
- [x] `tsc -p tsconfig.typecheck-core.json` — zero new errors on changed files (9 pre-existing unrelated errors in `omniglyph` compression code, confirmed via `git diff` to be outside this change).
- [x] `eslint` on changed files — zero new errors (2 pre-existing unrelated violations confirmed outside this diff).


diegosouzapw:
Thanks for this, @trieuvietthang. Having a real "Restore default" for AgentBridge closes a gap people have been hitting.

This branch was created from `main`, so against `release/v3.8.51` it shows about 1,300 changed files and can't be merged as-is. Instead of asking you to rebuild it, we re-landed it in #14730. Your commit is cherry-picked there with your authorship, and you'll be credited as co-author on the squash.

Three small follow-ups were needed to make it pass on the release branch:
- `flushWindowsDnsCache` used `IS_WIN`, which only exists on `main`. The release branch uses `isWin32()`, which fixed the 500 from the reset route.
- The route test no longer depends on the machine's real `/etc/hosts`.
- The four new UI strings were added to the other 65 locales.

Once #14730 merges, we'll close this PR and point back here.

Issue #10524, opened 2026-08-16
[BUG] SSE stream trailers (`: x-omniroute-*` comments) break strict SSE clients (e.g. WorkBuddy)


ikairz:
### OmniRoute Version

3.8.49

### Installation Method

npm (global)

### Operating System

macOS

### Description

OmniRoute 3.8.49 injects telemetry trailers as SSE comment lines (: x-omniroute-*) at the end of /v1/chat/completions streaming responses, right before or after the final data: [DONE] event. Per the SSE spec, lines beginning with : are comments and should be ignored by clients. However, strict SSE parsers (such as WorkBuddy — see https://www.workbuddy.ai/docs/workbuddy/Overview) attempt to JSON-parse every non-data: line, which causes a Invalid JSON body 400 error because these comment lines are not valid JSON payloads. The telemetry data is for server-side observability only and provides no value to the calling client. There is currently no environment variable, configuration flag, or documented option to suppress or disable this behavior.

### Steps to Reproduce

1. Send a POST /v1/chat/completions request to OmniRoute 3.8.49 with "stream": true
2. Consume the SSE stream until the final data: [DONE] event
3. Observe that 9 additional comment lines appear in the response body, for example:

: x-omniroute-cache-hit=false
: x-omniroute-latency-ms=1207
: x-omniroute-version=3.8.49
: x-omniroute-model=deepseek-ai/deepseek-v4-flash-0731
: x-omniroute-provider=nvidia
: x-omniroute-decision=provider=nvidia; latency_ms=1207

4. Use WorkBuddy or any SSE client that parses each line and attempts JSON.parse on non-data: lines. The client will fail with Invalid JSON body 400.

### Expected Behavior

- The stream consumer receives only standard SSE data: event lines and a final data: [DONE], with no extra non-standard lines appended.
- Or, the project provides a configuration switch (e.g. OMNIROUTE_DISABLE_SSE_TELEMETRY=true) that allows callers to suppress these trailers.
- Alternatively, telemetry is emitted using a proper SSE event: type (e.g. event: x-omniroute-telemetry) that SSE parsers naturally skip, rather than as raw comment lines that some strict parsers mis-handle.

### Actual Behavior

- The SSE stream is polluted with 9 unsolicited :-prefixed comment lines at the end.
- Strict SSE clients (WorkBuddy and similar) receive these lines, attempt to parse them as JSON, and surface a Invalid JSON body 400 error to the user.
[...]

PR #13850, opened 2026-09-16
fix(sse): emit valid responses keepalives

TheDemonTuan:
## Summary
- Replace malformed synthetic `response.in_progress` heartbeat events with SSE comments
- Preserve real Responses API event sequence numbers and payloads without emitting schema-violating placeholders
- Update early- and mid-stream regression coverage

## Verification
- `node --import tsx/esm --test tests/unit/sseHeartbeat.test.ts tests/unit/sse-heartbeat.test.ts tests/unit/earlyStreamKeepalive.test.ts tests/unit/early-stream-keepalive.test.ts tests/unit/chatcore-streaming-pipeline.test.ts tests/unit/responses-route-early-keepalive-wiring.test.ts` (all passed)


diegosouzapw:
Re-reviewed this against the fuller bug report that has since landed (#14330), and two things need
changing before it can merge.

First, the new keepalive shape bypasses the opt-out from #10524. The gate at
`open-sse/utils/sseHeartbeat.ts:107` keys on `shape === HEARTBEAT_SHAPES.COMMENT`, and the new
`openai-responses-keepalive` shape emits a `:` comment without being that shape — so
`OMNIROUTE_SSE_COMMENTS=off` no longer suppresses it, which re-breaks the strict clients that flag
was added for. Either reuse the `COMMENT` shape or widen the gate to any comment-shaped payload.

Second, #14330's third criterion is the invalid *failure* frames, and this PR doesn't touch them:
six emitters ship `response.failed`/error events with no `sequence_number` or a hardcoded `0` while
the translator numbers from 1 (`streamHandler.ts:454,464-472`, `diagnostics.ts:143-159`,
`codex.ts:515-520` and `:982-989`, `streamErrorFormat.ts:296-307`,
`earlyStreamKeepalive.ts:83-90,128`), and the same invalid keepalive is also emitted mid-stream at
`responsesHandler.ts:91-94`. Either extend the PR or say in the body that #14330 stays open for
them — and please use `Refs #14330` rather than `Closes`, since our default branch is the release
branch and `Closes` would auto-close it.

One coordination note: #14377 asks for the keepalive *shape* to be configurable, which this PR
decides one way by default. Fixing the invalid frames is unconditional and welcome; the shape choice
is probably better behind the env switch that issue asks for. Reds currently on the branch are
inherited from base-red #14496.

PR #13855, opened 2026-09-16
fix(providers): normalize non-function tool types for allowlisted built-in OpenAI-format providers

### LINKED ISSUE #13789: fix(providers): built-in OpenAI-format providers forward non-function tool types unchanged — agentrouter GLM rejects "tools[0].type is illegal"
## Summary

On built-in providers that use the OpenAI Chat format, a client tool whose `type` is not `"function"` reaches the upstream unchanged. An example is a hosted web search tool, or a Claude server tool without `input_schema`. The step that converts or drops these tools only runs for custom `openai-compatible-*` providers.

AgentRouter's GLM backend rejects such a request with `400 tools[0].type:type is illegal`.

## Evidence

- `open-sse/handlers/chatCore.ts:2448-2457` calls `normalizeOpenAICompatibleTools()` only when `provider?.startsWith("openai-compatible-")`. Built-in providers such as `agentrouter`, which uses the alternate `openai` format at `https://agentrouter.org/v1/chat/completions`, skip it.
- `open-sse/translator/helpers/openaiHelper.ts:211-246` runs for every OpenAI target. It converts function tools, Claude-shaped tools that carry `input_schema` or `description`, and Gemini `functionDeclarations`. Every other shape is passed through unchanged by the final `return tool;`.
- `open-sse/translator/request/claude-to-openai.ts:214-241` is not affected. It converts every named Claude tool to `type: "function"`, except native web search when the target is the Responses format. The gap is on OpenAI-format client requests.

## Report

Discussion #13346 (reporter @SL-in):

- `agentrouter/glm-5.3`, connection switched to the OpenAI-compatible format, after the Payload Rule workaround suggested for #13696 (the earlier `thinking.type` 400 no longer appears in the report).
- The request then fails with `[400]: tools[0].type:type is illegal`.
- The same model works from OpenCode.
- **Not yet confirmed:** the reporter was asked for `body.tools[0]` from the Provider Request log section.

## Open question before fixing

Normalizing every non-function tool globally is not safe. OpenAI's own Chat Completions accepts `type: "custom"` tools, so blanket conversion could alter requests to providers that support them. Options:

1. Gate the conversion on a per-provider or per-model capability ("function tools only"), set for AgentRouter GLM and similar backends.
2. Extend the existing `openai-compatible-*` normalization to a list of built-in providers known to accept only function tools.

## Acceptance

- A unit test where an OpenAI-format request with a `{ type: "web_search" }` tool, targeting `agentrouter` + `glm-*`, reaches the executor with no non-function tool. The test fails on the current tip.
- The same request targeting `openai` keeps its `custom` tools untouched.

Refs #13346, #13696

xiechimon:
Closes #13789.

The normalization step only ran for custom `openai-compatible-*` connections, so built-in providers on the OpenAI Chat wire (agentrouter) forwarded non-function tool types verbatim and GLM rejected the whole request with `400 tools[0].type:type is illegal`.

This is option 2 from the issue: the gate moves into `shouldNormalizeFunctionToolsOnly()`, with a conservative allowlist of built-in providers (agentrouter only, for now) that applies on the OpenAI Chat target. Responses targets keep native hosted tools, Claude targets keep their own dispatch normalization, and openai stays off the list so its `custom` tools pass through untouched.

Tests: an end-to-end chatCore test where agentrouter+glm receives a mixed tool catalog and every tool reaching the executor is a function tool (named server tools convert, nameless hosted tools drop, `web_search` lands as the fallback function). It fails on the current tip and passes here. Plus the acceptance control: openai keeps `custom` tools. Three helper-level tests cover the gate itself.

Validated: focused suites 21/21 green, `npm run lint` clean on touched files, prettier clean, `typecheck:core` clean.


diegosouzapw:
Thanks @xiechimon — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #13857, opened 2026-09-16
feat(dashboard): expose comboTimeoutMs next to Target timeout

HouMinXi:
## Status

Combo failover already honors `config.comboTimeoutMs` as the whole-combo wall-clock budget. `0` (the default) keeps unlimited iteration and falls back to the 10-minute hang-stop (`COMBO_LOOP_SAFETY_TIMEOUT_MS`). A positive value replaces that hang-stop for the combo.

The dashboard only painted Target timeout (`targetTimeoutMs`, per-step). Operators who needed a longer 15-step failover ceiling had to PUT JSON by hand.

## Problem

`comboRuntimeConfigSchema` listed dozens of knobs but not `comboTimeoutMs`. The key survived PUT only because the schema ends in `.passthrough()`. Combo editor Advanced and Combo defaults had no input, so a thinking-heavy combo with Target timeout = 600s still died at the 600s hang-stop after the first target.

## Change

- Declare `comboTimeoutMs` on `comboRuntimeConfigSchema` (`0..MAX_TIMER_TIMEOUT_MS`).
- Extract Target timeout + Combo timeout into `ComboTimeoutFields` so `combos/page.tsx` stays under the frozen loc cap (5071 → 5035).
- Add the same Combo timeout field to Combo defaults.
- English copy plus `docs/reference/ENVIRONMENT.md` distinguish the two knobs: keep Combo timeout longer than Target timeout so failover still has time.

## Tests

- `tests/unit/combo-config.test.ts`: accept 1_200_000, reject `MAX_TIMER_TIMEOUT_MS + 1`, accept the key on combo defaults, and assert the editor/defaults pages mount the field.
- `tests/unit/ui/combo-timeout-fields.test.tsx`: seconds input writes milliseconds; emptying the field clears the key.
- Bug injection: drop the `<ComboTimeoutFields>` mount → the wiring test fails; restore → pass.

## Out of scope

Does not change `REQUEST_TIMEOUT_MS` or the hardcoded hang-stop constant. A combo that leaves Combo timeout empty still uses the 10-minute safety net.


HouMinXi:
Looked at `8522fe12a0` (the product commit) and the follow-up `1dbb89d1fe`.

The combo editor and Combo defaults now expose `comboTimeoutMs` next to Target timeout. Empty still means the 10-minute hang-stop; a positive value is the whole-combo wall-clock budget. Schema caps it the same way as `targetTimeoutMs`.

One leftover: the UI test title said the field writes seconds while the assertion was `1_200_000` (milliseconds). Renamed the test. Behavior unchanged. `tests/unit/ui/combo-timeout-fields.test.tsx` is the lock — it goes red if the input stores seconds instead of milliseconds.

Left alone on purpose:
- The editor wiring test reads `page.tsx` / `ComboDefaultsTab.tsx` so dropping the mount fails.
- HTML `max` is 86400 seconds, same as the existing Target timeout control. The schema cap is `MAX_TIMER_TIMEOUT_MS` (~24.8 days). API values above one day still display; the spinner cannot reproduce them. Same as Target timeout today.

PR #13880, opened 2026-09-16
fix(ci): heal release/v3.8.51 base-reds for agent-skills sync and reasoning-effort test (#13866)

HouMinXi:
Heals two inherited base-red defects on `release/v3.8.51` identified in tracking issue #13866 (#13866 stays open).

### Root Causes & Fixes

1. **Agent skills drift (`check:agent-skills-sync` exit 2):**
   - In #13770 (`694c1b74eb`), `mcp enable` and `mcp disable` subcommands were introduced to `bin/cli/commands/mcp.mjs`.
   - `skills/cli-mcp/SKILL.md` was not refreshed with the generator, causing the dry-run check in `Merge integrity` to detect stale documentation and exit 2.
   - Fixed by applying the canonical output from `node scripts/skills/generate-agent-skills.mjs --apply`.

2. **Reasoning effort clamp and retry test failure (5/5 failed):**
   - In #13798 (`2bbe6e575b`), `requireCompatibleBaseUrl` was added to prevent unhydrated `openai-compatible-*` connections from routing to public OpenAI endpoints.
   - `tests/unit/reasoning-effort-clamp-and-retry.test.ts` uses a custom `SimpleExecutor` with dummy provider name `openai-compatible-chat-eaff6869` and dispatched with empty credentials (`credentials: {}`), throwing `'provider node "openai-compatible-chat-eaff6869" has no baseUrl'`.
   - Fixed by hydrating `providerSpecificData.baseUrl` during `SimpleExecutor.execute()`.

### Verification

- `npm run check:agent-skills-sync` passes (0 errors, 46 skills up-to-date).
- `node --test --import tsx tests/unit/reasoning-effort-clamp-and-retry.test.ts` passes 5/5.
- `npm run typecheck:core` clean (0 errors).
- ESLint clean on all changed files.
- `node scripts/check/check-changelog-integrity.mjs --base upstream/release/v3.8.51` passes.

Issue #12958, opened 2026-09-07
fix(providers): GitLab Duo (gitlab.com): health check 403 "direct access scope is unavailable" even with valid seat + default Duo namespace


Rahulsharma0810:
### Summary
The `gitlab-duo` credential health check fails with 403 on gitlab.com, surfaced as **“GitLab Duo direct access scope is unavailable”**, for a token that was just issued by the OAuth flow, even though the user has Duo active and a Default GitLab Duo namespace set.

### Repro
1. Connect a gitlab-duo provider via OAuth (PKCE) against gitlab.com with scopes `ai_features read_user api`.
2. Token exchange succeeds (connected).
3. CredentialHealth probe to `POST https://gitlab.com/api/v4/code_suggestions/direct_access` returns **403**, `error_code = 403`, `last_error = Access denied`.
4. Inference routing (`_1j_edf1`) then returns `403 "GitLab Duo direct access scope is unavailable"` for every request — the graceful fallback branch (`shouldFallbackToPublicCodeSuggestions`) is **not** taken because the 403 body is not exactly `"direct connections are disabled"`.

### Observations
- The request is sent with only `Authorization: Bearer <token>` — **no `namespace_path`/`project_path` body**, so gitlab.com must resolve Duo entitlement purely from the profile’s Default GitLab Duo namespace. That namespace **is** set (verified: only namespace `i1-1806`), yet still 403.
- The raw 403 response body is never logged/surfaced anywhere in the UI (a generic “Access denied” is stored instead), which is why users cannot distinguish entitlement vs. instance-config vs. scope issues.
- GitLab’s `direct_access` endpoint is **rate-limited to 10 req / 5 min per user**. OmniRoute calls it on every inference (60s cache) plus every health probe plus token-refresh checks; overlapping relinks/health checks trip real HTTP **429s** that are reported confusingly as connection errors.

### Suggested fix
- Send the user’s default/configured Duo namespace (e.g. `namespace_path` from profile preferences or `provider_specific_data`) in the `direct_access` body so gitlab.com can resolve entitlement for API-only clients.
- When `direct_access` returns 403/429, log and surface the **full response body** (as GitLab’s own extension does) instead of the generic “Access denied” / “scope is unavailable”.
- Respect the 10-req/5-min rate limit explicitly (retry/backoff), rather than charing 429s as connection failures.

PR #13900, opened 2026-09-16
fix(api): escape the raw control bytes #13758 left in the provider-test sanitizer

diegosouzapw:
Heals a base-red on `release/v3.8.51` that reddens the **`Unit Tests fast-path (1/4)`** shard of **every** open PR.

### Root cause

`#13758` (commit `5ff85c6d`, the #12958 GitLab Duo fallback) introduced `sanitizeUpstreamBodyText` in `src/app/api/providers/[id]/test/route.ts` with the control-character range written as **literal raw bytes** instead of escapes:

```
/[\r\n\t<0x00>-<0x1f>]+/g      ← raw NUL and raw 0x1f in the source
```

The regex is byte-identical at runtime, so nothing misbehaved — but a raw NUL in tracked source trips `tests/unit/source-no-raw-nul-bytes.test.ts`, whose whole job is to keep NUL bytes out of `src/`, `open-sse/` and `tests/`. Byte counts confirm the origin: the file had **0** NUL bytes at `3e2a6d8f` (parent) and **1** at `5ff85c6d`.

### Fix

One line — the raw bytes become their `\x00-\x1f` escapes. No behavior change.

### Validation (Hard Rule #18)

Reproduced on the pure tip (`origin/release/v3.8.51`, no other change):

```
node --import tsx/esm --test tests/unit/source-no-raw-nul-bytes.test.ts
  ✖ no source file under src/, open-sse/ or tests/ contains a raw NUL byte
    actual: [ 'src/app/api/providers/[id]/test/route.ts:247' ]   expected: []
```

After the fix: **3/3 pass**. `tests/unit/issue-12958-gitlab-duo-403-entitlement-fallback.test.ts` still **2/2 pass** — the sanitizer's behavior is unchanged.

### Scope

Does **not** cover the other inherited reds currently on the tip — `providers-constants-split`, `provider-node-reserved-prefix` and `hard-session-lease-bypass-inventory` all fail on the pure tip too and belong to the provider-registry drift, not here. `#13880` covers the `cli-mcp` skill drift and the reasoning-effort test.

Refs #12958


diegosouzapw:
Fechando como subsumida. O tip já recebeu a correção equivalente pela #13947 (`0d089e7e`, "fix(quality): clear the release/v3.8.51 base-reds").

As duas regexes são a mesma faixa de caracteres:

```
esta PR:  /[\r\n\t\x00-\x1f]+/g
tip:      /[\r\n\t\u0000-\u001f]+/g
```

`\x00-\x1f` e `\u0000-\u001f` são idênticos. Como esta PR não traz teste e o fragmento de changelog dela não agrega sobre o que já entrou, não há o que portar.

PR #13223, opened 2026-09-10
fix(mcp): load audit sqlite via runtime helper

chatchawan-simplewish:
## Summary

- Load the MCP audit native SQLite driver through the existing standalone-safe runtimeRequire helper.
- Prevent emitted CommonJS bundles from invoking an invalid dynamically imported createRequire binding.
- Add a focused regression test for the runtime-loader path.

## Related Issues

- Related to the packaged MCP audit error: a is not a function.

## Validation

- [x] Change type: DB / build-deploy
- [x] Focused tests and category gates from the golden path
- [ ] npm run lint (full lint left to CI; scoped ESLint passed)
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR
- SonarQube is temporarily opt-in while the private project has no quota; it is not a PR gate.

Focused validation:
- vitest run open-sse/mcp-server/__tests__/audit.test.ts - 4/4 passed
- scoped ESLint on the two changed files - passed
- production backend build and emitted-runtime write/read were verified during deployment acceptance

## Tests Added Or Updated

- open-sse/mcp-server/__tests__/audit.test.ts

## Coverage Notes

- The focused audit test exercises native-driver loading through runtimeRequire, audit statistics mapping, shutdown, and the Node SQLite fallback.

## Reviewer Notes

- The change reuses the existing runtime loader; no new dependency, migration, or feature flag is introduced.
- The packaged fix has been live-verified with authenticated MCP audit write/read, healthy database integrity, and no audit-loader errors.

PR #13903, opened 2026-09-16
fix(mcp): fall back when better-sqlite3 export is not callable

HouMinXi:
## Bug Description

Packaged MCP audit logging dies on a minified TypeError (`a is not a function`) when `better-sqlite3` resolves to a non-callable export. The classifier did not treat that shape as a native load failure, so the open was not cached. Dashboard MCP status then retried the same failed load on every poll.

Related to #13223 (OPEN). That PR switches the native load to `runtimeRequire`. This PR classifies the TypeError, caches a failed open, and refuses to ship an image without `better_sqlite3.node`. Keep both if they still apply after rebase.

## Root Cause

Standalone webpack can hand `better-sqlite3` a stub or a non-function export. `new (mod.default || mod)(path)` then throws `TypeError: a is not a function` (the minified identifier is `a`). `isNativeSqliteLoadError` only matched DLOPEN / MODULE_NOT_FOUND / bindings-file text, so `getDb()` logged and returned without caching. `GET /api/mcp/status` and `GET /api/mcp/audit` call that path; the MCP page polls status every 30s.

## Fix

- Treat `is not a function` and `is not a constructor` as native load failures in `isNativeSqliteLoadError` (`src/lib/db/core.ts` and the bootstrap-env copy).
- If the export is not a function, throw that TypeError so the classifier fires and `getDb` falls back to `node:sqlite` on Node 22.5+.
- Cache a failed open as `null` (`undefined` means never tried). `closeAuditDb()` resets to `undefined` so a later open can retry.
- Dockerfile: `test -f` the `.node` addon in the builder rebuild step and again in the runtime image.

## How to Verify

1. `npx vitest run --config vitest.mcp.config.ts open-sse/mcp-server/__tests__/audit.test.ts` — 5 passed.
2. `node --import tsx/esm --test tests/unit/db-core-native-error.test.ts` — 12 passed (includes the minified TypeError cases).
3. `node tests/unit/build/dockerfile-better-sqlite3-native.test.mjs` — 1 passed (builder and runtime `test -f` both required).

## Test Plan

- [x] Added regression tests for the classifier, the failed-open cache, and the Dockerfile `.node` gates
- [x] Existing focused tests still pass (counts above)
- [ ] Image rebuild on a host that previously shipped without `better_sqlite3.node`

## Risk Assessment

Low. Classifier widening only matches those two TypeError substrings already produced by a non-callable export. Miss-cache is limited to the MCP audit connection. The Docker `test -f` fails the build if the addon is missing, which is the intended gate.

## Not in this PR

- Does not replace #13223's `runtimeRequire` load path.
- Does not change the MCP Streamable HTTP server or the SDK client import.


HouMinXi:
## CI split on 3fad92f344

`src/lib/db/core.ts` is no longer in the file-size failure list (tree count 1757, frozen 1788). The extract in this head is the unique part of that gate.

Fast Quality Gates on this run still fails four checks. None of the named files are in this PR's unique diff (`Dockerfile`, `audit.ts`, `core.ts`, `sqliteLoadError.ts`, tests, changelog):

- file-size: `chat.ts`, `chatHelpers.ts`, `chatCore.ts`, `proxyFetch.ts`, `stream.ts`, `batch_api.test.ts`
- mutation-test-coverage: `combo-identical-error-streak.test.ts`, `13601-header-drop-count-surfaced.test.ts`
- complexity-ratchets new-code: 138 files vs merge-base `1510fb0d` (this branch is behind current `release/v3.8.51`)
- open-sse-typecheck: `openai-responses.ts` TS2677

Compare: https://github.com/diegosouzapw/OmniRoute/compare/release/v3.8.51...HouMinXi:fix/mcp-audit-sqlite-fallback — ahead 3, behind 69. Tip `0d089e7e` is #13947 (clear the v3.8.51 base-reds), which already touches `openai-responses.ts` and the file-size baseline.

Rebase onto that tip is in progress. Unique tests stay the same after rebase.

HouMinXi:
## Rebased onto 0d089e7e (#13947)

Head is now `73541ab0a7`. Unique range is still three commits (range-diff all `=` vs the previous unique range). Compare against `release/v3.8.51`: ahead 3, behind 0.

Tree count for `src/lib/db/core.ts` is 1757 (frozen 1788). Unique tests after rebase: native-error 39/39, audit 5/5, dockerfile native 1/1. `check:file-size` on this tree: 0 violations (the previous chat/stream/chatCore overages were on the stale merge-base).

CI on this SHA is restarting.

HouMinXi:
Rebased onto `release/v3.8.51` (`2fa6ef0bdd`) and added one commit while reviewing the branch.

The native-load classifier had been copied into `scripts/build/bootstrap-env.mjs` next to the runtime one, but the two files read its verdict in opposite directions. In `src/lib/db/sqliteLoadError.ts` a true verdict means "the driver is unusable, cascade to `node:sqlite`", so matching a non-callable export is exactly right. In the bootstrap the verdict feeds `hasEncryptedCredentials`, where true means "no encrypted credentials found" and clears the way to generate a fresh `STORAGE_ENCRYPTION_KEY`.

So with the TypeError patterns present in the bootstrap copy, a binding that loads but exports something non-callable, over a database full of `enc:v1:` rows, reads as an empty database and the operator silently loses access to every stored credential. I dropped those two patterns from the bootstrap copy only and left a comment in both files explaining why the pair is deliberately not identical, so nobody harmonises them later. A corrupt binding there still throws, now carrying the database path, the underlying message, and a `npm rebuild better-sqlite3` hint, so the narrower classifier costs nothing in diagnosability.

Guard test `tests/unit/build/bootstrap-env-sqlite-classifier.test.mjs` pins the asymmetry in both directions plus the hint text and the throw-not-return contract. Injection, four arms, each confirmed red: re-widen the bootstrap classifier; narrow the runtime classifier; drop the hint; return `false` for a broken binding instead of throwing. Clean tree green before and after.

Two things I looked at and deliberately did not bundle:

- The `bun:sqlite` branch has no analogous hint. It is untouched by this change and `bun:sqlite` is a built-in with no native rebuild step, so the hint would not mean anything there.
- `code === "MODULE_NOT_FOUND"` still returns `false`, which is the intended fresh-install path (no driver, no database, generate a key). The severe case is a missing binding alongside an existing encrypted database. That mechanism predates this branch, and widening or narrowing it inside a data-loss fix seemed like the wrong place, so I left it as a follow-up.

Verification on the rebased tree: 21 node tests and 5 vitest audit tests pass, prettier and `node --check` clean, lockfile check OK, all four commits signed and semantically identical across the rebase (`git range-diff` reports `=` for each).

On review coverage: verdict PASS with two consecutive clean rounds, all five findings dismissed. A third round died mid-flight without advancing state, the same failure that hit two earlier attempts on this diff at the same point, so I adjudicated the findings against the source instead of relaunching. The two "the hint could fire on a SQL error" findings are disproved by enumeration: that try block contains only `require`, `new Database`, `prepare(<string literal>).get()`, and `close()`, with no user callback or dynamic dispatch, so every reachable source of `is not a function` inside it is a genuine broken-binding symptom, while real SQL failures (`no such table`, `SQLITE_CORRUPT`, `SQLITE_BUSY`, readonly write) stay loud and get no hint.

HouMinXi:
CI on this branch shows six failures. I reproduced each one on a pristine `release/v3.8.51` checkout (`2fa6ef0bdd`) with no commits from this branch applied, and they all appear there identically, so none of them come from this work.

**API Route Typecheck** and **Docs Gates (fast-path)** — on pristine base: `apiTypecheckErrors=294`, the same two regressions (`src/app/api/internal/codex-responses-ws/route.ts` TS2339 baseline 7 live 12, `src/sse/handlers/chat.ts` TS2345 baseline 0 live 1), and the same env contract break: `OMNIROUTE_STRIP_SYSTEM_PREAMBLE` is read in code but absent from `.env.example` and `docs/reference/ENVIRONMENT.md`. `git log -S` traces that variable to `b7192b72e2` (#12905); this branch does not reference it and does not touch either typecheck file.

**Unit Tests fast-path (1-4/4)** — the eight failures are in `tests/unit/compression/body-adapter.test.ts` (RTK compression and Responses custom-tool metadata) and `tests/unit/issue-7793-env-doc-sync-repro.test.ts`, which is the test form of the same missing env var. Running just those two files on pristine base gives 7 failures out of 12 tests. Neither file is in this branch's diff.

The tests this branch does own all pass on the rebased tree: 21 node tests across `bootstrap-env-sqlite-classifier`, `bootstrap-env`, and `db-core-native-error`, plus 5 vitest tests in the MCP audit suite. Prettier, `node --check`, and the lockfile check are clean.

Happy to rebase again once the base is green, if that helps separate signal from noise here.

Issue #13178, opened 2026-09-10
feat(backend): compression: make Lite's hardcoded MAX_TOOL_LENGTH (2K-char tool-output cap) configurable via env / settings


Nord1cWarr1or:
## Summary

The Lite engine's per-tool-output cap `MAX_TOOL_LENGTH` is hardcoded at `2000` chars and is not exposed anywhere (env, settings panel, engine config). `compressToolResults()` truncates **every** `tool`-role message above 2000 chars down to 2000 + `"\n...[truncated]"` suffix. For coding-agent workloads (large file reads, crash reports, log dumps arriving as tool results) this silently removes the middle of the payload, and there is no supported way to tune it.

## Current behavior (v3.8.50)

- `open-sse/services/compression/lite.ts`:
  - `const MAX_TOOL_LENGTH = 2000;`
  - `compressToolResults()`: `if (msg.content.length <= MAX_TOOL_LENGTH) return msg;` → else `content.slice(0, backOffToWordBoundary(...)) + "\n...[truncated]"`
- The `lite` engine settings only expose an on/off toggle (`compressToolResults`); no length knob exists in `rtkConfig`/`languageConfig`-style config or env.
- Observed on a 330-line (~12.1 KB) tool result: the model received only ~1.9K chars of it (self-reported "cut at line 37"). E2E via `/v1/chat/completions`: `prompt_tokens 972` for a payload that should be ~3100.
- Temporarily raising the constant to `5000` restored full delivery (`prompt_tokens → 5317`, model quoted the entire 330-line report), with the rest of the stacked pipeline unaffected — so the knob is safe to expose.
- Notably, the engine already has an `isDocumentLikeRead` fast-path that skips truncation for document-like reads — but it does not cover outputs carrying `Error:`/`Exception:`/`Traceback:` markers, so crash reports and stack-trace dumps still get head/tail-truncated mid-data.

Also a deployment note: this constant ends up duplicated across ~30 compiled bundle copies (Turbopack chunks + worker + SSR variants), so a user-side hot-patch to raise it must be re-applied ~30× after every `npm i -g omniroute` update. An official knob makes that entire class of hacks unnecessary.

## Feature request

1. Env override, e.g. `OMNIROUTE_LITE_MAX_TOOL_LENGTH` (read where `MAX_TOOL_LENGTH` is declared), default `2000` — zero behavior change for existing installs.
2. Optionally surface it next to `compressToolResults` in the Settings UI.
[...]

PR #13915, opened 2026-09-16
feat(compression): make Lite tool-result truncation length configurable

HouMinXi:
## Summary

Lite truncated tool results at a hardcoded 2000 characters. File reads and crash dumps arriving as tool payloads lost the middle of the content, and there was no supported way to raise the cap.

This adds `lite.maxToolLength` next to the existing `compressToolResults` toggle. Resolution order:

1. explicit option / step config / dashboard setting
2. `OMNIROUTE_LITE_MAX_TOOL_LENGTH`
3. 2000

Default stays 2000, so existing installs do not change. The truncation-allowlist idea stays a follow-up.

Related to #13178.
#13178 stays open.

## Test plan

- [x] `compressToolResults` honors an explicit cap below 2000
- [x] a 1500-char payload is kept when the cap is 4000
- [x] `OMNIROUTE_LITE_MAX_TOOL_LENGTH` is read when no option is passed
- [x] an explicit option wins over the env fallback
- [x] settings round-trip `maxToolLength`
- [x] schema accepts an integer cap and rejects a string
- [x] Lite engine schema exposes `maxToolLength` (default 2000, min 256)
- [x] dashboard Save PUTs `{ compressToolResults, maxToolLength }` when the operator set a cap
- [x] dashboard Save omits `maxToolLength` when settings never stored one, and when the number input is cleared
- [x] defect injection: hardcoding 2000 again fails the two new truncation tests; restoring the resolver passes


HouMinXi:
Inline review of `af4a921e90` found three real issues. They are fixed in `0e0d775964`.

1. A pipeline step with `maxToolLength: 10` (or `NaN` / `Infinity`) is still a number, so it hid a valid global cap of 8000 and fell through to env. A 3000-character tool result was truncated at env 400 instead of kept. Only an in-range integer now overrides the global cap.

2. `PUT { lite: { compressToolResults: false } }` replaced the whole lite JSON row, so a stored 8000 disappeared. Confirmed on a temp sqlite. Omit the key to keep the stored cap; send `null` to clear it.

3. Save treated an out-of-range number (`100`) like a cleared field and omitted it, which also wiped the stored cap. Save now refuses and does not PUT.

Copy that still said "over 2,000 characters" is updated. Lite Preview still posts a user message, so it cannot show tool truncation. That is existing preview scope, not this change.

Tests: node 65/65, vitest 15/15. Injected the three defects at the call sites; each test failed, then passed after restore.

Related to #13178. #13178 stays open.

HouMinXi:
Follow-up on the last review of `0e0d775964`.

The three runtime issues stayed fixed. Two leftovers were still real:

1. The dashboard still said "over 2,000 characters". Schema copy had changed, but the page overlays `src/i18n/messages` when that key exists, so operators never saw the schema string. All 66 catalogs now point at the Maximum tool-result length field / env.

2. The out-of-range Save test matched `/save/i`, which is the Save button. It now looks for "Failed to save configuration."

Lite Preview still posts a user message and does not send `config.lite`, so it cannot show tool truncation. That is existing preview scope.

Related to #13178. #13178 stays open.

HouMinXi:
Third look at `4caf3a40a6`.

The runtime cap, the settings merge, the dashboard copy, and the out-of-range Save error all still match the earlier follow-ups. No new defects.

Lite Preview still posts a user message and does not send `config.lite`, so it cannot show tool truncation. Same as before; not this change.

Related to #13178. #13178 stays open.

HouMinXi:
Fourth look at `4caf3a40a6`.

Same SHA as the third look. Runtime cap, settings merge, dashboard copy, and out-of-range Save still hold. No new defects.

Lite Preview still posts a user message and does not send `config.lite`. Same as before; not this change.

Related to #13178. #13178 stays open.

HouMinXi:
Fifth look at `4caf3a40a6`.

Same SHA as the third and fourth looks. Runtime cap, settings merge, dashboard copy, and out-of-range Save still hold. No new defects.

Lite Preview still posts a user message and does not send `config.lite`. Same as before; not this change.

Related to #13178. #13178 stays open.

Issue #13460, opened 2026-09-12
feat(providers): A subscription's tunnelled nodes all register as one proxy, so its pool has a single egress


maxmad64bis:
## Problem / Use Case

Paste a subscription link holding 300 nodes, bind it to a pool, pick a rotation strategy — and every request still leaves by the same address. Nothing fails and nothing warns: the pool page reports 300 nodes while every request leaves by the same address, so the rotation strategy on screen has nothing to rotate over.

It lands hardest on the users this feature exists for. `docs/ops/PROXY_GUIDE.md` opens on developers in Russia, China, Iran, Cuba and Turkey — and a network that blocks plain http/socks5 is exactly where people run vless, vmess or shadowsocks instead, which are the nodes that collapse into a single entry. The operators with the fewest alternatives get the degenerate case, silently.

Per #9611, serving one account per route needs one distinct route per account. A subscription can't supply that today, however many nodes it carries.

The mechanism, for the record: tunnelled nodes are registered as **one** row for the whole set, pointing at `localCoreEndpoint` (`subscriptionService.ts:433-450`), while http/socks5 nodes each get their own (`:417-431`). So the cursor rotates over one member, under a UI that offers "a rotation strategy that cycles egress IPs" (`ProxyRegistryManager.tsx:67-68`).

## Proposed Solution

What an operator would get: one pasted link becomes a pool with as many real exits as it has working nodes — no commercial rotating proxy, no hand-written core config, and the pool page finally means what it says.

Underneath the node count, this is about who decides the route. Everything tunnelled goes into one local port today, and the core alone picks which node the request leaves by — so the pool, the rotation strategies and the per-IP cooldown all reason about a choice the app never made. Plural entries hand that decision back to OmniRoute, which is what makes every other proxy feature mean something on a subscription.

Mechanically, a subscription declares several loopback entries instead of one, and each pool member binds to its own. The existing gate (`coreEndpoint.ts:9-32`) applies unchanged per entry, and a single entry behaves exactly as today, so nobody has to migrate anything.

This adds no capacity anywhere — it stops the app from presenting capacity that isn't there.

I'm after a direction before writing code, not a spec, and I'm happy to shape it however you prefer.

## Alternatives Considered

Hand-building a multi-entry core config and adding each local port as a manual proxy. That's the current workaround; it drifts on every subscription refresh and duplicates a node list the app already holds.

## Acceptance Criteria

- One entry: identical behavior to today, no config change required.
[...]

PR #7299, opened 2026-07-15
feat(proxy): operator-level proxy subscriptions (Karing-style) — hardened, ready for review

xier2012:
## Operator-level proxy subscriptions (Karing-style)

The operator pastes a subscription URL; OmniRoute parses + syncs the nodes into
`proxy_registry` (tagged `source='subscription', subscription_id=…`) and binds
the pool through the **existing** scope resolution (`global` or selected
`provider` scopes). Rotation, health checks, and the fail-closed guard are all
inherited from the current pipeline — **no new routing surface**.

> This PR started as the v1 feature and has since been hardened across
> **13 review-driven iterations** (UX/i18n → P0 security → P1 robustness →
> P2 i18n error codes → P3 enhancements). It is now **ready for maintainer
> review**. Summary of everything below.

---

### v1 scope
- **Modes**: `global` (all traffic) or `rule` (only selected providers).
- **Protocols**: `http/https/socks5` + edge relay go straight in;
  `ss/vmess/vless/trojan/hysteria/tuic/wireguard` need a local core
  (`sing-box`/`clash` exposing SOCKS5/HTTP — operator-supplied endpoint).
- **Per-subscription toggle**, **status** (`ok`/`error`/`empty`), **redacted
  node summary**, **auto-refresh** on a configurable interval.
- **Multi-tenant per-API-key overrides: explicitly out of scope.**

### Architecture (why this is small)
Subscription nodes are upserted into `proxy_registry` and bound via the existing
scope-pool API. So existing rotation, health checks, fail-closed guards, and
dashboards (`ProxyPoolTab`, `GlobalConfigTab`) all work unchanged — subscription
nodes just appear with a `source` badge. Disabling/deleting a subscription
cleanly removes only its own bindings.

---

## Security hardening (P0)
- **Input-side SSRF guard** (`fetchGuard.ts`): the subscription **fetch** URL is
  validated before every request — scheme ∈ {http,https}; IP literals rejected if
  loopback / private / link-local (incl. `169.254.x` cloud metadata) /
  unspecified. `fetch` uses `redirect:"manual"` and re-validates the `Location`
  target on each hop (max one redirect), so a public URL can't 302 into an
  internal address.
- **DNS multi-record guard (P3-1)**: hostnames are resolved with
  `dns.lookup(host, { all:true })` and the fetch is rejected if **any** resolved
  A/AAAA address is internal — closes the "internal address on the 2nd A record"
  bypass.
[...]

Issue #2331, opened 2026-05-17
[BUG] GPT-5.5 Codex suffix aliases are overridden by client default reasoning.effort


terence71-glitch:
### OmniRoute Version

3.8.0

### Installation Method

Docker / Docker Compose

### Operating System

Windows

### OS Version

Windows 11 25H2

### Provider(s) Involved

Codex

### Model(s) Involved

gpt-5.5

### Client Tool

Opencode

### Description

OmniRoute exposes Codex model aliases such as:
cx/gpt-5.5-xhigh
cx/gpt-5.5-high
cx/gpt-5.5-medium
cx/gpt-5.5-low
These aliases appear to encode the intended reasoning effort in the model ID suffix.
We verified through OmniRoute detailed/pipeline logs that OpenCode automatically appends a default reasoning effort of medium to GPT-5-family requests. For example, even when selecting:
cx/gpt-5.5-xhigh
OpenCode sends:
```
{
  "model": "cx/gpt-5.5-xhigh",
  "reasoning": {
    "effort": "medium"
  }
}
```
OmniRoute currently gives request-level reasoning priority over the model suffix. As a result, cx/gpt-5.5-xhigh is sent upstream with medium effort instead of xhigh.
This makes the -xhigh / -high / -medium / -low suffix aliases ineffective for OpenCode and any other client that sends a default reasoning effort.

OpenCode automatically appends medium reasoning to GPT-5-family requests. This means selecting cx/gpt-5.5-xhigh in OpenCode does not actually result in xhigh reasoning unless the user also configures OpenCode to explicitly send xhigh.
The model alias name suggests that effort is controlled by the suffix, but in practice it is overridden by OpenCode’s default request parameter.

### Steps to Reproduce

1. Configure OpenCode to use OmniRoute with a Codex suffix alias, for example:
```
      {
     "provider": {
       "omniroute": {
         "models": {
           "cx/gpt-5.5-high": {
             "name": "cx/gpt-5.5-xhigh"
           }
         }
       }
     }
   }
```
2. Send a request from OpenCode.
3. Enable OmniRoute detailed/pipeline logs.
4. Inspect the client raw request. It contains:
```
  {
     model: cx/gpt-5.5-xhigh,
     reasoning: {
       effort: medium
     }
  }
```
5. Inspect the provider request sent upstream.

### Expected Behavior

Because the selected model alias is:
cx/gpt-5.5-xhigh
[...]

PR #13923, opened 2026-09-16
feat(proxy): support multiple local core endpoints, one per line

maxmad64bis:
## Summary

A subscription holding tunnelled nodes registers them as one pool member, so rotation has nothing to rotate over. This change lets an operator list several loopback core entries (one per line) and syncs each entry into its own registry row and pool member. A single entry behaves exactly as today.

On the two open questions in #13460: the delimiter is newline (one entry per line in a textarea) — a comma is legal inside userinfo (`socks5://user,name:pass@127.0.0.1:1080` parses today), so splitting on commas would break a working entry. On the error shape: neither a new subscription-level code nor a per-member `error`/`dead` through the health check, but the existing `LOCAL_CORE_ENDPOINT_INVALID` carrying a redacted per-entry `detail`. The health-check branch is inert by default — `resolveHealthCheckStatusWrite` returns null without `PROXY_HEALTH_AUTO_DEACTIVATE`, and policy C in `decision.ts` never mutates a status — so an unreachable member would never show `error`/`dead` on a default install; the proposed signal would have had no producer.

Before: single invalid entry → `NEEDS_CORE_NOT_CONFIGURED` + node list. After: single invalid entry → `LOCAL_CORE_ENDPOINT_INVALID` + redacted entry; several entries with one rejected → `LOCAL_CORE_ENDPOINT_INVALID` + rejected entries while the valid ones keep syncing.

## Related Issues

- Related to #13460
- Related to #7299
- Related to #13581

## Validation

- [x] Change type: provider
- [x] `npm run lint`

## Tests Added Or Updated

- `tests/unit/proxySubscription.coreEndpoint.test.ts` — parse (multi-line, CRLF, blanks, comma in userinfo kept intact, `null`/`""`), redact (userinfo, malformed input, path segment kept).
- `tests/unit/proxySubscription.pluralCore.test.ts` — new file: single-entry identical behavior, N entries into N rows visited by rotation, gate-rejected vs unreachable distinguished in `detail`, re-sync keeps row id and assignments, entry removed from the field deletes its row, scheme collision on one key keeps one row with a mention, `detail` carries no password or userinfo.

## Coverage Notes

- `src/lib/proxySubscription/coreEndpoint.ts` — covered by the extended unit table (parse + redact + gate).
- `src/lib/proxySubscription/subscriptionService.ts` — covered by the new sync suite over real loopback TCP servers and closed ports (no ESM stubs): per-entry gate, dedup, informative probe (`Promise.all`, verdict feeds `detail` only), conditional naming, stale-removal unchanged. #13581 (pool egress observation) is the post-merge check that the pool finally serves distinct exits.
- `SubscriptionTab.tsx` + `en`/`vi` strings — textarea swap only, no logic; placeholder/desc updated.

## Reviewer Notes

- Sync probes the normalized URL (row's effective port: `https` → 443 else 8080), not the raw entry, because the probe's own scheme default (`socks5` → 1080) disagrees with the row rule on port-less entries.
- Probes run under `Promise.all` (independent TCP reads); upserts stay sequential. Verdict feeds the warning `detail` only — a row is created even when its core is unreachable, so a transient outage never deletes a row and its operator assignments. This is safe at selection time since #13578: a failing entry is skipped on its own at request time.
- No new error code and no migration: the field stays a string, `LOCAL_CORE_ENDPOINT_INVALID` carries the redacted per-entry `detail`, one entry keeps the historic name and behavior.
- Limits: a TCP-open non-proxy squatter still yields a false-healthy probe (same as any existing TCP check); removing userinfo from the field is a new identity under the `(host, port, username)` keying, so the old row is replaced like any key change.
- CI `API Route Typecheck` red is inherited, not this PR: `src/lib/providerModels/modelDiscovery.ts TS2345 (baseline 0, live 1)` reproduces on the bare base with zero commits of this PR (ephemeral worktree at `a928ea8762`; PR rebased on `f1e7148c19` with focused checks re-run green), introduced upstream without a rebaseline. Unit + gates qualified 2026-09-16 (§12: normalized `✖` lists vs same-base PR #13919 + local re-runs at the head): `codex #2331` (`getForcedReasoningEffort` chain, tip codex commit) and `settings-i18n` (`ReasoningRoutingRules` missing EN, tip reasoning commit) both fail locally with zero PR files involved; no `✗` gate line cites a PR file — all reds inherited, none actionable in this PR.


diegosouzapw:

PR #13934, opened 2026-09-17
fix(test): allow the manual-entry guard in the OAuth loopback source assertion

diegosouzapw:
## What broke

`tests/unit/oauth-lan-loopback-guidance.test.ts` → *"OAuthModal renders the structured panel instead of the generic error step"* has been failing on `release/v3.8.51` since #9944 landed.

It is a source-regex test. It asserted:

```js
/else if \(isLocalhost\) \{[\s\S]{0,300}buildPkceLoopbackMismatchHint/
```

#9944 added an explicit manual-entry bypass to that arm, so the condition now reads `isLocalhost && !opts?.manualLoopback` (`src/shared/components/OAuthModal.tsx:535`) — choosing "Enter URL manually" skips the loopback warning for that one attempt, which is the whole point of that PR.

## Why this is a test fix and not a code fix

The behaviour the test guards is intact. The `isLocalhost` arm still calls `buildPkceLoopbackMismatchHint(provider, loopbackLocation)` and still does `setStep("loopback-mismatch")` — the structured panel, not the generic red error step. What the regex pinned was the *literal shape of the guard*, which is not what the test name claims to assert.

So the regex is widened to `/else if \(isLocalhost[^)]*\) \{…/` and a comment records why, naming #9944.

## Proof this does not mask the contract

Mutation check — replaced the hint call with a generic `setError` and re-ran:

```
ℹ tests 11 · ℹ pass 10 · ℹ fail 1
AssertionError: the isLocalhost arm of PKCE_CALLBACK_SERVER_PROVIDERS must build the structured hint
```

The assertion still fires when the real behaviour is gone. Source restored afterwards; this PR touches one test file and nothing else.

With the fix: **11/11 pass**.

## How it was found

Bisected across the release tip while auditing an unrelated PR:

| commit | result |
|---|---|
| `f1e7148c` (before #9944) | ✅ passes |
| `a2c6f718` (after #9944/#9908/#7138) | ✖ fails |
| `5acac802` (current tip) | ✖ fails |

## Two other base-reds, NOT addressed here

The same bisect shows these already failing at `f1e7148c`, before that batch — they predate this and are out of scope:

- `#2331 codex.ts still prioritizes modelEffort first in rawEffort chain`
- `direct translation calls have English messages`


diegosouzapw:
Good catch and a clean mutation-tested fix, but this landed already — the same regex widening (same rationale, #9944's manualLoopback bypass) merged via #13947 while this branch was in flight. Closing as subsumed; no rework needed on your end.

PR #4105, opened 2026-06-17
fix(responses): clear /v1/responses keepalive timer on cancel/abort (timer + CPU leak)

artickc:
## Problem
`createResponsesApiTransformStream` (used by the `/v1/responses` / Codex path) starts a 3s keepalive `setInterval` in `start()` but clears it **only in `flush()`**. `flush()` runs only when the writable side closes *normally*. When the client disconnects mid-stream the writable side is **aborted**, so `flush()` never runs and the interval is orphaned.

The orphaned interval keeps firing every 3s and calls `controller.enqueue()` on a now-closed controller, which **throws on every tick forever**. Each aborted `/v1/responses` stream therefore leaks one live timer (plus the retained transformer `state` closure). On a long-running proxy these accumulate, steadily increasing CPU usage and memory — a 1-core-pinned / sluggish backend after hours of use.

Every sibling per-stream timer in the codebase already guards against this (`sseHeartbeat.ts`, `earlyStreamKeepalive.ts`, `stream.ts` idle watchdog clear on abort/cancel + enqueue-failure and `unref()`); only this keepalive was missing it.

## Fix
- Wrap the keepalive `enqueue` in try/catch and **self-clear** the interval on the first failed enqueue (backstop for any transport where neither `flush` nor `cancel` runs).
- Add a transformer `cancel()` handler that clears the timer on stream abort/cancel.
- `unref()` the timer so it cannot keep the event loop alive on its own.

No behavior change on the normal (completed) path — the existing `flush()` clear is unchanged.

## Tests
Added two regression tests to `tests/unit/responses-transformer.test.ts`:
- timer is cleared when the readable side is cancelled (client disconnect),
- keepalive self-clears when `enqueue` throws on a torn-down controller.

All 10 tests in the file pass.

Issue #13935, opened 2026-09-17
[BUG]  OpenCode free  provider returns 403 for all API requests through OmniRoute


stormsia:
## https://github.com/diegosouzapw/OmniRoute/issues/13935#issuecomment-5725013266

> [!NOTE]
> **Update:** The first 12 hexadecimal characters of the session ID are time-ordered data derived from the timestamp and an internal counter. The remaining characters are random.
>
> OpenCode generates session IDs in:
>
> - [`identifier.ts`](https://github.com/anomalyco/opencode/blob/dev/packages/schema/src/identifier.ts)
> - [`session-id.ts`](https://github.com/anomalyco/opencode/blob/dev/packages/schema/src/session-id.ts)

### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Windows

### OS Version

10

### Provider(s) Involved

OpenCode Free 

### Model(s) Involved

all

### Client Tool

OpenCode app| Omni route Playground| Any api call

### Description

All requests to the OpenCode Free provider fail with the same HTTP 403 error, regardless of which OmniRoute client is used.

The issue is reproducible in both of the following scenarios:

1. OmniRoute Playground → OpenCode Free provider
2. OpenCode app → OmniRoute using an OmniRoute API key → OpenCode Free provider

The failure is not limited to a specific model. All tested models return the same error.

This indicates that the issue is likely related to the OpenCode Free provider integration or the way OmniRoute authenticates and forwards requests, rather than to the OpenCode app itself.

### Steps to Reproduce

1. Start OmniRoute.
2. Configure the OpenCode Free provider.
3. Open OmniRoute Playground.
4. Send a request using any OpenCode Free model.
5. Observe the HTTP 403 error.
6. Configure the OpenCode app to use OmniRoute with an OmniRoute API key.
7. Send a request using any OpenCode Free model.
8. Observe the same HTTP 403 error.

### Expected Behavior

The request should be accepted and routed successfully through OmniRoute, or OmniRoute should clearly identify that the selected free-tier model cannot be proxied and prevent the request with an actionable configuration message.

### Actual Behavior

[Error: [403]: Error from provider (Console): OpenCode's free tier can only be used from within OpenCode] 
Every API request fails with HTTP 403, regardless of the selected model or client.

### Test Impact

Unsure

### Error Logs / Output

```shell
[Error: [403]: Error from provider (Console): OpenCode's free tier can only be used from within OpenCode]
```

[...]

PR #13937, opened 2026-09-17
fix(opencode): send canonical free-tier identity headers

AStupidBear:
## Summary

OpenCode Zen free-tier requests now require the current OpenCode CLI identity contract. This change aligns OmniRoute with the behavior documented by [9router Issue #4101](https://github.com/decolua/9router/issues/4101) and implemented in [9router PR #4105](https://github.com/decolua/9router/pull/4105).

- use a versioned `opencode/1.18.31` User-Agent, while preserving valid newer OpenCode versions
- generate canonical `ses_` and `msg_` identifiers and deterministically translate foreign session IDs
- send `Authorization: Bearer public` for anonymous `opencode` / `opencode-zen` Zen requests without changing `opencode-go` key handling
- keep conversation sessions stable across retries and remove the Muse Responses UUID overwrite

## Validation

- targeted OpenCode unit tests: 101 passed
- `npm run typecheck:core`
- changelog integrity and public-credentials checks
- live OpenCode Zen request with canonical headers: HTTP 200

Reference URLs:
- https://github.com/decolua/9router/issues/4101
- https://github.com/decolua/9router/pull/4105


linuxtopia:
Header spoofing is already in place. The next step should not be further header changes, but using the same Bun/TLS/HTTP/2 transport as the official OpenCode CLI and validating it with a one-to-one integration test. In addition, FreeTierError 403 responses must not be marked as successful; they should trigger the appropriate retry and cooldown logic.

This looks like the right direction, and the live 200 OK validation is a strong signal that the canonical identity contract is the missing piece.

One thing I think should be fixed before merging: there is still a Muse Responses-specific path in open-sse/executors/opencode.ts that overwrites x-opencode-session with randomUUID():

```
if (
  this._requestFormat === "openai-responses" &&
  model.startsWith("muse-spark") &&
  !/^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[8-9ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(
    headers["x-opencode-session"] || ""
  )
) {
  headers["x-opencode-session"] = randomUUID();
}
```


That seems to conflict with the canonical session contract introduced by this PR. A UUID is not a valid ses_[0-9a-f]{12}[0-9A-Za-z]{14} OpenCode session ID.

If Muse Responses really requires a different session format, I think that behavior should be verified against the current upstream OpenCode implementation rather than falling back to a UUID. Otherwise this path can undo the fix and still produce a FreeTierError for muse-spark-*.

Would it make sense to remove this UUID overwrite and let normalizeSessionId() always produce the canonical OpenCode session format?

Also, it would be useful to add a regression test specifically covering:

- muse-spark-* + openai-responses

- no client-provided session

- generated x-opencode-session matches ^ses_[0-9a-f]{12}[0-9A-Za-z]{14}$

- the generated session is not replaced by a UUID

Apart from that, the canonical ses_/msg_ generation and versioned User-Agent changes look consistent with the 9router findings.

stormsia:
> [!NOTE]
> **Update:** The first 12 hexadecimal characters of the session ID are time-ordered data derived from the timestamp and an internal counter. The remaining characters are random.
>
> OpenCode generates session IDs in:
>
> - [`identifier.ts`](https://github.com/anomalyco/opencode/blob/dev/packages/schema/src/identifier.ts)
> - [`session-id.ts`](https://github.com/anomalyco/opencode/blob/dev/packages/schema/src/session-id.ts)

maxmad64bis:
Heads-up since #14013 now carries part of this work: it keeps your session-shape regex and your minimum-version rule, and you are credited with `Co-authored-by` on the commit plus a line in the body and the changelog fragment. The maintainer picks whichever of the two to land — I am not asking for this one to be closed.

What that PR adds on top of the identity headers is the request body, which the same gate also checks. Measured on the live endpoint across three models and both surfaces: `stream: true` and a non-empty `tools` array are required as well, so after the headers alone a request that carries no tools — Playground, title generation, compaction — still answers 403. That is also why the failure is compaction-specific for users of the official client in anomalyco/opencode#49433.

Two measurements that are useful to this PR if it is the one that lands:

- `Authorization: Bearer public` is not needed. An anonymous request with no authorization header at all answers 200 once the four conditions are met.
- The session id algorithm does not need to be reproduced. The upstream checks the shape and not the value: 12 arbitrary hex digits pass, so the timestamp-plus-counter derivation is not load-bearing.

diegosouzapw:
Thanks for the fast turnaround and the live 200 verification — the canonical id/UA work here is what got the whole #13935 investigation moving. The combined fix just merged as #14013 (squash `0349627c`), which carries your session regex and 1.17+ minimum-version rule and credits you with a `Co-authored-by` trailer, plus the request-body half of the contract (stream:true + non-empty tools) that the upstream also gates on. Closing this one with the credit preserved there. Appreciate it.

Issue #13691, opened 2026-09-15
fix(providers): devin-cli-agentic: SWE-2 models return <summary> envelopes as final content (finish_reason=stop), breaking agent tool loops


june8572-design:
## Summary

When using the `devin-cli-agentic` provider with Cognition **SWE-2** models (`swe-2`, `swe-2-medium`, `swe-2-high`, `swe-2-max`), multi-turn tool loops technically work, but the model's final assistant response is frequently an internal `<summary>...</summary>` report instead of user-facing text. OmniRoute forwards this verbatim with `finish_reason: stop`, so downstream agents (e.g. Telegram bots, Hermes-style tool loops) surface the internal progress report as the answer and stop.

## Environment

- OmniRoute: **v3.8.50** (Docker, `omniroute-custom:latest`)
- Devin CLI: **3000.10.21** (`devin acp` mode, `--agent-type summarizer`)
- Provider: `devin-cli-agentic`, model `swe-2-max`
- Client: OpenAI-compatible `/v1/chat/completions` with `tools` defined

## Reproduction

1. Register a `devin-cli-agentic` connection with a valid Devin CLI credential.
2. Route a tool-enabled chat completion (any agent loop: model emits `tool_calls`, client executes, sends `role: tool` results back).
3. Observe that after one or more successful tool-call rounds, the model returns a `<summary>` envelope as the assistant message with `finish_reason: stop`.

Example final assistant content received by the client (actual production response):

```
<summary>
Overview
사용자는 OKF 지식베이스를 검색해 Proxmox 서버부터 ... (a progress report, not an answer)

Key Details & Breadcrumbs
- ...

Current State
- 다음 단계는 ... (task not finished, but stream ended with finish_reason=stop)
</summary>
```

## Root cause analysis

`devin-cli-agentic` spawns `devin acp --agent-type summarizer`. The summarizer agent type exists to produce `<summary>`-formatted work reports, and SWE-2 in particular is heavily tuned toward emitting that envelope — even when the bridge prompt instructs it to respond with `<tool>` envelopes or plain final text.

This creates a structural conflict:

| Model (via `devin-cli-agentic`) | `<tool>` envelope | Final answer format |
|---|---|---|
| `swe-2`, `swe-2-high`, `swe-2-max` | ✅ works | ❌ wraps output in `<summary>` progress report |
| `claude-sonnet-4-6` | ✅ works | ✅ clean plain-text answer |
[...]

Issue #13600, opened 2026-09-13
feat(providers): Empty assistant response after tool_calls completion (104×, Gemini flash variants)


afonsoft:
**OmniRoute Version:** 3.8.50
**Installation Method:** npm (global)
**Operating System:** Linux
**OS Version:** Ubuntu / Linux 6.17.0-1020-oracle (arm64)
**Node.js Version:** v24.16.0
**Provider(s) Involved:** OpenRouter, TokenRouter, BAI
**Model(s) Involved:** `google/gemini-3.5-flash-lite` (70×), `google/gemini-3.6-flash` (22×), `google/gemini-3-flash-preview` (8×), `qwen3.8-27b` (3×), `z-ai/glm-5.3-free` (1×)
**Client Tool:** Any client routed through OmniRoute

## Description

After a model returns `tool_calls` and the client submits tool results, the streamed assistant response comes back **empty**: `[STREAM] Empty assistant response after tool_calls completion` — **104 warnings** in ~41 h.

Concentration on Gemini-flash variants via OpenRouter/TokenRouter suggests either (a) OmniRoute mishandles the post-tool stream continuation for these models, or (b) the upstream legitimately returns empty content and OmniRoute should auto-retry/failover rather than passing an empty completion to the client.

## Steps to Reproduce

1. Run an agentic client through OmniRoute to `openrouter:google/gemini-3.5-flash-lite`.
2. Let the model emit `tool_calls`, then submit results.
3. Observe `[STREAM] Empty assistant response after tool_calls completion` — the client receives an empty assistant turn.

## Expected Behavior

Post-tool-call continuations return actual assistant content; if the upstream legitimately returns empty, OmniRoute should retry/failover to an alternate rather than forwarding an empty response.

## Actual Behavior

Empty assistant responses reach the client (~104×/41h), degrading agentic flows — tools execute but the model never narrates/continues.

## Test Impact

Needs a new integration test — tool_calls → tool_result → streamed continuation returns non-empty assistant content for the affected Gemini variants.

## Error Logs / Output

```shell
[STREAM] Empty assistant response after tool_calls completion (openrouter:google/gemini-3.5-flash-lite) — sessionId=...   # ×70
[STREAM] Empty assistant response after tool_calls completion (openrouter:google/gemini-3.6-flash)   # ×22
[STREAM] Empty assistant response after tool_calls completion (tokenrouter:google/gemini-3-flash-preview)   # ×8
```

## Possible Solutions

[...]

PR #13865, opened 2026-09-16
fix(deps): restore npm 11 installs on main

HouMinXi:
Related to #13706. #13706 stays open until the rest of main-green is actually green.

`main` still has the global `brace-expansion: ^5.0.9` override plus the contradictory `minimatch → ^1.1.18` mapping. `release/v3.8.51` already dropped both (#11259). npm 11 fails on that tree (`Missing: brace-expansion@1.1.21 from lock file`); npm 10 does not, so a local npm 10 check misses it.

## What this PR actually contains

Five follow-up commits on top of the npm 11 lockfile refresh. Unique vs `main`:

- `package.json` / `package-lock.json`: drop the two brace-expansion overrides; refresh the lock. **96 `resolved` lines added, 73 removed**, not 24. Direct versions move with the refresh: `next` 16.3.1 → 16.3.3, `sharp` ^0.35.3 → ^0.35.4, `vitest` ^4.1.7 → ^4.1.11, `eslint-config-next` 16.3.1 → 16.3.3, plus new overrides (`browserslist`, `joi`, `csv-parse`) and a few other direct bumps (`qs`, `hono`, `fast-uri`, `adm-zip`, `js-yaml`). `next` / `sharp` match what release already ships.
- `src/lib/memory/embedding/transformersLocal.ts`: the dynamic import is assembled as `"@huggingface/" + "transformers"` and marked `/* webpackIgnore: true */` so Next does not resolve the optional package while compiling `/health`. Concatenated string literals can still be folded; the ignore comment is the load-bearing part. This is a production change, not lockfile hygiene.
- `electron/package.json` + lock: desktop advisory cleanup.
- `open-sse/utils/streamTiming.ts`: stop asserting wall-clock `setTimeout` against `Date.now`.
- Tests/setup for the optional transformers runtime.

Isolating only the first lockfile commit does not regenerate a lock that matches the four commits after it. Leaving the bumps in this PR is intentional: they converge with release rather than inventing a second main-only tree.

Supply chain on the new lock entries: all 96 resolve to `registry.npmjs.org`; no new package names.

This does not claim a full main-green pass. Other gates on #13706 still need to run.

PR #13954, opened 2026-09-17
feat(combo): add agentic-stall response classifier

patrykkopycinski:
## Summary

Adds a standalone **agentic-stall response classifier** for the combo failover path.

Some upstream providers answer a *mid-agentic-turn* request — `tools` present, conversation tail a tool result — with `finish_reason: stop` and a progress report as content, and **zero tool calls**. Agent harnesses read `stop` + no tool call as turn-over, so the client re-prompts, the model reports progress again, and the loop never converges. Every existing quality gate passes the response: it is a valid, non-empty completion.

This PR lands the **classification layer only**: pure functions that decide whether a 200 response is that stall shape. It adds no imports, no call sites, and changes no existing behaviour.

## Related Issues

- Related to #13691 (devin-cli-agentic: SWE-2 models return `<summary>` envelopes as final content with `finish_reason=stop`, breaking agent tool loops)
- Related to #13600 (empty assistant response after `tool_calls` completion — same "post-tool-call turn is not usable" class)

I did **not** claim `Closes #13691`: this PR classifies the stall but does not yet act on it. The classification has to land before a retry/failover caller can consume it. See *Reviewer Notes*.

## Validation

- [x] Change type: routing
- [x] Focused tests and category gates from the golden path — `bash scripts/quality/test-scoped.sh` → **66/66 pass**
- [ ] `npm run lint` — **cannot run in my environment**: ESLint 10.9.1 crashes loading `react/display-name` (`contextOrFilename.getFilename is not a function`) on *any* file, including pristine `origin/main` files. Repo/environment debt, not this diff. `prettier --check` passes.
- [x] Reconciled with the current active release base; focused checks rerun afterward — branched from `origin/main` @ `a3ca33fa6`, which was still the tip after fetch

### Classifier

| Condition | Requirement |
|---|---|
| Request shape | `tools` present **and** conversation tail is a tool result (OpenAI `role: "tool"`; Anthropic last user message carries `tool_result`) |
| Response shape | finished with `stop` / `end_turn` **and** contains no tool calls |
| Content shape | summary/narration block (`<summary>…`, `# Summary`, `Summary: …`) or short structureless prose reporting progress |

Three terminal protocols are exempt because their response is turn-over **by design**:

1. a genuine blocked-state report (`Blocked: … still OPEN`, `Cannot proceed …`),
2. an intentional-silence control token (`[SILENT]`, `NO_REPLY`, and the CJK equivalents) — an autonomous lane's "nothing to report" reply,
3. **a literal the request itself quoted as the required terminal reply** — a lane briefed with *"if nothing is worth saving, just say 'Nothing to save.' and stop"* is answered exactly that way by design; failing over re-runs the same deterministic turn.

Exemption 3 is deliberately narrow: verbatim, case-insensitive, whole-response match against a literal **present in the request**, under a length cap. A model drifting into a summary cannot satisfy it, because summarization text is not quoted in the brief. If the brief contains no literal, the function returns `false` and the normal path is unchanged.

## Tests Added Or Updated

- `tests/unit/agentic-stall.test.ts` — new, 66 tests

Covers: the request-shape gate in both API shapes, the narration heuristic, full classification across OpenAI/Anthropic × JSON/SSE, the positive controls (no tools pending / a real tool call / structured output must **not** classify), the kill switch, the clone-only contract on the response body, and the three exemption classes including their negative controls.

## Coverage Notes

- `open-sse/services/combo/agenticStall.ts` is covered by the new unit suite; there is no uncovered branch reachable from the exported API.
- **Mutation-tested** (the suite must fail when the logic breaks, not merely pass):

| Mutation | Result |
|---|---|
| silence exemption forced always-true | **killed** (9 failures) |
| instructed exemption forced always-true | **killed** (9 failures) |
| instructed exemption forced never-true | **killed** (6 failures) |
| length cap removed at all three enforcement layers | **killed** (2 failures) |

The cap is enforced at three points (the extraction filter, the whole-response check, and the extraction regex). Disabling any *single* layer is an equivalent mutant — the others still enforce it — which is intentional defence in depth; removing all three is what the tests catch.

## Reviewer Notes

- **This PR deliberately adds no call sites.** `open-sse/services/combo/combo.ts` is where the failover decision belongs, and that file has diverged substantially on active integration branches. Landing the classifier first keeps this change reviewable in isolation and avoids a large conflict-prone diff. A follow-up PR can wire it into the combo loop once a call site is agreed.
- **The failover must not retry blindly.** #13691 reports that re-prompting SWE-2 for a proper final answer frequently returns `<summary>` again. Failover to the *next combo member* is the remedy shape; retrying the same member is not.
- One divergence worth calling out for reviewers: exemption 2 mirrors the marker set an agent harness already treats as silence, so the two cannot drift on what counts as silence. Those markers are listed in `SILENCE_MARKERS` — if that list should be narrower or wider for OmniRoute, say so and I will adjust.


patrykkopycinski:
CI attribution — all 22 failures are pre-existing on the PR base, not this diff.

**This PR touches 2 files** (`open-sse/services/combo/agenticStall.ts`, `tests/unit/agentic-stall.test.ts`). It does not touch `package.json`, `package-lock.json`, or any workflow.

**Root cause of the 15 `check:node-runtime` failures (single cascade):**

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'tsx'
```
`node_modules` is never populated, so `node --import tsx scripts/check/check-supported-node-runtime.ts` cannot start. Every downstream job is then skipped.

**Why `node_modules` is empty — the install step is a false green.** `./.github/actions/npm-ci-retry` runs `npm ci` 3 times, all 3 fail, and the step still reports success:

```
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json or npm-shrinkwrap.json are in sync.
npm error Please update your lock file with `npm install` before continuing.
...
npm ci attempt 3/3 after transient failure
##[end-action id=__self.__run;outcome=success;conclusion=success;duration_ms=72497]
```

The retry loop's last line is `exit "$exit_code"` on attempt 3, so a genuine sync error is retried twice (20s + 40s) and then reported as a **pass**. The step's own `conclusion=success` is what hides it.

**Attribution — reproduced on the untouched base commit:**

| Check | Result |
| --- | --- |
| PR base | `a3ca33fa6` (= current `main` tip) |
| This PR touches lockfile/manifest/workflows | **no** — 2 source/test files only |
| `npx npm@11 ci --dry-run --ignore-scripts` on pristine `a3ca33fa6` | **RC=1** — "can only install packages when your package.json and package-lock.json … are in sync" |
| Same command with local npm 10.9.8 | RC=0 |

The failure is npm-major-specific (Node 24 ships npm 11; this machine has npm 10.9.8). Drift is in `packages/browser-pool`: manifest requires `playwright: 1.62.1`, and npm 11 rejects the tree while npm 10 accepts it.

**Already being fixed upstream:** #13865 (`fix(deps): restore npm 11 installs on main`) targets exactly this. This PR should go green once that lands; no duplicate lockfile change here.

**Other failures, also not this diff:** `Quality Gates (Extended)` (osv-scanner vulnerability ratchet), `Docs Sync (Strict)` (`check:docs-all`), and `Vitest`/`Bun SQLite`/`Quality Ratchet` — all in the same dependency-bootstrap blast radius or already red on base.

**What this PR does verify locally, on the same base commit:** the classifier suite passes through the repo's own gate, `bash scripts/quality/test-scoped.sh` → `66 pass / 0 fail`, plus `prettier --check` clean. `open-sse` typecheck reports 2 errors on pristine `a3ca33fa6` as well (pre-existing, baseline not widened).

PR #13955, opened 2026-09-17
fix(ci): parse main-green workflow YAML with js-yaml

HouMinXi:
## Summary

`scripts/quality/validate-release-green.mjs` imported the `yaml` package, which is not a direct dependency on `main`. The nightly main-green job then crashed before it printed the verdict fence, so the tracker comment was an empty code block.

This switches that import to `js-yaml` (`load` as `parseYaml`). `js-yaml` is already declared on `main`. The existing `extractCiGates` fixture still parses. A source assertion fails if the undeclared import comes back.

Does not touch lockfile overrides. Does not replace #13865.

Related to #13706
#13706 stays open

## Test plan

- [x] `node --import tsx/esm --test tests/unit/validate-release-green.test.ts` (32/32)
- [x] Inject `from "yaml"` → the new assertion FAIL; restore → PASS
- [x] `extractCiGates` still returns `lint` from the CI fixture


HouMinXi:
## CI on 7dffc0d448

This PR does not touch the lockfile. Unique files are `validate-release-green.mjs`, its unit test, and the changelog fragment.

Every job that runs `npm ci` dies on:

```
Missing: brace-expansion@1.1.21 from lock file
```

`npm-ci-retry` still reports success, then `check:node-runtime` fails with `Cannot find package 'tsx'` because the tree was never installed. That is the same main-branch install break tracked by #13706 / #13865. #13865's Lint job gets past `npm ci` and `check:node-runtime`.

This PR cannot go green on `main` until that lockfile companion lands. It only changes the empty main-green comment (undeclared `yaml` import). #13706 stays open.

Issue #12072, opened 2026-08-29
fix(providers): Tinycms Web signer installs permanent DOM shims, poisoning SSR with 500 / invalid JSON


Temp-tempp:
### OmniRoute Version

3.8.50

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11

### Node.js Version

22.23.2

### Provider(s) Involved

Tinycms Web

### Model(s) Involved

All

### Client Tool

Omniroute model test

### Description

Every time I attempt to use Tinycms Web it spams this error Failed to load interception settings: Unexpected token 'I', "Internal S"... is not valid JSON and then on refresh says Internal Server Error.

### Steps to Reproduce

1. Test any model
2. Get a bunch of errors and a Internal Server Error

### Expected Behavior

The models not crashing and spamming errors

### Actual Behavior

The models crash and spam errors

### Test Impact

Unsure

### Error Logs / Output

```shell

```

Temp-tempp:
Also it does let you prompt it once but it doesn't know it can read your code at all (Using Cline).

diegosouzapw:
There are two layers in what you're seeing, and separating them is what makes this actionable:

1. **The client-side symptom** — `Failed to load interception settings: Unexpected token 'I', "Internal S"… is not valid JSON` — means the dashboard called `GET /api/providers/tinycms-web/interception-rules` (`src/app/api/providers/[id]/interception-rules/route.ts`), the server answered with a plain-text `Internal Server Error` page, and the UI tried to `JSON.parse` it. The spammy message is a real (small) client defect: that fetch should guard non-JSON error bodies. But it's only the messenger.

2. **The actual defect is the server 500** — something in the Tinycms Web test/rules path is throwing. That part we can't pin down without your server log: could you paste the OmniRoute *server console* output (the terminal where `omniroute` runs, or set `APP_LOG_LEVEL=debug`) from the moment you click "Test" until the Internal Server Error? The stack line that precedes the 500 is exactly what's needed.

On the Cline note ("doesn't know it can read your code") — that's a separate concern: tinycms-web models are web-session-backed and don't support native tool calling, so agentic clients like Cline lose their file-tools with them; pick a tool-calling model for Cline.

Marking as needs-info for the server log — with that stack trace this becomes directly fixable.

diegosouzapw:
Found it — no server log needed.

[...]

PR #13957, opened 2026-09-17
fix(providers): keep TinyCMS DOM stub safe for Next.js SSR

aref-alapour:
﻿## Summary

Durable fix for the TinyCMS wasm-bindgen Node DOM shim poisoning the OmniRoute process so **every Next.js route returns Internal Server Error** after TinyCMS is used once.

The crash is:

```
TypeError: Cannot destructure property 'protocol' of 'window.location' as it is undefined
```

at Next.js SSR `getLocationOrigin()`. Running containers must be **rebuilt and restarted** to pick this up; a live poisoned process will keep 500ing until restart even after the image is updated.

## Root cause timeline

1. TinyCMS's signer is browser-targeted Rust compiled with wasm-bindgen. In Node, the generated glue expects `window`, `document`, `HTMLCanvasElement`, and `CanvasRenderingContext2D`.
2. OmniRoute historically installed those with `setupDomMocks()` doing **`g.window = g`** (alias `window` to the Node `global` object) **without installing `location`**.
3. `#12072` scoped install/restore around `initTinyCmsWasm()` / `generateSecurePayload()`, so a *successful* call no longer leaked `window` for the rest of the process lifetime.
4. That was not enough. `initTinyCmsWasm()` **awaits** WASM compile/instantiate **with the stubs still installed**. The await yields; concurrent Next.js SSR in the same Node process can observe a live `window`.
5. Next treats `typeof window !== "undefined"` as "we are in a browser", then `getLocationOrigin()` does `const { protocol, hostname, port } = window.location`. A `window` that is the Node global (or any stub) **without `location`** throws. Every subsequent dashboard/API SSR render is an Internal Server Error until the process restarts.

## Why it only appears after TinyCMS is used once

The shim is **not** a module-load side effect. Nothing installs `global.window` until the first TinyCMS request actually calls `initTinyCmsWasm()`. A freshly started OmniRoute process is healthy. After one TinyCMS challenge/sign, `window` exists on the process (at least for the duration of the WASM init await, and previously for the process lifetime). From that moment on, Next SSR on the same process is in the "browser" branch and crashes.

## Why bare `window = global` is toxic under Next SSR

Node is not a browser. Aliasing `window` to `global`:

- Makes `typeof window !== "undefined"` true for the whole process.
- Does **not** give `window.location` (`protocol` / `hostname` / `port`), which Next's `getLocationOrigin` destructures unconditionally once it thinks it is in a browser.
- Also left a fake `document.createElement()` that only returns a canvas, which is how `#12072` first showed up as a plain-text 500.

A leftover `window` without a Location-shaped object is enough to take down every route.

## What changed

Belt-and-suspenders in `open-sse/executors/tinycmsDomMocks.ts`, used by `tinycmsSigner.ts`:

1. **Never alias `window` to the Node global.** Install a dedicated stub object (`window !== global`).
2. **Always attach a Location-shaped stub** (`protocol`, `hostname`, `port`, `href`, `origin`, …) so even a live stub during the WASM `await` cannot crash `getLocationOrigin`.
3. **Restore after init/payload** via `withTinyCmsDomMocks` / `withTinyCmsDomMocksAsync` (keeps `#12072` leak protection).
4. **Heal leftovers:** if an older build already did `window === global`, replace that alias with the dedicated stub and do not put the toxic alias back on restore. If a leftover dedicated `window` remains, never strip `location` off it.

## Tests Added Or Updated

- `tests/unit/tinycms-dom-shim-ssr.test.ts` (new) — regression for no global alias, Location-shaped stub, async yield safety, no leak after init
- Existing `#12072` leak tests: `tests/unit/repro-12072-tinycms-dom-shim-leak.test.ts`
- Existing TinyCMS tests: `tests/unit/provider-tinycms-web.test.ts`, `tests/unit/tinycms-secure-nonce-randomness.test.ts`
- Focused suite: **29/29 pass**

## How to verify

```bash
  tests/unit/tinycms-dom-shim-ssr.test.ts \
  tests/unit/repro-12072-tinycms-dom-shim-leak.test.ts \
  tests/unit/provider-tinycms-web.test.ts
```

Manual (deployed instance that already 500s):

1. Rebuild and **restart** the OmniRoute container/process.
2. Hit any dashboard route — should render, not Internal Server Error.
3. Send one TinyCMS request, then hit dashboard again — must stay healthy.

## Reviewer Notes

- **Running containers need rebuild + restart.** Process-global bug; old Node process keeps 500ing until replaced.
- Does not change auth, security headers, or wipe data.
- Diff: 6 files, +416 / −103. Base `release/v3.8.51`. Head `aref-alapour:cursor/tinycms-dom-shim-044a`.

## Related Issues

- Related to #12072 (closes remaining Next SSR `location` hole during WASM init await)


diegosouzapw:
Thanks @aref-alapour — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

Issue #7764, opened 2026-07-19
fix: Weekly and Session utilization inconsistent position across quota cards


hichamza:
### OmniRoute Version

3.8.48

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### Description

<img width="796" height="528" alt="Image" src="https://github.com/user-attachments/assets/3391da4d-4239-4c60-b49f-46eb9c781b03" />

check attached to see the problem. the indicators are not located in the same position per card.

### Steps to Reproduce

go to provider quota with several different renewal date accounts. 

### Expected Behavior

session up top, week below

### Actual Behavior

can be anything

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

diegosouzapw:
Partial progress landed in `release/v3.8.49` via #7834, but **this issue stays open** — that PR does not fix what you screenshotted. Sharing the audit so the next pass starts from the right place:

1. `QuotaCardBody.tsx` (the only caller of `topQuotas()`, which #7834 hardened) is **not in the render tree**. `QuotaCard.tsx` renders `QuotaCardExpanded` exclusively, and that component handles both the collapsed and expanded states internally.
2. The live path already honors fixed ordering via `resolveQuotaDisplayOrder(providerId, quotas)` (added in #6722).
3. The real cause is narrower: `hasFixedQuotaOrder()` returns true only for `codex` and the GLM family. Every other provider falls through to `sortQuotasByRemaining()`, so the session/weekly rows get ordered by remaining-% and land in different positions per card — exactly the inconsistency in your screenshot.

So the fix needs to happen in the live path: give session/weekly a stable relative order for all providers (your "session up top, week below"), rather than sorting those two by remaining-%. Re-triaging with that scope.

diegosouzapw:
Confirmed this is still open after the #7834 partial fix. I reproduced it with a fresh TDD probe.

`hasFixedQuotaOrder()` (`quotaParsing.ts:19-22`) still only covers `codex` and the GLM family, so any other provider reporting both `session` and `weekly` windows — Grok, MiniMax, opencode/ollama-go — falls through to `sortQuotasByRemaining()`. Session and Weekly then swap position depending on whichever window currently has more headroom left, which is exactly what your screenshot shows.

[...]

PR #13962, opened 2026-09-17
fix(usage): read Anthropic's scoped weekly limits for Claude OAuth seats

dpozimski:
## Summary

- Anthropic's `GET /api/oauth/usage` now reports per-model weekly caps in a `limits[]` array (`kind: "weekly_scoped"` with `scope.model.display_name`), while every legacy `seven_day_<codename>` key comes back `null`. `getClaudeUsage` only read the legacy keys, so a scoped cap never reached the quota map. On a Claude Team seat captured on 2026-09-17 the Fable cap sat at 69% and was the active limit (`is_active: true`) while the shared weekly window showed 39%, and the dashboard, Provider Limits cache and fair-share saturation all saw only the 39%.
- `getClaudeUsage` now maps each `weekly_scoped` row into `quotas["weekly <model> (7d)"]` (percent used, reset time), the same key shape the legacy codename path produces, so the dashboard's Claude quota parser, the canonical window ordering and every consumer of the quota map pick it up with no further change. `session` and `weekly_all` rows are skipped because `five_hour` / `seven_day` already cover them; a legacy `seven_day_<codename>` value wins when both shapes name the same model; rows without a usable name or percent are ignored. The scope name falls back to `scope.model.id` and then `scope.surface`, lowercased to match the existing codename keys.

Captured payload shape (trimmed):

```json
{
  "five_hour": { "utilization": 37.0, "resets_at": "2026-09-17T10:30:00.674190+00:00" },
  "seven_day": { "utilization": 39.0, "resets_at": "2026-09-21T19:59:59.674212+00:00" },
  "seven_day_opus": null,
  "seven_day_sonnet": null,
  "limits": [
    { "kind": "session", "group": "session", "percent": 37, "scope": null },
    { "kind": "weekly_all", "group": "weekly", "percent": 39, "scope": null },
    { "kind": "weekly_scoped", "group": "weekly", "percent": 69, "resets_at": "2026-09-21T19:59:59.674396+00:00",
      "scope": { "model": { "id": null, "display_name": "Fable" }, "surface": null }, "is_active": true }
  ]
}
```

## Related Issues

- None found; `weekly_scoped` has no hit in the repo or the tracker.

## Validation

- [x] Change type: provider (usage fetcher only, no registry or executor change)
- [x] Focused tests and category gates from the golden path: `tests/unit/usage-claude-scoped-weekly-limits.test.ts`, `tests/unit/quota-saturation-anthropic-oauth.test.ts`, `tests/unit/usage-families-split.test.ts`, `tests/unit/repro-7764-collapsed-quota-order.test.ts` (35 tests, 0 failures)
- [x] `npm run lint` (prettier and eslint through the pre-commit hook, plus the docs-sync and any-budget checks)
- [x] Reconciled with the current active release base; rebased onto `release/v3.8.51` at b6373506

## Tests Added Or Updated

- `tests/unit/usage-claude-scoped-weekly-limits.test.ts` (new): scoped limit mapped into `weekly fable (7d)`; `session` / `weekly_all` rows ignored; legacy codename value kept when both shapes name one model; rows without a name or percent skipped, `scope.surface` used as the fallback name; payload without `limits[]` unchanged.

## Coverage Notes

- `open-sse/services/usage/claude.ts` gains one helper (`scopedLimitName`) and one loop inside the OAuth success branch; both are covered by the new test through `getClaudeUsage` with a stubbed `fetch`. Coverage in the file goes up.

## Reviewer Notes

- The new key is lowercase (`weekly fable (7d)`) to match the existing `weekly <codename> (7d)` keys; `MODEL_DISPLAY_NAMES` is not applied because `display_name` is already human readable.
- No behaviour change for payloads without `limits[]`, and a legacy `seven_day_<codename>` value still wins when both shapes are present, so nothing changes for accounts Anthropic has not migrated.
- The `limits[]` array also carries `severity` and `is_active`, which would let the dashboard show which window is currently binding; left out of this PR to keep the quota map shape unchanged.


dpozimski:
@diegosouzapw please CR

diegosouzapw:
Thanks for this — the diagnosis is right and the payload capture (Fable at 69% on the scoped window vs 39% on the shared weekly) is exactly the evidence we needed. Generalizing the `weekly_scoped` read to every model is the correct direction.

One change before this can merge: the new loop writes the scoped limits into `quotas`, which is the **account-wide routing** map. On the current tip, `open-sse/services/usage/claude.ts:136-137` already handles the Fable `weekly_scoped` case and deliberately stores it in a separate `modelQuotas` map — the comment there says why: *"Display-only model limits must not enter account-wide routing quotas."* A per-model ceiling landing in `quotas` would make the router treat the whole seat as exhausted when only one model is.

So the ask is: keep your generalized loop, but have it populate `modelQuotas` (extending the existing entry shape at :141) rather than `quotas`, and adjust the test at `tests/unit/usage-claude-scoped-weekly-limits.test.ts:73-80` to assert on `modelQuotas`. That also removes the duplication with the Fable-only block, which your loop can then replace.

Heads-up: #13888 (merge-ready, queued ahead of this one) restructures the same file into a new `claudeQuota.ts`. You'll likely want to rebase onto it once it lands to avoid a conflict — happy to help with that if it gets messy.

dpozimski:
Thanks, you're right about routing. Changes pushed:

- Rebased onto `release/v3.8.51` (18bbb101). I replaced the Fable-only block instead of adding a second loop: `FABLE_WEEKLY_LIMIT_SCHEMA` is now `WEEKLY_SCOPED_LIMIT_SCHEMA`, which accepts any model name and writes only into `modelQuotas`. `quotas` still only gets `session (5h)`, `weekly (7d)` and the legacy `seven_day_<codename>` keys.
- The name falls back from `model.display_name` to `model.id` to `surface`, lowercased. If a legacy `seven_day_<codename>` key already put the same `weekly <model> (7d)` entry in `quotas`, the scoped row is skipped. The dashboard merges `{ ...quotas, ...modelQuotas }`, so without that check the same model would show up twice.
- `usage-claude-scoped-weekly-limits.test.ts` now asserts on `modelQuotas` and checks that `quotas` stays at session/weekly, including a second scoped model named only by `model.id`. In `claude-fable-usage.test.ts` I changed one invalid-input case: `display_name: "Something else"` is a valid scoped model now and gets its own row, so that case uses a blank name instead.
- 40 node tests pass (scoped limits, Fable usage, saturation, families split, #7764 order), plus the vitest Fable rendering test.

#13888 is still open, so I rebased on the current release tip. Once it lands I'll rebase onto `claudeQuota.ts`. That should be a small move since the change is one schema and one loop.

dpozimski:
@diegosouzapw friendly reminder on this one: the `modelQuotas` change you asked for is in (details in my comment above), rebased on `release/v3.8.51`. The red checks are the same set #14740 fails on this base (tracked in #14547), nothing owned by this PR. Could you take another look when you get a chance?

PR #13984, opened 2026-09-17
fix(grok-cli): treat rolling 24h free-usage 429 as quota exhaustion

stormsia:
## Problem

Grok Build free-tier returns a 429 when the per-model rolling 24h token cap is hit (live body):

`You've used all the included free usage for model grok-4.6 for now. Usage resets over a rolling 24-hour window — tokens (actual/limit): 513161/500000.`

Before this change classify429 classified it as plain 
ate_limit, so combo waited ~30s (comboCooldownWait.maxWaitMs) and retried the **same drained grok-4.6 login** instead of treating it as quota exhaustion and advancing.

## Fix

- src/shared/utils/classify429.ts: new quota-exhaustion patterns for the rolling 24h free-usage message.
- open-sse/services/quotaTextCooldowns.ts: isRolling24hUsageLimitText + uildRolling24hQuotaFallback (24h cooldown, QUOTA_EXHAUSTED).
- open-sse/services/accountFallback.ts: wire the rolling-24h fallback into checkFallbackError next to the session-quota check.

Because grok-cli is passthroughModels, the result stays a model-only lockout for grok-4.6 on that login — sibling logins and other models remain eligible (asserted in tests).

## Tests

	ests/unit/grok-cli-free-usage-429.test.ts — RED before the fix (4 failing), GREEN after:

- classifier returns quota_exhausted for the exact body (string and JSON-wrapped)
- checkFallbackError → QUOTA_EXHAUSTED, 24h cooldown
- combo does not wait 30s (non-retryable)
- model parked on that login, connection stays active
- sibling grok-cli login stays eligible
- combo exhaustion does not skip sibling accounts


diegosouzapw:
Thanks @stormsia — merging via the release merge-train. Validated in local merge-train (mt-train8e) on the devbox @ train tip b38c9340f092c8852ecfa7f6e0f9592fd6f7fc35 with the 26 sibling PRs of the owner-approved file-size rebaseline batch: typecheck:core, file-size (rebaselined entry _rebaseline_2026_09_18_merge_train_8_frozen_growth), complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 722/722; vitest 479/482 where the 3 reds were 5s/20s timeouts under devbox load 18 and pass when re-run alone on the same train tip (flake class). Merged --admin per merge-gates §7.

Issue #6147, opened 2026-07-04
feat: routing/settings UX clarity — slider effective-share %, 'Cloud Sync' rename, built-in provider base_url override


diegosouzapw:
## Summary

Consolidated tracking issue for the concrete UX/clarity improvements the maintainer agreed to in discussion #4040 (structured feedback from @Nogg-aholic). Grouped here so the commitments survive the thread.

### 1. Weight sliders -- show effective share %
The 0-100 routing weight sliders are relative shares, but the UI lets the total reach e.g. 800%, which reads as absolute percentages and confuses users. Display the resulting share next to each slider (A=100, B=100, C=50 -> "~40% / ~40% / ~20%"). No change to the underlying weighted-random model -- purely a clarity affordance.

### 2. Rename "Cloud Sync" -> "Remote Settings Sync"
"Cloud Sync" implies a managed multi-tenant hosted service syncing self-hosted instances. It is actually a device-to-device settings sync via a server the user controls (off by default, no telemetry, no license server). Rename the settings section to "Remote Settings Sync" (or similar) to remove the managed-cloud implication.

### 3. Base-URL (+ auth) override for built-in providers
Custom "Compatible Node" providers already have a fully editable base URL per connection. Built-in providers (e.g. GitLawb) do not, so an upstream URL change requires a code fix. Add a base-URL (and where feasible, auth-scheme) override to the built-in connection edit form so users can hot-fix broken presets from the UI.

## Out of scope (tracked separately if reporter confirms)

- **GitLawb registry fix** -- awaiting exact upstream URL + error body from the reporter (`open-sse/config/providers/registry/gitlawb/index.ts`).
- **IBM Watson legacy-import edit/delete** -- 9router DB import surfaced as a `custom_providers` entry; verify Edit/Delete surfaces for imported entries.

## Source

- Discussion: #4040
- Areas: Dashboard routing settings UI, sync settings naming, provider connection edit form

Iammilansoni:
Hi @diegosouzapw, I'd love to take this on!

I recently merged the docs normalization PR (#6105 ) and I'm ready to dive into the dashboard UI. I can handle the React state calculations for the relative slider percentages, update the sync naming across the locales, and wire up the base-URL override in the provider edit form.

Could you assign this to me?

diegosouzapw:
[...]

Issue #5899, opened 2026-07-02
fix(providers): Api Airforce Model Fetch


alicomert:
### OmniRoute Version

3.8.38

### Installation Method

npm (global)

### Operating System

Linux

### OS Version

Ubuntu 22.04

### Description

Redirect blocked for GET https://api.airforce/v1/v1/models (308)

### Steps to Reproduce

Free models > Api airforce > Add key and fetch model

### Expected Behavior

Redirect blocked for GET https://api.airforce/v1/v1/models (308)

### Actual Behavior

Redirect blocked for GET https://api.airforce/v1/v1/models (308)

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

diegosouzapw:
Fixed in `release/v3.8.44` via #5904 (thanks @hamsa0x7). Root cause: a base URL ending in `/v1/chat/completions` (e.g. `https://api.airforce/v1/chat/completions`) was only stripped of `/chat/completions`, leaving a trailing `/v1`; the endpoint builder then produced `…/v1/v1/models`, whose 308 redirect surfaced as `REDIRECT_BLOCKED` and aborted the whole discovery probe loop before the correct `…/v1/models` candidate. The `/v1` suffix is now stripped independently, and a `REDIRECT_BLOCKED` on one candidate continues to the next endpoint. Already merged into the active development branch — feel free to pull and test. Ships in the next release (v3.8.44).

PR #14001, opened 2026-09-17
fix(providers): honor the base URL override in OpenRouter model discovery

tiangao88:
## Summary

Model discovery for the built-in `openrouter` provider resolved its catalog URL only from `PROVIDER_MODELS_CONFIG`, which is pinned to the **global** catalog:

```ts
// src/app/api/providers/[id]/models/discovery/providerModelsConfig.ts
openrouter: {
  url: "https://openrouter.ai/api/v1/models",
```

The per-connection base-URL override (`providerSpecificData.baseUrl`, exposed in the UI as "Advanced -> override base URL") was never consulted on the **discovery** path, while the **inference** path has honored it since #6147 (`open-sse/executors/base.ts` -> `resolveBaseUrl`, "Operator's manual override always wins").

So a connection pointed at a different OpenRouter region kept importing the global catalog: the per-connection model list, and the auto-sync that maintains it, advertised model ids the configured endpoint cannot serve. The mismatch only surfaced later, at inference time, so a region/catalog mismatch presented as a provider outage rather than a configuration problem.

The two catalogs differ materially. Measured today:

- `https://openrouter.ai/api/v1/models` -> **444** model ids
- `https://eu.openrouter.ai/api/v1/models` -> **58** model ids (a strict subset: 0 ids unique to the EU host)

Requests for ids present in the global catalog but absent from the configured endpoint fail closed with `404`.

## Change

In the discovery URL resolution (`src/app/api/providers/[id]/models/route.ts`), `openrouter` now prefers the connection override, reusing the existing `addModelsSuffix()` normalization from `src/lib/providers/validation/urlHelpers.ts` (drops a trailing chat/responses/messages path, appends `/models`, leaves an existing `/models` untouched). When no override is set the built-in global catalog is still used - unchanged behavior.

This mirrors the `openai` handling added for the same class of bug in #5899. Other providers were deliberately left alone: several resolve their discovery URL from a region- or registry-specific endpoint, so a blanket rewrite would regress them.

## Related Issues

- None found: searches for OpenRouter + base URL / discovery / region / `eu.openrouter.ai` returned no matching issue. Happy to file a tracking issue if you would prefer this tracked separately.

## Validation

- Change type: provider / routing
- **Automated tests could not be executed in my environment.** The local clone's `node_modules` is incomplete (`tsx`, `better-sqlite3`, `xxhash-wasm` and others are missing or empty; a full `npm install` is OOM-killed on this machine), so I could not run the focused test loop or `npm run lint` locally. I am relying on CI for the suite, lint and build, and flagging this explicitly rather than ticking boxes I did not run.
- The URL-shaping this change depends on was verified independently against the relevant inputs: `.../api/v1` -> `.../api/v1/models`, a trailing slash, an already-`/models` URL, `.../api/v1/chat/completions`, and the empty string (-> `""`, which falls back to the pinned catalog).

## Tests Added Or Updated

- `tests/unit/openrouter-models-baseurl-override.test.ts` (new) - asserts discovery hits the overridden base URL for `openrouter`, and that the built-in global catalog is still used when no override is set. Follows the harness of the existing `tests/unit/poe-provider-models-baseurl.test.ts` and `tests/unit/siliconflow-model-sync.test.ts`.

## Coverage Notes

- The production change is confined to the `openrouter` branch of the discovery URL resolution in `src/app/api/providers/[id]/models/route.ts` (one branch, reusing an already-exported helper). If coverage moves down in that file, it is because the new branch is exercised through the route-level test rather than a direct unit call.

## Reviewer Notes

- The `openai` branch above was intentionally not folded into a shared path: its normalizer differs (explicit `/v1` stripping to build `/v1/models`), and merging the two would change `openai` behavior inside a bug-fix PR. Worth a follow-up if you want a single code path for both.
- No migrations, no feature flag, no schema change.


tiangao88:
Heads-up on the red checks: they're inherited, not from this PR — it only adds a changelog fragment, a new test file, and an 18-line branch in `src/app/api/providers/[id]/models/route.ts`.

- **Unit Tests (3/4)** — the new test executed and **passes 2/2**.
- **API Route Typecheck** — 2 regressed errors, both in files untouched here: `codex-responses-ws/route.ts` (TS2339) and `sse/handlers/chat.ts` (TS2345 — that file was modified by the base commit `872376bdc` itself).
- **Fast Quality Gates** — one gate only: `[test-file-size]` on `tests/integration/chat-pipeline.test.ts` (1740 > frozen 1736, already over its cap on the base); source `[file-size]` reports `OK` in PR mode for the file this PR edits.
- **Docs Gates** — `OMNIROUTE_STRIP_SYSTEM_PREAMBLE` present in code but missing from `.env.example` (absent at the current tip too).
- **Unit Tests (1/4, 2/4, 4/4)** — webpack / RTK-compression / SSE-stream-buffer suites; unrelated areas.

Rebasing onto tip `83fa4328f` clears none of these, so I left the head as is — happy to rebase on request.

diegosouzapw:
Thanks @tiangao88 — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #14009, opened 2026-09-17
fix(perplexity-web): keep runs of spaces when cleaning non-streaming answers

### LINKED ISSUE #13968: fix(providers): Perplexity Web collapses runs of spaces, corrupting code indentation
### OmniRoute Version

3.8.50

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 Pro 10.0.26200

### Node.js Version

v24.14.1

### Provider(s) Involved

perplexity-web

### Model(s) Involved

pplx-sonnet, pplx-opus, pplx-gpt-5.6-sol, pplx-gpt-5.6-terra, pplx-kimi, pplx-grok-4.6

### Client Tool

pi coding agent 0.85.1, plus direct requests to `/v1/chat/completions`

---

## Description

Perplexity Web collapses every run of spaces into a single space in model output. This affects both plain-text responses and tool-call arguments, so any file an agent writes through `perplexity-web` has broken indentation.

For Python this makes every generated file unimportable (`IndentationError`). For other languages it silently destroys formatting.

This is related to, but distinct from, #12104. In my setup structured tool calls **are** emitted correctly (9/9 `pplx-*` models produced a valid `write_file` call with parseable JSON). The problem is that the *content* of those calls, and of normal text replies, loses whitespace.

Tabs are **not** collapsed, which suggests a whitespace normalization step (for example, text extracted from rendered markdown/HTML) rather than a model behavior.

## Steps to Reproduce

Connection: `perplexity-web` with a valid Pro `__Secure-next-auth.session-token` (confirmed logged in: normal questions are answered correctly, no "Sign up and repeat your request").

Reference code sent in each prompt:

```python
def f(x):
    if x:
        return 1
    return 0
```

**A. Plain text, no tools.** Ask `pplx-sonnet` to output that code in a code block, preserving indentation.

**B. Tool call.** Provide a `write_file(path, content)` tool and ask `pplx-sonnet` to write `f.py` with exactly that code.

**C. Tool call with tabs.** Same as B, but ask it to indent with TAB characters.

Then inspect the leading whitespace of each returned line.

## Expected Behavior

Leading whitespace is preserved exactly: 4 spaces for the first level and 8 for the second in A and B, and tabs in C.

## Actual Behavior

| Test | Leading whitespace returned |
|---|---|
| A. plain text | `" "` (a single space, at every nesting level) |
| B. tool call arguments | `" "` (a single space, at every nesting level) |
| C. tool call, tabs requested | `"\t"` and `"\t\t"` (preserved correctly) |

All nesting information is lost with spaces and preserved with tabs.

### Impact on a real agent task

I ran a multi-step coding task (write a 3-function Python module) through the `pi` agent with each model, then graded the result against 30 hidden unit tests:

| Model | Result |
|---|---|
| pplx-sonnet | File written; all 103 indented lines have exactly 1 space; `IndentationError` |
| pplx-gpt-5.6-sol | File written; `IndentationError` |
| pplx-kimi | File written; `IndentationError` |
| pplx-grok-4.6 | File written; `IndentationError` |
| pplx-opus | Emitted the tool call as plain text (`<tool>{...}`) instead of a structured call; no file |
| pplx-gpt-5.6-terra | No file written |

For comparison, the same task through the same OmniRoute instance with other providers (OpenRouter, OpenCode, Cohere, Cloudflare Workers AI) scored 29–30/30 with correct indentation, so the client and gateway pipeline are otherwise fine.

### Attempted workaround (not viable)

Adding a system-prompt rule to "indent with tabs" (via `pi --append-system-prompt`) caused both `pplx-sonnet` and `pplx-gpt-5.6-sol` to stop calling tools entirely: they replied "Created `tools.py`…" with no tool call in the session, and no file existed. That matches the symptom described in #12104.

## Notes

- The collapse happens in plain-text responses as well, so it looks like it is applied to the whole text stream from Perplexity Web, not only to the tool-call emulation layer.
- Suggested area to check: any text/HTML/markdown extraction or `\s+` normalization applied to Perplexity Web answer text before it is returned or parsed for tool calls.

costajohnt:
## Summary

- `cleanResponse(text, strip = true)` in `open-sse/executors/perplexity-web/protocol.ts` replaced every run of two or more spaces with a single space (`MULTI_SPACE = / {2,}/g`). The non-streaming path runs it with `strip` on, and tool mode always takes that path because it buffers the full completion before turning `<tool>` text into `tool_calls`. So every file an agent wrote through `perplexity-web` came back with one space per nesting level, and generated Python was unimportable. Tabs survived, which is what pointed at this normalization rather than the model.
- The only reason `MULTI_SPACE` existed was to tidy the double space left behind when a `[n]` citation marker between two words was removed. That is now handled by `CITATION_RE` itself (`/ ?\[\d+\]/g`), which takes the space before the marker with it, and `MULTI_SPACE` is gone. `text [1] more` still cleans to `text more`; `text [3].` now cleans to `text.` instead of `text .`.
- Blank-line squashing (`MULTI_NL`) and the final trim are unchanged. Streaming deltas already ran with `strip = false` and are unaffected apart from the citation spacing improvement.

## Related Issues

- Closes #13968

## Validation

- [x] Change type: provider
- [x] Focused test: `tests/unit/perplexity-web-preserve-spaces-13968.test.ts` (3 tests, 2 of them fail on `release/v3.8.51` before the fix) plus every `tests/unit/*perplexity*.test.ts` suite: 80 pass, 0 fail
- [x] eslint (with `config/quality/eslint-suppressions.json`) and prettier clean on the touched files; `node scripts/check/check-open-sse-typecheck.mjs` reports 0 errors
- [x] Branched from `release/v3.8.51` at 83fa4328f
- [x] Changelog fragment: `changelog.d/fixes/13968-perplexity-web-preserve-spaces.md`

## Tests Added Or Updated

- `tests/unit/perplexity-web-preserve-spaces-13968.test.ts` (new): a fenced Python block with 4 and 8 space indentation and a line with an aligned trailing comment pass through `cleanResponse` untouched; citation markers with and without a leading space, chained `[1][2]`, and a marker before punctuation all clean to single spacing; trim, blank-line squashing and `strip = false` passthrough are pinned so the rest of the cleaner is shown unchanged.

## Coverage Notes

- `open-sse/executors/perplexity-web/protocol.ts`: the changed regex and the `strip` branch of `cleanResponse` are both exercised directly by the new test. No other file changed.

## Reviewer Notes

- Removing `MULTI_SPACE` rather than gating it (for example, skipping it inside code fences) because the collapse also hit tool-call arguments, which carry no fence to detect. Trading a cosmetic cleanup for correct output seemed the right call; if there is a case where Perplexity emits stray double spaces that you want normalized, a narrower rule that leaves leading whitespace alone would be the next step.
- Not related to #12104 (tool calls not emitted); the reporter confirmed structured calls arrive, only their content lost whitespace.
- I have not run this against a live Perplexity session; the fix is a pure string transform and the unit test reproduces the reporter's exact input offline.
- Pre-existing and out of scope: `CITATION_RE` still strips any `[n]` token, so array indexing like `arr[0]` in code output is still mangled. Worth a follow-up issue; this PR only stops the whitespace collapse.
- Edge: two spaces before a citation marker (`text  [1] more`) now keep one extra space instead of collapsing; consistent with not touching runs of spaces.


diegosouzapw:
Thanks @costajohnt — merging via the release merge-train. Validated in local merge-train (.claude/worktrees/merge-train-20260918-111718-suite.log) on the devbox @ train tip 7bb373fba5e241964c0ffb17bd03a804700839bb, boarded with 55 sibling PRs: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; 747/747 changed-area node:test cases + 476/476 vitest green (fast parity mode — the full suite ran today on the release tip via the base-red train and runs again on the 3b train). Merged --admin per merge-gates §7.

PR #14022, opened 2026-09-17
fix(ci): document OMNIROUTE_STRIP_SYSTEM_PREAMBLE — the env/docs base red blocking every PR

diegosouzapw:
Base-red repair for `release/v3.8.51` — documentation only, no behavior change.

## The red

`Docs sync + fabricated-docs (strict)` fails on the base tip and therefore on every open PR
(tracked in #14004):

```
✗ In code but missing from .env.example: 1
   - OMNIROUTE_STRIP_SYSTEM_PREAMBLE
✗ Env / docs contract is out of sync.
```

Reproduced against a clean `release/v3.8.51` checkout with no local changes.

## The gap

`OMNIROUTE_STRIP_SYSTEM_PREAMBLE` is read at
`open-sse/translator/response/openai-to-claude.ts:413` and gates the system-echo preamble
stripper in `open-sse/utils/directivePreambleStripper.ts`, but it was never added to
`.env.example` / `docs/reference/ENVIRONMENT.md`.

## The fix

Documented in both, next to its sibling `OMNIROUTE_SYSTEM_INSTRUCTION_APPEND`, with the default
(`0`, off) and the reason it is opt-in: the stripper matches by English-prose heuristics and does
mutate the response payload, so a reply that legitimately opens with such a section would lose it.

After: `✓ Env / docs contract is in sync.`

Note: the file-size ratchet drift on `src/sse/handlers/chatHelpers.ts` listed in #14004 is
non-blocking and rebaselined at release — not touched here.


diegosouzapw:
Validated in local merge-train /tmp/mt-basereds.log on 192.168.0.113 @ train tip c2fb6404606064c56520d7b2a148e8770dc70cd0 (boarded #14022 #14016 #14079 #14082): static gates green; test:unit 40635 tests, 17 failing — every one reproduces on the pure release tip 1603c86e (41 failing there; list in the base-red sweep), zero new reds. Merged with --admin per merge-gates §4/§7 (inherited base-reds only).

PR #14023, opened 2026-09-17
fix(providers): detect Lemonade labels[] vision capability

### LINKED ISSUE #13918: fix(providers): detectVisionInput ignores Lemonade labels[], importing vision models as text-only
> Companion to #13847. That issue is `projectCombo` **losing** vision on members that already have it. This issue is the upstream half: `detectVisionInput` never **acquiring** vision for Lemonade models in the first place.

## OmniRoute Version

3.8.50 (`package.json` inside `diegosouzapw/omniroute:latest-web`)

## Installation Method

Docker (`diegosouzapw/omniroute:latest-web`, compose service, port 20128)

## Operating System

macOS

## OS Version

macOS (Apple Silicon)

## Node.js Version

n/a — running the published Docker image

## Provider(s) Involved

`lemonade` (Lemonade Server, built-in local provider — remote instance fronted over Tailscale at `http://100.88.99.116:13305/api/v1`)

## Model(s) Involved

`lemonade/Gemma-4-26B-A4B` (vision-capable; reports `labels: [... "vision" ...]`)

Same gap applies to any other Lemonade model carrying the `vision` label; also affects `lemonade/Qwen3.8-Flash-Next-UD-Q2_K_XL` / `lemonade/DeepSeek-V4-Flash-0731-UD-IQ2_XXS` for whatever their labels declare.

## Client Tool

Any OpenAI-compatible client or tool that reads `/v1/models` (or a combo derived from it) to decide whether to enable image input.

## Description

Lemonade Server does not broadcast capability booleans. Its `GET /v1/models` exposes capabilities **only** as a `labels[]` string array:

```
id, created, object, owned_by, checkpoint, recipe, size,
max_context_window, context_length, downloaded, suggested,
update_available, labels
```

where `labels` may contain `chat`, `vision`, `reasoning`, `tool-calling`, `coding`, `transcription`, `embeddings`, `reranking`, `image`, …

OmniRoute's import-time vision detection, `detectVisionInput(record)` (`src/lib/providerModels/modelDiscovery.ts:50`, called from the sync loop at `:330` and stored as `supportsVision` at `:359`), recognizes exactly four shapes:

```ts
export function detectVisionInput(record: JsonRecord): boolean {
  if (record.supportsVision === true) return true;

  const architecture = asRecord(record.architecture);
  if (modalitiesIncludeImage(architecture.input_modalities)) return true;
  if (modalitiesIncludeImage(record.input_modalities)) return true;

  const modality = toNonEmptyString(architecture.modality) || toNonEmptyString(record.modality);
  if (modality) {
    const [inputPart] = modality.toLowerCase().split("->");
    if ((inputPart || "").includes("image")) return true;
  }
  return false;
}
```

`labels` is not among them, and nothing else in the import path reads it either — an exhaustive grep across `src/` and `open-sse/` for a `labels` capability read returns no hits in `src/lib/providerModels/`, `src/lib/db/models/synced.ts`, or `src/app/api/providers/` (the only `labels` occurrences are UI copy, issue labels, and unrelated settings).

Consequently a Lemonade model that Lemonade itself labels `vision` is imported with `supportsVision` unset. With `model_capabilities` empty for the provider (it is a local provider, not models.dev-synced) and no static spec covering the id, **every** downstream vision source resolves `null` — so the model is advertised as text-only in `/v1/models`, and any combo containing it loses `multimodal` (see #13847).

**Critically, this is metadata-only.** Vision works end-to-end at request time; only the advertised capability is wrong:

- text-only request to `lemonade/Gemma-4-26B-A4B` → HTTP 200, served `Gemma-4-26B-A4B`
- identical request with a solid-red 64×64 PNG attached → HTTP 200, served `Gemma-4-26B-A4B`, response `"Red"`

The model received and correctly interpreted the image through OmniRoute. Only the flag is missing.

## Steps to Reproduce

1. Configure a `lemonade` provider connection pointing at a Lemonade Server instance whose catalog includes a vision-labelled model.
2. Trigger/allow a model sync.
3. Inspect the imported entry:

   ```bash
   curl -s http://localhost:20128/v1/models | python3 -c '
   import json,sys
   for m in json.load(sys.stdin)["data"]:
       if m["id"].startswith("lemonade/"):
           print(m["id"], m.get("capabilities"), m.get("input_modalities"))'
   ```

4. Compare against what Lemonade itself reports for the same model:

   ```bash
   curl -s http://<lemonade-host>:13305/api/v1/models | python3 -c '
   import json,sys
   for m in json.load(sys.stdin).get("data", []):
       print(m.get("id"), m.get("labels"))'
   ```

## Expected Behavior

When an upstream model record declares a `vision` label (Lemonade's documented capability channel), the import should set `supportsVision: true` on the synced model, exactly as it does for `architecture.input_modalities` from an OpenRouter-style catalog. The model should then advertise `capabilities.vision: true` in `/v1/models` and count as vision-capable in any combo that includes it.

## Actual Behavior

Lemonade's `vision` label is silently dropped. The model is imported without `supportsVision`, resolves to `null` on every vision source, and is advertised as text-only — while still functioning correctly when an image is actually sent.

```
lemonade/Gemma-4-26B-A4B  →  capabilities {'tool_calling': True, 'reasoning': True}   # no vision key
                             input_modalities: absent
```

## Test Impact

- Vision-capable Lemonade models cannot be selected for image requests by capability-aware clients, and combos containing them are advertised non-multimodal even when every member supports images.
- Users are pushed toward per-model custom-model overlays or manual DB writes purely to restore a flag the upstream already declared.

## Error Logs / Output

No error is raised — the label is dropped silently at import. Raw catalog responses above.

## Screenshots

n/a (API-level)

## Additional Context

- `detectVisionInput` was introduced by **#4264** ("capture image-input (vision) capability at sync time"), whose stated goal was exactly this class of problem: *"Without this, synced models reached the catalog with no vision flag and vision-capable models (which work at request time) showed up as non-vision after import."* #4264 closed the gap for OpenRouter-shaped catalogs but left other documented capability channels — including Lemonade's `labels[]` — uncovered.
- Lemonade's label vocabulary is documented at https://lemonade-server.ai/docs/api/openai/ (see also `lemonade-sdk/lemonade`).
- This is a **distinct** defect from #13847: #13847 is a *resolution time* bug (combo projection passing a bare model string that fails spec/registry/fragment lookup for variants like `-free` and path-qualified ids). This one is an *import time* bug (a source format never parsed). Fixing only one leaves the other path broken — a Lemonade model needs both this fix and the #13847 fix to advertise vision both individually and inside a combo.
- Related divergence class: `resolveVisionCapability` documents that vision sources historically drifted (#4071, #4072, #8250, #9195). This adds Lemonade's `labels[]` as a source that was never wired in.

## Validation Plan

Unit-test `detectVisionInput` directly against a realistic Lemonade record:

```ts
// Lemonade Server /v1/models record shape
const lemonadeRecord = {
  id: "Gemma-4-26B-A4B",
  object: "model",
  owned_by: "lemonade",
  checkpoint: "amd/gemma-4-26b-a4b",
  recipe: "llamacpp",
  size: 26000000000,
  max_context_window: 262144,
  context_length: 262144,
  downloaded: true,
  labels: ["chat", "vision", "reasoning", "tool-calling"],
};

expect(detectVisionInput(lemonadeRecord)).toBe(true);
```

And a negative case to pin the conservative behaviour (a Lemonade text-only model must stay `false`):

```ts
const textOnly = { ...lemonadeRecord, labels: ["chat", "tool-calling"] };
expect(detectVisionInput(textOnly)).toBe(false);
```

Then assert the end-to-end invariant: after sync, `lemonade/Gemma-4-26B-A4B` advertises `capabilities.vision: true` in `/v1/models`, and `computeComboCapabilities(comboContainingIt).multimodal === true`.

Suggested fix shape (matching the existing four shapes — a case-insensitive membership test, not a substring match, to avoid the false positives that led to bare `gemma` being removed from the id-fragment heuristic):

```ts
// Lemonade Server exposes capabilities as a `labels` string array.
const labels = Array.isArray(record.labels) ? record.labels : [];
if (labels.some((l) => typeof l === "string" && l.trim().toLowerCase() === "vision")) return true;
```

Note `labels` is also used by other OpenAI-compatible local servers (e.g. Lemonade's Ollama-compat surface), so the fix is likely to generalise beyond this one provider.

diegosouzapw:
## Summary

`detectVisionInput()` (`src/lib/providerModels/modelDiscovery.ts`) only recognized four upstream vision-declaration shapes — an explicit `supportsVision` flag, OpenRouter's `architecture.input_modalities` array, a top-level `input_modalities` array, and `architecture.modality`/`modality` string. Lemonade Server's `GET /v1/models` exposes capabilities only through a `labels[]` string array (e.g. `["chat", "vision", "reasoning", "tool-calling"]`) — none of the fields above — so a Lemonade vision model was imported with `supportsVision` unset and advertised as text-only in `/v1/models`, even though it correctly handles image input at request time.

Fix: add a fifth, purely additive branch that reads `record.labels`, guards it as an array, and does a case-insensitive, trimmed **exact** membership test for `"vision"` (not a substring match — see the earlier false-positive lesson with bare `gemma` id-fragment matching). All four existing shapes are untouched.

Closes #13918

## Root cause

- `src/lib/providerModels/modelDiscovery.ts:50-63` (before fix): `detectVisionInput()` never read `record.labels`.
- Call site at `src/lib/providerModels/modelDiscovery.ts:334`: `const supportsVision = detectVisionInput(record);`, stored on the `SyncedAvailableModel` at sync time — so the miss is permanent for every synced Lemonade vision model until re-synced after this fix.

`tests/unit/lemonade-vision-labels-13918.test.ts` (new file):

RED (before the fix, against unmodified `origin/release/v3.8.51`):
```
✖ detectVisionInput recognizes a Lemonade record whose labels[] includes vision
  AssertionError [ERR_ASSERTION]: false !== true
✔ detectVisionInput stays false for a Lemonade record without the vision label
✖ detectVisionInput is case-insensitive and trims whitespace in labels[]
  AssertionError [ERR_ASSERTION]: false !== true
```

GREEN (after the fix):
```
✔ detectVisionInput recognizes a Lemonade record whose labels[] includes vision
✔ detectVisionInput stays false for a Lemonade record without the vision label
✔ detectVisionInput is case-insensitive and trims whitespace in labels[]
✔ detectVisionInput does not substring-match labels[] entries
✔ detectVisionInput ignores non-array labels without throwing
✔ detectVisionInput leaves non-Lemonade shapes byte-identical (architecture.modality path)
ℹ tests 6, pass 6, fail 0
```

Also re-ran the pre-existing tests that exercise `detectVisionInput`/the sync path, both still green: `tests/unit/provider-models-vision-override-1904.test.ts` (5/5) and `tests/unit/openrouter-vision-sync-4264.test.ts` (2/2, confirming other providers' detection paths stay byte-identical).

## Gates run

- `node scripts/check/check-file-size.mjs` — no ✗ on touched files (only pre-existing, unrelated base-red drift on `chatHelpers.ts`/`chatCore.ts`/`chatcore-translation-paths.test.ts`)
- `node scripts/check/check-complexity.mjs` — OK (2850 violations, baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1289 violations, baseline 1437)
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json src/lib/providerModels/modelDiscovery.ts tests/unit/lemonade-vision-labels-13918.test.ts` — exit 0
- `node --import tsx/esm --test tests/unit/lemonade-vision-labels-13918.test.ts` — 6/6 pass
- `node --import tsx/esm --test tests/unit/provider-models-vision-override-1904.test.ts` — 5/5 pass (unchanged)
- `node --import tsx/esm --test tests/unit/openrouter-vision-sync-4264.test.ts` — 2/2 pass (unchanged; proves non-Lemonade providers' behavior is untouched)
- Pre-commit hooks (lint-staged, docs-sync, any-budget, tracked-artifacts) — all passed on commit

## Test plan

- [x] `node --import tsx/esm --test tests/unit/lemonade-vision-labels-13918.test.ts`
- [x] Existing `detectVisionInput`/sync-path tests re-run and confirmed green


diegosouzapw:

PR #14026, opened 2026-09-17
fix(sse): deepseek-flash resolves to 1M context, not the 128k default (#13922)

### LINKED ISSUE #13922: [BUG] deepseek-flash falls back to 128000 context limit (actual window is 1M)
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

macOS

### OS Version

27.0

### Node.js Version

_No response_

### Provider(s) Involved

DeepSeek

### Model(s) Involved

deepseek-flash (ds/deepseek-flash), also built-in ds/deepseek-v4-flash

### Client Tool

Claude Code (Anthropic Messages, /v1/messages)

### Description

Requests to `ds/deepseek-flash` are rejected by OmniRoute once the prompt exceeds 128k tokens. The request never reaches DeepSeek (request log shows account `-`, 58 ms, 0 tokens in).

`deepseek-flash` is the model name DeepSeek introduced on 2026-09-10 for DeepSeek-V4.1-Flash. Official docs list a 1M context window: https://api-docs.deepseek.com/quick_start/pricing

The model was imported from the provider's `/models` list. With no metadata for it, `resolveTokenLimit` in `open-sse/services/contextManager.ts` reaches the generic fallback (`DEFAULT_LIMITS.default`, `specific: false`) and returns 128000.

### Steps to Reproduce

1. Add a DeepSeek API key and import models (`deepseek-flash` appears).
2. Add `ds/deepseek-flash` to a combo (or call it directly).
3. Send a request with more than 128k input tokens, e.g. a long Claude Code session.

### Expected Behavior

The context limit for `deepseek-flash` should be 1,000,000, same as `deepseek-v4-pro`, and the request should be forwarded to DeepSeek.

### Actual Behavior

The request is rejected locally with HTTP 400 at ~128k input tokens.

### Test Impact

Unsure

### Error Logs / Output

```shell
[400] Input exceeds context window for deepseek/deepseek-flash: estimated 128696 input tokens, limit 128000. Reduce the prompt or route to a model with a larger context window.

Model 'deepseek-v4-flash' is not available in the active live catalog for provider 'deepseek'.
```

### Screenshots

_No response_

### Additional Context

Workaround: setting `CONTEXT_LENGTH_DEEPSEEK=1000000` on the container fixes it.

Related: the built-in `ds/deepseek-v4-flash` entry fails with "not available in the active live catalog", because DeepSeek no longer lists that name in `/models`. It is now a legacy alias routed to V4.1-Flash. See also #13505.

Setup: single DeepSeek account, used in a combo with Priority strategy.

### Validation Plan

1. Without `CONTEXT_LENGTH_DEEPSEEK` set, send a >128k-token request to `ds/deepseek-flash`: it should be forwarded to DeepSeek instead of returning 400.
2. Run the new unit test for `resolveTokenLimit("deepseek", "deepseek-flash")`.

diegosouzapw:
## Root cause

`REGISTRY['deepseek']` (`open-sse/config/providers/registry/deepseek/index.ts`) never set
`defaultContextLength`. `resolveTokenLimit()` (`open-sse/services/contextManager.ts:397-436`)
walks env override → models.dev synced DB → registry `defaultContextLength` → model-name
heuristics → generic 128k catch-all. Any DeepSeek model id not individually listed in the
registry (or not yet synced from models.dev) — such as `deepseek-flash`
(DeepSeek V4.1 Flash, announced 2026-09-10) — fell all the way through to the generic 128k
default instead of DeepSeek's real 1M window, so requests over 128k tokens were rejected
locally before ever reaching DeepSeek.

## Where the 1M figure comes from

Not invented: both existing DeepSeek entries in the same registry file
(`deepseek-v4-pro`, `deepseek-v4-flash`) already declare `contextLength: 1_000_000`. Corroborated
by two of this project's own local reference implementations checked out under
`_references/_sistemas_proxys/` (gitignored, not part of this diff):
- `9router`: `open-sse/providers/capabilities.js:352` — wildcard rule
  `{ pattern: "*deepseek-v4*", caps: { contextWindow: 1000000 } }` covering unlisted
  DeepSeek-V4-family model ids the same way.
- `sub2api`: `openai_codex_models_service.go:337` hardcodes
  `configuredCodexDeepSeekV4Context = 1_000_000` for the whole DeepSeek V4 family.

## Fix

- `open-sse/config/providers/registry/deepseek/index.ts`: added
  `defaultContextLength: 1_000_000` to the provider entry (mirrors every other registry entry,
  e.g. `registry/openai/index.ts:13`, `registry/glm/index.ts:10`) — this alone fixes
  `resolveTokenLimit` step 3 for any current or future unlisted DeepSeek model id, not just
  `deepseek-flash`.
- Added an explicit `deepseek-flash` model entry (`contextLength: 1_000_000`,
  `maxOutputTokens: 384_000`, reasoning/tool-calling flags matching the two existing DeepSeek
  entries) so its own catalog metadata is accurate rather than only inherited from the provider
  default.

## Regression test (TDD, RED → GREEN)

`tests/unit/deepseek-flash-context-limit-13922.test.ts` (new, permanent):
- `resolveTokenLimit("deepseek", "deepseek-flash")` → `{ limit: 1_000_000, specific: true }`
- an arbitrary unlisted `deepseek-*` model id also resolves via the new provider default

RED (pre-fix, on `origin/release/v3.8.51`):
```
AssertionError [ERR_ASSERTION]: expected deepseek-flash to resolve to the 1M DeepSeek window, got 128000 (specific=false)
actual: 128000, expected: 1000000, operator: 'strictEqual'
```
GREEN (post-fix):
```
✔ #13922: deepseek-flash (new 1M model, no registry/synced metadata) resolves to DeepSeek's 1M window
✔ #13922: any unlisted deepseek model id inherits the provider's 1M defaultContextLength
ℹ pass 2
ℹ fail 0
```

## Gates run

- `node scripts/check/check-file-size.mjs` — no ✗ on touched files (pre-existing ✗ on
  `chatHelpers.ts` / `chatCore.ts` / `chatcore-translation-paths.test.ts` are the known
  base-red drift, untouched here).
- `node scripts/check/check-complexity.mjs` — OK (2850 violations vs 3218 baseline, no new).
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1289 vs 1437 baseline, no new).
- `npm run check:open-sse-typecheck` — 0 errors.
- New test: `node --import tsx/esm --test tests/unit/deepseek-flash-context-limit-13922.test.ts` — 2/2 pass.
- Existing token-limit/combo/DeepSeek tests re-run: `combo-explicit-context-length-routing`,
  `combo-context-generic-default-10734`, `chatcore-combo-context-limit-8378`,
  `deepseek-pow-slot-leak-13094`, `opencode-deepseek-json-schema-fallback` — all pass in
  isolation. Note: `combo-context-generic-default-10734.test.ts`'s
  `GET /v1/models does not advertise 128k...` case is intermittently flaky on this shared,
  heavily-parallel devbox (an unrelated pre-existing async-cleanup race after the test
  completes, unaffected by this change — reproduced failing/passing with and without this
  diff across ~16 repeated runs); untouched, not weakened.

Closes #13922


diegosouzapw:

Issue #7746, opened 2026-07-19
fix(backend): CCR compressor can replace the entire user prompt with a bare [CCR retrieve hash=...] marker on non-MCP upstreams


beingshafin:
### OmniRoute Version

3.8.48

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 1125H2

### Node.js Version

24.12.0

### Provider(s) Involved

Any provider (Opencode Zen, Agent Router)

### Model(s) Involved

Any model (hy-3:free)

### Client Tool

Opencode

### Description

When using OmniRoute, the upstream model intermittently receives only a CCR retrieval marker instead of the actual user prompt.

This issue is **not client-specific**. I have reproduced it using both:
- OpenCode
- Claude Code

Both clients are configured to use the same OmniRoute OpenAI-compatible endpoint.

The observable symptom is that the model responds with something similar to:

> I only received this from you as your message:
>
> [CCR retrieve hash=5add8e7eb7d240d9b11d3af9]

which indicates that, from the model's perspective, the original prompt was not available.

I'm intentionally **not making assumptions about the intended internal CCR implementation**. This report is only about the externally observable behavior: the model occasionally receives only the CCR marker instead of the user's prompt.

### Steps to Reproduce

1. Start OmniRoute with CCR enabled.
2. Connect either OpenCode or Claude Code to the OmniRoute endpoint.
3. Start a new conversation.
4. Send a larger prompt (approximately 4–6 KB). (e.g. I use this prompt)
```
I want you to build a Playwright automation script for AgentRouter with the following requirements.

## Goal

Automate claiming the daily sign-in bonus for my two AgentRouter accounts.

## Environment

* Language: JavaScript (Node.js)
* Framework: Playwright
* Operating System: Windows
* Use Playwright only. Do not use Selenium or Puppeteer.

## Accounts

I have two different AgentRouter accounts.

* Account 1 logs in via **GitHub Account A**
* Account 2 logs in via **GitHub Account B**

Currently:

* GitHub Account A is logged into Microsoft Edge.
* GitHub Account B is logged into Google Chrome.

However, if using Playwright's own persistent browser profiles is more reliable than reusing my existing browser profiles, use Playwright-managed profiles instead.

## Requirements

1. Create a project structure.
2. Install all required dependencies.
3. Create persistent browser profiles so login is only required once.
4. Create a one-time setup script that:

   * Opens the browser.
   * Navigates to AgentRouter.
[...]

PR #14028, opened 2026-09-17
fix(sse): recognize MCP-gateway-namespaced CCR retrieve tool names (#13781, #13897)

### LINKED ISSUE #13781: [BUG]  CCR tool matching is too strict for MCP environments that prefix tool names
### OmniRoute Version

3.8.51

### Installation Method

Docker / Docker Compose

### Operating System

Windows

### OS Version

_No response_

### Node.js Version

_No response_

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

Hi! I’m trying to use CCR through Docker MCP, which exposes tool names with a prefix, e.g.:

expected by CCR: omniroute_ccr_retrieve
actual exposed tool: mcp__docker__omniroute__omniroute_ccr_retrieve

From the current implementation, CCR appears to require an exact tool-name match rather than normalizing or matching suffixed canonical names. This makes CCR incompatible with MCP environments that namespace/prefix tools.

Would you consider making tool resolution more flexible, for example by:
stripping known MCP prefixes before comparison, or
matching by canonical suffix (e.g. endsWith("omniroute_ccr_retrieve"))?

A safer approach might be:
normalize tool names first
then do exact matching on the canonical name

This would preserve determinism while improving compatibility with Docker MCP and similar tool providers.

Thanks!

tl;dr

change `===` in  
https://github.com/diegosouzapw/OmniRoute/blob/cad0fcc65d7733a3ab292bac7ced5d40c0a850cc/open-sse/services/compression/engines/ccr/protocolInstruction.ts#L44 

to ` endsWith("CCR_RETRIEVE_TOOL_NAME ")`

### Steps to Reproduce

Add omniroute mcp to mcp docker
Send requests to omniroute with CCR compression
Notice that CCR instructions are not injected as docker mcp breakes strict === 'omniroute_ccr_retrieve' tool name.
docker exposes `docker_mcp_omniroute_ccr_retrieve` or with toolname_prefix `docker_mcp_omniroute_omniroute_ccr_retrieve`

### Expected Behavior

should match "endsWith?()" instead of ===

### Actual Behavior

doesnt match

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

### LINKED ISSUE #13897: fix(compression): CCR protocol instruction never injected when the MCP gateway namespaces the retrieve tool name
## Problem

`callerSupportsCcrRetrieve()` matches the retrieve tool by **exact name equality**:

```ts
return flatName === CCR_RETRIEVE_TOOL_NAME || nestedName === CCR_RETRIEVE_TOOL_NAME;
```

`open-sse/services/compression/engines/ccr/protocolInstruction.ts:44`

Callers that reach OmniRoute's MCP server through a gateway that **namespaces** tool names never match. Reported from the Discord (EN) support channel with Docker MCP, where the advertised tool is:

```
mcp__docker__omniroute__omniroute_ccr_retrieve
```

## Impact

For those callers the CCR protocol instruction is **never injected**, even though the client can in fact call the tool. The model then receives `[CCR retrieve hash=<24hex> chars=N]` markers with no explanation of what they mean or which tool recovers the original text — the compression silently degrades the conversation instead of being transparent.

The namespacing is not exotic: `mcp__<server>__<tool>` is the shape Claude Code, Docker MCP and other gateways use.

## Proposed fix

Match the tool by its **trailing segment** rather than by full equality, e.g. accept a name that is `omniroute_ccr_retrieve` or ends with a separator + that name (`__`, `.`, `/`, `:`), so gateway prefixes still resolve while unrelated tools do not.

## Acceptance criteria

- A request advertising `mcp__docker__omniroute__omniroute_ccr_retrieve` gets the CCR protocol instruction injected.
- A request advertising only unrelated tools still does **not** get it (plain OpenAI-compatible callers must never be told to call the tool).
- Unit tests cover: exact name, `mcp__…__omniroute_ccr_retrieve`, dotted/slashed prefixes, and a near-miss name that must not match (e.g. `omniroute_ccr_retrieve_v2`).

## Area

Compression / CCR

## Additional context

Reported in the Discord (EN) support channel, with the caller's real tool definition.

diegosouzapw:
## Root cause

`callerSupportsCcrRetrieve()` (`open-sse/services/compression/engines/ccr/protocolInstruction.ts`)
matched the CCR retrieve tool by exact string equality against `omniroute_ccr_retrieve`. MCP
aggregators/gateways (Docker MCP Toolkit, Claude Code, and similar) rename tools when re-exposing
them to the model to avoid cross-server name collisions — typically as
`mcp__<gateway>__<server>__omniroute_ccr_retrieve`, or with a dotted/slashed namespace prefix
(`omniroute.omniroute_ccr_retrieve`). Because the match was exact, a caller that is fully
MCP-capable — it CAN and WILL call the tool, just under a gateway-assigned namespaced name — was
treated as if it had no access at all. Per the #7746 safety guarantee this check gates
`apply()` itself, not just the injected instruction, so the entire CCR compression engine was
silently skipped for such callers.

Fixes both #13781 and #13897, which report the same root cause from two angles (tool-matching
too strict for MCP environments / protocol instruction never injected when the MCP gateway
namespaces tools).

## Fix

Added `matchesCcrRetrieveToolName()`: true on an exact match, or when the name ends with one of
the separators `__`, `.`, `/`, `:` immediately followed by `omniroute_ccr_retrieve` — a
separator-bounded trailing-segment match, never a bare substring/`.endsWith()` test, so a
near-miss like `omniroute_ccr_retrieve_v2` still correctly does not match. Wired into both the
flat and nested (`{type:"function",function:{name}}`) branches of `callerSupportsCcrRetrieve()`.

`tests/unit/ccr-protocol-instruction.test.ts`

RED (confirmed against `origin/release/v3.8.51`, unfixed `protocolInstruction.ts` swapped back in
via `git checkout -- <file>`, never a branch switch — new test cases only):

```
✖ recognizes MCP-gateway-namespaced tool names (#13781, #13897) (54.472614ms)
  AssertionError [ERR_ASSERTION]: Docker MCP Toolkit style double-prefix name
  false !== true
ℹ tests 11
ℹ pass 10
ℹ fail 1
```

GREEN (fix restored):

```
✔ recognizes MCP-gateway-namespaced tool names (#13781, #13897) (4.263963ms)
✔ does NOT loosen matching into a plain substring test (2.320859ms)
ℹ tests 11
ℹ pass 11
ℹ fail 0
```

Also re-ran the existing suite in full (all 11 cases, including the pre-existing
`some_other_tool → NO instruction` and `no tools → NO instruction` cases) and the related
`tests/unit/compression/ccr-non-mcp-full-prompt-loss-7746.test.ts` (8/8) — both green, no
regression to the #7746 safety guarantee.

## Gates run

- `node --import tsx/esm --test tests/unit/ccr-protocol-instruction.test.ts` — 11/11 pass
- `node --import tsx/esm --test tests/unit/compression/ccr-non-mcp-full-prompt-loss-7746.test.ts` — 8/8 pass
- `node scripts/check/check-file-size.mjs` — no violation on touched files
- `node scripts/check/check-complexity.mjs` — OK (2850 violations vs baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1289 violations vs baseline 1437)
- `npm run check:open-sse-typecheck` — OK, 0 pre-existing errors

Verified the commit message and this PR body carry no AI-generation footer / Co-authored-by trailer.

Closes #13781
Closes #13897


diegosouzapw:

PR #14029, opened 2026-09-17
[defer] fix(build): drop orphaned httpClientAbortGuard.mjs pack-artifact entries

maxmad64bis:
> ⚠️ base-red inherited: #14004

## Summary

`check:pack-artifact` fails on `dist/httpClientAbortGuard.mjs`: the two pack-artifact allowlists require a file that nothing produces or consumes anymore. The #13636 crash-guard wiring was removed as undocumented rework collateral, leaving the policy entries orphaned. This PR drops the two orphaned entries so the gate passes again.

## Related Issues

- Related to #13636
- Related to #12732

## Validation

- [ ] Change type: build-deploy
- [x] `npm run lint` — 0 errors on the touched test file (`pack-artifact-policy.ts` is eslint-ignored by repo config, warning only)
- [x] File-size gate: clean on PR files (`chatcore-translation-paths` violation is base-inherited, file untouched by this PR)
- [x] Api-typecheck: no new regression (`chat.ts` TS2345 is base-inherited, file untouched by this PR (`OMNIROUTE_STRIP_SYSTEM_PREAMBLE`, introduced by #12905 without `.env.example` entry — file untouched by this PR))

## Tests Added Or Updated

- `tests/unit/pack-artifact-policy.test.ts`: removed the orphaned `dist/httpClientAbortGuard.mjs` line from the snapshot expected list.
- No new test needed: the three derived closure tests (`pack-artifact-server-ws-closure` 3/3, `pack-artifact-entrypoint-closures` 6/6, `assemble-standalone` 6/6) prove the purge is safe — they stay green with and without the entries.

## Coverage Notes

- No `src/`, `open-sse/`, `electron/` or `bin/` production code changed. Coverage unaffected.

## Reviewer Notes

- Proof the file is dead twice over: no producer (no `EXTRA_MODULE_ENTRIES` entry, no `prepublish.ts`/`colocate-standalone.mjs` reference) and no consumer (the wrapper imports 8 `./` siblings plus `./server.js`, none pointing at the guard; the three `src/` consumers import via the `@/` alias, bundled by Next, never the loose dist file). Full evidence trail in the linked issues.
- `check:pack-artifact --policy-only` passes (3672 entries); the full gate with a real build stays with CI.
- A sibling PR restores the #12861 settled-guard and #13636 predicates that the same rework removed; this PR touches none of those files.


diegosouzapw:
Review note on merge ordering — the premise of this PR is about to change under it.

Today the tip really is in the state this PR describes: `scripts/build/pack-artifact-policy.ts:58,206` still allow-list `httpClientAbortGuard.mjs` while nothing produces or consumes it. But that is not because the wiring was intentionally removed — it is collateral damage from the `e7999c477b` squash of #13717 (tracked in #14062), which deleted the #13636 producer (`scripts/build/assembleStandalone.mjs`) and consumer (`scripts/dev/standalone-server-ws.mjs`).

PR #14064 restores exactly those two files, verified byte-for-byte against the original #13636 merge. Once it lands, the file is produced and shipped again, and this PR would then be removing allow-list entries for a file that is present — `check:pack-artifact` would flag it.

So this is being **deferred rather than merged**: the purge is correct against a broken tip, and restoring the crash guard is the higher-value fix. No fault in the diff itself — the analysis was right for the tree you saw. If after #14064 lands there is still something orphaned here, this reopens immediately; otherwise it will be closed with credit noted on #14062.

diegosouzapw:

PR #14038, opened 2026-09-17
fix(guardrails): resolve nested combo-ref hops before vision-bridge decision (#13927)

### LINKED ISSUE #13927: [BUG] combo-ref hop strips image parts: nested combo advertises vision: true but leaf never receives the image
### OmniRoute Version

3.8.50

### Installation Method

Electron desktop app

### Operating System

macOS

### OS Version

26.6

### Node.js Version

v24.18.0 (bundled with OmniRouteTray.app)

### Provider(s) Involved

qwen-cloud-token-plan, command-code, opencode-go

### Model(s) Involved

deepseek-v4.1-flash (via preset/default → preset/deepseek-flash-latest)

### Client Tool

aside browser

### Description

**TL;DR:** Any request whose routing path contains a `combo-ref` (a combo member of `kind: "combo-ref"`) reaches the leaf model with the image parts removed. The call still returns 200, so clients silently degrade to text-only.

## The XY problem that led here

**X, what I actually want:** one stable model id in the router that all my connected apps can point at, so that when I change which model sits underneath, I don't have to touch any client. I also want to inject a system prompt at that stable id.

**Attempt 1, model alias.** Blocked by design. `/v1/models` advertises zero bare aliases: 69 of 88 catalog models carry a bare `alias`, and none of them appear in `/v1/models` (814 ids, all provider-prefixed). Aliases still route fine if you call them, they are just not advertised, which matches the lifecycle policy from #8627 (retired ids must not be silently rerouted to a model with different behaviour, capability, or cost). So any client that builds its model picker from `/v1/models` never sees the alias.

**Attempt 2, use a combo as the stable id.** Combos *are* advertised in `/v1/models`, combos support `system_message` (which gives me the prompt injection), and combos let me swap members without touching clients. This is the only mechanism I found that gives all three properties.

**The shape I landed on,** keep the public id stable and push the volatile part one level down:

```
preset/default                                  <- stable id my apps use (priority, system_message set)
  ─ combo-ref → preset/deepseek-flash-latest   <- the swap point
                   ├─ qwen-cloud-token-plan/deepseek-v4.1-flash
                   ├─ command-code/deepseek/deepseek-v4.1-flash
                   └─ opencode-go/deepseek-v4.1-flash
```

**Y, the bug this exposed:** text works, images don't. The image is dropped at the `combo-ref` hop.

## Reproduction

```bash
curl http://127.0.0.1:20128/v1/chat/completions \
  -d '{"model":"preset/default","max_tokens":8,"messages":[{"role":"user","content":[{"type":"text","text":"ok"},{"type":"image_url","image_url":{"url":"data:image/png;base64,<...>"}}]}]}'
```

## Evidence

Same image, same prompt, only `model` differs. `prompt_tokens_details.image_tokens` is the tell:

| model | status | image_tokens | text_tokens | prompt_tokens |
|---|---|---|---|---|
| `qwen-cloud-token-plan/deepseek-v4.1-flash` (leaf) | 200 | 434 | 338 | 772 |
| `preset/deepseek-flash-latest` (flat combo) | 200 | 434 | 338 | 772 |
| `preset/default` (nested combo) | 200 | **absent** | 384 | 384 |

The leaf model is identical in all three rows. Flat combos preserve the image, and the only nested combo on this instance loses it. Asking the leaf to describe the image works when the leaf is called directly, and returns "no vision provider connected" when called through `preset/default`.

I also found that #4536 documents that combo-refs are historically **flattened** before routing, so this may live in that flatten path.

## Related: capability metadata contradicts itself

For the same combo the two catalog endpoints disagree:

- `GET /v1/models` → `preset/default` reports `capabilities.vision: true`
- `GET /v1/combos` → `preset/default` reports `capabilities.multimodal: false` (the `combo-ref` isn't expanded into the multimodal intersection)

A client keyed on `/v1/models` believes it may send images, a client keyed on `/v1/combos` believes it may not, and neither request errors. Related: #13847, #12799.

### Steps to Reproduce

1. Have a combo whose member is a `combo-ref` (mine: `preset/default` → `preset/deepseek-flash-latest`).
2. POST an image to that combo:
```bash
curl http://127.0.0.1:20128/v1/chat/completions \
  -d '{"model":"preset/default","max_tokens":8,"messages":[{"role":"user","content":[{"type":"text","text":"ok"},{"type":"image_url","image_url":{"url":"data:image/png;base64,<...>"}}]}]}'
```
3. Inspect `usage.prompt_tokens_details.image_tokens` in the response.
4. Repeat the identical request against the leaf model id and against the inner flat combo.

### Expected Behavior

The image reaches the leaf model, and `usage.prompt_tokens_details.image_tokens` is populated, exactly as it is for a flat combo or a direct leaf call. A combo that advertises `capabilities.vision: true` in `/v1/models` should actually deliver the image to its targets.

### Actual Behavior

The image is silently dropped at the `combo-ref` hop. The response is a normal 200 with no error, but `prompt_tokens_details.image_tokens` is absent and the text token count is lower than the flat-combo equivalent. The model then replies that it cannot see an image.

Because nothing errors, any client that relies on the model catalog to decide whether to send images degrades to text-only without a signal. In my case the effect was that every image the connected app sent came back as "no vision-capable provider connected", with no indication that the gateway was at fault.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
Instrumented response for `preset/default` (image sent, image dropped):


{"usage":{"prompt_tokens":384,"completion_tokens":81,"total_tokens":465,"prompt_tokens_details":{"cached_tokens":0,"text_tokens":384}}}


Identical request against the flat inner combo `preset/deepseek-flash-latest` (image preserved):


{"usage":{"prompt_tokens":772,"completion_tokens":9,"total_tokens":781,"prompt_tokens_details":{"cached_tokens":0,"image_tokens":434,"text_tokens":338}}}


`image_tokens` present in the second, absent in the first; `prompt_tokens` 384 vs 772 for the same image and prompt.
```

### Screenshots

No error is raised, which is the core problem. Full request is a single POST to `/v1/chat/completions` with a `data:image/png;base64,` image part; nothing is logged at warn/error level. The only signal is the absent `image_tokens` field in `usage`.

## Why a single-entry combo (and a thought on special handling)

This instance has 13 combos. Exactly one is single-entry, and its only entry is the `combo-ref`: `preset/default`. That is the one that breaks.

**Why I built it that way.** I want a stable public id that my apps can point at, the ability to inject a system prompt there, and the freedom to swap the underlying model later without touching any client. `preset/default` is a pure indirection layer: it holds the `system_message` and the stable name, and the volatile part (which models, which order, which accounts) lives one level down in `preset/deepseek-flash-latest`. That way I re-point `preset/default`'s single ref when I want to change models, and everything downstream keeps working.

**Special handling for single-entry combos.** Worth considering, because a single-entry combo is unambiguous: there is no strategy decision to make, no failover to express, and no ordering to resolve. A combo with exactly one member is by definition a pass-through, and I'd expect `execute` and `flatten` to be indistinguishable for it. So if the `combo-ref` hop is where the image is lost, single-entry combos are the cheapest place to make the behaviour correct, because collapsing them cannot change routing semantics at all.

That said, the flat inner combo and the leaf both preserve images, so I suspect the fix belongs in the expansion path generally rather than as a special case. Mentioning it because a single-entry `combo-ref` is a natural fixture for the regression test.

**Capability metadata also disagrees with itself.** For the same combo:
- `GET /v1/models` → `preset/default` reports `capabilities.vision: true`
- `GET /v1/combos` → `preset/default` reports `capabilities.multimodal: false` (the `combo-ref` is not expanded into the multimodal intersection)

A client keyed on `/v1/models` believes it may send images, a client keyed on `/v1/combos` believes it may not, and neither request errors. Related: #13847, #12799, #4536.

### Validation Plan

A unit test over combo target expansion: given a combo whose only member is a `combo-ref`, assert the expanded target list preserves `image_url` parts. Assert `usage.prompt_tokens_details.image_tokens` is non-zero for a combo-ref-routed request and equals the value from the same request sent to the inner combo directly. A flat-vs-nested fixture pair catches this regression, since the flat path already behaves correctly.

diegosouzapw:
## Summary

`getComboVisionBridgeDecision()` treated any top-level `combo-ref` step in a combo's model list as an unconditional `"process"`, without ever resolving the referenced combo to check whether its real leaf targets support vision. For a pass-through combo whose only member is a `combo-ref` to another combo made entirely of vision-capable models (the reporter's `preset/default -> preset/deepseek-flash-latest` shape), this wrongly triggered the describe-and-replace path. With no vision-bridge describer model configured, every image got replaced with the literal stub `"[Image N]: (unavailable — no vision-capable provider connected)"` — the exact text the reporter saw the model answer with.

**Fix**: `getComboVisionBridgeDecision` now recursively resolves each `combo-ref` step to its real leaf models (depth-guarded by the same `MAX_COMBO_DEPTH` constant used by the flatten dispatch path in `open-sse/services/combo/comboStructure.ts`, plus a visited-set cycle guard) and folds their proven vision capability into the same accumulation used for direct `model` steps. An unresolvable combo-ref (combo not found, empty models, circular chain, or depth exceeded) is conservatively treated as a single non-vision-capable leaf rather than forcing the whole combo back to `"process"` — this keeps the safe-by-default behavior for genuinely unknown nesting while fixing the common all-vision-capable case.

Closes #13927

## Root cause

- `src/lib/guardrails/visionBridge.ts` (before fix): `if (s.kind === "combo-ref") return "process";` fired on the first combo-ref step and returned immediately, never recursing into the referenced combo's own model list.

## Regression test

`tests/unit/combo-ref-vision-bridge-13927.test.ts` (new file, 5 cases):
1. nested combo-ref to an all-vision-capable inner combo → `skip` (the reported bug)
2. combo-ref to a mixed-capability inner combo → `process`
3. combo-ref to a zero-vision inner combo → `no-vision`
4. circular combo-ref chain (no real leaf) → falls back safely to `no-vision`, never throws
5. circular combo-ref mixed with a real vision-capable leaf → `process`

RED (before fix, case 1):
```
✖ #13927: nested combo-ref to an all-vision-capable inner combo should resolve to skip, not process
  AssertionError [ERR_ASSERTION]: 'process' !== 'skip'
```

GREEN (after fix): all 5 cases pass.

## Gates run

- `node scripts/check/check-file-size.mjs` — OK, no violation on touched files (pre-existing `tests/unit/chatcore-translation-paths.test.ts` violation is unrelated/untouched)
- `npm run check:open-sse-typecheck` — OK, 0 errors (touched file imports a constant from `open-sse/`)
- `node --import tsx/esm --test tests/unit/combo-ref-vision-bridge-13927.test.ts` — 5/5 pass
- `node --import tsx/esm --test tests/unit/guardrails/visionBridge-combo-reroute.test.ts` (existing, decision-matrix coverage) — 12/12 pass, no regressions
- `node --import tsx/esm --test tests/unit/guardrails/combo-vision-provider-id-12112.test.ts` (existing, providerId resolution) — 2/2 pass, no regressions
- pre-commit hooks (lint-staged, docs-sync, any-budget, tracked-artifacts) — all passed


diegosouzapw:

PR #14040, opened 2026-09-17
fix(db): cache getPricingForModel() to stop /api/usage/history event-loop stall (#13891)

### LINKED ISSUE #13891: fix(backend): /api/usage/history blocks the event loop ~5s — uncached getPricing() rebuild per row
### What is happening

`GET /api/usage/history` (backed by `getUsageStats()` in `src/lib/usage/usageStats.ts`) takes **~5s per call and blocks the Node event loop for the entire duration** — stalling all concurrent traffic on a single-process deployment (LLM proxy streams, `/healthz`, everything).

Measured on v3.8.49 (Docker, ~45k rows in `usage_history`):

```
usage/history:        4895ms (single call, idle system)
healthz DURING call:  4841ms   <- event loop fully blocked
healthz otherwise:    3-5ms
```

### Root cause

The SQL is not the problem (31ms for the grouped scans, all indexes present). The cost is in JS:

1. `getUsageStats()` runs three GROUP BY scans over `usage_history` (byModel / byAccount / byApiKey) → in our DB that yields **531 groups** (163 + 254 + 114).
2. For **every group** it awaits `calculateAggregateCost()` → `calculateCost()` → `getPricingForModel()` (src/lib/usage/costCalculator.ts:186).
3. `getPricingForModel()` (src/lib/db/settings/pricing.ts:123) calls **`getPricing()` uncached** — which re-runs `getPricingLayers()`: 3 fresh `SELECT`s against `key_value` (defaults + `pricing_synced` (100 rows) + `models_dev_pricing` (223 rows) + user overrides), re-`JSON.parse`s every row, and re-merges all layers — **531 times per request**.
4. `readCache.ts:88` already provides `getCachedPricing()` ("Longer TTL since pricing rarely changes") — but `getPricingForModel()` does not use it. `usageStats.ts` also re-computes everything on every poll with no response-level cache.

### Impact

- Any dashboard poll of `/api/usage/history` hard-blocks the gateway for ~5s; with a 5s poll interval the event loop is effectively saturated continuously while that page is open.
- On self-hosted single-container setups this stalls live inference streaming, health probes (k8s probe flaps, see closed #10052), and every management API call.

### Suggested fixes (any one of these removes most of the stall)

1. **Route `getPricingForModel()` through the existing `getCachedPricing()`** (5s TTL is plenty; `invalidateDbCache("pricing")` already busts it on pricing writes) — turns 531 pricing rebuilds into 1.
2. Memoize cost per `(provider, model, service_tier)` within a single `getUsageStats()` run — the 531 groups collapse to a few dozen distinct rate lookups.
3. Add a short server-side response cache for `/api/usage/history` (e.g. 30–60s) — the dashboard polls it but data is write-append only.
4. (Optional) `calculateCost`/`calculateAggregateCost` don't need to be `async` once pricing is cached — removing the per-group `await` also avoids interleaving overhead.

### Environment

- OmniRoute v3.8.49 (official image), Docker Desktop, single container
- `aggregation.enabled=true`, `retention.usageHistory=365`, `usage_history` ≈ 45,915 rows
- Repro: `time curl -H "Authorization: Bearer <key>" http://127.0.0.1:18081/api/usage/history` while pinging `/healthz` every 250ms

diegosouzapw:
Closes #13891

## Root cause

`getPricingForModel()` (`src/lib/db/settings/pricing.ts`) called the uncached
`getPricing()` on every invocation instead of the `getCachedPricing()` helper
that already exists in `src/lib/db/readCache.ts` (30s TTL, invalidated on
every pricing write via `touchPricing()` -> `invalidateDbCache("pricing")`).

`usageStats.getUsageStats()` calls `calculateAggregateCost()` -> `calculateCost()`
-> `getPricingForModel()` once per GROUP BY row — up to 531 times on the
reporter's 45k-row `usage_history` table — so every `/api/usage/history` request
re-ran a synchronous 3-SELECT + `JSON.parse` + merge cycle against `key_value`
hundreds of times, blocking the event loop for ~5s.

Fix: swap the internal call inside `getPricingForModel()` from `getPricing()` to
`getCachedPricing()`. The exported `getPricing()` / `getPricingWithSources()`
functions are untouched — callers that need a guaranteed-fresh read (pricing
admin route, sync jobs) keep using them directly.

I audited every pricing writer (`pricing.ts`, `pricingSync.ts`,
`modelsDevSync.ts`) — all three namespaces (`pricing`, `pricing_synced`,
`models_dev_pricing`) already call `invalidateDbCache("pricing")` on write, so
this does not introduce any new staleness window beyond the existing 30s TTL
that other callers of `getCachedPricing()` already tolerate.

## Regression test

`tests/unit/pricing-for-model-cached-13891.test.ts`

- RED (pre-fix, run against unfixed code): 20 calls to `getPricingForModel()`
  triggered 60 `db.prepare()` calls (3 SELECTs × 20, zero caching).
- GREEN (post-fix): the same 20 calls trigger ≤3 `db.prepare()` calls.
- A second test proves cache invalidation: a lookup before `updatePricing()`
  returns the old value, and the very next lookup after the write returns the
  fresh value — confirming a stale price cannot silently persist past a write.

```
node --import tsx/esm --test tests/unit/pricing-for-model-cached-13891.test.ts
# tests 2, pass 2, fail 0
```

## Gates run

- `node scripts/check/check-file-size.mjs` — no violation on touched files
- `node scripts/check/check-complexity.mjs` — OK (2857 ≤ baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1291 ≤ baseline 1437)
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json src/lib/db/settings/pricing.ts tests/unit/pricing-for-model-cached-13891.test.ts` — exit 0
- New test: 2/2 pass (RED→GREEN evidence above)
- Existing pricing regression suite — all pass:
  `pricing-sync-memoization.test.ts` (3/3), `pricing-route-sources.test.ts` (1/1),
  `pricing-sync-extended.test.ts` (6/6), `tier-pricing-cache.test.ts` (4/4),
  `db-read-cache.test.ts` (10/10)


diegosouzapw:

PR #14041, opened 2026-09-17
fix(providers): convert agent_message input items for non-Codex-native Responses upstreams (#13698)

### LINKED ISSUE #13698: fix(providers): OpenCode Go Muse Spark 1.3 rejects Codex `agent_message` input items on the Responses endpoint (400)
I am facing the same issue as #12674 on OmniRoute 3.8.50. Muse Spark 1.3 contributor works fine directly in OpenCode, but fails when called through OmniRoute. I was able to fix it with the mapping below.

### Issue

The model is available in the catalog, but a request to `opencode-go/muse-spark-1.3-contributor` returns:

```text
[opencode-go/muse-spark-1.3-contributor] [500]: Internal server error (reset after 3s)
```

To reproduce, send a chat completion request using that model with a simple message such as `Reply exactly MUSE_CHECK_OK.`

The dynamic catalog includes the model, but the static target-format lookup has no mapping for it. `getModelTargetFormat()` returns null and `resolveOpencodeTargetFormat()` falls back to `openai`. The executor then selects Chat Completions instead of the Responses endpoint required by this model.

### Patch that fixed it

In `open-sse/config/providerModels.ts`, inside `getModelTargetFormat()`, add this after the existing explicit `targetFormat` check:

```ts
if (found?.targetFormat) return found.targetFormat;

if (
  alias === "opencode-go" &&
  bareModelId === "muse-spark-1.3-contributor"
) {
  return "openai-responses";
}
```

This keeps existing explicit mappings first and only changes the OpenCode Go model. It also covers provider-prefixed IDs after the existing prefix normalization.

I applied the equivalent change to the compiled runtime as a build-time patch, since changing the source file alone does not update an already-built image.

### Tested

- The same request failed before the patch and returned `MUSE_CHECK_OK` after it.
- Normal and streaming-mode requests both returned the expected replies without fallback.
- Checks for provider-prefixed IDs passed.
- Other providers and DeepSeek routing were unchanged by the mapping.

Can this mapping be included upstream so the model works without maintaining a separate patch? This fix covers the base model ID; I have not tested the effort-tier variants.

diegosouzapw:
Closes #13698

## Root cause

The shared Responses-API input sanitizer (`open-sse/services/responsesInputSanitizer.ts::sanitizeInputItem`) had no branch for Codex's proprietary `agent_message` input-item type, so it passed such items through untouched. OpenCode Go Muse Spark 1.3 implements a strict/plain OpenAI Responses endpoint that does not know this Codex-only type and rejects the request with `input[N] did not match any supported type`.

## Fix

- `sanitizeInputItem` now converts an `agent_message` item (`type: "agent_message"` or `role: "agent_message"`) into a plain `{type: "message", role: "user", ...}` item before it reaches any non-Codex-native Responses upstream — a leading context part carries `author`/`recipient`, followed by the original content flattened to input-safe parts (string/array/non-array content, `encrypted_content` → placeholder, `image_url`/`input_image` preserved via the existing `sanitizeContentPart`).
- A new `preserveAgentMessages` option opts a call site out of the conversion. `CodexExecutor` (the only caller that also passes `dropInternalAssistantMessages`) now derives both flags from the same `nativeCodexPassthrough` boolean via a new `sanitizeCodexResponsesInput()` helper — this keeps the real Codex/ChatGPT native passthrough path receiving `agent_message` untouched (it understands the type), while every other Responses executor (base/OpenCode/GitHub) picks up the new default (convert) with no call-site change.
- The helper was extracted into `responsesInputSanitizer.ts` instead of inlined in `open-sse/executors/codex.ts` because that file is at its frozen file-size cap (1530 lines) — the extraction keeps `codex.ts` at 1525 lines (net shrink) instead of growing it.

## Regression test

`tests/unit/responses-input-sanitizer-agent-message.test.ts` (9 cases): type-based and role-based conversion with context preserved, string content, encrypted_content placeholder, image_url/input_image preservation, non-array/non-string content fallback, `preserveAgentMessages: true` no-op (native Codex), idempotence, and the default (no options) call-site path used by base/GitHub executors.

RED (pre-fix, via the plan-file's probe):
```
✖ #13698: sanitizeResponsesInputItems leaves an agent_message item untouched (BUG)
AssertionError [ERR_ASSERTION]: expected sanitizer to convert agent_message to a supported Responses item type
  actual: 'agent_message'
  expected: 'agent_message'
```

GREEN (post-fix):
```
tests 9
pass 9
fail 0
```

Also re-ran the existing suites touching this sanitizer / Codex's native-passthrough path to confirm no regression: `tests/unit/responses-input-sanitizer-name.test.ts`, `tests/unit/codex-system-message-image-8089.test.ts`, `tests/unit/executor-codex.test.ts`, `tests/unit/executor-codex-responses-lite-translated-path.test.ts`, `tests/unit/responses-chat-translation-gaps.test.ts`, `tests/unit/codex-drop-nonstandard-events.test.ts` — 89/89 pass, including the case exercising `nativeCodexPassthrough` (`preserves compact requests and native passthrough semantics`).

## Gates run

- `node scripts/check/check-file-size.mjs` → OK (no ✗ on touched files; `codex.ts` 1525 < 1530 frozen cap)
- `node scripts/check/check-complexity.mjs` → OK — 2857 violations (baseline 3218)
- `node scripts/check/check-cognitive-complexity.mjs` → OK — 1291 violations (baseline 1437)
- `npm run check:open-sse-typecheck` → OK, 0 errors
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json <changed files>` → clean
- Pre-commit hooks (lint-staged, docs-sync, any-budget, tracked-artifacts) → passed


diegosouzapw:

PR #14042, opened 2026-09-17
fix(quota): resolve Quota Sharing plan from pool's primary connection

### LINKED ISSUE #13876: fix(backend): quota-share dashboard consumed amount does not update (needs repro)
The Quota Sharing page does not change the "consumed" , even if the usage page can determine how many users and how they consumed the tokens.

diegosouzapw:
## Root cause

Multi-connection Quota Sharing pools resolve a DIFFERENT provider plan depending on which member connection actually served a request (write path) vs. the pool's primary connection (dashboard read path). The wizard's "Limite" step PUTs a manual plan override only to the primary connection (`connectionIds[0]`); any other pool member falls back to a different (catalog/empty) plan shape. Since the `quota_consumption` dimension key is `poolId:unit:window`, a different unit/window means `recordConsumption()` writes to a different bucket than the one the dashboard's usage route reads — real traffic never appears as "consumed" even though it genuinely happened.

## Fix

- `src/lib/quota/enforce.ts::enforceQuotaShare` — resolve the plan from `pool.connectionId` (the pool's canonical primary connection) instead of `input.connectionId` (the connection that actually served the request).
- `src/lib/quota/enforce.ts::recordConsumption` — keep the matched `QuotaPool` object (not just its id) from the pool-search loop, then resolve the plan from `matchedPool.connectionId` for the same reason.
- `getSaturation(input.connectionId, ...)` is left untouched — that signal is legitimately per-connection.

## Regression tests

- `tests/unit/quota-share-multiconn-plan-mismatch.test.ts` — RED before the fix (`AssertionError: dashboard should show the 5000 tokens actually consumed via conn-secondary, got 0` — `0 !== 5000`), GREEN after.
- `tests/unit/quota-share-multiconn-enforce-cap.test.ts` — companion test proving a cap configured on the primary connection's plan is now enforced for a request served by a non-primary pool member (also confirmed RED before / GREEN after during development).

```
node --import tsx/esm --test tests/unit/quota-share-multiconn-plan-mismatch.test.ts tests/unit/quota-share-multiconn-enforce-cap.test.ts
# tests 2, pass 2, fail 0
```

## Gates run

- `node scripts/check/check-file-size.mjs` — no violation on touched files.
- `node scripts/check/check-complexity.mjs` — OK (2857 ≤ 3218 baseline).
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1291 ≤ 1437 baseline).
- `npm run typecheck:core` — exit 0.
- Existing quota regression suite re-run and green (58 + 25 tests): `quota-sqlite-store`, `quota-pool-usage-equal-split`, `quota-pool-usage-summed-budget`, `quota-enforce-await-store`, `quota-enforce`, `quota-equal-split`, `quota-summed-budget`, `quota-multiprovider`, `quota-per-key-model`, `quota-per-key-model-hotpath`, `quota-spend-recorder`, `combo-quota-soft-penalty`, `quota-epsilon-unconfigured-allow`, `serial/quota-division-blocks`, `quota-group-allocations`.

Closes #13876


diegosouzapw:

PR #14043, opened 2026-09-17
fix(providers): map agentrouter GLM thinking.type adaptive to enabled

### LINKED ISSUE #13696: fix(providers): agentrouter GLM models reject thinking.type "adaptive" (400) — map to enabled for glm-*
## Problem

GLM models served through AgentRouter's OpenAI-compatible format fail with a 400 when the client sends Anthropic-style adaptive thinking:

```
[400]: thinking.type "adaptive" is not supported by glm models; must be one of enabled, disabled
```

`adaptive` is Anthropic's thinking mode. GLM's OpenAI-compatible API also has a `thinking` field but only accepts `enabled` or `disabled`. Clients built around Claude send `thinking: {type: "adaptive"}`, and on AgentRouter's `openai` alternate format OmniRoute forwards it unchanged.

## Evidence

Discussion #13346 by @SL-in: `agentrouter/glm-5.3` failed from every agent client while the same request worked from OpenCode.

1. On the default AgentRouter format (Anthropic Messages at `/v1/messages`, with the Claude Code wire image), the request failed with an upstream 500 "Upstream rejected the request as invalid".
2. After switching the connection's target format to the OpenAI-compatible alternate, the error became the 400 above.

`agentrouter` is declared `passthroughModels: true` with only `claude-opus-4-8`, `claude-opus-5` and `gpt-5.6-sol` in its static catalog (`open-sse/config/providers/registry/agentrouter/index.ts`), so GLM ids reach upstream with no model-specific handling.

## Precedent

`open-sse/translator/paramSupport.ts:43-48` already handles the same clash for NVIDIA NIM's GLM wrapper (`{ provider: "nvidia", match: /z-ai\/glm-5\.2\b/i, drop: ["reasoning", "thinking"] }`, 9router#2023).

## Proposed solution

1. Add a `paramSupport` rule for `agentrouter` + `/glm-/i` that **maps** `thinking.type: "adaptive"` to `"enabled"` rather than dropping it, so reasoning stays on when the client asked for it. Fall back to dropping if the existing mechanism only supports drops.
2. Add `glm-*` entries to the AgentRouter catalog with the OpenAI target format, so GLM routing works without a manual per-connection format switch.

## Workaround until then

A payload rule under Settings -> Advanced -> Payload Rules:

```json
{
  "override": [
    { "models": [{ "name": "glm-*", "protocol": "agentrouter" }], "params": { "thinking.type": "enabled" } }
  ]
}
```

## Acceptance criteria

- [ ] A unit test sends `thinking: {type: "adaptive"}` for `agentrouter/glm-5.3` and asserts the upstream body carries `enabled` (or no `thinking`), while `agentrouter/claude-opus-5` keeps `adaptive`.
- [ ] Live validation of one GLM request through AgentRouter recorded in the PR (Hard Rule #18).

## Additional context

- Discussion: #13346
- Related: #13397 (other AgentRouter failures on 3.8.50, different models)

diegosouzapw:
## Root cause

AgentRouter routes GLM models through the generic `DefaultExecutor` (executor: `"default"`), which has no GLM-specific handling. When the connection's OpenAI-compatible alternate format is used and the client request already carries a Claude-style `thinking: {type: "adaptive"}` field (source format == target format == `"openai"`, so `translateRequest`'s hub-and-spoke conversion is skipped), that field survived `stripUnsupportedParams` untouched — `STRIP_RULES` in `open-sse/translator/paramSupport.ts` had no rule for provider `agentrouter` + `glm-*` — and reached AgentRouter's upstream GLM endpoint verbatim, which 400s with `thinking.type "adaptive" is not supported by glm models; must be one of enabled, disabled`.

The native GLM/ZAI executor (`open-sse/executors/glm.ts:404-420`) already performs exactly this `adaptive` → `enabled` remap for its own provider; AgentRouter just never reaches that code path since it uses the generic `DefaultExecutor` instead of `GlmExecutor`.

## Fix

Extended the `StripRule` shape in `open-sse/translator/paramSupport.ts` with an optional `mapThinkingType` map (alongside the existing `drop`/clamp mechanisms), and added a rule scoped to `provider: "agentrouter"` + `match: /glm-/i` that remaps `thinking.type` from `"adaptive"` to `"enabled"`, preserving any other `thinking` fields (e.g. `budget_tokens`) instead of dropping the whole field. Scoped to the whole `glm-*` family (not a single model id), since AgentRouter's static catalog carries no `glm-*` entries at all — `passthroughModels: true` lets arbitrary GLM ids through, so the fix must not be tied to one model.

## Regression test

`tests/unit/agentrouter-glm-thinking-adaptive-map.test.ts` (new):
- `agentrouter` + `glm-5.3` with `thinking.type: "adaptive"` → asserts `thinking` becomes `{type: "enabled", budget_tokens: ...}` (other fields preserved)
- `agentrouter` + `claude-opus-5` (regression guard) → `thinking.type` stays `"adaptive"`
- non-`agentrouter` provider with a `glm-*` model id → `thinking.type` stays `"adaptive"` (native GLM executor handles its own mapping separately)
- `agentrouter` + `glm-5.3` with no `thinking` field → no-op, nothing introduced

RED (before fix): `AssertionError: adaptive must be mapped to enabled for agentrouter GLM models, preserving other thinking fields` (actual `{type: 'adaptive', budget_tokens: 1024}` vs expected `{type: 'enabled', budget_tokens: 1024}`).

GREEN (after fix): all 4 new tests pass; `npm run test:unit` equivalent focused run — `node --import tsx/esm --test tests/unit/agentrouter-glm-thinking-adaptive-map.test.ts` → 4/4 pass.

Also updated the existing structural assertion in `tests/unit/executors-strip-unsupported-params.test.ts` ("STRIP_RULES is non-empty and every rule has a drop list or a clamp mechanism") to also accept the new `mapThinkingType` mechanism — it previously only recognized drop/clamp rules and would otherwise fail on the new rule shape. This is an alignment to the corrected contract (a `StripRule` may now sanitize by remapping a value, not only by dropping/clamping), not a weakening: the assertion still requires every rule to do *something* (drop, clamp, or map), just enumerates the new legitimate option.

## Gates run

- `node scripts/check/check-file-size.mjs` — OK (no violation on touched files; the reported test-file-size violation on `tests/unit/chatcore-translation-paths.test.ts` is pre-existing/untouched by this PR)
- `npm run check:open-sse-typecheck` — OK, 0 errors
- Existing param-support tests re-run individually (DB-per-file to avoid SQLITE_BUSY contention): `azure-max-output-clamp`, `chatcore-unsupported-params-strip`, `default-reasoning-effort-6879`, `executors-strip-unsupported-params`, `glm-executor-max-tokens-clamp-7364`, `nvidia-minimax-thinking-strip`, `nvidia-strip-prompt-cache-key`, `param-filters-apply`, `param-filters-db`, `zai-glm-max-tokens-clamp-7364` — all pass

## Live validation (Hard Rule #18)

Not performed against the production VPS in this pass — the fix is validated via TDD (a failing-then-passing regression test proving the exact reported 400 condition), which is the preferred validation path per the bug-fix protocol. The reported error string and root cause were confirmed by the prior triage's TDD probe and by the existing in-repo precedent (`glm.ts`'s identical adaptive→enabled mapping for its own provider, and the identical NVIDIA NIM `z-ai/glm-5.2` clash already fixed in `paramSupport.ts`).

Closes #13696


diegosouzapw:

PR #14044, opened 2026-09-17
fix(db): rotate proxy pools on the chat path like the registry does

### LINKED ISSUE #13575: fix(resilience): Proxy pools can't pick another member on the chat path: the per-connection proxy cache freezes the choice
**Current state** — in the default configuration this report describes the behavior as it is. With `PROXY_SKIP_RECENTLY_FAILED` enabled, a member that has been set aside is no longer served from the cache, while a member that keeps working stays pinned; the cache key is `connection:apiKey:provider`, not the connection alone. Proposed fix, and the two populations it has to respect: [this comment](https://github.com/diegosouzapw/OmniRoute/issues/13575#issuecomment-5702645413).

---

### OmniRoute Version

3.8.51 (`release/v3.8.51`, `152d95108c`)

### Installation Method

Built from source

### Operating System

Linux

### Provider(s) Involved

any (the pool is resolved before the provider call)

### Description

A proxy pool is meant to let OmniRoute choose among several egress routes, but on the chat path that choice is made once per connection and then frozen. Chat requests resolve their proxy through `resolveProxyForConnection`, which caches the result per connection (`src/lib/db/settings.ts:506-525`, stored at `:647`, `:671`, `:701` and `:748`). The pool strategy from #6395 runs on the first request; after that the cached member comes back, whatever the strategy or the state of that member.

The cache is cleared only when the proxy config generation is bumped (`settings.ts:39-41`). That happens on settings and registry writes, and also on any connection state write: `updateProviderConnection`, `resetConnectionBackoff`, the rate-limit write (`src/lib/db/providers/rateLimit.ts:305`). So the member a connection uses changes when some unrelated connection hits an error or a cooldown, for every connection at once, and never when this connection's own situation would call for it. On a quiet instance it doesn't change at all.

The existing tests call `resolveProxyForScopeFromRegistry` directly (`tests/unit/proxy-pool-rotation-6365.test.ts:75-78`), so they don't go through the cache.

### Steps to Reproduce

1. Create three proxies and add them to an `account` pool for one connection (`addProxyToScopePool`).
2. Call `resolveProxyForConnection(connectionId)` six times.
3. Call `resolveProxyForScopeFromRegistry("account", connectionId)` six times.

```ts
for (let i = 0; i < 6; i++) {
  hosts.push((await resolveProxyForConnection(conn.id)).proxy?.host);
}
// ["10.9.0.1", "10.9.0.1", "10.9.0.1", "10.9.0.1", "10.9.0.1", "10.9.0.1"]

for (let i = 0; i < 6; i++) {
  hosts.push((await resolveProxyForScopeFromRegistry("account", conn.id)).proxy.host);
}
// cycles through the three hosts
```

### Expected Behavior

#6365 asks for outbound requests to be able to "rotate their egress IP (e.g. a new IP each hour / per request)", and the pool UI offers `round-robin`, `random` and `sticky`. At minimum, a request should be able to get a different member when there's a reason to: the strategy says so (a `sticky` window ended), or the member it would get is known to be unusable.

### Actual Behavior

One member per connection, re-chosen only when any settings, registry or connection state write clears the whole cache.

### Test Impact

Needs a new unit test

A test that goes through `resolveProxyForConnection` fails today; the registry-level tests keep passing.

### Question before a fix

What should the default be? I'm not asking for a new IP on every request: changing egress for no reason would break setups that count on a stable IP per account (OAuth accounts in particular). The point is to make re-selection possible, and leave the decision to a rule.

My suggestion is a pool setting, off by default, stored next to the strategy and the sticky window in `proxy_scope_rotation`: "re-evaluate the member on every request". Off, a connection keeps its member as today. On, each chat request asks the pool again, and the pool answers with the same member unless something says otherwise: the strategy (`sticky` holds a member for its window, not for the life of the cache), or a member known to be unusable. Background callers (token refresh, warmup) would keep the cached member so they don't consume turns. Since pools are already per scope, that gives a per-provider switch (provider pool), a global one (global pool), and per-account or per-combo ones with no separate setting.

The alternative is a single server-wide switch, simpler but all or nothing.

One detail if re-selection happens on every request: the cursor update is read-then-write today (`src/lib/db/proxies/rotation.ts:98-128`, `:179-183`), so it would need an atomic `INSERT … ON CONFLICT DO UPDATE … RETURNING`. Happy to send the PR once you pick.

### Related

- #13578 stops the per-connection cache from re-serving a pool member that just failed. That's one rule of the kind described above, applied to a single case; the question here is about letting the pool decide in general.

diegosouzapw:
Closes #13575

## Root cause

`resolveProxyForConnection` (`src/lib/db/settings.ts`) caches the first pool-resolution result for a connection and, outside the narrow `#13578` "member was just set aside" escape hatch, returns that same cached member on every subsequent chat request — regardless of the pool's rotation strategy (round-robin/sticky/random). `resolveProxyForScopeFromRegistry`, which every existing rotation test (`#6365`) calls directly, re-runs the strategy on every call and rotates correctly. The chat path is the only consumer whose freeze bypasses that.

## Fix

Extend the cache-validity check with a second term alongside the existing `isCachedPoolMemberSetAside` escape hatch: a cached result is now also stale (falls through to the same cascade the direct registry callers use) whenever it came from a live scope pool (`source: "registry"`) **and** the connection does not need a stable egress. Two populations are excluded and keep their pinned member exactly as before:

- `EGRESS_BUCKETED_LOCK_PROVIDERS` (`open-sse/config/providerErrorRules.ts`) — opencode's free-tier quota is bucketed by egress IP; rotating would fragment one connection's quota across several IPs.
- `grok-web` — its `cf_clearance` cookie is pinned to the IP/User-Agent/TLS fingerprint that earned it (`src/shared/providers/webSessionCredentials.ts`); rotating the egress turns every subsequent request into a Cloudflare 403.

No new rotation strategy is introduced — the pool's existing round-robin/sticky/random logic in `src/lib/db/proxies/rotation.ts` is unchanged; the chat path now simply reaches it the same way the registry-direct callers already do.

## Regression test

`tests/unit/proxy-pool-chat-path-rotation-13575.test.ts` (new):

- RED before the fix: `resolveProxyForConnection` returned the same host 6/6 times for a 3-member `account`-scope pool.
- GREEN after: rotates across all 3 members; an `opencode` connection and a `grok-web` connection each stay pinned to a single member across 6 calls.

```
node --import tsx/esm --test tests/unit/proxy-pool-chat-path-rotation-13575.test.ts
```

`tests/unit/proxy-pool-skips-refused-member.test.ts` (updated): three of its assertions encoded the old frozen-cache contract as correct behavior (`resolveProxyForConnection` returning the exact same cached object on a second call with no set-aside event). Updated those three to the corrected always-rotates-except-pinned contract; every other case in that file, and every case in `tests/unit/proxy-pool-rotation-6365.test.ts`, passes unchanged.

## Gates run

- `node --import tsx/esm --test tests/unit/proxy-pool-chat-path-rotation-13575.test.ts tests/unit/proxy-pool-skips-refused-member.test.ts tests/unit/proxy-pool-rotation-6365.test.ts` — 23/23 pass
- Full existing proxy-resolution suite (`8385-perkey-proxy-global-toggle`, `combo-scope-proxy-dead-7149`, `db-proxies-crud`, `db-proxies-split`, `db-settings-crud`, `db-settings-extended`, `db-settings-split`, `fixes-p1`, `issue-13470-token-refresh-proxy-bypass`, `proxy-assigned-unavailable-6246`, `proxy-noauth-provider-6272`, `proxy-registry`, `proxy-resolution-status-filter`, `proxySubscription.service`, `repro-8995`, `resolve-proxy-family`) — 196/196 pass
- `node scripts/check/check-file-size.mjs` — no ✗ on touched files
- `npm run check:open-sse-typecheck` — 0 errors (settings.ts imports from `open-sse/config/providerErrorRules.ts`)
- `npx eslint --suppressions-location config/quality/eslint-suppressions.json` on all changed files — 0 errors
- `changelog.d/fixes/13575-proxy-pool-chat-path-rotation.md` added


diegosouzapw:

PR #14047, opened 2026-09-17
fix(sse): stop direct fetch retry reusing pooled flat response-start budget (#13703)

### LINKED ISSUE #13703: fix(resilience): flat 30s direct response-start watchdog 504s healthy slow-TTFB upstreams after ~60s (ignores timeout cascade, retry reuses same budget)
### Summary

Since v3.8.50, every **direct (no-proxy) egress** request is bounded by a flat **30 s response-start watchdog** (`open-sse/utils/directResponseStartTimeout.ts`, `DEFAULT_DIRECT_HEADERS_TIMEOUT_MS = 30_000`) with exactly **one retry on a fresh no-keep-alive socket** that gets the **same 30 s budget** (`open-sse/utils/proxyFetch.ts`, `maxAttempts = 2`). When a healthy upstream simply needs >30 s to produce its first byte — common for reasoning models and buffered gateways — the result is a **synthetic 504 after exactly ~60–61 s**, provider- and model-independent. The combo then fails over, wasting a full minute of latency and pushing load onto fallback targets.

Three concrete defects, all verified in `v3.8.50` and unchanged on current `main`:

1. **The watchdog ignores the entire timeout cascade.** OmniRoute has a well-designed precedence chain — connection `providerSpecificData.timeoutMs` → model registry `timeoutMs` → provider `getTimeoutMs()` → global `FETCH_TIMEOUT_MS` (`open-sse/handlers/chatCore/upstreamTimeouts.ts`). `resolveDirectHeadersTimeoutMs()` reads **only** `OMNIROUTE_DIRECT_HEADERS_TIMEOUT_MS` and consults none of them. An operator who raises a per-connection or per-model timeout still gets cut at 30 s.
2. **The fresh-socket retry reuses the same flat budget instead of the remaining target budget.** The zombie-socket rationale from #10214 only justifies 30 s for the *pooled* attempt (a pooled keep-alive socket that has silently died). The *retry* is by construction a brand-new socket — there is no zombie to detect, yet it gets another flat 30 s. With `DEFAULT_COMBO_TARGET_TIMEOUT_MS = 120_000` (`open-sse/services/comboConfig.ts`) the remaining ~90 s of target budget go unused.
3. **The failure is logged misleadingly.** When the watchdog fires, the executor catch logs `Fetch timeout after 600000ms on <url>` (`open-sse/executors/base.ts`, `err.name === "TimeoutError"` branch prints `this.getTimeoutMs()`), while the actual cut was 30 s. This sent us down the wrong path initially and will mislead any operator diagnosing it.

### Evidence (single production instance, one day: 2026-09-14, v3.8.50 stock)

**193 synthetic 504s out of 10,005 calls**, all in the 60.0–61.4 s band, across unrelated providers:

| provider (direct egress) | model | 504s | total | 200s | duration band |
|---|---|---|---|---|---|
| openai-compatible (Gemini via local CLIProxyAPI) | gemini-3.8-flash-high | 86 | 1607 | 1493 | 60.1–63.4 s |
| openai-compatible (api.apikey.fun) | gemini-3.8-flash | 90 | 902 | 811 | 60.1–64.7 s |
| openai-compatible | mimo-v2.5-pro | 7 | 82 | 65 | 85.1–100.8 s¹ |
| zai | glm-5.3 | 4 | 914 | 905 | 60.1–61.1 s |
| grok-cli | grok-4.6 | 3 | 918 | 887 | 60.1–60.9 s |

¹ a *different* watchdog (`Stream produced no non-ping SSE event within 80000ms`) — same TTFB class, listed for completeness.

Example call: `POST /v1/chat/completions`, requested `gemini-3.8-flash-high`, duration **61.4 s**, error body:

```json
data: {"error":{"message":"[504]: Direct response did not start within 30000ms — retrying on a fresh socket","type":"upstream_timeout","code":"DIRECT_RESPONSE_START_TIMEOUT"}}
```

Supporting observations:

- The upstream (CLIProxyAPI in front of Gemini) logged the aborted attempts as **499 (client abort) at ~28–30 s** — Google was still thinking, not down.
- Successful calls on the *same* targets the same day routinely show first-byte/total times of **26–30 s**, i.e. directly under the 30 s edge. This is slow-TTFB, not zombie sockets: the fresh-socket retry also times out, which per #10214's own contract means a genuinely slow response, not a dead pooled socket.
- Combo failover works (`failoverBeforeRetry: true`), so the client usually gets a 200 from the next target — but only after ~60 s of dead latency per attempt, and with unnecessary load shifted onto fallback models.

### Reproduction

Any direct-egress target whose TTFB exceeds 30 s reproduces this deterministically: send a streaming chat/completions request through a combo to a slow-first-byte upstream (or an artificial 45 s header delay). Timeline in the OmniRoute log:

```text
[ProxyFetch] Direct response-start timeout (30000ms) on pooled dispatcher — retrying on fresh no-keep-alive dispatcher: <host>
[ERROR] [504]: Direct response did not start within 30000ms — retrying on a fresh socket
{"tag":"TIMEOUT","msg":"Fetch timeout after 600000ms on <url>"}     <-- misleading: real cut was 30 s
```

Total: ~60–61 s, then combo failover.

### Proposed fix direction

1. Keep the flat 30 s (or `OMNIROUTE_DIRECT_HEADERS_TIMEOUT_MS`) budget **only for the pooled attempt** — that is the zombie-socket detection #10214 introduced.
2. For the **fresh-socket retry**, use the remaining request/target budget (e.g. derive from the combo `targetTimeoutMs` / the existing timeout cascade) instead of another flat 30 s. Env override stays a floor, never a ceiling cut.
3. Route the resolved budget through the existing cascade in `upstreamTimeouts.ts` so connection/model/provider overrides are honored by the direct layer.
4. Log the *actual* elapsed/budget values when the watchdog fires instead of `getTimeoutMs()` (600000), so operators can tell a 30 s direct-start cut from a genuine fetch timeout.

### Related

- #10214 / #10528 — introduced this watchdog (fixed a real stall); this issue is the over-firing follow-up, not a regression report against its purpose.
- PR #12906 — raises the ceiling to 180 s only when `reasoning_effort: high|max` is present in the body. That is a **subset**: our failing requests (e.g. `gemini-3.8-flash-high`, `grok-4.6`, `glm-5.3` without explicit effort) carry no such field. Attempt-budget allocation would fix the general case.
- #12861 / PR #12942 — the same timeout escaping as an unhandledRejection (crash). Different symptom, same function; fixing the budget here does not fix the crash there, but both touch the same code path and should be coordinated.
- #12655 — fetch-start cap in the executor layer (110 s); complementary layer, same TTFB problem class.

Environment: `v3.8.50` (docker, stock), verified unchanged on current `main` (`directResponseStartTimeout.ts` / `proxyFetch.ts`).

diegosouzapw:
Closes #13703

## Root cause

Three defects were confirmed in `release/v3.8.51`:

1. `resolveDirectHeadersTimeoutMs()` (`open-sse/utils/directResponseStartTimeout.ts`) only read `OMNIROUTE_DIRECT_HEADERS_TIMEOUT_MS`.
2. **The actionable bug**: `open-sse/utils/proxyFetch.ts`'s fresh-socket retry loop resolved `directHeadersTimeoutMs` once and passed the *identical* value to `directFetchWithBoundedResponseStart` on both the pooled attempt (attempt 0) and the fresh-socket retry (attempt 1). #10214's zombie-socket rationale only justifies a short watchdog on the pooled attempt (a possibly-dead keep-alive socket); the retry is by construction a brand-new socket with nothing to detect, so it should not be cut by the same short flat window when the caller already has a much larger deadline — this doubled the effective flat budget (60s in production for the reported case) and 504'd healthy slow-TTFB reasoning models.
3. `open-sse/executors/base.ts`'s `TimeoutError` catch logs the outer `fetchStartTimeoutMs`, not the actual direct-path budget.

This PR fixes defect #2, the one with a proven TDD repro and the one causing the reported 504s. Defects #1 and #3 are out of scope for this surgical fix (not proven necessary by the failing test, and #1 is already partially mitigated by the existing `hasHighReasoningEffort` body-heuristic bump in the same file).

## Fix

`resolveDirectHeadersTimeoutMs()` now takes two additional optional parameters, `attempt` and `hasCallerDeadline`:
- **Attempt 0** (pooled dispatcher): unchanged — bound by the flat `OMNIROUTE_DIRECT_HEADERS_TIMEOUT_MS` floor (default 30s), preserving #10214's zombie-socket detection.
- **Attempt 1+** (fresh-socket retry): when the caller already attached its own deadline signal (in production, the resolved connection/model/provider/`FETCH_TIMEOUT_MS` cascade merged in by the executor layer before `proxyFetch` is invoked), the retry now defers to a generous backstop ceiling (`OMNIROUTE_DIRECT_RESPONSE_RETRY_TIMEOUT_MS`, default 600s) instead of the identical short flat window. That caller signal is the real bound and fires first in the intended path — the backstop only matters if it doesn't. When there is no caller-supplied deadline at all, the retry keeps the same flat floor as before (no regression for that case).

The new bound:
- Is fully configurable via the new `OMNIROUTE_DIRECT_RESPONSE_RETRY_TIMEOUT_MS` env var (documented in both `.env.example` and `docs/reference/ENVIRONMENT.md`, per the env/docs contract).
- Never aborts a stream that has already produced bytes — the watchdog only ever wraps the pre-headers `directFetchWithBoundedResponseStart` window; its timer is cleared as soon as the fetch promise resolves (before any body streaming begins), for both attempts, unchanged from the pre-existing #10214 behavior.

## Regression test (TDD, Hard Rule #18)

`tests/unit/proxyfetch-direct-response-start-flat-retry-budget-13703.test.ts`

**RED** (against unmodified `release/v3.8.51`, confirmed in this worktree before the fix):
```
pooled attempt should be cut at the flat 80ms budget, was cut after 85ms  ✓ (attempt 0 correct)
EXPECTED (fixed) behavior: fresh-socket retry should use the remaining caller
budget (~1920ms available), not the identical flat 80ms window. Got 81ms —
proves the retry reuses the same flat budget as the pooled attempt instead of
deriving from the timeout cascade / remaining target timeout.
```

**GREEN** after the fix — plus no regression on the existing suite for this file:
```
  tests/unit/proxyfetch-direct-response-start-flat-retry-budget-13703.test.ts \
  tests/unit/proxyfetch-direct-response-start-timeout-10214.test.ts \
  tests/unit/direct-response-start-timeout-reasoning-504.test.ts \
  tests/unit/proxy-fetch.test.ts \
  tests/unit/proxyfetch-retry-fresh-socket-4252.test.ts \
  tests/unit/proxyfetch-caller-abort-log-suppression.test.ts \
  tests/unit/proxyfetch-undici-retry.test.ts \
  tests/unit/proxyfetch-bun.test.ts \
  tests/unit/proxyfetch-upstream-status-capture.test.ts \
  tests/unit/proxyfetch-vercel-relay-2743.test.ts \
  tests/unit/proxy-fetch-dns-retry-10443.test.ts \
  tests/unit/proxy-concurrency-keepalive-regression.test.ts
# tests 55, pass 55 (1 pre-existing skip), fail 0
```

`#10214`'s own regression suite (no-caller-signal retry, zombie-socket detection, dispatcher identity) and `#12861`'s unhandled-rejection guard on the same function both stay green — verified by direct re-run of their test files above.

## Also in this PR

- `tlsProfileForProvider`'s return type in `proxyFetch.ts` is pulled into a named type alias (`TlsProfileResult`) so its signature fits on one line under Prettier's canonical formatting — otherwise Prettier's mandatory lint-staged reformat of this **frozen** file (`check:file-size`) grows it past the baseline on any touch. Purely a formatting/line-count fix, no behavior change (confirmed by `check:open-sse-typecheck` and the existing test suite).

## Gates run

- `node scripts/check/check-file-size.mjs` — clean on all touched files (only the pre-existing, untouched `tests/unit/chatcore-translation-paths.test.ts` violation remains, unrelated to this diff).
- `node scripts/check/check-complexity.mjs` — OK (2857 violations vs baseline 3218).
- `node scripts/check/check-cognitive-complexity.mjs` — OK (1291 violations vs baseline 1437).
- `npm run typecheck:core` — exit 0.
- `npm run check:open-sse-typecheck` — OK, 0 pre-existing errors.
- `node scripts/check/check-env-doc-sync.mjs` — the new env var is fully in sync (only the two pre-existing base-red entries remain, unrelated to this diff).

## Base status

⚠️ base-red inherited: #14004 — docs env/docs contract (fixed separately in #14022), `chatHelpers.ts` file-size drift. Neither touched by this PR.

Verified this commit and PR body carry no AI-generation footer or bot co-author trailer (project Hard Rule #16).


diegosouzapw:

Issue #13893, opened 2026-09-16
feat(providers): opt-in raw passthrough for Claude OAuth connections — no way to send an unmodified request today


diegosouzapw:
## Summary

There is no way to send a Claude request unmodified when the connection uses OAuth. The rewrite is not behind a setting: it triggers on the credential itself, so every operator on an `sk-ant-oat` token gets it, whatever the client.

This issue proposes an opt-in, default-off escape hatch per connection, and documents what it would have to skip.

## Current behavior

Gate: `open-sse/executors/base.ts:970-980` — any `accessToken` starting with `sk-ant-oat` with no `apiKey`. The comment on `:967-969` states the intent: the cloak applies regardless of the upstream client.

Applied unconditionally from there:

- Billing line and the `You are Claude Code, Anthropic's official CLI for Claude.` sentinel prepended to `system` (`base.ts:1192`), with the `cch=` value recomputed per request as a hash of the whole body (`open-sse/services/claudeCodeCCH.ts:40-47`).
- `cache_control` deleted from every tool (`base.ts:1003-1007`).
- Missing `cache_control.ttl` defaulted to `1h` (`open-sse/services/claudeCodeConstraints.ts:169-202`).
- Synthetic `metadata.user_id` identity, ignoring the client's (`base.ts:1143-1164`, `open-sse/services/claudeIdentity.ts:257-267`).
- Header rewrite: `user-agent`, `x-app`, `X-Stainless-*`, `anthropic-version`, `anthropic-beta`, with the client's own versions dropped before the merge (`base.ts:1242-1296`).
- Header and body field reordering to the CLI wire order (`base.ts:1403-1410`).

What an operator can turn off today covers only part of it: `alwaysPreserveClientCache: "always"` (API only, no dashboard control), the system-block transform pipeline in Settings → Routing, and `x-omniroute-thinking: off` / `x-omniroute-effort: off` per request (`base.ts:1013-1035`). None reaches the tool `cache_control` strip, the billing line, the sentinel, the identity or the fingerprint.

## Proposal

A per-connection `rawPassthrough` flag, default off, that skips the block at `base.ts:975-1325` when the client already speaks the Claude format and the target is Claude (the `isClaudePassthrough` path at `open-sse/handlers/chatCore.ts:2205`).

[...]

PR #14053, opened 2026-09-18
feat(oauth): per-connection Claude OAuth raw passthrough (#13893)

HouMinXi:
## Summary

Claude OAuth traffic that is not Claude Code currently still gets the CLI
cloak, fingerprint rewrite, and CCH body signing. Anthropic then 400s
`sk-ant-oat` tokens on extra-usage / cache-control for those clients
(#13893).

This adds a per-connection opt-in, `providerSpecificData.rawPassthrough`
(boolean, default false, no migration). When on, a Claude OAuth connection:

- skips CLI cloak, static fingerprint headers, and `signRequestBody`
- rewrites outgoing headers at the final merge (hop-by-hop + CLI leftovers)
- keeps the gateway's own auth headers if the client Connection list
  nominates them
- retries 400s without re-signing the body
- resolves through the native claude executor instead of CLIProxyAPI / Dario

The dashboard toggle is limited to Claude OAuth. Save writes the top-level
boolean and drops nested `passthrough.raw` / `passthrough.rawPassthrough`
aliases so an import cannot re-arm the hatch after the operator turns it off.

API-key Claude connections keep the existing cloak even if the flag is set.

Related to #13893. #13893 stays open.

## Test plan

- [x] `node --import tsx/esm --test tests/unit/claude-oauth-raw-passthrough.test.ts` (27/27)
- [x] adjacent cloak / thinking-guard / schema suites (93/93 combined)
- [x] i18n key completeness (`65` locales)
- [ ] maintainer: enable the toggle on one Claude OAuth connection, send a
      non-CLI Chat Completions request, confirm Anthropic 200 and that
      `x-anthropic-billing` / `cch=` are absent on the wire

⚠️ base-red inherited: #13866


HouMinXi:
CI on this branch matches the inherited set on pristine `release/v3.8.51` (`1603c86e06`): API Route Typecheck, Docs Gates, Fast Quality Gates, and the four unit shards. Merge integrity, Vitest, and ESLint are green.

Unique tests: 50/50 locally (`tests/unit/claude-oauth-raw-passthrough.test.ts` + `tests/unit/provider-specific-data-schema.test.ts`). Unique files lint clean.

Default stays off (`isConnectionRawPassthrough` returns false on missing/malformed input). An explicit top-level `false` beats leftover nested aliases. Gateway auth headers cannot be nominated off the Connection list. No product change in this turn.

HouMinXi:
Thinned this PR to the plan on issue 13893.

The hatch is now a small resolver next to `resolveConnectionCacheOverride`, a four-condition AND gate at the cloak block (`isClaudePassthrough`, genuine `claude` provider, OAuth token, flag), and the existing dashboard toggle. Nested `passthrough.raw` aliases no longer arm the hatch; save still writes a top-level boolean and strips them so an import cannot turn the switch back on.

Removed the standalone 177-line helper, the hop-by-hop / Connection header sanitizer, and the proxy-executor bypass. Those were extra surface. Fingerprint leftovers from `buildHeaders` (`x-app`, `x-stainless-*`, `anthropic-dangerous-direct-browser-access`) are still dropped on the raw path, which is the acceptance check that no Stainless header is injected.

On the two open questions: the flag hard-requires `isClaudePassthrough` (non-Claude-format clients keep the cloak), and the stored shape is a flat boolean.

Unique tests 20/20 after the cut. Head `9346e0c646`.

HouMinXi:
Holding this behind the chatCore re-extract (#13065). Unique diff includes `open-sse/handlers/chatCore.ts`. Merging it first would add another hunk for the extraction to place. No maintainer review yet; nothing to adopt.

HouMinXi:
Rebased onto current `release/v3.8.51` after #14078.

The five locale files that conflicted kept the reviewed translations and the three raw-passthrough keys. Unit tests 20/20.

HouMinXi:
Rebased onto current `release/v3.8.51` (`36493a6270`). Unique commits unchanged.

PR #14056, opened 2026-09-18
docs(cli): fix non-existent `config show` and `models list` commands

### LINKED ISSUE #13997: docs(cli): README and CLI-TOOLS document two commands that do not exist (`config show`, `models list`)
Two commands printed in our own documentation do not exist in the CLI. Found while triaging #13982, where an outside contributor had picked them up and repeated them in a proposed guide — so this is actively misleading people today.

## 1. `omniroute config show` — `docs/reference/CLI-TOOLS.md:721`

```
omniroute config show                  # Display current configuration
```

There is no `show` subcommand. `config` registers exactly: `list`, `get <tool>`, `set <tool>`, `validate <tool>`, `opencode`, `lang` (`bin/cli/commands/config.mjs:285-367`). The intended command is `omniroute config list`.

Worth noting while fixing: `config` manages **external CLI tool** configuration (Claude Code, Cline, OpenCode…), not OmniRoute's own settings — which are `omniroute env show | get <k> | set <k> <v>`. The one-line description in the docs ("Display current configuration") reads as if it were the latter, which is probably how the confusion started.

## 2. `omniroute models list` — `README.md:865`

```
omniroute models list                     # ← runs against the REMOTE server
```

The command is `models [provider]` (`bin/cli/commands/models.mjs:8`), so `list` is parsed as a **provider filter**. There is no provider named `list`, so this **prints an empty model list and exits 0** — it fails silently, which is worse than erroring. The correct forms are `omniroute models` (all), `omniroute models groq` (filter by provider), or `omniroute models --search <query>` (`models.mjs:10-11`).

## Fix

Two line edits plus `npm run check:docs-all`. A regression guard would be better than the edit alone: the doc-sync gates already parse docs, so asserting that every `omniroute <cmd>` appearing in `README.md` and `docs/reference/CLI-TOOLS.md` resolves against the Commander registry (`bin/cli/commands/registry.mjs`) would stop the whole class — this is the second time a wrong command has reached a user.

Good first issue: the two edits are trivial; the gate is optional and separable.

Mistertelecom:
Fixes incorrect CLI commands documented in the README and CLI-TOOLS reference that do not exist or behave incorrectly.

Closes #13997.

`config show` does not exist, and the config subcommand actually manages external CLI integrations, not OmniRoute's own envs. Changed to `config list`.
`models list` parses `list` as a provider filter, causing unexpected behaviors. The correct command is just `models`.


diegosouzapw:

PR #14010, opened 2026-09-17
fix(docker): copy the app with node ownership instead of a second chown layer

costajohnt:
## Summary

The `runner-base` stage copies the standalone build as root and then runs `RUN chown -R node:node /app`. On overlayfs a chown rewrites every file it touches into the new layer, so the published image carries the ~2 GB standalone tree twice. `docker history diegosouzapw/omniroute:latest` shows it directly:

```
2.04GB  RUN /bin/sh -c chown -R node:node /app
16.4kB  COPY /app/scripts/dev/healthcheck.mjs ./healthcheck.mjs
28.1MB  COPY /app/node_modules/better-sqlite3 ./node_modules/better-sqlite3
2.03GB  COPY /app/.build/next/standalone ./
```

This PR sets `--chown=node:node` on the three `COPY --from=builder` lines, removes the recursive chown, and hands `/app` and `/app/data` to `node` non-recursively next to the existing `mkdir -p /app/data` (`/app` itself is created root-owned by `WORKDIR /app` in the `base` stage, and the runtime user still needs to create entries under it and write `/app/data` when no volume is mounted). `runner-web` and `runner-cli` are untouched; they already switch back to root for their own COPYs after this point.

## Related Issues

- Closes #13990

## Validation

- [x] Change type: build-deploy
- [x] Focused tests and category gates from the golden path: `node --import tsx/esm --test tests/unit/dockerfile-*.test.ts tests/unit/docker-build-memory-budget.test.ts` (22 pass), `npm run check:build-scope`, `npm run check:changelog-integrity`
- [x] `npm run lint`
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR
- `hadolint --failure-threshold error Dockerfile` (the CI lint) exits 0 before and after with the same warning set.

### Image size

CI does not build the image on PRs, so this was measured locally by rebuilding the `runner-base` COPY/chown sequence against the real `/app` tree from `diegosouzapw/omniroute:latest` (root-owned source files, same `node:26-trixie-slim` base, linux/arm64). Same content and same instructions as the Dockerfile, without the 10+ minute Next.js build.

| | `docker history` layers | `docker image inspect --format '{{.Size}}'` |
|---|---|---|
| before (`COPY` + `RUN chown -R`) | `COPY` 2.04 GB + `chown -R` 2.04 GB, 4.38 GB total | 1,220,846,106 |
[...]

PR #14057, opened 2026-09-18
fix(docker): prevent duplicate layer storage by using COPY --chown

### LINKED ISSUE #13990: [BUG] Docker image stores the app twice (~2 GB) because of RUN chown -R node:node /app
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04.4 LTS, kernel 6.17.0-1016-oracle (aarch64), Docker Engine 29.5.2

### Node.js Version

_No response_

### Provider(s) Involved

_No response_

### Model(s) Involved

_No response_

### Client Tool

_No response_

### Description

The published image is 5.67 GB on disk, and about 2 GB of that is a duplicate of the app.

In the `runner-base` stage, the app is copied in as root and then re-owned in a separate layer:

```dockerfile
COPY --from=builder /app/.build/next/standalone ./                                  # line 251
COPY --from=builder /app/node_modules/better-sqlite3 ./node_modules/better-sqlite3  # line 256
COPY --from=builder /app/scripts/dev/healthcheck.mjs ./healthcheck.mjs              # line 262
RUN chown -R node:node /app                                                         # line 267
```

Image layers are copy-on-write: changing a file's owner writes a full new copy of that file into the next layer. So the `chown -R` layer contains the whole standalone bundle again, and the original copy stays in the layer below it.

### Steps to Reproduce

```sh
docker pull ghcr.io/diegosouzapw/omniroute:latest
docker image history --format '{{.Size}}\t{{.CreatedBy}}' ghcr.io/diegosouzapw/omniroute:latest | sort -rh | head -3
```

### Expected Behavior

Image should be around 3 GB

### Actual Behavior

Image is almost 5.6GB in size

### Test Impact

Unsure

### Error Logs / Output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Validation Plan

_No response_

Mistertelecom:
Fixes #13990.

The previous `RUN chown -R node:node /app` duplicated the entire app in the image because modifying ownership in a new layer copies all touched files. This replaces it with `COPY --chown=node:node` to set ownership at extraction, bringing the published image size down by ~2GB.

Closes #13990


Mistertelecom:
Closing this one: #14010 already landed the same approach (`COPY --chown=node:node` at copy time instead of a second chown layer), and the Dockerfile on `release/v3.8.51` carries it with the 'Do NOT add a RUN chown -R' note. My branch is redundant. Thanks!

Issue #14062, opened 2026-09-18
fix(backend): squash of #13717 (e7999c477b) collaterally reverted ~14 merged PRs at the release tip


HouMinXi:
## What happened

The squash merge of #13717 (`e7999c477b`, 250 files, +2177/-4344) restored overlapping files to the PR branch's stale base. Every PR that had landed on `release/v3.8.51` between that branch point and the merge lost the files it shared with #13717.

#13717 is mine. I found this after a production deploy from the current tip started failing, and I traced the missing code back through that merge. Sorry.

## How I verified

For each suspected victim I took the PR's merge commit, extracted its added lines, and checked two things:

1. Are those lines present at the current tip (`1603c86e06`)?
2. If not, does `e7999c477b`'s own diff contain the exact deletions?

Only files where both hold are listed below. Later legitimate edits (VACUUM replacement, i18n locale churn) are called out separately and are **not** in the table.

## Confirmed reverts (code deleted by e7999c477b, still missing at tip)

| PR | What was lost | Files hit |
|---|---|---|
| #13655 | Agnes intl thinking-effort tier declarations | `open-sse/config/providers/registry/agnes/index.ts` (all three `supportedThinkingEfforts` arrays), `tests/unit/agnes-provider.test.ts` (-90 lines) |
| #13636 | Process crash guard extension | `src/shared/utils/httpClientAbortGuard.mjs` (99 lines), `open-sse/utils/directResponseStartTimeout.ts`, `scripts/build/assembleStandalone.mjs`, `scripts/dev/standalone-server-ws.mjs` |
| #13701 | Release acceptance shadow pipeline (entire feature) | workflow, 7 scripts under `scripts/release-acceptance/`, combo schema, 8 test files, fixtures |
| #13463 | gemini-3.8 think-level support | `open-sse/services/thinkingBudget.ts` (37/37 added lines), 2 translators, `tests/unit/gemini-38-think-level.test.ts` (-148) |
| #13572 | Claude assistant prefill restore | `open-sse/executors/base.ts`, `open-sse/services/contextManager.ts`, test |
| #13560 | combo probe no-thinking fix | `open-sse/services/diagnostics.ts`, `open-sse/services/testHealth.ts`, 3 test files |
| #13445 | groq TLS fingerprint fix | `open-sse/utils/proxyFetch.ts` (17/17), docs, `tests/unit/tls-proxy-context.test.ts` (-68) |
[...]

PR #14063, opened 2026-09-18
fix(providers): restore Agnes thinking-effort tiers deleted by #13717

HouMinXi:
## Summary

Restores the Agnes intl thinking-effort tier declarations that landed in #13655 and were later deleted by the #13717 squash (`e7999c477b`). Suffix ids such as `agnes/agnes-3.0-flash-xhigh` again resolve to `agnes-3.0-flash` via `supportedThinkingEfforts` in `open-sse/config/providers/registry/agnes/index.ts`.

## Test plan

- `tests/unit/agnes-provider.test.ts`: 20/20.
- Defect injection: rename the three `supportedThinkingEfforts` arrays -> 4 fail; restore -> 20/20.

---


diegosouzapw:
Reviewed and verified. This restores exactly the code that #13717's squash collaterally deleted from your original #13655 — I diffed the restored hunks in `open-sse/config/providers/registry/agnes/index.ts` and `tests/unit/agnes-provider.test.ts` byte-for-byte against the state right after #13655 merged, and they match exactly. No scope creep, nothing else on the current tip is touched or reverted.

Ran `tests/unit/agnes-provider.test.ts` in an isolated worktree: 20/20 passing. I also broke the fix on purpose (nulled out `agnes-3.0-flash`'s `supportedThinkingEfforts`) to confirm the new tests actually fail without it — they do, so this isn't a tautological test. `typecheck:core` is clean.

The two red checks (`API Route Typecheck`, `Docs Gates`) are not caused by this PR — your branch was cut before #14022 and #14079 landed on the release tip fixing exactly those two gates (`OMNIROUTE_STRIP_SYSTEM_PREAMBLE` env-doc sync and the API-route TS baseline). I reproduced the identical 3 unrelated TS errors on your branch and confirmed the current tip is clean on both gates. A rebase onto the current release/v3.8.51 tip should turn both green — no code change needed on your end for that.

Merging as-is once rebased/re-run is green.

PR #14030, opened 2026-09-17
fix(resilience): restore #12861 settled-guard and #13636 crash-guard predicates

maxmad64bis:
> ⚠️ base-red inherited: #14004

## Summary

A recoverable direct-fetch timeout could kill the whole server process: when the response-start timer fired after the fetch had already settled, the abort reason escaped as an `unhandledRejection`, promoted to `uncaughtException`, process exit. This restores the settled-guard (the timer becomes a no-op once the attempt settled) plus the crash-guard predicates that recognize this code, combo hedge cancellations and upstream network failures. Related to #12942, which covers the same crash but additionally rewires the production wrapper — this PR deliberately does not rewire the wrapper, keeping the change to the race plus the guard predicates.

## Related Issues

- Closes #12861
- Related to #12942
- Related to #13636

## Validation

- [ ] Change type: routing
- [ ] Focused tests and category gates from the golden path
- [x] `npm run lint` — 0 errors on all touched files
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR
- [x] File-size gate: clean on PR files (`chatcore-translation-paths` violation is base-inherited, file untouched by this PR)
- [x] Api-typecheck: no new regression (`chat.ts` TS2345 is base-inherited, file untouched by this PR (`OMNIROUTE_STRIP_SYSTEM_PREAMBLE`, introduced by #12905 without `.env.example` entry — file untouched by this PR))

## Tests Added Or Updated

- `tests/unit/direct-response-start-timeout-settled-guard-12861.test.ts` (restored, 184 lines): race harness with deterministic timer neutering plus `unhandledRejection` listener — fails without the guard, passes with it.
- `tests/unit/http-client-abort-guard-direct-timeout-12861.test.ts` (restored, 144 lines): predicate coverage for the three restored classifiers.
- `tests/unit/httpClientAbortGuard.test.mjs` (+35 lines) and `tests/unit/httpClientAbortGuard-default-logger.test.mjs` (2 wording assertions): neighbors realigned with the restored guard.
- Result: 33 pass / 0 fail / 3 cancelled (`proxyfetch-10214` cancellations are a pre-existing sandbox note, identical before and after).

## Coverage Notes

[...]

PR #14064, opened 2026-09-18
fix(resilience): restore process crash guard deleted by #13717

HouMinXi:
## Summary

Restores the process crash-guard extension from #13636 that #13717 (`e7999c477b`) deleted: `httpClientAbortGuard.mjs`, the settled-guard on `directResponseStartTimeout`, and the two standalone-server wiring sites. Combo hedge cancels and recoverable upstream fetch failures are swallowed again instead of becoming unhandled rejections.

## Test plan

- `tests/unit/http-client-abort-guard-direct-timeout-12861.test.ts` + `tests/unit/direct-response-start-timeout-settled-guard-12861.test.ts`: 19/19.
- Defect injection: force `shouldSwallowUncaught` to always return false -> 3 fail; restore -> 13/13 on the abort-guard file.

---


diegosouzapw:
Reviewed #14064. This is a clean, complete restore of the original #13636 fix that e7999c477b collaterally deleted — verified the file set matches f36fcd8c312f exactly (all 10 files), ran the PR's own test files plus the pack-artifact-policy/closure suite in an isolated worktree (33/33 and 23/23 passing respectively), and confirmed typecheck:core and eslint are clean.

One thing worth flagging for coordination, not a defect in this PR: #14030 restores the same core predicates (open-sse/utils/directResponseStartTimeout.ts, src/shared/utils/httpClientAbortGuard.mjs) but does NOT restore scripts/build/assembleStandalone.mjs or scripts/dev/standalone-server-ws.mjs — so on its own it would leave the production entrypoint (dist/server-ws.mjs) without the crash guard installed, which is the actual point of the original fix. This PR (#14064) is the more complete restoration and is the one we intend to merge; #14030 becomes redundant once this lands. Separately, #14029 removes the httpClientAbortGuard.mjs pack-artifact allowlist/required-path entries on the assumption they're permanently orphaned — that assumption only holds if this PR (or #14030's missing half) never lands. We'll sequence #14064 before #14029 (or drop #14029) to avoid re-breaking the pack-artifact policy. No action needed from you here — merging as-is.

PR #13224, opened 2026-09-10
fix(codex): route discovered reasoning tiers through catalog and dispatch

TPOHH:
## Summary

Codex discovery currently drops object-shaped `supported_reasoning_levels` entries and the upstream default. Newly discovered models consequently lack effort variants in the dashboard and `/v1/models`, while native `max` can silently become `xhigh` during dispatch.

This preserves native effort metadata through discovery, persistence, model resolution, and the executor. Discovered models receive deduplicated catalog/dashboard variants without adding their names to the static registry. Exact upstream IDs, including names ending in `-high` or `-max`, take precedence over synthetic aliases. The Chat route carries the resolved metadata and original explicit choice through translation so `max` reaches Codex unchanged.

Ultra is preserved in raw discovery metadata but excluded from newly generated API aliases and advertised wire tiers: the live Codex Responses endpoint rejects `ultra` and accepts `max`. Existing built-in Ultra aliases retain their historical Max mapping. Connection defaults and explicit thinking policies retain precedence over discovered defaults.

## Related Issues

- Related partial work: #12933 (safe discovery/static Astra), #12759 (static Astra support), and #12686 (reasoning-rule vocabulary). This addresses the generic discovery-to-dispatch path and object-shaped metadata rather than adding another model-name rule.
- ⚠️ base-red inherited: #12732

## Validation

- [x] Change type: provider / routing
- [x] Reconciled with active `release/v3.8.51` at `81bf3cc36e614574c65f5746a571342b8f1fbbcf`; its Kimi K3 capability carve-out is preserved.
- [x] Focused Node suite on the PR base: **127 tests passed**, including the new regression file, Codex executor, discovery split, synced efforts/capabilities, and `kimi-k3-effort-tiers-12299.test.ts`.
- [x] `npm run typecheck:core`, `check:provider-consistency`, `check:provider-assets`, and `check:file-size` passed.
- [x] Changed-file Prettier/ESLint and all pre-commit gates passed.
- [ ] Full `npm run lint -- --quiet`: three existing `no-explicit-any` errors in unchanged `tests/unit/volcengine-plan-binding-upsert.test.ts` (70, 96, 120), plus stale suppressions. No baseline changes in this PR.
- [x] Production-code changes include automated regressions.

[...]

PR #14065, opened 2026-09-18
fix(codex): restore reasoning-object whitelist deleted by #13717

HouMinXi:
## Summary

Restores the Codex reasoning-object key whitelist from #13643, which #13717 (`e7999c477b`) deleted. Unknown OpenRouter-style keys (`enabled`, `max_tokens`, ...) are stripped before the Responses API wire.

Conflict with current tip: keep the post-#13643 force-rule (`getForcedReasoningEffort` still wins) and keep `enabled: false` as a connection-default override. File-size freeze for `open-sse/executors/codex.ts` moved 1530 -> 1552 with a restore note.

## Test plan

- `tests/unit/codex-reasoning-wire-whitelist.test.ts`: 8/8.
- Defect injection: hard-code `clientDisabledReasoning = false` -> 2 fail; restore -> 8/8.

---


diegosouzapw:
Reviewed. The restore itself is correct and clean — I diffed your head against the current tip and the only change to `codex.ts` is the isolated whitelist + `clientDisabledReasoning` re-addition; every legitimate change that landed on `codex.ts` after the original #13643 (the `getForcedReasoningEffort` precedence from #13556, `stripCodexPassthroughRejectedParams`, the WS onclose hardening) is intact. All 8 new tests pass, and I re-ran 70 pre-existing Codex executor tests plus `typecheck:core` — all clean.

One concrete issue that will fail CI's file-size gate: the `config/quality/file-size-baseline.json` update sets `open-sse/executors/codex.ts` to 1552, but the gate script counts lines via `content.split("\n").length` (which is `wc -l` + 1 for a file with a trailing newline), so it actually measures 1553. Please bump both entries (and the `1530->1552` note) to 1553 — that's the only outstanding item; I'll re-check once it's pushed.

Separately, flagging for coordination rather than as something to fix here: #13224 rewrites this same block into a new `applyCodexReasoningSelection()` helper that spreads the raw client `reasoning` object without a key whitelist, so if it merges after this without picking up the fix, the OpenRouter-key 400 comes back a third time. I'll note this on that PR's review too.

PR #14102, opened 2026-09-18
fix(providers): allow full URLs for compatible chatPath and modelsPath

artickc:
## Why

Some OpenAI/Anthropic-compatible gateways do not keep model listing and chat on the same host. Operators already have Chat Path and Models Path fields for this, but those fields currently only accept relative suffixes of Base URL.

That fails in three ways:

1. Zod requires `chatPath` / `modelsPath` to start with `/`, so a full `http(s)` URL is rejected.
2. Executors, validation, and discovery concatenate `baseUrl + path`, so an absolute models URL becomes a garbage probe such as `https://chat-host/https://models-host/...`.
3. OpenAI-compatible model discovery often ignores `modelsPath` and probes `/v1/models` on the chat host instead.

The result is an empty models catalog even when the listing endpoint is valid.

## What changed

- Join `chatPath` / `modelsPath` as complete URLs when they start with `http://` or `https://`. Relative paths such as `/v1/models` keep the previous behavior.
- Accept full URLs in the provider-node Zod schema.
- Use the configured models URL for compatible listing and validation instead of always probing the chat host.
- Accept `model_name` records from non-OpenAI catalogs so listing still syncs when the payload has no `id`.

## Tests

```bash
node --import tsx/esm --test tests/unit/join-endpoint-url.test.ts tests/unit/custom-endpoint-paths.test.ts tests/unit/executor-default-base.test.ts tests/unit/openrouter-context-length-3202.test.ts tests/unit/custom-headers-provider-nodes.test.ts tests/unit/provider-page-helpers-3501.test.ts
```

135 passing.


artickc:
Closing — opened without operator confirmation, and the listing path is still failing for the reported compatible provider.

PR #14109, opened 2026-09-18
fix(translator): guard against empty tool_calls arrays closing message items early (#14008)

### LINKED ISSUE #14008: [BUG] Responses translator closes the message item early when a provider sends `tool_calls: []`, breaking streaming clients with `text part ... not found`
**OmniRoute Version:** 3.8.50
**Operating System:** Linux
**OS Version:** Ubuntu 24.04.4 LTS (kernel 6.8.0-139, x86_64)
**Node.js Version:** v26.7.0
**Provider(s) Involved:** CodeBuddy CN (`codebuddy-cn`, upstream `copilot.tencent.com`)
**Model(s) Involved:** `codebuddy-cn/deepseek-v4-flash`, `codebuddy-cn/glm-5.3-flash` (any model on this provider)
**Client Tool:** ZCode (OpenAI **Responses** API client, `/v1/responses`)

## Description

A provider that sends `tool_calls: []` (an **empty array**) on every delta makes the Responses translator close the assistant message item **immediately after the first content chunk**, then keep streaming more `response.output_text.delta` events for that same, already-closed `item_id`.

Strict Responses clients (ZCode's AI SDK adapter) key per-item state by `item_id` and delete the entry on `response.output_item.done` (= text-end). The late deltas then fail with:

```
UnknownError: text part msg_resp_cmb-<hex>_0 not found
```

which surfaces to the user as `AiSdkModelAdapterError: Model request failed.` (`reason=unknown`, `errorPhase=stream`, non-retryable) and kills the turn.

**Root cause** — truthiness check on an array (empty array is truthy in JS).

`open-sse/translator/response/openai-responses.ts:290` (current `main`):

```ts
  // Handle tool_calls
  if (delta.tool_calls) {                 // <-- `[]` is truthy -> enters branch
    if (state.reasoningId && !state.reasoningDone) {
      closeReasoning(state, emit);
    }
    const msgIdx = state.reasoningId ? state.reasoningIndex + 1 : idx;
    closeMessage(state, emit, msgIdx);    // <-- closes the message item early
    for (const tc of delta.tool_calls) {
      emitToolCall(state, emit, tc);
    }
  }
```

`closeMessage()` emits `response.output_text.done` + `response.content_part.done` + `response.output_item.done` for `msg_<responseId>_0`. The streaming loop keeps calling `emitTextContent()` for later content chunks, which re-emits `response.output_text.delta` with the same `item_id` because `state.msgItemAdded[idx]` is still set — producing deltas after the item was finalized.

Note `open-sse/utils/stream.ts:1919` already does this correctly (`delta?.tool_calls && delta.tool_calls.length > 0`), so the intent is clear and the other sites are just inconsistent.

## Steps to Reproduce

1. Configure a provider whose chat-completions SSE deltas include `"tool_calls": []` on every chunk. CodeBuddy CN does this; most providers send `null` or omit the field, which is why this is provider-specific.

   ```
   data: {"choices":[{"index":0,"delta":{"content":"你","tool_calls":[]},"finish_reason":null}]}
   data: {"choices":[{"index":0,"delta":{"content":"好","tool_calls":[]},"finish_reason":null}]}
   ...
   ```

2. Send a streaming Responses request through OmniRoute:

   ```bash
   curl -N https://<omniroute>/v1/responses      -H "Authorization: Bearer $KEY" -H "Content-Type: application/json"      -d '{"model":"codebuddy-cn/deepseek-v4-flash","stream":true,
          "input":[{"role":"user","content":[{"type":"input_text","text":"Say hello in one sentence."}]}]}'
   ```

3. Observe the event order:

   ```
   response.created
   response.in_progress
   response.output_item.added      item.type=message id=msg_resp_xxx_0   <-- opened
   response.content_part.added
   response.output_text.delta      delta="Hello"
   response.output_text.done       text="Hello"                          <-- closed after ONE chunk
   response.content_part.done
   response.output_item.done       item.id=msg_resp_xxx_0                 <-- item finalized
   response.output_text.delta      delta=" world"   <-- deltas continue AFTER the item is done
   response.output_text.delta      delta=", nice"
   ...
   response.completed
   ```

   The full text only ever appears in `response.output_text.done` as the truncated prefix; the remainder arrives as orphaned deltas.

4. On a strict client (ZCode / any AI SDK Responses consumer), the turn fails with `text part msg_resp_xxx_0 not found`.

**Reproducibility:** 4/4 attempts on this provider, including a no-tools control request, so it is not related to actual tool calls.

Measured (`response.output_text.delta` events arriving after the message item's `output_item.done`):

| Request | Unpatched 3.8.50 | Patched |
|---|---|---|
| `codebuddy-cn/deepseek-v4-flash` + tools | 13 violations | **0** |
| `codebuddy-cn/deepseek-v4-flash`, no tools | 58 violations | **0** |
| `codebuddy-cn/glm-5.3-flash` + tools | 30 violations | **0** |
| `th/deepseek-flash` (comparison, sends `null`) | 0 violations | 0 |

## Expected Behavior

`tool_calls: []` should be treated as "no tool calls": the message item must stay open until the real end of the assistant turn, and `output_text.done` / `output_item.done` must be emitted only after the last delta.

## Suggested Fix

Guard the array by length, consistent with `open-sse/utils/stream.ts:1919`:

```ts
if (Array.isArray(delta.tool_calls) && delta.tool_calls.length > 0) {
```

The same truthiness pattern exists in 4 other places on `main` (all currently unguarded):

| File | Line |
|---|---|
| `open-sse/translator/response/openai-responses.ts` | 290 |
| `open-sse/transformer/responsesTransformer.ts` | 798 |
| `open-sse/handlers/sseParser.ts` | 225 |
| `open-sse/translator/response/openai-to-antigravity.ts` | 56 |
| `open-sse/translator/response/openai-to-claude.ts` | 335 (`delta?.tool_calls`) |

These may have different blast radii (the `openai-responses.ts` one is the one proven to break clients), but they look like the same latent bug.

## Workaround (deployed, verified)

Replaced the guard in the 6 emitted Next.js chunk copies of `openai-responses.ts` inside the image and rebuilt it. Production regression after the patch: message item stays open across all deltas, `output_text.done` carries the complete text, 0 violations on all three test cases above, with tool calls and without.

I have not opened a PR because the fix touches a frozen/ratcheted file set (`config/quality/file-size-baseline.json`) and I am unsure whether you want the guard applied across all 5 sites or only the Responses translator. Happy to send a PR if you confirm the scope.

Mistertelecom:
Fixes a bug involving the OpenAI Responses API translator where empty `tool_calls: []` arrays (sent by some providers like CodeBuddy CN) were evaluated as truthy. This would trigger an early `closeMessage()` call and cause subsequent content deltas to reference an already-closed item, breaking strict AI SDK clients with "text part not found" errors.

The fix safely checks for `delta.tool_calls.length > 0` instead.

Closes #14008


Mistertelecom:
Closing this one: the guard is already on `release/v3.8.51` now — the translator reads `delta.tool_calls?.length`, so the empty-array case no longer closes the message item early. My branch is redundant. Thanks!

Issue #13973, opened 2026-09-17
fix(backend): three 6h schedulers fire in the same second; TRUNCATE checkpoint SIGBUSes under live mmap


HouMinXi:
## Summary

On a long-running production instance the server process dies with SIGBUS almost exactly every 6 hours after startup. There is no error output at all: no exception, no FATAL line, no shutdown log. The only trace is a systemd-coredump entry. I pulled the core and the faulting address maps exactly into a memory-mapped window of `storage.sqlite-shm` (the SQLite WAL-index file). The crash lands 1-2 seconds after the scheduled `wal_checkpoint(TRUNCATE)` fires.

## Timeline (two crashes, EDT)

Container started 13:42 on 2026-09-16.

- 19:42:05 `[DB] Periodic SQLite WAL checkpoint completed`
- 19:42:06 `[Cleanup] Starting auto-cleanup...` and `[ModelSync] Starting model sync cycle — 112 connection(s)` in the same second
- 19:42:06 systemd-coredump: process terminated with signal 7/BUS
- 19:42:14 container restart

Same sequence the next morning: checkpoint 04:58:16, cleanup + model sync 04:58:16, SIGBUS 04:58:17.

## Core dump evidence

The coredumpctl stack shows libc memcpy called from better-sqlite3 native code, and the thread stack bottoms out at `main` / `__libc_start_call_main`, so the fault is on the event loop thread.

gdb on the core:

```
si_addr = 0x7f21ec348000
...
0x00007f21ec348000 0x00007f21ec350000 0x8000  0x8000  /app/data/storage.sqlite-shm
```

The faulting address is the first byte of a 32KB mapping of the WAL-index shm file, at file offset 0x8000. SIGBUS on an mmap'd file means the page is no longer backed: the file was truncated or re-initialized smaller after the mapping was made.

## Mechanism

Three timers are anchored to process start with the same 6h period, so they fire in the same second:

- `src/lib/db/walMaintenance.ts:39`: `DEFAULT_WAL_TRUNCATE_INTERVAL_MS = 6h` -> `wal_checkpoint(TRUNCATE)`
- `src/lib/db/cleanup.ts:985`: `CLEANUP_INTERVAL_MS = 6h` -> batched DELETEs on large tables (200k+ rows)
- `src/shared/services/modelSyncScheduler.ts:17`: 6h model sync over all connections

[...]

PR #14113, opened 2026-09-18
fix(startup): bound model-catalog sync fan-out at boot

### LINKED ISSUE #13975: fix(startup): initial model sync fans out to every connection at once, OOM-ing the V8 heap
## Summary

With a large number of provider connections (~112, autoSync enabled), the process can enter a V8 out-of-memory crash loop during the first minute after boot. We watched two consecutive containers die with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory` at ~36 and ~41 seconds into their lives. The third boot survived; our best explanation is that the startup cleanup commits in batches, so each crashed boot left less cleanup work for the next one.

## Evidence (journal, EDT)

Second container of the crash loop:

```
06:32:41 [ModelSync] Scheduler started — interval: 6h
06:32:41 [Cleanup] Background cleanup scheduler started (every 6 hours).
06:32:47 [ModelSync] Starting model sync cycle — 112 connection(s)
06:33:16 FATAL ERROR: Ineffective mark-compacts near heap limit ...
```

The heap died at 3061MB against a `--max-old-space-size=3072` cap. The previous container had died the same way about a minute earlier.

## Why the first minute carries the most risk

- `src/shared/services/modelSyncScheduler.ts:293` fires the first sync cycle at +5s. `runSyncCycle` (`:248-252`) launches `syncConnectionModels` for every connection in one unbounded `Promise.allSettled(connections.map(...))`. Each sync fetches and parses a full model-list JSON. With 112 connections that is 112 fetch/parse cycles in flight at once on a cold heap.
- `src/lib/db/cleanup.ts:958-959` fires the first auto-cleanup at +30s: batched DELETEs on large tables (one of ours holds 234k rows with no index on the filter column), right on top of the sync.
- Both land while the server is still warming up and all caches are cold.

So the process does some of its heaviest work at +5s to +60s with the least warm state, and the heap cap is the only guard. The same instance survived for weeks with a 4096MB cap; 3072MB is not enough at this connection count.

## Related observation

A sibling container hit the same FATAL after 1h33m of uptime during a WAN outage storm, with dozens of egress retry errors in the preceding minutes. So the cap is reachable outside startup too; startup is just the guaranteed spike. Worth knowing for anyone lowering the heap cap to stay under a cgroup limit: it can trade a kernel OOM for a V8 crash loop.

## Suggested fixes

1. Bound the concurrency in `runSyncCycle` (process connections in small batches) instead of one unbounded map.
2. Delay the initial model sync or make it on-demand, so boot serves traffic first; consider a longer initial delay for the first cleanup too.
3. Document expected memory vs connection count, or leave more margin by default: the V8 heap cap does not cover native buffers, so the container limit needs real headroom above it.

## Environment

- OmniRoute built from release/v3.8.51; code refs verified at tip 0d089e7e
- Node 26, 112 provider connections with autoSync enabled
- `NODE_OPTIONS=--max-old-space-size=3072`, container cgroup memory limit 5G
- rootless podman 5.8.4, 1.1GB sqlite DB in WAL mode

## Status

We keep observing this in production with crash-time state capture in place. I intend to open a PR myself (bounded concurrency in `runSyncCycle`, deferred initial sync/cleanup). If a maintainer prefers a different shape, tell me here and I will build it that way.

HouMinXi:
## Summary

`runSyncCycle` used to fire `syncConnectionModels` for every autoSync connection in one unbounded `Promise.allSettled`. On a host with 112 connections that meant 112 catalog fetches/parses at once, on a cold heap, 5s after boot. Two consecutive containers died with a V8 heap OOM around +36s / +41s.

Diego's contract on #13975:

1. Cycle-wide concurrency cap of 4. Heap cost is total catalog JSON in flight, not pressure on one upstream, so the cap is on the cycle, not per provider.
2. First cycle at +90s instead of +5s. Boot can serve traffic, and the first cleanup (+30s) finishes before the sync starts. Delay alone would only move the explosion; with the cap it matches what the old "avoid startup congestion" comment was trying to do.

Reactive / on-demand `syncConnectionModels` is unchanged. Memory-per-connection docs stay a follow-up, as agreed.

Leftover from #13973 (stagger the three 6h timers; `last_seen_at` index) is not in this PR.

Closes #13975.

## Test plan

- 12 fake autoSync connections, each fetch held open until a gate drops. Peak in-flight is 4 with the cap, 12 if you restore the unbounded `map`. Same test: fail (12 !== 4) -> restore -> pass (peak 4).
- Existing scheduler tests still pass (15/15). Startup delay now asserts `MODEL_SYNC_STARTUP_DELAY_MS` (90_000) instead of a hard-coded 5000.
- Cleanup first pass stays at 30s; not moved.

⚠️ base-red inherited: #13866


diegosouzapw (approved):
Segue o contrato da #13975 exatamente: cap **do ciclo** (não por provider), adiamento como complemento e não substituto, e o unitário com prova red→green (12 → 4 em voo).

Revisei `mapWithConcurrencySettled`: índice compartilhado incrementado entre `await`s é atômico no event loop, resultados voltam na posição original, `Promise.all(workers)` só resolve depois do último — o formato de `PromiseSettledResult[]` que o resto do ciclo consome fica idêntico. `Math.min(Math.max(1, n), values.length)` cobre lista vazia e cap absurdo. Nada a objetar.

Os vermelhos de CI são a quinta onda de base-red (#13866) — a branch nasceu antes da #14101; atualizei a base para o CI re-rodar. O que sobrar é herdado e não bloqueia.

HouMinXi:
A quick note documenting the operational rationale behind the timing and concurrency numbers:

**Startup delay shifted from 5s to 90s**
The first 30 seconds after boot handle database bootstrap, routing initialization, and the initial session cleanup pass (+30s). Kicking off a full catalog sync at 5s competed directly with server warmup. Moving the first cycle to 90s lets the container settle and begin serving traffic before background sync starts.

**Global cycle concurrency capped at 4**
The OOM on #13975 was driven by cold-heap memory pressure: 112 connections firing simultaneously meant 112 upstream catalog responses being fetched and parsed concurrently in V8 heap memory. Capping active syncs to 4 processes the queue in small batches without memory spikes.

**Operational behavior**
In the first 90 seconds after boot, the server serves existing cached/persisted models. Dynamic models for autoSync connections populate once the 90s cycle runs. Manual sync (saving a connection or triggering sync in UI) is reactive and remains immediate; it does not wait for the 90s timer.

PR #12, opened 2026-02-13
fix(ui): fix Select dropdown dark theme inconsistency

diegosouzapw:
Use bg-surface instead of bg-white for select and option elements. Prevents white dropdown in dark mode.

PR #14115, opened 2026-09-18
fix(api): sanitize /v1/batches create error body (Hard Rule #12)

### LINKED ISSUE #14089: fix(api): /v1/batches returns raw error.message in the response body (Hard Rule #12)
`src/app/api/v1/batches/route.ts:66` puts the raw `error.message` into the HTTP response body, contrary to Hard Rule #12 (`AGENTS.md`) — every HTTP / SSE / executor error response must go through `buildErrorBody()` or `sanitizeErrorMessage()` from `open-sse/utils/error.ts`. Found while triaging #14035, whose "errors pass through shared sanitization" bullet turned out to be true for every other `/v1/*` route checked and false here.

```ts
// src/app/api/v1/batches/route.ts:62-68
} catch (error) {
  console.error("[BATCHES] Create failed:", error);
  return NextResponse.json(
    {
      error: {
        message: error instanceof Error ? error.message : "Create failed",   // ← raw
        type: "invalid_request_error",
```

An `Error` thrown from the DB layer or a validation helper can carry file paths, SQL fragments or internal identifiers; this sends them to the client verbatim.

## Fix

One line: `message: sanitizeErrorMessage(error instanceof Error ? error.message : "Create failed")` (or replace the whole block with `buildErrorBody(...)`). Check the other handlers in the same file (`GET`, cancel, delete-completed) for the same pattern while there.

Per the "Adding a New API Route" checklist, add an assertion that the error response does not leak a stack/path (`!body.error.message.includes("at /")`). Reference: `docs/security/ERROR_SANITIZATION.md`.

Good first issue: single-file, existing helper, existing test pattern.

xiechimon:
## What

`POST /api/v1/batches` now routes its catch-block error message through `sanitizeErrorMessage()` before it reaches the client, instead of forwarding the raw `error.message`.

In my own words: a DB-layer failure during batch create could ship file paths, SQL fragments or internal identifiers to the caller verbatim; this puts the response through the same shared sanitizer every other `/v1` route already uses, and the 400 shape stays exactly as before.

## Why

Closes #14089. Hard Rule #12 (AGENTS.md): every HTTP / SSE / executor error response must go through `buildErrorBody()` or `sanitizeErrorMessage()` from `open-sse/utils/error.ts`. This route's catch block was the one `/v1/*` spot that skipped it.

The one-line core fix:

```ts
message:
  sanitizeErrorMessage(error instanceof Error ? error.message : "Create failed") ||
  "Create failed",
```

The `|| "Create failed"` keeps the message non-empty when sanitization fully redacts (the sanitizer returns `""`/`[REDACTED]` shapes depending on input), matching the `agents/credentials/route.ts` pattern. I left the Zod `validation.error.message` path (line 30) alone on purpose: that is a schema-generated message with no runtime data, and every sibling route (`audio/speech`, `classify`, `embeddings`, `images/generations`) passes it through unchanged.

Also checked the other handlers in the file per the issue: `GET` has no catch block and `parseBatchListLimit`'s message is a static string, so the POST catch was the only leak path.

## Testing

New `tests/unit/14089-batches-route-error-sanitization.test.ts` drives the real `POST` handler with a real API key and a real seeded input file. The harness does not support `mock.module()` and `mock.method` cannot redefine the sealed ESM namespace export, so the batch INSERT is made to throw through the repo's own `setDbInstance` DI seam: a thin wrapper delegates every read (auth lookup, file ownership, validation) to the real adapter and only the `INSERT INTO batches` statement throws a path-and-stack-laden `SQLITE_CANTOPEN`.

- Leak test: the 400 body contains neither the DB path, nor the `storage.sqlite` filename, nor a stack frame (`\n    at `); `type` stays `invalid_request_error`.
- Control test: a clean, non-sensitive message (`Batch endpoint is not enabled on this instance`) survives sanitization intact, so the fix does not turn every error into `[REDACTED]`.

Red on base: the leak test fails with `response leaked the DB path: SQLITE_CANTOPEN: unable to open database file /tmp/.../storage.sqlite`; after the fix, 2 pass.

Adjacent regression: all six existing batch test files green (88 pass / 0 fail, incl. the GHSA-2jm2/GHSA-m3hp ownership suite that exercises the same route). `npx eslint` and `npx prettier --check` clean on both touched files; `npm run typecheck:core` exits 0.


diegosouzapw:

PR #14120, opened 2026-09-18
fix(api): drop stale catalog content-length when re-serializing model responses (#14092)

### LINKED ISSUE #14092: fix(api): /api/v1/providers/{provider}/models forwards the unified catalog's content-length onto a re-serialized filtered body — response never completes (playground model dropdown stuck loading)
## Environment
- Image: `diegosouzapw/omniroute:next` (3.8.51-dev, 2026-09-17 build); the same code is on current `main` (`src/app/api/v1/providers/[provider]/models/route.ts`, last touched by #13831).
- Any provider (verified with `opencode` and `nvidia`).

## Symptom
On the provider detail page (`/dashboard/providers/{id}`), the Playground (游乐场) model `<select>` stays on "加载中..." forever. The dashboard hook fetches `` `/api/v1/providers/${provider}/models` `` with a plain `fetch` (no timeout), so a request whose body never completes pins the dropdown in the loading state indefinitely.

## Repro
```
$ curl -N -H 'Authorization: Bearer <key>' http://127.0.0.1:20128/api/v1/providers/opencode/models
<headers arrive in ~2s>
HTTP/1.1 200 OK
content-type: application/json
content-length: 833150     <-- full unified-catalog size
<body never arrives; connection is killed ~66s later>
```
`/v1/models` and `/api/v1/models` respond normally (833 KB body), so the catalog itself is healthy.

## Root cause
`route.ts` re-serializes a **filtered** subset of the catalog but forwards the upstream response headers verbatim:

```ts
const response = await getUnifiedModelsResponse(request);   // cached unified catalog
...
return Response.json(
  { object: payload.object || "list", data: [...deduped.values()] },  // much smaller body
  { status: response.status, headers: response.headers }              // <-- keeps content-length: 833150
);
```
The declared `content-length` (full catalog, ~833 KB) never matches the actual body (a few KB), so the response body never completes.

## Suggested fix
Strip length-describing headers from the forwarded set and let the runtime recompute them, e.g.:

```ts
const headers = new Headers(response.headers);
headers.delete("content-length");
headers.delete("transfer-encoding");
return Response.json({ object: payload.object || "list", data: [...deduped.values()] },
  { status: response.status, headers });
```

Note the dashboard's empty-result auto-retry path (`sync-models` + refetch in the providers hook) is affected the same way.

Workaround we currently run (startup patch on the standalone bundle, no-op once this is fixed upstream): rewrite `{status:X.status,headers:X.headers}` in the minified route to filter out `content-length`/`transfer-encoding`.

Mistertelecom:
Fixes #14092 — the provider-models route re-serializes a filtered catalog subset but forwarded the catalog's own `content-length` (full catalog) onto the smaller body, so the response never completes for the client and the dashboard Playground model picker hangs on "loading" forever.

**Fix** — drop the length headers after re-serialization, same idiom already used in `textCompletionTransform.ts` / `specialtyCatalog.ts`:

- `src/app/api/v1/providers/[provider]/models/route.ts` — the reported route: build the forwarded header set without `content-length` / `transfer-encoding`.
- `src/app/api/v1/vscode/[token]/models/route.ts::getVscodeModelsCatalogResponse` — same bug class on the vscode catalog surface: the helper builds its header set straight from the catalog response, and both consumers (`vscode/[token]/models`, `vscode/raw/[token]/models`) re-serialize `body.data` (filtered/expanded) before responding.

**Evidence — regression test `tests/unit/provider-models-content-length-14092.test.ts`** (real handlers, temp `DATA_DIR`, loopback request — run with `node --import tsx/esm --test tests/unit/provider-models-content-length-14092.test.ts`):

On `release/v3.8.51` (pre-fix, files reverted to the base commit):

```
not ok 1 - provider models route re-serializes without the catalog's stale content-length
    providers route: content-length must describe the re-serialized body
  expected: 3502
  actual: 179799
not ok 2 - vscode catalog helper drops stale length headers and keeps the cache headers
  actual: '179799'
# pass 0
# fail 2
```

With the fix applied:

```
ok 1 - provider models route re-serializes without the catalog's stale content-length
ok 2 - vscode catalog helper drops stale length headers and keeps the cache headers
# pass 2
# fail 0
```

(179799 = the full catalog's byte length in a fresh instance; the re-serialized body is 3502 bytes — the declared length could never match.)

Closes #14092.


diegosouzapw:
Great catch, and thanks for going further than the reported route — the vscode catalog route really does have the same bug class. Before merging, could you swap your manual header-delete calls for the existing `stripStaleEncodingHeaders()` / `filterUpstreamResponseHeaderEntries()` helpers in open-sse/utils/upstreamResponseHeaders.ts? They already strip `content-encoding` too, which your current diff doesn't, and it'll keep both routes consistent with how the rest of the codebase does this. There's a narrower duplicate PR (#14146) on just the provider-models route that we'll close in favor of this one once that's in.

Mistertelecom:
@diegosouzapw Thanks for the review!

The requested helper reuse is already present in the current PR head, `c2b7215ed46950436473c02fef2ec9c13f975d22`:

- `src/app/api/v1/providers/[provider]/models/route.ts` uses `stripStaleEncodingHeaders(response.headers)`.
- `src/app/api/v1/vscode/[token]/models/route.ts` uses `filterUpstreamResponseHeaderEntries(response.headers.entries())` inside `getVscodeModelsCatalogResponse()`.

Both routes now reuse the shared helpers from `open-sse/utils/upstreamResponseHeaders.ts`, removing `content-length`, `transfer-encoding`, and `content-encoding` while preserving unrelated headers and the existing VS Code cache headers.

The regression file `tests/unit/provider-models-content-length-14092.test.ts` also includes a test exercising both helpers with stale encoding/length headers and checking that unrelated headers are preserved.

Could you please take another look at the latest diff? If the fork workflows are still awaiting approval, could you also approve them so CI can run?

Thanks!

PR #14122, opened 2026-09-18
fix(providers): keep array subscripts in perplexity-web code output

### LINKED ISSUE #14121: fix(providers): Perplexity Web strips array subscripts from code: print(arr[0]) becomes print(arr)
### OmniRoute Version

3.8.51 (`release/v3.8.51`, verified against a 3.8.50 install with `8fbc85e` applied)

### Installation Method

npm (global)

### Operating System

Windows

### OS Version

Windows 11 Pro 10.0.26200

### Node.js Version

v24.14.1

### Provider(s) Involved

perplexity-web

### Model(s) Involved

pplx-sonnet, pplx-gpt-5.6-sol, pplx-kimi, pplx-grok-4.6

### Client Tool

direct requests to `/v1/chat/completions`

---

## Description

`CITATION_RE` in `open-sse/executors/perplexity-web/protocol.ts` strips any `[n]` token, and `cleanResponse()` runs over the whole answer before tool mode parses `<tool>` text into `tool_calls`. Subscript indexing is spelled exactly like a citation marker, so array access is removed from code — in rendered code blocks and in tool-call arguments alike.

Flagged as out of scope in the #14009 reviewer notes ("`CITATION_RE` still strips any `[n]` token, so array indexing like `arr[0]` in code output is still mangled. Worth a follow-up issue"), so this is that follow-up.

## Steps to Reproduce

Ask any `pplx-*` model to emit a line containing subscripts, for example:

> Output exactly this line in a code block, nothing else: `print(arr[0], arr[12])`

## Expected Behavior

`print(arr[0], arr[12])`

## Actual Behavior

`print(arr, arr)` — reproduced on `pplx-sonnet`, `pplx-gpt-5.6-sol`, `pplx-kimi` and `pplx-grok-4.6`.

List literals are affected the same way (`x = [0]` becomes `x =`), and because the cleaner runs before tool parsing, the same loss lands inside `write_file` arguments, so generated code is corrupted on disk rather than only in the rendered answer.

## Notes

PR follows: keep `CITATION_RE` as-is for prose, but skip it inside fenced code blocks, inline code spans and `<tool>` payloads.

Zicocoder:
## Summary

- `CITATION_RE` strips any `[n]` token, and `cleanResponse()` runs over the whole answer before tool mode parses `<tool>` text into `tool_calls`. Subscript indexing is spelled exactly like a citation marker, so array access was removed from code: asking any `pplx-*` model for the literal line `print(arr[0], arr[12])` returned `print(arr, arr)`, and the same loss landed inside `write_file` arguments, so generated code was corrupted on disk rather than only in the rendered answer. List literals went the same way (`x = [0]` -> `x =`).
- Citation cleanup now skips protected regions instead of running over the whole string: fenced code blocks, `<tool>` payloads and inline code spans, via `CODE_SPAN_RE` + `stripCitations()`. Prose behaviour is unchanged, including the leading space folded into `CITATION_RE` by #14009.
- `CODE_SPAN_RE` alternation order is deliberate: closed regions first, then the unterminated tails so a stream cut off mid-answer still protects the partial fence or `<tool>` block, and the inline span last so the third backtick of a fence is never taken for an empty `` `` `` span. That last ordering point is load-bearing — with the inline alternative first, ` ```py\nz = c[7] ` lost its subscript.

Follow-up to the reviewer note on #14009 ("`CITATION_RE` still strips any `[n]` token, so array indexing like `arr[0]` in code output is still mangled. Worth a follow-up issue").

## Related Issues

- Closes #14121
- Follow-up to #14009 / #13968

## Validation

- [x] Change type: provider
- [x] Focused test: `tests/unit/perplexity-web-code-brackets.test.ts` (8 tests) — fenced code, inline spans, `<tool>` payloads, citations adjacent to protected regions, unterminated tails, and the #13968 indentation cases pinned again
- [x] Every `tests/unit/*perplexity*.test.ts` suite: 91 pass, 0 fail
- [x] `prettier --check` clean on the touched files; `eslint --suppressions-location config/quality/eslint-suppressions.json` exits 0
- [x] Branched from `release/v3.8.51`
- [x] Production-code change includes a new automated test in this PR
- [x] Changelog fragment: `changelog.d/fixes/14121-perplexity-web-code-brackets.md`

## Tests Added Or Updated

- `tests/unit/perplexity-web-code-brackets.test.ts` (new): `print(arr[0], arr[12])` and `x = [0]` survive inside a fence; `` `arr[0]` `` survives inline; a `<tool>` payload containing `arr[0]` and `m[12]` is untouched; prose citations (`text [1] more`, `[1][2]`, `text [3].`, `text[4] more`) still clean exactly as before; citations immediately before and after a protected region are still stripped; an unterminated fence or `<tool>` block protects its tail; the #13968 indentation and `strip = false` passthrough cases are re-pinned.

## Coverage Notes

- `open-sse/executors/perplexity-web/protocol.ts`: `CODE_SPAN_RE` and every branch of `stripCitations()` are exercised directly by the new test. No other file changed.

## Reviewer Notes

- Reproduced on `pplx-sonnet`, `pplx-gpt-5.6-sol`, `pplx-kimi` and `pplx-grok-4.6` against a live Perplexity session; the fix itself is a pure string transform, and the new unit test covers it offline.
- Deliberately not narrowing `CITATION_RE` by lookbehind (for example, "not preceded by a word character"). That would keep `arr[0]` but still eat a bare `x = [0]`, and it would change prose behaviour; skipping code regions leaves the prose path byte-identical.
- Inline spans are matched as `` `[^`\n]+` `` — single-backtick, non-empty, single-line. Multi-backtick inline spans (`` ``a`b`` ``) are not handled; they seemed vanishingly rare in this path next to the complexity of a full inline-code parser.
- Unrelated and still open: `pplx-sonnet`, `pplx-gpt-5.6-sol` and `pplx-kimi` currently refuse the emulated tool protocol outright (#12104); `pplx-grok-4.6` still emits calls, which is how the `<tool>` payload case above was confirmed end to end.


diegosouzapw:
Verified the root cause exactly — `CITATION_RE` is a bare `[n]` matcher with no code-awareness, so any subscript got eaten. The protected-region split approach is correct and the alternation ordering in `CODE_SPAN_RE` (closed regions → unterminated tails → inline span last) is a real subtlety that's easy to get backwards; good that it's called out in the comment. Merging as-is.

diegosouzapw:
Thanks @Zicocoder — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #14124, opened 2026-09-18
feat(providers): add Unifically as an OpenAI-compatible gateway

iliadg:
Adds [Unifically](https://unifically.com) as an OpenAI-compatible API-key gateway. Disclosure: I run Unifically.

Registry entry in the `eurouter` / `greenpt` shape (#13025, #13024): default executor, no translator, `passthroughModels: true`, `models: []`.

## Checked live, 2026-09-18

```
GET  https://api.unifically.com/v1/models                -> 200, OpenAI-shaped list, 155 models
GET  https://api.unifically.com/v1/models?category=llm   -> 200, 30 models
GET  https://api.unifically.com/v1/models  (no key)      -> 401
POST https://api.unifically.com/v1/chat/completions      -> 200, chat.completion
POST .../chat/completions  stream:true                   -> SSE chat.completion.chunk, final usage chunk, data: [DONE]
```

Auth is `Authorization: Bearer <key>`. Streaming was exercised with `google/gemini-3.5-flash`.

## Why `modelsUrl` carries `?category=llm`

The catalog is one list for text, image, video and audio (`category`: 30 llm, 41 image, 50 video, 34 audio). Only the `llm` models are chat-completion targets. The media models run through an async task API (`POST /v1/tasks`, then poll or webhook), which the default executor does not speak. An unscoped listing would offer 125 models that fail on every request, so discovery is scoped and there is a test pinning it.

## What is not claimed

- No tool or vision capability is declared. Neither was exercised.
- `hasFree` is `false`. New accounts get a one-time starting balance, not a recurring free tier, so no badge.
- `serviceKinds` is `["llm"]` only. Media support would need an executor in the shape of `open-sse/executors/kie.ts` plus `videoRegistry` / `audioRegistry` entries. I can send that as a follow-up if it is wanted.
- One known gap: in the streamed response the final usage chunk reported `completion_tokens: 0` while the non-streamed call reported 3. That is on our side and I am fixing it upstream; it does not affect routing.

Added to `AGGREGATOR_PROVIDER_IDS`, since it routes to third-party model providers.

## Verification

`tests/unit/unifically-provider.test.ts`: 7 passed. `tests/unit/eurouter-provider.test.ts` still 7 passed. Prettier and ESLint clean on the touched files. Changelog fragment included. `gateways.ts` grows by 19 lines of catalog data, so `file-size-baseline.json` will want the usual rebaseline note on merge.


iliadg:
Closing this. I will resubmit it from the project's own account.

PR #14125, opened 2026-09-18
docs(DOCKER_GUIDE): explain the `-web` image variant where image users will find it (#14105)

### LINKED ISSUE #14105: docs(docker): explain the `-web` image variant where image users will find it
## Problem / Use Case

The `-web` requirement is documented — but only in places a user of the **published image** never opens.

`Dockerfile`, stage `runner-web`:

> `Use runner-web when you need web-cookie providers (gemini-web, claude-web, claude-turnstile).`
> `For all other providers runner-base is sufficient.`

`contrib/podman/README.md` repeats the same list. Both are build-from-source paths.

Someone who runs `docker pull diegosouzapw/omniroute:<version>` and reads `docs/guides/DOCKER_GUIDE.md`
sees the `-web` tags in the **Release Channels** table (`:<version>-web`, `:latest-web`, `:next-web`,
`:main-web`) with no explanation of what the suffix means or who needs it. Searching that whole guide for
`playwright`, `chromium` or `browser` returns only two hits, both about `NEXT_PUBLIC_BASE_URL` as a
browser-facing origin for OAuth callbacks. The `Dockerfile Stages` table lists `builder`, `runner-base`
and `runner-cli` — `runner-web` is not among them, so the guide does not connect the stage to the tag.

Picking the tag without the suffix fails **at request time, not at startup**: the models are listed and
the provider shows as connected in the dashboard, and only sending a request produces

```
[500]: Failed to load external module playwright: Error: Cannot find module
'/app/node_modules/playwright/node_modules/playwright-core/browsers.json'
```

That message points inside `node_modules`, which reads like a corrupt dependency tree rather than a
deliberate packaging split, and it names no remedy. The remedy also differs by install method — a
different image tag on Docker, a browser download on npm.

Reproducible against the published images (both pulled and run on `linux/amd64`, v3.8.50):

```bash
docker run --rm diegosouzapw/omniroute:3.8.50 \
  node -e "require('/app/node_modules/playwright')"
# Error: Cannot find module '.../playwright/node_modules/playwright-core/browsers.json'

docker run --rm diegosouzapw/omniroute:3.8.50-web \
  node -e "require('/app/node_modules/playwright')"
# exits 0
```

On `3.8.50-web` a real `gweb/*` completion then succeeds through the gateway.

This is **not** a bug report and **not** a request to change how the images are built. The split is
right: most deployments never touch a browser-automating provider and should not carry Chromium.
The ask is only that the guide say the split exists and who is on each side of it.

## Proposed Solution

Add one `###` subsection to `docs/guides/DOCKER_GUIDE.md`, under `## Image Tags`, right after the
`### Release Channels` table. It is an H3, so the H2-only table of contents needs no change. No sizes or
browser builds are baked in, so it will not go stale per release.

---

### The `-web` variant

Every channel above is published twice: a default image and a `-web` image built from the same release.
The `-web` image additionally bundles Playwright and a Chromium build, with `PLAYWRIGHT_BROWSERS_PATH`
pointing at the preinstalled browsers, and is correspondingly larger.

| Tag | Playwright + Chromium | Use it when |
| --- | --- | --- |
| `:<version>`, `:latest`, `:next`, `:main` | No | Default choice for API-key providers. |
| `:<version>-web`, `:latest-web`, `:next-web`, `:main-web` | Yes | You enable a web-cookie provider that drives a real browser — `gemini-web`, `claude-web` or `claude-turnstile`. |

Web-cookie providers are listed and appear connected on an image without the `-web` suffix, but their
requests fail at runtime with:

```
[500]: Failed to load external module playwright: Error: Cannot find module
'/app/node_modules/playwright/node_modules/playwright-core/browsers.json'
```

Switch to the matching `-web` tag and recreate the container. No other configuration change is required.

```bash
docker pull diegosouzapw/omniroute:latest-web
```

---

Happy to open the PR for this if you would rather review a diff than a proposal — just say so.

## Alternatives Considered

- **Leave it to the Dockerfile comment.** It is accurate and already there, but it only reaches people
  who build from source. Users who pull the published image never see it.
- **Make the runtime error name the remedy.** More useful, but it touches the provider loader rather
  than the docs, and the fix differs per install method. Out of scope here; happy to file it separately
  if that is wanted.
- **Ship Playwright in the default image.** Rejected — it would add several hundred MB for the majority
  of deployments that never use these providers.

## Acceptance Criteria

- `docs/guides/DOCKER_GUIDE.md` states what the `-web` suffix adds to the image.
- It names the providers that require it, consistent with the `runner-web` comment in `Dockerfile`.
- It shows the runtime error that appears on the wrong variant, so the message is searchable and lands
  on this section.
- A reader choosing a tag for the first time can decide between the two variants without opening the
  `Dockerfile`.

## Area

`Documentation`

## Related Provider(s)

`gemini-web`, `claude-web`

## Additional Context

Observed sizes on Docker Hub for 3.8.50 are ~1238 MB and ~1705 MB (uncompressed). The `Dockerfile`
comment cites ~500 MB and ~800 MB for the corresponding stages; I did not investigate the difference and
it does not affect the request — no figures appear in the proposed text for that reason.

Related issues, as context rather than duplicates:

- #3379 — `[BUG] Include Playwright on docker images` (closed, completed).
- #3516 — `[BUG] Gemini Web provider fails when Playwright Chromium executable is missing` (closed, completed).
- #9599 — `gemini-web models fail to load Playwright (Incorrect nested path in dist)` (closed, not planned).
  Same error string as above, but reported on a global npm install rather than Docker.

If the `-web` variant is explained somewhere else in the docs site or the release notes, point me at it
and I will close this — my search covered `DOCKER_GUIDE.md`, the README and `docs/providers/`.

Mistertelecom:
Documents the `-web` variant where a **published-image user** actually looks (#14105):

- **Release Channels** — new `-web` subsection: what the suffix adds (Playwright + Chromium, built from `runner-web`), who needs it (`gemini-web`, `claude-web`, `claude-turnstile`), the deferred failure symptom (the provider lists its models and shows as connected; the *first request* returns the `Failed to load external module playwright … browsers.json` 500), and the remedy per install method — pull the `:<channel>-web` tag on Docker, `npx playwright install chromium` on npm/CLI installs.
- **Dockerfile Stages** — adds the `runner-web` row (the table previously stopped at `runner-cli`), fixes the stage count, and adds the `docker build --target runner-web` example.
- **Available Profiles** — adds the `web` profile row (the row `SELF_HOST_GUIDE` already links to from its web-cookie troubleshooting item) and drops the stale "four profiles" count.

Docs gates run locally:

```
[doc-links] scanned 156 docs, checked 1140 internal links — PASS
[docs-frontmatter] OK — 126 compiled doc(s) carry a valid frontmatter title.
[fabricated-docs] ✓ No fabricated API/env/CLI/hook/file references found.
```

(The file also went through `prettier --write`, matching the repo's `lint-staged` rule for `*.md`.)

Closes #14105.


diegosouzapw:
Good catch putting this where actual image users will see it instead of only in the Dockerfile comments. Verified the runner-web stage and web compose profile match your description exactly. Merging.

diegosouzapw:

PR #14140, opened 2026-09-18
fix(combo): exempt per-model-quota 403 from connection exhaustion (#14136)

### LINKED ISSUE #14136: [BUG] Model-scoped 403 on per-model-quota providers marks the whole connection exhausted in combos (reproduced with alibaba)
### OmniRoute Version

3.8.50 (official image `diegosouzapw/omniroute:3.8.50`). The same logic is still present on `release/v3.8.51`.

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Node.js Version

_No response_

### Provider(s) Involved

`alibaba` (reproduced: Model Studio / DashScope, International region `global-sg`, API key auth). By code reading, any provider where `hasPerModelQuota()` is true.

### Model(s) Involved

`qwen3.8-max` (built-in alias → `qwen3.8-max-preview`) fails. `deepseek-v4-pro` and `qwen3.7-max` on the same connection are healthy.

### Client Tool

Open WebUI (OpenAI-compatible `/v1/chat/completions`, streaming, with tools)

### Description

In combos, a **model-scoped** 403 on a per-model-quota provider marks the whole connection exhausted, so every other target on that connection is skipped, even when those models work with the same key.

`applyComboTargetExhaustion` sends every 401/403 (except Cloudflare 1010, quota/credits bodies, and the Alibaba free-quota-exhausted text) to `markAuthLevelExhaustion`, which adds `provider:connectionId` to `exhaustedConnections`. All remaining targets on that connection are then skipped via `getExhaustedTargetSkipReason`. The check does not depend on the provider:
https://github.com/diegosouzapw/OmniRoute/blob/7cc454d931da506722bcbfa93873be6d97a1d5fc/open-sse/services/combo/targetExhaustion.ts#L204-L225

This is inconsistent with the auth layer. Since #3027, `src/sse/services/auth.ts` treats a 403 on a per-model-quota provider (`hasPerModelQuota(provider, model, connectionPassthroughModels)`) as a model-only lockout: the connection stays active. The combo layer never got the same rule, so the two layers disagree on the same response:

```
AUTH  Model-only lockout for alibaba:qwen3.8-max — 403 forbidden 6s (per-model quota provider, connection stays active)
COMBO Provider alibaba connection <id> auth failure (403) — marking for skip on remaining targets (#8133)
```

**Reproduced with `alibaba`.** Alibaba Model Studio returns `403 Access denied` when the account/workspace is not authorized for a *specific* model (e.g. a preview model that has not been granted). The key itself is valid: other models on the same connection return 200 in the same second. The only Alibaba exemption in the combo layer today is `isAlibabaFreeQuotaExhaustedError`, and a plain "Access denied" does not match it.

I have only reproduced this with `alibaba`. From reading the code, the same path applies to every provider where `hasPerModelQuota()` is true (gateways/passthrough, compatible, gemini, github, …) whenever it returns a model-scoped 403.

### Steps to Reproduce

1. Add an `alibaba` API-key connection whose account can call `deepseek-v4-pro` and `qwen3.7-max` but is **not** authorized for `qwen3.8-max-preview`.
2. Create a `priority` combo: `alibaba/qwen3.8-max` → `alibaba/deepseek-v4-pro` → `alibaba/qwen3.7-max`, all on that same connection.
3. Send a chat completion to the combo.

### Expected Behavior

The 403 is scoped to `qwen3.8-max` on that connection, as the auth layer already does for per-model-quota providers. The combo falls through to `deepseek-v4-pro` and returns 200.

### Actual Behavior

The whole connection is marked exhausted. Targets 2 and 3 are skipped without being attempted, and the client receives the 403.

The combo only succeeds when a request arrives while the model-only cooldown (6–24s) is active: target 1 is then skipped *before* dispatch and target 2 runs. From the client's side this looks like random, intermittent failures.

Workaround: put the fallbacks first, or split the failing model onto a second connection with the same key.

### Test Impact

Needs a new unit test

### Error Logs / Output

```shell
# Failing request:
COMBO Trying model 1/3: alibaba/qwen3.8-max
ALIAS Model alias applied: qwen3.8-max → qwen3.8-max-preview
AUTH  Model-only lockout for alibaba:qwen3.8-max — 403 forbidden 12s (per-model quota provider, connection stays active)
COMBO Provider alibaba connection <id> auth failure (403) — marking for skip on remaining targets (#8133)
COMBO Model alibaba/qwen3.8-max failed, trying next
COMBO Skipping alibaba/deepseek-v4-pro — connection <id> for provider alibaba had connection error (#1731v2)
COMBO Skipping alibaba/qwen3.7-max — connection <id> for provider alibaba had connection error (#1731v2)
COMBO All models failed | alibaba/qwen3.8-max: auth — [403]: Access denied (HTTP 403)

# A few seconds earlier, same combo, while the model-only cooldown was active:
COMBO Skipping alibaba/qwen3.8-max — model locked by resilience (cooldown active)
COMBO Trying model 2/3: alibaba/deepseek-v4-pro
COMBO Model alibaba/deepseek-v4-pro succeeded (11167ms, 0 fallbacks)
```

### Screenshots

_No response_

### Additional Context

Related:
- #3027: the auth layer's rule for this case (per-model 403 on a per-model-quota provider → model-only lockout, connection stays active). The combo layer should follow the same rule.
- #8133 / PR #8195: introduced the connection-wide auth skip in combos.
- #12441 / PR #12449: exempted quota/credits bodies from the auth skip. #12441 also mentioned 403s being over-skipped, but the fix only covered quota/credits text. This is the per-model-access variant and is not covered.

Proposed fix: in the 401/403 branch of `applyComboTargetExhaustion`, a **403** should stay model-scoped (return `false`, so the existing model-only lockout handles it) under the same condition `auth.ts` uses: `hasPerModelQuota(provider, model, connectionPassthroughModels)`, including its exclusions (e.g. `isVertexConnectionWidePermissionDenied`). Sharing one predicate keeps both layers in agreement. **401** keeps the connection-wide skip.

Trade-off: if a per-model-quota provider returns 403 for a truly revoked key, the combo would try each sibling target once (each then gets its own model lockout) instead of skipping them at once. That costs one extra upstream call per sibling target on the same connection before the combo gives up. 401, the usual invalid-key status, is unaffected. If a smaller change is preferred, a narrower alternative is to apply the exemption only when `isAlibabaModelStudioProvider(provider)`.

I'm happy to open the PR.

### Validation Plan

- `node --import tsx --test tests/unit/combo/combo-target-exhaustion.test.ts` with new cases: `403` + per-model-quota provider (`alibaba`) + non-quota body + `connectionId` → returns `false` and does not add to `exhaustedConnections`; `401` on the same provider still marks the connection; existing #8133 cases (401 for any provider, 403 for non-per-model providers) keep passing.
- Manual (alibaba): the combo above returns 200 from `deepseek-v4-pro` on the first request, and logs show no `#1731v2` skip for the sibling targets.

rafacpti23:
## Summary
Fixes #14136

Exempts HTTP 403 errors on per-model-quota providers (such as Gemini, Vertex, Codex, Antigravity, and passthrough models) from marking connection-level or provider-level exhaustion in combo target dispatching (pplyComboTargetExhaustion), unless verified as a connection-wide Vertex denial (SERVICE_DISABLED or non-models IAM denial).

For per-model-quota providers, a 403 indicates a model-scoped tier restriction or access denial rather than an invalid credential, allowing sibling combo legs on the same connection to remain eligible.

## Testing
- Added unit tests in 	ests/unit/combo/combo-target-exhaustion.test.ts verifying:
  - 403 on per-model-quota provider does NOT mark connection or provider exhausted
  - 403 on vertex with model-scoped permission denial does NOT exhaust connection
  - 403 on vertex with connection-wide permission denial DOES exhaust connection
- All 44 unit tests in combo-target-exhaustion.test.ts pass.


abhisheksharma2411 (commented):
Correct fix, and the part I'd call out as right is that it **mirrors the existing precedent instead of inventing one**. `src/sse/services/auth.ts:3190-3196` already applies the same shape to the single-request path:

```ts
isPerModelQuotaProvider && (status === 403 || status === 402) && … &&
!(provider === "vertex" && isVertexConnectionWidePermissionDenied(errorText))
```

and this adds the combo-dispatch path the same way. Two sites deciding the same question differently is exactly how the model-scoped/connection-scoped distinction goes wrong, so matching it is better than a cleverer local answer.

The failure this prevents is the one the codebase already documents for the analogous provider — a connection marked exhausted by a model-scoped denial takes its healthy sibling legs down with it, and a combo then has nothing left to route to even though only one model was actually restricted.

**One thing worth raising as a follow-up rather than against this PR**, since it applies equally to both call sites: `isVertexConnectionWidePermissionDenied` is written against **Google's standard error envelope**, not anything Vertex-specific. It parses `error.details[]` for `SERVICE_DISABLED` and for `IAM_PERMISSION_DENIED` whose `metadata.resource` lacks `/models/`. `SERVICE_DISABLED` is what *any* Google API returns when the service is not enabled on the project — including `generativelanguage.googleapis.com` for a plain Gemini key.

So a Gemini connection whose Generative Language API is disabled returns a connection-wide 403 that both sites will now read as model-scoped, and the combo will burn one 403 per remaining Gemini leg before giving up. Fail-slow rather than fail-wrong, and rarer than the case this PR fixes — but the `provider === "vertex"` gate is the only thing scoping a format-generic classifier to one provider, and the function name is the reason nobody notices.

I'd land this as-is (consistency with `auth.ts` is worth more than a one-off divergence here) and treat "should the connection-wide check be provider-agnostic, applied wherever the error carries Google's `details[]`?" as its own issue touching both call sites. Happy to file it if you want it tracked.

Test coverage looks right — asserting both `exhaustedConnections.size` and `exhaustedProviders.size` rather than just the return value is what makes it a real regression guard, since the return value alone would still pass if the sets were populated as a side effect.

diegosouzapw:
Thanks @rafacpti23 — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #14143, opened 2026-09-18
fix(combos): use provider node prefix alias in combo builder qualifiedModel (#14135)

### LINKED ISSUE #14135: fix(api): /v1/models catalog emits a degraded duplicate entry keyed by the node UUID (vision dropped, context_length 128000) and input_modalities=null on the aliased entry
## Summary

When a combo member is picked in the GUI for a **custom `openai-compatible` provider**, the combo builder stores the member under the connection's **node UUID** (`openai-compatible-chat-<uuid>/Model`) instead of the provider's **prefix/alias** (`of/Model`).

That node-scoped id:

1. produces a **second, degraded `/v1/models` entry** for the same model (vision capability dropped, `context_length` 128000 instead of 1048576), because the synced-capability / canonical-metadata lookups are keyed by alias/canonical provider and **miss** on the node UUID;
2. is what downstream consumers (e.g. the OpenCode plugin resolving combo members) see — so a vision-capable combo member resolves as **non-vision**, and any strict "all members must be vision-capable" capability intersection publishes the whole combo as text-only.

Net effect: a vision-capable combo silently loses its vision capability (and reports a wrong context window) purely because of the stored id form.

## Environment

- OmniRoute self-hosted (current main), `data/storage.sqlite`
- Custom provider node: `type: openai-compatible`, `name: "Openference"`, `prefix: "of"`
- Affected model: `MiniMax M3` (vision-capable, 1M context)

## Evidence chain (verified against live DB + live `/v1/models`)

**1. Provider node** (`provider_nodes`):
```json
{"id":"openai-compatible-chat-01f72ee6-b216-48f7-b4d9-e279a7479953",
 "type":"openai-compatible","name":"Openference","prefix":"of"}
```

**2. Connections** (`provider_connections`) — keyed to the **node UUID**, not `of`:
```json
{"provider":"openai-compatible-chat-01f72ee6-…","name":"main","is_active":1}
```

**3. Combo `vision`** (`combos.data`) stores the member **node-scoped**:
```json
{
  "kind": "model",
  "model": "openai-compatible-chat-01f72ee6-b216-48f7-b4d9-e279a7479953/MiniMax M3",
  "providerId": "openai-compatible-chat-01f72ee6-b216-48f7-b4d9-e279a7479953",
  "weight": 0
}
```

**4. Live `/v1/models`** emits **two entries for the same model**, same `owned_by: "of"`, different metadata:
```jsonc
// healthy (alias form)
{"id":"of/MiniMax M3","owned_by":"of",
 "capabilities":{"vision":true,"tool_calling":true,"reasoning":true},
 "context_length":1048576}

// degraded (node-scoped form) ← what the combo member resolves to
{"id":"openai-compatible-chat-01f72ee6-…/MiniMax M3","owned_by":"of",
 "capabilities":{"tool_calling":true,"reasoning":true},
 "context_length":128000}
```

**5. Synced capabilities** (`model_capabilities`) hold the correct data, but keyed to a canonical provider/model pair (e.g. `provider:"opper", model_id:"minimax/m3"`, `modalities_input:["text","image","video"]`, `limit_context:1048576`). A query for `provider = 'of' OR provider LIKE '%01f72ee6%' OR model_id = 'MiniMax M3'` returns **zero rows** — so `getSyncedCapability(nodeUuid, "MiniMax M3")` misses and the catalog falls back to a 128000 default with no vision.

## Root cause

`src/lib/combos/builderOptions.ts` → `getComboBuilderOptions()` groups models by `connection.provider` (the **node UUID** for custom compatible providers) and builds `` qualifiedModel: `${providerId}/${modelId}` ``. `addModelOption()` stores that node-scoped value as the combo step's `model`/`providerId`. There is **no canonicalization of the node UUID back to the node's `prefix`/alias** at build time.

Downstream, `src/app/api/v1/models/catalog.ts` resolves capabilities via `getSyncedCapability(providerId, modelId)` / `getCanonicalModelMetadata(...)`, keyed by alias/canonical provider. The node UUID misses → degraded duplicate entry (vision dropped, `context_length` → 128000 default fallback).

## Two coupled defects

| # | Location | Defect |
|---|----------|--------|
| A (root) | `builderOptions.ts` | Stores node-scoped `openai-compatible-chat-<uuid>/Model` as combo member instead of alias `of/Model` |
| B (amplifier) | `catalog.ts` | Node-scoped id misses `model_capabilities` (keyed by canonical provider) → emits degraded duplicate (vision dropped, context defaults to 128000) |

## Recommended fix

**Primary (fixes both at the source):** in `builderOptions.ts`, canonicalize the provider key to the node's prefix/alias before building `qualifiedModel` and the stored step `providerId`:

```ts
// providerIdToPrefix is already built in catalog.ts from provider_nodes.prefix
const canonicalProviderId = providerIdToPrefix[providerId] || providerIdToAlias[providerId] || providerId;
qualifiedModel: `${canonicalProviderId}/${modelId}`   // → "of/MiniMax M3"
```

**Defense-in-depth:** in `catalog.ts`, when emitting a combo-target / node-scoped entry, resolve capabilities through the canonical provider (`resolveCanonicalProviderId` + `model_capabilities` fallback) so a node-scoped id never silently degrades to the 128000/no-vision default.

**Migration:** existing combos already store node-scoped member ids — a one-time normalization pass over `combos.data` (rewrite `openai-compatible-chat-<uuid>/X` → `<prefix>/X` using `provider_nodes.prefix`) is needed, otherwise already-saved combos stay broken after the builder fix.

## Impact

Any vision-capable model added to a combo via a custom `openai-compatible` provider is treated as text-only by capability-intersection consumers (e.g. the OpenCode plugin omits the combo from image-capable routing), and is listed in `/v1/models` with a wrong context length (128k vs real 1M) and dropped vision/attachment modalities. Built-in providers are unaffected because their `connection.provider` is already the canonical alias.

rafacpti23:
## Summary
Fixes #14135

For custom provider nodes (e.g. `openai-compatible-chat-<uuid>`) configured with a prefix alias (e.g. `of`), the combo builder previously used the raw provider ID when constructing `qualifiedModel` for provider node options, causing:
1. Combo steps to serialize with the raw node UUID instead of the prefix alias.
2. Capability and catalog resolution (such as in `/v1/models`) to fail looking up the canonical prefix.

This change ensures that custom provider nodes (`providerVisual.source === "provider-node"`) with a configured alias use their alias prefix as `routingPrefix` in `getComboBuilderOptions()`.

## Testing
- Added unit test in `tests/unit/combo-builder-custom-provider-alias-14135.test.ts` verifying prefix alias qualifiedModel assignment, global model list step building, and manual combo step resolution.
- Validated with `node --import tsx/esm --test tests/unit/combo-builder-custom-provider-alias-14135.test.ts` (100% pass).
- Verified related combo tests pass without regression.


diegosouzapw:
Solid catch and a clean fix — gating on `providerVisual.source === "provider-node"` is exactly right, so system providers and the existing opencode special-case keep their current behavior while custom nodes get their configured alias instead of the raw DB UUID. I reverted just this hunk locally to confirm the new test actually fails without it (raw `openai-compatible-chat-<uuid>/...` instead of `of/...`), then reapplied it and it passes — good regression coverage. Merging.

diegosouzapw:
Thanks @rafacpti23 — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #14144, opened 2026-09-18
fix(auth): support OIDC IdP issuer with trailing slash (#14119)

### LINKED ISSUE #14119: fix(auth): OIDC callback strips trailing slash from issuer, rejecting ID tokens from IdPs whose iss ends in / (Authentik)
### OmniRoute Version

3.8.50 (also present on `release/v3.8.51`)

### Installation Method

Docker / Docker Compose (Kubernetes, `diegosouzapw/omniroute` image)

### Description

Native OIDC dashboard login cannot validate ID tokens from any identity provider whose `issuer` ends with a trailing slash. Authentik is the common case: every Authentik OAuth2 provider advertises `issuer: "https://<host>/application/o/<slug>/"` (per-provider mode) or `"https://<host>/"` (global mode). Both end in `/`, and the slash is part of the `iss` claim Authentik signs.

Both OIDC routes normalise the configured issuer by stripping one trailing slash:

- `src/app/api/auth/oidc/login/route.ts` L14-15 (v3.8.50):
  https://github.com/diegosouzapw/OmniRoute/blob/5458026c216f77a3da68ea49152dc33470cfe2cb/src/app/api/auth/oidc/login/route.ts#L14-L15
- `src/app/api/auth/oidc/callback/route.ts` L72-73 (v3.8.50):
  https://github.com/diegosouzapw/OmniRoute/blob/5458026c216f77a3da68ea49152dc33470cfe2cb/src/app/api/auth/oidc/callback/route.ts#L72-L73

```ts
const issuer =
  typeof settings.oidcIssuer === "string" ? settings.oidcIssuer.trim().replace(/\/$/, "") : "";
```

That stripped value is used for two different things. Building discovery URLs (`${issuer}/.well-known/openid-configuration`, L101) is fine. It is also passed as the expected issuer to jose (callback L161-163):

https://github.com/diegosouzapw/OmniRoute/blob/5458026c216f77a3da68ea49152dc33470cfe2cb/src/app/api/auth/oidc/callback/route.ts#L161-L163

```ts
const { payload } = await jwtVerify(idToken, JWKS, {
  issuer,
  audience: clientId,
});
```

jose compares `iss` by exact string match. The token carries `https://<host>/application/o/<slug>/` while OmniRoute expects `https://<host>/application/o/<slug>`, so every login ends in `/login?oidc_error=id_token_invalid`.

There is no configuration-only workaround. Entering the issuer with a double slash (`.../<slug>//`) survives the strip, but discovery then requests `.../<slug>//.well-known/openid-configuration`, which Authentik answers with 404. Authentik's `issuer_mode: global` does not help either: that issuer is `https://<host>/` (still a trailing slash), and Authentik serves no discovery document at the host root (404).

Authentik source for reference (2026.8.2, `authentik/providers/oauth2/models.py` `OAuth2Provider.get_issuer`): per-provider mode returns `build_absolute_uri(reverse("authentik_providers_oauth2:provider-root", ...))`, and global mode returns `build_absolute_uri(reverse("authentik_core:root-redirect"))`. Both resolve to URLs that end in `/`.

### Steps to Reproduce

1. Create an Authentik OAuth2/OpenID provider (confidential, authorization code) with redirect URI `https://<omniroute>/api/auth/oidc/callback`.
2. In OmniRoute settings set `oidcEnabled: true`, `oidcIssuer: https://<authentik>/application/o/<slug>/` (the exact `issuer` from Authentik's discovery document), client id/secret, and `oidcAllowedSubjects`.
3. Click "Continue with OIDC" on `/login` and complete the Authentik login.

### Expected Behavior

The ID token validates against the issuer the IdP advertises, and the dashboard session is created.

### Actual Behavior

Redirect to `/login?oidc_error=id_token_invalid`. jose rejects the token with an `iss` claim mismatch: `https://<authentik>/application/o/<slug>/` from the token vs the expected `https://<authentik>/application/o/<slug>`.

### Test Impact

Needs a new unit test

### Additional Context

Suggested fix: keep the trimmed-but-unstripped issuer for token validation, and use the slash-stripped form only to build URLs. Preferably, validate against the `issuer` value returned by the discovery document when it is present (OIDC Discovery 1.0 §4.3 requires that value to match the `iss` in tokens exactly). For example:

```ts
const configuredIssuer = settings.oidcIssuer.trim();
const issuerBase = configuredIssuer.replace(/\/$/, "");
// discovery: `${issuerBase}/.well-known/openid-configuration`
// validation: jwtVerify(idToken, JWKS, { issuer: discovered.issuer ?? configuredIssuer, audience })
```

`tests/unit/oidc-callback.test.ts` only uses `oidcIssuer: "https://idp.test"` (no trailing slash), so it does not cover this case. A case with `oidcIssuer: "https://idp.test/app/"` and a token whose `iss` is `"https://idp.test/app/"` would reproduce it.

### Validation Plan

- `node --import tsx --test tests/unit/oidc-callback.test.ts` with a new trailing-slash issuer case
- End-to-end login against an Authentik 2026.x per-provider issuer

rafacpti23:
## Summary
Fixes #14119

Some OIDC Identity Providers (such as Authentik) format their canonical `issuer` with a trailing slash (e.g., `https://idp.domain/application/o/slug/`).

Previously:
1. Discovery requested `${issuerBase}/.well-known/openid-configuration` but ignored `data.issuer` returned by discovery when validating ID tokens.
2. `jwtVerify` was called with only `issuerBase` (trailing slash stripped), causing `JWTClaimValidationFailed: "iss" claim mismatch` when Authentik returned tokens with a trailing slash in the `iss` claim.

This fix:
1. Parses `discoveredIssuer` from OIDC discovery response if present.
2. Generates candidates for `expectedIssuers` matching configured raw issuer, base issuer, discovery issuer, and their trailing/non-trailing slash variants.
3. Passes `expectedIssuers` to `jwtVerify` issuer option so tokens from providers with or without trailing slash pass claim verification seamlessly.

## Testing
- Unit test suite `tests/unit/oidc-callback.test.ts` updated to include Authentik-style trailing slash test case (`OIDC callback handles issuer with trailing slash in settings and token (#14119)`).
- All 12 unit tests in `tests/unit/oidc-callback.test.ts` pass (100%).
- All unit tests in `tests/unit/oidc-login-state.test.ts` pass (100%).


diegosouzapw:
This is a solid, well-scoped fix — discovering token/JWKS endpoints against the trimmed issuer and then accepting the discovered issuer alongside both trailing-slash variants in `jwtVerify`'s `issuer` option (which `jose` accepts as a `string[]`) correctly handles IdPs like Authentik that publish a trailing-slash issuer. Your two new tests cover both the matching-issuer and mismatched-slash cases well. The only blocker is that the branch is very stale and currently shows as conflicting on GitHub — I checked with `git merge-tree` against the current tip and the OIDC route files themselves have zero real conflicts, so a rebase onto `release/v3.8.51` should be mechanical. Once it's green against the current tip, this is good to merge.

PR #14145, opened 2026-09-18
fix(api): sanitize error message in /v1/batches POST handler (Hard Rule #12)

### LINKED ISSUE #14089: fix(api): /v1/batches returns raw error.message in the response body (Hard Rule #12)

rafacpti23:
## Summary
Fixes #14089

In `POST /api/v1/batches`, the `catch` block was returning `error.message` directly in the HTTP 400 response body without routing it through `sanitizeErrorMessage()`. If an underlying database, filesystem, or driver error was caught during batch creation (e.g., SQLite connection/IO error or filesystem write failure), raw error messages containing internal filesystem paths, stack trace frames, or driver details could leak in the API response.

This fix:
1. Imports `sanitizeErrorMessage` from `@omniroute/open-sse/utils/error`.
2. Applies `sanitizeErrorMessage(error instanceof Error ? error.message : "Create failed")` when constructing the `invalid_request_error` response body.
3. Aligns `src/app/api/v1/batches/route.ts` with repository Hard Rule #12 and the error handling pattern in sibling routes (`/batches/delete-completed`, `/relay/chat/completions`, etc.).

## Testing
- Added regression test in `tests/unit/batch_api.test.ts` verifying that unexpected errors thrown during batch creation are sanitized and do not leak filesystem paths or stack traces.
- All 23 unit tests in `tests/unit/batch_api.test.ts` pass (100%).
- TypeScript and ESLint checks pass cleanly.


abhisheksharma2411 (commented):
Correct and necessary — and the reason it's necessary is worth putting in the PR body, because it isn't obvious from the diff.

**This route hand-rolls the error envelope**, so it never reaches the shared sanitizer:

```ts
return NextResponse.json(
  { error: { message: error instanceof Error ? error.message : "Create failed",
             type: "invalid_request_error" } },
  { status: 400, headers: CORS_HEADERS }
);
```

Everywhere else in `/v1` that goes through `buildErrorBody` is already covered — it calls `sanitizeErrorMessage(message)` internally at `open-sse/utils/error.ts:372`. I checked the two other `/v1` routes that looked unsanitized and both are fine: `models/catalog.ts` passes its raw message *into* `buildErrorBody` (sanitized by the builder), and `messages/count_tokens/route.ts` only puts it in a `log.debug`, not a response. So `batches` really is the outlier, not one of a set.

That points at the more durable fix: route this through `buildErrorBody` like the other 13 files in `/v1` do, rather than adding a second explicit sanitizer call at a hand-rolled site. You'd get sanitization, the standard `type`/`code` classification, and immunity to the next hand-rolled envelope drifting — the split is currently 13 files each way, so this pattern will recur.

**Second thing, and I'd argue it matters more than the leak:** this catch reports an internal failure as `400 invalid_request_error`. The test's own fixture is a filesystem/SQLite write failure — the client did nothing wrong, and telling them their request was invalid sends them to re-check a payload that was fine. `buildErrorBody(500, …, { type: "server_error" })` is both the honest status and the shared path. Out of scope for a sanitization fix, but it's the same three lines.

The test is well built — using a real thrown message with an absolute path *and* a stack frame means it would catch a sanitizer that only stripped one of the two, and stubbing `db.prepare` on the specific `INSERT INTO batches` statement keeps the rest of the fixture working. Worth also asserting the response **status**, so a later refactor that changes 400→500 is a deliberate decision rather than a silent one.

diegosouzapw:
This is a correct fix for a real bug (confirmed the leak on the tip), thank you for catching it. Another PR (#14115) landed the same fix slightly earlier and additionally guards the edge case where the sanitizer can return an empty string (matching the pattern already used in open-sse/utils/error.ts), so we're going with that one and closing this as a duplicate. Appreciate you working on Hard Rule #12 compliance here.

PR #14146, opened 2026-09-18
fix(api): strip stale encoding and content-length headers in provider models route

### LINKED ISSUE #14092: fix(api): /api/v1/providers/{provider}/models forwards the unified catalog's content-length onto a re-serialized filtered body — response never completes (playground model dropdown stuck loading)

rafacpti23:
## Summary
Fixes #14092

When `GET /api/v1/providers/[provider]/models` calls `getUnifiedModelsResponse(request)`, the catalog response may carry `content-length`, `content-encoding`, or `transfer-encoding` headers corresponding to the full uncompressed/compressed catalog payload.

The route filters down the catalog models for the specific provider and returns a re-serialized JSON response. If `response.headers` from the catalog response is passed through unmodified, the original `content-length` (e.g. dozens of kilobytes) is sent with a much smaller response body, causing HTTP clients to hang waiting for remaining bytes until connection timeout.

This fix:
1. Imports `stripStaleEncodingHeaders` from `@omniroute/open-sse/utils/upstreamResponseHeaders`.
2. Wraps `response.headers` with `stripStaleEncodingHeaders(response.headers)` before returning the response.
3. Strips `content-length`, `content-encoding`, and `transfer-encoding` while preserving non-stale headers.

## Testing
- Added regression test in `tests/unit/provider-models-v1-route.test.ts` verifying that stale `content-length`, `transfer-encoding`, and `content-encoding` headers are stripped.
- All 9 unit tests in `tests/unit/provider-models-v1-route.test.ts` pass (100%).
- TypeScript (`npm run typecheck:core`) and ESLint checks pass cleanly.


diegosouzapw:
Nice fix, and good instinct reusing `stripStaleEncodingHeaders()` here instead of hand-rolling the header strip — that's the right idiom. Another PR (#14120) fixes the exact same route plus a second instance of the same bug on the vscode catalog route, so we're taking that one as the base and asking its author to adopt your cleaner approach of reusing the shared utility. Closing this as a duplicate, but your utility-reuse pattern is going in either way.

PR #14148, opened 2026-09-18
fix(opencode): replay a request in the other tool shape when its tools are refused

maxmad64bis:
⚠️ base-red inherited: #13866

## Summary

Default change: none. Behavior change: a request whose tools OmniRoute added itself, answered 403 upstream, is replayed once without them; the accepted replay returns 200 (before: 403). The short title-generation request OpenCode sends beside every chat turn carried a placeholder tool because #14013 filled an empty `tools` array unconditionally, and the upstream accepts a "Never use tools" prompt only with no tools at all and any other only with the client's core tool names; refusals replay in the other shape, with the working shape remembered per prompt.

## Related Issues

- Related to #14013

## Validation

- [x] Change type: provider
- [ ] `npm run lint` — ESLint on the touched files is clean; the full run is red on the base (5 errors in files this PR doesn't touch)

## Tests Added Or Updated

- `tests/unit/opencode-request-shape-retry.test.ts` (new, 34 cases): replay on refusal, remembered shape + reversal, 64-prompt bound, 60 second pause, account health after rescued/double refusals, overlapping requests, compaction case, every new-path `catch`.
- `tests/unit/opencode-free-tier-refusal-rotation.test.ts` (updated): its request now declares its own tool list, see Reviewer Notes.

## Coverage Notes

- `open-sse/executors/opencodeRequestShape.ts` (new): every branch has a failing-without-it test; mutations (replay off, unbounded memory, unset pause, ignored abort) fail the named test. `prepareFreeTierRequest` takes an optional origin; without it the behavior is unchanged (36 existing contract cases). No coverage moved down.

## Reviewer Notes

- Tests use a fake upstream applying the rule measured on 2026-09-18 (third-party gateway); run live, the first title went 403 then 200 in two sends and the same title again went 200 in one send. `opencode-free-tier-refusal-rotation.test.ts` now declares a tool list, because "a refusal is returned as it came" no longer holds once the contract adds the tools itself, with the invariant unchanged for the client's own tools and six cases passing with and without this change.
- A replay can go to another account, with no cooldown or failure mark from a rescued refusal. When both shapes are refused, replays for that prompt pause for 60 seconds (two sends per prompt per minute max, 64 prompts tracked) and the caller gets the second refusal. The attempt is now looked up by the identity of its body and released at request end instead of a field of the shared executor where overlapping requests read each other's (`opencode.ts` 1195 of 1200 lines), while compaction requests already carry the conversation's own tools via #14013 (3 of 3 accepted live) and stay untouched with two pinning tests; unmeasured are compactions above the 52 thousand characters seen and clients other than the official one.


diegosouzapw:

PR #14149, opened 2026-09-18
fix(opencode): keep the target format and client session per request, not on the shared executor

maxmad64bis:
> **Stacked on #14148.** Only the last commit is this PR's; I'll rebase once it lands.

⚠️ base-red inherited: #13866

## Summary

Default change: none. Behavior change: a JSON caller of a Responses model no longer gets the raw event stream back when another request finishes while it waits. The shared executor kept target format and client session as instance fields written at request start and read after later awaits, so a slow Responses JSON request got the event stream (`event: response.completed …`) once a Chat request finished first; both values now live in a per-`execute()` context, with the public fields unchanged outside `execute()`.

## Related Issues

- Related to #14148

## Validation

- [x] Change type: provider
- [ ] `npm run lint` — ESLint on the touched files is clean; the full run is red on the base (5 errors in files this PR doesn't touch)

## Tests Added Or Updated

- `tests/unit/opencode-request-format-race.test.ts` (new, 4 cases): slow Responses JSON caller with a Chat request finishing meanwhile (fails before the change), two in flight each keeping session and format, a request ending not clearing what another still uses, plain fields unchanged outside `execute()`.

## Coverage Notes

- `opencodeRequestContext.ts` (new) and the accessors in `opencode.ts`: mutating the code (format or session always in the shared field, `execute` without a context) makes the named tests fail. The 41 test files that touch the executor pass unchanged. No coverage moved down.

## Reviewer Notes

- It depends on #14148: the wrapper it hooks into, `withRequestShapeRetry` around `executeOnce`, is introduced there. The public `_requestFormat` and `_clientSession` stay as accessors: inside `execute()` they use the request's own context, outside it plain fields, keeping the ten existing tests that assign them unchanged.
- The Muse Spark output-budget helpers move to `opencodeMuseSpark.ts` unchanged and re-exported, taking `opencode.ts` from 1195 to 1125 lines; `isResponsesTerminalLine` is now exported and `reportedCompletionTokens` leaves `normalizeMuseSparkFinishReason`, putting its complexity under the limit. Not changed here: the account list and cursor are also rebuilt per request on the shared instance (same class of problem), and reads of `_requestFormat` after a request ended still see the cleared value, as before.


diegosouzapw:

Issue #13483, opened 2026-09-12
fix(auth): Model-only lockout for no-auth providers is recorded but never enforced — locked member is retried on every request


Xore:
### OmniRoute Version
3.8.51 (built from source, `release/v3.8.51`)

### Installation Method
Built from source

### Operating System
Linux

### Node.js Version
22.23.2

### Provider(s) Involved
`opencode` (no-auth provider); likely every provider listed in `NOAUTH_PROVIDERS` / `WEB_COOKIE_PROVIDERS`

### Model(s) Involved
`opencode/deepseek-v4-flash-free` (upstream currently returns `400 "Model is unavailable"`)

### Client Tool
Any OpenAI-compatible client (`POST /v1/chat/completions`)

### Description
A `400 model_capacity` from a no-auth provider records a model-only lockout (`recordModelLockoutFailure(provider, "noauth", model, "model_capacity", 400, 1800s)` in `src/sse/services/auth.ts`, logged as `Model-only lockout for opencode:deepseek-v4-flash-free — 400 model_capacity 1800s (rule scope=model, connection stays active)`), but the lockout is never consulted for that provider again. On the very next request the combo tries the same member, sends the upstream call, gets the same 400, and records the same 1800 s lockout — every request, forever.

Cause as far as I can tell: in `getProviderCredentials` (`src/sse/services/auth.ts`, the `providerMaps.some((map) => map[resolvedId]?.noAuth)` branch around line 1190) no-auth providers return the synthetic `"noauth"` connection early, before the per-connection status pass that classifies real connections as `modelLocked` (~line 1552, `getModelLockoutInfo(provider, c.id, requestedModel)`). So the lockout keyed on `(opencode, "noauth", deepseek-v4-flash-free)` exists, but nothing on the selection path reads it. The comment for #3061 already notes that the synthetic connection has "no cooldown to brake it" for the in-request fallback loop; the same gap applies across requests for model lockouts.

Impact: each request through the combo pays a wasted upstream round-trip (~2.2–2.5 s here) plus an error row before reaching the healthy member, and the dashboard shows a steady stream of 400s for a member that OmniRoute itself declared locked out for 30 minutes.

### Steps to Reproduce
1. Add a no-auth provider member that the upstream currently rejects with a model-scoped 400 (`opencode/deepseek-v4-flash-free` → `Upstream request failed: Model is unavailable.`) to a priority combo, followed by a healthy member.
[...]

PR #14151, opened 2026-09-18
fix(auth): serve a locked no-auth model as a retryable cooldown

maxmad64bis:
⚠️ base-red inherited: #13866

## Summary

Behavior change: a request for a no-auth model inside a short model lock now waits out the lock (retry wait on by default) or gets `429 model_cooldown` with `Retry-After`, instead of `401 No active credentials`. After one failure (a 502, or the 403s OpenCode started sending on 2026-09-17, #13935) every request for that model got a fatal 401 until expiry, since #13547 makes selection return `null` during the lock. The no-auth path now returns the same "cooling down for this model" outcome as real connections, and a 401 or 403 from a no-auth provider no longer starts a token refresh that can only fail (3 to 6 seconds each, and repeated failures opened the refresh breaker for 30 minutes).

## Related Issues

- Related to #13483
- Related to #13935

## Validation

- [x] Change type: routing
- [ ] `npm run lint` — ESLint on the touched files is clean with the repository suppressions; I haven't run the full repository lint

## Tests Added Or Updated

- `tests/unit/noauth-model-lockout-retryable.test.ts` (new): selection returns the model-scoped cooldown outcome; not-found locks still `null`; others and expired locks untouched.
- `tests/unit/chat-noauth-model-cooldown.test.ts` (new): a real 502 lock is waited out and served through `handleChat`; `429` when the wait is off/shorter, `499` on disconnect; combo skips the locked member at once.
- `tests/unit/noauth-refresh-guard.test.ts` (new): no refresh after a 401 or 403 on a no-auth provider or after a request-scoped refusal; a keyed provider still refreshes.
- `tests/unit/noauth-model-lockout.test.ts` (updated): `null` assertions now expect the cooldown outcome.

## Coverage Notes

- `auth.ts`, `chatCore.ts`, `providerExecutionPipeline.ts` and the two new helpers are covered by these tests; neutralizing either fix makes them fail. No coverage moved down.

## Reviewer Notes

- The lock itself is unchanged, since combos and #13547 rely on it; selection now returns the existing `allRateLimited` shape while it is active so the handler needs no change, applying to every `noAuth` provider as #13547 did. The refresh guard sits at both refresh sites, `providerExecutionPipeline.ts` and `chatCore.ts`. The opencode client retries a 429 and honors `Retry-After`, and does not retry a 401 (`packages/opencode/src/session/retry.ts`); the retry wait (90 s per wait, 300 s total) is off inside combos.
- `getProviderCredentials` for a no-auth provider, 20k calls: 57-64 µs unlocked (57-58 before), 71-76 µs locked (65-67 before). Red on the base too, reproduced on `4d1282be31`: `check-api-typecheck` (`TS2677` in two files), `check:docs-all`, `check:agent-skills-sync` and the mutation gate; `check:changelog-integrity` compares against a stale `origin` on a fork.


diegosouzapw:

PR #12553, opened 2026-09-03
feat(dashboard): show exact token counts on hover in usage analytics cards and tables

ZaimMarzuki:
## Summary
Replaces the plain browser-native `title` tooltip with a styled, portal-based custom tooltip for metric cards in the Usage Analytics dashboard (mirroring the clean chart tooltip design).

### Enhancements
- **Rich Custom Tooltip (`src/shared/components/Tooltip.tsx`):**
  - Upgraded `TooltipProps` to accept `ReactNode` in addition to strings, supporting rich multi-line card layouts.
  - Styled with dark backdrop (`bg-[#10141e]/95`, `border-white/10`, `rounded-lg`, `shadow-xl`, `backdrop-blur-sm`) to match OmniRoute's dashboard palette.
- **Usage KPI Cards (`StatCard` & `CompactStatGrid` in `charts.tsx`):**
  - Displays a clean two-line popover card on hover:
    - **Header:** Card metric label (e.g. `Total Tokens`, `Input Tokens`, `Est. Cost`).
    - **Detail:** Formatted value with indicator (e.g. `tokens : 3,632,356,859 tokens`, `cost : $1,180.27`) using the card's accent color.
  - Also enhanced `MostActiveDay7d` to render matching custom hover popovers.
- **Data Tables (`ModelTable`, `ProviderTable`, `ApiKeyTable`, `RequestCountTable`):**
  - Keeps full formatted number tooltips for fast data inspection.
- **Unit Tests (`tests/unit/ui/analytics-token-hover-tooltip.test.tsx`):**
  - 7 unit tests covering custom tooltip rendering, title/metric content, and portal attachment.

PR #13848, opened 2026-09-16
fix(translator): pair Gemini tool responses per turn to prevent cross-turn ID collision

TheDemonTuan:
## Summary
- Scope tool call responses per turn from adjacent tool messages instead of using a conversation-wide ID map
- Scope tool_call_id -> function name lookup to the specific assistant turn
- Prevents 400 Bad Request caused by functionResponse name/output mismatch when tool_call_ids collide across different turns in long sessions

## Verification
- `node --import tsx/esm --test tests/unit/translator-openai-to-gemini.test.ts` (all 47 tests passed, including regression tests for native and context modes)
- `node --import tsx/esm --test tests/unit/antigravity-orphan-tooluse-7752.test.ts tests/unit/issue-11510-parallel-tool-thought-signature-openai.test.ts tests/unit/translator-antigravity-signature-bypass.test.ts tests/unit/t43-gemini-tool-call-no-thought-signature.test.ts`

PR #14152, opened 2026-09-19
fix(lint): drain the 5 no-unused-vars errors blocking the release-green hard gate

gonisulaimann:
The release-green validation on `release/v3.8.51` currently reports 5 hard ESLint errors (tracked on #13866). All five are `no-unused-vars` violations with no entry in `config/quality/eslint-suppressions.json`, so the frozen suppressions can't absorb them and the gate stays red:

- `src/shared/services/cliRuntime.ts:5` — `execFileSync` imported, never used (#12553)
- `tests/unit/arena-elo-sync-redesign.test.ts:40` — `getArenaEloSyncStatus` destructured, never used (#13446)
- `tests/unit/arena-elo-sync-redesign.test.ts:79` — `makeLeaderboardMap` defined, never called (#13446)
- `tests/unit/compression/compressionAnalyticsWriterFlatRate.test.ts:14` — `rmSync` imported, never used (#13446)
- `tests/unit/waitForServer-slow-first-response.test.mjs:30` — unused `req` handler parameter (#12553)

Each fix is the minimal dead-code removal: drop the dead import or local (and the now-orphaned `ArenaLeaderboardMap` type import), and `_`-prefix the handler parameter per the repo's unused-args convention. No runtime behavior changes.

Verified on the current tip (4d1282be3):

- `npx eslint . --format json --suppressions-location config/quality/eslint-suppressions.json --pass-on-unpruned-suppressions` reports 0 errors, 0 warnings (the release-green lint command)
- affected suites re-run green: arena-elo-sync-redesign 5/5, compressionAnalyticsWriterFlatRate 2/2, waitForServer 3/3, cli-runtime-detection + imports-packaged 20/20
- `check:complexity-ratchets` OK, `check:secrets` 0 findings

The `cli-runtime-extended` suite has one pre-existing failure at the pristine tip (`npm config get prefix` mock assertion) that reproduces identically without this diff — same class as the other 09-18 base-reds #14101 was draining.


xiaoyaner0201:
Confirming from the other side: this PR is the remaining half of the `ESLint errors: 92` hard failure on `release/v3.8.51`.

I traced the other 87 in #14254 — they are a single test file (`tests/unit/translator-openai-to-gemini.test.ts`) whose bulk-suppression entry was voided because #13848 pushed its `no-explicit-any` count from 73 past the frozen cap of 74. Fix in #14255.

With that branch applied, the `No new ESLint warnings` gate reports **exactly the 5 errors this PR removes**, and nothing else:

```
✗ src/shared/services/cliRuntime.ts:5 — 'execFileSync'
✗ tests/unit/arena-elo-sync-redesign.test.ts:40 — 'getArenaEloSyncStatus'
✗ tests/unit/arena-elo-sync-redesign.test.ts:79 — 'makeLeaderboardMap'
✗ tests/unit/compression/compressionAnalyticsWriterFlatRate.test.ts:14 — 'rmSync'
✗ tests/unit/waitForServer-slow-first-response.test.mjs:30 — 'req'
```

87 + 5 = 92. The two PRs touch disjoint files and can land in either order.

Also worth stating plainly for whoever reviews this: the red checks here are **not** caused by this diff. The unit-test failures are the inherited wave-5/6 base-reds catalogued on #13866 (`handleChat` emergency fallback, `i18n pt-BR integrity`, `media-page-client-browser-bundle`, the `pack-artifact` wrapper allowlist, `provider-assets-generic-fallback`) — all in code this PR does not touch. Same for `API Route Typecheck` (`rerankProviderNodes.ts` / `antigravity.ts` TS2677, the latter owned by #14164) and `Merge integrity` (the stale generated `omni-version-manager` skill).

Thanks for writing this one — it had been open a day and I nearly duplicated it before checking.

gonisulaimann:
Closing this one - the maintainer's own drain PR #14164 now carries the same fixes.

I diffed the two branches before closing: cliRuntime.ts, compressionAnalyticsWriterFlatRate.test.ts and waitForServer-slow-first-response.test.mjs come out blob-identical, and the arena-elo change is the same edit (his version just runs the imports through prettier on the way). Everything this PR does is already in #14164, and if this merged first it would hand that PR four conflicts for nothing.

So the 5-error half of the 92 breakdown from #14254 is covered either way.

gonisulaimann:
Sorry for the slow ack, and thanks for the trace — your 87+5 breakdown is what prompted me to actually diff the other candidate instead of leaving this open.

I closed this a few hours after your comment: the maintainer's drain PR #14164 carries all four fixes. Three of them (cliRuntime, compressionAnalyticsWriterFlatRate, waitForServer) are blob-identical to this branch, and the arena-elo one is the same edit with the imports just run through prettier. Merging here first would only have handed #14164 four conflicts.

I also re-verified the compose from the other side: #14255 touches only tests/unit/translator-openai-to-gemini.test.ts plus its changelog fragment, so with #14164 carrying these five there is zero file overlap anywhere in the chain. Either order still works.

PR #14155, opened 2026-09-19
fix(ci): run the env/doc sync gate on checkouts whose paths need URL encoding

gonisulaimann:
The CLI entry of `scripts/check/check-env-doc-sync.mjs` guarded itself with

```js
if (import.meta.url === `file://${process.argv[1]}`) {
```

`process.argv[1]` is absolutized by Node but never URL-encoded, while `import.meta.url` percent-encodes characters like spaces. In a checkout living under a path such as `/Users/dev/My Projects/OmniRoute`, the two strings can never match, so `main()` never ran: `npm run check:env-doc-sync` (and the same step inside `check:docs-all`) printed nothing and exited 0 regardless of the actual contract state. On encoding-free paths like the CI runner's `/home/runner/...` the comparison happens to hold, which is why the gate still catches drift in CI while silently passing on affected local checkouts.

The fix encodes the entry path with `pathToFileURL` before comparing, matching what the other gate scripts already do (`check-api-docs-refs.mjs`, `check-api-typecheck.mjs`, `check-bundle-size.mjs`, `check-circular-deps.mjs`, `check-codeql-ratchet.mjs`). The comparison lives in a small exported `isMainEntry(argv1, moduleUrl)` helper so it can be tested without spawning.

Tests cover the plain absolute path, the percent-encoding case, the negative cases (different entry target, missing `argv[1]`), and a spawned-process test asserting the report banner actually prints. On a checkout path with a space the spawned-CLI test fails before the fix and passes after.


diegosouzapw:

Issue #10702, opened 2026-08-19
[BUG] Vision Bridge fails with "No vision-capable provider connected" despite active settings


Stazyu:
### OmniRoute Version

3.8.50 (next)

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Provider(s) Involved

Command-Code, Opencode

### Model(s) Involved

deepseek-flash-v4, mimo v2.5

### Client Tool

Zoo Code, Playground OmniRoute

### Description

When making a request with an image payload to a text-only model (e.g., `deepseek/deepseek-v4-flash` via `command-code`), Vision Bridge fails to process the image pre-call and logs the following error:

`{"level":40,"time":"...","service":"omniroute","module":"sse","tag":"VISION-BRIDGE","msg":"Failed to get description for image 1: No vision-capable provider connected, cannot process image request"}`

This issue occurs even when Vision Bridge is explicitly enabled in `Dashboard > Settings > Modality Bridge > Vision` and configured with active provider credentials. OmniRoute fails to resolve or route the image description request to the connected vision-capable provider pool.

### Steps to Reproduce

1. Go to Dashboard > Settings > Modality Bridge > Vision.
2. Enable "Vision Bridge" and "Reroute text-only models".
3. Select a Vision model and save settings.
4. Send an OpenAI-compatible `/v1/chat/completions` POST request with an image payload targeting a non-vision model (e.g., `deepseek/deepseek-v4-flash`).
5. Observe the console / SSE logs.

### Expected Behavior

Vision Bridge should intercept the image, send it to the designated vision provider to extract a text description, and then pass that text context to the target model.

### Actual Behavior

Vision Bridge throws `[VISION-BRIDGE] Failed to get description for image 1: No vision-capable provider connected...` and rejects processing the request.

### Test Impact

Needs a new e2e test

### Error Logs / Output

```shell
{"level":40,"time":"2026-08-19T00:53:15.826Z","service":"omniroute","module":"sse","tag":"VISION-BRIDGE","msg":"Failed to get description for image 1: No vision-capable provider connected, cannot process image request"}
{"level":30,"time":"2026-08-19T00:53:15.840Z","service":"omniroute","module":"sse","tag":"ROUTING","msg":"Provider: command-code, Model: deepseek/deepseek-v4-flash"}
```

### Screenshots

[...]

Issue #10052, opened 2026-08-10
fix(ops): /healthz (and other HTTP) times out while event loop is busy (k8s probes flap)


RaviTharuma:
## Summary

On OmniRoute **v3.8.49** (Docker Hub image `diegosouzapw/omniroute:3.8.49`), Kubernetes readiness/liveness probes against **`GET /healthz`** occasionally hit **client timeout / no headers** while the process is still alive. This coincides with CPU-bound work on the main thread (long-context compression, large catalog builds). The probe is the stock `/healthz` body `ok` — not a custom path.

Related (distinct):
- Closed #7849 — long-context compression can exhaust V8 heap (OOM/abort). Here the process often **survives** but **stops answering HTTP** briefly.
- Open #9685 — `GET /v1/models` pegs the event loop (pricing re-parse). Same class of failure for *all* HTTP handlers when the loop is busy.
- Closed #7848 / #6800 — healthz / startup timing history.

## Environment
- OmniRoute: **3.8.49** (official image)
- Runtime: Node (image default), single replica on Kubernetes (Talos)
- Probes: HTTP GET `/healthz` on the main service port (`PORT=20128`)
- Resources (order of magnitude): multi‑GiB heap (`--max-old-space-size` ~12 Gi), multi‑core node

## Steps to reproduce
1. Run stock OmniRoute with Kubernetes startup + readiness + liveness probes on `/healthz` (timeout 2–5s).
2. Cold-start with a non-trivial `storage.sqlite` (tens–hundreds of MB) **or** serve concurrent long-context agent requests that trigger compression/catalog work.
3. Observe kubelet events such as:
   - `Startup probe failed: ... /healthz: context deadline exceeded (Client.Timeout exceeded while awaiting headers)`
   - intermittent readiness failures while the process still has a running main PID

## Expected behavior
`/healthz` (or a dedicated lightweight liveness path) remains responsive within a few hundred ms even when heavy request handling is in progress, so orchestrators can distinguish **busy** from **dead**.

## Actual behavior
While the event loop is busy, probes time out waiting for response headers. That causes:
- temporary **NotReady** / endpoint removal (traffic blackhole on a single replica)
- eventual **liveness kill** if the stall exceeds failure thresholds
- false “gateway down” alerts for operators even when the process is not crashed

## Suggested directions (non-prescriptive)
[...]

PR #14161, opened 2026-09-19
fix(guardrails): cache the vision-bridge no-candidate outcome

cryptiklemur:
## What

`getBestVisionModel()` caches only a successful pick. A `null` outcome is never cached, so an install that sends images with no vision-capable describer re-runs `getVisionCapableModels()` for every image of every request: a full `PROVIDER_MODELS` walk plus a credential lookup per candidate, on the request-serving event loop. It cannot succeed until credentials change, so repeating it buys nothing.

This caches the `null` outcome for 30 seconds, drops it the moment a candidate becomes usable, and keys it the same way as the positive cache so a different `excludedModels` set scans on its own.

## Gating

`OMNIROUTE_VISION_BRIDGE_NEGATIVE_CACHE_MS` sets the TTL. `0` disables the negative cache and restores today's behaviour exactly. The value is also exposed as `noCandidateCacheTtlMs` on `VisionBridgeRouterConfig` for per-call override.

## Measured

A homelab instance running a Claude-only combo with images in most turns. After #14038 the combo-ref path stops triggering this for all-vision combos, but any install that still reaches a `null` pays the full cost per image.

| | before | after |
|---|---|---|
| `/api/monitoring/health` p95 | 20.7s | 688ms |
| 30s timeouts | 22 of 300 | 0 |
| RSS | 7.07 GB | 912 MB |

Both runs are 300 samples on the same endpoint, and the after-run carried about 2.4x the model-call load. The main thread was pegged at 100% CPU in 25-30 second stretches while the scan ran.

## Related

Same class as #13149, where a synchronous scan on the request loop starved health polling. #10702 reports this failure from the correctness side. #10052 covers the symptom of `/healthz` timing out while the event loop is busy.

## Tests

`tests/unit/vision-bridge-negative-cache.test.ts` covers the scan running once, the `0` escape hatch restoring per-call scanning, expiry resuming the scan, recovery once credentials work, and cache-key isolation. The first case fails on `release/v3.8.51` without the source change and passes with it.

## Not included

Memoizing the in-flight `getVisionCapableModels` promise so two concurrent images share one scan. Deduping across calls with differing `deps` needs its own design, so it is out of scope here.


diegosouzapw:
Thanks @cryptiklemur — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #14167, opened 2026-09-19
Ollama Cloud quota: bearer /api/usage instead of settings cookie scrape

### LINKED ISSUE #3709: [Feature] Quota/usage visibility for Ollama Cloud connections: usage fetcher, or local request-counter enforcement of manual provider_plans
## Problem

Ollama Cloud free-tier accounts have hard **hourly + weekly request limits**. When an account crosses its weekly cap, the upstream answers instantly with:

```
429: you (<account>) have reached your weekly usage limit, upgrade for higher limits: https://ollama.com/upgrade
```

OmniRoute (v3.8.22) currently has no way to see this coming:

- `quota_snapshots` is only populated for providers with a registered quota fetcher (antigravity/agy, codex, opencode, deepseek, bailian, crof, ...). There is **no fetcher for ollama-cloud** (`src/lib/quota/*`, `registerGenericQuotaFetchers()`: nothing matches `ollama`).
- `GET /api/providers/quota-windows` consequently exposes no windows for ollama-cloud, so the Dashboard > Limits cutoff modal (`quota_window_thresholds_json` / `quotaPreflight`) cannot enforce anything for it.
- `planRegistry.ts` has no known plan for `ollama-cloud`, and `provider_plans` is empty unless manually seeded.

Result: the router keeps routing to an exhausted account until the upstream 429s, and operators only learn the account state from error logs. On our deployment one weekly-exhausted account took 285 upstream 429 hits in 48h while the sibling accounts on the same provider had zero; account usage IS visible on ollama.com's account page, but nothing in OmniRoute pulls it.

## What we'd like

Either, in order of preference:

1. **Remote usage pull**: an ollama-cloud usage fetcher registered into the quotaPreflight/quotaMonitor system, like `registerOpencodeQuotaFetcher()`. Caveat we already know: ollama.com does not document a public usage API; the usage shown at https://ollama.com/settings is rendered server-side. If there is any authenticated endpoint behind that page (even unofficial), a best-effort fetcher with graceful degradation would be enough. If there is truly no endpoint, option 2 still solves the operational problem.

2. **Local request-counter preflight for plan dimensions without a fetcher**: the Quota Sharing Engine already accepts manual per-connection plans via `provider_plans` / `POST /api/quota/plans` with exactly the right shape:

   ```json
   { "connectionId": "<id>", "provider": "ollama-cloud",
     "dimensions": [
       { "unit": "requests", "window": "hourly", "limit": N },
       { "unit": "requests", "window": "weekly", "limit": M }
     ], "source": "manual" }
   ```

   ...but enforcement (`enforceQuotaShare` / `recordConsumption`) only runs through quota **pools + per-API-key allocations**, i.e. it is designed for sharing one upstream account between downstream keys. There is currently no path that says: "this connection has a manual `requests/weekly` plan; count every routed request against it (OmniRoute already logs them all in `usage_history`) and skip the connection in credential selection once it crosses the cutoff threshold."

   That would give proactive rotation across multi-account providers (we run 6-7 ollama-cloud connections) with zero upstream API support needed, and it would work for every other provider without a fetcher too.

3. **429-message-aware cooldown** as the smallest possible improvement: when the upstream 429 body matches "weekly usage limit", set the connection's `rate_limited_until` to the next weekly reset (or at minimum hours, not the generic transient backoff), so an exhausted account stops being retried on every request.

## Why it matters

Multi-account free-tier providers are a core OmniRoute use case. Today an exhausted ollama account stays in rotation; combined with combo failover this wastes a failover hop (and a queue slot at `requestsPerMinute`/`concurrentRequests` limits) on every request until the weekly reset.

## Our current workaround

A small external script that buckets `usage_history` per connection per rolling window and compares against a hand-maintained limits file. It works, but it lives outside the router, so routing decisions cannot use it.

## Related issues

- #3630 (open) proposes checking the in-memory quota-exhaustion cache during credential selection. Complementary: that covers state learned from prior 429s, while this asks for quota visibility for providers where no fetcher exists at all (and plan-based local counting so the limit is known before the first 429).
- #1952 (closed) implemented weekly/5h quota skip for Codex accounts; this asks for the same idea generalized to providers without a usage API.
- #3027 (closed) fixed ollama-cloud 403 cooldown scoping; this is about quota visibility for the same provider.

NaNomicon:
Closes #3709

## Problem
The Ollama Cloud quota path scraped ollama.com/settings with the user's `__Secure-session` cookie: brittle (breaks on sign-in, redirects, HTML layout changes), and it required users to hand over a full browser-session credential just for a percentage readout.

## Approach
Replaced it with the documented bearer-auth quota endpoint:

- `open-sse/services/opencodeOllamaUsage.ts`: parses `GET https://ollama.com/api/usage` (env `OMNIROUTE_OLLAMA_API_USAGE_URL` to override) with `Authorization: Bearer <apiKey>`, manual redirect handling (redirect → sign-in message), 401/403 → "API key rejected", 10s timeout, non-JSON → clear error. Fraction values (≤1) are ×100 to a percentage, clamped at 100.
- `open-sse/services/usage.ts`: `ollama-cloud` quota dispatch now passes `apiKey || accessToken` only.

## Cookie path retirement
The cookie path is retired entirely (operator decision): no env-cookie fallback. Removed `ollamaCloudUsageCookie` + the 4 cookie env vars (`OMNIROUTE_OLLAMA_CLOUD_USAGE_URL`, `OLLAMA_USAGE_COOKIE`, `OMNIROUTE_OLLAMA_USAGE_COOKIE`, `OLLAMA_CLOUD_USAGE_URL`) everywhere:
- schema (providerSpecificData), request-default sanitizer deletes, dashboard QuotaScrapingFields / EditConnectionModal / quotaScrapingFieldValues
- 65 locale ENVIRONMENT.md copies + English (new `OMNIROUTE_OLLAMA_API_USAGE_URL` row added)
- providerText i18n keys

## Tests
- ollama-cloud-usage.test.ts: rewritten to the API path — 9 tests (no-key message, Bearer header, fraction→percent + clamp, details from limits.monthly.models[], 401/403, redirect, non-JSON)
- 7 affected test files updated, all green locally per-file

## Verification
- Live-tested with a real Ollama API key: `/api/usage` returns `limits.monthly{usage, models[]}`; parsed to `{plan:"Ollama Cloud", quotas.monthly{used, total:100, remaining}}` correctly.
- npm run check:open-sse-typecheck: 0 errors
- Acceptance grep: zero `ollamaCloudUsageCookie|__Secure-session|OLLAMA_USAGE_COOKIE` hits outside an unrelated t3-chat test that legitimately uses `__Secure-session` as a generic web-credential fixture.

## Note on migration
No action expected for existing monitorers: a connection actively tracking Ollama usage already has an API key configured, and the bearer path uses exactly that key. The no-key error message ("Ollama Cloud quota requires an API key. Set the connection's API key to fetch monthly usage.") only surfaces for connections that never had one.


NaNomicon:
Parking for now: without an active Ollama subscription on the test key, the aggregate usage fraction stays 0, so there is nothing meaningful to verify live end-to-end (per-model request counts track, but the dashboard bar stays at 100%). Implementation is complete and verified locally (10/10 unit tests, typecheck 0 errors, bearer call smoke-tested live); branch stays on the fork for revival when a paid key is available. Context: issue #3709.

PR #14170, opened 2026-09-19
fix: resolve OIDC token validation error with trailing slash issuers

### LINKED ISSUE #14119: fix(auth): OIDC callback strips trailing slash from issuer, rejecting ID tokens from IdPs whose iss ends in / (Authentik)

shubhayu-dev:
## Summary

## Related Issues

- Closes #14119
- Related to #

## Validation

- [ x] Change type: provider / routing / UI / i18n / CLI / DB / build-deploy / other
- [ x] Focused tests and category gates from the golden path
- [ x] `npm run lint`
- [ x] Reconciled with the current active release base; focused checks rerun afterward
- [ x] Production-code changes include a new or updated automated test in this PR

## Tests Added Or Updated

## Coverage Notes

## Reviewer Notes


diegosouzapw:
Thanks for chasing this — trailing-slash issuers are a real bug (#14119) and your repro is good. We already have #14144 open for the same issue with a slightly more defensive fix (it accepts both the raw and trimmed issuer plus the discovered one, so it degrades gracefully if discovery is unreachable). Your branch also picked up some unrelated changes along the way — a chatCore.ts hunk implementing Codex quota-header stripping that doesn't match the current function signature, plus some eslint-suppressions.json reverts — which would need to come out before this could merge on its own. I'd like to close this in favor of #14144 rather than ask you to rework it; let me know if you'd rather I keep it open.

shubhayu-dev:
Thanks for taking a look and explaining the context! I completely agree that the #14144's fallback approach makes perfect sense, and my branch definitely picked up some tangled history from a previous local branch.
Please go ahead and close this in favor of #14144. I'm just glad this trailing-slash bug is getting resolved.  Thanks again!!

PR #14171, opened 2026-09-19
fix: add modelsUrl to cheaperinference registry entry

### LINKED ISSUE #14168: feat(providers): add missing modelsUrl for cheaperinference
The cheaperinference registry entry defines the provider's base URL and a static model catalog, but omits modelsUrl:

- Registry entry: open-sse/config/providers/registry/cheaperinference/index.ts
- Base endpoint: https://api.cheaperinference.com/v1/chat/completions
- Expected models endpoint: https://api.cheaperinference.com/v1/models

This prevents the provider from participating in the standard live model-discovery path and leaves the dashboard/API dependent on the bundled static catalog. Please add:

modelsUrl: "https://api.cheaperinference.com/v1/models",

The endpoint is already documented/used by the provider's catalog capture comment (GET /v1/models).

Related but distinct: #14050 concerns enabling Cheaper Inference models, and #9094/#9043 covered initial provider registration.

shubhayu-dev:
## Summary

- Adds `modelsUrl` to the Cheaper Inference provider registry entry. This allows the provider to participate in the standard live model-discovery path (`GET /v1/models`) instead of relying solely on the static catalog.

## Related Issues

- Closes #14168

## Validation

- [x] Change type: provider
- [x] `npm run lint`

## Tests Added Or Updated

- No production logic changed; this is a purely declarative configuration update to a registry entry. Existing provider registry validation tests cover the schema.

## Coverage Notes

- Modifies `open-sse/config/providers/registry/cheaperinference/index.ts`. Since this only adds a static string property to an existing configuration object, coverage is unaffected.

## Reviewer Notes

- The endpoint `https://api.cheaperinference.com/v1/models` was already documented as the catalog source in the file's header comments, but was missing from the actual `RegistryEntry` object. 
- Included the required changelog fragment.


diegosouzapw:
Nice catch — the /v1/models endpoint was already documented in the file's header comment but never wired into the RegistryEntry itself. This looks good to merge as-is, thanks for the clean fix.

Two things before merge (we can do both in this branch): (1) a test — even a one-line registry/discovery-class assertion that `cheaperinference` now resolves as `openai-compat` rather than `static-only` (Hard Rule #8: production change without a test); (2) the live `/v1/models` there returns three `type: "image"` entries (`grok-imagine`, `nano-banana-*`), and the synced list replaces the static one without filtering by `type` — a sync fixture with an image row proving `filterChatSelectableModels` drops it would confirm those ids don't surface as chat models.

shubhayu-dev:
Thanks for the review and the catch on the image models! I will add the discovery-class assertion and set up a sync fixture to verify that filterChatSelectableModels drops the type: "image" entries. I'll push those updates to this branch shortly.

Issue #6222, opened 2026-07-05
feat(gemini): end-to-end tool-calling support on the /v1beta Gemini endpoint (request tools + response functionCall)


diegosouzapw:
## Summary

Tool/function calling is not supported end-to-end when a client speaks the **native Gemini format** against the `/v1beta/models/[...path]` endpoint. Both directions of the Gemini↔OpenAI conversion currently drop tool data, so a Gemini client can neither declare tools nor receive tool calls back.

## Details

**Response side — `tool_calls` dropped.**
`convertOpenAIResponseToGemini` in `open-sse/translator/response/openai-to-gemini-sse.ts` (around L310) maps only `reasoning_content` (→ `thought`) and `content` (→ `text`) into Gemini `parts`. It never emits `parts[].functionCall`, so `message.tool_calls` from the upstream completion are silently lost on the non-streaming path — the Gemini client gets a text-only answer even when the model requested a tool call.

**Request side — `functionDeclarations` dropped.**
`convertGeminiToInternal` in `src/app/api/v1beta/models/[...path]/route.ts` (~L143-171) drops request-side `tools[].functionDeclarations` and any prior `functionCall` / `functionResponse` parts when converting Gemini → OpenAI. So even if the response side were fixed, the model would never receive the tool definitions.

Because both sides drop tool data, this is a full feature gap, not a one-line response-projection fix (the Claude/OpenAI-Responses client paths already project tool calls correctly via `responseTranslator.ts` + `convertOpenAINonStreamingToClaude`; only the Gemini-endpoint path is affected).

## Scope

- [ ] Request: translate `body.tools[].functionDeclarations` → OpenAI `tools`, and prior `functionCall` / `functionResponse` parts → assistant `tool_calls` / `tool` messages, in `convertGeminiToInternal`.
- [ ] Response (non-streaming): emit `parts[].functionCall { name, args }` from `message.tool_calls` in `convertOpenAIResponseToGemini`; set the correct `finishReason`.
- [ ] Response (streaming): verify `openAIChunkToGeminiChunk` emits `functionCall` deltas (symmetry).
- [ ] Regression tests under `tests/unit/` (node:test): a Gemini-format tool-call round-trip (request declares a tool → response returns `functionCall`).

## Notes

[...]

PR #14173, opened 2026-09-19
fix(v1beta): keep inlineData and parts sent next to a functionResponse

datrixlab:
## Summary

- `convertGeminiToInternal()` (the `/v1beta/models/{m}:generateContent` route) reads only `text`, `functionCall` and `functionResponse` parts. Two things are lost that the Gemini request translator (`gemini-to-openai.ts`) keeps:
  - **`inlineData`**: an image, PDF or audio sent inline never reaches the provider. An image-only turn becomes `{ role: "user", content: "" }`, so the model is asked about a picture it never sees.
  - **Parts next to a `functionResponse`**: a content with a response returns its tool messages and drops every other part. gemini-cli sends a binary file a tool read (`read_file` on a PNG or PDF) exactly that way: the functionResponse, then the file as `inlineData` in the same content. The translator has split such contents since #6376 (`splitCoLocatedFunctionResponses()`).

Measured with the real module, then through `translateRequest` to a Gemini upstream:

| input content (role `user`) | before | after |
| --- | --- | --- |
| `text` + `inlineData` image | `content: "What is in this picture?"` | `[text, image_url data URL]`, same as `geminiToOpenAIRequest` |
| `inlineData` only (PDF) | `content: ""` | `[image_url data:application/pdf…]` |
| `functionResponse` + `inlineData` (gemini-cli `read_file`) | tool message only | tool message, then a user message with the image |
| `functionResponse` + `text` | tool message only | tool message, then the text as a user message |
| upstream Gemini request for the first row | no `inlineData` part | `inlineData { mimeType: "image/png" }` |

- Fix:
  - `inlineData` / `inline_data` in a non-model turn becomes an `image_url` data URL, the mapping the translator uses. Text-only contents keep a plain string.
  - A content with `functionResponse` parts is split into the responses and the rest **before** conversion, as the translator does, so `beginContent` sees each piece. A leftover piece with nothing in it (an empty text part) adds no message.
  - Model turns keep their text only. Chat Completions accepts image parts in user content, and replaying an image-model turn as assistant `image_url` content would fail on OpenAI targets. The base code did not read `inlineData` in any turn, so model turns convert exactly as before.

## Related Issues

- Follows #6376 (co-located `functionResponse` parts in the Gemini translator) and #6222 (tool calling on `/v1beta`).

## Validation

- [x] Change type: provider (`/v1beta` route)
- [x] Focused tests: the new file plus `v1beta-gemini-tool-calling-6222`, `gemini-tool-result-without-id`, `v1beta-models-route`, `gemini-to-openai-function-response` (39/39)
- [x] Wide run: all translator, gemini, v1beta and antigravity unit test files (232 files, 1740/1744). The four failures are `agy-antigravity-shared-catalog-12724`, which fails the same way on a clean `release/v3.8.51` on Windows.
- [x] `tsc` on `src/app/api/v1beta/models/**` (temp tsconfig extending the root one): no errors in the changed file; the six it reports are in other files and appear on a clean `release/v3.8.51` too. `eslint` clean; `prettier --check` on the changed lines

## Tests Added Or Updated

- `tests/unit/v1beta-gemini-inline-data.test.ts` (new):
  - text + image gives `[text, image_url]`, deep-equal to what `geminiToOpenAIRequest` produces for the same input;
  - an image-only turn (snake_case `inline_data`, PDF) is not an empty message;
  - gemini-cli's `read_file` shape: the tool message answers the id-less call, then the image follows as a user message;
  - text next to a response follows the tool message as a user message;
  - a response next to a call in the same content still answers the earlier call, and the new call is kept;
  - a functionResponse alone still yields only its tool message;
  - an image in a model turn keeps the assistant content as the text string;
  - text-only contents keep a plain string;
  - through `translateRequest(OPENAI → GEMINI)`, the image reaches the upstream request as `inlineData`.

With `convertGeminiToInternal.ts` reverted, the six behaviour tests fail and the three guards (response alone, model-turn image, text only) pass.

## Coverage Notes

- `splitFunctionResponses()` is exercised with no response, a response alone, a response with text, with `inlineData`, and with a `functionCall`.

## Reviewer Notes

- `thought` parts are still not split into `reasoning_content` on this route, as before; the translator does that in `convertGeminiContentWithReasoning()`. I kept this PR to the parts that were dropped.


diegosouzapw:
Thanks @datrixlab — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #14179, opened 2026-09-19
fix(build): combos page drops server-only import

maxmad64bis:
⚠️ base-red inherited: #13866

## Summary

The combos dashboard page broke the production client build with missing server-only modules. The page imported a resolver from the server-side model service, which lazily loads the database layer that the client bundler then follows. The page now resolves provider prefixes through the client-safe alias map, and the client-bundle guard follows first-party dynamic imports.

## Related Issues

- Related to #13474 — same family of production build broken by a server-side import reaching a client component.
- Related to #13704 — open change touching the same client-bundle guard; merge coordination only, no code reused.

## Validation

- [x] Change type: build-deploy
- [x] `npm run lint` — ESLint on the touched files is clean; the full run is red on the base (#13866)
- [x] Reconciled with the current active release base; focused checks rerun afterward — `rev-list --count HEAD..upstream/release/v3.8.51` = 0, focused checks rerun at this head

## Tests Added Or Updated

- `tests/unit/client-bundle-no-server-only-10692.test.ts` — follows first-party dynamic import edges and covers the specifier extractor; fails before the fix on the combos page trail and passes after.
- `tests/unit/combos-allowed-provider-resolution.test.ts` — proves alias-map resolution matches the previous canonical provider field for known, unknown, and non-string inputs.

## Coverage Notes

- Two-line call-site swap plus guard widening; coverage from the widened guard and the equivalence test; no coverage-sensitive branches added.

## Reviewer Notes

- The previous guard ignored first-party dynamic import edges, and the header claiming they cannot break a bundle edge was incorrect for the bundler in use and has been corrected. Behavior is identical for every input the page can pass because the old code read only the provider field, which equals the alias-map result, with unknown prefixes falling through to identity in both paths.
- Only literal dynamic-import specifiers are followed while variable and package specifiers are ignored, mirroring the existing static handling, with runtime discovery and no hardcoded module list. Potential merge conflict with #13704 on the same guard file; this change will be rebased if that change merges first.


diegosouzapw:

Issue #11503, opened 2026-08-25
fix(backend): catalog hygiene — FITNESS_TABLE ranks retired models at 0.98 and BUILT_IN_ALIASES forwards retired Claude ids to other retired ids


MumuTW:
## Why a new issue

#8777 was closed by #8790 (`maxContextWindow`), which does not touch `FITNESS_TABLE`; #8778's maintainer asked for the catalog-hygiene work to be filed separately from the Artificial Analysis question. This is that issue: the **static fitness table** and the **deprecation alias table**, both hand-maintained, both now pointing at models the vendors have retired. Full audits with reproduction are in #8777's last two comments; the numbers below are from `release/v3.8.51` and first-party lifecycle pages (Anthropic, OpenAI, Google, xAI, Groq), not models.dev (whose `status` flagged 0 of the 22 retired ids).

## 1. `FITNESS_TABLE` (`open-sse/services/autoCombo/taskFitness.ts:24`) ranks retired models above live ones

Resolved through the live chain (`getTaskFitnessWithSource`, `taskType="coding"`):

| id | vendor status | coding score | via |
|---|---|---|---|
| `gpt-5-codex` … `gpt-5.2-codex` (6 ids) | OpenAI shut down 2026-07-23 | **0.98** | `fitness_table` (`codex` pattern — the table's highest value) |
| `o3-mini`, `o3-2025-04-16` | OpenAI shutdown 2026-10/12 | 0.95 | `fitness_table` |
| `chatgpt-4o-latest` | shut down 2026-02-17 | 0.90 | `fitness_table` |
| `gpt-4-turbo`, `o4-mini` | shutdown 2026-10-23 | 0.88 | `fitness_table` |
| `claude-fable-5-thinking-max`, `gpt-5.6-sol-xhigh` (live flagships) | active | **0.50** | wildcard |

Recognition by the table, not availability, decides rank: a dead model the table knows beats a live one it doesn't by ~0.5. 13 retired ids and 9 with shutdown dates inside four months are still routable. #11492 fixes the variant side (effort/alias ids now inherit their base model's DB score) but deliberately does not touch this table's contents.

## 2. `BUILT_IN_ALIASES` (`open-sse/services/modelDeprecation.ts:13`) forwards retired ids to other retired ids

Applied on every request (`modelLifecyclePolicy.ts:43` rewrites `body.model`), so a stale target is a guaranteed 404:

| source | → target | target status |
|---|---|---|
[...]

PR #14183, opened 2026-09-19
fix(providers): dispatch opencode-go qwen3.8-max unchanged

### LINKED ISSUE #14181: fix(opencode-go): stop rewriting qwen3.8-max to preview
### OmniRoute Version

3.8.51

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

_No response_

### Node.js Version

_No response_

### Provider(s) Involved

OpenCode Go

### Model(s) Involved

opencode-go/qwen3.8-max and qwen3.8-max-preview

### Client Tool

OpenCode

### Description

OpenCode Go dynamically imports the current upstream model `opencode-go/qwen3.8-max`, and the dashboard displays it as an IMPORTED model. However, OmniRoute v3.8.51 applies the stale global deprecation mapping `qwen3.8-max` → `qwen3.8-max-preview` before dispatch. OpenCode Go then rejects the rewritten model with `401 Model qwen3.8-max-preview is not supported`. The static model specification also incorrectly treats the GA model as an alias of the Preview model.

### Steps to Reproduce

1. Use OmniRoute v3.8.51 with an active OpenCode Go connection.
2. Allow OpenCode Go to dynamically import the upstream `qwen3.8-max` model.
3. Send a request using `opencode-go/qwen3.8-max`.
4. Observe that the model is rewritten to `qwen3.8-max-preview` before dispatch.
5. Observe the upstream 401 response.

### Expected Behavior

Requests targeting OpenCode Go's current GA model should preserve the exact model ID and dispatch `qwen3.8-max` unchanged. Providers that still expose `qwen3.8-max-preview`, including Qoder and Bailian, should continue using the Preview ID without regressions.

### Actual Behavior

`opencode-go/qwen3.8-max` is rewritten to `qwen3.8-max-preview`, causing OpenCode Go to return `401 Model qwen3.8-max-preview is not supported`. The model-spec table also declares `qwen3.8-max` as an alias of the Preview entry.

### Test Impact

Unsure

### Error Logs / Output

```shell
401 Model qwen3.8-max-preview is not supported
```

### Screenshots

_No response_

### Additional Context

The fix should be provider-safe and narrowly scoped: remove the obsolete global deprecation mapping; remove the GA ID from the Preview aliases; add a separate GA model spec; register `qwen3.8-max` in the OpenCode Go registry with the existing Qwen routing requirements (`targetFormat: "claude"`, reasoning support, and the correct vision setting); and verify whether dynamically imported provider models are included by `hasKnownProviderModel()`.

Do not remove `qwen3.8-max-preview` from Qoder, Bailian, or other providers that still legitimately expose it. Search all occurrences before changing them and distinguish provider entries from documentation and obsolete alias/deprecation mappings.

### Validation Plan

Add regression coverage for `resolveModelAlias("qwen3.8-max", "opencode-go") === "qwen3.8-max"` and `resolveModelAlias("qwen3.8-max-preview", "qoder") === "qwen3.8-max-preview"`. Confirm that `qwen3.8-max` has its own model spec, the Preview spec no longer aliases the GA ID, and Preview providers remain registered. Run the relevant unit tests and report the exact passing test commands.

gonisulaimann:
OpenCode Go ships the GA `qwen3.8-max` upstream, but the global deprecation map in `modelDeprecation.ts` still forwards it to `qwen3.8-max-preview`. That mapping predates the GA model (its own comment says to drop it once Alibaba ships a distinct `qwen3.8-max` — that condition is now met), so every request for `opencode-go/qwen3.8-max` was rewritten and rejected upstream with `401 Model qwen3.8-max-preview is not supported`.

What changed:

- `open-sse/services/modelDeprecation.ts` — removed the stale `qwen3.8-max` → `qwen3.8-max-preview` mapping.
- `src/shared/constants/modelSpecs.ts` — `qwen3.8-max` gets its own spec with the same 1M-context envelope, so the context-preflight coverage the rewrite used to provide is preserved through the spec itself; the preview spec no longer aliases the GA id.
- `open-sse/config/providers/registry/opencode/go/index.ts` — registered the GA row with the qwen routing flags (`targetFormat: "claude"`, `supportsReasoning: true`, `supportsVision: false`) so `hasKnownProviderModel` also holds for the direct path.

Providers that legitimately expose the preview id are untouched: `qwen3.8-max-preview` still resolves and dispatches on qoder and bailian-coding-plan.

Tests (`tests/unit/qwen3-8-max-ga-14181.test.ts`, written first — all four failed on the base tip):

- `resolveModelAlias("qwen3.8-max")` and with the `opencode-go` provider both return the id unchanged.
- `qwen3.8-max-preview` still resolves on qoder and bailian-coding-plan, and both registries keep the preview row.
- opencode-go declares `qwen3.8-max` with the routing flags above.
- `qwen3.8-max` has its own `MODEL_SPECS` entry and the preview spec no longer aliases the GA id.

Results:

- `node --import tsx/esm --test tests/unit/qwen3-8-max-ga-14181.test.ts` — 4/4 pass.
- Related suites (`model-deprecation`, `model-deprecation-aliases-11503`, `resolve-model-alias-index-8697`, `bailian-coding-plan-provider`, `qwen-cloud-token-plan-discovery`, `model-capabilities-registry`) — 135/135 pass.
- `tsc --noEmit` clean on the touched files; eslint clean; prettier clean.

Fixes #14181


gonisulaimann:
Closing in favor of #14242, which fixes the same issue and does it more correctly.

When I re-verified the registries at the current tip, the ground truth is: qoder and bailian-coding-plan serve only qwen3.8-max-preview, while qwen-cloud-token-plan, alibaba, qwen-cloud, kilocode, clinepass and xkiro list the GA id. My branch deleted the global mapping outright - which fixes opencode-go but would regress bare-id qwen3.8-max requests against qoder and bailian. #14242 keeps the rewrite for the preview-only providers and lets the provider-aware exemption from #11503 pass the GA id through on opencode-go, which is the mechanism the repo already built for exactly this. Its test suite also pins the qoder rewrite staying intact, which mine asserted away.

The parts of my branch that were right (the GA MODEL_SPECS row, the opencode-go registry entry) are both in #14242, so nothing is lost here.

PR #14184, opened 2026-09-19
fix(proxy): probe http proxies on their default port 80

### LINKED ISSUE #14157: fix(resilience): http proxy on port 80 always fails the fast-fail probe (probe goes to 8080)
http proxies on port 80 always fail with `[Proxy Fast-Fail] Proxy unreachable`, even though the proxy is fine and `POST /api/settings/proxy/test` passes

## what happens

my setup: webshare rotating endpoint `p.webshare.io:80` (type `http`), registered in the proxy registry and assigned to one antigravity connection (scope `account`). every request through that connection dies instantly:

```
[Proxy Fast-Fail] Proxy unreachable: http://p.webshare.io:80
```

and then the connection goes into the ~5 min cooldown, so follow-up requests return `ALL_TARGETS_SKIPPED` even after i unassign the proxy

same host, same container, same credentials:
- `curl -x http://user:pass@p.webshare.io:80 https://cloudcode-pa.googleapis.com/` works
- a raw `net.createConnection({host:"p.webshare.io",port:80})` from inside the container connects in ~20ms
- `POST /api/settings/proxy/test` returns `{"success":true,"latencyMs":1387}`

changing nothing but the port to `3128` (webshare listens there too) makes everything work

## cause

`isProxyReachable` in `src/lib/proxyHealth.ts` does

```ts
url = new URL(proxyUrl);
const port = parseInt(url.port || defaultPortForScheme(url.protocol), 10);
```

the WHATWG URL parser drops the port when it equals the scheme default, so

```js
new URL("http://u:p@p.webshare.io:80").port   // ""
new URL("http://u:p@p.webshare.io:3128").port // "3128"
```

and the fallback is wrong for http:

```ts
case "http":
default:
  return "8080";
```

so the tcp probe goes to `p.webshare.io:8080`, nothing listens there, the probe resolves `false`, and the non-blocking race in `open-sse/utils/proxyFetch.ts` aborts the in-flight request with `PROXY_UNREACHABLE` (503)

the log label still says `:80` because it comes from a different helper, which makes this pretty confusing to debug

`https` on 443 is fine (fallback is `"443"`), socks5 keeps its port because it is not a special scheme, so only `http` + port 80 is broken

## fix

`defaultPortForScheme("http")` should return `"80"`

`8080` is a common proxy port but it is not what an empty `url.port` means for an `http:` url

## environment

- omniroute 3.8.51 (`diegosouzapw/omniroute:next`, docker), node v26
- still present on `main` today (`src/lib/proxyHealth.ts`, `defaultPortForScheme`)

gonisulaimann:
`isProxyReachable` parses the proxy URL and falls back to `defaultPortForScheme` when `url.port` is empty. The WHATWG URL parser drops the port when it equals the scheme default, so `http://user:pass@p.webshare.io:80` produced `port === ""` — and the `http` case handed back **8080** instead of 80. The probe then missed, every request through that connection died with `Proxy Fast-Fail`, and the connection went into cooldown even though the proxy was reachable (the dashboard's `POST /api/settings/proxy/test` uses a different path, which is why it reported the proxy as healthy).

The `https` (443) and socks5 (1080) defaults were already correct; only the `http` case was wrong. Proxies with explicit non-default ports never hit the fallback and are unaffected.

Tests (`tests/unit/proxy-health-default-port-14157.test.ts`, written first — both failed on the base):

- an `http://…:80` proxy URL probes port 80 (asserted via the existing `__setProxyHealthTcpCheckForTesting` seam).
- explicit ports (3128, 8443) still probe as configured, and schemeless `http`/`https` URLs resolve to 80/443.

Results:

- focused test file — 2/2 pass (red → green confirmed).
- proxy-health and fast-fail suites (`t14-proxy-fast-fail`, `proxy-health-6246`, `proxy-health-ipv6`, `proxy-health-blocked-outcome`, `proxy-health-blocked-streak-sweep`, `proxy-health-decide-action-6246`, `proxy-health-egress-line`, `proxy-health-refusal-memory`, `probe-10720-proxy-password-only-auth`, `proxy-probe-target`) — 62/62 pass.
- registry/connection suites touching explicit 8080 configs (`8385-perkey-proxy-global-toggle`, `9232-purge-proxy-assignments-on-delete`, `proxy-connection-test`, `proxy-registry-manager`) — 73/73 pass.
- `tsc --noEmit` clean, eslint clean, prettier clean.

Fixes #14157


diegosouzapw:

PR #14185, opened 2026-09-19
fix(api): detect /v1beta ingress bodies as openai, not claude

### LINKED ISSUE #14165: fix(api): /v1beta returns an empty body when generationConfig.maxOutputTokens is set (detectFormat guesses claude), agy loops forever
any gemini-format request that sets `generationConfig.maxOutputTokens` gets an empty reply from the `/v1beta` ingress, which makes the antigravity cli (agy) and anything built on the google genai sdk loop forever

## repro (one curl)

```bash
curl -N "http://HOST:20128/v1beta/models/gemini-3.8-flash:streamGenerateContent?alt=sse" \
  -H "x-goog-api-key: $KEY" -H 'content-type: application/json' \
  -d '{"contents":[{"role":"user","parts":[{"text":"Reply with exactly: OK-1"}]}],"generationConfig":{"maxOutputTokens":1024}}'
```

- streaming: `200`, `content-type: text/event-stream`, **0 bytes** of body
- same body on `:generateContent` (non-stream): `200` but the body is anthropic shaped, `{"id":"chatcmpl-...","type":"message","role":"assistant",...}`, not a gemini `candidates` response
- drop `maxOutputTokens` and both work (proper gemini sse chunks with `finishReason: STOP`)
- any value does it, i tried 1024, 8192, 32768, 65535, 65536
- the call log shows the model did answer correctly (`finish_reason: stop`, right text), the reply is just lost on the way back

## cause

`src/app/api/v1beta/models/[...path]/route.ts` converts the gemini body with `convertGeminiToInternal` and re-enters `handleChat` with an openai-ish body:

```ts
max_tokens: geminiBody.generationConfig?.maxOutputTokens,
```

then `detectFormat` in `open-sse/services/provider.ts` looks at that body and hits this heuristic:

```ts
if (body.max_tokens && !body.stream_options && !body.response_format) {
  return "claude";
}
```

so the source format becomes `claude`, the pipeline answers in anthropic sse, and `transformOpenAISSEToGeminiSSE` finds no openai chunks to translate and emits nothing

the route already knows the request is gemini→openai, so it should not let `handleChat` guess again from the body shape

passing the source format explicitly (or at least marking the converted body as openai) fixes it

## real world impact

my setup: antigravity cli 1.2.4 in gemini api key mode (`modelProvider: "gemini"` in settings.json, `GEMINI_API_KEY` = gateway key, `GOOGLE_GEMINI_BASE_URL` = gateway root)

agy always sends `maxOutputTokens: 65536`, so every turn comes back empty, agy then appends

```
Error: model output error: model output must contain either output text or tool calls, these cannot both be empty, please try again
```

and asks again, forever

one `agy -p` call produced 47 upstream requests in 100 seconds, all `200`, all billed to the antigravity accounts, until i killed it

from the client side it just looks like a hang

## workaround i am running

patched the compiled route so the converted body carries an openai-only marker, which flips `detectFormat` back to `openai`:

```js
// before
stream:c,max_tokens:a.generationConfig?.maxOutputTokens,
// after
stream:c,n:1,max_tokens:a.generationConfig?.maxOutputTokens,
```

with that, the same curl returns 353 bytes of valid gemini sse, and agy works end to end (plain reply in 3s, tool calls fine, no loop)

## environment

- omniroute 3.8.51 (`diegosouzapw/omniroute:next`, docker)
- both snippets above are still on `main` today

gonisulaimann:
Any Gemini-format request with `generationConfig.maxOutputTokens` got a broken reply from `/v1beta`: 200 SSE with zero bytes when streaming, and an anthropic-shaped body when not — which made the antigravity CLI and anything on the Google GenAI SDK loop forever.

Root cause: the `/v1beta` route converts the body to OpenAI chat format (`convertGeminiToInternal`) and re-enters `handleChat`, but the request URL keeps its `/v1beta` path. `detectFormatFromEndpoint` had no `/v1beta` branch, so detection fell through to body heuristics — and the converted body (`messages` + `max_tokens`) trips `detectFormat`'s "`max_tokens` is a required Claude field" heuristic, returning `claude`. `sourceFormat` then drives response shaping: the non-streaming path produced a claude body the route's OpenAI→Gemini converter doesn't understand (silently dropped), and the streaming path produced claude SSE events the OpenAI→Gemini SSE translator never converts (zero bytes).

Fix: `detectFormatFromEndpoint` now treats a `/v1beta` path as `openai` chat — matching the ingress's own contract — unless the body still carries the raw Gemini `contents` envelope, so client-raw-request contexts that pass unconverted Gemini bodies keep the existing `gemini` detection.

Tests (`tests/unit/v1beta-format-detection-14165.test.ts`, written first — the two detection tests failed on the base):

- a converted body (`messages` + `max_tokens`) on a `/v1beta/…:streamGenerateContent` path detects as `openai`, not `claude`.
- `detectFormatFromUrl` with the full ingress URL (including query string) detects as `openai`.
- a raw Gemini body (`contents`) on a `/v1beta` path still detects as `gemini`.
- `convertGeminiToInternal` emits the shape the branch relies on (`messages`, `max_tokens`, no `contents`).

Results:

- focused test file — 4/4 pass (red → green confirmed on the two detection tests).
- detection suites (`provider-detect-format-endpoint`, `antigravity-format-detection`, `translator-format-detection-2949`, `plan3-p0`) — 58/58 pass.
- `tsc --noEmit` clean, eslint clean, prettier clean.

Fixes #14165


diegosouzapw:
Verified the placement of the new /v1beta branch against the live provider.ts source — it sits correctly ahead of the max_tokens fallback and the raw-gemini guard preserves client-raw-request behavior. The failing checks on this PR are the same inherited base-red as the rest of your PRs this week (issue #13866), not caused by this change.

diegosouzapw:

Issue #3974, opened 2026-06-16
[BUG] tool-search-tool-2025-10-19 beta dropped on Claude OAuth path → 400 "Tool reference not found" with ENABLE_TOOL_SEARCH


huohua-dev:
### OmniRoute Version

3.8.26 (latest at time of report; root cause also confirmed on current `main` @ `ca1e17f`). Affected host runs the Electron desktop build on macOS.

### Installation Method

Electron desktop app

### Operating System

macOS

### Provider(s) Involved

Claude (`cc`) — `authType: oauth`, claude.ai subscription backend (`api.anthropic.com`).

### Model(s) Involved

`claude-opus-4-8` (affects all Claude models on the OAuth path).

### Client Tool

Claude Code 2.1.177, with `ENABLE_TOOL_SEARCH=true` (or `auto`, when MCP tool definitions exceed the ~10% context threshold).

### Description

When Claude Code's tool-search feature is active it sends tools with `defer_loading: true`, a `tool_search_tool_*` tool, and negotiates the header `anthropic-beta: tool-search-tool-2025-10-19`. OmniRoute strips that beta flag on **both** Claude code paths, so the upstream claude.ai backend receives deferred tools / `tool_reference` blocks **without** the enabling beta and rejects the request with:

```
400 Tool reference '<tool_name>' not found in available tools
```

Every MCP tool is affected — reproduced with both `mcp__codegraph__*` and `mcp__github__*`. The upstream (claude.ai OAuth) **does** support this beta: stock Claude Code talking directly to `api.anthropic.com` works fine. The breakage is purely OmniRoute discarding the client-negotiated header.

This is the beta-flag analogue of #2766 / #3033 (where the `tool_search` *tool* was dropped on Codex/Responses paths); here it is the enabling *beta flag* that is dropped on the **Claude** path.

### Root Cause

Two independent code paths both drop the flag:

**1. `default` executor path** — the one the `claude` provider actually uses (`providerRegistry.ts` `claude` entry: `executor: "default"`).
- `default.ts buildHeaders()` sets `anthropic-beta` from the hardcoded registry header `ANTHROPIC_BETA_CLAUDE_OAUTH` (`config/anthropicHeaders.ts`), which does **not** contain `tool-search-tool-2025-10-19` (verified: the slice-built OAUTH set has no `tool-search` entry).
[...]

Issue #3415, opened 2026-06-08
[BUG] Forced anthropic-beta rewrite on opus models (interleaved-thinking / advanced-tool-use) corrupts tool_use stream; Claude Code fails with "tool call could not be parsed (retry also failed)"


Forcerecon:
### OmniRoute Version

3.8.9

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24.04

### Provider(s) Involved

Claude

### Model(s) Involved

claude-opus-4-8

### Client Tool

Claude Code (CLI 2.1.161 / 2.1.162)

### Description

For Claude Code traffic to `claude-opus-4-8` (claude→claude, "Standard passthrough mode"), OmniRoute **discards the client's inbound `anthropic-beta` header and rebuilds it from scratch**, unconditionally forcing beta flags the client never requested — specifically `interleaved-thinking-2025-05-14`, plus `advanced-tool-use-2025-11-20` and `effort-2025-11-24` for opus + tools + system requests.

Under this forced interleaved-thinking semantics, a subset of opus-4-8 turns return **HTTP 200** with `stop_reason: "tool_use"` but a **malformed / incomplete `tool_use.input` JSON**. Claude Code cannot finalize the tool call, auto-retries once, hits the same condition, and aborts the turn with:

```
The model's tool call could not be parsed (retry also failed).
```

These turns are slow (single request **27s–362s**) with small output (527–3258 tokens), i.e. long thinking followed by a broken/stalled tool_use, not a token-limit truncation.

Net effect for the operator: long-running opus-4-8 sessions in Claude Code intermittently die mid-task. The same client, pointed at a proxy that forwards the client's own `anthropic-beta` header verbatim, never reproduces it.

Note: effort and thinking are *not* escalated when the client already sends them (guarded by `void 0 === …`). The only forced mutation is the **beta header rewrite**.

### Steps to Reproduce

1. Deploy OmniRoute 3.8.9 (Docker), claude→claude passthrough, with a working Anthropic provider connection.
2. Point a Claude Code CLI (2.1.16x) client at OmniRoute and select model `claude-opus-4-8`.
3. Run a tool-heavy session: many tools defined (~25), a system prompt present, adaptive thinking enabled (Claude Code defaults), with prompts that induce long thinking.
4. Send a series of turns.
[...]

Issue #2454, opened 2026-05-21
fix: Sanitize Claude Thinking Block Signatures & Beta Headers in Proxy Passthrough


havockdev:
# fix: Sanitize Claude Thinking Block Signatures & Beta Headers in Proxy Passthrough

## Summary

When OmniRoute proxies Claude API requests using OAuth tokens in passthrough mode, **two distinct 400 errors** occur depending on the scenario:

1. **Starting a new session with Haiku**: The proxy sends incompatible beta headers (`context-1m-2025-08-07`) that Haiku with OAuth authentication rejects
2. **Switching models mid-session** (e.g., Sonnet → Haiku): Historical `thinking` blocks carry cryptographic signatures bound to the **original auth token** — the proxy uses a different token, so Anthropic rejects the mismatched signatures

Both errors are **specific to the Claude passthrough path** (`sourceFormat === CLAUDE && targetFormat === CLAUDE`) and only manifest when using OAuth-based authentication through the proxy.

---

## Error 1 — Haiku Beta Header Incompatibility

```
[400]: This authentication style is incompatible with the long context beta header.
```

**When it happens**: User starts a **new session** selecting `claude-haiku-4-5-20251001` as the model. On the very first message, the API returns 400 immediately.

**What triggers it**: The function `selectBetaFlags()` in `claudeIdentity.ts` was applying the **same set of beta headers** to all Claude models indiscriminately. For Haiku with OAuth authentication, the `context-1m-2025-08-07` header is invalid because Haiku does not support 1M context and Anthropic explicitly rejects this combination.

## Error 2 — Thinking Block Invalid Signature

```
[400]: messages.11.content.0: Invalid `signature` in `thinking` block
```

**When it happens**: User starts a session with a **thinking-capable model** (Sonnet or Opus), sends several messages (building up thinking blocks in the conversation history), then **switches to a different model** (e.g., Haiku). The first message after switching fails with 400.

**What triggers it**: Claude's `thinking` blocks contain a cryptographic `signature` field that is **bound to the auth token that generated them**. When the proxy forwards the request using its own auth token (different from the client's original token), Anthropic's API validates the signature against the proxy's token and rejects it as invalid.

---

### Phase 1 — Intercepting Real Claude Desktop Behavior

To understand the correct behavior, we intercepted the **real Claude Desktop (v2.1.142)** HTTP requests for both Sonnet and Haiku using a local HTTP proxy capture. This gave us a ground-truth comparison:

[...]

PR #14186, opened 2026-09-19
fix(claude): forward the auto mode safeguards betas

cryptiklemur:
## What

Claude Code is changing auto mode so the server runs its permission checks inside the session's own requests, and doesn't bill for them. Through OmniRoute, every auto mode session shows this instead:

> We're changing auto mode to no longer charge for classifier requests in Claude Code. However, this session isn't eligible because your requests go through <gateway>, which isn't compatible with this update.

Claude Code asks for the checks with a `safeguards` body field plus two betas, `afk-mode-2026-01-31` and `dangerous-tool-use-2026-09-03`. The body field already passes through. The betas aren't in `FORWARDABLE_CLIENT_BETAS`, so `mergeClientAnthropicBeta` drops them on both the default executor and the Claude Code / OAuth path. The server never sees the request, and Claude Code falls back to its own separately billed classifier calls.

This adds both betas to the allowlist. Nothing else changes, and the allowlist still guards against #3415 and #2454.

Anthropic's guide for gateways: https://code.claude.com/docs/en/auto-mode-classifier-billing

## How I found the betas

They aren't in the public beta reference yet. I pointed Claude Code 2.1.278 at a local listener that logged headers and body keys, and ran one auto mode prompt. It sent `safeguards` in the body, and these were the only two betas tied to auto mode that OmniRoute doesn't already know.

## Measured

This ran on a homelab instance on the Claude OAuth path. The build was `1a74ab908` with only this diff applied.

| | before | after |
|---|---|---|
| "isn't eligible" notice | shown in every auto mode session | not shown |
| `/status` Auto mode server | `Disabled` after the fallback, per the docs | `Enabled` |

I also checked the response side. `safeguard_results` and `tool_use` ids already pass through unchanged on the claude-to-claude streaming and non-streaming paths.

## Tests

`tests/unit/claude-auto-mode-safeguards-beta.test.ts` covers four things:

- `mergeClientAnthropicBeta` forwards both betas once
- `selectBetaFlags` plus the merge keeps them on the Claude Code path
- `DefaultExecutor("claude")` sends them in its outbound headers
- `transformRequest` leaves the `safeguards` body field alone

The three beta cases fail on `release/v3.8.51` without the source change and pass with it. The body case passes either way. It guards against a future strip. The existing beta suites still pass: #3415, #3974, context-1m, and thinking-binding.

## Not included

`parseSSEToClaudeResponse` rebuilds a non-streaming Claude response from a fixed set of fields. If a Claude upstream is forced to stream for a `stream: false` client, it would drop `safeguard_results`. The `claude` and `anthropic` providers don't force streaming, so I left it alone.


diegosouzapw:
Thanks for tracking this down with a header capture against Claude Code 2.1.278 — that's solid diagnostic work. #14312 landed just after yours and covers the same dangerous-tool-use-2026-09-03 forwarding plus a second real bug (anthropic-compatible-* providers were dropping the client's beta entirely, not just filtering it), so I'm going to close this one in favor of that PR and credit your finding in the discussion there. One open question for you: afk-mode-2026-01-31 isn't in #14312 — its author deliberately left it out, and it's also the beta an older regression test (#2454) explicitly says shouldn't appear in our own default set. If your capture still shows it's required for the eligibility check specifically (not just present in the request), a small follow-up PR re-adding just that token with the capture evidence would be very welcome.

PR #14193, opened 2026-09-19
docs(cli): fix invalid config and models commands

### LINKED ISSUE #13997: docs(cli): README and CLI-TOOLS document two commands that do not exist (`config show`, `models list`)

prabhtheone:
## Summary

Fixes the two invalid CLI examples documented in #13997.

- Replace `omniroute config show` with `omniroute config list` in `docs/reference/CLI-TOOLS.md`.
- Replace `omniroute models list` with `omniroute models` in `README.md`.

These examples now match the actual CLI command definitions.

## Related Issue

Closes #13997

## Validation

- Documentation-only change.
- No production code changed.
- Diff contains exactly two files and two one-line corrections.
- Compared against `release/v3.8.51`: 2 commits ahead, 0 behind.

## Reviewer Notes

This PR intentionally does not include unrelated repository-wide CI fixes.


diegosouzapw:
Thanks for catching these — you're right that both examples are wrong. #14056 is already open fixing the exact same two lines for the same issue (#13997), and its comment for `config list` is a closer match to what the command actually does. I'll close this in favor of that one; happy to reopen if it stalls.

PR #14195, opened 2026-09-19
Fix/14191 pack artifact policy

prabhtheone:
## Summary

## Related Issues

- Closes #
- Related to #

## Validation

- [ ] `npm run lint`

## Tests Added Or Updated

## Coverage Notes

## Reviewer Notes


diegosouzapw:
Thanks — this is the right fix, and it matches what #14164 (the `release/v3.8.51` base-red drain, commit `5492be60`) already carries for the same two entries plus `bin/antigravity-bridge.mjs` and the snapshot test. Two things about where it lands:

- `main` does not have `httpClientAbortGuard.mjs` at all right now (#14029 removed it there; #14064 restored it on the release branch only), so a policy entry alone would make every `npm pack` on `main` fail on the *required* path. The owner's standing decision is that `main` is refreshed by the v3.8.51 release rather than by individual ports, so the fix will reach `main` with the release.
- I will add `Co-authored-by: Prabhjot Singh <jotgill1522@gmail.com>` to the #14164 squash so the credit for the policy repair is yours as well. Once #14164 merges this PR can be closed as landed; I'll link back here when that happens.

diegosouzapw:
Landed on `release/v3.8.51` via #14164 (squash `d610c24c`) with your co-authorship trailer — thank you. Closing this one as superseded; `main` receives it with the v3.8.51 release.

PR #14201, opened 2026-09-19
fix(quota): route Moonshot coding hosts to Kimi /usages

HouMinXi:
## Bug Description

A Moonshot connection whose `providerSpecificData.baseUrl` is `https://api.kimi.com/coding/v1` still went down the Open Platform balance path (`GET {origin}/v1/users/me/balance`). For Kimi Coding / Allegro API keys that origin is `api.moonshot.ai` after registry fallback, and the balance call returns 401. The dashboard then shows empty quota even though `GET https://api.kimi.com/coding/v1/usages` returns the membership windows the website displays.

Related to the Kimi Coding `/usages` fetcher already used by `kimi-coding` / `kimi-coding-apikey`.

## Root Cause

1. `resolveMoonshotOrigin` treated a missing Open Platform host as "use the moonshot registry default". An explicit `api.kimi.com/coding` URL is not Open Platform, so it must not fall back to `api.moonshot.ai`.
2. `getUsageForProvider` dispatched `provider === "moonshot"` (and uuid-shaped custom nodes) to the balance fetcher before looking at that URL.
3. Even on the Coding `/usages` path, membership percent comes from `usages.limit_7d.used_ratio` / `usages.limit_5h.used_ratio`. The parser only read integer `usage` / `limits[]`, so the website's 13% week / 9% rate-limit windows never became `code_7d` / `code_5h`.

## Fix

- Host check: `api.kimi.com` + path `/coding` or `/coding/...` is Coding Plan. Explicit coding URL is never an Open Platform origin.
- Dispatcher and `fetchMoonshotQuota` send those connections to `getKimiUsage` (`/usages` + `x-api-key`).
- Custom-node registration still wires a coding `baseUrl` so uuid ids keep a fetcher.
- Parser overlays `usages.limit_*` `used_ratio` onto `code_7d` / `code_5h`. Ratio 0 is a real empty window. Four-decimal percent rounding matches the website (`ratio * 100` is not binary-exact).
- Combo preflight caches a coding miss for the 60s TTL and goes through the shared quota-fetch throttle, same as other providers.
- `convertUsageToQuotaInfo` maps `code_5h` / `code_7d` onto the structural 5h/7d windows reset-aware scoring already reads.

## How to Verify

1. Unit: `tests/unit/moonshot-open-platform.test.ts`, `tests/unit/moonshot-quota-fetcher.test.ts`, `tests/unit/kimi-coding-billing.test.ts`, `tests/unit/generic-quota-fetcher.test.ts` (56 tests).
2. A `moonshot` connection with `baseUrl: https://api.kimi.com/coding/v1` must fetch `/usages`, not `/v1/users/me/balance`.
3. Fixture `usages.limit_7d.used_ratio = 0.147051` must surface as Code 7d ~14.7% used.

## Test Plan

- [x] Added regression tests for host classification, dispatcher URL, used_ratio overlay, used_ratio 0, miss cache, and window mapping
- [x] Existing tests in those files still pass (56/56)
- [ ] Manual: refresh Provider Limits on a moonshot row pointed at `api.kimi.com/coding` and confirm Code 7d/5h appear

## Risk Assessment

Low. Open Platform `api.moonshot.cn` / `api.moonshot.ai` hosts are unchanged. Only an explicit coding URL leaves the balance path. Fail-open on `/usages` errors still returns null and now caches that miss for 60s so a dead key cannot hammer the endpoint.

⚠️ base-red inherited: #13866


diegosouzapw:
Thanks @HouMinXi — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #7878, opened 2026-07-20
feat: support Bun bundled SQLite runtime

Arul-:
## Summary

- use Bun's bundled `bun:sqlite` as the first synchronous SQLite driver
- preserve better-sqlite3, node:sqlite, and sql.js fallback behavior
- resolve `sql-wasm.wasm` from the installed package for global/Bun layouts
- let the Bun CLI spawn the server with Bun instead of the `node` executable and incompatible V8 flags
- make CLI setup/backup and OmniRoute credential storage work with Bun SQLite
- add Bun CRUD, named-parameter, transaction, pragma, and close coverage

## Why

A global Bun install can contain a valid OmniRoute runtime and `sql.js`, but the current serve preflight exits on the incompatible bundled better-sqlite3 binary before the existing fallback chain is reached. The fallback also assumed `cwd/node_modules/sql.js`, which is not true for Bun's global install layout.

## Verification

- `bun test tests/unit/db-adapters/bunSqliteAdapter.test.ts tests/unit/db-adapters/nodeSqliteShared.test.ts` passes
- `bun test tests/unit/db-adapters/driverFactory.test.ts` passes with the installed sql.js package linked into the isolated checkout
- direct Bun CRUD probe selects `bun:sqlite`
- global-install sql.js probe successfully locates and loads `sql-wasm.wasm`
- Bun syntax/import checks pass for the modified CLI modules

The full upstream suite was not run because this isolated checkout intentionally has no dependency tree installed; the repository currently declares Node as its supported engine.

PR #14203, opened 2026-09-19
fix(db): return undefined for empty bun:sqlite .get() results

Ercaner1988:
## Summary

- `bun:sqlite`'s `Statement.get()` returns `null` when no row matches. `better-sqlite3`, `node:sqlite` and `sql.js` return `undefined`, and the rest of the code is written against that (`get(...) as Row | undefined`; `isExclusiveConnectionActivelyLeased` compares with `!== undefined`).
- Under Bun the lease guard therefore reports every connection as exclusively leased. Observed on a Bun-run server (a v3.8.50 install; the same code is on `release/v3.8.51`): the dashboard connection test returned `LEASE_ACTIVE` (skipped, 0 ms) for every connection tried (antigravity and mistral), usage refresh was deferred with 409, and the model-sync scheduler logged `No connections with autoSync enabled`, while `exclusive_connection_leases` was empty and no API key had the `lease:exclusive` scope.
- The Bun adapter now maps the no-row `null` to `undefined`, so all drivers share one contract. The change is confined to the best-effort Bun adapter; the Node drivers are untouched, and it does not change the runtime support policy (Node remains the supported runtime).
- Checked on Bun 1.4.0 and 1.4.2: both return `null` for a `.get()` with no matching row.

## Related Issues

- Related to #11468 and #7878 (Bun SQLite adapter work)

## Validation

- [x] Change type: DB (Bun adapter)
- [x] Focused tests and category gates from the golden path: `check-db-rules` passes; the `test:bun:db` file set (`bunSqliteAdapter`, `driverFactory`, `cliSqlite`) passes 24/24 on Bun 1.4.2. The two new tests failed before the change (`actual: null, expected: undefined`) and pass after it.
- [x] `npm run lint`: ESLint (repo config and suppressions) and Prettier were run on the two changed files and are clean. The full-repo lint is left to CI.
- [x] Reconciled with the current active release base: the branch was cut from the `release/v3.8.51` tip (`7a921299c`, 2026-09-19).

## Tests Added Or Updated

- `tests/unit/db-adapters/bunSqliteAdapter.test.ts`: two new tests. One uses a fake driver whose `get()` returns `null`, so it runs under any runtime, including the Node shards. The other runs against the real `bun:sqlite` and is skipped outside Bun (it is picked up by `test:bun:db`).

## Coverage Notes

- `src/lib/db/adapters/bunSqliteAdapter.ts` changes by one expression. The fake-driver test covers it on every runtime; the real-driver test covers it under Bun.
- Not run locally: the Node unit shards, `typecheck:core`, the full-repo lint and the production build. CI runs them on this PR.

## Reviewer Notes

- ⚠️ base-red inherited: #13866
- The fix is at the adapter on purpose: the `!== undefined` checks and the `| undefined` typings elsewhere assume the shared contract, so normalizing once here covers every caller instead of patching call sites one by one.
- No migration or feature flag.


diegosouzapw:
Thanks @Ercaner1988 — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #14205, opened 2026-09-19
fix(sse): read Responses-style usage names in OpenAI to Claude streaming

### LINKED ISSUE #14204: fix(backend): OpenAI to Claude streaming reports 0/0 usage when upstream uses Responses token names
Claude clients behind OmniRoute see 0 input / 0 output tokens when the OpenAI-compatible upstream reports usage with Responses API names (input_tokens / output_tokens) instead of chat-completions names (prompt_tokens / completion_tokens).

Version: 3.8.51 (release/v3.8.51 @ 7a921299)
Install: built from source
OS: macOS, node v26.8.1

Steps:
1. Point a Claude client (e.g. Claude Code) at OmniRoute with an OpenAI-format upstream that sends usage as {input_tokens, output_tokens, total_tokens} on the trailing choices-empty chunk (stream_options include_usage style).
2. Complete one streaming request.
3. Check the message_stop usage block and the call log token counts.

Expected: real input/output token counts.
Actual: input_tokens 0, output_tokens 0. The rest of the pipeline already accepts both namings (stream.ts usage aggregation, usageTracking.ts, openai-responses.ts) but trackUsageFromChunk in open-sse/translator/response/openai-to-claude.ts only reads prompt_tokens / completion_tokens / prompt_tokens_details, so the Claude leg drops the numbers.

Happy to put up a small fix that reads both namings the way the sibling code does.

soroush5:
The streaming translator only read prompt_tokens/completion_tokens, so upstreams reporting Responses-style input_tokens/output_tokens left Claude clients with zeroed usage. It now reads both namings like the rest of the pipeline.

Verified: 4 new tests fail before and pass after, neighbors green.

Fixes #14204


diegosouzapw:
Nice catch — this brings the Claude-format streaming translator in line with what open-sse/utils/stream.ts already does for the same prompt_tokens/input_tokens ambiguity, so it's a consistency fix rather than a new pattern. The precedence test (OpenAI naming wins when both are present) is the right call for safety. Merging.

One small item: please add a `changelog.d` fragment (the merge-integrity gate requires one) — or we'll add it at merge time.

soroush5:
Added the changelog fragment. Thanks for merging.

PR #14207, opened 2026-09-19
fix(combos): avoid server-only model import in dashboard

VardoKen:
## Summary

- Fix the Combos dashboard importing the server-only `open-sse/services/model.ts` module into the client bundle.
- Use `resolveProviderAlias()` directly from the client-safe `providerAlias.ts` module because this code path only needs the canonical provider ID.
- Prevent server-only dependencies, including Playwright and Node.js built-ins, from being pulled into the browser bundle.
- Add a regression test guarding the client/server module boundary.

## Related Issues

- Related to #13951

## Validation

- [x] Change type: UI
- [ ] `npm run lint`

## Tests Added Or Updated

- Added `tests/unit/dashboard/combos-client-boundary.test.mjs`.
- Regression test verifies that the Combos dashboard:
  - imports `resolveProviderAlias()` from the client-safe `providerAlias.ts`;
  - does not import server-only `open-sse/services/model.ts`;
  - does not reference `resolveCanonicalProviderModel()`.
- Focused regression test passed: 1 test, 1 passed, 0 failed.

## Coverage Notes

- This PR changes `src/app/(dashboard)/dashboard/combos/page.tsx`.
- The new regression test directly guards the client/server import boundary responsible for the production build failure.
- The production Docker build completed successfully after the fix.

## Reviewer Notes

- Root cause was introduced when the Combos dashboard began importing `resolveCanonicalProviderModel()` from `open-sse/services/model.ts`.
- That import transitively pulled server-only dependencies into the Client Component browser bundle, including Playwright and Node.js built-ins such as `fs`, `child_process`, `net`, and `async_hooks`.
- This code path only needs provider alias resolution, so `resolveProviderAlias()` provides the required behavior without importing the server-only model service.
- `npm run check:dashboard-typecheck` was executed but currently fails on three unrelated pre-existing/regressed files:
  - `src/app/(dashboard)/dashboard/providers/[id]/__tests__/proxySaveRefresh.test.tsx`
  - `src/mitm/handlers/antigravity.ts`
  - `src/shared/components/Sidebar.tsx`
- `npm run lint` was executed after fixing the PR-local unused variable and now fails only on the unrelated existing issue in `src/shared/services/cliRuntime.ts`.
- No migrations, database changes, or feature flags are involved.


VardoKen:
@diegosouzapw CI workflows are awaiting maintainer approval for this fork PR. Could you please approve and run the workflows? Thanks!

VardoKen:
@Mergifyio queue

VardoKen:
@diegosouzapw I cannot queue this PR because Mergify requires sender-permission >= write. Please approve the pending workflows and queue/merge the PR when ready. Thanks!

diegosouzapw:
Confirmed the server-only import was the real cause (Playwright + Node builtins leaking
into the client bundle via `open-sse/services/model.ts`) and `resolveProviderAlias()` is the
right client-safe replacement. One note: the new test guards the import boundary via a
source-text regex rather than exercising the resolution behavior — worth a follow-up
behavioral test confirming `resolveProviderAlias(prefix)` matches what
`resolveCanonicalProviderModel(...).provider` would have returned for any ambiguous gateway
prefixes in the table, so the `allowedProviders` restriction UI doesn't silently regress.
Not blocking — looks merge-ready as is.

Issue #6304, opened 2026-07-05
fix(providers): Auggie (Augment CLI) provider fails with spawn EINVAL on Windows


rogerelmaalouf:
### OmniRoute Version

3.8.44

### Installation Method

npm (global)

### Operating System

Windows

### Description

Title: Auggie (Augment CLI) provider fails with spawn EINVAL on Windows, regardless of Node version or build tools

Environment:
- OS: Windows
- OmniRoute version: 3.8.44
- Node.js: tested on both v24.18.0 and v22.23.1, same result on both
- Auggie CLI: v0.32.0, installed globally via npm, logged in via OAuth

Steps taken to rule out causes:
1. Confirmed Auggie CLI works standalone (auggie --version returns 0.32.0)
2. Confirmed OAuth login to Augment account completed successfully
3. Switched Node from v24 to v22 via nvm-windows, reinstalled omniroute and auggie under v22, same error
4. Installed Visual Studio Build Tools 2026 with Desktop development with C++ workload
   (cl.exe confirmed present via vcvarsall.bat x64)
5. Manually rebuilt wreq-js with build tools available, no build errors shown, still fails

Error when testing the Augment provider (Test button in dashboard, and via combo):
[502]: spawn EINVAL

omniroute doctor consistently shows:
WARN CLI Tools: Could not run CLI tool checks: Cannot find package '@/shared' imported from
.../omniroute/src/lib/cli-helper/tool-detector.ts

This suggests an unresolved TypeScript path alias (@/shared) that was not bundled to a
relative import before the package was published to npm. This may be the same code path
that fails when actually spawning the Auggie CLI process (spawn EINVAL), not just the
doctor diagnostic.

All other providers (OpenCode Zen, MiMoCode, DuckDuckGo AI Chat) work correctly on the
same machine and Node version.

### Steps to Reproduce

Title: Auggie (Augment CLI) provider fails with spawn EINVAL on Windows, regardless of Node version or build tools

Environment:
- OS: Windows
- OmniRoute version: 3.8.44
- Node.js: tested on both v24.18.0 and v22.23.1, same result on both
- Auggie CLI: v0.32.0, installed globally via npm, logged in via OAuth

Steps taken to rule out causes:
1. Confirmed Auggie CLI works standalone (auggie --version returns 0.32.0)
2. Confirmed OAuth login to Augment account completed successfully
3. Switched Node from v24 to v22 via nvm-windows, reinstalled omniroute and auggie under v22, same error
[...]

PR #14215, opened 2026-09-19
fix(auggie): resolve spawn EINVAL when running the CLI shim on Windowsfix(auggie): resolve spawn EINVAL when running the CLI shim on Windows

andreyzagid-tech:
## Problem

On Windows, `auggie` requests fail with `[502]: spawn EINVAL`. Since Node's CVE-2024-27980 fix (>=18.20.2 / 20.12.2 / 21.7.3), `spawn()` on a `.cmd`/`.bat` shim **without** `shell: true` throws `spawn EINVAL` synchronously.

`#6304` fixed the two request-time spawn sites (`spawnAuggie`/`runStreaming`) by routing them through the shared `buildAuggieSpawnOptions()` helper, which sets `shell: true` on win32. But two other CLI spawn sites were left bypassing it:

- `initAuggieModels()` — `auggie model list`
- `checkAuggieCliVersion()` — `auggie --version`

Both still threw `spawn EINVAL` on win32, so provider discovery and version checks kept failing even after `#6304`.

## Fix

- Generalize `buildAuggieSpawnOptions<S extends readonly string[]>(stdio)` so any stdio layout can be routed through it, and add `windowsHide: true`.
- Route `initAuggieModels` and `checkAuggieCliVersion` through `buildAuggieSpawnOptions` (shell enabled on win32).

## Test

New `tests/unit/auggie-win32-spawn-init-models-and-version.test.ts` (Windows-only, `skip: !isWin32`) uses real `.cmd` batch shims and spawns end-to-end:

- `checkAuggieCliVersion` runs a `.cmd` shim → fails with `spawn EINVAL` before this fix, passes after
- `initAuggieModels` discovers models from a `.cmd` shim → same before/after

Verified: new test 2/2 ✅, existing `auggie-win32-spawn-6304` 3/3 ✅, `tsc --pretty false -p tsconfig.typecheck-core.json` ✅, eslint ✅.


diegosouzapw:
Thanks @andreyzagid-tech — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #14220, opened 2026-09-19
fix(proxy-logs): keep the HTTP status the provider actually returned on search rows

maxmad64bis:
⚠️ base-red inherited: #13866

## Summary

Search proxy-log rows always read "no response received", even when the provider answered with a 429, 403, or 500. The chat writer already captures the received status; the search writer never forwarded it. After this change, search rows carry the status the provider actually returned, and keep null only when no response arrived.

## Related Issues

- Related to #13580 (the chat writer this extends to search).

## Validation

- [x] Change type: provider
- [ ] `npm run lint` — ESLint 0 on the two touched source/test files; the full `eslint .` run is red on the base, unrelated to this diff.
- [x] Reconciled with the current active release base; focused checks rerun afterward — `rev-list --count HEAD..upstream/release/v3.8.51` = 0, gates re-run after the last amend.
- [x] Production-code changes include a new or updated automated test in this PR — new `tests/unit/search-proxy-upstream-status.test.ts` (6 cases, RED-then-GREEN).

## Tests Added Or Updated

- New `tests/unit/search-proxy-upstream-status.test.ts` (6 cases, local HTTP server, journal capture): real 429/200/500 arrive on the row (RED before: null), transport timeout and synthesized envelope codes stay null; neighbor `search-432-plan-limit-cooldown.test.ts` still green (7/7).

## Coverage Notes

- Single-file change plus test: `executeProviderFetch` is the only search path, so every search provider is covered without a list.

## Reviewer Notes

- Never copies a locally fabricated code into `upstream_status` (envelope 402/502, transport 502/504, and post-hoc derived statuses stay null). Local gates: targeted ESLint 0, `typecheck:core` 0, new test 6/6 RED-then-GREEN, C4 re-run at push time (P2 merged, migration 179 present on the base).
- CI: red jobs are inherited from the red base (#13866), not from this diff — the third-party #14219 at the same base fails the same gates (api-typecheck 2 ✖ plus quality/docs/eslint/unit fast-path reds); all three files of this diff are outside the cited files. Non-blocking for this PR.


diegosouzapw:

PR #14180, opened 2026-09-19
feat(sse): per-egress pacing + fleet-wide backoff for opencode rotation (opt-in)

maxmad64bis:
⚠️ base-red inherited: #13866

## Summary

When several concurrent requests share the same network egress, the multi-account rotation fires them all at once, so a burst hits the provider simultaneously. After ten burst 429s in a minute the fleet is treated as suspect for 60-120 s, during which each request tries at most two accounts before handing over to the wait-and-replay path of #13924 (heartbeat plus one capped leg; last upstream answer only when that path is off), and the first success clears the state. Real rate limits (parseable Retry-After or rate-limit wording) never feed the suspect counter — they keep the existing early-stop path. Everything fails open and the behavior is opt-in via `OPENCODE_EGRESS_THROTTLE_ENABLED=1` (default off: rotation unchanged).

Related to #10880 (IP-bucketed 429s waste one upstream call per sibling account) and #10920 (stop retrying sibling connections after an IP-scoped 429): this paces dispatches per egress before they fire and backs the fleet off once burst pressure is proven, instead of locking accounts after the fact.

## Related Issues

- Related to #10880
- Related to #10920
- Related to #13924

## Validation

- [x] Change type: provider
- [x] Focused tests and category gates from the golden path
- [x] `npm run lint`
- [x] Reconciled with the current active release base; focused checks rerun afterward
- [x] Production-code changes include a new or updated automated test in this PR

## Tests Added Or Updated

- `tests/unit/opencode-egress-throttle.test.ts` (new, 22 tests: config defaults and bounds, cap and FIFO, jitter ranges, fleet window and expiry, fail-open paths, first-success reset, simulated burst wave, per-arm leak guards, caps eviction)

## Coverage Notes

- `open-sse/executors/opencodeEgressThrottle.ts` (new, ~545 lines) is covered by the 22 tests above; `open-sse/executors/opencode.ts` seam (3 call points, request-local budget) is covered by the leak-guard and budget tests. No existing test needed changes.

## Reviewer Notes

[...]

PR #14231, opened 2026-09-20
fix(proxies): share scopeId guard across validation and stored assignments

maxmad64bis:
⚠️ base-red inherited: #13866

## Summary

Non-global proxy scopes previously repeated the same `scopeId` check in several places with drift between them, so this change shares one validation rule and one runtime guard while keeping every existing 400 message byte-identical. Two deliberate default changes ship here: whitespace-only `scopeId` values on stored assignments are now rejected, and clearing a non-global scope without a `scopeId` is now rejected.

## Related Issues

- Related to #13575 (proxy pool could not switch members on the chat path; this change keeps the pool-scope validation contract stable while pool routing evolves)

## Validation

- [x] Change type: provider
- [x] Focused tests and category gates from the golden path (29 guard tests + 52 focused (guard/health/crud/version suites), typecheck 0, file-size/complexity OK)
- [ ] `npm run lint` — touched files clean locally; the full run is red on the base (inherited, non-blocking, see Reviewer Notes)
- [x] Reconciled with the current active release base; focused checks rerun afterward (0 behind, no realignment needed)

## Tests Added Or Updated

- `tests/unit/proxy-scope-id-guard.test.ts` (new, 29 tests: shared refinement across the five validation schemas with frozen messages, whitespace rejected, raw-scope guard including unknown scopes, pool route coverage, stored-assignment guards via public functions).

## Coverage Notes

- `src/shared/validation/schemas/proxy.ts` covered by the new guard matrix (all five schemas, both frozen messages, whitespace input); `src/lib/db/proxies/mappers.ts` plus its four call sites covered through direct guard assertions and public-function throws.
- No coverage regression expected: every touched production file ships with its dedicated test file.

## Reviewer Notes

- Default changes are explicit: (1) whitespace-only `scopeId` on stored assignments now throws instead of being accepted (validation schemas already trim, so only the stored path changes); (2) clearing a non-global scope without a `scopeId` is now rejected (guard placed before the detach branch).
- The shared guard takes the scope raw: the pool route passes the alias-normalized value so `?scope=bogus` without `scopeId` still returns 400, while stored-assignment sites pass normalized values. Unknown scopes keep the existing 400, global scope without `scopeId` is unchanged, the guard is a pure predicate with no shared state, and the same TS2677 pair reproduces on sibling #14180.


diegosouzapw:

PR #14233, opened 2026-09-20
fix(proxies): classify refusal cause and hang, add bounded opt-in recovery pass

maxmad64bis:
⚠️ base-red inherited: #13866

## Summary

A destination refusing the probe was misread as a dead proxy, and refused proxies were never re-tested. This PR carries the refusal cause next to the verdict without touching proxy status, plus an opt-in recovery pass. Defaults: stalled handshakes count like frank failures, unproven 403s stay neutral, recovery off unless `PROXY_HEALTH_RECOVERY_ENABLED` is set.

## Related Issues

- Related to #10654 (refused egress IP is not a healthy proxy).
- Related to #6246 (probe outcomes stay advisory; operator owns proxy status).

## Validation

- [x] Change type: routing
- [x] Focused tests and category gates from the golden path (25 new/updated proxy-health tests + neighbours (12+7+7), tsc filtered 0, eslint perimeter clean)
- [ ] `npm run lint` — touched files clean locally; the full run is red on the base (inherited, non-blocking, see Reviewer Notes)
- [x] Reconciled with the current active release base; focused checks rerun afterward (0 behind, no realignment needed)
- [x] Production-code changes include a new or updated automated test in this PR

## Tests Added Or Updated

- `tests/unit/proxy-health-refusal-cause.test.ts` (new, 5): 401/403/429/451 matrix over proven/unproven/absent; 403 without evidence neutral, never mutates.
- `tests/unit/proxy-health-probe-error.test.ts` (new, 3): stalled handshake vs plain local abort vs frank connection failure.
- `tests/unit/proxy-health-recovery-plan.test.ts` (new, 4): quota-refusal only, drops unreachable (even with 429 observation) and null keys; oldest-first with volume cut.
- `tests/unit/proxy-health-recovery-pass.test.ts` (updated, 4): fake probe plus real refusal memory; success lifts, failure doubles recurrence, off means zero probes.
- `tests/unit/proxy-health-recovery-default.test.ts` (new, 1): real default path, no injected probe (local 429 relay then 200, real registry/memory).
- `tests/unit/proxy-health-decide-action-6246.test.ts` (extended, 8): stalled handshake equals frank failure at/below threshold, opt-ins on/off.
- Neighbours green (12+7+7); filtered `tsc` 0, ESLint clean; RED-then-GREEN (new tests failed on missing imports first; default-path RED by stash, GREEN after).

## Coverage Notes

[...]

Issue #13459, opened 2026-09-12
fix(backend): usage_history stores provider aliases verbatim, splitting telemetry between alias and canonical id


mmzac1:
## Problem

`usage_history` records whatever provider string the caller used, so requests to the same backend land under two different names depending on whether the alias or the canonical id was used. Usage, cost and latency analytics are then split across both.

Observed on a local install with a single Ollama connection (`ollama-local`, alias `ollama`):

```
SELECT provider, COUNT(*) FROM usage_history GROUP BY provider;

ollama-local   19
ollama          2     <-- same connection, same model, same backend
```

Both rows came from one `ollama-local` provider connection serving `qwen3.5:2b`. The only difference is the model string in the request: `ollama/qwen3.5:2b` vs `ollama-local/qwen3.5:2b`.

## Steps to reproduce

1. Register a single `ollama-local` provider connection.
2. `POST /v1/chat/completions` with `"model": "ollama-local/qwen3.5:2b"`.
3. `POST /v1/chat/completions` with `"model": "ollama/qwen3.5:2b"` (the documented alias).
4. Group `usage_history` by `provider` — two rows for one connection.

## Expected

Both requests recorded under the canonical provider id (`ollama-local`), since `ollama` is declared as an alias, not a separate provider.

## Root cause

`resolveProviderId(aliasOrId)` already exists and does exactly this mapping:

- `src/shared/constants/providers.ts:368`

Neither `usage_history` writer calls it; the provider string is bound straight through:

- `src/lib/usage/usageHistory.ts:684` — `entry.provider || null`
- `src/lib/usage/migrations.ts:316` — `@provider`

Neither file imports `resolveProviderId`.

## Suggested fix

Normalise at the write boundary so historical rows stay consistent regardless of what callers send:

```ts
import { resolveProviderId } from "@/shared/constants/providers";
// ...
entry.provider ? resolveProviderId(entry.provider) : null,
```

Alias table for reference (`src/shared/constants/providers/*.ts`): `ollama`→`ollama-local`, `lmstudio`→`lm-studio`, `cc`→`claude`, `gh`→`github`, `cx`→`codex`, `kmc`→`kimi-coding`, `xao`→`xai-oauth`, `gc`→`grok-cli`.

A one-off UPDATE can fold existing alias rows into their canonical ids for anyone already affected.

## Notes

- Cosmetic rather than functional: routing, dispatch and billing all work correctly. Only the analytics grouping is affected.
[...]

Issue #12708, opened 2026-09-04
[Feature]  LLM Telemetry and Shadow Routing


thermxpertro-hue:
### Problem / Use Case

Add a built-in observability and evaluation layer for LLM usage.

The feature should support per-agent model credentials, request correlation IDs, task-type tagging, cost/token/latency tracking, tool-call success metrics, and persistent usage logs.

It should also support shadow routing: a configurable percentage of production requests are duplicated to a secondary local or alternative model for evaluation only. Shadow responses must never affect the user-facing response, execute tools, or perform external actions.

This would enable teams to compare cloud and local models on real workloads and generate reliable 30-day reports covering:

cost per agent/task/model;
latency and cold vs warm performance;
success and fallback rates;
tool-call reliability;
local-vs-cloud quality;
estimated savings;
recommended routing per task type.

The goal is to make Rakazo capable of data-driven model routing and safe migration from expensive cloud models to local or cheaper models without risking production quality.

### Proposed Solution

This would enable teams to compare cloud and local models on real workloads and generate reliable 30-day reports covering:

cost per agent/task/model;
latency and cold vs warm performance;
success and fallback rates;
tool-call reliability;
local-vs-cloud quality;
estimated savings;
recommended routing per task type.

The goal is to make Rakazo capable of data-driven model routing and safe migration from expensive cloud models to local or cheaper models without risking production quality.

### Acceptance Criteria

cost per agent/task/model;
latency and cold vs warm performance;
success and fallback rates;
tool-call reliability;
local-vs-cloud quality;
estimated savings;
recommended routing per task type.

### Area

Proxy / Routing

diegosouzapw:
Thanks — there is a real feature request in here, but it needs slicing before anyone can act on it, and part of it already exists.

First, a housekeeping note: the body asks to "make **Rakazo** capable of data-driven model routing". There is no Rakazo in this codebase, and the "Proposed Solution" section repeats the "Problem" section verbatim. If this was written against another product, it would help to restate what you specifically need from OmniRoute.

**What already exists**, so it should come off the list:

- Per-request cost, token and latency accounting, persisted in `call_logs`.
- Request correlation ids.
- Durable usage logs with export to external sinks (BigQuery and friends) — see `docs/frameworks/LOG-EXPORT.md`.
- An evaluation framework with A/B comparison between targets — `docs/frameworks/EVALS.md`.
[...]

Issue #12617, opened 2026-09-03
feat(api): map X-OmniRoute cost/usage/tok-s headers into OpenCode plugin telemetry


RaviTharuma:
## Feature

Map **gateway-reported cost, token usage, and tokens-per-second** into the host harness so a run shows real `$` and `tok/s`, not catalog guesses.

## Why

OmniRoute already emits (non-stream) `X-OmniRoute-Response-Cost`, `X-OmniRoute-Tokens-In/Out`, `X-OmniRoute-Latency-Ms`, `X-OmniRoute-Model`, `X-OmniRoute-Provider`, `X-OmniRoute-Decision`. CLIProxyAPI often has upstream `usage` and, on some builds, internal TPS — but plugins typically register models and stop there.

Harness cost UIs (OpenCode session cost, Pi usage, OMP status) therefore:

- price the **requested** model from models.dev / `/api/pricing` even after a combo/credential failover
- omit tok/s, or compute it client-side from deltas (wrong on buffered streams and tool pauses)

Live generation speed cannot come from models.dev. It has to ride on the completion.

## Requested behavior

1. On each completion (stream final chunk included), read gateway usage/cost/tok/s headers or `usage` fields and write them into the host’s usage/cost/speed slots.
2. If the gateway updates the routed model mid-request (combo / credential failover), update the host’s displayed model id for that turn.
3. Prefer gateway numbers over catalog `cost` × token estimates when both exist.
4. If the gateway has no tok/s yet, do **not** invent one from `tokens / latency_ms` (that includes TTFT). Show `—` until the gateway field exists.
5. Keep catalog sync (models.dev / `/v1/models` / `/api/pricing`) for picker metadata; this issue is **per-request** telemetry.

Related gateway asks (please link rather than re-implement scraping):

- OmniRoute: tok/s + streaming cost headers
- CLIProxyAPI: tok/s / usage on the inference response

## Acceptance

- One streamed turn shows host cost + tok/s that match the gateway response.
- After failover, the turn is attributed to the model that answered.
- No extra HTTP calls to dashboards or log APIs.

RaviTharuma:
Gateway-side request for tok/s + streaming cost/usage that plugins can map:

- https://github.com/diegosouzapw/OmniRoute/issues/12616

Plugin mapping requests (do not invent tok/s from `tokens / latency_ms`):

- Official OpenCode plugin: https://github.com/diegosouzapw/OmniRoute/issues/12617
[...]

PR #14235, opened 2026-09-20
feat(plugins): register host sdk telemetry hook behind opt-in flag

maxmad64bis:
⚠️ base-red inherited: #13866

## Summary

The editor plugin registers the host `sdk` hook only when the new `telemetry` option is enabled, and hosts without the `sdk` domain keep loading the catalog untouched. When enabled, the hook marks matching inference calls in their options only — filtered by provider and SDK package — without wrapping fetch, without assigning a client, and any registration failure is reduced to a warning; off by default and inert on older hosts.

## Related Issues

- No linked issue — no open upstream issue covers host `sdk` hook registration (neighbors #12708, #12617, #13459 are closed gateway-side items).

## Validation

- [x] Change type: other
- [x] Focused tests and category gates from the golden path (5 hook tests + full package suite 244/244, eslint 0 on touched files)
- [ ] `npm run lint` — touched files clean locally; the full run is red on the base (inherited, non-blocking, see Reviewer Notes)
- [x] Reconciled with the current active release base; focused checks rerun afterward (0 behind, no realignment needed)

## Tests Added Or Updated

- `@omniroute/opencode-plugin-v2/tests/telemetry-sdk-hook.test.ts` (new, 5 tests: older host loads, option-off registers nothing, option-on registers, foreign-provider options untouched, refusing host keeps its catalog).

## Coverage Notes

- Touches only the editor plugin package (`@omniroute/opencode-plugin-v2/src/index.ts` hook registration plus a 3-line opt-in option in `options.ts`), outside the `src/` coverage gate paths.
- Every new branch (domain absent, option off/on, filter match/mismatch, registration refusal) is covered by the new test file; no existing test needed changes.

## Reviewer Notes

- Split from a three-concern draft: this PR keeps only the telemetry-hook concern (1 commit, 4 files: hook registration, opt-in option, new test file, changelog fragment). Strict fallback is the whole implementation — the verification gate found no provable options-only marking against the pinned host contract, so the hook only marks options and the pure telemetry core is neither created nor imported. Gate re-anchored on repo-verifiable facts: the host mock in `gemini-language.test.ts` mounts a callable `sdk` domain, and `@opencode-ai/plugin 1.18.29` in `package.json` is the pinned contract reference whose `sdk`-event option shape is treated as unknown.
- Reserve: the callback reads `input.package` and `input.model.providerID` without optional chaining, so a malformed host event would throw inside the callback (the registration-time try/catch does not cover it); kept as-is for parity with the existing `languageHook` filter. Otherwise fail-safe: domain absent yields catalog only, option off never touches the hook, foreign providers leave options untouched, and a refusing host warns with the catalog kept — no path wraps fetch (failure lists identical to #14233 on the same base).


diegosouzapw:
Reconciled against the tip, as flagged on #14370.

#14370 landed on `release/v3.8.51` first and rewrote `src/index.ts`: the plugin now imports the stable `@opencode/plugin` contract instead of the beta `@opencode-ai/plugin`, catalog publishing goes through `Plugin.define` + `ctx.provider.transform`, and the per-event `aisdk` domains collapsed into one named entry point, `ctx.aisdk.hook(name, cb)`.

Your registration was written against the old `ctx.aisdk.sdk(cb)` shape, so it was **re-applied** on the new one rather than merged as-is: `hook("sdk", …)`, typed from the published `AISDKHooks["sdk"]` shape (`model` / `package` / `options`) — the same three fields your callback already read. Nothing about the design moved:

- gated on the opt-in `telemetry` option, still `z.boolean().default(false)`;
- inert on a host that exposes no `aisdk` domain (catalog loads untouched);
- marks only this provider's models whose package contains `@ai-sdk/openai-compatible`;
- `options`-only — no fetch wrapping, no `sdk` assignment;
- a refused registration (sync throw) and a failed one (rejected promise) both degrade to a warning.

`tests/telemetry-sdk-hook.test.ts`: all five `it(...)` bodies and every assertion are byte-identical to what you wrote. Only the host fake was re-shaped to the stable contract — `catalog` → `provider` + `model` (what `assertContext` now requires) and the two callable `aisdk` domains → one `hook(name, cb)` router that records only the `"sdk"` registration.

Validation on the reconciled tree: your hook suite 5/5; the full `@omniroute/opencode-plugin-v2` suite **244 pass / 0 fail** (the 239 from #14370 plus your 5); `npm run build` (the tsup DTS step is a strict `tsc` over `src`) and `tsc --noEmit -p tsconfig.json` (src + tests) clean; `typecheck:core`, `check-file-size` and `check-changelog-integrity` green.

The strict-fallback reasoning held up exactly as you wrote it — only the plumbing underneath it moved. Nothing for you to do.

PR #14237, opened 2026-09-20
feat(providers): surface new Claude, xAI, and Grok Build chat ids

HouMinXi:
## Summary

Claude's `/v1/models` GET now sends `limit=1000`, so a single request covers the current catalog instead of Anthropic's default page size. xAI API-key connections accept unknown model ids the same way `xai-oauth` already does. Live OAuth catalog fetch for `xai-oauth` is on by default (`XAI_OAUTH_LIVE_MODEL_DISCOVERY`); set the flag to false to keep the frozen seed. New Grok Build OAuth imports persist `autoSync: true` so the next published grok id shows up without a registry seed bump.

Claude does not get `passthroughModels`. A 429 on Claude still cools the whole account. There is no `after_id` loop.

## Test plan

- `t tests/unit/flagship-0day-discovery.test.ts tests/unit/xai-oauth-discovery.test.ts tests/unit/l1-oauth-autosync-default.test.ts tests/unit/combo-claude-per-model-scope.test.ts tests/unit/feature-flags-doc-sync-static.test.ts tests/unit/grok-cli-oauth.test.ts` — 87/87.
- Inject: drop `?limit=1000` from the Claude URL → the exact-URL test fails. Restore → pass.
- Inject: drop `passthroughModels` on `xaiProvider` only → xAI apikey half fails, oauth half still green.
- Inject: drop `autoSync` on `mapImportedToken` → JWT/auth.json tests fail, browser path still green. Drop it on `mapGrokBuildBrowserTokens` → the reverse.
- `combo-claude-per-model-scope` still treats Claude 429 as account-wide.

## Notes

`xai` / `xai-oauth` / `xao` stay out of `CATALOG_SIBLING_IDS` and `PROVIDER_SEARCH_PAIRS`. Seeds do not add `grok-4.7`, `gemini-4.0-pro`, or `claude-opus-5-1`.

⚠️ base-red inherited: #13866


HouMinXi:
Adding context on the semantic updates here and the boundaries we kept:

**xAI live discovery on by default**
`XAI_OAUTH_LIVE_MODEL_DISCOVERY` was previously off by default, leaving xai-oauth users on static seed lists unless they knew to export the flag. Since xAI rotates model names regularly, seeds go stale quickly. The live discovery path in `route.ts` already falls back to the static catalog if xAI is unreachable, so flipping the default gives users fresh chat IDs without breaking offline or degraded startups.

**autoSync on new Grok Build imports**
Browser and `auth.json` token imports used to omit `autoSync`, so imported accounts never updated their model list unless re-imported from scratch. Stamping `autoSync: true` on import lets the existing sync loop keep them current.

**xAI API key model passthrough**
We enabled `passthroughModels` for xAI API keys to match `xai-oauth`. Unrecognized model IDs pass through to xAI rather than getting blocked by our local schema validator. If an ID is invalid, xAI returns 404, and `perModelFailureScope` already isolates that failure to the single model ID.

**Why Claude was intentionally excluded from passthrough**
We specifically did not enable `passthroughModels` on Claude. Enabling passthrough on Claude turns `hasPerModelQuota("claude")` on, which changes 429 handling from cooling the entire account to locking just the failed model name. That would let combos keep pounding an already-exhausted Anthropic key with sister models. For Claude, requesting `limit=1000` on `/v1/models` gets the full current catalog in a single call without touching the quota failure domain.

HouMinXi:
Merge-order note for whoever lands second between this PR and #14525.

#14525 (grok build-object supportedThinkingEfforts) intentionally seeds grok-4.7 into the registry and keeps xhigh on grok models discovered from object catalogs. If it merges first, two guard tests in this PR need updating:

- tests/unit/flagship-0day-discovery.test.ts, test 9 ("zero-day models absent from support surface"): grok-4.7 has to come out of the banned list, since it will be present on purpose.
- tests/unit/provider-models-discovery-split.test.ts, "preserves exact supported reasoning efforts": grok now keeps xhigh, so the expected lists for 4.5/4.6/4.7 gain "xhigh", and the third fixture model (grok-4.8, object form with only xhigh declared) yields ["xhigh"] instead of undefined under the explicit-list filtering semantics.

If this one merges first instead, #14525 needs the reciprocal updates on its side. I hit both failures while building an integration branch carrying the two PRs together, so this is measured, not speculative.

Issue #2822, opened 2026-05-28
[BUG] qwen3.7-max (opencode-go) returns 500 on image inputs — upstream model does not support vision


jeferssonlemes:
### Description

**Environment**
- Provider: `opencode-go` (`https://opencode.ai/zen/go/v1/messages`)
- Model: `qwen3.7-max`
- targetFormat: `claude` (per #2292, qwen3.x rejects oa-compat)
- Client: any OpenAI-format client routing through OmniRoute (reproduced with opencode TUI 1.15.10 + `@ai-sdk/openai-compatible`)

**Bug**

Sending an image attachment to `qwen3.7-max` via OmniRoute → `opencode-go` consistently returns HTTP 500 from upstream:

```json
{"type":"error","error":{"type":"error","message":"Internal server error"}}
```

The client side surfaces this as `[opencode-go/qwen3.7-max] [500]: Internal server error (reset after 3s) [retrying 1s attempt #3]` and never recovers, burning the per-URL retry budget.

Text-only requests to the same model on the same endpoint succeed normally — only the presence of an `image` block triggers the 500.

**Root cause (upstream)**

Independent test confirms the underlying Bailian `qwen3.7-max` model itself does not currently process images. When the model finally responded directly, it replied verbatim:

> "Não consigo visualizar imagens — o modelo atual (qwen3.7-max) não suporta input de imagem."

Despite this, the `opencode-go` proxy returns an opaque `500 Internal server error` (instead of e.g. `400 "Model does not support image inputs"`), so OmniRoute treats it as a transient failure and the retry loop fires.

**OmniRoute side is correct**

I instrumented `open-sse/executors/base.ts` to dump the outgoing request body for `opencode-go` + `qwen3.x` to `/app/data/debug-qwen-body/`. The shape is the canonical Anthropic image block:

```json
{
  "type": "image",
  "source": { "type": "base64", "media_type": "image/png", "data": "..." }
}
```

`media_type` is correct (`image/png`), no `image_url` blocks leak through, body ~210 KB (image ~120 KB base64 / ~90 KB raw). Translation in `openai-to-claude.ts:543-555` handles `data:image/...;base64,` correctly. No bug in the translator.

**Proposed workaround**

Until upstream fixes either the proxy (return 400 with a clean message) or the model (add real vision support), OmniRoute should avoid the 500/retry loop for these models. Options:

[...]

Issue #2292, opened 2026-05-15
[BUG] + [FIX] opencode-zen: upstream returns wrong format (Claude) on OpenAI endpoint + rejects non-streaming for some models


raccoonwannafly:
## Problem

The `opencode-zen` provider has two upstream behavioral issues that cause failures for certain models (observed with `qwen3.6-plus-free` and `qwen3.6-plus`):

### 1. Response format mismatch (streaming)

The upstream returns **Claude/Anthropic format** responses (`type: "message_start"`, no `choices` array) even when the request is sent to the **OpenAI-compatible endpoint** (`/v1/chat/completions`) with OpenAI format headers. The HTTP status is `200`, so the executor treats it as a successful OpenAI response and passes it through untranslated.

**Client error:**
```
Type validation failed: Value: {"message":{"model":"qwen3.6-plus",...,"type":"message_start"}}.
Error: expected "choices" (array), received undefined
```

The client (expecting OpenAI format) receives raw Claude format and Zod validation fails.

### 2. Non-streaming requests return 500

The same models that work with `stream: true` return `500 Internal Server Error` when `stream: false` is set. The format fallback chain (`openai → claude`) tries both formats, but both return 500 for non-streaming requests. This means the dashboard test function (which uses `stream: false`) always fails for these models, even though they work fine via streaming clients like Kilo Code.

**Server log (non-streaming test):**
```
qwen3.6-plus-free: openai returned 500, trying next format
qwen3.6-plus-free: claude returned 500, trying next format
post_executor t=2068ms {"status":500}
```

**Server log (streaming via Kilo — works):**
```
post_executor t=3842ms {"status":200}
Translation mode: claude → openai
```

## Root Cause

The `OpencodeExecutor` format fallback mechanism correctly retries on 500 and switches request format, but it doesn't detect when a **200 response** has the wrong body format (Claude on an OpenAI endpoint). The response translation layer uses the pre-determined target format for translation, not the actual format returned by the upstream.

## Working Fix (tested on main branch)

We have a working fix on a local branch based on `main`. The approach has three parts:

### Part 1: Track the actual format used by the executor

In `open-sse/executors/opencode.ts`, add a `_lastSuccessfulFormat` field that persists after `execute()` returns, and set it based on the actual response:

```typescript
export class OpencodeExecutor extends BaseExecutor {
  _requestFormat: string | null = null;
  _lastSuccessfulFormat: string | null = null;

  async execute(input: ExecuteInput) {
[...]

PR #14242, opened 2026-09-20
fix(providers): stop rewriting GA qwen3.8-max to preview on opencode-go (#14181)

### LINKED ISSUE #14181: fix(opencode-go): stop rewriting qwen3.8-max to preview

xiechimon:
## What

OpenCode Go now serves a GA `qwen3.8-max` (it shows up as an imported model), but OmniRoute's built-in deprecation table (written when the model shipped only under the `-preview` id) rewrote it to `qwen3.8-max-preview` before dispatch. The upstream rejects the preview id with `401 Model qwen3.8-max-preview is not supported`.

The provider-aware exemption in `resolveModelAlias` (#11503) never fired here: the opencode-go static registry didn't declare the GA id, so `hasKnownProviderModel("opencode-go", "qwen3.8-max")` returned false and the global rewrite applied.

## Changes

- Declare `qwen3.8-max` in the opencode-go registry. Qwen models on opencode-go are text-only and route through the Claude translator (#2292, #2822); the new entry follows that. Base id only, upstream doesn't advertise effort-tier variants.
- Give the GA model its own `MODEL_SPECS` row instead of aliasing it to the preview spec (same limits: 1M context, 65k output).

The built-in rewrite stays for providers that only serve the preview id (qoder, bailian-coding-plan, qwen-cloud-token-plan) and for callers with no provider in hand.

## Verification

- New test `tests/unit/14181-opencode-go-qwen-ga-alias.test.ts`, 4 cases: the registry declares the GA id with the qwen family flags; `resolveModelAlias("qwen3.8-max", "opencode-go")` returns the id unchanged; preview-only providers (qoder) and no-provider callers still resolve to the preview id; the GA id resolves to its own spec row.
- Red-green checked against base: with the src changes reverted and only the test kept, 3 of 4 cases fail (the qoder control passes); with the fix, 4/4 pass.
- Adjacent suites: alias/deprecation/registry tests 29/29, qwen and model-specs area tests 77/77.
- `prettier --check`, `eslint`, `typecheck:core` clean on all touched files.

Fixes #14181


diegosouzapw:

Issue #13461, opened 2026-09-12
[BUG] Gateway returning bad response


arjav1181:
### OmniRoute Version

3.8.50

### Installation Method

Docker / Docker Compose

### Operating System

Linux

### OS Version

Ubuntu 24 LTS

### Provider(s) Involved

Perplexity web, pollinations and some more

### Model(s) Involved

auto/best-coding

### Client Tool

Opencode

### Description

Ok so what is happened is that i was using opencode theough the omniroute gateway hosted on a hugging face space, suddenly it said go to perplexity.ai and sign up and then finished, second my pollination credits ran out so it again said out of credits then finished and this type of issue is with some other providers also. So what i think is happening here is that when there is a problem with the provider like out of credits, or expired cookies they send a error to omniroute but with status code 200 and a clean body so what omniroute does then is froward it to me and thinks that the work is done, so i think the fix is to add erroe parsing with 200 request codes also.

### Steps to Reproduce

1. Go to perplexity web or pollinations ai then set them up with expired credentials or out of credits key
2. Send a request to the pollinations or perplexity web provider

### Expected Behavior

The request should come with a 200 status code but will display error in the content like out of credits or sign in at perplexity.ai

### Actual Behavior

Same as expected behaviour

### Test Impact

Unsure

### Error Logs / Output

```shell

```

diegosouzapw:
Thanks for reporting! Fixed on the `release/v3.8.51` **branch** (PR #13910). Note that `v3.8.51` has **not been published yet** — `npm view omniroute version` is the authority on what is actually released, and it will not show `v3.8.51` until the release closes. To test the fix before then you have to build from that branch; otherwise it reaches you when `v3.8.51` publishes.

PR #14243, opened 2026-09-20
fix(sse): treat antigravity empty completions with a normal stop as valid 200s (#14160)

### LINKED ISSUE #14160: fix(resilience): empty gemini completion is treated as provider failure: 502 + model lockout drains the antigravity pool
an empty (but successful) completion from antigravity gemini gets treated as a provider failure: the client gets `502 Provider returned empty content` and the model gets locked out on that account, which drains the pool for every other client

## my setup

- omniroute 3.8.51 (`:next`, docker), 22 active antigravity connections
- combo `gemini38high` (priority): `antigravity/gemini-3.8-flash-tiered` → `gemini-3.7-flash-tiered` → `gemini-3.1-flash-lite` → `gemini-pro-agent`
- two kinds of clients on the same pool: codex agents with huge `/v1/responses` sessions, and claude-mem (a background memory observer) calling `/v1/chat/completions` through the openai-compatible endpoint

## what happens

claude-mem sends a single user message, non-streaming, `max_tokens: 4096`, `temperature` set, no tools, no system message (~14k chars, starts with "Hello memory agent, you are continuing to observe the primary Claude session")

for that prompt the model very often has nothing to say and returns an empty completion

from `call_logs` for one day, grouped by prompt:

| request shape | 200 | 502 empty content |
|---|---|---|
| that single-message observer prompt | 17 | 397 |
| every other request from the same api key (multi-message, same combo, same accounts) | 369 | 0 |

so it is not flaky accounts or a broken translation in general, it is one prompt that legitimately produces no text ~96% of the time

each failure takes 2-9s and zero tokens

## why it hurts

every one of those is logged as a failure against the account + model:

```
[antigravity/gemini-3.8-flash-tiered] Provider returned empty content (reset after 5m)
Model-only lockout for antigravity:gemini-3.8-flash-tiered — 502 server_error 20s (failureCount=3, connection stays active)
Account e2afe858... unavailable (502), trying fallback
```

~400 of these a day is enough to keep most of the pool excluded

selection logs for the codex agents show things like `active=22, excluded=15` and `quota-aware: 3 with quota`, and those agents then end in `503 Maximum combo retry limit reached` after 8-9 minutes of walking accounts

a cheap harmless request is starving the expensive ones

## what i would expect

an empty completion with a normal finish reason is a valid answer, not an upstream outage

at minimum it should not count toward account/model lockout or the breaker

ideally the empty response is just returned to the client as `200` with empty content (maybe behind a setting), since retrying the same prompt on the next account gives the same empty answer anyway

the fake-success check in `open-sse/handlers/chatCore.ts` (`isEmptyContentResponse` → `persistFailureUsage(502, "empty_content")` → family fallback) makes sense for web-scraping providers that return an empty shell, but for a first-party api like antigravity/gemini an empty candidate with `finishReason: STOP` is a real result

## what i do not know yet

the call log artifact has `responseBody: null` for these, so i have not confirmed what the upstream candidate looks like (truly empty parts vs thought-only parts vs a block reason)

happy to capture a raw upstream response if you tell me the best flag for it

related but different: #13600 (empty assistant response after tool_calls, streaming), #7037 (antigravity non-streaming tool-only responses)

xiechimon:
## What

An empty (but successful) completion from antigravity's Gemini is treated as a provider failure: the client gets `502 Provider returned empty content` and the model is locked out on that account. The reporter sees ~400 of these a day from a single observer prompt the model legitimately answers with no text, which keeps most of a 22-connection pool excluded and starves unrelated clients (their words: a cheap harmless request was starving the expensive ones).

The fake-success guard exists for free-tier/scraping providers whose failure mode is an empty 200 shell (#13461). Antigravity is Google's first-party Gemini API: an empty candidate with `finishReason: STOP` is a real result, and retrying the same prompt on the next account returns the same empty answer anyway.

## Changes

- `isEmptyContentResponse()` gains an optional `provider` and a `TRUSTED_EMPTY_STOP_PROVIDERS` allowlist (same pattern as the provider-allowlisted `classifyFakeSuccessBody` from #13461). On antigravity, an empty completion with a normal terminal stop reason (openai `stop`, claude `end_turn`) is no longer flagged.
- The non-streaming leg passes its `provider` through, so these completions flow to the client as a valid 200 with empty content instead of a synthetic 502. No failure accounting, no model lockout, and no retrying the same prompt on the next account to get the same empty answer.

Unchanged: every other provider keeps the guard, including on empty stop completions; antigravity responses with no terminal stop reason are still flagged; the existing legit-empty carve-outs (`length`, `tool_calls`, `content_filter`, `max_tokens`, `tool_use`) are untouched.

## Verification

- New test `tests/unit/14160-antigravity-empty-stop-completion.test.ts`, 5 cases: antigravity empty+stop and empty+end_turn pass through; untrusted providers and the no-provider call keep flagging them; an antigravity empty shell with no finish reason is still flagged; existing carve-outs unchanged.
- Red-green checked against base: with the src changes reverted and only the test kept, the 2 antigravity cases fail and the 3 controls pass; with the fix, 5/5 pass.
- Adjacent suites: 199/199 across the error-classifier, non-streaming-leg, empty-content, fake-success, and silent-empty web-provider suites (incl. `combo-empty-content-failover-5085`, `diagnostics-fake-success-13461`, `copilot-m365-web-silent-empty-7858`, `zai-web-silent-empty-repro`).
- `prettier --check`, `eslint`, `typecheck:core` clean on all touched files.

Out of scope: the streaming variant (#13600) is a different code path (`streamEmptyChoices.ts`) and is not touched here. If you'd rather have the passthrough behind a settings toggle, happy to add one.

Fixes #14160


diegosouzapw:
This is a well-scoped fix — reusing the same trusted-provider-allowlist shape as classifyFakeSuccessBody rather than inventing a new mechanism, and keeping every other provider's guard untouched. The 5-case test (both wire shapes, an untrusted-provider control, and the no-terminal-reason control) covers exactly the boundary that matters. Good root-cause tie-back to #13461's original intent too.

Optional follow-up question: `gemini` / `gemini-cli` go through the same `geminiResponse.ts` aggregation and hit the same finish-reason gap — would you include them in `TRUSTED_EMPTY_STOP_PROVIDERS`, or keep this scoped to antigravity on purpose?

xiechimon:
I checked the two providers separately.

`gemini-cli` is discontinued and not routable (tokenRefresh.ts:142: "The gemini-cli provider was discontinued and is not routable"), so a whitelist entry for it would be dead config. I would leave it out.

`gemini` (registry id `gemini`) is the real case. It routes its non-streaming SSE through the same parseSSEToGeminiResponse aggregator as antigravity (nonStreamingSse.ts:35-36 maps both FORMATS.GEMINI and FORMATS.ANTIGRAVITY to it), and that accumulator defaults finishReason to "stop" (geminiResponse.ts:225). An empty-string text part sets sawContent but leaves textContent empty, so it emits the same empty-content + finish_reason "stop" 200 that triggered #14160 for antigravity. Since `gemini` is first-party Google, the allowlist's own rationale (a first-party empty stop is a real answer, not a scraping-provider shell) applies to it just as much, so I would include it.

One nuance worth confirming: the allowlist matches on the provider id, not the wire format, so adding "gemini" does not loosen the guard for the other format: "gemini" entries (hcnsec, vertex, vertex/partner). Those stay guarded, which is what we want for third-party gateways.

If that matches your intent, I will add "gemini" to TRUSTED_EMPTY_STOP_PROVIDERS and mirror the test: a gemini empty-stop case asserting a valid 200, plus an hcnsec control (same wire format, untrusted) asserting it still gets flagged. Otherwise I am fine keeping this antigravity-only and widening on a confirmed gemini report.

diegosouzapw:

PR #14005, opened 2026-09-17
fix(db): never TRUNCATE-checkpoint a live WAL (SIGBUS under traffic)

HouMinXi:
Refs #13973.

Long-running servers died about every six hours with SIGBUS (exit 135), no JS stack. A coredump put the fault on the mapped `storage.sqlite-shm` (WAL-index). The journal showed `wal_checkpoint(TRUNCATE)` one or two seconds before each death.

## Cause

`walMaintenance.ts` ran `wal_checkpoint(TRUNCATE)` on a six-hour timer **and** from the 256 MB size guard. TRUNCATE rewrites the wal-index while other handles still have it mapped. A later read of that mapping is SIGBUS; JS cannot catch it.

No interval makes a live TRUNCATE safe.

## Change

- Drop the periodic TRUNCATE scheduler. PASSIVE checkpoints still run every five minutes; a busy tick records telemetry and retries after 60s.
- Size guard (`OMNIROUTE_WAL_GUARD_MAX_MB`, default 256) now runs `wal_checkpoint(RESTART)` instead of TRUNCATE or a warning-only log. RESTART starts a new WAL without rewriting the mapped index, so the file stays bounded.
- Shutdown still truncates (`closeDbInstance`).
- `OMNIROUTE_WAL_TRUNCATE_INTERVAL_MS` is ignored; a positive value logs a one-time deprecation warning.

This implements one of the three items on #13973 (stop live TRUNCATE). It does not stagger the other 6h timers and does not add the `conversation_turn_nodes(last_seen_at)` index, so the issue stays open.

## Tests

WAL suites: 38 pass, 0 fail, 1 pre-existing bun:sqlite skip. Injecting the old warn-only guard fails the RESTART assertion; restoring it passes.

## Note

`OMNIROUTE_STRIP_SYSTEM_PREAMBLE` / `COMBO_LOOP_SAFETY_TIMEOUT_MS` env-doc drift is on the base tip (#14004 / #14054), not this PR.

---

⚠️ base-red inherited: #13866 — the 16 unit-test failures reproduce on the clean tip `7cc454d9` (6/6 sampled) and are identical across PRs with disjoint content; none touches what this PR changes.

PR #14244, opened 2026-09-20
fix(db): avoid live WAL TRUNCATE after incremental reclaim

Bl0ck154:
## Summary

`reclaimFreedPages()` still ends an incremental cleanup pass with `wal_checkpoint(TRUNCATE)` even though #14005 removed live TRUNCATE from WAL maintenance because changing the live WAL geometry can invalidate mapped handles and crash with SIGBUS.

This switches the cleanup-path final checkpoint to `PASSIVE`. The periodic WAL maintenance path already owns WAL high-water management, while PASSIVE is enough to fold reclaimed pages back without mutating the live WAL file geometry.

## Why

- #14005 explicitly establishes that live `TRUNCATE` is unsafe under concurrent handles.
- `reclaimFreedPages.ts` predates that change and still retained a live `TRUNCATE` call.
- The cleanup scheduler runs while the server is live, so it can overlap with other readers/handles.

## Validation

- `npx tsx --test tests/unit/db/cleanup-reclaim-freed-pages.test.ts`
- 8/8 tests pass.
- The existing INCREMENTAL regression still confirms `page_count` shrinks after reclamation with a final PASSIVE checkpoint.


diegosouzapw:
Thanks @Bl0ck154 — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

Issue #11128, opened 2026-08-22
[feature] ChaosSearch — extend chaos/fan-out combo dispatch to search providers


diegosouzapw:
## Feature Request

**Source:** Discussion #11112 by @psylligent

## Problem
Chaos mode (`open-sse/services/autoCombo/chaosEngine.ts`) only dispatches **chat** combo targets. Search-capable providers are a separate routing surface (web-search tool-call rewriting, cf. #9772), so there is no way to fan-out/race search providers the way chaos races chat models.

## Proposed Solution
Treat search-capable targets as first-class combo members so chaos (and potentially other fan-out strategies) can dispatch across them — "ChaosSearch": race multiple search providers and return the best/fastest result set.

### Implementation Ideas
- `open-sse/services/autoCombo/chaosEngine.ts` — currently assumes chat-shaped targets/results
- `open-sse/services/combo.ts` — strategy dispatch prelude
- Web-search tool rewriting surface (see #9772 for how custom search providers are wired today)

### Current Workarounds
None — search routing is 1:1 per request.

## Additional Context
- Discussion: https://github.com/diegosouzapw/OmniRoute/discussions/11112
- Related: #9772 (custom search provider rewriting)

diegosouzapw:
Analyzed and catalogued in the feature backlog — planned for an upcoming cycle. Verified against the code: chaos fan-out currently dispatches every panel member through the single-model chat pipeline (chaosEngine.ts wraps handleSingleModel and emits OpenAI-style SSE chunks), while search providers sit on a separate surface (webSearchRouting.ts layer-2 per-tool routing + webSearchFallback.ts /v1/search fallback). Racing search targets as combo members is therefore a real gap today — thanks @psylligent for the discussion thread. We'll scope it as a search-target adapter for the chaos/fan-out dispatcher. Closing to keep the tracker lean; the implementing PR will reference this issue.

PR #14248, opened 2026-09-20
fix(sse): skip +1s cooldown re-send for no-fresh-sibling burst-throttle 429 (#11128-followup)

adevwithpurpose:
### Problem
DSH achieves success on a throttled CodeBuddy account by re-sending late: the client's *spaced retry* lands after the upstream bucket refills. But when all credentials are rate-limited and the provider-execution pipeline fail-fasts (the 11128 burst throttle is time-based, so rotating to a cooling sibling cannot succeed), OmniRoute then saw the fail-fast 429 as an *upstream quota signal* and scheduled a +1s COOLDOWN_RETRY re-send. That re-send:
1. is a locally-synthesized restatement of an upstream response the upstream already gave, and
2. lands 0.5-1s into a ~60s throttle window, so it can never succeed — it only triples per-request latency.

### Fix
Tag the fail-fast burst-throttle 429 with a noFreshSiblingThrottle marker and suppress the local cooldown re-send, so the pipeline returns the restated 429 to the client immediately and the client's own spaced retry becomes the next attempt.

- ChatCoreErrorResult gains noFreshSiblingThrottle? (src/lib/skills/toolLoopTypes.ts)
- Pipeline fail-fast tags the outcome (open-sse/handlers/chatCore/providerExecutionPipeline.ts)
- Leg bridge propagates the marker (it rebuilds the error result inline — without this the field would be dropped)
- getCooldownAwareRetryDecision accepts suppressLocalCooldownRetry -> shouldRetry:false (src/sse/services/cooldownAwareRetry.ts)
- chat.ts captures the flag at all 4 error-set sites, passes it into the allRateLimited gate, logs COOLDOWN_RETRY_SKIP, and falls through to return the restated 429

### Proof / verification
- npm run typecheck:core exit 0
- New unit test tests/unit/cooldown-retry-suppress-local-throttle.test.ts (4/4 green): suppression forces no-retry even with budget/Retry-After due; flag absent retains legacy retry behavior
- Field propagation verified: pipeline -> bridge -> open-sse/handlers/chatCore.ts L5147 `return err` -> execution.result in chat.ts
- VPS live-trace (Tab 1, extension-based E2E) to follow


diegosouzapw:
The fix itself (commit `455f7355` — skip the +1s cooldown re-send for a no-fresh-sibling burst-throttle 429) is right, and it's the follow-up #11128 needed. The problem is the branch: it carries **43 commits from a separate long-lived fork branch (`release/v3.8.51-astra`)** — a whole new provider (WorkBuddy/cbai OAuth), a Codex CLI bump, the Astra Copilot catalog, a compression-worker rewrite — none of which belong to this PR, and they conflict with the release tip.

Could you open a fresh branch off `origin/release/v3.8.51` with only `455f7355` cherry-picked (`git cherry-pick 455f7355`), and push it here (or as a new PR)? That's a one-commit, ~1-file change and it merges the same day. Closing nothing on my side — this PR stays open for you to repoint. Thanks!

PR #13930, opened 2026-09-17
fix(deepseek-web): parse DeepSeek V4 native DSML markup and bare parameter tool calls

EkoSaputro14:
### Summary
This PR fixes tool-calling failures and text corruption when using `deepseek-web/deepseek-v4-flash` or `deepseek-web/deepseek-v4-pro`.

### Root Cause
DeepSeek V4 (particularly on chat.deepseek.com) natively emits **DSML (DeepSeek Markup Language)** markup or hybrid ad-hoc tags during tool invocations:
1. **DSML Tag Leaks:** The model emits `<｜｜DSML｜｜ calls>`, `<｜｜DSML｜｜ invoke name="...">`, and `<｜｜DSML｜｜ parameter name="...">` using fullwidth vertical bars (`｜`). OmniRoute's `TAG_TOKEN_RE` only looked for `<tool>` and `<tool_call>`, causing the entire DSML block to be treated as plain message content.
2. **Web Disclaimer Leak:** DeepSeek appends `"This response is AI-generated, for reference only."` at the end of outputs.
3. **Hallucinated Closing Parameter Tag:** The model frequently produces valid JSON followed by an invalid closing tag: `<tool>{"name": "...", "arguments": {...}}</parameter></tool>`. Because of the trailing `</parameter>`, `JSON.parse` failed with SyntaxError.
4. **Bare Parameter Blocks:** On long agentic trajectory prompts (e.g. 30k+ tokens in coding agents), the model occasionally skips the enclosing `<tool>` or `<invoke>` wrapper and outputs bare `<parameter name="...">...</parameter>` blocks directly.

### Solution
In `open-sse/translator/deepseekWebTools.ts`:
- Stripped web disclaimer string.
- Stripped outer DSML envelope tags (`<｜｜DSML｜｜ calls>`, etc.).
- Normalized `<invoke>` tags into `<tool>` and `<parameter>` tags into canonical parameter format.
- Stripped trailing hallucinated `</parameter>` closing tags right after JSON closing braces (`}`).
- Auto-wrapped bare `<parameter>` sequences into `<tool>...</tool>` so that OmniRoute's existing schema-based nameless-block matcher can map the parameter to the requested tool.

### Testing & Verification
Tested against live prompts with Hermes Agent:
- Verified parallel tool calling with `<｜｜DSML｜｜ invoke>` (e.g. `vision_analyze` + `terminal`).
- Verified bare `<parameter name="code">` correctly maps to `execute_code`.
[...]

PR #14249, opened 2026-09-20
fix(translator): parse DeepSeek Web DSML tool calls

### LINKED ISSUE #14208: fix(): deepseek-web doesn't support format of commands
| Field | Value |
|---|---|
| **OmniRoute Version** | `3.8.50` |
| **Installation Method** | npm (global) |
| **Operating System** | Windows |
| **OS Version** | 11 Pro 25H2 |
| **Node.js Version** | `22.22.3` |
| **Provider(s) Involved** | `deepseek-web` |
| **Model(s) Involved** | `ds-web/DeepSeek V4 Pro` |
| **Client Tool** | DeepSeek Harness |
| **Deepseek Harness** |   `0.1.1-rc.2` |

### Description

When routing a DeepSeek Web model (ds-web/DeepSeek V4 Pro) through OmniRoute, tool-call responses are sometimes not parsed and executed. Instead of invoking the requested tool, OmniRoute returns the raw, unparsed text containing DSML markup to the user.

The model produces an otherwise valid tool call (e.g. write / create_file with JSON arguments), but it is immediately followed by malformed / corrupted pseudo-tags such as <｜｜DSML｜｜ parameter> and </｜｜DSML｜｜ invoke> instead of a clean closing tag. Because of that trailing garbage, the parser rejects the whole block, silently drops the tool call, and shows the raw tagged text to the user — so the file is never created.

This looks closely related to the already-open PR #13226 ("Malformed tool-call replies"), but the reproduction below uses a different tool name (write), a non-ASCII (Russian) content payload, and a Windows-style file path, which may exercise additional edge cases in the parser.

### Steps to Reproduce

1. Configure OmniRoute with the DeepSeek Web provider (ds-web) and select model ds-web/DeepSeek V4 Pro.
2. Connect a harness / agent client that exposes a write (or create_file) tool.
3. Send the following request to the agent:

`create 1.txt with random text`

4. The model responds with a tool call intended to write 1.txt, but the response contains malformed DSML tags.
5. Observe that the file 1.txt is not created on disk, and the raw DSML text is displayed to the user.

### Expected Behavior

OmniRoute should parse the tool call, extract the JSON arguments (file_path, content), and forward a valid tool_calls payload to the harness so that the write tool is actually invoked and 1.txt is created. The user should receive the tool result, not raw markup.

### Actual Behavior

OmniRoute fails to parse the tool call and returns the raw, unparsed DSML text to the user. The file is not created. The output looks like this:

```
<｜｜DSML｜｜ calls> <｜｜DSML｜｜ invoke name="write"> <｜｜DSML｜｜ parameter name="file_path" string="true">D:\Projects\maxnadeev.ru\1.txt</｜｜DSML｜｜ parameter> <｜｜DSML｜｜ parameter name="content" string="true">Однажды в четверг, когда чайник уже второй раз закипал без причины, рыжий кот по имени Пельмень решил, что пора основать собственную метеостанцию. Прогнозы он давал простые: если спит на подоконнике — ясно, если на клавиатуре — переменная облачность с вероятностью случайных символов. Соседский голубь с этим не соглашался и настаивал на синоптическом подходе, но его аргументы тонули в ворчании холодильника.

К вечеру над городом прошёл дождь из чисел: 7, 42, 13, 88, 3, 17. Никто не удивился — все давно привыкли, что понедельники бывают странными. Дворник собрал несколько чисел в ведро и обещал вернуть их владельцам, если те докажут, что числа действительно их.

Так и закончился этот день: тихо, нелепо и вполне себе случайно. </｜｜DSML｜｜ parameter> </｜｜DSML｜｜ invoke> </｜｜DSML｜｜ calls>
```

### Test Impact

Unsure

### Error Logs / Output

```shell
OmniRoute version: 3.8.50
Node version:      22.22.3
OS:                Windows 11 Pro 25H2
Provider:          deepseek-web (ds-web)
Model:             ds-web/DeepSeek V4 Pro
Deepseek Harness:  0.1.1-rc.2
Harness/tool:      write (or create_file)

Raw model response (unparsed):
<｜｜DSML｜｜ calls> <｜｜DSML｜｜ invoke name="write"> <｜｜DSML｜｜ parameter name="file_path" string="true">D:\Projects\maxnadeev.ru\1.txt</｜｜DSML｜｜ parameter> <｜｜DSML｜｜ parameter name="content" string="true">...</｜｜DSML｜｜ parameter> </｜｜DSML｜｜ invoke> </｜｜DSML｜｜ calls>

Result: tool call dropped, file not created, raw text shown to user.
```

### Screenshots

Image

### Additional Context

- The issue appears to be a parsing/compatibility problem between the DSML markup emitted by DeepSeek Web and OmniRoute's tool-call parser, not a harness-side error.

- The corrupted tags use double ASCII pipes (<｜｜DSML｜｜ ...>) rather than the canonical full-width delimiter, which may be a tokenization artifact from the DeepSeek Web session.

- The content argument is a multi-line, non-ASCII (Cyrillic) string containing spaces, punctuation, and newlines, and file_path is a Windows path with backslashes. These may be additional stressors for the JSON/loose-object parser.

- This closely matches the description in PR #13226 ("Malformed tool-call replies"), which mentions a valid create_file JSON body followed by corrupted pseudo-tags such as <｜｜DSML｜｜ parameter> and </｜｜DSML｜｜ invoke>. Please confirm whether that fix covers the write + Cyrillic-content + Windows-path case described here.

- Related references:

    - PR #13226 — malformed tool-call replies / trailing DSML garbage.

    - PR #12905 — parse/scrub DSML tool-call markers.

    - Issue #4712 — DeepSeek-web tool-result (role: "tool") loss (fixed in v3.8.35).

### Validation Plan

_No response_

lorenzozanee:
## Summary

Parse DeepSeek Web's native DSML `invoke`/`parameter` responses into OpenAI-compatible tool calls, preserving Windows paths and non-ASCII argument content.

## Related Issues

Fixes #14208

## Validation

Choose change type focused loop [Contribution Golden Path](../docs/ops/CONTRIBUTION_GOLDEN_PATH.md). The full unit suite, Vitest, 60% coverage gate, and production build run in CI on PR (#8329):

- [x] Focused tests category gates golden path
- [ ] `npm run lint`
- [x] Reconciled current active release base; focused checks rerun afterward
- [x] Production-code changes include new or updated automated test in PR
- SonarQube is temporarily opt-in while private project has no quota; not PR gate.

## Tests Added Or Updated

- `tests/unit/deepseek-web-tools-dsml.test.ts` — covers native DSML calls with Windows paths and non-ASCII content.
- `node --import tsx/esm --test tests/unit/deepseek-web-tools-dsml.test.ts`
- `node --import tsx/esm --test tests/unit/deepseek-web*.test.ts`
- `npm run check:open-sse-typecheck`
- ESLint, Prettier, and `git diff --check` passed for the changed files.

## Coverage Notes

- The focused regression and DeepSeek Web test suites cover the translator change. No coverage movement was recorded for this change.

## Reviewer Notes

- The parser is scoped to DeepSeek Web's native full-width DSML envelope; the existing generic parser and other provider adapters are unchanged.
- Upstream CI, live DeepSeek behavior, Windows runtime behavior, and maintainer intent remain unverified.


diegosouzapw:
Thank you for tackling the DSML parsing gap from #14208 — the tokenizer-based approach with proper depth tracking is a nice design. Testing it against the exact string from the bug report, though, the DSML_TAG_RE pattern only matches the closing-tag form `<｜｜DSML｜｜ /parameter>` (slash after the marker); the real upstream output — and the issue's own repro — uses `</｜｜DSML｜｜ parameter>` (slash before the marker), which this regex never matches, so the original bug report still fails to parse cleanly with this change (file_path ends up swallowing the rest of the payload). #13930 already lands a more permissive normalization that handles both tag-slash placements and is ahead in the queue, so recommending closing this one in its favor — but I'd genuinely welcome a follow-up that ports your depth-tracking tokenizer onto that broader tag matching if you want to keep pushing on it.

Issue #10880, opened 2026-08-20
feat(resilience): lock all connections of a shared egress IP together on IP-bucketed 429


maxmad64bis:
### Problem / Use Case

Some providers bucket their free-tier quota by egress IP — the maintainer
himself documented this for opencode in #9611 ("the opencode free tier is
IP-based, not account-based"). When several connections share one egress IP,
they all draw from the same budget, and when it's gone the upstream answers
429 to every one of them.

Today the lock is per connection (`rateLimitedUntil` in `markAccountUnavailable`,
`src/sse/services/auth.ts`). So with N connections on the same IP, the first
429 locks one connection, the next request rotates to a sibling on the same IP
and burns another guaranteed-to-fail upstream call — N-1 wasted calls per
episode, each adding latency and load with zero chance of success.

Same shape as #10460 (three accounts, three guaranteed-fail 400s), which was
fixed in #10525: the signal is scoped to the real unit, not to the account.
An IP-bucketed 429 is scoped to the egress IP.

### Proposed Solution

Question first: **when several connections share an egress IP and the
provider's quota is IP-bucketed, should a 429 lock them all together instead
of one per request?**

If that's not a case worth supporting, close this — it just means operators
are expected to guarantee one distinct IP per account (the #9611 position),
and shared-IP pools are unsupported.

If it is, the smallest change that fits the existing design:

1. On `quota_exhausted` 429 (`checkFallbackError` → `markAccountUnavailable`),
   look up the connection's recent egress IP in `proxy_logs` (`egress_ip` is
   persisted per request, `src/lib/proxyLogger.ts`) and apply the same
   temporary cooldown to every connection seen on that IP.
2. Gate it behind the existing `HONORS_RULE_LOCK_SCOPE_PROVIDERS` allowlist
   (`open-sse/config/providerErrorRules.ts`) — providers opt in, everyone
   else keeps today's behavior byte-for-byte (same pattern as #10419).
3. Never terminal: temporary cooldown only (`rateLimitedUntil` +
   `testStatus: "unavailable"`), same self-recovering semantics as
   `isAgentrouterConnectionQuotaScope()`. No permanent state from an IP-level
   signal.

This doesn't fake a fallback or mask a real upstream 429 — the original 429
still surfaces, siblings are locked because they share the exact unit the
upstream limits.

### Alternatives Considered

- **Configuration only (#9611's current answer):** one distinct proxy IP per
[...]

Issue #10677, opened 2026-08-18
feat(resilience): Surface anonymous egress-IP sharing in the proxy health sweep and the egress API


maxmad64bis:
## Problem / Use Case

OmniRoute can already detect a dangerous situation: several accounts of the same rotation group exiting through one egress IP. `analyzeEgressSharing()` (src/lib/proxyEgress.ts:154) groups egress observations by IP and flags IPs shared by two or more accounts of the same rotation group — the condition behind codex anomaly revocation. Yet it has a single consumer: the management-authenticated `GET /api/settings/proxies/egress` endpoint (src/app/api/settings/proxies/egress/route.ts).

Nothing routine surfaces the result. The periodic proxy-health sweep does not report it, and the process log omits IPs and account prefixes by default since #10348/#10539 (src/lib/proxyLogger.ts). An operator cannot tell whether egress diversity is real without manually curling the authenticated endpoint.

Why it matters: proxy pools are meant to spread requests across distinct egress IPs, but proxies can share a real IP without the operator knowing. When that happens, a rate limit hits every account behind the shared IP at once, and the pool's announced diversity is illusory.

## Proposed Solution

1. Extend the existing `GET /api/settings/proxies/egress` response with an anonymous summary, leaving the current payload unchanged: distinct egress-IP count, sharing counts per rotation group, and the largest number of accounts sharing one IP. No IP literals in the summary; the raw map stays behind management auth.
2. Log a one-line anonymous summary in the periodic proxy-health sweep (src/lib/proxyHealth/scheduler.ts), e.g. `egress: <G> rotation group(s) share an egress IP (max <N> accounts)`. No IPs and no account identities by default; `PROXY_LOG_INCLUDE_IPS=true` remains the only raw-detail opt-in.
3. Compute the aggregate from recent `proxy_logs` rows over a rolling window — `egress_ip` is always persisted, even when the log line is redacted — without probing connections live.

## Alternatives Considered

- A dashboard card showing raw IPs: contradicts the redaction decision implemented in #10348 and merged in #10539, and UI strings cost 43 locale files.
- Documenting the curl call: the signal stays off every routine path.
- Probing each connection's egress IP in the sweep: expensive, and the data already exists in `proxy_logs`.

## Acceptance Criteria

- `GET /api/settings/proxies/egress` returns the anonymous summary; the existing payload shape is unchanged (additive change).
[...]

PR #14250, opened 2026-09-20
feat(db): add per-proxy egress IP lookup by host and port

maxmad64bis:
⚠️ base-red inherited: #13866
## Summary
Two pool members can leave through the same address, so the pool counts N members but serves fewer exits. The lookup normalized only the query side while `proxy_host` was stored raw, so a stored mixed-case or bracketed host never matched. This normalizes the host at write time in the proxy logger, keeps the per-proxy reader next to its connection-level sister, and shows each pool member's last seen egress address next to the pool totals.
## Related Issues
- Related to #6365 (pool rotation promise), #10677 (anonymous sharing summary, per-connection only), #10880 (per-connection exit lockout on IP-bucketed 429), #13460 (pool observation as useful diagnostic once fixed, not the fix). Follow-up of closed-unmerged #14250 (owner conditions 2026-09-21: consumer in-PR, write-side normalization, linked symptom — this body is the link).
## Validation
- [x] Change type: DB
- [ ] `npm run lint` — ESLint on the touched files is clean; the full run is red on the base (#13866), unrelated to this diff
- [x] Reconciled with the current active release base; focused checks rerun afterward (`668061c3`, 0 behind, gates re-run at this head)
## Tests Added Or Updated
- `tests/unit/proxy-logs-egress-ip-by-proxy.test.ts` (extended 8 cases: divergent stored `Proxy.Example.COM` / `[::1]` resolve after write-time normalization, idempotent canonical round-trip, latest wins, unknown pair null, port/host isolation, NULL-egress skip, 48 h window, empty/invalid guard)
- `tests/unit/proxy-pool-member-egress-route.test.ts` (new, 6 cases through the call-site: coincident members same key, member without rows null, stale absent, session-gateway limit written, `GET /pool` parity, flag-off/fail null)
- `tests/unit/ui/PoolMemberEgressLines.test.tsx` (new, 4 cases: renders lines, null renders placeholder, null body renders nothing, empty pool renders nothing)
- Neighbors re-run green: `proxy-logs-egress-ip` (5), `proxy-logs-egress-lookup-10880` (4), `egress-ip-lock-10880` (17), P6 observation route (7) + lib (8), P6 component (5), `i18n-vi-completeness` (5)
## Coverage Notes
- `src/lib/db/proxyLogs.ts` reader covered via the route tests (call-site, not helper-alone); `src/lib/proxyLogger.ts` normalization covered by the divergent stored-vs-query cases; `src/lib/proxyPoolEgressObservation.ts` member reader covered by the 6 route cases.
- No coverage movement beyond the addition; file-size rebaseline is mount-point only.
## Reviewer Notes
- File-size gate: rebaseline `_rebaseline_2026_09_21_14250_member_egress_lines` (`ProxyRegistryManager.tsx` 1477→1479, import + mount only, irreducible, covered by the route + component tests). Complexity gate: clean after extracting member-cache and per-member collection helpers (was 18, now 0 new violations). `check-api-typecheck` is clean on owned files (283 pre-existing errors within frozen baseline). Failing-gate file lists outside this diff.
- `PROXY_POOL_EGRESS_OBSERVATION` (default off) gates both the pool observation line and the new per-member lines: operators with P6 already on will see the member lines appear with no separate opt-out. No migration and no index: legacy rows keep their spelling and age out of the 24 h window, and the recorded plan is a window-bounded scan via `idx_pl_timestamp` like the sister helper.
- CI reds are inherited from the red base (#13866), not from this diff: every file cited by the failing gates is outside this diff. Non-blocking for this PR.


diegosouzapw:
Closing this one for now — not because the idea is wrong, but because as shipped it's scaffolding: `getRecentEgressIpForProxy(host, port)` has no caller anywhere in `src/` or `open-sse/`, and there's no linked issue describing who needs it. We don't merge helpers that nothing exercises, since they rot silently.

There's also a latent correctness gap worth fixing before it ever gets a caller: the lookup normalizes the *query* side (lowercases the host, strips IPv6 brackets) but `proxy_host` is stored raw at insert time (`src/lib/proxyLogger.ts` — only trimmed by the Zod schema, never lowercased). The new test passes only because its fixture is already normalized; a stored `Proxy.Example.COM` or `[::1]` would never match. Normalization has to happen on both sides, or on the write side only.

To get this accepted, the PR should:
1. Ship the consumer in the same PR (the proxy-health decision, the dashboard proxy row, wherever the recent egress IP is actually needed) so the helper has a real call site and a test that goes through it.
2. Normalize `proxy_host` consistently — ideally at write time in `proxyLogger.ts`, with a one-line migration note if existing rows need it — and add a test whose stored host differs in case/bracket style from the query.
3. Link the issue or describe the operator-visible symptom in the body.

Your other proxy PRs in this batch are excellent, so this is clearly a quick turnaround if you want it. Thanks.

maxmad64bis:
Superseded by #14364 (reopen refused after force-push: `422 state cannot be changed` on `PATCH state=open` and on `reopenPullRequest`, so a new PR carries the work). #14364 answers the three conditions above in one PR: the helper ships its dashboard consumer in the same PR with tests through the call-site, `proxy_host` is normalized at write time with a divergent stored-vs-query test, and the operator-visible symptom plus related issues are in the body.

PR #14260, opened 2026-09-20
fix(providers): stop treating Anthropic max_tokens as the context window

HouMinXi:
## Summary

`normalizeDiscoveredModels` treated Anthropic `max_tokens` as the context window. That field is the output cap. The window is `max_input_tokens`. Claude Opus 5 then showed 128K on `/v1/models` instead of 1M.

The window list now reads `max_input_tokens` / `maxInputTokens`. `max_tokens` is an output-limit candidate, next to `max_output_tokens`.

OpenRouter `context_length` and vLLM `max_model_len` still win when a record has them. Already-synced connections keep 128K until they re-sync.

## Test plan

- `node --import tsx/esm --test tests/unit/anthropic-discovery-window-max-tokens.test.ts tests/unit/vllm-max-model-len-12858.test.ts tests/unit/openrouter-context-length-3202.test.ts` — 16/16.
- Inject: put `record.max_tokens` back on the window list. The three Anthropic-shape tests fail (`128000` vs `1000000`). Restore → 5/5.

⚠️ base-red inherited: #13866


diegosouzapw:
Thanks @HouMinXi — merging via the release merge-train. Validated in local merge-train (mt-train10c) on the devbox @ train tip 4d841aa1c740bbaa03868dc0a403c62099a99a42 with the 72 sibling PRs of this batch: typecheck:core, file-size, complexity, cognitive-complexity, changelog-integrity green; changed-area node:test 831/831 (0 failing) and vitest 480/482 — the two reds are `autoCombo/provider-family-combos.test.ts` timing out at 20s, which reproduces on the PURE release tip under the full vitest suite (and is already tracked by the Release-Green issue #13866), so it is inherited, not this batch's. Merged --admin per merge-gates §3/§4/§7.

PR #14255, opened 2026-09-20
fix(tests): type the #13848 Gemini turn-pairing assertions — clears the 87-error ESLint base-red on release/v3.8.51

xiaoyaner0201:
## Summary

Clears the **`ESLint errors: 92` hard failure on `release/v3.8.51`** — the longest-standing item on base-red #13866, red on every release-green run since 2026-09-19 03:26.

Root cause analysis in #14254. In short: 87 of the 92 are a single file whose bulk suppression stopped applying, not 87 new defects.

`tests/unit/translator-openai-to-gemini.test.ts` is capped in `config/quality/eslint-suppressions.json` at a frozen `@typescript-eslint/no-explicit-any` count of **74**. #13848 (`c299e6c18`) added 158 lines with **14 further `any` annotations**, taking the live count to 87. ESLint bulk suppressions match by **count**, so exceeding the cap voids the entry for the whole file — the 73 previously-suppressed violations surface alongside the 14 new ones. The file reports `errorCount: 87` with `suppressedMessages: 0`.

```
 5 pre-existing no-unused-vars (untouched here — separate owners)
+87 this file (73 previously suppressed + 14 new)
────
 92
```

## Change

Typed the 14 new annotations using the conventions this file already establishes at the top (`UnknownRecord`, `GeminiRequestWithConfig`, and the `getFunctionCall` / `getFunctionResponse` helpers) — #13848's tests simply did not use them. Added three sibling type aliases in the same style:

```ts
type GeminiPart = {
  text?: unknown;
  functionCall?: { id?: string; name?: string; args?: unknown };
  functionResponse?: { id?: string; name?: string; response?: { result?: unknown } };
};
type GeminiContent = { role?: string; parts?: GeminiPart[] };
type GeminiRequestWithContents = { contents: GeminiContent[] };
```

The `(result as any)` casts become `as GeminiRequestWithContents` / `as GeminiRequestWithThinking`, and the `(c: any)` / `(p: any)` callback parameters infer from there.

Where `any` was silently permitting a possibly-undefined `.functionResponse` deref, the narrowing is now explicit:

```diff
-  const turn1Resp = turn1User.parts.find((p: any) => p.functionResponse);
+  const turn1Resp = turn1User.parts?.find((p) => p.functionResponse);
+  assert.ok(turn1Resp?.functionResponse, "Turn 1 functionResponse part must exist");
```

That is an added assertion, not a relaxed one — it fails loudly where the `any` version would have thrown a `TypeError` on undefined.

[...]

PR #14266, opened 2026-09-20
fix(ci): clear the two TS2677 errors from the API Route Typecheck baseline

HouMinXi:
## Summary

Two `TS2677` errors made API Route Typecheck red on `release/v3.8.51` (tracked in #13866).

`RerankProviderNodeRow` was an interface without an index signature, so it could not be assigned to `Record<string, unknown>` in the type predicate. Converted to a type alias in `src/app/api/v1/_shared/rerankProviderNodes.ts`.

The catalog `.map()` return in `src/mitm/handlers/antigravity.ts` was inferred as `{ id, displayName, description?: string } | null`. Annotated as `DynamicCatalogModel | null` and spread `description` only when present.

`config/quality/eslint-suppressions.json` is not in this PR. #14255 types the Gemini turn-pairing assertions instead of raising the frozen any-count from 74 to 87.

#14164 also edits the two TypeScript files (a `NonNullable<typeof c>` narrow). Whichever lands first, the other rebases.

## Verification

- `node scripts/check/check-api-typecheck.mjs`: OK, 283 pre-existing, 0 new
- antigravity + rerank unit tests: 29/29

Related to #13866. #13866 stays open.


diegosouzapw:
Thanks for tracking down both TS2677 root causes precisely — the `interface` vs `Record<string, unknown>` incompatibility explanation is spot on. Two things before this can merge: (1) `#14255` is also fixing the same `translator-openai-to-gemini.test.ts` suppression entry, but at the root (bringing the file's violations back under the frozen 74 cap) rather than raising the cap to 87 — could you drop the `eslint-suppressions.json` hunk here and let that one own it, so we don't end up with a stale over-suppressed entry (cap 87, real 73) once both land? (2) `#14244`... sorry, `#14164` is also independently fixing `antigravity.ts` and `rerankProviderNodes.ts` (differently — a `NonNullable<typeof c>` narrow instead of the type-alias change). Whichever of the two lands first, the other will need a quick rebase to drop the now-redundant hunk. Once the suppressions.json piece is dropped this is a clean, welcome fix for the base-red.

HouMinXi:
Dropped `config/quality/eslint-suppressions.json` from this PR. #14255 owns the translator-openai-to-gemini any-count (type the assertions, keep the frozen 74). What's left is the two TS2677 edits plus a changelog fragment.

#14164 is still open and also touches `antigravity.ts` / `rerankProviderNodes.ts`. I kept those hunks here. If that PR merges first I will rebase and drop the overlap.

Local: `node scripts/check/check-api-typecheck.mjs` OK (283 frozen, 0 new). antigravity 16/16, rerank 13/13.

HouMinXi:
The two TS2677 errors targeted here were resolved in upstream commit `d610c24c62` (#14164) using the `NonNullable<typeof ...>` narrowing approach in both files.

Verified on the latest `release/v3.8.51` tip: `node scripts/check/check-api-typecheck.mjs` passes cleanly with 0 new errors (283 baseline).

Since #14164 already landed the resolution, this PR is superseded.

diegosouzapw:
Thanks @HouMinXi — the two TS2677 errors this PR cleared from the API Route Typecheck baseline were already resolved on `release/v3.8.51` by #14164 (merged 2026-09-21), so this change no longer has anything to fix on the tip. Closing as superseded; your diagnosis matched the fix that landed. (merge-batch 2026-09-24)

